PR #5374, opened 2022-11-11
chore(ci): add timeout to all ci jobs

natemoo-re:
## Changes

- Adds reasonable timeouts to all jobs in `ci.yml`
- Checked our logs to find roughly average times.
- Prevents infinite loops from clogging CI forever

## Testing

This is the test

## Docs

N/A


jasikpark (approved):
lgtm, hopefully it’s obvious in the log when a job gets timed out

Issue #5373, opened 2022-11-11
Updating `tailwind.config.cjs` Kills The Astro Dev Server


neelsudhakaran:
### What version of `astro` are you using?

1.6.7

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

MacOS Big Sur v11.4

### Describe the Bug

Every time I make changes to the `tailwind.config.cjs` file, the dev server does the following things:

```
08:48:37 AM [astro] update /tailwind.config.cjs
08:48:37 AM [astro] Configuration updated. Restarting...

08:48:37 AM [astro] Continuing with previous valid configuration

$WORKING_DIR/node_modules/vite/dist/node/chunks/dep-51c4f80a.js:57069
          cb(new Error('The server is not running'));
             ^

Error: The server is not running
    at WebSocketServer.<anonymous> ($WORKING_DIR/node_modules/vite/dist/node/chunks/dep-51c4f80a.js:57069:14)
```

`$WORKING_DIR` is simply something I've come up with for this report to represent the root directory of the Astro project. Also, I'm using `@astrojs/tailwind` v2.1.2

### Link to Minimal Reproducible Example

I can't even seem to get the Tailwind plugin set up in CodeSanbox, GitPod, or StackBlitz—but that's due to my lack of experience with those tools.

### Participation

I'm more than happy to help fix this issue, although I may need some guidance on doing so!

neelsudhakaran:
**Update:** When trying to create a brand new minimal project (even on the same local machine), everything seems to work fine... I guess that means its something unrelated to the `astro`, `@astrojs`, and `tailwindcss` dependencies.

neelsudhakaran:
**Update:** Okay, so the error seems to occur if you have `@astrojs/react` and `@astrojs/tailwind` installed at the same time. It occurs regardless of the order they are listed in the `astro.config.ts` file. 

[Here's a git repo with a basic Astro + Tailwind project setup](https://github.com/neelsudhakaran/astro-tailwind-hmr-error). If you start the dev server with `npm start` and then modify the `tailwind.config.cjs` file the dev server will crash.

neelsudhakaran:
**Update:** I'm a bit lost now 😅.  Uninstalling `react`, `react-dom`, and `@astrojs/react`, then deleting `node_modules` and reinstalling everything still results in the same issue—but now the only dependencies are Tailwind related!

matthewp:
[...]

PR #5377, opened 2022-11-11
Use Vite to load non JS astro configs

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/5373
- When using a `astro.config.ts` proload was not able to load `@astrojs/tailwind` on a reload correctly (not sure why).
- Switches to using Vite for all non (m)js configs. 

## Testing

- Restart test added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/core/config/vite-load.ts
+++ b/packages/astro/src/core/config/vite-load.ts
@@ -1,0 +1,74 @@
+import * as vite from 'vite';
+import npath from 'path';
+
+export interface ViteLoader {
+	root: string;
+	viteServer: vite.ViteDevServer;
+}
+
+async function createViteLoader(root: string): Promise<ViteLoader> {
+	const viteServer = await vite.createServer({
+		server: { middlewareMode: true, hmr: false },
+		optimizeDeps: { entries: [] },
+		clearScreen: false,
+		appType: 'custom',
+		ssr: {
+			// NOTE: Vite doesn't externalize linked packages by default. During testing locally,
+			// these dependencies trip up Vite's dev SSR transform. In the future, we should
+			// avoid `vite.createServer` and use `loadConfigFromFile` instead.
+			external: ['@astrojs/tailwind', '@astrojs/mdx', '@astrojs/react']
+		}
+	});
+
+	return {
+		root,
+		viteServer,
+	};
+}
+
+interface TryLoadResult {
+	value: Record<string, any>;
+	filePath?: string;
+}
+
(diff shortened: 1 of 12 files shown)


matthewp:
@preview new-config

matthewp:
!preview new-config

matthewp:
!preview new-config

matthewp:
!preview new-config

packages/astro/src/core/config/vite-load.ts
```diff
@@ -0,0 +1,140 @@
...
+		for(const file of paths) {
+			// First verify the file event exists
+			try {
+				await fs.promises.stat(file);
+			} catch {
+				continue;
+			}
+
+			// Try loading with Node import()
+			if(/\.(m?)js$/.test(file)) {
```
natemoo-re, line 85: `cjs` can be loaded with `import()` as well
```suggestion
			if(/\.[cm]?js$/.test(file)) {
```

natemoo-re (approved):
LGTM! It makes sense to deprecate Proload and remove in v2.

packages/astro/src/core/config/vite-load.ts
```diff
@@ -0,0 +1,141 @@
...
+	let loader: ViteLoader | null = null;
+
+	try {
+		for(const file of paths) {
+			// First verify the file event exists
+			try {
+				await fs.promises.stat(file);
+			} catch {
+				continue;
+			}
```
bluwy, line 83: Should we only search for one Astro config file that exists and load that only? This seems to support multiple Astro config files, and would try to use the others if one config file fails to load (e.g. malformed JS)

packages/astro/src/core/config/vite-load.ts
```diff
@@ -0,0 +1,141 @@
...
+				const res = await load('astro', {
+					mustExist: true,
+					cwd: root,
+					filePath: file,
+				});
+				return {
+					value: res?.value ?? {},
+					filePath: file
+				};
+			} catch {}
```
bluwy, line 121: These load attempt swallows the error which perhaps we want to log out instead? e.g. for malformed JS

I think we should throw and exit too so that we don't startup the server with an empty `{}` astro config?

packages/astro/src/core/config/vite-load.ts
matthewp: Ah yeah, good point, divide this into a search and then a load.

packages/astro/src/core/config/vite-load.ts
matthewp: Yeah I can change this.

bluwy (approved):

Issue #5028, opened 2022-10-08
[@astrojs/image] gif formats are not supported


DJRHails:
### What version of `astro` are you using?

1.4.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Linux

### Describe the Bug

Remote `.gif` files are unsupported.

Local `.gif` files a semi-supported using Sharp, but break in `<Picture/>` components due to avif formatting.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-lmp1qc?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

panwauu:
Related to #4936

You could always set a custom formats option in picture: `formats={['webp']}` than you dont have the problem with `<Picture />` and sharp

From my experience it does always works with GIF if you use sharp and not avif. It does not work with squoosh and avif

DJRHails:
Yes - that's my work around at the moment:

```ts
let formats = [meta.format, "webp"];

if (meta.format !== "gif") {
  formats.push("avif");
}
``` 

But at the very least it could do with throwing an error / meaningful error message rather that a subtle break.

panwauu:
Or maybe the default formats should depend on the input format like you do already 🤔

I totally agree on the error messages! 

avif should support animated content correct?

DJRHails:
Yes - via `.avifs` I believe

matthewp:
Thanks, is anyone willing to submit a PR for this?

panwauu:
I am willing, but not sure if I am capable. I will look into it.

wulinsheng123:
mark,
I will contribute

wulinsheng123:
![image](https://user-images.githubusercontent.com/27261522/201592831-68d74bbe-e74d-4db5-9230-ca4e7c7eb9a0.png)

matthewp:
@bluwy assigning to you since you are reviewing the PR

wulinsheng123:
> it does always works with GIF if you use sharp and not avif.

The GIF setup `formats={['avif']}` then can't work with `sharp`?

wulinsheng123:
> 

hello 
Could your offered `code` run without set `image({ serviceEntryPoint: '@astrojs/image/sharp' })`?

wulinsheng123:
@bluwy I did both methods.

PR #5391, opened 2022-11-14
cant appeared when extension name was gif

wulinsheng123:
## Changes

the image can't appear when the extension name is `gif` because the `avif` was incompatible with `gif`

## Docs

#5028


packages/integrations/image/components/Picture.astro
```diff
@@ -41,7 +41,7 @@ const {
 	fit,
 	background,
 	position,
-	formats = ['avif', 'webp'],
+	formats = ['webp', 'avif'],
```
bluwy, line 44: I think we should only change the defaults when the user pass in a GIF, similar to https://github.com/withastro/astro/issues/5028#issuecomment-1272390635

This also won't close #5028 as you would need sharp to process GIFs, we should document this separately too for GIF handling if it's not possible to support ootb.

bluwy (changes requested):

packages/integrations/image/components/Picture.astro
wulinsheng123: Actually, The GIF can be handled by the `squoosh`.
image
look like this GIF picture was changed into a long picture.

wulinsheng123: @bluwy hello places looking at if you are available

bluwy: It's working for me only when I use sharp. Squoosh gives me `Buffer has an unsupported format` in my vscode terminal, and the network request showed a 500 in dev.

I'm still confused what this fixes here. It would be great if you can write some tests for this to verify it.

wulinsheng123: > It's working for me only when I use sharp. Squoosh gives me `Buffer has an unsupported format` in my vscode terminal, and the network request showed a 500 in dev.
> 
> I'm still confused what this fixes here. It would be great if you can write some tests for this to verify it.

There is not a error with squoosh In my computer

bluwy: here's a repro: https://stackblitz.com/edit/github-lmp1qc-tcfjrz?file=astro.config.mjs,src%2Fpages%2Findex.astro

wulinsheng123: I got it , I replicated this bug ,thanks ,I will fix it

PR #5393, opened 2022-11-14
Log error if failed to fetch remote image

### LINKED ISSUE #5392: I can't catch an error in the `loadRemoteImage` function when the fetch function occurs error

1.6.7

none

pnpm

mac

### Describe the Bug

I can't catch an error in the `loadRemoteImage` function when the fetch function occurs error

https://stackblitz.com/edit/github-lmp1qc?file=src/pages/index.astro

### Participation

wulinsheng123:
## Changes

I can't catch an error in the loadRemoteImage function when the fetch function occurs error

fix #5392


packages/integrations/image/src/endpoint.ts
```diff
@@ -14,7 +14,8 @@ async function loadRemoteImage(src: URL) {
 		}
 
 		return Buffer.from(await res.arrayBuffer());
-	} catch {
+	} catch (err: unknown) {
+		console.error(err)
```
bluwy, line 18: We don't report the full error, but it's currently being handled at 

https://github.com/withastro/astro/blob/eab4a0279fd28163cea9da2b010c5ccde5645e6f/packages/integrations/image/src/endpoint.ts#L35-L37

too. Are you seeing any errors that need this specific log?

bluwy: https://github.com/withastro/astro/issues/5392#issuecomment-1313425804

bluwy, line 18: I think we should do the same at https://github.com/withastro/astro/blob/eab4a0279fd28163cea9da2b010c5ccde5645e6f/packages/integrations/image/src/build/ssg.ts#L55

bluwy (approved):

PR #5398, opened 2022-11-14
WIP: Simplifies @astrojs/image API for a 1.0 release

tony-sull:
> ⚠️  Work in progress!  Using this PR to test a simplified `@astrojs/image` API via tagged preview releases.  If the API tests out, I'll raise this through the RFC process to propose a 1.0 release of the integration

## Changes

All image transformation options are now optional!  No more distinction between local and remote images
    - if no `width` or `height` is included, the image will be optimized but the original dimensions will be kept
    - if only `width` or `height` is included, the image will be resized with the original aspect ratio
    - if `width` and `height` are both included, the image will be cropped
    
**Motivation**

The original design was testing an experimental API that would make sure content layout shift (CLS) is always avoid by requiring `width` and `height` to always be known.  That trade-off didn't end up being worth it, CLS was avoided **but** you always had to know the dimensions needed for an image file - this gets tricky with remote images.

## Testing

Current tests are passing, extra coverage will be needed if the RFC is approved

## Docs

This will definitely require docs work, thankfully a majority of the work should be removing sections explaining the subtle differences between local and remote images 🎉


tony-sull:
!preview image-next

tony-sull:
!preview image-api

PR #5404, opened 2022-11-15
fix(create-astro): fix error when don't have template input

liruifengv:
## Changes

- fix error when don't have template input

## Before
When input `npm create astro@latest -- --template`, `create-astro` will throw error:

image

## After
image


bluwy (approved):

bluwy:
Thanks!

PR #3417, opened 2022-05-20
Fix: support `FormData` object on fetch body

bholmesdev:
## Changes

- Resolves #2741 
- Move `node-fetch` to external dependency for `@astrojs/webapi` bundle. This was causing strange issues with body parsing that I still haven't figured out. Still, since `@astrojs/telemetry` already pulls in `node-fetch` as a standard dependency, this shouldn't be a problem!
- moving away from using `node-fetch/src/index.js` exposed type issues. A few pieces to resolve:
  - `fetch` no longer accepts `URL` as a resource type. Funny enough, you _can_ still pass a `URL` for node-fetch to run `toString()` implicitly. Still, you get red squigglies in VS Code using URL today, so removing this type option is for consistency if anything.
  - `FetchInit` now uses node-fetch's body type in its type signature

### Why not switch to undici like all the cool kids?

@FredKSchott (a cool kid) suggested using [undici](https://www.npmjs.com/package/undici) since it mirrors Node's incoming fetch standard. This seemed possible at first... until I discovered their node versioning requirements. In short: **they[ don't expose `fetch` or `Request` for node versions below 16.5.](https://github.com/nodejs/undici#undicifetchinput-init-promise)** Also poked the source code to confirm. This would mean dropping Node 14.X support, which feels a bit _too_ far as we approach 1.0. Keep it in mind for the future though!

## Testing

N/A

## Docs

N/A

PR #5405, opened 2022-11-15
fix: bring fetch behaviour closer to webapi

okikio:
## Changes

- Brings fetch closer to the webapi by:
  * Adding URL support to fetch
  * Use webapi compatible types (e.g. fixes other type errors)
  * Removes support `node-fetch` peerDeps due to possible polyfill version conflicts
  * Switches`@astrojs/telemetry` from `node-fetch` to the `@astrojs/webapi` fetch polyfill instead (context: https://github.com/withastro/astro/pull/3417)

## Testing

No, new tests required. Current tests should suffice.

## Docs

This isn't particularly user facing, but I guess we could document that Astro's webapi will ignore `node-fetch` version updates


Princesseuh:
Now that we've dropped Node 14 and fully migrated to undici, the work done here is not relevant anymore, so I'll close this!

Thank you for this PR, nonetheless!

Issue #5390, opened 2022-11-14
Astro with tailwind for MPA needs multiple css files


srini-leanfolks:
### What version of `astro` are you using?

1.6.7

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

When using tailwindcss with astro for multiple pages, all the css are converted into one file of ~70kB. Any other ways to split the css for each page separately?

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-csbhoi

### Participation

- [ ] I am willing to submit a pull request for this issue.

mandar1jn:
Addition to this: using @apply doesn’t strip the classes from the tailwind files but just makes it so you have duplicate styling. inline doesn’t work at all with tailwindcss since no transformation is done

bluwy:
I think that's how tailwindcss works. It scans the files for classes and generates them under a single css file with:

```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```

Vite can only code-split when CSS is define in several modules. It's a caveat of tailwindcss unless you use `@apply` in different style tags across pages and components.

srini-leanfolks:
Using ```@apply``` means by having a class name for required elements and having a separate css files for each page and writing styles in it?

mandar1jn:
> Yeah. It's not the greatest and probably not tailwind-ish, but that's the only way to code-split CSS files that I can think of. Otherwise it would need to analyze the CSS, perform selector matching on each HTML files, and try to split it out, but I'm not sure how possible is that.

Actually, using @apply still generates the same tailwind file too. Causing even more code duplication

bluwy:
Oh maybe tailwind's scanner extracts the classes from `@apply` too 😦 That's.. a problem. I guess the only way now is to deal with a single css file. If CSS analysis & code-splitting were to be implemented, it would be in Vite. I'll close this for now as there's not much we can do here.

Issue #5384, opened 2022-11-13
dark theme font not recognized unless specified both in class and html tag


afewell:
### What version of `astro` are you using?

1.6.7

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

npm

### What operating system are you using?

linux

### Describe the Bug

I am using the astro ink theme, and discovered that the dark theme does not apply correctly unless it is specified in the class and also specified in the html tag. 

For example, in the [Header.astro](https://github.com/one-aalam/astro-ink/blob/main/src/components/Header.astro) file, the home page header specifies both the `header_title` class and the `dark:text-theme-dark-secondary`:
```
        <h3 class="header__title dark:text-theme-dark-secondary">
            <a href="">{ SITE.name }</a>
        </h3>
```

This seems strange because in the style section of the same Header.astro file, the `header_title` class also specifies the same dark theme:
```
    .header__title {
        @apply text-4xl font-extrabold md:text-5xl text-theme-secondary dark:text-theme-dark-secondary
    }
```

If i remove the `dark:text-theme-dark-secondary` class from the h3 tag, the dark theme no longer works, it has to be specified in both places. I am not a super expert, but it seems to me it should only need to be specified in the style. 

Also not the astro ink repo package.json specifies packages that are pretty old, so I updated them to the latest, and still have this same issue. Here is my package.json:
```
{
  "name": "astro-ink",
  "version": "1.0.1",
  "private": true,
  "scripts": {
    "dev": "astro dev",
    "build": "astro build",
    "check": "astro check",
    "postbuild": "node ./scripts/search/prepare-index.js",
    "preview": "astro preview"
  },
  "devDependencies": {
    "@astrojs/mdx": "^0.11.1",
    "@astrojs/rss": "^1.0.0",
    "@astrojs/sitemap": "^1.0.0",
    "@astrojs/svelte": "^1.0.0",
    "@astrojs/tailwind": "^2.1.2",
    "@fontsource/fira-sans": "^4.5.9",
    "@tailwindcss/aspect-ratio": "^0.4.0",
    "@tailwindcss/forms": "^0.5.0",
    "@tailwindcss/line-clamp": "^0.4.2",
    "@tailwindcss/typography": "^0.5.2",
    "@types/node": "^18.7.14",
    "astro": "^1.1.5",
    "globby": "^13.1.2",
    "gray-matter": "^4.0.3",
    "lunr": "^2.3.9",
    "mdx": "^0.2.3",
[...]

PR #5406, opened 2022-11-15
Update tailwindcss `@apply` suggestion

bluwy:
## Changes

There has been a couple issues about tailwindcss recently:

- #5384 
- #5390

This PR clarifies that inline classes should be preferred instead of `@apply` as it does not work for classes like `dark:text-pink` and duplicates the global and Astro scoped style output.

Question: Should we mentioned that it's not possible to code-split tailwind CSS output per-page too? All pages can currently only share one stylesheet file.

## Testing

N/A

## Docs

Updated tailwindcss README


matthewp:
Update looks good to me.

packages/integrations/tailwind/README.md
```diff
@@ -127,7 +127,9 @@ export default {
 
 You can now [import your own `base.css` as a local stylesheet](https://docs.astro.build/en/guides/styling/#import-a-local-stylesheet).
 
-If you are using Vue, Svelte, or another component integration with Astro, `@apply` directives used in component `<style>`s may generate errors about your custom Tailwind class not existing and cause your builds to fail. [Instead of using `@layer` directives in a a global stylesheet](https://tailwindcss.com/docs/functions-and-directives#using-apply-with-per-component-css), define your custom styles by adding a plugin to your Tailwind config:
+When using the `@apply` directive in an Astro, Vue, Svelte, or another component integration's `<style>` tag, it may generate errors about your custom Tailwind class not existing and cause your builds to fail. Using nested classes like `dark:text-pink` would also not work in scoped `<style>`.
```
delucis, line 130: Do we have a clearer idea about what exactly triggers these errors? I’ve used `@apply` in `.astro` successfully, e.g. <https://github.com/withastro/astro.build/blob/a4d8460976a36e335f2b0b4289e277b190750e19/src/components/CodeHero.astro#L65-L89>

Is it only nested classes that are an issue?

delucis (commented):
Thanks @bluwy! If I understand correctly, this text and the code sample that follow are not strictly speaking related to the `config.applyBaseStyles` option, correct? We just added these there because we were already talking about `@layer`?

I wonder if we should move this to the “Troubleshooting” section with a dedicated subheading like “Using `@apply` with components”?

packages/integrations/tailwind/README.md
bluwy: If I understand correctly from previous reports, classes in `@apply` are also extracted into the main tailwind css file, so there's two copies of the class. Maybe it's not obvious in a large project where it's used in several places.

bluwy:
> I wonder if we should move this to the “Troubleshooting” section with a dedicated subheading like “Using `@apply` with components”?

Make sense to me. I wonder why the previous docs were there too but this definitely sounds like a troubleshooting guide. Will move it there.

packages/integrations/tailwind/README.md
delucis: Ahh, I see. Sounds like there are actually two (or three?) issues:

1. Some usage triggers “custom Tailwind class not existing” errors, breaking builds.

2. _Any_ usage can result in duplicate CSS, because Tailwind scans the `.astro` file to know which classes to include in the main bundle and is not smart enough to distinguish between a class name in frontmatter, a `class` attribute, etc. or inside `<style>`.

3. Nested classes like `dark:text-pink` are not compatible with `@apply` (is this actually the issue that causes 1?)

I think for **1** it would be amazing if we have more specific guidance than “may generate” if we know which usage is the issue (we may not of course).

For **2**, I think this can probably framed in terms of best practices like you’re doing in the second paragraph you added. Maybe we also state clearly there how you end up with duplicate CSS.

bluwy: Just tested no2 and it's false. Turns out tailwind strips of classes from `@apply` so not sure why someone mentioned it doesn't 😄 Will remove that.

For no1, it seems to be added by https://github.com/withastro/astro/pull/4353 about custom classes.

For no3, it's a separate issue from no1. `dark:text-pink` classes specifically is generated as `.dark > .text-pink`, but this doesn't match any element in the Astro markup so the compiler shakes them out.

bluwy:
Updated the docs and moved the two `@apply` issues under troubleshooting. GitHub's diff isn't showing it right, but I didn't touch the examples section.

tony-sull:
For issue #1, is that a Tailwind issue in general, or does it have something to do with how Tailwind picks up the config file in an astro project?

For #3, I think something like `h1 { @apply dark:text-gray-50; }` should be compiled directly to the code below, with our scoped class added as usual.  Maybe there are certain edge cases where it breaks?

```css
@media (prefers-color-scheme: dark) {
 h1:where(.astro-GFDZHAND) {
  --tw-text-opacity: 1;
  color:rgb(249 250 251 / var(--tw-text-opacity))
 }
}
```

bluwy:
> For issue (1), is that a Tailwind issue in general, or does it have something to do with how Tailwind picks up the config file in an astro project?

It's a tailwind issue in general added in https://github.com/withastro/astro/pull/4353. The tailwind docs also outlined this caveat: https://tailwindcss.com/docs/functions-and-directives#using-apply-with-per-component-css

Yup re no3, it only happens for `darkMode: "class"`

packages/integrations/tailwind/README.md
```diff
@@ -146,14 +164,18 @@ module.exports = {
 }
 ```
 
-## Examples
+### Class-based directives like `dark:text-pink` with `@apply` does not work
```
delucis, line 167: It looks like [the Tailwind docs](https://tailwindcss.com/docs/hover-focus-and-other-states) call `dark:*`, `hover:*` etc. “modifiers”. I’ve also tweaked this heading to match the previous heading format.

```suggestion
### Class with modifiers does not work with `@apply` directives
```

packages/integrations/tailwind/README.md
```diff
@@ -146,14 +164,18 @@ module.exports = {
...
 
-- The [Astro Tailwind Starter](https://github.com/withastro/astro/tree/latest/examples/with-tailwindcss?on=github) gets you up and running with a base for your project that uses Tailwind for styling
-- Astro's homepage uses Tailwind. Check out its [Tailwind configuration file](https://github.com/withastro/astro.build/blob/main/tailwind.config.cjs) or an [example component](https://github.com/withastro/astro.build/blob/main/src/components/IntegrationCard.astro)
-- The [Astro Ink](https://github.com/one-aalam/astro-ink), [Sarissa Blog](https://github.com/iozcelik/SarissaBlogAstroStarter), and [Creek](https://github.com/robertguss/Astro-Theme-Creek) themes use Tailwind for styling
-- [Browse Astro Tailwind projects on GitHub](https://github.com/search?q=%22%40astrojs%2Ftailwind%22+filename%3Apackage.json&type=Code) for more examples!
+Astro `<style>` tags are scoped by default. As `dark:text-pink` compiles into `.dark .text-pink`, it does not match any markup in the Astro file and is hence removed, causing the class to not be applied. This may also happen for integrations like Vue and Svelte that supports scoped styles.
```
delucis, line 169: Reworked this slightly to move the example class down from the subheading.

```suggestion
Class names using modifiers like `dark:text-pink` or `focus:ring` are not compatible with the `@apply` directive in Astro `<style>` tags. For example, `dark:text-pink` compiles to `.dark .text-pink`, which is removed from build output because it does not match any markup in the `.astro` file. The same issue may also happen in framework components that support scoped styles like Vue and Svelte.
```

delucis (approved):
Thanks @bluwy! I think these are really good and provide clear examples of both the problem and what to do instead. I’ve suggested some wording tweaks but then should be good to go 🙌

packages/integrations/tailwind/README.md
bluwy: The issue is that only some modifiers do not work though. `hover:` and `focus:` etc works because they generate `.class:hover`. `dark:` however generates `.dark .class` which is causing problems in scoped styles.

Maybe we should call them class-based modifiers?

delucis: Ahhhh, thank you! I had not understood that nuance.

Then, yes, maybe this heading would be more accurate:

```suggestion
### Class-based modifiers do not work with `@apply` directives
```

packages/integrations/tailwind/README.md
delucis: Given the above, maybe this is more accurate? Are there any other good examples where this is the case? I guess the sample applies with the group classes like `group-hover:*`?

```suggestion
Certain Tailwind classes with modifiers rely on combining classes across multiple elements. For example, `dark:text-pink` compiles to `.dark .text-pink`. When this is used with the `@apply` directive in Astro `<style>` tags, the compiled styles are removed from the build output because they do not match any markup in the `.astro` file. The same issue may also happen in framework components that support scoped styles like Vue and Svelte.
```

bluwy: Yep `group-hover:*` should also have the same issue. I think that's a better example too as `dark:` could be configured to use media queries. I'll swap that out after committing your suggestion

delucis (approved):
Thanks for working through this with me @bluwy!

PR #5409, opened 2022-11-15
Allow Code component to run in Vercel

### LINKED ISSUE #5357: 🐛 BUG: Code Component breaks on Vercel SSR
### What version of `@astrojs/compiler` are you using?

latest

pnpm

mac

### Describe the Bug

The `<Code>` component works in dev, and works in all other build output formats that I tried (static, SSR /w `@astrojs/node`). 

However, when I use `@astrojs/vercel` and deploy, the build is successful but then I get the following error when I request a route that uses the `<Code>` component:

```
2022-11-10T00:14:54.906Z	74074352-f9c1-4006-b3d2-f50c3ddf0e0b	ERROR	Unhandled Promise Rejection 	{"errorType":"Runtime.UnhandledPromiseRejection","errorMessage":"Error: ENOENT: no such file or directory, open '/var/task/node_modules/shiki/themes/github-dark.json'","reason":{"errorType":"Error","errorMessage":"ENOENT: no such file or directory, open '/var/task/node_modules/shiki/themes/github-dark.json'","code":"ENOENT","errno":-2,"syscall":"open","path":"/var/task/node_modules/shiki/themes/github-dark.json","stack":["Error: ENOENT: no such file or directory, open '/var/task/node_modules/shiki/themes/github-dark.json'"]},"promise":{},"stack":["Runtime.UnhandledPromiseRejection: Error: ENOENT: no such file or directory, open '/var/task/node_modules/shiki/themes/github-dark.json'","    at process.<anonymous> (file:///var/runtime/index.mjs:1131:17)","    at process.emit (node:events:539:35)","    at process.emit (node:domain:475:12)","    at emit (node:internal/process/promises:140:20)","    at processPromiseRejections (node:internal/process/promises:274:27)","    at processTicksAndRejections (node:internal/process/task_queues:97:32)"]}
Unknown application error occurred
Runtime.Unknown
```

It looks like this is the same issue as this: https://github.com/code-hike/codehike/issues/283#issuecomment-1264786447

The problem is that Astro/Vite does not detect the dependency on these files, since Shiki uses `fs` to load then instead of `require/import`. Therefore, they never end up in the final build. Next.js had to solve this problem more globally with a workaround called `unstable_includeFiles`, we may want to do the same.

https://gist.github.com/FredKSchott/b875fec67552289e75ebc594499ccc32

matthewp:
## Changes

- The Code component now includes the default languages and themes as part of the bundle.
- This makes it possible to use the Code component in SSR environments that do not support dynamic loading, such as Vercel, Cloudflare, and Deno Deploy.
- Closes #5357
- This adds a couple of scripts to generate the files that import the themes and languages. This is because (1) there are a lot of languages and (2) so we can update without having to do a lot of manual checking.

## Testing

- ❌ Broken site (main): https://vercel-code-test.vercel.app/
- ✅ Working site (next--vercel-code): https://vercel-code-test-matthew6.vercel.app/

## Docs

N/A, bug fix


matthewp:
!preview vercel-code

matthewp:
!preview vercel-code

matthewp:
!preview vercel-code

matthewp:
!preview vercel-code

packages/astro/scripts/shiki-gen-themes.mjs
```diff
@@ -0,0 +1,28 @@
...
+for(const [key, imp] of themeImports) {
+	if(i > 0) {
+		code += ',';
+	}
+	code += `\n\t'${key}': () => ${imp}`;
+	i++;
+}
+code += '\n};';
+
+console.log(code);
```
natemoo-re, line 23: Nit: fix lint warning

packages/astro/scripts/shiki-gen-languages.mjs
```diff
@@ -0,0 +1,40 @@
...
+for(const [key, imp] of langImports) {
+	if(i > 0) {
+		code += ',';
+	}
+	code += `\n\t'${key}': () => ${imp}`;
+	i++;
+}
+code += '\n};';
+
+console.log(code);
```
natemoo-re, line 35: Nit: fix lint warning

natemoo-re (approved):
This seems reasonable, but it is pretty unfortunate that this has the potential to blow up our bundle. Can't really think of a better option, though.

matthewp:
@natemoo-re I don't love it either. If it makes you feel better, these files were all being loaded anyways, they were just being loaded dynamically with `fs` and not through imports.

Also this only affects sites that use Code, it's not pulled in with all Astro usage.

natemoo-re:
That does make me feel better! Seems like the least bad option.

LGTM!

Issue #5145, opened 2022-10-21
HMR: Installing sass when in error state does not reload page


matthewp:
### What version of `astro` are you using?

1.5.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

1. Do not install `sass`.
2. Add `lang="scss"` to a style attribute.
3. You will get an error.
4. `npm install sass`
5. Should reload the browser, does not. Reloading manually fixes it.

### Link to Minimal Reproducible Example

n/a

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
This is blocked by this Node core bug: https://github.com/nodejs/node/issues/44663

bluwy:
I think it would be cool if this works, but I'm not sure if Vite currently works when installing packages on the fly too. Traditionally whenever installing or updating a package, it would be safer to reboot everything.

matthewp:
Vite would work in this case if not for the Node core bug. Another alternative is to launch Astro as a subprocess and to kill that subprocess on any package.json changes. That would be heavy, however.

matthewp:
Another solution being discussed is installing sass when it is attempted to be used, see https://discord.com/channels/804011606160703521/1037805603956998304.

matthewp:
Assigning to @bluwy who's going to update Vite to first check if the package exists before trying to require it.

bluwy:
Ref Vite PR: https://github.com/vitejs/vite/pull/10812. Still needs to be cherry-picked into the Vite v3 branch and released before we can make the change here.

PR #5412, opened 2022-11-15
Restart dev server on package.json changes

matthewp:
## Changes

- Adds the package.json to the list of watch files. We restart the dev container any time one of these files changes.
- This is needed to get `sass` to work when installed after the dev server has already been started.
- This now works in Vite 3.2.4
- Closes https://github.com/withastro/astro/issues/5145

## Testing

- Unit test added which confirms the dev server is restarted when the package.json changes.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/core/config/settings.ts
+++ b/packages/astro/src/core/config/settings.ts
@@ -1,7 +1,7 @@
 import type { AstroConfig, AstroSettings, AstroUserConfig } from '../../@types/astro';
 import { SUPPORTED_MARKDOWN_FILE_EXTENSIONS } from './../constants.js';
 
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 import jsxRenderer from '../../jsx/renderer.js';
 import { createDefaultDevConfig } from './config.js';
 import { loadTSConfig } from './tsconfig.js';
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/config/settings.ts
```diff
@@ -24,9 +24,16 @@ export function createBaseSettings(config: AstroConfig): AstroSettings {
 export function createSettings(config: AstroConfig, cwd?: string): AstroSettings {
 	const tsconfig = loadTSConfig(cwd);
 	const settings = createBaseSettings(config);
+
+	const watchFiles = tsconfig?.exists ? [tsconfig.path, ...tsconfig.extendedPaths] : [];
+
+	if(cwd) {
+		watchFiles.push(fileURLToPath(new URL('./package.json', pathToFileURL(cwd))));
+	}
```
natemoo-re, line 32: This seems like an odd edge case, but should we check if `package.json` exists first?

Since Deno is landing `npm:` support, I think we're not far off from an Astro project that doesn't have a `package.json` file being a feasible thing.

natemoo-re (approved):
LGTM! One comment about being a bit more defensive.

packages/astro/src/core/config/settings.ts
matthewp: We don't need to check for that. This is used that when a file does change and it matches one of the files in `watchFiles`, it will trigger a restart. If the file doesn't exist then it will never match. It doesn't cause some other process to try and read the (potentially not existing) files.

Here's the only place this information is used: https://github.com/withastro/astro/blob/7191ed1c7c80c9d70d7d6f88e40ebc9a89787411/packages/astro/src/core/dev/restart.ts#L55-L57

Issue #5415, opened 2022-11-16
Hydrating dynamic framework components throws "Unable to resolve a valid export" error


paulwag:
### What version of `astro` are you using?

1.6.8

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux Ubuntu

### Describe the Bug

Using dynamic framework components brakes with hydration (see minimal reproduction). 

What works:
- import and then hydrate framework components
- assign components to variable and render (no hydration)

What fails:
- assigning a component to a variable and then hydrating it

_Use case for reference:_
Looping over a list of content parts and render components dynamically based on the type of the part.
This works great without hydration, but fails with error `Unable to resolve a valid export for "Component"!` on hydration.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-sewhq2?file=src%2Fpages%2Findex.astro,src%2Fcomponents%2FVueComponent.vue,src%2Fcomponents%2FSvelteComponent.svelte,package.json&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
This is a known limitation. The way that Astro knows which code to load in the browser is by matching up the component being used to which import statement it comes from. Reassigning breaks that connection so we don't know what to do.

paulwag:
Alright thanks for the quick info :)

SarcevicAntonio:
> This is a known limitation.

ahh bummer, i was working on something were this would be very handy.
any plans to support this in the future?

PR #5417, opened 2022-11-16
Prevent dev from crashing when there are errors in template

matthewp:
## Changes

- When an error occurs in a template only throw the first error that occurs. This is because latter errors can't be caught by the dev server. The user will see the first error in the error overlap (see below)
- Fixes https://github.com/withastro/astro/issues/5415

Screen Shot 2022-11-16 at 11 48 07 AM

## Testing

- unit test added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/wild-falcons-sparkle.md
+++ b/.changeset/wild-falcons-sparkle.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Makes sense, nice `catch`!

PR #5418, opened 2022-11-16
Add response iterator to node adapter

jbanety:
## Changes

Sometimes Astro sends a [ReadableStream](https://github.com/withastro/astro/blob/main/packages/astro/src/runtime/server/render/page.ts#L105) or a Buffer as a response body and it raise an error `TypeError: body is not async iterable.`
I added a function to get a response iterator from different response types (sourced from apollo-client).
With this, node adapter can handle all the Astro response types.

## Testing

Not implemented. I tested this in my project.

## Docs

N/A

### DIFF
--- a/.changeset/blue-rabbits-hear.md
+++ b/.changeset/blue-rabbits-hear.md
@@ -1,0 +1,9 @@
+---
+'@astrojs/node': minor
+---
+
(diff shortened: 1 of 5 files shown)


matthewp (approved):
thanks!

PR #5421, opened 2022-11-16
feat: provide HOST env variable at runtime

Scttpr:
## Changes

Provides a way to handle runtime HOST environment variable on Node Standalone builds

## Testing

I did not see any test for this section, should I write some ?

## Docs

Hi! I think this could be documented next to the host definition mechanism, I can participate if needed!

### DIFF
--- a/packages/integrations/node/README.md
+++ b/packages/integrations/node/README.md
@@ -109,6 +109,16 @@ node ./dist/server/entry.mjs
 
 For standalone mode the server handles file servering in addition to the page and API routes.
 
+
+#### Runtime HOST and PORT
+
+It is possible to override built HOST and PORT values with [runtime environment variables](https://docs.astro.
+build/en/guides/environment-variables/#using-the-cli)
+
+```shell
+HOST=0.0.0.0 PORT=3000 node ./dist/server/entry.mjs
+```
+
 #### HTTPS
 
 By default the standalone server uses HTTP. This works well if you have a proxy server in front of it that does HTTPS. If you need the standalone server to run HTTPS itself you need to provide your SSL key and certificate.
--- a/packages/integrations/node/README.md
+++ b/packages/integrations/node/README.md
@@ -110,10 +110,9 @@ node ./dist/server/entry.mjs
 For standalone mode the server handles file servering in addition to the page and API routes.
 
 
-#### Runtime HOST and PORT
+#### Custom host and port
 
-It is possible to override built HOST and PORT values with [runtime environment variables](https://docs.astro.
-build/en/guides/environment-variables/#using-the-cli)
+You can override the host and port the standalone server runs on by passing them as environment variables at runtime:
 
 ```shell
 HOST=0.0.0.0 PORT=3000 node ./dist/server/entry.mjs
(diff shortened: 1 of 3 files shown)


matthewp:
Change sounds good. Yes, please do update the readme to explain this. And thank you!

packages/integrations/node/src/standalone.ts
```diff
@@ -39,7 +39,8 @@ export default function startServer(app: NodeApp, options: Options) {
 	const { client } = resolvePaths(options);
 	const handler = middleware(app);
 
-	const host = getResolvedHostForHttpServer(options.host);
+	// Allow to provide host value at runtime
+	const host = getResolvedHostForHttpServer(process.env.HOST !== undefined ? process.env.HOST : options.host);
```
rishi-raj-jain, line 43: I think an empty string for HOST would still return true, would it make sense to update it to process.env.HOST ? process.env.HOST : options.host, instead?

Scttpr: I took some time to think about the logic but we have also the `false` case which is falsy and would skip the runtime variable

Scttpr: I can manage the condition with better edge cases management:
- `undefined` or empty string would mean no runtime variable
- else we have a value (`true`, `false`, `string`)

Scttpr: What do you think ?

Scttpr:
A job is failing but I'm not sure it is linked to my contribution, could you double check it please ?

Scttpr:
> Change sounds good. Yes, please do update the readme to explain this. And thank you!

Done, I was not sure about where to write it :p

packages/integrations/node/README.md
```diff
@@ -109,6 +109,16 @@ node ./dist/server/entry.mjs
 
 For standalone mode the server handles file servering in addition to the page and API routes.
 
+
+#### Runtime HOST and PORT
+
+It is possible to override built HOST and PORT values with [runtime environment variables](https://docs.astro.
+build/en/guides/environment-variables/#using-the-cli)
```
delucis, line 116: Thanks @Scttpr! I think this is a good place for this.

Made a couple of tweaks here. I don’t think we need all-caps when referring to the concept of host/port. I’m also not sure the link is that helpful as that section of the docs describes running the Astro dev server. How does this look?

```suggestion
#### Custom host and port

You can override the host and port the standalone server runs on by passing them as environment variables at runtime:
```

Scttpr: You are totally right!

packages/integrations/node/src/standalone.ts
rishi-raj-jain: > I can manage the condition with better edge cases management:
> 
> * `undefined` or empty string would mean no runtime variable
> * else we have a value (`true`, `false`, `string`)

@Scttpr 
That looks ideal!

Scttpr: > > I can manage the condition with better edge cases management:
> > 
> > * `undefined` or empty string would mean no runtime variable
> > * else we have a value (`true`, `false`, `string`)
> 
> @Scttpr That looks ideal!

Fixed, don't hesitate if it is not fine :)

rishi-raj-jain: @Scttpr Looking all set!

Issue #5424, opened 2022-11-17
Astro Partytown not resolved correctly with base path config in SSR node standalone


harrytran998:
### What version of `astro` are you using?

lastest

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

yarn 3

### What operating system are you using?

Mac

### Describe the Bug

```js
export default defineConfig({
  output: 'server',
  base: '/ok',
  server: {
    port: 3000,
    host: '0.0.0.0',
  },
  adapter: node({
    mode: 'standalone',
  }),
  integrations: [
     ...,
    AstroPartytown({ 
      config: { debug: true, forward: ['dataLayer.push'] }
    }),
],
})
```

When running dev or server mode --> Display this error in terminal

`[serve]    404        /~partytown/debug/partytown-sw.js`

And display this error in the browser

`TypeError: Failed to register a ServiceWorker for scope ('http://localhost:3000/ok/~partytown/debug/') with script ('http://localhost:3000/ok/~partytown/debug/partytown-sw.js'): A bad HTTP response code (404) was received when fetching the script.`

### Link to Minimal Reproducible Example

This link stackblitz https://stackblitz.com/github/harrytran998/reproduction-astro-base-path?file=package.json  

This link to Github if case Stackblitz does not work https://github.com/harrytran998/reproduction-astro-base-path

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
fixed by #5425

harrytran998:
@matthewp Do you have the plan to release this 😁?

PR #5425, opened 2022-11-17
fix: partytown base path

brenelz:
Closes issue https://github.com/withastro/astro/issues/5424

## Changes

When serving up the partytown-sandbox-sw.js it is already scoped to the base path. The code that matches the request url is at the following.

https://github.com/withastro/astro/blob/main/packages/integrations/partytown/src/sirv.ts#L209

## Testing

I was able to reproduce the problem locally and then tinkered around a bit until I figured out the problem. Might need some more tesing since I am still very new.

## Docs

Shouldn't have any effect on the docs. It is just a fix with the expected behavior.

### DIFF
--- a/.changeset/popular-melons-pretend.md
+++ b/.changeset/popular-melons-pretend.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/partytown': patch
+---
+
+fix partytown when base path specified
--- a/packages/integrations/partytown/src/index.ts
+++ b/packages/integrations/partytown/src/index.ts
@@ -36,7 +36,7 @@ export default function createPlugin(options: PartytownOptions): AstroIntegratio
 				config = _config;
 			},
 			'astro:server:setup': ({ server }) => {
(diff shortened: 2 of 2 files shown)


matthewp:
Thanks @brenelz! Can you add a changeset? `pnpm changeset`. This would be a patch I believe.

brenelz:
> Thanks @brenelz! Can you add a changeset? `pnpm changeset`. This would be a patch I believe.

Okay I've added one. Hopefully I did it right

joshuaiz:
Still seeing the issue mentioned in #5424:

```
bad HTTP response code (404) was received when fetching the script.
TypeError: Failed to register a ServiceWorker for scope ('http://localhost:5555/~partytown/debug/') with script ('http://localhost:5555/~partytown/debug/partytown-sw.js'): A bad HTTP response code (404) was received when fetching the script.
```

Is this something else?

dsacramone:
@joshuaiz Did you solve this, I am seeing this exact error now....

Issue #5428, opened 2022-11-17
Fail to import TypeScript integration since Astro 1.6.9


fviolette:
### What version of `astro` are you using?

1.6.9

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

Prior to 1.6.9, I could import and use .ts integrations (e.g. Astro asides); now I'm getting this : `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for <PATH>/integrations/astro-asides.ts

I don't use TypeScript elsewhere in my project; here's tsconfig.json:

{
  "compilerOptions": {
    "allowJs": true,
    "baseUrl": ".",
    "target": "es2020",
    "module": "esnext",
    "moduleResolution": "node",
    "jsx": "preserve",
    "paths": {
      "~/*": ["src/*"]
    },
  },
}

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-pf1rr5-nhssod

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #5431, opened 2022-11-17
Fix regression with loading .ts in .mjs config

matthewp:
## Changes

- Previously if importing a `astro.config.mjs` via dynamic import failed, we just failed completely.
- Didn't realize that people were relying on being able to import non-JS in these types of configs.
- Change is to fallback to the Vite flow.
- Closes https://github.com/withastro/astro/issues/5428

## Testing

- Test added
- Had to clean up the mock fs fallback plugin a bit.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/purple-tips-flash.md
+++ b/.changeset/purple-tips-flash.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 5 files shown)


bluwy (approved):
Missed the try catch 🙈

Issue #5167, opened 2022-10-22
imported css file classes don't get loaded on SSG build when using client:only=react


davidvpe:
### What version of `astro` are you using?

1.5.2

### Are you using an SSR adapter? If so, which one?

None, just SSG

### What package manager are you using?

npm v8.15.0 & node v16.17.0

### What operating system are you using?

macOS Monterrey 12.6

### Describe the Bug

When styles are added using local css files on client:only="react" components, styles don't get loaded into the page when building the page and using preview `npm run build && npm run preview` which more or less replicates my production deployment situation, using an S3 bucket and CloudFront on AWS. This doesn't affect other client: directives nor does it when using they "styles" prop on the html component or tailwind classes. This just happens when css files are being imported on cliend:only="react" components (not sure about other integrations though),

These are the steps to reproduce:

I've created a react component `CoolComponent.tsx`

```css
.first {
  background-color: red;
  width: 100px;
  height: 100px;
}

.second {
  background-color: blue;
  width: 50px;
  height: 50px;
}

```

```tsx
import React from 'react'
import styles from './styles.module.css' // also happens when using styles.css
type Props = {}

const CoolComponent = (props: Props) => {
  return (
    <div>
      <h1>Cool Component</h1>
      <div className={styles.first}>
        <div className={styles.second}></div>
      </div>
    </div>
  )
}

export default CoolComponent
```

and I am importing it on the index page

```astro
---
import Layout from "../layouts/Layout.astro";
import CoolComponent from "../components/CoolComponent";
---

<Layout title="Welcome to Astro.">
  <CoolComponent client:only="react" />
</Layout>
```
Since I'm using SSG, and the uploading it to an S3 Bucket I need to run `npm run build` and then I upload all the files to the bucket

I notice that there is a file that contains the css styles for Cool Component

<img width="273" alt="image" src="https://user-images.githubusercontent.com/1934159/197364573-b13d04f6-01d1-49c7-a96e-5de14cb94dcb.png">

But on the actual page that file never gets imported

Note: I just realised that I can replicate this bug result using `npm run preview` that way I don't need to upload it to an s3 bucket. 

Expected result:
![image](https://user-images.githubusercontent.com/1934159/197364528-c6e8ebda-527f-4836-bc05-20fd90112812.png)

Actual result:
![image](https://user-images.githubusercontent.com/1934159/197364598-1eecc9f5-062b-45fc-855e-06747aaf1060.png)
[...]

PR #5434, opened 2022-11-17
Use Vite's resolve to resolve paths for client:only

matthewp:
## Changes

- Use Vite (Rollup) `resolve` method to resolve within the compiler.
- This is just used for client:only resolution.
- Fixes case where importing a tsx component without the extension.
- Fixes https://github.com/withastro/astro/issues/5167

## Testing

- New test case added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/pink-cheetahs-heal.md
+++ b/.changeset/pink-cheetahs-heal.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Use Vite's resolve to resolve paths for client:only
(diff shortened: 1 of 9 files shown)


bluwy:
It doesn't use Vite resolve because it doesn't need to when compiling. We only need to make sure that path is import-able when the astro-island imports it. 

I'm not sure if this PR is the right move here. I think we might need to fix how vite-plugin-css links the styles up, maybe the missing extension is causing incorrect checks.

matthewp:
@bluwy Yeah, you're right. I removed that and instead made it so that the analyzer always fully resolves via the specifier to keep track of these.

tony-sull (approved):
let Vite handle the resolve so we don't worry about extensions, makes sense!

PR #5436, opened 2022-11-18
Add dark mode support for blog template

pavelkeyzik:
## Changes

- I've updated styles to support dark mode. I took colors from https://bearblog.dev

https://user-images.githubusercontent.com/17102399/202660193-3383d912-d34d-4e58-b919-b4a2ee21f37a.mp4

## Testing

I just reviewed all pages and updated colors, you can check them as well in a video I attached above

## Docs

- It'll affect users who use Dark Mode system preferences, now they'll have dark mode version of website. Do we need to update any docs for that? 🤔 

### DIFF
--- a/examples/blog/README.md
+++ b/examples/blog/README.md
@@ -20,6 +20,7 @@ Features:
 - ✅ Sitemap support
 - ✅ RSS Feed support
 - ✅ Markdown & MDX support
+- ✅ Dark and Light Mode
 
 ## 🚀 Project Structure
 
--- a/examples/blog/src/pages/blog.astro
+++ b/examples/blog/src/pages/blog.astro
@@ -25,7 +25,6 @@ const posts = (await Astro.glob('./blog/*.{md,mdx}')).sort(
 			ul li time {
 				flex: 0 0 130px;
 				font-style: italic;
-				color: #595959;
 			}
 			ul li a:visited {
 				color: #8e32dc;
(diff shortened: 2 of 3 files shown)


natemoo-re:
Hey @pavelkeyzik, thanks for the PR! I think dark mode was an intentional omission from this theme, since it's meant to be a minimal starting point. We didn't want users to accidentally deploy without even realizing there was a dark mode theme they forgot to update!

I personally think this is a great addition that's safe to add, but I'll double check with the rest of the team!

pavelkeyzik:
@natemoo-re Thanks for letting me know that. I just tried to use this template and needed dark mode as well, and I thought it might be useful for someone as well, so I created a PR. But I totally understand why you might wanted to ignore that, because sometimes user might want to disable dark mode for specific website, that's why this template might need toggle button as well, and so many things, that's not the purpose of the minimal template and I totally get that.

Let me know what's the team decision, I'm okay with both cases 😊

natemoo-re:
Hey @pavelkeyzik, after discussing with the team, we decided to keep this starter as minimal as possible and let everyone implement their own dark mode according to their specific needs.

We do really appreciate the PR, even though we weren't able to merge this one!

Side note: this conversation inspired me to build a little demo of how users could implement [dark mode](https://github.com/natemoo-re/astro-theme) without a flash of inaccurate theme—I might end up releasing that as a package that anyone can use!

pavelkeyzik:
Thank you! I'm happy to hear that 😊

PR #4986, opened 2022-10-05
Add new fields to API route context

bluwy:
## Changes

Add new `site`, `generator`, `url`, `clientAddress`, `props` and `redirect` fields to API route context to match closer to `Astro` global API.

Support second optional parameter `status` for `redirect()` (both api and Astro global).

Also fixes: https://github.com/withastro/astro/issues/4920

Refactored internally to share types between Astro global too.

## Testing

Added a test in ssr-api-routes.

## Docs

Docs: https://github.com/withastro/docs/pull/1731

Added links in jsdoc referencing the new docs.

PR #5446, opened 2022-11-21
Allow all redirection status codes in redirect()

jyasskin:
## Changes

This only changes the types file, not the implementation. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages for the list of redirection status codes.

## Testing

Tested manually.

## Docs

No docs; the restriction wasn't documented before.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1187,7 +1187,7 @@ interface AstroSharedContext<Props extends Record<string, any> = Record<string, 
 	/**
 	 * Redirect to another page (**SSR Only**).
 	 */
-	redirect(path: string, status?: 301 | 302 | 308): Response;
+	redirect(path: string, status?: 301 | 302 | 303 | 307 | 308): Response;
 }
 
 export interface APIContext<Props extends Record<string, any> = Record<string, any>>
--- a/.changeset/chilly-pumpkins-fetch.md
+++ b/.changeset/chilly-pumpkins-fetch.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix redirect() typing to allow all redirection status codes.


packages/astro/src/@types/astro.ts
```diff
@@ -1187,7 +1187,7 @@ interface AstroSharedContext<Props extends Record<string, any> = Record<string,
 	/**
 	 * Redirect to another page (**SSR Only**).
 	 */
-	redirect(path: string, status?: 301 | 302 | 308): Response;
+	redirect(path: string, status?: 301 | 302 | 303 | 307 | 308): Response;
```
bluwy, line 1190: This makes sense to me. Maybe we could add `300` and `304` here too? Here's the [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) for convenience.

We chose only the few status in https://github.com/withastro/astro/pull/4986#discussion_r991544356. Might be good to get thoughts from @matthewp  in case I miss something.

jyasskin: 300 and 304 don't really make sense with the rest of this signature. The signature allows callers to provide a single URL which fills in the `Location:` header, but:
* [300](https://httpwg.org/specs/rfc9110.html#status.300) says "the server SHOULD generate content in the 300 response containing a list of representation metadata and URI reference(s) from which the user or user agent can choose the one most preferred", which needs more than just a URL, and
* [304](https://httpwg.org/specs/rfc9110.html#status.304) doesn't need a `Location` at all, since it says that the client already has the resource cached.

I'd certainly rather get too many options than too few; just wanted to mention that those don't fit well.

jyasskin:
Sorry, did you want me to add 300 and 304, or to leave them out?

packages/astro/src/@types/astro.ts
bluwy: Sorry I haven't been catching up on the PR reviews. Thanks for the explanation! I think we can put 300 and 304 aside for now until someone asks for it with a reason. Let's keep it as-is.

jyasskin:
Thanks!

Issue #5439, opened 2022-11-19
temp tailwind.config.js file not removed.


chenxsan:
### What version of `astro` are you using?

1.6.10

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

Making any syntax error in `tailwind.config.cjs`, you'll see a `.temp.xxxx.tailwindcss.config.cjs` generated. Fixing that error in `tailwind.config.cjs` won't clean up that temp file.

<img width="515" alt="" src="https://user-images.githubusercontent.com/1091472/202842227-e9efe71c-ccaa-4c10-aebe-ab5a9d08fbdc.png">

I think we need to `unlink` that temp file in both `try` and `catch` here https://github.com/withastro/astro/blob/4b188132ef68f8d9951cec86418ef50bb4df4a96/packages/integrations/tailwind/src/index.ts#L51

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-1gh3v5?file=tailwind.config.cjs&on=stackblitz

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
@chenxsan Yes, thank you! Probably this should be done in the `finally` (maybe in its own try/catch in case it fails). Would you be willing to submit a PR for this?

PR #5450, opened 2022-11-21
Remove temp tailwind config file in finally

chenxsan:
## Changes

Closes https://github.com/withastro/astro/issues/5439

## Testing

Tested in local project.

## Docs

N/A


packages/integrations/tailwind/src/index.ts
```diff
@@ -47,10 +47,10 @@ async function getUserConfig(root: URL, configPath?: string, isRestart = false)
 				cwd: resolvedRoot,
 				filePath: tempConfigPath,
 			});
-
+		} catch (err) {
+			console.error(err)
```
chenxsan, line 51: I found myself checking the terminal more often than the editor for errors, so would like to print the error here.

matthewp (approved):
Thanks!

PR #5455, opened 2022-11-22
Integration READMEs code block sweep 🧹

delucis:
## Changes

Quick sweep through integration READMEs to fix up some code blocks

Includes:

- Fix an unclosed code block in Tailwind README.
- Move titles to comments instead of custom `title=""` attribute to make that context available on GitHub/npm (including updating any line number annotations to take this into account).
- Rework Netlify example to avoid reliance on `ins`/`del` annotations that won’t work outside of the docs site.
- Rework a Deno example for the same reason.
- A couple other small fixes.

## Testing

Docs only.

## Docs

Docs only /cc @withastro/maintainers-docs for feedback!

### DIFF
--- a/packages/integrations/alpinejs/README.md
+++ b/packages/integrations/alpinejs/README.md
@@ -44,7 +44,8 @@ npm install alpinejs @types/alpinejs
 Then, apply this integration to your `astro.config.*` file using the `integrations` property:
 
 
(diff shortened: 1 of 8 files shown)


sarah11918:
These all look great, @delucis, thanks!

Issue #5343, opened 2022-11-10
Stale CSS is displayed after modifying .astro files located outside of the project root


jamesknelson:
### What version of `astro` are you using?

1.6.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

When an Astro app loads CSS from a <style> tag in an `.astro` file located *outside* of the project root, then while the initially rendered styles will be correct, modifying the `.astro` file will cause stale CSS (but up-to-date markup) to be displayed. 

This occurs upon changes to both the CSS *and* the markup:

- If the contents of the `<style>` tag are modified (e.g. in the linked reproduction, changing one of the background colors), then the changes will not be reflected until the server is reloaded.
- If the markup is modified, the rendered markup will be updated - causing any and all scoped styles to *immediately disappear*.

Upon some investigation:

- Modifying the markup will render a new the scope hash in CSS classes.
- But the rendered CSS is static, even after updates. It continues to use the initial styles and scope hash.

## Motivation

External .astro imports already *mostly* work, and are used in popular packages such as astro-icon. It's already possible to use them, but development requires reloading the webserver after each change. Having working HMR for external packages too would markedly improve development velocity.

In particular, separating shared .astro files into separate packages allow for standard "theme" packages that can be shared between blogs (or other types of websites). For example, see the linked minimal reproduction monorepo.

### Link to Minimal Reproducible Example

https://github.com/jamesknelson/astro-stale-external-css-minimal-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #5460, opened 2022-11-23
Fix linked Astro library style HMR

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5343

Remove `/@fs` when resolving third-party Astro component style virtual module. Ideally we can pass it through for Vite's resolver to resolve itself, but I found that still causes other HMR problems. So sticking with this simple fix for now.

We could revisit this again and fix as a whole, but this bug alone took quite a while to figure out ~~that I almost considerd retiring~~ so putting it aside for now.

## Testing

Added new test in `astro-component`

## Docs

N/A. bug fix

### DIFF
--- a/.changeset/ten-emus-raise.md
+++ b/.changeset/ten-emus-raise.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix linked Astro library style HMR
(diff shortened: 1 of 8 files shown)


bluwy:
Windows strikes again. Marking as draft.

PR #5478, opened 2022-11-25
Updated according to new configuration

nemo0:
Astro imports the `defineConfig` function from `astro/config`. The `integrations` key needs to be passed into the `defineConfig` function, but it is not shown in the README. Updated the README according to the CLI example.

## Changes

- The `astro.config.mjs` example shown in the README is changed

### DIFF
--- a/packages/integrations/image/README.md
+++ b/packages/integrations/image/README.md
@@ -455,12 +455,12 @@ __`astro.config.mjs`__
 import { defineConfig } from 'astro/config';
 import image from '@astrojs/image';
 
-export default {
+export default defineConfig({
   integrations: [image({
     // Example: The entrypoint for a third-party image service installed from NPM
     serviceEntryPoint: 'my-image-service/astro.js'
   })],
-}
+});
 ```
 
 ### config.logLevel
@@ -479,7 +479,7 @@ export default defineConfig({
     // default: 'info'
     logLevel: 'debug'
   })],
-})
+});
 ```
 
 ### config.cacheDir
--- a/packages/integrations/image/README.md
+++ b/packages/integrations/image/README.md
@@ -63,13 +63,14 @@ Then, apply this integration to your `astro.config.*` file using the `integratio
 
 __`astro.config.mjs`__
 
-```js
+```js ins={2} "image()"
+import { defineConfig } from 'astro/config';
 import image from '@astrojs/image';
 
-export default {
+export default defineConfig({
   // ...
   integrations: [image()],
-}
+});
 ```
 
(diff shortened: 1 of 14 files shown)


packages/integrations/solid/README.md
```diff
@@ -44,12 +44,13 @@ Now, apply this integration to your `astro.config.*` file using the `integration
...
 import solid from '@astrojs/solid-js';
 
-export default {
-  // ...
-  integrations: [solid()],
-}
+export default defineConfig({
+   // ...
+   integrations: [solid()],
+});
```
bluwy, line 53: Looks like there are other readmes like svelte, turbolinks, react, lit, image, etc that could use this update too to be consistent. Also maybe we should keep the indent two spaces instead of three.

delucis: Agree with @bluwy — a good change to make across all READMEs using the outdated config style. And +1 on indenting by two spaces.

matthewp:
@nemo0 can you update the other readmes?

packages/integrations/solid/README.md
natemoo-re: I just went through and updated all of these!

sarah11918 (approved):
LGTM, @natemoo-re!

packages/integrations/image/README.md
```diff
@@ -446,8 +449,10 @@ The integration can be configured to run with a different image service, either
...
 The `serviceEntryPoint` should resolve to the image service installed from NPM. The default entry point is `@astrojs/image/squoosh`, which resolves to the entry point exported from this integration's `package.json`.
 
+__`astro.config.mjs`__
+
 ```js
-// astro.config.mjs
 import image from '@astrojs/image';
 
 export default {
```
MoustaphaDev, line 458: This one was left out

natemoo-re: Great catch, thank you!

MoustaphaDev:
Last occurence here
https://github.com/nemo0/astro/blob/10f518cf2234421cc3fb5fc104b99cd4b3000b76/packages/integrations/svelte/README.md#L102

MoustaphaDev (approved):

sarah11918:
Just a heads up that because this PR was opened so long ago, one of its original pages was outdated. Since it was accepted into the main branch for its minor changes to the config example, it brought back with it the "old" page.

And, this merging was done after I had looked at the changes and approved. I'm not sure how we avoid this from happening again, but maybe when addressing a super-old PR we make sure the author first updates their branch?

cc @natemoo-re

PR #5479, opened 2022-11-26
Improve README install guide

xyntechx:
## Changes

Separates the recommended and manual install commands for ease of copying.

## Testing

No tests were added as only the README was changed.

## Docs

No docs were updated as only the README was changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,15 @@
 
 ## Install
 
+The **recommended** way to install the latest version of Astro is by running the command below!
 
 ```bash
-# Recommended!
 npm create astro@latest
+```
 
-# Manual:
+However, you can also install Astro **manually** by running this command instead.
+
+```bash
 npm install --save-dev astro
 ```
 
--- a/README.md
+++ b/README.md
@@ -9,13 +9,13 @@
 
 ## Install
 
-The **recommended** way to install the latest version of Astro is by running the command below!
+The **recommended** way to install the latest version of Astro is by running the command below:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,12 +9,15 @@
 
 ## Install
 
+The **recommended** way to install the latest version of Astro is by running the command below!
```
sarah11918, line 12:
```suggestion
The **recommended** way to install the latest version of Astro is by running the command below:
```

README.md
```diff
@@ -9,12 +9,15 @@
...
 
 ```bash
-# Recommended!
 npm create astro@latest
+```
+
+However, you can also install Astro **manually** by running this command instead.
```
sarah11918, line 18:
```suggestion
You can also install Astro **manually** by running this command instead:
```

sarah11918 (commented):
Thanks for this, @xyntechx!  We do like keeping our instructions short and sweet, especially when docs has all the details. But, you are absolutely right that when the commands are separate, they are easier to one-click copy.

The only change I'd suggest is that in long, sentence form, we probably don't need the exclamation point anymore!  Now it feels like we're shouting! 😅  With the punctuation changes, Team Docs is officially happy with this PR.

xyntechx:
@sarah11918 thanks for the feedback! I've made and pushed the requested changes to this PR.

sarah11918 (approved):
Looks Good To Me!  Thank you for this, @xyntechx!  🥳

PR #5480, opened 2022-11-26
Fix astro preview not working on Windows

sapphi-red:
## Changes

When running `astro preview` on Windows, it didn't work with the following error.

> Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'

This PR fixes this error.

## Testing

I added a test. Also I tested with adapter-node manually.

## Docs

Docs isn't needed as it's a bug fix.

### DIFF
--- a/.changeset/swift-papayas-kiss.md
+++ b/.changeset/swift-papayas-kiss.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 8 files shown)


bluwy (approved):
Nice tests!

PR #5482, opened 2022-11-26
Documented how to access the base config in the app

alvinometric:
## Changes

While the `site` property in the Astro config can be accessed in the app via `Astro.site`, the `base` property is only accessible via `import.meta.env.BASE_URL`. It wasn't documented, so this PRs fixes this.

## Testing

No tests since it's just a docs change.

## Docs

This updates the types that are used to generate the reference pages (https://docs.astro.build/en/reference/).

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -432,6 +432,8 @@ export interface AstroUserConfig {
 	 * @description
 	 * The base path to deploy to. Astro will build your pages and assets using this path as the root. Currently, this has no effect during development.
 	 *
+	 * You can access this value in your app via `import.meta.env.BASE_URL`.
+	 *
 	 * ```js
 	 * {
 	 *   base: '/docs'


packages/astro/src/@types/astro.ts
```diff
@@ -432,6 +432,8 @@ export interface AstroUserConfig {
 	 * @description
 	 * The base path to deploy to. Astro will build your pages and assets using this path as the root. Currently, this has no effect during development.
 	 *
+	 * You can access this value in your app via `import.meta.env.BASE_URL`.
```
aFuzzyBear, line 435:
```suggestion
	 * You can alternatively access this value in your app via `import.meta.env.BASE_URL`.
```

aFuzzyBear: Im sure this would help users understand that the `import.meta.env` variables are available to them for using outside the scope of Astro files.

aFuzzyBear (commented):
Ill ask @sarah11918 to come take a wee look at this for us @alvinometric, Thank you for putting this in for us 👊

sarah11918:
Thanks for contributing this, @alvinometric! I'm taking a look around the site, and there are two things that come up for me:

1. we *do* already document this with the rest of the environment variables here: https://docs.astro.build/en/guides/environment-variables/#default-environment-variables

image


2. In the place you've PR'd this change, we *don't* describe how to access the other settings, like `site` via `Astro.site` directly above it. So, I'm not sure that this would be the place to add that information?

Can you describe the problem with the current documentation with the other environment variables, so that I know what we're trying to solve here? Is it that you had difficulty finding the information you needed about using `base` in your project? Do you think the info we already have on the environment variables page is insufficient or not helpful enough? It helps to know what the problem is so we can figure out what kind of docs solution works best!

sarah11918 (approved):
This does appear to be a helpful change, so approving!

Issue #4309, opened 2022-08-13
When creating a page with `getStaticPaths` which return a `number` param, later `Astro.params` returns a `string`


marceloverdijk:
### What version of `astro` are you using?

1.0.5

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

When creating a page with `getStaticPaths` which return a `number` param, later `Astro.params` returns a `string`.

```
---
import Layout from '../../layouts/Layout.astro';

export function getStaticPaths() {
  return [
    { params: { year: 1990 } }, // year is a number
    { params: { year: 1991 } },
    { params: { year: 1992 } },
  ]
}

const { year } = Astro.params
---

<Layout title="Season">
  {JSON.stringify(year)}
</Layout>
```

when accessing `/seasons/1990` this prints:
```
"1990"
```

although expected:
```
1990
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ydxdyc?file=src/pages/seasons/[year].astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

marceloverdijk:
Also trying with explicit typing the `year` to a `number` using a `Params` interface, the `year` value is still a `string`.

```
---
import Layout from '../../layouts/Layout.astro';

export function getStaticPaths() {
  return [
    { params: { year: 1990 } }, // year is a number
    { params: { year: 1991 } },
    { params: { year: 1992 } },
  ]
}

interface Params extends Record<string, string | number | undefined> {
  year: number
}

const { year } = Astro.params as Params
---

<Layout title="Season">
  {JSON.stringify(year)}
</Layout>

```

matthewp:
`Astro.params` get pulled from the URL. They will always be strings. You can return `props` as well which will be on `Astro.props` and those can be numbers or anything else.

marceloverdijk:
Hi @matthewp , thx for giving feedback.

I'm wondering if `Astro.params` are alway strings why is the type of `Astro.params` then:

```
export declare type Params = Record<string, string | number | undefined>;
```

Also note this PR , merged earlier this year:
https://github.com/withastro/astro/pull/3087 - Fix/numeric path params
which included the `number` type to `Params`...

```
- export type Params = Record<string, string | undefined>;
+ export type Params = Record<string, string | number | undefined>;
```


Although I now also notice this:

❯ npm run dev

> my-frontend@0.1.0 dev
[...]

PR #3087, opened 2022-04-12
Fix/numeric path params

tony-sull:
## Changes

Adds support for numeric route parameters in `getStaticPaths()`, for example when including years/months/days in a blog post URL

This also adds type validation for route parameters to throw an error when `getStaticPaths()` is given a parameter that isn't a string or number

## Testing

`getStaticPaths()` test updated to include a route using a number for the `year` parameter

## Docs

Related docs [PR #345](https://github.com/withastro/docs/pull/345)

PR #5484, opened 2022-11-27
Narrow the type of Params

### LINKED ISSUE #5442: Astro.params.foo is typed as string|number|undefined, but it can only be string|undefined, which forces extra TypeScript typing

1.6.10

node

npm

Linux (WSL)

### Describe the Bug

https://github.com/withastro/astro/blob/4b188132ef68f8d9951cec86418ef50bb4df4a96/packages/astro/src/%40types/astro.ts#L1149 says that `Astro.params` can contain numbers, which https://docs.astro.build/en/reference/api-reference/#params says doesn't actually happen. This forces TypeScript programs to either handle a `number` type that never actually happens or to cast it away.

I believe #3087 introduced this in order to loosen typing in `getStaticPaths()`, which produces `Params` instances, but it broke consumption of the types. You probably need two different types for inbound vs outbound.

Run `astro check` in https://stackblitz.com/edit/github-fht7tc?file=src/pages/products/[name].astro to see the error.

https://stackblitz.com/edit/github-fht7tc?file=src/pages/products/[name].astro

### Participation

Pimm:
Fixes #5442.

/cc @tony-sull & @bholmesdev

## Changes

[#3087 changed the type of `Params`](//github.com/withastro/astro/pull/3087) from `Record<string, string | undefined>` to `Record<string, string | number | undefined>` to formerly allow `getStaticPaths` to return numbers.

However, as the change was made to `Params` and not `GetStaticPathsResult` or `GetStaticPathsItem`, it also affects `Astro.params`.

This PR essentially moves the change made in #3087 from `Params` to `GetStaticPathsItem`, reverting the type of `Astro.params` while keeping the type of `getStaticPaths`.

**Note** that while looking into this, I [found this comment](https://github.com/withastro/astro/issues/4309#issuecomment-1239779706) by @bholmesdev:

> * if you pass a number to `params.year` in your `getStaticPaths`, it should _still_ be a number when destructuring `Astro.params`. If not, that's a bug.
> * if you created an **SSR dynamic route** like `[year].astro`, `year` will _always_ be a `string` (even if it could be parsed to a number). This is where string -> number conversion would be manual.

In my experience, `Astro.params` never contains numbers, not even in static mode. While reviewing, please consider that this PR changes the types to match the behaviour, although judging by this comment, we might want to change the behaviour to match the types instead.

## Testing

This is testable like so:
```typescript
const artist: string | undefined = Astro.params.artist;
```
I haven't found any type definition tests to which this one could be added.

## Docs

From what I understand, this change is [in line with the current docs](https://docs.astro.build/en/reference/api-reference/#params):

> `params` are encoded into the URL, so only strings are supported as values.

### DIFF
--- a/packages/astro/src/core/routing/params.ts
+++ b/packages/astro/src/core/routing/params.ts
@@ -1,4 +1,4 @@
-import type { Params } from '../../@types/astro';
+import type { GetStaticPathsItem, Params } from '../../@types/astro';
 import { validateGetStaticPathsParameter } from './validation.js';
 
 /**
@@ -27,12 +27,12 @@ export function getParams(array: string[]) {
  * values and create a stringified key for the route
  * that can be used to match request routes
  */
-export function stringifyParams(params: Params, routeComponent: string) {
+export function stringifyParams(params: GetStaticPathsItem['params'], routeComponent: string) {
 	// validate parameter values then stringify each value
 	const validatedParams = Object.entries(params).reduce((acc, next) => {
 		validateGetStaticPathsParameter(next, routeComponent);
 		const [key, value] = next;
-		acc[key] = typeof value === 'undefined' ? undefined : `${value}`;
+		acc[key] = value?.toString();
 		return acc;
 	}, {} as Params);
 
--- a/.changeset/nervous-laws-knock.md
+++ b/.changeset/nervous-laws-knock.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Narrow type of `Params`, as its values cannot be numbers
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1047,7 +1047,7 @@ export type GetHydrateCallback = () => Promise<() => void | Promise<void>>;
 	rss(): never;
 }
 
-export type GetStaticPathsItem = { params: Params; props?: Props };
+export type GetStaticPathsItem = { params: { [K in keyof Params]: Params[K] | number }; props?: Props };
 export type GetStaticPathsResult = GetStaticPathsItem[];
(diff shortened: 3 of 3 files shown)


packages/astro/src/core/routing/params.ts
```diff
@@ -27,12 +27,12 @@ export function getParams(array: string[]) {
...
  * that can be used to match request routes
  */
-export function stringifyParams(params: Params, routeComponent: string) {
+export function stringifyParams(params: GetStaticPathsItem['params'], routeComponent: string) {
 	// validate parameter values then stringify each value
 	const validatedParams = Object.entries(params).reduce((acc, next) => {
 		validateGetStaticPathsParameter(next, routeComponent);
 		const [key, value] = next;
-		acc[key] = typeof value === 'undefined' ? undefined : `${value}`;
+		acc[key] = value?.toString();
```
Pimm, line 35: This is more concise and works for `undefined`, numbers, and strings, but does require support for the `?.` operator. Since this support was added in Node.js 14.0.0, this should be OK.

bluwy (approved):

bholmesdev (approved):
I `par`am very impressed! Great change 👍

sarah11918:
While our API reference page does in fact say that `params` must be strings, we have this in our new Error reference guide:

Screenshot 2022-11-28 18 10 11
https://docs.astro.build/en/reference/error-reference/#invalid-value-for-getstaticpaths-route-parameter

So, no objections from docs if we're now all rallying behind Team String, but our error reference will have to change to match. (cc @Princesseuh @matthewp  )

Pimm:
Thanks @sarah11918.

The error reference looks correct to me as-is. `getStaticPaths` may still return numbers (although those numbers will reincarnate as strings in `Astro.params`).

PR #5495, opened 2022-11-29
Add a new error overlay

Princesseuh:
## Changes

This adds a new error overlay that is much more Astro-y and cuter than Vite's default one. It comes with a few features to make errors much more informative:

- A link to the [error reference](https://docs.astro.build/en/reference/error-reference/) for every known Astro error
- Basic support for Markdown to highlight and link to relevant information
- A full, syntax-colored code window 
- Support for dark and light mode
- Houston in the corner

**Before:**
image

**After:**
image

## Testing

Updated error tests for the new overlay's DOM structure

## Docs

N/A


packages/astro/src/core/errors/overlay.ts
```diff
@@ -0,0 +1,531 @@
...
+    </section>
+  </div>
+</div>
+`;
+
+const openNewWindowIcon =
+	/* html */
+	'<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="none"><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M14 2h-4m4 0L8 8m6-6v4"/><path stroke="currentColor" stroke-linecap="round" stroke-width="1.5" d="M14 8.667V12a2 2 0 0 1-2 2H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h3.333"/></svg>';
+
+const { HTMLElement = class {} } = globalThis;
```
matthewp, line 406: what's this about?

packages/astro/src/core/errors/overlay.ts
```diff
@@ -0,0 +1,531 @@
...
+
+		const { HTMLElement = class {} } = globalThis;
+		${ErrorOverlay.toString()}
+	`;
+}
+
+export function patchOverlay(code: string) {
+	return (
+		getOverlayCode() +
+		code.replace(
```
matthewp, line 535: I can't remember, did we look and see if there was another way to provide a custom overlay? Did that not work for some reason?

Princesseuh: I investigated taking over the entire client, but since Vite has a lot of code there (mostly related to HMR), we'd need to redo a lot of parts when we really just want to override the overlay. So I figured it wasn't worth it

I changed some code around this part right before sending it in, we could definitely make the replace more reliable

packages/astro/src/core/errors/overlay.ts
Princesseuh: Ah, I figured that in dev we could have a neater experience by writing the overlay code directly in JS inside of inside a string and then using `toString()` on the class when injecting it

packages/astro/src/core/errors/overlay.ts
matthewp: Ok, I was under the impression you could configure an alternative custom element constructor and not need to patch the code at all, but looking now I don't see this.

packages/astro/src/core/errors/overlay.ts
matthewp: I'm probably not understanding, but can you not just ts-ignore?

Princesseuh: Tested this again and I misremembered, it's not a build error but a runtime one, it'll complain that HTMLElement doesn't exist (since well, it doesn't exists on Node)

[Here's where this is originally in Vite's code](https://github.com/vitejs/vite/blob/db8df14d8934f11fa64e8b33e376f816b3b16f15/packages/vite/src/client/overlay.ts#L133-L135)

matthewp: No no, don't want to go to that trouble over this. This is a minor thing we can talk about later.

matthewp: I get it now. You are serializing the class using toString(). There are other ways to do this to avoid this code running in the browser (outside of making it a string), but not important.

matthewp (dismissed):

Princesseuh:
!preview

FredKSchott:
Any idea why the bot is continuing to post? Looks like there is a check in the CI script, but it's not being hit: https://github.com/withastro/astro/blob/main/.github/workflows/check-merge.yml#L16-L26

matthewp:
> Any idea why the bot is continuing to post?

I'm not sure. It used to work. Something maybe changed... I guess this line is maybe what's failing:

```js
if(review.user.id === 41898282 && review.state === 'CHANGES_REQUESTED') {
```

Which I guess it's possible that the user id is not the same. I'll add some more logging.

packages/astro/src/core/errors/overlay.ts
bluwy: I think what this PR does now might be the best way to get complete control of it. Vite could perhaps expose `vite/overlay` in the future to export some utilities, which could be done if more frameworks want to implement their own error overlay.

jasikpark:
!preview error-overlay

FredKSchott:
@matthewp with Erika going out, what are your thoughts about this PR going stale? I'd still like to merge and ship this behind an `--experimental-error-overlay` flag, and from quickly checking out the branch it looks like this can be done with just a single flag check where `patchOverlay()` is called. Erika also mentioned wanting to make sure that telemetry was properly disabled in the test suite.

Happy to leave it open until Erika returns, or help you get it over the finish line on her behalf, whichever you think is best.

matthewp:
This PR doesn't include telemetry so we don't need to worry about it. I think getting this out in the next minor makes sense, I'll assign myself to complete.

PR #5497, opened 2022-11-30
Refactor internal plugins code

bluwy:
## Changes

Making some incremental refactorings through separate PRs.

The commits can be reviewed individually to see the changes.

## Testing

Existing tests should pass.

## Docs

Added and updated the internal READMEs for new contributors.

### DIFF
--- a/packages/astro/src/vite-plugin-html/transform/index.ts
+++ b/packages/astro/src/vite-plugin-html/transform/index.ts
@@ -6,7 +6,6 @@ import slots, { SLOT_PREFIX } from './slots.js';
 
 export async function transform(code: string, id: string) {
 	const s = new MagicString(code, { filename: id });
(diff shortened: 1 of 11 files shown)


packages/astro/src/vite-plugin-html/transform/index.ts
```diff
@@ -16,14 +15,6 @@ export async function transform(code: string, id: string) {
 	);
 	s.append('`\n\t}\n}');
 
-	if (imports.size > 0) {
-		let importText = '';
-		for (const [path, importName] of imports.entries()) {
-			importText += `import ${importName} from "${path}";\n`;
-		}
-		s.prepend(importText);
-	}
```
bluwy, line 25: The `imports` variable doesn't seem used. So the size is always 0

Issue #5211, opened 2022-10-27
Telefunc not running with Astro


phiberber:
### What version of `astro` are you using?

1.6.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

Telefunc is a RPC library that works in Vite, Next, Nuxt, SvelteKit and others frameworks. The author of Telefunc (@brillout) thinks that this problem is currently with Astro.

Telefunc runs fine in development mode, the problem is when building it. We already solved an issue related to the Posix Path but now there's a problem with the `commonjs-resolver`.

This is the stack trace of it (I've deleted my user path for convenience):

`
[commonjs--resolver] id.endsWith is not a function
 error   id.endsWith is not a function
TypeError: id.endsWith is not a function
    at isWrappedId (file:///telefuncreproduction/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:7571:40)
    at Object.resolveId (file:///telefuncreproduction/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:7773:11)
    at file:///telefuncreproduction/node_modules/rollup/dist/es/shared/rollup.js:22710:40
`

### Link to Minimal Reproducible Example

https://github.com/phiberber/telefunc-reproduction

### Participation

- [X] I am willing to submit a pull request for this issue.

brillout:
I checked and indeed it does seem to be a problem on Astro's side. But it seems like only a minor thing.

Let me know if you have any question!

bluwy:
I had to remove the `appType: 'mpa'` config to get it working with Astro 1.6.1 and above. But yeah it seems to be working in dev only. Logging the `id` of the build error, I get:

```
{
  telefuncFilesGlob: '/Users/bjorn/Work/repros/telefunc-reproduction/node_modules/.pnpm/telefunc@0.1.28/node_modules/telefunc/dist/cjs/node/vite/telefuncFilesGlob.js'
}
```

And it seems to be caused by `telefunc` injecting a [Rollup input](https://github.com/brillout/telefunc/blob/b0db26477c58f15a04d4afdf9e2791af110009fc/telefunc/node/vite/build.ts#L23-L33)

haikyuu:
I also changed the vite config but this issue happened (in Vite not marko in particular) because I gave it an empty `build.rollupOptions.input` array

brillout:
[...]

PR #5499, opened 2022-11-30
Handle custom injected entry files during build

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5211

For the SSR build, allows different `entryFileName` if a Vite plugin injects additional rollup inputs.

### Notes

A brief flow of how Astro currently builds (static):
1. It first [runs an SSR build](https://github.com/withastro/astro/blob/c13c2b41d3421bbc7321310c84152a7dd839b8a3/packages/astro/src/core/build/static-build.ts#L69) so that the "generate page" phase can execute the SSR runtime to generate the pages.
2. Then it [runs the client build](https://github.com/withastro/astro/blob/c13c2b41d3421bbc7321310c84152a7dd839b8a3/packages/astro/src/core/build/static-build.ts#L89) only if there are [components to hydrate](https://github.com/withastro/astro/blob/c13c2b41d3421bbc7321310c84152a7dd839b8a3/packages/astro/src/core/build/static-build.ts#L76-L81).
3. [Generate the pages](https://github.com/withastro/astro/blob/c13c2b41d3421bbc7321310c84152a7dd839b8a3/packages/astro/src/core/build/static-build.ts#L93) from the SSR output
4. Remove SSR output since no server code needed for SSG.

### Notes 2

In c13c2b4, I removed executing the SSR build via checking Vite's build `outDir` as if that's change, it would break many other deployment integrations. So Vite's `outDir` must equal to Astro's `outDir`.

I also made `entryFileName` default to `[name].mjs` instead of `[name].js`. Previously telefunc only worked with the latter, but we want to make sure all SSR builds use the `.mjs` extension for cleanup.

## Testing

Tested manually with the repro. Existing tests should still pass.

## Docs

n/a. Plugins that inject new entries should probably work so it's a bug fix.


bluwy:
!preview telefunc

matthewp (approved):

natemoo-re (approved):

Issue #5475, opened 2022-11-25
[@astrojs/lit] Slotted content is wrapped in `<astro-slot>` elements which breaks native slot functionality


ba55ie:
### What version of `astro` are you using?

1.16.11

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows, Linux

### Describe the Bug

When using a client directive or the Class constructor for Lit Web Components an extraneous `astro-slot` element is created which interferes with styling of content within native `slot` elements.

When looking at the example `MyCard` elements, the first two don't get the correct css styling. The `h1` should be `hotpink` and the footer text should be `gray`.

The added `<astro-slot><h1>...</h1></astro-slot>` element interferes with the styling, because the `::slotted` selector only supports compound top-level selectors: https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted

There are also some render issues. I also tried the `is:raw` directive, but that doesn't seem to make a difference.

![image](https://user-images.githubusercontent.com/1845545/204014062-721da903-1be6-4f8f-b03b-f30e61b840fb.png)

![image](https://user-images.githubusercontent.com/1845545/204014332-7ef14ba2-207e-444a-bdf7-becd66537c24.png)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wqbjru?file=src/components/my-card.ts

### Participation

- [x] I am willing to submit a pull request for this issue.

hecspc:
We are facing the same problem. Not only the `<astro-island>` element make the `::slotted` selector useless, but also the pattern to inspect the slot elements.

We do have some components that have the next structure:
```html
<tab-bar>
  <tab>Item 1<tab>
  <tab>Item 2</tab>
<tab-bar>
```
This pattern is quite regular with other components like `Switcher`, `List` and similar. As you can imagine, the container component, in the example above the `tab-bar` component, needs to know its children. We do so querying the `slots` elements or using the event `@slotchange`, as  suggested in the [Lit documentation](https://lit.dev/docs/components/shadow-dom/#accessing-slotted-children).

Also these components could grow inserting more items. Normally will be inserting as children of the container element giving a strange structure diffucult to parse

```html
<list>
  <astro-slot>
     <list-item>Old Item<list>
     <list-item>Old Item<list>
  </astro-slot>
   <list-item>New Item<list>
   <list-item>New Item<list>
```
Here there is an example
[...]

PR #5500, opened 2022-11-30
Fix Lit `astro-slot` bug

ba55ie:
## Changes

- This removes the extraneous  `<astro-slot>` element which surrounds native slotted content.
- See issue #5475

## Testing

- There was already a test for `Lit` slotted content and that test still succeeds.
- I used my own test project.

## Docs

While this is a bug fix, it can cause rendering issues for users upgrading their project.

 It is a bug fix and restoring standard functionality.

### DIFF
--- a/.changeset/rich-mayflies-trade.md
+++ b/.changeset/rich-mayflies-trade.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/lit': patch
+---
+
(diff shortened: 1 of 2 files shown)


ba55ie:
Well, it seems this solution isn't 100% bulletproof. E.g. the regex is too greedy and the `<astro-island>` probably needs some refactoring too.

Issue #4090, opened 2022-07-29
🐛 No good error message when adding a client directive to an Astro component


sandervspl:
### What version of `astro` are you using?

1.0.0-rc.2

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

When you have a Preact component and a regular Astro component in the same project, and both have `client:idle` (or any other version of this), it will run into a syntax error on Safari. The location of the components does not matter. They can be in different pages, or on the same page. This bug seems to only happen after a production build, and only on Safari. Both desktop and mobile versions. It blocks all JavaScript from running.

Adding a `<script>` tag, a frontmatter, etc, does not resolve the issue. The reproduction is the most simple example I could create. Adding more code will still result in the error.

I have not tried using other integrations. I have only tested it with Preact.

#### Reproduction steps
1. Download and install the linked reproduction example, package manager doesn't matter.
2. Open Safari
3. Go to `http://localhost:3000`
4. Open terminal

There should be an error in the terminal

```
Unhandled Promise Rejection: SyntaxError: Invalid regular expression: invalid group specifier name
```

It seems to point to this part of the code

```html
<script>(self.Astro=self.Astro||{}).idle=a=>{const e=async()=>{await(await a())()}; ...
```

It specifically points to this part `await a()`

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-bzqxfw?file=src%2Fcomponents%2FPreactComponent.tsx,src%2Fpages%2Findex.astro&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

Remcoman:
I think this Regex is the issue: https://github.com/withastro/astro/blob/4819e7b071b20911dbb16b0701138655ef97cd62/packages/astro/src/runtime/server/index.ts#L653

Negative-lookbehind regexes are not supported in Safari.

matthewp:
That code doesn't run in the browser.

matthewp:
Oh. You have `<AstroComponent client:idle />`. Astro components cannot run in the client. I can't believe we don't have a proper error message for this already.

sandervspl:
Oh that's weird. The component that triggered this error is months old and hasn't changed for weeks. Error only popped up yesterday. Did something change?

hippotastic:
I think having the warning message is great, and it seems to work in `dev`. I think we can still improve on that in two aspects though:

[...]

PR #5316, opened 2022-11-07
Migrate error messages to new format (and misc error improvements)

Princesseuh:
## Changes

Follow up to https://github.com/withastro/astro/pull/5206!

This moves most(see below) of our errors to a new object that contain the error code, message and hint for known errors. The reason for this was to achieve the following goals:

- Much easier contributions from users who are not experienced with Astro's codebase (including our lovely docs team)
- Encourage and partly enforce a consistent format for error message
- Will make it possible to do doc generation from that object

While I was there, I studied ~~the blade~~ errors and came up with a little design document on how to best write error messages. A lot of our errors were already partly following these new guidelines, but some were enhanced with additional information / cleaned up when possible. Further work on improving the messages themselves will be done in smaller incremental PRs

Fix https://github.com/withastro/astro/issues/5150
Fix https://github.com/withastro/astro/issues/4634
Fix https://github.com/withastro/astro/issues/5151
Fix https://github.com/withastro/astro/issues/5152
Fix https://github.com/withastro/astro/issues/5276

## Caveats

Some errors were not touched, namely internal and CLI errors. This PR was slowly becoming humongous and I'd rather tackle those areas in separate PRs (especially since they both have slightly different requirements from public-facing errors)

## Testing

Current error tests still pass! Updated tests that needed changes

## Docs

Currently no direct docs part, I'll collaborate with docs for the next part of this (documenting the errors there)

cc @withastro/maintainers-docs for feedback on the error messages and the design guidelines for errors

PR #5501, opened 2022-11-30
Show warning about hydrating Astro components in build

Princesseuh:
## Changes

Bit of a lazy fix. I was hoping to be able to provide more information in the warning, but much like what I faced in https://github.com/withastro/astro/pull/5316, we don't have access to much there. In the future, I'd like to investigate a larger endeavour there to see if we can drill down info for error / warnings reporting (or bubble up warnings and errors, maybe?)

For now this is still an improvement. It's not that big of an issue that the location isn't in the message as the warning will show up when you visit the page / build it so you'll know either way in which file it is. In the future we'll also add an editor warning for this, which should also help users avoid this situation!

Fix https://github.com/withastro/astro/issues/4090

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/.changeset/yellow-lizards-shout.md
+++ b/.changeset/yellow-lizards-shout.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Added a warning in build when trying to hydrate an Astro component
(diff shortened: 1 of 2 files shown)


bluwy (approved):
Looks like we're also doing build-time warnings in other places so this seems good to me too 👍

PR #5506, opened 2022-12-01
Refactor Astro plugin and compile flow

bluwy:
## Changes

- Refactor `vite-plugin-astro` to split of core logic out of `index.ts` so they can be tested.
- Remove `/@fs` usage when compiling (uses `resolve.dedupe` to achieve this)

It's easier to review the changes by checking each commit individually.

## Testing

All existing tests should pass.

## Docs

N/A. refactors.

### DIFF
--- a/packages/astro/src/core/compile/cache.ts
+++ b/packages/astro/src/core/compile/cache.ts
@@ -1,0 +1,42 @@
+import type { AstroConfig } from '../../@types/astro';
+import { compile, CompileProps, CompileResult } from './compile';
+
+type CompilationCache = Map<string, CompileResult>;
+
+const configCache = new WeakMap<AstroConfig, CompilationCache>();
+
+export function isCached(config: AstroConfig, filename: string) {
+	return configCache.has(config) && configCache.get(config)!.has(filename);
+}
+
+export function getCachedCompileResult(
+	config: AstroConfig,
+	filename: string
+): CompileResult | null {
+	if (!isCached(config, filename)) return null;
+	return configCache.get(config)!.get(filename)!;
+}
+
+export function invalidateCompilation(config: AstroConfig, filename: string) {
+	if (configCache.has(config)) {
+		const cache = configCache.get(config)!;
+		cache.delete(filename);
+	}
+}
+
+export async function cachedCompilation(props: CompileProps): Promise<CompileResult> {
+	const { astroConfig, filename } = props;
+	let cache: CompilationCache;
+	if (!configCache.has(astroConfig)) {
+		cache = new Map();
+		configCache.set(astroConfig, cache);
+	} else {
+		cache = configCache.get(astroConfig)!;
+	}
+	if (cache.has(filename)) {
+		return cache.get(filename)!;
+	}
+	const compileResult = await compile(props);
+	cache.set(filename, compileResult);
+	return compileResult;
+}
--- a/packages/astro/src/core/compile/cache.ts
+++ b/packages/astro/src/core/compile/cache.ts
@@ -1,5 +1,5 @@
 import type { AstroConfig } from '../../@types/astro';
-import { compile, CompileProps, CompileResult } from './compile';
+import { compile, CompileProps, CompileResult } from './compile.js';
 
 type CompilationCache = Map<string, CompileResult>;
 
--- a/packages/astro/src/vite-plugin-astro/compile.ts
+++ b/packages/astro/src/vite-plugin-astro/compile.ts
@@ -1,10 +1,9 @@
 import { fileURLToPath } from 'url';
 import { ESBuildTransformResult, transformWithEsbuild } from 'vite';
 import { AstroConfig } from '../@types/astro';
-import { cachedCompilation, CompileProps } from '../core/compile';
-import { CompileResult } from '../core/compile/compile';
-import { LogOptions } from '../core/logger/core';
-import { getFileInfo } from '../vite-plugin-utils';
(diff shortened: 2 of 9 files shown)


packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -103,35 +102,22 @@ export default function astro({ settings, logging }: AstroPluginOptions): vite.P
...
-			}
-
-			let source = getCachedSource(config, raw.id);
-			if (!source) {
+			// For CSS / hoisted scripts, the main Astro module should already be cached
+			const filename = normalizeFilename(parsedId.filename, config);
+			const compileResult = getCachedCompileResult(config, filename);
+			if (!compileResult) {
 				return null;
 			}
```
bluwy, line 110: I found `this.resolve` and `normalizeFilename` seems to guard the same thing conceptually, so I opt for `normalizeFilename` as it's done by other plugins too.

packages/astro/src/vite-plugin-astro/compile.ts
```diff
@@ -0,0 +1,147 @@
...
+	logging,
+}: CachedFullCompilation): Promise<FullCompileResult> {
+	let transformResult: CompileResult;
+	let esbuildResult: ESBuildTransformResult;
+
+	try {
+		transformResult = await cachedCompilation(compileProps);
+		// Compile all TypeScript to JavaScript.
+		// Also, catches invalid JS/TS in the compiled output before returning.
+		esbuildResult = await transformWithEsbuild(transformResult.code, rawId, {
```
bluwy, line 40: Swapping our use of `esbuild.transform` to `transformWithEsbuild` so we use Vite's esbuild instance, and get the benefits of respecting the tsconfig.

I'm planning to refactor this more for the JSX plugin to when I backport https://github.com/vitejs/vite/pull/11120 to Vite 3.

packages/astro/src/vite-plugin-astro/compile.ts
```diff
@@ -0,0 +1,147 @@
...
+			loader: 'ts',
+			target: 'esnext',
```
bluwy, line 42: Use `esnext` as we don't need down-levelling (same behaviour as `esbuild.transform` and how Vite works currently).

packages/astro/src/core/compile/cache.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function invalidateCompilation(config: AstroConfig, filename: string) {
+	if (configCache.has(config)) {
+		const cache = configCache.get(config)!;
+		cache.delete(filename);
+	}
+}
+
+export async function cachedCompilation(props: CompileProps): Promise<CompileResult> {
```
matthewp, line 27: I like that the caching part is split into a separate module now!

packages/astro/src/vite-plugin-astro/compile.ts
matthewp: Oh interesting, how was our esbuild not respecting it?

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -103,35 +102,22 @@ export default function astro({ settings, logging }: AstroPluginOptions): vite.P
...
-			if (!raw) {
-				return null;
-			}
-
-			let source = getCachedSource(config, raw.id);
-			if (!source) {
+			// For CSS / hoisted scripts, the main Astro module should already be cached
+			const filename = normalizeFilename(parsedId.filename, config);
+			const compileResult = getCachedCompileResult(config, filename);
+			if (!compileResult) {
```
matthewp, line 108: I'm not sure this is right. At least looking at the old code we used to do `await cachedCompilation(compileProps);` for styles. Although thinking about it now, there shouldn't be a time where styles are loaded before we have compiled 🤔

matthewp:
I love how much smaller the module has gotten! Good refactor, not sure why the tests are failing though.

packages/astro/src/vite-plugin-astro/compile.ts
bluwy: Vite has [this code](https://github.com/vitejs/vite/blob/8d00673bdccdb9eec248f68bf790798f1d7e018e/packages/vite/src/node/plugins/esbuild.ts#L104) that passes the resolved tsconfig for that file during transform, as esbuild doesn't handle that by default. It should make the compiled output a bit more accurate, but in most case not noticable.

packages/astro/src/vite-plugin-astro/index.ts
bluwy: I noticed that for styles too but we always [guarded this part](https://github.com/withastro/astro/blob/3c44033e4ebafd28472d5c6a43e55c0363c6b555/packages/astro/src/vite-plugin-astro/index.ts#L114-L117) with a:

```js
let source = getCachedSource(config, raw.id);
if (!source) {
	return null;
}
```

So technically we're expecting the main module to be compiled before we process the styles, otherwise we bail.

matthewp: I would verify that all of the tests pass with this.

bluwy:
Windows has blessed me with a green check 😌 (though I had to rerun once but I'll take it)

Turns out I made some changes to `normalizeFilename` that accidentally preserved backslashes. (Should be normalized as forward slashes). And it caused errors in the compiled output as the compiler [inlines pathname as-is without escaping backslashes](https://github.com/withastro/compiler/blob/e03886ef7485eb8448bdc188aa464c52988336a5/internal/printer/printer.go#L409).

matthewp (approved):

PR #5508, opened 2022-12-01
improve a11y in new error overlay

jasikpark:
we should also make the "open in editor" button a button, not a link, if we are using JS for interactions

### DIFF
--- a/packages/astro/src/core/errors/overlay.ts
+++ b/packages/astro/src/core/errors/overlay.ts
@@ -389,20 +389,20 @@ ${style.trim()}
       </section>
       <div id="houston-overlay"></div>
       <div id="houston">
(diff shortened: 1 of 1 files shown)


packages/astro/src/core/errors/overlay.ts
```diff
@@ -389,20 +389,20 @@ ${style.trim()}
...
       </div>
     </header>
 
     <section id="message-hints">
       <section id="message">
         <span id="message-icon">
-          <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none"><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M12 7v6m0 4.01.01-.011M12 22c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10Z"/></svg>
+          <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" width="24" height="24" fill="none"><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M12 7v6m0 4.01.01-.011M12 22c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10Z"/></svg>
```
jasikpark, line 399: maybe there should be alt text here?

PR #5511, opened 2022-12-01
Head propagation

matthewp:
## Changes

- This implements the primitives for head propagation. See the test case to see how this can be used.
- Any module in the graph can be marked as doing head injection. Right now this is based on a comment syntax (which will be used for `renderEntry`. In the future Astro components could be marked as doing head injection by the compiler.
- A Vite plugin walks the graph to mark the tree as having head injection.
- At runtime components will mark themselves as doing injection. When we go to render the page we will recursively run their frontmatter scripts and collect all of the head content.
- When we go to render the head content is populated and renders out separately from the rest of the page.

## Testing

- Added a test case using the low-level APIs. This is what `renderEntry` will do as well.

## Docs

- This is a low-level implementation to be used by `renderEntry` and in the future, a head propagation API.


packages/astro/package.json
```diff
@@ -100,7 +100,7 @@
     "test:e2e:match": "playwright test -g"
   },
   "dependencies": {
-    "@astrojs/compiler": "^0.29.15",
+    "@astrojs/compiler": "next--module-id",
```
matthewp, line 103: Note that there is an associated compiler change. This should not be merged until that change is in.

packages/astro/src/runtime/server/index.ts
```diff
@@ -96,3 +90,8 @@ export function defineStyleVars(defs: Record<any, any> | Record<any, any>[]) {
 	}
 	return markHTMLString(output);
 }
+
+
+
+
+//////////////////////// new stuff
```
matthewp, line 97: oops, remove

packages/astro/src/runtime/server/render/page.ts
```diff
@@ -45,6 +51,21 @@ async function iterableToHTMLBytes(
 	return parts.toArrayBuffer();
 }
 
+async function bufferHeadContent(result: SSRResult) {
+	debugger;
```
matthewp, line 55: remove

packages/astro/src/vite-plugin-astro/types.ts
```diff
@@ -1,9 +1,11 @@
 import type { TransformResult } from '@astrojs/compiler';
+import type { PropagationHint } from '../@types/astro';
 
 export interface PluginMetadata {
 	astro: {
 		hydratedComponents: TransformResult['hydratedComponents'];
 		clientOnlyComponents: TransformResult['clientOnlyComponents'];
 		scripts: TransformResult['scripts'];
+		headInjection: PropagationHint;
```
matthewp, line 9: Should rename this to `propagation` to match the main types.

packages/astro/src/runtime/server/render/astro/instance.ts
```diff
@@ -0,0 +1,82 @@
...
+			yield * value.content;
+		} else {
+			yield * renderChild(value);
+		}
+	}
+}
+
+// In dev mode, check props and make sure they are valid for an Astro component
+function validateComponentProps(props: any, displayName: string) {
+	if (import.meta.env?.DEV && props != null) {
```
bluwy, line 53: Merge note: `import.meta.env?.DEV` was removed in https://github.com/withastro/astro/pull/5501

packages/astro/src/vite-plugin-head-propagation/index.ts
```diff
@@ -0,0 +1,53 @@
+import type { AstroSettings } from '../@types/astro';
+import type { AstroPluginMetadata } from '../vite-plugin-astro/index';
+import type { ModuleInfo } from 'rollup';
+
+import * as vite from 'vite';
+
+const injectExp = /^\/\/[ ]*astro-head-inject/;
```
bluwy, line 7:
```suggestion
const injectExp = /^\/\/\s*astro-head-inject/;
```

I think we can use `\s` for a space here

bluwy (commented):
Tried to digest this and it looks great aside from some comments below

packages/astro/src/runtime/server/render/astro/instance.ts
matthewp: Oh ok, how do we know we are running in dev now?

bluwy: The idea was that the warning should be called on build too. Does this need to be dev only?

packages/astro/src/core/compile/compile.ts
```diff
@@ -42,6 +44,7 @@ async function compile({
 		site: astroConfig.site?.toString(),
 		sourcefile: filename,
 		sourcemap: 'both',
+		moduleId,
```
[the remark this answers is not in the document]
matthewp: This was removed after I rebased with your branch.

matthewp:
!bench

packages/astro/src/runtime/server/render/astro/instance.ts
matthewp: Build is ok, but in SSR this is going to be called on every render which is quite expensive.

matthewp:
!bench

matthewp:
!bench

packages/astro/src/vite-plugin-head-propagation/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const injectExp = /^\/\/\s*astro-head-inject/;
+
+/**
+ * If any component is marked as doing head injection, walk up the tree
+ * and mark parent Astro components as having head injection in the tree.
+ * This is used at runtime to determine if we should wait for head content
+ * to be be populated before rendering the entire tree.
+ */
+export default function configAliasVitePlugin({
```
natemoo-re, line 15: Wrong name?

packages/astro/src/@types/astro.ts
```diff
@@ -1398,10 +1398,15 @@ export interface SSRMetadata {
 	hasRenderedHead: boolean;
 }
 
+export type PropagationHint = 'none' | 'self' | 'in-tree';
```
natemoo-re, line 1401: Add comment explaining what these do?

natemoo-re (approved):
Two small nits, but otherwise LGTM.

packages/astro/src/vite-plugin-head-propagation/index.ts
matthewp: woops, yeah

packages/astro/src/runtime/server/render/head.ts
```diff
@@ -27,16 +34,28 @@ export function renderHead(result: SSRResult): Promise<string> {
...
+	const baseHeadContent = markHTMLString(links.join('\n') + styles.join('\n') + scripts.join('\n'))
+
+	if(result.extraHead.length > 0) {
+		return renderExtraHead(result, baseHeadContent);
+	} else {
+		return baseHeadContent;
+	}
+}
+
+export function renderHead(result: SSRResult) {
```
bholmesdev, line 47: nit: maybe name `createRenderHead` to clarify this returns another function?

packages/astro/test/units/dev/head-injection.test.js
```diff
@@ -0,0 +1,160 @@
+import { expect } from 'chai';
+import * as cheerio from 'cheerio';
+
+import { runInContainer } from '../../../dist/core/dev/index.js';
+import { createFs, createRequestAndResponse } from '../test-utils.js';
+
+const root = new URL('../../fixtures/alias/', import.meta.url);
+
+describe('head injection', () => {
```
bholmesdev, line 9: Couldn't quite tell from these test cases, but will head injection support inline styles alongside standard stylesheets? I know that `renderEntry` needs to handle both inline styles and links from the render result today, since MDX outputs as inline styles in dev, stylesheets in production.

bholmesdev (commented):
Small nit and a clarifying question, but propagreat work!

packages/astro/test/units/dev/head-injection.test.js
matthewp: This makes it possible to create a component that does head injection. So `renderEntry` will need to use the right APIs, such as the `createComponent` object form, to return the CSS it needs.

So if I'm understanding you right, that work will be done in the renderEntry PR.

bholmesdev: @matthewp got it. I'm asking since I know a component may want to inject an inline `style` or a `link` tag. It sounds like this API gives you the freedom to inject _any_ HTML element + element children that you want?

matthewp: Right, you can return any string or iterable, essentially the same thing you can do in a template. This PR contains a helper for rendering a stylesheet: https://github.com/withastro/astro/pull/5511/files#diff-a2d99b846ae3297fe01c808e11546d92b906ffafdbc17134bca73edf276fd3b5R17

I'll probably add more helpers in the other PR.

PR #4136, opened 2022-08-03
Support Fragment in MDX files

natemoo-re:
## Changes

- Passes `Fragment` into MDX files.
- Fixes https://github.com/withastro/astro/issues/4130

## Testing

Tested manually

## Docs

Bug fix only

PR #5522, opened 2022-12-03
Support rendering `<Fragment>` in MDX `<Content />` component

### LINKED ISSUE #5519: Rendering `<Content />` fails for MDX file containing `<Fragment>`

1.6.12

n/a

npm

macOS / Stackblitz

### Describe the Bug

(Ran into this bug as part of migrating docs to MDX and it’s a bit of blocker given how much we use `<Fragment>` to pass content to slots.)

1. MDX supports using `<Fragment>`, which can be useful for example to pass stuff to a named slot:
   ```mdx
   <Component>
     <Fragment slot="box">Box content</Fragment>
   </Component>
   ```
   This renders fine for an MDX page, e.g. `src/pages/index.mdx`.

2. If you import (or glob) the same page and use it’s `<Content />` component however:
   ```astro
   ---
   import { Content } from './index.mdx';
   ---
   <Content />
   ```
   Astro will throw an error:
   ```
   Expected component `Fragment` to be defined: you likely forgot to import, pass, or provide it.
   ```

I guess somewhere `<Fragment>` support is being wired up when a page is rendered but that same wiring is missing for the `<Content />` render.

https://stackblitz.com/edit/github-zq6zwz?file=src/pages/index.mdx

### Participation

delucis:
## Changes

- Fixes #5519 
- Wraps MDX’s default `MDXContent` render function so we can pass in our definition of `Fragment` — otherwise `Fragment` is unsupported when rendering MDX using `<Content />`
- Reverts astro core workaround from #4136 which is no longer needed because this PR fixes the general case of fragments in MDX

## Testing

Expanded the `mdx-component` fixture/test suite to include an example `WithFragment.mdx` file which we test renders as expected.

## Docs

Bug fix only, should match existing docs expectations.

### DIFF
--- a/packages/astro/src/core/render/core.ts
+++ b/packages/astro/src/core/render/core.ts
@@ -113,6 +113,7 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 	}
 
 	// HACK: expose `Fragment` for all MDX components
+	// TODO: Remove in Astro v2 — redundant as of @astrojs/mdx@>0.12.0
 	if (typeof mod.default === 'function' && mod.default.name.startsWith('MDX')) {
 		Object.assign(pageProps, {
 			components: Object.assign((pageProps?.components as any) ?? {}, { Fragment }),
--- a/packages/astro/src/core/render/core.ts
+++ b/packages/astro/src/core/render/core.ts
@@ -3,7 +3,7 @@ import type { LogOptions } from '../logger/core.js';
 import type { RenderContext } from './context.js';
 import type { Environment } from './environment.js';
 
-import { renderPage as runtimeRenderPage } from '../../runtime/server/index.js';
+import { Fragment, renderPage as runtimeRenderPage } from '../../runtime/server/index.js';
 import { attachToResponse } from '../cookies/index.js';
 import { AstroError, AstroErrorData } from '../errors/index.js';
 import { getParams } from '../routing/params.js';
@@ -110,6 +110,13 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 	// Support `export const components` for `MDX` pages
 	if (typeof (mod as any).components === 'object') {
 		Object.assign(pageProps, { components: (mod as any).components });
+	}
+
+	// HACK: expose `Fragment` for all MDX components
+	if (typeof mod.default === 'function' && mod.default.name.startsWith('MDX')) {
+		Object.assign(pageProps, {
+			components: Object.assign((pageProps?.components as any) ?? {}, { Fragment }),
+		});
 	}
 
 	const response = await runtimeRenderPage(
--- a/packages/astro/src/core/render/core.ts
+++ b/packages/astro/src/core/render/core.ts
@@ -3,7 +3,7 @@ import type { LogOptions } from '../logger/core.js';
 import type { RenderContext } from './context.js';
 import type { Environment } from './environment.js';
 
-import { Fragment, renderPage as runtimeRenderPage } from '../../runtime/server/index.js';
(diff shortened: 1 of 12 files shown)


packages/integrations/mdx/src/index.ts
```diff
@@ -156,9 +152,19 @@ export default function mdx(mdxOptions: MdxOptions = {}): AstroIntegration {
 										)}) };`;
 									}
 									if (!moduleExports.includes('Content')) {
-										code += `\nexport const Content = MDXContent;`;
+										// Make `Content` the default export so we can wrap `MDXContent` and pass in `Fragment`
+										code = code.replace('export default MDXContent;', '');
+										code += `\nexport const Content = (props = {}) => MDXContent({
+											...props,
+											components: { Fragment, ...props.components },
```
delucis, line 159: In my tests, compiled MDX already includes an unused `Fragment` import, so using that here.

`parseESM` does return a list of imports too, so we could potentially use that to be extra sure the import exists if we think there are circumstances where it might not?

packages/astro/src/core/render/core.ts
```diff
@@ -112,13 +112,6 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 		Object.assign(pageProps, { components: (mod as any).components });
 	}
 
-	// HACK: expose `Fragment` for all MDX components
-	if (typeof mod.default === 'function' && mod.default.name.startsWith('MDX')) {
-		Object.assign(pageProps, {
-			components: Object.assign((pageProps?.components as any) ?? {}, { Fragment }),
-		});
-	}
```
delucis, line 120: With the changes to the MDX integration in this PR, this code is unnecessary (but harmless) so could be removed.

But, could be a breaking change for people who use fragments in MDX pages currently and update `astro` but _don’t_ update `@astrojs/mdx`. Should I revert and we can remove in an `astro` major?

Another option could be to release the MDX changes as a minor (`0.13.0`) and add a minimum supported `astro` version to `peerDependencies`.

packages/integrations/mdx/src/index.ts
bluwy: Looks like it's coming from https://github.com/withastro/astro/blob/a1885ea2f59f26cbdae10c298e1e0d1063d9dca1/packages/astro/src/jsx/babel.ts#L150-L157. I think checking from `parseESM` would be great since it's free and gives better errors in case we miss it. 👍

packages/astro/src/core/render/core.ts
bluwy: Yeah this seems technically a breaking change. +1 on reverting for now and adding a TODO to remove in v2

packages/integrations/mdx/src/index.ts
```diff
@@ -132,11 +132,18 @@ export default function mdx(mdxOptions: MdxOptions = {}): AstroIntegration {
...
+									// Fragment import should already be injected, but check just to be safe.
+									const importsFromJSXRuntime = moduleImports
+										.filter(({ n }) => n === 'astro/jsx-runtime')
+										.map(({ ss, se }) => code.substring(ss, se));
+									const hasFragmentImport = importsFromJSXRuntime.some((statement) =>
+										/[\s,{](Fragment,|Fragment\s*})/.test(statement)
+									);
+									if (!hasFragmentImport) {
+										code = 'import { Fragment } from "astro/jsx-runtime"\n' + code;
+									}
```
delucis, line 146: Turns out the parsed imports aren’t in the friendliest of formats. This is pretty thorough and will inject the import if it doesn’t find one. If you think an error in that scenario would be more appropriate, let me know.

bholmesdev (approved):
It was `Frag`meant to be that you'd find this 🙏

Issue #5526, opened 2022-12-03
Error reporting fails on undefined error index


jasikpark:
### What version of `astro` are you using?

astro@0.0.0-error-overlay-20221130171441

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

Getting an undefined property access on `errors[i].message`

https://github.com/withastro/astro/blob/a1885ea2f59f26cbdae10c298e1e0d1063d9dca1/packages/astro/src/core/errors/dev/utils.ts#L51

may be useful to enable "noUnindexedAccess" in the repo for typescript..

```
file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/core/errors/dev/utils.js:41
      err[i].message = text;
                     ^

TypeError: Cannot set properties of undefined (setting 'message')
    at file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/core/errors/dev/utils.js:41:22
    at Array.forEach (<anonymous>)
    at collectErrorMetadata (file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/core/errors/dev/utils.js:38:14)
    at onError (file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/vite-plugin-astro-server/request.js:59:33)
    at runWithErrorHandling (file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/vite-plugin-astro-server/controller.js:68:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async handleRequest (file:///Users/calebjasik/Git/aoc-astro/node_modules/.pnpm/astro@0.0.0-error-overlay-20221130171441/node_modules/astro/dist/vite-plugin-astro-server/request.js:39:3)
 ELIFECYCLE  Command failed with exit code 1.

aoc-astro on  main [!] via  v16.17.1 took 4s 
❯ 
```

### Link to Minimal Reproducible Example

[...]

PR #5539, opened 2022-12-06
fix: Error reporting fails on undefined error index

wulinsheng123:
## Changes

fix: #5526

## Docs

fix: #5526


Princesseuh:
Thank you for contributing! This fix seems fine to me, could you add a changeset?

wulinsheng123:
> Thank you for contributing! This fix seems fine to me, could you add a changeset?

I done

natemoo-re (approved):

PR #5543, opened 2022-12-06
Fix bench workflow

matthewp:
## Changes

- Need to pipe stderr to the file as there is where the results are printed

## Testing

- just locally

## Docs

N/A

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -48,9 +48,10 @@ jobs:
       - name: Run benchmark
         id: benchmark-pr
         run: |
-          pnpm run --silent benchmark > ./bench-result.md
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -48,9 +48,10 @@ jobs:
       - name: Run benchmark
         id: benchmark-pr
         run: |
-          pnpm run --silent benchmark > ./bench-result.md
+          pnpm run --silent benchmark 1&> ./bench-result.md
           result=$(awk '/requests in/' ./bench-result.md)
           echo "::set-output name=BENCH_RESULT${{matrix.node-version}}::$result"
+          echo "$result"
```
matthewp, line 54: this is to make it easier to debug

PR #5547, opened 2022-12-06
Update telemetry to use new errors information

Princesseuh:
## Changes

This PR updates telemetry events so it uses our new AstroError's data when possible. Most importantly, this allows us to have much better information about our own errors in our stats, since we can safely extract the entire error message anonymized + have a cool name and errorCode to filter errors by. The previous anonymizing logic is kept for unknown errors, so we can still get a basic idea of what happened just in case

Additionally, this PR adds telemetry for errors encountered in dev, as previously we'd only get telemetry for dev server stops.

## Testing

Added a test for a error event using our own errors, updated tests that needed it. Made it so telemetry doesn't work in tests to avoid triggering it unnecessarily

## Docs

N/A


packages/astro/src/events/error.ts
```diff
@@ -65,11 +66,47 @@ export function eventError({
...
+				: anonymizeErrorMessage(err.message),
 	};
 	return [{ eventName: EVENT_ERROR, payload }];
 }
+
+/**
+ * Safely get the error message from an error, even if it's a function.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+function getSafeErrorMessage(message: string | Function): string {
```
Princesseuh, line 86: The logic here is the same one that generates the [error reference](https://docs.astro.build/en/reference/error-reference/) on the docs, which has been proven to work:tm:

packages/astro/test/units/vite-plugin-astro-server/request.test.js
```diff
@@ -51,6 +51,10 @@ describe('vite-plugin-astro-server', () => {
 				logging
 			);
 
+			// `handleRequest` below has a call to telemetry when an error is found. Telemetry should already be disabled
+			// in CI by default, but we'll make sure to disable it here too just in case.
+			process.env.ASTRO_TELEMETRY_DISABLED = true;
```
matthewp, line 56: I don't like that we now need to remember to do this in every test that does dev server requests. That would mean all of these tests need it too: https://github.com/withastro/astro/tree/main/packages/astro/test/units/dev Plus any fixture-based tests that run dev.

What I would probably suggest doing is making this an option on the `Environment` object and then having the tests pass through `false`.

Princesseuh: That's understandable, will do!

matthewp:
Is this the first time we are sending errors to telemetry? Assuming so, can we have some tests that include non-anonymous error messages to verify that we are stripping them correctly (or just not sending the message, if that is the intent).

Princesseuh:
> Is this the first time we are sending errors to telemetry? Assuming so, can we have some tests that include non-anonymous error messages to verify that we are stripping them correctly (or just not sending the message, if that is the intent).

It's actually not the first time! The code for reporting errors is from pre-1.0 and hasn't been changed (apart from the fact that we now exclusively use it for unknown errors, whereas before it was used for both known and unknown errors)

delucis (approved):
Approving as docs because this doesn’t actually require docs.

matthewp:
@Princesseuh Looked over the tests and the only one I think we should add for now is one to verify that the stack trace is not included. Looking at the code I'm pretty sure it will not be, but it would be good to have a test to verify and enforce that going forward.

packages/astro/src/events/error.ts
```diff
@@ -64,12 +65,51 @@ export function eventError({
...
+	if (typeof message === 'string') {
+		return message;
+	} else {
+		return String.raw({
+			raw: extractStringFromFunction(message.toString()),
```
aFuzzyBear, line 94: I know this method 😃

packages/astro/src/events/error.ts
```diff
@@ -64,12 +65,51 @@ export function eventError({
...
+		});
+	}
+
+	function extractStringFromFunction(func: string) {
```
aFuzzyBear, line 98: Im not sure about this one... from reading it and please correct me, but this extracts the name from the function, 🤔
but this only works with the use case of anonymous functions, what about named functions? or is it outwith the scope of the method?
Also killer ninja regexing 🐱‍👤

packages/telemetry/src/index.ts
```diff
@@ -145,7 +145,6 @@ export class AstroTelemetry {
 			// to preview what data would be sent.
 			return Promise.resolve();
 		}
-
```
aFuzzyBear, line 148: 😎

aFuzzyBear (commented):
`tel` you what, this looks really good...💚

packages/astro/src/events/error.ts
Princesseuh: Yes! Named functions are outside the scope of this function, in our error definitions, we only use arrow functions (named functions would be totally unnecessary, as we'd never refer to them by name). The functions there are also on purpose simple, as they're also used for documentation

natemoo-re (approved):

Issue #5462, opened 2022-11-23
Styles are added to head in wrong order in prod


IanVS:
### What version of `astro` are you using?

1.6.10

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

npm

### What operating system are you using?

mac

### Describe the Bug

I got bitten by a really strange error when trying to add a new page to our site.  The hamburger button in the top nav started showing up in desktop sizes, but only in production builds.  I traced it down to an issue in the order that styles were being added to the `head`, global styles were being added _after_ component styles.  I've created as minimal of a reproduction that I could, hopefully it's helpful.  

I've noticed a lot of times where the styles in dev don't match production, which is pretty concerning, so I'm hopeful that this reproduction can expose an issue that can be solved and might help bring dev and prod builds more into line.

### Link to Minimal Reproducible Example

https://github.com/IanVS/astro-style-order-reproduction

### Participation

- [ ] I am willing to submit a pull request for this issue.

prodkt:
Happening to us as well! Watching in case we can provide additional detail as a discussion around this continues.

wulinsheng123:
![image](https://user-images.githubusercontent.com/27261522/203886718-90c48f31-7969-482c-9e6b-db23d0cb5a3a.png)
Help me check out

IanVS:
@wulinsheng123 thanks for taking a look, I just tested again in a fresh `git clone`, and the button is green for me.  What is the `Simple Browser` that you're opening this in?  The URL looks strange.  What happens if you open `http:localhost:3000` in firefox or chromium?

wulinsheng123:
I used `gitpod` that online editor, to open your project.
It runs on Chrome

IanVS:
Did you try cloning the project and building it locally, without gitpod?

jasikpark:
<img width="1624" alt="Screenshot 2022-11-27 at 10 57 41 PM" src="https://user-images.githubusercontent.com/10626596/204197274-e33a899f-aac4-4d9d-b27e-7b532a6f3023.png">
looks like it doesn't reproduce for me on gitpod either?

IanVS:
@wulinsheng123 @jasikpark can either of you reproduce in a standard environment?  Running in gitpod could be a confounding factor.

wulinsheng123:
I will try late. Recently I have had a little busy

wulinsheng123:
[...]

PR #5549, opened 2022-12-06
Use accumulated sort order when order production CSS

matthewp:
## Changes

- To determine how to order CSS bundles we use import order; if you want your CSS to have priority make it lower in the `import` than other CSS.
- However in the past this was only accounted for at the page component level. So if you used layouts the sorting within the layout component wasn't considered.
- This changes it so that the sort order is an accumulation of each module all the way up the tree. So that it takes into account layouts and other nested components.
- Fixes https://github.com/withastro/astro/issues/5462

## Testing

- Test added based on the test case in https://github.com/withastro/astro/issues/5462

## Docs

N/A, bug fix, this is already the documented way we do CSS sorting.


pnpm-lock.yaml
```diff
@@ -18424,7 +18430,7 @@ packages:
         optional: true
     dependencies:
       '@types/node': 18.11.9
-      esbuild: 0.15.18
+      esbuild: 0.15.14
```
natemoo-re, line 18433: Not sure how this snuck in here

natemoo-re (approved):
Looks good! Lockfile change snuck in, tho.

pnpm-lock.yaml
matthewp: hm, maybe i need to rebase with main

matthewp: this is fixed now.

jasikpark:
woooo glad this issue is fixed && that also now the CSS is weighted correctly based on import side-effect order :)

jasikpark:
!preview fix-css-accumulated-sort-order

jasikpark:
ope. that command fails when the backing branch is gone 😅

PR #5550, opened 2022-12-07
astro-rss: Generate feed with proper XML escaping

andersk:
## Changes

Generating XML using string concatenation is dangerous and can lead to underescaping.

Just to preemptively dismiss some excuses for this unsafety 😛:

- _“But most of the content was surrounded with with `<![CDATA[…]]>`, so it was fine!”_ Wrong. The content might contain `]]>` itself and break out of the `<![CDATA[` early.
- _“But URLs don’t need to be escaped, so it was fine!”_ Wrong. `&` commonly appears in URLs and needs to be escaped for XML or HTML serialization.
- _“But we were validating the output XML, so it was fine!”_ Wrong. A bug that’s caught at generation time is still a bug; also, some cases of underescaping could lead to incorrect output that isn’t invalid.
- _“But there’s no user-generated content for an attacker to control, so it was fine!”_ Wrong. Underescaping is still a bug even when it’s not a security vulnerability.

So let’s take the `fast-xml-parser` library that we were using to validate our hand-rolled XML, and use it to properly generate XML instead.

## Testing

I updated the existing tests to compare the XML using `chai-xml`, so that the snapshotted XML strings didn’t need to change. This verifies that the new output is semantically equivalent.

## Docs

This doesn’t need any documentation changes. I even preserve support for the existing `customData` option by parsing and re-serializing the XML string.

### DIFF
--- a/.changeset/hungry-snakes-live.md
+++ b/.changeset/hungry-snakes-live.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/rss': patch
+---
+
+Generate RSS feed with proper XML escaping
--- a/packages/astro-rss/package.json
+++ b/packages/astro-rss/package.json
@@ -31,6 +31,7 @@
     "astro-scripts": "workspace:*",
     "chai": "^4.3.6",
     "chai-as-promised": "^7.1.1",
+    "chai-xml": "^0.4.0",
     "mocha": "^9.2.2"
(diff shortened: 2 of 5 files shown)


natemoo-re (approved):
Makes sense to me, thanks for contributing!

PR #5559, opened 2022-12-08
Add Astro.joinBase

matthewp:
Note that this PR is **just an experiment**. If it is successful then this might become an RFC. 

## Motivation

This PR is intended to aid when using `base` configuration in Astro. A common issue is that because Astro does not mutate your templates when you set things like `<a href=` or `<img src=`, you have to remember to include the base yourself. This raises the problem of what do you do if your base changes? Or if it is dynamic per-environment?

### Astro.joinBase

With this PR there is a new utility method `Astro.joinBase`. With this method you simply pass it in a path (starting with a `/`) and it will join your base onto it. Here's an example usage:

```astro
---
const contactsLink = Astro.joinBase('/contacts');
---

<a href={contactsLink}>Contacts page</a>
```

When using the base of `/docs/` that will produce `/docs/contacts`.

### join-base: directive

However, since this is so commonly needed inside of the template this change also includes a directive `join-base:` which is used within your template to get the same effect, but without needing to create an expression and call the function.

Here's the same example but used with the directive:

```astro
<a join-base:href="/contacts">Contacts page</a>
```

## Example

This Stackblitz demonstrates usage: https://stackblitz.com/edit/github-lzehwm?file=astro.config.mjs,src%2Fpages%2Findex.astro,src%2Fpages%2Fabout.astro&on=stackblitz

## Installation

This change is in preview. You can install with a special prerelease tag:

```shell
npm install astro@next--join-base
```

## Changes

- Adds `Astro.joinBase` which is created in the Astro global.
  - Note that this problem can/should be created as the top-level global instead.
- Updates the compiler which includes `join-base:` directive.

## Testing

- None yet. Compiler has a test

## Docs

- Docs added in Astro types
- Need to update types for the directive
- Needs real docs on the site, obviously

### DIFF
--- a/.changeset/good-balloons-add.md
+++ b/.changeset/good-balloons-add.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Adds Astro.joinBase
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -100,7 +100,7 @@
     "test:e2e:match": "playwright test -g"
   },
   "dependencies": {
-    "@astrojs/compiler": "^0.30.0",
+    "@astrojs/compiler": "0.0.0-join-base-20221208164627",
     "@astrojs/language-server": "^0.28.3",
     "@astrojs/markdown-remark": "^1.1.3",
     "@astrojs/telemetry": "^1.0.1",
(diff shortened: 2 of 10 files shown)


matthewp:
!preview join-base

matthewp:
Closing as this was an experiment. Will use learnings from this to inform the next step of improving base in Astro, thanks!

PR #5560, opened 2022-12-08
Node: better serverEntrypoint error

natemoo-re:
## Changes

- We were accidentally logging `${fileURLToPath}` instead of `${fileURLToPath(serverEntrypoint)}`

## Testing

Tested manually

## Docs

N/A, bug fix only


matthewp (approved):

bholmesdev (commented):
At first I didn't see the (entry)point, but I get it now!

PR #5564, opened 2022-12-09
Add `server.headers` option

riywo:
With this new `server.headers` option, the users can specify custom headers for `astro dev` and `astro preview` servers.

This is useful when they want to build a website requiring specific response headers such as `Cross-Origin-Opener-Policy`.

## Changes

Adding a new `server.headers` option to allow the users to specify HTTP response headers for both `astro dev` and `astro preview`.

## Testing

Added new unit tests for both `astro dev` and `astro preview`.

## Docs

This document should be updated: https://docs.astro.build/en/reference/configuration-reference/#server-options

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -144,6 +144,11 @@ export async function handleRoute(
 		clientAddress: buildingToSSR ? req.socket.remoteAddress : undefined,
 	});
 
+	// Set user specified headers to response object.
+	for (const [name, value] of Object.entries(config.server.headers ?? {})) {
+		if (value) res.setHeader(name, value);
+	}
+
 	// attempt to get static paths
 	// if this fails, we have a bad URL match!
 	const paramsAndPropsRes = await getParamsAndProps({
--- a/.changeset/light-moose-attend.md
+++ b/.changeset/light-moose-attend.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Add `server.headers` option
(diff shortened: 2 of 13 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -144,6 +144,11 @@ export async function handleRoute(
 		clientAddress: buildingToSSR ? req.socket.remoteAddress : undefined,
 	});
 
+	// Set user specified headers to response object.
+	for (const [name, value] of Object.entries(config.server.headers ?? {})) {
+		if (value) res.setHeader(name, value);
+	}
+
```
riywo, line 151: I'm not sure why but passing `server.headers` to `vite` config didn't work. Hence, writing the headers here directly.

riywo:
Can anyone rerun the failed job? It failed at post steps, so it actually succeeded. https://github.com/withastro/astro/actions/runs/3654029967/jobs/6174109857

riywo:
Thank you for kicking a retry. Now all checks have passed.

It would be great if someone can review and merge this change. I'd like to build a website using `SharedArrayBuffer` which requires [some HTTP headers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements) and this change will help such use cases. Thanks!

riywo:
Done. This patch was needed when invoked from CLI actually. https://github.com/withastro/astro/compare/a34a19e8c92b5fcb92fe2651369e68465f090405..121ed37bdde07213628d9347405581d0c5a187c8

Not sure we should add such unit tests (or E2E tests?), though.

matthewp:
I think this is good since this is done manually. In production you'll have to make sure your server is configured to do the same, but I think given this is an opt-in approach that becomes obvious.

matthewp (approved):

matthewp (dismissed):
Blocking because this is a `minor` change. It will go out with the next minor release (likely on Thursday)

Issue #5502, opened 2022-11-30
Hashed build assets end up in multiple places in build output


delucis:
### What version of `astro` are you using?

1.6.12

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

macOS / Stackblitz

### Describe the Bug

Astro outputs various assets from its build and adds a hash to the file names to help with immutable caching. Currently these end up in various places making it less simple than it could be to set up cache headers for static builds:

- `dist/assets/<name>.<hash>.css`
  - styles in `<style>` tags or imported as ES modules

- `dist/chunks/<name>.<hash>.js`
  - scripts from UI integrations (Preact in this reproduction)
  - code-split modules (dynamically imported `canvas-confetti` in this repro)

- `dist/hoisted.<hash>.js` (also `dist/client.<hash>.js`?)
  - hoisted scripts from Astro components (if they’re bigger than Vite’s size limit for inlining — forced in the reproduction using `vite.build.assetsInlineLimit`)

- `dist/<ComponentName>.<hash>.js`
  - code for hydrated framework components

Proposal here is to move these all under `assets/` so that caching is simpler to set up. Vercel automatically sets `immutable` cache headers for the `assets/` folder of static Astro sites, so this would also ensure their default set-up covers everything.

Alternative would be to make PRs to Vercel and other providers for a more complex immutable caching set-up.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-p5qbcj?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

wulinsheng123:
Could I move all these files to the folder `assets `?
like this 
![image](https://user-images.githubusercontent.com/27261522/206135069-17bbbb1f-f49c-412f-886d-ac4905267264.png)

wulinsheng123:
I want to try fix this bug

delucis:
I think that’s one option, sure — or just move the _contents_ of `chunks/` into `assets/`. And we’d also need to make sure generated links to CSS, etc. end up pointing to the new location too.

wulinsheng123:
> I think that’s one option, sure — or just move the _contents_ of `chunks/` into `assets/`. And we’d also need to make sure generated links to CSS, etc. end up pointing to the new location too.

let me try to move all to the `assets` folder

wulinsheng123:
[...]

PR #5584, opened 2022-12-12
feat: change path into assets

wulinsheng123:
## Changes

fix bug #5502

### DIFF
--- a/.changeset/lovely-worms-invite.md
+++ b/.changeset/lovely-worms-invite.md
@@ -1,5 +1,8 @@
 ---
+'@astrojs/deno': major
+'@astrojs/netlify': major
+'@astrojs/image': minor
 'astro': major
 ---
 
-change assets of chunks path into assets folder
+Builds chunks into the `assets` folder. This simplifies configuring immutable caching with your adapter provider as all files are now in the same `assets` folder.
--- a/.changeset/lovely-worms-invite.md
+++ b/.changeset/lovely-worms-invite.md
@@ -1,5 +1,5 @@
 ---
-'astro': major
+'astro': patch
 ---
 
 change assets of chunks path into assets folder
--- a/.changeset/lovely-worms-invite.md
+++ b/.changeset/lovely-worms-invite.md
@@ -1,5 +1,5 @@
 ---
-'astro': patch
+'astro': major
 ---
 
 change assets of chunks path into assets folder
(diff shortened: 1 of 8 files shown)


.changeset/lovely-worms-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
```
delucis, line 2: Don’t know if this needs a major bump or not. If it does, we’ll need to wait a while before merging this. @matthewp & @bluwy are probably best place to confirm.

wulinsheng123: > Don’t know if this needs a major bump or not. If it does, we’ll need to wait a while before merging this. @matthewp & @bluwy are probably best place to confirm.

My bad,  I will chang into patch

bluwy: I think the changes look good, but this indeed seems like a major bump as it would invalidate all Astro projects cache, and affect integrations that relied on the old default paths (e.g. like in this PR the image, deno, netlify integrations).

We should probably do this for Astro 2.0, and also add changesets major for the image, deno, netlify integrations

delucis (dismissed):
Blocking until Astro v2. Needs major changesets as suggested by @bluwy.

.changeset/lovely-worms-invite.md
wulinsheng123: Ok, I will change it to major late because I have had a positive recently

packages/astro/src/core/build/static-build.ts
```diff
@@ -231,7 +231,7 @@ async function clientBuild(
 				output: {
 					format: 'esm',
 					entryFileNames: '[name].[hash].js',
```
SvizelPritula, line 233: Quick question: Why aren't entry files moved into `assets` as well?

matthewp: What do you think @wulinsheng123? That sounds right to me too.

natemoo-re (approved):
This LGTM! We decided to do an additional change and output to an `_astro` directory instead of `assets`, but I can make that change on top of these.

Many thanks for the PR!

bluwy (approved):

bluwy:
@natemoo-re maybe we should make those changes directly in this PR so we have one changeset that doesn't mention this intermediate change 🤔

natemoo-re:
@bluwy I was going to do that, but I think my changes will be quite involved. I'll be sure to update the changelog in my PR as well.

PR #5585, opened 2022-12-12
fix link 404 in Vue integration README

molychn:
## Changes

- Changed the link to the new URL due to vitejs/vite has moved framework plugins out of core.

### DIFF
--- a/packages/integrations/vue/README.md
+++ b/packages/integrations/vue/README.md
@@ -74,7 +74,7 @@ This package is maintained by Astro's Core team. You're welcome to submit an iss
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


molychn:
> Thanks for catching this @molychn and moving your PR from the docs repo — really appreciate it!
> 
> This looks good to me. One question: what do you think about linking to the package on npm instead? https://www.npmjs.com/package/@vitejs/plugin-vue-jsx
> 
> That way we wouldn’t be tied to changes in the repo structure.

@delucis , It is really a good way to loose the changes in the repo structure, I have changed it, by the way, U really teach me more.

delucis (approved):
Thanks!

PR #5590, opened 2022-12-12
Update contributing guide test-running examples

mattstein:
Existing example doesn’t work because RSS is handled in a separate package. (See [Discord](https://discord.com/channels/830184174198718474/845430950191038464/1051964090853117993).)

## Changes

This is a simple Markdown edit that changes `pnpm run test:match` example to one that works, adds an example of package filtering using that previous RSS case.

## Testing

Example commands were tested locally, but there’s too much irony in the fact that I can’t test the docs related to testing.

## Docs

This _is_ a documentation update. 😎


natemoo-re (approved):
🙌🏻 Thanks for the PR!

PR #5591, opened 2022-12-12
Fix missing RSS item `customData`

### LINKED ISSUE #5582: @astrojs/rss item-level `customData` omitted in 1.1.0

1.6.14

None

npm

macOS 13.0.1

### Describe the Bug

When I update the @astrojs/rss package to 1.1.0, the new RSSFeedItem’s `content` property works as expected but anything passed to `customData` is omitted from the rendered output.

It looks like the cause is [this commit](https://github.com/withastro/astro/commit/fe0da0185a85762ac5ac5bf66ea91947af1c329d#diff-976c8de545f5072ecad8b2e77b256cfd63086a70c5c8eb7feb2628b8471d0db3R166), which seems like it’s mistakenly looking at `rssOptions.customData` when it should be `result.customData`.

In the StackBlitz example, you can see a `<dc:creator>` tag for “Third Post” when the RSS package is pinned to 1.0.3:

```xml
<item>
<title>Third post</title>
<link>https://example.com/blog/third-post/</link>
<guid>https://example.com/blog/third-post/</guid>
<dc:creator>Buster Bluth</dc:creator>
</item>
```

Updating to 1.1.0, that tag vanishes. (Which has no relationship to the `content` parameter.) Anything passed to `customData` will fail to appear in the rendered feed:

```xml
<item>
<title>Third post</title>
<link>https://example.com/blog/third-post/</link>
<guid>https://example.com/blog/third-post/</guid>
</item>
```

https://stackblitz.com/edit/github-7ijnap-7fzwp3?file=src/pages/rss.xml.js

### Participation

mattstein:
## Changes

This fixes #5582 by basically correcting a typo, but it also adds a test so RSS item `customData` is slightly harder to break in the future.

## Testing

I added another variation of hardcoded RSS output that follows my use-case of appending a `<dc:creator>` node for the item. Adds a top-level namespace to support the node, uses `RSSFeedItem`’s `customData` property to append it, and makes sure the resulting output includes both.

## Docs

Nope! Just fixes a regression. When merged, the feature will work as currently described.

### DIFF
--- a/.changeset/ninety-socks-chew.md
+++ b/.changeset/ninety-socks-chew.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/rss': minor
+---
+
+Fixes a bug that prevented an item’s `customData` from being included.
--- a/packages/astro-rss/src/index.ts
+++ b/packages/astro-rss/src/index.ts
@@ -163,8 +163,9 @@ export async function generateRSS({ rssOptions, items }: GenerateRSSArgs): Promi
 		if (typeof result.content === 'string') {
 			item['content:encoded'] = result.content;
 		}
-		if (typeof rssOptions.customData === 'string')
-			Object.assign(item, parser.parse(`<item>${rssOptions.customData}</item>`).item);
+		if (typeof result.customData === 'string') {
+			Object.assign(item, parser.parse(`<item>${result.customData}</item>`).item);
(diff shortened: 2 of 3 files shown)


matthewp (approved):
This is great, thank you!

PR #5594, opened 2022-12-13
feat(cli): add `--output` flag

liruifengv:
## Changes

add `--output` CLI flag

```
astro build --output static
```


delucis:
I’m curious, how you’re planning to use this? Would you mind explaining your use case a bit?

If you pass `--output server`, Astro would also need to have an adapter for the deployment service you’re using — `node()`, `vercel()`, `cloudflare()` etc. — in `astro.config.mjs`. Is the main aim to toggle _off_ an SSR build?

Wondering if using environment variables as is already possible today might be a better fit for that use case:

```js
// astro.config.mjs

export default {
  output: process.env.BUILD_OUTPUT === 'ssr'
    ? 'server'
    : 'static',
  adapter: process.env.BUILD_OUTPUT === 'ssr'
    ? cloudflare()
    : undefined,
}
```

Then you would run:

```bash
# builds an SSR site
BUILD_OUTPUT=ssr astro build
# builds a static site
astro build
```

liruifengv:
Yes, I want to turn off ssr builds when using github pages.

Using environment variables seems to solve this case, I will try it.

liruifengv:
It works. Thank you.

PR #5608, opened 2022-12-14
fix(astro/core): Do not automatically add trailing slash

### LINKED ISSUE #5604: automatically add trailing slash against astroConfig.site

1.6.15

None

yarn

Mac

### Describe the Bug

In astro.config.mjs, the value specified in defineConfig is forced to be inserted with a trailing slash. The reason for this is that the transform of zod is used in `core/config/schema.ts`.

https://github.com/withastro/astro/blob/d7da0996b6b80499a6f8f2c86a5cfc3e39f741d2/packages/astro/src/core/config/schema.ts#L72-L76

This state is not intended by the user, so that ogp, for example, requires an absolute path to be specified, so it is written as `${import.meta.env.SITE}/ogp.png`, but the actual built HTML will contain `https://example.com//ogp.png` is written in the actual built HTML. Browsers will interpret this URL as convenient, but it may conflict with various other implementations.

After listening to the intent, the `transform` implementation of zod should be removed if it is not needed, and the URL should be left as specified by the user. If this is necessary, the docs should describe this behavior.

However, if you change it, you have to think about how to do it, or it will cause a bug if you simply stop using the trailing slash.

https://github.com/konojunya/astro-import-meta-env-site

### Participation

konojunya:
## Changes

fix #5604

Fixed a problem in which trailing slashes were automatically inserted for URLs set as `site` key in astro.config.

## Docs


.changeset/twelve-cooks-tickle.md
```diff
@@ -0,0 +1,9 @@
+---
+'astro': minor
```
bluwy, line 2: I think this should be a major as some could rely on the old behaviour, which this PR would break them. Since we're starting the 2.0 beta now, this seems like a great time to merge this.

.changeset/twelve-cooks-tickle.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+Fixed a problem in which trailing slashes were automatically inserted for URLs set as 'site' key in astro.config.
+Implementers should be aware that, for example, setting `${import.meta.env.SITE}ogp.png` to load an OGP will be converted to `https://example.comogp.png` and the image will not be loaded. As a result, the image may not load.
+
+## What should consumer do?
+Consumer should search `import.meta.env.SITE` when raising to this version and change it if they are writing code that expects the trailing slash to be added automatically.
```
bluwy, line 9: I think we can simplify this description with something like:
```suggestion
A trailing slash will not be automatically appended to `import.meta.env.SITE`. Instead, it will be the value of the `site` config as is. This may affect usages of `${import.meta.env.SITE}image.png`, which will need to be updated accordingly.
```

.changeset/twelve-cooks-tickle.md
konojunya: @bluwy I'll fix soon

bluwy (approved):
LGTM. cc @matthewp, I think we still have some time to add this breaking change.

matthewp:
Yeah I concur, `site` should be what the user provides that we do not touch, as long as it is a valid URL. But since this is a breaking change I want to bring it up with core first.

matthewp (approved):

PR #5609, opened 2022-12-15
[Image] Make `alt` paramater optional

zadeviggers:
## Changes

- This fixes the `alt` parameter being required. 
- 
The actual code works fine without an `alt` attribute. This is solely an issue with the types.

Originally (pre `v0.12.x`) `alt` wasn't required. Introducing it as a required type is a breaking change and shouldn't have been done in a minor patch (although `@astrojs/image` is still pre `1.0.0`, so I don't think it's technically a semver violation). 

Aside from the semver issue, there are many cases where images might be used without requiring an alt text (e.g. a background image).

## Testing

I don't think any tests need to be updated.

## Docs

No docs are needed.

### DIFF
--- a/packages/integrations/image/src/lib/get-image.ts
+++ b/packages/integrations/image/src/lib/get-image.ts
@@ -11,7 +11,7 @@ import type { ImageMetadata } from '../vite-plugin-astro-image.js';
 
 export interface GetImageTransform extends Omit<TransformOptions, 'src'> {
 	src: string | ImageMetadata | Promise<{ default: ImageMetadata }>;
-	alt: string;
+	alt?: string;
 }
 
 function resolveSize(transform: TransformOptions): TransformOptions {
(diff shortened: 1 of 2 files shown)


bivens-dev:
Curious as to why you would want to do this? It’s been very standard practice to require alt text on images for many years now and is considered a fundamental accessibility requirement.

natemoo-re:
This was definitely intentional! Alt text on images _is_ best practice, and we want to enforce that best practice. You can always set `alt=""` if you want to silence the type errors, which will also signal that your image is purely decorative.


From MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-alt)
> Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it's decoration or a tracking pixel), and that non-visual browsers may omit it from [rendering](https://developer.mozilla.org/en-US/docs/Glossary/Rendering_engine). Visual browsers will also hide the broken image icon if the alt is empty and the image failed to display.

PR #5611, opened 2022-12-15
Use less memory in .astro compilation

matthewp:
## Changes

- Reduce memory usage in the .astro vite plugin in 2 ways:
  - Avoid creating extra closures in style preprocessing (this was the biggest reduction).
  - Avoid temporary objects and use function arguments instead.
  - Avoid object spread which creates larger objects just to add a couple of properties to an object.

## Testing

- Tests should all still pass
- Tested with an example app to reduce memory.

## Docs

N/A

### DIFF
--- a/.changeset/real-apples-mix.md
+++ b/.changeset/real-apples-mix.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Reduce memory usage in .astro compilation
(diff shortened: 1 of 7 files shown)


natemoo-re (commented):
Code LGTM but there's a build error!

matthewp:
Yep, working on other stuff but I'll get back to this, it's non-critical.

natemoo-re:
@matthewp is this something we're still interested in getting over the line?

matthewp:
I would like to, but given the conflicts and not keeping an old PR around, I can fix later and resubmit.

PR #5612, opened 2022-12-16
fix: filter out draft item from glob result

### LINKED ISSUE #5505: @astrojs/rss includes drafts in output

1.6.12

None

npm

Mac

### Describe the Bug

When an item is marked as a draft in the frontmatter, it is still included in the XML output. You can see this behavior in the default blog example.

They should be excluded by default, and ideally added based on [integration configuration or build flags](https://docs.astro.build/en/guides/markdown-content/#enable-building-draft-pages).

https://stackblitz.com/edit/github-exbbav?file=src%2Fpages%2Fblog%2Ffirst-post.md,dist%2Frss.xml&on=stackblitz

### Participation

equt:
## Changes

- This PR should fix #5505 
- In brief, RSS will now respect the `draft` in the Markdown/MDX frontmatter

## Testing

A corresponding test has been added to ensure that any Markdown/MDX has `draft: true` in its frontmatter will be ignored, and a missing one or `false` will not.

## Docs

I'm not sure if this is a new feature or just a fix. Maybe we should add a note in the RSS doc?


.changeset/modern-dodos-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/rss': patch
```
delucis, line 2: This would probably need to be a major version change as it would break someone’s site if they currently have posts with `drafts: true` and expect those to be included in output.

```suggestion
'@astrojs/rss': major
```

If we don’t think this merits a v2 release of the RSS package, one alternative would be to make _including_ the drafts the default for backwards compatibility and introduce filtering behind an option. _Not_ including drafts does seem like the sensible default though, so perhaps a v2 is warranted.

delucis (commented):
Thanks @equt.

Could we make this configurable via an option to the `rss()` function? Or perhaps follow the [top-level `markdown.drafts` option](https://docs.astro.build/en/reference/configuration-reference/#markdowndrafts) although not sure we can easily access that in this context.

I know some people like to do stuff like include drafts when running `astro dev` to test stuff out, but exclude them in `astro build`, so making this configurable would support that use case.

equt:
Agree with that, but I can't figure out a way to get that option in the top-level config unless we parse the `astro.config` file again in the RSS package.

🤔 How about we take a step back instead? Rather than filtering for the user, we [expose the `draft`](https://github.com/withastro/astro/pull/5612/files#diff-976c8de545f5072ecad8b2e77b256cfd63086a70c5c8eb7feb2628b8471d0db3R46-R47) in the `RSSFeedItem`. This would allow something like a `preview` mode for previewing draft posts, and `release` mode for excluding all drafts.

fflaten:
> Agree with that, but I can't figure out a way to get that option in the top-level config unless we parse the `astro.config` file again in the RSS package.
> 
> 🤔 How about we take a step back instead? Rather than filtering for the user, we [expose the `draft`](https://github.com/withastro/astro/pull/5612/files#diff-976c8de545f5072ecad8b2e77b256cfd63086a70c5c8eb7feb2628b8471d0db3R46-R47) in the `RSSFeedItem`. This would allow something like a `preview` mode for previewing draft posts, and `release` mode for excluding all drafts.

There's really no need to expose anything in `RSSFeedItem` is it? You can already get `draft` directly from frontmatter. Ex.

```js
const postImportResult = import.meta.glob('./posts/**/*.{md,mdx}', { eager: true });
const posts = Object.values(postImportResult).filter((post) => import.meta.env.DEV || !post.frontmatter.draft);
```

If anything, I'd prefer `RSSOptions` to have a `includeDrafts`-bool like @delucis suggested. However, having content collections in mind, filtering is even easier (see below) + users might rename the `draft`-frontmatter/data to `isDraft` etc. So maybe not the best time to hardcode a filter at all, but rather include it in docs? 

```js
const posts = (await getCollection('blog', ({ data }) => {
    return import.meta.env.DEV || !data.draft
  }));
```

equt:
> There's really no need to expose anything in RSSFeedItem

Indeed, we don't need to expose the `draft` again, and somehow I just missed that.

> users might rename `draft` to `isDraft` etc

Considering the markdown integration has already hardcoded this, I guess most users will just go with it.

> So maybe not the best time to hardcode a filter at all, but rather include steps in docs?

We have to do something by default. Users who have put the `draft: true` in their posts will always expect the draft content to be filtered out.

delucis:
Fair enough if we can’t get the top-level Markdown config option.

In that case I’d still suggest adding an option to the `getRSS()` function, so you could do:

```js
import rss from '@astrojs/rss';

export const get = () => rss(
  drafts: true, // Include drafts
  // other options ...
});
```

delucis (approved):
Nice! LGTM — thanks @equt 🙌 

Would love a second opinion from the core team on whether this is in fact a major bump (the safe choice) or if we can get away with calling this a bug fix.

packages/astro-rss/src/index.ts
```diff
@@ -28,6 +28,8 @@ type RSSOptions = {
 	stylesheet?: string | boolean;
 	/** Specify custom data in opening of file */
 	customData?: string;
+	/** Whether to include drafts or not */
+	drafts?: boolean;
```
bholmesdev, line 32: nit: I'd prefer the name `includeDrafts` for naming a boolean option

bholmesdev (approved):
Left a nit on naming, but I'm happy with this PR's final `draft`!

To be a good semver samaritan, I'd stick with a `major` for this one. We _could_ switch the default of `includeDrafts` to `true` if we want to merge as a patch. Still, we'd probably flip back to `false` soon after for a major release, so that doesn't seem appealing to me.

bholmesdev:
> So maybe not the best time to hardcode a filter at all, but rather include steps in docs? 🙂

@fflaten Great point on Content Collections! I think the RSS integration will see some changes as content collections are more widely adopted. A loose set of ideas:
- Remove the `import.meta.glob` auto-resolver
- Introduce an `rssSchema` people can apply to their existing content config. This allows rich error messages on which properties are required by RSS, instead of the opaque "random entry broke RSS" error you see today:

```ts
import { rssSchema } from '@astrojs/rss';

  schema: {
    ...rssSchema,
    /* other properties you may want */
  }
});
```

Excited to work on this more with the community!

packages/astro-rss/src/index.ts
delucis: I like this but it would then be inconsistent with our own markdown config which uses `drafts: true` to build draft files as pages.

I guess we _could_ also rename the markdown flag in the future although `markdown.includeDrafts` is a less obvious home run in terms of naming.

bholmesdev: Fair enough! I personally didn't make the connection between the markdown config and the RSS config, so the different naming didn't bother me. I'm okay keeping it as `drafts` if it makes documentation easier.

tony-sull:
This looks great!  One quick note on the semver question - this should either be a `major` (if drafts are filtered out by default) or a `minor` (if drafts are included by default)

The second option is backwards compatible, but since it's really a new feature instead of a bug fix it's better to avoid using a `patch` release for it :+1:

sarah11918 (approved):
Approving the README on behalf of Astro Docs Maintainers!  🥳

PR #5615, opened 2022-12-16
Sanitize dynamic tags

natemoo-re:
## Changes

- Updates `renderComponent` to sanitize dynamic tags before printing
- This was an unintentional way to inject attributes, not a feature. This PR will close off that security hole.

## Testing

Added a unit test

## Docs

N/A, just a security patch

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -239,12 +239,14 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 	// This is a custom element without a renderer. Because of that, render it
 	// as a string and the user is responsible for adding a script tag for the component definition.
 	if (!html && typeof Component === 'string') {
+		// Sanitize tag name because some people might try to inject attributes 🙄
+		const Tag = Component.trim().split(/\s+/)[0].trim();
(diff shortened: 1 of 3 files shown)


packages/astro/src/runtime/server/render/component.ts
```diff
@@ -239,12 +239,14 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 	// This is a custom element without a renderer. Because of that, render it
 	// as a string and the user is responsible for adding a script tag for the component definition.
 	if (!html && typeof Component === 'string') {
+		// Sanitize tag name because some people might try to inject attributes 🙄
+		const Tag = Component.trim().split(/\s+/)[0].trim();
```
matthewp, line 243: Can we do an `indexOf(' ')` first to check that the tag name contains a space? This code looks a little expensive so would be good for well-intended code.

matthewp:
!bench

packages/astro/src/runtime/server/render/component.ts
natemoo-re: I switched this to a regex check because I realized there are other cases than whitespace.

natemoo-re:
!bench

matthewp (approved):

PR #5626, opened 2022-12-17
Docs: Fix `prerender` error message

natemoo-re:
## Changes

- Updates `message` to avoid stringifying the function.
- Updates `description` for clarity.

## Testing

N/A

## Docs

This is only a docs change

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -397,7 +397,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	 * @docs
 	 * @message
 	 * **Example error messages:**<br/>
-	 * InvalidPrerenderExport: A \`prerender\` export has been detected, but its value cannot be statically analyzed.
+	 * InvalidPrerenderExport: A `prerender` export has been detected, but its value cannot be statically analyzed.
 	 * @description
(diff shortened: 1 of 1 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -395,8 +395,11 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	},
 	/**
 	 * @docs
+	 * @message
+	 * **Example error messages:**<br/>
+	 * InvalidPrerenderExport: A `prerender` export has been detected, but its value cannot be statically analyzed.
 	 * @description
-	 * A `prerender` export was detected, but the value was not statically analyzable. Values computed at runtime are not supported, so `export const prerender` can only be set to `true` or `false`. Variables are not supported.
+	 * The `prerender` feature only supports a subset of valid JavaScript—be sure to use exactly `export const prerender = true` so that our compiler can detect this directive at build time. Variables, `let`, and `var` declarations are not supported.
```
sarah11918, line 402:
```suggestion
	 * The `prerender` feature only supports a subset of valid JavaScript — be sure to use exactly `export const prerender = true` so that our compiler can detect this directive at build time. Variables, `let`, and `var` declarations are not supported.
```
I suspect the spaces will help for readability. (These could also be two separate sentences.) But otherwise, I like the explicit "use exactly this" message!

sarah11918 (approved):

matthewp:
Changeset?

matthewp:
Oh, I guess that's not needed for docs?

aFuzzyBear (approved):
A lot more clearer 👍

PR #5633, opened 2022-12-18
Add JSDoc comments for error overlay to match other experimental flags

delucis:
## Changes

Add configuration docs for the `experimental.errorOverlay` flag to match those for `prerender` and `contentCollections`.

Currently in docs we render out those other flags but not the `errorOverlay` flag: https://docs.astro.build/en/reference/configuration-reference/#experimental-flags

## Testing

n/a docs change only

## Docs

Yes. Is there a reason we wouldn’t want to document this like we do other experimental features?

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -907,8 +907,23 @@ export interface AstroUserConfig {
 	 */
 	experimental?: {
 		/**
-		 * @hidden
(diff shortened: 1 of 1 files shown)


sarah11918 (approved):
Approving the text on behalf of Docs Maintainers!

bluwy (approved):

PR #5638, opened 2022-12-19
fix traced warnings from unexpected token

andreademasi:
This PR fixes building a website when using @astrojs/vercel/serverless . It's a temporary fix until https://github.com/vercel/nft/issues/311 gets fixed

## Changes

- Added an if check to ignore "Failed to parse" errors from @vercel/nft

## Testing
I don't think tests are needed for this small change, if they are let me know.

## Docs
It shouldn't affect user behavior

### DIFF
--- a/.changeset/lazy-tomatoes-call.md
+++ b/.changeset/lazy-tomatoes-call.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': minor
+---
+
+Ignore warnings when traced file fails to parse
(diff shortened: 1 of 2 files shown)


JuanM04 (approved):
Thank you very much!
Before merging, could you leave the imports as they were? Just a styling preference :smile:

andreademasi:
> Thank you very much! Before merging, could you leave the imports as they were? Just a styling preference smile

Done!

Issue #5624, opened 2022-12-17
@astro/vercel: Astro emits code that has `client:only` on SSR since vite 3.2.5, when importsNotUsedAsValues is set


WesSouza:
### What version of `astro` are you using?

1.7.1 (broke starting on 1.6.13)

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

When using `tsconfig.json`'s `"importsNotUsedAsValues": "error"` option, as set on [packages/astro/tsconfigs/strict.json](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/strict.json#L7), building with the `@astro/vercel` integration emits code for components that have `client:only="react"` inside the `vercel/output/functions/render.func/dist/entry.js` serverless function.

That is incorrect and causes the runtime to break when packages are incompatible with server rendering.

With that setting removed, the code is not present.

Working example:

[WesSouza/astro-vite-vercel-issue](https://github.com/WesSouza/astro-vite-vercel-issue)
https://astro-vite-vercel-issue.vercel.app/

Breaking change:

https://github.com/WesSouza/astro-vite-vercel-issue/pull/2
https://astro-vite-vercel-issue-git-broken-wessouza.vercel.app/
[Logs are public here](https://vercel.com/wessouza/astro-vite-vercel-issue/7oCJp4uxBMrdD5n55zPSCmZiY1u9/functions)

Runtime error on Vercel:
```
ERROR	Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/var/task/node_modules/date-fns/formatDistanceToNow' is not supported resolving ES modules imported from /var/task/dist/entry.js
Did you mean to import date-fns/formatDistanceToNow/index.js?
    at new NodeError (node:internal/errors:393:5)
    at finalizeResolution (node:internal/modules/esm/resolve:324:17)
    at moduleResolve (node:internal/modules/esm/resolve:965:10)
    at moduleResolveWithNodePath (node:internal/modules/esm/resolve:909:12)
    at defaultResolve (node:internal/modules/esm/resolve:1173:79)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:841:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:18)
[...]

Issue #5541, opened 2022-12-06
Astro 1.6.13 ReferenceError: window is not defined when use with stickthead module


patrickdung:
### What version of `astro` are you using?

1.6.13

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

Astro 1.6.12 is fine. It has problem when upgrade to Astro 1.6.13
I use this module https://www.npmjs.com/package/stickythead

 building client

🌼 daisyUI components 2.42.1  https://github.com/saadeghi/daisyui
  ✔︎ Including:  base, components, themes[29], utilities

8:24:23 PM [vite-plugin-svelte] dom compile done.
package                 files    time      avg
cloudnative-quest          16   0.17s   10.8ms
Completed in 3.37s.

```
 generating static routes
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

ReferenceError: window is not defined
    at Object.<anonymous> (/data/projects/webdev/astro/projects/cloudnative-quest-astro/node_modules/stickythead/dist/stickythead.js:1:204)
    at Module._compile (node:internal/modules/cjs/loader:1120:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1174:10)
    at Module.load (node:internal/modules/cjs/loader:998:32)
    at Module._load (node:internal/modules/cjs/loader:839:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:170:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
    at async generatePages (file:///data/projects/webdev/astro/projects/cloudnative-quest-astro/node_modules/astro/dist/core/build/generate.js:52:20)
    at async staticBuild (file:///data/projects/webdev/astro/projects/cloudnative-quest-astro/node_modules/astro/dist/core/build/static-build.js:61:5)
    at async AstroBuilder.build (file:///data/projects/webdev/astro/projects/cloudnative-quest-astro/node_modules/astro/dist/core/build/index.js:86:5)
    at async AstroBuilder.run (file:///data/projects/webdev/astro/projects/cloudnative-quest-astro/node_modules/astro/dist/core/build/index.js:127:7)
[...]

PR #5639, opened 2022-12-19
Fix client only import with importsNotUsedAsValues error

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/5541
fix https://github.com/withastro/astro/issues/5624

If a framework component as side effects that access browser globals, and is used as `client:only`, Astro should treeshake the import of that component import to prevent the side effect from running in SSR.

#5553 accidentally break this as `"importsNotUsedAsValues": "error"` is respected when specified in the tsconfig, causing the import to be preserved and not treeshaken.

This PR forces `"importsNotUsedAsValues": "error"`

## Testing

updated client-only test

## Docs

n/a bug fix

### DIFF
--- a/.changeset/two-pants-enjoy.md
+++ b/.changeset/two-pants-enjoy.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix `client:only` imports with `"importsNotUsedAsValues": "error"` tsconfig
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
Clean fix, thanks!

Issue #5643, opened 2022-12-19
Astro 1.7.x deletes directory `./chunks` completely from `dist` after SSG


christian-hackyourshack:
### What version of `astro` are you using?

1.7.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

After SSG, the new Astro erases `./chunks` completely from the `dist` directory. But some `hoisted.<checksum>.js` files reference into chunks for some shared code. This code is then not available in production on the client-side. In dev mode it still works fine.

My hoisted files look like this:

```js
import"./chunks/index.astro_astro_type_script_index_0_lang.c5b3d70c.js";const l=document.querySelectorAll("[data-picture-selectable]");l.forEach(e=>{e.addEventListener("click",()=>c(e.id))});function c(e){const t=document.querySelector(`[data-id="${e}"]`);t&&(t.style.display="flex")}const n=document.querySelectorAll(".picture--full");n.forEach(e=>{e.addEventListener("click",()=>e.style.display="none")});
```

I assume, this is closely related to #5632 and #5621.

### Link to Minimal Reproducible Example

This would need a bigger example, because small ones are inlined.

### Participation

- [ ] I am willing to submit a pull request for this issue.

Issue #5632, opened 2022-12-18
Version 1.7.1 breaks client:only vue components.


songololo:
### What version of `astro` are you using?

1.7.1

### Are you using an SSR adapter? If so, which one?

Vue

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Everything works fine in dev mode and seems to build cleanly. However, when running preview against a built project then `client:only` vue components don't load.

This is the browser error:
<img width="1320" alt="Screenshot 2022-12-18 at 11 53 43" src="https://user-images.githubusercontent.com/4347878/208296948-d963958e-d1d0-41e8-8803-bc11f254a258.png">

Downgrading to version 1.6 fixes the issue.

I have not had a chance to create a minimal reproducible example so please feel free to close if you wish. However, this issue should probably be noted by the maintainers as the current release introduces breaking behaviour.

### Link to Minimal Reproducible Example

none provided

### Participation

- [ ] I am willing to submit a pull request for this issue.

Issue #5621, opened 2022-12-17
TypeScript files in <script> don't run on Vercel


aleksa-codes:
### What version of `astro` are you using?

1.7.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Windows

### Describe the Bug

After upgrading the `mdx`, `rss` integrations and `astro` the website works on the local machine but doesn't show up when deployed to Vercel.

I upgraded my other website which is deployed with GitHub pages and it works just fine.

I tried rolling back to the previous dependencies which worked before and now they don't even work, kinda strange, I am probably missing something.

Here is the latest commit, upgrading dependencies:
https://github.com/aleksa-codes/astro-portfolio/commit/b568d7913cdf8d7ced615c8f381f68e7b7b292df

Website when everything was working, state to which I tried reverting back with old dependencies versions but failed:
https://aleksacodes-e7dmuw6l9-aleksacodes.vercel.app/

Here is what the console returns:

![Capture](https://user-images.githubusercontent.com/64015178/208241117-c0244382-8ee6-4776-92f8-1fe9ea5d1dd7.PNG)

Also Vercel dashboard usually displayed my favicon, now I got this:

![astro](https://user-images.githubusercontent.com/64015178/208240545-58ecb80e-4e10-4fd5-a2fc-1a8e06b76a45.png)

Edit:
Just tried upgrading Node version from 16 to 18.12.1 LTS on Vercel and locally and the build fails, that's maybe even more concerning than my current problem. But I rolled it back to 16, so this could get solved first.

Error message with Node 18:

![error-astro](https://user-images.githubusercontent.com/64015178/208242337-e12d72f6-08a1-4726-9cc2-5708f13f9e44.png)

### Link to Minimal Reproducible Example

https://aleksa.codes

### Participation

- [ ] I am willing to submit a pull request for this issue.

aleksa-codes:
After some testing looks like none of my scripts using TypeScript are working anymore when deployed to Vercel.

One of the scripts I have is a FOUC fix that applies `document.body.style.visibility = 'visible';` after the DOM is ready, before that I have style='visibility: hidden;' applied to `body` and that removes FOUC in most of the cases. That is why there is no content seen on my website because the script is never run.

I tried commenting out everything about the FOUC fix and leaving the rest, content shows now, but then the .ts script for snow, copyToClipboard and typewriter don't run as well.

aleksa-codes:
[...]

PR #5644, opened 2022-12-19
Patch static build regression

natemoo-re:
## Changes

- Can't say I fully understand what changed this behavior, but it looks like due to some timing changes in the build, the `chunks` directory is empty before it's populated by the client build.
- If we remove the `chunks` directory because it is empty, the client assets also disappear.
- Closes https://github.com/withastro/astro/issues/5643, closes https://github.com/withastro/astro/issues/5632, closes https://github.com/withastro/astro/issues/5621

## Testing

Tested manually, need to figure out how to test this

## Docs

Bug fix only

### DIFF
--- a/.changeset/small-camels-mix.md
+++ b/.changeset/small-camels-mix.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix static build regression where chunks would not be generated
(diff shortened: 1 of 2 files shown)


bholmesdev (approved):
Good `chun`do on that 👍

aFuzzyBear (commented):
Another example of `explicit` > `implicit` 😉

PR #5648, opened 2022-12-19
[Content] Throw on relative image usage

bholmesdev:
## Changes

We [decided relative image resolution was out-of-scope](https://github.com/withastro/rfcs/blob/main/proposals/0027-content-collections.md#out-of-scope--future) for the Content Collections RFC. Throw on relative image paths in `src/content/` for Markdown and MDX.

## Testing

Manual testing

## Docs

N/A - Already documented that Markdown and MDX are the only allowable files.


packages/integrations/mdx/src/plugins.ts
```diff
@@ -112,6 +114,30 @@ export function rehypeApplyFrontmatterExport(pageFrontmatter: Record<string, any
...
+		const relImagePaths = new Set<string>();
+		visit(tree, 'image', function raiseError(node: Image) {
+			if (isRelativePath(node.url)) {
+				relImagePaths.add(node.url);
+			}
+		});
+		if (relImagePaths.size === 0) return;
+
+		const errorMessage =
+			`Relative image paths are not support in the content/ directory. Please update to absolute paths:\n` +
```
natemoo-re, line 134: It's not super clear what "absolute paths" means here... can we be more helpful?

pnpm-lock.yaml
```diff
@@ -14770,6 +14774,11 @@ packages:
       unist-util-visit: 3.1.0
     dev: true
 
+  /mdast/3.0.0:
+    resolution: {integrity: sha512-xySmf8g4fPKMeC07jXGz971EkLbWAJ83s4US2Tj9lEdnZ142UP5grN73H1Xd3HzrdbU5o9GYYP/y8F9ZSwLE9g==}
+    deprecated: '`mdast` was renamed to `remark`'
```
natemoo-re, line 14779: 🤔 Hmm, can you get `Image` from `remark` or is it only available in the deprecated package?

packages/integrations/mdx/src/plugins.ts
bholmesdev: Ah, tough to be concise yet useful when we don't have the luxury of `AstroError`. I'll redraft

pnpm-lock.yaml
bholmesdev: Ah, [looks like the remark docs recommend `@types/mdast` without installing the core dep.](https://github.com/remarkjs/remark#types) Only `mdast` is deprecated while the types are stable, so I uninstalled that.

packages/integrations/mdx/src/plugins.ts
natemoo-re: I guess I'd just be a bit confused. Do we mean to put images in `public/` and use a base-relative absolute URL?

packages/markdown/remark/src/remark-content-rel-image-error.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function isRelativePath(path: string) {
+	return startsWithDotDotSlash(path) || startsWithDotSlash(path);
+}
+
+function startsWithDotDotSlash(path: string) {
+	const c1 = path[0];
+	const c2 = path[1];
+	const c3 = path[2];
+	return c1 === '.' && c2 === '.' && c3 === '/';
```
andersk, line 40: FYI: `path.startsWith('../')` (ES6, Node.js ≥ 4).

packages/markdown/remark/src/remark-content-rel-image-error.ts
```diff
@@ -0,0 +1,47 @@
...
+import { visit } from 'unist-util-visit';
+import type { VFile } from 'vfile';
+
+/**
+ * `src/content/` does not support relative image paths.
+ * This plugin throws an error if any are found
+ */
+export default function remarkContentRelImageError() {
+	return (tree: any, vfile: VFile) => {
+		if (!vfile.path.includes('content/')) return;
```
andersk, line 11: This could have false positives—would ``vfile.path.startsWith(`${vfile.cwd}/src/content/`)`` be more precise?

bholmesdev: Ah, stubbed that out and thought I fixed it 🤦‍♂️ We'd probably pull in the `srcDir` from the user config and join up here. Will update!

packages/integrations/mdx/src/plugins.ts
bholmesdev: Yep, `public` should be the recommendation here 👍

packages/markdown/remark/src/remark-content-rel-image-error.ts
bholmesdev: This is pulled directly from our core utility, so don't want to modify!

natemoo-re: The reason we do it this way is that it's a very tiny perf optimization!

natemoo-re (approved):
LGTM!

natemoo-re:
> Can this check be optional? Right now relative paths are the only way to reference images from `*.md` that works with arbitrary values of `BASE_URL`. It’s not ideal that they’re relative to the output URL rather than the input file, but better than impossible.

I think unfortunately, this is why this setting _can't_ be optional. We're trying to protect against anyone from relying on the current unspecified behavior. The eventual plan is to specify the _correct_ behavior for relative paths, which likely means that they'll be resolved and treated as actual assets. That would also solve the `BASE_URL` issue, but have lots of other benefits as well.

PR #5651, opened 2022-12-20
Add canonical URL to blog example

ShivamJoker:
## Changes
- Adds canonical URL which is important for SEO
- Helps with page duplicacy (eg. www -> non-www)

## Testing
It was tested on my landing page 
image

## Docs

No docs needs to be updated

### DIFF
--- a/examples/blog/src/components/BaseHead.astro
+++ b/examples/blog/src/components/BaseHead.astro
@@ -9,6 +9,8 @@ export interface Props {
 	image?: string;
 }
 
+const canonicalURL = new URL(Astro.url.pathname, Astro.site);
+
 const { title, description, image = '/placeholder-social.jpg' } = Astro.props;
 ---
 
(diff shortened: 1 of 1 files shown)


aFuzzyBear (commented):
I can understand why you had added this, but as it being an example ,is this pr more to make the blog example as a template hold some seo as standard over allow the user to implement it themselves? 
Personally no objections to this PR, thank you for putting this in @ShivamJoker

ShivamJoker:
@aFuzzyBear if someone wants to start using this template but don't have much idea about SEO, they will miss the `canonical` link and there site will have duplication.

> is this pr more to make the blog example as a template hold some seo as standard over allow the user to implement it themselves?

Yes

ShivamJoker:
I have taken the example from astro docs itself.

aFuzzyBear:
awesome, thanks for the follow up, Ill wait till @delucis could take a wee look over this before merging,

bluwy (approved):
We also added the canonical link at

https://github.com/withastro/astro/blob/0b50987584120e0c0e549f9ff838dc8879dbfa30/examples/docs/src/components/HeadSEO.astro#L16-L17

so this seems fine to me.

delucis (approved):
Yup, looks good. These examples are supposed to be minimal but I think canonical URLs are important enough to include 👍

PR #5652, opened 2022-12-20
Use acorn to postprocess Astro globs

bluwy:
## Changes

Use `acorn` to process `Astro.glob`. This allows parsing BigInts as `esprima` does not support it. This should be more efficient too as `recast` reprints the AST while this only makes tweaks with MagicString.

## Testing

Existing glob test should pass. Enable BigInt test.

## Docs

n/a. Internal refactor and bug fix.


natemoo-re (approved):
LGTM, thanks for cleaning this up!

PR #5654, opened 2022-12-20
MD/MDX collect headings refactor

delucis:
## Changes

- [x] Consolidate MD/MDX collect headings logic
- [x] Expose `rehypeHeadingSlugs` to Astro users via `@astrojs/markdown-remark`
- [x] Flip order of execution for MDX: inject heading IDs _after_ user plugins to match Markdown behaviour
- [x] Reuse `rehypeHeadingSlugs` in MDX integration instead of copy/paste duplication
- [x] Minor releases for both `@astrojs/markdown-remark` and `@astrojs/mdx` (minor == breaking pre v1)

Supersedes #5646 

## Testing

Added tests for:

- being able to override ID generation with the default config
- being able to inject IDs early by using `rehypeHeadingSlugs` directly

## Docs

Todo: will probably require an update to the Markdown guide

### DIFF
--- a/packages/markdown/remark/src/types.ts
+++ b/packages/markdown/remark/src/types.ts
@@ -68,6 +68,12 @@ export interface MarkdownMetadata {
 	html: string;
 }
 
+export interface MarkdownVFile extends VFile {
+	data: {
+		__astroHeadings?: MarkdownHeading[];
+	};
+}
+
 export interface MarkdownRenderingResult {
 	metadata: MarkdownMetadata;
 	vfile: VFile;
--- a/.changeset/fast-baboons-prove.md
+++ b/.changeset/fast-baboons-prove.md
@@ -8,18 +8,18 @@ Run heading ID injection after user plugins
 
 If you are using a rehype plugin that depends on heading IDs injected by Astro, the IDs will no longer be available when your plugin runs by default.
 
-To inject IDs before your plugins run, import and add the `rehypeHeadingSlugs` plugin to your `rehypePlugins` config:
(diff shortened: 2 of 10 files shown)


packages/markdown/remark/src/types.ts
```diff
@@ -68,6 +68,12 @@ export interface MarkdownMetadata {
 	html: string;
 }
 
+export interface MarkdownVFile extends VFile {
+	data: {
+		__astroHeadings?: MarkdownHeading[];
+	};
```
bholmesdev, line 74:
```suggestion
	data: {
		__astroHeadings?: MarkdownHeading[];
		astro: {
      frontmatter: Record<string, any>;
    }
	};
```
Nit: if we're typing our VFile data, we can add a type for Astro frontmatter as well! See [Astro core type](https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts#L1506)

bholmesdev (approved):
remarkable 👏

packages/markdown/remark/src/types.ts
delucis: Hmm. This type is intentionally only used internally as it’s kind of a private API. Are you thinking someone might import `MarkdownVFile` and use it in their custom plugins as a convenience when injecting frontmatter?

bholmesdev: Hm, honestly wasn't sure. We can definitely ignore since `astro.frontmatter` is superfluous for the type's use case.

PR #5291, opened 2022-11-03
Content Collections

bholmesdev:
## Changes

- This introduces a new reserved directory called `src/content/`, with the helper functions `renderContent` and `fetchContent` to efficiently retrieve entries
- Adds a new `with-content` example to showcase these APIs

[See RFC for full details and usage instructions!](https://github.com/withastro/rfcs/pull/373)

## Testing

- Collection query tests
- `getStaticPaths` integration tests
- SSR integration tests
- `render()` style bleed tests

## Docs

Guide: https://github.com/withastro/docs/pull/2141
API reference: TODO

PR #5655, opened 2022-12-20
Unwrap templates that only include a single function expression

natemoo-re:
## Changes

- There was a change to how we handle slots in 1.7.x https://github.com/withastro/astro/pull/5291/files#diff-b28cb24c3ad433b2aad5f08fff9eea8039cb4d6f517d427f401890e09d4acec9R35-R38
- This had the effect of _wrapping_ function slots with a template instance, which means any tools **unsafely** relying on our function signature might break.
- Changing this behavior is totally valid and not covered by semver because these are internal APIs, but in this case the fix is straightforward and helps make our function signature easier to reason about. It's also a very tiny performance optimization.

## Testing

Added a unit test to catch regressions

## Docs

N/A, this is an entirely private API

### DIFF
--- a/.changeset/itchy-items-shave.md
+++ b/.changeset/itchy-items-shave.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix a regression in 1.7.x that changed the signature of the internal slots API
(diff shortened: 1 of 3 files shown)


natemoo-re:
!bench

Issue #5598, opened 2022-12-13
Astro’s dev server can't bundle when there is a space in a parent folder 


ffraenz:
### What version of `astro` are you using?

1.6.15

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

M1 macOS

### Describe the Bug

I just started to compose a bug report without being able to reproduce the issue outside my local environment when I found the culprit: An innocent space character in one of the parent folder's name.

Steps to reproduce:

1. Create a folder with a space in it, e.g. `mkdir Project\ Name` and change into it.
2. Run `npm create astro@latest` and select 'an empty project', say yes to 'install npm dependencies', say no to 'initialize a new git repository', choose the 'Strict' TypeScript setup, and change into the newly created folder.
3. Append the following lines to `src/pages/index.astro`:

   ```html
   <script>
     console.log('Hello, World.')
   </script>
   ```
4. Run `npm run dev`.
5. Visit `http://127.0.0.1:3000/` as advertised in the command line.

The following exception is thrown in the CLI, blocking the page from being rendered in the browser.

```
error   Could not import `/src/pages/index.astro?astro&type=script&index=0&lang.ts`.
  Hint:
    This is often caused by a typo in the import path. Please make sure the file exists.
  Stacktrace:
Error: failed to load module for ssr: /src/pages/index.astro?astro&type=script&index=0&lang.ts
    at instantiateModule (file:///Users/user/Projects/Project%20Name/majestic-mass/node_modules/vite/dist/node/chunks/dep-5605cfa4.js:53281:15)
```

Note that building the project (running `npm run build` and then `npm run preview`) works just fine and yields the expected result. Thus, this seems to be related to esbuild only.

After some work I was able to reproduce the issue on Stackblitz, so the issue is no longer restricted to my local environment.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-q52vqt?file=project%20name/src/pages/index.astro

(I needed to manually replace `project name` by `project%20name` as even Stackblitz does not seem to work properly with spaces)

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Bumping the priority on this as we're getting many many reports in our support threads

reksaandhika:
This also happens to me, when trying to update Astro version from 1.6.11 to latest version.
I'm using Windows 10.

PR #5666, opened 2022-12-22
Fix/content spaces

bholmesdev:
## Changes

- chore: bump to `github-slugger` 2.0. This migrates from CommonJS to ESM but does not introduce API changes ([see release notes](https://github.com/Flet/github-slugger/releases/tag/2.0.0)).
- Introduce `github-slugger` to calculate content slugs. Previously, we simply chopped the extension from the file name. Slugifying each segment should handle capitalization and spaces for use in `getStaticPaths`.
  - **Before:** `english translations/Using Mdx.mdx` -> `english translations/Using Mdx`
  - **After:** `en/Using Mdx.mdx` -> `english-translations/using-mdx`

## Testing

TODO

## Docs

TODO: update `slug` API reference and guide to reflect changes


packages/astro/src/content/types-generator.ts
```diff
@@ -118,13 +120,12 @@ export async function createContentTypesGenerator({
...
-			entry: event.entry,
-			contentDir: contentPaths.contentDir,
-		});
-		// Not a valid `src/content/` entry. Silently return.
-		if (entryInfo instanceof Error) return { shouldGenerateTypes: false };
 		if (fileType === 'unknown') {
+			const entryInfo = getEntryInfo({
+				entry: event.entry,
+				contentDir: contentPaths.contentDir,
+				allowFilesOutsideCollection: true,
```
Princesseuh, line 127: Is this intended? I'm not sure I quite follow the logic here being unfamiliar with this code part

bholmesdev: Ha, thought that would make this code _more_ readable instead of less 😅 
In short: we'll allow underscore `_` files outside of a collection (aka files with the `unknown `type). If that's unclear looking at the context, I can refactor another way.

Princesseuh: Oh no, it's fine. I just was missing some context on the changes being done here.

I do believe some comments would maybe make this a bit more readable, but that's outside the scope of this PR, in my opinion

Princesseuh (approved):
Looks good to me! Appreciate the tests, awesome work

packages/astro/src/content/types-generator.ts
bholmesdev: Gotcha, comment added 👍

.changeset/stupid-shoes-complain.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': minor
```
matthewp, line 2: This seems like a bug fix to me rather than a new feature. Am I wrong?

bholmesdev: This is changing the way we compute `slug`s based on today's documentation. I chose a minor since this is breaking!

matthewp: Ok, I think this is fine for now. We should probably discuss as a team how to handle semver for experimental features. In the beta period it will be equivalent to patches I believe.

matthewp (approved):
code looks good, just need to resolve merge conflicts

.changeset/stupid-shoes-complain.md
surjithctly: Experimental features should not be coverd by `semver` right? I read this from Next.js

surjithctly:
Does this PR closes issue with spaces in a parent folder name? eg: `./My Works/astro/project`

The error was: 
```
[plugin:astro-delayed-asset-plugin] ENOENT: no such file or directory, open '/Users/My%20Works/astro/project/src/content/blog/article-name.md'
```

bholmesdev:
@surjithctly Yes it should! Probably worth an extra unit test for sanity though 👍

surjithctly:
Sadly, this doesn't fix the space issue in a parent folder. 👎

I think there should be a unit-test as you said to make sure it doesn't have a regression.

bholmesdev:
@surjithctly Well, you're right there 😓 I made the compromise yesterday of merging this fix as-is to ship our minor release, since this PR _does_ solve other existing problems with content collections. Was worth delaying our release for further testing in hindsight.

It also looks like an issue outside of content collections after seeing #5598. We can follow-up there as we investigate a fix!

PR #5698, opened 2022-12-29
[Content Collections] Move config and generated types to project root

bholmesdev:
## Changes

Moves the Content Collections config and generated types out of `src/content/` and to the project root.

### Why move these?

We process your content collections config separately from other files in `src`. This means Vite transformations **will not be run** as with other modules in `src`. For instance, this prevents importing constants from `.jsx` or `.mdx` files. Allowing the content config file to exist in `src` makes this feel less similar to the `astro.config` and more similar to a `utils` file.

We've also received support forum confusion over `types.generated` appearing in a `src` directory. Is this expected behavior? Should I ignore this file?

To avoid confusion on each of these points, I propose moving the content config and generated types to your project root:

**Before**

```bash
src/content/
  blog/
  config.ts
  types.generated.d.ts
astro.config.mjs
```

**After**

```bash
src/content/
  blog/
types.generated.d.ts
content.config.ts
astro.config.mjs
```

## Testing

Update test fixtures and generated types tests to respect new file locations.

## Docs

https://github.com/withastro/docs/pull/2257


surjithctly:
Is `types.generated.d.ts` really needed? Can't we avoid it? or move it to a dedicated cache directory like `.astro` so that we can gitignore them. If its really needed in the root, how about we rename them to `types.content.d.ts` or simply `types.d.ts`

bluwy:
@surjithctly That could work too. I'm not sure what else we may want to put in `.astro` in the future, but I think it's a futureproof solution if having it in `src/contents` is confusing. (Though most Astro templates have `src/env.d.ts` so I don't quite find it confusing)

Re `contents.config.ts`, it does feel a bit "unofficial" to me (naming-wise). Perhaps we can have it as `astro-content.config.ts` so the file explorer should show them side-by-side, it's a long name though. `astro.content.ts`? 🤔

bholmesdev:
Closing this PR per our community call discussion! Next steps:
1. Draft a plan to move `types.generated.d.ts` to a `.astro` directory, as in the original RFC
2. Keep the content config where it lives today. If the community has suggestions on this before the feature is baselined, [open a discussion](https://github.com/withastro/rfcs)!

Issue #5535, opened 2022-12-05
node ssr dont showing 404 page


polioan:
### What version of `astro` are you using?

1.6.12

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

Windows (but linux have same behavior)

### Describe the Bug

This behavior is similar on standalone and middleware mode. When using vite dev server - all ok, default or custom 404 page are showing. But when using build Instead of 404 page - plain text are showing.

run "npm run serve" for starting server in production (build) mode

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-cddzmd

### Participation

- [ ] I am willing to submit a pull request for this issue.

wulinsheng123:
mark

wulinsheng123:
![image](https://user-images.githubusercontent.com/27261522/209598254-6b2ddd33-33ac-4f4a-ac4f-cb91faad128a.png)

Does this page only appear in the dev?
Does It not appear in the Astro bundle?

bluwy:
@wulinsheng123 Yes I think that's intentional for dev only if the user doesn't provide a custom 404 page. For production, it should default to what the server would respond for 404.

Looking into the issue, looks like this is because we don't pass `{ matchNotFound: true }` [here](https://github.com/withastro/astro/blob/0b50987584120e0c0e549f9ff838dc8879dbfa30/packages/integrations/node/src/middleware.ts#L13). But we can't pass it because it might be a middleware, other middlewares might want to handle the request for something else instead of 404.

I think the fix here is that, if we're building for `standalone` mode, we set `{ matchNotFound: true }`. For `middleware` mode, we can leave it as is, but a nice bonus feature is if the server build output export a second middleware that handles 404.html serving only.

wulinsheng123:
Thanks for your tips.

bluwy:
Fixed in #5701

schroedan:
@bluwy Well, the problem was only half fixed. The 404 handler is missing in the middleware module exports. We can work around this in our case by building for `standalone` mode, setting `ASTRO_NODE_AUTOSTART` env to `disabled` and using the handler from the `standalone` module. But this is not a very clean way in my opinion.

wulinsheng123:
[...]

PR #5701, opened 2022-12-30
can jump 404 when that page does not exist

wulinsheng123:
## Changes

fix bug #5535


packages/integrations/node/src/middleware.ts
```diff
@@ -2,16 +2,17 @@ import type { NodeApp } from 'astro/app/node';
 import type { IncomingMessage, ServerResponse } from 'http';
 import type { Readable } from 'stream';
 import { responseIterator } from './response-iterator';
+import type { Options } from './types';
 
-export default function (app: NodeApp) {
+export default function (app: NodeApp, options: Options) {
```
bluwy, line 7: It would be great if we pass `mode` instead of `options` since we only use the `mode`.

bluwy (commented):
I have one nit but this looks good to me! Could we add some tests too to make sure we don't make regressions in the future?

packages/integrations/node/src/middleware.ts
wulinsheng123: > I have one nit but this looks good to me! Could we add some tests too to make sure we don't make regressions in the future?



> It would be great if we pass `mode` instead of `options` since we only use the `mode`.

ok no problem

wulinsheng123:
> I have one nit but this looks good to me! Could we add some tests too to make sure we don't make regressions in the future?

Let me research how to add a test for this code

wulinsheng123:
@bluwy I added some tests

bluwy (commented):
The new test looks great! Thanks for adding it. I have a few comments below but once done it should be all set.

wulinsheng123:
> The new test looks great! Thanks for adding it. I have a few comments below but once done, it should be all set.

sorry about that, 

> The new test looks great! Thanks for adding it. I have a few comments below but once done it should be all set.

My bad. I haven't rechecked. 
I will fix it as soon as possible

sarah11918:
Docs was pinged on this one, but I don't see anything needed from us here! I'm removing the review request, but if I missed something I should be looking at, just let me know!

bluwy (approved):
Awesome, thanks for making the fix.

PR #5706, opened 2022-12-30
Fix docs smoke error

bluwy:
## Changes

Add `@astrojs/sitemap` to `ssr.external` when loading config. It should be simplified when https://github.com/vitejs/vite/pull/10939 lands.

## Testing

Astro smoke test should pass.

## Docs

n/a bug fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "format:imports": "organize-imports-cli ./packages/*/tsconfig.json ./packages/*/*/tsconfig.json",
     "test": "turbo run test --output-logs=new-only --concurrency=1 --filter=astro --filter=create-astro --filter=\"@astrojs/*\"",
     "test:match": "cd packages/astro && pnpm run test:match",
-    "test:smoke": "turbo run build --filter=\"@example/*\" --filter=\"astro.build\" --filter=\"docs\" --output-logs=new-only --concurrency=1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "format:imports": "organize-imports-cli ./packages/*/tsconfig.json ./packages/*/*/tsconfig.json",
     "test": "turbo run test --output-logs=new-only --concurrency=1 --filter=astro --filter=create-astro --filter=\"@astrojs/*\"",
     "test:match": "cd packages/astro && pnpm run test:match",
-    "test:smoke": "turbo run build --filter=\"@example/*\" --filter=\"astro.build\" --filter=\"docs\" --output-logs=new-only --concurrency=1",
+    "test:smoke": "turbo run build --filter=\"@example/*\" --filter=\"docs\" --output-logs=new-only --concurrency=1",
```
bluwy, line 21: There doesn't look like a `astro.build` package to test so removed here

bholmesdev (approved):
Smost excellent fix of 2023 👏

package.json
bholmesdev: Hm, there _was_ an `astro.build` package but the checkout code was removed. We should add that back in a separate PR

aFuzzyBear:
> Smost excellent fix of 2023 👏

... so far 😉

PR #5717, opened 2023-01-02
Refactor tailwind integration setup

bluwy:
## Changes

As discussed in the [maintainers chat](https://discord.com/channels/830184174198718474/857704189723541514/1057581910051794994), the `style.postcss` Astro config can be removed in favour of `vite.css.postcss`. They do the same thing internally, and is only a property used by the tailwind integration.

I also disabled autoprefixer in dev. I'm not sure if this was intentional, but enabling it slows down dev a lot so I changed that in the PR too.

## Testing

It doesn't seem like there's existing tests for tailwind, but I've tested manually with the `with-tailwindcss` example and it works.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/integrations/svelte/src/index.ts
+++ b/packages/integrations/svelte/src/index.ts
@@ -1,6 +1,6 @@
 import type { Options } from '@sveltejs/vite-plugin-svelte';
 import { svelte } from '@sveltejs/vite-plugin-svelte';
-import type { AstroConfig, AstroIntegration, AstroRenderer } from 'astro';
+import type { AstroIntegration, AstroRenderer } from 'astro';
 import preprocess from 'svelte-preprocess';
 import type { UserConfig } from 'vite';
 
@@ -15,21 +15,16 @@ function getRenderer(): AstroRenderer {
 type ViteConfigurationArgs = {
 	isDev: boolean;
 	options?: Options | OptionsCallback;
-	postcssConfig: AstroConfig['style']['postcss'];
 };
 
-function getViteConfiguration({
-	options,
-	postcssConfig,
-	isDev,
-}: ViteConfigurationArgs): UserConfig {
+function getViteConfiguration({ options, isDev }: ViteConfigurationArgs): UserConfig {
 	const defaultOptions: Partial<Options> = {
 		emitCss: true,
 		compilerOptions: { dev: isDev, hydratable: true },
 		preprocess: [
 			preprocess({
 				less: true,
-				postcss: postcssConfig,
+				postcss: true,
 				sass: { renderSync: true },
 				scss: { renderSync: true },
 				stylus: true,
@@ -78,13 +73,12 @@ export default function (options?: Options | OptionsCallback): AstroIntegration 
 		name: '@astrojs/svelte',
 		hooks: {
 			// Anything that gets returned here is merged into Astro Config
-			'astro:config:setup': ({ command, updateConfig, addRenderer, config }) => {
+			'astro:config:setup': ({ command, updateConfig, addRenderer }) => {
 				addRenderer(getRenderer());
 				updateConfig({
 					vite: getViteConfiguration({
 						options,
 						isDev: command === 'dev',
-						postcssConfig: config.style.postcss,
 					}),
 				});
 			},
--- a/packages/integrations/tailwind/src/index.ts
+++ b/packages/integrations/tailwind/src/index.ts
@@ -66,6 +66,20 @@ async function getUserConfig(root: URL, configPath?: string, isRestart = false) 
 	}
 }
 
+function getViteConfiguration(isBuild: boolean, tailwindConfig: TailwindConfig) {
+	const postcssPlugins = [tailwindPlugin(tailwindConfig)];
+	if (isBuild) {
+		postcssPlugins.push(autoprefixerPlugin());
+	}
+	return {
+		css: {
+			postcss: {
+				plugins: postcssPlugins,
+			},
+		},
+	};
+}
+
 type TailwindOptions =
 	| {
 			config?: {
@@ -92,7 +106,14 @@ export default function tailwindIntegration(options?: TailwindOptions): AstroInt
 	return {
 		name: '@astrojs/tailwind',
 		hooks: {
-			'astro:config:setup': async ({ config, injectScript, addWatchFile, isRestart }) => {
+			'astro:config:setup': async ({
+				command,
+				config,
+				updateConfig,
+				injectScript,
+				addWatchFile,
+				isRestart,
+			}) => {
 				// Inject the Tailwind postcss plugin
 				const userConfig = await getUserConfig(config.root, customConfigPath, isRestart);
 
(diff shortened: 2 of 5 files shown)


packages/integrations/svelte/src/index.ts
```diff
@@ -78,13 +73,12 @@ export default function (options?: Options | OptionsCallback): AstroIntegration
...
 		hooks: {
 			// Anything that gets returned here is merged into Astro Config
-			'astro:config:setup': ({ command, updateConfig, addRenderer, config }) => {
+			'astro:config:setup': ({ command, updateConfig, addRenderer }) => {
 				addRenderer(getRenderer());
 				updateConfig({
 					vite: getViteConfiguration({
 						options,
 						isDev: command === 'dev',
-						postcssConfig: config.style.postcss,
```
bluwy, line 87: I didn't add a changeset for the svelte integration as it should be overidden by https://github.com/withastro/astro/pull/5685 specifically [this change](https://github.com/withastro/astro/pull/5685/files#diff-03dddc60579af21eb48fb3156c2466b43ac2e80bd6cdb5a399b271d77336cbc1)

bluwy:
Marking as draft. This depends on https://github.com/withastro/astro/pull/5685

Princesseuh (approved):
Looks good to me! I fear a little bit that some users will get confused by autoprefixer not being on in dev, but if we communicate it enough it'll be fine, I think

bluwy:
Good point. I've updated the tailwind readme to reflect that. It also doesn't look like we mention autoprefixer anywhere before, so this seems like a good time to bring up. cc @withastro/maintainers-docs for review!

packages/integrations/tailwind/README.md
```diff
@@ -68,6 +68,8 @@ export default defineConfig({
 
 When you install the integration, Tailwind's utility classes should be ready to go right away. Head to the [Tailwind docs](https://tailwindcss.com/docs/utility-first) to learn how to use Tailwind, and if you see a utility class you want to try, add it to any HTML element to your project!
 
+[Autoprefixer](https://github.com/postcss/autoprefixer) is also setup automatically for production builds so Tailwind classes work in older browsers.
```
sarah11918, line 71:
```suggestion
[Autoprefixer](https://github.com/postcss/autoprefixer) is also setup automatically for production builds so Tailwind classes will work in older browsers.
```
Tiniest of nits. And thanks for adding this, @bluwy! Great thinking @Princesseuh!

sarah11918 (approved):

natemoo-re (approved):

packages/integrations/tailwind/src/index.ts
```diff
@@ -66,6 +66,20 @@ async function getUserConfig(root: URL, configPath?: string, isRestart = false)
...
 
+function getViteConfiguration(isBuild: boolean, tailwindConfig: TailwindConfig) {
+	const postcssPlugins = [tailwindPlugin(tailwindConfig)];
+	if (isBuild) {
+		postcssPlugins.push(autoprefixerPlugin());
+	}
+	return {
+		css: {
+			postcss: {
+				plugins: postcssPlugins,
```
artstorm, line 77: @bluwy Setting an inline PostCSS config in Vite, has the effect that Vite no longer looks for `postcss.config.cjs`. 

See: 
> Note if an inline config is provided, Vite will not search for other PostCSS config sources.
> https://vitejs.dev/config/shared-options.html#css-postcss 

That results in by adding the Tailwind Integration 3.0.0-beta.x, `postcss.config.cjs` no longer work as described in the Astro Docs.  
https://docs.astro.build/en/guides/styling/#postcss

bluwy: Hmm that's true, but I think that happens before this PR too as Astro was assigning to the property internally:

https://github.com/withastro/astro/pull/5717/files#diff-025f8338e5be413b0f65c74bc50d6484c6bc9bb4f78847da0ba896b8e9569390L153-L155

I guess for the tailwind integration we need to read `postcss.config.js` too to resolve this.

LexSwed: Confirming, trying latest `@beta` packages makes Tailwind to ignore `postcss.config.js` (f.e. to enable nesting)

MoustaphaDev: > Seem to be addressed in: #5908

Came up with a possible fix for the issue but there's an issue. Applying changes to the `postcss.config.js` file only has effect after restarting the dev server.
I just wanted to know if it was an issue before, so that it will get addressed in another PR and doesn't block the current one.

bluwy: I think it was looking at the old code. I'd be fine merging it first, then handle the restart later.

MoustaphaDev: > I think it was looking at the old code. I'd be fine merging it first, then handle the restart later.

Sounds good, thanks!

artstorm:
@bluwy 

> I also disabled autoprefixer in dev. I'm not sure if this was intentional, but enabling it slows down dev a lot so I changed that in the PR too.

I found another side effect from this PR. There are classes in Tailwind that relies on Autoprefixer to work in all current browsers. One being [Background Clip](https://tailwindcss.com/docs/background-clip). `background-clip` only works in Chrome with `-webkit-` prefix. See [CanIUse](https://caniuse.com/?search=background-clip).

Tailwind styling that works perfectly fine in Astro 1.9 in dev, doesn't work in the current beta if using the lastest Chrome.

Example:
image

It would probably be a good idea to keep autoprefixer enabled in dev, as some Tailwind classes rely on it to work across all browsers. Otherwise it will probably be quite confusing when working in dev f certain parts of Tailwind does not work in a popular browser like Chrome.

bluwy:
I think I saw this issue in Discord too. We can definitely add an option to turn on autoprefixer in dev, but I still think it's good to turn it off by default to have good performance. Usually it's expected that developers work in evergreen browsers, but `background-clip` seems to be a special case that's not supported by most evergreen browsers.

Another option is to turn on `background-clip` processing only in dev. Could you help create an issue so we can track this? 🙏

chenxsan:
> Usually it's expected that developers work in evergreen browsers

That's not true if your project targets any old browsers, you would need to test them when developing, not after building.

I think Astro can provide an option for users who want to disable it in dev env, but please don't turn it on by default as it could break things like the issue https://github.com/withastro/astro/issues/5989

williameckert1:
> We can definitely add an option to turn on autoprefixer in dev, but I still think it's good to turn it off by default to have good performance.
Please do, but ideally with Autoprefixer on by default.
The issue with _background-clip_ was confusing me a lot today (my fault for not reading the changelog in detail). But this new default makes it impossible to preview the UI correctly across browsers.

badalsaibo:
Hey, I upgraded from v1.0.0 to v2.6.0 and there was an error regarding this `config.style.postcss` not found? Any work arounds for that? Thanks

bluwy:
Make sure to upgrade all the astro-related dependencies together.

badalsaibo:
So the astro update itself happens when adding the official netflify adapter. Is there any command to upgrade all astro required dependencies?

bluwy:
`npm update @astrojs/tailwind @astrojs/other-stuff`. Please open a support thread in Discord for further support 😬

PR #5720, opened 2023-01-02
fix(astro/core): Do not add base to hoisted script body

umarov:
## Changes

Remove inclusion of the base path from config into a hoisted script body. 

Before the hoisted script would be like this if the config has this base value: `base: 'astro-base-bug'`

```js
astro-base-bug/console.log("hello");
```

This would break when the page loads.

After my changes, things are as expected:

```js
console.log("hello");
```

## Testing

Added a test to confirm the fix. While writing the tests, I had the previous version in `dist` and it fail and include the base path value in the script as well.

## Docs

No changes to docs necessary

### DIFF
--- a/.changeset/good-carpets-confess.md
+++ b/.changeset/good-carpets-confess.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Do not add base path to a hoisted script body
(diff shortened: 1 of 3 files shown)


matthewp:
Thanks!

PR #3286, opened 2022-05-04
feat: expose pages

thepassle:
## Changes

- What does this change?
Exposes a `pages` property in the `astro:build:setup` hook as discussed in: https://github.com/withastro/rfcs/discussions/188#discussioncomment-2682523 The comment suggests to add a new hook, but it seemed like it makes more sense to just add it to the setup hook instead rather than giving it it's entire own special hook.

## Testing

Tested locally

## Docs

The `astro:build:setup` hook wasnt yet documented at all, so I added it here as well: https://github.com/withastro/docs/pull/433

PR #2879, opened 2022-03-24
Netlify adapter

matthewp:
## Changes

- Adds the Netlify adapter that works with Netlify Functions.
- Allows integrations to configure the folders where client/server code is written to.

## Testing

Not testing adapters, yet.

## Docs

- This will be part of https://github.com/withastro/docs/pull/255
- The readme has docs

PR #2820, opened 2022-03-17
Astro Integration System

FredKSchott:
## Changes

- Implements this RFC: https://github.com/withastro/rfcs/blob/main/proposals/0015-integrations.md

## How to Review

This is all interconnected to the point that I couldn't easily break the PR into smaller, individual, functional PRs. But, I was able to break down this PR into 3 different commits for a bit easier review:

1. [Review updated usage in examples](https://github.com/withastro/astro/pull/2820/commits/b3158ca9d853d53087821300dbbe4dd4e37d4483)
2. [Review initial integrations implementations](https://github.com/withastro/astro/pull/2820/commits/7aa0247f1ac210b38309c2b83a13a4eea394cfd2) 
3. [Update tests](https://github.com/withastro/astro/pull/2820/commits/978572a405a27308fdc9fbf59f071eee59085603)
4. [Review integration support in Astro](https://github.com/withastro/astro/pull/2820/commits/924634616d9e7b40e53fe151fff6e637251e7e34)

## Testing

- Updated.

## Docs

- https://github.com/withastro/docs/pull/239

PR #5727, opened 2023-01-03
Remove `pages` param from `astro:build:done`

bholmesdev:
## Changes

This PR removes the `pages` parameter from the `astro:build:done` hook and associated core logic. `pages` is an undocumented parameter that only the `@astrojs/sitemap` integration depends on. Its behavior has been replaced by `routes` across our other integrations and adapters, and requires a `pageNames` calculation that's unused by core.

## Testing

Ensure existing sitemap integration tests still pass

## Docs

- Remove `pages` from API quick reference (didn't match `astro:build:done` full length docs!) https://github.com/withastro/docs/pull/2277

### DIFF
--- a/.changeset/green-mangos-fix.md
+++ b/.changeset/green-mangos-fix.md
@@ -1,0 +1,12 @@
+---
+'astro': major
+'@astrojs/sitemap': patch
+---
+
+Remove the `pages` parameter from the `astro:build:done` integration hook.
+
+#### Migration
+
+The `pages` parameter was an undocumented parameter used by the `@astrojs/sitemap` integration. If your integration relied on `pages`, use the `routes` parameter and associated `pathname` property instead.
+
+Note: The `pathname` property on `routes` does _not_ respect [the `trailingSlash` property](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) in your Astro config. If needed, ensure you append this slash after reading the user's Astro config.
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1440,11 +1440,7 @@ export interface AstroIntegration {
 			updateConfig: (newConfig: vite.InlineConfig) => void;
 		}) => void | Promise<void>;
 		'astro:build:generated'?: (options: { dir: URL }) => void | Promise<void>;
-		'astro:build:done'?: (options: {
-			pages: { pathname: string }[];
-			dir: URL;
-			routes: RouteData[];
-		}) => void | Promise<void>;
+		'astro:build:done'?: (options: { dir: URL; routes: RouteData[] }) => void | Promise<void>;
 	};
 }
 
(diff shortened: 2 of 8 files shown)


matthewp:
It was also added to setup here: https://github.com/withastro/astro/pull/3286

matthewp (changes requested):
Blocking for discussion with core.

bholmesdev:
@matthewp To clarify, `pages` is a completely different shape in the `setup` hook! `pages` was just a set of `pathname` objects in the `astro:build:done` hook, duplicated by `routes`.

matthewp:
Doing some research on the history here.

- `pages` was added when the integrations API first released: https://github.com/withastro/astro/pull/2820/files
- `routes` was added when the Netlify adapter was added, because it needs the routes to generate its manifest: https://github.com/withastro/astro/pull/2879/files

matthewp:
So it looks like `pages` is the same as `routes.map(route => route.component)`, does that sound right to you @bholmesdev?

.changeset/green-mangos-fix.md
```diff
@@ -0,0 +1,12 @@
...
+'@astrojs/sitemap': patch
+---
+
+Remove the `pages` parameter from the `astro:build:done` integration hook.
+
+#### Migration
+
+The `pages` parameter was an undocumented parameter used by the `@astrojs/sitemap` integration. If your integration relied on `pages`, use the `routes` parameter and associated `pathname` property instead.
+
+Note: The `pathname` property on `routes` does _not_ respect [the `trailingSlash` property](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) in your Astro config. If needed, ensure you append this slash after reading the user's Astro config.
```
matthewp, line 12: Given this is such a small thing is it really worth removing? I understand that there is some overlap but this seems like a difference that some integrations might be relying on.

bholmesdev:
@matthewp Hm, it looks like the `trailingSlash` handling was specific to `pageName` and not respected by `route.component` either. It still feels odd to me that we have the `addPageName` behavior in core when it's only used by the sitemap integration.

.changeset/green-mangos-fix.md
bholmesdev: I understand, and maybe we can revert! Though I think having both `pages` and `routes` side-by-side is pretty confusing since they both mean the same thing conceptually. We avoided documenting `pages` in favor of `routes` for this reason. If we keep it around, I'd prefer a clearer name so we can document when it should be used _instead_ of routes.

matthewp: I'd feel better about this if we knew that no one was using it. It's too bad that we don't have telemetry data on this yet (something to add).

I do agree that `pages` is kind of weird, especially that it only applies to SSG. We need to discuss it might though I think.

bholmesdev:
Closing in favor of a more holistic fix for Astro 3.0. Notes to consider:
- `pages` respects the `trailingSlash` option while `routes` does not. How can we reconcile the behavior?
- `RouteData` exposes internal properties like `segments`, which most library authors don't need. How can we reshape `RouteData` into an intuitive object for integration authors?

PR #5728, opened 2023-01-03
Unflag experimental features

natemoo-re:
## Changes

Unflags `--experimental-error-overlay` and `--experimental-prerender`

## Testing

CI

## Docs

N/A

### DIFF
--- a/.changeset/chilled-geese-worry.md
+++ b/.changeset/chilled-geese-worry.md
@@ -9,4 +9,4 @@ You'll notice that the error overlay during `astro dev` has a refresh visual des
 The `prerender` feature is now enabled by default when using `output: 'server'`. To prerender a particular page, add `export const prerender = true` to your frontmatter.
 
 > **Warning**
-> Integration authors that previously relied on the exact structure of Astro's v1.0 build output may notice some changes to our output file structure. Please should test your integrations to ensure compatability.
+> Integration authors that previously relied on the exact structure of Astro's v1.0 build output may notice some changes to our output file structure. Please test your integrations to ensure compatability.
--- a/.changeset/chilled-geese-worry.md
+++ b/.changeset/chilled-geese-worry.md
@@ -1,0 +1,12 @@
+---
+'astro': major
+---
+
+The previously experimental features `--experimental-error-overlay` and `--experimental-prerender`, both added in v1.7.0, are now the default.
+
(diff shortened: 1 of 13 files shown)


matthewp:
The test failing here seems really bizarre...

natemoo-re:
> The test failing here seems really bizarre...

This seems to have something to do with the `prerender` changes—I'll investigate.

.changeset/chilled-geese-worry.md
```diff
@@ -0,0 +1,12 @@
+---
+'astro': major
+---
+
+The previously experimental features `--experimental-error-overlay` and `--experimental-prerender`, both added in v1.7.0, are now the default.
+
+You'll notice that the error overlay during `astro dev` has a refresh visual design and provides more context for your errors.
```
delucis, line 7: Typo:

```suggestion
You'll notice that the error overlay during `astro dev` has a refreshed visual design and provides more context for your errors.
```

packages/astro/src/@types/astro.ts
```diff
@@ -927,46 +925,6 @@ export interface AstroUserConfig {
 	 * These flags are not guaranteed to be stable.
 	 */
 	experimental?: {
-		/**
```
delucis, line 930: One thing to note is that removing these will remove them from docs too — https://docs.astro.build/en/reference/configuration-reference/#experimental-flags

Maybe that doesn’t really matter, but we should make sure to say v2-beta required when documenting `prerender` (error overlay’s probably less important docs-wise). I don’t think we got a `prerender` docs PR yet, right @natemoo-re?

natemoo-re: Correct, will let you know when that's ready!

natemoo-re: https://github.com/withastro/docs/pull/2279

natemoo-re:
> The test failing here seems really bizarre...

Figured this out. It was a test that we could customize `vite.build.rollupOptions.output.chunkFileNames`, but the `package.json` didn't include `"type": "module"` and the output files were plain `.js` (which would be treated as CJS in this case). 

Since prerender splits each page out to a separate chunk _instead of a single file_, this broke. Made a specific note in the changelog for this situation.

matthewp (approved):

delucis (approved):

PR #5736, opened 2023-01-03
update frontmatter injection link to match docs changes

sarah11918:
## Changes

Docs changed some headings, so this link needs to be updated:

frontmatter injection -> modifying frontmatter

I don't know whether we want to change the error message name or anything else in the description now that we decided to say "modifying frontmatter" vs "frontmatter injection", but at least this corrects the outdated link.

cc @bholmesdev @Princesseuh 

## Testing

No tests.

## Docs

Will reflect in docs.

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -523,7 +523,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	/**
 	 * @docs
 	 * @see
-	 * - [Frontmatter injection](https://docs.astro.build/en/guides/markdown-content/#example-injecting-frontmatter)
+	 * - [Frontmatter injection](https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically)
 	 * @description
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! Thanks @sarah11918

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -523,7 +523,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	/**
 	 * @docs
 	 * @see
-	 * - [Frontmatter injection](https://docs.astro.build/en/guides/markdown-content/#example-injecting-frontmatter)
+	 * - [Frontmatter injection](https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically)
```
Princesseuh, line 526:
```suggestion
	 * - [Modifying frontmatter programmatically](https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically)
```

In other errors I made it so the link name is the header title, thoughts on doing the same here?

delucis: Yes, that does make sense 👍

PR #5756, opened 2023-01-04
Fix for hoisted scripts in project with spaces in the file path

matthewp:
Note that this will be backported to 1.0 after being cherry-picked.

## Changes

- Fixes https://github.com/withastro/astro/issues/5598

## Testing

- Test added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/vite-plugin-utils/index.ts
+++ b/packages/astro/src/vite-plugin-utils/index.ts
@@ -3,10 +3,9 @@ import type { AstroConfig } from '../@types/astro';
 import {
 	appendExtension,
 	appendForwardSlash,
-	prependForwardSlash,
 	removeLeadingForwardSlashWindows,
 } from '../core/path.js';
-import { fileURLToPath } from 'url';
+import { viteID } from '../core/util.js';
 
 /**
  * Converts the first dot in `import.meta.env` to its Unicode escape sequence,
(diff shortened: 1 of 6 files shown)


packages/astro/src/vite-plugin-utils/index.ts
```diff
@@ -47,7 +49,8 @@ export function normalizeFilename(filename: string, config: AstroConfig) {
 	if (filename.startsWith('/@fs')) {
 		filename = filename.slice('/@fs'.length);
 	} else if (filename.startsWith('/') && !ancestor(filename, config.root.pathname)) {
-		filename = new URL('.' + filename, config.root).pathname;
+		const url = new URL('.' + filename, config.root);
+		filename = prependForwardSlash(fileURLToPath(url));
```
natemoo-re, line 53: Will this work on Windows? I assume `pathname` was being used here to ensure that we were getting the normalized `/path/to/file` but `fileURLToPath(url)` will give a platform-dependent file path. Maybe the result of that needs to be normalized before the forward slash is prepended?

natemoo-re: If Windows tests pass, then I digress.

matthewp: Yeah there's something wrong here. We should be using fileURLToPath though, otherwise you get `%20` for the spaces. Something else is wrong.

natemoo-re: I think `prependForwardSlash(normalizePath(fileURLToPath(url)))` would be correct? Using [`normalizePath`](https://vitejs.dev/guide/api-javascript.html#normalizepath) from Vite.

matthewp: viteID did it. the problem was that fileURLToPath uses back slashes. viteID fixes it.

natemoo-re (approved):
Looks great! Insert [space] pun here.

PR #5757, opened 2023-01-04
[ci] Skip OG images in docs smoke test

natemoo-re:
## Changes

- Adds `SKIP_OG` env flag for `docs` smoke test.
- This should speed up our smoke tests by avoiding Open Graph image generation in the docs!

## Testing

N/A, tooling change only

## Docs

N/A, tooling change only


natemoo-re:
cc @delucis for a review, I think this should work!

matthewp (approved):

PR #5758, opened 2023-01-04
Setup changesets to backport to 1.0 legacy branch

matthewp:
This is so that we can run a patch release for 1.0 without doing so manually. Once the legacy branch has been created we will revert this change.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -4,7 +4,7 @@
   "commit": false,
   "linked": [],
   "access": "public",
-  "baseBranch": "main",
+  "baseBranch": "1-legacy",
   "updateInternalDependencies": "patch",
   "ignore": ["@example/*", "@test/*"],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -4,7 +4,7 @@
   "commit": false,
   "linked": [],
   "access": "public",
-  "baseBranch": "1-legacy",
+  "baseBranch": "main",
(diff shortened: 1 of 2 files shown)


.changeset/config.json
```diff
@@ -4,7 +4,7 @@
   "commit": false,
   "linked": [],
   "access": "public",
-  "baseBranch": "main",
+  "baseBranch": "1-legacy",
```
natemoo-re, line 7: I think we can change this only on the `1-legacy` branch rather than main?

.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     branches:
-      - main
+      - 1-legacy
```
natemoo-re, line 6: I think we can include both of these? This action would then run on pushes to `main` or `1-legacy`, which would keep us from updating this workflow all the time.

matthewp:
Oh interesting idea. So basically make the change for the workflow to `main` and the config.json to the legacy branch, so changes to either one would result in the bot running and reading the respective config file. Worth trying.

natemoo-re:
> Oh interesting idea. So basically make the change for the workflow to `main` and the config.json to the legacy branch, so changes to either one would result in the bot running and reading the respective config file. Worth trying.

Yep! 80% sure it will work like we intend, but worth trying either way!

natemoo-re (approved):

matthewp:
This is wrong, should point to both branches, fixing...

matthewp:
Ok, should be right now.

PR #5761, opened 2023-01-04
Log pretty error when MDX integration is missing

bholmesdev:
## Changes

image

Addresses https://github.com/withastro/rfcs/discussions/420

Add "MDX integration missing" error when the MDX renderer is missing. This addresses a common source for bugs, where the user finds an obscure syntax error when the MDX integration isn't configured.

## Testing

Manual testing

## Docs

Add new error-data entry

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -534,6 +534,23 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 			'A remark or rehype plugin attempted to inject invalid frontmatter. Ensure "astro.frontmatter" is set to a valid JSON object that is not `null` or `undefined`.',
 		hint: 'See the frontmatter injection docs https://docs.astro.build/en/guides/markdown-content/#modifying-frontmatter-programmatically for more information.',
 	},
+	/**
+	 * @docs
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -534,6 +534,23 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * @description
+	 * Unable to find the official `@astrojs/mdx` integration. This error is raised when using MDX files without an MDX integration installed.
+	 */
+	MdxIntegrationMissingError: {
+		title: 'MDX integration missing.',
+		code: 6004,
+		message: (id: string) => {
+			return `Unable to render ${JSON.stringify(
+				id
+			)}. Ensure that the \`@astrojs/mdx\` integration is applied.`;
```
sarah11918, line 550:
```suggestion
			)}. Ensure that the \`@astrojs/mdx\` integration is installed.`;
```
I think a user might be more familiar with "installed" than "applied," since installing is what the user does, and installing is the fix for this?

Otherwise, LGTM!

sarah11918 (approved):

natemoo-re (approved):

PR #5763, opened 2023-01-05
fix: auto update field name in `package.json`

ansidev:
## Changes

Update field `name` in `package.json` while generating new project to match the user input.

- [x] NPM v6 and older: ignore updating the NPM package name.
- [x] NPM v7 and newer: supported.

The value for the new NPM package name is taken from:
1. `--package` argument.
2. `projectDir` if `--package` is not given.
3. request user input if the `--package` or `projectDir` is an invalid NPM package name.
4. ignore updating the NPM package name if user cancels the operation.

## Test cases

- [x] should the package name is changed if no value is given by the user.
- [x] should the package name is not changed if the user cancels the operation.
- [x] should the package name matches the passed argument `--package`.
- [x] should the package name matches the user input.

### DIFF
--- a/packages/create-astro/src/index.ts
+++ b/packages/create-astro/src/index.ts
@@ -10,6 +10,7 @@ import ora from 'ora';
 import { platform } from 'os';
 import path from 'path';
 import prompts from 'prompts';
+import validate from 'validate-npm-package-name';
 import detectPackageManager from 'which-pm-runs';
 import yargs from 'yargs-parser';
 import { loadWithRocketGradient, rocketAscii } from './gradient.js';
@@ -89,6 +90,19 @@ function isValidProjectDirectory(dirPath: string) {
 	return conflicts.length === 0;
 }
 
+async function validatePackageName(pkgName: string) {
+	const isValidPackageName = await validate(pkgName);
+	const isValid = isValidPackageName.validForNewPackages
+	const notices: string[] = []
+	if (!isValid) {
+		notices.push(...[
+			...isValidPackageName.warnings || [],
+			...isValidPackageName.errors || []
+		]);
+	}
+	return { isValid, notices }
+}
+
 const FILES_TO_REMOVE = ['.stackblitzrc', 'sandbox.config.json', 'CHANGELOG.md']; // some files are only needed for online editors when using astro.new. Remove for create-astro installs.
 
 // Please also update the installation instructions in the docs at https://github.com/withastro/docs/blob/main/src/pages/en/install/auto.md if you make any changes to the flow or wording here.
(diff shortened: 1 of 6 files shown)


packages/create-astro/src/index.ts
```diff
@@ -208,6 +209,9 @@ export async function main() {
 				}
 			})
 		);
+		
+		// Change project name in package.json
+		await execaCommand(`${pkgManager} pkg set name="${projectDir}"`, { cwd });
```
natemoo-re, line 214: Oh wow, I didn't know this was a thing! Clever.

natemoo-re (approved):
Looks good to me!

packages/create-astro/src/index.ts
```diff
@@ -208,10 +229,55 @@ export async function main() {
...
+					{ onCancel: printIgnoreUpdatingPkgNameMsg }
+				);
+				pkgName = pkgNameResponse.package as string;
+			}
+
+			if (typeof pkgName === 'string' && !!pkgName) {
+				// Change project name in package.json
+				await execaCommand(`${pkgManager} -v`, { cwd });
+				await execaCommand(`npm -v`, { cwd });
+				await execaCommand(`npm pkg set name=${pkgName}`, { cwd });
```
natemoo-re, line 276: Do we really need to run all three of these? I'd expect to use `pkgManager` and fail silently if `pkg set name` is not supported.

packages/create-astro/test/utils.js
```diff
@@ -37,16 +37,24 @@ export function promiseWithTimeout(testFn) {
...
 		stdin,
 		stdout,
 	};
 }
+
+export async function getNPMVersion() {
+	const { stdout: npmVersion } = await execa('npm', ['-v']);
+	const [major, minor, patch] = npmVersion.split('.').map(Number);
+	return { major, minor, patch }
+}
```
natemoo-re, line 60: This is unused AFAICT

natemoo-re (changes requested):

packages/create-astro/src/index.ts
ansidev: @natemoo-re Let me check this!

packages/create-astro/test/utils.js
ansidev: @natemoo-re `getNPMVersion()` is used by `package-step.test.js`.

I need this function because the sub command `npm pkg` is not supported in NPM v6 and older. Currently, I don't find out anyway to implement test case that run on specific Node version except using `if-else` condition.

I will move it into `package-step.test.js` since it will be used in that test file at this moment.

ansidev:
@natemoo-re

I have tested the `main` branch using `GitHub Codespaces`.

Running `pnpm run build`: failed randomly.

Error message: `command (packages/${random-package-name}) pnpm run build exited (1)`.

I don't know why. However, all test cases will pass if running `pnpm run build && pnpm run test` in the folder `create-astro`.

natemoo-re:
Hey @ansidev thanks for this PR! I did some refactoring of `create-astro` for testability and made sure to include this feature! See https://github.com/withastro/astro/pull/6082

PR #5765, opened 2023-01-05
Implement redesign of portfolio example

delucis:
## Changes

- Replace our current portfolio example with a new design by @doodlemarks

This is a more fully featured example than some of our more basic starters, so closer to a real “theme”. I think that was the general idea with this, so hopefully that’s OK. The starter does not include any dependencies apart from `astro` but does add some custom components including an `<Icon>` that contains the icons Mark suggested from the [Phosphor](https://phosphoricons.com/) icon set.

### Questions

- The original design used [Public Sans](https://fonts.google.com/specimen/Public+Sans) for body text. I’ve stuck with this for now, but **is there an argument for dropping Public in favour of a system font?** Public is a fairly similar sans to the default sans-serif bundled on most devices and we’re already loading Rubik as a custom font for headings.

- **Should this be our first full example that uses Content Collections?** This example already includes a type for frontmatter _and_ suffers from the style bleed issue when globbing projects for the homepage/work index, so could make sense to set those up a schema?

- There is some complexity getting all the background gradient images to work nicely, including multiple resolutions per background and some custom lazy loading for below the fold backgrounds. **Is this too much complexity for a theme like this?** Maybe there’s some alternative design approach that is lighter and we use background images just for the page headers.

### To-do

- [x] Some `astro check` errors in CI — will clean those up

### Light mode 
Screen Shot 2023-01-05 at 13 20 26

### Dark mode 
Screen Shot 2023-01-05 at 13 20 17

## Testing

- Lots of manual testing during development across browsers & devices — more eyes always appreciated. 
- All pages pass accessibility audits.

## Docs

We don’t generally document much for these, so guess we stick to that pattern? The screenshot on astro.new will need updating.


examples/portfolio/src/layouts/BaseLayout.astro
```diff
@@ -0,0 +1,101 @@
...
+		<div class="stack backgrounds">
+			<Nav />
+			<slot />
+			<Footer />
+		</div>
+
+		<script>
+			// Add “loaded” class once the document has completely loaded.
+			addEventListener('load', () => document.documentElement.classList.add('loaded'));
+		</script>
```
delucis, line 32: Here’s the lazy loading technique I mentioned. Background images are only set once the page has the `.loaded` class. This means we can prioritize above the fold background images and only load others after the page load is complete.

examples/portfolio/src/types.ts
```diff
@@ -1,8 +1,8 @@
 export interface Project {
 	title: string;
-	client: string;
 	description: string;
 	publishDate: string;
 	tags: string[];
 	img: string;
+	img_alt?: string;
 }
```
delucis, line 8: We’re already using this frontmatter interface, so maybe content collections would make sense.

examples/portfolio/src/components/Nav.astro
```diff
@@ -1,99 +1,324 @@
...
-	.socialIcon:hover {
-		fill: var(--t-active);
+	@media (min-width: 60em) {
+		.socials {
+			display: flex;
+			justify-content: flex-end;
+			gap: 0;
+		}
 	}
 </style>
```
Yan-Thomas, line 216: More high-contrast specific styles to fix the current page styles.
| Before | After |
| image | image |

Note that I choose to change the text color instead of the `background-color` because it was causing a weird background over the text and also it wasn't looking good on the hamburger menu:

image

image

```suggestion
	@media (forced-colors: active) {
		.link.active {
			color: SelectedItem;
		}
	}
</style>
```

Yan-Thomas (commented):
Great job @delucis, you and @doodlemarks made the portfolio example even better! I added a couple a11y (high contrast related) and UI suggestions for you to consider. 🙌

doodlemarks:
Screen Shot 2023-01-05 at 9 18 44 AM

I would add a bit more space here between the subheader and image. [Check out the mocks](https://www.figma.com/file/qiACDNgYrQrwEMj4VkX8Cx/Theme---Portfolio?node-id=36%3A903&t=AKFlY3xNisAcgfua-1). 

And one super nitpicky thing, can we expand the example number of works from 3 to 4?

examples/portfolio/src/layouts/BaseLayout.astro
```diff
@@ -0,0 +1,101 @@
...
+
+			.backgrounds {
+				isolation: isolate;
+				background: url('/assets/backgrounds/noise.svg') top center/1500px repeat,
+					var(--bg-image-main-curves) top center/var(--bg-gradient-size) no-repeat,
+					var(--bg-image-main) top center/var(--bg-gradient-size) no-repeat,
+					var(--bg-image-footer) bottom center/var(--bg-gradient-size) no-repeat, var(--gray-999);
+				background-blend-mode: overlay, var(--bg-svg-blend-mode), normal, normal, normal;
+			}
+		</style>
```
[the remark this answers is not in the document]
delucis:
```suggestion
			@media (forced-colors: active) {
				/* Deactivate custom backgrounds for high contrast users. */
				.backgrounds {
					background: none;
					background-blend-mode: none;
					--bg-gradient-size: none;
				}
			}
		</style>
```

examples/portfolio/src/components/Icon.astro
```diff
@@ -0,0 +1,76 @@
...
+	icon: keyof typeof iconPaths;
+	color?: string;
+	gradient?: boolean;
+	size?: string;
+}
+
+const { color = 'currentcolor', gradient, icon, size } = Astro.props;
+const iconPath = iconPaths[icon];
+
+const attrs: astroHTML.JSX.HTMLAttributes = {};
```
natemoo-re, line 36: Could also use our utility type here.
```
import type { HTMLAttributes } from 'astro/types';

const attrs: HTMLAttributes<'svg'> = {};
```

examples/portfolio/src/components/Skills.astro
```diff
@@ -0,0 +1,62 @@
+---
+import Icon from './Icon.astro';
+---
+
+<section class="box skills">
+	<div class="stack gap-2 lg:gap-4">
+		<Icon icon="terminal-window" color="var(--accent-regular)" size="2.5rem" gradient />
+		<h2>Full Stack</h2>
+		<p>Some sorta real clever text here but im too tired to think of it now.</p>
```
natemoo-re, line 9: Maybe basic Lorem here? This is a little casual for my tastes.

examples/portfolio/src/pages/about.astro
```diff
@@ -1,77 +1,130 @@
...
-				width: 100%;
-			}
+			<section>
+				<h2 class="section-title">Background</h2>
+				<div class="content">
+					<p>
+						Cream cheese say cheese stinking bishop. <a href="https://astro.build/">Astro</a> makes people
+						happy. Brie fondue hard cheese bocconcini feta camembert de normandie babybel airedale. Red
+						leicester swiss manchego mascarpone pepper jack airedale fromage frais ricotta. Cheese and
+						biscuits cauliflower cheese boursin.
```
natemoo-re, line 30: Lol, I know this is from the original template, but maybe basic Lorem is more approachable?

examples/portfolio/src/pages/work.astro
```diff
@@ -0,0 +1,39 @@
...
+import ContactCTA from '../components/ContactCTA.astro';
+import PortfolioPreview from '../components/PortfolioPreview.astro';
+import type { Project } from '../types';
+import Hero from '../components/Hero.astro';
+import Grid from '../components/Grid.astro';
+
+const projects = (await Astro.glob<Project>('./work/**/*.md')).sort(
+	(a, b) =>
+		new Date(b.frontmatter.publishDate).valueOf() - new Date(a.frontmatter.publishDate).valueOf()
+);
```
natemoo-re, line 13: IMO refactoring to Content Collections would be the way to go here

examples/portfolio/src/styles/global.css
```diff
@@ -70,25 +142,116 @@ img {
...
-	color: var(--t-active);
+	color: var(--link-color);
+}
+
+h1,
+h2,
+h3,
+h4,
+h5 {
+	/* TODO: Reinstate this default margin for Markdown content. */
```
natemoo-re, line 153: Was this TO-DONE?

natemoo-re (commented):
This is looking amazing! Had a few small copy comments, recognizing this is all placeholder content.

I also think using Content Collections would be ideal here since we're excited about it and people will be looking for examples of it 👍🏻

examples/portfolio/src/styles/global.css
delucis: Well spotted! It is now 🙌

tony-sull:
This is looking great!  Followed up with Chris offline to dig into the background images and perf testing, notes below

- looked like the SVG used for background noise was still going to be a perf issue, moving to a tiled PNG
- sticking with Google Fonts to avoid extra NPM deps, lighthouse scores look good enough for the trade-off in simplicity for a starter
- content collections API should clean up style bleed issues

natemoo-re (approved):

PR #5770, opened 2023-01-05
[Content Collections] Allow Zod unions, objects, and transforms as schemas

bholmesdev:
## Changes

- Allow Zod objects, unions, discriminated unions, intersections, and `transform` results as schemas
- Add helpful warning when `z.object(...)` wrapper is missing

### Why?

By enforcing the `z.object(...)` wrapper, we open the door for:
- using a `transform` to add and modify frontmatter properties

```ts
  schema: z.object({
    youtube: z.string().url(),
  }).transform(data => ({
    embedUrl: toEmbedUrl(data.youtube),
    thumbnailUrl: toThumbnailUrl(data.youtube),
  })
});
```

- using unions and discriminated unions for supporting multiple schemas on the same collection. This is common when splitting multiple options based on a `type` property:

```ts
const posts = defineCollection({
	schema: z.discriminatedUnion('type', [
		z.object({
			type: z.literal('post'),
			title: z.string(),
			description: z.string(),
		}),
		z.object({
			type: z.literal('newsletter'),
			subject: z.string(),
		}),
	]),
});
```

## Testing

Add a test for Zod unions

## Docs

https://github.com/withastro/docs/pull/2335

### DIFF
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -4,7 +4,7 @@ import type fsMod from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { createServer, ErrorPayload as ViteErrorPayload, normalizePath, ViteDevServer } from 'vite';
-import { z } from 'zod';
+import { z, ZodType } from 'zod';
 import { AstroSettings } from '../@types/astro.js';
 import { AstroError, AstroErrorData } from '../core/errors/index.js';
 import { astroContentVirtualModPlugin } from './vite-plugin-content-virtual-mod.js';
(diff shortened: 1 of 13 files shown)


matthewp:
Why do we have to require the use of `z.object()` to get those things?

packages/astro/src/content/utils.ts
```diff
@@ -68,8 +68,20 @@ export async function getEntrySlug(entry: Entry, collectionConfig: CollectionCon
 export async function getEntryData(entry: Entry, collectionConfig: CollectionConfig) {
 	let data = entry.data;
 	if (collectionConfig.schema) {
+		// TODO: remove for 2.0 stable release
+		if (
+			typeof collectionConfig.schema === 'object' &&
+			!('safeParseAsync' in collectionConfig.schema)
+		) {
```
natemoo-re, line 75: Seems like we could allow POJOs (plain ole' JS objects) here and automatically wrap them with `z.object` if necessary? Small convenience if you don't need any of the advanced `z.object` features.

natemoo-re (commented):
Definitely in favor of _supporting_ this but I'm not convinced we need to deprecate the current raw object support?

bholmesdev:
@matthewp @natemoo-re Alright, reasons I'm in favor of deprecating the raw object:

Requiring `z.object` makes makes shared schemas easier to document. Allowing POJOs today, we encourage this pattern:

```ts
const schema1 = {...}
const schema2 = {...}

  schema: { ...schema1, ...schema2, moreProperties: true },
});
```
Now, say schema1 needs a transform. schema1 would need to change from a POJO to an object:

```ts
const schema1 = z.object({ ... }).transform(...)
const schema2 = {...}

  schema: ???,
});
```
...and now schema1 and schema2 are difficult to combine. This problem is more pronounced when schemas come from third party libraries.

Requiring schemas to be Zod types, we can always document and encourage a `union` like so:

```ts
const schema1 = z.object({ ... }).transform(...)
const schema2 = z.object({ ... })

  schema: z.union([schema1, schema2, z.object({ moreProperties: true })]),
});
```
I'm also thinking about how much we'd explain if we document transforms or unions. Allowing POJOs, we'd need an "if you're using an object, change to a Zod object first" warning that'd be nice to avoid.

These are fairly minor complaints I know! I just prefer a single standard for doing things to make features discoverable and easy-to-document.

delucis:
+1 to a single consistent API here. Omitting the `z.object()` wrapper is a fairly small aesthetic benefit but sounds like it would increase documentation complexity significantly.

natemoo-re (approved):
I've been convinced! It's a bit boilerplate-y for the simple use case, but I agree that increasing the documentation complexity isn't worth it.

matthewp (approved):

PR #5772, opened 2023-01-05
Output assets to `_astro` directory

### LINKED ISSUE #5745: Move /assets folder to /_astro
Move `/assets/` to `/_astro/`

## Reasons

- Avoid route conflicts
- Sets up a namespace for more endpoints like the Image integration
- Helps httparchive with fingerprinting

Note, it would be nice if this were more easily configured. For ex.:

```js
export default {
  build: {
    assets: 'my-assets/'
  }
};
```

This would match what we already have with `build.server`, `build.client`, etc.

natemoo-re:
## Changes

- Changes output directory from `assets/` to `_astro`
- Better handling for output file naming
- Closes #5745.

## Testing

E2E tests added to verify the build output and `build.assets` configuration

## Docs

Guidance wanted!


packages/astro/src/core/build/static-build.ts
```diff
@@ -133,8 +134,8 @@ async function ssrBuild(opts: StaticBuildOptions, internals: BuildInternals, inp
 				input: [],
 				output: {
 					format: 'esm',
-					chunkFileNames: 'chunks/[name].[hash].mjs',
-					assetFileNames: 'assets/[name].[hash][extname]',
+					chunkFileNames: `chunks/[name].[hash].mjs`,
+					assetFileNames: `${settings.config.build.assets}/[name].[hash][extname]`,
```
matthewp, line 138: Should chunks go on there as well?

matthewp: Oh, this is server so I guess not in this case.

matthewp (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -586,6 +586,23 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		server?: string;
+		/**
+		 * @docs
+		 * @name build.assets
+		 * @type {string}
+		 * @default `'_astro'`
+		 * @description
```
delucis, line 594: We should show when this feature was added:

```suggestion
		 * @version 2.0.0
		 * @description
```

packages/astro/src/@types/astro.ts
```diff
@@ -586,6 +586,23 @@ export interface AstroUserConfig {
...
+		 * Specifies the directory where Astro-generated assets should live.
```
delucis, line 595: Might want to help people understand this is within the build output rather than being able to output to `src/assets` or something. I’ve also suggested some examples for what “Astro-generated assets” are. (I’m also unsure if “generated” is the ideal word vs “processed” or “managed” or… ? But can’t make up my mind so left that as-is for now.)

```suggestion
		 * Specifies the directory in the build output where Astro-generated assets (bundled JS and CSS for example) should live.
```

packages/astro/src/@types/astro.ts
```diff
@@ -586,6 +586,23 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		server?: string;
+		/**
+		 * @docs
+		 * @name build.assets
+		 * @type {string}
+		 * @default `'_astro'`
```
delucis, line 593: Do we also want to add `outDir` as a “see also” here given the final location is `${outDir}/${build.assets}`? 

```suggestion
		 * @default `'_astro'`
		 * @see outDir
```

delucis (commented):
Left some docs suggestions, but looks good! We don’t have any more detailed docs on build output currently but eventually it might make sense to mention this as part of the deployment guides or something. Maybe something like this:

> Static assets will be built to `/_astro` by default (configurable using the `build.assets` option), so that directory can be safely cached immutably if your deployment platform supports it.

delucis (approved):

PR #5773, opened 2023-01-06
Cleanup dependencies

bluwy:
## Changes

- Removed unused dependencies
- Refactor dependencies that're only used once, so it can be removed

I wanted to remove more, but the `astro.js` bin script, and `astro add` command uses a fair amount of packages that are only used once. And they can't be removed or it'll break the core feature.

## Testing

Existing tests should pass.

## Docs

Internal refactor. n/a


packages/astro/src/core/config/schema.ts
```diff
@@ -180,35 +176,6 @@ export const AstroConfigSchema = z.object({
 	legacy: z.object({}).optional().default({}),
 });
 
-interface PostCSSConfigResult {
-	options: Postcss.ProcessOptions;
-	plugins: Postcss.Plugin[];
-}
-
-async function resolvePostcssConfig(inlineOptions: any, root: URL): Promise<PostCSSConfigResult> {
```
bluwy, line 188: This function is unused. The `postcss` and `postcss-load-config` deps can be removed along.

packages/astro/src/core/create-vite.ts
```diff
@@ -43,21 +42,7 @@ const ALWAYS_NOEXTERNAL = new Set([
...
-		try {
-			resolveDependency(dep, projectRoot);
-			noExternalDeps.push(dep);
-		} catch {
-			// ignore dependency if *not* installed / present in your project
-			// prevents hard error from Vite!
-		}
-	}
-	return noExternalDeps;
-}
```
bluwy, line 60: Vite doesn't have an issue now if the deps are not installed. This enables removing `resolveDependency` and the `resolve` package.

packages/astro/src/core/render/ssr-element.ts
```diff
@@ -1,14 +1,15 @@
...
 
 function getRootPath(base?: string): string {
 	return appendForwardSlash(new URL(base || '/', 'http://localhost/').pathname);
 }
 
 function joinToRoot(href: string, base?: string): string {
-	return npath.posix.join(getRootPath(base), href);
+	const rootPath = getRootPath(base);
+	const normalizedHref = slashify(href);
+	return appendForwardSlash(rootPath) + removeLeadingForwardSlash(normalizedHref);
```
bluwy, line 12: This is the only place we use the `path-browserify` dep. I've emulated the posix join with this.


For reference, here's it's implementation: https://github.com/browserify/path-browserify/blob/872fec31a8bac7b9b43be0e54ef3037e0202c5fb/index.js#L180-L197

packages/astro/src/template/4xx.ts
```diff
@@ -1,4 +1,4 @@
-import { encode } from 'html-entities';
+import { escape } from 'html-escaper';
```
bluwy, line 1: `html-entities` and `html-escaper` do the same thing. Since we use `html-escaper` more, I've swap to it. `html-entities` was only used here once.

packages/astro/src/vite-plugin-jsx/tag.ts
```diff
@@ -18,9 +16,6 @@ export default async function tagExportsWithRenderer({
 	rendererName: string;
 	root: URL;
 }): Promise<PluginObj> {
-	const astroServerPath = fileURLToPath(
-		await importMetaResolve('astro/server/index.js', root.toString())
-	);
```
bluwy, line 23: This is not needed anymore as I made a change to dedupe `astro`, which means we can refer to the string unresolved. We do the same for Astro compile too:

https://github.com/withastro/astro/blob/4e37e82b4f4623e99bd58ebd70b5da0879d1713b/packages/astro/src/core/compile/compile.ts#L46

packages/astro/package.json
```diff
@@ -164,7 +156,6 @@
     "strip-ansi": "^7.0.1",
     "supports-esm": "^1.0.0",
     "tsconfig-resolver": "^3.0.1",
-    "typescript": "*",
```
Princesseuh, line 167: This is necessary for `astro check` to work, but it does bring up a good question of if Astro 2.0 is a good time to change this!

bluwy: Ah I couldn't find the peer dep for this. Thanks for pointing out, I'll revert this for now. I do like if Astro doesn't install TypeScript by default, but we could investigate that separately!

Princesseuh: Yep! Spun up a thread on Discord asking for people's opinion on this. We can tackle separately for sure!

bluwy: I've reverted this 👍

matthewp (approved):

PR #5776, opened 2023-01-06
[@astrojs/lit] Fix slotted content

### LINKED ISSUE #5475: [@astrojs/lit] Slotted content is wrapped in `<astro-slot>` elements which breaks native slot functionality

ba55ie:
## Changes

- This fixes #5475.
- It reverts the injected `<astro-slot>` components.
- And restores the `slot="name"` attributes.

I use [`parse5`](https://parse5.js.org/) to parse the HTML string into a _Fragment_ and then add the missing named slots. This is safer than using regex, but if there's a better parser, don't hesitate to change it.

```html
// Before
<my-element>
    <astro-slot slot="header">
        <h1>Header</h1>
    </astro-slot>

    <astro-slot>
        <p>Content</p>
        <p>Some more content</p>
    </astro-slot>

    <astro-slot slot="footer">
        <small>Legal stuff</small>
        <small>© 2023</small>
    </astro-slot>
</my-element>

// After
<my-element>
    <h1 slot="header">Header</h1>

    <p>Content</p>
    <p>Some more content</p>

    <small slot="footer">Legal stuff</small>
    <small slot="footer">© 2023</small>
</my-element>
```

## Testing

- Tests have been added for nested slotted HTML and Web Components.
- The code has been tested in my own projects.

## Docs

The docs do not need updating.

### DIFF
--- a/.changeset/kind-beers-give.md
+++ b/.changeset/kind-beers-give.md
@@ -1,0 +1,6 @@
+---
+'astro': patch
+'@astrojs/lit': patch
+---
+
+Fix Lit slotted content
--- a/packages/astro/test/fixtures/lit-element/src/pages/slots.astro
+++ b/packages/astro/test/fixtures/lit-element/src/pages/slots.astro
@@ -7,9 +7,33 @@ import {MyElement} from '../components/my-element.js';
   <title>LitElement | Slot</title>
 </head>
 <body>
-  <MyElement>
-		<div>default</div>
-		<div slot="named">named</div>
+  <MyElement id="root">
+		<div class="default">my-element default 1</div>
+		<div class="default">my-element default 2</div>
+
+		<h1 slot="named">my-element named 1</h1>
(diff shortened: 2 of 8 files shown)


matthewp:
Thanks!

PR #5786, opened 2023-01-06
[Content collections] Move generated types to `.astro` directory

bholmesdev:
## Changes

- Moves generated types from `src/content/types.generated.d.ts` -> `.astro/types.d.ts`
- Generate a `src/env.d.ts` if none exists with the `astro/client` and `.astro` types
- Add `.astro` reference path to `src/env.d.ts` if one already exists
  
### Why move `types.generated`?

Outputting generated types inside `src/` has some drawbacks:
- Linters and prettier will try to format the file. This has lead to issues in our CI already, as the type generator may not respect your indentation rules.
- Putting generated types alongside your content can be confusing. We've already received support forum requests asking whether to ignore the file. By consolidating all generated types to `.astro`, we can have future-proof recommendations there!

## Testing

- `astro check` on updated `@examples/with-content`
- check that `writeFile` is called on the `env.d.ts` file during `astro sync`

## Docs

https://github.com/withastro/docs/pull/2313


.changeset/curvy-foxes-reply.md
```diff
@@ -0,0 +1,16 @@
...
+
+If you're using Git for version control, we recommend ignoring this generated directory by adding `.astro` to your .gitignore.
+
+#### Migration
+
+You will need a [TypeScript reference path](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-) to include `.astro` types in your project. Running `astro dev`, `astro build`, or `astro sync` will configure this automatically if your project has a `src/env.d.ts` file. Otherwise, you can add a `src/env.d.ts` file manually with the following contents:
+
+```diff
+/// <reference path="astro/client" />
++ /// <reference types="../.astro/types.d.ts" />
```
matthewp, line 15: Does this mean you don't get types without manually creating a file? Or was this already required? Is there anything we can do about that?

bholmesdev: This will be automatically generated if a `src/env.d.ts` is present! Otherwise, we assume you've deleted or moved this file to manage ambient types yourself. Since this file is home to the `astro/client` types, I'd assume most users have this file in their Astro project.

bholmesdev:
Fair question @fflaten! There's a couple reasons for it:
1. To stay consistent with existing frameworks that use generated types, namely [Nuxt](https://nuxtjs.org/docs/directory-structure/nuxt) and [SvelteKit](https://kit.svelte.dev/docs/types#generated-types).
2. To ensure editor tooling keeps up with your edits. Types in `node_modules` are cached by editors like VS Code, so you'd need to restart your TS server periodically while working in `dev`.

That second reason definitely pushed us to types in your project root. Luckily, `.astro/` will be the landing place for any generated code in the future, making it easy to gitignore.

fflaten:
Thanks. Just not used to have generated data (cache) in a config-folder, but I agree the second reason alone is a good enough reason. 👍

matthewp (approved):

packages/astro/test/fixtures/content-collections/.gitignore
```diff
@@ -1 +1 @@
-types.generated.d.ts
+.astro
```
natemoo-re, line 1: This could probably go in the root `.gitignore`?

natemoo-re (approved):
Code LGTM, small comment about simplifying the `.gitignore`

packages/astro/test/fixtures/content-collections/.gitignore
bholmesdev: Agreed! I'll move

PR #5788, opened 2023-01-07
Refactor `Props` of `Image` and `Picture` component to support type checking

MoustaphaDev:
## Changes

- Refactor types to support props auto-completion for the `Image` and `Picture` components.
- Pass previously missing `alt` prop to the `getPicture` function

## Testing

Tested locally. There's only a small runtime change (see `alt` prop passed to the `getPicture` function). All tests should pass

## Docs

N/A

### DIFF
--- a/packages/integrations/image/components/Image.astro
+++ b/packages/integrations/image/components/Image.astro
@@ -2,29 +2,11 @@
 // @ts-ignore
 import { getImage } from '../dist/index.js';
 import { warnForMissingAlt } from './index.js';
-import type { ImgHTMLAttributes } from './index.js';
-import type { ImageMetadata, TransformOptions, OutputFormat } from '../dist/index.js';
-
-interface LocalImageProps
-	extends Omit<TransformOptions, 'src'>,
-		Omit<ImgHTMLAttributes, 'src' | 'width' | 'height'> {
-	src: ImageMetadata | Promise<{ default: ImageMetadata }>;
-	/** Defines an alternative text description of the image. Set to an empty string (alt="") if the image is not a key part of the content (it's decoration or a tracking pixel). */
-	alt: string;
-}
-
-interface RemoteImageProps extends TransformOptions, astroHTML.JSX.ImgHTMLAttributes {
-	src: string;
-	/** Defines an alternative text description of the image. Set to an empty string (alt="") if the image is not a key part of the content (it's decoration or a tracking pixel). */
-	alt: string;
-	format?: OutputFormat;
-	width: number;
-	height: number;
-}
+import type { LocalImageProps, RemoteImageProps } from './index.js';
 
 export type Props = LocalImageProps | RemoteImageProps;
 
-const { loading = 'lazy', decoding = 'async', ...props } = Astro.props as Props;
+const { loading = 'lazy', decoding = 'async', ...props } = Astro.props;
 
 if (props.alt === undefined || props.alt === null) {
 	warnForMissingAlt();
--- a/packages/integrations/image/components/Image.astro
+++ b/packages/integrations/image/components/Image.astro
@@ -2,9 +2,9 @@
 // @ts-ignore
 import { getImage } from '../dist/index.js';
 import { warnForMissingAlt } from './index.js';
-import type { LocalImageProps, RemoteImageProps } from './index.js';
+import type { ImageComponentLocalImageProps, ImageComponentRemoteImageProps } from './index.js';
 
-export type Props = LocalImageProps | RemoteImageProps;
+export type Props = ImageComponentLocalImageProps | ImageComponentRemoteImageProps;
 
 const { loading = 'lazy', decoding = 'async', ...props } = Astro.props;
 
--- a/packages/integrations/image/components/Picture.astro
+++ b/packages/integrations/image/components/Picture.astro
@@ -32,6 +32,7 @@ const { image, sources } = await getPicture({
 	fit,
 	background,
 	position,
+	alt,
 });
 
 delete image.width;
--- a/packages/integrations/image/components/Picture.astro
+++ b/packages/integrations/image/components/Picture.astro
@@ -1,36 +1,9 @@
 ---
 import { getPicture } from '../dist/index.js';
 import { warnForMissingAlt } from './index.js';
-import type { ImgHTMLAttributes, HTMLAttributes } from './index.js';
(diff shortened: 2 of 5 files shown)


packages/integrations/image/components/Image.astro
```diff
@@ -2,29 +2,11 @@
 // @ts-ignore
 import { getImage } from '../dist/index.js';
 import { warnForMissingAlt } from './index.js';
-import type { ImgHTMLAttributes } from './index.js';
-import type { ImageMetadata, TransformOptions, OutputFormat } from '../dist/index.js';
+import type { ImageComponentLocalImageProps, ImageComponentRemoteImageProps } from './index.js';
 
-interface LocalImageProps
-	extends Omit<TransformOptions, 'src'>,
```
MoustaphaDev, line 9: For some reason, when generics are used directly inside the astro component, its props become Record<string, any>. When extracted to another file, it works as expected

packages/integrations/image/components/Picture.astro
```diff
@@ -59,6 +32,7 @@ const { image, sources } = await getPicture({
 	fit,
 	background,
 	position,
+	alt,
```
MoustaphaDev, line 35: Ommiting `alt` was a mistake I guess?

packages/integrations/image/components/Image.astro
natemoo-re: Our language server doesn't support generics at the moment, but it will soon. Your workaround seems valid, though!

cc @Princesseuh for a TS review 😄

natemoo-re (approved):
Looks good to me, thanks for the PR!

tony-sull (approved):
looks good to me!

This code actually was on par with other types back in the day, looks like it shows it's age now with all the improvements to the language server having much stronger support across the board

packages/integrations/image/components/index.ts
```diff
@@ -2,6 +2,54 @@
 export { default as Image } from './Image.astro';
 export { default as Picture } from './Picture.astro';
 
+import type { TransformOptions, OutputFormat } from '../dist/loaders/index.js';
+import type { ImageMetadata } from '../dist/vite-plugin-astro-image.js';
+
+export interface ImageComponentLocalImageProps
+	extends Omit<TransformOptions, 'src'>,
+		Omit<ImgHTMLAttributes, 'src' | 'width' | 'height'> {
```
Princesseuh, line 10: Should be `astroHTML.JSX.ImgHTMLAttributes` too?

natemoo-re: Maybe `HTMLAttributes<'img'>` from [`astro/types`](https://github.com/withastro/astro/blob/main/packages/astro/types.d.ts#L7)?

MoustaphaDev: Sounds good to me!
So we could refactor [ImgHTMLAttributes and  HTMLAttributes](https://github.com/withastro/astro/pull/5788/files#diff-10f5ca07e9c63bb301d686880dfcb4497bd60f41a1f6d64a168f9950f960c7d4R54-R61) too perhaps?

MoustaphaDev: > Maybe `HTMLAttributes<'img'>` from [`astro/types`](https://github.com/withastro/astro/blob/main/packages/astro/types.d.ts#L7)?

Sounds good to me!
So we could refactor [ImgHTMLAttributes and  HTMLAttributes](https://github.com/withastro/astro/pull/5788/files#diff-10f5ca07e9c63bb301d686880dfcb4497bd60f41a1f6d64a168f9950f960c7d4R54-R61) too perhaps?

MoustaphaDev: > Should be `astroHTML.JSX.ImgHTMLAttributes` too?

Probaly not? I don't think we want directives here

packages/integrations/image/components/index.ts
```diff
@@ -1,16 +1,57 @@
 /// <reference types="astro/astro-jsx" />
 export { default as Image } from './Image.astro';
 export { default as Picture } from './Picture.astro';
+import type { HTMLAttributes as AllHTMLAttributes } from '../../../astro/types';
```
Princesseuh, line 4: This import won't work when deployed
```suggestion
import type { HTMLAttributes as AllHTMLAttributes } from 'astro/types';
```

packages/integrations/image/components/index.ts
```diff
@@ -1,16 +1,57 @@
...
+	src: ImageMetadata | Promise<{ default: ImageMetadata }>;
+	/** Defines an alternative text description of the image. Set to an empty string (alt="") if the image is not a key part of the content (it's decoration or a tracking pixel). */
+	alt: string;
+	sizes: HTMLImageElement['sizes'];
+	widths: number[];
+	formats?: OutputFormat[];
+}
+
+export interface PictureComponentRemoteImageProps
+	extends Omit<HTMLAttributes, 'src' | 'width' | 'height'>,
```
Princesseuh, line 39: No point in omitting those here, since they're not on HTMLAttributes

packages/integrations/image/components/index.ts
```diff
@@ -1,16 +1,57 @@
...
+	sizes: HTMLImageElement['sizes'];
+	widths: number[];
+	aspectRatio: TransformOptions['aspectRatio'];
+	formats?: OutputFormat[];
+	background: TransformOptions['background'];
+}
+
+export type ImgHTMLAttributes = AllHTMLAttributes<'img'>;
+
+export type HTMLAttributes = Omit<astroHTML.JSX.HTMLAttributes, 'set:text' | 'set:html' | 'is:raw'>;
```
Princesseuh, line 54: In my opinion, it'd be better to leave the `astro/types` `HTMLAttributes` as is, without renaming it. And this one can be called `GlobalHTMLAttributes`, since it's only the global ones that are included here

Additionally, the list of exclusions can be built from Astro's types, like done in `astro/types`:

```ts
export type HTMLAttributes<Tag extends HTMLTag> = Omit<
	keyof Omit<AstroBuiltinAttributes, 'class:list'>
>;
```

Princesseuh:
Left a few comments! Let me know if you'd rather I push this PR over the line myself, as you've done plenty work already!

MoustaphaDev:
> Left a few comments! Let me know if you'd rather I push this PR over the line myself, as you've done plenty work already!

Thanks! your suggestions where helpful enough, I just need approval now 🙂

Princesseuh (approved):

PR #5797, opened 2023-01-08
Set the AdapterModeNotInitialized error to the proper way

RoadRoller01:
## Changes

- replaced node mode error with one in errors-data
- added errors to the exports so adapters can use the official errors

## known problems
the added error title and message are statically typed with node adapter,
so when a new adapter uses the mode property we should make dynamic

## Testing

i will test with github actions

## Docs
https://docs.astro.build/en/guides/integrations-guide/node/#mode

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -49,6 +49,7 @@
     "./components/*": "./components/*",
     "./content/internal": "./dist/content/internal.js",
     "./debug": "./components/Debug.astro",
+    "./errors": "./dist/core/errors/index.js",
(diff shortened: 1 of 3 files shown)


sarah11918 (changes requested):
Just leaving this here, so that Docs will come back and review the error message/description if it's decided that this will be merged!

Princesseuh:
Thinking a bit more about it, I think the way I highlighted is better, and will allow for overall a better experience for users.

Thank you for contributing, even if your PR wasn't accepted, we're very grateful you choose to contribute to Astro!

PR #5803, opened 2023-01-09
Handle compiler breaking change

bluwy:
## Changes

- Bump a new minor for the compiler
- Handle compiler breaking changes
  - Use the new `filename` and `astroGlobalArgs` option
- Fix minor styles HMR

## Testing

Existing tests should pass. But I also tested the HMR issue manually to make sure it still works.

## Docs

n/a. internal API


packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -17,32 +22,24 @@ export interface HandleHotUpdateOptions {
...
-			// All styles are the same, we can skip an HMR update
-			const styles = new Set(newResult.css);
-			for (const style of oldResult.css) {
-				if (styles.has(style)) {
-					styles.delete(style);
-				}
-			}
-			if (styles.size === 0) {
-				return [];
-			}
```
bluwy, line 45: Since scope hash now derives from `normalizedFilename` only, I created a `isStyleOnlyChanged` function to handle it. It's a bit tedious than before, but I think leaving the HMR logic out of the compiler could be nice for now.

I also remove the `styles.size === 0` handling. The logic isn't quite right given a scenario where:

1. `oldResult` has two `<style>`s
2. User removes one `<style>`
3. `newResult` now has one `<style>`
4. This logic makes `styles.size === 0` and skips HMR.
5. BUG: The removed style's side effect isn't removed as no HMR or reload is done.

It seems that this does handle skipping HMR if no source code change, which I handled on line 37 instead.

packages/astro/test/fixtures/third-party-astro/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
-    "astro": "^1.0.0",
+    "astro": "workspace:*",
```
bluwy, line 6: The 1.0.0 was used to bypass the peer dep warning, but I found this test to use the old 1.0 code instead, breaking the compiler upgrade as both 1.0 and 2.0 code exist at the same time.

matthewp (approved):
Looks great, thanks!

PR #5785, opened 2023-01-06
Remove Astro-flavored Markdown from `@astrojs/markdown-remark`

delucis:
## Changes

Cleans up `@astrojs/markdown-remark` removing syntax tooling that is no longer needed now that we are not supporting Astro-flavored Markdown.

- Removes our custom remark/rehype plugins that powered AFMD
- Removes dependencies for those plugins that aren’t used elsewhere
- Updates our headings plugin to remove some AFMD-specific code
- Removes the `astroFlavoredMd` option from the `renderMarkdown` method

## Testing

I’ve removed AFMD-specific tests.

## Docs

We’re already documenting AFMD’s demise, so no additional docs associated with this PR.

PR #5804, opened 2023-01-09
[markdown-remark] Remove unused dependencies

delucis:
## Changes

Follow up to #5785. I missed that `acorn` and `hast-util-to-html` are no longer used by `@astrojs/markdown-remark` and can be removed as dependencies.

Does this need a patch changeset separate to #5785 or is it minimal enough to ignore?

## Testing
Existing tests should pass.

## Docs

No docs needed.


bluwy (approved):
LGTM! Looks like the lockfile needs another update before merging

delucis:
I tried removing the lock file and running `pnpm install --lockfile-only` like the lockfile CI task does but that doesn’t seem to produce any changes. Not sure what the simple way to resolve this is. Maybe it doesn’t matter?

bluwy:
Looking into the lockfile, it seems to be reverting the changes of the recent lockfile update. I usually do a merge, and during conflicts, run `pnpm i` as `pnpm` supports ootb conflict resolve. Maybe it's worth resetting the branch to `main`, remove the remark deps, and run `pnpm i` again to fix it 🤔

delucis:
This was a good tip. Checked out `main`, ran `pnpm i`, checked out my branch, removed the lockfile, ran `pnpm i` again and it generated the expected lockfile 👍 

I actually didn’t see any conflicts while merging, maybe because I have `pnpm-merge-driver` configured as suggested in [CONTRIBUTING.md](https://github.com/withastro/astro/blob/main/CONTRIBUTING.md#setting-up-your-local-repo)? Anyway, should be good now!

bluwy:
Yeah I never quite understood why `pnpm-merge-driver` was recommended 😅 Glad that's fixed now. PR looks good to me!

PR #5806, opened 2023-01-09
Add Astro as a peerDependency

matthewp:
## Changes

- This adds `astro` as a peerDependency to many of our integrations.
- Only adds for integrations that are already doing a `major` release, in order to avoid unnecessarily creating breaking releases for packages that are not otherwise changing.

## Testing

- N/A, I think.

## Docs

- N/A

### DIFF
--- a/.changeset/thin-seahorses-worry.md
+++ b/.changeset/thin-seahorses-worry.md
@@ -13,4 +13,4 @@
 
 Make astro a peerDependency of integrations
 
-This marks `astro` as a peerDependency of serveral packages that are already getting `major` version bumps. This is so we can more properly track the dependency between them and what version of Astro they are being used with.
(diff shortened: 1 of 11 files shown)


.changeset/thin-seahorses-worry.md
```diff
@@ -0,0 +1,16 @@
...
+'@astrojs/tailwind': major
+'@astrojs/vercel': major
+'@astrojs/vue': major
+'@astrojs/markdown-remark': major
+'@astrojs/image': minor
+---
+
+Make astro a peerDependency of integrations
+
+This marks `astro` as a peerDependency of serveral packages that are already getting `major` version bumps. This is so we can more properly track the dependency between them and what version of Astro they are being used with.
```
bluwy, line 16:
```suggestion
This marks `astro` as a peerDependency of several packages that are already getting `major` version bumps. This is so we can more properly track the dependency between them and what version of Astro they are being used with.
```

bluwy (approved):

PR #5808, opened 2023-01-09
Remove old migration guide link from error docs

Yan-Thomas:
## Changes

The current migration docs (pre-v1) is going to be replaced by a new page, and although we have a redirect for it, the generated error pages in docs trigger an error in our link checker now that the page doesn't exist anymore, therefore we need to remove this link for now (we can later add the new migration docs page).

## Docs

100% docs!


sarah11918:
LGTM, Yan! :rocket:

PR #5811, opened 2023-01-10
Simplify HMR handling

bluwy:
## Changes

Simplifies `vite-plugin-astro`'s `resolveId` hook. It looks like we were adding special cases there to get HMR working, but I found it possible to do less. I'll elaborate more in the review comments below.

## Testing

Existing tests that cover the HMR behaviour should pass. I checked the git history and made sure or edge case changes we have did have tests, so this should be safe.

## Docs

n/a. internal change.


packages/astro/src/core/compile/compile.ts
```diff
@@ -37,7 +35,7 @@ export async function compile({
 		// use `sourcemap: "both"` so that sourcemap is included in the code
 		// result passed to esbuild, but also available in the catch handler.
 		transformResult = await transform(source, {
-			moduleId,
+			moduleId: filename,
 			pathname: filename,
```
bluwy, line 39: I found the `filename` and the `moduleId` passed are always the same (absolute paths or virtual ids). They could only differ in the `resolveId` hook, but since we're compiling in the `transform` hook, it doesn't affect us.

In a future compiler PR, I'll merge `moduleId` and `pathname` options into one.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -27,85 +20,28 @@ interface AstroPluginOptions {
...
-
-	return {
+	const prePlugin: vite.Plugin = {
 		name: 'astro:build',
 		enforce: 'pre', // run transforms before other plugins can
 		configResolved(_resolvedConfig) {
 			resolvedConfig = _resolvedConfig;
 		},
-		// note: don’t claim .astro files with resolveId() — it prevents Vite from transpiling the final JS (import.meta.glob, etc.)
-		async resolveId(id, from, opts) {
```
bluwy, line 63: This entire `resolveId` hook is removed in favour of the next review comment.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -253,6 +182,19 @@ export default function astro({ settings, logging }: AstroPluginOptions): vite.P
...
+	const normalPlugin: vite.Plugin = {
+		name: 'astro:build:normal',
+		resolveId(id) {
+			// If Vite resolver can't resolve the Astro request, it's likely a virtual Astro file, fallback here instead
+			const parsedId = parseAstroRequest(id);
+			if (parsedId.query.astro) {
+				return id;
+			}
+		},
+	};
```
bluwy, line 195: ... Continuing. This `resolveId` hook should cover all our cases instead.

The reason we can remove the pre plugin `resolveId` completely is because the paths can already be resolved by Vite's resolver plugin. Pre plugins run before Vite's resolver. Previously we try to resolve it ourselves, and the edge cases compound because we had to emulate the Vite resolver algorithm to get the flow right (which isn't feasible)

Now we let Vite handle instead, and only catch those that don't work with a normal plugin, which runs after Vite's resolver.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -196,20 +132,14 @@ export default function astro({ settings, logging }: AstroPluginOptions): vite.P
 				return;
 			}
 
-			const filename = normalizeFilename(parsedId.filename, config);
```
bluwy, line 199: In other diffs, you'll see that I had removed `normalizeFilename`. That's because `normalizeFilename` was previously used to emulate Vite's resolver algorithm, and only needed in a pre plugins `resolveId`.

Since this is a `transform` hook, we can remove it as it's already normalized by Vite.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -27,85 +20,28 @@ interface AstroPluginOptions {
...
-			}
-		},
 		async load(id, opts) {
 			const parsedId = parseAstroRequest(id);
 			const query = parsedId.query;
 			if (!query.astro) {
 				return null;
 			}
 			// For CSS / hoisted scripts, the main Astro module should already be cached
-			const filename = normalizeFilename(parsedId.filename, config);
```
bluwy, line 108: This still needs a `normalizeFilename` in `load` as it may receive a root-relative path, like `/src/components/Foo.astro`. This seems to only happen in tests (likely the mocked fs used?), but I leave it here for now just in case.

bholmesdev (approved):
We need HMoRe quality PRs like this 👏 
Learned a bit about Vite's `resolveId` hook as well. Making a note of that. Change seems good assuming tests pass.

packages/astro/src/core/compile/compile.ts
matthewp: So filename *is* the module ID right? To do head propagation we rely on the moduleId being passed to the compiler to match what is found in the module graph. Just want to make sure that is still going to be happening here.

bluwy: Yup, I tested it and saw it's always identical, even for virtual files. I think the unit test should cover if not as I was able to get it to fail when I pass the wrong id during manual testing.

bluwy: I did another quick review, and I think this part I guess I'm slightly unsure:

https://github.com/withastro/astro/blob/840412128b00a04515156e92c314a929d6b94f6d/packages/astro/src/vite-plugin-astro/index.ts#L244-L245

When you added the `context.modules[0]?.id`, is it because it differs from `context.file`? If I understand the module graph right, [Vite's code](https://github.com/vitejs/vite/blob/896475dc6c7e5f1168e21d556201a61659552617/packages/vite/src/node/server/hmr.ts#L85-L92) should have them equal in most case.

matthewp: I think I did that out of "correctness" rather than a concern about `context.file`. So I think we can go with the change for now and see what happens.

natemoo-re (approved):

FDiskas:
It does not work. Try changing `Hello` within the component and it should change without reloading the main page.
https://stackblitz.com/edit/github-m9axad-vp3ckj?file=src%2Fpages%2Findex.astro,src%2Fcomponents%2FCard.astro

PR #5812, opened 2023-01-10
test(create-astro): correct wrong argument

ansidev:
## Changes

- Correct wrong argument: change `--dryrun` to `--dry-run`.

## Testing

- N/A

## Docs

- N/A


.changeset/rotten-onions-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+correct wrong argument in test
```
bluwy, line 5: I don't think we need a changeset for internal test changes.

ansidev: @bluwy I reverted this change

PR #5782, opened 2023-01-06
Drop Node 14 support

Princesseuh:
## Changes

Now that Node 14 is going EOL, we can remove some old stuff we had to support it. I, for one, welcome our new Node 16, 18 and 20 overlords.

## Testing

- Tests updated

## Docs

We'll make a PR to docs indicating that we now have 16 as a minimum version

PR #5814, opened 2023-01-10
Remove more unnecessary polyfills from webapi for Node 16

Princesseuh:
## Changes

Follow-up to https://github.com/withastro/astro/pull/5782 removing more unnecessary polyfills from Webapi. In the future, Node 18 will allow us to remove some other ones (`structuredClone`, `FormData` etc), but that'll be for later

~~Alternatively we move Astro to deno and remove this entire package alltogether~~

## Testing

Tests still pass! I also removed the tests in webapi testing previously polyfilled things, since testing that Node actually has the stuff it says it does seemed a bit redundant

## Docs

N/A


bholmesdev (approved):
Glad to polyfil-ter out these Node 14 deps! Don't have much experience with the webapi package internals, though these deletions seem correct to me

matthewp (approved):

PR #5823, opened 2023-01-10
Run sync as part of `astro check`

delucis:
## Changes

Sites using content collections will usually need their generated types for `astro check` to pass. This PR runs the `sync` command at the start of the check script to avoid users needing to always do `astro sync && astro check` (especially in CI where generated types are less likely to already be present).

@bholmesdev Is `astro sync` a safe no-op if no content collections are configured?

For context: this is already an issue for #5765 where `astro check` is failing in our CI because content types aren’t generated. The other approach would be to only fix this in our CI script but given other users might run into this, I thought this combo would make most sense.

## Testing

No additional testing just yet. Running `pnpm test:match "check"` passes but only shows two relevant tests currently.

## Docs

No docs needed I don’t think?


Princesseuh (approved):
Since we do this already on `dev` and `build`, might as well. If it ever becomes a problem, we could have a `--no-sync` flag or something similar. +1 from the `astro check` end of things!

packages/astro/src/cli/check/index.ts
```diff
@@ -17,9 +18,12 @@ interface Result {
 	hints: number;
 }
 
-export async function check(settings: AstroSettings) {
+export async function check(settings: AstroSettings, { logging }: { logging: LogOptions }) {
 	console.log(bold('astro check'));
 
+	const { sync } = await import('../sync/index.js');
+	await sync(settings, { logging, fs });
```
bholmesdev, line 25:
```suggestion
	const syncRet = await sync(settings, { logging, fs });
	// early exit on sync failure
	if (syncRet === 1) return syncRet;
```
Super pedantic, but check for an early `process.exit` if necessary!

bholmesdev (commented):
Small comment but this PR `check`s out. Also +1 to Erika's comment on adding a `--no-sync` or `--skip-sync` flag if users request it

PR #5825, opened 2023-01-10
[Content collections] Remove experimental flag

bholmesdev:
## Changes

- Removes `experimental.contentCollections` flag

## Testing

- Remove flag wherever used

## Docs

https://github.com/withastro/docs/pull/2354

### DIFF
--- a/packages/integrations/mdx/src/plugins.ts
+++ b/packages/integrations/mdx/src/plugins.ts
@@ -161,9 +161,8 @@ export async function getRemarkPlugins(
 	remarkPlugins = [...remarkPlugins, ...ignoreStringPlugins(mdxOptions.remarkPlugins)];
 
 	// Apply last in case user plugins resolve relative image paths
-	if (config.experimental.contentCollections) {
-		remarkPlugins.push(toRemarkContentRelImageError(config));
-	}
+	remarkPlugins.push(toRemarkContentRelImageError(config));
+
 	return remarkPlugins;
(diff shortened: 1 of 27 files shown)


packages/integrations/mdx/src/plugins.ts
```diff
@@ -161,9 +161,8 @@ export async function getRemarkPlugins(
 	remarkPlugins = [...remarkPlugins, ...ignoreStringPlugins(mdxOptions.remarkPlugins)];
 
 	// Apply last in case user plugins resolve relative image paths
-	if (config.experimental.contentCollections) {
-		remarkPlugins.push(toRemarkContentRelImageError(config));
-	}
+	remarkPlugins.push(toRemarkContentRelImageError(config));
+
```
natemoo-re, line 165: Don't we still need this flag? Relative assets should only be disallowed in the `src/content/` directory?

natemoo-re (commented):
Had a question about disallowing relative assets with this setup!

packages/integrations/mdx/src/plugins.ts
bholmesdev: Hm, `remarkContentRelImageError` checks if a file's in the `contentDir` internally. Shouldn't need a flag before applying (see https://github.com/withastro/astro/blob/main/packages/markdown/remark/src/remark-content-rel-image-error.ts#L13)

PR #5829, opened 2023-01-11
Fix `Code.astro` shiki css class replace logic

### LINKED ISSUE #5828: Code.astro component no longer set `.astro-code` css class

1.9.1

None

pnpm

Windows

### Describe the Bug

The builtin component `Code.astro` doesn't set `.astro-code` css class to the rendered `pre` as it did before.

Starting from shiki v0.12, a new feature reported here https://github.com/shikijs/shiki/pull/376 adds the theme name as css class. Even if the author added it as optional, in the end it's forced, see https://github.com/shikijs/shiki/pull/376#issuecomment-1356843275

This causes the bug, as the replace logic of `Code.astro` doens't work anymore. This bug happens when `astro build`.

**Important**
Actually because the shiki version is pinned to v0.11, this bug doens't occur. When updated to v0.12, the bug will happen. The PR I submitted, just resolves this issue. It's retro compatible with the v0.11 so it can be merged nonetheless.

https://stackblitz.com/edit/github-hwrkkq?file=src%2Fpages%2Findex.astro,package.json&on=stackblitz

### Participation

giuseppelt:
## Changes
The replace uses a regex witch matches anything extra and adds it back to the css class.

Closes #5828

## Testing
As mentioned in the linked issue #5828, the actual pinned shiki version v0.11 does work. The v0.12 causes the bug, as a new feature https://github.com/shikijs/shiki/pull/376 adds an extra css class, making the actual logic to fail.

There's a test already in place https://github.com/withastro/astro/blob/4b16e9ec9929269a16e7950d8fed78779149b0fc/packages/astro/test/astro-component-code.test.js#L37-L41 and it will fail when shiki gets updated to v0.12. This PR resolves the issue.

### DIFF
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -37,7 +37,7 @@ const { code, lang = 'plaintext', theme = 'github-dark', wrap = false } = Astro.
 /** Replace the shiki class name with a custom astro class name. */
 function repairShikiTheme(html: string): string {
 	// Replace "shiki" class naming with "astro"
-	html = html.replace('<pre class="shiki"', '<pre class="astro-code"');
+	html = html.replace(/<pre class="(.*)shiki(.*)"/, '<pre class="$1astro-code$2"');
 	// Handle code wrapping
 	// if wrap=null, do nothing.
 	if (wrap === false) {
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -37,7 +37,7 @@ const { code, lang = 'plaintext', theme = 'github-dark', wrap = false } = Astro.
 /** Replace the shiki class name with a custom astro class name. */
 function repairShikiTheme(html: string): string {
 	// Replace "shiki" class naming with "astro"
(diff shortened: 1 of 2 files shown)


packages/astro/components/Code.astro
```diff
@@ -37,7 +37,7 @@ const { code, lang = 'plaintext', theme = 'github-dark', wrap = false } = Astro.
 /** Replace the shiki class name with a custom astro class name. */
 function repairShikiTheme(html: string): string {
 	// Replace "shiki" class naming with "astro"
-	html = html.replace('<pre class="shiki"', '<pre class="astro-code"');
+	html = html.replace(/<pre class="(.*)shiki(.*)"/, '<pre class="$1astro-code$2"');
```
bluwy, line 40:
```suggestion
	html = html.replace(/<pre class="(.*?)shiki(.*?)"/, '<pre class="$1astro-code$2"');
```

I think it would be nice to make this a bit stricter so it doesn't match `<pre class="shiki boom"><code class="bam">` incorrectly. If shiki ever decide to do that one day

bluwy (approved):
Thanks for the fix!

Issue #5783, opened 2023-01-06
HMR not working when triggered from vite plugin


UstymUkhman:
### What version of `astro` are you using?

1.9.0, 1.9.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm, yarn

### What operating system are you using?

Windows

### Describe the Bug

`vite-plugin-glsl` (for example) is using vite's `server.restart` method to trigger HMR when a shader file is changed and saved. When using this plugin with `astro`, after saving a file, dev server crashes unless the following setting is added to `astro.config.mjs`:

```
export default defineConfig({
  vite: {
    server: {
      port: 3000
    }
  }
});
```

Original bug report [here](https://github.com/UstymUkhman/vite-plugin-glsl/issues/29);

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ctfwyn?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

DaniloArantesF:
It also seems to work fine if you disable watch in the plugin's config. Not sure if it's standard behavior but my dev server also crashed silently and continued to update on file changes.

UstymUkhman:
> It also seems to work fine if you disable watch in the plugin's config.

@DaniloArantesF Yeah, good point, thanks for testing that. With `watch: false` option we don't rely on vite's `server.restart` and astro seems to handle file changes automatically. But I also don't know if that's a standard behavior.

> ...my dev server also crashed silently and continued to update on file changes.

Sorry, but I didn't get this. With plugin's `watch: false` option, it seems to be working just like with vite's server option, so in my tests I haven't see the dev server crashing, but it continued to fetch the latest version of the shader file just like if the plugin's `watch` option was set to `true`.

bluwy:
#5849 should fix this. Note that the restart would cause a weird blank page between page reloads as `vite-plugin-glsl` isn't properly handling the changes. It should use `handleHotUpdate` instead of `configureServer` to detect the changes. Otherwise, `handleHotUpdate` would fallback to a full reload, while the server is still starting up, causing the weird interim blank page.

UstymUkhman:
Thanks for the hint @bluwy ! I will take a look at `handleHotUpdate` on the plugin's side.

PR #5849, opened 2023-01-13
Handle server restart from Vite plugins

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5783

Override Vite's `server.restart` API with our implementation.

## Testing

Added a unit test, and manually tested the repro that it works.

## Docs

n/a. bug fix.


natemoo-re (approved):
Clever, this is exciting!

PR #5851, opened 2023-01-13
[RSS] Get ready for content collections

bholmesdev:
## Changes

- Expose `rssSchema` to type-check content collections for RSS feeds
- Deprecate internal `import.meta.glob` handling and move to a `pagesGlobToRssItems()` helper. This has a few benefits:
  - Simplifies our `items` array to accept a single type. Better for JS docs and docs.astro.build.
  - Makes the "only glob `src/pages/` entries" requirement more explicit with the function name.
  - Makes handling globs and handling content collections feel equally supported. When `import.meta.glob` handling was implicit, this felt like the more encouraged solution.
- Refactor internals to use Zod all the way down. This makes error handling on bad config much clearer!

### README changes

- Change glob example to a content collections example
- Add `rssSchema` and `pagesGlobToRssItems()` reference
- Chore: change `import.meta.env.SITE` recommendation to new `context.site`

## Testing

- Update testing for new APIs
- Migrate glob testing to separate `pagesGlobToRssItems()` tests

## Docs

- update README
- update docs and tutorial https://github.com/withastro/docs/pull/2401

### DIFF
--- a/.changeset/fluffy-cups-travel.md
+++ b/.changeset/fluffy-cups-travel.md
@@ -1,0 +1,52 @@
+---
+'@astrojs/rss': major
+---
+
+Update RSS configuration with our content collections in mind.
+
+1. Expose an `rssSchema` for use with content collections. This ensures are RSS feed properties are present in your frontmatter:
+
+```ts
+import { defineCollection } from 'astro:content';
+import { rssSchema } from '@astrojs/rss';
+
+const blog = defineCollection({
+  schema: rssSchema,
+});
+
+export const collections = { blog };
+```
+
+2. Move implicit `import.meta.glob` handling to a separate `pagesGlobToRssItems()` helper. This simplifies our `items` configuration option to accept a single type, without losing existing functionality:
+
+```ts
+// src/pages/rss.xml.js
+import rss, { pagesGlobToRssItems } from '@astrojs/rss';
+
+export function get(context) {
+  return rss({
+    items: pagesGlobToRssItems(
+      import.meta.glob('./blog/*.{md,mdx}'),
+    ),
+  });
+}
+```
+
+#### Migration
+
+If you rely on our `import.meta.glob` handling, add the `pagesGlobToRssItems()` wrapper to your RSS config:
+
+```diff
+// src/pages/rss.xml.js
+import rss, {
++  pagesGlobToRssItems
+} from '@astrojs/rss';
+
+export function get(context) {
+  return rss({
++    items: pagesGlobToRssItems(
+      import.meta.glob('./blog/*.{md,mdx}'),
++    ),
+  });
+}
+```
--- a/.changeset/fluffy-cups-travel.md
+++ b/.changeset/fluffy-cups-travel.md
@@ -2,7 +2,7 @@
 '@astrojs/rss': major
 ---
 
-Update RSS configuration with our content collections in mind.
+Update RSS configuration with content collections in mind.
 
 1. Expose an `rssSchema` for use with content collections. This ensures are RSS feed properties are present in your frontmatter:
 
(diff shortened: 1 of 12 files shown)


bholmesdev:
Updating to keep `import.meta.glob()` handling around as a deprecated API, and dropping to a non-breaking minor!

.changeset/fluffy-cups-travel.md
```diff
@@ -0,0 +1,52 @@
+---
+'@astrojs/rss': major
+---
+
+Update RSS configuration with content collections in mind.
```
natemoo-re, line 5: I'd suggest putting these points last and moving the migration guide first. Coming from someone who is migrating from the previous major to this major, I'd want to know the relevant breaking changes that I need to make first. I might not know what "content collections" is.

packages/astro-rss/src/index.ts
```diff
@@ -1,113 +1,127 @@
...
+type ValidatedRSSOptions = z.infer<typeof rssOptionsValidator>;
+
+const rssFeedItemValidator = rssSchema.extend({ link: z.string(), content: z.string().optional() });
+const rssOptionsValidator = z.object({
+	title: z.string(),
+	description: z.string(),
+	site: z.string(),
+	items: z.array(rssFeedItemValidator),
+	xmlns: z.record(z.string()).optional(),
+	drafts: z.boolean().default(false),
```
natemoo-re, line 60: Typo?
```suggestion
	draft: z.boolean().default(false),
```

natemoo-re (commented):
Seems like a great refactor! Had a few comments around messaging and maybe found a typo?

packages/astro-rss/src/index.ts
bholmesdev: Actually not! This mirrors the naming of `markdown.drafts`, aka "include drafts yes or no." I fought to rename to `includeDrafts` and lost 😅

.changeset/fluffy-cups-travel.md
bholmesdev: Understood. Redrafted and bumped down to a minor! This is now deprecated, not removed 👍

sarah11918 (commented):
Quick Docs README review:

I've reviewed the accompanying PR to docs, which I know contains some of the same examples and sections. I'll let you look over those comments, and decide which ones are applicable here. It's OK if this README is more technical than the docs RSS page, which should be more oriented to practical usage. 

But, I want to give you a chance to look over that review and 
a) see if there's any suggestion there that you'd choose to carry into here. 
b) decide whether there's anything you're happy to leave exclusively to docs, and cut from the README. (We favour as little duplication as possible, and keeping the READMES to a minimum, where it makes sense. This document won't often have to stand on its own since readers are far more likely to encounter the docs page, and search only works in docs proper, etc.)

Then, I'll take another last look over this, but I don't see anything that's going to hold this up!

packages/astro-rss/README.md
```diff
@@ -177,6 +177,58 @@ Will inject the following XML:
 <rss xmlns:h="http://www.w3.org/TR/html4/"...
 ```
 
+### content
+
+The `content` key contains the full content of the post as HTML. This allows you to make your entire post content available to RSS feed readers.
+
+**Note:** Whenever you're using HTML content in XML, suggest using a package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) in order to make sure that your content is properly sanitized, escaped, and encoded.
```
sarah11918, line 184: WE suggest using a package... (or, whatever fix you made in the Docs PR)

packages/astro-rss/README.md
```diff
@@ -152,6 +177,62 @@ Will inject the following XML:
...
```
sarah11918, line 184:
```suggestion
**Note:** Whenever you're using HTML content in XML, we suggest using a package like [`sanitize-html`](https://www.npmjs.com/package/sanitize-html) in order to make sure that your content is properly sanitized, escaped, and encoded.
```

sarah11918 (approved):
Just re-add that missing "we" and good to go from Docs, @bholmesdev!

PR #5852, opened 2023-01-13
Support envPrefix in Astro config

### LINKED ISSUE #5827: Astro ignores vite.envPrefix only during production build

1.9.1

None

npm

Mac

### Describe the Bug

change your configuration to add `vite.envPrefix`:
```
export default defineConfig({vite: {
    envPrefix: 'VITE_',
}});
```

For both scripts `PUBLIC_P=1 VITE_P=2` was set in environment, both those values are accessible during SSR, but only one of them is going to be ignored on client (depending on the way you run it).

Running `npm run dev`:
image
Running `npm run build && npm run preview`:
image

My explanation is that it seems like Astro is ignoring `astro.vite` configuration during builds, but uses it during dev

https://github.com/JLarky/astro-env-prefix-bug

### Participation

rishi-raj-jain:
Fixes #5827 

## Changes

- What does this change?

### DIFF
--- a/.changeset/twenty-llamas-type.md
+++ b/.changeset/twenty-llamas-type.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+fix envPrefix in Astro.config
--- a/.changeset/twenty-llamas-type.md
+++ b/.changeset/twenty-llamas-type.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
-fix envPrefix in Astro.config
+Respect `vite.envPrefix` if provided
--- a/packages/astro/src/core/build/static-build.ts
+++ b/packages/astro/src/core/build/static-build.ts
@@ -161,7 +161,7 @@ async function ssrBuild(opts: StaticBuildOptions, internals: BuildInternals, inp
 			// SSR needs to be last
 			ssr && vitePluginSSR(internals, settings.adapter!),
 		],
-		envPrefix: 'PUBLIC_',
+		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
 		base: settings.config.base,
 	};
 
@@ -230,7 +230,7 @@ async function clientBuild(
 			}),
 			...(viteConfig.plugins || []),
 		],
-		envPrefix: 'PUBLIC_',
+		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
 		base: settings.config.base,
 	};
 
--- a/packages/astro/src/core/build/static-build.ts
+++ b/packages/astro/src/core/build/static-build.ts
@@ -230,7 +230,7 @@ async function clientBuild(
 			}),
 			...(viteConfig.plugins || []),
 		],
-		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
+		envPrefix: viteConfig.envPrefix ?? 'PUBLIC_',
 		base: settings.config.base,
 	};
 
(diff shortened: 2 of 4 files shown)


packages/astro/src/core/build/static-build.ts
```diff
@@ -161,7 +161,7 @@ async function ssrBuild(opts: StaticBuildOptions, internals: BuildInternals, inp
 			// SSR needs to be last
 			ssr && vitePluginSSR(internals, settings.adapter!),
 		],
-		envPrefix: 'PUBLIC_',
+		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
```
natemoo-re, line 164: I don't think we need to add our own top-level `envPrefix` option, we just need to support Vite's `envPrefix` option.

```suggestion
		envPrefix: viteConfig.envPrefix ?? 'PUBLIC_',
```

packages/astro/src/core/build/static-build.ts
```diff
@@ -230,7 +230,7 @@ async function clientBuild(
 			}),
 			...(viteConfig.plugins || []),
 		],
-		envPrefix: 'PUBLIC_',
+		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
```
natemoo-re, line 233: Same here
```suggestion
		envPrefix: viteConfig.envPrefix ?? 'PUBLIC_',
```

packages/astro/src/core/config/schema.ts
```diff
@@ -14,6 +14,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 	publicDir: './public',
 	outDir: './dist',
 	base: '/',
+	envPrefix: 'PUBLIC_',
```
natemoo-re, line 17: Would prefer not to add this
```suggestion
```

packages/astro/src/core/config/schema.ts
```diff
@@ -62,6 +63,7 @@ export const AstroConfigSchema = z.object({
 		.url()
 		.optional()
 		.transform((val) => (val ? appendForwardSlash(val) : val)),
+	envPrefix: z.string().optional().default(ASTRO_CONFIG_DEFAULTS.envPrefix),
```
natemoo-re, line 66:
```suggestion
```

packages/astro/src/core/create-vite.ts
```diff
@@ -110,7 +110,7 @@ export async function createVite(
 		],
 		publicDir: fileURLToPath(settings.config.publicDir),
 		root: fileURLToPath(settings.config.root),
-		envPrefix: 'PUBLIC_',
+		envPrefix: settings.config.envPrefix ?? 'PUBLIC_',
```
natemoo-re, line 113:
```suggestion
		envPrefix: settings.config.vite?.envPrefix ?? 'PUBLIC_',
```

natemoo-re (changes requested):
I think we can fix this without adding a new top-level `envPrefix` option. We try to rely on the `vite` config option for things that are already supported by Vite.

.changeset/twenty-llamas-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+fix envPrefix in Astro.config
```
natemoo-re, line 5:
```suggestion
---
'astro': patch
---

Respect `vite.envPrefix` if provided
```

natemoo-re (approved):
Amazing, thanks for the fix!

PR #5854, opened 2023-01-13
fix: Assets generated to public during build don't get copied to the final result

### LINKED ISSUE #5669: Assets generated to `public` during build don't get copied to the final result

1.7.2

N/A

pnpm

macOS

### Describe the Bug

If a page generate an asset in the `public` folder at build time (ex: using `writeFileSync`), the asset won't be copied to `dist` when building unless already present before building. I suspect this happens because `public` gets copied before the routes are built

In the reproducible example, make sure the `epic.data` file does not exists in `public` before running `astro build` to see the issue

This notably affects [astro-eleventy-img](https://github.com/Princesseuh/astro-eleventy-img), a decently popular user-land integration

https://stackblitz.com/edit/github-apd1w3?file=src/pages/index.astro

### Participation

rishi-raj-jain:
Fixes #5669 

We don't need to do this with output: 'server' as the pages aren't built while creating the build.


natemoo-re (commented):
I'm not positive if this is supposed to be supported or not, but it does seem like a small enough change... Still, is this a pattern we want to enable? Would love for others to weigh in.

Princesseuh:
> I'm not positive if this is supposed to be supported or not, but it does seem like a small enough change... Still, is this a pattern we want to enable? Would love for others to weigh in.

I'm obviously biased, since I maintain an integration that's affected, but I think that enabling this pattern is good! 

It's actually unintuitive to me that the public copy happens before the build. Due to the nature of it (just a copy of a folder), I would've expected it to be the last step, funnily enough 

For what it's worth, this also used to work correctly, I'm not sure what version broke it, but it worked fine when I first made `astro-eleventy-img` (which admittedly, was pre 1.0)

natemoo-re (approved):

FredKSchott:
How are the files copied today? public already ends up in `dist`, so it feels weird that the fix here is to *add* more copy logic vs. changing existing copy logic. Assuming that this fixes the problem by adding a second, redundant copy step?

rishi-raj-jain:
@FredKSchott 

So yeah 2 copies occur (in this case), one before app is built, the other after the whole app is built. The second one is required for this case as during building the app, some assets are dynamically generated.

Say during building a given page, it removes a file from the public directory (which is already copied before building). This process of copying twice will ensure assets initially and generated both are included.

FredKSchott:
I want others to weigh in, but this is a pretty heavy-handed workaround that will impact all users, just to solve a more rare use-case of public assets being generated from inside the Astro during the running of Astro. I would argue that that is not safe to do or rely on anyway, and would give the false impression that it is supported in Astro. 

I guess we could decide that we do want to support this, but there are plenty of reasons not to. For one, this app would break when switching to SSR, and one goal of the project is that SSG -> SSR should be as "drop in" as possible.

The workaround is to generate these assets in a separate script, outside of Astro, before Astro starts up. You can also call this script in your `astro.build.js` config on one of the plugin startup hooks.

FredKSchott (changes requested):
marking as "changes requested" just to make sure this doesn't get merged accidentally until issue is resolved, since it otherwise has an approval.

matthewp:
I also don't think we should support this. An integration can copy directly to to `ourDir` during the build, as many integrations already do (such as Netlify to create the `_redirects` file). 

I think the desire to write to public comes from before we had integrations and components wanted to do some build step. Now that we have integrations, that is the way to go.

matthewp:
Closing as we decided against this solution.

PR #5855, opened 2023-01-13
Remove legacy compiler handling

natemoo-re:
## Changes

- The compiler has been improved to the point where nobody really hits unexpected Wasm stacktraces anymore. We can remove our special handling of compiler wasm errors now.

## Testing

N/A, untested code path

## Docs

Refactor only

### DIFF
--- a/.changeset/angry-spoons-flow.md
+++ b/.changeset/angry-spoons-flow.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Remove legacy compiler error handling
--- a/packages/astro/src/vite-plugin-astro/compile.ts
+++ b/packages/astro/src/vite-plugin-astro/compile.ts
@@ -124,31 +124,5 @@ async function enhanceCompileError({
 		}
 	}
 
-	// improve compiler errors
-	if (err.stack && err.stack.includes('wasm-function')) {
-		const search = new URLSearchParams({
-			labels: 'compiler',
-			title: '🐛 BUG: `@astrojs/compiler` panic',
-			template: '---01-bug-report.yml',
-			'bug-description': `\`@astrojs/compiler\` encountered an unrecoverable error when compiling the following file.
-
-**${id.replace(fileURLToPath(config.root), '')}**
-\`\`\`astro
-${source}
-\`\`\``,
-		});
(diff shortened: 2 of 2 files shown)


matthewp (approved):

natemoo-re:
Actually, we might want to hang on to this, but the error is pretty inscrutable at the moment. Going to refactor this to use the new error overlay info.

This issue still repros and throws the error https://github.com/withastro/compiler/issues/553

sarah11918 (approved):
Astro Docs approves this message!

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -17,6 +17,7 @@ export const AstroErrorData = defineErrors({
 	UnknownCompilerError: {
 		title: 'Unknown compiler error.',
 		code: 1000,
+		hint: 'This is almost always a problem with the Astro compiler, not your code. Please open an issue at https://astro.build/issues.',
```
bluwy, line 20: Do we want to add a new vanity link like https://astro.build/issues/compiler so it directs to the compiler repo instead?

natemoo-re: https://github.com/withastro/astro.build/pull/506

packages/astro/src/core/errors/overlay.ts
```diff
@@ -522,6 +522,15 @@ class ErrorOverlay extends HTMLElement {
 
 		const el = this.root.querySelector(selector);
 
+		if (html) {
+			// Automatically detect links
+			text = text.split(' ').map(v => {
+				if (!v.startsWith('https://')) return v;
+				if (v.endsWith('.')) return `<a href="${v.slice(0, -1)}">${v.slice(0, -1)}</a>.`
+				return `<a href="${v}">${v}</a>`
```
Princesseuh, line 530:
```suggestion
				if (v.endsWith('.')) return `<a target="_blank" href="${v.slice(0, -1)}">${v.slice(0, -1)}</a>.`
				return `<a target="_blank" href="${v}">${v}</a>`
```

This isn't super relevant at the moment, but links shouldn't open in the same tab, or users will lose their HMR state (the see docs link is also `target="_blank"`, so are Markdown links)

PR #5864, opened 2023-01-15
Add missing `noPropertyAccessFromIndexSignature` to strictest tsconfig profile

### LINKED ISSUE #5861: Missing `noImplicitAny` rule in "strict" tsconfig
This is not a real bug as it doesn't affect runtime, but I feel like the strictest TypeScript configuration should enable a few rules by default: `noImplicitAny`, `noImplicitThis`, `noPropertyAccessFromIndexSignature`, `useUnknownInCatchVariables` and `strictPropertyInitialization`.

Affects latest version.

### Participation

ClementNerma:
Add missing `noPropertyAccessFromIndexSignature` to strictest tsconfig profile

Fixes #5861.

cc @Princesseuh

### DIFF
--- a/.changeset/smooth-cycles-sing.md
+++ b/.changeset/smooth-cycles-sing.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 The `noPropertyAccessFromIndexSignature` typescript configuration option was added to the `strictest` profile.
--- a/.changeset/smooth-cycles-sing.md
+++ b/.changeset/smooth-cycles-sing.md
@@ -1,0 +1,7 @@
+---
+'astro': minor
+---
+
+The `noPropertyAccessFromIndexSignature` typescript configuration option was added to the `strictest` profile.
+
+This will have no impact at runtime, but may affect code checks on codebases which do not already use this check.
--- a/packages/astro/tsconfigs/strictest.json
+++ b/packages/astro/tsconfigs/strictest.json
@@ -8,6 +8,8 @@
     "noImplicitOverride": true,
     // Force functions to specify that they can return `undefined` if a possible code path does not return a value.
     "noImplicitReturns": true,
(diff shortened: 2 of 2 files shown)


ClementNerma:
I don't know if I should put anything in the change set, given it won't apply any modification to existing projects, just stricten TypeScript's rules?

bluwy:
I think it's still good to add a changeset since this affects existing projects that extends from Astro's tsconfig. So the changelog should reflect it in case they notice new errors with it.

ClementNerma:
Just added a changeset.

.changeset/smooth-cycles-sing.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': minor
```
bluwy, line 2:
```suggestion
'astro': patch
```

bluwy (approved):

bluwy:
This seems to cause CI issues after merging: https://github.com/withastro/astro/actions/runs/3935549889/jobs/6731329981. Reading the errors, it seems that some can't really be fixed now. E.g. there doesn't seem to be a way to type `Astro.params`, so you'd have to access with the `['key']` syntax, which I find a bit awkward 🤔 

I'll revert this for now as this may be surprising change for those with the strictest config. We could try to include it again in the future.

ClementNerma:
> This seems to cause CI issues after merging: https://github.com/withastro/astro/actions/runs/3935549889/jobs/6731329981. Reading the errors, it seems that some can't really be fixed now. E.g. there doesn't seem to be a way to type `Astro.params`, so you'd have to access with the `['key']` syntax, which I find a bit awkward 🤔
> 
> I'll revert this for now as this may be surprising change for those with the strictest config. We could try to include it again in the future.

I find this behavior perfectly normal as it's indeed misleading to read some property access on an untyped object.

But I agree this can be surprising for such a common case for now.

PR #5871, opened 2023-01-16
Fix image integration not working on Node 18+

Princesseuh:
## Changes

When we removed support for Node 14, we migrated to undici. This caused an issue with our image integration because `squoosh` uses a fetch to load a local `.wasm` file, which is not supported by undici (and is unlikely to be for some time). 

This PR removes the code for that and instead opt to use the fallback only. This is also what Next.js does, so it should be okay!

## Testing

Tests still pass + manual tests

## Docs

N/A

### DIFF
--- a/packages/integrations/image/package.json
+++ b/packages/integrations/image/package.json
@@ -48,8 +48,7 @@
     "kleur": "^4.1.5",
     "magic-string": "^0.25.9",
     "mime": "^3.0.0",
-    "slash": "^4.0.0",
-    "web-streams-polyfill": "^3.2.1"
+    "slash": "^4.0.0"
   },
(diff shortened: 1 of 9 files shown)


Princesseuh:
!preview image-node18

Princesseuh:
!preview imageeight

bluwy (approved):
Looks like it needs a changeset, otherwise this looks good!

I was curious if this would drop support for worker or Deno environments, but it looks like these files are meant to run in Node.js (especially given the file names include `_node_`), so I think we're safe here too.

aFuzzyBear (commented):
Surgical precision 👩‍⚕️

Issue #5061, opened 2022-10-12
Cannot find namespace 'JSX'


chshouyu:
### What version of `astro` are you using?

1.4.7

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

I create a project use `npm create astro@latest`, and select the `blog` template, after it is done, I use this build script:
```
"build": "astro check && tsc --noEmit && astro build"
```
it is from https://docs.astro.build/en/guides/typescript/#type-checking

it report some typescript error:

 ```
node_modules/@types/mdx/index.d.ts:66:58 - error TS2503: Cannot find namespace 'JSX'.

66     export default function MDXContent(props: MDXProps): JSX.Element;
                                                            ~~~

node_modules/@types/mdx/index.d.ts:76:14 - error TS2300: Duplicate identifier 'default'.

76     export { default } from '*.mdx';
                ~~~~~~~

  node_modules/astro/client-base.d.ts:17:17
    17  export default load;
                       ~~~~
    'default' was also declared here.

node_modules/@types/mdx/types.d.ts:4:51 - error TS2503: Cannot find namespace 'JSX'.

4 type FunctionComponent<Props> = (props: Props) => JSX.Element | null;
                                                    ~~~

node_modules/@types/mdx/types.d.ts:6:52 - error TS2503: Cannot find namespace 'JSX'.

6 type ClassComponent<Props> = new (props: Props) => JSX.ElementClass;
                                                     ~~~

node_modules/@types/mdx/types.d.ts:11:65 - error TS2503: Cannot find namespace 'JSX'.

11     [key: string]: NestedMDXComponents | Component<any> | keyof JSX.IntrinsicElements;
                                                                   ~~~

node_modules/@types/mdx/types.d.ts:22:19 - error TS2503: Cannot find namespace 'JSX'.

22     [Key in keyof JSX.IntrinsicElements]?: Component<JSX.IntrinsicElements[Key]> | keyof JSX.IntrinsicElements;
                     ~~~

node_modules/@types/mdx/types.d.ts:22:54 - error TS2503: Cannot find namespace 'JSX'.

22     [Key in keyof JSX.IntrinsicElements]?: Component<JSX.IntrinsicElements[Key]> | keyof JSX.IntrinsicElements;
                                                        ~~~

node_modules/@types/mdx/types.d.ts:22:90 - error TS2503: Cannot find namespace 'JSX'.

[...]

PR #5872, opened 2023-01-16
Enable skipLibCheck by default

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5061

Disable typechecking libraries and .d.ts files. This change was also debatable in Vite before, but the tradeoff (of not typechecking local .d.ts files) was worth it as it's otherwise surfacing type errors unintentionally from a library. (More context: https://github.com/vitejs/vite/issues/7867)

## Testing

n/a

## Docs

n/a

### DIFF
--- a/.changeset/witty-pugs-drum.md
+++ b/.changeset/witty-pugs-drum.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Enable `skipLibCheck` by default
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
It's unfortunate that it's needed, but everyone does it for a reason. LGTM!

PR #5284, opened 2022-11-02
fix(astro): add `class:list` to `HTMLAttributes`

herteleo:
## Changes

- This adds the missing `class:list` to the `HTMLAttributes` type as explained in #5279

## Testing

Tested manually within `examples/basic/src/components/Card.astro` by extending the `Props` with `HTMLAttributes<"div">` and destructuring `'class:list': classList` from `Astro.props as Props`.

## Docs

The docs don't explicitly mention any props of HTMLAttributes.

PR #5879, opened 2023-01-17
fix(astro): add define:vars to HTMLAttributes

ZerdoX-x:
## Changes

- This adds the missing `define:vars` to the `HTMLAttributes` type as explained the same in #5284

## Testing

Was not tested.

## Docs

The docs don't explicitly mention any props of HTMLAttributes.

### DIFF
--- a/.changeset/beige-dots-pay.md
+++ b/.changeset/beige-dots-pay.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Include missing `define:vars` within `HTMLAttributes` type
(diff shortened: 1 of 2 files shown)


ZerdoX-x:
Not sure why this is not fixed yet or even implemented initially.
image

Princesseuh:
Hello, thank you for this contribution!

The `define:vars` attribute is only available on `style` and `script`, hence why you get an error when using it somewhere else. Your PR also wouldn't fix this, as it updates the `HTMLAttributes` type that people can use manually, not the one that is used for checking the template.

ZerdoX-x:
Oh.
That's cool. That is actually more convenient. Sorry for my misunderstanding and thank you for your response <3

PR #4756, opened 2022-09-14
Only pass through children prop if there are children

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/4753

## Testing

- Test added

## Docs

N/A, bug fix

PR #5886, opened 2023-01-17
Support passing `children` as props to a React component

### LINKED ISSUE #5493: Passing `children` as props to a react component breaks in production.

1.6.11

None

yarn

linux

### Describe the Bug

If you're passing `children` as a prop to a react component from a `.astro` file, instead of doing it via markup, eg:
```
<MyComponent children="test" />
```

The component will receive `{children: null}` when building. It works fine in `astro dev`, but breaks in `astro build` when the island is loaded. Since this is perfectly viable React code, i'd regard this as a bug.

https://stackblitz.com/edit/github-l4gigd

### Participation

- [x] I am willing to submit a pull request for this issue. As long as someone can point me in the right direction.

HiDeoo:
## Changes

This PR fixes #5493.

Passing `children` as props to a React component, e.g. `<WithChildren client:load children="test" />` would break in production with the component disappearing after hydration and React complaining with the following error:

```
Hydration failed because the initial UI does not match what was rendered on the server.
```

When creating the static HTML element for `children`, the `props.children` props was [previously ignored](https://github.com/withastro/astro/blob/28556a89fe1b63c88951834effd39630c7a52f90/packages/integrations/react/server.js#L73). This PR introduces a change to fallback to `props.children` if `children` is not defined which fixes the issue.

This PR also adds this change to `packages/integrations/react/server-v17.js` for the React 17 integration, altho, while adding this, I noticed that the changes introduced in #4756 were not applied to the React 17 integration so I added them as well.

## Testing

I updated the `react-component` fixture for `astro` to include a new `WithChildren` component rendering the `children` props and updated the associated test to render this component twice, once with `children` and once with `props.children` and assert that the rendered HTML is the same (which would fail without the changes in this PR as the `<astro-slot>` was missing in the second case).

I also manually tested locally this change in a local repro using the `file:` pnpm protocol and confirmed a component using `children` passed down as props is now properly rendered in production and the React hydration error is gone.

## Docs

I could not find any specific mention regarding this specific behavior in the docs and I'm not sure if it's worth mentioning it.

### DIFF
--- a/.changeset/fresh-eels-speak.md
+++ b/.changeset/fresh-eels-speak.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+Support passing `children` as props to a React component
--- a/packages/astro/test/fixtures/react-component/src/components/WithChildren.jsx
+++ b/packages/astro/test/fixtures/react-component/src/components/WithChildren.jsx
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export default function ({ children }) {
+  return <div className="with-children">{children}</div>;
+}
--- a/packages/astro/test/fixtures/react-component/src/pages/index.astro
+++ b/packages/astro/test/fixtures/react-component/src/pages/index.astro
@@ -7,6 +7,7 @@ import {Research2} from '../components/Research.jsx';
 import Pure from '../components/Pure.jsx';
 import TypeScriptComponent from '../components/TypeScriptComponent';
 import CloneElement from '../components/CloneElement';
+import WithChildren from '../components/WithChildren';
 
 const someProps = {
   text: 'Hello world!',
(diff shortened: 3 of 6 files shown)


HiDeoo:
Not sure what is going on regarding the CI failure only on macOS as I cannot repro locally (even on macOS) and it looks like it is related to a timeout issue in a Astro CLI related test which should not be impacted by this pull request 😕

Princesseuh:
> Not sure what is going on regarding the CI failure only on macOS as I cannot repro locally (even on macOS) and it looks like it is related to a timeout issue in a Astro CLI related test which should not be impacted by this pull request 😕 

The test is probably just flaky, no worries!

natemoo-re (approved):
This is a great fix, thank you!

PR #5888, opened 2023-01-18
Fix image integration crash on Netlify Functions due to `import.meta.url`

Princesseuh:
## Changes

Netlify Functions are still on CJS, however they support ESM files by running them through `esbuild`, this is great, but `esbuild` doesn't polyfill `import.meta.url` so our function just immediately crash. 

This PR add a helper that either return `import.meta.url` if supported, or an URL using `__dirname` so it works both directly in ESM and hazardly transpiled CJS

**Note**

There's apparently a way to get Netlify functions to run in pure ESM mode already, using a per-function config file, but it's undocumented and I could never get it to work despite multiple attempts 🤷‍♀️

Netlify is shipping pure ESM functions natively with no hidden flag soon, so this is really just a temporary workaround that we'll be able to remove in the future (or keep if it affects other hosts, who knows)

## Testing

Tested manually, this issue only shows up when running directly on Netlify

## Docs

N/A

### DIFF
--- a/packages/integrations/image/src/vendor/squoosh/emscripten-utils.ts
+++ b/packages/integrations/image/src/vendor/squoosh/emscripten-utils.ts
@@ -1,5 +1,5 @@
-// 
-import { fileURLToPath } from 'node:url'
+//
+import { fileURLToPath, pathToFileURL } from 'node:url'
 
 export function pathify(path: string): string {
   if (path.startsWith('file://')) {
@@ -28,4 +28,17 @@ export function instantiateEmscriptenWasm<T extends EmscriptenWasm.Module>(
 
 export function dirname(url: string) {
 	return url.substring(0, url.lastIndexOf('/'))
+}
+
+/**
+ * On certain serverless hosts, our ESM bundle is transpiled to CJS before being run, which means
+ * import.meta.url is undefined, so we'll fall back to __dirname in those cases
+ * We should be able to remove this once https://github.com/netlify/zip-it-and-ship-it/issues/750 is fixed
+ */
+export function getModuleURL(url: string | undefined): string {
+	if (!url) {
+		return pathToFileURL(__dirname).toString();
+	}
+
+	return url
 }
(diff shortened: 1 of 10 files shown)


packages/integrations/image/src/vendor/squoosh/emscripten-utils.ts
```diff
@@ -29,3 +29,16 @@ export function instantiateEmscriptenWasm<T extends EmscriptenWasm.Module>(
...
 }
+
+/**
+ * On certain serverless hosts, our ESM bundle is transpiled to CJS before being run, which means
+ * import.meta.url is undefined, so we'll fall back to __dirname in those cases
+ * We should be able to remove this once https://github.com/netlify/zip-it-and-ship-it/issues/750 is fixed
+ */
+export function getModuleURL(url: string | undefined): string {
+	if (!url) {
+		return pathToFileURL(__dirname).toString();
```
bluwy, line 40: The `__dirname` refers relative to this utils file. Would it work correctly if it's used but the consumer file later on?

I see some doing 
```js
scriptDirectory = dirname(getModuleURL(import.meta.url)) + '/'
```

which might have a different value depending on whether `import.meta.url` is supported.

Princesseuh: It's admittedly not ideal, but it does work because the __dirname call is only used in SSR where everything gets bundled 😅

If you have a better solution, I'll definitely take! Got lost in between the `__dirname`, `__filename` and `import.meta.url` while working on this so really I settled on the solution that actually did work 😵‍💫

bluwy: Ah I see. Didn't know this was bundled so it would always have the same value.

This trick sounds good to me then, though I wonder if it should be `__filename` instead since that reflects `import.meta.url`. If you've tested with `__dirname` and it works then I'll be fine by it too 👍

bluwy (approved):

PR #5891, opened 2023-01-18
Cleanup deprecated Markdown APIs from 0.X

bholmesdev:
## Changes

- Remove `content.astro` warning messages
- Remote deprecated `getHeaders()` function
- Remove `rawContent()` and `compiledContent()` errors from MDX and type signature. Users expect that checking `file.rawContent === undefined` will allow safe calling on MD and MDX files. By exporting these functions _just_ to throw errors, we made it harder to use them safely!

## Docs

All deprecated APIs, so no docs changes!

### DIFF
--- a/.changeset/calm-emus-raise.md
+++ b/.changeset/calm-emus-raise.md
@@ -1,6 +1,6 @@
 ---
-'astro': patch
-'@astrojs/mdx': patch
+'astro': major
+'@astrojs/mdx': minor
 ---
 
 Remove deprecated Markdown APIs from Astro v0.X. This includes `getHeaders()`, the `.astro` property for layouts, and the `rawContent()` and `compiledContent()` error messages for MDX.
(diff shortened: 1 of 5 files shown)


.changeset/calm-emus-raise.md
```diff
@@ -0,0 +1,6 @@
+---
+'astro': patch
```
matthewp, line 2: This should be a major.

.changeset/calm-emus-raise.md
```diff
@@ -0,0 +1,6 @@
...
+'@astrojs/mdx': patch
```
matthewp, line 3: This should be a minor (I think)

.changeset/calm-emus-raise.md
bholmesdev: Hm, I'm confused why this would be a major? These APIs were removed in 1.0 and are undocumented, and only serve as warning logs today. Happy to bump but curious for future reference!

matthewp (approved):

Issue #5860, opened 2023-01-14
[Prerender] Pages with dynamic routes in the same folder don't work in @2.0.0-beta.2


juanmiguelguerrero:
### What version of `astro` are you using?

2.0.0-beta.2

### Are you using an SSR adapter? If so, which one?

Netlify

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

If we have more than one page with dinamic routes in the same folder we get an error.

I have created 2 pages with getStaticPatch [cat].astro and [dog].astro (and prerender = true).

All dog routes doesn't works... 

And if you input any route then cat page will appear... never enter in the dog page.

Demo: https://stackblitz.com/github/juanmiguelguerrero/astro-prerender-staticpaths-beta2

If you take out the Netlify adapter configuration... everything works fine.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/juanmiguelguerrero/astro-prerender-staticpaths-beta2

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
This looks like a problem with the way the Netlify adapter is generating the `_redirects` file. We should be generating a single redirect for each `.html` file that is prerendered rather than each route that is prerendered.

bluwy:
It looks like the `_redirects` is:

```
/    /.netlify/functions/entry    200
/*   /cat3/index.html    200
/*   /dog3/index.html    200
```

I don't quite understand what's the desired output though, but I feel like the two latter redirects should be removed?

---

Also it doesn't seem like this is a build-only issue. During dev, the dogs page will always use the cats Astro files to render since Astro's routing is first-match (IIRC) [via the filename-as-regex](https://github.com/withastro/astro/blob/47be52c97a052c5b087b4697ef823611f9ec6c00/packages/astro/src/core/routing/match.ts#L4-L7), and the `geStaticPaths` is only considered later 🤔

PR #5904, opened 2023-01-19
Support prerender in Netlify redirects

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/5860
- A little bit of a refactor of the logic that generates the _redirects file to be a little more structured.
- Rather than use the build-generated "distURL", which is always going to be wrong for dynamic routes, builds it itself and uses Netlify's placeholder syntax. For spread routes it uses the * syntax.
- Prettifys it a bit so that each part, the input, target, and status all line up.

## Testing

- Test added for the same scenario that the bug describes.

## Docs

N/A, bug fix.

### DIFF
--- a/packages/integrations/netlify/src/shared.ts
+++ b/packages/integrations/netlify/src/shared.ts
@@ -5,6 +5,7 @@ type RedirectDefinition = {
 	dynamic: boolean;
 	input: string;
 	target: string;
+	weight: 0 | 1;
 	status: 200 | 404;
 };
 
@@ -26,15 +27,17 @@ export async function createRedirects(
 				definitions.push({
 					dynamic: false,
 					input: route.pathname,
-					target: route.distURL.toString().replace(dir.toString(), ''),
-					status: 200
+					target: prependForwardSlash(route.distURL.toString().replace(dir.toString(), '')),
+					status: 200,
+					weight: 1
 				});
 			} else {
 				definitions.push({
 					dynamic: false,
 					input: route.pathname,
 					target: `/.netlify/${kind}/${entryFile}`,
-					status: 200
+					status: 200,
+					weight: 1,
 				});
 
 				if (route.route === '/404') {
@@ -42,7 +45,8 @@ export async function createRedirects(
 						dynamic: true,
 						input: '/*',
 						target: `/.netlify/${kind}/${entryFile}`,
-						status: 404
+						status: 404,
+						weight: 0
 					});
 				}
 			}
@@ -67,14 +71,16 @@ export async function createRedirects(
 					dynamic: true,
 					input: pattern,
 					target,
-					status: 200
+					status: 200,
+					weight: 1
 				});
 			} else {
 				definitions.push({
 					dynamic: true,
 					input: pattern,
 					target: `/.netlify/${kind}/${entryFile}`,
-					status: 200
+					status: 200,
+					weight: 1
 				});
 			}
 		}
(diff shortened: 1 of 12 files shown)


bluwy (commented):
Ah I see what needs to be fixed now. I tested this locally, I wonder if the `functions/fixtures/404/dist/_redirects` is correct now though. It is:

```
/*      /.netlify/functions/entry    404
/       /.netlify/functions/entry    200
/404    /.netlify/functions/entry    200
```

where the `/*` is the first and I think that causes a 404 always? Before this PR it's:

```

  /    /.netlify/functions/entry    200
  /404    /.netlify/functions/entry    200
  /*    /.netlify/functions/entry    404
```

I think we need to sort to like before for this case.

matthewp:
Ah yeah, you're right. I think I was actually wrong about needing to sort dynamic routes first. What you really want is to sort any exact route first. It's the wildcard/spread routes that need to come last. I think I'll add a `specificity` integer and sort that way. I'll add a test for this too.

packages/integrations/netlify/test/functions/dynamic-route.test.js
```diff
@@ -21,6 +21,13 @@ describe('Dynamic pages', () => {
...
 	it('Dynamic pages are included in the redirects file', async () => {
 		const redir = await fixture.readFile('/_redirects');
-		expect(redir).to.match(/\/products\/\*/);
+		expect(redir).to.match(/\/products\/:id/);
+	});
+
+	it('Prerendered routes are also included using placeholder syntax', async () => {
+		const redir = await fixture.readFile('/_redirects');
+		expect(redir).to.include('/pets/:cat       /pets/:cat/index.html        200');
+		expect(redir).to.include('/pets/:dog       /pets/:dog/index.html        200');
```
natemoo-re, line 30: Will these conflict? I expect that they might and only `/pets/:cat` would be matched.

Since all of these are generating actual files we might need to track _any_ files that the dynamic route has generated, then add redirects for the exact route?

```
expect(redir).to.include('/pets/cat1       /pets/cat1/index.html        200');
expect(redir).to.include('/pets/cat2       /pets/cat2/index.html        200');
expect(redir).to.include('/pets/cat3       /pets/cat3/index.html        200');
expect(redir).to.include('/pets/dog1       /pets/dog1/index.html        200');
expect(redir).to.include('/pets/dog2       /pets/dog2/index.html        200');
expect(redir).to.include('/pets/dog3       /pets/dog3/index.html        200');
```

matthewp:
Ok, I think I got it now @bluwy here: https://github.com/withastro/astro/pull/5904/commits/00224f314344a891f94a0e359bd5b86b7b1f3077

The changes are:

- Astro already sorts the route array so we don't need to resort what is produced for `_redirects`.
- *But*, the wildcard 404 route is the exception. Since this is a new route that we create we need to sort it. So this change gives it a lower weight so its always sorted last. This is what you want, any defined routes are hit and then there's a final wildcard 404.

packages/integrations/netlify/src/shared.ts
```diff
@@ -10,37 +19,120 @@ export async function createRedirects(
...
+		if(b.target.length > minTargetLength) {
+			minTargetLength = b.target.length;
+		}
+
+		// Sort dynamic routes on top
+		if(a.weight > b.weight) {
+			return -1;
+		} else {
+			return 1;
+		}
```
bluwy, line 121: I don't think this matters from some quick tests, but maybe we can make this
```suggestion
		return b.weight - a.weight;
```

so that 2 equal weights return 0 so it retains the order.

bluwy (approved):

PR #5908, opened 2023-01-19
fix(@astrojs/tailwind): manually load postcss config file

MoustaphaDev:
## Changes

Fixes a bug spotted from the Astro discord: https://discord.com/channels/830184174198718474/1065463699508297758

The `@astrojs/tailwind` integration internally inlines the tailwind and autoprefixer plugins in the vite config, which causes vite to ignore loading postcss configuration files.
This change aims to bring back the default behavior of vite, which is to load postcss config files *only* when there's no inline postcss configuration in the vite user config.

## Testing

Manually tested

## Docs

No docs needed, this is a bug fix


matthewp:
Thanks! Can you add a changeset?

packages/integrations/tailwind/src/index.ts
```diff
@@ -1,12 +1,15 @@
...
 import type { AstroIntegration } from 'astro';
+import type { CSSOptions, UserConfig } from 'vite';
 import autoprefixerPlugin from 'autoprefixer';
 import fs from 'fs/promises';
 import path from 'path';
 import tailwindPlugin, { Config as TailwindConfig } from 'tailwindcss';
 import resolveConfig from 'tailwindcss/resolveConfig.js';
 import { fileURLToPath } from 'url';
 
+type AddWatchFile = (path: string | URL) => void;
```
bluwy, line 11: Seems like we can remove this for now since it's unused.

pnpm-lock.yaml
```diff
@@ -74,7 +74,7 @@ importers:
       '@astrojs/sitemap': ^1.0.0
       astro: ^2.0.0-beta.3
     dependencies:
-      '@astrojs/mdx': link:../../packages/integrations/mdx
+      '@astrojs/mdx': 1.0.0-beta.2_astro@packages+astro
```
bluwy, line 77: I think this incorrect `@astrojs/mdx` changes should be fixed once you merged up `main` into this branch and run `pnpm i` again. Otherwise maybe it's fine to leave it as is as CI would update the lockfile again (not quite sure).

bluwy (commented):
Looks great! with a couple of nits. Thanks for handling this unintentional regression 🙏

MoustaphaDev:
Made all requested changes

bluwy (approved):

PR #5910, opened 2023-01-19
Ignore `types.generated.d.ts` in tests

jasikpark:
## Changes

- Adds some generated type files to the gitignore. Normally these should be checked in I believe? But for these tests I don't think we want them checked in.

## Testing

n/a

## Docs

n/a


jasikpark:
these files are no longer generated

Issue #5900, opened 2023-01-19
SSR Streaming is disabled on Cloudflare


AngusMorton:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### Describe the Bug

Astro doesn't support streaming when deployed to Cloudflare Pages. It was disabled in https://github.com/withastro/astro/pull/3777.

ReadableStream is implemented in the Cloudflare Worker runtime now as of the end of last year (https://blog.cloudflare.com/standards-compliant-workers-api/), so, we should be able to revert those changes and enable streaming on Cloudflare again.

### Link to Minimal Reproducible Example

N/A

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #3777, opened 2022-06-30
Adding an option to disable HTTP streaming

tony-sull:
Closes #3764 

## Changes

This adds an option to Astro's `app` server to disable HTTP streaming (enabled by default)

Cloudflare Page Functions don't fully support `ReadableStream` yet - it is defined so a polyfill wouldn't really do the trick, but the server will throw an error when the `ReadableStream` constructor is called.  This change updates the Cloudflare adapter to always disable streaming

## Testing

- Added tests for disabling streaming in dev and production
- Manually tested the fix deploying to Cloudflare Pages

## Docs

Doc comments added for new config

Issue #3764, opened 2022-06-29
🐛 BUG: Streaming (1.0.0-beta.55) breaks Cloudflare Pages


nrgnrg:
### What version of `astro` are you using?

1.0.0-beta.55

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

pnpm

### What operating system are you using?

mac

### Describe the Bug

I'm super excited that streaming has landed in Astro, such a great feature!

Unfortunately Cloudflare Pages are currently stuck on a fairly old version of the workers runtime that doesn't support streams. There is as of right now no way to set the compatibility date or the flag to enable streams. I've asked about this again on the Cloudflare discord but the last update was "it's coming soon", which isn't a concrete timeframe.

What this means is that the Cloudflare adaptor, for now anyway, will only work with versions of Astro pre `1.0.0-beta.55`.

I've tried the `web-streams polyfill` and buffering the response to a string in the Cloudflare adaptor but I think the issue lies in the fact that the Response constructor doesn't know what to do when passed a stream and just outputs `[object ReadableStream]`.

I don't think this is a bug or an issue with Astro, more Cloudflare Pages itself, but I've raised the issue here because I think it should be documented on the Cloudflare adaptor which I'm happy to raise a PR to do, since it's now somewhat broken. Or on the off chance there is a way to workaround the streams changes that went into Astro core?

### Link to Minimal Reproducible Example

n/a

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
**Notes from Triage Call**

- Cloudflare adapter might need to use the `@astrojs/webapi` polyfill. If that works, we're good!
- If not, we might need to make our streaming support detection is properly working.
- As a final option, we may need to build-in a flag for core to toggle streaming on/off.

matthewp:
Can the adapter not just buffer the response and create a new one with the final HTML?

PR #5914, opened 2023-01-20
Re-enable streaming on Cloudflare Pages

AngusMorton:
Originally, we disabled streaming on Cloudflare Pages because the Cloudflare Worker runtime didn't support ReadableStream (https://github.com/withastro/astro/issues/3764, https://github.com/withastro/astro/pull/3777). At the end of last year, the Worker Runtime was updated and now supports ReadableStream ([Standards Compliant Workers API](https://blog.cloudflare.com/standards-compliant-workers-api/)). So we can now remove the workarounds we had in place for Cloudflare. 

Closes https://github.com/withastro/astro/issues/5900.

## Changes

- Re-enables streaming in the Cloudflare Pages integration.

## Testing

I haven't added any tests - I couldn't see any examples of tests of streaming in other integrations.

I have deployed an example to Cloudflare Pages, available here: https://astro-streaming-cf.pages.dev/

The example is the basics example, except that the `<head>` meta tags have an await in them for ~2 seconds. You can see in the waterfall that we stream the first part and then wait for 2 seconds:

image

## Docs

This brings the Cloudflare Pages integration back to the expected and documented streaming behaviour when SSR is enabled.


natemoo-re (approved):
This is great to see, thanks for contributing!

PR #5917, opened 2023-01-20
Remove duplicate CSS in dev

### LINKED ISSUE #5817: CSS Sourcemap Duplication of References in Dev

1.9.1

None

npm

Mac

### Describe the Bug

This is part of an old closed issue that needs addressing since Astro 1.0.0-beta42. 
Original issue: https://github.com/withastro/astro/issues/3647

I thought it might work itself out with future Astro versions. Even with Astro 2 beta which utilizes Vite v4, the issue persists.

With CSS sourcemaps enabled, there is duplicate CSS being referenced from both the css file and page file. This can make debugging css extremely tedious. I thought maybe it was a Vite bug. After using Vite as standalone compiler and Sveltekit 1.0, that doesn't seem to be the case as there is no CSS sourcemap duplication with these libraries. 

```js
// astro.config.mjs
vite: {
    css: {
      devSourcemap: true,
    },
  },
```
Screenshot 2023-01-10 at 11 45 47 AM

https://stackblitz.com/edit/github-g7v2dq

### Participation

natemoo-re:
## Changes

- Closes #5817
- Better handling for CSS in dev
- Includes the Vite ID of SSR'd styles so they can be removed once Vite takes over on the client

## Testing

- E2E test added for default handling
- E2E test added to confirm this also works when `vite.css.devSourcemap` is enabled

## Docs

Bug fix only

### DIFF
--- a/packages/astro/src/core/render/dev/index.ts
+++ b/packages/astro/src/core/render/dev/index.ts
@@ -10,7 +10,7 @@ import { PAGE_SCRIPT_ID } from '../../../vite-plugin-scripts/index.js';
 import { enhanceViteSSRError } from '../../errors/dev/index.js';
 import { AggregateError, CSSError, MarkdownError } from '../../errors/index.js';
 import type { ModuleLoader } from '../../module-loader/index';
-import { isPage, resolveIdToUrl } from '../../util.js';
+import { isPage, resolveIdToUrl, viteID } from '../../util.js';
 import { createRenderContext, renderPage as coreRenderPage } from '../index.js';
 import { filterFoundRenderers, loadRenderer } from '../renderer.js';
 import { getStylesForURL } from './css.js';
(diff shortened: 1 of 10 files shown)


packages/astro/src/core/render/dev/index.ts
```diff
@@ -133,7 +133,11 @@ async function getScriptsAndStyles({ env, filePath }: GetScriptsAndStylesParams)
 		});
 		// But we still want to inject the styles to avoid FOUC
 		styles.add({
-			props: {},
+			props: {
+				type: 'text/css',
```
matthewp, line 137: is this because vite needs the attributes to match? if not, `type` is not need on script or style, `<style>` is always text/css.

natemoo-re: No, `data-astro-dev-id` is the only _necessary_ change here. I just added this for consistency with the styles Vite injects.

matthewp (approved):

PR #5920, opened 2023-01-20
Improve error message for third-party template 404s

delucis:
## Changes

- Adds a helpful error message when a third-party template passed to `create astro` is not found.

- Inspired by https://github.com/withastro/docs/issues/2372

- Currently if a user passes `--template username/repo` and `giget` fails to load that repo, the command fails and says "does not exist!":
    image

- A common cause is that the template repo does not have a `main` branch (which `giget` assumes by default and we have no way of guessing without calling the GitHub API), so the updated error message adds some pointers

    > Could not find template **username/repo**!
    > Are you sure this GitHub repo exists? This command uses the **main** branch by default.
    > If the repo doesn't have a main branch, specify a custom branch name:
    > **username/repo#branch-name**

## Testing

Don’t think this needs tests as it’s just a log message?

## Docs

Arguably entirely docs!

### DIFF
--- a/.changeset/large-hotels-give.md
+++ b/.changeset/large-hotels-give.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Improve error message for third-party template 404s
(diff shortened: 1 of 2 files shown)


sarah11918 (approved):
Arguably, entirely approved text by Docs!

Princesseuh (approved):
Looks good to me! Love good error messages 😍

PR #5921, opened 2023-01-20
Content Collections: remove `data` and `body` from custom `slug()`

bholmesdev:
## Changes

- Removes `data` and `body` from the custom `slug()` configuration option

### Why remove this?

Allowing access to the parsed Zod schema introduces a bottleneck when looking up entries by slug. Today, this function is slower than we'd like it to be, since it needs to look through all items in a collection to give you the right one. By allowing access to the data object, it is more difficult (though not impossible) to make this function more performant in a non-breaking way.

We recognize the usefulness of looking up entries by custom `data` attributes. However, for the most performant and futureproof story in Astro 2.0, removing this to revisit in a future minor feels like the best path.

## Testing

Update tests expecting `data`

## Docs

TODO: docs updates

### DIFF
--- a/.changeset/happy-peaches-divide.md
+++ b/.changeset/happy-peaches-divide.md
@@ -1,0 +1,19 @@
+---
+'astro': major
+---
+
+Remove `data` and `body` from the custom collection `slug()` mapper.
+
+### Migration
+
+If you computed slugs based on frontmatter data, we suggest reading this property separately where your content is used:
+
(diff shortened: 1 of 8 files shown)


FredKSchott (approved):
LGTM once you fix tests

matthewp (approved):

bholmesdev:
Closing in favor of #5941

PR #5932, opened 2023-01-22
fix(image): allow usage of image from any directory

rasendubi:
Currently, @astrojs/image allows *importing* images from srcDir only. Importing images from outside srcDir fails miserably *in dev mode* and produces incorrect src.

This happens because `path.relative(fileURLToPath(config.srcDir), id)` resolves to "../something" and when joined with '/@astroimage' cancels it (`join('/@astroimage', '../../something')` => `'/something'`).

Rework /@astroimage URL scheme to be similar to "/@fs/" scheme—always export absolute path to the target file. This allows images to be places anywhere on the filesystem—even outside project root.

## Testing

Added tests for `@astrojs/image`.

## Docs

This lifts the restriction of only placing images in src or public directory. Images can be imported from anywhere—even outside project root.


rishi-raj-jain:
This is great!

packages/integrations/image/src/vite-plugin-astro-image.ts
```diff
@@ -65,22 +64,17 @@ export function createPlugin(config: AstroConfig, options: Required<IntegrationO
...
 		},
 		configureServer(server) {
 			server.middlewares.use(async (req, res, next) => {
 				if (req.url?.startsWith('/@astroimage/')) {
-					const [, id] = req.url.split('/@astroimage/');
+					// Reconstructing URL to get rid of query parameters in path
+					const url = new URL(req.url.slice('/@astroimage'.length), 'file:');
 
-					const url = new URL(id, config.srcDir);
 					const file = await fs.readFile(url);
```
rasendubi, line 78: If you are reviewing a PR, this is the core and the only meaningful change—the rest is just tests

rasendubi:
What's my next step to get this PR moving?

cc @matthewp @natemoo-re (btw, congrats on 2.0 launch!)

natemoo-re:
Thanks for the PR @rasendubi! I'm going to pull in @Princesseuh who is doing some cleanup on the image integration. This seems like a reasonable change!

Princesseuh (approved):

Issue #5934, opened 2023-01-22
img tag src with @astrojs/image doesn't work in certain circumstances


gvkhna:
### What version of `astro` are you using?

1.9.2

### Are you using an SSR adapter? If so, which one?

Deno? Or SSG? I believe it's not affected by adapter.

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

Simply adding astrojs/image to a project doesn't correctly render <img tags in certain circumstances.

For example:

```
import {MyImage} from '@images/myimage.png'

<img src={MyImage} alt="" />
```

In output for dev/build is turned into 

```
<img src="[object Object]" alt="">
```

It does seem like static img paths are not affected (`<img src="myimage.jpg` ..), this has something to do with resolving the import url and the astrojs/image plugin possibly in vite-plugin-astro-image.js

After extensive tracing I was able to narrow down the scenario in more detail that may help with a fix.

https://github.com/withastro/astro/blob/02ba7c2250a950593961d833201e46adb3e0873c/packages/astro/src/vite-plugin-astro/compile.ts#L37

Here you can add `console.log(esbuildResult)` and you'll see the following:

```
addAttribute as $$addAttribute

<img${$$addAttribute(MyImage, "src")} alt="img
```

Tracing this through to here:

https://github.com/withastro/astro/blob/37cb2fc02a03754d454b243579bc55e55cf72904/packages/astro/src/runtime/server/render/util.ts#L52

Where `key` is `src` and `value` is:

```
{
  src: '/assets/img-02.9ece27da.png',
  width: 95,
  height: 100,
  format: 'png'
}
```

I believe the issue here is that `value` is an object, through my tracing most other values are strings. `value` is generated from the `metadata` function in astrojs/image. As for how it gets there I'm not sure.

This Object is being sent from somewhere as part of astrojs/image I believe. But it's unclear exactly where at this point.

Here's a sample stack trace:

```
  at addAttribute (file:///home/~/dist/entry.mjs?time=1674364337555:1470:13)
    at AstroComponentInstance.factory (file:///~/dist/entry.mjs?time=1674364337555:3530:23)
    at AstroComponentInstance.init (file:///~dist/entry.mjs?time=1674364337555:994:29)
    at AstroComponentInstance.render (file:///~dist/entry.mjs?time=1674364337555:999:18)
    at render.next (<anonymous>)
    at renderChild (file:///~dist/entry.mjs?time=1674364337555:1058:18)
[...]

PR #5950, opened 2023-01-24
Utilize query params for image metadata

gvkhna:
First PR! Continuation of Issue: https://github.com/withastro/astro/issues/5934

## Changes

- Pass metadata of image file to component via query params
- Update getImage/getPicture to interpret metadata from the url search params if it exists
- Minimal example to get this working
- Removes typescript types for *.png,*.jpg that breaks typescript checking file exists at path
- Fixes `` tags even with astrojs/image loaded

**WIP at this point.** 

### NOTE

`getImage` and `getPicture` probably need to be cleaned up. With sufficient buy-in they could be moved to soley accessing metadata from query params. But right now this minimal set of changes acts as a bridge to trying to accept the old format and new format of data. Which makes these much more messy and difficult to reason about. 

## Synopsis

Currently 
```jsx
import {myImage} from './img.png'
// myImage = {src: 'images/img.png',width: 100, height: 100, format: 'png'}
```
Change to 
```jsx
import {myImage} from './img.png'
// myImage = 'images/img.png?w=100&h=100&f=png'
```

Params as part of a url search query are easy and versatile to parse in any context, for convenience `getImage` etc can be used.

## Testing

Manually tested 
- ``
- `<Image src={myImage} />`
- `<Picture src={myImage} alt="" sizes="((min-width: 50em) and (max-width: 60em)) 50em,((min-width: 30em) and (max-width: 50em)) 30em,(max-width: 30em) 20em" widths={[400, 800, 1200]} />`

## Docs

More than likely docs will have to be updated if docs suggest usage of the following code:

```jsx
import { myImage} from './myimage.png'
<img src={myImage.src} width={myImage.width} />
```

I don't see this in the docs, but I do see the following which should be the suggested alternative.

```jsx
import { getImage } from '@astrojs/image'
import { myImage } from './myimage.png'
const img = getImage(myImage)
console.log(img.src, img.width)
```


gvkhna:
The extensive set of tests is amazing! I'll continue work on fixing this, I did not thoroughly test these changes on SSR.

Princesseuh:
Closing for inactivity. Additionally, our new plan for images include making the `@astrojs/image` shape the default, which should reduce the confusion that lead to this PR being created

Thank you for contributing!

Issue #5902, opened 2023-01-19
Astro <Code> component is not able to use a custom theme anymore


Ivluengo:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

none

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### Describe the Bug

Hi! I have been using your `<Code>` component so far and apart from the issue #5595 that is different from this one, I have found a problem when I upgrade Astro. Since the upgrade I can't use my custom theme anymore. I have been trying all versions and the last that worked was 1.6.8. Since that onwards is not working. It doesn't show any error, but you can see the styles are not applied anymore.

I'm sending you some captures

## My code:
![code](https://user-images.githubusercontent.com/10440026/213409539-0ef2bac3-77f1-4060-a99a-c286d17b0010.png)

## The result in version 1.6.8:
![before](https://user-images.githubusercontent.com/10440026/213409669-68b76747-ff3f-4c1d-878a-6be526dea439.jpg)

## The result now:
![after](https://user-images.githubusercontent.com/10440026/213409702-54afa92b-dc67-4d86-bdfd-70a817fbc4e3.jpg)

Thanks a lot!

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-7dhldu?file=components/MyCode.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

delucis:
I think https://github.com/withastro/astro/commit/9f80a4046ff90e379be68bf03c3c4dd4dd5d6d87 likely caused this.

As far as I can tell we always expect `theme` to be a string here, whereas previously it could also be a custom theme object:

https://github.com/withastro/astro/blob/ce5c5dbd46afbe738b03600758bf5c35113de522/packages/astro/components/Shiki.js#L20-L24

Ivluengo:
Hi!! So do you think it will never be a custom object anymore? And if is that so, how can I pass my custom theme then?

Thanks!

delucis:
The bug is that we used to support `<Code theme={{ /* options */ }}>`

But the commit I linked assumes it only needs to handle `<Code theme="theme-name">` and so an object of options is ignored.

wulinsheng123:
I got it. I will fix it this week

Ivluengo:
You are amazing guys! Thanks for all your work! You have such an amazing framework so far!

💪

wulinsheng123:
> 

I think so

[...]

PR #5952, opened 2023-01-24
fix: component is not able to use a custom theme anymore

wulinsheng123:
## Changes

fix: #5902


natemoo-re (approved):
Code looks good to me, thanks for the fix!

PR #5962, opened 2023-01-24
Fix Content Collections not loading config file when there are spaces in the folder tree

### LINKED ISSUE #5961: Content Collections can't load `config.ts` file when there are spaces in the folder tree

2.0.0

None

pnpm

Ubuntu 22.04

### Describe the Bug

When there are spaces in the folder tree leading to the config file loaded by the content collections API, loading the config file fails

https://stackblitz.com/edit/github-moeeyn?file=package.json&on=stackblitz

### Participation

MoustaphaDev:
## Changes

Fix #5961 

## Testing

TODO

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -227,7 +227,9 @@ export async function loadContentConfig({
 		return undefined;
 	}
 	try {
-		unparsedConfig = await tempConfigServer.ssrLoadModule(contentPaths.config.pathname);
+		unparsedConfig = await tempConfigServer.ssrLoadModule(
+			decodeURIComponent(contentPaths.config.pathname)
+		);
 	} catch (e) {
 		throw e;
 	} finally {
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -227,9 +227,8 @@ export async function loadContentConfig({
 		return undefined;
 	}
 	try {
-		unparsedConfig = await tempConfigServer.ssrLoadModule(
-			fileURLToPath(contentPaths.config.pathname)
-		);
+		const configPathname = decodeURIComponent(contentPaths.config.pathname);
+		unparsedConfig = await tempConfigServer.ssrLoadModule(configPathname);
 	} catch (e) {
 		throw e;
 	} finally {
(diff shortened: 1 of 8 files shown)


matthewp:
Can you add a test? Also I think it should be done like its done here: https://github.com/withastro/astro/blob/main/packages/astro/src/core/render/dev/index.ts#L60

using `fileURLToPath()`

MoustaphaDev:
> Also I think it should be done like its done here: https://github.com/withastro/astro/blob/main/packages/astro/src/core/render/dev/index.ts#L60
> 
> using `fileURLToPath()`

Ah ok, testing that

MoustaphaDev:
@matthewp, some debugging and I think `decodeURIComponent` would do it here, in the console it looks like the logged file is indeed a path not a file url

packages/astro/src/content/utils.ts
```diff
@@ -228,7 +228,7 @@ export async function loadContentConfig({
 	}
 	try {
 		unparsedConfig = await tempConfigServer.ssrLoadModule(
-			decodeURIComponent(contentPaths.config.pathname)
+			fileURLToPath(contentPaths.config.pathname)
```
MoustaphaDev, line 231: Meant to do 
```js
new URL(fileURLToPath(contentPaths.config)).pathname
```
But that results in the majority of tests to fail. I'll revert back to using `decodeURIComponent`

matthewp: Are tests failing with just `fileURLToPath(contentPaths.config.pathname)`? (without the URL wrapper)?

MoustaphaDev: Yes, in both cases

matthewp: Oh sorry, I would expect it to be `fileURLToPath(contentPaths.config)`. config is a file URL. That converts it to a regular path which Vite can take.

matthewp: Essentially you almost never need/want to use `.pathname` anywhere that takes a file path.

MoustaphaDev: Ah thanks for your help, that did it!

matthewp (approved):
looks great, thanks!

bholmesdev (approved):
Nice fix, _and_ with tests! Love the `config`dence 👏

matthewp:
The failing test is the flaky one, will restart.

MoustaphaDev:
> The failing test is the flaky one, will restart.

Oops I thought the same, merged anyway as it's been ignore a lot of time
I can revert

PR #5970, opened 2023-01-24
[Content collections] Support type guards on `getCollection()` filter

bholmesdev:
## Changes

- Add type guard support for `getCollection()` filters

## Testing

Manual testing

## Docs

N/A

### DIFF
--- a/.changeset/strong-hotels-sort.md
+++ b/.changeset/strong-hotels-sort.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Add type guard support to filters on `getCollection()`
--- a/packages/astro/src/content/template/types.d.ts
+++ b/packages/astro/src/content/template/types.d.ts
@@ -44,10 +44,10 @@ declare module 'astro:content' {
 	): E extends ValidEntrySlug<C>
 		? Promise<CollectionEntry<C>>
 		: Promise<CollectionEntry<C> | undefined>;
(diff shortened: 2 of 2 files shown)


natemoo-re (approved):

andersk:
This broke the use of plain boolean filters.

```
/home/anders/zulip/zulip-blog/src/pages/author/[...author].astro:22:44 Error: Argument of type '(post: CollectionEntry<"posts">) => boolean' is not assignable to parameter of type '(entry: CollectionEntry<"posts">) => entry is CollectionEntry<"posts">'.
  Signature '(post: CollectionEntry<"posts">): boolean' must be a type predicate.
22  const posts = await getCollection("posts", (post) => post.data.author === author);
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It also doesn’t conceptually make sense to require a type guard. A type guard `(entry) => entry is E` is expected to return `true` when given an `E`, *and also* return `false` when not given an `E` ([documentation](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)), so it can only be used to filter for *all* members of `E`.

This is why TypeScript arrays have [two `filter` overloads](https://github.com/microsoft/TypeScript/blob/v4.9.4/lib/lib.es5.d.ts#L1442-L1453).

PR #5971, opened 2023-01-25
change @astrojs/solid to @astrojs/solid-js

JLarky:
## Changes

- just a typo, because @astrojs/solid package doesn't exist

before:
image

### DIFF
--- a/.changeset/serious-turtles-explode.md
+++ b/.changeset/serious-turtles-explode.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/serious-turtles-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
bluwy, line 2: Not sure if this is right. I think it should be
```suggestion
'astro': patch
```

natemoo-re (approved):

Issue #5967, opened 2023-01-24
Astro 2.0 Node Version Check Issue


SteveEdson:
### What version of `astro` are you using?

2.0.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac / Linux

### Describe the Bug

Getting the following errors:

## Locally
```
Node.js v16.13.1 is not supported by Astro!
Please upgrade Node.js to a supported version: ">=16.12.0"
```

## On Cloudflare pages
```
Node.js v16.19.0 is not supported by Astro!
Please upgrade Node.js to a supported version: ">=16.12.0"
```

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Pinging @bluwy in case he has any ideas. I haven't been able to reproduce this personally.

ZackarySantana:
https://github.com/withastro/astro/blob/a8d3e79246605d252dcddad159e358e2d79bd624/packages/astro/astro.js#L26

It's possible 'semver' isn't installed/properly and is throwing a module not found, the catch does not handle this and always says it is a version of node that doesn't work, that's the only way I was able to reproduce this

arkinmodi:
I also experienced issue this when trying to upgrade to Astro 2.0.

```sh
Node.js v16.18.0 is not supported by Astro!
Please upgrade Node.js to a supported version: ">=16.12.0"
```

I was able to reproduce it with this `package.json`.

```json
{
  "name": "@example/basics",
  "type": "module",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "dev": "astro dev",
    "start": "astro dev",
    "build": "astro build",
    "preview": "astro preview",
    "astro": "astro"
  },
  "dependencies": {
    "@commitlint/cli": "^16.3.0",
    "astro": "^2.0.0"
  }
}
```

It might be some sort of dependency conflict? Updating `@commitlint/cli` to the latest fixes the issue.

netanel-haber:
> I also experienced issue this when trying to upgrade to Astro 2.0.
> 
> ```shell
> Node.js v16.18.0 is not supported by Astro!
> Please upgrade Node.js to a supported version: ">=16.12.0"
> ```
> 
> I was able to reproduce it with this `package.json`.
> 
> ```json
> {
>   "name": "@example/basics",
>   "type": "module",
>   "version": "0.0.1",
[...]

PR #5972, opened 2023-01-25
Upgrade semver dependency

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5967

Upgrade semver to bring in the fix from https://github.com/npm/node-semver/pull/383, to allow `await import` to work properly.

## Testing

Tested manually.

## Docs

n/a. bug fix.


matthewp:
Nice research!

Issue #5912, opened 2023-01-19
Code component breaks build when using PNPM


sean-au:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

pnpm v7.2.0 node.js 19.2.0

### What operating system are you using?

Windows

### Describe the Bug

After running pnpm astro create@latest

adding 

`import { Code } from 'astro/components';`

and 

<Code code={`const foo = 'bar';`} lang="js" />  (both lines taken from the docs)

to pages/index.astro

On running `pnpm astro build` (or `npm astro build`) i get the following error

```
PS C:\github\astro-bootstrap\build-test> pnpm astro build

> @example/basics@0.0.1 astro C:\github\astro-bootstrap\build-test
> astro "build"

10:25:09 pm [build] output target: static
10:25:09 pm [build] Collecting build info...
10:25:09 pm [build] Completed in 13ms.
10:25:09 pm [build] Building static entrypoints...
10:25:15 pm [build] Completed in 5.80s.

 generating static routes 
 error   Cannot find package 'shiki' imported from C:\github\astro-bootstrap\build-test\dist\entry.mjs
  Did you mean to import shiki@0.11.1/node_modules/shiki/dist/index.js?
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'shiki' imported from C:\github\astro-bootstrap\build-test\dist\entry.mjs
Did you mean to import shiki@0.11.1/node_modules/shiki/dist/index.js?
    at new NodeError (node:internal/errors:393:5)
    at packageResolve (node:internal/modules/esm/resolve:823:9)
    at moduleResolve (node:internal/modules/esm/resolve:872:20)
    at defaultResolve (node:internal/modules/esm/resolve:1087:11)
    at nextResolve (node:internal/modules/esm/loader:161:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:831:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:413:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:77:40)
    at link (node:internal/modules/esm/module_job:76:36)
 ELIFECYCLE  Command failed with exit code 1.
```

Stackblitz link below works as I assume its using npm.

Here is a link to the same minimal example with PNPM 
https://github.com/powerfulweb-dev/build-test

### Link to Minimal Reproducible Example

[...]

PR #5973, opened 2023-01-25
[astro/component] 

Lofty-Brambles:
## Changes
This adds an inline attribute to render code in the `<Code />` component as inline code without the pre tag.
This PR is related to the issue: #5912 

## Testing
No tests were present for the `<Code />` component.

## Docs
Is this, along with the other attributes necessary to be documented? The current docs on it is pretty barebones. I'm willing to add a PR, if necessary.

### DIFF
--- a/.changeset/orange-bottles-attack.md
+++ b/.changeset/orange-bottles-attack.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Adds an inline attribute to render code in the <Code /> component as inline code without the pre tag.
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -1,5 +1,6 @@
 ---
 import type * as shiki from 'shiki';
+import { renderToHtml } from 'shiki';
 import { getHighlighter } from './Shiki.js';
 
 export interface Props {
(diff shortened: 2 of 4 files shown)


matthewp:
Looks like your branch has some conflicts. I would fix those and rebase with `main` which I think will take care of your build failing as well.

Lofty-Brambles:
That should do it? The ubuntu-latest build is failing due to a type mismatch between `string` and an union of predefined shiki themes.

Lofty-Brambles:
This issue seems to arise from https://github.com/withastro/astro/blob/5e3d538812793a952953e7bc1aa65010dd7538fa/packages/astro/src/core/config/schema.ts#L127
Which is resolving to `string | IShikiTheme` instead of `Theme | IShikiTheme`. Not sure as to how to fix this though- particularly why `BUNDLED_THEMES` is typed like `[Theme, ...Theme[]]`

matthewp:
Why would that be causing build issues? Your change doesn't touch that file?

Oh, is it just because you updated the Shiki version?

Lofty-Brambles:
Yep, the bump seems to be the reason

matthewp (approved):

matthewp:
Closing as this is stale. The changes look good! Please resubmit when you get the time.

PR #5992, opened 2023-01-26
Fix `Astro.url.protocol` when using the @astrojs/node SSR adapter with HTTPS

### LINKED ISSUE #5890: Astro.url.protocol returning http event when in https

8.19.3

Node adapter, standalone

npm

Arch Linux btw

### Describe the Bug

In the --- ... --- zone (how do you call it ?) of an Astro page, I'm trying to access the current URL origin of my page using Astro.url.origin. It returns the good domain name, but prefixed with http, not https (I started my preview server with SERVER_KEY_PATH & SERVER_CERT_PATH env variables, https works fine in the browser). Same issue for Astro.url.protocol.

https://stackblitz.com/edit/github-8d8wxa?file=src%2Fpages%2Findex.astro,package.json&on=stackblitz

### Participation

HiDeoo:
## Changes

This PR fixes #5890.

When using Astro with the `@astrojs/node` SSR adapter, `Astro.url.protocol` is always `http:` even when the server is running with HTTPS.

When creating the request used during rendering based on the Node.js request, the protocol is currently [hardcoded to `http:`](https://github.com/withastro/astro/blob/d47a9075bf13617fb5545aec50b46a76044bf85d/packages/astro/src/core/app/node.ts#L12).

This pull request fixes this issue by checking if the Node.js request is using HTTPS and updating the protocol accordingly. To do so, we check if the request socket is either an instance of `tls.TLSSocket` or if the [`X-Forwarded-Proto` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) is set to `https`.

While testing this, I also realised that when starting the server, the console is always logging a URL with the HTTP protocol, e.g. `Server listening on http://127.0.0.1:3000`. I decided to include a small change in this PR to log the proper protocol by checking if the server is an instance of `https.Server` or not. I thought this was not worth a different PR or even mentioning it in the changeset but maybe I was wrong?

## Testing

I added new `@astrojs/node` tests in a `url-protocol` suite using a fixture rendering a page with `Astro.url.protocol` and checking the rendered HTML based on the protocol of the Node.js request.

I also manually tested locally this change in a local repro using the `file:` pnpm protocol and confirmed that logging `Astro.url.protocol` correctly return `https:` when running with `SERVER_KEY_PATH=./key.pem SERVER_CERT_PATH=./cert.pem node ./dist/server/entry.mjs`

## Docs

I don't think this change needs to be documented.

### DIFF
--- a/.changeset/odd-rats-drop.md
+++ b/.changeset/odd-rats-drop.md
@@ -1,0 +1,6 @@
+---
+'astro': patch
+'@astrojs/node': patch
+---
+
+Fix `Astro.url.protocol` when using the @astrojs/node SSR adapter with HTTPS
--- a/packages/astro/src/core/app/node.ts
+++ b/packages/astro/src/core/app/node.ts
@@ -3,13 +3,18 @@ import type { SerializedSSRManifest, SSRManifest } from './types';
 
 import * as fs from 'fs';
 import { IncomingMessage } from 'http';
+import { TLSSocket } from 'tls';
 import { deserializeManifest } from './common.js';
 import { App, MatchOptions } from './index.js';
 
 const clientAddressSymbol = Symbol.for('astro.clientAddress');
 
 function createRequestFromNodeRequest(req: IncomingMessage, body?: Uint8Array): Request {
-	let url = `http://${req.headers.host}${req.url}`;
+	const protocol =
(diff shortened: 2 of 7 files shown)


matthewp (approved):
This is a fantastic PR. thank you!

Issue #5831, opened 2023-01-11
Cloudflare adapter with `prerender = true` breaks all routing


ni554n:
### What version of `astro` are you using?

1.9.1

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

After enabling the experimental `prerender` flag and adding `export const prerender = true;` in the _index.astro_ file; running `astro build` and `astro dev` works as expected.
But all route paths completely fails with **404** with Wrangler Dev as well as on the CF Pages after a successful deployment.

I think there's an issue with the generated `_worker.js` function. Routing starts working again if I delete that file or, set `prerender = false`.

Upgrading to `astro@2.0.0-beta.1` with `@astrojs/cloudflare@6.0.0-beta.0` didn't fix the issue also.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/node-3m5z2c

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
I think this is fixed by https://github.com/withastro/astro/pull/5822, going to test it once that is released.

gacek1123:
Running into the same issue with ` "astro": "2.0.0-beta.3"` and `"@astrojs/cloudflare": "6.0.0-beta.1"`.

matthewp:
Thanks for the reminder, this might be similar to https://github.com/withastro/astro/issues/5860, i'll take a look.

matthewp:
My WIP branch is here: https://github.com/withastro/astro/tree/cloudflare-prerender

just debugging so far, no fix found.

ni554n:
The fix is working great. Thank you. But I've bumped into another issue.

In one of my API routes, I have a library for generating JWT on CF that uses Web Crypto API; which is not available on Node 18.
As soon as I turn on the `prerender` flag on _index.astro_, `astro build` fails throwing `prerendering static routes.  error  SubtleCrypto not supported!` even though I have not enabled prerendering on that API route. Is this an intended behavior or a bug?

Please take a look at this example: https://stackblitz.com/edit/node-6ayeta

PR #5993, opened 2023-01-26
Support for prerendering in the Cloudflare integration

matthewp:
## Changes

- Fixes the Cloudflare integration to support prerendering

## Testing

- Test added that checks the generated `_routes.json`, verifies that prerendered route is included in the `excludes` config. This makes it so that the worker is bypassed in that route.
- Fixes https://github.com/withastro/astro/issues/5831

## Docs

N/A, bug fix


natemoo-re (approved):
Looks great!

PR #5998, opened 2023-01-26
fix: Don’t require `getCollection()` filter to be a type guard

andersk:
## Changes

Commit dabce6b8c684f851c3535f8acead06cbef6dce2a (#5970) broke the use of a plain boolean filter. Add an overload similar to TypeScript’s [`Array#filter` overload](https://github.com/microsoft/TypeScript/blob/v4.9.4/lib/lib.es5.d.ts#L1442-L1453).

https://github.com/withastro/astro/pull/5970#issuecomment-1405596846 (cc @bholmesdev)

## Testing

Tested with my company blog, which uses `const posts = await getCollection("posts", (post) => post.data.author === author)`.

## Docs

No documentation change needed.

### DIFF
--- a/packages/astro/src/content/template/types.d.ts
+++ b/packages/astro/src/content/template/types.d.ts
@@ -48,6 +48,10 @@ declare module 'astro:content' {
 		collection: C,
 		filter?: (entry: CollectionEntry<C>) => entry is E
 	): Promise<E[]>;
+	export function getCollection<C extends keyof typeof entryMap>(
+		collection: C,
+		filter?: (entry: CollectionEntry<C>) => unknown
+	): Promise<CollectionEntry<C>[]>;
 
 	type InferEntrySchema<C extends keyof typeof entryMap> = import('astro/zod').infer<
 		Required<ContentConfig['collections'][C]>['schema']
--- a/.changeset/rotten-dogs-hide.md
+++ b/.changeset/rotten-dogs-hide.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Update `getCollection()` filter to support type guards _or_ unknown values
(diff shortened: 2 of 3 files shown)


natemoo-re (approved):
Ahh nice catch, looks good to me! Sorry for the breakage.

packages/astro/src/content/template/types.d.ts
```diff
@@ -48,6 +48,10 @@ declare module 'astro:content' {
 		collection: C,
 		filter?: (entry: CollectionEntry<C>) => entry is E
 	): Promise<E[]>;
+	export function getCollection<C extends keyof typeof entryMap>(
+		collection: C,
+		filter?: (data: CollectionEntry<C>) => unknown
```
bholmesdev, line 53:
```suggestion
		filter?: (data: CollectionEntry<C>) => boolean
```
Not sure if this breaks type inferencing? Seems odd to require `unknown`.

bholmesdev (commented):
Eep! I remember testing this case but I'll trust if it broke on your end. Let me know if this nit also solves your issue.

packages/astro/src/content/template/types.d.ts
natemoo-re: I don't think it _requires_ `unknown`, it allows you to return any value rather than only `boolean`, just like a native `filter`.

bholmesdev: got it! In that case, approved ❤️

bholmesdev (approved):

packages/astro/src/content/template/types.d.ts
andersk: Right. `boolean` is a subtype of `unknown`, and native `filter` [allows](https://github.com/microsoft/TypeScript/blob/v4.9.4/lib/lib.es5.d.ts#L1442-L1453) `unknown`.

Issue #5989, opened 2023-01-26
@astro/tailwind is not adding prefixes in devmode


royeden:
### What version of `astro` are you using?

2.0.1

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

When updating to Astro 2.0 and the latest `@astro/tailwind` integration, vendor prefixes are no longer being added to classes when the dev server is running.

To specify:

```html
<h1 class="bg-clip-text text-transparent bg-gradient-to-r from-cyan-500 to-blue-500">Test</h1>
```

poduces the following `<style>` injected into astro by vite:

```css
.bg-clip-text {
  background-clip: text;
}
```

instead of the expected:

```css
.bg-clip-text {
  -webkit-background-clip: text;
  background-clip: text;
}
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hbvmrj

### Participation

- [x] I am willing to submit a pull request for this issue.

marcjulian:
I have noticed this behavioir too. I think this is because `autoprefixer` is only used during `build` and not in `dev` mode.

https://github.com/withastro/astro/blob/088f5194c55a6ec15b2eaf2cfb97f9ef45a24a33/packages/integrations/tailwind/src/index.ts#L103-L105

royeden:
It's actually documented on [this PR](https://github.com/withastro/astro/commit/a3a7fc9298e6d88abb4b7bee1e58f05fa9558cf1) on the changelog, this seems like an intentional fix, which strikes me as odd.

PR #6002, opened 2023-01-27
fix: autoprefixer on dev mode for tailwind plugin

royeden:
A recent PR (#5717) seems to have removed the autoprefixer postcss plugin for the TailwindCSS integration when it's not running in build mode. This breaks styles in development mode (#5989) because vendor prefixes aren't applied when it's running.

## Changes
- Removed `isBuild` check before running the autoprefixer plugin. It will run in every build mode.
- Added `appearance-none` class to examples to test the components.

## Testing
- Added `appearance-none` tests, as this is a vendor prefixed css value (this is tested with the -webkit prefix as playwright runs on chrome).

## Docs

This change wouldn't affect the current docs.

### DIFF
--- a/packages/integrations/tailwind/README.md
+++ b/packages/integrations/tailwind/README.md
@@ -68,7 +68,7 @@ export default defineConfig({
 
 When you install the integration, Tailwind's utility classes should be ready to go right away. Head to the [Tailwind docs](https://tailwindcss.com/docs/utility-first) to learn how to use Tailwind, and if you see a utility class you want to try, add it to any HTML element to your project!
 
-[Autoprefixer](https://github.com/postcss/autoprefixer) is also setup automatically for dev and builds so Tailwind classes will work in older browsers.
+[Autoprefixer](https://github.com/postcss/autoprefixer) is also set up automatically when working in dev mode, and for production builds, so Tailwind classes will work in older browsers.
 
 https://user-images.githubusercontent.com/4033662/169918388-8ed153b2-0ba0-4b24-b861-d6e1cc800b6c.mp4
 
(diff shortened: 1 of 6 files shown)


bluwy (approved):
We have discussed and decided to re-enable autoprefixer for now.

packages/integrations/tailwind/README.md
```diff
@@ -68,7 +68,7 @@ export default defineConfig({
 
 When you install the integration, Tailwind's utility classes should be ready to go right away. Head to the [Tailwind docs](https://tailwindcss.com/docs/utility-first) to learn how to use Tailwind, and if you see a utility class you want to try, add it to any HTML element to your project!
 
-[Autoprefixer](https://github.com/postcss/autoprefixer) is also setup automatically for production builds so Tailwind classes will work in older browsers.
+[Autoprefixer](https://github.com/postcss/autoprefixer) is also setup automatically for dev and builds so Tailwind classes will work in older browsers.
```
sarah11918, line 71:
```suggestion
[Autoprefixer](https://github.com/postcss/autoprefixer) is also set up automatically for dev and production builds so Tailwind classes will work in older browsers.
```
Fixed an existing typo!  :smile: And, I think people are used to hearing the phrase "production builds" (even :if you mean every time you build, even just locally). So just checking: did you mean to take that out? I'll let you decide what's most accurate!

sarah11918 (approved):
Comment below!

packages/integrations/tailwind/README.md
bluwy: Yeah I removed "production" since it kinda felt like it mean "dev builds" too, which isn't a thing. Trying to imply the dev server and production builds at the same time 🤔

njacob1001:
We really need this, the development mode is broken 💀

packages/integrations/tailwind/README.md
sarah11918: OK, then how about ..???

> set up automatically when working in dev mode, and for production builds

(I updated my suggestion above!)

bluwy: The Oxford comma saves again 😄 Yeah that works for me!

PR #6008, opened 2023-01-27
`@astrojs/image`: Allow passing `undefined` to transform options

### LINKED ISSUE #6001: Image integration types are wrong for an astro project w/ the strictest settings enabled.

2

none

npm

mac

### Describe the Bug

- `@astrojs/image` does not support `exactOptionalProperties` in its Props
- the error message for no `widths` passed to `Picture` is confusing
- `ts-expect-error` does not work for ignoring types problems

When using Picture, I'm passing a variable `imgPosition` that may be undefined to `position` attribute of `Picture`

```astro
 {/** @ts-expect-error @astrojs/image doesn't have the types quite right. */}
    <Picture
      widths={[
        150, 200,
        // for 2x pixel density displays
        300, 400,
      ]}
      background="white"
      sizes="150px (min-width 992px) 200px"
      src={hero.src}
      alt={hero.alt}
      class="BlogDetail_image"
      aspectRatio={1}
      position={imgPosition}
      loading=
```

image

This fails to typecheck b/c of

```js
 // Report an error when the value `undefined` is given to an optional property that doesn't specify `undefined` as a valid value.
    "exactOptionalPropertyTypes": true,
```

in `astro/tsconfigs/strictest.json`

corollary: "at least one width should be provided" does not make it obvious that the prop that expects a width is called `widths` - might be good to update that error

Reproduce:

Open the stackblitz and run `pnpm astro check`

get this output:

```
Error: Type '{ alt: string; src: string; aspectRatio: number; widths: number[]; position: undefined; }' is not assignable to type 'IntrinsicAttributes & Props' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Type '{ alt: string; src: string; aspectRatio: number; widths: number[]; position: undefined; }' is not assignable to type 'IntrinsicAttributes & PictureComponentRemoteImageProps' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
    Type '{ alt: string; src: string; aspectRatio: number; widths: number[]; position: undefined; }' is missing the following properties from type 'PictureComponentRemoteImageProps': sizes, background
16    <Picture alt="astro logo" src="https://astro.build/assets/press/full-logo-dark.png" aspectRatio={1} widths={[100]} position={undefined}/>
       ~~~~~~~
17   </body>

```

https://stackblitz.com/edit/github-nnelwr?file=src/pages/index.astro

### Participation

jasikpark:
## Changes

- This fixes #6001, allowing undefined passed as a variable in addition to not passing the property at all
- Updates TransformOptions for @astrojs/image

There may be more places in the integration where this is needed, idk.

## Docs

I don't think the docs need an update? This is just a change for users w/ `astro/tsconfigs/strictest` enabled

### DIFF
--- a/packages/integrations/image/src/lib/get-picture.ts
+++ b/packages/integrations/image/src/lib/get-picture.ts
@@ -51,7 +51,7 @@ export async function getPicture(params: GetPictureParams): Promise<GetPictureRe
 	}
 
 	if (!widths || !Array.isArray(widths)) {
-		throw new Error('[@astrojs/image] at least one `width` is required');
+		throw new Error('[@astrojs/image] at least one `width` is required. ex: `widths={[100]}`');
 	}
 
 	const aspectRatio = await resolveAspectRatio(params);
--- a/.changeset/polite-wombats-play.md
+++ b/.changeset/polite-wombats-play.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/image': patch
+---
+
+Updated error message for missing `widths` prop to provide an example
--- a/.changeset/strange-olives-rest.md
+++ b/.changeset/strange-olives-rest.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/image": patch
+---
+
+Allow passing `undefined` to TransformOptions, this is a types fix for users with `exactOptionalTypes` enabled in their tsconfig
--- a/packages/integrations/image/src/loaders/index.ts
+++ b/packages/integrations/image/src/loaders/index.ts
@@ -98,23 +98,23 @@ export interface TransformOptions {
 	 *
 	 * @default undefined The original image format will be used.
 	 */
-	format?: OutputFormat;
+	format?: OutputFormat | undefined;
 	/**
 	 * The compression quality used during optimization.
 	 *
 	 * @default undefined Allows the image service to determine defaults.
 	 */
-	quality?: number;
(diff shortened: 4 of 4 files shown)


packages/integrations/image/src/lib/get-picture.ts
```diff
@@ -51,7 +51,7 @@ export async function getPicture(params: GetPictureParams): Promise<GetPictureRe
 	}
 
 	if (!widths || !Array.isArray(widths)) {
-		throw new Error('[@astrojs/image] at least one `width` is required');
+		throw new Error('[@astrojs/image] at least one `width` is required. ex: `widths={[100]}`');
```
jasikpark, line 54: Maybe I should put this in a separate PR 😓 

I've been confused every time I've gotten this error, trying to pass `width="100vw"` then `widths="100vw"` until I get to `widths=[100]` and then `widths={[100]}` 😅 

weird that astro devtools doesn't seem to pick it up as an error

Princesseuh: Yeah, this message isn't great. I'm fine with this being in the same PR, it's just a very simple string change.

Princesseuh (approved):
Looks good to me, thank you!

Issue #5834, opened 2023-01-11
Duplicate stylesheet injection in production build


kiley-mitti:
### What version of `astro` are you using?

1.9.1

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac, Linux

### Describe the Bug

When using a layout which extends another layout, CSS stylesheets get injected twice in the production build of the site. This occurs on any page using the extended layout.

The first injection occurs in the `<head>` of the page.

The second injection occurs at the beginning of the `<body>` of the page.

Thank you so much for your attention to this issue! If you need any further information let me know.

### Link to Minimal Reproducible Example

[https://stackblitz.com/edit/github-2ugymh-oqddw9](https://stackblitz.com/edit/github-2ugymh-oqddw9)

**UPDATE (1/19/23):** We've found that this also extends to `injectScript` as well, so built a second example to account for that too. Hopefully this is useful to you!
 
[https://stackblitz.com/edit/github-2ugymh-e6stgx](https://stackblitz.com/edit/github-2ugymh-e6stgx)

### Participation

- [ ] I am willing to submit a pull request for this issue.

nicholasray:
I think this issue and https://github.com/withastro/astro/issues/5715 might be related

z336:
Adding a +1 to this bug, I'm seeing link and script tags injected twice on pages using one of the layouts. Thanks! 

### What version of astro are you using?
1.9.0

### Are you using an SSR adapter? If so, which one?
No

### What package manager are you using?
npm

### What operating system are you using?
Mac

### Describe the Bug
Stylesheet and script tags are being injected into the body with one layout but not both. The layout it's happening to is PageLayout.astro in my repo. 

### Link to Minimal Reproducible Example
https://astro-docs-prototype.netlify.app/
https://github.com/z336/astro-docs-prototype

matthewp:
Thanks for the repro! I'll be working on this soon.

marcfilleul:
I had this issue with nested layouts so I reverted back to a unique layout for now.

Now I have the same issue in a component with a <slot /> so it looks like the imported js and css are duplicated as soon as we use a <slot /> (in a layout or in a component).

matthewp:
I have a fix for this, just working out some details to avoid breaking some corner case layouts. Should have this in a patch next week.

z336:
[...]

Issue #5715, opened 2023-01-01
Duplicate css and js bundle tags when mdx is rendered in sub-layout


fflaten:
### What version of `astro` are you using?

1.8.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

When MDX imports a component and is rendered using a sub-layout/nested layout (ex. `BlogPostLayout` which again uses `BaseLayout`), link tags for bundled stylesheets and bundled JS module may be injected twice in `astro build`. The duplicates are placed in the body before the component.

**Run `astro build` in repro and inspect generated HTML.**

Repro includes three layouts:
- One (base). This always works as expected - no duplicates
  - Two (inherits One).
    - Three (inherits Two)

Pages: 
- `index*.astro` imports `<Content />` from MDX. `render()` with content collections behaves the same.
- `test*.mdx` uses layout-frontmatter. This only fails using layout Three for some unknown reason.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ske6w4

### Participation

- [ ] I am willing to submit a pull request for this issue.

thejediknight:
Facing same issue, I amusing astro 1.9.

nicholasray:
Also seeing this issue on [my website](https://www.nray.dev/) which uses Astro 1.9.1

sarajw:
Another one here - not just in dev but in the built static site too. Makes editing CSS in devtools a little annoying!

Just a thank you in advance, at least I hope this is the root of the issue I'm having :)

matthewp:
Sorry about this one. There's a couple of CSS ordering / duplication bugs floating around and I'm tackling all of those now.

fflaten:
No more duplicates (🎉) but it injects link+style in body now. Is that intentional?

PR #6034, opened 2023-01-30
Simplified head injection

matthewp:
## Changes

- This simplifies conditional head injection, fixing a lot of bugs related to CSS being out of place when using multiple levels of layouts with slots.
- Instead of rendering the head to a string in place, calls to renderHead and maybeRenderHead now instead yield out an instruction. At the page level this instruction is checked and if the head hasn't been rendered it will be.
- This means that head rendering is handled at the page level. This is the same mechanism used to inject the client directives and that seems to have worked well.
- Fixes https://github.com/withastro/astro/issues/5715
- Fixes https://github.com/withastro/astro/issues/5834

## Testing

- Unit tests added for multiple levels of layouts, including layouts that do not include explicit head tags.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/runtime/server/render/head.ts
+++ b/packages/astro/src/runtime/server/render/head.ts
@@ -1,8 +1,6 @@
 import type { SSRResult } from '../../../@types/astro';
-import type { RenderHeadInstruction } from './types';
 
 import { markHTMLString } from '../escape.js';
-import { renderChild } from './any.js';
 import { renderElement } from './util.js';
 
 // Filter out duplicate elements in our set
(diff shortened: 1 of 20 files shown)


packages/astro/src/runtime/server/render/head.ts
```diff
@@ -13,14 +12,8 @@ const uniqueElements = (item: any, index: number, all: any[]) => {
 	);
 };
 
-async function* renderExtraHead(result: SSRResult, base: string) {
```
matthewp, line 16: Removing this function because it causes head rendering to be async. We weren't utilizing this capability anyways. When/if we allow components to inject head content we'll likely want to bring this back, through a small refactor.

bholmesdev (approved):
Change makes sense to me! All for sticking with existing patterns. Should get a`head` of future issues

natemoo-re (approved):

robknight:
There seems to be a bug in the MDX integration test here:
```
 it('only injects contents into head', async () => {
                        const html = await fixture.readFile('/indexThree/index.html');
                        const { document } = parseHTML(html);

                        const links = document.querySelectorAll('link[rel=stylesheet]');
                        expect(links).to.have.a.lengthOf(1);

                        const scripts = document.querySelectorAll('script[type=module]');
                        expect(scripts).to.have.a.lengthOf(1);
                });
```
The selector is not scoped to the head tag, so it will match on _any_ link or script tag anywhere in the document.

As per #6057, it seems that this update causes scripts and styles to be inserted into the body, both within MDX documents and components that include layouts, causing a flash of unstyled content. I've also seen this in my own site.

PR #6038, opened 2023-01-30
Increase test timeout for flaky macOS test

natemoo-re:
## Changes

What it says in the title

## Testing

Yes

## Docs

No


natemoo-re:
> Will increasing the timeout fix this? I thought it was flaky because it would sometimes completely hang, not because it took a long time.

Ah, thought it was worth trying, but maybe this won't help/isn't the correct fix!

natemoo-re:
Welp, that didn't work! Closing.

PR #6040, opened 2023-01-30
Surface `allPages` build output in the `astro:build:done` hook.

ob6160:
## Changes

- We expose the `allPages` build output as `pageBuildData` in the arguments object passed to `astro:build:done`.
- This solves my scenario in which the `hoistedScript` information isn't set yet in the `pages: PageBuildData;` passed to the `astro:build:setup` hook.
- By exposing this the page build data after the build is complete on call of the `astro:build:setup` hook, integrations are able to make use of this information as `hoistedScript` will be set by this point.

## Testing

- Not sure which level is best to add tests for this or if they are needed for this kind of change as I'm a new contributor. Pointers welcome!

## Docs

I believe that the integration API docs page will need to be updated to document this new argument property if the change goes ahead.


matthewp (changes requested):
Thanks for the suggestion! I agree that surfacing this information is a good idea. But the `AllPagesData` object is very much an internal object that will likely break and doesn't make a lot of sense if you don't know how Astro's internal build works.

What I would love to see is reworking the `routes` object to provide more of this type of information. I know that @bholmesdev was interested in doing that as well.

I think this is complicated enough that it needs to go through our RFC process: https://github.com/withastro/roadmap

ob6160:
> Thanks for the suggestion! I agree that surfacing this information is a good idea. But the `AllPagesData` object is very much an internal object that will likely break and doesn't make a lot of sense if you don't know how Astro's internal build works.
> 
> What I would love to see is reworking the `routes` object to provide more of this type of information. I know that @bholmesdev was interested in doing that as well.
> 
> I think this is complicated enough that it needs to go through our RFC process: [withastro/roadmap](https://github.com/withastro/roadmap)

Thanks for the extra context around it being an internal object, it makes sense that the public API should be a subject of discussion in an RFC 👍 @bholmesdev I'd be happy to do set aside some time to collaborate on putting together an RFC for an improved `routes` object 😊

__Cheeky extra question (only if you have the time of course, or can think of a quick set of examples I could look at):__

In our specific use case we're making an integration that shifts the Astro build output into a set of files / directories that fit the format expected by a piece of internal tooling. Specifically, a single html, js, and css file for the whole build.

Without a change like this to provide metadata such as `hoistedScript` I'm quite stuck in my thinking around how we could easily traverse the build output and get the contents of each file without going into the file contents and grepping for the script to manually locate and coalesce them which feels.. fragile 🤔 . Am I missing another point of integration that would make this a little easier?

bholmesdev:
@ob6160 Thanks for the details! So it sounds like you want to un-tangle the bundle that Astro generates to isolate each page as an HTML, CSS, and JS chunk correct? If so, you might want to check [rollup's manual chunks option](https://rollupjs.org/configuration-options/#output-manualchunks) to manually specify the names of generated files to guarantee a predictable output. This is configurable from the `vite` option:

```ts
export default defineConfig({
  vite: {
    build: {...}
  }
})
```

I'll confess though, I'm not sure what the best destination for these files would be for your use case. I'd suggest playing around to see if that solution is feasible!

Otherwise, I'm happy to help review and answer questions for a proposal. I'd avoid loosely exposing build internals like this though, since these internals were built with core testing in mind rather than user-facing intuitive-ness. We should also be sure to expose options with a specific use case we can call out in our documentation. Hope that helps!

ob6160:
@bholmesdev aha! thanks for the pointer around the rollup manual chunks option, will give that a go and see if it's a good fit for our use-case here.

Yeah I think I'm a bit confused by the `pages` property exposed as an argument to the `astro:build:setup` hook. That seems to expose a lot of the same information as the `allPages` object that was the subject of this PR. Would a proposal involve thinking around all of the data that's exposed through the hook system?

bholmesdev:
@ob6160 Yep, I agree our `routes` param exposes a lot of build internals today, some options making... little sense to expose 😅 We realized this coming into 2.0, and would love to standardize what we expose more smartly going forward. I think an RFC for this could guide future refactors!

PR #6045, opened 2023-01-30
Handle invalid argument errors

natemoo-re:
## Changes

- Previously, the dev server completely crashed when doing something like
```astro
---
import Foo from '../components/Foo.astro';
// Called directly via function call
console.log(Foo())
---
// Called inside an expression without `<Foo />` syntax
{items.map(Foo)}
```
- This PR fixes this situation by validating the input arguments before attempting to render
- If the arguments are invalid, we now throw a new `InvalidComponentArgs` error with helpful info
- 
CleanShot 2023-01-30 at 13 20 57@2x

## Testing

Tested manually since this is a smaller one. Can add a test if needed

## Docs

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -415,6 +415,20 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		},
 		hint: 'Mutable values declared at runtime are not supported. Please make sure to use exactly `export const prerender = true`.',
 	},
+	/**
+	 * @docs
+	 * @message
+	 * **Example error messages:**<br/>
+	 * InvalidComponentArgs: Invalid arguments passed to <MyAstroComponent> component.
+	 * @description
+	 * Astro components cannot be rendered manually via function call, such as `Component()` or `{items.map(Component)}`. Prefer the component syntax `<Component />` or `{items.map(item => <Component {...item} />)}`.
+	 */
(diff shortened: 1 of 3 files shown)


matthewp (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -415,6 +415,20 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		},
 		hint: 'Mutable values declared at runtime are not supported. Please make sure to use exactly `export const prerender = true`.',
 	},
+	/**
+	 * @docs
+	 * @message
+	 * **Example error messages:**<br/>
+	 * InvalidComponentArgs: Invalid arguments passed to <MyAstroComponent> component.
+	 * @description
+	 * Astro components cannot be rendered manually via function call, such as `Component()` or `{items.map(Component)}`. Prefer the component syntax `<Component />` or `{items.map(item => <Component {...item} />)}`.
```
Yan-Thomas, line 424: Small nit: `via function call` -> `via a function call`
```suggestion
	 * Astro components cannot be rendered manually via a function call, such as `Component()` or `{items.map(Component)}`. Prefer the component syntax `<Component />` or `{items.map(item => <Component {...item} />)}`.
```

Yan-Thomas (commented):
Nice work @natemoo-re, I added a small docs related suggestion, but overall LGTM!

PR #6052, opened 2023-01-31
error overlay: show `cause` if available

### LINKED ISSUE #6027: Error overlay not showing `cause`

2.0.2

N/A

npm

Windows / Stackblitz

### Describe the Bug

I noticed that the new error overlay is hiding two types of info:
- anything in the [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property of the error
- anything printed to the console before the error

These two ways are often used to add additional context to the error before rethrowing. Would it be possible to show these in the overlay?

https://stackblitz.com/edit/github-rmhv9a?file=src%2Fpages%2Findex.astro

### Participation

mayank99:
## Changes

Added ability to display an error's [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) if available. Closes #6027

## Testing

Tested in a local project by changing code inside node modules.

Test fixture currently pending.

## Docs

Not needed I think? I guess we could document how to use the `cause` property.

### DIFF
--- a/.changeset/clever-panthers-end.md
+++ b/.changeset/clever-panthers-end.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Error overlay will now show the error's `cause` if available.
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
Awesome stuff, thank you very much!

PR #6055, opened 2023-01-31
[Lit] Forwards compatiblity for streaming Declarative Shadow DOM

e111077:
## Changes

-  Adds forwards compat for DSD
  - [Context](https://groups.google.com/a/chromium.org/g/blink-dev/c/xzT-vN-bx0s)
- Renders DSD templates with both the soon-to-be legacy `shadowroot` attribute and the `shadowrootmode` attribute
- This change to the spec is happening in Safari TP and Chrome 111

## Testing

Tests will continue to pass in Chrome 111 and Safari

Updated tests to reflect new expected templates

## Docs

This is an implementation detail that doesn't need to be surfaced to the user.


matthewp:
Wow, I didn't realize they were changing this attribute name. A bit unfortunate. Thanks for the context!

PR #6056, opened 2023-01-31
[Lit] Fix hydration not having the same reactive values as SSR template

### LINKED ISSUE #5777: Lit Element loses prop state for subsequent renders when rendered as island

1.9.0

None

npm

Mac

### Describe the Bug

Hi,

I've come across an issue that doesn't render lit elements correctly when rendering as an island.
The inital render is just fine, though subsequent renders fail to remember the current value / state of props that do not have default values defined in the lit constructor function. This just works fine when using lit elements in other environments than Astro.

We've put together a codesandbox that demo's the undesired behavior. It consists of 3 similar lit elements that are rendered as islands. One of them has a constructor value, a second one doesn't and a third one also doesn't but reflects the attribute (which also seems to fix the issue at hand).

When the button is pressed you can see that 1 of the elements loses its correct state.

https://codesandbox.io/p/sandbox/astro-lit-update-bug-h1imu8?file=%2Fsrc%2Fpages%2Findex.astro

### Participation

e111077:
fixes #5777

## Changes

- Lit SSR will only bind to attributes and rely solely on attribute conversion
  - Fixes issue in most cases where hydration fails to match template
  - **REASONING**
    - Otherwise would need client-side JS to update the properties to match the SSR'd content
    - Since Astro is SSG, it doesn't make sense to introduce scripts to update the nodes's prop values
    - Handling all JSX bindings as attribute bindings serializes state into the DOM
- **RESTRICTIONS**
  - If your island needs a property binding to function, it's not an SSR-able lit component
- **BREAKING CHANGES**
  - user must bind to attribute if different from property
    - e.g. `@property({attribute: 'str-attr'}) str = ''` must bind to `str-attr` instead of `str`

## Testing

Refactored tests to reflect this new behavior of setting attributes instead of properties

## Docs

Docs need to be added to mention that Lit integration JSX bindings do the following conversions:

- **Boolean**

```jsx
// JSX
<MyElement bool={true} boolFalse={false} boolStr="false"/>
// HTML
<my-element bool boolStr="false"></my-element>
```

- **Object**

Objects bound to a Lit element all go through `JSON.stringify`

- **String & number**

Bound directly as string

Additionally docs should note that all JSX to Lit bindings are only attribute bindings. That means if you have a custom attribute set, you must now bind to that attribute rather than the property as before. e.g.

```tsx
// my-element.ts
@customElement('my-element')
export class MyElement extends LitElement {
  @property({attribute: 'str-attr'}) str = '';
}

// page.astro - new
<MyElement str-attr={'yoloswag'} />

// page.astro - old, no longer works
<MyElement str={'yoloswag'} />
```


e111077:
Better solution found. Will refactor

PR #6065, opened 2023-01-31
fix(webapi): Polyfill with undici imports directly

Princesseuh:
## Changes

When I tried this in my original Node 14 PR, it didn't work... but now it seems like it does? I don't really get it 🤔

## Testing

Current test covers this

## Docs

N/A


Princesseuh:
!preview netlify-undici-fix

natemoo-re (approved):

bluwy (approved):
Should we add a changeset to trigger a release?

Princesseuh:
> Should we add a changeset to trigger a release?

It doesn't do anything on the user side (or at least, shouldn't). So not too worried about it going out or later, personally

PR #6068, opened 2023-01-31
Correct usage of `getImage` function

MoustaphaDev:
## Changes

Show correct usage or `getImage` function.

When providing a string to the `getImage` function, it is taken as a remote image. To use local images we need to import them.
We need to provide `alt` too as it's required

## Testing

N/A docs change only

## Docs

Docs change to match correct usage


sarah11918 (approved):
LGTM, @MoustaphaDev! Thanks so much!  :raised_hands:

Issue #5915, opened 2023-01-20
SolidJS integration breaks app when used in Vercel Edge


pilcrowOnPaper:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

Vercel (edge)

### What package manager are you using?

pnpm

### What operating system are you using?

MacOS

### Describe the Bug

When using Solid components in an app hosted on Vercel edge, it throws the following error:

```
TypeError: Failed to execute 'encode' on 'TextEncoder': parameter 1 is not of type 'String'.
    at (entry.mjs:41:5485)
    at (entry.mjs:93:13)
```

### Link to Minimal Reproducible Example

https://github.com/pilcrowOnPaper/astro-vercel-solid

### Participation

- [ ] I am willing to submit a pull request for this issue.

jmn:
I ran into the same thing with Preact just now.

bholmesdev:
Thanks for reporting! Wouldn't be the first component issue on Vercel edge. @pilcrowOnPaper @jmn have either of you been able to repro with Astro 2.0?

jmn:
Yes. I'm on Astro 2.0

helmisatria:
I ran into the same thing, not using any specific UI framework/library (only .astro) and deployed on Cloudflare pages

AirBorne04:
I narrowed it down to the `chunkToByteArray` function beeing called with an object (with type: 'head') which is causing the issue. This is an issue happening besides ui framework and adapters just the vercel adapter does error out on it.

https://github.com/withastro/astro/blob/eed81fa213bc0bd26384074bf0007323f04623f7/packages/astro/src/runtime/server/render/common.ts#L80

AirBorne04:
@pilcrowOnPaper is there a way to test vercel edge runtime locally?

pilcrowOnPaper:
@AirBorne04 I don't think so. I know it's based on Cloudflare workers but I'm not sure if there's a way to run that locally as well.

AirBorne04:
I managed to get what i wanted. Pretty funky all this :smile: Cloudflare has wrangler which runs the project locally. For vercel i managed now to upload the prebuilt and debug with the error messages. I have two minor issues so far and reached a render issue with Solid-js which we did fix for the cloudflare adapter, but vercel is building different and therefore i need to check how we can fix this as well.

HiDeoo:
> is there a way to test vercel edge runtime locally?

[...]

PR #6070, opened 2023-01-31
fix: Failed to execute 'encode' on 'TextEncoder': parameter 1 is not of type 'String' in Edge Runtime SSR

AirBorne04:
## Changes

- fix small issue in SSR mode #5915
- this is not fixing the Solid build for vercel, so for Solid it will still state something like `Error: Pa is not supported in the browser, returning undefined` that is a vercel adapter issue only and we need to update the build steps there.

## Testing

Tested example applications for react and solid on vercel.

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -261,9 +261,11 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 
 			if (isPage || renderer?.name === 'astro:jsx') {
 				yield html;
-			} else if (html) {
+			} else if(html && html.length > 0) {
 				yield markHTMLString(html.replace(/\<\/?astro-slot\>/g, ''));
-			}
+			} else {
+        yield '';
+       }
 		})();
 	}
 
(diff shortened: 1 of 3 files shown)


AirBorne04:
These fixes are only for the issues in core / to fix eg cloudflare for now. The solid + vercel needs an additional fix in the vercel adapter.

bholmesdev (approved):
NWTWWHB (not worse than what we had before!)

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -261,7 +261,7 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 
 			if (isPage || renderer?.name === 'astro:jsx') {
 				yield html;
-			} else {
+			} else if (html) {
```
bholmesdev, line 264: Could this cause the server to hang if there is _not_ HTML? Could we have an `else` case with a detailed internal error, or does fall-through work?

bholmesdev (commented):
Changing my review sorry! Comment before merging

packages/astro/src/runtime/server/render/component.ts
AirBorne04: yeah you are right i replace it with an empty string return now, that should do.

bholmesdev (approved):

Issue #6066, opened 2023-01-31
Functions directory for Cloudflare Adapter on directory mode should use the configured root directory instead of the cwd.


NachoVazquez:
### What version of `astro` are you using?

2.0.4

### Are you using an SSR adapter? If so, which one?

Cloudflare, 6.1.1

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

When choosing `directory` mode with the Cloudflare SSR adapter and running the **build** command, the `functions` folder is created on the `process.cwd` instead of using the configurable root folder.

This is problematic when using a monorepo where multiple projects with different functions folders live together, and the build command is run from the root of the monorepo instead each application.

### Link to Minimal Reproducible Example

N/A

### Participation

- [X] I am willing to submit a pull request for this issue.

NachoVazquez:
Yes, exactly!

NachoVazquez:
I have PR in progress here, but I need a bit of help with tests. https://github.com/withastro/astro/pull/6075

PR #6075, opened 2023-01-31
fix: use the root of the project as the functions location

NachoVazquez:
## Changes

- Use the configuration root as the location of the Cloudflare Pages functions when using the `directory` mode.

Fixes: #6066 

## Testing

Test the location of the functions folder creation.

## Docs

Docs don't need to be updated, docs already state that the root of the project is the place to generate the functions folder.


NachoVazquez:
I need help with the tests here. The fixture doesn't seem to work correctly when creating the functions folder, and the previous tests were naive.

matthewp:
I assume the problem is with using relative paths in the test, right? Those are defined here: https://github.com/withastro/astro/blob/main/packages/astro/test/test-utils.js#L168-L171 if it helps with debugging.

NachoVazquez:
Part of the problem is that the functions folder is not being created when I read the elements of the folder in the `console.log`.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -203,7 +203,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				}
 
 				if (isModeDirectory) {
-					const functionsUrl = new URL(`file://${process.cwd()}/functions/`);
+					const functionsUrl = new URL(`file://${_config.root}/functions/`);
```
AirBorne04, line 206: This is not working for me. `_config.root` is a URL itself.
Changing this line to
`const functionsUrl = new URL("functions", _config.root);`
works for me, also the test is running fine than.

AirBorne04 (changes requested):

packages/integrations/cloudflare/src/index.ts
NachoVazquez: Awesome, that helps! Thanks for the feedback! Making the changes now.

NachoVazquez: I made this change, but tests keep failing locally for me. I will push the update so we can have the same reference.

NachoVazquez: It does seem to work, interesting. 

Thank you :)

packages/integrations/cloudflare/test/directory.test.js
```diff
@@ -1,20 +1,21 @@
...
 			root: './fixtures/basics/',
+			output: 'server',
 			adapter: cloudflare({ mode: 'directory' }),
 		});
+		await fixture.build();
 	});
 
-	it('Builds', async () => {
-		await fixture.build();
+	it('generates the functions folder on under the config root', async () => {
```
AirBorne04, line 18: can change the text to
`generate functions folder inside the project root`

packages/integrations/src/env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="astro/client" />
```
AirBorne04, line 1: can you remove this file from the commit?

AirBorne04 (changes requested):
Sorry for being picky 😄

packages/integrations/src/env.d.ts
NachoVazquez: Sure thing! Sorry for the small details; I'm new to Astro, and I wasn't sure if this was needed.

packages/integrations/cloudflare/test/directory.test.js
NachoVazquez: That sounds better :)

NachoVazquez:
> Sorry for being picky 😄

Thanks for being picky :)

packages/integrations/cloudflare/functions/[[path]].js
```diff
@@ -0,0 +1,105 @@
+globalThis.process = {
```
Comment (AirBorne04) [inline packages/integrations/cloudflare/functions/[[path]].js:1]: sorry I did miss this one, can you also delete the functions folder, anything that got build should not be added to the repo

Comment (NachoVazquez) [reply packages/integrations/cloudflare/functions/[[path]].js:1]: Good catch!

AirBorne04 (approved):
Nice job!

NachoVazquez:
> Nice job! Thank you for making Astro better!

Thanks for your help!!!

NachoVazquez:
Hi @AirBorne04 , sorry to bother you, but do you have any info about when the next release of the Cloudflare adapter will happen? I need this changes :) thank you!

NachoVazquez:
Nevermind, I think it is out already! Thank you!

PR #6076, opened 2023-01-31
Add JSDOC `@message` param to broken error pages

Yan-Thomas:
## Changes

Added JSDOC `@message` param to `InvalidContentEntrySlugError` & `ContentSchemaContainsSlugError` errors to fix their generated docs pages.

| Before | After |
| image | image |
| image | image |

The function we use in docs to extract the error messages from this file expects functions to follow this style:

```ts
message: (params: string) => `Message ${params}`,
```

These and a few other errors follow this, instead:

```ts
message: (configFile: string) => {
        return message: (params: string) => `Message ${params}`;
},
```

Not sure if this function style makes a difference in the core side of things, if possible I'd advocate going with the former function style so we avoid possible errors (and if it can't be avoided, to use `@message` for these error messages).

When available, the docs doc gen script uses the JSDOC `@message` param instead, so I added it to the problematic functions as a fix. Also because they have dynamic values we can't bring to docs (like the current `collection` or `entryId`), in which case, it's better to use JSDoc anyway to overwrite these values with `COLLECTION_NAME` and `ENTRY_ID`, for extra clarity.

## Testing

Only JSDOC comments, shouldn't affect code behavior.

## Docs

It's all docs baby 😎 

cc @withastro/maintainers-docs for feedback!

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -629,6 +629,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	},
 	/**
 	 * @docs
+	 * @message `COLLECTION_NAME` → `ENTRY_ID` has an invalid slug. `slug` must be a string.
 	 * @see
 	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/)
 	 * @description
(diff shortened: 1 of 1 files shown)


hippotastic (approved):
LGTM and fixes the issue in docs!

bholmesdev (approved):
Gotcha, will `doc`tor up our errors with `@message` in the future!

PR #6082, opened 2023-02-01
Refactor `create-astro`

natemoo-re:
## Changes

- Creating a draft to watch CI
- Resigned UI for `create-astro`, with more features and much better CLI flags

## Testing

- `create-astro` had very poor test coverage previously. We were testing the CLI output directly, which was flaky and led to most tests being disabled.
- This PR refactors `create-astro` completely so that `prompt` and other internals can be mocked and the logic of each step can be tested independently

## Docs

TODO!


matthewp:
Like how the refactor allows better testing, love all of the new tests that are coming out of this.

Not sure I feel about using vitest here. Putting aside whether it's appropriate in this context (given that it compiles code differently than in prod), having multiple test libraries in the same repo does make it harder to work with. I'm open to changing test libraries, but I don't think a refactor of one package is the right place to make that decision.

I see that some of the tests are relying on vitest's mocking capabilities, was that the reason to switch? I would think this code could be changed to not need the mocking. For example there is a test that checks for process exit. Can this be refactored so that process.exit() is not called directly in the code, but instead you pass it a function (just as an example) for when the process should exit. That way you can test without needing to mock.

packages/create-astro/src/actions/dependencies.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export async function dependencies(ctx: Pick<Context, 'install'|'yes'|'prompt'|'pkgManager'|'cwd'|'dryRun'>) {
+	let deps = ctx.install ?? ctx.yes;
+	if (deps === undefined) {
+		({ deps } = await ctx.prompt({
+			name: 'deps',
+			type: 'confirm',
+			label: title('deps'),
+			message: `Install dependencies?`,
+			hint: 'recommended',
```
Yan-Thomas, line 15: Pointing out this is the only prompt with `hint: 'recommended'` rather than `hint: '(recommended)'`, if this wasn't intentional I'd change for consistency's sake:

```suggestion
hint: '(recommended)',
```

Yan-Thomas (approved):
Great work @natemoo-re, docs LGTM! I do have a small nitpick regarding a prompt message that can be dismissed if intentional 🙌

natemoo-re:
> I see that some of the tests are relying on vitest's mocking capabilities, was that the reason to switch? I would think this code could be changed to not need the mocking. For example there is a test that checks for process exit. Can this be refactored so that process.exit() is not called directly in the code, but instead you pass it a function (just as an example) for when the process should exit. That way you can test without needing to mock.

Yes, I was using `vitest` to allow mocking `console.log` and `process.stdout.write`, but I can definitely refactor this with a similar setup to our `logger` in `astro` itself. I will do that and stick with Mocha!

natemoo-re:
@matthewp tests refactored to stick with `mocha`/`chai` as we currently use, with the added benefit of zero mocking! 

Everything is accessed through a shared `Context` object and `process.stdout.write` is also handled internally so we can intercept them at the test level.

matthewp (approved):
Code looks good, love all of the new testing and how easy that seems to be.

Adding Sarah and Fred as reviewers since they have context about the UI changes.

FredKSchott:
@natemoo-re can you share a screen recording of the whole thing? I'll push to make sure this convo doesn't get too bike-sheddy but I think it's still an important part of evaluating this PR. I'll save my review to only look at that!

packages/create-astro/src/messages.ts
```diff
@@ -1,126 +1,183 @@
...
-	}
+export const title = (text: string) => align(label(text), 'end', 7) + ' ';
+
+export const welcome = [
+    `Let's claim your corner of the internet.`,
+    `I'll be your assistant today.`,
+    `Let's build something awesome!`,
+    `Let's build something great!`,
+    `Let's build something fast!`,
+		`Let's build the web we want.`,
```
sarah11918, line 30:
```suggestion
    `Let's build the web we want.`,
```
Just in case this was a mistake!

packages/create-astro/src/messages.ts
```diff
@@ -1,126 +1,183 @@
...
+    `Let's make the web weird!`,
+    `Let's make the web a better place!`,
+    `Let's create a new project!`,
+    `Let's create something unqiue!`,
```
sarah11918, line 34:
```suggestion
    `Let's create something unique!`,
```
I mean, I almost didn't want to change it!  :joy:

natemoo-re: Haha, incredible!

sarah11918 (approved):
Approve of the new TypeScript messages! Looking forward to dealing with this less in docs!  :sweat_smile:

FredKSchott:
Looks great! I may still push to tweak in the future (I'd still love to find a way to remove that "strictness" question) but this is absolutely an improvement and I am HERE for it.

FredKSchott (approved):

natemoo-re:
!preview refactor

bholmesdev (approved):
Creating an astronomical improvement 👏

Princesseuh (approved):
I'll fight with Fred, scream and throw tantrums for the "strictness" question to be kept forever, but just wanted to add my approval to the wonderful work you've done here Nate!

Issue #5924, opened 2023-01-21
@astrojs/image overrides import type for images (including code in other frameworks like react)


JLarky:
### What version of `astro` are you using?

1.9.2

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

importing images in React gives typescript error because `@astrojs/image` overrides the type (but not value) from `string` to `ImageMetadata`

code:

```
import logo from './dummy.png';

export default function Counter() {
  // img is expecting a string but we get ImageMetadata type instead
  return (
      <img src={logo} />
  );
}

```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xdmbqd?file=src/components/Counter.tsx

### Participation

- [ ] I am willing to submit a pull request for this issue.

# Update

Thanks for @aFuzzyBear [message in discord](https://discord.com/channels/830184174198718474/1068743304721748039/1069617899683184732) that fixes this issue by adding `?url` in image import

```
import logo from './dummy.png?url'; // ?url makes sure that import is a string

export default function Counter() {
  return (
      <img src={logo} />
  );
}
```

bluwy:
The value should be of shape `ImageMetadata` if you pass the `image()` in the Astro config. If you're not using the integration intentionally, then you probably shouldn't reference the images types in the `env.d.ts` file, which should fix it.

JLarky:
> The value should be of shape `ImageMetadata` if you pass the `image()` in the Astro config. If you're not using the integration intentionally, then you probably shouldn't reference the images types in the `env.d.ts` file, which should fix it.

This example was intentionally simplified. In a real project I use astro image in astro files, but it messes up with stuff in React files now. So if I remove those types to fix React now I will have the error in Astro

bluwy:
Maybe we could make the image integration to work in `.astro` and `.mdx` files only. But it's still not possible to type `.png` etc imports differently depending if you're in an `.astro` file or `.tsx` file, since TypeScript only allows you to define one.

_Maaaybe_ you can create 2 tsconfigs, one targets `.astro` files and the `env.d.ts` (with `@astrojs/image/client`); the other targets the rest without the `env.d.ts`, and TypeScript might be able to type the import differently then.

JLarky:
[...]

PR #6083, opened 2023-02-01
fix(@astrojs/image): ignore vue public image imports

### LINKED ISSUE #6071: Astro image attempting to load images from Vue components

2.0.4

None

npm

Windows

### Describe the Bug

I have a Vue component that is using some standard ``  and when I try to run build on my machine or Netlify I get...

[@astrojs/image] Could not load /images/old-ute-valley-park-photos/old-ute-1.jpg (imported by src/components/OldPhotos.vue?vue&type=script&setup=true&lang.ts): ENOENT: no such file or directory, open 'D:\images\old-ute-valley-park-photos\old-ute-1.jpg'
 error   Could not load /images/old-ute-valley-park-photos/old-ute-1.jpg (imported by src/components/OldPhotos.vue?vue&type=script&setup=true&lang.ts): ENOENT: no such file or directory, open 'D:\images\old-ute-valley-park-photos\old-ute-1.jpg'
Error: Could not load /images/old-ute-valley-park-photos/old-ute-1.jpg (imported by src/components/OldPhotos.vue?vue&type=script&setup=true&lang.ts): ENOENT: no such file or directory, open 'D:\images\old-ute-valley-park-photos\old-ute-1.jpg'

Not sure why Astro image is trying to get involved here.  I'm not importing it anywhere in the minimal repro. The Vue component is just using standard HTML img tags that are pointing to an image in the /public/images directory.

https://github.com/drewtownchi/image-test

### Participation

MoustaphaDev:
## Changes
Fix #6071 

During build, vite-plugin-vue's `template.transformAssetUrls.includeAbsolute` option is [hard-coded as `true`](https://github.com/vitejs/vite-plugin-vue/blob/main/packages/plugin-vue/src/template.ts#L131)
That causes images in vue components like `` to generate import requests for `/images/blah.jpg`.

Since the id ends with `.jpg` it will get picked up by the image integration which will try to load `/images/blah.jpg` which won't resolve to anything.
This change makes the image integration ignore those imports.

## Testing

Check if building a project with the same failing conditions stated earlier doesn't throw an error

## Docs

N/A bug fix only


.changeset/four-ducks-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/image': patch
+---
+
+Ignore of vue public imports during build
```
MoustaphaDev, line 5: will correct this typo

packages/integrations/image/src/vite-plugin-astro-image.ts
```diff
@@ -39,8 +49,8 @@ export function createPlugin(config: AstroConfig, options: Required<IntegrationO
 			}
 		},
 		async load(id) {
-			// only claim image ESM imports
-			if (!filter(id)) {
+			// only claim image ESM imports and ignore vue's absolute imports
+			if (!filter(id) || absoluteImageImports.has(id)) {
```
bluwy, line 53: I don't think we can change how we load files based on `resolveId` and the `source`/`importer`. This can cause a race condition where if we're trying to load an image in both an Astro file and Vue file.

For example, if we encounter the Vue file first, the image import would returning a URL as default export. Then, when we encounter the Astro file, it wouldn't transform the image as an object anymore, it uses the cached URL default export.

Unfortunately I don't have an idea of how to fix this yet. Perhaps @Princesseuh's work on image integration next could fix it.

A workaround could be to do https://github.com/withastro/astro/issues/5924#issuecomment-1410744335

PR #6084, opened 2023-02-01
Add option to control automatic doctype prepend

Sertion:
## Changes

Adds a setting to control the automatic doctype prepender as [discussed here](https://github.com/withastro/roadmap/discussions/266).

## Testing

Added a new test to the doctype tests that verifies that no doctype is added.

## Docs

The current proposed implementation adds a new option (`prependDoctype` type `boolean`) to the build (?) configuration.


.changeset/metal-dingos-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
matthewp, line 2: this would be a `minor` since it's a new feature.

packages/astro/src/@types/astro.ts
```diff
@@ -469,6 +469,17 @@ export interface AstroUserConfig {
 	 */
 	trailingSlash?: 'always' | 'never' | 'ignore';
 
+	/**
+	 * @docs
+	 * @name prependDoctype
+	 * @type {boolean}
+	 * @default true
```
Yan-Thomas, line 476: For better rendering in docs, we wrap default values with backticks.
```suggestion
	 * @default `true`
```

Yan-Thomas (commented):
Thanks for this @Sertion! I added a small docs suggestion for you 🙌

matthewp:
trying to understand the use-case better. You want to omit the doctype so you can use Astro pages as partials; does this mean that you also omit the head and body tags? I guess that works since we don't automatically append those.

Sertion:
> You want to omit the doctype so you can use Astro pages as partials; does this mean that you also omit the head and body tags?

That is right. I want to be able to generate HTML output that _only_ contains what I made it to contain without any unexpected side effects.

The use case is rendering content for technology such as [the nginx SSI module](https://nginx.org/en/docs/http/ngx_http_ssi_module.html).

Sertion:
Thanks for the update @natemoo-re. I looked at [@astrojs/compiler](https://github.com/withastro/compiler) to see if I could make the change but I could not.

What would be the next step?

matthewp:
I think I'm going to have to come out against this feature. Astro doesn't at present support partial generation. Any hacks we add are going to make everything more brittle and confusing for users.

It's called the `pages/` folder and expects to generate full pages. Partials are something we *could* support, I have heard some interest from people on the team, so it's not out of the realm of possibility. But it needs to go through the proposal process and be it's own feature if we are going to do it (which probably, imo, means not going through the pages folder). https://github.com/withastro/roadmap/discussions

Sertion:
> This is more complicated than a simple change and should go through the roadmap process. https://github.com/withastro/roadmap/discussions

Should it continue in [the existing proposal](https://github.com/withastro/roadmap/discussions/266) or would creating a new proposal be preferred?

natemoo-re:
Thanks for being flexible @Sertion! I think we can continue the discussion in https://github.com/withastro/roadmap/discussions/266 and hopefully come up with a more fleshed out proposal from there.

PR #6085, opened 2023-02-01
fix: solidjs integration for vercel edge build (adopting same mechanics as cloudflare)

### LINKED ISSUE #5915: SolidJS integration breaks app when used in Vercel Edge

AirBorne04:
## Changes

- A second build step has been added to allow the separation of framework and library builds
- Resolves #5915 

## Testing

With the example react and solid projects from the repo.

### DIFF
--- a/.changeset/wild-seas-happen.md
+++ b/.changeset/wild-seas-happen.md
@@ -1,5 +1,5 @@
 ---
-'@astrojs/vercel': major
+'@astrojs/vercel': patch
 ---
 
 Added second build step through esbuild, to allow framework defined build (vite build) and target defined bundling (esbuilt step)
--- a/.changeset/wild-seas-happen.md
+++ b/.changeset/wild-seas-happen.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': major
+---
+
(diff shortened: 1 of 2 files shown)


bholmesdev (approved):
esbuilt for success!

.changeset/wild-seas-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vercel': major
```
matthewp, line 2: i think this is a patch. even though it feels like a big change there isn't a reason to believe it's going to break anyone's code, or is there?

AirBorne04: I agree, that it can be a patch, since we know it is working well for cloudflare there should be little risk to make it a patch release.

Issue #6087, opened 2023-02-01
Request resource fail when the path has other language characters on SSR mode


QingXia-Ela:
### What version of `astro` are you using?

2.0.2

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

The resource will request fail when the path has Chinese on SSR mode with node. Only appear on build output server.

![Snipaste_2023-02-01_23-35-37](https://user-images.githubusercontent.com/86217807/216089048-3ad1bd75-a702-4c05-89a2-f0c37222f5d6.jpg)
![Snipaste_2023-02-01_23-35-16](https://user-images.githubusercontent.com/86217807/216089067-1062ab22-46e8-4274-b592-2f2752170885.jpg)

Maybe other language also has this problem?

I found that in the output server, the path will be encode one time:

https://github.com/withastro/astro/blob/f05eb9f8ba078a063e87c60309bbab491a043be4/packages/integrations/node/src/http-server.ts#L22

But when the browser send the request with Chinese path, it has been encoded. But with second time encoded, it makes path become incorrent.

![图片](https://user-images.githubusercontent.com/86217807/216093784-eccb7770-3893-4334-bb42-a3c2152ff924.png)

Maybe should change like this:
```js
const stream = send(req, encodeURI(decodeURI(pathname)), {
	root: fileURLToPath(client),
	dotfiles: pathname.startsWith('/.well-known/') ? 'allow' : 'deny',
});
```

### Link to Minimal Reproducible Example

https://github.com/QingXia-Ela/astro-bug-demo

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #6088, opened 2023-02-01
Fix incorrent encoded when path has other language characters

QingXia-Ela:
## Changes

- fixes https://github.com/withastro/astro/issues/6087
- more details: https://github.com/withastro/astro/issues/6087

## Testing

- tested with the examples at my local

## Docs

N/A, bug fix


matthewp:
Thanks!

PR #6092, opened 2023-02-01
[Content collections] Load content config with full Vite setup

### LINKED ISSUE #5990: Cannot resolve aliased module from `src/content/config.ts`

2.0.1

None

npm

Mac

### Describe the Bug

I have a module `src/data.ts` and `tsconfig.json` configured to resolve aliased names:

```json
{
  "extends": "astro/tsconfigs/strict",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "~/*": ["src/*"]
  }
}
```

I'm able to import `~/data` just fine from pages and other files, but not from `src/content/config.ts`. I get the error `Cannot find module '~/data' imported from '.../src/content/config.ts'`.

It all works however if I use relative import `../data` instead of the path alias.

no yet

### Participation

bholmesdev:
## Changes

- Resolves #5990 
- Dev: Load content config with existing dev server, instead of creating a new one
  - Moves type generation to _after_ dev server startup instead of before
  - Ensures `[content]` logs happen after server port logs, which should make them more visible
- Build: Load config by calling the `astro sync` command internally, with full `createVite` result instead of condensed config

## Testing

Ensure existing tests still pass

## Docs

N/A


bholmesdev:
Circling back to that failing unit test tomorrow. I discovered the `renderentry` configs _weren't getting loaded_ previously because it used an invalid config format (`schema: {` instead of `schema: z.object({`). Fixing this, it looks like the layout test fails when the config is present. The test _passes_ if I remove the `config.ts` mock 🤷

PR #6097, opened 2023-02-02
Return html when render markdown content

bluwy:
## Changes

When `render()` a markdown file (not MDX), return the `html` too.

Implements [this feature](https://github.com/orgs/withastro/projects/2/views/24?pane=issue&itemId=17970390) for better RSS composing experience. 

## Testing

Added new tests to make sure it renders

## Docs

- Changeset
- RSS guide update https://github.com/withastro/docs/pull/2545
- API reference update https://github.com/withastro/docs/pull/2547


.changeset/purple-glasses-pretend.md
```diff
@@ -0,0 +1,14 @@
+---
+'astro': minor
+---
+
+Return raw `html` when calling `render()` on a markdown content.
```
bholmesdev, line 5:
```suggestion
Expose `html` from the Content Collection `render()` result for Markdown entries. This is similar to the `compiledContent()` helper present on Markdown imports today.
```
Some rewording to clarify the usecase!

packages/astro/test/content-collections-render.test.js
```diff
@@ -195,5 +212,15 @@ describe('Content Collections - render()', () => {
...
+			const response = await fixture.fetch('/render-with-html', { method: 'GET' });
+			expect(response.status).to.equal(200);
+
+			const html = await response.text();
+			const $ = cheerio.load(html);
+
+			expect($('p').text()).to.equal('This is a post');
+		});
 	});
 });
```
bholmesdev, line 226: nit: as a sanity check, I might add a unit test from a `.ts` file as well. Just to confirm `html` is accessible when generating RSS feeds.

bholmesdev (commented):
Code looks good! Just nits on the changeset and about a potential unit test

bluwy:
Makes sense! Thanks for the explanation. I'll update the docs later for no2, and set up the unit test.

packages/astro/test/content-collections-render.test.js
bluwy: I'm not exactly sure what you mean about "unit test from a `.ts` file", but I'm guessing creating a `.ts` endpoint and set up an integration test for it? Not sure how a unit test would work with a `.ts` file with our current setup 😅

bholmesdev: @bluwy Yes, exactly. I've set up some JSON endpoint unit tests before, and they're honestly simpler than our HTML tests. [Example here.](https://github.com/withastro/astro/blob/main/packages/astro/test/content-collections.test.js#L18-L19) This uses the devalue package to parse not-JSON-friendly properties, but you could use `JSON.parse()` for these tests instead.

delucis:
Is there any concern about introducing this API before we can offer something similar for MDX? Is the intention that we might one day figure this out for MDX too? Render components but exclude scripts & styles?

packages/astro/test/content-collections-render.test.js
bluwy: 83b0815 should do it. Thanks for the tip!

bluwy:
After further discussion with the maintainers, we decided to close this for now until we have a proper way to expose `html` for MDX files too. This avoids confusion when a user switches `.md` to `.mdx` that a property is missing. Closing the docs PR too.

oliverpool:
I would really like to have this feature for my RSS feed (I don't use mdx at all).

Wouldn't it be reasonable to make `html` a function (`renderResult.html = mod.compiledContent` in case of `.md`) which throws a (nice) error in case of mdx? (maybe pointing to some issue here about `proper way to expose html for MDX files`).

For now I have to use `import.meta.glob` on the feed page (which return different objects than `getCollection` and without proper type)

OverflowCat:
So why isn't `compiledContent()` present in MDX? I have extensively used MDX and components in my blog, but most of the pages are static. Only a few pages contain <script> tags.

bholmesdev:
@OverflowCat In short, because MDX needs to be "executed" as Astro components, and we need an API to safely render Astro components to strings of HTML. This is being [tracked here](https://github.com/withastro/roadmap/issues/533)!

PR #6112, opened 2023-02-03
Fix remote image example

BryceRussell:
## Changes

The remote image examples for `@astrojs/image` do not have a `format` prop and do not mention `format` as a requirement for remote images

- Mentions format as a requirement for remote images
- Add required `format` prop to remote image examples
- Switch remote image url from a Google logo to a Astro logo
- Remove redundant example

## Testing

No tests, docs change

## Docs


Princesseuh:
AFAIK, format is only a requirement if you're using Squoosh, so it might be better to add that it's required to the format property only

petermekhaeil:
It may be worth combining the content on these 2 sections:
- https://docs.astro.build/en/guides/images/#image-
- https://docs.astro.build/en/guides/integrations-guide/image/#examples

I was puzzled why my image wasn't loading until I found on the other page that the `format` was required.

BryceRussell:
@Princesseuh Thanks for the recommendation. I tested it out and you are right, `format` is only required when using squoosh. I added a mention of this in the `format` prop section.

I think it would still be good to have the `format` props on the component examples in the "Remote Images" section. Squoosh is the default transformer and it may confuse people using this default that a remote image example has no `format` prop

Also this is my first PR that is more than a typo so let me know if I did anything wrong

natemoo-re (approved):
This looks good to me!

Princesseuh (approved):

delucis (approved):
LGTM2!

PR #6113, opened 2023-02-03
astro with tailwind incorrect style css order in subpath 

### LINKED ISSUE #5676: astro with tailwind incorrect style css order in subpath

1.7.2

Netlify

yarn

Mac

### Describe the Bug

css files (with tailwind)  in subpath (like http://website/subfolder) are imported in the wrong order and custom styles get overridden

it only happen when building for production, local dev is correct, I tried with deno and netlify

it works correctly with astro 1.6.14

i set up a sample in netlify:

Compare: 
https://verdant-kashata-5415d3.netlify.app/test/
https://verdant-kashata-5415d3.netlify.app/

=> TEST H1 in https://verdant-kashata-5415d3.netlify.app/test/
 is not styled with global css, it is overridden, it should be like in https://verdant-kashata-5415d3.netlify.app/

source code : https://github.com/herve-ch/astro-tw

https://github.com/herve-ch/astro-tw

### Participation

wulinsheng123:
## Changes
fix #5676

@matthewp 
I have found out how to fix this bug, but this module was a bit complicated. 
I'm not sure what a better idea was.

### DIFF
--- a/.changeset/thirty-queens-help.md
+++ b/.changeset/thirty-queens-help.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix Cannot resolve aliased module from src/content/config.ts
--- a/packages/astro/src/core/build/graph.ts
+++ b/packages/astro/src/core/build/graph.ts
@@ -26,7 +26,7 @@ export function* walkParentInfos(
 		if (seen.has(imp)) {
 			continue;
 		}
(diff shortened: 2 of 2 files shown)


wulinsheng123:
> Thanks @wulinsheng123 We'll want to get a test in too. I'll see if I can add that.

I will add a unit test if you confirm with this pr.

matthewp:
Hm, yeah I don't think this is the right solution. Going to close and work on another method to fix.

Issue #5996, opened 2023-01-26
Upgrading astro, astro-storyblok, and netlify adapter results in build error in Netlify (SSR)


SandraRodgers:
### What version of `astro` are you using?

2.0.0

### Are you using an SSR adapter? If so, which one?

Netlify

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

I updated my project to Astro 2.0. I'm using the netlify adapter for SSR. The dependencies have also been updated:

```
"dependencies": {
    "@astrojs/netlify": "^2.0.0",
    "@astrojs/tailwind": "^3.0.0",
    "@storyblok/astro": "^2.0.5",
    "astro": "^2.0.1",
    "tailwindcss": "^3.2.4"
  }
```
For netlify builds, Netlify is running Node v16.19.0. 

When I run it locally, the site builds fine. In storyblok I can edit and publish and see the updates, both the visual editor and on my localhost. 

Previously, it would build fine when I pushed to main (before I upgraded to Astro 2.0)

Now, when I push my updated project to main, the netlify deployment happens. It builds the site, but results in an error. 

Attaching screenshot of error and build logs where related warning happens

<img width="853" alt="Screen Shot 2023-01-26 at 12 33 06 PM" src="https://user-images.githubusercontent.com/45321563/214935963-a703270e-2073-45bf-8f78-a370e7ba366b.png">

<img width="1195" alt="Screen Shot 2023-01-26 at 1 52 07 PM" src="https://user-images.githubusercontent.com/45321563/214936360-53d7e6a3-e283-4b67-9fc9-444aef6cd874.png">

### Link to Minimal Reproducible Example

https://github.com/SandraRodgers/astro-storyblok-example

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
@Princesseuh when we removed Node 14 support did was stop shimming `fetch` and `Headers` globals? We might need to bring that back if so, since we document using things like `new Headers()` in code.

Princesseuh:
> @Princesseuh when we removed Node 14 support did was stop shimming `fetch` and `Headers` globals? We might need to bring that back if so, since we document using things like `new Headers()` in code.

No, we still provide those through undici. https://github.com/withastro/astro/blob/main/packages/webapi/src/ponyfill.ts#L74-L78

natemoo-re:
[...]

PR #6117, opened 2023-02-03
Fix polyfills not being available to imports on Netlify

Princesseuh:
## Changes

Previously, we'd always externalize SSR adapters, kinda as a coincidence more than a need, but we did. This made it so our import was always first in the end bundle, in turn making it so our polyfills were always applied in time. Adding `astro` as a peer deps of the adapters broke this, as they're now getting bundled.

This PR adds an escape hatch for packages to mark themselves as needing external, despite triggering the heuristics to mark Astro-related packages as not-external

Fix https://github.com/withastro/astro/issues/5996

## Testing

Tested manually. I tried adding a test for this for almost 2 hours and couldn't figure it out, because of the monorepo the deps works differently in a way that Vite don't see them correctly and it was super annoying,

## Docs

N/A, bug fix


natemoo-re (approved):

packages/integrations/netlify/package.json
```diff
@@ -23,6 +23,8 @@
     "./netlify-functions.js": "./dist/netlify-functions.js",
     "./edge-functions": "./dist/integration-edge-functions.js",
     "./netlify-edge-functions.js": "./dist/netlify-edge-functions.js",
+    "./polyfill.js": "./dist/polyfill.js",
+    "./polyfill": "./dist/polyfill.js",
```
bluwy, line 27: Are these two lines needed in this change? I can't seem to find the new file in dist

Princesseuh: Oops, left over from a previous attempt at making this work. Good catch, will update

packages/integrations/netlify/src/netlify-functions.ts
```diff
@@ -1,12 +1,9 @@
-import { polyfill } from '@astrojs/webapi';
+import '@astrojs/webapi/polyfill-ssr.js';
```
bluwy, line 1: IIUC netlify [handles dependencies this way](https://docs.netlify.com/functions/build/?fn-language=ts#manage-dependencies), if the build build script uses pnpm, and since `@astrojs/webapi` isn't a direct dep in a user's project by default, there's a chance it might break? 🤔 

Maybe we should re-export the polyfill from `@astrojs/netlify/polyfill` like the other (removed) change.

Princesseuh: That's what I tried first, but then I couldn't get the external to work, it would either crash with a rollup build error or just not work. I couldn't figure it out 😦

If you have a solution, I'll gladly take it!

bluwy: Ah I see the issue too. Getting:

```
[vite]: Rollup failed to resolve import "@astrojs/netlify/polyfill" from "/Users/bjorn/Work/oss/astro/packages/integrations/netlify/dist/netlify-functions.js".
```

Which should technically work but I'm not sure why not (maybe it's a monorepo issue). Maybe Matthew's idea of injecting the import in `generateBundle` or `renderChunk` for the output entry would be the way.

packages/integrations/netlify/package.json
```diff
@@ -50,5 +50,8 @@
     "cheerio": "^1.0.0-rc.11",
     "mocha": "^9.2.2",
     "vite": "^4.0.3"
+  },
+  "astro": {
+    "external": true
```
natemoo-re, line 55: 😞 Not sure how I feel about this... Maybe it would be better to hardcode the list if they're all first-party?

Princesseuh: The problem is that they're not necessarily first party, any SSR adapter can need it

matthewp: We talked about this and said we weren't going to document it for now. If the concern is about having an API to maintain; I had the same concern.

.changeset/silent-dragons-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
```
matthewp, line 2: this needs an `'astro': patch` as well since an astro change is required.

natemoo-re:
```suggestion
'@astrojs/netlify': patch
'astro': patch
```

packages/integrations/netlify/package.json
natemoo-re: Cool, think we're all on the same page. Not ideal but it makes sense. Resolving my comment.

natemoo-re (approved):

Issue #6101, opened 2023-02-02
@astrojs/image: support for SVG files missing


ggounot:
### What version of `astro` are you using?

2.0.5

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

SVG files can be used through `<img />` but the `@astrojs/image` integration doesn't permit it through `<Image />`.

A `*.svg` module declaration should be added in `@astrojs/image/client.d.ts` (as it exists in `astro/client.d.ts`) and metadata extraction support for SVG files should be added so that `<Image />` can generate an `<img />` tag with `width` and `height` attributes. The SVG file itself doesn't have to be converted.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rgnkk6?file=src/pages/index.astro

### Participation

- [x] I am willing to submit a pull request for this issue.

ggounot:
Fixed in @astrojs/image@0.15.0

gkatsanos:
@ggounot can we actually use the image integration and INLINE svg's instead if having them as <img />'s?

gkatsanos:
@ggounot I dont think so.. what would be the example code for this?

gkatsanos:
sorry, I think I may not have been clear:
import and svg and INLINE it in the html:

<html>
... <svg> <path> whatever </path> </svg>

ggounot:
@gkatsanos I don't think you need the image integration to do that. But I don't have much experience with Astro so I invite you to ask on the Discord directly.

gkatsanos:
Indeed, using Fragment and set:html it's done! Thank you

PR #6118, opened 2023-02-03
Add support for SVG images to  `@astrojs/image`

ggounot:
This fixes issue https://github.com/withastro/astro/issues/6101.

## Changes

Add support for SVG images to `@astrojs/image`:
- add `'*.svg'` module declaration;
- add `svg` in `InputFormat` and `OutputFormat`;
- update `SharpService.transform()` and `SquooshService.transform()` to directly return the input image if the format is `svg`.

This allows importing SVG files and displaying them through `<Image />`.

Sharp and Squoosh cannot output SVG so no transformation is performed if the output format is SVG.

The SVG's metadata can be read and used to generate the right attributes in the resulting ``.

## Testing

Added an SVG image in the `basic-image` fixture and tests for it in `image-ssg.test.js`, `image-ssr-build.test.js` and `image-ssr-dev.test.js`.

## Docs

Added information on SVG format usage in `@astrojs/image`'s `README.md`.


packages/integrations/image/README.md
```diff
@@ -169,14 +169,16 @@ Set to an empty string (`alt=""`) if the image is not a key part of the content
...
-**Type:** `'avif' | 'jpeg' | 'png' | 'webp'`<br>
+**Type:** `'avif' | 'jpeg' | 'jpg' | 'png' | 'svg' | 'webp'`<br>
 **Default:** `undefined`
 </p>
 
 The output format to be used in the optimized image. The original image format will be used if `format` is not provided.
 
 This property is required for remote images only, because the original format cannot be inferred.
 
+The `svg` format is only valid when the original image is already in SVG format. The image itself won't be processed but the resulting `<img />` will have the right attributes.
```
sarah11918, line 180: I'm tempted to make this a note, since this format works differently from all the others. When using the `svg` option:
- the original image must be in SVG format already
- the image integration cannot transform a non-SVG image into an SVG
- remote SVG images are not "processed" (does this mean optimized? or does this mean "changed to this format"? Processed could mean different things.) If the image is not optimized by the integration, that's another difference when setting the type to `svg`.

This seems like enough to call out as something that really is "not like the others" and might even by helpful to mention when/why you would choose to use the image integration for an svg, instead of choosing just an `` tag. It's not clear to me from this that the image integration does much for you here.  :smile:

ggounot: Thank you @sarah11918 for looking into this. :)

Yes, by "processed" I meant transformed/optimized.

The image integration not only transforms the images but also reads the width and height of the image to put them into the `width` and `height` attributes of the `` tag, which optimizes the cumulative layout shift (CLS). 

Transformation with SVG output is not possible but CLS optimization is, so the image integration is still useful for SVG images.

Moreover, making `<Image />` accept all image formats that work with `` allows the developer to not worry about SVG's specific case when using a bunch of images of various formats.

I hope I was clearer. I can add more details to the README if necessary.

ggounot: I updated the README. Hope it's better now :)

TechAurelian: > The image integration not only transforms the images but also reads the width and height of the image to put them into the `width` and `height` attributes of the `` tag, which optimizes the cumulative layout shift (CLS).

@ggounot You are absolutely right; I am reading more and more about how specifying `width` and `height` attributes on `` tags is so important now.

(I was also trying to do this on a site built with Astro, but as all images (binary and SVGs) are already optimized, I was only trying to automatically extract the width and height of each image and add them to the `` tag. BTW, what are you using to extract the SVG width and height in your code?)

Princesseuh (approved):
Sorry for the wait, we've been discussing this internally making sure it's something we wanted.

This looks great, thank you very much for doing all the work (code, documentation and tests? In one PR? That's awesome)

PR #6120, opened 2023-02-03
update node integration README with Fastify code example

edv-ns:
## Changes

Adding astrojs/node + Fastify minimal code example to the docs, with the same functionality as express example has.

## Docs

Will be automatically generated inside docs repo.

### DIFF
--- a/packages/integrations/node/README.md
+++ b/packages/integrations/node/README.md
@@ -98,6 +98,27 @@ app.use(ssrHandler);
 app.listen(8080);
 ```
 
+Or, with Fastify (>4):
+
+```js
+import Fastify from 'fastify';
+import fastifyMiddie from '@fastify/middie';
+import fastifyStatic from '@fastify/static';
+import { fileURLToPath } from 'url';
+import { handler as ssrHandler } from './dist/server/entry.mjs';
+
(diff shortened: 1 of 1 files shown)


edv-ns:
Ready to review.

matthewp:
Didn't Fastify remove `.use()`? https://github.com/fastify/fastify/pull/3506

natemoo-re (approved):
LGTM but also cc'ing @matthewp who built [`astro-fastify`](https://github.com/matthewp/astro-fastify)

edv-ns:
@matthewp yes, Fastify removed `.use()` and do not support middleware, but there are two packages `@fastify/express` and `@fastify/middie` which gives the Express-like syntax to use node as middleware.

ref: https://www.fastify.io/docs/latest/Reference/Middleware/

I have chose `@fastify/middie` because of mentioned improved performance, while `@fastify/express` is not meant to be used for long-term (see readme: https://github.com/fastify/fastify-express) as it is meant for transition period.

sarah11918 (approved):
The README updates are fine by docs!

matthewp:
@edv-ns Got it, thanks that helps a lot.

PR #6123, opened 2023-02-03
fix example Card component style

malteehrlen:
Card component has uneven border on hover, and in some cases the border is visible even when not hovering. This commit fixes those two issues. 
<details>
<summary>uneven border</summary>

uneven_border
</details>
<details>
<summary>background visible even when not hovering</summary>

background_visible
See the right border, a thin line appears when not hovering
</details>

## Changes

For the example Card component styling, the following was changed:
* increase border size slightly so symmetry can be achieved
* hide background image and only show it on hover

<details>
<summary>appearance after fix</summary>

issues_fixed
</details>

## Testing

No tests added, this is a css change for one of the example components.

## Docs

No docs added, this is a css change for one of the example components.

### DIFF
--- a/.changeset/neat-taxis-thank.md
+++ b/.changeset/neat-taxis-thank.md
@@ -1,5 +1,0 @@
----
-'astro': patch
----
-
-Fix hover styling of Card example component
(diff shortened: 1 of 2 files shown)


malteehrlen:
The failing tests seem unrelated to the changes, could that be?

natemoo-re (approved):
Thanks for the fix!

PR #6126, opened 2023-02-04
fix(portfolio): keyboard a11y for ThemeToggle

marcysutton:
## Changes

Fixed keyboard accessibility for the Portfolio example's ThemeToggle.

Making the focus outline transparent for the ThemeToggle button “except for Windows High Contrast Mode” makes this component inaccessible to keyboard users. Removing the CSS declaration allows the default focus outline to show.

The example previously failed WCAG 2.4.7 Focus Visible (Level A in WCAG 2.2).

Before (with Chrome DevTools open):

Portoflio page with focus forced on ThemeToggle, showing a transparent outline by default

After (focus is on the ThemeToggle button):

Cropped shot of the header on the Portfolio example, with focus visible on the ThemeToggle

## Testing

Tested locally with the keyboard in-browser before and after the change. I also ran the e2e tests but all the failures looked unrelated (framework/cross-platform errors).

### DIFF
--- a/examples/portfolio/src/components/ThemeToggle.astro
+++ b/examples/portfolio/src/components/ThemeToggle.astro
@@ -19,8 +19,6 @@ import Icon from './Icon.astro';
 		background-color: var(--gray-999);
 		box-shadow: inset 0 0 0 1px var(--accent-overlay);
 		cursor: pointer;
-		/* Outline visible only to high contrast users */
-		outline: 1px solid transparent;
 	}
 
 	.icon {


natemoo-re (approved):
Thank you so much for submitting a PR!

Yan-Thomas:
Thanks for finding this @marcysutton, I really appreciate your work and would love to see more contributions from you! 

For context, I was the one who added this line when reviewing the new portfolio template, I was using StackBlitz to test the changes I was making and probably forgot to add part of it as a suggestion in the PR, so I didn't notice this issue (as it didn't happen in my ~machine~ StackBlitz instance).

The transparent outline is a "hack" to have a border in High Contrast without causing layout shifts (it also has better compatibility, considering the `forced-colors` media query doesn't work in all browsers), this was done to make clear that it is a toggle component rather than two different buttons and avoid confusion.

My original intention was to have these lines in addition to the transparent outline:

```css
button:focus {
	outline: 2px solid white;
	outline-offset: 4px;
}
```

Resulting in the following focus states:

| Default  | High Contrast (ON) |
| Focused ThemeToggle component  | Focused ThemeToggle component in High Contrast Mode |

I'd love to know if adding this and bringing back the transparent outline would raise other a11y issues or if it's OK, we use the same technique for the ThemeToggle in our Error Overlay and would love to fix it in case there's something else I'm missing regarding accessibility. 🙌

PR #6149, opened 2023-02-06
Move pagination error to AstroErrorData

bloycey:
## Changes

This PR moves the error message in the `paginate()` function into `AstroErrorData` for consistency.

image

## Testing

Manually tested using the `blog` demo project. Can add test file if necessary.

## Docs

I haven't been terribly creative and have kept the error message as is. However I believe that this refactor means that it will appear on the docs so I would appreciate some eyes on this from the doc team 🙏

### DIFF
--- a/.changeset/breezy-worms-retire.md
+++ b/.changeset/breezy-worms-retire.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Moved pagination error to AstroErrorData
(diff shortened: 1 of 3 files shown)


Yan-Thomas (approved):
Nice work @bloycey, thanks for this. Docs LGTM!

natemoo-re (approved):
Code LGTM, thanks for contributing!

PR #6150, opened 2023-02-06
Add getStaticPaths type helpers to infer params and props

morellodev:
## Changes

- Adds `InferGetStaticParamsType` to infer params returned by `getStaticPaths()`
- Adds `InferGetStaticPropsType` to infer props returned by `getStaticPaths()`

## Testing

This is a type-only addition, no tests are needed.

## Docs

We probably need to update our docs in order to make people aware of these type helpers.


Princesseuh (approved):
> Curious if @Princesseuh has thoughts on this! Is this the right entrypoint or should these go in astro/types?

I'm a bit torn because while the new HTMLAttribute stuff is in `astro/types`, all the other getStaticPaths stuff is in `@types/astro`, so, not sure. I think either way is fine and it doesn't matter much. 

Ultimately, we should definitely infer this automatically in our language tooling, but this is a nice addition in the meantime! Awesome stuff

natemoo-re (approved):

matthewp:
Thanks @morellodev, we are hoping to get this out for the 2.1 release.

matthewp:
Docs are up at https://github.com/withastro/docs/pull/2765

Issue #6121, opened 2023-02-03
Content collection MDX styles missing from prod builds for non-page components


bholmesdev:
### What version of `astro` are you using?

2.0.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm, pnpm

### What operating system are you using?

Mac

### Describe the Bug

When using content collections, MDX styles are missing in production builds when rendering from a non-page `.astro` component.

1. Create an MDX content collection entry
2. Import styles from that entry
3. Try to render that entry from a component _that is not_ the `src/pages` component (any level of nesting will work).
4. Notice styles are present in dev, but not production build previews.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-jjdbqw?file=src/components/UsingMdx.astro,src/content/blog/using-mdx.mdx,src/pages/index.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

bholmesdev:
Note the above example includes a hoisted script test as well. Hoisted scripts _are_ correctly included, but I left this in for testing after the hoisted script PR goes in!

Issue #6057, opened 2023-01-31
MDX files with components insert script in layout slot


agrahamg:
### What version of `astro` are you using?

2.0.4

### Are you using an SSR adapter? If so, which one?

none

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Scripts from Astro Components are placed in the layout slot instead of in head when using MDX

This is a new issue since 2.0.2

### Link to Minimal Reproducible Example

https://github.com/agrahamg/astro-example

### Participation

- [ ] I am willing to submit a pull request for this issue.

tombennet:
Looks like this was introduced by #6034 in 2.0.3.

For me it doesn't just affect scripts from components, but also the site's global CSS generated using the `astrojs/tailwind` plugin.

In 1.9.2, stylesheets were injected in both head and body (see #5715). Duplication issue is now fixed, but stylesheets appear at the top of the componont in the body, meaning a flash of unstyled content before the browser finds and downloads the resources.

z336:
+1 here

### What version of `astro` are you using? 
Astro @ 2.0.4
@astrojs/mdx @ 0.15.2

### Are  you using an SSR adapter? If so, which one? 
no

### What package manager are you using? 
npm

### What operating system are you using? 
Mac

### Describe the bug
Stylelsheet and script tags are injected into the body of pages using MDX components. 

### Link to Minimal Reproducible Example
This page has MDX components and is showing the behavior: https://astro-docs-prototype.netlify.app/docs/en/tutorial/a-tutorial/

This page does not have MDX components and is injecting the tags into the head correctly: https://astro-docs-prototype.netlify.app/docs/en/tutorial/a-second-tutorial/

This is my repo: https://github.com/z336/astro-docs-prototype

minht11:
I am experiencing same issue of styles and scripts being inserted into body and not head, but while using [astro-remote](https://github.com/natemoo-re/astro-remote/tree/main/packages/astro-remote)


```
<Markdown
      components={{
	p: Paragraph,
      }}>
**Removing p component fixes the problem**
</Markdown>

// Paragraphp.astro
<p>
  <slot />
</p>
```

[Reproduction](https://stackblitz.com/edit/github-pcqbox?file=src/pages/index.astro) run `npm run build` and then `npm run preview`

Not quite sure if this is the same issue, because `astro-remote` independently parses markdown, but the end result is the same and both are related to markdown, let me know if i need to report separate bug for this.

[...]

PR #6152, opened 2023-02-06
Add additional scoping for head buffering

matthewp:
## Changes

- This is a continuation of https://github.com/withastro/astro/pull/6107
- The idea is that when rendering to keep track of the scope with which a render occurs. We want to know that we are rendering JSX, or we are rendering Astro, or if we are rendering for head buffering.
- The conditional head injection uses this information to determine if it should render head content or defer. 
- We know that when we are rendering JSX inside of an Astro layout that the Astro layout will do head injection, so we should not do it eagerly in the JSX content which will result in head content in the body.
- Closes https://github.com/withastro/astro/issues/6057
- Closes https://github.com/withastro/astro/issues/6121

## Testing

- New test case as provided by https://github.com/withastro/astro/issues/6057

## Docs

N/A, bug fix.


matthewp:
!preview mdx-links

matthewp:
!preview mdx-links

packages/astro/package.json
```diff
@@ -99,7 +99,7 @@
     "test:e2e:match": "playwright test -g"
   },
   "dependencies": {
-    "@astrojs/compiler": "^1.0.1",
+    "@astrojs/compiler": "next--slot-pass",
```
matthewp, line 102: use a release

matthewp:
@bluwy let me explain the flow for future readers then:

1. When calling JSX, Slots, head buffering (used to know what CSS to propagate), and slots.render(), append bit flags to `result.scope`.
2. *Also* created a scoped version of the `result` object (ie, `Object.create(result)`) so we can "lock in" `result.scope` so it applies only to children and not siblings.
3. In `maybeRenderHead` logic, use the bit flags to determine if we should render the head. The idea is that there are known scenarios where we should not; for example inside of a slot that's inside of JSX. 

This all became necessary because of a combination of:

1. Implicit head injection.
2. Content collections and head propagation. 

This code *works* but does add some bloat to the codebase that would be great to get rid of. For that reason I'm going to create a proposal that removes implicit head injection (likely in combination with a new feature like head propagation).

Issue #6102, opened 2023-02-02
Maximum call stack size exceeded with big site


jdmr:
### What version of `astro` are you using?

2.0.5

### Are you using an SSR adapter? If so, which one?

none

### What package manager are you using?

pnpm

### What operating system are you using?

mac

### Describe the Bug

I'm trying to migrate a big website from Gatsby to astro. I add a folder with 500 images and videos (376MB) in the public folder, I then try to build the project and I get the following:

06:59:56 AM [build] output target: static
06:59:56 AM [build] Collecting build info...
06:59:56 AM [build] Completed in 34ms.
06:59:56 AM [build] Building static entrypoints...
06:59:58 AM [build] Completed in 1.86s.

 generating static routes
▶ src/pages/index.astro
  └─ /index.html (+4ms)
Completed in 7ms.

 error   Maximum call stack size exceeded
![astro](https://user-images.githubusercontent.com/148428/216334305-2fa54962-d82e-42db-b505-6621f595d7e5.jpg)

### Link to Minimal Reproducible Example

https://git.swau.edu/astro/swau

### Participation

- [ ] I am willing to submit a pull request for this issue.

AirBorne04:
Ok, i did not have a problem with my own test case with many files.

I identified 2 issues related to file naming here:
- this file named [#:~:text=...](https://git.swau.edu/astro/swau/-/blob/master/public/assets/%23%3A~%3Atext%3D...) is problematic
- files named `John %26 Joan Curnow.webp` and `Paul L. %26 Doroth` are also problematic, as the copy process encodes the name and decodes it again and therefore turns the `%26` into the `&` and than can't find that file when trying to copy.

Changing those names makes the build run successful!

jdmr:
Thank you Daniel!

matthewp:
Looks like you got it worked out. As there is nothing actionable to be done here, going to close.

jdmr:
Thank you!

AirBorne04:
Well there is actually something, that might need a fix, i just could not wrap my head around a solution yet. During the file copy the file path from fs.readDirSync is passed to new Url(path, ..) and than converted with urlToFilePath, which is fine generally. Here we had an edge case were the filename contained ' %26 ' which got converted to '%20%26%20' (by new Url) and than converted to ' & ' (by urlToFilePath). Than that new filename could not be found for processing.

PR #6154, opened 2023-02-06
Updates to core fs operations

AirBorne04:
## Changes

- rewrite emptyDir + removeEmptyDir functions from 'core/fs' to use async functions and run in loop instead of recursive to prevent #6102 maximum callstack errors
- fixing the url encode/decode for file copy #6102 
- adding tests for special chars in file names

## Testing

- adding tests for empty folder cleanup
- adding tests for special chars in file names

## Docs

just a fix/cleanup does not impact user behaviour


matthewp:
This is a little stale, so closing. Please submit a new one when/if it's ready, thanks!

PR #6157, opened 2023-02-06
Fix duplicate alt attribute on Picture component.

chasm:
## Changes

In `packages/integrations/image/components/Picture.astro` on line 44, `{...image}` already includes the `alt` attribute, so including it again creates a duplicate which causes the output to fail HTML validation: `<img alt="" alt="" ...`

Removing `{alt}` (see changes below) solves the problem.

## Testing

Do you really want to add a test to test for duplicate attributes? The current tests should confirm that the `alt` attribute is not missing, which should be sufficient to prove that the `{alt}` code mentioned above was superfluous.

I ran the tests (thank you for using pnpm!) and they passed. I'm sure you'll run them, too.

## Docs

There is no reason for this to be discussed in the docs. It's a simple mistake, and this PR fixes it.

### DIFF
--- a/.changeset/tidy-buses-mate.md
+++ b/.changeset/tidy-buses-mate.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/image": patch
+---
+
+Fix duplicate `alt` attribute on Picture component.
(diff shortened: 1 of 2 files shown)


chasm:
This is so tiny that it might be easier to just go make the change yourselves.

natemoo-re:
> This is so tiny that it might be easier to just go make the change yourselves.

Nah, you made the code, you should get credit! Thanks for contributing!!

natemoo-re (approved):
Looks great! Just waiting for CI to pass 😊

chasm:
You’re welcome.

Issue #6145, opened 2023-02-06
Build breaks when importing CSS as inline


santicros:
### What version of `astro` are you using?

2.0.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

MacOS 12.6

### Describe the Bug

When trying to import a CSS file inline, it works on development, but it breaks on build.

Vite supports importing CSS inline with the `?inline` param: https://vitejs.dev/guide/features.html#disabling-css-injection-into-the-page like this:

```js
import otherStyles from './bar.css?inline' // will not be injected into the page
```

When doing this on Astro, it works in development mode, but when doing build it breaks, saying:

```sh
 Cannot find module '/home/projects/github-5xrgxx/dist/chunks/index.83f085fc.2f5bdb15.mjs' imported from /home/projects/github-5xrgxx/dist/chunks/pages/all.c4bb7b53.mjs
```

I personally need this to share some styles to a canvas and embed them inline. I need to import a stylesheet as text and use it inside JavaScript. Thanks!

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-5xrgxx?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
I ran into this as well so I'm biased in favor of fixing. :-D

matthewp:
Pretty sure this is a Vite bug, css @bluwy. Looks like it's deleting the chunk because it's "pure". But this should not be treated as pure.

PR #6161, opened 2023-02-07
Prevent ?inline and ?raw CSS from being bundled as CSS

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6145

## Testing

- Test added for `?inline` and `?raw` usage.

## Docs

N/A, bug fix


bluwy (commented):
Was about to suggest the incorrect regex test, but that's fixed now. LGTM once the tests pass.

Also, according to the issue, is this still a Vite bug? I don't think Vite is removing pure css chunks for `?inline` 🤔 

Also, I noticed the `cssRe` check can be replaced with `import { isCSSRequest } from 'vite'` since Vite 4.0. We could clean that up here, or I'm fine tackling that in a separate PR too.

matthewp:
@bluwy does the vite version of `isCSSRequest` exclude `?inline` and `?raw`? If so I agree that it would be good to use that.

This could technically happen in Vite but it would require that an `?inline` CSS module be in the same bundle as another (non-inline) CSS bundle and without any JS. It happened to us because we control manualChunks to force all CSS into the common bundle. But it wouldn't happen in normal Vite, again unless someone was using manualChunks.

bluwy:
> This could technically happen in Vite but it would require that an `?inline` CSS module be in the same bundle as another (non-inline) CSS bundle and without any JS. It happened to us because we control manualChunks to force all CSS into the common bundle. But it wouldn't happen in normal Vite, again unless someone was using manualChunks.

Ah I see. So what happened before is that all types of CSS imports are combined as a CSS chunk, even if they're not really CSS like for `?inline` and `?raw`. This _is_ kinda an edge case, maybe it's worth taking a look in Vite still, but I think this PR as the fix for now is great too.

matthewp:
Yep, that's exactly right. I can't think of a scenario where this would happen in regular Vite without someone manipulating manualChunks like we were. In this case the manualChunks part is the error, we don't actually want this type of CSS in that same chunk. If it were it would prevent it from being split into a CSS file!

bluwy (approved):

nicholasray:
@matthewp It seems like styles imported with the `?raw` query param are being bundled in the dev server. For example, in the following codesandbox, I'd expect the background to not be red since those styles are imported with `?raw`, but it is red:

https://codesandbox.io/p/sandbox/rough-shape-l7t6qk?file=%2Fpackage.json&selection=%5B%7B%22endColumn%22%3A20%2C%22endLineNumber%22%3A13%2C%22startColumn%22%3A20%2C%22startLineNumber%22%3A13%7D%5D

I noticed this behavior began with Astro 2.0.7. Is that intended?

bluwy:
Seems like a bug to me. Feel free to open an issue so we can track it. It could also likely be a Vite bug but I'm not sure yet.

nicholasray:
@bluwy Thank you. I created https://github.com/withastro/astro/issues/6422

Issue #6162, opened 2023-02-07
Hydrated framework components in MDX in content collections break `astro build`


delucis:
### What version of `astro` are you using?

2.0.6 (also tested 2.0.2)

### Are you using an SSR adapter? If so, which one?

n/a

### What package manager are you using?

npm/pnpm

### What operating system are you using?

macOS / Stackblitz

### Describe the Bug

There seems to be an issue resolving paths when bundling hydration scripts for framework components used in MDX in content collections. Running into this as part of the docs migration — https://github.com/withastro/docs/pull/2546.

Minimal example that triggers an error when rendered via `<Content />` during `astro build` (`astro dev` works fine):

**src/content/posts/post.mdx**

```mdx
---
title: MDX post with hydrated component
---

import Component from '../../components/PreactComponent';

<Component client:load />
```

Example error stack trace from the docs migration branch:

```
Error: Cannot find the built path for /.../src/components/tabs/Tabs
    at Object.resolve (file:///.../node_modules/.pnpm/astro@2.0.6_lo4g36gygrjg64j6jydpsd7m54/node_modules/astro/dist/core/build/generate.js:252:15)
    at generateHydrateScript (file:///.../dist/chunks/astro.615ddc00.mjs:951:48)
    at renderFrameworkComponent (file:///.../dist/chunks/astro.615ddc00.mjs:1667:24)
    at async renderChild (file:///.../dist/chunks/astro.615ddc00.mjs:1029:11)
    at async RenderTemplateResult.[Symbol.asyncIterator] (file:///.../dist/chunks/astro.615ddc00.mjs:716:7)
    at async renderAstroTemplateResult (file:///.../dist/chunks/astro.615ddc00.mjs:724:20)
    at async renderToString (file:///.../dist/chunks/astro.615ddc00.mjs:757:20)
    at async renderJSXVNode (file:///.../dist/chunks/astro.615ddc00.mjs:1241:31)
    at async Promise.all (index 16)
    at async renderJSX (file:///.../dist/chunks/astro.615ddc00.mjs:1206:10)
```

My best hunch so far is this has to do with path resolution. Notably the example above will work if we replace the relative path with an alias, e.g.:

```js
import Component from '~/components/PreactComponent';
```

However it goes deeper than this. In the reproduction the hydration script fails to build even if the Preact component is imported indirectly via an Astro component (which is what we do in docs):

**src/content/posts/post.mdx**

```mdx
import AstroComp from '~/components/AstroComp.astro'
```

[...]

PR #6168, opened 2023-02-07
Fix mixed usage of aliases and relative for client hydration

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6162
- Projects that use relative imports for client components mixed with aliases can run into a problem where the build is unable to find the client component for hydration.
- Underlying cause is we assumed a 1-1 relationship between final resolved id and the specifier used. This fixes that false assumption.

## Testing

- Updated the alias test to add a case where we had relative usage.

## Docs

N/A, bug fix


delucis:
Looks plausible to me! Thanks for looking into this @matthewp.

bholmesdev (approved):
Just what we aliasked for 👏

PR #6171, opened 2023-02-08
Suggest correct path when incorrect import path supplied

bloycey:
## Changes

This PR adds a small DX improvement to the error screen that displays when an incorrect import path has been supplied. 

When a broken path is found we test a series of similar paths to see if they find the correct file. If we find a match we ask `Did you mean [file]?`.

image

This solution presumes that most path errors are due to the prefix being incorrectly typed (e.g. `./` instead of `../`). Thus, the focus of this PR is trying various other prefixes to see if we can discover a path that works.

This solution isn't perfect in that it won't account for drastically incorrect paths (if the desired file is many levels away), however this trade-off means that the we don't need to do a repo-wide search - keeping the code relatively straightforward.


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -461,7 +461,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		title: 'Could not import file.',
 		code: 4001,
 		message: (importName: string) => `Could not import \`${importName}\`.`,
-		hint: 'This is often caused by a typo in the import path. Please make sure the file exists.',
+		hint: (suggestedPaths: string[]) => suggestedPaths.length > 0 ? `Did you mean ${suggestedPaths.join(" or ")}?` : 'This is often caused by a typo in the import path. Please make sure the file exists.',
```
natemoo-re, line 464: Dynamic message/hints always trip me up so cc @Princesseuh for another set of eyes.

packages/astro/src/core/errors/dev/vite.ts
```diff
@@ -187,3 +191,18 @@ export async function getViteErrorPayload(err: ErrorWithMetadata): Promise<Astro
 		return errorName.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
 	}
 }
+
+function findSuggestedImportPaths(failedImportString: string, currentFilePath: string): string[] {
+	const suggestedPaths: string[] = [];
+	const prefixes = ["../../", "../", "./", "/", "~/"]
```
natemoo-re, line 197: I don't think `~/` is always supported, assuming that's a custom alias?

packages/astro/src/core/errors/dev/vite.ts
```diff
@@ -187,3 +191,18 @@ export async function getViteErrorPayload(err: ErrorWithMetadata): Promise<Astro
...
+	const importPrefix = prefixes.find(prefix => failedImportString.startsWith(prefix))
+	const prefixesToTry = prefixes.filter(prefix => prefix !== importPrefix)
+	prefixesToTry.forEach(newPrefix => {
+		const modifiedImportPath = importPrefix ? failedImportString.replace(importPrefix, newPrefix) : `${newPrefix}${failedImportString}`
+		const fullPath = path.join(path.dirname(currentFilePath), modifiedImportPath)
+		if(fs.existsSync(fullPath)){
+			suggestedPaths.push(`\`${modifiedImportPath}\``)
+		}
```
natemoo-re, line 205: Does this only handle incorrect prefixes (`./` instead of `../`)? 

Would really love if we could support fuzzy finding as well, but that's probably non-trivial. We can always tackle that in a follow-up.

natemoo-re (commented):
Nice work, I can't wait for this to land!

Had a few questions and would love another core reviewer before signing off on this!

packages/astro/src/core/errors/errors-data.ts
Princesseuh: No problem on hint, only messages are used to build the docs 👍 I still generally recommend to avoid too much logic in the messages themselves, but I have no issues with the snippet here

packages/astro/src/core/errors/dev/vite.ts
Princesseuh: I was wondering if it wouldn't be faster / easier to list all the files in `src` and sort by Levenshtein distance? I believe Astro depends on `fast-glob` already, so it shouldn't be too hard to list the files.

bloycey: My initial plan was to do fuzzy find the files, but I was unsuccessfully trying to get the vite `import.meta.glob` to work. `fast-glob` might be the solution. I'll take another look!

packages/astro/src/core/errors/dev/vite.ts
bloycey: Removed e3388d8082d4823a3fa638cd75a099a0d098af26 👍

packages/astro/src/core/errors/dev/vite.ts
bloycey: Just an update on where I am with this. I tried using `fast-glob` and it does a good job of finding files. The current issue, however, is that I don't know how to determine the **component** that the error is coming from - only the page route (through the `filePath` variable in `vite.ts`). I need the component so I can suggest the right path relative to the component making the request. Any suggestions here?

bloycey: > Does this only handle incorrect prefixes (`./` instead of `../`)?

Yes, so currently it operates on the assumption that the error is _probably_ in the prefix. So for an erroring request to `./global.css` we would try `../../global.css`,  `../global.css`, and `/global.css` and see if we find a `global.css` file at any of those locations. If so, we ask `did you mean [file]`, otherwise we display the default existing error message.

Admittedly this approach is a little limited in that respect, however I think it can catch the low-hanging fruit with a fairly small footprint.

natemoo-re: Ooh, yeah I'm not sure if we can surface the actual source file rather than the current route, but obviously we'll need that info. Otherwise I'd say we're blocked here?

bloycey: Yeah, if it's not possible to get the component path then we are blocked for the robust solution (although my limited solution still works 😛).

Feel free to close the PR if this one isn't going ahead at this stage.

natemoo-re: Thanks for opening the PR! We'll have to circle back to this at some point, I would really love to add this feature.

PR #2517, opened 2022-02-01
Remove docs from monorepo

jonathantneal:
## Changes

- Removes `docs` from this project.
- Docs migrated to https://github.com/withastro/docs
- Migration notice added to `docs` directory.
- References to `docs` are removed.
- References to `docs` are updated in PR template.

PR #6183, opened 2023-02-08
Fix root directory spread pagination `url.prev` for first page

### LINKED ISSUE #6147: Root-directory pagination `url` values use `''` instead of `'/'` for first page

2.0.4

n/a

pnpm

macOS / Stackblitz

### Describe the Bug

When using the `getStaticPaths()` `paginate` helper, each page includes a `url` object for providing URLs to the previous, current, and next pages (`{ current, prev?, next? }` — see [API docs](https://docs.astro.build/en/core-concepts/routing/#complete-api-reference)).

Pagination works by using `/` for page 1, and then `/2` `/3` etc. for subsequent pages. But when paginating in a `src/pages/[...page].astro` file (as opposed to in a subdirectory), `url.current` on page 1 and `url.prev` on page 2 are `''` instead of `'/'` so can’t be used cleanly.

This works as expected in subdirectories, e.g. `src/pages/nested/[...page].astro` produces `/nested`, `/nested/1` etc.

Workarounds include:

- Catching the empty string case like this: `<a href={url.prev || '/'}>`
- setting `trailingSlash: 'always'`

But as in #4426, it would probably make sense for this to always use `/` for this edge case.

https://stackblitz.com/edit/github-gy6h4v?file=src/pages/[...page].astro

### Participation

Jutanium:
## Changes

- Closes #6147
- Closes https://github.com/withastro/docs/issues/2517
- When we're in "spread mode" (i.e., `includesFirstPageNumber` is set to `false`), and we're returning the `prev` value that refers to the first page, return an empty string. This gets `/` appended to it.
Previously, we returned `undefined`, which caused the bug.

## Testing

- Created an `astro-pagination-root-spread` test fixture

## Docs

Removes the need for [#2517](https://github.com/withastro/docs/issues/2517)

### DIFF
--- a/.changeset/calm-readers-learn.md
+++ b/.changeset/calm-readers-learn.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes the first-page value of `url.prev` when paginating a spread route at the root
--- a/packages/astro/src/core/render/paginate.ts
+++ b/packages/astro/src/core/render/paginate.ts
@@ -63,9 +63,7 @@ export function generatePaginateFunction(routeMatch: RouteData): PaginateFunctio
 									: routeMatch.generate({
 											...params,
 											page:
-												!includesFirstPageNumber && pageNum - 1 === 1
-													? undefined
-													: String(pageNum - 1),
+												!includesFirstPageNumber && pageNum - 1 === 1 ? '' : String(pageNum - 1),
 									  }),
 						},
 					} as Page,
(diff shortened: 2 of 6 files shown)


delucis (commented):
Not familiar with this bit of the core codebase, but the test looks solid! Thanks @Jutanium 🙌

matthewp:
Looks great, thanks!

matthewp (approved):

Issue #5602, opened 2022-12-13
Node SSR with Express JSON middleware fails on POST


erg208:
### What version of `astro` are you using?

1.6.15

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

In the Node/Express SSR app we need to use the Express JSON middleware like so.
```
  app.use(express.json()) // for parsing application/json
```

We need this for other Express middleware to work, like for Azure AD SSO. When we do any call to a POST end-point, it does not complete with a response.
Found if we change in https://github.com/withastro/astro/blob/main/packages/astro/src/core/app/node.ts The following code in the class NodeApp render function from:
```
			let body = Buffer.from([]);
			let reqBodyComplete = new Promise((resolve, reject) => {
				req.on('data', (d) => {
					body = Buffer.concat([body, d]);
				});
				req.on('end', () => {
					resolve(body);
				});
				req.on('error', (err) => {
					reject(err);
				});
			});
```
To:
```
			let body = Buffer.from([]);
                        let reqBodyComplete = null;
                        if (req.body) {
                          reqBodyComplete = new Promise((resolve, reject) => {
                            body = Buffer.from(JSON.stringify(req.body));
                            resolve(body);
                          });
                        } else {
                          reqBodyComplete = new Promise((resolve, reject) => {
                            req.on("data", (d) => {
                              body = Buffer.concat([body, d]);
                            });
                            req.on("end", () => {
                              resolve(body);
                            });
                            req.on("error", (err) => {
                              reject(err);
                            });
                          });
                        }
```
Doing this allowed us to utilize the Express JSON middleware and the also allows the Astro end-points have access to the request data. We hope this is a feasible change and would not break anything else.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/erg208-node-astro-ssr

### Participation

- [X] I am willing to submit a pull request for this issue.

erg208:
Also it would be nice to add a reference in the Request Object to the original Node request in the same function. Like so.
```
  let request = new Request(url, {
    method,
    headers: new Headers(entries),
    body: ["HEAD", "GET"].includes(method) ? null : body
  });
  request.req = req;
```

erg208:
[...]

PR #6192, opened 2023-02-09
Fix for Node SSR with Express JSON middleware fails on POST

erg208:
## Changes

- This change is to allow SSR with Node.js and Express.js to utilize Express middleware, especially the JSON one, and not break the Astro end-points that handle a POST request.
- This change is related to [Issue 5602](https://github.com/withastro/astro/issues/5602#issue-1495169317)
- Also this change includes adding the original IncomingMessage (req) as a property of the Request that is created in createRequestFromNodeRequest, so that in Astro end-points there is a way to access the Express Session middleware.

## Testing

- There was no test added, just tested to make sure that the original behavior was not broken.

## Docs

- The documentation might need to be updated to include that in the Request that is available to Astro end-points that the original IncomingMessage request is available as the new property "req".

https://github.com/withastro/docs

### DIFF
--- a/.changeset/spotty-islands-study.md
+++ b/.changeset/spotty-islands-study.md
@@ -1,0 +1,5 @@
+---
+'astro': major
+---
+
+Fix for Node SSR with Express JSON middleware fails on POST
--- a/packages/astro/src/core/app/node.ts
+++ b/packages/astro/src/core/app/node.ts
@@ -23,6 +23,8 @@ function createRequestFromNodeRequest(req: IncomingMessage, body?: Uint8Array): 
 		headers: new Headers(entries),
 		body: ['HEAD', 'GET'].includes(method) ? null : body,
 	});
+	// @ts-ignore
+	request.req = req;
 	if (req.socket?.remoteAddress) {
 		Reflect.set(request, clientAddressSymbol, req.socket.remoteAddress);
 	}
@@ -36,18 +38,27 @@ export class NodeApp extends App {
 	render(req: IncomingMessage | Request, routeData?: RouteData) {
 		if ('on' in req) {
 			let body = Buffer.from([]);
-			let reqBodyComplete = new Promise((resolve, reject) => {
-				req.on('data', (d) => {
-					body = Buffer.concat([body, d]);
+			let reqBodyComplete = null;
+			// @ts-ignore
+      if (req.body) {
+        reqBodyComplete = new Promise((resolve, reject) => {
+					// @ts-ignore
+					body = Buffer.from(JSON.stringify(req.body));
+          resolve(body);
+        });
+      } else {
+				reqBodyComplete = new Promise((resolve, reject) => {
+					req.on('data', (d) => {
+						body = Buffer.concat([body, d]);
+					});
+					req.on('end', () => {
+						resolve(body);
+					});
+					req.on('error', (err) => {
+						reject(err);
+					});
 				});
-				req.on('end', () => {
-					resolve(body);
-				});
-				req.on('error', (err) => {
-					reject(err);
-				});
-			});
-
+			}
 			return reqBodyComplete.then(() => {
 				return super.render(
 					req instanceof Request ? req : createRequestFromNodeRequest(req, body),


packages/astro/src/core/app/node.ts
```diff
@@ -23,6 +23,8 @@ function createRequestFromNodeRequest(req: IncomingMessage, body?: Uint8Array):
 		headers: new Headers(entries),
 		body: ['HEAD', 'GET'].includes(method) ? null : body,
 	});
+	// @ts-ignore
+	request.req = req;
```
matthewp, line 27: why is this needed?

.changeset/spotty-islands-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
```
matthewp, line 2: If I understand correctly, this is a bug fix and not a breaking change. If I'm wrong, what does this break?

packages/astro/src/core/app/node.ts
matthewp: Oh, I guess for this reason:

> Also this change includes adding the original IncomingMessage (req) as a property of the Request that is created in createRequestFromNodeRequest, so that in Astro end-points there is a way to access the Express Session middleware.

We don't want to do that in the official adapter. You are of course welcome to do this in your own adapter. You could even build one that simply wraps the Node adapter.

We do want to find a solution for middleware, see https://github.com/withastro/roadmap/discussions/174

matthewp (changes requested):
Do not want to attach the IncomingMessage to the request.

packages/astro/src/core/app/node.ts
erg208: I have done a local wrapper around the Node adapter to allow the Astro end-points to have more information from the Request provided that is in the original request. Just thought there where others in the community that might have the same need. Sorry, I will remove it and wait for the solution mentioned.

.changeset/spotty-islands-study.md
erg208: Yes, this is a bug fix. I have never did a PR like this in an open source project before. Sorry.

packages/astro/src/core/app/node.ts
matthewp: No problem! Using a wrapper is the way to go, for now.

packages/astro/src/core/app/node.ts
```diff
@@ -36,18 +36,27 @@ export class NodeApp extends App {
...
 			let body = Buffer.from([]);
-			let reqBodyComplete = new Promise((resolve, reject) => {
-				req.on('data', (d) => {
-					body = Buffer.concat([body, d]);
+			let reqBodyComplete = null;
+			// @ts-ignore
+      if (req.body) {
+        reqBodyComplete = new Promise((resolve, reject) => {
+					// @ts-ignore
+					body = Buffer.from(JSON.stringify(req.body));
```
matthewp, line 44: I'm guessing this is what is causing the tests to fail. req.body is not a string here, I imagine.

erg208: The Express JSON middleware is used by several other Express middleware and it parses the POST request data into the req.body. This is how I fixed it to allow the Astro POST end-points to work. Maybe it is not the best way to fix this issue or I could try changing the if statement. Is there there a way for me to just run the unit test that are failing locally?

erg208: I have figure how to run the node unit test locally, so will work on a better fix.

erg208: I changed the if statement to:
if ((typeof(req.body) === 'object') && (Object.keys(req.body).length > 0)) {

matthewp: Are you expecting `req.body` to be a string here? If so just check that directly: `if(req.body === 'string')`...

erg208: With these standard Express middleware being used when a POST request is being processed the data is parsed into the req.body property as an object with properties.
app.use(express.json()) // for parsing application/json
app.use(express.urlencoded({ extended: true })) // for parsing application/x-www-form-urlencoded

matthewp: Ah, I understand now. In that case might I suggest using something like this: https://github.com/sindresorhus/is-plain-obj . That will probably get around the problem you are running into where the body is an object but not one you can stringify.

erg208: I am doing the test for properties "(Object.keys(req.body).length > 0)", because found in the unit testing that req.body is being set to an empty object and causing the unit test to fail. I am not seeing that is-plain-obj has an option to test if an object is plain and not empty.

matthewp: Where does the unit test do that?

erg208: @matthewp I am not familiar with JavaScript unit testing, I am more with Java, I just know when I was debugging while running the test what I was seeing. I will try to find the JavaScript unit test and pin point what was causing the test to fail. Also I have never worked with a JavaScript project of this magnitude and do not know how to install a dependency for a particular module. Any help in these areas would be appreciated.

erg208: @matthewp I am making some progress with the unit test. Found that the unit test packages\integrations\node\test\api-route.test.js on line 20 "let { req, res, done } = createRequestAndResponse({" it calls createRequestAndResponse from import of packages\integrations\node\test\test-utils.js. Then in createRequestAndResponse on line 20 "let req = httpMocks.createRequest(reqOptions);" it calls httpMocks.createRequest from import of node-mocks-http. Looking at the source for [httpMocks.createRequest](https://github.com/howardabrams/node-mocks-http/blob/master/lib/mockRequest.js) on line 73 "mockRequest.body = options.body ? options.body : {};" the Request that is created will always have a body property that is at least an empty object. I know this is a lot to share, but hopefully helps us.

Geoffrey-Pliez:
I have the same issue but in my case body is a string.

So, I suggest the following code

`render(req, routeData) {

    if (typeof (req.body) === 'string' && req.body.length > 0) {
      return super.render(
        req instanceof Request ? req : createRequestFromNodeRequest(req, body),
        routeData
      );
    }

    if ((typeof (req.body) === 'object') && (Object.keys(req.body).length > 0)) {
      return super.render(
        req instanceof Request ? req : createRequestFromNodeRequest(req, JSON.stringify(body)),
        routeData
      );
    }

    if ('on' in req) {
      let body = Buffer.from([]);
      const reqBodyComplete = new Promise((resolve, reject) => {
        req.on('data', (d) => {
          body = Buffer.concat([body, d]);
        });
        req.on('end', () => {
          resolve(body);
        });
        req.on('error', (err) => {
          reject(err);
        });
      });
      return reqBodyComplete.then(() => {
        return super.render(
          req instanceof Request ? req : createRequestFromNodeRequest(req, body),
          routeData
        );
      });
    }

    return super.render(
      req instanceof Request ? req : createRequestFromNodeRequest(req),
      routeData
    );
  }`

erg208:
@Geoffrey-Pliez - Thank you for your code example of how to solve this issue. I have been busy with corporate stuff and finally had a chance to try it out. Will be committing your code as part of this PR.

packages/astro/src/core/app/node.ts
```diff
@@ -34,6 +34,24 @@ export class NodeApp extends App {
 		return super.match(req instanceof Request ? req : createRequestFromNodeRequest(req), opts);
 	}
 	render(req: IncomingMessage | Request, routeData?: RouteData) {
+		// @ts-ignore
```
matthewp, line 37: why is this ts-ignore? If it's the types of `req.body` I'd prefer if you cas to `any`

Issue #6185, opened 2023-02-09
Styles are rendered inside the body, instead of the header in certain combinations using slots


prototypa:
### What version of `astro` are you using?

2.0.9

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Since **Astro** version **2.0.4** something changed that causes for certain uses of slots (https://docs.astro.build/en/reference/api-reference/#astroslotsrender), the main styles (link) are rendered inside tags in the body instead of in the header. This causes page speeds problems with CLS (cumulative shift).

You can check the problem in the following repository: https://github.com/prototypa/astro-styles-in-body-with-slots

Once the build is done, the result of the html page is the following:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  </head>

  <body>
    <main>
      <section class="bg-slate-50">
        <div class="mx-auto text-center max-w-3xl px-4 py-8 md:py-16">
          <h1 class="text-4xl md:text-5xl font-bold leading-tighter tracking-tighter">Lorem ipsum lorem</h1>
          <p class="mt-4 text-xl text-gray-500">
            At vero eos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti
          </p>
          <div class="mt-4">
            <link rel="stylesheet" href="/_astro/good.857198b4.css" />
            <p>
              Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
              dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
              ea commodo consequat.
            </p>
            <p class="mt-4">
              Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
              Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est
              laborum.
            </p>
          </div>
        </div>
      </section>
      <div class="text-center p-4">
        <a class="text-blue-500 underline" href="good">To good example »</a>
      </div>
    </main>
  </body>
</html>
```

[...]

Issue #6181, opened 2023-02-08
When MDX contains an Astro component, script tags are not placed in head


kevinleedrum:
### What version of `astro` are you using?

2.0.9

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

If you have an MDX page that imports an Astro component, all of the `<script>` tags are placed right before that imported component in the DOM instead of inside `<head>`.  It's very strange.  It does not happen with the MDX article in the official blog template that uses a content collection.  I suspect it only happens with a regular file path-based route.  

I was able to create a minimal reproduction. The problem can be seen on the `/test/` page.

![Screenshot 2023-02-08 at 4 03 52 PM](https://user-images.githubusercontent.com/3342530/217651306-9cf11b9e-d2d9-4685-a1d0-cd3dbb64ed05.png)

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/inspiring-goldberg-wuwq0f?on=codesandbox&file=%2Fsrc%2Fpages%2Ftest%2Findex.mdx&selection=%5B%7B%22endColumn%22%3A16%2C%22endLineNumber%22%3A18%2C%22startColumn%22%3A16%2C%22startLineNumber%22%3A18%7D%5D

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
I'll take a look. Meanwhile, if you use a layout this likely won't occur.

kevinleedrum:
This happened on my personal site with a layout, too.  I ended up not using the component and just using html+markdown, so no pressure from me, but I figured I'd report it for posterity.

angelokeirsebilck:
I am encountering the same issue. 

Using mdx with content collection, the .mdx file has imported .astro components and I render the md inside a Layout.astro file using <Content />

matthewp:
@angelokeirsebilck that scenario might be different (likely is). If you could create a stackblitz and share it with me, I'll take a look.

angelokeirsebilck:
> @angelokeirsebilck that scenario might be different (likely is). If you could create a stackblitz and share it with me, I'll take a look.

@matthewp I could not create a stackblitz but managed to make a minimal repo (starting from the blog starter) to reproduce the error.

https://github.com/angelokeirsebilck/mdx-collection-vercel-head-misplacement
https://mdx-collection-vercel-head-misplacement.vercel.app/blog/bug/

[...]

PR #6196, opened 2023-02-09
Fix head injection misplacement with Astro.slots.render()

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6185
- Fixes https://github.com/withastro/astro/issues/6181
- When rendering a head, check if we are within a `Astro.slots.render()`, these should never inject the head unless at the top-level.

## Testing

- Tested for slots in slots (unaffected by this bug, but tested anyways).
- Tested slots directly within Astro.slots.render()
- Tested slots inside of slots inside of Astro.slots.render()
- This should cover every possible way that Astro.slots.render() could mess up head injection.

## Docs

N/A, bug fix


bholmesdev (approved):
Injust get this merged already

PR #6197, opened 2023-02-09
Fix types for TableHTMLAttributes

BryceRussell:
- Add `string` type to `border` attribute for `TableHTMLAttributes`
- Add `'false' | 'true'` types to `frame` attribute for `TableHTMLAttributes`

Thanks @Princesseuh  for helping

## Changes

**What does this change?**

Fix types for TableHTMLAttributes

## Testing

I am not sure if this needs testing

## Docs

I don't think this adds or changes any docs

### DIFF
--- a/.changeset/sixty-scissors-refuse.md
+++ b/.changeset/sixty-scissors-refuse.md
@@ -2,4 +2,5 @@
 'astro': patch
 ---
 
-Fix `TableHTMLAttributes` types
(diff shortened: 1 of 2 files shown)


BryceRussell:
I pushed a small change but not sure if it is still descriptive enough maybe this?

> Fix `border` and `frame` attributes type for `<table>` elements

natemoo-re (approved):
Looks good to me, thanks for updating the changeset!

Issue #6174, opened 2023-02-08
Vercel adapter doesn't work with <Code /> component


Ehesp:
### What version of `astro` are you using?

2.0.8

### Are you using an SSR adapter? If so, which one?

Vercel (^3.0.1)

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

When including the `<Code />` component in pages, the shiki themes are not being uploaded to Vercel:

```
[ssr] Error: ENOENT: no such file or directory, open '/var/task/node_modules/shiki/themes/css-variables.json' (x13)
```

```
<Code
  theme="css-variables"
  lang="js"
  code="console.log('Hello, world!');"
/>
```

See repro link for an example. This works in dev, but fails when deployed to Vercel.

### Link to Minimal Reproducible Example

https://github.com/Ehesp/astro-vercel-shiki-error

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Thanks! It looks like this is due to the fact that shiki dynamically loads json files. I feel like we have solved this issue in the past, let me do some research and see what changed here.

matthewp:
Ah yes, we did previously fix this here: https://github.com/withastro/astro/pull/5409

matthewp:
@Ehesp the reproduction doesn't seem to use Code, or vercel. Can you update it?

Ehesp:
Oh woops I pushed the empty repo. Updated now

matthewp:
Just dropping in to say I can recreate the error. Digging into why our solution isn't working any more.

matthewp:
Pretty sure it's caused by this change: https://github.com/withastro/astro/pull/5952

Working on a fix.

matthewp:
Should be fixed by https://github.com/withastro/astro/pull/6198 Going to deploy a test app to confirm.

matthewp:
Being released now, thanks again for bringing this important issue to our attention @Ehesp 

https://github.com/withastro/astro/actions/runs/4137734219/jobs/7153272973

Ehesp:
@matthewp thanks, although I've just upgraded both `astro` and `@astro/vercel`, and I get another error now:

```
[...]

PR #6198, opened 2023-02-09
Fixes usage of Code component in Vercel

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6174

## Testing

- Created some unit tests for this, so we can better test how we generate options in the future.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/components/Shiki.js
+++ b/packages/astro/components/Shiki.js
@@ -15,14 +15,14 @@ function stringify(opts) {
 
 /**
  * @param {import('shiki').HighlighterOptions} opts
- * @returns {Promise<import('shiki').Highlighter>}
+ * @returns {Promise<import('shiki').HighlighterOptions>}
  */
-async function resolveHighlighter(opts) {
+export async function resolveHighlighterOptions(opts) {
 	const resolvedThemes = [];
-	if (Object.keys(opts.theme).length) {
-		resolvedThemes.push(opts.theme);
-	} else if (opts.theme && opts.theme in themes) {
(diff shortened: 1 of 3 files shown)


matthewp:
!preview shiki-vercel

matthewp:
Working: https://astro-vercel-shiki-error-kljl7rwie-matthew6.vercel.app/

packages/astro/components/Shiki.js
```diff
@@ -15,14 +15,14 @@ function stringify(opts) {
 
 /**
  * @param {import('shiki').HighlighterOptions} opts
- * @returns {Promise<import('shiki').Highlighter>}
+ * @returns {Promise<import('shiki').HighlighterOptions>}
  */
-async function resolveHighlighter(opts) {
+export async function resolveHighlighterOptions(opts) {
```
natemoo-re, line 20: So the `export` here is the fix?

natemoo-re (commented):
Code LGTM, but would love to hear more about how this fix works

packages/astro/components/Shiki.js
matthewp: No, the order swapping is the fix: https://github.com/withastro/astro/pull/6198/files#diff-766f75438aeb2486b5bf2c8f6f3a6521fb48e42b6b51b9be6077cdad5f85c162R22-R25

the export is to unit test this function.

natemoo-re: Ahhh OK thanks!

natemoo-re (approved):

Issue #6206, opened 2023-02-10
Cloudflare mode directory outputs [[path]] to root instead of ./function


vampaz:
### What version of `astro` are you using?

2.0.10

### Are you using an SSR adapter? If so, which one?

Cloudflare 6.1.2

### What package manager are you using?

npm 

### What operating system are you using?

mac/Linux

### Describe the Bug

Expected:
After build the file [[path]].js is saved at ./functions

Result:
After the build the file [[path]].js is saved at the root instead of the expected destination ./functions

When it started:
This was introduced when updating both astro and the cf adaptor(6.1.2) to the latest

How to reproduce:
`npm run build` in the provided stackblitz

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-qbsvtj?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

mfrachet:
I think I m able to reproduce locally, but I don't know if it s a bug or not. Seems related to this LOC: https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/src/index.ts#L208

Also, adding the following assertion:

```js
	it.only('generates functions folder inside the project root', async () => {
		expect(await fixture.pathExists('../functions')).to.be.true;
		expect(await fixture.pathExists('../functions/[[path]].js')).to.be.true; // this one
	});
```

In https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/test/directory.test.js#L19 could safe guard this behaviour.

Also, on a side note, running `npm run build && npm test` in cloudflare folder end up generating the `[[path]].js` file as a sibling of `functions` (probably the behaviour you mention).

I should be able to fix it (let me know if someone else wants to try fixing this one)

PR #6208, opened 2023-02-10
Fix 6206: Cloudflare function path resolving one step above

mfrachet:
Fixes https://github.com/withastro/astro/issues/6206

## Changes

Make sure the `[[path]].js` file is generated inside the `functions` folder in cloudflare.

## Testing

Assertion on a test has been added

## Docs

I don't think it has an impact

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -205,8 +205,9 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				if (isModeDirectory) {
 					const functionsUrl = new URL('functions', _config.root);
 					await fs.promises.mkdir(functionsUrl, { recursive: true });
-					const directoryUrl = new URL('[[path]].js', functionsUrl);
-					await fs.promises.rename(finalBuildUrl, directoryUrl);
+
+					const directoryUrl = new URL('functions/[[path]].js', _config.root);
+					await fs.promises.rename(finalBuildUrl, directoryUrl.pathname);
 				}
 			},
 		},
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -203,10 +203,10 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				}
 
 				if (isModeDirectory) {
-					const functionsUrl = new URL('functions', _config.root);
+					const functionsUrl = new URL('functions/', _config.root);
 					await fs.promises.mkdir(functionsUrl, { recursive: true });
 
-					const directoryUrl = new URL('functions/[[path]].js', _config.root);
+					const directoryUrl = new URL('[[path]].js', functionsUrl);
 					await fs.promises.rename(finalBuildUrl, directoryUrl);
 				}
 			},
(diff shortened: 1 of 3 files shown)


matthewp:
@AirBorne04 can you review?

@mfrachet Please add a changeset with `pnpm changeset`, thanks!

packages/integrations/cloudflare/src/index.ts
```diff
@@ -205,8 +205,9 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				if (isModeDirectory) {
 					const functionsUrl = new URL('functions', _config.root);
 					await fs.promises.mkdir(functionsUrl, { recursive: true });
-					const directoryUrl = new URL('[[path]].js', functionsUrl);
-					await fs.promises.rename(finalBuildUrl, directoryUrl);
+
+					const directoryUrl = new URL('functions/[[path]].js', _config.root);
+					await fs.promises.rename(finalBuildUrl, directoryUrl.pathname);
```
AirBorne04, line 210: This change to pathname seems to cause errors on windows builds.

packages/integrations/cloudflare/test/directory.test.js
```diff
@@ -17,5 +17,6 @@ describe('mode: "directory"', () => {
 
 	it('generates functions folder inside the project root', async () => {
 		expect(await fixture.pathExists('../functions')).to.be.true;
+		expect(await fixture.pathExists('../functions/[[path]].js')).to.be.true;
```
AirBorne04, line 20: Good addition

AirBorne04 (changes requested):
Something weird is ongoing here  at least on windows it does not work.

packages/integrations/cloudflare/src/index.ts
mfrachet: Fixed and pushed, thanks 🙏🏻

packages/integrations/cloudflare/src/index.ts
```diff
@@ -205,7 +205,8 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				if (isModeDirectory) {
 					const functionsUrl = new URL('functions', _config.root);
```
AirBorne04, line 206: actually it is enough to add a trailing slash behind `functions` this you can fix the issue easily

packages/integrations/cloudflare/src/index.ts
```diff
@@ -205,7 +205,8 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 					await fs.promises.mkdir(functionsUrl, { recursive: true });
-					const directoryUrl = new URL('[[path]].js', functionsUrl);
+
+					const directoryUrl = new URL('functions/[[path]].js', _config.root);
```
AirBorne04, line 209: this can revert back to the former version

AirBorne04 (changes requested):
I only now had the change to take a good look at the issue, it seems we were only missing a trailing `/` behind the `functions` path, fixing this should be sufficient.

packages/integrations/cloudflare/src/index.ts
mfrachet: Aww, thanks for the guidance 😊

AirBorne04 (approved):
thank you for the change!

PR #6209, opened 2023-02-10
@astrojs/markdoc

bholmesdev:
https://user-images.githubusercontent.com/51384119/223143111-cf22c71d-c28c-44fa-9637-586e9c6a7f2f.mp4

## Changes

- Implements the Markdoc proposal on our roadmap https://github.com/withastro/roadmap/discussions/478
- Introduce new `@astrojs/markdoc` integration
  - Add Markdoc support to content collections with either `.mdoc` or `.md` files
  - Set up `Content` component to render based on Markdoc `config` and `components` properties
- Add new (private) helper to the `astro:config:setup` integration hook: `addContentEntryType`
  - Migrate Markdown and MDX to use internally
  - Remove hardcoded content collection processing code to make this generic
  - Remove hardcoded `render()` type signature in favor of generated types per-extension. This allows type checking for `Content` props that are specific to Markdoc

## Testing

Add Markdoc + content collections tests for dev and build
[Add `performance/` package](https://github.com/withastro/astro/pull/6209/files#diff-8a940103b8dbb9a37f2ccc1d3abc5b186337082493fcc622257f64ccd23e77b1) with content collection test benchmark
- Introduce `ASTRO_PERFORMANCE_BENCHMARK` flag to disable nonessential remark plugins for Markdown and MDX (GFM, Smartypants, Shiki, heading IDs). This ensures a level playing field comparing against Markdoc.
- Add `build.mjs` script to generate 1000 files of each file type and compare build times
- TODO: compare memory usage

## Docs

- with-markdoc example README
- `@astrojs/markdoc` README


matthewp (changes requested):
Blocking merge as this is pending RFC acceptance.

.changeset/breezy-cats-grab.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': patch
+'@astrojs/markdoc': patch
```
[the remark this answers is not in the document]
matthewp: I think it has to be, yeah. You could set the current version to 0.0.0 if you want the first release to be 0.0.1, but meh, versions are free.

bholmesdev:
!preview markdoc

bholmesdev:
!preview markdoc

bholmesdev:
!preview markdoc

louiss0:
Is there going to be a global config?

matthewp:
This has been moved to stage 2: https://github.com/withastro/roadmap/issues/496

packages/integrations/markdoc/components/RenderNode.astro
```diff
@@ -0,0 +1,30 @@
...
+				<Astro.self node={child} />
+			))}
+		</Node.component>
+	) : (
+		<Fragment>
+			<Fragment set:html={`<${Node.tag} ${stringifyAttributes(Node.attributes)}>`} />
+			{Node.children.map((child) => (
+				<Astro.self node={child} />
+			))}
+			<Fragment set:html={`</${Node.tag}>`} />
```
bholmesdev, line 27: Attempted to clean this up attribute stringifier by calling our JSX runtime instead. i.e. `const Component = jsx(Node.tag, Node.attributes)`. However, calling the JSX runtime creates an infinite render loop in development, whether called within the `astroNode.ts` file or the `.astro` frontmatter 😕 Note sure if @natemoo-re has better context on this.

bholmesdev:
Benchmark findings here! This tests Markdoc against Markdown and MDX across multiple test suites. TL;DR: **Markdoc is blazingly fast** 🔥 
https://gist.github.com/bholmesdev/07ec840d0bcdc24c0114268e16b2d0b0

bholmesdev:
!preview markdoc

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Markdoc files can be used within content collections. Add entries to any content collection using the `.mdoc` extension:
+
+```sh
+src/content/docs/
+  why-markdoc.mdoc
+  quick-start.mdoc
+```
+
+Then, query for these files using the [Content Collection APIs](https://docs.astro.build/en/guides/content-collections/#querying-collections):
```
sarah11918, line 67: Just checking "query for" ? The entire page of Content Collections, we typically say "query your collection" or "query this collection." (We ony say "query for" once, in a code comment.)

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,160 @@
...
+<h1>{entry.data.title}</h1>
+<!--Render Markdoc contents with the Content component-->
+<Content />
+```
+
+📚 See the [Astro Content Collection docs][astro-content-collections] for more information.
+
+## Configuration
+
+You can configure how your Markdoc content is rendered using props via the `Content` component. This component is returned by [a content collection `render()` result](https://docs.astro.build/en/guides/content-collections/#rendering-content-to-html).
```
sarah11918, line 87:
```suggestion
Once the MDX integration is installed, no configuration is necessary to use .mdx files in your Astro project.

You can configure how your Markdoc content is rendered by passing props to the `Content` component.
```
Something like this is more in keeping with the MDX content. We should probably first establish that we're not configuring *the Markdoc integration* itself in any way. Then, we can describe configuring the rendered content specifically. (And, I don't think the second sentence is necessary. It feels more about how content collections work than how this integration does.)

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }}
+/>
+```
+
+### `components` prop
+
+The `components` prop defines mappings from an HTML element name to an Astro or UI framework component (React, Vue, Svelte, etc).
+
+:::note
```
sarah11918, line 116: I'm thinking the note here should go *after* the mapping example, and then should have its own example of creating a wrapper as described. This interrupts the flow of demonstrating what the components prop *can* do.

In fact, instead of a note, if this concept gets its own example, then rather than being an aside, it can be its own h4 section, and a documented example of something people may very well want to do.

sarah11918 (commented):
I'm so looking forward to playing around with this, Ben!  :partying_face: 

Here are some initial thoughts and questions on a first read through. Nothing too major, and with this being experimental, we can absolutely iterate on these docs as people start using the integration and asking questions. But I think a little bit upfront could be helpful.

I think think after addressing these comments (some are clarifying questions), *I* think Team Docs is ready to just take this and run with it? How do you feel about that? I don't think there's much technical here that we'd have trouble with, nor more than the occasional clarifying question we might have as we otherwise mostly go deep into checking for consistency with other integration docs, and other nits. We'd be happy to absolve you of any further responsibility if you're OK with that!

examples/with-markdoc/README.md
```diff
@@ -0,0 +1,50 @@
...
+│   └── content/docs/
+│       └── intro.mdoc
+│   └── components/
+│       └── DocsContent.astro
+│   └── pages/
+│       └── index.astro
+└── package.json
+```
+
+Markdoc can be used in content collections with the `.mdoc` extension. See `content/docs/` for an example.
```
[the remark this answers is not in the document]
bholmesdev: It _cannot_ be used in `src/pages/`. Will make that explicit 👍

examples/with-markdoc/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+You can also apply Astro components and server-rendered UI components (React, Vue, Svelte, etc) to your Markdoc files. See `src/content/DocsContent.astro` for an example configuration.
```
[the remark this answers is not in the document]
bholmesdev: Yes, Markdoc does not use imports! "apply" is a better term I think.

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,160 @@
...
+- <strong>[Usage](#usage)</strong>
+- <strong>[Configuration](#configuration)</strong>
+- <strong>[Examples](#examples)</strong>
+- <strong>[Troubleshooting](#troubleshooting)</strong>
+- <strong>[Contributing](#contributing)</strong>
+- <strong>[Changelog](#changelog)</strong>
+
+## Why Markdoc?
+
+Markdoc allows you to enhance your Markdown with [UI components][astro-ui-frameworks]. If you have existing content authored in Markdoc, this integration allows you to bring those files to your Astro project using content collections.
```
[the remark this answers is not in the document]
bholmesdev: We suggest using Astro components actually! We should update this.

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,160 @@
...
+```js ins={2} "markdoc()"
+import { defineConfig } from 'astro/config';
+import markdoc from '@astrojs/markdoc';
+
+export default defineConfig({
+  // ...
+  integrations: [markdoc()],
+});
+```
+
```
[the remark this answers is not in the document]
bholmesdev: Ah interesting, hadn't added this back. I've personally installed the Markdoc VS Code extension without adding a file association. May need to check if that's also recommended, though adding a note about the extension sounds like a good idea 👍

examples/with-markdoc/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+[![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/github/withastro/astro/tree/latest/examples/with-markdoc)
+[![Open with CodeSandbox](https://assets.codesandbox.io/github/button-edit-lime.svg)](https://codesandbox.io/p/sandbox/github/withastro/astro/tree/latest/examples/with-markdoc)
+
+> 🧑‍🚀 **Seasoned astronaut?** Delete this file. Have fun!
+
+## 🚀 Project Structure
+
+Inside of your Astro project, you'll see the following folders and files:
```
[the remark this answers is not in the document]
bholmesdev: Ah yes, was just focused on documenting the files that deserve clear explanation. Now that this starter is more-or-less stable, we can include all of these.

packages/integrations/markdoc/components/astroNode.ts
```diff
@@ -0,0 +1,78 @@
...
+		return String(node);
+	} else if (node === null || typeof node !== 'object' || !Markdoc.Tag.isTag(node)) {
+		return '';
+	}
+
+	if (isCapitalized(node.name) && node.name in components) {
+		const component = components[node.name];
+		const props = node.attributes;
+
+		const children = node.children.map((child) => createAstroNode(child, components));
```
ematipico, line 35: Can we avoid recursion? Recursions put a lot of pressure on memory, especially on big files.

bholmesdev:
!preview markdoc

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+async function benchmark({ fixtures, templates }) {
+	for (const fixture of fixtures) {
+		const root = new URL(`./fixtures/${fixture}/`, import.meta.url);
+		await generatePosts({
+			postsDir: fileURLToPath(new URL('./src/content/generated/', root)),
+			numPosts: NUM_POSTS,
+			ext: extByFixture[fixture],
+			template: templates[fixture],
+		});
+		console.log(`[${fixture}] Generated posts`);
```
matthewp, line 28: can you add the eslint ignore here?

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+
+		const { build } = await loadFixture({
+			root,
+		});
+		const now = performance.now();
+		console.log(`[${fixture}] Building...`);
```
matthewp, line 34: same

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+		await build();
+		console.log(cyan(`[${fixture}] Built in ${bold(getTimeStat(now, performance.now()))}.`));
```
matthewp, line 36: same

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+
+		const formats = Array.isArray(flags.format)
+			? flags.format
+			: typeof flags.format === 'string'
+			? [flags.format]
+			: ['md', 'mdx', 'mdoc'];
+
+		if (test.includes('simple')) {
+			const fixtures = formats;
+			console.log(
```
matthewp, line 58: eslint ignore

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+					md: 'simple.md',
+					mdx: 'simple.md',
+					mdoc: 'simple.md',
+				},
+			});
+		}
+
+		if (test.includes('with-astro-components')) {
+			const fixtures = formats.filter((format) => format !== 'md');
+			console.log(
```
matthewp, line 73: eslint ignore

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,114 @@
...
+				templates: {
+					mdx: 'with-astro-components.mdx',
+					mdoc: 'with-astro-components.mdoc',
+				},
+			});
+		}
+
+		if (test.includes('with-react-components')) {
+			const fixtures = formats.filter((format) => format !== 'md');
+			console.log(
```
matthewp, line 89: eslint ignore

packages/integrations/markdoc/components/astroNode.ts
bholmesdev: Hm, I'm not sure if we can with this implementation. Markdoc ASTs are represented as nested trees, so we'd need to flatten the tree into a different shape. Not something I've attempted before, but curious if you have suggestions?

packages/astro/src/integrations/index.ts
```diff
@@ -100,11 +101,22 @@ export async function runHookConfigSetup({
 				const exts = (input.flat(Infinity) as string[]).map((ext) => `.${ext.replace(/^\./, '')}`);
 				updatedSettings.pageExtensions.push(...exts);
 			}
+
+			// Semi-private `addContentEntryType` hook
```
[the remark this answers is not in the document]
matthewp: I think he means that it's not actually private, but it is (intentionally) undocumented and therefore not subject to semver. But we know that some might still use it so we're unlikely to break it without careful consideration.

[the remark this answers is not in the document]
bholmesdev: Correct! I copied this type from the `addExtension` comment just above this. I think both could have a fuller explanation 👍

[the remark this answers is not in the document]
bholmesdev: Created `applySemiPrivateHooks` function with @matthewp's comment to clarify!

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,302 @@
...
+  components={{
+    aside: ClientAside,
+  }}
+/>
+```
+
+
+## Examples
+
+*   The [Astro Markdoc starter template](https://github.com/withastro/astro/tree/latest/examples/with-mdx) shows how to use Markdoc files in your Astro project.
```
sarah11918, line 282:
```suggestion
*   The [Astro Markdoc starter template](https://github.com/withastro/astro/tree/latest/examples/with-markdoc) shows how to use Markdoc files in your Astro project.
```

sarah11918 (commented):
Just leaving some nits here for now!

examples/with-markdoc/README.md
```diff
@@ -0,0 +1,59 @@
...
+│   └── pages/
+│       └── index.astro
+|   └── env.d.ts
+├── astro.config.mjs
+├── README.md
+├── package.json
+└── tsconfig.json
+```
+
+Markdoc can be used in content collections _exclusively_ with the `.mdoc` extension. See `content/docs/` for an example file.
```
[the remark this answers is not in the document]
bholmesdev: I wanted to use language here that didn't feel overly discouraging. This makes what you _can't_ do more clear than what you _can_ do. I'd consider removing the "must be located within `src/content/`" piece since it's implied. If you prefer "only" to "exclusively," that's fine!

packages/astro/performance/content-benchmark.mjs
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable no-console */
```
bluwy, line 1: Maybe we can migrate this performance directory to the new `/benchmark` directory at the later stage (happy to help too), but I think it's great to check this performance in at the meantime.

packages/integrations/markdoc/components/RenderNode.astro
bluwy: Maybe it's good to avoid reusing the JSX runtime for now as according to some recent benchmarks we found that JSX rendering is a bit slow.  https://github.com/withastro/astro/pull/6415

packages/integrations/markdoc/components/astroNode.ts
ematipico: I think we can skip it for now then. I thought there was already a working solution that we could use. 
Indeed traversing and rebuilding a tree without recursion is not that simple - but there are already proved solutions [out there](https://www.geeksforgeeks.org/inorder-tree-traversal-without-recursion/).

packages/markdown/remark/src/index.ts
```diff
@@ -64,8 +67,10 @@ export async function renderMarkdown(
 		.use(toRemarkInitializeAstroData({ userFrontmatter }))
 		.use([]);
 
-	if (gfm) {
-		parser.use(remarkGfm);
+	if (!isPerformanceBenchmark) {
```
ematipico, line 70: nit: `if (!isPerformanceBenchmark && gfm) {}`

No need for two nested `if`

packages/markdown/remark/src/index.ts
```diff
@@ -79,18 +84,20 @@ export async function renderMarkdown(
 		parser.use([[plugin, pluginOpts]]);
 	});
 
-	if (scopedClassName) {
-		parser.use([scopedStyles(scopedClassName)]);
-	}
+	if (!isPerformanceBenchmark) {
```
ematipico, line 87: Can't we use only one single `if (!isPerformanceBenchmark) {}` block?

packages/integrations/mdx/src/index.ts
```diff
@@ -26,8 +27,32 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
...
 		name: '@astrojs/mdx',
 		hooks: {
-			'astro:config:setup': async ({ updateConfig, config, addPageExtension, command }: any) => {
+			'astro:config:setup': async ({
+				updateConfig,
+				config,
+				addPageExtension,
+				addContentEntryType,
+				command,
+			}: any) => {
```
ematipico, line 36: Any chance to not use `any` here?

packages/integrations/markdoc/src/index.ts
```diff
@@ -0,0 +1,107 @@
...
+import Markdoc from '@markdoc/markdoc';
+import { prependForwardSlash, getAstroConfigPath, MarkdocError, parseFrontmatter } from './utils.js';
+import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
+
+export default function markdoc(markdocConfig: Config = {}): AstroIntegration {
+	return {
+		name: '@astrojs/markdoc',
+		hooks: {
+			'astro:config:setup': async ({ updateConfig, config, addContentEntryType }: any) => {
```
ematipico, line 13: Any chance to not use `any` here? I've seen that sometimes we use `any` and sometimes we don't. Ideally, we should be consistent and document when we're forced to use `any`

packages/integrations/markdoc/components/astroNode.ts
```diff
@@ -0,0 +1,86 @@
...
+	if (typeof node === 'string' || typeof node === 'number') {
+		return String(node);
+	} else if (node === null || typeof node !== 'object' || !Markdoc.Tag.isTag(node)) {
+		return '';
+	}
+
+	if (isCapitalized(node.name) && node.name in components) {
+		const component = components[node.name];
+		const props = node.attributes;
+		const children = node.children.map((child) => createAstroNode(child, components));
```
ematipico, line 35: Please help me to understand this code:
1. `validateComponents` doesn't have a signature of what's returned; what does return? (I'm not familiar with `zod`)
2. once we have `components =` assigned, we pass it to ALL the children of the node and validate again and again... it seems inefficient. 

	This means that if we have a node with 15 children, we validate the same components 15 times. 

	 Maybe the validation should be moved out of `createAstroNode`, or make sure that we validate them only **ONCE**

packages/astro/src/core/errors/dev/vite.ts
```diff
@@ -150,9 +150,15 @@ export async function getViteErrorPayload(err: ErrorWithMetadata): Promise<Astro
 		: undefined;
 
 	const highlighter = await getHighlighter({ theme: 'css-variables' });
+	let highlighterLang = err.loc?.file?.split('.').pop();
+	if (['cjs', 'mjs'].includes(highlighterLang ?? '')) {
```
ematipico, line 154: nit: you could move this array in a `const` at the top of the file, so it's not recreated every time

packages/astro/src/integrations/index.ts
```diff
@@ -95,16 +96,31 @@ export async function runHookConfigSetup({
...
+					value: addPageExtension,
+					writable: false,
+					enumerable: false,
+				});
+				Object.defineProperty(hooks, 'addContentEntryType', {
+					value: addContentEntryType,
+					writable: false,
+					enumerable: false,
+				});
+			})();
```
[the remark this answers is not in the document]
bholmesdev: Just to make the code self-documenting. I could see people wondering if the function has a reason though. I'll just remove it

packages/integrations/markdoc/CHANGELOG.md
```diff
@@ -0,0 +1,614 @@
+# @astrojs/mdx
```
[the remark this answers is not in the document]
bholmesdev: ack! Total accident. Removing

packages/markdown/remark/src/index.ts
bholmesdev: Not if we want to apply plugins in the right order! If we truncate to a single `if`, plugins outside the `if` will apply out of order. We could implement a sorter by name, but seems a bit much

packages/integrations/markdoc/src/index.ts
```diff
@@ -0,0 +1,107 @@
...
+				const viteConfig: InlineConfig = {
+					plugins: [
+						{
+							name: '@astrojs/markdoc',
+							async transform(code, id) {
+								if (!id.endsWith('.mdoc')) return;
+
+								validateRenderProperties(markdocConfig, config);
+								const body =
+									getEntryInfo({ fileUrl: new URL(prependForwardSlash(id), 'file://'), contents: code }).body;
```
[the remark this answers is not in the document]
bholmesdev: Wasn't sure if `pathToFileUrl` would work in windows, given Vite forces forward slashes for IDs (I think?). I'll give it a shot to see if tests pass 👍

examples/with-markdoc/README.md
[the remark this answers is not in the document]
sarah11918: The real problem is just that the adjective is in the wrong place and doesn't actually convey what you mean to convey. If you want it to sound more positive, I can do that! 

(As written, this is says you can "only" use Markdoc in content collections if you use the `.mdoc` extension. That's not the meaning we're going for.)

packages/integrations/markdoc/components/astroNode.ts
bholmesdev: Good questions!
1. `validateComponents` will return the result of Zod's `parse` function, which will just be an identity function if no transforms are being applied (in this case, there are none). There's arguably no need for the variable assignment because of this, so I'll remove it 👍 
2. You're right, I missed this! We should move the validation step outside of `createAstroNode` to the `Renderer.astro` level before create is called

ematipico (approved):
Great work!

.changeset/breezy-cats-grab.md
```diff
@@ -0,0 +1,7 @@
...
+'@astrojs/mdx': patch
+---
+
+Introduce the (experimental) `@astrojs/markdoc` integration. This unlocks Markdoc inside your Content Collections, and brings support for Astro and UI components in your content. This also improves internals to make Content Collections extensible to more file types in the future.
```
matthewp, line 7: can you expand this changeset a little bit? Like maybe show the `astro add` command and/or show the config diff that adds this.

packages/integrations/mdx/src/index.ts
```diff
@@ -149,14 +185,20 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
...
+const defaultMdxOptions = {
 	extendMarkdownConfig: true,
 	recmaPlugins: [],
-	remarkPlugins: [],
-	rehypePlugins: [],
-	remarkRehype: {},
-};
+}
+
+function markdownConfigToMdxOptions(markdownConfig: typeof markdownConfigDefaults): MdxOptions {
```
bholmesdev, line 193: Removing the `any` type per @ematipico's suggestion uncovered a type error! This refactor ensures all Markdown config options are properly converted _before_ getting passed down to the MDX compiler

.changeset/breezy-cats-grab.md
bholmesdev: Done! Included links to where the docs and `with-markdoc` starter will go as well

matthewp (approved):

examples/with-markdoc/README.md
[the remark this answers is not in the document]
sarah11918:
```suggestion
Markdoc (`.mdoc`) files can be used in content collections to author your Markdown content alongside Astro and server-rendered UI framework components (React, Vue, Svelte, and more). See `src/content/docs/` for an example file.
```
What about something like this to replace the two paragraphs?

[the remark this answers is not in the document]
bholmesdev: That sounds good to me!

Yan-Thomas (approved):
That's amazing @bholmesdev, LGTM!

louiss0:
Be on the lookout for official language server support in Markdoc. [Discussion Here](https://github.com/markdoc/markdoc/discussions/5).

louiss0:
I think this question is best answered with a demo. But I'm wondering how the global config will work with the local component. Will there be an error when it comes to conflicts or will there be an automatic override?

kauhat:
I have been experimenting with the using added  **_private_** `addContentEntryType()` helper on the setup hook to support an additional file type. As far as I can tell this is working well!

I haven't figured out rendering this yet, but it is generating collections of items with the expected slug and metadata.

Here's my usage:
https://github.com/kauhat/astro/blob/cooklang/packages/integrations/cooklang/src/index.ts

.changeset/breezy-cats-grab.md
```diff
@@ -0,0 +1,15 @@
+---
+'astro': patch
+'@astrojs/markdoc': patch
+'@astrojs/mdx': patch
```
bluwy, line 4: Since `@astrojs/mdx` is updated to use the new APIs from next Astro's patch (so previous version won't work), should we mark this a minor to make a breaking release?

bluwy (approved):

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,342 @@
...
+```astro
+---
+import { getEntryBySlug } from 'astro:content';
+import Heading from '../components/Heading.astro';
+
+const entry = await getEntryBySlug('docs', 'why-markdoc');
+const { Content } = await entry.render();
+---
+
+<Content
```
sarah11918, line 195: Should this be passed a `level` prop in this example?

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,342 @@
...
+import { getEntryBySlug } from 'astro:content';
+import Aside from '../components/Aside.astro';
+
+const entry = await getEntryBySlug('docs', 'why-markdoc');
+const { Content } = await entry.render();
+---
+
+<Content
+  components={{
+    Aside: Aside,
```
sarah11918, line 148: Because these are exactly the same name, and both capitalized, I have to admit I'm not sure which `Aside` is which. I know they're "wired" together, but it's not obvious which one is "defining" the other.

a) They don't *have* to have the same name, right? We can call our component `CustomAside` for example? (Which I think is cleaner than trying to write more words to describe the explicit relationship)

b) I think also seeing an example of this being used in a Markdoc file is useful here. What we can see is that we're rendering "all our Markdoc content, with a custom aside figured". I think then seeing a Markdoc snippet is helpful. In their own docs, their tags seem to be lower case (e.g. the`callout` example). But we have two upper case `Asides`, so I might be confused about using a tag.. do I do lower case becase that seems to be how you write Markdoc? Is it upper case because that's how we've wired it here?

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,342 @@
...
+import { getEntryBySlug } from 'astro:content';
+import ClientAside from '../components/ClientAside.astro';
+
+const entry = await getEntryBySlug('docs', 'why-markdoc');
+const { Content } = await entry.render();
+---
+
+<Content
+  components={{
+    aside: ClientAside,
```
sarah11918, line 233: Now I'm left wondering why this `aside` is lower case, when the others are upper case. Is this a typo? Were the others? (tbh, this is more what I would have expected!)

packages/integrations/markdoc/README.md
```diff
@@ -0,0 +1,342 @@
...
+```sh
+# Using NPM
+npx astro add @markdoc/markdoc
+# Using Yarn
+yarn astro add @markdoc/markdoc
+# Using PNPM
+pnpm astro add @markdoc/markdoc
+```
+
+Then, call Markdoc's rendering functions where you plan to define your configuration. This example defines an `abTestGroup` Markdoc variable based on an SSR param, transforming the raw entry `body` to ignore `astro.config.mjs` configuration. The result is rendered using the `Renderer` component provided by `@astrojs/markdoc`:
```
sarah11918, line 298:
```suggestion
Astro recommends running the Markdoc transform step manually for this. This allows you to define your configuration and call Markdoc's rendering functions in a `.astro` file directly, bypassing `astro.config.mjs`.

You will need to install the `@markdoc/markdoc` package into your project.

```sh
# Using NPM
npx astro add @markdoc/markdoc
# Using Yarn
yarn astro add @markdoc/markdoc
# Using PNPM
pnpm astro add @markdoc/markdoc
```
Now, you can define Markdoc configuration options using `Markdock.transform()`.

This example defines an `abTestGroup` Markdoc variable based on an SSR param, transforming the raw entry `body`. The result is rendered using the `Renderer` component provided by `@astrojs/markdoc`:
```

I just felt the 2nd paragraph here was a little overloaded. So I tried to explain the what/why above, then leave this sentence to just explain exactly what is happening below. Something like this?

File: packages/integrations/markdoc/README.md (hunk 36 above)
Comment (bholmesdev) [reply packages/integrations/markdoc/README.md:233]: Just a typo! Should be uppercase

File: .changeset/breezy-cats-grab.md (hunk 33 above)
Comment (natemoo-re) [reply .changeset/breezy-cats-grab.md:4]: I'm a little confused why `@astrojs/mdx` is even included in this changeset that is all about `@astrojs/markdoc`. 

If you could create multiple changesets that describe what actually changed in each package that would be ideal.

File: packages/integrations/markdoc/components/RenderNode.astro (hunk 2 above)
Comment (natemoo-re) [reply packages/integrations/markdoc/components/RenderNode.astro:27]: Another option would be creating this component with our internal JS API, but this seems fine too.

File: packages/integrations/markdoc/src/index.ts [hunk 38]
```diff
@@ -0,0 +1,127 @@
...
+								validateRenderProperties(markdocConfig, config);
+								const body = getEntryInfo({
+									// Can't use `pathToFileUrl` - Vite IDs are not plain file paths
+									fileUrl: new URL(prependForwardSlash(id), 'file://'),
+									contents: code,
+								}).body;
+								const ast = Markdoc.parse(body);
+								const content = Markdoc.transform(ast, markdocConfig);
+
+								return `import { jsx as h } from 'astro/jsx-runtime';\nimport { Renderer } from '@astrojs/markdoc/components';\nconst transformedContent = ${JSON.stringify(
```
natemoo-re, line 66: Are we using the `jsx-runtime` here for any reason other than it having a convenient signature? I would expect this to use our JS API for components.

packages/integrations/markdoc/src/utils.ts
```diff
@@ -0,0 +1,145 @@
...
+	public hint: string | undefined;
+	public frame: string | undefined;
+
+	type = 'MarkdocError';
+
+	constructor(props: ErrorProperties, ...params: any) {
+		super(...params);
+
+		const {
+			code = 99999,
```
natemoo-re, line 48: What is the expectation with these error codes? Would be good to leave a comment.

packages/integrations/markdoc/src/utils.ts
```diff
@@ -0,0 +1,145 @@
...
+const componentsPropValidator = z.record(
+	z
+		.string()
+		.min(1, 'Invalid `components` prop. Component names cannot be empty!')
+		.refine(
+			(value) => isCapitalized(value),
+			(value) => ({
+				message: `Invalid \`components\` prop: ${JSON.stringify(
+					value
+				)}. Component name must be capitalized. If you want to render HTML elements as components, try using a Markdoc node [TODO: DOCS LINK]`,
```
natemoo-re: TODO

.changeset/breezy-cats-grab.md
bholmesdev: Discussed on discord! Sounds like adding a legacy check to make this non-breaking would be the best path 👍

packages/markdown/remark/src/index.ts
```diff
@@ -39,6 +39,9 @@ export const markdownConfigDefaults: Omit<Required<AstroMarkdownOptions>, 'draft
 	smartypants: true,
 };
 
+// Skip nonessential plugins during performance benchmark runs
+const isPerformanceBenchmark = Boolean(process.env.ASTRO_PERFORMANCE_BENCHMARK);
```
natemoo-re, line 43: Don't _love_ this but if there's not a cleaner way to do it, that's fine.

packages/integrations/markdoc/README.md
bholmesdev: Discussed offline. Changing to the shortened JS object syntax (`components={{ Aside }}`) to follow best practices and assume an intermediate programmer knowledge

packages/integrations/markdoc/README.md
bholmesdev: Much better yes! 👏

examples/with-markdoc/.vscode/extensions.json
```diff
@@ -0,0 +1,4 @@
+{
+  "recommendations": ["astro-build.astro-vscode"],
```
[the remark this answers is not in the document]
bholmesdev: Discussed with the Markdoc team. There's an unofficial extension from an individual contributor, though they recommend just using Markdown highlighting for now.

packages/integrations/markdoc/src/utils.ts
bholmesdev: Honestly no clue 😅 We don't have a system for error codes for integrations outside of Astro core. Just used `99999` as a stand-in. I'll leave a comment for now so we can discuss best practices on discord.

packages/markdown/remark/src/index.ts
bholmesdev: Yeah, couldn't come up with a clearer way. Originally had a private CLI flag with prop drilling, but wasn't sure that would be maintainable in the long run. Want to touch base with @bluwy about refactoring this performance package now that his benchmark package is in.

natemoo-re: Cool, resolving for now.

sarah11918 (approved):
Docs is signing off, @bholmesdev! Some lovely work here. 🥳

examples/with-markdoc/.vscode/extensions.json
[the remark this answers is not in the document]
mohitsinghs: Hi @bholmesdev ! I'm that individual contributor. I added basic syntax highlighting but locally I use a work in progress lsp branch which supports formatting, completions and linting. Let me know if I can help. The source lives [here](https://github.com/markdoc-extra/vscode-markdoc)

Issue #6156, opened 2023-02-06
Unhelpful Error: Astro.redirect and Response in child component


Jutanium:
### What version of `astro` are you using?

2.0.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac, Linux

### Describe the Bug

We [document](https://docs.astro.build/en/guides/server-side-rendering/#astrorequestheaders) that you can't modify headers from child components, but doing so presents an unhelpful error that does not link to the documentation.
<img width="1260" alt="image" src="https://user-images.githubusercontent.com/4033662/217088890-388927ec-972b-43a2-9e19-77e6c1016f33.png">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rxpb4p?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

mfrachet:
Hey there, I'm experimenting on the side and I was wondering what should be the expected output after dealing with this issue.

Since the `Test` component is trying to redirect but is also trying to render content, do we expect it to be showing something or to silently be hidden and not showing?

Thanks for the guidance (let me know if other folks want to dive into this one for faster resolution 😊 )

Jutanium:
The `Test` component is returning a Redirect, so it doesn't end up rendering content. This is supposed to fail, but we should fail with a nice error message rather than the generic one we do now.

mfrachet:
Can I try fixing it? Let me know if you prefer dealing with it with the team 😊. (Not wasting time of the team and mine hahaha)

Jutanium:
@mfrachet The platform team marked it as lower priority, so they might not be able to get to it soon. Have at it!

matthewp:
New message is much clearer.

<img width="1438" alt="Screen Shot 2023-06-15 at 4 48 33 PM" src="https://github.com/withastro/astro/assets/361671/7857e576-49ae-4ee5-85da-319334aea302">

PR #6210, opened 2023-02-10
Fix 6156: remove warning when child component redirects

mfrachet:
Fixes https://github.com/withastro/astro/issues/6156

## Changes

Safeguard against a value that could not be defined when converting a chunk into a byte array


matthewp:
Hey @mfrachet This change makes sense. Adding a comment explaining why it's doing this would be good. Also this is in draft status. Since this PR is a bit old I'm going to close, but please open a new one if you get a chance to work on it again.

PR #4138, opened 2022-08-04
make Remark rehype options available in astro config

gtnbssn:
## Changes

- Allows to set remark-rehype options in astro.config.mjs
- See options here https://github.com/remarkjs/remark-rehype#options
- Astro has {allowDangerousHtml: true, passThrough: ["raw", "mdxTextExpression", "mdxJsxTextElement", "mdxJsxFlowElement"]} by default, this keeps them and will extend the values with what goes in the astro.config.mjs
- This should allow to set other options, notably footerLabel to close #3163

## Testing

Add this in astro.config.mjs

```
markdown: {
  remarkRehype: {
    footnoteLabel : "other title for footnotes",
    footnoteBackLabel: 'other text for the link back to content',
  }
}
```

Create a markdown page with a least one footnote and check that the title for the footnotes section has been updated.

## Docs

I'll update (if necessary) and reopen [docs#445](https://github.com/withastro/docs/pull/445) if this passes.

PR #6211, opened 2023-02-10
remove change in type of original Options from remark-rehype

pedro199288:
## Changes

- What does this change?
This PR removes a change that was introduced in the type of original Options from `remark-rehype` package. This was done in this PR https://github.com/withastro/astro/pull/4138  in `packages/markdown/remark/src/types.ts`.
The reason for this PR is that when passing a valid configuration for remark-rehype (that actually works correctly), typescript is returning an error.

**Before the PR change:**
image

The above error is:
```
Type '{ heading: (state: Parameters<Handler>[0], node: Parameters<Handler>[1], parent: Parameters<Handler>[2]) => { type: string; tagName: string; properties: { someProp: string; }; children: never[]; }; }' is not assignable to type '(state: State, parent: MdastNodes) => ElementContent[]'.
  Object literal may only specify known properties, and 'heading' does not exist in type '(state: State, parent: MdastNodes) => ElementContent[]'.
```

**After the PR change:**
image

No error appears

Maybe I missed something from the original PR but as it is is giving a typescript error with a config that actually works.

## Testing

No test was updated because this is just related to typescript

## Docs

Haven't add docs but I can if necessary


pedro199288:
I don't know why that error appear but it continues appearing even if I left the code as it was before 😕

bluwy:
The change makes sense to me, but I'm stumped with the TypeScript error too. Looks like it's a bug there: https://github.com/microsoft/TypeScript/issues/42873

If I do:

```ts
export type RemarkRehype = Omit<RemarkRehypeOptions, ''>
```

instead, which is kinda silly, it works. Maybe you can try this and merge up main? I can't seem to merge main locally, there's some sort of history mismatch.

matthewp:
@pedro199288 do you plan to still work on this?

pedro199288:
@bluwy thanks for the support. What you suggested looks silly, but the fact that it doesn't work without that, also looks silly to me. So I will give a try, thanks! :)

pedro199288:
> @pedro199288 do you plan to still work on this?

Yeah I will give it a third and last try this week or the next one. If don't manage to fix it I'll let you know

pedro199288:
I've tried what @bluwy suggested but it didn't work just with that. I've changed two files in my last commit. Curiously, only doing one of these changes didn't fix the issue with the build step but adding the two changes did so (I don't know why). 

You can see that I set `preserveSymlinks: true` but probably there is a reason why this was with its default to `false`? That "solution" was one of the suggestions in the typescript bug previously mentioned https://github.com/microsoft/TypeScript/issues/42873#issuecomment-1381702742 but I'm not sure if it is a valid solution for this project

matthewp:
@pedro199288 all tests are passing now, are you saying this isn't working? Or just that you're unsure about the option you added?

pedro199288:
@matthewp yeah it works fine for me also. It's just that I'm not sure if the change in tsconfig might be a problem for whatever reason that I'm not aware of. Other than that it looks good to me :)

natemoo-re:
Hey @pedro199288, sorry for the delay on this! This change _seems_ good to me, but I agree that the `"preserveSymlinks": true` and `Omit` changes are a bit concerning.

Happy to try to resolve these if you're still interesting in getting this change in.

pedro199288:
Hey @natemoo-re ! sorry for the late response. But yeah sure, it'd be nice if you can help with this

ematipico:
@pedro199288 I don't think the team has the bandwidth to help you with this PR. It's OK if you aren't interested anymore to this PR

natemoo-re:
We're planning to fix this in https://github.com/withastro/astro/pull/9147 and release in Astro 4.0.

bluwy:
Closing as https://github.com/withastro/astro/pull/9147 is merged

PR #6213, opened 2023-02-10
Update compilation target for Node 16

Princesseuh:
## Changes

Despite dropping support for Node 14, we were still building for that target. This fixes that.

This is (hopefully) the final nail in the Node 14 coffin, as after this, Astro won't include the downlevelling necessary to run on Node 14 at all

This makes Astro slightly smaller: Before: `794kb` and after: `743kb` and the build slightly faster.

## Testing

Hopefully the CI pass 🤞 

## Docs

N/A

### DIFF
--- a/.changeset/wicked-pumpkins-punch.md
+++ b/.changeset/wicked-pumpkins-punch.md
@@ -1,0 +1,30 @@
+---
+'astro': minor
+'@astrojs/prism': minor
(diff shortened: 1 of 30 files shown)


natemoo-re (approved):

matthewp (approved):

delucis (commented):
Needs a changeset?

Issue #5937, opened 2023-01-22
Duplicate DOM elements with conditional rending in preact and using SSR


kymppi:
### What version of `astro` are you using?

1.9.2

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

When using shared state (nanostores) with at least preact components and SSR in astro. Sometimes if the data changes fast enough or you hard-reload the browser tab it renders the component twice. One time with the pre-rendered data and the second time with the correct data from nanostores. I suspect it might be also happening with other UI integrations but haven't tested them yet.

Correctly rendered output:
![image](https://user-images.githubusercontent.com/48528700/213931357-d107db8c-0ae4-4b23-848d-29aa074bbbb2.png)

The unexpected result from the bug:
![image](https://user-images.githubusercontent.com/48528700/213931348-5bf174fb-2de5-475f-a4fd-d1db28cd90fc.png)

I didn't get it to work inside stackblitz, but when using the exact same code locally it sometimes happens 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ypfqeo

### Participation

- [ ] I am willing to submit a pull request for this issue.

lloydjatkinson:
See also the conversation in Discord for some more context - not a new problem it seems! https://discord.com/channels/830184174198718474/1066475401033031700/1066475401033031700

matthewp:
It was easiest for me to recreate by throttling the CPU 4x.

matthewp:
This seems to happen if `userInfo.loggedIn` is true on the first run. What is happening is we render some DOM like this:

```html
<astro-island>
  <astro-slot><h1>Login</h1></astro-slot>
</astro-island>
``

And then our component is defined like:

```js
export const Preact = (props: { children: JSX.Element }) => {
  const userInfo = useStore(user);

  if (userInfo.loggedIn)
    return (
      <div>
        <h2>Preact {userInfo.name}</h2>
      </div>
    );
  else return props.children;
};
```

And we call render like:

```js
render(
  h(Preact, props, h(StaticHtml, { value: '<h1>Login</h1>' }),
  element
);
```

For some reason the diff/patch is not removing the `<astro-slot>`. Digging into why.

matthewp:
I've reached out to the Preact team to understand this better.

PR #6215, opened 2023-02-10
Prevent hydration mismatches in Preact

matthewp:
## Changes

- Fix hydration mismatches by using the 3rd argument form of render().
- This provides the `<astro-island>` as wrapped as a Preact component, for the same of guaranteeing a root match in render. Doing this ensures that children will all be fixed in case of a mismatch.
- Fixes https://github.com/withastro/astro/issues/5937

## Testing

- Tested in the demo app. An automated test is not really possible as it takes CPU throttling to recreate.

https://user-images.githubusercontent.com/361671/218183429-09d0ba58-e7de-44a3-8e83-0a918a106448.mp4

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/orange-chicken-end.md
+++ b/.changeset/orange-chicken-end.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/preact': patch
+---
+
+Prevent hydration mismatches in Preact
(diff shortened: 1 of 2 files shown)


developit:
LGTM!

Issue #6214, opened 2023-02-10
Misplacement of head content in nested .astro components rendered from content collection MDX.


angelokeirsebilck:
### What version of `astro` are you using?

2.0.10

### Are you using an SSR adapter? If so, which one?

Vercel adapter but same problem in Node adapter

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

This repo uses vercel adapter but the same bug occurs when using node adapter.

Rendering nested .astro components misplaces the head links inside the body (right after where you import the nested component)

<img width="641" alt="image" src="https://user-images.githubusercontent.com/42614667/218161999-e0c9c466-f321-4b5d-8896-2853b3445a99.png">

First component in .mdx: https://github.com/angelokeirsebilck/mdx-collection-vercel-head-misplacement/blob/main/src/content/blog/bug.mdx?plain=1#L10

nested astro component: https://github.com/angelokeirsebilck/mdx-collection-vercel-head-misplacement/blob/main/src/components/blocks/home/HomeBanner.astro#L14

Live example: https://mdx-collection-vercel-head-misplacement.vercel.app/blog/bug/

### Link to Minimal Reproducible Example

https://github.com/angelokeirsebilck/mdx-collection-vercel-head-misplacement

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Fixed in https://github.com/withastro/astro/issues/6214

matthewp:
You can try this out with:

```
npm install astro@next--headbody
```

angelokeirsebilck:
Can confirm that this works

PR #6216, opened 2023-02-10
Fix head injection in body with slots.render() and head buffering

matthewp:
## Changes

- Fixes case where `Astro.slots.render()` is called within head buffering. Prevents head content from being rendered in the body instead of the head.
- Closes https://github.com/withastro/astro/issues/6185
- Closes https://github.com/withastro/astro/issues/6214

## Testing

- New test case added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/test/fixtures/head-injection/src/components/UsesSlotRender.astro
+++ b/packages/astro/test/fixtures/head-injection/src/components/UsesSlotRender.astro
@@ -1,0 +1,7 @@
+---
+import SlotRenderComponent from "./SlotRenderComponent.astro";
+---
+
+<SlotRenderComponent>
+    <p slot="slot-name">Paragraph.</p>
+</SlotRenderComponent>
--- a/.changeset/eight-cameras-itch.md
+++ b/.changeset/eight-cameras-itch.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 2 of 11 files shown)


matthewp:
!preview headbody

matthewp:
!preview

matthewp:
!preview headbody

bholmesdev (approved):
Feeling some deja vu, but gotta `slot` in more test cases sometimes

packages/astro/test/fixtures/head-injection/src/components/UsesSlotRender.astro
```diff
@@ -0,0 +1,7 @@
+---
+import SlotRenderComponent from "./SlotRenderComponent.astro";
+---
+
+<SlotRenderComponent>
+    <p slot="slot-name">Paragraph.</p>
```
jamesarosen, line 6: I don't see an assertion that this ever ends up in the DOM anywhere. (That may be OK for this; I'm just calling it out in case it's an accidental omission.)

matthewp: That part doesn't matter here. The slots just exist to create the condition where head was inserted improperly. So we only need to test that the links were injected in the head and not the body.

PR #5874, opened 2023-01-16
feat(@astrojs/netlify): Add on-demand builders Netlify functions 

juanmiguelguerrero:
## Changes

- A new configuration option is added to the Netlify adapter to enable On-Demand Builder feature.

## Testing

- I don't know how to add a test with `chai`.
- In any case, by adding this configuration, it is possible to test the functionality:

```javascript
import { defineConfig } from 'astro/config';
import netlify from '@astrojs/netlify/functions';

export default defineConfig({
  output: 'server',
  adapter: netlify({
    builders: true
  }),
});

```

## Docs

- A new heading "On-Demand Builders" must be added to the "Configuration" section, including the above example and a link to the official documentation (https://docs.netlify.com/configure-builds/on-demand-builders/)

/cc @withastro/maintainers-docs for feedback!

PR #6219, opened 2023-02-11
Add ttl option to netlify adapter for builders

betabong:
## Changes

https://github.com/withastro/astro/pull/5874 added an option to render astro pages via Netlify On-Demand Builders.

- Extends new builders option for Netlify adapter
- Allows to set a TTL (Time to Live) option:
    - As a number for all pages / routes
    - As a function per request

```js
// astro.config.mjs
import netlify from '@astrojs/netlify/functions';

export default defineConfig({
  output: 'server',
  adapter: netlify({
    builders: {
      ttl(event) {
        return event.path.startsWith('/blog') ? 3600 : undefined;
      },
    },
  }),
});
```

## Testing

I honestly don't know how to test this. I was not even able to try it out by myself, as I don't know how to link a git subpackage as a dependency. If anybody can hint me on how to do this, I'm happy to test things with an actual project.

## Docs

@withastro/maintainers-docs I have updated the Readme with instructions.

### DIFF
--- a/packages/integrations/netlify/README.md
+++ b/packages/integrations/netlify/README.md
@@ -13,7 +13,6 @@ Learn how to deploy your Astro site in our [Netlify deployment guide](https://do
 - <strong>[Contributing](#contributing)</strong>
 - <strong>[Changelog](#changelog)</strong>
 
-
 ## Why Astro Netlify
 
 If you're using Astro as a static site builder—its behavior out of the box—you don't need an adapter.
@@ -21,7 +20,6 @@ If you're using Astro as a static site builder—its behavior out of the box—y
 If you wish to [use server-side rendering (SSR)](https://docs.astro.build/en/guides/server-side-rendering/), Astro requires an adapter that matches your deployment runtime.
 
 [Netlify](https://www.netlify.com/) is a deployment platform that allows you to host your site by connecting directly to your GitHub repository. This adapter enhances the Astro build process to prepare your project for deployment through Netlify.
-
 
 ## Installation
 
@@ -40,22 +38,22 @@ If you prefer to install the adapter manually instead, complete the following tw
 
 1. Install the Netlify adapter to your project’s dependencies using your preferred package manager. If you’re using npm or aren’t sure, run this in the terminal:
 
-    ```bash
-      npm install @astrojs/netlify
-    ```
+   ```bash
+     npm install @astrojs/netlify
+   ```
 
 1. Add two new lines to your `astro.config.mjs` project configuration file.
 
-    ```js ins={3, 6-7}
-    // astro.config.mjs
-    import { defineConfig } from 'astro/config';
-    import netlify from '@astrojs/netlify/functions';
+   ```js ins={3, 6-7}
+   // astro.config.mjs
+   import { defineConfig } from 'astro/config';
+   import netlify from '@astrojs/netlify/functions';
 
-    export default defineConfig({
-      output: 'server',
-      adapter: netlify(),
-    });
-    ```
+   export default defineConfig({
+     output: 'server',
+     adapter: netlify(),
+   });
+   ```
 
 ### Edge Functions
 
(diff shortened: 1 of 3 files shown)


packages/integrations/netlify/README.md
```diff
@@ -129,14 +126,31 @@ import netlify from '@astrojs/netlify/functions';
...
   output: 'server',
   adapter: netlify({
-    builders: true
+    builders: true,
   }),
 });
 ```
 
-On-demand Builders are only available with the `@astrojs/netlify/functions` adapter and are not compatible with Edge Functions.
+On-demand Builders optionally let you return a [TTL](https://docs.netlify.com/configure-builds/on-demand-builders/#time-to-live-ttl) (time-to-live) value for the cached page. The astro adapter allows you to set a default TTL value for all pages, or a custom TTL value for specific pages via a function call.
```
Yan-Thomas, line 134: A few wording suggestions:
1. I believe that "Astro adapter" is redundant, considering the context it's clear we're talking about the current adapter, so I would remove "Astro".
2. We're repeating "TTL value" a few times in this sentence, what do you think of simplifying it a bit?
```suggestion
On-demand Builders optionally let you return a [TTL](https://docs.netlify.com/configure-builds/on-demand-builders/#time-to-live-ttl) (time-to-live) value for the cached page. The adapter allows you to set a default TTL value for all pages or a custom one for specific pages via a function call.
```

packages/integrations/netlify/README.md
```diff
@@ -129,14 +126,31 @@ import netlify from '@astrojs/netlify/functions';
...
+    builders: {
+      ttl(event) {
+        return event.path.startsWith('/blog') ? 3600 : undefined;
+      },
+    },
+  }),
+});
+```
 
+On-demand Builders are only available with the `@astrojs/netlify/functions` adapter and are not compatible with Edge Functions.
```
Yan-Thomas, line 153: The sentence above also starts with "On-demand Builders" and I think we should call a bit more attention to it, so I changed it a bit adding "Note that" so people notice this part more!
```suggestion
Note that On-demand Builders are only available with the `@astrojs/netlify/functions` adapter and are not compatible with Edge Functions.
```

Yan-Thomas (commented):
Thanks for this @betabong, I added a few docs related suggestions, let me know your thoughts! 🙌

packages/integrations/netlify/README.md
betabong: @Yan-Thomas Great suggestions, thanks 🙏

juanmiguelguerrero:
@betabong I think it would be good to add an example where the `ttl` is a number. Reading the new doc it's not quite clear how it should be done.

```javascript
// astro.config.mjs
import netlify from '@astrojs/netlify/functions';

export default defineConfig({
  output: 'server',
  adapter: netlify({
    builders: {
      ttl: 3600
      },
    },
  }),
});
```

packages/integrations/netlify/src/netlify-functions.ts
```diff
@@ -8,7 +8,11 @@ polyfill(globalThis, {
 });
 
 export interface Args {
-	builders?: boolean;
+	builders?:
+		| boolean
+		| {
+				ttl?: number | ((event: HandlerEvent) => number | undefined);
```
matthewp, line 14: Unfortunately this kind of API won't work. `Args` gets serialized and included in the build. Your astro.config.mjs does not get bundled with the build. So only plain objects/primitives can be passed into the actual integration. Functions won't work.

How we usually fix this is by instead having the code be defined in a separate module and then building that module as part of the build. Something: `builders: 'path/to/implementation.ts'`. It's not as nice of an API but it prevents us from needing to build/bundle the config file

betabong: @matthewp Oh wow, I wasn't aware of that. Does this only apply to adapter configurations? After all we can use Vite plugins in `astro.config` that again are (partly) functions: https://docs.astro.build/en/reference/configuration-reference/#vite – on the other hand I can't find any example of passing a path to .ts or .js file. If you know of any other usage, that would be very helpful indeed.

matthewp: We don't build the config, so any functions that need to exist in production cannot be provided via a config option.

As for examples, all of the renderers work this way, see for example: https://github.com/withastro/astro/blob/e58a92527f54cb29d3515d544fad833a5ce1061d/packages/integrations/vue/src/index.ts#L13-L17

I'd suggest a similar pattern if you want to allow the user to customize runtime values.

matthewp:
Closing as the current approach doesn't work with SSR. You could just omit the function callback approach, which would be serializable. Or go with a pattern where the user passes a path to a file that gets built. I'll leave that decision up to you.

PR #6229, opened 2023-02-13
fix: add support for `js/mjs` file extensions for Content Collections config file

### LINKED ISSUE #6179: Unsupported config.js file type for content collection schema

2.0.8

None

npm

Stackblitz

### Describe the Bug

When adding a `content/config.js` file to [define a schema](https://docs.astro.build/en/guides/content-collections/#defining-a-collection-schema) for a content collection, I see the following output:

```shell
09:16:39 AM [content] Unsupported file types found. Prefix with an underscore (`_`) to ignore:
- config.js
```

I am able to use the content collection, but it is not clear to me that the schema in the `content/config.js` file is being used.

https://stackblitz.com/edit/github-fba5em?file=README.md

### Participation

MoustaphaDev:
## Changes

Closes #6179

## Testing

Check that an error is thrown when the frontmatter doesn't match the schema of the collection defined in the `config.mjs` file.

I could add a test case for a `config.js` but it seems redundant as the config file has to be a module anyway.

## Docs

Bug fix, we already document this


.changeset/three-peaches-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
MoustaphaDev, line 2: Patch because this feature should already be present as [we document that](https://docs.astro.build/en/guides/content-collections/#defining-a-collection-schema)

bholmesdev (commented):
This code looks great! _However,_ there's a caveat to using `mjs` configs: intellisense will not work unless `"allowJs: true` is present in your `tsconfig.json`, or if you switch to a `jsconfig.json`.

I'm curious if we can check for `allowJs` in your `tsconfig`, and log a warning if it isn't present. This would guide users in the right direction to avoid future support tickets. A callout in our docs would be great too!

MoustaphaDev:
Ah thanks for the heads-up! I found this commit on discord https://github.com/withastro/docs/commit/bb99ae0a8afcf05d70ce57bf16ea034c2c92b9f5 where `allowJs` was removed from the `tsconfig.json` because it would break the `astro:content` module. Did find a way to get around the side effects of `allowJs`?

bholmesdev:
@MoustaphaDev Ah good catch. I've been unable to replicate that issue _outside_ of docs but it's probably worth more testing. I verified `astro:content` works as expected using `allowJs` in the `blog` starter on my end. Mind trying on your end as a sanity check?

MoustaphaDev:
I don't see anything breaking @bholmesdev! Working on implementing the warnings

packages/astro/src/content/server-listeners.ts
```diff
@@ -71,3 +76,51 @@ export async function attachContentServerListeners({
...
+	if (!['info', 'warn'].includes(logging.level))
+		warn(
+			logging,
+			'content',
+			`Make sure you have the ${bold('allowJs')} compiler option set to ${bold(
+				'true'
+			)} in your ${bold(tsConfigFileName)} file to have autocompletion in your ${bold(
+				contentConfigFileName
+			)} file.
+See ${bold('https://www.typescriptlang.org/tsconfig#allowJs')} for more information..
```
bholmesdev, line 98:
```suggestion
See ${bold('https://www.typescriptlang.org/tsconfig#allowJs')} for more information.
```

bholmesdev (approved):
Super small nit but looks amazing! Feeling `config`dent 👏

tschaub:
Thanks for the work on this, @MoustaphaDev!

PR #6235, opened 2023-02-13
add mithril integration

pep108:
## Changes

- Added integration to render Mithril components

## Testing

- An example was created in example/framework-mithril that duplicates the React example and was smoke tested

## Docs

A README.md was included in integrations/mithril with documentation, but not yet added to the docs repo.


pep108:
close because build is failing

Issue #6010, opened 2023-01-27
`@astrojs/image` with SSR doesn't prerender images for prerendered pages


eostrom:
### What version of `astro` are you using?

2.0.2

### Are you using an SSR adapter? If so, which one?

Netlify; also observed on Node

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

With SSR, exporting `prerender = true` from a page generates the HTML at build time. I think `@astrojs/image` should also generate image files at build time.

I noticed this with Netlify: when I switched to SSR, my home page hero image disappeared. Looking at the browser console, it was requesting a URL that started with `/_image`. I opened the URL in a new tab, and found that Netlify was running a function that timed out after 10 seconds.

I'm submitting a Gitpod URL for a [repo that demonstrates the issue](https://github.com/eostrom/astro-image-prerender) with the `@astrojs/node` integration. `yarn preview` isn't working for me on Gitpod – not sure why, but there are instructions in the README for looking in the `dist` folder to see what is and isn't there.

Tested with `@astrojs/image` 0.13.0 and `@astrojs/node` 5.0.1 (and `@astrojs/netlify` 2.0.0).

I'm not checking the "willing to submit a PR" box because I don't know where to start, but if someone can point me in the right direction I'm happy to give it a try.

### Link to Minimal Reproducible Example

https://www.gitpod.io/#https://github.com/eostrom/astro-image-prerender

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
I know neither this part of image integration or the pre-rendering stuff too well, but I'd assume [this part](https://github.com/withastro/astro/blob/45b41d98f50dc9f76a5004a8b3346f393f1a6cb6/packages/integrations/image/src/index.ts#L138-L140) needs to be smarter than just checking if mode is static or not for this to work

rener084:
Same problem. Using Vercel..

etmartinkazoo:
Having the same issue

etmartinkazoo:
Admittedly I haven't dug deeply into @astro/image but would adding an `or` to the conditional like the following help?
```
// Helpers for building static images should only be available for SSG
if (_config.output === 'static' || _config.output === 'ssr') {
  globalThis.astroImage.addStaticImage = addStaticImage;
}
```

etmartinkazoo:
[...]

PR #6237, opened 2023-02-13
Fix static images on prefetch pages for SSR

gshively11:
## Changes

This change fixes the `@astrojs/images` static image build/cache for prefetched pages when using server (SSR) mode. Previously, this was only working for static mode. Now, the build will process and cache static images with server mode for prefetched pages.

This isn't the most graceful change, but it's functional. If there is a better way to handle this, please let me know. I'm still very unfamiliar with astro.

## Testing

I tested this locally with my SSR astro project. Before the change, all image requests were being processed on-demand server-side. With the change, the images are processed at build time and pulled from ~/assets.

## Docs

### DIFF
--- a/.changeset/tasty-pugs-pay.md
+++ b/.changeset/tasty-pugs-pay.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/image': patch
+---
+
+Fix static images for prefetched pages with SSR mode. Images on prefetched pages will now be pulled in their post-processed form from ~/assets, instead of being processed server-side with each request.
--- a/packages/integrations/image/src/build/ssg.ts
+++ b/packages/integrations/image/src/build/ssg.ts
@@ -146,7 +146,11 @@ export async function ssgBuild({
 			inputBuffer = res?.data;
 			expires = res?.expires || 0;
 		} else {
-			const inputFileURL = new URL(`.${src}`, outDir);
+			let inputFileBase = outDir;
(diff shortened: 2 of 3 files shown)


gshively11:
This PR fixes https://github.com/withastro/astro/issues/6010

natemoo-re:
Interesting fix! Thanks for exploring this and opening a PR! 

I think this definitely solves the `prerender` problem for SSR, but I'm curious if there are other unintended side-effects of this change... I'll have to think about this one and get back to you!

matthewp:
Closing as this is breaking tests. Please resubmit if you have the time and can fix them, thanks!

gshively11:
Hey @matthewp, I'd be happy to fix the tests if I can get a general :+1: or :-1: from the astro maintainers as to whether the approach I took here makes sense or not.  I'd like to avoid going through the trouble of fixing the tests to find out the changes I made can't be merged.  FWIW, I've been using this change on my mixed SSR/SSG astro site using the node adapter without issue.

matthewp:
Oh yes, sure thing @gshively11. My only feedback would be that `client/` and `server/` are not hardcoded defaults, so you want to check `config.build.server` and `config.build.client` which have the actual values after user configuration. But aside from that, the approach seems fine to me.

gshively11:
Gotcha, thanks @matthewp.  I'll work on cleaning up and re-submitting this in the next week or so.

Issue #5796, opened 2023-01-07
MDX with rehype-mathjax creates xlinkhref attributes instead of xlink:href


vigonotion:
### What version of `astro` are you using?

1.9.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

I want to render latex equations in md and mdx files. To set this up, my config looks like this:

```
import remarkMath from "remark-math"
import rehypeMathjax from "rehype-mathjax"

// https://astro.build/config
export default defineConfig({
  site: 'https://example.com',
  markdown: {
    remarkPlugins: [
      remarkMath
    ],
    rehypePlugins: [
      rehypeMathjax
    ]
  },
  integrations: [mdx({
    
  }), sitemap(), tailwind()],
});
```

When I create a md file containing this:
	$$
	L = \frac{1}{2} \rho v^2 S C_L
	$$

it renders just fine:

<img width="352" alt="Bildschirm­foto 2023-01-07 um 23 57 23" src="https://user-images.githubusercontent.com/1506970/211173303-b15c7033-48b4-4e8d-a1f9-839345856916.png">

But when I change the file to mdx (which I need), the svg doesn't render correctly anymore:

<img width="416" alt="Bildschirm­foto 2023-01-07 um 23 57 48" src="https://user-images.githubusercontent.com/1506970/211173310-d826d2d8-cd07-47f7-9505-c3343d114a9b.png">

After some debugging I found out that when using mdx, all attributes containing colons are wrong:

	correct (in md):
	<g data-mml-node="mo" transform="translate(958.8,0)">
	    <use data-c="3D" xlink:href="#MJX-3-TEX-N-3D"></use>
	</g>
	
	wrong (in mdx):
	<g data-mml-node="mo" transform="translate(958.8,0)">
	    <use data-c="3D" xlinkhref="#MJX-3-TEX-N-3D"></use>
	</g>

`xlinkhref` is wrong and results in a faulty svg.

Now I'm stuck and don't know where the problem is, but I think it must be in the mdx render pipeline. How can I fix this or how do I debug where the wrong attribute thing even happens?

(came here from the support discord)

### Link to Minimal Reproducible Example

https://github.com/vigonotion/mdx-mathjax-reproduction

### Participation

- [X] I am willing to submit a pull request for this issue.

wassfila:
## Test repo
[...]

PR #6243, opened 2023-02-14
Support rehype plugins that inject namespaced attributes

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/5796

Thanks to https://github.com/mdx-js/mdx/pull/2255, there's a way to properly render namespaced attributes injected by rehype plugins, without Astro needing to manually map camelCase namespaced attributes, to the normal ones.

With this PR, Astro should render `<use xlink:href` instead of `<use xlinkhref`

## Testing

Added test to mdx integration

## Docs

n/a. bug fix.

### DIFF
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -74,6 +74,7 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 									const { data: frontmatter, content: pageContent } = parseFrontmatter(code, id);
 									const compiled = await mdxCompile(new VFile({ value: pageContent, path: id }), {
 										...mdxPluginOpts,
+										elementAttributeNameCase: 'html',
 										remarkPlugins: [
 											// Ensure `data.astro` is available to all remark plugins
 											toRemarkInitializeAstroData({ userFrontmatter: frontmatter }),
--- a/packages/integrations/mdx/test/mdx-plugins.test.js
+++ b/packages/integrations/mdx/test/mdx-plugins.test.js
@@ -63,6 +63,20 @@ describe('MDX plugins', () => {
 		expect(selectRehypeExample(document)).to.not.be.null;
 	});
 
+	it('supports custom rehype plugins with namespaced attributes', async () => {
+		const fixture = await buildFixture({
+			integrations: [
+				mdx({
+					rehypePlugins: [rehypeSvgPlugin],
+				}),
+			],
+		});
+		const html = await fixture.readFile(FILE);
+		const { document } = parseHTML(html);
+
+		expect(selectRehypeSvg(document)).to.not.be.null;
+	});
+
 	it('extends markdown config by default', async () => {
 		const fixture = await buildFixture({
 			markdown: {
@@ -207,6 +221,23 @@ function rehypeExamplePlugin() {
 	};
 }
 
+function rehypeSvgPlugin() {
+	return (tree) => {
+		tree.children.push({
+			type: 'element',
+			tagName: 'svg',
+			properties: { xmlns:"http://www.w3.org/2000/svg" },
+			children: [
+				{
+					type: 'element',
+					tagName: 'use',
+					properties: { 'xlink:href': '#icon' }
+				}
+			]
+		});
+	};
+}
+
 function recmaExamplePlugin() {
 	return (tree) => {
 		estreeVisit(tree, (node) => {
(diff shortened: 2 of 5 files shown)


packages/integrations/mdx/src/index.ts
```diff
@@ -74,6 +74,7 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 									const { data: frontmatter, content: pageContent } = parseFrontmatter(code, id);
 									const compiled = await mdxCompile(new VFile({ value: pageContent, path: id }), {
 										...mdxPluginOpts,
+										elementAttributeNameCase: 'html',
```
wooorm, line 77: Wouldn’t this break React users?

packages/integrations/mdx/test/mdx-plugins.test.js
```diff
@@ -207,6 +221,23 @@ function rehypeExamplePlugin() {
...
+	return (tree) => {
+		tree.children.push({
+			type: 'element',
+			tagName: 'svg',
+			properties: { xmlns:"http://www.w3.org/2000/svg" },
+			children: [
+				{
+					type: 'element',
+					tagName: 'use',
+					properties: { 'xlink:href': '#icon' }
```
wooorm, line 234: The correct hast casing is `xLinkHref`

bluwy: Ah right forgot about that
```suggestion
					properties: { 'xLinkHref': '#icon' }
```

packages/integrations/mdx/src/index.ts
bluwy: The compiled MDX will be rendered by Astro's JSX runtime only so it would be fine. React JSX happens outside of MDX, where they are imported and compiled separately (without MDX).

bholmesdev (approved):
I'll trust MDX fixes this internally! Merge in the `name` of stability

packages/integrations/mdx/src/index.ts
wooorm: Oh right, that answers most of my Qs, but I do wonder whether there’s overlap.

How does the passing of props work between Astro’s own JSX runtime, and React’s JSX runtime? Maybe this point only relates to JSX written by the author tho, and the above option is for generated elements, not components.

Though, does Astro allow component overwrites? (`h1` -> `MyComponent`)? In that case, the `MyComponent` currently gets `className`, and would now get `class`?

matthewp (approved):

packages/integrations/mdx/src/index.ts
bluwy: > How does the passing of props work between Astro’s own JSX runtime, and React’s JSX runtime?

The props should be passed as-is from Astro's JSX to React. Yes this option only affects generated elements from rehype plugins.




> Though, does Astro allow component overwrites? (`h1` -> `MyComponent`)? In that case, the `MyComponent` currently gets `className`, and would now get `class`?

I believe you can with https://docs.astro.build/en/guides/markdown-content/#assigning-custom-components-to-html-elements. I'm not very sure about how the entire MDX flow works, but from a [quick stackblitz test](https://stackblitz.com/edit/github-hhkk2d?file=src%2Fcontent%2Fblog%2Fusing-mdx.mdx,src%2Fcomponents%2FHeading2.astro,astro.config.mjs&on=stackblitz), it looks like `className` is passed down as is.

wooorm: The stackblitz is on the stable release of astro right?
I am worried that this change breaks that

bluwy: Ah you're right. I've tested it locally now and it seems to be the case. We probably can't revert this as it's fixing a bug, I'll make another PR to make this a breaking minor.

It kinda is a bit awkward if the component is a React component, but I think it's a tradeoff to take for now, as the user can still workaround it.

wooorm: This is likely breaking everyone that is using react, and "component overwrites". 
This feature in MDX lets users specify their framework, through two options.
Those options need to be configured by the end user.
If Astro is in the middle, it would *also* need similar functionality: if it wants to hijack things, it needs to deal with different props expected by different frameworks

bluwy: I think it's fine for Astro to take a stance on the default options since you can mix Astro and React components in component overwrites. It's also difficult to know the casing type ahead of time when compiling. Exposing another option isn't bulletproof too since casing wouldn't be right when using an Astro component override. So I think it's worth the tradeoff for now.

PR #6244, opened 2023-02-14
Fix hydration path for circular imports

### LINKED ISSUE #6014: Circular imports evaluated twice client-side?

2.0.2

Node Standalone

npm

Linux

### Describe the Bug

I wanted to use a SolidJS context provider for a component also rendered client-side (with `client:load`).

I noticed that my first approach was failing to get the proper context value, apparently due to a circular import in my component, resulting the in context initialization code [^1] being run twice client-side, hence resolving into different symbols when used in my `ContextProvider` and `SimpleDiv` components (instead of resolving into the same symbol as I was expecting).

The behavior is correct during SSR, but incorrect during CSR.

Moving the context initialization code [^1] to a separate file (imported by both the `ContextProvider` and `SimpleDiv` components solves the issue.

See the console output (by opening the dev tools) from this [reproduction example on Stackblitz](https://stackblitz.com/edit/github-cbhyay-u2rz7y):

```
// Output for ./src/components/without-bug/
[SSR][Good] createContext() call
[SSR][Good] Language: fr

[CSR][Good] createContext() call
[CSR][Good] Language: fr

// Output for ./src/components/with-bug/
[SSR][Bad] createContext() call
[SSR][Bad] Language: fr

[CSR][Bad] createContext() call
[CSR][Bad] createContext() call
[CSR][Bad] Language: en
```

It shows:
- The `createContext()` call (incorrectly?) being done twice client-side when using the `./src/components/with-bug/` components, resulting in the incorrect context being used.
- The `createContext()` call (correctly?) being done once client-side when using the `./src/components/without-bug/` components, resulting in the correct context being used.

I am not sure if it's an expected behavior from the ESM module imports resolution (in this case, sorry for the bug report), or a Vite limitation, or an Astro/SolidJS limitation.

[^1]: Context initialization code: `export const ApplicationContext = createContext([{ lng: 'en' }, {}]);`

https://stackblitz.com/edit/github-cbhyay-u2rz7y

### Participation

bluwy:
## Changes

Fix #6014 

Ensure we load hydration URLs from `/src/Component.jsx`, instead of `/@fs/...`. This is because Vite's import analysis would import from `/src/Component.jsx` instead if a circular import happens. Keeping the same path deduplicates them so only one exist at a time.

## Testing

Added a test following the repro's case

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/core/render/dev/resolve.ts
+++ b/packages/astro/src/core/render/dev/resolve.ts
@@ -1,17 +1,17 @@
 import type { ModuleLoader } from '../../module-loader/index';
 import { resolveIdToUrl } from '../../util.js';
 
-export function createResolve(loader: ModuleLoader) {
+export function createResolve(loader: ModuleLoader, root: URL) {
 	// Resolves specifiers in the inline hydrated scripts, such as:
 	// - @astrojs/preact/client.js
 	// - @/components/Foo.vue
 	// - /Users/macos/project/src/Foo.vue
 	// - C:/Windows/project/src/Foo.vue (normalized slash)
 	return async function (s: string) {
-		const url = await resolveIdToUrl(loader, s);
+		const url = await resolveIdToUrl(loader, s, root);
 		// Vite does not resolve .jsx -> .tsx when coming from hydration script import,
 		// clip it so Vite is able to resolve implicitly.
-		if (url.startsWith('/@fs') && url.endsWith('.jsx')) {
+		if (url.startsWith('/') && url.endsWith('.jsx')) {
 			return url.slice(0, -4);
 		} else {
 			return url;
--- a/packages/astro/src/core/util.ts
+++ b/packages/astro/src/core/util.ts
@@ -161,7 +161,7 @@ export function emoji(char: string, fallback: string) {
  */
 // NOTE: `/@id/` should only be used when the id is fully resolved
 // TODO: Export a helper util from Vite
-export async function resolveIdToUrl(loader: ModuleLoader, id: string) {
+export async function resolveIdToUrl(loader: ModuleLoader, id: string, root?: URL) {
 	let resultId = await loader.resolveId(id, undefined);
 	// Try resolve jsx to tsx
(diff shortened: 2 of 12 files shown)


packages/astro/src/core/render/dev/resolve.ts
```diff
@@ -1,17 +1,17 @@
...
 	// - @/components/Foo.vue
 	// - /Users/macos/project/src/Foo.vue
 	// - C:/Windows/project/src/Foo.vue (normalized slash)
 	return async function (s: string) {
-		const url = await resolveIdToUrl(loader, s);
+		const url = await resolveIdToUrl(loader, s, root);
 		// Vite does not resolve .jsx -> .tsx when coming from hydration script import,
 		// clip it so Vite is able to resolve implicitly.
-		if (url.startsWith('/@fs') && url.endsWith('.jsx')) {
+		if (url.startsWith('/') && url.endsWith('.jsx')) {
```
bluwy, line 14: Changed to a simple slash check, but I think this shouldn't cause issues in practice. This is needed because we need to strip `.jsx` for root-relative paths too, that doesn't start with `/@fs`

packages/astro/src/core/util.ts
```diff
@@ -161,7 +161,7 @@ export function emoji(char: string, fallback: string) {
  */
 // NOTE: `/@id/` should only be used when the id is fully resolved
 // TODO: Export a helper util from Vite
-export async function resolveIdToUrl(loader: ModuleLoader, id: string) {
+export async function resolveIdToUrl(loader: ModuleLoader, id: string, root?: URL) {
```
bluwy, line 164: Ideally the `root` should always be needed, but I find getting this info is tricky for some cases, so I made it optional for now. The other cases should work fine without this fix.

matthewp (approved):
So glad you were able to figure this one out!

bluwy:
To be fair you figured the bug at https://github.com/withastro/astro/issues/6014#issuecomment-1423196337 first which gave a nice head start. I only went in with the fix 😄

PR #6254, opened 2023-02-15
Clarify SVG image gotcha info

Yan-Thomas:
## Changes

Changed the wording about SVG images when using the `format` property to make it more clear to understand, also exemplifying what's a raster image.

## Testing

No code changes 😜 

## Docs

100% docs.

### DIFF
--- a/packages/integrations/image/README.md
+++ b/packages/integrations/image/README.md
@@ -177,7 +177,7 @@ The output format to be used in the optimized image. The original image format w
 
 This property is required for remote images when using the default image transformer Squoosh, this is because the original format cannot be inferred.
 
-> When using the `svg` format, the original image must be in SVG format already (raster images cannot be converted to vector images). The SVG image itself won't be transformed but the final `<img />` element will get the optimization attributes.
+Note that when using the `svg` option, it's expected for the original image to already be in this format, as raster images (like `.png` or `.jpg`) cannot be converted into vectors. The image itself won't be transformed, but the final `<img />` element will get the optimization attributes.
 
 #### quality
 
--- a/packages/integrations/image/README.md
+++ b/packages/integrations/image/README.md
@@ -177,7 +177,7 @@ The output format to be used in the optimized image. The original image format w
 
 This property is required for remote images when using the default image transformer Squoosh, this is because the original format cannot be inferred.
 
-Note that when using the `svg` option, it's expected for the original image to already be in this format, as raster images (like `.png` or `.jpg`) cannot be converted into vectors. The image itself won't be transformed, but the final `<img />` element will get the optimization attributes.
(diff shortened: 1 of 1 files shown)


packages/integrations/image/README.md
```diff
@@ -177,7 +177,7 @@ The output format to be used in the optimized image. The original image format w
 
 This property is required for remote images when using the default image transformer Squoosh, this is because the original format cannot be inferred.
 
-> When using the `svg` format, the original image must be in SVG format already (raster images cannot be converted to vector images). The SVG image itself won't be transformed but the final `<img />` element will get the optimization attributes.
+Note that when using the `svg` option, it's expected for the original image to already be in this format, as raster images (like `.png` or `.jpg`) cannot be converted into vectors. The image itself won't be transformed, but the final `<img />` element will get the optimization attributes.
```
Yan-Thomas, line 180: Another flavour: not mention raster vs vector at all.
```suggestion
Note that when using the `svg` option, it's expected for the original image to already be in this format, as different formats cannot be converted into SVG. The image itself won't be transformed, but the final `<img />` element will get the optimization attributes.
```

sarah11918, line 180:
```suggestion
Added in v 1.x: You can use the `Image` component when working with SVG images, but the `svg` option can only be used when the original object image is a `.svg` file. Other image formats (like `.png` or `.jpg`) cannot be converted into vector images. The `.svg` image itself will not be transformed, but the final `<img />` element will have the optimization attributes.
```
I'm wondering about something like this? (which also includes the idea that this was added? Maybe not very *gracefully*, but ...) just throwing this out there to generate discussion. I don't think this wording is "it" but I like leaning towards including WHY this is useful/helpful. (You can use the integration, even if you don't get ALL its features, so you don't have to change and write `` for only these images!)

sarah11918: (Also, I'm not sure about the original "optimization attributes" wording. If it passes everyone else's approval, I'm fine. It's just something I'd flag to think about.)

Yan-Thomas: Not sure what "original object image" means here, it's a way to refer to the imported image? So maybe just using "original imported image", here could work?

Also tried changing the wording of the last bit and changed `Image` to `<Image />`, so it's following the same pattern as the rest of the README. Let me know what you think of these changes.

```suggestion
Added in v0.15.0: You can use the `<Image />` component when working with SVG images, but the `svg` option can only be used when the original image is a `.svg` file. Other image formats (like `.png` or `.jpg`) cannot be converted into vector images. The `.svg` image itself will not be transformed, but the final `<img />` will be properly optimized by the integration.
```

sarah11918: I like this Yan! Commit this change when you're happy, then we'll let @delucis look at it like this and see what changes he wants to make.

natemoo-re (approved):

Issue #6230, opened 2023-02-13
Vercel analytics file missing


fschroiff:
### What version of `astro` are you using?

2.0.10

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

On my site https://www.fsch.ro I get a 404 on the file `/_vercel/insights/script.js`. 
also when building locally I don’t see the script.js file. 

This is my config:

```
import { defineConfig } from 'astro/config';
import sitemap from '@astrojs/sitemap';
import tailwind from "@astrojs/tailwind";
import prefetch from "@astrojs/prefetch";
import vercel from "@astrojs/vercel/static";

export default defineConfig({
  site: 'https://fsch.ro',
  integrations: [sitemap(), tailwind(), prefetch()],
  adapter: vercel({
    analytics: true
  })
});
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-4s4gze?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

mfrachet:
(I may be wrong) I'm feeling that this issue is about this specific loc from `@vercel/analytics`: https://github.com/vercel/analytics/blob/main/packages/web/src/generic.ts#L20

From my understanding, Vercel should add this script (see the "else" part of the ternary above).

Having you tried enabling analytics and the re-building (without cache)?

fschroiff:
I had to also enable "Audiences" in the Vercel Analytics settings and redeploy. Now the script.js gets added. 
Maybe some additional check is needed. Or it could be noted in the docs that you need to activate both options (web vitals and audiences) in Vercel Analytics.

fschroiff:
Could it be that for Web Vitals you don't need to add anything to your Astro setup? That the analytics package is only needed for the audiences feature?

delucis:
@fschroiff I see the script loading successfully on your site:
<img width="414" alt="image" src="https://user-images.githubusercontent.com/357379/219066810-6012997a-efd4-4f16-bd62-f8cf40d29443.png">

It could be that you have a browser extension that blocks tracking scripts? I also see a 404 when enabling Ghostery as Ghostery recognises Vercel’s analytics scripts and blocks it.

Locally this script is not available though as it is injected by Vercel (and you probably don’t want to collect analytics from local development). Maybe we should avoid injecting the analytics script during development?

fschroiff:
[...]

PR #6258, opened 2023-02-15
Only inject Vercel analytics in production

delucis:
## Changes

- Only injects the script for Vercel analytics on `build`
- Vercel’s analytics scripts rely on some extra scripts Vercel injects magically to projects built on their platform but these aren’t available locally so 404 (reported in #6230 along with other stuff)
- Folks likely don’t want to track analytics locally anyway

Philosophical question: is this difference between dev and build OK? Generally differences between the two can trip people up — what if this script interferes with some other part of the project somehow and you only notice in production?

## Testing

Existing tests should pass. Do we want a test that asserts the script is _not_ included in `dev`?

## Docs

No docs needed I don’t think.

### DIFF
--- a/.changeset/dull-wasps-decide.md
+++ b/.changeset/dull-wasps-decide.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': patch
+---
+
+Don’t inject analytics script in dev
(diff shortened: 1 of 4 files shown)


bluwy (approved):
I think it's fine for the dev and build to work differently, especially that analytics should be "addons" that a site can work with or without at a time. 

It also looks like Vercel analytics doesn't have an API for tracking custom events, so the changes is safer too.

natemoo-re (approved):

PR #6259, opened 2023-02-16
fix: add property name to rss errors

y-nk:
## Changes

- What does this change?
Add more information to zod error report

When generating a rss feed with the rss function, if one prop of an item is missing, the only feedback will be `Invalid input` without much of _where_ is the invalid input.

## Testing

I've added a prop on a string based on zod's doc here: https://github.com/colinhacks/zod/blob/master/ERROR_HANDLING.md#zodissue

Not sure if this must be tested.

## Docs

No doc as well, it's rather enhanced error handling.

### DIFF
--- a/.changeset/good-snails-attend.md
+++ b/.changeset/good-snails-attend.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/rss": patch
+---
+
+Improve RSS schema errors with additional property name context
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
This LGTM but I'd love for @bholmesdev to take a look as well

bholmesdev (approved):
Ah, valigreat fix! Thanks for this 👍

bholmesdev:
@y-nk Should be good to merge once you add a changeset. Run `pnpm changeset` at the repository root and choose a patch release.

y-nk:
thanks a lot @natemoo-re 🙏🙏🙏

PR #6268, opened 2023-02-16
Backwards compatible fix for `--camelCase` css vars in `style` attribute

### LINKED ISSUE #6264: Camel case CSS variables transformed to kebab case when used in style object

2.0.12

None

pnpm

Mac

### Describe the Bug

See [here](https://github.com/withastro/docs/issues/1539#issuecomment-1431759558). Camel case variables should be left alone, but are transformed when used in a style object. This does not occur with inline styles.

https://stackblitz.com/edit/github-2cyv6w?file=src/pages/index.astro

### Participation

natemoo-re:
## Changes

- Closes #6264 
- Removes a bug in our `style` logic that would convert `--camelCase` to `--camel-case`.
- This is done in a backwards-compatible way to avoid a breaking change.
- This should be removed in v3.

## Testing

Tested manually, small enough that we don't have tests

## Docs

Related to ...

### DIFF
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -27,7 +27,14 @@ const kebab = (k: string) =>
 	k.toLowerCase() === k ? k : k.replace(/[A-Z]/g, (match) => `-${match.toLowerCase()}`);
 const toStyleString = (obj: Record<string, any>) =>
 	Object.entries(obj)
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! Really smart way to handle this.

bluwy (approved):
```html
<div style="--my-value:var(--myValue);--myValue:red"></div>
```

Looks like the snake case var accesses the camelCase var that is declared later, but I tested this and it still worked fine too 👍 LGTM

Issue #6276, opened 2023-02-17
npm create astro@latest fails with 400 Bad Request


JohannesKlauss:
### What version of `astro` are you using?

2.0.13

### Are you using an SSR adapter? If so, which one?

N

### What package manager are you using?

npm and yarn

### What operating system are you using?

Mac

### Describe the Bug

I wanted to try out astro, but cannot install it, since it always fails during the create process when trying to download `https://github.com/withastro/astro/archive/latest.tar.gz`

This is the cli output with npm:

```
npm create astro@latest ./astro-poc

╭─────╮  Houston:
│ ◠ ◡ ◠  We're glad to have you on board.
╰─────╯

 astro   v2.0.13 Launch sequence initiated.

      ◼  dir Using ./astro-poc as project directory

  tmpl   How would you like to start your new project?
         Empty
Failed to download https://github.com/withastro/astro/archive/latest.tar.gz: 400 Bad Request
npm ERR! code 1
npm ERR! path /Users/johannesklauss/Documents/Spacetime/astro-test
npm ERR! command failed
npm ERR! command sh -c -- create-astro ./astro-poc

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/johannesklauss/.npm/_logs/2023-02-17T08_43_06_383Z-debug-0.log
```

And with yarn it is the same. Not sure if this is related to github or astro.

### Link to Minimal Reproducible Example

No reproducible applicable

### Participation

- [X] I am willing to submit a pull request for this issue.

julianbattye:
I'm also having this issue. Is there a simple way to bypass and get it installed such as a previous version?

eliseomurillo9:
Same issue here and I still dont find any way to solve it, even with previous versions

julianbattye:
It may be unrelated but I've just updated Node from 16 to 18 and it appears to be working now.

Princesseuh:
Hi everyone, sorry for the issue. We're currently investigating what's happening.

In the meantime, if you can't upgrade to Node 18 or higher, installing manually should work: https://docs.astro.build/en/install/manual/

Sorry for the inconvenience!

Princesseuh:
Hi, this seems to be caused by an issue upstream with the package we use to download the template: https://github.com/unjs/giget/issues/81

We'll be reverting the version in the meantime, sorry again for the inconvenience.

PR #6278, opened 2023-02-17
fix(create-astro): Pin giget to 1.0.0

Princesseuh:
## Changes

An upstream issue in giget is causing `downloadTemplate` to cause a 400 when hitting GitHub. (see https://github.com/unjs/giget/issues/81 ) Their own CI caught this, but they unfortunately still released the update. This PR pins the version to 1.0.0 in the meantime

Fix https://github.com/withastro/astro/issues/6276

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/.changeset/rare-lizards-grin.md
+++ b/.changeset/rare-lizards-grin.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Revert to giget 1.0.0 until upstream issue is fixed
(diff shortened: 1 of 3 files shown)


delucis (approved):
LGTM!

pi0:
Issues should be fixed in latest giget release v1.1.1 (https://github.com/unjs/giget/pull/82) :)

PR #6285, opened 2023-02-17
Add documentation for unknown errors

Princesseuh:
## Changes

We still generate docs links for unknown errors, so we should have documentation for them!

## Testing

N/A

## Docs

It's docs all the way down!

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -506,8 +506,6 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	// CSS Errors - 5xxx
 	/**
 	 * @docs
-	 * @message
-	 * Unknown CSS Error.
 	 * @see
 	 * 	- [Styles and CSS](https://docs.astro.build/en/guides/styling/)
 	 * @description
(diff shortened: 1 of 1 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -590,6 +633,13 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	 * @name CLI Errors
 	 */
 	// CLI Errors - 8xxx
+	/**
+	 * @docs
+	 * @description
+	 * Astro encountered an unknown error while starting one of its CLI command. The error message should contain more information.
```
Yan-Thomas, line 639: I believe that when we use "one of" we need to use a plural noun instead.
```suggestion
	 * Astro encountered an unknown error while starting one of its CLI commands. The error message should contain more information.
```

Yan-Thomas (commented):
Thanks @Princesseuh, great work tackling this. I have a small suggestion and then LGTM!

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -676,7 +734,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		hint: 'See https://docs.astro.build/en/guides/content-collections/ for more on the `slug` field.',
 	},
 
-	// Generic catch-all
+	// Generic catch-all - Only use this in extreme cases, like if there was a cosmic ray bit flip
```
natemoo-re, line 737: 😆 Nice clarification

natemoo-re (approved):
Looks great, nice job!

Yan-Thomas (approved):

PR #6286, opened 2023-02-17
Head bubble

matthewp:
## Changes

- Implements https://github.com/withastro/roadmap/discussions/481
- With this change `<head>` is treated differently depending on where it appears in the template. It can be:
  - Nested inside of `<html>` tag, in which case it acts like today.
  - Top-level, in which case it is hoisted and its contents are placed in the actual head when the page is sent out.
- Because the initial response is blocked waiting for head to propagate, this also means you can modify the response (to insert headers for example), by having a head in your template.

This can be tried out with:

```shell
npm install astro@next--head-bubbling
```

## Testing

- New test cases

## Docs

- TODO

### DIFF
--- a/examples/head-bubbling/package.json
+++ b/examples/head-bubbling/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "@example/head-bubbling",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "dev": "astro dev",
+    "start": "astro dev",
+    "build": "astro build",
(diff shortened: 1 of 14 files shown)


Princesseuh (changes requested):
This is all wrong, needs changes!!!

matthewp:
!preview head-bubbling

matthewp:
Closing as we are not pursuing this in the immediate future, but if we wind up doing it this code can be used as a starting point.

PR #6288, opened 2023-02-17
Enhanced Astro codespace support

bdmac:
## Changes
See https://github.com/withastro/roadmap/discussions/472 for further context.

Enhances the base Codespaces devcontainer and Dockerfile to provide a smoother development experience out of the box for contributors. Users of the base devcontainer will now be dropped into a fully functional development environment in their codespace and tests will automatically start running for them in a terminal tab.

Adds a devcontainer per example so that we can add buttons on astro.new that will launch users straight into the appropriate example app. Additional minimal Dockerfiles were created to support general example apps and the deno example app. A devcontainer was created for each example to ensure that the user is launched into the appropriate folder, has the appropriate initial file open, and that the example app is running on their behalf once the codespace loads. The built in browser preview should also show the running example by default.

Screenshot 2023-02-17 at 2 50 00 PM

Screenshot 2023-02-17 at 2 26 22 PM

## Testing

Only tweaking Codespaces devcontainer setups so no tests were added or updated.

## Docs

I've adjusted the readmes in all examples to include links to starting a codespace on that example as shown above in the screenshots. After this is merged I can also update astro.new to add these same links.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/javascript-node
+FROM mcr.microsoft.com/devcontainers/javascript-node:0-18
 
 # Install playwright
 RUN npm install -g @playwright/test
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,17 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node
+
+# Install playwright
+RUN npm install -g @playwright/test
+
+# Install latest pnpm
+RUN npm install -g pnpm
+
+# Install deno
+ENV DENO_INSTALL=/usr/local
+RUN curl -fsSL https://deno.land/x/install/install.sh | sh
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && curl -sSL https://dl.google.com/linux/direct/google-chrome-stable_current_$(dpkg --print-architecture).deb -o /tmp/chrome.deb \
+    && apt-get -y install /tmp/chrome.deb
+
+COPY welcome-message.txt /usr/local/etc/vscode-dev-containers/first-run-notice.txt
--- a/.devcontainer/basics/devcontainer.json
+++ b/.devcontainer/basics/devcontainer.json
@@ -1,5 +1,3 @@
-// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
-// https://github.com/microsoft/vscode-dev-containers/tree/v0.238.1/containers/docker-existing-dockerfile
 {
     "name": "Astro Examples",
     "build": {
@@ -34,8 +32,7 @@
       "vscode": {
         "extensions": [
           "astro-build.astro-vscode",
-          "esbenp.prettier-vscode",
-          "dbaeumer.vscode-eslint"
+          "esbenp.prettier-vscode"
         ]
       }
     }
--- a/.devcontainer/basics/devcontainer.json
+++ b/.devcontainer/basics/devcontainer.json
@@ -1,0 +1,43 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.238.1/containers/docker-existing-dockerfile
+{
+    "name": "Astro Examples",
+    "build": {
+      "dockerfile": "../examples.Dockerfile"
+    },
+
+    "workspaceFolder": "/workspaces/astro/examples/basics",
+
+    "portsAttributes": {
+      "3000": {
+        "label": "Application",
+        "onAutoForward": "openPreview"
+      }
+    },
+  
+    "forwardPorts": [3000],
+  
+    "postCreateCommand": "pnpm install && cd /workspaces/astro && pnpm run build",
+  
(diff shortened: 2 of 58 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node
```
samruddhikhandale, line 1: Can we pin this image to a specific node version and probably reference a semantic version ? 
This way we could avoid inconsistencies, and build failures due to future (even breaking) updates to this image.

```suggestion
FROM mcr.microsoft.com/devcontainers/javascript-node:0-18
```

.devcontainer/basics/devcontainer.json
```diff
@@ -0,0 +1,43 @@
...
+      "codespaces": {
+        "openFiles": [
+          "src/pages/index.astro"
+        ]
+      },
+      "vscode": {
+        "extensions": [
+          "astro-build.astro-vscode",
+          "esbenp.prettier-vscode",
+          "dbaeumer.vscode-eslint"
```
samruddhikhandale, line 38: We could remove this as this extension is already added by the image
https://github.com/devcontainers/images/blob/main/src/javascript-node/.devcontainer/devcontainer.json#L27

.devcontainer/blog/devcontainer.json
```diff
@@ -0,0 +1,43 @@
...
+      "codespaces": {
+        "openFiles": [
+          "src/pages/index.astro"
+        ]
+      },
+      "vscode": {
+        "extensions": [
+          "astro-build.astro-vscode",
+          "esbenp.prettier-vscode",
+          "dbaeumer.vscode-eslint"
```
samruddhikhandale, line 38: Same comment here and for every other devcontainer.json which is based upon the javascript-node image

.devcontainer/examples.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node
```
samruddhikhandale, line 1: > Can we pin this image to a specific node version and probably reference a semantic version ?

Same comment here and for every other dockerfile

.devcontainer/basics/devcontainer.json
```diff
@@ -0,0 +1,43 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.238.1/containers/docker-existing-dockerfile
```
samruddhikhandale, line 2: > https://github.com/microsoft/vscode-dev-containers/tree/v0.238.1/containers/docker-existing-dockerfile

Wondering why would we need a reference to `docker-existing-dockerfile`? 🤔 
Also, the container in `microsoft/vscode-dev-containers` are not currently supported and are moved to the [dev containers](https://github.com/devcontainers) org.

.devcontainer/example-welcome-message.txt
```diff
@@ -0,0 +1,3 @@
+👋 Welcome to Astro!
+
+The example app should automatically start soon in a new terminal tab.
```
samruddhikhandale, line 3:
```suggestion
👋 Welcome to "Astro" in GitHub Codespaces!

🛠️  Your environment is fully setup with all the required software.

🚀 The example app should automatically start soon in a new terminal tab.
```

samruddhikhandale: Probably we could make similar changes to other welcome messages as well?

samruddhikhandale (commented):
This is awesome, thank you so much! ✨ 
Left some comments/thoughts.

.devcontainer/basics/devcontainer.json
bdmac: Probably copied from somewhere.

.devcontainer/basics/devcontainer.json
bdmac: Sure, I didn't even realize the images could _have_ their own devcontainers nor how they interact with ours!

.devcontainer/example-welcome-message.txt
bdmac: Not following what you mean here.

samruddhikhandale: I am suggesting an update to the [example-welcome-message.txt](https://github.com/withastro/astro/pull/6288/files/7c4fce28e22b47fa85cca54600ac01d7a0737dc3#diff-c26282470ea314cfa922b5047b78b10fe214470a8284be71122f17c0ecab2c88) contents. If you decide to apply those changes, then you could also update other welcome files as well. (eg. devcontainer/welcome-message.txt)

natemoo-re (approved):
🙌 This is amazing, thanks so much for the PR!

This LGTM but I'd love one more review before merging.

Yan-Thomas (approved):
Docs LGTM, thanks @bdmac!

natemoo-re:
Cool, I'm going to go ahead and merge this @bdmac! Thanks so much for the PR. 🙌

Regarding `astro.new`, we just finalized the mockups for the planned redesign, so we'll probably be overhauling that site early in Q2. You're still welcome to create a PR to add these, but I just wanted to give you a heads up that the look and feel will be changing significantly in a month or two.

PR #6289, opened 2023-02-17
fix(examples/docs): Styles for left sidebar's group titles are not applied.

djobbo:
## Changes

- Fix left sidebar group titles css.

*Before*
image
*After*
image

> Another way to fix this would be to use the `heading` class defined in `examples/docs/src/styles/index.css` that is equivalent to `nav-group-title` and is used in the right sidebar. This would avoid having duplicate code, but would couple heading styles in left and right sidebars.

## Testing

- tested manually, minor visual fix.

## Docs

- no new docs needed.


Yan-Thomas (approved):
LGTM, thanks for catching and fixing this @djobbo 💜!

bluwy (approved):

PR #6296, opened 2023-02-18
Update MDX integration to es-module-lexer@1.1.1

RaphaelBossek:
## Changes
Updated es-module-lexer to 1.1.1 in MDX integration using new API.

## Testing
Standard astro test suite passed.

## Docs
It's only a internel package update with modified API usage.


natemoo-re (approved):
Ah, I'm surprised we missed this. Thanks so much for the PR!

Issue #6020, opened 2023-01-29
New prerender option breaks catch-all route on root when using preview


andycansdale:
### What version of `astro` are you using?

2.0.2

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

Using Node in standalone mode and given the following pages directory:
```
/pages
   myfirstpage.astro
   mysecondpage.astro
   [...params].astro
```
If we apply `export const prerender = true` to any page on the root (i.e. myfirstpage or mysecondpage) then the rest route [...params].astro returns a 404 in preview mode after a new build. When in dev mode the page displays correctly.

The page also displays correctly if nested in another folder. In other words if we move it off the root.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-n7aqnf?file=src/pages/mypage.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Prerender support was added to the Node adapter in 5.0.3. This should be working now!

andycansdale:
Hi! I've just tested this with the Node adapter version 5.0.3 and the problem persists. if I prerender any page, the [...params].astro page on the root still throws a 404. As mentioned, this only happens after building the app. I can work fine in dev mode and navigate to our [...paras].astro page without any issues.

matthewp:
Thanks! Reopening

florian-renfer:
I am having the same issue, glad you are already working on it.

wulinsheng123:
mark

wulinsheng123:
I have just done a bug related to prerender issues.
I can help to check this bug.

leomp12:
Just trying to give more two cents here: it's not only on preview mode, `prerender = true` is also breaking catch-all route on Node SSR with middleware mode on production (`astro dev` works fine)

wulinsheng123:
> Just trying to give more two cents here: it's not only on preview mode, `prerender = true` is also breaking catch-all route on Node SSR with middleware mode on production (`astro dev` works fine)

Thanks for your hint.
I will try out what you said in my local

PR #6298, opened 2023-02-18
Fix prrender

wulinsheng123:
## Changes
fix:#6020


wulinsheng123:
Sorry, submit more than one for #6020.
I take more time to this bug to find the best solution.

PR #6307, opened 2023-02-19
feat(content): bring back contentConfig.slug

y-nk:
Hello :) I've been talking a little about this on the community discord.

I was first interested to find a way to organize my content in my filesystem, which started to be messy because of growing amount of entries for a single collection.

At first I found that you can pass `slug:` in frontmatter ([as documented here](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs)) but having _many_ entries i was rather looking for something programmatic.

By chance I found that my content collection had a `slug: (entry: EntryInfo) => string | Promise<string>` type defined and tried to use it... but it didn't do anything!

After asking on Discord it seems like this feature was "removed at the 2.0 release"
image
but **i'm opening this PR to open a discussion with the Astro crew to propose to bring it back.**

My primary usecase is to keep my filesystem tidy by sorting entries by filename (`000-slug.md`), but also would prefer not to have the `000-` in the urls.

Of course i could go over each of my entries and inject the proper frontmatter data with a shell script, but i think it seems like a reasonable usecase, which could also be useful for people interested in having dates in frontmatter (`pubDate: 2022-02-19`) and generate slugs with dates inside it (a la wordpress) without having to duplicate the data in the filename (or the other way around!).

I did not fix (yet) the types and the `getEntryBySlug` part, but this minimal amount of code looks promising (at least to my knowledge). I've been patching my node_modules/ and it seems like working, at least for generating the types and serving content in SSG with `getCollection`.

I'll be happy to complete more tasks related to this if it seems something you'd like to bring back, otherwise i can just flip my PR to remove the generated `slug()` type in the `astro/src/content/templates/types.d.ts`.

### DIFF
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -172,6 +172,15 @@ export async function createContentTypesGenerator({
 
 		const { id, collection } = entryInfo;
 
+		// not sure this is the best way
+		const { collections = {} } = await loadContentConfig({
+			fs,
+			settings,
+			viteServer,
+		})
+		const collectionConfig = collections[collection] ?? {}
+		if (collectionConfig.slug) entryInfo.slug = collectionConfig.slug(entryInfo)
+
 		const collectionKey = JSON.stringify(collection);
 		const entryKey = JSON.stringify(id);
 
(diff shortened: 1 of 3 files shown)


matthewp:
@bholmesdev can you recall the reason for removing this? I remember but the reasons are a little vague. I know that we do not want to bring it back however, @y-nk. We should discuss the use-cases on https://github.com/withastro/roadmap/discussions and then talk about a solution (which is maybe this but maybe something pretty different).

y-nk:
thank you for explaining the process, i'll bring this to discussion.

natemoo-re:
Thanks @y-nk! Since discussion is happening in https://github.com/withastro/roadmap/discussions/516, I'm going to close this PR.

PR #6314, opened 2023-02-19
[create-astro] Execute the 'git' step as the final interaction

MilesPernicious:
This ensures the initialized repository has all configuration changes commited in the first commit

## Changes

Changes the order of the `create-astro` interactive steps.  Previously, it would run `git init`, `git add`, `git commit` *before* prompting for which tsconfig schema the project would use.  If `strict` or `strictest` are selected, this would mean that `tsconfig.json` would then be modified, and the initial repo is dirty.

## Testing

pnpm install && pnpm build && pnpm test: All (47) tests passed.

There could likely be an additional test for this, though I'm not entirely sure how to start writing tests for this ATM.

## Docs

Docs should not be needed.
Only affects ordering of the interactive script. Anyone using create-astro programmatically should be using the CLI --flags

### DIFF
--- a/.changeset/great-colts-accept.md
+++ b/.changeset/great-colts-accept.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Prompt for git initialization last, so all configurations can get added to the initial commit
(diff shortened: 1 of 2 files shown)


bluwy (approved):
Make sense to me 👍

natemoo-re (approved):
Yep, this makes sense to me! Thank you.

Issue #6316, opened 2023-02-20
There is a problem with the packaging of `xlink:href` dynamic attribute values


J-env:
### What version of `astro` are you using?

2.0.14

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

<img width="868" alt="image" src="https://user-images.githubusercontent.com/25259920/220066681-5f6cf647-4b70-435c-a70f-4b707d40f7d9.png">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-sdeict-nxhefd

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #6319, opened 2023-02-20
fix: inserted space before namespaced attributes values

MoustaphaDev:
## Changes

- Fix https://github.com/withastro/astro/issues/6316
- Adds an optional parameter to the `addAttribute` function that we can use to prevent adding a space before the attribute's key

To be merged after https://github.com/withastro/compiler/pull/732, then tests should pass

## Testing

Check if foreign element's namespaced attribute is present when rendered.
Tested with a compiler preview, encoutered one possibly unrelated issue, see in comments.

## Docs

Bug-fix only

### DIFF
--- a/packages/astro/test/astro-attrs.test.js
+++ b/packages/astro/test/astro-attrs.test.js
@@ -55,6 +55,13 @@ describe('Attributes', async () => {
 		expect($('img').attr('happy:smile')).to.equal('sweet');
 	});
 
+	it('Passes namespaced expression attributes as expected', async () => {
+		const html = await fixture.readFile('/namespaced-with-expression/index.html');
+		const $ = cheerio.load(html);
+
+		expect($('use').attr('xlink:href')).to.deep.equal('#test');
+	});
+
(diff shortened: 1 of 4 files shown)


packages/astro/test/astro-attrs.test.js
```diff
@@ -55,6 +55,13 @@ describe('Attributes', async () => {
 		expect($('img').attr('happy:smile')).to.equal('sweet');
 	});
 
+	it('Passes namespaced expression attributes as expected', async () => {
+		const html = await fixture.readFile('/namespaced-with-expression/index.html');
+		const $ = cheerio.load(html);
+
+		expect($('use').attr('xlink:href')).to.deep.equal('#test');
```
MoustaphaDev, line 62: cheerio somehow only sees `href`, not `xlink:href`. I tried with jsdom and the tests pass.
Also when `xlink:href` is set in a div, tests pass

MoustaphaDev:
Closed as not needed anymore due to a change in https://github.com/withastro/compiler/pull/732

PR #6320, opened 2023-02-21
fix #6020

### LINKED ISSUE #6020: New prerender option breaks catch-all route on root when using preview

wulinsheng123:
## Changes

fix:#6020

I found out about this bug's reason. Because when checking  whether the component is 'PRERENDER.`
This pr will be better than the #6298.
But this pr change more that I don't know whether there is a break.


matthewp:
Looks great, thank you!

PR #6322, opened 2023-02-21
Fix import of type HTMLAttributes in blog example

ChrisEke:
## Changes

- Use `import type` for `HTMLAttributes` in the "Headerlink" component from Blog examples. This conforms with docs examples [Typescript: Built-in HTML attributes](https://docs.astro.build/en/guides/typescript/#built-in-html-attributes)

## Testing

No tests added. Change only affect examples.

### DIFF
--- a/examples/blog/src/components/HeaderLink.astro
+++ b/examples/blog/src/components/HeaderLink.astro
@@ -1,9 +1,7 @@
 ---
 import type { HTMLAttributes } from 'astro/types';
 
-interface Props extends HTMLAttributes<'a'> {
-	class?: string;
-}
+type Props = HTMLAttributes<'a'>;
 
 const { href, class: className, ...props } = Astro.props;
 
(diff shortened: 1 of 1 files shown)


examples/blog/src/components/HeaderLink.astro
```diff
@@ -1,5 +1,9 @@
 ---
-export interface Props extends astroHTML.JSX.AnchorHTMLAttributes {}
+import type { HTMLAttributes } from 'astro/types';
+
+interface Props extends HTMLAttributes<'a'> {
+	class?: string;
```
Princesseuh, line 5: Is this line needed? Presumably, `HTMLAttributes<'a'>` already includes a definition for `class`.

examples/blog/src/components/HeaderLink.astro
```diff
@@ -1,5 +1,9 @@
...
+}
```
ChrisEke, line 6:
```suggestion
type Props = HTMLAttributes<'a'>;
```

PR #6323, opened 2023-02-21
fix(cookies): Update Undici to 5.20 and fix cookies behaviour

Princesseuh:
## Changes

Now that Undici released 5.20, we can upgrade again. This fix the `npm audit` warning

## Testing

Tests will pass!

## Docs

N/A

### DIFF
--- a/packages/integrations/netlify/src/netlify-functions.ts
+++ b/packages/integrations/netlify/src/netlify-functions.ts
@@ -102,34 +102,11 @@ export const createExports = (manifest: SSRManifest, args: Args) => {
 			isBase64Encoded: responseIsBase64Encoded,
 		};
 
-		// Special-case set-cookie which has to be set an different way :/
-		// The fetch API does not have a way to get multiples of a single header, but instead concatenates
-		// them. There are non-standard ways to do it, and node-fetch gives us headers.raw()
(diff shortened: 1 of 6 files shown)


packages/integrations/netlify/src/netlify-functions.ts
```diff
@@ -102,34 +102,11 @@ export const createExports = (manifest: SSRManifest, args: Args) => {
 			isBase64Encoded: responseIsBase64Encoded,
 		};
 
-		// Special-case set-cookie which has to be set an different way :/
-		// The fetch API does not have a way to get multiples of a single header, but instead concatenates
-		// them. There are non-standard ways to do it, and node-fetch gives us headers.raw()
-		// See https://github.com/whatwg/fetch/issues/973 for discussion
-		if (response.headers.has('set-cookie')) {
-			if ('raw' in response.headers) {
```
Princesseuh, line 110: Old left over from node-fetch. When everything is gonna be updated to use `getSetCookie`, we'll also be able to remove the whole `splitCookiesString` thing

bluwy (approved):
LGTM! Great to be able to remove the old hacks. Looks like this would need a changeset too.

ematipico (approved):

Issue #6329, opened 2023-02-22
Unknown Content Collection Error when parent directory starts with an underscore


justtobetrendy:
### What version of `astro` are you using?

2.0.14

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Hello,

I was experimenting with Astro and was constantly getting an [Unknown Content Collection Error](https://docs.astro.build/en/reference/errors/unknown-content-collection-error/) when accessing the `/blog` URL of the `blog template` site.

**Steps to reproduce**

1. `mkdir _sandbox`, `cd _sandbox` (the underscore is important)
2. `npm create astro@latest`, create it in `./blog`, select `Use blog template`, install dependencies, initialize git repo and select `Strict` TypeScript configuration.
3. `cd blog`
4. `npm run dev` 
5. navigate to `http://localhost:3000/blog` 

What I noticed was:

1. the type definition were not generated in `./astro` when running `npm run dev` or `npm run astro sync`,
2. this works in astro@2.0.6 but stopped working in astro@2.0.7,
3. removing the underscore from the parent directory (create it in `sandbox` directory rather than `_sandbox`) resolves the issue in astro@2.0.14.

Note that I can't reproduce this on stackblitz since I do not have control over the parent directory. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-2gphvs?file=package.json,src%2Fcontent%2Fconfig.ts,.gitignore

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #6330, opened 2023-02-22
fix(content): ensure underscore ignores only child paths of the content directory.

MoustaphaDev:
## Changes

Fix https://github.com/withastro/astro/issues/6329

## Testing

Tested manually and added unit tests to check that the `getEntryType` function returns correctly when there's an underscore above the `content` directory folder tree

## Docs

Bug fix only


natemoo-re (approved):
Great fix, nice tests. Thank you! 🙌

Issue #6240, opened 2023-02-13
Node adapter should add mode too


ElianCodes:
### What version of `astro` are you using?

2.0.11

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

1. Bootstrapping a new project with `pnpm create astro@latest` and going with all defaults.
2. Adding Node adapter through `pnpx astro add node`
3. config gets updated, but misses the `mode` option in the Node adapter (see screenshot linked)

Shouldn't we add the `mode: 'standalone'` as a default with the config when installing the adapter?

![image](https://user-images.githubusercontent.com/15145918/218597036-d93baddf-04af-401b-9460-4657f42f7ce0.png)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-f7h71h

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
@ematipico this one comes from `src/core/add/`. I'm not sure if there's an easy/good solution here, so something to discuss.

matthewp:
Suggestion:

1. Figure out if any other integrations take arguments.
2. If so, do the same thing
3. Maybe come up with a generic way to pass integration specific config to `astro add`.

PR #6333, opened 2023-02-22
fix(astro): correctly change configuration when node adapter is added

ematipico:
## Changes

Fixes https://github.com/withastro/astro/issues/6240

## Testing

I tested manually. Unfortunately, I haven't found a way to test these commands.

Screenshot 2023-02-22 at 18 39 52

## Docs

N/A


natemoo-re (approved):
🎉 Looks great to me! Nice fix.

PR #6343, opened 2023-02-23
Improve portfolio example icon support

delucis:
## Changes

Recently saw a few people asking about how to add icons to the portfolio starter (specifically saw Discord & LinkedIn logos requested). This change aims to document the process in the template and add a few more common icons.

- Add 4 new icons to the out-of-the-box icons in the portfolio example (logos for Discord, LinkedIn, Instagram, and TikTok).
- Tweak `Icon.astro` so it works for shapes with `fill` not just `stroke`.
- Add step-by-step instructions on how users can add more icons.

## Testing

Manually tested the new icons using `astro dev` while adding them.

## Docs

Icon instructions are currently in a source code comment. Our other examples don’t have much in the way of documentation, so went for that, but in theory we could also have this in the README.


bluwy (approved):
Looks good!

PR #6344, opened 2023-02-23
Implement RFC "A core story for images"

Princesseuh:
## Changes

This implements the [`A plan for a core image story` RFC](https://github.com/withastro/roadmap/pull/500)

All of it is implemented under a `experimental.image` flag. Please refer to the RFC for specific details as to what it adds

## Testing

Added tests! More tests could be added, but since this is going under an experimental flag, we can tackle that later

## Docs

https://github.com/withastro/docs/pull/2779


matthewp:
Looks like currently this is not being added automatically:

```
/// <reference types="astro/client-image" />
```

Can we tie into the same mechanism that content collection uses to append to env.d.ts?

Also I think this needs to be in the `exports` map of package.json.

matthewp:
@Princesseuh Writing tests I ran into some small bugs around finding files in the right place in the various ways that build works; ssg, ssr, prerender. All of this is tested now so I think it's good, but if you see a problem it's likely caused by [2aec7ef](https://github.com/withastro/astro/pull/6344/commits/2aec7efde1f23e0c9f153ceb32d3116d2fbb1df5)

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -143,6 +143,8 @@ export default function assets({
 						? settings.config.build.server
 						: settings.config.outDir;
 
+				console.log('dir', dir);
```
matthewp, line 146: oh no

matthewp:
Should `getImage` require the `alt` prop?

matthewp:
I'm done with my round of tests. I think I tested everything in the RFC. When `quality` is ready I'll add a test for that as well.

Princesseuh:
> I'm done with my round of tests. I think I tested everything in the RFC. When `quality` is ready I'll add a test for that as well.

It should be ready!

matthewp:
Windows 😔 I think I know what's causing that, it's just the test I think.

packages/astro/src/assets/consts.ts
```diff
@@ -0,0 +1,14 @@
...
+export const VALID_INPUT_FORMATS = [
+	'heic',
+	'heif',
+	'avif',
+	'jpeg',
+	'jpg',
+	'png',
+	'tiff',
+	'webp',
+	'gif',
```
ematipico, line 12: nit: I wonder if there's a way to keep these formats in sync with `client-image.d.ts`

packages/astro/src/assets/types.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
+type ImageSharedProps<T> = T & {
+	/**
+	 * Width of the image, the value of this property will be used to assign the `width` property on the final `img` element.
+	 *
+	 * For local images, this value will additionally be used to resize the image to the desired width, taking into account the original aspect ratio of the image.
+	 *
+	 * **Example**:
```
ematipico, line 48: I think if you use the `@example` JSDoc tag, then the IDE will render the code block correctly

packages/astro/src/assets/utils/etag.ts
```diff
@@ -0,0 +1,44 @@
+/**
```
ematipico, line 1: Should we also copy the licence file? 

EDIT: they don't have a licence file, so maybe we could link: https://github.com/tjwebb/fnv-plus#license

ematipico (commented):
Some random questions and comments. Nothing blocking

packages/astro/src/assets/types.ts
ematipico: Oh neat! I didn't know that plain markdown could work!

packages/astro/src/assets/image-endpoint.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		if (!isLocalService(imageService)) {
+			throw new Error('Configured image service is not a local service');
+		}
+
+		const url = new URL(request.url);
+		const transform = await imageService.parseURL(url);
+
+		if (!transform || !transform.src) {
+			throw new Error('Incorrect transform returned by `parseURL`');
```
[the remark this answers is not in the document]
Princesseuh: We can definitely add more information to the message, however since this error is shown to users of the user's website, we actually need to be careful about what we say.

packages/astro/src/assets/internal.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+interface GenerationData {
+	weight: {
+		before: number;
+		after: number;
+	};
+}
+
+export async function generateImage(
```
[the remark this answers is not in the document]
Princesseuh: This API is not public, it's only called by our internal build.

packages/astro/src/assets/consts.ts
Princesseuh: Apart from type gen, I unfortunately don't think so :( Module declarations are unfortunately super rigid in TypeScript

packages/astro/components/Image.astro
```diff
@@ -0,0 +1,28 @@
+---
+import { getImage, type LocalImageProps, type RemoteImageProps } from 'astro:assets';
+import { AstroError, AstroErrorData } from '../dist/core/errors/index.js';
+
+// The TypeScript diagnostic for JSX props uses the last member of the union to suggest props, so it would be better for
+// LocalImageProps to be last. Unfortunately, when we do this the error messages that remote images get are complete nonsense
+// Not 100% sure how to fix this, seems to be a TypeScript issue. Unfortunate.
+type Props = LocalImageProps | RemoteImageProps;
+
+const { ...props } = Astro.props;
```
matthewp, line 10: is there a need for the spread?

packages/astro/src/assets/services/service.ts
```diff
@@ -0,0 +1,155 @@
...
+		return '/_image?' + searchParams;
+	},
+	parseURL(url) {
+		const params = url.searchParams;
+
+		if (!params.has('href')) {
+			return {};
+		}
+
+		let transform: Record<string, any> = { src: params.get('href')! };
```
matthewp, line 135: Should this object have more of a defined shape? If so you could initialize the other values as `undefined`. This will also help with perf (if this function is called often enough for that to matter). Disregard if it shouldn't have more of a shape.

packages/astro/src/assets/types.ts
```diff
@@ -0,0 +1,111 @@
...
+import { ImageService } from './services/service.js';
+
+export type ImageQualityPreset = 'low' | 'mid' | 'high' | 'max' | (string & {});
+export type ImageQuality = ImageQualityPreset | number;
+export type InputFormat = (typeof VALID_INPUT_FORMATS)[number] | (string & {});
+export type OutputFormat = (typeof VALID_OUTPUT_FORMATS)[number] | (string & {});
+
+declare global {
+	// eslint-disable-next-line no-var
+	var astroAsset: {
```
matthewp, line 12: I have thoughts 😅, for later.

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -0,0 +1,229 @@
...
+const resolvedVirtualModuleId = '\0' + VIRTUAL_MODULE_ID;
+
+export default function assets({
+	settings,
+	logging,
+	mode,
+}: AstroPluginOptions & { mode: string }): vite.Plugin[] {
+	let resolvedConfig: vite.ResolvedConfig;
+
+	globalThis.astroAsset = {};
```
matthewp, line 28: In the meantime, what do you think about tucking away the use of this global into its own module. So the nastiness can be contained to within there? Something like:

```js
export function createGlobal() {
  globalThis.astroAsset = {};
}

export default function() {
  return globalThis.astroAsset;
}

export function destroyGlobal() {
  delete globalThis.astroAsset;
}
```

Then, here you would do:

```js
import { createGlobal } from './asset.js';


....
createGlobal();
```

And anywhere that just needs to access it does:

```js
import astroAsset from '/path/to/assets/asset.js';

// Get the image service
astroAsset().imageService; // ....

// Call the addStaticImage function
astroAsset().addStaticImage(/* ... */);
```

etc. Then when we refactor to remove the global hopefully anything using it wouldn't need to change.

Just an idea, let me know if you don't think it'll help enough.

matthewp:
Most of my comments are somewhere between nits and nice-to-haves, nothing majorly blocking. Looking forward to using this feature!

packages/astro/src/assets/vite-plugin-assets.ts
Princesseuh: I tried to do this, because it seemed like a good idea, but unfortunately TypeScript made this hell and it lead to needing to add a bunch of unnecessary checks everywhere 😦

Princesseuh:
!preview experimental-assets

packages/astro/src/assets/services/squoosh.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { ImageQualityPreset, OutputFormat } from '../types.js';
+import { baseService, BaseServiceTransform, LocalImageService, parseQuality } from './service.js';
+import { processBuffer } from './vendor/squoosh/image-pool.js';
+import type { Operation } from './vendor/squoosh/image.js';
+
+const baseQuality = { low: 25, mid: 50, high: 80, max: 100 };
+const qualityTable: Record<Exclude<OutputFormat, 'png'>, Record<ImageQualityPreset, number>> = {
+	avif: {
+		// Squoosh's AVIF encoder has a bit of a weird behavior where `62` is technically the maximum, and anything over is overkill
+		max: 100,
```
natemoo-re, line 12: If this is the case, shouldn't `62` be `max` for consistency?

packages/astro/src/content/template/types.d.ts
```diff
@@ -3,6 +3,17 @@ declare module 'astro:content' {
 	export type CollectionEntry<C extends keyof typeof entryMap> =
 		(typeof entryMap)[C][keyof (typeof entryMap)[C]] & Render;
 
+	export const image: (options: {
+		format?: string;
+		width?: number;
+		height?: number;
```
natemoo-re, line 9: Are these types still relevant with the change to prefer `refine`?

packages/astro/src/content/utils.ts
```diff
@@ -49,6 +49,14 @@ export const msg = {
 		`${collection} does not have a config. We suggest adding one for type safety!`,
 };
 
+export function extractFrontmatterAssets(data: Record<string, any>): string[] {
+	const assets = Object.values(data).filter(
+		(value) => typeof value === 'object' && value['__astro_asset'] == true
+	);
```
natemoo-re, line 55: Looks like this only supports `image()` at the top level. Does this need to be recursive?

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -91,11 +93,18 @@ export function astroContentImportPlugin({
 					? await getEntryData(partialEntry, collectionConfig)
 					: unparsedData;
 
+				const images = extractFrontmatterAssets(data).map(
+					(image) => `'${image}': await import('${normalizePath(image)}'),`
+				);
```
natemoo-re, line 98: Same comment, only supported at the top-level.

packages/astro/src/core/path.ts
```diff
@@ -64,3 +64,7 @@ export function removeQueryString(path: string) {
 	const index = path.lastIndexOf('?');
 	return index > 0 ? path.substring(0, index) : path;
 }
+
+export function isRemotePath(src: string) {
+	return /^(https?:)?\/\//.test(src);
```
natemoo-re, line 69: Should probably also handle `data:` URIs

packages/astro/test/fixtures/core-image-ssg/src/pages/index.astro
```diff
@@ -0,0 +1,18 @@
...
+<head>
+
+</head>
+<body>
+	<div id="local">
+		<Image src={myImage} alt="a penguin" />
+	</div>
+	
+	<div id="remote">
+		<Image src="https://avatars.githubusercontent.com/u/622227?s=64" alt="fred" width="48" height="48" />
```
natemoo-re, line 15: Probably want a test for `data:` URIs as well

natemoo-re (commented):
Looking really great, just had a few small comments!

packages/astro/src/core/config/schema.ts
```diff
@@ -115,6 +118,17 @@ export const AstroConfigSchema = z.object({
 			.optional()
 			.default({})
 	),
+	image: z
```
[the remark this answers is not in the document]
matthewp: Seems like something we can debate in the experimental phase :) There was a similar comment in the RFC IIRC. Feels like we can clean this up before stable.

mandar1jn:
Just want to say it's really cool that you guys are improving the workflow around content so much

packages/astro/src/content/utils.ts
```diff
@@ -50,11 +51,19 @@ export const msg = {
...
-	);
+	function findAssets(potentialAssets: Record<string, any>): ImageMetadata[] {
+		return Object.values(potentialAssets).map((entry: any) => {
+			if (typeof entry === 'object') {
+				if (entry.__astro === true) {
+					return entry;
+				} else {
+					return findAssets(entry);
+				}
+			}
```
ematipico, line 62: 1. if `typeof entry === 'object'` is `false`, we don't map to anything, and we return `undefined`, is that correct?
2. Since we are using recursion with `findAssets`, maybe this is a good situation to use [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

Princesseuh: Yeah, logic is a bit faulty, as tests have shown... Will fix

How would you use `flatMap` in this context? It only flatten a single level

ematipico: If we know that we can only have one level, maybe we don't need to use a recursive approach. At first I thought we could have many levels.

Princesseuh: The images can be infinitely deep, as Markdown frontmatters can include nested elements

Princesseuh:
!preview experimental-assets

packages/astro/src/assets/vite-plugin-assets.ts
matthewp: ok, we'll revisit later.

matthewp (approved):

packages/astro/src/content/utils.ts
ematipico: You could `.flat(Infinity)`. `.flat` accepts a number that tells the function how many levels deep the flattening needs to occur.

.changeset/gold-rocks-cry.md
```diff
@@ -0,0 +1,15 @@
...
+
+Add a new experimental flag (`experimental.assets`) to enable our new core Assets story.
+
+This unlocks a few features:
+- Image component and JavaScript API to transform and optimize images.
+- Support for images stored relatively to your content in Markdown.
+- Support for validating assets using Content Collections.
+- and more!
+
+See [Assets (Experimental)](https://docs.astro.build/en/guides/assets/) on our docs for more information on how to use this feature!
```
delucis, line 15: Small typo, probably should be “in our docs” or “on our docs site”

```suggestion
See [Assets (Experimental)](https://docs.astro.build/en/guides/assets/) on our docs site for more information on how to use this feature!
```

packages/astro/components/Image.astro
```diff
@@ -0,0 +1,28 @@
+---
+import { getImage, type LocalImageProps, type RemoteImageProps } from 'astro:assets';
+import { AstroError, AstroErrorData } from '../dist/core/errors/index.js';
+
+// The TypeScript diagnostic for JSX props uses the last member of the union to suggest props, so it would be better for
+// LocalImageProps to be last. Unfortunately, when we do this the error messages that remote images get are complete nonsense
+// Not 100% sure how to fix this, seems to be a TypeScript issue. Unfortunate.
+type Props = LocalImageProps | RemoteImageProps;
```
delucis, line 8: Does `RemoteImageProps['src']` always have a protocol? Is it possible to brand an interface that way so that `http(s)://`-prefixed `src` indicates we should use `RemoteImageProps`? I guess not because we don’t have a way to make `Props` take a type argument?

packages/astro/src/@types/astro.ts
```diff
@@ -918,7 +932,27 @@ export interface AstroUserConfig {
...
-	experimental?: object;
+	experimental?: {
+		/**
+		 * @docs
+		 * @name experimental.assets
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 2.1.0
+		 * @description
+		 * Enable experimental support for enhanced images support and ability to optimize and resize images directly in Astro. With this enabled, a new `astro:assets` module will be exposed
```
delucis, line 943: Does this read OK? I’m trying to avoid the double “support”.

```suggestion
		 * Enable experimental image processing support, including the ability to optimize and resize images directly in Astro. With this enabled, a new `astro:assets` module will be exposed.
```

.changeset/gold-rocks-cry.md
```diff
@@ -0,0 +1,15 @@
...
+'@astrojs/markdown-remark': minor
+---
+
+Add a new experimental flag (`experimental.assets`) to enable our new core Assets story.
+
+This unlocks a few features:
+- Image component and JavaScript API to transform and optimize images.
+- Support for images stored relatively to your content in Markdown.
+- Support for validating assets using Content Collections.
+- and more!
```
sarah11918, line 13:
```suggestion
Add a new experimental flag (`experimental.assets`) to enable our new core Assets story.

This unlocks a few features:
- A new built-in image component and JavaScript API to transform and optimize images.
- Relative images with automatic optimization in Markdown and MDX files.
- Support for validating assets using content collections.
- and more!
```
Just suggesting some edits to make this more closely align with what we have on the assets page itself. 

I think emphasizing new and built-in for the image component could be helpful (since we're going to have to remind people they can't use the existing integration), and I think "relative images" in Markdown/MDX is the hot phrase everyone wants to hear!  🥳

packages/astro/src/@types/astro.ts
delucis: It’s also not clear here what you enable _apart_ from optimisation/resizing, so we could also just compress it down to that:

```suggestion
		 * Enable experimental support for optimizing and resizing images. With this enabled, a new `astro:assets` module will be exposed.
```

packages/astro/src/assets/services/service.ts
```diff
@@ -0,0 +1,174 @@
...
+	return result;
+}
+
+interface SharedServiceProps {
+	/**
+	 * Return the URL to the endpoint or URL your image are generated from.
+	 *
+	 * For a local service, your service should expose an endpoint handling the image requests, or use Astro's at `/_image`.
+	 *
+	 * For external services, this should point to the URL where your images are coming from, for instance, `/_vercel/image`
```
delucis, line 31: Language nit:

```suggestion
	 * Return the URL to the endpoint or URL your images are generated from.
	 *
	 * For a local service, your service should expose an endpoint handling the image requests, or use Astro's at `/_image`.
	 *
	 * For external services, this should point to the URL your images are coming from, for instance, `/_vercel/image`
```

packages/astro/src/assets/services/service.ts
```diff
@@ -0,0 +1,174 @@
...
+	 *
+	 */
+	getURL: (options: ImageTransform) => string;
+	/**
+	 * Return any additional HTML attributes separate from `src` your service require to show the image properly.
+	 *
+	 * For example, you might want to return the result `width` and `height` to avoid CLS, or a particular `class` or `style`.
+	 * In most cases, you'll want to return directly what your user supplied you, minus the attributes that were used to generate the image.
```
delucis, line 39: Nit:

```suggestion
	 * Return any additional HTML attributes separate from `src` that your service requires to show the image properly.
	 *
	 * For example, you might want to return the `width` and `height` to avoid CLS, or a particular `class` or `style`.
	 * In most cases, you'll want to return directly what your user supplied you, minus the attributes that were used to generate the image.
```

sarah11918 (approved):
Looks great! Most of the excitement will happen in the docs PR itself, but some thoughts on the changeset for you. Congrats, @Princesseuh 🥳

packages/astro/src/assets/services/service.ts
```diff
@@ -0,0 +1,174 @@
...
+export type ExternalImageService = SharedServiceProps;
+
+type LocalImageTransform = {
+	src: string;
+	[key: string]: any;
+};
+
+export interface LocalImageService extends SharedServiceProps {
+	/**
+	 * Parse the requested parameters passed along the URL from `getURL` back into an object to be used later by `transform`
```
delucis, line 53: Nit:

```suggestion
	 * Parse the requested parameters passed in the URL from `getURL` back into an object to be used later by `transform`
```

.changeset/gold-rocks-cry.md
Princesseuh: Just removed the `and MDX` part since this isn't actually supported at the moment, otherwise looks awesome! Thank you very much!

delucis (commented):
Some small comments!

packages/astro/components/Image.astro
Princesseuh: It does for our services (http(s), ftp(s) and data:), but other services might use `src` for any kind of keys, so we can't do that 😔

packages/astro/tsconfigs/base.json
```diff
@@ -17,6 +17,11 @@
     // Properly support importing CJS modules in ESM
     "esModuleInterop": true,
     // Skip typechecking libraries and .d.ts files
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    // Add alias for assets folder for easy reference to assets
+    "baseUrl": ".",
+    "paths": {
+      "~/assets/*": ["src/assets/*"]
```
delucis, line 24: Has there been much discussion about this? How does this merge with user-defined aliases? I notice we don’t show use of this at all in the new Assets docs.

packages/astro/src/assets/types.ts
```diff
@@ -0,0 +1,125 @@
...
+ */
+export interface ImageMetadata {
+	src: string;
+	width: number;
+	height: number;
+	format: InputFormat;
+}
+
+/**
+ * All the transformations possible to an image
```
delucis, line 30: Is this the most accurate description of this type or would “transformations” better be framed as “paramaters” given that `src` isn’t really a transformation?

Something like this maybe?

```suggestion
 * Options accepted by the image transformation service.
```

packages/astro/src/assets/types.ts
```diff
@@ -0,0 +1,125 @@
...
+	 * **Example**:
+	 * ```astro
+	 * <Image src={...} format="avif" alt="..." />
+	 * ```
+	 */
+	format?: OutputFormat;
+	/**
+	 * Desired quality for the image. Value can either be a preset such as `low` or `high`, or a numeric value from 0 to 100.
+	 *
+	 * Ultimately, the ending perceivable quality is loader-specific.
```
delucis, line 101:
```suggestion
	 * The perceptual quality of the output image is loader-specific.
```

delucis (commented):
A few more small comments!

packages/astro/tsconfigs/base.json
Princesseuh: I haven't received a single piece of feedback on this, I'll admit 🤷‍♀️ I don't necessarily like aliases myself, but I get the usage in Markdown (where it can be a bit cumbersome to know where `src/assets` is)

packages/astro/src/assets/types.ts
Princesseuh: I like it!

packages/astro/tsconfigs/base.json
delucis: I definitely understand the convenience — we have `src/` aliased in docs to allow easy referencing of components etc. from MDX — but just wasn’t sure about setting a default if we don’t for anything else. People have opinions on how these should look, and if we haven’t documented so far, maybe this is not required as a core feature. That said, if it went through the RFC process and people think it’s important, that’s cool!

Princesseuh: Definitely something we can look at post release, we have until 3.0 to figure out the alias 😄

PR #6347, opened 2023-02-23
Use getSetCookie, if available

natemoo-re:
## Changes

- Related to https://github.com/withastro/astro/pull/6323
- Node 18's `fetch` implementation may have been updated to the broken version of `undici`, so we need to keep this manual workaround for now.
- Once Node 18 is on `undici@5.20.0+` this should be fine to remove?
  - **UPDATE**: `18.14.1` is using the outdated version of `undici`, `18.14.2` is fine. We could wait for the CI images to update, but I think this is worth being defensive about.

## Testing

Covered by existing tests

## Docs

Bug fix only

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/response.ts
+++ b/packages/astro/src/vite-plugin-astro-server/response.ts
@@ -57,6 +57,12 @@ export async function writeWebResponse(res: http.ServerResponse, webResponse: Re
 
 	const _headers = Object.fromEntries(headers.entries());
 
+	// Undici 5.20.0+ includes a `getSetCookie` helper that returns an array of all the `set-cookies` headers.
+	// Previously, `headers.entries()` would already have these merged, but it seems like this isn't the case anymore.
+	if ('getSetCookie' in headers && typeof headers.getSetCookie === 'function') {
+		_headers['set-cookie'] = headers.getSetCookie();
+	}
+
 	// Attach any set-cookie headers added via Astro.cookies.set()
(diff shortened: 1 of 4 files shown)


Princesseuh (approved):

packages/astro/src/vite-plugin-astro-server/response.ts
```diff
@@ -57,6 +57,12 @@ export async function writeWebResponse(res: http.ServerResponse, webResponse: Re
 
 	const _headers = Object.fromEntries(headers.entries());
 
+	// Undici 5.20.0+ includes a `getSetCookie` helper that returns an array of all the `set-cookies` headers.
+	// Previously, `headers.entries()` would already have these merged, but it seems like this isn't the case anymore.
+	if ('getSetCookie' in headers && typeof headers.getSetCookie === 'function') {
```
matthewp, line 62: Not important since this is a dev server, but this is an annoying thing about TS where it wants you to use `in` which is slow, when just a directly `typeof headers.getSetCookie === 'function'` is what you really want to do here. That's why I tend to pair `typeof` with `as any`, ala `typeof (something as any).prop`. Just needed to rant on that, I'll walk myself out now.

matthewp (approved):

packages/astro/src/vite-plugin-astro-server/response.ts
natemoo-re: Fair! That's a great thing to keep in mind, but yes I was doing it for the type safety

PR #6348, opened 2023-02-23
CI pin `pnpm/action-setup` to `v2`

natemoo-re:
## Changes

- We previously pinned `pnpm/action-setup` to `v2.2.1` exactly, which was throwing warnings in our Actions tab.
- This updates the tag to `v2` so we're on the latest.

## Testing

N/A

## Docs

N/A


bluwy (approved):
I remember 2.2.2 had issues that was pinned in Vite too, but looks like that's fixed now. I'm using `v2` in other projects too.

PR #6351, opened 2023-02-23
[Lit] render DSD attributes based on `shadowRootOptions`

hrmcdonald:
## Changes

- Update `@astrojs/lit`’s `server.js` to properly render elements with `shadowRootOptions` that include setting [delegatesFocus](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus) to `true`.
- Logic is based on `@lit-labs/ssr` [latest implementation as found here](https://github.com/lit/lit/blob/b0c3f82ef0f97326a205e77e7e1043b75a5cc53f/packages/labs/ssr/src/lib/render-value.ts#L738-L748)

## Testing

A test was added to ensure an element with `delegatesFocus` set to `true` has this attribute properly included in the rendered static markup.

## Docs

No documentation changes are needed. This is expected functionality that currently isn't properly supported by `@astrojs/lit`.

### DIFF
--- a/.changeset/old-fireants-allow.md
+++ b/.changeset/old-fireants-allow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/lit': patch
+---
+
+Render DSD attributes based on `shadowRootOptions`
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Awesome, thanks for the fix!

natemoo-re (approved):
Awesome, thanks for the fix!

PR #6353, opened 2023-02-24
Improve error message for endpoint getStaticPaths with undefined value

### LINKED ISSUE #6095: `undefined` params within getStaticPaths cause Vercel build to fail

2.0.0

Vercel

pnpm

Mac

#### Expectation

When using `params` in a `getStaticPaths` function, I would expect `undefined` to result in the creation of an index route.

#### Actual Behaviour

The build fails in Vercel

##### Success: Without `undefined` `params`

- Deployment: https://vercel.com/joebell/astro-bug-params-undefined/4G9AqahtnJTsvXzCP6Z97H55ze73

```sh
19:36:18.324	▶ src/pages/example/[...slug].astro
19:36:18.345	  ├─ /example/one/index.html (+21ms)
19:36:18.347	  ├─ /example/two/index.html (+23ms)
19:36:18.348	  └─ /example/three/index.html (+25ms)
19:36:18.349	λ src/pages/api/example/[...slug].ts
19:36:18.350	  ├─ /api/example/one (+1ms)
19:36:18.351	  ├─ /api/example/two (+2ms)
19:36:18.351	  └─ /api/example/three (+3ms)
19:36:18.351	Completed in 59ms.
```

##### Error: With `undefined` `params`

- Deployment: https://vercel.com/joebell/astro-bug-params-undefined/Fj1im5wqsKA6XdVn2VhGkWrkSFja

```sh
▶ src/pages/example/[...slug].astro
19:39:10.815	  ├─ /example/one/index.html (+18ms)
19:39:10.817	  ├─ /example/two/index.html (+19ms)
19:39:10.818	  ├─ /example/three/index.html (+21ms)
19:39:10.820	  └─ /example/index.html (+22ms)
19:39:10.820	λ src/pages/api/example/[...slug].ts
19:39:10.821	  ├─ /api/example/one (+1ms)
19:39:10.822	  ├─ /api/example/two (+2ms)
19:39:10.822	  ├─ /api/example/three (+3ms)
19:39:10.911	 error   EISDIR: illegal operation on a directory, open '/vercel/path0/.vercel/output/static/api/example'
19:39:10.911	Error: EISDIR: illegal operation on a directory, open '/vercel/path0/.vercel/output/static/api/example'
19:39:10.926	 ELIFECYCLE  Command failed with exit code 1.
```

https://github.com/joe-bell/astro-bug-params-undefined

> **Note**
>
> More details have been provided in the reproduction

### Participation

---

Not part of the template, but just want to say a huge thanks to the Astro team – I love using this product and I appreciate you all!

wulinsheng123:
**Update by bluwy**:

## Changes

Fixes #6095

Provide a better error message for #6095 as it's not possible to prerender the endpoint with a specific edge case shown in the changeset message. The error looks something like this:

```
 error   Could not render `/api/catch/[...slug]` with an `undefined` param as the generated path will collide during prerendering. Prevent passing `undefined` as `params` for the endpoint's `getStaticPaths()` function, or add an additional extension to the endpoint's filename.
  Hint:
    Rename `src/pages/api/catch/[...slug].ts` to `src/pages/api/catch/[...slug].json.ts`
  File:
    src/pages/api/catch/[...slug].ts
  Stacktrace:
```

## Testing

Test added in Astro's integration tests

## Docs

This will create a new Astro error for https://docs.astro.build/en/reference/error-reference/#astro-errors. Appreciate a review!

**End of update**

---

## Changes
related to #6095

Use the `undefined ` to ignore an `index` in the function `getStaticPaths` when you want to access a website link, such as `api/example/index `.
I think there can't appear an `undefined` in the value of returned getStaticPaths.
For one, when an `undefined` in the array will break the bundle.
such as when config is
```js
export default defineConfig({
  output: "server",
  adapter: node({mode:'standalone'}),
  build: {
    format: 'file'
  }
});
```
And there is an `[...slug].astro` in the directory `pages/index`

When `[...slug].astro` is 

```astro 
---
export const prerender = true;

export function getStaticPaths() {
  const slugs = ["one", undefined];

  return slugs.map((u) => ({ params: { slug: u } }));
}

const { slug } = Astro.params;
---

<h1>{slug || "index"}</h1>
```
Result:
The `page/index/index.html` will replace Index.html.

image

For another: 
This situation also happens in the`Endpoints`.
It will fight to create the same file of trailing slash each other when the `getStaticPaths` return values of more than one, including an `undefined`.this also will cause a bug of [#6095](https://github.com/withastro/astro/issues/6095).
when it is 
```js 
import type { APIRoute } from "astro";

const slugs = ["one", undefined];

export const prerender = true;

export const get: APIRoute = ({ params }) => {
  return {
    body: JSON.stringify({
      slug: params.slug || "index",
    }),
  };
};

export function getStaticPaths() {
  return slugs.map((u) => ({ params: { slug: u } }));
}

```
The path like `api/example`.
When the one will create a path as `api/example/one`.
When the undefined  will create a path as `api/example`.
And will get an error:
` error   EISDIR: illegal operation on a directory, open '/Users/wuls/Desktop/issue/astro-bug-params-undefined/dist/client/api/example'`.

Resolution:

My idea is to tell the users uses to replace the `undefined` with the `index` and check the value of returned getStaticPaths.
Automating to replace `undefined` with `index` or throwing an error message if there includes an `undefined`    
or send a warn to users.

Such as:
```js
export function getStaticPaths() {
  const slugs = ["one", "index"];

  return slugs.map((u) => ({ params: { slug: u } }));
}

```

## Docs

https://docs.astro.build/en/reference/configuration-reference/#buildformat


wulinsheng123:
There will have a compatible problem.
Such as users who previously used `api/exmple` to obtain data from the endpoints need to use `api/exmple/index`to get data.

bluwy:
Ah alright, I see the issue now. Thanks for the write-up!

I don't think we can remove support for `undefined` yet because there are actual use case for it with slugs. Even though it has the conflict behaviour, e.g. `[...slug].astro` (with undefined) and `index.astro`, I think that is still correct because you're opting in to that. In which case we should log a warning that there's a conflict instead (probably future PR).

For endpoints, the problem seems to be that in static generation, we generate files without extensions, which conflict with folder names. I don't think we can add an extension here, that changes how it's served.

So I think the solution here is to log an error if `undefined` is used in endpoints `getStaticPaths`, where the endpoint file name is `*.js` or `*.ts`, and skip generating it statically. Because endpoints with `*.json.js` or `*.json.ts` will still work as that has an extension.

wulinsheng123:
Thanks for your reply.
I think this idea is better.
I will fix it late.

wulinsheng123:
> Ah alright, I see the issue now. Thanks for the write-up!
> 
> I don't think we can remove support for `undefined` yet because there are actual use case for it with slugs. Even though it has the conflict behaviour, e.g. `[...slug].astro` (with undefined) and `index.astro`, I think that is still correct because you're opting in to that. In which case we should log a warning that there's a conflict instead (probably future PR).
> 
> For endpoints, the problem seems to be that in static generation, we generate files without extensions, which conflict with folder names. I don't think we can add an extension here, that changes how it's served.
> 
> So I think the solution here is to log an error if `undefined` is used in endpoints `getStaticPaths`, where the endpoint file name is `*.js` or `*.ts`, and skip generating it statically. Because endpoints with `*.json.js` or `*.json.ts` will still work as that has an extension.

don't we need any tips if meet the `*.js`?
do we directly skip to generate ?

wulinsheng123:
> Ah alright, I see the issue now. Thanks for the write-up!
> 
> I don't think we can remove support for `undefined` yet because there are actual use case for it with slugs. Even though it has the conflict behaviour, e.g. `[...slug].astro` (with undefined) and `index.astro`, I think that is still correct because you're opting in to that. In which case we should log a warning that there's a conflict instead (probably future PR).
> 
> For endpoints, the problem seems to be that in static generation, we generate files without extensions, which conflict with folder names. I don't think we can add an extension here, that changes how it's served.
> 
> So I think the solution here is to log an error if `undefined` is used in endpoints `getStaticPaths`, where the endpoint file name is `*.js` or `*.ts`, and skip generating it statically. Because endpoints with `*.json.js` or `*.json.ts` will still work as that has an extension.

Your second solution will break a lot of tests, this one cannot be skipped.

bluwy:
> don't we need any tips if meet the `*.js`?
> do we directly skip to generate ?

Yeah we can report a warning if so.

> Your second solution will break a lot of tests, this one cannot be skipped.

I'm not sure how it's breaking the tests? Ideally following the issue, instead of a fs error, we replace it with a more user-friendly error if we know a specific case will fail.

wulinsheng123:
> > don't we need any tips if meet the `*.js`?
> > do we directly skip to generate ?
> 
> Yeah we can report a warning if so.
> 
> > Your second solution will break a lot of tests, this one cannot be skipped.
> 
> I'm not sure how it's breaking the tests? Ideally following the issue, instead of a fs error, we replace it with a more user-friendly error if we know a specific case will fail.

https://github.com/wulinsheng123/astro/blob/f6452f12305ecd778881d984b2e36ba61cec5cd6/packages/astro/test/fixtures/ssr-dynamic/src/pages/api/products/%5Bid%5D.js#L4
Look here.
We will break this test If we directly skip generating a file.
I tested it the whole day. 
I checked the route data and if it was the endpoint and *.ts(*.js). I wouldn't put route data into an array. And it will be ignored whatever dev and pro.

bluwy:
> https://github.com/wulinsheng123/astro/blob/f6452f12305ecd778881d984b2e36ba61cec5cd6/packages/astro/test/fixtures/ssr-dynamic/src/pages/api/products/%5Bid%5D.js#L4
> Look here.
> We will break this test If we directly skip generating a file.

It shouldn't break that test because that file doesn't have a `getStaticPaths` though, so things should still work as-is. We only need to provide an error in dev and build if:

1. The endpoint ends with `.js`/`.ts`, and not `.json.js` etc
2. And the endpoint has `getStaticPaths` with `undefined` value

---

I ran this PR locally and it seems to be applying this idea:

> I don't think we can remove support for `undefined` yet because there are actual use case for it with slugs. Even though it has the conflict behaviour, e.g. `[...slug].astro` (with undefined) and `index.astro`, I think that is still correct because you're opting in to that. In which case we should log a warning that there's a conflict instead (probably future PR).

which is great! But I think we could tweak it as a warning for now, since it doesn't really break anything, but only cause confusing behaviour for users. Ideally in dev and build, if:

1. The endpoint or Astro file has `getStaticPaths` (regardless of `undefined`)
2. And there's another file matches that path, then log a warning.

I realize this case isn't for `undefined` only as you could have `/[...slug].astro` and `/blog.astro`. If `[...slug].astro` has `getStaticPaths` of `blog`, then there's a clash too.

I think we could really split these two issues as two PRs though, depending on how you'd like to tackle them.

---

I have a feeling that the second one might be trickier to do, but the first one would be easier. Happy to discuss further if you have any questions!

wulinsheng123:
image
image

wulinsheng123:
@bluwy 
I think you wouldn't include a demo if you added this condition `route.type === 'endpoint' && mod.getStaticPaths`.

https://stackblitz.com/edit/github-upsdcx-f9dvfq?file=README.md

bluwy:
That's the second solution that we discussed about, which I suggested to do in another PR. That's lower priority since it doesn't error on build. Btw gonna mark this as draft for a bit as I noticed some more things to fix.

packages/astro/src/core/render/core.ts
```diff
@@ -35,6 +35,29 @@ export async function getParamsAndProps(
...
+				// is `undefined`, throw an error as we can't generate the `/api` file and `/api` directory
+				// at the same time. Using something like `[slug].json.ts` instead will work.
+				if (route.type === 'endpoint' && mod.getStaticPaths) {
+					const lastSegment = route.segments[route.segments.length - 1];
+					// Check last segment is solely `[slug]` or `[...slug]` case (dynamic). Make sure it's not
+					// `foo[slug].js` by checking segment length === 1. Also check here if the params are undefined.
+					if (
+						lastSegment.length === 1 &&
+						lastSegment[0].dynamic &&
+						Object.values(params).some((v) => v === undefined)
```
bluwy, line 49: TODO: I think this could be improved a little to check the params of the last segment only. Maybe assuming `params` is ordered and getting the last value could be enough.

bluwy (commented):
Noticed some more changes to be made, I can make them when I get around if you'd like.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -454,6 +454,38 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 *		 { params: { slug: undefined },
+	 * 		 { params: { slug: "about" } }
+	 * 	 ];
+	 * }
+	 * ```
+	 *
+	 * When prerendering this endpoint, Astro will generate both the `/api` file and `/api/about` file. As `/api`
+	 * is both a file and directory in the filesystem, it will cause a path collision when building.
+	 */
+	PrerenderDynamicEndpointPathCollide: {
```
bluwy, line 478: TODO: this entry needs to be moved down, and rename the code to avoid conflict.

wulinsheng123:
@bluwy I have done but I think we need to merge quickly. T.T
now there is a conflict.

bluwy:
This should be good now. Moving this to review 👍

wulinsheng123:
> This should be good now. Moving this to review 👍

thank

.changeset/small-knives-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Throw better error when a dynamic endpoint without additional extensions is prerendered with `undefined` params. For example, `src/pages/blog/[slug].js` and `src/pages/blog/[...slug].js` can't have a `getStaticPaths` function that returns `undefined` for `slug`, Astro would generate both a `/blog` file and `/blog` directory (for thruthy slugs), causing a build error. To fix this, you can add an extension like `[...slug].json.js` to generate `/blog.json` instead.
```
sarah11918, line 5:
```suggestion
Throw better error when a dynamic endpoint without additional extensions is prerendered with `undefined` params.
```
I think for the changeset, we'd normally just stick to something like this? The change is "a better error." I don't *think* you need to explain in such detail?

If I'm wrong, ping me again and I'll fix a typo/run on sentence!  :smile:

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -483,6 +483,38 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 			)} are supported for optimization.`,
 		hint: "If you do not need optimization, using an `img` tag directly instead of the `Image` component might be what you're looking for.",
 	},
+	/**
+	 * @docs
+	 * @see
+	 * - [`getStaticPaths()`](https://docs.astro.build/en/reference/api-reference/#getstaticpaths)
+	 * - [`params`](https://docs.astro.build/en/reference/api-reference/#params)
+	 * @description
```
sarah11918, line 491: This description doesn't really follow our standard format for error messages. Erika has written a helpful guide here so you can get a feel for the style we're aiming for: https://github.com/withastro/astro/blob/main/packages/astro/src/core/errors/README.md

This feels more like teaching someone *about* what's happening vs being user-focused by saying what went wrong and what the user needs to do next to resolve the error.

Please take a look at the error writing guide and try following that model. I'll take another look, and be happy to help, when this is a bit more user-focused!  🙌

sarah11918 (changes requested):
Thanks for this contribution @wulinsheng123, I do love error messages!  🥳 

See my comment below with the link to the error writing guide so we can work on getting this in the most helpful shape possible for our confused, frustrated users who might encounter this error!

packages/astro/src/core/render/core.ts
```diff
@@ -35,6 +35,27 @@ export async function getParamsAndProps(
 			const paramsMatch = route.pattern.exec(pathname);
 			if (paramsMatch) {
 				params = getParams(route.params)(paramsMatch);
+
+				// If we have an endpoint at `src/pages/api/[slug].ts` that's prerendered, and the `slug`
+				// is `undefined`, throw an error as we can't generate the `/api` file and `/api` directory
+				// at the same time. Using something like `[slug].json.ts` instead will work.
+				if (route.type === 'endpoint' && mod.getStaticPaths) {
+					const lastSegment = route.segments[route.segments.length - 1];
```
ematipico, line 43: Could `lastSegment` be `undefined`? Do we have at least one segment here?

ematipico (approved):

packages/astro/src/core/render/core.ts
wulinsheng123: @bluwy help check

bluwy: Yeah since we're handling a dynamic route, it would always have a segment in the path here, so it should not be `undefined`

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -532,6 +532,26 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * You should add an additional extension to the endpoint's filename 
+	 * and A dynamic endpoint with `getStaticPaths` should not return `params` with `undefined`.
```
bluwy, line 542: Thanks for updating this part @wulinsheng123. I'm checking the errors-data readme again, I think it is clearer when understanding how this gets rendered by docs:

image

Knowing that this gets rendered under "What went wrong?", maybe we could update the description to answer it like:
```suggestion
	 * @description
	 * The endpoint is prerendered with an `undefined` param but the generated file and folder names will collide.
	 * An additional extension can be added to the endpoint file name to generate the file with a different name.
	 * For example, renaming `pages/api/[slug].ts` to `pages/api/[slug].json.ts`.
```

I _think_ this follows the message format. cc @sarah11918 if this is right

wulinsheng123: > Thanks for updating this part @wulinsheng123. I'm checking the errors-data readme again, I think it is clearer when understanding how this gets rendered by docs:
> 
> image
> 
> Knowing that this gets rendered under "What went wrong?", maybe we could update the description to answer it like:
> 
> I _think_ this follows the message format. cc @sarah11918 if this is right

My bad.
I will remember it. I'll fix it tomorrow

wulinsheng123: Thanks for your suggestion

wulinsheng123:
@altano cc @bluwy

bluwy (approved):
Changes now look good to me! One more pass from docs and we should be good to go.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,27 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
 	},
+	/**
+	 * @docs
+	 * @see
+	 * - [`getStaticPaths()`](https://docs.astro.build/en/reference/api-reference/#getstaticpaths)
+	 * - [`params`](https://docs.astro.build/en/reference/api-reference/#params)
+	 * @description
+	 * The endpoint is prerendered with an `undefined` param but the generated file and folder names will collide.
+	 *
+	 * An additional extension can be added to the endpoint file name to generate the file with a different name. For example, renaming `pages/api/[slug].ts` to `pages/api/[slug].json.ts`.
```
sarah11918, line 541:
```suggestion
		 * The endpoint is prerendered with an `undefined` param so the generated path will collide with another route.
	 *
	 * If you cannot prevent passing `undefined`, then an additional extension can be added to the endpoint file name to generate the file with a different name. For example, renaming `pages/api/[slug].ts` to `pages/api/[slug].json.ts`.
```
So much better, folks! Would you say this slight tweak is still accurate? @bluwy @wulinsheng123 (Just also looking below at the message which suggests that not passing `undefined` is a way out of this error.)

bluwy: Sounds good to me too 👍

PR #6356, opened 2023-02-24
feat(cli): add `--watch` to `astro check` command

ematipico:
## Changes

See https://github.com/withastro/roadmap/issues/473 for more context.

This PR adds a new argument to the command `astro check` called `--watch`.

Coding-wise, I refactored the file, but the command's core logic is the same.

This PR  takes advantage of the `watcher` provided by `vite`. `vite` uses [`chokidar`](https://www.npmjs.com/package/chokidar) as a file watcher. When a file is added/modified, we upsert the document and run the checker. When a file is deleted - `unlink` - we remove and rerun the checker.

I had to change the part where we created the `vite` configuration. Passing `watch.ignore: undefined` was breaking the watcher. 

## Testing

I added a new test case to make sure we receive an error if we modify a file that was previously valid.

I also tested the feature manually, locally.

## Docs

Requires documentation


ematipico:
!preview check-watch

packages/astro/test/cli.test.js
```diff
@@ -1,8 +1,15 @@
...
+import { cli, parseCliDevStart, cliServerLogSetup, loadFixture } from './test-utils.js';
+import stripAnsi from 'strip-ansi';
+import { promises as fs, writeFileSync, readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { isIPv4 } from 'node:net';
+import { join } from 'node:path';
+import { Writable } from 'node:stream';
+
+export function wait(ms) {
+	return new Promise((resolve) => setTimeout(resolve, ms));
```
matthewp, line 11: is this util being used anywhere?

packages/astro/test/fixtures/astro-check-watch/tsconfig.json
```diff
@@ -0,0 +1,2 @@
+// Having a `tsconfig.json`, even empty speeds up the test massively since TypeScript does not have to look for one
```
matthewp, line 1: good comment to add

packages/astro/test/test-utils.js
```diff
@@ -141,13 +143,45 @@ export async function loadFixture(inlineConfig) {
...
+		 * 		});
+		 * 		let { check, stop, watch } = fixture.check({
+		 * 			flags: { watch: true },
+		 * 		});
+		 * ```
+		 * @param opts {any} Override options sent to the check command
+		 */
+		check: async (opts) => {
+			checkServer = await check(settings, { logging, viteServer: devServer, ...opts });
+			return {
```
matthewp, line 172: any reason not to just return `checkServer` here? Since the object you are returning mirrors its API anyways.

matthewp:
Looks really good overall! Just left a few comments.

packages/astro/test/test-utils.js
ematipico: True, it's some leftover code. It makes sense to return the server itself! Good catch!

packages/astro/test/cli.test.js
ematipico: Leftover code, a good catch!

matthewp (approved):

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,339 @@
...
+import { createServer, ViteDevServer } from 'vite';
+import { createVite } from '../../core/create-vite.js';
+import type { CreateViteOptions } from '../../core/create-vite';
+import type { SyncOptions, ProcessExit } from '../../core/sync';
+import fsMod from 'fs';
 
-interface Result {
+type DiagnosticResult = {
 	errors: number;
 	// The language server cannot actually return any warnings at the moment, but we'll keep this here for future use
```
Princesseuh, line 22: This comment is actually no longer true, as it now returns warnings in some cases from the compiler

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,339 @@
...
+	let checkFlags = parseFlags(flags);
+	let options: CreateViteOptions = { settings, logging, mode: 'build', fs };
+	if (checkFlags.watch) {
+		info(logging, 'check', 'Checking files in watch mode');
+		options.isWatcherEnabled = true;
+	} else {
+		info(logging, 'check', 'Checking files');
+	}
+	// We create a server to start doing our operations
+	const viteServer = await createServer(
```
Princesseuh, line 81: Just wondering, was there a specific reason a Vite server was chosen over other similar tools (chokidar or other watchers)? Does it have a better understanding of the project perhaps?

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,339 @@
...
+	logging: Readonly<LogOptions>;
+
+	fileSystem: typeof fsMod;
+};
+
+/**
+ * Responsible to check files - classic or watch mode - and report diagnostics.
+ *
+ * When in watch mode, the class does a whole check pass, and then starts watching files.
+ * When a change occurs to an `.astro` file, the checker
```
Princesseuh, line 130: Unfinished sentence?

packages/astro/src/cli/check/index.ts
matthewp: I suggested Vite just because we already use it's file watcher for dev, so it's going to work the same, and be one fewer dep.

ematipico: Ditto! One thing that I don't like very much is that while we listen to file changes, we also run the `sync` operation, which also uses a vite server under the hoods. 

I evaluated the idea of using just one server for both operations, but that didn't turn out very well:
- content collections and check servers use different options/settings;
- I don't think it's possible to change options when a server is instantiated;

bluwy: Actually I'm with Erika here too. `createServer` takes a lot of resource to start up, it also sets up a different chokidar instance so they aren't really shared in terms of performance. Plus according to https://github.com/withastro/roadmap/discussions/502, we might also want to make `astro check` a separate package, and it would be quite heavy to rely on Vite instead of only chokidar.

So maybe it's better that we use chokidar directly for now.

ematipico: @Princesseuh @bluwy Please have another review. I just wanted to let you know that I applied your suggestion. 

> createServer takes a lot of resource to start up

Is there some article/resource/finding that I can read about it?

bluwy: > Is there some article/resource/finding that I can read about it?

It's mostly through reading what Vite does when calling `createServer()` ([code](https://github.com/vitejs/vite/blob/38ce81ceb86d42c27ec39eefa8091f47f6f25967/packages/vite/src/node/server/index.ts#L338)), which is a lot of stuff 😬

packages/astro/src/core/build/index.ts
```diff
@@ -68,7 +68,6 @@ class AstroBuilder {
 			logging,
 		});
 		this.manifest = createRouteManifest({ settings: this.settings }, this.logging);
-
```
bluwy, line 71: I think we can undo this line removal, as it makes reading the code harder when compact.

packages/astro/src/core/build/index.ts
```diff
@@ -80,7 +79,6 @@ class AstroBuilder {
 			{ settings: this.settings, logging, mode: 'build', command: 'build' }
 		);
 		await runHookConfigDone({ settings: this.settings, logging });
-
```
bluwy, line 83: Here as well

packages/astro/test/test-utils.js
```diff
@@ -141,13 +143,33 @@ export async function loadFixture(inlineConfig) {
...
+		 * ```js
+		 * 		let fixture = await loadFixture({
+		 * 			root: './fixtures/astro-check-watch/',
+		 * 		});
+		 * 		let { check, stop, watch } = fixture.check({
+		 * 			flags: { watch: true },
+		 * 		});
+		 * ```
+		 * @param opts {any} Override options sent to the check command
+		 */
```
bluwy, line 169: This JSDoc should be moved up following the previous comment

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+ *
+ * If the `--check` flag is provided, the command runs indefinitely and provides diagnostics
+ * when `.astro` files are modified.
+ *
+ * Every time an astro files is modified, content collections are also generated.
+ *
+ * @param {AstroSettings} settings
+ * @param {CheckPayload} options
+ * @param {Flags} options.flags
+ * @param {LogOptions} options.logging
```
bluwy, line 77: I'm not sure about these JSDoc's, as they're duplicating TypeScript types and might cause a chance of it being outdated. I think we can remove it if we're not adding documentation for a specific param.


There's some other similar pattern in this file which I think we can remove.

ematipico:
I had to disable the test. It became unreasonable: very few times it passes, then it fails. Locally it passes every time. Unfortunately logs are not written to console when a test fails, so it gets difficult to debug it.

With vite server implementation, it worked like a charm.

packages/astro/src/cli/check/index.ts
Princesseuh: That's also my opinion on JSDoc, it barely improves the hover information unless you add descriptions to every property and often just ends up being a (often outdated) duplicate of the types. Not a fan

ematipico: This is more about hinting the correct information to the user, without navigating too much the "types":

Without
Screenshot 2023-03-03 at 15 29 45

With 
Screenshot 2023-03-03 at 15 30 05

I'm OK with removing the JSDoc, but I believe only TypeScript types are not enough to help developers.

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
 import glob from 'fast-glob';
 import * as fs from 'fs';
 import { bold, dim, red, yellow } from 'kleur/colors';
 import { createRequire } from 'module';
 import ora from 'ora';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { printDiagnostic } from './print.js';
+import type { Arguments as Flags } from 'yargs-parser';
+import { debug, info } from '../../core/logger/core.js';
+import type { SyncOptions, ProcessExit } from '../../core/sync';
```
Princesseuh, line 14: This is not a problem because it's a type only import, but we typically only use `.js` imports

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+import fsMod from 'fs';
+import type { FSWatcher } from 'chokidar';
+import { join } from 'node:path';
 
-interface Result {
+type DiagnosticResult = {
```
Princesseuh, line 19: Any reasons for using a type instead of an interface here?

packages/astro/src/cli/check/index.ts
matthewp: I don't think we've really standardized that yet. For type imports I often omit it. If we're all doing it differently we should pick one, I'd be happy with either way. Maybe always using `.js` safer.

Princesseuh: Considering we're forced to use `.js` imports for normal imports, imo we should do the same for all imports

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+	Listen,
+}
+
+const ASTRO_GLOB_PATTERN = '**/*.astro';
 
-	const spinner = ora(` Getting diagnostics for Astro files in ${fileURLToPath(root)}…`).start();
+/**
+ * Checks `.astro` files for possible errors.
+ *
+ * If the `--check` flag is provided, the command runs indefinitely and provides diagnostics
```
Princesseuh, line 69:
```suggestion
 * If the `--watch` flag is provided, the command runs indefinitely and provides diagnostics
```

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+
+	/**
+	 * Stops the watch. It terminates the inner server.
+	 */
+	public async stop() {
+		await this.#watcher?.close();
+	}
+
+	/**
+	 * Weather the checker should run in watch mode
```
Princesseuh, line 186:
```suggestion
	 * Whether the checker should run in watch mode
```

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+			});
+		});
+
+		return result;
+	}
 }
 
 /**
- * Open all Astro files in the given directory and return the number of files found.
+ * Open all Astro files in the given directory and return the number of files found.*
```
Princesseuh, line 340:
```suggestion
 * Open all Astro files in the given directory and return the number of files found.
```

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
 ): Promise<number> {
-	const files = await glob('**/*.astro', {
+	const files = await glob(ASTRO_GLOB_PATTERN, {
 		cwd: fileURLToPath(workspaceUri),
 		ignore: ['node_modules/**'].concat(filePathsToIgnore.map((ignore) => `${ignore}/**`)),
 		absolute: true,
 	});
 
 	for (const file of files) {
+		debug('check', `Adding file ${file} to the list of files to check.`);
```
Princesseuh, line 357: I like this 😄

packages/astro/src/cli/check/index.ts
ematipico: `interface` comes with the feature to use `extends`/`implements`, which is not a feature I'm interested in. Generally, `type` is "lighter" compared to interfaces. Usually, I use `type` unless I need the interface features!

packages/astro/src/cli/check/index.ts
ematipico: `import type` is elided at compilation time, so I thought it didn't matter, and I let the IDE do its thing without thinking about it. I can add `.js` for consistency!

ematipico: It turns out that adding `.js` emits the TS error... I have no idea why.

Princesseuh (approved):
Looks great, awesome work!

matthewp:
!preview check-watch

packages/astro/test/test-utils.js
```diff
@@ -141,13 +162,17 @@ export async function loadFixture(inlineConfig) {
 
 	let fixtureId = new Date().valueOf();
 	let devServer;
+	let checkServer;
```
bluwy, line 165: Looks like we can remove this as it's unused.

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+		this.#syncCli = syncCli;
+		this.#logging = logging;
+		this.#settings = settings;
+		this.#fs = fileSystem;
+		this.#filesCount = 0;
+	}
+
+	/**
+	 * Check all `.astro` files once and then finishes the operation.
+	 * @returns {Promise<CheckResult>}
```
bluwy, line 162: I think we can remove this `@return` type as it's already annotated in TypeScript below.

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+	 */
+	public async check(): Promise<CheckResult> {
+		return await this.#checkAllFiles(true);
+	}
+
+	/**
+	 * Check all `.astro` files and then start watching for changes.
+	 * @returns {Promise<CheckResult>}
```
bluwy, line 170: This can be removed too

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+	/**
+	 * Stops the watch. It terminates the inner server.
+	 */
+	public async stop() {
+		await this.#watcher?.close();
+	}
+
+	/**
+	 * Whether the checker should run in watch mode
+	 * @returns {boolean}
```
bluwy, line 187: Here as well.

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+	}
+
+	/**
+	 * Lint all `.astro` files, and report the result in console. Operations executed, in order:
+	 * 1. Compile content collections.
+	 * 2. Optionally, traverse the file system for `.astro` files and saves their paths.
+	 * 3. Get diagnostics for said files and print the result in console.
+	 *
+	 * @param openDocuments Whether the operation should open all `.astro` files
+	 * @private
```
bluwy, line 208: I'm not sure if this `@private` is needed as the function is already prefixed with `#`. The other private functions doesn't have this annotation too.

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+		this.#diagnosticsChecker.upsertDocument({
+			uri: pathToFileURL(filePath).toString(),
+			text,
+		});
+	}
+
+	/**
+	 * Logs the result of the various diagnostics
+	 *
+	 * @param {Readonly<DiagnosticResult>} result Result emitted by AstroChecker.#breakDownDiagnostics
```
bluwy, line 288: I think we can remove the duplicated types here too:
```suggestion
	 * @param result Result emitted by AstroChecker.#breakDownDiagnostics
```

packages/astro/src/cli/check/index.ts
```diff
@@ -1,101 +1,360 @@
...
+
+		return result;
+	}
 }
 
 /**
  * Open all Astro files in the given directory and return the number of files found.
+ * @param {URL} workspaceUri
+ * @param {string[]} filePathsToIgnore
+ * @param {AstroCheck} checker
```
bluwy, line 343: I think we can remove these too as they are already annotated in TypeScript below.

bluwy (commented):
The flow looks great to me. I have some nitpicks again about the jsdoc 😬 I'm not sure if it's good to have JSDoc overlap with TypeScript if we're already using TypeScript as the primary way to document types.

bluwy (approved):
Awesome. Thanks for making the changes 💚 Looking forward for this to land!

Issue #6336, opened 2023-02-23
Tailwind is not loaded for layout of markdown pages


hieudo-dev:
### What version of `astro` are you using?

2.0.14

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

I'm facing an issue where TailwindCSS is not available in the Markdown pages of my app.

I created some `.md` pages that has my content and I specified a layout component through the property `layout` like below:
```astrojs
---
layout: ../layouts/Base.astro
---
```
the **Base.astro** component uses TailwindCSS but I when access my pages in the browser none of the TailwindCSS classes work.

I'm note sure if this is a bug or a misusage on my part. As of now im using content collections and the `astro:content` to work around this. A reproducible version is provided below.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xkptsr-yprlfg

### Participation

- [X] I am willing to submit a pull request for this issue.

rafael-pinheiro:
Also struggling with that.
It's worse that it only happens on build and not on dev. Found out after deploy =/

matthewp:
I think this may be the same issue as https://github.com/withastro/astro/issues/6318 Will add a test for both

Issue #6318, opened 2023-02-20
Astro.slots.render() makes the link tag be inserted in the body at build time


tetracalibers:
### What version of `astro` are you using?

2.0.14

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm, yarn

### What operating system are you using?

Mac

## Summary

If I use `Astro.slots.render('default')` to extract the slots and `set:html` to display them, the CSS `link` tags extracted at build time will be inserted within the `body` tag.

```astro
---
const content = await Astro.slots.render('default')
---

<Fragment set:html={content} />
```

When displayed using the `slot` tag, the CSS is inserted within the `head` tag.

```astro
<slot />
```

This problem occurs when an `astro build` is performed.
（When `astro dev` is executed, both cases are normal.）

## Demo

After viewing it in a browser, open the developer tools and see if the `link` tag is in the `body` tag or in the `head` tag.

### use `Astro.slots.render('default')`

- https://tetracalibers.github.io/astro-slot-bug-test/slots-render/

The `link` tag is inserted within the `body` tag.

<img width="563" alt="use-astro-slots-render" src="https://user-images.githubusercontent.com/92695929/220084763-b02e66a4-e1ba-4c32-b500-2535a0ef22b0.png">

### use `<slot />`

- https://tetracalibers.github.io/astro-slot-bug-test/slot-tag/

The `link` tag is inserted within the `head` tag.

<img width="563" alt="use-slot-tag" src="https://user-images.githubusercontent.com/92695929/220084934-736dcb0d-e2b3-497a-b2c9-b237a0913dcf.png">

## Expected Results

The `link` tag that loads the CSS file should be placed within the `head` tag.

### Problems with initial site display

The placement of the `link` tag within the `body` tag delays the loading of the CSS file and disrupts the initial display of the site.

### Problems with `astro-icon` package's `Sprite` feature

The internal implementation of `Sprite.Provider` in [astro-icon](https://github.com/natemoo-re/astro-icon) uses `Astro.slots.render('default')`.

- https://github.com/natemoo-re/astro-icon/blob/main/packages/core/lib/SpriteProvider.astro

Users of astro-icon's sprite functionality will have problems with initial display turbulence.

- https://github.com/natemoo-re/astro-icon/issues/83

### Link to Minimal Reproducible Example

https://github.com/tetracalibers/astro-slot-bug-test

### Participation

- [ ] I am willing to submit a pull request for this issue.

[...]

PR #6363, opened 2023-02-24
Only apply head propagation in trees that need it

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6318
- Fixes https://github.com/withastro/astro/issues/6336
- We were mistakenly marking *all* modules as need head propagation. This should only be applied to those that have head propagation in their tree (currently this is only content collection usage).

## Testing

- New test case added.

## Docs

N/A, bug fix


natemoo-re (approved):
Nice tests!

PR #6366, opened 2023-02-24
Do not apply user plugins when spawning temporary Vite server

### LINKED ISSUE #6364: Running Vite server during build breaks plugin assumptions

2.0.12+

N/A

pnpm

macOS

### Describe the Bug

In https://github.com/withastro/astro/pull/6238, we broke [`vite-plugin-pwa`](https://github.com/vite-pwa/vite-plugin-pwa/) (and possibly others) because we started running a temporary Vite server during the build to load Content Collections. `vite-plugin-pwa` uses [`apply: 'serve'`](https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L32) as a heuristic for running the dev server, which no longer works when we start a temporary server _and pass all plugins_ to that server.

cc @userquin 

https://stackblitz.com/edit/github-kyecnu?file=integration.js,astro.config.mjs,package.json

### Participation

natemoo-re:
## Changes

- Closes #6364
- Adds flag to `createVite` that disallows user-provided plugins. We need to do this when spawning a temporary server during `astro build` because some Rollup/Vite plugins are stateful and have different behavior if a server is detected.
- cc @bholmesdev not sure if this has unintended consequences!

## Testing

Tested manually with `vite-plugin-pwa` repro

## Docs

Bug fix only

### DIFF
--- a/.changeset/quick-items-switch.md
+++ b/.changeset/quick-items-switch.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix regression that caused some stateful Vite plugins to assume they were running in `dev` mode during the `build`.
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -31,6 +31,7 @@ interface CreateViteOptions {
 	logging: LogOptions;
 	mode: 'dev' | 'build' | string;
 	fs?: typeof nodeFs;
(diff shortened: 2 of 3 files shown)


userquin:
It seems any plugin also being applied on `preview` command :grimacing:

natemoo-re:
Closing in favor of #6368

PR #6368, opened 2023-02-25
fix: don't mix Vite plugins when spawning temporary Vite server

### LINKED ISSUE #6364: Running Vite server during build breaks plugin assumptions

userquin:
## Changes

- Closes #6364
- Supersedes #6366
- Filter plugins using the Astro command
- /cc @bholmesdev
- /cc @natemoo-re

## Testing

Added simple test with disparate plugins for static build.

## Docs

Bug fix only: maybe we need to add some hints for integrations registering plugins, check for example `@vite-pwa/astro` integration logic: https://github.com/vite-pwa/astro/blob/main/src/index.ts#L31

### DIFF
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -172,14 +172,21 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
-	// command will be undefined when running sync
+	// PR #6238 includes `astro sync` command, which is not a Vite command with an additional vite server.
+	// AstroBuilder::setup: will call createVite twice:
+	// - with `command` set to `build/dev` (src/core/build/index.ts L72)
+	// - and again in the `sync` module to generate `Content Collections` (src/core/sync/index.ts L36)
+	// We need to check if the command is `build` or `dev` before merging the user-provided vite config.
+	// We also need to filter out the plugins that are not meant to be applied to the current command:
+	// - If the command is `build`, we filter out the plugins that are meant to be applied for `serve`.
+	// - If the command is `dev`, we filter out the plugins that are meant to be applied for `build`.
 	if (command) {
 		let plugins = settings.config.vite?.plugins;
 		if (plugins) {
 			const { plugins: _, ...rest } = settings.config.vite
 			const applyToFilter = command === 'build' ? 'serve' : 'build'
 			const applyArgs = [{...settings.config.vite, mode}, { command, mode }]
-			// @ts-ignore we know what are doing
+			// @ts-expect-error ignore TS2589: Type instantiation is excessively deep and possibly infinite. 
 			plugins = plugins.flat(Infinity).filter((p) => {
 				if (!p || p?.apply === applyToFilter) {
 					return false;
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -172,8 +172,8 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
-	// PR #6238 includes `astro sync` command, which is not a Vite command with an additional vite server.
-	// AstroBuilder::setup: will call createVite twice:
+	// PR #6238 Calls user integration `astro:config:setup` hooks when running `astro sync`.
+  // Without proper filtering, user integrations may run twice unexpectedly:
 	// - with `command` set to `build/dev` (src/core/build/index.ts L72)
 	// - and again in the `sync` module to generate `Content Collections` (src/core/sync/index.ts L36)
 	// We need to check if the command is `build` or `dev` before merging the user-provided vite config.
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -170,7 +170,28 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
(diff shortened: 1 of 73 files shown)


userquin:
We need the command, not the mode.

userquin:
for preview, integrations should take care, and for vite user plugins we should also filter them before starting Vite preview

packages/astro/src/core/create-vite.ts
```diff
@@ -170,7 +172,31 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
-	result = vite.mergeConfig(result, settings.config.vite || {});
+	// command will be undefined when running sync
+	if (command && command !== 'preview') {
+		let plugins = settings.config.vite?.plugins;
+		if (plugins) {
+			const { plugins: _, ...rest } = settings.config.vite
+			const applyToFilter = mode === 'build' ? 'serve' : 'build'
```
userquin, line 180: this should be `const applyToFilter = command === 'build' ? 'serve' : 'build'`

userquin: we can `build` in `dev` mode

packages/astro/src/core/create-vite.ts
```diff
@@ -30,6 +30,8 @@ interface CreateViteOptions {
 	settings: AstroSettings;
 	logging: LogOptions;
 	mode: 'dev' | 'build' | string;
+	// will be undefined when using `sync`
+	command?: 'dev' | 'build' | 'preview';
```
userquin, line 34: remove `'preview'`, Astro will use Vite `preview`

packages/astro/src/core/create-vite.ts
```diff
@@ -170,7 +172,31 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
-	result = vite.mergeConfig(result, settings.config.vite || {});
+	// command will be undefined when running sync
+	if (command && command !== 'preview') {
```
userquin, line 176: onde remoed `'preview'` from `CreateViteOptions` changethis to `if (command) {`

userquin:
don't merge yet, preparing test

userquin:
We've another problem, when a plugin is included it can be excluded on sync call since we don't have command/action when using apply with function, we need to review it.

packages/astro/src/core/create-vite.ts
```diff
@@ -170,7 +172,31 @@ export async function createVite(
...
+				}
+
+				return true;
+			});
+
+			result = vite.mergeConfig(result, { ...rest, plugins });
+		} else {
+			result = vite.mergeConfig(result, settings.config.vite || {});
+		}
+	}
```
userquin, line 199: add else case without plugins

natemoo-re (approved):
Great fix and awesome test! 

Test failure is unrelated.

packages/astro/src/core/create-vite.ts
```diff
@@ -170,7 +172,31 @@ export async function createVite(
...
 	let result = commonConfig;
-	result = vite.mergeConfig(result, settings.config.vite || {});
+	// command will be undefined when running sync
+	if (command) {
+		let plugins = settings.config.vite?.plugins;
+		if (plugins) {
+			const { plugins: _, ...rest } = settings.config.vite
+			const applyToFilter = command === 'build' ? 'serve' : 'build'
+			const applyArgs = [{...settings.config.vite, mode}, { command, mode }]
+			// @ts-ignore we know what are doing
```
ematipico, line 182: Please use `@ts-expect-error`, so we can remove the comment if the compiler gets better.

Would mind changing the comment too? Just saying "we know what we are doing" doesn't seem useful to other developers. Can we explain _what_ we are doing?

packages/astro/src/core/create-vite.ts
```diff
@@ -170,7 +172,31 @@ export async function createVite(
 	//   3. integration-provided vite config, via the `config:setup` hook
 	//   4. command vite config, passed as the argument to this function
 	let result = commonConfig;
-	result = vite.mergeConfig(result, settings.config.vite || {});
+	// command will be undefined when running sync
+	if (command) {
```
ematipico, line 176: I'd like to see a comment that explains what this new code does.

ematipico (commented):
I'm not against the change, but I'd like to see better comments

packages/astro/src/core/create-vite.ts
userquin: imagen

userquin: we need to traverse flat plugins array to filter them

ematipico (approved):

bholmesdev (commented):
Added some suggestions too clear up in-line comments. Great fix though, was unaware Vite plugins have an `apply` utility!

packages/webapi/mod.d.ts
```diff
@@ -1,13 +1,13 @@
 // organize-imports-ignore
```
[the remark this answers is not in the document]
natemoo-re: I can't figure out why this is here. Git seems to think the file is up-to-date with `main`. Resolving so we can merge.

PR #6369, opened 2023-02-25
Update doc link in non-html-pages example README

weakish:
## Changes

Replace the following dead link:

/en/core-concepts/astro-pages/#non-html-pages

Also, mentions `.js` or `.ts` instead of `.md`,
since this is a non-html-pages example (`.md` are for html pages). 

## Testing

Update README, thus no tests are added.

## Docs

This commit itself is a documentation change.


delucis (approved):
Thanks for catching that @weakish!

PR #6376, opened 2023-02-27
Refactor benchmark script

bluwy:
## Changes

Update root `pnpm benchmark` script and CI `!bench` command. See https://github.com/bluwy/astro/pull/2 for examples, e.g. https://github.com/bluwy/astro/pull/2#issuecomment-1448419909

Also added a new `/benchmark` folder with READMEs within it that should hopefully help clarify how it works.

The Astro codebase is also tweaked a bit to collect timing/memory logs via `process.env.ASTRO_TIMER_PATH`, and if it exists collect and write the results to the path. This makes passing the results easier between child processes. It's only used for benchmarking.

---

Locally, you can run `pnpm benchmark`, `pnpm benchmark memory`, `pnpm benchmark server-stress`, etc

In PR comments, you can comment `!bench`, `!bench memory`, `!bench server-stress`, etc

## Testing

Manual testing

## Docs

n/a. This doesn't touch on user documentation


bluwy:
!bench

bluwy:
Looks like I need to fork this to properly test the workflow. Now it's using the main workflow.

natemoo-re:
> Looks like I need to fork this to properly test the workflow. Now it's using the main workflow.

Yep that's correct. I use a personal fork for testing actions/CI.

matthewp (approved):
Looks great! Excited for this.

benchmark/bench/server-stress.js
```diff
@@ -0,0 +1,82 @@
...
+
+	console.log('Writing results to', fileURLToPath(outputFile));
+	await fs.writeFile(outputFile, JSON.stringify(result, null, 2));
+
+	console.log('Result preview:');
+	console.log('='.repeat(10));
+	console.log(`#### Server stress\n\n`);
+	let text = autocannon.printResult(result);
+	if (process.env.CI) {
+		text = text.match(/^.*?requests in.*?read$/m)?.[0];
```
ematipico, line 50: Would you mind adding a small comment explaining what we're doing here and why?

benchmark/ci-helper.js
```diff
@@ -0,0 +1,10 @@
+const benchLogs = process.argv[2];
```
ematipico, line 1: Could you add a comment to this file explaining what it is for? It seems to be a particular file used only in CI for doing something that I can't understand 😅

packages/astro/src/core/config/timer.ts
```diff
@@ -0,0 +1,65 @@
...
+	startHeap: number;
+}
+
+/**
+ * Timer to track certain operations' performance. Used by Astro's scripts only.
+ * Set `process.env.ASTRO_TIMER_PATH` truthy to enable.
+ */
+export class AstroTimer {
+	private enabled: boolean;
+	private ongoingTimers: Map<string, OngoingStat> = new Map();
```
ematipico, line 21: Maybe we could use `WeakMap`? It doesn't seem we iterate them

ematipico (approved):

benchmark/make-project/README.md
```diff
@@ -0,0 +1,7 @@
+# make-project
+
+This `make-project` folder contains different files to programmatically create a new Astro project. They are created inside the `projects` folder and are gitignored. These projects are used by benchmarks for testing.
+
+Each benchmark can specify the default project to run in it's `defaultProject` export, but it can be overriden if `--project <project-name>` is passed through the CLI.
```
sarah11918, line 5:
```suggestion
Each benchmark can specify the default project to run in its `defaultProject` export, but it can be overriden if `--project <project-name>` is passed through the CLI.
```

sarah11918 (commented):
Just a heads up that Docs usually removes the review request to Astro Docs Maintainers ourselves! (We get pinged, so we do monitor these requests.)

We are automatically pinged whenever a README file is included in a PR because this is content that someone might... read! And even if internal, we do care about its quality.

benchmark/bench/README.md
```diff
@@ -0,0 +1,7 @@
+# bench
+
+This `bench` folder contains different benchmarking files that you can run via `astro-benchmark <bench-file-name>`, e.g. `astro-benchmark memory`. Files that start with an underscore are not benchmarking files.
+
+Benchmarking files will run against a project to measure it's performance, and write the results down as JSON in the `results` folder. The `results` folder is gitignored and it's result files can be safely deleted if you're not using it.
```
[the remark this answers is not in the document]
bluwy: Ah sorry about that! Noted 👍 Thanks for catching these grammar mistakes!

benchmark/bench/server-stress.js
bluwy: Good idea 👍 It's here mostly to truncate the markdown text in the comments so it's not huge. I'll update this.

benchmark/ci-helper.js
bluwy: Yup! I've written this mostly because I can't write the same thing in bash 😅 I'll note what it's doing

packages/astro/src/core/config/timer.ts
bluwy: Weak maps only allows objects as keys, we also need to refetch the `OngoingStat` when calling the `end()` method, so we need it to persists indefinitely as there isn't a reference to pass around.

bluwy:
Going to merge this to unblock the memory optimizations soon. Happy to improve further if there's more changes to add. Thanks for the reviews!

PR #6386, opened 2023-02-28
ci: add workflow_dispatch to workflows

ematipico:
## Changes

This is a housekeeping PR. It adds the `workflow_dispatch:` to some of our workflows. This enables the possibility to run workflows manually against specific branches. This help us with debugging github actions using a PR. 

This a screenshot of the menu we can have under the actions that have `workflow_dispatch:`

Screenshot 2023-02-28 at 17 12 07

I added the new property only to the workflows that seemed more "internal". Let me know which workflows need to be updated.

## Testing

After merging the PR, we should see the new dropdown in the workflows that are updated.

## Docs

N/A

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -3,6 +3,7 @@ name: Benchmark
 on:
   issue_comment:
     types: [created]
+  workflow_dispatch:
 
 env:
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,7 @@
 name: Examples astro check
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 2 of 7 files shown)


matthewp (approved):

.github/workflows/benchmark.yml
```diff
@@ -3,6 +3,7 @@ name: Benchmark
 on:
   issue_comment:
     types: [created]
+  workflow_dispatch:
```
bluwy, line 6: Question: Some of the workflows (like this one) rely on the `github.event` variable to work. IIUC this workflow relies on specific properties to the `issue_comment` event only, does it work for `workflow_dispatch` too?

bluwy, line 6: Question: Some of the workflows (like this one) rely on the `github.event` variable to work. IIUC this workflow relies on specific properties to the `issue_comment` event only, does it work for `workflow_dispatch` too?

bluwy: Alright. I'm happy to try this out if it would be helpful! Let's go with it.

bluwy (approved):

PR #6391, opened 2023-03-01
Teardown compiler to improve rendering perf

bluwy:
## Changes

Use `@astrojs/compiler` `teardown()` API to remove the internal WASM instance and cache after we're finished with the Vite build.

## Testing

All existing test should pass. `!bench memory` _should_ show an improvement too, but it doesn't because the technique to capture "heap used" includes the garbage collection root (for some reason).

But when testing with docs locally, I can see the optimization working where Vite build memory gets offloaded during rendering.

## Docs

n/a. internal refactor.

### DIFF
--- a/.changeset/bright-forks-add.md
+++ b/.changeset/bright-forks-add.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Teardown compiler after Vite build to free up memory when rendering pages
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -99,7 +99,7 @@
     "test:e2e:match": "playwright test -g"
   },
   "dependencies": {
-    "@astrojs/compiler": "next--teardown",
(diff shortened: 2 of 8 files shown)


bluwy:
!bench memory

bluwy:
This is good for review now. The `!bench memory` results above should be the same as the latest commit as they're using the same compiler.

natemoo-re (approved):

bluwy:
Hmm. This seems to make CI tests even longer! I guess it's a double-edge sword with the `teardown()` API, where running builds many times causes the WASM instance to instantiate more times, slowing it down 🤔 

Maybe we can skip teardown for tests for now.

bluwy:
!bench memory

natemoo-re (approved):
Change to toggle the setting makes sense, still approved! 👍

PR #6238, opened 2023-02-13
Run integration config hook on `astro sync`

bholmesdev:
## Changes

- Run `astro:config:setup` hooks on `astro sync`. This was preventing `preact-compat` from mounting in the docs repo
- Refactor: move `sync` out of `cli/` into `core/`

PR #6392, opened 2023-03-01
Fix astro sync config merge

bluwy:
## Changes

Recently merged but unreleased change https://github.com/withastro/astro/pull/6368 regresses https://github.com/withastro/astro/pull/6238. Causing docs to fail to build because integrations that updates the Vite config isn't affecting Astro's internal Vite config.

This PR makes `astro sync` to run in `build` mode, meaning plugins with `apply: 'serve'` will not run in `astro sync`. I _think_ this is what https://github.com/withastro/astro/pull/6368 was trying to fix (the issue:  https://github.com/withastro/astro/issues/6364), so this shouldn't be regressing that. And opens up the path for `astro sync` to update Astro's internal Vite config.

cc @userquin 

## Testing

I made sure the test added in https://github.com/withastro/astro/pull/6368 still passed

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -30,8 +30,7 @@ interface CreateViteOptions {
 	settings: AstroSettings;
 	logging: LogOptions;
 	mode: 'dev' | 'build' | string;
-	// will be undefined when using `sync`
-	command?: 'dev' | 'build';
+	command: 'dev' | 'build';
 	fs?: typeof nodeFs;
 }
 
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/create-vite.ts
```diff
@@ -180,32 +179,28 @@ export async function createVite(
...
 
-			result = vite.mergeConfig(result, { ...rest, plugins });
-		} else {
-			result = vite.mergeConfig(result, settings.config.vite || {});
-		}
+			return true;
+		});
+		result = vite.mergeConfig(result, { ...rest, plugins });
+	} else {
+		result = vite.mergeConfig(result, settings.config.vite || {});
```
bluwy, line 203: The main change in this diff is that previously when `command` is not set, it means it's in `astro sync`. However, it did not run `result = vite.mergeConfig(result, settings.config.vite || {});`, causing the issue.

We can't run that `mergeConfig` if `command` is undefined because that calls Vite plugins with `apply: 'serve'` in builds. So I made `astro sync` run in `build` mode only to fix this.

Issue #4935, opened 2022-09-30
`astro build --help` should print a help message about the `astro build` command


jasikpark:
### What version of `astro` are you using?

1.4.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

If you run `npx astro build -h` it will just build and ignore the flag
If you run `npx astro build --help` it will print the global help message

I would expect that both commands print help details for the `build` command.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rq9ofs?file=example&view=editor

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
@jasikpark are you able to submit a PR for this?

jasikpark:
what would be required to make this happen?

matthewp:
Help is printed here: https://github.com/withastro/astro/blob/main/packages/astro/src/cli/index.ts so updating that to account for it being for a subcommand.

jasikpark:
okay thx -- I guess this issue is two things:

A that `-h` is ignored as a help flag for subcommands

B that it'd be great to have more in-depth help output for particular subcommands more than the single-line that's printed for global help

ematipico:
I will be working on this. I will take the opportunity to add the `--help` flag to ALL commands.

PR #6394, opened 2023-03-01
feat(cli): add help flags to various commands

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/4935

While the issue was specific to `build` command, the same was valid for other commands.

This PR adds the `--help` flag to all commands. 

## Testing

I tested manually. At the moment, it would make little sense to have a unit test. At the moment, help output and flags are disconnected, which means that adding a new flag to the logic and not to the help command would make the test still pass.

## Docs

For the description, I copied the text from the website.

cc @withastro/maintainers-docs

### DIFF
--- a/packages/astro/src/cli/check/index.ts
+++ b/packages/astro/src/cli/check/index.ts
@@ -2,7 +2,6 @@
 import { AstroCheck, DiagnosticSeverity } from '@astrojs/language-server';
 import type { AstroSettings } from '../../@types/astro';
 import type { LogOptions } from '../../core/logger/core.js';
-
 import glob from 'fast-glob';
 import * as fs from 'fs';
 import { bold, dim, red, yellow } from 'kleur/colors';
@@ -10,6 +9,8 @@ import { createRequire } from 'module';
 import ora from 'ora';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { printDiagnostic } from './print.js';
+import { Arguments } from 'yargs-parser';
+import { printHelp } from '../../core/messages.js';
 
 interface Result {
 	errors: number;
@@ -18,11 +19,25 @@ interface Result {
 	hints: number;
 }
 
-export async function check(settings: AstroSettings, { logging }: { logging: LogOptions }) {
+export async function check(
+	settings: AstroSettings,
+	{ logging, flags }: { logging: LogOptions; flags: Arguments }
+) {
+	if (flags.help || flags.h) {
+		printHelp({
+			commandName: 'astro check',
+			usage: '[...flags]',
+			tables: {
+				Flags: [['--help (-h)', 'See all available flags.']],
+			},
+			description: `Runs diagnostics against your project and reports errors to the console.`,
+		});
+		return;
+	}
 	console.log(bold('astro check'));
 
 	const { syncCli } = await import('../../core/sync/index.js');
-	const syncRet = await syncCli(settings, { logging, fs });
+	const syncRet = await syncCli(settings, { logging, fs, flags });
 	// early exit on sync failure
 	if (syncRet === 1) return syncRet;
 
--- a/packages/astro/src/cli/check/index.ts
+++ b/packages/astro/src/cli/check/index.ts
@@ -23,7 +23,7 @@ export async function check(
 	settings: AstroSettings,
 	{ logging, flags }: { logging: LogOptions; flags: Arguments }
 ) {
-	if (flags.help || flags.h) {
+	if (flags.help) {
 		printHelp({
 			commandName: 'astro check',
 			usage: '[...flags]',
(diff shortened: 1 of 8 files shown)


packages/astro/src/core/preview/index.ts
```diff
@@ -6,17 +6,35 @@ import { runHookConfigDone, runHookConfigSetup } from '../../integrations/index.
...
 }
 
 /** The primary dev action */
 export default async function preview(
 	_settings: AstroSettings,
-	{ logging }: PreviewOptions
-): Promise<PreviewServer> {
+	{ logging, flags }: PreviewOptions
+): Promise<PreviewServer | undefined> {
+	if (flags?.help) {
```
MoustaphaDev, line 24:
```suggestion
	if (flags?.help || flags?.h) {
```

packages/astro/src/core/sync/index.ts
```diff
@@ -11,13 +11,27 @@ import { getTimeStat } from '../build/util.js';
...
+import { Arguments } from 'yargs-parser';
 
 type ProcessExit = 0 | 1;
 
 export async function syncCli(
 	settings: AstroSettings,
-	{ logging, fs }: { logging: LogOptions; fs: typeof fsMod }
+	{ logging, fs, flags }: { logging: LogOptions; fs: typeof fsMod; flags?: Arguments }
 ): Promise<ProcessExit> {
+	if (flags?.help) {
```
MoustaphaDev, line 23:
```suggestion
	if (flags?.help || flags?.h) {
```

MoustaphaDev (commented):
This looks great, there are just a couple of `-h` flags uncaught

packages/astro/src/core/dev/dev.ts
```diff
@@ -32,7 +34,26 @@ export interface DevServer {
...
+	if (options.flags?.help || options.flags?.h) {
+		printHelp({
+			commandName: 'astro dev',
+			usage: '[...flags]',
+			tables: {
+				Flags: [
+					['--port', `Specifies which port to run on. Defaults to 3000.`],
+					['--host', `listen on all addresses, including LAN and public addresses.`],
+					['--host <custom-address>', `expose on a network IP address at <custom-address>`],
+					['--help (-h)', 'See all available flags.'],
```
sarah11918, line 47:
```suggestion
					['--port', `Specifies which port to run on. Defaults to 3000.`],
					['--host', `listen on all addresses, including LAN and public addresses.`],
					['--host <custom-address>', `expose on a network IP address at <custom-address>`],
					['--help (-h)', 'See all available flags.'],
```
Since these are messages to print, I'll just point out FYI that these are not consistent. Some begin with capital letters, some end with periods. The starting verb tense isn't consistent.  :)

If you wanted consistency, you could choose the "command" tense of each (Specify, Listen, Expose, See) or the "descriptive" tense of what the flag does (Specifies, Listens, Exposes, Displays/Prints/Lists) but See is the tricky one to replace here, and it's probably not horrible if that one just stays as "See". But, ideally the top 3 should at least be consistent, to me.

Up to you! Just noting!

jasikpark: Would be pretty cool to generate the Docs descriptions from the help commands, though those are definitely different contexts

ematipico:
I had to disable the test. It became very flaky and unreliable. Most of the time doesn't pass, few times it passes. And I have no idea why.

It's a pity that when a test fails, the logs are not shown at all :(

packages/astro/src/cli/check/index.ts
```diff
@@ -18,11 +19,25 @@ interface Result {
 	hints: number;
 }
 
-export async function check(settings: AstroSettings, { logging }: { logging: LogOptions }) {
+export async function check(
+	settings: AstroSettings,
+	{ logging, flags }: { logging: LogOptions; flags: Arguments }
+) {
+	if (flags.help) {
```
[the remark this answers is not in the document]
jasikpark: i’m confused, is adding “-h” as an option not necessary?

ematipico:
> Can "-h" in addition to "--help" support be added? looks great otherwise 😄

Sorry, what do you mean by support? 🤔 

If you meant something like `astro build -h`, then yes, this PR will support it!

packages/astro/src/core/dev/dev.ts
```diff
@@ -32,7 +34,26 @@ export interface DevServer {
 export default async function dev(
 	settings: AstroSettings,
 	options: DevOptions
-): Promise<DevServer> {
+): Promise<DevServer | undefined> {
+	if (options.flags?.help || options.flags?.h) {
```
jasikpark, line 38: don't you need to duplicate this logic elsewhere?

most of the if statements don't check for `-h`

ematipico: I see what you mean. I had applied the suggestions, but then they were removed because I force pushed. Thank you for catching that, I'll fix the code ASAP!

jasikpark (approved):
thx for building this! lgtm 👍

PR #6396, opened 2023-03-01
Remove unused benchmarks

bluwy:
## Changes

Removed `/scripts/memory/` and `/packages/astro/test/benchmark/` benchmarks. Also remove it's related command and documentation.

They are now superseded by the new `/benchmark/` folder, and `pnpm run benchmark` command in root.

## Testing

n/a

## Docs

I've updated CONTRIBUTING.md


ematipico (approved):
The best PRs are the ones that have red bigger than green! 🚀

natemoo-re (approved):

CONTRIBUTING.md
```diff
@@ -124,19 +124,16 @@ pnpm exec changeset
...
 
 ```shell
-pnpm run benchmark --filter astro
+pnpm run benchmark
 ```
 
-Which will fail if the performance has regressed by **10%** or more.
-
-To update the times cd into the `packages/astro` folder and run the following:
+This will run all available benchmarks sequentially. To run a specific benchmark, you can run (for example):
```
sarah11918, line 130: Just a shout out that it can sometimes be confusing to mix what happens for two different commands on the same line like this. 

Because of the deleted lines, I can't actually suggest here, but I would propose a structure closer to the following so you can scan for actions like "To do X..." or "Use Y" etc at the beginnings of lines:

``````
We have benchmarks to keep performance under control.**

You can run all available benchmarks sequentially by running (from the project root):

```shell
pnpm run benchmark
```

To run a specific benchmark only, you can add the name of the benchmark after the command:

```shell
pnpm run benchmark memory
```

Use `pnpm run benchmark --help` to see all available options.
``````

** (not SUPER required to start a new line here! Feel free to keep these sentences combined! But especially if you wanted to add any sentence with generic benchmark info, or if you still had that note about failure for example, this would be the place to do it. Then start each new line with an action someone might want to do, followed by the command.)

Comment (bluwy) [reply CONTRIBUTING.md]: I like the convention! I've split them up to follow the suggestion. I've also added more info of the `benchmark` directory, and the `!bench` command. Let me know what you think!

**Review by Yan-Thomas — APPROVED** [review]
LGTM, thanks @bluwy!

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -121,22 +121,23 @@ pnpm exec changeset
...
 ```shell
-node test/benchmark/build.bench.js --save
-node test/benchmark/dev.bench.js --save
+pnpm run benchmark memory
 ```
 
-Which will update the build and dev benchmarks.
+Use `pnpm run benchmark --help` to see all available options.
+
+In a PR on GitHub, you can also comment `!bench`, `!bench memory`, etc, to run the benchmarks on CI. The benchmark will run on both the PR branch and the `main` branch, and the results will be posted as a comment.
```
matthewp, line 140: Can you explain the difference between `!bench` and `!bench memory`? Maybe a list of commands like:

- `!bench` - Benchmarks rendering performance against the PR branch and main and prints out a table of each result.
- `!bench memory` - Benchmarks memory usage against the PR branch and main and prints out a table of each result.

bluwy: I was hoping the paragraphs before this already implied what they do, otherwise it would be duplicating the docs 🤔 Ideally we would expand more commands than `!bench` too so they can use `pnpm run benchmark --help` to view the full list of them.

Maybe we can rename the npm script as `pnpm run bench` so it's clearer?

matthewp: Ah ok, so the !bench command maps directly to the npm command? I didn't realize that. If so then I guess you can disregard this comment.

sarah11918, line 140:
```suggestion
To run these benchmarks in a PR on GitHub instead of using the CLI, you can comment `!bench`. The benchmarks will run on both the PR branch and the `main` branch, and the results will be posted as a new comment.

To run only a specific benchmark on CI, add its name after the command in your comment, for example, `!bench memory`.
```
What about something like this? I'm hoping adding something like "these benchmarks" makes it a little more clear that it's all the same benchmarks available!

bluwy: Sounds good to me 👍

PR #6397, opened 2023-03-01
ci: correctly update comment in action

ematipico:
## Changes

This PR updates the workflow `check-merge.yml` to use a different action to add comments: https://github.com/peter-evans/create-or-update-comment

Used this action before and it's really handy and simple to use! With this change, our PRs won't be bloated by messages anymore! :)

## Testing

We have to merge it and see how it behaves, unfortunately. 

## Docs

N/A


matthewp (approved):

benchmark/package.json
```diff
@@ -15,4 +15,4 @@
     "mri": "^1.2.0",
     "port-authority": "^2.0.1"
   }
-}
\ No newline at end of file
+}
```
ematipico, line 18: That's a change coming from `pnpm format`

PR #6401, opened 2023-03-02
feature: pass tsconfig to vite alias

okikio:
## Changes

This changes `vite-plugin-config-alias` so that its aliases are passed directly to Vite.

By passing aliases from `jsconfig.json` or `tsconfig.json` files directly into Vite’s built-in alias configuration, non-JS files like CSS can now aliases.

Meaning in CSS you can now use aliases, e.g.

```scss
@import "src:styles/shared.css";
```

## Testing

The current test for `tsconfig.json` only targets JS resolution, and so I’ve added a test for `jsconfig.json` that also target CSS resolution.

For an example of the issue, here is what current happens when an alias is used inside a CSS file:
https://stackblitz.com/edit/github-khftu4?file=src/pages/_index.css

For an example of the fix, here is what happens if we push the tsconfig to aliases instead:
https://stackblitz.com/edit/github-khftu4-d255tn?file=src/pages/_index.css

## Docs

### DIFF
--- a/.changeset/strong-colts-study.md
+++ b/.changeset/strong-colts-study.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Pass config aliases directly to Vite alias
--- a/packages/astro/src/vite-plugin-config-alias/index.ts
+++ b/packages/astro/src/vite-plugin-config-alias/index.ts
@@ -64,14 +64,6 @@ const getConfigAlias = (settings: AstroSettings): Alias[] | null => {
 		}
 	}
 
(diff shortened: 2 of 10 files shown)


matthewp:
The author if this commit is banned from Astro. Asking someone else to submit a PR on his behalf is an attempt to evade the ban. We cannot accept the PR regardless of the code merits for this reason. Please do not submit code on their behalf in the future.

Jon is banned from the Astro GitHub repositories until such a time that the core team decides to relax the ban. That will happen when/if the team decides to do so. Attempts to evade the ban will only lessen the likelihood of that happening.

PR #6406, opened 2023-03-02
ci: use one action for comments

ematipico:
## Changes

Chore PR.  Follow up of #6397 

It updates the benchmark workflow to use the same action we use for comments.

cc @bluwy 

## Testing

I should be able to trigger the action manually and see the message, although I am unsure if that will work. Alternatively, we would need to merge it and then use the magic comment to trigger the benchmark.

## Docs

N/A

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -91,11 +91,13 @@ jobs:
       pull-requests: write
     steps:
       - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v1
+        uses: peter-evans/create-or-update-comment@v2
+        continue-on-error: true
         with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          pr_number: ${{ github.event.issue.number }}
+          issue-number: ${{ github.event.issue.number }}
           message: |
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -91,11 +91,13 @@ jobs:
...
         with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          pr_number: ${{ github.event.issue.number }}
+          issue-number: ${{ github.event.issue.number }}
           message: |
             ${{ needs.benchmark.outputs.PR-BENCH }}
 
             ${{ needs.benchmark.outputs.MAIN-BENCH }}
+            - name: Send PR review
+          edit-mode: replace
```
bluwy, line 103:
```suggestion
          edit-mode: replace
```

bluwy (approved):
Small typo, but LGTM. I don't think it should be a big deal if we run `!bench` and then `!bench memory` later (which have different results). We can view the edit history in that case.

natemoo-re (approved):

ematipico:
FYI, manually triggering the action didn't have any effect, unfortunately.

PR #6408, opened 2023-03-02
Update MDX VSCode extension recommendation

kanadgupta:
## Changes

The MDX VSCode extension recommended in the MDX example was [deprecated last August](https://github.com/silvenon/vscode-mdx/commit/c68ac445814d472b7120cb706f5f64d8c33bdfe1) in favor of [the official MDX extension](https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx). This PR updates the VSCode recommendation accordingly!

## Testing

No actual tests here, but I made the change in a local example and confirmed that the recommendations showed up properly 🎊 

Before:

_(note that the recommended extensions list deliberately filters out extensions that have been deprecated)_

CleanShot 2023-03-02 at 14 16 09@2x

After:

CleanShot 2023-03-02 at 14 12 38@2x

## Docs

No docs updates needed, just a tiny VSCode config change (but please correct me if I'm wrong!)


natemoo-re (approved):
Thanks for the fix, that's great!

PR #6409, opened 2023-03-02
Update devcontainer names

bdmac:
## Changes
Sets unique name per example app devcontainer for enhanced usability.

## Testing

N/A

## Docs

Docs-only change.


natemoo-re (approved):
Great change, thank you so much! ♥️

PR #6410, opened 2023-03-02
Fix marketing typo by just removing extra info

bdmac:
## Changes
Noticed there was a typo ported over from the original devcontainer with "Codespaces" written "CodeSpaces". I was going to just correct the casing but realized the extra context is probably unnecessary now so instead I just opted to remove it instead.

## Testing
N/A

## Docs
N/A


natemoo-re (approved):
Makes sense, thanks!

PR #6412, opened 2023-03-03
Remove redundant comments when `astro add` update `astro.config.mjs`

liruifengv:
## Changes

- `astro add svelte` will update `astro.config.mjs`, but will generate redundant comments `/ https://astro.build/config`.

image

## Testing
None

## Docs
None


natemoo-re (approved):
Looks good to me, folks will be very happy about this one!

PR #6413, opened 2023-03-03
Use undici's FormData for polyfilling

Princesseuh:
## Changes

Previously we used a random polyfill package for FormData, but undici actually provide this so we can use theirs instead.

## Testing

Tests should pass!

## Docs

Added some comments telling us which polyfills to remove in future node versions


Yan-Thomas (approved):
Docs LGTM!

natemoo-re (approved):
Code looks good! CI needs to be rerun, but once tests are passing this should be good to go.

PR #6415, opened 2023-03-03
Add page render benchmark

bluwy:
## Changes

Add new `pnpm run benchmark render` CLI command, and `!bench render` comment command.

Question: Is creating a private `@astrojs/timer` integration in `/packages/integrations/` fine? Does it affect docs generation?

Example output locally:

| Page   | Avg (ms) | Stdev (ms) | Max (ms) |
| :----- | -------: | ---------: | -------: |
| /astro |     0.58 |       0.92 |     9.52 |
| /md    |     2.75 |       0.31 |     4.21 |
| /mdx   |    63.81 |      14.06 |   113.71 |

## Testing

Tested manually locally, and through https://github.com/bluwy/astro/pull/1

## Docs

I've added a new private `@astrojs/timer` integration that has a README. Ideally this should be kept private from users too.

### DIFF
--- a/packages/integrations/timer/src/server.ts
+++ b/packages/integrations/timer/src/server.ts
@@ -1,0 +1,20 @@
+import { polyfill } from '@astrojs/webapi';
+import type { SSRManifest } from 'astro';
+import { NodeApp } from 'astro/app/node';
+
+polyfill(globalThis, {
+	exclude: 'window document',
+});
+
+export function createExports(manifest: SSRManifest) {
+	const app = new NodeApp(manifest);
+	return {
+		handler: async (req, res) => {
+			const start = performance.now();
+			await app.render(req);
+			const end = performance.now();
+			res.write(end - start + '');
+			res.end();
+		},
+	};
+}
--- a/benchmark/bench/render.js
+++ b/benchmark/bench/render.js
@@ -1,0 +1,117 @@
+import fs from 'fs/promises';
+import http from 'http';
+import path from 'path';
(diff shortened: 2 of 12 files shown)


.github/workflows/benchmark.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         continue-on-error: true
         with:
           issue-number: ${{ github.event.issue.number }}
-          message: |
+          body: |
```
bluwy, line 98: This was a typo that prevented from CI commenting

packages/integrations/timer/src/server.ts
```diff
@@ -0,0 +1,21 @@
...
+export function createExports(manifest: SSRManifest) {
+	const app = new NodeApp(manifest);
+	return {
+		handler: async (req: IncomingMessage, res: ServerResponse) => {
+			const start = performance.now();
+			await app.render(req);
+			const end = performance.now();
+			res.write(end - start + '');
+			res.end();
+		},
```
bluwy, line 19: The main code in the `@astrojs/timer` integration is here. The rest are largely copied from `@astrojs/node` but stripped out the unneeded parts.

ematipico (approved):
Looks good to me. One thing I struggled with: it took me a while to understand **how** the render benchmark works. Perhaps, there should be a description somewhere (even the README) to explain how we run the benchmarks.

bluwy:
Oh crap I forgot there's a requested review for docs on this. Let me check with them real quick 😅

PR #6429, opened 2023-03-06
chore: use directive `@ts-expect-error` instead of `@ts-ignore`

ematipico:
## Changes

This is a maintenance PR. It replaces `@ts-ignore` with `@ts-expect-error`.

The benefit of `@ts-expect-error` is that it tells the compiler that a compiler error is expected in the next line, and **if there isn't any error**, the compiler fails.

This directive is useful in cases the "compiler is not powerful enough", but then a new version comes around and fixes the issue.

The places where `@ts-ignore` was removed were the places where there wasn't a compiler error anymore.

## Testing

The CI should pass, there's not change in logic.

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/add/babel.ts
+++ b/packages/astro/src/core/add/babel.ts
@@ -3,12 +3,12 @@ import parser from '@babel/parser';
 import traverse from '@babel/traverse';
 import * as t from '@babel/types';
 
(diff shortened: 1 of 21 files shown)


bluwy (approved):
Nice cleanup!

PR #6430, opened 2023-03-06
Ignore `@astrojs/timer` in changesets

bluwy:
## Changes

Looks like `@astrojs/timer` (private package for benchmark only) is still handled in changeset's release: https://github.com/withastro/astro/pull/6414. Hopefully setting this ignore will not bump its version and generate changelogs.

## Testing

n/a

## Docs

n/a. tweaks to changesets workflow.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,7 +6,7 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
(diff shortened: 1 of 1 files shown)


natemoo-re (approved):

bluwy:
After discussion with docs, it might be good to move `@astrojs/timer` into `/benchmark` instead and rename it differently.

PR #6433, opened 2023-03-06
Move benchmark package and update changeset config

bluwy:
## Changes

This PR moves `packages/integrations/timer` to `benchmark/packages/timer` so that benchmark packages are all in one place. And also `packages/integrations` are for public integrations only.

This also update the changeset config to prevent detecting `@benchmark/*`, `astro-scripts`, and `astro-benchmark` from releasing. The latter two are private packages and uses `workspace:*` so they don't need a release.

## Testing

I made sure `pnpm run benchmark render` still works as it relies on `@benchmark/timer`.

## Docs

The `@benchmark/timer` package has a README with a short description of it.


matthewp (approved):

delucis (approved):
Docs look ok! Thanks for moving the around @bluwy

PR #6435, opened 2023-03-06
Expose the ssr manifest

matthewp:
## Changes

- Makes the manifest exposed through `astro:ssr-manifest` so that plugins can get its information.
- Requested via the bling project which does its own routing and wants to know what scripts exist.
- Also names the dev server middleware handler.

Can be tried out by installing `astro@next--ssr-manifest`

## Testing

- Test added

## Docs

Undocumented for now.


matthewp:
!preview ssr-manifest

PR #6454, opened 2023-03-08
Add support new `astro:assets` ESM import shape for SVGs 

Princesseuh:
## Changes

We don't support optimizing SVGs at this time, however, supporting the new ESM shape for them is in scope.

## Testing

Added a test for refusing unsupported formats.

## Docs

N/A, the docs actually already wrongly documented that SVGs were supported 😅

### DIFF
--- a/packages/astro/src/assets/vite-plugin-assets.ts
+++ b/packages/astro/src/assets/vite-plugin-assets.ts
@@ -197,7 +197,7 @@ export default function assets({
 				resolvedConfig = viteConfig;
 			},
 			async load(id) {
-				if (/\.(heic|heif|avif|jpeg|jpg|png|tiff|webp|gif)$/.test(id)) {
+				if (/\.(jpeg|jpg|png|tiff|webp|gif|svg)$/.test(id)) {
 					const url = pathToFileURL(id);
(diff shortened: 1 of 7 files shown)


packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -197,7 +197,7 @@ export default function assets({
 				resolvedConfig = viteConfig;
 			},
 			async load(id) {
-				if (/\.(heic|heif|avif|jpeg|jpg|png|tiff|webp|gif)$/.test(id)) {
+				if (/\.(jpeg|jpg|png|tiff|webp|gif|svg)$/.test(id)) {
```
Princesseuh, line 200: For the removed formats here: It turns out that `image-size` does not actually support them! The popular alternative `probe-image-size` does, but it's also multiple times larger... 

It's unlikely that people will import heic and heif, but avif could happen. Maybe we could send a fix upstream? Or maybe we could shake the world with a new `image-size` replacement, who knows.

ematipico (approved):

delucis (approved):
Docs side of things LGTM!

PR #6455, opened 2023-03-08
Fix configuration reference docs

delucis:
## Changes

- Adds documentation for the `image.service` config option
- Adds the `@docs` tag to include the “Experimental Flags” heading in docs

## Testing

n/a docs only

## Docs

/cc @Princesseuh for feedback on how I documented `image.service`

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -721,7 +721,7 @@ export interface AstroUserConfig {
 		 * ```js
 		 * {
 		 *   image: {
-		 *     // Example: Enable the Sharp image processor
+		 *     // Example: Enable the Sharp-based image service
 		 *     service: 'astro/assets/services/sharp',
 		 *   },
 		 * }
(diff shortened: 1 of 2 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -706,6 +706,27 @@ export interface AstroUserConfig {
...
+		 * @description
+		 * Set which image service is used for Astro’s experimental assets support.
+		 *
+		 * The value should be a module specifier for the image service to use:
+		 * either one of Astro’s two built-in services, or a third-party implementation.
+		 *
+		 * ```js
+		 * {
+		 *   image: {
+		 *     // Example: Enable the Sharp image processor
```
Princesseuh, line 724:
```suggestion
		 *     // Example: Enable the Sharp-based image service
```
Maybe? I think there's an important distinction between "image processor" and "image service" since we document image services as a concept:tm:

Princesseuh (approved):
Looks great from my end!

delucis:
Does this require a patch changeset?

Princesseuh:
> Does this require a patch changeset?

Since the docs is user facing (as in, you can see it in your editor), I'd say that it does!

Issue #6443, opened 2023-03-07
fetch not available in endpoints


mef:
### What version of `astro` are you using?

2.0.17

### Are you using an SSR adapter? If so, which one?

Node.js adapter

### What package manager are you using?

npm

### What operating system are you using?

linux

### Describe the Bug

Using `fetch` in an endpoint works well in development mode, but fails when executing the build using the node.js SSR adapter (`@astrojs/node@5.0.4`) with node.js version `v16.17.1`.

`npm run build` runs successfully. However, running either `npm run preview`  or `node dist/server/entry.mjs`  fails with the following error message:

```
file:///home/me/development/project-name/dist/server/chunks/pages/all.35ef890f.mjs:3021
        fetch(thirdPartyURL + '?locale=en', fetchOptions)
        ^

ReferenceError: fetch is not defined
    at getData (file:///home/me/development/project-name/dist/server/chunks/pages/all.35ef890f.mjs:2898:8)
    at file:///home/me/development/project-name/dist/server/chunks/pages/all.35ef890f.mjs:2902:1
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:533:24)
    at async Module.preview (file:///home/me/development/project-name/node_modules/@astrojs/node/dist/preview.js:13:23)
    at async preview (file:///home/me/development/project-name/node_modules/astro/dist/core/preview/index.js:33:18)
    at async runCommand (file:///home/me/development/project-name/node_modules/astro/dist/cli/index.js:170:22)
    at async cli (file:///home/me/development/project-name/node_modules/astro/dist/cli/index.js:180:5)
```

Possible workarounds:

* upgrade to node.js v18, which ships with an implementation of  `fetch` (flagged as experimental, also based on undici ([details](https://nodejs.org/en/blog/announcements/v18-release-announce/))
* import a node.js implementation of `fetch` from the dependency `undici` - which is installed with Astro - inside the endpoint file:

```
import { fetch } from 'undici'
```

Looking at the [tests for global fetch](https://github.com/withastro/astro/blob/main/packages/astro/test/fetch.test.js), it seems that the support is only tested in astro components and framework components, but not in endpoint files.

Is this the expected behaviour? I would have assumed that fetch would also be supported in endpoint files in SSR mode, just as it works in dev mode.

### Link to Minimal Reproducible Example

[...]

PR #6457, opened 2023-03-08
fix(@astrojs/node): inject polyfill when running preview server

ematipico:
## Changes

Partial fix for https://github.com/withastro/astro/issues/6443

## Testing

I manually tested that the preview server works.

## Docs

N/A

### DIFF
--- a/packages/integrations/node/src/preview.ts
+++ b/packages/integrations/node/src/preview.ts
@@ -3,7 +3,11 @@ import http from 'http';
 import { fileURLToPath } from 'url';
 import { createServer } from './http-server.js';
 import type { createExports } from './server';
+import { polyfill } from '@astrojs/webapi';
 
+polyfill(globalThis, {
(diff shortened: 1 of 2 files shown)


packages/integrations/node/src/preview.ts
```diff
@@ -3,7 +3,11 @@ import http from 'http';
 import { fileURLToPath } from 'url';
 import { createServer } from './http-server.js';
 import type { createExports } from './server';
+import { polyfill } from '@astrojs/webapi';
 
+polyfill(globalThis, {
+	exclude: 'window document',
+});
```
matthewp, line 10: if you look at, for ex. the build, this is inside of the exported function, not a the top-level.

Princesseuh:
Closing since the issue isn't in the preview server. Hopefully the changes we're currently making to bundling will also help with this issue!

Issue #6404, opened 2023-03-02
MDX and rehype-mathjax not rendering properly


ewzpack:
### What version of `astro` are you using?

2.0.16

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### Describe the Bug

rehype-mathjax (CHTML) creates html structure and inject CSS, but the problem is that the CSS gets escaped

for example

```css
mjx-container[jax=&quot;CHTML&quot;] {
  line-height: 0;
}
mjx-stretchy-v &gt; mjx-beg {
  height: 0;
}
```

instead of 

```css
mjx-container[jax="CHTML"] {
  line-height: 0;
}
mjx-stretchy-v > mjx-beg {
  height: 0;
}
```

Its works correctly when using markdown only or importing a `.md` file (containing math expression) into a `.mdx` file.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-r1w9pn?file=src/content/blog/math.mdx

### Participation

- [ ] I am willing to submit a pull request for this issue.

y-nk:
definitely related to https://github.com/withastro/astro/issues/6327

bluwy:
Yeah this seems like the same issue as #6327. Keeping it separate for now to make sure we fix both script and style tags.

wulinsheng123:
@bluwy, could you let me take a look at this bug?
I want to learn something about content modules.
thanks

bluwy:
I've already started working on this, so you might have to pick a different one 😬

Issue #6327, opened 2023-02-22
inject script tag in mdx content


y-nk:
### What version of `astro` are you using?

latest

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

npm@8

### What operating system are you using?

macOS@latest

## Problem

When trying to add a script tag in an mdx file, the script tag isn't parsed properly.

**Without valid JSX syntax**, as expected the content gets parsed as html, quotes are properly respected:

```html
<!-- input -->
<script>
window.alert('foo')
</script>

<!-- output -->
<script><p>window.alert('foo')</p></script>
```

**With　valid JSX syntax**, _surprisingly_ the content of the tag gets escaped:

```html
<!-- input -->
<script>{`window.alert('foo')`}</script>

<!-- output -->
<script>window.alert(&#39;foo&#39;)</script>
```

I also tried `is:raw`, `is:inline` and the `client:*` combinations without success.

## Goal

To inject client only small scripting without having anything else to do than `<script>`. I know that remark-mdx would replace script tags to div, but since it works here... why not push it to the max?

## Workaround

Add a framework integration and rely on them by creating a third party script component such as:

```tsx
export default function Script({ value }: { value: string }) {
  return <script dangerouslySetInnerHTML={{ __html: value }} />;
}
```

...but i wished i didn't have to rely on [React/Vue/etc...] integration to do this.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vuknlr?file=src/pages/index.mdx&on=stackblitz

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Sorry for the frustration! I'm curious if `set:html` works? That's our equivalent to `dangerouslySetInnerHTML`

y-nk:
@natemoo-re after testing, the `set:html` works but then it'll still be a plain string, so it's not that different from the workaround i've proposed (with the React component).

I made a rewrite in Astro with a little enhancement. It feels a bit more like `is:inline` now. The DX is much better _but_ still "not 🐐"-rated and bundling is still not possible so it's pretty limited.

Here's for posterity:

```astro
---
import { readFileSync } from "fs"
import { dirname, join } from "path"

const { src, url } = Astro.props

export type Props = {
  src: string
  url?: string
}

const dir = url ? dirname(new URL(url).pathname) : process.cwd()
const code = readFileSync(join(dir, src), 'utf-8')
---
<script set:html={code} />
```

to be used as:

[...]

PR #6459, opened 2023-03-08
Prevent HTML-escape of raw strings in JSX script/style tags

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/6404
Fixes https://github.com/withastro/astro/issues/6327

This PR allows writing

```jsx
<script>
{`console.log('raw script')`}
</script>

<style>
{`
h1[id="script-style-raw"] {
  color: red;
}
`}
</style>
```

Without the inner string contents to be HTML-escaped incorrectly. This also unblocks remark/rehype plugins that injects AST that would theoretically represent something like the above. 

## Testing

Added tests in MDX integration

## Docs

Technically this allows the syntax of the above, but I'm not sure this is a pattern we want to show in the docs yet.


packages/astro/src/runtime/server/jsx.ts
```diff
@@ -220,11 +220,27 @@ async function renderElement(
...
 	);
 }
 
+/**
+ * Pre-render the children with the given `tag` information
+ */
+function prerenderElementChildren(tag: string, children: any) {
+	// For content within <style> and <script> tags that are plain strings, e.g. injected
+	// by remark/rehype plugins, or if user explicitly does `<script>{'...'}</script>`,
+	// we mark as HTML string to prevent the content being html-escaped.
```
ematipico, line 236:
```suggestion
	// For content within <style> and <script> tags that are plain strings, e.g. injected
	// by remark/rehype plugins, or if a user explicitly does `<script>{'...'}</script>`,
	// we mark it as an HTML string to prevent the content from being HTML-escaped.
```

packages/astro/src/runtime/server/jsx.ts
```diff
@@ -220,11 +220,27 @@ async function renderElement(
...
-				: `>${children == null ? '' : await renderJSX(result, children)}</${tag}>`
+				: `>${
+						children == null ? '' : await renderJSX(result, prerenderElementChildren(tag, children))
+				  }</${tag}>`
 		)}`
 	);
 }
 
+/**
+ * Pre-render the children with the given `tag` information
```
ematipico, line 231: nit nit nit: the description and/or the function's name don't do justice to the actual implementation. Saying "pre-render" is a bit reductive, considering what the function actually does. Maybe `escapeContentTags` or something like that?

ematipico (approved):

matthewp (approved):

packages/astro/src/runtime/server/jsx.ts
bluwy: I chose `prerenderElementChildren` since it's being called by the `renderElement` function, so kinda like a pre-pass phase. So ideally in the future this function could do more than handling this specific case, similar to the other render functions in this file. So I'm not sure if we need to rename it to something specific 🤔

Issue #6444, opened 2023-03-07
Initial scaffolded project with @astrojs/lit erroring: module is not defined


cbarratt:
### What version of `astro` are you using?

^2.0.18

### Are you using an SSR adapter? If so, which one?

Netlify, but happens with no SSR

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### NOTE:

This appears to be ok when using yarn, but not pnpm

If I do `pnpm install --force --shamefully-flatten`, it works

I've just gone to setup a brand new astro project using the astro cli to scaffold it so, i then proceeded to add the lit integration using `pnpm astro add lit`.

Once the lit() integration was added, the site will error:

```
ReferenceError: module is not defined
    at eval (/Users/callumbarratt/projects/checkout-web/node_modules/.pnpm/node-domexception@1.0.0/node_modules/node-domexception/index.js:18:1)
    at instantiateModule (file:///Users/callumbarratt/projects/checkout-web/node_modules/.pnpm/vite@4.1.4/node_modules/vite/dist/node/chunks/dep-ca21228b.js:52420:15)
```

### Link to Minimal Reproducible Example

https://github.com/dankochetov/astro-test-2

### Participation

- [X] I am willing to submit a pull request for this issue.

bluwy:
`lit` requires this `.npmrc` to work with pnpm: https://github.com/withastro/astro/blob/e1858e633484d0e1eca0e3ad72af679fc12271e0/examples/framework-lit/.npmrc#L1

Maybe `astro add` needs to be tweaked to add that if it doesn't exist, for pnpm.

PR #6460, opened 2023-03-08
Add default .npmrc for `astro add lit`

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6444

Lit dependencies need to be hoisted to work at the meantime. This PR surfaces this requirement in a few more places.

Currently it doesn't edit the `.npmrc` if there's an existing one, but I think this would be enough for now.

## Testing

Tested the CLI manually in the examples folder. Confirmed that it works for pnpm only.

## Docs

Updated the lit integration readme to note this workaround in its troubleshooting section.

### DIFF
--- a/.changeset/tough-tigers-drum.md
+++ b/.changeset/tough-tigers-drum.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Add default `.npmrc` file when adding the lit integration through `astro add lit` and using `pnpm`
+Add default `.npmrc` file when adding the Lit integration through `astro add lit` and using `pnpm`.
--- a/.changeset/tough-tigers-drum.md
+++ b/.changeset/tough-tigers-drum.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Add default `.npmrc` file when adding the lit integration through `astro add lit` and using `pnpm`
--- a/packages/astro/src/core/add/index.ts
+++ b/packages/astro/src/core/add/index.ts
@@ -61,7 +61,7 @@ export default {
 };
 `;
 const LIT_NPMRC_STUB = `\
-# Lit libraries are required to be hoisted for now due to dependency issues
+# Lit libraries are required to be hoisted due to dependency issues.
 public-hoist-pattern[]=*lit* 
 `;
 
--- a/packages/astro/src/core/add/index.ts
+++ b/packages/astro/src/core/add/index.ts
@@ -60,6 +60,10 @@ export default {
 	preprocess: vitePreprocess(),
 };
 `;
+const LIT_NPMRC_STUB = `\
+# Lit libraries are required to be hoisted for now due to dependency issues
+public-hoist-pattern[]=*lit* 
+`;
 
 const OFFICIAL_ADAPTER_TO_IMPORT_MAP: Record<string, string> = {
 	netlify: '@astrojs/netlify/functions',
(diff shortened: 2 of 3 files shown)


.changeset/tough-tigers-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Add default `.npmrc` file when adding the lit integration through `astro add lit` and using `pnpm`
```
sarah11918, line 5:
```suggestion
Add default `.npmrc` file when adding the Lit integration through `astro add lit` and using `pnpm`.
```
It looks like we capitalize Lit when we use the name in a sentence!

packages/astro/src/core/add/index.ts
```diff
@@ -60,6 +60,10 @@ export default {
 	preprocess: vitePreprocess(),
 };
 `;
+const LIT_NPMRC_STUB = `\
+# Lit libraries are required to be hoisted for now due to dependency issues
```
sarah11918, line 64: I wasn't sure whether this was a public-facing user message, but just in case:

```suggestion
# Lit libraries are required to be hoisted due to dependency issues.
```
I would tend to take out things like "for now"... gives us cover!  ;) If this is a public-facing statement, then we usually do make them sentences with punctuation.

packages/integrations/lit/README.md
```diff
@@ -141,6 +141,14 @@ export default defineConfig({
 
 The correct order might be different depending on the underlying cause of the problem. This is not guaranteed to fix every issue however, and some libraries cannot be used if you are using the Lit integration because of this.
 
+### Strict package managers
+
+When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may get an error like `ReferenceError: module is not defined` when running your site. To fix this, you can hoist Lit dependencies with an `.npmrc` file:
```
sarah11918, line 146:
```suggestion
When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like `pnpm`, you may get an error such as `ReferenceError: module is not defined` when running your site. To fix this, hoist Lit dependencies with an `.npmrc` file:
```

sarah11918 (commented):
Looks great, @bluwy! Just some tiny nits!  :raised_hands:

packages/integrations/lit/README.md
```diff
@@ -141,6 +141,14 @@ export default defineConfig({
...
+
+```ini title=".npmrc"
+public-hoist-pattern[]=*lit* 
+```
```
ematipico, line 150:
```suggestion
```ini 
title=".npmrc"
public-hoist-pattern[]=*lit* 
```
```

**Review by ematipico — APPROVED** [review]

**Review by matthewp — APPROVED** [review]

File: .changeset/tough-tigers-drum.md (hunk 1 above)
Comment (bluwy) [reply .changeset/tough-tigers-drum.md:5]: Thanks! All the suggestions make sense to me

File: packages/integrations/lit/README.md (hunk 4 above)
Comment (sarah11918) [reply packages/integrations/lit/README.md:150]: Sorry, just seeing the notification now!

This format above works for docs, yes. And it's what we prefer if the code is *only* displayed in docs. 

But, if it messes up the GitHub README, then docs should *also* be able to parse:
````
```lang
// filename.here

code snippet
```
````
So that's a reasonable backup, too!

bluwy: Cool! I think because the paragraph pre-mentioned that it is a `.npmrc` file, it being omitted in GitHub readme would be fine. And the extra addition for rendered docs is a plus, so I'll stick with the change for now.

PR #6465, opened 2023-03-08
Fixes ESM imported assets to be root relative

matthewp:
## Changes

- Fixes ESM imported assets to be root relative instead of using a file URL. This means they can be used with a normal `` tag.

## Testing

- New test added

## Docs

N/A, bug fix


natemoo-re (approved):

matthewp:
!preview image-root-relative

Issue #6471, opened 2023-03-09
Cloudflare Adapter in Directory Mode doesn't respect `base`


RichiCoder1:
### What version of `astro` are you using?

v2.1.2

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

When generating the function and SSR `_routes.json` for Cloudflare, the adapter will not respect the `base` option resulting in basic routes 404-ing.

The applies to both SSR and prerender'd routes (due to an issue both with the generated function, and the `_routes.json` not being mapped to base.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-4ru86g?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
@RichiCoder1 is this fixed with #6473 merged?

RichiCoder1:
@bluwy yup, was fully resolved with https://github.com/withastro/astro/pull/6550

PR #6473, opened 2023-03-09
fix: rebase _routes.json for Cloudflare when config.base set

RichiCoder1:
## Changes

Fixes and issue where `base` was not being respected for the generation of `_routes.json`.

Part of https://github.com/withastro/astro/issues/6471

## Testing

Manual validation

## Docs

Shouldn't need a doc fix as this fixes the _routes to align with expectations.


packages/integrations/cloudflare/src/index.ts
```diff
@@ -150,7 +150,11 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						.map((file: string) => `/${file}`);
 
 					for (let page of pages) {
-						staticPathList.push(prependForwardSlash(page.pathname));
+						let pagePath = prependForwardSlash(page.pathname);
+						if (config.base !== '/') {
+							pagePath = `${config.base}${pagePath}`;
```
RichiCoder1, line 155: This is making an assumption that `base` doesn't have a trailing slash. Should I account for that?

bluwy (approved):

natemoo-re (approved):
Appreciate the patch, thank you!

PR #6474, opened 2023-03-09
add browser to bug reports

ElianCodes:
## Changes

Adds a browser field to issue template. I think this could be useful in trying to reproduce bugs.

## Testing

No testing required

## Docs

No docs required


ElianCodes:
Yeah! I think this could be useful especially when running into rendering issues or hydration bugs.

.github/ISSUE_TEMPLATE/---01-bug-report.yml
```diff
@@ -39,6 +39,13 @@ body:
       placeholder: Mac, Windows, Linux
     validations:
       required: true
+  - type: input
+    id: browser
+    attributes:
+      label: What browser are you using?
+      placeholder: Chrome, Firefox, Safari
+    validations:
+      required: true
```
natemoo-re, line 48: I don't think this needs to be required?

```suggestion
```

ElianCodes: I do think you might be right, but of course there's no stopping someone from inputting something like "N.A." Maybe it's a better idea to include "N.A or none" in the placeholder.  

The reason i'm suggesting the above, is because OS is also required while it might be N.A. And 2. Doesn't this simplify. 

But you guys are the experts, so of course, I'll follow your lead!

natemoo-re: Fair, something to maybe revisit. I'm good with that reasoning!

Issue #6469, opened 2023-03-09
Astro build crashes if any front matter fields are null


NJKode:
### What version of `astro` are you using?

2.1.2

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

MacOS 12.2

### Describe the Bug

I have some blog posts written in Markdown with YAML front matter, e.g:

```markdown
---
title: Cool post!
author:
---
Nice content!
```

I've also set up my Astro schemas to allow null values in my front matter.

However! Since Astro `2.1.0`, if *ANY* of these fields are  nullish (such as the `author` field above) , `astro build` will crash with the following error:

```
Cannot read properties of null (reading '__astro')
```

The problem appears to be in `content/utils.ts`. 
In the `extractFrontmatterAssets()` function there is this block of code:
```javascript
if (typeof curr === 'object') {
  if (curr.__astro === true) {
    // continues
  }
}
```

In the lovingly-crafted world of JavaScript, `null` is of type `object`. 
So my `null` front matter fields  satisfy the first `if` statement. 😎 
However in the very next step, we try to access the `__astro` field on this `null` object - resulting in explosions. 😢 

Seems like a possible fix would be to also check `curr !== null`.

### Link to Minimal Reproducible Example

https://github.com/NJKode/astro-build-error

### Participation

- [X] I am willing to submit a pull request for this issue.

Princesseuh:
JavaScript 🤦‍♀️

Will fix, thank you for the report!

cjreinhardt:
Is this still happening to anyone else? I’m running into this error on a project. The linked repo above also generates an error despite this issue being closed. Will post more details shortly.

Princesseuh:
The code that was causing this error does not exist at all anymore since many versions ago. If you're encountering something similar in a current version, it is most definitely not this issue and we'd love if you could open a new issue!

cjreinhardt:
> The code that was causing this error does not exist at all anymore since many versions ago. If you're encountering something similar in a current version, it is most definitely not this issue and we'd love if you could open a new issue!

Thank you! Will do a little more investigation to make sure it’s not something I’m doing wrong, will open another issue if it seems appropriate.

PR #6480, opened 2023-03-09
fix(images): Fix empty frontmatter keys crashing Astro

Princesseuh:
## Changes

This would've never had happened in Rust

Fix https://github.com/withastro/astro/issues/6469

## Testing

Added one. That fixture gets built, so the build will fail if we don't handle this correctly. We could test in the dev server too presumably, but it's the same codepath either ways...

## Docs

N/A

### DIFF
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -46,7 +46,7 @@ export const msg = {
 export function extractFrontmatterAssets(data: Record<string, any>): string[] {
 	function findAssets(potentialAssets: Record<string, any>): ImageMetadata[] {
 		return Object.values(potentialAssets).reduce((acc, curr) => {
-			if (typeof curr === 'object') {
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):

packages/astro/src/content/utils.ts
```diff
@@ -46,7 +46,7 @@ export const msg = {
 export function extractFrontmatterAssets(data: Record<string, any>): string[] {
 	function findAssets(potentialAssets: Record<string, any>): ImageMetadata[] {
 		return Object.values(potentialAssets).reduce((acc, curr) => {
-			if (typeof curr === 'object') {
+			if (typeof curr === 'object' && !Array.isArray(curr) && curr !== null) {
```
bholmesdev, line 49: Gotta love checking if things are really actually truly objects huh 😓

bholmesdev (approved):
`object`ively perfect

Princesseuh:
Superseded by https://github.com/withastro/astro/pull/6483

PR #6483, opened 2023-03-09
Fix using images in content collections

Princesseuh:
## Changes

Previous to this PR, the images we would return from Content Collections's frontmatters wouldn't be built by Vite, as such it would break in build. Additionally, the path was wrong so trying to use them directly through an `img` tag would fail in both modes. TL:DR: It never worked, ha!

Fix https://github.com/withastro/astro/issues/6469

## Testing

Added a whole bunch of tests

## Docs

N/A


packages/astro/src/content/utils.ts
```diff
@@ -43,21 +42,24 @@ export const msg = {
...
-				if (curr.__astro === true) {
-					acc.push(curr);
+export function patchFrontmatterAssets(data: Record<string, any>): Record<string, any> {
+	function patchAsset(potentialAssets: Record<string, any>): Record<string, any> {
+		Object.keys(potentialAssets).forEach((entry) => {
+			if (typeof potentialAssets[entry] === 'object' && potentialAssets[entry] !== null) {
+				if (potentialAssets[entry].__astro_asset) {
+					potentialAssets[entry] = `$$ASSET_(await import('${normalizePath(
+						potentialAssets[entry].src
+					)}')).default_ASSET$$`;
```
Princesseuh, line 52: This is super ugly, but I couldn't figure out how to do in a better way. Passing and trying to use `this.load()` here failed in various ways

matthewp (approved):

PR #6488, opened 2023-03-09
Add fallback for hosts without import.meta.url

matthewp:
## Changes

- Fixes bug reported in discord: https://discord.com/channels/830184174198718474/1083154052264759398

## Testing

- Just manually. Not sure if there aren't other Netlify bugs underneath this.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/assets/utils/metadata.ts
+++ b/packages/astro/src/assets/utils/metadata.ts
@@ -1,8 +1,8 @@
 import { createRequire } from 'module';
 import fs from 'node:fs/promises';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { ImageMetadata, InputFormat } from '../types.js';
-const require = createRequire(import.meta.url);
+const require = createRequire(getModuleURL(import.meta.url));
 const sizeOf = require('image-size');
 
 export interface Metadata extends ImageMetadata {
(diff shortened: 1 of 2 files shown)


packages/astro/src/assets/utils/metadata.ts
```diff
@@ -1,8 +1,8 @@
 import { createRequire } from 'module';
 import fs from 'node:fs/promises';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { ImageMetadata, InputFormat } from '../types.js';
-const require = createRequire(import.meta.url);
+const require = createRequire(getModuleURL(import.meta.url));
 const sizeOf = require('image-size');
```
Princesseuh, line 6: Would using a normal import fix the problem? If so, it might be worth investigating that. I only used `require` to workaround a weird issue I had going on during dev, but it might not be required. The image integration actually imported the same package without a require

matthewp: Yeah I noticed that. I assumed it was necessary for a reason. I agree, I'll try that and see what happens.

matthewp:
!preview image-image-size

packages/astro/src/assets/utils/metadata.ts
matthewp: Switched to `import`, thanks for the suggestion

natemoo-re (approved):
Changes look good but PR title and changelog seem confusing for the final changes made

PR #6493, opened 2023-03-09
File abstraction

matthewp:
## Changes

- Adds a new abstract, a `File` that is intended to be used any time we are receiving a string that we *think* is a file. It can take files in the form of:
  - File URLs (string and object)
  - Absolute filesystem paths.
  - Root-relative paths
  - Vite's `/@fs` file path prefixed paths.
- And then convert from and to any of the above. 

## Testing

- Testing the class itself in unit tests
- Updated some of the usage that felt safe to swap out, should not break existing tests of course.

## Docs

N/A, internal refactor

### DIFF
--- a/packages/astro/src/core/file/index.ts
+++ b/packages/astro/src/core/file/index.ts
@@ -80,6 +80,13 @@ class File {
 		return '/@fs' + abs;
 	}
 
+	/**
+	 * Converts to an id that is used as the key in Vite's module graph
+	 */
+	toViteID() {
+		return viteID(this.toFileURL());
+	}
+
 	static getPathType(raw: string, root: URL): PathType {
 		if(raw.startsWith('/@fs')) {
 			return 'vite-fs-path';
--- a/packages/astro/src/core/file/index.ts
+++ b/packages/astro/src/core/file/index.ts
@@ -1,0 +1,113 @@
+import type { AstroConfig } from '../../@types/astro';
+import { pathToFileURL, fileURLToPath } from 'node:url';
+import { normalizePath } from 'vite';
+import { prependForwardSlash } from '../path.js';
+import { unwrapId, viteID } from '../util.js';
+
+export type PathType =
+	/**
+	 * file:///Users/name/projects/todos/src/pages/index.astro
+	 */
+	'url' |
+	/**
+	 * /Users/name/projects/todos/src/pages/index.astro
+	 */
+	'absolute' |
+	/**
+	 * /@fs/Users/name/projects/todos/src/pages/index.astro
+	 */
+	'vite-fs-path' |
+	/**
+	 * /src/pages/index.astro
+	 */
+	'root-relative' |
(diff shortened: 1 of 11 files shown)


natemoo-re (commented):
Seems nice, I'm definitely in favor of this. 

Expect it will take a while to find all the places we're using different abstractions at the moment, but once this propagates through the codebase it'll be great!

Princesseuh:
Same comment as Nate, this is awesome!

packages/astro/src/core/file/index.ts
```diff
@@ -0,0 +1,120 @@
+import type { AstroConfig } from '../../@types/astro';
+import { pathToFileURL, fileURLToPath } from 'node:url';
+import { normalizePath } from 'vite';
+import { prependForwardSlash } from '../path.js';
+import { unwrapId, viteID } from '../util.js';
+
+export type PathType =
+	/**
+	 * file:///Users/name/projects/todos/src/pages/index.astro
+	 */
```
Princesseuh, line 10: In addition to those comments, would it be possible to add a brief explanation as to when you should use each path types? This abstraction makes it easier to use the different ones, but I fear that without explanations, we'll still use the wrong ones at times.

Princesseuh (approved):

packages/astro/src/core/file/index.ts
```diff
@@ -0,0 +1,120 @@
...
+			return 'vite-fs-path';
+		}
+		
+		if(raw.startsWith('/')) {
+			const normalRoot = normalizePath(fileURLToPath(root))
+			if(raw.startsWith(normalRoot)) {
+				return 'absolute';
+			} else {
+				return 'root-relative';
+			}
```
ematipico, line 101: This should fall back to `unknown` if we give a `URL` containing actual website URLs.

In fact, passing a website URL will throw an error:
```js
const { fileURLToPath } = require('url');
let url = new URL('https://example.com');
console.log(fileURLToPath(url)); // error
```

packages/astro/test/units/file/file.test.js
```diff
@@ -0,0 +1,96 @@
...
+			let path = 'some-virtual-id';
+			let file = new File(path, root);
+			try {
+				file.toFileURL();
+				expect(false).to.equal(true);
+			} catch(err) {
+				expect(err).to.be.an.instanceOf(Error);
+				expect(err.message.startsWith('Cannot create file URL')).to.equal(true);
+			}
+		});
```
ematipico, line 94: We should test against `URL` that are not file paths, e.g. website URLs

ematipico (changes requested):

ematipico:
@matthewp The only changes I would request are the tests against non-file paths, the rest is just nitpick/bike sheeding

matthewp:
Closing as this is a little bit stale. I do want to get back to trying this out again in the future.

PR #6494, opened 2023-03-09
Consistency improvements to several package descriptions

Yan-Thomas:
## Changes

Changed several package descriptions so they all are following the same rules and display better in the `astro.build` site.

## Testing

I hope changing a description doesn't break packages 😬 

## Docs

Docs, docs, and docs.


natemoo-re (approved):
Thanks for this, looks great!

ematipico (approved):

packages/integrations/alpinejs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@astrojs/alpinejs",
-  "description": "The official Alpine.js integration for Astro.",
+  "description": "Use Alpine within Astro",
```
sarah11918, line 3: "Within" sounds a little formal here. Is there another word that's accurate enough and doesn't sound weird? "with" or "in"... :laughing:

sarah11918 (approved):
Looks good to me, Yan! I do like the more active verb descriptions.  Would be nice to maybe be able to get rid of that one "within" but non-blocking!

packages/integrations/alpinejs/package.json
Yan-Thomas: I think "with" is fine, I used "within" just to be consistent with the other UI framework integration packages, they all follow "Use (React) components within Astro". The "components" part doesn't apply to Alpine because it's more of additional methods and properties to use in HTML.

sarah11918: @Yan-Thomas I am generally happier using "with" unless you really mean to emphasize *inside*. The place I mostly use "within" in docs is to say that something has to be located "within" `src/pages/` (meaning, anywhere inside, including a subdirectory.)  I could also see using it "within a component..." or "within your template..." again, meaning inside.

So if you want to update others to say "with" I think that's great! I also think if people are searching, they're more likely to type "How do I use Alpine with Astro" (or "and Astro") not "within."

PR #6502, opened 2023-03-10
fix: Enforce the usage of type imports when possible

Princesseuh:
## Changes

Being explicit with type imports only has benefits (apart from needing to type 4 more characters, at least), it helps prevent side effects and weird edge cases.

As such, this PR change all the imports that should've been type imports to be type imports and set `importsNotUsedAsValues` to `error`. It should be noted that this option will change in TypeScript 5.0 (for `verbatimModuleSyntax`, which is the same thing but better, really), so we'll need to update this then.

## Testing

CI should pass!

## Docs

N/A


ematipico (approved):
Great work!

bholmesdev (approved):
This is my `type` of PR!

PR #6505, opened 2023-03-10
fix(images): Move emitESMImage util to another file

Princesseuh:
## Changes

`emitESMImage` depends on `rootRelativePath` which depends on `vite`. As such, we need to be careful when importing it because importing it in the wrong context will cause a bunch of weird issues due to Vite trying to load things it shouldn't be loading (such as itself). Since it was in `internal.ts`, which gets imported from other sources, it would get imported sometimes in the wrong contexts.

This was very confusing, but I think this work

## Testing

Tests should still pass! I tested manually that this fix the issue I was running into, as it couldn't be reproduced either in local or in CI

## Docs

N/A


ematipico (approved):
@Princesseuh from your point of view, how do you think we can prevent this issue from happening again?

Princesseuh:
> @Princesseuh from your point of view, how do you think we can prevent this issue from happening again?

Unfortunately, I do not know. Generally we avoid this kind of situation by having things in `dev` folders so we know that we can use them safely, but this case is different because it was because the file was honestly getting mixed up in some weird import mess. The error also only happened in local, in certain situations, nether tests or CI replicated it 🤷‍♀️

PR #6508, opened 2023-03-10
[Content collections] Better error formatting

bholmesdev:
## Changes

This fixes a few problems in today's error formatter:
- The collection -> entry path is bolded for visibility
- Frontmatter keys are now always placed at the start of an error message. This fixes inconsistencies where the key name could be at the start of the end of a message.
- Union errors now log "Did not match union" instead of "Invalid input." Will also log "type" and "invalid literal" errors for keys shared by all members of a union.
- Unhandled errors now correctly fall through to Zod's error message. This would log "Invalid Input" for all unhandled messages before.

Example: a union between `{ key: z.literal('tutorial') }` and `{ key: z.literal('blog') }` will raise a single error when `key` does not match:
```md
Did not match union.
> key: Expected `'tutorial' | 'blog'`, received 'foo'
```

### Before / after

Here is a before and after of the error experience when authoring in docs.astro.build. These errors are [based on the union schema configured here](https://github.com/withastro/docs/blob/main/src/content/config.ts#L107-L117).

Example invalid frontmatter:

```yaml
---
type: guide
i18nReady: maybe?
---
```

**Before**

image

**After**

image

## Testing

- Unit test error map

## Docs

- Small change to errors data: formatting and slight message change

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -771,7 +771,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		code: 9001,
 		message: (collection: string, entryId: string, error: ZodError) => {
 			return [
-				`${String(collection)} → ${String(entryId)} frontmatter does not match collection schema.`,
+				`**${String(collection)} → ${String(entryId)}** frontmatter is invalid.`,
 				...error.errors.map((zodError) => zodError.message),
 			].join('\n');
 		},
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -771,7 +771,9 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		code: 9001,
 		message: (collection: string, entryId: string, error: ZodError) => {
 			return [
-				`**${String(collection)} → ${String(entryId)}** frontmatter is invalid.`,
+				`**${String(collection)} → ${String(
+					entryId
+				)}** frontmatter does not match collection schema.`,
 				...error.errors.map((zodError) => zodError.message),
 			].join('\n');
 		},
--- a/.changeset/quick-turtles-decide.md
+++ b/.changeset/quick-turtles-decide.md
@@ -1,0 +1,8 @@
+---
+'astro': patch
+---
+
+Improve content collection error formatting:
+- Bold the collection and entry that failed
+- Consistently list the frontmatter key at the start of every error
+- Rich errors for union types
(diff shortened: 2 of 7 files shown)


jasikpark:
Curious if there's anything to upstream / open source as a general solution once this merges 😁 looking amazing

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -771,7 +771,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		code: 9001,
 		message: (collection: string, entryId: string, error: ZodError) => {
 			return [
-				`${String(collection)} → ${String(entryId)} frontmatter does not match collection schema.`,
+				`**${String(collection)} → ${String(entryId)}** frontmatter is invalid.`,
```
bholmesdev, line 774: Flagging for docs review @sarah11918 👀

bholmesdev:
@jasikpark Already want to bring this to our RSS config handler! Also throws a big ole `Invalid input` for some of the types.

jasikpark (commented):
this is CLEAN 👏 quick glance over looks great to me, though I haven't tested or thought through the logic

jasikpark:
reads similar to https://github.com/causaly/zod-validation-error - maybe there's interesting prior art there for how to generalize this, tho ig it probably looks exactly the same lol

bholmesdev:
@jasikpark I gave that library a try! Didn't like that keys were listed at the _end_ of an error instead of the beginning (i.e. `Required at "key"` instead of `"key": Required`). Also doesn't have the union error handling and bolding that we have here. Thanks for mentioning that though, not sure how we could align this with their approach without an explosion of config options 😅

packages/astro/src/core/errors/errors-data.ts
sarah11918: @bholmesdev I love the change overall! And, I know you wanted something shorter here, but I don't *love* "invalid" vs "does not match what you said you wanted here." (which is *why* it's invalid)

If you feel the need to shorten, something like ...

"unexpected frontmatter" feels a little more user friendly?

Your call! Non-blocking!

sarah11918 (approved):
Left a comment! I might try to go for more surprise than judgement, but non-blocking!

matthewp (approved):

jasikpark:
@bholmesdev yep didn't think it'd be a drop-in at all, just wanted to make you aware / show that there's existing interest in the concept of "give me pretty error messages from a zod validation error"

packages/astro/src/core/errors/errors-data.ts
bholmesdev: Understood! Didn't feel too strongly about this change. I think I'll revert to the old wording, with the bolding added 👍

Yan-Thomas (approved):
That's a great improvement, nice work @bholmesdev! Docs LGTM 🙌

Issue #6446, opened 2023-03-07
Vercel rewrites not working


Ehesp:
### What version of `astro` are you using?

2.0.18

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Rewrites inside of a `vercel.json` file are not applied (note, redirects do seem to work).

Also see https://discord.com/channels/830184174198718474/1077242845406167112/1077242845406167112

To reproduce, add this to your `vercel.json` and deploy:

```json
{
  "rewrites": [
    {
      "source": "/",
      "destination": "/page2"
    }
  ]
}
```

See repro link below. The index page should rewrite to page2:

https://astro-vercel-rewrites.vercel.app/

### Link to Minimal Reproducible Example

https://github.com/Ehesp/astro-vercel-rewrites

### Participation

- [ ] I am willing to submit a pull request for this issue.

ivoilic:
This is also true for rewrites in Vercel middleware.
Related discussion: https://github.com/withastro/roadmap/discussions/261

cruzlutor:
Same issue here, but in my case, it fails only with the root. For sub-directories it works well

ivoilic:
@cruzlutor are those sub directories static? Vercel itself handles rewrites for static content. It’s SSR content where Astro needs to handle the rewrite headers.

natemoo-re:
Hey y'all, circling back to this! We've landed middleware support in Astro and edge middleware support for our Vercel adapter, so I'd recommend leaning on those features to accomplish this for now.

After discussing this with the team, we agree that respecting `vercel.json` would be a nice enhancement, but since don't have a rewrite primitive like Next.js does, it would be quite a bit of work to add. Definitely a future enhancement we're considering! But hopefully middleware solves your need for now.

kidonng:
From what I can observe, rewrites for non-existent pages do work, i.e. existing Astro pages takes precedence when they conflict with Vercel config.

firxworx:
I was reporting a different issue and saw just how many open issues relate to Vercel and figured I'd share the only way I figured out to ship a recent (April/May) Astro project for my client was to ditch Vercel. 

I had so many issues with Vercel, middleware, etc -- Really _any_ nontrivial Vercel feature with an Astro deployment. I had an i18n scenario with translated URL's + RTL where the whole nine yards of rewrites, edge functions, etc played a role.

[...]

PR #6517, opened 2023-03-11
Added a caution about rewrites only working for static files

ivoilic:
Currently using rewrite with static files works but server rendered files with either edge or serverless function don't get rewritten. See: https://github.com/withastro/astro/issues/6446

## Changes

Vercel integration doc update

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/integrations/vercel/README.md
+++ b/packages/integrations/vercel/README.md
@@ -175,6 +175,8 @@ You can use Vercel middleware to intercept a request and redirect before sending
     ```
 1. While developing locally, you can run `vercel dev` to run middleware. In production, Vercel will handle this for you.
 
+:::caution[Trying to rewrite?] Currently rewriting a request with middleware only works for static files. :::
+
 ## Troubleshooting
 
 **A few known complex packages (example: [puppeteer](https://github.com/puppeteer/puppeteer)) do not support bundling and therefore will not work properly with this adapter.** By default, Vercel doesn't include npm installed files & packages from your project's `./node_modules` folder. To address this, the `@astrojs/vercel` adapter automatically bundles your final build output using `esbuild`.


sarah11918:
Noting that this PR itself is only a Docs change (to the README) cautioning that something doesn't work. 

But, Docs needs to know whether this is something that should be documented as not working, or whether the team instead has plans to make this work...  so we can't sign off on this yet! :smile:

ivoilic:
@sarah11918 This note makes no judgement as to wether this feature will be added in the future. I personally wasted hours of my time yesterday assuming I had not implemented rewrite correctly before I looked at the code for @astro/vercel and saw there’s nothing handling the rewrite header. This note can always be updated but not including any mention of this feature being missing does a great disservice to your users.

natemoo-re:
Alright, I was hoping this one would be a quick bugfix and we could close this issue as solved, but it does look like we'd need to implement a rewrite primitive on the framework level. See https://github.com/withastro/astro/pull/6539 for my failed experiments. 

It isn't out of the question to add this feature, but I'm not comfortable unilaterally adding it without weighing the tradeoffs against the rest of the work on our plates.

If @sarah11918 is cool with this note, I think that's probably the best way forward for right now. We can circle back to supporting this functionality in the future.

ivoilic:
@natemoo-re I can take a swing at implementing this sometime this week/weekend if I have some free time. I have a decent idea how to handle the middleware rewrites at least. Not sure about the vercel config ones.

tonyski:
@ivoilic Any updates? Looking forward to this feature!

packages/integrations/vercel/README.md
```diff
@@ -175,6 +175,8 @@ You can use Vercel middleware to intercept a request and redirect before sending
     ```
 1. While developing locally, you can run `vercel dev` to run middleware. In production, Vercel will handle this for you.
 
+:::caution[Trying to rewrite?] Currently rewriting a request with middleware only works for static files. :::
```
sarah11918, line 178:
```suggestion
> **Warning**
> Trying to rewrite? Currently rewriting a request with middleware only works for static files.
```
I think because this README will also be displayed on GitHub, we need to use their syntax here, and probably lose the fancy title. Otherwise, Sarah is cool with this note!

sarah11918 (approved):

sarah11918:
Approving this note, with the syntax change!

Issue #6467, opened 2023-03-08
Whole ui package get bundled in astro


QuentinDutot:
### What version of `astro` are you using?

2.1.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### Describe the Bug

In a monorepo with an astro project and a solidjs ui package, all the components get bundled when running astro build.

In bigger projects with more components it results in a ton of useless js loaded.

Steps to reproduce in the given repro:
```
- npm i
- npm run build
- open apps/web/dist/_astro/index.xxx.js
- find both button and textarea - even only button is used on index page
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/astro-hjo4xk?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Looks like this is because for each client components imported, the import path is passed directly as Rollup inputs of the client build. Ideally we need to generate a virtual file that takes in account of the named export so that the client build treeshakes the output. This would need a bit of refactor but it sounds doable and feels like it would've affected many component libraries used today.

PR #6527, opened 2023-03-13
Treeshake exported client components that are not imported

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6467

When a hydrated or `client:only` component is passed to the client build, only build those that are imported. e.g. `import { Foo } from './ManyComponents.jsx'`, the bundled entry should only export `Foo` and not other unused components.

This PR changes our client component collection logic to also track exports names imported. A custom Vite plugin is used to proxy entries to something like `export { Foo } from '/resolved/path/ManyComponents.jsx'`, so that we treeshake the unused components.

## Testing

Added a bundling test to make sure unused components are treeshaken.

## Docs

n/a. Bug fix.

### DIFF
--- a/.changeset/cyan-camels-swim.md
+++ b/.changeset/cyan-camels-swim.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Treeshake exported client components that are not imported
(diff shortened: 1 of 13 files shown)


natemoo-re (commented):
Nice test, this change looks great. Thanks for tackling!

natemoo-re (approved):

Issue #6529, opened 2023-03-13
Cloudflare Adapter unable to resolve `node:fs`


jackcarey:
### What version of `astro` are you using?

2.1.2

### Are you using an SSR adapter? If so, which one?

Cloudflare v6.2.0

### What package manager are you using?

npm

### What operating system are you using?

Windows & CF Workers

### What browser are you using?

Chrome

### Describe the Bug

- When disabling the CF adapter and switching to `output: "static"`, the site builds and can be previewed, except I am obviously  missing the Cloudflare functions I'm developing.
- When running `astro dev` with the CF adapter and `output: "server"` the site builds and loads correctly in my browser. 
- When trying `astro build --verbose`, the build fails both locally and when pushing to CF Pages, with errors about built-in node modules that couldn't be found and warnings about packages being marked with `sideEffects: false`.
- I've tried changing node versions and switching the adapter mode to `advanced`, to no avail.
- I'm unable to reproduce the error in Stackblitz, as it just returns "Upgrade Required" with no errors in the console.
-My repo is private for now, but I can make it public if necessary.

Here's the info I think is needed:

```
// astro.config.mjs
import { defineConfig } from 'astro/config';
import tailwind from "@astrojs/tailwind";
import compress from "astro-compress";
import mdx from "@astrojs/mdx";
import sitemap from "@astrojs/sitemap";
import critters from "astro-critters";
import cloudflare from "@astrojs/cloudflare";

// https://astro.build/config
export default defineConfig({
  site: 'https://astro-site.pages.dev/',
  experimental: { assets: true },
  integrations: [mdx(), tailwind(), sitemap(), critters(), compress()],
  output: "server",
  adapter: cloudflare({ mode: 'directory' }),
  // output: "static",
});
```

```
//package.json
{
	"name": "astro-site",
	"type": "module",
	"version": "0.1.1",
	"scripts": {
		"dev": "astro dev --verbose",
		"start": "astro dev",
		"build": "astro build",
		"build-verbose": "astro build --verbose",
		"preview": "astro preview",
		"preview-wrangler": "wrangler pages dev ./dist",
		"wrangler-deploy": "wrangler",
		"astro": "astro"
	},
	"dependencies": {
		"@astrojs/cloudflare": "^6.2.0",
		"@astrojs/mdx": "^0.17.2",
		"@astrojs/sitemap": "^1.1.0",
		"@astrojs/tailwind": "^3.0.1",
		"astro": "^2.1.2",
[...]

PR #6533, opened 2023-03-13
Add warning when using assets with non-node adapters

Princesseuh:
## Changes

What the title says!

Closes https://github.com/withastro/astro/issues/6529

## Testing

N/A, just a warning

## Docs

### DIFF
--- a/packages/astro/src/assets/vite-plugin-assets.ts
+++ b/packages/astro/src/assets/vite-plugin-assets.ts
@@ -1,3 +1,4 @@
+import { bold } from 'kleur/colors';
 import MagicString from 'magic-string';
 import mime from 'mime';
 import fs from 'node:fs/promises';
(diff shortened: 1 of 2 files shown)


packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -28,6 +29,30 @@ export default function assets({
 
 	globalThis.astroAsset = {};
 
+	const UNSUPPORTED_ADAPTERS = new Set([
```
Princesseuh, line 32: It's okay for the error to be here, as this Vite plugin is only loaded when `experimental.assets` is enabled. 

It's currently unclear if we'll add non-Node support before 3.0 (mostly depends on deps), but if we don't, we'll have to move this warning somewhere else where we have access to the Astro config, maybe in the `astro:assets` virtual module

PR #6534, opened 2023-03-13
Sitemap: support SSR routes

atilafassina:
## Changes

- [x] Adds support for SSR routes
- [x] Changeset added

## Testing

I tested using the examples, replacing in `package.json` the `@astrojs/sitemap` version with `workspace:*` and making sure it's using `output: "server"` in the `astro.config.mjs`.

Optional: adding `const prerender = true` to a few routes.

Run `astro build` and make sure a `sitemap.xml` is created with no duplicated routes and every SSR route is created.

Optional-2: Add a hardcoded `customPages` definition to sitemap `options`.

## Docs

If you wish to use this implementation, I'm happy to update the docs.

### DIFF
--- a/packages/integrations/sitemap/src/index.ts
+++ b/packages/integrations/sitemap/src/index.ts
@@ -94,7 +94,7 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
 							 * remove the initial slash from relative pathname
 							 * because `finalSiteUrl` always has trailing slash
 							 */
-							const path = finalSiteUrl.pathname + r.pathname.substring(1);
+							const path = finalSiteUrl.pathname + r.generate(r.pathname).substring(1);
 							const newUrl = new URL(path, finalSiteUrl).href;
 							urls.push(newUrl);
 						}
--- a/packages/integrations/sitemap/src/index.ts
+++ b/packages/integrations/sitemap/src/index.ts
@@ -5,14 +5,11 @@ import {
 	type LinkItem as LinkItemBase,
 	type SitemapItemLoose,
 } from 'sitemap';
-import { resolve } from 'path';
 import { fileURLToPath } from 'url';
-import fg from 'fast-glob';
 import { ZodError } from 'zod';
 
 import { generateSitemap } from './generate-sitemap.js';
 import { Logger } from './utils/logger.js';
-import { isRoutePrerendered } from './utils/is-route-prerendered.js';
 import { validateOptions } from './validate-options.js';
 
 export type ChangeFreq = `${EnumChangefreq}`;
@@ -61,61 +58,15 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
 
 		hooks: {
 			'astro:config:done': async ({ config: cfg }) => {
-				if (cfg.site) {
-					config = cfg;
-				} else {
-					// eslint-disable-next-line no-console
-					console.warn(
-						'The Sitemap integration requires the `site` astro.config option. Skipping.'
-					);
-					return;
-				}
+				config = cfg;
 			},
 
-			'astro:build:start': async () => {
-				if (config.output !== 'server' || !config.site) {
-					return;
-				}
-
-				const srcPath = fileURLToPath(config.srcDir);
-				const pagesPath = resolve(srcPath, 'pages');
-
-				const pageFiles = await fg(`${pagesPath}/**/*.{astro,ts,js}`);
-
-				const routes = (
-					await Promise.all(
-						pageFiles.map(async (filePath) => {
-							const isPrerendered = await isRoutePrerendered(filePath);
-							const index = filePath.indexOf('pages/') + 6;
-							const routeSegment = filePath
-								.substring(index)
-								.replace(/\.(astro|ts|js)/, '')
-								.replace(/index$/, '');
-
-							/**
-							 * @TODO
-							 * figure out how to run `getStaticPaths` here.
-							 */
-							const isDynamicRoute = routeSegment.endsWith(']');
-							const shouldIndex = !isDynamicRoute && !isPrerendered;
-
-							return shouldIndex ? `${config.site}/${routeSegment}` : undefined;
-						})
-					)
-				).filter((route): route is string => Boolean(route));
-				const opts = validateOptions(config.site, options);
-
-				opts.customPages = opts.customPages
-					? Array.from(new Set([...routes, ...opts.customPages]))
-					: routes;
-				options = opts;
-
-				logger.info(`build is starting + ${JSON.stringify(opts.customPages, null, 2)}`);
-			},
-
-			'astro:build:done': async ({ dir, pages }) => {
+			'astro:build:done': async ({ dir, routes }) => {
 				try {
 					if (!config.site) {
+						logger.warn(
+							'The Sitemap integration requires the `site` astro.config option. Skipping.'
+						);
 						return;
 					}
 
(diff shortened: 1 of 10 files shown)


packages/integrations/sitemap/src/index.ts
```diff
@@ -51,18 +54,71 @@ const OUTFILE = 'sitemap-index.xml';
...
+							 * @TODO
+							 * figure out how to run `getStaticPaths` here.
+							 */
+							const isDynamicRoute = routeSegment.endsWith(']');
+							const shouldIndex = !isDynamicRoute && !isPrerendered;
+
+							return shouldIndex ? `${config.site}/${routeSegment}` : undefined;
+						})
+					)
+				).filter((route): route is string => Boolean(route));
```
natemoo-re, line 105: We should have all of this information already, because Astro has to do the same logic internally. I'd be uncomfortable duplicating this logic here and keeping it in sync with the routing logic in the main `astro` package.

packages/integrations/sitemap/src/index.ts
```diff
@@ -51,18 +54,71 @@ const OUTFILE = 'sitemap-index.xml';
...
+
+				opts.customPages = opts.customPages
+					? Array.from(new Set([...routes, ...opts.customPages]))
+					: routes;
+				options = opts;
+
+				logger.info(`build is starting + ${JSON.stringify(opts.customPages, null, 2)}`);
+			},
 
+			'astro:build:done': async ({ dir, pages }) => {
```
natemoo-re, line 116:
```suggestion
			'astro:build:done': async ({ dir, routes }) => {
```

Check out the `routes` object which has our full routing manifest. This might have the info you need!

packages/integrations/sitemap/src/utils/is-route-prerendered.ts
```diff
@@ -0,0 +1,8 @@
+const REGEX_PRERENDER = /const prerender = true/g;
+import { readFile } from 'fs/promises';
+
+export async function isRoutePrerendered(filePath: string) {
+	const contents = await readFile(filePath, 'utf-8');
+
+	return REGEX_PRERENDER.test(contents);
+}
```
natemoo-re, line 8: Same comment about logic duplication. This info is handled in `astro` and surfaced in the `routes` map.

natemoo-re (commented):
I don't think we need a `build:start` hook with custom logic for this! `build:done` probably has most of what you need already, and if there's anything missing we can work on exposing it.

atilafassina:
> I don't think we need a `build:start` hook with custom logic for this! `build:done` probably has most of what you need already, and if there's anything missing we can work on exposing it.

Thanks for the insights, @natemoo-re . I refactored accordingly.
I must admit I still can't find a way to run `getStaticPaths` or to output the paths for the SSR'ed dynamic routes. Did I missed something perhaps?

delucis:
`getStaticPaths` is not used for SSR sites so that’s maybe the issue? The model is quite different for dynamic routes in SSR. Instead of declaring a static array of supported routes via `getStaticPaths`, a dynamic route is expected to handle _any_ matching parameter, e.g. `src/pages/[lang]/index.astro` will be used to serve any request that matches whether it is `/en`, `/zh` or `/foo`. It is then up to the developer to check `Astro.params.lang` and decide how that changes what is returned (i.e. whether `/foo` should be a 404 etc.)

I’d guess this makes it impossible for us to support these dynamic SSR routes in the sitemap.

atilafassina:
> `getStaticPaths` is not used for SSR sites so that’s maybe the issue? The model is quite different for dynamic routes in SSR. Instead of declaring a static array of supported routes via `getStaticPaths`, a dynamic route is expected to handle _any_ matching parameter, e.g. `src/pages/[lang]/index.astro` will be used to serve any request that matches whether it is `/en`, `/zh` or `/foo`. It is then up to the developer to check `Astro.params.lang` and decide how that changes what is returned (i.e. whether `/foo` should be a 404 etc.)
> 
> I’d guess this makes it impossible for us to support these dynamic SSR routes in the sitemap.

Oh 🤯. Thanks a lot for this clarification @delucis 
I was thinking on determining preemptively which routes should throw `404` and which should render.
But if it happens at render time, you're absolutely right.

The ideal scenario would be (1) for the developer to use `customPages` if they want to include such routes in the sitemap, or (2) create a additional entry to the sitemap where they append an entry to the xml when the page is generated. But  case n.2 should not be covered by this integration, imho.

---

In this case I'll just clean up my code and make CI pass 😊

natemoo-re (commented):
This is looking really good! Happy to help debug CI failures if you're stuck, just let me know.

atilafassina:
> This is looking really good! Happy to help debug CI failures if you're stuck, just let me know.

Thanks, @natemoo-re 

I'm afraid I'll need your input on this one.
I fell into sort of a rabbit hole and here's what I found out:

`route.pathname` and `route.route` don't take `addTrailingSlash` into consideration. So I started using `route.generate(route.pathname)` which seemed to be the right way.

But the generator does not take into consideration `buildFormat`. So when it's "directory" it should add the trailing slash even when `addTrailingSlash: ignore`.

At this point I wanted to stop and ask you: is it desirable to patch the [`getRouteGenerator`](https://github.com/withastro/astro/blob/main/packages/astro/src/core/routing/manifest/generator.ts#L34-L38). But before asking I decided to try it out, the solution which seems to work for me is:

```ts
	/**
	 * If `addTrailingSlash` is 'always'
	 * or `buildFormat` is 'directory' and `addTrailingSlash` is not "never"
	 */
	const shouldAddTrailingSlash =
		addTrailingSlash === 'always' ||
                (buildFormat === 'directory' && addTrailingSlash !== 'never');
	let trailing: '/' | '' = '';
	if (shouldAddTrailingSlash && segments.length) {
		trailing = '/';
	}

```

but this breaks 1 test: [astro-root-pagination-spread](https://github.com/withastro/astro/blob/main/packages/astro/test/astro-pagination-root-spread.test.js).

Additionally, [`deserializeRouteData`](https://github.com/withastro/astro/blob/main/packages/astro/src/core/routing/manifest/serialization.ts#L24) doesn't have build format available there, so my patch into the generator had to add "directory" as a default value for testing my theory, so I'm not sure if that's a reliable solution.


---

So, what do you think I should do?
Of course, alternatively I can simply add my check on the integration level and tests will pass.

I just would like to have context if `routeData` not using build format for generating the path is intentional.

matthewp:
@atilafassina sorry for not getting back to you. routeData is not "normalized" to take into account trailing slash config, you are right on that. We do that at the "end" of things, such as when static files are generated. So I think you are right to do it in the integration itself.

Let me know when this is ready for another review.

atilafassina:
> @atilafassina sorry for not getting back to you. routeData is not "normalized" to take into account trailing slash config, you are right on that. We do that at the "end" of things, such as when static files are generated. So I think you are right to do it in the integration itself.
> 
> Let me know when this is ready for another review.

Hi, @matthewp 

thanks for the reply.
The PR should be ready for review. I don't exactly understand now why the build is failing.
I fixed the conflicts, but build is failing at:
```
Error: @astrojs/markdown-remark:build: src/frontmatter-injection.ts(1,15): error TS2305: Module '"vfile"' has no exported member 'Data'.
```

and lint is failing at:
```
[error] packages/astro/src/core/errors/errors-data.ts: SyntaxError: ',' expected. (919:12)
[error]   917 | 		code: 99999,
[error]   918 | 	},
[error] > 919 | } as const satisfies Record<string, ErrorData>;
[error]       |            ^
[error]   920 |
[error]   921 | type ValueOf<T> = T[keyof T];
[error]   922 | export type AstroErrorCodes = ValueOf<{
```

neither of those files are touched in this PR. Do you have any advice to offer me? I'm a bit confused. 😅

atilafassina:
pinging both @ElianCodes and @matthewp since both of y'all showed interest in it.
Thanks a lot, folks. Tests are now 🟢. 

I had some weird situation with corrupted `pnpm-lock.yaml` from one of the rebases I did and that was causing unrelated tasks to fail (tests or build in other packages).

ElianCodes:
Awesome! Thanks @atilafassina! I will take a look at it when I'm back from Copenhagen, Denmark!

matthewp (approved):

matthewp:
@atilafassina Saw the new commits, is this ready to go?

atilafassina:
> @atilafassina Saw the new commits, is this ready to go?

Yes. I'd say so.
Just updated and everything looks green now! :)

atilafassina:
> This looks awesome @atilafassina! Before this is merged, are there docs we need to adjust as well?

@ElianCodes 
I don't think so. I couldn't find any mention of the plugin _not_ supporting SSR routes at the [current docs](https://docs.astro.build/en/guides/integrations-guide/sitemap/#usage). So I think no change is needed.

PR #6537, opened 2023-03-13
Prevent astro:content from depending on Node builtins

matthewp:
## Changes

- Fixes issue where non-Node runtimes were breaking when using astro:content due to the Node dependency that `--experimental-assets` has.

## Testing

- In the content collections test, add a plugin that verifies that no Node builtins are depended on in the build.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/test/content-collections.test.js
+++ b/packages/astro/test/content-collections.test.js
@@ -222,6 +222,24 @@ describe('Content Collections', () => {
 				root: './fixtures/content-ssr-integration/',
 				output: 'server',
 				adapter: testAdapter(),
+				vite: {
+					plugins: [
+						// Verifies that `astro:content` does not have a hard dependency on Node builtins.
+						// This is to verify it will run on Cloudflare and Deno
+						{
+							name: 'verify-no-node-stuff',
+							generateBundle() {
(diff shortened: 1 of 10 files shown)


packages/astro/test/content-collections.test.js
```diff
@@ -222,6 +222,24 @@ describe('Content Collections', () => {
...
+						{
+							name: 'verify-no-node-stuff',
+							generateBundle() {
+								const nodeModules = ['node:fs', 'node:url', 'node:worker_threads', 'node:path'];
+								nodeModules.forEach(name => {
+									const mod = this.getModuleInfo(name);
+									if(mod) {
+										throw new Error(`Node builtins snuck in: ${name}`)
+									}
+								});
```
natemoo-re, line 238: This is very smart! I wonder if this would be useful as a more general (not CC-specific) test as well?

natemoo-re (approved):

packages/astro/test/content-collections.test.js
bholmesdev: Totally. I remember multiple breaking bugs with the Markdown plugin a while back too

bholmesdev (approved):
runtime funtime 🙌

PR #6542, opened 2023-03-14
ci: format add owner judgment

btea:
## Changes
Update format.yml


bluwy (approved):
Makes sense to me. I've definitely seen this error before when forking.

btea:
Yes, I also encountered this error when forking.

PR #6543, opened 2023-03-14
fix: fix compiler issues url

liruifengv:
## Changes

- `https://astro.build/issues/compiler` return a 404 page.
- fix compiler issues url

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -24,16 +24,16 @@ export const AstroErrorData = defineErrors({
 	 * @message
 	 * Unknown compiler error.
 	 * @see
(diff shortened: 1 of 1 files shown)


bluwy:
We used to have that vanity link added in https://github.com/withastro/astro.build/pull/506, looks like it was incorrectly reverted through some changes. The right fix would be to re-add that redirect at https://github.com/withastro/astro.build

liruifengv:
> We used to have that vanity link added in [withastro/astro.build#506](https://github.com/withastro/astro.build/pull/506), looks like it was incorrectly reverted through some changes. The right fix would be to re-add that redirect at https://github.com/withastro/astro.build↳

I will do it. 

But shouldn't `compiler issues list` and `create an issue` be two different links？

PR #6544, opened 2023-03-14
fix #6420

### LINKED ISSUE #6420: Node and Rollup assetFileNames output to wrong path

2.0.17

Node

npm

macOS

### Describe the Bug

Ran `npm run build`, Astro generate CSS appear in `server` which is incorrect, the correct should be in `client` folder.
Note: path's `dir` variable returns is an empty string in astro.config.mjs, so by default `assetFileNames` is pointed to `server` folder:

```
import path from "path";

// https://astro.build/config
import node from "@astrojs/node";

// https://astro.build/config
export default defineConfig({
  vite: {
    build: {
      cssCodeSplit: false,
      assetsInlineLimit: 0,
      rollupOptions: {
        output: {
          entryFileNames: 'assets/script/a.[hash].js',
          assetFileNames: ({
            name
          }) => {
            const { ext, dir, base } = path.parse(name);
            if (ext == ".css") return path.join(dir, "assets/css", 'a.[hash].css');
            return "assets/img/[name].[ext]";
          }
        }
      }
    }
  },
  output: "server",
  adapter: node({
    mode: "standalone"
  })
});

```

NA

### Participation

wulinsheng123:
## Changes

fix #6420
It can't find this file cause the node throws an error if the users custom a path that includes the folder path


wulinsheng123:
image
I try a different version of the nodejs. But there is nothing on my computer.

ematipico (approved):
We should update the template under the "Testing" section

PR #6547, opened 2023-03-14
Fix local images having the wrong width and height when both dimensions were provided

Princesseuh:
## Changes

What the title says. This is a regression from when I fixed images having the wrong width and height if neither were provided, kinda funny

## Testing

Added tests for all situations

## Docs

N/A


delucis (approved):
Looks great! Thanks @Princesseuh

bluwy (approved):

PR #6548, opened 2023-03-14
Change how squoosh is loaded in the image integration

matthewp:
## Changes

- Moves away from loading the `.wasm` files in the image integration and instead creates base64 encoded versions of them. This ensures that they are bundled with the JS source code.
- Fixes usage in Netlify.
- Long-term I'm not sure if this is the best approach, so not doing the same for the core image integration yet. We should look into ways where we can guarantee better control over files that need to be read at runtime in SSR environments.

## Testing

- https://the-great-matthewp-site.netlify.app/

## Docs

N/A, bug fix


matthewp:
!preview netlify-image

matthewp:
!preview netlify-image

matthewp:
!preview netlify-image

matthewp:
!preview netlify-image

matthewp:
!preview netlify-image

matthewp:
!preview netlify-image

packages/integrations/image/src/vite-plugin-astro-image.ts
```diff
@@ -31,15 +31,30 @@ export function createPlugin(config: AstroConfig, options: Required<IntegrationO
...
 			resolvedConfig = viteConfig;
 		},
-		async resolveId(id) {
+		async resolveId(id, parent) {
 			// The virtual model redirects imports to the ImageService being used
 			// This ensures the module is available in `astro dev` and is included
 			// in the SSR server bundle.
 			if (id === virtualModuleId) {
 				return await this.resolve(options.serviceEntryPoint);
+			} else if(id.endsWith('?inline-bytes')) {
```
matthewp, line 40: unused

packages/integrations/image/src/vendor/squoosh/emscripten-utils.ts
```diff
@@ -10,19 +10,14 @@ export function pathify(path: string): string {
...
-      // The glue code generated by emscripten uses the original
-      // file names of the worker file and the wasm binary.
-      // These will have changed in the bundling process and
-      // we need to inject them here.
-      if (requestPath.endsWith('.wasm')) return pathify(path)
-      if (requestPath.endsWith('.worker.js')) return pathify(workerJS)
-      return requestPath
-    },
+		// @ts-expect-error
+		wasmBinary: bytes,
```
matthewp, line 17: The emscripten wrapper code accepts this property and it's the way to provide raw bytes, but the TS types do not allow it.

packages/integrations/image/package.json
```diff
@@ -38,7 +38,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc && pnpm run postbuild",
     "build:ci": "astro-scripts build \"src/**/*.ts\" && pnpm run postbuild",
     "postbuild": "astro-scripts copy \"src/**/*.wasm\"",
-    "dev": "astro-scripts dev \"src/**/*.ts\"",
+    "dev": "astro-scripts dev --copy-wasm \"src/**/*.ts\"",
```
MoustaphaDev, line 41: Isn't this just to copy [vendored wasm files](https://github.com/withastro/astro/blob/694918a56b01104831296be0c25456135a63c784/scripts/cmd/build.js#L114) of the core assets feature?

matthewp: Yes you're right, this does nothing here. I'll remove it.

MoustaphaDev (approved):
Looks good to me

packages/integrations/image/src/vendor/squoosh/emscripten-utils.ts
bluwy: I think it would be nice to have this comment in the code comment!

bluwy (approved):
LGTM!

PR #6550, opened 2023-03-15
fix(cloudflare): base strip logic

RichiCoder1:
## Changes

Fixes the strip logic for _config.base. Apparently the negative indexing is a feature of the deprecated substr, not substring.

Apologies for not catching this the first time around.

## Testing

Manually tested via `patch-package` locally. Verified behavior is correct now.

## Docs

N/A

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -156,7 +156,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						let pagePath = prependForwardSlash(page.pathname);
 						if (_config.base !== '/') {
 							const base = _config.base.endsWith('/')
-								? _config.base.substring(0, _config.base.length - 1)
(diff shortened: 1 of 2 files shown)


packages/integrations/cloudflare/src/index.ts
```diff
@@ -156,7 +156,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						let pagePath = prependForwardSlash(page.pathname);
 						if (_config.base !== '/') {
 							const base = _config.base.endsWith('/')
-								? _config.base.substring(0, -1)
+								? _config.base.substring(0, _config.base.length - 1)
```
bluwy, line 159: Maybe we can use slice to simplify it?
```suggestion
								? _config.base.slice(0, -1)
```

bluwy (approved):

bluwy:
Thanks for following up!

Issue #6540, opened 2023-03-13
[assets] Optimized image uses original file extension


fflaten:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Edge

### Describe the Bug

When using the new `Image` component from the experimental assets-feature, the optimized image is saved using the original file extension.

In the sample below the new image will encoded as webp, but file is named ex. `me.40e6410b_1OoVqG.png` and not `me.40e6410b_1OoVqG.webp` as expected.

```ts
---
import { Image } from 'astro:assets';
import avatarImage from '@assets/me.png';
---
<Image src={avatarImage} width={300} alt="Cartoon image of me" />
```

### Link to Minimal Reproducible Example

Unable due to binary files limitation without subscription

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #6555, opened 2023-03-15
Add validateOptions hook to Image Service API

Princesseuh:
## Changes

Add a `validateOption` hook that can be used to validate / augment the passed options by the user. This is necessary because certain options (namely format) need to be augmented before reaching other hooks / `addStaticImage`.

For instance, for `format`, if we change the file format inside `transform` like we did previously, the final file will have the wrong file extension because the path has already been decided before `transform`. And while it's possible to correct the output path (since `transform` returns the outputted format), then we would have the wrong file path in the HTML file as it would point to the previous file extension. This hook solves that. I tried to find another solution, but this was the easiest one.

Fun-fact: This hook was present in the first version of the RFC, but it was removed during internal discussions as I could not remember the use case, oops.

Fix https://github.com/withastro/astro/issues/6540

## Testing

Added a test testing the fix, which uses the hook

## Docs

https://github.com/withastro/docs/pull/2859


bluwy (approved):
Nice! I like how all the validations are in one spot now.

Issue #6511, opened 2023-03-11
Get 500 error with a prerender 404 page and cloudflare SSR


sperjar:
### What version of `astro` are you using?

2.1.0, 2.1.2

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

npm,pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

# Problem: Get this Error When prerender 404 page

```
GET /_astro/index.11c12019.css 500 Internal Server Error (142.91ms)
15:22:51 [ssr] Error: Expected an exported Astro component but received typeof undefined
at Va (/home/projects/github-9dgysh/dist/\_worker.js:60:9174)
at Se.dt (/home/projects/github-9dgysh/dist/\_worker.js:60:15323)
at Se.render (/home/projects/github-9dgysh/dist/\_worker.js:60:14118)
at Object.fetch (/home/projects/github-9dgysh/dist/\_worker.js:92:2761)
at ServiceWorkerGlobalScope.[kDispatchFetch] (/home/projects/github-9dgysh/node_modules/@miniflare/core/src/standards/event.ts:385:13)
```

# My config

1. @astrojs/cloudflare in SSR mode.
2. create a prerendered 404 page.
3. preview (not dev) mode access `/aaa` (or any other path will cause 404 error)

# Reproduction

```
npm run build && npm run dev
```

the Press `b` open a browser, access `/aaa` (or any other path will cause 404 error).

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-9dgysh?file=README.md

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Does this happen when deployed or only when running `astro preview`?

sperjar:
@natemoo-re, deployment and preview mode both cause the problem when using Cloudflare Integration.

bluwy:
I haven't dig completely into it, but the error seems to come from

https://github.com/withastro/astro/blob/04dddd783da3235aa9ed523d2856adf86b792b5f/packages/astro/src/core/app/index.ts#L131-L134

where when hitting a 404, `mod` is a noop `() => {}` instead of the `404.astro` file being rendered. I'm not sure why yet but this seems like a bug.

This also looks similar to https://github.com/withastro/astro/issues/6177#issuecomment-1422779509

PR #6558, opened 2023-03-15
Fix prerendered 404 page handling in SSR

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6511

When falling back to `404.astro`, and that `404.astro` is prerendered, skip rendering it in SSR to prevent runtime errors. Instead, respond with a code 404 so that the environment can fallback and use the prerendered `404.html`.

## Testing

I've added a test to capture the first part. But I've not confirmed if environments _do_ fallback to the `404.html` if we return a 404 status code.

But in a way, if a person opt to prerendering `404.html`, it probably means that the deployed environment supports falling back to it.

## Docs

Perhaps we can document which environments support falling back to `404.html`. It might be good to get the community to fill in this information as I find it rarely documented (especially paired with functions/workers), so it needs some trial and error.

### DIFF
--- a/.changeset/yellow-grapes-design.md
+++ b/.changeset/yellow-grapes-design.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix prerendered 404 page handling in SSR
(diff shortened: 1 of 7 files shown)


natemoo-re (approved):
LGTM!

PR #6559, opened 2023-03-15
Vendor image-size

Princesseuh:
## Changes

`image-size` has caused us a lot of migraine because it's CJS. For some reasons, Vite does not handle it correctly and you get a bunch of `require is not defined` and other errors depending on your package manager, type of install, phase of the moon etc. This vendors it and removes the CJS part of it so everything works.

Vendoring `queue`, a dep of `image-size` was also needed because it ran into the same issues.

## Testing

Tests still pass + tested manually!

## Docs

N/A


Yan-Thomas (commented):
Removed the maintainers-docs review since the only "docs change" is the vendor README that isn't user-facing (and LGTM!) 🙌

matthewp (approved):
Needs a changeset. Otherwise looks good, thanks for remembering the license.

PR #6562, opened 2023-03-15
Integration readme: Replace tabs with double spaces

Tc-001:
## Changes

In multiple `README`s there was a mix of spaces and tabs used, this caused weird and incosistent formatting like seen here in npm:
image

Docs site seems to convert tabs, so was not affected. 

This PR replaces all `/src/packages/**/README.md` tabs with double spaces, making it consistent (as far as I can tell) with the rest of the docs.

## Testing

Docs change, no testing needed

## Docs

This should only affect github and npm, https://docs.astro.build/ should remain the same.


sarah11918:
Thanks @Tc-001 !  :tada:

PR #6572, opened 2023-03-16
Fix error in empty markdown files with content

MoustaphaDev:
## Changes

When the markdown page is empty, `rawFrontmatter` is undefined which causes `rawFrontmatter.indexOf` to throw
image-raw-fm

The fix is to return early when the frontmatter is falsy.

## Testing

Check that we get a correct error
## Docs

Bug fix only


natemoo-re (approved):
Thanks for the great fix!

PR #6579, opened 2023-03-17
feat(all): Migrate to TypeScript 5.0

Princesseuh:
## Changes

This migrates the codebase to use TS 5.0 and the new settings it provides.

Unfortunately, we cannot use `node16` `moduleResolution` at this time because it breaks a lot of our deps and required more time than I had to fix. Probably worth exploring, though.

This makes our builds slightly faster, but it's not THAT noticeable either. The diff looks big because esbuild needed to be upgraded and they added support for more archs

## Testing

Tests should pass!

## Docs

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,10 +84,10 @@
     "@changesets/cli": "2.23.0",
     "@octokit/action": "^3.18.1",
     "@types/node": "^18.7.21",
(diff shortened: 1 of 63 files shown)


bluwy (approved):
Looks great. TS 5.0 🚀

ematipico (approved):

PR #6586, opened 2023-03-17
fix: Add missing --watch flag when running astro check --help

solelychloe:
## Changes

**What does this change?**
- The following change adds the missing `--watch` flag for `astro check` when one runs `astro check --help`.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/tender-emus-eat.md
+++ b/.changeset/tender-emus-eat.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 fix: Add missing --watch flag for astro check when running astro check --help
--- a/.changeset/tender-emus-eat.md
+++ b/.changeset/tender-emus-eat.md
@@ -1,5 +1,5 @@
 ---
-'astro': major
+'astro': minor
 ---
 
 fix: Add missing --watch flag for astro check when running astro check --help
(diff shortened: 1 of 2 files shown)


packages/astro/src/cli/check/index.ts
```diff
@@ -85,7 +85,10 @@ export async function check(
 			commandName: 'astro check',
 			usage: '[...flags]',
 			tables: {
-				Flags: [['--help (-h)', 'See all available flags.']],
+				Flags: [
+					['--watch', 'Check input files in watch mode.'],
```
bluwy, line 89:
```suggestion
					['--watch', 'Watch Astro files for changes and re-run checks.'],
```

I wonder if this would better explain the behaviour `--watch` enables 🤔

.changeset/tender-emus-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
bluwy, line 2:
```suggestion
'astro': patch
```

This should be a patch as it fixes a missing flag description.

solelychloe: Thank you!

packages/astro/src/cli/check/index.ts
solelychloe: You're right. I loosely followed `tsc --watch`'s description but I think this change better explains the `astro check --watch` behavior.

PR #6588, opened 2023-03-17
feat: Expose `$entry` variable to Markdoc

bholmesdev:
## Changes

- Expose content collection entry information to Markdoc files. This allows you to render the parsed frontmatter based on your schema, alongside the generated `id`, `slug`, and `collection` name.

```mdx
---
title: Hello Markdoc!
---

# {% $entry.data.title %}
```

- Introduce a new `getRenderModule()` helper on our internal `addContentEntryType` API. This exposes the parsed `entry` to your Vite loader for use while generating the module (ex. to expose an `$entry` variable to the Markdoc transformer). This also standardizes how the result of `.render()` is generated. Before, integration authors would need to wire up a separate Vite plugin that resolves modules of your preferred extension and hope this meshes with Astro's content collection API ([see diff here](https://github.com/withastro/astro/compare/feat/markdoc-entry-data?expand=1#diff-6f1a5db26f5ed6e5118cf55b197f2368253e83450989fcd71540ebfe56b5eb38L43-L73)). By standardizing this, we can change how and when rendered modules are loaded in the future without breaking changes.
- Introduce a caching layer for entry info, so this information can be passed to `getRenderModule()`.

## Testing

- Add Markdoc integration test for using the `$entry` variable

## Docs

- README section on `$entry`

### DIFF
--- a/packages/astro/src/content/vite-plugin-content-imports.ts
+++ b/packages/astro/src/content/vite-plugin-content-imports.ts
@@ -129,21 +129,25 @@ export const _internal = {
 	}
 
 	// Used by the `render-module` plugin to avoid double-parsing your schema
-	const contentEntryModuleByIdCache = new Map<string, ContentEntryModule | 'loading'>();
-	const awaitingCacheById = new Map<string, ((val: ContentEntryModule) => void)[]>();
+	const contentEntryModuleByIdCache = new Map<
+		string,
+		ContentEntryModule | AwaitingCacheResultQueue
+	>();
+	type AwaitingCacheResultQueue = {
+		awaitingQueue: ((val: ContentEntryModule) => void)[];
+	};
+
 	function getContentEntryModuleFromCache(id: string) {
-		const value = contentEntryModuleByIdCache.get(id);
+		const cacheEntry = contentEntryModuleByIdCache.get(id);
 		// It's possible for Vite to load modules that depend on this cache
 		// before the cache is populated. In that case, we queue a promise
 		// to be resolved by `setContentEntryModuleCache`.
-		if (value === 'loading') {
+		if (typeof cacheEntry === 'object' && cacheEntry != null && 'awaitingQueue' in cacheEntry) {
 			return new Promise<ContentEntryModule>((resolve, reject) => {
-				const awaiting = awaitingCacheById.get(id) ?? [];
-				awaiting.push(resolve);
-				awaitingCacheById.set(id, awaiting);
+				cacheEntry.awaitingQueue.push(resolve);
 			});
-		} else if (value) {
-			return Promise.resolve(value);
+		} else if (cacheEntry) {
+			return Promise.resolve(cacheEntry);
 		}
 		return Promise.resolve(undefined);
 	}
@@ -155,7 +159,7 @@ export const _internal = {
 		fileId: string;
 		pluginContext: PluginContext;
 	}): Promise<ContentEntryModule> {
-		contentEntryModuleByIdCache.set(fileId, 'loading');
+		contentEntryModuleByIdCache.set(fileId, { awaitingQueue: [] });
 
 		const contentConfig = await getContentConfigFromGlobal();
 		const rawContents = await fs.promises.readFile(fileId, 'utf-8');
@@ -205,14 +209,14 @@ export const _internal = {
 			body,
 			_internal,
 		};
-		contentEntryModuleByIdCache.set(fileId, contentEntryModule);
-		const awaiting = awaitingCacheById.get(fileId);
-		if (awaiting) {
-			for (const resolve of awaiting) {
+
+		const cacheEntry = contentEntryModuleByIdCache.get(fileId);
+		if (typeof cacheEntry === 'object' && cacheEntry != null && 'awaitingQueue' in cacheEntry) {
+			for (const resolve of cacheEntry.awaitingQueue) {
 				resolve(contentEntryModule);
 			}
-			awaitingCacheById.delete(fileId);
 		}
+		contentEntryModuleByIdCache.set(fileId, contentEntryModule);
 		return contentEntryModule;
 	}
 
--- a/packages/astro/src/content/vite-plugin-content-imports.ts
+++ b/packages/astro/src/content/vite-plugin-content-imports.ts
@@ -285,7 +285,8 @@ async function getContentConfigFromGlobal() {
 				if (ctx.status === 'loaded') {
 					resolve(ctx.config);
 					unsubscribe();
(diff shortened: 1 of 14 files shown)


bluwy (approved):
LGTM! For a second I thought it's a wordplay on Sentry 🙈

Yan-Thomas (approved):
Good job @bholmesdev, docs LGTM

packages/integrations/markdoc/README.md
```diff
@@ -237,6 +237,20 @@ const { Content } = await entry.render();
...
+
+```md
+---
+title: Welcome to Markdoc 👋
+---
+
+# {% $entry.data.title %}
+```
+
+The `$entry` object matches [the `CollectionEntry`type](https://docs.astro.build/en/reference/api-reference/#collection-entry-type), excluding the `.render()` property.
```
Yan-Thomas, line 252: oops, forgot to send this small suggestion!
```suggestion
The `$entry` object matches [the `CollectionEntry` type](https://docs.astro.build/en/reference/api-reference/#collection-entry-type), excluding the `.render()` property.
```

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -41,116 +45,201 @@ export function astroContentImportPlugin({
...
+				awaiting.push(resolve);
+				awaitingCacheById.set(id, awaiting);
+			});
+		} else if (value) {
+			return Promise.resolve(value);
+		}
+		return Promise.resolve(undefined);
+	}
+
+	async function setContentEntryModuleCache({
```
ematipico, line 152: Could you document this function? A lot is going on, and I struggle to understand the purpose.

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -41,116 +45,201 @@ export function astroContentImportPlugin({
...
+		fileId,
+		pluginContext,
+	}: {
+		fileId: string;
+		pluginContext: PluginContext;
+	}): Promise<ContentEntryModule> {
+		contentEntryModuleByIdCache.set(fileId, 'loading');
```
ematipico, line 159: I admit that having a "loading" value inside a cache is not something you see often. Usually, you either have or do not have value. Would you mind explaining how this caching works?

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -41,116 +45,201 @@ export function astroContentImportPlugin({
...
+			contents: rawContents,
+		});
+		const generatedInfo = getEntryInfo({
+			entry: pathToFileURL(fileId),
+			contentDir: contentPaths.contentDir,
+		});
+		if (generatedInfo instanceof NoCollectionError) throw generatedInfo;
+
+		const _internal = { filePath: fileId, rawData: info.rawData };
+		// TODO: move slug calculation to the start of the build
```
ematipico, line 213: Is there an issue for tracking this TODO?

ematipico (commented):
I admit that I'm struggling to understand the function `setContentEntryModuleCache`. It does a lot of stuff. Of course, I don't have much knowledge of the tool yet, so I lack a lot of contexts.

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -41,116 +45,201 @@ export function astroContentImportPlugin({
...
-				// Escape before Rollup internal transform.
-				// Base on MUCH trial-and-error, inspired by MDX integration 2-step transform.
-				return { code: escapeViteEnvReferences(code) };
+	];
+
+	if (settings.contentEntryTypes.some((t) => t.getRenderModule)) {
+		plugins.push({
+			name: 'astro:content-render-imports',
+			async load(viteId) {
+				if (!contentEntryExts.some((ext) => viteId.endsWith(ext))) return;
```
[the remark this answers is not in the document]
bholmesdev: Deserves a comment! We're ensuring this module should be handled by content collections at all. If not, short circuit to avoid processing the `fileId`.

packages/astro/src/content/vite-plugin-content-imports.ts
bholmesdev: Yep, sorry for the unconventional implementation! Trying to track 3 possible states here:
1. Module has not been cached yet (`undefined`)
2. Module is being resolved, but the async operation hasn't completed yet (`'loading'`)
3. Module has been resolved and is stored (`ContentEntryModule`)

This gets around an annoying problem I hit in production builds, where Vite tries to load a content entry's rendered content _before_ its frontmatter module has been resolved. By tracking which cached entries haven't resolved yet with `'loading'`, we can queue up a promise to "wait" for that frontmatter module to be resolved before rendering the post's content module. I tried implementing this with cached promises instead, but truthfully couldn't wrap my head around the complexity for a friday evening 😅 

If this is too confusing, I'm happy to refactor. Let me know if there's an implementation I'm not seeing

packages/astro/src/content/vite-plugin-content-imports.ts
bholmesdev: Nope, but there should be!

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -41,116 +45,201 @@ export function astroContentImportPlugin({
...
+			// Wait for config to load
+			contentConfig = await new Promise((resolve) => {
+				const unsubscribe = globalContentConfigObserver.subscribe((ctx) => {
+					if (ctx.status === 'loaded') {
+						resolve(ctx.config);
+						unsubscribe();
+					} else if (ctx.status === 'error') {
+						resolve(undefined);
+						unsubscribe();
+					}
```
ematipico: Here we have an `if` and `else if`, which means that the two conditions might hit `false`. Is it intentional?

bholmesdev: Yes, this is intentional! Just means we'll keep waiting until one of these two conditions are met. We can change the `else if` to just an `if`

bholmesdev:
@ematipico Alright, rewrote the cache implementation and [added a thorough code comment](https://github.com/withastro/astro/pull/6588/files#diff-1db1c6ced41edd0c121f7794c3a7b2f67edcdbcb1778371752a83f0e4cff23bbR149). It's not _quite_ the suggestion you made (missed that while I was working) but I think this puts us in a better spot. Opening for re-review if you want to take a look! We can refine from here in a follow-up PR as well.

ematipico (approved):
The implementation is way better now! Thank you for taking the time for refactoring it!

PR #6591, opened 2023-03-18
Added prop type definition for FormattedDate in Blog example

matthewmercuri:
## Changes
Added a type definition for `FormattedDate` in the Blog template. This will ensure users of the template have their `date` prop correctly typed.

## Testing
No new functionality added.

## Docs
I don't believe docs should need to be updated. This does not affect behaviour in any way.


bluwy (approved):
Looks good 👍

PR #6592, opened 2023-03-19
Add option to disable streaming

pilcrowOnPaper:
## Changes

- Add `server.streaming` config (default: `true`)

## Testing

No testing done. 

An option to disable streaming already exists internally so this PR just exposes it.

## Docs

- Update config reference: https://docs.astro.build/en/reference/configuration-reference/

### DIFF
--- a/.changeset/gold-dolphins-burn.md
+++ b/.changeset/gold-dolphins-burn.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Add `server.streaming` config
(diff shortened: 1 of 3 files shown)


pilcrowOnPaper:
For me, it's a styling issue. Streaming causes the background color of the document/html to flash even if no promise is awaited during rendering. My site doesn't benefit at all from streaming so disabling it was the easiest solution. I was told this shouldn't be happening on Discord and maybe that it's a bug.

matthewp:
@pilcrowOnPaper could be. Can you provide an example of this? It doesn't make sense to me. The CSS should be defined in the head, so they should block before any of the page is visible. You should see your background and then the content. Would love to see an example of this, maybe there's something else wrong.

matthewp:
Hey @pilcrowOnPaper, let's take this to a Roadmap discussion. If we wind up wanting to do this we can use this PR to implement.

One consideration here is that middleware is coming soon, and you could use that to disable streaming (buffer the entire page before sending the response). So this could be achieved without a new separate API.

PR #6594, opened 2023-03-19
chore: wrap the projectDir in quotes if it contains spaces

btea:
## Changes
When outputting projectDir in the terminal, if it contains spaces, wrap it in quotes.

image

### DIFF
--- a/.changeset/tall-beans-own.md
+++ b/.changeset/tall-beans-own.md
@@ -1,0 +1,5 @@
+---
+'create-astro': minor
+---
+
+wrap projecDir in quptes if it contains spaces
--- a/.changeset/tall-beans-own.md
+++ b/.changeset/tall-beans-own.md
@@ -1,5 +1,5 @@
 ---
-'create-astro': minor
+'create-astro': patch
 ---
 
 wrap projecDir in quptes if it contains spaces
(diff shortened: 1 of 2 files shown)


packages/create-astro/src/messages.ts
```diff
@@ -121,6 +121,9 @@ export const nextSteps = async ({ projectDir, devCmd }: { projectDir: string; de
 
 	await sleep(100);
 	if (projectDir !== '') {
+		if (projectDir.includes(' ')) {
+			projectDir = `"${projectDir}"`;
+		}
 		const enter = [
 			`\n${prefix}Enter your project directory using`,
 			color.cyan(`cd ./${projectDir}`, ''),
```
andersk, line 129: This should use a real shell-quoting function like ``quote(["cd", `./${projectDir}`])`` from the [shell-quote](https://www.npmjs.com/package/shell-quote) library, rather than a one-off hack for just one of many shell metacharacters that might be misinterpreted.

.changeset/tall-beans-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': minor
```
bluwy, line 2:
```suggestion
'create-astro': patch
```

This should be a patch as it fixes a bug.

ematipico (changes requested):
Could you please fill the template around testing? If you tested manually, perhaps you could attach a screenshot/video

btea:
@ematipico I added a screenshot https://github.com/withastro/astro/pull/6594#issue-1630903156

bluwy (approved):

ematipico (approved):

Issue #6598, opened 2023-03-19
Image asset in Markdown crashes `astro dev` with ENOENT error


andersk:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

With experimental assets enabled, if add an image in `src/assets` and reference it from `src/pages/index.md`:

```md
![image](../assets/astro-icon-dark.png)
```

the dev server crashes when reloading the page:

```console
~/projects/github-s4wjpw 4s
❯ npm start
$ astro dev
  🚀  astro  v2.1.3 started in 79ms
  
  ┃ Local    http://localhost:3000/
  ┃ Network  use --host to expose
  
04:02:59 PM [serve]    404                                  /_image
[ENOENT: no such file or directory, open '/home/projects/github-s4wjpw/.file:///home/projects/github-s4wjpw/src/assets/astro-icon-dark.png'] {
  code: 'ENOENT',
  errno: -2,
  path: '/home/projects/github-s4wjpw/.file:///home/projects/github-s4wjpw/src/assets/astro-icon-dark.png',
  syscall: 'open'
}

~/projects/github-s4wjpw 8s
❯
```

This seems to be due to this `'.' +` added in #6466. As far as I can tell, that commit broke exactly what it purported to fix, so I must be missing something.

https://github.com/withastro/astro/blob/43daac7a9b4cc793369bbeb3f722f8de9d8f64c8/packages/astro/src/assets/vite-plugin-assets.ts#L105

This may be the same issue described in https://github.com/withastro/astro/issues/6462#issuecomment-1460737911, but the main issue of that report was different so I’m opening a new one.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-s4wjpw?file=src%2Fpages%2Findex.md

### Participation

- [ ] I am willing to submit a pull request for this issue.

mfrachet:
I've tried something very naive locally and it seems to work:

```js
const isAbsolutlyReferenced = filePath.startsWith('/');
const filePathURL = isAbsolutlyReferenced
	? new URL('.' + filePath, settings.config.root)
	: new URL(filePath, settings.config.root);
```

at https://github.com/withastro/astro/pull/6466

I can open a PR if necessary, but I'm pretty sure there's a clever way to fix this 😅

Princesseuh:
[...]

Issue #6564, opened 2023-03-16
Image asset referenced from one Markdown file in a collection is redundantly optimized many times


andersk:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

[This project](https://stackblitz.com/edit/github-w4ctjf-kmbr3n) has a collection with 10 empty Markdown files and one Markdown file referencing a single image asset (using experimental assets). If I run `astro build`, it optimizes this single image 38 times:

```
❯ astro build
12:10:56 AM [content] Types generated 2.35s
12:10:56 AM [build] output target: static
12:10:56 AM [build] Collecting build info...
12:10:56 AM [build] Completed in 2.41s.
12:10:57 AM [build] Building static entrypoints...
12:11:02 AM [build] Completed in 5.34s.

 generating static routes 
▶ src/pages/index.astro
  └─ /index.html (+74ms)
▶ src/pages/blog/[slug].astro
  ├─ /blog/empty-0/index.html (+50ms)
  ├─ /blog/empty-1/index.html (+77ms)
  ├─ /blog/empty-2/index.html (+105ms)
  ├─ /blog/empty-3/index.html (+133ms)
  ├─ /blog/empty-4/index.html (+165ms)
  ├─ /blog/empty-5/index.html (+190ms)
  ├─ /blog/empty-6/index.html (+217ms)
  ├─ /blog/empty-7/index.html (+248ms)
  ├─ /blog/empty-8/index.html (+273ms)
  ├─ /blog/empty-9/index.html (+293ms)
  └─ /blog/with-image/index.html (+317ms)

 generating optimized images 
  ▶ /_astro/astro-logo-dark.af92bf69_Zp0Xz8.png (before: 10kb, after: 3kb) (+8.74s)
  ▶ /_astro/astro-logo-dark.af92bf69_Zp0Xz8.png (before: 10kb, after: 3kb) (+212ms)
  ▶ /_astro/astro-logo-dark.af92bf69_Zp0Xz8.png (before: 10kb, after: 3kb) (+196ms)
  ▶ /_astro/astro-logo-dark.af92bf69_Zp0Xz8.png (before: 10kb, after: 3kb) (+247ms)
  ▶ /_astro/astro-logo-dark.af92bf69_Zp0Xz8.png (before: 10kb, after: 3kb) (+173ms)
[...]

Issue #6462, opened 2023-03-08
assets: relative paths in markdown files work in dev, but not in SSG build


mashehu:
### What version of `astro` are you using?

2.1.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

Hi,

I am trying to use the new assets feature and I have markdown files (which I fetch from a different local repo using `Astro.glob`) with relative image paths there. These work fine during dev mode, but when I try to make a static build, I get
```
Error [RollupError]: Failed to parse Markdown file "/Users/matho180/MultiQC/docs/core/development/plots.md":
The "fileName" or "name" properties of emitted files must be strings that are neither absolute nor relative paths, received "../MultiQC/docs/images/cbrewer_scales.png".
    at error (file:///Users/matho180/MultiQC_website/node_modules/rollup/dist/es/shared/node-entry.js:2105:30)
    at FileEmitter.emitFile (file:///Users/matho180/MultiQC_website/node_modules/rollup/dist/es/shared/node-entry.js:23804:24)
    at Object.resolveImage (file:///Users/matho180/MultiQC_website/node_modules/astro/dist/vite-plugin-markdown/index.js:55:23)
    at async file:///Users/matho180/MultiQC_website/node_modules/@astrojs/markdown-remark/dist/remark-collect-images.js:16:20 {
  code: 'VALIDATION_ERROR'
}
```
The relative paths in this error messages are already resolved from `../../images` and correct from the perspective of the website repo.

I linked the respective repo branch below.

### Link to Minimal Reproducible Example

https://github.com/mashehu/MultiQC_website/tree/switch-to-assets

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
I'm finding a dev bug, but not the build bug you are reporting. What is different here? https://stackblitz.com/edit/github-mnzpmy-p1csgv?file=astro.config.mjs,src%2Fpages%2Findex.md&on=stackblitz

mashehu:
found a way to reproduce the error. works in dev, but not on build: https://stackblitz.com/edit/github-mnzpmy-gckpwj?file=astro/src/pages/[slug].astro

mashehu:
okay, 2.1.2 made it also stop working in dev mode 😞

matthewp:
Thanks for the repro! Will look into it tomorrow. This is an experimental feature. Thanks for your patience.

PR #6573, opened 2023-03-17
Fix processing same image multiple times

ken0x0a:
## Changes

- add a check before generating image with asset service to prevent generating same image multiple time

Fix #6564, close #6489

## Testing

I didn't test only this change.
I'm using astro with few more patches and just extracted only this part, but this changes are very simple and don't even need to verify, I think.

## Docs

No changes for usage

PR #6604, opened 2023-03-20
Fix using optimized images in Markdown

Princesseuh:
## Changes

Previously, we had a bit of a wonky dance to make optimized images in Markdown. We'd load the image service a second time outside of Vite, pass it to Remark along a lot of other stuff, resolve the images in Remark, augment the `img` in Rehype and do some other stuff.. Only for it to not work anyway. It had numerous issues, there was a lot of duplicated logic and the final URL was anyway wrong, it pointed to the original file instead of the optimized one.

This PR completely changes this for the following workflow:

In Remark/Rehype:
- A Remark plugin extracts all the `` calls to find the images to ESM imports later
- A Rehype plugin adds a temporary attribute to the generated `` so we can find them later

In Vite:
- In Vite (`vite-plugin-markdown.ts`), we then add an ESM import for every image we found in our Remark plugin
- And we replace every temporary attribute added by the rehype plugin with the result of a call to `getImage()`

Overall, this leads us to a usage that's very similar to how you would actually use `getImage` yourself in Markdown. Additionally, there's no duplicated code, it all goes through the normal pipeline without needing any additional calls to the image services

Supersede https://github.com/withastro/astro/pull/6596
Supersede https://github.com/withastro/astro/pull/6573
Fix https://github.com/withastro/astro/issues/6598 
Fix https://github.com/withastro/astro/issues/6564
Fix https://github.com/withastro/astro/issues/6462

However, https://github.com/withastro/astro/issues/6489 is still an issue. It's unrelated to Markdown.

## Testing

Some tests already covered part of this (notably the `getHTMLAttributes` attributes being added). I added some more tests to the parts that weren't tested, such as the SSG build and using a custom service in Markdown

## Docs

N/A. Feature is already documented


packages/astro/src/vite-plugin-markdown/index.ts
```diff
@@ -143,17 +109,26 @@ export default function markdown({ settings, logging }: AstroPluginOptions): Plu
...
 
-				const images = {
-					${imagePaths.map((entry) => `'${entry}': await import('${entry}')`)}
+				function updateImageReferences(html) {
+					return html.replaceAll(
+						/__ASTRO_IMAGE_=\"(.+)\"/gm,
+						(full, imagePath) =>
+							\`src="\${images[imagePath].src}" \${Object.entries(images[imagePath].attributes)
+								.map((entry) => \`\${entry[0]}="\${entry[1]}"\`)
+								.join(" ")}\`
```
natemoo-re, line 127: This could use our internal [`spreadAttributes`](https://github.com/withastro/astro/blob/28373452503bc6ca88221ffd39a5590e015e4d71/packages/astro/src/runtime/server/index.ts#LL68C17-L68C33) function if you want to clean this up! Handles escaping quotes and sanitization.

natemoo-re (approved):
Clean refactory, looks great!

PR #6607, opened 2023-03-20
[Markdoc] Refactor Renderer internals to use `renderComponent()`

bholmesdev:
## Changes

- Update the Markdoc render internals to remove an unneeded dependency: `stringify-attributes`. Should also be more readable than the JSX implementation, using the same underlying methods as our compiler!

## Testing

N/A

## Docs

N/A


natemoo-re (approved):
This is a nice refactor—much more readable!

PR #6617, opened 2023-03-21
Fix tsconfig alias regression

### LINKED ISSUE #6613: Internal server error: Failed to resolve import "lib/getClient" from...

2.1.4

None

pnpm

Mac

Chrome

### Describe the Bug

I'm using astro with TypeScript. I set `compilerOptions.baseUrl` and `import some from 'lib/thing'`. 2.1.3 works fine, but 2.1.4 broken below error

```
Internal server error: Failed to resolve import "lib/getClient" from...
```

thanks

https://stackblitz.com/edit/github-u39yd1?file=package.json

### Participation

MoustaphaDev:
## Changes

Fix #6613

## Testing

Check that Typescript file without `.ts` extension can be imported

## Docs

N/A, bug fix only

### DIFF
--- a/packages/astro/src/vite-plugin-config-alias/index.ts
+++ b/packages/astro/src/vite-plugin-config-alias/index.ts
@@ -73,18 +73,17 @@ export default function configAliasVitePlugin({
 			}
 		},
 		async resolveId(id, importer, options) {
-			if (id.startsWith('.') || id.startsWith('/')) return;
+			if (id.startsWith('.') || path.isAbsolute(id)) return;
 
 			// Handle baseUrl mapping for non-relative and non-root imports.
 			// Since TypeScript only applies `baseUrl` autocompletions for files that exist
 			// in the filesystem only, we can use this heuristic to skip resolve if needed.
 			const resolved = path.posix.join(resolvedBaseUrl, id);
 
-			const resolvedAliasedId = await this.resolve(resolved, importer, {
+			return await this.resolve(resolved, importer, {
 				skipSelf: true,
 				...options,
 			});
-			if (resolvedAliasedId) return resolvedAliasedId;
 		},
 	};
 }
--- a/packages/astro/src/vite-plugin-config-alias/index.ts
+++ b/packages/astro/src/vite-plugin-config-alias/index.ts
@@ -72,16 +72,19 @@ export default function configAliasVitePlugin({
 				};
 			}
 		},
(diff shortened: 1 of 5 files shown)


packages/astro/src/vite-plugin-config-alias/index.ts
```diff
@@ -72,16 +72,19 @@ export default function configAliasVitePlugin({
...
 		},
-		resolveId(id) {
+		async resolveId(id, importer, options) {
 			if (id.startsWith('.') || id.startsWith('/')) return;
 
 			// Handle baseUrl mapping for non-relative and non-root imports.
 			// Since TypeScript only applies `baseUrl` autocompletions for files that exist
 			// in the filesystem only, we can use this heuristic to skip resolve if needed.
 			const resolved = path.posix.join(resolvedBaseUrl, id);
-			if (fs.existsSync(resolved)) {
```
MoustaphaDev, line 82: When importing TypeScript files without the `.ts` extension, this condition is always false, as `/path/to/ts-flie` doesn't exist.
I thought about checking for TypeScript files but, I realized that there could be other file types that could be resolved when not specifying the extension, maybe via vite plugins pushed before the ts resolver?

ematipico: Great call!

packages/astro/test/alias-tsconfig.test.js
```diff
@@ -44,10 +44,16 @@ describe('Aliases with tsconfig.json', () => {
 
 		it('works in css @import', async () => {
 			const html = await fixture.fetch('/').then((res) => res.text());
-			console.log(html);
 			// imported css should be bundled
 			expect(html).to.include('#style-red');
 			expect(html).to.include('#style-blue');
 		});
+
+		it('can load load typescript files without .ts extension', async () => {
```
ematipico, line 52: nit: ".ts or .d.ts files extensions."

ematipico (approved):

MoustaphaDev:
There's a weird path issue occuring on Windows, I'll debug this on my windows machine

bluwy (approved):
Windows passes now! It was because it checked for absolute imports via `.startsWith('/')` which doesn't match on Windows. I've changed that to `path.isAbsolute`.

MoustaphaDev:
Thanks for jumping in @bluwy!

Issue #6620, opened 2023-03-21
can't run astro build when use vite-pwa tailwind and set ts import alias


deemsang:
### What version of `astro` are you using?

2.1.4

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

got this error when run astro build
```
[vite-plugin-pwa:build] Could not load /home/projects/github-nyhn6u/src/astrojs/tailwind/base.css (imported by astro:scripts/page-ssr.js): ENOENT: no such file or directory, open '/home/projects/github-nyhn6u/src/astrojs/tailwind/base.css'
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-nyhn6u?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

wulinsheng123:
Let me dig into this bug.

deemsang:
change prefix path in tsconfig to anything else except `@` to fix the issue. but maybe its still a bug

wulinsheng123:
> change prefix path in tsconfig to anything else except `@` to fix the issue. but maybe its still a bug

Why do you need to code such as ` "@*": ["src/*"]`?
Why not `"@": ["src/*"]`?

deemsang:
`"@": ["src/*"]` cant import like this `import Layout from "@layouts/Layout.astro";`

wulinsheng123:
> `"@": ["src/*"]` cant import like this `import Layout from "@layouts/Layout.astro";`

Do you have any success with the example in another builder tool?

wulinsheng123:
> Looks like this is a regression for a usecase I didn't expect. We need to pass a custom resolver to `vite.resolve.alias` to optional resolve the path if it fails.

Let me try it out?

bluwy:
Fixed in https://github.com/withastro/astro/pull/6633

PR #6626, opened 2023-03-22
Fix tsconfig alias edgecase with "@*" path key

bluwy:
## Changes

Partial revert of https://github.com/withastro/astro/pull/6566. Fixes https://github.com/withastro/astro/issues/6620

This PR reverts the strategy of using Vite aliases to map tsconfig aliases. Instead it hacks Vite's `createResolver` so that it would still work inside CSS files.

Not proud of what I've done, but this works knowing how Vite uses `createResolver` in practice.

If this feels a bit too hacky, I'm happy to fallback to dropping support for working in CSS files, since we only supported it in two patches and it can be considered a regression.

## Testing

Added a test based on the issue's repro

## Docs

n/a. bug fix.


bluwy:
The core tsconfig alias handling will be reverted, so this PR will be on hold for now until I get to the bottom of things.

bluwy:
Closing this for now. Will open a new PR with the fix.

PR #6627, opened 2023-03-22
Resolve images in the frontmatter relative to the current file

Princesseuh:
## Changes

Previously, all images in the frontmatter were relative to `src/assets`, this PR changes it so it's relative to the current file instead. This is a proper resolve, so you can use aliases (including `~/assets/`) and stuff.

This PR does not cover using images in certain cases, such as arrays. This is intentional as this is not supported at the moment in general so I'll tacle it in a dedicated PR

## Testing

Updated tests

## Docs

https://github.com/withastro/astro/pull/6627

### DIFF
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -3,8 +3,8 @@ import matter from 'gray-matter';
 import fsMod from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
-import type { EmitFile } from 'rollup';
-import { normalizePath, type ErrorPayload as ViteErrorPayload, type ViteDevServer } from 'vite';
+import type { EmitFile, PluginContext } from 'rollup';
+import { normalizePath, type ViteDevServer, type ErrorPayload as ViteErrorPayload } from 'vite';
 import { z } from 'zod';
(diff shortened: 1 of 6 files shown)


bholmesdev:
I see this is all WIP, but love this implementation! Pretty simple all things considered 👍

packages/astro/src/content/utils.ts
```diff
@@ -117,6 +118,37 @@ export async function getEntryData(
...
+		 */
+		async function preprocessAssetPaths(object: Record<string, any>) {
+			if (typeof object !== 'object' || object === null) return;
+
+			for (let [schemaName, schema] of Object.entries<any>(object)) {
+				if (schema._def.description === '__image') {
+					object[schemaName] = z.preprocess(
+						async (value: unknown) => {
+							if (!value || typeof value !== 'string') return value;
+							return (await resolver(value))?.id;
```
natemoo-re, line 133: Do we have a test for aliased paths?

Princesseuh: We do! [This file](https://github.com/withastro/astro/blob/8b2af558b481015a93100b1811f4c85959a42ce0/packages/astro/test/fixtures/core-image-ssg/src/content/blog/one.md) has it all: a relative path, a nested path and an aliased path.

matthewp (approved):

PR #6630, opened 2023-03-22
[Markdoc] Support automatic image optimization with `experimental.assets`

bholmesdev:
## Changes

- Process all Markdoc images with the new `Image` component from `astro:assets`
- Expose Vite's `pluginContext` to content entry types
- Handle all aliased and relative paths based on `src/assets/` (may change based on #6627)

## Testing

- Add tests for absolute, relative, and aliased paths

## Docs

Summarized in changeset. The experimental asset docs should cover expected behavior I think!


packages/integrations/markdoc/package.json
```diff
@@ -32,19 +32,23 @@
   },
   "dependencies": {
     "@markdoc/markdoc": "^0.2.2",
+    "astro": "workspace:*",
```
bholmesdev, line 35: Note: a direct dependency on Astro _will_ cause changesets to auto-bump the Markdoc package on core changes. This is because of a direct dependency on the `emitESMImage` utility function. We should discuss if this is acceptable for an experimental integration or not.

Princesseuh: Hmm, I know you gave two options when making this, but would it be perhaps better to pass `emitESMImage` somehow?

bholmesdev: Yeah this bothers me... I can pass down `emitESMImage` via the `getRenderModule()` context on `this`. Just feels a bit dirty

matthewp: If you make it be a peerDependency it won't bump automatically. All of the other integrations are peerDeps. Is there a reason for this one to have a direct?

Princesseuh (approved):
Amazing work. It looks really great!! I wish the Markdown version was this simple!

packages/integrations/markdoc/package.json
bholmesdev: Moved back to a peer dep! It's only pulled in to access an experimental function (`emitESMImage`), so as long as this is tracked, we should be safe to use peer deps alone.

PR #6638, opened 2023-03-23
Avoid implicit head injection when there is a head element in the tree

matthewp:
## Changes

- Uses the compiler to mark components that have the `<head>` element.
- Then uses the Vite module graph to know if a page contains a component with a `<head>` element.
- If it does, we can avoid implicit head injection heuristics. This avoids all accidental body injection bugs entirely.

## Testing

TBD

## Docs

N/A


matthewp:
!preview head-prop

matthewp:
Verified this fixes the docs issue.

packages/astro/src/core/render/dev/metadata.ts
```diff
@@ -0,0 +1,47 @@
...
+function addMetadata(
+	map: SSRResult['componentMetadata'],
+	modInfo: ModuleInfo | null
+) {
+	if (modInfo) {
+		const astro = getAstroMetadata(modInfo);
+		if(astro) {
+			let metadata: SSRComponentMetadata = {
+				containsHead: false,
+				propagation: 'none'
```
natemoo-re, line 36: Big fan of this change!

packages/astro/src/runtime/server/render/common.ts
```diff
@@ -50,52 +49,9 @@ export function stringifyChunk(result: SSRResult, chunk: string | SlotString | R
...
-					}
-
-					// Astro.slots.render() should never render head content.
-					case ScopeFlags.RenderSlot | ScopeFlags.Astro:
-					case ScopeFlags.RenderSlot | ScopeFlags.Astro | ScopeFlags.JSX:
-					case ScopeFlags.RenderSlot | ScopeFlags.Astro | ScopeFlags.JSX | ScopeFlags.HeadBuffer: {
-						return '';
-					}
-				}
-
```
natemoo-re, line 98: 👋🏻 Byeeeee

natemoo-re (approved):
Looks great, much easier to follow!

Issue #6504, opened 2023-03-10
Image assets paths not correctly resolved if using URL base in site config


songololo:
### What version of `astro` are you using?

2.1.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I'm trying the new experimental assets optimisation for images.

During development everything works as intended.

When building the site, everything works as intended when using a base path prefix of `/`.

However, the build breaks if using a base path other than `/`, e.g. if using `test` as a basepath, then it tries to use the following style of path:
`error   ENOENT: no such file or directory, open '/path/to/repo/astro-minimal/dist/test/_astro/ginger_cat.509f05b4.jpeg'`

Note that `test` has been inserted between `dist` and `_astro`.

### Link to Minimal Reproducible Example

https://github.com/songololo/astro-minimal

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Thank you for submitting this issue! I just renamed the issue to clear up any possible confusion with the old `@astrojs/image` integration.

PR #6643, opened 2023-03-24
Fix using optimized images with `base`

Princesseuh:
## Changes

I did not understand how `base` work, which honestly, is not that surprising considering how confused a lot of us are about it 😔

Fix https://github.com/withastro/astro/issues/6504

## Testing

Added a bunch of tests

## Docs

N/A


Princesseuh:
Hmm tests pass in local but fails in CI... Will investigate

packages/astro/src/assets/internal.ts
```diff
@@ -104,7 +106,15 @@ export async function generateImage(
 		clientRoot = buildOpts.settings.config.outDir;
 	}
 
-	const fileData = await fs.promises.readFile(new URL('.' + options.src.src, serverRoot));
+	const fileData = await fs.promises.readFile(
+		new URL(
+			'.' +
+				prependForwardSlash(
+					join(buildOpts.settings.config.build.assets, basename(options.src.src))
```
matthewp, line 113: Could you alias `options.src.src` to some variable before using it? Reading it here I don't know what it points to. Give it some descriptive name.

Also, this needs to use `path.posix.join` or there will be backslashes on Windows.

PR #6649, opened 2023-03-24
fix(images): Improve error handling around the new assets feature

Princesseuh:
## Changes

This adds some more error handling around recently revamped features, notably Markdown (both frontmatter and content) will now error properly when images are not found.

This PR also does a very small logic fix to images in the frontmatter where images would sometimes be resolved from the wrong collection.

## Testing

Added tests

## Docs

cc @withastro/maintainers-docs ! Some new errors to look at in the classic `errors-data`


packages/astro/src/content/utils.ts
```diff
@@ -130,10 +130,12 @@ export async function getEntryData(
...
 						async (value: unknown) => {
 							if (!value || typeof value !== 'string') return value;
-							return (await resolver(value))?.id;
+							return (
+								(await resolver(value))?.id ?? path.join(path.dirname(entry._internal.filePath))
+							);
 						},
 						schema,
-						{ description: undefined }
+						{ description: '__image' }
```
Princesseuh, line 138: Unrelated bug fix. Since we resetted the `description`, the paths would always be relative to the first collection loaded instead of the one currently being rendered

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,72 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+		message: (options: string) =>
+			`Expected \`src\` property to be either an ESM imported image or a string with the path of a remote image. Received \`${options}\`.`,
+		hint: 'This error can often happen because of a wrong path. Make sure the path to your image is correct.',
+	},
+	/**
+	 * @docs
+	 * @see
+	 * - [Assets (Experimental)](https://docs.astro.build/en/guides/assets/)
+	 * @description
+	 * `getImage()` first parameter should be an object with the different properties to apply to your image.
```
Yan-Thomas, line 564: Nit: small change to be consistent with the error above, that says "getImage()'s first parameter"
```suggestion
	 * `getImage()`'s first parameter should be an object with the different properties to apply to your image.
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,72 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * @docs
+	 * @see
+	 * - [Assets (Experimental)](https://docs.astro.build/en/guides/assets/)
+	 * @description
+	 * Astro could not find an image you included in your Markdown content. Usually, this is simply caused by a typo in the path.
+	 *
+	 * Images in Markdown are relative to the current file, as such, to refer to an image that is directly next to the `.md` file, your path should start with `./`
+	 */
+	MarkdownImageNotFound: {
+		title: 'Image not found',
```
Yan-Thomas, line 591: Missing dot
```suggestion
		title: 'Image not found.',
```

Yan-Thomas (commented):
Good job @Princesseuh, added a few suggestions 🙌

Princesseuh:
Investigating a Windows thing... Not too sure of the cause.

ematipico (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,72 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 			)} are supported for optimization.`,
 		hint: "If you do not need optimization, using an `img` tag directly instead of the `Image` component might be what you're looking for.",
 	},
+	/**
+	 * @docs
+	 * @see
+	 * - [Assets (Experimental)](https://docs.astro.build/en/guides/assets/)
+	 * @description
+	 * The `src` property, either on `getImage`'s first parameter or on the Image component needs to be either an image that has been ESM imported or a string.
```
sarah11918, line 538:
```suggestion
	 * An image's `src` property is not valid.  The Image component requires the `src` attribute to be either an image that has been ESM imported or a string. This is also true for the first parameter of `getImage()`.
```
Just wondering whether something like this structure maybe this breaks up all the possible combinations here vs trying to squish it all into one sentence? (Image component/	`getImage` and ESM import/string)

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,72 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+			`Expected getImage() parameter to be an object. Received \`${options}\`.`,
+	},
+	/**
+	 * @docs
+	 * @see
+	 * - [Assets (Experimental)](https://docs.astro.build/en/guides/assets/)
+	 * @description
+	 * Astro could not find an image you included in your Markdown content. Usually, this is simply caused by a typo in the path.
+	 *
+	 * Images in Markdown are relative to the current file, as such, to refer to an image that is directly next to the `.md` file, your path should start with `./`
```
sarah11918, line 588:
```suggestion
	 * Images in Markdown are relative to the current file. To refer to an image that is located in the same folder as the `.md` file, the path should start with `./`
```
Maybe a tiny bit cleaner?

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -530,6 +530,72 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 */
+	MarkdownImageNotFound: {
+		title: 'Image not found.',
+		code: 3028,
+		message: (imagePath: string, fullImagePath: string | undefined) =>
+			`Could not find requested image \`${imagePath}\`${
+				fullImagePath ? ` at \`${fullImagePath}\`.` : '.'
+			}`,
+		hint: 'This is often caused by a typo in the image path. Please make sure the file exists.',
```
sarah11918, line 597:
```suggestion
		hint: 'This is often caused by a typo in the image path. Please make sure the file exists, and is spelled correctly.',
```
I feel like if we're suggesting that the typo is the most common reason, then the file probably DOES actually exist, and we should check the spelling!

sarah11918 (approved):
I am loving all these error message PRs!  :partying_face: 

Thanks, @Princesseuh for being such a champion of these! I made some non-blocking suggestions for you, to use as you see fit!

packages/astro/src/core/errors/errors-data.ts
Princesseuh: Seems fine to me! Admittedly, this is not an error you're supposed to hit often, it's extra cushion if the user is trying / not intentionally doing something weird

Issue #6641, opened 2023-03-24
Astro.clientAddress returns nothing with Netlify SSR Edge adapter in production


christianpetty:
### What version of `astro` are you using?

2.1

### Are you using an SSR adapter? If so, which one?

Netlify Edge

### What package manager are you using?

npm

### What operating system are you using?

Linux Ubuntu 22.04.2 WSL

### What browser are you using?

Firefox

### Describe the Bug

Astro.clientAddress does not return anything in production on Netlify. It was pointed out to me that it might be due to this:

https://answers.netlify.com/t/edge-functions-dont-expose-clients-ip/56623/6

The address is only accessible via the Netlify Context object, not the header.

### Link to Minimal Reproducible Example

https://rad.cm/ip

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #6651, opened 2023-03-24
Use Deno API to set Astro.clientAddress

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6641
- Uses same logic as the Deno adapter.

## Testing

No :(

## Docs

N/A


packages/integrations/netlify/src/netlify-edge-functions.ts
```diff
@@ -6,7 +6,7 @@ const clientAddressSymbol = Symbol.for('astro.clientAddress');
 export function createExports(manifest: SSRManifest) {
 	const app = new App(manifest);
 
-	const handler = async (request: Request): Promise<Response | void> => {
+	const handler = async (request: Request, connInfo?: any): Promise<Response | void> => {
```
bholmesdev, line 9: nit: is there a `handler` type we could import from Netlify here instead of using `any`?

bholmesdev (commented):
Comment on type checking. Edjust a suggestion (oh boy that pun was bad...)

packages/integrations/netlify/src/netlify-edge-functions.ts
matthewp: done, thanks for the suggestion

PR #6653, opened 2023-03-24
[Markdoc] New config format with runtime variable support!

bholmesdev:
## Changes

https://user-images.githubusercontent.com/51384119/227636133-d500abfb-5ae8-4dae-b5b4-5e07d1cced38.mp4

- Move Markdoc configuration options from your `astro.config.mjs` to a new `markdoc.config.mjs`
- Remove the `components` property from the `<Content />` component. Components can now be imported from the `markdoc.config.mjs` and applied to the `render `property directly. No need for magic string mapping anymore 👏 
- Add deps to `@astrojs/markdoc`: esbuild for processing the `markdoc.config`, `kleur` for formatted logging
- Support passing variables to your Markdoc files from the `Content` component. These variables can be passed as props:

```astro
---
import { getEntryBySlug } from 'astro:content';

const entry = await getEntryBySlug('docs', 'why-markdoc');
const { Content } = await entry.render();
---

<Content name={Astro.params.username} />
```

This example `name` variable is available as `$name`:

```md
Hey {% $name %}! Let's learn about Markdoc.
```

## Testing

- Update tests to use new Markdoc config
- Variable passing test

## Docs

- Update core and example READMEs


packages/integrations/markdoc/components/TreeNode.ts
```diff
@@ -47,44 +44,30 @@ export const ComponentNode = createComponent({
...
 			type: 'component',
 			component,
 			props,
 			children,
 		};
-	} else if (isCapitalized(node.name)) {
-		throw new MarkdocError({
-			message: `Unable to render ${JSON.stringify(node.name)}.`,
-			hint: 'Did you add this to the "components" prop on your <Content /> component?',
-		});
```
bholmesdev, line 81: Bye bye wacky error handling 👋

packages/integrations/markdoc/test/content-collections.test.js
```diff
@@ -37,70 +36,20 @@ describe('Markdoc - Content Collections', () => {
...
-				simplePostEntry,
-				withComponentsEntry,
-				withConfigEntry,
+				post1Entry,
+				post2Entry,
+				post3Entry,
 			]);
 		});
-
-		it('renders content - simple', async () => {
```
bholmesdev, line 54: Render tests and fixtures moved to a separate `render.test.js`

packages/integrations/markdoc/README.md
```diff
@@ -99,100 +99,69 @@ const { Content } = await entry.render();
...
-```astro
----
-import { getEntryBySlug } from 'astro:content';
-import Aside from '../components/Aside.astro';
+{% aside type="tip" %}
 
-const entry = await getEntryBySlug('docs', 'why-markdoc');
-const { Content } = await entry.render();
----
+Use tags like this fancy "aside" to add some *flair* to your docs.
```
sarah11918, line 139: You caught Sarah in the right mood. Flair approved! :nail_care:

sarah11918 (approved):
Uh, I had only one review comment... and it's praise! :tada:

(Am I slipping....?) Great job with the docs, Ben! :purple_circle:

packages/integrations/markdoc/README.md
bholmesdev: Thought I could get away with some flair in the example code 😉

packages/integrations/markdoc/src/load-config.ts
```diff
@@ -0,0 +1,101 @@
...
+	};
+}
+
+/**
+ * Forked from Vite's `bundleConfigFile` function
+ * with added handling for `.astro` imports,
+ * and removed unused Deno patches.
+ * @see https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L961
+ */
+async function bundleConfigFile({
```
bluwy, line 42: 👀  I think it would be great if we can prevent using esbuild for the config. IIUC the markdoc config is evaluated in node so it can be validated before rendering: https://github.com/withastro/astro/blob/b75570d34ca0b528874b255b791df58175524b2e/packages/integrations/markdoc/src/index.ts#L56-L61

Could we add a validation call in `Renderer.astro` instead? https://github.com/withastro/astro/blob/b75570d34ca0b528874b255b791df58175524b2e/packages/integrations/markdoc/components/Renderer.astro#L13-L14

That way the config is only loaded as a normal module by Vite.

bholmesdev: @bluwy Understood! My reason against this: **we would be validating your Markdoc is correct at runtime.** This means, if you deploy to Astro SSR, you wouldn't see any validation errors until the first user makes a request 😬 Loading at buildtime ensures all Markdoc is valid _before_ deploying to the runtime. This mirrors the CI setup that Stripe uses for their deploys as well. I agree the esbuild loading isn't ideal, but this is the only solution to get the best of all worlds.

bholmesdev: I'll add a code comment to clarify!

bholmesdev: @bluwy I... still disagree with this, since validating when it's too late isn't really useful. The markdoc config is also at the base of your project outside of `src`, so it should be assumed that the same rules apply as the `astro.config` (no `import.meta.env` access, no aliases, no vite plugins respected).

I know this is a tough line to walk, but I feel pretty strongly about keeping buildtime validation for SSR. It would be worse to push to runtime so all-the-things work, then walk _back_ what you can do in a config if we hit problems later.

bluwy: Since markdoc is still in v0, let's go ahead with this then. I still have concerns with this technique but we can see how it goes 👍

packages/integrations/markdoc/src/load-config.ts
```diff
@@ -0,0 +1,101 @@
+import type { AstroConfig } from 'astro';
+import type { Config as MarkdocConfig } from '@markdoc/markdoc';
+import { build as esbuild } from 'esbuild';
+import { fileURLToPath } from 'node:url';
+import * as fs from 'node:fs';
+
+const SUPPORTED_MARKDOC_CONFIG_FILES = [
+	'markdoc.config.js',
+	'markdoc.config.mjs',
+	'markdoc.config.ts',
```
ematipico, line 10: Maybe `.mts` should be added too

packages/integrations/markdoc/src/load-config.ts
ematipico: I pick the "validate as soon as possible" between the two evils. If there's a way to prevent errors from being reported to the end users, that should be the top priority. Developers should handle possible edge cases coming from esbuild/vite. Sure, it's a burden for the developers and us, but the end users shouldn't pay the cost.

packages/integrations/markdoc/src/index.ts
```diff
@@ -1,48 +1,40 @@
...
 	addContentEntryType: (contentEntryType: ContentEntryType) => void;
 };
 
-export default function markdocIntegration(
-	userMarkdocConfig: ReadonlyMarkdocConfig = {}
-): AstroIntegration {
+export default function markdocIntegration(legacyConfig: any): AstroIntegration {
+	if (legacyConfig) {
+		// eslint-disable-next-line no-console
+		console.log(
```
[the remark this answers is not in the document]
bholmesdev: Since this is an integration, we don't have access to our internal logger unfortunately. This has happened enough times that I'm thinking an `@astrojs/logger` package for integration authors could be useful 🤷

packages/integrations/markdoc/src/default-config.ts
```diff
@@ -0,0 +1,18 @@
...
+		entry: ContentEntryModule;
+	}
+): MarkdocConfig {
+	return {
+		...config,
+		variables: {
+			entry: ctx.entry,
+			...config.variables,
+		},
+		// TODO: heading ID calculation, Shiki syntax highlighting
```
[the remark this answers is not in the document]
bholmesdev: Yes, reflected on my taskboard right now!

packages/integrations/markdoc/package.json
```diff
@@ -32,17 +35,19 @@
   },
   "dependencies": {
     "@markdoc/markdoc": "^0.2.2",
+    "esbuild": "^0.17.12",
     "gray-matter": "^4.0.3",
+    "kleur": "^4.1.5",
```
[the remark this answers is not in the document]
bholmesdev: Added a note to the PR description: Add deps to @astrojs/markdoc: esbuild for processing the markdoc.config, kleur for formatted logging

packages/integrations/markdoc/src/experimental-assets-config.ts
```diff
@@ -0,0 +1,29 @@
+import type { Config as MarkdocConfig } from '@markdoc/markdoc';
+import Markdoc from '@markdoc/markdoc';
+//@ts-expect-error Cannot find module 'astro:assets' or its corresponding type declarations.
+import { Image } from 'astro:assets';
+
+// Separate module to only import `astro:assets` when
+// `experimental.assets` flag is set in a project.
+// TODO: merge with `./default-config.ts` when `experimental.assets` is baselined.
```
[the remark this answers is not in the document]
bholmesdev: Done. Made a blocked task on my board 👍

packages/integrations/markdoc/src/index.ts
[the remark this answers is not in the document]
ematipico: Oh, I see. I thought we could have access to some `astro` API inside the integrations. Thank you for clarifying!

FredKSchott:
I really love this new API. Sorry I couldn't give feedback earlier, but just wanted to confirm:

- Is `markdoc.config.js` used by anyone else? Is the hope that this is a general format/concept that other projects could use as well and eventually could become a standard, or is it really something that will only ever make sense in this form in an Astro project? I have some thoughts on this.
- Are there any bundling concerns with multiple asides all included in the same global config file? Ex: Say I have 3 tags: `A.astro` (with scoped styles)`, `B.jsx` (or more likely `B.astro` containing a JSX `client:load` or `client:only` island), and `C.astro` (contains a hoisted script). Are we building Astro projects properly so that the assets only show up for each component on the pages where that component/tag is used?

bholmesdev:
Thanks @FredKSchott! Discussed offline, but I'll summarize here:
1. Yes, the `markdoc.config.js` is modeled on the simplest case for NextJS' integration (they offer multiple config options). Feeling pretty comfortable following their lead.
2. The renderer is built to only hoist scripts and styles for components actually _used_ on the page. So there should be no danger of global config meaning global hoisting anywhere you render content. I'll make a PR with a test case to confirm!

PR #6659, opened 2023-03-26
feature: configuration for css inlining behavior

lilnasy:
## Changes
**RFC**: [withastro/roadmap#343](https://github.com/withastro/roadmap/discussions/343#discussioncomment-5539608)

### Major points of review:
- adds an experimental configuration option: AstroConfig.experimental.inlineStylesheets :
```ts
| "never"  // all styles necessary are sent in external stylsheets (default, follows current behavior)
```
```ts
| "auto"   // stylesheets smaller than `ViteConfig.build.assetsInlineLimit` (default: 4kb) are inlined
```
```ts
| "always" // all styles necessary are inlined into <style> tags 
```
- css ordering: head rendering was refactored to allow styles to be ordered across `<link>` (external) and `<style>` (inline) tags
- cleanup: there are various places where links-related utility functions are now redundant

## Testing
Screenshot (88)

## Docs
Docs changes will be required regarding the addition of a configuration option.


lilnasy:
@matthewp could I get some feedback on this?

lilnasy:
!preview css-inline

packages/astro/src/@types/astro.ts
```diff
@@ -1011,6 +1011,26 @@ export interface AstroUserConfig {
 		 * }
 		 */
 		assets?: boolean;
+		/**
+		 * @docs
+		 * @name experimental.inlineStylesheets
+		 * @type {'always' | 'auto' | 'never'}
+		 * @default `external`
```
lilnasy, line 1018: external -> never

sarah11918 (commented):
Just some quick docs comments here!  

- I changed some formatting to match the `trailingSlash` existing example, since that seemed to be a relevant comparison. (Things like single quotation marks, bullet points for the config options.) Do adjust if anything didn't properly copy over, or if spacing is off etc.!

- Since this is experimental, I don't think we need additional documentation other than what's here. This content will show up in the configuration reference page as an experimental option, so it will be documented in our complete list of config options. After this PR is merged (or even after it's no longer experimental) we could separately choose to add a note/link to this reference documentation from our Styling/CSS page saying that this option is available. But, I'd consider that non-blocking for this feature.

Yan-Thomas (approved):
Agree with @sarah11918 here (non-blocking), docs LGTM!

packages/astro/src/core/build/internal.ts
```diff
@@ -240,39 +240,54 @@ export function* eachServerPageData(internals: BuildInternals) {
...
+			return -1;
+		} else if (depthB === -1) {
+			return 1;
+		} else {
+			return depthA > depthB ? -1 : 1;
+		}
+	}
+}
+
+export function mergeInlineCss(
```
YoniFeng, line 279: perf nit (premature optimization + I'm not familiar with Astro)

I'd guess that for sufficiently large projects with a non-trivial amount of CSS, the excess copying won't be trivial.
A regular `for` loop and pushing merged/non-inline `StyleSheetAsset`s to an array avoids the issue.

matthewp:
@lilnasy Heads up that we are planning our next `minor` release next week (next Thursday). If this PR is ready it could go out then. No rush though, thanks!

lilnasy:
@matthewp it's all done

matthewp (changes requested):
Blocked as this is a minor. Block will be released when the next release occurs.

matthewp (approved):
This looks really great. I appreciate the amount of work this took and how thorough you were with testing and everything else.

lilnasy:
@matthewp I found a mistake in one of the tests I wrote. I'd prefer to rebase and keep the commit history clean, but I'm hesitating because the review is done. Is rebasing now fine?

matthewp:
When PRs get merged they are squashed into a single commit. So you can just push new commits.

packages/astro/src/core/build/plugins/plugin-css.ts
```diff
@@ -25,236 +25,312 @@ interface PluginOptions {
...
-						// CSS is prioritized based on depth. Shared CSS has a higher depth due to being imported by multiple pages.
-						// Depth info is used when sorting the links on the page.
-						if (pageData?.css.has(importedCssImport)) {
-							// eslint-disable-next-line
-							const cssInfo = pageData?.css.get(importedCssImport)!;
-							if (depth < cssInfo.depth) {
-								cssInfo.depth = depth;
-							}
+				// Chunks that have the viteMetadata.importedCss are CSS chunks
+				if (meta.importedCss.size < 1) continue;
```
bluwy, line 114: nit: The comment here is also contradicting the condition check, but since this helps prevent a large indent, maybe we can update the comment:

```js
// Skip if the chunk has no CSS, we want to handle CSS chunks only
```

packages/astro/src/core/build/plugins/plugin-css.ts
```diff
@@ -25,236 +25,312 @@ interface PluginOptions {
...
-		for (const id of hashIds) {
-			hash.update(id, 'utf-8');
-		}
-		const h = hash.digest('hex').slice(0, 8);
-		const proposedName = baseName + '.' + h;
-		return proposedName;
-	}
+	// stylesheet filenames are kept in here until "post", when they are rendered and ready to be inlined
+	const pagesToCss = new Map<string, Map<string, { order: number; depth: number }>>();
+	const pagesToPropagatedCss = new Map<string, Map<string, Set<string>>>();
```
bluwy, line 63: nit: I personally find keeping a temporary map instead of updating the `pageData` directly a bit hard to follow. Is it necessary to refactor this for this feature to work?

I don't have a strong opinion of going back to the previous pattern since the work is already done, but seeing the `.set().get()` is causing an itch in me 😬

packages/astro/src/core/build/plugins/plugin-css.ts
```diff
@@ -25,236 +25,312 @@ interface PluginOptions {
...
+				const assetSize = new TextEncoder().encode(stylesheet.source).byteLength;
+
+				const toBeInlined =
+					inlineConfig === 'always'
+						? true
+						: inlineConfig === 'never'
+						? false
+						: assetSize <= assetsInlineLimit;
+
+				if (toBeInlined) delete bundle[id];
```
bluwy, line 241: Unless I've missed something, this _could_ delete CSS chunks that are only used by framework components. I have not tested if it's true though.

packages/astro/src/core/build/internal.ts
```diff
@@ -224,39 +224,54 @@ export function hasPrerenderedPages(internals: BuildInternals) {
...
+	current: StylesheetAsset
+): Array<StylesheetAsset> {
+	const lastAdded = acc.at(acc.length - 1);
+	const lastWasInline = lastAdded?.type === 'inline';
+	const currentIsInline = current?.type === 'inline';
+	if (lastWasInline && currentIsInline) {
+		const merged = { type: 'inline' as const, content: lastAdded.content + current.content };
+		return [...acc.slice(0, acc.length - 1), merged];
+	}
+	return [...acc, current];
```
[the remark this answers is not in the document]
YoniFeng: https://github.com/withastro/astro/pull/6659#discussion_r1176994205 :(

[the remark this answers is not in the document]
matthewp: I agree with @bluwy here, we really don't use immutable patterns and spread syntax all that much in Astro (at least not in performance critical parts). So even without knowing the impact, and assuming that this is probably not going to kill perf, I'd be more comfortable with this being mutable.

packages/astro/src/core/build/plugins/plugin-css.ts
matthewp: Why do you think that? A chunk used by a framework component is still injected into the page.

bluwy: I just check it locally and this doesn't seem to be an issue, so ignore me 😄 I thought that CSS chunk could be not linked to any page, but seems like it'll always be.

packages/astro/test/css-inline-stylesheets.js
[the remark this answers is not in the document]
bluwy: From discord, there's a weird caching issue when sharing a single fixture, so keeping as is for now. For the test name, I handled it at https://github.com/withastro/astro/pull/6983

PR #6662, opened 2023-03-26
Reduced Playwright test flakiness with click() events in few components

RaphaelBossek:
## Changes

- Reduced Playwright test flakiness with click() events by 150ms delay and enabling 3 retries by default
- Introduced few new script commands to analyse Playwright results:
  - pnpm run test:e2e:report
  - pnpm run test:e2e:debug <filter>
- HTML reports are enabled too, but not shown by default after test run

## Testing

This patch is about testing with Playwright. With this change, I have a more or less reliable test suite. Before no Playwright test run was successful. Now only 2 or 3 tests are affected by retries but run through.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
     "test:vite-ci": "turbo run test --filter=astro --output-logs=new-only --no-deps --concurrency=1",
     "test:e2e": "cd packages/astro && pnpm playwright install && pnpm run test:e2e",
     "test:e2e:match": "cd packages/astro && pnpm playwright install && pnpm run test:e2e:match",
+    "test:e2e:report": "cd packages/astro && pnpm run test:e2e:report",
+    "test:e2e:debug": "cd packages/astro && pnpm playwright install && pnpm run test:e2e:debug",
     "benchmark": "astro-benchmark",
     "lint": "eslint --cache .",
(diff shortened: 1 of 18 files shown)


bluwy:
I don't think setting a `150ms` delay is the way to fix flakiness, if an element needs to be waited longer due to an action, we should wait for the element instead. And 0 tries locally is intentional so tests don't run too long when making quick tests. But I'm curious what others think about this too.

RaphaelBossek:
Getting involved with testing is probably not a reputable topic, judging by the turnout here.

Consulting the Playwright Discord #help-playwright channel with this topic resulted in following answer:

> Alternative to retrying the click action or setting a fixed delay on it would be to:
> - Change the app (system under test) so that the elements don't pass the auto-waiting conditions when the app is not truly ready to handle input. https://playwright.dev/docs/actionability
> - Change the tests to detect when the app is ready to handle input. Add for example an await locator.waitFor(), an auto-waiting assertion or wait for a network response to be received. 

In conclusion that means that the static text elements have to changed to programmatic ones in order to be able to recognise if the component framework is up and running.

matthewp:
Agree with @bluwy here, let's make the tests deterministic by waiting on whatever condition needs to exist before the button can be clicked.

RaphaelBossek:
Will find a solution for that. Good playgroud to improve my skills about Playwright.

PR #6665, opened 2023-03-26
Improve mobile support by adding initial-scale to the viewport meta tag

cevdetardaharan:
## Changes

- Improve mobile support by adding initial-scale to the viewport meta tag.

- What does this change?

## Testing

The changes were tested locally by creating a new Astro project with `npm astro create@latest` and checking the mobile responsiveness of the page.

## Docs

This change should not affect user behavior, and no updates to the documentation are required.


bluwy (commented):
Isn't `initial-scale` already `1` by default? https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#initial-scale

PR #6673, opened 2023-03-27
Remove X-UA-Compability and fix meta tag attributes

cevdetardaharan:
## Changes

- Removed X-UA-Compability meta tag.
- Changed charset attribute to use lowercase 'utf-8'.
- Removed initial-scale attribute from viewport meta tag as it's set to 1 by default.

## Testing

This change was manually tested in various browsers.

## Docs

This change does not affect user behavior and does not require documentation updates.


packages/astro/astro-jsx.d.ts
```diff
@@ -822,7 +822,6 @@ declare namespace astroHTML.JSX {
 	interface MetaHTMLAttributes extends HTMLAttributes {
 		charset?: string | undefined | null;
 		content?: string | URL | undefined | null;
-		'http-equiv'?: string | undefined | null;
```
Princesseuh, line 825: Why was this removed?

cevdetardaharan:
Thank you for your feedback! The reason behind this PR was to make sure that the charset attribute is consistent across all files, using the recommended utf-8 value. As for the X-UA-Compatibility tag, it is unnecessary as IE has been dropped, and its inclusion can cause unexpected behavior in certain situations. Finally, the initial-scale=1 attribute was removed because it is already the default value, and specifying it can lead to unnecessary markup. Here are some articles that support these decisions:

    Regarding the X-UA-Compatibility tag: [Internet Explorer is dead](https://blogs.windows.com/windowsexperience/2022/06/15/internet-explorer-11-has-retired-and-is-officially-out-of-support-what-you-need-to-know/)
    Regarding the initial-scale attribute: [MDN - initial-scale](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#initial-scale)

sarah11918:
Removed the need for a docs review! I checked the change in capitalization to the README, and all's fine!

cevdetardaharan:
It's ready to go but the branch has conflicts
`
packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-simple.astro
packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-with-components.astro
packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-with-config.astro
`
and I can't have `write access`.

ematipico:
> It's ready to go but the branch has conflicts `packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-simple.astro packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-with-components.astro packages/integrations/markdoc/test/fixtures/content-collections/src/pages/content-with-config.astro` and I can't have `write access`.

From you local branch

```shell
git fetch upstream
git merge upstream/main
```

Then resolve the conflicts, and run

```shell
git add .
git commit
git push
```

And you should be good to go.

PR #6676, opened 2023-03-27
Fix: Paginate adds unexpected trailing slash on index route

### LINKED ISSUE #6603: `paginate` create broken `page.url.prev` link for second page (added trailing slash)

2.1.3

Github Pages

yarn

Mac, Linux

Any

### Describe the Bug

`paginate` create broken `page.url.prev` link for second page (added trailing slash)

How to reproduce:

add in astro config:

```
build: {
    format: 'file',
  },
```

Next create `[...page].astro` in `posts` subfolder and add pagination links:

```
{page.url.prev ? <a href={page.url.prev}>Previous</a> : null}
{page.url.next ? <a href={page.url.next}>Next</a> : null}
```

On page `/posts/2` link "Previous" will contain in `href=/posts/` (with trailing slash), all other links works as expected `/posts/number` (no trailing slash). This will work in dev mode, but after build you website will contain broken link (because static hosting often don't have mode `trailingSlash: 'ignore'`)

https://stackblitz.com/edit/github-gfz7vj-kbpqqf?file=src/pages/posts/[...page].astro

### Participation

bholmesdev:
## Changes

- Resolves #6603
- Use `undefined` for index routes during pagination. Otherwise, this inserts an expected trailing `/`. found this also generates `/posts//` by default!

## Testing

- Add tests for next and prev pagination links

## Docs

N/A

### DIFF
--- a/.changeset/eight-humans-push.md
+++ b/.changeset/eight-humans-push.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix next nad previous links for index routes when using pagination
--- a/packages/astro/src/core/render/paginate.ts
+++ b/packages/astro/src/core/render/paginate.ts
@@ -63,7 +63,9 @@ export function generatePaginateFunction(routeMatch: RouteData): PaginateFunctio
 									: routeMatch.generate({
 											...params,
 											page:
-												!includesFirstPageNumber && pageNum - 1 === 1 ? '' : String(pageNum - 1),
+												!includesFirstPageNumber && pageNum - 1 === 1
+													? undefined
+													: String(pageNum - 1),
 									  }),
 						},
 					} as Page,
--- a/packages/astro/src/core/render/paginate.ts
+++ b/packages/astro/src/core/render/paginate.ts
@@ -39,6 +39,21 @@ export function generatePaginateFunction(routeMatch: RouteData): PaginateFunctio
 				...additionalParams,
 				[paramName]: includesFirstPageNumber || pageNum > 1 ? String(pageNum) : undefined,
 			};
+			const current = correctIndexRoute(routeMatch.generate({ ...params }));
+			const next =
+				pageNum === lastPage
+					? undefined
+					: correctIndexRoute(routeMatch.generate({ ...params, page: String(pageNum + 1) }));
+			const prev =
+				pageNum === 1
+					? undefined
+					: correctIndexRoute(
+							routeMatch.generate({
+								...params,
+								page:
+									!includesFirstPageNumber && pageNum - 1 === 1 ? undefined : String(pageNum - 1),
+							})
+					  );
 			return {
 				params,
 				props: {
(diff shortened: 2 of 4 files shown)


bholmesdev:
Will resolve that failed unit test tomorrow! Looks like we expect a `/` for the root path even with trailing slashes removed. This passed with a false positive since we added `/`s everywhere, so might special case if the generated route is still `''` after generation.

.changeset/eight-humans-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix next nad previous links for index routes when using pagination
```
MoustaphaDev, line 5:
```suggestion
Fix next and previous links for index routes when using pagination
```

packages/astro/test/astro-pagination.test.js
```diff
@@ -41,12 +41,28 @@ describe('Pagination', () => {
...
+					expect(prevHref).to.be.undefined;
+					expect(nextHref).to.be.undefined;
+				}
+				if (color === 'blue' && p === 1) {
+					expect(prevHref).to.be.undefined;
+					expect(nextHref).to.equal('/posts/blue/2');
+				}
+				if (color === 'blue' && p === 2) {
+					expect(prevHref).to.equal('/posts/blue/1');
+					expect(nextHref).to.be.undefined;
```
MoustaphaDev, line 64: Do these tests every run? Say when `p = '1'`  ===> `p  ===  1` is false which will cause the test to not run

bholmesdev: Sorry, this is a super imperative way to check if the `params` entry matches one of the items in the list ([see the lines just above](https://github.com/withastro/astro/pull/6676/files#diff-febabdfee1785d50a08874c42cbcb3d8eb6e6f420fb37843dcdce0511cb6f72fR39-R41)). Brain-dead solution I know, but oddly the simplest I could find 😅

MoustaphaDev: I meant that `p` is always a string so comparing it the numbers will return `false`
'1' !=== 1

I'm I overlooking something?

packages/astro/src/core/render/paginate.ts
ematipico, line 1: I was triaging this issue last week, and I managed to track down the issue to this [file](https://github.com/withastro/astro/blob/main/packages/astro/src/core/routing/manifest/generator.ts), which the function that generates the route.

Shouldn't we apply the fix there? The reason why I bailed on this issue was because I am not good with regex :(

ematipico (approved):

packages/astro/src/core/render/paginate.ts
bholmesdev: Ah yep, I was poking this file for a _long_ time. Eventually decided to keep the fix to paginate, since it _was_ passing an invalid parameter (`''` instead of `undefined`). I also don't see a problem with the static paths output that also relies on the generate function. If we change how it behaves, we'd probably want a more fundamental change to cleaning up our trailing slash handling 🤷

packages/astro/test/astro-pagination.test.js
bholmesdev: Ohhhh right you are! I'm a fool. Will fix

MoustaphaDev (approved):

PR #6677, opened 2023-03-27
[create-astro] Error on `--template` that does not exist

bholmesdev:
This reverts commit c13d428a7804b5b9809dbea94a1b17c36714a1e1.

## Changes

- Raise an appropriate error when cloning templates that don't exist. Looks like the giget library only throws when a remote repo does not exist, but _will not_ throw if a specific example inside that repo is missing. This adds an empty dir check, since that's a good chance it didn't work as expected.

## Testing

N/A - dry run prevents actually fetching to avoid flakiness

## Docs

N/A

### DIFF
--- a/packages/create-astro/src/actions/template.ts
+++ b/packages/create-astro/src/actions/template.ts
@@ -35,8 +35,13 @@ export async function template(ctx: Pick<Context, 'template' | 'prompt' | 'dryRu
 			while: () =>
 				copyTemplate(ctx.template!, ctx as Context).catch((e) => {
 					// eslint-disable-next-line no-console
-					error('error', e);
-					process.exit(1);
+					if (e instanceof Error) {
+						error('error', e.message);
+						process.exit(1);
+					} else {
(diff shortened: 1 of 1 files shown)


packages/create-astro/src/actions/template.ts
```diff
@@ -81,11 +86,18 @@ export default async function copyTemplate(tmpl: string, ctx: Context) {
 		} catch (err: any) {
 			fs.rmdirSync(ctx.cwd);
 			if (err.message.includes('404')) {
-				await error('Error', `Template ${color.reset(tmpl)} ${color.dim('does not exist!')}`);
+				throw new Error(`Template ${color.reset(tmpl)} ${color.dim('does not exist!')}`);
```
bholmesdev, line 89: This used to log the error and close the spinner before `ctx.exit` could be called. This throws so the higher scope can catch and log appropriately.

bluwy (approved):

PR #6679, opened 2023-03-27
Fix incorrect path in error overlay on Win

### LINKED ISSUE #6678: Incorrect path to file in the error overlay on windows

2.1.7

None

pnpm

Windows

Chrome

## Incorrect path is shown on Windows in the Error overlay. 

Instead of `C:\astro\examples\minimal\src\pages\index.astro:19:60` it will show `C:\astro\examples\minimal\C:\astro\examples\minimal\src\pages\index.astro:19:60`. 
It appears this is because of a bug [here](https://github.com/withastro/astro/blob/c13d428a7804b5b9809dbea94a1b17c36714a1e1/packages/astro/src/core/errors/dev/utils.ts#L38).

## Steps to reproduce (*Windows only!*)
1. Install the dependencies and run `npm start`.
2. Observe the incorrect path. Try to use the `Open in editor` button unsuccessfully.

https://github.com/fcFn/astro-bad-path-on-win

### Participation

fcFn:
## Changes

Fix #6678.

## Testing

An e2e test was added.

## Docs
This just fixes a bug so no docs changes are needed.

### DIFF
--- a/packages/astro/e2e/errors.test.js
+++ b/packages/astro/e2e/errors.test.js
@@ -52,6 +52,28 @@ test.describe('Error display', () => {
 		expect(await page.locator('vite-error-overlay').count()).toEqual(0);
 	});
 
+	test('shows correct file path when a page has an error', async ({ page, astro }) => {
+		await page.goto(astro.resolveUrl('/import-not-found'));
+
+		const { fileLocation, absoluteFileLocation } = await getErrorOverlayContent(page);
+		const absoluteFileUrl = 'file://' + absoluteFileLocation.replace(/:\d+:\d+$/, '');
+		const fileExists = astro.pathExists(absoluteFileUrl);
+
+		expect(fileExists).toBeTruthy();
+		expect(fileLocation).toMatch(/^pages\/import-not-found\.astro/);
+	});
+
+	test('shows correct file path when a component has an error', async ({ page, astro }) => {
+		await page.goto(astro.resolveUrl('/preact-runtime-error'));
+
+		const { fileLocation, absoluteFileLocation } = await getErrorOverlayContent(page);
+		const absoluteFileUrl = 'file://' + absoluteFileLocation.replace(/:\d+:\d+$/, '');
+		const fileExists = astro.pathExists(absoluteFileUrl);
+
+		expect(fileExists).toBeTruthy();
+		expect(fileLocation).toMatch(/^components\/PreactRuntimeError.jsx/);
+	});
+
 	test('framework errors recover when fixed', async ({ page, astro }) => {
 		await page.goto(astro.resolveUrl('/svelte-syntax-error'));
 
--- a/packages/astro/src/core/errors/dev/utils.ts
+++ b/packages/astro/src/core/errors/dev/utils.ts
@@ -4,11 +4,13 @@ import * as fs from 'node:fs';
 import { isAbsolute, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import stripAnsi from 'strip-ansi';
+import { normalizePath } from 'vite';
 import type { ESBuildTransformResult } from 'vite';
 import type { SSRError } from '../../../@types/astro.js';
 import { AggregateError, type ErrorWithMetadata } from '../errors.js';
 import { codeFrame } from '../printer.js';
 import { normalizeLF } from '../utils.js';
+import { removeLeadingForwardSlashWindows } from '../../path.js';
 
 type EsbuildMessage = ESBuildTransformResult['warnings'][number];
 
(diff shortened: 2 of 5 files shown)


fcFn:
Ah, it seems that the title in the file path that is shown in the error overlay is supposed to be absolute while the text is supposed to be relative to the project root? That's how it is on Linux. 

Well, I fixed the double issue path on Windows so that now it at least shows a valid Windows path and the `Open in editor` button works on Windows (and the tests now pass on both Linux and Windows). 

Should I work some more on this and make it show the relative path in the text content like it does on Linux?

fcFn:
I just discovered that while fixing the path on Windows when the error is in `pages` the path will now break if the error is in a component 😅 I will investigate further.

fcFn:
I fixed it and added another test to check the path when an error is in a component.

packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -32,10 +32,15 @@ export function collectErrorMetadata(e: any, rootFolder?: URL | undefined): Erro
...
 		// - The "Open in Editor" button won't work
+		const isWindows = process.platform === 'win32';
+		if (isWindows && error.loc?.file) {
+			error.loc.file = error.loc.file.replace(/^\//, '');
+		}
 		if (
 			error.loc?.file &&
 			rootFolder &&
-			(!error.loc.file.startsWith(rootFolder.pathname) || !isAbsolute(error.loc.file))
+			((!error.loc.file.startsWith(rootFolder.pathname) && !isWindows) ||
```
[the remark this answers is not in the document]
Princesseuh: The path utilities in `packages/astro/src/core/path.ts` might help you make this a bit cleaner. An explanation of what it is doing would also be great!

packages/astro/src/core/errors/overlay.ts
```diff
@@ -636,7 +636,8 @@ class ErrorOverlay extends HTMLElement {
 			const codeContent = code.querySelector<HTMLDivElement>('#code-content');
 
 			if (codeHeader) {
-				const cleanFile = err.loc.file.split('/').slice(-2).join('/');
+				const separator = err.loc.file.includes('/') ? '/' : '\\';
```
fcFn, line 639: Overlay didn't account for Windows separators here which and this change fixes it.

packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -32,10 +33,15 @@ export function collectErrorMetadata(e: any, rootFolder?: URL | undefined): Erro
 		// - We'll fail to show the file's content in the browser
 		// - We'll fail to show the code frame in the terminal
 		// - The "Open in Editor" button won't work
+
+		// Normalize the paths so that we can correctly detect if it's absolute on any platform
+		const normalizedFile = error.loc?.file?.replace(/\\/g, '/');
+		const normalizedRootFolder = removeLeadingForwardSlashWindows(rootFolder?.pathname || '');
```
fcFn, line 39: Remove a leading forward slash which `rootFolder.pathname` comes with which is invalid on Windows (e.d. `/c:/`) and will always fail the check for the path being absolute below which results in the bug that is described in the issue.

packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -32,10 +33,15 @@ export function collectErrorMetadata(e: any, rootFolder?: URL | undefined): Erro
 		// - We'll fail to show the file's content in the browser
 		// - We'll fail to show the code frame in the terminal
 		// - The "Open in Editor" button won't work
+
+		// Normalize the paths so that we can correctly detect if it's absolute on any platform
+		const normalizedFile = error.loc?.file?.replace(/\\/g, '/');
```
fcFn, line 38: Normalize the separators because if we are on Windows the path sometimes uses `/` (if the error is in a component) and sometimes uses `\\` (if the error is in a page).

packages/astro/e2e/errors.test.js
```diff
@@ -52,6 +52,28 @@ test.describe('Error display', () => {
 		expect(await page.locator('vite-error-overlay').count()).toEqual(0);
 	});
 
+	test('shows correct file path when a page has an error', async ({ page, astro }) => {
```
fcFn, line 55: Two different tests because on Windows the path separators are different depending on the location of the error (i.e. component vs page).

fcFn:
@Princesseuh Thank you for the hint about the path utilities! 

I think I found the underlying cause and made the fix more robust. I added a comment for the less clear part and I added more explanations in the comments for the PR as I'm not sure which of those should be added to the code.

packages/astro/src/core/errors/dev/utils.ts
fcFn: @Princesseuh Thank you so much for pointing that out! Pushed a fix.

Princesseuh (approved):

PR #6680, opened 2023-03-27
Fix image regeneration after changing image service

koriwi:
## Changes
I experienced very strange behaviour which resulted in images being rotated or not depending on how i change the widths or alt tags of images. I tracked this down to the image cache not being regenerated after i switched to sharp.
The behaviour of sharp differs in some ways. In my case sharp rotates the image while removing the exif data. squoosh does remove the exif data without rotating the images. When my image was the exact same image as before I got served the cached version built previously by squoosh so some images were still up side down etc. Changing the alt tag invalidated the cache, as it is used for hashing.

- Don't use the alt tag for hashing as it has nothing to do with the actual content of the image
- Include the serviceEntryPoint in the hashKey to invalidate the image cache when changing from squoosh to sharp for example.

## Testing

I did not add unit tests as my experience in this repo is very slim right now.
I verified that the behaviour is as intended by reproducing the mentioned behaviour manually.

## Docs

Now the behaviour is as intended so i don't think with a fix like this we need any additional docs

### DIFF
--- a/.changeset/lovely-owls-sniff.md
+++ b/.changeset/lovely-owls-sniff.md
@@ -1,0 +1,6 @@
+---
+'astro': patch
+'@astrojs/image': patch
+---
+
+Invalidates cache when changing serviceEntryPoint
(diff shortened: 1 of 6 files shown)


Princesseuh (approved):
Looks great, thank you for making the change in both `astro:assets` and `@astrojs/image`

PR #6681, opened 2023-03-27
Update lit-ssr dependency

e111077:
## Changes

- Updates lit-ssr dependency to match that of `lit@^2.7.0`'s requirements
- 

## Testing

Usual test run

## Docs

Should not require behavior change


packages/integrations/lit/server-shim.js
```diff
@@ -3,12 +3,11 @@ import { installWindowOnGlobal } from '@lit-labs/ssr/lib/dom-shim.js';
 if (typeof fetch === 'function') {
 	const _fetch = fetch;
 	installWindowOnGlobal();
-	globalThis.fetch = window.fetch = _fetch;
+	globalThis.fetch = _fetch;
 } else {
 	installWindowOnGlobal();
 }
```
augustjk, line 10: I think this entire block may be unnecessary now as the dom shim is built into the node builds of `lit`.

packages/integrations/lit/server-shim.js
```diff
@@ -3,12 +3,11 @@ import { installWindowOnGlobal } from '@lit-labs/ssr/lib/dom-shim.js';
...
 
-window.global = window;
 document.getElementsByTagName = () => [];
 // See https://github.com/lit/lit/issues/2393
 document.currentScript = null;
```
augustjk, line 13: These also may be unneeded. Not sure about `document.getElementsByTagName` but `document.currentScript` was definitely needed due to the introduction of `window` in global, which will not be needed with the shims in the node builds.

packages/integrations/lit/server-shim.js
e111077: In runtime for some reason it does seem that `globalThis.fetch !== _fetch` for some reason, so I included it. WDYT?

packages/integrations/lit/server-shim.js
e111077: `document.gEBTN` definitely needed in server.js

packages/astro/e2e/lit-component.test.js
```diff
@@ -8,10 +8,6 @@ const test = testFactory({
 // TODO: configure playwright to handle web component APIs
 // https://github.com/microsoft/playwright/issues/14241
 test.describe('Lit components', () => {
-	test.beforeAll(() => {
-		delete globalThis.window;
```
matthewp, line 12: No longer needs to shim the window?

matthewp (approved):

packages/astro/e2e/lit-component.test.js
e111077: All of the shims needed for lit are baked into Lit's node build. I just pushed a commit though because we do need to actually re-shim only `customElements.define`because our baked-in shims in ReactiveElement's node build is forgiving and tries to call anything anyone else has shimmed, but we have a small hack we are relying on so we re-shim that

.changeset/gold-windows-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/lit': patch
```
augustjk, line 2: It is technically a breaking change due to the changes in the available dom shim. would it make sense to do a new major?

packages/integrations/lit/server-shim.js
```diff
@@ -1,20 +1,28 @@
...
+if (globalThis.document) {
+	document.currentScript = null;
 }
 
-window.global = window;
-document.getElementsByTagName = () => [];
-// See https://github.com/lit/lit/issues/2393
-document.currentScript = null;
+// Astro seems to have a DOM shim and the only real difference that we need out
+// of the Lit DOM shim is that the Lit DOM shim does something that triggers
```
augustjk, line 13: Perhaps worth noting instead of "something" that we specifically need `observedAttributes` to be accessed. https://github.com/lit/lit/blob/a4eb5e60d393db3cd9d9a7f82c0c37457cd368eb/packages/labs/ssr-dom-shim/src/index.ts#L121-L123

.changeset/gold-windows-fly.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Bump deps to lit-ssr 3.0 and polyfill supoprt for template[shadowrootmode] support. Also fix breakage caused by lit@2.7.0 hydration behavior requiring lit-ssr 3.0
```
augustjk, line 5:
```suggestion
Update to use `@lit-labs/ssr@^3`
**[BREAKING]** DOM shim required for Lit SSR has been greatly reduced. `window`, `document`, and other objects are no longer available in global. Most SSR-ready component code should not be affected but, if so, they can be fixed with optional chaining or by using the `isServer` environment checker from the `lit` package. See [lit.dev docs on authoring components for SSR].(https://lit.dev/docs/ssr/authoring/#browser-only-code)
**[BREAKING]** Adds compatibility with `lit@2.7.0` hydration behavior. Do not update if you're using `lit@2.6.1` or lower.
Includes support for template[shadowrootmode] support.
```

augustjk (commented):
This part of the readme should be updated
https://github.com/withastro/astro/blob/fc71c3f18819ac3ad62809a7eeff5fe7840f2c4b/packages/integrations/lit/README.md?plain=1#L123-L142

e111077:
@matthewp would also like to note that we have to re-shim HTMLElement because Astro's DOM shim doesn't seem to do anything when you do

```ts
element.setAttribute('foo', 'bar')
element.setAttribute('baz', 'quux')
element.getAttribute('foo') // bar
element.attributes // ['foo', 'baz']
```

https://github.com/withastro/astro/blob/main/packages/webapi/src/lib/Element.ts#L25-L32

I wonder WYT about implementing that in Astro?

e111077:
You also might want to squash merge this one 🙈

matthewp:
Thanks! I'm not sure how we feel about having the DOM shim in long term. Sort of feel like the renderer should bring its own if needed. So I think what you are doing here is fine for now, thanks!

matthewp:
@augustjk is this ok with you now?

augustjk:
Looking good, thank you!!

PR #6699, opened 2023-03-28
tailwind integrations guide: use <PackageManagerTabs> component

Nuc1eoN:
Update doc page to make use of the nicer looking \<PackageManagerTabs> for package manager selection.

Same as used in https://github.com/withastro/docs/blob/main/src/content/docs/en/install/auto.mdx?plain=1#L32


delucis:
Thanks so much for this thoughtful contribution! It would be awesome if we _could_ use the `<PackageManagerTabs>` on this page, but unfortunately this README needs to be viewed on GitHub and npm as well as in our docs site, so we can’t really use our custom components here. One path we could explore is developing some kind of custom syntax the docs site understands, where HTML comments get mapped to components but are invisible in regular Markdown contexts. But for now unfortunately this isn’t possible.

Thanks again! Maybe some of our [“good first issues” in docs](https://github.com/withastro/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) would be interesting if you’re looking to get involved in contributing.

PR #6703, opened 2023-03-29
Move `image()` to be passed as part of `schema` 

Princesseuh:
## Changes

As the title says, this moves `image()` to be provided through `schema`, like such `schema: ({image}) => z.object({...})` instead of something you import from `astro:content`. This has fairly consequential benefits on our end, as we can now directly pass information about the current entry to schema instead of trying to clumsily patch it

## Testing

Tests still pass!

## Docs

WIP

### DIFF
--- a/packages/astro/src/content/runtime-assets.ts
+++ b/packages/astro/src/content/runtime-assets.ts
@@ -1,20 +1,28 @@
-import { pathToFileURL } from 'url';
+import type { PluginContext } from 'rollup';
 import { z } from 'zod';
-import {
-	imageMetadata as internalGetImageMetadata,
-	type Metadata,
-} from '../assets/utils/metadata.js';
+import type { AstroSettings } from '../@types/astro.js';
+import { emitESMImage } from '../assets/index.js';
 
-export function createImage(options: { assetsDir: string; relAssetsDir: string }) {
+export function createImage(
+	settings: AstroSettings,
+	pluginContext: PluginContext,
+	entryFilePath: string
+) {
+	if (!settings.config.experimental.assets) {
+		throw new Error('Enable `experimental.assets` in your Astro config to use image()');
+	}
+
 	return () => {
-		if (options.assetsDir === 'undefined') {
-			throw new Error('Enable `experimental.assets` in your Astro config to use image()');
-		}
+		return z.string().transform(async (imagePath, ctx) => {
+			const resolvedFilePath = (await pluginContext.resolve(imagePath, entryFilePath))?.id;
+			const metadata = await emitESMImage(
+				resolvedFilePath,
+				pluginContext.meta.watchMode,
+				pluginContext.emitFile,
+				settings
+			);
 
-		return z.string({ description: '__image' }).transform(async (imagePath, ctx) => {
-			const imageMetadata = await getImageMetadata(pathToFileURL(imagePath));
-
-			if (!imageMetadata) {
+			if (!metadata) {
 				ctx.addIssue({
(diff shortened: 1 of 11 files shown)


.changeset/many-dancers-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Move `image()` to come from `schema` instead to fix it not working with refine and inside complex types
```
bholmesdev, line 5: Could had a migration section here to show necessary code changes. [Here's an example](https://github.com/withastro/astro/pull/6653/files#diff-b20a481d3c7673a1da8222b0c6428b3b6fac2d1e97316aa4a59efdc122be21adR8) from the Markdoc 0.1 change

packages/astro/src/content/runtime-assets.ts
```diff
@@ -1,20 +1,28 @@
...
+import type { AstroSettings } from '../@types/astro.js';
+import { emitESMImage } from '../assets/index.js';
+
+export function createImage(
+	settings: AstroSettings,
+	pluginContext: PluginContext,
+	entryFilePath: string
+) {
+	if (!settings.config.experimental.assets) {
+		throw new Error('Enable `experimental.assets` in your Astro config to use image()');
```
bholmesdev, line 12: nit!
```suggestion
		throw new Error('Enable `experimental.assets` in your Astro config to use `image()`');
```

bholmesdev (commented):
Some callouts on documentation. Couldn't `image`ine our internals would get this simple with the change!

bluwy (approved):
The new flow looks great 👍

Princesseuh:
!preview schema-image

Princesseuh:
!preview schema-image

ematipico (approved):

packages/astro/src/content/template/virtual-mod.mjs
```diff
@@ -11,6 +11,13 @@ export function defineCollection(config) {
 	return config;
 }
 
+// TODO: Remove this when having this fallback is no longer relevant. 2.3? 3.0? - erika, 2023-04-04
+export const image = () => {
+	throw new Error(
+		'Importing image() from `astro:content` is no longer supported. See https://docs.astro.build/en/guides/assets/#update-content-collections-schemas for more information.'
```
bholmesdev, line 17:
```suggestion
		'Importing `image()` from `astro:content` is no longer supported. See https://docs.astro.build/en/guides/assets/#update-content-collections-schemas for our new import instructions.'
```

Nit: new wording to make it clear there's a _new_ way to do it, and the feature didn't disappear

bholmesdev (approved):

packages/astro/src/content/template/types.d.ts
```diff
@@ -13,8 +13,27 @@ declare module 'astro:content' {
 	export type CollectionEntry<C extends keyof typeof entryMap> =
 		(typeof entryMap)[C][keyof (typeof entryMap)[C]];
 
+	// TODO: Remove this when having this fallback is no longer relevant. 2.3? 3.0? - erika, 2023-04-04
+	/**
+	 * @deprecated
```
Xetera, line 18: Should the existing behavior here be changing if it's being marked as deprecated? Or is it just here for the visual cue in editors

Princesseuh: Just there for the visual cue. Since the feature is experimental, we don't care much about API stability and maintaining both options doesn't really work anyway.

PR #6706, opened 2023-03-30
TDD pattern development

wulinsheng123:
## Changes

related to [pr](https://github.com/withastro/roadmap/issues/531)
- What does this change?

### DIFF
--- a/packages/astro/test/astro-minification-html.test.js
+++ b/packages/astro/test/astro-minification-html.test.js
@@ -1,0 +1,24 @@
+import { expect } from 'chai';
+import * as cheerio from 'cheerio';
+import { loadFixture, isWindows } from './test-utils.js';
+
+describe('minification html', () => {
+	let fixture;
+	const regex = /[ \n]/;
+	before(async () => {
+		fixture = await loadFixture({ root: './fixtures/minification-html/' });
+		await fixture.build();
+	});
+
+	describe('Build', () => {
+		before(async () => {
+			await fixture.build();
+		});
+
+		it('validating the html is or not is minification ', async () => {
+			const html = await fixture.readFile('/index.html');
+			expect(regex.test(html)).to.equal(false);
+		});
+
+	});
+});
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -99,6 +99,7 @@
     "dev": "astro-scripts dev --copy-wasm --prebuild \"src/runtime/server/astro-island.ts\" --prebuild \"src/runtime/client/{idle,load,media,only,visible}.ts\" \"src/**/*.{ts,js}\"",
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
     "test:unit": "mocha --exit --timeout 30000 ./test/units/**/*.test.js",
+    "t": "mocha --exit --timeout 30000 ./test/astro-minification-html.test.js",
     "test:unit:match": "mocha --exit --timeout 30000 ./test/units/**/*.test.js -g",
     "test": "pnpm run test:unit && mocha --exit --timeout 20000 --ignore **/lit-element.test.js && mocha --timeout 20000 **/lit-element.test.js",
     "test:match": "mocha --timeout 20000 -g",
--- a/packages/astro/test/fixtures/minification-html/astro.config.mjs
+++ b/packages/astro/test/fixtures/minification-html/astro.config.mjs
@@ -1,0 +1,6 @@
+import { defineConfig } from 'astro/config';
+
+// https://astro.build/config
+export default defineConfig({
+	compact: true
+});
--- a/packages/astro/test/fixtures/minification-html/package.json
+++ b/packages/astro/test/fixtures/minification-html/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "@test/minification-html",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
+    "astro": "workspace:*"
+  }
+}
--- a/packages/astro/test/fixtures/minification-html/src/pages/index.astro
+++ b/packages/astro/test/fixtures/minification-html/src/pages/index.astro
@@ -1,0 +1,14 @@
+---
+
+---
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width" />
+    <title>minimum html</title>
+  </head>
+  <body>
+    <main>
+    </main>
+  </body>
+</html>
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.4
+lockfileVersion: '6.0'
 
 overrides:
   tsconfig-resolver>type-fest: 3.0.0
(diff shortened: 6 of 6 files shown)


packages/astro/src/core/config/schema.ts
```diff
@@ -22,6 +22,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 		assets: '_astro',
 		serverEntry: 'entry.mjs',
 	},
+	compact: true,
```
matthewp, line 25: Default should be `false`

wulinsheng123:
@matthewp cc

packages/astro/src/@types/astro.ts
```diff
@@ -438,6 +438,23 @@ export interface AstroUserConfig {
...
 	site?: string;
 
+
+	/**
+	 * @docs
+	 * @name compressHTML
+	 * @type {boolean}
+	 * @default `true`
+	 * @description
+	 * Set to `true` to enable compressHTML mode. This will remove all whitespace from your final HTMl file, including newlines. This is useful for reducing the size of your final build html bundle
```
sarah11918, line 448:
```suggestion
	 * Astro removes all whitespace from your final HTML file, including line breaks, by default. This is useful for reducing the size of your final build's HTML bundle. To disable this, set the `compressHTML` flag to `false`. 
```
Is "newline" a proper term? I couldn't find a definition, or see this term used anywhere as one word, but more commonly I saw reference to removing "line breaks."  If that is what's meant here, then I think it seems to be a more standard phrase.

Also, it seemed odd to say "set to true" when the default is already true. We should probably standardize this, but right now for other Booleans, we've used "Astro does X by default. Set to false to disable." 


Alternatively, we also have precedence wording like 
> Controls whether or not to remove all whitespace from your final HTML file, including line breaks. This is useful for reducing the size of your final build's HTML bundle.

So, we're not standardized yet (though we should be!) but in either case, if the default is true, we should not be showing/instructing setting to true. If in fact `default: true` is wrong, and it should be false, then the "Controls whether..." wording also works. Or, change the above suggestion to:

> Astro can remove all whitespace from your final HTML file, including line breaks, by default. This is useful for reducing the size of your final build's HTML bundle. To enable this, set the `compressHTML` flag to `true`.

packages/astro/src/@types/astro.ts
```diff
@@ -438,6 +438,23 @@ export interface AstroUserConfig {
...
+	 *
+	 * ```js
+	 * {
+	 *   compressHTML: true
```
sarah11918, line 452: Be sure to only show the non-default option here, the one that actually requires being set.

(See my comment above: if the default is true, then there's no reason to show true here as the user would never do that. If the default is actually false, then this example is correct, but the default above needs to be changed.)

sarah11918 (commented):
Thanks for this, @wulinsheng123! One little paragraph raised a couple of questions below.  :sweat_smile: 

I'm going to take this opportunity to converge on some standardizing of how we define these Booleans, so thanks for being my guinea pig and putting a little extra work in here! I've offered a couple of suggestions for wording that match the structure of other terms on this page, and want to double check what the default value actually is (true or false) in order to make the example as helpful and realistic as possible.

I leave it in y'all's hands to come up with something reasonable based on my feedback!  :raised_hands:

matthewp:
I submitted the stage 3 RFC here: https://github.com/withastro/roadmap/pull/581

packages/astro/src/@types/astro.ts
```diff
@@ -462,6 +462,24 @@ export interface AstroUserConfig {
 	 */
 	site?: string;
 
+
+	/**
+	 * @docs
+	 * @name compressHTML
+	 * @type {boolean}
+	 * @default `true`
```
ematipico, line 470:
```suggestion
	 * @default `false`
```

The RFC says that by default the option should be `false`

packages/astro/src/@types/astro.ts
```diff
@@ -462,6 +462,24 @@ export interface AstroUserConfig {
...
+	 * @description
+	 * Astro removes all whitespace from your final HTML file, including line breaks, by default. This is useful for reducing the size of your final build's HTML bundle. 
+	 * To disable this, set the `compressHTML` flag to `false`.
```
ematipico, line 473: Let's rephrase this. The option is opt-in; the user must enable it.

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,42 @@
+import { expect } from 'chai';
+import * as cheerio from 'cheerio';
+import { loadFixture, isWindows } from './test-utils.js';
+
+describe('minification html', () => {
+	describe('in the dev', () => {
```
ematipico, line 6:
```suggestion
	describe('in DEV enviroment', () => {
```

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,42 @@
...
+				root: './fixtures/minification-html/',
+			});
+			devServer = await fixture.startDevServer();
+		});
+
+		after(async () => {
+			devServer.stop();
+		});
+
+		it('Verify that the HTML code is compressed in the dev', async () => {
```
ematipico, line 21:
```suggestion
		it('should emit compressed HTML in the emitted file', async () => {
```

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,42 @@
...
+			let res = await fixture.fetch(`/`);
+			expect(res.status).to.equal(200);
+			const html = await res.text();
+			expect(regex.test(html.slice(-100))).to.equal(false);
+		});
+
+	})
+	describe('build', () => {
```
ematipico, line 29:
```suggestion
	describe('in SSG mode', () => {
```

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,42 @@
...
+		let fixture;
+		const regex = /[\r\n]+/gm;
```
ematipico, line 31: nit: no need to repeat the regex, we can define it once at the top of the file, and we can give it a better name

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,42 @@
...
+		before(async () => {
+			fixture = await loadFixture({ root: './fixtures/minification-html/' });
+			await fixture.build();
+		});
+
+		it('Verify that the HTML code is compressed in the pro', async () => {
+			const html = await fixture.readFile('/index.html');
+			expect(regex.test(html.slice(20))).to.equal(false);
```
ematipico, line 39: Could you explain to me how the assertion works? What's `20`? I would find more understandable the followings:
- assign `20` to a variable with a good name, so I understand what's that about
- update the description of the test that explains the assertion
- change the assertion to match the **expected text** instead

ematipico (changes requested):

ematipico (commented):
I believe we should add a test for SSR too, what do you think?

packages/astro/test/astro-minification-html.test.js
ematipico: I don't know what we are testing here. It's a genuine question; I don't understand! 😅

wulinsheng123: I agree with what you said.
I will get a good name for that 20.
thanks for you review

ematipico (approved):
Nice work! 🚀

packages/astro/test/astro-minification-html.test.js
matthewp: Explained offline, but what is happening is that doctype is on a separate line, so the output HTML is like this:

```html
<!doctype html>
<html> ..... 
```

So this is grabbing everything after the doctype and verifying it's all just one line. I'll clean up the tests and add some comments so it's more obvious.

packages/astro/test/astro-minification-html.test.js
bluwy, line 1: nit: we usually name the test file with the same name as the fixture, so I think we can rename this to `minification-html.test.js`. It makes it easier to find the link between the test file and the fixture too.

packages/astro/test/astro-minification-html.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+		after(async () => {
+			devServer.stop();
+		});
+
+		it('should emit compressed HTML in the emitted file', async () => {
+			let res = await fixture.fetch(`/`);
+			expect(res.status).to.equal(200);
+			const html = await res.text();
+			expect(NEW_LINES.test(removeDoctypeLineInDev())).to.equal(false);
```
bluwy, line 43:
```suggestion
			expect(NEW_LINES.test(removeDoctypeLineInDev(html))).to.equal(false);
```

The `html` wasn't passed in here, causing the test to fail.

bluwy (commented):
Implementation looks great! Have one small nit below.

bluwy (approved):

matthewp (dismissed):
Blocking to prevent accidental merge before release day.

packages/astro/src/@types/astro.ts
```diff
@@ -462,6 +462,24 @@ export interface AstroUserConfig {
...
 
+
+	/**
+	 * @docs
+	 * @name compressHTML
+	 * @type {boolean}
+	 * @default `false`
+	 * @description
+	 * When enabled Astro removes all whitespace from your final HTML, including line breaks. This is useful for reducing the size of HTML transfered over the wire.
+	 * To enable this, set the `compressHTML` flag to `true`.
```
matthewp, line 473: From docs, we should explain that this feature is enabled in both development and production.

packages/astro/src/@types/astro.ts
```diff
@@ -462,6 +462,24 @@ export interface AstroUserConfig {
...
 	site?: string;
 
+
+	/**
+	 * @docs
+	 * @name compressHTML
+	 * @type {boolean}
+	 * @default `false`
+	 * @description
+	 * When enabled Astro removes all whitespace from your final HTML, including line breaks. This is useful for reducing the size of HTML transfered over the wire in the development and the production.
```
sarah11918, line 472:
```suggestion
	 * This is an option to minify your HTML output and reduce the size of your HTML files. When enabled, Astro removes all whitespace from your HTML, including line breaks, from `.astro` components. This occurs both in development mode and in the final build.
```
It felt like the changeset had more (and important!) details than the actual docs did.  :smile: Would something like this be correct?

matthewp: Yes!

matthewp (approved):

PR #6714, opened 2023-03-30
Add `build.assetsPrefix` option for CDN support

bluwy:
## Changes

RFC: https://github.com/withastro/roadmap/pull/545

The changes are honestly `base` -> `assetsPrefix || base` everywhere for asset links. And judiciously handle `/` where `base` needs to be prepended by `/`, and `assetsPrefix` doesn't.

## Testing

Added tests for links, images (experimental and `@astrojs/image`), content collections, and SSR.

Manual live testing: https://github.com/bluwy/astro-cdn-test and https://bluwy.github.io/astro-cdn-test/

## Docs

Document `import.meta.env.ASSETS_PREFIX` at https://github.com/withastro/docs/pull/2971.

Documentation for `build.assetsPrefix` is in this PR's `@astro/types.ts`


packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -134,8 +134,13 @@ function buildManifest(
...
 
-	const bareBase = removeTrailingForwardSlash(removeLeadingForwardSlash(settings.config.base));
-	const joinBase = (pth: string) => (bareBase ? bareBase + '/' + pth : pth);
+	const prefixAssetPath = (pth: string) => {
+		if (settings.config.build.assetsPrefix) {
+			return joinPaths(settings.config.build.assetsPrefix, pth);
+		} else {
+			return prependForwardSlash(joinPaths(settings.config.base, pth));
+		}
+	};
```
bluwy, line 143: I'm not sure if this is exactly a breaking change for the SSR manifest. Before asset paths:

- `foo.css`
- `custom-base/foo.css`
- `https://example.com/_astro/foo.css`

After asset paths:

- `/foo.css`
- `/custom-base/foo.css`
- `https://example.com/_astro/foo.css`

This was changed because these raw paths are passed to 

https://github.com/withastro/astro/blob/06ddd216133fa19aa1a9ee7cf747091855730eab/packages/astro/src/core/render/ssr-element.ts#L5-L13

without `base` and `assetsPrefix` params (how SSR works). But if we let the "After" paths into the function, the `else` condition needs to check includes `://` before prepending a `/` which feels kinda hacky, so I changed this to the "Before" paths to simplify it.

packages/astro/src/@types/astro.ts
```diff
@@ -612,6 +612,27 @@ export interface AstroUserConfig {
...
+		 * @docs
+		 * @name build.assetsPrefix
+		 * @type {string}
+		 * @default `undefined`
+		 * @version 2.2.0
+		 * @description
+		 * Specifies the prefix for Astro-generated asset links, this can be used if assets are served from a different domain than the current site.
+		 * 
+		 * If this is set to `https://cdn.example.com`, assets will be fetched from `https://cdn.example.com/_astro/...` (regardless of the `base` option).
+		 * To change the `_astro` path, you can use the `build.assets` option to rename it.
```
sarah11918, line 625:
```suggestion
		 * Specifies the prefix for Astro-generated asset links. This can be used if assets are served from a different domain than the current site.
		 * 
		 * For example, if this is set to `https://cdn.example.com`, assets will be fetched from `https://cdn.example.com/_astro/...` (regardless of the `base` option).
		 * To rename the `_astro` path, specify a new directory in `build.assets`.
```
What do you think of the shorter renaming instruction?

sarah11918 (commented):
Looking good! Left some tiny edits for consideration!  :raised_hands:

packages/astro/src/@types/astro.ts
bluwy: Much nicer. Thanks!

matthewp (changes requested):
Looks like our `minor` GH action isn't work.

Yan-Thomas (approved):
Docs LGTM!

packages/astro/src/core/build/plugins/plugin-ssr.ts
bluwy: Yeah it did have the `/` for `manifest.assets` added here:

https://github.com/withastro/astro/blob/9846b3e93588626b73026b3f6499c1a62df79a24/packages/astro/src/core/build/plugins/plugin-ssr.ts#L215

but not for scripts/styles derived from here:

https://github.com/withastro/astro/blob/9846b3e93588626b73026b3f6499c1a62df79a24/packages/astro/src/core/build/plugins/plugin-ssr.ts#L165-L196

The code above are linked to `main`, which used the `joinBase` helper, and the helper strips the leading slash.

bluwy:
[DOCS] I've pushed another commit to explain the process of serving assets through a third-party domain as suggested by @delucis

ematipico (commented):
I think this looks good to me. What's the expectation of the value should be? It's marked as a `string`, but if the user puts "anything_fancy", what's the outcome of the build? Should we do some validation? Should we warn the user?

bluwy:
It would act as a normal prefix so it could be anything. I'm mostly testing values like `https://cdn.example.com` for now, but it's intentionally lax to allow values like `/some/other/root` if they're deployment setup relies on it.

ematipico:
Perfect! Thank you for clarifying!

matthewp:
@bluwy I think in config.ts we specify `site` to be a URL (it gets refined with new URL()` so maybe it makes sense to do the same here, so that any non-URLs would throw?

bluwy:
The idea is that it doesn't necessarily need to be a URL, it can be any string prefix depending on their deployment, so I'm not sure if we need that restriction, unless we want to only support URL for now?

gkatsanos:
> It would act as a normal prefix so it could be anything. I'm mostly testing values like `https://cdn.example.com` for now, but it's intentionally lax to allow values like `/some/other/root` if their deployment setup relies on it.

I think I found an issue with this second type of values: https://github.com/withastro/astro/issues/6844

PR #6716, opened 2023-03-31
Handle Vite 4.3 module graph changes

bluwy:
## Changes

**DRAFT until the Vite PR is merged**

Handle the changes for https://github.com/vitejs/vite/pull/11973

We use the module graph to scan for CSS to prevent FOUC. We crawl it using `mod.importedModules`, but the field is meant for client-side imports only. Vite had mixed it with SSR and is now un-mixing it in https://github.com/vitejs/vite/pull/11973.

Although this feels like a breaking change, I think it can be justified that we should've used `mod.ssrTransformResult.deps` instead for SSR stuff, and that `mod.importedModules` were only supposed to be client-side only.

Vite 4.3 will introduce `mod.ssrImportedModules` instead so I've updated to anticipate it when we upgrade to Vite 4.3.

## Testing

Existing tests should pass. Though we're not using Vite 4.3 yet, vite-ecosystem-ci will test against it.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/selfish-adults-tie.md
+++ b/.changeset/selfish-adults-tie.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Handle Vite 4.3 module graph changes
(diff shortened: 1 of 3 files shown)


bluwy:
Closing as the Vite change will happen in Vite 5.

Issue #6644, opened 2023-03-24
Unclear `Astro.redirect` behavior in SSR


y-nk:
### What version of `astro` are you using?

latest

### Are you using an SSR adapter? If so, which one?

vercel@latest

### What package manager are you using?

yarn

### What operating system are you using?

doesn't matter

### What browser are you using?

chromium

### Usecase

I'm trying to create a "Protected Route" pattern where a page would be wrapped in a `<ProtectedRoute>` component which would redirect to `/login` if a user is not logged in.

### Problem

1. I've tried with a layout → there's an error
2. I've tried with a component → there's an error
3. I've tried with a separate page → it gets ignored

### Question

Since we must `return Astro.redirect()`, it seems we can't wrap things up a level upper. **How to proceed then?**

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-fnw2f4?file=src/pages/from-page.astro,src/pages/index.astro,src/pages/redirected.astro,src/pages/from-layout.astro,src/pages/from-component.astro,src/pages/from-other-page.astro&on=stackblitz

### Participation

- [X] I am willing to submit a pull request for this issue.

Issue #6632, opened 2023-03-23
Response cookies can't be set when page is .astro


sausir1:
### What version of `astro` are you using?

2.1.4

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Linux, Ubuntu 20.04

### What browser are you using?

Firefox, Chrome

### Describe the Bug

I think the title says it all. But to be more clear - I have a `Layout.astro` component in my `/layouts` directory, where a cookie is set based on given props/params (eg. `Astro.cookies.set('layout', 'test')`. Everything works fine when requesting a page that is not another `.astro` file - Cookie is being sent back and can be seen in response.  But, if I navigate to `index.astro` (does not matter if it set's any cookies or not), the response does not include the cookie that is being  set in `Layout.astro` file anymore. Don't really know if it has to work like this, but the behavior is quite unexpected I think.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-7j549u?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
You can only set cookies inside of page components due to streaming. By the time the Layout runs the stream has already started, headers are sent. We should probably give you an error when you attempt to set them lower in the page.

sausir1:
Well, if I would get an error when setting cookies in a layout component, how do I set them from `.vue`, `.md` and other page files? Maybe a warning is better indication, rather then just stopping execution?

matthewp:
You can't set cookies from framework components.

Skn0tt:
I spent a couple hours debugging why my cookies set in a component didn't work, until I've stumbled across this issue. It worked fine in local dev, so I was expecting it to work fine in prod as well! I'm assuming there's some underlying race condition here.

Would it be possible to add some kind of runtime detection to this? A console warning would've been enough for me to catch this a lot earlier. If you can give me some pointers as to how to implement it, i'm happy to open a PR :D

PR #6719, opened 2023-03-31
Better errors for when response is already sent

matthewp:
## Changes

- Adds better error messaging for when trying to do `Astro.cookies.set()` and `Astro.redirect()` on a response that has already been sent.
- Closes https://github.com/withastro/astro/issues/6632
- Closes https://github.com/withastro/astro/issues/6644

## Testing

- Unit test added for cookies.
- SSR test added for redirects

## Docs

N/A, error message

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -900,6 +900,18 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		hint: 'See https://docs.astro.build/en/guides/content-collections/ for more on the `slug` field.',
 	},
 
+	/**
+	 * @docs
+	 * @message The response cannot be altered after it has already been sent.
+	 * @description
+	 * Making changes to the response, such as setting headers, cookies, and the status code cannot be done outside of page components.
+	 */
+	ResponseSentError: {
+		title: 'Unable to set response',
+		code: 9004,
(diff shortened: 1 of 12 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -900,6 +900,18 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 		hint: 'See https://docs.astro.build/en/guides/content-collections/ for more on the `slug` field.',
 	},
 
+	/**
+	 * @docs
+	 * @message The response cannot be altered after it has already been sent.
```
Princesseuh, line 905: No need for this property here, this is only if the actual message is too complex

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -900,6 +900,18 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * @description
+	 * Making changes to the response, such as setting headers, cookies, and the status code cannot be done outside of page components.
+	 */
+	ResponseSentError: {
+		title: 'Unable to set response',
+		code: 9004,
```
Princesseuh, line 911: 9xxx are for content collections, you want this in the 3xxx section. It needs to be "physically" moved up there as well so it appears in the right order in the docs

sarah11918 (approved):
Docs-approved error message! :partying_face:

PR #6721, opened 2023-03-31
feat(astro): experimental middleware

ematipico:

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -906,6 +906,29 @@ export interface AstroUserConfig {
 	/**
 	 * @docs
 	 * @kind heading
+	 * @name Integrations
+	 * @description
+	 *
+	 * Extend Astro with custom integrations. Integrations are your one-stop-shop for adding framework support (like Solid.js), new features (like sitemaps), and new libraries (like Partytown and Turbolinks).
+	 *
+	 * Read our [Integrations Guide](https://docs.astro.build/en/guides/integrations-guide/) for help getting started with Astro Integrations.
+	 *
+	 * ```js
+	 * import react from '@astrojs/react';
+	 * import tailwind from '@astrojs/tailwind';
+	 * {
+	 *   // Example: Add React + Tailwind support to Astro
+	 *   integrations: [react(), tailwind()]
+	 * }
+	 * ```
+	 */
+	middlewares?: Array<
+		AstroIntegration | (AstroIntegration | false | undefined | null)[] | false | undefined | null
+	>;
+
+	/**
+	 * @docs
+	 * @kind heading
 	 * @name Vite
 	 * @description
 	 *
@@ -1034,6 +1057,9 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
 	// This is a more detailed type than zod validation gives us.
 	// TypeScript still confirms zod validation matches this type.
 	integrations: AstroIntegration[];
+
+	// Order of the middlewares
+	middlewareOrder: string[];
 }
 
 export type ContentEntryModule = {
@@ -1328,6 +1354,7 @@ export interface Page<T = any> {
 export type PaginateFunction = (data: any[], args?: PaginateOptions) => GetStaticPathsResult;
 
 export type Params = Record<string, string | undefined>;
+export type Locals = Record<string, string | undefined>;
 
 export interface AstroAdapter {
 	name: string;
@@ -1369,6 +1396,16 @@ interface AstroSharedContext<Props extends Record<string, any> = Record<string, 
 	 * Redirect to another page (**SSR Only**).
 	 */
 	redirect(path: string, status?: 301 | 302 | 303 | 307 | 308): Response;
+
+	/**
+	 * TODO documentation
+	 */
+	locals: Locals;
+
+	/**
+	 * TODO documentation
+	 */
+	setLocal: (key: string, value: any) => void;
 }
 
 export interface APIContext<Props extends Record<string, any> = Record<string, any>>
@@ -1442,6 +1479,11 @@ export interface APIContext<Props extends Record<string, any> = Record<string, a
 	 * [context reference](https://docs.astro.build/en/guides/api-reference/#contextredirect)
 	 */
 	redirect: AstroSharedContext['redirect'];
+
+	/**
+	 * TODO documentation
+	 */
+	locals: AstroSharedContext['locals'];
 }
 
 export type Props = Record<string, unknown>;
@@ -1482,7 +1524,7 @@ export interface SSRLoadedRenderer extends AstroRenderer {
 	};
 }
 
-export type HookParameters<
+export type HookIntegrationParameters<
 	Hook extends keyof AstroIntegration['hooks'],
 	Fn = AstroIntegration['hooks'][Hook]
 > = Fn extends (...args: any) => any ? Parameters<Fn>[0] : never;
@@ -1529,6 +1571,21 @@ export interface AstroIntegration {
 		}) => void | Promise<void>;
 	};
 }
+
+export type MiddlewareResolve = (context: APIContext) => Promise<Response>;
+
+export type OnBeforeRequestHook = (context: APIContext, resolve: Resolve) => Promise<Response>;
+
+export type Resolve = (context: APIContext) => Promise<Response>;
+
+export type OnAfterRequestHook = (
+	context: Readonly<APIContext>,
+	response: Response
+) => Promise<Response>;
+
+export type AstroMiddlewareInstance = {
+	onRequest?: OnBeforeRequestHook;
+};
 
 export interface AstroPluginOptions {
 	settings: AstroSettings;
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -616,6 +616,37 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 			}`,
 		hint: 'This is often caused by a typo in the image path. Please make sure the file exists, and is spelled correctly.',
 	},
+
+	/**
+	 * TODO [PLT-101] documentation
+	 */
+	MiddlewareNotFound: {
+		title: 'Middleware not found.',
+		code: 3030,
+		message: (middlewareName: string, middlewarePath: string) =>
+			`Can't find the middleware ${middlewareName} at path ${middlewarePath}. Make sure the file exists and has the correct name.`,
+	},
+
+	/**
+	 * TODO [PLT-101] documentation
+	 */
+	LocalsNotAvailable: {
+		title: '`Astro.locals` is not available in current adapter.',
+		code: 3030,
+		message: (adapterName: string) =>
+			`\`Astro.locals\` is not available in the "${adapterName}" adapter. File an issue with the adapter to add support.`,
+	},
+
+	/**
+	 * TODO [PLT-101] documentation
+	 */
+	LocalsNotSerializable: {
+		title: '`Astro.locals` are not serializable.',
+		code: 3031,
+		message: (href: string) => {
+			return `Information stored in \`Astro.locals\` are not serializable when visiting "${href}" path. Make sure you store only data that are compatible.`;
+		},
+	},
 	// No headings here, that way Vite errors are merged with Astro ones in the docs, which makes more sense to users.
 	// Vite Errors - 4xxx
 	/**
--- a/packages/astro/src/core/render/core.ts
+++ b/packages/astro/src/core/render/core.ts
@@ -1,4 +1,11 @@
-import type { ComponentInstance, Params, Props, RouteData } from '../../@types/astro';
+import type {
+	APIContext,
+	ComponentInstance,
+	OnBeforeRequestHook,
+	Params,
+	Props,
+	RouteData,
+} from '../../@types/astro';
 import type { LogOptions } from '../logger/core.js';
 import type { RenderContext } from './context.js';
 import type { Environment } from './environment.js';
@@ -9,6 +16,7 @@ import { AstroError, AstroErrorData } from '../errors/index.js';
 import { getParams } from '../routing/params.js';
 import { createResult } from './result.js';
 import { callGetStaticPaths, findPathItemByKey, RouteCache } from './route-cache.js';
+import { isValueSerializable } from '../util.js';
 
 interface GetParamsAndPropsOptions {
 	mod: ComponentInstance;
@@ -82,7 +90,12 @@ export async function getParamsAndProps(
 	return [params, pageProps];
 }
 
-export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env: Environment) {
+export async function renderPage(
+	mod: ComponentInstance,
+	ctx: RenderContext,
+	env: Environment,
+	apiContext?: APIContext
+) {
 	const paramsAndPropsRes = await getParamsAndProps({
 		logging: env.logging,
 		mod,
@@ -108,6 +121,16 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 	if (!Component)
 		throw new Error(`Expected an exported Astro component but received typeof ${typeof Component}`);
 
+	let locals = {};
+	if (apiContext) {
+		if (!isValueSerializable(apiContext.locals)) {
+			throw new AstroError({
+				...AstroErrorData.LocalsNotSerializable,
+				message: AstroErrorData.LocalsNotSerializable.message(ctx.pathname),
+			});
+		}
+		locals = apiContext.locals;
+	}
 	const result = createResult({
 		adapterName: env.adapterName,
 		links: ctx.links,
@@ -127,6 +150,7 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 		scripts: ctx.scripts,
 		ssr: env.ssr,
 		status: ctx.status ?? 200,
+		locals,
 	});
 
 	// Support `export const components` for `MDX` pages
@@ -134,7 +158,7 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
 		Object.assign(pageProps, { components: (mod as any).components });
 	}
 
-	const response = await runtimeRenderPage(
+	let response = await runtimeRenderPage(
 		result,
 		Component,
 		pageProps,
--- a/.changeset/pretty-bears-deliver.md
+++ b/.changeset/pretty-bears-deliver.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+New middleware API
--- a/packages/astro/src/core/config/schema.ts
+++ b/packages/astro/src/core/config/schema.ts
@@ -29,6 +29,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 		open: false,
 	},
 	integrations: [],
+	middlewares: [],
 	markdown: {
 		drafts: false,
 		...markdownConfigDefaults,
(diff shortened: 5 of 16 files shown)


ematipico:
!preview middleware

ematipico:
!preview middleware

ematipico:
!preview middleware

ematipico:
!preview middleware

ematipico:
!preview middleware

packages/astro/src/@types/astro.ts
```diff
@@ -1685,3 +1732,7 @@ export type CreatePreviewServer = (
 export interface PreviewModule {
 	default: CreatePreviewServer;
 }
+
+export declare namespace AstroMiddleware {
```
ematipico, line 1736: I tried this approach for the interface expansion inside a pet project, but unfortunately, that doesn't work :(

Any help would be appreciated!

packages/astro/src/core/middleware/sequence.ts
```diff
@@ -0,0 +1,43 @@
...
+							return applyHandleResult;
+						} else {
+							return next();
+						}
+					} else {
+						return next();
+					}
+				} else {
+					return next();
+				}
```
ematipico, line 38: Differently from SvelteKit, we have different requirements. Our middleware can return nothing, call `next()` and still work. So we have to do checks whether a middleware returns `void` or not.

packages/astro/src/@types/astro.ts
ematipico: Fixed it!

packages/astro/package.json
```diff
@@ -64,6 +64,10 @@
     "./zod": {
       "types": "./zod.d.ts",
       "default": "./zod.mjs"
+    },
+    "./middleware": {
+      "default": "./middleware.mjs",
+      "types": ""
     }
```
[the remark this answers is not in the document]
ematipico: There are two reasons why I created a new file: 
- I copied the same pattern we have for `astro/config` and `astro/zod`;
- I wanted to export only **a subset of APIs** to user-land: `sequence` and `defineMiddleware`. The other functions are for internal usage. 

If had to apply your suggestion, where should I define the **internal** functions?

packages/astro/src/@types/astro.ts
```diff
@@ -1487,6 +1513,31 @@ export interface APIContext<Props extends Record<string, any> = Record<string, a
...
+	 * ```ts
+	 * // src/middleware.ts
+	 * import {defineMiddleware} from "astro/middleware";
+	 *
+	 * export const onRequest = defineMiddleware((context, next) => {
+	 *   context.locals.greeting = "Hello!";
+	 *   next();
+	 * });
+	 * ```
+	 * Inside the an astro file:
```
sarah11918, line 1531:
```suggestion
	 * Inside a `.astro` file:
```
We keep debating whether people say "a dot astro file" or "an astro file" even when written `.astro`. :sweat_smile:  We think most people pronounce the dot, so we tend to use this.

packages/astro/src/@types/astro.ts
```diff
@@ -1487,6 +1513,31 @@ export interface APIContext<Props extends Record<string, any> = Record<string, a
...
+	 * ```astro
+	 * ---
+	 * // index.astro
```
sarah11918, line 1534:
```suggestion
	 * // src/pages/index.astro
```
Whenever possible and reasonable, we show full file paths. Since we can reasonably infer it here, I'd suggest this.

sarah11918 (commented):
Whew, @ematipico!  :raised_hands: 

I know there are more docs in my future, and I know where my Issue is for me to start tackling them... but I just snagged a couple of tiny things here!

lilnasy:
Could a middleware tell if it's running in dev mode?

The use case I'm thinking of is a middleare that caches responses. It should not act during dev.

lilnasy:
Could a middleware communicate with an integration?

The use case I'm thinking of is an integration that hashes inline resources, and a middleare that uses the hashes to create a strict `Content-Security-Policy` header.

I understand letting middleware be added by integrations was made out of scope for now, but simply receiving data from an independent integration would be alone be valuable.

packages/astro/package.json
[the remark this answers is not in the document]
bluwy: Maybe we can create a `src/core/middleware/_entry.ts` file to define the entrypoints? It _is_ a new convention, but it'll reduce maintenance for the `.mjs` and `.d.ts` file.

packages/astro/src/@types/astro.ts
```diff
@@ -1527,7 +1578,7 @@ export interface SSRLoadedRenderer extends AstroRenderer {
 	};
 }
 
-export type HookParameters<
```
matthewp, line 1530: This rename would be a breaking change. Can you add this back as an alias (maybe with a TODO to remove in 3.0)?

```ts
export type HookParameters = HookIntegrationParameters;
```

ematipico: Good catch! I will revert the change, it's an old code

packages/astro/src/core/app/index.ts
```diff
@@ -201,8 +213,46 @@ export class App {
...
+			const paramsAndPropsResp = await getParamsAndProps({
+				mod: mod as any,
+				route: renderContext.route,
+				routeCache: this.#env.routeCache,
+				pathname: renderContext.pathname,
+				logging: this.#env.logging,
+				ssr: this.#env.ssr,
+			});
+
+			if (paramsAndPropsResp === GetParamsAndPropsError.NoMatchingStaticPath) {
```
matthewp, line 225: We don't want this error because static paths are a SSG only thing. `getStaticPaths` is not called in SSR. 

Actually I'm not sure why you added `getParamsAndProps` call in the first place. Is it needed for some reason? How were params being populated before this?

ematipico: This is just code that I copied from somewhere else to create `APIContext`. I am happy to remove the error if you think it's unnecessary.

packages/astro/src/core/app/index.ts
```diff
@@ -232,7 +282,13 @@ export class App {
 			status,
 		});
 
-		const result = await callEndpoint(handler, this.#env, ctx, this.#logging);
+		const result = await callEndpoint(
+			handler,
+			this.#env,
+			ctx,
+			this.#logging,
+			this.#manifest.middleware as AstroMiddlewareInstance<Response | EndpointOutput>
```
matthewp, line 290: Isn't middleware going to be `undefined` if the feature is not enabled?

packages/astro/src/core/app/index.ts
ematipico: > Actually I'm not sure why you added getParamsAndProps call in the first place. Is it needed for some reason? How were params being populated before this?

In order to create the APIContext, I needed props and params. Should I just pass empty objects?

matthewp: No, we do need those. I should want to make sure we're not calling this code more than once. Where is `params` being populated normally, before this change? I guess you are moving this code up the tree and removing it from the previous place?

The error is needed in SSG mode but not in SSR. This module is SSR-only (currently).

packages/astro/src/core/endpoint/dev/index.ts
```diff
@@ -18,5 +26,11 @@ export async function call(options: SSROptions, logging: LogOptions) {
 		route: options.route,
 	});
 
-	return await callEndpoint(endpointHandler, env, ctx, logging);
+	return await callEndpoint(
+		endpointHandler,
+		env,
+		ctx,
+		logging,
+		middleware as AstroMiddlewareInstance<Response | EndpointOutput>
```
matthewp, line 34: Could we avoid the cast if SSROptions.middleware was typed to this? Currently it is typed as `AstroMiddlewareInstance<unknown>`.

packages/astro/src/core/render/core.ts
```diff
@@ -153,3 +168,57 @@ export async function renderPage(mod: ComponentInstance, ctx: RenderContext, env
...
+
+/**
+ * Checks whether any value can is serializable.
+ *
+ * A serializable value contains plain values. For example, `Proxy`, `Set`, `Map`, functions, etc.
+ * are not serializable objects.
+ *
+ * @param object
+ */
+export function isValueSerializable(value: unknown): boolean {
```
matthewp, line 180: these two functions might be good to put into the util module `src/core/util.ts`.

ematipico: They were in that module fore, but unfortunately `utils` imports a function from the `vite` module, which will trigger rollup errors in some parts of our application.

packages/astro/src/core/app/index.ts
ematipico: I solved the issue without casting

packages/astro/src/core/render/core.ts
```diff
@@ -84,65 +109,74 @@ export async function getParamsAndProps(
...
+	params: Params;
+};
 
+export async function renderPage({
+	mod,
+	renderContext,
+	env,
+	apiContext,
+	params,
+	props,
```
matthewp, line 129: can you check and see if it would be possible for `params` and `props` to go on the `renderContext` object? That way there are fewer things that need to be passed to ever function. Conceptually RenderContext is for things specific to a single render, which params and props fits.

I think that renderContext is created *before* params and props, but they could be initialized as empty objects and then be replaced when params and props are gotten.

packages/astro/src/core/render/core.ts
matthewp: Ok, that makes sense, disregard this comment.

packages/astro/src/core/render/core.ts
ematipico: It seemed that `props` and `params` **can** be created when the `renderContext` is created. We just needed the `mod` and the `env`, which are information we already have and are needed for the rendering phase.

My last commit shows the refactor. There's still one more opportunity for another refactor, but we can do it later.

matthewp (approved):

ematipico:
@sarah11918 could you double-check again if the docs are correct, please?

packages/astro/package.json
```diff
@@ -81,6 +85,8 @@
     "config.mjs",
     "zod.d.ts",
     "zod.mjs",
+    "middleware.d.ts",
+    "middleware.mjs",
     "env.d.ts",
```
bluwy, line 90: We can remove this with the updated export setup

packages/astro/src/core/middleware/sequence.ts
```diff
@@ -0,0 +1,37 @@
+import type { APIContext, MiddlewareResponseHandler } from '../../@types/astro';
+import { defineMiddleware } from './index.js';
+import { AstroError, AstroErrorData } from '../errors/index.js';
+
+/**
+ * From SvelteKit: https://github.com/sveltejs/kit/blob/master/packages/kit/src/exports/hooks/sequence.js
+ *
+ * It accepts one or more middleware handlers and makes sure that they are run in sequence.
+ */
+export function sequence(...handlers: MiddlewareResponseHandler[]): MiddlewareResponseHandler {
```
bluwy, line 10: This JSDoc will be exposed to users, are we fine linking it to SvelteKit? 😅

packages/astro/src/core/build/static-build.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from '../../core/build/internal.js';
 import { emptyDir, removeEmptyDirs } from '../../core/fs/index.js';
 import { appendForwardSlash, prependForwardSlash } from '../../core/path.js';
-import { isModeServerWithNoAdapter } from '../../core/util.js';
+import { isModeServerWithNoAdapter, viteID } from '../../core/util.js';
```
bluwy, line 16: Looks like we can revert the changes in this PR as the added imports are unused.

There's also a few places that have unused imports. Running `pnpm format:ci` fixes it (a bit slow), but it's not a big deal if we don't fix it as CI will auto-format and remove it later. But it would help keep the git history clean.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -628,6 +628,94 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 *   return "string"
+	 * });
+	 * ```
+	 */
+	MiddlewareNotAResponse: {
+		title: 'The middleware returned something that is not a `Response`',
+		code: 3032,
+		message:
+			'When returning something from a middleware, you must return a valid `Response`. Failing doing so would cause errors..',
+	},
```
bluwy, line 670: Maybe it can be improved in the future, but it would be nice to know what is returned that is not a response, that's causing this error.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -628,6 +628,94 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * 	// doesn't return anything or doesn't call next
+	 * 	context.locals.someData = false;
+	 * });
+	 * ```
+	 */
+	MiddlewareNoDataOrNextCalled: {
+		title: "The middleware didn't return a response or called `next`",
+		code: 3031,
+		message: 'The middleware needs to return a `Response` object or call the `next` function.',
+	},
```
sarah11918, line 650:
```suggestion
	 * Thrown when the middleware does not return any data or call the `next` function.
	 *
	 * For example:
	 * ```ts
	 * import {defineMiddleware} from "astro/middleware";
	 * export const onRequest = defineMiddleware((context, _) => {
	 * 	// doesn't return anything or call `next`
	 * 	context.locals.someData = false;
	 * });
	 * ```
	 */
	MiddlewareNoDataOrNextCalled: {
		title: "The middleware didn't return a response or call `next`",
		code: 3031,
		message: 'The middleware needs to either return a `Response` object or call the `next` function.',
	},
```
I'm not sure how we want to refer to "a piece of middleware" but "a middleware" doesn't sound quite right.  :smile:

I've suggested "the middleware" but if there's a more descriptive word like "a middleware request" or "a middleware call" to describe the action, or one that describes the thing like "a middleware service", those would be places we could use "a".

This is a little difficult to word because these are two different things that both need to not happen, and they're not easy to describe together using the same words.  :sweat_smile: But I think this is clear for now?

packages/astro/src/core/errors/errors-data.ts
sarah11918, line 670:
```suggestion
	 * Thrown in development mode when middleware returns something that is not a `Response` object.
	 *
	 * For example:
	 * ```ts
	 * import {defineMiddleware} from "astro/middleware";
	 * export const onRequest = defineMiddleware(() => {
	 *   return "string"
	 * });
	 * ```
	 */
	MiddlewareNotAResponse: {
		title: 'The middleware returned something that is not a `Response` object',
		code: 3032,
		message:
			'Any data returned from middleware must be a valid `Response` object.',
	},
```
Would this be accurate to say? I feel like we don't have to say that NOT doing the thing you have to do will cause errors. I think we can just say, "You have to do this!" and errors are otherwise implied?  :smile:

packages/astro/src/core/middleware/sequence.ts
bluwy: I agree that giving credits is good too, but maybe at the bottom of the jsdoc or at the astro docs, otherwise it gives the vibe of "we re-implement SvelteKit's feature, go check out their docs".

packages/astro/src/core/errors/errors-data.ts
bluwy: Using `JSON.stringify(obj, null, 2)` could help with that, or maybe a `typeof obj`, would give a bit of hint too. But I don't think this is necessarily needed for now so we can look into it again.

bluwy (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -628,6 +628,94 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 *   return next();
+	 * });
+	 * ```
+	 */
+	LocalsNotAnObject: {
+		title: 'Value assigned to `locals` is not accepted.',
+		code: 3033,
+		message: `The \`locals\` can only be assigned to an object. Other values like numbers, strings, etc. are not accepted.`,
+		hint: 'If you tried to remove some information from the `locals` object, try to use `delete` or set the property to `undefined`.',
+	},
```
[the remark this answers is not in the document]
ematipico: >  might consider locals singular, even though the word is plural? If it's one locals object, then I think singular fits? So I'd say "locals is..."
Yes, I would consider `locals` like a name, other than something that we quantify. 

> locals could be "overriden" or could be "overwritten" depending on what's actually happening. I just want to make sure that "override" is what you're going for, and you didn't actually prefer or mean "overWRITE". (I think of override in the sense of controls, or what's going to take precedence. But, I would probably think "overwrite" if the object contents are in fact being changed/rewritten.)

I mean "override". The user can overwrite the contents of `locals` as they want. The user is free. But they can't do something like `locals = "new string"` (override). That is forbidden and will throw this very error.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -628,6 +628,94 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 *   };
+	 *   return next();
+	 * });
+	 * ```
+	 */
+	LocalsNotSerializable: {
+		title: '`Astro.locals` are not serializable.',
+		code: 3034,
+		message: (href: string) => {
+			return `The information stored in \`Astro.locals\` are not serializable when visiting "${href}" path.\nMake sure you store only data that are serializable.`;
```
sarah11918, line 717:
```suggestion
			return `The information stored in \`Astro.locals\` is not serializable when visiting "${href}" path.\nMake sure you store only serializable data.`;
```
Information is singular, and I'm not sure about "when visiting... path" If that renders to e.g. "when visiting /about/ path" then that sentence wouldn't read very well in English.

What exactly is the signifiance of including the part about the path? Would the warning work without it, or is the message depending on knowing which path is causing the error? In that case, instead of "when visiting the path" what is the actual problem? The route can't be created? The path doesn't exist? 

> The information stored in Astro.locals for the path "${href} is not serializable. Mke sure...

Would something closer to that be possible?

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -628,6 +628,94 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 *     foo() {
+	 *       alert("Hello world!")
+	 *     }
+	 *   };
+	 *   return next();
+	 * });
+	 * ```
+	 */
+	LocalsNotSerializable: {
+		title: '`Astro.locals` are not serializable.',
```
sarah11918, line 714:
```suggestion
	 * Thrown in development mode when a user attempts to store something that is not serializable in `local`.
	 *
	 * For example:
	 * ```ts
	 * import {defineMiddleware} from "astro/middleware";
	 * export const onRequest = defineMiddleware((context, next) => {
	 *   context.locals = {
	 *     foo() {
	 *       alert("Hello world!")
	 *     }
	 *   };
	 *   return next();
	 * });
	 * ```
	 */
	LocalsNotSerializable: {
		title: '`Astro.locals` is not serializable',
```
Just the same little things for consistency with the other docs e.g. no period for the title, no comma after "thrown in development mode"

packages/astro/src/core/errors/errors-data.ts
ematipico: > What exactly is the signifiance of including the part about the path?

The logic of the middleware can change based on the page the user is visiting. So having a path can help the developer understand the issue.

> > The information stored in Astro.locals for the path "${href} is not serializable. Make sure...
> 
> Would something closer to that be possible?

That's perfect!

JLarky:
@ematipico you have a typo in description `MiddlewareHanlder` instead of `MiddlewareHandler` (that's why we use TS BTW 😅)

Issue #6695, opened 2023-03-28
@astrojs/tailwind ignores Tailwind CSS v3.3 TypeScript config


phenomen:
### What version of `astro` are you using?

2.1.8

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### Describe the Bug

When I use Tailwind CSS v3.3 TypeScript config file `tailwind.config.ts`, Astro ignores it and uses default config instead.

**Reference:** https://tailwindcss.com/blog/tailwindcss-v3-3#esm-and-typescript-support

I think you just need to bump Tailwind dependency to use updated `resolveConfig`.

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Thanks @phenomen, would accept a PR if you wanted to fix this.

alf5:
it looks like tailwind 3.3 is already picked up by @astrojs/tailwind (either new project or pnpm upgrade)
but fails to load tailwind config file
Maybe this is as easy as getting @proload/core to use its typescript plugin - @proload/plugin-typescript ?

Princesseuh:
I wonder if using proload is the best solution for loading the config. I'd expect Tailwind to provide something for that

juliusmarminge:
> I wonder if using proload is the best solution for loading the config. I'd expect Tailwind to provide something for that

They have. `tailwindcss/loadConfig`: https://github.com/tailwindlabs/tailwindcss/blob/master/src/lib/load-config.ts

We just shipped support for this in prettier-plugin-tailwindcss: https://github.com/tailwindlabs/prettier-plugin-tailwindcss/pull/137#issuecomment-1488545934

FinnDore:
Gonna try make a pr now

FinnDore:
So ive got it kind of working. Just struggling to get custom colors to load in the example for some reason 
<img width="797" alt="image" src="https://user-images.githubusercontent.com/34718806/228627260-25de5e0d-eb6a-4761-9252-aad915819273.png">

FinnDore:
ah it was simply a questionable skill issue, for 34 minuites the idea that `bg` needs to be applied to `t3-purple-100` for it to change the bg color had escaped my mind.... we have a working TS config!

<img width="1510" alt="image" src="https://user-images.githubusercontent.com/34718806/228635184-14925be2-6c8d-4728-98bc-4bc96d489394.png">

[...]

PR #6724, opened 2023-04-01
@astrojs/tailwind: simplify, upgrade & fix support for ts config file

TomPichaud:
First participation to Astro! Love your project and looking forward to its evolution!

Fix: https://github.com/withastro/astro/issues/6695

## Changes

Commit https://github.com/withastro/astro/pull/6724/commits/8a8be4221edb136af505d17f0f20ec1d341f6a1a : add support for ESM/TS config files while maintaining current behaviour: providing a default tailwind config for astro if none is present.

Knowing that Astro can already suggest and generate a default tailwind config trough the command `astro add tailwind`, I think we can simplify this integration further, reducing its dependencies and making it more robust.

In light of the above, I've made the following commit https://github.com/withastro/astro/pull/6724/commits/4a515e3dc402a705c8ffcbe35a839904c2f6e190 : only setup postcss, supporting custom config file path and the base css import

Also, I feel like any integration setup should be inside the integration package and not hard coded into astro to be able to use `astro add ...`.

## Testing

Tested locally with a `tailwind.config.ts` and custom `custom-tailwind.config.ts`

## Docs

No further documentation is needed, but we could add ESM/TS support


TomPichaud:
Commit https://github.com/withastro/astro/pull/6724/commits/8a8be4221edb136af505d17f0f20ec1d341f6a1a add support for ESM/TS config files while maintaining current behaviour: providing a default tailwind config file for astro

Commit https://github.com/withastro/astro/pull/6724/commits/4a515e3dc402a705c8ffcbe35a839904c2f6e190 simplify this integration by only setting up postcss, supporting custom config file path and the base css import

.changeset/real-spies-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/tailwind': major
```
matthewp, line 2: Mistake?

matthewp:
Hey @TomPichaud are you still working on this change? Thanks!

TomPichaud:
I think I'm done. Did you read all of my commit ? I was hoping for some review or discussion around this but I'm happy with my work. Maybe someone can review and approve ?

I'm not sure about the version as I've made some changes that affect users with no tailwind config, if any ¿? But like I said, Astro is already pushing the user to have a tailwind config file and can do it through the CLI.

chav-aniket:
+1 for getting this reviewed asap 🙇🏽

TomPichaud:
@matthewp Any review would be welcome please ? It's working perfectly on all my projects.

helbing:
@TomPichaud Hi guy, are you still dealing with this PR?

TomPichaud:
Just waiting for someone to review and approve...

matthewp:
We usually don't review the code until the PR is ready; meaning merge conflicts are resolved and tests are passing. 

As for the overall theme of the PR, definitely in favor of simplifying! One thing that's important is HMR. Have you verified that this change still allows Tailwind changes to HMR in the client? Have you verified that changing the config triggers HMR as well (the `addWatchFile` removal might have hurt that).

And just to clarify, the main new feature / bug fix of this PR is that TypeScript and ESM configs are now supported, is that correct?

helbing:
I see this PR want to reactor the Tailwind integration and then support the TypeScript config to fix errors. I think we should separate fix things and refactor things. Support the TypeScript config is high priority. We should deal with it first and do refactor later in other PR. How do you think? @TomPichaud @matthewp

helbing:
I think we should separate the fix thing and refactor thing. To support TypeScript config is high priority, we should deal with it first. And then later we can refactor the tailwind integration at other PR. How do you think? @TomPichaud @matthewp

TomPichaud:
Like I said, there's little point doing redundant work. To my opinion it's best to let tailwind load the config and not do it ourself, adding more dependencies, bug possibilities and the necessity to update it when tailwind add a feature. Simple|Less is better, and if we follow tailwind's recommendations we just need to configure postcss + vite, like I'm suggesting we do.

Also they are working on automatic 'content' detection so it will already be ready for that.

So I disagree on the fix first and then refactor because the fix is adding dependencies as the refactoring is removing some dependencies. To get the same result.
Also Astro'doc already recommend and explain how to create a tailwind config file that support astro, and it even do it for you through the command line using tailwind init feature. So we should not overthink it and load a custom config that we edit just for the content section to support astro.

helbing:
> Like I said, there's little point doing redundant work. To my opinion it's best to let tailwind load the config and not do it ourself, adding more dependencies, bug possibilities and the necessity to update it when tailwind add a feature. Simple|Less is better, and if we follow tailwind's recommendations we just need to configure postcss + vite, like I'm suggesting we do.
> 
> Also they are working on automatic 'content' detection so it will already be ready for that.
> 
> So I disagree on the fix first and then refactor because the fix is adding dependencies as the refactoring is removing some dependencies to get the same result. Also Astro'doc already recommend and explain how to create a tailwind config file that support astro, and it even do it for you through the command line using tailwind init feature. So we should not overthink it and load a custom config that we edit just for the content section to support astro.
> 
> I suggest you test this pr thoroughly and merge it. I've got no issues locally.

Yes, I test it locally, it can works. I see the branch has conflicts need be resolved, can you deal with it?

bluwy:
I've rebased the PR and it should be good to go now. I also think it's nice to clean up the config by moving `config.path` and `config.applyBaseStyles` to the root, which that could be made after this is merged.

TomPichaud:
Thanks for your help!

.changeset/real-spies-pretend.md
ematipico: I can't see the history of the file. @TomPichaud is this change **intentional**? Shouldn't this be a `patch` change, considering the PR says  "fix"? 

If this is a `major`, and breaks existing APIs, the changeset should mention a migration path for what's changed.

matthewp:
Let's do a preview release and test it out. In particular we need to verify there are no regressions in terms of HMR, which is something we fixed in the previous implementation.

matthewp:
!preview tailwind-config

matthewp:
For some reason the preview release didn't work. So I manually did a beta release which you can test with:

```shell
npm install @astrojs/tailwind@beta
```

matthewp:
HMR looks good here.

ematipico (approved):
Let's 🚢 it!

.changeset/real-spies-pretend.md
bluwy: I've updated this part with what changed and the migration path.

PR #6748, opened 2023-04-04
chore: use `workspace` where possible

ematipico:
## Changes

There were some fixtures that were using packages from `npm`, instead we should always `workspace:*" whenever possible in our fixtures.

## Testing

Current CI should pass

## Docs

N/A

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -659,12 +659,12 @@ importers:
 
   packages/astro/e2e/fixtures/astro-component:
     specifiers:
-      '@astrojs/preact': ^1.1.0
+      '@astrojs/preact': workspace:*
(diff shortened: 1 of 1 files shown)


Princesseuh:
Would love @matthewp review here because I'm not 100% sure if the packages in the `performance` folder should use `workspace:*`

bluwy (approved):
Looks fine to me. I think these packages are already using the workspace version because we have this config:

https://github.com/withastro/astro/blob/fe646e2fe0c3f49754500afcdafef50b26591602/.npmrc#L2

but it's good to be more explicit. I ran the performance test locally and it seems to run fine too. Except that markdoc benchmark has an error because of it's recent breaking change.

PR #6758, opened 2023-04-05
[Draft][Markdoc] Fix global asset bleed

bholmesdev:
## Changes

- What does this change?

### DIFF
--- a/packages/astro/src/core/render/dev/vite.ts
+++ b/packages/astro/src/core/render/dev/vite.ts
@@ -19,11 +19,14 @@ export async function* crawlGraph(
 	loader: ModuleLoader,
 	_id: string,
 	isRootFile: boolean,
+	stopAtPropagatedAssets: boolean,
 	scanned = new Set<string>()
 ): AsyncGenerator<ModuleNode, void, unknown> {
 	const id = unwrapId(_id);
 	const importedModules = new Set<ModuleNode>();
-	if (new URL(id, 'file://').searchParams.has(PROPAGATED_ASSET_FLAG)) return;
+	if(stopAtPropagatedAssets) {
+		if (new URL(id, 'file://').searchParams.has(PROPAGATED_ASSET_FLAG)) return;
+	}
 
 	const moduleEntriesForId = isRootFile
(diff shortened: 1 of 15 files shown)


packages/astro/src/vite-plugin-head/index.ts
```diff
@@ -9,7 +9,7 @@ import { getTopLevelPages, walkParentInfos } from '../core/build/graph.js';
 import type { BuildInternals } from '../core/build/internal.js';
 import { getAstroMetadata } from '../vite-plugin-astro/index.js';
 
-const injectExp = /^\/\/\s*astro-head-inject/;
+const injectExp = /(^\/\/|\/\/!)\s*astro-head-inject/;
```
bholmesdev, line 12: Maybe worth a comment on what this Regex checks?

delucis (approved):
Seeing a review request, but I don’t _think_ this needs a docs maintainers review? This is mostly invisible to users and just a bug fix right? Nice work, Ben!

packages/astro/src/core/render/dev/vite.ts
```diff
@@ -10,9 +9,10 @@ import { isCSSRequest } from './util.js';
  * List of file extensions signalling we can (and should) SSR ahead-of-time
  * See usage below
  */
-const fileExtensionsToSSR = new Set(['.astro', ...SUPPORTED_MARKDOWN_FILE_EXTENSIONS]);
+const fileExtensionsToSSR = new Set(['.astro', '.mdoc', ...SUPPORTED_MARKDOWN_FILE_EXTENSIONS]);
```
bluwy, line 12: Is there a reason we're hardcoding `.mdoc` here?

matthewp: These are modules that we need to ensure are imported before we do rendering in dev. We need to do this so that we can discover which CSS to include on the page.

matthewp: Yeah, I mean it definitely could. We need a generic solution if we keep going down that path. Maybe using metadata or something like that... probably would rather hold off until the next time to tackle that.

PR #6759, opened 2023-04-05
Upgrade to Vite 4.2

bluwy:
## Changes

Upgrade to Vite 4.2 to use its newer options and remove some hacky code.

## Testing

existing tests should pass

## Docs

n/a. version bump.

### DIFF
--- a/packages/astro/src/core/build/static-build.ts
+++ b/packages/astro/src/core/build/static-build.ts
@@ -148,6 +148,9 @@ async function ssrBuild(
 		logLevel: opts.viteConfig.logLevel ?? 'error',
 		build: {
 			target: 'esnext',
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/build/static-build.ts
```diff
@@ -148,6 +148,9 @@ async function ssrBuild(
 		logLevel: opts.viteConfig.logLevel ?? 'error',
 		build: {
 			target: 'esnext',
+			// Vite defaults cssMinify to false in SSR by default, but we want to minify it
+			// as the CSS generated are used and served to the client.
+			cssMinify: viteConfig.build?.minify == null ? true : !!viteConfig.build?.minify,
```
matthewp, line 153: Ah nice! Glad to see this new option.

PR #6760, opened 2023-04-05
ci: correctly update comment

ematipico:
## Changes

The `Check mergeability` workflow is always run at every commit, and we don't properly check if the bot has already left a message.

This updates in a way that the bot always updates the same message. The code is taken from this example: https://github.com/peter-evans/create-or-update-comment#where-to-find-the-id-of-a-comment

## Testing

Merge and check if it works 

## Docs

N/A


.github/workflows/check-merge.yml
```diff
@@ -62,11 +62,20 @@ jobs:
...
+          issue-number: ${{ github.event.number }}
+          comment-author: 'github-actions[bot]'
+
       - name: Send PR review
         if: steps.find-blockers.outputs.found == 'true'
-        uses: peter-evans/create-or-update-comment@v2
+        uses: peter-evans/create-or-update-comment@v3
         continue-on-error: true
         with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
```
matthewp, line 78: what happens if there isn't a comment yet? Is this like undefined or something?

ematipico: I don't know the answer, but In the repo they have an example where they split create and update, I can use that one

PR #6765, opened 2023-04-05
Fix experimental.assets's Squoosh service not working on Netlify and Vercel SSR

Princesseuh:
## Changes

I have tried a thousand ways to bundle `.wasm` files on SSR hosts and concluded that: There's no way. Every runtime has its own way of doing it (Vercel supports `.wasm?module` imports, Netlify doesn't have an option, who knows what other are doing).

A promising way was to use Vite's WASM imports, but this doesn't work either for multiple reasons, notably those don't work in SSR the way you would expect them to. Both `?init` and `?url` don't work because the `.wasm` files end at the wrong place and for the former, it only supports fetching the wasm file so it's ultimately unsuitable for SSR. Even if that worked, some SSR hosts just won't include the `.wasm` files in their final bundle (ex: Netlify)

So I went the shameful way. I inlined them.

## Testing

Tested manually. See <https://mechanical-magnitude-princesseuh.vercel.app/

## Docs

N/A


Princesseuh:
!preview image-wasm-ssr

ematipico (approved):

PR #6766, opened 2023-04-06
Exporting ImageFunction from astro:content

Xetera:
As of now, if you want to declare your schemas separate from the collection function, some hacks like these are needed:

```ts
type Schema = Parameters<typeof defineCollection>[0]['schema']

const blogSchema: Schema = ({ image }) => z.object({
  title: z.string(),
  date: z.coerce.date(),
  description: z.string(),
  tags: z.array(z.string()),
  draft: z.boolean().default(false),
  banner: image(),
})
```

Doing this might be needed if you want to use the schema in other places so you can export the type of the schema.

## Changes

Exports the `ImageFunction` and gives the object passed to the schema function a name.

## Testing

Just a type change so no tests should be needed.

## Docs

This should be more of a QoL update and types aren't documented in general from what I know.

### DIFF
--- a/.changeset/big-poems-march.md
+++ b/.changeset/big-poems-march.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Exporting the ImageFunction in astro:content and grouping it under a SchemaContext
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
Looks great, thank you!

PR #6771, opened 2023-04-06
Implement scopedStyleStrategy

matthewp:
## Changes

- Adds `scopedStyleStrategy` option to the config.
- Uses this option to pass into the compiler.

## Testing

- Tests that `'where'` is the current default and works as expected.
- Tests that `'class'` omits the where pseudo-selector.

## Docs

- Included in the Astro config.


ematipico (changes requested):
Blocking because it's a minor

matthewp:
!preview scopedStyleStrategy

bluwy (approved):
Changes looks good to me 👍

.changeset/green-cups-hammer.md
```diff
@@ -0,0 +1,21 @@
...
+'astro': minor
+---
+
+Implements a new class-based scoping strategy
+
+This implements the [Scoping RFC](https://github.com/withastro/roadmap/pull/543), providing a way to opt-in to increased style specifity for Astro component styles.
+
+This prevents bugs where global styles override Astro component styles due to CSS ordering and the use of element selectors.
+
+To enable class-based scoping you can set it in your config:
```
ematipico, line 11:
```suggestion
Implements a new class-based scoping strategy

This implements the [Scoping RFC](https://github.com/withastro/roadmap/pull/543), providing a way to opt-in to increased style specificity for Astro component styles.

This prevents bugs where global styles override Astro component styles due to CSS ordering and the use of element selectors.

To enable class-based scoping, you can set it in your config:
```

.changeset/green-cups-hammer.md
```diff
@@ -0,0 +1,21 @@
...
+
+```js
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  scopedStyleStrategy: 'class'
+});
+```
+
+Note that the 0-specifity `:where` pseudo-selector is still used as the default strategy. The intent is to change `'class'` to be the default in 3.0.
```
ematipico, line 21:
```suggestion
Note that the 0-specificity `:where` pseudo-selector is still the default strategy. The intent is to change `'class'` to be the default in 3.0.
```

ematipico (approved):
Looks good to me!

packages/astro/src/@types/astro.ts
```diff
@@ -494,6 +494,22 @@ export interface AstroUserConfig {
...
+	 * @type {('where' | 'class')}
+	 * @default `'where'`
+	 * @description
+	 *
+	 * Specify the strategy used for scoping styles within Astro components. Choose from:
+	 *   - `'where'` - Use `:where` selectors, causing no specifity increase.
+	 *   - `'class'` - Use class-based selectors, causing a +1 specifity increase.
+	 *
+	 * Using `'class'` is helpful when using global stylesheets that set defaults on element selectors, and then wanting to use element selectors to override those global styles within an Astro component.
+	 * Using `'where'` gives you more control over specifity, but requires that you use higher-specifity selectors, layers, and other tools to control which selectors are applied.
```
sarah11918, line 509:
```suggestion
	 * Using `'class'` is helpful when you want to ensure that element selectors within an Astro component override global style defaults (e.g. from a global stylesheet).
	 * Using `'where'` gives you more control over specifity, but requires that you use higher-specifity selectors, layers, and other tools to control which selectors are applied.
```
Would this be accurate? Edit as necessary. I think focusing on the need (i.e. that your style tags in components override global defaults) is both helpful to the reader, and can be expressed in a shorter sentence.

packages/astro/src/@types/astro.ts
```diff
@@ -494,6 +494,22 @@ export interface AstroUserConfig {
 	 */
 	trailingSlash?: 'always' | 'never' | 'ignore';
 
+	/**
+	 * @docs
+	 * @name scopedStyleStrategy
+	 * @type {('where' | 'class')}
+	 * @default `'where'`
+	 * @description
+	 *
```
sarah11918, line 503:
```suggestion
	 * @version 2.3.5
	 *
```
This might be a good candidate for the "Added In" field?

sarah11918 (commented):
Looks good, @matthewp! I left some suggestions to consider below.

Issue #6125, opened 2023-02-04
[@astrojs/image] sizes and background props should be optional on <Picture>


jkjustjoshing:
### What version of `astro` are you using?

2.0.6

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

The `<Picture>` component has 2 required props that don't make sense to me as being required:

### `sizes`

This is a pass-through prop that gets added to the underlying HTML elements. An HTML `<picture>` element doesn't require this attribute, so why does Astro? My understanding is that this attribute is only used for different art direction at different sizes, so requiring it for use cases where the different images are identical except for resolution is unnecessary.

### `background`

The documentation says that "it’s advisable to include a background color, otherwise black will be used as default replacement for transparent pixels". If this prop has a default value ("black"), why am I being made to add this prop, especially if I'm only using JPEGs (with no transparency)?

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/picture-requires-sizes-and-background-3ugksq (see the Typescript error on the `<Picture>` element usage)

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #6773, opened 2023-04-06
Optional `Sizes` prop on Picture component

doganalper:
## Changes

Set `sizes` prop as optional as it is not needed by default on 
fixes https://github.com/withastro/astro/issues/6125

## Testing

N/A

## Docs

N/A


Princesseuh (approved):
Looks good to me, thank you!

Issue #6561, opened 2023-03-15
Layout convention prevents HTML streaming


tropperstyle:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

Node.js

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Curl

### Describe the Bug

HTML streaming pauses immediately after the opening `<body>` tag, even if the async component is deep in the document.

In this example, the server should output `<h1>` before pausing for the `<Async>` component.

The `<h1>` tag _is_ output first if the layout component is not used.

It is not clear from the documentation if this is a bug or an inherent limitation of Astro components

----

layout.astro
```
---
const { title } = Astro.props;
---

<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8" />
        <title>{title}</title>
    </head>

    <body>
        <slot />
    </body>
</html>
```

async.astro
```
---
await new Promise((resolve) => setTimeout(resolve, 5000));
---

<p>Async</p>
```

page.astro
```
---
import Layout from "./layout.astro";
import Async from "./async.astro";
---

<Layout title="Layout Blocks Streaming">
    <h1>Page</h1>
    <Async />
</Layout>
```

$ curl -v http://localhost:3000/page
```
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <title>Layout Blocks Streaming</title>
    </head>

    <body>

        <!-- Pauses here for 5s -->

        <h1>Page</h1>
        <p>Async</p>
    </body>
</html>
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rduhhb-u6t9nb?file=src/pages/index.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
Thanks, I'll be looking into this next week.

bholmesdev:
I ran into this during a tech demo at a meetup the other day. Wasn't sure if this was intended behavior; sounds like not?

matthewp:
@bholmesdev layouts are weird because we render them to a string. I think that's likely what's going on here, we are rendering it to a string before injecting it. I am going to look into whether anything can be done about this. IIRC the primary reason why we render them to a string is for the Astro.slots APIs.

PR #6775, opened 2023-04-06
Support streaming inside of slots

matthewp:
## Changes

- In order to support the `Astro.slots.render('name')` API previously `renderSlot` returned a string. This meant no streaming of slots.
- Refactored that function to `renderSlotToString`. It's still used a bit internally.
- Changed `renderSlot`, which is now just used inside the template (for `<slot></slot>`) to stream.
- Depends on the compiler fix to preserve whitespace and newlines inside slots: https://github.com/withastro/compiler/pull/770
- Closes https://github.com/withastro/astro/issues/6561

## Testing

- Added test case to our existing streaming tests

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/test/streaming.test.js
+++ b/packages/astro/test/streaming.test.js
@@ -9,6 +9,8 @@ describe('Streaming', () => {
 	/** @type {import('./test-utils').Fixture} */
 	let fixture;
 
+	let decoder = new TextDecoder();
+
 	before(async () => {
 		fixture = await loadFixture({
 			root: './fixtures/streaming/',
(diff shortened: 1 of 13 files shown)


natemoo-re (approved):
Makes sense to me, nice fix!

packages/astro/test/streaming.test.js
```diff
@@ -33,11 +35,21 @@ describe('Streaming', () => {
...
 		});
+
+		it('Body of slots is chunked', async () => {
+			let res = await fixture.fetch('/slot');
+			let chunks = [];
+			for await (const bytes of streamAsyncIterator(res.body)) {
+				let chunk = decoder.decode(bytes);
+				chunks.push(chunk);
+			}
+			expect(chunks.length).to.be.greaterThan(2);
```
ematipico, line 51: Is `2` exported the number of slots in the test fixture? Is it because we render two `Wait` components?

ematipico (approved):

packages/astro/test/streaming.test.js
matthewp: 3 is the expected. You get the first initial chunk. Then the chunk after the first wait. Then the chunk after the second wait. I guess I could change this to equal because it should always be 3.

ematipico: It might help. Having a strict equal can prevent possible regressions

matthewp: The reason for avoiding it is that Node buffers writes. If you send 2 chunks with no wait it will combine them into a single write on the network. So the chunks can be less than the number you expect, but never more. In this case, the wait is sufficiently long enough that it should always split into 3 chunks. SO I've updated the test to reflect that.

PR #6776, opened 2023-04-06
Revert vercel env fix

nblackburn:
Apologies, in my efforts to find out why the key wasn't being included, allowed myself to lose track of the fact that only environment variables that are prefixed with PUBLIC_ will be bundled at build time.

## Changes

- Reverts #6751 

## Testing

Didn't have the desired effect, this simply reverts it back to how it was.

## Docs

- Fixes a previously unfixed patch (specifically in the case where it was working before)
- Remains broken for those it was broken for before

### DIFF
--- a/.changeset/shaggy-chefs-build.md
+++ b/.changeset/shaggy-chefs-build.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': patch
+---
+
+Revert change to environment variable
--- a/packages/integrations/vercel/src/analytics.ts
+++ b/packages/integrations/vercel/src/analytics.ts
@@ -39,7 +39,7 @@ const sendToAnalytics = (metric: Metric, options: Options) => {
 };
 
 function webVitals() {
-	const analyticsId = (import.meta as any).env.VERCEL_ANALYTICS_ID;
+	const analyticsId = (import.meta as any).env.PUBLIC_VERCEL_ANALYTICS_ID;
 	if (!analyticsId) {
 		console.error('[Analytics] VERCEL_ANALYTICS_ID not found');
 		return;


bluwy:
Ah right I forgot that private env vars are only replaced when transforming files in SSR, which the analytics script is client only. As a fix, the Vercel adapter could update the Vite config to add `VERCEL_ANALYTICS_ID` to the `envPrefix` array. That should do the trick, considering `VERCEL_ANALYTICS_ID` is public.

matthewp:
@bluwy that's a fix for the initial problem, right? I'm going to assume that we need to merge this first.

bluwy:
The `envPrefix` trick would fix this regression, keeping the `(import.meta as any).env.VERCEL_ANALYTICS_ID;` part. But if we can't get around the fix soon, merging this as a stop gap sounds fine to me too.

bluwy (approved):
Let's merge this for now then, and we can revisit this again.

bluwy:
`@delucis` mentioned yesterday that Vercel seems to also set the `PUBLIC_` env vars on build: https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#framework-environment-variables. So this seems to be a non-issue for now.

The docs doesn't mention `PUBLIC_VERCEL_ANALYTICS_ID` though, but I'd assume it also would. Maybe if it's reproducible that it isn't working, we could check with Vercel if it's correctly set.

onebarloop:
> Maybe we could simply make the variable public by passing it to the client side.
> 
> As a workaround/POC, this works by editing the `.env` file in the Astro project:
> 
> ```
> PUBLIC_VERCEL_ANALYTICS_ID=${VERCEL_ANALYTICS_ID}
> ```
> 
> So maybe it would also be achievable directly in the adapter?
> 
> Otherwise, maybe the "define" could also work for this case. The [Vite documentation](https://vitejs.dev/config/shared-options.html#define) states:
> 
> > If you would like to expose an unprefixed variable, you can use [define](https://vitejs.dev/config/shared-options.html#define) to expose it:
> 
> ```js
> define: {
>   'import.meta.env.ENV_VARIABLE': JSON.stringify(process.env.ENV_VARIABLE)
> }
> ```
> 
> This is simply food for though, as I assume adding a `envPrefix` is also a perfectly viable solution 👍

I can confirm vite-define works for this case

PR #6781, opened 2023-04-07
Fix: `astro:server:setup` middleware

### LINKED ISSUE #6729: Error using Astro and Partytown plugin: 404 Not Found for partytown-sandbox-sw.js and partytown-sw.js

2.1.9

no

npm

MacOS

Chrome, Safari, Firefox

### Describe the Bug

I am encountering an error when using the Astro framework version 2.1.9 and the Partytown plugin. The error message is:

```bash
09:41:25 PM [serve]    404 /~partytown/debug/partytown-sandbox-sw.js
09:41:26 PM [serve]    404        /~partytown/debug/partytown-sw.js
```
I have tested this issue with Partytown versions 1.1.1 and 1.2.0. However, when using Astro version 2.1.7, both versions of Partytown work without any issues.

I am unsure whether this is a problem with Astro or the Partytown plugin.

https://github.com/alexsedeke/astropartytown.git

### Participation

bholmesdev:
## Changes

- Resolves #6729 
- #6693 removed a duplicate `astro:server:setup` call from `create-vite`. Turns out, _this_ is where the setup hook should be run! By only calling from the `container`, we were attaching middlewares after Vite had setup the server. Moving this call to the original plugin with `configureServer` solves the issue.

## Testing

- Add middlewares test to make sure integration API example works https://docs.astro.build/en/reference/integrations-reference/#astroserversetup

### DIFF
--- a/examples/minimal/src/env.d.ts
+++ b/examples/minimal/src/env.d.ts
@@ -1,2 +1,1 @@
-/// <reference path="../.astro/types.d.ts" />
 /// <reference types="astro/client" />
--- a/examples/minimal/src/env.d.ts
+++ b/examples/minimal/src/env.d.ts
@@ -1,1 +1,2 @@
+/// <reference path="../.astro/types.d.ts" />
 /// <reference types="astro/client" />
--- a/.changeset/wet-chefs-love.md
+++ b/.changeset/wet-chefs-love.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix `astro:server:setup` middlewares not applying. This resolves an issue with the Partytown integration in dev.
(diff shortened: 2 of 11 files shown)


examples/minimal/src/env.d.ts
```diff
@@ -1 +1,2 @@
+/// <reference path="../.astro/types.d.ts" />
```
ematipico, line 1: Is the change intended?

ematipico (approved):

examples/minimal/src/env.d.ts
bholmesdev: Nope! Reverting

PR #6782, opened 2023-04-07
fix(error-overlay): force error overlay direction to be LTR

amirhhashemi:
## Changes

The error overlay is designed for LTR (Left to Right) documents so when I change my document's direction to RTL (`<html dir="rtl"`) it breaks. I forced the direction to be LTR because it doesn't make sense to be RTL. The error message and the codes would be hard to read in RTL.

Before:

Screenshot from 2023-04-07 18-21-51

After:

Screenshot from 2023-04-07 18-22-38

## Testing

Not needed

## Docs

Not needed

### DIFF
--- a/.changeset/beige-humans-study.md
+++ b/.changeset/beige-humans-study.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Force error overlay direction to be LTR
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
This looks great. Sorry for not considering this when designed the overlay, we really should have.

PR #6791, opened 2023-04-08
Render async SolidJS components

patdx:
## Changes

- Initially discussed in https://github.com/withastro/roadmap/discussions/519
- Changes the server renderer of Solid components from renderToString to renderToStringAsync.
- This allows you to use *async* APIs inside of your Solid components, such as getImage() from Astro Image or remote APIs.
- The renderer will generate the official Solid hydration script together with this first eligible component of the page. This is required to handle possible data emissions related to resources resolving or error boundaries triggering.

@johannesspohr also contributed code and tests to this PR.

## Testing

I designed some test scenarios and wrote some tests conforming the existing/nonexistence of hydration scripts depending on the client directive.

## Docs

This PR changes how server-only and hydrating SolidJS components may resolve async resources. Before, they may have resolved them on the client side. Now, they may be resolved on the server side before rendering. I tried to cover the changes in the changeset and the SolidJS integration readme.


sarah11918:
I'm going to tag @Jutanium here as the Docs Maintainer to decide whether this would require an update to the Solid integration README if accepted.

packages/astro/src/@types/astro.ts
```diff
@@ -1617,6 +1617,7 @@ export interface SSRMetadata {
 	renderers: SSRLoadedRenderer[];
 	pathname: string;
 	hasHydrationScript: boolean;
+	hasSolidHydrationScript: boolean;
```
matthewp, line 1620: Rather than have framework-specific properties in this internal Result object, this should happen within the renderer itself. You can do this by using the Result object as a weakmap key. We do this in some of the integrations already. See for example in the Preact integration: https://github.com/withastro/astro/blob/main/packages/integrations/preact/src/server.ts#L47

Jutanium:
I would want to edit the SolidJS Integration Readme if this gets accepted

packages/astro/src/@types/astro.ts
patdx: Thank you. I found the equivalent `getContext` in the Solid integration and changed it to use that. In the process I discovered another issue with the PR so I have temporarily changed it to "draft" status until I solve that issue.

patdx:
Thank you for the initial feedback. I tried to address the comment about [how to track page context](https://github.com/withastro/astro/pull/6791#discussion_r1164546343). In the process I noticed a new bug in the implementation [where it is still generating "hydration events/patches"  even though `client::none` should be totally static](https://github.com/withastro/astro/blob/689b2367800300dc1fb838ee67617ad89b2e6adc/packages/astro/test/solid-component.test.js#L47). I have temporarily changed the PR to draft until I have solved that issue.

matthewp:
Would fix https://github.com/withastro/astro/issues/6912

matthewp:
Awesome @patdx can you update to the latest Solid version in that case?

patdx:
Hi @matthewp I updated all the references to the latest SolidJS version, 1.7.4, at the time of writing.

I also added a test based on the #6912 reproduction example.

#6912 lists three possible outcomes:

> 1. (bad) The component fails to render completely, usually on first startup only. The lazy component is not in the source HTML.
> 2. (bad) The component flashes for a second before hydration, and then after hydration occurs, it disappears. Usually this happens on the second refresh and much more frequently in my more complex production app where hydration takes more time.
> 3. (good) The component successfully renders itself, usually after several refreshes. The component is in the source HTML.

As far as I can tell (1) seems to be resolved no matter how times I refresh in dev or preview mode. Most of the time I got the successful result (3), however I could still still observe some issue like (2) that randomly occurred when refreshing. So I believe this PR may improve the behavior of #6912 but not fix it entirely.

sarah11918 (changes requested):
Just blocking this one for docs so it doesn't slip past me. If accepted, this has potential breaking changes for existing users and will require an update to the Solid integration README to document this behaviour. 

@patdx - When we are confident in this being accepted, and the user-facing behaviour is stable enough to know what we need to explain to readers (behind the scenes implementation details don't need to be documented, so it doesn't matter to docs *how* you get this behaviour working! :smile:), please take an initial stab at documenting the new behaviour in the README (finding the logical place, adding the behaviour) and @Yan-Thomas and I will review!

patdx:
I made a super quick attempt to rebase the PR with the latest version.

Depending on the implementation of the new [parallelized rendering](https://astro.build/blog/astro-250/), it may be necessary adjust the Server render logic. If there are hydratable Solid components on the page, the official [SolidJS hydration script](https://www.solidjs.com/guides/server#hydration-script) needs to be inserted as an inline script on the page before the first hydratable Solid component in the HTML.

Also, I didn't realize but I guess it is possible to have Astro Solid islands rendered inside of Astro Solid islands?

```html
<!-- index.astro -->
<SolidA>
  <SolidB />
</SolidA>
<SolidC/>
```
If SolidB may be rendered before SolidA then I think I can't just rely on "which render function was called first", need to use extraHead or something to inject the hydration script.

natemoo-re:
Hey @patdx thanks for tackling this! Sorry things have been quiet! Are you still interested in working on this? If so, we'd be happy to help get this back on track.

patdx:
> Hey @patdx thanks for tackling this! Sorry things have been quiet! Are you still interested in working on this? If so, we'd be happy to help get this back on track.

Hi @natemoo-re sorry yes I am still interested in working on this, but I only have a few hours per week max to work on this kind of stuff, and I ran out of energy to keep up recently, I guess I am a little re-energized now. But still, if somebody is interested to take over I am happy to step back :)

I tried to rebase and resolve [my own open question](https://github.com/withastro/astro/pull/6791#issuecomment-1553913189) about nested components. I did some testing and found it actually worked mostly correctly by default. But I was able to come up with a pretty contrived situation where the inner SolidJS component was rendered first, which caused the SolidJS hydration script to be rendered in the wrong spot (or not at all). So I moved the SolidJS hydration script to `extraHead` and added a new test to validate that the hydration script comes before any hydration events in the output HTML. Does this approach sound okay? This change is tracked in this commit: https://github.com/withastro/astro/pull/6791/commits/1bb4a7d802d3f6db75416bc74b5bda9a508a86a3

patdx:
I updated the readme to clarify that that Suspense is only automatically wrapped around hydrating components, not non-hydrating components (`client:only` components). 

I set it up this way because I feel like it is a good default to make the renderToStringAsync on the server wait for `lazy()` and `createResource()` to resolve, and a Suspense boundary seems necessary to make this happen.

However, if we automatically wrapped Suspense around a client:only component that used `createResource()` API or similar, since there is no server rendering, the default blank/null Suspense fallback would definitely render. That doesn't seem like a great default. In conclusion, if you want to use these Suspense-y APIs for client:only components, I think you probably would want to add your own custom Suspense component and fallback anyway, so there is no special benefit to adding it ourselves.

howagain:
Hey @patdx wanted to bump this tread because I'd love to see this get merged in. Also, follow-up questions but what would be required to get this to work with solid's streaming. Astro already streams its page down so would that even be possible or would that break Astro's current component model?

patdx:
> Hey @patdx wanted to bump this tread because I'd love to see this get merged in.

@howagain I did a quick rebase, hmm I think the PR is basically feature complete, with all the tests I can think of. I need help from the Astro team to review the technical strategy and documentation when they have some time.

> Also, follow-up questions but what would be required to get this to work with solid's streaming. Astro already streams its page down so would that even be possible or would that break Astro's current component model?

Hmm, I'm not sure how that would work. I don't understand the full rendering story, but I think that Astro expects the Renderers (React, Solid, etc) to provide a function called [rendertoStaticMarkup](https://github.com/search?q=repo%3Awithastro%2Fastro%20renderToStaticMarkup&type=code) that returns `Promise<string>`. First of all Astro would need to add support for a way to return a streaming value back instead, and Renderers could implement that streaming API.

There would still be some other issues to overcome:

1. I'm not sure how you would support streaming with components nested inside components, especially if different frameworks. Although many frameworks support streaming output, I don't think many (or any?) frameworks support streaming arbitrary HTML inside of them so nesting components may deopt streaming.
2. Through my testing, it seemed possible in Astro for a nested component to start "rendering" before its parent started rendering component. SolidJS also needs a [Hydration Script](https://www.solidjs.com/guides/server#hydration-script) injected in the HTML before the first top-level SolidJS component rendered on the page. I solved this problem in this PR using [extraHead](https://github.com/withastro/astro/commit/1bb4a7d802d3f6db75416bc74b5bda9a508a86a3) to make sure the Hydration Script got injected before the first SolidJS component regardless of which one started rendering first. But if Astro started streaming everything, I think we would have to assume that the `<head>...</head>` was already rendered and streamed to the client before rendering started for the SolidJS components. So we could no longer use `extraHead`. We would need a better way to detect the first top level SolidJS component in the page and inject the Hydration Script before that, regardless of actual rendering order.

howagain:
@sarah11918 what all is required to get this merged in? I'd love to start taking advantage of solid's `lazy` to help with code splitting our larger solid components.

sarah11918:
Hi @howagain!

I know it's my block you're seeing, but that's because this PR would cause breaking changes and we want to make sure nothing goes out without appropriate guidance in documentation. Normally, when a feature is approved and the implementation is confirmed by the dev team, then the docs team would be pinged to oversee documentation. (Not every PR we receive is accepted and approved.)

The maintainer who would normally oversee this documentation is no longer with the project, but I'll put this one on @ElianCodes ' radar. We would look at it once @natemoo-re or @matthewp tells us it's likely to be approved! So, I guess docs is the bottleneck, but at the same time is not the bottleneck here.  :smile:

natemoo-re:
Sorry for the delay @patdx @howagain! We've been heads-down working on Astro 3.0. This would be a great addition to our revamped Solid adapter to go along with that major version. Hoping we can circle back to this soon!

patdx:
I rebased my code on the latest changes again.

I am having trouble figure out the source but I noticed that according to the tests this file:

```
packages/astro/test/fixtures/slots-solid/src/pages/mdx.mdx
```

is not rendering correctly, the output after `astro build` is just:

```
<!DOCTYPE html>undefined
```

Need to figure out why this is happening...

packages/integrations/solid/src/server.ts
```diff
@@ -21,27 +43,91 @@ function renderToStaticMarkup(
...
+			// *before* the first `await` in the function, so the hydration script will
+			// be prefixed to the first hydratable component on the page, regardless of
+			// the order in which the components finish rendering.
+
+			this.result._metadata.extraHead.push(generateHydrationScript());
+			ctx.hasHydrationScript = true;
+		}
+	}
+
+	html +=
```
johannesspohr, line 126: This is causing the test failures, as `undefined` return from `renderToString` will be stringified here and cause the resulting html to be a string and not `undefined`.

patdx: Thank you! I tried changing this code and it seemed to help with some tests. I still had a few unresolved issues. I also found that the Solid renderer `check()` function also erroneously returns true for MDX page components, overriding the MDX page component rendering and causing the output of a string "undefined". I found at least a basic fix for that. Finally there seem to be some test failures in `packages/astro/test/astro-slots-nested.test.js`, still investigating that. I hope to push these changes soon.

johannesspohr: I removed the `let html = '';` a couple of lines before, and replaced the `html += ` with `const html =`, and all tests pass for me. I don't know if the issues you are talking about are covered by test cases or if I'm missing something, but the error you're talking about sounds exactly like the issue this fixes.
Thanks for your efforts btw, this feature would really elevate the solid integration to a new level :-)

patdx: Got it, thanks. I had tried rebasing on the latest commit in `main` branch at the time, so maybe that had somehow caused a difference in our results.

Originally I just added `?? ''` to handle the "undefined" issue, but I was still getting the `astro-slots-nested ` error for some reason. I rebased again and tried more like your style of using `=` instead of `+=`.   In the last test, the astro-slots-nested issue seemed to resolve itself.

I do feel like the renderer detection logic is not quite working correctly (eg, picking the wrong renderer in some cases), but I can't reproduce it reliably...and out of scope for this task I suppose.

johannesspohr: If I look at the current implementation I still have some questions.
The `check` function relies on the render function returning **not** a string for `html` in case it can't be rendered (see `server.tsx` line 48). So, we need to make sure that we pass any `undefined` from `renderToString` as is, and not convert it to a string. However, by doing `?? ''` or concatenating (like now) will convert it into a string (`"undefined"`), and that will make the `check` function falsely assume that anything is a solid component. As said, by that one change (not concatenating `html` in any case, but just assigning it), my tests were passing consistently.

patdx: Hmm, good point. I wrote that comment after this commit:  https://github.com/withastro/astro/commit/0837b54450392cf7cd246d3d19a3e179bb099878

Where I passed the html value "untouched", just assigning it:

```ts
let html: string;

if (renderStrategy === "async") {
  html = await renderToStringAsync(renderFn, { renderId });
} else {
  html = renderToString(renderFn, { renderId });
}

return {
  attrs: {
    "data-solid-render-id": renderId,
  },
  html,
};
```

That commit seemed to pass the unit test suite, but after I pushed it I realized that the e2e test suite was failing: https://github.com/withastro/astro/actions/runs/6486448279/job/17614736586

Upon further investigation, it seems that in some of the e2e tests, the [Solid Hydration Script](https://www.solidjs.com/guides/server#hydration-script) failed to be injected into the page before Solid elements, causing some of the e2e tests to have an initialization error and just time out.

johannesspohr: I had a proper look at the hydration script issue and some things make more sense to me now, namely:
- The `check` call is a render "dry run", just to find out if we could render this component with solid. It relies on the result `html` being a string if successful, `undefined` if not. Only if the render was successful will the render function be called again for the "real" render - because we're passing different `renderStrategies`, we can use this to determine whether we are in a dry run right now (we could consider renaming it).
- The check for the mdx extentions only hides the fact that because of this concatenation, the check is not properly working. So we still need to make sure to return `undefined` if the dry run (sync) fails.
- In SSR, the head is always already sent, so just using `extraHeads` doesn't work - we need to print to HTML as you also did.

So the solution I came up with looks like this (can surely be optimized and include the extraHead thing again in case it's not yet rendered):
```typescript
        /*
	 * The first component to render needs to also render the hydration script. 
	 * Since components render in parallel, we already indicate to print the hydration
	 * script, as soon as we know it will be printed here. 
	 */
	const includeHydrationScript =
		!ctx.hasHydrationScript && needsHydrate && renderStrategy === 'async';
	if (includeHydrationScript) {
		ctx.hasHydrationScript = true;
	}

	let html =
		renderStrategy === 'async'
			? await renderToStringAsync(renderFn, { renderId })
			: renderToString(renderFn, { renderId });
        // we can assume that with includeHydrationScript = renderStrategy async it always works, because the dry run returned proper HTML
	if (includeHydrationScript) {
		html = generateHydrationScript() + html;
	}
```

There's one more caveat: Since components render in parallel, it might be that the order of rendering is not actually the order that the components appear in. Imagine two astro components
```
<Comp1 />
<Comp2 />
```
where both are delayed (fetching sth. from the database for example), but Comp2 is ready earlier. If both render a solid component, the hydration script will actually be printed *only* before Comp2. Luckily, still both components hydrate successfully, if I read that hydration script correctly it's only that events (click, input) on the first one wouldn't be caught and handled later. From my understanding, with the current architecture, it's not possible to find out from the render function which component is the first on the page, so we would need to live with that.

patdx: Thank you, I think (or hope) I finally get it about the undefined/string thing. I see now that is how the code in main branch is working too. I should try to keep this part unchanged to help keep this PR focused.

Oh, interesting, so you mean SSR mode as opposed to SSG/static rendering mode. In this SSR mode extraHead does not work?

Unfortunately the placement of the hydration script is a little more important than just buffering some client/iinput events. It also sets up a data store at `_$HY` used to pass serialized data from the server to the client, In particular, I've found that Suspense, createResource and ErrorBoundary tend to use it. If the `_$HY` object is not initialized first it may cause errors or break hydration, like this:

image

But yeah, there does not seem to be an obvious/simple way to take care of that.

johannesspohr: Good points. I just tested putting some `createResource` in my scenario with the two delayed components and, as you predicted, it threw (`_$HY` is undefined). It still worked somehow, but I guess we shouldn't rely on that.

One thing I noticed is that Astro manages to put the prescripts (the Astro hydration logic) before the actual first component that uses a directive like `client:load` - as far as I can see there is no way to hook in from a renderer to inject the solid hydration script. But that could probably be achieved somehow.
Also, we might just add this small script to the head section everytime when the solid integration is activated, also not ideal but probably the easiest solution.

sarah11918:
Hey @patdx ! I see more activity on this! 🥳 

Can you let me know whether the changeset/README are likely to change, or ready to be reviewed? As I wrote before, in docs we don't really care *how* you get it working, we only care about how someone will use the feature.

So that means if the usage instructions are correct advice to the user (whenever all the technical details are worked out, even if they're not quite yet now) then we can put this on our radar again to review!  🙌

johannesspohr:
I looked into the `RenderInstructions` and found that would be the best way to implement it. It needs some minor additions in the core but feels quite clean. This fixes the delay issue and fix the slots problem as well.
https://github.com/johannesspohr/astro/commit/fc6e08d3446145f86b16812dd9e2bfc65de3e81d
(don't know if I could push this to your branch, but you might wanna have a look first anyway)

patdx:
> I looked into the `RenderInstructions` and found that would be the best way to implement it. It needs some minor additions in the core but feels quite clean. This fixes the delay issue and fix the slots problem as well. [johannesspohr@fc6e08d](https://github.com/johannesspohr/astro/commit/fc6e08d3446145f86b16812dd9e2bfc65de3e81d) (don't know if I could push this to your branch, but you might wanna have a look first anyway)

Thank you @johannesspohr. I liked your solution so I cherry picked your commit into this PR, hope that is okay. I do wonder if it will be acceptable to modify the core like this. Let's see if there is any additional feedback about this approach from the core team.

> Hey @patdx ! I see more activity on this! 🥳
> 
> Can you let me know whether the changeset/README are likely to change, or ready to be reviewed? As I wrote before, in docs we don't really care _how_ you get it working, we only care about how someone will use the feature.
> 
> So that means if the usage instructions are correct advice to the user (whenever all the technical details are worked out, even if they're not quite yet now) then we can put this on our radar again to review! 🙌

Hi @sarah11918 it's pretty complete. Even if there are more changes after technical review I don't think they affect the docs very much. I took another look and can't think what else to say, hopefully it is on the right track. Would greatly appreciate some feedback on the documentation. Thank you!

sarah11918:
Thanks, just a heads up this is on my list for this week (reviewing the docs!)

packages/integrations/solid/README.md
```diff
@@ -98,6 +98,73 @@ export default defineConfig({
 });
 ```
 
+## Rendering strategy
+
+Hydrating SolidJS components are automatically wrapped in Suspense boundaries and rendered on the server using the [renderToStringAsync](https://www.solidjs.com/docs/latest/api#rendertostringasync) function. This means that lazy components will be resolved and rendered on the server. For example the following source files:
```
sarah11918, line 103: In general, we prefer not to have multiple code samples just thrown at the reader with no commentary, and then at the very end ask them to make a connection about what they saw. We do try to guide the reader by priming them for each component they're about to see.

Here, you're not guiding the reader at all as to what to even look for, so they have to read every. single. thing. super closely to see what's important and what is actually happening. Is the name of the file important? (turns out it is!) Is the fact that the name `Hello` in the function important? (Turns out it's not!) 

These should really be telling the reader what they are *going to see*, so that they already know the _what_ before they get there, and they can just focus on the _how_.

So, can you try to do that here, something like (and I'm not really familiar with Solid, so this is just an example of how to guide the reader. It's your job to make these statements accurate!)

Before you do that for this first example, though, please read my comment for the second example. I suspect something similar is happening here, where there might be more code samples than are needed to simply convey that "you can lazily import other components, and they'll be available." I found this super long and complicated to get that idea across, and no priming for what to look for. The second example is a bit more concrete, and I think whatever strategy you end up using there will help this example.

packages/integrations/solid/README.md
```diff
@@ -98,6 +98,73 @@ export default defineConfig({
...
+<LazyHelloAstro />
+```
+
+Will be rendered into the server's HTML output as
+
+```html
+<div>Hello Astro</div>
+```
+
+Resources will also be resolved. For example:
```
sarah11918, line 132: ... continuing from the idea in my earlier comment,

For example, in this case, I might suggest something like:

> Resources will also be resolved. For example, a function that awaits a data fetch will be able to render the returned information:

or 

> Resources will also be resolved. The following example shows a function that awaits a data fetch, and then renders the returned information:

(I'm not familiar with Solid, so it's your job to make sure the information is correct. But, we do like to pre-load our explanations so that people know what they should be looking at/for when you put a code sample in front of them.)

I would also argue that the three snippets here are more distracting than helpful. It takes work to figure out what is going on when I think it's kind of assumed that if you write a Solid component, you'll render in a `.astro` file somewhere, and then it will render to HTML. It's probably ok to completely skip the middle code here, because I think what you're really trying to say is "a component written like this will render to this HTML". 

In fact, if *all* you were trying to say is that it would successfully return the data from the fetch, I would argue that you wouldn't even need the final HTML snippet at all. I think it's pretty clear from the function alone that you're *trying* to return `<div>Name: [some actual name]</div>` , so if the key point is just "this works" then I don't think we need three whole code snippets to say that.

.changeset/chilly-badgers-push.md
```diff
@@ -0,0 +1,29 @@
+---
+'@astrojs/solid-js': major
+---
+
+Render SolidJS components using [renderToStringAsync](https://www.solidjs.com/docs/latest#rendertostringasync).
```
sarah11918, line 5:
```suggestion
Render SolidJS components using [`renderToStringAsync`](https://www.solidjs.com/docs/latest#rendertostringasync).
```
Just checking, because I don't see these formatted as code anywhere, but seems like something that should be? If so, I'd update throughout.

.changeset/chilly-badgers-push.md
```diff
@@ -0,0 +1,29 @@
...
+This increases the flexibility of the SolidJS integration. Server-only SolidJS components may now call async functions directly using the `createResource` API, like loading data from another API or using async Astro server functions like `getImage()`. It is very unlikely that a server only component would have used the Suspense feature until now, so this should not be a breaking change for server-only components.
+
+This could be a breaking change for components that meet the following conditions:
+
+- The component uses Suspense APIs like `Suspense`, `lazy` or `createResource`, and
+- The component is mounted using a *hydrating* directive:
+  - `client:load`
+  - `client:idle`
+  - `client:visible`
+  - `client:media`
```
sarah11918, line 22:
```suggestion
This increases the flexibility of the SolidJS integration. Server-only SolidJS components may now call async functions directly using the `createResource` API, like loading data from another API or using async Astro server functions like `getImage()`.

It is very unlikely that a server only component would have used the Suspense feature until now, so this should not be a breaking change for server-only components.

However, this could be a breaking change for components that meet the following conditions:

- The component uses Suspense APIs like `Suspense`, `lazy` or `createResource`, and
- The component is mounted using a *hydrating* directive:
  - `client:load`
  - `client:idle`
  - `client:visible`
  - `client:media`
```
I always like to call out more attention when we're talking about breaking changes, and so that first sentence probably shouldn't be tacked on to the end of an unrelated paragraph. The "however" gives that extra bit of warning, and doesn't make it sound like you're contradicting yourself! :smile:

packages/integrations/solid/README.md
```diff
@@ -98,6 +98,73 @@ export default defineConfig({
...
+
+Will generate the following HTML output:
+
+```html
+<div>Name: Luke Skywalker</div>
+```
+
+### Wrapping in Suspense
+
+Server-only or hydrating components are automatically wrapped in top level Suspense boundaries, so it is not necessary to add a top level Suspense boundary around async components.
```
sarah11918, line 162:
```suggestion
Server-only or hydrating components are automatically wrapped in top-level Suspense boundaries, so it is not necessary to add a top-level Suspense boundary around async components.
```
nit

sarah11918 (commented):
Hi @patdx ! Thanks for your patience with docs! This looks like an amazing feature, and I'm sure it will be a welcome update to the Solid integration!

The changeset was *really* nicely written! I appreciate your attention to breaking changes, and I felt like there was good context provided for a major upgrade! :raised_hands: 

I made some suggestions re: tackling this topic in the README. I think my main point is self-explanatory, (it's the same for both examples) but feel free to ask if you have any questions about it!

packages/integrations/solid/README.md
patdx: Addressed in https://github.com/withastro/astro/pull/6791/commits/5a810a9b73463d93b3b7eeee92ea1dd1ca9f7111

.changeset/chilly-badgers-push.md
patdx: Addressed in https://github.com/withastro/astro/pull/6791/commits/5a810a9b73463d93b3b7eeee92ea1dd1ca9f7111

packages/integrations/solid/README.md
patdx: Wow, thanks for deep feedback. :pray: 

I cut down the examples a lot and decided to focus on `createResource`, as that seems easier to explain and understand.

I left one line in about `lazy`. Maybe it's less critical to explain to the end user, but I think this PR should fix the lazy-related bug #6912, so I feel like it should at least get a small mention.

Addressed in: https://github.com/withastro/astro/pull/6791/commits/5a810a9b73463d93b3b7eeee92ea1dd1ca9f7111

packages/integrations/solid/README.md
patdx: Addressed in https://github.com/withastro/astro/pull/6791/commits/5a810a9b73463d93b3b7eeee92ea1dd1ca9f7111

patrick-entinux:
I made an update to address the documentation feedback.

I also changed the version of Solid JS everywhere to 1.8.5 This is for two reasons:

1. It seems that at some point last week or earlier, the resolved version of Solid (eg, the version shown in pnpm-lock.yml)  in this workspace changed from 1.7.x to 1.8.3. This messed up some of the tests, which were checking for signatures of the hydration script which has changed. Since the tests now formally depend on v1.8.x, I thought I should formally update the versions specified in package-lock.json files too.
2. There was also a broken test where, even in server-only component mode, it was generating some inline hydration scripts. I found that I could resolve this by upgrading Solid JS to 1.8.5 and making use of the new `noScripts: true` option: https://github.com/solidjs/solid/issues/1931

patdx:
(Sorry, I accidentally posted from the wrong account, so I have deleted and am reposting the same message)

I made an update to address the documentation feedback.

I also changed the version of Solid JS everywhere to 1.8.5 This is for two reasons:

1. It seems that at some point last week or earlier, the resolved version of Solid (eg, the version shown in pnpm-lock.yml)  in this workspace changed from 1.7.x to 1.8.3. This messed up some of the tests, which were checking for signatures of the hydration script which has changed. Since the tests now formally depend on v1.8.x, I thought I should formally update the versions specified in package-lock.json files too.
2. There was also a broken test where, even in server-only component mode, it was generating some inline hydration scripts. I found that I could resolve this by upgrading Solid JS to 1.8.5 and making use of the new `noScripts: true` option: https://github.com/solidjs/solid/issues/1931

sarah11918 (approved):
Thanks so much for addressing the documentation feedback! I think what you've got is super clear, and this seems like a great benefit to our SolidJS users!  :partying_face:

florian-lefebvre:
What's the status of this PR apart from the conflicts? AFAIK docs is fine, what about implementation itself?

patdx:
@florian-lefebvre The code is completed and passing the tests.

I think the next step is to get a technical review from Astro core team. I'm not sure who would be in charge. Maybe @natemoo-re or @matthewp could help take a look?

While the code is complete to the best of my knowledge, we ended up introducing a new `renderHydrationScript()` mechanic to Astro's core. The purpose was to solve the issue of getting Solid's hydration script injected on the page before the first Solid component (even if components are nested, etc). I think we need to make sure to get the team's approval or feedback on this part of the change especially.

I just resolved the conflicts again. Looks like the Solid JS version got bumped to 1.8.5 in main branch of Astro last week, which makes this PR smaller :)

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -375,6 +375,15 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 				}
 			}
 			destination.write(createRenderInstruction({ type: 'directive', hydration }));
+			if (hydration.directive !== 'only' && renderer?.ssr.renderHydrationScript) {
+				destination.write(
+					createRenderInstruction({
+						type: 'renderer-hydration',
+						rendererName: renderer.name,
+						render: renderer?.ssr.renderHydrationScript,
```
ematipico, line 383: I don't think the optional chaining operator `?` is needed, because `render` is a mandatory property

packages/astro/src/runtime/server/render/common.ts
```diff
@@ -89,6 +89,13 @@ function stringifyChunk(
 				}
 				return renderAllHeadContent(result);
 			}
+			case 'renderer-hydration': {
+				if (!result._metadata.hasRendererSpecificHydrationScript[instruction.rendererName]) {
```
ematipico, line 93: I don't understand this check, why do we check against `false` values? Shouldn't we **remove** the `if` block?

matthewp (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -2480,6 +2485,10 @@ export interface SSRResult {
  */
 export interface SSRMetadata {
 	hasHydrationScript: boolean;
+	/**
+	 * Keep track of which renderers already injected their specific hydration script
```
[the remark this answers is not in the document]
patdx: Noted. I am editing the comment. Also, I'm going to change it to a `Set<string>` as this matches the pattern of `hasDirectives` and `propagators` which might make it easier to understand.

The naming is tricky. The name `hydrationScripts` could almost work, however it may be too confusing next to the unrelated `hasHydrationScript`. I think that's why @johannesspohr, who helped write this part of the PR, originally added `RendererSpecific`, to make it more clearly different from `hasHydrationScript`.

Thinking of it as a `Set`, I'm planning to revise it slightly to `rendererSpecificHydrationScripts`. This lets we rewrite the code in `packages/astro/src/runtime/server/render/common.ts` in a slightly easier to read way:

```ts
const { rendererSpecificHydrationScripts } = result._metadata;
const { rendererName } = instruction;

if (!rendererSpecificHydrationScripts.has(rendererName)) {
  rendererSpecificHydrationScripts.add(rendererName);
  return instruction.render();
}
return '';
```

packages/astro/src/runtime/server/render/common.ts
patdx: The [Solid hydration script](https://www.solidjs.com/guides/server#hydration-script) only needs to be rendered once per page, so this is a check to make sure it has not been rendered yet. If we removed the `if`, it would get rendered once per component instead, which would probably still work, but would not be ideal as it would be sending some redundant data to the client.

packages/astro/src/runtime/server/render/instruction.ts
```diff
@@ -11,18 +11,32 @@ export type RenderHeadInstruction = {
 	type: 'head';
 };
 
+/**
+ * Render a renderer-specific hydration script before the first component of that
+ * framework
+ */
+export type RendererHydrationInstruction = {
+	type: 'renderer-hydration';
```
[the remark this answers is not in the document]
patdx: You're right, the thing it's actually doing is rendering a hydration *script*, not doing the actual hydration. (Hydration happens after.) However, I feel that `hydration-script` would be confusingly similar to the unrelated `hasHydrationScript` thing. For the benefit of future developers I think it would be good to keep these clearly separated. I will rename it to `renderer-hydration-script` for now. Please take a look again after I push the next commit and let me know what you think?

packages/integrations/solid/src/client.ts
```diff
@@ -1,14 +1,12 @@
+import { Suspense } from 'solid-js';
 import { createComponent, hydrate, render } from 'solid-js/web';
 
 export default (element: HTMLElement) =>
 	(Component: any, props: any, slotted: any, { client }: { client: string }) => {
-		// Prepare global object expected by Solid's hydration logic
-		if (!(window as any)._$HY) {
-			(window as any)._$HY = { events: [], completed: new WeakSet(), r: {} };
-		}
```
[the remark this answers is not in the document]
patdx: This code has been moved to the `renderHydrationScript` directive thing and now using the [official hydration script](https://www.solidjs.com/guides/server#hydration-script) generated by Solid's `generateHydrationScript()` function.

Background: The hydration script creates a global `_$HY` object. During Solid's SSR phase, it may serialize data and inject it into inline scripts on the page that depend on the existence of the `_$HY` object. These hydration event scripts look similar to the snippet below:

```html
<script>(self.$R=self.$R||{})["s0"]=[]_$HY.r["s00-2-0"]="$$f"`</script>
```

Hydration events are uncommon in sync SSR, but very common in async SSR (especially when using createResource, Suspense or ErrorBoundary). This file `packages/integrations/solid/src/client.ts` loads too late to handle any actual hydration event scripts. Therefore it becomes necessary to inject the hydration script into HTML directly, before any of the hydration event scripts.

packages/integrations/solid/src/context.ts
```diff
@@ -11,10 +11,13 @@ export function getContext(result: RendererContext['result']): Context {
...
 		return contexts.get(result)!;
 	}
-	let ctx = {
+	let ctx: Context = {
 		c: 0,
 		get id() {
-			return 's' + this.c.toString();
+			// a hyphen at the end makes it easier to distinguish the island
+			// render id from component tree portion of the hydration key `data-hk="..."`
+			// https://github.com/solidjs/solid-start/blob/f0860887030e0632949b3f497e279aecb6ed5afd/packages/start/islands/mount.tsx#L41
```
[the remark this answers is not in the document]
patdx: I'll just remove it from the PR because I don't want the PR to get distracted by this. I changed this because it made debugging hydration errors a little bit easier and forgot about it, but now that you mention it, I realize that technically it's not related or essential. Maybe me or someone else can make a PR for this some other time. :)

patdx:
I rebased the PR and added a new commit to address @ematipico's feedback. Thanks!

matthewp (approved):

packages/integrations/solid/README.md
```diff
@@ -98,6 +98,41 @@ export default defineConfig({
 });
 ```
 
+## Async rendering strategy
```
sarah11918, line 101: @patdx Making a note that this change should now be PR'd to the docs repo directly!

We have updated all the integration READMEs to a minimal format, so instead, please make a PR in the docs repo for this content change at: https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/integrations-guide/solid-js.mdx and do not edit the Solid-js README. Thank you!

patdx: Got it, I made a PR for the docs here: https://github.com/withastro/docs/pull/6027

patdx:
I merged and resolved conflicts in the last commit, including reverting the readme changes.

I created a separate PR for the Solid docs change here: https://github.com/withastro/docs/pull/6027

packages/integrations/solid/README.md
sarah11918: Thanks, got it! We'll pick this part of the conversation up over there! :partying_face:

Issue #6058, opened 2023-01-31
Astro 2.0 warning when using sourcemaps


JLarky:
### What version of `astro` are you using?

2.0.4

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

Starting from 2.0 running `astro build` causes this warning when using sourcemaps. I haven't tested it thoroughly, but it seems like this is just a warning and sourcemaps for my code are still there.

Warning:
```
09:45:56 PM [build] Building static entrypoints...
Sourcemap is likely to be incorrect: a plugin (astro:scanner) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
Sourcemap is likely to be incorrect: a plugin (astro:scanner) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
09:45:56 PM [build] Completed in 328ms.
```

stackblitz is just a fork of default astro template where I enabled sourcemaps and ran `build`:

<img width="674" alt="image" src="https://user-images.githubusercontent.com/7026/215668089-7877375b-98ca-45d7-8363-1d71be0129a2.png">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ytzf7d?file=astro.config.mjs&startScript=build

### Participation

- [ ] I am willing to submit a pull request for this issue.

JLarky:
hey, didn't mean to rain on your parade, but :)

while the simplest possible case (the stackblizt link that I created) was fixed, I still see those errors in my production website

<img width="1783" alt="image" src="https://user-images.githubusercontent.com/7026/218286517-bb5dc352-5ac5-4283-8172-b8a2c7fc298c.png">

here's an example of latest (2.0.10) build that produces the same warnings https://github.com/JLarky/qgp.app/actions/runs/4153799688/jobs/7185669325

<img width="1046" alt="image" src="https://user-images.githubusercontent.com/7026/218286627-32fb4d7a-34bf-4157-8603-790d373b22b6.png">

bluwy:
Seems to be caused by https://github.com/withastro/astro/blob/8b7cb64dadfca93c65d62df54754633d398cb2ed/packages/astro/src/content/vite-plugin-content-imports.ts#L130

That might need magic-string to replace and generate a new sourcemap.

wulinsheng123:
mark

PR #6808, opened 2023-04-10
Remove unnecessary transform for content imports

bluwy:
## Changes

In `vite-plugin-content-imports`, there's a transform to escape Vite references. However, in the load hook we already did that, so we should be able to remove the transform hook.

Fix https://github.com/withastro/astro/issues/6058 (there was a sourcemap warning in the transform hook)

## Testing

Existing tests should pass.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/ninety-frogs-impress.md
+++ b/.changeset/ninety-frogs-impress.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Remove unnecessary escape transform for content imports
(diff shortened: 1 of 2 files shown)


bluwy:
Looks like this break docs 👀

bholmesdev:
Thanks @bluwy! Was actually refactoring this while working on data collections, and I found a better fix: **don't use `load` at all.** Always use the `transform` hook to handle flags like this to avoid Vite post-processing for certain files. I was finding that Vite's JSON transform was super aggressive even on flag imports and virtual modules. Moving everything to a transform, I was able to avoid this _and_ keep our tests passing. Think I'll make a code refactor PR so we don't need to wait on data collections for the change.

bluwy:
Awesome! That makes sense to me, and thanks for chiming in. I'll close this for now then so we can fix it in a different way.

Issue #6629, opened 2023-03-22
2.1.5 regression: `astro check` fails with “`astro sync` command failed to generate content collection types”


andersk:
### What version of `astro` are you using?

2.1.5

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

`astro check` fails in a project using collections. It worked in 2.1.4.

This can be reproduced in the standard blog example (https://astro.new/blog).

```console
❯ astro check
01:33:17 PM [check] Checking files
 error   `astro sync` command failed to generate content collection types.
  Hint:
    Check your `src/content/config.*` file for typos.
  File:
    /home/projects/aonipylmz.github/node_modules/astro/dist/core/errors/errors.js:26:5
  Code:
    25 |   }
    > 26 |   setLocation(location) {
         |     ^
      27 |     this.loc = location;
      28 |   }
      29 |   setName(name) {
  Stacktrace:
Error
    at AstroError (file:///home/projects/aonipylmz.github/node_modules/astro/dist/core/errors/errors.js:26:5)
    at sync (file:///home/projects/aonipylmz.github/node_modules/astro/dist/core/sync/index.js:83:11)
```

### Link to Minimal Reproducible Example

https://astro.new/blog

### Participation

- [ ] I am willing to submit a pull request for this issue.

bholmesdev:
Confirmed this was introduced by #6617, which resolved a _different_ urgent issue 😓 

Tagging @bluwy and @MoustaphaDev in case they have context on this change. Seems to be an issue resolving the `astro:content` virtual module.

PR #6811, opened 2023-04-10
Fix check CLI fs load fallback behaviour

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6629

In `vite-plugin-load-fallback`, the `fs` check wasn't right when the `fs` is from `import * as fs from 'fs'`. This PR fixes it. 

I also simplified the `resolveId` hook as the `this.resolve` and always `slashify` part doesn't seem necessary. Other plugins should've already resolve before the `load-fallback` plugin, and always `slashify` makes it impossible for a plugin to run `this.resolve` to check if an id is resolveable (e.g. `tsconfig-alias` plugin `baseUrl`)

## Testing

Existing tests should pass. The edgecase in #6629 is a little hard to test, but once I do some tsconfig aliases soon, the regression shouldn't happen again.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/angry-experts-compete.md
+++ b/.changeset/angry-experts-compete.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix check CLI fs load fallback behaviour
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
I don't completely understand the code here (I'm not 100% sure why sync needs to be passed an instance of fs, I assume it's for tests?), but nonetheless the code make sense and the tests pass, so it looks good to me

matthewp (approved):

bluwy:
Yeah sync is passed with fs so it can be unit-tested to swap out the fs

PR #6813, opened 2023-04-10
Update issue template with link

ElianCodes:
## Changes

Updates note on *minimal example* with link to creating minimal examples on the astro docs: https://docs.astro.build/en/guides/troubleshooting/#creating-minimal-reproductions

## Testing

Not really tested

## Docs

No docs needed, since this is a ref to docs


bluwy (approved):
Nice guide!

Issue #6551, opened 2023-03-15
Questions about style import of scss+svelte+astro


mydracula:
### What version of `astro` are you using?

latest 

### Are you using an SSR adapter? If so, which one?

none

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

1.How to import `base.scss` globally，I want to use ` @include` or `@extend`
```
import { defineConfig } from 'astro/config';
import mdx from '@astrojs/mdx';
import sitemap from '@astrojs/sitemap';
import svelte from '@astrojs/svelte';

saving
export default defineConfig({
  integrations: [mdx(), sitemap(), svelte()],
  vite: {
    css: {
      preprocessorOptions: {
        scss: {
          additionalData: '@use "src/styles/base.scss" as *;',
        },
      },
    },
  },
});
```
2.if the import is displayed，I use`@import '$styles/variables.scss';` 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-v55sfg?file=src/components/Test.svelte

### Participation

- [x] I am willing to submit a pull request for this issue.

bluwy:
I don't think aliases from tsconfig work in css/scss at the moment, is that the issue you're facing?

mydracula:
> 

yes I want to know how to use the path with astro and scss

PR #6812, opened 2023-04-10
Add more tsconfig alias tests

bluwy:
## Changes

Add more tests to tsconfig aliases, which covers previous regressions from:
- https://github.com/withastro/astro/issues/6620
- https://github.com/withastro/astro/issues/6613
- https://github.com/withastro/astro/pull/6566

Also added tests for `@import` in todo that I'll work on in a later PR.

## Testing

New tests should pass.

## Docs

n/a. tests only.

PR #6816, opened 2023-04-11
Support tsconfig aliases in CSS `@import`

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6551 (take 2)
Supersedes https://github.com/withastro/astro/pull/6626

This PR refactors `vite-plugin-config-alias`, and adds a `config.createResolver` patch that would make this plugin apply to CSS `@import` too. Even though this is a bit hacky, it's the only way to make it work completely.

## Testing

https://github.com/withastro/astro/pull/6812 added tests that previously caused regressions.

This PR also adds more tests for dev and build, and runs on Windows too.

## Docs

n/a. [Import aliases in the docs](https://docs.astro.build/en/guides/typescript/#import-aliases) should already implicitly apply for style `@import` too, so I don't think this needs to be explicitly documented.


Princesseuh:
Is there a world in which using a plugin like https://github.com/aleclarson/vite-tsconfig-paths would reduce the amount of code we need to do ourselves?

bluwy:
`vite-tsconfig-paths` actually has this issue too (https://github.com/aleclarson/vite-tsconfig-paths/issues/41), but if it get fixed upstream, I think it makes sense to use the plugin. It's quite exhausting to maintain this 🥲

If the patch works in this PR, I can upstream the changes too.

Princesseuh (approved):
Looks good! Would love to see us use an upstream plugin for this in the future to reduce the maintenance burden though

PR #6817, opened 2023-04-11
Refactor content collection transforms

### LINKED ISSUE #6058: Astro 2.0 warning when using sourcemaps

bholmesdev:
## Changes

- Resolve #6058
- Migrate `load` step for content collection imports and assets to the `transform` step. This reduces complexity and unblocks support for content extensions with existing transforms from Vite core, like `.json` ([reason for unblock here](https://github.com/withastro/astro/pull/6808#issuecomment-1501876918)).
- Add check for `vite.build.sourcemap` option to content collections and the MDX plugin. When present, generate source maps. Avoid generating otherwise to prevent perf bottlenecks.

### DIFF
--- a/packages/astro/src/content/vite-plugin-content-imports.ts
+++ b/packages/astro/src/content/vite-plugin-content-imports.ts
@@ -9,7 +9,6 @@ import { AstroErrorData } from '../core/errors/errors-data.js';
 import { AstroError } from '../core/errors/errors.js';
 import { escapeViteEnvReferences, getFileInfo } from '../vite-plugin-utils/index.js';
 import { CONTENT_FLAG } from './consts.js';
-import MagicString from 'magic-string';
 import {
 	getContentEntryExts,
 	getContentPaths,
@@ -85,11 +84,7 @@ export function astroContentImportPlugin({
 							rawData: ${JSON.stringify(_internal.rawData)},
 						};`);
 
-					if (settings.config.vite.build?.sourcemap) {
-						const s = new MagicString(code);
-						return { code: s.toString(), map: s.generateMap() };
-					}
-					return code;
+					return { code, map: { mappings: '' } };
 				}
 			},
 			configureServer(viteServer) {
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -1,4 +1,3 @@
-import MagicString from 'magic-string';
 import { markdownConfigDefaults } from '@astrojs/markdown-remark';
 import { toRemarkInitializeAstroData } from '@astrojs/markdown-remark/dist/internal.js';
 import { compile as mdxCompile } from '@mdx-js/mdx';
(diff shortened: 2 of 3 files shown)


packages/astro/src/vite-plugin-jsx/index.ts
```diff
@@ -104,6 +105,11 @@ interface AstroPluginJSXOptions {
 	logging: LogOptions;
 }
 
+// Format inspired by https://github.com/vitejs/vite/blob/main/packages/vite/src/node/constants.ts#L54
+const SPECIAL_QUERY_REGEX = new RegExp(
```
bholmesdev, line 109: I discovered our JSX runtime will try to process files with special flags like `?raw` and `?astroContent`. This regex should prevent that

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -65,26 +66,30 @@ export function astroContentImportPlugin({
...
+						export const data = ${devalue.uneval(data) /* TODO: reuse astro props serializer */};
+						export const _internal = {
+							filePath: ${JSON.stringify(_internal.filePath)},
+							rawData: ${JSON.stringify(_internal.rawData)},
+						};`);
+
+					if (settings.config.vite.build?.sourcemap) {
+						const s = new MagicString(code);
+						return { code: s.toString(), map: s.generateMap() };
+					}
```
bluwy, line 91: I believe for virtual files or generated content, we can return `map: { mappings: '' }` to note that there's no sourcemap (it's the file itself). This might be faster than using `MagicString`.

https://rollupjs.org/plugin-development/#source-code-transformations

packages/integrations/mdx/src/index.ts
```diff
@@ -168,7 +173,12 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 											import.meta.hot.decline();
 										}`;
 									}
-									return escapeViteEnvReferences(code);
+									code = escapeViteEnvReferences(code);
+									if (config.vite.build?.sourcemap) {
+										const s = new MagicString(code);
+										return { code: s.toString(), map: s.generateMap() };
+									}
```
bluwy, line 180: This part does transform the code so it needs a map, but it's hard to generate a mapping here. Using `MagicString` isn't quite accurate too I believe. I think we could do `return { code, map: null }` to preserve the mapping as suggested in the linked rollup docs.

The transformation in `escapeViteEnvReferences` does column-based changes, so it should be fine to preserve tha sourcemap.

bholmesdev: Understood. Changed!

packages/astro/src/content/vite-plugin-content-imports.ts
bholmesdev: Sounds good to me! It's virtual anyways so I'm not sure what use a sourcemap would have

bluwy (approved):
Nice refactor!

Issue #6788, opened 2023-04-08
Slot fallback is showing even when passing in a slot prop.


edoardofelici:
### What version of `astro` are you using?

2.2.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

The slot fallback is now showing even when a slot prop is provided to the 'slottable' component. This bug will break a lot of UIs. 

The regression version is '2.2.1'. Version '2.2.0' is working fine. 

See provided stackblitz example, where installing astro@2.2.0 fixes the problem. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-ombtg4?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

lukeshiru:
The same thing is happening to me. Here's a simple code example:

```astro
<!-- Child.astro -->
<strong>
  <slot name="foo">Default value</slot>
</strong>

<!-- Parent.astro -->
<Child>
  <span slot="foo">Hi</span>
</Child>
```

Results in:

```html
<strong>
  <span>Hi</span>Default value
</strong>
```

Expected (and previous behavior):

```html
<strong>
  <span>Hi</span>
</strong>
```

I'll roll back to `2.2.0` in the meantime. Thanks for reporting @edoardofelici!

kamranahmedse:
I can confirm the same behavior. Rolling back to `2.2.0` fixed the issue.

PR #6819, opened 2023-04-11
Fix slot fallbacks unexpectedly showing up in some cases

MoustaphaDev:
## Changes

Fix https://github.com/withastro/astro/issues/6788

Check if there's no slotted content before showing the fallback

## Testing

Check if the fallback content doesn't appear next to the slotted content

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/runtime/server/render/slot.ts
+++ b/packages/astro/src/runtime/server/render/slot.ts
@@ -35,7 +35,7 @@ export async function* renderSlot(
 		yield* iterator;
 	}
 
-	if (fallback) {
+	if (fallback && !slotted) {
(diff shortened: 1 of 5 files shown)


packages/astro/src/runtime/server/render/slot.ts
```diff
@@ -35,7 +35,7 @@ export async function* renderSlot(
 		yield* iterator;
 	}
 
-	if (fallback) {
+	if (fallback && !slotted) {
```
MoustaphaDev, line 38: The `yield` keyword doesn't return so we also have to check that there's no slotted content

Princesseuh (approved):

PR #6836, opened 2023-04-13
Ensure unique workspace name

bluwy:
## Changes

A small chore to ensure workspace packages have a unique name. This is required when I upgrade turbo, but for some reason turbo also scans gitignored packages (e.g. build output), and those build output have empty names causing https://github.com/vercel/turbo/issues/3340

So I stick with the old turbo version for now, but thought it would be good to do this cleanup.

## Testing

Existing tests should pass.

## Docs

n/a. cleanup.


MoustaphaDev (approved):
Fan`test`ic!

PR #6847, opened 2023-04-14
Use Markdoc on `*.mdoc.md` files instead of `.mdoc`

ancaemcken:
## Changes

Allows Markdoc parsing in `*.mdoc.md` files, removing editor choice restrictions for adopters of Markdoc. 
Modified the `examples/with-markdoc` example app:
  - with a plain `.md` file that is supposed to use the regular Astro markdown with remark & rehype plugins.
  - `package.json` with needed plugins for rendering a table of contents in plain `.md` file. 
  - `astro.config.mjs` with `markdown` configuration. 

## Testing

Updated @astrojs/markdoc tests to work for `*.mdoc.md` and ran the full suite of tests.

## Docs

Not updated yet. This is a POC. But willing to do so.


bholmesdev:
Took a note of this idea! There's definitely implications to _avoiding_ our `.md` parser when preceded by `.mdoc`. I'll be taking a closer look at Markdoc features these next couple weeks and I'll offer feedback then 👍

bholmesdev:
Discussing with the team now! Though I'll share there's some fundamental changes we'd need in this PR:
- **We shouldn't [let Markdoc own the base `.md` extension](https://github.com/withastro/astro/pull/6847/files#diff-6f1a5db26f5ed6e5118cf55b197f2368253e83450989fcd71540ebfe56b5eb38R49).** This will cause conflicts with our Markdown pipeline, where we want `.md` files to run through our remark pipeline and Markdoc to run through the integration (at least today). We include TypeScript completions for `*.md` files, and a Markdoc override would break this contract. Instead, `extensions` should be `['.mdoc', '.mdoc.md']`.
- **[This special check for `.mdoc` files](https://github.com/withastro/astro/pull/6847/files#diff-0d71e7ce516d20f04d81efa8379a6a88bd0ded81c5ba60ad1ac99f719756f0fbR150-R151) is dangerous** for the Markdown conflicts mentioned above. To ensure `.mdoc.md` runs through Markdoc before getting caught by our Markdown renderer, we likely need to sort `extensions` where doubled extensions `.mdoc.md` are caught before single extensions like `.md`.
- **Markdoc should not be processed by remark.** I noticed quite a few remark plugins were added to the `with-markdoc` example, and I wouldn't expect any of these to be run on Markdoc content. Can these be removed?

bholmesdev:
Discussing with the team now! After giving this PR a look, I see some more fundamental issues to work out:
- **We shouldn't [let Markdoc own the base `.md` extension](https://github.com/withastro/astro/pull/6847/files#diff-6f1a5db26f5ed6e5118cf55b197f2368253e83450989fcd71540ebfe56b5eb38R49).** This will cause conflicts with our Markdown pipeline, where we want `.md` files to run through our remark pipeline and Markdoc to run through the integration (at least today). Instead, the configured content entry `extensions` should be `['.mdoc', '.mdoc.md']`.
- **[This special check for `.mdoc` files](https://github.com/withastro/astro/pull/6847/files#diff-0d71e7ce516d20f04d81efa8379a6a88bd0ded81c5ba60ad1ac99f719756f0fbR150-R151) is also problematic** for the Markdown conflicts mentioned above. To ensure `.mdoc.md` runs through Markdoc before getting caught by our Markdown renderer, we likely need to sort `extensions` where doubled extensions like `.mdoc.md` are respected before single extensions like `.md`.
- **Markdoc should not be processed by remark.** I noticed quite a few remark plugins were added to the `with-markdoc` example, and I wouldn't expect any of these to be run on Markdoc content. Can these be removed?
- **`.mdoc.md` will conflict with our [Markdown import type](https://github.com/withastro/astro/blob/main/packages/astro/client-base.d.ts#L87).** We allow ESM imports for Markdown files today, and `*.mdoc.md` imports will give false type information. I've tried excluding `*.mdoc.md` files from the declare module statement, but I haven't found a syntax that TypeScript supports. We're still debating if this is a blocker 😕 

Hope this feedback makes sense! Thanks again for the contribution

matthewp:
Closing as we don't want to do this right now.

PR #6848, opened 2023-04-14
feat(image): Update image service config to allow passing settings to the service

Princesseuh:
## Changes

Certain image services need a separate config, this is applicable for both remote and local image services. For instance, you might want a setting to restrict the remote patterns supported. Or for remote services, maybe your service only support a certain amount of widths and heights, or you can set defaults etc etc.

It looks like this:
```ts
export default defineConfig({
  experimental: {
    assets: true,
  },
  image: {
    service: { entrypoint: "./src/imageService.ts", config: {widths: [300, 600]}}
  },
});
```

and the settings are passed to all the hooks as a second parameter:
```ts
const service: ExternalImageService = {
  validateOptions(options, serviceOptions) {
  	if (!serviceOptions.widths.includes(options.width)) {
  		throw new Error("Specified width is not currently present in your service config")
  	}
  }
}
```

---

Since we have a new shape, I also added utilities functions to `astro/assets` to set the service:

```ts
import { squooshImageService } from "astro/assets/;

export default defineConfig({
  experimental: {
    assets: true,
  },
  image: {
    service: squooshImageService()
  },
});
```

## Testing

Added a test

## Docs

https://github.com/withastro/docs/pull/3096


packages/astro/src/@types/astro.ts
```diff
@@ -746,26 +752,26 @@ export interface AstroUserConfig {
 		/**
 		 * @docs
 		 * @name image.service (Experimental)
-		 * @type {'astro/assets/services/sharp' | 'astro/assets/services/squoosh' | string}
-		 * @default `'astro/assets/services/squoosh'`
+		 * @type {{entrypoint: 'astro/assets/services/sharp' | 'astro/assets/services/squoosh' | string, config: Record<string, any>}}
+		 * @default `{entrypoint: 'astro/assets/services/squoosh', config?: {}}`
 		 * @version 2.1.0
```
sarah11918, line 757: I'm just curious because I don't know: Do we need to worry about updating the version number here since this specific configuration will only be appropriate for 2.3.x? (Even though `image.service` itself was added in 2.1.0?)

Are we at all worried about this being misleading for someone who happens to be on a slightly older version of Astro before this update to the image service config?

packages/astro/src/assets/index.ts
```diff
@@ -1,5 +1,21 @@
...
 export { imageMetadata } from './utils/metadata.js';
+
+export function getSharpImageService(): ImageServiceConfig {
+	return {
+		entrypoint: 'astro/assets/services/sharp',
+		config: {},
+	};
+}
+
+export function getSquooshImageService(): ImageServiceConfig {
```
bluwy, line 16: I'd maybe lean into removing the `get` so it's only `sharpImageService` and `squooshImageService`, as I'd imagine third-party image service would've omitted the `get` too in their API, e.g. `cloudinaryImageService`. But curious to hear other opinions too.

bluwy (approved):
The general implementation looks good to me

Issue #6829, opened 2023-04-12
Issues with Markdoc


louiss0:
### What version of `astro` are you using?

2.2.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm 

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

There are a few problems with Mardoc. I decided to test it and see if things were going to be ok. But I have found a few problems with it 
- Functions can't be used I get an error called invalid child when I attempt to use a function
- Syntax highlighting is not automatically enabled 
- Code Comments are added to each page.

To see the first two problems go to the `markdown-style-guide.mdoc` page. 
I created the Code component with the fence so that it could just be implemented later on. Other than that I will write a discussion soon.   

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-zdtvav

### Participation

- [] I am willing to submit a pull request for this issue.

louiss0:
I have recently discovered that markdoc wraps each page in an article tag I hope this is not the way this integration will work in the near future. This is where I discovered it from [here](https://github.com/withastro/docs/issues/2904). This is another problem.

louiss0:
I recently found out that Markdoc Errors are not displayed because of the Astro Markdoc Implementatiion. I tried to talk to them about error messages but instead I was told this. 

> Can you tell me how you are performing the validation? It seems like you are just seeing the error id, but Markdoc itself provides very explicit messages. For example, an invalid type shows the [message](https://github.com/markdoc/markdoc/blob/main/src/validator.ts#L209) "Attribute '${key}' must be type of '${t}" and an error on a matches rule shows the [message](https://github.com/markdoc/markdoc/blob/main/src/validator.ts#L223) "Attribute '${key}' must match one of ${matches}".
> Ryan Paul Maintainer of Stripe 

I think there is a way of transferring errors from one catcher to another. I think you decided to override the Markdoc errors with your own one.

paulrudy:
> I have recently discovered that markdoc wraps each page in an article tag I hope this is not the way this integration will work in the near future. This is where I discovered it from [here](https://github.com/withastro/docs/issues/2904). This is another problem.

[...]

PR #6850, opened 2023-04-14
Data collections and references

bholmesdev:
## Demo

https://user-images.githubusercontent.com/51384119/235770585-66ebc2dd-423a-49bc-8283-c44b0c63dfe3.mp4

Implementation for the data collection and collection reference RFCs: https://github.com/withastro/roadmap/pull/569

## Changes

Introducing data collections! These are JSON-based collections you can query throughout your app, and "reference" from existing content collections.
- **Introduce a reserved `src/data/` directory** for storing collections of JSON entries. This uses the same file-based conventions as content collections.
- **Add a new `type: 'data'` property to `defineCollection`** to configure a data collection from your `src/content/config.ts`.
- **Add a new query API** to fetch data and content entries by ID: `getEntry()`. Data collections do _not_ have slugs, so we cannot use the existing `getEntryBySlug()` helper.
- **Add a new `reference()` extension function** to any configured collection. This can be used like a table join in a database, referencing other collections by key.

## Testing

- Test data collections
- Test collection references
- Test mixed content errors on `sync`

## Docs

https://github.com/withastro/docs/pull/3233


bholmesdev:
!preview data-collections

bholmesdev:
!preview data-collections

bholmesdev:
!preview data-collections

louiss0:
I'm concerned about the fact that each file format might become a new folder name is this the only folder for data that is not markdown? What about other files?

louiss0:
> !preview data-collections

I know you guys said you were going to support content collections. But I wonder what other formats are you going to support? I wonder whether or not other folders would be taken from people. I think from now on it's better to stop taking top level folders from people and instead just take folders in content collections. 

I'm saying this because with this PR four folders are taken from people. One of them being `assets`. That folder was used for more than images by some people, especially React Devs.

bholmesdev:
@louiss0 To clarify, we do _not_ require or own an `assets/` directory! This is a convention we like to demo to show alias support, but it's not required to use image optimization.

As for the proliferation of folders, that _is_ a good concern to keep in mind. As we see it, there won't be further "collection" types beyond content and data, so I'd expect `content/` and `data/` to be the only two reserved directories (other than `pages/` of course). If we do find a reason for more content types, I agree combining into a single dir or using config-based collections would be best!

louiss0:
> @louiss0 To clarify, we do _not_ require or own an `assets/` directory! This is a convention we like to demo to show alias support, but it's not required to use image optimization.
> 
> As for the proliferation of folders, that _is_ a good concern to keep in mind. As we see it, there won't be further "collection" types beyond content and data, so I'd expect `content/` and `data/` to be the only two reserved directories (other than `pages/` of course). If we do find a reason for more content types, I agree combining into a single dir or using config-based collections would be best!
Ok then I wonder Is there anyone working on Markdoc? I have found some issues with it and no one has responded so far.

louiss0:
Oh, I'm pro moving the `config.ts` out of the content file. Maybe it will work like a regular config file instead. Is there any other options that will be added to define Collection in the future?

louiss0:
I'm also concerned about Markdoc. They are written [here](https://github.com/withastro/roadmap/discussions/554).

bholmesdev:
@louiss0 I'm also the active maintainer of `@astrojs/markdoc` 🙃 I'm trying not to divide my attention too much, so I've left Markdoc proposals in my unread until next week. Happy to take a look at your roadmap discussion!

louiss0:
> @louiss0 I'm also the active maintainer of `@astrojs/markdoc` 🙃 I'm trying not to divide my attention too much, so I've left Markdoc proposals in my unread until next week. Happy to take a look at your roadmap discussion!

That thing has problems I tried to file an issue but that was shut down by Matt. I think you are the one to solve the problems from that issue. He told me to spilt up each of the problems into different issues But I think they are too small. Here's the link to the issues https://github.com/withastro/astro/issues/6829#issuecomment-1527946589.

louiss0:
> @louiss0 I'm also the active maintainer of `@astrojs/markdoc` 🙃 I'm trying not to divide my attention too much, so I've left Markdoc proposals in my unread until next week. Happy to take a look at your roadmap discussion!

Keep going bud. I'm in no rush to use Markdoc. I care more about Assets right now. I just decided to put my thoughts out there. I don't know how much you are going to extend it or if Remark is going to be involved but I just hope we don't lose the DX of using mdx too much.

packages/astro/src/content/utils.ts
```diff
@@ -146,33 +178,45 @@ export function getContentEntryConfigByExtMap(settings: Pick<AstroSettings, 'con
 	return map;
 }
 
-export class NoCollectionError extends Error {}
-
-export function getEntryInfo(
-	params: Pick<ContentPaths, 'contentDir'> & {
-		entry: string | URL;
-		allowFilesOutsideCollection?: true;
+export function getEntryCollectionName({
```
bholmesdev, line 181: Refactored wacky `NoCollectionError` and function overloading to a separate query for collection names. This hopefully cleans up some readability problems!

packages/astro/src/core/sync/index.ts
```diff
@@ -99,6 +110,9 @@ export async function sync(
 		}
 	} catch (e) {
 		const safeError = createSafeError(e);
+		if (isAstroError(e)) {
+			throw e;
+		}
```
bholmesdev, line 115: Avoid swallowing AstroError hints and error codes

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1000,6 +1000,52 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * @description
+	 * A content collection cannot contain a mix of content and data entries. You must store entries in separate collections by type.
+	 */
+	MixedContentDataCollectionError: {
+		title: 'Content and data cannot be in same collection.',
+		code: 9005,
+		message: (collection: string) => {
+			return `**${collection}** contains a mix of content and data entries. All entries must be of the same type.`;
+		},
+		hint: 'Store data entries in a new collection separate from your content collection, and share a `schema` in your content config.',
```
sarah11918, line 1031:
```suggestion
		hint: 'Store data entries in a new collection separate from your content collection.',
```
As we discussed in the data collections PR, I hesitate to mention "sharing a schema" without an explicit example of what that means. Right now, the docs there show writing the schema and importing it in a separate file. I wouldn't be opposed to expanding that existing example to show TWO collections, one content and one data, using the same imported schema! (With a note that explicitly mentions "sharing a schema" when the structures are the same, even if the content type is different.)

So, I'd either truncate this (personally), OR add sharing a schema to the collections docs.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1000,6 +1000,52 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	},
+	/**
+	 * @docs
+	 * @message `COLLECTION_NAME` contains entries of type `ACTUAL_TYPE`, but is configured as a `EXPECTED_TYPE` collection.
+	 * @see
+	 * - [Defining content collections](https://docs.astro.build/en/guides/content-collections/#defining-collections)
+	 * @description
+	 * Content collections must contain entries of the type configured. If your collection config is missing a `type`, it will be considered `type: 'content'` by default. Try adding `type: 'data'` to your collection config for data collections.
```
sarah11918, line 1039:
```suggestion
	 * Content collections must contain entries of the type configured. Collections are `type: 'content'` by default. Try adding `type: 'data'` to your collection config for data collections.
```
I think this could be totally shortened!

sarah11918 (commented):
These error messages are super-nicely done, @bholmesdev! I would be honoured to make a mistake with my collections!  Just a couple of thoughts for you to consider below!

packages/astro/src/core/sync/index.ts
```diff
@@ -74,6 +75,16 @@ export async function sync(
...
 
+	// Patch `ws.send` to bubble up error events
+	// `ws.on('error')` does not fire for some reason
+	const wsSend = tempViteServer.ws.send;
+	tempViteServer.ws.send = (payload: HMRPayload) => {
+		if (payload.type === 'error') {
+			throw payload.err;
+		}
+		return wsSend(payload);
+	};
```
bluwy, line 86: I think it's because the Vite server `logLevel` is set to silent?

https://github.com/withastro/astro/blob/4516d7b22c5979cde4537f196b53ae2826ba9561/packages/astro/src/core/sync/index.ts#L71

.changeset/early-eyes-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+'astro': minor
+'@astrojs/markdoc': patch
```
bluwy, line 3: Should this be a minor as the new version won't work in older versions of Astro anymore? (Due to content flag change)

bluwy (commented):
Tried my best to review this huge PR 😄 Code looks good to me. Have a couple extra suggestions below

packages/astro/src/content/runtime.ts
```diff
@@ -110,6 +134,92 @@ export function createGetEntryBySlug({
...
+	getEntryImport: GetEntryImport;
+	getRenderEntryImport: GetEntryImport;
+}) {
+	return async function getEntry(
+		collectionOrLookupObject:
+			| string
+			| { collection: string; id: string }
+			| { collection: string; slug: string },
+		lookupId?: string
+	) {
```
ematipico, line 171: nit: I find this part of the code difficult to read, probably because of these inline types. And there's also the fact that this function doesn't have a return type... so while the compiler can infer what's returning, I honestly can't understand it 😆 

Any chance to avoid inline types, at least in this case?

packages/astro/src/content/runtime.ts
```diff
@@ -110,6 +134,92 @@ export function createGetEntryBySlug({
...
+		let collection: string, id: string;
+		if (typeof collectionOrLookupObject === 'string') {
+			collection = collectionOrLookupObject;
+			id = lookupId!;
```
ematipico, line 175: Could you add a comment saying why it's safe to assume that `lookupId` is not `undefined`?

packages/astro/src/content/template/virtual-mod.mjs
```diff
@@ -3,28 +3,33 @@ import {
...
-	throw new Error(
-		'Importing `image()` from `astro:content` is no longer supported. See https://docs.astro.build/en/guides/assets/#update-content-collections-schemas for our new import instructions.'
-	);
-};
-
 const contentDir = '@@CONTENT_DIR@@';
 
-const entryGlob = import.meta.glob('@@ENTRY_GLOB_PATH@@', {
-	query: { astroContent: true },
+const contentEntryGlob = import.meta.glob('@@CONTENT_ENTRY_GLOB_PATH@@', {
```
ematipico, line 16: We should have these unique strings saved somewhere, all together. Or at least follow one pattern for consistency: `'@@CONTENT_DIR@@'` is held in a variable; instead, this one is not.

packages/astro/src/content/utils.ts
```diff
@@ -146,33 +183,45 @@ export function getContentEntryConfigByExtMap(settings: Pick<AstroSettings, 'con
...
+	contentDir,
+	entry,
+}: Pick<ContentPaths, 'contentDir'> & { entry: string | URL }) {
+	const entryPath = typeof entry === 'string' ? entry : fileURLToPath(entry);
+	const rawRelativePath = path.relative(fileURLToPath(contentDir), entryPath);
+	const collectionName = path.dirname(rawRelativePath).split(path.sep).shift() ?? '';
```
ematipico, line 192: nit: We could use [`at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) :)

```js
path.dirname(rawRelativePath).split(path.sep).at(-1)
```

More readable IMHO

packages/astro/src/content/utils.ts
```diff
@@ -146,33 +183,45 @@ export function getContentEntryConfigByExtMap(settings: Pick<AstroSettings, 'con
...
-		fileURLToPath(contentDir),
-		typeof entry === 'string' ? entry : fileURLToPath(entry)
-	);
-	const rawCollection = path.dirname(rawRelativePath).split(path.sep).shift();
-	const isOutsideCollection = rawCollection === '..' || rawCollection === '.';
+	collection,
+}: Pick<ContentPaths, 'contentDir'> & { entry: URL; collection: string }): string {
+	const rawRelativePath = path.relative(fileURLToPath(contentDir), fileURLToPath(entry));
+	const rawId = path.relative(collection, rawRelativePath);
+	const rawIdWithoutFileExt = rawId.replace(new RegExp(path.extname(rawId) + '$'), '');
```
ematipico, line 210: Out of curiosity, what's that `$` about?

packages/astro/src/content/utils.ts
```diff
@@ -146,33 +183,45 @@ export function getContentEntryConfigByExtMap(settings: Pick<AstroSettings, 'con
...
+	entry,
+	contentDir,
+	collection,
+}: Pick<ContentPaths, 'contentDir'> & { entry: URL; collection: string }): {
+	id: string;
+	slug: string;
+} {
+	const rawRelativePath = path.relative(fileURLToPath(contentDir), fileURLToPath(entry));
+	const rawId = path.relative(collection, rawRelativePath);
 	const rawIdWithoutFileExt = rawId.replace(new RegExp(path.extname(rawId) + '$'), '');
```
ematipico, line 225: This code seems to be repeated. Any chance of having it in a function?

packages/astro/src/content/utils.ts
```diff
@@ -278,6 +325,18 @@ export function parseFrontmatter(fileContents: string, filePath: string) {
...
 
+// TODO: hoist to proper error
+const InvalidDataCollectionConfigError = {
+	...AstroErrorData.UnknownContentCollectionError,
+	message: (dataExtsStringified: string, collection: string) =>
+		`Found a non-data collection with ${dataExtsStringified} files: **${collection}.** To make this a data collection, use the \`defineDataCollection()\` helper or add \`type: 'data'\` to your collection config.`,
+};
+
+export function hasContentFlag(viteId: string, flag: (typeof CONTENT_FLAGS)[number]) {
+	const flags = new URLSearchParams(viteId.split('?')[1]);
```
ematipico, line 336: Would you mind leaving a comment saying why it's safe to query the index 1 of the array?

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -55,14 +68,46 @@ export function astroContentImportPlugin({
...
+		}
+	}
+
 	const plugins: Plugin[] = [
 		{
 			name: 'astro:content-imports',
 			async transform(_, viteId) {
-				if (isContentFlagImport(viteId)) {
+				if (hasContentFlag(viteId, DATA_FLAG)) {
+					const fileId = viteId.split('?')[0];
```
ematipico, line 87: Same here, we should leave a comment saying why it's safe to query the index 0.

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -297,3 +352,57 @@ async function getContentConfigFromGlobal() {
...
+	fs,
+	pluginContext,
+	settings,
+}: {
+	fs: typeof fsMod;
+	fileId: string;
+	contentPaths: Pick<ContentPaths, 'contentDir'>;
+	pluginContext: PluginContext;
+	dataEntryExtToParser: Map<string, DataEntryType['getEntryInfo']>;
+	settings: Pick<AstroSettings, 'config'>;
```
ematipico, line 369: Ben likes inline types! :D

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -297,3 +352,57 @@ async function getContentConfigFromGlobal() {
...
+}): Promise<DataEntryModule> {
+	const contentConfig = await getContentConfigFromGlobal();
+	const rawContents = await fs.promises.readFile(fileId, 'utf-8');
```
ematipico, line 372: I/O operations are always a hazard... we should always assume that they could throw an error :(

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -297,3 +352,57 @@ async function getContentConfigFromGlobal() {
...
+	const { contentDir } = contentPaths;
+	const collection = getEntryCollectionName({ entry, contentDir });
+	if (collection === undefined) throw new AstroError(AstroErrorData.UnknownContentCollectionError);
+
+	const id = getDataEntryId({ entry, contentDir, collection });
+
+	const _internal = { filePath: fileId, rawData };
+
+	const collectionConfig = contentConfig?.collections[collection];
+	let data = collectionConfig
```
ematipico, line 398: nit: `const`?

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -29,34 +33,41 @@ export function astroContentVirtualModPlugin({
...
 		.readFileSync(contentPaths.virtualModTemplate, 'utf-8')
+		.replace(
+			'@@COLLECTION_NAME_BY_REFERENCE_KEY@@',
+			new URL('reference-map.json', contentPaths.cacheDir).pathname
+		)
 		.replace('@@CONTENT_DIR@@', relContentDir)
-		.replace('@@ENTRY_GLOB_PATH@@', entryGlob)
-		.replace('@@RENDER_ENTRY_GLOB_PATH@@', entryGlob);
+		.replace('@@CONTENT_ENTRY_GLOB_PATH@@', `${relContentDir}**/*${getExtGlob(contentEntryExts)}`)
+		.replace('@@DATA_ENTRY_GLOB_PATH@@', `${relContentDir}**/*${getExtGlob(dataEntryExts)}`)
```
ematipico, line 46: I see that this string is repeated... one more reason to move it inside a constant somewhere! (avoid typos)

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -70,52 +81,65 @@ export function astroContentVirtualModPlugin({
...
-				// Log warning during type generation, safe to ignore in lookup map
-				if (info instanceof NoCollectionError) return;
+		contentGlob.map(async (filePath) => {
+			const entryType = getEntryType(filePath, contentPaths, contentEntryExts, dataEntryExts);
+			// Globbed ignored or unsupported entry.
+			// Logs warning during type generation, should ignore in lookup map.
+			if (entryType !== 'content' && entryType !== 'data') return;
+
+			const collection = getEntryCollectionName({ contentDir, entry: pathToFileURL(filePath) });
+			if (!collection) throw UnexpectedLookupMapError;
```
ematipico, line 125: Love it! ❤️

packages/astro/src/core/config/settings.ts
```diff
@@ -21,6 +24,38 @@ export function createBaseSettings(config: AstroConfig): AstroSettings {
 				: [],
 		pageExtensions: ['.astro', '.html', ...SUPPORTED_MARKDOWN_FILE_EXTENSIONS],
 		contentEntryTypes: [markdownContentEntryType],
+		dataEntryTypes: [
+			{
+				extensions: ['.json'],
+				getEntryInfo({ contents }) {
+					if (contents === undefined || contents === '') return { data: {} };
+
+					const data = JSON.parse(contents);
```
ematipico, line 33: This function can throw a runtime error if `contents` is not a valid JSON object. We should have a `try/catch` block

ematipico (commented):
There are some `TODO` in the code, but I don't understand when you aim to address them. This PR? Next PR? Maybe a mention in the description of the PR could shed some light;

There are a few places where we need to catch errors:
- I/O operations, it's minor but you never the permissions of the file system
- `JSON.parse`, **very important**. If the contents of a file are not a valid JSON object, the function will throw a runtime error.

matthewp (approved):

.changeset/early-eyes-bow.md
bholmesdev: Agreed, will bump

.changeset/early-eyes-bow.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Introducing data collections and references. These offer a way to store JSON file as collection entries under the new `src/data/` directory. You can reference this data from existing content collections using "references." Learn more on our docs (TODO!)
```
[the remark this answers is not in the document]
bholmesdev: Yes we did! This changeset needs some love

packages/astro/src/content/runtime.ts
bholmesdev: Added an error for this!

packages/astro/src/content/runtime.ts
bholmesdev: Refactored to explicit return types and added an inline comment for the weird naming

packages/astro/src/content/vite-plugin-content-imports.ts
bholmesdev: Ugh, exceptions are so hard to track down in JS... I can add something here

packages/astro/src/core/sync/index.ts
bluwy: I was looking at [this Vite code](https://github.com/vitejs/vite/blob/d09bbd093a4b893e78f0bbff5b17c7cf7821f403/packages/vite/src/node/server/middlewares/error.ts#L43-L58) where I found the `logLevel`. Not a big deal since we're pushing to get this over the line, and we can follow up later 👍

packages/astro/src/content/template/virtual-mod.mjs
bholmesdev: Ah, we sadly can't change this for glob paths. Vite requires all glob imports to be static strings, and **does not support variables.** Messy, but it's the only way to use regex replace 🤷

packages/astro/src/content/utils.ts
bholmesdev: So you like `.at` but not `??=` huh... 😉

ematipico: I am an old man, I was thought I could use - in **Java** - `[-1]` to get the last element of an array, and then JavaScript ruined everything 😆

bholmesdev: Ah, realized it's `[0]` actually! Probably telling that I can't remember what `shift()` vs `unshift()` is. Changed

packages/astro/src/content/utils.ts
bholmesdev: Well `URLSearchParams` accepts `undefined`, so this should always be safe. Would rather avoid noisy comments here... would adding `?? ''` help?

packages/astro/src/content/vite-plugin-content-imports.ts
bholmesdev: JS returns `undefined` even when an index is out-of-bounds, so I'm not sure it's worth the comment?

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
bholmesdev: Per comment above, we can't hoist these to a `consts` file due to Vite's limitations 😕

packages/astro/src/content/vite-plugin-content-imports.ts
ematipico: Looks great! Thank you!

PR #6851, opened 2023-04-15
add warn message when using unsupported file types in pages/

timozander:
## Changes

- add warning message when files with unsupported extensions are used in pages/
console output

<details><summary>Motivation</summary>
<p>
Occasionally, I messed up the file extensions and wondered why the page is not found. Especially since Vite just shows "updated file xyz", I think this warning message might be useful to the user
</p>
</details> 

## Testing

Tested with the projects from `examples/`, unit testing logging does not really make sense

## Docs

Docs are not needed since this feature serves as a way to alert the user that something is wrong. Docs are already clear on which file extensions are allowed

### DIFF
--- a/.changeset/hip-bottles-hide.md
+++ b/.changeset/hip-bottles-hide.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Added warning message when using unsupported file extensions in pages/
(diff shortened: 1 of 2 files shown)


bluwy:
This seems like a nice addition, but it does make setups that "intentionally put images beside the page file so it can be imported relatively" show the warning. I don't think it's a big deal since we're introducing the `src/assets` convention, so I'm fine with this change.

Curious to hear what others think too.

matthewp (approved):

ematipico (approved):

PR #6857, opened 2023-04-16
Shortcuts in `astro dev` like in vite

ronanru:
## Changes

In vite projects while the dev server is running, I can press `r` in the terminal to restart it, or press `u` to show the server URL again. I was missing this feature in astro. Today I thought, why don't I add it myself.

## Testing

not sure how to test this

## Docs

I added an info message when the dev server starts. Don't think any additions to docs are necessary.

### DIFF
--- a/.changeset/shiny-hounds-buy.md
+++ b/.changeset/shiny-hounds-buy.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+added shortcuts to `astro dev`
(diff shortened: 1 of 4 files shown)


matthewp (changes requested):
Thanks for submitting! Marking this as "Request changes" because we like to release minors all together on a schedule. We'll review this separately and release the block if approved and once we are ready for our next minor.

florian-lefebvre:
Any news about this?

bluwy:
We'd like to support this when we use Vite 5 as we can re-use its exposed shortcuts logic. I'll close this for now but will be sure to setup co-authors when we get to it (It's in our internal roadmap). Thanks for kicking off the discussion at the meantime @ronanru

PR #6862, opened 2023-04-17
fix: astro island urls missing assetsPrefix in SSR mode

### LINKED ISSUE #6852: assetsPrefix is not prepended to astro island component-url and renderer-url

2.3.0

Node.js

npm

Mac

Chrome

### Describe the Bug

I'm using the new assetPrefix and it works for most of the assets. However, it seems that some of the js assets were missing the assetPrefix, so they continue to be served from the app server instead of from the CDN.

The stackblitz below can be used to reproduce. Adding a svelte component, this is what gets generated:

```
<astro-island uid="byYzy" component-url="/_astro/Test.dfb48ca8.js" component-export="default" renderer-url="/_astro/client.c4e17359.js" ...
```

Notice that `component-url` and `renderer-url` are still relative paths. They are served from node.js rather than from the assetPrefix.

https://stackblitz.com/edit/github-q7ujbk-kqvmcl

### Participation

jcdogo:
## Changes

- Fixes #6852
- In SSR mode, `assetsPrefix` is not being prepended to component-url and renderer-url.

## Testing

We added a test to `astro/test/astro-assets-prefix.test.js`.

```js
it('react component astro-island should import from assetsPrefix', async () => {
	const request = new Request('http://example.com/custom-base/');
	const response = await app.render(request);
	expect(response.status).to.equal(200);
	const html = await response.text();
	const $ = cheerio.load(html);
	const island = $('astro-island');
	expect(island.attr('component-url')).to.match(assetsPrefixRegex);
	expect(island.attr('renderer-url')).to.match(assetsPrefixRegex);
});
```

A similar test case already existed for the SSG case, but not for SSR. I just copied and adapted the test case for SSR. When the test case was added for SSR, it was failing until we implemented the fix.

## Docs

This change should not really affect the user's behavior since it is a bug fix.

### DIFF
--- a/.changeset/large-pens-rhyme.md
+++ b/.changeset/large-pens-rhyme.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Fixes bug with assetsPrefix not being prepended to component-url and renderer-url in astro islands when using SSR mode.
--- a/.changeset/large-pens-rhyme.md
+++ b/.changeset/large-pens-rhyme.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 Fixes bug with assetsPrefix not being prepended to component-url and renderer-url in astro islands when using SSR mode.
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -21,6 +21,7 @@ import {
 } from '../render/index.js';
 import { RouteCache } from '../render/route-cache.js';
 import {
+	createAssetLink,
 	createLinkStylesheetElementSet,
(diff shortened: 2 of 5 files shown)


Princesseuh (approved):

.changeset/large-pens-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
Princesseuh, line 2:
```suggestion
'astro': patch
```

This is just a bug fix, no need for a minor changeset

bluwy (approved):

bluwy:
Thanks for the fix!

PR #6867, opened 2023-04-17
docs: improve CONTRIBUTING.md

nix6839:
## Changes

- Update prerequisite to latest information (See: #6823 and #5782).
- Fix typo.
- Add language for code blocks.
- Add angle brackets around the bare URLs.
- Update changesets repository URL.
- Fix invalid fragment link.
- Improve grammar.

## Testing

Only document changed.


CONTRIBUTING.md
```diff
@@ -271,7 +271,7 @@ When in prerelease mode, the automatic PR release process will no longer release
 1. Run `pnpm exec release` to publish your new release.
 1. Run `git push && git push --tags` to push your new release to GitHub.
 1. Run `git push release/0.X:latest` to push your release branch to `latest`.
-1. Go to https://github.com/withastro/astro/releases/new and create a new release. Copy the new changelog entry from https://github.com/withastro/astro/blob/latest/packages/astro/CHANGELOG.md.
+1. Go to <https://github.com/withastro/astro/releases/new> and create a new release. Copy the new changelog entry from <https://github.com/withastro/astro/blob/latest/packages/astro/CHANGELOG.md>.
```
bluwy, line 274: I think GH markdown should work without the angle brackets, so I'm not sure if this needs to be added here.

ElianCodes: While true that it is not required in the Github flavoured markdown, I would opt to add them, we also do this everywhere else IIRC

CONTRIBUTING.md
```diff
@@ -10,14 +10,14 @@ We welcome contributions of any size and skill level. As an open source project,
...
-pnpm: "^7.9.5"
+node: "^>=16.12.0"
+pnpm: "^8.2.0"
 # otherwise, your build will fail
 ```
 
 ### Setting up your local repo
 
-Astro uses pnpm workspaces, so you should **always run `pnpm install` from the top-level project directory.** running `pnpm install` in the top-level project root will install dependencies for `astro`, and every package in the repo.
+Astro uses pnpm workspaces, so you should **always run `pnpm install` from the top-level project directory.** Running `pnpm install` in the top-level project root will install dependencies for `astro`, and every package in the repo.
```
ElianCodes, line 20:
```suggestion
Astro uses pnpm workspaces, so you should **always run `pnpm install` from the top-level project directory**. Running `pnpm install` in the top-level project root will install dependencies for `astro`, and every package in the repo.
```

CONTRIBUTING.md
```diff
@@ -10,14 +10,14 @@ We welcome contributions of any size and skill level. As an open source project,
 ### Prerequisite
```
ElianCodes, line 10:
```suggestion
### Prerequisites
```

Yan-Thomas (approved):
Thanks @nix6839, LGTM!

CONTRIBUTING.md
bluwy: Interesting, I'm not sure if we're using angle brackets everywhere now, but if it's a pattern the docs prefer, I'm fine with adding it.

nix6839: It might be a good idea to spend some time looking at other documents later. For now, I resolve this conversation

Yan-Thomas: I personally haven't seen the usage of angle brackets in the docs before, we rarely expose links like this, we rather associate it with a piece of content using the `[link title](link)` syntax.

> We should also consider people viewing through other markdown viewers

I would not mind adding them, I'm just not sure where someone would read this if not inside a GitHub product (GitHub.com, GitHub Mobile, etc.) and I am a bit worried if we change this in READMEs that are parsed into our docs site, as I'm not sure if it consider this syntax.

nix6839: It seems better not to consider other viewers (with little demand) for the sake of consistency and safety. I'll remove the angle brackets.

sarah11918 (commented):
Docs approves the changes!

PR #6874, opened 2023-04-18
Align adapter setup

nblackburn:
## Changes

- Refactored static adapter setup to use updateConfig
- Added changeset

## Testing

- Ran unit tests

## Docs

- No externally facing changes are present.

### DIFF
--- a/.changeset/hungry-lizards-float.md
+++ b/.changeset/hungry-lizards-float.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': patch
+---
+
+Refactor static adapter to use updateConfig method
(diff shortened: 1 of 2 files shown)


.changeset/hungry-lizards-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vercel': minor
```
bluwy, line 2:
```suggestion
'@astrojs/vercel': patch
```

This should be a patch since it refactors internal stuff.

nblackburn: You're right, I have updated the branch.

PR #6877, opened 2023-04-20
Upgrade to Vite 4.3

bluwy:
## Changes

Upgrade to Vite 4.3

## Testing

Existing tests should pass

## Docs

n/a


.changeset/purple-jars-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
ematipico, line 2: Does astro expose some `vite` APIs? If so, this should count as `minor`.

ematipico (approved):

.changeset/purple-jars-pay.md
birkskyum: I guess having the vite config inside astro config expose vite apis https://docs.astro.build/en/guides/configuring-astro/#customising-output-filenames

bluwy: Yeah I think it'll only be through the `vite` config option, but since users could already use Vite 4.3 by running `npm update vite` to change the version Astro will use, I made it as a patch.

bluwy:
> For some reason, the astro docs always cold starts. Probably worth checking again. Build times were quite similar too.

Fixed in https://github.com/withastro/astro/pull/6902

matthewp (approved):

PR #6902, opened 2023-04-24
Disable Vite optimizer for sync and config loading

bluwy:
## Changes

Disable the optimizer with `optimizeDeps.disabled` so that the optimize metadata is not overwritten

### Explanation

In the astro docs repo, what's happening is that:

1. Load `astro.config.ts` with Vite server, sees `./node_modules/.vite/deps/_metadata.json`, but hash mismatched so remove the file.
2. Start Astro dev server, which sees that there's no `./node_modules/.vite/deps/_metadata.json`.

This causes the Astro dev server startup to always re-optimize on step 2, which makes startup slow. This PR prevents step 1 from happening.

### Result

This removes 200ms of scanning and 600ms of prebundling for the Astro docs repo for warm startups.

## Testing

Manual testing in astro docs repo

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/shiny-roses-tap.md
+++ b/.changeset/shiny-roses-tap.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Disable Vite optimizer for sync and config loading. Improve first page load time for warm server startup.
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
Awesome!

PR #6915, opened 2023-04-26
bugfix: rewrite lit.js package resolution to use node export for Cloudflare SSR

### LINKED ISSUE #6887: @astrojs/lit breaks @astrojs/cloudflare builds

2.3.0

Cloudflare

yarn

Mac

Brave

### Describe the Bug

When deploying via GitHub, the astro project builds successfully, but fails at deploying the CF function.

No lit components are used in the project; simply adding lit to the integrations field in the config seems to break deployments.

The error from Cloudflare:
`Error: Failed to publish your Function. Got error: Uncaught ReferenceError: window is not defined at functionsWorker-xxxx.js:2140:10`

A snippet of what could be the offending code in the build:
```js
var ea = window,
  aa =
    ea.ShadowRoot &&
    (ea.ShadyCSS === void 0 || ea.ShadyCSS.nativeShadow) &&
    "adoptedStyleSheets" in Document.prototype &&
    "replace" in CSSStyleSheet.prototype,
  Fn = Symbol(),
  Sn = new WeakMap(),
  ta = class {
    constructor(t, a, u) {
      if (((this._$cssResult$ = !0), u !== Fn))
        throw Error(
          "CSSResult is not constructable. Use `unsafeCSS` or `css` instead.",
        );
      (this.cssText = t), (this.t = a);
    }
    get styleSheet() {
...
```

This is apparently from the Lit library itself.

I'll keep digging in to see what I can find out, but I'd greatly appreciated any help!

https://github.com/chaffinated/astro-lit-cloudflare-repro

### Participation

chaffinated:
## Changes

fix #6887

Rewrite lit packages entrypoint paths to work in Cloudflare SSR

## Testing

Added test in cloudflare integration

## Docs

No documentation was added because this should not directly affect developer experience.


packages/integrations/cloudflare/src/index.ts
```diff
@@ -97,7 +97,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 				await esbuild.build({
 					target: 'es2020',
-					platform: 'browser',
+					platform: 'node',
```
bluwy, line 100: This isn't right. The Cloudflare environment is not node, but workers, which the closest platform is `browser`. Otherwise you'd get `fs`, `path` etc modules bundled, which woudln't work in Cloudflare.

It doesn't look like lit supports worker environments too: https://github.com/lit/lit/issues/1983

chaffinated: >  Cloudflare environment is not node

While this is true, two different lit packages do not build correctly for Cloudflare with platform set to `browser`. The correct (node) versions of the broken lit packages actually compile correctly with this setting (and don't include node libs).

I've tried `platform: 'neutral', conditions: ['node']`; that attempts to build the node versions of `@lit/reactive-element` and `lit-html` (the problem packages) in a more V8-like context, but the build fails.

I'll see if there's any way to get Lit to build correctly with `platform: 'neutral'`, because, as of now, Lit is completely broken on Cloudflare.

chaffinated: @bluwy , I've updated this to something that might be more amenable to Cloudflare's Worker environment. Do you know if this would break any other integrations? The tests for this package passed, but I'm not sure how deep the tests go for integrations.

johannesspohr:
The same happened for `solid-js`, which actually contains a `worker` condition in the exports. I wonder if that should also be included in the conditions here?

Also, the vercel edge integration basically uses the same esbuild config and will probably suffer from the same issue (as described in #6989), I guess the same fix could be applied there as well.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -97,7 +97,9 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 				await esbuild.build({
 					target: 'es2020',
-					platform: 'browser',
+					platform: "neutral",
+          conditions: ['node'],
+          mainFields: ['main', 'module', 'browser'],
```
bluwy, line 102: Sorry for the late reply, we still can't use the `node` condition here though because Cloudflare's environment doesn't support node, but a V8-ish environment. It will break packages that are also packaged correctly. So we still need to keep the `platform: 'browser'` for correctness.

I'm not really sure how we could fix this, besides nudging lit to export a worker condition and we can set the worker condition here. This is what SvelteKit does too: https://github.com/sveltejs/kit/blob/fb68c253a59402df2b2b1b5db7a54fad504cdd32/packages/adapter-cloudflare/index.js#L48-L50

chaffinated: Apologies for my misunderstanding about the initial PR. Using 'node' as a platform or in the export conditions is not correct. That said...

@bluwy There's nothing correct about using `platform: 'browser'`, either. Neither Cloudflare Workers or V8 itself are actually browser-like environments, and using browser-like settings is very obviously breaking things. In my opinion, the most correct path is using `platform: 'neutral'`.

Building the lit library with the `node` condition works in Cloudflare; building it with the `browser` condition breaks it. It's definitely not correct from a nomenclature perspective, but this highlights an issue: **we can't expect other package maintainers to use package exports like the Astro team wants them to. We can dig our heels in about this stuff, but the only people that suffer are Astro users.**

How about this as a workaround:
1. In the integration package for each framework, we use esbuild to "prepackage" dependencies with their own conditions to a temp directory common to all integrations
2. In the Cloudflare adapter, we add an alias to the final esbuild step to prefer those prepackaged modules over what's in the node_modules directory

Maybe that's overkill, but it would allow each integration to set its own build conditions. This would also allow each integration to build according to how its package maintainers set it up, and would make for a better Astro user experience (which I think we can all agree is the goal here).

Is this a viable solution?

johannesspohr: I completely agree to your point about the suffering Astro users. So we should focus on making it working for them. And I think we should make it work out of the box for libraries using standards (e.g. the `worker` export condition). Until the adoption of that, I think it would be a good solution to:

- hardcode aliases for officially supported frameworks, we could maybe add unit tests that break when the aliasing is not required anymore to avoid cluttering
- offer a way to customize the aliases from the adapter config, so people can get arbitrary libraries working

In my opinion, the additional build step adds a lot of complexity, and if I got this correctly, the runtime isn't even known at this stage, so I'm not sure whether it even offers a good solution for that price.

chaffinated: > the additional build step adds a lot of complexity

> the runtime isn't even known at this stage

Both good points.

I've tried the alias with the lit-html package (one of the ones causing the problem with cloudflare), and esbuild's module resolution doesn't work; it doesn't seem to recognize that the `./node` folder is a module without relying on the "exports" property of the package.json file. I could have set it up wrong, so I might need someone else to try it out and confirm that behavior.

It would be ugly, but maybe there's a way to override certain packages' "exports" properties before build?

For example, if we could change the lit-html "node" export property name to "worker", that would solve our problem without adding too much complexity. And technically, that package's "node" export is more similar to a V8/worker environment than their browser export, so...

@johannesspohr What are your thoughts on this?

johannesspohr: @chaffinated I also couldn't get it to work with aliases, it seems like esbuild only accepts aliases that are also exported (`solid-js` actually exports everything like `/dist/*`, so that works)

By hooking into the resolve logic with a plugin, it is possible though (although I'm really not proud of this). So this will basically do a manual makeshift resolution of the module to an actual file, by completely bypassing esbuilds internal resolution. I'd love if we could specify something like "use node condition for package x", but I haven't found anything. There is an inbuilt resolve helper function, but that doesn't accept any differing conditions from what I can see.

```
let replaceLitPlugin = {
  name: 'replaceLit',
  setup(build) {
    build.onResolve({ filter: /^lit-html/ }, args => {
      console.log(args);
      return { path: path.join(process.cwd(), args.path.replace('lit-html', 'node_modules/lit-html/node') + (args.path.endsWith('.js') ? '' : '/lit-html.js')) }
    })
    build.onResolve({ filter: /^@lit\/reactive-element/ }, args => {
      console.log(args);
      return { path: path.join(process.cwd(), args.path.replace('@lit/reactive-element', 'node_modules/@lit/reactive-element/node') + (args.path.endsWith('.js') ? '' : '/reactive-element.js')) }
    })
  },
}```

Comment (bluwy) [reply packages/integrations/cloudflare/src/index.ts:102]: > @bluwy There's nothing correct about using `platform: 'browser'`, either. Neither Cloudflare Workers or V8 itself are actually browser-like environments, and using browser-like settings is very obviously breaking things. In my opinion, the most correct path is using `platform: 'neutral'`.

Webworker environments like Cloudflare are known to rely on "browser"-ish behaviour, and the convention is supported by Vite and many frameworks. Even Cloudflare wrangler [uses it](https://github.com/cloudflare/workers-sdk/blob/133c0423ccb4c2b35a1dd26157ce9a24c6a743bb/packages/wrangler/src/bundle.ts#L368-L423). We can't change that to fix the lit compatibility issue, lit would have fail to bundle in a vanilla Cloudflare wrangler setup too.

The alias would be the easiest workaround for now, but it's still reasonable to expect `lit` to properly export their files so it works on other platforms too. @johannesspohr's esbuild plugin is a fairly good implementation of the workaround. We need to resolve the import to an absolute path to bypass the `exports` field restriction.

Comment (chaffinated) [reply packages/integrations/cloudflare/src/index.ts:102]: @bluwy I'm not sure if you're being deliberately inflexible, or if you're actually hearing what I'm saying.

> Webworker environments like Cloudflare are known to rely on "browser"-ish behaviour, and the convention is supported by Vite and many frameworks.

Yes, but you are forgetting the *use case*.

Vite's tagline is literally "next generation frontend tooling". Vite uses these settings as a compromise, because it's *more likely* to work with their main use case.

What we're doing here is *server-side rendering* in a worker environment. If there's *any* usage of the DOM at all, the entire build will break.

Because of that, I still believe opting for a more conservative "neutral" platform is the right move. Supporting a subset of APIs is easier than supporting a superset.

> Even Cloudflare wrangler [uses it](https://github.com/cloudflare/workers-sdk/blob/133c0423ccb4c2b35a1dd26157ce9a24c6a743bb/packages/wrangler/src/bundle.ts#L368-L423).

Not seeing the `platform: 'browser'` you're talking about. Also, the `conditions` property relies on fallbacks; the first options, `workerd` and `worker` are the appropriate conditions we'd expect for package exports on workers.

>but it's still reasonable to expect lit to properly export their files so it works on other platforms too

These exports from the lit packages are literally for server-side rendering, so I don't know why you'd expect the platform export name to be "browser". "Worker" would be ideal, but I think what we're doing here is less common and more of an edge case than the majority of users of their package.

So, that means *we* need to be the one to adapt.

Comment (bluwy) [reply packages/integrations/cloudflare/src/index.ts:102]: > Not seeing the platform: 'browser' you're talking about.

https://esbuild.github.io/api/#platform

> By default, esbuild's bundler is configured to generate code intended for the browser.

Comment (chaffinated) [reply packages/integrations/cloudflare/src/index.ts:102]: @johannesspohr This seems like a good option for lit. Do you think this should be internal only, or should this be surfaced to an option in the adapter?

```typescript
export default defineConfig({
  //...
  adapter: cloudflare({
    mode: 'directory',
    alias: [
      { from: /^lit-html/, to: 'lit-html/node' },
      { from: /^@lit\/reactive-element/, to: '@lit/reactive-element/node' },
    ],
  }),
});

It seems silly that a user would need a custom option for lit to work out of the box; it would also require an update to the docs. That said, it might be a useful feature to have for other integrations.

What are your thoughts about having an internal list of known and required aliases (including those lit packages), but also surfacing that option to extend that list?

Comment (johannesspohr) [reply packages/integrations/cloudflare/src/index.ts:102]: That's exactly my thoughts, having the configuration option to also be able to rewrite the resolutions of other libraries. Imho we could hardcode the rules for lit, because I think it should be supported out of the box. However, we should make sure to cover the case where lit introduces the right exports, so it won't suddenly break again in that case.

My most recent iteration looked like that:
```js
let rewriteResolutionPlugin(rewrites) = {
  name: 'resolutionRewrite',
  setup(build) {
    for (rewrite of rewrites) {
      build.onResolve({ filter: rewrite.filter }, async ({path, ...options}) => {
        // avoid endless loops (this function will be called again by resolve())
        if (options.pluginData === "skip-rewrite") {
          return undefined; // continue with original resolution
        }
        // get original resolution
        const resolution = await build.resolve(path, {...options, pluginData: 'skip-rewrite'});
        resolution.path = filter.rewrite(resolution.path);
        return resolution;
      })
    }
  },
}

...
plugins: [rewriteResolutionPlugin([
            {filter: /^lit-html/, rewrite: (path) => path.includes('/lit-html/node/') ? path : path.replace('/lit-html/', '/lit-html/node/'),
            {filter: /^@lit\/reactive-element/, rewrite: (path) => path.includes('/@lit/reactive-element/node/') ? path : path.replace('/@lit/reactive-element/', '/@lit/reactive-element/node/'),
          ])
...
```

... but I'm happy with any variation. If you want feel free to include it into your PR, or I can open one tomorrow.

bluwy: Got swamped in notifications again 😢 Catching up, I understand that it's weird that Cloudflare Workers would pick the `browser` condition (and behaviour) even though it isn't actually one. I should've pointed earlier to Cloudflare's source code that it's intentional to follow the standard (they set) instead of the convention (by others).

I'm leaning towards hardcoding for now as we have first-party support for lit. We could revisit exposing an option if the issue becomes common.

I also noticed @johannesspohr opened https://github.com/withastro/astro/pull/7092 which is related. But for this PR we could resolve lit's issue only for now.

chaffinated: @bluwy Sticking with CloudFlare's standard settings sounds good, but I do foresee this coming up again, especially with 3rd party integrations.

I've gone ahead and added an optional property to the CloudFlare adapter's config object to support additional rewrites if necessary as well as including the appropriate export conditions to support "worked" and "worker" builds.

johannesspohr:
@chaffinated great! In #7092 I added a test for solid, do you think it makes sense to also include sth like that for lit?

chaffinated:
One more thing... Does anyone know where the `astro-scripts` command in the package.json scripts comes from? I'd like to be able to build and test locally, but I'm missing that piece of the puzzle.

chaffinated:
@johannesspohr Yup! And I just found the `astro-scripts` in the workspace, so I'll get those added ASAP

packages/integrations/cloudflare/src/index.ts
```diff
@@ -1,12 +1,13 @@
...
-import esbuild from 'esbuild';
+import esbuild, { type Plugin } from 'esbuild';
 import * as fs from 'fs';
 import * as os from 'os';
 import glob from 'tiny-glob';
 import { fileURLToPath, pathToFileURL } from 'url';
 
 type Options = {
 	mode: 'directory' | 'advanced';
+	rewrites?: RewriteResolutionOption[];
```
bluwy, line 10: Following my comment before, I don't think we should expose this option. We're only special-casing lit because it's a first-party integration. For other integrations that don't export right, they'd need to fix it themselves. Otherwise this option would only prolong the incorrect exports.

The rest of the changes looks great though.

chaffinated: @bluwy I agree with encouraging organizations to export more correctly. I've made the changes you requested as well as added tests.

Though, I want to point out again that the Lit.js team would argue they've exported everything correctly. It would be most correct if they had a `worker` export, but I don't know how common that use case is yet. Lit only breaks the build because the `browser` export has DOM references, *which is completely expected and justifiable*.

I generally don't agree that we should opt for de facto conventions (i.e. Cloudflare using `browser` as the build platform) over semantic conventions (i.e. SSR using `worker` or `neutral` platform). We'll end up with language that obfuscates rather than clarifies what's happening, and that will become more difficult to maintain.

Anyways, enough of my indignant protests.

bluwy:
The new changes looks good to me. I've also opened https://github.com/lit/lit/issues/3907 to hear what their thoughts are on this before we move on with the patch.

bluwy:
Looks like running two wrangler CLIs caused the CI to be flaky (even before my cleanup). Gonna have to investigate it a bit before merging.

chaffinated:
@matthewp I agree with this, but I think you may need to confer with @bluwy about that. Btw, who has the authority to make the final call on this?

JosefJezek:
@bluwy @chaffinated any progress?

matthewp:
Since this is quite a bit outdated, will need a new PR.

dengel29:
So what's the right way to go about this, since all related issues and PRs seem to have been closed after a lack of consensus or direction? As a user, how can I register with the Astro team that this would be a nice feature?

matthewp:
@dengel29 My comment here is still true: https://github.com/withastro/astro/pull/6915#issuecomment-1587704166

The Cloudflare adapter is community maintained by the maintainer is extremely active. If you brought up this issue and asked for such a config I bet he would be willing to work with you. That adapter lives here: https://github.com/withastro/adapters

PR #6916, opened 2023-04-26
[Content collections] Fast `getEntryBySlug()` lookup

bholmesdev:
## Changes

- Generate a `.astro/lookup-map.json`. This maps collection entry slugs to their local file path for glob imports.
- Update `getEntryBySlug()` to look up entries from this lookup map, instead of filtering across the entire collection.

## Testing

N/A - check that `getEntryBySlug()` tests still pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -14,6 +14,7 @@ import {
 
 type GlobResult = Record<string, () => Promise<any>>;
 type CollectionToEntryMap = Record<string, GlobResult>;
+type GetEntryImport = (collection: string, lookupId: string) => () => Promise<any>;
 
 export function createCollectionToGlobResultMap({
 	globResult,
@@ -28,9 +29,8 @@ export function createCollectionToGlobResultMap({
 		const segments = keyRelativeToContentDir.split('/');
 		if (segments.length <= 1) continue;
 		const collection = segments[0];
-		const entryId = segments.slice(1).join('/');
 		collectionToGlobResultMap[collection] ??= {};
-		collectionToGlobResultMap[collection][entryId] = globResult[key];
+		collectionToGlobResultMap[collection][key] = globResult[key];
 	}
 	return collectionToGlobResultMap;
 }
@@ -38,10 +38,10 @@ export function createCollectionToGlobResultMap({
 const cacheEntriesByCollection = new Map<string, any[]>();
 export function createGetCollection({
 	collectionToEntryMap,
-	collectionToRenderEntryMap,
+	getRenderEntryImport,
 }: {
 	collectionToEntryMap: CollectionToEntryMap;
-	collectionToRenderEntryMap: CollectionToEntryMap;
+	getRenderEntryImport: GetEntryImport;
 }) {
 	return async function getCollection(collection: string, filter?: (entry: any) => unknown) {
 		const lazyImports = Object.values(collectionToEntryMap[collection] ?? {});
@@ -64,7 +64,7 @@ export function createGetCollection({
 							return render({
 								collection: entry.collection,
 								id: entry.id,
-								collectionToRenderEntryMap,
+								renderEntryImport: await getRenderEntryImport(collection, entry.slug),
 							});
 						},
 					};
@@ -82,10 +82,10 @@ export function createGetCollection({
 
 export function createGetEntryBySlug({
 	getCollection,
-	collectionToRenderEntryMap,
+	getRenderEntryImport,
 }: {
 	getCollection: ReturnType<typeof createGetCollection>;
-	collectionToRenderEntryMap: CollectionToEntryMap;
+	getRenderEntryImport: GetEntryImport;
 }) {
 	return async function getEntryBySlug(collection: string, slug: string) {
 		// This is not an optimized lookup. Should look into an O(1) implementation
@@ -114,7 +114,7 @@ export function createGetEntryBySlug({
 				return render({
 					collection: entry.collection,
 					id: entry.id,
-					collectionToRenderEntryMap,
+					renderEntryImport: await getRenderEntryImport(collection, entry.slug),
 				});
 			},
 		};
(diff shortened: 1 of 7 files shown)


packages/astro/src/content/runtime.ts
```diff
@@ -114,7 +103,7 @@ export function createGetEntryBySlug({
 				return render({
 					collection: entry.collection,
 					id: entry.id,
-					collectionToRenderEntryMap,
+					renderEntryImport: await getRenderEntryImport(collection, slug),
```
bluwy, line 106:
```suggestion
					renderEntryImport: getRenderEntryImport(collection, slug),
```

packages/astro/src/content/runtime.ts
```diff
@@ -81,29 +81,18 @@ export function createGetCollection({
...
-		// as it's probably that people will have very large collections.
-		const entries = await getCollection(collection);
-		let candidate: (typeof entries)[number] | undefined = undefined;
-		for (let entry of entries) {
-			if (entry.slug === slug) {
-				candidate = entry;
-				break;
-			}
-		}
+		const entryImport = await getEntryImport(collection, slug);
```
bluwy, line 91: and here
```suggestion
		const entryImport = getEntryImport(collection, slug);
```

packages/astro/src/content/utils.ts
```diff
@@ -358,3 +387,59 @@ function search(fs: typeof fsMod, srcDir: URL) {
...
+
+		const { id, collection, slug: generatedSlug } = info;
+		filePathByLookupId[collection] ??= {};
+		const { slug: frontmatterSlug } = await contentEntryConfig.getEntryInfo({
+			fileUrl: pathToFileURL(filePath),
+			contents: await fs.promises.readFile(filePath, 'utf-8'),
+		});
+		const slug = getEntrySlug({ id, collection, generatedSlug, frontmatterSlug });
+		filePathByLookupId[collection][slug] = rootRelativePath(root, filePath);
+	}
```
bluwy, line 435: IIUC this is where we move most of the runtime work to build time. Since this does an `await` for each file, should we `.map` and use a `Promise.all`?

bluwy (commented):
Looks great! Got a few nits below. (GitHub pre-comments by initial reviews for some reason)

packages/astro/src/content/utils.ts
bholmesdev: We're usually wary of `Promise.all` for memory usage. But since this is a one-time cost, I'll give it a try. Also curious if a `frontmatterSlug` cache is worthwhile for dev server runtime performance

packages/astro/src/content/runtime.ts
bholmesdev: Ah, this `await` is needed actually! I got the type signature wrong 🙈

bluwy (approved):
Looks great! Thanks for double checking the types.

packages/astro/src/content/utils.ts
```diff
@@ -358,3 +387,90 @@ function search(fs: typeof fsMod, srcDir: URL) {
 	}
 	return { exists: false, url: paths[0] };
 }
+
+export async function getStringifiedLookupMap({
```
ematipico, line 391: Would it be possible to document this function a bit?

packages/astro/src/content/utils.ts
```diff
@@ -358,3 +387,90 @@ function search(fs: typeof fsMod, srcDir: URL) {
...
+		globOpts
+	);
+	let filePathByLookupId: {
+		[collection: string]: Record<string, string>;
+	} = {};
+
+	await Promise.all(
+		contentGlob.map(async (filePath) => {
+			const info = getEntryInfo({ contentDir, entry: filePath });
+			if (info instanceof NoCollectionError) return;
```
ematipico, line 424: Is there a reason why we are silencing an error?

packages/astro/src/content/utils.ts
```diff
@@ -358,3 +387,90 @@ function search(fs: typeof fsMod, srcDir: URL) {
...
+			filePathByLookupId[collection] ??= {};
+			const slug = await getEntrySlug({
+				id,
+				collection,
+				generatedSlug,
+				fs,
+				fileUrl: pathToFileURL(filePath),
+				contentEntryType,
+			});
+			filePathByLookupId[collection][slug] = rootRelativePath(root, filePath);
```
ematipico, line 438: Since `filePathByLookupId[collection]` is going to have a `[slug]` either way, should this work too?

```js
const slug = await getEntrySlug({
	id,
	collection,
	generatedSlug,
	fs,
	fileUrl: pathToFileURL(filePath),
	contentEntryType,
});
filePathByLookupId[collection] = {
	...filePathByLookupId[collection],
	[slug]: rootRelativePath(root, filePath)
}
```

If so, I like more this approach because it's easier to read compared to the previous (the `??=` was unexpected XD)

packages/astro/src/content/utils.ts
```diff
@@ -358,3 +387,90 @@ function search(fs: typeof fsMod, srcDir: URL) {
...
+	fs: typeof fsMod;
+	id: string;
+	collection: string;
+	generatedSlug: string;
+	fileUrl: URL;
+	contentEntryType: Pick<ContentEntryType, 'getEntryInfo'>;
+}) {
+	const { slug: frontmatterSlug } = await contentEntryType.getEntryInfo({
+		fileUrl,
+		contents: await fs.promises.readFile(fileUrl, 'utf-8'),
```
ematipico, line 466: If we can't read the file, an error will be fired. Do you know if we handle the error case already?

ematipico (commented):
I know very little about the logic of the content collections, but I wish the PR description would have better explained the new change. 

I left some comments, code-wise.

packages/astro/src/content/utils.ts
bholmesdev: Ah, this shouldn't be an error in hindsight. Just means the entry isn't inside a collection directory. I'm refactoring this implementation to something more readable in the data collections PR. Will add a clarifying comment here.

packages/astro/src/content/utils.ts
bholmesdev: Ugh every time I use `??=` people complain 😅  Yeah, this should work too but it's 50% less fun

packages/astro/src/content/utils.ts
bholmesdev: Ah, this definitely assumed errors were an impossible case. Could throw an unknown content collection error maybe?

packages/astro/src/content/utils.ts
bholmesdev: Sure! Will also move to the `vite-plugin-content-virtual-mod` file where it's used; confusing to make a reusable util

packages/astro/src/content/utils.ts
ematipico: Your assumptions are entirely valid! Although, you never know what the users/rights of the scripts that install/create the files/directory. 

My golden rule is to assume that when working with files, the code might need to have the rights proper rights to run the operation.

ematipico (approved):

PR #6911, opened 2023-04-25
Upgrade turbo

bluwy:
## Changes

- Upgrade turbo to latest.
- Improve build prime: Build packages once and will be shared in all OS tests.
- Remove workspace package naming conflict to make turbo work.

## Testing

Existing tests should pass

## Docs

n/a. internal test changes.

slightly related docs optimization: https://github.com/withastro/docs/pull/3102

PR #6917, opened 2023-04-27
Fix changeset version run

bluwy:
## Changes

Changeset is currently failing as I had removed the `package.json` `workspaces` field that it used to use in #6911. I don't think using that field is right, so I've tweak the changeset config, and provide a package name for one fixture to fix it.

## Testing

Tested `pnpm changeset version --dry-run` to make sure there's no errors.

## Docs

n/a. changeset changes.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,7 +6,7 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@example/*", "@test/*"],
+  "ignore": ["!(astro|create-astro|@astrojs/*|astro-scripts)"],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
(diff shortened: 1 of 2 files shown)


.changeset/config.json
```diff
@@ -6,7 +6,7 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@example/*", "@test/*"],
+  "ignore": ["!(astro|create-astro|@astrojs/*|astro-scripts)"],
```
bluwy, line 9: I switched to using an opt-in syntax so we don't accidentally publish other package names in the future. This pattern is referenced from https://github.com/sveltejs/kit/blob/fb68c253a59402df2b2b1b5db7a54fad504cdd32/.changeset/config.json#L9

Princesseuh (approved):

PR #6921, opened 2023-04-27
Speed up docs smoke tests

bluwy:
## Changes

Uses a new `PUBLIC_TWO_LANG` field from https://github.com/withastro/docs/pull/3102 to speed up docs smoke tests. It will only generates pages for two languages to build faster, as mostly the content varies and not the functionality.

## Testing

Existing smoke tests should pass and be faster.

## Docs

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,7 +240,7 @@ jobs:
         run: pnpm run build
 
       - name: Remove docs translations except for English and Korean
-        run: find ./smoke/docs/src/content/docs ! -name 'en' ! -name 'ko' -type d -depth 1 -exec rm -rf {} +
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM!

I'll be curious to see how much faster this makes things for you.

delucis:
I'm not sure what the baseline is but this doesn't look much faster than a full build to me — a lot of time is spent generating the bundle before rendering and the language changes here don't impact that, only what follows. Maybe you could update the smoke test to `rm -rf` everything in `src/content/docs` that isn't English or Korean?

bluwy:
I'm seeing a speed improvement on Windows from 774s to 246s, but you're right that the bundling is still taking most of the time. I didn't thought about `rm -rf` it though, sounds like a good idea! Will try it.

bluwy:
It seems to be improved again and now takes 126s!

matthewp (approved):

Issue #5084, opened 2022-10-13
Heading elements in MDX not being passed to Astro.props 


fmonper1:
### What version of `astro` are you using?

1.5.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

Npm

### What operating system are you using?

Mac

### Describe the Bug

I am authoring some MDX pages where I render a list of tables with a heading for each of the tables using an .astro component.

I would expect this headings from the astro components to be included in the headings property of Astro.props in my layout, but the headings are not included in the props and thus not appear on the table of content.

![image](https://user-images.githubusercontent.com/16008141/195712900-a6876a63-286f-4e8c-96fe-8976c0bc268b.png)

I would expect the `this is not` heading to appear on the table of content but it's not appearing.

Noktasizi from [this discord discussion ](https://discord.com/channels/830184174198718474/1019713903481081876) might be able to provide more content

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qhpm6w?file=src%2Fpages%2Fen%2Fintroduction.mdx,astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

kcole93:
To add a bit more detail based on the discussion taking place on the Discord server, I believe this behavior is unexpected because Astro's built-in `collect-headings` plugin is described as a Rehype plugin that would presumably gather headings from the HTML rendered from a MD or MDX page.

However, in practice *only headings defined in Markdown syntax (e.g., `## Heading 2`) are correctly picked up* by the built-in plugin while headings defined as HTML (e.g., `<h2>Heading 2</h2>`) are not identified or given an ID. This behavior also seems to hold true when using a pipeline of Rehype plugins like `rehypeSlug`, `rehypeAutolinkHeadings` and `toc`. A minimal production of this behavior can be found [here](https://stackblitz.com/edit/github-t1qyad?file=src/pages/example.mdx).

Given this information, it seems like there's potentially something funky going on in the processing pipeline that results in elements defined via HTML markup not being sent further down the line.

matthewp:
@ematipico this is related to the MDX integration in `packages/integrations/mdx`.

tudorpopams:
This issue is happening for `md` files as well, in pre 2.0 versions (currently using 1.6).

imchell:
[...]

PR #6922, opened 2023-04-27
fix: collect HTML headings in MDX documents

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/5084

I had to pull/update new libraries to new updated types of the `rehype` AST.

Unfortunately types there are very weird, and I am not very familiar with the logic of the function, so I had to some downcast - `as any` - in order to keep the existing logic and avoid TS to scream at me.

## Testing

Added new test cases

### DIFF
--- a/packages/integrations/mdx/test/fixtures/mdx-get-headings/src/pages/test-html.mdx
+++ b/packages/integrations/mdx/test/fixtures/mdx-get-headings/src/pages/test-html.mdx
@@ -1,0 +1,6 @@
+<h1>Heading test</h1>
+<h2>Subsection 2</h2>
+<h3>Subsection 3</h3>
+<h4>Subsection 4</h4>
+<h5>Subsection 5</h5>
+<h6>Subsection 6</h6>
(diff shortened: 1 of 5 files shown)


delucis:
Oh, I’m only now seeing the original issue. We actually rely on the different behaviour for `<h2>text</h2>` vs `## text` to exclude headings from table of contents in docs (we can use `<h2>` to avoid a heading being included in Astro’s `headings`). I’m guessing this “fix” would break our usage?

ematipico:
> Oh, I’m only now seeing the original issue. We actually rely on the different behaviour for `<h2>text</h2>` vs `## text` to exclude headings from table of contents in docs (we can use `<h2>` to avoid a heading being included in Astro’s `headings`).

Oh, I wasn't aware of this usage! You raised a good point: is it a fix or a feature? 

> I’m guessing this “fix” would break our usage?

Yes, and it would potentially emit HTML that would break SEO.

delucis:
> is it a fix or a feature?

I don’t know! I’ve sometimes thought it’s a bit weird that two things that are kind of equivalent behave different and thought that could confuse users. But it is helpful in our case. Docusaurus has the same pattern and even recommend using `<h>` to avoid inclusion in the table of contents in their docs: https://docusaurus.io/docs/markdown-features/toc#customizing-table-of-contents-generation

PR #6923, opened 2023-04-27
Fix changeset publish

bluwy:
## Changes

Use `package.json` `workspaces` field to configure changesets like before. And make sure test packages have `{"private": true}`

## Testing

Ran `pnpm changeset version --dry-run`, `pnpm changeset publish --dry-run`, `pnpm changeset`

## Docs

n/a

### DIFF
--- a/packages/astro/performance/package.json
+++ b/packages/astro/performance/package.json
@@ -2,6 +2,7 @@
   "name": "@test/performance",
   "version": "1.0.0",
   "description": "",
+  "private": true,
   "main": "index.js",
   "scripts": {
     "benchmark": "node ./content-benchmark.mjs"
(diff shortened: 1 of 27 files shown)


.changeset/config.json
```diff
@@ -6,7 +6,6 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["!(astro|create-astro|@astrojs/*|astro-scripts)"],
```
bluwy, line 9: This can be removed because the `workspaces` field only captures packages that we want to publish, otherwise changeset would error on unused ignores.

Princesseuh (approved):

PR #6926, opened 2023-04-27
Refactored test suite for use outside Astro itself

RaphaelBossek:
## Changes

- What does this change?
Allows foreign projects to have thier own Astro fixture test suite.

## Testing
The Astro test suite works as expected, of course.

This implementation is used by the https://github.com/RaphaelBossek/astro-static-cms 

## Docs

If the chagene would be accepted an howto would be the next step. As exptected we have a lot of examples in the Astro repo.


RaphaelBossek:
Will fix the build issues and come back.

PR #6928, opened 2023-04-28
Support `--host` for Node adapter preview

### LINKED ISSUE #6661: host: true with Node adapter

2.1.7

node

pnpm

linux

firefox

### Describe the Bug

Using the following configuration:

```json
{
  server: {
    host: true // "0.0.0.0",
  },
  integrations: [tailwind()],
  output: 'server',
  adapter: node({
    mode: 'standalone',
  }),
}
```

Instead of listening on all addresses, it says "http://undefined:3000" with "astro preview".

https://codesandbox.io/p/sandbox/thirsty-ride-r3726r?on=codesandbox&file=%2Fpackage.json&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A15%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A15%7D%5D

### Participation

JerryWu1234:
## Changes

- Closes #6661
- Supports `--host` flag and/or respects `server.host` setting when using `astro preview` 

## Testing

Tested manually

## Docs

Bug fix! We already say that the `server.host` setting changes `astro dev` and `astro preview` behavior.


bluwy:
I don't think this is the right fix, otherwise this PR makes `--host` only listens to one network host, instead of all available network host. Plus it gives preference to ipv4 only too. I think the issue is with the logging, and you should be able to get the information after the server starts with `server.server.address()` and log with that information.

wulinsheng123:
> I don't think this is the right fix, otherwise this PR makes `--host` only listens to one network host, instead of all available network host. Plus it gives preference to ipv4 only too. I think the issue is with the logging, and you should be able to get the information after the server starts with `server.server.address()` and log with that information.

`server.server.address()` will return an `undefined`.

https://github.com/wulinsheng123/astro/blob/2a33667e94ba60dd40322ad468762743dc001c15/packages/integrations/node/src/http-server.ts#L17
This host always gets `undefined` when we have passed `--host.`  in nodejs adapter 
So I even pass a network address

bluwy:
Sorry for the late reply. When I log before [here](https://github.com/withastro/astro/blob/34202616c2ad0bd4b7a808217718ee74d1b20eae/packages/integrations/node/src/preview.ts#L71-L72) with `console.log(server.server.address())`, I got this though:

```
> astro preview "--host"

{ address: '::', family: 'IPv6', port: 3000 }
Preview server listening on http://undefined:3000
```

nodejs also documented it [here](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback) so I'm not sure why it isn't working. But either way I think we should still leave `host` `undefined` so the server listens on all active connections rather than a specific one. We can move the fix solely in the nodejs integration only.

Maybe we can re-implement how Vite [does it here](https://github.com/vitejs/vite/blob/d09bbd093a4b893e78f0bbff5b17c7cf7821f403/packages/vite/src/node/utils.ts#L892-L914)?

wulinsheng123:
> Sorry for the late reply. When I log before [here](https://github.com/withastro/astro/blob/34202616c2ad0bd4b7a808217718ee74d1b20eae/packages/integrations/node/src/preview.ts#L71-L72) with `console.log(server.server.address())`, I got this though:
> 
> ```
> > astro preview "--host"
> 
> { address: '::', family: 'IPv6', port: 3000 }
> Preview server listening on http://undefined:3000
> ```
> 
> nodejs also documented it [here](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback) so I'm not sure why it isn't working. But either way I think we should still leave `host` `undefined` so the server listens on all active connections rather than a specific one. We can move the fix solely in the nodejs integration only.
> 
> Maybe we can re-implement how Vite [does it here](https://github.com/vitejs/vite/blob/d09bbd093a4b893e78f0bbff5b17c7cf7821f403/packages/vite/src/node/utils.ts#L892-L914)?

actually, I simulate the code of Vite. 
Let me conclude.
1. move the code to the node adapter
2. re-implement code that got a network address

wulinsheng123:
image

wulinsheng123:
image

wulinsheng123:
image

packages/integrations/node/src/get-network-address.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const wildcardHosts = new Set([
+  '0.0.0.0',
+  '::',
+  '0000:0000:0000:0000:0000:0000:0000:0000',
+])
+type Protocol = 'http' | 'https'
+
+export function getNetworkAddress(protocol: Protocol = 'http', hostname: string, port:number, base?:string) {
```
bluwy, line 14: Should the `hostname` be `hostname: string | undefined` since you're handling `undefined` in the code too?

---

Also it would be good to comment where this function is referenced from (Vite)

packages/integrations/node/src/preview.ts
```diff
@@ -67,9 +69,23 @@ const preview: CreatePreviewServer = async function ({
...
+	
+	if(host === undefined ){
+		// eslint-disable-next-line no-console
+		console.log(
+			`Preview server listening on 
+			\t
+			local: ${address.local[0]}
+			\t
+			network: ${address.network[0]}
+			` );
```
bluwy, line 82: It would be nice if this is formatted better, something like:
```
Preview server listening on:
  local:   http://..
  network: http://...
```

packages/integrations/node/src/standalone.ts
```diff
@@ -53,11 +54,24 @@ export default function startServer(app: NodeApp, options: Options) {
...
+			\t
+			local: ${address.local[0]}
+			\t
+			network: ${address.network[0]}
+			` );
+	}else{
+		// eslint-disable-next-line no-console
+		console.log(`Preview server listening on \t 
+		${address.local[0]}`);
+	}
```
bluwy, line 74: Similar here

packages/integrations/node/test/url-protocol.test.js
```diff
@@ -2,20 +2,41 @@ import { TLSSocket } from 'tls';
...
+		});
+		after(async () => {
+			await devPreview.stop();
+		});
+		it('test host is true ', async () => {
+			const address = getNetworkAddress('http', undefined, 3000)
+			const res = await fetch(address.network[0])
+			expect(res.status).to.equal(200);
+		});
+	})
```
bluwy, line 39: Maybe we can move this test to `node-middleware.test.js`? So we can share the preview server there so it tests faster

.changeset/soft-pianos-happen.md
```diff
@@ -0,0 +1,6 @@
+---
+'astro': minor
```
bluwy, line 2: I think we can remove this now.

natemoo-re (commented):
These code changes look alright to me, but I'm not familiar enough with Node's networking logic to know for sure if this is the correct approach.

Would love @bluwy to re-review!

natemoo-re (approved):
Cleaned up the formatting, LGTM!

matthewp (approved):

beltonk:
astro preview "--host" is not working
npm run preview -- --host 
also does not expose host

PR #6929, opened 2023-04-28
Upgrade undici to v5.22.0

bluwy:
## Changes

Bump undici to hopefully reduce the `fetch failed` errors that sometimes happen in CI.

## Testing

Existing tests should pass.

## Docs

n/a

### DIFF
--- a/.changeset/polite-pears-kneel.md
+++ b/.changeset/polite-pears-kneel.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/telemetry': patch
+'@astrojs/webapi': patch
+---
+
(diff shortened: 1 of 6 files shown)


ematipico (approved):

Princesseuh (commented):
any reasons it added a new dep instead of replacing 5.20.0?

bluwy:
It looks like some Cloudflare dependencies still rely on undici 5.20.0. I tried `pnpm update undici -r` but it doesn't seem to update the internal nested deps, so not touching those for now to keep the scope down.

Princesseuh (approved):

PR #6930, opened 2023-04-28
Update dependencies to remove peer dep warnings

bluwy:
## Changes

- Update dependencies to prevent the postcss peer dep warnings on install.
- Also update the `@sveltejs/vite-plugin-svelte` dep as that removes the stats log clutter in the CI logs

There's one remaining peer dep warning though:

```
 WARN  Issues with peer dependencies found
packages/integrations/svelte
└─┬ svelte2tsx 0.5.11
  └── ✕ unmet peer typescript@^4.1.2: found 5.0.2
```

We need to upgrade to svelte2tsx v6 to fix it, but that bumps the peer dep of `svelte` to 3.55.0. Our svelte peer dep is 3.54.0. So we'd need to cut a major if we want to fix it, which doesn't seem worth it for now.

## Testing

Existing tests should pass.

## Docs

n/a. deps update

### DIFF
--- a/.changeset/wise-steaks-wash.md
+++ b/.changeset/wise-steaks-wash.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/tailwind': patch
+'@astrojs/svelte': patch
+---
+
+Update dependencies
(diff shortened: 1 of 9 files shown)


Princesseuh:
fwiw, the `typescript` dep of `svelte2tsx` in our context is optional, we only use it in contexts where we the user provide TypeScript themselves

ematipico (approved):

Issue #6899, opened 2023-04-24
Certain Shiki themes unsupported by the Astro Code component 


ohansemmanuel:
### What version of `astro` are you using?

2.3.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

The `<Code />` component does not seem to support all Shiki themes as documented. 

![CleanShot 2023-04-24 at 08 05 22](https://user-images.githubusercontent.com/10930234/233913761-955cfea3-bd91-4264-a0b3-c3875921214e.png)

![CleanShot 2023-04-24 at 08 07 35](https://user-images.githubusercontent.com/10930234/233913828-8e6059d5-9c8d-401a-8f61-26056c6fbbfd.png)

From the Shiki [theme list](https://github.com/shikijs/shiki/blob/main/docs/themes.md#all-themes),  the following themes aren't supported: 

-  'material-theme-darker'
  -  'material-theme-lighter'
  - 'material-theme-ocean'
  -  'material-theme-palenight'
  -  'material-theme'

Please see the reproduction link below

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rakqbl?file=src/pages/index.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

ohansemmanuel:
Took a look at the source, and seems the theme names are wrong. Perhaps this was updated on Shiki's end. See https://github.com/withastro/astro/blob/main/packages/astro/components/shiki-themes.js#L12

![CleanShot 2023-04-24 at 08 20 55](https://user-images.githubusercontent.com/10930234/233915127-db529251-c8b9-4553-a57f-26c3a37535d1.png)

ohansemmanuel:
Shiki change: https://github.com/shikijs/shiki/commit/391713be6dec7b9484b06da0bc5bf27c33e5fee6

ematipico:
Yeah, I tried to upgrade shiki some time ago, and there few breaking changes in the typescript types.

PR #6932, opened 2023-04-28
Upgrade shiki to v0.14.1

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/6899

- Upgrade shiki
- Add compat code to handle shiki breaking changes
  - AFAICT the only breaking change is the theme name change, so I added compat to remap the old theme name to the newer one.
- Clean up shiki theme and language codegen
- Add theme name to `pre` tag class, e.g. `<pre class="astro-code github-dark">` (feature by shiki itself)

NOTE: while this could break theme types in `shikiConfig`, they will still work in runtime so I don't think this is a breaking change.

## Testing

Existing tests should pass. I also tested the `<Code` component locally so it still works, including the compat stuff.

## Docs

The only new thing to document is probably the theme name in the `pre` tag, but I'm not sure if it needs to be documented specifically.


packages/astro/scripts/shiki-gen-languages.mjs
bluwy, line 1: This file used to generate a language handler function for each language, causing a lot of line generation. I've refactored it to a single `handleLang` function so it can be shared.

packages/astro/scripts/shiki-gen-languages.mjs
```diff
@@ -4,31 +4,34 @@ const dir = await fs.promises.readdir('packages/astro/node_modules/shiki/languag
...
 		return undefined;
 	}
-})`,
-	];
-});
-let code = `import { BUNDLED_LANGUAGES } from 'shiki';
+}
 
+// prettier-ignore
 export const languages = {`;
```
bluwy, line 31: Used `// prettier-ignore` here as the codegen code is more readable without prettier's line breaks.

packages/astro/scripts/shiki-gen-themes.mjs
```diff
@@ -2,18 +2,36 @@ import fs from 'fs';
...
+};
+
+let code = `\
+/**
+ * This file is prebuilt from packages/astro/scripts/shiki-gen-themes.mjs
+ * Do not edit this directly, but instead edit that file and rerun it to generate this file.
+ */
+
+// prettier-ignore
+export const themes = {`;
```
bluwy, line 27: Also used `// prettier-ignore` here as the codegen code is more readable without prettier's line breaks.

packages/astro/scripts/shiki-gen-themes.mjs
```diff
@@ -2,18 +2,36 @@ import fs from 'fs';
...
-		code += ',';
-	}
-	code += `\n\t'${key}': () => ${imp}`;
-	i++;
+	code += `\n\t'${key}': () => ${imp},`;
+}
+code += `\n\t// old theme names for compat`;
+for (const oldName in compatThemes) {
+	code += `\n\t'${oldName}': () => ${toThemeImport(compatThemes[oldName])},`;
 }
```
bluwy, line 35: This line is where we add the additional compat code for shiki

packages/markdown/remark/src/remark-shiki.ts
```diff
@@ -10,10 +10,30 @@ import type { ShikiConfig } from './types.js';
...
 const highlighterCacheAsync = new Map<string, Promise<shiki.Highlighter>>();
 
+// Map of old theme names to new names to preserve compatibility when we upgrade shiki
+const compatThemes: Record<string, string> = {
+	'material-darker': 'material-theme-darker',
+	'material-default': 'material-theme',
+	'material-lighter': 'material-theme-lighter',
+	'material-ocean': 'material-theme-ocean',
+	'material-palenight': 'material-theme-palenight',
+};
```
bluwy, line 20: Had to use duplicated code here because the entire file seems to be duplicated. I could see deduplicating it in a separate PR in the future.

bluwy:
This is ready for review now. I added additional comments on some changes above that are minimized to keep this page clean.

matthewp (approved):

Issue #6709, opened 2023-03-30
Astro loads two copies of SolidJS


lxsmnsyc:
### What version of `astro` are you using?

2.1.8

### Are you using an SSR adapter? If so, which one?

None (if @astrojs/node doesn't count)

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

![image](https://user-images.githubusercontent.com/4783372/228759116-9c35efe9-c3b1-4d00-867d-95559cda1b84.png)

As you see in the image, Astro loads two copies of SolidJS, one through the `file:///` URL (see `createResource`), the other with the  POSIX path (see `createMemo`).

Based on what I know with Vite's module resolution, since this is close to the browser behavior, two copies are loaded because they come from different URLs, despite pointing to the same file locally. This behavior breaks SolidJS packages that uses SolidJS APIs that relies on the existence of a root (e.g. `createUniqueId`, `createResource`). The code basically thinks that the APIs are called "outside a root".

I do think this same issue might be prevalent to other integrations that rely on the same mechanism (e.g. React), but I don't have any evidence for that yet. 

Another potential issue is how `@astrojs/solid-js` has [SolidJS as both a peer dependency and a dependency](https://github.com/withastro/astro/blob/main/packages/integrations/solid/package.json#L41-L45), I do think it should be both a peer and a dev dependency?

### Link to Minimal Reproducible Example

https://github.com/lxsmnsyc/astro-solid-spa-nested

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
This fixes it:

```js
    ssr: {
      noExternal: ['solid-swr-store']
    }
```

Since this is an ecosystem package I'll add this to the Solid integration.

PR #6934, opened 2023-04-28
Update Solid ecosystem package logic to include packages with peerDep…

matthewp:
… of Solid

## Changes

- Closes https://github.com/withastro/astro/issues/6709

## Testing

Tested in the example project

## Docs

N/A

### DIFF
--- a/.changeset/purple-lamps-sleep.md
+++ b/.changeset/purple-lamps-sleep.md
@@ -2,4 +2,4 @@
 "@astrojs/solid-js": patch
 ---
 
-Update Solid ecosystem package logic to include packages with peerDep of Solid
+Allow Solid ecosystem packages to not need special export map configuration. By default Solid is now treated as an external package in SSR, so any other dependent packages will receive the same instance.
--- a/.changeset/purple-lamps-sleep.md
+++ b/.changeset/purple-lamps-sleep.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/solid-js": patch
+---
+
+Update Solid ecosystem package logic to include packages with peerDep of Solid
--- a/packages/integrations/solid/src/dependencies.ts
+++ b/packages/integrations/solid/src/dependencies.ts
@@ -8,7 +8,7 @@ export async function getSolidPkgsConfig(isBuild: boolean, astroConfig: AstroCon
 		isBuild,
 		viteUserConfig: astroConfig.vite,
 		isFrameworkPkgByJson(pkgJson) {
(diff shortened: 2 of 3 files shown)


packages/integrations/solid/src/dependencies.ts
```diff
@@ -8,7 +8,7 @@ export async function getSolidPkgsConfig(isBuild: boolean, astroConfig: AstroCon
 		isBuild,
 		viteUserConfig: astroConfig.vite,
 		isFrameworkPkgByJson(pkgJson) {
-			return containsSolidField(pkgJson.exports || {});
+			return containsSolidField(pkgJson.exports || {}) || !!(pkgJson.peerDependencies || {})['solid-js'];
```
Princesseuh, line 11: The logic here is a tiny bit hard to read with the numerous pipe. Reviewing this, I was actually confused that the code added was in the parameter of `containsSolidField` and it took me a few reads before getting the logic and that it wasn't a parameter. Maybe a comment could help?

matthewp: Yeah, that's very true. I'm actually going to hold off on this one until I have a chance to speak with @bluwy because maybe making all packages with a peerDep isn't the right call here. Might just want to hardcode this one instead.

bluwy: We'd probably want to align with `vite-plugin-solid`'s [implementation](https://github.com/solidjs/vite-plugin-solid/blob/66955a81de1e6b84e02271e4f8603500b45656cb/src/index.ts#L301-L308) too, so I think we should avoid this as the package would also have issues in other Vite SSR Solid setups.

Maybe a solution is to remove the `solid-js` string here:

https://github.com/withastro/astro/blob/dc062f6695ce577dc569781fc0678c903012c336/packages/integrations/solid/src/index.ts#L50

Or only apply it in dev.

matthewp: Sounds reasonable since that's what vite-plugin-solid does. I've update the PR to remove `'solid'` as a noExternal, seems to work from my testing.

.changeset/purple-lamps-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/solid-js": patch
+---
+
+Update Solid ecosystem package logic to include packages with peerDep of Solid
```
bluwy, line 5: This probably needs to be changed with the new PR changes.

bluwy (approved):
LGTM. Looks like the tests pass too.

Issue #6900, opened 2023-04-24
Errors not caught in imported components / streaming broken chunks


povilas-juodzevicius:
### What version of `astro` are you using?

2.3.0

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

Currently Astro doesn't handle errors that are thrown in imported components. The node adapter tries rendering the component, but throws out only an error: `net::ERR_INVALID_CHUNKED_ENCODING 200 (OK)`. So the server response is 200, though the runtime error is not caught. 

The problem seems to be in streaming chunks where one of them is broken. The node adapter's `asyncIterator` [function](https://github.com/withastro/astro/blob/main/packages/integrations/node/src/response-iterator.ts#L196) can be extended to catch these errors. It ends up in the `dist/server/entry.mjs`.

It would be best if we could either disable streaming or have some error catching in the adapter code so not to render a blank screen.  

### Link to Minimal Reproducible Example

https://github.com/povilas-juodzevicius/astro-runtime-error-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Yes, this is a problem with the Node adapter I believe.

PR #6935, opened 2023-04-28
Catch errors that occur within the stream in the Node adapter

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6900

## Testing

- Test added

## Docs

N/A, bug fix

### DIFF
--- a/packages/integrations/node/src/middleware.ts
+++ b/packages/integrations/node/src/middleware.ts
@@ -51,8 +51,12 @@ async function writeWebResponse(app: NodeApp, res: ServerResponse, webResponse: 
 
 	res.writeHead(status, Object.fromEntries(headers.entries()));
 	if (webResponse.body) {
-		for await (const chunk of responseIterator(webResponse) as unknown as Readable) {
-			res.write(chunk);
+		try {
+			for await (const chunk of responseIterator(webResponse) as unknown as Readable) {
+				res.write(chunk);
+			}
+		} catch(err: any) {
+			res.write(err + '');
 		}
 	}
 	res.end();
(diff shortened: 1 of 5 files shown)


packages/integrations/node/src/middleware.ts
```diff
@@ -51,8 +51,12 @@ async function writeWebResponse(app: NodeApp, res: ServerResponse, webResponse:
...
 	res.writeHead(status, Object.fromEntries(headers.entries()));
 	if (webResponse.body) {
-		for await (const chunk of responseIterator(webResponse) as unknown as Readable) {
-			res.write(chunk);
+		try {
+			for await (const chunk of responseIterator(webResponse) as unknown as Readable) {
+				res.write(chunk);
+			}
+		} catch(err: any) {
+			res.write(err + '');
```
Princesseuh, line 59: Is it safe to write the error like this? In dev we usually don't care, but in prod you might not want errors to show in the user's browser

MoustaphaDev (approved):
LGTM!

packages/integrations/node/src/middleware.ts
matthewp: I was thinking this is what we did if the error happened before the response, but I guess not: https://github.com/withastro/astro/blob/main/packages/astro/src/core/app/index.ts#L210

So maybe log and write out `'Internal server error'` instead?

Princesseuh: That's what I would personally go for, with a 500 status code

matthewp: Can't set the status code at this point, the response is already sent. So this is just a matter of what goes in the HTML. Really the application should be handling errors themselves, this is a fallback case.

Princesseuh: Ah yes, fair. Well, I stil think show Internal Server Error is safer, just in case

PR #6940, opened 2023-04-28
Show injected custom 404 route in dev

### LINKED ISSUE #6937: Custom 404 page injected using `injectRoute` not used

2.3.2

None

npm/pnpm

macOS / Stackblitz

Firefox

### Describe the Bug

When a project includes `src/pages/404.astro`, this route gets used as a custom 404 page, including in dev.

However, if an integration uses the `injectRoute` hook to inject a custom page at `/404`, Astro does not treat this the same way: the injected page is only available at `/404`, other 404 errors continue to show the standard Astro 404 screen during dev. (The injected route _does_ get correctly built as `404.html` though, so should work in production on static hosts that support that pattern like Netlify.)

I’d expect injecting a route to `/404` to allow an integration to set the 404 users see during dev.

https://stackblitz.com/edit/github-3u16qb?file=astro.config.mjs

### Participation

delucis:
## Changes

- Fixes #6937
- Updates `getCustom404Route` logic to find 404s added via `injectRoute`
- As a side effect, also adds support for `src/pages/404.html`, which we don’t have currently.

## Testing

Added a fixture + test like we have for the current custom 404 strategies and also a dev server unit test.

I also applied the same changes in a patch inside a monorepo project to test the logic when the injected route comes from a separate package.

## Docs

Bug fix, no docs needed.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -35,11 +35,12 @@ interface MatchedRoute {
 	mod: ComponentInstance;
 }
 
-function getCustom404Route({ config }: AstroSettings, manifest: ManifestData) {
-	// For Windows compat, use relative page paths to match the 404 route
-	const relPages = resolvePages(config).href.replace(config.root.href, '');
-	const pattern = new RegExp(`${appendForwardSlash(relPages)}404.(astro|md)`);
-	return manifest.routes.find((r) => r.component.match(pattern));
+function getCustom404Route(
+	{ config }: AstroSettings,
+	manifest: ManifestData
+): RouteData | undefined {
+	const pathname = config.trailingSlash === 'always' ? '/404/' : '/404';
+	return manifest.routes.find((r) => r.pattern.test(pathname));
 }
 
 export async function matchRoute(
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -1,6 +1,6 @@
 import type http from 'http';
 import mime from 'mime';
-import type { AstroSettings, ComponentInstance, ManifestData, RouteData } from '../@types/astro';
+import type { ComponentInstance, ManifestData, RouteData } from '../@types/astro';
 import type {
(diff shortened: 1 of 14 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -35,11 +33,9 @@ interface MatchedRoute {
...
 }
 
-function getCustom404Route({ config }: AstroSettings, manifest: ManifestData) {
-	// For Windows compat, use relative page paths to match the 404 route
-	const relPages = resolvePages(config).href.replace(config.root.href, '');
-	const pattern = new RegExp(`${appendForwardSlash(relPages)}404.(astro|md)`);
-	return manifest.routes.find((r) => r.component.match(pattern));
+function getCustom404Route(manifest: ManifestData): RouteData | undefined {
+	const route404 = /^\/404\/?$/;
+	return manifest.routes.find((r) => route404.test(r.route));
```
delucis, line 38: I think this is safe? Previously we only supported `404.astro`, `404.md`, `404.mdx`, and `404.mdoc` as custom 404 routes. Those would all have a `route` of `/404` or `/404/`, so are still supported. This change adds custom 404 support for `src/pages/404.html` or for `injectRoute('404', ...)` in dev.

packages/astro/test/units/dev/dev.test.js
```diff
@@ -158,6 +158,68 @@ describe('dev container', () => {
...
+					expect(r.res.statusCode).to.equal(200);
+				}
+				{
+					// `/404` serves the custom 404 page as expected.
+					const r = createRequestAndResponse({ method: 'GET', url: '/404' });
+					container.handle(r.req, r.res);
+					await r.done;
+					const doc = await r.text();
+					expect(doc).to.match(/<h1>Custom 404<\/h1>/);
+					expect(r.res.statusCode).to.equal(200);
```
delucis, line 208: Not quite sure whether we should expect a 404 status here or whether this request/response mock always returns 200?

delucis:
Looks like the logic that renders custom 404s is failing for injected routes on Windows only 🤔

matthewp (approved):

PR #6943, opened 2023-04-29
"-y" parameter

hamer1818:
i used -y parameter for faster loading

## Changes

- What does this change?
  ```npx astro add vue``` 

## Testing

it's work

### DIFF
--- a/packages/integrations/vue/README.md
+++ b/packages/integrations/vue/README.md
@@ -16,7 +16,7 @@ To install `@astrojs/vue`, run the following from your project directory and fol
 
 ```sh
 # Using NPM
-npx astro add vue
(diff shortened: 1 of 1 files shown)


ankit-gautam23 (commented):
LGTM

bluwy:
I'm not sure about this. For first-time users, this would auto-accept prompts that they might not want to accept.

delucis:
I agree with @bluwy. Maybe we should document `-y` in the CLI reference in docs but in this context we should show the basic command that lets people choose whether they want to accept changes or not.

bluwy:
Closing this for now as the docs changes can be done separately.

PR #6946, opened 2023-04-29
New feature: Abbility to test foreign Astro integrations outside of Astro

RaphaelBossek:
## Changes

Allow foreign projects create a fixture with Astro to test their own implementation against Astro.

## Testing

The main part of the change was to extract the required source code to the testing subdirectory and modify the test suite to apply these changes. Thus, the Astro test suite primarily uses this approach itself. If there are any sugesstions how to test the test suite itself, please comment.

## Docs

If this change would be accepted a tutorial how the Astro test suite does work would be the next step.

I've used this modification in my own project you can find here: https://github.com/grewety/astro-static-cms/tree/main/test


RaphaelBossek:
Regarding semver: I would increate the PATCH number but nothing more. It's only relevant for developers who want to use the Astro test suite. The Changeset is "broken" because is not rebased on the latest changes done on Astro main. But I can create new one if this PR will be accepted. Regarding the dependencies Vs devDependencies, no changes are mandatory from my test.

natemoo-re:
Thanks for the PR @RaphaelBossek! I know that testing Astro Integrations can be a major pain at the moment, but I don't think we can commit to exposing our internal testing utilities as a public API. If we did, that would mean those APIs would be subject to semver and we'd struggle to refactor them if needed. We are also looking into possibly changing our internal testing strategy to make things more performant, which this would complicate significantly.

I'm going to close this PR as it's something we can't commit to right now, but we'd love to see you submit a proposal in the https://github.com/withastro/roadmap repository so we can collaborate on a proper API for testing integrations!

PR #6950, opened 2023-04-30
Update base.json

MrHBS:
Just small stylistic edit if you agree

## Changes

- What does this change?


bluwy:
`ESNext` is actually more accurate as defined by https://json.schemastore.org/tsconfig

PR #6952, opened 2023-05-01
fix: fix sharp semver range to allow for 0.32.0

Princesseuh:
## Changes

Semver range for Sharp didn't allow Sharp's latest version. This fixes that. 

This PR additionally updates our code to Sharp's 0.32.0, which removes the need for `@types/sharp`

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -201,7 +201,7 @@
     "unified": "^10.1.2"
   },
   "peerDependencies": {
-    "sharp": ">=0.32.0"
+    "sharp": ">=0.31.0"
   },
   "peerDependenciesMeta": {
     "sharp": {
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -181,7 +181,6 @@
     "@types/rimraf": "^3.0.2",
     "@types/send": "^0.17.1",
     "@types/server-destroy": "^1.0.1",
-    "@types/sharp": "^0.31.1",
     "@types/unist": "^2.0.6",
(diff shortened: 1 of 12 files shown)


matthewp (approved):

packages/astro/package.json
```diff
@@ -196,13 +195,13 @@
...
     "sass": "^1.52.2",
-    "sharp": "^0.31.3",
+    "sharp": "^0.32.1",
     "srcset-parse": "^1.1.0",
     "undici": "^5.22.0",
     "unified": "^10.1.2"
   },
   "peerDependencies": {
-    "sharp": "^0.31.3"
+    "sharp": ">=0.32.0"
```
bluwy, line 204: Isn't bumping the peer dep version a breaking change? Unless we keep it `>=0.31.3`

Princesseuh: Sharp has a bit of a weird relationship with versions, it doesn't really follow semver nor does it changes its API (and either way, is very unlikely to affect us since we only use the very very basic options, users don't have direct access to sharp)

bluwy: The issue is that when the user installs the new version of Astro, the package manager would then warn about the version mismatch, even though it will still work. IIRC some package managers would error out entirely too. So I think we could keep the version to avoid that.

Princesseuh: ah, yeah that's annoying. Will revert to >=0.31.0 then

bluwy (approved):

PR #6956, opened 2023-05-01
Make prerendering decision based on `PageBuildData` instead of `BuildInternals`

lilnasy:
Integration developers might expect to able to control prerendering behavior by mutating PageBuildData.route.prerender in the `'astro:build:setup'` hook. However, the decision to prerender is currently made by inspecting BuildInternals.

This PR changes where various parts of the build pipeline look to decide if a page should be prerendered. They now exclusively consider PageBuildData, allowing integrations to participate in the decision.

### DIFF
--- a/.changeset/popular-bats-pump.md
+++ b/.changeset/popular-bats-pump.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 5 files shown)


matthewp:
Merged as this all seems right to me. Would want to avoid documenting this however until we have a test where an integration mutates `route.prerender`.

lilnasy:
@matthewp tests implemented here #6957

PR #6957, opened 2023-05-01
Implement tests for integrations mutating prerendering

lilnasy:
Screenshot (124)


bluwy (commented):
The tests looks great! I think we can remove the changeset as it doesn't cause any changes from the end-user's point of view.

bluwy (approved):

PR #6960, opened 2023-05-02
Improve undici fetch failed errors in tests

bluwy:
## Changes

Sometimes undici fails with `fetch failed` without any other hints, this PR adds the URL that might have caused the fail, which I'm guessing it's the issue here.

## Testing

Shouldn't break existing tests.

## Docs

n/a


matthewp (approved):

bluwy:
Pushed another change to refactor the prerender tests so that only a single build and single dev server is used for the entire test file. I believe starting up multiple dev servers was causing the fetch errors before due to port auto-increment / port clashes.

The git diff isn't too kind, but I mainly re-ordered tests to share a single fixture, and tweaked some strings to accommodate it.

Issue #6525, opened 2023-03-12
Can't set multiple cookies in response on v2


michael-land:
### What version of `astro` are you using?

2.1.2

### Are you using an SSR adapter? If so, which one?

tried on vercel and node 

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

```ts
    Astro.cookies.set(`is_auth`, '1', {
      httpOnly: true,
      secure: true,
      sameSite: 'strict',
      expires: addMinutes(new Date(), 5),
      path: '/',
    });
    Astro.cookies.set(`authorization`, accessToken, {
      httpOnly: true,
      secure: true,
      sameSite: 'strict',
      expires: addMinutes(new Date(), 5),
      path: '/',
    });
```

<img width="812" alt="image" src="https://user-images.githubusercontent.com/33362998/224557706-b6913071-b267-4eb8-9742-5278f6ffd27c.png">
<img width="1044" alt="image" src="https://user-images.githubusercontent.com/33362998/224557748-7de4c5cf-a1f4-4bd8-8482-82a166ab3499.png">

### Link to Minimal Reproducible Example

todo

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Hello! What version of Node are you using? We've recently had a whole dilemma around cookie handling where specific versions of Node behaved differently and stuff, so it would be helpful to know!

jeffdrumgod:
I'm using 18.12 and have the same problem here

romanzy313:
Please reopen the issue, I am having the same problem as @jeffdrumgod. Setting multiple cookies works in build but not in dev. Really makes it impossible to use cookies while developing.

Running `node 18.14.0`, `astro 2.3.3`, `@astrojs/node 5.1.2`

matthewp:
Ok, will recheck.

matthewp:
Think a regression happened here. https://github.com/withastro/astro/pull/6355/files

alex-sherwin:
I'm seeing multiple set-cookie's work in dev and not in prod (nodejs adapter) using current Node LTS v18.16.0 and latest Astro 2.5.5

I'm not even using Astro.cookies at all, this is me directly setting up a Response with a Headers object with multiple set-cookie headers, e.g.

```typescript
const h = new Headers();
h.append("set-cookie", "cookie1=abc");
h.append("set-cookie", "cookie2=abc");
[...]

PR #6973, opened 2023-05-02
Ensure multiple cookies set in dev result in multiple set-cookie headers

matthewp:
## Changes

- Using `res.setHeader` which accepts an array of headers, which results in separate header lines. This is required for `set-cookie`.
- Fixes https://github.com/withastro/astro/issues/6525

## Testing

- Updated an existing test. Using raw sockets so that we can examine the raw response and see that there are 2 separate set-cookie headers. You can't examine this for sure using `fetch` because you can't see the raw response.

## Docs

N/A bug fix

### DIFF
--- a/.changeset/twelve-feet-switch.md
+++ b/.changeset/twelve-feet-switch.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Ensure multiple cookies set in dev result in multiple set-cookie headers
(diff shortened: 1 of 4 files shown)


ematipico (approved):

Princesseuh:
The test that failed is related to the change here. Since all other pipelines passed, I'll just rerun it to see if it's flaky, but due to undici differences between versions I wouldn't necessarily be surprised if there's a real change

PR #6974, opened 2023-05-03
Remove comments from base.json

joelleclerc:
Remove otherwise helpful JS-style comments causing various linting errors when base.json config is brought into Astro project by "extends" (comments not compatible with .json files)

## Changes

- What does this change?

## Testing
Deleting comments fixed the lint errors in my local project. No other testing done.

## Docs
I don't think doc updates are needed. I assume this issue was unintentional.

### DIFF
--- a/packages/astro/tsconfigs/base.json
+++ b/packages/astro/tsconfigs/base.json
@@ -1,30 +1,19 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
   "compilerOptions": {
-    // Enable top-level await, and other modern ESM features.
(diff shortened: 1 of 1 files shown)


bluwy:
tsconfig supports comments in JSON though, I don't think we should remove it. The linter probably needs to be configured to ignore that file.

Princesseuh:
Yeah, TypeScript officially support comments in JSON for their config. I believe that the default config you get with their init command even has comments in it.

Many editors will set the language as "JSONC" when opening a `tsconfig.json`, your linter should probably do a similar thing.

Issue #6849, opened 2023-04-14
React useId cannot generate unique identifiers for islands


SudoCat:
### What version of `astro` are you using?

2.2.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Firefox

### Describe the Bug

When rendering React components in separate islands on one page, React's `useId` hook generates duplicate IDs.

The react `useId` hook generates incrementing IDs based on the react component tree. As islands have separate roots, they will generate their IDs separately, resulting in collisions.

Many accessibility-focused React UI libraries make use of `useId` to generate aria attributes. This functionality is currently not compatible with Astro.

React does provide a configuration option, `identifierPrefix`, which can be passed to `createRoot`, `hydrateRoot`, `renderToString` and similar functions. 

If the react renderer adapter passed a unique `identifierPrefix` when rendering each island, this issue could be resolved. 

---

I attempted adding this in to the react adapter, but I wasn't sure on the best method to generate a consistent unique ID per component during server rendering which could be accessible to client rendering.

At first I considered using the UID attribute on the astro-island wrapping element, unfortunately that isn't available at server render time, as it's generated from hashing the server renderinng response. This value also isn't guaranteed to be unique - if the same component is rendered in two places on a page with the same props, the generated UID is identical. 

I'm not sure what the best solution would be, but I'd be more than happy to write up a PR if anyone has some better thoughts on how to achieve this!

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-lriyir?file=src/pages/index.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
Thanks @SudoCat We do something similar for Preact here: https://github.com/withastro/astro/blob/main/packages/integrations/preact/src/signals.ts#L43

I think the same method could work here. Use `this.result` which is a unique object for each render, and increment a counter. Then You can pass data attributes to use in the client.

SudoCat:
Excellent, this seems like a more practical solution with fewer issues than serializing and hashing the props.

I'll work on porting the preact context solution over to the react adapter, and shall raise a PR if I get everything working. Thanks for pointing me in the right direction!

PR #6976, opened 2023-05-03
Generate unique ids within each React island

SudoCat:
Resolves https://github.com/withastro/astro/issues/6849

React's useId hook generates unique IDs by incrementing down the component tree. When multiple react roots are on a single page, these IDs can collide. React provides an option, `identifierPrefix` to prefix ids per root.

Using preact adapters context solution, increment an ID per component rendered on a request, preventing collisions. 

## Changes

- Ensures each react island can generate unique ids without collisions, improve accessibility
- follows the context pattern used by preact-adapter to increment id per component
- incremental id is added to astro-island element attributes, named `prefix`
- id is passed to react server rendering methods directly as `identifierPrefix` option
- id is read from attributes client side, and passed as options to createRoot/hydrateRoot as `indetifierPrefix`

## Testing

- tested via local linking to validate results
- test case added to react test suite to validate that two ids generated in separate islands are unique.

## Docs

Unsure if documentation is needed; shouldn't change anything for users - something that didn't work properly will now work as intended.

### DIFF
--- a/.changeset/happy-ears-call.md
+++ b/.changeset/happy-ears-call.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+Prevent ID collisions in React.useId
--- a/packages/integrations/react/context.js
+++ b/packages/integrations/react/context.js
@@ -1,0 +1,24 @@
+const contexts = new WeakMap();
+
+const ID_PREFIX = 'r';
+
+function getContext(rendererContextResult) {
+	if (contexts.has(rendererContextResult)) {
+		return contexts.get(rendererContextResult);
+	}
+	const ctx = {
+		currentIndex: 0,
+		get id() {
+			return ID_PREFIX + this.currentIndex.toString();
+		},
+	};
+	contexts.set(rendererContextResult, ctx);
+	return ctx;
+}
+
+export function incrementId(rendererContextResult) {
+	const ctx = getContext(rendererContextResult)
+	const id = ctx.id;
+	ctx.currentIndex++;
+	return id;
+}
--- a/packages/integrations/react/server.js
+++ b/packages/integrations/react/server.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom/server';
 import StaticHtml from './static-html.js';
+import { incrementId } from './context.js';
 
 const slotName = (str) => str.trim().replace(/[-_]([a-z])/g, (_, w) => w.toUpperCase());
 const reactTypeof = Symbol.for('react.element');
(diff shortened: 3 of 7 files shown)


.changeset/happy-ears-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/react': minor
```
ematipico, line 2: Should this be a `patch` instead?

packages/integrations/react/context.js
```diff
@@ -0,0 +1,21 @@
+const contexts = new WeakMap();
+
+export function getContext(result) {
+	if (contexts.has(result)) {
+		return contexts.get(result);
+	}
+	let ctx = {
```
ematipico, line 7: nit: `let` => `const`

packages/integrations/react/context.js
```diff
@@ -0,0 +1,21 @@
+const contexts = new WeakMap();
+
+export function getContext(result) {
```
ematipico, line 3: Can we call `result` with a better name? It's a result of what?

packages/integrations/react/context.js
```diff
@@ -0,0 +1,21 @@
...
+	if (contexts.has(result)) {
+		return contexts.get(result);
+	}
+	let ctx = {
+		c: 0,
+		get id() {
+			return 'r' + this.c.toString();
```
ematipico, line 10: What are `c` and `"r"`? Sorry for being pedantic, but I think we can have better names.

Like, we can assign `'r'` to a constant and give it a good name, so we know what it's that for.

packages/integrations/react/server.js
```diff
@@ -58,6 +59,12 @@ async function getNodeWritable() {
 }
 
 async function renderToStaticMarkup(Component, props, { default: children, ...slotted }, metadata) {
+	let prefix;
+	if (this && this.result) {
+		prefix = incrementId(getContext(this.result))
```
ematipico, line 64: nit: we can avoid exposing the `getContext` function and use it inside `incrementId` directly. `getContext` is only used in this place, so we don't need to export it.

.changeset/happy-ears-call.md
SudoCat: I wasn't sure, so I went probably overly cautious. I couldn't really decide if this counted as a bug fix or additional functionality. I suppose it makes something work that didn't work before, as opposed to exposing new possibilities or options, so a patch is more appropriate? 
I'll update this. _Somehow_. 😅

packages/integrations/react/server.js
SudoCat: resolution: removed getContext

packages/integrations/react/context.js
SudoCat: resolution: renamed `c` to `currentIndex`. assigned `'r'` to `const ID_PREFIX = 'r'` and interpolated.

packages/integrations/react/context.js
SudoCat: renamed `result` to `rendererContextResult`

SudoCat:
No worries at all! Your pedantry aligns well with my own. I'm very grateful for a prompt review, anything to help get this accessibility issue fixed and shipped 🎉

PR #6981, opened 2023-05-04
Add global crypto to @astrojs/webapi

### LINKED ISSUE #6870: `crypto` global is only partially polyfilled

2.3.0

(none)

pnpm

Mac

(none)

### Describe the Bug

Global `crypto` seems to be polyfilled by Astro but only `crypto.getRandomValues()` exists. `crypto.subtle` and `crypto.randomUUID()` is undefined. 

https://github.com/pilcrowOnPaper/astro-crypto

### Participation

andremralves:
## Changes

fixes #6870

- Add the global `crypto` to be polyfilled in the webapi package.

<details>

I need some opinions to finish this issue

Currently the global crypto will not be polyfilled even if I add it to the webapi package (using `node version<=18`).

That is because it's already partially polyfilled by `@astrojs/compiler` in the global scope here: https://github.com/withastro/compiler/blob/bedced877522e3a14ec9416b592a2c185b8fd0d5/packages/compiler/node/wasm_exec.ts#L25-L33

The code above just polyfills the `getRandomValue` method and It runs before the `applyPolyfill` function is called in Astro's core. So the polyfill for `crypto` is skiped as it already exists (but with only one method).

I thought about some alternatives to solve that:

1. Polyfill the entire crypto interface in the `@astrorjs/compiler` package.
2. Call the polyfill from `webapi` before the compiler's files are imported.
3. Create a HACK in the `webapi` package to polyfill just the missing properties from crypto.
4. Remove the global polyfill from `@astrorjs/compiler` (I tested and nothing broke, but I don't know if can affect in something else).

</details>

## Testing
...
## Docs

No docs necessary


Princesseuh:
While https://github.com/withastro/compiler/pull/790 fixes the issue in dev, I would assume that in prod SSR, this is still an issue, right? I'm personally for this PR.

andremralves:
You are right @Princesseuh. Good catch! I tested locally and in prod SSR we still can't use crypto globally.

image

I also noticed that I get a type error when I import webcrypto because the `@types/node` package used by `webapi` is still on version `v14.18.21`, maybe it should be updated as v14 is no longer supported  and Astro uses version `>=16.12`.

PR #6990, opened 2023-05-04
Add caching for optimized images

Princesseuh:
## Changes

This PR adds a new setting `cacheDir` (default is `./node_modules/.astro` (same as `@astrojs/image`)) that is then used to cache optimized images in order to be able to reuse the build artifacts in subsequent builds.

This should unblock larger website from using `astro:assets`, as the previous implementation wasn't performant enough

## Testing

Added tests

## Docs

N/A

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -431,6 +431,23 @@ export interface AstroUserConfig {
 
 	/**
 	 * @docs
+	 * @name cacheDir
+	 * @type {string}
+	 * @default `"./node_modules/.astro"`
+	 * @description Set the directory used to cache build artifacts to. Files in this directory will be used in subsequent builds to speed up the build time.
+	 *
+	 * The value can be either an absolute file system path or a path relative to the project root.
+	 *
+	 * ```js
+	 * {
+	 *   cacheDir: './my-custom-cache-directory'
+	 * }
+	 * ```
+	 */
+	cacheDir?: string;
+
+	/**
+	 * @docs
 	 * @name site
 	 * @type {string}
 	 * @description
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -434,7 +434,7 @@ export interface AstroUserConfig {
 	 * @name cacheDir
 	 * @type {string}
 	 * @default `"./node_modules/.astro"`
-	 * @description Set the directory used to cache build artifacts to. Files in this directory will be used in subsequent builds to speed up the build time.
+	 * @description Set the directory for caching build artifacts. Files in this directory will be used in subsequent builds to speed up the build time.
 	 *
 	 * The value can be either an absolute file system path or a path relative to the project root.
 	 *
--- a/packages/astro/src/assets/internal.ts
+++ b/packages/astro/src/assets/internal.ts
@@ -73,12 +73,19 @@ export function getStaticImageList(): Iterable<
 	return globalThis.astroAsset.staticImages?.entries();
 }
 
-interface GenerationData {
+interface GenerationDataUncached {
+	cached: false;
 	weight: {
 		before: number;
 		after: number;
 	};
 }
+
+interface GenerationDataCached {
+	cached: true;
+}
+
+type GenerationData = GenerationDataUncached | GenerationDataCached;
 
 export async function generateImage(
 	buildOpts: StaticBuildOptions,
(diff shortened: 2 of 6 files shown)


packages/astro/src/core/config/schema.ts
```diff
@@ -13,6 +13,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 	srcDir: './src',
 	publicDir: './public',
 	outDir: './dist',
+	cacheDir: './node_modules/.astro',
```
matthewp, line 16: Why not the top-level `.astro` folder?

Princesseuh: This would require users to set up caching manually on their host, while `node_modules` is already cached by most hosts (same reason why Vite's cache is in `node_modules`, as far as I know)

matthewp: Ok, I didn't realize this. Makes sense in that case.

matthewp (approved):

packages/astro/src/assets/internal.ts
```diff
@@ -89,7 +96,8 @@ export async function generateImage(
 		return undefined;
 	}
 
-	const imageService = (await getConfiguredImageService()) as LocalImageService;
+	// Ensure that the cache directory exists
+	await fs.promises.mkdir(buildOpts.settings.config.cacheDir, { recursive: true });
```
ematipico, line 100: We need to make sure we have the permissions to create the directory. If we don't, is the error already handled?

packages/astro/src/assets/internal.ts
```diff
@@ -112,19 +133,20 @@ export async function generateImage(
...
 		buildOpts.settings.config.image.service.config
 	);
 
-	const finalFileURL = new URL('.' + filepath, clientRoot);
-	const finalFolderURL = new URL('./', finalFileURL);
-
 	await fs.promises.mkdir(finalFolderURL, { recursive: true });
-	await fs.promises.writeFile(finalFileURL, resultData.data);
+	await fs.promises.writeFile(cachedFileURL, resultData.data);
+	await fs.promises.copyFile(cachedFileURL, finalFileURL);
```
ematipico, line 146: Are these FS operations safe? Do you know if errors are handled already?

packages/astro/src/assets/internal.ts
Princesseuh: They're definitely not, but we don't actually handle any kind of FS error at this stage, even outside assets. 

I'd say that it doesn't really matter and we'd rather the build crash with the error than attempt to continue, so handling it would pretty much just be rethrowing it

packages/astro/src/assets/internal.ts
Princesseuh: Same answer as the other comment, we don't handle FS errors in build in general, we'd just rethrow it really

packages/astro/src/assets/internal.ts
matthewp: Good call out. I would expect us to gracefully handle this. If we don't have permission on the cache dir then it should proceed uncached.

ematipico (approved):

packages/astro/src/assets/internal.ts
Princesseuh: It does seems superfluous to me considering that: 
- We don't handle not having permissions to `dist` anyway in the end
- In most cases, the user will have just ran `npm install` before running the build, proving that they have write permissions to `node_modules`

But, I can definitely do it if it feels necessary

matthewp (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -429,6 +429,23 @@ export interface AstroUserConfig {
 	 */
 	outDir?: string;
 
+	/**
+	 * @docs
+	 * @name cacheDir
+	 * @type {string}
+	 * @default `"./node_modules/.astro"`
+	 * @description Set the directory used to cache build artifacts to. Files in this directory will be used in subsequent builds to speed up the build time.
```
sarah11918, line 437:
```suggestion
	 * @description Set the directory for caching build artifacts. Files in this directory will be used in subsequent builds to speed up the build time.
```
I think this is shorter and sweeter!

sarah11918 (approved):

sebastienbarre:
Howdy. I realize this is old, but I can’t find an answer in the docs. Provided I set `cacheDir` to, say, a `cache/` directory at the top of my source tree, outside of `node_modules/`, is it safe to commit this cache directory to my repo? I would make sure to only commit the images, not other assets or json files. And of course I would include a local build step before every deploy.

My use case here is that I do not want to build assets on the server; I want to do it locally. Space is cheap, and I have all the time in the world locally to build them. Ideally, I would assume that during the build step server-side, the prebuilt assets found in the cache directory would be used straight away, as if they had been built on the server instead of locally.

I’m mostly asking because, looking at my cache directory, I see that my filenames contain a hash, for example `typography.DE1iCK6d_Zbsv3q.webp`, and I’m wondering if this hash is stable. Meaning, is its value deterministic based only on the properties of the file, or does it include data from the machine the asset was built on? In other words, would a different hash be computed for the same image when built remotely versus locally? If that’s the case, that would be an unfortunate deal breaker with respect to committing my cache directory.

Thank you very much.

PR #6991, opened 2023-05-04
feat: hybrid output

MoustaphaDev:
## Changes

Implementation of the [hybrid rendering RFC](https://github.com/withastro/roadmap/blob/213cf99cd9ded5ac03b300366e79da21bb5f4332/proposals/hybrid-rendering.md).

This introduces the `hybrid` output format which enables prerendering by default and allows opting out of it with `export const prerender = false`.

## Testing

We will use the existing prerendering test fixtures, but we will make prerendering a function of a `PRERENDER` environment variable, doing `export const prerender = import.meta.env.PRERENDER` so that we can set prerendering to `false` or `true` depending on which feature we're testing.

## Docs

There's an open docs issue https://github.com/withastro/docs/issues/3226


packages/astro/src/@types/astro.ts
```diff
@@ -1095,6 +1096,40 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		middleware?: boolean;
+		
+		/**
+		 * @docs
+		 * @name experimental.hybridOutput
+		 * Enable experimental support for opt-out prerendering
```
sarah11918, line 1103: Just checking that we're missing some fields here like `@type`, `@default`, `@version`, and `@description`?

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -6,12 +6,12 @@ import { eachPageData, hasPrerenderedPages, type BuildInternals } from '../inter
 import type { AstroBuildPlugin } from '../plugin';
 import type { StaticBuildOptions } from '../types';
 
-export function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): VitePlugin {
+function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): VitePlugin {
 	return {
 		name: '@astro/plugin-build-pages',
 
 		options(options) {
-			if (opts.settings.config.output === 'static' || hasPrerenderedPages(internals)) {
```
MoustaphaDev, line 14: `vitePluginPage` runs before `vitePluginPrerender` which is responsible for parsing the `prerender` metadata of astro files, so there's no way to know if there are prerendered pages or not.
This would always return `undefined` so I removed it.

packages/astro/src/@types/astro.ts
sarah11918, line 1103:
```suggestion
		 * Enable experimental support for hybrid SSR with pre-rendering enabled by default.
```
I'm thinking something like this is a little more descriptive if people are reading to figure out what this is? Adjust as needed to be totally accurate, but I think this sentence could more explain what the option does in terms the users will think of their project.

packages/astro/src/@types/astro.ts
```diff
@@ -1095,6 +1096,40 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 *  output: 'hybrid',
+		 * 	experimental: {
+		 *		hybridOutput: true,
+		 * 	},
+		 * }
+		 * ```
+		 * Then add `export const prerender =  false` to any page or endpoint you want to opt-out of prerendering.
```
sarah11918, line 1115:
```suggestion
		 * Then add `export const prerender =  false` to any page or endpoint you want to opt-out of pre-rendering.
```
I noticed that we are not totally consistent in our own docs (I'll make a task to go fix that!) but everywhere else online seems to use the hypen for `pre-rendering`. So if that seems right to you, let's make that our standard.

sarah11918 (commented):
Just a few quick comments here! Congrats on this, @MoustaphaDev! People are going to love this!  :partying_face: :rocket:

packages/astro/src/@types/astro.ts
MoustaphaDev: Agreed, this explains the feature better

packages/astro/src/core/config/config.ts
```diff
@@ -351,3 +357,9 @@ export function mergeConfig(
 ): Record<string, any> {
 	return mergeConfigRecursively(defaults, overrides, isRoot ? '' : '.');
 }
+
+// TODO: remove after the experimetal phase when
+// it won't be needed to edit two config fields to enable the feature
+function isHybridMalconfigured(config: AstroConfig) {
+	return config.experimental.hybridOutput ? config.output !== 'hybrid' : config.output === 'hybrid';
```
matthewp, line 364: This is an example of another function that might be nice to live in a prerender folder.

matthewp:
Code looks really great overall. I left some comments more about code organization. But functionality looks perfect. Really appreciate how much work this was and how thorough the testing is.

packages/astro/src/core/build/static-build.ts
```diff
@@ -294,7 +300,9 @@ async function cleanStaticOutput(opts: StaticBuildOptions, internals: BuildInter
 		if (pageData.route.prerender)
 			allStaticFiles.add(internals.pageToBundleMap.get(pageData.moduleSpecifier));
 	}
-	const ssr = opts.settings.config.output === 'server';
+	const ssr =
+		opts.settings.config.output === 'server' ||
+		(opts.settings.config.experimental.hybridOutput && opts.settings.config.output === 'hybrid');
```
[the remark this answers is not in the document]
MoustaphaDev: Good idea, I'll refactor this

matthewp:
Also @MoustaphaDev can you add a changeset via `pnpm changeset`?

MoustaphaDev:
Sure, I'll add that

.changeset/mighty-shoes-scream.md
```diff
@@ -0,0 +1,39 @@
+---
+'astro': minor
+'@astrojs/cloudflare': patch
+'@astrojs/netlify': patch
+'@astrojs/vercel': patch
+'@astrojs/image': patch
+'@astrojs/deno': patch
+'@astrojs/node': patch
```
MoustaphaDev, line 8: Marked these as 'patch' because I updated the warning we get when using these adapters with an unexpected output mode.

packages/astro/src/core/build/generate.ts
```diff
@@ -231,7 +232,7 @@ async function getPathsForRoute(
 			route: pageData.route,
 			isValidate: false,
 			logging: opts.logging,
-			ssr: opts.settings.config.output === 'server',
+			ssr: opts.settings.config.output === 'server' || isHybridOutput(opts.settings.config),
```
MoustaphaDev, line 235: I could refactor these further with an `isSSR` utility but in my option it will make the code more difficult to understand, without stepping into the function.
Just my option, I'm open to changing it!

sarah11918 (approved):
Approving the docs edits!

PR #6995, opened 2023-05-04
fix: move service functions from astro/assets to astro/config so it can be imported

Princesseuh:
## Changes

The service functions were available from `astro/assets`, but it's not intended that people import things from there, as it's an internal file first and foremost. This PR moves the utilities methods to `astro/config`, so they can be imported easily in the same breath as the other utility methods.

## Testing

Tested manually

## Docs

https://github.com/withastro/docs/pull/3174

### DIFF
--- a/.changeset/mean-houses-exist.md
+++ b/.changeset/mean-houses-exist.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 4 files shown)


TheOtterlord (commented):
Works locally for me :+1:

matthewp (approved):

ematipico (approved):

PR #7000, opened 2023-05-04
fix: don't require 'paths' object in tsconfig to alias paths

craigjennings11:
In Astro versions < 2.3.0, simply declaring a `baseUrl` in your tsconfig would allow for path aliasing. However, commit 8539eb164 added a requirement that `paths` also had to exist for the aliases to be processed. This commit removes that unnecessary requirement so that aliasing can be achieved solely through defining a `baseUrl` again.

## Changes

It remove the need for a `paths` key in a tsconfig file for path aliasing.

## Testing

Added tests to ensure that only defining a `baseUrl` works

## Docs

This could probably cause changes for the https://docs.astro.build/en/guides/aliases/ page to mention that the 'paths' object isn't necessary to achieve path aliasing.

### DIFF
--- a/.changeset/dry-garlics-push.md
+++ b/.changeset/dry-garlics-push.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Remove 'paths' requirement for tsconfig path aliasing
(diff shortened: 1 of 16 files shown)


bluwy (approved):
Thanks for fixing this! I forgot to leave a review 😅

Issue #6998, opened 2023-05-04
astro@latest installing incorrect version


darbyjack:
### What version of `astro` are you using?

2.3.2

### Are you using an SSR adapter? If so, which one?

N/A

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

Running `pnpm create astro@latest` does not install the latest version (2.4.1 at the time of the post).

I was able to successfully reproduce this each time I attempted it.

![image](https://user-images.githubusercontent.com/21014720/236325441-3f9c5430-c3a3-4c7a-b6fa-9c840cb6a094.png)

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

sarah11918:
Just adding that I also tried, and got the same result! It installed v 2.3.2 for me, also!

TheOtterlord:
Same here.

Not sure what the requirement is for it to update the package.json because it does in https://github.com/withastro/astro/pull/6914 but not in https://github.com/withastro/astro/pull/6954

bluwy:
It's a pnpm 8 issue: https://github.com/pnpm/pnpm/issues/6463. The gist is that pnpm 8 now installs the lowest matching version instead of the highest, so `^2.3.2` => `2.3.2` and not `2.4.0`. There are discussions of a solution which pnpm would install the highest only if the lockfile is missing, but IMO it's a strange pnpm behaviour in the first place.

Anyways, what we can do is:

1. Add `.npmrc` of `resolution-mode=highest` for templates using pnpm.
2. Or, manually bump the versions in the templates for now.

I'd lean towards no2 to keep it simple, and hope that pnpm fixes this soon.

PR #7002, opened 2023-05-05
Update template deps to latest

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6998 for now

- First commit updates non-major deps: `pnpm update --filter "@example/*"`
- Second commit update deps with breaking changes: manual with `pnpm outdated --filter "@example/*"`

## Testing

I made sure the templates with breaking change deps still works by manually running dev and building + previewing them.

## Docs

n/a. dep update.

### DIFF
--- a/examples/middleware/package.json
+++ b/examples/middleware/package.json
@@ -12,12 +12,8 @@
     "server": "node dist/server/entry.mjs"
   },
   "dependencies": {
-    "@astrojs/node": "workspace:*",
-    "astro": "workspace:*",
-    "concurrently": "^7.6.0",
-    "html-minifier": "^4.0.0",
-    "svelte": "^3.58.0",
-    "unocss": "^0.15.6",
-    "vite-imagetools": "^4.0.19"
+    "@astrojs/node": "^5.1.2",
+    "astro": "^2.4.1",
+    "html-minifier": "^4.0.0"
   }
 }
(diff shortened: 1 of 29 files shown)


examples/middleware/package.json
```diff
@@ -12,12 +12,8 @@
...
   "dependencies": {
-    "astro": "workspace:*",
-    "svelte": "^3.48.0",
-    "@astrojs/node": "workspace:*",
-    "concurrently": "^7.2.1",
-    "unocss": "^0.15.6",
-    "vite-imagetools": "^4.0.4",
+    "@astrojs/node": "^5.1.2",
+    "astro": "^2.4.1",
     "html-minifier": "^4.0.0"
```
bluwy, line 17: @ematipico I changed the `workspace:*` to the usual semver identifier since these templates would be cloned as-is by `create-astro`, and it'll need the raw identifiers in order to install normally.

---

Also, not entirely related, but when running the middleware example (`pnpm dev`), following the login flow with username and password as `"a"`, I get in the terminal:

```
01:22:07 PM [middleware] Middleware doesn't work for endpoints that return a simple body. The middleware will be disabled for this page.
```

Seems like it's because we're returning nothing here that's causing the issue:

https://github.com/withastro/astro/blob/6a748958fcc54baa54d390d762a0252e76838811/examples/middleware/src/middleware.ts#L68

ematipico: We should probably remove the endpoint from the example... The warning was added later on, after a discussion with the team.

ematipico (approved):

ematipico (approved):

Issue #7003, opened 2023-05-05
Cannot find module '*?raw' or its corresponding type declarations.


J-env:
### What version of `astro` are you using?

2.4.0+

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm,pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

```
astro check && npx tsc --noEmit && astro build
```

<img width="901" alt="image" src="https://user-images.githubusercontent.com/25259920/236383333-5b33f8da-1630-445f-89e0-827414541086.png">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-sdeict-tyjyqg?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #7009, opened 2023-05-05
fix: declare namespace so client-base.d.ts stays an ambient definition

Princesseuh:
## Changes

In declaration files, you can't export a namespace, you need to declare it. Otherwise the file isn't considered an ambient declaration anymore and you can't refer to it with a triple slash

Fix https://github.com/withastro/astro/issues/7003

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/.changeset/giant-planes-try.md
+++ b/.changeset/giant-planes-try.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


matthewp (approved):

ematipico (approved):
thank you!

Issue #7007, opened 2023-05-05
Middleware can affect css bundling?


TheOtterlord:
### What version of `astro` are you using?

2.4.1

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

It appears middleware can cause a bug with CSS bundling. When using middleware with the tailwind integration (or as I also found later with style tags in the page) there is a split second of no CSS in development, and in a production build no CSS is present.

As I write this, I've noticed than in the build, the CSS is present is the bundle, but isn't included in the head of the HTML source.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-8rjwkk?on=stackblitz

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
That is definitely unexpected. Is this just with Tailwind or have you noticed anything with component CSS?

If it's Tailwind, that happens via `injectScript` API. I still can't think of how the middleware change would effect this at the moment.

TheOtterlord:
Normal style tags also fail. I've updated the stackblitz with an example

ematipico:
For now, you can fix the issue by writing `return next()` instead. I will look at why just calling `next()` doesn't work.

matthewp:
It's a requirement that a middleware returns either a `Response` or the return value of `next()` (which is a Response). Do we have a warning in place already about not returning anything from the middleware?

ematipico:
We don't. I am going to add it and make sure that we handle this case. Eventually we should adhere to one single way to do things, I think.

PR #7010, opened 2023-05-05
fix(middleware): emit warning if `next` is called and nothing is returned

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/7007

There was a small bug where just calling `next()` without returning anything led to some errors.

Our `next` function, internally, calls an `await` and that can lead to an error if the user **doesn't** await the `next` function.

I applied the following changes:
- added a warning in case the user does `next()` without returning anything;
- fixed this very case; now calling `next()` and returning anything **will throw an error**;
## Testing

I added a test case

## Docs

I removed from the examples:
- `api/`, middleware doesn't support endpoints that return `{ "body": "" }`
- the incorrect case that we want to prevent


matthewp:
To be clear, it is still required that you return the value of `next()` (or your own Response), right? We are not making it work that you just call next() without returning anything?

ematipico:
> To be clear, it is still required that you return the value of `next()` (or your own Response), right? We are not making it work that you just call next() without returning anything?

We will have to throw an Astro error instead of a warning, is that OK with you? (for now I made this case to work)

ematipico:
I updated the PR. Now calling `next()` and returning anything will throw an error.

matthewp:
@ematipico The error would be that they called `next` *after* the function completes, right? If so then that makes sense to me.

I think we should hold off on allowing this though, as it makes our code more complex and makes "how do you use next()" more difficult to explain. Good errors/warnings for now should help.

ematipico:
> @ematipico The error would be that they called `next` _after_ the function completes, right? If so then that makes sense to me.

Yes, the error will be emitted at the end of the function.

PR #7011, opened 2023-05-05
Add errors for unknown experimental keys

TheOtterlord:
## Changes

Adds validation for experimental keys, throwing a custom zod error if any unknown keys are present.

```
[astro] Unable to load /home/otterlord/Documents/Repos/astro/examples/basics/astro.config.mjs

 error   Invalid experimental key: `notARealExperiment`. 
  Make sure the spelling is correct, and that your Astro version supports this experiment.
  See https://docs.astro.build/en/reference/configuration-reference/#experimental-flags for more information.
```

<details>
<summary>Preview</summary>

</details>

## Testing

Manually tested

## Docs

The error links to existing related documentation. (see [related docs](https://docs.astro.build/en/reference/configuration-reference/#experimental-flags))


packages/astro/src/core/config/schema.ts
```diff
@@ -201,6 +201,17 @@ export const AstroConfigSchema = z.object({
...
+    .refine(d => {
+      const validKeys = Object.keys(ASTRO_CONFIG_DEFAULTS.experimental)
+      const invalidKeys = Object.keys(d).filter(key => !validKeys.includes(key))
+      if (invalidKeys.length > 0) return false
+      return true
+    }, d => {
+      const validKeys = Object.keys(ASTRO_CONFIG_DEFAULTS.experimental)
+      const invalidKeys = Object.keys(d).filter(key => !validKeys.includes(key))
+      return { message: `Invalid experimental key: \`${invalidKeys.join(', ')}\`. \nMake sure the spelling is correct, and that your Astro version supports this experiment.\nSee https://docs.astro.build/en/reference/configuration-reference/#experimental-flags for more information.` };
+    })
```
bluwy, line 214: Could we use [`.strict()`](https://github.com/colinhacks/zod#strict) here to automatically handle this for us? Or was the reason to generate friendlier error messages?

TheOtterlord: Yep, just as a friendlier message. The default `.strict()` behaviour wasn't clear enough to deal with the issues people were reporting (experiments not working because they forgot to update as an example)

bluwy: That make sense. I just found out that the config zod errors are pretty bad in general 😬 With `root: true` config, I get:

```
09:05:37 PM [astro] Unable to load /Users/bjorn/Work/oss/astro/examples/basics/astro.config.mjs

 error   Expected string, received boolean
```

I was hoping we could augment the zod errors to provide a better error message for the experimental options, so we don't have to re-implement the `.strict()` behaviour.

TheOtterlord: I tried that at first, but if I remember it didn't allow me to use the object itself, so the error didn't even specify which key was the issue.

bluwy (approved):
Forgot to followup on this. Seems like a nice addition!

PR #7032, opened 2023-05-08
fix middleware typing export

### LINKED ISSUE #7017: Middleware imports are not being resolved properly in a TS project

2.4.1

None

pnpm

Mac

not relevante

### Describe the Bug

Hey peps. 👋 

First, thanks for finally adding `middleware` to Astro. This was a feature I had been waiting for so long, mainly because Vercel's middleware is very restrictive in what we can import and do.

## Problem

Anyhow, I have an Astro project in TS, and while trying to follow the documentation example (the sequence one, for example), I noted one thing: the types are not being resolved:

Screenshot 2023-05-06 at 10 00 22

That's a bit odd. I went down to Astro's module to see if the exports are declared in the package.json, and actually they are:

Screenshot 2023-05-06 at 10 01 55

If it's declared... I wasn't sure why the heck it wasn't working.

Then, I remembered I did some investigation a while ago about how the conditional types export (`exports[number].types`) are not always resolved in the IDE, kind of forcing us to rely on a single `.d.ts` export.

> [You can check my strategy here](https://raulmelo.dev/en/blog/build-javascript-library-with-multiple-entry-points-using-vite-3#generating-type-definitions)

But visiting this, I found a possible solution based on this issue (https://github.com/microsoft/TypeScript/issues/50794) opened on the TypeScript repository.

In summary, `"moduleResolution": "node"` can't handle these conditional exports and that's the option we have set in Astro's `base.json`:

Screenshot 2023-05-06 at 10 11 16

@andrewbranch states:

> package.json exports are only supported in --moduleResolution node16 and --moduleResolution nodenext. This is expected behavior, but we are working on expanding it. A few points of common confusion:
>
> - If you’re using Node 12 or 14, you should use --moduleResolution node16. >The features added to 16 have fully or mostly been backported to 12 and 14.
> - `--moduleResolution node` is for Node 11 and older. Its name is currently bad. We plan to rename it accordingly.
> - If you are using a bundler or a non-Node runtime, we do not currently have a mode for you. node16 and nodenext are only for Node.*

Then, I have overridden the `moduleResolution` to `node16` to see if that would fix the problem, and it did:

Screenshot 2023-05-06 at 10 16 23

## Solution

If it's said that Astro's project has to use Node 14+, does it make sense to change the `base.json` to `moduleResolution: node16`?

## Steps to reproduce 

1. access https://stackblitz.com/github/devraul/astro-middleware-types?file=src/middleware.ts
2. you'll see that the `astro/middleware` import isn't working (for TS only);
3. open `tsconfig.json` and uncomment the `compilerOptions` that sets `moduleResolution` to `node16` and save the file;
4. open the middleware again (`src/middleware.ts`);
5. the problem was gone

https://stackblitz.com/github/devraul/astro-middleware-types?file=src/middleware.ts

### Participation

raulfdm:
> Closes #7017

## Changes

This PR fixes the middleware's typing export. Read about the problem on #7017.

https://user-images.githubusercontent.com/12464600/236786864-744c4f50-1390-4950-81bc-b8a03828573e.mp4

## Testing

With the fix applied:

1. Open the `examples/middleware`;
2. check the `astro/middleware` import (the type should be there)

## Docs

No docs change is required since this is a functional change.

### DIFF
--- a/.changeset/new-seals-hear.md
+++ b/.changeset/new-seals-hear.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 fix middleware typing export for "moduleResolution: node"
--- a/.changeset/new-seals-hear.md
+++ b/.changeset/new-seals-hear.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+fix middleware typing export for "moduleResolution: node"
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -20,6 +20,9 @@
       ],
       "app/*": [
         "./dist/core/app/*"
+      ],
+      "middleware": [
+        "./dist/core/middleware/index.d.ts"
       ]
     }
   },


packages/astro/package.json
```diff
@@ -66,7 +69,6 @@
       "default": "./zod.mjs"
     },
     "./middleware": {
-      "types": "./dist/core/middleware/index.d.ts",
```
bluwy, line 69: We still need to keep this to support node16 and above support. The `typesVersion` trick only exists to support versions lower than that.

raulfdm: Hmm... I thought it wouldn't be a problem.

I added it back :)

.changeset/new-seals-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
bluwy, line 2: This should be a patch as it fixes a bug
```suggestion
'astro': patch
```

bluwy (approved):
Looks great! Thanks for the video showing that it works.

ematipico (approved):
Thank you for looking after this!

matthewp:
Closes https://github.com/withastro/astro/issues/7017

PR #7042, opened 2023-05-08
HTML minification

matthewp:
## Changes

- Adds the `compressHTML` flag to the config, defaulting to false.
- This value is passed to the compiler's `compact` flag.

## Testing

- Test case added for dev and build which verifies the HTML is compressed by checking the number of lines in the output (should be 1).

## Docs

- Config value is documented in the types.


matthewp:
This PR is not needed.

PR #3934, opened 2022-07-15
Add context to Node SSR adapter

laptou:
## Changes

- Added a new property `Astro.context` which can be accessed by components and a new `context` property on the argument passed into endpoints
- This context can be set by passing a third parameter `context` into the handler in the Node.js adapter

## Testing

- Added a unit test which verifies that the `Astro.context` global works as expected during SSR.

## Docs

https://github.com/withastro/docs/pull/1003

PR #7049, opened 2023-05-10
Allow `Astro.locals` to be set by adapters

wrapperup:
## Changes

This is the spiritual successor to https://github.com/withastro/astro/pull/3934
Astro Middleware RFC https://github.com/withastro/roadmap/issues/531

Adapters can now pass in an object to be used in `Astro.locals`. This is super useful for passing in data from a server into Astro, like user sessions and other per-request data derived on the server outside of Astro.

- Add ability for adapters to pass in a `locals` object into various render APIs
- RenderContext points to the request's `locals` property, just like APIContext.
- APIContext is no longer passed into `renderPage`, since `locals` can be accessed from the RenderContext
- In dev mode, `locals` can be passed in the request via `Symbol.for('astro.locals')`

APIContext being removed from renderPage may be a controversial change. As far as I can tell, the only thing it was passed in for was to access the locals object, but I'm not sure if there was any planned additions that will require the entire APIContext to be passed in.

## Testing

I wrote a few SSR tests for Astro core and the Node adapter.

## Docs

https://github.com/withastro/docs/pull/3224
Also updated the readme for the node integration with an example.


packages/astro/src/core/render/context.ts
```diff
@@ -49,12 +53,29 @@ export async function createRenderContext(
...
 		...options,
 		origin,
 		pathname,
 		url,
 		params,
 		props,
 	};
+
+	// We define a custom property, so we can check the value passed to locals
+	Object.defineProperty(context, 'locals', {
```
ematipico, line 67: I noticed this is the same code we use somewhere else. Maybe it's worth having a shared function and making it more "DRY". This would allow us to **not** have repeated logic to maintain.

ematipico (commented):
> APIContext being removed from renderPage may be a controversial change. As far as I can tell, the only thing it was passed in for was to access the locals object, but I'm not sure if there was any planned additions that will require the entire APIContext to be passed in.

That's precisely the case, the context is passed only to access `locals` and validate it (because that's **when `locals` is actually used in the rendering phase**.

The changes look good to me, although I'd like to have more tests:
- a test case where an error is thrown if `locals` *is not serialisable*
- a test case where an error is thrown if `locals` *is not an object* 

A bonus, maybe that can be followed by another PR (not mandatory, just a nice to have), an example where we show it can be used by our users. What do you think?

matthewp:
We are currently planning on this for the 2.6 release, about 3 weeks from now. At that time middleware will go stable and the docs changes can be accepted (with some possible improvements) how they are written now.

matthewp (dismissed):
Blocking to prevent accidental merges. The current plan is, if ready, this would go out with 2.6.

packages/integrations/node/README.md
```diff
@@ -119,6 +119,25 @@ app.use(ssrHandler);
...
+
+```js
+import express from 'express';
+import { handler as ssrHandler } from './dist/server/entry.mjs';
+
+const app = express();
+app.use(express.static('dist/client/'))
+app.use((req, res, next) => {
+  const locals = {
+    foo: 'bar'
```
sarah11918, line 132: Would love if this didn't use `foo: bar` but instead something real?  (In docs, we've been using `title: "New title"` to demonstrate updating a property.)

But, non-blocking!

sarah11918 (approved):
Docs has had their say! Thanks for providing us great stuff to work with, @wrapperup! :partying_face:

matthewp:
@wrapperup I'm not able to push changes to your repository. This PR needs to be merged with main in order to be mergeable. We'd love to get this in for 2.6 if possible.

matthewp:
@wrapperup Hey, we are in the process of merging things today. If you can resolve the conflicts we can get this in for 2.6. Thank you!

ematipico:
It seems we haven't managed to merge this PR. @wrapperup are you still interested in working on this PR? If not, please let us know, we can takeover your work.

ematipico:
Closing in favour of https://github.com/withastro/astro/pull/7385

Issue #6600, opened 2023-03-20
Dependencies installing with npm... error: Request timed out after one minute


WintrCodes:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac and Windows

### What browser are you using?

N/A

### Describe the Bug

When creating a new Astro project via 'npm create astro@latest', installing deps via the interactive CLI times out if you select 'yes' to install deps.

![image](https://user-images.githubusercontent.com/110974960/226331817-9292b659-421e-4056-92bf-c1a814fba6b7.png)

Workaround was to select 'no' and run 'npm i' after switching to the directory

Happened on Windows and MacOS Ventura

Requesting logs from Discord user (Peak)

Discord thread:
https://discordapp.com/channels/830184174198718474/1086276455773778020

### Link to Minimal Reproducible Example

Unable to replicate in StackBlitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Looks like it's coming from:

https://github.com/withastro/astro/blob/1f749c942af4fcf138f9c4be07ce6b189c0a88fa/packages/create-astro/src/actions/dependencies.ts#L46

Maybe it's because the `npm install` actually took more than a minute to finish? Does that happen when you run `npm  i` separately?

ematipico:
@WintrCodes, if you could share the contents of the npm logs, that would be amazing. It could help us understand what's the underlying issue.

WintrCodes:
> Looks like it's coming from:
> 
> https://github.com/withastro/astro/blob/1f749c942af4fcf138f9c4be07ce6b189c0a88fa/packages/create-astro/src/actions/dependencies.ts#L46
> 
> Maybe it's because the `npm install` actually took more than a minute to finish? Does that happen when you run `npm i` separately?

When running `npm i` outside of the installer CLI, it installed with zero issue

WintrCodes:
> I couldn't find the npm logs on my Mac; if someone knows where they'll be, I'll double check!

I couldn't find the npm logs on my Mac; if someone knows where they'll be, I'll double check!

ematipico:
> > I couldn't find the npm logs on my Mac; if someone knows where they'll be, I'll double check!
> 
> I couldn't find the npm logs on my Mac; if someone knows where they'll be, I'll double check!

[...]

PR #7052, opened 2023-05-10
fix: don't exit the program is dependencies don't install

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/6600

This change affects only the "dependencies" phase, but in my opinion, we should do this for each step.

## Testing

I manually changed the code to throw an error

Screenshot 2023-05-10 at 12 19 38


Princesseuh:
Hmm, I'm not sure I agree with this change. I think the warning is too subtle for users (we have to think that it's new users using create astro) and will result in many people asking for help on Discord due to thinking the deps installed when they didn't.

In my opinion, increasing the timeout so it happens less often would be a better first step

ematipico:
> In my opinion, increasing the timeout so it happens less often (if that is in fact the issue) would be a better first step 

It could be a timeout issue, but it could also be another issue that we can't control. Even if we change the timeout, we are back to square one. 

We can change the messaging, surely. I think we should not exit the program.

ematipico:
> > > In my opinion, increasing the timeout so it happens less often (if that is in fact the issue) would be a better first step
> > 
> > 
> > It could be a timeout issue, but it could also be another issue that we can't control. Even if we change the timeout, we are back to square one.
> > We can change the messaging, surely. I think we should not exit the program.
> 
> I think I'd at least go with an error instead of a warning maybe?

Done!

packages/create-astro/src/actions/dependencies.ts
```diff
@@ -25,12 +25,15 @@ export async function dependencies(
...
-				install({ pkgManager: ctx.pkgManager, cwd: ctx.cwd }).catch((e) => {
-					// eslint-disable-next-line no-console
+			while: () => {
+				return install({ pkgManager: ctx.pkgManager, cwd: ctx.cwd }).catch((e) => {
 					error('error', e);
-					process.exit(1);
-				}),
+					error(
+						'error',
+						'Dependencies failed to install, please install them manually after setup.'
```
delucis, line 33: Could we specify the command users need to run here something like this?

```suggestion
						`Dependencies failed to install, please run "${ctx.pkgManager} install" to install them manually after setup.`
```

Not sure if we usually have a style for code like this? Bold text?

packages/create-astro/src/messages.ts
```diff
@@ -94,6 +94,15 @@ export const info = async (prefix: string, text: string) => {
...
 
+export const warn = async (prefix: string, text: string) => {
+	if (stdout.columns < 80) {
+		log(`${' '.repeat(5)} ${color.yellow('◼')}  ${color.yellow(prefix)}`);
+		log(`${' '.repeat(9)}${color.dim(text)}`);
+	} else {
+		log(`${' '.repeat(5)} ${color.yellow('◼')}  ${color.yellow(prefix)} ${color.dim(text)}`);
+	}
+};
+
```
delucis, line 105: Should this be removed if it’s no longer needed?

delucis (approved):
LGTM! Do you have a screenshot of the output after the changes?

ematipico:
Unfortunately, with the current APIs of `cli-kit`, we can't avoid that "Dependencies installed" message, and I didn't want to change `cli-kit` for the time being :( 

What do you guys think?

delucis:
> Unfortunately, with the current APIs of `cli-kit`, we can't avoid that "Dependencies installed" message, and I didn't want to change `cli-kit` for the time being :(

Oh right… that’s unfortunate. I guess this PR is still much better than the whole thing crashing, but it would definitely be nice to avoid the confusing “Dependencies failed to install… Dependencies installed” message. How tricky is it to update this in a followup PR to `cli-kit`?

ematipico:
@delucis We can definitely do a follow up PR wit the change!

packages/create-astro/src/actions/dependencies.ts
```diff
@@ -25,12 +25,26 @@ export async function dependencies(
...
+					error(
+						'error',
+						`Dependencies failed to install, please run ${color.bold(
+							ctx.pkgManager + ' install'
+						)} to install them manually after setup.`
+					);
+				});
+				return install({ pkgManager: ctx.pkgManager, cwd: ctx.cwd }).catch((e) => {
 					error('error', e);
-					process.exit(1);
```
MoustaphaDev, line 32: Seems like removing this + the error message should've been enough to prevent exiting

ematipico: Yes, but it's definitely **not** a good DX printing the cause of the error without giving an explanation to the user :)

PR #7059, opened 2023-05-10
`@astrojs/image`: Add `fetchpriority` to `Picture` types

jasikpark:
## Changes

- Adds `fetchpriority` to allowed `Picture` attributes


jasikpark:
offtopic thought: why not have Picture pass the aspectRatio attribute? i can't seem to get that to pass through to the base image so i can reduce CLS

PR #7062, opened 2023-05-11
Bug 6672

### LINKED ISSUE #6672: Rendering content collection entries through a loop with imported astro components causes broken scripts and styles

2.1.7

None

npm

Mac

Chrome

### Describe the Bug

The new Content Collection API lets you import mdx files and render their content anywhere using the `.render()` method, and only imports those components and their styles and scripts when they're used on a page. This works perfectly when generating routes from content - on each page an entry, its scripts and styles are executed and bundled. But when I try to render multiple entries on one page through a loop, scripts and styles don't work.

```js
// pages/page.astro
---
const entries = await getCollection('someColletion');
---
{entries.map(async entry => {
	// Scripts and styles don't work in <Content />
	const { Content } = await entry.render();
	return <Content mdProps="work" />; 
})}
// content/someCollection/some-mdx.mdx
import AstroComponent from 'pathToComponent/AstroComponent.astro';
<AstroComponent />
// components/AstroComponent.astro
<!-- Some html -->
<script>/* Some scripts */</script>
<style>/* Some styles */</style>
```

Everything works when explicitly tell astro which components to render (see reproduction):

```js
---
const { Content } = await entries[0].render();
const { Content: Content2 } = await entries[1].render();
---
<Content />
<Content2 />
```

I want to render my mdx components with imported nested astro components through an array, as this is the only way to use astro components and prevent bundling of unused scripts and styles (https://github.com/withastro/astro/issues/6328)

https://stackblitz.com/edit/github-b1xdf7-ptottr?file=src%2Fcomponents%2FRenderEntry.astro,src%2Fpages%2Findex.astro,src%2Fpages%2Ffinal.astro

### Participation

wulinsheng123:
## Changes
fix #6672 

https://github.com/withastro/astro/issues/6672#issuecomment-1543892103

making the component renders the css and js, because `the bufferHeadContent` function will ignore css and js from component.


.changeset/small-wombats-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
```
matthewp, line 2: This is a bug fix, so `patch`

matthewp (approved):

PR #7067, opened 2023-05-11
Redirects

matthewp:
## Preview release

Note that this is now available as a preview release. You can install with:

```bash
npm install astro@next--redirects
```

Then you can set up redirects in your config:

```js

export default defineConfig({
  redirects: {
    '/other': '/place'
  }
});
```

Additionally you can use `Astro.redirect` in components.

### Output

With this preview release the output works in 2 ways:

- An HTML file is written out with: `<meta http-equiv="refresh" content="0;url=/" />`
- If you use the Netlify `netlifyStatic` adapter it will also write out the appropriate `_redirects` file for you.

## Changes

- Implements the Stage 3 Redirects RFC: https://github.com/withastro/roadmap/pull/587


matthewp:
!preview redirects

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -45,6 +45,7 @@ export const AstroErrorData = {
 	 * To redirect on a static website, the [meta refresh attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta) can be used. Certain hosts also provide config-based redirects (ex: [Netlify redirects](https://docs.netlify.com/routing/redirects/)).
 	 */
 	StaticRedirectNotAvailable: {
+		// TODO remove
```
Princesseuh, line 48: Just a heads up to not remove errors, but deprecate them (`@deprecated since version x`)

matthewp:
!preview redirects

.changeset/chatty-actors-stare.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+You can also specify spread routes using the same syntax as in file-based routing:
+
+```js
+import { defineConfig } from 'astro/config';
+
+export defineConfig({
+  redirects: {
+    '/blog/[...slug]': '/articles/[...slog]'
```
delucis, line 28: I’m guessing this is a typo and not a Freudian slip for what a slog it was putting this together 😅 
```suggestion
    '/blog/[...slug]': '/articles/[...slug]'
```

packages/astro/src/@types/astro.ts
```diff
@@ -733,6 +755,28 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		serverEntry?: string;
+		/**
+		 * @docs
+		 * @name build.redirects
+		 * @type {boolean}
+		 * @default `true`
```
delucis, line 762:
```suggestion
		 * @default `true`
		 * @version 2.6.0
```

packages/astro/src/@types/astro.ts
```diff
@@ -452,6 +453,27 @@ export interface AstroUserConfig {
 	 */
 	cacheDir?: string;
 
+	
+
+	/**
+	 * @docs
+	 * @name redirects
```
delucis, line 460: I’m noticing that the `image.service` option includes “(Experimental)” after the name and mentions it in the description too: 

Image service docs excerpt

Is that something we want to do for redirects too so it’s clear you need to enable the flag to get these settings?

packages/astro/src/@types/astro.ts
```diff
@@ -452,6 +453,27 @@ export interface AstroUserConfig {
...
+	 * and the value is the path to redirect to. Either of these an also be a dynamic route
+	 * following the same convention as in file-based routes.
+	 * 
+	 * ```js
+	 * {
+	 *   redirects: {
+	 *     '/old': '/new'
+	 *   }
+	 * }
+	 * ```
```
delucis, line 473: I guess we should probably document some more of the config details here too?

````suggestion
	 * @description Specify a mapping of redirects where the key is the route to match
	 * and the value is the path to redirect to. Either of these can also be a dynamic route
	 * following the same convention as in file-based routes.
	 * 
	 * ```js
	 * {
	 *   redirects: {
	 *     '/old': '/new',
	 *     '/blog/[...slug]': '/articles/[...slug]',
	 *   }
	 * }
	 * ```
	 *
	 * Astro will serve redirected GET requests with a status of `301`
	 * and use a status of `308` for any other request method.
	 * You can customize the status code using an object in the redirect config:
	 * 
	 * ```js
	 * {
	 *   redirects: {
	 *     '/other': {
	 *       status: 302,
	 *       destination: '/place',
	 *     },
	 *   }
	 * }
	 * ```
	 *
	 * Status codes are only set when using SSR or with a static adapter that
	 * supports it (Netlify, Cloudflare, and Vercel).
	 * Elsewhere, builds will fall back to a client redirect using a
	 * [`<meta http-equiv="refresh">` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv)
	 * and do not support status codes.
````

**Review by delucis — COMMENTED** [review]
Docs-only review from me. Thanks Matthew!

File: packages/internal-helpers/package.json [hunk 6]
```diff
@@ -0,0 +1,41 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/withastro/astro.git",
+    "directory": "packages/internal-helpers"
+  },
+  "bugs": "https://github.com/withastro/astro/issues",
+  "exports": {
+    "./path": "./dist/path.js"
+  },
```
ematipico, line 16: Any particular reason why `types` and `main` fields are missing?

packages/internal-helpers/src/path.ts
```diff
@@ -0,0 +1,81 @@
...
+	return idx === -1 ? path : path.slice(0, idx);
+}
+
+export function removeQueryString(path: string) {
+	const index = path.lastIndexOf('?');
+	return index > 0 ? path.substring(0, index) : path;
+}
+
+export function isRemotePath(src: string) {
+	return /^(http|ftp|https):?\/\//.test(src) || src.startsWith('data:');
```
ematipico, line 80: Should we add `ws://`?

packages/internal-helpers/src/path.ts
ematipico, line 1: A couple of non-blocking thoughts:
- it would be nice to have an explanation of the `path` module somewhere, what's that for; Or we could have some JsDoc attached to the functions that show some examples of how to use the functions.
- I noticed that these functions work on `string`, but internally we often use `URL` to load files, should we have some utility to support that too, or some example of how we can use these functions when we have `URL` at hand?

packages/astro/src/core/build/generate.ts
```diff
@@ -518,10 +565,29 @@ async function generatePath(
 			}
 			throw err;
 		}
-		throwIfRedirectNotAllowed(response, opts.settings.config);
-		// If there's no body, do nothing
-		if (!response.body) return;
-		body = await response.text();
+
+		switch(true) {
+			case (response.status >= 300 && response.status < 400): {
```
ematipico, line 570: Why not an `if/else`?

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -29,6 +30,11 @@ export function getVirtualModulePageNameFromPath(path: string) {
 	)}`;
 }
 
+export function getVirtualModulePageIdFromPath(path: string) {
+	const name = getVirtualModulePageNameFromPath(path);
+	return '\x00' + name;
```
ematipico, line 35: nit: we tend to use `\0` across the code base for virtual paths

ematipico (approved):
Some nitpick, not blocking!

packages/internal-helpers/package.json
matthewp: This module is a bunch of subpackage helpers so there isn't a main module.

packages/astro/src/core/build/generate.ts
matthewp: Not sure, can change it if that's the preference. My assumption was that this might expand to more status code handling but that might not be the case.

packages/astro/src/@types/astro.ts
```diff
@@ -452,6 +453,27 @@ export interface AstroUserConfig {
...
+	
+
+	/**
+	 * @docs
+	 * @name redirects
+	 * @type {RedirectConfig}
+	 * @default `{}`
+	 * @description Specify a mapping of redirects where the key is the route to match
+	 * and the value is the path to redirect to. Either of these an also be a dynamic route
+	 * following the same convention as in file-based routes.
```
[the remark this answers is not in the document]
matthewp: Added a sentence about this.

sarah11918 (commented):
Left some docs comments, for your consideration @matthewp!

.changeset/twenty-suns-vanish.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/netlify': minor
+---
+
+Support for experimental redirects
+
+This adds support for the redirects RFC in the Netlify adapter. No changes are necessary, simply use configured redirects and the adapter will update your `_redirects` file.
```
sarah11918, line 7:
```suggestion
This adds support for the redirects RFC in the Netlify adapter, including a new `@astrojs/netlify/static` adapter for static sites. 

No changes are necessary when using SSR. Simply use configured redirects and the adapter will update your `_redirects` file.
```
Given our discussion, would updating the changeset something like this make sense? (And, if the Cloudflare one is new, too, something like this in its changeset also?)

packages/integrations/netlify/README.md
```diff
@@ -74,6 +74,25 @@ export default defineConfig({
 });
 ```
 
+### Static sites
+
+For static sites you usually don't need an adapter. However, if you use `redirects` configuration in your Astro config, the Netlify adapter can be used to translate this to the proper `_redirects` format.
```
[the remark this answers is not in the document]
matthewp: They get merged. I'll add a note.

packages/internal-helpers/src/path.ts
matthewp: Done! Thanks for the suggestion.

natemoo-re (approved):
This looks great, excited for this feature!

packages/astro/src/@types/astro.ts
```diff
@@ -452,6 +453,53 @@ export interface AstroUserConfig {
...
+	 *     '/blog/[...slug]': '/articles/[...slug]',
+	 *   }
+	 * }
+	 * ```
+	 *
+	 *  
+	 * For statically-generated sites with no adapter installed, this will produce a client redirect using a [`<meta http-equiv="refresh">` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) and does not support status codes.
+	 *
+	 * When using SSR or with a static adapter in `output: static`
+	 * mode (Netlify, Cloudflare, and Vercel), status codes are supported.
```
sarah11918, line 484:
```suggestion
	 * mode, status codes are supported.
```
I think we decided to remove this list.

PR #7071, opened 2023-05-11
Parallelize rendering of sibling components to avoid async waterfalls

johannesspohr:
## Changes

This changes the rendering behavior of Astro components to render child components in parallel.
In cases where multiple components in different subtrees fetch data, this data fetching can be done in parallel and will improve loading times in those scenarios.

## Testing

A test is added to confirm the parallel execution of the artificial delays. Also, an existing test (number of stream chunks) had to be changed, because the parallel execution lead to 2 chunks instead of the expected 3.

## Docs

Docs & changeset will be added as this is just a draft

### DIFF
--- a/.changeset/clever-garlics-relate.md
+++ b/.changeset/clever-garlics-relate.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Render sibling components in parallel
--- a/packages/astro/src/runtime/server/render/astro/render-template.ts
+++ b/packages/astro/src/runtime/server/render/astro/render-template.ts
@@ -3,6 +3,7 @@ import type { RenderInstruction } from '../types';
 import { HTMLBytes, markHTMLString } from '../../escape.js';
 import { isPromise } from '../../util.js';
 import { renderChild } from '../any.js';
+import { EagerAsyncIterableIterator } from '../util.js';
 
 const renderTemplateResultSym = Symbol.for('astro.renderTemplateResult');
 
@@ -35,12 +36,16 @@ export class RenderTemplateResult {
 	async *[Symbol.asyncIterator]() {
 		const { htmlParts, expressions } = this;
 
+		let iterables: Array<AsyncIterable<any>> = [];
+		for (let i = 0; i < htmlParts.length; i++) {
+			iterables.push(new EagerAsyncIterableIterator(renderChild(expressions[i])));
+		}
 		for (let i = 0; i < htmlParts.length; i++) {
 			const html = htmlParts[i];
-			const expression = expressions[i];
+			const iterable = iterables[i];
 
 			yield markHTMLString(html);
-			yield* renderChild(expression);
+			yield* iterable;
 		}
 	}
 }
--- a/packages/astro/src/runtime/server/render/astro/render-template.ts
+++ b/packages/astro/src/runtime/server/render/astro/render-template.ts
@@ -41,8 +41,10 @@ export class RenderTemplateResult {
 		for (let i = 0; i < htmlParts.length; i++) {
 			iterables.push(new EagerAsyncIterableIterator(renderChild(expressions[i])));
 		}
-		// once the execution is interrupted, we start buffering the other iterators
+		// once the execution of the next for loop is suspended due to an async component,
+		// this timeout triggers and we start buffering the other iterators
 		setTimeout(() => {
+			// buffer all iterators that haven't started yet
 			iterables.forEach((it) => !it.isStarted() && it.buffer());
 		}, 0);
 		for (let i = 0; i < htmlParts.length; i++) {
--- a/packages/astro/src/runtime/server/render/astro/render-template.ts
+++ b/packages/astro/src/runtime/server/render/astro/render-template.ts
@@ -36,10 +36,15 @@ export class RenderTemplateResult {
 	async *[Symbol.asyncIterator]() {
 		const { htmlParts, expressions } = this;
 
-		let iterables: Array<AsyncIterable<any>> = [];
+		let iterables: Array<EagerAsyncIterableIterator> = [];
+		// all async iterators start running in non-buffered mode to avoid useless caching
 		for (let i = 0; i < htmlParts.length; i++) {
 			iterables.push(new EagerAsyncIterableIterator(renderChild(expressions[i])));
 		}
+		// once the execution is interrupted, we start buffering the other iterators
+		setTimeout(() => {
+			iterables.forEach((it) => !it.isStarted() && it.buffer());
+		}, 0);
 		for (let i = 0; i < htmlParts.length; i++) {
 			const html = htmlParts[i];
 			const iterable = iterables[i];
--- a/packages/astro/src/runtime/server/render/page.ts
+++ b/packages/astro/src/runtime/server/render/page.ts
@@ -157,6 +157,10 @@ export async function renderPage(
 										}
 									}
 								}
+
+								// `chunk` might be a Response that contains a redirect,
+								// that was rendered eagerly and therefore bypassed the early check
+								// whether headers can still be modified. In that case, throw an error
 								if (chunk instanceof Response) {
 									throw new AstroError({
 										...AstroErrorData.ResponseSentError,
--- a/packages/astro/src/runtime/server/render/page.ts
+++ b/packages/astro/src/runtime/server/render/page.ts
@@ -157,6 +157,11 @@ export async function renderPage(
 										}
 									}
 								}
+								if (chunk instanceof Response) {
+									throw new AstroError({
+										...AstroErrorData.ResponseSentError,
+									});
+								}
 
 								const bytes = chunkToByteArray(result, chunk);
 								controller.enqueue(bytes);
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -136,24 +136,40 @@ export function renderElement(
 	return `<${name}${internalSpreadAttributes(props, shouldEscape)}>${children}</${name}>`;
 }
 
-// This eagering consumes an AsyncIterable so that its values are ready to yield out
-// ASAP. This is used for list-like usage of Astro components, so that we don't have
-// to wait on earlier components to run to even start running those down in the list.
+// This wrapper around an AsyncIterable can eagerly consume its values, so that
+// its values are ready to yield out ASAP. This is used for list-like usage of
+// Astro components, so that we don't have to wait on earlier components to run
+// to even start running those down in the list.
 export class EagerAsyncIterableIterator {
 	#iterable: AsyncIterable<any>;
 	#queue = new Queue<IteratorResult<any, any>>();
 	#error: any = undefined;
 	#next: Promise<IteratorResult<any, any>> | undefined;
+	/**
+	 * Whether the proxy is running in buffering or pass-through mode
+	 */
+	#isBuffering = false;
+	#gen: AsyncIterator<any> | undefined = undefined;
+	#isStarted = false;
+
 	constructor(iterable: AsyncIterable<any>) {
 		this.#iterable = iterable;
-		this.#run();
 	}
 
-	async #run() {
-		let gen = this.#iterable[Symbol.asyncIterator]();
+	/**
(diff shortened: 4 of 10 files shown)


packages/astro/src/runtime/server/render/page.ts
```diff
@@ -157,6 +157,11 @@ export async function renderPage(
 										}
 									}
 								}
+								if (chunk instanceof Response) {
+									throw new AstroError({
+										...AstroErrorData.ResponseSentError,
+									});
+								}
```
johannesspohr, line 164: Because we now eagerly render components, even redirects in nested components can now render successfully, as the response hasn't started sending at that point. So a response returned from a nested component, which will cause the following function to crash. So I'm catching this here again, but I'm not sure if this is the right place or the right way.

packages/astro/test/parallel.js
```diff
@@ -0,0 +1,24 @@
...
+	});
+
+	it('renders fast', async () => {
+		let html = await fixture.readFile('/index.html');
+		let $ = cheerio.load(html);
+
+		let firstStart = Number($('.start').first().text());
+		let lastStart = Number($('.start').last().text());
+		let timeTook = lastStart - firstStart;
+		expect(timeTook).to.be.lessThan(50, 'the components in the list were kicked off more-or-less at the same time.');
```
lilnasy, line 22: I know this was written this way in the original PR as well, but this message will be shown if the test fails.

johannesspohr: Thanks for the heads up, I've seen both interpretations (putting the expected behaviour or the error message in there), but I'm happy to make it more explicit 👍

lilnasy:
This addresses a major concern. Thanks for working on this.

packages/astro/test/parallel.js
```diff
@@ -0,0 +1,24 @@
...
+		await fixture.build();
+	});
+
+	it('renders fast', async () => {
+		let html = await fixture.readFile('/index.html');
+		let $ = cheerio.load(html);
+
+		let firstStart = Number($('.start').first().text());
+		let lastStart = Number($('.start').last().text());
+		let timeTook = lastStart - firstStart;
```
lilnasy, line 21: All components are kicked off at the same time, so timeTook here will be always be 0 or 1.
Screenshot (157)

I think the test should measure the longest time difference instead.
Screenshot (158)

```suggestion
		const startTimes  = Array.from($('.start')).map(element => Number(element.children[0].data))
		const finishTimes = Array.from($('.finished')).map(element => Number(element.children[0].data))

		const timeTook = Math.max(...finishTimes) - Math.min(...startTimes);
```

johannesspohr: For my understanding, the fact that all kick off at the same time is the result of this change. Otherwise, they would render in order and the start times would differ by more than 50ms. The big threshold of 50ms is a bit misleading though (I just copied that from the old pr as well)

lilnasy: Yeah I agree the number is misleading. It's what made me think the intention is to measure the render time. Maybe it could reduced and renamed to maybe `kickedOffWithin`?

johannesspohr: I also liked your idea, so I added the check as well. It's maybe not the cleanest unit test, as it tests more than one thing, but I think it's still OK.

johannesspohr:
I changed the behavior a bit to not start buffering right away, but only when an async component is reached (e.g. the event loop picks up the `setTimeout`). This could be optimized even further if we had a way to know whether a subcomponent tree has any async components in it, but for now, the current implementation should at least not cause any impact for components, that only have children that resolve immediately.

I ran benchmarks with `autocannon` against the portfolio example:
- before this PR: 204 req / s
- with caching all eagerly: 168 / s
- with current solution: 199 req / s

smnbbrv:
Thank you very much, when this gets merged the world becomes a lot better! This would close the discussion https://github.com/withastro/roadmap/discussions/577

> This could be optimized even further if we had a way to know whether a subcomponent tree has any async components in it, but for now, the current implementation should at least not cause any impact for components, that only have children that resolve immediately.

If I get it right, you mean every async subcomponent would have its own `setTimeout` as of now (and this is not optimal)? Or this actually fixes only top-level async components?

.changeset/clever-garlics-relate.md
ematipico, line 1: We should not have a changeset, because this is an internal change that doesn't change the behaviour of the public APIs

packages/astro/src/runtime/server/render/astro/render-template.ts
```diff
@@ -35,12 +36,21 @@ export class RenderTemplateResult {
...
 
+		let iterables: Array<EagerAsyncIterableIterator> = [];
+		// all async iterators start running in non-buffered mode to avoid useless caching
+		for (let i = 0; i < htmlParts.length; i++) {
+			iterables.push(new EagerAsyncIterableIterator(renderChild(expressions[i])));
+		}
+		// once the execution is interrupted, we start buffering the other iterators
+		setTimeout(() => {
+			iterables.forEach((it) => !it.isStarted() && it.buffer());
+		}, 0);
```
ematipico, line 47: Could you explain this `setTimeout`, please? 

My understanding is that with this function, the code becomes "asynchronous", meaning that it will be executed in the next "tick" and there's the risk that the next `for` loop will be executed before the timeout kicks in. Is my understanding correct? If so, could you explain why it's safe? If not, could you please explain to me? :)

packages/astro/src/runtime/server/render/page.ts
ematipico: I believe it's really valuable to add your comment **in the code**, so we don't lose this piece of knowledge!

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -135,3 +135,107 @@ export function renderElement(
...
+		this.#iterable = iterable;
+	}
+
+	/**
+	 * Starts to eagerly fetch the inner iterator and cache the results.
+	 * Note: This might not be called, once next() has been called once, e.g. the iterator is started
+	 */
+	async buffer() {
+		if (this.#gen) {
+			throw new Error('Cannot not switch from non-buffer to buffer mode');
```
ematipico, line 165: We should explain why and/or how a developer can fix the error.

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -135,3 +135,107 @@ export function renderElement(
...
+			this.head = { item };
+			this.tail = this.head;
+		} else {
+			this.tail!.next = { item };
+			this.tail = this.tail!.next;
+		}
+	}
+
+	isEmpty() {
+		return this.head === undefined;
```
ematipico, line 233: Without tests, I fail to understand when `this.head` can be `undefined`. Could you show me an example?

johannesspohr:
> If I get it right, you mean every async subcomponent would have its own `setTimeout` as of now (and this is not optimal)? Or this actually fixes only top-level async components?

To add a bit of context: the generic rendering operation is async, so it can either return a promise, or return an immediately resolved promise, that will basically synchronously continue the code execution. Caching a resolved promise is not only superfluous, but adds performance overhead, as we (synchronously) write everything into a buffer, just to read it again later, but we could have as well just streamed the output to the client.

In this scenario, we don't know whether a rendering operation will immediately resolve or not. So I opted for a middleground. We will instantiate the `EagerAsyncIterableIterator` for all children (because we won't know if an iterator started otherwise), but they are set to operate in pass-through / non-buffered mode. This means, that they will simply forward the `next()` call to the inner iterator and cause minimal impact. Now we register the timeout, which will run as soon as a child iterator does something _actually asynchronous_. Only then will the next "tick" be available and we have some slack time to start rendering the other children. So we will check which iterators haven't started yet, and start buffering those. In an ideal world, we could skip buffering the iterators that are also synchronous, as we gain nothing from buffering those (streaming the output directly is as fast). But other children, which are actually async, will start doing their stuff (network requests or similar).

Say the layout has 4 components:
- `Header` resolves immediately
- `Content` fetches some data
- `SeoContent` resolves immediately
- `Footer` fetches some data

We start by wrapping all in the `EagerAsyncIterableIterator`, but don't buffer yet, so nothing actually starts rendering.
- The `Header` gets rendered, as it synchronously resolves, it gets directly streamed to the output
- `Content` is rendered, but it needs to wait for the data, so it is async
- Then the `setTimeout` fires, which starts buffering all the iterators that haven't started, so `SeoContent` and `Footer`
- `SeoContent` resolves immediately, so we write that into the buffer (this step is useless, but we have no way of telling afaik)
- `Footer` is buffered, so it fires its network request
- Notice that at this point, we have the network requests for both async components running in parallel, the only delay is waiting for the tick and rendering the `SeoContent`, but this is not avoidable without knowing the async style
- When the `Content` resolves, the result will be streamed to the client
- The `SeoContent` is immediately streamed from the buffer
- The `Footer` is either directly streamed from the cache (if the network request resolved already) or streamed once its ready.

I hope this explains the process and also why the `setTimeout` is needed. There might be a better way to do it, but due to my limited knowledge about iterators etc., I don't see any.

Note also that this happens recursively in every child component.

.changeset/clever-garlics-relate.md
bluwy: I think we should keep it so we document what changed. If a regression happens, it would be easy to know which PR caused it from the changelog.

packages/astro/src/runtime/server/render/util.ts
johannesspohr: I can add a test if that helps to clear things up. The two situations when a queue `isEmpty()` are:
- initially (no items added so far)
- after the last entry has been `shift`ed (which has `.next` undefined, to `this.head` will be undefined afterwards)

johannesspohr:
@ematipico Thanks for the detailed review. I added more documentation. I find it kinda hard to estimate what is understandable to somebody who didn't think about this for hours 😅 so please keep bugging me if there still is something not clear from the comments.

bluwy:
On top of Ema's review, I did some perf runs using `wrk` with the `portfolio` example, and didn't notice any regressions or improvements, which I think is a good thing as it should be more noticable for async tasks that runs longer.

I guess a tiny concern is that running many async stuff in parallel could pressure the CPU and memory a bit, but I think in practice pages/components wouldn't be doing many intensive stuff at once.

.changeset/clever-garlics-relate.md
ematipico: Good call :) thank you @bluwy !

matthewp:
@johannesspohr It seems like we are pretty close on this one. We are doing the 2.5 release tomorrow. If you think you can address the review comments this could possibly go out then. Otherwise 2.6 will be in a couple of weeks.

johannesspohr:
@matthewp I was able to resolve everything except for that one thing where I'm unsure if this needs further documentation or tests. However, from a functionality standpoint, we should be good to go.

packages/astro/src/runtime/server/render/astro/render-template.ts
```diff
@@ -35,12 +36,23 @@ export class RenderTemplateResult {
...
 		const { htmlParts, expressions } = this;
 
+		let iterables: Array<EagerAsyncIterableIterator> = [];
+		// all async iterators start running in non-buffered mode to avoid useless caching
+		for (let i = 0; i < htmlParts.length; i++) {
+			iterables.push(new EagerAsyncIterableIterator(renderChild(expressions[i])));
+		}
+		// once the execution of the next for loop is suspended due to an async component,
+		// this timeout triggers and we start buffering the other iterators
+		setTimeout(() => {
```
matthewp, line 46: Why do this in a setTimeout and not immediately?

smnbbrv: @matthewp I guess that's the comment above https://github.com/withastro/astro/pull/7071#issuecomment-1549978562

ematipico (approved):
Thank you for addressing my pedantic feedback! 😅 💪 🚀

matthewp:
Thanks so much @johannesspohr, happy to have this in.

smnbbrv:
Hi guys, just tested the solution, it works for single components. However, it doesn't for fragment arrays, here is the follow-up issue https://github.com/withastro/astro/issues/7127

PR #7074, opened 2023-05-12
Implement custom client directives

bluwy:
## Changes

RFC: https://github.com/withastro/roadmap/pull/583

- Add `experimental.customClientDirectives` option.
- Support integration `astro:config:setup` hook `addClientDirective()` API to add custom client directive.
- Pass down `clientDirectives` map deep down to `SSRResult` to be rendered an runtime for `.astro` files.
- Updated `astro-scripts prebuild` to have special bundling for client directives (tried my best in implementation 😅)

## Testing

Added e2e tests

## Docs

https://github.com/withastro/docs/issues/3241

Also updated the types for `experimental.customClientDirectives` as the docs site will pull this info here. ([Link to diff](https://github.com/withastro/astro/pull/7074/files#diff-cec1de65bb8dd2f8330cb37fe3698afd66162bc0f311aa11e0b3dd573f6a2b9fR1082-R1102))


scripts/cmd/prebuild.js
```diff
@@ -39,11 +39,29 @@ export default async function prebuild(...args) {
 	}
 
 	async function prebuildFile(filepath) {
-		const tscode = await fs.promises.readFile(filepath, 'utf-8');
-		const esbuildresult = await esbuild.transform(tscode, {
-			loader: 'ts',
+		let tscode = await fs.promises.readFile(filepath, 'utf-8');
+		// If we're bundling a client directive, modify the code to match `packages/astro/src/core/client-directive/build.ts`.
+		// If updating this code, make sure to also update that file.
+		if (filepath.includes(`runtime${path.sep}client`)) {
```
bluwy, line 45: This code change isn't great, but I couldn't import `astro/src/core/client-directive/build.ts` for this to work, as this is a JS file (+ circular dep). We could also choose to not prebundle this, but it feels a bit wasteful to bundle on the user-side everytime if not.

---

Also, side-note I found `esbuild.build` to output smaller code than `esbuild.transform`. IIFE output is also needed so they don't bleed in `<script>` tags.

matthewp (changes requested):
Blocked because of minor release

packages/astro/e2e/fixtures/custom-client-directives/client-password.js
```diff
@@ -0,0 +1,20 @@
...
+  let consecutiveMatch = 0
+
+  const handleKeydown = async (e) => {
+    if (e.key === password[consecutiveMatch]) {
+      consecutiveMatch++
+    } else {
+      consecutiveMatch = 0
+    }
+
+    if (consecutiveMatch === password.length) {
```
matthewp, line 13: Don't forget to remove eventListener here :D

packages/astro/src/integrations/index.ts
```diff
@@ -97,6 +99,19 @@ export async function runHookConfigSetup({
 				addWatchFile: (path) => {
 					updatedSettings.watchFiles.push(path instanceof URL ? fileURLToPath(path) : path);
 				},
+				addClientDirective: ({ name, entrypoint }) => {
+					if (!settings.config.experimental.customClientDirectives) {
+						throw new Error(
```
matthewp, line 104: Can/should we be using `AstroError` here? cc @Princesseuh

packages/astro/src/integrations/index.ts
```diff
@@ -138,6 +153,11 @@ export async function runHookConfigSetup({
 			) {
 				addContentEntryType(mdxContentEntryType);
 			}
+
+			// Add custom client directives to settings, waiting for compiled code by esbuild
+			for (const [name, compiled] of addedClientDirectives) {
+				updatedSettings.clientDirectives.set(name, await compiled);
```
matthewp, line 159: Any perf concerns that waiting on these slows down dev startup time? What would be the downsides to saving the Promise instead? Code complexity is one thing that I would imagine.

matthewp:
The proposal mentions that overriding built-in directives is a non-goal. In the past some people expressed that this would be bad thing if we allowed it. I didn't notice in the code, are we doing anything that prevents someone from, for example, overriding `client:idle`?

packages/astro/src/integrations/index.ts
Princesseuh: Nope, since this particular error will be removed when it'll be unflagged. Temporary errors and errors in integrations are the two that don't / shouldn't use AstroError

bluwy:
> I didn't notice in the code, are we doing anything that prevents someone from, for example, overriding `client:idle`?

Yeah, I added a guard [here](https://github.com/withastro/astro/blob/24fef0fb2310fe69be6577417b57de0eec92f179/packages/astro/src/integrations/index.ts#L108-L112) (beside the experimental check). It probably also goes in hand of whether it should use `AstroError` instead of `Error`. I thought about that, but didn't go with it as the error happens on settings/config init (maybe it still make sense to use `AstroError` here?)

packages/astro/src/integrations/index.ts
bluwy: Yeah the first few tries I went with a thunk and promise approach, and code complexity was the main issue that I continue that.
1. It async poisons the renderer.
2. It touches a lot of types (the map has to be passed/drilled down to the renderer).

That was also why I kept prebundling the builtin directives so default apps don't have to always build it. But for custom directives, since we recommend keeping it small, I think the bundling is fast enough that it won't be noticable too.

Also, since we prebundle the custom directives eagerly above, this `for await` pattern should execute similarly to a `Promise.all`

sarah11918 (approved):
LGTM, @bluwy!  :partying_face: Just left a possible link you might want to include!

packages/astro/src/@types/astro.ts
```diff
@@ -1075,6 +1079,28 @@ export interface AstroUserConfig {
...
 		inlineStylesheets?: 'always' | 'auto' | 'never';
 
+		/**
+		 * @docs
+		 * @name experimental.customClientDirectives
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 2.5.0
+		 * @description
+		 * Allow integrations to use the experimental `addClientDirective` API in the `astro:config:setup` hook
```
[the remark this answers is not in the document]
bluwy: Makes sense to me! I didn't add it as the other two didn't have a link either 😬

.changeset/tall-eyes-vanish.md
```diff
@@ -0,0 +1,51 @@
...
+```astro
+<Counter client:click />
+```
+
+The client directive file (e.g. `astro-click-directive/click.js`) should export a function of type `ClientDirective`:
+
+```ts
+import type { ClientDirective } from 'astro'
+
+const clickDirective: ClientDirective = (load, opts, el) => {
```
ematipico, line 43: I believe in TS we would need to types for `load`, `opts` and `el`?

packages/astro/src/core/client-directive/build.ts
```diff
@@ -0,0 +1,33 @@
+import { build } from 'esbuild';
```
ematipico, line 1: nit: we could move the import of `esbuild` inside the function via dynamic import, so we load `esbuild` only when we create client directives.

packages/astro/test/test-utils.js
```diff
@@ -170,17 +177,17 @@ export async function loadFixture(inlineConfig) {
 	let devServer;
 
 	return {
-		build: (opts = {}) => {
+		build: async (opts = {}) => {
 			process.env.NODE_ENV = 'production';
-			return build(settings, { logging, telemetry, ...opts });
+			return build(await getSettings(), { logging, telemetry, ...opts });
```
ematipico, line 182: Do we need to call `getSettings` every time we call a function? Can't this be moved before the return at line 179?

scripts/cmd/prebuild.js
```diff
@@ -39,11 +39,29 @@ export default async function prebuild(...args) {
...
+			tscode = tscode.replace(
+				/export default (.*?)Directive/,
```
ematipico, line 47: Can you leave a comment for the regex too? This seems to be an unwritten rule where our directives must be called with a postfix "Directive"; otherwise, this script will fail. Also, they must have an `export default`. We should document it somewhere.

Ideally, this script should throw an error if the regex fails, with a message that tells the developer what went wrong.

.changeset/tall-eyes-vanish.md
bluwy: With the `ClientDirective` type, the params should be able to auto-infer the types.

packages/astro/src/core/client-directive/build.ts
bluwy: Vite starts up `esbuild` by default so I didn't went with the dynamic import approach to simplify it.

packages/astro/test/test-utils.js
bluwy: I documented why I did this at https://github.com/withastro/astro/blob/2bf1e2a9b88c2d6e9af93fb73f5e22a444a268bc/packages/astro/test/test-utils.js#L131-L142

The build/sync/check/dev commands could mutate `settings` so we'd have to re-create the settings to ensure the test isn't polluted.

scripts/cmd/prebuild.js
bluwy: Yeah I think adding some more guards and comments make sense. I kept it simple since we don't change directive code a lot.

Princesseuh (approved):
This looks great, awesome work @bluwy! I can't wait to see what the community do with this 😄

packages/astro/test/test-utils.js
ematipico: I read the comment, but I didn't understand that it was the reason why it has to be called every time. I thought it was the description of **what** the function does.

.changeset/tall-eyes-vanish.md
ematipico: I had trouble with that in the past, though.

ematipico (approved):
Let's ship it! 🚢 Great work! :)

matthewp (approved):

packages/astro/test/test-utils.js
bluwy: I updated the comment at 8769d3c. Hope it's clearer now!

PR #7078, opened 2023-05-12
Fix #7056 Splashing Navigation Menu in portfolio template 

### LINKED ISSUE #7056: Splashing Navigation Menu in portfolio template (example/portfolio)

2.4.4

None

pnpm

Linux

Chrome

### Describe the Bug

On mobile devices the expanded menu is visible by default for browsers that don't support JavaScript (or when the JavaScript is blocked), but is hidden for browsers that do support it. This is causing a brief **splash** effect when the page is initially loaded or refreshed, as the menu will be visible before JavaScript hides it.

Screen_Recording_20230510-201437_Chrome (1)

Also when JS is blocked the hero section's text is being hidden by overlapping menu, rendering it invisible.

localhost_3000_(iPhone XR) (1)

https://stackblitz.com/github/withastro/astro/tree/latest/examples/portfolio

### Participation

imNel:
## Changes

- Closes #7056 
- Added hidden to the menu-content div to fix the flickering on initial load. Unfortunately though this hides the menu when not using JS so I...
- Added a `<noscript>` tag to the Nav component to display the menu properly when not using Javascript.
  - Unfortunately this creates a repeated code situation where we have the header twice, so one block of code is always going to be hidden no matter the situation. I don't use vanilla JS much so I'm open to suggestions here on a better way to do this :)
- Fixed bug where menu was hiding hero text in non-JS

Before (No JS):
image

After (No JS):
image

Wasn't able to screen record the refresh bug happening as the flickering was a lot faster than in the issue's gif, but it's gone now.

## Testing

I was able to replicate the bug on my browser and phone with --host. Tested manually on both devices after changes were made.

## Docs

Purely a frontend change, shouldn't affect any docs.


krishna-santosh:
https://github.com/withastro/astro/issues/7056#issuecomment-1546538044

krishna-santosh:
Hey @imNel, Thanks you for submitting the PR for my issue! I really appreciate your work on this.

 Small issues though, I used brave browser's **Block Scripts** feature to disable JS, This is how it looks, I've tested this on my Linux Laptop and Windows Desktop, with pnpm, npm and bun 

1) When JS is disabled in desktop view

image

2) Social Icons when JS is disabled in Mobile view 

localhost_3000_(iPhone XR)

imNel:
Hey @krishna-santosh, fixed both issues now (see below)

image
image


While this does fix the issue you raised, I'm wondering if this is a good idea for an example project. Having a repeated element in a `<noscript>` tag makes for a confusing example, imo.

Do you think it's worth the extra complexity to keep this? Maybe it's simpler just to fix the hero text overlap and not worry about the flicker...

krishna-santosh:
Can you confirm if the appearance is normal when JavaScript is enabled?

krishna-santosh:
> Do you think it's worth the extra complexity to keep this? Maybe it's simpler just to fix the hero text overlap and not worry about the flicker...

We could possibly move the repeating code into its own component, this can reduce redundancy, but as you said it still makes it complex for an example template.

The whole point is to get rid of that flicker, when testing it out... I stumbled upon this overlapping issue,

natemoo-re:
Thanks for this PR, @imNel! I know this fix is technically correct and helpful, but I agree that it's a bit complex for a starter template. Not sure what we want to do about that, but maybe @delucis has some thoughts as the person who built this template!

krishna-santosh:
image

krishna-santosh:
@natemoo-re @imNel @delucis how about switching to a checkbox version and skipping JS  all together? This can hopefully make it less complex for a starter template and it can work even when JS is disabled. Getting rid of Splashing Navigation Menu.

natemoo-re (approved):
Thanks for the PR! Let's go ahead and merge this. Really appreciate the fix! ♥️

Issue #7070, opened 2023-05-11
Middleware does not work with API Endpoints


alex-sherwin:
### What version of `astro` are you using?

2.4.4

### Are you using an SSR adapter? If so, which one?

NodeJS

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

If you have an API Endpoint which returns a `Response` with **ANY** non-null body, middleware is skipped entirely and prints warning:

> Middleware doesn't work for endpoints that return a simple body. The middleware will be disabled for this page.

This is caused by this line of code: https://github.com/withastro/astro/blob/a44e755cc666cfbcff22eb80872265cd4298e5e5/packages/astro/src/core/endpoint/index.ts#L113

I can't think of why there's a conditional block around whether to run middleware or not for API Endpoints, but I'm pretty confident that this is not the intended behavior.

Maybe this conditional was trying to naively determine if `response` was of the optional `EndpointOutput` object instead of a `Response`?  But that doesn't seem likely since in the code path which does call the middleware types the middleware generics to expect the union type of `Response | EndpointOutput`.

It feels to me like the conditional around running middleware for API Endpoints should probably be removed entirely and just always run if middleware is defined.

However, if there is some legitimate some reason why middleware *should* be skipped, I don't think it should be based on whether `Response.body` is set or not, I don't think middleware should care about this for any particular reason

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-avaykm-qoanvo?file=src/pages/api-notok.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

ematipico:
Thank you for opening an issue! Yes, this is a valid bug and instead of checking against `null`, we should do this check

```js
if (response instanceof Response)
```

Since you're willing to open a PR, I am going to assign the issue to you! Thank you :) 
Let me know if you have further questions

alex-sherwin:
Hi @ematipico 

I was in the process of putting together this PR and unit test coverage, and realized that I think maybe there's more broken about middleware + endpoint's than I first thought.

Making the simple change to `response instanceof Response` will make middleware run the way it's currently coded, however.. this doesn't really seem right.

Middleware in Astro, from what I can tell, is [properly] designed to be chainable in the sense that the middleware chooses when the actual requested route is executed by calling `next()` (allowing middleware wrap a request and run logic before and after the **actual** request is executed (e.g. endpoint or page))

[...]

PR #7079, opened 2023-05-12
Bugfix #7070 Middleware not working with endpoints

alex-sherwin:
## Changes

* when middleware is used with endpoints, defer execution of `renderEndpoint` until the middleware invokes `next()` (making it properly chainable, like it is for pages)
* standalone type definitions for endpoint return types so that the types can be reused for a few purposes
* abstract conversion of endpoint "simple" return type into a Response into a reusable function
* modify endpoint middleware `next()` implementation, which invokes `renderEndpoint`, to convert any simple result into a Response and log a warning. This is necessary because middleware handlers are typed to only support receiving a `Response` instance from the result of `next()`.
* add unit test coverage
* expand the @examples/middleware to show usage with endpoint scenarios

I looked into changing middleware handlers to all needing to handle `Response | EndpointOutput`, but that was a deep rabbit hole that would have potentially breaking changes for any existing middleware handler code out in the wild.

So, instead, to make middleware "just work" for all use cases, in a sane way people would expect (which is that middleware should *ALWAYS* run, so you can trust it for things like auth/session checks, etc), I added an automatic conversion from any endpoint "simple" result into a `Response` immediately after it's resolved via `next()` using the same process that `App.callEndpoint` used, abstracting that functionality into a common utility function.

## Testing

Unit tests were added which cover:

* middleware + endpoint with `Response` result
* middleware + endpoint where middleware intercepts and returns a `Response` directly, never invoking `next()` (thus the API endpoint is never executed)
* middleware + endpoint with "simple" result using `text/plain;charset=utf-8` encoding
* middleware + endpoint with "simple" result using `application/json;charset=utf-8` encoding derived from path extension

## Docs

### DIFF
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -16,21 +16,24 @@ import { AstroError, AstroErrorData } from '../errors/index.js';
 import { warn, type LogOptions } from '../logger/core.js';
 import { callMiddleware } from '../middleware/callMiddleware.js';
 import { isValueSerializable } from '../render/core.js';
+import { simpleEndpointOutputToResponse } from '../util.js';
 
 const clientAddressSymbol = Symbol.for('astro.clientAddress');
 const clientLocalsSymbol = Symbol.for('astro.locals');
 
-type EndpointCallResult =
-	| {
-			type: 'simple';
-			body: string;
-			encoding?: BufferEncoding;
-			cookies: AstroCookies;
-	  }
-	| {
-			type: 'response';
-			response: Response;
-	  };
+export type EndpointCallResultSimple =  {
+  type: 'simple';
+  body: string;
+  encoding?: BufferEncoding;
+  cookies: AstroCookies;
+};
+
+export type EndpointCallResultResponse = {
+  type: 'response';
+  response: Response;
+};
+
+export type EndpointCallResult = EndpointCallResultSimple | EndpointCallResultResponse;
 
 export function createAPIContext({
 	request,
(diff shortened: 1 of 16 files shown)


alex-sherwin:
I'm not sure why the checks are failing on:

```
@astrojs/deno:test: [commonjs--resolver] Cannot bundle Node.js built-in "fs" imported from "../../../../../astro/dist/core/util.js". Consider disabling ssr.noExternal or remove the built-in dependency.
@astrojs/deno:test:  error   Cannot bundle Node.js built-in "fs" imported from "../../../../../astro/dist/core/util.js". Consider disabling ssr.noExternal or remove the built-in dependency.
```

I didn't change this import on util.js, and the tests run locally OK for me.

I re-pushed a commit to get these checks to re-run and they failed again for the same reason

Are the GitHub CI tests flaky, or did my PR really change something tangible causing this?

packages/astro/src/core/endpoint/index.ts
```diff
@@ -108,27 +111,39 @@ export async function call<MiddlewareResult = Response | EndpointOutput>(
...
+        });
+      }
+      // we can only support Response here, and *NOT* EndpointResponse, since this value will be passed back to the middleware handlers
+      const endpointResponse = await renderEndpoint(mod, context, env.ssr);
+      if (endpointResponse instanceof Response) {
+        return endpointResponse;
+      } else {
+        warn(
+          logging,
+          'ssr',
```
ematipico, line 132: Any particular reason to use `ssr` as logging instead of `middleware`?

packages/astro/src/core/endpoint/index.ts
```diff
@@ -16,21 +16,24 @@ import { AstroError, AstroErrorData } from '../errors/index.js';
 import { warn, type LogOptions } from '../logger/core.js';
 import { callMiddleware } from '../middleware/callMiddleware.js';
 import { isValueSerializable } from '../render/core.js';
+import { simpleEndpointOutputToResponse } from '../util.js';
```
ematipico, line 19: We can't import `util` in this file because this file relies on `node:` specific utilities. It's preferable to move `simpleEndpointOutputToResponse` somewhere else, where `node` is not used.

packages/astro/src/core/util.ts
```diff
@@ -221,3 +224,31 @@ export function resolvePath(specifier: string, importer: string) {
 		return specifier;
 	}
 }
+
+// EndpointOutput.body is typed as "Body" (not exported) which is only allowed to be a string
+export function isEndpointOutput(o: any): o is EndpointOutput {
```
ematipico, line 229: nit: I think we can give a better name instead of `o`. Maybe "responseOutput" works better

ematipico:
> I'm not sure why the checks are failing on:
> 
> ```
> @astrojs/deno:test: [commonjs--resolver] Cannot bundle Node.js built-in "fs" imported from "../../../../../astro/dist/core/util.js". Consider disabling ssr.noExternal or remove the built-in dependency.
> @astrojs/deno:test:  error   Cannot bundle Node.js built-in "fs" imported from "../../../../../astro/dist/core/util.js". Consider disabling ssr.noExternal or remove the built-in dependency.
> ```
> 
> I didn't change this import on util.ts, and the tests run locally OK for me.
> 
> I re-pushed a commit to get these checks to re-run and they failed again for the same reason
> 
> Are the GitHub CI tests flaky, or did my PR really change something tangible causing this?

Here's [why](https://github.com/withastro/astro/pull/7079#discussion_r1195051568)

alex-sherwin:
Ok I got it. 

Based on your issue comment here: https://github.com/withastro/astro/issues/7070#issuecomment-1549608841

Then, I'm happy to wait for an update from you/astro team.. as it stands right now I put some code in there which converts `EndpointOutput` to `Response` during the middleware processing, which I think is not desired (however, maybe it's ok in the interim until `EndpointOutput` is removed?)

For now I will update this PR so it works how I intended and let you decide what direction to go in, I'm happy to contribute more changes to align with that

alex-sherwin:
Closing in favor of #7106

PR #7086, opened 2023-05-14
Fix `create-astro` regression

MoustaphaDev:
## Changes

Fix a regression on `create-astro` where the dependency installation step always fails

## Testing

Tested manually

## Docs

N/A, bug fix


packages/create-astro/src/actions/dependencies.ts
```diff
@@ -26,15 +26,6 @@ export async function dependencies(
 			start: `Dependencies installing with ${ctx.pkgManager}...`,
 			end: 'Dependencies installed',
 			while: () => {
-				return Promise.reject('Unknown error').catch((e) => {
```
MoustaphaDev, line 29: Prematurely returning from the function here caused that the install step would never run

bluwy (approved):
Thanks for catching this!

PR #7092, opened 2023-05-15
Add `workerd` and `worker` to cloudflare adapter bundling

johannesspohr:
## Changes

This adds the `workerd` and `worker` import condition to bundling for Cloudflare, similar to how [SvelteKit](https://github.com/sveltejs/kit/blob/fb68c253a59402df2b2b1b5db7a54fad504cdd32/packages/adapter-cloudflare/index.js#L50) and [Cloudflare](https://github.com/cloudflare/workers-sdk/blob/133c0423ccb4c2b35a1dd26157ce9a24c6a743bb/packages/wrangler/src/bundle.ts#LL391C1-L391C48) is doing it.

## Testing

Added a test for SolidJS, which was broken on Cloudflare deployments before.

## Docs

As this is only a bugfix, probably no docs changes are needed.


bluwy (approved):
Thanks for fixing this! Looks like npm was down so the CI failed, will rerun it again.

Issue #6669, opened 2023-03-27
Deep `<slot />` in Vue component not rendering when hydrated


spacedawwwg:
### What version of `astro` are you using?

2.1.7

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome, Firefox

### Describe the Bug

Deep `<slot />` in Vue component not rendering when hydrated when nested within another Vue component that isn't hydrated.

It looks like it was an issue in v1 last year, as mentioned here in this issue https://github.com/withastro/astro/issues/4970

But it also looks like @matthewp said it should be fixed in this comment https://github.com/withastro/astro/issues/4970#issuecomment-1272015210 , but looking at the stackblitz in that comment, the same "Hydration completed but contains mismatches." are happening

I've put together a reproduction show what works and does not work here: https://stackblitz.com/edit/github-f6ayhq-eqxoez

The expectation is that a Vue component with a slot that **isn't** hydrated should simply render as HTML and the nested Vue component that **should** hydrate renders without mismatch.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-f6ayhq-eqxoez

### Participation

- [ ] I am willing to submit a pull request for this issue.

spacedawwwg:
Willing to add a PR if the issue in the astro code can be pointed out :)

wulinsheng123:
![image](https://user-images.githubusercontent.com/27261522/233817185-e20fafc4-a88f-4dc6-b89a-208373fa387c.png)
Why I can do it ?

spacedawwwg:
> ![image](https://user-images.githubusercontent.com/27261522/233817185-e20fafc4-a88f-4dc6-b89a-208373fa387c.png) Why I can do it ?

@wulinsheng123 if you read the issue at the top of that demo

"Issue: "Slot Text!" paragraph should be visible in every example below"

as you can see in your screengrab, the final two counters have no slot text rendering when they should

wulinsheng123:
> > ![image](https://user-images.githubusercontent.com/27261522/233817185-e20fafc4-a88f-4dc6-b89a-208373fa387c.png) Why I can do it ?
> 
> @wulinsheng123 if you read the issue at the top of that demo
> 
> "Issue: "Slot Text!" paragraph should be visible in every example below"
> 
> as you can see in your screengrab, the final two counters have no slot text rendering when they should

thank for you explain

[...]

Issue #6597, opened 2023-03-19
Experiencing hydration errors when passing strings as children to nested client React components


lokkeestudios:
### What version of `astro` are you using?

2.1.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I recently encountered some hydration error messages (server / client mismatch).
It seems like you cannot pass string children to a React component nested in another React component, which has any client directive but `client:load` (e.g. `<Container><Button>This is text</Button></Container>`).
Since this looks like quite the edge case it seems more like a bug than anything else to me, so I decided to create an issue on this.
Passing in the text as a prop or without the React component being the child of another React component, I don't experience these issues.
Furthermore, you can also pass strings as children to components, if they are not React, but Astro components.

Thank you in advance! 😊 

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/mystifying-mayer-6ozvmp?file=%2FREADME.md

### Participation

- [ ] I am willing to submit a pull request for this issue.

lokkeestudios:
After doing some further testing on this it seems like this issue applies to any children which are being passed from an Astro component into a React component (which is nested in another React component).

lokkeestudios:
@matthewp hmm.. weird. Let me know if I can help you in any way to get this issue resolved 🙂
I mean you technically just have to have two nested client React components in an Astro component, where the nested Component has children as well and you should experience the same hydration error.

matthewp:
I don't think framer is required to show this bug. I think this is happening here too: https://stackblitz.com/edit/github-jr2n4y?file=src%2Fpages%2Findex.astro&on=stackblitz

lokkeestudios:
@matthewp yes you are right. It does not have anything to do with framer-motion. Just wanted to stress that the Component is hydrated on the client-side.

matthewp:
So this is one is really tough given the way we currently do slotting. An astro-slot is injected so that we can find the inner HTML to rehydrate in the client. However a server-rendered framework component has its astro-slots stripped. When this happens it's also stripping the slots of a child island, which is causing this bug.

I'm thinking we can avoid adding the `astro-slot` for server rendered components, and then we won't need to remove them after. @natemoo-re worked on this part of the code so I'm going to wait until he gets back (he's on vacation this week) to talk to him about it.

lokkeestudios:
[...]

Issue #6301, opened 2023-02-18
Astro + Solid issue with hydration: nested islands being rendered partially


alexander-lozovsky:
### What version of `astro` are you using?

2.0.14

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

I'm building an application with Astro + Solid + node.js SSR, and I was facing very interesting issue with hydration

Sometimes in the app we're using nested islands, something like:
```
<Tabs client:load> // 
  <TabPanel>
     ...
     <Modal client:load>{/* modal content */}</Modal>
  </TabPanel>
  ...
<Tabs/>
```

And once we moved the app to different infrastructure with http caching enabled, we've started experiencing issues with such components being rendered partially in browsers (Chrome, Safari). In Chrome the issue happened from time to time, but stable reproducing in Safari 

I managed to create a minimal reproducible example https://github.com/alexander-lozovsky/astro-solid-cache-bug, where I'm using node.js ssr with a custom express server and http cache configured for static files.

First time you open the page in Safari, it is rendered as expected:
![image](https://user-images.githubusercontent.com/20779617/219874417-2b049d54-9c77-4781-892a-7debf7f21eda.png)

But when you reload the page, `.js` files are loaded from the cache, and the page is not rendering correctly:
![image](https://user-images.githubusercontent.com/20779617/219875638-bb21031e-ba1d-4639-afe1-52424d1ca6a2.png)

I was debugging hydration logic, and technically it happens because Astro starts hydration on `Wrapper` **before** Safari renders the `NestedComponent`: when it starts hydration on `Wrapper`, it takes `innerHTML` with `<h1>Hello world 1</h1>` and insert it to the page. So `NestedComponent` never renders

As this happened only when `.js` files are loaded from http cache, I assume it has something to do with browser event loop and how it prioritise tasks

In this example issue is reproducible only in Safari, but we observe similar things in Chrome as well. It looks like there are some gaps in how Astro hydrates such Solid islands, so when you enable http caching there are possible race conditions between hydration logic and browser rendering the page 

I'm leaning towards to not use Solid nested islands in our app at all, as there are too many issues with it at the moment (https://github.com/withastro/astro/issues/6263 one more issue I've raised). But since Astro allows to do this, it would be nice to improve hydration logic to avoid such race conditions if it is possible. If not - maybe framework can restrict this pattern or at least throw some warning that nested Solid islands might not safe to use

[...]

Issue #6263, opened 2023-02-16
If you nest two Solid islands into each other, the nested one is mounted twice


alexander-lozovsky:
### What version of `astro` are you using?

2.0.12

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

When you render two Solid islands nested into each other, the nested one is mounted twice.

So in case you have something like this:
```
<Wrapper client:idle>
  <NestedComponent client:idle>
    <p>Hello world</p>
  </NestedComponent>
</Wrapper>
```

`NestedComponent` runs twice in the browser, which means that if you run some logic inside `onMount`, it runs twice.

See the attached stackblitz: click "Open in New Tab" and observe "NestedComponent" logged twice in the console

Same composition with React components works as expected https://stackblitz.com/edit/github-tgt7ck?file=src/pages/index.astro

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-gcp42c?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

idiglove:
![image](https://user-images.githubusercontent.com/29911508/220032330-b2a5eada-b4af-4677-b720-0d6f51d55d20.png)
I don't see "Nested Component" being logged twice

alexander-lozovsky:
@idiglove you're looking at the dev server logs, please check the browser logs

![image](https://user-images.githubusercontent.com/20779617/220038327-7303ebde-8f8d-4ba2-a4e8-178e09db397f.png)
![image](https://user-images.githubusercontent.com/20779617/220038111-349baf96-290d-4d70-b535-cc213bbc651c.png)

SerekKiri:
Problem might be with a fallback to `load` event from `requestIdleCallback` as that what astro does -> https://docs.astro.build/en/reference/directives-reference/#clientidle

`window.requestIdleCallback` is not supported on safari
<img width="300" alt="image" src="https://user-images.githubusercontent.com/29735836/221299048-15ffe9be-aa1b-4215-8135-576fff44925c.png">

Question though as there is no `load` being used but a `setTimeout` instead if I'm looking into proper places. 
<img width="400" alt="image" src="https://user-images.githubusercontent.com/29735836/221300003-bfb621d2-cbad-433d-a7de-0c655bd1bc1e.png">
[...]

PR #7093, opened 2023-05-15
Prevent removal of nested slots within islands

matthewp:
## Changes

- This uses the tag name `astro-static-slot` instead of `astro-slot` when the slots are purely SSR. Doing this ensures that we only remove slots from SSR-only components and not from islands.
- This is done in a backwards compatible way because we only do this if a renderer includes the `supportsAstroStaticSlot` flag. Otherwise we preserve the old behavior.
- This is a temporary flag that we can remove in 3.0.
- Fixes https://github.com/withastro/astro/issues/6597
- Fixes https://github.com/withastro/astro/issues/6669

## Testing

- Used existing nested slot tests, but expanded and tested every affected framework:
  - React
  - Vue
  - Preact
  - Solid
  - Svelte

## Docs

N/A, bug fix


MoustaphaDev (changes requested):
Blocking as this is a minor

bluwy (approved):
Changes look good to me. It looks like https://github.com/withastro/astro/issues/6301 and https://github.com/withastro/astro/issues/6263 are kinda similar issues, but I don't think this PR fixes it.

Issue #7068, opened 2023-05-11
Different behavior with slot vs Astro.slots.render


ido-pluto:
### What version of `astro` are you using?

2.4.4

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Firefox

### Describe the Bug

When using the `<slot/>` syntax and rendering the same component twice, the slot content will be empty.
Not happening when doing with `Astro.slots.render`

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/priceless-solomon-sxv6r4?on=codesandbox&file=%2Fsrc%2Fcomponents%2FCardContainer.astro%3A1%2C1

### Participation

- [ ] I am willing to submit a pull request for this issue.

wulinsheng123:
let me take a look

lilnasy:
I don't see `Astro.slots` being used anywhere in the repro.

ido-pluto:
I created a new repo for this:
https://github.com/ido-pluto/astro-render-issue/tree/main

It seems like in Astro version 3, the issue changed form: instead of empty slots, all the slot renders are the same

ido-pluto:
This is working, but it is not the expected behaviour when I want to rerender a slot.
I need to make a special agistment to every nested slot there is.

I need this functionality for a framework I built for Astro that uses this for input validation when I want to render the same thing twice and It will have 2 two different outputs.

I think it still is performant without cache, because if someone wants to render twice, he expects a different output, if not, why not just store the output yourself in a variable?

lilnasy:
From looking at the compiler output, I would expect that behavior as well. This is probably a bug, I'll take a look.

PR #7099, opened 2023-05-16
fix bug 7068 that can't render two slots at the same time 

JerryWu1234:
## Changes

Loading two slots in the async pattern will lead to a component of nested which can not be loaded.  
fix bug #7068


packages/astro/src/runtime/server/render/astro/instance.ts
```diff
@@ -31,8 +31,9 @@ export class AstroComponentInstance {
 		this.factory = factory;
 		this.slotValues = {};
 		for (const name in slots) {
-			const value = slots[name](result);
-			this.slotValues[name] = () => value;
+			// collections-renderentry.test.js this units needs to early get the css from the children component and the slot  
+			slots[name](result)
+			this.slotValues[name] = () => slots[name](result);
```
matthewp, line 36: Calling this twice feels wrong to me. I can believe it fixes the test but it's puzzling why that would be.

wulinsheng123: I have tried a lot but I still haven’t found the way it broke that unit test.
It's very weird that it only breaks that one-unit test

wulinsheng123: https://github.com/withastro/astro/blob/31cbf4357e866abab62bb4ccf949f5a59aec06ac/packages/astro/test/units/dev/collections-renderentry.test.js#L157
There isn’t CSS, but the CSS style tag even appears in the bundle.

JerryWu1234: @matthewp coule help me check this

JerryWu1234:
it's weird 
everything is going well.
however, I added some tests for this part.

matthewp:
Hey @JerryWu1234, thanks as always. Going to close since the PR is old. Happy to have a new one if you can figure out the test problem, thanks!

Issue #6989, opened 2023-05-04
SolidJS integration on Vercel Edge breaks with "Ua is not supported in the browser"


Coretteket:
### What version of `astro` are you using?

2.3.4

### Are you using an SSR adapter? If so, which one?

Vercel Edge

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome, Firefox

### Describe the Bug

It appears that issue #5915 has re-appeared recently. Using versions `astro@^2.3.4`, `@astrojs/solid-js@^2.1.1`, `@astrojs/vercel@^3.3.0`, I'm getting the following error in Vercel logs, and an empty SolidJS component on client. Note that this error only appears when deployed.

```
Error: Ua is not supported in the browser, returning undefined
    at (entry.mjs:56:32412)
    at (entry.mjs:56:32522)
    at (entry.mjs:56:33846)
    at (entry.mjs:42:1044)
```

The minimal reproducible example is identical to that of issue #5915, except for the updated packages as described above.

### Link to Minimal Reproducible Example

https://github.com/pilcrowOnPaper/astro-vercel-solid

### Participation

- [ ] I am willing to submit a pull request for this issue.

wkelly17:
I don't have a min example to contribute, but I may be getting the same thing on Cloudflare as well.  Same deps cause the same blank issue, and locking to 
astro/cloudflare 6.2.2 and @astrojs/solid-js 2.1 seems to fix it. So, not sure if was the bump to 2.1.1 in astro/solid or some shared logic for the edge betwen 6.2.2 and 6.2.4, but seems like it was solid-js 2.1 is my guess.

johannesspohr:
I tried to mess around with a little, and I was able to make some progress, but still not 100%:

- The conditional exports for `solid-js/web` contains a `worker` section, that is listed before the `browser` exports, so it should be used when adding `worker` to Vite's `resolve.conditions`. If I try to do that in the cloudflare `astro:build:setup` hook, it still uses `web/dist/web.js`, which is causing the error, and not `web/dist/server.js`, as it should imho (but I might be missing something)

[...]

PR #7101, opened 2023-05-16
Build edge/worker runtime with webworker ssr target

bluwy:
## Changes

Part of the fix for https://github.com/withastro/astro/issues/6989

The solid integration was setting the SSR target to `node`, but in a Cloudflare / Vercel edge setup, it should be `webworker`. This PR fixes that.

Also fixes a missing esbuild dep in `@astrojs/vercel`

## Testing

Manually tested with the PR repro.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/quiet-gifts-pay.md
+++ b/.changeset/quiet-gifts-pay.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': patch
+---
+
+Add missing esbuild dependency
--- a/.changeset/silver-ties-vanish.md
+++ b/.changeset/silver-ties-vanish.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/cloudflare': patch
+'@astrojs/vercel': patch
+'@astrojs/solid-js': patch
+---
+
+Always build edge/worker runtime with Vite `webworker` SSR target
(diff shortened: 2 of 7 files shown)


ematipico (approved):

davycoe:
Hey this update seems to have killed graphql-request from making API calls once deployed to cloudflare. 
If I roll back the following @astrojs/cloudflare 6.3.0 > 6.2.4 / @astrojs/solid-js 2.2.0 > 2.1.1 it works.

Will this be a problem with other libs targeting node that worked pre SSR target change?

bluwy:
@davycoe sorry for the breaking change. It's tough to prevent it since it involves two packages that does separate things, and it was unfortunately coupled before that made it worked.

The new behaviour now is technically correct though since bundling for Cloudflare, we shouldn't target `node` following Cloudflare's runtime standard. It should also allow libraries that didn't work before to work now.

You could perhaps try adding the problematic packages to `config.vite.ssr.noExternal` in the astro config to workaround things, but it would be best to figure what packages are having trouble with Cloudflare specifically.

davycoe:
> 

I've worked out that the way environment variables were imported previously must have changed in the last update, due to not getting errors regarding my token my initial assumption was wrong!

Sorry for wasting your time :)

PR #7106, opened 2023-05-16
fix: middleware for API endpoints

ematipico:
## Changes

This PR fixes a bug where the function `renderEndpoint` was called too early, making the middleware effectively not work for ALL API endpoints.

cc @alex-sherwin

Partially addresses https://github.com/withastro/astro/issues/7070

- I renamed the function `call` to `callEndpoint`;
- I moved the warning about "simple" endpoints inside the `callMiddleware` function

## Testing

I created two test cases

## Docs

N/A

### DIFF
--- a/.changeset/long-starfishes-raise.md
+++ b/.changeset/long-starfishes-raise.md
@@ -1,0 +1,7 @@
+---
+'astro': patch
+---
+
+Fix middleware for API endpoints that use `Response`.
+
+Astro, will log a warning for endpoints that don't use `Response`. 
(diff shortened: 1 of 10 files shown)


.changeset/long-starfishes-raise.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': patch
+---
+
+Fix middleware for API endpoints that use `Response`.
+
+Astro, will log a warning for endpoints that don't use `Response`.
```
bluwy, line 7: tiny nit: the `Astro, ` part feels a bit out of place, maybe
```suggestion
Fix middleware for API endpoints that use `Response`, and log a warning for endpoints that don't use `Response`. 
```

bluwy (approved):

PR #7109, opened 2023-05-17
feat: remove experimental flag middleware

ematipico:
## Changes

This PR removes the experimental flag for the middleware.

While doing so, I had to change how types work inside the `manifest`. The middleware is a piece of code that is shared among all pages/entry points, but this doesn't work very well in SSR.

The way the SSR bundling logic works is logic this: 
1. the plugin `plugin-pages.ts` generates a virtual module that looks like this `export { middleware, page, renderers }` and `middleware` is imported from a shared chunk. 
2. the plugin `plugin-ssr` **tries** to import all the virtual modules from the previous plugin, and then tries to take the `middleware` code and expose it via `manifest`. Doing so could work better with the current plugin architecture.

So, to ensure that the code generated works with the current runtime, I removed the `middleware` variable from the manifest. This is safe because the virtual page emitted by `plugin-pages.ts` also has the middleware code.

This change made me realise that the current architecture is brittle and confusing, although I already know how I can change it. I will follow up with different PRs to make the change less confusing.

## Testing

Current tests should all pass.

The examples should be updated **after** this PR lands.

## Docs

N/A


matthewp (dismissed):
Block until 2.6

packages/astro/src/@types/astro.ts
```diff
@@ -108,7 +108,6 @@ export interface CLIFlags {
 	drafts?: boolean;
 	open?: boolean;
 	experimentalAssets?: boolean;
-	experimentalMiddleware?: boolean;
```
sarah11918, line 111: Just commenting that I think there's an entire section of docs to remove from this file! lines 1123 - 1142ish?

sarah11918 (commented):
Just a note below re: removing middleware from the list of experimental flags in docs!

packages/astro/src/@types/astro.ts
ematipico: Thank you @sarah11918 , I completely missed that part!  I updated the PR and removed that section now

sarah11918 (approved):

PR #7113, opened 2023-05-17
Cloudflare workerd - Static file compatibility

petertjmills:
## Changes

- What does this change?

Cloudflare workerd is the runtime that Cloudflare Workers and subsequently Pages runs on. `_worker.js` that is output by the astro cloudflare integration runs readily on workerd with little setup needed, with the only caveat being that static files aren't handled correctly. This is because by default workerd returns a `application/octet-stream` content type for all files. I've added some basic handling to re-wrap the file in a new content type. 

## Testing

Running `npx workerd serve config.capnp` with the following config.capnp:
```
using Workerd = import "/workerd/workerd.capnp";

const config :Workerd.Config = (
  services = [
    (name = "main", worker = .mainWorker),
    (name = "site-files", disk = "./astro/dist")
  ],

  sockets = [
    # Serve HTTP on port 8080.
    ( name = "http",
      address = "*:8080",
      http = (),
      service = "main"
    ),
  ]
);

const mainWorker :Workerd.Worker = (
  modules = [
    (name = "main", esModule = embed "./astro/dist/_worker.js")
  ],
  compatibilityDate = "2023-02-28",
  bindings = [
    (name = "ASSETS", service = "site-files"),
  ]
);
```

### EDIT: Adding tests

Created workerd test that runs astro with a static file using workerd.
- Checks that astro serves correctly with h1 element
- Checks for a string in a statically served SVG. 

## Docs

No docs needed to be added, this probably doesn't make a difference for services running on Cloudflare Pages, as they use the _routes.json file and don't fallback to serving static files.

### DIFF
--- a/.changeset/empty-months-pump.md
+++ b/.changeset/empty-months-pump.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Updating the static asset fallback handling, to rewrite the headers. This shouldn't affect cloudflare pages deployments as static file handling shouldn't hit fallback, but this does add compatibility with cloudflare's workerd, the open source worker runtime, as by default files are sent as `application/octet-stream` https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd.capnp#LL665C29-L665C53
--- a/.changeset/empty-months-pump.md
+++ b/.changeset/empty-months-pump.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': minor
 ---
 
-Updating the static asset fallback handling, to rewrite the headers. This shouldn't affect cloudflare pages deployments as static file handling shouldn't hit fallback, but this does add compatibility with cloudflare's workerd, the open source worker runtime, as by default files are sent as `application/octet-stream` https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd.capnp#LL665C29-L665C53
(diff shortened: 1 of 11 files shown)


ematipico:
Isnt't it there a way to create a test case in our test suite?

.changeset/empty-months-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Updating the static asset fallback handling, to rewrite the headers. This shouldn't affect cloudflare pages deployments as static file handling shouldn't hit fallback, but this does add compatibility with cloudflare's workerd, the open source worker runtime, as by default files are sent as `application/octet-stream` https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd.capnp#LL665C29-L665C53
```
ematipico, line 5:
```suggestion
Fix the static asset fallback handling to rewrite the headers. This shouldn't affect Cloudflare pages deployments as static file handling shouldn't hit fallback, but this does add compatibility with Cloudflare's workerd, the open source worker runtime, as by default files are sent as `application/octet-stream` https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd.capnp#LL665C29-L665C53
```

ematipico (commented):
Code looks good. I just wonder if it's possible to add a test case in our test suite

petertjmills:
Are there any docs on implementing tests? I haven't done much unit testing before 😄

ematipico:
> Are there any docs on implementing tests? I haven't done much unit testing before 😄

We already have some tests in our Cloudflare integration; could you start from there? (FYI, I am not very familiar with Cloudflare tests)

petertjmills:
Tests added. Not too sure why ubuntu failed check on post-setup

petertjmills:
Am closing now, because it seems there is a better way to do this with workerd, and if so, this patch does more harm than good. 
No longer required.

Issue #6785, opened 2023-04-08
Poor error message when collection doesn't have a folder / doesn't have entries


JoshuaKGoldberg:
### What version of `astro` are you using?

2.2.1

### Are you using an SSR adapter? If so, which one?

n/a

### What package manager are you using?

n/a

### What operating system are you using?

MacOS Ventura 13.2

### What browser are you using?

n/a

### Describe the Bug

If you create a new collection in `src/content/config.ts` that doesn't yet have a corresponding folder, Astro's message is:

```plaintext
[content] "example" is not a collection. Check your content config for typos.
```

That's not very informative. I stumbled onto this by accident when setting up a new Astro project; if I hadn't had @ElianCodes helping, it probably would have kept me confused for a while.

How about emitting an error message that indicates what's wrong?

```plaintext
[content] The "example" collection has no corresponding directory at src/content/example.
```

```plaintext
[content] The "example" collection has no entries under src/content/example.
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xdijqm?file=src/content/config.ts

Run `npm run astro -- sync` in the terminal to see the complaint.

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #7118, opened 2023-05-18
Improve content collection error message for empty folders

Princesseuh:
## Changes

This PR does two things:
- Enhanced the error message to be more clear that it means there's no folder for a collection, and recommend the user to  either create the folder or check their config
- Fix empty folders not being added to the list of collections at start, they'll now be added the same way they would've had they'd been added post launch

Fix https://github.com/withastro/astro/issues/6785

## Testing

Tested manually

## Docs

cc @withastro/maintainers-docs for feedback on the error message

### DIFF
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -509,9 +509,9 @@ function warnNonexistentCollections({
 			warn(
 				logging,
 				'content',
-				`${JSON.stringify(
+				`The ${JSON.stringify(
 					configuredCollection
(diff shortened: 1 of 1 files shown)


bluwy (approved):

bluwy:
Looks like we also need a changeset (maybe after the test fix PR is merged)

ematipico (approved):

bholmesdev (approved):
Direct-only you could be this thorough 👏

Yan-Thomas (approved):
Docs LGTM!

PR #7119, opened 2023-05-18
Fix customClientDirectives config default typo

bluwy:
## Changes

Seems like this is the only change that's causing the tests to fail.

Don't think it needs a changeset as it fixes an unreleased feature.

## Testing

Ran the test locally and passes

## Docs

n/a. bug fix.


delucis (approved):
LGTM! Agree re: the changeset 👍

PR #7104, opened 2023-05-16
Add files field to packages

bluwy:
## Changes

I found that many of out packages don't specify the `"files"` field, resulting in src, tests, and turbo logs being published to npm. I've added the field for the packages missing it.

## Testing

Ran `npm pack` for each packages to verify.

## Docs

n/a. refactor / bug fix.

PR #7123, opened 2023-05-18
fix(prefetch): Fix the inclusion of `@types/network-information`

connor-baer:
## Changes

https://github.com/withastro/astro/pull/7104 added the `"files"` field to all packages to limit the files that are published to npm. The `@astro/prefetch` package only includes the `dist` folder, however, the code tries to import from the `@types` folder, causing the following error:

```
Could not resolve "../@types/network-information.d.ts" from "node_modules/@astrojs/prefetch/dist/client.js"
```

~I added the `@types` folder to the `"files"` field to fix it.~

I changed the import to a triple-slash directive as [suggested](https://github.com/withastro/astro/pull/7123#issuecomment-1554172407) by @bluwy.

## Testing

I did not test the change as it seems straightforward.

## Docs

This fixes an expected behavior, so an update to the docs doesn't seem necessary.

### DIFF
--- a/.changeset/brown-goats-serve.md
+++ b/.changeset/brown-goats-serve.md
@@ -3,4 +3,4 @@
 '@astrojs/prefetch': patch
 ---
 
-Include the `@types` directory in the published `"files"`.
+Fix the inclusion of `@types/network-information`.
--- a/.changeset/brown-goats-serve.md
+++ b/.changeset/brown-goats-serve.md
@@ -1,6 +1,6 @@
-
 ---
 '@astrojs/prefetch': patch
 ---
 
 Fix the inclusion of `@types/network-information`.
+
--- a/.changeset/brown-goats-serve.md
+++ b/.changeset/brown-goats-serve.md
@@ -1,0 +1,6 @@
+
+---
+'@astrojs/prefetch': patch
+---
+
+Include the `@types` directory in the published `"files"`.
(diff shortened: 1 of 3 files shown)


bluwy:
Hi, thanks for catching this. I didn't expect the types file to be imported, and it doesn't seem right that it's imported by a JS file. 

I think instead we can remove the import here:
https://github.com/withastro/astro/blob/31cbf4357e866abab62bb4ccf949f5a59aec06ac/packages/integrations/prefetch/src/client.ts#L2
and add this to the top of the file instead.
```js
/// <reference types="../@types/network-information.d.ts" />
```

connor-baer:
@bluwy That makes sense! I've changed the approach as suggested.

.changeset/brown-goats-serve.md
```diff
@@ -0,0 +1,6 @@
+
+---
+'@astrojs/prefetch': patch
+---
+
+Fix the inclusion of `@types/network-information`.
```
bluwy, line 6:
```suggestion
---
'@astrojs/prefetch': patch
---

Fix the inclusion of `@types/network-information`.

```

small formatting tweak so i hope it doesn't trip up changesets

millette:
Are released automated/on a schedule? Downgrading for now.

bluwy:
I've cut a new release to fix it.

Issue #7073, opened 2023-05-12
Too many open files / UnknownContentCollectionError: site does not build


fviolette:
### What version of `astro` are you using?

2.4.5

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Max

### What browser are you using?

Chrome

### Describe the Bug

Site does not build using Astro 2.4.5 (with the latest MDX integration).
The issue seems to start with Astro 2.4.4.

I've used the docs site as an example.
My private site has the same issue.

`EMFILE: too many open files, open '/home/projects/github-xwpzeh/src/pages/[lang]/[...slug].astro'`

`UnknownContentCollectionError: [object Object]
    at Module.getEntrySlug (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/astro@2.4.5_@types+node@18.6.4_sass@1.54.3/node_modules/astro/dist/content/utils.js:298:11)
    at async eval (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/astro@2.4.5_@types+node@18.6.4_sass@1.54.3/node_modules/astro/dist/content/vite-plugin-content-virtual-mod.js:99:20)
    at async Promise.all (index 103)
    at async getStringifiedLookupMap (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/astro@2.4.5_@types+node@18.6.4_sass@1.54.3/node_modules/astro/dist/content/vite-plugin-content-virtual-mod.js:90:3)
    at async Context.load (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/astro@2.4.5_@types+node@18.6.4_sass@1.54.3/node_modules/astro/dist/content/vite-plugin-content-virtual-mod.js:52:36)
    at async Object.load (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/vite@4.3.2_@types+node@18.6.4_sass@1.54.3/node_modules/vite/dist/node/chunks/dep-7efa13d7.js:42887:32)
    at async loadAndTransform (file://file:///home/projects/github-xwpzeh/node_modules/.pnpm/vite@4.3.2_@types+node@18.6.4_sass@1.54.3/node_modules/vite/dist/node/chunks/dep-7efa13d7.js:53326:24)`
    
The troubleshooting link provided is https://docs.astro.build/en/reference/errors/unknown-content-collection-error/

### Link to Minimal Reproducible Example

[...]

PR #7125, opened 2023-05-19
Make vite-plugin-content-virtual-mod run `getEntrySlug` 10 at a time

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7073

Make vite-plugin-content-virtual-mod run `getEntrySlug` serially as it reads the filesystem. Running in parallel all at once could cause the OS to lock reading files.

## Testing

Manually tested with docs to see if there's a first-page load slowdown. I didn't notice a big difference so this should be good.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/dull-teachers-work.md
+++ b/.changeset/dull-teachers-work.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Make vite-plugin-content-virtual-mod run `getEntrySlug` serially to prevent `EMFILE: too many open files` error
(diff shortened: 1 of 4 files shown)


ematipico (approved):
The code looks good. Should we make a preview release and ask users to test it to make sure that this is the correct fix?

bluwy:
!preview content-serial

matthewp (approved):

ematipico:
I wonder if the issue really the watcher and how many files we are processing... since the error says "too many **open** files", maybe we should use try to use [`fs.close()`](https://nodejs.org/api/fs.html#filehandleclose) to close the file at the end of the operation.

PR #7128, opened 2023-05-19
Fix injectRoute for SSR

johannesspohr:
Use manifest routes for SSR app manifest instead of page components to enable injected routes with SSR.

I don't know whether this is the right way to handle it, but it seemed logical to me looking at the data structures.

## Changes

- Before, routes created by `injectRoute` didn't work for SSR. Now the injected manifest reuses the routes form the manifest, which also included injected routes.

## Testing

One test added to the `ssr-dynamic` suite

## Docs

Docs shouldn't be affected

### DIFF
--- a/packages/astro/src/core/build/plugins/plugin-ssr.ts
+++ b/packages/astro/src/core/build/plugins/plugin-ssr.ts
@@ -151,34 +151,29 @@ function buildManifest(
 		}
 	};
 
-	for (const pageData of eachPageData(internals)) {
-		if (!pageData.route.prerender) continue;
-		if (!pageData.route.pathname) continue;
+	for (const route of opts.manifest.routes) {
+		if (!route.prerender) continue;
+		if (!route.pathname) continue;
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -151,34 +151,29 @@ function buildManifest(
...
 	}
 
-	for (const pageData of eachPageData(internals)) {
-		if (pageData.route.prerender) continue;
+	for (const route of opts.manifest.routes) {
+		const pageData = internals.pagesByComponent.get(route.component);
+		if (route.prerender) continue;
+		if (!pageData) {
+			continue;
+		}
```
ematipico, line 176:
```suggestion
		if (route.prerender || !pageData) continue;
```

ematipico (commented):
Code looks good. I defer the final decision to @matthewp

PR #7129, opened 2023-05-19
Fix e2e flaky tests again

bluwy:
## Changes

- Fix `client:visible` test by manually scrolling the view ourselves
- Assign unique port for each e2e tests. I found even if they're running serially, they could still somehow conflict.
  - This technically means we can run the tests in parallel now, but enabling it still leave some tests flaky, especially the HMR tests and the Solid nested React test (`TypeError: Cannot read properties of undefined (reading 'getCurrentStack')` error)

## Testing

Ran e2e tests locally.

## Docs

n/a. test refactor.

### DIFF
--- a/packages/astro/playwright.config.js
+++ b/packages/astro/playwright.config.js
@@ -24,9 +24,6 @@ const config = {
 	use: {
 		/* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
 		actionTimeout: 0,
-		/* Base URL to use in actions like `await page.goto('/')`. */
-		baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL || 'http://localhost:3000',
-
 		/* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
 		trace: 'on-first-retry',
 	},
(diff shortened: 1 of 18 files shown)


packages/astro/playwright.config.js
```diff
@@ -24,9 +24,6 @@ const config = {
 	use: {
 		/* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
 		actionTimeout: 0,
-		/* Base URL to use in actions like `await page.goto('/')`. */
-		baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL || 'http://localhost:3000',
-
```
bluwy, line 29: Remove this `baseURL` to discourage navigating `goto` without a base url / host (e.g. always use `astro.resolveUrl`). So that each test can navigate with their own unique port.

packages/astro/test/test-utils.js
```diff
@@ -151,7 +151,7 @@ export async function loadFixture(inlineConfig) {
 	};
 
 	const resolveUrl = (url) =>
-		`http://${config.server.host}:${config.server.port}${url.replace(/^\/?/, '/')}`;
+		`http://${config.server.host || 'localhost'}:${config.server.port}${url.replace(/^\/?/, '/')}`;
```
bluwy, line 154: Sometimes `config.server.host` can be `false`, so fallback to `localhost` in that case.

ematipico (approved):
Hopefully, peace and consistency will reign again!

matthewp (approved):

PR #7136, opened 2023-05-19
Parallel array rendering

### LINKED ISSUE #7127: Parallel rendering does not respect arrays

2.5.0

None

npm

Mac

Chrome

### Describe the Bug

I tested the version 2.5.0 with parallel component rendering with the setup where I have originally discovered the parallel rendering issue and it does not seem to be fully fixed yet. It does not work properly with `array.map` async fragments.

Here is the rendering diagram:

image

I could easily reproduce this in a generic astro application.

```astro
// AsyncComponent.astro

---
export interface Props {
  delay: number;
}

const {delay} = Astro.props as Props;

await new Promise(resolve => setTimeout(resolve, delay));
---
<div>{delay}</div>
```

```astro
// index.astro

---
import Layout from '../layouts/Layout.astro';
import Card from '../components/Card.astro';
import AsyncComponent from '../components/AsyncComponent.astro';

const asyncComponentDelays = [500, 1000, 1000, 2000, 1000];
---

<Layout title="Welcome to Astro.">
	<main>
		<h1>Welcome to <span class="text-gradient">Astro</span></h1>
		<p class="instructions">
			To get started, open the directory <code>src/pages</code> in your project.<br />
			<strong>Code Challenge:</strong> Tweak the "Welcome to Astro" message above.
		</p>
		
		{asyncComponentDelays.map((delay) => <AsyncComponent delay={delay} />)}

// ...
```

The result is 5.5s rendering (expected 2s):

image

If I add the following components

```astro
// index.astro
// ...
		<AsyncComponent delay={1000} />
		<AsyncComponent delay={1000} />
		<AsyncComponent delay={1000} />

		{asyncComponentDelays.map((delay) => <AsyncComponent delay={delay} />)}
// ...
```

Then it results in 6.5s rendering time, which also should not be the case (should remain 5.5s, of course if 5.5s were the right time):

image

Is there something to work this around?

https://github.com/smnbbrv/astro-parallel-rendering-issue

### Participation

johannesspohr:
## Changes

This uses the existing rendering parallelization for array children as well
Fixes #7127

## Testing

Extended the parallel rendering test

## Docs

Changeset added


packages/astro/src/runtime/server/render/astro/render-template.ts
```diff
@@ -36,23 +36,15 @@ export class RenderTemplateResult {
...
+		let iterables = bufferIterators(expressions.map((e) => renderChild(e)));
 		for (let i = 0; i < htmlParts.length; i++) {
 			const html = htmlParts[i];
 			const iterable = iterables[i];
 
 			yield markHTMLString(html);
-			yield* iterable;
+			if (iterable) {
+				yield* iterable;
+			}
```
bluwy, line 47: Is there a reason this is wrapped with an `if`?

bluwy (commented):
One small question, but the fix looks great!

packages/astro/src/runtime/server/render/astro/render-template.ts
johannesspohr: I'm glad you ask :-)
There might be more `htmlParts` than `expressions`, and so far it didn't matter because apparently `renderChild` doesn't mind undefined values. However, if we map the `expressions` based on their length, and not the length of `htmlElements` as before, we will end up with less entries, and while we can pass `undefined` to `renderChild`, it will crash if we `yield* undefined`. That's why we need an additional check here.

johannesspohr: `renderChild(undefined)` will `yield undefined` (TIL you can `await undefined`), so that's why it worked before. Do you think my change will improve the readability?

ematipico (approved):
Looks good to me!

Issue #6635, opened 2023-03-23
image asset: ENOENT error for remote markdown files with relative paths


mashehu:
### What version of `astro` are you using?

2.1.5

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

MacOS

### What browser are you using?

Chrome

### Describe the Bug

With the new image assets, relative links to images in a remote markdown are not resolved well and generate a ENOENT error which stops the server. See example below

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-mnzpmy-qdvhnk?file=astro/package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
It's unfortunately unlikely that we can support this. Importing files from outside the Astro project is not a supported use case we have at the moment, and is unfortunately unlikely to happen to be due to technical restrictions regarding where files can live.

PR #7139, opened 2023-05-19
feat: support images outside the project

Princesseuh:
## Changes

Changes the returned `src` for imported images to be absolute path with a Vite `/@fs/` thing so we can support imports from outside the project. This also allows us to add an error message when someone make the common mistake of passing an imported image's `src` to `Image` and `getImage` instead of passing the full import

This PR also includes some minor fixes to error messages that I noticed while adding a new error.

Fix https://github.com/withastro/astro/issues/6635

## Testing

Added tests, modified tests that tested the URL precisely, and tests should still pass!

## Docs

All the usages of `src` that are documented should still work, so no change necessary.


packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -107,13 +107,13 @@ export default function assets({
...
 						const url = new URL(req.url, 'file:');
-						const filePath = url.searchParams.get('href');
+						const filePath = url.searchParams.get('href')?.slice('/@fs'.length);
 
 						if (!filePath) {
 							return next();
 						}
 
-						const filePathURL = new URL('.' + filePath, settings.config.root);
+						const filePathURL = new URL('.' + filePath, 'file:');
```
matthewp, line 116: Won't this make it so that it's always relative to your filesystem root?

matthewp: Maybe it's becoming relative to the `pwd`?

Princesseuh: > Won't this make it so that it's always relative to your filesystem root?

It's always relative to root, but this is not a problem as far as I could tell

ematipico (approved):

bluwy (approved):
The `/@fs` handling looks good to me 👍

matthewp (approved):

packages/astro/src/assets/utils/emitAsset.ts
```diff
@@ -40,34 +39,14 @@ export async function emitESMImage(
 		url.searchParams.append('origHeight', meta.height.toString());
 		url.searchParams.append('origFormat', meta.format);
 
-		meta.src = rootRelativePath(settings.config, url);
+		meta.src = `/@fs` + prependForwardSlash(fileURLToNormalizedPath(url));
```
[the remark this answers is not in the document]
Princesseuh: Hmm, I'm willing to take the risk. I think the consistency of it is great, even though I don't necessarily want users to rely on it. Will definitely reconsider if it becomes an issue!

PR #7141, opened 2023-05-19
Fix: Heading ID CI flakiness

bholmesdev:
## Changes

- Switch from a global `headingSlugger` to a context variable scoped to the config. Adding arbitrary variables to the Markdoc config _was_ endorced by the Stripe team, who uses a `services` pattern for context. We've used `ctx` here.
- Rename `applyDefaultConfig()` to `setupConfig()`. Better captures why this function should be run for each file's transform, and the fact that it sets up context

## Testing

Add test for heading cache leaks for documents with the same heading IDs

## Docs

N/A

### DIFF
--- a/.changeset/eleven-tables-speak.md
+++ b/.changeset/eleven-tables-speak.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdoc': patch
+---
+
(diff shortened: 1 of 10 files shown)


natemoo-re (approved):
Nice tests, this looks good!

matthewp (approved):

PR #7107, opened 2023-05-16
Client-side Routing with `Outlet`

natemoo-re:
## Changes

Just opening for preview release

## Testing

Manually

## Docs

N/A

PR #7145, opened 2023-05-19
Add experimental SPA router

natemoo-re:
## Changes

- Adds an experimental `router: 'mpa' | 'spa'` option to the config
- SPA router persists matching islands by default
- SPA router has a built-in `route-announcer` for accessibility
- SPA router fetches next page and automatically merges into current page efficiently
- SPA router is controlled by `data-` attributes on the `<a>` tag.
  - `<a data-router-ignore>` will skip SPA navigation and perform a traditional page load
  - `<a data-router-noscroll>` will not scroll to the top of the page upon navigation
  - `<a data-router-keepfocus>` will not reset focus upon navigation
- TODO: resource prefetching for the incoming page
- TODO: resource caching during router lifetime
- TODO: imperative router API

Replaces exploration done in #7107

## Getting Started

This PR is still a work in progress. If you'd like to give feedback, here's how you can try it out!

**Install the preview release**

```bash
npm i astro@next--spa
```

**Update your `astro.config.mjs` file**

```js
export default defineConfig({
  experimental: {
    router: 'spa'
  }
})
```

**Have fun!**

## Testing

Manually for now

## Docs

Tracking in https://github.com/withastro/docs/issues/3314

### DIFF
--- a/.changeset/nice-eels-thank.md
+++ b/.changeset/nice-eels-thank.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Enable experimental SPA router
--- a/examples/blog/astro.config.mjs
+++ b/examples/blog/astro.config.mjs
@@ -7,4 +7,7 @@ import sitemap from '@astrojs/sitemap';
 export default defineConfig({
 	site: 'https://example.com',
 	integrations: [mdx(), sitemap()],
+	experimental: {
+		router: 'spa'
+	}
 });
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -146,7 +146,7 @@
     "js-yaml": "^4.1.0",
     "kleur": "^4.1.4",
     "magic-string": "^0.27.0",
-    "micromorph": "^0.4.4",
+    "micromorph": "^0.4.5",
     "mime": "^3.0.0",
     "ora": "^6.1.0",
     "path-to-regexp": "^6.2.1",
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -146,6 +146,7 @@
     "js-yaml": "^4.1.0",
     "kleur": "^4.1.4",
     "magic-string": "^0.27.0",
+    "micromorph": "^0.4.4",
     "mime": "^3.0.0",
     "ora": "^6.1.0",
     "path-to-regexp": "^6.2.1",
--- a/packages/astro/src/core/config/schema.ts
+++ b/packages/astro/src/core/config/schema.ts
@@ -43,6 +43,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 		customClientDirectives: false,
 		inlineStylesheets: 'never',
 		middleware: false,
+		router: 'mpa'
 	},
 };
 
@@ -209,6 +210,10 @@ export const AstroConfigSchema = z.object({
 				.default(ASTRO_CONFIG_DEFAULTS.experimental.inlineStylesheets),
 			middleware: z.oboolean().optional().default(ASTRO_CONFIG_DEFAULTS.experimental.middleware),
 			hybridOutput: z.boolean().optional().default(ASTRO_CONFIG_DEFAULTS.experimental.hybridOutput),
+			router: z
+				.enum(['mpa', 'spa'])
+				.optional()
+				.default(ASTRO_CONFIG_DEFAULTS.experimental.router),
 		})
 		.passthrough()
 		.refine(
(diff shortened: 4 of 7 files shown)


natemoo-re:
!preview spa

NaCoLiu:
why draft bro?

MoustaphaDev:
Hey @NaCoLiu! It's a PR in progress, it will be ready for review when all the necessary changes and features have been implemented

tony-sull:
> TODO: resource prefetching for the incoming page

I wonder if this could be used in conjunction with `@astrojs/prefetch` for page and resource prefetching?

kyjus25:
Dude it transitions too!?!?!?  CLEAN 🔥

I have a complex dynamic layout mechanic i'm testing out where the layout changes based on a query param in the URL. Like, dynamic astro file imports and stuff. And I was like... man, Astro works better than any of the other frameworks out there for my very niche use case, I just wish it CSR'd. And I stumbled upon your PR. And I tried it. And my jaw legitimately dropped. To the floor.

VERY excited for this PR. Round of applause and hats off to you 👏

francoromanol:
This is amazing 👏! One question, is there any way to manage/remove transitions?

NaCoLiu:
https://github.com/withastro/docs/issues/3314#issuecomment-1560555356

unbiased-dev:
From what I can tell currently the whole site is replaced with the new page while the islands persist which is cool.

Partial updates would be better ofc but the current main concern atm is the transition animations.

Currently they apply to all elements including the persisted islands, which is a bit awkward _because_ they are persistent. 
Like if you have a counter and you navigate to another page, the counter state stays but only after the element disappears and reappears. This would be less noticeable if islands could be able to opt-out of the transition animation.

Although that could theoretically be even worse if the location of said island is in a different place in the DOM causing odd layout shifts.

I think the API that you were tinkering with via `Outlet`s feels cleaner and gives more flexibility and control over said transitions and hydrations. This would also give more control over which islands actually keep state, as currently unless you opt out of CSR via the `data-router-ignore` which would cause a full page reload, you can't control whether you want an island to not be persistent.

Just my current observations, I might be off course on the end goal here.

Brian-McBride:
Is there a way to help/contribute to this feature? 

I think, like others who are interested, that while there is a lot of value in MPA patterns for some sites - there is also value in client-side routing. My dev teams have been doing a few client jobs with Next's APP router and React's Server Components are turning out to be a mess. But, those projects still need the CSR for the "app" portions of the platform. 

The idea of having islands remain mounted is just awesome.

JLarky:
I plaid with this a bit. So far my feedback:
- I wish I would opt into this instead of it current behavior where it intercepts everything (and causes issues), like if I could mark pages where I want to use it, similar to `export const csr = false;` in SvelteKit (https://kit.svelte.dev/docs/page-options)
- currently there's no way to interact with the router programmatically, I tried `history.replaceState(123, null)` and it did nothing. So instead I created invisible link with `data-router-noscroll data-router-keepfocus` and called it with `click()` method, which did in fact call `navigate` function, but it wasn't the best "API" :)
- So I tried it on a route where I have an SPA and it went into infinite loop once I clicked on `Link` component in the app 😬
- Also navigation from page wihout react SPA to react SPA seemed to not work 100%, some stuff was stuck in "lading" state
- Overall the current behaviour means that you can't gradually implement this mode in your existing app, and that puts it into pretty much "I can probably use it for a small demo app, but if I want to use it in a real app I have to write my own router" :(
- but I see the potential :)

<details>
<summary>
example app instructions
</summary>
So I don't have the best example of issues that I had, but I pushed the code that I plaid to the github, so you can look at it here:
https://github.com/fogbender/b2b-saaskit/commits/idea--astro-spa-router
this app requires auth, so it might be hard to test everything without a full setup
because SPA (without SSR that doesn't require auth) on the http://localhost:3000/survey seem to work fine, but the SPA+SSR does require auth :(
otherwise
- clone it
- switch to `idea--astro-spa-router`
- yarn && yarn dev # if you use npm or pnpm it will pick up the wrong version of astro package
- click around :)

I'm playing with it at about 48 minutes in this video https://www.youtube.com/watch?v=7UfYD5NAl5I
</details>

matthewp:
We're not going to pursue this implementation as is. It's very likely that pieces of this will play into our eventual implementation however (the diffing part in particular). Thanks for all of the feedback everyone!

Our current plan going forward is to experiment more with the view transition APIs to see how best Astro's CSR solution can align with and extend that. So things like per-route opt-in is something we definitely agree with.

Let's continue the discussion in the stage 1 proposal: https://github.com/withastro/roadmap/issues/532

PR #7148, opened 2023-05-20
fix(prefetch): distribute `@types` dir

DerYeger:
## Changes

The `@astrojs/prefetch` integration references a file in its dist `*.js` files that is not included in NPM downloads.
Hence, [my builds are failing](https://github.com/DerYeger/jan-mueller/actions/runs/5029812698/jobs/9021789049?pr=169#step:3:78).

This PR updates the `package.json` `files` section to include the required files.

## Testing

I did not test this change in this library, as only the distribution is affected.
Instead, I replicated the change in my affected repository and verified that the fix does work.

## Docs

No changes to the docs are required, IMO.

### DIFF
--- a/.changeset/popular-pots-confess.md
+++ b/.changeset/popular-pots-confess.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/prefetch': patch
+---
+
+distribute @types dir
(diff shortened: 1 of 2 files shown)


Princesseuh:
Already covered by https://github.com/withastro/astro/pull/7123

PR #7150, opened 2023-05-20
fix(hybrid-output): no matched route when using `getStaticPaths`  

### LINKED ISSUE #7140: Pre-rendering error on dynamic routes with “hybrid” output

2.5.0

Vercel

yarn

Mac

Safari

### Describe the Bug

Implicit pre-rendering without `export const prerender = true` using the new `output: "hybrid"` mode doesn’t work for dynamic routes. An error is shown in the console, and when building the site no html files are generated for the routes.

```sh
[getStaticPaths] getStaticPaths() is ignored when "output: server" is set.
```

https://stackblitz.com/edit/github-nheqm9

### Participation

MoustaphaDev:
## Changes

Fix #7140

With the original prerendering feature, we had to export a `prerender` constant in order to prerender a page. Because of that we could get the prerendering state of a module by accessing its `prerender` export (`mod.prerender`)

With the new hybrid rendering feature however, we can't do that because it doesn't require a `prerender` export.

To get the `prerender` state of a route we should retrieve it from the metadata set by the [`astro:scanner`](https://github.com/withastro/astro/blob/5c3c672d2657d29e96aaaf7d9b1529f2ea432ce0/packages/astro/src/vite-plugin-scanner/index.ts#L8) vite plugin, which should erase inconsistencies on that front between dev and build.

## Testing

Copy the test fixtures of `ssr-prerender-get-static-paths`, and remove the `export const prerender = true` from pages/endpoints.
In this case, creating a new test fixture seems like the easier choice.

Also fixed the old prerender tests, which weren't actually testing the prerendering feature.

## Docs

Bug fix, N/A


packages/astro/src/prerender/status.ts
MoustaphaDev, line 1: Put this in its own file to avoid weird bundling issues that caused most previous CI errors.

packages/astro/src/prerender/metadata.ts
MoustaphaDev, line 1: Put this in its own file to avoid weird bundling issues that caused most previous CI errors.

packages/astro/test/ssr-prerender-get-static-paths.test.js
```diff
@@ -1,131 +1,282 @@
...
-	before(async () => {
-		fixture = await loadFixture({
-			root: './fixtures/ssr-prerender-get-static-paths/',
+describe('Prerender', () => {
+	describe('output: "server"', () => {
+		const sharedConfig = {
 			site: 'https://mysite.dev/',
+			adapter: testAdapter(),
 			base: '/blog',
+			output: 'server',
```
MoustaphaDev, line 12: The old prerendering tests didn't have `output: server`, which is needed.

packages/astro/test/ssr-prerender-get-static-paths.test.js
```diff
@@ -1,131 +1,282 @@
...
-		res = await fixture.fetch('/c');
-		expect(res.status).to.equal(200);
-	});
+				it('resolves 404 on pattern match without static path - rest params', async () => {
+					const res = await fixture.fetch('/blog/pizza/pizza-hut');
+					const html = await res.text();
 
-	describe('404 behavior', () => {
-		it('resolves 200 on matching static path - named params', async () => {
-			const res = await fixture.fetch('/pizza/provolone-sausage');
```
MoustaphaDev, line 72: paths didn't include the `base`

packages/astro/test/ssr-prerender-get-static-paths.test.js
```diff
@@ -1,131 +1,282 @@
...
+			});
 		});
 	});
 
-	describe('route params type validation', () => {
-		it('resolves 200 on nested array parameters', async () => {
-			const res = await fixture.fetch('/nested-arrays/slug1');
-			expect(res.status).to.equal(200);
-		});
+	describe('output: "hybrid"', () => {
```
MoustaphaDev, line 143: Same tests as the above. Refactoring them to reduce code length will result in less readable tests

bluwy:
> Copy the test fixtures of `ssr-prerender-get-static-paths`, and remove the `export const prerender = true` from pages/endpoints. In this case, creating a new test fixture seems like the easier choice.

We were discussing about testing recently, and would want to prevent duplicated test since it's getting slower. Would it be possible to share a single fixture to test it?

Besides that, the code changes looks good to me!

matthewp (approved):

PR #7151, opened 2023-05-20
Add MDX optimize option

bluwy:
## Changes

Add `optimize` option for `@astrojs/mdx` to optimize the MDX output for faster builds and rendering.

Most of the rehype code is copied from https://github.com/withastro/docs/blob/6b6f629e2241dce12dd4763124d00eeb223aa1f4/plugins/rehype-optimize-static.ts, but:
- Removed special heading handling as this rehype plugin will always run last
- Add custom components handling (`export const components` and from a config option)

## Testing

Added tests to `@astrojs/mdx`. Also manually tested with Astro docs.

## Docs

Added docs for the new option in the `@astrojs/mdx` README.

### DIFF
--- a/.changeset/dirty-singers-enjoy.md
+++ b/.changeset/dirty-singers-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/mdx': patch
+---
+
+Add `optimize` option for faster builds and rendering
--- a/packages/integrations/mdx/README.md
+++ b/packages/integrations/mdx/README.md
@@ -83,6 +83,7 @@ You can configure how your MDX is rendered with the following options:
 - [Options inherited from Markdown config](#options-inherited-from-markdown-config)
 - [`extendMarkdownConfig`](#extendmarkdownconfig)
 - [`recmaPlugins`](#recmaplugins)
+- [`optimize`](#optimize)
 
 ### Options inherited from Markdown config
 
@@ -185,9 +186,34 @@ We suggest [using AST Explorer](https://astexplorer.net/) to play with estree ou
 
 ### `optimize`
 
-Whether to optimize the MDX output for faster builds and rendering. An internal rehype plugin will be injected last to transform static parts of the `hast` into [`set:html` properties](https://docs.astro.build/en/reference/directives-reference/#sethtml). As this modifies the intermediate `estree` and leaves some generated HTML to be un-escaped, the option is disabled by default.
+- **Type:** `boolean | { customComponentNames?: string[] }`
 
-As the optimizer works within individual MDX files, if you're [passing custom components to an imported MDX content](https://docs.astro.build/en/guides/markdown-content/#custom-components-with-imported-mdx), you'll need to configure `customComponentNames` to prevent the optimizer from handling the custom components:
+This is an optional configuration setting to optimize the MDX output for faster builds and rendering via an internal rehype plugin. This may be useful if you have many MDX content and notice slow builds. However, this option may generate some unescaped HTML, so make sure your site's interactive parts still work correctly after enabling it.
+
+This is disabled by default. To enable MDX optimization, add the following to your MDX integration configuration:
+
+__`astro.config.mjs`__
+
+```js
+import { defineConfig } from 'astro/config';
+import mdx from '@astrojs/mdx';
+
+export default defineConfig({
+  integrations: [
+    mdx({
+      optimize: true,
+    })
+  ]
+});
+```
+
+### `customComponentNames`
+
+- **Type:** `string[]`
+
+An optional property of `optimize` to prevent the MDX optimizer from handling any [custom components passed to imported MDX content](https://docs.astro.build/en/guides/markdown-content/#custom-components-with-imported-mdx).
+
+You will need to exclude these components from optimization as the optimizer eagerly converts content into a static string, which will break custom components that needs to be dynamically rendered. For example, with the below configuration, the MDX output should be `<Heading>...</Heading>` instead of `"<h1>...</h1>"`.
 
 ```astro
 ---
@@ -197,6 +223,10 @@ import Heading from '../Heading.astro';
 
 <Content components={{...components, h1: Heading }} />
 ```
+
+Note that if you use `export const components = { ... }` in your MDX file to configure custom components, you don't need to manually configure this option as the optimizer will automatically detect them.
+
+To configure this option, specify an array of HTML element names that are custom components:
 
 __`astro.config.mjs`__
 
@@ -215,9 +245,6 @@ export default defineConfig({
   ]
 });
 ```
-
-> **Note**
-> If you're using `export const components = { ... }` in your MDX file, the optimizer will automatically detect and add them to `customComponentNames` so you don't have to manually configure it.
 
 ## Examples
 
--- a/packages/integrations/mdx/README.md
+++ b/packages/integrations/mdx/README.md
@@ -188,7 +188,7 @@ We suggest [using AST Explorer](https://astexplorer.net/) to play with estree ou
 
 - **Type:** `boolean | { customComponentNames?: string[] }`
 
-This is an optional configuration setting to optimize the MDX output for faster builds and rendering via an internal rehype plugin. This may be useful if you have many MDX content and notice slow builds. However, this option may generate some unescaped HTML, so make sure your site's interactive parts still work correctly after enabling it.
+This is an optional configuration setting to optimize the MDX output for faster builds and rendering via an internal rehype plugin. This may be useful if you have many MDX files and notice slow builds. However, this option may generate some unescaped HTML, so make sure your site's interactive parts still work correctly after enabling it.
 
 This is disabled by default. To enable MDX optimization, add the following to your MDX integration configuration:
 
@@ -211,9 +211,11 @@ export default defineConfig({
 
 - **Type:** `string[]`
 
-An optional property of `optimize` to prevent the MDX optimizer from handling any [custom components passed to imported MDX content](https://docs.astro.build/en/guides/markdown-content/#custom-components-with-imported-mdx).
+An optional property of `optimize` to prevent the MDX optimizer from handling any [custom components passed to imported MDX content via the components prop](https://docs.astro.build/en/guides/markdown-content/#custom-components-with-imported-mdx).
 
-You will need to exclude these components from optimization as the optimizer eagerly converts content into a static string, which will break custom components that needs to be dynamically rendered. For example, with the below configuration, the MDX output should be `<Heading>...</Heading>` instead of `"<h1>...</h1>"`.
+You will need to exclude these components from optimization as the optimizer eagerly converts content into a static string, which will break custom components that needs to be dynamically rendered. 
+
+For example, the intended MDX output of the following is `<Heading>...</Heading>` in place of every `"<h1>...</h1>"`:
 
 ```astro
 ---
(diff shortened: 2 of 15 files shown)


.changeset/dirty-singers-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/mdx': patch
```
matthewp, line 2: shouldn't this be a minor?

packages/integrations/mdx/README.md
```diff
@@ -183,6 +183,42 @@ These are plugins that modify the output [estree](https://github.com/estree/estr
...
+        // Prevent the optimizer from handling `h1` elements
+        customComponentNames: ['h1'],
+      },
+    })
+  ]
+});
+```
+
+> **Note**
+> If you're using `export const components = { ... }` in your MDX file, the optimizer will automatically detect and add them to `customComponentNames` so you don't have to manually configure it.
```
sarah11918, line 220: I get this, and I think this is fine! But, an explanation above will help the reader understand why they shouldn't want these components optimized. Optimization always sounds like a good thing!  :smile:

sarah11918 (commented):
Thanks @bluwy ! I left some suggestions for your consideration below because I think we can try to make the explanations less implementation-y and more explanatory to the reader when and why to use this config option! See what you think, and let me know where I've gotten things wrong or missed something!  :sweat_smile:

.changeset/dirty-singers-enjoy.md
ematipico: Breaking change -> major

I also think this should be a **minor**.

bluwy: I'm confused. There's no breaking changes, and `@astrojs/mdx` is v0. So breaking changes -> minor, features/fixes -> patch?

packages/integrations/mdx/README.md
```diff
@@ -183,6 +183,42 @@ These are plugins that modify the output [estree](https://github.com/estree/estr
 
 We suggest [using AST Explorer](https://astexplorer.net/) to play with estree outputs, and trying [`estree-util-visit`](https://unifiedjs.com/explore/package/estree-util-visit/) for searching across JavaScript nodes.
 
+### `optimize`
```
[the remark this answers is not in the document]
bluwy: Good catch! I'm thinking of putting this last as a _special feature_ we can point people to turn on if they're having issues. Ideally in the next breaking release when we get good feedback, we can make it the default so the option won't stay here long.

packages/integrations/mdx/README.md
```diff
@@ -183,6 +183,42 @@ These are plugins that modify the output [estree](https://github.com/estree/estr
...
+
+Whether to optimize the MDX output for faster builds and rendering. An internal rehype plugin will be injected last to transform static parts of the `hast` into [`set:html` properties](https://docs.astro.build/en/reference/directives-reference/#sethtml). As this modifies the intermediate `estree` and leaves some generated HTML to be un-escaped, the option is disabled by default.
```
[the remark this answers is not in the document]
bluwy: Good advice. I didn't notice the readme has the `**Type:**` pattern, I'll use that. The type is actually a `boolean | object`. I'll add a few more examples too.

I think there's still some useful information in the technical parts because it explains why the object option exists, but I'll try to move things around.

bluwy:
Thanks for the review @sarah11918. I've updated the docs with _hopefully_ friendlier explanations now 😄

packages/integrations/mdx/src/plugins.ts
```diff
@@ -145,6 +146,12 @@ export function getRehypePlugins(mdxOptions: MdxOptions): MdxRollupPluginOptions
 		// computed from `astro.data.frontmatter` in VFile data
 		rehypeApplyFrontmatterExport,
 	];
+
+	if (mdxOptions.optimize) {
+		const options = mdxOptions.optimize === true ? undefined : mdxOptions.optimize;
```
ematipico, line 151: Why do we pass `undefined` here? Can you add a comment to the code? It looks "strange" from a beginner (me) point to view

packages/integrations/mdx/src/rehype-optimize-static.ts
```diff
@@ -0,0 +1,97 @@
...
+ * do not include any MDX elements.
+ *
+ * This optimization reduces the JS output as more content are represented as a
+ * string instead, which also reduces the AST size that Rollup holds in memory.
+ */
+export function rehypeOptimizeStatic(options?: OptimizeOptions) {
+	return (tree: any) => {
+		// Read `export const components` to make sure the components are not optimized.
+		const customComponentNames = new Set<string>(options?.customComponentNames);
+		for (const child of tree.children) {
```
ematipico, line 24: What do we collect in this `for` loop?

packages/integrations/mdx/src/rehype-optimize-static.ts
```diff
@@ -0,0 +1,97 @@
...
+				// `elementStack` from the `allPossibleElements` set.
+				if (node.type.startsWith('mdx') || isCustomComponent) {
+					for (const el of elementStack) {
+						allPossibleElements.delete(el);
+					}
+					// Micro-optimization: While this destroys the meaning of an element
+					// stack for this node, things will still work but we won't repeatedly
+					// run the above for other nodes anymore. If this is confusing, you can
+					// comment out the code below when reading.
+					elementStack.length = 0;
```
ematipico, line 57: nit: What confuses me is that we drain the `elementStack` inside an `if` block before another `if` block. **Logically**, we can enter both `if`, which means we can drain the stack, push nodes, and then do it again. 

The comment fails to address that.

packages/integrations/mdx/src/rehype-optimize-static.ts
```diff
@@ -0,0 +1,97 @@
...
+		visit(tree, {
+			enter(node) {
+				// @ts-expect-error read tagName naively
+				const isCustomComponent = node.tagName && customComponentNames.has(node.tagName);
+				// For nodes that can't be optimized, eliminate all elements in the
+				// `elementStack` from the `allPossibleElements` set.
+				if (node.type.startsWith('mdx') || isCustomComponent) {
+					for (const el of elementStack) {
+						allPossibleElements.delete(el);
+					}
```
ematipico, line 52: Are we handling cases where we have this type of code? 

```jsx
<p>
	<SomeComponent>
</p>
```

Sorry, it's not straightforward

packages/integrations/mdx/src/rehype-optimize-static.ts
```diff
@@ -0,0 +1,97 @@
...
+					elementStack.length = 0;
+				}
+				// For possible subtree root nodes, record them
+				if (node.type === 'element' || node.type === 'mdxJsxFlowElement') {
+					elementStack.push(node);
+					allPossibleElements.add(node);
+				}
+			},
+			leave(node, _, __, parents) {
+				// Similar as above, but pop the `elementStack`
```
ematipico, line 66: nit: Similar to what, sorry?

packages/integrations/mdx/src/rehype-optimize-static.ts
bluwy: I usually use https://mdxjs.com/playground/ to test it. In this case, the nodes are two nested `mdxJsxFlowElement` nodes, so this should also handle it.

I took me a long while to craft this algorithm, so happy to answer any other questions 😄

packages/integrations/mdx/src/rehype-optimize-static.ts
bluwy: Yeah the idea is that we can and want to enter both if block, because if this node is non-optimizable, that doesn't mean we can't optimize the children _into_ this node. 

For example:
```mdx
<Foo>
  <Bar>
    # this is markdown
  </Bar>
</Foo>
```

Can be turned into:
```mdx
<Foo>
  <Bar set:html="<h1>this is markdown</h1>"></Bar>
</Foo>
```

So `Bar` here is something we want to track because in the `leave` hook later, we can add it to `allPossibleElements`.

---

I don't think it's easy to explain this as a comment because this optimization only comes clear when you read it last after understanding the (unfortunately complex 😅) algorithm.

ematipico: I wouldn't mind having a `README.md` next to this file that explains a bit about what we do :) it's not always easy to explain what's on our mind in a few comments! I completely understand

ematipico (approved):

ematipico:
Great PR! 🚀

Yan-Thomas (approved):
Docs LGTM!

packages/integrations/mdx/README.md
```diff
@@ -183,6 +184,68 @@ These are plugins that modify the output [estree](https://github.com/estree/estr
 
 We suggest [using AST Explorer](https://astexplorer.net/) to play with estree outputs, and trying [`estree-util-visit`](https://unifiedjs.com/explore/package/estree-util-visit/) for searching across JavaScript nodes.
 
+### `optimize`
+
+- **Type:** `boolean | { customComponentNames?: string[] }`
+
+This is an optional configuration setting to optimize the MDX output for faster builds and rendering via an internal rehype plugin. This may be useful if you have many MDX content and notice slow builds. However, this option may generate some unescaped HTML, so make sure your site's interactive parts still work correctly after enabling it.
```
sarah11918, line 191:
```suggestion
This is an optional configuration setting to optimize the MDX output for faster builds and rendering via an internal rehype plugin. This may be useful if you have many MDX files and notice slow builds. However, this option may generate some unescaped HTML, so make sure your site's interactive parts still work correctly after enabling it.
```
either "many MDX files" or "a large amount of MDX content" works here! Your choice!

Also, I KNOW people are going to ask "what's many? what's a lot?" They ALWAYS do...  :sweat_smile: But, I'm not going to care right now!

packages/integrations/mdx/README.md
```diff
@@ -183,6 +184,68 @@ These are plugins that modify the output [estree](https://github.com/estree/estr
...
+  integrations: [
+    mdx({
+      optimize: {
+        // Prevent the optimizer from handling `h1` elements
+        customComponentNames: ['h1'],
+      },
+    })
+  ]
+});
+```
```
sarah11918, line 247:
``````suggestion
- **Type:** `string[]`

An optional property of `optimize` to prevent the MDX optimizer from handling any [custom components passed to imported MDX content via the components prop](https://docs.astro.build/en/guides/markdown-content/#custom-components-with-imported-mdx).

You will need to exclude these components from optimization as the optimizer eagerly converts content into a static string, which will break custom components that needs to be dynamically rendered. 

For example, the intended MDX output of the following is `<Heading>...</Heading>` in place of every `"<h1>...</h1>"`:

```astro
---
import { Content, components } from '../content.mdx';
import Heading from '../Heading.astro';
---

<Content components={{...components, h1: Heading }} />
```

To configure optimization for this using the `customComponentNames` property, specify an array of HTML element names that should be treated as custom components:

__`astro.config.mjs`__

```js
import mdx from '@astrojs/mdx';

export default defineConfig({
  integrations: [
    mdx({
      optimize: {
        // Prevent the optimizer from handling `h1` elements
				// These will be treated as custom components
        customComponentNames: ['h1'],
      },
    })
  ]
});
```

Note that if your MDX file [configures custom components using `export const components = { ... }`](https://docs.astro.build/en/guides/markdown-content/#assigning-custom-components-to-html-elements), 
 then you do not need to manually configure this option. The optimizer will automatically detect them.
``````

See what you think about this! I think swapping the order and leaving the note at the bottom lets the example not be interrupted.

sarah11918 (commented):
Looks great, @bluwy! Thanks for revisiting this! Just left a couple of thoughts!  :raised_hands:

bluwy:
Thanks @sarah11918. The couple suggestions look great to me. I made some small whitespace adjustments before merging your suggestions.

PR #7166, opened 2023-05-22
refactor: move renderers to their own plugin

ematipico:
## Changes

This PR moves the code generated for the renderers in its own plugin.

The reason why this is required is that this is code that will be shared among pages, so it makes more sense to have an entry point just for that.

This will be helpful shortly when we'll split pages into multiple entry points, and they will need to import the same code

## Testing

The current tests should pass

## Docs

N/A


packages/astro/test/sourcemap.test.js
```diff
@@ -14,9 +14,4 @@ describe('Sourcemap', async () => {
 		const counterMap = dir.find((file) => file.match(/^Counter\.\w+\.js\.map$/));
 		expect(counterMap).to.be.ok;
 	});
-
-	it('Builds non-empty sourcemap', async () => {
```
ematipico, line 18: @bluwy I noticed you were the last person to change this file. Could you please confirm that this is no longer true? Now the `entry.mjs` imports `renderers.mjs`, which means we should have `sources: []`?

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -25,8 +26,12 @@ function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): V
 		async load(id) {
 			if (id === resolvedPagesVirtualModuleId) {
 				let importMap = '';
-				let imports = [];
+				const imports = [];
+				const exports = [];
+				const content = [];
```
bluwy, line 31: nit: I think we can type this as `string[]` since it can't be inferred by default.
```suggestion
				const imports: string[] = [];
				const exports: string[] = [];
				const content: string[] = [];
```

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -36,31 +41,15 @@ function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): V
...
-
-export const pageMap = new Map([${importMap}]);
-export const renderers = [${rendererItems}];
-${opts.settings.config.experimental.middleware ? `export const middleware = _middleware;` : ''}
-`;
+				content.push(`export const pageMap = new Map([${importMap}]);`);
 
-				return def;
+				const result = [imports.join('\n'), content.join('\n'), exports.join('\n')];
+				return result.join('\n');
```
bluwy, line 52: Maybe we should directly construct the string here instead of creating a new array and joining it? e.g.
```suggestion
				return `\
${imports.join('\n')}
${content.join('\n')}
${exports.join('\n')}`
```

Made a [perf test](https://perf.link/#eyJpZCI6InMxajVtbDIxYWF4IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGZvbyA9IFwiTG9yZW0gSXBzdW0gaXMgc2ltcGx5IGR1bW15IHRleHQgb2YgdGhlIHByaW50aW5nIGFuZCB0eXBlc2V0dGluZyBpbmR1c3RyeS4gTG9yZW0gSXBzdW0gaGFzIGJlZW4gdGhlIGluZHVzdHJ5J3Mgc3RhbmRhcmQgZHVtbXkgdGV4dFwiXG5jb25zdCBiYXIgPSBcIkxvcmVtIElwc3VtIGlzIHNpbXBseSBkdW1teSB0ZXh0IG9mIHRoZSBwcmludGluZyBhbmQgdHlwZXNldHRpbmcgaW5kdXN0cnkuIExvcmVtIElwc3VtIGhhcyBiZWVuIHRoZSBpbmR1c3RyeSdzIHN0YW5kYXJkIGR1bW15IHRleHRcIiIsInRlc3RzIjpbeyJuYW1lIjoiVGVzdCBDYXNlIiwiY29kZSI6ImZvbyArICdcXG4nICsgYmFyIiwicnVucyI6WzQ1OTMwMDAsNTY4MDAwLDEyMDkwMDAsNjg4MDAwMCw2ODU2MDAwLDQ1NTYwMDAsNjk4MDAwLDI5NjMwMDAsMjk2MTAwMCw4NzgwMDAsOTA0MzAwMCwzMDY2MDAwLDczMDAwMDAsNjI3NzAwMCw3ODIyMDAwLDYyOTMwMDAsNjEyMjAwMCwxMTkxODAwMCwyNDkyMDAwLDY0NTAwMCw3MjU2MDAwLDQwMDIwMDAsNTc2MzAwMCw0MTMyMDAwLDMxOTIwMDAsNTIzNTAwMCw2NzkwMDAsNjE5MDAwMCwyNTI4MDAwLDUxMTgwMDAsMzAyODAwMCwxMDI5NzAwMCwzOTkzMDAwLDI3MDAwMDAsMTA0MjcwMDAsMjM4MjAwMCw3NDk3MDAwLDMzODIwMDAsOTA4NDAwMCwzNTM1MDAwLDY1MjAwMDAsNjgxODAwMCwxMjMzNTAwMCwxOTkxMDAwLDcxNTIwMDAsMzYwMTAwMCw3NTUyMDAwLDY4NTUwMDAsMTU2MzcwMDAsMTEyNzYwMDAsMjcyNTAwMCwxNjE0MjAwMCwxMzc1MzAwMCw2NDgwMDAwLDY1NjkwMDAsNTM4MTAwMCwyMjU2MDAwLDE0NTY1MDAwLDEwOTgwMDAsNjk2NDAwMCw2NTMzMDAwLDEzMDU2MDAwLDQ2NjAwMDAsMTY1NDAwMCwxMjI1NzAwMCw1NDEwMDAwLDkwMDkwMDAsNjc3OTAwMCwyMTUwMDAsMTU3NDMwMDAsNzQyMDAwMCwxNjg4NDAwMCw4OTY5MDAwLDY5NTAwMCwxMjY0OTAwMCw1OTMyMDAwLDE2MDYwMDAwLDg0ODkwMDAsNzg2MDAwLDExNTM1MDAwLDM1MDcwMDAsMTU3MTkwMDAsNDk3NDAwMCw4ODYwMDAwLDU4OTAwMDAsODc1MDAwLDQ5NTAwMCwxMDA2MTAwMCw1MDI3MDAwLDUwNjAwMCw5NzM3MDAwLDUwNzEwMDAsODIyMTAwMCw3MDI5MDAwLDk1MjEwMDAsMTI1MTAwMCwxMTc2ODAwMCw5OTkwMDAwLDcyNjUwMDAsNDIzNzAwMF0sIm9wcyI6NjQzOTY5MH0seyJuYW1lIjoiVGVzdCBDYXNlIiwiY29kZSI6ImAke2Zvb31cXG4ke2Jhcn1gIiwicnVucyI6WzEwMTEwMDAsNDk4MDAwLDIxNzcwMDAsNzYzOTAwMCwxNjY2MDAwLDYxNDIwMDAsMTUwMjEwMDAsMjMyNzAwMCw0NjkzMDAwLDkzMjAwMCw3NTcxMDAwLDI1NDEwMDAsNjI0NjAwMCw0MjU1MDAwLDM5MDcwMDAsODU1ODAwMCwzNTIxMDAwLDgyNTIwMDAsNjg1OTAwMCw2NjMwMDAsNzU2MjAwMCw1MDI3MDAwLDYxMzAwMDAsMTUzNzAwMCwyMDg1MDAwLDEwMDYzMDAwLDEzMDYwMDAsNzU1ODAwMCwyNTU3MDAwLDUyNTkwMDAsNDAyMzAwMCw2NjczMDAwLDI5NDcwMDAsMTExNzAwMCw0MjMyMDAwLDM0MDYwMDAsOTE1MDAwMCw1MjEwMDAsOTg3MTAwMCw2NTU4MDAwLDUwMjgwMDAsNDAzNjAwMCw5NTIxMDAwLDMzMDUwMDAsMzg3MzAwMCwzODcyMDAwLDEyNTA1MDAwLDUyMjYwMDAsNDAzMDAwLDg5NTIwMDAsNTkzMDAwLDEwMjAwMCw1NDYzMDAwLDM0NjkwMDAsNjc0OTAwMCwyOTUwMDAwLDIyNjUwMDAsOTM3NTAwMCwxNTM2MDAwLDY5MzEwMDAsNDE5MTAwMCwxMzA1OTAwMCwzODAxMDAwLDI1NzcwMDAsNzQzNDAwMCw3MjUyMDAwLDEwNTE0MDAwLDY4MzYwMDAsMTEwNzAwMCwxNjA1NTAwMCw5Mjc5MDAwLDE5MjA2MDAwLDk1NDUwMDAsMTYzNzAwMCwxMTQ5ODAwMCw4MDM5MDAwLDE0NjE1MDAwLDc5NTEwMDAsNDEzNDAwMCwxMDUyMDAwMCw1NDgyMDAwLDE4MDQ0MDAwLDQwMzIwMDAsNzQ1NTAwMCw2MjMxMDAwLDI4NDgwMDAsOTY3MDAwLDk0OTcwMDAsNzE1NzAwMCw3MjAwMDAsODAxNDAwMCw2NjM4MDAwLDEwNDYxMDAwLDUyNjIwMDAsNjAzNDAwMCwxMzk0MDAwLDgzNDMwMDAsMTA0MTQwMDAsNzU0NTAwMCw0NjY3MDAwXSwib3BzIjo1ODg2NzAwfSx7Im5hbWUiOiJUZXN0IENhc2UiLCJjb2RlIjoiW2ZvbywgYmFyXS5qb2luKCdcXG4nKSIsInJ1bnMiOls2NjUwMDAsOTcwMDAwLDI4NzYwMDAsMjg3NjAwMCwyODc2MDAwLDIxMzIwMDAsNTE2MDAwLDE5NDgwMDAsMjQzOTAwMCw1MDIwMDAsMjg3NjAwMCwxNDczMDAwLDIzOTYwMDAsMjg3NjAwMCwyMTg3MDAwLDI4NzYwMDAsMTgyNDAwMCwyODc2MDAwLDEwNjUwMDAsNjYzMDAwLDI4NzYwMDAsMjIxMzAwMCwyODc2MDAwLDE3NjMwMDAsMjAxODAwMCwyODc2MDAwLDU2NzAwMCwyODc2MDAwLDE4MTUwMDAsMjM2NTAwMCwyMjA4MDAwLDI4NzYwMDAsMjE5MTAwMCw1NzQwMDAsMjg3NjAwMCwxNzc0MDAwLDI4NzYwMDAsODcwMDAwLDI4NzYwMDAsMjcyNzAwMCwyMzkyMDAwLDIxNzQwMDAsMjg3NjAwMCw5NTkwMDAsMjg3NjAwMCwyMzcxMDAwLDI4NzYwMDAsMTY1MTAwMCwyOTkxMDAwLDI4NzYwMDAsMTcwNzAwMCwyOTAwMDAwLDI1ODUwMDAsMTkwMDAwMCwyODc2MDAwLDY2NzAwMCwyNTcwMDAsMjg3NjAwMCw2OTIwMDAsMjg3NjAwMCwyODc2MDAwLDI4NzYwMDAsMjg3NjAwMCwxMDA2MDAwLDI4NzYwMDAsMjg3NjAwMCwyODc2MDAwLDI4NzYwMDAsMTAwMCwyODc2MDAwLDI4NzYwMDAsMTIxMDAwLDI4NzYwMDAsODI4MDAwLDI4NzYwMDAsMTUyNDAwMCwyODkzMDAwLDI4NzYwMDAsMjI4NzAwMCwyODc2MDAwLDI4NzYwMDAsMzEyNDAwMCwyNjQ1MDAwLDI4NzYwMDAsMjg3NjAwMCwyMzYzMDAwLDI1NzAwMCwyODc2MDAwLDI4NzYwMDAsNTcyMDAwLDI4NzYwMDAsOTY5MDAwLDI4NzYwMDAsMjY0MTAwMCwyODc2MDAwLDc4MjAwMCwyODc2MDAwLDI4NzcwMDAsMjg3NjAwMCwyNjI1MDAwXSwib3BzIjoyMjAwNDYwfV0sInVwZGF0ZWQiOiIyMDIzLTA1LTIzVDA4OjM2OjQyLjk1M1oifQ%3D%3D) here too

packages/astro/src/core/build/plugins/plugin-renderers.ts
```diff
@@ -0,0 +1,68 @@
...
+						imports.push(`import ${variable} from '${renderer.serverEntrypoint}';`);
+						rendererItems += `Object.assign(${JSON.stringify(renderer)}, { ssr: ${variable} }),`;
+						i++;
+					}
+
+					exports.push(`export const renderers = [${rendererItems}];`);
+
+					const result = [imports.join('\n'), exports.join('\n')];
+
+					return result.join('\n');
```
bluwy, line 47: Probably same here to construct a string directly.

bluwy (commented):
Some small nits, but I really like this refactor!

packages/astro/test/sourcemap.test.js
bluwy: Yeah it looks like with the new chunks, the sources are moved to `renderers.mjs.map` instead. Maybe we can update the test to check that file instead?

packages/astro/src/core/build/plugins/plugin-pages.ts
ematipico: The main reason I used a temporary variable is debugging purpose only - so I can log the result code.

I can remove if you think this is an hot path

bluwy: I don't think it's a hot path, but I usually lean towards writing performant code whenever possible.

packages/astro/test/sourcemap.test.js
ematipico: Great! I will do that :) thank you!

matthewp (approved):

PR #7171, opened 2023-05-23
Prevent Vite watching on Astro config load

bluwy:
## Changes

From @ematipico's suggestion https://github.com/withastro/astro/issues/7073#issuecomment-1557144748, we could prevent Vite watching when spinning up the Vite dev server.

It looks like the watching is already disabled for content collections / `astro sync` via:

https://github.com/withastro/astro/blob/6112671100122904db78aa9f880dbfe957aa0c24/packages/astro/src/core/create-vite.ts#L155

But I updated the Astro config load flow too as that will benefit from it.

## Testing

Logged `viteServer.watcher.getWatched()` to verify that directories aren't watched.

## Docs

n/a. internal improvement.

### DIFF
--- a/packages/astro/src/core/sync/index.ts
+++ b/packages/astro/src/core/sync/index.ts
@@ -65,7 +65,7 @@ export async function sync(
 	const tempViteServer = await createServer(
 		await createVite(
 			{
-				server: { middlewareMode: true, hmr: false },
+				server: { middlewareMode: true, hmr: false, watch: { ignored: ['**/*'] } },
 				optimizeDeps: { disabled: true },
 				ssr: { external: [] },
 				logLevel: 'silent',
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/sync/index.ts
```diff
@@ -65,7 +65,7 @@ export async function sync(
 	const tempViteServer = await createServer(
 		await createVite(
 			{
-				server: { middlewareMode: true, hmr: false },
+				server: { middlewareMode: true, hmr: false, watch: { ignored: ['**/*'] } },
```
bluwy, line 68: As explained in the PR description, this isn't needed, but I updated anyways so it's consistent with the `createViteLoader` code.

ematipico (approved):

PR #7173, opened 2023-05-23
Implement support for dynamic routes in redirects

matthewp:
## Changes

- Adds support for dynamic routes. Syntax is:

```js
redirects: {
  '/blog/[...slug]': '/team/articles/[...slug]'
}
```

Same as for file-based router. This does not introduce any new syntax.

## Testing

- Base test case added in core
- Netlify tested and implemented.

## Docs

Not yet, docs will be done as a separate issue.


packages/astro/src/core/build/static-build.ts
```diff
@@ -56,6 +57,11 @@ export async function viteBuild(opts: StaticBuildOptions) {
 		// Track the page data in internals
 		trackPageData(internals, component, pageData, astroModuleId, astroModuleURL);
 
+		// Redirects are config-only, do not add them as input
+		if(routeIsRedirect(pageData.route)) {
+			//continue;
```
ematipico, line 62: Should we remove the comment?

ematipico (approved):

packages/astro/src/core/build/static-build.ts
matthewp: Oops, yeah.

PR #7175, opened 2023-05-23
fix: :adhesive_bandage: update partytown options ts type

AkashRajpurohit:
## Changes

- Fixes the typescript type for partytown options, no need to pass `undefined` when you don't want to add any config

## Testing

N.A

## Docs

N.A

### DIFF
--- a/.changeset/curvy-hotels-study.md
+++ b/.changeset/curvy-hotels-study.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/partytown': minor
+---
+
+fix typescript type for partytown options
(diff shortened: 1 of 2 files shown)


.changeset/curvy-hotels-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/partytown': minor
```
bluwy, line 2:
```suggestion
'@astrojs/partytown': patch
```

I think it's a small change that a patch is fine here.

bluwy (approved):

PR #7180, opened 2023-05-23
Stabilize inline stylesheets

lilnasy:
## Changes
Stabilize `AstroConfig.build.inlineStylesheets`.

## Testing
Existing ones should pass.

## Docs
I think [Configuration reference](https://docs.astro.build/en/reference/configuration-reference/#experimentalinlinestylesheets) as-is would be sufficient.

### DIFF
--- a/.changeset/stupid-pumpkins-perform.md
+++ b/.changeset/stupid-pumpkins-perform.md
@@ -1,0 +1,24 @@
+---
+'astro': major
+---
+
+The Inline Stylesheets RFC is now stable!
+
+You can now control how Astro bundles your css with a configuration change:
+
+```ts
+export default defineConfig({
+    ...
+    build: {
+        inlineStylesheets: "auto"
+    }
+    ...
+})
+```
+
+The options include:
+- `inlineStylesheets: "never"`: This is the behavior you are familiar with. Every stylesheet is external, and added to the page via a `<link>` tag. Default.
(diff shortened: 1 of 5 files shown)


matthewp (dismissed):
We want this to be a non-breaking change for now, just out of experimental.

.changeset/stupid-pumpkins-perform.md
```diff
@@ -0,0 +1,24 @@
+---
+'astro': major
```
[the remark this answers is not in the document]
lilnasy: You're right, I forgot what major means.

packages/astro/src/@types/astro.ts
```diff
@@ -733,6 +733,27 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		serverEntry?: string;
+		/**
+		 * @docs
+		 * @name build.inlineStylesheets
+		 * @type {('always' | 'auto' | 'never')}
+		 * @default `never`
+		 * @version 2.4.0
```
sarah11918, line 741:
```suggestion
		 * @version 2.6.0
```
I think this should be 2.6 because any mention of it being an experimental option is going away. So someone with e.g. 2.4 or 2.5 but who never enabled the flag would not have access to this.

sarah11918 (approved):
Docs are good! Just update the version to reflect when it was stable, I think is the safer option?

packages/astro/src/@types/astro.ts
lilnasy: Yeah, this makes sense. Thanks!

PR #7187, opened 2023-05-23
Markdoc - Shiki

bholmesdev:
## Changes

- Add a `shiki()` extension to add syntax highlighting to Markdoc
- Introduce a new `extends` property to the Markdoc config. This accepts an array of objects to be merged and applied before user-specified config. `shiki()` is the first dogfooding!

## Testing

- Add unit tests for the Shiki syntax highlighter

## Docs

- Update README with syntax highlighting section


packages/integrations/markdoc/package.json
```diff
@@ -61,6 +61,7 @@
     "linkedom": "^0.14.12",
     "mocha": "^9.2.2",
     "rollup": "^3.20.1",
+    "shiki": "^0.14.1",
```
ematipico, line 64: Shouldn't we add `shiki` as a peer dependency instead? And made it optional via `peerDependenciesMeta`?

packages/integrations/markdoc/src/extensions/shiki.ts
```diff
@@ -0,0 +1,126 @@
...
+				'#000004': 'var(--astro-code-token-constant)',
+				'#000005': 'var(--astro-code-token-string)',
+				'#000006': 'var(--astro-code-token-comment)',
+				'#000007': 'var(--astro-code-token-keyword)',
+				'#000008': 'var(--astro-code-token-parameter)',
+				'#000009': 'var(--astro-code-token-function)',
+				'#000010': 'var(--astro-code-token-string-expression)',
+				'#000011': 'var(--astro-code-token-punctuation)',
+				'#000012': 'var(--astro-code-token-link)',
+			});
```
ematipico, line 58: It's best to move this object to the top of the file so we don't create it every time :)

packages/integrations/markdoc/src/extensions/shiki.ts
```diff
@@ -0,0 +1,126 @@
...
+		return theme;
+	}
+};
+
+/**
+ * Cache each Shiki highlighter by theme name.
+ * Note: cache only needed for dev server reloads, internal test suites, and manual calls to `Markdoc.transform` by the user.
+ * Otherwise, `shiki()` is only called once per build, NOT once per page, so a cache isn't needed!
+ */
+const highlighterCacheAsync = new Map<string, Promise<shikiTypes.Highlighter>>();
```
ematipico, line 33: It's really weird caching the **promises**. Shouldn't we cache the actual values?

packages/integrations/markdoc/src/extensions/shiki.ts
```diff
@@ -0,0 +1,126 @@
...
+					let html = highlighter.codeToHtml(attributes.content, { lang });
+
+					// Q: Could these regexes match on a user's inputted code blocks?
+					// A: Nope! All rendered HTML is properly escaped.
+					// Ex. If a user typed `<span class="line"` into a code block,
+					// It would become this before hitting our regexes:
+					// &lt;span class=&quot;line&quot;
+
+					// Replace "shiki" class naming with "astro-code"
+					html = html.replace(/<pre class="(.*?)shiki(.*?)"/, `<pre class="$1astro-code$2"`);
```
ematipico, line 101: Nit: we could store these regexes at the top of the file so they are created only once :)

ematipico (approved):

packages/integrations/markdoc/package.json
bholmesdev: Ah ha, sounds right to me! Will update

packages/integrations/markdoc/src/extensions/shiki.ts
bholmesdev: I took this from our existing implementation and it took me a bit to see _why._ In short: this should cache promises for multiple parallel requests with lower likelihood of cache misses. This benefit is lower because of my comment there, so I'd be open to refactoring

packages/integrations/markdoc/src/extensions/shiki.ts
bholmesdev: What I get for copying code Fred wrote (gottem)

packages/integrations/markdoc/README.md
```diff
@@ -203,6 +203,39 @@ export default defineMarkdocConfig({
...
+      // Default: false
+      wrap: true,
+      // Pass custom languages
+      // Note: Shiki has countless langs built-in, including `.astro`!
+      // https://github.com/shikijs/shiki/blob/main/docs/languages.md
+      langs: [],
+    })
+  ],
+})
+```
```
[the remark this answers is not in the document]
sarah11918: (And, sorry it's such a huge section of the code! I should have had more faith in you that I wasn't going to need to edit it as a whole.)  :sweat_smile:

packages/integrations/markdoc/src/extensions/shiki.ts
bluwy: Yeah we have a few shiki code copy-pasted around the codebase. We should probably consolidate it as part of the markdown packages in the future.

packages/integrations/markdoc/src/extensions/shiki.ts
bholmesdev: Gonna be some _fun_ const naming...

packages/integrations/markdoc/README.md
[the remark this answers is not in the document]
bholmesdev: This is all correct! When using Astro SSR, Markdoc **will ship Shiki as a dependency wherever you deploy.** This is different from Markdown and MDX, which don't ship Shiki (even when you use it). Hence the need for a manual install. Happy with this more explicit instruction!

Yan-Thomas (approved):
Docs LGTM!

PR #7189, opened 2023-05-23
fix: add astro-static-slot to the list of inert tags in astro css

elevatebart:
## Changes

In a recent version of Astro we introduced the `<astro-static-slot>` element. Prior to this only `<astro-slot>` and `<astro-island>`. 

When we have this kind of code, we expect the slot to not have an effect ad the div and the image to be on the same line.

```html
<div style="display: flex;">
  <astro-static-slot>
    <img src="test.jpg" />
    <div>Button Content</div>
  </astro-static-slot>
</div>
```

## Testing

It's a simple bug fix, could be tested if browser rendering had apis.

## Docs

Bug fix

### DIFF
--- a/.changeset/static-slot-css.md
+++ b/.changeset/static-slot-css.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+fix: add astro-static-slot to the list of inert tags in astro css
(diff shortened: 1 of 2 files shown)


.changeset/static-slot-css.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
bluwy, line 2:
```suggestion
'astro': patch
```

bluwy (approved):

PR #7193, opened 2023-05-24
refactor: emit pages as physical entry points

ematipico:
## Changes

This PR is another internal refactor of our internal bundling strategy.

Before, the pages were bundled in a single entry point. This entry point was called `entry.mjs`. 

With this refactor, now each page **becomes** an entry point. 

For SSG builds, this should result in **no changes** whatsoever in its business logic. The build should work as it was, and the emitted files should still work as they were. The only effective change is that Astro needs to load each entry point when generating the build.

For SSR builds, this should result in an internal change. Before, all the pages were bundled inside the `entry.mjs` file. The pages are **NOT** included in the final bundle, but they are loaded via dynamic import. This is an internal change of the business logic, and the SSR build still needs to work.

## Testing

The current tests MUST all pass.

## Docs

I created a file inside the `build/plugins/` folder. I noticed that there's some knowledge that I don't want to lose, and explaining it inside the code can be very tedious. 

Unfortunately, I don't know ALL plugins, so consider this file as a starting point and a WIP. In fact, the `plugin-ssr` does **more** than I described, so probably someone else will chime in and write down something more.

cc @withastro/maintainers-docs for feedback, but I am not really sure if we need your review for internal knowledge, I don't want to give you more work than you already have! 😅

### DIFF
--- a/packages/astro/src/core/build/plugins/plugin-middleware.ts
+++ b/packages/astro/src/core/build/plugins/plugin-middleware.ts
@@ -8,7 +8,6 @@ import type { StaticBuildOptions } from '../types';
 export const MIDDLEWARE_MODULE_ID = '@astro-middleware';
 export const RESOLVED_MIDDLEWARE_MODULE_ID = '\0@astro-middleware';
 
-let inputs: Set<string> = new Set();
 export function vitePluginMiddleware(
 	opts: StaticBuildOptions,
 	_internals: BuildInternals
@@ -42,10 +41,6 @@ export function vitePluginMiddleware(
 
 				return result.join('\n');
 			}
-		},
-
-		outputOptions(options) {
-			options.entryFileNames = 'middleware.mjs';
 		},
 	};
 }
--- a/packages/astro/src/core/build/plugins/plugin-middleware.ts
+++ b/packages/astro/src/core/build/plugins/plugin-middleware.ts
@@ -20,26 +20,14 @@ export function vitePluginMiddleware(
 			}
 		},
 
-		resolveId(id) {
+		async resolveId(id) {
 			if (id === MIDDLEWARE_MODULE_ID && opts.settings.config.experimental.middleware) {
-				return RESOLVED_MIDDLEWARE_MODULE_ID;
-			}
-		},
-
-		async load(id) {
-			if (id === RESOLVED_MIDDLEWARE_MODULE_ID && opts.settings.config.experimental.middleware) {
-				const imports: string[] = [];
-				const exports: string[] = [];
-				let middlewareId = await this.resolve(
+				const middlewareId = await this.resolve(
 					`${opts.settings.config.srcDir.pathname}/${MIDDLEWARE_PATH_SEGMENT_NAME}`
 				);
 				if (middlewareId) {
-					imports.push(`import { onRequest } from "${middlewareId.id}"`);
-					exports.push(`export { onRequest }`);
+					return middlewareId.id;
 				}
-				const result = [imports.join('\n'), exports.join('\n')];
-
-				return result.join('\n');
 			}
 		},
 	};
--- a/packages/astro/src/core/build/plugins/plugin-middleware.ts
+++ b/packages/astro/src/core/build/plugins/plugin-middleware.ts
@@ -43,6 +43,10 @@ export function vitePluginMiddleware(
 				return result.join('\n');
 			}
 		},
+
+		outputOptions(options) {
+			options.entryFileNames = 'middleware.mjs';
+		},
 	};
 }
 
--- a/packages/astro/src/core/build/plugins/plugin-middleware.ts
+++ b/packages/astro/src/core/build/plugins/plugin-middleware.ts
@@ -6,7 +6,6 @@ import type { AstroBuildPlugin } from '../plugin';
 import type { StaticBuildOptions } from '../types';
 
 export const MIDDLEWARE_MODULE_ID = '@astro-middleware';
-export const RESOLVED_MIDDLEWARE_MODULE_ID = '\0@astro-middleware';
 
 export function vitePluginMiddleware(
 	opts: StaticBuildOptions,
--- a/packages/astro/src/core/build/plugins/plugin-pages.ts
+++ b/packages/astro/src/core/build/plugins/plugin-pages.ts
@@ -13,6 +13,22 @@ export const ASTRO_PAGE_RESOLVED_MODULE_ID = '\0@astro-page:';
 // This is an arbitrary string that we are going to replace the dot of the extension
 export const ASTRO_PAGE_EXTENSION_POST_PATTERN = '@_@';
 
+/**
+ * 1. We add a fixed prefix, which is used as virtual module naming convention;
+ * 2. We replace the dot that belongs extension with an arbitrary string.
+ *
+ * @param path
+ */
+export function getVirtualModulePageNameFromPath(path: string) {
+	// we mask the extension, so this virtual file
+	// so rollup won't trigger other plugins in the process
+	const extension = extname(path);
+	return `${ASTRO_PAGE_MODULE_ID}${path.replace(
+		extension,
+		extension.replace('.', ASTRO_PAGE_EXTENSION_POST_PATTERN)
+	)}`;
+}
+
 function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): VitePlugin {
 	return {
 		name: '@astro/plugin-build-pages',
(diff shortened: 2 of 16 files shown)


packages/astro/src/core/build/static-build.ts
```diff
@@ -172,8 +175,15 @@ async function ssrBuild(
 					assetFileNames: `${settings.config.build.assets}/[name].[hash][extname]`,
 					...viteConfig.build?.rollupOptions?.output,
 					entryFileNames(chunkInfo) {
-						if (chunkInfo.facadeModuleId === resolvedPagesVirtualModuleId) {
-							return opts.buildConfig.serverEntry;
```
matthewp, line 176: This is a configurable value by the user, so we need to preserve it some how for SSR usage. If the user isn't using serverless mode I think we can still use this as the entry point probably, right?

packages/integrations/mdx/src/index.ts
```diff
@@ -12,6 +12,7 @@ import { VFile } from 'vfile';
 import type { Plugin as VitePlugin } from 'vite';
 import { getRehypePlugins, getRemarkPlugins, recmaInjectImportMetaEnvPlugin } from './plugins.js';
 import { getFileInfo, ignoreStringPlugins, parseFrontmatter } from './utils.js';
+import { extname } from 'node:path';
```
matthewp, line 15: is this unused?

packages/astro/src/core/build/static-build.ts
ematipico: This code path wasn't used for SSR but for SSG. In fact, `resolvedPagesVirtualModuleId` is the virtual module name assigned by `plugin-pages`. 

`serverEntry` is never used for SSR, so I belive we have a bug on hand (I haven't seen a test case with `serverEntry` anywhere).

matthewp: Oh right, disregard then.

packages/astro/src/core/build/static-build.ts
```diff
@@ -172,7 +176,16 @@ async function ssrBuild(
 					assetFileNames: `${settings.config.build.assets}/[name].[hash][extname]`,
 					...viteConfig.build?.rollupOptions?.output,
 					entryFileNames(chunkInfo) {
-						if (chunkInfo.facadeModuleId === resolvedPagesVirtualModuleId) {
+						if (chunkInfo.facadeModuleId?.startsWith(ASTRO_PAGE_RESOLVED_MODULE_ID)) {
+							return `${chunkInfo.facadeModuleId
```
matthewp, line 180: Can this renaming code be moved to its own function and some explanation for what it's doing?

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -1,62 +1,95 @@
...
+				const pageMap: string[] = [];
+
+				for (const path of Object.keys(allPages)) {
+					const extension = extname(path);
+					const virtualModuleName = `${ASTRO_PAGE_MODULE_ID}${path.replace(
+						extension,
+						extension.replace('.', ASTRO_PAGE_EXTENSION_POST_PATTERN)
+					)}`;
+					let module = await this.resolve(virtualModuleName);
+					if (module) {
```
bluwy, line 66: It would be nice to document why we're constructing a module name this way, and how it'll ba handled later (in `entryFileNames`)

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -77,6 +110,8 @@ const _start = 'start';
 if(_start in adapter) {
 	adapter[_start](_manifest, _args);
 }`;
+				const result = [imports.join('\n'), contents.join('\n'), content, exports.join('\n')];
+				return result.join('\n');
 			}
```
bluwy, line 115: Would be nice to also directly concat a string here 😬

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -1,54 +1,76 @@
...
+				for (const path of Object.keys(opts.allPages)) {
+					const extension = extname(path);
+
+					// we mask the extension, so this virtual file
+					// so rollup won't trigger other plugins in the process
+					const virtualModuleName = `${ASTRO_PAGE_MODULE_ID}${path.replace(
+						extension,
+						extension.replace('.', ASTRO_PAGE_EXTENSION_POST_PATTERN)
+					)}`;
+
```
bluwy, line 33: Looks like this duplicates a similar code in `plugin-ssr`. Maybe we should abstract as a function?

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -1,54 +1,76 @@
...
-					const variable = `_page${i}`;
-					imports.push(
-						`const ${variable} = () => import(${JSON.stringify(pageData.moduleSpecifier)});`
-					);
-					importMap += `[${JSON.stringify(pageData.component)}, ${variable}],`;
-					i++;
-				}
+				// split by ":", the second element is the page name, which will start with "src/..."
+				const [, pageName] = id.split(':');
+				// We replaced the `.` of the extension with ASTRO_PAGE_EXTENSION_POST_PATTERN, let's replace it back
```
bluwy, line 53: nit: usually the pattern with handling virtual files is using `id.slice(ASTRO_PAGE_RESOLVED_MODULE_ID.length)`, which I think would be safer too.

packages/astro/src/core/build/static-build.ts
```diff
@@ -172,8 +177,10 @@ async function ssrBuild(
...
 					entryFileNames(chunkInfo) {
-						if (chunkInfo.facadeModuleId === resolvedPagesVirtualModuleId) {
-							return opts.buildConfig.serverEntry;
+						if (chunkInfo.facadeModuleId?.startsWith(ASTRO_PAGE_RESOLVED_MODULE_ID)) {
+							return makeAstroPageEntryPointFileName(chunkInfo.facadeModuleId);
+						} else if (chunkInfo.facadeModuleId === SSR_VIRTUAL_MODULE_ID) {
+							return opts.settings.config.build.serverEntry;
 						} else if (chunkInfo.facadeModuleId === RESOLVED_MIDDLEWARE_MODULE_ID) {
 							return 'middleware.mjs';
 						} else if (chunkInfo.facadeModuleId === RESOLVED_RENDERERS_MODULE_ID) {
```
bluwy, line 186: With the `plugin-middleware` refactor, this condition wont pass anymore. It'll still pass if you have a `src/middleware.ts` because the else condition here catches it. But if I have `src/middleware/index.ts`, it won't output `middleware.mjs` anymore (`index.mjs` instead) (tested in `middleware` example). Maybe it's simpler to keep the `plugin-middleware` changes to handle this for now.

packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -1,54 +1,76 @@
...
 import type { StaticBuildOptions } from '../types';
 import { MIDDLEWARE_MODULE_ID } from './plugin-middleware.js';
 import { RENDERERS_MODULE_ID } from './plugin-renderers.js';
+import { extname } from 'node:path';
+
+export const ASTRO_PAGE_MODULE_ID = '@astro-page:';
+export const ASTRO_PAGE_RESOLVED_MODULE_ID = '\0@astro-page:';
+
+// This is an arbitrary string that we are going to replace the dot of the extension
+export const ASTRO_PAGE_EXTENSION_POST_PATTERN = '@_@';
```
bluwy, line 14: Nice pick 😄

packages/astro/src/core/build/plugins/plugin-ssr.ts
ematipico: I did, in the `README.md` I added

packages/astro/src/core/build/static-build.ts
ematipico: Good catch! Completely missed that

packages/astro/src/core/build/plugins/plugin-ssr.ts
bluwy: With the function refactor I guess we can skip this then. It would be nice still since (I assume) the README is for higher level stuff, and inline code comments add the guard rails when reading.

ematipico: Done!

matthewp (approved):

packages/astro/src/core/build/static-build.ts
bluwy: @ematipico I might've pointed the wrong thing, but with the new commit, this scenario still happens:

> But if I have `src/middleware/index.ts`, it won't output `middleware.mjs` anymore (`index.mjs` instead) (tested in `middleware` example).

In the `middleware` example with `src/middleware/index.ts`, the middleware is output as `dist/server/index.mjs`, which I don't think is intentional, and we want `dist/server/middlewares.mjs` instead?

I was suggesting keeping the previous `plugin-middleware` code, or maybe there's a property in `chunkInfo` that we can use to identify is a middleware?

bluwy: Maybe:

```js
if (chunkInfo.facadeModuleId?.includes('middleware') && chunkInfo.exports.includes('onRequest')) {
  return 'middleware.mjs'
}
```

A bit fragile though.

packages/astro/src/core/build/plugins/plugin-middleware.ts
```diff
@@ -8,7 +8,6 @@ import type { StaticBuildOptions } from '../types';
 export const MIDDLEWARE_MODULE_ID = '@astro-middleware';
 export const RESOLVED_MIDDLEWARE_MODULE_ID = '\0@astro-middleware';
```
bluwy, line 9: One small cleanup we could do is to remove this since it's unused.

bluwy (approved):
Looks great! Thanks for refactoring and documenting this section of the codebase.

PR #7197, opened 2023-05-24
Fix nested astro-island hydration race condition

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6301

- Setup `astro:hydrate` event listener and client directive init only after one macrotask when `MutationObserver` detected a change.
- Don't hydrate elements that are disconnected from the DOM.
- Setup prebuilt target to es2018 so we can write bleeding-edge code, while the prebuilt output is downleveled.

## Testing

The issue author helped confirmed this worked. The existing e2e tests should also pass.

## Docs

n/a. bug fix.


packages/astro/src/runtime/server/astro-island.ts
```diff
@@ -129,7 +134,7 @@ declare const Astro: {
 					window.dispatchEvent(new CustomEvent('astro:hydrate'));
 				};
 				attributeChangedCallback() {
-					if (this.hydrator) this.hydrate();
+					this.hydrate();
```
bluwy, line 137: NOTE: `this.hydrate` already has the `this.hydrator` guard, so this can be simplified.

PR #7198, opened 2023-05-24
Implement redirects in Cloudflare

matthewp:
## Changes

- Adds a new package `@astrojs/underscore-redirects` which is used to build the `_redirects` format, shared by Netlify and Cloudflare, from Astro route definitions.

## Testing

- Unit tests in new package
- Integration test with Cloudflare

## Docs

N/A, documented as part of release.

### DIFF
--- a/packages/underscore-redirects/package.json
+++ b/packages/underscore-redirects/package.json
@@ -1,0 +1,41 @@
+{
+  "name": "@astrojs/underscore-redirects",
+  "description": "Utilities to generate _redirects files in Astro projects",
+  "version": "0.1.0",
+  "type": "module",
+  "author": "withastro",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/withastro/astro.git",
+    "directory": "packages/underscore-redirects"
+  },
+  "bugs": "https://github.com/withastro/astro/issues",
+  "main": "./dist/index.js",
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "prepublish": "pnpm build",
+    "build": "astro-scripts build \"src/**/*.ts\" && tsc -p tsconfig.json",
+    "build:ci": "astro-scripts build \"src/**/*.ts\"",
+    "postbuild": "astro-scripts copy \"src/**/*.js\"",
+    "dev": "astro-scripts dev \"src/**/*.ts\"",
+    "test": "mocha --exit --timeout 20000"
+  },
+  "devDependencies": {
+    "astro": "workspace:*",
+    "astro-scripts": "workspace:*",
+    "@types/chai": "^4.3.1",
+    "@types/mocha": "^9.1.1",
+    "chai": "^4.3.6",
+    "mocha": "^9.2.2"
+  },
+  "keywords": [
+    "astro",
+    "astro-component"
+  ]
+}
--- a/packages/underscore-redirects/package.json
+++ b/packages/underscore-redirects/package.json
@@ -12,6 +12,7 @@
   },
   "bugs": "https://github.com/withastro/astro/issues",
   "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": "./dist/index.js"
   },
--- a/packages/underscore-redirects/src/astro.ts
+++ b/packages/underscore-redirects/src/astro.ts
@@ -11,18 +11,31 @@ function getRedirectStatus(route: RouteData): ValidRedirectStatus {
 	return 301;
 }
 
-export function createRedirectsFromAstroRoutes(
-	config: Pick<AstroConfig, 'output' | 'build'>,
-	routes: RouteData[],
-	dir: URL,
-	dynamicTargetValue: string
-) {
+interface CreateRedirectsFromAstroRoutesParams {
+	config: Pick<AstroConfig, 'output' | 'build'>;
+	routes: RouteData[];
+	dir: URL;
+	dynamicTarget?: string;
+}
+
+/**
+ * Takes a set of routes and creates a Redirects object from them.
+ */
+export function createRedirectsFromAstroRoutes({
+	config,
+	routes,
+	dir,
+	dynamicTarget = '',
+}: CreateRedirectsFromAstroRoutesParams) {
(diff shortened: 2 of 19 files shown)


packages/underscore-redirects/src/print.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function print(
+	definitions: RedirectDefinition[],
+	minInputLength: number,
+	minTargetLength: number
+) {
+	let _redirects = '';
+
+	// Loop over the definitions
+	definitions.forEach((defn, i) => {
```
ematipico, line 11: nit: `for..of` is usually [faster](https://perf.link/#eyJpZCI6ImRzZDZxMjVwcTl0IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGRhdGEgPSBbLi4uQXJyYXkoMTAwMDAwKS5rZXlzKCldIiwidGVzdHMiOlt7Im5hbWUiOiJGaW5kIGl0ZW0gMTAwIiwiY29kZSI6ImxldCBzdHJpbmcgPSBcIlwiO1xuXG5kYXRhLmZvckVhY2goeCA9PiBzdHJpbmcgKz0geCkiLCJydW5zIjpbMTA4LDk5LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTE3LDExNywxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCw5OSwxMDgsMTA4LDEwOCw5OSwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTE3LDEwOCwxMDgsOTksMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDExNywxMDgsMTE3LDEwOCwxMTcsMTE3LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMDgsMTA4LDEwOCwxMTcsMTA4LDEwOCwxMDgsMTA4LDExNywxMDgsMTA4LDEwOCwxMDgsMTE3LDExNywxMDgsMTE3LDk5LDEwOCwxMTcsMTE3LDEwOCwxMDgsMTA4LDExNywxMTcsMTE3LDExNywxMTcsMTA4LDEwOCwxMTcsMTE3LDExNywxMDgsMTA4LDEwOCwxMTcsMTE3LDEwOF0sIm9wcyI6MTA5fSx7Im5hbWUiOiJGaW5kIGl0ZW0gMjAwIiwiY29kZSI6ImxldCBzdHJpbmcgPSBcIlwiO1xuZm9yIChjb25zdCBzIG9mIGRhdGEpIHtcbiAgc3RyaW5nICs9IHM7XG59IiwicnVucyI6WzEzNSwxNjIsMTcxLDEyNiwyMTYsMTQ0LDEzNSwxODksMTgwLDEzNSwxNjIsMTgwLDE5OCwyMTYsMTQ0LDE4OSwxMDgsMjE2LDEzNSwyMTYsMTM1LDE4MCwyMDcsMjA3LDE1MywyMDcsMjE2LDE4MCwyMTYsMTUzLDE5OCwxODksMTk4LDE4OSwyMTYsMTI2LDEzNSwxMzUsMTYyLDE4MCwyMDcsMTM1LDEzNSwyMTYsMTM1LDIxNiwxOTgsMTgwLDEzNSwxODAsMTUzLDEzNSwxMzUsMjE2LDE5OCwxMzUsMTUzLDE2MiwyMTYsMTgwLDE4MCwxNDQsMTM1LDIxNiwyMjUsMTgwLDEzNSwxMzUsMTQ0LDIwNywxMzUsMTQ0LDE4OSwxODAsMjM0LDE5OCwyMTYsMjE2LDE0NCwxMzUsMTg5LDIxNiwxODAsMTcxLDE0NCwxNTMsMjE2LDIxNiwxODAsMTgwLDIwNywxODAsMjE2LDIxNiwxMjYsMTM1LDE2MiwyMTYsMTM1LDE1M10sIm9wcyI6MTc0fV0sInVwZGF0ZWQiOiIyMDIzLTA1LTI1VDA5OjA0OjQ0LjY3OVoifQ%3D%3D) than `.forEach`

packages/underscore-redirects/src/print.ts
```diff
@@ -0,0 +1,26 @@
+import type { RedirectDefinition } from './redirects';
+
+export function print(
```
ematipico, line 3: We could have a comment that shows how's the printed output. Easier to "debug" for us humans :)

packages/underscore-redirects/src/astro.ts
```diff
@@ -0,0 +1,123 @@
...
+const pathJoin = posix.join;
+
+function getRedirectStatus(route: RouteData): ValidRedirectStatus {
+	if(typeof route.redirect === 'object') {
+		return route.redirect.status;
+	}
+	return 301;
+}
+
+export function createRedirectsFromAstroRoutes(
```
ematipico, line 14: I find it difficult to understand the logic of this function:
- the PR doesn't explain much about the shared code
- there aren't any comments that help

If I would have to fix this code in the future, I would have hard time 😅

packages/underscore-redirects/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "scripts": {
+    "prepublish": "pnpm build",
+    "build": "astro-scripts build \"src/**/*.ts\" && tsc -p tsconfig.json",
+    "build:ci": "astro-scripts build \"src/**/*.ts\"",
+    "postbuild": "astro-scripts copy \"src/**/*.js\"",
+    "dev": "astro-scripts dev \"src/**/*.ts\"",
+    "test": "mocha --exit --timeout 20000"
+  },
+  "devDependencies": {
+    "astro": "workspace:*",
```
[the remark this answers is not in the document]
matthewp: No, because it's not needed at all at runtime. Given that this is used by integrations I don't know what have a peer would even do, because integrations do not directly depend on Astro either. I think peer applies when the user is using Astro, which isn't the case here.

ematipico (approved):
Thank you for addressing my comments!

Issue #5552, opened 2022-12-07
Error `fn is not a function` when Astro.glob in a TS file


goulvenclech:
### What version of `astro` are you using?

1.6.13

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

NPM

### What operating system are you using?

Tested on Linux, Mac & Windows

### Describe the Bug

Hi guys !

I tried to use `await Astro.glob` inside a TS file by passing Astro as an argument. Throw the error "FN is not a function", looks like a Vite-related problem. 

The same function works like a charm when in an Astro file. 

### Error log

```
fn is not a function

/home/projects/github-xlqiz4/globHandler/<@file:/file:/home/projects/github-xlqiz4/node_modules/astro/dist/runtime/server/astro-global.js:28:47

globHandler/<@file://file:///home/projects/github-xlqiz4/node_modules/astro/dist/runtime/server/astro-global.js:28:47
globHandler@file://file:///home/projects/github-xlqiz4/node_modules/astro/dist/runtime/server/astro-global.js:28:35
getStaticPathFromContentType@/src/utils.ts:4:32
getStaticPaths@/home/projects/github-xlqiz4/src/pages/blog/[article].astro:10:32
callGetStaticPaths@file://file:///home/projects/github-xlqiz4/node_modules/astro/dist/core/render/route-cache.js:38:27
getParamsAndProps@file://file:///home/projects/github-xlqiz4/node_modules/astro/dist/core/render/core.js:41:47
matchRoute@file://file:///home/projects/github-xlqiz4/node_modules/astro/dist/vite-plugin-astro-server/route.js:43:59
````

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xlqiz4?file=src/utils.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

Princesseuh:
Tried to debug this on my end and it seems like [importMetaGlobResult](https://github.com/withastro/astro/blob/4f7f20616ed2b63f94ebf43bc5fdc1be55062a94/packages/astro/src/runtime/server/astro-global.ts#L13) here is somehow an array of strings instead of being an array of imports. I'm not too familar with how Vite's `import.meta.glob` works, but I guess it comes from there?

bluwy:
I think this is expected as documented at https://docs.astro.build/en/reference/api-reference/#astroglob. Closing this for now.

PR #7204, opened 2023-05-25
Add error message if `Astro.glob` is called outside

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/5552

`Astro.glob` only works in .astro files. Error if otherwise.

I did a really quick fix for this one. Should we use the `AstroError` class here?

## Testing

Added a unit test.

## Docs

There's an existing docs for this at https://docs.astro.build/en/reference/api-reference/#astroglob

### DIFF
--- a/.changeset/moody-coats-develop.md
+++ b/.changeset/moody-coats-develop.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Add error message if `Astro.glob` is called outside of an Astro file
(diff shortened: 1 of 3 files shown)


Princesseuh:
> I did a really quick fix for this one. Should we use the AstroError class here?

I would say yes! It'll also allow us to point to any relevant docs etc.

ematipico (approved):
I also agree about using an `AstroError` if possible :)

matthewp (approved):

bluwy:
I'll merge this one first, and followup converting both the errors there as `AstroError`. Thanks for the reviews!

PR #7209, opened 2023-05-25
Markdoc - improve syntax highlighting support

bholmesdev:
## Changes

- Add the Prism syntax highlighter as an extension
  - Exported as `@astrojs/markdoc/prism`
  - Introduces `@astrojs/prism` as `@astrojs/markdoc` dependency
- Make shiki a **bundled** dependency to remove the manual `shiki` installation requirement
  - Change shiki extension export from `@astrojs/markdoc/config` to `@astrojs/markdoc/shiki`. This lets us mark the Shiki package as `ssr.external`, and removes the need for a manual `shiki` install
  - ⭐️ **Note:** this _will_ add `shiki` to your `node_modules` when deploying to SSR. Though with code splitting, shiki won't be imported at runtime unless `@astrojs/markdoc/shiki` is used. Verified manually building with the node and vercel adapters and checking bundled imports

## Testing

- Add Prism test
- Update Shiki import in tests

## Docs

- Add new syntax highlighting information to README
- Update original shiki changeset with new package location

### DIFF
--- a/packages/integrations/markdoc/src/extensions/prism.ts
+++ b/packages/integrations/markdoc/src/extensions/prism.ts
@@ -1,0 +1,22 @@
+// @ts-expect-error Cannot find module 'astro/runtime/server/index.js' or its corresponding type declarations.
+import { unescapeHTML } from 'astro/runtime/server/index.js';
+import { runHighlighterWithAstro } from '@astrojs/prism/dist/highlighter';
+import { Markdoc, type AstroMarkdocConfig } from '../config.js';
+
+export default async function prism(): Promise<AstroMarkdocConfig> {
+	return {
+		nodes: {
+			fence: {
+				attributes: Markdoc.nodes.fence.attributes!,
+				transform({ attributes: { language, content } }) {
+					const { html, classLanguage } = runHighlighterWithAstro(language, content);
+
+					// Use `unescapeHTML` to return `HTMLString` for Astro renderer to inline as HTML
+					return unescapeHTML(
+						`<pre class="${classLanguage}"><code class="${classLanguage}">${html}</code></pre>`
+					);
+				},
+			},
+		},
+	};
+}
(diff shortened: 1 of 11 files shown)


matthewp:
Does this mean Shiki is bundled when you're using Markdoc or for non-Markdoc projects as well? One thing to note is that some adapters like Cloudflare bundle everything into a single file, so conditionally loading doesn't help in those cases. But if this only applies to users of the Markdoc integration then that seems acceptable.

packages/integrations/markdoc/src/extensions/prism.ts
```diff
@@ -0,0 +1,22 @@
+// @ts-expect-error Cannot find module 'astro/runtime/server/index.js' or its corresponding type declarations.
+import { unescapeHTML } from 'astro/runtime/server/index.js';
```
ematipico, line 2: This type of comment broke the PR this morning. You should leave an empty newline after this import, so import sorting won't move the suppression comment. [Context](https://github.com/withastro/astro/pull/7206/files)

ematipico (approved):
A small comment that needs to be addressed before merge

packages/integrations/markdoc/src/extensions/prism.ts
bholmesdev: Got it, was unaware of that! Will fix 👍

bholmesdev:
@matthewp Just verified, and it only bundles shiki when `@astrojs/markdoc/shiki` is explicitly imported into your markdoc config (same goes for prism). When I remove the import, shiki is no longer bundled! Did a manual test with the cloudflare integration using new syntax highlighter, prism, and shiki. And correct, this is never bundled for projects that don't use Markdoc.

matthewp (approved):
With the explanation this approach sounds good to me.

PR #7215, opened 2023-05-25
fix(astro/core/app/node): support http2 requests

### LINKED ISSUE #7135: SSR Nodejs (middleware):  HTTP2 requests have an undefined hostname

2.5.0

Node

npm

Linux

Firefox

### Describe the Bug

Run as a middleware using [@fastify/middie](https://www.npmjs.com/package/@fastify/middie), [@astro/image](https://github.com/withastro/astro/tree/main/packages/integrations/image) integration fails to transform images (no matters which transformers is used) from HTTP/2 requests.

As a matter of fact, requests passed down to the Astro handler are properly routed to image endpoints, but local images can’t be loaded since hostname part of URLs are `undefined` (the `loadRemoteImage()` function receives `https://undefined/...` source URLs).

The issue comes from the `createRequestFromNodeRequest()` function:
https://github.com/withastro/astro/blob/main/packages/astro/src/core/app/node.ts#L17

HTTP2 Requests Headers from Fastify do not have the `host` property, but a `:authority` pseudo header (see also https://nodejs.org/docs/latest-v18.x/api/http2.html#note-on-authority-and-host).

How about using a fallback `req.headers.host || req.headers[':authority']` instead?

I’ll push a merge request later.

https://will.come.later/when/i/find/some/time

### Participation

bmenant:
## Changes

- Node adapter fallbacks to `:authority` http2 pseudo-header when `host` is nullish. 
  Fixes #7135 

## Testing

This actually _needs_ tests and I’d like to add some, but I do need some direction on where actually are the tests for the core/app/node pieces.
Plus it seems I can’t get a test environment up and running (loads of errors at build).

## Docs

No change to user’s behavior.


bluwy (approved):
Let's go ahead with this first then, we can add the tests for it later. Thanks for the PR!

Issue #6931, opened 2023-04-28
Style missing when using client:only directive


timseverien:
### What version of `astro` are you using?

2.3.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

## Summary

Given Svelte component with styles used in another Svelte component loaded with the `client:only` directive, styles of the inner component are missing on `astro build`.

## Example

Let’s say we have a component called `Flow` to create vertical rhythm:

```svelte
<div class="flow">
  <slot />
</div>

<style>
  .flow {
    display: flex;
    flex-direction: column;
    gap: 1em;
  }
</style>
```

Additionally, we have a component `Card` that uses `Flow`:

```svelte
<script>
  import Flow from "./Flow.svelte";
</script>

<Flow>
  <div>Header</div>
  <div>Content</div>
  <div>Footer</div>
</Flow>
```

Let’s say we want to use the `Card` component client-side only. In `index.astro`, we write:

```astro
<Card client:only />
```

Everything looks fine, but on production the `Flow` style is missing.

## Distilling the problem

- The problem doesn’t occur on `astro dev`
- The problem does occur when running `astro build && astro preview`
- When the `client:only` directive is removed, the `Flow` styles are applied again
- When the `client:only` directive is swapped for `client:loaded` or any of the other `client:*` directives, styles are applied again
- When adding a `<Flow>` component to the `index.astro`, the styles are applied again, regardless whether we keep the `client:only` directive on `Card` or not

It appears `astro build` performs an action that `astro dev` doesn’t, which may discard styles or cause classname mismatches.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/bug-astro-svelte-styles?file=src/pages/error.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

vampaz:
Can add that I've the same with Vue. I think it is related with this https://github.com/withastro/astro/commit/7653cf9e9fedc6edc6038603248351e276191c3a

amxmln:
I can confirm this is an issue with Vue as well. Maybe it has to do with scoped styles?

Another thing I noticed, that may be of relevance is that the ordering of styles is different in dev and build. Importing `normalize.css` causes the styles from normalize to override my styles during dev, because it gets added *after* my styles, but everything is fine after build.

This started roughly around the same time the component styles disappeared when using `client:only`.

josephmcg:
[...]

Issue #7124, opened 2023-05-18
#6582 introduced production build error


PxlBuzzard:
### What version of `astro` are you using?

`2.2.3`, and `2.5.0` still has the issue

### Are you using an SSR adapter? If so, which one?

Node, middleware

### What package manager are you using?

yarn

### What operating system are you using?

Mac

### Describe the Bug

From `2.2.3` onwards, there is an error during the `build` client step:

```
building client
vite v4.2.1 building for production...

"getDefaultExportFromCjs" is not exported by "commonjsHelpers.js", imported by "node_modules/react/index.js".
^ same error across many node_modules

RollupError: "getAugmentedNamespace" is not exported by "commonjsHelpers.js", imported by "repo/node_modules/@apollo/client/index.js?commonjs-proxy".
```

I don't believe `@apollo/client` is the culprit, it just happens to be the first package in `ssr.noExternal` list. I am going to dive into the exact code changes from [2.2.2 to 2.2.3](https://github.com/withastro/astro/compare/astro%402.2.2...astro%402.2.3) and see if I can find the issue.

Related dependency versions:
- `vite: 4.3.8` (latest, `4.2.1` saved in astro `2.2.3`)
- `rollup: 3.22.0` (latest, `3.18.0` saved in vite `4.2.1`)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-zj4xwa, but I don't have a working repro yet.

### Participation

- [X] I am willing to submit a pull request for this issue.

bluwy:
From the 2.2.3 changelog, https://github.com/withastro/astro/pull/6582 might be the likely culprit. As you mentioned there isn't a working repro, it would be a bit hard to find a fix for it, but perhaps that PR would be a clue if you'd like to debug it.

PxlBuzzard:
Thanks for the callout! When I revert the changes in `plugin-css.ts` and link it to my local project, it successfully compiles on Astro 2.5.0! I should be able to work towards a bug fix PR knowing that.

PxlBuzzard:
[No progress update] I've been trying on and off the last couple days to create a repro for this in the Astro test files, but I still can't get it to repro. I did try enabling `cssCodeSplit: true` in my failing repo and that compiles because it bypasses the failing logic. I'm not yet knowledgable enough about Vite's plugin system to how I might even make a fix.

PxlBuzzard:
#7218 fixes this issue, thank you so much for your work @bluwy. 

[...]

PR #6582, opened 2023-03-17
Fix CSS chunking between multiple framework components

bluwy:
## Changes

Fix #6524
Fix #6395

When we use `manualChunks` to give a name for CSS chunks, we based on all the parent top-level `.astro` page related to that CSS chunk to create a name and hash. 

However, for CSS chunks during the client build, where the build inputs are the framework components, it's not possible to do so. This PR skips the custom chunk handling for client builds.

But the above change indirectly regresses https://github.com/withastro/astro/pull/4566, so I fixed it by changing our CSS dedupe handling in a `transform()` hook instead of `renderChunk()`. 

## Testing

Added a test based on both issue repros, also updated some tests due to new chunking behaviour.

## Docs

n/a. bug fix.

PR #7218, opened 2023-05-26
Fix CSS deduping and missing chunks

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/6931
May also fix this issue: https://github.com/withastro/astro/issues/7124 (cc @PxlBuzzard appreciate if you can test this after it's released)

> Preface: "CSS module" here means the CSS rollup module, not [CSS modules](https://github.com/css-modules/css-modules)

The previous CSS deduping flow was incorrect, as it assumed a CSS module found in a SSR build applies to all pages, which isn't the case most of the time. This regression was introduced in #6582

I've fixed this by implementing a more correct CSS deduping strategy, that is to make sure SSR and client builds' generated CSS chunk ids are always derived the same way. That way the chunk name and content are always the same when written to fs.

I've put more in-code comments on how it works and why.

## Testing

Added test from #6931. Made sure all other tests passed.

I'm also trying a new way to create the test, moving the new fixture code in `0-css/src/pages/client-only-and-ssr/` only. This way each new test we add are only new folders in `pages/`

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/eighty-gifts-cheer.md
+++ b/.changeset/eighty-gifts-cheer.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix CSS deduping and missing chunks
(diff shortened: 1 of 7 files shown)


PxlBuzzard:
@bluwy This PR fixed it! Thank you for taking the time to find a fix for this, I appreciate it.

bluwy:
Awesome! Thanks for checking it.

PR #7220, opened 2023-05-26
feat: SSR `split` mode

ematipico:
## Changes

This PR adds a new SSR mode, called `split`.

Serverless mode creates an entry point for each page, and each entry point is a self-contained application meant to render one single page.

Here's a list of changes:
- added a new `split` boolean to the `build` object; by default, the value is `false`, which means that Astro will keep the current behaviour;
- `plugin-ssr.ts` will emit two plugins, one to emit SSR server and one for SSR split, and they can't coexist. One is used, and the other won't. The reason why the two plugins live in the same file is that they share the majority of the code;
- in `plugin-ssr.ts` the code that generates the deserialisation of the manifest was moved into a function, shared between the new plugins;
- the shape of the manifest now changes a bit based on the `split` value. `pageMap` is kept for SSR server, while SSR split has a simple `pageModule` function. Because of that, I created two types: `SSRServerManifest` and `SSRServerlessManifest`;
- I created a new option called `serverlessEntryPrefix`; by default it's `entry`

Here's an example of the emitted output

```
├── chunks
│   ├── @my-ssr.6147f6ca.mjs
│   ├── astro.60173704.mjs
│   └── pages
│       ├── _slug_.astro.bbdede9d.mjs
│       ├── about.astro.60a22a11.mjs
│       └── index.astro.23e80271.mjs
├── pages
│   ├── blog
│   │   ├── entry._slug_.astro.mjs
│   │   └── entry.about.astro.mjs
│   └── entry.index.astro.mjs
└── renderers.mjs
```

## Testing

Added some test case

## Docs

If you're OK with the APIs, then I will start drafting a PR in the docs repo, because now the payload of `astro:build:ssr` has an additional property


packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -73,12 +69,224 @@ function vitePluginSSR(
...
+					return;
+				}
+				if (options.settings.config.build?.mode === 'server') {
+					return;
+				}
+
+				if (internals.ssrServerlessEntryChunks.size === 0) {
+					throw new Error(`Did not generate an entry chunk for SSR serverless`);
+				}
+				// TODO: check this commented code
```
ematipico, line 256: This should work now, I will remove the `TODO` once there are more tests

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -73,12 +69,224 @@ function vitePluginSSR(
...
+				const {
+					settings: { config },
+					allPages,
+				} = options;
+				const imports: string[] = [];
+				const contents: string[] = [];
+				const exports: string[] = [];
+
+				const path = getPathFromVirtualModulePageName(RESOLVED_SERVERLESS_MODULE_ID, id);
+				const virtualModuleName = getVirtualModulePageNameFromPath(ASTRO_PAGE_MODULE_ID, path);
```
ematipico, line 181: Here we juggle from `\0@astro-serverless-page:src/pages/index@_@astro` to `@astro-page:src/pages/index@_@astro`

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -73,12 +69,224 @@ function vitePluginSSR(
...
+): AstroBuildPlugin {
+	const ssr =
+		options.settings.config.output === 'server' || isHybridOutput(options.settings.config);
+	return {
+		build: 'ssr',
+		hooks: {
+			'build:before': () => {
+				let vitePlugin =
+					// config.build object is optional, so we check NOT EQUAL against "serverless" instead
+					ssr && options.settings.config.build?.mode !== 'serverless'
```
ematipico, line 112: This is the only change to this function

packages/astro/src/@types/astro.ts
```diff
@@ -733,6 +733,27 @@ export interface AstroUserConfig {
...
 		serverEntry?: string;
+
+		/**
+		 * @docs
+		 * @name build.mode
+		 * @type {string}
+		 * @default `'server'`
+		 * @description
+		 * Defines how the SSR should be bundled. SSR code for "server"
+		 * will be built in one single file.
```
matthewp, line 744: Should also explain what happens with 'serverless'.

packages/astro/src/@types/astro.ts
```diff
@@ -733,6 +733,27 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		serverEntry?: string;
+
+		/**
+		 * @docs
+		 * @name build.mode
+		 * @type {string}
```
matthewp, line 740: I think you can do `{'server' | 'serverless'}` here.

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -73,12 +69,224 @@ function vitePluginSSR(
...
+	if (config.experimental?.middleware === true) {
+		imports.push(`import * as _middleware from "${MIDDLEWARE_MODULE_ID}";`);
+		middleware = 'middleware: _middleware';
+	}
+	let pageMap;
+	if (config.build.mode === 'serverless') {
+		pageMap = 'pageModule';
+	} else {
+		pageMap = 'pageMap';
+	}
```
ematipico, line 282: The diffing doesn't help, but this is the only dynamic part of the code generated in SSR. `pageMap` for server and `pageModule` for serverless

ematipico:
I rebase the branch. The PR should be ready for review

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,34 @@
...
+    }),
+    build: {
+        ssrMode: "serverless",
+        serverlessEntryPrefix: "main"
+    }
+})
+```
+
+- `ssrMode` accepts two values, `"server"` or `"serverless"`. Default value, `"server"`. 
+- `serverlessEntryPrefix` allows to change the prefix of a serverless
```
bluwy, line 34: I wonder if `serverless` is the right term here. The previous build output can and are also used on serverless platforms. SvelteKit's Vercel adapter calls it ["split"](https://kit.svelte.dev/docs/adapter-vercel#deployment-configuration), since we're doing on the build-level, maybe we can call it `build.pageCodeSplit`? Vite has a similar option called `vite.build.cssCodeSplit`.

I don't really understand the `serverlessEntryPrefix` part yet, but maybe I will when I finish reviewing this PR!

packages/astro/src/@types/astro.ts
```diff
@@ -822,6 +822,61 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 *   build: {
+		 *     serverlessEntryPrefix: 'main'
+		 *   }
+		 * }
+		 * ```
+		 */
+		serverlessEntryPrefix?: string;
```
matthewp, line 843: After talking offline we thought it might be premature to offer this option. Users who want to customize the filename probably want full customization. So we decided to omit this for now and then to add it later when/if people ask for it.

.changeset/wet-readers-join.md
matthewp: are you proposing `build.pageCodeSplit` be a boolean or what?

bluwy: Yeah, it would be a boolean which defaults to false.

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -71,12 +73,216 @@ function vitePluginSSR(
...
+				}
+				// TODO: check this commented code
+				// Mutate the filename
+				// internals.ssrEntryChunk.fileName = options.settings.config.build.serverEntry;
+
+				for (const [moduleName, chunk] of internals.ssrServerlessEntryChunks) {
+					const code = await injectManifest(options, internals, chunk);
+					mutate(chunk, 'server', code);
+				}
+			},
```
bluwy, line 265: Does this mean we inject a manifest for each page chunk? It would be nice if they share a chunk that solely contains the manifest, if that's possible 🤔 

The difference between `SSRServerManifest` and `SSRServerlessManifest` seems to be the module instance value `pageModule` vs `pageMap`. If we can share a chunk that uses `pageMap` only, then each page chunk could manually load their own page module from that map?

.changeset/wet-readers-join.md
ematipico: I probably misunderstand the meaning, because in my head it seems we split the code of a page in multiple files. 

But it makes sense to rename the option to something that's more familiar to the majority of users

packages/astro/src/core/build/plugins/plugin-ssr.ts
ematipico: > I mean the map should contain all the pages, but the chunk would access only one page from the map. There shouldn't be a huge implication since each module is a dynamic import? (Or maybe I'm missing the idea of splitting here)

When classic SSR, the `entry.mjs` file will have something like this (pseudo-code, simplified):

```js
const page0 = () => import("chunks/page0.0303030.mjs");
const page1 = () => import("chunks/page1.04ofio3030.mjs");

const pageMap = new Map([
	["src/pages/index.astro", page0],
	["src/pages/blog.astro", page1],
])

const manifest = {
	pageMap
}
```

Instead, with SSR split, each file will look like this (pseudo-code, simplified):

```js
// file dist/pages/entry.index.astro.mjs
const pageModule = () => import("chunks/page0.0303030.mjs").then(n => n.i); 

const manifest = {
	pageModule
}
```

```js
// file dist/pages/entry.blog.astro.mjs
const pageModule = () => import("chunks/page1.04ofio3030.mjs").then(n => n.i); 

const manifest = {
	pageModule
}
```

That's why the `pageMap` and `pageModule`. What do you think? Should we keep `pageMap` for both?

packages/astro/src/core/build/internal.ts
```diff
@@ -114,6 +116,8 @@ export function createBuildInternals(): BuildInternals {
 		discoveredScripts: new Set(),
 		staticFiles: new Set(),
 		componentMetadata: new Map(),
+		ssrServerlessEntryChunks: new Map(),
```
bluwy, line 119: nit: I think it would be nice to remove the term "serverless" around the codebase since it's not specifically for serverless environments.

I'm not sure what others think about this though, but it's a bit confusing for me if we keep the term here.

packages/astro/src/core/app/index.ts
```diff
@@ -175,14 +188,21 @@ export class App {
...
+						`Unexpectedly unable to find a component instance for route ${route.route}`
+					);
+				}
+				const pageModule = await importComponentInstance();
+				return pageModule;
+			} else if (this.#manifest.pageModule) {
+				const importComponentInstance = this.#manifest.pageModule;
+				return importComponentInstance;
+			} else {
+				throw new AstroError(AstroErrorData.FailedToFindPageMapSSR);
```
bluwy, line 204: I'm not sure if `AstroError` can be used in the `App`. IIUC this code is bundled into the user's server code. Checking it locally, I can see the entire AstroError handling is included so that might not be what we want as it increases the bundle size.

I think we can use a plain ol error here for now.

packages/astro/src/core/app/index.ts
```diff
@@ -52,7 +53,19 @@ export class App {
...
+		if (this.#manifest.pageMap) {
+			const pageModule = await this.#manifest.pageMap.get(routeData.component)!();
+			return await pageModule.page();
+		} else if (this.#manifest.pageModule) {
+			const pageModule = await this.#manifest.pageModule;
+			return await pageModule.page();
+		} else {
+			throw new AstroError(AstroErrorData.FailedToFindPageMapSSR);
+		}
+	}
```
bluwy, line 66: This function doesn't seem to be used, but if you do plan to keep it, private functions are often written at the bottom of the class, instead of here (before the constructor).

packages/astro/src/@types/astro.ts
```diff
@@ -822,6 +822,39 @@ export interface AstroUserConfig {
...
 		inlineStylesheets?: 'always' | 'auto' | 'never';
+
+		/**
+		 * @docs
+		 * @name build.ssrMode
+		 * @type {boolean}
+		 * @default {false}
+		 * @description
+		 * Defines how the SSR code should be bundled. SSR code for "server"
+		 * will be built in one single file.
```
bluwy, line 833: The docs here needs to be updated.

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,33 @@
+---
+'astro': minor
+---
+
+Shipped a new SSR mode, called `serverless`. 
+When enabled, Astro will emit a file for each page, which will render one single page.
+
+These files will be emitted inside `dist/pages`, and they will look like this:
```
bluwy, line 8: The changeset docs here also needs to be updated.

packages/astro/src/@types/astro.ts
sarah11918: Yes! I'm reviewing the docs PR and it mentions `build.split` so I'm assuming that's the correct name for this?

matthewp (approved):

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,30 @@
+---
+'astro': minor
+---
+
+Shipped a new SSR mode, called `split`. 
+When enabled, Astro will emit a file for each page, which will render one single page.
```
sarah11918, line 6:
```suggestion
When enabled, Astro will "split" the single `entry.mjs` file and instead emit a separate file to render each individual page during the build process.
```
I'm assuming this is where the name "split" came from?  :smile: It would be good to have something to connect this name to what it does.

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,30 @@
...
+
+These files will be emitted inside `dist/pages`, and they will look like this:
```
sarah11918, line 8:
```suggestion
These files will be emitted inside `dist/pages`, mirroring your page files in `src/pages/`, for example:
```
This is the wording we went with in docs, and is more descriptive than "they will look like this" so I prefer something like this!

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+├── pages
+│   ├── blog
+│   │   ├── entry._slug_.astro.mjs
+│   │   └── entry.about.astro.mjs
+│   └── entry.index.astro.mjs
+```
+
+To enable and customise this mode, new options are now available:
```
sarah11918, line 18:
```suggestion
To enable, set `build.split: true` in your Astro config:
```
I don't think this is customizable? Only enable-able? I would mention that this is a build option, one way or another here.

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,30 @@
...
+
+```js
```
sarah11918, line 20:
```suggestion
```js
// src/astro.config.mjs
```
Let's add a file title, whatever the convention is here!

packages/astro/src/@types/astro.ts
```diff
@@ -822,6 +822,39 @@ export interface AstroUserConfig {
...
 		inlineStylesheets?: 'always' | 'auto' | 'never';
+
+		/**
+		 * @docs
+		 * @name build.split
+		 * @type {boolean}
+		 * @default {false}
+		 * @version 2.7.0
+		 * @description
+		 * Defines how the SSR code should be bundled.
```
sarah11918, line 833:
```suggestion
		 * Defines how the SSR code should be bundled when built.
```
Does this make sense to add? Is it too redundant? Your call!

packages/astro/src/@types/astro.ts
```diff
@@ -822,6 +822,39 @@ export interface AstroUserConfig {
...
+		 *
+		 * When `split` is `true`, Astro will emit a file for each page.
+		 * Each file emitted will render only one page. The pages will be emitted
+		 * inside a `dist/pages/` directory, and the emitted files will keep the same file paths
+		 * of the `src/pages` directory.
```
sarah11918, line 838:
```suggestion
		 * inside a `dist/pages/` directory, mirroring the same file paths as the `src/pages/` directory.
```
I think this is tighter!

packages/astro/src/@types/astro.ts
```diff
@@ -822,6 +822,39 @@ export interface AstroUserConfig {
...
+		 *
+		 * Inside the `dist/` directory, the pages will look like this:
+		 * ```plaintext
+		 * ├── pages
+		 * │   ├── blog
+		 * │   │   ├── entry._slug_.astro.mjs
+		 * │   │   └── entry.about.astro.mjs
+		 * │   └── entry.index.astro.mjs
+		 * ```
```
sarah11918, line 847: I'm going to ask @delucis to weigh in on this one:

Chris: will this plaintext file tree be picked up and rendered with our `<Filetree />` component in docs? If not, this will look out of place by comparison.

I might consider just not showing it at all, since we do show this nice component file tree in docs itself on the SSR page. Maybe it's good enough for reference to just say that `dist/pages/` mirrors `src/pages/`?

sarah11918 (commented):
I'm not used to editing so much in a changeset. :smile:

This looks good @ematipico, just a few comments for your consideration where maybe we can make the text tighter!

bluwy (approved):
Code looks good to me! There's a few other places that still uses the word "serverless" that could be renamed "split-mode" perhaps, but I don't think their usage currently is too confusing, so happy as-is too.

Also have a small suggestion with test below.

.changeset/wet-readers-join.md
ematipico: Mirroring is not technically correct because the name of the files will look different. The only thing that technically we mirror is the **directory structure**. The name of the file will be different, e.g.: `blog.astro` -> `entry.blog.astro.mjs`

packages/astro/src/@types/astro.ts
ematipico: Makes sense!

ematipico:
Heads up @natemoo-re 

In [`f7e8c4d` (#7220)](https://github.com/withastro/astro/pull/7220/commits/f7e8c4d51313971a2b95f0ab3a66e5880db60d0d) I changed what we pass to compute the correct entry point. Before, we were passing `allPages`, but this is not ideal because that's an internal data structure that is not exposed to integrations. Usually, we pass `RouteData[]` to integration hooks, so I changed it. This allowed me to make our test adapter testable.

.changeset/wet-readers-join.md
sarah11918: OK, what about something like "mirroring the directory structure of your page files in `src/pages/`"... ?

packages/astro/src/@types/astro.ts
sarah11918: Same suggestion as above, I think you could say "mirroing the directory structure" if you wanted to!

packages/astro/src/@types/astro.ts
sarah11918: Ok, so just to clarify, this will NOT show up in the `@description`? If not, then do whatever you want as it won't get published in docs! :smile:

.changeset/wet-readers-join.md
```diff
@@ -0,0 +1,31 @@
+---
+'astro': minor
+---
+
+Shipped a new SSR build configuration mode: `split`. 
+When enabled, Astro will "split" the single `entry.mjs` file and instead emit a separate file to render each individual page during the build process.
+
+These files will be emitted inside `dist/pages`, and they will look like this:
```
sarah11918, line 8:
```suggestion
These files will be emitted inside `dist/pages`, mirroring the directory structure of your page files in `src/pages/`, for example:
```
Updated my suggestion to be more precise about the mirroring. Would this work?

packages/astro/src/@types/astro.ts
sarah11918: On closer inspection, I think this one is fine as is! Suggested a change only above!

sarah11918 (approved):
OK, updated my one suggestion in the changeset to be now more explanatory AND accurate for your consideration! :wink: But, I withdraw my suggested change in the docs itself. That one is fine as is! I've resolved and it's good to go!

packages/astro/src/@types/astro.ts
ematipico: Sorry that I was a wish of mine :D 
**Yes, it will be shown in the `@description`**

ematipico: @sarah11918 I removed the `plaintext` block from the documentation

PR #7227, opened 2023-05-28
#7226 - fixes NodeJS adapter for multiple set-cookie headers (and other header issues)

### LINKED ISSUE #7226: Node Adapter writes server.writeHead(..) headers incorrectly

2.5.5

node

pnpm

macOS, linux

Chrome

### Describe the Bug

In the NodeJS Adapter middleware https://github.com/withastro/astro/blob/main/packages/integrations/node/src/nodeMiddleware.ts#L52

There is this line to start writing the actual server response

```typescript
res.writeHead(status, Object.fromEntries(headers.entries()));
```

The line `Object.fromEntries(headers.entries())` is too naive and not accounting for the fact that `Headers` is a special purpose multi-value map (i.e. can contain > 1 value per key), it cannot be converted to a simple object in this manner.

For example:

```typescript
const h = new Headers();
h.append("set-cookie", "cookie1=value");
h.append("set-cookie", "cookie2=value");

// results in { "set-cookie": "cookie1=value" }
const naiveObject = Object.fromEntries(headers.entries());
```

Where it needs to be
```json
{ "set-cookie": ["cookie1=value", "cookie2=value"] }
```

In reading the node spec for `response.writeHead` https://nodejs.org/docs/latest-v18.x/api/http.html#responsewriteheadstatuscode-statusmessage-headers interestingly it does not explicitly state that the object form of supplying headers can accept an array of values (but the docs for `response.setHeader` do say this, so I would assume the behavior is the same)

So this needs to be smarter about building a multi-value object of header key/value pairs

There has been a bunch of recent issues opened up *specifically* around set-cookie behavior not working as expected (and indeed that's where I got tripped up on this), and I have a suspicion for most of them this is the root cause, and it's a general header problem, not specifically about cookies.

The linked stack blitz doesn't really work because this is only a problem once you build and run a prod build (because it's the nodejs adapter middleware that's the issue, so the dev server doesn't exhibit this problem) 

https://stackblitz.com/edit/github-chr9rk?file=src%2Fpages%2Fapi.ts

### Participation

alex-sherwin:
Fixes #7226

# What it does

Utilizes the new standard WebAPI Fetch [Headers.getSetCookie\()](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) function to safely handle multiple set-cookie headers when converting from a WebAPI Response to a NodeJS ServerResponse

Modifies the existing nodeMiddleware logic which used to first set AstroCookies on `ServerResponse.setHeader(...)` and then called `ServerResponse.writeHead(status, Response.headers)`, which meant any that if the WebAPI Response had any `set-cookie` headers on it, they would replace anything from `AstroCookies` (they were not merged, so any `AstroCookies` would get lost during response writing, and, multiple `set-cookie` headers from the Response.headers were written incorrectly as a CSV).

The new logic delegates appending `AstroCookies` values onto the WebAPI Response Headers object, so that a new single unified function safely converts from the WebAPI Response Headers into a NodeJS compatible `OutgoingHttpHeaders` object utilizing the new standard `Headers.getSetCookie()` function provided by the existing **undici** WebAPI polyfills.

Plus extensive test coverage.

# Notes

The Fetch WebAPI standard has been updated with a function `Headers.getSetCookie()` (see https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie)

The undici library that **@astrojs/webapi** utilizes for WebAPI polyfills on NodeJS added support for `Headers.getSetCookie()` in 5.19.0 (see https://github.com/nodejs/undici/releases/tag/v5.19.0)

Since Astro is already using undici@^5.22.0, this means Astro code already had access to this method without having to change any dependencies.

I think that this is probably the best most standards compliant way to fix this problem, and other [unrelated] existing pieces of code that are using copy/pasted set-cookie heading parsing logic (such as https://github.com/withastro/astro/blob/main/packages/integrations/netlify/src/netlify-functions.ts#LL140C26-L140C26) might want to consider changing to using `Headers.getSetCookie()` where appropriate since going forward it's a standards compliant way of dealing with the set-cookie special case in the WebAPI Headers type.


alex-sherwin:
I think these failures (around a Markdoc image asset test) are also present on main, I'm getting the same error when running tests locally on the main branch

natemoo-re (approved):
This PR is incredibly thorough with great tests, thanks for contributing!

alex-sherwin:
Thanks for merging!

I found the astro e2e/integration test harness stuff is really nicely done and very easy utilize for real-world integration/adapter test coverage

PR #7228, opened 2023-05-28
Update page.ts, removed newline on tag

proyb6:
## Changes

Removed '\n' on doctype tag to improve HTML minification.


packages/astro/src/runtime/server/render/page.ts
```diff
@@ -39,7 +39,7 @@ async function iterableToHTMLBytes(
 			if (i === 0) {
 				i++;
 				if (!/<!doctype html/i.test(String(chunk))) {
-					parts.append('<!DOCTYPE html>\n', result);
+					parts.append('<!DOCTYPE html>', result);
```
ematipico, line 42: This has the side effect of removing the newline even when **we are not minifying the HTML**. We read the configuration ad remove the newline only wen `compress: true`.

And we need to update the tests

ematipico (changes requested):

bluwy:
Superseded by https://github.com/withastro/astro/pull/7242

Issue #7051, opened 2023-05-10
Vitest doesn't work with Content Collection APIs


schietegal:
### What version of `astro` are you using?

2.4.1

### Are you using an SSR adapter? If so, which one?

-

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

-

### Describe the Bug

Writing unit tests with vitest and trying to access the [Content Collection APIs](https://docs.astro.build/en/reference/api-reference/#content-collections) will result in an error.

`bholmesdev` already pointed out that [`getViteConfig`](https://github.com/withastro/astro/blob/latest/examples/with-vitest/vitest.config.ts) might help, but it doesn't.

Just run `npm run test`on the reproduction and look at the console.

Error:
```bash
Error: Content config failed to load.
  ❯ _0x46a642@https:/github3bgwtf-dgzh.w-corp.staticblitz.com/blitz.9c984a48.js:44:536697
  ❯ AstroError@file:/file:/home/projects/github-3bgwtf/node_modules/astro/dist/core/errors/errors.js:26:5
  ❯ getContentConfigFromGlobal@file:/file:/home/projects/github-3bgwtf/node_modules/astro/dist/content/vite-plugin-content-imports.js:208:11
  ❯ setContentEntryModuleCache@file:/file:/home/projects/github-3bgwtf/node_modules/astro/dist/content/vite-plugin-content-imports.js:144:33
  ❯ transform@file:/file:/home/projects/github-3bgwtf/node_modules/astro/dist/content/vite-plugin-content-imports.js:65:73
  ❯ transform@file:/file:/home/projects/github-3bgwtf/node_modules/vite/dist/node/chunks/dep-934dbc7c.js:42935:44
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-3bgwtf

### Participation

- [ ] I am willing to submit a pull request for this issue.

mattwagl:
Hey Astro-Team 👋😀! Thanks for building this wonderful piece of software.

I am also experiencing this problem. While I was in the process of creating a bug report myself, this issue was opened. So I thought I'd add my reproducible example here.

**Are you using an SSR adapter? If so, which one?**
None

**What package manager are you using?**
npm

**What operating system are you using?**
Mac

**What browser are you using?**
Chrome

**Which Integrations are you using?**

```json
  "@astrojs/mdx": "0.19.1",
  "@astrojs/react": "2.1.3", // not included in the sandbox
[...]

PR #4154, opened 2022-08-04
Add `getViteConfig` helper

natemoo-re:
## Changes

- Fixes https://github.com/withastro/astro/issues/3856
- Adds a new export to `astro/config` entrypoint that exposes `getViteConfig`
- Adds a new `with-vitest` example
- Testing guide to come later

## Testing

Tested manually

## Docs

Future task

PR #7233, opened 2023-05-29
Fix Vitest with content collections

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7051

Make sure we also call `attachContentServerListeners` for `getViteConfig`. In `astro dev`, it's called in the CLI:

https://github.com/withastro/astro/blob/ae99753d73358ad685d7f6e5a8d52f7ae0164926/packages/astro/src/core/dev/dev.ts#L98

But it's not called for `getViteConfig`, hence there were issues when using content collections with it.

## Testing

Manually tested. It's a bit hard to test this without some integration test setup, the [initial implementation
](https://github.com/withastro/astro/pull/4154) was also manually tested.

## Docs

n/a. bug fix.


matthewp (approved):

packages/astro/src/config/vite-plugin-content-listen.ts
```diff
@@ -0,0 +1,41 @@
...
+		},
+		async buildStart() {
+			await attachContentServerListeners({
+				fs: fs,
+				settings,
+				logging,
+				viteServer: server,
+			});
+		},
+	};
```
bholmesdev, line 40: A bit confused when this runs. Is this potentially triggering the listeners _twice_ (once from the `dev.ts` file and once from here)? Or is this exclusively a patch for vitest?

bluwy: The `dev.ts` file would only run when we call `astro dev`, so this call here is exclusively for vitest, since a vitest/vite config setup doesn't call the Astro CLI.

bholmesdev: Got it. Another reason to remove the server requirement for loading content configs! Can ideally use a vite plugin for all cases instead of calling `attach-listeners` separately 👍

bholmesdev (approved):
collexcellent

packages/astro/src/config/vite-plugin-content-listen.ts
bluwy: I did tried a Vite plugin at first, but the timing when it logs "content types generated" for the CLI is off (logs before the CLI), and it's pretty hard to adjust that timing that I went with calling it in two places for now 😬

PR #7235, opened 2023-05-29
fix: prioritize dynamic prerendered routes over dynamic server routes 

### LINKED ISSUE #7200: Inconsistent route order priority in SSR/Hybrid output mode

2.5.5

Node

npm

Linux

Firefox

### Describe the Bug

[Route Priority Order](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) favours static pages over SSR, but all routes alphabetically in development. This causes a mismatch between what the user sees in development vs production.

As an example, we have two rest parameter routes. `[...page].astro` (SSR) and `[...slug].astro` (static).

- In development `[...page].astro` takes priority because of the alphabetical ordering. 
- In production (at least in the case of the Node adapter) generated static pages of `[...slug].astro` take priority, leaving `[...page.astro]` any non-static routes.

https://stackblitz.com/edit/github-dxo1jg?file=src%2Fpages%2Fblog%2F[...page].astro&on=stackblitz

### Participation

MoustaphaDev:
## Changes

Fix #7200

Currently the ordering conflicts of dynamic static and server routes are resolved alphabetically. However, dynamic static routes should always be prioritized in that case.

## Testing

- Added unit tests for the `matchAllRoutes` function to verify that the ordering of matched routes is correct
- Added unit tests to verify that individual routes are matched correctly

## Docs

The current behavior didn't change, but I thought it would be good to add some missing info to the docs 
https://github.com/withastro/docs/pull/3374


packages/astro/src/prerender/routing.ts
```diff
@@ -0,0 +1,67 @@
...
+	matches: RouteData[];
+	settings: AstroSettings;
+};
+async function preloadAndSetPrerenderStatus({
+	env,
+	matches,
+	settings,
+}: PreloadAndSetPrerenderStatusParams) {
+	const preloaded = await Promise.all(
+		matches.map(async (route) => {
```
bluwy, line 35: Putting a note: We had issues in the past that using `Promise.all` on promises that access the filesystem could slow things down and even error out. But IIUC `matches` here should usually be a low number, so this should be fine.

bluwy (approved):
Looks great!

packages/astro/src/prerender/routing.ts
MoustaphaDev: Yeah `matches` should generally be low, I'll keep an eye out if this every becomes problematic 👍️

PR #7242, opened 2023-05-30
remove the white space after the doctype according to the property co…

### LINKED ISSUE #7162: HTML minification with Doctype tag

2.5.1

Node

npm

macOS

Firefox

### Describe the Bug

After ran thru Astro HTML minification in 2.5.x release, I found curiosity why isn't docttype tag being minified?

```
<!DOCTYPE html>
<html ...>
```

NA

### Participation

JerryWu1234:
…mpressHTML

fix #7162

## Changes
remove the white space after the doctype according to the property compressHTML


.changeset/tasty-stingrays-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
matthewp, line 2: Let's make this be `patch` as well.

Princesseuh:
@matthewp Are there any streaming concerns with this? I feel like surely we made it so there's a newline after the doctype for a reason.

wulinsheng123:
> @matthewp Are there any streaming concerns with this? I feel like surely we made it so there's a newline after the doctype for a reason.

will it break something ?

ematipico (approved):

natemoo-re:
> will it break something ?

I don't think so. Newlines typically mean the browser will flush the buffer and attempt to parse the latest chunk as HTML, so in this case it will process the doctype when the first chunk is flushed. Seems like fine behavior to me.

natemoo-re (approved):

packages/astro/src/core/render/core.ts
```diff
@@ -108,9 +108,10 @@ export type RenderPage = {
 	renderContext: RenderContext;
 	env: Environment;
 	apiContext?: APIContext;
+	isCompressHTMl?: boolean
 };
 
-export async function renderPage({ mod, renderContext, env, apiContext }: RenderPage) {
+export async function renderPage({ mod, renderContext, env, apiContext, isCompressHTMl = false }: RenderPage) {
```
bluwy, line 114:
```suggestion
export async function renderPage({ mod, renderContext, env, apiContext, isCompressHTML = false }: RenderPage) {
```

The name isn't spelt right here (ant it's consumers).

Princesseuh:
@matthewp The rest of the function is still faulty, the suggestion only correct one instance

JerryWu1234:
> @matthewp The rest of the function is still faulty, the suggestion only correct one instance

I noticed that and i fixed it

bluwy:
Thanks!

PR #7243, opened 2023-05-30
fix:query not considered in directory redirection

Riki-WangJJ:
## Changes

- What does this change?


natemoo-re (approved):
Thanks for the patch! This looks great.

Added tests and a changeset for you.

PR #7245, opened 2023-05-30
Redirects: Allow preventing the output of the static HTML file

matthewp:
## Changes

- Adds the `build.redirects = false` option so that you can turn off the HTML generated redirects.
- This is for adapters that have their own redirect configuration files (like Netlify `_redirects` who do not want actual HTML files to be generated.
- Updated the relevant adapters.
- See the RFC change here: https://github.com/withastro/roadmap/pull/587/commits/dfd349eefc2251a532b306415ccc320568076e9d

## Testing

- Added a test case in the basic redirects fixture.

## Docs

Coming later.


delucis (approved):
Docs look good! Thanks @matthewp.

Issue #5465, opened 2022-11-24
Stacktrace reports wrong line number


truesri:
### What version of `astro` are you using?

1.6.10

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

When accessing data that is undefined the stacktrace reports the wrong line number. When using no integrations the line number seems to be off by 1, but when using the tailwindcss integration it is off by more.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vdxjk2?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

dsomel21:
@bluwy, are we sure that this is a p3 ("An edge case that only affects very specific usage"). 

It was happening to me in #6372 when I was basically doing:

```astro
---
const response = await fetch(
  "https://clean-dragons-hear-198-54-132-72.loca.lt/api/v1/books"
);
const data = await response.json();
const book = data.books.find((b) => b.id == id); // This is breaking cuz the fetch failed
---
```

Nothing _too_ specific. 

I can look into it, but can you share some details about where to look (is it the `astro` repo, or what).

JoshuaKGoldberg:
It this an appropriate place for me to suggest this be raised to a p2? I agree with @dsomel21, it's really quite a poor onboarding experience for folks like me new to Astro.

I'd love to look into helping fix this if you can give a few starting pointers 🙂

Princesseuh:
> It this an appropriate place for me to suggest this be raised to a p2? I agree with @dsomel21, it's really quite a poor onboarding experience for folks like me new to Astro.
> 
> I'd love to look into helping fix this if you can give a few starting pointers 🙂

Sorry for the late answer, I didn't see your comment. The causes of this are quite various, so it's generally hard to pinpoint the problem and give guidance.

In most cases, apart from CSS errors, the reason is source map related, either the source map that our compiler generated isn't accurate, or the source map isn't linked correctly / configured correctly.

For frontmatter errors in particular, there's a bit of a special situation where the frontmatter is reparsed to give a better error message: <https://github.com/withastro/astro/blob/72c6bf01fe49b331ca8ad9206a7506b15caf5b8d/packages/astro/src/vite-plugin-astro/compile.ts>, which I feel might be partially responsible? I'm not sure. 

[...]

PR #7273, opened 2023-06-02
Fix error stacktrace from Vite SSR runtime

bluwy:
## Changes

Part of the fix for https://github.com/withastro/astro/issues/5465

When we call `ssrLoadModule` to get a module processed by Vite, and call the module's runtime (e.g. it's exported function), we also need to fix the stacktrace there.

Previously we're [only fixing the stacktrace](https://github.com/withastro/astro/blob/80e86d65fa5c16247f6b6696152d0edea065947e/packages/astro/src/core/render/dev/index.ts#L62-L77) when we call `ssrLoadModule` (module init).

Also, when we call MagicString's `generateMap()` API, we need to pass `hires: true` too. Otherwise a low res map combined with another map (for multiple plugin transforms) will degrade the sourcemap accuracy/resolution. Causing offset issues.

## Testing

Manually tested with the issues repro

## Docs

n/a. bug fix.


bluwy:
Looks like Vite's SSR stacktrace rewrite made the file path worse 👀 Gonna have to look into Vite

PR #7275, opened 2023-06-02
Run example check in parallel with 5 at most

bluwy:
## Changes

The `test:check-examples` command in CI has been flaky recently. Try running only 5 examples at most in parallel instead  of 28, to ease the CPU.

## Testing

Tested manually locally.

## Docs

n/a.


Princesseuh:
I can feel the CPU being at ease after this change.

matthewp (approved):

PR #7276, opened 2023-06-02
Simplify Deno test

bluwy:
## Changes

Instead of spinning up a new Deno server for each individual test, we can share it similar to the describe-it pattern where each `it` test shares the same server.

In Deno, it's called [test-step](https://deno.com/manual@v1.34.1/basics/testing#test-steps), so I've refactored the code with this pattern. Locally It improves Deno test time from ~25s to ~5s.

## Testing

Ran the test locally.

## Docs

n/a. test refactor.


natemoo-re (approved):
This seems like a great change!

PR #7277, opened 2023-06-02
Add 'starlight' as template alias

natemoo-re:
## Changes

- Adds a built-in template alias for `starlight`.
- Refactors template logic for readability.
- Once released, can be used like so...

```
pnpm create astro --template starlight
```

## Testing

Manual

## Docs

Will be updating the Starlight docs once this is released!


delucis (approved):
LGTM! Thank you for setting this up @natemoo-re 🙌

PR #7288, opened 2023-06-03
Node built-in modules support for Deno adapter

lilnasy:
## Changes
It seems to be a use case that's receiving more and more interest. Deno can leverage built-in node modules, but it's blocked by an overly restrictive configuration set by the adapter.

## Testing
[nodecompat.astro](https://github.com/withastro/astro/pull/7288/files#diff-cde1d6bd05d7fffc08097404004b322be91bd41493cbbfbc49b961f64e73da2a)

## Docs
Not a user facing change, though a note somewhere could be useful.

### DIFF
--- a/.changeset/modern-lobsters-sparkle.md
+++ b/.changeset/modern-lobsters-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/deno': minor
+---
+
+The deno adapter now supports npm package that require built-in node modules.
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
This looks like a great change and the code is straightforward!

I'm definitely on board with this but would love another review before merging.

lilnasy:
> This looks like a great change and the code is straightforward!
> 
> 
> 
> I'm definitely on board with this but would love another review before merging.

Yeah, that sounds good.

lilnasy:
I would have liked to include an nom library that uses a supported node built-in module, but couldn't find one that's appropriate for CI.

matthewp (approved):

Issue #5665, opened 2022-12-22
Svelte island does not expect class prop by default, causing warning


Shmuppel:
### What version of `astro` are you using?

1.7.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

A lot of work has already gone into passing Astro classes to Svelte islands (#5397, #5112), currently it seems the default behaviour is to pass the scoped classname as a prop (class="astro-VFKJLXQS"). However, Svelte components by default don't expect this prop unless explicitly used in the Svelte component. This causes a console warning: `<Example> was created with unknown prop 'class'`.  

I'm hesitant to bring it up because it seems like more of a Svelte issue than an Astro one, but it does cause unexpected warnings which can be considered bad DX. 

Thanks guys!

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/sweet-james-kxt2e3

### Participation

- [ ] I am willing to submit a pull request for this issue.

narration-sd:
Just spent a time on this myself, without discovering an external workaround, but indeed tis is something apparently internal to Svelte...not to be fixed, as it's supposed to be a dev 'warning'' to 'help'.

Someone ought to speak to them about the psychology of this, but the good thing is that it and others do go away for production. You can prove it to yourself  by doing a  build and then preview with your Astro project.

Still agree well with @Princesseuh 's categorization...

SrJSDev:
Just came here from a Google search, after a similar incident while using Astro + Svelte:

`index.svelte? [sm]:20 <App> was created with unknown prop 'class'`

Not crazy bad or anything, but annoying, as I cannot even `export let class = ""` in my Svelte component - it's invalid JS syntax.

Is there a reason Astro must pass in a `class` prop?

HiDeoo:
A *dirty* workaround could be the `@astrojs/svelte` renderer `clientEntrypoint` containing a console filter like the ones already existing in Astro, e.g. [this one](https://github.com/withastro/astro/blob/a1c0cbe604c9f91cdc421b5606aab574999eba01/packages/astro/src/runtime/server/jsx.ts#L227-L249) or [the one in the preact integration](https://github.com/withastro/astro/blob/a1c0cbe604c9f91cdc421b5606aab574999eba01/packages/integrations/preact/src/server.ts#L72-L94), filtering out warnings specifically targeting the `class` prop.

[...]

PR #7291, opened 2023-06-04
Filter out Svelte's unknown class prop console warnings

kitschpatrol:
## Changes

Astro's hydration code passes a `class` prop to Svelte components, inducing Svelte to log a warning about an unknown prop.

Preempting this by exporting a `class` prop from the Svelte component isn't a viable workaround since `class` is a reserved identifier in JS.

This PR implements the console-filtering workaround suggested by @HiDeoo in #5665, borrowing the `useConsoleFilter` approach from the [preact integration](https://github.com/withastro/astro/blob/a1c0cbe604c9f91cdc421b5606aab574999eba01/packages/integrations/preact/src/server.ts#L72-L94).

It would probably be better to generalize the console filtering process so it could be shared across multiple integrations.

Ideally there would be a way to handle this in Svelte, but as was pointed out in the issue thread even they resort to [similar hackery](https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/client/client.js#L1974-L1996) in SvelteKit.

Note that you may need to clear your Vite cache to get this change to stick:
```sh
rm -rf ./node_modules/.vite/
```

## Testing

1. Create an Astro + Svelte project: 
    ```sh
    npm create astro@latest -- --template framework-svelte
    ````

2. Run the project and observe the browser console's warning output:
    ```sh
    npm run dev
    ```

    Console output:
    ```
    <Counter> was created with unknown prop 'class'
    ```

3. Apply the patch

4. Clear the vite cache:
    ```sh
    rm -rf ./node_modules/.vite/
    ```

2. Run the project and observe the browser console's (lack of) warning output:
    ```sh
    npm run dev
    ```

    Console output:
    ```
    ```

## Docs

This is a non-breaking change that offers a minor quality of life improvement, not sure documentation beyond what's included in the source comments is necessary.


packages/integrations/svelte/client.js
bluwy, line 1: I think we only need this filter in dev so we don't include the extra JS in prod. Perhaps we need another `client-dev.js` file specifically, which re-exports `client.js` and includes the `console.warn` filter globally similar to the SvelteKit code.

kitschpatrol: Oh good point, didn't realize it was dev only.

natemoo-re:
I know the console messages are irritating but I don't love console filtering as a general approach... if this is spreading to multiple renderer integrations, maybe we need to rethink how the Svelte renderer works?

kitschpatrol:
Yeah I'm not crazy about console filtering either, but if you decide it's worth it I've updated the PR with an `import.meta.env.DEV` check as suggested. Seems a little simpler than maintaining an additional `client-dev.js` file. Looks like DCE is working as expected.

wackbyte:
FYI, you *can* export a `class` prop, like so:
```js
let className;
export { className as class };
```

bluwy (approved):
I'm personally fine with the patching to avoid the confusion in the logs, but will hold off merging before discussing with @natemoo-re first.

natemoo-re:
Sorry for the delay! I'm fine with merging this since it should help avoid confusion, appreciate the PR!

natemoo-re (approved):

PR #7296, opened 2023-06-05
fix(html): Export HTML components as default export so their type is valid JSX

Princesseuh:
## Changes

The export shape for HTML component is an object with a `render` function on it, this is very cool but unfortunately, it's an invalid type for JSX. This PR changes it to be a default exported function instead, which doesn't change the usage.

This PR also includes a small change to the type reference in `env.d.ts` to point to `client-base.d.ts` instead of `client.d.ts`, as the later includes some definitions that we should be overwriting later on

Fix https://github.com/withastro/language-tools/issues/447
Fix https://github.com/withastro/language-tools/issues/557

## Testing

Tests should still pass!

## Docs

N/A

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -17,7 +17,7 @@ import {
 } from './astro/index.js';
 import { Fragment, Renderer, stringifyChunk } from './common.js';
 import { componentIsHTMLElement, renderHTMLElement } from './dom.js';
-import { renderSlots, renderSlotToString, type ComponentSlots } from './slot.js';
+import { renderSlotToString, renderSlots, type ComponentSlots } from './slot.js';
 import { formatList, internalSpreadAttributes, renderElement, voidElementNames } from './util.js';
 
 const rendererAliases = new Map([['solid', 'solid-js']]);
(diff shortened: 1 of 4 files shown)


bluwy:
The change makes sense to me, but it looks like the tests are failing.

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -51,7 +51,7 @@ function isFragmentComponent(Component: unknown) {
 }
 
 function isHTMLComponent(Component: unknown) {
-	return Component && typeof Component === 'object' && (Component as any)['astro:html'];
+	return Component && (Component as any)['astro:html'];
```
bluwy, line 54:
```suggestion
	return Component && (Component as any)['astro:html'] === true;
```

Changing to this fixes it for me. It's because of this proxy component that's tripping the check:

https://github.com/withastro/astro/blob/25c50300b52c680b64eea6d6a10557ec17029a73/packages/astro/test/fixtures/solid-component/src/components/ProxyComponent.jsx#L9-L14

where `Component['astro:html']` would return an actual JSX component.

Princesseuh: Ooh, that's hilarious. I saw the Proxy component but missed this, thank you very much!

bluwy (approved):

Issue #7222, opened 2023-05-26
Spreading possibly undefined attributes on HTML element throws an Exception


Edo-San:
### What version of `astro` are you using?

2.4.1

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Brave/Chrome

### Describe the Bug

Hi everyone 👋 

I created a [stripped-down example](https://stackblitz.com/edit/withastro-astro-stjnvx?file=src%2Fpages%2Findex.astro) of this issue I've been facing lately.

Basically, if I try to spread a possibly `undefined` object on an HTMLElement, Typescript does not complain since the object may also be defined (take a look at [this quick playground](https://www.typescriptlang.org/play?target=99#code/PTAEFEA8AcBsEsDG8AuoCuA7AJgUwGbya7YBQiA9pgM5oCGoAvBjgUSeVbaAEZOgBvAHQi6AX1KkQoALJ0Anr1ws8hYmUo00ifnJQALIQCc6OCgFsAFAEpQAHlAAGIQFZOW0Nn46A-IND6RgQAXKAARPgUFGFioKFYquwaXGjKzMIi2BLu1BSwuEKwFADmlgI8ADSguGLWkqQo8tDKAArw-ABKuJRG2Ha0RkTFVQNDAHwNTcoAiuj8AqAA2gDWuPKho5jFALobKINboGJAA) for reference), but at runtime, if the object is `undefined`, the [spreadAttribute](https://github.com/withastro/astro/blob/6c7df28ab34b756b8426443bf6976e24d4611a62/packages/astro/src/runtime/server/index.ts#L71) function throws an Error because `Object.entries` is called on `undefined`.

AFAIK this code would be valid in JSX, for instance, and Typescript does not complain about it, so I think that function should be able to handle `undefined` values as well, but I was wondering if the current implementation was indeed intended as it is or not 🤔 

If so, I'll close this issue, otherwise, I'd be more than happy to contribute!

Keep rocking! 🙌 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-stjnvx?file=src%2Fpages%2Findex.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

bluwy:
Makes sense to me! A simple `= {}` on this line should fix it:
[...]

PR #7310, opened 2023-06-06
Fix/update spread attribute

Edo-San:
## Changes

Add default value to `spreadAttributes` `values` argument, in order to prevent the function from throwing when spreading potentially undefined values.

This PR should fix issue #7222 

## Testing

No tests were added, as it is a fallback mechanism

## Docs

No harmful changes in user behaviour 😉


packages/astro/src/runtime/server/index.ts
```diff
@@ -69,7 +69,7 @@ export function __astro_tag_component__(Component: unknown, rendererName: string
 
 // Adds support for `<Component {...value} />
 export function spreadAttributes(
-	values: Record<any, any>,
+	values: Record<any, any> = {},
```
natemoo-re, line 72: Hmm, is this actually desirable behavior? I think the spread operator would throw in plain JavaScript and JSX if it is used on an `undefined` value.

Edo-San: Hey Nate 👋 TY for your reply.
Aside from the playgrounds I listed in the original issue #7222 I also tried it out on a JSX playground and I had no errors when spreading an `undefined` value.
I don't know if this is the desirable behaviour, but it seemed reasonable to me since JS, TS and JSX are not complaining (unless I've been missing something here) 🤔

natemoo-re: Ah, I totally missed the discussion in the original issue. I think this makes sense since TypeScript doesn't complain.

I'm cool with this change!

natemoo-re (approved):

PR #7314, opened 2023-06-06
@astrojs/svelte: Update to svelte2tsx 0.6.15 and vite-plugin-svelte 2.4.1

wackbyte:
## Changes

- Bump svelte2tsx from ^0.5.11 to ^0.6.15
- Bump @sveltejs/vite-plugin-svelte from ^2.1.1 to ^2.4.1
- Bump svelte from ^3.44.0 from ^3.55.0 (required by svelte2tsx)

Now that svelte2tsx supports TypeScript 5.x, I think its "packageExtensions" can be removed? (I'm not sure)

## Testing

I did not add any tests as this is a dependency bump.

## Docs

This is a small dependency bump and does not require any documentation changes.


natemoo-re (approved):
Seems like a reasonable change, thanks for updating these deps!

bluwy:
This is a breaking change as it changes the peer deps required by the dependencies. We'd need to mark it a major in changesets.

Issue #7279, opened 2023-06-02
create-astro: set custom registry during project scaffold


calebdwilliams:
### What version of `astro` are you using?

latest

### Are you using an SSR adapter? If so, which one?

N/A

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

N/A

### Describe the Bug

My workspace currently blocks access to the global npm registry, instead filtering all traffic through an internal registry. Unfortunately that means running the scaffold command `npm create astro@latest` results in the following error: 

```
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: Client network socket disconnected before secure TLS connection was established
    at connResetException (node:internal/errors:705:14)
    at TLSSocket.onConnectEnd (node:_tls_wrap:1594:19)
    at TLSSocket.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1358:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ClientRequest instance at:
    at TLSSocket.socketErrorListener (node:_http_client:494:9)
    at TLSSocket.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'ECONNRESET',
  path: null,
  host: 'registry.npmjs.org',
  port: 443,
  localAddress: undefined
}
```

I need some way to point the create script at a local registry rather than the global npm registry. This value can then be passed down to [`giget's --registry flag`](https://github.com/unjs/giget/blob/main/README.md#options). Unfortunately setting the environment variable doesn't seem to help. 

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
fixed in https://github.com/withastro/astro/pull/7326

PR #7326, opened 2023-06-07
Fixes an issue where create Astro doesn't respect custom npm registries

calebdwilliams:
## Changes

- Introduced a fix for #7279 by introducing a utility to check the preferred package manager config to find any custom package registry, defaulting to the public npm registry (which is currently in use)

## Testing

I'm not sure how to effectively test this fix without affecting the entire build process. Happy to accept suggestions.

## Docs

This should enable Astro to behave as expected for users with custom package registries (such as large enterprises who might block the public npm registry for security or efficiency reasons).

### DIFF
--- a/.changeset/thirty-books-smoke.md
+++ b/.changeset/thirty-books-smoke.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Ensure create-astro respects package manager registry configuration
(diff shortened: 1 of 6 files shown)


natemoo-re (approved):
This looks great, thanks for the PR!

Discussed on Discord and confirmed locally that these changes work both when a user has set a custom registry and when a user has no custom registry defined. Package managers properly return the default NPM registry in the latter case.

PR #7340, opened 2023-06-08
fix: encode spaces when generating `srcset`

### LINKED ISSUE #7338: `getPicture` may generate invalid `srcset` values after SSR

2.4.1

None

npm

Ubuntu (WSL)

Brave

### Describe the Bug

Filenames that contain space characters result in invalid `srcset` values. Images are loaded anyway, but it always render the `src` of the `img`, never ones defined in `srcset`.

Let's say the filename is `HELLO WORLD`. There're no problems with `getImage` and `img` because `` will request `HELLO%20WORLD`. The entire `src` value is considered as URL and thus properly encoded. But when it's `srcset`, `HELLO WORLD 100w` is invalid as spaces have meaning. The proper value should be `HELLO%20WORLD` here.

This only happens after SSR, because dev server puts the filename in search params which gets URL encoded anyway. When examining the example, you can refer to the console output that warns about `srcset` parsing error.

- StackBlitz (dev): https://stackblitz.com/edit/github-3mwabz?file=src%2Fpages%2Findex.astro
- GitHub Pages (prod, because SSG build won't work on StackBlitz): https://stackblitz.com/edit/github-3mwabz?file=src%2Fpages%2Findex.astro

https://github.com/wooseopkim/astro-picture-srcset-repro

### Participation

wooseopkim:
## Changes

- Fixes #7338.
  - `HELLO WORLD.png` currently will have `srcset` of `HELLO WORLD.png 100w`.
  - In that case any `source`s are ignored and `img`'s `src` will be rendered.
  - Browser complains, but it's just `console.warn` calls.

```
Failed parsing 'srcset' attribute value since it has an unknown descriptor.
Dropped srcset candidate "/_astro/HELLO"
```

## Testing

- `packages/integrations/image/test/picture-ssg.test.js`
- `packages/integrations/image/e2e/picture.test.js`
  - Is it okay to duplicate the whole `basic-picture` fixture?

## Docs

- Fixed bug in `Picture` and `getPicture` with space characters in filenames.


packages/integrations/image/src/lib/get-picture.ts
```diff
@@ -85,7 +85,7 @@ export async function getPicture(params: GetPictureParams): Promise<GetPictureRe
 					image = img;
 				}
 
-				return `${img.src} ${width}w`;
+				return `${img.src?.replaceAll(' ', '%20')} ${width}w`;
```
wooseopkim, line 88: AFAIK encoding the whole `img.src` was not required. I've tested the filename `@withastro:안녕하세요🚀+❤.jpg` which contains an emoji, URI reserved characters, and Korean chracters but it seemed okay before (and also after) applying this.

matthewp (approved):

Princesseuh:
The code looks great to me, however the testing strategy is very heavy-handed and we've been trying to improve our testing strategy recently.

I'd suggest not using a full E2E test for this and instead just adding a test to this file: https://github.com/withastro/astro/blob/main/packages/integrations/image/test/picture-ssg.test.js

wooseopkim:
@Princesseuh Yeah, agreed and understood. Actually [I've also updated `test/picture-ssg.test.js`](https://github.com/withastro/astro/pull/7340/files#diff-57138d4b656b3b7fea899435de544b8e87f9871f9f2440ce7932bd51e9ad28c6). In this case, would it be enough to delete the `e2e` directory?

bluwy:
I think we can remove playwright entirely with the new test. I'm also slightly leaning towards `encodeUri` as the more correct way of handling this

bluwy (approved):

PR #7347, opened 2023-06-09
fix(errors): Deprecate error codes

Princesseuh:
## Changes

Error codes are not useful when you have other static information that people can use. People remember HTTP error codes because they're ubiquitous, people have been seeing them for 30 years now and there's great marketing on some of them (funny 404 pages).

In our case, all of our errors have names that are static and people can find that are much easier to remember

This is not breaking: We never showed those codes to users in Astro itself.

## Testing

Tests should still pass!

## Docs

<https://github.com/withastro/docs/pull/3463>


matthewp (approved):

bluwy (approved):
Looks great! Seems like it's a continuation from https://github.com/withastro/astro/pull/7089#discussion_r1193837105

Princesseuh:
Something I just realized with this PR is that the logic that handles if we show the link to the docs or not needs to be updated since people will start using AstroError with unknown errors

bluwy:
Do you mean updating this part?

https://github.com/withastro/astro/blob/3830b6888108eb7c58f03792c8285ae5ddc855a6/packages/astro/src/core/errors/dev/vite.ts#L140-L149

The current logic is a bit strange, but I think it should also cover unknown errors.

bluwy:
Don't want to further distract your time off 😬 Should we leave it as is for now? Or refactor that to something like:
```js
const hasDocs = err.name in AstroErrorData
```
Not really sure what was the original intent of the code.

bluwy:
Merging in as the docs changes are only removing references of the error codes, and the rest are handled in https://github.com/withastro/docs/pull/3463

PR #7367, opened 2023-06-12
Update example dependencies versions with a script

bluwy:
## Changes

Add a script to update the example deps versions.

- Partial revert of https://github.com/withastro/astro/pull/7248, which reverted https://github.com/withastro/astro/pull/6868 (so 6868 gets re-instated)
- Remove changeset patch as it's not used to control publish filtering anymore

## Testing

Manually ran `pnpm run version` locally

## Docs

n/a


delucis (commented):
LGTM! Not quite sure what to make of the cancelled/skipped CI jobs given they‘re not really testing this PR. Existing tests should still pass though right?

bluwy:
Yeah it should still pass. I've been seeing similar failures recently here where CI would bail completely without any reason. I've merged the branch to re-trigger CI again.

scripts/deps/update-example-versions.js
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	await fs.writeFile(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
+}
+
+/**
+ * @param {string} packageJsonPath
+ */
+async function tryParsePackageJson(packageJsonPath) {
```
ematipico, line 62: nit: coming from Rust, usually `try*` functions are meant for those functions that **might** cause intentional errors, and the consumers need to handle the errors.

scripts/deps/update-example-versions.js
```diff
@@ -0,0 +1,66 @@
...
+const workspaceDirs = await glob(rootPackageJson.workspaces, {
+	onlyDirectories: true,
+	cwd: rootUrl,
+});
+for (const workspaceDir of workspaceDirs) {
+	const packageJsonPath = path.join(workspaceDir, './package.json');
+	const packageJson = await tryParsePackageJson(packageJsonPath);
+	if (!packageJson) continue;
+
+	packageToVersions.set(packageJson.name, packageJson.version);
```
ematipico, line 29: Should we throw an error if `name` or `version` don't exist?

scripts/deps/update-example-versions.js
```diff
@@ -0,0 +1,66 @@
...
+});
+for (const exampleDir of exampleDirs) {
+	const packageJsonPath = path.join(exampleDir, './package.json');
+	const packageJson = await tryParsePackageJson(packageJsonPath);
+	if (!packageJson) continue;
+
+	// Update dependencies
+	for (const depName of Object.keys(packageJson.dependencies ?? [])) {
+		if (packageToVersions.has(depName)) {
+			packageJson.dependencies[depName] = `^${packageToVersions.get(depName)}`;
```
ematipico, line 45: nit: should we just pin `~` - bug fixes?

Personally, I would pin the versions without ranges. It's known that there are libraries that break the CIs with patches.

ematipico (approved):

scripts/deps/update-example-versions.js
delucis: Worth bearing in mind we’re bumping `astro` or `@astrojs/*` integrations in our example templates here. If we break CI with patches, that’s probably on us, no?

bluwy: I think `^` should be good here as it's what we used to use too. Since these are our packages that we're bumping, I don't think we're making breaking changes in minors or patches that needs this restriction.

scripts/deps/update-example-versions.js
bluwy: Renamed this as `readAndParsePackageJson`

scripts/deps/update-example-versions.js
bluwy: Added checks for those!

Issue #7230, opened 2023-05-28
@astrojs/lit experimental-hydrate-support update


khmyznikov:
### What version of `astro` are you using?

2.4.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

experimental-hydrate-support.js [was removed](https://github.com/lit/lit/pull/3765) from lit package and now leaves in [@lit-labs/ssr](https://lit.dev/docs/ssr/client-usage/#loading-@lit-labsssr-clientlit-element-hydrate-support.js)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-uvfjud?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #7373, opened 2023-06-12
Upgrade lit deps

matthewp:
## Changes

- Closes https://github.com/withastro/astro/issues/7230

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/integrations/lit/hydration-support.js
+++ b/packages/integrations/lit/hydration-support.js
@@ -1,1 +1,2 @@
-import 'lit/experimental-hydrate-support.js';
+// @ts-check
+import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
--- a/packages/integrations/lit/package.json
+++ b/packages/integrations/lit/package.json
@@ -42,8 +42,8 @@
     "test": "mocha"
   },
   "dependencies": {
-    "@lit-labs/ssr": "^3.1.0",
-    "@lit-labs/ssr-dom-shim": "^1.1.0",
+    "@lit-labs/ssr": "^3.1.3",
+    "@lit-labs/ssr-dom-shim": "^1.1.1",
     "parse5": "^7.1.2"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


packages/integrations/lit/hydration-support.js
```diff
@@ -1 +1,2 @@
-import 'lit/experimental-hydrate-support.js';
+// @ts-check
+import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
```
bluwy, line 2: Do we need to add `@lit-labs/ssr-client` into `dependencies`?

packages/integrations/lit/package.json
```diff
@@ -42,21 +42,21 @@
...
     "cheerio": "^1.0.0-rc.11",
-    "lit": "^2.7.0",
+    "lit": "^2.7.5",
     "mocha": "^9.2.2",
     "sass": "^1.52.2"
   },
   "peerDependencies": {
     "@webcomponents/template-shadowroot": "^0.2.1",
-    "lit": "^2.7.0"
+    "lit": "^2.7.5"
```
bluwy, line 60: Updating the peer deps would be a breaking change (and a major). Do we need to bump the `lit` version here?

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
bluwy, line 1: I think you'd need to use pnpm 8.6 and above to generate the lockfile v6.1

packages/integrations/lit/package.json
matthewp: No, we don't nee to update this one, will downgrade.

pnpm-lock.yaml
matthewp: Hm, very odd:

> pnpm -v
> 8.6.2

matthewp: Got it.

ematipico (approved):
Looks good to me, after addressing @bluwy 's comment

PR #7377, opened 2023-06-12
Update image tags 

### LINKED ISSUE #7357: Update image tags in examples

^2.5.6

None

npm

Linux

Edge

### Describe the Bug

image
https://github.com/withastro/astro/blob/main/examples/portfolio/src/pages/index.astro

The img tags should be replaced with ``<Image />`` for better performance as explained here https://docs.astro.build/en/guides/integrations-guide/image/

I would like to be assigned to this issue

https://github.com/withastro/astro/blob/main/examples/portfolio/src/pages/index.astro

### Participation

dotarjun:
## Changes

- Replaces ```` tags with ``<Image>`` from ``@astrojs/image``

## Testing

Working on dev environment

## Docs

Docs already exist at https://docs.astro.build/en/guides/integrations-guide/image/

Closes #7357

### DIFF
--- a/examples/portfolio/astro.config.mjs
+++ b/examples/portfolio/astro.config.mjs
@@ -1,4 +1,8 @@
 import { defineConfig } from 'astro/config';
 
+import image from "@astrojs/image";
+
 // https://astro.build/config
-export default defineConfig({});
+export default defineConfig({
+  integrations: [image()]
+});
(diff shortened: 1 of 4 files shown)


matthewp:
Thanks, we don't want to encourage the use of the Image integration as we are migrating people to the core image story. Given that it's still experimental it doesn't seem like the right thing to add yet, either. As we approach 3.0 that's a good thing to change here.

Issue #7195, opened 2023-05-24
Built code has no access to the run-time environment variables


arty-name:
### What version of `astro` are you using?

2.5.5

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

yarn

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

I use environment variables in my code according to the documentation like this: `import.meta.env.TEST`. They work fine in the development mode: `TEST=Pass astro dev`. After I build the code with `astro build` and start the server as `TEST=Pass astro preview`, the running application has no access to the runtime variables. In the generated code, I see that `import.meta.env.TEST` has been replaced with `({}).TEST`. To reproduce, open the linked example, stop the development server in the console, run `turbo build && turbo preview` and reload the view.

What made debugging complicated was the presence of `.env` file (and the inability of Stackblitz to persist it). If the `.env` file exists when `astro build` is run, then `import.meta.env.TEST` is replaced with the value of this variable from the .env file. This makes the built code ignore the environment variables configured during run-time.

Yet another quirk appears when there is no `.env` file however the environment variable is provided at build time: `TEST=Nevermind astro build`. This results in the expected outcome: `import.meta.env.TEST` is replaced with `process.env.TEST` in the generated code. To reproduce, open the linked example, stop the development server in the console, run `turbo working-build && turbo preview` and reload the view.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-uegcud?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
The variable needs to be present during the build. If it is present then it will be fetched at runtime. For example just do: `TEST=whatever astro build` and it will be taken from the ENV at runtime.

arty-name:
@matthewp it will not be fetched at runtime, this is what my demo showcases

AxelLoupias:
Hello, sorry for commenting here, but if I want the env variable to be public?
I have tried to do what you said but if I add the "PUBLIC_" it will keep the value that was as placeholder and not the one I add in runtime.

arty-name:
The public variable can only be present in scripts, right?

The scripts are only generated during the build, and then are served as static files. Looks like the public variables can only be configured during build, and not during the runtime.

nemanjam:
Is it possible to prevent inlineing env vars at build time and load it in runtime in static builds, similarly like you can with `NEXT_PUBLIC_` prefix in Next.js?

[...]

PR #7378, opened 2023-06-12
Runtime environment variables docs for node integration

arty-name:
## Changes

This PR adds a section to the node integration docs on how to handle the runtime environment variables. It is not documented anywhere, and the current behavior even caused me to [file a bug about it](https://github.com/withastro/astro/issues/7195).

Since I am filing this PR via the GitHub website I cannot provide a changeset.

## Testing

I have added no tests for the modified README file because none existed.

## Docs

This is somehow a docs PR


packages/integrations/node/README.md
```diff
@@ -150,6 +150,10 @@ You can pass the path to your key and certification via the environment variable
 SERVER_KEY_PATH=./private/key.pem SERVER_CERT_PATH=./private/cert.pem node ./dist/server/entry.mjs
 ```
 
+#### Runtime environment variables
+
+If an `.env` file is present when the build process is run, the values of environment variables provided by this file will be hard-coded in the output, same as when generating a static website. Special handling is required for values that have to be provided at run-time. Such variables must exist with any value during the build process: `RUNTIME_VALUE=placeholder astro build`. The placeholder value will be ignored by the build process, however its existence will signal to Astro that the actual value will be available as an environment variable when the built application will be run. In the case of multiple variables, you can store them in a file like `.env.runtime` and then start the build with the following command: `export $(cat .env.runtime) && astro build`. Note that these variables **must be absent** in the `.env` file.
```
sarah11918, line 155:
``````suggestion
If an `.env` file containing environment variables is present when the build process is run, these values will be hard-coded in the output, just as when generating a static website. 

For a value to be provided at run-time, use the following variable with any placeholder value in your `.env` file:  `RUNTIME_VALUE=placeholder astro build`.  This variable will signal to Astro that the actual value will be available when the built application is run. The placeholder value will be ignored by the build process, and Astro will use the value provided at run-time. 

In the case of multiple run-time variables, store them in a seperate file (e.g. `.env.runtime`) outside of `.env`. Start the build with the following command: 
```sh
export $(cat .env.runtime) && astro build
```
``````
I've edited this to split up the content into easier to read chunks, and tried to tighten up some wording while I'm at it. Of course, when I do this, I might often cut something important or miss a nuance, so please have a read over this as a suggested model, and don't hesitate to correct so that it is actually factually accurate!  :smile:

**Review by sarah11918 — COMMENTED** [review]
Thanks for this docs contribution @arty-name! I think this is very useful information to have!  

See my take on an editing pass below to fit it in with our current docs style, and see what you think!

File: packages/integrations/node/README.md (hunk 1 above)
Comment (arty-name) [reply packages/integrations/node/README.md:155]: Thank you, Sarah, I’m always happy to get an input from a tech writer and a native speaker!

This topic is unfortunately overly complex, and I see that I could not explain it clearly enough. For one, it’s not the single `RUNTIME_VALUE` which needs to be provided: during the build the developer needs to provide *every* value which will also come in during the execution. Another important part which lost its emphasis is this: "Note that these variables **must be absent** in the `.env` file." 

What do you think about this improved version?

>If an `.env` file containing environment variables is present when the build process is run, these values will be hard-coded in the output, just as when generating a static website. 

>Special handling is required for values that have to be provided at run-time. During the build, they **must be absent** from the `.env` file. But you also need to tell Astro which environment variables to expect at run-time. For that you need to provide them during the build process: `VALUE1=placeholder VALUE2=placeholder astro build`. Feel free to replace `placeholder` with any non-falsy value, it will be ignored anyway.

>Providing multiple values in this way could be cumbersome. For convenience, store them in a seperate file (e.g. `.env.runtime`). Start the build with the following command: 
```sh
export $(cat .env.runtime) && astro build
```

TheOtterlord, line 155: > https://github.com/withastro/astro/pull/7378#discussion_r1233608066

After the first section, you explain how to use it with multiple variables. We wouldn't want to lead people down the wrong path before asking them to do it another way. We can probably still make this clear in the text with something like "you must provide Astro with every environment variable".

In the final section, we shouldn't need to mention it's convenience, because we're trying to teach the user the one, easy way to do something.

```suggestion
If an `.env` file containing environment variables is present when the build process is run, these values will be hard-coded in the output, just as when generating a static website.

During the build, the variable must be absent from the `.env` file, and you must provide Astro with every environment variable to expect at run-time: `VARIABLE_1=placeholder astro build`. This signals to Astro that the actual value will be available when the built application is run. The placeholder value will be ignored by the build process, and Astro will use the value provided at run-time.

In the case of multiple run-time variables, store them in a seperate file (e.g. `.env.runtime`) from `.env`. Start the build with the following command:

```sh
export $(cat .env.runtime) && astro build
```
```

Comment (TheOtterlord) [reply packages/integrations/node/README.md:155]: Sorry for the long wait. Our docs reviewing/editing has had a huge backlog and I'm picking this up for Astro Docs Maintainers!

This is looking much better! I have a few suggestions based on your revision, of which I've thrown in a new review.

File: packages/integrations/node/README.md [hunk 2]
```diff
@@ -150,6 +150,17 @@ You can pass the path to your key and certification via the environment variable
 SERVER_KEY_PATH=./private/key.pem SERVER_CERT_PATH=./private/cert.pem node ./dist/server/entry.mjs
 ```
 
+#### Runtime environment variables
+
+If an `.env` file containing environment variables is present when the build process is run, these values will be hard-coded in the output, just as when generating a static website.
+
+During the build, the variable must be absent from the `.env` file, and you must provide Astro with every environment variable to expect at run-time: `VARIABLE_1=placeholder astro build`. This signals to Astro that the actual value will be available when the built application is run. The placeholder value will be ignored by the build process, and Astro will use the value provided at run-time.
```
arty-name, line 157: Thank you for your input @TheOtterlord, this looks very nice! Before merging it, though, do you think it would make sense to make this clearer? "During the build, the variable must be absent from the `.env` file" does not say which variable. Could it be "the runtime variables"

TheOtterlord: Yep! That sounds like a great clarification

arty-name: I’ve adjusted it like this. How does the PR look to you now?

TheOtterlord (approved):
Thanks a lot @arty-name!
This is amazing! I'll update the branch, and get this merged.

Issue #7097, opened 2023-05-15
Unused Layouts' styles are being imported in Dev Mode


LefanTan:
### What version of `astro` are you using?

2.4.5

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome, Safari

### Describe the Bug

In an Astro project containing multiple Layouts, the global styles would bleed into each other even though a Page might only be using just one layout at a time. Because of the conflicts of CSS that happens only in Dev, this makes working with Layouts extremely difficult on Astro. 

Steps to reproduce:
1. Create a Layout with a `<style is:global />` tag. In my example, I had the following style tag
```
<style is:global>

main {
    background: rgb(255, 227, 196);
}

</style>
```
2. Create a Page (we call this `Page1`) that imports the Layout we just defined. 
3. Create another Page ( we call this `Page2`) that's just a traditional HTML page with a `<main>`
```
<html>
...
  <main>
      Page 2
  </main>
</html>
```
4. Start the dev server: `npm run dev`
5. You'll notice that Page2's `<main>` element will have the background color defined in the Layout, even though it isn't using or importing it. 

Thanks again to the Astro team.
### Link to Minimal Reproducible Example

https://github.com/LefanTan/astro-unused-layout

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Hi, can you provide the repro with stackblitz or a github repo? Codesandbox shows "task killed" for me and doesn't allow users to download the project :(

LefanTan:
@bluwy Hey there, I updated some configurations to the existing Codesandbox, can you try again?

LefanTan:
Just want to bump this issue. Unsure where this lies in priority, but it has made development a nightmare. 

There are cases when the "styling" works in development because some unrelated styles are bleeding into each other but when deployed with production build, things will fail because those styles aren't actually there.

matthewp:
This one is pretty interesting. It combines a few things to make it occur. Essentially this happens:

1. This project uses Tailwind
2. When postcss is run it picks up dependencies. Tailwind's glob pattern makes all of the Astro files a dependency.
3. We call addWatchFile for HMR
4. Internally Vite turns all calls to addWatchFile into the importedModules on the module entry.
5. We add styles in dev based on crawling the graph and looking at importedModules.

[...]

PR #7381, opened 2023-06-12
Prevent accidental CSS inclusion in dev

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/7097
- We can't trust `importedModules` to be true imports. It includes stuff tracked for HMR.
- Using `importers` to verify that an import is an import.
- This is slower, O(n), but necessary. A dev-only issue.

## Testing

- Added a new unit test

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/render/dev/vite.ts
+++ b/packages/astro/src/core/render/dev/vite.ts
@@ -41,6 +41,7 @@ export async function* crawlGraph(
 			continue;
 		}
 		if (id === entry.id) {
+			const urlDeps = entry.ssrTransformResult?.deps ?? [];
 			scanned.add(id);
 			const entryIsStyle = isCSSRequest(id);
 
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/render/dev/vite.ts
```diff
@@ -82,7 +82,7 @@ export async function* crawlGraph(
 						}
 					}
 				}
-				if (!isPropagationStoppingPoint) {
+				if (isImportedBy(id, importedModule) && !isPropagationStoppingPoint) {
 					importedModules.add(importedModule);
 				}
 			}
```
bluwy, line 88: On line 47, maybe we can do this instead?

```diff
-			for (const importedModule of entry.importedModules) {
+			for (const importedModule of entry.ssrTransformResult.deps) {
```

Assuming `ssrTransformResult` would always exist since we're running the Astro component in SSR. This is a bit similar to https://github.com/withastro/astro/pull/6716 when I was refactoring it.

Tailwind strikes again 😦

matthewp: I don't think this is a problem with Tailwind so much. They aren't doing anything wrong here. This comes down to the Vite API changing and us using the new one incorrectly. 

Thanks @bluwy I'll give that option a try.

matthewp: `ssrTransformResult.deps` will work. It is urls (paths), so need to compare against `entry.url`.

PR #7385, opened 2023-06-13
feat: expose locals to render api and from requests in dev mode

ematipico:
## Changes

Please read https://github.com/withastro/astro/pull/7049 for more context.

## Testing

Added test cases

## Docs

### DIFF
--- a/packages/astro/src/core/render/context.ts
+++ b/packages/astro/src/core/render/context.ts
@@ -7,7 +7,7 @@ import type {
 	SSRElement,
 	SSRResult,
 } from '../../@types/astro';
-import { AstroError, AstroErrorData } from '../errors';
+import { AstroError, AstroErrorData } from '../errors/index.js';
 import { getParamsAndPropsOrThrow } from './core.js';
 import type { Environment } from './environment';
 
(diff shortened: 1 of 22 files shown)


ematipico:
@matthewp do you think we should update the docs website and provide an example of how adapters can pass an object, which will be the `locals`?

bluwy (approved):

packages/astro/src/core/render/context.ts
```diff
@@ -51,12 +55,29 @@ export async function createRenderContext(
...
 		...options,
 		origin,
 		pathname,
 		url,
 		params,
 		props,
 	};
+
+	// We define a custom property, so we can check the value passed to locals
+	Object.defineProperty(context, 'locals', {
```
matthewp, line 69: can you add `enumerable: true`? Sometimes get support questions where people try to console.log the context.

matthewp:
@ematipico Yeah I think docs would be good, but I'm not sure how well or if the `App` class is documented at the moment.

Yan-Thomas (approved):
Docs LGTM!

PR #7386, opened 2023-06-13
[Cloudflare integration] Expose cf metadata and Cloudflare caches API

beynar:
## Changes
This pull request expose the Cloudflare `caches` API and cf object (request metadata provided by Cloudflare's edge) to the runtime object.

## Example
Cloudflare add some usefull informations about the request such as geolocation properties see [the docs](https://developers.cloudflare.com/workers/runtime-apis/request#incomingrequestcfproperties)

``` ts
const runtime = getRuntime(Astro.request);
const visitor_country  = runtime.cf?.country;
```

The Cloudflare cache API can now be used.

``` ts
const runtime = getRuntime(Astro.request);
const cachedObject = await runtime.caches?.default.match(Astro.request);
```

### DIFF
--- a/.changeset/angry-socks-sell.md
+++ b/.changeset/angry-socks-sell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Expose cf metadata and Cloudflare Worker Cache API through `caches` in runtime.
--- a/packages/integrations/cloudflare/package.json
+++ b/packages/integrations/cloudflare/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@astrojs/underscore-redirects": "^0.1.0",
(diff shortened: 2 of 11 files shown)


matthewp:
I think this sounds good. When https://github.com/withastro/astro/pull/7385 lands do you think it would make sense to deprecate `getRuntime` and instead put this stuff on `Astro.locals`?

beynar:
Yes, it will be very beneficial and make more sense imo. I'm actually doing it in a middleware. 

Also `getRuntime` is wrongly typed and returns undefined in dev mode. I would have suggest to return a default (` Reflect.get(request, Symbol.for("runtime")) || { name: "cloudflare", env: {}, waitUntil: (cb) => Promise.resolve(cb) }`) or to type its return as potentially undefined.

PR #7391, opened 2023-06-14
Rename tailwind integration options

bluwy:
## Changes

Following up #6724 (major), this renames:

- `config.path` to `configFile`
- `config.applyBaseStyles` to `applyBaseStyles`

The `config.` prefix is a bit redundant 🤔 I also chose `configFile` over `configPath` as that's the same name used by Vite and `vite-plugin-svelte`.

## Testing

Existing test should pass

## Docs

Updated the README to reflect the new option names.

### DIFF
--- a/.changeset/real-spies-pretend.md
+++ b/.changeset/real-spies-pretend.md
@@ -2,7 +2,7 @@
 '@astrojs/tailwind': major
 ---
 
-Let tailwind postcss plugin load its config file itself. This changes the `tailwind.config.js` loading behaviour where Tailwind would load the config file from `process.cwd()` instead of the project `root`. You can configure the integration's `config.path` option to load from a specific path instead.
+Let tailwind postcss plugin load its config file itself. This changes the `tailwind.config.js` loading behaviour where Tailwind would load the config file from `process.cwd()` instead of the project `root`. You can configure the integration's `configFile` option to load from a specific path instead.
 
 ```js
 import { defineConfig } from 'astro/config';
@@ -12,9 +12,7 @@ import { fileURLToPath } from 'url';
 export default defineConfig({
 	integrations: [
 		tailwind({
-			config: {
-				path: fileURLToPath(new URL('./tailwind.config.js', import.meta.url)),
-			},
+			configFile: fileURLToPath(new URL('./tailwind.config.js', import.meta.url)),
 		}),
 	],
 });
--- a/.changeset/real-spies-pretend.md
+++ b/.changeset/real-spies-pretend.md
@@ -2,9 +2,12 @@
 '@astrojs/tailwind': major
 ---
 
-Let tailwind postcss plugin load its config file itself. This changes the `tailwind.config.js` loading behaviour where Tailwind would load the config file from `process.cwd()` instead of the project `root`. You can configure the integration's `configFile` option to load from a specific path instead.
+Let tailwind postcss plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.
+
+If your Tailwind config file is not located in the current working directory, you will need to configure the integration's `configFile` option to load from a specific path:
 
 ```js
+// astro.config.mjs
 import { defineConfig } from 'astro/config';
 import tailwind from '@astrojs/tailwind';
 import { fileURLToPath } from 'url';
@@ -12,15 +15,16 @@ import { fileURLToPath } from 'url';
 export default defineConfig({
 	integrations: [
 		tailwind({
-			configFile: fileURLToPath(new URL('./tailwind.config.js', import.meta.url)),
+			configFile: fileURLToPath(new URL('./tailwind.config.cjs', import.meta.url)),
 		}),
 	],
 });
 ```
 
-This change also requires a Tailwind config file to exist in your project as Astro's fallback value is no longer provided. It is set up automatically during `astro add tailwind`, but you can also manually create a `tailwind.config.cjs` file in your project root:
+This change also requires a Tailwind config file to exist in your project as a fallback config is no longer provided. It is set up automatically during `astro add tailwind`, but if it does not exist, you can manually create a `tailwind.config.cjs` file in your project root:
 
 ```js
+// tailwind.config.cjs
 /** @type {import('tailwindcss').Config} */
 module.exports = {
 	content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
--- a/.changeset/real-spies-pretend.md
+++ b/.changeset/real-spies-pretend.md
@@ -2,7 +2,7 @@
 '@astrojs/tailwind': major
 ---
 
-Let tailwind postcss plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.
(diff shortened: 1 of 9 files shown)


ematipico:
> I also chose `configFile` over `configPath` as that's the same name used by Vite and `vite-plugin-svelte`.

> **Note**: Disclaimer: I don't want to bikeshed. 

Usually, the terminology "path" is used for types/structures that don't necessarily exist and they do not necessarily belong to a physical file. Instead, the terminology "file" is used for those types/structures that read/write to FS, which means they must be derived/computed from "path" that exists.

This means that when I read `configFile`, I assume I can read its contents. Is it the case?

bluwy:
Yes! `configFile` should always be a file path that exist in the filesystem. This value is passed to tailwind, which IIRC the internal code simply does `fs.readFile(path.resolve(configFile), 'utf-8')`.

ematipico:
That's what got me! It's a path, we don't pass the actual contents 😅

bluwy:
I'm confused if you're proposing either `configFile` or `configPath` here 😅 But I think I'm leaning towards the current one for now for consistency.

delucis:
Are we documenting migration steps somewhere if this is a major release with breaking changes? I think the changelog for the major release should include clear descriptions highlighting:

- What broke?
- Under which circumstances? 
- What do you need to do to your existing project to upgrade?

ematipico (approved):
Looks good. I would personally use `configPath`, but it's just a personal preference. No need to block it

bluwy:
@delucis I'm currently documenting them in the changesets. There's two migration needed:
1. option name change (this PR)
2. tailwind config loading behaviour change (https://github.com/withastro/astro/pull/6724)

In this PR, I've tweaked the changeset for no2 to reflect the new option name (`real-spies-pretend.md`). I suppose they don't currently answer your 3 points well that can be improved 😅

.changeset/real-spies-pretend.md
```diff
@@ -2,27 +2,29 @@
 '@astrojs/tailwind': major
 ---
 
-Let tailwind postcss plugin load its config file itself. This changes the `tailwind.config.js` loading behaviour where Tailwind would load the config file from `process.cwd()` instead of the project `root`. You can configure the integration's `config.path` option to load from a specific path instead.
+Let tailwind postcss plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.
```
sarah11918, line 5: Only because Matthew now needs better edited changesets... 😆 ...

"tailwind postcss plugin" is either missing capitalization of these proper nouns, OR if that's the name of the plugin, I'd use it directly in code backticks e.g. `tailwind-postcss` (< - I prefer this option if true, a little, not strongly.)

```suggestion
Let Tailwind PostCSS plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.
```

.changeset/real-spies-pretend.md
```diff
@@ -2,27 +2,29 @@
...
-				path: fileURLToPath(new URL('./tailwind.config.js', import.meta.url)),
-			},
+			configFile: fileURLToPath(new URL('./tailwind.config.cjs', import.meta.url)),
 		}),
 	],
 });
 ```
 
-This change also requires a Tailwind config file to exist in your project as Astro's fallback value is no longer provided. It is set up automatically during `astro add tailwind`, but you can also manually create a `tailwind.config.cjs` file in your project root:
+This change also requires a Tailwind config file to exist in your project as a fallback config is no longer provided. It is set up automatically during `astro add tailwind`, but if it does not exist, you can manually create a `tailwind.config.cjs` file in your project root:
```
sarah11918, line 24: I love the clarity of "if it does not exist!" .. otherwise, you are endorsing both an automatic and possibly a manual one.  :smile:

packages/integrations/tailwind/README.md
```diff
@@ -84,9 +84,9 @@ If it isn't there, you add your own `tailwind.config.(js|cjs|mjs)` file to the r
 
 The Astro Tailwind integration handles the communication between Astro and Tailwind and it has its own options. Change these in the `astro.config.mjs` file (_not_ the Tailwind configuration file) which is where your project's integration settings live.
 
-#### config.path
+#### configFile
```
sarah11918, line 87:
```suggestion
#### config.path (deprecated in vxxxxx)

In vxxxxx, `config.path` was replaced with [configFile](#configFile). Please update to the latest version of `@astrojs/tailwind` and [update your integration config](add a section showing the code example from the file diff and link to it!) accordingly.

#### configFile (added in vxxxx; replaces `config.path`)
```

What about something *like* this here? Not removing the field entirely but showing deprecation, and for which version?

Then, we'll make a section in the README itself for updating, exactly like you showed in the changeset, that can be used as reference.

I think leaving this in for a while is helpful enough to anyone troubleshooting, looking at the OTHER updated sections' code examples and asking themselves, "Wait, what the heck is `configFile`? I have `config.path`?? and then they come see this here and it's easy to see that it's a 1:1 swap of names.

packages/integrations/tailwind/README.md
```diff
@@ -100,14 +100,14 @@ import tailwind from '@astrojs/tailwind';
...
   integrations: [tailwind({
     // Example: Provide a custom path to a Tailwind config file
-    config: { path: './custom-config.cjs' },
+    configFile: './custom-config.cjs',
   })],
 });
 ```
 
-#### config.applyBaseStyles
+#### applyBaseStyles
```
sarah11918, line 108:
```suggestion
#### config.applyBaseStyles (deprecated in vXXXXX)

In vxxxxx, `config.applyBaseStyles` was replaced with [applyBaseStyles](#applyBaseStyles). Please update to the latest version of `@astrojs/tailwind` and [update your integration config](add a section showing the code example from the file diff and link to it!) accordingly.

#### applyBaseStyles
```
Same idea?

.changeset/real-spies-pretend.md
bluwy: Yeah a lot of this is written in a semi-informal way 😄 It is the `tailwindcss` PostCSS plugin, so we could use that instead:
```suggestion
Let the `tailwindcss` PostCSS plugin load its config file itself. This changes the Tailwind config loading behaviour where it is loaded from `process.cwd()` instead of the project `root`.
```

packages/integrations/tailwind/README.md
sarah11918: > If they're upgrading a major, they should check the changelog instead.

I feel like you're expecting a lot of users here. 😅 And remember, MOST people will see this page *in docs*, where they won't have easy/direct access to the changelog.

The above shorter is fine, BUT I would add something with a link to the changelog then that says something like:

> See the changelog for updating your config

? What about that? Then, whatever's decided here, do the same below? I think with a link to follow where they can see the example in the changelog is OK. I might later add something myself in the way of an example of the updated config, but as long as there's a clear path to that changelog, that works so this can be non-blocking!

bluwy: A link to the changelog works for me!

I think eventually everyone has to learn to find the changelog when updating majors. Not all package has docs explaining migrating X to Y. Maybe I'm a bit harsh to people 😄 

Anyways I give the changelog suggestion a go first.

sarah11918 (approved):
Thanks, @bluwy! I think these look good now!

PR #7392, opened 2023-06-14
Update astro.ts to clarify configuration settings for base

r-southworth:
## Changes

- What does this change? 
      - Clarifies expected behavior of import.meta.env.BASE_URL when using base configuration

- I don't think this change needs a changeset?

## Testing

Not sure what testing is supposed to be done since it is just a docs wording change just trying to move this pull request to the correct spot. https://github.com/withastro/docs/pull/3349

## Docs

This change is made to clarify the existing documentation.


ematipico (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -545,10 +545,12 @@ export interface AstroUserConfig {
 	 * ```
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
+	 * 
+	 * By default, the value of `import.meta.env.BASE_URL` includes a trailing slash. If you have [`trailingSlash`](/en/reference/configuration-reference/#trailingslash) set to `'never'`, then you will need to add the trailing slash manually in your static asset imports and URLs
```
Yan-Thomas, line 549: I'm not sure relative links will work accordingly here, from what I saw in the same file we're mostly using absolute links, so to be sure, I'd change that. I also did a few changes in the writing, let me know what you think!
````suggestion
	 * By default, the value of `import.meta.env.BASE_URL` includes a trailing slash. If you have the [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option set to `'never'`, you will need to add it manually in your static asset imports and URLs.
````

Yan-Thomas (commented):
Great work @r-southworth, thanks for this. I've made a suggestion for you! 🙌

packages/astro/src/@types/astro.ts
delucis: Yes, I think @Yan-Thomas is right here — will add his suggestion 👍

delucis (approved):
This looks good to me! Thanks for braving the gauntlet of Astro repos and exploring the hidey-holes where our docs live to help us improve this @r-southworth!

PR #7395, opened 2023-06-15
Update error message for custom slugs in collection

alvinometric:
## Changes

If you had a field called `slug` in a frontmatter in a Content Collection, you would get the following error message:

```
A content collection schema should not contain slug since it is reserved for slug generation. Remove this from your podcasts collection schema.
```

I understood it like the name `slug` was not allowed _at all_. But it _is_ allowed and can be used in frontmatter to override and get a custom slug.

So I asked @sarah11918 and she proposed a better error message.

## Testing

No testing, just a string change.

## Docs

Should not need a doc update since the beginning of the error message stays the same so the docs can keep referencing it.

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -1044,7 +1044,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	 * @see
 	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/)
 	 * @description
-	 * A content collection schema should not contain the `slug` field. This is reserved by Astro for generating entry slugs. Remove the `slug` field from your schema, or choose a different name.
+	 * A content collection schema should not contain the `slug` field. Remove `slug` from your schema in order to use custom slugs in your frontmatter.
 	 */
 	ContentSchemaContainsSlugError: {
 		title: 'Content Schema should not contain `slug`.',


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1044,7 +1044,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	 * @see
 	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/)
```
sarah11918, line 1045:
```suggestion
	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs)
```
More direct link!

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1044,7 +1044,7 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
 	 * @description
-	 * A content collection schema should not contain the `slug` field. This is reserved by Astro for generating entry slugs. Remove the `slug` field from your schema, or choose a different name.
+	 * A content collection schema should not contain the `slug` field. Remove `slug` from your schema in order to use custom slugs in your frontmatter.
```
sarah11918, line 1047:
```suggestion
	 * A content collection schema should not contain the `slug` field. Remove `slug` from your schema. You can still use custom slugs in your frontmatter.
```
I realized after our discussion that to be a little more generic (i.e. maybe some people are getting this error who are NOT trying to use custom slugs, but who think they *need* to specify `slug` when they don't), we could use something like this. What do you think?

sarah11918 (commented):
Thanks for jumping in with this @alvinometric!  🙌  

Had more thoughts, see below!

alvinometric:
Thanks Sarah 🙏

sarah11918 (approved):
Love it! Thanks for the great team work, @alvinometric !  🙌

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1042,9 +1042,9 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
 	/**
 	 * @docs
 	 * @see
-	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/)
+	 * - [The reserved entry `slug` field](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs)
 	 * @description
-	 * A content collection schema should not contain the `slug` field. This is reserved by Astro for generating entry slugs. Remove the `slug` field from your schema, or choose a different name.
+	 * A content collection schema should not contain the `slug` field. Remove `slug` from your schema. You can still use custom slugs in your frontmatter.
```
sarah11918, line 1047:
```suggestion
	 * A content collection schema should not contain the `slug` field. This is reserved by Astro for generating entry slugs. Remove `slug` from your schema. You can still use custom slugs in your frontmatter.
```
Oops, just noticed that we lost the "reserved" property, which I think is important, so I'm adding that back in.

PR #7412, opened 2023-06-18
feat: cache assets in Vercel adapter

kidonng:
## Changes

Add cache headers to assets in Vercel adapter's route config, following [Vercel's default config](https://github.com/vercel/vercel/blob/cfea31e6cff33c0a2dd3663025093fd9b8e7f1d8/packages/frameworks/src/frameworks.ts#L276-L280).

This provides better config out of the box and is more consistent with static only builds, which uses Vercel's default config.

## Testing

[Before](https://nikki-q2suzaqdf-kidonng.vercel.app/2023-01-27): assets are not cached

image

[After](https://nikki-brvcredwz-kidonng.vercel.app/2023-01-27): assets are strongly cached

image

### DIFF
--- a/packages/integrations/vercel/src/edge/adapter.ts
+++ b/packages/integrations/vercel/src/edge/adapter.ts
@@ -151,6 +151,11 @@ export default function vercelEdge({
 					version: 3,
 					routes: [
 						...getRedirects(routes, _config),
+						{
+							src: '^/_astro/(.*)$',
+							headers: { 'cache-control': 'public, max-age=31536000, immutable' },
+							continue: true,
+						},
 						{ handle: 'filesystem' },
 						{ src: '/.*', dest: 'render' },
 					],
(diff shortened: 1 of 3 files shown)


packages/integrations/vercel/src/edge/adapter.ts
```diff
@@ -151,6 +151,11 @@ export default function vercelEdge({
 					version: 3,
 					routes: [
 						...getRedirects(routes, _config),
+						{
+							src: '^/_astro/(.*)$',
```
bluwy, line 155: The `_astro` path can be configured through [`build.assets`](https://docs.astro.build/en/reference/configuration-reference/#buildassets), it would be nice to derive this string from that too.

natemoo-re:
Closing in favor of https://github.com/withastro/astro/pull/7729, but thank you for kicking this off @kidonng!

PR #7419, opened 2023-06-19
fix: cloudflare waitUntil not working as inteded with getRuntime

TorbjornHoltmon:
### What does this change?

- The cloudflare context was spread. This did not work and waitUntil was not passed on as intended.
( This might also be an issue for other context runtime functions )
- Wait until has to be "proxied" through a function, passing the reference alone will trigger illegal invocations.

## Testing

Tested on our own environment.
Has been tested on our `advanced` mode website.
I cannot speak for `directory` mode, but I assume the problem is the same.

## Docs

No need to update docs,  `waitUntil` should now work as intended.

### Comments

I did not bother to mock waitUntil since we already have a mock In our project.

Its up to the Astro team to decide from this point on. But you could do something like what we do:

```typescript
    waitUntil: (promise: Promise<any>) => {
      if (runtime?.waitUntil) {
        runtime.waitUntil(promise);
      } else {
        Promise.resolve(promise).catch((error) => {
          console.error(error);
          // noop
        });
      }
    },
```

### DIFF
--- a/.changeset/clever-apes-shop.md
+++ b/.changeset/clever-apes-shop.md
@@ -1,5 +1,5 @@
 ---
-'@astrojs/cloudflare': minor
+'@astrojs/cloudflare': patch
 ---
 
 Fixed issue with cloudflare runtime `waitUntil` not working as intended.
--- a/.changeset/clever-apes-shop.md
+++ b/.changeset/clever-apes-shop.md
@@ -2,5 +2,5 @@
 '@astrojs/cloudflare': minor
 ---
 
-Fixed issue with runtime wait until not working as intended.
+Fixed issue with cloudflare runtime `waitUntil` not working as intended.
 
--- a/.changeset/clever-apes-shop.md
+++ b/.changeset/clever-apes-shop.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/cloudflare': minor
+---
+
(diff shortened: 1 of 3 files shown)


packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -17,6 +17,7 @@ export function createExports(manifest: SSRManifest) {
 	}: {
 		request: Request & CFRequest;
 		next: (request: Request) => void;
+		waitUntil: (promise: Promise<any>) => void;
```
ematipico, line 20: Should we use `unknown` instead of `any`?

ematipico: Wow even better! Thank you!

.changeset/clever-apes-shop.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/cloudflare': minor
```
bluwy, line 2: Should this be a patch? as it fixes a bug
```suggestion
'@astrojs/cloudflare': patch
```

TorbjornHoltmon: Yeah, patch is probably more correct.

bluwy (approved):

ematipico (approved):

PR #7424, opened 2023-06-19
Fix up `pnpm dev`

bholmesdev:
## Changes

- **No more "astro" package red squiggles!** 🥳  Avoid deleting `.d.ts` files in dev. Turns out, we had a bad glob where we clean up files. We need `onlyFiles` to avoid deleting nested directories.
- Add `"persistent": true` with boosted concurrency [per turborepo's recommendation](https://turbo.build/repo/docs/handbook/dev#setup-with-turborepo).
- Move `src/content/templates` to the project root. These were getting crawled by `tsc`, which throws wacky build errors if you try to build twice!

## Testing

- Manual testing that `dev` runs and `.d.ts` files are preserved
- Checking that `"persistent": true` = faster dev perf needs extended testing


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "turbo run build --filter=astro --filter=create-astro --filter=\"@astrojs/*\" --filter=\"@benchmark/*\"",
     "build:ci": "turbo run build:ci --filter=astro --filter=create-astro --filter=\"@astrojs/*\" --filter=\"@benchmark/*\"",
     "build:examples": "turbo run build --filter=\"@example/*\"",
-    "dev": "turbo run dev --parallel --filter=astro --filter=create-astro --filter=\"@astrojs/*\" --filter=\"@benchmark/*\"",
+    "dev": "turbo run dev --concurrency=40 --parallel --filter=astro --filter=create-astro --filter=\"@astrojs/*\" --filter=\"@benchmark/*\"",
```
bluwy, line 14: I don't think we need the `--concurreny` here. The [docs](https://turbo.build/repo/docs/reference/command-line-reference/run#--concurrency) mentioned that:

> This option is ignored if the [--parallel](https://turbo.build/repo/docs/reference/command-line-reference/run#--parallel) flag is also passed.

Presumably the `--parallel` flag also made the `"persistent": true` config not needed, but I think it's good to have the latter anyways.

bholmesdev: Hm, I actually hit a build error stating the default `--concurrency` is `10`, and we needed to bump to (# tasks) to get it working. We have 33 CLI tasks which seems to hit the parallel default limit. I chose 40 to give us a bit of room 🤷

bluwy: Strange. Must be outdated docs then. Let's go with this for now!

bluwy (approved):

matthewp (approved):

natemoo-re (approved):

PR #7425, opened 2023-06-20
feat: use typescript-eslint@v6's reworked configs

JoshuaKGoldberg:
## Changes

Posting this as a reference for https://discord.com/channels/830184174198718474/845430950191038464/1120563248782114986:

> 👋 I'm working on testing out reworked preset configs for the next major version of typescript-eslint: https://github.com/typescript-eslint/typescript-eslint/issues/6760. I'd like to try them out on Astro because it's one of my favorite open source projects ❤️. Should I file an issue for this? (I could only find the bug template).
> Benefits of trying this out means I'll help out with the ESLint config in `withastro/astro` & listen even more closely & seriously to your feedback on the new preset configs!

_Edit: the response there was positive, so un-drafting! Thanks!_

Upgraded typescript-eslint to v6, with reworked ESLint configurations.

You can read https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta#user-facing-breaking-changes for the rationale behind the config changes. Essentially, the new recommended starter configs are:

* `"plugin:@typescript-eslint/recommended-type-checked"`
* `"plugin:@typescript-eslint/stylistic-type-checked"`

I've commented any changes I suspect you might disagree with inline, for discussion. Would love to know what you all think!


.eslintrc.cjs
```diff
@@ -1,23 +1,55 @@
...
+    '@typescript-eslint/no-unused-vars': ['error', { ignoreRestSiblings: true }],
+    'no-only-tests/no-only-tests': 'error',
+    '@typescript-eslint/no-shadow': ['error'],
+    'no-console': 'warn',
+
+    /*
+    interface: 404 results in 198 files
+    type: 129 results in 73 files
+    */
+    '@typescript-eslint/consistent-type-definitions': 'off',
```
JoshuaKGoldberg, line 24: It looks like you've _somewhat_ solidified on preferring `interface` over `type`, but both have a good amount of usage. Is using one over the other when possible something you'd want to standardize?

https://typescript-eslint.io/rules/consistent-type-definitions

.eslintrc.cjs
```diff
@@ -1,23 +1,55 @@
...
+    '@typescript-eslint/no-unsafe-call': 'off',
+    '@typescript-eslint/no-unsafe-member-access': 'off',
+    '@typescript-eslint/no-unsafe-return': 'off',
+    '@typescript-eslint/prefer-nullish-coalescing': 'off',
+    '@typescript-eslint/prefer-string-starts-ends-with': 'off',
+    '@typescript-eslint/require-await': 'off',
+    '@typescript-eslint/restrict-plus-operands': 'off',
+    '@typescript-eslint/restrict-template-expressions': 'off',
+    '@typescript-eslint/sort-type-constituents': 'off',
+    '@typescript-eslint/unbound-method': 'off',
```
JoshuaKGoldberg, line 49: Each of these had at least a few existing warnings, so I didn't want to overreach and solve for them without asking. Are any of these rules ones you feel particularly strongly about?

- https://typescript-eslint.io/await-thenable
- https://typescript-eslint.io/ban-ts-comment
- https://typescript-eslint.io/class-literal-property-style
- https://typescript-eslint.io/consistent-indexed-object-style
- https://typescript-eslint.io/dot-notation
- https://typescript-eslint.io/no-base-to-string
- https://typescript-eslint.io/no-confusing-void-expression
- https://typescript-eslint.io/no-empty-function
- https://typescript-eslint.io/no-floating-promises
- https://typescript-eslint.io/no-misused-promises
- https://typescript-eslint.io/no-this-alias
- https://typescript-eslint.io/no-unsafe-argument
- https://typescript-eslint.io/no-unsafe-assignment
- https://typescript-eslint.io/no-unsafe-call
- https://typescript-eslint.io/no-unsafe-member-access
- https://typescript-eslint.io/no-unsafe-return
- https://typescript-eslint.io/prefer-nullish-coalescing
- https://typescript-eslint.io/prefer-string-starts-ends-with
- https://typescript-eslint.io/require-await
- https://typescript-eslint.io/restrict-plus-operands
- https://typescript-eslint.io/restrict-template-expressions
- https://typescript-eslint.io/sort-type-constituents
- https://typescript-eslint.io/unbound-method

If not, I can go ahead and fix for all their complaints.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Build (ignoring failures)
+        run: pnpm run build || true
```
JoshuaKGoldberg, line 53: Explaining: some newly enabled lint rules require type information. I'm under the impression `types` won't be fully filled out (i.e. some packages are seen as `any` / not-existing in editors) until `pnpm run build` happens. Is that the case?

package.json
```diff
@@ -28,7 +28,7 @@
     "test:e2e": "cd packages/astro && pnpm playwright install && pnpm run test:e2e",
     "test:e2e:match": "cd packages/astro && pnpm playwright install && pnpm run test:e2e:match",
     "benchmark": "astro-benchmark",
-    "lint": "eslint --cache .",
+    "lint": "eslint . --report-unused-disable-directives",
```
JoshuaKGoldberg, line 31: Not strictly necessary, but I've found this to be a nice cleanup.

packages/astro/src/@types/astro.ts
```diff
@@ -1087,9 +1087,13 @@ export interface AstroUserConfig {
...
-	integrations?: Array<
-		AstroIntegration | (AstroIntegration | false | undefined | null)[] | false | undefined | null
-	>;
+	integrations?: (
+		| AstroIntegration
+		| (AstroIntegration | false | undefined | null)[]
+		| false
+		| undefined
+		| null
+	)[];
```
JoshuaKGoldberg, line 1096: Switching from `Array<...>` to `(...)[]` is very much a _`stylistic`_ choice (https://typescript-eslint.io/rules/array-type/). It seemed to me like `[]` was generally preferred already?

packages/markdown/remark/src/index.ts
```diff
@@ -160,9 +160,8 @@ function prefixError(err: any, prefix: string) {
 	const wrappedError = new Error(`${prefix}${err ? `: ${err}` : ''}`);
 	try {
 		wrappedError.stack = err.stack;
-		// @ts-expect-error
 		wrappedError.cause = err;
```
JoshuaKGoldberg, line 163: I added `target: ES2022` in the `tsconfig.json` so this line wouldn't need a `// @ts-expect-error`.

packages/astro/src/core/create-vite.ts
```diff
@@ -32,7 +32,7 @@ import { joinPaths } from './path.js';
 interface CreateViteOptions {
 	settings: AstroSettings;
 	logging: LogOptions;
-	mode: 'dev' | 'build' | string;
+	mode: string;
```
bluwy, line 35: Are we able to keep the previous type definition? It _is_ redundant for now since the `string` would widen the type, but we have discussion to circle back in refactoring `mode` in the future, and keeping `'dev' | 'build'` keeps the history here.

package.json
bluwy: I think we'd want to still keep the `--cache` too?

packages/astro/src/@types/astro.ts
bluwy: I usually use `(...)[]` so this is fine by me.

.eslintrc.cjs
bluwy: I usually prefer `interface` over `type`. I think some of us prefer the other way around too. Maybe we can enforce this in a later PR, or if we decide on a style here.

packages/telemetry/src/system-info.ts
```diff
@@ -52,7 +52,8 @@ export function getSystemInfo(versions: { viteVersion: string; astroVersion: str
 	}
 
 	const cpus = os.cpus() || [];
-	meta = {
+
+	return {
```
bluwy, line 56: I think there's a bug here. We want to assign to `meta` so that it is cached on the second run. The new code doesn't assign to it anymore

packages/integrations/image/src/utils/workerPool.ts
```diff
@@ -112,7 +112,7 @@ export default class WorkerPool<I, O> {
 		parentPort!.on('message', async (data) => {
 			const { msg, id } = data;
 			try {
-				const result = await cb(msg);
+				const result = cb(msg);
```
bluwy, line 115: I think we need an `await` here. It's usage at

https://github.com/joshuakgoldberg/astro/blob/e321967a408764407e0e4edd659d955a5f5db054/packages/integrations/image/src/vendor/squoosh/image-pool.ts#L148

returns a promise.

packages/integrations/image/src/index.ts
```diff
@@ -39,7 +39,7 @@ export interface IntegrationOptions {
 	/**
 	 * Entry point for the @type {HostedImageService} or @type {LocalImageService} to be used.
 	 */
-	serviceEntryPoint?: '@astrojs/image/squoosh' | '@astrojs/image/sharp' | string;
+	serviceEntryPoint?: string;
```
bluwy, line 42: I think we'd want to preserve the previous static strings too 🤔

.github/workflows/ci.yml
bluwy: Yeah `pnpm run build` is needed to generate the types. I'd be fine with this since our CI also builds packages. I think this can be refactored further later to share the turbo cache in a later PR.

.eslintrc.cjs
ematipico: Usually, I go the other way around. 

`type` by default, and `interface` if I need to use its features:
- lazy loading
- `extends`


I don't have strong opinions, though.

.eslintrc.cjs
ematipico: > [typescript-eslint.io/await-thenable](https://typescript-eslint.io/await-thenable)

I'll keep it turned off. We have cases where we don't want to `await` a promise on purpose (middleware is one).

> [typescript-eslint.io/no-empty-function](https://typescript-eslint.io/no-empty-function)

Maybe we have a few no-ops in the code base? Not sure. I'd like to add this rule if possible and fix the cases where we trigger the error.

packages/astro/src/assets/vendor/image-size/types/pnm.ts
```diff
@@ -19,7 +19,7 @@ const handlers: { [type: string]: Handler} = {
     let dimensions: string[] = []
 
     while (lines.length > 0) {
-      const line = lines.shift() as string
+      const line = lines.shift()!
```
ematipico, line 22: The bang operator is unsafe; I have a practice of leaving a `// SAFETY: <explanation>` comment where I explain why it's safe to do these kinds of casting.

ematipico (commented):
I noticed that some changes use the bang operator `!`. I consider this very unsafe, and I wonder if we should change our code there or at least leave some comments explaining why it's safe. cc @withastro/maintainers-core

packages/astro/src/assets/vendor/image-size/types/pnm.ts
ematipico: Yeah, both approaches require attention. I believe this is a chore from our side

.eslintrc.cjs
bholmesdev: I'm with Ema on that too. `type` is more ergonomic to write in my opinion, and I like that it avoids interface merging. Loose opinion strongly held :)

.eslintrc.cjs
bholmesdev: > https://typescript-eslint.io/rules/ban-ts-comment

This will cause issues when we reference experimental packages without generated types ([see here](https://github.com/withastro/astro/blob/main/packages/integrations/markdoc/src/index.ts#L16-L17)). I think these comments are rare, but have use cases we talk about during PR reviews. I'd prefer keeping this turned off!

packages/integrations/image/src/index.ts
bholmesdev: `string` will override those static strings in intellisense, so I understand the deletion. I actually _just_ went through an extensive twitter thread on this 😄 

TL;DR: if you want to show suggestions via intellisense, but _still_ allow arbitrary strings, you can [either append `& {}` to the definition or use the uppercase `String` object instead](https://twitter.com/BHolmesDev/status/1670575170027769856):

`serviceEntryPoint?: '@astrojs/image/squoosh' | '@astrojs/image/sharp' | String;`

.eslintrc.cjs
bluwy: I don't have a strong reason picking `interface` over `type`, other than I use them the same way as `function foo() {}` vs `const foo = () => {}`. E.g. `interface` for typing object shapes, `type` for quick one-offs. But I also somewhat like that `interface` generally provides nicer TS intellisense message, especially when unions/extends are used.

matthewp: I think we should avoid doing an upgrade and changing (or adding) lint rules in the same PR. Those are two separate ideas that should be discussed separately.

.eslintrc.cjs
matthewp: Same comment as above, would prefer that we don't change the rules in an upgrade PR.

packages/astro/src/@types/astro.ts
lilnasy: I'd argue against linting for either. `(...)[]` looks neat with smaller types but becomes easy to miss with larger unions and object types.

.eslintrc.cjs
JoshuaKGoldberg: I'll leave this as a todo then - agreed that if there isn't team consensus, this is not a great place to come to one. Thanks!

@bholmesdev looking forward to a whiteboard video on the nuanced differences between the two 😉

package.json
JoshuaKGoldberg: Ehh I've found the cache to be ... unreliable when working with type checking. Theoretically a file's lint result could be impacted by a couple of other factors -other files it imports types from, TSConfig settings- so caching becomes difficult.

packages/astro/src/@types/astro.ts
JoshuaKGoldberg: @lilnasy does https://typescript-eslint.io/rules/array-type/#array-simple match your preference?

packages/telemetry/src/system-info.ts
JoshuaKGoldberg: Ah, this is what I get for hand-editing 😅 good catch. Will undo.

packages/astro/src/core/create-vite.ts
JoshuaKGoldberg: Yup we can! I can see why you'd want this. I'll move https://typescript-eslint.io/rules/no-redundant-type-constituents to the _todo_ pile.

packages/integrations/image/src/utils/workerPool.ts
JoshuaKGoldberg: I'd initially had https://typescript-eslint.io/rules/require-await on the enabled list & fixed up its complaints, then realized you use `await` for places that _might_ give a Promise from middleware (as stated elsewhere in the review). I missed reverting this change. Will revert this and anything else similar I can spot.

packages/integrations/image/src/index.ts
JoshuaKGoldberg: As much as I love working around TypeScript 😄 I'll revert these string changes & move https://typescript-eslint.io/rules/no-redundant-type-constituents/ to the todo pile.

.eslintrc.cjs
JoshuaKGoldberg: > > [typescript-eslint.io/rules/ban-ts-comment](https://typescript-eslint.io/rules/ban-ts-comment)
> 
> This will cause issues when we reference experimental packages without generated types ([see here](https://github.com/withastro/astro/blob/main/packages/integrations/markdoc/src/index.ts?rgh-link-date=2023-06-20T11%3A05%3A25Z#L16-L17)). I think these comments are rare, but have use cases we talk about during PR reviews. I'd prefer keeping this turned off!

Maybe helpful?: it can be configured to allow the comment directives _if_ there's a description attached.

> Same comment as above, would prefer that we don't change the rules in an upgrade PR.

👍 I'll keep these all as _todo_.

packages/astro/src/assets/vendor/image-size/types/pnm.ts
JoshuaKGoldberg: Yup - `!` is precisely the same as `as string` when used on a nullish `string`. All this change is doing is going with a slightly more succinct syntax. https://typescript-eslint.io/rules/non-nullable-type-assertion-style

packages/astro/src/@types/astro.ts
lilnasy: That's perfect!

packages/astro/src/assets/vendor/image-size/types/pnm.ts
ematipico: Thank you. I suppose we can keep the existing changes, then we can enable the lint rule and then we can refactor the errors :) 

The bang operator is easily detectable by an analyzer

packages/astro/src/@types/astro.ts
bluwy: I'm not really sure about `array-simple`. If we're undecided about this, I think we can also leave it unchanged for now (sorry for causing more changes)

package.json
bluwy: I won't deny that I've seen `--cache` acting strange before, so I could be persuaded removing it for now if there's no rejections from others.

.eslintrc.cjs
nickmccurdy: Note that changing from `interface` to `type` can be a breaking change, both due to inability to merge declarations and potential performance regressions.

packages/astro/src/core/create-vite.ts
nickmccurdy: Note that allowing redundant string literals does still improve autocomplete.

packages/astro/src/assets/vendor/image-size/types/pnm.ts
nickmccurdy: Hi, I wrote the original PR to Vite replacing `as` with `!`. The reason for this is that while both can make potentially unsafe assertions, `as` can silence type errors when types are assignable, making `!` relatively more safe.

packages/astro/src/core/create-vite.ts
bholmesdev: @nickmccurdy Hm, doesn't seem to from my experience. Union-ing with a lowercase `string` will blow away other string literals in VS Code. There's a few hacks you can use to get around this like adding `& {}` or `String` to the type.

.eslintrc.cjs
JoshuaKGoldberg: FWIW at this point `type` has been optimized pretty significantly. I can't remember the last time I've seen it be a performance concern in any code (and would love to be pointed to a demonstrable example!). The TS wiki page is out of date.

packages/astro/src/@types/astro.ts
JoshuaKGoldberg: No worries 😄 I'll go through and undo the changes, then merge from `main` & re-request your review. Thanks!

JoshuaKGoldberg: Done. This is a lesson learned the hard way - when there are many complaint from a stylistic rule, leave it out in the proposal. 😄

JoshuaKGoldberg:
> **CI / Lint (pull_request)** Cancelled after 3m

...works on my machine! (as in, `pnpm lint` after `pnpm build` produces no errors. Undrafting 🤷

bluwy (approved):

.eslintrc.cjs
```diff
@@ -1,23 +1,53 @@
 module.exports = {
+  extends: [
```
ematipico, line 2: I wonder, since this is a breaking change, should we try to move to use the new eslint configuration format? https://eslint.org/docs/latest/use/configure/configuration-files-new

ematipico (approved):

PR #7430, opened 2023-06-20
Update safe dependencies

bluwy:
## Changes

Supersedes https://github.com/withastro/astro/pull/7417

Update all non-major dependencies through `pnpm update -r --interactive`, except TypeScript, Zod, and Shiki, which were causing hard-to-fix TypeScript errors. And except `@astrojs/compiler` and `@astrojs/language-server` that are causing check error and slowness.

Some deps also caused TypeScript errors, but there are easily fixable, which is fixed in the 2nd commit.

There's more we can upgraded later which can be checked with `pnpm outdated -r`, but planning to tackle it separately.

## Testing

Existing tests should pass. Also ran them locally.

## Docs

n/a


natemoo-re (approved):
Thanks for digging into this one!

ematipico (approved):

PR #7431, opened 2023-06-20
chore: update bug report template

ematipico:
## Changes

- update the bug report template by adding a new section called "What's the expected result", where the reporter should tell us what they expect. This is useful, because also the expectations are a form of feedback and dicussion
- fix the `config.yml`, the link to the discussions was pointing to the old repo (even though there's a redirect in place)

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/---01-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/---01-bug-report.yml
@@ -56,8 +56,8 @@ body:
   - type: textarea
     id: bug-expectation
     attributes:
-      label: What's the expected result
-      description: Describe what you expected to happen.
+      label: What's the expected result?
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/---01-bug-report.yml
```diff
@@ -53,6 +53,13 @@ body:
       description: A clear and concise description of what the bug is.
     validations:
       required: true
+  - type: textarea
+    id: bug-expectation
+    attributes:
+      label: What's the expected result
+      description: Describe what you expected to happen.
```
bluwy, line 60:
```suggestion
      label: What's the expected result?
      description: Describe what you expect to happen.
```

bluwy (approved):

PR #7432, opened 2023-06-20
feat: `astro info` command

ematipico:
## Changes

Hackaton time!

This PR adds new command called `astro info`; it's a medium meant for:
- users to share quickly their local environment, without typing too much in the bug report template (to update if this PR is accepted)
- maintainers to have a predictable set of information that is standard across projects 

For now, I printed only the info that we usually ask in the PR template, but we can print more :D 

## Testing

Not needed for now

## Docs

Will include the team after feedback


delucis:
lol, why is called `rage`?? 😂

ematipico:
> lol, why is called `rage`?? 😂

It's called `rage` because you usually use this command when everything goes wrong and you're upset! That's why `rage` 😃

natemoo-re:
This is great! Maybe `astro info` is a better command name? But we could definitely keep `astro rage` as an alias 😅

ematipico:
Renamed the command 😢 

carbon(10)


Here's the final result

packages/astro/src/cli/index.ts
```diff
@@ -71,6 +74,62 @@ async function printVersion() {
...
+			logging,
+		});
+		if (userConfig.adapter?.name) {
+			adapter = userConfig.adapter.name;
+		}
+		if (userConfig.integrations) {
+			for (const integration of userConfig.integrations) {
+				if (isIntegration(integration)) {
+					integrations.push(integration.name);
+				}
```
delucis, line 118: I believe this will run into the same bug I just fixed in #7429 — some integrations are `AstroIntegration[]` not `AstroIntegration`. Probably need to update the `isIntegration` helper to assert `is AstroIntegration | AstroIntegration[]` and then act accordingly?

ematipico: Good catch! Last commit should handle that

sarah11918:
Reminder to add docs here! https://docs.astro.build/en/reference/cli-reference/

bluwy (approved):
I like this! I suppose we also want to update the issue template when this gets released.

.changeset/happy-donuts-taste.md
```diff
@@ -0,0 +1,18 @@
+---
+'astro': minor
+---
+
+New command `astro rage`, useful to share debugging information about your current environment
```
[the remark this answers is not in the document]
bluwy: I think the PR has updated to `astro info` instead, unless `rage` is setup as an alias?

ematipico:
I added "Platform" and "Architecture" to the list of information returned

sarah11918:
Just checking up on docs for this! I didn't notice anything PR'd to the CLI reference page.

ematipico:
> Just checking up on docs for this! I didn't notice anything PR'd to the CLI reference page.

Going to create a PR for that

.changeset/happy-donuts-taste.md
```diff
@@ -0,0 +1,20 @@
+---
+'astro': minor
+---
+
+Adds a new command `astro info`, useful for sharing debugging information about your current environment when you need help!
```
sarah11918, line 5:
```suggestion
Environment info in your CLI

Adds a new command `astro info`, useful for sharing debugging information about your current environment when you need help!
```
Just adding something that will become a heading here, since I think this will end up in the blog post draft, and the existing line would be too long.

sarah11918 (approved):

.changeset/happy-donuts-taste.md
sarah11918: After discussing with Chris, I'm not *sure* about updating the changeset here! :sweat_smile: 

This is a good changeset for the changelog, but not for the blogpost. So, I'm not sure whether to do any optimizing here at all, since there's no good choice.

As a default, I'll NOT suggest headings and stuff, only edit for typos etc., and let Matthew battle that when it comes blog post time! In which case, this one is fine as is!

ematipico: No bother :) fee free to merge any suggestion you deem right

PR #7436, opened 2023-06-20
Fix minior issue in the documentation related to port customization

kossidts:
## Changes

Destructure the argument of the function to customize the Astro dev server based on the command run. (https://github.com/withastro/docs/issues/3532)

## Docs

The fix is made in the types file from which the Configuration Reference docs are generated. Hence the fix leads to an update of the documentation.


ElianCodes (approved):
Yup! You're totally right!

ematipico (approved):

PR #3392, opened 2022-05-17
Allow overriding build vite config options

matthewp:
## Changes

- Allows your `vite` config to override options that are used for SSR.
- Fixes #3046

## Testing

- New test added.

## Docs

N/A

PR #7438, opened 2023-06-21
Fix `astro:build:setup` hook `updateConfig` utility

bluwy:
## Changes

While working on some fixes today, I noticed `astro:build:setup` hook's `updateConfig` util doesn't actually update the config.

In contrary, the `astro:config:setup` hook's `updateConfig` util does actually update the astro config. So the above seems like a bug to me, which is fixed here.

## Testing

Added unit test.

## Docs

Looks like it's not documented 🤔 https://docs.astro.build/en/reference/integrations-reference/#astrobuildsetup. The feature is added in https://github.com/withastro/astro/pull/3392

### DIFF
--- a/.changeset/cyan-vans-clap.md
+++ b/.changeset/cyan-vans-clap.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix `astro:build:setup` hook `updateConfig` utility
--- a/.changeset/cyan-vans-clap.md
+++ b/.changeset/cyan-vans-clap.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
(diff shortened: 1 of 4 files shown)


ematipico (approved):
I am surprised we didn't have a test for that 😨

.changeset/cyan-vans-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix `astro:build:setup` hook `updateConfig` utility
```
ematipico, line 5:
```suggestion
Fix `astro:build:setup` hook `updateConfig` utility, there the configuration wasn't correctly updated when the hook was fired.
```

PR #7439, opened 2023-06-21
Format markdown files

bluwy:
## Changes

Format markdown files with prettier

## Testing

n/a

## Docs

n/a


packages/integrations/markdoc/README.md
```diff
@@ -54,14 +54,15 @@ export default defineConfig({
...
 [VS Code](https://code.visualstudio.com/) supports Markdown by default. However, for Markdoc editor support, you may wish to add the following setting in your VSCode config. This ensures authoring Markdoc files provides a Markdown-like editor experience.
 
 ```json title=".vscode/settings.json"
-"files.associations": {
+{
+  "files.associations": {
     "*.mdoc": "markdown"
+  }
 }
 ```
```
bluwy, line 67: Needed to use an object so it's valid JSON for prettier parsing.

packages/integrations/image/README.md
```diff
@@ -333,13 +348,14 @@ See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/size
...
 **Required:** `true`
+
 </p>
 
 The list of sizes that should be built for responsive images. This is combined with `aspectRatio` to calculate the final dimensions of each built image.
 
 ```astro
 // Builds three images: 400x400, 800x800, and 1200x1200
-<Picture src={...} widths={[400, 800, 1200]} aspectRatio="1:1" alt="descriptive text" />
+<Picture src={/* ... */} widths={[400, 800, 1200]} aspectRatio="1:1" alt="descriptive text" />
```
bluwy, line 358: Need a small tweak here so prettier-plugin-astro can parse this.

packages/integrations/sitemap/README.md
```diff
@@ -129,66 +129,69 @@ Sitemap: https://<YOUR SITE>/sitemap-index.xml
...
 
 export default defineConfig({
-  integrations: [sitemap({
-    filter: ...
-  })]
+  integrations: [
+    sitemap({
+      filter, // ...
+    }),
+  ],
```
bluwy, line 143: Need to use a slightly different syntax so it's valid JS. Was also thinking to remove `filter,` entirely and keep `// ...` only

bluwy (commented):
Mostly let prettier run as-is, but made some specific changes below to fix parsing errors.

ematipico (approved):

bluwy:
We're already running prettier in CI so markdown should be included together after this.

packages/integrations/sitemap/README.md
delucis: That would work — or something like `// configuration options` to be clear what it’s a placeholder for?

packages/integrations/vercel/README.md
```diff
@@ -214,26 +221,26 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+     }
+     return Response.redirect(url);
+   }
+   ```
+
 1. While developing locally, you can run `vercel dev` to run middleware. In production, Vercel will handle this for you.
 
-> **Warning**
-> **Trying to rewrite?** Currently rewriting a request with middleware only works for static files.
+> **Warning** > **Trying to rewrite?** Currently rewriting a request with middleware only works for static files.
```
delucis, line 243: This looks wrong? Should be across two lines. Didn’t happen to other places we use the `> **Warning**` or `> **Note**` syntax, nor sure what the issue here was.

packages/markdown/component/readme.md
```diff
@@ -3,15 +3,14 @@
 This package brings legacy support for the `<Markdown />` component to all Astro projects.
 
 > The `<Markdown />` component does not work in SSR. Consider [importing Markdown content](https://docs.astro.build/en/guides/markdown-content/#importing-markdown) instead.
-:::
+> :::
```
delucis, line 6: This `:::` looks like it got copied over from our docs syntax at some point, and should probably be removed:

```suggestion
```

packages/markdown/component/readme.md
```diff
@@ -3,15 +3,14 @@
 This package brings legacy support for the `<Markdown />` component to all Astro projects.
 
 > The `<Markdown />` component does not work in SSR. Consider [importing Markdown content](https://docs.astro.build/en/guides/markdown-content/#importing-markdown) instead.
```
delucis, line 5: While we’re here we can probably make this a [warning](https://github.com/orgs/community/discussions/16925) instead of a plain blockquote too:

```suggestion
> **Warning**
> The `<Markdown />` component does not work in SSR. Consider [importing Markdown content](https://docs.astro.build/en/guides/markdown-content/#importing-markdown) instead.
```

packages/webapi/README.md
```diff
@@ -108,48 +108,46 @@ The `exclude` option receives a list of WebAPIs to exclude from polyfilling.
...
 ```
 
-| Shorthand      | Excludes |
-|:-------------- |:-------- |
-| `Document+`    | `Document`, `HTMLDocument` |
-| `Element+`     | `Element`, and exclusions from `HTMLElement+` |
-| `Event+`       | `Event`, `CustomEvent`, `EventTarget`, `MediaQueryList`, `Window`, and exclusions from `Node+` |
-| `EventTarget+` | `Event`, `CustomEvent`, `EventTarget`, `MediaQueryList`, `Window`, and exclusions from `Node+` |
+| Shorthand      | Excludes                                                                                                                                                                                                                              |
+| :------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
```
delucis, line 140: Not sure this is _prettier_ but I guess we can roll with it 😅

delucis (commented):
Thanks @bluwy!

Generally we’ve avoided auto-formatting content in the docs repo because sometimes it’s nice to have full control of docs example code and there are some weird pain points I’ve found when formatting MDX (plain Markdown like this seems to do better).

I had a quick read through and this mostly LGTM! Left a couple of comments for spots that may need fixing.

packages/integrations/image/README.md
delucis: Could we just pass some variable here?

```suggestion
<Picture src={img} widths={[400, 800, 1200]} aspectRatio="1:1" alt="descriptive text" />
```

packages/integrations/vercel/README.md
bluwy: Good catch! Looks like a prettier bug when the second line starts with a bold word

bluwy:
> Generally we’ve avoided auto-formatting content in the docs repo because sometimes it’s nice to have full control of docs example code and there are some weird pain points I’ve found when formatting MDX (plain Markdown like this seems to do better).

I think the formatting benefit outweighs some of the caveats, which should be fixable with a `<!-- prettier-ignore -->`. But won't force this everywhere 😄

Yan-Thomas (approved):
Everything is looking fine to me!

PR #7445, opened 2023-06-22
Fix an error message for use in docs

delucis:
## Changes

- Uses a single template string instead of multiple joined together with `+`
- Docs doesn’t support parsing concatenated strings resulting in the following Markdown:
   ```md
   > **PrerenderDynamicEndpointPathCollide**: Could not render `PATHNAME` with an `undefined` param as the generated path will collide during prerendering. ` +
            `Prevent passing `undefined` as `params` for the endpoint's `getStaticPaths()` function, ` +
            `or add an additional extension to the endpoint's filename. (E03026)
            `or add an additional extension to the endpoint's filename.
   ```

## Testing

Existing tests should pass.

## Docs

Docs-fix only.


sarah11918 (approved):
Thanks @delucis!

PR #7446, opened 2023-06-22
add new directive `client:load="dom"` to support SSI rendering with delay

don41382:
The PR adds a new directive option for `client:load` -> `client:load="dom"`. This waits with the hydration process until the DOM is fully loaded.

When using SSI includes via Nginx content might be delivered delayed. An eager approach, like `client:load` breaks the Svelte hydration of the page, which results in duplicated elements.

I included an example in the example folder.

## Changes

Adds a new directive `client:load="dom"`, which starts hydration after dom load

## Testing

I am not sure how to test this, because it requires a nginx to be running. Ideas are welcome!

## Docs
see https://github.com/withastro/docs/pull/3546


don41382:
Do I need an extra issue to fix the broken checks? I don't understand why `check (pull request)` and `Bundle Size (pull request)` are failing.

Thanks for your help.

natemoo-re:
> Do I need an extra issue to fix the broken checks? I don't understand why check (pull request) and Bundle Size (pull request) are failing.

I think those actions are both failing because they can't run on forks? Just a guess, but either way it's not your fault. Don't worry about those checks!

We'll have to discuss this implementation as a team. I could definitely see an argument for waiting for `DOMContentLoaded` being the default behavior.

don41382:
@natemoo-re Thanks for the update and let me know, if I can support you.

don41382:
Any update on this?

natemoo-re:
Hey thanks for your patience, @don41382! I'm sorry for the delay.

Unfortunately, after discussing with the team, we've decided to close this PR for now. We're not opposed to the idea, but we'd love to discuss it through our roadmap process so we can consider all the design implications. Since we support a number of other directives, we want to make sure the APIs are considered holistically and don't diverge in their API surface.

Please [open a discussion](https://github.com/withastro/roadmap/discussions) and we can continue the conversation there.

In the meantime, I will point out our [`addClientDirective`](https://docs.astro.build/en/reference/integrations-reference/#addclientdirective-option) API which allows you to fully customize any hydration behavior. Hopefully that unblocks you for now.

Issue #7444, opened 2023-06-22
`trailingSlash: "always"` break on Vercel


anasrar:
### What version of `astro` are you using?

2.6.6

### Are you using an SSR adapter? If so, which one?

@astrojs/vercel@3.5.0

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

when `trailingSlash: "always"` is presented in config, access `/` will continuously redirect to `/`. Resulting infinite redirect when deployed on Vercel.

### What's the expected result?

See basics template on `/`.

### Link to Minimal Reproducible Example

https://github.com/anasrar/astro-vercel-trailingslash

### Participation

- [ ] I am willing to submit a pull request for this issue.

rishi-raj-jain:
Are you sure about this?
https://github.com/rishi-raj-jain/rishi.app/blob/main/src/routes/%2Blayout.ts doesn't break on rishi.app.

rishi-raj-jain:
That was SvelteKit, hence deleted. I love both the frameworks.

artt:
@bluwy seems like this is still an issue? What do you use for your setup now?
https://github.com/withastro/adapters/issues/418

PR #7447, opened 2023-06-22
Fix vercel redirects with trailingSlash always

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7444

It was generating redirects from `/` to `//`, which was causing infinite redirects. I've removed creating redirects for the root page, which should fix this.

I don't see `@astrojs/underscore-redirects` creating redirects based on `trailingSlash` so they might be unaffected.

## Testing

Added test in vercel

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/nervous-paws-knock.md
+++ b/.changeset/nervous-paws-knock.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': patch
+---
+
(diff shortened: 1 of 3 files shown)


ematipico (approved):
Let's 🚢 it

PR #7448, opened 2023-06-22
feat: build option `splitMiddleware`

ematipico:
## Changes

This PR adds a new build option called `build.splitMiddleware`. This is an advanced option needed for integrations, and it will be used later on in the Vercel adapter.

## Testing

I created a small test case, asserting that the locals won't be rendered because the middleware it's not called.

## Docs

If the team agrees, I'd like to skip the doc review for now, and review it only when the final code is ready to be merged on `main`

<-- /cc @withastro/maintainers-docs for feedback! -->


matthewp:
Happy to see this was maybe a smaller change. But it looks like from the code that if `opts.settings.config.build.splitMiddleware` is true, then middleware will not be built because it's not part of the module graph. Doesn't it need to become a top-level `input`? Or no?

ematipico:
>  Doesn't it need to become a top-level input? Or no?

It seems there's no need, the code emitted will look like this:

```js
import { e as sequence, f as defineMiddleware } from './chunks/astro.52e57b1a.mjs';

const third = defineMiddleware(async (context, next) => {
	if (context.request.url.includes('/broken-locals')) {
		context.locals = {
			fn() {},
		};
	} else if (context.request.url.includes('/does-nothing')) {
		return undefined;
	}
	return next();
});

const onRequest = sequence(first, second, third);

export { onRequest };
```

And it won't be imported by any page

packages/astro/src/core/build/plugins/plugin-middleware.ts
```diff
@@ -39,6 +39,13 @@ export function vitePluginMiddleware(
 		load(id) {
 			if (id === EMPTY_MIDDLEWARE) {
 				return 'export const onRequest = undefined';
+			} else if (id === MIDDLEWARE_MODULE_ID) {
+				this.emitFile({
+					type: 'chunk',
+					preserveSignature: 'strict',
+					fileName: 'middleware.mjs',
+					id,
+				});
```
bluwy, line 48: Noicee

packages/astro/src/core/build/index.ts
```diff
@@ -226,6 +226,25 @@ class AstroBuilder {
...
+			}
+		}
+		if (config.build.splitMiddleware === true) {
+			if (config.output === 'server') {
+				warn(
+					this.logging,
+					'configuration',
+					'The option `build.splitMiddleware` won\'t take effect, because `output` is not `"server"`.'
+				);
+			}
```
bluwy, line 246: Are the `config.output === 'server'` checks correct? Should it be `config.output === 'client'` instead since hybrid mode should also work too?

.changeset/strong-years-travel.md
```diff
@@ -0,0 +1,20 @@
+---
+'astro': minor
+---
+
+Introduced a new build option for SSR, called `build.splitMiddleware`.
```
bluwy, line 5: I'm a little concerned if `splitMiddleware` is the right term here. We're already splitting the middleware entrypoint out like how `build.split` handles the other entrypoints.

What about `build.handleMiddleware` / `build.callMiddleware` that are boolean `true` by default. Or just `build.middleware` but that's a bit bold.

In Astro 3, maybe it's good to revisit how adapters configure the build options, because `build.*` options are visible to users too and they likely don't want to configure this themselves.

packages/astro/src/core/build/index.ts
ematipico: Ah yeah, good catch!

.changeset/strong-years-travel.md
ematipico: Yeah, the name can be reviewed. This is a low-level configuration that I wouldn't want to expose to the user. I am not sure it's possible to do this now. 

What about `includeMiddleware` or `excludeMiddleware`?

bluwy: `excludeMiddleware` sounds good to me! I like it's an option that's `false` by default

packages/astro/src/core/config/schema.ts
```diff
@@ -25,6 +25,7 @@ const ASTRO_CONFIG_DEFAULTS: AstroUserConfig & any = {
 		redirects: true,
 		inlineStylesheets: 'never',
 		split: false,
+		excludeMiddlwar: false,
```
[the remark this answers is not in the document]
ematipico: Thank you, will fix in a next PR :)

PR #7454, opened 2023-06-23
Svelte Four

ghostdevv:
## Changes

Adds support for svelte 4 in `@astrojs/svelte`~~, we need support for this in language tools as well before it can be merged, so I created this issue https://github.com/withastro/language-tools/issues/589~~

## Testing

Tested locally

## Docs

n/a


bluwy:
I don't think this needs to be blocked if language-tools isn't ready. IIUC we only need to loosen the peer dep range.

ghostdevv:
I think these are all the changes needed then?

bluwy:
Yup 👍

bluwy (approved):

PR #7458, opened 2023-06-23
feat: expose middleware URL to integrations

ematipico:
## Changes

This is another feature towards the support of Vercel Edge Middleware.

This feature exposes the file path of the middleware file emitted during the astro build.

## Testing

I extended the functionality of the test adapter and created a test that makes sure that the file exists and it has content.

## Docs

I will document the new payload towards the end of the feature.

### DIFF
--- a/.changeset/chilly-pants-fix.md
+++ b/.changeset/chilly-pants-fix.md
@@ -11,8 +11,8 @@ function integration(): AstroIntegration {
     return {
         name: "fancy-astro-integration",
         hooks: {
-            'astro:build:done': ({ middlewarePath }) => { 
-                if (middlewarePath) {
+            'astro:build:done': ({ middlewareEntryPoint }) => { 
+                if (middlewareEntryPoint) {
                     // do some operations
                 }
             }
@@ -21,4 +21,4 @@ function integration(): AstroIntegration {
 }
 ```
 
-The `middlewarePath` is only defined if the user has created an Astro middleware.
+The `middlewareEntryPoint` is only defined if the user has created an Astro middleware.
--- a/.changeset/chilly-pants-fix.md
+++ b/.changeset/chilly-pants-fix.md
@@ -1,0 +1,24 @@
+---
+'astro': minor
+---
+
+Astro exposes the middleware file path to the integrations in the hook `astro:build:done`
+
+```ts
+// myIntegration.js
+import type { AstroIntegration } from 'astro';
+function integration(): AstroIntegration {
+    return {
+        name: "fancy-astro-integration",
+        hooks: {
+            'astro:build:done': ({ middlewarePath }) => { 
+                if (middlewarePath) {
+                    // do some operations
+                }
+            }
+        }
+    }
+}
+```
+
+The `middlewarePath` is only defined if the user has created an Astro middleware.
--- a/packages/astro/src/core/build/internal.ts
+++ b/packages/astro/src/core/build/internal.ts
@@ -84,7 +84,6 @@ export interface BuildInternals {
 	entryPoints: Map<RouteData, URL>;
 	ssrSplitEntryChunks: Map<string, Rollup.OutputChunk>;
 	componentMetadata: SSRResult['componentMetadata'];
-
 	middlewareEntryPoint?: URL;
 }
 
--- a/packages/astro/src/core/build/internal.ts
+++ b/packages/astro/src/core/build/internal.ts
@@ -84,6 +84,8 @@ export interface BuildInternals {
 	entryPoints: Map<RouteData, URL>;
 	ssrSplitEntryChunks: Map<string, Rollup.OutputChunk>;
 	componentMetadata: SSRResult['componentMetadata'];
+
+	middlewareEntryPoint?: URL;
 }
 
 /**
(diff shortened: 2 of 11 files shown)


packages/astro/src/core/build/plugins/plugin-middleware.ts
```diff
@@ -41,6 +41,17 @@ export function vitePluginMiddleware(
 				return 'export const onRequest = undefined';
 			}
 		},
+
+		writeBundle(_, bundle) {
+			for (const [chunkName, chunk] of Object.entries(bundle)) {
+				if (chunk.type === 'asset') {
+					continue;
+				}
+				if (chunk.fileName === 'middleware.mjs') {
```
ematipico, line 50: I will put this string in shared variable in the following PRs

packages/astro/src/core/build/internal.ts
```diff
@@ -84,6 +84,8 @@ export interface BuildInternals {
 	entryPoints: Map<RouteData, URL>;
 	ssrSplitEntryChunks: Map<string, Rollup.OutputChunk>;
 	componentMetadata: SSRResult['componentMetadata'];
+
+	middlewareEntryPoint?: URL;
```
ElianCodes, line 88: removed extra whitespace

```suggestion
	middlewareEntryPoint?: URL;
```

ElianCodes (commented):
Awesome! Just left a small formatting nit

packages/integrations/vercel/test/edge-middleware.test.js
```diff
@@ -10,9 +10,4 @@ describe('Serverless prerender', () => {
 			root: './fixtures/middleware/',
 		});
 	});
-
-	it('build successful', async () => {
-		await fixture.build();
-		expect(await fixture.readFile('../.vercel/output/static/index.html')).to.be.ok;
-	});
```
bluwy, line 17: Not sure if removing this test is intentional. It looks like this test file is added by a wip commit from the base branch.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -109,9 +109,6 @@ export default function vercelServerless({
 				// Remove temporary folder
 				await removeDir(buildTempFolder);
 
-				// Write middleware
-				await writeFile(new URL('./middleware.js', functionFolder), generateMiddlewareCode());
-
```
bluwy, line 114: Assuming these stuff were WIP too?

packages/astro/test/middleware.test.js
```diff
@@ -201,6 +210,19 @@ describe('Middleware API in PROD mode, SSR', () => {
 		const text = await response.text();
 		expect(text.includes('REDACTED')).to.be.true;
 	});
+
+	it('the integration should receive the path to the middleware', async () => {
+		expect(middlewarePath).to.not.be.undefined;
+		try {
+			const path = fileURLToPath(middlewarePath);
+			console.log(path);
```
bluwy, line 218: There's a `console.log` here.

.changeset/chilly-pants-fix.md
```diff
@@ -0,0 +1,24 @@
...
+import type { AstroIntegration } from 'astro';
+function integration(): AstroIntegration {
+    return {
+        name: "fancy-astro-integration",
+        hooks: {
+            'astro:build:done': ({ middlewarePath }) => { 
+                if (middlewarePath) {
+                    // do some operations
+                }
+            }
```
bluwy, line 18: Wonder if it makes sense to name this `middlewareEntryPoint` instead? So it's consistent with `entryPoints`

packages/astro/src/core/build/plugins/plugin-middleware.ts
```diff
@@ -41,6 +41,17 @@ export function vitePluginMiddleware(
...
+					internals.middlewareEntryPoint = new URL(chunkName, opts.settings.config.outDir);
```
bluwy, line 51: Should we combine `outDir` + `"server"` + `chunkName` here so we don't have to handle constructing the final `middlewareEntryPoint` again in`core/build/index.ts`? Otherwise this URL path would also point to a non existent path.

ematipico: Technically, at this point of the execution, the path **is correct**. The files are moved **later**.  If we decided to prefix `"server"` here, we would create a `URL` that point to a file that doesn't exist. If, in the future, we decide to expose this information to a hook that is called before the files are moved, we could incur an error.

packages/integrations/vercel/test/edge-middleware.test.js
ematipico: Yes it's intentional. I had pushed this code in the destination branch when I was testing stuff. This test doesn't do anything at the moment and the assertion always fails

packages/astro/src/core/build/plugins/plugin-middleware.ts
bluwy: That's fair 👍 I'm mostly concerned that the final constructed path takes a lot of work to do so, which might not be great for perf. I won't let this block the PR though, I'm also fine with this for now.

bluwy (approved):

bluwy:
Looks like there's a CI fail, but the direction is good for me.

PR #7459, opened 2023-06-23
Fix markdoc test fail in main

bluwy:
## Changes

Fix a small regression from https://github.com/withastro/astro/pull/7381, where virtual modules are not properly detected as `importedModules` as they are represented as `\0astro:content`, but the URLs are `/@id/__x00__astro:content`.

I fixed the issue by inverting the `/@id/` handling to the URLs, so that we can compare directly.

I made refactors along the way, the 3 commits each show step-by-step changes. The 3rd commit is the actual fix.

No changeset as it's fixing an unreleased bug.

## Testing

Existing test should pass

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/cool-birds-prove.md
+++ b/.changeset/cool-birds-prove.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix missing styles for Markdoc files in development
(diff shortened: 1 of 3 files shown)


bholmesdev (approved):
Ah ha, was _just_ about to ask about this! Thanks for the quick fix. No stopping (point) you 👍

bholmesdev:
Needs changeset though!

bluwy:
I noted in the description that we don't need a changeset though 😅

PR #7467, opened 2023-06-23
feat: restart server on markdoc config change

bholmesdev:
## Changes

- Update `server:setup` to fire `restart()` whenever a support Markdoc config type has a change.

## Testing

Manually test that server restarts

## Docs

N/A

### DIFF
--- a/packages/integrations/markdoc/src/index.ts
+++ b/packages/integrations/markdoc/src/index.ts
@@ -4,7 +4,7 @@ import Markdoc from '@markdoc/markdoc';
 import type { AstroConfig, AstroIntegration, ContentEntryType, HookParameters } from 'astro';
 import crypto from 'node:crypto';
 import fs from 'node:fs';
-import { fileURLToPath, pathToFileURL } from 'node:url';
+import { fileURLToPath } from 'node:url';
 import {
 	hasContentFlag,
 	isValidUrl,
@@ -15,11 +15,16 @@ import {
 } from './utils.js';
 // @ts-expect-error Cannot find module 'astro/assets' or its corresponding type declarations.
 import { emitESMImage } from 'astro/assets';
-import { bold, red, yellow } from 'kleur/colors';
+import { bold, red } from 'kleur/colors';
+import vitePluginRestart from 'vite-plugin-restart';
 import path from 'node:path';
 import type * as rollup from 'rollup';
 import { normalizePath } from 'vite';
-import { loadMarkdocConfig, type MarkdocConfigResult } from './load-config.js';
+import {
+	loadMarkdocConfig,
+	SUPPORTED_MARKDOC_CONFIG_FILES,
+	type MarkdocConfigResult,
+} from './load-config.js';
 import { setupConfig } from './runtime.js';
 
 type SetupHookParams = HookParameters<'astro:config:setup'> & {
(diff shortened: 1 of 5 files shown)


packages/integrations/markdoc/src/index.ts
```diff
@@ -204,15 +209,16 @@ export const Content = createComponent({
 
 				updateConfig({
 					vite: {
-						vite: {
-							ssr: {
```
bholmesdev, line 208: How long has that been there... :houston_panic:

packages/integrations/markdoc/src/index.ts
```diff
@@ -231,17 +237,6 @@ export const Content = createComponent({
...
 			},
-			'astro:server:setup': async ({ server }) => {
-				server.watcher.on('all', (event, entry) => {
-					if (prependForwardSlash(pathToFileURL(entry).pathname) === markdocConfigResultId) {
-						console.log(
-							yellow(
-								`${bold('[Markdoc]')} Restart the dev server for config changes to take effect.`
-							)
-						);
-					}
```
bluwy, line 242: I think what we can do is directly restart it here with `server.restart()`. It's also what `vite-plugin-restart` [does under the hood](https://github.com/antfu/vite-plugin-restart/blob/43461b5f362e31944502cefdc45ef9c1f3ebf9eb/src/index.ts#L90).

Before we had our custom container restart strategy, I wanted to make integrations call `server.restart()` too, but I was a bit too late. So if we could start here that would be great 😄

bholmesdev: Ah, I'll see if we can simplify to that instead!

bholmesdev: Update: done. Works great 🏆

packages/integrations/markdoc/src/index.ts
```diff
@@ -233,12 +233,11 @@ export const Content = createComponent({
...
 				server.watcher.on('all', (event, entry) => {
-					if (prependForwardSlash(pathToFileURL(entry).pathname) === markdocConfigResultId) {
-						console.log(
-							yellow(
-								`${bold('[Markdoc]')} Restart the dev server for config changes to take effect.`
-							)
-						);
+					for (const markdocConfigBase of SUPPORTED_MARKDOC_CONFIG_FILES) {
+						const absolutePath = new URL(markdocConfigBase, astroConfig.root);
+						if (entry === fileURLToPath(absolutePath)) {
```
bholmesdev, line 238: I assume this _will_ be the file path and not a Vite ID. Let me know if that's wrong!

bluwy, line 238: Not blocking the PR, but perhaps this can be optimized as 

```js
if (SUPPORTED_MARKDOC_CONFIG_FILES.some(f => entry.endsWith(s))) {
  server.restart()
}
```

(probably good enough) as this function would be triggered repeatedly for any file changes.

bluwy (approved):

packages/integrations/markdoc/src/index.ts
bholmesdev: Ah, was needlessly worried about a markdoc config located in the wrong place. This is probably fine!

ematipico (approved):

packages/integrations/markdoc/src/index.ts
bluwy: Nice!

PR #7468, opened 2023-06-23
feat: New Markdoc `render` API

### LINKED ISSUE #7317: @astrojs/markdoc v0.3.1 (reprise of issue #7170): Markdoc tag rendering Astro component breaks when component uses getEntryBySlug()

2.6.0

None

pnpm

Mac

Chrome, Safari

### Describe the Bug

This is a reprise of issue #7170, which shows up again with @astrojs/markdoc v0.3.1. Details basically identical to the previous issue, except that this time around, local Astro in dev mode doesn't seem to throw an error. In Stackblitz, it does.

This is a very particular bug that showed up with @astrojs/markdoc v0.3.1 , when all of the following are true:

1. @astrojs/markdoc is v0.3.1
2. Pages are served from a content collection of _at least two_ .mdoc files.
3. `markdoc.config.mjs` imports an Astro component (for the purpose of defining a tag, although no tag actually needs to be defined—or used in any .mdoc file)
4. The Astro component uses `getEntryBySlug()` to access data from another content collection (the new `getEntry()` also produces the error)

In local dev mode, I'm not seeing an error, the page just doesn't render. In Stackblitz looks like:

```
  error   Unexpected error while rendering pages → index.mdoc.
Error
    at AstroError (<anonymous>)
    at <anonymous> (<anonymous>)
    at render (<anonymous>)
```

Pages render as expected after downgrading to v0.3.0, or when any of the other items above are false!

https://stackblitz.com/edit/github-pq3pgo-xieer6?file=package.json

### Participation

bholmesdev:
## Changes

https://github.com/withastro/astro/assets/51384119/cda76802-80e1-45ea-adae-56b90aeec95d

Introduce a new API (yes, another one) for adding components to your Markdoc config. Instead of importing components directly, you can use the `component()` utility to reference components by pathname:

```ts
import { component, defineMarkdocConfig } from '@astrojs/markdoc/config';

  tags: {
    aside: {
      // For local files
      render: component('./src/components/MyAside.astro'),
    },
    tabs: {
      // For npm packages with named exports
      render: component('@astrojs/starlight/components', 'Tabs'),
    }
  }
});
```

### Why the change?

- Allows the Markdoc config to be read directly without esbuild hacks to "stub out" `.astro` component imports. This should unlock [Markdoc's language server](https://github.com/markdoc/language-server) as well.
- Allow Astro components from `npm` packages and `index` files. We used a naive assumption before to stub `.astro` imports by file extension. Without this requirement, you can now use Astro components from anywhere!
- Resolves #7317 - component imports are now inlined on the `.mdoc` module itself, instead of using a propagated assets flag, which removes Markdoc-specific content collection bugs. This removes our complex `resolveId()` code and leans on the same propagated assets logic as Markdown and MDX.

## Testing

- Update test fixtures to use new config format.

## Docs

- Update Markdoc README


bholmesdev:
!preview markdoc-component-config

bluwy:
Setting a very high bar for PR vids here 😲 

The changes looks good to me, but I'm curious if we're fine with losing autocompletion for the component imports. I'm leaning towards being fine here as it's lesser bundling and magic. Also, are we able to remove the esbuild bundling entirely now?

Side note: I noticed a pattern where external components tend to make rendering slower. MDX has a [new option](https://docs.astro.build/en/guides/integrations-guide/mdx/#customcomponentnames) similar to markdoc now likely for the same reason.

bholmesdev:
@bluwy 

> are we able to remove the esbuild bundling entirely now?

We still need it to process `.ts` configs to evaluate at build time, so not quite! We could use Vite's config loader instead of maintaining our own though 👀 

> I'm curious if we're fine with losing autocompletion for the component imports

I was working on a code-gen solution to glob any `.astro` files in `src` and offer autocomplete for those. This would actually be _better_ than our existing solution, because importing `.astro` files in a `.ts` file raises lint errors when you're outside of `src`. I removed it as out-of-scope so I could ship something though 🙃 

> Side note: I noticed a pattern where external components tend to make rendering slower. MDX has a [new option](https://docs.astro.build/en/guides/integrations-guide/mdx/#customcomponentnames) similar to markdoc now likely for the same reason.

This is interesting... but I'm not sure it applies to Markdoc? Markdoc doesn't "compile" components at all. We instead ship a tree of nodes to the runtime, and recurse through to find and render any components found. I'd assume this works the same regardless or where the component came from 🤷

sarah11918 (approved):
Looks great, @bholmesdev! Just a tiny nit and maybe a more explicit migration instruction to consider!

Yan-Thomas (approved):
Docs LGTM!

PR #7477, opened 2023-06-25
Bump dependency versions

TheOtterlord:
## Changes

- Bumps `semver` and `fast-xml-parser` dependencies for `astro` and `@astrojs/rss`.

## Testing

- Ran tests

## Docs

N/A


bluwy (approved):
Deleted the changesets as we don't usually add that for dep updates, and the semver ranges before also includes the one we're updating too. But also I believe changesets handles reporting deps update by default too.

PR #7487, opened 2023-06-26
Remove invalid code block language tag in sitemap README

delucis:
## Changes

- Removes use of `txt` as a language tag in a code block, which is [not valid](https://github.com/shikijs/shiki/blob/main/docs/languages.md#all-languages) and causes error logging in docs.
- Had to use a `<!-- prettier-ignore -->` comment because otherwise Prettier tries to remove the whitespace between the opening backticks and the metadata docs uses for insertion highlighting and filename.
- No changeset because in non-docs contexts this change is essentially invisible.

## Testing

Docs-only, n/a.

## Docs

### DIFF
--- a/packages/integrations/sitemap/README.md
+++ b/packages/integrations/sitemap/README.md
@@ -91,7 +91,8 @@ After verifying that the sitemaps are built, you can add them to your site's `<h
 </head>
 ```
 
-```txt ins={4} title="public/robots.txt"
+<!-- prettier-ignore -->
+``` ins={4} title="public/robots.txt"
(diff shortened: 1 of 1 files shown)


ElianCodes (approved):
Weird that prettier removes the whitespace, but this looks like a valid workaround to me!

delucis:
Thanks for reviewing!

I’m guessing Prettier’s assumption is that the whitespace is a typo between the backticks and the language tag (because our syntax is admittedly nonstandard) and thinks it’s fixing that?

PR #7488, opened 2023-06-26
Pass `compressHTML` settings to manifest

### LINKED ISSUE #7333: HTML Minification

2.6.1

Node

npm

macOS

Firefox

### Describe the Bug

In 2.6.0 update which has been patched with 
https://github.com/withastro/astro/pull/7242

I found the `<doctype>`  still contains newline while other part are minified, did the patch solved the issue?

https://stackblitz.com/edit/github-wjgoff?file=astro.config.mjs

### Participation

natemoo-re:
## Changes

- Resolves #7333
- `compressHTML` setting was not passed through to adapters because we weren't saving it in the manifest
- Now the manifest contains `compressHTML` and the base `App` will pass this setting through to `renderPage`

## Testing

Tested manually, but will add a test

## Docs

N/A, bug fix only


matthewp (approved):

ematipico (changes requested):
Isn't this, technically, a minor? We're adding a new property to the manifest, which is passed down to the integrations.

I think we should add a test, blocking before merging.

bluwy:
I think I can see if more as a bug fix as we expect HTML compression to work everywhere, but it didn't handle the manifest part. The new property is an implementation detail of the bug fix. It would be nice to have a test though.

ematipico (approved):

PR #7490, opened 2023-06-27
fix: correct URL for entry points

ematipico:
## Changes

There were two issues:
- the test case was using `for..in` instead of `for..of`, because of that, the test was running but the for loop wasn't executed. Fixing that showed the issue and it was failing;
- fixed the path of `entryPoints`, which now uses the correct `build.server` URL
- we forgot to document the introduction of `entryPoints` in a changeset; we can't create a new changeset now because the feature is already released. I changed `CHANGELOG.md`. Once accepted, I will change the text of https://github.com/withastro/astro/releases/tag/astro%402.7.0

## Testing

Fixed the current test case

## Docs

### DIFF
--- a/packages/astro/CHANGELOG.md
+++ b/packages/astro/CHANGELOG.md
@@ -35,9 +35,8 @@
     },
   });
   ```
-- [#7220](https://github.com/withastro/astro/pull/7220) [`459b5bd05`](https://github.com/withastro/astro/commit/459b5bd05f562238f7250520efe3cf0fa156bb45) Thanks [@ematipico](https://github.com/ematipico)! - Shipped a new SSR build configuration mode: `split`.
 
-  The Astro hook `astro:build:ssr` now receives a new option in their payload, called `entryPoints`.
+- [#7220](https://github.com/withastro/astro/pull/7220) [`459b5bd05`](https://github.com/withastro/astro/commit/459b5bd05f562238f7250520efe3cf0fa156bb45) Thanks [@ematipico](https://github.com/ematipico)! - The Astro hook `astro:build:ssr` now receives a new option in their payload, called `entryPoints`.
 
   `entryPoints` is defined as a `Map<RouteData, URL>`, where `RouteData` represents the information of a Astro route and `URL` is the path to the physical file emitted at the end of the build.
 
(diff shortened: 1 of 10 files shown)


ematipico:
cc @alexanderniebuhr

packages/astro/src/@types/astro.ts
```diff
@@ -111,21 +111,6 @@ export interface CLIFlags {
...
-	/**
-	 * @deprecated Use config.build.server instead.
-	 */
-	server: URL;
-	/**
-	 * @deprecated Use config.build.serverEntry instead.
-	 */
-	serverEntry: string;
-}
-
```
bluwy, line 128: Love this refactor!

packages/astro/CHANGELOG.md
```diff
@@ -35,6 +35,24 @@
     },
   });
   ```
+- [#7220](https://github.com/withastro/astro/pull/7220) [`459b5bd05`](https://github.com/withastro/astro/commit/459b5bd05f562238f7250520efe3cf0fa156bb45) Thanks [@ematipico](https://github.com/ematipico)! - Shipped a new SSR build configuration mode: `split`.
+
+  The Astro hook `astro:build:ssr` now receives a new option in their payload, called `entryPoints`.
```
bluwy, line 40: Tiny nit following how changesets usually format these:
```suggestion

- [#7220](https://github.com/withastro/astro/pull/7220) [`459b5bd05`](https://github.com/withastro/astro/commit/459b5bd05f562238f7250520efe3cf0fa156bb45) Thanks [@ematipico](https://github.com/ematipico)! - The Astro hook `astro:build:ssr` now receives a new option in their payload, called `entryPoints`.
```

1. Prepend a new line
2. Made `The Astro hook ...` the first changesets message instead.

bluwy (approved):
One small nit, but otherwise looks great!

NOTE: The changelog edit sounds fine to me. IIRC you can do that with changesets. Editing the release later should also be fine too.

packages/astro/src/@types/astro.ts
ematipico: Yeah it was long overdue

PR #7492, opened 2023-06-27
feat: bundle the Vercel Edge Middleware function

ematipico:
## Changes

This PR adds logic to create the real  Vercel Edge Middleware file. Check the [documentation](https://vercel.com/docs/concepts/functions/edge-middleware#create-edge-middleware).

The content of the function will be something like this (this is pseudo code, but almost like the real deal):

```js
import { onRequest } from './dist/middleware.hash.js';
import { createAPIContext } from 'astro/whatever';

export default async function middleware(request) {

 const next = async () => {
 	const response = await fetch(url, {
 		headers: {
 			"x-astro-locals": serializeLocals(ctx.locals)
 		}
 	});
 	return response;
 };

  const ctx = createAPIContext(request, whateverElse);

  return onRequest(ctx, next);
}
```

The first `import` is the Astro Middleware.
The second `import` is something we don't have yet, but we will be required to call `onRequest`. As you can see, the `next` function is where we pass the `locals`, serialize them and contact the "origin server" (the actual lambda that renders the page).

As for now, locals **are not part of this PR**. The only goal is the creation of the `middeware.js` file, bundled via `esbuild`.

Talking with @matthewp , we moved `middlewareEntryPoint` from `astro:build:done` to `astro:build:ssr`. We agreed that this information belongs to SSR, like `entryPoints`.

## Testing

I created a test case to check if the file emitted exists.
I also manually tested using a pet project.

## Docs

Still early

### DIFF
--- a/.changeset/good-pigs-fetch.md
+++ b/.changeset/good-pigs-fetch.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': minor
+---
+
+
--- a/packages/integrations/vercel/src/serverless/adapter.ts
+++ b/packages/integrations/vercel/src/serverless/adapter.ts
@@ -90,7 +90,10 @@ export default function vercelServerless({
 			'astro:build:ssr': async ({ middlewareEntryPoint }) => {
 				if (middlewareEntryPoint) {
 					const outPath = fileURLToPath(buildTempFolder);
-					const bundledMiddlewarePath = await generateEdgeMiddleware(middlewareEntryPoint, outPath);
+					const bundledMiddlewarePath = await generateEdgeMiddleware(
+						middlewareEntryPoint.path,
+						outPath
+					);
 					// let's tell the adapter that we need to save this file
 					filesToInclude.push(bundledMiddlewarePath);
 				}
--- a/packages/integrations/vercel/src/serverless/adapter.ts
+++ b/packages/integrations/vercel/src/serverless/adapter.ts
@@ -12,7 +12,7 @@ import { exposeEnv } from '../lib/env.js';
 import { getVercelOutput, removeDir, writeJson } from '../lib/fs.js';
 import { copyDependenciesToFunction } from '../lib/nft.js';
 import { getRedirects } from '../lib/redirects.js';
-import { generateEdgeMiddleware } from './middleware';
+import { generateEdgeMiddleware } from './middleware.js';
 import { fileURLToPath } from 'node:url';
 
 const PACKAGE_NAME = '@astrojs/vercel/serverless';
--- a/packages/integrations/vercel/src/serverless/adapter.ts
+++ b/packages/integrations/vercel/src/serverless/adapter.ts
@@ -9,9 +9,11 @@ import {
 	type VercelImageConfig,
 } from '../image/shared.js';
 import { exposeEnv } from '../lib/env.js';
-import { getVercelOutput, removeDir, writeFile, writeJson } from '../lib/fs.js';
+import { getVercelOutput, removeDir, writeJson } from '../lib/fs.js';
 import { copyDependenciesToFunction } from '../lib/nft.js';
 import { getRedirects } from '../lib/redirects.js';
+import { generateEdgeMiddleware } from './middleware';
+import { fileURLToPath } from 'node:url';
 
 const PACKAGE_NAME = '@astrojs/vercel/serverless';
 
@@ -42,6 +44,7 @@ export default function vercelServerless({
 	let buildTempFolder: URL;
 	let functionFolder: URL;
 	let serverEntry: string;
+	let serverBuildDir: URL;
 
 	return {
 		name: PACKAGE_NAME,
@@ -52,12 +55,14 @@ export default function vercelServerless({
 				}
 				const outDir = getVercelOutput(config.root);
 				const viteDefine = exposeEnv(['VERCEL_ANALYTICS_ID']);
+				serverBuildDir = new URL('./dist/', config.root);
 				updateConfig({
 					outDir,
 					build: {
 						serverEntry: 'entry.mjs',
 						client: new URL('./static/', outDir),
-						server: new URL('./dist/', config.root),
+						server: serverBuildDir,
+						excludeMiddleware: true,
 					},
 					vite: {
 						define: viteDefine,
@@ -80,7 +85,7 @@ export default function vercelServerless({
 	`);
 				}
 			},
-			'astro:build:done': async ({ routes }) => {
+			'astro:build:done': async ({ routes, middlewareEntryPoint }) => {
 				// Merge any includes from `vite.assetsInclude
 				const inc = includeFiles?.map((file) => new URL(file, _config.root)) || [];
 				if (_config.vite.assetsInclude) {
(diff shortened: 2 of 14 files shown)


ematipico:

ematipico:

ematipico:
I just tried the preview release in a pet project. I tracked a minor bug, which I fixed in the last commit. After the fix, I could see the bundled `middleware.js`.

The code contains the business logic of the astro middleware, plus some imported code because `astro/middleware` doesn't export a `createAPIContext` function, which will be done in a separate PR.

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Then this file gets bundled with esbuild. The bundle phase will inline the Astro middleware code.
+ *
+ * @param astroMiddlewareEntryPoint
+ * @param outPath
+ * @returns {Promise<URL>} The path to the bundled file
+ */
+export async function generateEdgeMiddleware(
+	astroMiddlewareEntryPoint: URL,
+	outPath: string
+): Promise<URL> {
```
bluwy, line 20: Is it correct that we're generating an edge middleware in [Vercel Serverless](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes) mode? Shouldn't the middleware run in the serverless runtime instead?

ematipico: Yes, that's correct. That's the intended logic. The middleware is an **edge** function, and it doesn't run in a serverless enviroment: https://vercel.com/docs/concepts/functions/edge-middleware#

From what I learnt, in Vercel an edge function is a Cloudflare worker, while a serverless function is an AWS lambda. Two different runtimes

matthewp: This is an opt-in feature.

The point of this feature is to run your middleware closer to the user. This means if you do something like a redirect to a language-specific subpath you do it close to the user. Then your regular app runs in serverless.

There wouldn't be a reason to have this feature for `@astrojs/vercel/edge`. You're already at the end. Splitting middleware into a separate function just makes it slower.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -52,12 +57,14 @@ export default function vercelServerless({
...
 				const viteDefine = exposeEnv(['VERCEL_ANALYTICS_ID']);
+				serverBuildDir = new URL('./dist/', config.root);
 				updateConfig({
 					outDir,
 					build: {
 						serverEntry: 'entry.mjs',
 						client: new URL('./static/', outDir),
-						server: new URL('./dist/', config.root),
+						server: serverBuildDir,
+						excludeMiddleware: true,
```
matthewp, line 67: Per our conversation, we are going to let the user toggle this themselves, right?

ematipico: Good catch! Yes, that's intention

packages/integrations/vercel/src/serverless/middleware.ts
bluwy: If it's opt-in then that's good by me then 👍 I thought it will be a feature automatically turned on.

ematipico: > If it's opt-in then that's good by me then 👍 I thought it will be a feature automatically turned on.

Yeah, that's the whole philosophy. The user **must** set the correct configuration in order to trigger the feature. I reverted the code in the test case, Matthew caught the bug.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -52,12 +57,13 @@ export default function vercelServerless({
...
 				const outDir = getVercelOutput(config.root);
 				const viteDefine = exposeEnv(['VERCEL_ANALYTICS_ID']);
+				serverBuildDir = new URL('./dist/', config.root);
 				updateConfig({
 					outDir,
 					build: {
 						serverEntry: 'entry.mjs',
 						client: new URL('./static/', outDir),
-						server: new URL('./dist/', config.root),
+						server: serverBuildDir,
```
bluwy, line 66: Not sure if it's intentional, but the `serverBuildDir` is extracted out, but it isn't used anywhere else than here.

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+		format: 'esm',
+		bundle: true,
+		minify: false,
+	});
+	return pathToFileURL(bundledFilePath);
+}
+
+function edgeMiddlewareTemplate(middlewarePath: string) {
+	return `
+import { onRequest } from "${middlewarePath}"
```
bluwy, line 46: Presumably if `middlewareEntryPoint` returns `URL` directly, this part could be:

```js
import { onRequest } from ${JSON.stringify(fileUrlToPath(middlewareUrl))}
```

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+export default function middleware(request) {
+\tconst url = new URL(request.url);
+\tconst next = async () => {
+\t\tconst response = await fetch(url);
+\t\treturn response;
+\t};
+
+\tconst ctx = createAPIContext(request);
+
+\treturn onRequest(ctx, next);
```
bluwy, line 57: Not a big issue, but a bit strange to see `\t` used instead of spaces, since it will get bundled too 😅

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export async function generateEdgeMiddleware(
+	astroMiddlewareEntryPointPath: string,
+	outPath: string
+): Promise<URL> {
+	const code = edgeMiddlewareTemplate(astroMiddlewareEntryPointPath);
+	// Temporary file, so `esbuild` can resolve it without too much trouble.
+	// Using `stdin` option is not very optimal
+	const middlewareEdgePath = join(outPath, 'middleware-temp.mjs');
+	await writeFile(middlewareEdgePath, code);
```
bluwy, line 25: I'm curious why `stdin` isn't optimal? It should be fine for small amount of code passed 🤔 I'm using `stdin` here too:

https://github.com/withastro/astro/blob/1a9b644e20b9ff1e8484aae3e97a9ee2fa81238c/packages/astro/src/core/client-directive/build.ts#L12-L21

.changeset/good-pigs-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vercel': minor
+---
+
+
```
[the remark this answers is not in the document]
ematipico: Yeah I used it only to trigger the preview release. I will update it towards the end once the new feature is ready.

packages/integrations/vercel/src/serverless/middleware.ts
ematipico: Oh, I noticed noticed that we do `middlewareUrl.toString()`. We use this trick somewhere else: https://github.com/withastro/astro/blob/2df9002a561a008cea0a37b52159365a3e74ae68/packages/astro/src/core/build/generate.ts#L152

ematipico:
I am going to merge it. The test fail is not related to my changes. Feel free to leave more feedback and I will address it in the next PR.

PR #7495, opened 2023-06-27
Run documentation generation as part of docs smoke test

delucis:
## Changes

- Updates CI to run the various `docgen` scripts in docs before building the docs site in smoke tests
- These scripts generate docs based on files in the `astro` repo and will run based on the current `astro` branch here so we can catch changes that would break docs early.

## Testing

It’s all tests — existing tests should pass in this PR. Will check to make sure it’s running the docgen in CI as expected.

## Docs

Not required I don’t think as it’s a testing change more than anything.


matthewp (approved):

ematipico (commented):
The change looks good to me, although I would move this to another workflow, and trigger it only when the known files are modified. 

As far as I understood, only few files trigger the documentation.

delucis:
> we can make the new workflow `workflow_dispatch`, so we can manually run the workflow from any PR we need.

If we do that people will generally forget to run it, so I’d be pretty strongly in favour of not doing that. Is 7 seconds of CI time really a deal breaker here? If so I can look at setting up filtering.

ematipico:
> You were suggesting workflow_dispatch in addition to when the filter is in place, right? Not only running it on workflow_dispatch. I’ll have a think if there’s a way we can do that.

Yes exactly, that's what I meant

ematipico (approved):

delucis:
OK, so I added the step to check if changes exist yesterday. Seems basically instant in the Ubuntu smoke test, so cut ~7s script run time to <1s. But took 4s on Windows, so roughly half of ~9s runs. If folks think those time savings are worth the added complexity, I’m happy to roll with it.

ematipico:
Looks good to me :) let's 🚢 it!

PR #7498, opened 2023-06-27
feat: export a new api called `createContext`

ematipico:
## Changes

Tiny change. For more context, the the PR https://github.com/withastro/astro/pull/7492

## Testing

I manually checked if the compiled code exports the new API

## Docs

Not needed for now

### DIFF
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -31,19 +31,26 @@ type EndpointCallResult =
 			response: Response;
 	  };
 
+type CreateAPIContext = {
+	request: Request;
+	params: Params;
+	site?: string;
+	props: Record<string, any>;
+	adapterName?: string;
+};
+
+/**
+ * Creates a context that holds all the information needed to handle an Astro endpoint.
+ *
+ * @param {CreateAPIContext} payload
+ */
 export function createAPIContext({
 	request,
 	params,
 	site,
 	props,
 	adapterName,
-}: {
-	request: Request;
-	params: Params;
-	site?: string;
-	props: Record<string, any>;
-	adapterName?: string;
-}): APIContext {
+}: CreateAPIContext): APIContext {
 	const context = {
 		cookies: new AstroCookies(request),
 		request,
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/endpoint/index.ts
```diff
@@ -31,19 +31,26 @@ type EndpointCallResult =
 			response: Response;
 	  };
 
+type CreateAPIContext = {
+	request: Request;
+	params: Params;
+	site?: string;
+	props: Record<string, any>;
+	adapterName?: string;
+};
```
bluwy, line 40: The feature sounds good to me. I have a small concern if this param for `createAPIContext` is too wide though. Other than `request`, is it easily possible for integrations to derive `params`, `site`, `props` and `adapterName`?

`site` and `adapterName` could be something Astro automatically handles.

bluwy (approved):

packages/astro/src/core/endpoint/index.ts
ematipico: @matthewp and I discussed it. We landed to these conclusions:
- `adapterName` is the name of the adapter itself (integration)
- `props`, `site` and `params` can be empty. `params` be derived from the `request`, but an adapter might have a different type from Astro `Params` - which is a `Record<string, string | undefined>`. 


Should we expose a small function that wraps the original `createAPIContext` with narrowed types? Ideally, only request, params and adapterName. What do you think? I like the idea

bluwy: Yeah I think it would be great to narrow the types. Could we also exclude the `adapterName`? The value (like `site`) can already be found from the config in build-time.

packages/astro/src/core/middleware/index.ts
```diff
@@ -1,9 +1,40 @@
...
+	 */
+	params?: Params;
+};
+
+/**
+ * Creates a context to be passed to Astro middleware `onRequest` function.
+ *
+ * @param request
+ * @param params
+ * @param adapterName
```
bluwy, line 28: I think the `@param`s here can be removed, since there's only a single object param

bluwy (approved):

PR #7500, opened 2023-06-27
fix: make `Picture` generate valid dev URLs

### LINKED ISSUE #7481: @astrojs/image@0.17.1 Picture component double encoded srcset value

2.7.0

None

npm

Linux

chrome, firefox

### Describe the Bug

Passing `GetImageResult` to `<Picture/>` component results in image 404 error (image not shown) if any of the `<source>` are loaded instead of the original `img` src attribute. The URL-s in the srcset attributes are double encoded.

Picture HTML tag is rendered with desired sources (formats and sizes).

https://stackblitz.com/edit/github-cvzpwg-hrfg8c?file=src%2Fpages%2Findex.astro

### Participation

wooseopkim:
## Changes

- Fixes #7481
  - `encodeURI`ing the whole `src` has unwanted side effects other than the original purpose.
  - Encoding only ` `(space characters) seems to be the only way to make sure there are no side effects.
  - See #7340 for the original/wanted change.

## Testing

- Added some assertions to the unit tests.

## Docs

- Fixed bug in `Picture` and `getPicture` that generates invalid dev URLs.

### DIFF
--- a/.changeset/few-ants-brake.md
+++ b/.changeset/few-ants-brake.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/image": patch
+---
+
+fix: make `Picture` generate valid dev URLs
--- a/packages/integrations/image/src/lib/get-picture.ts
+++ b/packages/integrations/image/src/lib/get-picture.ts
@@ -86,7 +86,6 @@ export async function getPicture(params: GetPictureParams): Promise<GetPictureRe
 				}
 
 				return `${img.src?.replaceAll(' ', encodeURI)} ${width}w`;
-				// return `${encodeURI(img.src ?? '')} ${width}w`;
 			})
 		);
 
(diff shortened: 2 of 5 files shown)


wooseopkim:
This feels a little different than making dumb mistakes while working on my company's product. This also feels guilty like it does, but somehow more... shameful or degrading? Not being a native speaker, I don't which word to choose. Contributing to the project with magnificent DX feels great though. Kudos to Astro maintainers and any other open source developers.

bluwy (approved):
I think I understand the issue better now. When we're given `/foo space?something-query`, we only want to encode the `/foo space` part of the URL, not the `?...` query.

So the `encodeUri` should probably be applied somewhere here:

https://github.com/withastro/astro/blob/3df71fd34fe2f50d81776e5ccdb7a340a4b2ca41/packages/integrations/image/src/lib/get-image.ts#L144-L159

Given that this fixes a regression and we're phasing out `@astrojs/image` in favour of experimental assets, let's get this quick fix in first.

Issue #7442, opened 2023-06-21
`<Picture />` ignoring src's `IFD0:Orientation` metadata


auroraanna:
### What version of `astro` are you using?

2.5.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Linux

### What browser are you using?

Brave (Chromium)

### Describe the Bug

The <Picture /> element is ignoring the `IFD0:Orientation` image metadata of the src attribute which results in falsely rotated images and a stretched image.

## Example

The following image with width 2998, height 3998 and orientation 90 degrees clockwise as displayed by gThumb:
![image](https://github.com/withastro/astro/assets/81317317/8acf9a97-0dab-40ac-bc9f-24de5e1ca41d)

The result of passing it through Astro's `<Picture>` element:
![image](https://github.com/withastro/astro/assets/81317317/8c0cdbb6-c90c-4463-9746-ff7979993d9c)

### What's the expected result?

The image metadata orientation should be taken into account.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-m6jojv

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Thanks for opening an issue! We are focusing our efforts on the built-in [`assets`](https://docs.astro.build/en/guides/assets/) support rather than `@astrojs/image` bugs, but we'd be happy to review a PR that fixes this!

Mrowa96:
@natemoo-re I created PR to fix the issue, can I ask you for CR?

natemoo-re:
Yep, taking a look now. Thanks so much @Mrowa96!

PR #7510, opened 2023-06-28
Fix image metadata generation

Mrowa96:
## Changes

Fixes https://github.com/withastro/astro/issues/7442 
- Error was only happening with squoosh, not with sharp
- It was affecting both `Image` and `Picture` components
- It was caused because `metadata` function is called with `/@astroimage/...` src after transformations are done. But `fileURLToPath` cannot accept this kind of url and it throws an error.

## Testing

Manually, here are the results.
Before:
image
After:
image

## Docs

Api didn't change so I don't think there is a need to update the docs.


natemoo-re (approved):
Thanks for the PR!

PR #7511, opened 2023-06-28
Initial View Transition Support

matthewp:
This is an experiment in View Transition support in Astro. You can enable the preview release to play around with:

```shell
npm install astro@next--view-transitions --save
```

There are a couple of parts to this to understand:

## Animation directives

This PR adds a new directive `transition:animate` which takes a few values `'morph' | 'slide' | 'fade'`. 

You use it like so:

```astro
<body transition:animate="slide">
...
```

> **Important**: These are *not required* to use. You can also specify your own animations in CSS. This is intended to be a suite of built-in styles.

Each element you animation gets a `view-transition-name` CSS property. Astro will auto-generates these for you based on the position within the DOM. Some times if you are morphing two elements that are located in different places the auto-generation won't work, in which case you can manually specify a transition name:

__index.astro__

```astro
<div class="card" transition:animate="morph" transition:name="card">
```

__post.astro__

```astro
<div class="hero" transition:animate="morph" transition:name="card">
```

## Enabling View Transitions

The feature needs to be enabled on a per-page basis. If you are using a layout or head component you can enable it cross-site that way.

### SPA mode

This works in all Chromium based browsers (Chrome, Edge, Opera, Vivaldi), stable.

```astro
---
import { ViewTransitions } from 'astro/components';
---
<html>
  <head>
    <title>My App</title>
    <ViewTransitions />
  </head>
  <body transition:animate="fade">
    ...
  </body>
</html>
```

### MPA mode

This works only in Chrome Canary after enabling a couple of flags.

```
chrome://flags#view-transition
chrome://flags#view-transition-on-navigation
```

```astro
<html>
  <head>
    <title>My App</title>
    <meta name="view-transition" content="same-origin" />
  </head>
  <body transition:animate="fade">
    ...
  </body>
</html>
```

Notice that this is the same as SPA mode, but instead of using our ViewTransitions component you add a meta tag instead. This means 0 JS.

## Browser support

This PR targets browsers that support view transitions. In the future we would like to provide similar abilities via some sort of fallback implementation. How that works is TBD.

For this release it's relevant to decide what to do about non-supported browsers. We could either:

- Do client-side routing but without any animations.
- Fallback to regular MPA routing.

---

## Changes

- Adds experimental support for [View Transitions](https://developer.chrome.com/docs/web-platform/view-transitions/)
- The goal is to make View Transitions (something you can already use in Astro) feel deeply integrated with Astro.

## Testing

- Not done yet. This should be done via playwright tests.

## Docs

Not yet.


matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

packages/astro/tsconfig.json
```diff
@@ -6,6 +6,8 @@
     "declarationDir": "./dist",
     "module": "ES2022",
     "outDir": "./dist",
-    "target": "ES2021"
+    "target": "ES2021",
+    "jsx": "preserve",
+    "types": ["@types/dom-view-transitions", "network-information-types"]
```
Princesseuh, line 11: Using `types` make type acquisition explicit, which could affect other things (ex: we use `@types/node`), something to be on the lookout for

matthewp: What's the alternative?

Princesseuh: It depends how you need the types: 
- If you need them globally: Importing them in a `.d.ts` (like how our `env.d.ts` works)
- If you only need them in specific files: Importing the types in said files

Princesseuh: I'll note that it's not a problem, making type acquisition explicit, it's even good for performance. It's just something to be aware of

matthewp:
Thanks @TheOtterlord, I'm betting I need to use `ev.currentTarget` rather than `ev.target`. Will check and let you know.

packages/astro/tsconfig.json
matthewp: Changed this to an import.

matthewp:
!preview view-transitions

TheOtterlord:
Thanks! Can confirm that's fixed it on both test sites.

TheOtterlord:
Quick update. After updating, it appears to break fallback support in Firefox. It replaces the pathname in the URL, but doesn't update the DOM. The [Stackblitz](https://stackblitz.com/github/TheOtterlord/astro-view-transitions-bug) should show this problem on the latest preview release

matthewp:
@TheOtterlord Thanks! I had combined the view transition selectors with the fallback using a comma, but it turns out that if a browser doesn't understand part of a selector it ignores the whole thing. Fixed it, new preview release will come out in a minute.

https://github.com/withastro/astro/assets/361671/7675d2e2-b2f8-4711-a198-6d19e63742da

matthewp:
!preview view-transitions

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		let el = document.querySelector('[name="astro-view-transitions-fallback"]');
+		if(el) {
+			return el.getAttribute('content') as Fallback;
+		}
+		return 'animate';
+	}
+
+	const parser = new DOMParser();
+
+  async function updateDOM(dir: Direction, html: string, fallback: Fallback = 'swap') {
```
ematipico, line 40: Here the default `fallback` is `swap`, but we return `'animate'` inside the function `getFallback`. Is it intended?

sarah11918 (commented):
Just left a couple of quick comments for docs/changeset, @matthewp !

Also, I noticed that this PR adds the experimental flag to a few of the example starter templates (basics, blog, hackernews, portfolio and with-tailwindcss), so just wanted to call attention to that and make sure it's intentional!

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		if(fallback === 'animate') {
+			let isAnimating = false;
+			addEventListener('animationstart', () => isAnimating = true, { once: true });
+
+			document.documentElement.dataset.astroTransitionFallback = 'old';
+			doc.documentElement.dataset.astroTransitionFallback = 'new';
+			// Trigger the animations
+			addEventListener('animationend', swap, { once: true });
+			// If there are no animations, go ahead and swap on next tick
+			setTimeout(() => !isAnimating && swap());
```
ematipico, line 54: Could you leave a comment explaining why `setTimeout` is needed? Usually, `setTimeout` is an hint of "something is not right"

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		} else {
+			swap();
+		}
+  }
+
+	async function navigate(dir: Direction, href: string) {
+		let htmlPromise = getHTML(href);
+		let finished: Promise<void>;
+		const { html, ok } = await htmlPromise;
```
ematipico, line 63:
```suggestion
		const finished: Promise<void>;
		const { html, ok } = await getHTML(href);
```

We don't use `htmlPromise` anywhere else.

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		// If there is a problem fetching the new page, just do an MPA navigation to it.
+		if(!ok) {
+			location.href = href;
```
ematipico, line 66: Should we add a `return` here?

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		document.head.append(link);
+	}
+
+	if(supportsViewTransitions || getFallback() !== 'none') {
+		document.addEventListener('click', (ev) => {
+		let link = ev.target;
+		if(link instanceof Element && link.tagName !== 'A') {
+			link = link.closest('a');
+		}
+		if (link &&
```
ematipico, line 99: Could you add a comment that explains what this check does?

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+			!ev.ctrlKey && // new tab (windows)
+			!ev.altKey && // download
+			!ev.shiftKey &&
+			!ev.defaultPrevented &&
+			transitionEnabledOnThisPage()
+			) {
+			ev.preventDefault();
+    	navigate('forward', link.href);
+			history.pushState({index: currentIndex + 1}, '', link.href);
+			currentIndex++;
```
ematipico, line 115:
```suggestion
			currentIndex += 1;
			history.pushState({index: currentIndex }, '', link.href);
```

packages/astro/components/ViewTransitions.astro
```diff
@@ -0,0 +1,137 @@
...
+		}
+  });
+	window.addEventListener('popstate', () => {
+		if(!transitionEnabledOnThisPage()) return;
+		const index = history.state?.index ?? (currentIndex + 1);
```
ematipico, line 120: nit: `index` => `nextIndex`

packages/astro/src/transitions/vite-plugin-transitions.ts
```diff
@@ -0,0 +1,40 @@
+import type { AstroConfig } from '../@types/astro';
+import * as vite from 'vite';
+import { AstroError } from '../core/errors/index.js';
+
+const virtualModuleId = 'astro:transitions';
+
+// The virtual module for the astro:transitions namespace
+export default function astroTransitions({ config }: { config: AstroConfig; }): vite.Plugin {
+	const resolvedVirtualModuleId = '\0' + virtualModuleId;
```
bluwy, line 9: I think we can move this to the top level of the file.

packages/astro/src/runtime/server/transition.ts
```diff
@@ -0,0 +1,155 @@
...
+	let animations: TransitionDirectionalAnimations | null = null;
+	switch(animationName) {
+		case 'fade': {
+			animations = fade();
+			break;
+		}
+		case 'slide': {
+			animations = slide();
+			break;
+		}
```
bluwy, line 34: Note: I was thinking if this not being tree-shakable would be an issue, but our client directives are also not treeshakable so I think this is also fine for now.

examples/with-tailwindcss/src/layouts/main.astro
```diff
@@ -8,8 +9,9 @@ const { content } = Astro.props;
 		<meta name="viewport" content="width=device-width" />
 		<link rel="icon" type="image/svg+xml" href="/favicon.svg" />
 		<title>{content.title}</title>
+		<ViewTransitions />
 	</head>
-	<body>
+	<body transition:animate="morph">
```
bluwy, line 14: Do we want to remove the view transitions added in the examples before merging? These are used for create-astro templates

packages/astro/components/ViewTransitions.astro
matthewp: Ah yeah, the browser doesn't have an API to tell you if there is an animation queued up. If there is no animation (if the user enabled view transitions but did not add any animations) then the `animationend` event will never fire. So the setTimeout is a fallback in case there are no animations, we still want to swap in the new DOM.

I'll add a comment explaining.

packages/astro/components/ViewTransitions.astro
matthewp: Finished is used though, it needs to be a `let`.

packages/astro/components/ViewTransitions.astro
matthewp: I don't think you need to, but I can for readability.

.changeset/silly-garlics-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+View Transitions
```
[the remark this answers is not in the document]
matthewp: Updated

examples/with-tailwindcss/src/layouts/main.astro
matthewp: I was thinking we wanted to promote the usage of this by having it in the examples. Will ask docs what they think.

packages/astro/components/ViewTransitions.astro
matthewp: This was really a weird case to handle non-fallback. I made the argument optional instead.

.changeset/silly-garlics-live.md
```diff
@@ -0,0 +1,60 @@
...
+    viewTransitions: true,
+  },
+})
+```
+
+This enables you to use the new APIs added.
+
+## <ViewTransitions />
+
+This is a component which acts as the *router* for transitions between pages. You need to add it to each page, in the `<head>` section, in which you want to have transitions away from be done in-client. To enable support throughout your entire app, add the component in some common layout or component.
```
sarah11918, line 25:
```suggestion
This is a component which acts as the *router* for transitions between pages. Add it to the `<head>` section of each individual page where transitions should occur *in the client* as you navigate away to another page, instead of causing a full page browser refresh. To enable support throughout your entire app, add the component in some common layout or component that targets the `<head>` of every page.
```

bluwy (approved):
Looks great! Also tested locally and it works. I notice there's a transition difference between native and fallback in the `blog` example, but I don't think that can be easily fixed without more complexity.

.changeset/silly-garlics-live.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Animations
+
+Using `transition:animate` you can use some of our built-in animations. For example:
+
+```astro
+<header transition:animate="slide">
+```
+
+Will perform an animation in which the element slides out to the left, and slides in from the right. You can also customize these animations by specifying a duration like so:
```
sarah11918, line 49:
``````suggestion
Add `transition:animate` to any element to use Astro's built-in animations.

```astro
<header transition:animate="slide">
```

In the above, Astro's `slide` animation will cause the `<header>` element to slide out to the left, and then slide in from the right when you navigate away from the page.

You can also customize these animations using any CSS animation properties, for example, by specifying a duration:
``````

We try to keep descriptions of different code samples on their own lines, so that we don't describe both the above and below together!

Check my wording here. It's a little more verbose but I do like explicitly mentioning:
- *add this to an element* vs "use" so that usage is as clear as possible
- showing off that you can add any valid CSS! (Check my wording: "any CSS animation properties" <- is that exactly correct, or the best way to say it?)

File: .changeset/silly-garlics-live.md [hunk 15]
```diff
@@ -0,0 +1,60 @@
...
+```astro
+---
+import { slide } from 'astro:transition';
+---
+<header transition:animate={slide({ duration: 200 })}>
+```
+
+## Continue learning
+
+Checkout the [docs](https://docs.astro.build/en/guides/client-side-routing/) to learn more.
```
sarah11918, line 60:
```suggestion
Check out the [client-side routing docs](https://docs.astro.build/en/guides/client-side-routing/) to learn more.
```
We try to make link text as explicit as we can! It's especially helpful for assistive technologies, but also provides a solid visual to the eye on where you'll be taken.

sarah11918 (commented):
Some suggestions for the changeset!  :raised_hands:

examples/with-tailwindcss/src/layouts/main.astro
sarah11918: Docs thinks remove! :smile:

PR #7513, opened 2023-06-28
Added `readImageFile` helper in astro:assets

FineWolf:
## Changes

- Added `readImageFile(src: ImageMetadata)` helper in `astro:assets`. Returns a `Buffer` of the imported file when passed an `ImageMetadata`.
- Handles the following cases:
  - `readImageFile()` is called before client is generated, while magic strings (`__ASTRO_ASSET_IMAGE__[hash]__`) are still in use.
  - `readImageFile()` is called during building pages, after magic strings are replaced with the final relative urls (`/_astro/file.ext`).
  - `readImageFile()` is called using the dev server, while `.src` returns URLs in the format of `/@fs/<path>` (on Windows and POSIX)

## Testing

This change was tested in a private project of mine where I'm hitting all the use cases described above. I did not adjust any test fixtures to test if `readImageFile()` returns a proper buffer.

However, I could modify the `astro-assets` and `astro-assets-prefix` fixtures to add an attribute to the `img` derived from the content of the buffer (a hash, maybe?).

## Docs

The assets page document will need to be updated to add information about this new method; what use cases it can accomplish (generate small placeholder images on the fly using `plaiceholder`, or maybe reading EXIF information from the source images for a portfolio website).

/cc @Princesseuh This is related to [our discussion on the official Discord](https://discord.com/channels/830184174198718474/1123600493848047670)

### DIFF
--- a/packages/astro/client-base.d.ts
+++ b/packages/astro/client-base.d.ts
@@ -52,6 +52,9 @@ declare module 'astro:assets' {
 		) => Promise<import('./dist/assets/types.js').GetImageResult>;
 		getConfiguredImageService: typeof import('./dist/assets/index.js').getConfiguredImageService;
 		Image: typeof import('./components/Image.astro').default;
+		readImageFile: (
+			src: import('./dist/assets/types.js').ImageMetadata
+		) => Promise<Buffer>;
 	};
 
 	type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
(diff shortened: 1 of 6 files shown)


lilnasy:
This seems heavily tied into Node. Some deployment targets dont have `Buffer`, or even a filesystem. What would happen on those runtimes?

Princesseuh:
Yeah, this is unfortunately not the proper way to add this due to SSR. I'll properly review this when I'm back to work in a few weeks

FineWolf:
Can someone point me in a direction of an example where such pattern IS implemented properly? 

I'm currently blocked on my project until this feature is available, so I would love to provide a PR that is compliant with all runtimes.

lilnasy:
Taking a cursory glance at [src/assets/internal.ts](https://github.com/withastro/astro/pull/7513/files#diff-b49dbbca6a48cc398542ddd82a7b394a428beebad0d8348e97706b875cc0cac9), I feel like it could be made into an integration.

FineWolf:
I'm closing this pull request. Since I'm mostly concerned with SSG in my use-case, I've managed to rewrite it in an integration.

PR #7518, opened 2023-06-29
Refactor component and renderers loading flow

bluwy:
## Changes

In the `preload` call, we're passing `[renderers, component]` everywhere. The `preload`'s responsibility is to only load the component, so it should return the component directly instead of a tuple.

The `renderers` were returned because it's used to assign to `env.renderers` in dev.

In this PR, I made `preload` return `component` directly, and assign `env.renderers` itself to not leak it.

## Testing

Existing test should pass.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/core/endpoint/dev/index.ts
+++ b/packages/astro/src/core/endpoint/dev/index.ts
@@ -5,12 +5,8 @@ import { createRenderContext } from '../../render/index.js';
 import { callEndpoint } from '../index.js';
 
 export async function call(options: SSROptions, logging: LogOptions) {
-	const {
-		env,
-		preload: [, mod],
-		middleware,
-	} = options;
-	const endpointHandler = mod as unknown as EndpointHandler;
+	const { env, preload, middleware } = options;
+	const endpointHandler = preload as unknown as EndpointHandler;
 
 	const ctx = await createRenderContext({
 		request: options.request,
@@ -18,7 +14,7 @@ export async function call(options: SSROptions, logging: LogOptions) {
 		pathname: options.pathname,
 		route: options.route,
 		env,
-		mod: endpointHandler as any,
+		mod: preload,
 	});
 
 	return await callEndpoint(endpointHandler, env, ctx, logging, middleware?.onRequest);
--- a/packages/astro/src/core/render/dev/index.ts
+++ b/packages/astro/src/core/render/dev/index.ts
@@ -5,17 +5,14 @@ import type {
 	MiddlewareResponseHandler,
 	RouteData,
 	SSRElement,
-	SSRLoadedRenderer,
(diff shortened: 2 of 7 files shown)


packages/astro/src/core/render/dev/index.ts
```diff
@@ -156,11 +148,7 @@ async function getScriptsAndStyles({ env, filePath }: GetScriptsAndStylesParams)
 }
 
 export async function renderPage(options: SSROptions): Promise<Response> {
-	const [renderers, mod] = options.preload;
-
-	// Override the environment's renderers. This ensures that if renderers change (HMR)
-	// The new instances are passed through.
-	options.env.renderers = renderers;
```
bluwy, line 163: The `renderers` were mainly used here, but anywhere else doesn't need it.

packages/astro/src/core/render/dev/index.ts
```diff
@@ -44,36 +41,31 @@ export interface SSROptions {
...
+}: {
+	env: DevelopmentEnvironment;
+	filePath: URL;
+}): Promise<ComponentInstance> {
 	// Important: This needs to happen first, in case a renderer provides polyfills.
-	const renderers = await loadRenderers(env.loader, env.settings);
+	const renderers = await loadRenderers(env.settings, env.loader);
+	// Override the environment's renderers. This ensures that if renderers change (HMR)
+	// The new instances are passed through.
+	env.renderers = renderers;
```
bluwy, line 55: I moved it here instead, immediately `loadRenderers` so we don't have to pass it around.

packages/astro/src/core/render/dev/index.ts
```diff
@@ -44,36 +41,31 @@ export interface SSROptions {
...
 
 	try {
 		// Load the module from the Vite SSR Runtime.
 		const mod = (await env.loader.import(viteID(filePath))) as ComponentInstance;
 
-		return [renderers, mod];
+		return mod;
```
bluwy, line 61: With this simplification, we can directly return the `mod` instead. The rest of the changes are to accommodate this type change.

packages/astro/src/core/endpoint/dev/index.ts
```diff
@@ -5,20 +5,16 @@ import { createRenderContext } from '../../render/index.js';
...
+	const endpointHandler = preload as unknown as EndpointHandler;
 
 	const ctx = await createRenderContext({
 		request: options.request,
 		origin: options.origin,
 		pathname: options.pathname,
 		route: options.route,
 		env,
-		mod: endpointHandler as any,
+		mod: preload,
```
ematipico, line 17: Love it!

ematipico (approved):
Looks good to me!

PR #7521, opened 2023-06-29
Make APIRoute generic like APIContext

knpwrs:
## Changes

`APIContext` is generic so that `props` can be typed, but `APIRoute`, which composes `APIContext`, is not likewise generic. This PR changes that.

## Testing

No tests, just a backwards-compatible change to a type.

## Docs

No docs added, it was intuitive to me to expect `APIRoute` to be generic and I was rather surprised that it wasn't already.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1772,8 +1772,8 @@ export interface EndpointOutput {
 	encoding?: BufferEncoding;
 }
 
-export type APIRoute = (
-	context: APIContext
+export type APIRoute<Props extends Record<string, any> = Record<string, any>> = (
+	context: APIContext<Props>
 ) => EndpointOutput | Response | Promise<EndpointOutput | Response>;
 
 export interface EndpointHandler {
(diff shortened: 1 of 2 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1772,8 +1772,8 @@ export interface EndpointOutput {
 	encoding?: BufferEncoding;
 }
 
-export type APIRoute = (
-	context: APIContext
+export type APIRoute<RouteProps extends Props = Props> = (
+	context: APIContext<RouteProps>
```
bluwy, line 1776: There's a little breaking change with this, as `Props` is `Record<string, unknown>` and previously the default type is `Record<string, any>`. I think we should have something like this instead:

```suggestion
export type APIRoute<Props extends Record<string, any> = Record<string, any>> = (
	context: APIContext<Props>
```

bluwy: Ah didn't notice that. Moving the `export type Props = Record<string, unknown>;` after the `APIRoute` type should fix it.

knpwrs: Done in 61af58e

knpwrs:
Looks like CI is passing now and the breaking change should be handled.

ematipico (approved):

bluwy (approved):

PR #7522, opened 2023-06-29
Refactor telemetry usage flow

bluwy:
## Changes

- Use telemetry singleton everywhere.
- Refactor how tests disable telemetry.

We currently have code that passes it through several functions, because tests can then pass a telemetry stub to disable telemetry. I think it's simpler to set `process.env.ASTRO_TELEMETRY_DISABLED = true` instead to disable this so we don't leak the test implementation details.

## Testing

Tested manually.

## Docs

n/a.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1375,7 +1375,6 @@ export interface AstroSettings {
 	tsConfig: TsConfigJson | undefined;
 	tsConfigPath: string | undefined;
 	watchFiles: string[];
-	forceDisableTelemetry: boolean;
 	timer: AstroTimer;
 }
 
--- a/packages/astro/src/core/render/dev/environment.ts
+++ b/packages/astro/src/core/render/dev/environment.ts
@@ -32,7 +32,6 @@ export function createDevelopmentEnvironment(
 		site: settings.config.site,
 		ssr: isServerLikeOutput(settings.config),
 		streaming: true,
-		telemetry: Boolean(settings.forceDisableTelemetry),
 	});
 
 	return {
--- a/packages/astro/test/test-utils.js
+++ b/packages/astro/test/test-utils.js
@@ -22,6 +22,9 @@ polyfill(globalThis, {
 	exclude: 'window document',
 });
 
+// Disable telemetry when running tests
+process.env.ASTRO_TELEMETRY_DISABLED = true;
+
 /**
  * @typedef {import('undici').Response} Response
  * @typedef {import('../src/core/dev/dev').DedvServer} DevServer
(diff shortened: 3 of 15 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1375,7 +1375,6 @@ export interface AstroSettings {
 	tsConfig: TsConfigJson | undefined;
 	tsConfigPath: string | undefined;
 	watchFiles: string[];
-	forceDisableTelemetry: boolean;
```
bluwy, line 1378: For the code review, you many notice `disableTelemetry`, `forceDisableTelemtry`, etc gets removed. I found these options only exists for tests to pass a value.

I removed it to simplify the code and rely on `process.env.ASTRO_TELEMETRY_DISABLED = true` instead.

packages/astro/src/core/render/dev/environment.ts
```diff
@@ -32,7 +32,6 @@ export function createDevelopmentEnvironment(
 		site: settings.config.site,
 		ssr: isServerLikeOutput(settings.config),
 		streaming: true,
-		telemetry: Boolean(settings.forceDisableTelemetry),
```
bluwy, line 35: Note: There's a bug here before. This should be `telemetry: !settings.forceDisableTelemetry`. Luckily we aren't recording any telemetry in the tests now that affects this.

packages/astro/test/test-utils.js
```diff
@@ -22,6 +22,9 @@ polyfill(globalThis, {
 	exclude: 'window document',
 });
 
+// Disable telemetry when running tests
+process.env.ASTRO_TELEMETRY_DISABLED = true;
```
bluwy, line 26: This is where I disable the telemetry. It should affect other integration tests as they import this file too.

packages/astro/test/units/shiki/shiki.test.js
```diff
@@ -9,7 +9,7 @@ describe('<Code />', () => {
 		let container;
 		let mod;
 		before(async () => {
-			container = await createContainer({ root, disableTelemetry: true });
+			container = await createContainer({ root });
```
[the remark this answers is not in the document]
ematipico: How do we discover if the unit tests will start triggering telemetry events?

ematipico (approved):

packages/astro/test/units/shiki/shiki.test.js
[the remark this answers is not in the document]
bluwy: I don't know 😄 Currently I do a global search for `.record` and `.notify`, and there's only a few places used that I know are not unit-testable.

I wouldn't mind adding a flag to the npm script to disable it completely if you prefer too.

PR #7523, opened 2023-06-29
feat: connect vercel edge middleware to astro

ematipico:
## Changes

This PR connects the Vercel Edge Middleware to the Astro middleware:
- created a new API called `trySerializeLocals` to serialise the `locals`. This is important because these `locals` will travel via headers, and we need to keep them when they are transferred from the vercel middleware to the Astro middleware;
- ~created a new API inside the `@astrojs/vercel` adapter, this API is called `createLocals` and allows users to initialize their `locals` with anything they want. The function accepts an object that contains `request` and `context`. You can check [Vercel documentation](https://vercel.com/docs/concepts/functions/edge-middleware/middleware-api#edge-middleware-signature);~ As per @bluwy suggestion, the new API is actually a new key file called `vercel-edge-middleware`, it receives an object with `request` and `context` and `esbuild` will bundle the file.
- updated the adapter to pass `locals` to `app.render`. These `locals` are extracted from the headers of the request;

## Testing

I plan to create a preview release, create a new project and deploy it on Vercel.

Also, I added a snapshot test for the generated middleware. The reason is that it's very difficult to see how's the code generated, especially for reviewers. Please let me know how do feel about it; I am happy to remove it.

## Docs

N/A Later when I know the APIs are stable and settled among the team members.

### DIFF
--- a/packages/astro/src/core/middleware/index.ts
+++ b/packages/astro/src/core/middleware/index.ts
@@ -32,5 +32,76 @@ function createContext({ request, params }: CreateContext) {
 	});
 }
 
+/**
+ * Checks whether the passed `value` is serializable.
+ *
+ * A serializable value contains plain values. For example, `Proxy`, `Set`, `Map`, functions, etc.
+ * are not accepted because they can't be serialized.
+ */
+function isLocalsSerializable(value: unknown): boolean {
+	let type = typeof value;
+	let plainObject = true;
+	if (type === 'object' && isPlainObject(value)) {
+		for (const [, nestedValue] of Object.entries(value)) {
+			if (!isLocalsSerializable(nestedValue)) {
+				plainObject = false;
+				break;
+			}
+		}
+	} else {
+		plainObject = false;
+	}
+	let result =
+		value === null ||
+		type === 'string' ||
+		type === 'number' ||
+		type === 'boolean' ||
+		Array.isArray(value) ||
+		plainObject;
+
+	return result;
+}
+
+/**
+ *
+ * From [redux-toolkit](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/isPlainObject.ts)
+ *
+ * Returns true if the passed value is "plain" object, i.e. an object whose
+ * prototype is the root `Object.prototype`. This includes objects created
+ * using object literals, but not for instance for class instances.
+ */
+function isPlainObject(value: unknown): value is object {
+	if (typeof value !== 'object' || value === null) return false;
+
+	let proto = Object.getPrototypeOf(value);
+	if (proto === null) return true;
+
+	let baseProto = proto;
+	while (Object.getPrototypeOf(baseProto) !== null) {
+		baseProto = Object.getPrototypeOf(baseProto);
+	}
+
+	return proto === baseProto;
+}
+
+/**
+ * It attempts to serialize `value` and return it as a string.
+ *
+ * ## Errors
+ *  If the `value` is not serializable if the function will throw a runtime error.
+ *
+ * Something is **not serializable** when it contains properties/values like functions, `Map`, `Set`, `Date`,
+ * and other types that can't be made a string.
+ *
+ * @param value
+ */
+function trySerializeLocals(value: unknown) {
+	if (isLocalsSerializable(value)) {
+		return JSON.stringify(value);
+	} else {
+		throw new Error("The passed value can't be serialized.");
+	}
+}
+
 // NOTE: this export must export only the functions that will be exposed to user-land as officials APIs
-export { sequence, defineMiddleware, createContext };
+export { sequence, defineMiddleware, createContext, trySerializeLocals };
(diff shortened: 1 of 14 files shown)


packages/astro/src/core/middleware/index.ts
```diff
@@ -32,5 +32,76 @@ function createContext({ request, params }: CreateContext) {
...
+
+/**
+ *
+ * From [redux-toolkit](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/isPlainObject.ts)
+ *
+ * Returns true if the passed value is "plain" object, i.e. an object whose
+ * prototype is the root `Object.prototype`. This includes objects created
+ * using object literals, but not for instance for class instances.
+ */
+function isPlainObject(value: unknown): value is object {
```
matthewp, line 73: We have packages/internal-helpers now. I could see some of the things like this going there. Not blocking though, just an idea.

matthewp (approved):

packages/integrations/vercel/test/fixtures/middleware/astro.config.mjs
```diff
@@ -2,6 +2,12 @@ import {defineConfig} from "astro/config";
...
 
 export default defineConfig({
-    adapter: vercel(),
+    adapter: vercel({
+        createLocals: ({request}) => {
+            console.log(request);
+            return {
+                "foo": "bar"
+            }
+        }}),
```
bluwy, line 11: I don't quite understand the `createLocals` option here. Is there a use case for this? Could the user initialize their locals in  `src/middleware.js` instead?

My main concern is that it's being serialized into the generated code like:

```js
	const localsFn = createLocals
 		? `(${createLocals})({ request, context })`
 		: '(() => undefined)({ request, context })';
```

And that means `createLocals` need to be self-contained (The function can't access anything outside of it).

bluwy:
> And I'm not sure why...

That's because after the SSR build step, `dist/entry.mjs` contains `import 'esbuild';` causing Vercel's nft tool to think esbuild needs to be "bundled". But you can't really bundle esbuild.

So somewhere is accidentally importing esbuild in the compiled runtime code maybe. I'm not really sure and it's hard to debug these stuff.

packages/integrations/vercel/test/fixtures/middleware/astro.config.mjs
ematipico: These are excellent questions! 

> Could the user initialize their locals in src/middleware.js instead?

No, that's because of how the architecture of this feature works, which I plan to document in the `README.md` right before the end.

[Vercel sets the architecture of the middleware](https://vercel.com/_next/image?url=%2Fstatic%2Fdocs%2Fconcepts%2Ffunctions%2Fedge-middleware-dark.png&w=1200&q=75&dpl=dpl_7X9G6SyKxrcDHCME4XbfLGf79bmF).


The Vercel middleware will always run **first**, even before Astro, and there's no way to change the order. The vercel middleware will be the **first** to receive the `Request` and the **last** to receive `Response`

> Is there a use case for this? 

Based on the constraints of the architecture, I figured out the following:
- vercel edge middleware receives a context and a request, and maybe the user wants to do something with them;
- maybe the user wants to send down to Astro some information, and this info will be used to inside the Astro page;
- maybe the user wants to run some geolocation info, and pass down the Astro page info about the localisation;

> And that means createLocals need to be self-contained (The function can't access anything outside of it).

Yes, that's the most significant constraint. Unfortunately, I couldn't find a more innovative way to do so.

packages/astro/src/core/middleware/index.ts
ematipico: I thought about it, and wasn't sure what to do. I will move the function there!

packages/integrations/vercel/test/fixtures/middleware/astro.config.mjs
bluwy: Interesting, thanks for the explanation! I think the most robust way is to introduce a new special file for this, e.g. `src/vercel-edge-middleware.js` and then the esbuild build will import that file and bundle the exported handler. A bit like Vue's [`appEntrypoint` pattern](https://docs.astro.build/en/guides/integrations-guide/vue/#appentrypoint).

It might be better in the long run 🤔

ematipico:

ematipico:
I tested the feature and the middleware works! I could see my logs, and I could see the locals working and rendered in the Astro page, I could see the new `vercel-edge-middleware` file working.

Here's the plan: I will merge this PR once the CI passes. Then I will create a final PR that will merge all the new code to `main`, and we will involve the docs team to craft some decent changelogs and documentation.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -31,6 +42,7 @@ export interface VercelServerlessConfig {
 	analytics?: boolean;
 	imageService?: boolean;
 	imagesConfig?: VercelImageConfig;
+	createLocals?: CreateLocals;
```
bluwy, line 45: Do we want to remove this option since we're using the file now?

PR #7531, opened 2023-06-30
fix: don't serialize `undefined` as `null`

### LINKED ISSUE #7528: `undefined` props are serialized as `null`

2.7.2

Node

pnpm

Linux (NixOS)

Firefox (Librewolf)

### Describe the Bug

When a Svelte component is hydrated, any props that are `undefined` become `null`. This has caused unexpected behavior and unsound typing (`prop === null` vs. `prop === undefined`) for me in the past.

In my example, it is shown that `undefined` props are indeed `undefined` when rendered on the server, but are hydrated with such props being `null` instead.

I have not tested whether this happens in other frameworks.

Any `undefined` props should remain `undefined` when hydrated.

<https://gitlab.com/wackbyte/astro-serializes-undefined-as-null>

### Participation

wackbyte:
## Changes

I added a special case for the serialization of `undefined` so that `serializeProps` doesn't serialize it as `null`.

Closes #7528.

## Testing

I expanded on the existing serialization tests and added a new case specifically for `undefined`.

I also tested it in [the example I provided for the issue](https://gitlab.com/wackbyte/astro-serializes-undefined-as-null) by changing its `astro` dependency to the `astro` package from this PR's branch.

## Docs

AFAIK the specific behavior of serialization is not documented, so no docs were added.

I decided to make the changeset a patch bump, but I'm not sure if it would be considered a breaking change instead.

### DIFF
--- a/.changeset/moody-singers-develop.md
+++ b/.changeset/moody-singers-develop.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix serialization of `undefined` in framework component props
--- a/packages/astro/e2e/fixtures/pass-js/src/components/React.tsx
+++ b/packages/astro/e2e/fixtures/pass-js/src/components/React.tsx
@@ -1,10 +1,14 @@
 import type { BigNestedObject } from '../types';
-import { useState } from 'react';
 
 interface Props {
-	obj: BigNestedObject;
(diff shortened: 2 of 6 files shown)


natemoo-re (approved):
Thanks for the clean fix and great tests!

PR #7532, opened 2023-06-30
feat: vercel edge middleware support

ematipico:
## Changes

This PR adds support for [vercel edge middleware](https://vercel.com/docs/concepts/functions/edge-middleware) via Astro middleware.

This is an opt-in feature, and the `@astrojs/vercel/serverless` adapter will automatically create the edge middleware function for you.

Please refer to the [readme for limitations and constraints](https://github.com/withastro/astro/blob/feat/vercel-edge-middleware/packages/integrations/vercel/README.md#limitations-and-constraints).

To quickly opt in to the feature, you need to use the new option `build.excludeMiddleware` and - obviously - have a middleware file in your code base:

```js
// astro.config.mjs
import {defineConfig} from "astro/config";
import vercel from "@astrojs/vercel";
export default definConfig({
   output: "server",
   adapter: vercel(),
   build: {
       excludeMiddleware: true
   }
})
```

This PR should be ready to be merged to `main`. I will create another preview release to test it a little bit, although I tested it locally, and the feature works as intended. There might be some edge cases that I didn't take into consideration.

## Testing

Test cases were added along the way.

## Docs

@withastro/maintainers-docs It's now time for your help to review the changes :) There are a bunch of changesets because a few stuff were added.


ematipico:

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,61 @@ You can use Vercel middleware to intercept a request and redirect before sending
 > **Warning**
 > **Trying to rewrite?** Currently rewriting a request with middleware only works for static files.
 
+### Vercel Edge Middleware with Astro middleware
+
+The `@astrojs/vercel/serverless` adapter can automatically create the vercel edge middleware when you have an Astro middleware in you code base.
+
+This is an opt-in feature, and the `build.excludeMiddleware` option needs to be set to `true`:
```
bluwy, line 269: Discussion: I talked with Matthew a bit on this. I'm still not sure configuring this option to control an integration behaviour makes sense from an API standpoint.

I think it's more natural if they can configure as:

```js
vercel({
  edgeMiddleware: true
})
```

because `build.excludeMiddleware` can't really be used for all integrations. Maybe we can polish this up in Astro 3.0 again, where I think it's nicer if this can only be configured from integration hooks.

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,61 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+```js
+// src/vercel-edge-middleware.js
+
+export default function({ request, context }) {
+    // do something with request and context
+    return {
+        title: "Spider-man's blog"
+    }
+}
+```
```
bluwy, line 295: Question: Is there a way to _type_ this function? e.g.
```ts
import type { EdgeMiddlewareHandler } from '@astrojs/vercel/serverless`
```

packages/integrations/vercel/README.md
ematipico: So you think that when passing `edgeMiddleware` to the integration, the integration itself should update the configuration and enable `build.split`?

packages/integrations/vercel/README.md
ematipico: Yes, there is, I will update the example :)

packages/integrations/vercel/src/lib/nft.ts
```diff
@@ -23,6 +21,7 @@ export async function copyDependenciesToFunction({
 		base = new URL('../', base);
 	}
 
+	const { nodeFileTrace } = await import('@vercel/nft');
```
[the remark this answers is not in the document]
ematipico: For some reason, `esbuild` was assing `@vercel/nft` to the bundled file. To be honest, I don't know why. When I changed this to use the dynamic import, the problem was gone.

packages/integrations/vercel/README.md
bluwy: Yeah, but I'm not sure if it's too late now as the same goes for `build.split`. Maybe better to keep this as-is and change it again in the next major.

In the future, we can move the new APIs to the `astro:config:setup` perhaps, similar to `addClientDirectives()`. Maybe named `setBuildMode({ excludeMiddleware: boolean })`.

I think `build.split` could be something always enabled in general though.

packages/astro/src/@types/astro.ts
```diff
@@ -849,6 +849,29 @@ export interface AstroUserConfig {
...
 		split?: boolean;
+
+		/**
+		 * @docs
+		 * @name build.excludeMiddleware
+		 * @type {boolean}
+		 * @default {false}
+		 * @version 2.8.0
+		 * @description
+		 * Defines how the SSR code should be bundled when built.
```
sarah11918, line 860:
```suggestion
		 * Defines whether or not any SSR middleware code will be bundled when built.
```
Since this is a boolean, I think the initial description should present exactly what the choice is as a "whether or not." Also, since this refers specifically to middleware code, that should be somewhere in the sentence. This is just an example of what that sentence could look like. Please edit so that it's correct/has the right nuance!

packages/astro/src/@types/astro.ts
```diff
@@ -849,6 +849,29 @@ export interface AstroUserConfig {
...
+		 *
+		 * When enabled, the middleware code is **not** going to be imported by the pages.
+		 *
+		 * It's going to be responsibility of the end user to execute import the code and execute it.
```
sarah11918, line 864:
```suggestion
		 * By default, middleware code is bundled and imported by all pages during the build. To instead execute and import middleware code manually, set `build.excludeMiddleware: true`:
```

```suggestion
		 * When enabled, middleware code is not bundled and imported by all pages during the build. To instead execute and import middleware code manually, set `build.excludeMiddleware: true`:
```
What do you think about something of this structure? I *think* prefer the one starting with default behaviour, then saying "if you want to do this instead, set true". That feels more like a definition/explanation of the config option to me, and then when to change it. I think it also helps with the "enabled means NOT" issue.  :smile:

But, something like the second could also work.

packages/integrations/vercel/README.md
```diff
@@ -233,9 +233,9 @@ export default defineConfig({
 });
 ```
 
-### Vercel Middleware
+### Vercel Edge Middleware
 
-You can use Vercel middleware to intercept a request and redirect before sending a response. Vercel middleware can run for Edge, SSR, and Static deployments. You don't need to install `@vercel/edge` to write middleware, but you do need to install it to use features such as geolocation. For more information see [Vercel’s middleware documentation](https://vercel.com/docs/concepts/functions/edge-middleware).
+You can use Vercel edge middleware to intercept a request and redirect before sending a response. Vercel middleware can run for Edge, SSR, and Static deployments. You don't need to install `@vercel/edge` to write middleware, but you do need to install it to use features such as geolocation. For more information see [Vercel’s middleware documentation](https://vercel.com/docs/concepts/functions/edge-middleware).
```
sarah11918, line 238:
```suggestion
You can use Vercel Edge middleware to intercept a request and redirect before sending a response. Vercel middleware can run for Edge, SSR, and Static deployments.  You may not need to install this package for your middleware. `@vercel/edge` is only required to use some middleware features such as geolocation. For more information see [Vercel’s middleware documentation](https://vercel.com/docs/concepts/functions/edge-middleware).
```
Maybe something more of this style?

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,82 @@ You can use Vercel middleware to intercept a request and redirect before sending
 > **Warning**
 > **Trying to rewrite?** Currently rewriting a request with middleware only works for static files.
 
+### Vercel Edge Middleware with Astro middleware
+
+The `@astrojs/vercel/serverless` adapter can automatically create the vercel edge middleware when you have an Astro middleware in you code base.
```
sarah11918, line 267:
```suggestion
The `@astrojs/vercel/serverless` adapter can automatically create the Vercel Edge middleware from an Astro middleware in your code base.
```
Would it be correct to say it like this? That you write it in Astro, but it gets "turned into" Vercel middleware code, executed by them? If so, I think this explains that.

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,82 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+export default defineConfig({
+   output: "server",
+   adapter: vercel(),
+   build: {
+       excludeMiddleware: true
+   }
+})
+```
+
+Optionally, you can create a known file called `vercel-edge-middleware.(js|ts)` in the [`srcDir`](https://docs.astro.build/en/reference/configuration-reference/#srcdir) folder to create [`Astro.locals`](https://docs.astro.build/en/reference/api-reference/#astrolocals).
```
sarah11918, line 284:
```suggestion
Optionally, you can create a file recognized by Vercel named `vercel-edge-middleware.(js|ts)` in the [`srcDir`](https://docs.astro.build/en/reference/configuration-reference/#srcdir) folder to create [`Astro.locals`](https://docs.astro.build/en/reference/api-reference/#astrolocals).
```
suggest replacing "known" with "recognized by..." whatever understands/recognizes this file!

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,82 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+/**
+ * 
+ * @param options.request {Request}
+ * @param options.context {import("@vercel/edge").RequestContext}
+ * @returns {object}
+ */
+export default function({ request, context }) {
+    // do something with request and context
+    return {
+        title: "Spider-man's blog"
```
sarah11918, line 299: I can't believe you made me spell check this! :joy:

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,82 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+The function:
+- must export a **default** function;
+- must **return** an `object`;
+- accepts an object with a `request` and `context` as properties;
+- `request` is typed as [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request);
+- `context` is typed as [`RequestContext`](https://vercel.com/docs/concepts/functions/edge-functions/vercel-edge-package#requestcontext);
+
+#### Limitations and constraints 
+
+When you opt in this feature, there are few constraints to take into consideration:
```
sarah11918, line 329:
```suggestion
When you opt in to this feature, there are few constraints to note:
```

packages/integrations/vercel/README.md
```diff
@@ -262,6 +262,82 @@ You can use Vercel middleware to intercept a request and redirect before sending
...
+- the Vercel Edge middleware will always be the **first** function to receive the `Request` and the last function to receive `Response`.
+
+  This an architectural constraint that follows the [boundaries set by Vercel](https://vercel.com/docs/concepts/functions/edge-middleware).
+- you're allowed to use `request` and `context` only to produce an `Astro.locals` object. 
+
+  Operations like redirects, etc. should be delegated to Astro middleware
+- `Astro.locals` **must be serializable**, failing to do so will result in a **runtime error**. 
+ 
+  This means that you **can't** store complex types like `Map`, `function`, `Set`, etc.
```
sarah11918, line 338:
```suggestion
- The Vercel Edge middleware will always be the **first** function to receive the `Request` and the last function to receive `Response`. This an architectural constraint that follows the [boundaries set by Vercel](https://vercel.com/docs/concepts/functions/edge-middleware).
- Only `request` and `context` may be used to produce an `Astro.locals` object. Operations like redirects, etc. should be delegated to Astro middleware.
- `Astro.locals` **must be serializable**. Failing to do so will result in a **runtime error**. This means that you **cannot** store complex types like `Map`, `function`, `Set`, etc.
```
I worry about what this will look like visually, so suggest making these all in the bullet points for now. If it needs changing, we can do it after we see it in docs.

sarah11918 (commented):
Looks pretty good @ematipico ! Left some copy edits to consider!  :raised_hands:

packages/integrations/vercel/src/lib/nft.ts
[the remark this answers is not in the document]
bluwy: I tried looking into the issue again, and I'm still confused why the issue is happening 😄 Maybe we can put a comment like:

```js
// The Vite bundle includes an import to `@vercel/nft` for some reason,
// and that trips up `@vercel/nft` itself during the adapter build. Using a
// dynamic import helps prevent the issue.
// TODO: investigate why
```

.changeset/strong-years-travel.md
```diff
@@ -0,0 +1,20 @@
...
+export default defineConfig({
+    build: {
+        excludeMiddleware: true
+    }
+})
+```
+
+When enabled, the code that belongs to be middleware **won't** be imported
+by the final pages/entry points. The user is responsible to import it and 
+call it manually.
```
sarah11918, line 20:
```suggestion
When enabled, the code that belongs to be middleware **won't** be imported
by the final pages/entry points. The user is responsible for importing it and 
calling it manually.
```

packages/integrations/vercel/README.md
ematipico: I believe so!

packages/integrations/vercel/README.md
ematipico: It's not recognized by Vercel, but it's recognized by the adapter (us) :)

packages/integrations/vercel/README.md
ematipico: That's what you get from  me when I have to write content that is not "foo", "bar" or "lorem" 🤣

bluwy (approved):

PR #7533, opened 2023-06-30
fix: error TS2578: Unused '@ts-expect-error' directive When user global config(.npmrc) sets shamefully-hoist true

jc1144096387:
## Changes

- What does this change?
I add ```shamefully-hoist=false``` into the npm config(.npmrc) of the project. Because if shamefully-hoist is true in user global config(.npmrc), build will throw @astrojs/node:build: src/response-iterator.ts(67,3): error TS2578: Unused '@ts-expect-error' directive.
When shamefully-hoist is true, pnpm will put @types/web into the root of node_modules, and so typescript will find the interface ReadableStreamDefaultReader, and then  '@ts-expect-error' directive will be unused.I run `pnpm run build` and get the error below.
ts-error
build-error

## Testing

I add ```shamefully-hoist=false``` into the root npm config(.npmrc). Then I run ```pnpm install``` and ```pnpm run build```. Building successfully.


bluwy:
I don't think we need this. `shamefully-hoist=false` is already the default for pnpm, and it's rare to set this globally as it conflicts with one of pnpm's core motive.

jc1144096387:
> I don't think we need this. `shamefully-hoist=false` is already the default for pnpm, and it's rare to set this globally as it conflicts with one of pnpm's core motive.

Thank you very much to reply me. You are right. Because I am not familiar with pnpm, so I set shamefully-hoist true globally before when I install some incompatible dependency.
I will close this pr, thank you again.

Issue #7538, opened 2023-06-30
[astrojs/cloudflare] use Astro.locals for env variable


alexanderniebuhr:
- updating cloudflare so that it's "env" is available on Astro.locals
- should be available in pages via Astro.locals.env
- Which means you don't have to do getRuntime() any more

### Participation

- [X] I am willing to submit a pull request for this issue.

alexanderniebuhr:
👋 i'm working on this

alexanderniebuhr:
just figured that we need to expose runtime in addition to keep feature parity with `getRuntime()`

PR #7541, opened 2023-06-30
feat(@astrojs/cloudfalre): add cloudflare envs to `Astro.locals`

alexanderniebuhr:
## Changes

- Add the Cloudflare Runtime Variables as `Astro.locals.runtime`
- Add deprecation comment in code, to make sure we remove `getRuntime()` future release
- Add migration info in changeset
- Update Docs
- closes https://github.com/withastro/astro/issues/7538

## Testing

- manually tested with Cloudflare Pages Deployments.
- duplicated integration test, which was used for `getRuntime`

## Docs

## Notes:

- It does not export a type yet. Users have to type the `locals` themself.

### DIFF
--- a/.changeset/fuzzy-jobs-explain.md
+++ b/.changeset/fuzzy-jobs-explain.md
@@ -4,6 +4,10 @@
 
 Expose the runtime bindings & environmental variabels via `Astro.locals`
 
-  ```js
-  const env = Astro.locals.runtime.env;
+  ```diff
+  - import { getRuntime } from '@astrojs/cloudflare/runtime';
+  - getRuntime(Astro.request);
+  
+  + const runtime = Astro.locals.runtime;
+  + const env = runtime.env;
   ```
--- a/.changeset/fuzzy-jobs-explain.md
+++ b/.changeset/fuzzy-jobs-explain.md
@@ -2,12 +2,11 @@
 '@astrojs/cloudflare': minor
 ---
 
-Expose the runtime bindings & environmental variabels via `Astro.locals`
+The `getRuntime` utility has been deprecated and should be updated to the new [`Astro.locals`](https://docs.astro.build/en/guides/middleware/#locals) API.
 
-  ```diff
-  - import { getRuntime } from '@astrojs/cloudflare/runtime';
-  - getRuntime(Astro.request);
+```diff
+- import { getRuntime } from '@astrojs/cloudflare/runtime';
+- getRuntime(Astro.request);
   
-  + const runtime = Astro.locals.runtime;
-  + const env = runtime.env;
-  ```
++ const runtime = Astro.locals.runtime;
+```
--- a/.changeset/fuzzy-jobs-explain.md
+++ b/.changeset/fuzzy-jobs-explain.md
@@ -1,0 +1,9 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Expose the runtime bindings & environmental variabels via `Astro.locals`
+
+  ```js
+  const env = Astro.locals.runtime.env;
+  ```
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -89,11 +89,7 @@ export function createAPIContext({
 			if (typeof val !== 'object') {
 				throw new AstroError(AstroErrorData.LocalsNotAnObject);
 			} else {
-				if (Reflect.has(request, 'runtime') && !Reflect.has(val, 'runtime')) {
-					throw new AstroError(AstroErrorData.LocalsNoRuntime);
-				} else {
-					Reflect.set(request, clientLocalsSymbol, val);
-				}
+				Reflect.set(request, clientLocalsSymbol, val);
 			}
 		},
 	});
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -89,7 +89,11 @@ export function createAPIContext({
 			if (typeof val !== 'object') {
 				throw new AstroError(AstroErrorData.LocalsNotAnObject);
 			} else {
-				Reflect.set(request, clientLocalsSymbol, val);
+				if (Reflect.has(request, 'runtime') && !Reflect.has(val, 'runtime')) {
+					throw new AstroError(AstroErrorData.LocalsNoRuntime);
+				} else {
+					Reflect.set(request, clientLocalsSymbol, val);
+				}
 			}
 		},
 	});
--- a/packages/integrations/cloudflare/src/runtime.ts
+++ b/packages/integrations/cloudflare/src/runtime.ts
@@ -1,3 +1,4 @@
+// TODO: remove `getRuntime()` in Astro 3.0
 import type { Cache, CacheStorage, IncomingRequestCfProperties } from '@cloudflare/workers-types';
 
 export type WorkerRuntime<T = unknown> = {
@@ -21,6 +22,16 @@ export type PagesRuntime<T = unknown, U = unknown> = {
 	cf?: IncomingRequestCfProperties;
 };
 
+/**
+ * @deprecated since version 2.0
+ * The `getRuntime` utility has been deprecated and should be updated to the new [`Astro.locals`](https://docs.astro.build/en/guides/middleware/#locals) API.
+ * ```diff
+ * - import { getRuntime } from '@astrojs/cloudflare/runtime';
+ * - getRuntime(Astro.request);
+ *  
+ * + const runtime = Astro.locals.runtime;
+ * ```
+ */
 export function getRuntime<T = unknown, U = unknown>(
 	request: Request
 ): WorkerRuntime<T> | PagesRuntime<T, U> {
--- a/packages/integrations/cloudflare/src/runtime.ts
+++ b/packages/integrations/cloudflare/src/runtime.ts
@@ -23,7 +23,7 @@ export type PagesRuntime<T = unknown, U = unknown> = {
 };
 
 /**
- * @deprecated since version 2.0
+ * @deprecated since version 6.8.0
  * The `getRuntime` utility has been deprecated and should be updated to the new [`Astro.locals`](https://docs.astro.build/en/guides/middleware/#locals) API.
  * ```diff
  * - import { getRuntime } from '@astrojs/cloudflare/runtime';
--- a/packages/integrations/cloudflare/src/server.advanced.ts
+++ b/packages/integrations/cloudflare/src/server.advanced.ts
@@ -12,15 +12,13 @@ type Env = {
 	name: string;
 };
 
-interface CloudflareContext {
-	locals: {
+interface WorkerRuntime {
 		runtime: {
 			waitUntil: (promise: Promise<any>) => void;
 			env: Env;
 			cf: CFRequest['cf'];
 			caches: typeof caches;
 		};
-	};
 }
 
 export function createExports(manifest: SSRManifest) {
@@ -46,7 +44,7 @@ export function createExports(manifest: SSRManifest) {
 				request.headers.get('cf-connecting-ip')
 			);
 
-			// @deprecated: getRuntime() can be removed in the next major release, after testing
+			// @deprecated: getRuntime() can be removed in the next major release, after testing runtime in locals
 			Reflect.set(request, Symbol.for('runtime'), {
 				env,
 				name: 'cloudflare',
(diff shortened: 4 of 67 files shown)


matthewp:
Hey @alexanderniebuhr 👋 this is looking good, but is still in draft. Are you ready for people to review it?

alexanderniebuhr:
> Are you ready for people to review it?

@matthewp soon! Need to fix some types and double check a test! Will ping you here & in Discord :)

packages/astro/src/core/endpoint/index.ts
```diff
@@ -91,7 +91,11 @@ export function createAPIContext({
 			if (typeof val !== 'object') {
 				throw new AstroError(AstroErrorData.LocalsNotAnObject);
 			} else {
-				Reflect.set(request, clientLocalsSymbol, val);
+				if (Reflect.has(request, 'runtime') && !Reflect.has(val, 'runtime')) {
+					throw new AstroError(AstroErrorData.LocalsNoRuntime);
+				} else {
+					Reflect.set(request, clientLocalsSymbol, val);
+				}
```
alexanderniebuhr, line 98: This should prevent users to overwrite `locals.runtime` if it exists.

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -42,7 +44,20 @@ export function createExports(manifest: SSRManifest) {
...
-			let response = await app.render(request, routeData);
+
+			let response = await app.render(request, routeData, {
+				runtime: {
+					// request: Request; // not needed because we have Astro.request, thus they are minor differences
+					// functionPath: string; // not needed
+					waitUntil: (promise: Promise<any>) => {
+						context.waitUntil(promise);
+					},
+					// passThroughOnException: () => void; // probably not needed ??
```
alexanderniebuhr, line 55: Should we include it anyway?

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -10,25 +10,16 @@ if (!isNode) {
...
 
 		const { pathname } = new URL(request.url);
 		// static assets fallback, in case default _routes.json is not used
 		if (manifest.assets.has(pathname)) {
-			// we need this so the page does not error
-			// https://developers.cloudflare.com/pages/platform/functions/advanced-mode/#set-up-a-function
-			return (runtimeEnv.env as EventContext<unknown, string, unknown>['env']).ASSETS.fetch(
-				request
-			);
+			return env.ASSETS.fetch(request)
```
alexanderniebuhr, line 22: This is actually a bug fix, which otherwise breaks CF. See `[server.advanced.ts](https://github.com/withastro/astro/pull/7541/files#diff-854537ee160391dda208e76dbea4c1471189e0368d2d76537485a0039d103da3)`, where this fix was already introduced weeks ago.

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -38,17 +29,35 @@ export function createExports(manifest: SSRManifest) {
...
+
+			let response = await app.render(request, routeData, {
+				env: env,
+				cf: request.cf,
+				runtime: {
+					// request: Request; // not needed because we have Astro.request, thus they are minor differences
+					// functionPath: string; // not needed
+					waitUntil: (promise: Promise<any>) => { context.waitUntil(promise); },
+					// passThroughOnException: () => void; // probably not needed ??
+					// next: (input?: Request | string, init?: RequestInit) => Promise<Response>; // probably not needed ??
```
alexanderniebuhr, line 53: Should we include them anyways, I can't think of a use-case but other users might have some.

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -38,17 +29,35 @@ export function createExports(manifest: SSRManifest) {
...
+					env: context.env,
+					params: context.params, // Isn't that the same as Astro.props ??
```
alexanderniebuhr, line 55: Are Astro.props 1-1 equally? Haven't found anything.

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -38,17 +29,35 @@ export function createExports(manifest: SSRManifest) {
...
+					// data: Data; // Should we include it ??: https://community.cloudflare.com/t/what-is-context-data-in-pages-functions/476559/7
```
alexanderniebuhr, line 56: What is your opinion, should we include it for users who want to use it, I can't see any use-case.

alexanderniebuhr:
@matthewp I'm ready to get initial reviews :)

packages/astro/src/core/endpoint/index.ts
ematipico, line 98: This should be in core. Astro doesn't know anything about "runtime" and it should stay that way. We should move this code in the adapter.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -657,6 +657,20 @@ See https://docs.astro.build/en/guides/server-side-rendering/ for more informati
...
+	 * Thrown when `locals` include runtime envs, but is overwritten with something that does not include them
+	 *
+	 * ```
+	 */
+	LocalsNoRuntime: {
+		title: 'Value assigned to `locals` is not accepted.',
+		message:
+			'`locals` include runtime information. Overwriting them will make them unaccissable.',
+		hint: 'If you tried to remove some information from the `locals` object, try to use `delete` or set the property to `undefined`.',
+	},
```
ematipico, line 673: We should remove this error, because it's not part of core.

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -10,25 +10,16 @@ if (!isNode) {
...
-		request: Request & CFRequest;
-		next: (request: Request) => void;
-		waitUntil: EventContext<unknown, any, unknown>['waitUntil'];
-	} & Record<string, unknown>) => {
-		process.env = runtimeEnv.env as any;
+	const onRequest = async (context: EventContext<unknown, string, unknown>) => {
+		const request = context.request as CFRequest & Request
+		const { next, env } = context
+
+		process.env = env as any; // would love to remove this any cast in the future
```
ematipico, line 17: It would be beneficial to add a comment that explains **why** the cast to `any`, like this your comment would give some context to other developers.

packages/integrations/cloudflare/src/server.directory.ts
ematipico: Usually it's best to pass a tiny set of APIs, and then include new ones on the go, if users need them.

packages/integrations/cloudflare/src/server.directory.ts
ematipico: No. `params` come from the URL the page. `props` belongs to a component.

packages/integrations/cloudflare/src/server.directory.ts
ematipico: We can include them **if** users need it. For new, we can hide them and add them as we go if users request them.

ematipico (changes requested):

packages/integrations/cloudflare/src/server.directory.ts
alexanderniebuhr: My bad, I meant [Astro.params](https://docs.astro.build/en/reference/api-reference/#astroparams)

packages/astro/src/core/endpoint/index.ts
alexanderniebuhr: @ematipico in [Discord](https://discord.com/channels/830184174198718474/1124318874175819847/1126171117309136899) you said, we should make sure that `locals` are not overwritten by a middleware function. How would that be possible if the check is happening in the adapter?

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -12,6 +12,17 @@ type Env = {
...
+interface CloudflareContext {
+	locals: {
+		runtime: {
+			waitUntil: (promise: Promise<any>) => void;
+			env: Env;
+			cf: CFRequest['cf'];
+			caches: typeof caches;
+		};
+	};
+}
```
alexanderniebuhr, line 24: this also prepares for a future, where we export an interface for users to extend in their locals type definition.

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -47,17 +58,32 @@ export function createExports(manifest: SSRManifest) {
...
-					env: env,
-					cf: request.cf,
-					caches: caches,
 				},
 			});
 
+			// We should freeze the runtime object, to make sure it&nested properties are not mutated
+			Object.freeze(cloudflareContext.locals.runtime);
+
+			let response = await app.render(request, routeData, cloudflareContext.locals);
```
alexanderniebuhr: this is a little different implemented than in the reference, but AFAIK this actually puts all the protections in place. Since using the adapter, locals now has a value all the time, the user should not ever write to `context.locals`!, even not in object.

alexanderniebuhr: @ematipico let me know if that thought makes actually sense or not.

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -42,7 +57,32 @@ export function createExports(manifest: SSRManifest) {
...
+						},
+						env: env,
+						cf: request.cf,
+						caches: caches,
+					},
+				},
+			});
+
+			// We should freeze the runtime object, to make sure it&nested properties are not mutated
+			Object.freeze(cloudflareContext.locals.runtime);
```
ematipico, line 83: While it makes sense to use `Object.freeze`, it provide really **poor DX**. You'll get something like:

```
// TypeError: Cannot freeze ...
```

Which doesn't give info about *why it happened* and *how to fix it*. We should create an object with `Object.create` and throw errors using the `set()` function.

ematipico (commented):
It would be great if we can add some tests, where we have a middleware.

Inside the middleware, should do something like:
- `context.locals.user.name = "Nick"` -> no errors, it should work
- `context.locals.runtime = {}` -> runtime error, can't override
- `context.locals = {}`, allowed.. no runtime errors but log a warning to the user that wiping out `locals` is unsafe

packages/integrations/cloudflare/src/server.advanced.ts
alexanderniebuhr: I agree, but isn't the issue with using a `set()` function, that it will only protect the exact property _(in this case `runtime`)_ and not any nested once _(e.g. `runtime.env`)_?

ematipico: Yes. In that case you want to manually create each property:

```js
const locals = Object.create({}, {
	env: {
		enumerable: true,
		value: env,
		set() {
			throw new Error()
		}
	},
	waitUntil: {
		enumerable: true,
		value: /* value */,
		set() {
			throw new Error()
		}
	}
});
```

It's a bit lengthy, but I think it's the best solution to offer a better DX

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -42,7 +57,32 @@ export function createExports(manifest: SSRManifest) {
...
-			let response = await app.render(request, routeData);
+
+			const cloudflareContext = {} as CloudflareContext;
+
+			// We define a custom property, so we can check the value passed to locals
+			Object.defineProperty(cloudflareContext, 'locals', {
+				enumerable: true,
+				// we should protect top-level locals from being overwritten
+				// users should just use nested properties, e.g. locals.test = "foo"
+				writable: false,
```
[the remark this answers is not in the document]
alexanderniebuhr: As discussed via Discord, we will only show a warning if users try to overwrite locals with `locals = {}`. Anything other warning or protection is to be added in the future if needed.

sarah11918 (approved):
Docs approves the changeset and README! 🥳

lilnasy:
Would `Astro.locals` in the frontmatter have identical functionality to `context.locals` in the middleware?

For context, trying to decide on a different API for #7975, and this PR seems related.

alexanderniebuhr:
> Would `Astro.locals` in the frontmatter have identical functionality to `context.locals` in the middleware?

AFAIK, no. This just passes the cf runtime to `locals` using the `app.render()` function.

ematipico (approved):
Almost there! I think we're good. Just one more thing: could you add a `TODO` comment to the `getRuntime` function as a reminder to remove it in the next major release of the library??

lilnasy:
I would like to establish a convention for adapters introducing runtime controls through locals.

For #7975, I went with `Astro.locals.netlify.setBuildersTtl()` because I wanted it to be immediately clear that you might need to revisit it if you switch providers. But, I am not opposed to `Astro.locals.runtime.setBuildersTtl()`.

What I consider important is that there is consistency.

alexanderniebuhr:
> I would like there to be a convention for adapters introducing runtime controls through locals.

I am also fine with both. However I think this needs some RFC to align across everything. I really want this PR to get merged soon, so not sure if we can wait for the RFC. Any opinions cc @ematipico @matthewp 

- `Astro.locals.<PROVIDER>.<XYZ>`
- `Astro.locals.runtime.<XYZ>`

matthewp:
lgtm! No opinion on namespacing from me, integrations can essentially do what they want here, I don't think an RFC is valid for this sort of thing. Adapters can coordinate if they'd like / think it makes things easier. I'm not sure if it does though, but that's my opinion.

.changeset/fuzzy-jobs-explain.md
```diff
@@ -0,0 +1,13 @@
...
+
+Expose the runtime bindings & environmental variabels via `Astro.locals`
+
+  ```diff
+  - import { getRuntime } from '@astrojs/cloudflare/runtime';
+  - getRuntime(Astro.request);
+  
+  + const runtime = Astro.locals.runtime;
+  + const env = runtime.env;
+  ```
```
natemoo-re, line 13:
```suggestion

The `getRuntime` utility has been deprecated and should be updated to the new [`Astro.locals`](https://docs.astro.build/en/guides/middleware/#locals) API.

```diff
- import { getRuntime } from '@astrojs/cloudflare/runtime';
- getRuntime(Astro.request);
  
+ const runtime = Astro.locals.runtime;
+ const env = runtime.env;
```
```

**Review by natemoo-re — APPROVED** [review]
Looks great! 

One request from me: could we mark the `getRuntime` export as `/** @deprecated */` so that users get this feedback in their IDE?

**Comment by lilnasy** [comment]
In that case, I'll go with locals.runtime because I'd only have to change one line.

File: packages/integrations/cloudflare/src/runtime.ts [hunk 14]
```diff
@@ -21,6 +22,16 @@ export type PagesRuntime<T = unknown, U = unknown> = {
 	cf?: IncomingRequestCfProperties;
 };
 
+/**
+ * @deprecated since version 2.0
```
natemoo-re, line 26:
```suggestion
 * @deprecated since version 6.8.0
```

natemoo-re (approved):

Issue #7460, opened 2023-06-23
Over 10k `setTimeout()` on Route Rendering


itsmatteomanf:
### What version of `astro` are you using?

2.7.0

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Safari

### Describe the Bug

When publishing to Pages, which has a 10k `setTimeout()` limit, I'm hitting this limit when rendering ~3.5k components.

```json
 {
	"name": "QuotaExceededError",
	"message": "You have exceeded the number of timeouts you may set.10000",
	"timestamp": 1687523060530
}
```

The issue seems to be related to this function, which is from Astro's handling of components. None of these components handle any actual data fetching, just rendering HTML.

```js
function $a(e) {
	let t = e.map((n) => new $t(n));
	return (
		setTimeout(() => {
			t.forEach((n) => !n.isStarted() && n.buffer());
		}, 0),
		t
	);
}
```

I can't really provide a minimal example, it needs to be deployed on Pages and you need to look at their logs to see the error (apart from the page failing to render).

### What's the expected result?

The page to load.

### Link to Minimal Reproducible Example

null

### Participation

- [x] I am willing to submit a pull request for this issue.

bluwy:
The code comes from https://github.com/withastro/astro/blob/dfa70357d3ce936995e0dfaf4336a2bc1e9b8136/packages/astro/src/runtime/server/render/util.ts#L147-L157

This seems to be a really niche case though. Can you show how you're rendering 3.5k components? Are they scattered or rendered side-by-side? cc @johannesspohr

itsmatteomanf:
It's a grid of components.

```js
{data.map((item) => (
	<FilterItem
		href={`/page/${item.url}/`}
		string={item.name}
	/>
))}
```

Can I create them from within the main page? Sure. Is it ideal? No. This component is used in tons of pages.
I could make the component render multiple children itself, probably, but it would still be kind of the opposite of what you would expect the code to look like.

johannesspohr:
I guess a lot of the `setTimeout` are called in the same synchronous execution, so we could probably reduce the amount of setTimeouts a lot by batching those calls as well. I won't be able to try this before next week though (but of course anyone is welcome to give it a go). Other than that, I don't know if some replacement (like `setImmediate`) is available on workers, it seems like it's not.

itsmatteomanf:
[...]

PR #7494, opened 2023-06-27
Update `setTimeout()` to `queueMicrotask()`

itsmatteomanf:
## Changes

Replaces the instance of `setTimeout()` in the runtime to use `queueMicrotask()`, to resolve limitations on Cloudflare Workers.

Solves https://github.com/withastro/astro/issues/7460.

## Testing

No tests were added per https://github.com/withastro/astro/issues/7460#issuecomment-1606954052.

## Docs

No changes in documentation is required, no user-facing changes nor different behaviour is expected.

PR #7544, opened 2023-06-30
Batch iterator buffering to reduce number of timeout calls

johannesspohr:
## Changes

- What does this change?
This batches iterator buffering to solve #7460

## Testing

Difficult due to inability to count the number of cals to `setTimeout`. Will be tackled in a different PR.

## Docs

Changeset is included

### DIFF
--- a/.changeset/chilled-books-shave.md
+++ b/.changeset/chilled-books-shave.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Batch async iterator buffering to reduce numbers of calls to `setTimeout`
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -139,6 +139,26 @@ export function renderElement(
 	return `<${name}${internalSpreadAttributes(props, shouldEscape)}>${children}</${name}>`;
 }
 
+const iteratorQueue: EagerAsyncIterableIterator[][] = [];
+
+/**
+ * Takes an array of iterators and adds them to a list of iterators to start buffering
+ * as soon as the execution flow is suspended for the first time. We expect a lot
+ * of calls to this function before the first suspension, so to reduce the number
+ * of calls to setTimeout we batch the buffering calls.
+ * @param iterators
+ */
+function queueIteratorBuffers(iterators: EagerAsyncIterableIterator[]) {
+	if (iteratorQueue.length === 0) {
+		setTimeout(() => {
+			// buffer all iterators that haven't started yet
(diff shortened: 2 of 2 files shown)


natemoo-re:
!bench

itsmatteomanf:
I run the same benchmarks against my own PR which added the `queueMicrotask()` implementation, everything here seems either equal or, more often, better than that case.

https://github.com/withastro/astro/pull/7494

To me this looks good. Might be worth pushing soon, to catch the next release to revert the performance regression.

bluwy:
I think this PR is in a ready state as the benchmark are sort-of the test, and it does show a big perf improvement! Rendering `.astro` files take avg 1.6ms from 4.3ms.

itsmatteomanf:
It had to get worse to get better, but it might be a nice performance improvement. Once it’s actually released I’ll compare with my massive page that caused the issues.
v2.7.1 vs. v2.7.2 vs. this one.

johannesspohr:
Thanks for the feedback everybody, I'd like to have a proper test that everything works as expected, but I didn't find a good solution to spy on `setTimeout`. Any ideas how to test this? (I guess we could import / inject setTimeout somehow to replace it with a counting function)

I also threw in a couple of good ol' `console.log`s and saw the number of setTimeouts drop from 14 to 7 with the new version. So I assume it works, but it's really hard to really track what's going on and how many calls it should be.

itsmatteomanf:
> but it's really hard to really track what's going on and how many calls it should be.

Yeah, I had like 3k components which triggered over 10k, not sure how, why or when. I traced it to that function, but actually counting them was *hard*. To me a reduction sounds good, as long as the actual rendering works correctly, like it should.

`queueMicrotask()` is slower than `setTimeout()`, that we figured out, and I'm not knowledgeable enough on Astro's codebase to say if there is a better way. I'd assume not, but improvements can always be done, usually ahah

> I'd like to have a proper test that everything works as expected

What are ou worried about not working?

johannesspohr:
I mostly wanted to verify that the batching is in fact happening and make this expectation explicit with a test, so future changes won't break it (as it's kind of implicit at the moment). So if everybody is OK with it we can merge it for now, and try adding some regression tests in a follow-up PR.

itsmatteomanf:
Fine by me, I feel like at the very least resolving the perf regression is a nice thing. This should bring performance even higher than it was. Testing wise I’d have to think about it.

bluwy (approved):
I'm still a bit worried having a timing-based test on this. When we added the parallel rendering timing tests, it became flaky now and then that we had to bump the timeout. But if you can figure out a way that doesn't cause flakiness, that would be great!

PR #7555, opened 2023-07-03
fix(astrojs/cloudflare): SSR split file renaming misses ts endpoints

alexanderniebuhr:
## Changes

- fixed an bug, where `ts` endpoint files are not renamed to .js after bundle step
## Testing

- updated test suite
- manually

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -133,12 +133,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						})
 					)
 
+					console.log(outputFiles)
+
 					// loop through all new bundled files and write them to the functions folder
 					for (const outputFile of outputFiles) {
 
 						// split the path into an array
 						const path = outputFile.split('/');
-
+						console.log(path)
 						// replace dynamic path with [path]
 						const pathWithDynamics = path.map((segment) => segment.replace(/(\_)(\w+)(\_)/g, (_, __, prop) => {
 							return `[${prop}]`;
@@ -166,7 +168,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 						const oldFileUrl = new URL(`$astro/${outputFile}`, outputUrl);
 						const newFileUrl = new URL(finalPath, functionsUrl);
-
+						console.log(oldFileUrl, newFileUrl)
 						await fs.promises.rename(oldFileUrl, newFileUrl);
 					}
 				} else {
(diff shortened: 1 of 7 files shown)


alexanderniebuhr:
cc @matthewp @ematipico

packages/integrations/cloudflare/src/index.ts
```diff
@@ -138,6 +138,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						const fileName = entryPointsRouteData[index].component
 							.replace('src/pages/', '')
 							.replace('.astro', '.js')
+							.replace('.ts', '.js')
```
ematipico, line 141: This logic is not very robust, because a page in Astro can have many extensions: js, ts, mjs, cjs, md, mdx, markdoc, etc

Maybe we should review the logic

ematipico (changes requested):

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: That's true, I'll refactor that logic to be more robust

alexanderniebuhr: @ematipico updated the logic to make it more robust. currently it has to a lot of memory allocations for clear code. let me know if I should consolidate all the steps into one `.map()`?

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -24,7 +24,13 @@ export function createExports(manifest: SSRManifest) {
 		const { pathname } = new URL(request.url);
 		// static assets fallback, in case default _routes.json is not used
 		if (manifest.assets.has(pathname)) {
-			return next(request);
+			// we need this so the page does not error
+			// https://developers.cloudflare.com/pages/platform/functions/advanced-mode/#set-up-a-function
+			return (runtimeEnv.env as unknown & {
+				ASSETS: {
+					fetch: typeof fetch;
+				};
```
alexanderniebuhr, line 32: this is already done for `advanced` mode but was missing for `directroy` mode. please also note that I'll remove the type assertion on the `Astro.locals` PR, where I refactored the `EventContext` typing

packages/integrations/cloudflare/src/index.ts
```diff
@@ -126,28 +125,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+						const finalDirPath = finalSegments.slice(0, -1).join('/');
+						const finalPath = finalSegments.join('/');
+
+						const newDirUrl = new URL(finalDirPath, functionsUrl);
+						await fs.promises.mkdir(newDirUrl, { recursive: true })
+
+						const oldFileUrl = new URL(`$astro/${outputFile}`, outputUrl);
+						const newFileUrl = new URL(finalPath, functionsUrl);
+						await fs.promises.rename(oldFileUrl, newFileUrl);
 					}
```
alexanderniebuhr, line 171: This loop has a lot of memory allocation, let me know if I should consolidate the steps into more abstract steps?
@ematipico @matthewp

alexanderniebuhr:
had to redo the PR, due to git history conflicts: https://github.com/withastro/astro/pull/7568

Issue #7529, opened 2023-06-30
CSS HMR is broken on 2.7.2


jojojojojoj5564656465465:
### What version of `astro` are you using?

2.7.2

### Are you using an SSR adapter? If so, which one?

no

### What package manager are you using?

pnpm

### What operating system are you using?

linux

### What browser are you using?

edge, vivaldi

### Describe the Bug

When i am in dev mode, the hot reload is not working for css. It's works for html but not on css even if it's tailwind class or css in file.

### What's the expected result?

i want to be abble to use dev mode and have css compile along the html

### Link to Minimal Reproducible Example

https://stackblitz.com/github.com/jojojojojoj5564656465465/astro-bug2

### Participation

- [X] I am willing to submit a pull request for this issue.

Dunqing:
There is the same problem in `UnoCSS`, a related issue in #2812.

I did some research and guessed that it was broken in https://github.com/withastro/astro/pull/7381. Hope it helps `astro` to fix this.

delucis:
Noting that we’ve also seen reports of this with Astro v2.7.1 so may have been introduced before 2.7.2.

jojojojojoj5564656465465:
I used my unocss config on qwik and didn't get those bugs so I don't blame unocss

delucis:
I’ve definitely seen similar reports for a range of set-ups with/without UnoCSS, so yeah, don’t think that’s the issue here. I’d recommend anyone who’s hitting this downgrades to astro v2.7.0 for now:

```
npm i astro@2.7.0
```

jojojojojoj5564656465465:
> I’ve definitely seen similar reports for a range of set-ups with/without UnoCSS, so yeah, don’t think that’s the issue here. I’d recommend anyone who’s hitting this downgrades to astro v2.7.0 for now:
> 
> ```shell
> npm i astro@2.7.0
> ```

I used 2.7.0 and it 's working perfectly

gjaynes-squiz:
FYI, I'm getting this too with Tailwind CSS, I'm not using UnoCSS. 

It's particularly bad when saving `.mdx` file inside a content collection. 

I downgraded to 2.7.0 and it seems to fix the issue.

florivdg:
> It's particularly bad when saving `.mdx` file inside a content collection.

I can definitely replicate that! I'm using Tailwind CSS and .mdx files, but this combination is proving to be not working as well.

PR #7565, opened 2023-07-04
Fix style crawling logic for CSS HMR

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7529

Used the not-so-performant technique initially used in the first commit of https://github.com/withastro/astro/pull/7381 to fix the issue for now.

The issue seems to be that UnoCSS is invalidating the Astro files at https://github.com/unocss/unocss/blob/757b7f18b060f32387689545c878aea1e63203b4/packages/vite/src/modes/global/dev.ts#L59. This breaks our expectations where they are already loaded and exist as `ssrTransformResult`. The invalidation clears out `ssrTransformResult`, causing the issue.

A fix I intially thought is to force `loader.import()` on all modules while we're traversing the graph to counteract UnoCSS, but that feels more non-performant than the initial PR idea, so I went with the initial PR idea instead.

## Testing

Existing tests should pass. Also tested the repro locally. Unfortunately time is a bit tight for me now to add a real test with unocss.

## Docs

n/a.

### DIFF
--- a/.changeset/red-turtles-drum.md
+++ b/.changeset/red-turtles-drum.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix style crawling logic for CSS HMR
--- a/packages/astro/src/core/module-loader/loader.ts
+++ b/packages/astro/src/core/module-loader/loader.ts
@@ -48,6 +48,7 @@ export interface ModuleNode {
 	} | null;
 	ssrError: Error | null;
 	importedModules: Set<ModuleNode>;
+	importers: Set<ModuleNode>
 }
 
 export interface ModuleInfo {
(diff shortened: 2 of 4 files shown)


ematipico (approved):

duxphp:
[unocss-hmr] failed to get unocss hash, hmr might not work

The error is still reported when using unocss

bluwy:
Can you provide a repro where that's happening?

marvin-j97:
> Can you provide a repro where that's happening?

https://stackblitz.com/edit/withastro-astro-yrtpg7?file=package.json

bluwy:
Thanks for the repro. Seems like it's a browser warning log, not an error. And it's also happening in 2.7.0, which is out of scope of this PR as it's fixing a regression from 2.7.1.

IIUC this needs to be fix upstream at https://github.com/unocss/unocss/blob/757b7f18b060f32387689545c878aea1e63203b4/packages/vite/src/modes/global/dev.ts#L187C6-L198, skipping the warning if `__vite_css__` is an empty string.

PR #7568, opened 2023-07-05
fix(@astrojs/cloudflare): SSR split file renaming misses ts endpoints

alexanderniebuhr:
## Changes

- fixed an bug, where `ts` endpoint files are not renamed to .js after bundle step
## Testing

- updated test suite
- manually

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -133,12 +133,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						})
 					)
 
+					console.log(outputFiles)
+
 					// loop through all new bundled files and write them to the functions folder
 					for (const outputFile of outputFiles) {
 
 						// split the path into an array
 						const path = outputFile.split('/');
-
+						console.log(path)
 						// replace dynamic path with [path]
 						const pathWithDynamics = path.map((segment) => segment.replace(/(\_)(\w+)(\_)/g, (_, __, prop) => {
 							return `[${prop}]`;
@@ -166,7 +168,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 						const oldFileUrl = new URL(`$astro/${outputFile}`, outputUrl);
 						const newFileUrl = new URL(finalPath, functionsUrl);
-
+						console.log(oldFileUrl, newFileUrl)
 						await fs.promises.rename(oldFileUrl, newFileUrl);
 					}
 				} else {
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -3,6 +3,7 @@ import type { AstroAdapter, AstroConfig, AstroIntegration, RouteData } from 'ast
 import esbuild from 'esbuild';
 import * as fs from 'fs';
 import * as os from 'os';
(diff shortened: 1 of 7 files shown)


packages/integrations/cloudflare/src/index.ts
```diff
@@ -126,28 +125,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+						const finalDirPath = finalSegments.slice(0, -1).join('/');
+						const finalPath = finalSegments.join('/');
+
+						const newDirUrl = new URL(finalDirPath, functionsUrl);
+						await fs.promises.mkdir(newDirUrl, { recursive: true })
+
+						const oldFileUrl = new URL(`$astro/${outputFile}`, outputUrl);
+						const newFileUrl = new URL(finalPath, functionsUrl);
+						await fs.promises.rename(oldFileUrl, newFileUrl);
 					}
```
alexanderniebuhr, line 171: this loop adds a lot of memory allocation, should I refactor it to be more consolidated?
@ematipico @matthewp

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -24,7 +24,13 @@ export function createExports(manifest: SSRManifest) {
...
 		// static assets fallback, in case default _routes.json is not used
 		if (manifest.assets.has(pathname)) {
-			return next(request);
+			// we need this so the page does not error
+			// https://developers.cloudflare.com/pages/platform/functions/advanced-mode/#set-up-a-function
+			return (runtimeEnv.env as unknown & {
+				ASSETS: {
+					fetch: typeof fetch;
+				};
+			}).ASSETS.fetch(request);
```
alexanderniebuhr, line 33: this change was already done in `advanced` mode, but missing in `directory` mode. So I did this. 
Note I will also get rid of the type assertion once I finish the PR for runtime in `Astro.locals`

alexanderniebuhr (commented):
just some notes, for better understanding

packages/integrations/cloudflare/src/index.ts
ematipico: What do you think it adds a lot of memory? Also, I am unfamiliar with Cloudflare, so I ask you a question in return: what does this operation do?

packages/integrations/cloudflare/src/server.directory.ts
ematipico: The example in TypeScript shows that we can type `ASSETS` like this:

```
interface Env {
  ASSETS: Fetcher;
}
```

Can we do the same here?

alexanderniebuhr: Will get the type of `EnventContext` itself, which is coming from `@cloudflare/workers-types``

alexanderniebuhr: Will get the type of `EnventContext` itself, which is coming from `@cloudflare/workers-types``

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: took discussion onto discord: https://discord.com/channels/830184174198718474/1126159202725924877

ematipico (approved):

PR #7569, opened 2023-07-05
Added preact-prepass-render to @astrojs/preact

aleksandrjet:
## Changes

Added `preact-ssr-prepass` library to preact integraiont. This library allow to use lazy components in preact code. Without this i get unknown error in SSR.

While execute SSR `preact-ssr-prepass` will wait to load all dynamic imports so that `preact-render-to-string` can use they. More details about [preact-ssr-prepass](https://github.com/preactjs/preact-ssr-prepass)

## Testing

Added `examples/lazy-preact` and `preact-lazy-component.test` with lazy loading of component.

## Docs

I don't think that it needs documentation, because i expected that Suspense and lazy functions will work. 

It does not affect the current api, but it can slow build execution of preact components, because `preact-ssr-prepass` will add extra pass of tree. This pass will wait for all dynamic imports to be loaded, even if they don't exist in the code.


.changeset/big-pets-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/preact': patch
+---
+
+Added preact-ssr-prepass library to preact integraiont
```
sarah11918, line 5:
```suggestion
Suspense on the server with Preact

Adds `preact-ssr-prepass` library to the official Preact integration.
```
If we're looking more closely at changesets that might become part of the automated blog post, we could do something like this for a header. If not, just use the corrections to the second sentence!  :smile:

sarah11918 (commented):
Agreed that I don't think the Preact integration README needs updating for this, as it was never explicitly stated that you *couldn't* do this before, and we're just adding some functionality under the hood.

aleksandrjet:
do I need to do something else? I see that not all checks passed, but I can't restart them

natemoo-re:
> do I need to do something else? I see that not all checks passed, but I can't restart them

No worries @aleksandrjet, we just have a couple of flaky CI tests that we're trying to track down! We're working through our backlog of PRs right now, hoping to give this a proper review soon!

natemoo-re:
Also I reached out to Marvin on Twitter about this one to check if `preact-ssr-prepass` is still the official recommendation from the Preact team. Hopefully it is!

https://twitter.com/n_moore/status/1678825870364930050

matthewp:
@natemoo-re did you ever hear back about this?

PR #7571, opened 2023-07-05
refactor: use only manifest to compute information

ematipico:
## Changes

This is the first step towards a unified way to "render a page or an endpoint".

In this PR, I refactored the generation of the page using the `SSRManifest`. The manifest has already ALL the information needed to render a page.

In this PR we create a `SSRManifest` using the settings, the configuration and build internals. 

## Testing

Current CI should pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/app/types.ts
+++ b/packages/astro/src/core/app/types.ts
@@ -37,7 +37,7 @@ export type SSRManifest = {
 	adapterName: string;
 	routes: RouteInfo[];
 	site?: string;
-	base?: string;
+	base: string;
 	compressHTML?: boolean;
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/app/types.ts
```diff
@@ -37,7 +37,7 @@ export type SSRManifest = {
 	adapterName: string;
 	routes: RouteInfo[];
 	site?: string;
-	base?: string;
+	base: string;
```
ematipico, line 40: I traced this type, and confirmed that `base` is always created by the function `buildManifest`

bluwy (approved):
Looks great!

PR #7575, opened 2023-07-05
Refactor `SSRResult` and `RenderContext`

bluwy:
## Changes

- Mainly removing props that are unused
- Also make `SSRResult._metadata` only contain ephemeral and mutable state

I made isolated commits that can be reviewed individually. But the whole changes aren't too big either. 

## Testing

Existing tests should pass

## Docs

n/a. internal refactor.

### DIFF
--- a/.changeset/many-nails-arrive.md
+++ b/.changeset/many-nails-arrive.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdoc': patch
+---
+
+Handle internal access change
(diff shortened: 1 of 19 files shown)


ematipico (approved):
Love it!

bluwy:
Pushed a commit to fix a markdoc fail. It's sort of a breaking change, but I feel like it's an internal API which would've been hard to refactor if we don't change this.

matthewp (approved):

PR #7582, opened 2023-07-05
Draft: Add an "allowHTML" feature for Markdoc

alex-sherwin:
## Changes

* Adds an *allowHTML* option for the @astrojs/markdoc integration which enables HTML tokenization/rendering in Markdoc markup

## Testing

With extensive unit tests on parsed HTML DOM elements


packages/integrations/markdoc/package.json
```diff
@@ -27,6 +27,10 @@
       "types": "./dist/extensions/shiki.d.ts",
       "default": "./dist/extensions/shiki.js"
     },
+    "./html": {
+      "types": "./dist/extensions/html/index.d.ts",
+      "default": "./dist/extensions/html/js"
+    },
```
bholmesdev, line 33: It looks like users won't import `@astrojs/markdoc/html` directly. I'd suggest removing this export.

packages/integrations/markdoc/src/extensions/html/htmlattributes/mutate-raw-html-attributes-to-react-compatible-variants.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Given a Record<string, string> of raw HTML attributes, fix various cases to be React
```
bholmesdev, line 2: Are React-specific transforms necessary for our Astro x Markdoc renderer? I'd be curious if tests continue to pass without this 🤷

bholmesdev (commented):
Overall, very impressed with the thorough documentation and unit testing. Thanks for this! I've left some comments below, and have some general comments to address:
- **Folder structure:** I'm not sure `html` belongs in `extensions/`, since it isn't applied using the `extends` configuration option. I think we can move `html/` to the base `src/` directory. I also think we can remove the `3rd-party/` directory; a code comment at the top denoting the source should be sufficient.
- **Attributing MIT licensed code:** Speaking with core, I don't think "taken from" is the right verbage for these comments. I'm discussing with core how we can best attribute external code. I'll follow-up with a better answer!

bholmesdev:
Update: Here's our recommendation for citing external code!

>  When your licensed code only lives in a single file, I'd do a comment in that one file instead. `/** @license MIT ${COPY-PASTE-THE-ENTIRE-LICENSE-HERE} */`
either at the top of the file, or as close to the one snippet of code as possible.

This may seem excessive, though it makes guidelines clear for code usage. I'd use this in the files under `3rd-party/` and the react transforms.

packages/integrations/markdoc/src/extensions/html/transform/html-token-transform.ts
```diff
@@ -0,0 +1,239 @@
+// third party
```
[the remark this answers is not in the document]
alex-sherwin: Oh, just to clarify, this is my personal convention annotating blocks of imports (I group imports like this):

```typescript
// third party
import * as React from "react";

// first party
import { x } from "@mycompany/pacakage";

// local
import { z } from "../../project/local";
```

It is not an indicator this file is from a third party.  But I will happily clean this up to remove confusion

packages/integrations/markdoc/src/extensions/html/htmlattributes/mutate-raw-html-attributes-to-react-compatible-variants.ts
alex-sherwin: Added test coverage of randomly cased HTML attributes in Markdoc markup to ensure this was not problematic

packages/integrations/markdoc/src/extensions/html/transform/html-token-transform.ts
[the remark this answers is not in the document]
alex-sherwin: See other similar comment, this is *not* an indicator of third party code, just a marker that this was a block of third party import statements (personal coding convention)

alex-sherwin:
I really screwed up this branch with a rebase to fix my commit author info :)

I'll leave this repo up for short-term posterity but will open up a PR from a clean branch

bholmesdev:
@alex-sherwin agh, just happened to me on a Markdoc release too. No problem, link when it's available!

alex-sherwin:
New PR is ready: https://github.com/withastro/astro/pull/7597

PR #7584, opened 2023-07-06
Simpify passing endpoint logging option

bluwy:
## Changes

Small refactor to avoid an extra `logging` parameter, because we can use `env.logging` instead.

## Testing

Existing test should pass.

## Docs

n/a


bluwy:
The smoke test seems to be unrelated and seeing it more often recently. It's coming from a failed request from npmjs while building the docs site.

ematipico (approved):

bluwy:
Wrong button 😄

PR #7585, opened 2023-07-06
Refactor passing compressHTML

bluwy:
## Changes

The `compressHTML` flow is a bit out of place. It's being passed manually to the rendering flow instead of throught `SSRResult` like the other configs. This PR:

- Moves `compressHTML` to `SSRResult`
- Adds `compressHTML` to `Environment` (to be passed to `SSRResult`)
- Make `compressHTML` type in `SSRManifest` non-optional. The value should always exist.

## Testing

Existing tests should pass.

## Docs

n/a


ematipico (approved):
I am loving these refactors, things are slowly starting to make sense

PR #7587, opened 2023-07-06
refactor: use SSR manifest in dev

ematipico:
## Changes

There isn't a great change in logic. In this PR we create an `SSRManifest` in dev mode, which we pass down until the `handleRoute`.

This should "normalize" the code across dev, SSG and SSR. I also changed a bit the `App` class to prepare some methods that will be needed in the near future.

## Testing

Current CI should pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -61,26 +61,41 @@ export class App {
 			routes: manifest.routes.map((route) => route.routeData),
 		};
 		this.#routeDataToRouteInfo = new Map(manifest.routes.map((route) => [route.routeData, route]));
-		this.env = createEnvironment({
-			adapterName: manifest.adapterName,
+		this.baseWithoutTrailingSlash = removeTrailingForwardSlash(this.#manifest.base);
+		this.env = this.#createEnvironment(streaming);
+	}
+
+	set setManifest(newManifest: SSRManifest) {
+		this.#manifest = newManifest;
+	}
+
+	/**
+	 * Creates an environment by reading the stored manifest
+	 *
+	 * @param streaming
+	 * @private
+	 */
+	#createEnvironment(streaming = false) {
+		return createEnvironment({
+			adapterName: this.#manifest.adapterName,
 			logging: this.logging,
-			markdown: manifest.markdown,
+			markdown: this.#manifest.markdown,
 			mode: 'production',
-			compressHTML: manifest.compressHTML,
-			renderers: manifest.renderers,
-			clientDirectives: manifest.clientDirectives,
-			async resolve(specifier: string) {
-				if (!(specifier in manifest.entryModules)) {
+			compressHTML: this.#manifest.compressHTML,
+			renderers: this.#manifest.renderers,
+			clientDirectives: this.#manifest.clientDirectives,
+			resolve: async (specifier: string) => {
+				if (!(specifier in this.#manifest.entryModules)) {
 					throw new Error(`Unable to resolve [${specifier}]`);
 				}
-				const bundlePath = manifest.entryModules[specifier];
+				const bundlePath = this.#manifest.entryModules[specifier];
 				switch (true) {
 					case bundlePath.startsWith('data:'):
 					case bundlePath.length === 0: {
 						return bundlePath;
 					}
 					default: {
-						return createAssetLink(bundlePath, manifest.base, manifest.assetsPrefix);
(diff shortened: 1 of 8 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -40,59 +40,78 @@ export interface MatchOptions {
...
+	 */
+	protected env: Environment;
 	#manifest: SSRManifest;
-	#manifestData: ManifestData;
+	protected manifestData: ManifestData;
 	#routeDataToRouteInfo: Map<RouteData, RouteInfo>;
-	#encoder = new TextEncoder();
-	#logging: LogOptions = {
+	protected encoder = new TextEncoder();
+	protected logging: LogOptions = {
```
ematipico, line 52: New classes that will extend this class will have access to this data; that's why I marked them `protected`.

packages/astro/src/vite-plugin-astro-server/plugin.ts
```diff
@@ -70,3 +78,29 @@ export default function createVitePluginAstroServer({
...
 }
+
+/**
+ * It creates a `SSRManifest` from the `AstroSettings`.
+ *
+ * Renderers needs to be pulled out from the page module emitted during the build.
+ * @param settings
+ * @param renderers
+ */
+export function generateDevelopmentManifest(settings: AstroSettings): SSRManifest {
```
bluwy, line 89: I'm curious to see how this would be used later. IIUC the `SSRManifest` is only used to pass info from build to the SSR runtime, where it has the special restriction of implementing [serialization](https://github.com/withastro/astro/blob/ab922c7f8e3fc367a3ff0f7921dbe01c54c58985/packages/astro/src/core/build/plugins/plugin-ssr.ts#L475) and [deserialization](https://github.com/withastro/astro/blob/ab922c7f8e3fc367a3ff0f7921dbe01c54c58985/packages/astro/src/core/app/common.ts#L4-L27) 🤔

packages/astro/src/core/app/index.ts
bluwy: Are there existing super-classes that uses these fields? Or will you be creating those as a followup?

We might want to be cautious a bit as JS doesn't have a `protected` concept, and they will now be public with this change ([TypeScript example](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQMQAsCmIISAvNAORnqZYDu8ATiACbSkAMVGADvfAC55gAlgDNErcpUzQe-QcOgAjMPQkVUAXyA)). Not blocking this change though if you do plan to use this later.

packages/astro/src/core/build/generate.ts
```diff
@@ -647,14 +646,14 @@ async function generatePath(
...
-export function generateRuntimeManifest(
+export function createBuildManifest(
 	settings: AstroSettings,
 	internals: BuildInternals,
 	renderers: SSRLoadedRenderer[]
 ): SSRManifest {
 	return {
 		assets: new Set(),
-		entryModules: {},
+		entryModules: Object.fromEntries(internals.entrySpecifierToBundleMap.entries()),
```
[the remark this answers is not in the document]
ematipico: I believe this is the code that is breaking the tests :(

packages/astro/src/core/app/index.ts
ematipico: Yeah, maybe it's a bit premature this change. I am going to revert it.

packages/astro/src/vite-plugin-astro-server/plugin.ts
bluwy: I also like that idea. On my initial attempts, I tried to create a `renderRoute` too, but it turns out we need to pass in a lot of options. Right now there seem to be 3 key things needed to render a page: `mod` (the runtime module instance), `RenderContext` (created per request), and `Environment` (created once).

IIUC it would be nice to lessen these 3 things. Maybe we can discuss more about your plans with involving `SSRManifest` into the mix. Maybe it'll merge into one of the 3 things?

packages/astro/src/core/build/generate.ts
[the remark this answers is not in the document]
bluwy: Looks like it's passing now 🚀

bluwy (approved):
LGTM. Open to see progress around `SSRManifest` for now!

matthewp (approved):

PR #7588, opened 2023-07-06
Refactor endpoint response object handling

bluwy:
## Changes

- Make `callEndpoint` return a response directly, even converting simple object form to a response so we don't duplicate work.
- For SSG build, we still need to return the object form. `callEndpoint` now accepts an extra `returnObjectFormIfAvailable` param to return that. Mostly for compat as we _might_ remove object form support in Astro 3.0

## Testing

Existing tests should pass.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -9,6 +9,7 @@ import type {
 } from '../../@types/astro';
 import type { Environment, RenderContext } from '../render/index';
 
+import mime from 'mime';
 import { isServerLikeOutput } from '../../prerender/utils.js';
 import { renderEndpoint } from '../../runtime/server/index.js';
 import { ASTRO_VERSION } from '../constants.js';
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/endpoint/index.ts
```diff
@@ -100,12 +97,29 @@ export function createAPIContext({
...
+	returnObjectFormIfAvailable?: boolean
+): Promise<Response | SimpleEndpointObject>;
+// Base implementation
+export async function callEndpoint<MiddlewareResult = Response | EndpointOutput>(
+	mod: EndpointHandler,
+	env: Environment,
+	ctx: RenderContext,
+	onRequest?: MiddlewareHandler<MiddlewareResult> | undefined,
+	returnObjectFormIfAvailable?: boolean
+): Promise<Response | SimpleEndpointObject> {
```
bluwy, line 122: Had to create this type abomination so we get the correct types depending if `returnObjectFormIfAvailable` is passed.

packages/astro/src/core/endpoint/index.ts
```diff
@@ -128,38 +142,54 @@ export async function callEndpoint<MiddlewareResult = Response | EndpointOutput>
...
+
+		const bytes = encoder.encode(response.body);
+		headers.set('Content-Length', bytes.byteLength.toString());
+
+		response = new Response(bytes, {
+			status: 200,
+			headers,
+		});
+
+		attachToResponse(response, context.cookies);
```
bluwy, line 189: The two places in `vite-plugin-astro-server/route.ts` and `core/app/index.ts` that has this same code is now consolidated here.

bluwy (commented):
Note: it's easier to review with whitespaces off.

bluwy:
Well this didn't really go well. Looks like in dev and prod, we handle simple objects with `encoding` differently.

Dev: Use Node's `Buffer.from`
Prod: Use `TextEncoder` (only works with `utf8` strings)

Dev behaviour encompasses prod behaviour, which is fine because it covers static mode that writes the content to a file with different encodings. I don't think this is a nice refactor until if we decide to remove the simple object syntax. Will comment on that linear issue instead.

PR #7589, opened 2023-07-06
feat: support features for adapters

ematipico:
## Changes

Recently we added new options to Astro, but these options are only valid if the adapter can support them. From a [discord discussion](https://discord.com/channels/830184174198718474/1120744365350993930/1126129156455927838), we thought that there should be a system where the adapter tells Astro if it's able to support certain features.

Since I am about to add one of these features to the Netlify adapter, I thought adding this system was an excellent chance.

## Testing

Added some test case

## Docs

Not now, I will prepare some something in the next PRs.

### DIFF
--- a/.changeset/sharp-elephants-vanish.md
+++ b/.changeset/sharp-elephants-vanish.md
@@ -1,0 +1,17 @@
+---
+'astro': minor
+---
+
+Now you can tell Astro if an adapter supports certain features.
+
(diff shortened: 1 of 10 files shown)


ematipico:
Should we also add other features that are considered stable for adapters?

matthewp (approved):

ematipico:
Closing, I am going to reopen it with a different approach for the Astro 3.0 release

PR #7593, opened 2023-07-07
Fix: Markdoc v0.4.0 docs

bholmesdev:
## Changes

- Add note on named exports to README
- Add link to docs from error hint

### DIFF
--- a/packages/integrations/markdoc/README.md
+++ b/packages/integrations/markdoc/README.md
@@ -136,6 +136,31 @@ Use tags like this fancy "aside" to add some _flair_ to your docs.
 {% /aside %}
 ```
 
+### Use Astro components from npm packages and TypeScript files
+
+You may need to use Astro components from TypeScript or JavaScript files with named imports. This is common when using npm packages and design systems.
+
+You can pass the import name as the second argument to the `component()` function like so:
+
+```js
+// markdoc.config.mjs
+import { defineMarkdocConfig, component } from '@astrojs/markdoc/config';
+
+export default defineMarkdocConfig({
+  tags: {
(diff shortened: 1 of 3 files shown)


packages/integrations/markdoc/README.md
```diff
@@ -136,6 +136,31 @@ Use tags like this fancy "aside" to add some _flair_ to your docs.
 {% /aside %}
 ```
 
+### Use Astro components from npm packages and TypeScript files
+
+You may need to use Astro components from TypeScript or JavaScript files with named imports. This is common when using npm packages and design systems.
+
+You can pass the import name as the second argument to the `component()` function like so:
```
delucis, line 143: No need for the “like so” here:
```suggestion
You can pass the import name as the second argument to the `component()` function:
```

delucis (approved):
LGTM, @bholmesdev! Left a couple of small comments but feel free to merge once you’ve taken a look.

packages/integrations/markdoc/README.md
```diff
@@ -136,6 +136,31 @@ Use tags like this fancy "aside" to add some _flair_ to your docs.
 {% /aside %}
 ```
 
+### Use Astro components from npm packages and TypeScript files
+
+You may need to use Astro components from TypeScript or JavaScript files with named imports. This is common when using npm packages and design systems.
```
delucis, line 141: Nit: I’ll let you decide if this is better or if you prefer your version.
```suggestion
You may need to use Astro components exposed as named exports from TypeScript or JavaScript files. This is common when using npm packages and design systems.
```

PR #7599, opened 2023-07-07
Fix: Hyphens breaking Markdoc tags

### LINKED ISSUE #7595: Tag name can't contain hyphens in @astrojs/markdoc v0.4.1

2.8.0

None

pnpm

Linux

Firefox

### Describe the Bug

Using tag names with hyphens is not possible in `@astrojs/markdoc` `v0.4.1`.

I get the following error (when running with `pnpm dev`):
```
Parse failure: Unexpected token (7:15)
At file: /path/to/index.en.mdoc
Contents of line 7: import Tagfloat-img from "/src/components/FloatImg.astro";
```

`markdoc.config.ts`:
```ts
import { component, defineMarkdocConfig, nodes } from "@astrojs/markdoc/config";

  nodes: {
    document: {
      ...nodes.document,
      render: undefined,
    },
  },
  tags: {
    "float-img": { render: component("./src/components/FloatImg.astro") },
  },
});
```

Seems like it's about the generated default import name as it can't contain hyphens. It worked in `v0.3.3`.

Building (with `pnpm build`) gives just a `Unexpected token (Note that you need plugins to import files that are not JavaScript)` error in the frontmatter for a markdoc containing the tag.

It should work with tags that contain hyphens (like in `v0.3.3`).

According to the [docs](https://markdoc.dev/docs/attributes#create-a-custom-attribute)  hyphens are allowed.

https://stackblitz.com/edit/github-7zrydw-n8uwg1?file=markdoc.config.mjs

### Participation

bholmesdev:
## Changes

- Resolves #7595

## Testing

- Add tagname with hyphen to test suite

### DIFF
--- a/.changeset/witty-pandas-behave.md
+++ b/.changeset/witty-pandas-behave.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdoc': patch
+---
+
+Fix hyphens in Markdoc tag names causing build failures
--- a/packages/integrations/markdoc/src/content-entry-type.ts
+++ b/packages/integrations/markdoc/src/content-entry-type.ts
@@ -225,14 +225,19 @@ function getStringifiedImports(
 	let stringifiedComponentImports = '';
 	for (const [key, config] of Object.entries(componentConfigMap)) {
 		const importName = config.namedExport
-			? `{ ${config.namedExport} as ${componentNamePrefix + key} }`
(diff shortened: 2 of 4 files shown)


natemoo-re (approved):
NWTWWHB! Thanks for jumping on this Ben!

PR #7607, opened 2023-07-09
feat: added types to allow the creation of user space content services

FineWolf:
## Changes

- Add new exported types in `astro:content` in order to allow user spaces services to effectively type content services.
- `CollectionKey` allows to target the name of all collections
- `ContentCollectionKey` allows to target the name of content collections
- `DataCollectionKey` allows to target the name of data collections

There is currently no way to cleanly `infer` a type that is equivalent to `keyof AnyEntryMap` in user space. This PR aims to rectify that.

## Testing

This is a typing change, no test is currently present to check the generated typings.

## Docs

The documentation doesn't currently cover the typings of `astro:content`.


ematipico:
Since we don't have a documentation page of this kind of stuff, we could use a changelog to show the usage of these new types.

Princesseuh (approved):

Issue #7154, opened 2023-05-21
.json files prefixed with underscores don't get ignored inside content collections on build


Princesseuh:
### What version of `astro` are you using?

2.5.1

### Are you using an SSR adapter? If so, which one?

N/A

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### What browser are you using?

Arc

### Describe the Bug

I have the following folder structure:

```tree
src/
└─ content/
   ├─ blog/
   │   └─ my-article/
   │     ├─ my-article.md
   │     └─ _some-data.json
   └─ config.ts
```
(I refer to `_some-data.json` from my layout through some arcane way that does not matter in this context.)

In dev, this works, `_some-data.json` is ignored, as files prefixed with underscores should be. In build however, `_some-data.json` is being passed through the Zod schema as if it was a file that needs to respect the `blog` schema. I would like for this file to be completely ignored, like it is in build.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-nxjeur?file=src/content/blog/my-article/my-article.md,src/pages/index.astro,src/content/blog/my-article/_my-data.json&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

birtles:
I believe this is a regression from 2.5. This used to work with 2.4.

Princesseuh:
> I believe this is a regression from 2.5. This used to work with 2.4.

Yes, 2.5 added data collections. Before that, JSON files were never processed, underscore or not

tonydangblog:
Hi @bholmesdev, I just updated to Astro 2.5.6 and seem to be having a similar, but slightly different issue.

I have the following folder structure:

```
src/
└─ content/
   ├─ blog/
   │   └─ my-article/
   │     ├─ my-article.md
   |     ├─ _my-data-directory/
   │       └─ some-data.json
   └─ config.ts
```

My json data (`some-data.json`) is NOT prefixed with an underscore, however it is in a directory (`_my-data-directory/`) that is prefixed with an underscore. 

I am getting the same issue where the build fails because it does not match the collection schema. May I ask if this is the intended behavior and it is required for json files to be always prefixed with an underscore even if it's already in a directory prefixed with an underscore?

Princesseuh:
This issue is back since `2.7.1`. Not sure what commit broke it exactly

PR #7611, opened 2023-07-10
Ignore content .json files prefixed with underscores

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/7154

Fix the glob to include nested folders/files with `_`

cc @Princesseuh my bad 😅 

## Testing

Added an integration test for this.

## Docs

n/a

### DIFF
--- a/.changeset/funny-tomatoes-guess.md
+++ b/.changeset/funny-tomatoes-guess.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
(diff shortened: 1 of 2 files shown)


Princesseuh:
Code looks great! Only comment I have would be to maybe add a test for ignoring a folder too, that way the entire feature is tested

ematipico (approved):

PR #7617, opened 2023-07-10
Add CLI startup speed benchmark

bluwy:
## Changes

- Support `pnpm astro-benchmark cli-startup` in root
- Support `!bench cli-startup` in PR comments (once merged)

Example output:

```
Making project: render-default
Finished making project: render-default
Benchmarking `astro --help`...
Done
Benchmarking `astro info`...
Done
Result preview:
==========
#### CLI Startup


| Command      | Avg (ms) | Stdev (ms) | Max (ms) |
| :----------- | -------: | ---------: | -------: |
| astro --help |   434.98 |      21.17 |   492.63 |
| astro info   |   468.74 |      28.54 |   527.03 |
==========
```

## Testing

Ran `pnpm astro-benchmark cli-startup` locally. I did not test if the `!bench` PR comment would work, but it _should_ following the other patterns.

## Docs

n/a. The `!bench` command should be documented in `CONTRIBUTING` in general.

### DIFF
--- a/packages/astro/src/cli/index.ts
+++ b/packages/astro/src/cli/index.ts
@@ -113,7 +113,7 @@ async function printInfo({
 	} catch (_e) {}
 	console.log();
 	printRow('Astro version', `v${ASTRO_VERSION}`);
-	printRow('Package manager', packageManager.name);
+	printRow('Package manager', packageManager?.name);
 	printRow('Platform', platform());
 	printRow('Architecture', arch());
 	printRow('Adapter', adapter);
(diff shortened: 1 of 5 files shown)


ematipico (approved):

packages/astro/src/cli/index.ts
```diff
@@ -113,7 +113,7 @@ async function printInfo({
 	} catch (_e) {}
 	console.log();
 	printRow('Astro version', `v${ASTRO_VERSION}`);
-	printRow('Package manager', packageManager.name);
+	printRow('Package manager', packageManager?.name);
```
bluwy, line 116: Forgot to put a note. This was needed because running `node /path/to/astro.js` directly yields no package manager, so adding the `?.` to prevent an error while running the benchmark.

ematipico: We could fix it by adding a "Could not determine." label, I'll send a PR

PR #7619, opened 2023-07-11
Refactor simple CLI commands

bluwy:
## Changes

Each commit should show an isolated change, but the overview is:

- The `docs`, `info`, `telemetry`, `add` commands are now located under `src/cli/{command}/index.ts`
- Remove `logging` option from `openConfig` as it's unused, this simplifies the `info` command.

In the following PR, I'll handle the more complex commands like `dev`, `build`, `preview`, `sync`, and `check`

## Testing

Ran the CLI manually to test

## Docs

n/a. internal refactor.


packages/astro/src/cli/index.ts
```diff
@@ -69,62 +67,11 @@ function printAstroHelp() {
 }
 
 /** Display --version flag */
-async function printVersion() {
+function printVersion() {
 	console.log();
 	console.log(`  ${colors.bgGreen(colors.black(` astro `))} ${colors.green(`v${ASTRO_VERSION}`)}`);
 }
 
-async function printInfo({
```
bluwy, line 77: This chunk of code is moved to `src/info/index.ts`, replacing dynamic imports as normal imports, as the entire module would be dynamically imported instead.

packages/astro/src/cli/index.ts
```diff
@@ -171,64 +118,56 @@ async function handleConfigError(
...
-		dest: nodeLogDestination,
-		level: 'info',
-	};
+
+	// These commands can run directly without parsing the user config.
 	switch (cmd) {
 		case 'help':
 			printAstroHelp();
-			return process.exit(0);
+			return;
```
bluwy, line 126: I don't think we need the `process.exit` here to short-circuit the CLI. The continuing paths after a `return` should immediately exit node too. Exit code should still be 0 because of https://github.com/withastro/astro/blob/34918c9e043ac434d219910ea5599491f2504f22/packages/astro/astro.js#L74-L76

ematipico (approved):

PR #7621, opened 2023-07-11
fix(@astrojs/vercel): improve file detection

ematipico:
## Changes

The file detection of the handler file needed to be corrected.

## Testing

I improved the testing cases by adding an assertion for the contents inside the middleware file. I noticed that the snapshots are not always reliable. 

## Docs

N/A


packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -50,7 +50,7 @@ function edgeMiddlewareTemplate(middlewarePath: string, vercelEdgeMiddlewareHand
 	const filePathEdgeMiddleware = fileURLToPath(vercelEdgeMiddlewareHandlerPath);
 	let handlerTemplateImport = '';
 	let handlerTemplateCall = '{}';
-	if (existsSync(filePathEdgeMiddleware) + '.js' || existsSync(filePathEdgeMiddleware) + '.ts') {
+	if (existsSync(filePathEdgeMiddleware + '.js') || existsSync(filePathEdgeMiddleware + '.ts')) {
```
bluwy, line 53: Wow I missed this too 🙈

packages/integrations/vercel/test/edge-middleware.test.js
```diff
@@ -4,27 +4,38 @@ import chaiJestSnapshot from 'chai-jest-snapshot';
...
+before(function () {
+	chaiJestSnapshot.resetSnapshotRegistry();
+});
 
-	beforeEach(function () {
-		chaiJestSnapshot.configureUsingMochaContext(this);
-	});
+beforeEach(function () {
+	chaiJestSnapshot.configureUsingMochaContext(this);
+});
```
bluwy, line 13: Do we still need this code since there's the `setup.js` file now?

bluwy (approved):

bluwy:
> One more thing I noticed is, snapshot tests are very slow, even on my machine (MacBook M2). I wonder if there's a better testing library out there

`chai-jest-snapshot` is a little old and using an older version of `jest-snapshot`, which maybe attributes to the slowness. Maybe we could fork a more up-to-date version?

packages/integrations/vercel/test/edge-middleware.test.js
ematipico: Oh yeah, good catch. We can remove it

PR #7622, opened 2023-07-11
Refactor config loading

bluwy:
## Changes

- Simplify config loading via nodejs and vite
- Move config merging logic to separate file
- Bug fix with telemetry report on config error

## Testing

Tested manually for config loading and restarting logic. The smoke tests should also cover this.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/cli/index.ts
+++ b/packages/astro/src/cli/index.ts
@@ -102,13 +102,10 @@ async function handleConfigError(
 	error(logging, 'astro', `Unable to load ${path ? colors.bold(path) : 'your Astro config'}\n`);
 	if (e instanceof ZodError) {
 		console.error(formatConfigErrorMessage(e) + '\n');
+		telemetry.record(eventConfigError({ cmd, err: e, isFatal: true }));
 	} else if (e instanceof Error) {
 		console.error(formatErrorMessage(collectErrorMetadata(e)) + '\n');
 	}
-	const telemetryPromise = telemetry.record(eventConfigError({ cmd, err: e, isFatal: true }));
-	await telemetryPromise.catch((err2: Error) =>
-		debug('telemetry', `record() error: ${err2.message}`)
-	);
 }
 
 /**
--- a/packages/astro/src/core/config/config.ts
+++ b/packages/astro/src/core/config/config.ts
@@ -5,9 +5,7 @@ import fs from 'fs';
 import * as colors from 'kleur/colors';
 import path from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
-import { mergeConfig as mergeViteConfig } from 'vite';
 import { AstroError, AstroErrorData } from '../errors/index.js';
-import { arraify, isObject, isURL } from '../util.js';
 import { createRelativeSchema } from './schema.js';
 import { loadConfigWithVite } from './vite-load.js';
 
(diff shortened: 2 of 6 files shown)


packages/astro/src/cli/index.ts
```diff
@@ -102,13 +102,10 @@ async function handleConfigError(
 	error(logging, 'astro', `Unable to load ${path ? colors.bold(path) : 'your Astro config'}\n`);
 	if (e instanceof ZodError) {
 		console.error(formatConfigErrorMessage(e) + '\n');
+		telemetry.record(eventConfigError({ cmd, err: e, isFatal: true }));
```
bluwy, line 105: `eventConfigError` only accepts `ZodError`

packages/astro/src/core/config/config.ts
```diff
@@ -226,54 +220,24 @@ export async function openConfig(configOptions: LoadConfigOptions): Promise<Open
...
-		if (configOptions.isRestart) {
-			// Hack: Write config to temporary file at project root
-			// This invalidates and reloads file contents when using ESM imports or "resolve"
-			const tempConfigPath = path.join(
-				root,
-				`.temp.${Date.now()}.config${path.extname(configPath)}`
-			);
-
-			const currentConfigContent = await fsMod.promises.readFile(configPath, 'utf-8');
-			await fs.promises.writeFile(tempConfigPath, currentConfigContent);
```
bluwy, line 256: This custom write flow is converted to 
```ts
await import(pathToFileURL(configPath).toString() + '?t=' + Date.now());
```
in `vite-load.ts`

packages/astro/src/core/config/config.ts
```diff
@@ -226,54 +220,24 @@ export async function openConfig(configOptions: LoadConfigOptions): Promise<Open
...
-	}
+	const configPath = await resolveConfigPath({
+		cwd: configOptions.cwd,
+		flags: configOptions.flags,
+		fs: fsMod,
+	});
+	if (!configPath) return undefined;
+
+	// Create a vite server to load the config
+	return await loadConfigWithVite({
```
bluwy, line 236: Since we have the `configPath` guard here, we can refactor `loadConfigWithVite` to return the config directly, instead of `{ value, configPath }`, because we already know the `configPath`.

packages/astro/src/core/config/config.ts
```diff
@@ -226,54 +220,24 @@ export async function openConfig(configOptions: LoadConfigOptions): Promise<Open
...
-interface TryLoadConfigResult {
-	value: Record<string, any>;
-	filePath?: string;
-}
-
 async function tryLoadConfig(
 	configOptions: LoadConfigOptions,
 	root: string
-): Promise<TryLoadConfigResult | undefined> {
+): Promise<Record<string, any> | undefined> {
```
ematipico, line 226: While we're refactoring, I want to propose the following change, with two options:
1. rename `tryLoadConfig` to `loadConfig` and always return a `Record<string, any>`. The function will return `{}` if no `configPath` is returned;
2. keep the name `tryLoadConfig` but throw a runtime error, and let the consumer handle the error;

What do you think?

It's a pity that there's no way to signal that a function can throw an error to a consumer :(

ematipico (approved):
Let's 🚢  it!

packages/astro/src/core/config/config.ts
bluwy: Good call! I didn't spot the `loadConfig` opportunity (option 1) which we can definitely do. Went with that instead since throwing errors are a bit expensive.

Issue #7602, opened 2023-07-08
Astro's Response objects can't be cloned on Node (dev server, adapter)


lilnasy:
### What version of `astro` are you using?

2.8

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

Calling `.clone()` on the response returned from middleware's `next()` returns a new `Response` that is missing its body.
- The issue exists on the dev server.
- The issue exists on the node adapter.
- The issue does not exist on Deno or Cloudflare adapters.
- The issue does not exist on non-Astro-generated responses or those reconstructed from an Astro-generated response.
<img width="418" alt="image" src="https://github.com/withastro/astro/assets/69170106/e73a0ac7-8ae6-4a49-ba55-edace3cfab75">
<img width="588" alt="image" src="https://github.com/withastro/astro/assets/69170106/e3e3d12d-8ff9-4495-b073-89508c78c634">

# Workaround
A response reconstructed from Astro's response can be cloned.
<img width="392" alt="image" src="https://github.com/withastro/astro/assets/69170106/b089737a-a4ab-4c47-9afd-ab8fbf715c4f">
<img width="581" alt="image" src="https://github.com/withastro/astro/assets/69170106/6686ad4a-68f0-4038-93b2-eab355c036b3">

### What's the expected result?

See the screenshot showing terminal output in Workaround

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-eixftf?file=src%2Fmiddleware.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
What version of Node are you using?

lilnasy:
> What version of Node are you using?

v20.3.0, but I can reproduce on v18.16.1 and v16.20.1.

PR #7623, opened 2023-07-11
Fixes Response not being cloneable by middleware

matthewp:
## Changes

- Implements `clone()` in our response wrapper.
- We should probably investigate removing this in 3.0.
- Fixes https://github.com/withastro/astro/issues/7602

## Testing

- Updated middleware test with this test case.

## Docs

N/A


ematipico (approved):
Looks good to me!

PR #7626, opened 2023-07-11
feat: bumped typescript-eslint to v6 stable 🥳

JoshuaKGoldberg:
## Changes

Bumps to the stable config now that [typescript-eslint v6 is released as stable](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6).

I don't think this needs a changeset, since the only package code it touches is to remove a couple of comments?

## Testing

```shell
pnpm lint
```

Linting-only change, no tests needed I think?

## Docs

Internal-only change, no docs needed I think?


packages/astro/src/events/error.ts
```diff
@@ -72,8 +72,6 @@ export function eventError({
 		cliCommand: cmd,
 		isFatal: isFatal,
 		anonymousMessageHint:
-			// https://github.com/typescript-eslint/typescript-eslint/issues/4820
-			// eslint-disable-next-line @typescript-eslint/prefer-optional-chain -- errorData may be false
```
JoshuaKGoldberg, line 76: Explaining: I disable the rule in `.eslintrc.cjs`, so it doesn't need to be disabled here.

bluwy (approved):

ematipico (approved):
Thank you for your hard work @JoshuaKGoldberg !

PR #7630, opened 2023-07-12
Fix `build.excludeMiddleware` default type docs

Yan-Thomas:
## Changes

The default value should be enclosed by backticks, rather than curly braces to avoid a rendering issue in docs.

image

## Docs

It's all about the docs


sarah11918 (approved):
Thanks, @Yan-Thomas!

PR #7634, opened 2023-07-13
Move all CLI commands to their subfolders

bluwy:
## Changes

- Move the `dev`, `build`, `preview`, `sync` commands to their own folder
- Move loading `settings` to each command itself

---

This currently creates more code, but only a stop point for now to make further refactoring easier. Each command can now start the actual work with only the `flags` and `logging` params.

## Testing

Tested manually. This should also improve startup perf a little. Will run the benchmark

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/cli/build/index.ts
+++ b/packages/astro/src/cli/build/index.ts
@@ -1,0 +1,21 @@
+import type yargs from 'yargs-parser';
+import _build from '../../core/build/index.js';
+import type { LogOptions } from '../../core/logger/core.js';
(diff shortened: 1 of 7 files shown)


bluwy:
!bench cli-startup

ematipico (approved):
Great numbers!

bluwy:
Looks like it brought down `astro --help` from 650ms to 330ms! `astro info` stays the same at the moment, but I'll work on removing the base cost which should help both again later.

PR #7635, opened 2023-07-13
Convert CLI top-level imports as dynamic imports

bluwy:
## Changes

- Move top-level imports in `src/cli/index.ts` to separate files, and make them dynamically imported instead

NOTE to self: This is the last change for the CLI speedup refactor, there's still a little more to improve by splitting up the error messages from the `src/core/message.ts` file. But I don't think it's worth it now. The CLI for `astro --help` and `astro docs` starts around 150-200ms now.

## Testing

Tested manually. Will also run benchmark command for this.

## Docs

n/a. internal refactor.


bluwy:
!bench cli-startup

ematipico (approved):

Princesseuh (approved):
Awesome work!

Issue #7304, opened 2023-06-06
Images are in wrong orientation


Akxe:
### What version of `astro` are you using?

2.5.5

### Are you using an SSR adapter? If so, which one?

Github pages

### What package manager are you using?

npm

### What operating system are you using?

Windows & Ubuntu (for build)

### What browser are you using?

Chrome

### Describe the Bug

I was pulling images from `.md` as a list, so I had to process them in `/projects/[...slug].astro` file

I used the snippet code below:

```typescript
const photos = (await Promise.all(data.photos.map(photo => getImage({
  src: photo.img,
  alt: photo.alt || `Vizualizace projektu "${data.title}"`,
})))).map(photo => {
  if (typeof photo.options.src == 'string') {
    throw new Error('Image not loaded');
  }

  return {
    src: photo.src,
    alt: photo.attributes.alt as string,
    style: {
      aspectRatio: `${photo.options.src.width}/${photo.options.src.height}`, 
      objectFit: 'cover',
    },
  };
});
```

I was the optimizing the code to run more in parallel and fixed it. To my surprise, the fix is when you switch from `getImage` options object to a string.

```typescript
const photos = await Promise.all(data.photos.map(photo => getImage(photo.img).then(result => {
  return {
    src: result.src,
    alt: photo.alt || `Vizualizace projektu "${data.title}"`,
    style: {
      aspectRatio: `${result.options.width}/${result.options.height}`,
      objectFit: 'cover',
    },
  };
})));
```

The bug is present regardless if you use Squoosh or Sharp.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-jfzz89?file=src%2Fpages%2Findex.astro&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

millette:
For comparison, see:

<https://github.com/withastro/astro/blob/main/packages/integrations/image/src/loaders/sharp.ts#L18-L19>

PR #7637, opened 2023-07-13
fix(assets): Fix images not following EXIF rotation

Princesseuh:
## Changes

As the title says

Fix https://github.com/withastro/astro/issues/7304

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/packages/astro/src/assets/services/squoosh.ts
+++ b/packages/astro/src/assets/services/squoosh.ts
@@ -1,11 +1,13 @@
 // TODO: Investigate removing this service once sharp lands WASM support, as libsquoosh is deprecated
 
 import type { ImageOutputFormat, ImageQualityPreset } from '../types.js';
+import { imageMetadata } from '../utils/metadata.js';
 import {
 	baseService,
 	parseQuality,
 	type BaseServiceTransform,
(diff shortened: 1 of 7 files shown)


packages/astro/src/assets/services/squoosh.ts
```diff
@@ -28,6 +30,28 @@ const qualityTable: Record<
...
+		case 3:
+		case 4:
+			return { type: 'rotate', numRotations: 2 };
+		case 5:
+		case 6:
+			return { type: 'rotate', numRotations: 1 };
+		case 7:
+		case 8:
+			return { type: 'rotate', numRotations: 3 };
+	}
```
bluwy, line 52: Would be nice to link to https://exiftool.org/TagNames/EXIF.html for an explanation of these magic numbers. Or an inline comment of what each number represent.

bluwy (approved):

packages/astro/src/assets/services/squoosh.ts
Princesseuh: Pff, imagine not knowing the EXIF rotation numbers by heart 

(I agree, will update, thank you!)

Issue #7581, opened 2023-07-05
Experimental redirects: Dynamic routes interfere with redirects in SSR mode 


paulrudy:
### What version of `astro` are you using?

2.7.4

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome, Safari

### Describe the Bug

In SSR mode, I'm using a dynamic route `[...slug]` to serve pages from a content collection. I'm also trying to use experimental redirects to redirect specific routes. But the dynamic route seems to supersede the redirects, making them useless.

Here's an experimental redirect. `foo.md` doesn't exist, and `bar.md` does exist: 
```
redirects: {
    '/foo': '/bar',
  },
```

The redirect is ignored, and navigating to `/foo` results in an error in rendering the dynamic route: `Cannot read properties of undefined (reading 'render')`.

I'm not sure whether or not this issue is based on a misreading of [this line in the docs](https://github.com/withastro/docs/blob/7997df0959c3f806bc9372515235ef1d54f1f1a1/src/content/docs/en/core-concepts/routing.mdx#L324):

> Redirects also follow the same rules, but are prioritized *last*; if there is a file-based route and a redirect with the same route rule, the file-based route is chosen.

[Edit: [now updated](https://github.com/withastro/docs/pull/3659) to say "same route _priority level_"]

The way I interpret this line is that, since `Static routes without path parameters will take precedence over all other routes`, if the redirect is a static route, it should be a higher priority than a file-based route with a path or rest parameter. I can't see it making sense any other way.

### What's the expected result?

The redirect's destination should be the updated parameter for the dynamic route.

From the example, navigating to `/foo` should redirect to `/bar`, thus `[...slug]` should receive `bar` as its parameter.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-z4igp1-ulwjc3?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
@paulrudy Here's how the proposal tackles the dynamic routes: https://github.com/withastro/roadmap/blob/redirects/proposals/redirects.md#dynamic-routes

I believe your use case is not handled per design.

paulrudy:
> @paulrudy Here's how the proposal tackles the dynamic routes: https://github.com/withastro/roadmap/blob/redirects/proposals/redirects.md#dynamic-routes
> 
> 
> 
[...]

Issue #7478, opened 2023-06-25
Experimental redirect: "Failed to resolve entry for package"


sanderbongers:
### What version of `astro` are you using?

2.7.0

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

pnpm (but happens with all)

### What operating system are you using?

Mac

### What browser are you using?

Firefox

### Describe the Bug

My aim is to use /blog as the homepage instead of /, which I tried using aredirect, like this:

astro.config.mjs
```
export default defineConfig({
	experimental: {
		redirects: true,
	},
	redirects: {
		"/": "/blog",
	},
});
```

But whenever the redirect key is `"/"` or `""`, the build fails (node v20.3.0):

```
09:28:38 PM [build] Building static entrypoints...
[@astro/plugin-build-internals] Failed to resolve entry for package "/usr/local/var/www/blog". The package may have incorrect main/module/exports specified in its package.json.
 error   Failed to resolve entry for package "/usr/local/var/www/blog". The package may have incorrect main/module/exports specified in its package.json.
  File:
    /usr/local/var/www/blog/node_modules/.pnpm/vite@4.3.9/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:23382:11
  Code:
    23381 | function packageEntryFailure(id, details) {
    > 23382 |     throw new Error(`Failed to resolve entry for package "${id}". ` +
            |           ^
      23383 |         `The package may have incorrect main/module/exports specified in its package.json` +
      23384 |         (details ? ': ' + details : '.'));
      23385 | }
  Stacktrace:
```

`pnpm run dev` works though, it's only `pnpm run build` that fails.

I've implemented a workaround by using a double slash, like so: `"//": "/blog"`, then it behaves as expected.

### What's the expected result?

Being able to build the site with the intended redirect.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/withastro/astro/tree/latest/examples/blog?file=astro.config.mjs&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

JerryWu1234:
let me take a look

JerryWu1234:
because '/' was occupied by the vite.
I want to add the '/' after the '/' if the user has used '/' as  a redirect path.but I feel that it's maybe not a good idea. I'm afraid of my work being useless.
could you share you idea ,if you are available

awcot:
Similar config & goals, same issue

sanderbongers:
[...]

PR #7644, opened 2023-07-13
Fixes for redirects config

matthewp:
## Changes

- Updates the prioritization logic so that redirects are inserted before dynamic routes of the same base.
- Fixes https://github.com/withastro/astro/issues/7581
- Makes it so that we don't try to resolve the `/` redirect.
- Fixes https://github.com/withastro/astro/issues/7478
- Implements recommendations from the RFC for static files built: https://github.com/withastro/roadmap/issues/466#issuecomment-1595940678
  - Adds `<meta name="robots" content="noindex">` so that search engines do not index the page.
  - Adds a delay for 302 redirects because Google interprets that as temporary.
  - Adds a `<a>` for bots to follow (and humans to see).
  - Adds `<link rel="canonical">`.

## Testing

- Everything is tested.

## Docs

N/A, bug fixes.


packages/astro/src/core/build/generate.ts
```diff
@@ -619,9 +619,18 @@ async function generatePath(
...
+			// https://developers.google.com/search/docs/crawling-indexing/301-redirects#metarefresh
+			const delay = response.status === 302 ? 2 : 0;
 			body = `<!doctype html>
 <title>Redirecting to: ${location}</title>
-<meta http-equiv="refresh" content="0;url=${location}" />`;
+<meta http-equiv="refresh" content="${delay};url=${location}">
+<meta name="robots" content="noindex">
+<link rel="canonical" href="${location}">
+<body>
+	<a href="${location}">Redirecting from <code>${fromPath}</code> to <code>${location}</code></a>
```
natemoo-re, line 632: This is a nice touch!

natemoo-re (approved):
Looks great!

PR #7646, opened 2023-07-13
nit: export props has been unnecessary.. forever

Princesseuh:
## Changes

People always get confused by `export interface Props` since the `export` has never, ever, been needed. This removes it from every public-facing files in the repo

## Testing

N/A

## Docs

N/A

### DIFF
--- a/examples/basics/src/components/Card.astro
+++ b/examples/basics/src/components/Card.astro
@@ -1,5 +1,5 @@
 ---
-export interface Props {
+interface Props {
 	title: string;
 	body: string;
 	href: string;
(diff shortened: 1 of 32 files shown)


bluwy (approved):
I suppose exporting it makes it easier to import the `Props` type from other components? But I also don't mind this change.

Princesseuh:
> I suppose exporting it makes it easier to import the `Props` type from other components? But I also don't mind this change.

The pattern is definitely still supported, this mostly avoids users thinking the export is necessary

ematipico:
Why do you think users do that?

ematipico (approved):

Princesseuh:
> Why do you think users do that?

We made a mistake once in the documentation, at the total beginning of Astro and users propagated it so much everyone got Mandela effect on it being required

PR #7647, opened 2023-07-14
Add new internal `resolveConfig` API

bluwy:
## Changes

**Note: This merges into the `js-api` branch while I work on these step-by-step**

- Add new `resolveConfig` API. Does the same thing as `openConfig`, but duplicating to easier migrate step-by-step.
- Add `AstroInlineConfig` type. Ideally this will be the config type the `dev` / `build` JS API will accept. ("inline" borrowed from Vite)
- Refactor `resolveConfigPath` to not accept the raw `flags`. Easier to divide between JS and CLI code.

## Testing

Existing tests should pass.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1309,6 +1309,9 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
 	// TypeScript still confirms zod validation matches this type.
 	integrations: AstroIntegration[];
 }
+export interface AstroInlineConfig extends AstroUserConfig {
+	configFile?: string | false;
+}
 
 export type ContentEntryModule = {
 	id: string;
--- a/packages/astro/src/core/config/config.ts
+++ b/packages/astro/src/core/config/config.ts
@@ -1,5 +1,5 @@
 import type { Arguments as Flags } from 'yargs-parser';
-import type { AstroConfig, AstroUserConfig, CLIFlags } from '../../@types/astro';
+import type { AstroConfig, AstroUserConfig, CLIFlags, AstroInlineConfig } from '../../@types/astro';
 
 import fs from 'fs';
 import * as colors from 'kleur/colors';
@@ -10,7 +10,7 @@ import { mergeConfig } from './merge.js';
 import { createRelativeSchema } from './schema.js';
 import { loadConfigWithVite } from './vite-load.js';
 
-export const LEGACY_ASTRO_CONFIG_KEYS = new Set([
+const LEGACY_ASTRO_CONFIG_KEYS = new Set([
 	'projectRoot',
 	'src',
 	'pages',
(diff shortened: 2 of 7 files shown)


packages/astro/src/core/config/config.ts
```diff
@@ -23,11 +23,7 @@ export const LEGACY_ASTRO_CONFIG_KEYS = new Set([
 ]);
 
 /** Turn raw config values into normalized values */
-export async function validateConfig(
-	userConfig: any,
-	root: string,
-	cmd: string
-): Promise<AstroConfig> {
+async function validateConfig(userConfig: any, root: string, cmd: string): Promise<AstroConfig> {
```
bluwy, line 26: Un-export as it's not used externally

packages/astro/src/core/config/config.ts
```diff
@@ -157,82 +155,74 @@ interface LoadConfigOptions {
...
 		fs: fsMod,
-		root,
 	});
 }
 
 /** Attempt to resolve an Astro configuration object. Normalize, validate, and return. */
-export async function resolveConfig(
+// NOTE: This function will be removed in a later PR. Renamed with "Legacy" suffix to make way
+// for the new `resolveConfig` API.
+async function resolveConfigLegacy(
```
bluwy, line 225: Un-export as it's not used internally. Also rename to make way for new API name.

packages/astro/src/core/config/config.ts
bluwy: Nevermind it's actually used in tests.

packages/astro/src/@types/astro.ts
```diff
@@ -1309,6 +1309,9 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
 	// TypeScript still confirms zod validation matches this type.
 	integrations: AstroIntegration[];
 }
+export interface AstroInlineConfig extends AstroUserConfig {
+	configFile?: string | false;
```
ematipico, line 1313: This is a weird type. What' the relation between a `string` and a `false`? I would have expected a `string` or `undefined`

packages/astro/src/core/config/config.ts
```diff
@@ -248,5 +238,25 @@ export function createDefaultDevConfig(
...
+	userConfig: AstroUserConfig;
+	astroConfig: AstroConfig;
+}
+
+// NOTE: This is the new API to load an Astro config. Use this instead of the other functions!
+export async function resolveConfig(
+	inlineConfig: AstroInlineConfig,
+	command: string,
+	fsMod = fs
+): Promise<ResolveConfigResult> {
```
ematipico, line 254: I think it's worth documenting :)

ematipico (approved):

packages/astro/src/@types/astro.ts
bluwy: `false` would mean not loading the `astro.config.js` at all. It would be most helpful when we have the JS API, where we want to use this inlined config only, and not load any external `astro.config.js` file. I copied this pattern from Vite.

We can't use `undefined` as it implicitly means "search for an astro config and load it"

packages/astro/src/core/config/config.ts
bluwy: Definitely after I get get all the refactors done 😄 I'm not completely sure this is the final shape yet so sort-of an interim code.

packages/astro/src/@types/astro.ts
ematipico: It's still a weird pattern that requires some time to understand for a person that sees it for the first time. There are three states that we need to document:
- the absence of `configFile`
- `configFile` passed as `false`
- `configFile` passed as `string`

bluwy: Thanks yeah I'll keep note of that to be documented later 👍

PR #7651, opened 2023-07-14
Handle flaky tests

bluwy:
## Changes

- Update cloudflare tests to prevent test timeout. Move CLI startup and teardown to before/after hook to spread out the timeout limit
  - Handles `"renders the solid component"` and `"Load cf and caches API"` fails
- Set integration test timeout to 30s. Some builds take longer than expected.
  - Handles `"Code component inside static build"` and `"Static build: dir takes the URL path to the output directory"` fails
- Retry getStaticPath test twice
  - Handles `"only calls getStaticPaths once"` fail

## Testing

Existing tests should still pass.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -107,8 +107,8 @@
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
     "test:unit": "mocha --exit --timeout 30000 ./test/units/**/*.test.js",
     "test:unit:match": "mocha --exit --timeout 30000 ./test/units/**/*.test.js -g",
-    "test": "pnpm run test:unit && mocha --exit --timeout 20000 --ignore **/lit-element.test.js && mocha --timeout 20000 **/lit-element.test.js",
-    "test:match": "mocha --timeout 20000 -g",
+    "test": "pnpm run test:unit && mocha --exit --timeout 30000 --ignore **/lit-element.test.js && mocha --timeout 30000 **/lit-element.test.js",
+    "test:match": "mocha --timeout 30000 -g",
     "test:e2e": "playwright test",
     "test:e2e:match": "playwright test -g"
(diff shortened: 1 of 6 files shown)


packages/integrations/cloudflare/test/basics.test.js
```diff
@@ -5,28 +5,29 @@ import * as cheerio from 'cheerio';
 describe.skip('Basic app', () => {
```
Princesseuh, line 5: Is the skip intended?

bluwy: Looks like it was skipped in https://github.com/withastro/astro/pull/4117. I found the bug though that had us skipped the test and will submit that separately.

Princesseuh (approved):

packages/astro/package.json
```diff
@@ -107,8 +107,8 @@
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
     "test:unit": "mocha --exit --timeout 30000 ./test/units/**/*.test.js",
     "test:unit:match": "mocha --exit --timeout 30000 ./test/units/**/*.test.js -g",
-    "test": "pnpm run test:unit && mocha --exit --timeout 20000 --ignore **/lit-element.test.js && mocha --timeout 20000 **/lit-element.test.js",
-    "test:match": "mocha --timeout 20000 -g",
+    "test": "pnpm run test:unit && mocha --exit --timeout 30000 --ignore **/lit-element.test.js && mocha --timeout 30000 **/lit-element.test.js",
+    "test:match": "mocha --timeout 30000 -g",
```
Princesseuh, line 111: Every time we raise this number, I cry a little

bluwy: If it helps, the unit test was the first to bump to 30000 first 😄

PR #7657, opened 2023-07-14
Added support for a `maxWidth` property in assets

alvinometric:
## Changes

This updates the base service in Assets to support a `maxWidth` parameter. With this, all images over a certain width would be resized. 

### Why? 

I got used to this with Gatsby, and if the user puts a giant 6000x6000 image it would never get sent to the browser at that size.

Usage:
```
// astro.config.mjs
image: {
    service: {
      entrypoint: 'astro/assets/services/sharp',
      config: {
        maxWidth: 590,
      },
    },
  },
```

## Testing

Locally using a project. I'm happy to add a test.

## Docs

It will need to be added to the docs, if accepted. I'm not sure if it should be in the API docs or just the Assets docs.

/cc @withastro/maintainers-docs for feedback

### DIFF
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -108,7 +108,7 @@ export type BaseServiceTransform = {
  *
  */
 export const baseService: Omit<LocalImageService, 'transform'> = {
-	validateOptions(options) {
+	validateOptions(options, serviceConfig) {
 		// `src` is missing or is `undefined`.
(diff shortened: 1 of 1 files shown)


Princesseuh:
Closing since we won't do this. Our current recommendation for anyone trying to do this is to either create a wrapper around Image / getImage if you're using those and if you need it everywhere, a wrapper around our base image services.

Thank you for contributing nonetheless!

PR #7669, opened 2023-07-16
Update README.md

Boston343:
## Changes

Add a short note under "Static sites" that you do not need the Netlify adapter to use Netlify functions.

## Testing

No tests. Small docs README text change only.

## Docs

This is a small documentation update. I was confused by this when I was learning how to setup Netlify functions with Astro, and discovered I did not need the adapter.

### DIFF
--- a/packages/integrations/netlify/README.md
+++ b/packages/integrations/netlify/README.md
@@ -92,7 +92,7 @@ export default defineConfig({
 
 ### Static sites
 
-For static sites you usually don't need an adapter. However, if you use `redirects` configuration (experimental) in your Astro config, the Netlify adapter can be used to translate this to the proper `_redirects` format.
+For static sites you usually don't need an adapter, and you are able to use Netlify functions without the adapter. However, if you use `redirects` configuration (experimental) in your Astro config, the Netlify adapter can be used to translate this to the proper `_redirects` format.
 
 ```js
 import { defineConfig } from 'astro/config';


packages/integrations/netlify/README.md
```diff
@@ -92,7 +92,7 @@ export default defineConfig({
 
 ### Static sites
 
-For static sites you usually don't need an adapter. However, if you use `redirects` configuration (experimental) in your Astro config, the Netlify adapter can be used to translate this to the proper `_redirects` format.
+For static sites you usually don't need an adapter, and you are able to use Netlify functions without the adapter. However, if you use `redirects` configuration (experimental) in your Astro config, the Netlify adapter can be used to translate this to the proper `_redirects` format.
```
bluwy, line 95: Do you mean you can use Netlify functions manually for a static app, e.g. creating the [netlify/functions folder](https://docs.netlify.com/functions/create/?fn-language=ts) yourself?

I suppose that's true for all adapters in general, e.g. Vercel Edge, Cloudflare workers, etc. So I'm not sure if it's needed to explain here.

Boston343: Correct. It was not clear to me that you could do so - the docs made it seem like you need the adapter. If it's the case for all of them then a general comment could be made somewhere about serverless functions.

bluwy: I'm not really sure where we can add that comment. We aren't quite saying that you need an adapter to use serverless functions for example, but more: you need an adapter for server-side rendering, and that server logic could be in serverless functions.

If you think some note should be added still, I think https://github.com/withastro/docs/issues would be the best place to get feedback on it.

ElianCodes:
Hey there @Boston343,

As pointed out by @bluwy, we currently don't think that this line should be added to the adapter README.

Thanks for your suggestion!

Issue #7667, opened 2023-07-15
Vercel adapter 3.7.0 or higher fails to build when `includeFiles` is present


raulfdm:
### What version of `astro` are you using?

2.8.3

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

not relevant

### Describe the Bug

I notice after upgrading `@astrojs/vercel` to `3.7.0` or higher because I use `includeFiles`, it throws an error:

```
Cannot read properties of undefined (reading 'root')
  File:
    /home/projects/withastro-astro-768xre/node_modules/.pnpm/@astrojs+vercel@3.7.2_astro@2.8.3_react@18.2.0/node_modules/@astrojs/vercel/dist/serverless/adapter.js:70:108
```

To reproduce, open the stackblitz repository and run `pnpm run build`.

If you switch to using `vercel36` (another vercel adapter version [3.6.0]), it'll work just fine.

### What's the expected result?

Not throwing an error while building.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-768xre?file=README.md

### Participation

- [X] I am willing to submit a pull request for this issue.

raulfdm:
https://github.com/withastro/astro/assets/12464600/cec3bde5-3040-4d01-aad6-9d0af89dd6ca

PR #7677, opened 2023-07-17
Fix vercel build error when passing `includeFiles`

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7667

The `_config` was accessed before it was assigned in the integration hook.

## Testing

Added a new test in `serverless-prerender`

## Docs

n/a. bug fix.


packages/integrations/vercel/test/fixtures/serverless-prerender/astro.config.mjs
```diff
@@ -2,6 +2,9 @@ import { defineConfig } from 'astro/config';
 import vercel from '@astrojs/vercel/serverless';
 
 export default defineConfig({
-	adapter: vercel(),
+	adapter: vercel({
+		// Pass some value to make sure it doesn't error out
+		includeFiles: ['package.json']
+	}),
```
ematipico, line 8: Shouldn't we assert that this file is copied over?

bluwy: I guess we could, but it won't be testing what the original issue had encountered.

bluwy: Added a test for it.

Issue #7137, opened 2023-05-19
Cloudflare adapter no longer includes environment variables using "import.meta.env" @build


davycoe:
### What version of `astro` are you using?

2.5.0

### Are you using an SSR adapter? If so, which one?

Cloudflare 6.3.0

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

import.meta.env.MY_VAR is no longer defined at build time when deployed to Cloudflare.

Example I've included:

- Clone https://github.com/davycoe/astro_cf_test
- Rename example.env to .env
- npm i && npm run dev
- You'll see the import.meta.env.VITE_ENV_TEST on home_page

Deploy to Cloudflare and set the environment variables:
- ENV_TEST_TWO = hello from process env!
- VITE_ENV_TEST = hello from vite env!

Once set and deployed only the process.env.ENV_TEST_TWO will render.

If you trust that I didn't mess anything up you can see the result deployed here:
https://astro-cf-test.pages.dev/

I've added some config options to get this process.env to work:

```
import { defineConfig } from "astro/config";

import cloudflare from "@astrojs/cloudflare";

// https://astro.build/config
export default defineConfig({
  output: "server",
  adapter: cloudflare(),
  vite: {
    define: {
      "process.env.ENV_TEST_TWO": JSON.stringify(process.env.ENV_TEST_TWO),
    },
  },
});
```

Is this expected/known?

### Link to Minimal Reproducible Example

https://github.com/davycoe/astro_cf_test

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #7679, opened 2023-07-17
Fix cloudflare runtime env var handling

bluwy:
## Changes

Follow up from https://github.com/withastro/astro/pull/7651#discussion_r1263856991
Fix https://github.com/withastro/astro/issues/7137

Fix cloudflare runtime env var handling. We're using Vite's `ssr.target = 'webworker'` config, and it's converting our `process.env.RUNTIME_VAR` to `({}).RUNTIME_VAR`, causing it to not work.

This also tweaked Astro's env handling for the `PRERENDER` test, as it's indirectly passing with the bug above. Since it was passed a boolean value, I had to update `vite-plugin-env` to handle it too.

## Testing

Un-skipped a test that revealed this bug.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -94,7 +94,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 					// Cloudflare env is only available per request. This isn't feasible for code that access env vars
 					// in a global way, so we shim their access as `process.env.*`. We will populate `process.env` later
-					// in it's fetch handler.
+					// in its fetch handler.
 					vite.define = {
 						'process.env': 'process.env',
 						...vite.define,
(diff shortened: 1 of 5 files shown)


packages/integrations/cloudflare/src/index.ts
```diff
@@ -91,6 +91,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
 					}
 					vite.ssr ||= {};
 					vite.ssr.target = 'webworker';
+
+					// Cloudflare env is only available per request. This isn't feasible for code that access env vars
+					// in a global way, so we shim their access as `process.env.*`. We will populate `process.env` later
+					// in it's fetch handler.
```
ematipico, line 97:
```suggestion
					// in its fetch handler.
```

ematipico (approved):

Issue #7663, opened 2023-07-15
Unable to find the manifest for the module file: entry.index.astro.mjs


josfaber:
### What version of `astro` are you using?

2.8.3

### Are you using an SSR adapter? If so, which one?

node

### What package manager are you using?

npm

### What operating system are you using?

macOS Ventura 13.4.1

### What browser are you using?

Chrome 114.0.5735.198

### Describe the Bug

On a new project created with `npm create astro@latest` and mode is `hybrid` (installed node with `npx astro add node`), as soon as I set the build to be splitted, I get the following error (full cli output):
```
> astro build

09:34:27 AM [content] No content directory found. Skipping type generation.
09:34:27 AM [build] output target: hybrid
09:34:27 AM [build] deploy adapter: @astrojs/node
09:34:27 AM [build] Collecting build info...
09:34:27 AM [build] Completed in 26ms.
09:34:27 AM [build] Building hybrid entrypoints...
09:34:28 AM [build] Completed in 1.15s.

 prerendering static routes
Server listening on http://127.0.0.1:3000
 error   Unable to find the manifest for the module file:///Users/jos/workspace/astro-test-1/dist/server/src/pages/entry.index.astro.mjs?time=1689406468435. This is unexpected and likely a bug in Astro, please report.
Error: Unable to find the manifest for the module file:///Users/jos/workspace/astro-test-1/dist/server/src/pages/entry.index.astro.mjs?time=1689406468435. This is unexpected and likely a bug in Astro, please report.
    at generatePages (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/core/build/generate.js:112:19)
    at async staticBuild (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/core/build/static-build.js:94:7)
    at async AstroBuilder.build (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/core/build/index.js:118:5)
    at async AstroBuilder.run (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/core/build/index.js:149:7)
    at async build (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/core/build/index.js:37:3)
    at async build (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/cli/build/index.js:7:3)
    at async runCommand (file:///Users/jos/workspace/astro-test-1/node_modules/astro/dist/cli/index.js:117:7)
[...]

PR #7680, opened 2023-07-17
fix: throw an error for incorrect configuration

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/7663

`build.split` is - at the moment - meant to be used only when `output` is `"server"`. This PR throw an error in case the configuration doesn't meet the criteria. 

## Testing

I locally tested that the error is thrown

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/build/index.ts
+++ b/packages/astro/src/core/build/index.ts
@@ -233,7 +233,7 @@ class AstroBuilder {
 		if (config.build.split === true) {
 			if (config.output !== 'server') {
 				throw new Error(
-					'The option `build.split` can be used only when `output` is set to `"server"`.'
+					'The option `build.split` can only be used when `output` is set to `"server"`.'
 				);
 			}
 		}


packages/astro/src/core/build/index.ts
```diff
@@ -229,6 +229,14 @@ class AstroBuilder {
 				);
 			}
 		}
+
+		if (config.build.split === true) {
+			if (config.output !== 'server') {
+				throw new Error(
+					'The option `build.split` can be used only when `output` is set to `"server"`.'
```
bluwy, line 236:
```suggestion
					'The option `build.split` can only be used when `output` is set to `"server"`.'
```

I think the `only` in this spot flows a bit natural 🤔

bluwy (approved):
One small grammar nit. LGTM!

In the future, we could also move these validation to https://github.com/withastro/astro/blob/d69fe3a8d2e1c5d49fe08aa3d974fc37cee9ca93/packages/astro/src/core/config/config.ts#L25-L30 once I've finished up the config refactoring.

PR #7682, opened 2023-07-17
Replace openConfig with resolveConfig

bluwy:
## Changes

**This PR merges to the js-api branch**

- Changes internal `openConfig` with `resolveConfig` (new API)

NOTE: There's one more place using `openConfig`, that is `dev/restart.ts`. I'd like refactor how it passes around configs during restarts in general, so holding that change for the next PR. After that, we can remove `openConfig`.

## Testing

Existing tests should pass.

## Docs

n/a. internal change.

### DIFF
--- a/packages/astro/test/test-utils.js
+++ b/packages/astro/test/test-utils.js
@@ -129,9 +129,6 @@ export async function loadFixture(inlineConfig) {
 	 * command functions below to prevent tests from polluting each other.
 	 */
 	const getSettings = async () => {
-		if (isWindows) {
-			console.log('VERCEL', config.root, root);
-		}
 		let settings = createSettings(config, root);
 		if (config.integrations.find((integration) => integration.name === '@astrojs/mdx')) {
 			// Enable default JSX integration. It needs to come first, so unshift rather than push!


bluwy:
Fixed it! Awaiting https://github.com/withastro/astro/pull/7701 and https://github.com/withastro/astro/pull/7704 to be merged first, then rebasing it, and should be good.

packages/integrations/mdx/test/mdx-plugins.test.js
```diff
@@ -96,7 +96,7 @@ describe('MDX plugins', () => {
 	it('ignores string-based plugins in markdown config', async () => {
 		const fixture = await buildFixture({
 			markdown: {
-				remarkPlugins: [['remark-toc']],
+				remarkPlugins: [['remark-toc', {}]],
```
bluwy, line 99: The tuple needs two values or it won't pass config validation. This should be what users have to do right now too it seems.

packages/integrations/mdx/test/mdx-get-static-paths.test.js
```diff
@@ -23,7 +23,7 @@ describe('getStaticPaths', () => {
 		const $ = cheerio.load(html);
 		expect($('p').text()).to.equal('First mdx file');
 		expect($('#one').text()).to.equal('hello', 'Frontmatter included');
-		expect($('#url').text()).to.equal('/src/content/1.mdx', 'url is included');
+		expect($('#url').text()).to.equal('src/content/1.mdx', 'url is included');
```
bluwy, line 26: The `base` handling in the tests previously patches the config, causing the leading slash. In a normal Astro app, there shouldn't be a leading slash.

packages/integrations/mdx/test/mdx-plugins.test.js
ematipico: Is it a change in the public APIs?

ematipico (approved):

packages/integrations/mdx/test/mdx-plugins.test.js
bluwy: No it shouldn't be. Currently if I use the same configuration in the `basics` example, I get

```
10:06:40 PM [astro] Unable to load /Users/bjorn/Work/oss/astro/examples/basics/astro.config.mjs

[config] Astro found issue(s) with your configuration:
  ! markdown.remarkPlugins.0  Invalid input.
```

It appears now because the tests will run the zod schema validation for each fixture options.

PR #7684, opened 2023-07-17
refactor: move page rendering in one single function

ematipico:
## Changes

Refactor.

I moved the rendering of the page inside a unique function. I called `tryRenderPage` to make it more clear that that's an operation that could throw an error. In fact, there's a place where we handle the thrown error.

## Testing

I updated the unit tests. Current CI should pass

## Docs

N/A


bluwy (approved):
Love this refactor! Nice to see 3 places now merge into 1.

There's also a flaky test in CI, but I think that should already be fixed in `main`.

PR #7687, opened 2023-07-17
Update esbuild target for Deno

kwhinnery:
## Changes

Updates the [build target](https://esbuild.github.io/api/#target) for esbuild from `es2020` to `esnext` to allow for more advanced language features that are available in Deno.

## Testing

Added a test that includes a language feature supported in `esnext`, in this case a top-level await.

## Docs

This change doesn't require updates to the documentation - a Deno user would reasonably assume that all runtime features that work in Deno would work during the Astro build process as well.


kwhinnery:
There's a set of tests failing in an area of the code I didn't touch - are we seeing this on other PRs as well? Thanks!

Princesseuh:
> There's a set of tests failing in an area of the code I didn't touch - are we seeing this on other PRs as well? Thanks!

Yep, sorry - my fault. We had a regression in our Markdoc integration. I went ahead and updated your branch.

bluwy (approved):
I think this is probably fine as Deno 1.0 already supports many modern syntax from the start 🤔

bluwy:
I was hoping to get another eye on this from another maintainer, in case I miss something, but looks like it should be safe. Will merge this one.

kwhinnery:
Thanks a ton @bluwy!

Issue #7407, opened 2023-06-16
Regression: @astrojs/vercel/edge not compatible with imageService


alexanderniebuhr:
### What version of `astro` are you using?

2.6.4

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

> Hi, I am trying to enable vercel's imageService, which is said to be supported on the edge in the @astrojs/vercel docs ([link](https://docs.astro.build/en/guides/integrations-guide/vercel/#imageservice)). To do that, I am required to enable experimental: { assets: true }, however by doing that I think that some node packages get imported into the build and that causes it to fail because of the edge runtime

Regression. https://github.com/withastro/astro/issues/7057 seems to be back.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/verce-edge-image-service-2xuprn?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Tried to figure out where the nodejs modules are imported, and it's surprisingly hard. I can't seem to find where in the import chain it's being used. I made this Vite plugin to identify what files are included in a chunk, but it's still as hard:

```js
{
  name: 'track',
  renderChunk(code, chunk) {
    console.log(chunk)
  }
}
```

Adding this in the `astro:build:setup` hook of the vercel integration didn't help either:

```js
vite.build ||= {};
vite.build.rollupOptions ||= {};
vite.build.rollupOptions.output ||= {};
// Disable optimization as the output is bundled by esbuild, which nullifies the benefit
if (Array.isArray(vite.build.rollupOptions.output)) {
	vite.build.rollupOptions.output.forEach((output) => {
		output.hoistTransitiveImports = false;
	});
} else {
	vite.build.rollupOptions.output.hoistTransitiveImports = false;
}
```

PR #7691, opened 2023-07-17
fix(assets): Remove Node deps from index of assets

Princesseuh:
## Changes

I'm not sure why this fixes it, but it does, sorry @bluwy  

Fix https://github.com/withastro/astro/issues/7407

## Testing

Tested manually, both in local and actually on Vercel

## Docs

N/A

### DIFF
--- a/.changeset/shy-garlics-join.md
+++ b/.changeset/shy-garlics-join.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix not being able to build on Vercel Edge when `astro:assets` was enabled even when using a non-Node image service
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):

bluwy:
I'm just glad this is fixed 🎉

Princesseuh:
> Bummer, I was using `imageMetadata` but it's no longer exported.

Arf, I feared this was gonna happen, ha. We kinda on purpose don't expose it too publicly since it's internal and doesn't run on non-Node SSR hosts, but I'll make a PR exporting it from `astro/assets/utils`, it's fine there

birtles:
Thank you so much 🙏

PR #7692, opened 2023-07-17
nit: use `node:` prefix everywhere

Princesseuh:
## Changes

As we head into a future of multiple runtimes, climate change, AI taking over etc. we might want to run Astro on other runtimes. The `node:` prefix allows tools to identify Node deps more easily. It's also useful for us, because we sometimes use Node deps where we shouldn't and the prefix makes it easier to see at a glance where Node is hiding

## Testing

Tests should all pass!

## Docs

N/A

### DIFF
--- a/.github/scripts/bundle-size.mjs
+++ b/.github/scripts/bundle-size.mjs
@@ -1,5 +1,5 @@
 import { build } from 'esbuild';
-import { existsSync } from 'fs';
+import { existsSync } from 'node:fs';
 
 const CLIENT_RUNTIME_PATH = 'packages/astro/src/runtime/client/';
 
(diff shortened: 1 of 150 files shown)


natemoo-re (approved):

Princesseuh:
Huh.. The Content Collections that tests no Node built-in snuck-in is complaining, but not because I added a Node dep, but because it didn't test for `fs`, only `node:fs`, ha

Princesseuh:
`node:fs` is the only Node import that gets in the bundle... There's only two files in Astro that only have this import, as far as I can tell

- `packages/astro/src/core/config/timer.ts`
- `packages/astro/src/core/compile/style.ts`

matthewp (approved):

PR #7697, opened 2023-07-18
feat: allow more type of props to be parsed in the scanner

y-nk:
## Changes

- What does this change?

It widen the range of the `vite-astro-scanner` plugin to allow parsing of more properties, and of type `string`.

This is mainly for plugin/integrations authors. There are many behaviors that could be made if we could let users expose some "route based props" which could be later picked up in integrations.

On the side, i'd like to bring another approach to this using recma instead of regular expression and split of strings which may break easily.

## Testing, Docs

The goal of this PR is to have an open discussion, not necessarily to be merged. I'm happy to add tests/docs if this can be merged as-is.

### DIFF
--- a/packages/astro/src/core/build/types.ts
+++ b/packages/astro/src/core/build/types.ts
@@ -21,13 +21,14 @@ export type StylesheetAsset =
 	| { type: 'external'; src: string };
 
 export interface PageBuildData {
-	component: ComponentPath;
-	route: RouteData;
-	moduleSpecifier: string;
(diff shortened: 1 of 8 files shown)


matthewp:
Seems similar to https://github.com/withastro/astro/pull/8260. I think we agree on having some sort of way to have custom metadata in pages would be useful.  But this is a big enough change that it needs to go through the RFC process: https://github.com/withastro/roadmap/discussions

PR #7632, opened 2023-07-12
feat(@astrojs/netlify): edge middleware support

ematipico:
## Changes

This PR adds support for Netlify edge middleware. The edge middleware, in Netlify, is simply an edge function. The function will call the server origin - the page itself - and eventually return the response.

Here is some context and documentation: https://docs.netlify.com/edge-functions/api/#return-a-redirect

The code is very similar to the one implemented in Vercel. It was slightly adapted to Netlify.

## Testing

- new test cases added
- cut a preview release and see if I can use it in a pet project

## Docs

 /cc @withastro/maintainers-docs for feedback!

PR #7700, opened 2023-07-18
Add changeset & docs for next `@astrojs/netlify` release

delucis:
## Changes

- Follow-up PR to #7698
- Adds docs and changeset for #7632, which we removed to avoid releasing before docs were polished.

Would love @ematipico to make sure none of what I wrote is lies 😀 

## Testing
N/A — docs only.

## Docs

### DIFF
--- a/packages/integrations/netlify/README.md
+++ b/packages/integrations/netlify/README.md
@@ -98,7 +98,7 @@ Netlify Edge Functions provide a [context object](https://docs.netlify.com/edge-
 
 To expose values from this context to your site, create a `netlify-edge-middleware.ts` (or `.js`) file in your project’s [source directory](https://docs.astro.build/en/reference/configuration-reference/#srcdir). This file must export a function that returns the data to add to [Astro’s `locals` object](https://docs.astro.build/en/guides/middleware/#locals), which is available in middleware and Astro routes.
 
-In this example, `visitorCountry` and `hasEdgeMiddleware` would both be added to the Astro’s `locals` object:
+In this example, `visitorCountry` and `hasEdgeMiddleware` would both be added to Astro’s `locals` object:
 
 ```ts
 // src/netlify-edge-middleware.ts
--- a/packages/integrations/netlify/README.md
+++ b/packages/integrations/netlify/README.md
@@ -57,7 +57,7 @@ If you prefer to install the adapter manually instead, complete the following tw
 
 ### Edge Functions
 
-Netlify has two serverless platforms, Netlify Functions and [Netlify's experimental Edge Functions](https://docs.netlify.com/netlify-labs/experimental-features/edge-functions/#app). With Edge Functions your code is distributed closer to your users, lowering latency.
+Netlify has two serverless platforms, [Netlify Functions](https://docs.netlify.com/functions/overview/) and [Netlify Edge Functions](https://docs.netlify.com/edge-functions/overview/). With Edge Functions your code is distributed closer to your users, lowering latency.
 
 To deploy with Edge Functions, use `netlify/edge-functions` in the Astro config file instead of `netlify/functions`.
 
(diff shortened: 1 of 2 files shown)


packages/integrations/netlify/README.md
```diff
@@ -72,6 +72,56 @@ export default defineConfig({
 });
 ```
 
+### Run middleware in Edge Functions
+
+When deploying to Netlify Functions, you can choose to use an Edge Function to run your Astro middleware.
+
+To enable this, set the `build.excludeMiddleware` Astro config option to `true`:
+
+```js ins={9}
```
ematipico, line 81:
```suggestion
```js
```

Is it needed?

packages/integrations/netlify/README.md
```diff
@@ -72,6 +72,56 @@ export default defineConfig({
...
+
+To expose values from this context to your site, create a `netlify-edge-middleware.ts` (or `.js`) file in your project’s [source directory](https://docs.astro.build/en/reference/configuration-reference/#srcdir). This file must export a function that returns the data to add to [Astro’s `locals` object](https://docs.astro.build/en/guides/middleware/#locals), which is available in middleware and Astro routes.
+
+In this example, `visitorCountry` and `hasEdgeMiddleware` would both be added to the Astro’s `locals` object:
+
+```ts
+// src/netlify-edge-middleware.ts
+import type { Context } from 'https://edge.netlify.com';
+
+export default function ({ request, context }: { request: Request; context: Context }) {
```
ematipico, line 107:
```suggestion
export default function ({ request, context }: { request: Request; context: Context }): object {
```

ematipico (approved):
Looks good!

packages/integrations/netlify/README.md
delucis: It will be picked up in the docs site to highlight that line as an “insertion”! Doesn’t do anything in regular Markdown on GitHub/npm.

packages/integrations/netlify/README.md
delucis: I removed this because I thought `object` was considered bad practice in TypeScript — but now I’m seeing TypeScript ESLint actually relaxed this in v5: https://github.com/typescript-eslint/typescript-eslint/pull/3818

I’m still not sure it adds much though? `object` gives you very weak type checking. `Record<string, unknown>` would be slightly better (prevents you returning a `Map` or a function for example):

```suggestion
export default function ({
  request,
  context,
}: {
  request: Request;
  context: Context;
}): Record<string, unknown> {
```

But not sure it’s worth letting Prettier explode one line into seven 😅

ematipico: I don't have strong opinions. The most important thing is that we should communicate to the user that they can type the retuned type as they want, as long as the returned type is an object :)

delucis: OK, cool. I think we have it covered then. We say both “a function that returns the data to add to Astro’s `locals` object” above and “must return a JSON-serializable object” below the code sample, plus include a comment inside the code sample about returning serializable data, so I think that should be enough.

packages/integrations/netlify/README.md
```diff
@@ -72,6 +72,56 @@ export default defineConfig({
...
+});
+```
+
+#### Pass edge context to your site
+
+Netlify Edge Functions provide a [context object](https://docs.netlify.com/edge-functions/api/#netlify-specific-context-object) including metadata about the request, such as a user’s IP, geolocation data, and cookies.
+
+To expose values from this context to your site, create a `netlify-edge-middleware.ts` (or `.js`) file in your project’s [source directory](https://docs.astro.build/en/reference/configuration-reference/#srcdir). This file must export a function that returns the data to add to [Astro’s `locals` object](https://docs.astro.build/en/guides/middleware/#locals), which is available in middleware and Astro routes.
+
+In this example, `visitorCountry` and `hasEdgeMiddleware` would both be added to the Astro’s `locals` object:
```
Yan-Thomas, line 101: I feel like there shouldn't be a "the" here?
```suggestion
In this example, `visitorCountry` and `hasEdgeMiddleware` would both be added to Astro’s `locals` object:
```

Yan-Thomas (approved):
Looks great @delucis, I just have a small nit, overall LGTM!

delucis:
OK, let’s get this released 🚀 

Thanks for your patience with me slowing you down @ematipico!

Issue #7322, opened 2023-06-07
Astro redirects not working


Erlandas:
### What version of `astro` are you using?

2.6.1

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

yarn

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

Tried out new experimental redirects. 
```javascript
experimental: {
    redirects: true
  },
redirects: {
    '/a': '/b'
  }
 ```
Went trough all paths possible (for both new page and old page with slash and without) Yet visiting page "/a" no redirects being made (both in dev and production).

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-njit3y

### Participation

- [ ] I am willing to submit a pull request for this issue.

**Comment by jlarmstrongiv** [comment]
The types and validation for `redirects` are also incorrect

**Comment by matthewp** [comment]
I can recreate the dev problem. Fixing now.

**Comment by pijusz** [comment]
Would like to reopen this @matthewp
In astro docs, redirect with non-301 status doesn't work.

Goal is to be able to redirect with parameters status 308 (f.e. `/b?type=b`, `/a?type=b`)
```javascript
  output: "hybrid",
  adapter: vercel(),
  experimental: {
    redirects: true,
  },
  redirects: {
    "/b": {
      destination: "/a",
      status: 308,
    },
  },
```

received error:
```
[config] Astro found issue(s) with your configuration:
  ! redirects./b  Expected string, received object.
```

bluwy:
Can confirm this is still happening. I'm refactoring some tests that now validates the config, and it fails with the error too. We need to fix the validation here: https://github.com/withastro/astro/blob/ed20154a5c89f07e87784c5fe80a028ceef741fa/packages/astro/src/core/config/schema.ts

PR #7701, opened 2023-07-18
Fix redirects map object-form value validation

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7322

Extend `redirects` config validation to pass configs like:
```js
	redirects: {
		'/': {
			status: 302,
			destination: '/home',
		},
	},
```

## Testing

Tested manually for now. I'm planning to add config validation tests in the `js-api` branch and PR so we catch this in the future.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/core/config/schema.ts
+++ b/packages/astro/src/core/config/schema.ts
@@ -149,7 +149,12 @@ export const AstroConfigSchema = z.object({
 			.optional()
 			.default({})
 	),
-	redirects: z.record(z.string(), z.string()).default(ASTRO_CONFIG_DEFAULTS.redirects),
+	redirects: z
+		.record(
+			z.string(),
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/config/schema.ts
```diff
@@ -149,7 +149,12 @@ export const AstroConfigSchema = z.object({
 			.optional()
 			.default({})
 	),
-	redirects: z.record(z.string(), z.string()).default(ASTRO_CONFIG_DEFAULTS.redirects),
+	redirects: z
+		.record(
+			z.string(),
+			z.union([z.string(), z.object({ status: z.number(), destination: z.string() })])
```
bluwy, line 155: Not sure if we need the extra strictness for allowed `status` numbers. I've made it only a `z.number()` for now.

ematipico (approved):

natemoo-re (approved):

PR #7702, opened 2023-07-18
Export AstroConfigType properly as type and not as a Zod schema value

shishkin:
This makes the intended usage of Zod generated types actually possible

## Changes

- ~~Export AstroConfigSchema as value instead of type~~
- Export AstroConfigType properly as type and not as a Zod schema value

## Testing

Previously, type autocomplete was broken on `AstroConfig` (e.g. inside an integration hook). With this change the zod schema value is retained and can be used to generate the proper type, so autocomplete works (e.g. in `packages/astro/src/@types/astro.ts` for `AstroConfig` type).

## Docs

No docs impacted. This fixes previously intended behavior that wasn't working.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -13,7 +13,6 @@ import type { AddressInfo } from 'node:net';
 import type * as rollup from 'rollup';
 import type { TsConfigJson } from 'tsconfig-resolver';
 import type * as vite from 'vite';
-import type { z } from 'zod';
 import type { SerializedSSRManifest } from '../core/app/types';
 import type { PageBuildData } from '../core/build/types';
 import type { AstroConfigType } from '../core/config';


shishkin:
@ematipico thanks for prompting me. I actually think my fix doesn't solve the problem I had.

The issue is that `AstroConfig` typing is not working in 3rd party integrations:

Screenshot 2023-07-18 at 18 40 04

I see only `integrations` defined in [@types/astro.ts](https://github.com/withastro/astro/blob/a8cbd7c71a5ca319c0ebc6e9649428c0d4afc8fe/packages/astro/src/%40types/astro.ts#L1306-L1311).

I thought this was due to the zod schema instance being exported as a type while it must be a value for the `z.output<typeof ...>` magic to work.

However, now even with my change I'm not able to get the autocompletion in an external integration project, though types now work internally within the astro monorepo for me.

It seems the zod type generation combined with interface inheritance doesn't survive in the exported type definitions. I finally had to resort to _zod way of inheritance_ to generate the final `AstroConfig` type.

delucis:
I’m not sure this is currently not working. For example, in an integration I maintain, I already get autocomplete:
image

shishkin:
> @delucis: I’m not sure this is currently not working. For example, in an integration I maintain, I already get autocomplete

I've tested your [delucis/astro-embed](https://github.com/delucis/astro-embed) repo and indeed autocomplete worked out of the box. But when I upgraded typescript to 5.1.6 and switched tsconfig [module resolution](https://www.typescriptlang.org/tsconfig#moduleResolution) to `nodenext` I only saw `integrations` in the autocomplete.

delucis:
Good to know! I tested in https://github.com/withastro/starlight but still helpful to know this may be something introduced with specific config/dependency versions.

shishkin:
Is there an intention to merge this? Even if it's not an immediate concern for developers not using module resolution `bundle` or `node16` there is still an error in how `AstroConfigSchema` is imported. If it's used as a value it can't be imported as a type.

bluwy:
@shishkin Could the issue be that our dts exported is not node16 compatible? There's work regarding it there at https://github.com/withastro/astro/pull/7787. I don't think we'd want to export Zod schema as value, that's leaning towards internal API. And we probably don't want runtime in the `astro.ts` file too.

shishkin:
Yes, that makes sense. The issue is due to [changes in module resolution](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler) `bundler`/`node16`. I can try to update the PR to ensure that the config package only exports a proper type instead of a zod schema value.

shishkin:
@bluwy I've updated the PR to export the type and not the schema value. Hope this makes sense now and can be merged.

PR #7703, opened 2023-07-18
refactor: unify `renderPage` and `callEndpoint` in one single function

ematipico:
## Changes

This PR renames `tryRenderPage` to `tryRenderRoute` and moves the rendering of pages and endpoints under this same very function.

This requires refactoring inside the `core/app/index` module. 

In the next PR I will move files inside `core/render/dev` inside the `vite-plugin-astro-server`.

## Testing

Updated current tests. Current CI should pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/render/core.ts
+++ b/packages/astro/src/core/render/core.ts
@@ -3,15 +3,17 @@ import type {
 	ComponentInstance,
 	MiddlewareHandler,
 	MiddlewareResponseHandler,
+	RouteType,
 } from '../../@types/astro';
 import { renderPage as runtimeRenderPage } from '../../runtime/server/index.js';
 import { attachToResponse } from '../cookies/index.js';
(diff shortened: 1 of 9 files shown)


packages/astro/src/core/render/core.ts
```diff
@@ -101,26 +107,50 @@ export async function tryRenderPage<MiddlewareReturnType = Response>(
...
+			const result = await callEndpoint(
+				mod as any as EndpointHandler,
+				env,
+				renderContext,
+				onRequest
+			);
+			return result;
+		}
+		default:
+			throw new Error('');
```
bluwy, line 146: Would a later PR populate this error message?

bluwy (approved):

packages/astro/src/core/render/core.ts
ematipico: I will update the error in this PR. Totally forgot it. Thank you!

PR #7704, opened 2023-07-18
Fix absolute path handling for config validation in windows

bluwy:
## Changes

Fix passing absolute paths to `root`, `srcDir`, `publicDir`, `outDir`, `cacheDir`, `build.client`, and `build.server` on Windows. We currently have code like this:

https://github.com/withastro/astro/blob/b3b9fc52b33ae8d9bd3f205758051814a1ed4add/packages/astro/src/core/config/schema.ts#L251

And they fail on Windows because `val` looks like `C:\\User\\bob`. The `C:` looks like a protocol when passed to `new URL` and that creates an invalid final URL. Switching to `path.resolve` fixes this for now.

## Testing

Tested in https://github.com/withastro/astro/pull/7682 where I passed the absolute `root` path in tests, and it failed. The [fa32ff0](https://github.com/withastro/astro/pull/7682/commits/fa32ff029c68272f57b4299501f6f083152daed2) commit there fixes it, which I ported it to here to `main` first.

Besides that, no actual tests here yet, but CI should continue to pass.

Also thanks `@hippotastic` for testing that this works.

## Docs

The docs for `srcDir`, `publicDir`, `outDir`, and `cacheDir` mentioned that they accept absolute paths. `root`, `build.client`, and `build.server` don't exactly mention it, but it would be nice if they work still for consistency.

### DIFF
--- a/.changeset/loud-hairs-tell.md
+++ b/.changeset/loud-hairs-tell.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix absolute path handling when passing `root`, `srcDir`, `publicDir`, `outDir`, `cacheDir`, `build.client`, and `build.server` configs in Windows
--- a/packages/astro/src/core/config/config.ts
+++ b/packages/astro/src/core/config/config.ts
@@ -4,7 +4,7 @@ import type { AstroConfig, AstroUserConfig, CLIFlags } from '../../@types/astro'
 import * as colors from 'kleur/colors';
 import fs from 'node:fs';
 import path from 'node:path';
-import { fileURLToPath, pathToFileURL } from 'node:url';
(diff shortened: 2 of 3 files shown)


ematipico (approved):

Geoffrey-Pliez:
Hi,

Is it normal for the pathname of these paths to start with "/"?
Because it doesn't allow you to use the pathname in other projects like [this one](https://github.com/firebase/firebase-tools/pull/5709)

bluwy:
I don't think that's common on Windows. If you're dabbling with path `file:` URLs, you should use `fileUrlToPath` or `pathToFileUrl` to make sure they work across OS.

Geoffrey-Pliez:
Ok, I'll try to explore the idea.
Thanks.

Issue #7705, opened 2023-07-18
@astrojs/markdoc integration fails import


MrMacStripe:
### What version of `astro` are you using?

2.8.4

### Are you using an SSR adapter? If so, which one?

none

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

irrelevant

### Describe the Bug

After updating Astro to 2.8.4 and @astrojs/markdoc to 0.4.2 the Astro build command fails.

`[vite] Error when evaluating SSR module /Users/.../astro.config.mjs: failed to import "@astrojs/markdoc"
|- file:///Users/.../node_modules/@astrojs/markdoc/dist/content-entry-type.js:6
import { emitESMImage } from "astro/assets";
         ^^^^^^^^^^^^
SyntaxError: The requested module 'astro/assets' does not provide an export named 'emitESMImage'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)

04:40:10 PM [astro] Unable to load /Users/.../astro.config.mjs

 error   The requested module 'astro/assets' does not provide an export named 'emitESMImage'
file:///Users/.../node_modules/@astrojs/markdoc/dist/content-entry-type.js:6
import { emitESMImage } from "astro/assets";
         ^^^^^^^^^^^^
SyntaxError: The requested module 'astro/assets' does not provide an export named 'emitESMImage'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)`

### What's the expected result?

The build should process without error

### Link to Minimal Reproducible Example

cant reproduce this on a new one

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #7706, opened 2023-07-18
Fix/markdoc assets

Princesseuh:
## Changes

A previous PR removed `emitESMImage` from `astro/assets/index.js` to fix Vercel Edge support, however, the Markdoc integration imported that. We didn't catch it for two reasons:

- Due to `moduleResolution: 'node'` not being able to get the types properly, there's a `@ts-expect-error` there, so type checking didn't fail
- For CI performance reason, we don't run the CI of every single integration on all core changes, so the tests fails weren't caught

Fix https://github.com/withastro/astro/issues/7705

## Testing

Tests should pass!

## Docs

N/A

### DIFF
--- a/.changeset/smart-hairs-lick.md
+++ b/.changeset/smart-hairs-lick.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/markdoc': patch
+'astro': patch
+---
+
+Fix Markdoc integration not being able to import `emitESMImage` from Astro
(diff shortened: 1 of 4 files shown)


bluwy (approved):
It might be a slight breaking change for markdoc still, since it now requires the latest Astro version to work, but I think it's better that reverting the other fix that causes another regression.

Princesseuh:
> It might be a slight breaking change for markdoc still, since it now requires the latest Astro version to work, but I think it's better that reverting the other fix that causes another regression.

Surface area is small, it's only broken in specifically 2.8.4, so users can choose to stay on 2.8.3 or update the Markdoc integration. It's a tiny bit annoying, but it's not *too bad*

ematipico (approved):

PR #7707, opened 2023-07-18
Determine hoisted scripts via AST analysis

### LINKED ISSUE #7673: Hoisted scripts break when components are re-exported from a typescript file

2.8.3

None

npm

Linux

Firefox

### Describe the Bug

When components are re-exported, like so:
```ts
import Component from './Component.astro';
import Layout from './Layout.astro';

export { Component, Layout };
```

Astro loses track of which components are included on which pages, leading to hoisted scripts being included on a page where the component does not exist.

This is impactful on a monorepo that includes a library of Astro components, as there is no easy way to avoid re-exporting them.

Re-exporting components should function identically to importing them directly.

https://stackblitz.com/edit/github-15jfdd?file=src%2Fpages%2Findex.astro

### Participation

ottomated:
## Changes

- Fixes #7673 
- Previous behavior: If a component is anywhere in the import tree of the current page, include its script in the hoisted bundle
- New behavior: Components must be actually imported by the current page for their hoisted scripts to be included.

## TODO

- [ ] Investigate performance - there are opportunities for early returns and caching.
- [ ] Attempt to move MDX-specific logic out of the core.
- [ ] Add documentation
- [ ] Make this work for `<style>` tags as well
- [x] ~~This will current break in the case where a component is imported by another component. Fix would be to switch import tracking to the importer in this case?~~
- [x] ~~Handle dynamic imports~~
- [x] ~~depthsToExportNames should be a Map<number, string[]> because technically the same component could be exported twice from the same file under different names~~

## Testing

A new test was added that tests various combinations of re-exporting style and checks to see if the built pages contain the expect components and no others.

## Docs

TODO: document the exact behavior in the hoisted scripts section


ottomated:
This would break the edge case where a component is exported indirectly, like so:
```
import Component from './Component.astro';

export default getComponent() {
  return Component;
}
```

Not sure if that's an issue or not.

matthewp:
As long as this doesn't cause regressions I think a small change to support the common re-export pattern makes sense. I would just want the code cleaned up a little bit, move the AST logic into a separate, clearly named function.

ottomated:
Last change is the component tracking switcher needs to handle MDX. @matthewp which file extensions specifically can render astro components? Is it just .astro and .mdx?

packages/astro/src/core/build/plugins/plugin-analyzer.ts
```diff
@@ -17,6 +19,102 @@ function isPropagatedAsset(id: string) {
...
+			}
+		}
+	}
+	if (!importNames.length) return 'no';
+
+	// If the component is imported by another component, assume it's in use
+	// and start tracking this new component now
+	if (parentInfo.id.endsWith('.astro')) {
+		exportNames.push('default');
+	} else if (parentInfo.id.endsWith('.mdx')) {
```
ottomated, line 111: I assume this is unacceptable - mdx logic shouldn't be in the core. Unsure on how to resolve.

lilnasy:
If we only need to scan the imports, maybe we can avoid creating and walking the full AST by using es-module-lexer instead.

ottomated:
> If we only need to scan the imports, maybe we can avoid creating and walking the full AST by using es-module-lexer instead.

Well, the AST is already created, but this is worth looking into I suppose

packages/astro/src/core/build/plugins/plugin-analyzer.ts
matthewp: Yeah, given we already have this in the code. Not ideal.

matthewp:
This looks really good! I'm going to update the branch to see if that kicks off CI to run fully.

packages/astro/test/hoisted-imports.test.js
```diff
@@ -0,0 +1,93 @@
...
+		const html = await fixture.readFile(page);
+		const $ = cheerio.load(html);
+		const scriptText = [];
+
+		const importRegex = /import\s*?['"]([^'"]+?)['"]/g;
+		async function resolveImports(text) {
+			const matches = text.matchAll(importRegex);
+			for (const match of matches) {
+				const importPath = match[1];
+				console.log('- ', importPath);
```
bluwy, line 24: Should we remove this `console.log`?
```suggestion
```

packages/astro/src/core/build/plugins/plugin-analyzer.ts
```diff
@@ -29,17 +127,49 @@ export function vitePluginAnalyzer(internals: BuildInternals): VitePlugin {
 		>();
 
 		return {
-			scan(this: PluginContext, scripts: AstroPluginMetadata['astro']['scripts'], from: string) {
+			async scan(
+				this: PluginContext,
+				scripts: AstroPluginMetadata['astro']['scripts'],
+				from: string
+			) {
+				// console.log('\n\n==== ', from, ' ====\n\n');
```
bluwy, line 135:
```suggestion
```

packages/astro/src/core/build/plugins/plugin-analyzer.ts
```diff
@@ -29,17 +127,49 @@ export function vitePluginAnalyzer(internals: BuildInternals): VitePlugin {
...
+
+							const doesImport = await doesParentImportChild.call(
+								this,
+								parentInfo,
+								childInfo,
+								childExportNames
+							);
+
+							if (doesImport === 'no') {
+								// console.log(`[astro] ${parentInfo.id} does not import any of ${childExportNames} from ${childInfo?.id}`);
```
bluwy, line 166:
```suggestion
```

bluwy (approved):
Clean tests! I think our strategy for hoisting by statically analyzing things won't always be correct, so having a best-effort implementation for now is fine by me.

There would be a slight performance hit with this as reading the `ast` in Rollup introduces an acorn parse, but running some build tests doesn't yield a big time difference, so maybe it is negligible.

packages/astro/src/core/build/plugins/plugin-analyzer.ts
bluwy: I haven't tested this yet, but shouldn't the `ast` already contain these exports and added them to `exportNames`?

.changeset/real-drinks-melt.md
```diff
@@ -0,0 +1,26 @@
...
+
+Then let's say you have a page that uses just the Accordion like so:
+
+```astro
+---
+import { Accordion } from '@matthewp/my-astro-lib';
+---
+```
+
+Previously Astro would bundle `<script>`s from both the Tabs and Accordion onto this page, even though on the Accordion is used. Now Astro is a little smarter and can support this re-export pattern that is common among libraries.
```
Yan-Thomas, line 26: A small typo here!
```suggestion
Previously Astro would bundle `<script>`s from both the Tabs and Accordion onto this page, even though only the Accordion is used. Now Astro is a little smarter and can support this re-export pattern that is common among libraries.
```

Yan-Thomas (commented):
Thanks, this looks great. Just found a small typo in the changeset! 🙌

packages/astro/src/core/build/plugins/plugin-analyzer.ts
ottomated: This is to handle a slightly different case. If component `Title.astro` is imported and used by `TitleBar.astro`, `TitleBar.astro` is not going to re-export `Title`. Therefore, this line basically switches from tracking Title to TitleBar, so the page will include `Title`'s scripts.

ottomated:
Do we want to make this also work for styles before merging? It shouldn't be that hard

.changeset/real-drinks-melt.md
```diff
@@ -0,0 +1,26 @@
...
+
+Then let's say you have a page that uses just the Accordion like so:
+
+```astro
+---
+import { Accordion } from '@matthewp/my-astro-lib';
+---
+```
+
+Previously Astro would bundle `<script>`s from both the Tabs and Accordion onto this page, even though only the Accordion is used. Now Astro is a little smarter and can support this re-export pattern that is common among libraries.
```
sarah11918, line 26:
```suggestion
Improved hoisted script bundling

Astro's static analysis to determine which `<script>` tags to bundle together just got a little smarter!

Astro create bundles that optimize script usage between pages and place them in the head of the document so that they are downloaded as early as possible. One limitation to Astro's exiting approach has been that you could not dynamically use hoisted scripts. Each page received the same, all-inclusive bundle whether or not every script was needed on that page.

Now, Astro has improved the static analysis to take into account the actual imports used. 

For example, Astro would previously bundle the `<scripts>` from both the `<Tab>` and `<Accordian>`  component for the following library that re-exports multiple components:

__@matthewp/my-astro-lib__

```js
export { default as Tabs } from './Tabs.astro';
export { default as Accordion } from './Accordion.astro';
```

Now, when an Astro page only uses a single component, Astro will send only the necessary script to the page. A page that only imports the `<Accordian>` component will not receive any `<Tab>` component's scripts:

```astro
---
import { Accordion } from '@matthewp/my-astro-lib';
---
```

You should now see more efficient performance with Astro now supporting this common library re-export pattern.
```
OK, took a shot at something that both emphasized the change a little earlier (since this is a changeset) and I think works as a narrative for a blog post!

Also, maybe I'm the only one who is picky about this, but telling a story between two code snippets is always tricky when you want to discuss them in tandem, (visually, grammatically...) so I tried to rework to make them each a little more standalone. This helps to make clear what each snippet's meaning or part of the story is, rather than, "Say you had all these pieces...." and then trying to draw a conclusion.

What you had wasn't bad! This is totally for your consideration! But if I had my choice, I'd structure a little something like this.

**Review by sarah11918 — COMMENTED** [review]
Hi @matthewp ! 

Now that Yan has made the minor fixes, this would be fine as-is **after making `<script>` plural in line 7** (either with an "s" or "script TAGS")... I just think it could be better!  ;)

See my updated take for your consideration and the reasons I'd be inclined to restructure. But totally your call! Take what you like from it, or take nothing at all!

File: .changeset/real-drinks-melt.md [hunk 7]
```diff
@@ -0,0 +1,30 @@
...
+
+Improved hoisted script bundling
+
+Astro's static analysis to determine which `<script>` tags to bundle together just got a little smarter!
+
+Astro create bundles that optimize script usage between pages and place them in the head of the document so that they are downloaded as early as possible. One limitation to Astro's existing approach has been that you could not dynamically use hoisted scripts. Each page received the same, all-inclusive bundle whether or not every script was needed on that page.
+
+Now, Astro has improved the static analysis to take into account the actual imports used. 
+
+For example, Astro would previously bundle the `<scripts>` from both the `<Tab>` and `<Accordian>`  component for the following library that re-exports multiple components:
```
sarah11918, line 13:
```suggestion
For example, Astro would previously bundle the `<script>`s from both the `<Tab>` and `<Accordian>`  component for the following library that re-exports multiple components:
```

.changeset/real-drinks-melt.md
```diff
@@ -0,0 +1,30 @@
+---
+'astro': minor
+---
+
+Improved hoisted script bundling
+
+Astro's static analysis to determine which `<script>` tags to bundle together just got a little smarter!
+
+Astro create bundles that optimize script usage between pages and place them in the head of the document so that they are downloaded as early as possible. One limitation to Astro's existing approach has been that you could not dynamically use hoisted scripts. Each page received the same, all-inclusive bundle whether or not every script was needed on that page.
```
ottomated, line 9: The existing approach actually did work in most cases (delivering different bundles per page) - this change only fixes the re-export case below.

sarah11918: @ottomated Can you suggest a rewording that is true? What would this paragraph have to look like so that it is accurate?

ottomated: reworded in #7733 !

Issue #7590, opened 2023-07-06
"res.writeHead is not a function" in Express/Node middleware.


vergonha:
### What version of `astro` are you using?

v2.7.4

### Are you using an SSR adapter? If so, which one?

Node Adapter, Middleware Mode

### What package manager are you using?

NPM

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

Custom middleware and error handling aren't catching Syntax Errors. 

As [official Express documentation](https://expressjs.com/guide/error-handling.html) describes, these are the middlewares to handling errors:

```js
function clientErrorHandler(err, req, res, next) {
  if (req.xhr) {
    res.status(500).send({ error: 'Something failed!' });
  } else {
    next(err);
  }
}

function errorHandler(err, req, res, next) {
  if (res.headersSent) {
    return next(err);
  }
  res.status(500);
  res.render('error', { error: err });
}

app.use(clientErrorHandler);
app.use(errorHandler);
```

However, there is a built-in error in the "dist/server/entry.mjs" file built by astro.

```
TypeError: res.writeHead is not a function
    at eval (file:///home/projects/middleware-astro-error/dist/server/entry.mjs:2639:13)
    at Layer.handle_error (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/layer.js:71:5)
    at trim_prefix (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:326:13)
    at eval (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:286:9)
    at Function.process_params (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:346:12)
    at next (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:280:10)
    at Layer.handle_error (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/layer.js:67:12)
    at trim_prefix (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:326:13)
    at eval (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:286:9)
    at Function.process_params (file:///home/projects/middleware-astro-error/node_modules/express/lib/router/index.js:346:12)
```

### What's the expected result?

Error will be handled by express and will be returned as a response in the API endpoint.

### Link to Minimal Reproducible Example

Run with `❯ npm run build && node server.mjs`

[...]

PR #7708, opened 2023-07-18
Fix "res.writeHead is not a function" in Express/node middleware

DixCouleur:
## Changes
Fix https://github.com/withastro/astro/issues/7590

add "error" argument to nodeMiddleware handler function

## Testing

https://github.com/withastro/astro/issues/7590 works correctly

## Docs

bug fix


natemoo-re:
Thanks so much for the PR, we really appreciate it! This change looks good but there seems to be a type error. Once that's fixed, this should be good to go!

DixCouleur:
> Thanks so much for the PR, we really appreciate it! This change looks good but there seems to be a type error. Once that's fixed, this should be good to go!

i forgot to build before submit, i will fix it

natemoo-re:
> My idea is that I need to infer the type of parameters received by the handler in createExports in runtime 🤔

That seems to be overcomplicating things? Couldn't the handler accept any number of arguments (`...args`) and act differently depending on the shape of `args`?

DixCouleur:
> > My idea is that I need to infer the type of parameters received by the handler in createExports in runtime 🤔
> 
> That seems to be overcomplicating things? Couldn't the handler accept any number of arguments (`...args`) and act differently depending on the shape of `args`?

Like `args[0] instanceof IncomingMessage` ?

ematipico:
> > > My idea is that I need to infer the type of parameters received by the handler in createExports in runtime 🤔
> > 
> > 
> > That seems to be overcomplicating things? Couldn't the handler accept any number of arguments (`...args`) and act differently depending on the shape of `args`?
> 
> Like `args[0] instanceof IncomingMessage` ?

Yeah that's the idea

DixCouleur:
> > > > My idea is that I need to infer the type of parameters received by the handler in createExports in runtime 🤔
> > > 
> > > 
> > > That seems to be overcomplicating things? Couldn't the handler accept any number of arguments (`...args`) and act differently depending on the shape of `args`?
> > 
> > 
> > Like `args[0] instanceof IncomingMessage` ?
> 
> Yeah that's the idea

I committed it with this idea

natemoo-re (approved):
Thanks for making that change, this looks great!

packages/integrations/node/src/nodeMiddleware.ts
```diff
@@ -1,17 +1,27 @@
 import type { NodeApp } from 'astro/app/node';
-import type { IncomingMessage, ServerResponse } from 'node:http';
+import type { ServerResponse } from 'node:http';
 import type { Readable } from 'stream';
 import { createOutgoingHttpHeaders } from './createOutgoingHttpHeaders';
 import { responseIterator } from './response-iterator';
-import type { Options } from './types';
+import type { ErrorHandlerParams, Options, RequestHandlerParams } from './types';
 
 export default function (app: NodeApp, mode: Options['mode']) {
```
natemoo-re, line 8:
```suggestion
// Disable no-unused-vars to avoid breaking signature change
// eslint-disable-next-line @typescript-eslint/no-unused-vars
export default function (app: NodeApp, _mode: Options['mode']) {
```

Issue #7717, opened 2023-07-19
Cannot deploy Vercel serverless function when using Node 20


Princesseuh:
### What version of `astro` are you using?

2.9.0

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

macOS

### What browser are you using?

N/A

### Describe the Bug

[The function here](https://github.com/withastro/astro/blob/4eba967d2b7e0593ed49f191b70dfea7a43e9829/packages/integrations/vercel/src/serverless/adapter.ts#L193) does [not take into account the versions of Node that Vercel actually supports](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version)

### What's the expected result?

Perhaps show a warning if the current version of Node is unsupported, or fallback to the nearest version

### Link to Minimal Reproducible Example

N/A

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Closed by #7718

PR #7718, opened 2023-07-19
fix(vercel): Switch to node 18 when local version is not supported

lilnasy:
## Changes

- #7717 
- Switch to Node.js 18 runtime when the version being used locally isn't supported by Vercel Serverless Functions.
- Warn when the local version being used is deprecated by Vercel.

#### What users will see when using an unsupported version
image

#### What users will see when using a deprecated version
image

## Docs
- Should not be required as this does not affect usage.


natemoo-re (approved):
This looks great, thank you!

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -19,6 +19,12 @@ const PACKAGE_NAME = '@astrojs/vercel/serverless';
 export const ASTRO_LOCALS_HEADER = 'x-astro-locals';
 export const VERCEL_EDGE_MIDDLEWARE_FILE = 'vercel-edge-middleware';
 
+const SUPPORTED_NODE_VERSIONS: Record<string, { status: 'current' } | { status: 'deprecated', removal: string }> = {
```
bluwy, line 22: Would be nice to add a comment and link to https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version so we can keep this up-to-date in the future.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -19,6 +19,12 @@ const PACKAGE_NAME = '@astrojs/vercel/serverless';
...
+	14: { status: 'deprecated', removal: 'August 15th 2023'  },
+	16: { status: 'deprecated', removal: 'February 6th 2024' },
```
bluwy, line 24: Ah the US date format 😅

bluwy (approved):

packages/integrations/vercel/src/serverless/adapter.ts
lilnasy: Yes, that's a great idea. Meant to do that actually.

packages/integrations/vercel/src/serverless/adapter.ts
lilnasy: Using Intl.DateTimeFormat now 🫡

bluwy: Meant that mostly as a joke, but now it's even better! Thanks 🙌

bluwy (approved):

ematipico (approved):

PR #7722, opened 2023-07-19
Ensure sitemap excludes nested error pages

natemoo-re:
## Changes

- There was a regression introduced in https://github.com/withastro/astro/pull/7655
- This PR fixes the regression, so that nested `[...slug]/404` and `[...slug]/500` pages are excluded as well
- Changeset added

## Testing

Test updated to include nested 404 page

## Docs

N/A, bug fix only

### DIFF
--- a/.changeset/fuzzy-toes-float.md
+++ b/.changeset/fuzzy-toes-float.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/sitemap': patch
+---
+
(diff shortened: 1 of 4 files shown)


delucis (approved):
Code makes sense to me, but I wasn’t able to test locally with Starlight’s docs where I first noticed the regression.

Thanks for the quick fix Nate! 💜

PR #7723, opened 2023-07-19
Fix ts error on `main`

natemoo-re:
## Changes

- Missed a type import
- Fixes CI

## Testing

Builds locally

## Docs

N/A


delucis (approved):
Not sure I’m qualified but approving this just because I’m impressed by Nate.

natemoo-re:
> Not sure I’m qualified but approving this just because I’m impressed by Nate.

We'll allow it 😅

PR #7730, opened 2023-07-20
Refactor Astro page rendering

bluwy:
## Changes

This PR creates a new `RenderDestination` abstraction for SSR render APIs to `write` to. This allows us to move away from generators and write to the string/stream directly.

Gist of change:
- Formalize `renderToString` and `renderToReadableStream` for Astro components
- Update `renderPage` to use those two APIs

Right now, everything is still using generators. In a later PR, I'll move the APIs to use the `RenderDestination` instead.

## Testing

Existing rendering tests should pass.

## Docs

n/a. internal refactor.


packages/astro/src/runtime/server/jsx.ts
```diff
@@ -91,7 +91,11 @@ Did you forget to import the component or is it possible there is a typo?`);
 						props[key] = value;
 					}
 				}
-				const html = markHTMLString(await renderToString(result, vnode.type as any, props, slots));
+				const str = await renderToString(result, vnode.type as any, props, slots);
+				if (str instanceof Response) {
+					throw str;
+				}
```
bluwy, line 97: This is not ideal, but I'm following the API we had before:

https://github.com/withastro/astro/blob/d8bae784b4a3b555c213febe6334373d89c7e089/packages/astro/src/runtime/server/render/astro/factory.ts#L32-L35

packages/astro/src/runtime/server/render/astro/factory.ts
```diff
@@ -20,29 +16,6 @@ export function isAstroComponentFactory(obj: any): obj is AstroComponentFactory
 	return obj == null ? false : obj.isAstroComponentFactory === true;
 }
 
-// Calls a component and renders it into a string of HTML
-export async function renderToString(
```
bluwy, line 24: This old `renderToString` API is moved to `render.ts` and along side the `renderToReadableStream` API.

ematipico (approved):

PR #7734, opened 2023-07-20
test: add test in the Node adapter for astro:assets

Princesseuh:
## Changes

What the title says. This PR also includes a very small fix where `globalThis.astroAsset.addStaticImage` could leak out of the program. This didn't affect normal usage, but was surfaced in the Node test.

## Testing

It's tests!

## Docs

N/A


packages/astro/src/core/build/generate.ts
```diff
@@ -130,7 +130,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		? opts.settings.config.build.server
 		: getOutDirWithinCwd(opts.settings.config.outDir);
 
-	if (ssr && !hasPrerenderedPages(internals)) return;
+	// HACK! `astro:assets` relies on a global to know if its running in dev, prod, ssr, ssg, full moon
+	// If we don't delete it here, it's technically not impossible (albeit improbable) for it to leak
+	if (ssr && !hasPrerenderedPages(internals)) {
+		delete globalThis?.astroAsset?.addStaticImage;
```
ematipico, line 136: nit: it's usually better to use `something.bar = undefined` instead of `delete`, [performance-wise](https://docs.rome.tools/lint/rules/nodelete/#nodelete-since-v070).

ematipico (approved):

packages/astro/src/core/build/generate.ts
Princesseuh: oh weird, didn't know that. This is in the build part so it (kinda) matters, will update

Princesseuh: Aah, but then I have to check if the property exists, that's annoying and probably not worth it then

Princesseuh: ah, but then I'd need to check that `astroAsset` exists or it's unsafe. Oh well, the global is supposed to be temporary so I guess it's fine

PR #7746, opened 2023-07-21
Bump vite to 4.4.6

birkskyum:
## Changes

This bumps Vite to 4.4.6 (Astro is currently at 4.3.9). It includes patch bumps to all the adapters.

This brings experimental support for [Lightning CSS](https://vitejs.dev/guide/features.html#lightning-css) (vite 4.4.0) and improved support for [Bun](https://github.com/vitejs/vite/pull/13901) (vite 4.4.6).

Two ts-expect-error flags are obsolete and removed, because [Vite now use esbuild 0.18](https://github.com/vitejs/vite/blob/v4.4.0/packages/vite/CHANGELOG.md#esbuild-018-update)

## Testing

Existing unit tests and e2e tests should still pass.

## Docs

There is currently no section on [Lightning CSS](https://lightningcss.dev/).


.changeset/sweet-lamps-cover.md
```diff
@@ -0,0 +1,12 @@
+---
+'@astrojs/tailwind': minor
+'@astrojs/markdoc': minor
+'@astrojs/netlify': minor
+'@astrojs/svelte': minor
+'@astrojs/image': minor
+'@astrojs/mdx': minor
+'@astrojs/vue': minor
+'astro': minor
```
ematipico, line 9: These should all be patches because we are not adding new features.

ematipico (changes requested):

Princesseuh:
@bluwy was there any reasons we weren't already on 4.4?

.changeset/sweet-lamps-cover.md
```diff
@@ -0,0 +1,12 @@
+---
+'@astrojs/tailwind': patch
+'@astrojs/markdoc': patch
+'@astrojs/netlify': patch
+'@astrojs/svelte': patch
+'@astrojs/image': patch
+'@astrojs/mdx': patch
+'@astrojs/vue': patch
+'astro': patch
```
bluwy, line 9: I don't think we need all of these. Most of the integrations use Vite as a dev dep only. I think we can keep `'astro': patch` itself.

bluwy:
> @bluwy was there any reasons we weren't already on 4.4?

Mostly that we're not using any 4.4 specific features, and users can bump to 4.4 if they want to themselves. But also, 4.4 started with quite some regressions that I initially hold off, but it's safer now.

The e2e failing seems to be legit though. Might need to address that before merging, could be a Vite regression.

birkskyum:
I tried updating solid-js 1.7.6 -> 1.7.8 (latest), but it doesn't resolve "Circular imports with Solid › Context" test breaks in E2E.

birkskyum:
In the existing main branch (Vite 4.3.9), there are already these similar Preact E2E cyclic errors, but they are being muted:

```sh
  ✓  20 [Chrome Stable] > e2e/custom-client-directives.test.js:74:2 › Custom Client Directives - build server › client:password should work (150ms)
 error   Cyclic reference detected while serializing props for <PreactCounter client:load>!
  
  Cyclic references cannot be safely serialized for client-side usage. Please remove the cyclic reference.
Error: Cyclic reference detected while serializing props for <PreactCounter client:load>!

Cyclic references cannot be safely serialized for client-side usage. Please remove the cyclic reference.

  ✓  21 [Chrome Stable] › e2e/error-cyclic.test.js:20:2 › Error: Cyclic Reference › overlay (671ms)
 error   [sass] Undefined variable.
    ╷
  5 │     color: $red; // undefined!
    │            ^^^^
    ╵
    e2e/fixtures/error-sass/src/pages/index.astro 5:12  root stylesheet
  File:
    /Users/admin/repos/astro/packages/astro/e2e/fixtures/error-sass/src/pages/index.astro:5:12
  Code:
    Error: Undefined variable.
      ╷
    5 │     color: $red; // undefined!
      │            ^^^^
      ╵
      e2e/fixtures/error-sass/src/pages/index.astro 5:12  root stylesheet
  Stacktrace:
Error: Undefined variable.
  ╷
5 │     color: $red; // undefined!
  │            ^^^^
```

birkskyum:
@bluwy , to debug this further, how can I as an Astro user use a custom version of Vite like you suggest?

bluwy:
You can run `npm update vite` in your project to use a different Vite version for Astro.

I can try to take a look at the issue over the weekend. A bit swamped at the moment 😅

birkskyum:
@bluwy , I've found that removing the `client:only` in the `<ContextProvider client:only />` makes the test pass.

birkskyum:
Turns our being more strict about using `import type` when only importing types such as the solid-js Component was all it took.

.changeset/sweet-lamps-cover.md
birkskyum: This has been updated, and all tests are now green.

bluwy (approved):
Thanks for fixing the tests! This make sense to me as Vite's esbuild now understand our tsconfig.json's `verbatimModuleSyntax: true` config which causes the type elision to no work correctly. Not really sure if this is a Vite regression, but I'll keep watch of this.

birkskyum:
Rebased. Tests will pass if rerun (they were just cancelled).

.changeset/sweet-lamps-cover.md
birkskyum: @ematipico , is there anything else that should be changed here?

Issue #7739, opened 2023-07-20
Enabling view transitions breaks back functionality


MikeCarbone:
### What version of `astro` are you using?

2.9.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome, Firefox

### Describe the Bug

Enabled the new view transitions and the back arrow no longer functions. The URL changes, but the view does not. Tested in latest chrome and firefox (view transition doesn't work in firefox, but back functionality still broken).

Details:
On my homepage `/`, I followed the steps [outlined here](https://docs.astro.build/en/guides/view-transitions/#built-in-animation-directives). On another page, I did not add any of the view configuration setup. One with the transition, one without. When I navigate from the page with the transitions enabled to the non-enabled page, the transition will fire and take me to the next page. When I try to use the browser back arrow, either by clicking it or swiping, the URL will change but the view will not.

### What's the expected result?

The back and forward arrows should function as expected.

The linked stackblitz isn't rendering anything and I don't have the patience to debug. But it's a simple implementation.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-b79kq3?file=src%2Fpages%2Fblog%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Thanks. So if I got this right, it's

A - VT
B - No VT

A -> B (back) -> A that causes the bug?

PR #7750, opened 2023-07-21
Handle back navigation from page without VT

matthewp:
## Changes

- When navigation from a page without VT enabled we want a full page refresh.
- Change is to do `location.reload()` so that a refresh is fired.
- Fixes https://github.com/withastro/astro/issues/7739

## Testing

- New E2E case added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -128,7 +128,12 @@ const { fallback = 'animate' } = Astro.props as Props;
 			}
 		});
 		window.addEventListener('popstate', () => {
-			if (!transitionEnabledOnThisPage()) return;
+			if (!transitionEnabledOnThisPage()) {
+				// The current page doesn't haven't View Transitions,
+				// respect that with a full page reload
+				location.reload();
+				return;
+			}
 			const nextIndex = history.state?.index ?? currentHistoryIndex + 1;
 			const direction: Direction = nextIndex > currentHistoryIndex ? 'forward' : 'back';
 			navigate(direction, location.href);
--- a/.changeset/sixty-peas-join.md
+++ b/.changeset/sixty-peas-join.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Trigger full page refresh on back nav from page without VT enabled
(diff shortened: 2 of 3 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -128,7 +128,12 @@ const { fallback = 'animate' } = Astro.props as Props;
 			}
 		});
 		window.addEventListener('popstate', () => {
-			if (!transitionEnabledOnThisPage()) return;
+			if (!transitionEnabledOnThisPage()) {
+				// The current page doesn't haven't View Transitions,
+				// respect that with a full page reload
+				location.reload();
```
matthewp, line 134: I was unsure if this was the best way but from research it seems like it probably is? popstate is a weird API.

ematipico: I suppose that's fine

ematipico (approved):

W1M0R:
I am using htmx in an Astro component. Htmx does history management and can restore state when pressing the back button. However, the Astro popstate implementation reloads the page before Htmx gets an opportunity to do its work. I'm still investigating a solution, but I think that I might either need to trick Astro into not doing a reload (maybe just enable ViewTransitions), or somehow unregister the Astro popstate handler to prevent interference.

@matthewp @ematipico What do you think would be a good approach to deal with this situation?

Here is more information:

> The hx-push-url attribute allows you to push a URL into the browser [location history](https://developer.mozilla.org/en-US/docs/Web/API/History_API). This creates a new history entry, allowing navigation with the browser’s back and forward buttons. htmx snapshots the current DOM and saves it into its history cache, and restores from this cache on navigation.

1. https://htmx.org/attributes/hx-push-url/
2. https://htmx.org/attributes/hx-boost/

W1M0R:
Here is a debugging screenshot that might help:

image


Notice that Unpoly, HTMX and Astro are responding to the popstate event (see the Global Listeners tab).

Also notice how Unpoly first checks to see if the state was saved by it or someone else. Maybe Astro could also do something similar?

W1M0R:
HTMX also does a similar check:

image

W1M0R:
Here is where the breakpoint hits:

image

Notice that `event.state.htmx` is set to true.

W1M0R:
One "workaround" might be to set the `astro-view-transitions-fallback` to `none`:

image

image

W1M0R:
It turns out that I had `<ViewTransitions />` in my `<head>`. Removing it resolved the issue for me.

W1M0R:
For anyone looking to improve the onpopstate handling, here is a clip on the handler: https://youtu.be/javGxN-h9VQ?si=cYxRkzQnlO7Y9DGr&t=1080

Issue #7735, opened 2023-07-20
@astrojs/vercel with edge function target doesn't render 404.astro page


susickypavel:
### What version of `astro` are you using?

2.9.0

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac, Windows

### What browser are you using?

Chrome, Firefox, Safari

### Describe the Bug

Deployed app: https://astro-vercel-404-edge-bug.vercel.app/
Code: https://stackblitz.com/edit/github-z7x12r?file=src%2Fpages%2F404.astro

Astro website deployed on Vercel using edge adapter doesn't recognize 404.astro page and browser native 404 page is shown instead.

It works in development, it doesn't work on Vercel only. It doesn't matter whether the 404.astro is prerendered or not.

### What's the expected result?

Navigating to a page that doesn't exist shows the content of 404.astro.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-z7x12r?file=src%2Fpages%2F404.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

alanescarcha:
Bump!
It seems that in 2.8.0 it worked correctly.

sahilsuman933:
@susickypavel, I believe this issue has been resolved in one of the latest PRs. If you visit https://gullible-galaxy.vercel.app/, you can see that the 404 page is rendering correctly.

susickypavel:
> @susickypavel, I believe this issue has been resolved in one of the latest PRs. If you visit https://gullible-galaxy.vercel.app/, you can see that the 404 page is rendering correctly.

Can you provide stackblitz project? I'm running the latest version of Vercel adapter and Astro and still seeing this issue.

sahilsuman933:
I am also experiencing the same issue in the latest version of Astro. However, I have not encountered any problems when linking to the Astro repository using "pnpm link astro" with all the latest pull requests merged. Therefore, please wait for the latest release, as it will address the issue. @susickypavel 

// @Princesseuh  Just to confirm again, can you test that from your end?

matthewp:
We have deprecated the Vercel Edge and Netlify Edge adapters due to compatibility issues and are encouraging people to instead use the serverless versions of these adapters, along with Edge Middleware, which we feel gives you the best of both worlds. Our adapter migration plan is laid out in this [roadmap issue](https://github.com/withastro/roadmap/discussions/631).

[...]

Issue #7616, opened 2023-07-10
Unable to find custom 404 page with trailing slash


516310460:
### What version of `astro` are you using?

v2.8.0

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

When output: "hybrid", accessing a non-existent page, the custom 404 page cannot be found.
Automatically add '/' after the URL
Step:
1.npm run build
2.HOST=0.0.0.0 PORT=3000 node dist/server/entry.mjs
3.localhost:3000/en/test/asd
4.Custom 404 page not found

### What's the expected result?

1. Find the custom 404 page
2. Remove the trailing slash

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-nhc9gb-malcwc?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

516310460:
I have a question, I don't know how to implement it.
[lang]/page/[id].astro
Expect:
1. url: /en/page/10
2. url: /page/10
I don't know how to achieve it

About another question
Use `output: "hybrid"`
The route will prioritize matching SSG before matching SSR.
Can I set it to match SSR first and then match SSG, because in actual website visits, I need a dynamic route. If the SSR application crashes, I can use SSG

natemoo-re:
I am looking into this and can see that we aren't handling 404 pages properly in hybrid mode. Working on a fix.

- Trailing slash behavior can be configured with the [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option
- Please open new issues for other bugs and visit [our Discord](https://astro.build/chat) for support questions.

516310460:
> I am looking into this and can see that we aren't handling 404 pages properly in hybrid mode. Working on a fix.
> 
> * Trailing slash behavior can be configured with the [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option
> * Please open new issues for other bugs and visit [our Discord](https://astro.build/chat) for support questions.

`TrailingSlash 'can only be used in development. In output: "hybrid" and SSG mode, it will automatically add at the end`/`

lilnasy:
Fixed in #8070 (astro v2.10.10+)
<img width="423" alt="image" src="https://github.com/withastro/astro/assets/69170106/e1c5c841-9d9d-4908-80a5-1a1f748a75de">

PR #7754, opened 2023-07-21
Refactor 404 and 500 approach

natemoo-re:
## Changes

- Refactors our base `app` to handle `404` and `500` routes more consistently
- Should now support pre-rendered `404` and `500` pages by making a new request to the origin server
- Treating as a patch, but can release as minor if we're concerned about the approach?
- Related to https://github.com/withastro/astro/issues/7735, https://github.com/withastro/astro/issues/7616

## Testing

Tests updated!

## Docs

N/A


matthewp (approved):

ematipico (approved):
I like this approach a lot. Great refactor!

packages/astro/src/core/app/index.ts
```diff
@@ -307,6 +262,62 @@ export class App {
...
+
+	#mergeResponses(newResponse: Response, oldResponse?: Response) {
+		if (!oldResponse) return newResponse;
+		const { status, statusText, headers } = oldResponse;
+		return new Response(newResponse.body, {
+			status,
+			statusText,
+			headers: new Headers(Array.from(headers))
+		})
+	}
```
natemoo-re, line 312: Not sure if this is totally needed? We had a test that confirmed that `status` and `statusText` from the original `Response` were preserved.

Trying to think through edge cases here—a bit unclear if headers from the `newResponse` should be preserved/have precedence?

packages/integrations/node/test/fixtures/prerender-404/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@test/nodejs-prerender-404",
```
natemoo-re, line 2: Moved this from the `astro` package to the `@astrojs/node` package so we can test that the prerendered 404 page is properly handled in prod.

natemoo-re (commented):
A few updates! Moved some tests around, added a couple small fixes

packages/astro/src/core/app/index.ts
matthewp: I'm not sure either but preserving them feels like the right thing to do...

packages/astro/src/core/app/index.ts
```diff
@@ -307,6 +264,63 @@ export class App {
...
+		Reflect.set(response, responseSentSymbol, true);
+		return response;
+	}
+
+	#mergeResponses(newResponse: Response, oldResponse?: Response) {
+		if (!oldResponse) return newResponse;
+		const { status, statusText, headers } = oldResponse;
+
+		return new Response(newResponse.body, {
+			status: status === 200 ? newResponse.status : status,
```
ematipico, line 311: This is weird; why do we return `newRespnose.status` if we check the original `staus`? Can you add a comment?

ematipico:
@natemoo-re I left some feedback, could you please address it?

PR #7755, opened 2023-07-21
feat: Add a Check for `prefers-reduced-motion` to the View Transitions component.

jazzypants1989:
## Changes

Before, the animations still automatically ran even when the user has the `prefers-reduced-motion` setting set to `reduce`.

https://github.com/withastro/astro/assets/102560332/cb617192-87c3-48cf-b6c5-2c7760f2d474

Now, we do a quick check to see if this setting is enabled, and skip the animation if so.

https://github.com/withastro/astro/assets/102560332/1143d629-9394-4444-9fc4-f0a87c8e9ff2

I wrote this carefully so that the fallback animation follows suit. Here we have it in Firefox:

https://github.com/withastro/astro/assets/102560332/0c53717c-dbed-48c3-bbfa-e49559787ea0

I had to increase the setTimeout to 50ms to avoid a DOM exception. It still worked fine, but it was angry about it haha.

## Testing

I am admittedly an absolute idiot, but I could not get the codespace container to work or even get the tests to run locally on my machine. So, I didn't include a test because I have no idea if it would work. I spent three hours in vain. I really apologize. 

And, honestly, I don't even know how to test for this. With the fallback we could just check for an animationstart event or something, but is there a special view transition event that we can check for (the absence of)? I checked the spec and I didn't see anything that you could observe on the window... maybe a special data attribute or something? I have no clue. The ::view-transition element will be at the root, so maybe we could look for that somehow.

## Docs

Perhaps we could just add an extra note that we take care of this automatically for the users? Perhaps a little extra info box that says something like,

"Don't worry! If you're worried about how all these animations will affect users that prefer reduced motion, we automatically check for this setting and do a more subtle swap of the DOM elements to better accommodate these users."


matthewp:
Hey @jazzypants1989, does this need to be done in JS? I did another PR (sorry, I didn't see yours first), that just did the minimal thing in CSS: https://github.com/withastro/astro/pull/7757

jazzypants1989:
Yeah, per my comment there, these two pull requests actually work very well together because mine would cover the SPA version and yours would cover the MPA version. 💖

sarah11918:
Just adding a comment here that I would absolutely want some kind of note added to the Docs page. (Whichever of these two PRs goes through!)

Assuming that either solution is an automated, behind-the-scenes check (nothing to set or configure in your Astro project), then a Docs PR can be made either way, without Docs caring which implementation is used.

@jazzypants1989 Would you like to PR the View Transitions page in docs? I would expect a note to indicate:
- that Astro's view transitions respects a user's `prefers-reduced-motion` preference (set in the browser?)
- ANY situation in which View Transitions would NOT respect this setting (e.g. unsupported browser, if it's a Tuesday... whatever.)

jazzypants1989:
@sarah11918 Will do! I'll have that done in the next hour or so. Meanwhile, we can close this PR because @matthewp found an elegant CSS solution instead.

PR #7759, opened 2023-07-21
Misc improvements for `astro:assets`

Princesseuh:
## Changes

- The types for `SharedImageService` were wrong and assumed every hook to be sync only, but we do support all of them being optionally async
- This led me to discover some spots where we in fact, did not support them being async
- Exported all the utils from `assets/utils`, they're undocumented, not an official thing, can break and change etc, but some users like to use them (especially imageMetadata, which we'll be able to make an official export once `image-size` is 2.0 if someone needs it)
- I realized that I completely misunderstood a part of `@astrojs/image`, and that we could completely remove the endpoint used only for dev. Now, the same endpoint is used in SSR and dev, making it much easier for users to try their custom endpoints in dev

## Testing

Tests should still pass!

## Docs

N/A

### DIFF
--- a/packages/astro/test/units/dev/restart.test.js
+++ b/packages/astro/test/units/dev/restart.test.js
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import * as cheerio from 'cheerio';
 import { fileURLToPath } from 'node:url';
 
+import { createSettings, openConfig } from '../../../dist/core/config/index.js';
 import {
 	createContainerWithAutomaticRestart,
 	isStarted,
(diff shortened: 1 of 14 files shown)


matthewp:
needs changeset, otherwise looks good

packages/astro/test/units/dev/restart.test.js
```diff
@@ -133,7 +133,7 @@ describe('dev container restarts', () => {
 			cmd: 'dev',
 			logging: defaultLogging,
 		});
-		const settings = createSettings(astroConfig);
+		const settings = createSettings(astroConfig, 'dev');
```
matthewp, line 136: isn't the second arg the cwd?

Princesseuh: Not anymore: https://github.com/withastro/astro/blob/b46dfdb12fae70fb6f126b6105968d371073ea63/packages/astro/src/core/config/settings.ts#L113

matthewp (approved):

PR #7760, opened 2023-07-21
Basics theme styling updates

doodlemarks:
Ugh, who let the designer have git access.

I wanted to do a light styling update to bring Basics more inline with our updated branding while not piling on too much crud. 

Before:
image

After:
image


delucis (commented):
Love it @doodlemarks! Looks super slick and a nice update to this template that looked a bit out of date.

One note: I _think_ the “Astro” with the gradient probably has too low a contrast — a11y tooling doesn’t measure clipped gradients like this easily, but definitely the `--accent-dark` colour here is too low.

A dumb quick fix I spotted was just bumping the `background-position` in the `.text-gradient` style up to 25% so that less of the dark accent part of the gradient is used:

| Before | After |
|---|---|
| image | The text “Welcome to Astro” with a higher contrast gradient applied to the word “Astro” |

Or we could of course do something totally different.

Other than that, 10/10, no notes!

doodlemarks:
great callout @delucis! once again, busted by the a11y police

doodlemarks:
updated: image

examples/basics/src/layouts/Layout.astro
```diff
@@ -22,12 +22,15 @@ const { title } = Astro.props;
...
+		--accent: 136, 58, 234;
+		--accent-light: 224, 204, 250;
+		--accent-dark: 49, 10, 101;
+		--accent-gradient: linear-gradient(45deg, rgb(var(--accent)), rgb(var(--accent-light)) 30%, white 60%);
 	}
 	html {
 		font-family: system-ui, sans-serif;
-		background-color: #F6F6F6;
+		background: #13151A url('astro_a.svg') center -32px no-repeat;
+		background-size: 224px;
```
natemoo-re, line 33: Hmm not sure if I love this! Maybe [inlining the SVG](https://yoksel.github.io/url-encoder/) to avoid having another file users have to delete?

natemoo-re (commented):
Nice, thanks for doing this Mark!

examples/basics/src/layouts/Layout.astro
doodlemarks: updated!

PR #7768, opened 2023-07-23
Blog theme styling updates

doodlemarks:
## Changes

Another goblin weekend push. Wanting to still keep this styling "light" but seeing how often this is minimally styled beyond what we present it'd be nice to have a bit of a more modern base. 

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |

## Testing
Tested on chrome, ff, safari, and mobile. 

## Docs
n/a


examples/blog/src/styles/global.css
```diff
@@ -3,33 +3,84 @@
   https://github.com/HermanMartinus/bearblog/blob/297026a877bc2ab2b3bdfbd6b9f7961c350917dd/templates/styles/blog/default.css
```
tony-sull, line 3:
```suggestion
 Based on https://github.com/HermanMartinus/bearblog/blob/297026a877bc2ab2b3bdfbd6b9f7961c350917dd/templates/styles/blog/default.css
```

examples/blog/src/layouts/BlogPost.astro
```diff
@@ -30,22 +56,28 @@ const { title, description, pubDate, updatedDate, heroImage } = Astro.props;
 
 	<body>
 		<Header />
-		<main>
+		<main class="blog-post">
```
tony-sull, line 59: Super tiny nit that's more personal preference, feel free to ignore!

I always try to lean on Astro's scoped styles when possible.  Here the style should be able to target `main { }` directly, avoiding the extra `.blog-post` class all together

doodlemarks: tbh not sure why I did that ha, goblin mode things

natemoo-re (commented):
LGTM! I would love @Princesseuh to weigh in on the images!

Princesseuh:
> LGTM! I would love @Princesseuh to weigh in on the images!

We'll probably be updating some of the templates to use `astro:assets` for 3.0, but we can tackle this outside this PR for sure, so LGTM!

PR #7769, opened 2023-07-23
feat: implement partial build

y-nk:
## Changes

1. added `astroConfig.build.filterPage` option
2. used this option at `collectPagesData` stage to allow skipping pages without dynamic params in it
3. used this option at `getPathsForRoute` stage to allow skipping pages with dynamic params in it

## Goal

The feature gives a way for users to skip the generation of some pages from their astro config. It's a quick win to allow static generation to do _partial_ generation easily.

The feature proposes to filter all pages (_before_ **and** _after_ expansion from `getStaticPaths`). It has to be done so that a user could filter and master 100% the content which is statically built.

**This feature is intended for pages enrolled for prerendering only, since we aim to target only the pages generated at build time.** (not for SSR)

## Common usecases

1. It is possible to prefix routes with `_` which would disable them but it is currently not possible to have routes available only in dev mode (which are excluded from build), but with this we can do:

```ts
{
  build: {
    filterPage: (pageURL) => pageURL.startsWith('/dev/')
  }
}
```

2. For people using Astro as an SSG/hybrid, we can use this feature for regenerating static content from a 3rd party CMS easily:

```ts
{
  build: {
    filterPage: (pageURL) => {
      const regexps = process.env.LIST_OF_EXPR_OF_UPDATED_PAGES_FROM_CMS
        .split(',')
        .map(expr => new RegExp(expr))
    
      return !!regexps.find(regexp => regexp.test(pageURL))
    }
  }
}
```

## Testing

I've tested this manually:

1. cloned astro on my machine
4. built astro from root
5. `cd packages/astro` && `yarn link`
6. in my test project: `yarn link astro`
7. implemented by change, compiled again astro
8. added `filterPage` function in my astro user config

I tried:

1. `filterPage: (pagePath) => false` → no pages built at all
2. `filterPage: (pagePath) => true` → default, no effect
3. `filterPage: (pagePath) => !pagePath.startsWith('/dev/')` → my dev pages weren't built

## Docs

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -883,6 +883,26 @@ export interface AstroUserConfig {
 		 * Please see your [SSR adapter's documentation](/en/guides/integrations-guide/#official-integrations) for using `edgeMiddleware` to define whether or not any SSR middleware code will be bundled when built.
 		 */
 		excludeMiddleware?: boolean;
+
+		/**
+		 * @docs
+		 * @name build.filterPage
+		 * @type {(pagePath: string) => boolean}
+		 * @default `undefined`
+		 * @version 2.10.0
+		 * @description
+		 * When defined, will be called for every single page to be generated (before generation) and gives opportunity to skip the generation of the page (based on user defined criterias)
+		 *
+		 * ```js
+		 * {
+		 *   build: {
+		 *     // will exclude all /dev/ from being generated
+		 *     filterPage: (pagePath) => !pagePath.startsWith('/dev/')
+		 *   }
+		 * }
+		 * ```
+		 */
+		filterPage?: (pagePath: string) => boolean;
 	};
 
 	/**
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -891,9 +891,10 @@ export interface AstroUserConfig {
 		 * @default `undefined`
 		 * @version 2.10.0
 		 * @description
-* Defines a function that will check which pages should be filtered out from the build while returning `false`.
-*
-* When defined, its function will be called for every single page about to be generated and gives the opportunity to skip the page. The criteria defined by the user based on `pageURL`, while returning `false`, will exclude any generated pages from the static build (including prerendered pages).
+		 *
+		 * Defines a function that will check which pages should be filtered out from the build while returning `false`.
+		 *
+		 * When defined, its function will be called for every single page about to be generated and gives the opportunity to skip the page. The criteria defined by the user based on `pageURL`, while returning `false`, will exclude any generated pages from the static build (including prerendered pages).
 		 *
 		 * ```js
 		 * {
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -887,22 +887,22 @@ export interface AstroUserConfig {
 		/**
 		 * @docs
 		 * @name build.filterPage
-		 * @type {(pagePath: string) => boolean}
+		 * @type {(pageURL: string) => boolean}
 		 * @default `undefined`
 		 * @version 2.10.0
 		 * @description
-		 * When defined, will be called for every single page to be generated (before generation) and gives opportunity to skip the generation of the page (based on user defined criterias)
+		 * When defined, will be called for every single page about to be generated and gives opportunity to skip the page. When returning `false`, the page matching the finale `pageURL` will be excluded from the static build.
 		 *
 		 * ```js
 		 * {
 		 *   build: {
-		 *     // will exclude all /dev/ from being generated
-		 *     filterPage: (pagePath) => !pagePath.startsWith('/dev/')
+		 *     // will exclude all /dev/* pages from being generated
+		 *     filterPage: (pageURL) => !pageURL.startsWith('/dev/')
 		 *   }
 		 * }
 		 * ```
 		 */
-		filterPage?: (pagePath: string) => boolean;
+		filterPage?: (pageURL: string) => boolean;
 	};
 
 	/**
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -891,7 +891,9 @@ export interface AstroUserConfig {
 		 * @default `undefined`
 		 * @version 2.10.0
 		 * @description
-		 * When defined, will be called for every single page about to be generated and gives opportunity to skip the page. When returning `false`, the page matching the finale `pageURL` will be excluded from the static build.
+* Defines a function that will check which pages should be filtered out from the build while returning `false`.
+*
+* When defined, its function will be called for every single page about to be generated and gives the opportunity to skip the page. The criteria defined by the user based on `pageURL`, while returning `false`, will exclude any generated pages from the static build (including prerendered pages).
 		 *
 		 * ```js
 		 * {
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -279,26 +279,39 @@ async function generatePage(
 		mod: pageModule,
 	};
 
-	const icon = pageData.route.type === 'page' ? green('▶') : magenta('λ');
-	if (isRelativePath(pageData.route.component)) {
-		logger.info(null, `${icon} ${pageData.route.route}`);
-	} else {
-		logger.info(null, `${icon} ${pageData.route.component}`);
-	}
-
 	// Get paths for the route, calling getStaticPaths if needed.
 	const paths = await getPathsForRoute(pageData, pageModule, pipeline, builtPaths);
 
+	// We do this here to get the stat `filteredPaths / paths` for info logging.
+	const filteredPaths = paths.filter((path) => pathFilter?.(path));
+
+	// Cleanup entries added from `getPathsForRoute`
+	const skippedPaths = paths.filter((path) => !filteredPaths.includes(path));
+	skippedPaths.forEach((path) => builtPaths.delete(removeTrailingForwardSlash(path)));
+
+	const routesSkipped =
+		filteredPaths.length === 0
+			? dim('[skipped fully]')
+			: filteredPaths.length < paths.length
+			? dim(`[skipped ${paths.length - filteredPaths.length} paths of ${paths.length}]`)
+			: '';
+
+	const icon = pageData.route.type === 'page' ? green('▶') : magenta('λ');
+	if (isRelativePath(pageData.route.component)) {
+		logger.info(null, `${icon} ${pageData.route.route} ${routesSkipped}`);
+	} else {
+		logger.info(null, `${icon} ${pageData.route.component} ${routesSkipped}`);
+	}
+
 	let prevTimeEnd = timeStart;
-	for (let i = 0; i < paths.length; i++) {
-		const path = paths[i];
-		if (pathFilter && !pathFilter(path)) continue;
+	for (let i = 0; i < filteredPaths.length; i++) {
+		const path = filteredPaths[i];
 		await generatePath(path, generationOptions, pipeline);
 		const timeEnd = performance.now();
 		const timeChange = getTimeStat(prevTimeEnd, timeEnd);
 		const timeIncrease = `(+${timeChange})`;
 		const filePath = getOutputFilename(pipeline.getConfig(), path, pageData.route.type);
-		const lineIcon = i === paths.length - 1 ? '└─' : '├─';
+		const lineIcon = i === filteredPaths.length - 1 ? '└─' : '├─';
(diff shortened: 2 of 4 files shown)


ematipico:
@y-nk Thank you for proposing this feature. A few things that I am missing; could you please address them?

1. The function seems to be applied only when we **generate** the page, although we build it anyway. Should we also **not build** it? It would make the whole build faster;
2. I believe the filtering should be applied [here](https://github.com/withastro/astro/blob/main/packages/astro/src/core/build/generate.ts#L141-L181). Doing so, we avoid **loading** the page to generate. What do you think?
3. What about SSR? Is this feature meant also for SSR?

y-nk:
@ematipico thanks for your time spent on reviewing this.

1. I'm honestly lost at the generate/build nuance. i'm not sure to have enough knowledge to understand this.

2. i first thought on putting the filtering where you pointed, but it does not expand the filtering to paths generated from `getStaticPaths` (with dynamic params in it). For this to happen it needs to be within `generatePage`. i'm thinking we could also add the "raw" path - containing the params - too and decide to skip import it if the function returned false (like a grouped filter, if you follow me)

3. the feature is definitely not meant for SSR, that's why i put the config in `build` rather than root level. it is intended for build only, and for filtering pre-renderable (prerender=true) route paths only (writing this i realize that api routes are not - yet - covered here)

The usecase addressed is to skip pages which should not be built, defined by a criteria in userland. this might not make sense when using content collection, but it does when using a 3rd party CMS (but one could still use the output of a git diff of `src/content` maybe?)

Applied to my current usecase, my project has `src/pages/[...path].astro` where all route paths are pulled from a strapi backend and generated at built. i want to wire strapi's webhook to my build github action so that only updated pages get rebuilt and deployed.

lilnasy:
@y-nk  Regarding your first point,  the two processes can be thought of as compiling, and rendering (to HTML). From looking at the PR, I can tell it prevents some pages from being rendered, but we still pay the cost of `.astro` to `.ts` compilation, css and js compilation and optimizations for the entire project, etc. In effect, this PR is like SSR + prerendering but only the prerendered pages are kept and no server runtime is created.

Regarding the third point, the `build` section of the astro config has implications on both server and static _builds_. A more appropriate place might be output (`"output": "static-partial"`). It sounds drastic but that's the category of change this PR implies.

I would look into how this change could be made into an integration. And if that's not possible, what could we change in the Integrations API to make it possible.

y-nk:
@lilnasy i've tried to make this an integration prior making the change here, and followed @natemoo-re advice (given in discord) that making a PR here would be a better option given what i wanted to achieve.

for the cost of ts i agree, but there's sadly no way to make it not happening since there's no function encapsulation in astro components ; at this point it's a design constraint in astro files since there's just no way for anyone to run `getStaticPaths` and skip one of the pages without all the steps you mentioned.

for the part of cost of css etc, I didn't check on this - it was my expectation that skipping a page in `generatePage` that early would result in the rest of the process being trimmed out as well.

i will add an earlier filtering for pages which don't have dynamic params so that the process can be optimized to its maximum.

y-nk:
to bring a bit of light on this further, the project i'm testing this on has ~@1000 pages (~100x10 locales) pulled from a distant strapi backend.

the full build with `output: "hybrid"` is ~5mn (i also have further optimizations to help speed up the build process such as cached/deduped fetch calls)

when i add `filterPage: (path) => path.startsWith("/fr/")` in my condig i get only the expected pages built, and the whole process under 1mn.

what i mean is that even tho we pay the cost of js/css, the performances are still quite acceptable and the output is as expected ; at least from what I'm testing... could you propose some cases which would help me tackle corner cases?

ematipico:
> @ematipico thanks for your time spent on reviewing this.
> 
>     1. I'm honestly lost at the generate/build nuance. i'm not sure to have enough knowledge to understand this.
> 

We can prevent bundling a page easily, that's where you can write your code. `pageData` has everything you need

https://github.com/withastro/astro/blob/25e04a2ecbda7952a68220ce6739ae1c75144858/packages/astro/src/core/build/plugins/plugin-pages.ts#L45-L47

>     2. i first thought on putting the filtering where you pointed, but it does not expand the filtering to paths generated from `getStaticPaths` (with dynamic params in it). For this to happen it needs to be within `generatePage`. i'm thinking we could also add the "raw" path - containing the params - too and decide to skip import it if the function returned false (like a grouped filter, if you follow me)
> 

Oh, so your feature needs to work on the final URL? It wasn't very clear from your description. Then we should not called `pagePath` but `pageURL`, or `generatedPageUrl`. Maybe we should update the description to highlight the fact that the feature is meant to filter generated pages after the execution of `getStaticPaths`. It's really important.

Given this, I suppose my first point doesn't apply anymore.

>     3. the feature is definitely not meant for SSR, that's why i put the config in `build` rather than root level. it is intended for build only, and for filtering pre-renderable (prerender=true) route paths only (writing this i realize that api routes are not - yet - covered here)
> 

That's definitely another aspect of your proposed feature that should be highlighted in the description.

> 
> The usecase addressed is to skip pages which should not be built, defined by a criteria in userland. this might not make sense when using content collection, but it does when using a 3rd party CMS (but one could still use the output of a git diff of `src/content` maybe?)
> 
> Applied to my current usecase, my project has `src/pages/[...path].astro` where all route paths are pulled from a strapi backend and generated at built. i want to wire strapi's webhook to my build github action so that only updated pages get rebuilt and deployed.

packages/astro/src/core/build/generate.ts
```diff
@@ -257,24 +257,39 @@ async function generatePage(
...
+	const skippedPaths = paths.filter((path) => !filteredPaths.includes(path));
+	skippedPaths.forEach((path) => builtPaths.delete(removeTrailingForwardSlash(path)));
+
+	const routesSkipped =
+		filteredPaths.length === 0
+			? dim('[skipped fully]')
+			: filteredPaths.length < paths.length
+			? dim(`[skipped ${paths.length - filteredPaths.length} paths of ${paths.length}]`)
+			: '';
+
```
y-nk, line 276: the `.filter` call could be moved later (at L339) if there's no value in displaying the `(filtered/total)` stat in the logs, which would make the chunk of L266-L268 and L273-L274 useless.

y-nk: with a `pagePath => pagePath.startsWith('/fr')`, the render looks like this:

```
▶ src/pages/blog/category/[...path].astro [skipped fully]
▶ src/pages/blog/[...path].astro [skipped fully]
▶ src/pages/[...path].astro [skipped 161 paths of 191]
  ├─ /fr/index.html (+2.28s)
^C
```

y-nk:
@ematipico Thanks for your feedback. I'll proceed with the renaming of `pagePath` to be `pageURL`. You're right that it's more proper. Also I'll update the description of the PR accordingly, i _think_ you're right and most people reviewing got the wrong idea due to a bad description.

packages/astro/src/@types/astro.ts
```diff
@@ -892,6 +892,26 @@ export interface AstroUserConfig {
...
 		excludeMiddleware?: boolean;
+
+		/**
+		 * @docs
+		 * @name build.filterPage
+		 * @type {(pageURL: string) => boolean}
+		 * @default `undefined`
+		 * @version 2.10.0
+		 * @description
+		 * When defined, will be called for every single page about to be generated and gives opportunity to skip the page. When returning `false`, the page matching the finale `pageURL` will be excluded from the static build.
```
Yan-Thomas, line 903: Some docs wording suggestions, I've tried to make it closer to the format we have in other APIs docs, while adding some extra clarification. 

If I got it right, this should also apply to prerendered pages in SSR, right? That's something I'd like to clarify if true.
```suggestion
* Defines a function that will check which pages should be filtered out from the build while returning `false`.
*
* When defined, its function will be called for every single page about to be generated and gives the opportunity to skip the page. The criteria defined by the user based on `pageURL`, while returning `false`, will exclude any generated pages from the static build (including prerendered pages).
```

Yan-Thomas (commented):
Thanks @y-nk, that sounds like a nice feature, I've a docs comment/suggestion for you!

packages/astro/src/@types/astro.ts
y-nk: @Yan-Thomas thank a lot for your suggestions, i applied them right away.

bluwy:
From a brief look, I think this would be a nice feature as experimental for now. I think it's safer too to only skip on generation rather than build since we're tracking inter-page information during build, and skipping some might affect the final chunk names that otherwise the filtered pages would wrongly use.

The docs should probably mention that it only works for static builds though.

y-nk:
@bluwy hello and thanks for your time spent on the review.

1. as i understand you would rather remove the code added in `page-data`?
2. i'll try to update the docs again!
3. i'm not sure on how to create an "experimental feature"; could you briefly (bullet-point style) explain to me the big steps? from there i'm sure i can pick it up.

🙏

bluwy:
> 1. as i understand you would rather remove the code added in `page-data`?

I _think_ so. In your case with CMSes, they likely won't import any of your local style or scripts so you'd less likely see the issue (?). But for those who specifically wanted to build certain pages only, they might see the inconsistency. We currently have [this logic](https://github.com/withastro/astro/blob/db8c04002f51e89df57d44680e9e562105505aa4/packages/astro/src/core/build/plugins/plugin-css.ts#L78-L104) to determine what name to give a CSS chunk deriving from pages, and having missing pages could cause different names to appear.

> 3\. i'm not sure on how to create an "experimental feature"; could you briefly (bullet-point style) explain to me the big steps? from there i'm sure i can pick it up.

Making it experimental is probably my opinion/instinct for now as I feel like ultimately we could have a better implementation like ISR? So optional unless another maintainer chimes in, but to acheive that:

1. It would be very similar changes you did in `schema.ts`, but to the `experimental:` field, e.g. [here](https://github.com/withastro/astro/blob/db8c04002f51e89df57d44680e9e562105505aa4/packages/astro/src/core/config/schema.ts#L233-L240).
2. Then it's mostly changing `build.filterPages` to `experimental.buildPages`. We did this similarly for `inlineStylesheets`.

---

Thanks for keeping up on the PR!

bluwy:
Thanks I might have confused how ISR works, but what I had in mind is definitely in line with yours. I was thinking writing the build metadata somewhere and using it on the next run to identify which page needs to be re-generated.

ShivamJoker:
I would love to have an option which generates the pages/paths which was updated.

florian-lefebvre:
I haven't read every single comment but I think it could make sense to also apply this for SSR pages. Here is my usecase:

I'm using [Keystatic](https://keystatic.com/) and following [their guide](https://keystatic.com/docs/installation-astro#astro-s-ssr-hybrid-mode) we make this page ssr + client only.
However, because I'm using [local strategy](https://keystatic.com/docs/local-vs-git-hub-storage-strategy#local-strategy), it only makes sense to keep it in dev mode.

So this PR is a great start, but I think having a way to disable SSR pages as well would be great.

Another thought: I like the `export const prerender = boolean` api to opt out of prerendering. I think it could also make sense in this context with something like `export const exclude = boolean`. Other names could be: include, generate

y-nk:
@florian-lefebvre to be clear my goal is *not* to introduce a "do not include this route in the build" api. 

my goal is to have static routes generated from dynamic params (with `getStaticPaths`, which could he 10000s) to avoid being all generated when only a subset needs to be regenerated.

if you don't want a ssr route in the build, and ou can simply prefix with a `_` before you build your project. you can even automate this with a prebuild/postbuild npm script (`mv src/pages/my-ssr.astro src/pages/_my-ssr.astro` and back)

florian-lefebvre:
Thanks for the clarification! Please just ignore my usecase then as it is out of scope 👍

y-nk:
@ematipico @bluwy sorry for not updating earlier as i've been away for a long weekend afk ; i'm back and will take some time later this week to address this PR (as well as the other one).

bluwy:
No worries! We can always review and release it whenever you're ready.

ematipico:
@y-nk if the PR is not ready, can you convert it to a draft? It's very difficult to know when the PR is ready.

natemoo-re:
Hey @y-nk! Thank you for opening this PR and kicking this discussion off. We just [accepted the stage 2 proposal for incremental builds](https://github.com/withastro/roadmap/issues/698) which will supersede this PR. It is still in a very early exploratory phase, but it's a feature we've committed to exploring and, assuming all goes well, implementing.

I believe this will address your use case, but [the ISR proposal](https://github.com/withastro/roadmap/discussions/228) is still open as well. We can continue this conversation on whichever proposal you feel is more relevant.

y-nk:
@natemoo-re super happy to see this going forward. of course i would have loved for my PR to be merged but i understand the processes and the bigger picture! i'll try to stay involved in there

PR #7776, opened 2023-07-24
Match the .well-known directory when generating routes.json

aditsachde:
## Changes

This changeset causes the glob engine to match files in directories that start with a dot when generating a _routes.json file. This is necessary for matching files in a `.well-known` directory.

## Testing

image

When run with this change, the build process successfully generates an entry for "/.well-known/keybase.txt", which is otherwise excluded. 

## Docs

The docs currently state that all files in the `dist/` folder will be listed in the `_routes.json` file, and this change brings the code in line with the expected behavior listed in the documentation.

```
By default, @astrojs/cloudflare will generate a _routes.json file that lists all files from your dist/ folder and redirects from the _redirects file in the exclude array.
```


ematipico (approved):

.changeset/clean-planets-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
```
ematipico, line 2:
```suggestion
'@astrojs/cloudflare': patch
```

I am going to flag it as a patch, because it aligns with the behaviour we already mention in the docs

ematipico (commented):
It would be great if we could add a test though. Is it possible?

bluwy (approved):

Issue #7775, opened 2023-07-23
Minimal example fails with @astrojs/cloudflare


TheOtterlord:
### What version of `astro` are you using?

2.9.1

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

A minimal Astro project with the Cloudflare adapter installed, fails in production with this error:
```json
  "exceptions": [
    {
      "name": "TypeError",
      "message": "Failed to execute 'encode' on 'TextEncoder': parameter 1 is not of type 'String'.",
      "timestamp": 1690145312706
    }
  ],
```

Steps to reproduce:
1. Start a new Astro project. 
2. Add the Cloudflare adapter.
3. Deploy to CF Pages.

### What's the expected result?

Renders the minimal template page as expected.

### Link to Minimal Reproducible Example

https://github.com/TheOtterlord/astro-cf-repro/tree/9b9f172858f3c967d914a26b42bd8350d4eb947f

### Participation

- [ ] I am willing to submit a pull request for this issue.

wolves:
I am experiencing this as well and commented on an older related(closed) issue with my details as well.
https://github.com/withastro/astro/issues/6081#issuecomment-1646600176

PR #7777, opened 2023-07-24
Fix rendering TextEncoder encoding error regression

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7775

My bad. My PR (https://github.com/withastro/astro/pull/7730) accidentally reverted the fix from https://github.com/withastro/astro/pull/6070. I tried making a test, but I can't reproduce it. Might be a difference between wrangler and cloudflare prod?

## Testing

Not sure how to test this. I can't reproduce it with wrangler.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/dry-cycles-appear.md
+++ b/.changeset/dry-cycles-appear.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix rendering TextEncoder encoding error regression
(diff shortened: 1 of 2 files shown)


reneklacan:
@bluwy Thanks for fixing this. Any easy way to add a regression test to prevent it from happening in the future?

reneklacan:
> Might be a difference between wrangler and cloudflare prod?

I personally experienced error both with Wrangler and in CF production

Princesseuh (approved):

bluwy:
@reneklacan that would be much appreciated! I'll merge this for now so we can get a release out to fix the regression.

PR #7778, opened 2023-07-24
feat(assets): support remote images

y-nk:
## Changes

This enable the processing of remote images with the `astro:assets`. It will download the images into a cache folder and process them as if it was any other image. 

EDIT: after discussing with @Princesseuh on the subject, I've added a round of security measures through the `domains` and `remotePatterns` configuration options, which works similarly to `next/image`'s one (described here: https://nextjs.org/docs/app/api-reference/components/image#remotepatterns)

## Testing

I've forked astro locally, iterated and tested manually both for `dev` and `build` mode with a sample route such as:

```astro
---
export const prerender = true

import { Image } from 'astro:assets'

import localImage from '~/assets/600x400.png'

const remoteImage = 'https://placehold.co/600x400.png'
---

<>
  <Image src={localImage} width={60} height={40} alt="" />
  <Image src={remoteImage} width={60} height={40} alt="" />
</>
```

In my config:

```
{
  image: {
    service: sharpImageService(),
    domains: ['placehold.co'],
  }
}
```

## Docs

I didn't add docs yet because at this point I'm not sure the PR will be approved. I'd rather wait for round of feedbacks and approval of code, settle on API and _then_ add docs before merging 🙏


ematipico:
Hi @y-nk , thank you for your contribution. Considering that there isn't an RFC for this feature, could you please describe the feature you're implementing in detail? 

The code seems to do much more (remote patterns? domain? etc.) compared to the description.

y-nk:
@ematipico sure can do. if you want the full discussion, you can refer to [this discord thread](https://discord.com/channels/830184174198718474/1131988547172302938). Let me know if the following is clear enough ; if not i'll be glad to explain further.

---

## Original problem

The current state of `astro:assets` does not allow for remote images to be processed. It is a problem for me (who need to import images from a strapi backend) but it has also been pointed out 2 or 3 times over the last week as "something necessary/something we should do" in the discord server... I opened a thread to discuss the issue and found good support and guidance on how to kickstart this so i thought... why not trying to solve my problem and help the team?

## First implementation

The original scope of the PR (thanks to @Princesseuh's guidances and contributions) only proposed to add the processing of remote images as well. Basically, when writing:

```tsx
import { Image } from 'astro:assets'
<Image src="https://placehold.co/640x480.png />
```

the result will be an `` which url is going through the asset pipeline (`/_image?href=`)

## Wrapping up

After I've done this, I discussed it with @Princesseuh who suggested that for security reasons we should add a level of filtering for remote images, in the same way that `next/image` works. I personally have no culture for how this feat is implemented, so I thought that the vercel way would be a good starting point.

The `next/image` feature comes with 2 levels of allowances for enabling the processing of remote images called `domains` and `remotePatterns`. You can find the documentation here: https://nextjs.org/docs/app/api-reference/components/image#remotepatterns

Basically, `remotePatterns` is an object-based url matching algorithm while `domains` which is simpler by design is a string based strict equality on the domain requesting the image.

So I asked Erika which of the 2 levels she would like, and she answered "`domains` first, but ultimately we'll need both" so i thought that with a good design i could do both in one shot (and i think i did well).

## What's in the PR

**The final result** is adding remote image processing to `astro:assets`, but prohibited by default, hidden behind either `domains` or `remotePatterns` options, available in the `AstroUserConfig` under `config.image.domains` or `config.image.remotePatterns`.

### Side notes

As for the algorithm behind `remotePatterns`, i did not want to do the vercel way ([here](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/match-remote-pattern.ts)) because _i think_ (?) my way is faster and does not require installing a 3rd party package (`micromatch`). To make sure all is covered I added unit tests, but i'll be happy to add more cases if you require some.

pijusz:
Hi, just wanted to ask if there is any progress with remote images? As y-nk has mostly finished this feature

Princesseuh:
> Hi, just wanted to ask if there is any progress with remote images? As y-nk has mostly finished this feature

I'm hoping to get to it this week, it seems like the current implementation break some things looking at the test and there's also some docs work needed for it.

y-nk:
@Princesseuh sorry for not updating earlier as i've been away for a long weekend afk ; i'm back and will take some time later this week to address this.

tusamni:
@y-nk Thanks for working on this, I've been hoping something like this would happen eventually.

futurefabric:
Really pleased this work is (hopefully) about to drop, thank you for looking into this 🙏
(I'm v. keen to move to `astro:assets` ahead of the deprecation of `astrojs/image` but can't do so without this support for remote images.)

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
...
+		 * @docs
+		 * @name image.domains (Experimental)
+		 * @type {Array<string>}
+		 * @default `{domains: []}`
+		 * @version 2.9.2
+		 * @description
+		 * Defines a list of domains allowed to be optimized locally
+		 *
+		 * The value should be an array of domain strings of images allowed to be optimized locally.
+		 * Wildcards are not permitted (see image.remotePatterns).
```
sarah11918, line 1026:
```suggestion
		 * Defines a list of permitted image source domains for local image optimization. No other remote images will be optimized by Astro.
		 *
		 * This option requires an array of individual domain names as strings. Wildcards are not permitted. Instead, use [`image.remotePatterns`](#imageremotepatterns-experimental) to define a list of allowed source URL patterns.
```
I'd like to see something more in this style maybe? (Make the wording accurate!)

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 *   image: {
+		 *     // Example: Enable the Sharp-based image service
+		 *     domains: ['astro.build'],
+		 *   },
+		 * }
```
sarah11918, line 1034:
```suggestion
		 * // astro.config.mjs
		 * {
		 *   image: {
		 *     // Example: Allow remote image optimization from a single domain
		 *     domains: ['astro.build'],
		 *   },
		 * }
```
I'll also point out that the changeset shows a trailing slash, and this example doesn't, so just make them both consistent with whatever it should be!

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
...
+		domains?: string[];
+
+		/**
+		 * @docs
+		 * @name image.remotePatterns (Experimental)
+		 * @type {Array<RemotePattern>}
+		 * @default `{remotePatterns: []}`
+		 * @version 2.9.2
+		 * @description
+		 * Defines a list of url patterns allowed to be optimized locally
```
sarah11918, line 1046:
```suggestion
		 * Defines a list of permitted image source URL patterns for local image optimization.
```
Again, make this accurate, but I'm trying to hint at the idea of a remote image source. (The original descriptions of these didn't mention images at all, for example.)

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
...
+		 *
+		 * The patterns can contain 4 properties:
```
sarah11918, line 1048:
```suggestion
		 * `remotePatterns` can be configured with four properties:
```

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
...
+		 *   image: {
+		 *     // Example: statically process all images from your aws s3 bucket
+		 *     remotePatterns: [{
+		 *       protocol: 'https',
+		 *       hostname: '**.amazonaws.com',
+		 *     }],
+		 *   },
+		 * }
+		 * ```
+		 */
```
sarah11918, line 1078:
``````suggestion
		 * ```js
		 * {
		 *   image: {
		 *     // Example: statically process all images from your aws s3 bucket
		 *     remotePatterns: [{
		 *       protocol: 'https',
		 *       hostname: '**.amazonaws.com',
		 *     }],
		 *   },
		 * }
		 * ```
		 * 
		 * You can use wildcards to define the permitted `hostname` and `pathname` values as described below. Otherwise, only the exact values provided will be configured:
		 * `hostname`:
		 *   - Start with '**.' to allow all subdomains ('endsWith').
		 *   - Start with '*.' to allow only one level of subdomain.
		 * 
		 * `pathname`:
		 *   - End with '/**' to allow all sub-routes ('startsWith').
		 *   - End with '/*' to allow only one level of sub-route.
		 */
``````
I think showing the example first is more helpful, so I've swapped these. Also, I'm not sure, but I seem to see "subroute`"more often as one word or hyphenated when I look it up. Making it one word here gave me red squiggles, but a hyphen didn't. So, arbitrarily, I'm suggesting that one. But if you know for sure, then choose that one.

Also, the reader is probably less interested in reading this from a "what your feature's matching patterns are" standpoint as they are in knowing *how to configure* for what they want. So I've made a suggestion where you can see what framing it this way this looks like. Bonus, lines are shorter, more direct, and are instructions to the users.

File: .changeset/sour-frogs-shout.md [hunk 6]
```diff
@@ -0,0 +1,25 @@
+---
+'astro': patch
+---
+
+Added support for optimizing remote images to `astro:assets`. This comes with two new parameters to specify which domains and host patterns are authorized for remote images.
```
sarah11918, line 5:
```suggestion
Added support for optimizing remote images from authorized sources when using `astro:assets`. This comes with two new parameters to specify which domains (`image.domains`) and host patterns (`image.remotePatterns) are authorized for remote images.
```
I think adding "from authorized sources" right in the first sentence is helpful for "skimmers", because that will immediately make them realize that they need to somehow *do* something re: specifying authorization, vs just "Yay, remote images now!"

.changeset/sour-frogs-shout.md
```diff
@@ -0,0 +1,25 @@
...
+
+For example, the following configuration will only allow remote images from `astro.build` to be optimized:
+
+```ts
+export default defineConfig({
+  image: {
+    domains: ["astro.build/"],
+  }
+});
```
sarah11918, line 14:
```suggestion
// astro.config.mjs
export default defineConfig({
  image: {
    domains: ['astro.build'],
  }
});
```
Added the file name, and also changed the value here to match the docs example (though, THIS one may in fact be correct, in which case, change the other one so that they match). One had a trailing slash and one didn't, and even if they're both OK, we should show the same example in both places.

packages/astro/src/@types/astro.ts
```diff
@@ -1010,6 +1012,71 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		service: ImageServiceConfig;
+
+		/**
+		 * @docs
+		 * @name image.domains (Experimental)
+		 * @type {Array<string>}
+		 * @default `{domains: []}`
+		 * @version 2.9.2
```
sarah11918, line 1021:
```suggestion
		 * @version 2.10.10
```
?? Check this! I'm guessing though that 2.9.2 is not what you intended.

sarah11918 (commented):
Thanks @y-nk and @Princesseuh for this cool feature! I've left some suggestions below for your consideration re: docs!  :raised_hands:

packages/astro/src/assets/build/generate.ts
```diff
@@ -0,0 +1,172 @@
...
+					cached: true,
+				};
+			}
+		}
+	} catch (e) {
+		// no-op. We'll generate the image below if copying from cache failed
+	}
+
+	// The original filepath or URL from the image transform
+	const originalImagePath = isESMImportedImage(options.src) ? options.src.src : options.src;
```
ematipico, line 91: No need to call this again, we already did and its value is `isLocalImage`

packages/astro/src/assets/build/remote.ts
```diff
@@ -0,0 +1,55 @@
...
+		// calculate an expiration date based on the response's TTL
+		const policy = new CachePolicy(webToCachePolicyRequest(req), webToCachePolicyResponse(res));
+		const expires = policy.storable() ? policy.timeToLive() : 0;
+
+		return {
+			data: Buffer.from(await res.arrayBuffer()),
+			expires: Date.now() + expires,
+		};
+	} catch (err: unknown) {
+		console.error(err);
```
ematipico, line 27: We should use our logger

packages/astro/src/assets/build/remote.ts
```diff
@@ -0,0 +1,55 @@
+import CachePolicy from 'http-cache-semantics';
+
+export type RemoteCacheEntry = { data: string; expires: number };
+
+export async function loadRemoteImage(src: string) {
+	try {
+		if (src.startsWith('//')) {
+			src = `https:${src}`;
```
ematipico, line 8: I don't think we should prefix the protocol. Many developers use `//` on purpose because internal environments (`local`, `dev`, `test`, `staging`, `pre-prod`, etc.) might not be under `https`.

packages/astro/src/assets/build/remote.ts
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		const req = new Request(src);
+		const res = await fetch(req);
+
+		if (!res.ok) {
+			return undefined;
```
ematipico, line 15: It would be nice if we could log a warning/debug that says that the image returned a non-200 status code.

packages/astro/src/assets/internal.ts
```diff
@@ -22,6 +23,30 @@ export function isESMImportedImage(src: ImageMetadata | string): src is ImageMet
...
+	{
+		domains = [],
+		remotePatterns = [],
+	}: Partial<Pick<AstroConfig['image'], 'domains' | 'remotePatterns'>>
+): boolean | undefined {
+	let url: URL;
+	try {
+		url = new URL(src);
+	} catch (e) {
+		return undefined;
```
ematipico, line 41: We should **at least** comment and say what are the cases where this operation can fail.

packages/astro/src/assets/build/generate.ts
```diff
@@ -0,0 +1,172 @@
...
+		if (isLocalImage) {
+			await fs.promises.copyFile(cachedFileURL, finalFileURL);
+
+			return {
+				cached: true,
+			};
+		} else {
+			const JSONData = JSON.parse(readFileSync(cachedFileURL, 'utf-8')) as RemoteCacheEntry;
+
+			console.log(JSONData.expires, Date.now());
```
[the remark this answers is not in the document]
Princesseuh: That one should actually not be there at all, oops

packages/astro/src/assets/build/remote.ts
Princesseuh: Well, the problem is that there's no way for us to know the protocol from here, we have to assume one

ematipico: I'm not very familiar with the business logic, but why do we need to prepend the https?

Princesseuh: To actually fetch the remote image 😄 In SSR, it's totally fine for users to do `//...` because we can assume the protocol from the current request, however in SSG, `//` means nothing, because there's no request to infer it from.

Doing `fetch("//...")` will throw with an invalid URL (of course), so we need to prepend a protocol.

.changeset/sour-frogs-shout.md
```diff
@@ -0,0 +1,27 @@
+---
+'astro': patch
+---
+
+Added support for optimizing remote images from authorized sources when using `astro:assets`. This comes with two new parameters to specify which domains (`image.domains`) and host patterns (`image.remotePatterns) are authorized for remote images.
```
sarah11918, line 5:
```suggestion
Added support for optimizing remote images from authorized sources when using `astro:assets`. This comes with two new parameters to specify which domains (`image.domains`) and host patterns (`image.remotePatterns`) are authorized for remote images.
```

.changeset/sour-frogs-shout.md
```diff
@@ -0,0 +1,27 @@
...
+```ts
+// astro.config.mjs
+export default defineConfig({
+  image: {
+    domains: ["astro.build"],
+  }
+});
+```
+
+The following configuration will only allo remote images from HTTPs hosts:
```
sarah11918, line 18:
```suggestion
The following configuration will only allow remote images from HTTPS hosts:
```
maybe the S should be capitalized here, too?

sarah11918:
I'm noticing now that there's no mention in the description as to whether this is *exclusively* for the `<Image />` component, or whether it also works with the `` tag.

It doesn't have to affect *this* PR, but for the v3 docs I'm writing, I'd like to know so I can be explicit! Can someone mention me here and let me know one way or the other? Thanks!  :raised_hands: 

@Princesseuh @y-nk

Princesseuh:
> I'm noticing now that there's no mention in the description as to whether this is *exclusively* for the `<Image />` component, or whether it also works with the `` tag.
> 
> It doesn't have to affect *this* PR, but for the v3 docs I'm writing, I'd like to know so I can be explicit! Can someone mention me here and let me know one way or the other? Thanks!  :raised_hands: 
> 
> @Princesseuh @y-nk 

This is only for image processing, so only when using `Image` or `getImage`. For normal img tags, everything is allowed and we do no processing at all. 

The technical reasons for all the fluff of domains and remote patterns regarding processing remote images is that the server (not the user's browser) has to fetch the images to process them, so you don't want users to be able to make your server fetch any random URLs

packages/astro/src/assets/build/remote.ts
ematipico: Oh, I see. It's a gamble. Shouldn't we try to fetch the image with both protocols?

ematipico (approved):
Thank you for addressing my comments! Let's 🚢 it!

packages/astro/src/assets/build/generate.ts
```diff
@@ -0,0 +1,174 @@
...
+			await fs.promises.copyFile(cachedFileURL, finalFileURL);
+
+			return {
+				cached: true,
+			};
+		} else {
+			const JSONData = JSON.parse(readFileSync(cachedFileURL, 'utf-8')) as RemoteCacheEntry;
+
+			// If the cache entry is not expired, use it
+			if (JSONData.expires < Date.now()) {
```
notjosh, line 78: @y-nk sorry to comment on a closed PR, but is this logic correct? I _think_ it should be `>`, but I may be missing something obvious.

(but big ❤️❤️❤️ for this PR overall - it's super helpful to have remote images optimised locally!)

y-nk: @notjosh i'll be honest i don't remember writing that line. maybe @Princesseuh have more clue about it. from reading the if it seems you're right, but i'm not so sure.

Princesseuh: This was fixed in a PR after this one!

PR #7780, opened 2023-07-24
Remove support for Node 16

Princesseuh:
## Changes

Move our lowest supported version to v18.14.1. That specific version of Node 18 was chosen as previous versions had issues in their fetch implementation. While those issues don't necessarily affect the majority of users, it's safer for us to recommend a version we know work in every situation.

Moving up a Node version also allow us to uses TypeScript's newest features, as such this also migrates us to `moduleResolution: 'bundler'`.

Changes to `@astrojs/webapi` were skipped in this PR, I'll tackle those separately to make sure I don't remove needed polyfills and stuff.

## Testing

Tests should all pass!

## Docs

Will do a PR to docs to update the installation page


lilnasy:
Should we address this?
https://github.com/withastro/astro/blob/astro@2.9.2/packages/webapi/src/ponyfill.ts#L18
```ts
import { File, FormData, Headers, Request, Response, fetch } from 'undici' // Remove when Node 16 is dropped for Node 18.
```

It would help deno and bun if they could use their own implementation and slightly reduce how long users have to wait when installing astro.

Princesseuh:
> Should we address this? https://github.com/withastro/astro/blob/astro@2.9.2/packages/webapi/src/ponyfill.ts#L18
> 
> ```ts
> import { File, FormData, Headers, Request, Response, fetch } from 'undici' // Remove when Node 16 is dropped for Node 18.
> ```
> 
> It would help deno and bun if they could use their own implementation and slightly reduce how long users have to wait when installing astro.

Intentionally skipped in this PR, I plan to tackle the `webapi` changes separately

ematipico (approved):

packages/telemetry/package.json
```diff
@@ -13,7 +13,10 @@
   "bugs": "https://github.com/withastro/astro/issues",
   "homepage": "https://astro.build",
   "exports": {
-    ".": "./dist/index.js",
+    ".": {
+      "default": "./dist/index.js",
+      "types": "./dist/types/index.d.ts"
```
bluwy, line 18: The `"types"` need to come before `"default"` since exports condition is order-sensitive.

tsconfig.base.json
```diff
@@ -3,7 +3,8 @@
     "declaration": true,
     "emitDeclarationOnly": true,
     "strict": true,
-    "moduleResolution": "node",
+    // All packages are built with ESBuild, so we can use `moduleResolution: 'bundler'`
+    "moduleResolution": "Bundler",
```
bluwy, line 7: Even though they are built with esbuild, the runtime would still break if we import a path without `.js`. esbuild doesn't append the `.js` automatically it seems. Would it be more accurate to use `node16` instead? Or maybe figure out adding the `.js`?

I think `bundler` make more sense for user's astro projects though.

bluwy: Well not blocking the PR for sure, we can also tackle this existing issue in a later PR.

packages/integrations/vue/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "allowJs": true,
     "module": "ES2022",
     "outDir": "./dist",
-    "target": "ES2021"
+    "target": "ES2022",
+    "verbatimModuleSyntax": false
```
Princesseuh, line 9: Q: Why this? (and same in Svelte's)
A: `verbatimModuleSyntax` prevents you from using ESM syntax in CJS files. So it creates an error in the `.cts` file asking it to use CJS syntax. Which is fine, but actually using CJS syntax causes more errors to appear (probably because our codebase assume full ESM). I couldn't figure it out, so I just disabled it.

Hopefully in the future VS Code supports ESM extensions.

PR #7781, opened 2023-07-24
feat: feature map for adapters

ematipico:
## Changes

This PR adds a new API that adapters can use when registering one. It's called `supportedFeatures`, and it's a new way to tell Astro if and when an adapter is able to support certain features that are enabled via configuration. 

The adapter can also tell Astro the kind of support:
- stable
- experimental
- unsupported
- deprecated

For now, this new API/object is not mandatory, but it will be in Astro 4.0.

The PR also updates the adapters that we maintain, and it adds the feature map to them.

### About `assets`

There's a small caveat for the assets. Astro assets is a feature that is already turned on by default, meaning that Astro sets already a default service. This means that adapters that **don't support** assets will always get an error, which is inconvenient. I came up with a solution where we set the default **after** the adapter validation, although I don't like the solution very much, and I wonder if there's a better way to handle a case like this.

The assets feature is just an example; we might have the same problem with future features like this. What do you suggest?

## Testing

I added new unit test cases and updated the existing ones.

## Docs

I will create the changelog in another PR

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1660,12 +1660,55 @@ export type PaginateFunction = (data: any[], args?: PaginateOptions) => GetStati
 
 export type Params = Record<string, string | undefined>;
 
+export type SupportsKind = 'Unsupported' | 'Stable' | 'Experimental' | 'Deprecated';
+
+export type AstroAdapterFeatureMap = {
+	/**
+	 * The adapter is able serve static pages
+	 */
+	staticOutput?: SupportsKind;
+	/**
+	 * The adapter is able to serve pages that are static or rendered via server
+	 */
+	hybridOutput?: SupportsKind;
+	/**
+	 * The adapter is able to serve SSR pages
+	 */
+	serverOutput?: SupportsKind;
+	/**
+	 * Support for emitting a SSR file per page
+	 */
+	functionPerPage?: SupportsKind;
+	/**
+	 * Support when `build.ecludeMiddleware` is enabled.
+	 */
+	edgeMiddleware?: SupportsKind;
+	/**
+	 * The adapter can emit static assets
+	 */
+	assets?: {
+		supportKind?: SupportsKind;
+		/**
+		 * Whether if this adapter deploys files in an enviroment that is Node.js compatible.
+		 *
+		 * The default services used by Astro needs Node.js.
+		 */
+		isNodeCompatible?: boolean;
+	};
+};
+
 export interface AstroAdapter {
 	name: string;
 	serverEntrypoint?: string;
 	previewEntrypoint?: string;
 	exports?: string[];
 	args?: any;
+	/**
+	 * List of features supported by an adapter.
+	 *
+	 * If the adapter is not able to handle certain configurations, Astro will throw an error.
+	 */
+	supportedFeatures?: AstroAdapterFeatureMap;
 }
 
 type Body = string;
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1676,14 +1676,6 @@ export type AstroAdapterFeatureMap = {
 	 */
 	serverOutput?: SupportsKind;
 	/**
-	 * Support for emitting a SSR file per page
-	 */
-	functionPerPage?: SupportsKind;
-	/**
-	 * Support when `build.ecludeMiddleware` is enabled.
-	 */
-	edgeMiddleware?: SupportsKind;
-	/**
 	 * The adapter can emit static assets
 	 */
 	assets?: {
--- a/packages/astro/src/assets/generate.ts
+++ b/packages/astro/src/assets/generate.ts
@@ -27,6 +27,11 @@ export async function generateImage(
 	options: ImageTransform,
 	filepath: string
 ): Promise<GenerationData | undefined> {
+	if (typeof buildOpts.settings.config.image === 'undefined') {
+		throw new Error(
+			"Astro hasn't set a default service for `astro:assets`. This is an internal error and you should report it."
+		);
+	}
 	if (!isESMImportedImage(options.src)) {
 		return undefined;
 	}
--- a/packages/astro/src/assets/vite-plugin-assets.ts
+++ b/packages/astro/src/assets/vite-plugin-assets.ts
@@ -26,6 +26,12 @@ export default function assets({
 	logging,
 	mode,
 }: AstroPluginOptions & { mode: string }): vite.Plugin[] {
+	const imageService = settings.config.image;
+	if (typeof imageService === 'undefined') {
+		throw new Error(
+			"Astro hasn't set a default service for `astro:assets`. This is an internal error and you should report it."
+		);
+	}
 	let resolvedConfig: vite.ResolvedConfig;
 
 	globalThis.astroAsset = {};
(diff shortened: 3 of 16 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1660,12 +1660,47 @@ export type PaginateFunction = (data: any[], args?: PaginateOptions) => GetStati
 
 export type Params = Record<string, string | undefined>;
 
+export type SupportsKind = 'Unsupported' | 'Stable' | 'Experimental' | 'Deprecated';
```
matthewp, line 1663: Can you explain the reason for this?

lilnasy: I think `undefined` is inevitably going to become a `SupportsKind`. For example, when a new capability is added and a user updates astro but not the adapter.

bluwy, line 1663: Is there a reason we're not hiding `build.split`, `build.excludeMiddleware`, etc, but introducing a guard like this instead?

I was hoping we can support `vercel({ edgeMiddleware: true })` for example, which internally toggles some feature flag, e.g. `supportedFeatures.edgeMiddleware: true` in place of `build.excludeMiddleware: true`. This also makes the feature supported by adapters explicit in the types.

ematipico: @matthewp, I thought we could have some flexibility in the future with the features that expose to the adapters. Here's what I had in mind:

- Unsupported -> needed to throw an error
- Experimental -> a way to tell the user that the feature is new and might be buggy/change. The user will see a terminal log
- Deprecated -> to gracefully tell the user that some feature might be removed from the adapter
- Stable -> to tell Astro that a feature is supported

ematipico: > Is there a reason we're not hiding build.split, build.excludeMiddleware, etc, but introducing a guard like this instead?

Do you want to remove them? If so, it would be a breaking change.

> I was hoping we can support `vercel({ edgeMiddleware: true })` for example, which internally toggles some feature flag, e.g. `supportedFeatures.edgeMiddleware: true` in place of `build.excludeMiddleware: true`. This also makes the feature supported by adapters explicit in the types.

The main idea is to tell Astro what an adapter is capable of, and prevent the user from using some adapter/configuration when using x, y or z adapter.

The idea that you suggest can still be implemented. What I want to make clear is that there's a distinction:
- the support map is to tell Astro, "Hey, I - Vercel Adapter - am able to handle the `excludeMiddleware` option.
- your option is more like "Hey, I - user that is using Vercel Adapter - want **to use** this feature from the map.

This means that **wanting to use a feature** doesn't necessarily mean that the adapter is able to support it.

bluwy: Yeah I'm thinking more of no2 that makes more sense to me 🤔 What this becomes is not an Astro feature, but an adapter feature, and for features like "edge middleware" it sounds more of a responsibility/feature of an adapter, rather than Astro.

The feature set I'm thinking off, are a way to tweak Astro to output in a certain way. In most cases, adapters should trying tweaking the astro/vite config before resorting to that.

> Do you want to remove them? If so, it would be a breaking change.

Yeah I think we should remove them. Like above, I don't think they're Astro features, but adapter features. The options are glue code to tweak the Astro output in specific ways. It would be breaking, but worth the newer API design? Just my 2c 😬 

---

I think what this PR has now it's a bit of bouncing responsibilities, and makes it hard to tell what features the adapter supports until you run the adapter (where the error appears)

matthewp: Yeah, like @lilnasy said we are going to have to support `undefined` here. IIRC the reason to have this feature is so that when we add new features we can tell if an adapter is supporting it.

To that end, I don't think any adapter is going to set `Unsupported` on purpose. More likely it just never gets set at all until they come to support it.

So I'm wondering if having these different options might be premature. Would it not be better to make it a boolean `false | true` and default it to `false`? Then we could add additional flags like `Deprecated` in the future. But I'm not sure if we should add them until we have a use-case for them.

matthewp: How would we hide `build.split` and `build.excludeMiddleware`? I think we originally did it this way to avoid having another flag to support. But now it makes more sense; since users can set these things anyways don't we want a good way to know if they are going to work? Unless there's a real means to make these "private" to integrations.

ematipico: > Yeah, like @lilnasy said we are going to have to support undefined here. IIRC the reason to have this feature is so that when we add new features we can tell if an adapter is supporting it.

`Unsupported` will be exactly `undefined`. Use case: we release the `build.split` feature, and it's only added to the node adapter. Then, a user tries to use the feature for - for example - the deno adapter. Astro will throw an error because no value has been set, and the `Unsupported` is the default value. (I will have to fix the assignment of the default value). 

> To that end, I don't think any adapter is going to set Unsupported on purpose. More likely it just never gets set at all until they come to support it.

Actually, there are cases where features are **intetionally unsupported**. For example Vercel Edge won't support the edge middleware, so it makes sense to explicitly flag `Unsupported`. Another example is Cloudflare, which doesn't support `"static"` output.

ematipico: I still want to make sure that @bluwy 's proposal is unrelated to what I am proposing in this PR.

In this PR, I want to build a common foundation where adapters can tell Astro they support arbitrary features. It's **not** a proposal of how config/options should be passed to Astro.

So far, we have yet to have a standard way to validate certain things. Some example:
- https://github.com/withastro/astro/blob/main/packages/integrations/cloudflare/src/index.ts#L66-L71
- https://github.com/withastro/astro/blob/e5e5cc8e70bd7e842d8cdeb8614552e621e3aea0/packages/astro/src/assets/vite-plugin-assets.ts#L39-L61

That's what this PR wants to address.

bluwy: > How would we hide `build.split` and `build.excludeMiddleware`?

We can have an internal config like `settings._adapterFeatures.split` or `settings.config._adapterFeatures.split`.

> But now it makes more sense; since users can set these things anyways don't we want a good way to know if they are going to work?

I think I'm proposing that users should set these config from the adapter options instead. If it's available in the config types, it would definitely work for the adapter. Perhaps you're talking about the `staticOutput` feature stuff, which in response to Ema:

> In this PR, I want to build a common foundation where adapters can tell Astro they support arbitrary features. It's **not** a proposal of how config/options should be passed to Astro.

Putting it this way, yeah I think that make sense. For features like `staticOutput`, `hybridOuput`, `assets`, I think this PR creates a good way to log that since they are core Astro features.

I think I'm mostly prefer removing the `edgeMiddleware` and `functionsPerPage` features in this PR, since they're not core Astro features, but more of glue logic to make adapters output differently?

ematipico: > I think I'm mostly prefer removing the edgeMiddleware and functionsPerPage features in this PR, since they're not core Astro features, but more of glue logic to make adapters output differently?

You are right, and I agree with you. Although, technically, we still need it. Few examples:

- `edgeMiddleware`: the Vercel adapter can support this feature in "serverless" but doesn't for the "edge". So we need to tell Astro (and the user);
- `functionsPerPage`: we recently implemented this feature - `build.split` - for the Netlify adapter. Although, we implemented it only for the "functions". A few days ago, a user filed an issue because the "edge-functions" didn't support this feature. With this feature map, we could print an error "for free". And eventually, enable this feature once we implement the code to support it.

PR #7782, opened 2023-07-24
Refactor Astro rendering to write results directly

bluwy:
## Changes

Follow-up from https://github.com/withastro/astro/pull/7730, this PR uses the new `RenderDestination` abstraction to render the HTML directly, instead of using generators.

This is a big PR, and unfortunately hard to split off, but I'll try to leave comments for each changes.

Gist of changes:
1. Remove generators/yield, use `destination.write` instead.
2. Some generators are left as-is because things like `RenderInstruction`s are re-sorted and hoisted.
3. Parallel rendering simplified as `renderComponent` API now returns a promise that eagerly renders.

## Testing

Existing tests should pass.

## Docs

n/a.


bluwy:
!bench server-stress

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -349,67 +382,154 @@ function sanitizeElementName(tag: string) {
...
+}
+
+export async function renderComponent(
 	result: SSRResult,
 	displayName: string,
 	Component: unknown,
 	props: Record<string | number, any>,
 	slots: any = {}
-): Promise<ComponentIterable> | ComponentIterable | AstroComponentInstance {
+): Promise<RenderInstance> {
```
bluwy, line 448: `renderComponent` is the core API for rendering any component. Now it will return a promise so it's rendered in parallel, and return a `RenderInstance` - a new concept added, but basically an object with `{ render: (destination) => void }`. (destination is the `RenderDestination`)

packages/astro/src/runtime/server/render/slot.ts
```diff
@@ -49,17 +50,21 @@ export async function renderSlotToString(
...
 			}
-			instructions.push(chunk);
-		} else {
-			content += chunk;
-		}
-	}
+		},
+	};
+	const renderInstance = renderSlot(result, slotted, fallback);
+	await renderInstance.render(temporaryDestination);
```
bluwy, line 67: Slots are very tricky to render. They are split as `content` and `instructions`. `content` is the basic HTML, `instructions` are like astro island scripts/styles, and head content.

This out-of-order sorting makes it that we need a temporary destination to "collect" and sort the chunks.

packages/astro/test/streaming.test.js
```diff
@@ -48,7 +48,7 @@ describe('Streaming', () => {
 				let chunk = decoder.decode(bytes);
 				chunks.push(chunk);
 			}
-			expect(chunks.length).to.equal(3);
+			expect(chunks.length).to.equal(2);
```
bluwy, line 51: I _think_ because we render more stuff in sync now, it's now streamed in 2 chunks? This is the only test I had to tweak.

ematipico (approved):
I like this new API and way of rendering things, especially the `.write` function!

I like it so much that when I see things like `str +=`, it makes me wanna cry 😢 😆 

I am not a big fan of the closure approach, because it seems more a side effect. I wish we could find a better approach, to make it more "linear". 

Just sharing my thoughts, overall I am happy with the implementation :)

bluwy:
Thanks for reviewing! Yeah I'm also not quite liking the closures, it's kinda the same as generator yields, but without generators. Tried to avoid that, but it's a bit hard with the slots and hoisting. Maybe we can revisit that again with this new flow.

bluwy:
Since this is a large-ish change, I'll add a changeset to be safe so it's easier to track if it caused a regression somewhere.

PR #7783, opened 2023-07-24
feat: upper case the name of the endpoints

ematipico:
## Changes

> **Note**: Astro 3.0 breaking change

> **Note**: PR to `next` branch

Now the endpoints must be uppercase, and we don't support anymore the `del` function, but it will be `DELETE`.

## Testing

Updated all the tests that use the old endpoints. Current tests should all pass.

## Docs

### DIFF
--- a/.changeset/twelve-coats-rush.md
+++ b/.changeset/twelve-coats-rush.md
@@ -2,9 +2,9 @@
 'astro': major
 ---
 
-The endpoints that have **lower case** name are **deprecated**.
+The endpoints that have **lowercase** names are **deprecated**.
 
-The endpoints methods **needs** to be in **upper case**.
+The endpoints methods **need** to be in **uppercase**.
 
 ```diff
 --export function get() {
--- a/.changeset/twelve-coats-rush.md
+++ b/.changeset/twelve-coats-rush.md
@@ -2,7 +2,9 @@
 'astro': major
 ---
 
-The endpoints methods are not in **capital case**.
+The endpoints that have **lower case** name are **deprecated**.
+
+The endpoints methods **needs** to be in **upper case**.
 
 ```diff
 --export function get() {
--- a/.changeset/twelve-coats-rush.md
+++ b/.changeset/twelve-coats-rush.md
@@ -2,9 +2,9 @@
 'astro': major
 ---
 
-The endpoints that have **lowercase** names are **deprecated**.
+Lowercase names for endpoint functions are now deprecated.
 
-The endpoints methods **need** to be in **uppercase**.
+Rename functions to their uppercase equivalent:
 
 ```diff
 - export function get() {
--- a/.changeset/twelve-coats-rush.md
+++ b/.changeset/twelve-coats-rush.md
@@ -7,29 +7,29 @@ The endpoints that have **lowercase** names are **deprecated**.
 The endpoints methods **need** to be in **uppercase**.
 
 ```diff
---export function get() {
-++export function GET() {
+- export function get() {
++ export function GET() {
     return new Response(JSON.stringify({ "title": "Bob's blog" }));
 }
 
---export function post() {
-++export function POST() {
+- export function post() {
++ export function POST() {
     return new Response(JSON.stringify({ "title": "Bob's blog" }));
 }
 
---export function put() {
-++export function PUT() {
+- export function put() {
++ export function PUT() {
     return new Response(JSON.stringify({ "title": "Bob's blog" }));
 }
 
---export function all() {
-++export function ALL() {
+- export function all() {
++ export function ALL() {
     return new Response(JSON.stringify({ "title": "Bob's blog" }));
 }
 
-// you can use use the whole word "DELETE"
---export function del() {
-++export function DELETE() {
+// you can use the whole word "DELETE"
+- export function del() {
++ export function DELETE() {
     return new Response(JSON.stringify({ "title": "Bob's blog" }));
 }
 ```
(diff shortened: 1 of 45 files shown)


natemoo-re:
Just double checking [SvelteKit's endpoints](https://kit.svelte.dev/docs/routing#server) and making sure we support all the same things.

- [x] `GET`
- [x] `POST`
- [x] `PATCH`
- [x] `PUT`
- [x] `DELETE`
- [ ] `OPTIONS`
- [ ] `HEAD`

ematipico:
> Should we add OPTIONS and HEAD?

**In this PR**, I don't think we should. It's a new feature that we don't have. It can be implemented without a breaking change.

packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -1,28 +1,56 @@
...
+	const lowerCaseMethod = method.toLowerCase();
+
+	// NOTE: remove in Astro 4.0
+	if (mod[lowerCaseMethod]) {
+		warn(
+			logging,
+			'astro',
+			"Lower case endpoint points are deprecated. In Astro 4.0 it won't work anymore."
+		);
+		info(logging, 'astro', `Rename the endpoint ${lowerCaseMethod} to ${method}`);
```
bluwy, line 14: Maybe we can combine the two messages into one?
```suggestion
		warn(
			logging,
			'astro',
			`Lower case endpoint names are deprecated and will not be supported in Astro 4.0. Rename the endpoint ${lowerCaseMethod} to ${method}.`
		);
```

bluwy (approved):

.changeset/twelve-coats-rush.md
```diff
@@ -0,0 +1,35 @@
+---
+'astro': major
+---
+
+The endpoints that have **lower case** name are **deprecated**.
+
+The endpoints methods **needs** to be in **upper case**.
```
Yan-Thomas, line 7: Plural fixes, plus, we tend to use the lowercase/uppercase variant (one word) in docs, so while this is not incorrect, I would change for consistency.
```suggestion
The endpoints that have **lowercase** names are **deprecated**.

The endpoints methods **need** to be in **uppercase**.
```

Yan-Thomas (commented):
Thanks @ematipico, added a suggestion to improve the changeset 🙌

.changeset/twelve-coats-rush.md
```diff
@@ -0,0 +1,35 @@
...
+++export function ALL() {
+    return new Response(JSON.stringify({ "title": "Bob's blog" }));
+}
+
+// you can use use the whole word "DELETE"
+--export function del() {
+++export function DELETE() {
+    return new Response(JSON.stringify({ "title": "Bob's blog" }));
+}
+```
```
delucis, line 35: I’m not sure we usually use this double `--`/`++` syntax in diffs? Probably safest to stick with a single `-`/`+` at the start of the line. Also fixed the repeated “use” in the comment above `DELETE`.

````suggestion
```diff
- export function get() {
+ export function GET() {
    return new Response(JSON.stringify({ "title": "Bob's blog" }));
}

- export function post() {
+ export function POST() {
    return new Response(JSON.stringify({ "title": "Bob's blog" }));
}

- export function put() {
+ export function PUT() {
    return new Response(JSON.stringify({ "title": "Bob's blog" }));
}

- export function all() {
+ export function ALL() {
    return new Response(JSON.stringify({ "title": "Bob's blog" }));
}

// you can use the whole word "DELETE"
- export function del() {
+ export function DELETE() {
    return new Response(JSON.stringify({ "title": "Bob's blog" }));
}
```
````

File: packages/astro/src/runtime/server/endpoint.ts [hunk 4]
```diff
@@ -1,28 +1,55 @@
 import type { APIContext, EndpointHandler, Params } from '../../@types/astro';
+import { type LogOptions, warn, info } from '../../core/logger/core.js';
 
-function getHandlerFromModule(mod: EndpointHandler, method: string) {
+function getHandlerFromModule(mod: EndpointHandler, method: string, logging: LogOptions) {
+	const lowerCaseMethod = method.toLowerCase();
+
+	// NOTE: remove in Astro 4.0
```
delucis, line 7: Will we remember to look for `NOTE`? I’ve usually seen `TODO` used in the past.

packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -1,28 +1,55 @@
...
+	if (mod[lowerCaseMethod]) {
+		warn(
+			logging,
+			'astro',
+			`Lower case endpoint names are deprecated and will not be supported in Astro 4.0. Rename the endpoint ${lowerCaseMethod} to ${method}.`
```
delucis, line 12: Does this logging tell users which file/path they need to update?

.changeset/twelve-coats-rush.md
```diff
@@ -0,0 +1,35 @@
+---
+'astro': major
+---
+
+The endpoints that have **lowercase** names are **deprecated**.
+
+The endpoints methods **need** to be in **uppercase**.
```
delucis, line 7: I think we can rephrase this to emphasise the action users need to take, i.e. when thinking about a CHANGELOG as a guide that users need to act upon, we can focus on users _renaming_ functions rather than the passive state functions should have:

```suggestion
Lowercase names for endpoint functions are now deprecated.

Rename functions to their uppercase equivalent:
```

delucis (commented):
Thanks for this @ematipico!

A few quick comments, mainly focused on DX stuff like the CHANGELOG & deprecation logging.

packages/astro/src/runtime/server/endpoint.ts
ematipico: Unfortunately we don't have this information in that point of the code :(

PR #7785, opened 2023-07-24
Update `tsconfig.json` presets for TS 5.0

Princesseuh:
## Changes

TypeScript has this annoying behaviour where if you use an unknown compiler option, it'll.. completely refuse to do anything. I feel like they could've just added a warning instead, but whatever. This PR adds cool new options like `verbatimModuleSyntax` and updates it to use `moduleResolution: 'bundler'`, unlocking supports for the `exports` field. It's neat

## Testing

Tested manually

## Docs

N/A, we don't document this

### DIFF
--- a/packages/astro/tsconfigs/base.json
+++ b/packages/astro/tsconfigs/base.json
@@ -10,9 +10,13 @@
     "allowImportingTsExtensions": true,
     // Enable JSON imports.
     "resolveJsonModule": true,
+    // Enforce the usage of type-only imports when needed, which helps avoiding bundling issues.
+    "verbatimModuleSyntax": true,
+    // Ensure that each file can be transpiled without relying on other imports.
+    "isolatedModules": true,
     // Astro directly run TypeScript code, no transpilation needed.
     "noEmit": true,
-    // Report an error when importing a file using a casing different from the casing on disk.
+    // Report an error when importing a file using a casing different from another import of the same file.
     "forceConsistentCasingInFileNames": true,
     // Properly support importing CJS modules in ESM
     "esModuleInterop": true,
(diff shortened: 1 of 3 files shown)


packages/astro/tsconfigs/base.json
```diff
@@ -5,11 +5,11 @@
...
     "module": "ESNext",
     // Enable node-style module resolution, for things like npm package imports.
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
+    // Allow importing TypeScript files using their native extension (.ts(x)).
+    "allowImportingTsExtensions": true,
     // Enable JSON imports.
     "resolveJsonModule": true,
-    // Enable stricter transpilation for better output.
-    "isolatedModules": true,
```
lilnasy, line 12: Why is this no longer needed?

Princesseuh: You can't have `isolatedModules` and `verbatimModuleSyntax` on at the same time, but I guess since they're in separate preset we could enable it and disable it in strict, maybe. Or maybe `verbatimModuleSyntax` should be part of base anyways...

Princesseuh: Oh, it seems like they changed it https://github.com/microsoft/TypeScript/pull/53611

bluwy (approved):
Should this be merged to the `next` branch?

ematipico (approved):
Looks good to me, let's 🚢  it!

howagain:
I'm seeing a behavior on webstorm where `"allowImportingTsExtensions": true` is adding the `.tsx` [extension automatically](https://youtrack.jetbrains.com/issue/WEB-60917/Typescript-Use-file-extension-Never-is-not-respected-when-auto-importing-with-allowImportingTsExtensions-enabled). It seems like a bug on their side but why is it included in the base tsConfig for astro? I don't see any discussion over that decision.

Issue #7765, opened 2023-07-22
View Transition resets html attributes


florian-lefebvre:
### What version of `astro` are you using?

^2.9.1

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Brave

### Describe the Bug

I have a script that toggles the `.dark` class on the html element. However, using view transitions, the html class is reset to its original state (ie. class specified in the layout) but the script is not re-run.

To reproduce:
1. The page background will be red if your os is in dark mode
2. navigate to another page, the background will be white again

### What's the expected result?

I'd expect one of the following:

1. html attributes are not reset to default because scripts are not re-run
2. scripts re-run

I think 1. makes more sense to avoid flashes

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-8a4stc

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Replacing the HTML attribute is what you would expect with View Transitions MPA mode so that I think is what we are going for. I would expect a script to run, though. Let me look into that.

matthewp:
Oh right, these are module scripts that get bundled. So those will not necessarily rerun (maybe if they are inline...). So we'll need some solution for that.

One person had suggested firing the `load` event. I'll have to see if other similar routers do that. Another option is a namespaced `astro:load` or something like that. Then you could listen to this event and do what you need. Want to first research what other similar routers do, if you happen to know and can provide context that would be appreciated.

matthewp:
Turbo uses a `turbo:load` event: https://turbo.hotwired.dev/reference/events

florian-lefebvre:
Awesome thank you!

Tc-001:
Sadly this doesn't seem solve the theme toggle usecase unless I am missing something.

The current chain seems to be `Old page -> Fetch new page -> Start crossfade -> End crossfade -> Dispatch astro:load`

This means that the non-dark version of the page is still visible until the transition is completed

```astro
<script is:inline>
	document.addEventListener("astro:load", () => document.body.setAttribute("data-dark", "dark"))
</script>
```
[Screencast from 2023-07-25 23-07-40.webm](https://github.com/withastro/astro/assets/55956895/df1cd83a-cd89-40bc-8844-9759767bb441)

matthewp:
Ah good point, I think we need to run astro:load inside updateDOM before the transition is finished.

[...]

PR #7786, opened 2023-07-24
Execute scripts on page navigation (view transitions)

matthewp:
## Changes

- DOMParser creates scripts that are non executable, event after inserting into the page.
- This makes it so that scripts are always executed. If they have already run we reinsert them. That means for inline classic scripts those will re-run. Module scripts will not.
- This approach was chosen mostly because it's the last amount of code. It also aligns with what Swup does it.
- Adds the `astro:load` even that you can listen to if you need to rerun something on page navigation.
- Fixes https://github.com/withastro/astro/issues/7765

## Testing

- Test added to verify the `astro:load` event fire on forward navigation.
- Test added to verify `astro:load` fires on direct page navigation (MPA).

## Docs

- https://github.com/withastro/docs/pull/3810

### DIFF
--- a/.changeset/sharp-seas-smile.md
+++ b/.changeset/sharp-seas-smile.md
@@ -2,11 +2,11 @@
 'astro': patch
 ---
 
-Execute scripts when navigating to a new page
+Execute scripts when navigating to a new page.
 
-When navigating to an new page with client-side navigation, scripts are executed (and re-executed). This makes it so that any new scripts on the incoming page are run and DOM can be updated.
+When navigating to an new page with client-side navigation, scripts are executed (and re-executed) so that any new scripts on the incoming page are run and the DOM can be updated.
 
-Module scripts won't re-execute because they never do. To support cases where you want to modify the DOM this also adds a `astro:load` event that you can listen to:
+However, `type=module` scripts never re-execute in Astro, and will not do so in client-side routing. To support cases where you want to modify the DOM, a new `astro:load` event listener been added:
 
 ```js
 document.addEventListener('astro:load', () => {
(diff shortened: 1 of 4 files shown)


.changeset/sharp-seas-smile.md
```diff
@@ -0,0 +1,15 @@
+---
+'astro': patch
+---
+
+Execute scripts when navigating to a new page
+
+When navigating to an new page with client-side navigation, scripts are executed (and re-executed). This makes it so that any new scripts on the incoming page are run and DOM can be updated.
+
+Module scripts won't re-execute because they never do. To support cases where you want to modify the DOM this also adds a `astro:load` event that you can listen to:
```
sarah11918, line 9:
```suggestion
Execute scripts when navigating to a new page.

When navigating to an new page with client-side navigation, scripts are executed (and re-executed) so that any new scripts on the incoming page are run and the DOM can be updated.

However, `type=module` scripts never re-execute in Astro, and will not do so in client-side routing. To support cases where you want to modify the DOM, a new `astro:load` event listener been added:
```
Some tiny grammar tidying, plus some suggestions for concise wording (if accurate!) for your consideration!

sarah11918 (commented):
Quick comment on the changeset text, @matthewp, and just a quick question whether we should be adding `astro:load` as part of the docs PR (since I don't think there's anything existing there) that I'll also bring up there!

sarah11918 (approved):
Approved by docs!

doganalper:
I think this would solve problem I'm facing currently. I wanted to add view transitions but since I have dark/light theme it breaks logic I have and themes not working properly. Good work @matthewp!

PR #7787, opened 2023-07-24
Use `moduleResolution: 'node16'`

Princesseuh:
## Changes

**Blocked by https://github.com/vitejs/vite/pull/13947**

https://github.com/withastro/astro/pull/7780 made it so we use `moduleResolution: 'bundler'`, this is mostly fine but since we still require extensions in the runtime, it's better for us to use `moduleResokution: 'node16'` instead. However, this has, a lot of requirements. This PR attempts to do it

## Testing

Tests should pass!

## Docs

N/A


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.preferences.importModuleSpecifierEnding": "js"
```
bluwy, line 2: I just realized I set this per your recommendation but with `javascript.` 🤦 That's why it was not working.

bluwy (commented):
Besides the Vite issue, the changes looks good to me. Thanks for tackling this!

bluwy:
Not sure if Vite is the only blocker for node16, but it has been fixed and released in 4.4.7

Princesseuh:
Closing this, will have to restart from scratch pretty much

PR #7791, opened 2023-07-25
feat: pass logger to integrations

ematipico:
## Changes

In this PR, all the Astro hooks now have an additional object called "bag".

In this "bag", we can find a set of utility integrations that can use in that hook. Ideally, a bag should be the same for each hook. Because of that, I had to create an internal map of the loggers, otherwise, we risk creating an abysmal about of loggers: `n*m`m where `n` is the number of integrations and `m` is the number of hooks. 

I created two classes:
- `Logger`, basic class. The objective of this class is to be used internally in our codebase. Ideally, we will pass around an instance of this class instead of passing `logging`. I plan to make this change gradually in the future.
- `AstroIntegrationLogger`, similar to `Logger`, this class has the `label` stored inside of it. The main idea is that integrations should not know anything about labels.
	If an integration needs to do some logging with a different label - which I assume it's rare - they can use the `fork` function, which returns a new logger which has the same logging options. 

## Testing

Current tests should pass.

## Docs

N/A I will create the changeset in the `next` branch

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1857,86 +1857,69 @@ export interface AstroIntegration {
 	name: string;
 	/** The different hooks available to extend. */
 	hooks: {
-		'astro:config:setup'?: (
-			options: {
-				config: AstroConfig;
-				command: 'dev' | 'build' | 'preview';
-				isRestart: boolean;
-				updateConfig: (newConfig: Record<string, any>) => void;
-				addRenderer: (renderer: AstroRenderer) => void;
-				addWatchFile: (path: URL | string) => void;
-				injectScript: (stage: InjectedScriptStage, content: string) => void;
-				injectRoute: (injectRoute: InjectedRoute) => void;
-				addClientDirective: (directive: ClientDirectiveConfig) => void;
-				// TODO: Add support for `injectElement()` for full HTML element injection, not just scripts.
-				// This may require some refactoring of `scripts`, `styles`, and `links` into something
-				// more generalized. Consider the SSR use-case as well.
-				// injectElement: (stage: vite.HtmlTagDescriptor, element: string) => void;
-			},
-			bag: AstroIntegrationBag
-		) => void | Promise<void>;
-		'astro:config:done'?: (
-			options: {
-				config: AstroConfig;
-				setAdapter: (adapter: AstroAdapter) => void;
-			},
-			bag: AstroIntegrationBag
-		) => void | Promise<void>;
-		'astro:server:setup'?: (
-			options: { server: vite.ViteDevServer },
-			bag: AstroIntegrationBag
-		) => void | Promise<void>;
-		'astro:server:start'?: (
-			options: { address: AddressInfo },
-			bag: AstroIntegrationBag
-		) => void | Promise<void>;
-		'astro:server:done'?: (bag: AstroIntegrationBag) => void | Promise<void>;
-		'astro:build:ssr'?: (
-			options: {
-				manifest: SerializedSSRManifest;
-				/**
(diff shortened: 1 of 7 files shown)


delucis (commented):
Exciting @ematipico! Quick API question: what’s the motivation for exposing this in a second argument instead of adding it to the existing object hooks pass? It adds some complexity for users and I’m not sure I understand why it’s necessary or how to explain it.

ematipico:
> Exciting @ematipico! Quick API question: what’s the motivation for exposing this in a second argument instead of adding it to the existing object hooks pass? It adds some complexity for users and I’m not sure I understand why it’s necessary or how to explain it.

Thinking long term, the "bag" will contain a set of APIs provided by Astro at every hook. This bag will always be available at every hook, and it **doesn't change** regardless of the hook. Considering that `options` will change based on the hook, I thought it was semantically different and worth the distinction with a second parameter.

delucis:
This is my personal opinion, so curious to hear feedback from others, but I suspect this distinction — one object changes, the other doesn't — is an implementation detail that users don't really care about. The result is just: now I have to remember which of these two objects my hook receives has what I need.

With one object, I open the curly brace in my editor, start typing, get autocomplete, find what I want, done.

With two objects, I do that, start typing... Huh, not showing up, hmmm 🤔 ah, I think I read there's another object, maybe it's there?

I do think for documentation it's helpful to be able to tell people "here are things you'll _always_ get in your integration hooks" but we don't need a distinct object to do that.

matthewp:
I think I agree with @delucis here. I think the biggest downside to this implementation is that it makes destructing more difficult and that's often how we document using the options. I think documentation + types already helps figure out what options are always provided to the integration.

packages/astro/src/integrations/index.ts
```diff
@@ -38,6 +38,19 @@ async function withTakingALongTimeMsg<T>({
 	return result;
 }
 
+// Internally used to store instances of loggers.
+const Loggers = new Map<string, AstroIntegrationLogger>();
```
matthewp, line 42: I think this is being used to avoid creating a new AstroIntegrationLogger each time, is that right?

If so, this is leaking memory. Can you use `WeakMap<AstroIntegration, AstroIntegrationLogger>` instead?

ematipico: I wanted to use WeakMap in the first place, although it needs an object as key. I can use the whole integration object then

ematipico:
@delucis @matthewp addressed your feedback, please review again

packages/astro/src/@types/astro.ts
```diff
@@ -1874,10 +1876,17 @@ export interface AstroIntegration {
 		'astro:config:done'?: (options: {
 			config: AstroConfig;
 			setAdapter: (adapter: AstroAdapter) => void;
+			logger: AstroIntegrationLogger;
 		}) => void | Promise<void>;
```
bluwy, line 1880: I wonder if we can/should only expose the logger here. Then integrations can assign this to a variable outside of the hook and use it everywhere, as I think that's commonly needed too when abstracting functions.

ematipico: We would force the consumers to use this "trick" to save the logger instance. I prefer to give our users everything they need in the hooks.

It's a pity we don't have a high-order function to define integrations because we could pass the logger there.

bluwy: I think it's mostly me accustomed to Vite's flow, which is something like: 

```js
let resolvedConfig

return {
  name: 'vite-plugin',
  configResolved(config) {
    resolvedConfig = config
  },
  transform() {
    resolvedConfig.logger.warn('...')
  }
}
```

Not fully oppose if we're ok with exposing `logger` in most of the hooks though.

PR #7792, opened 2023-07-25
changelog for logger

ematipico:
## Changes

This is the changelog for https://github.com/withastro/astro/pull/7791

## Testing

N/A

## Docs


.changeset/twelve-baboons-obey.md
```diff
@@ -0,0 +1,20 @@
+---
+'astro': minor
+---
+
+Astro integrations now accept a logger. The logger is available to all hooks as 
+an additional parameter:
```
delucis, line 6: Quick review, but commented on #7791 that it might be preferable to include `logger` in the first argument if we can.

```suggestion
Integrations can now log messages using Astro’s built-in logger.
The logger is available to all hooks as an additional parameter:
```

bluwy:
I don't quite understand this flow of making the changes in `main`, and submitting the changesets in `next`. Wouldn't the logic in `main` already be released if we need to release something? It becomes undocumented until in 3.0.

But since this is also non-breaking, I think we can cut this in v2?

Issue #7793, opened 2023-07-25
astro add cli doesn't inherit flags


Elod-T:
### What version of `astro` are you using?

2.9.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

With the `astro add` cli I you can install astro integrations and automatically add them to the config file. There are integrations that are only needed as devDependency but the cli doesn't provide a way to install them as one. ( For example `astro-compress`, `astro-preload` or `@astrojs/sitemap` )

![astro-add](https://github.com/withastro/astro/assets/33983644/0d2ab5c6-ad32-44b5-b7f6-1e46c3240cd4)

### What's the expected result?

I would expect that the arguments get passed down to the install command.

Example:

![2023-07-25_16-50](https://github.com/withastro/astro/assets/33983644/f6d12933-b068-44df-8153-6bc247e2dc1d)

### Link to Minimal Reproducible Example

any astro project

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Thanks for opening an issue! I see that you already opened a PR, so we can continue the discussion there.

PR #7794, opened 2023-07-25
Fixes withastro/astro#7793 astro add cli pass down arguments to install cmd 

Elod-T:
## Changes

- get arguments from the `astro add` cli
- pass them down to the actual package manager

## Testing

I tested it locally, running `astro add astro-compress -D` ran `pnpm add astro-compress -D true`

## Docs

I think it should be in the docs

### DIFF
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -641,13 +641,22 @@ async function tryToInstallIntegrations({
 }): Promise<UpdateResult> {
 	const installCommand = await getInstallIntegrationsCommand({ integrations, cwd });
 
+	const inheritedFlags = Object.entries(flags)
+		.map(([flag, value]) => {
+			if (flag == '_') return;
+
+			return [`-${flag}`, value.toString()];
+		})
+		.filter(Boolean)
+		.flat();
+
 	if (installCommand === null) {
 		return UpdateResult.none;
 	} else {
 		const coloredOutput = `${bold(installCommand.pm)} ${installCommand.command}${[
 			'',
 			...installCommand.flags,
-		].join(' ')} ${cyan(installCommand.dependencies.join(' '))}`;
+		].join(' ')} ${cyan(installCommand.dependencies.join(' '))} ${inheritedFlags.join(' ')}`;
 		const message = `\n${boxen(coloredOutput, {
 			margin: 0.5,
 			padding: 0.5,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18666,7 +18666,6 @@ packages:
     resolution: {directory: packages/astro, type: directory}
     id: file:packages/astro
     name: astro
-    version: 2.9.2
     engines: {node: '>=16.12.0', npm: '>=6.14.0'}
     hasBin: true
     peerDependencies:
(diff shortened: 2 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -18666,7 +18666,6 @@ packages:
     resolution: {directory: packages/astro, type: directory}
```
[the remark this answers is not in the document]
bluwy: I think it's caused by the latest pnpm version

packages/astro/src/cli/add/index.ts
```diff
@@ -641,13 +641,22 @@ async function tryToInstallIntegrations({
...
 	const installCommand = await getInstallIntegrationsCommand({ integrations, cwd });
 
+	const inheritedFlags = Object.entries(flags)
+		.map(([flag, value]) => {
+			if (flag == '_') return;
+
+			return [`-${flag}`, value.toString()];
+		})
+		.filter(Boolean)
+		.flat();
```
bluwy, line 651: I don't think we should pass everything down. To be safe, we can scope to `--save-dev`, `-D` etc only. In case we support more options in the future, this part won't be glossed over.

Elod-T:
@ematipico Yes I did test it locally, sorry I wasn't clear enough, I meant no unit tests. I'll make the changelog tomorrow

packages/astro/src/cli/add/index.ts
Elod-T: I don't get why we should only pass those down. It's essentially giving back the normal functionality with flags to the cli. It will work as pnpm without the `astro add` wrapper. Please let me know if I am missing something

bluwy: I think `astro add` shouldn't be seen as `pnpm add` but smarter. It does a lot more like editing configs, adding integration-specific configs etc. The installation phase is one of the steps for `astro add`. So passing the args directly to that single step is a bit off to me.

ematipico:
> @ematipico Yes I did test it locally, sorry I wasn't clear enough, I meant no unit tests. I'll make the changelog tomorrow

Could you please update the template, then? It's important to state that you tested the feature. Even a manual test is important.

packages/astro/src/cli/add/index.ts
natemoo-re: I agree with @bluwy that we should only support `--save-dev` and `-D` for now.

natemoo-re (changes requested):
Code looks good but I'd also like to reduce the scope of this change to only allow `-D` and `--save-dev` for now.

Elod-T:
I'm working on the pr, and I'm thinking that we could allow all flags supported by `npm install`.
```
-P, --save-prod: Package will appear in your dependencies. This is the default unless -D or -O are present.

-D, --save-dev: Package will appear in your devDependencies.

-O, --save-optional: Package will appear in your optionalDependencies.

--no-save: Prevents saving to dependencies.

When using any of the above options to save dependencies to your package.json, there are two additional, optional flags:

-E, --save-exact: Saved dependencies will be configured with an exact version rather than using npm's default semver range operator.

-B, --save-bundle: Saved dependencies will also be added to your bundleDependencies list.
```
[Source](https://docs.npmjs.com/cli/v8/commands/npm-install#description)

natemoo-re:
Closing in favor of https://github.com/withastro/astro/pull/8032 so we can get this merged! Unfortunately I wasn't able to push to your fork so I had to push your commits to one of our branches.

PR #7798, opened 2023-07-25
[docs] Markdoc README syntax fixes

sarah11918:
## Changes

Fixes wonky rendering in docs because of the `:::caution` syntax. 

Replaces with a GitHub caution. Also moves the caution below so it's not between code sample and its explanation.

## Testing

No tests

## Docs

Only Docs!

### DIFF
--- a/packages/integrations/markdoc/README.md
+++ b/packages/integrations/markdoc/README.md
@@ -424,11 +424,10 @@ export default defineConfig({
 });
 ```
 
-:::caution
-When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible.
-
-Ensure that any HTML markup comes from trusted sources.
-:::
+> **Caution**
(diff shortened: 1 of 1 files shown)


packages/integrations/markdoc/README.md
```diff
@@ -429,6 +424,12 @@ export default defineConfig({
 });
 ```
 
+:::caution
+When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible.
+
+Ensure that any HTML markup comes from trusted sources.
+:::
```
sarah11918, line 431:
```suggestion
> **Caution**
> When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible.
>
> Ensure that any HTML markup comes from trusted sources.
```

Yan-Thomas (approved):
LGTM!

PR #7799, opened 2023-07-25
[docs] Markdoc README

sarah11918:
## Changes

Tries again to invoke the correct GitHub aside syntax

## Testing

No tests.
## Docs

Only docs!

### DIFF
--- a/packages/integrations/markdoc/README.md
+++ b/packages/integrations/markdoc/README.md
@@ -424,7 +424,7 @@ export default defineConfig({
 });
 ```
 
-> **Caution**
+> **Warning**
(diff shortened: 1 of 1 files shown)


Yan-Thomas (approved):
Taking advantage of GitHub's preview, 100% sure we are *in fact* right this time, sorry for forgetting this in the last one.

> **Warning**
> When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible.
>
> Ensure that any HTML markup comes from trusted sources.

PR #7800, opened 2023-07-25
Scroll position restoration

matthewp:
## Changes

- This implements scroll position restoration in the ViewTransitions router.
- State is preserved in `history.state`
  - when clicking a link we add the previous page's position
  - when a scroll event occurs we record the state
  - When the user clicks back we use the state to set `scrollTo`
  - When the user clicks forward we use the state to set `scrollTo`

## Testing

- Back button test added
- Forward button test added

## Docs

N/A, transparent to developer


packages/astro/components/ViewTransitions.astro
```diff
@@ -194,5 +207,14 @@ const { fallback = 'animate' } = Astro.props as Props;
...
 		addEventListener('load', onload);
+		// There's not a good way to record scroll position before a back button.
+		// So the way we do it is by listening to scroll and just continuously recording it.
+		addEventListener('scroll', () => {
+			const state: State | undefined = history.state;
+			if(state) {
+				state.scrollY = scrollY;
+				history.replaceState(state, '');
+			}
+		}, { passive: true });
```
bluwy, line 218: Even though with passive, I wonder if there's a perf penalty calling `replaceState` repeatedly. Maybe it can use a throttle.

Could keep an eye on this and leave it as is too now though.

bluwy (approved):

packages/astro/components/ViewTransitions.astro
ematipico: I agree with Bjorn, but I feel more strongly on this one. I think **we have** to add a bouncing mechanism.

matthewp: If it were a regular Map would you still feel this way? I would assume that `replaceState` is just a map under the hood. I don't want to add extra code, increasing the bundle size, unless we know that we need to.

matthewp: Given that this is an experimental feature I think it's ok to be less safe in this case. If it is a problem we'll hear about it and fix before it becomes stable.

matthewp: Did some investigation and Safari considers calling replaceState too many times a security issue. I think I understand why; malicious sites could call pushState a lot to prevent you from backing out of the site. But I don't see why that would be the case for replaceState.

In any event, will need to look for another solution here.

bluwy: I remember very vaguely hitting this issue, and was able to fix it with a `throttle` helper with 300ms timeout (i think). This SO post seems to explain it: https://stackoverflow.com/questions/38419867/os-x-ios-safari-history-replacestate-limit-throws-securityerror-dom-exception

matthewp: Yeah, that's what I wound up doing. I did 500ms to be safe, but can bump it down. Otherwise ready for another review.

Akxe:
This has a 2 bugs:

## In prod - back animation no longer plays
1. Open https://gingerarchitecture.cz/
2. Click the first box
3. Navigate back
4. Observe: **no animation back**

## In dev (and probably in prod once the previous bug is fixed) - 
1. Clone and start this repository: https://github.com/Akxe/gingerarchitecture.cz
2. Scroll down
3. Click the last box
4. Navigate back
4. Navigate forward
6. Observe: **`scrollTop = 0` is applied to the old page causing the animated image to "fly" from the bottom of the page to the top**

millette:
From what I can see in prod, if I follow a link from the footer, the page opens at the footer instead of the top. Is this expected? Can it be overriden?

Akxe:
@millette I do not see that, but going back and forth using navigation does seem to break the page... sometimes... this is worrying.

millette:
For example, if you click `Français` on https://bus.waglo.com/en/ you'll stay at the top, but if you go the to bottom of the page and click one of the footer links, the page stays at the bottom.

Issue #7731, opened 2023-07-20
Astro redirects should not generate HTML stubs when using an adapter


mashehu:
### What version of `astro` are you using?

2.8.5

### Are you using an SSR adapter? If so, which one?

Netlify

### What package manager are you using?

npm

### What operating system are you using?

MacOS

### What browser are you using?

Chrome, Firefox

### Describe the Bug

When using configured redirects on netlify I see the placeholder HTML with "Redirecting to: " for a moment, before the browser is redirected to the correct site. 

### What's the expected result?

Based on the wording in the documentation:
> When running astro build, Astro will output HTML files with the [meta refresh](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#examples) tag by default. Supported adapters will instead write out the host’s configuration file with the redirects.

 I assumed that the netlify adapter, would only write to the `_redirects` file, without generating these placeholder files, which give a quick flash of an unstyled page between pages.

See my minimal example deployment: https://main--thriving-dusk-04bc5b.netlify.app/

### Link to Minimal Reproducible Example

https://github.com/mashehu/astro-quickstart/tree/main

### Participation

- [ ] I am willing to submit a pull request for this issue.

jdozierezell:
I just dropped by to create the same issue. This began appearing after a recent update of the adapter and was not present prior to that.

natemoo-re:
Ah yep, this behavior became more noticable when https://github.com/withastro/astro/pull/7644 was merged, but I suspect this has never worked properly in `hybrid` mode.

We should be bypassing these files in `hybrid` mode as well and only generate the `_redirects` file so Netlify can handle it.

natemoo-re:
This is also relevant for the Vercel and Node adapters

fprl:
Before creating a new issue:

I am trying to implement redirects but its not working in prod (in dev all good):
```javascript
redirects: {
  '/my-account/my-bookings/[hash]': '/nl/my-account/bookings/[hash]',
  '/my-account/my-profile': '/nl/my-account/profile',
  '/en/my-account/my-bookings/[hash]': '/my-account/bookings/[hash]',
  '/en/my-account/my-profile': '/my-account/profile',
  },
```
Everything gets redirected to `/`. Maybe this feature is for static builds? I'm using node adapter (I also see the placeholder).

Is this related to this PR? Thanks

matthewp:
[...]

PR #7805, opened 2023-07-25
Prevent building .html files in hybrid mode

matthewp:
## Changes

- In Netlify, do not build .html files because it prioritizes those
- This only is an issue with hybrid mode
- Fixes: https://github.com/withastro/astro/issues/7731

## Testing

- Test case added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/ninety-kids-fail.md
+++ b/.changeset/ninety-kids-fail.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Prevent building .html file redirects in hybrid mode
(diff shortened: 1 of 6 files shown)


natemoo-re:
Isn’t this also relevant for Vercel and Node? IMO any adapter in hybrid mode should use server redirect behavior.

bluwy (approved):

ematipico (approved):

matthewp:
Yeah, it might need it in both of those too. I'll do those in a follow-up.

PR #7809, opened 2023-07-26
Refactor dev CLI flow

bluwy:
## Changes

**Note: this merges into the js-api branch**

Refactor the `dev` command CLI flow

1. Remove `openConfig` usage, switch to `resolveConfig` instead
2. Make a somewhat cleaner `dev()` JS API (now accepts `AstroInlineConfig`)
3. Update tests with new signature.

Next will be making the same changes as no2 for all other commands, which should be easier than `dev`.

## Testing

Existing tests should pass. I also tested the dev command locally, especially restarts when editing `astro.config.mjs`.

## Docs

n/a. internal change for now. will document more when we expose a JS API.

### DIFF
--- a/packages/astro/src/cli/dev/index.ts
+++ b/packages/astro/src/cli/dev/index.ts
@@ -1,9 +1,10 @@
-import fs from 'node:fs';
+import { cyan } from 'kleur/colors';
 import type yargs from 'yargs-parser';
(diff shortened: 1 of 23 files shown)


ematipico (commented):
Way better now!

ematipico (approved):
Way better now!

PR #7811, opened 2023-07-26
Make build/preview/sync API accept AstroInlineConfig

bluwy:
## Changes

**Note: This PR merges into the js-api branch**

- Make `build`/`preview`/`sync` API accept AstroInlineConfig.
- Mostly moving code around without logic change.

The JS API is almost done! This finishes the APIs above besides `check` and `add` as those will be refactored in 3.0. Two more things to clean up in later PRs:

1. Initial config loading error handling can be improved.
2. `logging` can be tucked into `settings` so the user don't have to pass that in.

## Testing

Existing tests should pass. Also tested locally for the commands to all work.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/core/sync/index.ts
+++ b/packages/astro/src/core/sync/index.ts
@@ -35,13 +35,7 @@ export async function sync(
 	const { userConfig, astroConfig } = await resolveConfig(inlineConfig ?? {}, 'sync');
 	telemetry.record(eventCliSession('sync', userConfig));
 
-	const _settings = createSettings(astroConfig, 'dev', fileURLToPath(astroConfig.root));
-
-	const settings = await runHookConfigSetup({
-		settings: _settings,
-		logging: options.logging,
-		command: 'build',
-	});
+	const settings = createSettings(astroConfig, 'dev', fileURLToPath(astroConfig.root));
 
 	return await syncInternal(settings, options);
 }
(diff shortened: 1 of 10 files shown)


ematipico:
> Make build/preview/sync API accept AstroInlineConfig.

Will these be the official APIs? Just trying to understand where we are going :)

packages/astro/src/core/sync/index.ts
```diff
@@ -1,62 +1,66 @@
...
+	 * Only used for testing
+	 * @internal
+	 */
+	fs?: typeof fsMod;
 };
 
-export async function syncCli(
-	settings: AstroSettings,
-	{ logging, fs, flags }: { logging: LogOptions; fs: typeof fsMod; flags?: Arguments }
+export async function sync(
```
ematipico, line 31: Will this be the official API? Just trying to understand where we are going

ematipico (approved):

packages/astro/src/core/sync/index.ts
bluwy: Yes I think it would be nice to make this public. I also noticed the PR description had a typo. It's besides `check` and `add` that I had not refactored, not `sync` and `add` 😅 Will update that.

ematipico: Perfect, thank you. I am excited to see the final result!

PR #7815, opened 2023-07-26
feat: astro features

ematipico:
## Changes

Introducing **Astro features**. Astro features are features that astro has out the box, and sometimes users can opt-in. 
Few examples:
- `staticOutput` if a user uses a specific adapter, the adapter needs to be able to serve those static assets. An example is Cloudflare, which **doesn't support it**.
- `assets`: not supported by those adapters that are incompatible with node
 
___

This PR adds a new API that adapters can use when registering one. It's called `supportedFeatures`, and it's a new way to tell Astro if and when an adapter is able to support certain features that are enabled via configuration. 

The adapter can also tell Astro the kind of support:
- stable
- experimental
- unsupported
- deprecated

For now, this new API/object is not mandatory, but it will be in Astro 4.0.

The PR also updates the adapters that we maintain, and it adds the feature map to them.

### About `assets`

There's a small caveat for the assets. Astro assets is a feature that is already turned on by default, meaning that Astro sets already a default service. This means that adapters that **don't support** assets will always get an error, which is inconvenient. I came up with a solution where we set the default **after** the adapter validation, although I don't like the solution very much, and I wonder if there's a better way to handle a case like this.

The assets feature is just an example; we might have the same problem with future features like this. What do you suggest?

## Testing

I added new unit test cases and updated the existing ones.

## Docs

I will create the changelog in another PR

### DIFF
--- a/.changeset/dirty-lies-cover.md
+++ b/.changeset/dirty-lies-cover.md
@@ -1,0 +1,32 @@
+---
+'@astrojs/cloudflare': minor
+'@astrojs/netlify': minor
+'@astrojs/vercel': minor
+'@astrojs/deno': minor
+'@astrojs/node': minor
+'astro': minor
+---
+
+Introduced the concept of feature map. A feature map is a list of features that are built-in in Astro, and an Adapter
+can tell Astro if it can support it.
+
+```ts
+import {AstroIntegration} from "./astro";
+
+function myIntegration(): AstroIntegration {
+    return {
+        name: 'astro-awesome-list',
+        // new feature map
+        supportedAstroFeatures: {
+            hybridOutput: 'experimental',
+            staticOutput: 'stable',
+            serverOutput: 'stable',
+            assets: {
+                supportKind: 'stable',
+                isSharpCompatible: false,
+                isSquooshCompatible: false,
+            },
+        }
+    }
+}
+```
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1669,6 +1669,39 @@ export type PaginateFunction = (data: any[], args?: PaginateOptions) => GetStati
 
 export type Params = Record<string, string | undefined>;
 
+export type SupportsKind = 'unsupported' | 'stable' | 'experimental' | 'deprecated';
+
+export type AstroFeatureMap = {
+	/**
+	 * The adapter is able serve static pages
+	 */
+	staticOutput?: SupportsKind;
+	/**
+	 * The adapter is able to serve pages that are static or rendered via server
+	 */
+	hybridOutput?: SupportsKind;
+	/**
+	 * The adapter is able to serve SSR pages
+	 */
+	serverOutput?: SupportsKind;
+	/**
+	 * The adapter can emit static assets
+	 */
+	assets?: AstroAssetsFeature;
+};
+
+export interface AstroAssetsFeature {
+	supportKind?: SupportsKind;
+	/**
+	 * Whether if this adapter deploys files in an enviroment that is compatible with the library `sharp`
+	 */
+	isSharpCompatible?: boolean;
+	/**
+	 * Whether if this adapter deploys files in an enviroment that is compatible with the library `squoosh`
+	 */
+	isSquooshCompatible?: boolean;
+}
+
 export interface AstroAdapter {
 	name: string;
 	serverEntrypoint?: string;
@@ -1676,6 +1709,12 @@ export interface AstroAdapter {
 	exports?: string[];
 	args?: any;
 	adapterFeatures?: AstroAdapterFeatures;
+	/**
+	 * List of features supported by an adapter.
+	 *
+	 * If the adapter is not able to handle certain configurations, Astro will throw an error.
+	 */
+	supportedAstroFeatures?: AstroFeatureMap;
 }
 
 type Body = string;
--- a/packages/astro/src/integrations/index.ts
+++ b/packages/astro/src/integrations/index.ts
@@ -18,7 +18,7 @@ import type { SerializedSSRManifest } from '../core/app/types';
 import type { PageBuildData } from '../core/build/types';
 import { buildClientDirectiveEntrypoint } from '../core/client-directive/index.js';
 import { mergeConfig } from '../core/config/index.js';
-import { info, type LogOptions, AstroIntegrationLogger } from '../core/logger/core.js';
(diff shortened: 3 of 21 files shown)


bluwy (commented):
The general validation logic looks great! Got some questions/comments about some things

packages/astro/src/integrations/index.ts
```diff
@@ -178,13 +186,183 @@ export async function runHookConfigDone({
...
 				logging,
 			});
 		}
 	}
+	// Astro doesn't set a default on purpose when loading the configuration file.
+	// Not setting a default allows to check if the adapter doesn't support the feature in case the user overrides
+	// Astro's defaults.
+	//
+	// At the end of the validation, we push the correct defaults.
+	if (typeof settings.config.image === 'undefined') {
```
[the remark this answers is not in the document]
bluwy: I'm actually not sure if we can get the adapter config from here. It might be too late 🤔

packages/astro/src/@types/astro.ts
```diff
@@ -1660,12 +1660,47 @@ export type PaginateFunction = (data: any[], args?: PaginateOptions) => GetStati
...
+	 */
+	assets?: {
+		supportKind?: SupportsKind;
+		/**
+		 * Whether if this adapter deploys files in an enviroment that is Node.js compatible.
+		 *
+		 * The default services used by Astro needs Node.js.
+		 */
+		isNodeCompatible?: boolean;
+	};
```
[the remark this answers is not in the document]
Princesseuh: Hmm, definitely not for `Squoosh`, especially since we plan on removing it. But for Sharp I guess it could be a thing? It does have native deps, albeit nowadays it kinda supports everything and it'll hopefully support WASM soon so it's unlikely for a platform to have a specific issue with it anymore (even Stackblitz and other web envs have it working now)

packages/astro/src/integrations/index.ts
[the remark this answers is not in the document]
ematipico: Oh, shoot, that's old code that I have to remove :D good catch!

ematipico:
I updated the code:
- the validation doesn't throw an error, but it **logs** an error. This was making testing impossible.
- I removed the hacks I needed for deno tests
- @bluwy, I updated the code based on your suggestion; please check again
- @Princesseuh, could you please check the **adapters** and make sure the values are correct?

packages/integrations/vercel/src/edge/adapter.ts
```diff
@@ -27,6 +27,15 @@ function getAdapter(): AstroAdapter {
...
 		serverEntrypoint: `${PACKAGE_NAME}/entrypoint`,
 		exports: ['default'],
+		supportedAstroFeatures: {
+			hybridOutput: 'Stable',
+			staticOutput: 'Stable',
+			serverOutput: 'Stable',
+			assets: {
+				supportKind: 'Unsupported',
+				isSharpCompatible: true,
+			},
```
Princesseuh, line 37: On Vercel Edge, the only image service supported is Vercel's (when using `vercel({imageService: true})`)

.changeset/dirty-lies-cover.md
```diff
@@ -0,0 +1,31 @@
...
+        name: 'astro-awesome-list',
+        // new feature map
+        supportedFeatures: {
+            hybridOutput: 'Experimental',
+            staticOutput: 'Stable',
+            serverOutput: 'Stable',
+            assets: {
+                supportKind: 'Unsupported',
+                isNodeCompatible: false,
+            },
```
Princesseuh, line 27: Outdated?

packages/integrations/vercel/src/edge/adapter.ts
Princesseuh:
```suggestion
			assets: {
				supportKind: 'Stable',
				isSharpCompatible: true,
				isSquooshCompatible: false,
			},
```

packages/integrations/cloudflare/src/index.ts
```diff
@@ -24,11 +24,31 @@ export function getAdapter(isModeDirectory: boolean): AstroAdapter {
...
 				serverEntrypoint: '@astrojs/cloudflare/server.advanced.js',
 				exports: ['default'],
+				supportedAstroFeatures: {
+					hybridOutput: 'Stable',
+					staticOutput: 'Unsupported',
+					serverOutput: 'Stable',
+					assets: {
+						supportKind: 'Unsupported',
+						isSharpCompatible: false,
+						isSquooshCompatible: false,
```
Princesseuh, line 49:
```suggestion
						supportKind: 'Stable',
						isSharpCompatible: false,
						isSquooshCompatible: false,
```

Every adapter supports `assets` the feature, what they don't support is the image services. Someone could write an image service powered by a supported library (or Cloudflare's Image API in this case)

packages/integrations/netlify/src/integration-edge-functions.ts
```diff
@@ -11,6 +11,16 @@ export function getAdapter(): AstroAdapter {
 		name: '@astrojs/netlify/edge-functions',
 		serverEntrypoint: '@astrojs/netlify/netlify-edge-functions.js',
 		exports: ['default'],
+		supportedAstroFeatures: {
+			hybridOutput: 'Stable',
+			staticOutput: 'Stable',
+			serverOutput: 'Stable',
+			assets: {
+				supportKind: 'Unsupported',
```
Princesseuh, line 19:
```suggestion
				supportKind: 'Stable',
```

Same as Cloudflare, every adapter supports `assets`, it's the services they don't support

matthewp:
Should the values be lower-case? When we have string flags like this elsewhere in the code I'm pretty sure they are lower-case.

packages/astro/src/integrations/index.ts
```diff
@@ -178,6 +179,31 @@ export async function runHookConfigDone({
 								`Integration "${integration.name}" conflicts with "${settings.adapter.name}". You can only configure one deployment integration.`
 							);
 						}
+						if (!adapter.supportedAstroFeatures) {
+							// NOTE: throw an error in Astro 4.0
+							warn(
+								logging,
+								'astro',
+								`The adapter ${adapter.name} doesn't provide a feature map. From Astro 3.0, an adapter 
+		can provide a feature map. Not providing a feature map will cause an error in Astro 4.0.`
```
bluwy, line 188: nit: I think the string should be a single line so it can be line-wrapped in the terminal naturally.

bluwy (approved):

PR #7816, opened 2023-07-26
feat: pass logger to integrations

ematipico:
## Changes

I created two classes:
- `Logger`, basic class. The objective of this class is to be used internally in our codebase. Ideally, we will pass around an instance of this class instead of passing `logging`. I plan to make this change gradually in the future.
- `AstroIntegrationLogger`, similar to `Logger`, this class has the `label` stored inside of it. The main idea is that integrations should not know anything about labels.
	If an integration needs to do some logging with a different label - which I assume it's rare - they can use the `fork` function, which returns a new logger which has the same logging options. 

## Testing

Current tests should pass.

## Docs

### DIFF
--- a/.changeset/odd-books-live.md
+++ b/.changeset/odd-books-live.md
@@ -1,0 +1,23 @@
+---
+'astro': minor
+---
+
+Integrations can now log messages using Astro’s built-in logger.
+
+The logger is available to all hooks as an additional parameter:
+
+```ts
+import {AstroIntegration} from "./astro";
+
+// integration.js
+export function myIntegration(): AstroIntegration {
+    return {
+        name: "my-integration",
+        hooks: {
+            "astro:config:done": ({ logger }) => {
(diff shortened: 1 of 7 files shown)


delucis (commented):
Changelog and public API look good to me! Thanks @ematipico 💜

matthewp:
The diff is showing all changes unrelated to a logger. Like the capitalization change to API routes. Is that a mistake?

bluwy (approved):
I'm still somewhat on the side that we should only expose the logger in `astro:config:done`, but if the rest are fine with exposing it in more hooks, I'm fine with it too 👍

jlarmstrongiv:
Thank you @ematipico !  I just updated my integration to use the new `AstroIntegrationLogger` in https://github.com/jlarmstrongiv/astro-i18n-aut/commit/f7468c1fac43ea12ef77a65e9e6d534d00995c0f 🎉

PR #7817, opened 2023-07-26
[docs] Markdoc README problematic aside notation

sarah11918:
## Changes

This takes yet another stab at making an aside using GitHub notation that imports properly into Docs.

## Testing

This tested my patience.

## Docs

Only docs!


Yan-Thomas (approved):
Tested this locally, it's working, if this doesn't work I'm burning this cursed README.
image

PR #7821, opened 2023-07-26
Fix astro/app import

ottomated:
## Changes

- Fixes a regression with importing `'astro/app'`

## Testing

- Added a test that imports astro/app, confirmed that it fails in the case of missing a `.js` in the imports

## Docs

N/A


natemoo-re (approved):
Thanks a lot for the PR!

PR #7829, opened 2023-07-26
fix(assets): Fix endpoint not being injected when an integration would enable `experimental.assets`

Princesseuh:
## Changes

What the title says. We'd only check for `experimental.assets` before running the integration's config setup hooks, so we'd miss it if an integration would enable it for the user (such as Starlight

## Testing

Tested manually, we'll be removing this in 3.0 anyway, so

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/build/index.ts
+++ b/packages/astro/src/core/build/index.ts
@@ -4,12 +4,14 @@ import { performance } from 'node:perf_hooks';
 import type * as vite from 'vite';
 import type yargs from 'yargs-parser';
 import type { AstroConfig, AstroSettings, ManifestData, RuntimeMode } from '../../@types/astro';
+import { injectImageEndpoint } from '../../assets/internal.js';
 import {
 	runHookBuildDone,
 	runHookBuildStart,
 	runHookConfigDone,
 	runHookConfigSetup,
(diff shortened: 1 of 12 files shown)


packages/astro/src/core/build/index.ts
```diff
@@ -89,6 +91,13 @@ class AstroBuilder {
 			command: 'build',
 			logging,
 		});
+
+		// HACK: Since we only inject the endpoint if `experimental.assets` is on and it's possible for an integration to
```
delucis, line 95: Is it worth a “Remove in 3.0” comment or would we pick it up anyway because `experimental.assets` will go away so this would be a type error?

delucis (approved):
Thanks Erika! LGTM. Plus I trust you.

packages/astro/src/core/build/index.ts
Princesseuh: In 3.0 it's likely that we'll need an option to disable the endpoint (ex: maybe you have your own, maybe you don't want any `assets` stuff), so we'll still need to check for something in the config still here. So for now I'll leave it like this and update whenever we'll add that option

bluwy (approved):

PR #7838, opened 2023-07-27
Fix head propagation for MDX components

bluwy:
## Changes

Revert this part https://github.com/withastro/astro/pull/7782#discussion_r1272460169

Fixes MDX fail (Presumably also [fixes markdoc](https://github.com/withastro/astro/blob/73eb4dfe2f1ed8bceeba6fb57d60c583f61d78b3/packages/integrations/markdoc/components/TreeNode.ts#L94-L102) but it doesn't seem to have tests cover head propagation)

I'm actually not sure how MDX components can get their head propagated, because it involves attaching themselves to `result._metadata.propagators` (which I didn't see any for MDX specifically). But reverting this part for now fix it.

## Testing

Ran MDX and markdoc test locally. Others tests should also pass.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/runtime/server/render/astro/instance.ts
+++ b/packages/astro/src/runtime/server/render/astro/instance.ts
@@ -37,6 +37,7 @@ export class AstroComponentInstance {
 	}
 
 	async init(result: SSRResult) {
+		if (this.returnValue !== undefined) return this.returnValue;
 		this.returnValue = this.factory(result, this.props, this.slotValues);
 		return this.returnValue;
 	}
@@ -72,7 +73,7 @@ function validateComponentProps(props: any, displayName: string) {
 	}
 }
 
-export async function createAstroComponentInstance(
+export function createAstroComponentInstance(
 	result: SSRResult,
 	displayName: string,
 	factory: AstroComponentFactory,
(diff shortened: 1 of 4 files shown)


packages/astro/src/runtime/server/render/astro/instance.ts
```diff
@@ -37,6 +37,7 @@ export class AstroComponentInstance {
 	}
 
 	async init(result: SSRResult) {
+		if (this.returnValue !== undefined) return this.returnValue;
```
bluwy, line 40: This part makes sure `this.returnValue` is cached. Otherwise we're calling `render()` then `init()`, which causes double head rendering.

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -144,3 +148,20 @@ async function callComponentAsTemplateResultOrResponse(
...
+		if (done) {
+			break;
+		}
+		// Call component instances that might have head content to be propagated up.
+		const returnValue = await value.init(result);
+		if (isHeadAndContent(returnValue)) {
+			result._metadata.extraHead.push(returnValue.head);
+		}
+	}
+}
```
bluwy, line 167: This code is what we had before I did the refactor, so no changes made here.

ematipico (approved):
Looks good. There are conflicts to solve

bluwy:
Yup, the conflicts should go away after I revert https://github.com/withastro/astro/pull/7843

bluwy:
Nevermind, I can't seem to revert that PR now. Will manually do it in this PR, then merge.

PR #7839, opened 2023-07-27
feat: adapter features, deprecate astro configs

ematipico:
## Changes

This PR deprecates `build.split` and `build.excludeMiddleware`.

After an internal discussion, we realised that these two config values should not live inside Astro, because they are only usable only with an adapter that can understand it

Because of that, we introduced the concept of `adapterFeatures`. These features are toggles that signal Astro to change the output of the build, and then the adapter needs to do something with those. 

## Testing

I updated the current test to use the new configuration. The tests should pass.

## Docs

### DIFF
--- a/.changeset/fair-emus-divide.md
+++ b/.changeset/fair-emus-divide.md
@@ -1,0 +1,39 @@
+---
+'astro': major
+'@astrojs/netlify': minor
+---
+
+The configuration `build.split` and `build.excludeMiddleware` are deprecated.
+
+Configuration that were inside the astro configuration, are now moved inside the adapter:
+
+```diff
+import {defineConfig} from "astro/config";
+import netlify from "@astrojs/netlify/functions";
+
+export default defineConfig({
+-    build: {
+-        excludeMiddleware: true
+-    },
+-    adapter: netlify()
++    adapter: netlify({
++        edgeMiddleware: true
++    })
+})
+```
+
+```diff
+import {defineConfig} from "astro/config";
+import vercel from "@astrojs/vercel/serverless";
+
+export default defineConfig({
+-    build: {
+-        split: true
+-    },
+-    adapter: netlify()
++    adapter: netlify({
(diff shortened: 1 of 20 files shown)


.changeset/fair-emus-divide.md
```diff
@@ -0,0 +1,39 @@
...
+export default defineConfig({
+-    build: {
+-        split: true
+-    },
+-    adapter: netlify()
++    adapter: netlify({
++        functionPerRoute: true
++    })
+})
+```
```
bluwy, line 38: I think we only need one diff per changeset for netlify and vercel? Looks like it's duplicated here.

bluwy (approved):
Thanks for making the changes! This looks great to me.

Should we also validate if the deprecated configs are used in runtime, somewhere around here?

https://github.com/withastro/astro/blob/73eb4dfe2f1ed8bceeba6fb57d60c583f61d78b3/packages/astro/src/core/config/config.ts#L65-L77

Also, the docs team were also interested in the API discussion for this, so might be good to run by them if this new syntax is better. cc @withastro/maintainers-docs

matthewp (approved):

ematipico:
@bluwy 

> Should we also validate if the deprecated configs are used in runtime, somewhere around here?

Two reasons why we can't do that:
- these keys **throw an error**, which means it won't be a deprecation;
- the code reads only the first level of keys, which doesn't fit our case because both keys are inside `build`

TheOtterlord (commented):
Not core docs, but I found a few typos

bluwy:
> Two reasons why we can't do that:
> 
> * these keys **throw an error**, which means it won't be a deprecation;
> * the code reads only the first level of keys, which doesn't fit our case because both keys are inside `build`

I mean we can to the warning deprecations, and validate `build.*` around that area of the code, not necessarily error-ing like it. But it would be a nice to have for me and not something needed.

PR #7840, opened 2023-07-27
feat: remove webapi in favor of a smaller polyfill

Princesseuh:
## Changes

This PR removes the `@astrojs/webapi` package completely, instead replacing it with a very small polyfill. `@astrojs/webapi` used to be necessary, but Node has gotten pretty good over the years at supporting many of those APIs natively, and for the rest, we mainly provided no-op anyway.

image

## Testing

Tests should still pass!

## Docs

N/A


matthewp (approved):

ematipico:
> This is to be merged in next, so we can do this whenever we'll actually release 3.0

I hope you know how, because I wouldn't know how to do if the files don't exist anymore 🤣 (even in `next` branch)

PR #7843, opened 2023-07-27
Track added links/styles/scripts to prevent duplicates

matthewp:
## Changes

- https://github.com/withastro/astro/commit/80e3d4d3d0f7719d8eae5435bba3805503057511#diff-3b2bb624348e4447bb08d85298dfd18bbd849514839f175cf0ccbc1de04c2a33 introduced a regression where head content was not propagated.
- It turns out not to be a small thing, but unique style/script rendering was working by mistake perhaps. 
- Uses string keys during rendering to tell if we've rendered a script/style/link or not. And only render when we have.

## Testing

- packages/integrations/mdx/test/css-head-mdx.test.js was failing but now should pass.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/yellow-olives-sing.md
+++ b/.changeset/yellow-olives-sing.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes head propagation regression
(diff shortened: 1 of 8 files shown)


matthewp:
Going to merge this since it's a regression. Happy to revert post-release if anyone sees a problem with it!

bluwy:
@matthewp I actually had the fix at https://github.com/withastro/astro/pull/7838. I also noticed the duplicate links but it was fixed in a simpler way. The regression was due to my rendering refactoring.

I think we can revert this and merge my PR instead? 😬

Issue #7188, opened 2023-05-23
Improve Cloudflare Integration in Hybrid Mode to Optimize Invocations


tonyaellie:
### What version of `astro` are you using?

2.5.4

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

Currently, in the hybrid mode of our application, the Cloudflare integration utilizes an exclusion-based approach to determine which pages should be excluded from Cloudflare's function invocation. This means that by default, 404s will result in a function invocation, by switching to only including the opted in pages invocations will be decreased.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hb99m1

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Related: https://github.com/withastro/astro/issues/6516

natemoo-re:
I believe this was fixed in https://github.com/withastro/astro/pull/7846, but if that's not the case, please feel free to reopen!

PR #7846, opened 2023-07-27
more efficient _routes.json for cloudflare integration

### LINKED ISSUE #6516: Very large `exclude` attribute on  `_routes.json` file, crashing deployments on Cloudflare Pages

2.1.2

Cloudflare @6.2.0

npm

Linux

Firefox

### Describe the Bug

I have my fonts locally and there are at most 500 files. These files get bundled onto `_astro` and added to `_routes.json` to the `excluded` param, producing then the following error:

```text
Error: Failed to publish your Function. Got error: Error 8000057: Rules in `_routes.json` are over the 100 rule limit. Refer to https://cfl.re/3FsE4aF.
```

I'm a bit confused about what is actually required to be on the `_routes.json`...

On the https://stackblitz.com/edit/github-jllrss?file=dist/_routes.json I show that the woff is excluded (even though it is a fake file). Imagine having 500+ files? Throws the error above

https://stackblitz.com/edit/github-jllrss?file=dist/_routes.json

### Participation

schummar:
## Changes
Fixes #6516

The cloudflare integration tends to create large _routes.json files, which quickly hits the limit of 100 combined include/exclude entries that Cloudflare imposes even for small projects.

This PR should create smaller _routes.json by
- Creating include entries for non prerendered endpoints/pages
- Creating exclude entries only for those static files, that are matched by one of the includes

That means if you have many assets and/or static pages, instead of
```json
{
  "include": ["/*"],
  "exclude": [
    "<too many entries>"
  ]
}
```
you'll get
```json
{
  "include": [
    "/someRoute",
    "/anotherRoute/*"
  ],
  "exclude" [
    "/anotherRoute/staticPart/",
    "/anotherRoute/staticPart/index.html",
  ]
}
```

For cases, where the "exclude all static files" strategy produces less entries - e.g. when using `server` mode and not prerender any/most of the pages - the old output will be used.

There is one interesting edge case, where I am really not sure what the best solution would be. If there are no dynamic routes, the `include` array would be empty. Cloudflare does not allow this however. Usually you would not even need to use the integration for such a case, but if you do, we need a fallback value in `include`. I have chosen a magic value `/MATCH_NONE_INCLUDE_PATTERN` for now, but maybe someone has a better idea...

## Testing

Added a test case.
Coincidentally had to fix an actual pre existing bug to make it pass 😉 (Windows file paths ended up in _routes.json)

## Docs


ematipico:
Thank you! I am about to review. In the meantime, a question for you: What's `/MATCH_NONE_INCLUDE_PATTERN`?

packages/integrations/cloudflare/src/index.ts
```diff
@@ -39,6 +39,8 @@ const SHIM = `globalThis.process = {
 
 const SERVER_BUILD_FOLDER = '/$server_build/';
 
+const matchNoneIncludePattern = '/MATCH_NONE_INCLUDE_PATTERN';
```
ematipico, line 42: It would be nice to add some comment here. Ideally, a brief explanation or a link to the documentation that explains this magic value.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -233,14 +235,39 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				// cloudflare to handle static files and support _redirects configuration
 				// (without calling the function)
 				if (!routesExists) {
+					const functionEndpoints = routes
+						.filter((route) => ['endpoint', 'page'].includes(route.type) && !route.prerender)
```
ematipico, line 239: You can move `['endpoint', 'page']` inside a top-level `const`, so we don't create the same array at each iteration. Also, could you leave a comment that explains why we have this filter?

packages/integrations/cloudflare/src/index.ts
```diff
@@ -324,3 +375,21 @@ export default function createIntegration(args?: Options): AstroIntegration {
 function prependForwardSlash(path: string) {
 	return path[0] === '/' ? path : '/' + path;
 }
+
+function deduplicatePatterns(patterns: string[]) {
```
ematipico, line 379: I think a comment that explains what this function does is important. We can explain how the deduplication is done, because it seems there's some specific logic here.

packages/integrations/cloudflare/test/fixtures/routesJson/astro.config.mjs
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from 'astro/config';
+import cloudflare from '@astrojs/cloudflare';
+
+
+
+export default defineConfig({
```
ematipico, line 6:
```suggestion
import cloudflare from '@astrojs/cloudflare';

export default defineConfig({
```

schummar:
> Thank you! I am about to review. In the meantime, a question for you: What's `/MATCH_NONE_INCLUDE_PATTERN`?

I have it in the description:

> There is one interesting edge case, where I am really not sure what the best solution would be. If there are no dynamic routes, the include array would be empty. Cloudflare does not allow this however. Usually you would not even need to use the integration for such a case, but if you do, we need a fallback value in include. I have chosen a magic value /MATCH_NONE_INCLUDE_PATTERN for now, but maybe someone has a better idea...

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
+import { expect } from 'chai';
+import { loadFixture } from './test-utils.js';
+
+/** @type {import('./test-utils.js').Fixture} */
+describe('generate _routes.json', () => {
```
ematipico, line 5:
```suggestion
describe('_routes.json generation', () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+	after(() => {
+		delete process.env.SRC;
+	});
+
+	describe('both functions and static files', () => {
```
ematipico, line 10:
```suggestion
	describe('of both functions and static files', () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+
+		before(async () => {
+			process.env.SRC = './src/mixed';
+			fixture = await loadFixture({
+				root: './fixtures/routesJson/',
+			});
+			await fixture.build();
+		});
+
+		it('creates includes for functions and excludes for static files where needed', async () => {
```
ematipico, line 21:
```suggestion
		it('creates `include` for functions and `exclude` for static files where needed', async () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+
+			expect(routes).to.deep.equal({
+				version: 1,
+				include: ['/a/*'],
+				exclude: ['/a/', '/a/redirect', '/a/index.html'],
+			});
+		});
+	});
+
+	describe('only functions', () => {
```
ematipico, line 33:
```suggestion
	describe('of only functions', () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+
+		before(async () => {
+			process.env.SRC = './src/dynamicOnly';
+			fixture = await loadFixture({
+				root: './fixtures/routesJson/',
+			});
+			await fixture.build();
+		});
+
+		it('creates a wildcard include and excludes only for the redirect', async () => {
```
ematipico, line 44:
```suggestion
		it('creates a wildcard `include` and `exclude` only for the redirect', async () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+
+			expect(routes).to.deep.equal({
+				version: 1,
+				include: ['/*'],
+				exclude: ['/a/redirect'],
+			});
+		});
+	});
+
+	describe('only static files', () => {
```
ematipico, line 56:
```suggestion
	describe('of only static files', () => {
```

packages/integrations/cloudflare/test/routesJson.js
```diff
@@ -0,0 +1,78 @@
...
+
+		before(async () => {
+			process.env.SRC = './src/staticOnly';
+			fixture = await loadFixture({
+				root: './fixtures/routesJson/',
+			});
+			await fixture.build();
+		});
+
+		it('create only one include that is supposed to match nothing', async () => {
```
ematipico, line 67:
```suggestion
		it('create only one `include` that is supposed to match nothing', async () => {
```

ematipico:
@withastro/maintainers-docs friendly ping to review the small change in documentation

schummar:
Thanks for the review @ematipico. I found #7188, which is also related. There unnecessary invocations for 404 pages are discussed, which should also be solved by using actual includes.

So with that in mind, should we always use includes by default and only fall back to the reverse approach when the list gets over 100 entries, while the alternative has less? Right now we would choose the approach that produces the smaller list.
Or maybe a setting?

ematipico:
> > Thank you! I am about to review. In the meantime, a question for you: What's `/MATCH_NONE_INCLUDE_PATTERN`?
> 
> I have it in the description:
> 
> > There is one interesting edge case, where I am really not sure what the best solution would be. If there are no dynamic routes, the include array would be empty. Cloudflare does not allow this however. Usually you would not even need to use the integration for such a case, but if you do, we need a fallback value in include. I have chosen a magic value /MATCH_NONE_INCLUDE_PATTERN for now, but maybe someone has a better idea...
> 
> 

I read the description, but I still don't understand this value. Is it just a placeholder or something that cloudflare recognises?

schummar:
> I read the description, but I still don't understand this value. Is it just a placeholder or something that cloudflare recognises?

Then I have not been clear enough, sorry. Assume you are using hybrid mode and don't opt out of prerender for any page/endpoint (yet). That means we would get this `_routes.json`:
```json
{
  "version": 1,
  "include": [],
  "exclude": []
}
```

Cloudflare does not allow this and throws an error on deployment:
```
Invalid _routes.json file found at: <path>/dist
Routes must have at least 1 include rule, but no include rules were detected.
```

The limitation is documented: https://developers.cloudflare.com/pages/platform/functions/routing/#limits
So to solve this, I had to add some entry to `include`. I wanted something that will likely not match anything real, hence the weird magic value.

**However, while writing this response, I had a better idea: Lets do this:**
```json
{
  "version": 1,
  "include": ["/"],
  "exclude": ["/"]
}
```
Still a workaround for Cloudflare's limitation but not as baffling. And it can actually not match anything since `exclude` trumps `include`.

ematipico:
Thank you @schummar , I had a different interpretation of "magic value", but I understand now. Thank you!

ematipico (approved):

packages/integrations/cloudflare/README.md
```diff
@@ -106,7 +106,8 @@ Cloudflare has support for adding custom [headers](https://developers.cloudflare
 
 ### Custom `_routes.json`
 
-By default, `@astrojs/cloudflare` will generate a `_routes.json` file that lists all files from your `dist/` folder and redirects from the `_redirects` file in the `exclude` array. This will enable Cloudflare to serve files and process static redirects without a function invocation. Creating a custom `_routes.json` will override this automatic optimization and, if not configured manually, cause function invocations that will count against the request limits of your Cloudflare plan.
+By default, `@astrojs/cloudflare` will generate a `_routes.json` file that lists function endpoints in the `include` array and files from your `dist/` folder, which would match one of the includes, in the `exclude` array.
```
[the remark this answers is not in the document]
schummar: If it's ok for you, I would actually try to get less specific here. The behavior is a bit more complex than before and describing it accurately would need more space. Stating that the file will be generated "based on your applications's dynamic and static routes" might be enough for the user. And I added a link to Cloudflare's docs for more in depth information about `_routes.json` in general.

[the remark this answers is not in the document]
schummar: @sarah11918 did you have a chance to look at this again?

sarah11918 (approved):
Hey @schummar ! This rewrite looks clearer, and I think it's helpful! Thanks!

vvashisht:
This approach disregards any routes present in the functions directory and not in astro's scope. I'm using the 'directory' mode and I have few routes in the functions directory that stopped working because of this change. 

Is it possible to factor those in? Or have a way of supplementing the 'includes' array somehow?

schummar:
That is true. The manual work around right now: You can define your own `_routes.json`.
https://docs.astro.build/de/guides/integrations-guide/cloudflare/#custom-_routesjson

To properly address the issue I see multiple possibilities:
1. Make the strategy configurable. As I mentioned, at the moment the strategy that produces less entries will be chosen. We could easily allow users to override it in the integrations config - e.g. `routes: { strategy: 'exclude' | 'include' | 'auto' }`

2. Allow modifying the routes in the config. E.g. `routes: { include: ['/myCustomFunction'], exclude: [] }` which could be merged with the automatically calculated routes.

3. Traverse the functions directory and include the routes automatically. I haven't done much with custom routes so far so I can't say how complex that can become. I imagine picking up simple routes like `functions/myFunction.js` would be easy, while handling dynamic routes there could be a problem.

The first two options are straight forward to implement but add to the config and docs complexity, so the team should voice their opinion about it. The third option is the most powerful but also most complex to implement. I suppose the recommended way to do functions is to write Astro endpoints and let the functions be generated (in which case everything works fine), so the third option might not be worth it.

My recommendation is implementing 1 and 2. Opinions?

vvashisht:
Yes, that is what I ended up doing indeed :)

Cloudflare's 100 rule constraint outweighs the hassle of manual route configuration. So, I do think that this PR serves a critical role.

On the other hand, if the current implementation doesn't evolve, anyone using the 'directory' mode to implement routes/pages/middleware outside astro's scope can not benefit from optimized automatic routing config.

Also, I might not be the only one facing this problem right now.

In my opinion, as a first step, the documentation should mention that opting for 'directory' mode necessitates manual routing config, if one intends to put any route in the functions directory. The possible points for this addition can be at one or both locations:
https://docs.astro.build/en/guides/integrations-guide/cloudflare/#mode
https://docs.astro.build/en/guides/integrations-guide/cloudflare/#custom-_routesjson

Further, a build log notice can be added to communicate that in absence of manual config, 'functions' routes will be ignored.

---

As for the route generation logic, I agree with your proposal. I think option 3 is both an overkill and a bug trap.

It would be ideal to implement both 1 and 2, but, and this is just my opinion, option 2 alone could suffice in a pinch.

Having said that, for option 2, merging of user defined config needs a bit of consideration regarding conflict resolution.

vvashisht:
@ematipico I understand that the team must be busy on multiple fronts. But if you could please spare a quick opinion:

Should this [side effect](https://github.com/withastro/astro/pull/7846#issuecomment-1693870028) be fixed? If so, should I create a separate issue for it?

schummar:
I am back from vacation and have created a pr with a proposal: #8459

PR #7852, opened 2023-07-28
Improve config error handling

bluwy:
## Changes

- Improve error handling on initial startup and dev restarts
- Handle error logging directly in the places where the error emits
- Simplify `dev()` public JS API (remove `beforeRestart` and `handleConfigError`)

## Testing

Existing tests should pass. We don't have a lot of these kind of tests, I also tested the commands manually.

## Docs

n/a. internal change, the error messages should still appear unchanged.


packages/astro/src/cli/flags.ts
bluwy, line 1: This file is a rename from `load-settings.ts` but Git doesn't see it that way.

packages/astro/src/core/config/config.ts
```diff
@@ -80,7 +84,21 @@ export async function validateConfig(
 	const AstroConfigRelativeSchema = createRelativeSchema(cmd, root);
 
 	// First-Pass Validation
-	const result = await AstroConfigRelativeSchema.parseAsync(userConfig);
+	let result: AstroConfig;
+	try {
+		result = await AstroConfigRelativeSchema.parseAsync(userConfig);
+	} catch (e) {
+		// Improve config zod error messages
+		if (e instanceof ZodError) {
```
bluwy, line 92: Errors are now logged more directly.

packages/astro/src/core/dev/dev.ts
```diff
@@ -36,9 +35,6 @@ export default async function dev(
 		inlineConfig,
 		logging: options.logging,
 		fs,
-		handleConfigError: options.handleConfigError,
-		// eslint-disable-next-line no-console
-		beforeRestart: () => console.clear(),
```
bluwy, line 41: The `beforeRestart` was only created because on `console.clear()`, as we don't want to call `console.clear` in tests (test results gets removed).

Personally I find clearing the console moves and cuts out the terminal a lot that it's better to not do it instead, so I removed it.

packages/astro/src/cli/flags.ts
```diff
@@ -0,0 +1,23 @@
...
+		server: {
+			port: typeof flags.port === 'number' ? flags.port : undefined,
+			host:
+				typeof flags.host === 'string' || typeof flags.host === 'boolean' ? flags.host : undefined,
+			open: typeof flags.open === 'boolean' ? flags.open : undefined,
+		},
+		experimental: {
+			assets: typeof flags.experimentalAssets === 'boolean' ? flags.experimentalAssets : undefined,
+		},
+	};
```
Princesseuh, line 22: do we have a `--experimentalViewTransitions` flag?

ematipico: I think we don't; I believe we forgot about it. We should probably add it.

packages/astro/src/core/errors/errors.ts
```diff
@@ -141,6 +141,12 @@ export class AggregateError extends AstroError {
 	}
 }
 
+/**
+ * Check if an error is a ZodError from an AstroConfig validation.
+ * Used to suppress formatting a ZodError if needed.
+ */
+export const astroConfigZodErrors = new WeakSet();
```
ematipico, line 148: Usually, I tend to **not expose** such internal types from a module. I tend to expose functions that internally mutate the object. Nothing against it, but it's a pattern that I don't feel comfortable about, it seems more a side effect.

packages/astro/src/core/dev/restart.ts
```diff
@@ -4,8 +4,11 @@ import * as vite from 'vite';
 import type { AstroInlineConfig, AstroSettings } from '../../@types/astro';
 import { eventCliSession, telemetry } from '../../events/index.js';
 import { createSettings, resolveConfig } from '../config/index.js';
+import { collectErrorMetadata } from '../errors/dev/utils.js';
+import { astroConfigZodErrors } from '../errors/errors.js';
 import { createSafeError } from '../errors/index.js';
-import { info, type LogOptions } from '../logger/core.js';
+import { error as _error, info, type LogOptions } from '../logger/core.js';
```
ematipico, line 10: nit: `_error` => `logError`

packages/astro/src/core/dev/dev.ts
ematipico: Totally agree, especially if there are scripts/users that rely on stdout/stderr

packages/astro/src/core/config/config.ts
```diff
@@ -172,11 +190,19 @@ async function loadConfig(
...
-	});
+	try {
+		return await loadConfigWithVite({
+			root,
+			configPath,
+			fs: fsMod,
+		});
+	} catch (e) {
+		const configPathText = configFile ? colors.bold(configFile) : 'your Astro config';
+		// Config errors should bypass log level as it breaks startup
```
ematipico, line 201: About the comment: Is it because `--silent` can potentially not log these errors?

ematipico (approved):

packages/astro/src/core/errors/errors.ts
bluwy: Yeah I figured it could be a function too, but then I'd have to export two `setAstroConfigZodError` and `isAstroConfigZodError` APIs instead of one. I'm leaning towards not putting it in `config.ts` to remove the need for `setAstroConfigZodError` because that leads to the CLI `throw-and-exit.ts` needing to take longer to be `import()`-ed.

packages/astro/src/core/config/config.ts
bluwy: Yeah `--silent` would otherwise not log that the config failed to load. Try to pass the `logging` down here is a bit tricky, especially the fact that an Astro config `logLevel` can also contribute to how `logging` works.

packages/astro/src/core/errors/errors.ts
bluwy: I pushed the changes. Admittedly I think it makes things more cleaner now (had type issues) so thanks for nudging me to that 😄

packages/astro/src/core/config/config.ts
bluwy: Actually there's no `logLevel` config in the Astro config, got it mixed up. I think it's still probably fine always logging it here as the other code in this file also does that, but I'll revisit this again if it's not ideal.

Issue #7833, opened 2023-07-27
View transitions - FOUC when using back button browser


thomasledoux1:
### What version of `astro` are you using?

2.9.3

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome 115.0.5790.102

### Describe the Bug

I activated View Transitions on my website, and I added a specific transition between the titles of my blog posts on the overview page, and the title on the blog detail page.
When I click on the blog preview on the overview, I can see the title transition, works fine.
But when I hit the back button in my browser, I see a flash of unstyled content, so the CSS seems to be applied later? 
Not sure how/why this is happening.
Can be tested on https://website-thomas-astro-58u9o7hec-thomasledoux1.vercel.app/blog.
GitHub: https://website-thomas-astro-58u9o7hec-thomasledoux1.vercel.app/blog

### What's the expected result?

No FOUC

### Link to Minimal Reproducible Example

https://github.com/thomasledoux1/website-thomas-astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Thanks for the repro! Will take a look today and see what's going on.

FrancoJavierGadea:
I have the same problem here: https://64c233ea8c4b4d00081f2eb7--libros1812.netlify.app/

In localhost, it works great, but when I deploy it, flashing appears.

Although applying an inline CSS patch helped, it's not the ideal solution.

matthewp:
I think the problem is likely because it naively swaps the entire documentElement. I think it probably needs to diff the head instead. Working on that.

lukeshafer:
Looks like the issue is understood here, but thought I'd mention that I'm experiencing the same/similar issue - FOUC when navigating, and only on the deployed site. I can provide more information if needed

davycoe:
> Looks like the issue is understood here, but thought I'd mention that I'm experiencing the same/similar issue - FOUC when navigating, and only on the deployed site. I can provide more information if needed

I've a similar FOUC when navigating using Chrome. Deployed to CloudflarePages, FireFox seems to work as expected,

Example here: [https://7974c724.astro-transitions.pages.dev/](url)

matthewp:
This will be fixed in #7861. With that change, links are left untouched if they exist on both pages, so there should be no FOUC. New links are preloaded before the DOM swap happens.

You can test it out with the preview release:

```
[...]

PR #7861, opened 2023-07-28
Persistent DOM in ViewTransitions

matthewp:
## Changes

- This enables islands or any DOM element to be persisted across page navigation. As described in the RFC: https://github.com/withastro/roadmap/blob/view-transitions/proposals/0040-view-transitions.md#persistent-islands
- Also fixes https://github.com/withastro/astro/issues/7833 . For links uses the same `data-astro-transition-persist` attr to keep links in the page, untouched.
- Compiler changes: https://github.com/withastro/compiler/pull/840/

## Testing

- Test added for a video element
- Test added for a client-side island

## Docs

- https://github.com/withastro/docs/pull/3993


matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

matthewp:
!preview view-transitions

matthewp:
Demo: https://stackblitz.com/edit/github-stfdbe-ikkp2e?file=src%2Fpages%2Findex.astro,src%2Fpages%2Ftwo.astro,src%2Fcomponents%2FVideo.astro,src%2Fcomponents%2FLayout.astro

matthewp:
!preview view-transitions

packages/astro/src/runtime/server/hydration.ts
```diff
@@ -166,5 +166,11 @@ export async function generateHydrateScript(
 		})
 	);
 
+	['data-astro-transition-scope', 'data-astro-transition-persist'].forEach(name => {
```
ematipico, line 169: nit: we could move this array in a top-level `const` variable, so we don't create it every single time

packages/astro/components/ViewTransitions.astro
```diff
@@ -97,17 +125,21 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+		const links: Promise<any>[] = [];
+		for(const el of doc.querySelectorAll('head link[rel=stylesheet]')) {
+			// Do not preload links that are already on the page.
+			if(!document.querySelector(`[${PERSIST_ATTR}="${el.getAttribute(PERSIST_ATTR)}"]`)) {
+				const c = document.createElement('link');
+				c.setAttribute('rel', 'preload');
+				c.setAttribute('as', 'style');
+				c.setAttribute('href', el.getAttribute('href')!);
+				links.push(new Promise<any>(resolve => {
 					['load', 'error'].forEach((evName) => c.addEventListener(evName, resolve));
```
ematipico, line 137: Out of curiosity, are these events removed when go to a new page?

ematipico (approved):

packages/astro/components/ViewTransitions.astro
```diff
@@ -87,8 +88,35 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+				const newEl = id && doc.head.querySelector(`[${PERSIST_ATTR}="${id}"]`);
+				// If the element exists in the document already, remove it
+				// from the new document and leave the current node alone
+				if(newEl) {
+					newEl.remove();
+				} else {
+					// Otherwise remove the element in the head. It doesn't exist in the new page.
+					el.remove();
+				}
+			}
```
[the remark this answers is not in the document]
matthewp: That would diff everything, right? This is not really want we want, we want to only keep nodes marked with `transition:persist`, and links which are implicitly marked as such, by us. I don't see this becoming more complex though.

packages/astro/components/ViewTransitions.astro
matthewp: No, these are preload links and are created for new stylesheets only. I could add a `{ once: true }` but I can't think of any scenario where they'd fire again.

packages/astro/src/runtime/server/render/head.ts
```diff
@@ -17,11 +18,20 @@ export function renderAllHeadContent(result: SSRResult) {
...
+						...style.props,
+						// Add this attribute so ViewTransitions keeps the link in the page
+						'data-astro-transition-persist': shorthash(style.props.href)
+					},
+					children: style.children
+				});
+			} else {
+				return renderElement('style', style)
+			}
+		});
```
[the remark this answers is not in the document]
matthewp: Maybe you're saying we should be persisting all links anyways, even those created by users that are in public?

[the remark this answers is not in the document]
bluwy: > I could have it only add this if viewTransitions is turned on, would that help?

Yeah that would be fine by me too, but ultimately I indeed mean this:

> Maybe you're saying we should be persisting all links anyways, even those created by users that are in public?

too. So links in general are "persisted" over by default. I think it's more part of the client-side routing needs, than special case we're handling. But I could be wrong that that isn't the intention?

packages/astro/components/ViewTransitions.astro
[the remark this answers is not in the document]
bluwy: Alright 👍 I thought for `head` we probably want to persist anything as much as default instead of only those marked (similar to the discussion above). But if it's not the case here, we can ignore this for now.

packages/astro/src/runtime/server/render/head.ts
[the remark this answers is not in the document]
matthewp: Ok, made this change in https://github.com/withastro/astro/pull/7861/commits/3e11ef2e742e0cb5814c7c797142a25b62d63ab2

bluwy (approved):
Looks great!

.changeset/empty-experts-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Persistent islands
```
matthewp, line 5: Update this!

.changeset/empty-experts-unite.md
```diff
@@ -0,0 +1,25 @@
+---
+'astro': minor
+---
+
+Persistent DOM and Islands in Experimental View Transitions
+
+If you have `viewTransitions: true` enabled in your Astro config's experimental section, you can now take advantage of the new `transition:persist` directive. With this directive you can keep the state of DOM elements and islands on the old page when transitioning to the new page.
```
Yan-Thomas, line 7: Comma nit
```suggestion
If you have `viewTransitions: true` enabled in your Astro config's experimental section, you can now take advantage of the new `transition:persist` directive. With this directive, you can keep the state of DOM elements and islands on the old page when transitioning to the new page.
```

.changeset/empty-experts-unite.md
```diff
@@ -0,0 +1,25 @@
...
+
+In 2.10 you can now add the `transition:persist` directive and this video will be moved over to the next page (if the video also exists on that page).
+
+Likewise, this feature works with any client-side framework component island. In this example, a counter's state is preserved and moved to the new page:
+
+```astro
+<Counter count={5} client:load transition:persist />
+```
+
+See the [docs](https://docs.astro.build/en/guides/view-transitions/#maintaining-state) to learn more on usage.
```
Yan-Thomas, line 25: Always good to have a more descriptive link for accessibility.
```suggestion
See our [View Transitions Guide](https://docs.astro.build/en/guides/view-transitions/#maintaining-state) to learn more on usage.
```

Yan-Thomas (commented):
Thanks Matthew, adding a few suggestions before Sarah's take on this!

sarah11918 (approved):
Just a couple of quick suggestions for you to consider! Looks good!

PR #7871, opened 2023-07-31
Move `logging` option to `logLevel` inline config option

bluwy:
## Changes

**NOTE: this merges into the js-api branch**

- Changes `dev({ root: '...' }, { logging: { ... } })` to `dev({ root: '...', logLevel: '...' })`
- Similar changes for `build`, `preview`, and `sync`
- However, you can still do `dev({ logging: { ... } })`. I added `logging` as an internal option because some tests passes a custom `logging` to collect logs.

NOTE: After this is merged, I think we can merge `js-api` into `main`, then decide if we want to expose the JS API as a feature since the changes so far are mostly refactors.

## Testing

Existing tests should pass. Also tested the commands manually.

## Docs

n/a. Internal change as JS API is not exposed yet.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -19,7 +19,7 @@ import type { PageBuildData } from '../core/build/types';
 import type { AstroConfigSchema } from '../core/config';
 import type { AstroTimer } from '../core/config/timer';
 import type { AstroCookies } from '../core/cookies';
-import type { LogOptions } from '../core/logger/core';
+import type { LogOptions, LoggerLevel } from '../core/logger/core';
 import type { AstroComponentFactory, AstroComponentInstance } from '../runtime/server';
 import type { SUPPORTED_MARKDOWN_FILE_EXTENSIONS } from './../core/constants.js';
 export type {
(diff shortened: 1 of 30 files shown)


packages/astro/test/test-utils.js
```diff
@@ -103,8 +104,8 @@ export const silentLogging = {
 export async function loadFixture(inlineConfig) {
 	if (!inlineConfig?.root) throw new Error("Must provide { root: './fixtures/...' }");
 
-	/** @type {import('../src/core/logger/core').LogOptions} */
-	const logging = defaultLogging;
+	// Silent by default during tests to not pollute the console output
+	inlineConfig.logLevel = 'silent';
```
bluwy, line 108: Set the default logging to silent here, so most of the tests I've removed the manual `logging: silentLogging` config.

packages/astro/src/@types/astro.ts
```diff
@@ -1331,9 +1331,15 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
...
-export interface AstroInlineConfig extends AstroUserConfig {
+export interface AstroInlineConfig extends AstroUserConfig, AstroInlineOnlyConfig {}
+export interface AstroInlineOnlyConfig {
 	configFile?: string | false;
 	mode?: RuntimeMode;
+	logLevel?: LoggerLevel;
+	/**
+	 * @internal for testing only
+	 */
+	logging?: LogOptions;
```
bluwy, line 1342: Adds two new options to the inline config. I'm not really happy to add `logging` here, but if we can get Ema's `Logger` proposal, maybe we can substitute to that instead which is more user-friendly.

ematipico: I totally agree

ematipico (approved):

PR #7872, opened 2023-07-31
Move MDX plugin re-ordering hack to MDX integration

bluwy:
## Changes

... instead of doing it in Astro core.

## Testing

Existing tests should pass. However, `next` might need to merge up `main` changes for the recent fixes to the MDX fails.

## Docs

n/a. This change should not be visible as long as they upgrade Astro core and MDX integration to latest

### DIFF
--- a/packages/integrations/mdx/package.json
+++ b/packages/integrations/mdx/package.json
@@ -53,6 +53,9 @@
     "unist-util-visit": "^4.1.2",
     "vfile": "^5.3.7"
   },
+  "peerDependencies": {
+    "astro": "workspace:^2.9.6"
+  },
   "devDependencies": {
     "@types/chai": "^4.3.5",
     "@types/estree": "^1.0.1",
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -95,6 +95,21 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 								enforce: 'pre',
 								configResolved(resolved) {
 									importMetaEnv = { ...importMetaEnv, ...resolved.env };
+
+									// HACK: move ourselves before Astro's JSX plugin to transform things in the right order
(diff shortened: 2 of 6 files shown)


ematipico:
> n/a. This change should not be visible as long as they upgrade Astro core and MDX integration to latest

Maybe we should enforce this via `package.json` using `peerDependencies`?

bluwy:
Good call, looks like most other integrations has the peer dep, will add it for MDX too.

packages/integrations/mdx/src/index.ts
```diff
@@ -95,6 +95,21 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
...
+											(p) => p.name === '@mdx-js/rollup'
+										);
+										if (myPluginIndex !== -1) {
+											const myPlugin = resolved.plugins[myPluginIndex];
+											// @ts-ignore-error ignore readonly annotation
+											resolved.plugins.splice(myPluginIndex, 1);
+											// @ts-ignore-error ignore readonly annotation
+											resolved.plugins.splice(jsxPluginIndex, 0, myPlugin);
+										}
+									}
```
bluwy, line 112: Note: This hack has been what I've been doing at https://github.com/bluwy/whyframe/blob/72e34498748f0b45d9272c886e123efd4dd24e17/packages/jsx/src/index.js#L26-L37 too

packages/integrations/mdx/package.json
```diff
@@ -53,6 +53,9 @@
     "unist-util-visit": "^4.1.2",
     "vfile": "^5.3.7"
   },
+  "peerDependencies": {
+    "astro": "workspace:^2.9.6"
```
ematipico, line 57: Should this become `3.0.0` before the release?

ematipico (approved):
Looks good! Left a question, not a blocker :)

packages/integrations/mdx/package.json
bluwy: I think changesets or pnpm will bump that before the release. We should definitely check our existing integrations also have that bumped before the initial release too.

PR #7874, opened 2023-07-31
feat: change default port to 4321

ematipico:
## Changes

Changes Astro's defualt port to `4321`

## Testing

Current tests should pass

## Docs

### DIFF
--- a/.changeset/unlucky-sheep-build.md
+++ b/.changeset/unlucky-sheep-build.md
@@ -2,4 +2,4 @@
 'astro': major
 ---
 
-Astro's default port is now `4321`.
+Astro's default port when running the dev or preview server is now `4321`.
--- a/.changeset/unlucky-sheep-build.md
+++ b/.changeset/unlucky-sheep-build.md
@@ -1,0 +1,5 @@
+---
+'astro': major
+---
+
+Astro's default port is now `4321`.
--- a/.changeset/unlucky-sheep-build.md
+++ b/.changeset/unlucky-sheep-build.md
@@ -3,3 +3,5 @@
 ---
 
 Astro's default port when running the dev or preview server is now `4321`.
+
+This will reduce conflicts with ports used by other tools.
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -326,7 +326,7 @@ type ServerConfig = {
 	/**
 	 * @name server.port
 	 * @type {number}
-	 * @default `3000`
+	 * @default `4321`
 	 * @description
 	 * Set which port the dev server should listen on.
 	 *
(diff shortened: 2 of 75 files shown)


fflaten:
Would you mind adding a line about why it's changed at all?

.changeset/unlucky-sheep-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
+---
+
+Astro's default port is now `4321`.
```
delucis, line 5: Maybe worth specifying this is related to `astro dev` and `astro preview`?

```suggestion
Astro's default port when running the dev or preview server is now `4321`.
```

delucis (commented):
Left a small note for the CHANGELOG. We’ll also have a bunch of places that need updating in docs: https://github.com/search?q=repo%3Awithastro%2Fdocs%20%3A3000&type=code

Have we considered what happens when the port is busy with this change? Currently, if I run `astro dev`, but another project is already running, Astro automatically increments the port number and uses `3001` instead (or `3002` etc. until it finds a free port). Is this now `4321`, `4322`, `4323`?

ematipico:
> Would you mind adding a line about why it's changed at all? 🙂

Someone suggested the idea and it made sense for us. 4, 3, 2, 1 ... launch! Also, 3000 is often used by other frameworks.

ematipico:
> Have we considered what happens when the port is busy with this change? Currently, if I run `astro dev`, but another project is already running, Astro automatically increments the port number and uses `3001` instead (or `3002` etc. until it finds a free port). Is this now `4321`, `4322`, `4323`?

That logic hasn't changed. So we incrementally find free ports: `4321, `4322`, etc.

delucis (approved):
Thanks Ema, looks good. We _could_ mention our motivation in relation to @fflaten's comment, e.g. adding "in order to avoid conflicts with ports used by other tools" or something to the changelog.

I've never really cared one way or the other about this change, but docs look ok to me!

packages/astro/src/@types/astro.ts
```diff
@@ -326,7 +326,7 @@ type ServerConfig = {
 	/**
 	 * @name server.port
 	 * @type {number}
-	 * @default `3000`
+	 * @default `4321`
```
bluwy, line 329: There's another change like this in line 939

bluwy (commented):
I think there's still more places to change when I global-search for `3000`, e.g.

- `sandbox.config.json` (`"port": 3000`)
- `astro.config.mjs` (`site: 'http://localhost:3000'`)
- Some READMEs

Since CI doesn't run all the integration tests, it would be good to confirm if they pass too locally.

fflaten:
All README files says 4000 atm

ematipico:
> All README files says 4000 atm

Cheers for the review. It should be OK now.

Issue #7869, opened 2023-07-30
Warning: The getStaticPaths() statement has been ignored because `output: "server"` is set.


zepang:
### What version of `astro` are you using?

2.9.6

### Are you using an SSR adapter? If so, which one?

@astrojs/node

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I added a Markdown formatted page in the src
directory and executing npm run build in server-side rendering mode I get the following warning message:

```
[getStaticPaths] The getStaticPaths() statement in
src/pages/Creating a blog using Next.js SSG.md has been ignored because `output: "server"` is set.
```
 I  found that it is caused by the logic judgment of `vite-plugin-scanner` under astro(the version of astro is 2.9.6), and the file path is `astro/dist/vite-plugin-scanner/index.js`, where there is a judgment of :`! pageOptions.prerender && isServerLikeOutput(settings.config) && code.includes("getStaticPaths")`
it so happens that my post mentions the getStaticPaths method in next.js. which, in astro server-side rendering mode, gave me the warning message mentioned above

### What's the expected result?

I hope the code will be more accurate in determining the inclusion of "getStaticPaths" and won't issue a similar warning!

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-gsrh1x?file=astro.config.mjs&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #7876, opened 2023-07-31
fix: scan `getStaticPaths` only in `.astro` files

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/7869

## Testing

I checked manually


packages/astro/src/vite-plugin-scanner/index.ts
```diff
@@ -43,12 +44,15 @@ export default function astroScannerPlugin({
...
+			const fileExtension = extname(filename);
 
 			// `getStaticPaths` warning is just a string check, should be good enough for most cases
 			if (
 				!pageOptions.prerender &&
 				isServerLikeOutput(settings.config) &&
-				code.includes('getStaticPaths')
+				code.includes('getStaticPaths') &&
+				// this should only be valid for `.astro` files
+				fileExtension === '.astro'
```
bluwy, line 55: nit: I think it'll be slightly more performant if we only call `extname` inline here, incase the other conditions short-circuit.

packages/astro/src/vite-plugin-scanner/index.ts
```diff
@@ -43,12 +44,13 @@ export default function astroScannerPlugin({
...
 			}
-
 			// `getStaticPaths` warning is just a string check, should be good enough for most cases
 			if (
 				!pageOptions.prerender &&
 				isServerLikeOutput(settings.config) &&
-				code.includes('getStaticPaths')
+				code.includes('getStaticPaths') &&
+				// this should only be valid for `.astro` files
+				extname(filename) === '.astro'
```
bluwy, line 53: Actually I just remembered this could happen for endpoints too, so we need to extend the extensions to `.js` and `.ts` too?

ematipico: If we include `.js` and `.ts` files, we are at square one 🤣

bluwy: Probably good enough to fix the issue since we skip `.md` and `.mdx`, but true it's not so robust 😄

bluwy (approved):
IIUC if we really want to do this right, we can also borrow the `exports` values from `es-module-lexer` 🤔 But this also works for me for now.

Issue #6623, opened 2023-03-21
import.meta.env.BASE_URL is including a trailing slash


sugarmaplemedia:
### What version of `astro` are you using?

2.1.4

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Firefox

### Describe the Bug

When I call `import.meta.env.BASE_URL` in my `src` files, it includes a trailing slash that renders incorrectly as `http://localhost/foo//bar`.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wccs8f?file=src/pages/index.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

sugarmaplemedia:
Update: setting `trailingSlash` in the config from the default `ignore` to `never` fixed the issue. Either way, its still a bug.

bluwy:
Makes sense to me. Looks like the behaviour was from https://github.com/withastro/astro/pull/5358, even though before it we had always prepend a slash. But I agree that if the default is `ignore`, it should probably leave the `base` as is.

There's a good chance someone might be relying on the old behaviour though so "fixing" this could be a breaking change.

natemoo-re:
Closing this since https://github.com/withastro/astro/pull/7878 has been merged to `next`. This will be fixed in Astro 3.0!

PR #7878, opened 2023-07-31
Preserve base slash when trailingSlash ignore

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/6623

When `trailingSlash` is set to `ignore`, the `import.meta.env.BASE_URL` will have a string that's the same as `config.base` as-is. Unlike `always` where it always adds a trailing slash, and unlike `never` where it'll never have a trailing slash.

Since the issue was created, looks like docs now [document the old, different behaviour](https://docs.astro.build/en/reference/configuration-reference/#base):

> By default, the value of import.meta.env.BASE_URL includes a trailing slash. If you have the [trailingSlash](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option set to 'never', you will need to add it manually in your static asset imports and URLs.

where `ignore` would work like `always` (Always have a trailing slash). Perhaps we should discuss if this change makes sense?

## Testing

Updated some tests to fix the old behaviour, the rest should still pass.

## Docs

Added changelog, and updated docs in `astro.ts`

### DIFF
--- a/.changeset/six-grapes-look.md
+++ b/.changeset/six-grapes-look.md
@@ -1,0 +1,15 @@
+---
+'astro': major
+---
+
+The value of `import.meta.env.BASE_URL`, which is derived from the `base` option, will no longer have a trailing slash added by default or when `trailingSlash: "ignore"` is set. The existing behavior of `base` in combination with `trailingSlash: "always"` or `trailingSlash: "never"` is unchanged.
+
+If your `base` already has a trailing slash, no change is needed.
+
+If your `base` does not have a trailing slash, add one to preserve the previous behaviour:
+
+```diff
+// astro.config.mjs
+- base: 'my-base',
++ base: 'my-base/',
+```
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -572,12 +572,7 @@ export interface AstroUserConfig {
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
 	 *
-	 * By default, the value of `import.meta.env.BASE_URL` includes a trailing slash. If you have the [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option set to `'never'`, you will need to add it manually in your static asset imports and URLs.
-	 *
-	 * ```astro
-	 * <a href="/docs/about/">About</a>
-	 * <img src=`${import.meta.env.BASE_URL}image.png`>
-	 * ```
+	 * The value of `import.meta.env.BASE_URL` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
 	 */
 	base?: string;
 
--- a/packages/astro/src/core/config/schema.ts
+++ b/packages/astro/src/core/config/schema.ts
@@ -8,7 +8,7 @@ import path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { BUNDLED_THEMES } from 'shiki';
 import { z } from 'zod';
-import { appendForwardSlash, prependForwardSlash, trimSlashes } from '../path.js';
+import { appendForwardSlash, prependForwardSlash, removeTrailingForwardSlash } from '../path.js';
 
 const ASTRO_CONFIG_DEFAULTS = {
 	root: '.',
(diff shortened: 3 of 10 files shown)


packages/astro/src/core/config/schema.ts
```diff
@@ -366,22 +366,14 @@ export function createRelativeSchema(cmd: string, fileProtocolRoot: string) {
...
 
-		// If there is no base but there is a base for site config, warn.
-		const sitePathname = config.site && new URL(config.site).pathname;
-		if (!trimmedBase.length && sitePathname && sitePathname !== '/') {
-			config.base = sitePathname;
-			/* eslint-disable no-console */
-			console.warn(`The site configuration value includes a pathname of ${sitePathname} but there is no base configuration.
-
-A future version of Astro will stop using the site pathname when producing <link> and <script> tags. Set your site's base with the base configuration.`);
-		}
```
bluwy, line 379: It looks like this had never worked for some time now. We do re-assign `config.base = sitePathname`, but the logic after this re-assign `config.base` based on something else again. So this compat feature had already been removed.

packages/astro/test/dev-routing.test.js
```diff
@@ -113,9 +113,9 @@ describe('Development Routing', () => {
 			expect(response.status).to.equal(200);
 		});
 
-		it('404 when loading subpath root without trailing slash', async () => {
+		it('200 when loading subpath root without trailing slash', async () => {
 			const response = await fixture.fetch('/blog');
-			expect(response.status).to.equal(404);
+			expect(response.status).to.equal(200);
 		});
```
bluwy, line 119: Had to update some tests behaviours when fetching a site with `base` without trailing slash. When you have `base: '/blog'` and `trailingSlash: 'ignore'`, fetching `http://mysite.com/docs` (no trailing slash) will now work.

ematipico (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -571,12 +571,7 @@ export interface AstroUserConfig {
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
 	 *
-	 * By default, the value of `import.meta.env.BASE_URL` includes a trailing slash. If you have the [`trailingSlash`](https://docs.astro.build/en/reference/configuration-reference/#trailingslash) option set to `'never'`, you will need to add it manually in your static asset imports and URLs.
-	 *
-	 * ```astro
-	 * <a href="/docs/about/">About</a>
-	 * <img src=`${import.meta.env.BASE_URL}image.png`>
-	 * ```
+	 * The value of `import.meta.env.BASE_URL` is the same as `base`, unless `trailingSlash: "always"` or `trailingSlash: "never"` is set, which will ensure or remove a trailing slash from the `base` respectively.
```
delucis, line 574: One query around "same as" — don't we add a leading slash to `base` if not present?
Maybe this is more accurate?
```suggestion
	 * The value of `import.meta.env.BASE_URL` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
```

delucis (commented):
Thanks @bluwy! Left some changeset and docs comments.

.changeset/six-grapes-look.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': major
+---
+
+The value of `import.meta.env.BASE_URL`, which is derived from the `base` option, will not be enforced with a trailing slash if `trailingSlash: "ignore"` is set. The existing behaviours of `trailingSlash: "always"` and `trailingSlash: "never"` are unchanged.
+
+To migrate to this new behaviour, if your `base` configuration already has a trailing slash, no change is needed. If your `base` configuration does not have a trailing slash, you can add one to preserve the previous behaviour, or make sure any usages of `import.meta.env.BASE_URL` works without the trailing slash.
```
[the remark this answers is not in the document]
bluwy: I initially had two migration paths since not everyone uses `import.meta.env.BASE_URL`, so not needing to add a trailing `/` would be fine too. But I guess this is already implied at the first paragraph so this works for me too.

packages/astro/src/@types/astro.ts
bluwy: > One query around "same as" — don't we add a leading slash to `base` if not present?

Yeah that's true. I'd perhaps like to keep that not mentioned to not overload how `base` works, and because the docs always recommend with a leading slash.

Your suggestion sounds good to me though 👍

bluwy:
Merging this in as the examples check is blocked by https://github.com/withastro/astro/pull/7998

PR #7879, opened 2023-07-31
Refactor and improve Astro config loading flow

bluwy:
## Changes

This is a feature branch of various PRs merging into `main`. There shouldn't be un-reviewed code here, except the added changeset, so I think we can merge this anytime?

Note that this does not add the JS API yet, but we could anytime we want. I figured we can decide that in a separate PR so this PR only contains the refactors.

## Testing

Tests should pass.

## Docs

n/a until the JS API is exported. The internal refactors shouldn't be visible, but I added a changeset just in case I regress something.


ematipico (approved):
If no changes were done in the meantime, I think we can merge it anytime!

PR #7887, opened 2023-07-31
feat: add logging for when hydrate's JSON parse fails

ffxsam:
## Changes

Per [this discussion](https://discord.com/channels/830184174198718474/830184175176122389/1135607503682547752), there seems to be an edge case(?) in production where hydration fails due to attempting to JSON.parse invalid props. This change simply adds logging so that people using Sentry—or similar error logging services—can see exactly what the issue is. This will likely either lead to devs finding issues in their code, or finding a possible bug in Astro which can then be addressed.

## Testing

No tests added, since this is just simple logging. But let me know if you feel something is needed here!

## Docs

N/A


packages/astro/src/runtime/server/astro-island.ts
```diff
@@ -127,9 +127,22 @@ declare const Astro: {
...
+					let props: Record<string, unknown>;
+
+					try {
+						props = this.hasAttribute('props')
+							? JSON.parse(this.getAttribute('props')!, reviver)
+							: {};
+					} catch (e) {
+						// eslint-disable-next-line no-console
+						console.error(
+							`[hydrate] Error parsing props for component ${this.Component?.__name}`,
```
natemoo-re, line 140: Not sure if `__name` will always exist, but we should also have `component-url` and `component-export` attributes to surface what (final, built) file is causing the problem.

natemoo-re (commented):
This looks great, thanks for the PR!

packages/astro/src/runtime/server/astro-island.ts
ffxsam: Cool, let me do some testing with those locally and I'll update the PR.

natemoo-re (approved):
Awesome, thanks for tackling this!

.changeset/gorgeous-starfishes-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
ematipico, line 2: This should be a `patch`

ematipico (changes requested):

bluwy (approved):
I suppose this increases the JS size a little, but if there's many reports of this I think it's worth it to track it down for now.

Otherwise perhaps we can shorten the error handling code more, if we want to keep the size smaller.

ffxsam:
IMO, all calls to `JSON.parse` should be wrapped in a try/catch block. It's one of the most commonly occurring errors in JavaScript applications because so much is assumed about the string being passed in, and the default error thrown by `JSON.parse` is useless (it's a shame the W3C spec didn't call for printing at least the first part of the string that caused the error to be thrown).

Maybe we create a `safeJsonParse' function that throws an actual useful error? In fact, I'm going to implement it in our application ASAP. 😄

packages/astro/src/runtime/server/astro-island.ts
```diff
@@ -127,9 +127,30 @@ declare const Astro: {
...
+
+					let props: Record<string, unknown>;
+
+					try {
+						props = this.hasAttribute('props')
+							? JSON.parse(this.getAttribute('props')!, reviver)
+							: {};
+					} catch (e) {
+						let componentName: string =
+							this.getAttribute('component-url') || this.Component?.__name || '<unknown>';
```
bluwy, line 139: Where's the `__name` coming from? If it's a framework-specific prop, I think we should remove it for now to simplify it.

ffxsam:
Addressed all final issues! Thanks for the review.

bluwy:
Thanks! Looks like there's an unrelated CI fail. I'll take a look at that before merging.

ematipico (approved):

PR #7888, opened 2023-07-31
Make Astro.cookies.get(key) return undefined

matthewp:
## Changes

- Astro.cookies.get(key) now returns `undefined` if the cookie isn't present.
- This is a breaking change, as previously you needed to use `has()` to check for existence.

## Testing

- Test updated

## Docs

- TBD

### DIFF
--- a/.changeset/twenty-cheetahs-deny.md
+++ b/.changeset/twenty-cheetahs-deny.md
@@ -1,0 +1,17 @@
+---
+'astro': major
+---
+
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
This will be a nice change!

ematipico (changes requested):
Branch must be rebased again, blocking to be safe

bluwy (approved):

PR #7892, opened 2023-08-01
feat: use external @astrojs/check

Princesseuh:
## Changes

- What does this change?


Princesseuh:
!preview astro-check

packages/astro/src/cli/install-package.ts
```diff
@@ -0,0 +1,118 @@
...
+	} catch (e) {
+		info(
+			logging,
+			'',
+			`To continue, Astro requires the following dependency to be installed: ${bold(packageName)}.`
+		);
+		const result = await installPackage([packageName, ...otherDeps], options, logging);
+
+		if (result) {
+			packageImport = await import(packageName);
```
bluwy, line 32: I think nodejs caches the result if it import fails the first time, so this second import of `packageName` would not work. I think we can do an initial `require.resolve` first to check the existence, then install to workaround it

Princesseuh: Oh? It worked in my tests, hmm

packages/astro/src/cli/check/index.ts
```diff
@@ -1,386 +1,39 @@
...
-	#filesCount: number;
-	#updateDiagnostics: NodeJS.Timeout | undefined;
-
-	constructor({
-		diagnosticChecker,
-		isWatchMode,
-		syncCli,
-		settings,
-		fileSystem,
+export async function check({ flags, logging }: CheckOptions) {
```
bluwy, line 14: Also a merge hint when rebasing `next` to align with `main`. The logging would need to be derived like so:

```js
const logging = createLoggingFromFlags(flags);
```

packages/astro/src/cli/install-package.ts
bluwy: Here's the related issue: https://github.com/nodejs/modules/issues/307. I'm pretty sure it would hit that unless you already have `@astrojs/check` installed somewhere 🤔

bluwy (approved):

packages/astro/src/cli/check/index.ts
```diff
@@ -1,386 +1,39 @@
...
-				bold(`Result (${this.#filesCount} file${this.#filesCount === 1 ? '' : 's'}): `),
-				bold(red(`${result.errors} ${result.errors === 1 ? 'error' : 'errors'}`)),
-				bold(yellow(`${result.warnings} ${result.warnings === 1 ? 'warning' : 'warnings'}`)),
-				dim(`${result.hints} ${result.hints === 1 ? 'hint' : 'hints'}\n`),
-			].join(`\n${dim('-')} `)
+	if (!checkPackage || !typescript) {
+		error(
+			logging,
+			'check',
+			'The `@astrojs/check` and `typescript` packages are required for this command to work. Please install them into your project and try again.'
```
ematipico, line 28: nit: I think the message is misleading. The `getPackage` method actually **tries** to install the package, so if it's `undefined` the cause could also be that, right?

packages/astro/src/cli/install-package.ts
```diff
@@ -0,0 +1,124 @@
...
+	].join(' ')} ${cyan(installCommand.dependencies.join(' '))}`;
+	const message = `\n${boxen(coloredOutput, {
+		margin: 0.5,
+		padding: 0.5,
+		borderStyle: 'round',
+	})}\n`;
+	info(
+		logging,
+		null,
+		`\n  ${magenta('Astro will run the following command:')}\n  ${dim(
```
ematipico, line 85: Should `message` be rendered after "...following command:"?

Princesseuh: I also thought that was weird while doing the code, but in the terminal it makes a bit more sense visually. This is also how it works for `astro add` (code was copied from there)

packages/astro/src/cli/check/index.ts
Princesseuh: Hmm, I don't think it's necessarily misleading, if the install fails, you'll get an error message from it before this message appear. I could add a "manually" in there though

matthewp (approved):
lgtm

PR #7900, opened 2023-08-01
Setup cleanup cache workflow

bluwy:
## Changes

Copied from https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

This will automatically cleanup the cache originating from a PR when the PR is closed or merged.

There doesn't seem to be a way to set the cache retention days. It's hardcoded to 7 days. Right now our caches are full in 4 days.

## Testing

None. I trust GitHub 🙈 

## Docs

n/a

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,35 @@
+name: Cleanup cache
+
+on:
+  pull_request:
+    types:
+      - closed
+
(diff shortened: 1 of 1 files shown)


ematipico:
I think our cache grows so much  because every time we create an integration test, the package lock is updated and a new cache is automatically generated.

Not sure how we can fix it though.

bluwy:
Yeah I don't think we can fix that without setting retention days too. But I think with this workflow, PRs should still be able to read the cache in `main` if the deps don't change.

ematipico (approved):

PR #7904, opened 2023-08-01
Refactor MDX components exports handling

bluwy:
## Changes

Move `export const components` [MDX feature](https://docs.astro.build/en/guides/markdown-content/#assigning-custom-components-to-html-elements) handling from Astro core to MDX integration.

## Testing

Existing Astro and MDX tests should pass.

## Docs

n/a. Should be transparent when upgraded to latest. Will note in the migration guide issue as so.

### DIFF
--- a/.changeset/loud-candles-admire.md
+++ b/.changeset/loud-candles-admire.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/mdx': patch
+---
+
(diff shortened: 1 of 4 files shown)


ematipico (approved):

natemoo-re (approved):
Nice! That's a great change.

PR #7918, opened 2023-08-02
feat: change `compressHTML` default value to `true`

ematipico:
## Changes

Changes the default value of `compressHTML` to `true`.

## Testing

I updated some tests to reflect the new default.

## Docs

### DIFF
--- a/.changeset/unlucky-ravens-type.md
+++ b/.changeset/unlucky-ravens-type.md
@@ -1,0 +1,5 @@
+---
+'astro': major
+---
+
+The property `compressHTML` is now `true` by default.
--- a/.changeset/unlucky-ravens-type.md
+++ b/.changeset/unlucky-ravens-type.md
@@ -2,4 +2,13 @@
 'astro': major
 ---
 
-The property `compressHTML` is now `true` by default.
+The property `compressHTML` is now `true` by default. Setting this value to `true` is no longer required.
+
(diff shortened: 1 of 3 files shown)


sarah11918:
If this happens, then we also need to update the types file so that this change is reflected in the docs configuration reference page here! https://github.com/withastro/docs/pull/3993

.changeset/unlucky-ravens-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
+---
+
+The property `compressHTML` is now `true` by default.
```
sarah11918, line 5:
``````suggestion
The property `compressHTML` is now `true` by default. Setting this value to `true` is no longer required.

If you do not want to minify your HTML output, you must set this value to `false` in `astro.config.mjs`.

```js
{
  compressHTML: false
}
```
``````
Especially since this is a change that reverses the behaviour of many existing projects, it is particularly important to provide guidance as to what people should *do* in response to this change in default to keep their original settings. 

(In general, it's *always* good to include any actions people might need to take in the changesets! This is where people will look to see what has changed, and what they might need to worry about, and docs only ever shows "how things are now.")

bluwy (approved):
LGTM besides Sarah's changesets review.

matthewp (approved):

PR #7919, opened 2023-08-02
feat: break down `AstroDataError` in multiple objects

ematipico:
## Changes

This PR removes the `AstroDataError` object and instead defines each error as an exported object.

Few things I had to change to adapt some logic:
- since we no longer have a top-level object, we cannot extract the name of the error at runtime. So I created a new `"name"` property, which is essentially the name of the error itself;
- `name` is now mandatory, so some `create*error` factory functions add that `"name"`. It's a bit redundant;
- the code that catches the error at runtime and shows the Astro overlay still needs to understand the name of the error, so I used a trick like `import * as AstroErrorData`. I tested locally with a project, and it still works;
- the CI job that checks documentation of the error now fails because the exported AST is different. I have PR ready, but I need help fixing this. Do you think we should turn off the action for now?

## Testing

- I updated some test case that was failing. The rest of the tests should pass.
- I manually used a project and intentionally created some code that fired an Astro error and checked that the overlays still showed all the relevant information.
- I manually built an edge middleware test case and checked that the bundle contained only the errors imported by the middleware. It only contains four objects, three shaking works :)

## Docs

I didn't add a changelog because this is an internal refactor. Should I create a changelog?


Princesseuh (commented):
Code looks good, unfortunate that this is needed, but I understand why ESBuild cannot tree shaking unused properties of objects safely.

Don't forget to update the README.md in the `src/core/errors` folder with the new `name` property!

packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -371,6 +371,7 @@ function stringifyEntryData(data: Record<string, any>): string {
 			});
 		} else {
 			throw new AstroError({
+				name: 'ContentImportsPluginError',
 				message: 'Unexpected error processing content collection data.',
 			});
```
Princesseuh, line 376: I'll note that this should be a defined error in error-data with perhaps common causes and stuff. But something to see in another PR perhaps.

The only time where throwing an AstroError that is not defined is ok is if the error is temporary (ex: experimental feature has not been enabled).

PR #7924, opened 2023-08-02
JSX refactor

matthewp:
## Changes

- Remove custom babel setup to support multiple JSX frameworks.
- Instead use each framework's own Vite plugins.
- To support multiple, new `include/exclude` objects lets you target only certain files.
- That's the only breaking change.

## Testing

- We had an existing test fixture for this, so it's just updated.

## Docs

- [3.0 migration guide](https://github.com/withastro/docs/issues/3756#issuecomment-1672025268)


package.json
```diff
@@ -76,6 +76,7 @@
     "astro-benchmark": "workspace:*"
   },
   "devDependencies": {
+    "@babel/plugin-transform-react-jsx": "^7.22.5",
```
matthewp, line 79: This was necessary because the Preact Vite plugin provides this to babel as a string. So babel must resolve it and doesn't know where to do so.

matthewp: Maybe @bluwy knows a workaround

packages/astro/src/core/config/settings.ts
```diff
@@ -95,7 +94,7 @@ export function createBaseSettings(config: AstroConfig): AstroSettings {
 				},
 			},
 		],
-		renderers: [jsxRenderer],
```
matthewp, line 98: No longer added by default, only used by the mdx plugin.

packages/astro/test/fixtures/jsx/src/components/preact/PreactCounter.tsx
```diff
@@ -7,6 +7,8 @@ export default function PreactCounter() {
   const add = () => setCount((i) => i + 1);
   const subtract = () => setCount((i) => i - 1);
 
+	debugger;
```
matthewp, line 10: remove

packages/integrations/preact/src/index.ts
```diff
@@ -1,96 +1,40 @@
...
 		hooks: {
 			'astro:config:setup': ({ addRenderer, updateConfig, command }) => {
-				const development = command === 'dev';
-				if (compat) addRenderer(getCompatRenderer(development));
-				addRenderer(getRenderer(development));
+				const preactPlugin = preact({include, exclude});
+
+				// If not compat, delete the plugin that does it
+				if(!compat) {
+					const pIndex = preactPlugin.findIndex(p => p.name == 'preact:config');
```
matthewp, line 23: This is a hack, otherwise Preact will try to turn on compat mode always, breaking usage next to React.

Should ask the Preact team to make this optional.

FredKSchott:
Amazing! This will be so huge for anyone building more complex interactive islands/SPAs with Astro!

What happens if a JSX file is unmatched/unclaimed by any integration? Is it possible that I could have one integration act as the "default"? This would map to a user who uses React for everything, except for a couple of Solid or Preact components that they have on the page.

```jsx
// Example:
// src/components/preact/*.jsx (the special preact components)
// src/components/**/*.jsx     (the rest of your codebase)
export default defineConfig({
  integrations: [
    preact({include: 'src/components/preact/**/*.jsx'}),
    react(),
  ]
})
```

packages/integrations/preact/src/index.ts
rschristian: Agreed! It's something I had ran into in the past but similarly spliced out. Opened https://github.com/preactjs/preset-vite/issues/78

matthewp:
I'm pretty sure this already works, if you don't use an include the Vite plugins don't filter anything out. So you need to put that integration last. I would expect your example to work as written. So i think this is a documentation thing.

package.json
bluwy: I don't think this will work in user's projects? They'd have to install it themselves too to workaround it like here.

One thing I did try in the past (which has a catch) is setting the babel [cwd option](https://babeljs.io/docs/options#cwd) to the preact plugin [babel options](https://github.com/preactjs/preset-vite/blob/3ed6fb180ed3845f26281675e45f06c1acbce906/src/index.ts#L129). Assigning it to the resolved path of `@preact/preset-vite` could work I think, so the plugins are resolved from that directory. The catch is that if we ever pass additional babel plugins, it won't work.

matthewp: I tried installing it into the test project and that didn't work. I guess I don't understand the problem properly because it feels like this would always be breaking for every user of this plugin and they would have fixed it by now, or there at least would be an issue with a lot of activity on it. But there's not, so that's why I thought it was related to our monorepo setup.

bluwy:
The example would work, but probably now ideal because the preact JSX file will be transformed twice by Preact and React plugins? It would be better to explicitly configure the includes.

matthewp:
@bluwy You mean for the Preact plugin? The React pass will occur after the JSX is already compiled away. So it's just a noop. Is that what you mean?

bluwy:
Yeah it's the noop that I'm concerned, since there's still effort and perf cost there to process it.

FredKSchott:
Cool, thanks for checking!

package.json
matthewp: This is fixed via `cwd` config.

matthewp:
This is ready for review. The lint error shown is wrong, eslint is drunk in CI (fine locally).

packages/astro/src/core/create-vite.ts
```diff
@@ -121,7 +121,8 @@ export async function createVite(
 			envVitePlugin({ settings }),
 			markdownVitePlugin({ settings, logging }),
 			htmlVitePlugin(),
-			jsxVitePlugin({ settings, logging }),
+			// jsxVitePlugin({ settings, logging }),
```
natemoo-re, line 124:
```suggestion
```

packages/astro/src/core/create-vite.ts
```diff
@@ -134,6 +135,7 @@ export async function createVite(
 			vitePluginSSRManifest(),
 			settings.config.experimental.assets ? [astroAssetsPlugin({ settings, logging, mode })] : [],
 			astroTransitions({ config: settings.config }),
+
```
natemoo-re, line 138:
```suggestion
```

packages/astro/src/vite-plugin-mdx/index.ts
```diff
@@ -0,0 +1,134 @@
...
+import {
+	transformWithEsbuild,
+	type Plugin,
+	type ResolvedConfig,
+} from 'vite';
+import type { AstroRenderer, AstroSettings } from '../@types/astro';
+import type { LogOptions } from '../core/logger/core.js';
+import type { PluginMetadata } from '../vite-plugin-astro/types';
+
+import babel from '@babel/core';
```
natemoo-re, line 11: Any chance we can drop `@babel/core` if this is the only place we're using it?

packages/integrations/solid/src/index.ts
```diff
@@ -34,30 +16,62 @@ async function getViteConfiguration(isDev: boolean, astroConfig: AstroConfig) {
 		 * We only need esbuild on .ts or .js files.
 		 * .tsx & .jsx files are handled by us
 		 */
-		esbuild: { include: /\.ts$/ },
+		//esbuild: { include: /\.ts$/ },
```
natemoo-re, line 19:
```suggestion
```

Can also remove the comment above this

.changeset/perfect-horses-tell.md
```diff
@@ -0,0 +1,25 @@
+---
+'astro': major
+---
+
+JSX refactor
+
+JSX in Astro has been refactored to better support each framework. In order to support multiple JSX frameworks at the same time, new `include` and `exclude` config options have been added to each integration to control which files to target. For example:
```
natemoo-re, line 7: Just a suggestion, but this wording could be a bit more precise IMO.

```suggestion
Astro's JSX handling has been refactored with better support for each framework.

Previously, Astro automatically scanned your components to determine which framework-specific transformations should be used. In practice, supporting advanced features like Fast Refresh with this approach proved difficult.

Now, rather than scanning your components, JSX framework integrations accept `include` and `exclude` config options. When using multiple JSX frameworks in the same project, users should manually control which files belong to each framework using the `include` and `exclude` options.
```

.changeset/perfect-horses-tell.md
```diff
@@ -0,0 +1,25 @@
...
+			include: ['**/react/*']
+		}),
+		solid({
+			include: ['**/solid/*'],
+		}),
+  ]
+});
+```
+
+This config is only needed in projects that use multiple JSX frameworks; if only using one no config is needed.
```
natemoo-re, line 25: Maybe this should be a comment in the code sample instead?

natemoo-re (commented):
Code looks great, super excited for this! Just a few small cleanup things.

natemoo-re:
Also could we add `// eslint:disable @typescript-eslint/no-unused-vars` to the `astro/packages/astro/src/core/build/plugins/plugin-analyzer.ts:1:23` file to silence that lint error?

packages/astro/src/vite-plugin-markdown/index.ts
```diff
@@ -115,8 +110,7 @@ export default function markdown({ settings, logging }: AstroPluginOptions): Plu
 				}
 
 				const code = escapeViteEnvReferences(`
-				import { Fragment, jsx as h } from ${JSON.stringify(astroJsxRuntimeModulePath)};
-				import { spreadAttributes } from ${JSON.stringify(astroServerRuntimeModulePath)};
+				import { unescapeHTML, spreadAttributes, createComponent, render, renderComponent } from ${JSON.stringify(astroServerRuntimeModulePath)};
```
bluwy, line 113: Great to see the separation with JSX!

packages/astro/src/vite-plugin-mdx/index.ts
```diff
@@ -0,0 +1,134 @@
...
+	logging: LogOptions;
+}
+
+// Format inspired by https://github.com/vitejs/vite/blob/main/packages/vite/src/node/constants.ts#L54
+const SPECIAL_QUERY_REGEX = new RegExp(
+	`[?&](?:worker|sharedworker|raw|url|${CONTENT_FLAG}|${PROPAGATED_ASSET_FLAG})\\b`
+);
+
+/** Use Astro config to allow for alternate or multiple JSX renderers (by default Vite will assume React) */
+export default function mdxVitePlugin({ settings }: AstroPluginJSXOptions): Plugin {
```
bluwy, line 89: In the future, could we move this MDX Vite plugin to `@astrojs/mdx`?

packages/integrations/preact/src/index.ts
```diff
@@ -1,94 +1,71 @@
...
+					viteConfig.optimizeDeps!.include!.push(
+						'preact/compat',
+						'preact/test-utils',
+						'preact/compat/jsx-runtime',
+					);
+					viteConfig.resolve = {
+						alias: [
+							{ find: 'react', replacement: 'preact/compat' },
+							{ find: 'react-dom/test-utils', replacement: 'preact/test-utils' },
+							{ find: 'react-dom', replacement: 'preact/compat' },
```
bluwy, line 53: I think we can remove these as it's [covered by the Preact plugin](https://github.com/preactjs/preset-vite/blob/4dfecb379c17fbb2f442987a8ff95536ff290cbd/src/index.ts#L184-L190)

packages/integrations/react/src/index.ts
```diff
@@ -1,5 +1,14 @@
...
 import { version as ReactVersion } from 'react-dom';
+import react, {type Options as ViteReactPluginOptions} from '@vitejs/plugin-react';
+
+const FAST_REFRESH_PREAMBLE = `
+import RefreshRuntime from '/@react-refresh'
+RefreshRuntime.injectIntoGlobalHook(window)
+window.$RefreshReg$ = () => {}
+window.$RefreshSig$ = () => (type) => type
+window.__vite_plugin_react_preamble_installed__ = true
+`;
```
bluwy, line 11: I think we can also get the preamble code from vite plugin react, like `react.preambleCode` ([source code](https://github.com/vitejs/vite-plugin-react/blob/4bebe5bd7c0267f6b088005293870cf69953b73a/packages/plugin-react/src/index.ts#L284)).

We also need to handle the `base` here so the preamble code is correctly loaded on different base. (Above the linked source code has an example)

packages/astro/src/vite-plugin-mdx/index.ts
matthewp: In favor of esbuild you mean?

packages/integrations/solid/src/index.ts
```diff
@@ -1,30 +1,12 @@
 import type { AstroConfig, AstroIntegration, AstroRenderer } from 'astro';
+import solid, { type Options as ViteSolidPluginOptions } from 'vite-plugin-solid';
 import { getSolidPkgsConfig } from './dependencies.js';
```
bluwy, line 3: We should be able to remove `getSolidPkgsConfig` and it's related dep (`vitefu`) since `vite-plugin-solid` also does the same internally. I brought over the logic from it initially.

packages/astro/src/vite-plugin-mdx/index.ts
matthewp: esbuild already runs. I'm not sure what implications there are if we remove this code to be honest.

packages/astro/src/vite-plugin-mdx/index.ts
matthewp: Yeah, I can try that. I'm not sure why this separation exists.

packages/integrations/solid/src/index.ts
matthewp: I think something was failing without this, but I'll give it another try.

packages/astro/src/vite-plugin-mdx/index.ts
bluwy: Cool, we can also do this as a follow-up if you prefer 😄

matthewp:
I think I've addressed everything I can here, would appreciate another review when people are ready.

.changeset/perfect-horses-tell.md
```diff
@@ -0,0 +1,27 @@
+---
+'astro': major
+---
+
+Astro's JSX handling has been refactored with better support for each framework.
+
+Previously, Astro automatically scanned your components to determine which framework-specific transformations should be used. In practice, supporting advanced features like Fast Refresh with this approach proved difficult.
+
+Now, rather than scanning your components, JSX framework integrations accept `include` and `exclude` config options. When using multiple JSX frameworks in the same project, users should manually control which files belong to each framework using the `include` and `exclude` options.
```
sarah11918, line 9:
```suggestion
Now, Astro determines which framework to use with `include` and `exclude` config options where you can specify files and folders on a per-framework basis. When using multiple JSX frameworks in the same project, users should manually control which files belong to each framework using the `include` and `exclude` options.
```
Is this correct? I feel like it's helpful to call out these config values. 

Also, Is this just an array? Can there be more than one item in an `include` or `exclude`? Can it be an individual file instead of a path? Can you BOTH include and exclude things for a framework? I think having one of these be a more complex example will be helpful at least by the time we get to the migration guide, so I'll want to do that in the Docs Issue, but I'll let you decide whether that's overloading this changeset!

matthewp: Yes, there can be any number of items in the include/exclude. Yes, you can add both. I didn't add an `exclude` example as I don't know a great use-case for it. But it's something that's provided by each framework plugin so we are just exposing that to the user.

packages/integrations/solid/src/index.ts
```diff
@@ -1,63 +1,72 @@
...
 	return {
 		name: '@astrojs/solid-js',
 		hooks: {
 			'astro:config:setup': async ({ command, addRenderer, updateConfig, config }) => {
 				addRenderer(getRenderer());
-				updateConfig({ vite: await getViteConfiguration(command === 'dev', config) });
+				updateConfig({
+					/*vite: {
+						plugins: [solid({ include, exclude, dev: command === 'dev', ssr: true })]
+					}*/
```
bluwy, line 67: I think we can remove this?

packages/integrations/solid/src/index.ts
```diff
@@ -1,63 +1,72 @@
...
 	// https://github.com/solidjs/vite-plugin-solid
 	// We inject the dev mode only if the user explicitly wants it or if we are in dev (serve) mode
 	const nestedDeps = ['solid-js', 'solid-js/web', 'solid-js/store', 'solid-js/html', 'solid-js/h'];
-	const solidPkgsConfig = await getSolidPkgsConfig(!isDev, astroConfig);
 	return {
 		/**
 		 * We only need esbuild on .ts or .js files.
 		 * .tsx & .jsx files are handled by us
 		 */
-		esbuild: { include: /\.ts$/ },
```
bluwy, line 37: This comment also seems to be not needed now.

packages/integrations/react/src/index.ts
```diff
@@ -73,13 +55,18 @@ function getViteConfiguration() {
...
 	return {
 		name: '@astrojs/react',
 		hooks: {
-			'astro:config:setup': ({ addRenderer, updateConfig }) => {
+			'astro:config:setup': ({ config, command, addRenderer, updateConfig, injectScript }) => {
 				addRenderer(getRenderer());
-				updateConfig({ vite: getViteConfiguration() });
+				updateConfig({ vite: getViteConfiguration({include, exclude}) });
+				if (command === 'dev') {
+					const preamble = FAST_REFRESH_PREAMBLE.replace(`__BASE__`, config.base)
```
bluwy, line 67: We might need to ensure `config.base` has a trailing slash here so it generates the right import path.

packages/integrations/preact/src/index.ts
```diff
@@ -1,94 +1,68 @@
...
-			return {
-				plugins: [jsx({}, { runtime: 'automatic', importSource: 'preact' })],
-			};
-		},
 	};
 }
 
-function getCompatRenderer(development: boolean): AstroRenderer {
+export type Options =Pick<VitePreactPluginOptions, 'include' | 'exclude'> & { compat?: boolean };
+// TODO: Add back compat support -- how would this work in the new world?
```
bluwy, line 16: I think the plugin now has compat support? Or are they missing features with this refactor?

bluwy (approved):
Some minor nits, but looks good overall!

packages/integrations/preact/src/index.ts
matthewp: Nope it has it, will remove the comment.

ematipico (approved):

PR #7926, opened 2023-08-02
Make the image decorative

mgifford:
## Changes

- This just makes the background icon clearly decorative

## Testing

Just need to see that `aria-hidden="true" focusable="false"` was added to the SVG file in the page.

## Docs

Should have no impacts on behavior. Just implementing a best practice to ensure that images are clearly marked as decorative.


examples/basics/src/pages/index.astro
```diff
@@ -12,6 +12,7 @@ import Card from '../components/Card.astro';
 			viewBox="0 0 495 623"
 			fill="none"
 			xmlns="http://www.w3.org/2000/svg"
+      aria-hidden="true" focusable="false"
```
natemoo-re, line 15:
```suggestion
      aria-hidden="true"
```

> `focusable="false"` is also used to ensure Internet Explorer won’t allow the Tab key to navigate into the SVG.

I don't think we need to be concerned with IE support since it's EOL.

natemoo-re (approved):
Thanks so much, this is great!

PR #7939, opened 2023-08-03
Handle error state for version

natemoo-re:
## Changes

- We noticed that `npm create astro` sometimes printed `vundefined`, which is embarrasing
- This adds an error handler to ensure that we don't print the version if the lookup fails

## Testing

Manually. It works!

## Docs

N/A


TheOtterlord:
I can't leave a comment directly for this part of the file, but does [this line](https://github.com/withastro/astro/pull/7939/files#diff-d871d8733c58908341170a58707fccc268e650dd6685b9aeb9d6dac387cb08f3L96) also need updating?

TheOtterlord (approved):
Looks good!

Issue #7470, opened 2023-06-24
tsconfig: `compilerOptions.baseUrl` should probably be "./" ?


hirasso:
### What version of `astro` are you using?

2.7.0

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Irrelevant

### Describe the Bug

This is not necessary a bug, but maybe it is one. I need to use `compilerOptions.paths` for my project. This is what I want:

```json
{
  "extends": "astro/tsconfigs/strict",
  "compilerOptions": {
    "strictNullChecks": true,
    "noImplicitAny": false,
    "paths": {
      "@packages/*": ["packages/*"],
    },
  }
}
```

When I try to import one of my packages, vscode complains:

```js
import MyPackage from "@packages/my-package/src/index.js";
```

![Screenshot 2023-06-24 at 16 03 19](https://github.com/withastro/astro/assets/869813/a1a24672-ae1d-4771-92a9-bcdee7a61f4e)

It seems like this is caused by the `baseUrl` being set to `.` in [astro/tsconfigs/base.json](https://github.com/withastro/astro/blob/main/packages/astro/tsconfigs/base.json#L22).

### What's the expected result?

When I change `compilerOptions.baseUrl` to `./` like so:

```json
{
  "extends": "astro/tsconfigs/strict",
  "compilerOptions": {
    "strictNullChecks": true,
    "noImplicitAny": false,
    "baseUrl": "./",
    "paths": {
      "@packages/*": ["packages/*"],
    },
  }
}
```

...it starts to work as expected:

![Screenshot 2023-06-24 at 16 03 01](https://github.com/withastro/astro/assets/869813/85f10fce-d212-40f6-a3cc-58fe3218a8cb)

### Link to Minimal Reproducible Example

none needed

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
I think it's because the `baseUrl` in `base.json` refers to that directory itself, e.g. `/node_modules/astro/tsconfig/`. It was added for the `~/assets/` alias but I'm not sure if it ever worked in that case 🤔 Maybe we should remove it and the preconfigured `paths` directly.

natemoo-re:
Would love @Princesseuh to weigh in on this. Notably the `~/assets` alias was part of our [RFC for `astro:assets`](https://github.com/withastro/roadmap/blob/main/proposals/0030-core-image-story.md#srcassets)

[...]

PR #7940, opened 2023-08-03
Removed ~/assets alias and baseUrl from tsconfig presets.

kanashimia:
## Changes

Removed ~/assets alias and baseUrl from tsconfig presets.
Instead of predefining ~/assets and baseUrl in our tsconfig presets user should now define them in their own in their tsconfig.json, as predefined aliases were too opinionated and didn't work without modifying tsconfig anyways.

By opinionated i mean that it not only forces a specific location, but also a specific prefix, some users would want to use assets like `@assets/blah.png`, it just doesn't make much sense to define this in the preset.

Original justification for defining src/assets as a stable directory was in https://github.com/withastro/roadmap/discussions/468#discussioncomment-4882421
Justification for defining ~/assets as an alias was in https://github.com/withastro/roadmap/discussions/468#discussioncomment-4882397

User has to add baseUrl into their tsconfig for this to work anyway, and this only adds alias for ~/assets and not ~/content or whatever, it just doesn't make much sense, src/assets isn't really any more special than src/content.

Fixes https://github.com/withastro/astro/issues/7470 (you should also read discussion there)

## Testing

No tests were added.
tsconfig.json was added at the base dir of all tests that use ~/assets alias, they weren't removed as it may be useful to test if aliases regular work there. 
Do you think tests (relevant parts) should be removed instead?

Tested with `pnpm --filter astro run test`

## Docs

Builtin ~/assets alias in the astro assets example should be removed or modified in the docs.

In theory this change can affect users that don't use astro:assets, only if they had `baseUrl` set in their tsconfig, i doubt it matters.

### DIFF
--- a/.changeset/happy-jokes-drive.md
+++ b/.changeset/happy-jokes-drive.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Remove ~/assets alias and baseUrl from tsconfig presets
--- a/packages/astro/src/assets/types.ts
+++ b/packages/astro/src/assets/types.ts
@@ -86,7 +86,7 @@ export type LocalImageProps<T> = ImageSharedProps<T> & {
 	 *
 	 * **Example**:
 	 * ```js
-	 * import myImage from "~/assets/my_image.png";
+	 * import myImage from "../assets/my_image.png";
 	 * ```
 	 * And then refer to the image, like so:
 	 * ```astro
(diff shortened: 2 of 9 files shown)


Princesseuh:
Superseded by https://github.com/withastro/astro/pull/8113, sorry for not using your PR, I didn't realize you had everything done already. I added you as a co-author of the commit so you'll have proper credit on your profile

PR #7945, opened 2023-08-03
Fix race condition when performing swap for fallback

matthewp:
## Changes

- In fallback mode for the router we need to wait for animations (if there are any) and also setTimeout (in case there are none). This causes a race condition that can case swap() to be called twice.
- Fix is to remove the event listener and the timeout whenever the first wins.

## Testing

- Tested manually. Maybe we should think about adding Firefox as a test runner some how.

## Docs

Bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -168,13 +168,18 @@ const { fallback = 'animate' } = Astro.props as Props;
 
 			// Trigger the animations
 			document.documentElement.dataset.astroTransitionFallback = 'old';
-			doc.documentElement.dataset.astroTransitionFallback = 'new';
+			const fallbackSwap = () => {
+				removeEventListener('animationend', fallbackSwap);
+				clearTimeout(timeout);
+				swap();
+				document.documentElement.dataset.astroTransitionFallback = 'new';
+			};
 			// If there are any animations, want for the animationend event.
-			addEventListener('animationend', swap, { once: true });
+			addEventListener('animationend', fallbackSwap, { once: true });
 			// If there are no animations, go ahead and swap on next tick
 			// This is necessary because we do not know if there are animations.
 			// The setTimeout is a fallback in case there are none.
-			setTimeout(() => !isAnimating && swap());
+			var timeout = setTimeout(() => !isAnimating && fallbackSwap());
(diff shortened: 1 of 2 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -168,13 +168,18 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+				document.documentElement.dataset.astroTransitionFallback = 'new';
+			};
 			// If there are any animations, want for the animationend event.
-			addEventListener('animationend', swap, { once: true });
+			addEventListener('animationend', fallbackSwap, { once: true });
 			// If there are no animations, go ahead and swap on next tick
 			// This is necessary because we do not know if there are animations.
 			// The setTimeout is a fallback in case there are none.
-			setTimeout(() => !isAnimating && swap());
+			var timeout = setTimeout(() => !isAnimating && fallbackSwap());
```
matthewp, line 182: This is a valid use of `var`!

bluwy: We could also hoist up a `let` 😄 `var` would bleed out of this if block.

bluwy (approved):

packages/astro/components/ViewTransitions.astro
ematipico: I also think that hoisting a `let` is more explicit. If it's impossible, we should leave a comment explaining why `var` needs to be used. E.g.: [like TS does](https://github.com/microsoft/TypeScript/blob/83f3abda153f0f567e296a3c75cdca37c7eb065d/src/compiler/sourcemap.ts#L38)

ematipico (approved):

packages/astro/components/ViewTransitions.astro
matthewp: Oh my I totally forgot that a let does hoist too 🤦‍♂️

matthewp: Ok fixed. Let's never speak of this again.

ethan-ou:
Thanks @matthewp for this one. I tried using ViewTransitions in 2.10.1 and on Firefox some pages returned a 'null' instead of loading the body of the page. I think this update has resolved it.

PR #7946, opened 2023-08-04
Fix missing CSS when 404 server Response redirects to a custom 404 page

### LINKED ISSUE #7910: 404 page missing CSS when 404 server Response redirects to static custom 404 page

2.9.7

Vercel

pnpm

Mac

Chrome, Safari

### Describe the Bug

Perhaps a bit of an edge case, but with the following setup:
* server mode
* dynamic route (alphabetically first priority) `[...ssrPath].astro` redirects non-existing pages to 404 using `new Response` but does no page rendering
* dynamic route (alphabetically second priority) `[...staticPath].astro` renders pages statically (includes `export const prerender = true`)
*  custom `/pages/404.md`

The issue is: When deployed on Vercel, navigating to a non-existent page renders the custom 404 page, but without CSS.

Additional notes:
* Going directly to the 404 page _does_ successfully include CSS
* If the `Response` is changed to an Astro redirect (`return Astro.redirect('/404')`), CSS is included in the 404 page after the redirect
* This issue does not occur locally. With `vercel dev` or `astro dev`, CSS is included after the 404 Response.

CSS should be included in the custom 404 page rendered after a 404 `Response`

[repo](https://github.com/paulrudy/astro-404-test.git)
[demo](https://astro-404-test.vercel.app)

### Participation

andremralves:
## Changes

Closes #7910

The regression  was introduced here #7703 because of a typo.

This PR changes the variable used in `#createRenderContext` back to the correct one: `errorRouteData`.

The change was simple (but really hard to find) and I also added some tests to prevent this error from happening again.

## Testing

One test to check imports in 404 redirections to a custom 404 page using `contentCollections`. 

## Docs

n/a


bluwy (approved):
Thanks for fixing this. Besides the merge conflicts, this lgtm

natemoo-re:
Good catch! I'm guessing the conflicts are coming from https://github.com/withastro/astro/pull/7754 which was merged quite recently. Please check out that PR. It _might_ have fixed this issue already?

andremralves:
> Good catch! I'm guessing the conflicts are coming from #7754 which was merged quite recently. Please check out that PR. It _might_ have fixed this issue already?

I also thought your PR might have fixed the issue, so I made some tests, but the problem was still showing up.

I believe that the line `const finalRouteData = routeData ?? errorRouteData;` can be the problem, shouldn't we always use `errorRouteData` to create the render context as `const mod` will aways use `errorRouteData`?

Because in my tests, the problem was being caused because a route other than `/404` was being passed to `#createRenderContext` inside the `finalRouteData` constant.

I just could reproduce this error using ssr with `contentCollections`. An example can be found in the tests that I made.

andremralves:
I made a new commit merging the branches and changing a little bit the behavior of `#renderError`, I believe the issue is fixed now. But a careful review would be good.

paulrudy:
Thanks for finding and addressing this!

PR #7957, opened 2023-08-04
chore: update prettier to `3.0`

ematipico:
## Changes

It updates prettier to `3.0`

## Testing

I run locally the script that formats the code, and it works as expected.

## Docs

N/A

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -6,7 +6,7 @@ module.exports = {
   tabWidth: 2,
   trailingComma: 'es5',
   useTabs: true,
-  plugins: ['./node_modules/prettier-plugin-astro/dist/index.js'],
+  plugins: ['prettier-plugin-astro'],
   overrides: [
     {
       files: ['.*', '*.json', '*.md', '*.toml', '*.yml'],
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -6,7 +6,7 @@ module.exports = {
   tabWidth: 2,
   trailingComma: 'es5',
   useTabs: true,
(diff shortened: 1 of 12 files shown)


.prettierrc.cjs
```diff
@@ -6,7 +6,7 @@ module.exports = {
   tabWidth: 2,
   trailingComma: 'es5',
   useTabs: true,
-  plugins: ['./node_modules/prettier-plugin-astro'],
+  plugins: ['./node_modules/prettier-plugin-astro/dist/index.js'],
```
bluwy, line 9: I think we can also simply specify `plugins: ['prettier-plugin-astro']` in Prettier 3

packages/astro/astro-jsx.d.ts
```diff
@@ -9,7 +9,7 @@
  * Adapted from React’s TypeScript definition from DefinitelyTyped.
  * @see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts
  */
-declare namespace astroHTML.JSX {
+namespace astroHTML.JSX {
```
bluwy, line 12: Is the `declare` removal caused by prettier?

.prettierrc.cjs
Princesseuh: Yes, I would recommend to do this actually, it's better

packages/astro/astro-jsx.d.ts
ematipico: Yes. The files changed are caused by running prettier locally (which is good, it seems we caught a bug?)

bluwy (approved):

PR #7960, opened 2023-08-04
docs: Update Lit README.md

ElianCodes:
## Changes

Following feedback acquired via Feedback Fish, updated the Lit integration README to be a bit more clear on what the difference between using tagName vs import is.

## Testing

only docs change

## Docs

- closes https://github.com/withastro/docs/issues/4041


packages/integrations/lit/README.md
```diff
@@ -63,33 +63,38 @@ To use your first Lit component in Astro, head to our [UI framework documentatio
...
-> Note that exporting the `tagName` is **required** if you want to use the tag name in your templates. Otherwise you can export and use the constructor, like with non custom element frameworks.
-
-In your Astro template import this component as a side-effect and use the element.
+Now, the component can be used via the element tag name
 
 ```astro
+// src/pages/index.astro
 ---
+---
+<!-- note: this way, you can't use client: directives -->
```
TheOtterlord, line 87: Could this be a note on the page itself? Explaining why this matters (won't have client-side interactivity)

packages/integrations/lit/README.md
```diff
@@ -63,33 +63,38 @@ To use your first Lit component in Astro, head to our [UI framework documentatio
...
+<my-element />
```
TheOtterlord, line 88: I tried this myself, and found it didn't work without an import. It does work briefly when removing an existing import, but restarting the dev server (or building) will reset this behaviour.

TheOtterlord (commented):
Left a few comments

packages/integrations/lit/README.md
ElianCodes: Oh damn, you're right! So we just should remove this example in its totality 🤔

ElianCodes: The funny thing is that it does work when you already put the component on the page using the import:

```astro
---
import {MyCounter} from 'my-counter.js'
---
<!-- both work -->
<MyCounter client:load/>
<my-counter/>
```

TheOtterlord: It does also work with a nameless import
```astro
---
import 'my-counter.js'
---
```

ElianCodes: yeah, I noted that too, that way, using the tag name works `<my-counter>` but not with `client:` directives 🤔

TheOtterlord:
Yeah, I think I'd agree. While I don't know much about Lit, this is the way all our integrations work. I do wonder if it's worth documenting though, just for the ability to, e.g. import them in a layout, and use them in any components on your site. But that sort-of falls apart since you can't add interactivity.

We don't happen to have a Lit fan on the team, do we? :sweat_smile:

ElianCodes:
Yup! exactly my thinking!

If both you and I agree, I think we can merge this PR.

We can always revisit this a little later to update with the other example using an unamed import and tagname.

TheOtterlord (approved):
Yep! Sounds good to me

PR #7964, opened 2023-08-04
Add URL to RSSOptions.site type

### LINKED ISSUE #7951: RSS: wrong site type

2.10.1

Vercel

pnpm

Windows

Brave

### Describe the Bug

When using `@astrojs/rss`, the `rss` function has a parameter `site` of type string. However, it expects a `URL` (from `context.site`) so we currently have to put `@ts-ignore` above.

Fix the type from `string` to `URL`

https://stackblitz.com/edit/github-b8xj5x?file=package.json,src%2Frss.xml.ts&on=stackblitz

### Participation

DerTimonius:
## Changes

- This PR adds the URL type to `RSSOptions.site`
- `site` was only of type `string` before and lead to TS errors
- Closes #7951 
- Ran `pnpm exec changeset`

## Testing

- Did manual testing in TS Playground to check for the best option (wanted to use zod instead of URL directly, but did not really work)
- Existing tests still run as expected

## Docs

- No changes in the docs necessary

### DIFF
--- a/.changeset/new-otters-sell.md
+++ b/.changeset/new-otters-sell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/rss': patch
+---
+
+Add URL to RSSOptions.site type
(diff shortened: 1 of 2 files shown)


bluwy (approved):

bluwy:
Thanks!

PR #7965, opened 2023-08-04
chore: fix issue in the check.js script

ematipico:
## Changes

For some reason sometimes `.config` is undefined.

## Testing

Run it locally, it works!

## Docs

N/A


bluwy (approved):
I haven't seen the issue before, but also make sense to guard it since the types mention it could be undefined.

PR #7975, opened 2023-08-06
Add Incremental Static Regeneration support for the Netlify's on-demand builders

### LINKED ISSUE #6654: Add TTL support for Netlify adapter for On Demand Builders

2.1.5

Netlify

### Describe the Feature

The Netlify On Demand Builders adapter should support Time To Live (TTL):

https://docs.netlify.com/configure-builds/on-demand-builders/#time-to-live-ttl

Relevant PR: #5874 

### Participation

- [X] I am willing to submit a pull request for this issue (but I won't be hurt if someone beats me to it)

lilnasy:
## Changes
- Closes #6654 
- Previous Effort #6219 
- Exposes Netlify on-demand builder's [ttl](https://docs.netlify.com/configure-builds/on-demand-builders/#time-to-live-ttl) configuration as a local
- Usage
image

- Adding types
image

- Available only when [builders mode](https://docs.astro.build/en/guides/integrations-guide/netlify/#builders) is enabled.

## Testing
Added a fixture
image

## Docs
Added to the package README
image


packages/integrations/netlify/package.json
```diff
@@ -23,6 +23,7 @@
     "./netlify-functions.js": "./dist/netlify-functions.js",
     "./edge-functions": "./dist/integration-edge-functions.js",
     "./netlify-edge-functions.js": "./dist/netlify-edge-functions.js",
+    "./builders": "./builders.d.ts",
```
bluwy, line 26: Maybe this can be `/builders-types` to not mix up with the other `/functions` and `/edge-functions` variants of adapters.
```suggestion
    "./builders-types": "./builders.d.ts",
```

packages/integrations/netlify/src/netlify-functions.ts
```diff
@@ -99,6 +102,8 @@ export const createExports = (manifest: SSRManifest, args: Args) => {
 			headers: responseHeaders,
 			body: responseBody,
 			isBase64Encoded: responseIsBase64Encoded,
+			// @ts-expect-error Property 'netlify' does not exist on type '{}'
+			ttl: locals.netlify?.builders?.ttl,
```
bluwy, line 106: Could we parse the `Cache-Control` here instead of exposing a new API? Looking at your ISR experiments and docs, that seems to be supported for Vercel and Cloudflare ootb, it would be nice to keep the same syntax for Netlify, but glueing this special `ttl` handling here.

ematipico (changes requested):
`locals` in an Astro component are read-only properties and you won't be able to set anything like this. The only place where you can write stuff in the locals is inside the middleware.

natemoo-re:
> `locals` in an Astro component are read-only properties and you won't be able to set anything like this. The only place where you can write stuff in the locals is inside the middleware.

Ah, that is a very important limitation of this API! Should we use a `Proxy` during `astro dev` that throws an error when users attempt to assign a value via `set()`?

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: That's a great point! I agree it would be nice to have a uniform approach for all the adapters, but that approach should be the subject of an RFC.

lilnasy:
There is no official word of this outside of Netlify's API reference, but most of on-demand builders' functionality has been brought into other functions.

I found this out from a [Netlify support thread](https://answers.netlify.com/t/netlify-function-with-query-strings-ignores-custom-cache-control-header/15390/62), and the gist of it is, its use case has become niche: single cache store used globally - instead of per-region, as it is for normal and edge functions.

This PR could still be merged for feature-completeness. I will be updating the Netlify part of the ISR guide to use `Cache-Control` headers instead.

ematipico:
> @ematipico It works. Do you mean it _should be_ read-only?

@lilnasy 

Yes. `Astro.locals` **must be** read-only. That's what `dev` and node adapter do in the first place. Making `Astro.locals` writeable in the netlify would cause different behaviour in the APIs. I understand that it works, but that's not how the APIs should be crafted. 

`locals` are meant to be written in the middleware layer. 

> > `locals` in an Astro component are read-only properties and you won't be able to set anything like this. The only place where you can write stuff in the locals is inside the middleware.
> 
> Ah, that is a very important limitation of this API! Should we use a `Proxy` during `astro dev` that throws an error when users attempt to assign a value via `set()`?

We don't use `Proxy` in core, but yeah, that's another way to do it.

packages/integrations/netlify/builders-types.d.ts
```diff
@@ -0,0 +1,9 @@
+declare namespace App {
+    interface Locals {
+        netlify: {
+            builders: {
+                ttl: number | undefined
+            }
+        }
+    }
+}
```
ematipico, line 9: I don't thinks this is the correct type we should export. 

When users want to type their `locals`, they do:

```ts
/// <reference types="astro/client" />
declare namespace App {
	interface Locals {
		user: {
			name: string;
			surname: string;
		};
	}
}
```

So I think we should export an interface like: 

```ts
export interface NetlifyLocals {
	netlifyBuilldersTtl: number | undefined
}
```

And the user just do :
```ts
/// <reference types="astro/client" />
declare namespace App {
	interface Locals extends NetlifyLocals {
		user: {
			name: string;
			surname: string;
		};
	}
}
```

packages/integrations/netlify/src/netlify-functions.ts
```diff
@@ -80,6 +80,9 @@ export const createExports = (manifest: SSRManifest, args: Args) => {
 				locals = JSON.parse(localsAsString);
 			}
 		}
+		if (builders) {
+			Object.assign(locals, { netlify: { builders: { ttl : undefined } } });
```
ematipico, line 84: I don't think there's a specific need to create such a deep object for **one value**. What about 

```js
{
	netlifyBuildersTtl: 45
}
```

packages/integrations/netlify/builders-types.d.ts
lilnasy: I like this API! I'll look into it.

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: I am concerned that this is prone to deprecation. For example, what if we introduce more runtime-specific controls that would benefit from a namespace.

lilnasy:
@ematipico Considering that serializability restrictions are relaxed, maybe a `setTtl()` function is more appropriate.
```astro
---
Astro.locals.netlify.builders.setTtl(45)
---
<h1>Astro</h1>
```

ematipico:
Still, I think we should remove this API via locals, and pass the value when initializing the adapter:

- Your example gives the impression that the value is per route, while it isn't

packages/integrations/netlify/src/netlify-functions.ts
ematipico: I would say that's not a concern because all these objects are controlled by the adapter, and we can have only one adapter at a time. 

Maybe `netlify.buildersTtl` should be enough.

ematipico:
> @ematipico Considering that serializability restrictions are relaxed, maybe a `setTtl()` function is more appropriate.
> 
> ```astro
> ---
> Astro.locals.netlify.builders.setTtl(45)
> ---
> <h1>Astro</h1>
> ```

Yeah that's way better I'd say

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: I agree, going with `netlify.setBuildersTtl()`.

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: I am hesitant because parsing the `Cache-Control` header is not without its caveats.

Builders only support server-side control, the closest directive is `s-maxage`, and the others would not just be thrown out but replaced with `public, max-age=0, must-revalidate`. Users might find it confusing to have their directives ignored or even inverted.

Also, I want to leave room for an RFC that abstracts over cache headers (and other ISR methods) similar to what Next and Nuxt have. By abstracting parsing the header, we are implying that `Cache-Control` _is_ the abstraction.

lilnasy:
I would like a pointer for adding a no-op for dev.

I see dev mode support was added in #7385, but the two tests added in it are for build.

ematipico:
> I would like a pointer for adding a no-op for dev.
> 
> I see dev mode support was added in #7385, but the two tests added in it are for build.

Yeah, the PR title is misleading (it was picked up from a third-party contributor). 

I am not sure how we can fix this case, with the code we have now. I think Astro should provide a **safe** way to retrieve the "adapter runtime", and if the adapter doesn't provide any runtime, it should be undefined.

E.g.

```js
const runtime = Astro.locals.getAdapterRuntime();
if (runtime) {
	runtime.setBuilderTtl(45)
}
```

lilnasy:
Should I add a `import.meta.env.PROD` check in the docs for now?

matthewp (approved):

ematipico (approved):
Looks good!

packages/integrations/netlify/src/netlify-functions.ts
bluwy: Alright. Thanks for the explanation. In that case we can go with the PR's approach first and we can revisit this later.

packages/integrations/netlify/builders-types.d.ts
lilnasy: Implemented

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: Implemented

packages/integrations/netlify/src/netlify-functions.ts
lilnasy: Yeah, I agree

packages/integrations/netlify/README.md
```diff
@@ -162,6 +162,30 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
 > **Note**
 > You can still include a `public/_redirects` file for manual redirects. Any redirects you specify in the redirects config are appended to the end of your own.
 
+### On-demand Builders
+
+[Netlify On-demand Builders](https://docs.netlify.com/configure-builds/on-demand-builders/) are serverless functions used to generate web content as needed that’s automatically cached on Netlify’s Edge CDN. You can enable their use using the [`builders` configuration](#builders).
+
+By default, all pages will be rendered on first visit and the rendered result will be reused for every subsequent visit until you redeploy. To set a revalidation time, call the `runtime.setBuildersTtl(ttl)` [local](https://docs.astro.build/en/guides/middleware/#locals) with the duration (in seconds) for which the page should be reused.
```
ElianCodes, line 169:
```suggestion
By default, all pages will be rendered on first visit and the rendered result will be reused for every subsequent visit until you redeploy. To set a revalidation time, call the [`runtime.setBuildersTtl(ttl)` local](https://docs.astro.build/en/guides/middleware/#locals) with the duration (in seconds).
```

ElianCodes (commented):
A couple of small changes, but after applying, good to go!

packages/integrations/netlify/README.md
```diff
@@ -162,6 +162,30 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
 > **Note**
 > You can still include a `public/_redirects` file for manual redirects. Any redirects you specify in the redirects config are appended to the end of your own.
 
+### On-demand Builders
+
+[Netlify On-demand Builders](https://docs.netlify.com/configure-builds/on-demand-builders/) are serverless functions used to generate web content as needed that’s automatically cached on Netlify’s Edge CDN. You can enable their use using the [`builders` configuration](#builders).
```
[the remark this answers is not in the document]
lilnasy: This doesn't sound right, can you check if this is the change you wanted to suggested?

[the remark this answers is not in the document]
lilnasy: Alright, good to go.

PR #7979, opened 2023-08-07
Export JS API

bluwy:
## Changes

Export `dev`, `build`, `preview`, and `sync` JavaScript APIs.

## Testing

Existing tests updated to use the entrypoint to test it.

## Docs

I copied each APIs jsdoc from the [CLI reference docs](https://docs.astro.build/en/reference/cli-reference/#astro-dev) because I ~~am lazy~~ wanted to keep it consistent with what we have now.

We'll likely need a dedicated page for it as we add more APIs in the future (e.g. there was discussion to move `defineConfig` to the main entrypoint), but it could be weird to have two places referring to similar info?

Appreciate some feedback on this, cc @withastro/maintainers-docs!

### DIFF
--- a/.changeset/many-pears-explode.md
+++ b/.changeset/many-pears-explode.md
@@ -1,0 +1,5 @@
+---
+'astro': major
+---
+
+Export `dev`, `build`, `preview`, and `sync` APIs from `astro`. This replaces the previous entrypoint code which runs the Astro CLI programmatically.
--- a/packages/astro/index.d.ts
+++ b/packages/astro/index.d.ts
@@ -1,0 +1,2 @@
+export * from './dist/@types/astro.js'
+export * from './dist/core/index.js'
--- a/packages/astro/src/core/README.md
+++ b/packages/astro/src/core/README.md
@@ -1,5 +1,7 @@
 # `core/`
 
-Code that executes within the top-level Node context. Contains the main Astro logic for the `build` and `dev` commands, and also manages the Vite server and SSR.
+Code that executes within the top-level Node context. Contains the main Astro logic for the `build`, `dev`, `preview`, and `sync` commands, and also manages the Vite server and SSR.
+
+The `core/index.ts` file is the main entrypoint for the `astro` package.
 
 [See CONTRIBUTING.md](../../../../CONTRIBUTING.md) for a code overview.
--- a/packages/astro/src/core/build/index.ts
+++ b/packages/astro/src/core/build/index.ts
@@ -37,15 +37,18 @@ export interface BuildOptions {
 	/**
 	 * Teardown the compiler WASM instance after build. This can improve performance when
 	 * building once, but may cause a performance hit if building multiple times in a row.
+	 *
+	 * @internal only used for testing
+	 * @default true
 	 */
 	teardownCompiler?: boolean;
 }
 
-/** `astro build` */
-export default async function build(
-	inlineConfig: AstroInlineConfig,
-	options: BuildOptions
-): Promise<void> {
+/**
+ * Builds your site for deployment. By default, this will generate static files and place them in a dist/ directory.
+ * If SSR is enabled, this will generate the necessary server files to serve your site.
+ */
+export default async function build(inlineConfig: AstroInlineConfig): Promise<void> {
 	applyPolyfill();
 	const logging = createNodeLogging(inlineConfig);
 	const { userConfig, astroConfig } = await resolveConfig(inlineConfig, 'build');
@@ -53,8 +56,12 @@ export default async function build(
 
 	const settings = createSettings(astroConfig, fileURLToPath(astroConfig.root));
 
+	// Use this hack to prevent the build options being public API
+	// eslint-disable-next-line prefer-rest-params
+	const internalOptions = arguments[1] as BuildOptions | undefined;
+
 	const builder = new AstroBuilder(settings, {
-		...options,
+		...internalOptions,
 		logging,
 		mode: inlineConfig.mode,
 	});
@@ -82,7 +89,7 @@ class AstroBuilder {
 		}
 		this.settings = settings;
 		this.logging = options.logging;
-		this.teardownCompiler = options.teardownCompiler ?? false;
+		this.teardownCompiler = options.teardownCompiler ?? true;
 		this.routeCache = new RouteCache(this.logging);
 		this.origin = settings.config.site
 			? new URL(settings.config.site).origin
--- a/packages/astro/src/core/sync/index.ts
+++ b/packages/astro/src/core/sync/index.ts
@@ -22,8 +22,7 @@ export type ProcessExit = 0 | 1;
 
 export type SyncOptions = {
 	/**
-	 * Only used for testing
-	 * @internal
+	 * @internal only used for testing
 	 */
 	fs?: typeof fsMod;
 };
@@ -32,10 +31,11 @@ export type SyncInternalOptions = SyncOptions & {
 	logging: LogOptions;
 };
 
-export async function sync(
-	inlineConfig: AstroInlineConfig,
-	options?: SyncOptions
-): Promise<ProcessExit> {
+/**
+ * Generates TypeScript types for all Astro modules. This sets up a `src/env.d.ts` file for type inferencing,
+ * and defines the `astro:content` module for the Content Collections API.
+ */
+export default async function sync(inlineConfig: AstroInlineConfig): Promise<ProcessExit> {
 	const logging = createNodeLogging(inlineConfig);
 	const { userConfig, astroConfig } = await resolveConfig(inlineConfig ?? {}, 'sync');
 	telemetry.record(eventCliSession('sync', userConfig));
@@ -48,7 +48,11 @@ export async function sync(
 		command: 'build',
 	});
 
-	return await syncInternal(settings, { logging, fs: options?.fs });
+	// Use this hack to prevent the sync options being public API
+	// eslint-disable-next-line prefer-rest-params
+	const internalOptions = arguments[1] as SyncOptions | undefined;
+
+	return await syncInternal(settings, { ...internalOptions, logging });
 }
 
 /**
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -12,7 +12,7 @@
   },
   "bugs": "https://github.com/withastro/astro/issues",
   "homepage": "https://astro.build",
-  "types": "./dist/@types/astro.d.ts",
+  "types": "./index.d.ts",
   "typesVersions": {
     "*": {
       "app": [
@@ -28,8 +28,8 @@
   },
   "exports": {
     ".": {
-      "types": "./dist/@types/astro.d.ts",
-      "default": "./astro.js"
+      "types": "./index.d.ts",
+      "default": "./dist/core/index.js"
     },
     "./env": "./env.d.ts",
     "./types": "./types.d.ts",
@@ -86,6 +86,7 @@
     "tsconfigs",
     "dist",
     "astro.js",
+    "index.d.ts",
     "config.d.ts",
     "config.mjs",
     "zod.d.ts",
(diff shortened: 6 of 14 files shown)


packages/astro/src/core/build/index.ts
```diff
@@ -37,24 +37,31 @@ export interface BuildOptions {
...
 	applyPolyfill();
 	const logging = createNodeLogging(inlineConfig);
 	const { userConfig, astroConfig } = await resolveConfig(inlineConfig, 'build');
 	telemetry.record(eventCliSession('build', userConfig));
 
 	const settings = createSettings(astroConfig, fileURLToPath(astroConfig.root));
 
+	// Use this hack to prevent the build options being public API
+	// eslint-disable-next-line prefer-rest-params
+	const internalOptions = arguments[1] as BuildOptions | undefined;
```
bluwy, line 61: Note: I chose to do this trick since I didn't want a wrapper `build` function that _hides_ the option. Otherwise there's a risk the implementation diverges and the jsdoc would be in an odd place.

If anyone feels it's better to wrap it instead, I'm also open to do so. I don't have a hard preference.

natemoo-re: Personally I'd prefer that we wrap it. I understand the concern about implementation divergence, but we can keep both functions in this file and make sure there are comments explaining what needs to happen when updating.

packages/astro/src/core/sync/index.ts
```diff
@@ -48,7 +48,11 @@ export async function sync(
 		command: 'build',
 	});
 
-	return await syncInternal(settings, { logging, fs: options?.fs });
+	// Use this hack to prevent the sync options being public API
+	// eslint-disable-next-line prefer-rest-params
+	const internalOptions = arguments[1] as SyncOptions | undefined;
+
+	return await syncInternal(settings, { ...internalOptions, logging });
```
natemoo-re, line 55: Same comment here.

natemoo-re (commented):
This looks great! Would definitely prefer having a public wrapper for `build` and `sync` though.

packages/astro/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from './dist/@types/astro.js'
```
Princesseuh, line 1: Should this be an `export type`?

ematipico (changes requested):
In the `AstroInlineConfig`, there's a `configFile` type. We should document it and explain how it behaves. I check its type and I don't remember why there's a `string | false`.

bluwy:
Forgot to document those, just pushed a commit adding some jsdocs.

ematipico (approved):
I still think we need to write some examples and tests for these APIs, to show how they work and **if they work** as intended. 

Maybe we can do a follow-up PR, but I think they are really important

bluwy:
I think we're already using them in the tests? I updated `test-utils.js` to use those. We could use more examples in the docs, but awaiting if the docs team have ideas on how to document those.

packages/astro/src/core/build/index.ts
```diff
@@ -37,14 +37,20 @@ export interface BuildOptions {
...
 	 */
 	teardownCompiler?: boolean;
 }
 
-/** `astro build` */
+/**
+ * Builds your site for deployment. By default, this will generate static files and place them in a dist/ directory.
+ * If SSR is enabled, this will generate the necessary server files to serve your site.
+ */
 export default async function build(
```
ematipico, line 51: Here's a thought that I am concerned about. 

It makes sense that `dev` and `preview` don't return anything, but I think `build` at least should return an array or a map of the assets that have been generated during the build, otherwise it wouldn't add enough value VS the CLI.

ematipico: > Maybe it would make sense to put these behind an experimental flag so we can change the returned signatures freely in the future.

Clever! BTW feel free to merge it; we can modify the return types, etc. in future PRs 😄

delucis (commented):
Hi @bluwy! Sorry it took me so long to get to this one.

The individual JSDoc comments here look OK in general but I’m missing example of how this would be used and if this is intended to be user-facing or not. Currently the changelog make it sound like an internal refactor.

While we think about where best to document this on the docs site itself, the changelog could be a good opportunity to sketch out some examples of running the commands and using the results?

One other comment from reading through this is that the config behaviour is not entirely clear to me. It seemed from the types that I can pass configuration directly inline, but then there’s also the comment about automatically looking for a config file or passing the path. Not clear to me if these are both the case, and if they are what happens if you have a config file _and_ inline config. Probably something else that examples would clarify.

bluwy:
No worries! I think we want this to be user-facing eventually, but it needs more polish along the way. Right now the return results of the functions are a bit lacking and will likely change, so leaning on not documenting those yet. BUT I think the parameters are more stable and are less likely to change.

I should also note in the changelog that they are experimental, and I can add more examples there.

> One other comment from reading through this is that the config behaviour is not entirely clear to me.

I'll also clarify this in the changeset! Inline configs should always take the highest priority, and you can indeed specify a custom config path, not load any config, or let Astro searches by default.

packages/astro/src/@types/astro.ts
```diff
@@ -1357,13 +1357,39 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
...
 export interface AstroInlineConfig extends AstroUserConfig, AstroInlineOnlyConfig {}
 export interface AstroInlineOnlyConfig {
+	/**
+	 * A custom path to the Astro config file. If relative, it'll resolve based on the current working directory.
+	 * Set to false to disable loading any config files.
+	 *
+	 * If this value is undefined or unset, Astro will search for an `astro.config.(js,mjs,ts)` file relative to
+	 * the `root` and load the config file if found.
+	 * 
+	 * The inline config passed in this object will take highest priority when merging with the loaded user config.
```
delucis, line 1373: This is clear — thank you! 🙌

delucis (commented):
Thanks Bjorn — this is great! Left a few comments.

One bit of API feedback to think about:

**Is it necessary to have the complexity of inline config _plus_ file loading _plus_ merging?**

I’m trying to think of the scenarios and it seems you’d _either_ use an on-disk config OR provide inline config, but not both. (In cases where you want some shared config which is overridden for a few values, I’d assume you can just import the shared config and spread it in?)

Almost like the usage can be:
```js
// Looks up config on disk
dev();

// Looks up config from custom location on disk
dev({ configFile: './custom-config.js' });

// Uses config inline
dev({ config: { site: 'https://example.com' } });
```

The high flexibility in the current design may seem user-friendly — they can do whatever they want! — but sometimes that leads to unintuitive corners, like:
- I, a user, set some inline config and run `build(myConfig)`.
- Hmm… build output looks like it used additional config I didn’t set 🤔 
- Why? Looks like it applied this on-disk config file 🤔 
- Why? I need to explicitly set `configFile: false` to disable that

I may be missing something though.

delucis:
Cool — OK, I’m happy you know what you’re doing. That all makes sense. (Although, for a tool built on top of Astro I might want user config to _override_ some of my defaults, so that might still end up needing to eject out of our config loading potentially? Because currently the tool’s inline config would always take precedence over the user’s config file.)

Making the auto-loading opt-in with `configFile: true` is a possibility! I’ll let you decide if you think that’s best or not. It sounds it from my example above, but I could also see the argument that `dev()` should “just work” and behave the same way a standard `astro dev` run does.

.changeset/many-pears-explode.md
```diff
@@ -0,0 +1,35 @@
+---
+'astro': major
+---
+
+Export experimental `dev`, `build`, `preview`, and `sync` APIs from `astro`. These APIs allow you to run Astro's commands programmatically, and replaces the previous entrypoint that runs the Astro CLI.
```
sarah11918, line 5: Just pointing out that I see the word "experimental" here and that typically means "behind an experimental flag" for us. If that's the case, then we should show the flag to enable and should include them in config-reference as an experimental entry.

If it's closer to something like "new" or "use at your own risk" or "good luck!", we might want to consider a different word or phrase that is closer to that meaning.

When it comes time for docs, we do have a CLI page in reference, and we could consider including this on that page. The page starts out pretty user-friendly, but then could progress into something like this.

bluwy: The APIs are directly imported so I don't think we can have a flag for it. It's closer to "use at your own risk" though, and I'm not exactly sure what's the alternative to "experimental", maybe "work-in-progress" or "unstable"? 🤔 

> When it comes time for docs, we do have a CLI page in reference, and we could consider including this on that page. The page starts out pretty user-friendly, but then could progress into something like this.

That sounds great. Since these APIs are _experimental_, and it's signature will likely change, do you think it's worth introducing a full section for these APIs as a start? Or I was thinking we can lightly touch on the JS API names for each CLI commands:

> `astro dev`
> ...
> You can also run the command programatically with the experimental `dev()` API exported from the `"astro"` package.

bluwy:
> Although, for a tool built on top of Astro I might want user config to _override_ some of my defaults, so that might still end up needing to eject out of our config loading potentially? Because currently the tool’s inline config would always take precedence over the user’s config file.

@delucis I think that could be better solved indeed if/once we export a `loadConfig` or `resolveConfig` API, so they can load it themselves, and another `mergeConfig` API to merge their defaults. I'm leaning towards exposing those in later PRs though. Their implementations aren't quite public-ready yet.

For now they could also add an integration in the inline config to manually set fallback values if it's unset.

.changeset/many-pears-explode.md
delucis: Our usual approach with experimental things is to give them an isolated page or section — makes it easier to avoid lots of repetition saying each time something is mentioned is experimental, makes it easier to update the docs as experimental features change, plus avoids drawing too much attention to a feature we’re still stabilising.

packages/astro/src/@types/astro.ts
```diff
@@ -1357,13 +1357,39 @@ export interface AstroConfig extends z.output<typeof AstroConfigSchema> {
...
+	 */
 	configFile?: string | false;
+	/**
+	 * The current mode used when building your site to generate either "development" or "production" code.
+	 */
 	mode?: RuntimeMode;
+	/**
+	 * The logging level to filter messages logged by Astro.
+	 * - "debug": Log all messages and debug information.
+	 * - "info": Log all messages.
```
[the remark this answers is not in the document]
bluwy: I think that works for me 👍

.changeset/many-pears-explode.md
bluwy: I've made a first draft of the API under the CLI commands page as a section at https://github.com/withastro/docs/pull/4234

sarah11918: I think that's a good place to start this, and keep it self-contained on that page as an Advanced section! Let's run with that idea for now.

sarah11918, line 5:
```suggestion

Export experimental `dev`, `build`, `preview`, and `sync` APIs from `astro`. These APIs allow you to run Astro's commands programmatically, and replaces the previous entry point that runs the Astro CLI.
```
From looking it up, I *think* entry point is two words. Unless you mean a very specific term/property, I'd choose two words here.

packages/astro/src/core/README.md
```diff
@@ -1,6 +1,8 @@
 # `core/`
 
-Code that executes within the top-level Node context. Contains the main Astro logic for the `build` and `dev` commands, and also manages the Vite server and SSR.
+Code that executes within the top-level Node context. Contains the main Astro logic for the `build`, `dev`, `preview`, and `sync` commands, and also manages the Vite server and SSR.
+
+The `core/index.ts` file is the main entrypoint for the `astro` package.
```
sarah11918, line 5:
```suggestion
The `core/index.ts` file is the main entry point for the `astro` package.
```

sarah11918 (approved):
I think this is fine, @bluwy , and I'll spend more time on the longer section that's going to be in docs!

Yan-Thomas (approved):
Went through the READMEs, changesets and comments, everything docs-wide LGTM!

PR #7980, opened 2023-08-07
Remove undici dep

bluwy:
## Changes

Remove `undici` dependency as Node.js 18 now includes it ootb.

## Testing

Existing tests should pass. I also ran the astro, node integration, and telemetry tests locally.

## Docs

n/a. internal change.

### DIFF
--- a/packages/integrations/node/src/response-iterator.ts
+++ b/packages/integrations/node/src/response-iterator.ts
@@ -6,7 +6,6 @@
 
 import type { ReadableStreamDefaultReadResult } from 'node:stream/web';
 import { Readable as NodeReadableStream } from 'stream';
-import type { Response as NodeResponse } from 'undici';
 
 interface NodeStreamIterator<T> {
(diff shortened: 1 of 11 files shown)


packages/integrations/node/src/response-iterator.ts
```diff
@@ -35,8 +34,8 @@ function isBuffer(value: any): value is Buffer {
 	);
 }
 
-function isNodeResponse(value: any): value is NodeResponse {
-	return !!(value as NodeResponse).body;
+function isNodeResponse(value: any): value is Response {
+	return !!(value as Response).body;
 }
```
natemoo-re, line 39: Do we even need this util anymore?

natemoo-re (approved):

packages/integrations/node/src/response-iterator.ts
bluwy: I think it's used to check the types below, and could probably be `instanceof Response` too. But I think I'll leave it for now to simplify the change.

PR #7981, opened 2023-08-07
Remove StreamingCompatibleResponse polyfill

bluwy:
## Changes

Node 18 should support streaming (ReadableStream) through undici now, so we don't have to polyfill it.

## Testing

Existing tests should pass.

## Docs

n/a. internal change.


natemoo-re (approved):
Hooray! Love that we're getting rid of this.

natemoo-re:
If we absolutely need to support native Node streams, we now have the option of using the native [`toWeb()`](https://nodejs.org/api/stream.html#streamreadabletowebstreamreadable-options) method as well.

PR #7943, opened 2023-08-03
Fix `injectRoute`

natemoo-re:
## Changes

- Adds an additional `resolvedI jectedRoutes` array to the internal `Settings` instance. 
- Updates the Vite plugin that manages Astro Integrations to ensure `resolvedInjectedRoutes` is populated with a fully resolved file URL for each entry point.
- Fixes #7563 
- Attempted to address #7561 but that’s going back to the drawing board.

## Testing

Manually confirmed, hard to test because it’s a workspace/dependency thing. 
 
## Docs

Bug fix only

PR #7986, opened 2023-08-07
Ensure `injectRoute` works during build

natemoo-re:
## Changes

- In https://github.com/withastro/astro/pull/7943, we missed the fact that the integrations container plugin was only applied during `dev`
- This updates the logic to ensure the integrations container plugin can always run
- Adds guard to only run `server:setup` hooks during `dev`
- Adds guard to only run `resolveInjectedRoutes` during server build (otherwise, `buildStart` runs twice)

## Testing

Tested manually because monorepo

## Docs

N/A

### DIFF
--- a/.changeset/red-bikes-return.md
+++ b/.changeset/red-bikes-return.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


delucis (approved):
May this PR be blessed with all good things as it sets sail under the starlight ✨

PR #7993, opened 2023-08-08
Add shorthand for Starlight templates to `create astro`

delucis:
## Changes

This adds some more special case handling for Starlight starter templates

Currently we hardcode `--template starlight` to `withastro/starlight/examples/basics`.

But https://github.com/withastro/starlight/pull/337 adds a new starter, and it would be nice to shorten the command needed to use it.

```sh
# current syntax
npm create astro@latest -- --template withastro/starlight/examples/tailwind

# with the changes in this PR
npm create astro@latest -- --template starlight/tailwind
```

If we think this is too much special-casing, we can close this.

## Testing

Tested the logic works manually, but not sure if it’s possible to run the local `create-astro`?

## Docs

Will only impact Starlight’s docs, which I’ll update if we merge this.

### DIFF
--- a/packages/create-astro/src/actions/template.ts
+++ b/packages/create-astro/src/actions/template.ts
@@ -67,9 +67,14 @@ const FILES_TO_UPDATE = {
 };
 
 function getTemplateTarget(tmpl: string, ref = 'latest') {
+	if (tmpl === 'starlight') tmpl = 'starlight/basics';
+	const starlightMatches = tmpl.match(/^starlight\/(.+)$/);
+	if (starlightMatches) {
+		const [, starter] = starlightMatches;
+		return `withastro/starlight/examples/${starter}`;
+	}
 	const isThirdParty = tmpl.includes('/');
 	if (isThirdParty) return tmpl;
-	if (tmpl === 'starlight') return `withastro/starlight/examples/basics`;
 	return `github:withastro/astro/examples/${tmpl}#${ref}`;
 }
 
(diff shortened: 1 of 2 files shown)


packages/create-astro/src/actions/template.ts
```diff
@@ -67,9 +67,14 @@ const FILES_TO_UPDATE = {
 };
 
 function getTemplateTarget(tmpl: string, ref = 'latest') {
+	if (tmpl === 'starlight') tmpl = 'starlight/basics';
+	const starlightMatches = tmpl.match(/^starlight\/(.+)$/);
+	if (starlightMatches) {
+		const [, starter] = starlightMatches;
+		return `withastro/starlight/examples/${starter}`;
+	}
```
natemoo-re, line 75: Just a suggestion for code clarity!
```suggestion
	if (tmpl.startsWith('starlight')) {
		const [_, starter = 'basics'] = tmpl.split('/');
		return `withastro/starlight/examples/${starter}`;
	}
```

natemoo-re (approved):

packages/create-astro/src/actions/template.ts
```diff
@@ -67,9 +67,12 @@ const FILES_TO_UPDATE = {
 };
 
 function getTemplateTarget(tmpl: string, ref = 'latest') {
+	if (tmpl.startsWith('starlight')) {
+		const [_, starter = 'basics'] = tmpl.split('/');
```
natemoo-re, line 71:
```suggestion
		const [, starter = 'basics'] = tmpl.split('/');
```

delucis:
Ah yeah, I did consider this but this version does potentially discard stuff if there are two slashes. Don't see when we'd need that though to be fair — thanks for merging!

Issue #7978, opened 2023-08-07
Exponential bloat in prop serialization


belluzj:
### What version of `astro` are you using?

2.10.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

When passing deeply nested JSON data to a client-side component (e.g. Vue) using client:load, the serialization of the props creates a quadratic amount of quote escaping with each level of nesting. 

See this repository reproducing the issue: https://github.com/daltonmaag/astro-prop-bloat-report/blob/main/src/pages/index.astro

Screenshot: 
![image](https://github.com/withastro/astro/assets/3616772/7bdc3fc3-d602-423d-91f0-da1ff6e3be43)

One can see on the screenshot that as the length of the JSON data in bytes increases linearly (second column) the length of the contents of the props="..." HTML attribute from Astro doubles from one line to the next.

The open devtools below shows that the quote escaping is the culprit.

### What's the expected result?

The length in bytes of the Astro props should be linearly proportional to the size of the input data.

### Link to Minimal Reproducible Example

https://github.com/daltonmaag/astro-prop-bloat-report

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Ah this is a great argument for moving these props out of the quoted `props` attribute and into a [`<script>`](https://github.com/withastro/roadmap/discussions/635) element, which is a proposal that I see you've already commented on!

Given this bloat problem, I would definitely consider this an important design change rather than a brand new feature. Since we're working on Astro 3.0 at the moment, that seems like a good time to make this change. Would you want to submit a PR against the in-progress [`next`](https://github.com/withastro/astro/tree/next) branch?

Here are the areas I imagine we need to update:
- The [`serializeProps` function](https://github.com/withastro/astro/blob/main/packages/astro/src/runtime/server/serialize.ts#L112-L115) currently generates JSON data that suffers from this bloat problem. This could be updated to generate a string of JavaScript that is written to an inline `script` element. Note that the generated `script` element should remain an inline (classic) script, because hydration should occur during streaming rather than be delayed until `type=module` scripts are executed.
- The corresponding [`serialize.test.js`](https://github.com/withastro/astro/blob/main/packages/astro/test/serialize.test.js) should be updated to verify this new format.
[...]

PR #7995, opened 2023-08-08
Remove nested JSON.stringify from props serialization

belluzj:
## Changes

- Only call `JSON.stringify` once when serializing props for an island
- Match this change when deserializing props for an island
- This fixes a quadratic increase in quote escaping when using nested data, see #7978 

## Testing

I added a test that the kind of nested data from #7978 results in a normal-sized serialized text.

I didn't add more tests in this branch because I'm not comfortable changing the code organization in the project but I committed to another branch how I tested the implementation, basically by adding roundtrip tests to the file `serialize.test.js`. However I had to move the deserialization to the same file as the serialization, which I'm not sure fits the project structure. See https://github.com/daltonmaag/astro/commit/a513407c9c5efb30cd7be70892c8e2c7e677db56

## Docs

I believe this change should be transparent to users so I'm not sure it needs documenting?

### DIFF
--- a/.changeset/great-icons-turn.md
+++ b/.changeset/great-icons-turn.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix quadratic quote escaping in nested data in island props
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
This looks good to me as a hotfix for `astro@2.x`. We'll want to [revisit this approach for `astro@3.x`](https://github.com/withastro/astro/pull/8004) to avoid the `&quote;` encoding issue entirely.

PR #7999, opened 2023-08-08
Adding Replay to the playwright e2e test suite

jaril:
## Changes

This adds the [Replay Playwright integration](https://www.replay.io/), so that we generate recordings of the playwright test suite.

These recordings (for example, [nested-in-react.test.js](https://app.replay.io/recording/svelte-counter--fd920640-0cf2-4413-b2dc-a95c2776112a) makes it easier to debug test failures with time-travel debugging. [Here's a quick demo](https://www.loom.com/share/3afc406a5c0a4585956b483881efbd53) of what this debugging workflow looks like in practice.

Replays are automatically recorded while the test suite runs, and are uploaded with `replayio/action-upload`. The logs of that step will contain the links to the uploaded replays.

To access them more conveniently on a per-run basis, you can join the Plausible CI team in Replay using [this link](https://app.replay.io/team/invitation?code=9695c7a3-4a53-46aa-a26b-3d16cfafd446).

## Testing

There's nothing to test — we're just changing the way the tests are run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Install Replay Playwright
+        run: npx @replayio/playwright install        
+
       - name: Build Packages
         run: pnpm run build
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,13 +179,26 @@ jobs:
         run: pnpm install
 
       - name: Install Replay Playwright
+        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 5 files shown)


natemoo-re:
Oh wow, this is really exciting! Let me chat with the team about this. Could be super helpful when we hit issues.

natemoo-re:
Alright @jaril, thanks for your patience! 

I discussed this with the team, and while this is very cool and seems useful, we're not sure it justifies the CI overhead if the team isn't actively taking advantage of Replay.

If we run into flaky E2E issues in the future and/or adopt Replay into our day-to-day workflows, we might reconsider this.

PR #8002, opened 2023-08-08
chore: check if ssr works on netlify

martcl:
## Changes

Added tests to validate that we are serving fresh content when using SSR on Netlify

## Testing

This test runs every week in a github action


natemoo-re (approved):
Thanks for the PR! I like the test

PR #8004, opened 2023-08-08
Adjust `astro-island` serialization approach

### LINKED ISSUE #7978: Exponential bloat in prop serialization

natemoo-re:
## Changes

- Closes #7978
- Overhaul the way `astro-island` handles component props. Instead of serializing the props, inlining them into the HTML (with attribute escaping), then deserializing the props on the client with a custom pipeline, this PR introduces [`seroval`](https://github.com/lxsmnsyc/seroval) for battle-tested and efficient behavior.
  - Seroval is used by Solid
  - Supports circular references
- Internal implementation details are no longer injected as attributes to `astro-island`, they are "hidden" inside a `<script async>` tag. This has multiple benefits:
  - Cleaner markup
  - Reduces footprint of large props
  - Eliminates deserialization cost

## Testing

Existing tests should pass. Also tested manually to ensure behavior is the same.

## Docs

This is an implementation detail, but probably should be called out in the migration guide. No user action needed, though.


natemoo-re:
Tests need to be updated to match this new format

packages/astro/package.json
```diff
@@ -153,14 +153,15 @@
...
     "network-information-types": "^0.1.1",
     "ora": "^6.3.1",
-    "sharp": "^0.32.1",
     "p-limit": "^4.0.0",
     "path-to-regexp": "^6.2.1",
     "preferred-pm": "^3.0.3",
     "prompts": "^2.4.2",
     "rehype": "^12.0.1",
     "semver": "^7.5.3",
+    "seroval": "^0.9.0",
```
ematipico, line 162: I am uncomfortable using a library that has not reached `1.0` yet.  

In the description, you say it's "battle-tested", but it's only used by solid. Could you please address my concerns?

lilnasy:
Would this block introducing strict content security policies down the line?

packages/astro/package.json
lilnasy: I am concerned by seroval's [use of `eval()`](https://github.com/lxsmnsyc/seroval/blob/v0.9.0/packages/seroval/src/index.ts#L100). Was [devalue](https://github.com/rich-harris/devalue) considered? It seems to be the SvelteKit equivalent.

natemoo-re:
> Would this block introducing strict content security policies down the line?

@lilnasy I don't think so. Similar to the existing inline scripts that we already generate, we can create a `nonce` for these `script` elements and expose an allowlist for CSP usage (somehow). It's not fundamentally different that what we already do.

packages/astro/package.json
ematipico: Do you think we could ask them to release a `1.0` version? Or at least ask them what's their plan with versioning.

natemoo-re: Looks like there are maybe some super-powerful streaming features planned prior to v1.0, but it's otherwise stable. https://twitter.com/lxsmnsyc/status/1689294233272737793

lilnasy: Would that be non-breaking?

natemoo-re: Yes, I think there are separate APIs for any `async` or `streaming` operations

sarah11918:
> This is an implementation detail, but probably should be called out in the migration guide. No user action needed, though.

If it's possible that someone's project *may* exhibit different behaviour with this under-the-hood change (though, there's nothing they should do differently), then we can do a:

In 2.0, `astro-island` component props were... (action/handled in a way...)
In 3.0, Astro uses X to more efficiently...

What should I do?
No change is needed to your project code, but you may notice/if you notice...

If you want something like this in the migration guide, then you can add it to the pile in the [changes tracking docs issue](https://github.com/withastro/docs/issues/3756). If something like this doesn't make sense to include in the guide, then probably OK to not include in the guide itself, and leave it to a changelog?

packages/astro/src/runtime/server/astro-island.ts
```diff
@@ -43,7 +25,14 @@ declare const Astro: {
 			class extends HTMLElement {
 				public Component: any;
 				public hydrator: any;
-				static observedAttributes = ['props'];
+
+				public opts: Record<string, any> = {};
```
bluwy, line 29: Would be nice to put a comment above here that these properties are coming from the `<script>` child.

bluwy (approved):
Looks great! I like this serialization approach more

ematipico (approved):

packages/astro/test/test-utils.js
```diff
@@ -243,6 +243,20 @@ export async function loadFixture(inlineConfig) {
...
+ * @returns {{ componentUrl: string, componentExport: string, rendererUrl: string, beforeHydrationUrl?: string, opts: Record<string, unknown>, props: Record<string, unknown> }}
+ */
+export function getIslandDataFromScript(text) {
+	const Astro = { assign: (_, props) => props };
+	const document = {};
+	const args = { Astro, document }
+
+	// Yes, I know, `eval` is bad! But our data is embedded as JS on purpose.
+	return new Function(Object.keys(args), `return ${text}`)(...Object.values(args));
+}
```
natemoo-re, line 258: Calling this out! New test util that extracts data from the `astro-island > script` text. It's using `new Function` because the data is now embedded as JavaScript rather than JSON, so we need to `eval` it. 

However, we don't serialize functions so this content is all safe, and the function is scoped to receive the local `Astro` and `document` args defined here.

matthewp (approved):

natemoo-re:
Damn, still struggling with some E2E tests pretty consistently. This approach seems to be causing HMR and `client:only` issues across the board.

Think we're going to have to punt this from the 3.0 launch and ship this as an experimental feature flag in 3.1. That's probably a safer approach anyway!

Princesseuh:
Marked this as a draft since I believe we're still doing this, it's just not ready yet. Let me know if I should close @natemoo-re

natemoo-re:
Thanks @Princesseuh! Still planning on picking this up, but yeah draft is great for now.

matthewp:
Closing the PR, but we hope to pick it back up again!

spacedawwwg:
Had my fingers crossed for this one 🤞🏻

lxsmnsyc:
I hope this gets picked up in the future. Seroval now has a cross-reference serializer that allows reference deduping across serialized objects, and so I wanted to see it in action in Astro. I'm certain Astro would benefit on that mechanism.

natemoo-re:
For the record, I would still like to make this happen! Just wasn't able to prioritize it. Hoping to pick it back up eventually...

spacedawwwg:
😢

PR #8007, opened 2023-08-09
Support Bun by adjusting how `@babel/plugin-transform-react-jsx` is imported.

paperdave:
## Changes

This brings Astro a step closer to running on Bun by working around an existing workaround done when loading `@babel/plugin-transform-react-jsx`.

In Bun and some other tooling, `import`ing a CommonJS module that has `module.exports.__esModule` will tell the module loader/bundler to set `module.exports.default` as the ESM default export, where without this annotation the default export will be set to `module.exports`. Node does not do this transformation, which turns transpiled-to-cjs modules like `@babel/plugin-transform-react-jsx` into `{ default: { default: Function } }`, where in bun this loads as originally intended like `{ default: Function }`

This PR fixes the three places astro uses `.default.default` and ensures it handles the case where it may be just `.default`.

Our tracking issue for Astro as a whole is https://github.com/oven-sh/bun/issues/3746. This PR doesn't fully close that yet, as there are some remaining stability issues in Bun (random segfaults) when using larger projects.

## Testing

I have been manually testing this on the astro.build docs locally on macos with node.js v20.5.0 and a development build of bun.

## Docs

This is an internal change to the codebase and has no user-facing side effects. Maybe I should've added a comment for future code readers to understand why checking both `.default.default` and `.default` is necessary.

### DIFF
--- a/.changeset/quick-actors-sing.md
+++ b/.changeset/quick-actors-sing.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/preact': patch
+'astro': patch
+---
+
+Support Bun by adjusting how `@babel/plugin-transform-react-jsx` is imported.
--- a/packages/astro/src/jsx/renderer.ts
+++ b/packages/astro/src/jsx/renderer.ts
@@ -3,10 +3,9 @@ const renderer = {
 	serverEntrypoint: 'astro/jsx/server.js',
 	jsxImportSource: 'astro',
 	jsxTransformOptions: async () => {
-		const {
-			default: { default: jsx },
-			// @ts-expect-error
-		} = await import('@babel/plugin-transform-react-jsx');
+		// @ts-expect-error types not found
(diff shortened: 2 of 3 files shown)


bluwy (approved):

matthewp:
I'm not sure if we should add special-casing code for Bun. I could see someone looking at this code and thinking it's unnecessary and removing it.

natemoo-re:
> I'm not sure if we should add special-casing code for Bun. I could see someone looking at this code and thinking it's unnecessary and removing it.

I don't mind this change, personally! It seems trivial to support Bun's divergent behavior here. Since the ultimate goal for Bun is full Node compat, the rest of compatability issues will be their responsibility.

natemoo-re (approved):
LGTM. Worth noting that this code is changing in #7924 for 3.0, but this is worth merging regardless IMO!

PR #8010, opened 2023-08-09
CollectionEntry type improvement 

OliverSpeir:
Provides more useful information when hovering variables typed with CollectionEntry<"collectionName"> 

* Does not change the functionality of existing code

* Flatten<AnyEntryMap[C]> is the same as AnyEntryMap[C][keyof AnyEntryMap[C]]

* Both will produce a union of all note entry types within the "collectionName" key

## Changes

- What does this change?

The CollectionEntry type which is a part of astro:content module that is autogenerated in .astro/types.d.ts

## Testing

Made this change locally and tried it with VSCode having made sure the types were updated with both astro build and npm run dev. Tried it with multiple collections, of different shapes. 

## Docs

No, it will only improve IDE experience.

This nuance typescript experience isn't currently documented, more of something a user would stumble upon, but it is somewhat expected behavior to have a useful hover for a typed variable.

### DIFF
--- a/packages/astro/content-types.template.d.ts
+++ b/packages/astro/content-types.template.d.ts
@@ -10,7 +10,9 @@ declare module 'astro:content' {
 
 declare module 'astro:content' {
 	export { z } from 'astro/zod';
-	export type CollectionEntry<C extends keyof AnyEntryMap> = AnyEntryMap[C][keyof AnyEntryMap[C]];
+	
+	type Flatten<T> = T extends { [K: string]: infer U } ? U : never;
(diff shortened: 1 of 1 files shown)


ematipico:
Thank @OliverSpeir , could you provide us with some screenshots of **how** this PR improves DX?

OliverSpeir:
Before: 
before change


After:
after change

bholmesdev (approved):
Nice change @OliverSpeir! TIL about infer. Definitely prefer this 👏

PR #7997, opened 2023-08-08
Revert hoisted scripts analysis (#7707)

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7744

Reverts #7707. Reason at https://github.com/withastro/astro/issues/7857#issuecomment-1669783879

## Testing

Skipped the added test in #7707 for now so we can re-use that when we revisit this.

## Docs

We mention this in https://astro.build/blog/astro-290/#hoisted-script-optimization, but I don't think it make sense to remove it? Or may worth providing an update? I don't think this is an easy issue to fix completely in the short term.

/cc @withastro/maintainers-docs for feedback!

Issue #7744, opened 2023-07-21
The component js is lost after packaging.


J-env:
### What version of `astro` are you using?

2.9.0+

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm,pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

Since there is so much code in the project, it is difficult to find out what is causing it.
The following is the details page after construction as an example.

/dist/detail/v5.html:

<img width="1127" alt="image" src="https://github.com/withastro/astro/assets/25259920/7dd679b3-1807-44df-b16a-0a921558260c">

The <Modal /> component is used in the details page, but the <Modal /> component js is not imported in this js file.
There is no such problem in the development environment.
/dist/static-dev/hoisted.470aeb3d.js:

<img width="974" alt="image" src="https://github.com/withastro/astro/assets/25259920/9864f668-2607-4f93-8c94-ff8f093b38dc">

### What's the expected result?

It was good before version 2.9.0

### Link to Minimal Reproducible Example

none

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Hmm It might be caused by https://github.com/withastro/astro/pull/7707, which changes hoisted scripts detection across Astro files.

natemoo-re:
A minimal reproduction would be extremely helpful if you're able to provide one.

tillprochaska:
I’ve been seeing the same issue. We have multiple documentation sites built using Astro and share a few common components between them.

While I haven’t had the time yet to set up a minimal reproduction example, the site is open source. Maybe that’s already helpful.

* This is one of the Astro sites: https://github.com/alephdata/aleph/tree/develop/docs
* All pages include the `AlgoliaSearch` component (via a layout component) from [this package](https://github.com/alephdata/astro-theme-docs/blob/main/src/components/page/AlgoliaSearch.astro).
* When running `astro dev`, the JS for the `AlgoliaSearch` component is included on all pages as expected.
* When running `astro build`, the JS is *not* included.

I have temporarily locked the Astro version to 2.8.5 which *does not* have this problem.

sermagnus:
Same issue

ottomated:
The root issue here is #7857

PR #8011, opened 2023-08-09
Move hoisted script analysis optimization as experimental

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7744
Supersedes and closes https://github.com/withastro/astro/pull/7997

Move the feature in https://github.com/withastro/astro/pull/7707 behind an experimental flag. And added extra comments to make the analysis easier to follow.

## Testing

Updated hoisted script test to enable the experimental flags.

## Docs

Added the jsdocs. cc @withastro/maintainers-docs for feedback!

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1280,9 +1280,10 @@ export interface AstroUserConfig {
 		 * @default `false`
 		 * @version 2.10.4
 		 * @description
-		 * Enable hoisted script analysis optimization to prevent unused components' script from being included in a page unexpectedly.
+		 * Prevents unused components' scripts from being included in a page unexpectedly.
 		 * The optimization is best-effort and may inversely miss including the used scripts. Make sure to double-check your built pages
 		 * before publishing.
+		 * Enable hoisted script analysis optimization by adding the experimental flag:
 		 *
 		 * ```js
 		 * {
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1272,6 +1272,27 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		viewTransitions?: boolean;
+
+		/**
+		 * @docs
+		 * @name experimental.optimizeHoistedScript
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 2.10.4
+		 * @description
+		 * Enable hoisted script analysis optimization to prevent unused components' script from being included in a page unexpectedly.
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/build/plugins/plugin-analyzer.ts
```diff
@@ -30,29 +30,28 @@ async function doesParentImportChild(
...
-		},
-	});
-	// All of the aliases the current component is imported as
+	for (const node of (parentInfo.ast as any).body) {
+		if (node.type === 'ImportDeclaration') {
+			imports.push(node);
+		} else if (node.type === 'ExportDefaultDeclaration' || node.type === 'ExportNamedDeclaration') {
+			exports.push(node);
+		}
+	}
```
bluwy, line 46: Import and exports declarations are top level only, so use this optimization.

ematipico (approved):

matthewp (approved):

natemoo-re (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -1272,6 +1272,27 @@ export interface AstroUserConfig {
...
+		/**
+		 * @docs
+		 * @name experimental.optimizeHoistedScript
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 2.10.4
+		 * @description
+		 * Enable hoisted script analysis optimization to prevent unused components' script from being included in a page unexpectedly.
+		 * The optimization is best-effort and may inversely miss including the used scripts. Make sure to double-check your built pages
+		 * before publishing.
```
sarah11918, line 1285:
```suggestion
		 * Prevents unused components' scripts from being included in a page unexpectedly.
		 * The optimization is best-effort and may inversely miss including the used scripts. Make sure to double-check your built pages
		 * before publishing.
		 * Enable hoisted script analysis optimization by adding the experimental flag:
```
Nice, @bluwy!  For future proofing, since presumably at some point this will no longer be experimental, we could maybe just start with the description of what it does, and move the instruction to enable closer to the code block? That way, this wording will continue to work and can be reused if the config option description is needed in the future. :raised_hands: !

sarah11918 (approved):
One small suggestion to make this A++, @bluwy! (OK, to make my life easier in the future... but that still counts as A++ in my books!!)  :wink:

packages/astro/src/@types/astro.ts
bluwy: Sounds great to me. I was following the description pattern in the other experimental options for this, but either way works for me.

PR #8012, opened 2023-08-09
feat: expose user astro error

ematipico:
## Changes

Adds a new `astro/errors` module, where the user can import `AstroUserError` and use it as they see fit.

The class accepts a mandatory `message` and an optional `hint`.

## Testing

N/A

## Docs

N/A


packages/astro/src/core/errors/userError.ts
```diff
@@ -0,0 +1,20 @@
...
+	hint?: string;
+}
+
+export function createAstroError({ message, hint }: CreateAstroError): AstroError {
+	return new AstroError({
+		name: 'AstroError',
+		message,
+		hint,
+	});
+}
```
bluwy, line 20: Maybe we should export the `AstroError` class directly? It could be used to `extend` and checked for `instanceof` by users, and errors are usually created like `throw new AstroError()` in JS.

ematipico: If you check the linear issue related to this implementation, Erika suggested avoiding exposing `AstroError` because it contains a lot of internal things that might not be needed. 

Do you still think it's worth exposing the whole `AstroError`?

bluwy: Thanks for the insights. I think it makes sense to limit the API surface if we want to be cautious, but if we're already returning an instance of `AstroError` that already exposes a bit of API 🤔 

For example, this PR limits to `message` and `hint` only, but someone could also do:

```js
const error = createAstroError({ ... })
error.setLocation(...)
```

From Erika's suggestion in Linear, maybe we should extend from `Error` only instead and not use `AstroError`?

bluwy (approved):
I'm ok with this approach! But also would like to hear Erika's thoughts if `AstroUserError` is a suitable name for this.

Also wonder if it's worth aliasing `AstroUserError` to `AstroError` during entrypoint re-export to make the API cleaner, but I'm fine either way.

Princesseuh:
I think aliasing it to AstroError would be fine yeah

Issue #7849, opened 2023-07-28
ViewTransition: Self link triggers animation


florian-lefebvre:
### What version of `astro` are you using?

2.9.4

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Brave

### Describe the Bug

When clicking on a link that is the same to the current page, animation is triggered even though we stay on the page.

To view the bug, go on a blog post and click "Link to self"

### What's the expected result?

I see 2 possibilities:

1. Don't trigger animation
2. Trigger and scroll top

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-d5ubcr

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Is this for hash links or something less? Hash links were just fixed (yesterday I believe).

florian-lefebvre:
I need to try with the scroll thing to see of it feels okay. I'll keep you updated

florian-lefebvre:
It does not feel right imo, wdyt?

https://github.com/withastro/astro/assets/69633530/5f85ed66-2dac-452c-997b-7b5d629c3dcd

matthewp:
yeah probably not, have you tried with 2.10?

florian-lefebvre:
Tried with latest (2.10.1) and it now scrolls to top as well, but feels a bit weird. I think that, when on the same page, it should scroll up without triggering the animation

martrapp:
Currently, no animations are started for links with hashes.
As a workaround, you could replace `<a href="">Link to self</a>` with `<a href="#top">Link to self</a>`.
Perhaps this should be changed so that all links to the same page (including `href=""` & `href="#"`) should not trigger animations.

@matthewp what do you think?

matthewp:
Fixed by https://github.com/withastro/astro/pull/8013, thanks @martrapp !

Issue #7882, opened 2023-07-31
viewTransitions: clicking a link doesn't reset scrollY to 0


martrapp:
### What version of `astro` are you using?

2.9.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

For viewTransitions: 
- The current page is larger than the viewport and was scrolled to the bottom.
- Navigation to another page shows the middle/bottom part of that page, not the beginning.

### What's the expected result?

Regardless of the scroll position of the referring page, the new page is displayed from the beginning

Possible solution: 		
initialize scrollY in the click handler when navigating: navigate('forward', link.href, { scrollY: 0 });

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-fmqrqm?file=src%2Fpages%2Fother-page.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #8013, opened 2023-08-09
View Transitions: do not animate same page navigation, but animate hash links to different pages

martrapp:
## Changes

- Formerly, links with a hash mark did not start view transitions. Now they will, if they lead to a different page.
- Formerly, links to the same page (without hashes) did start view transitions. Now they don't.  

## Testing

Tested with the examples from #7849 and #7882

## Docs

No doc updates required.
Fixes unexpected behavior

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -224,12 +224,6 @@ const { fallback = 'animate' } = Astro.props as Props;
 		document.head.append(link);
 	}
 
-	function isSamePage(location, link) {
-		const linkURL = new URL(link.href, location.href);
-		linkURL.hash = location.hash;
-		return linkURL.href.replace(/#$/, '') === location.href.replace(/#$/, '');
-	}
-
 	if (supportsViewTransitions || getFallback() !== 'none') {
 		document.addEventListener('click', (ev) => {
 			let link = ev.target;
@@ -245,7 +239,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 				link.href &&
 				(!link.target || link.target === '_self') &&
 				link.origin === location.origin &&
-				!isSamePage(location, link) &&
+				location.pathname !== link.pathname &&
 				ev.button === 0 && // left clicks only
 				!ev.metaKey && // new tab (mac)
 				!ev.ctrlKey && // new tab (windows)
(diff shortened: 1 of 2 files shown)


martrapp:
There is still some minor issue:  `samePage("name", "name#")`  returns false, which makes links with `href="#"` to start a transition. (`href=""` and href=`#top` will not start transitions). Will add an additional fix for that to this PR.

martrapp:
done.

packages/astro/components/ViewTransitions.astro
```diff
@@ -220,6 +224,12 @@ const { fallback = 'animate' } = Astro.props as Props;
 		document.head.append(link);
 	}
 
+	function isSamePage(location, link) {
+		const linkURL = new URL(link.href, location.href);
+		linkURL.hash = location.hash;
+		return linkURL.href.replace(/#$/, '') === location.href.replace(/#$/, '');
```
matthewp, line 230: Can you explain what this is doing? It's not obvious to me. Can we not just check if `location.pathname === link.pathname`?

martrapp: Ha :smile: When I started, it seemed like the natural way to check if two URLs are "equal modulo the fragment ID". It got more complicated when I noticed a problem with hashes at the end.

But I probably mostly avoided worrying about other URL properties like searchParams.

I only have SSG experience with Astro.  
I suppose there are also SSR examples with fixed pathnames, where people expect different searchParams to show the transition to different pages?

Maybe we should aim for a simple, generic solution and give users an additional link property for manual control.

pathname === pathname is fine with me. I will fix it that way.

Maybe you already have ideas on how to override link behavior to force transitions for same pathnames. I'll leave that to a later change :-)

matthewp:
@martrapp can you mark this as ready for review?

Also, does this close https://github.com/withastro/astro/issues/7849?

martrapp:
> Also, does this close #7849?

yes, the example included in that issue now works as expected: `href=""` jumps to top of the page without triggering a view transition.

ryansolid:
I think there are cases where you want to navigate to the same page and have it preserve.

1. Pagination using search parameters. This is where I noticed this fix messed up my Hackernews example. Things like `https://hackernews.com/top?page=2`. This also applies to search pages which will often refresh with different search parameters/filters.

2. Mutations. This one might be a different case but say posting a form to add a comment to a page while watching a video and not having the video unload. In this case there is no url difference but you just want to force it.

matthewp:
Thanks @ryansolid, fixed in `astro@2.10.7`

PR #8015, opened 2023-08-09
Remove the Vercel Edge adapter

matthewp:
## Changes

- Removes the `@astrojs/vercel/edge adapter, advices people to use Serverless + Edge Middleware instead.

## Testing

- Tests updated

## Docs

- Readme docs updated
- Changeset used as breaking change notification.


.changeset/plenty-keys-add.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/vercel': major
+---
+
+Remove the Vercel Edge adapter
+
+With the Vercel adapter now supporting Edge middleware, there's no longer a need for an adapter for Edge itself (deploying your entire app to the edge). This adapter had several known limitations and compatibility issues that prevented very many people from using it in production. To reduce maintenance costs and because we have a better story with Serveless + Edge Middleware, we are removing the Edge adapter.
```
sarah11918, line 7:
``````suggestion
 `@astrojs/vercel/serverless` now supports Edge middleware, so a separate adapter for Edge itself (deploying your entire app to the edge) is no longer necessary. Please update your Astro config to reflect this change:
 
 ```diff
 // astro.config.mjs
import { defineConfig } from 'astro/config';
- import vercel from '@astrojs/vercel/edge';
+ import vercel from '@astrojs/vercel/serverless';

export default defineConfig({
  output: 'server',
  adapter: vercel(),
});
```
 
This adapter had several known limitations and compatibility issues that prevented many people from using it in production. To reduce maintenance costs and because we have a better story with Serveless + Edge Middleware, we are removing the Edge adapter.
``````
What do you think about something like this? (I'm assuming the correct change to make is to now use the `/serverless` instead?)

sarah11918 (commented):
I think showing a recommendation re: what to do now that this is gone is helpful, too! Something in the style of below. (Update it as required for technical accuracy).

ematipico (approved):

natemoo-re (approved):
👍🏻 to Sarah's comment! Otherwise LGTM.

.changeset/plenty-keys-add.md
matthewp: Yeah, should also show using Edge Middleware since that is what we are recommending. I'll accept and then make that change.

PR #8019, opened 2023-08-10
Clear todos and remove css kebab handling

bluwy:
## Changes

- Clear some actionable TODO comments
- Remove kebab-case transform for camelCase CSS variables (#6264 for more context)
- Remove `astro add` error when match `package.json` files. Our config search code now only matches `astro.config.*`, so it'll never hit that condition.

## Testing

Removed some test skips marked todo. See if they pass now.

### DIFF
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -29,10 +29,6 @@ const toStyleString = (obj: Record<string, any>) =>
 	Object.entries(obj)
 		.map(([k, v]) => {
 			if (k[0] !== '-' && k[1] !== '-') return `${kebab(k)}:${v}`;
-			// TODO: Remove in v3! See #6264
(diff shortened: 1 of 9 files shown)


packages/astro/src/runtime/server/render/util.ts
```diff
@@ -29,10 +29,6 @@ const toStyleString = (obj: Record<string, any>) =>
 	Object.entries(obj)
 		.map(([k, v]) => {
 			if (k[0] !== '-' && k[1] !== '-') return `${kebab(k)}:${v}`;
-			// TODO: Remove in v3! See #6264
-			// We need to emit --kebab-case AND --camelCase for backwards-compat in v2,
-			// but we should be able to remove this workaround in v3.
-			if (kebab(k) !== k) return `${kebab(k)}:var(${k});${k}:${v}`;
```
natemoo-re, line 35: ♥️

natemoo-re (approved):
Great cleanup, thank you!

PR #8020, opened 2023-08-10
refactor: add pipeline concept

ematipico:
## Changes

This PR starts the refactors around how routes are rendered. 

The idea is the following: creating a "core" that does one single thing, render a route by returning a `Response`, and remove from the equation all the information that aren't part of a page. For example, the `Environment` is not information that belongs exclusively to a page. 

Then, in the future refactos, we are going to create **specialized** pipelines - or consumers - that compute the basic information in a different way, and then **handle** the finished `Response` in a different way. These specialized pipelines will be: the dev server, the build, SSR. In the future we will have containers and testing.

We are going to have a `Pipeline` base class, which exposes a method called `renderRoute`. It accepts the `RenderContext` and the `ComponentInstance` and returns a `Response`. Very simple. 

In this PR I created the first consumer, AKA `SSRRoutePipeline`. As you might have noticed, some logic was moved inside this new class. 
 

In order to make the internal developments better, I marked `route` inside `RenderContext` mandatory. The reason why I did so is because `RouteData` is supposed to contain the information of the **type** of the route we're about to render, but since could be `undefined`, we ended up having an API like this:

```js
tryRenderRoute('page', renderContext)
```

Which doesn't make much sense, because we're supposed to have the `route` already inside `renderContext`. I marked it mandatory, and the majority of the tests still pass, which means that the change in logic is correct. 

## Testing

The existing tests should pass. I updated the tests where we create the `RenderContext` manually, because now `route` is mandatory. The shape of `route` used in tests won't trigger the function calls `callGetStaticPaths`, which isn't needed in those tests.

## Docs

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -1,13 +1,13 @@
-import mime from 'mime';
 import type {
 	EndpointHandler,
 	ManifestData,
+	MiddlewareEndpointHandler,
 	RouteData,
 	SSRElement,
 	SSRManifest,
 } from '../../@types/astro';
 import type { SinglePageBuiltModule } from '../build/types';
-import { attachToResponse, getSetCookiesFromResponse } from '../cookies/index.js';
+import { getSetCookiesFromResponse } from '../cookies/index.js';
 import { consoleLogDestination } from '../logger/console.js';
 import { error, type LogOptions } from '../logger/core.js';
 import {
@@ -16,12 +16,10 @@ import {
 	removeTrailingForwardSlash,
 } from '../path.js';
 import { RedirectSinglePageBuiltModule } from '../redirects/index.js';
-import { isResponse } from '../render/core.js';
 import {
 	createEnvironment,
 	createRenderContext,
 	tryRenderRoute,
-	type Environment,
 	type RenderContext,
 } from '../render/index.js';
 import { RouteCache } from '../render/route-cache.js';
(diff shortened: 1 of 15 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -184,35 +184,8 @@ export class App {
 			}
 			Reflect.set(response, responseSentSymbol, true);
 			return response;
-		} else {
```
matthewp, line 187: wow, love all of this deletion

matthewp:
Looks great! Is the plan to later add a build pipeline and a dev pipeline?

matthewp (approved):

packages/astro/src/core/app/index.ts
natemoo-re: Agreed, all this cleanup in the `App` is great!

natemoo-re (approved):
Looks great! Excited to see how this centralizes our logic across different targets!

packages/astro/src/core/app/ssrPipeline.ts
```diff
@@ -0,0 +1,54 @@
...
+export class SSRRoutePipeline extends Pipeline {
+	encoder = new TextEncoder();
+
+	constructor(env: Environment) {
+		super(env);
+		this.setEndpointHandler(this.#ssrEndpointHandler);
+	}
+
+	// This function is responsible for handling the result coming from an endpoint.
+	async #ssrEndpointHandler(request: Request, response: EndpointCallResult): Promise<Response> {
```
bluwy, line 27: Small nit: Is `this.setEndpointHandler` and `this.#ssrEndpointHandler` something that every extended class of `Pipeline` needs to implement? I was thinking we can create an abstract method like `abstract endpointHandler()` in `Pipeline` instead, but I'm also fine with this pattern if you're still shaping up the final pattern.

ematipico: Technically, no.

Not all the future pipelines will have to handle endpoints. For example: a testing pipeline will need to be used only when needed, an Astro container won't need it (I presume)

PR #8021, opened 2023-08-10
[@astrojs/vercel] Individually enable Speed Insights and Web Analytics

chriswdmr:
## Changes

Major plugin changes

**Added:**
- `webAnalytics` property
- `speedInsights` property

Introduces dedicated properties on the plugin to enable Vercel Speed Insights and Vercel Web Analytics individually.

**Removed:**
- `analytics` property
It got removed because it toggled both individual features at once which was not ideal since not every user has both features enabled. 
It also threw an error when only Web Analytics was enabled since the environment variable for Speed Insights `VERCEL_ANALYTICS_ID` was not available.

## Testing
I've added basic tests to verify the features are loaded correctly.

## Docs
I've also updated the docs accordingly, let me know if I can make them better


natemoo-re:
Thanks @chriswdmr! Conceptually, this change LGTM.

To bikeshed on the exact API a bit, I wonder if we want to make this backwards compatible and allow `analytics` to accept a boolean or an object that configures these two separately? Seems like we could do that on a `minor` without forcing everyone to change their code.

I'm also a bit worried about exposing all the underlying options from `@vercel/analytics` to the `webAnalytics.config` property. IMO if users want to deeply customize the behavior of their analytics, they can always use `@vercel/analytics` directly. Also passing functions from the `astro.config.mjs` file to `injectRoute` probably doesn't work as expected since functions aren't serializable?

chriswdmr:
>To bikeshed on the exact API a bit, I wonder if we want to make this backwards compatible and allow analytics to accept a boolean or an object that configures these two separately? Seems like we could do that on a minor without forcing everyone to change their code.

Good point! I think it's super tricky since we can't know which features are available and enabled on the user side. Hence why I dedicated to not leave the `analytics` property there since it's just confusing with the updated naming of both Vercel features.

>I'm also a bit worried about exposing all the underlying options from @vercel/analytics to the webAnalytics.config property. IMO if users want to deeply customize the behavior of their analytics, they can always use @vercel/analytics directly. Also passing functions from the astro.config.mjs file to injectRoute probably doesn't work as expected since functions aren't serializable?

Oh, I didn't know that, thanks for the callout. I'm not sure how other adapters are handling that, do you have some pointers? Or is the general ideal to have some "simple" support but if you want to do more things you opt out of the adapter?

matthewp:
@chriswdmr The reason why this won't work is that we don't build the config file as part of the production bundle (as it often contains a lot of dev dependencies).

The pattern we often use instead is to allow the user to have a module that they specify some exports from. I do this in one of my own adapters, see this: https://github.com/matthewp/astro-fastify#entry

So in your example you might have them define a module like `vercel-analytics.{js,ts}` with an interface like:

```js
export const beforeSend = (event) => {
  // Ignore all events that have a `/private` inside the URL
  if (event.url.includes('/private')) {
    return null;
  }
  return event;
}
```

And then have them pass that module as a config option. Then in your injected script import it like so:

```js
import { inject } from '@vercel/analytics';
import { beforeSend } from '${userDefinedModule}';
inject({
  mode: ${config?.mode},
  beforeSend: beforeSend,
  debug: ${config?.debug}
});`;
```

This is pseudo-code of course, also want to account from them not defining such module.

packages/integrations/vercel/README.md
```diff
@@ -101,11 +102,56 @@ import vercel from '@astrojs/vercel/serverless';
...
+    webAnalytics: {
+      enabled: true,
+    },
   }),
 });
 ```
 
+#### `beforeSend`
+
+Since functions can't be passed down via `astro.config.mjs`, you need to export the `beforeSend` function in a separate file inside your root called `vercel-web-analytics.ts`.
```
ematipico, line 114: I don't think we should tell users about our design decisions. We should just tell them "that's how you use this hook/function".

packages/integrations/vercel/README.md
```diff
@@ -101,11 +102,56 @@ import vercel from '@astrojs/vercel/serverless';
...
+If you're not using TypeScript, you can define the function inside `vercel-web-analytics.js`.
+
+```js
```
ematipico, line 117:
```suggestion
```ts
```

packages/integrations/vercel/README.md
```diff
@@ -101,11 +102,56 @@ import vercel from '@astrojs/vercel/serverless';
...
+    // Ignore all events that have a `/private` inside the URL
+    if (event.url.includes('/private')) {
+      return null;
+    }
+    return event;
+  }
+}
+```
+
+````js
```
ematipico, line 131:
```suggestion

```

packages/integrations/vercel/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@astrojs/internal-helpers": "workspace:*",
-    "@vercel/analytics": "^1.0.2",
+    "@vercel/analytics": "~1.0.0",
```
ematipico, line 55: I don't think this change is needed.

packages/integrations/vercel/src/edge/adapter.ts
ematipico, line 1: We deprecated the edge adapter of vercel, so this file shouldn't be here,

packages/integrations/vercel/src/lib/web-analytics.ts
```diff
@@ -0,0 +1,56 @@
...
+			return {
+				beforeSend: functions.default.beforeSend,
+			};
+		}
+
+		return {
+			beforeSend: undefined,
+		};
+	} catch (e) {
+		return {
```
ematipico, line 38: I think we should still log `e`, so at least we know what went wrong during debugging. 

You can obtain the astro logger from the Astro hooks, it's called `logger`, and its type is `AstroIntegrationLogger`. 

You can pass it to `getInjectableWebAnalyticsContent`

packages/integrations/vercel/src/lib/web-analytics.ts
```diff
@@ -0,0 +1,56 @@
...
+async function getWebAnalyticsFunctions(root: URL) {
+	try {
+		const files = await Promise.all([
+			import(/* @vite-ignore */ fileURLToPath(new URL('./vercel-web-analytics.ts', root))).catch(
+				() => undefined
+			),
+			import(/* @vite-ignore */ fileURLToPath(new URL('./vercel-web-analytics.js', root))).catch(
+				() => undefined
+			),
+		]);
```
ematipico, line 18: I believe it's healthier first to check if these files exist on disk e.g. [`fs.existSync`](https://nodejs.org/api/fs.html#fsexistssyncpath), and then import the ones that exist.

Also, we should log a warning if **both** files exist. 

So we don't need to rely on `Promise.all` and a `try/catch`.

packages/integrations/vercel/src/lib/web-analytics.ts
```diff
@@ -0,0 +1,56 @@
...
+
+		const functions = files[0] || files[1];
+
+		if (functions?.default) {
+			if (typeof functions.default.beforeSend !== 'function') {
```
ematipico, line 23: The example that you wrote doesn't export a `default`. I think it's best **to not rely on `default`**. Without a `default`, we can make our APIs more flexible if, in the future, we want to add more exports.

packages/integrations/vercel/src/lib/web-analytics.ts
```diff
@@ -0,0 +1,56 @@
...
+				throw new Error(
+					`@astrojs/vercel: ./vercel-web-analytics.js should export a \`beforeSend\` function.`
+				);
```
ematipico, line 26: Astro now exposes `AstroError`, via:

```js
import { AstroError } from "astro/errors";
```

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -107,12 +116,23 @@ export default function vercelServerless({
 	return {
 		name: PACKAGE_NAME,
 		hooks: {
-			'astro:config:setup': ({ command, config, updateConfig, injectScript }) => {
-				if (command === 'build' && analytics) {
-					injectScript('page', 'import "@astrojs/vercel/analytics"');
+			'astro:config:setup': async ({ command, config, updateConfig, injectScript }) => {
+				if (webAnalytics?.enabled) {
+					injectScript(
+						'page',
```
ematipico, line 122: Does that mean that analytics are not used for endpoints and redirects? If this is intentional, I think we should document it.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -107,12 +116,23 @@ export default function vercelServerless({
...
+							{
+								...webAnalytics.config,
+								mode: command === 'dev' ? 'development' : 'production',
+							},
+							config.root
+						)
+					);
+				}
+				if (command === 'build' && speedInsights?.enabled) {
+					injectScript('page', 'import "@astrojs/vercel/speed-insights"');
```
ematipico, line 133:
```suggestion
					injectScript('page', 'import "@astrojs/vercel/speed-insights.js"');
```

Extension is important in an ESM environment.

ematipico (changes requested):
A couple of essential things we should address:

- This change removes an option **without deprecating it** first.. This is important for us, especially for libraries that we directly maintain. We should keep `analytics` and keep backwards compatibility for some weeks/months.
- This change **doesn't** provide any migration guide form the **old** option. We have to write this migration inside the changelog.

packages/integrations/vercel/README.md
chriswdmr: Makes sense. Updated!

packages/integrations/vercel/src/edge/adapter.ts
chriswdmr: Oh I missed that somehow, removed. Thanks!

packages/integrations/vercel/src/lib/web-analytics.ts
chriswdmr: Thanks for the heads up. I was not using `default` at all, I think it was a left over from testing. Totally agree, hence why it's not using `default`

packages/integrations/vercel/src/lib/web-analytics.ts
chriswdmr: Amazing, switched it to AstroError!

packages/integrations/vercel/src/serverless/adapter.ts
chriswdmr: That's how analytics typically works. It requires a client side script, hence why we can't track these things.

chriswdmr:
@ematipico thanks for the review!

> This change removes an option **without deprecating it** first. This is important for us, especially for libraries that we directly maintain. We should keep `analytics` and backwards compatibility for some weeks/months.
Makes sense, I'll update the PR. Is it fine if I add a `logger.warn` when it's used, to point users to the new options?

> * This change **doesn't** provide any migration guide form the **old** option. We have to write this migration inside the changelog.
Sorry that I missed that. Will add it to the changelog.

matthewp:
@chriswdmr a warning would be very much appreciated for this.

packages/integrations/vercel/src/lib/web-analytics.ts
chriswdmr: Amazing, didn't know that. Updated!

chriswdmr:
@matthewp @ematipico I've updated the PR based on your feedback.
I brought back the `analytics` property and added a warning for it when used.

packages/integrations/vercel/src/serverless/adapter.ts
ematipico: Thank you! :)

packages/integrations/vercel/src/serverless/adapter.ts
ematipico: Oh, it's in the `package.json`! I missed that. So, in practice, it should be fine as is. :) You can ignore my suggestion.

ematipico:
Thank you @chriswdmr ! We'll keep the PR open for a few days, we have a big release coming up, so we want to avoid merging big PRs unless they are important fixes. Please bear with us for a little while.

packages/integrations/vercel/README.md
```diff
@@ -85,13 +85,14 @@ vercel deploy --prebuilt
...
-**Type:** `boolean`<br>
-**Available for:** Serverless, Static<br>
-**Added in:** `@astrojs/vercel@3.1.0`
+**Type:** `VercelWebAnalyticsConfig`<br>
+**Available for:** Serverless, Edge, Static<br>
+**Added in:** `@astrojs/vercel@3.8.0`
 
-You can enable [Vercel Analytics](https://vercel.com/analytics) (including Web Vitals and Audiences) by setting `analytics: true`. This will inject Vercel’s tracking scripts into all your pages.
+You can enable [Vercel Web Analytics](https://vercel.com/docs/concepts/analytics) by setting `webAnalytics: { enabled: true }`. This will inject Vercel’s tracking scripts into all of your pages.
+Alternatively, you can also pass [configuration options](https://vercel.com/docs/concepts/analytics/package) (except functions like `beforeSend`) via the `config` property inside `webAnalytics`.
```
sarah11918, line 95:
```suggestion
You can also pass [Vercel's analytics configuration options](https://vercel.com/docs/concepts/analytics/package) `mode` and `debug` via the `config` property inside `webAnalytics`.
```
- Checking on "alternatively" because I wouldn't think that passing an additional config option is an *alternative* to enabling `webAnalytics`? Should this perhaps be "Additionally" (or, nothing at all, since there's already an "also"?)
- Following that link, there really are only three config options mentioned, so assuming that's all there is (?) it I think we can just state the two which *can* be passed here. Then, there is a section below for how to handle `beforeSend`.

packages/integrations/vercel/README.md
```diff
@@ -101,7 +102,50 @@ import vercel from '@astrojs/vercel/serverless';
...
+      enabled: true,
+    },
+  }),
+});
+```
+
+#### `beforeSend`
+
+To define the `beforeSend` function, you need to create a separate file inside your root called `vercel-web-analytics.ts`.
+If you're not using TypeScript, you can define the function inside `vercel-web-analytics.js`.
```
sarah11918, line 115:
```suggestion
`beforeSend` is a function that can modify analytics event data before it's sent to Vercel.

Define this function in a separate file at the root of your project called `vercel-web-analytics.ts` (or `vercel-web-analytics.js` if you're not using TypeScript).
```

You've said in the section above not to pass `beforeSend` as a config option. Does this mean that simply having this file at the root of your project is enough? The Vercel adapter will automatically find it and use it? Is there anything else that needs to happen other than creating this file?

sarah11918 (commented):
Thanks for the docs contributions for this @chriswdmr ! It's quite comprehensive and well done! Just two comments from me re: the README below!

@ematipico has already give you some great docs advice, and I'm sure whatever final decisions you and he make will be awesome!  🙌

packages/integrations/vercel/README.md
chriswdmr: Exactly, the user just needs to create the file and export the `beforeSend` function there

packages/integrations/vercel/README.md
chriswdmr: Thanks for the callout, you're right. I've updated the copy!

sarah11918 (approved):
Approving on behalf of docs!

matthewp:
!preview vercel-upgrade

matthewp:
!preview vercel-upgrade

matthewp:
I had to create a new PR, I think because this is from the fork's `main` branch. Anyways, this is now released as version `0.0.0-vercel-speed-insights-20230905180431`, or install with `npm install @astrojs/vercel@next--vercel-speed-insights`

.changeset/sixty-teachers-tap.md
```diff
@@ -0,0 +1,26 @@
...
+// astro.config.mjs
+export default defineConfig({
+	adapter: vercel({
+-   analytics: true,
++		webAnalytics: {
++			enabled: true
++		},
++		speedInsights: {
++			enabled: true
++		}
```
ematipico, line 20:
```suggestion
-		analytics: true,
+		webAnalytics: {
+			enabled: true
+		},
+		speedInsights: {
+			enabled: true
+		}
```

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -124,12 +138,25 @@ export default function vercelServerless({
...
 		name: PACKAGE_NAME,
 		hooks: {
-			'astro:config:setup': ({ command, config, updateConfig, injectScript }) => {
-				if (command === 'build' && analytics) {
-					injectScript('page', 'import "@astrojs/vercel/analytics"');
+			'astro:config:setup': async ({ command, config, updateConfig, injectScript, logger }) => {
+				if (webAnalytics?.enabled || analytics) {
+					if (analytics) {
+						logger.warn(
+							`@astrojs/vercel: the \`analytics\` property is deprecated. Please use the new \`webAnalytics\` and \`speedInsights\` properties instead.`
```
ematipico, line 145: You don't need `@astrojs/vercel`. The logger will print the message with the name of the adapter already: https://docs.astro.build/en/reference/integrations-reference/#astrointegrationlogger

packages/integrations/vercel/src/static/adapter.ts
```diff
@@ -44,20 +58,33 @@ export default function vercelStatic({
...
 		name: '@astrojs/vercel',
 		hooks: {
-			'astro:config:setup': ({ command, config, injectScript, updateConfig }) => {
-				if (command === 'build' && analytics) {
-					injectScript('page', 'import "@astrojs/vercel/analytics"');
+			'astro:config:setup': async ({ command, config, injectScript, updateConfig, logger }) => {
+				if (webAnalytics?.enabled || analytics) {
+					if (analytics) {
+						logger.warn(
+							`@astrojs/vercel: the \`analytics\` property is deprecated. Please use the new \`webAnalytics\` and \`speedInsights\` properties instead.`
```
ematipico, line 65: Same as the other comment, `@astrojs/vercel` is not needed anymore.

ematipico (approved):

.changeset/sixty-teachers-tap.md
```diff
@@ -0,0 +1,26 @@
+---
+'@astrojs/vercel': major
```
bluwy, line 2: Since we're only deprecating the `analytics` option now, could this be marked as a minor instead?

matthewp (changes requested):
Blocking for now as this is still being tested.

matthewp (approved):

matthewp (approved):

PR #8022, opened 2023-08-10
Fix getCollection return new instance in prod

bluwy:
## Changes

Fixes https://github.com/withastro/roadmap/discussions/604

This was planned for 3.0, but I'm not sure if we need to, since the bug only happen in prod. For now I'm pointing this to merge into `main`.

## Testing

Added a new test.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/smart-nails-push.md
+++ b/.changeset/smart-nails-push.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
Awesome! Glad you figured out this was a dev/prod inconsistency and can change on a `patch`

PR #8025, opened 2023-08-10
Remove unnecessary hyphen from svg tag

morinokami:
## Changes

- Remove unnecessary hyphen from svg tag in examples/blog

## Testing

Ran the blog example locally and confirmed that the appearance of the page did not change

## Docs

N/A because this is an update of example


natemoo-re (approved):
Thanks!

PR #8026, opened 2023-08-10
Update path to image in public/ to suppress warning

morinokami:
## Changes

- Update path to image in public/ to suppress warning (`files in the public directory are served at the root path. ...`)

## Testing

Ran the blog example locally and confirmed that the warning disappeared and the appearance of the page did not change

## Docs

N/A because this is an update of example


natemoo-re (approved):
Thank you!

PR #8027, opened 2023-08-10
Ensure dev server restart respects `base` removal

### LINKED ISSUE #6067: The link to public fails if you put the parameter "base" and then remove it

2.0.2

No

Tested with npm and pnpm

Windows

### Describe the Bug

The problem is that once you put the "base" parameter in the astro.config.mjs file, if you later remove it, it still keeps the base prefix.

Steps to reproduce:
1. Clone Minimal reproducible example or Create a basic app with astro.
`npm create astro@latest`
2. If you have created a new app add any svg or any other file in the "public" folder
step2
3. Use the file in any astro file.
step3
4. You run astro with "npm run dev" and everything works.
5. In the file "astro.config.mjs" you add the base parameter with any value.
6. It will warn you that you must change the url adding the value that you have entered in "base" so that it works in production. But it still works.
7. Remove the base parameter from the "astro.config.mjs" file
and it stops working as it still requires the "base". It shows a 404 code in the console.
step7
*Click on image to better view*

https://github.com/alejandrofan2/astro-bug-base

### Participation

natemoo-re:
## Changes

- Closes #6067
- The `base` middleware should always be injected, even if the base is set to the default
- Otherwise, dev server restarts don't respect when `base` is removed

## Testing

Tested manually since it's HMR/restart behavior

## Docs

N/A, bug fix only


packages/astro/src/template/4xx.ts
```diff
@@ -1,5 +1,4 @@
 import { escape } from 'html-escaper';
-import { baseCSS } from './css.js';
```
natemoo-re, line 2: This is unrelated cleanup, but I was in this file so it made sense

matthewp (approved):

PR #8031, opened 2023-08-10
fix(prerendering): never prerender custom 404 or 500 routes

### LINKED ISSUE #8005: Prerended custom 500 page does not include imported stylesheets

2.10.3

Node

npm

Mac

Chrome, Firefox, Safari

### Describe the Bug

Prerended custom 500 page does not include imported stylesheets in production mode. Changing prerender to false, fixes it. 

Stylesheets are imported correctly inside custom 500 page.

https://stackblitz.com/edit/github-pt4enm-cr35nf

### Participation

lilnasy:
## Changes
- Closes #8005
- Adds a check whether the route being prerendered is a custom 404 or 500 one.

## Future Work
- Custom 404 and 500 routes don't work in dev mode at all.

## Testing
Pending

## Docs
Does not affect usage


lilnasy:
To be clear, these routes were always rendered just-in-time.

This change prevents the manifest corruption that occurs when prerendering was attempted.

matthewp:
sgtm

lilnasy:
Astro core doesn't support prerendering 404.astro, it is prerendered into 404.html but the router renders it just-in-time anyway. However, the netlify adapter is able to achieve the effect by using `_redirects`, so it would be a regression to consider it ineligible for prerendering.

Only considering 500.astro ineligible still solves the immediate issue.

I think actually prerendering 404 and 500 routes can be revisited after the "pipeline" refactors.

PR #8035, opened 2023-08-11
fix: Removed extra double quotes from computed style in shiki code component

fyndor:
## Changes

Removed an invalid extra double quote that is added to the shiki wrapper component called Code.

## Testing

Sorry, I don't have the time to learn how to properly create unit tests for this within your ecosystem.  Sorry if this is unacceptable.  The test I would add I guess would be whether the style attribute is properly formed without the extra double quote.  I am not sure how to do that.

## Docs

I would assume no document changes should be needed for this.  It currently creates malformed html, but at least Chrome still renders properly even with the malformed HTML.  This should not affect the functionality other than possibly make stricter browsers (if they exist) work properly.

### DIFF
--- a/.changeset/tall-owls-act.md
+++ b/.changeset/tall-owls-act.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Removed extra double quotes from computed style in shiki code component
--- a/.changeset/tall-owls-act.md
+++ b/.changeset/tall-owls-act.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 Removed extra double quotes from computed style in shiki code component
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -74,9 +74,9 @@ const html = renderToHtml(tokens, {
 			// Handle code wrapping
 			// if wrap=null, do nothing.
 			if (wrap === false) {
-				style += '; overflow-x: auto;"';
+				style += '; overflow-x: auto;';
 			} else if (wrap === true) {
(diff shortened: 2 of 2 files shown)


fyndor:
Not sure what that CI error means

matthewp:
In what case does this create malformed HTML? Can you post an example?

fyndor:
I would say in all cases, but I guess technically if you pass null to the wrap prop then it wouldn't cause malformed html. Not sure the purpose of wrap == null.

<Code code="some code" />
<Code code="some code" wrap={false} />
<Code code="some code" wrap={true} />

These should all add an extra double quote in html.  But I guess the following would lead to valid html:

<Code code="some code" wrap={null} />

The code surrounding adding extra styling depending on the wrap prop value adds a double quote at the end of the appended style string.  Then in the return statement, the style string variable is inserted between two double quotes.  So if wrap != null, then there is always an extra double quote at the end of the style attribute.  If you check out the commit, the It's pretty obvious there is an erroneous double quote at the end of those two lines.  I was looking at the outputted html when I discovered the issue then backtracked it to the code fixed by this commit.

.changeset/tall-owls-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
bluwy, line 2:
```suggestion
'astro': patch
```

bluwy (approved):
Looks like a simple fix to me. Thanks for catching this!

fyndor:
No problem.  This might be my first non-documentation open source contribution :D  Glad to help improve Astro in my small way.  It's a great project.

PR #8042, opened 2023-08-11
Fix same-page transition with different query params

matthewp:
## Changes

- When navigating we prevent same-page transition normally. But when there are search params we should treat that as a different page and have a transition.

## Testing

- Test case added

## Docs

N/A, bug fix


FredKSchott (approved):
drive-by LGTM

PR #8051, opened 2023-08-13
fix: Cleanup blog template images

Greenheart:
## Changes

* Remove old unused image from the blog starter template.
* Optimize image size for `blog-placeholder-5.jpg` (-94% size reduction)

(This contains no changeset as per instructions in https://github.com/withastro/astro/blob/main/CONTRIBUTING.md#making-a-pull-request)

## Testing

* Verified that the blog stater template works as before by ensuring all five images display correctly at http://localhost:3000/blog when running the blog starter template
* The directory size should be slightly smaller than before.

## Docs

Not applicable since this is just a content change for the blog template.


bluwy (approved):
Looks great. Thanks!

PR #8066, opened 2023-08-14
feat(assets): Add support for passing non-awaited imports to the Image component and `getImage`

Princesseuh:
## Changes

What the title says. This is a pretty common request since when using inline imports, it's super annoying to have to await + `.default` it, and especially when using components in MDX and stuff, you kinda want your code to be as neat as possible. Also, this worked in `@astrojs/image` and people wanted it back

## Testing

Added a test

## Docs

N/A, the proper path is already documented, this is something people might stumble upon

### DIFF
--- a/.changeset/odd-plants-tie.md
+++ b/.changeset/odd-plants-tie.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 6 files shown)


matthewp (approved):

natemoo-re (approved):
No brainer! Thanks for adding this!

Issue #7973, opened 2023-08-06
Vercel adapter + prerender + build split leads to an error


raulfdm:
### What version of `astro` are you using?

2.10.1

### Are you using an SSR adapter? If so, which one?

Vercel

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Doesn't matter

### Describe the Bug

While trying to use Vercel adapter and `build.split = true`, pages with `export const prerender = true` throws the following error:

```
 prerendering static routes 
 error   Unable to find the manifest for the module file:///path/to/project/astro-vercel-build-split-prerender-error/apps/website/dist/src/pages/_lang_/entry.index.astro.mjs?time=1691298936496. This is unexpected and likely a bug in Astro, please report.
Error: Unable to find the manifest for the module file:///path/to/project/astro-vercel-build-split-prerender-error/apps/website/dist/src/pages/_lang_/entry.index.astro.mjs?time=1691298936496. This is unexpected and likely a bug in Astro, please report.
```

### What's the expected result?

Build without error or have instructions about pre-rendered pages.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qfges3

### Participation

- [ ] I am willing to submit a pull request for this issue.

raulfdm:
Related https://github.com/withastro/astro/issues/7753

AkashRajpurohit:
FYI this is not specific to Vercel adapter. Happening for Cloudflare adapter as well.

ematipico:
@raulfdm can you check if 2.10.10 fixes your issue?

raulfdm:
@ematipico I've updated my stackblitz repo to latest version of astro and astro vercel:

```json
  "dependencies": {
    "@astrojs/vercel": "3.8.2",
    "astro": "2.10.14"
  }
```

The error remains:
![CleanShot 2023-08-28 at 15 53 25](https://github.com/withastro/astro/assets/12464600/7e8b1651-b4cb-4ec5-8969-6406038fdd56)

raulfdm:
Well... the release will happen around the next days, right? As soon as it lands I'll check this out... thanks for getting back with this.

jacobthesheep:
Just checked. No errors.

ematipico:
Thank you! Closing as resolved

PR #8070, opened 2023-08-14
Prerendering corner cases

### LINKED ISSUE #8005: Prerended custom 500 page does not include imported stylesheets

lilnasy:
## Changes
- closes #8005 
- maybe also #7973
- [x] Makes sure prerendered pages are turned into no-ops
- [ ] Uses prerendered 500 routes using the approach introduced in #7754 

## Testing
- Adjusted `node/prerender-404` to actually check that it prerendered
- Added one that checks styles/scripts are included for 500 routes

## Docs
Does not affect usage


packages/astro/src/core/build/plugins/plugin-internals.ts
```diff
@@ -41,6 +41,7 @@ export function vitePluginInternals(input: Set<string>, internals: BuildInternal
 			}
 			await Promise.all(promises);
 			for (const [, chunk] of Object.entries(bundle)) {
+				// why is the distinction important?
 				if (chunk.type === 'chunk' && chunk.facadeModuleId) {
```
lilnasy, line 45: <details>
<summary>how it's used in the codebase</summary>

For distinguishing which chunks get added to `internals.pageToBundleMap` and which to `internals.entrySpecifierToBundleMap`. Only the chunks in `pageToBundleMap` are considered when turning prerendered pages into no-op modules.

</details>

`facadeModuleId` seems to be a barely documented property added by rollup. Sometimes, whether or not an astro page gets it depends on whether an unrelated page was prerendered. The end result is, we are not reliably turning prerendered components into no-ops.

matthewp: I think this is checking of the chunk is an entrypoint. Usually a page, I believe.

lilnasy:
The main work should be done. Testing is botched.

#7754 uses a [hack](https://github.com/withastro/astro/pull/7754/files#diff-d5f3e4eb9c050d49b171e7135d21ba49da7321efded2db0a486280d00118b1c4R11) to prevent node from reusing a cached `entry.mjs` but I think its now reusing a cached dependency of it.

lilnasy:
Ready for review.

The change did not fix #7973 unfortunately.

natemoo-re (approved):
This looks great, thank you so much for digging into this!

matthewp (approved):

PR #8072, opened 2023-08-14
Update types to specify that compress defaults to true

matthewp:
## Changes

- The option was already changed to default to true, but the types were not changed. This updates them, as well as the docs.

## Testing

N/A

## Docs

- In the types file

### DIFF
--- a/.changeset/tasty-camels-speak.md
+++ b/.changeset/tasty-camels-speak.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):

ElianCodes (approved):
Docs look good to me!

sarah11918 (approved):
Thanks Matthew! Just committed a change directly because I couldn't suggest over multiple lines. Happy with this now!

PR #8075, opened 2023-08-14
fix: prevent null prefixed results from React.useId when using client:only directive

SudoCat:
There was a low-impact bug with the unique id generation fix, causing client:only ids to be generated with `null`, e.g. `:nullr0:` 

The ids generated are actually still unique, and avoid clashes with server-side, but it was flagged by a team member as a possible bug, so I felt an upstream fix was wise.

## Changes

- removes the identifierPrefix from client:only components, to prevent client:only useId calls from being prefixed with `null`

## Testing

- quick test in browser via linking
- additional e2e test case added to validate clientside results are unique, and do not clash with static or hydrated server-rendered results.

## Docs

No docs required, no change to user behaviour.

### DIFF
--- a/.changeset/thin-plums-drop.md
+++ b/.changeset/thin-plums-drop.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+fix a bug where react identifierPrefix was set to null for client:only components causing React.useId to generate ids prefixed with null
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
Thanks for the fix, really appreciate the PR!

PR #4683, opened 2022-09-08
Fix TypeScript compilation errors

Princesseuh:
## Changes

Running `tsc` in Astro projects without `skipLibCheck` would fail because of some TS errors inside Astro. This fix those

## Testing

Tested manually 

## Docs

N/A

PR #8076, opened 2023-08-14
Refactor dependency usage

bluwy:
## Changes

- Move type-only deps to devdeps
- Remove unused dev deps
- `pnpm` auto sort dep names

## Testing

Existing tests should pass.

## Docs

n/a. Shouldn't have visible changes.


packages/astro/package.json
```diff
@@ -127,8 +127,6 @@
     "@babel/traverse": "^7.22.5",
     "@babel/types": "^7.22.5",
     "@types/babel__core": "^7.20.1",
-    "@types/dom-view-transitions": "^1.0.1",
-    "@types/yargs-parser": "^21.0.0",
```
Princesseuh, line 131: I believe this one is not a devdep unless something changed, see: https://github.com/withastro/astro/pull/4683

bluwy: Ah I'll have to check that. I was hoping if it's exposed but not public, the types would be `any` but the user would never see it anyways.

ematipico (approved):

packages/astro/package.json
bluwy: I don't see an error when using this branch locally for `astro check`. If I manually set `skipLibCheck: false` (Astro defaults to true), and run with Astro stable latest and this branch, and run `tsc` instead, I get this error in both versions:

image

Among many other errors (and including Vite). Maybe the `tsc` results are expected. So in the case of `astro check` we're safe here?

PR #8077, opened 2023-08-14
Reduce `create-astro` dependencies

natemoo-re:
## Changes

- Fixes PLT-777
- Moves `mocha` and `chai` to `devDependencies`
- Updates `giget` since https://github.com/unjs/giget/issues/81 was merged in February, but we never updated
- Moves from `execa` to a thin `child_process` wrapper since we don't need more advanced features
- Included as a minor for safety. Internal refactor, but better safe than sorry.

## Testing

Existing tests pass

## Docs

N/A, maintenance only


packages/create-astro/test/git.test.js
```diff
@@ -46,3 +29,29 @@ describe('git', () => {
...
+		};
+		await git(context);
+
+		expect(fixture.hasMessage('Git has already been initialized')).to.be.true;
+	});
+
+	after(() => {
+		rmSync(dir, { recursive: true, force: true });
+	})
+})
```
natemoo-re, line 57: Updated this test just for clarity

packages/create-astro/package.json
```diff
@@ -32,15 +32,14 @@
...
-    "giget": "1.0.0",
-    "mocha": "^9.2.2",
+    "giget": "^1.1.2",
     "node-fetch-native": "^1.2.0",
     "which-pm-runs": "^1.1.0"
   },
   "devDependencies": {
     "@types/which-pm-runs": "^1.0.0",
+    "chai": "^4.3.7",
+    "mocha": "^9.2.2",
```
natemoo-re, line 42: These should have always been dev deps!

packages/create-astro/package.json
```diff
@@ -32,15 +32,14 @@
   "//b": "DEPENDENCIES IS FOR UNBUNDLED PACKAGES",
   "dependencies": {
     "@astrojs/cli-kit": "^0.2.3",
-    "chai": "^4.3.7",
-    "execa": "^6.1.0",
```
natemoo-re, line 36: Removed this!

packages/create-astro/src/shell.ts
```diff
@@ -0,0 +1,44 @@
+// This is an extremely simplified version of [`execa`](https://github.com/sindresorhus/execa)
+// intended to keep our dependency size down
```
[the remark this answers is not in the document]
Princesseuh: Doing this for Astro itself would probably be also great

matthewp (approved):

packages/create-astro/src/shell.ts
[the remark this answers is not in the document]
natemoo-re: Agreed! We can follow-up on that

PR #8078, opened 2023-08-14
feature(@astrojs/cloudflare): port functionPerRoute

alexanderniebuhr:
## Changes

- it does keep support build.split, until it is fully removed
- ports https://github.com/withastro/astro/pull/7509 to next branch _(since the bugfix was missing in next)_
- ports https://github.com/withastro/astro/pull/7839 to cloudflare adapter
- closes #7948

## Testing

- see ported test

## Docs

Since the changelog is copied from an already merged PR, I don't think we need to review it again

PR #8081, opened 2023-08-14
Update wrangler

natemoo-re:
## Changes

- `wrangler` dependency was quite old, potentially causing flaky tests in https://github.com/withastro/astro/pull/8078
- updates devDependency to latest

## Testing

We'll see

## Docs

N/A, chore


alexanderniebuhr:
This would work:

https://discord.com/channels/830184174198718474/845430950191038464/1140741487710187570

packages/integrations/cloudflare/test/test-utils.js
```diff
@@ -21,20 +22,30 @@ const wranglerPath = fileURLToPath(
...
+ * @returns {Promise<WranglerCLI>}
  */
-export function runCLI(basePath, { silent, port = 8787 }) {
+export async function runCLI(basePath, { silent, port }) {
+	// Hack: force existing process on port to be killed
+	try {
+		await kill(port, 'tcp');
+	} catch {
+		// Will throw if port is not in use, but that's fine
+	}
```
natemoo-re, line 33: This was needed because apparently `wrangler` didn't always exit cleanly

ematipico (approved):

alexanderniebuhr (commented):
Just to make sure we don't forget it, this also needs to be opened again with base `next` too.
Because we want the change also in `next`

bluwy (approved):

packages/integrations/cloudflare/test/test-utils.js
bluwy: Nice fix. I also tried updating wrangler the last time but was stuck with this. Strange that wrangler doesn't properly teardown its server sometimes.

bluwy:
> Just to make sure we don't forget it, this also needs to be opened again with base `next` too.
> Because we want the change also in `next`

`next` is frequently updated from `main` so it'll get the changes soon too.

PR #8082, opened 2023-08-14
Add experimentalReactChildren option to React integration

matthewp:
## Changes

- Adds a new option `experimentalReactChildren` option to the React integration.
- A proposed solution for https://github.com/withastro/roadmap/discussions/465. Want to release as experimental to see if it works for the MDX use-case and to see if there are any downsides. Plan would be to keep it an option in the future.

## Testing

- Moved the existing React test over
- Updated to test that the children contain a `length` which is something you don't get without the option.

## Docs

- In the readme


packages/integrations/react/test/react-component.test.js
```diff
@@ -99,12 +99,18 @@ describe('React Components', () => {
 			const $ = cheerioLoad(html);
 			expect($('#cloned').text()).to.equal('Cloned With Props');
 		});
+
+		it('Children are parsed as React components, can be manipulated', async () => {
+			const html = await fixture.readFile('/children/index.html');
+			const $ = cheerioLoad(html);
+			expect($(".with-children-count").text()).to.equal('1');
```
natemoo-re, line 106: Children count would always be `1`, even with the option enabled. 

We should probably include a test where the React component has multiple children. The children count will be `1` without the experimental flag enabled, and `N` with it enabled.

natemoo-re (commented):
Looking good!

packages/integrations/react/server.js
```diff
@@ -43,6 +44,12 @@ async function check(Component, props, children) {
 		return React.createElement('div');
 	}
 
+	if (props['use:vnode']) {
+		const convert = await import('./vnode-children.js').then(mod => mod.default);
+		delete props['use:vnode'];
+		children = convert(children);
+	}
+
```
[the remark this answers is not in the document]
matthewp: Nope, you're right.

packages/integrations/react/test/react-component.test.js
matthewp: With out it enabled children is an object with no length. But I agree having multiple children is a better test so I'll update.

matthewp: Updated this test to have multiple children, better captures the intent of the test.

.changeset/yellow-snakes-jam.md
```diff
@@ -0,0 +1,21 @@
+---
+'@astrojs/react': minor
+---
+
+Optionally parse React slots as React children
+
+This adds a new flag `experimentalReactChildren` which can be set in the React component like so:
```
sarah11918, line 7:
```suggestion
Optionally parse React slots as React children.

This adds a new configuration option for the React integration `experimentalReactChildren`:
```

packages/integrations/react/README.md
```diff
@@ -61,6 +61,44 @@ To use your first React component in Astro, head to our [UI framework documentat
 - 💧 client-side hydration options, and
 - 🤝 opportunities to mix and nest frameworks together
 
+## Options
+
+### Children parsing
+
+When you pass children into a React component from an Astro component you only see a single child on the React side:
```
sarah11918, line 68:
```suggestion
Children passed into a React component from an Astro component are parsed as plain strings, not React nodes.

For example, the `<ReactComponent />` below will only receive a single child element:
```
I think we can tighten this explanation up? This might be a bit *too* terse? Make it accurate, but I think this kind of structure would work nicely!

packages/integrations/react/README.md
```diff
@@ -61,6 +61,44 @@ To use your first React component in Astro, head to our [UI framework documentat
...
+  <div>one</div>
+  <div>two</div>
+</ReactComponent>
+```
+
+This is because the children being passed to *are not* React nodes. Astro treats its templates as plain strings.
+
+If you are using a library that *expects* to be passed more than 1 child element; for example so that it can slot certain elements in different places, you might find this to be a blocker. You can set the experimental flag `experimentalReactChildren` to tell Astro to always pass children to React as React vnodes.
+
+There is some runtime cost to this, but it can help with compatibility. You can enable this option with:
```
sarah11918, line 85:
```suggestion
If you are using a library that *expects* more than one child element element to be passed, for example so that it can slot certain elements in different places, you might find this to be a blocker.

You can set the experimental flag `experimentalReactChildren` to tell Astro to always pass children to React as React vnodes. There is some runtime cost to this, but it can help with compatibility. 

You can enable this option in the configuration for the React integration:
```
Just separated out the ideas a little differently.

sarah11918 (commented):
Just made a couple of quick suggestions for your consideration, @matthewp !

sarah11918:
@matthewp - Normally we'd *also* expect to see a PR to config reference to document a new experimental flag. Should we do that here? Is there a reason not to?

PR #8085, opened 2023-08-15
Export astro/compiler-runtime and cleanup exports

bluwy:
## Changes

(Same as changeset)

Remove exports for `astro/internal/*` and `astro/runtime/server/*` in favour of `astro/runtime/*`. Add new `astro/compiler-runtime` export for compiler-specific runtime code.

Ultimately it would be nice to get rid of `astro/runtime/*` and make it all stricter, but that's longer term.

## Testing

Existing tests should pass.

## Docs

We currently don't reference these exports in the docs, which is great since they are internal APIs. The `@astrojs/compiler` README can use an update after 3.0 is released to recommend `astro/compiler-runtime` over `astro/runtime/server/index.js`

### DIFF
--- a/.changeset/neat-owls-run.md
+++ b/.changeset/neat-owls-run.md
@@ -1,0 +1,25 @@
+---
+'astro': major
+---
+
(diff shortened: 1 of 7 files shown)


ematipico (approved):

bluwy:
Merging so Ema can rebase next

PR #8086, opened 2023-08-15
Remove deepmerge-ts for simpler implementation

bluwy:
## Changes

Implement a simpler merge function (based on Astro's mergeConfig utility but simplified) to prevent needing to use a dep.

## Testing

Tested with a project running `astro add preact`, `astro add solid-js`, etc. Making sure the `tsconfig.json` updates still work.

## Docs

n/a. internal change.

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -137,7 +137,6 @@
     "common-ancestor-path": "^1.0.1",
     "cookie": "^0.5.0",
     "debug": "^4.3.4",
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
Sad to see it go, because `deepmerge-ts` is a really nice package, but happy for the dep reduction. `tsconfig.json` never gets super super complex so I assume we didn't need the extra features anyway.

PR #8088, opened 2023-08-15
refactor: build pipeline

ematipico:
## Changes

This PR refactors the build process using the `Pipeline` class.

Here's a list of significant changes:
- created a new `plugin-manifest` plugin to generate the manifest. The `manifest` object is now emitted in a separate file called `manifest.mjs`. The file is emitted in the `config.build.server` folder. This is an internal refactoring. I went with this change because the manifest was repeated for each emitted entry/route, giving the impression that a manifest belongs to an entry/route. This isn't correct. A manifest belongs to the whole app. It made more sense to have a separate file with this information. 
- created a new `BuildPipeline` class that extends `Pipeline`, which is responsible for giving all the essential information to its consumers; for example, `BuildInternals` and `StaticOuputOptions` are now inside `BuildPipeline` and with a series of getters, the consumers retrieve the information needed.
- changed how we retrieve the pages to render without using generators. Even though it wasn't necessary, now we reduced the number of loops.

## Testing

Current tests should pass.

## Docs

N/A internal refactor

### DIFF
--- a/packages/astro/src/core/build/buildPipeline.ts
+++ b/packages/astro/src/core/build/buildPipeline.ts
@@ -103,15 +103,21 @@ export class BuildPipeline extends Pipeline {
 			);
 		}
 
-		const renderersEntryUrl = new URL('renderers.mjs', staticBuildOptions.settings.config.outDir);
+		const renderersEntryUrl = new URL(
+			'renderers.mjs',
+			staticBuildOptions.settings.config.build.server
+		);
 		const renderers: SSRLoadedRenderer[] | undefined = await import(renderersEntryUrl.toString());
 		if (!renderers) {
 			throw new Error(
 				"Astro couldn't find the emitted renderers. This is an internal error, please file an issue."
 			);
 		}
-		manifest.renderers = renderers;
-		return manifest;
+		// manifest.renderers = renderers;
+		return {
+			...manifest,
+			renderers,
+		};
 	}
 
 	/**
--- a/packages/astro/src/core/build/buildPipeline.ts
+++ b/packages/astro/src/core/build/buildPipeline.ts
@@ -6,7 +6,7 @@ import { RESOLVED_SPLIT_MODULE_ID } from './plugins/plugin-ssr.js';
 import { ASTRO_PAGE_EXTENSION_POST_PATTERN } from './plugins/util.js';
 import type { SSRManifest } from '../app/types';
 import type { AstroConfig, AstroSettings, SSRLoadedRenderer } from '../../@types/astro';
-import { isServerLikeOutput } from '../../prerender/utils.js';
+import { getOutputDirectory, isServerLikeOutput } from '../../prerender/utils.js';
 import type { EndpointCallResult } from '../endpoint';
 import { createEnvironment } from '../render/index.js';
 import { BEFORE_HYDRATION_SCRIPT_ID } from '../../vite-plugin-scripts/index.js';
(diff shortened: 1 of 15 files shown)


packages/astro/src/core/build/buildPipeline.ts
```diff
@@ -0,0 +1,176 @@
...
+			return new Response(body, {
+				headers: { ...response.response.headers },
+			});
+		} else {
+			if (response.encoding) {
+				const headers = new Headers();
+				headers.set('X-Astro-Encoding', response.encoding);
+				return new Response(response.body, {
+					headers,
+				});
```
ematipico, line 170: I thought we could store the `encoding` coming from a `simple` response using an internal header. What do you think? When generating a page, we read it and use it when we write the file to file system.

matthewp (approved):

natemoo-re (approved):
Looks great! Glad to see the manifest changes and a whole bunch of cleanup.

packages/astro/src/core/build/plugins/plugin-manifest.ts
```diff
@@ -0,0 +1,243 @@
...
+		resolveId(id) {
+			if (id === SSR_MANIFEST_VIRTUAL_MODULE_ID) {
+				return RESOLVED_SSR_MANIFEST_VIRTUAL_MODULE_ID;
+			}
+		},
+		async load(id) {
+			if (id === RESOLVED_SSR_MANIFEST_VIRTUAL_MODULE_ID) {
+				const imports = [];
+				const contents = [];
+				const exports = [];
```
bluwy, line 38: Nit: we could type this as `string[]`

packages/astro/src/core/build/generate.ts
```diff
@@ -64,6 +58,8 @@ import type {
 	StylesheetAsset,
 } from './types';
 import { getTimeStat } from './util.js';
+import { BuildPipeline } from './buildPipeline.js';
+import type { BufferEncoding } from 'vfile';
```
bluwy, line 62: nit: I think we can remove this import as `BufferEncoding` is a global with `@types/node` loaded.

bluwy (approved):
Like the changes so far. Thanks for addressing my comments from the last PR!

Have some small nits below.

packages/astro/src/core/build/buildPipeline.ts
```diff
@@ -0,0 +1,211 @@
...
+				mode: staticBuildOptions.mode,
+				renderers: manifest.renderers,
+				clientDirectives: manifest.clientDirectives,
+				compressHTML: manifest.compressHTML,
+				async resolve(specifier: string) {
+					const hashedFilePath = manifest.entryModules[specifier];
+					if (typeof hashedFilePath !== 'string') {
+						// If no "astro:scripts/before-hydration.js" script exists in the build,
+						// then we can assume that no before-hydration scripts are needed.
+						if (specifier === BEFORE_HYDRATION_SCRIPT_ID) {
```
BlankParticle, line 47: I think this check is the reason that so many hybrid render builds are failing, 
in `packages/src/core/build/plugins/plugin-manifest.ts:238` you are setting `entryModules[BEFORE_HYDRATION_SCRIPT_ID] = ''`, then here you are checking `if (typeof hashedFilePath !== 'string')` and inside that check you are checking `if (specifier === BEFORE_HYDRATION_SCRIPT_ID)`, but as you have already set the `BEFORE_HYDRATION_SCRIPT_ID` in `plugin-manifest.ts` so this check is never called.

as a result, while resolving `BEFORE_HYDRATION_SCRIPT_ID` in `packages/astro/src/runtime/server/hydration.ts:157` we get `"/"` instead of `""`, for which this check passes and `before-hydration-url` is generated.
```js
if (beforeHydrationUrl.length) {
		island.props['before-hydration-url'] = beforeHydrationUrl;
	}
```

So the potential fixes could be to change the function to be
```js
async resolve(specifier: string) {
					const hashedFilePath = manifest.entryModules[specifier];
					// If no "astro:scripts/before-hydration.js" script exists in the build,
					// then we can assume that no before-hydration scripts are needed.
					if (specifier === BEFORE_HYDRATION_SCRIPT_ID) {
						return '';
					}
					if (typeof hashedFilePath !== 'string') {
						throw new Error(`Cannot find the built path for ${specifier}`);
					}
					return createAssetLink(hashedFilePath, manifest.base, manifest.assetsPrefix);
				},
```
or change the check at `packages/astro/src/runtime/server/hydration.ts:157` to be
```js
if (beforeHydrationUrl.length !== 0 && beforeHydrationUrl !== "/") {
		island.props['before-hydration-url'] = beforeHydrationUrl;
	}
```

I am down to make a PR to fix the issue (#8379 and many more), just waiting for confrimations from the maintainers

PR #8089, opened 2023-08-15
[create-astro] update shell logic

natemoo-re:
## Changes

- The timeout logic added in https://github.com/withastro/astro/pull/8077 caused some issues
- This PR refactors `shell` to rely on Node's built-in `timeout` support
- If a `ChildProcess` times out, it will emit the `close` event but the `exitCode` will be `null`.

## Testing

Tested manually because it's spawning a shell

## Docs

N/A, bug fix only

### DIFF
--- a/.changeset/angry-balloons-argue.md
+++ b/.changeset/angry-balloons-argue.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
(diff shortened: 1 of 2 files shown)


delucis (approved):
Don’t know enough to know judge the fix, but I’d say NWTWWHB — ship it!

PR #8092, opened 2023-08-15
Cleanup dotfiles in dist

### LINKED ISSUE #7933: `node_modules` in `dist` folder

2.10.0

None

pnpm

macOS

N/A (build related)

### Describe the Bug

This is related to the Starlight integration but appears to be an issue with Astro itself. This only showed up for Astro versions `2.5.6` and AFTER. `2.5.5` and before run as expected.

There appear to be 2 files that aren't cleaned up as part of running `pnpm astro build`:

- `dist/node_modules/.pnpm/@astrojs+starlight@0.6.1_astro@2.10.0/node_modules/@astrojs/starlight/404.astro.mjs`
- `dist/node_modules/.pnpm/@astrojs+starlight@0.6.1_astro@2.10.0/node_modules/@astrojs/starlight/index.astro.mjs`

The content inside the `index.astro.mjs` looks like this:

```js
export { renderers } from '../../../../../../renderers.mjs';
export { onRequest } from '../../../../../../_empty-middleware.mjs';
import '../../../../../../chunks/astro.2c0550c5.mjs';

const page = () => import('../../../../../../chunks/pages/index.astro.301b0b80.mjs');

export { page };
```

The contents of `404.astro.mjs` is the same but referencing the 404 files instead.

**Note**: I wasn't able to get this to reproduce in Stackblitz but got it locally on a macOS M1 Pro (macOS 13.4 (22F66)) and also on a GitHub Codespace environment.

Original issue in the Starlight repo: https://github.com/withastro/starlight/issues/458 
Potentially related: https://github.com/withastro/astro/issues/7397, https://github.com/withastro/astro/pull/7193

The `node_modules` folder shouldn't be included in the `dist` folder (which is resulting in some downstream tooling thinking a build output parameter isn't configured properly since there's a `node_modules` folder).

https://github.com/lorenzolewis/starlight-bug-test

### Participation

natemoo-re:
## Changes

- Closes #7933, plt-789
- Ensure dotfiles are cleaned up after static build
- This can happen when `node_modules/.pnpm` is present in the build output

## Testing

Manually tested the repro in https://github.com/withastro/astro/issues/7933, could no longer repro

## Docs

N/A, bug fix only


bluwy (approved):
LGTM. But I'm also confused how the `node_modules` ended in the `dist`.

natemoo-re:
> LGTM. But I'm also confused how the `node_modules` ended in the `dist`.

Yeah, that's a good point. I think the emitted file from the `injectRoute` kept the original source path? We can fix that in a follow-up for sure.

PR #8095, opened 2023-08-15
Add newline between HTML elements and <style> tags in example files

teddybradford:
## Changes

These changes provide more delineation between HTML and CSS in example files. Also matches existing delineation in `index.astro`.

- Add newline between </html> and <style> tags in example `components/Card.astro`
- Add newline between </html> and <style> tags in example `layouts/Layout.astro`

## Testing

No testing performed.

## Docs

No changes to docs needed.


ematipico:
Thank you, but we're not looking for these kinds of contributions.

PR #8099, opened 2023-08-16
Deprecate drafts feature

### LINKED ISSUE #6400: Draft frontmatter flag not respected in collection

2.0.2

None

pnpm

Mac

### Describe the Bug

`draft: true` in frontmatter of pages is respected, but in frontmatter of collections it isn't when generating static site.

As a workaround one can create a filter when calling `getCollection` and also `getStaticPath`

```
await getCollection("post", ({ data }) => import.meta.env.DEV || !data?.draft))
```

But that's not consistent with pages (where ItJustWorks) or with user expectations

https://stackblitz.com/edit/github-tlp8uq

### Participation

### LINKED ISSUE #7324: Drafts filtering behavior, configuration, flags, and docs needs cleanup.

2.6.1

None

pnpm

Mac

Firefox

### Describe the Bug

This *may* simply be a problem with my understanding of the behavior or a clarification in the docs.

## Background

I've looked at #440 and the docs regarding [the `build --drafts` flag](https://docs.astro.build/en/reference/cli-reference/#astro-build) and the [markdown.drafts](https://docs.astro.build/en/reference/configuration-reference/#markdowndrafts) configuration.

It looks to me like:
- the `--drafts` flag isn't wired in to code yet
- there's no strong convention for using the `markdown.drafts` flag. AFAICT it's ignored by the collections API, the stock blog template, and other 3rd party templates (my stackblitz example only demonstrates that the draft pages are built whether or not the flag is present).
- There's not clarity on how drafts filtering should happen (#7087, #5055)

Astro's collection API is a nice addition, and it prevents 1000 ad-hoc almost compatible similar implementations from sprouting up. The idea of having items in a collection that aren't ready for production (drafts) seems like a good thing to have a convention around.

## Proposal

This will take some discussion, but to start the discussion, I propose that:

- When an item is marked `draft` it is included in the by default in `astro dev` and excluded by default in `astro build`, which can be overridden by `config.markdown.drafts`, which can be overridden by the `--drafts` flag.

- When drafts are excluded, they're excluded *everywhere*. They don't exist--not their empty folders, their html, nor their RSS feeds. This prevents accidental info leaks and other embarrassments.

- This behavior should be implemented in the collections API.

## Next Steps

All those assertions are debatable, but I think form a good starting point. If we can agree on behavior and api, I'd be willing to submit a PR.

35055

https://stackblitz.com/github/bradobro/astro-test-drafts?file=README.md

### Participation

bluwy:
## Changes

Deprecate drafts feature in favour of content collections

- Closes #7324, closes #6400

## Testing

Tested manually that the warning is logged in dev and build.

## Docs

Added a changeset. The migration path is a bit bigger since it involves migrating to content collections first, then filtering `draft`. I linked to the related parts in docs instead.

I'm also not sure if we need more documentation in the added `@deprecated` jsdoc. If we have more info in the published migration guide later, those fields could point to it instead?


matthewp (approved):

.changeset/silent-bikes-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/rss': patch
+'astro': patch
+---
+
+Deprecate the drafts feature in favour of content collections. If you'd like to create draft pages that's visible in dev but not in prod, you can [migrate to content collections](https://docs.astro.build/en/guides/content-collections/#migrating-from-file-based-routing) and [manually filter out pages](https://docs.astro.build/en/guides/content-collections/#filtering-collection-queries) with the `draft: true` frontmatter property instead.
```
sarah11918, line 6:
```suggestion
Deprecate the `markdown.drafts` configuration option.

If you'd like to create draft pages that are visible in dev but not in production, you can [migrate to content collections](https://docs.astro.build/en/guides/content-collections/#migrating-from-file-based-routing) and [manually filter out pages](https://docs.astro.build/en/guides/content-collections/#filtering-collection-queries) with the `draft: true` frontmatter property instead.
```

ematipico (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -1017,6 +1017,7 @@ export interface AstroUserConfig {
 		 * @name markdown.drafts
 		 * @type {boolean}
 		 * @default `false`
+		 * @deprecated since version 3.0. Use content collections instead.
```
sarah11918, line 1020: @delucis - Is there any special formatting we need to adhere to for the `deprecated` entry here?

sarah11918 (commented):
Hey @bluwy!  Quick touch up to the changeset, and I've left a question for @delucis since I can't remember the last time we deprecated something on this page and I want to make sure we're doing it correctly.  :smile:

packages/astro/src/@types/astro.ts
delucis: We don’t actually display `@deprecated` comments in docs yet, so now’s our chance to decide a format if we want!

I can imagine generating something like this? Where the contents of the comments follow the leading `Deprecated:`.

image

Or even something like this depending on how highlighted we want it:

image

So given that, I guess it depends if you think specific wording would be clearest there @sarah11918?

bluwy: Actually, checking where I adopted the `since version X` pattern which is from `errors-data.ts`, it looks like we already render it today:

image

So perhaps another place to align if we want to make `@deprecated` work nicely everywhere.

delucis: Ah OK, yes, we could implement a similar pattern in the config reference if we like how the error generation script does this.

In this example I don’t love how basically a full sentence is split across heading and paragraph, so if we like the chunkier aside rendering, I would probably argue we should end up with something that looks more like:

> ⚠️ **Deprecated**
> Deprecated since version 2.6. Use content collections instead.

In which case, best for that to be reflected in the JSDoc comment too:

```suggestion
		 * @deprecated Deprecated since version 2.6. Use content collections instead.
```

bluwy: I pushed a commit to update the `@deprecated` tag to be a proper sentence instead 👍

ematipico (approved):

sarah11918:
Just noting that for v3.0, we will actually be using this here: https://github.com/withastro/astro/pull/8158

delucis (approved):
Thanks for updating the `@deprecated` tags @bluwy! LGTM!

PR #8101, opened 2023-08-16
`astro:` namespace for middleware and components

matthewp:
## Changes

- Adds aliases of `astro:middleware` and `astro:components`

## Testing

- All updated to use the aliases

## Docs

- [PR](https://github.com/withastro/docs/pull/4239)

### DIFF
--- a/.changeset/neat-mugs-end.md
+++ b/.changeset/neat-mugs-end.md
@@ -1,0 +1,8 @@
+---
+'astro': minor
+---
+
+
+`astro:`namespace aliases for middleware and components
+
+This adds aliases of `astro:middleware` and `astro:components` for the middleware and components modules. This is to make our documentation consistent between are various modules, where some are virtual modules and others are not. Going forward new built-in modules will use this namespace.
--- a/examples/middleware/src/middleware.ts
+++ b/examples/middleware/src/middleware.ts
@@ -1,4 +1,4 @@
-import { defineMiddleware, sequence } from 'astro/middleware';
+import { defineMiddleware, sequence } from 'astro:middleware';
 import htmlMinifier from 'html-minifier';
 
 const limit = 50;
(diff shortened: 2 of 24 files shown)


natemoo-re (approved):

.changeset/neat-mugs-end.md
```diff
@@ -0,0 +1,8 @@
+---
+'astro': minor
+---
+
+
+`astro:`namespace aliases for middleware and components
```
ematipico, line 6:
```suggestion
`astro:` namespace aliases for middleware and components
```

ematipico (commented):
About `astro:middleware` VS `astro/middleware`, here's some concerns we should take into account (even in the language tools cc @Princesseuh )

At the moment, `astro/middleware` exposes the following APIs: `createContext`, `trySerializeLocals`, `defineMiddleware` and `sequence`.

`defineMiddleware` and `sequence` can be part of `astro:middleware`, but `createContext`, `trySerializeLocals` can't be part of `astro:middleware`, because are **runtime** APIs that expose code that will be used by integrations for advanced features (e.g. edge middleware).

I wonder:
- is it possible to exclude `createContext`, `trySerializeLocals` from `astro:middleware`?
- is it possible to exclude `defineMiddleware` and `sequence` from `astro/middleware`?
- is it possible to suggest the correct import based on the context? Probably yes, because the middleware is a **known** file with fixed location and name.

bluwy:
> but `createContext`, `trySerializeLocals` can't be part of `astro:middleware`, because are **runtime** APIs that expose code that will be used by integrations for advanced features (e.g. edge middleware).

Couldn't `createContext` and `trySerializeLocals` still be part of `astro:middleware`, but integrations can import from `astro/middleware` instead? They'd be unused in the other environments, but I think it's probably fine so it's easier to maintain.

ematipico (approved):

PR #8104, opened 2023-08-16
Fix missing async/await in vite-plugin-integrations-container on astro:server:setup hook

### LINKED ISSUE #8097: create-vite code path does not wait for "astro:server:setup" hooks

2.10.7

node

npm

linux

chrome

### Describe the Bug

```diff
--- node_modules/astro/dist/vite-plugin-integrations-container/index.js 2023-08-15 19:10:11.965464887 -0500
+++ fixed.js 2023-08-15 19:10:04.233526608 -0500
@@ -6,10 +6,10 @@
 }) {
   return {
     name: "astro:integration-container",
-    configureServer(server) {
+    async configureServer(server) {
       if (server.config.isProduction)
         return;
-      runHookServerSetup({ config: settings.config, server, logging });
+      await runHookServerSetup({ config: settings.config, server, logging });
     },
     async buildStart() {
       if (settings.injectedRoutes.length === settings.resolvedInjectedRoutes.length)

```

Sorry for the inline patch, but I modified my node_modules file directly to see if this worked.

Basically, all integration hooks can return a Promise<void>, so I added a astro:server:setup hook, and added passport-js+oidc-client middleware.  The later requires an internal await(), to talk to the external server to auto-discover the endpoint urls.  I track down the bug to a missing async/await pair in the vite-plugin-integrations-container file.

I expect that if I have an async astro:server:setup hook, that `astro run dev` will wait for it to finish, before continuing on with the startup.

inline patch

### Participation

eigood:
## Changes

Closes #8097

Add async/await into configureServer so that integrations with an async astro:server:setup hook can function correctly.

## Testing

I edited the built file directly in my target app before creating this patch, and it worked for me in that scenario.  I don't yet know how to create a proper test case around this.  Please see discussion in #8097 

## Docs

No documentation change; the docs already have this hook defined to return a Promise<Void>, this change just fixes the code so it honors that.

### DIFF
--- a/packages/astro/src/vite-plugin-integrations-container/index.ts
+++ b/packages/astro/src/vite-plugin-integrations-container/index.ts
@@ -16,9 +16,9 @@ export default function astroIntegrationsContainerPlugin({
 }): VitePlugin {
 	return {
 		name: 'astro:integration-container',
-		configureServer(server) {
+		async configureServer(server) {
 			if (server.config.isProduction) return;
-			runHookServerSetup({ config: settings.config, server, logging });
+			await runHookServerSetup({ config: settings.config, server, logging });
 		},
 		async buildStart() {
 			if (settings.injectedRoutes.length === settings.resolvedInjectedRoutes.length) return;
--- a/packages/astro/test/fixtures/integration-server-setup/integration.js
+++ b/packages/astro/test/fixtures/integration-server-setup/integration.js
@@ -1,8 +1,12 @@
+import { setTimeout } from "node:timers/promises";
+
 export default function() {
 	return {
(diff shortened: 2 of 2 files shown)


natemoo-re:
Thanks, this looks great! If you could cherry-pick my test update onto your branch that would be awesome. I can't seem to push to your branch.

```
git cherry-pick 50454ed364bf360f2adce21a04cfc437346fe279
```

natemoo-re (approved):
Thank you for the PR!

natemoo-re:
Thanks again @eigood! The test failure is unrelated and should be fixed by https://github.com/withastro/astro/pull/8103

PR #8109, opened 2023-08-17
fix: reinsert attribute to specify direction of ViewTransition

martrapp:
## Changes

`astroTransition = dir` is now set for the current document.
The removal of the attribute from the DOM is put on hold for now, since the removal seems to prevent fallback animations.
Plus: fix for an of-by-one bug I introduced in a previous commit that prevented the correct detection of direction ;-)

## Testing

Manual tests only

## Docs

n/a bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -101,9 +101,8 @@ const { fallback = 'animate' } = Astro.props as Props;
 
 	const parser = new DOMParser();
 
-	async function updateDOM(dir: Direction, html: string, state?: State, fallback?: Fallback) {
+	async function updateDOM(html: string, state?: State, fallback?: Fallback) {
 		const doc = parser.parseFromString(html, 'text/html');
-		doc.documentElement.dataset.astroTransition = dir;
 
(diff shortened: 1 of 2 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -233,15 +232,17 @@ const { fallback = 'animate' } = Astro.props as Props;
 			location.href = href;
 			return;
 		}
+		document.documentElement.dataset.astroTransition = dir;
```
matthewp, line 235: why here and not inside of updateDOM?

martrapp: :) 
1. symmetry: removing the attribute is done a few lines further down in "Navigate". Same level.
1. When I removed `doc.documentElement.dataset.astroTransition = dir` from `updateDOM`, it was clear that updating the DOM does not depend on a direction. (this was the only usage of `dir` in this function)
1. simpler, less parameters for updateDOM

matthewp: Ok, trust your judgment here, thanks!

PR #8113, opened 2023-08-17
config: remove ~/assets alias

Princesseuh:
## Changes

What the title says, the alias didn't work properly in the editor because of how TypeScript works and it was a bit awkward to have now that `src/assets` isn't necessarily as special as it used to be. We'll revisit this in the future in a dedicated "should we have cool aliases?" RFC potentially. But for 3.0, no built-in aliases. Of course people can still define them themselves if they desire to.

## Testing

Kept the tests for aliases since it's good to test for people who configure them, but added the alias manually to those tests

## Docs

Already covered by Sarah's current work on the v3 page for assets

### DIFF
--- a/.changeset/clever-bats-breathe.md
+++ b/.changeset/clever-bats-breathe.md
@@ -1,5 +1,5 @@
 ---
-'astro': patch
+'astro': major
 ---
 
 This import alias is no longer included by default with astro:assets. If you were using this alias with experimental assets, you must convert them to relative file paths, or create your own [import aliases](https://docs.astro.build/en/guides/aliases/).
--- a/.changeset/clever-bats-breathe.md
+++ b/.changeset/clever-bats-breathe.md
@@ -4,10 +4,10 @@
 
 This import alias is no longer included by default with astro:assets. If you were using this alias with experimental assets, you must convert them to relative file paths, or create your own [import aliases](https://docs.astro.build/en/guides/aliases/).
 
-```astro
+```diff
 ---
 // src/pages/posts/post-1.astro
-import rocket from '~/assets/rocket.png'
-import rocket from '../../assets/rocket.png';
+- import rocket from '~/assets/rocket.png'
++ import rocket from '../../assets/rocket.png';
 ---
 ```
--- a/.changeset/clever-bats-breathe.md
+++ b/.changeset/clever-bats-breathe.md
@@ -1,0 +1,13 @@
+---
+'astro': patch
+---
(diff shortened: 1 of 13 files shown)


sarah11918:
Just a note that if there is going to be a changeset here, since there is a potentially user breaking change and you might want something to show up in the changelog, you can pull from the content here: https://deploy-preview-3739--astro-docs-2.netlify.app/en/guides/images/#remove-the-assets-import-alias

But yes, advice for this will be covered in docs!

sarah11918:
Awww... and it was a first-time contributor!  :smile:

Princesseuh:
> Awww... and it was a first-time contributor! 😄

We talk often on Discord about `astro:assets`, I'll credit them appropriately 😅

.changeset/clever-bats-breathe.md
```diff
@@ -0,0 +1,13 @@
...
+
+This import alias is no longer included by default with astro:assets. If you were using this alias with experimental assets, you must convert them to relative file paths, or create your own [import aliases](https://docs.astro.build/en/guides/aliases/).
+
+```astro
+---
+// src/pages/posts/post-1.astro
+import rocket from '~/assets/rocket.png'
+import rocket from '../../assets/rocket.png';
+---
+```
```
sarah11918, line 13:
``````suggestion
```diff
---
// src/pages/posts/post-1.astro
- import rocket from '~/assets/rocket.png'
+ import rocket from '../../assets/rocket.png';
---
```
``````
You'll need to do your diffing yourself here, since this isn't Astro docs. :cry:

**Review by matthewp — APPROVED** [review]

File: .changeset/clever-bats-breathe.md [hunk 2]
```diff
@@ -0,0 +1,13 @@
+---
+'astro': patch
```
ematipico, line 2: Should it be a `major`?

ematipico (approved):

.changeset/clever-bats-breathe.md
Princesseuh:
```suggestion
'astro': major
```

elliottmejia:
Hey, wrote a whole bunch of code with @components, @layout, etc. In a 2.x build.
**From what I understand here, this @ aliases are deprecated and I should instead be using ../ ?**
Or should I set the baseURL to the relative path of the TS package? What would this be where the dist is on a base of ~?
I would much prefer to use aliases.

PR #8115, opened 2023-08-17
refactor: use pipeline in Development mode

ematipico:
## Changes

This PR takes the pipeline concept into development mode. 

The `DevPipeline` is very similar to the `BuildPipeline`. Like in the previous PR, the `pipeline` object is passed around, and functions request from it all the information that is needed.

This refactor removes the need for `DevelopmentEnviroment`: `AstroSettings` and `ModuleLoader` are stored in the `DevPipeline`, and functions use its getters `getSettings` and `getModuleLoader`.

I also removed `tryRenderRoute`, which was deprecated.

## Testing

I updated some unit tests to use this new `DevPipeline`. Eventually, we should use a new `TestPipeline` for these kinds of tests, but not in this PR.

## Docs

N/A

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/devPipeline.ts
+++ b/packages/astro/src/vite-plugin-astro-server/devPipeline.ts
@@ -1,0 +1,128 @@
+import { Pipeline } from '../core/pipeline.js';
+import type { AstroConfig, AstroSettings, RouteData } from '../@types/astro';
+import type { ModuleLoader } from '../core/module-loader';
+import type { Environment } from '../core/render';
+import { createEnvironment, loadRenderer } from '../core/render/index.js';
+import { createResolve } from './resolve.js';
+import { RouteCache } from '../core/render/route-cache.js';
+import { isServerLikeOutput } from '../prerender/utils.js';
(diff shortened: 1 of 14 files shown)


packages/astro/src/vite-plugin-astro-server/devPipeline.ts
```diff
@@ -0,0 +1,128 @@
...
+	get logger(): Readonly<Logger> {
+		return this.#devLogger;
+	}
+
+	async loadRenderers() {
+		const renderers = await Promise.all(
+			this.#settings.renderers.map((r) => loadRenderer(r, this.#loader))
+		);
+		this.env.renderers = renderers.filter(Boolean) as SSRLoadedRenderer[];
+	}
```
ematipico, line 70: As you can notice, the side effects stay inside the pipeline, so we can better control them.

bluwy (approved):
Looks great! I notice this conflicts with the endpoint simple object deprecation + refactor I'm doing, so that `callEndpoint` can directly return a `Response`. Will hold that off for a bit until this is merged.

PR #8118, opened 2023-08-17
Switch `build.inlineStylesheets` default to auto

lilnasy:
## Changes
- Small stylesheets will now be inlined as <style>tags by default.

## Testing
- Explicitly use `inlineStylesheets: never` for tests that were written when it was the default
- [x] for packages/astro/test
- [x] for packages/integrations/*/test

## Docs
The `Bundle control` section will need to be updated.
PR: https://github.com/withastro/docs/pull/4253 ([live](https://docs.astro.build/en/guides/styling/#bundle-control), [updated](https://deploy-preview-4253--astro-docs-2.netlify.app/en/guides/styling/#bundle-control))

### DIFF
--- a/.changeset/smart-numbers-shout.md
+++ b/.changeset/smart-numbers-shout.md
@@ -1,0 +1,19 @@
+---
+'astro': patch
+---
+
+Astro is now smarter about CSS! Small stylesheets are now inlined by default, and no longer incur the cost of additional requests to your server. Your visitors will have to wait less before they see your pages, especially those in remote locations or in a subway.
+
+This may not be news to you if you had opted using the `build.inlineStylesheets` configuration. Stabilized in Astro 2.6 and set to "auto" by default for Starlight, this configuration allowed you to reduce the number of requests for stylesheets by inlining them into <style> tags. The new default is "auto", which selects assets smaller than 4kB and includes them in the initial response.
+
+To go back to the previous default behavior, change `build.inlineStylesheets` to "never".
+
+```ts
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+	build: {
+		inlineStylesheets: 'never',
+	},
+});
+```
--- a/.changeset/smart-numbers-shout.md
+++ b/.changeset/smart-numbers-shout.md
@@ -2,9 +2,9 @@
 'astro': patch
 ---
 
-Astro is now smarter about CSS! Small stylesheets are now inlined by default, and no longer incur the cost of additional requests to your server. Your visitors will have to wait less before they see your pages, especially those in remote locations or in a subway.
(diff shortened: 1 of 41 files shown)


lilnasy:
Ready for review

sarah11918:
Note from docs: there will also need to be a corresponding update to the [configuration-reference page](https://docs.astro.build/en/reference/configuration-reference/#buildinlinestylesheets) via [`astro.ts`](https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts#L839)!

lilnasy:
Forgot about it, thanks a lot for the reminder!

matthewp (approved):
thanks!

packages/astro/src/@types/astro.ts
```diff
@@ -847,7 +847,7 @@ export interface AstroUserConfig {
 		 * ```js
 		 * {
 		 * 	build: {
-		 *		inlineStylesheets: `auto`,
+		 *		inlineStylesheets: `never`,
```
sarah11918, line 850: Heads up that I also updated here, since we typically show a setting that would *need to be set*, and you wouldn't need to set `auto` specifically now that it's the default.

lilnasy: Saving me again!
🙏

sarah11918:
1. I'm assuming we should have a changeset since we're changing a default and this could be considered a breaking change for some projects? A notice of what to configure in order to get back to the same behaviour you had if you were relying on the default behaviour?

2. Asking about protocol here: This PR states there need to be updates to the docs repo. Should I be approving this PR BEFORE those changes are made, or do we think this PR should sit here (and think about what it did!) until there's a corresponding docs PR, so that we know it will happen?  :smile:

lilnasy:
I'm not sure about either of these  because I'm not familiar with the implications of merging into next.

Would the changeset become part of 3.0 release notes? Is there a `next` branch for docs I should open a PR for?

matthewp:
I would probably not approve this until the docs are updated. I can grab those tomorrow.

.changeset/smart-numbers-shout.md
```diff
@@ -0,0 +1,19 @@
+---
+'astro': patch
+---
+
+Astro is smarter about CSS! Small stylesheets are now inlined by default, and no longer incur the cost of additional requests to your server. Your visitors will have to wait less before they see your pages, especially those in remote locations or in a subway.
```
sarah11918, line 5: lol How can I not approve this?

sarah11918 (approved):
Docs is happy! Have also reviewed the corresponding docs PR, so we're all in good shape!  :raised_hands:

PR #8124, opened 2023-08-17
chore: resolve ShellCheck 2086 diagnostic in gitpod setup script

benelan:
## Changes

This resolves [ShellCheck 2086](https://www.shellcheck.net/wiki/SC2086) to prevent potential globbing and word splitting issues in the `gitpod-setup.sh` script.

## Testing

No tests were added because it just moved a quotation mark in a script's `echo` command.

## Docs

No documentation is required because it is one-character change to a tooling script.


natemoo-re (approved):
Thanks for the PR!

PR #8125, opened 2023-08-17
[error messages] Update image errors-data.ts for v3

sarah11918:
## Changes

This updates all the error messages that previously had links to the v2.x exerimental assets page. Links now point to the Images page (with appropriate anchor link if more targeted).

Also updates a message stating svg file format as unsupported (since it's not NOT supported now...)

## Testing

No tests, updates error messages

## Docs

All error messages!


sarah11918:
ugh, closing this because branches.

Issue #7923, opened 2023-08-02
" Unknown chunk type " when trying to display data from api in .astro component


lowzyyy:
### What version of `astro` are you using?

2,9,7

### Are you using an SSR adapter? If so, which one?

vercel

### What package manager are you using?

npm

### What operating system are you using?

windows

### What browser are you using?

firefox

### Describe the Bug

`Unknown chunk type:` name_of_some_attribute

When i fetch some data from API and try to display it, it shows error of unknown chunk.
This error has different error message on lower version of the astro like: `stringified is undefined`. But if you update to the latest version it should say the same as i mentioned in the beginning 

### What's the expected result?

It should display the data correctly

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-s5cpjh?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Pinging @bluwy for this regression

bluwy:
It doesn't seem to be a regression. The error happens in older versions, but the newer versions has a different error message.

bluwy:
The issue is that the data has the property `type` and internally we also used that to render things differently. A better fix is to have our renderer tag the internal objects to handle it differently, so userland data is rendered right.

But what would ultimately get rendered is an `[object Object]` and that isn't likely what you want. You can wrap it with a `JSON.stringify` before rendering instead.

natemoo-re:
Yeah seems like using `Symbol.for('astro:type')` would be a smart move.

jacobdejean:
I don't think Astro is doing anything wrong here- the issue really is that OP is trying to insert an object into the \<p\> element instead of a component or a string.

The MRE code can be fixed like this
```jsx
// from this
<p>{animeDetails.data.titles}</p>

// to this
{ animeDetails.data.titles.map(title => <p>{title.type /* or do logic with title.type */ + title.title}</p>)}
```

I can't imagine theres many cases where non-component objects need to be inserted into DOM elements in Astro. In React this case can be handled on a per-component basis, by reading the inserted object via `children` prop (And opting to transform that object into a renderable string internally). 

With Astro's \<slot \\> we don't really have the opportunity to perform transformations on inserted objects, so does it really make sense to accommodate these potential property name conflicts? 

While this error seems confusing at first, its stopping these arbitrary un-renderable objects from being rendered.

PR #8127, opened 2023-08-17
Stringify shouldn't throw on user object during rendering

natemoo-re:
## Changes

- Closes https://github.com/withastro/astro/issues/7923
- As https://github.com/withastro/astro/issues/7923 outlines, "Unknown chunk type" was a very generic error message that threw when a user passed an object like `{ type: 'something' }` in an expression.
- This PR adds a `createRenderInstruction` helper that includes a symbol for internal instructions so we can differentiate them from user objects

## Testing

Test added

## Docs

N/A, just a bug fix


bluwy (commented):
Looks great! Thanks for refactoring this.

bluwy (approved):

Issue #7966, opened 2023-08-04
[3.0.0-beta.0] Missing Sharp error during build


fflaten:
### What version of `astro` are you using?

3.0.0-beta.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Edge

### Describe the Bug

After upgrading a project using assets feature to `astro@3.0.0-beta.0`, `astro build` fails with:

```shell
 generating optimized images 
 error   Could not find Sharp. Please install Sharp manually into your project.
  File:
    /workspaces/Website/node_modules/.pnpm/astro@3.0.0-beta.0_@types+node@20.4.6_sass@1.64.2_typescript@5.0.4/node_modules/astro/dist/assets/generate.js:56:22
  Code:
    55 |   const imageService = await getConfiguredImageService();
    > 56 |   const resultData = await imageService.transform(
         |                      ^
      57 |     fileData,
      58 |     { ...options, src: originalImagePath },
      59 |     buildOpts.settings.config.image.service.config
  Stacktrace:
Error: Could not find Sharp. Please install Sharp manually into your project.
    at loadSharp (file:///workspaces/Website/dist/chunks/astro.7dbb7aca.mjs:3558:11)
    at async Object.transform (file:///workspaces/Website/dist/chunks/astro.7dbb7aca.mjs:3569:15)
    at async generateImage (file:///workspaces/Website/node_modules/.pnpm/astro@3.0.0-beta.0_@types+node@20.4.6_sass@1.64.2_typescript@5.0.4/node_modules/astro/dist/assets/generate.js:56:22)
    at async generateImage (file:///workspaces/Website/node_modules/.pnpm/astro@3.0.0-beta.0_@types+node@20.4.6_sass@1.64.2_typescript@5.0.4/node_modules/astro/dist/core/build/generate.js:154:26)
    at async generatePages (file:///workspaces/Website/node_modules/.pnpm/astro@3.0.0-beta.0_@types+node@20.4.6_sass@1.64.2_typescript@5.0.4/node_modules/astro/dist/core/build/generate.js:141:7)
    at async staticBuild (file:///workspaces/Website/node_modules/.pnpm/astro@3.0.0-beta.0_@types+node@20.4.6_sass@1.64.2_typescript@5.0.4/node_modules/astro/dist/core/build/static-build.js:85:7)
[...]

PR #8128, opened 2023-08-18
fix: better error when Sharp can't be resolved (ex: pnpm)

Princesseuh:
## Changes

Unfortunately, due to bundling, packaging, JavaScript, etc nonsense, on pnpm Sharp fails to be found. There's no proper way to fix this that wouldn't be: risky and/or unreliable. Trust me, I tried putting `sharp` in every single Vite config I found.

So, I did the next best thing: I upgraded the error message with every bit of info I could. I also added the passthrough / noop image service to the config so it's easier to use for SSR users who don't want to install Sharp into their projects. In the future, a more optimal solution for SSR would be a way to disable the endpoint, but that also requires either rethinking how we do images in Markdown, or at least a way to disable image processing in Markdown, it's a bit complex.

Fix https://github.com/withastro/astro/issues/7966

## Testing

Tested manually

## Docs

/cc @withastro/maintainers-docs for feedback on the error message and description

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -794,11 +794,13 @@ export const InvalidDynamicRoute = {
 /**
  * @docs
  * @see
- * - [Images](https://docs.astro.build/en/guides/images/)
+ * - [Default Image Service](https://docs.astro.build/en/guides/images/#default-image-service)
+ * - [Image Component](https://docs.astro.build/en/guides/images/#image--astroassets)
+ * - [Image Services API](https://docs.astro.build/en/reference/image-service-reference/)
  * @description
- * When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
+ * Sharp is the default image service used for `astro:assets`. When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
  *
- * If you are not using image processing, you can use a passthrough image service that does no processing, like such:
+ * If you are not using `astro:assets` for image processing, and do not wish to install Sharp, you can configure the following passthrough image service that does no processing:
  *
  * ```js
  * import { defineConfig, passthroughImageService } from "astro/config";
@@ -813,7 +815,7 @@ export const MissingSharp = {
 	name: 'MissingSharp',
 	title: 'Could not find Sharp.',
 	message: 'Could not find Sharp. Please install Sharp (`sharp`) manually into your project.',
-	hint: "See Sharp's installation instructions for more information: https://sharp.pixelplumbing.com/install. If you are not using image processing, you may want to use the passthrough image service instead, see https://docs.astro.build/en/reference/errors/missing-sharp for more information.",
+	hint: "See Sharp's installation instructions for more information: https://sharp.pixelplumbing.com/install. If you are not relying on `astro:assets` to optimize, transform, or process any images, you can configure a passthrough image service instead of installing Sharp. See https://docs.astro.build/en/reference/errors/missing-sharp for more information.",
 };
 // No headings here, that way Vite errors are merged with Astro ones in the docs, which makes more sense to users.
 // Vite Errors - 4xxx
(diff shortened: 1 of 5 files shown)


ematipico (approved):

ElianCodes (approved):
Looks great!

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -791,6 +791,30 @@ export const InvalidDynamicRoute = {
...
+ *    service: passthroughImageService(),
+ *  },
+ * });
+ * ```
+ */
+export const MissingSharp = {
+	name: 'MissingSharp',
+	title: 'Could not find Sharp.',
+	message: 'Could not find Sharp. Please install Sharp (`sharp`) manually into your project.',
+	hint: "See Sharp's installation instructions for more information: https://sharp.pixelplumbing.com/install. If you are not using image processing, you may want to use the passthrough image service instead, see https://docs.astro.build/en/reference/errors/missing-sharp for more information.",
```
sarah11918, line 816:
```suggestion
	hint: "See Sharp's installation instructions for more information: https://sharp.pixelplumbing.com/install. If you are not relying on `astro:assets` to optimize, transform, or process any images, you can configure a passthrough image service instead of installing Sharp. See https://docs.astro.build/en/reference/errors/missing-sharp for more information.",
```
I'm wondering whether this might be clearer to the reader than "image processing" which could make people wonder, wait, I have images? Can I not have any images in my project without Sharp?

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -791,6 +791,30 @@ export const InvalidDynamicRoute = {
...
 		`The ${invalidParam} param for route ${route} is invalid. Received **${received}**.`,
 } satisfies ErrorData;
+/**
+ * @docs
+ * @see
+ * - [Images](https://docs.astro.build/en/guides/images/)
+ * @description
+ * When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
+ *
+ * If you are not using image processing, you can use a passthrough image service that does no processing, like such:
```
sarah11918, line 801:
```suggestion
 * If you are not using `astro:assets` for image processing, and do not wish to install Sharp, you can configure the following passthrough image service that does no processing:
```
I think saying `astro:assets` here can be supported by the rest of docs where we can make the distinction of image processing vs not by saying whether `astro:assets` is involved.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -791,6 +791,30 @@ export const InvalidDynamicRoute = {
 	message: (route: string, invalidParam: string, received: string) =>
 		`The ${invalidParam} param for route ${route} is invalid. Received **${received}**.`,
 } satisfies ErrorData;
+/**
+ * @docs
+ * @see
+ * - [Images](https://docs.astro.build/en/guides/images/)
+ * @description
+ * When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
```
sarah11918, line 799:
```suggestion
 * Sharp is the default image service used for `astro:assets`. When using a [strict package manager](https://pnpm.io/pnpm-vs-npm#npms-flat-tree) like pnpm, Sharp must be installed manually into your project in order to use image processing.
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -791,6 +791,30 @@ export const InvalidDynamicRoute = {
 	message: (route: string, invalidParam: string, received: string) =>
 		`The ${invalidParam} param for route ${route} is invalid. Received **${received}**.`,
 } satisfies ErrorData;
+/**
+ * @docs
+ * @see
+ * - [Images](https://docs.astro.build/en/guides/images/)
```
sarah11918, line 797:
```suggestion
 * - [Default Image Service](https://docs.astro.build/en/guides/images/#default-image-service)
 * - [Image Component](https://docs.astro.build/en/guides/images/#image--astroassets)
 * - [Image Services API](https://docs.astro.build/en/reference/image-service-reference/)
```
These will all work in v3 docs

sarah11918 (commented):
Thanks for this great error, @Princesseuh ! I added some suggestions for details in case you wanted to make these even more super-explicit, as well as some links I know will be live in the new docs. (Some are now.)  Your call!

My only concern was that (as you know from my trying to write these new docs!) "image processing" can risk sounding vague if people don't really get that "processed" doesn't include "simply display on your site". So this message makes me wonder in general maybe whether referring to "using `astro:assets` or not?" is an easier way for readers to understand that "processing" is more than just "agreeing to display your image on your page."  😄

matthewp (approved):

PR #8132, opened 2023-08-18
Deprecate simple objects from endpoints

bluwy:
## Changes

- Deprecate returning simple objects from endpoints
- Refactor endpoint handling to within `callEndpoint`. This makes the `#endpointHandler` convention in Pipeline not used much (only `ssrPipeline` has a special case) cc @ematipico
- NOTE: much of the refactor was a port from https://github.com/withastro/astro/pull/7588

## Testing

Added test for `ResponseWithEncoding`, and tests for previous object form so they still work, until they're removed.

## Docs

Added changeset. Any mention of the object form in docs might need to be removed.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -23,6 +23,7 @@ import type { LogOptions, LoggerLevel } from '../core/logger/core';
 import type { AstroIntegrationLogger } from '../core/logger/core';
 import type { AstroComponentFactory, AstroComponentInstance } from '../runtime/server';
 import type { SUPPORTED_MARKDOWN_FILE_EXTENSIONS } from './../core/constants.js';
+import type { ResponseWithEncoding } from '../core/endpoint/index.js';
 
 export type {
 	MarkdownHeading,
@@ -1963,12 +1964,13 @@ export interface APIContext<Props extends Record<string, any> = Record<string, a
 	 * ```
 	 */
 	locals: App.Locals;
+	ResponseWithEncoding: typeof ResponseWithEncoding;
 }
 
 export type EndpointOutput =
 	| {
 			body: Body;
-			encoding?: Exclude<BufferEncoding, 'binary'>;
+			encoding?: BufferEncoding;
 	  }
 	| {
 			body: Uint8Array;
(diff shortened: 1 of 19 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1963,12 +1964,13 @@ export interface APIContext<Props extends Record<string, any> = Record<string, a
...
 }
 
 export type EndpointOutput =
 	| {
 			body: Body;
-			encoding?: Exclude<BufferEncoding, 'binary'>;
+			encoding?: BufferEncoding;
 	  }
 	| {
 			body: Uint8Array;
```
bluwy, line 1976: I think this was a mistake in https://github.com/withastro/astro/pull/8003, we technically only support binary strings, and not binary unit8arrays. We have a test for binary strings and the PR caused a type error.

But simple objects are being deprecated anyways so I don't think it's a big deal.

packages/astro/src/core/endpoint/index.ts
```diff
@@ -150,9 +168,50 @@ export async function callEndpoint<MiddlewareResult = Response | EndpointOutput>
 		}
 	}
 
-	return {
-		...response,
-		type: 'simple',
-		cookies: context.cookies,
-	};
+	let body: BodyInit;
```
bluwy, line 171: All the pipeline simple object handling converges into this function starting from this line. This should also make the dev/prod  behaviour more consistent.

ematipico (approved):
I love how much churn we removed, and we consolidated **one way** to process the responses from the endpoints!

matthewp (approved):

packages/astro/src/core/build/buildPipeline.ts
```diff
@@ -20,10 +18,6 @@ export class BuildPipeline extends Pipeline {
 	#internals: BuildInternals;
 	#staticBuildOptions: StaticBuildOptions;
 	#manifest: SSRManifest;
-	#currentEndpointBody?: {
-		body: string | Uint8Array;
-		encoding: BufferEncoding;
-	};
```
ematipico, line 26: One of the issues that forced me to keep this object is that we have a test that checks if we don't mangle the body or response, such as images. I

bluwy: Yeah I found that too (and CI is also erroring on that 😅), but I was able to make it not mangle for simple objects before. Currently it's erroring for `ResponseWithEncoding`. Thanks for the note!

bluwy: Aight I think I fixed it. I had to add the same encode logic inside `ResponseWithEncoding` so that it's on par with simple objects. I think that make sense for string responses, and otherwise the migration guide wouldn't be as nice, e.g. they have to do this themselves:

```js
new ResponseWithEncoding(Buffer.from('...', 'binary'), undefined, 'binary')
```

sarah11918 (approved):
Looks great, @bluwy ! 

As for the rest of docs, when you file an issue for this, please add a note there to remind us (inclusive! anyone can jump in here!) to do a "find all" instances of "Response" in the docs, then examine the surrounding context and update any text descriptions or code samples as needed!  :raised_hands:

PR #8136, opened 2023-08-18
Fix 404 response leading to an infinite loop when there is no 404 page

### LINKED ISSUE #8100: Returning a new Response with the Status Code 404 causes an Redirect to the (missing) /404.astro Page

2.10.9

Node

npm

Mac

Chrome

### Describe the Bug

In the past, with an older version of Astro, it was feasible to generate Responses bearing the 404 status code. However, in the current setup (2.10.9), this process triggers an infinite loop due to Astro's subsequent search for the missing /404.astro page. To circumvent this loop, I opt not to furnish a /404.astro page. This choice stems from my desire to refrain from redirecting users to a dedicated 404 page. Instead, I intend to furnish a 404 Code directly while retaining the requested URL's context.

```
return new Response("custom 404 html content", {
   status: 404
});
```

I would expect to be able to return any page with a 404 code.

https://stackblitz.com/edit/stackblitz-starters-6pmtt6?file=response-with-404-code%2Fsrc%2Fpages%2F[...path].astro

### Participation

andremralves:
## Changes

Fixes #8100 

Checks if 404 route exists before calling `matchRoute`.

`matchRoute` won't return undefined if the 404 route doesn't exist because we are in ssr. So we keep going through the recursion.

## Testing

Tested manually.

## Docs

n/a


natemoo-re (approved):
Thanks for the PR!

PR #8139, opened 2023-08-18
fix: polyfill File using undici instead of node:buffer

Princesseuh:
## Changes

What the title says

## Testing

We'll see!

## Docs

N/A


matthewp:
Changeset please

matthewp (approved):

PR #8141, opened 2023-08-18
fix(node): delegate preview's not found and error handling to core/app

### LINKED ISSUE #8054: 404 page doesn't render in production mode

2.10.7

Node

npm

Mac

Safari

### Describe the Bug

In SSR mode, when going to a page that doesn't exist, the 404 page should be rendered (the default or custom one). It does this correctly in dev mode, but **in production mode it only renders a blank page**.

Using @astrojs/node `5.3.3` with node `v18.16.1`

Also see the repro.
Run `npm run build && npm run preview` and go to `/doesntexist`.

The 404 page should be rendered correctly in SSR production mode.

https://stackblitz.com/edit/github-trfvka?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes
- Closes #8054
- Cause: the preview entrypoint passed a callback to nodeMiddleware, which overrode the recently refactored 404/500 handling.
- Fix: removed the callback to let src/core/app handle 404 and 500.

## Testing
Tested manually

## Docs
Does not affect usage

### DIFF
--- a/.changeset/rude-ducks-exist.md
+++ b/.changeset/rude-ducks-exist.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/node': patch
+---
+
+Fixed an issue where the preview mode handled 404 and 500 routes differently from running app with node directly.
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
Ah nice! Great to remove that special handling!

PR #8148, opened 2023-08-18
[docs] update scopedStyleStragegy default and description

sarah11918:
## Changes

Updates the default value in the config file to match the description. Also tightens up the `attribute` description.

## Testing

No tests, only docs

## Docs

Just docs!

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -608,7 +608,7 @@ export interface AstroUserConfig {
 	 * @docs
 	 * @name scopedStyleStrategy
 	 * @type {('where' | 'class' | 'attribute')}
-	 * @default `'where'`
+	 * @default `'attribute'`
 	 * @version 2.4
 	 * @description
 	 *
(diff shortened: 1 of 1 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -619,7 +619,7 @@ export interface AstroUserConfig {
 	 *
 	 * Using `'class'` is helpful when you want to ensure that element selectors within an Astro component override global style defaults (e.g. from a global stylesheet).
 	 * Using `'where'` gives you more control over specifity, but requires that you use higher-specifity selectors, layers, and other tools to control which selectors are applied.
-	 * Using `'attribute'` is useful in case there's manipulation of the class attributes, so the styling emitted by Astro doesn't go in conflict with the user's business logic.
+	 * Using 'attribute' is useful when you have scripts manipulating the `class` attribute of elements, and want to avoid conflicts between your own styling logic and Astro's styling.
```
sarah11918, line 622:
```suggestion
	 * Using 'attribute' is useful when you have scripts manipulating the `class` attribute of elements to avoid conflicts between your own styling logic and Astro's styling.
```

packages/astro/src/@types/astro.ts
```diff
@@ -619,7 +619,7 @@ export interface AstroUserConfig {
 	 *
 	 * Using `'class'` is helpful when you want to ensure that element selectors within an Astro component override global style defaults (e.g. from a global stylesheet).
 	 * Using `'where'` gives you more control over specifity, but requires that you use higher-specifity selectors, layers, and other tools to control which selectors are applied.
-	 * Using `'attribute'` is useful in case there's manipulation of the class attributes, so the styling emitted by Astro doesn't go in conflict with the user's business logic.
+	 * Using 'attribute' is useful when you have scripts manipulating the `class` attribute of elements to avoid conflicts between your own styling logic and Astro's application of styles.
```
matthewp, line 622: It's not really scripts. The only scenario where I think the class version causes a problem would be if a component was passed a `class` prop and expected it to equal some exact string. Not really related to client-side scripts. A client side script using `querySelector` is going to work either way.

sarah11918, line 622:
```suggestion
	 * Using 'attribute' is useful when you are manipulating the `class` attribute of elements and need to avoid conflicts between your own styling logic and Astro's application of styles.
```
OK, this might be a bit more neutral?

natemoo-re: I like that!

PR #7944, opened 2023-08-03
Support Bun in create-astro

colinhacks:
## Changes

Better support for Bun users when using `create-astro`.

- Prints the correct command for starting dev server (`bun run dev`)
- Add `bunx create-astro` to readme 

## Testing

I didn't see tests for `pnpm` or `yarn`. Bun sets `npm_config_user_agent` like Yarn/pnpm, and `which-pm-runs` detects this as expected. Here's a sample run showing everything working with Bun:

```sh
npm_config_user_agent=bun/1.0 node create-astro.mjs 

╭─────╮  Houston:
│ ◠ ◡ ◠  Let's build something great!
╰─────╯

 astro   v2.10.1 Launch sequence initiated.

   dir   Where should we create your new project?
         ./growing-gamma

  tmpl   How would you like to start your new project?
         Include sample files
      ✔  Template copied

  deps   Install dependencies?
         Yes
      ✔  Dependencies installed

    ts   Do you plan to write TypeScript?
         Yes

   use   How strict should TypeScript be?
         Strict
      ✔  TypeScript customized

   git   Initialize a new git repository?
         No
      ◼  Sounds good! You can always run git init manually.

  next   Liftoff confirmed. Explore your project!

         Enter your project directory using cd ./growing-gamma 
         Run bun run dev to start the dev server. CTRL+C to stop.
         Add frameworks like react or tailwind using astro add.

         Stuck? Join us at https://astro.build/chat

╭─────╮  Houston:
│ ◠ ◡ ◠  Good luck out there, astronaut! 🚀
╰─────╯
``` 

## Docs

Sister documentation PR: https://github.com/withastro/docs/pull/4052

PR #8150, opened 2023-08-18
Better support in detecting Bun

jacobthesheep:
## Changes

- Prints the correct command for adding an integration when using bun (`bunx astro add`)
- Bun is now detected as a package manager in all cases
- improves upon #7944
- added consistent wording

## Testing

Tested manually. Here's a screenshot.
works

## Docs

Sister documentation PR: https://github.com/withastro/docs/pull/4052
https://docs.astro.build/en/recipes/bun/

### DIFF
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -6,7 +6,6 @@ import fsMod, { existsSync, promises as fs } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import ora from 'ora';
-import whichPM from 'which-pm';
 import prompts from 'prompts';
 import type yargs from 'yargs-parser';
 import { loadTSConfig, resolveConfigPath, resolveRoot } from '../../core/config/index.js';
@@ -27,6 +26,7 @@ import { createLoggingFromFlags } from '../flags.js';
 import { generate, parse, t, visit } from './babel.js';
 import { ensureImport } from './imports.js';
 import { wrapDefaultExport } from './wrapper.js';
+import { detectAgent } from '@skarab/detect-package-manager';
 
 interface AddOptions {
 	flags: yargs.Arguments;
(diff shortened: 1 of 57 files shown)


packages/astro/src/cli/add/index.ts
```diff
@@ -629,6 +644,8 @@ async function getInstallIntegrationsCommand({
 			return { pm: 'yarn', command: 'add', flags: [], dependencies };
 		case 'pnpm':
 			return { pm: 'pnpm', command: 'add', flags: [], dependencies };
+		case 'bun':
+			return { pm: 'bun', command: 'add', flags: [], dependencies };
```
natemoo-re, line 648: This is great!

packages/astro/src/cli/add/index.ts
```diff
@@ -72,12 +71,28 @@ const OFFICIAL_ADAPTER_TO_IMPORT_MAP: Record<string, string> = {
...
+		return { name: 'yarn' };
+	}
+	if (existsSync(path.join(cwd, 'pnpm-lock.yaml'))) {
+		return { name: 'pnpm' };
+	}
+	if (existsSync(path.join(cwd, 'bun.lock'))) {
+		return { name: 'bun' };
+	}
+	return null;
+}
```
natemoo-re, line 88: Slightly hesitant about swapping out `preferred-pm` for this custom logic. It's essentially the same but I think `preferred-pm` has some fallbacks that this doesn't cover?

natemoo-re: I guess if Bun isn't supported by `preferred-pm`, we do kinda need to use our own!

jacobthesheep: I could implement additional checks, and error handling, and utilize which-pm as a fallback.

ematipico:
@jacobthesheep Thank you for the contribution. Have you tested the changes? It seems you haven't. Care to share some screenshots?

jacobthesheep:
I created a separate project and used my local Astro through [pnpm link](https://pnpm.io/cli/link) and it seems to work.
works

packages/astro/src/cli/add/index.ts
colinhacks: Bun I've submitted a PR to `preferred-pm` but I'm not sure when it'll land

PR #8153, opened 2023-08-18
 property in anchor element forces a reload of the target

### LINKED ISSUE #7947: PDF attachment not rendering right when ViewTransition is enabled.

2.10.1

None

pnpm

Mac

Chrome

### Describe the Bug

Say I have a pdf file containing Chinese texts under `public` folder, and I have a link to that pdf file in one of my pages:

```
<a href="/chinese.pdf">A PDF containing Chinese.</a>
```

With ViewTransition enabled, clicking that link would return something like this:
CleanShot-a46EcksI@2x

I would expect the browser to show the pdf content.

https://github.com/chenxsan/astro-view-transition-pdf

### Participation

martrapp:
## Changes

In an anchor element, the `data-astro-reload` property (with no value or with a value other than `'false'`) forces a reload of the target document (instead of a view transition).
  
Closes #7947 

## Testing

manually tested on the example given in #7947, outcome documented there.
[update 2023-08-22]: added e2e test 

## Docs

The documentation  of the (still experimental) View Transitions needs to be updated.

Not sure if the data property will be the final interface for users or if there will be syntactic sugar (e.g. `transition:reload`).
I plan to add one or more properties in the next few days to address other issues, in particular a property to control cross-page scrolling behavior (as opposed to on-page scrolling behavior)
  
### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -136,6 +136,18 @@ const { fallback = 'animate' } = Astro.props as Props;
 			// Remove them before swapping.
 			doc.querySelectorAll('head noscript').forEach((el) => el.remove());
 
+			// swap attributes of the html element
+			// - delete all attributes from the current document
+			// - insert all attributes from doc
+			// - reinsert all original attributes that are named 'data-astro-*'
+			const html = document.documentElement;
+			const astro = [...html.attributes].filter(
+				({ name }) => (html.removeAttribute(name), name.startsWith('data-astro-'))
+			);
+			[...doc.documentElement.attributes, ...astro].forEach(({ name, value }) =>
+				html.setAttribute(name, value)
+			);
+
 			// Swap head
 			for (const el of Array.from(document.head.children)) {
 				const newEl = persistedHeadElement(el);
--- a/.changeset/lovely-ducks-move.md
+++ b/.changeset/lovely-ducks-move.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix: View Transition: swap attributes of document's root element
--- a/packages/astro/e2e/fixtures/view-transitions/src/components/AttributedLayout.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/components/AttributedLayout.astro
@@ -1,0 +1,17 @@
+---
+import { ViewTransitions } from 'astro:transitions';
+import { HTMLAttributes } from 'astro/types';
+
+interface Props extends HTMLAttributes<'html'> {}
+---
+<html {...Astro.props}>
+	<head>
+		<title>Testing</title>
+		<ViewTransitions />
+	</head>
+	<body>
+		<main transition:animate="slide">
+			<slot />
+		</main>
+	</body>
+</html>
--- a/packages/astro/e2e/fixtures/view-transitions/src/pages/other-attributes.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/pages/other-attributes.astro
@@ -1,0 +1,11 @@
+---
+import AttributeLayout from '../components/AttributedLayout.astro';
+---
+<AttributeLayout
+lang="es"
+style="background-color: green"
+data-other-name="value"
+data-astro-fake="value"
+data-astro-transition="downward">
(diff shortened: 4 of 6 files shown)


matthewp:
Want to bikeshed on the name a little bit. `astro-reload` sounds a little off to me because we're not *reloading* the page; the page has never been loaded in the first place. But if this is a common name other similar implementations use then I think it's ok.

Just did a check:

- swup - `data-no-swup`: https://swup.js.org/api/markup/#ignore-links
- Turbo - `data-turbo="false"`: https://turbo.hotwired.dev/reference/attributes#data-attributes
- Phoenix - Opt-in via link component: https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#link/1

Could not find other examples.

martrapp:
force-load?
mpa-load?
…

ematipico:
Maybe `astro-load` or `astro-server-load`? Or `astro-force-load` like suggested. "mpa" uses an acronym that is not well known.

matthewp:
cc @delucis and @bluwy are naming experts.

delucis:
Given our current directives related to this are all `transition:*`, I’d imagine something like `transition:none` would make most sense here?

Although… would clicking a link with this property override any `transition:persist` and blow away that DOM? If so, that makes the `transition:*` namespace a much less good fit.

Do we have a description of use cases somewhere? Those might help with naming.

matthewp:
@delucis This just means you want to opt-out of client-side routing for a particular link. One use-case is that the link goes to a file rather than another page, for ex. `<a href="/chart.pdf">Chart</a>`.

Note that directives are used for server-side processing in some way; `transition:animate` creates a stylesheet. There's no server-side stuff happening in this one. The attribute will remain in the output HTML. Doesn't mean we can't use a directive like name, but something to consider.

bluwy:
SvelteKit uses `data-sveltekit-reload` so there's an initial pattern. Though I'd say it's confusing at times too, it's not really a reload, but letting the browser _do its thing_ instead. Maybe `data-astro-no-load`, `data-astro-skip-route`, `data-astro-skip-navigate`, `data-astro-skip`, `data-astro-ignore`? I'm not really sure what's the best name here 😄

delucis:
Yeah, if we’re reasonably happy with how [`astrojs/prefetch`](https://docs.astro.build/en/guides/integrations-guide/prefetch/) works currently, that could maybe help us pick a name that works well with those? Although that’s opt-in at the moment via `rel="prefetch"` and `rel="prefetch-intent"`, so perhaps doesn’t help.

matthewp:
What about `data-astro-no-transition`? Keep it specific to what happens otherwise?

martrapp:
There are only two hard things in Computer Science: cache invalidation and naming things.

My 2ct

I won’t use “transition” because technically it is not mainly about visual transition but whether or not the document is updated with new content (swap) or replaced by a new document. There could also be swap without transition.

link would sound good to me but none of the defined values (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) seem to fit our purpose

another try:
data-astro-fresh-document
data-astro-browser-default-navigation
data-astro-browser-default
data-astro-browser-navigation
sata-astro-default-navigation

These would also express that we leave it to to browser to choose viewers based on content-type

packages/astro/components/ViewTransitions.astro
```diff
@@ -276,6 +276,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 			if (
 				link &&
 				link instanceof HTMLAnchorElement &&
+				(link.dataset.astroReload ?? 'false') === 'false' &&
```
matthewp, line 279: is there a reason why someone would use a false value here? If not I think we can just check for it's existence to save a few bytes.

martrapp: For me it looks odd if someone writes: `reload="false"` and gets a reload. 
But at the end it is a matter of documentation: "Any value assigned to this attribute is ignored."
10 Bytes saved :-) !

matthewp:
I'm changing the base of this to `next` as we are not going to do another minor release of 2.x, so this will go out with 3.0.

I'm going to document this as part of https://github.com/withastro/docs/pull/4320

PR #8156, opened 2023-08-19
feat(ViewTransitions): use `scrollend` instead of `scroll` where supported

kurtextrem:
## Changes

The [scrollend](https://developer.chrome.com/blog/scrollend-a-new-javascript-event/#event-details) mechanism seems like a better way to record the scroll position compared to throttling, so we could use it whenever a browser supports it.

Additionally, I've removed the `{passive}` flag from the `scroll` event, as it does nothing ([source](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener?retiredLocale=de#improving_scrolling_performance_with_passive_listeners:~:text=You%20don%27t%20need%20to%20worry%20about%20the%20value%20of%20passive%20for%20the%20basic%20scroll%20event.%20Since%20it%20can%27t%20be%20canceled%2C%20event%20listeners%20can%27t%20block%20page%20rendering%20anyway.)).

- [x] TODO: run `pnpm exec changeset`

## Testing

I've tested it locally, but maybe I'm missing out how you are testing changes to this component.

## Docs

Open question: Does it make sense to document we're using `scrollend` in modern browsers? Personally, I don't think so (implementation details), but maybe it's interesting for some techy folks.

### DIFF
--- a/.changeset/long-chefs-jump.md
+++ b/.changeset/long-chefs-jump.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+The scrollend mechanism is a better way to record the scroll position compared to throttling, so we now use it whenever a browser supports it.
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -341,9 +341,9 @@ const { fallback = 'animate' } = Astro.props as Props;
 		});
 		addEventListener('load', onload);
 		// There's not a good way to record scroll position before a back button.
-		// So the way we do it is by listening to scroll and just continuously recording it.
+		// So the way we do it is by listening to scrollend if supported, and if not continuously record the scroll position.
     const updateState = () => {
-			// only updste history entries that are managed by us
+			// only update history entries that are managed by us
 			// leave other entries alone and do not accidently add state.
 			if (history.state) {
 				persistState({ ...history.state, scrollY });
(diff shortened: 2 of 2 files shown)


martrapp (commented):
Definitely an improvement! 😃

matthewp:
sgtm. Need a changeset and to not be in draft status. `pnpm changeset`

kurtextrem:
@martrapp @matthewp Thank you! I've added a changeset.

matthewp:
Ok, want to talk through this just a little more. Does `scrollend` wait until the user releases their fingers from the trackpad / screen? If so, is that what we want? When reading an article your fingers may remain the entire time as you scroll slowly. Does `scrollend` allow us to keep the scroll position as people do this slow scroll?

kurtextrem:
>  When reading an article your fingers may remain the entire time as you scroll slowly

The `scrollend` event only fires after you lift your finger (check out the demo at the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event#using_scrollend_with_an_event_listener) page), but at least I can not think of a manual way to trigger a link after scrolling without lifting your finger. A back gesture (on Android) can also not be triggered without lifting the finger at least once again.

However, if you scroll around, keep your finger at some position and e.g. some random non-Astro JS triggers a navigation after a specific timeout, `scrollend` will never fire. In that case the scroll position is lost, but I'm not sure if the implementation is trying to cover that (at least not from the comment in the code, as far as I can tell).

martrapp:
you could try to keep scrolling the mouse wheel while you press one handed Alt-<left-arrow>. That seems to be a corner case we can simply ignore.

martrapp:
… I assume, we could even define the router without listening to scroll events

matthewp (approved):

martrapp (commented):
Would be happy to approve, but ...
image
😃

matthewp:
Oh! Sorry about that. I'll take this comment as an approval. I guess access is only for maintainers (which you are on track to become imo).

PR #8159, opened 2023-08-19
feat(blog): add Disqus integration by default

suzaku:
## Changes

Add Disqus integration to blog template.


suzaku:
Makes sense. Maybe I should create a new theme based on the template.

PR #8160, opened 2023-08-20
feat(blog): add support for tags

suzaku:
## Changes

Add support for blog tags.

Screenshot 2023-08-20 at 9 56 47 AM


matthewp:
Hello! Thanks for submitting this. If this is something you think we should do, please discuss it on the discord first. For now I haven't heard interest amongst the team to have tags in the blog, so to be conservative I think we shouldn't accept the change right now.

PR #8166, opened 2023-08-20
Fixes in the client-side router

martrapp:
## Changes

There had been some corner cases for pages that did not include `<ViewTransitions>` in there `<head>`   

## Testing

added two e2e tests

## Docs

n/a bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -25,13 +25,14 @@ const { fallback = 'animate' } = Astro.props as Props;
 	// you can figure it using an index. On pushState the index is incremented so you
 	// can use that to determine popstate if going forward or back.
 	let currentHistoryIndex = history.state?.index || 0;
-	if (!history.state) {
+	if (!history.state && transitionEnabledOnThisPage()) {
 		persistState({ index: currentHistoryIndex, scrollY: 0 });
 	}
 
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
This seems totally reasonable, thanks for the fix!

packages/astro/components/ViewTransitions.astro
```diff
@@ -25,13 +25,14 @@ const { fallback = 'animate' } = Astro.props as Props;
...
 	let currentHistoryIndex = history.state?.index || 0;
-	if (!history.state) {
+	if (!history.state && transitionEnabledOnThisPage()) {
 		persistState({ index: currentHistoryIndex, scrollY: 0 });
 	}
 
 	const supportsViewTransitions = !!document.startViewTransition;
-	const transitionEnabledOnThisPage = () =>
-		!!document.querySelector('[name="astro-view-transitions-enabled"]');
+	function transitionEnabledOnThisPage() {
```
matthewp, line 33: went with arrow functions to save bytes, can we change this back?

martrapp: Sure ;-)
I had the choice to move the code up, or make it a function. Will revert.

martrapp: Done. This time, even the checks passed :)

Issue #7870, opened 2023-07-30
SSR node middleware not working on Express


marsidev:
### What version of `astro` are you using?

2.9.6

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

When enabling SSR (`output: "server"` or `output: "hybrid"`), with a middleware node adapter, the middleware is not working on Express (v4.18.2) when using `app.use(ssrHandler)`, as shown in [the docs](https://docs.astro.build/en/guides/integrations-guide/node/#middleware):

![image](https://github.com/withastro/astro/assets/56328053/ee9681a2-111f-4bd6-a369-39881030a533)

Instead, it works when using the middleware as follows:
```js
app.use((req, res, next) => {
  ssrHandler(req, res, next);
});
```

I noticed it works as expected on Fastify (v4.21.0) with just `app.use(ssrHandler)`.

This is my config:

```js
export default defineConfig({
  output: "server",
  adapter: node({
    mode: "middleware",
  }),
  server: {
    port: 8000,
  },
});
```

This is my express server:

```js
import express from "express";
import { handler as ssrHandler } from "./dist/server/entry.mjs";

const PORT = process.env.PORT || 8000;
const app = express();

app.use(express.static("dist/client/"));
app.use(ssrHandler); // <-- This is not working

// The following works:
// app.use((req, res, next) => {
//   ssrHandler(req, res, next);
// });

app.listen(PORT, () => {
  console.log(`Server running on http://localhost:${PORT}`);
});
```

I created [a repo](https://github.com/marsidev/astro-ssr-example) to reproduce the error:

1. Fork/clone the repo
2. Install deps (`pnpm i`)
3. Build the project (`pnpm build`)
4. Run express server (`pnpm preview`)
5. Go to http://localhost:8000/ssr or http://localhost:8000/api/ssr to check if works

### What's the expected result?

It should work with `app.use(ssrHandler)`. 

Please fix it or at least update the docs.

### Link to Minimal Reproducible Example

https://github.com/marsidev/astro-ssr-example

### Participation

- [ ] I am willing to submit a pull request for this issue.

andremralves:
It stopped working in version `v5.3.0` when the `locals` parameter was added to the middleware function. https://github.com/withastro/astro/pull/7385

Apparently `app.use()` can't handle functions with an extra parameter directly.

[...]

PR #8176, opened 2023-08-21
fix(@astrojs/node): handler should work with `express`

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/7870

## Testing

I added a new test case that uses `express` directly.

## Docs

N/A


matthewp (approved):
nice test

natemoo-re (approved):
Great test, thanks for fixing!

PR #8181, opened 2023-08-21
Finalize View Transition event names

matthewp:
> Do not merge! This is a draft as we biekshed these names. Will undraft once consensus is reached.

## Changes

- From the RFC: https://github.com/withastro/roadmap/pull/607
- Renames the `astro:load` to `astro:page-load`
   - `astro:load` is already taken.
- Renames `astro:beforeload` to `astro:after-swap`.
   - The event fires immediate after the DOM contents are swapped, so you can update the DOM to prevent FOUC
     - Most common use-case is dark mode.

## Testing

- Test updated

## Docs

- TBD, after the final naming decision is made.

## Points of contention

Based on other discussions from the RFC, I think these are the major points of contention to decide:

1. The `afterswap` name doesn't mention the transition. Some think it should.
2. The names are lowercased without dashes or camelcasing. The reasoning is that this follows the naming convention of built-in events.

### DIFF
--- a/.changeset/proud-fans-type.md
+++ b/.changeset/proud-fans-type.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Finalize View Transition event names
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -17,7 +17,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 		index: number;
 		scrollY: number;
 	};
-	type Events = 'astro:pageload' | 'astro:afterswap';
+	type Events = 'astro:page-load' | 'astro:after-swap';
 
 	const persistState = (state: State) => history.replaceState(state, '');
 
@@ -33,7 +33,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 	const transitionEnabledOnThisPage = () =>
 		!!document.querySelector('[name="astro-view-transitions-enabled"]');
 	const triggerEvent = (name: Events) => document.dispatchEvent(new Event(name));
-	const onPageLoad = () => triggerEvent('astro:pageload');
+	const onPageLoad = () => triggerEvent('astro:page-load');
 	const PERSIST_ATTR = 'data-astro-transition-persist';
 
 	const throttle = (cb: (...args: any[]) => any, delay: number) => {
@@ -175,7 +175,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 				persistState(state);
 			}
 
-			triggerEvent('astro:afterswap');
+			triggerEvent('astro:after-swap');
 		};
 
 		// Wait on links to finish, to prevent FOUC
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -17,7 +17,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 		index: number;
 		scrollY: number;
 	};
-	type Events = 'astro:load' | 'astro:beforeload';
+	type Events = 'astro:pageload' | 'astro:afterswap';
 
 	const persistState = (state: State) => history.replaceState(state, '');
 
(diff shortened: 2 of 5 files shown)


matthewp:
My major concern with using `transition` as part of the name for what is `afterswap` in this PR is that a transition is not just 1 step. From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API#the_view_transition_process) there are 6 steps. Additionally within our own code there are more steps. We might want to add events in the future for some of these steps and I want to avoid vague names that might clash with future event names.

matthewp:
For reference, here are other similar libraries and their event names:

- [htmx](https://htmx.org/docs/#events) uses `htmx:beforeSwap` and `htmx:afterSwap`.
- [swup](https://swup.js.org/hooks/#list) uses `content:replace`.
- [Turbo](https://turbo.hotwired.dev/reference/events) uses `turbo:render` (I think that's the one).

lilnasy:
> [Turbo](https://turbo.hotwired.dev/reference/events) uses `turbo:render`

I think that could be misconstrued because rendering (to HTML) is still happening on the server.

delucis:
> 1. The `afterswap` name doesn't mention the transition. Some think it should.

I think I am coming around to your argument. I don’t love the naming, but if we think we’ll need more fine-grained events and other names block that, we can aim to document the technical underpinnings so that `afterswap` is understood. Seeing the other tools using similar names is also reassuring.

> 2. The names are lowercased without dashes or camelcasing. The reasoning is that this follows the naming convention of built-in events.

We already break this convention by including a `:` in the event name, so I don’t think we have to adhere to it. (I’m also guessing this convention is mainly due to attribute form of `<el onclick="">` etc. which won’t apply to these events either.)

Reasons the convention is bad:
- less readable than a style where words are split up by casing differences (`afterSwap`) or a separator character (`after-swap`, `after_swap`)
- less accessible, because screen readers can’t detect the word boundary and are liable to mangle the pronunciation (quick VoiceOver test: `afterswap` — “afters waap”; `afterSwap` — “after swap”; `after-swap` — “after swap”; `after_swap` — “after underscore swap”)

matthewp:
Here's some docs. Want to see if this helps with the naming: https://github.com/withastro/docs/pull/4320

Going to update this PR to use dashes.

Akxe:
I would like to list all possible events we (you) may want to fire for a single transition:

| Order | Short | User usage | Description |
|--------|--------|--------|--------|
| 1. | beforeUnload | Use to save any component state that would be destroyed. `event.preventDefault()` would allow cancel the navigation/transition | Before anything is done, this would be  |
| 2. | transitionStart | Not sure, just an FWI? | It is basically at the same time, but it marks the start of animation |
| 3. | afterSwap | Modify HTML before it is rendered for the first time | The head is diffed, the body is swapped, and parts that should remain did remain |
| 4. | transitionEnd | _Is this the time that_  | The end of animation |
| 5. | pageLoad | Restore state to any component that may need it | New styles were added, and the script already ran | 


Is this right?

matthewp:
Yes, that's excellent @Akxe! The intent of this PR is not to add all possible events, but just to solidify the names. I like that you have a list though. I'd probably recommend adding that to a roadmap discussion issue, and we can use that as a guide when adding new events. I could also see some table like this making its way into the docs.

Akxe:
@matthewp My intent was not to add them, but to know what all the possible events would be. It could help us define what event would be used for what, and also it would help us name the events consistently across the whole action.

Akxe:
I do think that the addition of `beforeUnload` is a good idea because if a user keeps a form partially filled, the page should be able to prevent his departure.

Also, [ViewTransition API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API#the_view_transition_process) uses events `ready`, `finished` (and to some degree `updateDone`). I am not sure we want to use those, but worth considering, since they would mimic the naming from the standard that Astro ViewTransition is using/minicing.

matthewp:
Yeah, I intend to write up a new proposal shortly that includes some new events, I agree that's probably the next one to come, and some other additions such as a JS API to trigger a navigation to occur.

Issue #8161, opened 2023-08-20
Broken Portfolio Template


krishna-santosh:
### What version of `astro` are you using?

2.10.12

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

![image](https://github.com/withastro/astro/assets/75202541/8682ef81-1bd8-40c2-8e04-b3956687ef25)

Due to the recent merge #7078. 
I've also mentioned it before the merge in the Issue and PR's comment section https://github.com/withastro/astro/issues/7056#issuecomment-1640405566

### What's the expected result?

Website(Template) to look normal.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/withastro/astro/tree/latest/examples/portfolio?on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

kasutu:
After removing the loop Where the social links are located the issue has been reduced.
![image](https://github.com/withastro/astro/assets/64959978/386b27fe-442c-4576-8acf-13d5713d3006)

![image](https://github.com/withastro/astro/assets/64959978/a9015446-62dc-4dd6-855f-e88cc7ca4355)


I am running in `dev` mode and browsing in `edge`

PR #8185, opened 2023-08-22
Fix: changed inline-style to regular selector

dreyfus92:
## Changes

- Closes https://github.com/withastro/astro/issues/8161
- Replaced the inline-style to a regular selector, for some reason it was breaking the template. 

### Before:
image

### After:
image

### DIFF
--- a/examples/portfolio/src/components/Nav.astro
+++ b/examples/portfolio/src/components/Nav.astro
@@ -36,7 +36,7 @@ const iconLinks: { label: string; href: string; icon: keyof typeof iconPaths }[]
 			</template>
 		</menu-button>
 	</div>
-	<noscript class="menu-noscript">
+	<noscript>
 		<ul class="nav-items">
 			{
 				textLinks.map(({ label, href }) => (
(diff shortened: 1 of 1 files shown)


lilnasy:
Thanks for taking this on!

examples/portfolio/src/components/Nav.astro
```diff
@@ -60,7 +60,7 @@ const iconLinks: { label: string; href: string; icon: keyof typeof iconPaths }[]
 			}
 		</ul>
 	</noscript>
-	<noscript style="display: contents;">
+	<noscript style="menu-content">
```
lilnasy, line 63: The value for style seems invalid, does it do anything?

lilnasy: You're right, and the `<noscript>` tag doesn't need the style attribute (it's invalid and not doing anything), and the other one doesn't need a class (it doesn't match a rule anyway). They might be causes of confusion for users building on top of this example.

dreyfus92: So, should I remove both style selectors from both <noscript/> tags?

lilnasy: Yeah, that sounds good to me.

lilnasy (approved):

lilnasy:
This looks great, both with and without javascript!

lilnasy:
Congratulations on your first PR 🎉

and thanks again!

Issue #8059, opened 2023-08-14
In different `astro` versions, the execution order of parent and child components is inconsistent


J-env:
### What version of `astro` are you using?

2.8.2,2.10.7

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm,pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

astro2.8.5: [example](https://stackblitz.com/edit/github-sdeict-mdfwhi?file=src%2Fcomponents%2FParent.astro)

<img width="927" alt="image" src="https://github.com/withastro/astro/assets/25259920/30beeb26-102f-49d4-b90d-4e205094e90c">

astro2.10.7: [example](https://stackblitz.com/edit/github-sdeict-asom31?file=src%2Fcomponents%2FParent.astro)

<img width="914" alt="image" src="https://github.com/withastro/astro/assets/25259920/0b5ac6ec-5a81-4aa6-9f60-5f99931f3c6d">

This will cause descendant components not to get context data.

### What's the expected result?

Is there an api like react, preact to pass data across components
<img width="784" alt="image" src="https://github.com/withastro/astro/assets/25259920/80d0f42a-f61e-4ed3-9db4-44590ee0b075">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-sdeict-asom31?file=src%2Fcomponents%2FParent.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #8187, opened 2023-08-22
Fix Astro components parent-child render order

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/8059

Previously, the `renderComponent` API will kick off the render directly, however I didn't notice that component slots are inited eagerly (for head propagation) and it kicked off the rendering of slots eagerly (not good).

This PR reverts to the previous technique where the parent will render each child in parallel manually instead. It renders all the child into a buffer first, then flush it out when it's its turn to render.

NOTE: While the benchmark shows that this is faster, it's because I removed the `bufferChunks.length = 0` optimization which seems to do more harm than good for perf. If I ported that, the perf is actually slightly lower because things are less in parallel.

## Testing

Added a new test

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/mean-snakes-play.md
+++ b/.changeset/mean-snakes-play.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix Astro components parent-child render order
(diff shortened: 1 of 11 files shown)


bluwy:
!bench server-stress

matthewp (approved):

bluwy:
!bench server-stress

matthewp (approved):

Issue #8060, opened 2023-08-14
Env vriables with Vite and Cloudflare, the new replace of getRuntime()


Mahmoudgalalz:
### What version of `astro` are you using?

Latest

### Are you using an SSR adapter? If so, which one?

Cloudflare

### What package manager are you using?

npm

### What operating system are you using?

Mac,

### What browser are you using?

Chrome

### Describe the Bug

I have no access to environment variables via cloudflare, and I do override the connection string an a ts file and import it in a route

### What's the expected result?

I should be able to access the old getRuntime() out of the new usage Astro.Locals.runtime

as I cannot access this in the .ts files 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wj2hw8?file=src%2Fdb.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Ah, I think this consideration was overlooked when switching to `Astro.locals.runtime`. Ideally `import.meta.env` should still work, though. IMO you shouldn't need to use `getRuntime` for this.

alexanderniebuhr:
@Mahmoudgalalz `getRuntime()` should still be available as for now. We wanted to get feedback for potential issues, before removing it.

kelvinz:
I have a similar question, how can I access my cloudflare env variables at my API endpoint on the server?

Docs say,
```
import type { APIContext } from "astro";
import { getRuntime } from "@astrojs/cloudflare/runtime";

export async function get({request}: APIContext) => {
  const runtime = getRuntime(request);
  // the type KVNamespace comes from the @cloudflare/workers-types package
  const { MY_KV } = (runtime.env as { MY_KV: KVNamespace }));

  return {
    // ...
  };
};
```
But since getRuntime is gone, how do I access the env?

ematipico:
Probably we should write a better documentation. 

**Inside an `.astro` file**:

```astro
const runtime = Astro.locals.runtime;
```

**Inside an endpoint**

```js
export function get(context) {
	const runtime = context.locals.getRuntime();
}
```

kelvinz:
@ematipico this is the error I'm getting from ts, `Property 'getRuntime' does not exist on type 'Locals'`

ematipico:
> @ematipico this is the error I'm getting from ts, `Property 'getRuntime' does not exist on type 'Locals'`
> when I try running `const runtime = context.locals.getRuntime()`
> 
[...]

PR #8190, opened 2023-08-22
docs: better documentation for `runtime`

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8060

## Testing

N/A

## Docs

### DIFF
--- a/packages/integrations/cloudflare/README.md
+++ b/packages/integrations/cloudflare/README.md
@@ -75,8 +75,21 @@ It's then possible to update the preview script in your `package.json` to `"prev
 
 You can access all the Cloudflare bindings and environment variables from Astro components and API routes through `Astro.locals`.
 
-```js
+If you're inside an `.astro` file, you access the runtime using the `Astro.locals` global:
+
+```astro
 const env = Astro.locals.runtime.env;
+```
+
+From an endpoint:
+
+```js
+// src/pages/api/someFile.js
+export function get(context) {
+    const runtime = context.locals.runtime;
+    
+    return new Response("Some body");
+}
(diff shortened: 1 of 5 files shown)


ematipico:
cc https://github.com/alexanderniebuhr should we add something about how to type the runtime?

alexanderniebuhr:
> cc @alexanderniebuhr should we add something about how to type the runtime?

Sure thing, let me double check how typing would be best with the current types & will open another PR. Should not block this one.

packages/integrations/cloudflare/README.md
```diff
@@ -75,10 +75,23 @@ It's then possible to update the preview script in your `package.json` to `"prev
 
 You can access all the Cloudflare bindings and environment variables from Astro components and API routes through `Astro.locals`.
 
-```js
+If you're inside an `.astro` file, you access the runtime using the `Astro.locals` global:
+
+```astro
 const env = Astro.locals.runtime.env;
```
alexanderniebuhr, line 81: What do you think about this change? In the sentence above you talked about `runtime`, so I think it makes sense to update the example, not sure if we should show that `env` is inside it.
```suggestion
const runtime = Astro.locals.runtime;
```

alexanderniebuhr (commented):
This makes sense. 🎉

ematipico:
@alexanderniebuhr I committed a change where the types are now exported and improved the docs showing how users can type the runtime. Please have a look!

alexanderniebuhr:
I like it!

sarah11918 (approved):
Docs approves! :partying_face:

Issue #8192, opened 2023-08-22
Astro config conflict / misalignment with Vite config on server.open expected type ( boolean vs boolean & string)


gkatsanos:
### What version of `astro` are you using?

2.10.4

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

This works:
```
// astro.config.mjs
export default defineConfig({
  server: {
    port: 3000,
    open: true,
  },
```

looking into the vite config though: https://vitejs.dev/config/server-options.html#server-open 
this should also work:
```
export default defineConfig({
  server: {
    open: '/docs/index.html',
  },
})
```

but it doesnt: 
```

[config] Astro found issue(s) with your configuration:
  ! server.open  Expected boolean, received string.
```

Follow up question: in the astro docs, `open` is not documented at all. Yet it still works. This is a little confusing.

### What's the expected result?

Dev server should open at the designated path. 

### Link to Minimal Reproducible Example

not relevant

### Participation

- [X] I am willing to submit a pull request for this issue.
- Here you go: https://github.com/withastro/astro/pull/8193

gkatsanos:
from astro.d.ts:
```
    /**
     * @name server.open
     * @type {boolean}
     * @default `false`
     * @version 2.1.8
     * @description
     * Control whether the dev server should open in your browser window on startup.
     *
     * ```js
     * {
     *   server: { open: true }
     * }
     * ```
     */
```

from node_modules/.pnpm/vite@4.4.9_sass@1.62.0/node_modules/vite/dist/node/index.d.ts
```
    /**
     * Open browser window on startup
     */
    open?: boolean | string;
```

gkatsanos:
Would appreciate PR push permissions so I can make a PR on this. It's here waiting on my local :/

PR #8193, opened 2023-08-22
Fixing type to align with Vite configuration

gkatsanos:
## Changes

- What does this change? Fixes: https://github.com/withastro/astro/issues/8192 

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -349,7 +349,7 @@ type ServerConfig = {
 
 	/**
 	 * @name server.open
-	 * @type {boolean}
+	 * @type {boolean | string}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
@@ -360,8 +360,14 @@ type ServerConfig = {
 	 *   server: { open: true }
 	 * }
 	 * ```
+	 * or
+	 * ```js
+	 * {
+	 *   server: { open: '/about' }
+	 * }
+	 * ```
 	 */
-	open?: boolean;
+	open?: boolean | string;
 };
 
 export interface ViteUserConfig extends vite.UserConfig {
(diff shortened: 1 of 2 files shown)


bluwy (changes requested):
This doesn't fix the issue yet. We need to update the schema to allow passing strings:

https://github.com/withastro/astro/blob/a571a1ac07fad4a898158b04594e03ad0c1de83b/packages/astro/src/core/config/schema.ts#L144

And make sure we're also using the string:

https://github.com/withastro/astro/blob/a571a1ac07fad4a898158b04594e03ad0c1de83b/packages/astro/src/core/dev/container.ts#L59-L64

And also that the CLI accepts it:

https://github.com/withastro/astro/blob/a571a1ac07fad4a898158b04594e03ad0c1de83b/packages/astro/src/cli/flags.ts#L24

gkatsanos:
Quick question, @bluwy , if this is a core vite feature, why do we need to override / rewrite with our own logic? (I'm sorry if this is a dumb question, just figuring things out)

bluwy:
I don't have the full context, but it's likely because we want to present the dev server as Astro's dev server, and not Vite's. Vite would be an implementation detail. So common server options like host and port are top-level in the Astro config, but if you want to tap into Vite specific server options, like setting up proxies, then you'd have to use the `vite.server` config.

gkatsanos:
I see. I'm not sure I have the experience in terms of working with the core of the FW to do all the changes required..

matthewp:
Is this ready for another review? Or still being worked on?

gkatsanos:
@matthewp I don't think I can go through with this, some parts of the codebase are a little blackbox to me.

bluwy:
I'll close this for now to clean up the PRs and track the progress in the issue. Thanks for the PR to kick off the discussion

PR #8196, opened 2023-08-23
Prevent bundling sharp

bluwy:
## Changes

`sharp` can't be bundled and it fails in runtime if so. It's bundled by default with pnpm unless the user installs `sharp` themselves.

## Testing

I'm updating some dependencies and it causes the Vercel test to OOM, this fixes it.

I also tested this with the `ssr` example app and it still works.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/spicy-stingrays-cheer.md
+++ b/.changeset/spicy-stingrays-cheer.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent bundling sharp as it errors in runtime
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Maybe for later, but we should update the relative `README.md` that explains the plugins.

bluwy:
Yeah. I think this place isn't quite right to set the external/noExternal too, I'd probably put it in `static-build.ts` `ssrBuild` directly, but it's hard to merge it in with the user config.

I could tweak Vite's `mergeConfig` API so that it works better though and we can put it in that file later though.

Issue #8154, opened 2023-08-18
Astro.locals, set in middleware, available on 404 page in dev mode but not in production mode


xriter:
### What version of `astro` are you using?

2.10.12

### Are you using an SSR adapter? If so, which one?

Node

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Safari

### Describe the Bug

After some previous bugs around the 404 page have been solved, I still experienced problems with my 404 page not rendering. I now found the reason: another bug.

When using `middleware.ts` to set `Astro.locals`, you can then use those values in the `404.astro` page when it's in dev mode. But when in production mode, those values are undefined. (This caused the page to crash without me knowing, resulting in a blank 404 page).

👉 **So in short: the behaviour of `Astro.locals` availability in `404.astro` between dev and production mode is inconsistent.**

This problem started from Astro 2.9.7. Before that it worked as expected.

To repro and see the difference between the 2 modes, go to the stackblitz link.
Then first run `npm run dev` and then `npm run build && npm run preview`

Dev mode (the expected result):
<img width="1481" alt="Scherm­afbeelding 2023-08-19 om 01 37 38" src="https://github.com/withastro/astro/assets/70454441/87e24e26-9063-4522-bec1-aaf766b93afb">

Production mode:
<img width="1493" alt="Scherm­afbeelding 2023-08-19 om 01 38 03" src="https://github.com/withastro/astro/assets/70454441/34254afd-4c26-4dd5-8160-3e63ac615647">

(This is a continuation of the problem I first described in issue #8054, but turns out to be a separate bug.)

### What's the expected result?

I expect `Astro.locals`, set in `middleware.ts` to be available in `404.astro` in production mode.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-trfvka-twqzu5?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Might be related #7881

lilnasy:
@xriter Does middleware work at all for 404?

If not, this is the same issue I brought up in #7881.

lilnasy:
Thanks for opening this issue and confirming!

To clarify, this issue is **not** a duplicate of 7881. When I reported it in that issue, I should've created another issue instead.

lilnasy:
Is the [workaround](https://github.com/withastro/astro/issues/7881#issuecomment-1664540900) possible in your case?

[...]

PR #8200, opened 2023-08-23
fix(ssr): correctly call the middleware when rendering error pages

### LINKED ISSUE #7881: Middleware not called for non-custom 404 page

2.9.6

Node

npm

Mac

Safari

### Describe the Bug

When you have `middleware.ts` setup, and a **custom** 404 page, the middleware will get called when you hit a 404 page. However, if you don't have a custom 404 page but rely on the default, the middleware won't get called.

Wondering if this is a deliberate decision?

Consistency between the custom and non-custom 404 page. Both should call the middleware.

https://stackblitz.com/edit/github-pt4enm?file=src%2Fpages%2Findex.astro

### Participation

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8154

This fix goes into `next` because it requires some infrastructure that we have there but not in `main`.

This PR removes the `onRequest` function from the "ComponentInstance", instead we move its resolution inside the pipeline. The middleware now becomes 100% detached from a page/route. 

This change fits perfectly because the SSR build already creates a `middleware.mjs` file, and we already save the `middlewareEntryPoint` file.

## Testing

I created a new test case.

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/build/plugins/plugin-manifest.ts
+++ b/packages/astro/src/core/build/plugins/plugin-manifest.ts
@@ -237,6 +237,10 @@ function buildManifest(
 		// Set this to an empty string so that the runtime knows not to try and load this.
 		entryModules[BEFORE_HYDRATION_SCRIPT_ID] = '';
 	}
+	const isEdgeMiddleware =
+		// TODO: remove in Astro 4.0
+		settings.config.build.excludeMiddleware ||
+		settings.adapter?.adapterFeatures?.edgeMiddleware;
 
 	const ssrManifest: SerializedSSRManifest = {
 		adapterName: opts.settings.adapter?.name ?? '',
@@ -250,7 +254,7 @@ function buildManifest(
 		clientDirectives: Array.from(settings.clientDirectives),
 		entryModules,
 		assets: staticFiles.map(prefixAssetPath),
-		middlewareEntryPoint: internals.middlewareEntryPoint?.toString() ?? undefined,
+		middlewareEntryPoint: !isEdgeMiddleware ? internals.middlewareEntryPoint?.toString() : undefined,
 	};
 
 	return ssrManifest;
--- a/packages/astro/src/core/build/plugins/plugin-manifest.ts
+++ b/packages/astro/src/core/build/plugins/plugin-manifest.ts
@@ -93,14 +93,19 @@ export function pluginManifest(
 				}
 
 				const manifest = await createManifest(options, internals);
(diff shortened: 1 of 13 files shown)


packages/astro/src/core/redirects/component.ts
```diff
@@ -12,6 +12,5 @@ export const RedirectComponentInstance: ComponentInstance = {
 
 export const RedirectSinglePageBuiltModule: SinglePageBuiltModule = {
 	page: () => Promise.resolve(RedirectComponentInstance),
-	onRequest: (ctx, next) => next(),
```
ematipico, line 15: As you can see, there's no need for this stub anymore.

packages/astro/src/core/build/plugins/plugin-manifest.ts
```diff
@@ -245,6 +250,7 @@ function buildManifest(
 		clientDirectives: Array.from(settings.clientDirectives),
 		entryModules,
 		assets: staticFiles.map(prefixAssetPath),
+		middlewareEntryPoint: internals.middlewareEntryPoint?.toString() ?? undefined,
```
ematipico, line 253: We saved the entry point in the SSR manifest, as it was a route.

packages/astro/test/fixtures/middleware-dev/src/pages/404.astro
```diff
@@ -0,0 +1,6 @@
+---
```
lilnasy, line 1: If 404.astro exists, this issue might be fixing #8154, **not** #7881. I think the older issue may even be intended behavior.

ematipico: @lilnasy why do you think #7881 should be the intended behaviour? Genuinely asking! You might be onto something

lilnasy: I'm thinking about the cases where an adapter tries to handle unmatched requests by bypassing the core app.

Some serverless platforms may not even call `app.render()`: if they find no matching route and no prerendered page, they might fallback to a platform-specific 404 page.

On a different note, if we were to wrap middleware around our default response (`new Response(null, { status: 404 })`), it would likely cause another wave of issues because I doubt many users are accounting for `next()`'s body being null.

ematipico: That makes total sense! I agree with every point.

PR #8206, opened 2023-08-23
View Transition: swap attributes of document's root element

martrapp:
## Changes

fix: extends swap() to include the attributes of the documents root element

## Testing

added e2e test

## Docs
n/a
bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -136,6 +136,18 @@ const { fallback = 'animate' } = Astro.props as Props;
 			// Remove them before swapping.
 			doc.querySelectorAll('head noscript').forEach((el) => el.remove());
 
+			// swap attributes of the html element
+			// - delete all attributes from the current document
+			// - insert all attributes from doc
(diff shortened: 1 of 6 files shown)


martrapp:
@matthewp would like to ask you for review.

packages/astro/components/ViewTransitions.astro
```diff
@@ -136,6 +136,18 @@ const { fallback = 'animate' } = Astro.props as Props;
 			// Remove them before swapping.
 			doc.querySelectorAll('head noscript').forEach((el) => el.remove());
 
+			// swap attributes of the html element
+			// - delete all attributes from the current document
+			// - insert all attributes from doc
+			// - reinsert all original attributes that are named 'data-astro-*'
+			const html = document.documentElement;
+			const astro = [...html.attributes].filter(
+				({ name }) => (html.removeAttribute(name), name.startsWith('data-astro-'))
```
matthewp, line 145: ohhh very clever

matthewp:
This does fix a bug, although I don't think we have an issue for it. It was reported on discord IIRC, that the `lang` from the new page is not present.

PR #8207, opened 2023-08-23
Update built-in view transitions

natemoo-re:
## Changes

- Closes plt-809
- Updates animations of built-in `fade` and `slide` animations
- Adds built-in `none` animation, disabling the default user-agent animation
- ~~Adds built-in `crossfade` animation, similar to default user-agent animation but faster~~ `crossfade` behavior is now used by the default `fade`
- Defaults elements without a `transition:animate` directive to `fade`
- Renames `morph` to `initial` to more closely align with CSS terminology
- ~~Sets the default `root` view transition to `none`~~ removed this opinionated change

## Testing

Tested manually. Here's a demo!

First run is all using our default (`fade`), then `none` on `html`, then `slide`

https://github.com/withastro/astro/assets/7118177/17d4dd58-acf4-400c-a6b7-49129496a678

## Docs

Will coordinate with docs to get this into the update guide


FredKSchott (changes requested):
I may have missed a conversation, but I'd understood this work to just be to tweak the values of the existing animations. If you haven't yet, confirm that this change is in scope for @matthewp and docs given that we're post-RC / so close to 3.0 release.

Mainly looking at the proposal! Will leave the code review for others:
  
- Updates animations of built-in `fade` and `slide` animations
  - Fred: Yessssss, LGTM
- Adds built-in `none` animation, disabling the default user-agent animation
  - Fred: I think this makes sense... would defer to @matthewp on this one.
- Adds built-in `crossfade` animation, similar to default user-agent animation but faster
  - Fred: very confused about the difference between "fade" and "crossfade", both from a naming perspective (aren't these both the same?) and from looking at the code (the differences seem minor and basically personal preference). I see our responsibility as having a strong preference for a single good "fade", and if anyone wants more control they can customize their own.
  - Guidance: remove `crossfade`, have a single `fade` with whichever duration you recommend and `EASE_IN_OUT_QUINT` as the best default easing.
- Defaults elements without a `transition:animate` directive to `crossfade`
  - Fred: LGTM with the advice above to use a single `fade`
- Renames `morph` to `initial` to more closely align with CSS terminology
  - Fred: I see how you got here logically, but this probably will be pretty confusing to people.  Assuming 99% of people don't know much about View Transitions, and what the initial behavior is. It's growing on me vs. when I first saw it, but I think I would still recommend `morph` as the most clear API name. Possibly compromise: support `initial` as an alias to `morph`?
  - Guidance: I'm not really sure to be honest, I guess I could go either way. Would love @matthewp's thoughts.
- Sets the default `root` view transition to `none`
  - Our first piece of advice is usually to customize this, so this is our opinionated default
  - This can be reset by setting `transition:animate="initial"` on the `html` element
  - Fred: This feels in conflict with the reasoning above: in one change you bring us closer to CSS terminology/defaults and here you change them. I think I'm sold on the concept of "fewest surprises/changes vs. browser behavior" so I'd suggest walking this change out in favor of our faster `fade`.
  - Guidance: Walk this out, default `root` view transition should `fade` (similar to browser default, but our faster version).

matthewp:
Everything looks great to me here! The naming as well, I like the choice to switch the `morph` which was inaccurate.

As for the fade vs. crossfade + default discussion, my take is:

1. We got a bit of feedback that the default didn't feel great. @bholmesdev mentioned this, as did one prominent OS maintainer (whose name escapes me at the moment).
    1. Given that, changing the default seems like a good idea! We give people a way to opt-out so this seems reasonable.
    1. I can understand why `none` sort of makes sense here. You didn't opt-in this element to be transitioned, like you do when you use `transition:name`.
2. _But_, I do think we should use the crossfade (or fade if renamed) as the default here because:
    1. When you use `transition:name` you would get a different default animation. It feels like it should be the same.
    1. The `root` is an implicit transition name, even though you didn't type it yourself, I think once you learn more about VT you would expect it to be treated the same.
    1. Also worth mentioning that, this feature is built around *view transitions*, so I think we should embrace that an animation is indeed the default. It's not meant to act like a traditional CSR.

As for crossfade vs fade, I don't think I know enough about the difference to give an opinion. Do these things have different meanings? What was the reason for having both @natemoo-re ? I think if they are meaningfully different then we could have both, but I could also see it as being confusing. If there's just a small difference could we combine them into a single animation and provide some flag in the JS API to make it behave like the other one?

natemoo-re:
Thanks for all the feedback! 

Definitely agree with `fade` vs `crossfade` being confusing.

- `crossfade` is similar to the browser's default, where the outgoing element fades out WHILE the incoming element fades in. The elements overlap so it's a smooth transition (hence "cross"-fade).
- `fade` has no overlap, so the outgoing element fades out THEN the incoming element fades in. Visually, there's a small time where the element seems to have completely disappeared.

Fewer options is probably better, and I think `crossfade` is the best default. Happy to rename this one to `fade`. Or maybe we want to keep both but adjust the naming so there's not so much conceptual overlap?

Also agree with keeping the `root` transition. We can document that `transition:animate="none"` on the `html` element disables the animation, but it's opt-in to align with the default UA behavior.

natemoo-re:
Alright, resolved this feedback by making the following changes:
- `crossfade` has been renamed to `fade`, the existing `fade` behavior has been removed
- the `root` animation is no longer disabled by default. We can instruct users to add `<html transition:animate="none">` if they want to disable the page-level fade.
- Sounds like we're happy with renaming `morph` => `initial`, so I kept that

matthewp (approved):
code looks good!

bholmesdev:
Thanks for those changes @natemoo-re! Agree with all of the tweaks. I'll admit that `initial` isn't my favorite naming, since it doesn't describe what the animation will do once enabled (it's basically a "go read the docs" requirement). I'm confused why `morph` is misleading given this is the name for the view transitions tweening effect? Would a name like "flip" be a better description?

matthewp:
!preview vt-anim

matthewp:
Doing a preview release to play around with. Want to confirm that `<div transition:name>` still does the position transform. If we are defaulting to `crossfade` then it might not.

natemoo-re:
> I'll admit that `initial` isn't my favorite naming, since it doesn't describe what the animation will do once enabled (it's basically a "go read the docs" requirement). I'm confused why `morph` is misleading given this is the name for the view transitions tweening effect? Would a name like "flip" be a better description?

I find the `morph` / `flip` naming confusing because those aren't ever referenced in the specification. This setting is not _an animation_, it's the explicit _absence of a custom animation_, which will use the default [User Agent Stylesheet described in the specification](https://drafts.csswg.org/css-view-transitions-1/#ua-styles). Per the spec, the browser sets `transform: translate` values for the `::view-transition-old` and `::view-transition-new` snapshots when performing a View Transition.

So our custom `fade` animation still exhibits this "morph" behavior because we do NOT modify the `transform` property. So IMO it's misleading to imply that `morph` is any sort of special construct, it's just the default browser behavior.

natemoo-re:
> Want to confirm that `<div transition:name="whatever">` still does the position transform.

Yep, I was just testing this as well! It does as long as your animation doesn't specify a custom `transform` value.

natemoo-re:
Yes, `default` crossed my mind, but I the same concern. Astro's default is not the thing named `default` but is `fade`?! That just sounds extremely odd.

bholmesdev:
@natemoo-re Ah ha, that makes much more sense. I'm happy with the name `initial` then! To clarify though, when would a user reach for `initial` over `fade`? Are they the same thing?

.changeset/five-geese-crash.md
```diff
@@ -0,0 +1,9 @@
+---
+'astro': major
+---
+
+Adjust default [View Transition](https://docs.astro.build/en/guides/view-transitions/) animations.
```
sarah11918, line 5:
```suggestion
Change the [View Transition built-in animation](https://docs.astro.build/en/guides/view-transitions/#built-in-animation-directives) options.
```
I think this sounds a little more stand up and take notice. Also links directly to the section.

.changeset/five-geese-crash.md
```diff
@@ -0,0 +1,9 @@
...
+
+The `transition:animate` value `morph` has been renamed to `initial`. Elements with a `transition:name` directive but no `transition:animate` directive will now default to `fade`.
```
sarah11918, line 7:
```suggestion
The `transition:animate` value `morph` has been renamed to `initial`. Also, this is no longer the default animation.

 If no `transition:animate` directive is specified, your animations will now default to `fade`.
```
I'm not entirely sure this is correct, but it was a mouthful. Is it the case that `fade` is default instead of `initial` ALL the time? The "with a `transition:name` " threw me off. Is it only in this particular case that there IS a transition name? (not every element will have one, because it's not required to set.) What if there is NO transition name, either? Is that a red herring or is that actually important in deciding which animation is the default?

If the above is correct (`fade` is always the default) then there's no need to mention "with a transition:name". If there's a different default with no transition name, then we also need to specify that, and this becomes a little bit more involved of a definition.  :smile:

.changeset/five-geese-crash.md
```diff
@@ -0,0 +1,9 @@
...
+
+Astro also supports a new `transition:animate` values, `none`. This can be used to disable animated full page transitions by setting `transition:animate="none"` on your `html` element.
```
sarah11918, line 9:
```suggestion
Astro also supports a new `transition:animate` value, `none`. This value can be used on a page's `<html>` element to disable animated full-page transitions on an entire page.
```
maybe a little clearer?

sarah11918 (commented):
See what you think @natemoo-re !

FredKSchott (approved):
Love all the changes/updates, thanks for taking the feedback @natemoo-re! Big +1 with `initial` given the reasons not to use `morph`.

packages/astro/src/transitions/index.ts
```diff
@@ -11,28 +16,28 @@ export function slide({
 				{
 					name: 'astroFadeOut',
 					duration: duration ?? '90ms',
-					easing: 'cubic-bezier(0.4, 0, 1, 1)',
+					easing: EASE_OUT_QUART,
```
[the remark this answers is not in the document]
natemoo-re: Great catch, I missed this while refactoring. Switched to EASE_IN_OUT everywhere.

packages/astro/src/runtime/server/transition.ts
```diff
@@ -22,78 +21,80 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
-	const scope = createTransitionScope(result, hash);
+	result._metadata.extraHead.push(markHTMLString(`<style>${sheet.toString()}</style>`));
+	return scope;
+}
 
-	// Default transition name is the scope of the element, ie HASH-1
-	if (!transitionName) {
-		transitionName = scope;
-	}
+class ViewTransitionStyleSheet {
```
natemoo-re, line 65: Cleaned the style generation up significantly with a builer class—I highly recommend reviewing this file without the diff!

packages/astro/src/runtime/server/transition.ts
```diff
@@ -22,78 +21,80 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
-::view-transition-new(${transitionName}) {
-	${stringifyAnimation(animations.forwards.new)}
-}
-[data-astro-transition-fallback=new] [data-astro-transition-scope="${scope}"] {
-	${stringifyAnimation(animations.forwards.new)}
-}
+	toString() {
+		const { scope, name } = this;
+		const [modern, fallback] = [this.modern, this.fallback].map(rules => rules.join(''));
+		return [`[data-astro-transition-scope="${scope}"] { view-transition-name: ${name}; }`, this.layer(modern), fallback].join('')
```
natemoo-re, line 74: View Transition pseudos are now wrapped in `@layer astro` so that user CSS will always take precedence.

matthewp (approved):

PR #8217, opened 2023-08-24
View Transitions: adding data-astro-reload forces browser default behavior for navigation

martrapp:
THIS IS THE SECOND TRY for the failed PR #8153 

## Changes

Specify `data-astro-reload` (no value) on an anchor element to force the browser to ignore view transitions and fall back to default loading.

This is helpful when navigating to documents that have different content-types, e.g. application/pdf, where you want to use the build in viewer of the browser.
Example: `<a href='/my.pdf' data-astro-reload>...</a>`

## Testing

e2e test added

### DIFF
--- a/.changeset/light-badgers-mate.md
+++ b/.changeset/light-badgers-mate.md
@@ -1,0 +1,8 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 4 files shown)


matthewp (approved):

matthewp:
Fixed in https://github.com/withastro/astro/issues/7947

PR #8227, opened 2023-08-25
refactor: internal refactor to use the `Logger`

ematipico:
## Changes

This internal refactor ditches the use of `logging` and `LogOptions` in favour of the new `Logger` API. 

I believe there are still some parts of the code that still use the `logging`. I will refacto them later.

Closes PLT-845

## Testing

Current CI should pass.

## Docs

N/A


packages/astro/test/units/test-utils.js
```diff
@@ -187,7 +188,7 @@ export function createBasicEnvironment(options = {}) {
 		clientDirectives: getDefaultClientDirectives(),
 		resolve: options.resolve ?? ((s) => Promise.resolve(s)),
 		routeCache: new RouteCache(options.logging, mode),
-		logging: options.logging ?? defaultLogging,
+		logging: options.logging ?? defaultLogger,
```
bluwy, line 191: I think the two `logging` here should be `logger` since the `Environment` types had it updated too.

packages/astro/src/core/config/logging.ts
```diff
@@ -1,13 +1,12 @@
 import type { AstroInlineConfig } from '../../@types/astro.js';
-import type { LogOptions } from '../logger/core.js';
+import { Logger } from '../logger/core.js';
 import { nodeLogDestination } from '../logger/node.js';
 
-export function createNodeLogging(inlineConfig: AstroInlineConfig): LogOptions {
-	// For internal testing, the inline config can pass the raw `logging` object directly
-	if (inlineConfig.logging) return inlineConfig.logging;
+export function createNodeLogger(inlineConfig: AstroInlineConfig): Logger {
+	if (inlineConfig.logging) return new Logger(inlineConfig.logging);
```
bluwy, line 6: I think ideally `AstroInlineConfig` has a new `logger` option in place of `logging`, so we could do `if (inlineConfig.logger) return new Logger(inlineConfig.logger);`.

Might have to update tests that passes `logging` to `fixture.startDevServer()` and `fixture.build()` I think. If it turns out to be tricker and we need to keep logging, then I think we should keep the previous comment for now.

bluwy (approved):
Awesome!

Issue #7159, opened 2023-05-22
Page prop is type never with InferGetStaticPropsType


bilco105:
### What version of `astro` are you using?

2.5.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Safari

### Describe the Bug

This looks to be a repeat of #6507 which was closed, though I'm still seeing the same behaviour.

The following code in `src/pages/blog/[...page].astro` doesn't infer props correctly, causing `page` in the snippet below to be typed as `never`.

```
export const getStaticPaths = (async ({ paginate }) => {
  const posts = (await getCollection('blog')).sort(
    (a, b) => a.data.pubDate.valueOf() - b.data.pubDate.valueOf()
  );
  return paginate(posts, { pageSize: 10 })
}) satisfies GetStaticPaths

type Props = InferGetStaticPropsType<typeof getStaticPaths>

const { page } = Astro.props as Props
```

I can work around this by manually typing the Props as follows, but it would be good to not to have keep doing this throughout the project.

```
interface Props {
  page: Page<CollectionEntry<'blog'>>
}
```

Issue can be reproduced by running `npm run check` within the attached Stackblitz

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xpiyfp?file=src%2Fpages%2Fblog%2F%5B...page%5D.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #8048, opened 2023-08-13
fix getStaticPaths typing

hori-ryota:
## Changes
fix typing around getStaticPaths

- Be short and concise. Bullet points can help!

* resolve #7159 

## Testing

```typescript
import type { InferGetStaticPropsType, PaginateFunction } from "astro";

type Post = {
  title: string;
  body: string;
};

export function getStaticPaths({
  paginate,
}: {
  paginate: PaginateFunction<Post>;
}) {
  return paginate(
    [
      {
        title: "title1",
        body: "body1",
      },
    ],
    { pageSize: 10 },
  );
}
type Props = InferGetStaticPropsType<typeof getStaticPaths>;
```

before
```typescript
type Props = never
```

after

```typescript
type Props = Props & {
    page: Page<Post>;
}
```

## Docs

/cc @withastro/maintainers-docs for feedback!

PR #8229, opened 2023-08-25
feat(paginate): Return exact types from paginate()

Princesseuh:
## Changes

`paginate()` currently returns a naive `GetStaticPathsResult` which override the actual types it returns, this PR makes it so it actually returns what is being inputted, which because of the fact that it allows you to inject params and props, made this strangely hard. Maybe there's an easier way to do it, but I couldn't figure it out

This is similar to https://github.com/withastro/astro/pull/8048, but it doesn't require user to manually type `PaginateFunction`

This PR also fixes https://github.com/withastro/astro/issues/7159, mostly by accident

## Testing

Tested manually

## Docs

N/A. Docs document the return values of paginate(), so this really just makes it more accurate to what's currently described


packages/astro/client.d.ts
```diff
@@ -57,7 +57,7 @@ declare module 'astro:assets' {
 	};
 
 	type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
-	type Simplify<T> = { [KeyType in keyof T]: T[KeyType] };
+	type Simplify<T> = { [KeyType in keyof T]: T[KeyType] } & {};
```
Princesseuh, line 60: Just a small fix, eventually when we refactor types, we should probably have a "internal-type-utils" file somewhere with this ~~and all the other nonsense I add~~

packages/astro/src/core/render/route-cache.ts
```diff
@@ -50,7 +51,9 @@ export async function callGetStaticPaths({
 	// Calculate your static paths.
 	let staticPaths: GetStaticPathsResult = [];
 	staticPaths = await mod.getStaticPaths({
-		paginate: generatePaginateFunction(route),
+		// Q: Why the cast?
+		// A: So users downstream can have nicer typings, we have to make some sacrifice in our internal typings, which necessitate a cast here
+		paginate: generatePaginateFunction(route) as PaginateFunction,
```
Princesseuh, line 56: I think there is a way to make this work without the cast, but I couldn't for my life figure it out. I think the disconnect between the internal implementation and what the user actually uses makes it very hard to do in one type

natemoo-re: Seems like a reasonable tradeoff!

natemoo-re (approved):
Looks great, approved!

Would you mind closing out https://github.com/withastro/astro/pull/8048 and the related issue after merging this since they overlap?

matthewp (approved):

PR #8231, opened 2023-08-25
View Transitions: use history.scrollRestoration="manual"

justinbeaty:
## Changes

- When using the morph animation, there is a scroll position jump when navigating back / forward through history.
- This is because the browser is restoring the old scroll position before the DOM is swapped.
- This PR simply changes `history.scrollRestoration = "manual";` to tell the browser we'll take care of the scroll position ourselves.

## Testing

- Create a morph animation with `transition:name`
- Navigate back / forwards in history at various scroll positions and see the smooth animation

## Docs

- No doc updated needed, this was a bug.

### DIFF
--- a/.changeset/twenty-crabs-fry.md
+++ b/.changeset/twenty-crabs-fry.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+View Transitions: use history.scrollRestoration="manual"
--- a/.changeset/twenty-crabs-fry.md
+++ b/.changeset/twenty-crabs-fry.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
-View Transitions: use history.scrollRestoration="manual"
+Fixes scroll behavior when using View Transitions by enabling `manual` scroll restoration
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -35,12 +35,6 @@ const { fallback = 'animate' } = Astro.props as Props;
 		persistState({ index: currentHistoryIndex, scrollY: 0 });
 	}
 
-	// With the default "auto", the browser will jump to the old scroll position
-	// before the ViewTransition is complete.
-	if (history.scrollRestoration) {
-		history.scrollRestoration = "manual";
-	}
-
 	const throttle = (cb: (...args: any[]) => any, delay: number) => {
 		let wait = false;
 		// During the waiting time additional events are lost.
@@ -361,7 +355,16 @@ const { fallback = 'animate' } = Astro.props as Props;
 			// Just ignore stateless entries.
 			// The browser will handle navigation fine without our help
 			if (ev.state === null) {
+				if (history.scrollRestoration) {
+					history.scrollRestoration = "auto";
+				}
 				return;
+			}
+
+			// With the default "auto", the browser will jump to the old scroll position
+			// before the ViewTransition is complete.
+			if (history.scrollRestoration) {
+				history.scrollRestoration = "manual";
 			}
 
 			const state: State | undefined = history.state;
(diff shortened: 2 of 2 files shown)


matthewp (approved):

matthewp:
Perfect, thank you!

matthewp:
@martrapp this sound right to you?

justinbeaty:
Even more generally, this fixes page jumping for all SPA page transitions when using back / forward navigation. The morph transitions were just the most noticeable.

.changeset/twenty-crabs-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
natemoo-re, line 2:
```suggestion
'astro': patch
```

.changeset/twenty-crabs-fry.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+View Transitions: use history.scrollRestoration="manual"
```
natemoo-re, line 5:
```suggestion
Fixes scroll behavior when using View Transitions by enabling `manual` scroll restoration
```

natemoo-re (approved):
Thanks for the PR! Looks great!

martrapp:
Hi @justinbeaty! Thanks a lot for the PR. I wasn't really able to see what scrolling is suppressed by `scrollRestoration="manual"` for navigating to different pages. Do you have a minimal example where I can observe the difference in behavior? 

<ins>In its current form, the PR breaks the in-page history navigation</ins> (back/forward between links to fragments on the same page).

Currently, we only actively handle navigation between different pages.
We leave navigation on the same page to the browser. 
Switching all history entries to "manual" does not work for intra-page navigation.

justinbeaty:
@martrapp Here's a reproducible example: https://github.com/justinbeaty/astro-scroll-restoration-mwe

There's also the `with-pr` branch that uses the `ViewTransition.astro` file from this PR.

Here's a video before the PR. Notice how the colored box jumps around, and also you can see flashes of red (from the footer) in some cases.

https://github.com/withastro/astro/assets/51970393/98ab3958-d8ae-42b1-afbf-deb70bcafc53

Here's a video after: 

https://github.com/withastro/astro/assets/51970393/b8bf3084-de66-4706-93a8-fdfab27907fd

I will investigate the hash fragment issue, but I'm quite certain `history.scrollRestoration = "manual";` is the correct setting.

martrapp:
Quite a difference, thank you for the examples. 
Maybe we should set "manual" just before we add new entries with pushState and set it to "auto" when we fall through to the browsers default handling.

justinbeaty:
That might be an easy solution, otherwise we have to pushState on fragment changes, but avoid fetching the page. But that's a pretty significant refactor it seems like.

justinbeaty:
@martrapp Hi Martin, I just tried this on my minimal reproducible repo, which seems like it might in fact resolve the issue. I've only done light testing so I could be missing something (much like I missed the issues with the hash fragments in the first place.)

https://github.com/justinbeaty/astro-scroll-restoration-mwe/commit/844e635b16657e020a3506caca8f3d33cbdfe50b

I can update this PR with that change as well if there's not a cleaner way to do this. I'm also not opposed to a replacement PR or commit by the Astro team if the solution is more involved. I haven't fully grokked the `ViewTransitions.astro` file yet.

martrapp:
@justinbeaty I'm sure that resetting scrollRestoration to "auto" at the beginning of the click handler and adding "manual" scroll restoration just before the two pushStates is a real improvement! In fact, this might also be necessary to solve another scrolling issue. Currently I don't see any chance that your change would make anything worse ;-)

Thanks to your discovery, the issue with the hash fragments is now covered by the e2e tests.

You found the bug and I would be happy to see your updated PR in the code base ;-)

justinbeaty:
@martrapp Done, I'm glad to know that it might fix another bug too.

matthewp (approved):
lgtm

PR #8239, opened 2023-08-26
Vercel adapter default changes

matthewp:
## Changes

- Makes `functionPerRoute` default to `true`.
- Adds back in stubs for the edge adapter, to error and point people towards using serverless + edge middleware.

## Testing

- Tests should still work

## Docs

- Needs docs review.

### DIFF
--- a/.changeset/silly-dolphins-try.md
+++ b/.changeset/silly-dolphins-try.md
@@ -6,4 +6,4 @@ Vercel adapter now defaults to `functionPerRoute`.
 
 With this change, `@astrojs/vercel/serverless` now splits each route into its own function. By doing this, the size of each function is reduced and startup time is faster.
 
-You can disable this option, which will cause the code to be bundled into a single function, by setting `functionPerRoute` to false.
+You can disable this option, which will cause the code to be bundled into a single function, by setting `functionPerRoute` to `false`.
--- a/.changeset/silly-dolphins-try.md
+++ b/.changeset/silly-dolphins-try.md
@@ -1,0 +1,9 @@
+---
+'@astrojs/vercel': major
+---
+
+Vercel adapter now defaults to a function per route
+
+With this change, in `@astrojs/vercel/serverless`, now each route is split into its own function. By doing this the size of each function is reduced and startup time is faster.
+
+You can disable this option, which will cause the code to be bundled into a single function, by setting `functionPerRoute` to false.
--- a/.changeset/silly-dolphins-try.md
+++ b/.changeset/silly-dolphins-try.md
@@ -4,6 +4,6 @@
 
 Vercel adapter now defaults to a function per route
 
-With this change, in `@astrojs/vercel/serverless`, now each route is split into its own function. By doing this the size of each function is reduced and startup time is faster.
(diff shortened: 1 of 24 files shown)


.changeset/silly-dolphins-try.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/vercel': major
+---
+
+Vercel adapter now defaults to a function per route
+
+With this change, in `@astrojs/vercel/serverless`, now each route is split into its own function. By doing this the size of each function is reduced and startup time is faster.
```
sarah11918, line 7:
```suggestion
With this change, `@astrojs/vercel/serverless` now splits each route into its own function. By doing this, the size of each function is reduced and startup time is faster.
```
Take a comma, leave a comma!

sarah11918 (approved):
Would just tighten up the wording a bit, but LGTM!

packages/integrations/vercel/test/fixtures/serverless-prerender/astro.config.mjs
```diff
@@ -4,7 +4,8 @@ import vercel from '@astrojs/vercel/serverless';
 export default defineConfig({
 	adapter: vercel({
 		// Pass some value to make sure it doesn't error out
-		includeFiles: ['included.js']
+		includeFiles: ['included.js'],
+		functionPerRoute: false,
```
matthewp, line 8: @ematipico I had to turn this off for this test. I'm not sure why, it could be a bug with functionPerRoute with prerendering.

packages/integrations/vercel/src/edge/adapter.ts
```diff
@@ -0,0 +1 @@
+import './throw.js';
```
FredKSchott, line 1: nit: you can remove both of these and just update the package.json imports mapping to point both to `throw.js`

FredKSchott (approved):
LGTM with 1 small comment

packages/integrations/vercel/test/fixtures/serverless-prerender/astro.config.mjs
ematipico: This should be fixed now

.changeset/silly-dolphins-try.md
```diff
@@ -0,0 +1,9 @@
+---
+'@astrojs/vercel': major
+---
+
+Vercel adapter now defaults to a function per route
```
ematipico, line 5:
```suggestion
Vercel adapter now defaults to `functionPerRoute`.
```

.changeset/silly-dolphins-try.md
```diff
@@ -0,0 +1,9 @@
...
+
+With this change, `@astrojs/vercel/serverless` now splits each route into its own function. By doing this, the size of each function is reduced and startup time is faster.
+
+You can disable this option, which will cause the code to be bundled into a single function, by setting `functionPerRoute` to false.
```
ematipico, line 9:
```suggestion
You can disable this option, which will cause the code to be bundled into a single function, by setting `functionPerRoute` to `false`.
```

ematipico (approved):

PR #8245, opened 2023-08-27
fix(deno): add start to options and extract to new file (close #8221)

### LINKED ISSUE #8221: @astrojs/deno - "start" option is not typed as part of the adapter Options type.
### Astro info

```block
N/A
```

N/A

### Describe the Bug

When using the adapter, I get a typescript error stating that the property does not exist on the Options type for the adapter.

image

However, when I look further at the type declarations, the Option type only has the port and hostname options defined. The start option still works perfectly so it hasn't been taken out of the adapter (thankfully!), but it must have been overlooked when adjusting the adapter's type declaration.

image

N/A

N/A

### Participation

florian-lefebvre:
## Changes

- Adds `start` to the `Options` type of the deno adapter
- Extract types to a dedicated file
- Closes #8221

## Testing

Not tested

## Docs

No docs updated

### DIFF
--- a/.changeset/olive-moles-tan.md
+++ b/.changeset/olive-moles-tan.md
@@ -2,4 +2,4 @@
 '@astrojs/deno': patch
 ---
 
-Fix Options types to include start. Types are extracted to a new file to avoid duplication.
+TypeScript users now get better suggestions when configuring the Deno adapter.
--- a/.changeset/olive-moles-tan.md
+++ b/.changeset/olive-moles-tan.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/deno': patch
+---
+
+Fix Options types to include start. Types are extracted to a new file to avoid duplication.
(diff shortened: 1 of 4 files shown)


.changeset/olive-moles-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/deno': patch
+---
+
+Fix Options types to include start. Types are extracted to a new file to avoid duplication.
```
lilnasy, line 5: This becomes part of the changelog and gets shown to astro users. It would be helpful if it mentioned that it affects the config file.

lilnasy (changes requested):
This looks great! Just the changeset could use a little rephrasing.

.changeset/olive-moles-tan.md
florian-lefebvre: What about: Add `start` argument to the adapter's config type.

lilnasy (approved):

lilnasy:
Thanks for working  on this! LGTM.

PR #8251, opened 2023-08-28
feat(cli): Add docs error link to errors in the CLI

Princesseuh:
## Changes

Adds a link to the docs's error reference in the CLI, much like in the overlay

image

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/errors/dev/utils.ts
+++ b/packages/astro/src/core/errors/dev/utils.ts
@@ -9,6 +9,7 @@ import { normalizePath } from 'vite';
 import type { SSRError } from '../../../@types/astro.js';
 import { removeLeadingForwardSlashWindows } from '../../path.js';
 import { AggregateError, type ErrorWithMetadata } from '../errors.js';
+import { AstroErrorData } from '../index.js';
 import { codeFrame } from '../printer.js';
 import { normalizeLF } from '../utils.js';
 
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -206,13 +207,29 @@ function cleanErrorStack(stack: string) {
...
+}
+
 /**
  * Render a subset of Markdown to HTML or a CLI output
  */
 export function renderErrorMarkdown(markdown: string, target: 'html' | 'cli') {
 	const linkRegex = /\[(.+)\]\((.+)\)/gm;
 	const boldRegex = /\*\*(.+)\*\*/gm;
-	const urlRegex = / (\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|]) /gim;
+	const urlRegex = /(\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|])/gim;
```
Princesseuh, line 232: Unrelated fix, the regex required a space on both ends of the link, which broke the underline in some cases, I'm not quite sure why I did it that way in the first place..

ematipico (approved):
It worked! :) Sorry for the unsound regression I caused, thank you for fixing it

PR #8253, opened 2023-08-28
Lazily initialize ResponseWithEncoding

matthewp:
## Changes

- Due to StackBlitz not yet supporting Node 18, we cannot extend `Response` until the polyfills have ran.

## Testing

Just manually, this only affects stackblitz.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -95,7 +95,6 @@ type ResponseParameters = ConstructorParameters<typeof Response>;
 export let ResponseWithEncoding: ReturnType<typeof initResponseWithEncoding>;
 // TODO Remove this after StackBlitz supports Node 18.
 let initResponseWithEncoding = () => {
-	// eslint-disable-next-line @typescript-eslint/no-shadow
 	class LocalResponseWithEncoding extends Response {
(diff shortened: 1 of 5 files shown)


packages/astro/src/core/endpoint/index.ts
```diff
@@ -91,27 +92,39 @@ export function createAPIContext({
...
-export class ResponseWithEncoding extends Response {
-	constructor(body: ResponseParameters[0], init: ResponseParameters[1], encoding?: BufferEncoding) {
-		// If a body string is given, try to encode it to preserve the behaviour as simple objects.
-		// We don't do the full handling as simple objects so users can control how headers are set instead.
-		if (typeof body === 'string') {
-			// In NodeJS, we can use Buffer.from which supports all BufferEncoding
-			if (typeof Buffer !== 'undefined' && Buffer.from) {
-				body = Buffer.from(body, encoding);
+export let ResponseWithEncoding: ReturnType<typeof initResponseWithEncoding>;
+let initResponseWithEncoding = () => {
```
Princesseuh, line 96: Mind adding a "TODO" or "HACK" or something comment up to so we know to remove this once Stackblitz supports Node 18, similar to the comment in `polyfill.ts`?

ematipico (approved):

PR #8258, opened 2023-08-28
 Allow fallback animations on html element

matthewp:
## Changes

- Allows animations on the `html` element by fixing the selector so it targets both children and the root itself.
- Simplifies the fallback logic. Previous was not working in Safari because `animationstart` did not fire before the setTimeout.
  - Now using https://developer.mozilla.org/en-US/docs/Web/API/Document/getAnimations which *I think* is safe to use. Waits on their `finished` event.
- Removed `opacity: 0` for the `none` animation as that causes the element to be invisible.

## Testing

https://github.com/withastro/astro/assets/361671/3e6575bb-1825-4a38-96c2-9fc5124d6af7

## Docs
N/A

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -220,23 +220,16 @@ const { fallback = 'animate' } = Astro.props as Props;
 		links.length && (await Promise.all(links));
 
 		if (fallback === 'animate') {
-			let isAnimating = false;
-			addEventListener('animationstart', () => (isAnimating = true), { once: true });
-
 			// Trigger the animations
 			document.documentElement.dataset.astroTransitionFallback = 'old';
+			const finished = Promise.all(document.getAnimations().map(a => a.finished));
 			const fallbackSwap = () => {
-				removeEventListener('animationend', fallbackSwap);
-				clearTimeout(timeout);
-				swap();
 				document.documentElement.dataset.astroTransitionFallback = 'new';
+				swap();
+				
 			};
-			// If there are any animations, want for the animationend event.
-			addEventListener('animationend', fallbackSwap, { once: true });
-			// If there are no animations, go ahead and swap on next tick
-			// This is necessary because we do not know if there are animations.
-			// The setTimeout is a fallback in case there are none.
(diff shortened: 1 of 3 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -220,23 +220,15 @@ const { fallback = 'animate' } = Astro.props as Props;
 		links.length && (await Promise.all(links));
 
 		if (fallback === 'animate') {
-			let isAnimating = false;
-			addEventListener('animationstart', () => (isAnimating = true), { once: true });
-
 			// Trigger the animations
 			document.documentElement.dataset.astroTransitionFallback = 'old';
+			const finished = Promise.all(document.getAnimations().map(a => a.finished));
```
natemoo-re, line 225: Nice! This is the API I was talking about. Support matrix looks good to me.

getAnimations support matrix

packages/astro/src/runtime/server/transition.ts
```diff
@@ -52,7 +52,7 @@ export function renderTransition(
 			}
 		}
 	} else if (animationName === 'none') {
-		sheet.addAnimationRaw('old', 'animation: none; opacity: 0; mix-blend-mode: normal;');
+		sheet.addAnimationRaw('old', 'animation: none; mix-blend-mode: normal;');
```
natemoo-re, line 55: Hmm, I might be wrong but I from my testing the `opacity: 0` was needed because of stacking when using view transitions (not the fallback). This should be handled by `mix-blend-mode: normal` though.

matthewp: Yeah i wanted to get your opinion on this. I thought it was for that reason. Isn't the reason to do this so you can see the new element which is behind it? But for `none` is that necessary as both elements are the same?

I removed this because it makes it so that in fallback mode the element is invisible until both animations are complete. We might need a mechanism for VT only styles.

matthewp: Changeed it with this: https://github.com/withastro/astro/pull/8258/commits/00dd8ee665cf9a80ba1a66ce22be323604ef3789

natemoo-re (approved):
Nice, looks great!

PR #8260, opened 2023-08-28
Allow custom `const x = boolean` exports to be read from plugins

ottomated:
## Changes

- Any const boolean exports on a page can now be accessed on `RouteData`. This allows plugins to define custom per-page options.

## Testing

- Adds a new `custom-page-options` test

## Docs

- This should probably be added to plugin docs

### DIFF
--- a/.changeset/brown-tips-drop.md
+++ b/.changeset/brown-tips-drop.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+Plugins can now access a page's boolean exports on RouteData['customOptions']
--- a/packages/astro/src/vite-plugin-scanner/scan.ts
+++ b/packages/astro/src/vite-plugin-scanner/scan.ts
@@ -10,6 +10,9 @@ type BuiltInExports = Exclude<keyof PageOptions, 'custom'>;
 // Quick scan to determine if code includes recognized export
 // False positives are not a problem, so be forgiving!
 function includesExport(code: string) {
+	for (const name of BOOLEAN_EXPORTS) {
+		if (code.includes(name)) return true;
+	}
 	return code.includes('export const');
 }
 
--- a/packages/astro/src/vite-plugin-scanner/scan.ts
+++ b/packages/astro/src/vite-plugin-scanner/scan.ts
@@ -5,14 +5,12 @@ import * as eslexer from 'es-module-lexer';
 import { AstroError, AstroErrorData } from '../core/errors/index.js';
 
 const BOOLEAN_EXPORTS = new Set(['prerender']);
+type BuiltInExports = Exclude<keyof PageOptions, 'custom'>;
 
 // Quick scan to determine if code includes recognized export
 // False positives are not a problem, so be forgiving!
 function includesExport(code: string) {
-	for (const name of BOOLEAN_EXPORTS) {
-		if (code.includes(name)) return true;
-	}
-	return false;
+	return code.includes('export const');
 }
 
 // Support quoted values to allow statically known `import.meta.env` variables to be used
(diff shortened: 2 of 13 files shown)


packages/astro/src/vite-plugin-scanner/scan.ts
```diff
@@ -51,33 +49,41 @@ export async function scan(
...
-			// For a given export, check the value of the first non-whitespace token.
-			// Basically extract the `true` from the statement `export const prerender = true`
-			const suffix = code.slice(endOfLocalName).trim().replace(/\=/, '').trim().split(/[;\n]/)[0];
-			if (prefix !== 'const' || !(isTruthy(suffix) || isFalsy(suffix))) {
+		const { n: name, ls: startOfLocalName, le: endOfLocalName } = _export;
+		if (startOfLocalName === -1 || endOfLocalName === -1) continue;
+		// For a given export, check the value of the local declaration
+		// Basically extract the `const` from the statement `export const prerender = true`
+		const indexOfExport = code.lastIndexOf('export', startOfLocalName);
+		const prefix = code.slice(indexOfExport + 6, startOfLocalName).trim();
```
ematipico, line 57: This `6` should be saved into a top-level `const` with a name that explains **why** 6 (maybe with a comment);

packages/astro/src/vite-plugin-scanner/scan.ts
```diff
@@ -51,33 +49,41 @@ export async function scan(
...
+
+		// `export const name = "value";
+		//                      ^      ^
+		// indexOfFirstNonWhitespace  endOfLineIndex
+		const indexOfFirstNonWhitespace = code.slice(endOfLocalName).search(/[^=\s]/) + endOfLocalName;
+		const endOfLineIndex = code.slice(indexOfFirstNonWhitespace).search(/[\n;]/) + indexOfFirstNonWhitespace;
```
ematipico, line 63: This check won't work for carriage returns, e.g. `\r`

Also, we can't assume there's a semicolon.

packages/astro/test/fixtures/custom-page-options/src/pages/index.astro
```diff
@@ -0,0 +1,15 @@
+---
+export const prerender = true;
+
+export const customTrue = true;
+export const customFalse = false;
```
ematipico, line 5: We should add more cases:
- no semicolons
- more spaces between `=` and the value;

.changeset/brown-tips-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Plugins can now access a page's boolean exports on RouteData['customOptions']
```
ematipico, line 5: I would call it `externalOptions`. It's more sound and communicates that these options **don't** from Astro. Or `pluginOptions`.

matthewp: I don't think we should differentiate. If we accept this feature I think it should include all boolean options, including `prerender`.

The reason is, if this field is just for non-Astro defined booleans, that would mean that if Astro adds any new booleans itself, that becomes a breaking change. However if they are all under `customOptions` then there is no breaking change.

packages/astro/src/vite-plugin-scanner/scan.ts
ottomated: - In the case of a carriage return, this will return `"value"\r` and the carriage return will be trimmed off on line 64
- It doesn't assume there's a semicolon – it splits off at either the first semicolon OR the first newline

ottomated: Added a test case for this and it passes

packages/astro/src/vite-plugin-scanner/scan.ts
```diff
@@ -51,33 +49,41 @@ export async function scan(
...
-				.pop()!
-				.trim()
-				.replace('prerender', '')
-				.trim();
-			// For a given export, check the value of the first non-whitespace token.
-			// Basically extract the `true` from the statement `export const prerender = true`
-			const suffix = code.slice(endOfLocalName).trim().replace(/\=/, '').trim().split(/[;\n]/)[0];
-			if (prefix !== 'const' || !(isTruthy(suffix) || isFalsy(suffix))) {
+		const { n: name, ls: startOfLocalName, le: endOfLocalName } = _export;
+		if (startOfLocalName === -1 || endOfLocalName === -1) continue;
```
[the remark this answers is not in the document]
ottomated: This only occurs if there is no local variable being exported - i.e. `export default 1`. That's impossible to do in a `.astro` file because it is already exporting the component as default.

.changeset/brown-tips-drop.md
matthewp: Another option here is that custom options must have some prefix we define. Similar to how HTML has the `data-` attributes. If they were required to be `customFoo`, for example, we could ensure that Astro never accidentally breaks an integration defined option.

packages/astro/src/vite-plugin-scanner/scan.ts
ottomated: added an inline comment, don't think it needs to be top-level for this case

.changeset/brown-tips-drop.md
ottomated: I like that idea but `custom` as a prefix feels a bit bulky to me

natemoo-re (changes requested):
I'm blocking this until we all get a chance to discuss as a team.

Really, really love this idea. I can think of four different problems this API could solve right off the bat—cache configuration, fragment/partial support, streaming control, and ISR. 

That's very exciting, but it's also a good sign that we shouldn't rush to merge this without going through [the roadmap process](https://github.com/withastro/roadmap). We have a process in place precisely because PRs like this can have wide-ranging implications and we owe it to our users to think through all the consequences of them before merging.

ottomated:
> I'm blocking this until we all get a chance to discuss as a team.

Sounds good! I don't have time to put together an RFC, so I'll back out and leave this code to be used as it can!

natemoo-re:
> Sounds good! I don't have time to put together an RFC, so I'll back out and leave this code to be used as it can!

No worries at all, that's totally understandable! Hoping we can pick this up ASAP to explore for 3.1 or soon after.

sevfurneaux:
This is great @ottomated! @lilnasy pointed me to this PR, as it'll be a precursor to potential page specific adaptor configuration (ISR, Vercel On-Demand ISR, Vercel Draft mode, Netlify Builders for singular pages) as outlined in [this proposal](https://github.com/withastro/roadmap/discussions/656).

One thing worth considering, is allowing for object properties to be exportable, and not just booleans. This would allow for more advanced configurations such as ISR expiry time and other adaptor specific configurations as mentioned in [the proposal](https://github.com/withastro/roadmap/discussions/656):

```js
export const config = {
  isr: {
    expiration: 60,
    bypassToken: BYPASS_TOKEN,
    allowQuery: ['search']
  }
};
```

Exportable objects would also allow for more granular cache configuration, fragment/partial support, streaming control as you mentioned @natemoo-re.

matthewp:
Still really like this idea. Hope it can get rolled into an RFC in the future. Closing as there's nothing actionable to be done right now.

PR #8264, opened 2023-08-29
Unmount framework components when islands are destroyed

natemoo-re:
## Changes

- Adds an `astro:unmount` event when islands are destroyed
- Improves view transitions lifecycle because navigating to a new page without an island now properly runs framework lifecycle hooks
- Updates all the framework integrations

## Testing

Tested manually but I would love to add some E2E tests

## Docs

We don't really document how to build a new renderer, but this would be a detail for renderers to handle.


packages/astro/components/ViewTransitions.astro
```diff
@@ -220,15 +222,23 @@ const { fallback = 'animate' } = Astro.props as Props;
 		links.length && (await Promise.all(links));
 
 		if (fallback === 'animate') {
+			let isAnimating = false;
+			addEventListener('animationstart', () => (isAnimating = true), { once: true });
+
 			// Trigger the animations
 			document.documentElement.dataset.astroTransitionFallback = 'old';
-			const finished = Promise.all(document.getAnimations().map(a => a.finished));
```
matthewp, line 225: did you remove this change on purpose?

natemoo-re: Whoops, no. No idea how that got removed!

PR #8268, opened 2023-08-29
Prevent ViewTransition script from being added by mistake

matthewp:
## Changes

- Removes `ViewTransitions` from `astro/components`.
- This was added very early (before the first release I think), but we now say to import from `astro:transitions`.
- Having this as part of astro/components brings its script if you are just using, for example, `<Code />`.

## Testing

- Added a test to the Code component test

## Docs

Bug fix.

### DIFF
--- a/.changeset/swift-taxis-sing.md
+++ b/.changeset/swift-taxis-sing.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent ViewTransition script from being added by mistake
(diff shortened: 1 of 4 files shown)


natemoo-re (commented):
There’s not a way to export this from `astro/components` and have a script? I guess not as easily, but I do like having our components be in one spot.

matthewp:
@natemoo-re Not right now, that's what `experimental.optimizeHoistedScript` aims to solve, but it has some issues right now.

natemoo-re:
Gotcha. Okay this works for me, then!

PR #8271, opened 2023-08-29
Fix video persistence regression

matthewp:
## Changes

- Previous swap order was causing videos to reset.

## Testing

- Manual. We have a test that should have failed before. After this is released I'll work on fixing that test.

## Docs

N/A, bug fix


natemoo-re (approved):
Manually checked with my local unmount test, still works!

LGTM.

PR #8276, opened 2023-08-29
add leading and trailing slash sanitization for params

FredKSchott:
## Changes

image

- Someone reported an issue where getStaticPaths were silently failing to match expected params (see the error above)
- This was due to them constructing their own route params, and incorrectly assuming that they needed a trailing slash in the slug in order to match a trailing slash in the URL.
- Since we handle this for you at the Astro layer, I can't think of a world where `/` would be expected behavior either leading or trailing in some route param.
- Therefore, I got nerd sniped into adding some simple route param sanitization

## Testing

- Test added

## Docs

- No docs needed


natemoo-re (approved):
LGTM!

PR #8280, opened 2023-08-30
chore: use biome to format JSON and JavaScript files

ematipico:
## Changes

With this PR, I want to **propose** the usage of `biome` for format JSON and JS files instead of prettier.

Also, it proposed to **remove** the `organize-imports-cli` and use `biome` to do that.

Here's the benchmarks run on my machine.

Current `format:ci` command:

```
Benchmark 1: pnpm format:ci
  Time (abs ≡):        42.289 s               [User: 55.376 s, System: 7.550 s]

```

Format + organize imports of `biome`:

```
Benchmark 1: pnpm format:source
  Time (abs ≡):        404.7 ms               [User: 742.7 ms, System: 170.9 ms]
```

`biome` + prettier (we need to keep prettier to format `yml`, `astro` and `md` files):

```
Benchmark 1: pnpm format:ci
  Time (abs ≡):         2.402 s               [User: 2.658 s, System: 1.071 s]
```

This comes to a price though, having **another** tool inside the code base. Although, I think this is worth the price:
- CI is going to be snappier and faster;
- we removed `organize-imports-cli`, which was taking **a lot** of time to finish;
- the action `[ci] format` is going to take way less time to apply the commit;

## Testing

The current CI should pass.

## Docs

N/A


ematipico:
> It would be fantastic if biome can support md, yml and Astro so we can get rid of prettier altogether :D

I could, yes u.u 

Unfortunately, we're not there yet.

biome.json
```diff
@@ -0,0 +1,38 @@
...
+		"formatter": {
+			"quoteStyle": "single",
+			"trailingComma": "es5"
+		}
+	},
+	"json": {
+		"parser": {
+			"allowComments": true
+		}
+	}
```
bluwy, line 37: Can we keep using spaces for JSON like before?

bluwy (commented):
Skimming through the changes they look fine to me. Have one comment below about tabs in json. Other than that it looks great! We should also follow-up and remove the `[ci] format` commit so they formatting is enforced in CI.

biome.json
ematipico: Still not :( it's a feature we are developing right now. Is it a deal breaker? If so, we could not format JSON files

bluwy: I would prefer keeping the formatting style as close to before, so I'd vote skipping formatting for now 😬

natemoo-re: The only reason is for our example templates, I think. Two spaces is generally the default.

natemoo-re:
I'm in favor of this! Could you add [ad0d2b0](https://github.com/withastro/astro/pull/8280/commits/ad0d2b083bdae6ab6e35d3aea8ede641f83faa57) to [`.git-blame-ignore-revs`](https://github.com/withastro/astro/blob/6258cd1c3a45a56385b2dc24f0408eb751365743/.git-blame-ignore-revs)?

natemoo-re:
Just a heads up that I also opened https://github.com/withastro/astro/pull/8420, but we can merge either one first

biome.json
bluwy: Package managers usually formats with two spaces when installing new dependencies that updates the `package.json` too.

JoshuaKGoldberg:
> we need to keep prettier to format yml, astro and md files

If you really want to be snazzy and fast, you could replace it with [dprint](https://dprint.dev) 😄 . And at that point you're already working at low level language speeds, so perhaps dprint would be a better Prettier replacement altogether?

A couple of interesting dprint references:

* https://github.com/microsoft/TypeScript/pull/54820
* https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65993

biome.json
ematipico: That's fixed now

ematipico:
> I'm in favor of this! Could you add [ad0d2b0](https://github.com/withastro/astro/pull/8280/commits/ad0d2b083bdae6ab6e35d3aea8ede641f83faa57) to [`.git-blame-ignore-revs`](https://github.com/withastro/astro/blob/6258cd1c3a45a56385b2dc24f0408eb751365743/.git-blame-ignore-revs)?

I wanted to do that, but should we wait for the PR to be merged? I thought that when we merge the PR, GitHub creates a new commit.

.git-blame-ignore-revs
```diff
@@ -4,3 +4,5 @@
 1335797903a57716e9a02b0ffd8ca636b3883c62
 # Manually format .astro files in example projects (#3862)
 59e8c71786fd1c154904b3fefa7d26d88f4d92d2
+# New formatting using Biome
+07613b96bbe7a74179895c8cbe5f364133edd9ee
```
bluwy, line 8: Yeah I think we should only do this after merging. The commit hash will change

packages/integrations/lit/server.js
```diff
@@ -1,6 +1,6 @@
-import './server-shim.js';
 import { LitElementRenderer } from '@lit-labs/ssr/lib/lit-element-renderer.js';
 import * as parse5 from 'parse5';
+import './server-shim.js';
```
bluwy, line 3: I think we need the `import './server-shim.js';` to be first as it's side-effectful and presumably affects how `@lit-labs/ssr/lib/lit-element-renderer.js` is loaded, so the order matters.

Princesseuh: They should honestly be in separate group if the order is important, any formatter would make this mistake

bluwy: Yeah putting a space between the imports would be fine by me too. I guess if biome wants to be more cautious about this automatically, plain `import '...'` ordering should be left as-is.

jacobthesheep:
Will biome also be added to the astro sites?

Princesseuh:
> Will biome also be added to the astro sites?

No, this is only for our own repo.

jacobthesheep:
Why is this closed?

Princesseuh:
> Why is this closed?

Since Biome can't format a fair amount of our files, it feels weird to have multiple formatters with different styles and configs

ematipico:
> Why is this closed?

It seems there isn't enough interest and I don't want to keep the PR open, if someone else wants too, they are welcome

PR #8289, opened 2023-08-30
config: set proper default config on framework-react example

Princesseuh:
## Changes

Config has been outdated for the past two years. Luckily, `astro add react` did set the proper config

## Testing

N/A

## Docs

N/A


natemoo-re (approved):
Thanks!

Issue #8299, opened 2023-08-30
Middleware not correctly execupted in SSR


ematipico:
### Astro info

```block
Astro 3.0
```

### What browser are you using?

N/A

### Describe the Bug

The middleware is not executed when bundling for SSR, especially adapters like cloudflare, vercel and netlify

### What's the expected result?

The middleware should work on production

### Link to Minimal Reproducible Example

N/A

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #8300, opened 2023-08-30
fix(ssr): inline middleware during the build, not rely on file system

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8299

Reverts part of changes of https://github.com/withastro/astro/pull/8200

We can't load the middleware using a file system path because, in adapters like Vercel, the bundled files are eventually moved. For other adapters like cloudflare, that is not permitted.

## Testing

Current tests should pass.

I also tested locally and checked that the middleware is correctly bundled in SSR.

## Docs

N/A


packages/astro/src/core/build/plugins/plugin-manifest.ts
```diff
@@ -237,9 +237,6 @@ function buildManifest(
 		// Set this to an empty string so that the runtime knows not to try and load this.
 		entryModules[BEFORE_HYDRATION_SCRIPT_ID] = '';
 	}
-	const isEdgeMiddleware =
-		// TODO: remove in Astro 4.0
-		settings.config.build.excludeMiddleware || settings.adapter?.adapterFeatures?.edgeMiddleware;
```
natemoo-re, line 242: Did this logic get dropped? Looks like the `TODO` did, at least.

matthewp (approved):

packages/astro/src/core/build/static-build.ts
```diff
@@ -123,14 +123,14 @@ export async function staticBuild(opts: StaticBuildOptions, internals: BuildInte
 		case settings.config.output === 'static': {
 			settings.timer.start('Static generate');
 			await generatePages(opts, internals);
-			await cleanServerOutput(opts);
+			// await cleanServerOutput(opts);
```
natemoo-re, line 126: Debugging?

packages/astro/src/core/build/static-build.ts
```diff
@@ -123,14 +123,14 @@ export async function staticBuild(opts: StaticBuildOptions, internals: BuildInte
...
 			settings.timer.end('Static generate');
 			return;
 		}
 		case isServerLikeOutput(settings.config): {
 			settings.timer.start('Server generate');
 			await generatePages(opts, internals);
-			await cleanStaticOutput(opts, internals);
+			// await cleanStaticOutput(opts, internals);
```
natemoo-re, line 133: Debugging?

packages/integrations/vercel/test/edge-middleware.test.js
```diff
@@ -19,17 +19,11 @@ describe('Vercel edge middleware', () => {
 	});
 
 	// TODO: The path here seems to be inconsistent?
-	it.skip('with edge handle file, should successfully build the middleware', async () => {
+	it.skip('without edge handle file, should successfully build the middleware', async () => {
```
natemoo-re, line 22: Still using `it.skip`, should it?

natemoo-re (commented):
This seems reasonable! Still see a few changes to be cleaned up.

packages/integrations/vercel/test/edge-middleware.test.js
ematipico: Yes. For some reasons that I don't understand, the path of the emitted files inside the fixtures is not predictable.

natemoo-re (approved):

PR #8319, opened 2023-08-31
fix(@astrojs/vercel): warn user when `functionPerRoute` is `true`

ematipico:
## Changes

We log a warning to the user, because the number of functions emitted could exceed the limit of someone's plan.

## Testing

Current tests should pass

## Docs

N/A

/cc @withastro/maintainers-docs please check the warning


ElianCodes (approved):
I think this is a great addition!

PR #8331, opened 2023-08-31
Prevent View Transition fallback from waiting on looping animations

matthewp:
## Changes

- Fixes an issue some have reported where navigation does not complete in Safari.
- Before trigger animations, get a list of animations that currently exist. Then trigger the animation and filter out the existing animations. What you're left with should be animations that are part of the view transition.

## Testing

https://github.com/withastro/astro/assets/361671/a3d87c05-d1b5-468d-850e-eb6c8e23d42f

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/late-foxes-juggle.md
+++ b/.changeset/late-foxes-juggle.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent View Transition fallback from waiting on looping animations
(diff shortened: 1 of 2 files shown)


matthewp:
!preview vt-fallback-filter

matthewp:
!preview vt-fallback-filter

natemoo-re (approved):
Looks great!

PR #8339, opened 2023-08-31
Respect the download attribute in links when using view transitions

martrapp:
## Changes

No need for explicit setting `data-astro-reload` attribute on links that have the `download` attribute.

(diff for packages/astro/e2e/view-transitions.test.js looks messy because I pushed the tests at the end of the file back into the View Transition group)

## Testing

e2e test added

## Docs

n.a. / fits implicit expectation of user


matthewp (approved):
great test!

natemoo-re (approved):

Issue #5699, opened 2022-12-29
"fetch failed" for `<Image>` component when transforming local images


septatrix:
### What version of `astro` are you using?

1.8.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### Describe the Bug

When using the astro image feature with a local image, the images do not get loaded an instead an error is printed by the development server:

```js
TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11118:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async loadRemoteImage (/home/septatrix/Documents/programming/cv/node_modules/@astrojs/image/dist/endpoint.js:13:17)
    at async Module.get (/home/septatrix/Documents/programming/cv/node_modules/@astrojs/image/dist/endpoint.js:29:19)
    at async call (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/core/endpoint/index.js:70:20)
    at async call (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/core/endpoint/dev/index.js:15:10)
    at async handleRoute (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/vite-plugin-astro-server/route.js:117:20)
    at async run (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/vite-plugin-astro-server/request.js:45:14)
    at async runWithErrorHandling (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/vite-plugin-astro-server/controller.js:65:5)
    at async handleRequest (file:///home/septatrix/Documents/programming/cv/node_modules/astro/dist/vite-plugin-astro-server/request.js:39:3) {
  cause: Error: connect ECONNREFUSED 127.0.0.1:3000
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1300:16) {
    errno: -111,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '127.0.0.1',
    port: 3000
  }
}
```

A hint by @mundry (https://github.com/withastro/astro/issues/5051#issuecomment-1367467896) seemed to hint at this being a problem with IPv4/IPv6. I have been able to confirm this though I am not sure how to resolve this issue:

[...]

PR #8343, opened 2023-09-01
Properly generating local images with string src

leomp12:
With Astro 3.x (`astro:assets`) `getImage({ src: '/img/something.png', width, height })` is being ignored.

## Changes

This makes `getImage()` and `<Image>` work again (generates the transformed images) with src string as local path.
Also fixes https://github.com/withastro/astro/issues/5699

## Testing

TBH I've tested it only in a non-minimal context porting my customized <Picture> component to Astro v3 (assets), sorry 😬 
I've tested `build` without the change and and no image was generated, after the changes all expected images was generated as usually done with Astro v2 and `@astrojs/images`.

If you think it's importante for this change, I can try to create a minimal reproduction later. Or you can just pick the idea, close the PR and commit a fix yourself with the desired tests, if you prefer.

## Docs

> No docs changes...

Love Astro btw , hoping to use in production really soon ❤️

### DIFF
--- a/packages/astro/src/assets/internal.ts
+++ b/packages/astro/src/assets/internal.ts
@@ -36,7 +36,7 @@ export function isRemoteAllowed(
 		remotePatterns = [],
 	}: Partial<Pick<AstroConfig['image'], 'domains' | 'remotePatterns'>>
 ): boolean {
-	if (!isRemotePath(src)) return false;
+	if (!isRemotePath(src)) return true;
 
 	const url = new URL(src);
 	return (
--- a/.changeset/witty-cooks-exercise.md
+++ b/.changeset/witty-cooks-exercise.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix `astro:assets` to properly transforming local images with string src
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -1,5 +1,6 @@
 import fs, { readFileSync } from 'node:fs';
 import { basename, join } from 'node:path/posix';
+import { isRemotePath } from '@astrojs/internal-helpers/path';
 import type { BuildPipeline } from '../../core/build/buildPipeline';
 import { prependForwardSlash } from '../../core/path.js';
 import { isServerLikeOutput } from '../../prerender/utils.js';
@@ -52,7 +53,8 @@ export async function generateImage(
 		clientRoot = config.outDir;
 	}
 
-	const isLocalImage = isESMImportedImage(options.src);
+	const isESMImported = isESMImportedImage(options.src);
+	const isLocalImage = isESMImported || !isRemotePath(options.src);
 
 	const finalFileURL = new URL('.' + filepath, clientRoot);
 	const finalFolderURL = new URL('./', finalFileURL);
@@ -91,7 +93,7 @@ export async function generateImage(
 	}
 
 	// The original filepath or URL from the image transform
-	const originalImagePath = isLocalImage
+	const originalImagePath = isESMImported
 		? (options.src as ImageMetadata).src
 		: (options.src as string);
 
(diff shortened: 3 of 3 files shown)


Princesseuh:
This limitation is intentional, supporting this is not possible in SSR.

leomp12:
Hey @Princesseuh , thanks for the answer!
I was looking for open issues/PRs with the same problem to prevent opening this PR unnecessarily, then I misunderstood the mentioned issue, sorry about it.

About the impossibility with SSR, I actually use SSR with Node.js adapter, dev server is running perfectly after the changes, with the images transformed.

I'm currently deploying without the image transformation stuff for performance reasons, I pregenerate the images instead ([context](https://github.com/ecomplus/cloud-commerce/tree/main/packages/storefront/src/images) [2](https://github.com/ecomplus/cloud-commerce/blob/main/packages/storefront/astro.config.mjs#L142-L147)  [3](https://github.com/ecomplus/cloud-commerce/blob/main/packages/storefront/scripts/build-prod.sh) [4](https://github.com/ecomplus/cloud-commerce/blob/main/packages/storefront/astro.config.mjs#L176-L180), my build step actually run `astro build` twice, last one with SSR adapter), but I got to use `<Picture src="/{string}" ... />` with SSR runtime transformation (testes both with sharp and squoosh btw), used to work with `@astrojs/image`, how can it be impossible now? 🤔 
I'm deploying to Firebase with Node.js 18, it just make required to deploy with the _publicDir_ included to function source of course. Maybe with a new option config option?

In my case this feature is a must (will stick with Astro v2 if no possible here), `astro:assets` is useless to me without this behavior, and IMO it's really useful to anyone with git-based CMS (Tina CMS, Decap CMS, Keystatic...).
Also, there are several mentions in the documentation using `<Image>` this way.

---

By the way, in addition to server runtime image transformations being slow (bad LCP with not cached images), just the fact of adding `sharp` to function dependencies was increasing cold start considerably, and thus increasing TTFB of not CDN-cached routes. So I'm also a bit worried about sharp as astro pkg dependency (I'll probably move install `astro` to dev deps only and manually install dependencies required for SSR 🙃 ).

Princesseuh:
Ah, I see where we got confused. You mean images in `public`, not local images. What doesn't work in SSR is supporting, ex: `<Image src={"./relative_image.png"} alt="" />` since `./relative_image.png` doesn't exist anymore in SSR.

Honestly, we kinda made the choice on purpose not to support optimization for images in `public` both for technical and philosophical reasons. While it did work most of the time in `@astrojs/image`, it also led to a fair amount of users being confused when it didn't work like they would expect (you can very easily get Vite warnings and stuff if you misuse it). Philosophically, the folder is intended for `static` assets only, so not supporting optimization for those makes more sense.

However, since you can support this with remote optimization. I think I could be convinced, though I wouldn't recommend people to do it. We'd have to make sure it works with the same code in SSR too. I'll discuss it with the rest of the team, thank you for contributing!



> By the way, in addition to server runtime image transformations being slow (bad LCP with not cached images), just the fact of adding sharp to function dependencies was increasing cold start considerably, and thus increasing TTFB of not CDN-cached routes. So I'm also a bit worried about sharp as astro pkg dependency (I'll probably move astro to dev deps only and manually install dependencies required for SSR 🙃 ).

Yes, runtime optimization is terrible, especially in serverless. Using the Node SSR adapter you can cache at the server level, so that works better.

packages/astro/src/assets/internal.ts
```diff
@@ -36,7 +36,7 @@ export function isRemoteAllowed(
 		remotePatterns = [],
 	}: Partial<Pick<AstroConfig['image'], 'domains' | 'remotePatterns'>>
 ): boolean {
-	if (!isRemotePath(src)) return false;
+	if (!isRemotePath(src)) return true;
```
leomp12, line 39: @Princesseuh if you prefer, we can also ensure only images on public dir are allowed here by checking src as absolute path:

```suggestion
	if (!isRemotePath(src) && src.charAt(0) === '/') return true;
```

leomp12:
I was able to solve this by manually setting src as `ImageMetadata` with the dynamic filepath, and copying the image files manually to the output folder.
Closing here...

PR #8345, opened 2023-09-01
Refactor HTML Meta and DOCTYPE Declarations

cevdetarda:
## Changes

- Removed unnecessary `<!DOCTYPE html>` from `.astro` files for improved efficiency.
- Standardized naming conventions by changing `<!doctype html>` to `<!DOCTYPE html>`.
- Ensured consistent character encoding by changing `<meta charset="UTF-8">` to `<meta charset="utf-8">`.
- Removed the `<meta http-equiv="X-UA-Compatible">` tag, no longer needed for modern browsers.
- Simplified viewport meta tags by setting `<meta name="viewport" content="width=device-width">`.
- Replaced Twitter Card tags (`twitter:url`, `twitter:title`, `twitter:description`, `twitter:image`) with Open Graph (`og:*`) tags for better compatibility and added `prefix="og: http://ogp.me/ns#"` support packages.

References:
- https://death-to-ie11.com/
- https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag
- https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started##twitter-cards-and-open-graph

## Testing

This change was tested by thoroughly reviewing the modified code to ensure correctness and by running relevant test cases. It has undergone manual testing on various browsers and devices to verify compatibility.

## Docs

No need for docs change, it's a basic change and has no relationship with the documentation.

### DIFF
--- a/examples/basics/src/layouts/Layout.astro
+++ b/examples/basics/src/layouts/Layout.astro
@@ -6,15 +6,14 @@ interface Props {
 const { title } = Astro.props;
 ---
 
-<!DOCTYPE html>
 <html lang="en">
 	<head>
-		<meta charset="UTF-8" />
-		<meta name="description" content="Astro description">
+		<meta charset="utf-8" />
 		<meta name="viewport" content="width=device-width" />
 		<link rel="icon" type="image/svg+xml" href="/favicon.svg" />
 		<meta name="generator" content={Astro.generator} />
 		<title>{title}</title>
+		<meta name="description" content="Astro description">
 	</head>
 	<body>
 		<slot />
--- a/examples/blog/src/components/BaseHead.astro
+++ b/examples/blog/src/components/BaseHead.astro
@@ -9,14 +9,12 @@ interface Props {
 	image?: string;
 }
 
-const canonicalURL = new URL(Astro.url.pathname, Astro.site);
-
 const { title, description, image = '/blog-placeholder-1.jpg' } = Astro.props;
 ---
 
 <!-- Global Metadata -->
 <meta charset="utf-8" />
-<meta name="viewport" content="width=device-width,initial-scale=1" />
+<meta name="viewport" content="width=device-width" />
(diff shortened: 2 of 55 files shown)


jacobthesheep:
Use `<meta
      name="viewport"
      content="width=device-width,initial-scale=1,viewport-fit=cover"
    />` as it ensures better compatibility with modern devices

cevdetarda:
> `<meta name="viewport" content="width=device-width,initial-scale=1,viewport-fit=cover" />` would ensure better compatibility with modern devices.

Thank you for the suggestion! While <meta name="viewport" content="width=device-width,initial-scale=1,viewport-fit=cover" /> offers additional control, our current approach of <meta name="viewport" content="width=device-width"> prioritizes simplicity and broad compatibility with modern devices and browsers. It provides a standard configuration that suits most use cases. If there's a strong rationale for the added complexity and fine-tuning options, we're open to discussing it further.

- https://stackoverflow.com/questions/60154236/is-viewport-fit-cover-no-longer-working-on-the-ios-safari
- https://modelessdesign.com/backdrop/283
- https://stackoverflow.com/questions/69695538/viewport-fit-cover-not-extending-viewport-into-status-bar-area-around-notch-on

fflaten:
> - Removed unnecessary <!DOCTYPE html> from .astro files for improved efficiency.

Could you elaborate on the efficiency gains? Isn't doctype declaration considered a good practice to optimize performance? 

As for `viewport-fit=cover` I'd avoid it for fixtures and basic examples/templates without styles for simplicity.

For the more styled examples like `blog` it could make sense to include it **and** safe-area padding in the css to help users get started. Websites with non-solid backgrounds looks bad without it on iOS. Just my two cents 🙂

matthewp:
Looks like these are quite a bit outdated. If you'd like to submit this change again, I would recommend discussing on discord in the `#docs` channel to make sure it's a change they are on board with.

Issue #8312, opened 2023-08-31
Astro v3 pnpm build fails, symlink


pijusz:
### Astro info

```block
yarn run v1.22.19
$ astro info

Astro version            v3.0.3
Package manager          pnpm
Platform                 win32
Architecture             x64
Adapter                  @astrojs/vercel/serverless
Integrations             @astrojs/prefetch, @astrojs/react, @astrojs/tailwind, astro-robots-txt
```

### What browser are you using?

Chrome

### Describe the Bug

When building Astro v3 on Windows NVM and on Vercel:
With pnpm packages and node v18.17.1:
``` error   EEXIST: file already exists, symlink '.pnpm/react@18.2.0/node_modules/react' -> '/vercel/path0/.vercel/output/functions/entry._slug_.astro.func/node_modules/react'
Error: EEXIST: file already exists, symlink '.pnpm/react@18.2.0/node_modules/react' -> '/vercel/path0/.vercel/output/functions/entry._slug_.astro.func/node_modules/react'
 ELIFECYCLE  Command failed with exit code 1.
Error: Command "pnpm run build" exited with 1
```
with yarn and node v18.17.1 - works

### What's the expected result?

pnpm is able to build Astro v3 with vercel serverless on node 18+

Stackblitz example seems to work

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-gmnscw

### Participation

- [ ] I am willing to submit a pull request for this issue.

florian-lefebvre:
I also have the issue locally with the following env:
- OS: Win 11
- Node: 18.14.1 (nvm)
- PNPM: 8.7.0

newtoallofthis123:
I was facing a similar issue from `pnpm`, the issue was solved by switching to `npm`

raulfdm:
Can someone add a repository to reproduce it? I'm trying to convert to a minimal and scoped project but no success.

pijusz:
Someone else managed to reproduce, but it works for me. Might need to have static routes with react components:
"but in my case specifically, I got the slug error at an updates page of mine, which is a react component "
Other one, which works for me though
![image](https://github.com/withastro/astro/assets/20190386/eaee76c9-4c9f-4b66-9ca3-bb3f25d2d8b5)


Will try to make reproduction tomorrow if none else comes in.

lilnasy:
Here's that screenshot in better quality.

https://cdn.discordapp.com/attachments/1146475239967047680/1146729067962433617/image.png

[...]

PR #8348, opened 2023-09-01
fix(@astrojs/vercel): slowness and symbolic link

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8312

This PR addresses two issues:
- **slowness of the build**: our adapter, when bundling a function, needs to carry over the dependencies that are inlined. Unfortunately, some package managers rely on symbolic links, and we need to trace them. This means that most of the time, the base path to start looking for files is the root folder (`/` or `C:\`). On big projects (monorepos), this could cause a lot of overhead. Fortunately, `@vercel/nft` allows us to pass a `cache` object to reuse when bundling the functions. **The first function to bundle will always be slow**, but the next ones are way faster. 
- **symbolic link issues**: I suppose this is specific to `pnpm`. Essentially, our adapter carries over the symbolic link of our inlined (external, not bundled) dependencies. After the first run, we have the source of the symbolic link in the `pnpm` store, so we don't need to create a new one again. Hopefully, this should fix the issue. 

## Testing

Create a preview release and ask users to try it, and make sure to check if the deployed function works.

## Docs

N/A

### DIFF
--- a/packages/integrations/vercel/src/lib/fs.ts
+++ b/packages/integrations/vercel/src/lib/fs.ts
@@ -1,5 +1,6 @@
 import type { PathLike } from 'node:fs';
 import * as fs from 'node:fs/promises';
+import { existsSync } from 'node:fs';
 import nodePath from 'node:path';
 import { fileURLToPath } from 'node:url';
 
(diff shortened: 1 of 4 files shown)


ematipico:
!preview vercel-symbolic-link

packages/integrations/vercel/src/lib/fs.ts
```diff
@@ -74,11 +75,13 @@ export async function copyFilesToFunction(
...
 		if (isSymlink) {
 			const realdest = fileURLToPath(new URL(nodePath.relative(commonAncestor, realpath), outDir));
-			await fs.symlink(
-				nodePath.relative(fileURLToPath(new URL('.', dest)), realdest),
-				dest,
-				isDir ? 'dir' : 'file'
-			);
+			const target = nodePath.relative(fileURLToPath(new URL('.', dest)), realdest);
+			// NOTE: when building function per route, dependencies are linked at the first run, then there's no need anymore to do that once more.
+			// So we check if the destination already exists. If it does, move on.
```
Princesseuh, line 80: But doesn't every function get its own `node_modules`?

ematipico: Yes, but inside a function we have symbolic links: 

Screenshot 2023-09-01 at 13 16 08


This is `pnpm` doing its own thing I guess

ematipico: Full disclosure, it's still something a bit foggy to me, especially that part of the code. It's possible my comment is wrong

PR #8351, opened 2023-09-01
Allow vite to refer to inlined CSS

### LINKED ISSUE #8329: Inlined stylesheets are referenced by vite's preload helper
### Astro info

```block
Astro version            v3.0.3
Package manager          npm
Platform                 linux
Architecture             x64
Integrations             auto-import, @astrojs/tailwind, @astrojs/svelte, @astrojs/mdx, @astrojs/preact
```

Chrome

### Describe the Bug

image

There is supposed to be a rendered image in that box.

For some reason, vite attains references to css chunks and tries to manually load them using javascript. Some of these stylesheets were inlined and removed from the bundle. The css loads as expected but [javascript code added by vite throws](https://github.com/vitejs/vite/blob/v4.4.9/packages/vite/src/node/plugins/importAnalysisBuild.ts#L133), which prevents user code from executing.

Editing the vite code to not throw is sufficient to prevent this, but hopefully there's a better angle. 

image

https://github.com/threlte/threlte/tree/1dfc901/apps/docs

### Participation

lilnasy:
## Changes
Closes #8329 
Inlined CSS chunks are not deleted anymore, because when using dynamic imports with `import.meta.glob()`, vite tries to preload them.

## Testing
Added a fixture.

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/chatty-walls-happen.md
+++ b/.changeset/chatty-walls-happen.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixed a case where dynamic imports tried to preload inlined stylesheets.
--- a/.changeset/fair-countries-admire.md
+++ b/.changeset/fair-countries-admire.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/svelte': patch
+---
+
+Removed vite warnings.
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -13,7 +13,7 @@ module.exports = {
   rules: {
     // These off/configured-differently-by-default rules fit well for us
     '@typescript-eslint/array-type': ['error', { default: 'array-simple' }],
-    '@typescript-eslint/no-unused-vars': ['error', { ignoreRestSiblings: true }],
+    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: "^_", ignoreRestSiblings: true }],
     'no-only-tests/no-only-tests': 'error',
     '@typescript-eslint/no-shadow': ['error'],
     'no-console': 'warn',
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -121,8 +121,6 @@ export class App {
 		}
 		return pathname;
 	}
-	// Disable no-unused-vars to avoid breaking signature change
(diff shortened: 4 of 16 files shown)


lilnasy:
I'm hoping there's a better way.

Ideally vite would not attain references to files we might later delete. Alternatively, it could not throw an error when preloading fails as it is not critical.

I will try to look around their codebase to find something useful. If I don't, I'll mark this ready to review on monday.

lilnasy:
I haven't had time to look into fixing it on the vite side, but the compromise is small - small css files included in the server bundle. 

And the unnecessary requests vite makes is a minor annoyance. It does not affect performance.

bluwy (approved):
I'm ok with this fix for now. If you'd like to look further, I think there's maybe a couple ways for a direct fix:

1. Use [`modulePreload.resolveDependencies`](https://vitejs.dev/config/build-options.html#build-modulepreload) to filter out the removed CSS. A bit tricky to bridge this config option to `generateBundle` in plugin-css.
2. Filter out the deps [here](https://github.com/vitejs/vite/blob/db1f3a377658b512fd0507d95c4f19edc4817f59/packages/vite/src/node/plugins/importAnalysisBuild.ts#L554-L563) that does not exist in the `bundle`. Could be a better fix.

lilnasy:
@bluwy
I looked into 1, but css is special-cased and preloaded anyway - `build.modulePreload` only applies to scripts.

I will try to look into 2 later this week, but I am skeptic because vite's built-in hooks run before app hooks IIRC, so I expect it to have already created the preload script by the time.

Either way, not blocking. Will merge after node 18 tests pass.

bluwy:
@lilnasy thanks for looking into it.

> I will try to look into 2 later this week, but I am skeptic because vite's built-in hooks run before app hooks IIRC, so I expect it to have already created the preload script by the time.

Vite handles that through plugins, so as long as we order ours right before Vite's, I think it could work. You can check out the plugin ordering at [here](https://github.com/vitejs/vite/blob/2b80089491eb028468c54b3f1e7cf375c94a4432/packages/vite/src/node/plugins/index.ts#L42-L103) and [here](https://github.com/vitejs/vite/blob/2b80089491eb028468c54b3f1e7cf375c94a4432/packages/vite/src/node/build.ts#L436-L464). Not urgent of course, feel free to look if you're interested!

PR #8353, opened 2023-09-01
fix: no asset plugin w/ img is imported with query

elevatebart:
## Changes

- What does this change?

### DIFF
--- a/packages/astro/src/assets/vite-plugin-assets.ts
+++ b/packages/astro/src/assets/vite-plugin-assets.ts
@@ -14,9 +14,6 @@ import { hashTransform, propsToFilename } from './utils/transformToPath.js';
 
 const resolvedVirtualModuleId = '\0' + VIRTUAL_MODULE_ID;
 
-const rawRE = /(?:\?|&)raw(?:&|$)/;
-const urlRE = /(\?|&)url(?:&|$)/;
-
 export default function assets({
 	settings,
 	mode,
@@ -119,8 +116,9 @@ export default function assets({
 				resolvedConfig = viteConfig;
 			},
 			async load(id) {
-				// If our import has the `?raw` or `?url` Vite query params, we'll let Vite handle it
-				if (rawRE.test(id) || urlRE.test(id)) {
+				// If our import has any query params, we'll let Vite handle it
+				// See https://github.com/withastro/astro/issues/8333
+				if (id.includes('?')) {
 					return;
 				}
 


packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -119,8 +116,9 @@ export default function assets({
 				resolvedConfig = viteConfig;
 			},
 			async load(id) {
-				// If our import has the `?raw` or `?url` Vite query params, we'll let Vite handle it
-				if (rawRE.test(id) || urlRE.test(id)) {
+				// If our import has any query params, we'll let Vite handle it
+				// See https://github.com/withastro/astro/issues/8333
+				if (id.includes('?')) {
```
Princesseuh, line 121: Linux and macOS allow `?` in file names, what you might want to do is compare the original id with the cleaned one using `removeQueryString`

elevatebart: Will do. Thanks.

I was curious about how this function worked since the `?` could be in the path. 
Here is the code of the `removeQueryString` function.
It is not taking any steps to check if `?` is part of the fileName. 

```ts
export function removeQueryString(path: string) {
	const index = path.lastIndexOf('?');
	return index > 0 ? path.substring(0, index) : path;
}
```

elevatebart: I still have done the change even for no other reason than clarity.

Princesseuh: That function takes the last index of a `?`, so if there's multiple (ex: `file?name.png?query=something`), it'll return from the beginning to the last index, skipping over `?` in the file.

Princesseuh: 🙈 Well, that's something else to fix one day. But at least since we're all using the same function, we'll only have to fix it once

Princesseuh (approved):

elevatebart:
Is something missing from this PR? 
Is it ready to merge?
Should I try to understand why the Cloudflare tests are failing (wrangler timeout)?

thank youu

Princesseuh:
> Is something missing from this PR? 
> Is it ready to merge?
> Should I try to understand why the Cloudflare tests are failing (wrangler timeout)?
> 
> thank youu

Nothing wrong, I'll check it out tomorrow. Sorry for the delay!

PR #8354, opened 2023-09-01
fix(vercel): Don't output an error message for failing to resolve sharp

Princesseuh:
## Changes

What the title says. Sharp is always external, so it'll always output a warning but this is not a problem

## Testing

Tested it manually, it didn't output a warning, sir

## Docs

N/A


ematipico (approved):

ematipico (approved):
Changeset?

PR #8357, opened 2023-09-01
feat: add progress to asset generation

itsmatteomanf:
## Changes

Added counter to show progress for assets image generation.
Fixed small unit of measurement error.

## Testing

No tests are necessary, the change is minimal and just to logging.

## Docs

No documentation changes are needed, adds info to logging.

### DIFF
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -155,12 +155,12 @@ export async function generateImage(
 	};
 }
 
-export function getStaticImageList(): Iterable<
-	[string, { path: string; options: ImageTransform }]
+export function getStaticImageList(): Map<
+	string, { path: string; options: ImageTransform }
 > {
 	if (!globalThis?.astroAsset?.staticImages) {
-		return [];
+		return new Map();
 	}
 
-	return globalThis.astroAsset.staticImages?.entries();
+	return globalThis.astroAsset.staticImages;
 }
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -196,12 +196,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		}
 	}
 
-	const staticImageList = Array.from(getStaticImageList())
+	const staticImageList = getStaticImageList()
 
-	if (staticImageList.length) logger.info(null, `\n${bgGreen(black(` generating optimized images `))}`); let count = 0;
-	for (const imageData of staticImageList) {
+	if (staticImageList.size) logger.info(null, `\n${bgGreen(black(` generating optimized images `))}`); let count = 0;
+	for (const imageData of staticImageList.entries()) {
 		count++
-		await generateImage(pipeline, imageData[1].options, imageData[1].path, count, staticImageList.length);
+		await generateImage(pipeline, imageData[1].options, imageData[1].path, count, staticImageList.size);
 	}
 
 	delete globalThis?.astroAsset?.addStaticImage;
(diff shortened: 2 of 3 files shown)


packages/astro/src/core/build/generate.ts
```diff
@@ -196,9 +196,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		}
 	}
 
-	logger.info(null, `\n${bgGreen(black(` generating optimized images `))}`);
-	for (const imageData of getStaticImageList()) {
-		await generateImage(pipeline, imageData[1].options, imageData[1].path);
+	const staticImageList = Array.from(getStaticImageList())
+
+	if (staticImageList.length) logger.info(null, `\n${bgGreen(black(` generating optimized images `))}`); let count = 0;
+	for (const imageData of staticImageList) {
```
Princesseuh, line 202: You can get the index directly in the `for of` here, no need for a separate variable

packages/astro/src/core/build/generate.ts
```diff
@@ -196,9 +196,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		}
 	}
 
-	logger.info(null, `\n${bgGreen(black(` generating optimized images `))}`);
-	for (const imageData of getStaticImageList()) {
-		await generateImage(pipeline, imageData[1].options, imageData[1].path);
+	const staticImageList = Array.from(getStaticImageList())
```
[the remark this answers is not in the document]
itsmatteomanf: Oh, ok. I didn't want to actually change the return value, to keep the changes to a minimum.

Will do!

packages/astro/src/core/build/generate.ts
itsmatteomanf: You can? I am not aware, and haven't found, a way to get the index from a `for... of`.

I could call the `.entries()` method and destructive the array I get, but I get a separate variable reassigned each time.

packages/astro/src/core/build/generate.ts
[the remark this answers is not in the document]
Princesseuh: Yeah, no worries here since it's only called here. The biggest reason it's in a function is to "sandbox" the usage of the `globalThis.astroAsset` because we're hoping to change the implementation eventually

packages/astro/src/core/build/generate.ts
Princesseuh: Oh no, you're right. I forgot how Map works, derp

itsmatteomanf: Yeah, it's kind of annoying, but that's how it is. I'm gonna see what's the best way forward, changing the structure.

Diamondbrandblue:
Log improvements yes

packages/astro/src/assets/build/generate.ts
```diff
@@ -155,12 +155,12 @@ export async function generateImage(
...
 }
 
-export function getStaticImageList(): Iterable<
-	[string, { path: string; options: ImageTransform }]
+export function getStaticImageList(): Map<
+	string, { path: string; options: ImageTransform }
 > {
 	if (!globalThis?.astroAsset?.staticImages) {
-		return [];
+		return new Map();
```
itsmatteomanf, line 162: Returning a map forces me to create one here, which I presume isn't the worst, it's just once for an empty map.
The rest reduces a lot of overhead.

Princesseuh: That's fine with me!

Princesseuh (approved):
Thank you very much, this looks great!

itsmatteomanf:
My pleasure! :)

PR #8365, opened 2023-09-01
fix hydration in solid renderer

ryansolid:
## Changes

I was playing on stream and noticed top-level fragments weren't working properly in Solid Astro integration. Duplicated elements and other weirdness. On further exploration I realized this is because the renderer always creates new DOM nodes by cloning so it can't match them.

I then noticed that the hydration check wasn't working because it was being done before any hydrate calls were being done top-level. This meant it was always false so we were always creating the nodes as if client rendered as well.

Also hierarchical children weren't working as we'd consume all descendant slots and they'd override the parent. I'm hoping this treewalker is fine for performance but if you have ideas open to that.

The solution might seem a bit odd but given visibility changes (ie.. islands controlling when to render slots) I think that creating vs consuming isn't an either/or like it was before but per slot. We try to fill from the DOM first and failing that we fill out the templates. I suppose we could make it do that lazily for even more performance but them prop handling needs some more work.

## Testing

Honestly I only tested this in my Hackernews project which let me test recursion and visibility show and hide. But I didn't test client only or other modes so there could be gaps.

## Docs

This shouldn't require docs changes. The API is the same. It just fixes a bunch of random bugs and should improve performance for the shown element case.


natemoo-re (commented):
Thanks @ryansolid! I'll give this a more thorough test run, but the code looks totally reasonable.

🙌 Appreciate you jumping on this

natemoo-re (approved):
Thank you! No notes, this looks great. Merging as-is.

PR #8366, opened 2023-09-01
Ensure `chunkFileNames` do not include invalid characters

natemoo-re:
## Changes

- We were getting reports of the `@astrojs/netlify` adapter being totally broken for `server` and `hybrid` deployments.
- Turns out Astro would sometimes emit chunks with invalid characters `@_@` which Netlify choked on
- The fix is to ensure `chunkFileNames` sanitizes the filename first
- Also changes our chunk asset filenames from something like `dist/server/chunks/index@_@astro.[hash].[ext]` to something like `dist/server/chunks/index_hash.[ext]`
- Notable change is `[name].[hash].[ext]` to `[name]_[hash].[ext]`
- Changeset added!

## Testing

Tested manually with community reproductions. It works!

## Docs

Bug fix only, will alert folks!

### DIFF
--- a/.changeset/eighty-laws-brake.md
+++ b/.changeset/eighty-laws-brake.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Update `chunkFileNames` to avoid emitting invalid characters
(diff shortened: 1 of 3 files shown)


natemoo-re:
!preview sanitize-filenames

natemoo-re:
!preview sanitize-filenames

natemoo-re:
Tests will need to be updated to match this expected format `[name]_[hash]` instead of `[name].[hash]`

ematipico (approved):

FredKSchott (approved):

FredKSchott (approved):
RE-APPROVED

Issue #8364, opened 2023-09-01
Import Astro files in typescript (`.ts`) files does not work anymore (3.0.5)


leifmarcus:
### Astro Info

```block
Astro                    v3.0.5
Node                     v18.16.0
System                   macOS (arm64)
Package Manager          pnpm/npm
Output                   static
Adapter                  none
Integrations             @astrojs/partytown
                         @astrojs/vue
```

### Describe the Bug

When importing an astro file inside a `.ts` file and checking the types with `tsc`, there will be always an error message, that the types of the module `.astro` cannot be resolved.

In Astro 2.x, we implemented a custom module declaration, which did work as expected:
```ts
declare module '*.astro' {
  import type { AstroComponentFactory } from 'astro/runtime/server';
  const component: AstroComponentFactory;
  export default component;
}
```

After moving to Astro `3.0.5`, this does not work anymore when using `tsc --noEmit`.  When having the module declared, there is a typescript error in `node_modules/astro/components/index.ts` (Error: `Unused '@ts-expect-error' directive.`)

We also tested the `@astrojs/ts-plugin` plugin, but this does not change anything.

### What's the expected result?

- When importing astro files in `*.ts` files, typescript should be able to resolve the typings of the astro file.
- It should be possible to either typecheck typescript files using `astro check` or there should be a way, to run `tsc --noEmit` without failing when importing `.astro` files.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-eyk8i5

### Participation

- [ ] I am willing to submit a pull request for this issue.

vixalien:
So the only way to fix this is to use `// @ts-ignore `? Also, the linked PR just adds `ts-ignore` to files inside the astro repo, and does not actually fix this issue.

PR #8367, opened 2023-09-01
fix: tsc complaining about astro imports again

Princesseuh:
## Changes

Fix https://github.com/withastro/astro/issues/8364

## Testing

Tested manually

## Docs

N/A


natemoo-re (approved):
Womp wahhhh. Thanks for the fix!

leifmarcus:
Thanks for the fix!!!

PR #8382, opened 2023-09-03
Remove AstroError if content directory is empty

### LINKED ISSUE #8336: Allow for empty content collections
### Astro Info

```block
Astro version            v2.9.6
Package manager          npm
Platform                 darwin
Architecture             arm64
Integrations             @astrojs/react, @astrojs/tailwind, auto-import, @astrojs/mdx
```

_No response_

### Describe the Bug

I'm well aware some might see this as a new feature, but I have to disagree. Content collections should be able to handle a "empty state" – meaning you should be able to use a content collection which has no items. Currently, this will always throw an error, so the page basically breaks. But there are plenty valid use cases for empty collections and it should be possible to still use the site (e.g. throw a warning instead of an error).

Also the error is technically not correct. This is what it says:
```
 error   The collection examples does not exist. Ensure a collection directory with this name exists.
 ```

Yet, this is also the error, when the folder exists, but there's just no file in it.

Am I missing something or is this really an oversight in the concept of content collections?

Page should still render

https://stackblitz.com/edit/github-agpp3a?file=src%2Fpages%2Findex.astro

### Participation

### PR DESCRIPTION
**Pull request by DerTimonius**
## Changes

- Remove `AstroError` in `getCollection` if the targeted content directory is empty 
- Warns instead
- Closes #8336 

## Testing

Added a new test in the `content-collection.test.js`, currently work in progress 

## Docs

Docs are unaffected by this change

## Still Todo
Wanted to wait for some feedback before:
- [ ] Finish tests
- [ ] changeset

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -55,9 +55,11 @@ export function createGetCollection({
 		} else if (collection in dataCollectionToEntryMap) {
 			type = 'data';
 		} else {
-			throw new AstroError({
-				...AstroErrorData.CollectionDoesNotExistError,
-				message: AstroErrorData.CollectionDoesNotExistError.message(collection),
+			return zodString().transform((_, ctx) => {
+				ctx.addIssue({
+					code: ZodIssueCode.custom,
+					message: `The collection **${collection}** does not exist. Ensure a collection directory with this name exists.`,
+				});
 			});
 		}
 		const lazyImports = Object.values(
--- a/packages/astro/test/content-collections.test.js
+++ b/packages/astro/test/content-collections.test.js
@@ -230,6 +230,22 @@ describe('Content Collections', () => {
 		});
 	});
 
+	describe('With empty collections directory', () => {
+		it('Handles the empty directory correclty', async () => {
+			const fixture = await loadFixture({
+				root: './fixtures/content-collections-empty-dir/'
+			});
+			let error;
+			try {
+				await fixture.build();
+			} catch (e) {
+				error = e.message;
+			}
+			expect(error).to.be.undefined;
+			// TODO: try to render a page
+		})
+	})
+
 	describe('SSR integration', () => {
 		let app;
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by florian-lefebvre** [comment]
Any news on this? Maybe you can mark it as ready and you'll get reviews about the changeset etc?

**Comment by DerTimonius** [comment]
I wanted to wait for initial feedback, since I was not sure if I'm on the right track here, but I'll mark it as ready for review

File: .changeset/hungry-dancers-hug.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+remove AstroError for empty collection directory
```
ematipico, line 5:
```suggestion
Do not throw an error for an empty collection directory.
```

packages/astro/src/content/runtime.ts
```diff
@@ -55,9 +55,11 @@ export function createGetCollection({
...
 			type = 'data';
 		} else {
-			throw new AstroError({
-				...AstroErrorData.CollectionDoesNotExistError,
-				message: AstroErrorData.CollectionDoesNotExistError.message(collection),
+			return zodString().transform((_, ctx) => {
+				ctx.addIssue({
+					code: ZodIssueCode.custom,
+					message: `The collection **${collection}** does not exist. Ensure a collection directory with this name exists.`,
+				});
```
ematipico, line 62: I am unfamiliar with zod; what do we do here and why?

DerTimonius: To be honest, I'm not entirely happy with this. I think this would still lead to an error. 

My initial idea was to use something like `logger.warn` but I couldn't find a logger?

ematipico: You can get a logger instance from each astro hook: https://docs.astro.build/en/reference/integrations-reference/#astrointegrationlogger

Then, you can pass it around and log whatever you want :)

DerTimonius: Thanks for the hint! I hope this solution is better

ematipico (approved):
I believe we can merge this PR, once rebased

PR #8408, opened 2023-09-04
fix: include route prefix in vercel func names, fix #8401

### LINKED ISSUE #8401: New Vercel adapter generates conflicting lambda functions for routes with same filename but different directories
### Astro Info

```block
Astro                    v3.0.8
Node                     v18.14.2
Package Manager          yarn
                         @astrojs/svelte
```

_No response_

### Describe the Bug

New version of vercel adapter can generate invalid output if two routes have the same file name, but different directory structure.

Consider the structure of `pages` directory below. In some lines I added vercel function names generated by the adapter:
```
├── index.astro      // -> entry.index.astro.func
├── 404.astro        // -> entry.404.astro.func
├── share
│   └── [id].astro   // -> entry._id_.astro.func
├── api
    ├── secrets.ts   // -> entry.secrets.astro.func
    └── secrets
        └── [id].ts  // -> entry._id_.astro.func <- this route produces the same vercel func name as share/[id].astro
```

For each page Vercel adapter is going to build a separate route handler. However, their names may conflict with each other, producing invalid output.

I would expect each route to produce separate lambda function.

I narrowed the problem down to this line in Vercel adapter:
https://github.com/withastro/astro/blob/main/packages/integrations/vercel/src/serverless/adapter.ts#L211

It generates function's directory name based on the file name only. I assume this will brake for every route with the same file names nested in different directories.

I can create PR soon.

https://github.com/slawekkolodziej/secretsecret

### Participation

slawekkolodziej:
## Changes

Recently updated Vercel adapter has a bug. If two routes have identical file names but different directory structures, the adapter will produce only one serverless function and assign it to both routes.

This PR fixes the problem.

Some more details: https://github.com/withastro/astro/issues/8401

## Testing

Within vercel adapter I created 'serverless-with-dynamic-routes' fixture and added necessary tests. If you run those tests against current `main`, the will fail.

## Docs

No need to update docs. This PR only fixes a rather major bug and brings the expected behavior.


slawekkolodziej:
Since my proposed change injects route path to file name, I begun to wonder what is the maximum allowed length for serverless function name in [Vercel build output](https://vercel.com/docs/build-output-api/v3/primitives#serverless-functions)? Since I couldn't find the answer in their documentation, I had to run some manual tests.

The answer is: less than 240 characters. If `*.func` folder has 240 characters or more, Vercel responds with
```
Error: ENAMETOOLONG: name too long
```

With that in mind, I think that the proposed solution is fine for the time being, just to fix the original issue. However ideally there should be a mechanism that limits function name length.

slawekkolodziej:
I thought about limiting the file name to a sane number of characters and adding short content hash to the end.

I like to have route included in the function name - as you said, it simplifies debugging and makes debug.

Now when I think about it, we could ditch `entry-` and `-astro` parts of function names completely, to improve monitoring even more:

Consider what my original change generates:
what-code-generates-now

After some adjustments, Vercel can accept this too:
what-we-can-have

ematipico:
> Now when I think about it, we could ditch entry- and -astro parts of function names completely

- `entry` comes from the core; maybe we should remove it from there?
- `.astro` is just the extension of the file, you would need to remove `.js`, `.ts`, etc., right?

slawekkolodziej:
> * `.astro` is just the extension of the file, you would need to remove `.js`, `.ts`, etc., right?

The complete file extension is `.astro.mjs`, original code already removed `.mjs`.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -207,14 +207,31 @@ You can set functionPerRoute: false to prevent surpassing the limit.`
 
 				// Multiple entrypoint support
 				if (_entryPoints.size) {
-					for (const [route, entryFile] of _entryPoints) {
-						const func = basename(entryFile.toString()).replace(/\.mjs$/, '');
+					// TODO: Currently this function directly checks for a single route 
+					// injected in packages/astro/src/assets/internal.ts Ideally this
+					// should be done dynamically
+					const isInjectedRoute = (route: RouteData) =>
+						route.pathname === '/_image';
```
ematipico, line 214: I think we could change the name in `isImageRoute`, and remove the `TODO`. Here we want to specially allow-list our internal `_image` endpoint.

slawekkolodziej:
I simplified the code a bit, now my routes looks as clean as they can get 💚 

Screenshot 2023-09-05 at 14 54 57

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -207,14 +207,31 @@ You can set functionPerRoute: false to prevent surpassing the limit.`
...
+
+          const getInjectedFuncName = (entryFile: URL) =>
+            basename(entryFile.toString())
+              .replace(/^entry\./, '')
+              .replace(/\.mjs$/, '');
```
ematipico, line 219: Why a regex? Isn't `replace("entry.", "")` enough?

ematipico (approved):
I left some comments. I think overall, we could release this fix.

packages/integrations/vercel/src/serverless/adapter.ts
slawekkolodziej: This regex only matches "`entry.`" at the beginning of the string. You could ask the same question about the original `.replace(/\.mjs$/, '');` code. It could use `.replace(".mjs", "")` too.

packages/integrations/vercel/src/serverless/adapter.ts
slawekkolodziej: Makes sense.

I wonder if there should be any comment left regarding the todo here:
https://github.com/withastro/astro/blob/main/packages/astro/src/assets/internal.ts#L14-L19, just for the reference.

bluwy: Maybe if `route.component` doesn't start with `src/pages/`, we can then fallback for `getInjectedFuncName`? That way we don't have to hardcode this.

packages/integrations/vercel/src/serverless/adapter.ts
slawekkolodziej: Regarding your other comment, if we rename `isInjectedRoute` to `isImageRoute`, it would be natural to rename `getInjectedFuncName` to `getImageFuncName` too, then this function could return a fixed string.

packages/integrations/vercel/src/serverless/adapter.ts
ematipico: Way better

packages/integrations/vercel/src/serverless/adapter.ts
ematipico: Sorry, I ask the question in a different way: Why did you use a regex instead of a simple string? Do they have the same impact performance-wise?

packages/integrations/vercel/src/serverless/adapter.ts
bluwy: Yup looks great! Maybe the functions can be renamed so it's clearer, but the idea and flow is good.

packages/integrations/vercel/src/serverless/adapter.ts
slawekkolodziej: I used it to ensure precision. For example, it avoids incorrectly replacing 'hello.world.entry.astro.mjs' when `entryFile` doesn't start with 'entry.'. Regex allowed me to target replacements specifically at string's start. However, if you are certain that it always begins with `entry.` then that's enough!

Simple string replacement is generally faster, though the speed difference in this case is most likely negligible.

I am not that attached to this regex, I removed it. I just want this PR to get merged so I can upgrade to v3.

natemoo-re (approved):

matthewp (approved):

PR #8410, opened 2023-09-05
Fix some formatting problems and description in astro.ts

morinokami:
## Changes

- Fix some formatting problems.
- Update image options description.

## Testing

N/A because this is just a comment update.

## Docs

N/A because this is just a comment update.


sarah11918 (approved):
Perfect, these are excellent catches @morinokami !  🚀

Issue #8375, opened 2023-09-02
Astro 3 breaks support on `android-arm64v8`


vedxp:
### Astro Info

```block
Astro version            v3.0.6
Package manager          pnpm
Platform                 android
Architecture             arm64
Adapter                  Couldn't determine.
Integrations             unocss
```

### Describe the Bug

I've encountered an issue with Astro v3 where support for the `android-arm64v8` platform seems to have been broken due to recent breaking changes.

Starting from Astro v3, the default image service is set to Sharp. Unfortunately, prebuilt binaries for Sharp is not available for the `android-arm64v8` platform, leading to this problem. When trying to initiate an Astro project on this platform, the following error is encountered:
```sh
pnpm create astro@latest
```
```sh
node_modules/.pnpm/sharp@0.32.5/node_modules/sharp: Running install script, failed in 2.1s
.../sharp@0.32.5/node_modules/sharp install$ (node install/libvips && node install/dll-copy && prebuild-install) || (node install/can-comp…
│ sharp: Downloading https://github.com/lovell/sharp-libvips/releases/download/v8.14.4/libvips-8.14.4-android-arm64v8.tar.br
│ sharp: Installation error: Prebuilt libvips 8.14.4 binaries are not yet available for android-arm64v8
│ sharp: Please see https://sharp.pixelplumbing.com/install for required dependencies
└─ Failed in 2.1s at /data/data/com.termux/files/home/gst/node_modules/.pnpm/sharp@0.32.5/node_modules/sharp
 ELIFECYCLE  Command failed with exit code 1.
```

This issue prevents Astro projects from being properly initialized on the `android-arm64v8` platform. It would be greatly appreciated if this incompatibility could be addressed or if any guidance could be provided on potential workarounds or solutions.

### What's the expected result?

Astro v3 should seamlessly run on the `android-arm64v8` platform and enable project initiation without any hindrance, just as it did in previous versions.

### Link to Minimal Reproducible Example

Couldn't reproduce on web IDEs

### Participation

- [ ] I am willing to submit a pull request for this issue.

jacobthesheep:
The issue is due to the [sharp](https://sharp.pixelplumbing.com/install)  library not supporting android-arm64v8. A possible solution could be to modify Astro to use its "noop" service for platforms that don't support Squoosh or Sharp.

Princesseuh:
[...]

PR #8416, opened 2023-09-05
fix: make sharp an optional dep

Princesseuh:
## Changes

We have some error handling about importing Sharp, so we can make it a optionalDep so it doesn't crash the entire install when it fails to install.

Fix https://github.com/withastro/astro/issues/8375

## Testing

We'll see, planning to do a preview release

## Docs

N/A. I slightly updated the error message to mention you can switch to other image services though

### DIFF
--- a/.changeset/sweet-icons-stare.md
+++ b/.changeset/sweet-icons-stare.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Installing will no longer fail when Sharp can't be installed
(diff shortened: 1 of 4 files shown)


Princesseuh:
!preview optional-sharp

ematipico (approved):

natemoo-re (approved):

bluwy (approved):

Princesseuh:
!preview optional-sharp

PR #8417, opened 2023-09-05
Skip cloudflare CLI tests if fail to start

bluwy:
## Changes

cc @Princesseuh

I tried to use retries, but Mocha doesn't support it in the `before` hook. Implementing retries manually is also tricky, so I went with this PR instead - Try to run the test as usual, but if startup fails, skip the test suite.

## Testing

Ran the cloudflare tests locally. I've not seen it fail locally though.

## Docs

n/a.

### DIFF
--- a/packages/integrations/cloudflare/test/cf.test.js
+++ b/packages/integrations/cloudflare/test/cf.test.js
@@ -9,7 +9,7 @@ describe('Cf metadata and caches', () => {
 	/** @type {import('./test-utils').WranglerCLI} */
 	let cli;
 
-	before(async () => {
+	before(async function () {
 		fixture = await loadFixture({
 			root: './fixtures/cf/',
 			output: 'server',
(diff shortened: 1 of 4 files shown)


packages/integrations/cloudflare/test/cf.test.js
```diff
@@ -9,16 +9,20 @@ describe('Cf metadata and caches', () => {
...
 		fixture = await loadFixture({
 			root: './fixtures/cf/',
 			output: 'server',
 			adapter: cloudflare(),
 		});
 		await fixture.build();
 
-		cli = await runCLI('./fixtures/cf/', { silent: false, port: 8788 });
-		await cli.ready;
+		cli = await runCLI('./fixtures/cf/', { silent: true, port: 8786 });
```
bluwy, line 20: Changed the port number as it's close to the other test's `8789` port, which maybe was causing issues somehow.

natemoo-re (approved):
Nice, thank you for the QoL improvement!

Issue #8378, opened 2023-09-02
HMR does not work when editing markdown / .md files.


davidbernegger:
### Astro Info

```block
Astro                    v3.0.7
Node                     v18.15.0 / v20.3.1 (tried both)
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

HMR does not work on Astro 3.0.7 when editing `.md` files in the pages directory. Astro does recognise the change, as it shows `[astro] reload /src/pages/markdown.md` whenever i save, but the browser does not refresh unlike the `.astro` files. In the Minimal reproducible example, navigate to the markdown page and try to edit it.

### What's the expected result?

The browser should refresh with the current content 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ve9ttx?file=src%2Fpages%2Fmarkdown.md

### Participation

- [ ] I am willing to submit a pull request for this issue.

lloydjatkinson:
Did you try with the markdown file in /content, as part of content collections?

davidbernegger:
Yeah, this works, but putting `.md` in the pages directory is supported according to the [official documentation](https://docs.astro.build/en/core-concepts/astro-pages/#supported-page-files)

bluwy:
Looks like it's happening since 3.0.0-beta.2, works in beta.0 (beta.1 has a faulty dep). From a briefly look, it seems that `handleHotUpdate` here:

https://github.com/withastro/astro/blob/8352ee0a9a6d27a0823d0ff4b3df2526eb2b5ca5/packages/astro/src/vite-plugin-astro/index.ts#L183-L188

Should be returning an array with a single element of the markdown file module node, but it doesn't in the latest version. Vite's module graph seems to not have the markdown file module node.

PR #8418, opened 2023-09-05
Fix markdown page HMR

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/8378

The [JSX refactor PR](https://github.com/withastro/astro/pull/7924) updated vite-plugin-markdown so it returns an Astro component instance. When rendering Astro components, head is not rendered by default unless it calls `renderHead` or `maybeRenderHead`, which it doesn't.

This PR adds it to preserve the previous behaviour, and to fix the HMR issue.

Before the PR, markdown pages return non-Astro components and it has automatic head rendering, denoted by the `astro.needsHeadRendering` symbol. So it worked before.

## Testing

Added a new test to ensure to Vite script is added.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/twelve-cars-tell.md
+++ b/.changeset/twelve-cars-tell.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix markdown page HMR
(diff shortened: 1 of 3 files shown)


matthewp (approved):

natemoo-re (approved):
Clever fix!

PR #8424, opened 2023-09-05
fix: caching not working for `astro:assets`

itsmatteomanf:
## Changes

It reverses the check for remote images in `astro:assets`, `JSONData.expires` should be higher than `Date.now()`, not lower.

## Testing

No new tests, but tested locally by changing the sign.

## Docs

Actually makes what the docs say should happen.


natemoo-re (approved):
Thanks for the PR!

PR #8426, opened 2023-09-06
feat(@astrojs/cloudflare): add runtime support to `astro dev`

alexanderniebuhr:
## Changes

- **clean up**: there were still some references to old `getRuntime`
- **workaround**: until wrangler exposes for of their API for programatically usage, I just copied over the important logic
- **add**: `req.cf` or `Astro.locals.runtime.cf` support for `astro dev`, without the need to run wrangler
- **add**: `Astro.locals.runtime.env` support for enviromental vars defined according to CF docs in either `wrangler.toml` or `.dev.vars`

## Known issues (for another PR)

- **missing**: CF bindings mocking (R2, D1, KV, Durable Objects)
- **missing**: support for env vars defined in other files than `wrangler.toml`, `.dev.vars`
- **missing**: `wrangler.json` support

## Testing

- manually

## Docs

~~- would love to add them with a later PR, just want to stealth release it first~~


packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -44,19 +44,6 @@ export function createExports(manifest: SSRManifest) {
...
-				env,
-				name: 'cloudflare',
-				caches,
-				cf: request.cf,
-				...context,
-				waitUntil: (promise: Promise<any>) => {
-					context.waitUntil(promise);
-				},
-			});
-
```
lilnasy, line 59: Is this safe at this point? I expect users to be already depending on it in their astro 3.0 projects.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -105,17 +206,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+									// will be fetched from git dynamically once we support mocking of bindings
+									CF_PAGES_COMMIT_SHA: 'TBA',
+									CF_PAGES_URL: `http://${req.headers.host}`,
+									...vars,
+								},
+								cf: cf,
+								// will be supported once we support mocking of bindings
+								// waitUntil: (promise: Promise<any>) => {
+								//   context.waitUntil(promise);
+								// },
```
lilnasy, line 243: I think it would be appropriate for this to be a no-op.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -105,17 +206,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+								// will be supported once we support mocking of bindings
+								// caches: caches,
```
lilnasy, line 245: I would think this can be mocked right here as well. Probably an in-memory cache, because caching across `astro dev` invocations is asking for trouble.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -66,6 +74,90 @@ export function getAdapter({
...
+	if (runtimeMode === 'local') {
+		return CF_FALLBACK;
+	} else if (runtimeMode === 'remote') {
+		// add caching to stop extensive querying
+		// try {
+		// 	const storedCf = JSON.parse(await fs.readFile(cfPath, "utf8"));
+		// 	const cfStat = await fs.stat(cfPath);
+		// 	assert(Date.now() - cfStat.mtimeMs <= CF_DAYS * DAY);
+		// 	return storedCf;
+		// } catch {}
```
lilnasy, line 145:
```suggestion
```

packages/integrations/cloudflare/src/index.ts
```diff
@@ -66,6 +74,90 @@ export function getAdapter({
...
+		try {
+			const res = await fetch(CF_ENDPOINT);
+			const cfText = await res.text();
+			const storedCf = JSON.parse(cfText);
+			// // Write cf so we can reuse it later
+			// await mkdir(path.dirname(cfPath), { recursive: true });
+			// await writeFile(cfPath, cfText, "utf8");
+			// log.debug("Updated `Request.cf` object cache!");
```
lilnasy, line 153:
```suggestion

```

packages/integrations/cloudflare/src/index.ts
```diff
@@ -66,6 +74,90 @@ export function getAdapter({
...
 
+async function getCFObject(runtimeMode: string): Promise<IncomingRequestCfProperties | void> {
+	// Milliseconds in 1 day
+	const DAY = 86400000;
+	// Max age in days of cf.json
+	const CF_DAYS = 1;
+
+	const CF_ENDPOINT = 'https://workers.cloudflare.com/cf.json';
+
+	const CF_FALLBACK: IncomingRequestCfProperties = {
```
lilnasy, line 85: Can you add a comment here. Is this random data? Should the users be aware the cloudfalre adapter is making this up.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -1,18 +1,26 @@
...
 export type { AdvancedRuntime } from './server.advanced';
 export type { DirectoryRuntime } from './server.directory';
 
 type Options = {
 	mode: 'directory' | 'advanced';
 	functionPerRoute?: boolean;
+	/**
+	 * 'off': current behaviour (wrangler is needed)
+	 * 'local': use a static req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
+	 * 'remote': use a dynamic real-live req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
```
lilnasy, line 21: I'm not sure what "astro dev is enough" means here. Could you briefly mention the alternative in this typedoc?

packages/integrations/cloudflare/src/CFVars.ts
```diff
@@ -0,0 +1,141 @@
...
+	} else if (configPath?.endsWith('json')) {
+		// rawConfig = parseJSONC(fs.readFileSync(configPath), configPath);
+	}
+	const vars = getVarsForDev(rawConfig, configPath);
+	// // Process the top-level configuration.
+	// const { config, diagnostics } = normalizeAndValidateConfig(
+	// 	rawConfig,
+	// 	configPath,
+	// 	args
+	// );
```
lilnasy, line 139:
```suggestion
```

packages/integrations/cloudflare/src/CFVars.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+export async function CFVars() {
+	let rawConfig;
+	const configPath = findWranglerToml(process.cwd(), false); // false = args.experimentalJsonConfig
+
+	// Load the configuration from disk if available
+	if (configPath?.endsWith('toml')) {
+		rawConfig = parseTOML(fs.readFileSync(configPath).toString(), configPath);
+	} else if (configPath?.endsWith('json')) {
+		// rawConfig = parseJSONC(fs.readFileSync(configPath), configPath);
```
lilnasy, line 131:
```suggestion
```

packages/integrations/cloudflare/src/CFVars.ts
```diff
@@ -0,0 +1,141 @@
...
+export function loadDotEnv(path: string): DotEnv | undefined {
+	return tryLoadDotEnv(path);
+}
+function getVarsForDev(config: any, configPath: string | undefined): any {
+	const configDir = resolve(dirname(config.configPath ?? '.'));
+	const devVarsPath = resolve(configDir, '.dev.vars');
+	const loaded = loadDotEnv(devVarsPath);
+	if (loaded !== undefined) {
+		const devVarsRelativePath = relative(process.cwd(), loaded.path);
+		// logger.log(`Using vars defined in ${devVarsRelativePath}`);
```
lilnasy, line 114:
```suggestion
```
Comments can be removed. If you realise it's time to bring this code back, it would be possible to find this commit in this PR.

lilnasy (changes requested):
Love this! Setting the standard for adapters yet again.

As important as documentation is for this feature, it could be deferred. But there should at least be basic tests. If for nothing else, then ensuring that things don't wildly change.

packages/integrations/cloudflare/src/server.advanced.ts
alexanderniebuhr: As `runtime.ts` files was removed last week, `getRuntime()` wouldn't work at all. :)

lilnasy: OK thanks for clarifying

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: Sorry, I'm at a loss, what exactly do you mean by no-op?

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: Absolutely! The introduction of mocking is planned as a second phase, primarily because it necessitates more intricate architectural design. I aim to align this closely with the bindings feature.

As for caches, they will be excluded in the initial implementation. If you're dependent on cache functionality, the existing wrangler pages dev workaround remains a viable option for the time being.

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: This data serves as the fallback, consistent with what is used in both Miniflare and Workerd. I'll annotate the code to clarify this point.

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: For context, previously one had to either use wrangler in "proxy-mode" or build the Astro project and point to the dist folder. The phrase "astro dev is enough" signifies that running astro dev now provides runtime information, a feature previously exclusive to wrangler.

This comment is intended solely for contributors and code-reviewers; detailed documentation will be added in the second phase of this topic.

alexanderniebuhr:
@lilnasy, thank you for your insightful comments! You're correct that I included code comments intended for a later stage. I agree with your suggestion to remove them for the time being; they can always be reintroduced later.

packages/integrations/cloudflare/src/index.ts
lilnasy: .waitUntil() only makes sense for serverless environments, where the instance needs to be told to stay awake.

Astro's dev server is going to stay up anyway, so this function can be a non-operation, a function that does nothing.

packages/integrations/cloudflare/src/index.ts
lilnasy: In that case, I think it makes sense for it be introduced as no-ops as well.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -105,17 +195,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				_buildConfig = config.build;
 
 				if (config.output === 'static') {
-					throw new Error(`
-  [@astrojs/cloudflare] \`output: "server"\` or \`output: "hybrid"\` is required to use this adapter. Otherwise, this adapter is not necessary to deploy a static site to Cloudflare.
-
-`);
+					throw new Error(STATIC_ERROR);
```
ematipico, line 198: Use `AstroError` -> `import { AstroError } from "astro/error"`

packages/integrations/cloudflare/src/index.ts
```diff
@@ -105,17 +195,49 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 				}
 
 				if (config.base === SERVER_BUILD_FOLDER) {
-					throw new Error(`
-  [@astrojs/cloudflare] \`base: "${SERVER_BUILD_FOLDER}"\` is not allowed. Please change your \`base\` config to something else.`);
+					throw new Error(FOLDER_ERROR);
 				}
 			},
+			'astro:server:setup': ({ server }) => {
+				server.middlewares.use(async function middleware(req, res, next) {
```
ematipico, line 206: I think it's better to add a new middleware only if `runtimeMode` isn't `'off'`:

```js
if( runtimeMode !== 'off') {
	server.middlewares.use();
}
```

ematipico (commented):
Code-wise, it looks good. I agree with every comment @lilnasy made

We should add some documentation once we settle the business logic

alexanderniebuhr:
Fixed most of the comments. I actually implemented local `caches` support, using miniflare. Will add some tests & docs later today.

packages/integrations/cloudflare/README.md
```diff
@@ -159,6 +159,25 @@ export function GET({ params }) {
 }
 ```
 
+### Astro Dev Server
+
+The Adapter allows an opt-in flag, to enable the use of Environment Variables and the Cloudflare Request Object to be populated by the Astro Dev Server. There is no need to use wrangler in this case. This flag is `off` by default.
+
+There are two supported modes: `local` & `remote`. In the mode `local` environmental variables are available, but the request object is populated from a static placeholder value. In the mode `remote` environmental variables and the a live fetched request object are available.
```
sarah11918, line 166:
```suggestion
### `cloudflare.runtime` 

`runtime: "off" | "local" | "remote"`
default `"off"`

This optional flag enables the Astro dev server to populate environment variables and the Cloudflare Request Object, avoiding the need for Wrangler.

-  `local`: environmental variables are available, but the request object is populated from a static placeholder value. 
- `remote`: environmental variables and the live, fetched request object are available.
- `off`: the Astro dev server will populate neither environment variables nor the request object. Use Wrangler to access Cloudflare bindings and environment variables.
```

I don't think "dev server" is the key idea here, it's the config object. I took a stab at this from the structure of the `mode` section above. Please make this true/in the right jargon, but I think this kind of structure will work well here!

packages/integrations/cloudflare/README.md
```diff
@@ -87,7 +87,7 @@ It's then possible to update the preview script in your `package.json` to `"prev
 
 ## Access to the Cloudflare runtime
 
-You can access all the Cloudflare bindings and environment variables from Astro components and API routes through `Astro.locals`.
+You can access all the Cloudflare bindings and environment variables from Astro components and API routes through `Astro.locals`. Usually you need to use [Wrangler](https://github.com/cloudflare/workers-sdk) to be able to access the runtime.
```
sarah11918, line 90:
```suggestion
You can opt in to access all the Cloudflare bindings and environment variables from Astro components and API routes through `Astro.locals`. This replaces using [Wrangler](https://github.com/cloudflare/workers-sdk) to access the runtime.
```
I think this needs to be clearer, because the above phrase is a bit sophisticated English usage that requires familiarity with the two parts of that common structure, and only implies the (important) second half. It *could* potentially be interpreted as "Usually, you WILL (present tense! You will! this is our guidance!) use Wrangler to do this" NOT as you intended, "Usually you'd use Wrangler, **but now you don't have to!**") 

You could also add on to the existing sentence something that is explicitly clear, 

> ... to access the runtime, but configuring `cloudflare.runtime` will allow you to use Astro's dev server instead."

sarah11918 (commented):
Thanks @alexanderniebuhr! Just some docs tightening up to do, and I've given you a structure that you can adjust as necessary!  🙌

matthewp (approved):
Love this

ematipico (approved):
Great work!

packages/integrations/cloudflare/src/index.ts
```diff
@@ -1,18 +1,31 @@
...
 import { sep } from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import glob from 'tiny-glob';
+import { getEnvVars } from './parser.js';
 
 export type { AdvancedRuntime } from './server.advanced';
 export type { DirectoryRuntime } from './server.directory';
 
 type Options = {
 	mode: 'directory' | 'advanced';
```
lilnasy, line 21:
```suggestion
	mode?: 'directory' | 'advanced';
```
When users want to specify `runtime`, they shouldn't need to specify `mode` as well.

alexanderniebuhr: That makes sense, do we want to default to `advanced` mode?

lilnasy: Yeah, that seems to be the implied default already.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -105,15 +197,56 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+						'[@astrojs/cloudflare] `base: "${SERVER_BUILD_FOLDER}"` is not allowed. Please change your `base` config to something else.'
+					);
+				}
+			},
+			'astro:server:setup': ({ server }) => {
+				if (runtimeMode !== 'off') {
+					server.middlewares.use(async function middleware(req, res, next) {
+						try {
+							const cf = await getCFObject(runtimeMode);
+							const vars = await getEnvVars();
```
lilnasy, line 216: In my testing, this step is always failing and locals never get set.

lilnasy: Seems like when wrangler.toml doesn't exist, the runtime locals don't appear.

packages/integrations/cloudflare/README.md
alexanderniebuhr: While I concur with your comment, the technical implementation requires an additional PR to align with it. For now, I've removed the reference to wrangler in this section, as the other part of the Readme already contains all the necessary information.

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: You're right, definitely something I'll change in a future PR, since it kinda should be more explicit.

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: Do you have a `wrangler.toml` & `.dev.vars` file?

packages/integrations/cloudflare/README.md
```diff
@@ -159,6 +159,30 @@ export function GET({ params }) {
...
 
+### `cloudflare.runtime` 
+
+`runtime: "off" | "local" | "remote"`
+default `"off"`
+
+This optional flag enables the Astro dev server to populate environment variables and the Cloudflare Request Object, avoiding the need for Wrangler.
+
+-  `local`: environmental variables are available, but the request object is populated from a static placeholder value. 
+- `remote`: environmental variables and the live, fetched request object are available.
```
sarah11918, line 170:
```suggestion
-  `local`: environment variables are available, but the request object is populated from a static placeholder value. 
- `remote`: environment variables and the live, fetched request object are available.
```
Ignore if "environmental" really is the right word, we just don't use it anywhere else so I'm double checking!

alexanderniebuhr: Good catch ❤️

packages/integrations/cloudflare/src/index.ts
alexanderniebuhr: > missing: support for env vars defined in other files than wrangler.toml, .dev.vars

This limitation is recognized and is on my roadmap. I plan to address it in an upcoming PR alongside the introduction of bindings. Thanks for pointing it out; no action required at this moment.

sarah11918 (approved):
It's been a pleasure doing doc'ing with you!  🫡

lilnasy (approved):
This is great! Thanks for all the hardwork you put in here.

Issue #8352, opened 2023-09-01
Astro 3.0 dev server error when base path is set and React component loaded


dallyh:
### Astro Info

```block
Astro                    v3.0.5
Node                     v18.17.1
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/react
```

### Describe the Bug

When the base path is set in the Astro config, and an React component is used with the "client" directive, Astro fails to load the webpage only when using the dev server.  After the webpage gets build everything works fine. This kind of neglectes any attempts to quickly develop and debug React components. This issue seems a little widespread, as some people on Discord confirmed that it is not only me.

Config is set like this:
```javascript
// https://astro.build/config
export default defineConfig({
  integrations: [react()],
  base: "basepath",
});
```

The error which prevents load/render of the page is:

`
Failed to resolve import "/basepath/@react-refresh" from "astro:scripts/before-hydration.js". Does the file exist?
`

### What's the expected result?

Site with React components and a base path set in a config should work in a dev server without any errors.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/dallyh/astro-react-refresh-issue

### Participation

- [ ] I am willing to submit a pull request for this issue.

paoloose:
Relevant piece of code introduced in Astro 3.0: https://github.com/withastro/astro/pull/7924/commits/e07a4e5db6d24c016995c330a4720439b32e7d8f#diff-714fabdf53991de22f8549ac97613b5d49f9c6aaa096a57d8a29f220758978c4R66-R69

Plugin's side: https://github.com/vitejs/vite-plugin-react/blob/286360281992c425bf75cb0a18846f65fcdc5ef3/packages/plugin-react/src/fast-refresh.ts#L23-L29

Not replacing `__BASE__` with the Astro `config.base` but with a forward slash `/` completely solved the problem for me, but I'm not sure it's the proper fix

```diff
        if (command === "dev") {
          const preamble = FAST_REFRESH_PREAMBLE.replace(
            `__BASE__`,
-            appendForwardSlash(config.base)
+           '/'
          );
          injectScript("before-hydration", preamble);
        }
```

sasoria:
I'm getting the same error as well. This is a temporary fix I'm using:
``` ts
export default defineConfig({
  base: import.meta.env.DEV ? "" : "/base-path",
  ...
}); 
```

PR #8428, opened 2023-09-06
Fix React dev mode using a base

matthewp:
## Changes

- React Refresh is the module `@react-refresh`. Previously we were loading it as `/basepathname/@react-refresh`. Instead packages need to be loaded from the absolute root, regardless of whether a base is used or not.
- Fixes https://github.com/withastro/astro/issues/8352

## Testing

- Tested in a demo project.
- Can't be tested in our integration tests due to it requiring the browser loading the module.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/cold-knives-peel.md
+++ b/.changeset/cold-knives-peel.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+Fix React dev mode using a base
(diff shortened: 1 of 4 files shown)


Princesseuh (approved):

bluwy (approved):
LGTM. Seems like it's able to serve both with and without a basepath. The error was happening because we inject a script that's processed by Vite. Injecting as an inline script could also solve this.

matthewp:
Astro's 'head-inline' scripts are non-module scripts so that wouldn't work here. Going with this as it seems to work.

PR #8431, opened 2023-09-06
Temporarily pin Node 20 version in CI

matthewp:
## Changes

- Node 20 is currently broken, pinning our tests to the last working version to unblock PRs.


Princesseuh (approved):
Let's try to remember to remove this when Node 20 is fixed, ha

PR #8440, opened 2023-09-06
Fix `renderToFinalDestination` handling

### LINKED ISSUE #8434: I updated from v2 to V3  and now I get error when i do npm run dev
### Astro Info

```block
Astro                    v3.0.9
Package Manager          npm
Adapter                  none
                         @astrojs/tailwind
```

chrome

### Describe the Bug

I updated from v2 to V3 and now I get this error when I use npm run dev. Now using node v18.17.1 (npm v10.0.0) Deleted and node modules folder and package-lock.json and used npm install while switching between versions 20.5.1, 18.14.1, 18.17.1 and still got the same error.
astro2

I expect that the website shows up on localhost:4321

--

### Participation

natemoo-re:
## Changes

- Closes #8434
- Properly handles children that return `undefined`
- Seems to happen with Markdown content mainly? Not sure why.

## Testing

Manually tested with user reproduction

## Docs

Bug fix only

### DIFF
--- a/packages/astro/src/runtime/server/render/any.ts
+++ b/packages/astro/src/runtime/server/render/any.ts
@@ -18,6 +18,7 @@ export async function renderChild(destination: RenderDestination, child: any) {
 			});
 		});
 		for (const childRender of childRenders) {
+			if (!childRender) continue;
 			await childRender.renderToFinalDestination(destination);
 		}
 	} else if (typeof child === 'function') {
--- a/.changeset/tame-fans-happen.md
+++ b/.changeset/tame-fans-happen.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/astro/src/runtime/server/render/any.ts
```diff
@@ -18,6 +18,7 @@ export async function renderChild(destination: RenderDestination, child: any) {
 			});
 		});
 		for (const childRender of childRenders) {
+			if (!childRender) continue;
```
natemoo-re, line 21: We don't have to wait for @bluwy to weigh in, but it seems like some of the `childRender` objects are `undefined` which is kind of unexpected. Not sure I fully grok why that's happening? Is it Markdown-specific?

matthewp (approved):

lilnasy (approved):
Barely familiar with this code path, but lgtm.

PR #8441, opened 2023-09-06
Transition only between pages where both have ViewTransitions enabled

martrapp:
## Changes

Before: 
The history could contain states which are not handled by the client-side router when it dies during a visit to pages without `<ViewTransitions>`. The client-side router dies when a page w/o `<ViewTransitions>` is (re)loaded (no swap).  The load deletes the router's modules. This is not a problem if the user navigates back to a page with `<ViewTransitions>` and this is reloaded. But if the browser history contains an entry for that page that was originally created by the router, the browser expects the now-deceased router to force the reload for that history entry. This results in a broken backward navigation.

Change:
With this change, new history entries and (view transitions) will only be created if the target page also defines `<ViewTransitions>`.  This way the browser's backward navigation will always restore the router in time, even if lost on pages without `<ViewTransitions>`.  

## Testing

Added e2e test to prove that a killed router is revived correctly. 
Adjusted existing tests because navigation to pages without `<ViewTransition>` leads to reloads.

## Docs

n/a. bug fix


matthewp (approved):

matthewp:
Looks great! Should note that we checked `<meta name="view-transition" content="same-origin" />` and it behaves the same way as this PR. This prevents bugs were you can get "stuck" going forward to pages without VT and then trying to come back.

PR #8443, opened 2023-09-06
Fix "Cannot read properties of null" in cli/telemetry

the-dijkstra:
## Changes

Latest version of Astro is throwing the following error on CI:

```
error   Cannot read properties of null (reading 'name')
    File:
      node_modules/astro/dist/cli/telemetry/index.js:5:57
    Code:
      4 | async function notify() {
      > 5 |   const packageManager = (await whichPm(process.cwd())).name ?? "npm";
          |                                                         ^
        6 |   await telemetry.notify(() => {
        7 |     console.log(msg.telemetryNotice(packageManager) + "\n");
        8 |     return true;
    Stacktrace:
  TypeError: Cannot read properties of null (reading 'name')
      at notify (node_modules/astro/dist/cli/telemetry/index.js:5:57)
      at async runCommand (node_modules/astro/dist/cli/index.js:97:3)
      at async cli (node_modules/astro/dist/cli/index.js:144:5)
```

I will add changsets if this gets approved

### DIFF
--- a/.changeset/strong-colts-hang.md
+++ b/.changeset/strong-colts-hang.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix "Cannot read properties of null" error in CLI code
--- a/packages/astro/src/cli/telemetry/index.ts
+++ b/packages/astro/src/cli/telemetry/index.ts
@@ -9,7 +9,7 @@ interface TelemetryOptions {
 }
 
 export async function notify() {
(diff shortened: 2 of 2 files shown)


jacobthesheep:
Oops my mistake for forgetting to add the '?' to it.

Princesseuh (approved):
Thanks, this looks great! Surprised this wasn't a type error 🤔

Issue #8322, opened 2023-08-31
Vercel image service breaks svg when used with `astro:assets`


florian-lefebvre:
### Astro info

```block
Astro version            v3.0.3
Package manager          npm
Platform                 linux
Architecture             x64
Adapter                  @astrojs/vercel
Integrations             None or couldn't determine.
```

### What browser are you using?

Brave

### Describe the Bug

When using the Vercel static adapter and enabling the imageService locally, svg assets passed to the `Image` component are broken. Their link points to an error 500 with message `Server Error: Buffer has an unsupported format`.

### What's the expected result?

Be able to pass svgs to the Image component

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-aprtsf

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Can confirm that SVGs break when using

```js
vercel({ imageService: true })
```

I'm not too sure what the fix here will be. We might need a way for image services to configure which formats they support. cc @Princesseuh

Princesseuh:
> Can confirm that SVGs break when using
> 
> ```js
> vercel({ imageService: true })
> ```
> 
> I'm not too sure what the fix here will be. We might need a way for image services to configure which formats they support. cc @Princesseuh 

Is this in dev or production? In dev we don't use Vercel's service, we use Squoosh. I believe the error message comes from Squoosh too.

florian-lefebvre:
Broken in dev, haven't tried on vercel but I will do so

florian-lefebvre:
Can confirm it works on vercel so it must come from squoosh

PR #8445, opened 2023-09-07
feat(vercel): Use Sharp in dev instead of Squoosh by default

Princesseuh:
## Changes

This changes the image service in dev to be Sharp powered like Astro's base service instead of Squoosh. The Squoosh one is still available if needed for people who can't install Sharp for various reasons.

This also fixes https://github.com/withastro/astro/issues/8322

## Testing

Added a test for SVGs and current tests still pass

## Docs

The README was updated with the new config!

### DIFF
--- a/.changeset/cold-flies-clean.md
+++ b/.changeset/cold-flies-clean.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vercel': major
+---
+
+Use Sharp in dev mode instead of Squoosh by default
--- a/packages/integrations/vercel/README.md
+++ b/packages/integrations/vercel/README.md
@@ -170,6 +170,27 @@ import astroLogo from '../assets/logo.png';
 />
 ```
 
+### useSquooshDev
+
+**Type:** `boolean`<br>
+**Available for:** Serverless, Static
+**Added in:** `@astrojs/vercel@3.3.0`
+
+When enabled, in development mode a Squoosh-based image service will be used in development instead of Sharp. This is useful if you cannot install Sharp's dependencies on your development machine.
+
+```js
+import { defineConfig } from 'astro/config';
+import vercel from '@astrojs/vercel/serverless';
+
+export default defineConfig({
+  output: 'server',
+  adapter: vercel({
+    imageService: true,
+    useSquooshDev: true,
+  }),
+});
+```
+
 ### includeFiles
 
 **Type:** `string[]`<br>
--- a/packages/integrations/vercel/src/serverless/adapter.ts
+++ b/packages/integrations/vercel/src/serverless/adapter.ts
@@ -64,6 +64,7 @@ export interface VercelServerlessConfig {
 	analytics?: boolean;
 	imageService?: boolean;
 	imagesConfig?: VercelImageConfig;
+	useSquooshDev?: boolean;
 	edgeMiddleware?: boolean;
 	functionPerRoute?: boolean;
 }
@@ -74,6 +75,7 @@ export default function vercelServerless({
 	analytics,
 	imageService,
 	imagesConfig,
+	useSquooshDev,
 	functionPerRoute = true,
 	edgeMiddleware = false,
 }: VercelServerlessConfig = {}): AstroIntegration {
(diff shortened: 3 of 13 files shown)


ematipico (approved):

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -68,6 +68,7 @@ export interface VercelServerlessConfig {
 	analytics?: boolean;
 	imageService?: boolean;
 	imagesConfig?: VercelImageConfig;
+	useSquooshDev?: boolean;
```
bluwy, line 71: Would it make sense to change this as `devImageService: 'sharp' | 'squoosh'` instead? In case Astro supports a new image transformer in the future, we can reserve it here.

Princesseuh: That is definitely a better thing to do, thank you for not letting me ship this mistake

packages/integrations/vercel/README.md
```diff
@@ -172,6 +172,27 @@ import astroLogo from '../assets/logo.png';
 />
 ```
 
+### useSquooshDev
+
+**Type:** `boolean`<br>
+**Available for:** Serverless, Static
+**Added in:** `@astrojs/vercel@3.3.0`
+
+When enabled, in development mode a Squoosh-based image service will be used in development instead of Sharp. This is useful if you cannot install Sharp's dependencies on your development machine.
```
sarah11918, line 181:
```suggestion
Allows you to configure a Squoosh-based image service to be used in development instead of Sharp. This is useful if you cannot install Sharp's dependencies on your development machine.
```

.changeset/cold-flies-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vercel': major
+---
+
+Use Sharp in dev mode instead of Squoosh by default
```
sarah11918, line 5:
```suggestion
Adds a configuration option `useSquooshDev` to use use Squoosh in dev mode instead of Sharp by default.
```

packages/integrations/vercel/README.md
```diff
@@ -172,6 +172,27 @@ import astroLogo from '../assets/logo.png';
 />
 ```
 
+### useSquooshDev
+
+**Type:** `boolean`<br>
+**Available for:** Serverless, Static
+**Added in:** `@astrojs/vercel@3.3.0`
```
sarah11918, line 179: Maybe also add a `**Default**:  false` here?

sarah11918 (approved):
Hi @Princesseuh -- Double check because it sounded like the changeset was saying the opposite of what the README was? Or I just missed something?

Made some quick suggestions below for your consideration!

Princesseuh:
> Hi @Princesseuh -- Double check because it sounded like the changeset was saying the opposite of what the README was? Or I just missed something?
> 
> Made some quick suggestions below for your consideration!

Sorry, this PR went through a lot and some things ended up being outdated while I thought everything was done with it 😔 I'd love if you could re-review this, as the option changed, instead of being a boolean for Squoosh, it now allows you to choose between Sharp and Squoosh or your own image service, which leads to more flexibility

packages/integrations/vercel/README.md
```diff
@@ -172,6 +172,30 @@ import astroLogo from '../assets/logo.png';
...
 ```
 
+### devImageService
+
+**Type:** `'sharp' | 'squoosh' | string`<br>
+**Available for:** Serverless, Static
+**Added in:** `@astrojs/vercel@3.3.0`
+**Default**: 'sharp'
+
+Allows you to configure a which image service to use in development. This can be useful if you cannot install Sharp's dependencies on your development machine and prefer to use Squoosh instead.
```
sarah11918, line 182:
```suggestion
Allows you to configure which image service to use in development. This can be useful if you cannot install Sharp's dependencies on your development machine, but using another image service like Squoosh would allow you to preview images in your dev environment.
```
I know this might sound like a little much, but I wonder whether adding a purpose to this statement (pick the reason you think makes sense!) helps to describe the choice and also really emphasize that this is JUST for development. (Build is unaffected.  <-- and you could even add something like that, if you wanted!)

Issue #8399, opened 2023-09-04
In rare cases, CSS for transition scopes is not generated 


martrapp:
### Astro Info

```block
Astro                    v3.0.8
Node                     v20.5.1
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

A certain structure of nested components leads to CSS not being generated for transition scopes.
First found here: https://discord.com/channels/830184174198718474/1147944562682253394 by Nash

### What's the expected result?

CSS generation works independent of the component structure

### Link to Minimal Reproducible Example

https://github.com/martrapp/ronijaakkola-website.git

### Participation

- [X] I am willing to submit a pull request for this issue.

martrapp:
Preliminary analysis so far:
The required style sheet is pushed to SSRResult._metadata.extraHead in renderTransition().

In this example, however, the transition:name is not found until after the <head> rendering is complete.
Two maybe-head instructions follow after the transition:name is found, but they do nothing because the head is already closed.

lilnasy:
Related #7761

matthewp:
This is a head propagation bug, i'll take a look.

hkfi:
I ran into this issue as well. However, I had my `<ViewTransitions />` as a child of `<body>` instead of `<head>`. My issue was resolved after moving `<ViewTransitions />` to be a child of `<head>`.

matthewp:
I'm still looking at this. It wasn't what I was expecting but pretty confident a fix is close.

matthewp:
Think I have a solution here, working on a test.

matthewp:
Fix is in #8449.

adrianosmond:
I don't think that the fix that landed in 3.0.12 has fixed the stackblitz linked in the original post. If you go to the stackblitz link, update Astro to 3.0.12 and click the "Not working" link you'll see that you only get a full page cross-fade transition. On inspecting the <img /> (on the not working page, not the landing page) you can see it has a `data-astro-transition-scope` but no CSS with a view-transition-name for it has been generated.

martrapp:
3.0.12 fixes this issue (#8339) and the repository linked above now generates the required CSS:
```
<head>
  ...
  <style>
    [data-astro-transition-scope="astro-z5xyy7w3-1"] { view-transition-name: title-project-1; }@layer astro { ::view-transition-old(title-project-1) { 
  ...
```
@adrianosmond what stackblitz are you looking at?

[...]

Issue #8392, opened 2023-09-04
Nesting Astro components prevents transition directives from animating


sam-high:
### Astro Info

```block
Astro                    v3.0.12
Node                     v16.20.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Nested **View Transitions** stop animating when HTML with **Transition Directives** is nested too deep within Astro components.

**e.g. when nesting an image**

```
<img
  transition:name="test"
  transition:animate="initial"
  src="https://placehold.co/600x400/png"
/>
```

**This level of nesting works**
-
    Images (Astro Component)
        Image (Astro Component)
            HTML image (with Transition Directives)
            
**This level of nesting or more does not work**
-
    Images Container (Astro Component) ** Extra level
        Images (Astro Component)
            Image (Astro Component)
                HTML image (with Transition Directives)            
                
                
If I can help make this issue clearer please let me know.

### What's the expected result?

Nested **View Transitions** to animate no matter how deeply they are nested.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ojxahh?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

adrianosmond:
I don't think that the fix that landed in 3.0.12 has fixed the stackblitz linked in the original post. If you go to the stackblitz link, update Astro to 3.0.12 and click the "Not working" link you'll see that you only get a full page cross-fade transition. On inspecting the  (on the not working page, not the landing page) you can see it has a data-astro-transition-scope but no CSS with a view-transition-name for it has been generated.

sam-high:
> I don't think that the fix that landed in 3.0.12 has fixed the stackblitz linked in the original post. If you go to the stackblitz link, update Astro to 3.0.12 and click the "Not working" link you'll see that you only get a full page cross-fade transition. On inspecting the (on the not working page, not the landing page) you can see it has a data-astro-transition-scope but no CSS with a view-transition-name for it has been generated.

You are right. I have updated the [Link to Minimal Reproducible Example](https://stackblitz.com/edit/github-ojxahh?file=src%2Fpages%2Findex.astro) to Astro 3.0.12 and unfortunately, it does not resolve the issue.

Thanks for looking into the issue @matthewp

martrapp:
[...]

Issue #7761, opened 2023-07-21
Scripts and Styles in components included inside mdx content, loaded through another component, don't work


trafnar:
### What version of `astro` are you using?

2.9.1

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I'm looping through a content collection, using a component to render each content item. The content items themselves are defined in .mdx files, each with a component inside the mdx. That component does not get styling/scripts applied.

[Reproducible example Stackblitz project](https://stackblitz.com/edit/github-u2yta7?file=src%2Fpages%2Findex.astro)

<details><summary>Basic Example of what's in the linked project</summary>
<p>

index.astro (some imports removed for brevity)
```astro
---
const posts = (await getCollection("posts"));
---
{posts.map((post) => <Post post={post} />)}
```

components/Post.astro
```astro
---
const { post } = Astro.props;
const { Content } = await post.render();
---

<h2>
  {post.data.title}
</h2>

<Content />
```

components/Card.astro
```astro
<div class="card">This is a card</div>
<style>
.card{background-color:yellow; border:1px solid red;}
</style>
```

content/posts/my-post.mdx
```mdx
---
title: "My Post"
---
import Card from '../../components/Card.astro';
<Card />
```

The `<Card />` used in `my-post.mdx` will not have styling applied

</p>
</details> 

### What's the expected result?

The `<Card />` component should have it's local styles applied.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-u2yta7?file=src%2Fpages%2Findex.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Thank you for the clean reproduction, it was really helpful!

I'll let one of the other maintainers confirm, but this seems to be an architectural limitation.

Due to Astro's streaming, we only wait until the page's frontmatter has run before generating the `<head>` element. Frontmatter of a component (in your case, `components/Post.astro`) cannot add scripts or stylesheets because by the time it's run, the `<head>` is already sent to the browser so that it can start loading assets.

In case of SSG, it shouldn't matter, but the same architecture is used for both modes.

ChrisPlease:
Just started exploring Astro yesterday and ran into this today. My component is a 1 off, so as a work around I set `<style is:inline>`, but I can see how that's not useful in a loop.

kurtextrem:
[...]

PR #8449, opened 2023-09-07
Fix multi-layout head injection

matthewp:
## Changes

- How head propagation works in a nutshell:
  1. Every component that has head content to propagate has special metadata.
  2. When we begin to render a page we call `instance.init(result)`.
  3. Components with this metadata add it to the result.
  4. When we buffer head content we wait for each of these `init`s to complete. We do it recursively so that if they have children with head content those are also awaited.
  5. When there's nothing left to wait on, head content is done and rendering begins.
- The bug is that we were not calling `init()` until (5), so it trivially was working for single children, but not for deep children in layouts.
- The fix is to call init when an Astro template runs so that it can register itself.

- Fixes https://github.com/withastro/astro/issues/8399
- Fixes https://github.com/withastro/astro/issues/8392
- Fixes https://github.com/withastro/astro/issues/7761

## Testing

- Updated an existing test case but went deeper to add an extra layer of layout which triggers the need to call `renderAstroComponent` which is where the bug is.

## Docs

N/A, bug fix.

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -421,6 +421,9 @@ function renderAstroComponent(
 	slots: any = {}
 ): RenderInstance {
 	const instance = createAstroComponentInstance(result, displayName, Component, props, slots);
+	// Initialize the instance early so that child components with head propagation
+	// are able to register themselves.
+	instance.init(result);
 	return {
 		async render(destination) {
 			// NOTE: This render call can't be pre-invoked outside of this function as it'll also initialize the slots
--- a/.changeset/orange-cheetahs-happen.md
+++ b/.changeset/orange-cheetahs-happen.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix multi-layout head injection
--- a/packages/astro/test/units/dev/head-injection.test.js
+++ b/packages/astro/test/units/dev/head-injection.test.js
@@ -123,27 +123,38 @@ describe('head injection', () => {
 						});
 					}
 				`.trim(),
+				'/src/components/Content.astro': `
+				---
+				import { renderEntry } from '../common/head.js';
+				const ExtraHead = renderEntry();
+				---
+				<ExtraHead />
+				`,
+				'/src/components/Inner.astro': `
+				---
+				import Content from './Content.astro';
+				---
+				<Content />
+				`,
 				'/src/components/Layout.astro': `
-					---
-					import { renderEntry } from '../common/head.js';
-					const ExtraHead = renderEntry();
-					---
 					<html>
 						<head>
 							<title>Normal head stuff</title>
 						</head>
 						<body>
 							<slot name="title" />
-							<ExtraHead />
+							<slot name="inner" />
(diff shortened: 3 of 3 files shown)


natemoo-re (approved):

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -421,6 +421,9 @@ function renderAstroComponent(
 	slots: any = {}
 ): RenderInstance {
 	const instance = createAstroComponentInstance(result, displayName, Component, props, slots);
+	// Initialize the instance early so that child components with head propagation
+	// are able to register themselves.
+	instance.init(result);
```
bluwy, line 426: Interesting, I thought we were already calling it in `bufferHeadContent`:

https://github.com/withastro/astro/blob/0fa483283e54c94f173838cd558dc0dbdd11e699/packages/astro/src/runtime/server/render/astro/render.ts#L152-L167

which the propagators is added during component instance creation too:

https://github.com/withastro/astro/blob/0fa483283e54c94f173838cd558dc0dbdd11e699/packages/astro/src/runtime/server/render/astro/instance.ts#L77-L89

Could it be that the component we're rendering failed the `isAPropagatingComponent()` check?

natemoo-re:
That failing test looks like this change switches the rendering order again... I'm not sure if this is solvable without eagerly evaluating slots, rendering bottom-up rather than top-down.

That being said, there are top-down rendering is something we've really tried to stick with because lazy slot evaluation has some very nice implications for conditional logic. Folks _are_ relying on our current behavior, unfortunately.

packages/astro/src/runtime/server/render/component.ts
matthewp: Ah actually that's a good point. The parent should have marked as `'in-tree'`. I'll check on this.

matthewp:
Putting back to draft as I think there's another possible solution here.

lilnasy:
Do you mind if I add some tests. I made some a while ago based on #7761's repro.

packages/astro/src/runtime/server/render/component.ts
lilnasy, line 426: I am assuming this changes how long the server runtime waits before starting to stream <body>. It would be indented for cases where scripts and styles got missed, but could it change the timing for other users as well?

matthewp:
!preview propagation-metadata

matthewp:
!preview propagation-metadata

packages/astro/src/runtime/server/render/component.ts
matthewp: Check out the new changes.

matthewp: Ok, turns out that we were not bubbling up `propagation` that came through the compiler. In that case we don't need to early call to init.

lilnasy:
It does not seem to fix #7761.

matthewp:
@lilnasy thanks, removing that one from the list. Once I made the change to the compiler, this really only targets VT issues. That one might be similar, however.

natemoo-re (approved):

PR #8456, opened 2023-09-07
Improve `create-astro` startup performance

natemoo-re:
## Changes

- Avoids blocking `create-astro` startup on async network requests
- Adds a `console.clear()` call at start because PNPM always spits out a bunch of dependencies during `pnpm create astro`.
- First Houston now clears after greeting to keep the UI clear
- Minor change for safety. Also bumps minimum Node version to 16 (not 18 yet just to be safe)

## Testing

Tested manually. Existing test suite should work!

**Before** `time pnpm create astro` 2.03s to first message
**After** `time pnpm create astro@next--perf` 1.32s to first message

## Docs

N/A


natemoo-re:
!preview perf

natemoo-re:
!preview perf

bluwy (approved):
I have a slight preference to not clear the console for pnpm and the houston message as it makes the CLI jump around and potentially hides important information. But not a blocker for this PR.

packages/create-astro/src/index.ts
```diff
@@ -19,6 +19,8 @@ process.on('SIGTERM', exit);
 // https://github.com/withastro/docs/blob/main/src/pages/en/install/auto.md
 // if you make any changes to the flow or wording here.
 export async function main() {
+	// Clear console because PNPM startup is super ugly
+	console.clear();
```
ematipico, line 23: Does this contain important information for the user?

ematipico (approved):

packages/create-astro/src/index.ts
natemoo-re: In my experience no, it just logs out a stream while downloading packages. We'll only be able to clear if everything successfully downloaded and `create-astro` is running, so not a risk to clear IMO.

FredKSchott:
I've definitely seen users report losing something in their history that they needed or wanted from a console.clear() (IIRC snowpack used to do this in it's dev server output and we removed it for this reason). The subjective value of clearing out the npm/pnpm/yarn install noise doesn't outweigh the objective risk of losing useful data from the user's terminal history.

Similar to others this isn't a huge deal / something that will break immediately, but having gone through the full process of adding and removing something like this before I'd definitely recommend reverting the `console.clear()`. You can solve equally well IMO with a few newlines instead to separate the clean, beautiful `create-astro` output from the dirty package manager garbage :)

PR #8459, opened 2023-09-08
fix(cloudflare): added config for _routes.json generation

schummar:
## Changes
Follow up to #7846 

- Adds options to switch between the different strategies when generating `_routes.json`. By default the strategy producing the smaller file will be chosen. If users want more control over it, they can override it by setting `routes.strategy` to `"include"` (current default behavior) or `"exclude"` (old default behavior).
- Additional options to add manual `include` and `exclude` entries.

```diff
export default defineConfig({
    adapter: cloudflare({
        mode: 'directory',
+       routes: {
+           strategy: 'include',
+           include: ['/foo'],
+           exclude: ['/bar'],
+       },
    }),
});
```

## Testing

Added test cases ✅ 

## Docs

Added docs ✅ 


packages/integrations/cloudflare/test/routesJson.js
alexanderniebuhr, line 1: Could we rename this file to `routesJson.test.js` to align with all the other test files in the adapter?

schummar: Of course, done.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -373,39 +386,63 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 					staticPathList.push(...routes.filter((r) => r.type === 'redirect').map((r) => r.route));
 
+					const strategy = args?.routes?.strategy ?? 'auto';
+
 					// In order to product the shortest list of patterns, we first try to
 					// include all function endpoints, and then exclude all static paths
```
ematipico, line 392: I believe this comment is no longer valid, since the logic is changed

packages/integrations/cloudflare/src/index.ts
```diff
@@ -373,39 +386,63 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 
 					// Cloudflare requires at least one include pattern:
 					// https://developers.cloudflare.com/pages/platform/functions/routing/#limits
 					// So we add a pattern that we immediately exclude again
-					if (include.length === 0) {
-						include = ['/'];
-						exclude = ['/'];
+					if (includeStrategy?.include.length === 0) {
+						includeStrategy.include = ['/'];
+						includeStrategy.exclude = ['/'];
```
ematipico, line 416: Why do we override `exclude`?

packages/integrations/cloudflare/src/index.ts
```diff
@@ -373,39 +386,63 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 					}
 
 					// If using only an exclude list would produce a shorter list of patterns,
 					// we use that instead
```
ematipico, line 420: I believe the comment is no longer valid?

packages/integrations/cloudflare/src/index.ts
```diff
@@ -373,39 +386,63 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+					const excludeStrategy =
+						strategy === 'include'
+							? undefined
+							: {
+									include: ['/*'],
+									exclude: deduplicatePatterns(staticPathList.concat(args?.routes?.exclude ?? [])),
+							  };
+
+					const includeStrategyLength = includeStrategy
+						? includeStrategy.include.length + includeStrategy.exclude.length
```
ematipico, line 430: Why is the include strategy a sum of `include` and `exclude`?

packages/integrations/cloudflare/src/index.ts
```diff
@@ -373,39 +386,63 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+						: Infinity;
+
+					const excludeStrategyLength = excludeStrategy
```
ematipico, line 433: Why is the exclude strategy a sum of `include` and `exclude`?

packages/integrations/cloudflare/src/index.ts
schummar: I have updated the comment.

packages/integrations/cloudflare/src/index.ts
schummar: I have updated the comment.

packages/integrations/cloudflare/src/index.ts
schummar: Since Cloudflare does not allow an empty `include` array. Therefore, we add `'/'` and immediately remove it again. The result is the same as having both array empty, but it prevents the Cloudflare deployment error.

packages/integrations/cloudflare/src/index.ts
schummar: Both strategies produce both `include` and `exclude` entries. The difference is:
- One produces primarily `include` entries and `exclude` only where needed to resolve conflicts.
- The other produces primarily `exclude` entries and `include` only one wildcard entry.

Cloudflare's limit of 100 entries concerns both arrays combined. So, to choose the best strategy, we add the lengths for each strategy and select the one with the lesser number.

packages/integrations/cloudflare/src/index.ts
schummar: See the previous conversation.

packages/integrations/cloudflare/src/index.ts
ematipico: Thank you! Maybe it's worth adding a comment so that future people will know that :)

ematipico: Oh sorry, that's great!

.changeset/famous-seas-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+added config for `_routes.json` generation
```
sarah11918, line 5:
```suggestion
Adds a new config option for `_routes.json` generation: `routes.strategy`.
```
I personally like listing the new option here! Will make it easy for people to know what to look for if they're reading the changelog!

sarah11918 (commented):
Thanks for providing this documentation @schummar! I left a comment re: the changelog, but I have more questions regarding the README before I can continue reviewing.

I'm confused about whether the 2nd and 3rd items are in fact the options for `routes.strategy`. Would that not make them e.g. `routes.strategy.include` and `routes.strategy.exclude`? Or, are they in fact other `routes` configurations (so you're adding three new configurations to `routes` )?

Because they are named identically, it is easy to wonder whether these are in fact the same thing, so a code configuration example would be helpful. Can you show an example of a configuration here? And then, I'll review the README!  🙌

sarah11918 (commented):
OK, @schummar , I found your code example and included it!  😄 

Also, please see my comments re: documenting the config options. I'd also love for @alexanderniebuhr and @ematipico to take a look at my comments and provide guidance here!

(Admittedly, the details go beyond my direct knowledge and experience, but my role is to massage the docs to be as consistent as possible, and as accessible as possible to a wide audience and range of experience levels! So thank you in advance for helping me do that!  🙌 )

packages/integrations/cloudflare/README.md
```diff
@@ -75,6 +75,36 @@ export default defineConfig({
...
+If you want to use the automatic `_routes.json` generation, but want to include additional routes (e.g. when having custom functions in the `functions` folder), you can use the `routes.include` option to add additional routes to the `include` array.
+
+### routes.exclude
+
+`routes.exclude: string[]`
+
+default `[]`
+
+If you want to use the automatic `_routes.json` generation, but want to exclude additional routes, you can use the `routes.exclude` option to add additional routes to the `exclude` array.
+
```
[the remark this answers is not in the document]
alexanderniebuhr: I've only skimmed through the implementation and haven't delved into the details, but when using mode: 'directory', the functions folder (containing custom-written functions) should be included by default. @schummar, is there a need for a user to specify this in the include array? If so, I think we should rethink the logic here.

alexanderniebuhr (commented):
I have a few suggestions to enhance this:

1. To maintain consistency, could we rename the test file as follows?
```
renamed:    packages/integrations/cloudflare/test/routes-json.js -> packages/integrations/cloudflare/test/routes-json.test.js
``` 


2. I'd love to see a variety of test cases, such as:

- Tests for mode: 'advanced'
- Tests for both include & exclude strategy using a custom function (they should target the same files but with different syntax)
- Tests for functionPerRoute: true
- A test incorporating assets
 
3. Lastly, I'm concerned that the include strategy might not address all edge cases. (During my manual testing, I encountered some issues. I'd be happy to discuss this on a Discord thread in case I missed something.)
If that's the case, I'd suggest not setting the default to auto. Perhaps making exclude the default would be more appropriate.

packages/integrations/cloudflare/README.md
```diff
@@ -75,6 +75,36 @@ export default defineConfig({
...
+default `"auto"`
+
+If no [custom `_routes.json`](#custom-_routesjson) is provided, `@astrojs/cloudflare` will generate one for you. There are two ways to generate the `_routes.json`:
+
+1. For each page or endpoint in your application that is not prerendered, an entry in the `include` array will be generated. For each page that is prerendered and whoose path is matched by an `include` entry, an entry in the `exclude` array will be generated.
+
+2. One `"/*"` entry in the `include` array will be generated. For each page that is prerendered, an entry in the `exclude` array will be generated.
+
+Setting `routes.strategy` to `"include"` will generate a `_routes.json` with the first strategy. Setting it to `"exclude"` will use the second strategy. Setting it to `"auto"` will use the strategy that generates the least amount of entries.
```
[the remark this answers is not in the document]
alexanderniebuhr: I agree. Even with my current understanding of Cloudflare and its implementation, I believe it would be beneficial to have some example-based, easy-to-understand content.

sarah11918 (dismissed):
Just putting this here so I get notified when I should come back and look at updated documentation!  🙌

packages/integrations/cloudflare/README.md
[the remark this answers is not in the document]
schummar: I have updated the example. There is indeed no need to reference the functions folder, but rather custom functions withing the functions folder that were not generated by Astro. I hope the updated example clarifies this.

schummar:
I have made some update but did not address all feedback yet. Will continue later, I'll notify you, when I'm done.

packages/integrations/cloudflare/README.md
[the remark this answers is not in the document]
alexanderniebuhr: Alright, I think we can improve the adapter to even include custom functions in a future iteration. :)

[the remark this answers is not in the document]
schummar: To some degree, yes. They can have dynamic parts, like Astro pages too. So it's not strictly possible to know all correct routes from just looking at the files. One could however define, that static assets always beat dynamic path parts - again like Astro pages - and let users override `include` and `exclude` to implement anything more specific.

packages/integrations/cloudflare/README.md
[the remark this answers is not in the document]
schummar: Yes, understood. It's kind of hard to separate implementation from result here because caring about the contents of the `_routes.json` file is already rather low level. I've made another attempt with less details but more examples. Enough or should I slim it down even more?

schummar:
> I have a few suggestions to enhance this:
> 
> 1. To maintain consistency, could we rename the test file as follows?
> 
> ```
> renamed:    packages/integrations/cloudflare/test/routes-json.js -> packages/integrations/cloudflare/test/routes-json.test.js
> ```
> 

Sorry for missing the ".test" part. Maybe third time's the charm 😁. I did the same adjustment for the test's fixtures, renaming the path from `routesJson` to `routes-json`.


> 2. I'd love to see a variety of test cases _(non-blocking for me)_, such as:
> 
> * Tests for mode: 'advanced'

Done. Executing all test cases for both modes now.

> * Tests for both include & exclude strategy using a custom function (they should target the same files but with different syntax)

Is this not already covered by the "with additional `include` entries"/"with additional `exclude` entries" test cases? Files do not really play a role here. The content of the `functions` folder is not considered at all, the user supplied values are simply added to  the `include`/`exclude` arrays.

> * Tests for functionPerRoute: true

Do you expect that option has any impact on the resulting `_routes.json` file? It should only influence the content of the `functions` folder and that does not play a role here.

> * A test incorporating assets

I've added an asset file as well as a `_redirects` file to the fixtures' `public` folder.


> 3. Lastly, I'm concerned that the include strategy might not address all edge cases. (During my manual testing, I encountered some issues. I'd be happy to discuss this on a Discord thread in case I missed something.)
>    If that's the case, I'd suggest not setting the default to auto. Perhaps making exclude the default would be more appropriate.
> 
> **UPDATE:** Please ignore point 3 for now. I believe I've discovered an existing bug that I need to address. I'm still in the process of evaluating it.

That is still an interesting question: Defaulting to the `exclude` strategy might be safer. However when having mostly static pages and only a few dynamic ones the `include`/`auto` strategies usually produce a lot better `_routes.json` files. As far as I know this is the most common use case for Astro, right? People will run into the 100 entries limit very frequently.

alexanderniebuhr:
> Is this not already covered by the "with additional include entries"/"with additional exclude entries" test cases? Files do not really play a role here. The content of the functions folder is not considered at all, the user supplied values are simply added to the include/exclude arrays.

I believe that users shouldn't be required to add custom functions manually. Nonetheless, this discussion is beyond the scope of this PR. It looks good to me!

> Do you expect that option has any impact on the resulting _routes.json file? It should only influence the content of the functions folder and that does not play a role here.

Could you clarify this point? I'm uncertain why the `functions` folder is deemed irrelevant. The `functionPerRoute` will generate multiple files within the `functions` folder, as opposed to a singular `[[path]].js`. It's essential that these files are processed automatically to ensure accurate output for both modes _(include/exclude)_ in the `_routes.json`.

> As far as I know this is the most common use case for Astro, right? People will run into the 100 entries limit very frequently.

Indeed, the current default seems appropriate. However, if we start receiving feedback or reports, we can consider adjusting the default settings in the future.

schummar:
> Could you clarify this point? I'm uncertain why the `functions` folder is deemed irrelevant. The `functionPerRoute` will generate multiple files within the `functions` folder, as opposed to a singular `[[path]].js`. It's essential that these files are processed automatically to ensure accurate output for both modes _(include/exclude)_ in the `_routes.json`.

What I mean is, the content of the `functions` folder is not take into account when generating the `_routes.json` file. As far as I know, `functionPerRoute` also does not change the content of the `routes` and `pages` variables. However, since I don't understand all implications, maybe it's better to just test it as well. Better safe than sorry. => I have adjusted it now.

When a future PR indroduces analzying custom functions in the `functions` folder, testing with different `functionPerRoute` settings will become essential anyway.

alexanderniebuhr:
> does not change the content of the `routes` and `pages` variables

not sure, but at least you should be able to rely on them. But thanks for adding tests :)

schummar:
> Just putting this here so I get notified when I should come back and look at updated documentation! 🙌

@sarah11918 are the docs ok now?

packages/integrations/cloudflare/README.md
```diff
@@ -75,6 +75,92 @@ export default defineConfig({
...
+
+### routes.exclude
+
+`routes.exclude: string[]`
+
+default `[]`
+
+If you want to use the automatic `_routes.json` generation, but want to exclude additional routes, you can use the `routes.exclude` option to add additional routes to the `exclude` array.
+
+The following example automatically generates `_routes.json` while including and excluding additional routes. Note that that is only necessary if you have custom functions in the `functions` folder that are not handled by Astro.
```
sarah11918, line 148:
```suggestion
The following example automatically generates `_routes.json` using the `include` route generation strategy while additionally including and excluding additional routes specified directly in configuration. Note that that is only necessary if you have custom functions in the `functions` folder that are not handled by Astro.
```
The unfortunate duplication of naming makes this a little bit tricky to get just right, but I think adding some extra explicitness can help here.

Just to super clarify, the includes and excludes you've specified would not be handled AT ALL by the `include` route generation strategy? If they would, does mentioning them here override what happens during auto route generation, and should we say that this is the purpose? I want to make sure this example is demonstrating something people might actually need to do, and that it's clear why stating these includes and excludes are necessary when they're already auto-generating routes.

sarah11918 (commented):
Thank you @schummar , the description of the three options is so much more user-friendly! I appreciate the extra attention there! Just one tiny comment on the separate include and exclude options for clarification, but these docs are looking great, so I've removed the blocking review.

As soon as everyone here is confident my comment below is satisfactorily represented in the docs, this will be good to go!

packages/integrations/cloudflare/src/index.ts
```diff
@@ -386,6 +399,10 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 				// move cloudflare specific files to the root
 				const cloudflareSpecialFiles = ['_headers', '_redirects', '_routes.json'];
+				if (!isModeDirectory) {
+					cloudflareSpecialFiles.push('_worker.js');
+				}
```
alexanderniebuhr, line 404: Not blocking comment, however I'm not sure why we need this here? `_worker.js` should be at the root already.

packages/integrations/cloudflare/README.md
alexanderniebuhr: I'm happy with the current version, & don't think we need to improve it further for now.

alexanderniebuhr (approved):
Great contribution. LGTM!

ematipico (approved):

packages/integrations/cloudflare/src/index.ts
schummar: That is so that the _worker.js file does not make it into the `_routes.json`. It would be handled like a static asset otherwise. You are right, that trying to copy it to the root does not make sense. The error is ignored, but still it's not good practice. I've moved the block slightly, so that it does not affect the copying of files and added a comment.

PR #8463, opened 2023-09-08
doc: usage trend

cwtuan:
I am reaching out to request a small addition to the README file of the Astro family packages repository.

As an Astro.js user, I believe it would greatly benefit the community if we could include a link to the npm usage trend for the Astro.js family packages. Including this link in the README will provide a quick and convenient way for users to access this valuable information and stay updated with the Astro.js ecosystem.

Please consider this request, and if you find it reasonable. Thank you for considering my suggestion.


Princesseuh:
Hello! Thank you for this PR, however we'd prefer to avoid linking to such websites that might disappear randomly. We also don't really consider this information to be necessary super valuable + it's easy to find by yourself.

Thank you nonetheless!

PR #8464, opened 2023-09-08
feat(types): Add types for the object form of style

Princesseuh:
## Changes

What the title says! It shows both the camelCase version and the kebab-case version, since we support both. Users can choose their favorite, it's cool

image

## Testing

Tested manually

## Docs

Shouldn't need anything normally. There's a new export in `astro/types`, but we don't document all of our exports exhaustively tbh


natemoo-re (approved):
Nice!

PR #8468, opened 2023-09-08
[MDX] Support `img` component prop for optimized images

### LINKED ISSUE #7223: `img` MDX component override is not used in Astro collections

the issue comes from `@astrojs/mdx@0.19+`

Vercel

npm

Mac

Chrome

### Describe the Bug

When providing [custom components](https://docs.astro.build/en/guides/markdown-content/#assigning-custom-components-to-html-elements) for MDX file in content collections, `alt` will always be replaced by `astro:assets` `Image` component, hence not allowing to style it or customise the behaviour (for example, rendering the caption). The regression seems to come in some of [`@astrojs/mdx` updates](https://github.com/withastro/astro/pull/6824), reverting to `0.18` helps to mitigate the issue.

```js
export const components = {
  p: (props) => <p  {...props} class="custom-p" />
  img: (props) => <img {...props} class="custom-style" />
}
```

In `.mdx`:

```mdx
---
title: New post
---

export { components } from '../components';

![alt for image](assets/image.jpg)
```

`p` will be correctly picked up from `components`, but `img` will be ignored and `astro` image used instead.
Ideally, `alt` would be parsed as a local image, so `props` include `src`, `alt`, `width`, `height`, etc. So that they can be passed to `astro:assets/Image`

https://codesandbox.io/p/sandbox/eager-pascal-b30gwz?file=%2Fsrc%2Fpages%2Findex.astro%3A14%2C1

### Participation

bholmesdev:
## Changes

Respect the `img` component prop / export for images optimized via `astro:assets`. This will pass an `ImageMetadata` object as the `src` property to any `img` component when an optimized image is used.

- Resolves #7223
- Introduce an `'astro-image'` custom element to map via the `components` prop
- Check for user components passed as props or via an export before applying the Astro image optimizer

## Testing

- test `components` property as export, esm import, and content collection entry

## Docs

- docs draft in CHANGELOG
- official docs still TODO until we decide we like this!


bholmesdev:
I'm debating if this is a major, since it will now pass ImageMetadata for optimized images (see CHANGELOG). Could be breaking, or just a "new feature" since we didn't respect the img export at all before this change 🤔

bholmesdev:
I'm debating if this is a major, since it will now pass ImageMetadata for optimized images ([see changeset](https://github.com/withastro/astro/blob/fdd2b48eced71bb90eb393078201689bf312a7d3/.changeset/cyan-penguins-divide.md)). Could be breaking, or just a "new feature" since we didn't respect the `img` export at all before this change 🤔

Princesseuh (approved):
You went much further than I would've had I implemented this myself, awesome work!

Issue #8114, opened 2023-08-17
client:only ui component loses its style in a persistent island


abernier:
### What version of `astro` are you using?

2.10.9

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac, Linux

### What browser are you using?

Chrome

### Describe the Bug

I have a `transition:persist`ent island. In that island, I get a `client:only` React component.

screencast: https://storage.googleapis.com/abernier-portfolio/flyme3.mov

When I navigate to a new page:
- ✅ island is persisted (my React component not rerendered -- that's fine)
- ❌ BUT I lose the style imported from my React component

since the Player react-component is not re-rendered, and the island persistent (`<marquee>` proves it), it's weird my `import`ed style gets touched in the head...

NB: it only happens when `client:only`

### What's the expected result?

the client UI style should be untouched across navigation

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-l21t7u?file=src%2Flayouts%2FLayout.astro,src%2Fpages%2Fmoon.astro,src%2Fcomponents%2FPlayer.tsx,src%2Fpages%2Findex.astro,src%2Fcomponents%2Fplayer.css&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

erixweb:
Hi. I've been looking at the reproducible example and what I did is change from client:only to **client:load.** This has worked for me.

abernier:
> If you want to use client:only, documentation marks you must specify the correct framework for it to work.

I've [tested with `client:only="react"`](https://stackblitz.com/edit/github-l21t7u-5t1tam?file=src%2Flayouts%2FLayout.astro%3AL29,src%2Fpages%2Fmoon.astro,src%2Fcomponents%2FPlayer.tsx,src%2Fpages%2Findex.astro,src%2Fcomponents%2Fplayer.css&on=stackblitz) but same behaviour

martrapp:
Preliminary analysis:
There is a <style> element in the <head> of the page that contains the styling of the player.
During swap(), it is removed because it is not part of the target page.
The <style> element is not part of either the "from" or the "to" document as sent by the server. (client:only)
It is injected into the <head> by Vite's client lib (@vite/client:updateStyle()).
I suspect this happens only once for the original page, since the island is persistent.

matthewp:
I love this demo btw 😀

matthewp:
[...]

PR #8472, opened 2023-09-08
Prevent style removal of client:only components

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/8114
- This is a dev only problem. Because client:only styles are injected only when the island runs and will not rerun when inside of a persisted island.
- Solution is to not remove these style tags. This is dev-only code and will be tree-shaken.

## Testing

- New test case added. Styles should have the same count after transitioning to a new page.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/grumpy-hotels-heal.md
+++ b/.changeset/grumpy-hotels-heal.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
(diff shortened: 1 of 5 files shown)


martrapp:
LGTM

PR #8492, opened 2023-09-10
fix(types): make `image.service` optional

xiBread:
## Changes

Reflect the change made in #8266.

## Testing

Types changed only.

## Docs

N/A


Princesseuh (approved):
Looks great. Thank you!

PR #8494, opened 2023-09-11
Clear caches older than 5 days

bluwy:
## Changes

Spent some time figuring it out, looks like there's an action we can use to do this.

## Testing

n/a. I think it should work.

## Docs

n/a.

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,19 +1,28 @@
 name: Cleanup cache
 
 on:
+  schedule:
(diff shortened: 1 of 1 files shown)


ematipico (approved):
Only 5? 🫣

bluwy:
Not sure if you mean too high or too less 😄 I think it will just enough keep us under the limit.

bluwy:
I believe the cache key is based on the lockfile, so updating tests doesn't quite affect it. We can try 3 days if we're still seeing full caches though.

PR #8502, opened 2023-09-11
Use shikiji

bluwy:
## Changes

Use shikiji instead of shiki.

WIP because:
1. `shikiji` doesn't support `setColorReplacements` for `css-variables` theme.
2. `shikiji` doesn't render default theme colors for unknown/plaintext languages.

## Testing

Existing tests should pass, except the above two (2 test cases)

## Docs

Most docs around shiki should also apply to shikiji, so I don't think we need to update the docs. The ideal case is to also have shikiji's improvments upstreamed to shiki.


packages/astro/components/Code.astro
```diff
@@ -47,41 +54,64 @@ const {
...
+}
+
+const highlighter = await getCachedHighlighter({
+	langs: [lang],
+	themes: [theme],
 });
 
-// 2. Turn code into shiki theme tokens
-const tokens = highlighter.codeToThemedTokens(code, typeof lang === 'string' ? lang : lang.id);
+const html = highlighter.codeToHtml(code, {
```
bluwy, line 74: I refactored the old code to `getCachedHighlighter` + `highlighter.codeToHtml`. `codeToHtml` is the same as `codeToThemeTokens` + `renderToHtml`.

`codeToHtml` uses the hast tree which I refactored to take in account of it.

packages/astro/components/Code.astro
```diff
@@ -47,41 +54,64 @@ const {
...
+			// theme.id for shiki -> shikiji compat
+			const themeName = typeof theme === 'string' ? theme : theme.name;
+			if (themeName === 'css-variables') {
+				// Replace special color tokens to CSS variables
+				visit(node as any, 'element', (child) => {
+					if (child.properties?.style) {
+						child.properties.style = replaceCssVariables(child.properties.style);
+					}
+				});
+			}
```
bluwy, line 114: At the moment, shikiji doesn't have good support for `css-variables` theme, largely because there's builtin light/dark theme support now. Maybe we can remove it in the next major.

packages/astro/src/core/build/plugins/plugin-internals.ts
```diff
@@ -16,7 +16,6 @@ export function vitePluginInternals(input: Set<string>, internals: BuildInternal
 						// Except for these packages as they're not bundle-friendly. Users with strict package installations
 						// need to manually install these themselves if they use the related features.
 						external: [
-							'shiki', // For syntax highlighting
```
bluwy, line 19: `shikiji` can be bundled, so we don't have to externalize it, and the user don't have to manually install themselves if using pnpm.

packages/astro/src/core/config/schema.ts
```diff
@@ -228,11 +235,30 @@ export const AstroConfigSchema = z.object({
...
+								if (typeof lang === 'object') {
+									// `id` renamed to `name
+									if ((lang as any).id && !lang.name) {
+										lang.name = (lang as any).id;
+									}
+									// `grammar` flattened to lang itself
+									if ((lang as any).grammar) {
+										Object.assign(lang, (lang as any).grammar);
+									}
+								}
```
bluwy, line 253: Applied the same compat as in the `Code.astro` component.

.changeset/few-peas-hunt.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/markdoc': minor
+'@astrojs/markdown-remark': minor
+'astro': minor
+---
+
+Update internal `shiki` syntax highlighter to `shikiji`. Existing docs on `shiki` still stays the same, but `shikiji` will help make bundling simpler and easier to maintain.
```
ematipico, line 7: Are there APIs that change? If not, maybe this should be a patch. If yes, then we should briefly document what's changed.

packages/astro/components/Code.astro
```diff
@@ -47,41 +54,64 @@ const {
...
+		pre(node) {
+			// Swap to `code` tag if inline
+			if (inline) {
+				node.tagName = 'code';
+			}
 
-// 3. Get shiki theme object
-const _theme = highlighter.getTheme();
+			const classValue = (node.properties.class as string) ?? '';
+			const styleValue = (node.properties.style as string) ?? '';
```
ematipico, line 85: Is the cast safe? If so, can we add a document saying why it's safe?

packages/astro/components/shiki.ts
```diff
@@ -0,0 +1,46 @@
+import { type Highlighter, getHighlighter } from 'shikiji';
```
ematipico, line 1: Maybe we should move this file inside the `internal-helpers` package. Its content seems to be the same as we have in the markdoc integration.

ematipico (approved):
It looks good to me. I wonder if we could move shared code inside the `internal-helpers` pacakge.

sarah11918:
With respect to docs, I would only expect we need to add docs to the site if:

- someone upgrades to 3.3 and suddenly their site looks different, without doing anything (this would be a breaking change, and should be in the changeset)
- someone has additional config options, or more options to existing config values, that they can set
- there is something new they can do now that they couldn't before (or the opposite, some behaviour has been removed)
- there is a different way to do something that they could previously do

Outside of these conditions, there's probably not much we would document? We don't tend to explain a lot of inner workings, but rather how to do things. If the user has to act differently, or will notice different behaviour even as the result of changing nothing, these are the two big flags for docs.

.changeset/few-peas-hunt.md
bluwy: The external APIs are unchanged, only the inner workings. I put `minor` since it feels like a feature, just that its not very obvious 🤔

packages/astro/components/shiki.ts
bluwy: Yeah I'm planning to do a second round refactor once this is merged. It's in 3 places now. `shikiji` simplifies the usage a bit so I thought of duplicating the code like before for now, then refactor it as one again.

bluwy:
@sarah11918 

1. The site should look the same. The themes are copied from `shiki` from `shikiji`.
2. The existing options stay the same. `shikiji` does have extra APIs, but it's not exposed to the user yet.
3. For now there's no new or removed features.
4. Existing patterns stay the same.

Probably the only visible change is the markup where `style="color: #ffffff"` is now `style="color:#ffffff"`.

And `<span class="line" style="color: #ffffff"` is now `<span class="line"` + `<code class="astro-code" style="color:#ffffff"`. It's a new strategy where `span.line` inherits the color from `code` instead. Maybe I can document this in the changeset, but I'm not really sure if it'll affect people a lot.

.changeset/few-peas-hunt.md
```diff
@@ -0,0 +1,31 @@
+---
+'@astrojs/markdoc': minor
+'@astrojs/markdown-remark': minor
+'astro': minor
+---
+
+Update the internal `shiki` syntax highlighter to `shikiji`, an ESM-focused alternative that simplifies bundling and maintenance. There are no new options and changes to how you author code blocks and syntax highlighting.
```
sarah11918, line 7:
```suggestion
Updates the internal `shiki` syntax highlighter to `shikiji`, an ESM-focused alternative that simplifies bundling and maintenance. 

There are no new options and no changes to how you author code blocks and syntax highlighting.
```

I'm now enforcing a standard verb tense here! Add your "s"s, everyone!  :smile: 

Also splitting this out thinking of the auto-generated blog post heading from the first line. So for minors, if there's a good one-line headline, let's keep those separated out.

sarah11918 (approved):
Looks great, @bluwy ! Just some very minor changes to the opening line, but I think this is more than helpful for describing the change!

PR #8518, opened 2023-09-12
Support AVIF input assets

Princesseuh:
## Changes

Same thing as https://github.com/withastro/astro/pull/8380, but as a minor

## Testing

PR added a test

## Docs

N/A, we already document that images are imported this way


.changeset/five-doors-love.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Support AVIF input assets
```
matthewp, line 5: Given this is a minor can you make this a little more flowery... almost as if it were going into a blog post.

Princesseuh: Done!

PR #8519, opened 2023-09-12
config: migrate us to moduleResolution: 'node16'

Princesseuh:
## Changes

Blocked by https://github.com/withastro/cli-kit/pull/22

What the title says. Normally this is no change at all for users, however for us this has many benefits:
- We won't ever forget the `.js` anymore 😅
- More accurate types for packages that are properly typed.

This PR also cleans up the `tsconfig.json` a bit, some packages were being built weirdly because they had a different `tsconfig.json`. Now everyone apart from Deno is using the same settings.

## Testing

Tests should all pass!

## Docs

N/A


packages/integrations/markdoc/src/html/transform/html-token-transform.ts
```diff
@@ -1,5 +1,6 @@
 import type { Tokenizer } from '@markdoc/markdoc';
 import { Parser } from 'htmlparser2';
+// @ts-expect-error This type isn't exported
 import type * as Token from 'markdown-it/lib/token';
```
Princesseuh, line 4: Couldn't find this type in the public API at all, it seems like the only reason we were able to get it previously was from a raw file path.

packages/integrations/preact/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import preact, { type PreactPluginOptions as VitePreactPluginOptions } from '@preact/preset-vite';
+import { preact, type PreactPluginOptions as VitePreactPluginOptions } from '@preact/preset-vite';
```
Princesseuh, line 1: Removed the default export here, the types were annoying. Importing it like this works, the tests pass and the types are much easier

packages/integrations/preact/src/server.ts
```diff
@@ -1,10 +1,10 @@
 import type { AstroComponentMetadata } from 'astro';
 import { Component as BaseComponent, h, type VNode } from 'preact';
-import render from 'preact-render-to-string';
+import { render } from 'preact-render-to-string';
```
Princesseuh, line 3: Same thing here

packages/integrations/react/src/index.ts
```diff
@@ -7,6 +7,7 @@ export type ReactIntegrationOptions = Pick<ViteReactPluginOptions, 'include' | '
 	experimentalReactChildren?: boolean;
 };
 
+// @ts-expect-error
```
Princesseuh, line 10: The React Vite plugin only exports itself as a default export with a lying type to TypeScript's eye, so this doesn't work. It'd be easier if it exported itself normally, like the Preact plugin does.

packages/telemetry/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@astrojs/telemetry",
   "version": "3.0.1",
   "type": "module",
-  "types": "./dist/types/index.d.ts",
+  "types": "./dist/index.d.ts",
```
Princesseuh, line 5: I'm not sure why the telemetry package built itself this way, but it was the only package doing so. I changed it to work like all the other packages.

ematipico (approved):

Princesseuh:
> Why not nodenext? Also why not esnext for target?

Astro targets Node LTS, not latest. So we can't safely use esnext as a target. For example, we can't use `toSorted` since it doesn't exist in Node 18, however having the target at esnext would tell us that it's okay to use it and then we'd break all our Node 18 users.

bluwy (approved):
Awesome work!

PR #8522, opened 2023-09-12
Vt follow redirects

martrapp:
## Changes

With redirects, the location had to be updated to reflect the final destination in the browsers address bar.
Redirect errors, like same origin violation, are handled as any other errors of `fetch` by full load of the requested `href` 
(as same origin policy would prevent us from accessing the target DOM)

While we are looking at the response object now, I have also grabbed the content-type for the parser.

## Testing
two new e2e tests

## Docs

n/a bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -64,7 +64,9 @@ const { fallback = 'animate' } = Astro.props as Props;
 	};
 
 	async function getHTML(href: string) {
-		const res = await fetch(href);
+		const res = await fetch(href, {
+			redirect: 'manual'
+		});
 		const html = await res.text();
 		return { ok: res.ok, html };
 	}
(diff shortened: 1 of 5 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -280,13 +290,16 @@ const { fallback = 'animate' } = Astro.props as Props;
 	async function navigate(dir: Direction, loc: URL, state?: State) {
 		let finished: Promise<void>;
 		const href = loc.href;
-		const { html, ok } = await getHTML(href);
+		const { html, ok, contentType, redirected } = await getHTML(href);
+		// if there was a redirection, show the final URL in the browser's address bar
+		redirected && (loc = new URL(redirected));
 		// If there is a problem fetching the new page, just do an MPA navigation to it.
-		if (!ok) {
+		if (!ok || contentType !== 'text/html') {
```
matthewp, line 297: so nice we have this in this PR too!

martrapp: Not as clever as intended, see fix in PR #8537

PR #8523, opened 2023-09-12
Disallow node builtins in runtime

natemoo-re:
## Changes

- Updates our eslint settings to detect Node built-ins that accidentally make their way into our runtime code.
- Closes PLT-572
- Might catch a few issues in the future, but doesn't cover everything.

## Testing

Verified locally

## Docs

N/A


natemoo-re:
Got the #NWTWWHB for this on Discord, so I'm going to merge!

PR #8525, opened 2023-09-12
Add `@astrojs/upgrade` package for automatic package upgrades

natemoo-re:
## Changes

- Closes PLT-1214
- Adds a new CLI package named `@astrojs/upgrade`
- Usage is as follows:
  ```bash
  npx @astrojs/upgrade # upgrades to latest
  ```
  ```bash
  npx @astrojs/upgrade beta # upgrades to beta (or latest if beta isn't found)
  ```
  ```bash
  npx @astrojs/upgrade --dry-run # logs versions but no install
  ```
Terminal window that shows "Integration upgrade in progress..." for a mixed release

- For major or prereleases, automatically links to the migration guide for Astro (only if it exists) and the CHANGELOG for everything else.

Terminal window that shows "Integration upgrade in progress..." for a beta release

## Testing

Added some initial basic test suites in addition to lots of manual testing

## Docs

Updated the README! We should take another pass at this, though.


q1b:
Is this for hackathon @natemoo-re,
Can we contribute

jacobthesheep:
@natemoo-re I think it would also be great to either add a `@astrojs/uninstall` package or a CLI command to uninstall packages gracefully as currently, users have to remember to manually remove them from the Astro config before installing another Astro integration.

jasikpark:
I'm a big fan of storybook's upgrade script: https://storybook.js.org/docs/react/configure/upgrading#upgrade-script

It especially helps to reduce weird mismatched peer-deps!

https://github.com/antfu/taze is my default for general dep upgrades though: `pnpx taze@latest -I` is always in my zsh history 🙈

sarah11918 (commented):
So excited for this, Nate, and to be able to put this at the top of the Upgrade guide instead of "don't forget to update all your integrations , too!" 

Some tiny suggestions below for your consideration.

packages/upgrade/src/actions/install.ts
```diff
@@ -0,0 +1,125 @@
...
+import { color, say } from '@astrojs/cli-kit';
+import { pluralize, celebrations, done, error, info, log, spinner, success, upgrade, banner, title, changelog, warn, bye } from '../messages.js';
+import { shell } from '../shell.js';
+import { random, sleep } from '@astrojs/cli-kit/utils';
+
+export async function install(
+	ctx: Pick<Context, 'version' | 'packages' | 'packageManager' | 'prompt' | 'dryRun' | 'exit' | 'cwd'>
+) {
+	await banner();
+	log('')
```
ematipico, line 15: nit: I would create a function called `printHardLine()`, so we know that these lone `log("")` are intentional

packages/upgrade/src/actions/install.ts
```diff
@@ -0,0 +1,125 @@
...
+		await info(`${packageInfo.name}`, `is up to date on`, `v${tag}`)
+		await sleep(random(50, 150));
+	}
+	if (toInstall.length === 0 && !ctx.dryRun) {
+		log('')
+		await success(random(celebrations), random(done));
+		return;
+	}
+	const majors: PackageInfo[] = []
+	for (const packageInfo of [...dependencies, ...devDependencies].sort(sortPackages)) {
```
ematipico, line 29: Any reason why we create **another new** array, when we already have `toInstall`? Can we reason the same array somehow?

packages/upgrade/src/actions/install.ts
```diff
@@ -0,0 +1,125 @@
...
+
+function sortPackages(a: PackageInfo, b: PackageInfo): number {
+	if (a.isMajor && !b.isMajor) return 1;
+	if (b.isMajor && !a.isMajor) return -1;
+	if (a.name === 'astro') return -1;
+	if (b.name === 'astro') return 1;
+	if (a.name.startsWith('@astrojs') && !b.name.startsWith('@astrojs')) return -1;
+	if (b.name.startsWith('@astrojs') && !a.name.startsWith('@astrojs')) return 1;
+	return a.name.localeCompare(b.name);
+}
```
ematipico, line 89: Does this function **change** the order of the final `package.json` (in case there are some FS operations)?

packages/upgrade/src/actions/install.ts
```diff
@@ -0,0 +1,125 @@
...
+		while: async () => {
+			try {
+				if (dependencies.length > 0) {
+					await shell(ctx.packageManager, ['install', ...dependencies.map(({ name, targetVersion }) => `${name}@${(targetVersion).replace(/^\^/, '')}`)], { cwd, timeout: 90_000, stdio: 'ignore' })
+				}
+				if (devDependencies.length > 0) {
+					await shell(ctx.packageManager, ['install', '--save-dev', ...devDependencies.map(({ name, targetVersion }) => `${name}@${(targetVersion).replace(/^\^/, '')}`)], { cwd, timeout: 90_000, stdio: 'ignore' })
+				}
+			} catch {
+				const packages = [...dependencies, ...devDependencies].map(({ name, targetVersion }) => `${name}@${targetVersion}`).join(' ')
```
ematipico, line 107: Another case where we create a new array from an array we already have.

packages/upgrade/src/actions/install.ts
```diff
@@ -0,0 +1,125 @@
...
+				);
+				return ctx.exit(1);
+			}
+		},
+	});
+
+	await say([`${random(celebrations)} ${random(done)}`, random(bye)], { clear: false });
+}
+
+async function ensureYarnLock({ cwd }: { cwd: string }) {
```
ematipico, line 121: What does this function do? The code seems to write something on FS, but the name doesn't hint that. Also, why do we write a lock file? 🤔

packages/upgrade/src/actions/verify.ts
```diff
@@ -0,0 +1,153 @@
...
+		if (!packageInfo.targetVersion) {
+			return false;
+		}
+	}
+	return true;
+}
+
+async function resolveTargetVersion(packageInfo: PackageInfo, registry: string): Promise<void> {
+	const packageMetadata = await fetch(`${registry}/${packageInfo.name}`, { headers: { accept: 'application/vnd.npm.install-v1+json' }});
+	const { "dist-tags": distTags } = await packageMetadata.json();
```
ematipico, line 110: These two operations can fail, we should catch the error somewhere.

If we decide to handle the error inside the caller, we should call this function `tryResolveTargetVersion`.

Also, the `fetch` call can return a non-200 code, and we should check it first.

packages/upgrade/src/actions/verify.ts
```diff
@@ -0,0 +1,153 @@
...
+		if (packageInfo.name === 'astro') {
+			const upgradeGuide = `https://docs.astro.build/en/guides/upgrade-to/v${toVersion.major}/`;
+			const docsRes = await fetch(upgradeGuide);
+			if (docsRes.status === 200) {
+				packageInfo.changelogURL = upgradeGuide;
+				packageInfo.changelogTitle = `Upgrade to Astro v${toVersion.major}`;
+				return;
+			}
+		}
+		const latestMetadata = await fetch(`${registry}/${packageInfo.name}/latest`);
```
ematipico, line 138: Let's handle the status code of this call

matthewp (approved):

packages/upgrade/src/actions/install.ts
natemoo-re: Good feedback! Addressed in [377a543](https://github.com/withastro/astro/pull/8525/commits/377a543b9f5b62ffb08045b37f95326e0fbe535d)

packages/upgrade/src/actions/install.ts
natemoo-re: Nope, I think I just missed this in a refactor. Updated!

packages/upgrade/src/actions/install.ts
natemoo-re: Great question! Nope, this function is only for display purposes in this CLI. Package managers will respect the existing order of dependencies in the `package.json` file. Left a comment to this effect in [27d9409](https://github.com/withastro/astro/pull/8525/commits/27d940953b7e1638c81c86917906e60537edeaa8).

packages/upgrade/src/actions/install.ts
natemoo-re: I believe this one is actually necessary. This function handles `dependencies` and `devDependencies` separately but we're concatenating them into a single array for the error message.

packages/upgrade/src/actions/install.ts
natemoo-re: Added a comment in [f1a16c9](https://github.com/withastro/astro/pull/8525/commits/f1a16c9a99d1d6bb5c411db4dde596c1f590d5bc). This is needed for Yarn PnP support and was added to `create-astro` in https://github.com/withastro/astro/pull/8028

packages/upgrade/src/actions/verify.ts
natemoo-re: Good call. Refactored this to handle `>= 400` status codes in [203530f](https://github.com/withastro/astro/pull/8525/commits/203530febef74e9610f80f8cc484d45d6304ed71) and made sure that our existing error message was triggered if these calls fail.

PR #8533, opened 2023-09-13
Improve MDX rendering performance

bluwy:
## Changes

NOTE: this merges into `markdown-perf` branch. When https://github.com/withastro/astro/pull/8532 is merged, I'll rebase this then merge it.

- Improve markdown rendering performance by sharing the mdx processor.
- Refactor `recmaInjectImportMetaEnv` as its own file
- Refactor `rehypeApplyFrontmatterExport` as its own file

I tested it in the astro docs and there doesn't seem to be any improvements :( but this seems like a nice change overall otherwise.

## Testing

Ran astro & mdx tests locally

## Docs

n/a

### DIFF
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -1,6 +1,4 @@
-import { markdownConfigDefaults } from '@astrojs/markdown-remark';
-import { toRemarkInitializeAstroData } from '@astrojs/markdown-remark/dist/internal.js';
-import { compile as mdxCompile, type CompileOptions } from '@mdx-js/mdx';
+import { markdownConfigDefaults, setVfileFrontmatter } from '@astrojs/markdown-remark';
 import type { PluggableList } from '@mdx-js/mdx/lib/core.js';
 import type { AstroIntegration, ContentEntryType, HookParameters, SSRError } from 'astro';
 import astroJSXRenderer from 'astro/jsx/renderer.js';
@@ -8,10 +6,9 @@ import { parse as parseESM } from 'es-module-lexer';
 import fs from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
 import type { Options as RemarkRehypeOptions } from 'remark-rehype';
-import { SourceMapGenerator } from 'source-map';
 import { VFile } from 'vfile';
 import type { Plugin as VitePlugin } from 'vite';
-import { getRehypePlugins, getRemarkPlugins, recmaInjectImportMetaEnvPlugin } from './plugins.js';
+import { createMdxProcessor } from './plugins.js';
 import type { OptimizeOptions } from './rehype-optimize-static.js';
 import {
 	ASTRO_IMAGE_ELEMENT,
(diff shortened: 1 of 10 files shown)


packages/integrations/mdx/src/index.ts
```diff
@@ -1,17 +1,15 @@
 import { markdownConfigDefaults } from '@astrojs/markdown-remark';
-import { toRemarkInitializeAstroData } from '@astrojs/markdown-remark/dist/internal.js';
-import { compile as mdxCompile, type CompileOptions } from '@mdx-js/mdx';
+import { setAstroData } from '@astrojs/markdown-remark/dist/internal.js';
```
bluwy, line 2: Using `setAstroData` instead of `toRemarkInitializeAstroData` remark plugin helps avoid needing to re-create the processor everytime we render.

packages/integrations/mdx/src/index.ts
```diff
@@ -79,21 +77,7 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
...
-					jsxImportSource: 'astro',
-					// Note: disable `.md` (and other alternative extensions for markdown files like `.markdown`) support
-					format: 'mdx',
-					mdExtensions: [],
-				};
-
-				let importMetaEnv: Record<string, any> = {
-					SITE: config.site,
-				};
+				let processor: ReturnType<typeof createMdxProcessor>;
```
bluwy, line 80: All these options are moved to `createMdxProcessor`

packages/integrations/mdx/src/plugins.ts
```diff
@@ -4,101 +4,49 @@ import {
...
 import { rehypeInjectHeadingsExport } from './rehype-collect-headings.js';
 import rehypeMetaString from './rehype-meta-string.js';
 import { rehypeOptimizeStatic } from './rehype-optimize-static.js';
 import { remarkImageToComponent } from './remark-images-to-component.js';
-import { jsToTreeNode } from './utils.js';
 
 // Skip nonessential plugins during performance benchmark runs
 const isPerformanceBenchmark = Boolean(process.env.ASTRO_PERFORMANCE_BENCHMARK);
 
-export function recmaInjectImportMetaEnvPlugin({
```
bluwy, line 29: This plugin is moved to `recma-inject-import-meta-env.ts` as-is without changes

packages/integrations/mdx/src/plugins.ts
```diff
@@ -4,101 +4,49 @@ import {
...
-						raw: JSON.stringify(importMetaEnv[envVarName]),
-					};
-					Object.assign(node, envVarLiteral);
-				}
-			}
-		});
-	};
 }
 
-export function rehypeApplyFrontmatterExport() {
```
bluwy, line 56: This plugin is moved to `rehype-apply-frontmatter-export.ts` as-is without changes

packages/integrations/mdx/src/index.ts
matthewp: Why not expose this from the package?

bluwy: It still feels a bit internal-ish so I put it in `internal.js` without thinking much. Need a bit of time to think about the ideal API, which I can also do though 😬

matthewp: Yeah, I think we should either expose it as is, or design a better API, but not use internal methods. I think it's ok if it's not the ideal API, we can change it later.

bluwy: Aight I updated to export `setVfileFrontmatter` instead from the main export, and removed `setAstroData` in the latest force push rebased to `main`.

packages/integrations/mdx/src/rehype-apply-frontmatter-export.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { VFile } from 'vfile';
+import { jsToTreeNode } from './utils.js';
+
+export function rehypeApplyFrontmatterExport() {
+	return function (tree: any, vfile: VFile) {
+		const astroData = safelyGetAstroData(vfile.data);
+		if (astroData instanceof InvalidAstroDataError)
+			throw new Error(
+				// Copied from Astro core `errors-data`
+				// TODO: find way to import error data from core
```
ematipico, line 12: Now we have `AstroError`:

```js
import { AstroError } from "astro/errors";
```

ematipico (approved):
I'm not very familiar with this code. I just left one comment.

matthewp (approved):

packages/integrations/mdx/src/rehype-apply-frontmatter-export.ts
bluwy: I mostly copied the existing code over. We can follow-up a PR for this improvement 😬

PR #8536, opened 2023-09-13
feat: better errors for images

Princesseuh:
## Changes

Users makes some understandable mistakes around using `<Image />`:
- They pass filepaths, instead of imports
- They pass the result of a non-eager `import.meta.glob`

This PR adds better documentation and error messages around these cases to better inform users on the proper way to use `<Image />`. Hopefully it'll help!

This PR also includes a small styling change to the overlay, as informed by Mark.

## Testing

Tested manually the new messages, we have tests for errors otherwise already

## Docs

Some of this is docs!

### DIFF
--- a/packages/astro/src/assets/internal.ts
+++ b/packages/astro/src/assets/internal.ts
@@ -77,12 +77,12 @@ export async function getImage(
 
 	const service = await getConfiguredImageService();
 
-	// If the user inlined an import, something fairly common especially in MDX, await it for them
+	// If the user inlined an import, something fairly common especially in MDX, or passed a function that returns an Image, await it for them
 	const resolvedOptions: ImageTransform = {
 		...options,
 		src:
 			typeof options.src === 'object' && 'then' in options.src
-				? (await options.src).default
+				? (await options.src).default ?? (await options.src)
 				: options.src,
 	};
 
--- a/packages/astro/src/core/errors/dev/utils.ts
+++ b/packages/astro/src/core/errors/dev/utils.ts
@@ -227,21 +227,21 @@ export function getDocsForError(err: ErrorWithMetadata): string | undefined {
  * Render a subset of Markdown to HTML or a CLI output
  */
 export function renderErrorMarkdown(markdown: string, target: 'html' | 'cli') {
-	const linkRegex = /\[(.+)\]\((.+)\)/gm;
+	const linkRegex = /\[([^\[]+)\]\((.*)\)/gm;
 	const boldRegex = /\*\*(.+)\*\*/gm;
-	const urlRegex = /(\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|])/gim;
+	const urlRegex = / (\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|])/gim;
 	const codeRegex = /`([^`]+)`/gim;
 
 	if (target === 'html') {
 		return escape(markdown)
 			.replace(linkRegex, `<a href="$2" target="_blank">$1</a>`)
 			.replace(boldRegex, '<b>$1</b>')
-			.replace(urlRegex, ' <a href="$1" target="_blank">$1</a> ')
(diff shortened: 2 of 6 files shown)


packages/astro/src/assets/internal.ts
```diff
@@ -77,12 +77,12 @@ export async function getImage(
...
 	const service = await getConfiguredImageService();
 
-	// If the user inlined an import, something fairly common especially in MDX, await it for them
+	// If the user inlined an import, something fairly common especially in MDX, or passed a function that returns an Image, await it for them
 	const resolvedOptions: ImageTransform = {
 		...options,
 		src:
 			typeof options.src === 'object' && 'then' in options.src
-				? (await options.src).default
+				? (await options.src).default ?? (await options.src)
```
Princesseuh, line 85: This is a small behaviour change to support using non-eager `import.meta.glob` better. Saw some users confused that they couldn't pass the result of it to Image without awaiting multiple times and calling multiple functions in some cases. This was particularly relevant in MDX where you can't await everything the same way as in Astro files

packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -227,21 +227,21 @@ export function getDocsForError(err: ErrorWithMetadata): string | undefined {
  * Render a subset of Markdown to HTML or a CLI output
  */
 export function renderErrorMarkdown(markdown: string, target: 'html' | 'cli') {
-	const linkRegex = /\[(.+)\]\((.+)\)/gm;
+	const linkRegex = /\[([^\[]+)\]\((.*)\)/gm;
 	const boldRegex = /\*\*(.+)\*\*/gm;
-	const urlRegex = /(\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|])/gim;
+	const urlRegex = / (\b(https?|ftp):\/\/[-A-Z0-9+&@#\\/%?=~_|!:,.;]*[-A-Z0-9+&@#\\/%=~_|])/gim;
```
Princesseuh, line 232: A small regression fix in this regex. I changed it not too long ago and it broke some links, this fixes that

sarah11918 (approved):
Docs look good! Thanks @Princesseuh !

ematipico (approved):

natemoo-re (approved):

Marocco2:
This PR blocks production on my env as I rely on string based filepath for `og:image`

Minimum example:
```typescript
const { title, description, heroImage = "@assets/blog/introducing-astro.jpg" } = Astro.props;
const { src }= await getImage({
src: heroImage, width: 640, height: 426,
alt: ''
});
const socialImageURL = new URL(src, Astro.site);
```
```html
<meta property="og:image" content={socialImageURL} />
```

Princesseuh:
> This PR blocks production on my env as I rely on string based filepath for `og:image`
> 
> Minimum example:
> ```typescript
> const { title, description, heroImage = "@assets/blog/introducing-astro.jpg" } = Astro.props;
> const { src }= await getImage({
> src: heroImage, width: 640, height: 426,
> alt: ''
> });
> const socialImageURL = new URL(src, Astro.site);
> ```
> ```html
> <meta property="og:image" content={socialImageURL} />
> ```

How did this work before? Before this PR this would've ended up being an invalid path at build time anyway.

Marocco2:
> How did this work before? Before this PR this would've ended up being an invalid path at build time anyway.

Not at all. Basically I would pass `heroImage` from a frontmatter (for ex: "@assets/blog/hello-world.jpg"). `getImage()` would resolve the path and generate an optimized version of the image which I would use for `og:image`

Princesseuh:
> > How did this work before? Before this PR this would've ended up being an invalid path at build time anyway.
> 
> Not at all. Basically I would pass `heroImage` from a frontmatter (for ex: "@assets/blog/hello-world.jpg"). `getImage()` would resolve the path and generate an optimized version of the image which I would use for `og:image`

Hmm, `getImage` does not resolve images...

Marocco2:
Alright, I found a better way to deal with this stuff. Sorry for any inconvenience. Keep up the good work 👍

wddesousa:
> Alright, I found a better way to deal with this stuff. Sorry for any inconvenience. Keep up the good work 👍

can you tell me how? I'm struggling to making og:image work with cover images saved in the src folder but can't figure it out

PR #8537, opened 2023-09-13
Fix for checking media type in CSR

martrapp:
## Changes

Fix for https://discord.com/channels/830184174198718474/1151397066212388924/1151397066212388924
When checking the content type of the fetched document I confused media type with content type.
Fixed that

## Testing

Still passes all e2e tests :)

## Docs

n/a bug fix.


packages/astro/components/ViewTransitions.astro
```diff
@@ -65,19 +65,19 @@ const { fallback = 'animate' } = Astro.props as Props;
...
 		try {
-			res = await fetch(href);
+			const res = await fetch(href);
+			const html = await res.text();
+			return {
+				ok: res.ok,
+				html,
+				redirected: res.redirected ? res.url : undefined,
+				// drop potential charset (+ other name/value pairs) as parser needs the mediaType
+				mediaType: res.headers.get('content-type')?.replace(/;.*/, ''),
```
natemoo-re, line 76:
```suggestion
				mediaType: res.headers.get('content-type')?.replace(/;.*$/, ''),
```

Might just want to check that this replaces everything up to the end of the value? In practice, probably doesn't make much of a difference but it seems slightly more correct to me.

natemoo-re (approved):

natemoo-re (approved):
Awesome, thanks for the fix!

PR #8554, opened 2023-09-14
fix(create): update correct link for examples

Dunqing:
## Changes

- What does this change?

Update correct link


bluwy (commented):
I think we might have accidentally removed that vanity link. Updating https://github.com/withastro/astro.build/blob/main/vercel.json instead should fix it.

Princesseuh:
We fixed the link, so no need to change the link anymore.

PR #8560, opened 2023-09-14
feat(@astro/cloudflare): improve DX for runtime typing

alexanderniebuhr:
## Changes

- allow users to add env types using a generic, improving DX

## Testing

❌

## Docs

❌


matthewp (approved):

packages/integrations/cloudflare/src/server.advanced.ts
```diff
@@ -12,10 +12,10 @@ type Env = {
 	name: string;
 };
 
-export interface AdvancedRuntime {
+export interface AdvancedRuntime<T extends object> {
```
lilnasy, line 15: Because it's a patch release, maybe a default generic argument should be added.
```suggestion
export interface AdvancedRuntime<T extends object = {}> {
```

packages/integrations/cloudflare/src/server.directory.ts
```diff
@@ -7,10 +7,10 @@ if (!isNode) {
 	process.env = getProcessEnvProxy();
 }
 
-export interface DirectoryRuntime {
+export interface DirectoryRuntime<T extends object> {
```
lilnasy, line 10:
```suggestion
export interface DirectoryRuntime<T extends object = {}> {
```

lilnasy (approved):

sarah11918 (approved):
This looks good to me on behalf of docs! 

Make a note of the date: this is the rare time I'm going to wonder whether there is *too* much in the changeset, compared to the actual change in usage and documentation.  😅 

I would be tempted to say that just the line describing the change is enough, without the code snippet. (It's a minor change from existing README documentation, and the examples are clear there on how to type.)

But, I leave the final decision to Platform!

alexanderniebuhr:
I think I agree with @sarah11918 here.
This is also a `patch` change, so the smaller changeset should be fine. :)

PR #8563, opened 2023-09-15
Fix some typos and such [LOW PRIORITY]

Akul2010:
So far, I have only added the comment in. No need for tests because it just comments out the code (just learning JS, but I don't think that syntax was JS-like). I may add more typo and other things I notice, so I am marking this as draft for now.

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+// #!/usr/bin/env node
 export default async function run() {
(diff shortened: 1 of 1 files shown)


Princesseuh:
That's not a typo, it's a shebang. It's supposed to look like that: https://en.wikipedia.org/wiki/Shebang_%28Unix%29

Akul2010:
Oh, in that case, I will close it.

PR #8564, opened 2023-09-15
fix(@astrojs/cloudflare): support for nodejs_compat

### LINKED ISSUE #8520: Setting "nodejs_compat" for Cloudflare SSR API?
### Astro Info

```block
Astro                    v3.0.10
Node                     v20.5.1
Package Manager          npm
                         @astrojs/svelte
```

_No response_

### Describe the Bug

Hi guys, I'm running an API endpoint using the Cloudflare adaptor.
But I require the use of Buffer and hit the Buffer is not defined error.
I found that in Cloudflare documentation that I can run `compatibility_flags = [ "nodejs_compat" ]` and import Buffer this way `import { Buffer } from 'node:buffer'`, but I am unable to use this and build properly using Astro.
What am I missing or need to get this to work?

To be able to build and push to Cloudflare.

https://stackblitz.com/edit/github-5nhtag?file=src%2Fpages%2Fapi%2Fgenerate.ts

### Participation

alexanderniebuhr:
## Changes

- Allows import for supported [Node.js runtime APIs](https://developers.cloudflare.com/workers/runtime-apis/nodejs)

closes #8520 

## Docs


ematipico (commented):
Maybe we should list the node bindings that are supported in the documentation, plus we should highlight that only `node:*` is supported by the adapter, which means that importing `assert` will result in an error.

matthewp (approved):

.changeset/fifty-weeks-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Fix build issues for `node:*`, for supported runtime APIs
```
ematipico, line 5: It's a minor but we write as it was a patch 😅

ematipico (approved):
Looks good, just a small feedback on the changeset

.changeset/fifty-weeks-bake.md
alexanderniebuhr: Still waiting for docs, so can push an update. 🚀

.changeset/fifty-weeks-bake.md
```diff
@@ -0,0 +1,20 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Add support for following Node.js Runtime APIs, which are availabe in [Cloudflare](https://developers.cloudflare.com/workers/runtime-apis/nodejs) using the `node:` syntax.
```
sarah11918, line 5:
```suggestion
Add support for the following Node.js Runtime APIs, which are availabe in [Cloudflare](https://developers.cloudflare.com/workers/runtime-apis/nodejs) using the `node:` syntax.
```

packages/integrations/cloudflare/README.md
```diff
@@ -170,6 +170,21 @@ This will enable Cloudflare to serve files and process static redirects without
...
+- assert
+- AsyncLocalStorage
+- Buffer
+- Diagnostics Channel
+- EventEmitter
+- path
+- process
+- Streams
+- StringDecoder
+- util
```
sarah11918, line 186:
``````suggestion
Astro's Cloudflare adapter allows you to use the following Node.js runtime APIs (that are supported by Cloudflare):

- assert
- AsyncLocalStorage
- Buffer
- Diagnostics Channel
- EventEmitter
- path
- process
- Streams
- StringDecoder
- util

To use these APIs, your page or endpoint must be server-side rendered (not pre-rendered) and must use the the `import {} from 'node:*'` import syntax. 

```js
// put a file path for the endpoint here
export const prerender = false;
import { Buffer } from 'node:buffer';
```

Additionally, you'll need to enable the Compatibility Flag in Cloudflare. The configuration for this flag may vary based on where you deploy your Astro site. 

For detailed guidance, please refer to the [Cloudflare documentation](https://developers.cloudflare.com/workers/runtime-apis/nodejs).
``````
1. What would you think about a structure like this? (I always like my list to come directly after we've said, "the following list..."  😄 )

2. If I have the facts correct, then the first sentence can be tightened up a lot, too. (Is it really necessary to say that Cloudflare supports these? Is it sufficient to say that this is what the adapter does?) Another take on that sentence could be:

	> Use any Node.js runtime API supported by Cloudflare with the `@astrojs/cloudflare` adapter:
	
	I think the point you're getting at is that there's parity between the Node.js runtime APIs supported by Cloudflare and this adapter? If you really need to emphasize that, then this second sentence works well. If you don't, then just saying what the adapter does (as in the suggestion above) is probably fine.

3. That sentence about the import syntax seems a little lacking context, and once again, devs hiding the good stuff in the changeset (which someone might see once) instead of the actual documentation which people will use as a reference!  😄 

	So what about a code example here that optionally shows that this is for sure an SSR'd page (e.g. with the prerender set to false, if you really want to emphasize this) and the example of the proper import syntax as shown in the changeset? (Either a `.astro` page or endpoint would do.)

**Review by sarah11918 — COMMENTED** [review]
Another great contribution @alexanderniebuhr!  🙌  I had some thoughts about the docs, see below and see what you think!  (As always, up to you to capture the exact details/nuance, but this is what I'm envisioning as a nice structure/framing.)

File: packages/integrations/cloudflare/README.md (hunk 3 above)
Comment (alexanderniebuhr) [reply packages/integrations/cloudflare/README.md]: Thank you for your Feedback @sarah11918 🚀 

1. Agree :)
2. It was a question from other reviewers to list it, but you are right the adapter and Cloudflare have parity. I combined both suggestions to hopefully make everyone happy :)
3. That's right, given my experience I usually tend to look at changelogs rather documentations. Good that you point it out! Hope the example makes sense now :)

File: packages/integrations/cloudflare/README.md [hunk 4]
```diff
@@ -172,7 +172,7 @@ This optional flag enables the Astro dev server to populate environment variable
 
 ```js
 // astro.config.mjs
-import { defineConfig } from 'astro/config';
+import { defineConfig } from 'astro/config;'
```
florian-lefebvre, line 175:
```suggestion
import { defineConfig } from 'astro/config';
```

florian-lefebvre (commented):
There's a small typo in the import

alexanderniebuhr:
close in favor of https://github.com/withastro/astro/pull/8595 to unblock smoke tests

PR #8571, opened 2023-09-15
API for clientside router

martrapp:
## Changes

This is the first version of an API for the existing client side router of the view transition feature.
It provides a method to remote control the page navigation.
```
import { navigate } from 'astro:transitions/client';
...
   navigate('/somewhere');
or
  navigate('/somewhere', options);
...
````
It does not support any options yet.
Outlook: First supported option will be: `{history: 'push' | 'replace' | 'auto'}` (see also https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate)

For further discussion see https://github.com/withastro/roadmap/discussions/653

## Testing

added e2e test

## Docs

Needs an additional section in the documents. Would like to enrich the feature a bit more before announcing it.
I'm happy to provide documentation if I can get some guidance.

### CODE REVIEW

**Comment by matthewp** [comment]
Thanks! Will start reviewing next week.

File: packages/astro/src/transitions/router.ts [hunk 1]
```diff
@@ -0,0 +1,394 @@
...
+			}
+			scrollTo({ left: 0, top: 0, behavior: 'instant' });
+		}
+		return;
+	}
+	// these are the cases we will handle: same origin, different page
+	transition('forward', link);
+}
+
+addEventListener('popstate', (ev) => {
```
matthewp, line 343: Curious why this listener and everything below is in this file but the click listener is not. Should it all be in here? I could see leaving prefetch out, as that's not a router behavior.

matthewp:
Looks great! I'd be happy to take on the docs part of this. Do you think it makes sense to add the `history` option now? If it wouldn't be too difficult to support.

packages/astro/src/transitions/router.ts
martrapp: My intention was to move everything into router.ts that is necessary to use the API. For example, the popstate handler is essential because it consumes the history state set by `navigate()`. The click handler is not essential for the router. Rather the other way around. However, this scheme breaks with the meta elements. In the end, `router.ts` also depends on ViewTransition.astro, as it needs to know which pages will/could support view transitions.

matthewp:
I added docs here: https://github.com/withastro/docs/pull/4756

matthewp:
I'll also update the changeset here. We try to have longer changesets for new features.

.changeset/fresh-pots-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
matthewp, line 2: Can you make this be `minor`?

packages/astro/src/transitions/vite-plugin-transitions.ts
```diff
@@ -19,6 +24,11 @@ export default function astroTransitions(): vite.Plugin {
 				export { default as ViewTransitions } from "astro/components/ViewTransitions.astro";
 			`;
 			}
+			if (id === resolvedVirtualClientModuleId) {
+				return `
+				export * from "astro/transitions/router";
```
ematipico, line 29: Shouldn't this be `astro/transitions/router/client`? If not, why?

ematipico (commented):
I left some comments. Many of them are nitpicks, which I prefixed them with `nit:`, feel free to ignore them.

.changeset/fresh-pots-draw.md
martrapp: changed to minor and copied your very good example from the docs. Feel free to improve the text as I am not a native speaker (well at least not of English)

packages/astro/src/transitions/vite-plugin-transitions.ts
martrapp: Rational: in the future, there might be other transformation client stuff  besides the router

packages/astro/src/transitions/router.ts
```diff
@@ -0,0 +1,394 @@
...
+	const { html, ok, mediaType, redirected } = await getHTML(href);
+	// if there was a redirection, show the final URL in the browser's address bar
+	redirected && (loc = new URL(redirected));
+	// If there is a problem fetching the new page, just do an MPA navigation to it.
+	if (!ok || !(mediaType === 'text/html' || mediaType === 'application/xhtml+xml')) {
+		location.href = href;
+		return;
+	}
+
+	const doc = parser.parseFromString(html, mediaType);
```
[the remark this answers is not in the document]
martrapp: no, the function is auto-repairing whatever you throw in as long as the second parameter is a valid DOMParserSupportedType.

packages/astro/src/transitions/router.ts
```diff
@@ -0,0 +1,394 @@
...
+		swap();
+	}
+}
+
+async function transition(dir: Direction, loc: URL, state?: State) {
+	let finished: Promise<void>;
+	const href = loc.href;
+	const { html, ok, mediaType, redirected } = await getHTML(href);
+	// if there was a redirection, show the final URL in the browser's address bar
+	redirected && (loc = new URL(redirected));
```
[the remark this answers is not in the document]
martrapp: Absolutely! We had a discussion about the file size of the script, but minification will turn both the readable and cryptic versions into exactly the same unreadable stuff

packages/astro/src/transitions/router.ts
```diff
@@ -0,0 +1,394 @@
...
+if (import.meta.env.DEV) {
+	noopEl = document.createElement('div');
+}
+
+async function updateDOM(doc: Document, loc: URL, state?: State, fallback?: Fallback) {
+	// Check for a head element that should persist, either because it has the data
+	// attribute or is a link el.
+	const persistedHeadElement = (el: HTMLElement): Element | null => {
+		const id = el.getAttribute(PERSIST_ATTR);
+		const newEl = id && doc.head.querySelector(`[${PERSIST_ATTR}="${id}"]`);
```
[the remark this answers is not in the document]
matthewp: We can't, we only return early if there is an id, which is what happens immediately below this.

packages/astro/src/transitions/router.ts
```diff
@@ -0,0 +1,394 @@
...
+			if (newEl) {
+				// The element exists in the new page, replace it with the element
+				// from the old page so that state is preserved.
+				newEl.replaceWith(el);
+			}
+		}
+
+		// Simulate scroll behavior of Safari and
+		// Chromium based browsers (Chrome, Edge, Opera, ...)
+		scrollTo({ left: 0, top: 0, behavior: 'instant' });
```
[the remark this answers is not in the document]
martrapp: Good point! I've learned a lot about https://html.spec.whatwg.org these days. Actually Firefox is no different here, but what we're doing isn't quite compliant yet either. I'm going to fix that.

packages/astro/src/transitions/router.ts
```diff
@@ -0,0 +1,394 @@
...
+			// from the new document and leave the current node alone
+			if (newEl) {
+				newEl.remove();
+			} else {
+				// Otherwise remove the element in the head. It doesn't exist in the new page.
+				el.remove();
+			}
+		}
+		// Everything left in the new head is new, append it all.
+		document.head.append(...doc.head.children);
```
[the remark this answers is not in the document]
martrapp: fixed by Matthew

martrapp:
Will address the review comments on the main branch in another PR, keep this PR as draft and redo the refactoring for the API just in time for 3.2

matthewp (approved):

matthewp:
"Check mergeability" is ok to fail here, it's the thing that blocks PRs with minor changesets. I think there might be a problem with that script.

zanhk:
Thanks @martrapp

Issue #8321, opened 2023-08-31
Vercel serverless adapter logs too much


lilnasy:
### Astro info

```block
Astro version            v3.0.3
Package manager          pnpm
Platform                 linux
Architecture             x64
Adapter                  @astrojs/vercel/serverless
Integrations             None or couldn't determine.
```

### What browser are you using?

N/A

### Describe the Bug

For the `functionPerRoute` setting, message about node.js compatibility is logged once per route.

![image](https://github.com/withastro/astro/assets/69170106/b53e8b95-d48d-4ee8-929a-8192c7eae1a5)

### What's the expected result?

The compatibility message should run once per build.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-gmnscw?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Notes if you want to take this up:
- [`getRuntime()`](https://github.com/withastro/astro/blob/main/packages/integrations/vercel/src/serverless/adapter.ts#L229) is the function doing the logging
- It is called by [`createFunctionFolder()`](https://github.com/withastro/astro/blob/main/packages/integrations/vercel/src/serverless/adapter.ts#L101) which may run several times.
- Since 3.0, every integration hook has access to a `logger`, replacing console with it will make inserting `PACKAGE_NAME` in every message unnecessary.

rafael-almeida:
Hey @lilnasy,
 
I would like to give this a shot as a first contribution to the project.

lilnasy:
Hi, @rafael-almeida, thanks for the help!

I'll note that the vercel adapter was updated today, but I haven't checked if the issue was fixed. I think the compatibility data may still be outdated.

Please confirm and if the issue is still there, I encourage you to open a PR that mentions this issue. 🙏

rafael-almeida:
Hi @lilnasy,

Thanks for the update. 

I am having trouble reproducing the issue using the stackblitz link you provided. Could you provide more detailed steps or context to help replicate it?

As this is my first open source contribution, I might need some extra guidance. I hope that is not too much trouble 🙏

<details>
  <summary>Here are the logs I got</summary>

```
~/projects/github-gmnscw
❯ pnpm install && pnpm run dev
Lockfile is up to date, resolution step is skipped
Packages: +444
[...]

PR #8581, opened 2023-09-16
fix: Vercel serverless adapter logs too much

rishi-raj-jain:
fix: #8321

### DIFF
--- a/packages/integrations/vercel/src/serverless/adapter.ts
+++ b/packages/integrations/vercel/src/serverless/adapter.ts
@@ -245,6 +245,8 @@ You can set functionPerRoute: false to prevent surpassing the limit.`
 				const filesToInclude = includeFiles?.map((file) => new URL(file, _config.root)) || [];
 				filesToInclude.push(...extraFilesToInclude);
 
+				getRuntime(true);
+
 				// Multiple entrypoint support
 				if (_entryPoints.size) {
 					const getRouteFuncName = (route: RouteData) => route.component.replace('src/pages/', '');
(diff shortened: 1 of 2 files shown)


packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -312,29 +314,31 @@ You can set functionPerRoute: false to prevent surpassing the limit.`
 	};
 }
 
-function getRuntime() {
+function getRuntime(verbose?: boolean) {
```
bluwy, line 317: nit: It would be nice to extract the verbose part as a new function called `validateRuntime` instead.

ematipico (approved):

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -245,6 +245,8 @@ You can set functionPerRoute: false to prevent surpassing the limit.`
 				const filesToInclude = includeFiles?.map((file) => new URL(file, _config.root)) || [];
 				filesToInclude.push(...extraFilesToInclude);
 
+				getRuntime(true);
```
bluwy, line 248: I mean we should have this part as `validateRuntime()` instead, and `getRuntime()` doesn't have any parameters. The `version` and `major` variables would be duplicated, but I think that's fine.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -322,7 +324,6 @@ function getRuntime() {
 		);
 		console.warn(`[${PACKAGE_NAME}] Your project will use Node.js 18 as the runtime instead.`);
 		console.warn(`[${PACKAGE_NAME}] Consider switching your local version to 18.`);
-		return 'nodejs18.x';
```
bluwy, line 325: I think we need to keep an empty `return` here to pass the test

rishi-raj-jain: But I passed the same in the getRuntime function, aight though

bluwy: I mean we need to return an empty `return;`, otherwise it hits the next if condition below. Another option is to change the if condition below as an else-if

rishi-raj-jain: done!

bluwy (approved):
Thanks!

PR #8583, opened 2023-09-16
Tailwind configuration file changed from `cjs` to `ts`

mcarn:
## Changes

- What does this change?
Renamed a 404 page (file ending in cjs) to a valid one.
   * rename https://github.com/withastro/astro.build/blob/main/tailwind.config.cjs ⚠️ 
   * to https://github.com/withastro/astro.build/blob/main/tailwind.config.ts 🆗 


bluwy (approved):
We don't need a changeset as it's only updating the docs. Thanks for the fix!

PR #8588, opened 2023-09-18
docs(@astro/partytown): add docs for `resolveUrl` and Google Tag Manager

ethan-ou:
## Changes

There was some confusion on BuilderIO/partytown#72 about how to use Partytown with Google Tag Manager and Astro.

I've added some extra documentation on how to set up the `resolveURL` property and added a Google Tag Manager example for future users.

## Testing

N/A

## Docs

This is a docs change!

### DIFF
--- a/packages/integrations/partytown/README.md
+++ b/packages/integrations/partytown/README.md
@@ -68,6 +68,8 @@ Partytown should be ready to go with zero config. If you have an existing 3rd pa
 
 If you open the "Network" tab from [your browser's dev tools](https://developer.chrome.com/docs/devtools/open/), you should see the `partytown` proxy intercepting this request.
 
+To run analytics or ads, you may need extra configuration to allow network requests inside of the Partytown worker thread. See the [Configuration](#configuration) and [Examples](#examples) sections.
+
 ## Configuration
 
 To configure this integration, pass a 'config' object to the `partytown()` function call in `astro.config.mjs`.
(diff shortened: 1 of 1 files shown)


bluwy (commented):
My first hunch is that this new docs should be in the main partytown docs instead, since it's not exactly related to Astro. But it also looks like we have existing docs like `config.forward` that is not related to Astro either.

It seems like the docs for `config.resolveUrl` is already documented at https://partytown.builder.io/proxying-requests. And docs for Google Analytics is at https://partytown.builder.io/google-tag-manager albeit incomplete for proxying.

So I'll leave to the docs team to decide if we should have this here instead. But I think it's better in the long run if it's all documented in the core partytown docs.

sarah11918:
Hi @ethan-ou ! As bluwy mentioned, we do try to keep our documentation contained to what is truly needed to configure and integrate with Astro, or any Astro-specific caveats and not to duplicate usage documentation that is managed on your own site, which would be always up to date, not needing our maintenance etc. Also, since documentation here shows up in our site search results, having things like Google Tag Manager on this page might not be desirable. Maybe this is better as a recipe?

So, my question would be whether this is uniquely Astro instructions, something that would be set up differently in Astro etc. or is this some documentation that exists elsewhere that this might be duplicating. We can of course link to your documentation, and seeing that the Astro config option mirrors the Partytown config makes me think that we maybe don't need this documentation here, but can refer? 

Let me know, because of course we do want to have helpful information here, but we don't typically document more than configuration of our own integration and basic usage for Astro specifically, and not, for example, usage of the external service itself. Just trying to figure out where this lies to know how to proceed!

ethan-ou:
Thanks @sarah11918 I agree that the best approach is to put more of the setup guides on Partytown's end and just leave the plugin docs focused on how to match Partytown's config with Astro's config.

Maybe I'll submit a pull request in Partytown and see their appetite for this sort of guide. If that goes through, I'm thinking the Astro Partytown plugin should just have a note saying that analytics in general may require additional setup, and to look at the Partytown docs. What do you think?

sarah11918:
@ethan-ou Sounds good to me! We'd rather not document what probably makes sense to have in PartyTown's own docs, but again, we're happy to be helpful to Astro users here and we would happily link there. So if there are still any changes you want to make to our documentation, please notify me again to take a look. Otherwise, I'll let you decide whether closing this and working on their end makes more sense immediately? Just let me know!

Issue #7508, opened 2023-06-28
>=2.3.0 something changed in scss processing


marceloverdijk:
### What version of `astro` are you using?

`>=2.3.0`

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I'm using official getbootstrap.com themes in a Astro side by using the scss provided by the theme.

That always worked except when upgrading from `2.2.3` to `2.3.0` after it was released.
II notices some strange css behaviour; back then in the nouislider component.

Anyway I thought it was something minor and decided to wait for an upgrade.

I now switched to another getbootstrap.com theme (using `2.2.3) and the theme works as expected.

Decided to update Astro to latest `2.7.1` and I see some unexpected styling issues still unfortunately.

E.g. this how it's rendered with `2.7.1`

<img width="1004" alt="image" src="https://github.com/withastro/astro/assets/197004/963e406e-f93e-413b-9988-a39d50f6b29a">

Note the position of the 2nd dropdown, and also the height is different.

### What's the expected result?

But with `2.2.3` it is rendered as expected:

<img width="980" alt="image" src="https://github.com/withastro/astro/assets/197004/713aadf8-8894-4d67-a0f5-9d5edffd4909">

At the moment I'm unable to provide a reproducible example but if needed I will try to provide something.

I'm wondering if there changed something from `2.3.0` in terms of css/scss processing

My `astro.config.mjs` looks like:

```
import { defineConfig } from 'astro/config'

// https://astro.build/config
export default defineConfig({
  site: process.env.NODE_ENV === 'production' ? 'https://www.example.com/' : 'http://localhost:3000/',
  vite: {
    css: {
      preprocessorOptions: {
        scss: {
          quietDeps: true,
        },
      },
    },
  },
})
```

and `postcss.config.js`:

```
module.exports = {
  plugins: [require('autoprefixer')],
}
```
```
```

### Link to Minimal Reproducible Example

...

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
The biggest styling changes in 2.3.0 is https://github.com/withastro/astro/pull/6816, but I'm not sure how that break things. Maybe a path is incorrectly remapped now? It would be hard for us to debug without a repro.

marceloverdijk:
Thx @bluwy , I will try to create an isolated example repo over the weekend.

[...]

PR #8592, opened 2023-09-18
Fix alias plugin causing CSS ordering issue

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7508

I don't really know what's happening, for some reason if we eagerly resolve ids with aliases, it affects the CSS ordering of some CSS styles injected to the HTML. Doing it later (giving other plugins first) seems to fix it and is also what the plugin used to do before I changed it. (I didn't think they were issues with it before)

## Testing

Tested with the issue's repro only. I can't find a pattern that's causing the issue.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -111,8 +111,6 @@ export async function createVite(
 			exclude: ['astro', 'node-fetch'],
 		},
 		plugins: [
-			configAliasVitePlugin({ settings }),
-			astroLoadFallbackPlugin({ fs, root: settings.config.root }),
 			astroVitePlugin({ settings, logger }),
 			astroScriptsPlugin({ settings }),
 			// The server plugin is for dev only and having it run during the build causes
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/create-vite.ts
```diff
@@ -134,6 +132,8 @@ export async function createVite(
 			vitePluginSSRManifest(),
 			astroAssetsPlugin({ settings, logger, mode }),
 			astroTransitions(),
+			configAliasVitePlugin({ settings }),
+			astroLoadFallbackPlugin({ fs, root: settings.config.root }),
```
bluwy, line 136: Move these two to last as `configAliasVitePlugin` should give way for others to resolve first. And `astroLoadFallbackPlugin` is a fallback, it should always be last. The ordering only affects among other `enforce: 'post'` plugins.

matthewp (approved):

ematipico (approved):
Let's ship it 🚢

PR #8593, opened 2023-09-18
fix: Bare paths redirect to paths with a trailing slash in production

### LINKED ISSUE #7808: Bare paths redirect to paths with a trailing slash in production

9.8.0

Yes, Node.js

npm

Windows

Chrome

### Steps to Reproduce:
1. Create astro project
2. Run `npm astro add node`
3. Set `output: 'hybrid'` in astro config
4. Add `about.astro` to `pages` dir
5. Run `npm run build`
6. Run `node dist/server/entry.mjs`
7. In browser, navigate to `http://localhost:3000/about`
8. Notice redirect to `http://localhost:3000/about/`

### Notes:
- This seems to be at least related to https://github.com/pillarjs/send treating `/about/` as a file and `/about` as a directory
- Does not seem to be reproducible in stackblitz. Stackblitz's request handling may be masking the issue.

We should be able to decide (maybe with some config) to allow both URLs with and without trailing slashes, or to redirect one to the other (3 options -- both, trailing redirects to bare, bare redirects to trailing).

https://stackblitz.com/edit/github-xzttgu?file=sandbox.config.json

### Participation

rishi-raj-jain:
fix #7808

## Changes

Fix bare paths redirect to paths with a trailing slash in production


rishi-raj-jain:
I don't think the files I have edited have anything to do with the macOS test failing.

rishi-raj-jain:
@ematipico Can you guide on writing tests?

packages/integrations/node/src/http-server.ts
```diff
@@ -23,21 +23,36 @@ function parsePathname(pathname: string, host: string | undefined, port: number)
 
 export function createServer(
 	{ client, port, host, removeBase }: CreateServerOptions,
-	handler: http.RequestListener
+	handler: http.RequestListener,
+	trailingSlash: string
```
ematipico, line 27: We lose type information here. We should use `AstroUserConfig['trailingSlash']` here, so we use the same type of the configuration

packages/integrations/node/src/http-server.ts
```diff
@@ -23,21 +23,36 @@ function parsePathname(pathname: string, host: string | undefined, port: number)
...
+			let encodedURI = parsePathname(pathname, host, port);
 
 			if (!encodedURI) {
 				res.writeHead(400);
 				res.end('Bad request.');
 				return res;
 			}
 
-			const stream = send(req, encodedURI, {
+			let pathForSend = encodedURI;
```
ematipico, line 41: `pathForSend` => `pathToSend`

packages/integrations/node/src/http-server.ts
```diff
@@ -23,21 +23,36 @@ function parsePathname(pathname: string, host: string | undefined, port: number)
...
+			if (trailingSlash === 'never') {
+				if (pathname.endsWith('/')) {
+					encodedURI = parsePathname(pathname.substring(0, pathname.length - 1), host, port);
+					if (!encodedURI) {
+						res.writeHead(400);
+						res.end('Bad request.');
+						return res;
+					}
+				}
+				pathForSend = encodedURI + '/';
```
ematipico, line 52: Why are we adding a trailing slash if `trailingSlash` is `"never"`

ematipico (changes requested):
I think we have to add a couple of tests. For example, you added a `400` code and a `301` code, and we should create two test cases that hit those cases.

packages/integrations/node/src/http-server.ts
rishi-raj-jain: Hey @ematipico, we're adding a trailingSlash to pathForSend in case of trailingSlash is never so that the directory like behavior is maintained but URL redirection doesn't take place.

rishi-raj-jain:
Aiming to get this done by tomorrow.

florian-lefebvre:
#9080 was aiming to fix the same bug but since it was made later than this PR, we would have taken yours to fix the bug. However, it turns out your fix won't actually work so we asked PR 9080 author to make changes in core instead.

Issue #8547, opened 2023-09-13
Static builds stops at `generating static routes` without error when using optimized images


BryceRussell:
### Astro Info

```block
(stackblitz minimal reproduction)
Astro                    v3.0.8
Node                     v16.20.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

In some cases, when using optimized images, `astro build` will stop/fail at the `generating static route` stage and not throw any errors making production builds appear as if they were successful

**Cases**:
1) [Support Thread](https://discord.com/channels/830184174198718474/1148654506528231434/1148654506528231434) | [Example Repo](https://github.com/scottaw66/sierraalfawhiskey/blob/brokentest/src/content/posts/1.md?plain=1#L11)
    - Breaks when using an inline relative image path in markdown
1) [Support Thread](https://discord.com/channels/830184174198718474/1146556928856707133/1146556928856707133) | [Example](https://discord.com/channels/830184174198718474/1146556928856707133/1146781753516765307) | [Minimal Reproduction](https://github.com/BryceRussell/minimal-repo-md-imgs-in-public)
    - Breaks when using an inline relative image path in markdown to asset in `/public`
1) [Support Thread](https://discord.com/channels/830184174198718474/1137374189280571562/1151553338291269643) | [Github Pages Example](https://github.com/Pocee/PoceBlog/actions/runs/6175451494/job/16762328084#step:3:113) | [Minimal Reproduction (Stackblitz)](https://stackblitz.com/edit/github-threhe?file=src%2Fpages%2Findex.md) | [Minimal Reproduction v2](https://github.com/BryceRussell/minimal-repo-md-imgs-in-pages)
    - Breaks when using an inline relative image path in markdown located in `/src/pages`
1)  [Support Thread](https://discord.com/channels/830184174198718474/1151955934894096455/1151993152920625298) | [Example Repo](https://github.com/webpro/webpro.nl/blob/astro-issue/src/pages/og/%5B...route%5D.ts#L1)
    - Breaks when using `import.meta.glob()` to glob markdown containing inline relative image paths
1) [Support Thread](https://discord.com/channels/830184174198718474/1148668689911119922/1148668689911119922) | [Example](https://discord.com/channels/830184174198718474/1148668689911119922/1148671563286188083)
    - Breaks when importing a `.md` file with inline relative image paths
[...]

Issue #8454, opened 2023-09-07
Octal literal in strict mode with markdown images


jsnape:
### Astro Info

```block
Astro                    v3.0.10
Node                     v18.17.1
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
                         @astrojs/mdx
```

### Describe the Bug

From support discussion: https://discord.com/channels/830184174198718474/1149037613961977906

When a markdown document references an image that it cannot find but the past to that image has numbers in it then instead or saying the file can't be found the error returned is similar to below:

_Parse failure: Octal literal in strict mode (9:130)
At file: E:/Dev/snape.me/src/content/posts/2009/04/using-integration-services-to-populate-a-date-dimension.md
Contents of line 9:                     'image_3.png': await getImageSafely((await import("image_3.png")).default, "image_3.png", "/E:\Dev\snape.me\src\content\posts\2009\04\image_3.png")_

Is cause by the following in the .md file:

```Markdown
![Script Component Type](image_3.png)
```

But if I change it to:

```Markdown
![Script Component Type](./image_3.png)
```

Then the page loads correctly.

This only happens on Windows.

When I run it in https://codesandbox.io/ then the error message is (far easier to understand):

_Cannot find module 'image_3.png' imported from '/workspace/src/content/posts/2009/04/using-integration-services-to-populate-a-date-dimension.md'_

### What's the expected result?

Expected result is that the error message is easy to debug like the Linux version above.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-i55tcq?file=src%2Fcontent%2Fblog%2F2023%2F09%2Fbroken-image.md

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #8598, opened 2023-09-19
fix(markdown): Fix Markdown images breaking the build

Princesseuh:
## Changes

This PR reworks how we handle images in Markdown to workaround [a Rollup bug with top level await](https://github.com/rollup/rollup/issues/4708). This has the side effect of needing us to rework how error handling is done around these parts, which honestly is great because it had some Markdown specific parts that I was able to make more general, improving the overall experience.

Fix https://github.com/withastro/astro/issues/8547
Also fix https://github.com/withastro/astro/issues/8454, by accident

## Testing

Tests should all pass! The specific problem here is kinda hard to test because it depends specifically on how the website gets bundled, confusing

## Docs

Added two errors and deprecated an old one. One of added one is mostly a copy paste of the deprecated one.


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -620,8 +620,42 @@ export const ExpectedImageOptions = {
 	message: (options: string) =>
 		`Expected getImage() parameter to be an object. Received \`${options}\`.`,
 } satisfies ErrorData;
+
+/**
+ * @docs
+ * @see
+ * - [Images](https://docs.astro.build/en/guides/images/)
+ * @description
+ * Astro could not find an image you imported. Usually, this is simply caused by a typo in the path.
```
sarah11918, line 629:
```suggestion
 * Astro could not find an image you imported. Often, this is simply caused by a typo in the path.
```
I do like hedging our bets a little more with "often" here!

sarah11918 (approved):
Only one small nit I'd consider, otherwise, LGTM!

packages/astro/src/vite-plugin-markdown/index.ts
```diff
@@ -118,36 +120,25 @@ export default function markdown({ settings, logger }: AstroPluginOptions): Plug
...
+						return html.replaceAll(/__ASTRO_IMAGE_="([^"]+)"/gm, (full, imagePath) =>
+							spreadAttributes({
+								src: images[imagePath].src,
+								...images[imagePath].attributes,
+							})
+						);
+					});
 				}
 
 				const html = updateImageReferences(${JSON.stringify(html)});
```
bluwy, line 144: Since `updateImageReferences` now returns a promise, wouldn't `html` become a promise and affect `compiledContent` and the `Content` render code?

Seems like we need another TLA here but maybe that is able to side-step the Rollup bug.

Princesseuh: Ah yes, I didn't notice that it made `html` a promise since it worked in rendering... Unfortunately, putting a TLA anywhere in this file causes the issue.

bluwy: I think the Rollup bug is only caused when dynamically importing a module that causes an import loop, but this PR removed the dynamic import as a top-level import, so theoretically, I think we can keep using TLA?

Unless if you're already tested it, then pardon me 😅 We might have to refactor some modules to break the loop otherwise. Or fix in Rollup directly.

bluwy: By TLA, I mean directly on this line, like `const html = await ...`

Princesseuh: Doing `const html = await` causes the issue, unfortunately

bluwy: I guess I could try fixing this in Rollup directly

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -29,6 +30,17 @@ export default function assets({
 		// Expose the components and different utilities from `astro:assets` and handle serving images from `/_image` in dev
 		{
 			name: 'astro:assets',
+			outputOptions(outputOptions) {
+				// Specifically split out chunk for asset services to prevent TLA deadlock
+				// https://github.com/rollup/rollup/issues/4708
+				extendManualChunks(outputOptions, {
+					before(id) {
+						if (id.includes('astro/dist/assets/services')) {
```
Princesseuh, line 38: Will this hardcoding cause an issue for custom services?

bluwy: Would custom services import things like from `astro/assets/utils`? Then maybe we need to go one directory up and make it `astro/dist/assets` instead 🤔

Princesseuh: Hmm, admittedly, it's fairly unlikely

bluwy: I think either ways it should still work fine as we're primarily grouping our Astro code into one chunk. The issue before was that `services/sharp.ts` uses `services/service.ts`. But both were in two different chunks. `services/sharp.ts` was in it's own chunk, but `services/service.ts` got grouped with the markdown page chunk. And so there was an import loop, and the TLA in the markdown page made it worst.

So manually handling where `services/service.ts` should be chunked fixed it here. I don't mind if we want to go back one directory down though.

Princesseuh: I'm honestly not sure of the impact this can have, so I think whichever works just fine for me as long as it works, ha.

If it turns out to be a problem somehow, we can change it back fairly easily it seems like

bluwy: Changed it back to `astro/dist/assets/services/` for now since it feels a liitle safer that way 👍

PR #8600, opened 2023-09-19
telemetry: Improve config info object format

FredKSchott:
## Changes

- Removes outdated/unused/broken data from the `ASTRO_CLI_SESSION_STARTED` telemetry event
- Adds a new, more generic config info base object using "is set" boolean checks
- Adds more protection with `measureX` helper functions that prevent raw user input collective with input checking, type errors, and centralized sanitization logic.

## Testing

- Tests updated.

## Docs

- See https://astro.build/telemetry/ for information about how Astro collects anonymous telemetry and how we use it to improve our roadmap.

### DIFF
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -9,7 +9,12 @@ import ora from 'ora';
 import preferredPM from 'preferred-pm';
 import prompts from 'prompts';
 import type yargs from 'yargs-parser';
-import { loadTSConfig, resolveConfigPath, resolveRoot } from '../../core/config/index.js';
+import {
+	loadTSConfig,
+	resolveConfig,
+	resolveConfigPath,
+	resolveRoot,
+} from '../../core/config/index.js';
 import {
 	defaultTSConfig,
 	presets,
@@ -23,7 +28,7 @@ import { appendForwardSlash } from '../../core/path.js';
 import { apply as applyPolyfill } from '../../core/polyfill.js';
 import { parseNpmName } from '../../core/util.js';
 import { eventCliSession, telemetry } from '../../events/index.js';
-import { createLoggerFromFlags } from '../flags.js';
+import { createLoggerFromFlags, flagsToAstroInlineConfig } from '../flags.js';
 import { generate, parse, t, visit } from './babel.js';
 import { ensureImport } from './imports.js';
 import { wrapDefaultExport } from './wrapper.js';
@@ -87,7 +92,9 @@ async function getRegistry(): Promise<string> {
 }
 
 export async function add(names: string[], { flags }: AddOptions) {
-	telemetry.record(eventCliSession('add'));
+	const inlineConfig = flagsToAstroInlineConfig(flags);
+	const { userConfig } = await resolveConfig(inlineConfig, 'add');
+	telemetry.record(eventCliSession('add', userConfig));
 	applyPolyfill();
 	if (flags.help || names.length === 0) {
 		printHelp({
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/config/schema.ts
```diff
@@ -140,7 +140,6 @@ export const AstroConfigSchema = z.object({
 				.optional()
 				.default(ASTRO_CONFIG_DEFAULTS.build.excludeMiddleware),
 		})
-		.optional()
```
natemoo-re, line 143: Don't we need this?

packages/astro/src/cli/add/index.ts
```diff
@@ -87,7 +92,9 @@ async function getRegistry(): Promise<string> {
 }
 
 export async function add(names: string[], { flags }: AddOptions) {
-	telemetry.record(eventCliSession('add'));
+	const inlineConfig = flagsToAstroInlineConfig(flags);
+	const { userConfig } = await resolveConfig(inlineConfig, 'add');
+	telemetry.record(eventCliSession('add', userConfig));
```
natemoo-re, line 97: Is the current config actually important to record here?

packages/astro/src/core/config/schema.ts
```diff
@@ -158,7 +157,6 @@ export const AstroConfigSchema = z.object({
 				port: z.number().optional().default(ASTRO_CONFIG_DEFAULTS.server.port),
 				headers: z.custom<OutgoingHttpHeaders>().optional(),
 			})
-			.optional()
```
natemoo-re, line 161: Same. Seems necessary, unless I'm missing some context.

packages/astro/src/core/config/schema.ts
```diff
@@ -274,27 +272,11 @@ export const AstroConfigSchema = z.object({
...
-					)}\`. \nMake sure the spelling is correct, and that your Astro version supports this experiment.\nSee https://docs.astro.build/en/reference/configuration-reference/#experimental-flags for more information.`,
-				};
-			}
+		.strict(
+			`Invalid or outdated experimental feature.\nCheck for incorrect spelling or outdated Astro version.\nSee https://docs.astro.build/en/reference/configuration-reference/#experimental-flags for a list of all current experiments.`
 		)
-		.optional()
 		.default({}),
-	legacy: z.object({}).optional().default({}),
+	legacy: z.object({}).default({}),
```
natemoo-re, line 279: Also this one.

natemoo-re (commented):
Changes look good! I appreciate the typesafety updates.

packages/astro/src/core/config/schema.ts
natemoo-re: After doing some research, I guess not! `.optional()` + `.default({})` is redundant. There will always be a value.

packages/astro/src/cli/add/index.ts
natemoo-re: I don't have a problem if this info is statistically useful, but it seems like we'll get final data when users run `dev` or the `build` so would err on the side of not recording this.

natemoo-re (approved):
Questions withdrawn because I know how to Google.

Nothing blocking, the one comment I left is up to your discretion.

packages/astro/src/core/config/schema.ts
FredKSchott: Yup, this was a surprise for me as well! I wouldn't have made a change in this PR except that it actually changes the schema object's API when I wanted to get the shape of the schema back out for validation.

bluwy (approved):

packages/astro/src/cli/add/index.ts
bluwy: Maybe we can pass an empty user config object instead so we don't have to load the entire thing? e.g. `telemetry.record(eventCliSession('add', {}))`

FredKSchott: Is the reason that we don't track it here because it's slow? 

I see approvals so I'll merge but happy to revisit this post-PR if there's still interest in removing!

Issue #8482, opened 2023-09-10
Possible issue with JS and ViewTransition


sparxastronomy:
### Astro Info

```block
Astro                    v3.0.12
Node                     v18.16.0
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/sitemap
                         @astrojs/tailwind
                         @astrojs/mdx
```

### Describe the Bug

I'm trying out the new ViewTransition update. 

I've two basic  JS function,  one to toggle navbar menu and one to toggle themes.

```js
<script is:inline>
	// Definig the function to attach event to the element
	function attachEvent(selector, event, fn) {
		const matches = document.querySelectorAll(selector);
		if (matches && matches.length) {
			matches.forEach((elem) => {
				elem.addEventListener(event, () => fn(elem), false);
			});
		}
	}

	// Defining the function to check if the menu is open
	const menuCheck = () =>{
		const elem = document.querySelector('[data-aw-toggle-menu]');
		if (elem) {
			elem.classList.remove('expanded');
		}
		document.body.classList.remove('overflow-hidden');
		document.getElementById('header')?.classList.remove('h-screen');
		document.querySelector('#header nav')?.classList.add('hidden');
		return;
	}

	// Defining the function to check and set the theme
	const themeCheck = () =>{
		if (
		localStorage.theme === 'dark' ||
		(!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)
		) {
			document.documentElement.classList.add('dark');
			} else {
			document.documentElement.classList.remove('dark');
			}
		return;
	}

	// Check theme and menu on window load
	window.addEventListener("DOMContentLoaded", () => themeCheck());
	window.addEventListener("DOMContentLoaded", () => menuCheck());
	// Check theme and menu on after swap
	document.addEventListener("astro:after-swap", () => themeCheck());
	document.addEventListener("astro:after-swap", () => menuCheck());

	window.onload = function () {
		attachEvent('[data-aw-toggle-menu]', 'click', function (elem) {
			elem.classList.toggle('expanded');
			document.body.classList.toggle('overflow-hidden');
			document.getElementById('header')?.classList.toggle('h-screen');
			document.querySelector('#header nav')?.classList.toggle('hidden');


		});
		attachEvent('[data-aw-toggle-theme]', 'click', function () {
			document.documentElement.classList.toggle('dark');
[...]

Issue #8466, opened 2023-09-08
ViewTransition - Identifier 'theme' has already been declared


visualnaut:
### Astro Info

```block
Astro                    v3.0.7
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/mdx
```

### Describe the Bug

When using ViewTransition on global layout with a astro component that have inline script, it somehow logged a syntax error every time we traverse a page and the inline script did not worked. 

I search in the issues list and I found old similar case but with different root cause https://github.com/withastro/astro/issues/3643

### What's the expected result?

Inline script inside astro component worked and no syntax error being logged.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wqkwwm

### Participation

- [X] I am willing to submit a pull request for this issue.

dawichi:
### Exactly the same error here with ViewTransition

I have a blog with just a homepage `pages/index.astro` and 2 posts (content) visibles under `[...slug].astro`. With tailwind and mdx, nothing fancy, just wanted to play with this viewtransitions thing

After set the `<ViewTransitions/>` in the head, that error appeared each time i change the page.

```
VM2713:1  Uncaught SyntaxError: Identifier 'theme' has already been declared
    at runScripts (ViewTransitions.astro:102:11)
    at navigate (ViewTransitions.astro:276:10)
```

Also, this is my <DarkToggle /> component, nothing fancy neither


```
<button id="dark-toggle" class="bg-gray-300 dark:bg-zinc-700 rounded-full px-2 py-1">
    <i class="bi bi-moon-fill" />
</button>

<script is:inline>
    function toggleDarkMode() {
        const newMode = localStorage.getItem('theme') === 'dark' ? 'light' : 'dark'
        document.body.dataset.theme = newMode
        localStorage.setItem('theme', newMode)
        document.documentElement.classList.toggle('dark')
    }

    // Attach event listener
    document.getElementById('dark-toggle').addEventListener('click', toggleDarkMode)

    // Set initial theme
    const theme = localStorage.getItem('theme')
    if (theme === 'dark') {
        document.body.dataset.theme = theme
        document.documentElement.classList.toggle('dark')
    }
</script>
```

The weird thing is that if I delete the `<DarkToggle />` component, the error dissapears, but for example each time I swap between posts, i get a very similar one! And yes! it has an inline script too. (to generate TOCs for each post, particularly)

```
[...]

PR #8603, opened 2023-09-19
Prevent body scripts from re-executing on navigation

matthewp:
## Changes

- Closes https://github.com/withastro/astro/issues/8466
- Closes https://github.com/withastro/astro/issues/8482

## Testing

- New test case added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -129,18 +129,18 @@ const { fallback = 'animate' } = Astro.props as Props;
 		var noopEl = document.createElement('div');
 	}
 
-	async function updateDOM(doc: Document, loc: URL, state?: State, fallback?: Fallback) {
+	async function updateDOM(newDocument: Document, loc: URL, state?: State, fallback?: Fallback) {
 		// Check for a head element that should persist, either because it has the data
 		// attribute or is a link el.
 		const persistedHeadElement = (el: HTMLElement): Element | null => {
 			const id = el.getAttribute(PERSIST_ATTR);
-			const newEl = id && doc.head.querySelector(`[${PERSIST_ATTR}="${id}"]`);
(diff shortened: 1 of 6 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -202,6 +189,20 @@ const { fallback = 'animate' } = Astro.props as Props;
 			// Everything left in the new head is new, append it all.
 			document.head.append(...doc.head.children);
 
+			// Replace scripts
+			for(const s1 of document.scripts) {
```
matthewp, line 193: This happens here so that we get all scripts, not just the ones in the head.

packages/astro/components/ViewTransitions.astro
```diff
@@ -187,6 +174,22 @@ const { fallback = 'animate' } = Astro.props as Props;
 				html.setAttribute(name, value)
 			);
 
+			// Replace scripts in both the head and body.
+			for(const s1 of document.scripts) {
+				for (const s2 of doc.scripts) {
```
ematipico, line 179: Any chance to have a better name? `document` and `doc` are the same to me 😅

matthewp: Yeah, it's really the new document, i'll rename

matthewp: Well actually, it might be better to do that in https://github.com/withastro/astro/pull/8571 since that's a refactor. cc @martrapp

martrapp: Please feel free to make those chances directly on the current source. I will also do the improvements that ematipico suggested for # 8571 on the current state and update the refactoring just before 3.2

ematipico: Thank you

PR #8607, opened 2023-09-20
feat(i18n): first draft of configuration

ematipico:
## Changes

This is a first draft of the possible configuration for the upcoming i18n routing. Closes PLT-981

> [!NOTE]
> This PR is against the branch `i18n-routing`

### Configuration

```js
export default defineConfig({
	i18n: {
		defaultLocale: 'en' // mandatory. This locale must be in the locales array
		locales: [
			'en',
			'es', 
			'en-US', 
			'it-IT',
			'pt',
			'pt-BR'
		],
		fallback: {
			/**
			* The fallback system. The fallback system handles cases where certain pages 
			* aren't translated yet.
			* Fallback = Redirect
			* In this example, if a page for `pt-BR` isn't found, Astro will:
			* 1. fallback to `pt`
			* 2. fallback to `en-US`
			* 3. fallback to default locale
			*/
			'pt-BR': ['pt', 'en-US'] 
		},
		detectBrowswerLanguage: false
	}
})
```

### Initial runtime

~~Considering the proposal, we need the presence of some runtime in order to provide enough information to the client. 

By default, users will create files inside directories that have the name of the locale, and eventually the links in the website will look like this: `https://example.com/es/guide/getting-started`

Although, if a user needs a custom domain for a certain locale, the links need to change. So if in the example we opt-in the custom domain for `es` locale, the links will need to look like this: `https://es.example.com/guide/getting-started`

The runtime in this PR provides an example of the functions we should provide to the user to achieve that.~~

Domains was left out of scope for the initial, experimental, release. Instead, the browser detection has introduced as an opt-in feature.

> [!WARNING]
> The proposed code is not optimized yet, but the unit tests should provide enough information of the functionality of the runtime

## Testing

Some unit test. Current CI should pass for the new configuration.

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1331,21 +1331,65 @@ export interface AstroUserConfig {
 		 */
 		optimizeHoistedScript?: boolean;
 
+		// TODO review with docs team before merging to `main`
 		/**
 		 * @docs
 		 * @name experimental.i18n
-		 * @type {boolean}
 		 * @version 3.*.*
 		 * @description
-		 * Allows to configure
+		 *
+		 * Allows to configure the beaviour of the i18n routing
 		 */
 		i18n?: {
-			// TODO documentation
+			/**
+			 * @docs
+			 * @name experimental.i18n.defaultLocale
+			 * @type {string}
+			 * @version 3.*.*
+			 * @description
+			 *
+			 * The default locale of your website/application
+			 */
 			defaultLocale: string;
-			// TODO documentation
-			locales?: string[];
-			// TODO documentation
+			/**
+			 * @docs
+			 * @name experimental.i18n.locales
+			 * @type {string[]}
+			 * @version 3.*.*
+			 * @description
+			 *
+			 * A list of locales supported by the website.
+			 */
+			locales: string[];
+
+			/**
+			 * @docs
+			 * @name experimental.i18n.fallback
+			 * @type {Record<string, string[]>}
+			 * @version 3.*.*
+			 * @description
+			 *
+			 * The fallback system of the locales. By default, the fallback system affect the **content only**, and it doesn't
+			 * do any redirects.
+			 *
+			 * This means that when attempting to navigate to a page that hasn't been translated, Astro will pull the content
+			 * from the page of the default locale and render it. No redirects will happen.
+			 */
 			fallback?: Record<string, string[]>;
+
+			/**
+			 * @docs
+			 * @name experimental.i18n.detectBrowserLanguage
+			 * @type {boolean}
+			 * @version 3.*.*
+			 * @description
+			 *
+			 * Whether Astro should detect the language of the browser - usually using the `Accept-Language` header. This is a feature
+			 * that should be supported by the adapter. If detected, the adapter can decide to redirect the user to the localised version of the website.
+			 *
+			 * When set to `true`, you should make sure that the adapter you're using is able to provide this feature to you.
+			 */
+			detectBrowserLanguage: boolean;
 		};
 	};
 }
@@ -1807,6 +1851,11 @@ export type AstroFeatureMap = {
 	 * The adapter can emit static assets
 	 */
 	assets?: AstroAssetsFeature;
+
+	/**
+	 * List of features that orbit around the i18n routing
+	 */
+	i18n?: AstroInternalisationFeature;
 };
 
 export interface AstroAssetsFeature {
@@ -1819,6 +1868,13 @@ export interface AstroAssetsFeature {
 	 * Whether if this adapter deploys files in an environment that is compatible with the library `squoosh`
 	 */
 	isSquooshCompatible?: boolean;
+}
+
+export interface AstroInternalisationFeature {
+	/**
+	 * Wether the adapter is able to detect the language of the browser, usually using the `Accept-Language` header.
+	 */
+	detectBrowserLanguage?: SupportsKind;
 }
 
 export interface AstroAdapter {
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1335,6 +1335,7 @@ export interface AstroUserConfig {
 		/**
 		 * @docs
 		 * @name experimental.i18n
+		 * @type {object}
 		 * @version 3.*.*
 		 * @description
 		 *
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1330,6 +1330,25 @@ export interface AstroUserConfig {
 		 * ```
 		 */
 		optimizeHoistedScript?: boolean;
+
+		/**
+		 * @docs
+		 * @name experimental.i18n
+		 * @type {boolean}
+		 * @version 3.*.*
+		 * @description
+		 * Allows to configure
+		 */
+		i18n?: {
+			// TODO documentation
+			defaultLocale: string;
+			// TODO documentation
+			locales?: string[];
+			// TODO documentation
+			fallback?: Record<string, string[]>;
+			// TODO documentation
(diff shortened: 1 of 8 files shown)


packages/astro/src/i18n/index.ts
```diff
@@ -0,0 +1,20 @@
+import type { AstroConfig } from '../@types/astro.js';
+import type { Logger } from '../core/logger/core.js';
+
+/**
+ * The base URL
+ */
+export function getI18nBaseUrl(locale: string, config: AstroConfig, logger: Logger) {
+	if (!config.experimental.i18n) {
+		logger.debug('i18n', "The project isn't using i18n features, no need to use this function.");
+		return;
```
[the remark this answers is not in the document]
ematipico: Yeah you're right, although returning throwing an error feels too much. What do you think if we **log** and **error** and return your suggested snippet?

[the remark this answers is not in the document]
florian-lefebvre: So you mean this?
```ts
logger.warn(...)
logger.error(...)
return ...
```
What's the point of having both warning and error?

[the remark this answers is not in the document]
ematipico: Sorry I meant _log AN error_ 😆

packages/astro/src/i18n/index.ts
```diff
@@ -0,0 +1,20 @@
...
+	}
+	const base = config.base;
+
+	if (base) {
+		logger.debug('i18n', 'The project has a base directory, using it.');
+		return `${base}/${locale}/`;
+	} else {
+		return `/${locale}/`;
+	}
```
florian-lefebvre, line 19:
```suggestion
	const base = config.base;

	if (!config.experimental.i18n) {
		logger.error('i18n', "The project isn't using i18n features, no need to use this function.");
		return base ? `/${base}/` : '/';
	}

	if (base) {
		logger.debug('i18n', 'The project has a base directory, using it.');
		return `${base}/${locale}/`;
	} else {
		return `/${locale}/`;
	}
```

packages/astro/src/i18n/index.ts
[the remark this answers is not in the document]
florian-lefebvre: You can resolve this conversation this 👍

lilnasy:
There are unrelated commits in the diff. You may want to update the base branch `-routing`, or drop commits from `-routing-configuration`.

ematipico:
Thank you @lilnasy for the call out, I didn't see them

ematipico:
That should be OK now :)

packages/astro/src/@types/astro.ts
```diff
@@ -1330,6 +1330,68 @@ export interface AstroUserConfig {
...
+
+		// TODO review with docs team before merging to `main`
+		/**
+		 * @docs
+		 * @name experimental.i18n
+		 * @type {object}
+		 * @version 3.*.*
+		 * @description
+		 *
+		 * Allows to configure the beaviour of the i18n routing
```
bluwy, line 1342:
```suggestion
		 * Allows to configure the behaviour of the i18n routing
```

packages/astro/src/@types/astro.ts
```diff
@@ -1804,6 +1871,13 @@ export interface AstroAssetsFeature {
 	isSquooshCompatible?: boolean;
 }
 
+export interface AstroInternalisationFeature {
+	/**
+	 * Wether the adapter is able to detect the language of the browser, usually using the `Accept-Language` header.
```
bluwy, line 1876:
```suggestion
	 * Whether the adapter is able to detect the language of the browser, usually using the `Accept-Language` header.
```

packages/astro/src/core/config/schema.ts
```diff
@@ -271,6 +271,43 @@ export const AstroConfigSchema = z.object({
...
+						defaultLocale: z.string(),
+						locales: z.string().array(),
+						fallback: z.record(z.string(), z.string().array()).optional(),
+						detectBrowserLanguage: z.boolean().optional().default(false),
+					})
+					.optional()
+					.refine((i18n) => {
+						if (i18n) {
+							const { defaultLocale, locales, fallback } = i18n;
+							if (locales.includes(defaultLocale)) {
```
bluwy, line 286:
```suggestion
							if (!locales.includes(defaultLocale)) {
```

Looks like an incorrect check

packages/astro/src/i18n/vite-plugin-i18n.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const virtualModuleId = 'astro:i18n';
+const resolvedVirtualModuleId = '\0' + virtualModuleId;
+
+type AstroInternalization = {
+	settings: AstroSettings;
+	logger: Logger;
+};
+
+export default function astroInternalization({ settings }: AstroInternalization): vite.Plugin {
```
bluwy, line 13:
```suggestion
type AstroInternationalization = {
	settings: AstroSettings;
	logger: Logger;
};

export default function astroInternationalization({ settings }: AstroInternationalization): vite.Plugin {
```

I wouldn't mind if it's shorten as i18n here too 🙈

packages/astro/src/i18n/index.ts
```diff
@@ -0,0 +1,21 @@
+import type { AstroConfig } from '../@types/astro.js';
+import type { Logger } from '../core/logger/core.js';
+
+/**
+ * The base URL
+ */
+export function getI18nBaseUrl(locale: string, config: AstroConfig, logger: Logger) {
```
bluwy, line 7: Is `getI18nBaseUrl` user facing? It would be hard for them to obtain a value for `config` and `logger`. If it's only for integrations, then maybe we should leave it as `astro/i18n` like `astro/errors`.

packages/astro/src/i18n/vite-plugin-i18n.ts
```diff
@@ -0,0 +1,26 @@
...
+	return {
+		name: 'astro:i18n',
+		load(id) {
+			if (id === resolvedVirtualModuleId) {
+				return `
+					import { getI18nBaseUrl as getI18nBaseUrlInternal } from "astro/i18n";
+								
+					export getI18nUrl = (locale) => getI18nBaseUrlInternal(locale, ${settings.config});
```
bluwy, line 21: I'm not sure if this will work as `settings.config` isn't necessarily completely serializable as a string, and the logger isn't passed in too 👀

ematipico: > I'm not sure if this will work as settings.config isn't necessarily completely serializable as a string

Where am I expecting to retrieve this value then? 

> and the logger isn't passed in too 👀

Cheers, I'll fix it when I will tackle this part

packages/astro/src/i18n/vite-plugin-i18n.ts
ematipico: Will change that, cheers

packages/astro/src/i18n/vite-plugin-i18n.ts
bluwy: Since you're only reading the `base` and the truthy value of `config.experimental.i18n`, you can read both of them outside of this string, and then inject them individually into this string, since both are serializable (with `JSON.stringify(someValue)`).

But I'm not sure a logger is serializable though. You might need to resort to `console.warn` instead.

ematipico: Ah ok, I thought you were talking about something else. Yes that will be done when I tackle the virtual module

packages/astro/src/i18n/index.ts
ematipico: Yes, that would the idea yes

ematipico:
I am going to merge this one because **I need it**. Leave your feedback here and I will address it later

PR #8612, opened 2023-09-20
Ensure cookies are attached when middleware changes the Response

matthewp:
## Changes

- When middleware return `new Response` we need to attach cookies to it.
- This aligns with how components work as well, if you return a new Response from a page component the cookies that have been set via `Astro.cookies.set()` go along for the ride.

## Testing

New test case added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/curvy-owls-rest.md
+++ b/.changeset/curvy-owls-rest.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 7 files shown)


ematipico (approved):
Now that I look at the fix, it seemed so simple 😢 

Thank you for looking into it Matthew!

bluwy (approved):

PR #8615, opened 2023-09-21
chore(core): improve the logging of assets feature

alexanderniebuhr:
Seeing a lot of confusion about the `noop` logging. This issue has come up multiple times on Discord already (e.g., [Discord link](https://discord.com/channels/830184174198718474/1154082475547574385)). I believe we should adjust the logging to avoid confusing users.

## Changes

- **Changed** the logging approach in `validateAssetsFeature`
- **Removed** the error log stating that the `assets` feature is unsupported
- **Lowered** the severity of the log from error to warning for `sharp` or `squoosh` incompatibility. Since we automatically replace these with `noop`, the user shouldn't need to make any manual changes; therefore, an error is too severe.
- **Reworded** the warning log about automatically switching to `noop`

**For more details, see self-review comments**

## Docs


packages/astro/src/integrations/astroFeaturesValidation.ts
```diff
@@ -131,15 +131,15 @@ function validateAssetsFeature(
 		isSquooshCompatible = false,
 	} = assets;
 	if (config?.image?.service?.entrypoint === SHARP_SERVICE && !isSharpCompatible) {
-		logger.error(
+		logger.warn(
```
alexanderniebuhr, line 134: the user shouldn't need to make any manual changes; therefore, an error is too severe

packages/astro/src/integrations/astroFeaturesValidation.ts
```diff
@@ -131,15 +131,15 @@ function validateAssetsFeature(
...
 			'astro',
 			`The currently selected adapter \`${adapterName}\` is not compatible with the image service "Sharp".`
 		);
 		return false;
 	}
 
 	if (config?.image?.service?.entrypoint === SQUOOSH_SERVICE && !isSquooshCompatible) {
-		logger.error(
+		logger.warn(
```
alexanderniebuhr, line 142: the user shouldn't need to make any manual changes; therefore, an error is too severe

packages/astro/src/integrations/index.ts
```diff
@@ -221,17 +221,21 @@ export async function runHookConfigDone({
 								logger
 							);
 							for (const [featureName, supported] of Object.entries(validationResult)) {
-								if (!supported) {
+								// If `supported` / `validationResult[featureName]` only allows boolean,
+								// in theory 'asstes' false, doesn't mean that the feature is not supported, but rather that the chosen image service is unsupported
+								// in this case we should not show an error, that the featrue is not supported
+								// if we would refactor the validation to support more than boolean, we could still be able to differentiate between the two cases
+								if (!supported && featureName !== 'assets') {
```
alexanderniebuhr, line 228: As already laid out in the comment, I believe that in the long term, this message should not be based on a boolean. Instead, we should be able to filter whether a feature is unsupported or partially supported.

packages/astro/src/integrations/index.ts
```diff
@@ -221,17 +221,21 @@ export async function runHookConfigDone({
...
 									logger.error(
 										'astro',
 										`The adapter ${adapter.name} doesn't support the feature ${featureName}. Your project won't be built. You should not use it.`
 									);
 								}
 							}
 							if (!validationResult.assets) {
-								logger.info(
+								logger.warn(
```
alexanderniebuhr, line 236: the user shouldn't need to make any manual changes; therefore, an error is too severe

packages/astro/src/integrations/index.ts
```diff
@@ -221,17 +221,21 @@ export async function runHookConfigDone({
...
 									'astro',
-									`The selected adapter ${adapter.name} does not support Sharp or Squoosh for image processing. To ensure your project is still able to build, image processing has been disabled.`
+									`The selected adapter ${adapter.name} does not support image processing. To ensure your project continues to build successfully, the image service has been automatically switched to the 'noop' option, effectively disabling image processing.`
```
alexanderniebuhr, line 238: We don't need to mention `sharp` or `squoosh` again, as we've already stated in the log line above that the current adapter is incompatible with them.

alexanderniebuhr (commented):
self-review for background context

packages/astro/src/integrations/index.ts
```diff
@@ -221,17 +221,21 @@ export async function runHookConfigDone({
 								logger
 							);
 							for (const [featureName, supported] of Object.entries(validationResult)) {
-								if (!supported) {
+								// If `supported` / `validationResult[featureName]` only allows boolean,
+								// in theory 'asstes' false, doesn't mean that the feature is not supported, but rather that the chosen image service is unsupported
```
florian-lefebvre, line 225:
```suggestion
								// in theory 'assets' false, doesn't mean that the feature is not supported, but rather that the chosen image service is unsupported
```

packages/astro/src/integrations/index.ts
Princesseuh: Yeah, the support here is confusing. In my opinion, it would've been better for it to be more about the Node compatibility of the image service vs the specific services.

packages/astro/src/integrations/index.ts
sarah11918: Just checking that people feel that "does not support image processing/disabling image processing" is clear enough that it won't make people think they somehow can't even use/have images in their project. (Out of context of the inner workings of the image services API, "processing" could literally mean "handling/being able to accept".)

It might be worth something slightly more reassuring that mentions what WILL happen to/with your images (which sounds less like images are somehow a problem): they will not be optimized, they will be copied into your build as is, this allows you to use the image component and `getImage()` function although the functionality will be limited to preventing CLS and will not optimize or otherwise process your images.

alexanderniebuhr: @sarah11918 do you have a suggestion for me, or do we wait for other opinions?

ematipico: We can say something like "The selected adapter ${adapter.name} doesn't support image optimization, although they will still be served correctly. To ensure your project ..."

ematipico (approved):
The PR looks good to me. Let's wait for Sarah's approval

packages/astro/src/integrations/index.ts
```diff
@@ -221,17 +221,21 @@ export async function runHookConfigDone({
...
 										`The adapter ${adapter.name} doesn't support the feature ${featureName}. Your project won't be built. You should not use it.`
 									);
 								}
 							}
 							if (!validationResult.assets) {
-								logger.info(
+								logger.warn(
 									'astro',
-									`The selected adapter ${adapter.name} does not support Sharp or Squoosh for image processing. To ensure your project is still able to build, image processing has been disabled.`
+									`The selected adapter ${adapter.name} does not support image processing. To ensure your project continues to build successfully, the image service has been automatically switched to the 'noop' option, effectively disabling image processing.`
```
sarah11918, line 238:
```suggestion
									`The selected adapter ${adapter.name} does not support image processing. To allow your project to build with the original, unprocessed images, the image service has been automatically switched to the 'noop' option.`
```
I'm wondering whether something like this might be sufficient (and more reassuring that you do get images)? 

If it's possible to link for more info somewhere, then this would be a good place https://docs.astro.build/en/reference/configuration-reference/#imageservice (and, I did make myself a docs issue to document a noop option!)

alexanderniebuhr: I don't think we should add a link to the terminal log, but your version sounds better.

alexanderniebuhr: I've updated it & also added the URL, should be working.

sarah11918 (approved):
I think we have a winner! I am sufficiently less frightened about my images as a reader, so docs is signing off!

PR #8620, opened 2023-09-21
feat(assets): Add support for `srcset` and a Picture component

Princesseuh:
## Changes

- What does this change?

### DIFF
--- a/packages/astro/components/Image.astro
+++ b/packages/astro/components/Image.astro
@@ -23,6 +23,11 @@ if (typeof props.height === 'string') {
 }
 
 const image = await getImage(props);
+const additionalAttributes: Record<string, any> = {};
+
+if (image.srcSet.length > 0) {
+	additionalAttributes.srcset = image.srcSetValue;
+}
 ---
 
-<img src={image.src} {...image.attributes} />
+<img src={image.src} {...additionalAttributes} {...image.attributes} />
--- a/packages/astro/components/Picture.astro
+++ b/packages/astro/components/Picture.astro
@@ -1,0 +1,44 @@
+---
+import { LocalImageProps, RemoteImageProps, getImage } from 'astro:assets';
+import { GetImageResult, ImageOutputFormat } from '../dist/@types/astro';
+import { isESMImportedImage } from '../dist/assets/internal';
+import { AstroError, AstroErrorData } from '../dist/core/errors/index.js';
+import { HTMLAttributes } from '../types';
+
+type Props = (LocalImageProps | RemoteImageProps) & {
+	formats: ImageOutputFormat[];
+	fallbackFormat: ImageOutputFormat;
+	pictureAttributes: HTMLAttributes<'picture'>;
+};
+
+const props = Astro.props;
+const optimizedImages: Record<
+	ImageOutputFormat,
+	GetImageResult
+> = await Promise.all(
+	props.formats.map(async (format) => await getImage({ ...props, format: format, widths: props.widths, densities: props.densities }))
+);
+
+const fallbackFormat =
+	props.fallbackFormat ?? isESMImportedImage(props.src)
+		? ['svg', 'gif'].includes(props.src.format)
+			? props.src.format
+			: 'png'
+		: 'png';
+
+const originalImage = await getImage({ ...props, format: fallbackFormat, widths: props.widths, densities: props.densities });
+
+if (props.alt === undefined || props.alt === null) {
+	throw new AstroError(AstroErrorData.ImageMissingAlt);
+}
+---
+
+<picture {...props.pictureAttributes}>
+	{Object.entries(optimizedImages).map(([format, image]) => (
+		<source
+			srcset={image.srcSetValue}
+			type={image.srcSet[0].attributes?.type}
+		/>
+	))}
+	<img src={originalImage.src} {...originalImage.attributes} />
+</picture>
--- a/packages/astro/src/assets/consts.ts
+++ b/packages/astro/src/assets/consts.ts
@@ -24,4 +24,5 @@ export const VALID_SUPPORTED_FORMATS = [
 	'svg',
 	'avif',
 ] as const;
+export const DEFAULT_OUTPUT_FORMAT = 'webp' as const;
 export const VALID_OUTPUT_FORMATS = ['avif', 'png', 'webp', 'jpeg', 'jpg', 'svg'] as const;
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -1,7 +1,7 @@
 import type { AstroConfig } from '../../@types/astro.js';
 import { AstroError, AstroErrorData } from '../../core/errors/index.js';
 import { isRemotePath, joinPaths } from '../../core/path.js';
-import { VALID_SUPPORTED_FORMATS } from '../consts.js';
+import { DEFAULT_OUTPUT_FORMAT, VALID_SUPPORTED_FORMATS } from '../consts.js';
 import { isESMImportedImage, isRemoteAllowed } from '../internal.js';
 import type { ImageOutputFormat, ImageTransform } from '../types.js';
 
@@ -28,6 +28,12 @@ type ImageConfig<T> = Omit<AstroConfig['image'], 'service'> & {
 	service: { entrypoint: string; config: T };
 };
 
+type SrcSetValue = {
+	transform: ImageTransform;
+	descriptor?: string;
+	attributes?: Record<string, any>;
+};
+
 interface SharedServiceProps<T extends Record<string, any> = Record<string, any>> {
 	/**
 	 * Return the URL to the endpoint or URL your images are generated from.
@@ -38,6 +44,13 @@ interface SharedServiceProps<T extends Record<string, any> = Record<string, any>
 	 *
 	 */
 	getURL: (options: ImageTransform, imageConfig: ImageConfig<T>) => string | Promise<string>;
+	/**
+	 * TODO: Document
+	 */
+	getSrcSet?: (
+		options: ImageTransform,
+		imageConfig: ImageConfig<T>
+	) => SrcSetValue[] | Promise<SrcSetValue[]>;
 	/**
 	 * Return any additional HTML attributes separate from `src` that your service requires to show the image properly.
 	 *
@@ -174,6 +187,11 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 				});
 			}
 
+			if (options.widths && options.densities) {
+				console.warn('Cannot use `widths` and `densities` at the same time. Using `densities`.');
+				options.widths = undefined;
+			}
+
 			// We currently do not support processing SVGs, so whenever the input format is a SVG, force the output to also be one
 			if (options.src.format === 'svg') {
 				options.format = 'svg';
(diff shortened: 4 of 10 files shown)


Princesseuh:
!preview picture

Princesseuh:
!preview picture

packages/astro/src/assets/consts.ts
```diff
@@ -24,4 +24,5 @@ export const VALID_SUPPORTED_FORMATS = [
 	'svg',
 	'avif',
 ] as const;
+export const DEFAULT_OUTPUT_FORMAT = 'webp' as const;
```
matthewp, line 27: clever use of const

packages/astro/components/Picture.astro
```diff
@@ -0,0 +1,57 @@
+---
+import { getImage, type LocalImageProps, type RemoteImageProps } from 'astro:assets';
+import type { GetImageResult, ImageOutputFormat } from '../dist/@types/astro';
+import { isESMImportedImage } from '../dist/assets/internal';
+import { AstroError, AstroErrorData } from '../dist/core/errors/index.js';
+import type { HTMLAttributes } from '../types';
+
+type Props = (LocalImageProps | RemoteImageProps) & {
+	formats?: ImageOutputFormat[];
```
matthewp, line 9: Is `svg` ok here?

Princesseuh: Yes! It's okay for it to be used because the input might be a `svg`. Additionally, we don't support it ourselves (and our services does successfully prevents it), but an image service could implement a raster to vector function and fully support using a `svg` here no matter the input format

matthewp (approved):

lilnasy:
!preview picture

Princesseuh:
!preview picture

Princesseuh:
!preview picture

bluwy (approved):

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,49 @@
...
+<Image src={myImage} width={myImage.width / 2} densities={[2]} alt="My cool image"  />
+```
+
+```html
+<img src="..." srcset="... 2x, ... 3x" alt="My cool image" />
+```
+
+## Picture component
+
+The `Picture` component can be used to generate a `<picture>` element with multiple sources. It can be used as follow:
```
sarah11918, line 28:
```suggestion
The experimental `<Picture />` component can be used to generate a `<picture>` element with multiple `<source>` elements. 

The example below uses `format` property to generate a `<source>` in each of the specified image formats:
```
I kind of like making "sources" code here, to distinguish the tag from "multiple sources" sounding like "original sources/locations/files" etc. unless there's a good reason not to do this! (And if "elements" isn't the best word, pick another word like tag or attribute or whatever is technically accurate here. It's also ok just to pluralize as `<source>`s, but that doesn't always visually look the best.)

Also, should probably be re-upping experimental, right?  :sweat_smile:

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,49 @@
...
+```astro
+---
+import { Picture } from "astro:assets";
+import myImage from "./my-image.jpg";
+---
+
+<Picture src={myImage} formats={["avif", "webp"]} alt="My super image in multiple formats!" />
+```
+
+The above code will generate the following:
```
sarah11918, line 39:
```suggestion
The above code will generate the following HTML, and allow the browser to determine the best image to display:
```
I originally came just to add "HTML", but since this will be also material for the blog post draft, I think throwing in a benefit of some kind at the end of the sentence will be helpful for the release blog post AND gives us a bit of a nicer sentence here.

Pick whatever benefit/description you want, but I like a beefier sentence showing off how this is useful here!

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,49 @@
...
+'astro': minor
+---
+
+Add support for generating multiple widths when using the Image component and a Picture component for supporting multiple formats.
+
+## `srcset` support
+
+Two new properties have been added to `Image` and `getImage`: `densities` and `widths`.
+
+These props can be used to generate a `srcset` attribute with multiple sources. For example:
```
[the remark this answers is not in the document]
Princesseuh: The `width / 2` is actually required! Because densities should never upscale, so your base image needs to be smaller than the densities it generates. (You can think of `width` as controlling the base image, whereas `densities` control its variants)

If you have an image that is originally 1280x720, and you ask for density 2x (2560x1440) without specifying a smaller size for the base image, Astro would have to upscale the original image to reach that size, since we never upscale, `density={['2x']}` would just get ignored.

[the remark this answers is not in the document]
sarah11918: I kind of know I knew that at one point... and I agree, no need to get into that here. But it doesn't seem intuitive to me that an array with a single value of 2 produces both a 2x and a 3x in `srcset`. Admittedly, if I understood this better, I'd be better able to say what should/shouldn't go in a description here. Naively, I would have expected to need to specify two values in my array to get two generated.

packages/astro/src/assets/services/service.ts
```diff
@@ -216,6 +221,89 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
...
+		const { targetWidth, targetHeight } = getTargetDimensions(options);
+		const { widths, densities } = options;
+		const targetFormat = options.format ?? DEFAULT_OUTPUT_FORMAT;
+
+		const aspectRatio = targetWidth / targetHeight;
+		const imageWidth = isESMImportedImage(options.src) ? options.src.width : options.width;
+		const maxWidth = imageWidth ?? Infinity;
+
+		// REFACTOR: Could we merge these two blocks?
+		if (densities) {
```
zaknesler, line 235: Firstly, thank you @Princesseuh for your work on this! Such a lovely feature. I've been using the latest previews here for a gallery of mine, just calling `getImage` to use the srcset generation, and it works like a charm. Just one issue that I've come across, but maybe I'm just using it incorrectly!

I think the conditional here for when either `densities` or `widths` is defined means that something like the following is not possible:

```astro
<Picture
  alt="An image in two formats with one width"
  src={image}
  width={700}
  formats={['avif', 'webp']}
/>
```

Which I would think could generate something like the following:

```html
<picture>
  <source srcset="/_astro/my_image_700w.hash.avif" type="image/avif" />
  <source srcset="/_astro/my_image_700w.hash.webp" type="image/webp" />
  <img
    src="/_astro/my_image_700w.hash.jpg"
    width="700"
    height="350"
    decoding="async"
    loading="lazy"
    alt="An image in two formats with one width"
  />
</picture>
```

Where it wouldn't generate a srcset like `[url] [width]w, [url] [width]w`, etc. but instead as the single resized image.

Instead, it results in this error:

```
error   Cannot read properties of undefined (reading 'attributes')
Code:
  51 |         srcset={`${image.src}, ` + image.srcSet.attribute}
> 52 |         type={image.srcSet.values[0].attributes?.type}
     |                                      ^
```

Basically, I think that it's getting to this point and not generating any srcSet values. If instead I defined it as `widths={[700]}` it would produce the 700w image and the original size in the srcset, when I'd just like to resize the image to 700px and have multiple file types.

Is this a bug or is there another way to achieve this with this impl? Maybe just out of scope?

.changeset/smooth-goats-agree.md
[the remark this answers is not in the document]
Princesseuh: Oh, it doesn't sorry!

If you do `densities={[2]}`, it'll only have the original density (in the `src` attribute) and the 2x in `srcset`.

packages/astro/src/assets/services/service.ts
Princesseuh: Fixed in the latest commit, sorry for the inconvenience!

The problem wasn't the block of code you shared, but the fact that we used the `srcset` generated to find the attributes to use on the different formats, so when you had no srcset (like you shared, only one width), we wouldn't be able to find the proper attributes.

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,56 @@
...
+'astro': minor
+---
+
+Adds experimental support for generating `srcset` attributes and a new `<Picture />` component.
+
+## `srcset` support
+
+Two new properties have been added to `Image` and `getImage()`: `densities` and `widths`.
+
+These properties can be used to generate a `srcset` attribute, either based on absolute widths in pixels (e.g. [300, 600, 900]) or pixel density descriptors (e.g. `["2x"]` or `[1.5, 2]`).
```
sarah11918, line 11: love the pixel density descriptors showing both unit options!

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,56 @@
...
+import myImage from "./my-image.jpg";
+---
+
+<Image src={myImage} width={myImage.width / 2} densities={[1.5, 2]} alt="My cool image"  />
+```
+
+```html
+<img
+  src="/_astro/my_image.hash.webp"
+  srcset="/_astro/my_image.hash.webp 1.5x, /_astro/my_image.hash.webp 2x"
```
sarah11918, line 26: These look way more instructive now! :raised_hands:

.changeset/smooth-goats-agree.md
```diff
@@ -0,0 +1,56 @@
...
+  alt="My cool image"
+/>
+```
+
+## Picture component
+
+The experimental `<Picture />` component can be used to generate a `<picture>` element with multiple `<source>` elements.
+
+The example below uses `format` property to generate a `<source>` in each of the specified image formats:
```
sarah11918, line 35:
```suggestion
The example below uses the `format` property to generate a `<source>` in each of the specified image formats:
```
grammar nit

sarah11918 (approved):
I love where the changset landed! I just saw one tiny missing "the" somewhere, but good to go!

jasikpark (approved):
looks great! thanks so much for championing this!

packages/astro/components/Image.astro
```diff
@@ -23,6 +23,12 @@ if (typeof props.height === 'string') {
 }
 
 const image = await getImage(props);
+
+const additionalAttributes: Record<string, any> = {};
```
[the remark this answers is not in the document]
Princesseuh: Mostly so we don't have to check the attributes before using them, since it doesn't really matter (it's classic Astro serializing, so it supports everything)

packages/astro/src/assets/services/service.ts
```diff
@@ -260,3 +348,28 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
...
+			// If we have a width but no height, use width to calculate the height
+			targetHeight = Math.round(targetWidth / aspectRatio);
+		} else if (!targetWidth && !targetHeight) {
+			// If we have neither width or height, use the original image's dimensions
+			targetWidth = options.src.width;
+			targetHeight = options.src.height;
+		}
+	}
+
+	// TypeScript doesn't know this, but because of previous hooks we always know that targetWidth and targetHeight are defined
```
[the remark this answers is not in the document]
Princesseuh: Yeah, we could probably accept type params on image services, and have like a prop for validated props and stuff.

mks-h:
Thanks for your work! I was looking forward to it. I'll go and use it instead of my own little script, as soon as I get to it.

PR #8621, opened 2023-09-21
Add View Transitions announcer

matthewp:
## Changes

- Adds an announcer to announce the page for AT

## Testing

- Added a new example in `examples/view-transitions` that we can use to test this.

## Docs

- https://github.com/withastro/docs/pull/4786


packages/astro/components/ViewTransitions.astro
```diff
@@ -30,6 +30,15 @@ const { fallback = 'animate' } = Astro.props as Props;
...
 	const triggerEvent = (name: Events) => document.dispatchEvent(new Event(name));
 	const onPageLoad = () => triggerEvent('astro:page-load');
+	const announce = () => {
+		let div = document.createElement('div');
+		div.setAttribute('aria-live', 'assertive');
+		div.setAttribute('aria-atomic', 'true');
+		div.setAttribute('style', 'position:absolute;left:0;top:0;clip:rect(0 0 0 0);clip-path:inset(50%);overflow:hidden;white-space:nowrap;width:1px;height:1px');
+		let title = document.title || document.querySelector('h1')?.textContent || location.pathname;
+		div.textContent = title;
+		document.body.append(div);
```
Yan-Thomas, line 40: What I believe is happening is since this node is supposedly created for each page on load with its according title, there's not a change to be announced by `aria-live` in the first place. 

I've tried a few changes to fix this, note that doing this **DOES NOT SOLVE IT** (maybe JavaScript is doing some optimization and forcing the text to be added before appended?):
```js
document.body.append(div);
let title = document.title || document.querySelector('h1')?.textContent || location.pathname;
div.textContent = title;
```

What **DID WORK**, was forcing the change to happen only after the node is for sure appended:

```js
document.body.append(div);
setTimeout(() => {
	let title = document.title || document.querySelector('h1')?.textContent || location.pathname;
	div.textContent = title;
}, 50);
```

Yan-Thomas (changes requested):
Looks like the way we implemented the announcement in `ViewTransitions.astro` isn't working on the NVDA/Windows/Chrome combo!

packages/astro/components/ViewTransitions.astro
matthewp: Interesting, thanks will investigate further.

matthewp: @Yan-Thomas updated the implementation with your suggestion but omitted the 50 ms. Curious if that long of a wait is needed or not.

Yan-Thomas: Without the wait, or even, smaller waits, the behavior starts to get confusing and sometimes it is announced, some others not. I found a [helpful issue comment with explanation](https://github.com/nvaccess/nvda/issues/8873#issuecomment-432935068), TL;DR screen readers are coalescing the creation and update events and for that reason, it doesn't get announced as a change. Also, I'd even recommend going up to `60` so it's on the same refresh rate as browsers as suggested by the comment.

.changeset/small-rules-relax.md
```diff
@@ -0,0 +1,11 @@
...
+'astro': minor
+---
+
+Route Announcer in `<ViewTransitions />`
+
+The View Transitions router now does route announcement. When transitioning between pages with a traditional MPA approach assistive technologies will announce the page title when the page finishes loading. When you introduce client-side routing into an app this no longer naturally happens, so the user isn't aware that the page has changed.
+
+A route announcer runs after the page loads, for Astro this is the `astro:page-load` event, giving the title of the page to be the first thing announced. The ViewTransitions announcer first looks for the `<title>`, and if there is not one falls back to the first `<h1>` it finds, or otherwise announces the pathname. We recommend you always include a `<title>` in each page.
+
+See the ViewTransitions docs for more on how accessibility is handled.
```
sarah11918, line 11:
```suggestion
The View Transitions router now does route announcement. When transitioning between pages with a traditional MPA approach, assistive technologies will announce the page title when the page finishes loading. This does not automatically happen during client-side routing, so visitors relying on these technologies to announce routes are not aware when a page has changed.

The view transitions route announcer runs after the `astro:page-load` event, looking for the page `<title>` to announce. If one cannot be found, the announcer falls back to the first `<h1>` it finds, or otherwise announces the pathname. We recommend you always include a `<title>` in each page for accessibility.

See the [View Transitions docs](https://docs.astro.build/en/guides/view-transitions/) for more on how accessibility is handled.
```
Just tightened up some sentences and added the link to docs!

sarah11918 (commented):
Just suggested some tighter wording for the changeset, for your consideration!

.changeset/small-rules-relax.md
matthewp: Wonderful, thanks!

Yan-Thomas (approved):
Docs & Accessibility LGTM! 🙌

packages/astro/components/ViewTransitions.astro
```diff
@@ -30,6 +30,21 @@ const { fallback = 'animate' } = Astro.props as Props;
 		!!document.querySelector('[name="astro-view-transitions-enabled"]');
 	const triggerEvent = (name: Events) => document.dispatchEvent(new Event(name));
 	const onPageLoad = () => triggerEvent('astro:page-load');
+	const announce = () => {
+		let div = document.createElement('div');
+		div.setAttribute('aria-live', 'assertive');
+		div.setAttribute('aria-atomic', 'true');
+		div.setAttribute('style', 'position:absolute;left:0;top:0;clip:rect(0 0 0 0);clip-path:inset(50%);overflow:hidden;white-space:nowrap;width:1px;height:1px');
```
ematipico, line 37: Maybe let's leave a comment that tells us what this style does and why. +1 for a link that explains it.

packages/astro/components/ViewTransitions.astro
```diff
@@ -30,6 +30,21 @@ const { fallback = 'animate' } = Astro.props as Props;
 		!!document.querySelector('[name="astro-view-transitions-enabled"]');
 	const triggerEvent = (name: Events) => document.dispatchEvent(new Event(name));
 	const onPageLoad = () => triggerEvent('astro:page-load');
+	const announce = () => {
+		let div = document.createElement('div');
```
ematipico, line 34: Are we going to create this div every time? Is there a risk where this div is created more than once in the same page?

ematipico (approved):

packages/astro/components/ViewTransitions.astro
matthewp: Yeah, the incoming page wipes out the previous one, so it won't be retained between page loads.

examples/view-transitions/README.md
```diff
@@ -0,0 +1,31 @@
+# Astro Movies View Transitions Demo
+
+### 👉🏽 [Live Demo](https://astro-movies.pages.dev/)
+
+![Screenshot](./screenshot.png)
+
+## 🚀 Getting Started
+
+1. Clone this repository and install dependencies with `npm install`.
+2. Copy `.env.example` as `.env` and add your TMDB API Key.
```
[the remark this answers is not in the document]
matthewp: Ahh, will remove that. I simplified the example to not use the API.

matthewp:
Wait on permission from the original author of the example included here. If I don't get the permission I'll remove it from this PR.

matthewp:
Maxi said it was ok, so merging.

PR #8628, opened 2023-09-22
feat(i18n): default locale system for dev

ematipico:
## Changes

Depends on https://github.com/withastro/astro/pull/8607

Addresses PLT-982

This PR adds new logic to **development** to retrieve the content of the `defaultLocale` in case the page of a locale doesn't exist.

Given the following config:

```js
{
	defaultLocale: 'en',
	locales: ['en', 'pt']
}
```

And let's assume we have the following pages
```
/src/pages/en/welcome
/src/pages/en/first-steps
/src/pages/pt/hola
```

If we visit `/pt/first-steps`, Astro will **render the content** of the page `/src/pages/en/welcome`. URL will stay the same.

## Testing

Added new test cases.

## Docs

N/A for now

### DIFF
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -335,6 +335,11 @@ export function createRouteManifest(
 				const route = `/${segments
 					.map(([{ dynamic, content }]) => (dynamic ? `[${content}]` : content))
 					.join('/')}`.toLowerCase();
+				const locale = settings.config.experimental.i18n?.locales.find((currentLocale) => {
+					if (route.startsWith(`/${currentLocale}`)) {
+						return currentLocale;
+					}
+				});
 				routes.push({
 					route,
 					type: item.isPage ? 'page' : 'endpoint',
@@ -345,6 +350,7 @@ export function createRouteManifest(
 					generate,
 					pathname: pathname || undefined,
 					prerender,
+					locale,
 				});
 			}
 		});
@@ -407,6 +413,11 @@ export function createRouteManifest(
 					`An integration attempted to inject a route that is already used in your project: "${route}" at "${component}". \nThis route collides with: "${collision.component}".`
 				);
 			}
+			const locale = settings.config.experimental.i18n?.locales.find((currentLocale) => {
+				if (route.startsWith(`/${currentLocale}`)) {
+					return currentLocale;
+				}
+			});
 
 			// the routes array was already sorted by priority,
 			// pushing to the front of the list ensure that injected routes
@@ -421,6 +432,7 @@ export function createRouteManifest(
 				generate,
 				pathname: pathname || void 0,
 				prerender: prerenderInjected ?? prerender,
+				locale,
 			});
 		});
 
@@ -448,6 +460,11 @@ export function createRouteManifest(
 			.map(([{ dynamic, content }]) => (dynamic ? `[${content}]` : content))
 			.join('/')}`.toLowerCase();
 
+		const locale = settings.config.experimental.i18n?.locales.find((currentLocale) => {
+			if (route.startsWith(`/${currentLocale}`)) {
+				return currentLocale;
+			}
+		});
 		const routeData: RouteData = {
 			type: 'redirect',
 			route,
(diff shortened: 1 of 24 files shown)


packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -335,6 +335,11 @@ export function createRouteManifest(
 				const route = `/${segments
 					.map(([{ dynamic, content }]) => (dynamic ? `[${content}]` : content))
 					.join('/')}`.toLowerCase();
+				const locale = settings.config.experimental.i18n?.locales.find((currentLocale) => {
+					if (route.includes(`/${currentLocale}`)) {
+						return currentLocale;
+					}
+				});
```
bluwy, line 342: Do we allow the locale directory to be anywhere in the route? e.g. `src/pages/foo/bar/pt/content.astro`. I think it's most common being strictly `src/pages/pt/...` and maybe we should change the check here from `.includes` to `.startsWith`.

And perhaps check with `/${currentLocale}/` instead (two wrapping slashes), so `pt` doesn't accidentally match `/pt-br/...`.

Would be nice to wrap this as a function too.

packages/astro/src/core/routing/match.ts
```diff
@@ -9,3 +9,53 @@ export function matchRoute(pathname: string, manifest: ManifestData): RouteData
...
+): RouteData[] {
+	const matchedRoutes: RouteData[] = [];
+	// SAFETY: the function is called upon checkin if `experimental.i18n` exists first
+	const defaultLocale = config!.defaultLocale;
+
+	for (const route of manifest.routes) {
+		// we don't need to check routes that don't belong to the default locale
+		if (route.locale === defaultLocale) {
+			// we check if the current route pathname contains `/en` somewhere
+			if (route.pathname?.includes(`/${defaultLocale}`)) {
```
bluwy, line 40: Same check here.

packages/astro/src/core/routing/manifest/create.ts
matthewp: Yes, I believe the plan is that it has to be the route, as this is how it works in other frameworks and in Starlight. We discussed some ways to make it configurable but haven't come up with a good solution as of yet.

packages/astro/package.json
```diff
@@ -77,7 +77,8 @@
       "types": "./dist/core/middleware/namespace.d.ts",
       "default": "./dist/core/middleware/namespace.js"
     },
-    "./transitions": "./dist/transitions/index.js"
+    "./transitions": "./dist/transitions/index.js",
+    "./i18n": "./dist/i18n/index.js"
```
matthewp, line 81: Since we have had to do this so many times for our namespaced modules, maybe we should have a convention for this to avoid repeating the exports, like `"./modules/*": "./dist/modules/*"`. Then anything we put in that folder would be exposed. Not something to change in this PR, just a thought for a future refactor.

ematipico: I'd would to defer this change to someone else, preferably in another PR.

packages/astro/src/core/routing/manifest/create.ts
ematipico: > .includes to .startsWith

My initial implementation used `startsWith`,  but that wouldn't work in cases where the user decides to use `base` in their configuration. 

> And perhaps check with /${currentLocale}/ instead (two wrapping slashes), so pt doesn't accidentally match /pt-br/....

That wouldn't work in cases we a plain `index` that hasn't been translated.

So if you have `locales: ['en',  'it']` and you **don't have** the page `src/pages/it/index.mdx`, then when you call `example.com/it`, you'd want to show the english version.

ematipico: > We can get the base from settings.config.base here.

Yeah I know, but this code isn't meant to deal with `base` yet 😅 see the tests. Do you want me to deal with `base` in this PR?

bluwy: If the code here is meant as a quick prototype of the feature, then I can definitely close one eye here and have it polished later 😄 Just pointing out a bug that could happen here since I'm not sure in further reviews I'll see this diff again.

I don't mind if you want to skip this for now, and fix it later though. But I think it should be fixed after things are settled.

bluwy (approved):
Thanks for expanding on the base and locale checks. Looks fine by me for now as an early implementation if you need to merge.

packages/astro/src/core/routing/manifest/create.ts
ematipico: Added the checks

PR #8636, opened 2023-09-22
fix: no deletion of scripts during view transition

martrapp:
## Changes

there were some problems with the deletion of scripts during view transition (to prevent re-execution)
No they are not deleted anymore but simply marked as already executed

## Testing

pass existing e2e tests

## Docs

bug fix, n/a

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -219,13 +219,13 @@ const { fallback = 'animate' } = Astro.props as Props;
 				for (const s2 of newDocument.scripts) {
 					if (
 						// Inline
-						(s1.textContent && s1.textContent === s2.textContent) ||
+						(!s1.src && s1.textContent === s2.textContent) ||
 						// External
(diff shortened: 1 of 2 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -219,13 +219,13 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+						(!s1.src && s1.textContent === s2.textContent) ||
 						// External
-						(s1.type === s2.type && s1.src === s2.src)
+						(s1.src && s1.type === s2.type && s1.src === s2.src)
 					) {
-						s2.remove();
-					} else {
-						s1.remove();
+						// the old script is in the new document: we mark it as executed to prevent re-execution
+						s2.dataset.astroExec = '';
```
matthewp, line 227: very clever, this also means they will exist on the next page transition, so they won't be removed there either.

matthewp (approved):

PR #8640, opened 2023-09-22
Warn on empty content collections

matthewp:
## Changes

- There was a mistake in https://github.com/withastro/astro/pull/8382 that returned an integration when it should have just warned. This replaces that.
- This also deprecates the error that no longer occurs.

## Testing

- Test added in https://github.com/withastro/astro/pull/8382 will still pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -1,6 +1,5 @@
 import type { MarkdownHeading } from '@astrojs/markdown-remark';
 import { ZodIssueCode, string as zodString } from 'zod';
-import type { AstroIntegration } from '../@types/astro.js';
 import { AstroError, AstroErrorData } from '../core/errors/index.js';
 import { prependForwardSlash } from '../core/path.js';
 import {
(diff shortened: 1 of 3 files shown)


packages/astro/src/content/runtime.ts
```diff
@@ -56,7 +55,9 @@ export function createGetCollection({
 		} else if (collection in dataCollectionToEntryMap) {
 			type = 'data';
 		} else {
-			return warnOfEmptyCollection(collection);
+			// eslint-disable-next-line no-console
+			console.warn(`The collection **${collection}** does not exist or is empty. Ensure a collection directory with this name exists.`);
```
natemoo-re, line 59: Should this use the `logger` so it respects the configured `logLevel`?

matthewp: This is the runtime so there is no access to the logger. Same reason we use console.warn in src/runtime/*

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1158,6 +1158,7 @@ export const ContentSchemaContainsSlugError = {
 /**
  * @docs
  * @message A collection queried via `getCollection()` does not exist.
+ * @deprecated Collections that do not exist are no longer an error. A warning is omitted instead.
```
TheOtterlord, line 1161: error nit
```suggestion
 * @deprecated Collections that do not exist no longer result in an error. A warning is omitted instead.
```

TheOtterlord (commented):
looks good :+1:

PR #8641, opened 2023-09-22
Update index.ts

ohrrkan:
## Changes

Remove false warnings from Deno integration adapter  build.

Deno integration adapter  issues warnings when build with astro build. These warnings typically arise from packages labeled with "sideEffects": false," which are included in the Vite built output because they are considered external.


matthewp:
The Deno adapter is now maintained here: https://github.com/denoland/deno-astro-adapter

Will be removed from this repo soon.

PR #8642, opened 2023-09-22
Fix getCollection when collection is empty

TheOtterlord:
## Changes

Fixes `getCollection` when the collection is empty. `3.1.2` included #8382 which was meant to prevent an error from throwing, and replace it with a warning. The implementation instead returned an Astro integration from `getCollection`. This PR removes that, and replaces it with a `console.warn` before returning an empty array.

#### Other options

- We could revert the PR entirely,.
- We can return `undefined` if we want to explicitely show the collection does not exist.

## Testing

The previous PR had testing to confirm it was no longer throwing an error.

## Docs

I need to check if docs need updating, but in any case the current behaviour is already undocumented. @Princesseuh did mention that the previous PR deprecated an error, so that will likely also need addressing.


TheOtterlord:
Haha, looks like I missed #8640
Closing as duplicate!

PR #8645, opened 2023-09-22
fix getDataEntryById to lookup by base name

matthewp:
## Changes

- Found this bug while working on a VT demo.
- Was expecting the lookup map to be by base id, but it is not, so instead using `createGlobLookup` like the other lookup functions do.

## Testing

- New test case added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -137,13 +137,12 @@ export function createGetEntryBySlug({
 }
 
 export function createGetDataEntryById({
-	dataCollectionToEntryMap,
+	getEntryImport,
 }: {
-	dataCollectionToEntryMap: CollectionToEntryMap;
+	getEntryImport: GetEntryImport;
 }) {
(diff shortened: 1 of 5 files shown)


packages/astro/src/content/runtime.ts
```diff
@@ -137,13 +137,12 @@ export function createGetEntryBySlug({
...
 export function createGetDataEntryById({
-	dataCollectionToEntryMap,
+	getEntryImport,
 }: {
-	dataCollectionToEntryMap: CollectionToEntryMap;
+	getEntryImport: GetEntryImport;
 }) {
 	return async function getDataEntryById(collection: string, id: string) {
-		const lazyImport =
-			dataCollectionToEntryMap[collection]?.[/*TODO: filePathToIdMap*/ id + '.json'];
```
bholmesdev, line 146: Oh my lord... that shipped? 😰

bholmesdev (approved):
Thanks for fixing this. I oh you a whiskey sour jsomeday 🙏

PR #8646, opened 2023-09-22
Head propagation graph walking on new pages

matthewp:
## Changes

- We walk the graph when we encounter a module that has head content, such as `transition:` directives, walking up parents to mark them with metadata `in-tree`.
- If you load a page that contains a module with head content, and then later load another page (as happens in dev), we need to check dependencies so that we can mark the new page with the `in-tree` if needed.
- Fixes https://github.com/withastro/astro/issues/8392

## Testing

- New fixture test added mimicking https://github.com/withastro/astro/issues/8392

## Docs

N/A, bug fix


packages/astro/src/vite-plugin-head/index.ts
```diff
@@ -45,9 +45,34 @@ export default function configHeadVitePlugin(): vite.Plugin {
...
+			if(importer) {
+				// Do propagation any time a new module is imported. This is because
+				// A module with propagation might be loaded before one of its parent pages
+				// is loaded, in which case that parent page won't have the in-tree and containsHead
+				// values. Walking up the tree in resolveId ensures that they do
+				return this.resolve(source, importer, { skipSelf: true }).then(result => {
+					if(result) {
+						let info = this.getModuleInfo(result.id);
+						if(info && getAstroMetadata(info)) {
+							const astro = getAstroMetadata(info)!;
```
ematipico, line 63: Can't we save the result of `getAstroMetadata(info)` in a variable and avoid query the same info twice?

ematipico (approved):

packages/astro/src/vite-plugin-head/index.ts
matthewp: We could, it's just a property accessor though (`info.meta.astro`), there's no querying, so no performance issue here. I'd do it if you think it improves readability.

matthewp: I think it does improve it slightly so I went ahead and did so

packages/astro/src/vite-plugin-head/index.ts
```diff
@@ -45,9 +45,34 @@ export default function configHeadVitePlugin(): vite.Plugin {
 
 	return {
 		name: 'astro:head-metadata',
+		enforce: 'pre',
+		apply: 'serve',
 		configureServer(_server) {
 			server = _server;
 		},
+		resolveId(source, importer) {
```
bluwy, line 53: If we're handling the propagation in `resolveId`. Could we remove the similar code in the `transform` hook entirely? Or do they work differently?

matthewp: I was thinking about this... I think we probably could. I don't fully trust out test coverage though. I'd rather leave a TODO and try and remove it later when feeling more brave.

bluwy: Works for me 👍 Probably worth circle back whenever possible as it touches on perf

Issue #8257, opened 2023-08-28
Returning `Response(null, { status: 404 })` from middleware makes `astro dev` hang


itsmatteomanf:
### Astro info

```block
Astro version            v2.10.14
Package manager          pnpm
Platform                 darwin
Architecture             arm64
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/tailwind
```

### What browser are you using?

Safari, Chrome

### Describe the Bug

If I try and return `new Response(null, { status: 404 })` from the middleware (works in a page), the process `astro dev` hangs indefinitely, forcing a manual restart.

### What's the expected result?

Returning the `404.astro` page.

### Link to Minimal Reproducible Example

withastro-astro-issue-5186

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Couldn't reproduce, the dev server continues to serve requests and handle module updates as expected.

Could you check again?

itsmatteomanf:
Stackblitz's example fails, too... Other than telling you that's the case, I have no idea how to test again, given it behaves like that to me (and even in the minimal example above). The page shows blank, while it should return the 404 page given what the middleware tells it to return.

ematipico:
Are you testing in development or in production? Seeing you're using the Cloudflare adapter, it could be the case.

ematipico:
Is it possible that the middleware doesn't run?

itsmatteomanf:
> Is it possible that the middleware doesn't run?

I doubt that, given that if I put anything else in there it works just as well. Even returning `new Response("hello", { status: 200 })` works.

itsmatteomanf:
Adding that I am expecting to return the 404 page on a page that actually exists, so middleware should definitely run.
I'm pretty sure middleware runs even on routes that return 404, though.

lilnasy:
404.astro wasn't there when I tested yesterday. I can confirm the issue now.

itsmatteomanf:
Oh, I'm sure I added it... it might not have saved it and it did when I switched to that tab again after you posted.

lilnasy:
Just tested by adding the node adapter to the repro, it works fine.

ematipico:
I am going ahead and close the issue. This is a feature, and it requires some internal changes, which should also be reflected in the respective adapters.

One of our maintainers - @lilnasy - created a proposal that would solve this issue: https://github.com/withastro/roadmap/discussions/681

If you need it, please upvote it!

[...]

PR #8648, opened 2023-09-23
fix(rerouting): check that the new route is different

lilnasy:
## Changes
- Fixes closed issue #8257
- The issue has to do with implicit rerouting which will be deprecated and replaced with an [explicit API](https://github.com/withastro/roadmap/discussions/681) but that may take a while.

## Testing
Added 4 cases based on what I saw people encounter in the issue and on discord.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -220,6 +220,7 @@ export async function handleRoute({
 	let response = await pipeline.renderRoute(renderContext, mod);
 	if (response.status === 404 && has404Route(manifestData)) {
 		const fourOhFourRoute = await matchRoute('/404', manifestData, pipeline);
+		if (fourOhFourRoute?.route !== options.route)
 		return handleRoute({
 			...options,
 			matchedRoute: fourOhFourRoute,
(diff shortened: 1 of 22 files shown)


ematipico (approved):
Code looks good to me. I wonder if we can add a test case, the same we have in the issue

lilnasy:
Let's put a potential endless loop in our CI 🥳

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -220,6 +220,7 @@ export async function handleRoute({
 	let response = await pipeline.renderRoute(renderContext, mod);
 	if (response.status === 404 && has404Route(manifestData)) {
 		const fourOhFourRoute = await matchRoute('/404', manifestData, pipeline);
+		// if (fourOhFourRoute?.route !== options.route)
```
lilnasy, line 223: Commenting out the fix to see what happens to CI.

lilnasy: CI times out. Logs for `astro:test` [never appear](https://github.com/withastro/astro/actions/runs/6292593299/job/17092598249?pr=8648#step:9:747). It wouldn't be possible to tell if the cases failed.

lilnasy:
They seem to be synchronous, so there's not stopping them.

ematipico:
Well, let's not overthink it then. You tested manually :)

lilnasy:
And we don't have the bug in our codebase anymore! 🥳

_at least this particular instance of it_

matthewp (approved):

PR #8661, opened 2023-09-25
fix: bundle netlify functions as ESM to support top-level await

Skn0tt:
## Changes

- What does this change?

https://github.com/withastro/astro/pull/8598 included a change that makes Astro emit functions with top-level await, specifically this line:

https://github.com/withastro/astro/blob/863f5171e8e7516c9d72f2e48ea7db1dea71c4f5/packages/astro/src/vite-plugin-markdown/index.ts#L146

Top-Level await is supported in ESM, but not in CJS. By default, Netlify bundles `.mjs` files into CJS files under the hood, so this will result in build failures. To work around that, this PR adds a config file that tells Netlify to bundle into ESM.

I've tested this out on https://github.com/Hiroki1112/hiroki-dev, which is the site I first noticed this bug with. My change fixed their build process.

### DIFF
--- a/packages/integrations/netlify/src/integration-functions.ts
+++ b/packages/integrations/netlify/src/integration-functions.ts
@@ -1,5 +1,6 @@
 import type { AstroAdapter, AstroConfig, AstroIntegration, RouteData } from 'astro';
-import { extname } from 'node:path';
+import { extname, join } from 'node:path';
+import { writeFile } from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
 import { generateEdgeMiddleware } from './middleware.js';
 import type { Args } from './netlify-functions.js';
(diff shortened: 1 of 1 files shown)


packages/integrations/netlify/src/integration-functions.ts
```diff
@@ -85,6 +86,16 @@ function netlifyFunctions({
...
+				const functionsConfig = {
+						version: 1,
+						config: {
+							nodeModuleFormat: "esm"
+						}
+				}
+				const functionsConfigPath = join(fileURLToPath(_config.build.server), "entry.json")
+				await writeFile(functionsConfigPath, JSON.stringify(functionsConfig))
+
+				// await writeFile(_config.build.server)
```
ematipico, line 98: To remove?

Skn0tt: Yes, to remove! Not sure what happened there, should get some coffee :D

packages/integrations/netlify/src/integration-functions.ts
```diff
@@ -85,6 +86,16 @@ function netlifyFunctions({
 				}
 			},
 			'astro:build:done': async ({ routes, dir }) => {
+				const functionsConfig = {
+						version: 1,
+						config: {
+							nodeModuleFormat: "esm"
+						}
```
ematipico, line 93: Is this documented somewhere? 👀

ematipico (approved):

packages/integrations/netlify/src/integration-functions.ts
Skn0tt: No, sadly not! It's more of a stop-gap solution until we start outputting to ESM for all functions - but there's always sites that break in that migration, so we're pretty careful about it. There's something in the works on this though, i'm hoping we can remove this again in a couple months! Will let you know about it then :)

Skn0tt:
Thanks for the quick turnaround!

PR #8666, opened 2023-09-26
fix: print image path for NoImageMetadata

### LINKED ISSUE #8665: `NoImageMetadata` does not log the image path
### Astro Info

```block
Astro                    v3.1.2
Node                     v20.7.0
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When an image fails due to no metadata, you get the error message:

```
NoImageMetadata: Could not process image metadata for `${imagePath}`.
     at imageMetadata (file:///.../node_modules/astro/dist/assets/utils/metadata.js:6:11)
[...
   loc: { file: 'src/pages/[...slug].astro' },
   title: 'Could not process image metadata.',
   hint: 'This is often caused by a corrupted or malformed image. Re-exporting the image from your image editor may fix this issue.',
   frame: undefined,
   type: 'AstroError'
 }
```

I think the intention is to print the path of the image, but instead it's not printed due to this line:

https://github.com/withastro/astro/blob/60684fad72a625b1a7aee1ab9a9169574a1c780c/packages/astro/src/core/errors/errors-data.ts#L652

N/A

### Participation

nolanlawson:
Fixes #8665

## Changes

- Fixes the error message so that it includes the image path

## Testing

I did not add a test or repro, because the bug and fix seem fairly clear to me from the code.

## Docs

I don't think docs need to be updated.

### DIFF
--- a/.changeset/gentle-rocks-enjoy.md
+++ b/.changeset/gentle-rocks-enjoy.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix NoImageMetadata image path error message
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -649,7 +649,7 @@ export const NoImageMetadata = {
 	name: 'NoImageMetadata',
 	title: 'Could not process image metadata.',
 	message: (imagePath: string | undefined) =>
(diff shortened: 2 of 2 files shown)


bluwy (approved):
Looks like a simple change, thanks for fixing it!

PR #8670, opened 2023-09-26
Fix assets not getting optimized when `outDir` is outside the CWD

### LINKED ISSUE #8668: astro:assets breaks when outDir is set to a folder that's outside the project's folder
### Astro Info

```block
Astro                    v3.1.4
Node                     v20.5.0
Package Manager          pnpm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

In Astro's config, when specifying an `outDir` that's outside the project, Astro assets breaks & fails to build.

Output when trying to build the project, where `outDir: "../dist-folder-outside-the-project`
```
$ pnpm build

> astro-project@0.0.1 build /Users/michailk/Documents/astro-images-bug/astro-project
> astro build

12:02:00 PM [content] No content directory found. Skipping type generation.
12:02:00 PM [build] output target: static
12:02:00 PM [build] Collecting build info...
12:02:00 PM [build] Completed in 39ms.
12:02:00 PM [build] Building static entrypoints...
12:02:01 PM [build] Completed in 439ms.

 generating static routes 
▶ src/pages/index.astro
  └─ /index.html (+9ms)

 generating optimized images 
 error   ENOENT: no such file or directory, open '/Users/michailk/Documents/astro-images-bug/out-folder-outside-the-project/_astro/image.28b2a2f0.png'
Error: ENOENT: no such file or directory, open '/Users/michailk/Documents/astro-images-bug/out-folder-outside-the-project/_astro/image.28b2a2f0.png'
 ELIFECYCLE  Command failed with exit code 1.
```

It seems like something is expecting to find the unoptimized image inside the `outDir`, while in reality, the image is located in `astro-project/.astro/_astro/`.

I've experimented a little further and found that:
- The build will still succeed if the optimized images are cached. It'll fail when images get changed or added.
- interestingly, setting the `outDir` anywhere within the project, like `./a-folder/inside/the/project` seems to be working just fine.

For the project to... well, build... with optimized images.

https://github.com/MichailiK/astro-images-outdir-bug

### Participation

MichailiK:
## Changes

- Resolves #8668 
- Fixes a build failure when using `astro:assets` and setting `outDir` to a folder outside of the project. For example `outDir: '../outside-the-project'`
- It's caused by some assets being constrained in the project folder/cwd, by using `getOutDirWithinCwd`
- `astro:assets` does not use `getOutDirWithinCwd`, so it ends up looking for the unoptimised assets at the wrong place

I wrote a slightly deeper explanation in https://github.com/withastro/astro/issues/8668#issuecomment-1735465977.

## Testing

The minimal reproducible example in the issue above (https://github.com/MichailiK/astro-images-outdir-bug) was used to test this change. I don't think this would cause regressions. I believe this change just ensures that Astro is trying to obtain the unoptimised asset from the correct folder. 

## Docs

Since this is a bug, caused by an inconsistency between where `astro:assets` is looking for the assets & where they are actually located, I don't believe this needs docs changes.

### DIFF
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -49,7 +49,7 @@ export async function generateImage(
 		serverRoot = config.build.server;
 		clientRoot = config.build.client;
 	} else {
-		serverRoot = config.outDir;
+		serverRoot = getOutDirWithinCwd(config.outDir);
 		clientRoot = config.outDir;
 	}
 
@@ -104,13 +104,9 @@ export async function generateImage(
 
 	// If the image is local, we can just read it directly, otherwise we need to download it
 	if (isLocalImage) {
-		const outDir = isServerLikeOutput(config)
-			? config.build.server
-			: getOutDirWithinCwd(config.outDir);
-
 		imageData = await fs.promises.readFile(
 			new URL(
-				prependForwardSlash(join(outDir.pathname + config.build.assets, basename(originalImagePath))),
+				'.' + prependForwardSlash(join(config.build.assets, basename(originalImagePath))),
 				serverRoot
 			)
 		);
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -7,7 +7,7 @@ import { getConfiguredImageService, isESMImportedImage } from '../internal.js';
 import type { LocalImageService } from '../services/service.js';
 import type { ImageMetadata, ImageTransform } from '../types.js';
 import { loadRemoteImage, type RemoteCacheEntry } from './remote.js';
-import { getOutDirWithinCwd } from "#astro/core/build/common";
+import { getOutDirWithinCwd } from "../../core/build/common.js";
 
 interface GenerationDataUncached {
 	cached: false;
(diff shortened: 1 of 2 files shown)


packages/astro/src/assets/build/generate.ts
```diff
@@ -103,9 +104,13 @@ export async function generateImage(
 
 	// If the image is local, we can just read it directly, otherwise we need to download it
 	if (isLocalImage) {
+		const outDir = isServerLikeOutput(config)
+			? config.build.server
+			: getOutDirWithinCwd(config.outDir);
```
bluwy, line 109: It might be better to use `getOutDirWithinCwd()` on this line instead:

https://github.com/withastro/astro/blob/2365c124645d5067a12987f205cee23a45d1d13d/packages/astro/src/assets/build/generate.ts#L51

packages/astro/src/assets/build/generate.ts
```diff
@@ -103,9 +104,13 @@ export async function generateImage(
...
+
 		imageData = await fs.promises.readFile(
 			new URL(
-				'.' + prependForwardSlash(join(config.build.assets, basename(originalImagePath))),
+				prependForwardSlash(join(outDir.pathname + config.build.assets, basename(originalImagePath))),
 				serverRoot
```
bluwy, line 114: We can put the `outDir.pathname` here as that doesn't work on Windows. Updating `serverRoot` to the above suggestion should do the trick.

matthewp:
Thanks!

PR #8672, opened 2023-09-26
Support integrations added in updateConfig() in astro:config:setup

delucis:
## Changes

~~Adds an `addIntegration()` method available to integrations in their `astro:config:setup` hook. This adds support for dynamically adding integrations, which is currently not supported (integrations added via `updateConfig` are not picked up):~~

<details>
<summary>Outdated code sample</summary>

```js
import sitemap from '@astrojs/sitemap';

const MyIntegration = (): AstroIntegration => {
  return {
    name: 'my-integration',
    'astro:config:setup': ({ config, addIntegration }) => {
      const userSitemap = config.integrations.find(({ name }) => name === '@astrojs/sitemap');
      if (!userSitemap) {
        addIntegration(sitemap({ /* opts */ }));
      }
    },
  };
};
```

</details>

**Updated after discussion.** This PR now simply fixes the existing `updateConfig()` method to support integrations added to the `integrations` array.

## Testing

- Added a unit test

## Docs

- ~~We’ll need to update the [Integration API reference](https://docs.astro.build/en/reference/integrations-reference/#astroconfigsetup). Will make a PR if this API seems reasonable/after feedback (cc @withastro/maintainers-docs)~~
- No docs needed — now a bug fix!


matthewp:
I like the implementation here. As long as we don't try to deduplicate or anything like that which increases complexity, I think this is a nice addition.

delucis:
> As long as we don't try to deduplicate or anything like that which increases complexity, I think this is a nice addition.

Yeah, in general I don’t think we can know whether duplicates are intentional or not. In theory, some integrations might even make sense as multiple instances, so I think that’s outside of our responsibility.

delucis:
One question: this should probably be a `minor` change right?

bluwy:
Could we reuse the `updateConfig` API for adding new integrations? That way we don't need to introduce a new API. The for-of loop might need to be swapped to a regular for loop against `updatedConfig.integrations` for it to work in that case.

delucis:
> Could we reuse the updateConfig API for adding new integrations? That way we don't need to introduce a new API. The for-of loop might need to be swapped to a regular for loop against updatedConfig.integrations for it to work in that case.

I tried this first, but the way `mergeConfig` works it won’t work, because we avoid mutating anything, so `updatedConfig.integrations` will not be updated with the new value.

delucis:
Oh wait, I think I misread your comment @bluwy — you were actually suggesting avoiding `addIntegration()` altogether in favour of `updateConfig()`, I think? @matthewp wrote:

> probably if we were to do this, it would be a special new api, not just pushing to integrations array.

So that’s why I went this route.

There is already some special casing in `mergeConfig()`, so if we do prefer avoiding the new method in favour of fixing `updateConfig()` for integrations, we could do that there potentially. (Basically make sure that instead of `integrations: [...old.integrations, ...new.integrations]` we do something like `old.integrations.push(...new.integrations)`.)

delucis:
Here’s what it might look like to fix in `updateConfig()` instead: https://github.com/withastro/astro/compare/chris/integrations-update-config-alt

bluwy:
Yeah I mean preferring `updateConfig()` and avoiding `addIntegration()`! I'm thinking the mutating part shouldn't be a problem though as long as we iterate the integrations array with something like this?

```js
for (let i = 0; i < updatedConfig.integrations.length; i++) {
  const integration = updatedConfig.integrations[i]
  // ...
}
```

Since the `updatedConfig.integrations[i]` call should always get the actual new added integration.

delucis:
Ahhh, yes, that does work. Updated [my alternate branch](https://github.com/withastro/astro/compare/chris/integrations-update-config-alt) with that approach. I’m happy to go with that if @matthewp agrees.

matthewp:
I'd be fine with that.

delucis:
OK, updated this PR to use the `for` loop approach. Thanks to Bjorn for suggesting that!

I had to `eslint-disable` the for loop as there’s a rule set up to enforce use of `for..of`, and I included a comment explaining why we need it.

Added a minor changeset as that seems more honest about what this does even though it could also be considered a patch/bug fix.

.changeset/cold-schools-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Support integrations added by other integrations with `updateConfig()`
```
matthewp, line 5: Can you write a bit more, like a paragraph (this would go into the blog post).

bluwy (approved):
Looks great! I know Nate mentioned filtering falsy integrations and possibly needing to flatten it. If needed we can implement that while [merging the config here](https://github.com/withastro/astro/blob/37c4a0bd035661ec93dca5a4097b110fdae8c82e/packages/astro/src/core/config/merge.ts#L23-L39), but I don't think it's necessarily needed for now and we can keep watch for this in the future.

delucis:
Updated with a more detailed changeset!

.changeset/cold-schools-yell.md
```diff
@@ -0,0 +1,34 @@
+---
+'astro': minor
+---
+
+Add integrations dynamically
```
sarah11918, line 5: I'm torn, because while I think this is a good blog header, I think in our changesets starting with "add", typically we'd be saying what we added, which is the ability to add integrations dynamically. :sweat_smile: 

> Add the ability to add integrations dynamically

.changeset/cold-schools-yell.md
```diff
@@ -0,0 +1,34 @@
...
+
+Astro integrations can now dynamically add additional integrations during set-up. This makes it possible for integration authors to bundle integrations more intelligently for their users.
```
sarah11918, line 7:
```suggestion
Astro integrations can now themselves dynamically add and configure additional integrations during set-up. This makes it possible for integration authors to bundle integrations more intelligently for their users.
``` 
I just kind of like this for emphasis, since we have some level of recursion going on?

And just checking that if you only got one phrase to describe this feature, it would be "intelligent bundling" that's the key idea here, and not for example something about ensuring dependencies because now it can know what other integrations you have?

.changeset/cold-schools-yell.md
```diff
@@ -0,0 +1,34 @@
...
+
+For example, an integration can decide to add Astro’s official sitemap integration if it sees a user hasn’t already configured it:
```
sarah11918, line 9:
```suggestion
In the following example, an Astro integration's own configuration includes a check for Astro’s official sitemap integration. If not found, the integration will add `@astrojs/sitemap` to the existing project, passing any desired configuration options:
```
Was trying to see whether we could use our newer suggestion format here...

sarah11918 (commented):
Looks great @delucis ! I did try to rework the line introducing the code snippet, not entirely sure how successfully, in keeping with the new style we're going for. At any rate, it feels a little less anthropomorphic (trying not to use "sees" for example).

.changeset/cold-schools-yell.md
delucis: Ah yeah, definitely was thinking of the blog format here and you’re right it’s a strange fit for the changeset.

This might work? (Also avoids the “add … add”.)

```suggestion
Support adding integrations dynamically
```

.changeset/cold-schools-yell.md
delucis: Bit biased because I’m thinking about Starlight being able to switch from always adding MDX & Sitemap to only adding them when needed — this is what I had in mind for “intelligent bundling”. I think that makes sense seeing as technically integrations could provide their dependencies (like Starlight does today), but not “intelligently” — e.g. if a user adds sitemap themselves, things crash.

.changeset/cold-schools-yell.md
sarah11918: I'll let Mattthew be the final judge here on what's helpful for him! But sure, "support adding", "support for adding"... those work!

.changeset/cold-schools-yell.md
delucis: Tried to rework a bit to de-clunkify it?
```suggestion
In the following example, a custom integration checks if `@astrojs/sitemap` is already configured. If not, the integration adds Astro’s sitemap integration, passing any desired configuration options:
```

sarah11918: "checks whether" and yes, I think it's good!

PR #8673, opened 2023-09-26
Fix normalize inconsistency between load & transform for filename

hinshun:
- Some build systems symlink node_modules into the root, if normalized against config.root then the cache key between set & get will be different.

## Changes

- During `transform`: `cacheFullCompilation` normalizes filename via `normalizePath(parsedId.filename)`
- During `load`: `getCachedCompileResult` normalizes filename via `normalizePath(normalizeFilename(parsedId.filename, config.root))`
- This causes a cache miss when using a package manager (Nix in my case) that symlinks `node_modules` into project root
- Due to cache miss, `compileResult == null`, so the `code` portion doesn't become `/* client hoisted script, empty in SSR: ${id} */`
- Then, the entire `.astro` file gets sent to esbuild transform, which doesn't work because astro files aren't valid typescript.

Alternatively, we can go the other way and ensure during `transform` we also normalize via `normalizePath(normalizeFilename(parsedId.filename, config.root))`.

## Testing

Starlight doesn't work without this change. After this change applied, it was fixed. I also patched the code to `console.log` the `parserId.filename` before and after normalization to confirm that it indeed changed the cache key in a bad way.

## Docs

This affects the out-of-the-box experience with package managers that symlink `node_modules`.

### DIFF
--- a/packages/astro/src/vite-plugin-astro/index.ts
+++ b/packages/astro/src/vite-plugin-astro/index.ts
@@ -11,7 +11,6 @@ import {
 	type CompileProps,
 } from '../core/compile/index.js';
 import { isRelativePath } from '../core/path.js';
-import { normalizeFilename } from '../vite-plugin-utils/index.js';
 import { cachedFullCompilation } from './compile.js';
 import { handleHotUpdate } from './hmr.js';
 import { parseAstroRequest } from './query.js';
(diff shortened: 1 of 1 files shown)


natemoo-re:
Based on the failing tests, I'm thinking that this change doesn't work as expected. 

I will say that changing how path normalization works has bitten us on multiple occasions. There are almost always unexpected side-effects that break other features when making a "tiny" change like this. 

If you could provide a reproduction that shows us how the current solution is broken for you and ensure that all tests are passing, that would go a long way towards getting this fixed!

ematipico:
@hinshun are you still interested at this PR?

hinshun:
Yes, sorry been dealing with fires at work. I will provide a reproduction and investigate the CI failures.

florian-lefebvre:
@hinshun can you provide a reproduction for the issue you mentioned in the initial comment?

florian-lefebvre:
I'm closing this PR for inactivity and because according to https://github.com/withastro/astro/pull/8673#issuecomment-1758104944, this is probably not the right fix for your problem. Please open an issue with a reproduction to help us help you!

Issue #8178, opened 2023-08-21
Hydration Mismatch Due to Incomplete Content Download in Slow Connection 


howagain:
### What version of `astro` are you using?

2.9.0

### Are you using an SSR adapter? If so, which one?

NodeJS SSR adapter (in Docker container)

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome, Safari

### Describe the Bug

The page fails hydrating throwing an error of `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'nextSibling')` also visible with `firstChild` or `firstSibling` depending on where the hydration script out ran the HTML.

The issue occurs when hydration is executed before the content of a large astro island is fully downloaded. This leads to a hydration mismatch as the closing  tag is missing. It's particularly prevalent with slow network connections, and our setup where JS is fully cached but HTML is not (with a cache-control: no-cache header). It appears to be related to this race condition issue: https://github.com/withastro/astro/pull/7197 but specific to the way Solid works. The problem is not a regression and is found in Astro v2.9.0.

In our example, when hydration breaks, elements that were not hydrated do not `console.log` on click.

Picture of the error in the console:
<img width="1792" alt="Screenshot 2023-08-22 at 8 31 58 AM" src="https://github.com/withastro/astro/assets/59291747/5c675bf4-2af1-4839-ac0f-82ebdacfbc15">

### What's the expected result?

The hydration should successfully match the opening  and closing  tags, without any mismatches or breaks, even in slow network conditions with large astro islands.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qsvzyu?file=README.md

#### Instructions
1. Run `npm run build` and `npm run preview` to get the built server
2. Load the page, go to dev tools, turn on cache by unchecking "disable cache", turn on Fast 3G throttling for network speed
3. Refresh the page several times, occasionally it will throw the `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'nextSibling')` error

### Participation

- [ ] I am willing to submit a pull request for this issue.

jonyo:
@natemoo-re In the mean time, can you think of any temporary work-arounds to somehow delay hydration until the full DOM of the island is streamed in? We tried a few things, no success so far:

- split into smaller islands: worked a little, but we have a state where the complexity grows (and thus, more markup to send) in a way that is not possible to split up more than we have already, and in that state it still consistently gets the error.
[...]

PR #8680, opened 2023-09-27
Fix hydration on slow connection

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/8178

This implements a `<!--astro:end-->` comment marker added as the last child of the `astro-island` tag, used to detect if the children is ready for hydration.

As the comment marker could risk being stripped by HTML minifiers, I also added a fallback `DOMContentLoaded` listener to hydrate if it's triggered.

The downside with this approach is the increased code size for the astro-island component: 3053 chars -> 3362 chars.

## Testing

Tested manually with the repro in #8178. It's a bit hard to create a test for this without making it flaky.

## Docs

n/a. We currently don't document at the very specific time when we start hydrating.


bluwy:
Actually I need to benchmark multiple islands too to confirm it doesn’t stall them altogether. Marking as draft.

matthewp (approved):

bluwy:
I did some tests, it actually stalled all islands before rendering, which isn't a great experience for those on poor connections who have to wait for the entire page to load before they can interact with it. I switched to something similar to Nate's idea instead, adding a trailing child script to emit an event to signal the parent island that it's loaded. Wish there's something like `DOMContentLoaded` but for partial dom nodes instead.

bluwy:
This is ready for review. I updated the PR description of what this PR does now.

matthewp (approved):
Pretty clever solution!

PR #8682, opened 2023-09-27
fix(cloudflare): target es2022 instead of es2020 to fix esbuild incompatibility issues

dario-piotrowicz:
## Changes

The Cloudflare adapter currently uses esbuild targeting es2020, this works fine but this can be problematic when used with some astro integrations, like `@astrojs/mdx`, the issue being that mdx seems to use top-level awaits not supported by es2020

So if I have an astro config such as this:
```ts
export default defineConfig({
  site: 'https://example.com',
  integrations: [mdx(), sitemap()],
  output: "server",
  adapter: cloudflare(),
});
```

and try to build my application I get the following:
Screenshot 2023-09-27 at 19 48 40

> **Note**
> This happens with the blog template, if I start a brand new application, as soon as I add the Cloudflare adapter and try building my application I get this error

> **Note**
> This might also break in other cases as well, this with the mdx integration is the only one I noticed thus far

## Testing

I'm not sure if such a small change requires tests, please let me know if I should look into adding those

## Docs

There isn't really any documentation to add as this is a simple change that should not be visible to users.


alexanderniebuhr (approved):
@dario-piotrowicz Fantastic work! 🚀 

I've verified that CF Workers & Pages support es2022. Given that all existing tests have passed, there's no need for additional tests. Documentation updates aren't necessary either.

dario-piotrowicz:
Thanks a lot @alexanderniebuhr ❤️, and thanks for cleaning up the PR and adding the one spot I missed 🙈

alexanderniebuhr:
@dario-piotrowicz Checking in on this. I'd like to confirm if the documentation team has reviewed the changeset. Once that's done, this will be ready for merge, and I'll expedite the process.

sarah11918 (approved):
Docs approves the changeset! :partying_face:

Issue #6217, opened 2023-02-10
Percent in file name causes error in Content Collection


braden-w:
### What version of `astro` are you using?

2.0.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

Including a percent in a file name in the `Content` folder results in the following error:

```
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at https://aonipxqowgithub-1pqc.w-credentialless.staticblitz.com/blitz.ce29b75f029c88a5dcc9b3dccb923c5bcb86e937.js:6:144075
    at Module.fileURLToPath (https://aonipxqowgithub-1pqc.w-credentialless.staticblitz.com/blitz.ce29b75f029c88a5dcc9b3dccb923c5bcb86e937.js:6:144080)
    at isOnIgnoreList (file://file:///home/projects/aonipxqow.github/node_modules/astro/dist/content/utils.js:149:53)
    at Module.getEntryType (file://file:///home/projects/aonipxqow.github/node_modules/astro/dist/content/utils.js:138:39)
    at handleEvent (file://file:///home/projects/aonipxqow.github/node_modules/astro/dist/content/types-generator.js:94:32)
    at Object.queueEvent (file://file:///home/projects/aonipxqow.github/node_modules/astro/dist/content/types-generator.js:181:17)
    at FSWatcher.eval (file://file:///home/projects/aonipxqow.github/node_modules/astro/dist/content/server-listeners.js:59:24)
    at EventEmitter.emit (https://aonipxqowgithub-1pqc.w-credentialless.staticblitz.com/blitz.ce29b75f029c88a5dcc9b3dccb923c5bcb86e937.js:6:155690)
    at FSWatcher.emitWithAll (file://file:///home/projects/aonipxqow.github/node_modules/vite/dist/node/chunks/dep-3007b26d.js:50373:8)
```

While the error seems initially understandable, the Astro generated slug **already** strips non-alphanumeric characters, so the percent is not included in the end URL. It would be useful to be able to include percents, as my current workflow has several files that include % signs in their titles.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/withastro/astro/tree/latest/examples/blog?file=src/content/blog/problem%.md

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Can confirm that this is still a problem, even on the latest!

PR #8684, opened 2023-09-28
Support content collections with % in filename

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/6217
- This was more than I expected! In addition to not using pathToFileURL in some cases where we should have, we also need to make sure that the bundles we write to disk don't have the % character (unless it is part of an encoding), because `import()` will throw on those files.
- Had to borrow some code from Node core which handles something similar: https://github.com/nodejs/node/blob/3838b579e44bf0c2db43171c3ce0da51eb6b05d5/lib/internal/url.js#L1382-L1391

## Testing

- Test case added to content collections 

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/core/build/static-build.ts
+++ b/packages/astro/src/core/build/static-build.ts
@@ -4,7 +4,7 @@ import glob from 'fast-glob';
 import { bgGreen, bgMagenta, black, dim } from 'kleur/colors';
 import fs from 'node:fs';
 import path, { extname } from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import * as vite from 'vite';
 import type { RouteData } from '../../@types/astro.js';
 import {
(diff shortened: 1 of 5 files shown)


packages/astro/src/core/build/static-build.ts
```diff
@@ -186,6 +186,18 @@ async function ssrBuild(
...
+						// We do this because you cannot import a module with this character in it.
+						for(let i = 0; i < name.length; i++) {
+							if(name[i] === '%') {
+								const third = name.codePointAt(i + 2)! | 0x20;
+								if (name[i + 1] !== '2' || third !== 102) {
+									return `chunks/${name.replace(/%/g, '_percent_')}_[hash].mjs`;
+								}
+							}
+						}
+						//console.log("HERE", name)
```
alexanderniebuhr, line 200:
```suggestion
```

nit: should we remove the comment?

alexanderniebuhr (approved):
based on my trust in Node core code _(have to fully tried it locally)_, this lgtm.

packages/astro/src/core/build/static-build.ts
```diff
@@ -186,6 +186,17 @@ async function ssrBuild(
...
+						// This is borrowed from Node core: https://github.com/nodejs/node/blob/3838b579e44bf0c2db43171c3ce0da51eb6b05d5/lib/internal/url.js#L1382-L1391
+						// We do this because you cannot import a module with this character in it.
+						for(let i = 0; i < name.length; i++) {
+							if(name[i] === '%') {
+								const third = name.codePointAt(i + 2)! | 0x20;
+								if (name[i + 1] !== '2' || third !== 102) {
+									return `chunks/${name.replace(/%/g, '_percent_')}_[hash].mjs`;
+								}
+							}
+						}
```
bluwy, line 199: I feel like it should be possible to import modules with `%` but it needs to be escaped before importing. Is it happening inside the Rollup-generated code, or are we importing it wrongly?

matthewp: It's the rollup generated code, it has this exactly like `import('../chunks/four%.hash.mjs')` and Node cannot handle that. I think this might actually be a Node.js bug; not being able to import valid filename seems like a bug to me. But any event, if we escape it in the filename then everything works.

matthewp: We talked offline and this is either a Node.js bug or a Rollup bug, so going with the workaround for now. I'll be filing a bug with Node.js to see what they think.

Issue #8676, opened 2023-09-27
Astro v3 markdown page router Chinese get unreadable characters


liruifengv:
### Astro Info

```block
Windows 10

v3.0.0 - v3.1.4 are all have this bug

v2.10.15 was fine
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

## Bug description

In Astro v3, without using content collection and layout, only create markdown files under pages to use page routing, when `astro dev`, Chinese characters are unreadable. After `astro build`, the preview is correct.

## Steps to reproduce

- Create `test.md` file under pages, enter Chinese content
- Run `astro dev`
- View the page on `http://localhost:4321/test`, Chinese characters are unreadable

![image](https://github.com/withastro/astro/assets/25167721/37b9e08f-1bc0-48b2-b1c9-2d1fb74d990d)

### What's the expected result?

Chinese characters are readable

### Link to Minimal Reproducible Example

https://github.com/liruifengv/test-md-chinese

### Participation

- [x] I am willing to submit a pull request for this issue.

Princesseuh:
Your page / server needs to specify a charset to use. Typically, in HTML this is done using `<meta charset="utf-8" />`. Astro serves the page as-is.

liruifengv:
> Your page / server needs to specify a charset to use. Typically, in HTML this is done using `<meta charset="utf-8" />`. Astro serves the page as-is.

How should I specify the charset when I only have a test.md file as a page?
Astro v2.10.15 is working fine without doing anything.

matthewp:
@liruifengv You'll want to use a layout, see here: https://docs.astro.build/en/core-concepts/layouts/#markdownmdx-layouts

liruifengv:
I know the layout.

But in the docs tutorials, Astro beginners learn Astro step-by-step when learning [create-your-first-md-file](https://docs.astro.build/en/tutorial/2-pages/2/#%E5%88%9B%E5%BB%BA%E4%BD%A0%E7%9A%84%E7%AC%AC%E4%B8%80%E4%B8%AA-md-%E6%96%87%E4%BB%B6), they will encounter this problem, which will confuse the beginner until he learns [the layout chapter](https://docs.astro.build/zh-cn/tutorial/4-layouts/)

This isn't a problem I'm having, but it's a beginner's feedback to me.

Why does v2 work and v3 doesn't anymore, is something done to it to make it unsupported?

I think this needs to be fixed, not CLOSED.
[...]

PR #8686, opened 2023-09-28
fix: accidentally breakage of charset in response header

liruifengv:
## Changes

With [This][1] plugin change to use `renderComponent`, causing [this][2] branch to be false and we lost the `Content-Type: text/html; charset=utf-8` header below, and I think it's an accident change, right?

[1]: https://github.com/withastro/astro/blob/main/packages/astro/src/vite-plugin-markdown/index.ts#L180
[2]: https://github.com/withastro/astro/blob/f5c617e3a3ed8f010ff28f0cfe0f322ad54ed6e0/packages/astro/src/runtime/server/render/page.ts#L17C9-L17C9

### Before
image

### After
image

Fixed: https://github.com/withastro/astro/issues/8676
## Testing
None

## Docs
None

### DIFF
--- a/.changeset/nine-books-clap.md
+++ b/.changeset/nine-books-clap.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix: accidentally breakage of charset in response header
--- a/packages/astro/src/runtime/server/render/page.ts
+++ b/packages/astro/src/runtime/server/render/page.ts
@@ -58,6 +58,11 @@ export async function renderPage(
 	// Create final response from body
 	const init = result.response;
 	const headers = new Headers(init.headers);
+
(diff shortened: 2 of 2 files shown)


matthewp:
When intentionally do not set the charset because Astro is often used for SSG and many hosts do not set this header. You can fix this in your own app by adding:

```html
<meta charset="utf-8">
```

Into the `<head>` element of your pages. If you think we should change this behavior the place to discuss that is here: https://github.com/withastro/roadmap/discussions

I would note, we need strong evidence that hosts have changed here and that they do set the charset as a header. Otherwise we would be breaking production. As frustrating as this is for when you are starting out in dev, it would be much worse for dev to look fine but your deployed app to look wrong.

PR #8689, opened 2023-09-28
Update codesample for Alpine README

ElianCodes:
## Changes

Small update to change our Expressive Code to use `diff`'s instead of languages


sarah11918 (approved):
LGTM!

PR #8692, opened 2023-09-28
[@astrojs/preftech] Prefetch images as well

tex0l:
## Changes

Prefetch images references in `img` tags at the same time the stylesheets are prefetched

## Testing

This was tested manually, if it is approved in principle, I'll write a test suite.

## Docs

I did not expose it as an option, therefore the documentation should just state that it pre-fetches the images. If it is approved in principle, I'll write the necessary documentation.


ematipico:
Thank you, @tex0l , but we are moving `prefetch` into core so this code won't be relevant anymore.

PR #8693, opened 2023-09-28
chore(ci): pin CI Node version to 18.17.1

lilnasy:
## Changes
- CI - Node 18.18 breaks ~~playwright~~ (not just playwright) on windows.

## Testing
- Does not affect behavior.

## Docs
- Does not affect usage.


lilnasy:
It was neccessary for the "test" job as well, but that didn't fail the checks because branches in the `withastro` org use cache, and the "test" was skipped.

Issue #8637, opened 2023-09-22
Vite injected styles and Astro injected styles are conflicting


marceloverdijk:
### Astro Info

```block
Astro                    v3.1.2
Node                     v18.17.1
System                   macOS (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Yes I also noticed (partially) duplicated css.

E.g. I have a span like:

```
<span class="badge bg-primary bg-opacity-10 text-primary">Hotel</span>
```

and the generated css contains:

L28048:
```
.bg-primary {
  --bs-bg-opacity: 1;
  background-color: rgba(var(--bs-primary-rgb), var(--bs-bg-opacity)) !important;
}
```

L28116:
```
.bg-opacity-10 {
  --bs-bg-opacity: 0.1;
}
```

L34947:
```
.bg-primary {
  --bs-bg-opacity: 1;
  background-color: rgba(var(--bs-primary-rgb), var(--bs-bg-opacity)) !important;
}
```

The `.bg-primary` is for some reason duplicated, but the `bg-opacity-10` not.
The problem now is that the `--bs-bg-opacity` is overwritten again, causing a styling issue.

This new issue is created after discussion [here](https://github.com/withastro/astro/issues/7508) with @bluwy 

### What's the expected result?

Correct - non duplicate - styling.

### Link to Minimal Reproducible Example

https://github.com/marceloverdijk/astro-7508

### Participation

- [ ] I am willing to submit a pull request for this issue.

sarvex:
This is the GitHub repo https://github.com/sarvex/afraid-aurora containing the code 

The problem is seen with the current NodeJS version 20.7. with LTS 18.18, the problem is not there

marceloverdijk:
Thx @sarvex for sharing you repo. 
Note that I used Node v18.17.1 and with that version I had the issue as well.

sarvex:
@marceloverdijk there is something weird going on... 
The problem shows up on the local windows machine but is absent under WSL Ubuntu. 
The problem does not show up on Stackblitz

marceloverdijk:
On my side as well. This particular example mentioned above happens for me in 1 project but in another project.
So it is very unclear where it is coming from to be honest.

@bluwy also noticed this behavior (see discussion in #7508 )

marceloverdijk:
I also found this issue: https://github.com/withastro/astro/issues/6975
[...]

PR #4157, opened 2022-08-04
Fix duplicate CSS in `dev` mode

natemoo-re:
## Changes

- Closes #4102
- During `dev`, Vite treats all `.css` files as JS modules that inject themselves into the page.
- This is great for HMR, but it leads to flashes of unstyled content during `astro dev`.
- To avoid this, Astro started injecting the static CSS while server-side rendering the file. This avoids any flashes of unstyled content, but when the CSS is injected for HMR, it leads to duplicate styles. 
- This PR adds a tiny MutationObserver to our `client/hmr` logic that removes any SSR'd styles as soon as the HMR styles are present on the client. No more duplicates!
- Also removed the `data-astro-injected` attribute because it's no longer needed!

## Testing

Tests using `data-astro-injected` updated. HMR already covered by existing tests.

## Docs

N/A

PR #8706, opened 2023-09-29
Fix duplicated Astro and Vite injected styles

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/8637

The issue was that `<style data-astro-dev-id="..."` and `<style data-vite-dev-id="..."` have different IDs that they don't match and deduplicate correctly, causing duplicated styles. The main issue is that we're using `importedModule.url` instead of `importedModule.id`.

But I made a complete cleanup of CSS HMR, relying on Vite to handle it entirely. A lot of HMR handling was added by https://github.com/withastro/astro/pull/4157 before. It's possible to take this new approach today because Vite natively treats `data-vite-dev-id` as FOUC-prevention injected styles. 

## Testing

Existing CSS and CSS HMR tests should still work. Updated them where necessary for the slightly different markup. I also tested the https://github.com/withastro/astro/issues/8637 repro manually.

## Docs

n/a. internal change and bug fix.


packages/astro/e2e/css-sourcemaps.test.js
bluwy, line 1: This test file seems to duplicate the same test in `css.test.js`

packages/astro/e2e/view-transitions.test.js
```diff
@@ -631,7 +631,7 @@ test.describe('View Transitions', () => {
 	});
 
 	test('client:only styles are retained on transition', async ({ page, astro }) => {
-		const totalExpectedStyles = 8;
+		const totalExpectedStyles = 7;
```
bluwy, line 634: There's no more `<style data-astro-dev-id` as it's merged into `<style data-vite-dev-id`, so this number is reduced by one.

packages/astro/src/runtime/client/hmr.ts
```diff
@@ -1,46 +1,7 @@
...
-				if (isViteInjectedStyle(node)) {
-					injectedStyles.get(node.getAttribute('data-vite-dev-id')!)?.remove();
-				}
-			}
-		}
-	});
-	mo.observe(document.documentElement, { subtree: true, childList: true });
-
-	// Vue `link` styles need to be manually refreshed in Firefox
-	import.meta.hot.on('vite:beforeUpdate', async (payload) => {
```
bluwy, line 20: I manually tested this and it seems to not apply anymore, so I removed it. The rest of this file relates to our CSS handling, which is no longer needed as Vite handles all now, so pretty much the entire file is empty and not need to be loaded anymore.

matthewp (approved):
Amazing, thanks!

PR #8709, opened 2023-09-29
fix(cloudflare): broken link in docs

alexanderniebuhr:

### DIFF
--- a/packages/integrations/cloudflare/README.md
+++ b/packages/integrations/cloudflare/README.md
@@ -192,7 +192,7 @@ default `"off"`
 
 Determines whether and how the Cloudflare Runtime is added to `astro dev`.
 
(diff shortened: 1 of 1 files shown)


Yan-Thomas:
Thanks @alexanderniebuhr, nice catch! Since we're already on the "fix links" topic, our CI also found out we have a bad link description in regards to a11y, could you add the following change to line as well 172?:

```md
Add `wasmModuleImports: true` to `astro.config.mjs` to enable this functionality in both the Cloudflare build and the Astro dev server. Read more about [using Wasm modules](#use-wasm-modules)
```

With both done our docs integrations CI should be perfectly unblocked 🙌

sarah11918 (approved):
This checks out to me!

Issue #8669, opened 2023-09-26
ViewTransitions: Call to sharp().toBuffer() in nested component breaks view transitions


rkulinsky:
### Astro Info

```block
Astro                    v3.1.4
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

I am using sharp library to generate previews for the images. Preview is generated by calling to sharp(...).toBuffer() and then used as an inline style with base64 encoded image for "background-image: url(...)".
The issue is that view transitions work for the first image only.

Minimal reproducible example is provided.

### What's the expected result?

View transitions work for all images with transition:name.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-rbkdv6?file=src%2Flib.ts%3AL59,src%2Fpages%2Findex.astro%3AL10,src%2Fcomponents%2FImage.astro%3AL23

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #8710, opened 2023-09-29
Fixes View transition styles being missing when component used multip…

matthewp:
…le times

## Changes

- Fixes https://github.com/withastro/astro/issues/8669
- Using a Map was wrong, we want to wait on each *instance* of a component. If there are multiple wait on them all.  
- Using a Set fixes it

## Testing

- Test added that checks for the right number of styles.

## Docs

N/A, bug fix.

### DIFF
--- a/.changeset/wet-numbers-serve.md
+++ b/.changeset/wet-numbers-serve.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/markdoc': patch
+'astro': patch
+---
+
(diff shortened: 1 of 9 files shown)


martrapp:
At first I was surprised that only the initialization of the data structure and the collection of the data is affected by the change. It doesn't seem to have any effect on the use of the structure by the consumers. I guess we were only interested in `values()` for `Map` like `Set`.

LGTM

martrapp (approved):

PR #8715, opened 2023-10-01
Fixes: Shiki syntax highlighting adds is:raw attribute to the HTML output

cprass:
I recently ran my page through the W3C validator and it showed me a bunch of errors and warnings about `is:raw` being not allowed. Afaik `is:raw` is an Astro specific directive that should now appear on the static page output.

The raw-attribute appears only in the Shiki highlighted syntax.

To reproduce it, you can use this template https://stackblitz.com/github/withastro/astro/tree/latest/examples/with-markdown-shiki

## Changes

I removed the part of the `shiki` plugin that adds the `is:raw` to the `pre`-tags. The fact that `is:raw` is in the HTML output implies that `is:raw` or a similar mechanic is already used higher up the chain.

## Testing

I added a Mocha test to verify the fix.

## Docs

The attribute is hidden from the user and only visible when inspecting the page HTML. Browsers seem to ignore it and it only appeared as a "problem" when running a HTML validator on the page.

### DIFF
--- a/.changeset/cuddly-vans-reply.md
+++ b/.changeset/cuddly-vans-reply.md
@@ -2,4 +2,4 @@
 '@astrojs/markdown-remark': patch
 ---
 
-Remove is:raw from remark Shiki plugin
+Remove `is:raw` from remark Shiki plugin
--- a/.changeset/cuddly-vans-reply.md
+++ b/.changeset/cuddly-vans-reply.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
+Remove is:raw from remark Shiki plugin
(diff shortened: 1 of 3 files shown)


.changeset/cuddly-vans-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
+Remove is:raw from remark Shiki plugin
```
ematipico, line 5:
```suggestion
Remove `is:raw` from remark Shiki plugin
```

ematipico (approved):
It looks good to me, but I would prefer to defer the merge and review to @bluwy

bluwy (approved):
LGTM. Seems like an artifact from the Astro-flavoured markdown days.

PR #8725, opened 2023-10-02
fix(node app): only include undici in output when on stackblitz

lilnasy:
## Changes
- Allows polyfill to be removed with tree-shaking.
- Results from the blog template with `@astrojs/vercel`:

|            | Before  | After |
| ------------- | ------------- | ------------- |
Single Function| 1,312 kB  | 535 kB |
functionPerRoute | 1200~1300 kB  | 430~550 kB  |

## Testing
Does not attempt to change behavior; existing tests should pass.

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/thirty-ravens-fly.md
+++ b/.changeset/thirty-ravens-fly.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Node-based adapters now create less server-side javascript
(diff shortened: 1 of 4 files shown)


.changeset/thirty-ravens-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Node-based adapters now create less server-side javascript
```
lilnasy, line 5: Changeset is a little awkward.

- in astro, but benefits node and vercel
- borderlines patch and minor

lilnasy:
Closing as stackblitz now runs on node 18 and the undici polyfill can be removed altogether.

Issue #8674, opened 2023-09-26
Page styles not being removed when view transitions enabled in >= 3.0.13


adrianosmond:
### Astro Info

```block
Astro                    v3.1.4
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

Using the minimal reproducible example: https://stackblitz.com/edit/github-xrfje7

Steps to reproduce
- Open the index page, notice that the background is black
- Click the link to go to the white page
- Notice that the background is now white (this is expected)
- Click the link on the white page to go back to the first page with the black background
- Notice that the background is still white (this is not expected)

The issue is that the page which should have the black background now has the white background of the page it just visited. It seems like styles from the page with the white background are not being removed when leaving that page.

This only happens if you include ViewTransitions in the page Layout. 

It seems to have been introduced in v3.0.13 - v3.0.12 doesn't have this issue as far as I can tell.

### What's the expected result?

Styles that are specific to a single page shouldn't hang around after leaving that page

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xrfje7

### Participation

- [ ] I am willing to submit a pull request for this issue.

resource11:
I have this same issue @matthewp. I can send you a repo for testing if you need.

matthewp:
Thanks, taking a look now.

matthewp:
Probably caused by this: https://github.com/withastro/astro/pull/8472

Issue #8604, opened 2023-09-19
Commit `fa77fa63d` in Astro 3.0.13 breaks page after a view transitions with dev server


dallyh:
### Astro Info

```block
Astro                    v3.1.1
Node                     v18.17.1
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/react
                         astro-i18n-integration
                         @astrojs/sitemap
                         @astrojs/prefetch
```

### Describe the Bug

I've just updated packages for my project, which I currently mainly use to experiement with Astro 3. After upgrading to 3.1.1 I've noticed that navigation while using ViewTransitions in SPA mode with a **dev** server is currently broken - navigation to a page does not load styles properly, or the page is shifted down (like it was scrolled).

I have a repository here: https://github.com/dallyh/website-concept/. To reproduce the issue, just clone the `before-upgrade` and `broken-nav-transition` repos. Install packages for branch `before-upgrade`, and run the dev server. Switch to English language (in the footer, or navigate to `http://localhost:4321/website-concept/en/` - this is not mandatory for the bug to occur), then click the "Contact" or "Resume" card - everything works as expected. 

Do the same for branch `broken-nav-transition`, which uses newest Astro, and suddenly both pages are rendered incorrectly. If you then refresh the page manually, everything is fine.

For the page "Resume" the whole page is moved down and the scrollbar is at the top most position.

![image](https://github.com/withastro/astro/assets/6968534/8f98aa77-4bdd-486e-baac-26d251a2acb9)

For the page "Contact" some of the styles are not loaded (or it seems like it).

Broken styling:

![image](https://github.com/withastro/astro/assets/6968534/def7031c-d1b4-4299-b9d9-ecde556d9d81)

Expected result:
![image](https://github.com/withastro/astro/assets/6968534/d864b700-759b-41a4-aee7-2af95e8b2af0)

I cannot provide a different example, as I don't really know what triggers this. I have tracked down the problematic commit, which is: https://github.com/withastro/astro/commit/fa77fa63d944f709a37f08be93f0d14fe1d91188. After a revert, everything works smoothly.

I should also mention, that this only happens on a dev server, after the page gets build, everything is fine. I've noticed that the commit only changes things while in dev mode, so that is maybe why.

### What's the expected result?

Styles work, page starts from top.

### Link to Minimal Reproducible Example

https://github.com/dallyh/website-concept/tree/broken-nav-transition

### Participation

[...]

PR #8727, opened 2023-10-02
Fix view transitions with client:only components

martrapp:
## Changes

The original issue is #8114. Persistent `client:only` components lose their styling on transition.
This only happens with `astro dev`, not with `astro build`.
The reason was that the HTML files contained only statically generated styles, and the comparison between the old DOM and the new file removed the client-side generated styles.
Fix #8472 fixed this by attempting to identify these client-generated styles and retain them.
This caused several issues where old styles were not deleted during the transition, even though they should have been. (#8604, #8674)

## Testing

Manually tested with code from #8604.

## Docs

n/a, bug fix

### DIFF
--- a/.changeset/cuddly-vans-reply.md
+++ b/.changeset/cuddly-vans-reply.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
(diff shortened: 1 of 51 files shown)


martrapp:
Replaced by #8745

PR #8729, opened 2023-10-02
chore: remove undici polyfill

lilnasy:
## Changes
Removes undici polyfill

## Testing
Does not change behavior (hopefully.)

## Docs
Does not affect usage.


Princesseuh:
https://github.com/withastro/astro/blob/main/packages/astro/astro.js also requires changes

ematipico (approved):

.changeset/thirty-ravens-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
bluwy, line 2: I think we also need to add this for `@astrojs/node` and `@astrojs/telemetry` so they all have new versions published.

lilnasy: `@astrojs/telemetry` is a good catch, node and vercel shouldn't need a bump though. I think it would apply retroactively to them.

bluwy (approved):

PR #8730, opened 2023-10-02
Improve `astro info` compatability

natemoo-re:
## Changes

- On Windows terminals, `astro info`'s copy to clipboard feature was copying the entire JSON-encoded string rather than the unencoded string as expected.
- With some help from @lilnasy, we debugged this and landed on passing the input directly to `clip`.

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/packages/astro/src/cli/info/index.ts
+++ b/packages/astro/src/cli/info/index.ts
@@ -56,7 +56,8 @@ async function copyToClipboard(text: string) {
 	if (!shouldCopy) return;
 	const command = system === 'darwin' ? 'pbcopy' : 'clip';
 	try {
-		execSync(`echo ${JSON.stringify(text.trim())} | ${command}`, {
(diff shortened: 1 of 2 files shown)


packages/astro/src/cli/info/index.ts
```diff
@@ -56,7 +56,8 @@ async function copyToClipboard(text: string) {
 	if (!shouldCopy) return;
 	const command = system === 'darwin' ? 'pbcopy' : 'clip';
 	try {
-		execSync(`echo ${JSON.stringify(text.trim())} | ${command}`, {
+		execSync(command, {
+			input: text.trim(),
 			encoding: 'utf8',
 			stdio: 'ignore',
```
lilnasy, line 62:
```suggestion
```
It doesn't work unless this line is removed as well.

Issue #8470, opened 2023-09-08
Missing styles and scripts in dev mode on Windows when the drive letter is lowercase


erg208:
### Astro Info

```block
Astro                    v3.0.12
Node                     v18.17.1
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/vue
                         @astrojs/react
```

### If this issue only occurs in one browser, which browser is a problem?

Edge and Chrome

### Describe the Bug

Followed ["using script in astro documentation"](https://docs.astro.build/en/guides/client-side-scripts/#using-script-in-astro) and it does not work in our Windows environment. The JavaScript never gets added to the Web page. It does work on [https://stackblitz.com/edit/github-1egmuk](https://stackblitz.com/edit/github-1egmuk).

### What's the expected result?

Expect to see "Hello" on the Web page; and "test.js is being loaded..." in de the dev-tools console of the Web browser.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-1egmuk

### Participation

- [ ] I am willing to submit a pull request for this issue.

erg208:
I am still trying to debug the Windows 10 environment on why this feature of Astro is not working. What we really what to do is utilize external TypeScript modules with Classes and interfaces, but we thought we would start with with JavaScript.

What I have discovered is that removing all of the global npm modules did not resolve the issue.

I found that our workaround for the moment is to utilize the Windows Linux Subsystem, because we are not seeing the issue in this OS.

matthewp:
Hey @erg208 my guess is that this is a path issue. Is it possible that this *does* work?

```
<script>
import "../scripts/test.js";
</script>
```

erg208:
Hi @matthewp , thank you for the suggestion. I did try with the import and it still did not work in windows. I tried as admin in windows thinking maybe it was the cause, but no luck. I also tried changing to use the file:// protocol, but that did not work either. I have started to look at Vite to see if they have any known issues with rollup on Windows.

lilnasy:
I had to install react and vue because you have their integrations added but not the frameworks themselves, but afterwards I could not reproduce the issue.

<img width="500" src="https://github-production-user-asset-6210df.s3.amazonaws.com/69170106/270198513-b1b47be0-5e4b-472f-b4d2-cef76417118a.png">

matthewp:
[...]

PR #8741, opened 2023-10-03
fix(windows): ensure drive letter is uppercase

lilnasy:
## Changes
- Fix for #8470
- Issue still affects Vite.

## Testing
😶‍🌫️

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/astro.js
+++ b/packages/astro/astro.js
@@ -32,6 +32,13 @@ async function main() {
 		}
 	}
 
+  // windows drive letters can sometimes be lowercase, which vite cannot process
+  if (process.platform === 'win32') {
+    const cwd = process.cwd()
+    const correctedCwd = cwd.slice(0, 1).toUpperCase() + cwd.slice(1)
+    process.chdir(correctedCwd)
+  }
+
 	return import('./dist/cli/index.js')
 		.then(({ cli }) => cli(process.argv))
 		.catch((error) => {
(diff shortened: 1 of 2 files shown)


packages/astro/astro.js
```diff
@@ -32,6 +32,13 @@ async function main() {
 		}
 	}
 
+  // windows drive letters can sometimes be lowercase, which vite cannot process
+  if (process.platform === 'win32') {
+    const cwd = process.cwd()
+    const correctedCwd = cwd.slice(0, 1).toUpperCase() + cwd.slice(1)
+    process.chdir(correctedCwd)
```
bluwy, line 39: Should we skip this it's already uppercase? And does this also fix the issue if the drive letter is originally lowercase?

bluwy: IIUC there could be two case:

1. Originally `C:/something/...` but the terminal is on `c:/something/...` (this is fixed)
2. Originally `c:/something/...` but the terminal is on `C:/something/...`

I'm not sure if no2 is a real-world scenario, but if so I was thinking the the `correctedCwd` would have it lowercase. But if no2 isn't something we should worry about, then the current code should be fine then.

lilnasy: Yeah, while looking around github and discord threads, I did not come across a case where a lowercase drive letter is intentional on any software or user's part.

ematipico (approved):

Issue #8567, opened 2023-09-15
Unnecessary CSS files in dist


artursopelnik:
### Astro Info

```block
Astro                    v3.0.10
Node                     v20.5.1
System                   macOS (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/react
```

### If this issue only occurs in one browser, which browser is a problem?

Not browser-related

### Describe the Bug

When I update Astro from version 3.0.8 to 3.0.10, Vite bundles CSS files in my _astro directory in dist. These files are still bundled in the latest version, but the CSS files were never assigned. It's a bug.

## Screenshot
https://ibb.co/bvrvD3D

### What's the expected result?

No unnecessary CSS files that are not included in the dist.

### Link to Minimal Reproducible Example

#

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for opening this issue.

Due to a Vite behavior (#8329), CSS files had to be included (#8351).

Since the extra files are small and don't break anything, this would be low priority.

If someone would like to work on this issue, bluwy provided helpful pointers in #8351.

pratik-codes:
Hey @lilnasy I would like to give it a shot

lilnasy:
Thanks @pratik-codes
Go for it!

pratik-codes:
`When I update Astro from version 3.0.8 to 3.0.10, and run ´npm run dev´, Astro creates unnecessary CSS files in dist folder. These files are still bundled in the latest version, but the CSS files were never assigned. It's definitely a bug`

To check this bug I wanted to reproduce this so I created a new astro project and did the same thing. But it wasn't creating an unnecessary css file in the dist folder. 

To be sure about it I removed all the CSS from the project and when built it didn't even create a CSS file. 
<img width="473" alt="Screenshot 2023-09-25 at 11 40 31 PM" src="https://github.com/withastro/astro/assets/64960569/bc3a7b99-98bc-416b-9191-6bfbaf62557d">


Attaching a screenshot above. 

@lilnasy I think there is no issue or do you think I am doing anything wrong here?

lilnasy:
That is probably a typo - `npm run dev` does **not** affect the `dist` folder - we are conserved with `npm run build`.

pratik-codes:
@lilnasy correct I understood it was a typo. I tried npm run build only and it didn't generate anything when there was no CSS in Dist. I will also share a video for better understanding of this.

[...]

PR #8743, opened 2023-10-04
Remove unused CSS output files when inlined

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/8567

Fixes 2 issues:
1. Fix Vite preloading CSS references by updating `importedCss`.
2. CSS files inlined to any pages should not be deleted.

I'm actually unsure if there's a bug in no2, where in the `css-dangling-references` fixture, the CSS is not inlined/duplicated to both the pages.

## Testing

The existing tests (especially `css-dangling-references`) should pass

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/core/build/plugins/plugin-css.ts
+++ b/packages/astro/src/core/build/plugins/plugin-css.ts
@@ -200,7 +200,7 @@ function rollupPluginAstroBuildCSS(options: PluginOptions): VitePlugin[] {
 			const inlineConfig = settings.config.build.inlineStylesheets;
 			const { assetsInlineLimit = 4096 } = settings.config.vite?.build ?? {};
 
-			Object.entries(bundle).forEach(([_, stylesheet]) => {
+			Object.entries(bundle).forEach(([id, stylesheet]) => {
 				if (
 					stylesheet.type !== 'asset' ||
 					stylesheet.name?.endsWith('.css') !== true ||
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/build/plugins/plugin-css.ts
```diff
@@ -243,8 +248,21 @@ function rollupPluginAstroBuildCSS(options: PluginOptions): VitePlugin[] {
...
 						propagatedStyles.add(sheet);
+						sheetAddedToPage = true;
 					});
 				});
+
+				if (toBeInlined && sheetAddedToPage) {
+					// CSS is already added to all used pages, we can delete it from the bundle
+					// and make sure no chunks reference it via `importedCss` (for Vite preloading)
+					// to avoid duplicate CSS.
+					delete bundle[id];
```
ematipico, line 259: It's better to avoid `delete`
```js
bundle[id] = undefined
```
This page contains some information about why: https://biomejs.dev/linter/rules/no-delete

I predict that doing so will trigger the TS compiler, though

ematipico (approved):

artursopelnik:
+1

packages/astro/src/core/build/plugins/plugin-css.ts
bluwy: The convention when dealing with rollup bundles is to `delete` it rather than setting it to undefined. Otherwise it kinda means that "the chunk exists but it's undefined".

lilnasy (approved):
This is great!
I was unsure about the timing of our "post" plugin against vite's preload plugin, but I can now see that [it is "post"](https://github.com/vitejs/vite/blob/v4.4.10/packages/vite/src/node/build.ts#L452-L453) as well.

Tested against threlte docs as well, which is where the issue was discovered. Works perfectly!

lilnasy:
> 2. CSS files inlined to any pages should not be deleted.
That doesn't sound right.

> I'm actually unsure if there's a bug in no2, where in the css-dangling-references fixture, the CSS is not inlined/duplicated to both the pages.
You're right, the fixture does not work how I imagined. The html is not rendered either. I think what we might be noticing is that `import.meta.glob()` is not compatible with rendering on the server.

bluwy:
> > 3\. CSS files inlined to any pages should not be deleted.
> 
> That doesn't sound right.

Sorry I messed up my wording a bit 😅 I mean CSS files **not** inlined to any pages should not be deleted. That was why I added the `sheetAddedToPage` guard.

PR #8745, opened 2023-10-04
Support view transitions for client:only components

martrapp:
## Changes

The original issue is #8114. Persistent `client:only` components lose their styling on transition.
This only happens with `astro dev`, not with `astro build`.
The reason was that the HTML files contained only statically generated styles, and the comparison between the old DOM and the new file removed the client-side generated styles.
Fix #8472 fixed this by attempting to identify these client-generated styles and retain them.
This caused several issues where old styles were not deleted during the transition, even though they should have been. (#8604, #8674)

**My preferred solution would be to be able to read from each client:only astro-island (or each page) which head elements it dynamically creates. But unfortunately I have not been able to do that for dev mode. Any ideas how to achieve this?**

This fix identifies head elements that are dynamically inserted by persisted client:only components.
This is done by diffing the head elements that are parsed from file with those found in the DOM right before transitioning away from the document. Of course this only helps for the next transition to this page. Therefore the page is full reloaded if the information about dynamic head elements is not available yet.

Usage: visit all pages once by navigating to them using view transitions.
After that pre-loading, few transitions work as expected on following page visits.

## Testing

Manually tested with code from #8604.
Manually tested with code from #8114

## Docs

n/a, bug fix

### DIFF
--- a/.changeset/lazy-keys-shout.md
+++ b/.changeset/lazy-keys-shout.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Support view transitions for client:only components
--- a/packages/astro/e2e/fixtures/view-transitions/src/pages/client-only-one.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/pages/client-only-one.astro
@@ -3,6 +3,7 @@ import Layout from '../components/Layout.astro';
 import Island from '../components/Island';
 ---
 <Layout>
+	<p id="page-one">Page 1</p>
 	<a id="click-two" href="/client-only-two">go to page 2</a>
 	<div transition:persist="island">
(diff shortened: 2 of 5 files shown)


matthewp:
I don't love the amount of code needed to fix this and the fact that we need to reload some times. Would love to find another solution. Let me think on this one for a little bit...

martrapp:
I share your assessment regarding the code. It would be great if we could find a simpler solution!

martrapp:
Workaround while we strive to find a good solution:

- Note that the problem only occurs in DEV mode, not when creating/previewing for production.
- In DEV mode, you can import the missing CSS files explicitly on the page that uses the "client:only" component.

Issue #7469, opened 2023-06-24
Strange naming of css files


niklasgrewe:
### What version of `astro` are you using?

2.6.6

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

i use a `global.css` file in my layout `base.astro` like this:

```html
---
import '../styles/global.css'
---

<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
</head>
<body>
  <slot></slot>
</body>
</html>
```
i use this `base` layout in my two pages `index.astro` and `about.astro` like this:

```html
---
import Base from '../layouts/Base.astro'
---

<Base>
  <h1>Home page</h1>
</Base>

<style>
  h1 {
    color: red;
  }
</style>
```
```html
---
import Base from '../layouts/Base.astro'
---

<Base>
  <h1>About page</h1>
</Base>

<style>
  h1 {
    color: black;
  }
</style>
```
when i build my simple site, i get the following css files:

- `index.[hash].css` - homepage styles
- `about.[hash].css` - styles from `global.css` file, imported in `base.astro`
- `about.[hash].css` - about page styles

it is a bit strange that i have two `about.[hash].css` files.

### What's the expected result?

i like that the css files are named according to the pages, but why are imported css files, like my `global.css` file not named according to their file name or the layout file name?

i wan to get the following css files:
- `index.[hash].css` - homepage styles
- `global.[hash].css` or `base.[hash].css` - styles from `global.css` file, imported in `base.astro`
- `about.[hash].css` - about page styles 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wgxcan?file=src%2Fpages%2Fabout.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
This is currently expected. We derive the CSS file name by the first name of a list of pages that uses it, e.g. if both `index` and `about` uses it, the first (sorted alphabetically) page - `about` - would be the name.

I think it's somewhat possible to use the original file name, e.g. `global.css`, as the main name though, so will leave this open for now for investigation.

andywwright:
For now this is the only thing that feels wrong about Astro. This naming is very misleading and took me a while to investigate.

[...]

PR #8754, opened 2023-10-05
Make CSS chunk names less confusing

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/7469

This PR implements a lame algorithm to pick a chunk base name that is not confusing. Simply by avoiding the `404` and `500` names if there's others available.

I couldn't quite figure out a better naming-scheme without overcomplicating it. I also tried resorting this to Rollup, but there's edgecases where that's not possible. Our CSS handling that links CSS to HTML pages are very ingrained throughout the codebase. Also in practice, Rollup would also pick the confusing name otherwise.

## Testing

Tested manually.

## Docs

n/a.

### DIFF
--- a/packages/astro/src/core/build/css-asset-name.ts
+++ b/packages/astro/src/core/build/css-asset-name.ts
@@ -1,4 +1,4 @@
-import type { GetModuleInfo } from 'rollup';
+import type { GetModuleInfo, ModuleInfo } from 'rollup';
 
 import crypto from 'node:crypto';
 import npath from 'node:path';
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/build/css-asset-name.ts
```diff
@@ -1,25 +1,34 @@
...
 
 import crypto from 'node:crypto';
 import npath from 'node:path';
 import type { AstroSettings } from '../../@types/astro.js';
 import { viteID } from '../util.js';
 import { getTopLevelPages } from './graph.js';
 
+// These pages could be used as base names for the chunk hashed name, but they are confusing
+// and should be avoided it possible
+const confusingBaseNames = ['404', '500'];
```
Princesseuh, line 11: You might want to avoid dynamic routes as well, for instance, my website's CSS file is named `_...page_.1d4448e7.css` (with the underscore at the start, weirdly enough)

bluwy: I could perhaps look into prettify some names too 🤔

lilnasy:
> I couldn't quite figure out a better naming-scheme without overcomplicating it.
Do we need to include a real file's name at all? Could we let it be just the hash?

bluwy:
I think we still need some sort of memorable name so it's easier to debug things when needed.

packages/astro/src/core/build/css-asset-name.ts
bluwy: I updated it so `.` is converted to `_`, and multiple `_` are collapsed to a single `_`. So `_...page_.1d4448e7.css` should look like `_page_.1d4448e7.css` now. There could be same names if there's `[slug]` and `[...slug]` side-by-side, but I think it should be fine (hashes are still different so chunking is still correct).

matthewp (approved):

PR #8756, opened 2023-10-05
Fix scripts run with <ViewTransitions> on fallback swap and multiple inline <head> scripts 

leomp12:
Fixes for <ViewTransitions> with "SPA" mode on v2 (not sure about v3):

## Changes

1. Multiple <head> inline scripts are removed cause second one is being considered external with `(s1.type === s2.type && s1.src === s2.src)` (s1.src === s2.src === undefined), an [example real use case](https://github.com/ecomplus/cloud-commerce/blob/main/packages/storefront/src/lib/layouts/BaseHead.astro#L172-L176) with a persistent inline script (all site shared state) plus a page-context inline script (page state);

2. `runScripts` is called before DOM replacement on fallback mode (Firefox) cause fallback swap is async and not awaited, this cause scripts not being executed on first page transition, on next transition (3º page) the scripts of past page (2º) are executed...

## Testing

I'm using a custom <ViewTransitions> component [here](https://github.com/ecomplus/cloud-commerce/blob/main/packages/storefront/src/lib/components/ViewTransitions.astro) to [fix these problems](https://github.com/ecomplus/cloud-commerce/commit/2e5320718bfa76c6e7f6258031a546e3f9089aa4) , both bugs resolved after changes.

I can try to create a minimal reproduction on Stackblitz later if necessary.

## Docs

N/A

### DIFF
--- a/.changeset/stale-planes-prove.md
+++ b/.changeset/stale-planes-prove.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix scripts run with <ViewTransitions> on fallback swap and multiple inline <head> scripts
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -122,7 +122,7 @@ const { fallback = 'animate' } = Astro.props as Props;
 						// Inline
 						(s1.textContent && s1.textContent === s2.textContent) ||
 						// External
-						(s1.type === s2.type && s1.src === s2.src)
+						(s1.type === s2.type && s1.src && s1.src === s2.src)
 					) {
 						return s2;
 					}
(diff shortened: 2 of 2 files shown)


leomp12:
Hey @ematipico , it's still relevant for v2, not sure about v3 (main).
v2 is not being maintained anymore?

lilnasy:
@leomp12 That is correct. So far we have only ever released new versions for the currently stable version, which is v4 currently.

martrapp:
Hi @leomp12! I'm sorry that I didn't notice your PR. 

A lot has happend since 2.x. The whole structure of ViewTransition.astro has been refactored since then. The code you are referring to is no in router.ts. The first bug has been solved. Regarding the second one: Waiting for animations has been altered to better handle infinite animations. Also the whole simulation logic ha been updated. 

The bugs you addressed have been solved. Could you please try with a current 4.0 version of Astro if you still have issues?

matthewp:
Closing as we don't want to fix bugs in an experimental feature. Please do upgrade to Astro 4 where this should be working. Thank you for submitting!

leomp12:
Hey guys, tbh I'm using a customized version of `<ViewTransitions>` for a long time, now with Astro v3 (yet) because v4 broke on my first try (5 minutes thing), apparently because of some Vue integration bug (global instance). I'll try it again later.
Thank you all.

Issue #8164, opened 2023-08-20
Spaces in project path break Markdoc integration


suhaylmv:
### What version of `astro` are you using?

2.10.12

### Are you using an SSR adapter? If so, which one?

No

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Firefox

### Describe the Bug

Had a fully functional blog which uses .md files, but after installing markdoc and creating the config file, right when I rename a .md file to .mdoc  throws this error:

```
Failed to load url /mnt/D/Docs/Web%20projects/randomproject/markdoc.config.mjs (resolved id: /mnt/D/Docs/Web%20projects/randomproject/markdoc.config.mjs) in /mnt/D/Docs/Web projects/randomproject/src/content/blog/randomblogpost.mdoc. Does the file exist?
```

### What's the expected result?

It should keep working 

### Link to Minimal Reproducible Example

Link to my project repo: https://github.com/suhaylmv/minov-studio . Check the branch "test"

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for bringing this up!

It's probably going to be difficult to diagnose it from your real project that you linked. Could you create a minimal reproduction?

suhaylmv:
I installed this markdoc template  https://github.com/withastro/astro/tree/latest/examples/with-markdoc , and when I run npm run dev it throws exactly the same error. Could you please try this template and see if it works for you? If it does, then it must be problem of my machine

lilnasy:
It seems to work fine, both in dev and build.

<img width="960" alt="image" src="https://github.com/withastro/astro/assets/69170106/5c616698-75f2-4909-95dc-124ee7641220">

suhaylmv:
That's odd. I think the problem is the OS. Now I'm on Elementary OS (debian based), but before when I used Arch I never had such issues. I had a fuckton of other problems too because of Elementary. Distro hopping is never going to end). Thank you for help!

suhaylmv:
I made a fresh install of pop os and I'm having the same problem. Idk what to do

suhaylmv:
Ok I found a solution. For some reason when the project folder is inside an external hdd drive and the project uses markdoc, gives this error. I would suggest trying the same on other machines to see if the same error occurs. If it does, then maybe it's worth adding the information to the troubleshooting page in the documentation.

lilnasy:
You also seem to have a space in your project path, could that be the issue?

[...]

PR #8759, opened 2023-10-05
Fix: Markdoc Integration build when root folder contains spaces

lutaok:
Hello,
opening this PR to fix Issue #8164.

## Changes

- Added `decodeURI` function call to pathnames representing root folder in `markdoc` plugin

## Testing

- Added a test case for rendering and building with a root project folder that contains spaces
- Manually tested with an existing `markdoc` project with same characteristics:
  - `pnpm link` on local Astro project containing fix
  - `astro build` on the existing project
  - No errors shown, build completed successfully

## Docs

- No workaround needed for the root project folder's name
- No docs were added because this issue wasn't showing on docs

### DIFF
--- a/packages/integrations/markdoc/src/content-entry-type.ts
+++ b/packages/integrations/markdoc/src/content-entry-type.ts
@@ -106,7 +106,7 @@ export async function getContentEntryType({
 import { createGetHeadings, createContentComponent } from '@astrojs/markdoc/runtime';
 ${
 	markdocConfigUrl
-		? `import markdocConfig from ${JSON.stringify(markdocConfigUrl.pathname)};`
+		? `import markdocConfig from ${JSON.stringify(decodeURI(markdocConfigUrl.pathname))};`
 		: 'const markdocConfig = {};'
 }
 
@@ -232,7 +232,9 @@ function getStringifiedImports(
 		const resolvedPath =
 			config.type === 'local' ? new URL(config.path, root).pathname : config.path;
 
-		stringifiedComponentImports += `import ${importName} from ${JSON.stringify(resolvedPath)};\n`;
+		stringifiedComponentImports += `import ${importName} from ${JSON.stringify(
+			decodeURI(resolvedPath)
+		)};\n`;
 	}
 	return stringifiedComponentImports;
 }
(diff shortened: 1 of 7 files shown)


packages/integrations/markdoc/src/content-entry-type.ts
```diff
@@ -106,7 +106,7 @@ export async function getContentEntryType({
 import { createGetHeadings, createContentComponent } from '@astrojs/markdoc/runtime';
 ${
 	markdocConfigUrl
-		? `import markdocConfig from ${JSON.stringify(markdocConfigUrl.pathname)};`
+		? `import markdocConfig from ${JSON.stringify(decodeURI(markdocConfigUrl.pathname))};`
```
bluwy, line 109: I think it would be safer we call `fileUrlToPath(markdocConfigUrl)` instead, since this makes sure it's a valid fs path.

packages/integrations/markdoc/src/content-entry-type.ts
```diff
@@ -232,7 +232,9 @@ function getStringifiedImports(
 		const resolvedPath =
 			config.type === 'local' ? new URL(config.path, root).pathname : config.path;
 
-		stringifiedComponentImports += `import ${importName} from ${JSON.stringify(resolvedPath)};\n`;
+		stringifiedComponentImports += `import ${importName} from ${JSON.stringify(
+			decodeURI(resolvedPath)
+		)};\n`;
```
bluwy, line 237: We can also fix this similarly by only changing the `new URL(config.path, root).pathname` to `fileUrlToPath(new URL(config.path, root))`

bluwy (changes requested):

packages/integrations/markdoc/src/content-entry-type.ts
lutaok: Hello, thank you for taking the time to review my PR. 
I've updated the code to use the safer API you pointed out!

lilnasy (approved):
LGTM, thanks for the tests as well!

bluwy (approved):

bluwy:
Thanks for fixing this!

natemoo-re:
Closes #8164

PR #8766, opened 2023-10-06
fix(cloudflare): support for 'cloudflare:*' imports

jadbox:
## Changes

- Allow the use of the runtime library `cloudflare:sockets` without throwing a build error by adding it to the external list.

## WHY?

This update is needed for merge in order to further work on Postgres.js to allow for SQL socket connections for the CF tcp connec() api.

## Testing

Try to reference cloudflare:sockets in an import.

## Docs

https://blog.cloudflare.com/workers-tcp-socket-api-connect-databases/

### DIFF
--- a/.changeset/heavy-elephants-tan.md
+++ b/.changeset/heavy-elephants-tan.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Adding cloudflare:sockets compile support
--- a/.changeset/heavy-elephants-tan.md
+++ b/.changeset/heavy-elephants-tan.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Adding cloudflare:sockets compile support
+Adds `cloudflare:sockets` compile support
--- a/packages/integrations/cloudflare/README.md
+++ b/packages/integrations/cloudflare/README.md
@@ -357,6 +357,10 @@ import { Buffer } from 'node:buffer';
 
 Additionally, you'll need to enable the Compatibility Flag in Cloudflare. The configuration for this flag may vary based on where you deploy your Astro site. For detailed guidance, please refer to the [Cloudflare documentation on enabling Node.js compatibility](https://developers.cloudflare.com/workers/runtime-apis/nodejs).
 
+## Cloudflare module support
+
+All Cloudflare namespaced packages (like `cloudflare:sockets`) are whitelisted for use. Note that the package `cloudflare:sockets` does not work locally without using Wrangler dev mode.
+
 ## Preview with Wrangler
 
 To use [`wrangler`](https://developers.cloudflare.com/workers/wrangler/) to run your application locally, update the preview script:
--- a/packages/integrations/cloudflare/README.md
+++ b/packages/integrations/cloudflare/README.md
@@ -359,7 +359,7 @@ Additionally, you'll need to enable the Compatibility Flag in Cloudflare. The co
 
 ## Cloudflare module support
 
-All Cloudflare namespaced packages (like `cloudflare:sockets`) are whitelisted for use. Note that the package `cloudflare:sockets` does not work locally without using Wrangler dev mode.
+All Cloudflare namespaced packages (e.g. `cloudflare:sockets`) are allowlisted for use. Note that the package `cloudflare:sockets` does not work locally without using Wrangler dev mode.
 
 ## Preview with Wrangler
 
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -290,7 +290,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 								'node:stream',
 								'node:string_decoder',
 								'node:util',
-								'cloudflare:sockets'
+								'cloudflare:*'
 							],
 							entryPoints: pathsGroup,
 							outbase: absolutePagesDirname,
@@ -372,7 +372,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							'node:stream',
 							'node:string_decoder',
 							'node:util',
-							'cloudflare:sockets'
+							'cloudflare:*'
 						],
 						entryPoints: [entryPath],
 						outfile: buildPath,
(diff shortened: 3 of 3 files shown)


jadbox:
This update is needed to merging to further work on Postgres.js to allow for SQL socket connections for the CF tcp connec() api.

lilnasy (approved):
Clean implementation, thanks!

You should know the cloudflare adapter is in the middle of a migration to https://github.com/withastro/adapters. Hopefully, it will be done by the end of next week. This PR would probably be merged in that repo.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -290,6 +290,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 								'node:stream',
 								'node:string_decoder',
 								'node:util',
+								'cloudflare:sockets'
```
alexanderniebuhr, line 293:
```suggestion
								'cloudflare:*'
```

packages/integrations/cloudflare/src/index.ts
```diff
@@ -371,6 +372,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							'node:stream',
 							'node:string_decoder',
 							'node:util',
+							'cloudflare:sockets'
```
alexanderniebuhr, line 375:
```suggestion
							'cloudflare:*'
```

alexanderniebuhr (changes requested):
Thanks for the contribution, do you think it would be better to support all `cloudflare:` prefixed imports, given that there might be more right now and coming in the future.

Also I'm just going to block this for now, since as @lilnasy said, we would probably merge it in the new repo. I'll follow up at least every other day to keep you in the loop!

jadbox:
@alexanderniebuhr thanks for the update. This is a project blocker for us until resolved, but I understand the need to finish the adapter repo change.

If we can use wildcard for cloudflare:* packages (and that's esbuild valid syntax?), I think we should also wildcard all node:* packages too since CF continually adds new packages here.

On a large point outside of esbuild externals, it seems limiting presently to not be able to override any of the esbuild settings in the client from outside the adapter. Cloudflare updates frequently, and the adapter seems to need an update every time due to esbuild directives. The above pkg external wildcards would help, but not in all cases. Maybe the adapter(s) client interface should take an optional esbuild parameters override config object?

alexanderniebuhr:
> This is a project blocker for us until resolved.

Ok, I understand. We can also get this PR finished here, **but would need to open the same PR on the new repo too.**
I would like to see added documentation, for the feature _(depending if we go for the wildcard or just the `sockets`)_ & link to the related Cloudflare docs.

> If we can use wildcard for cloudflare:* packages (and that's esbuild valid syntax?)

I think it is valid syntax, would you mind to research it. IIRC the SvelteKit Cloudflare adapter uses `cloudflare:*`.

> I think we should also wildcard all node:* packages too since CF continually adds new packages here.

I don't think this is a valid approach. What happens if the user uses a Node.JS Api which is not supported, we would not warn the user, but running the project would fail. The difference here is that all possible imports from `cloudflare:*`, will be supported by Cloudflare. While for the `node:*` prefix only a subset of imports are supported, so we should be very explicit about the supported options and should only externalize those.

> On a large point outside of esbuild externals, it seems limiting presently to not be able to override any of the esbuild settings in the client from outside the adapter. Cloudflare updates frequently, and the adapter seems to need an update every time due to esbuild directives. The above pkg external wildcards would help, but not in all cases. Maybe the adapter(s) client interface should take an optional esbuild parameter selective override config object?

Thank you for the idea. While I personally don't like to allow esbuild settings overwrites, it's definitely outside the current PR's scope. Let's explore this in a separate PR to fully address its implications and documentation needs. I personally would like to see a broader discussion first, regarding this idea. IIRC the Svelte team already came up with some arguments in favor & against that idea. _(I think you even linked them yourself, somewhere)_

jadbox:
@alexanderniebuhr I've committed a patch to wildcard cloudflare:* packages.

alexanderniebuhr:
Thank you. Can you also add / update the documentation in the `README.md` file?

jadbox:
@alexanderniebuhr Readme updated under `## Cloudflare module support`

alexanderniebuhr:
@jadbox thank you so much. Once docs approve, we will get this merged here & I will make sure it get's moved over to the new repo!

alexanderniebuhr (approved):

.changeset/heavy-elephants-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Adding cloudflare:sockets compile support
```
sarah11918, line 5:
```suggestion
Adds `cloudflare:sockets` compile support
```
If we want to keep our practice of starting with singular, present tense verbs!

packages/integrations/cloudflare/README.md
```diff
@@ -357,6 +357,10 @@ import { Buffer } from 'node:buffer';
 
 Additionally, you'll need to enable the Compatibility Flag in Cloudflare. The configuration for this flag may vary based on where you deploy your Astro site. For detailed guidance, please refer to the [Cloudflare documentation on enabling Node.js compatibility](https://developers.cloudflare.com/workers/runtime-apis/nodejs).
 
+## Cloudflare module support
+
+All Cloudflare namespaced packages (like `cloudflare:sockets`) are whitelisted for use. Note that the package `cloudflare:sockets` does not work locally without using Wrangler dev mode.
```
sarah11918, line 362:
```suggestion
All Cloudflare namespaced packages (e.g. `cloudflare:sockets`) are allowlisted for use. Note that the package `cloudflare:sockets` does not work locally without using Wrangler dev mode.
```
So, it's mostly considered unacceptable to use the terms "whitelist" and "blacklist" now. I'm happy with any of the common alternatives for now, though admittedly, Astro hasn't standardized on one. We can just choose any acceptable alternative for now, and update if we do make a standard decision! 

"allowlist"  (vs "denylist") is the option I see most commonly used/recommended

sarah11918 (commented):
Thanks for adding these docs @jadbox ! Just a few tiny changes to language/formatting to get it in to docs. Then, Docs is happy!  🙌

jadbox:
@sarah11918 Requested doc changes committed.

PR #8771, opened 2023-10-07
fix(transitions router): no-op on the server

lilnasy:
## Changes
- Prevents import side-effects from running during SSR.
- Makes `navigate()` do nothing on the server.

## Testing
_Pending_.

## Docs
There should be a note about framework usage somewhere. _Pending_.

### DIFF
--- a/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
+++ b/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
 import { navigate } from "astro:transitions/client";
 export default function ClickToNavigate({ to, id }) {
-	return <button id={id} onClick={() => navigate(to)}>Navigate to `{to}`</button>;
+    return <button id={id} onClick={() => navigate(to)}>Navigate to `{to}`</button>;
 }
--- a/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
+++ b/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-
+import { navigate } from "astro:transitions/client";
 export default function ClickToNavigate({ to, id }) {
 	return <button id={id} onClick={() => navigate(to)}>Navigate to `{to}`</button>;
 }
--- a/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
+++ b/packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export default function ClickToNavigate({ to, id }) {
+	return <button id={id} onClick={() => navigate(to)}>Navigate to `{to}`</button>;
+}
--- a/packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
@@ -1,4 +1,12 @@
 ---
 import ClickToNavigate from "../components/ClickToNavigate.jsx"
+import { ViewTransitions } from "astro:transitions";
 ---
-<ClickToNavigate id="react-client-load-navigate-button" to="/two" client:load/>
+<html>
+	<head>
+		<ViewTransitions />
+	</head>
+	<body>
+		<ClickToNavigate id="react-client-load-navigate-button" to="/two" client:load/>
+	</body>
+</html>
--- a/packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
@@ -1,0 +1,4 @@
+---
+import ClickToNavigate from "../components/ClickToNavigate.jsx"
+---
+<ClickToNavigate id="react-client-load-navigate-button" to="/two" client:load/>
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -19,7 +19,7 @@ const inBrowser = import.meta.env.SSR === false;
 export const supportsViewTransitions = inBrowser && !!document.startViewTransition;
 
 export const transitionEnabledOnThisPage = () =>
-    supportsViewTransitions && !!document.querySelector('[name="astro-view-transitions-enabled"]');
+    inBrowser && !!document.querySelector('[name="astro-view-transitions-enabled"]');
 
 const samePage = (otherLocation: URL) =>
 	location.pathname === otherLocation.pathname && location.search === otherLocation.search;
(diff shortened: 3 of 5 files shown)


lilnasy:
The test I added is failing, not sure why. But at least the build and server render didn't fail as it would have before this PR.

packages/astro/src/transitions/router.ts
```diff
@@ -13,9 +13,12 @@ type Events = 'astro:page-load' | 'astro:after-swap';
 // only update history entries that are managed by us
 // leave other entries alone and do not accidently add state.
 const persistState = (state: State) => history.state && history.replaceState(state, '');
-export const supportsViewTransitions = !!document.startViewTransition;
+
+export const supportsViewTransitions = import.meta.env.SSR === false && !!document.startViewTransition;
+
 export const transitionEnabledOnThisPage = () =>
-	!!document.querySelector('[name="astro-view-transitions-enabled"]');
+	import.meta.env.SSR === false && !!document.querySelector('[name="astro-view-transitions-enabled"]');
```
ematipico, line 20: Can't we use `supportsViewTransitions && !!document...` here instead?

packages/astro/src/transitions/router.ts
```diff
@@ -365,6 +376,9 @@ async function transition(
 }
 
 export function navigate(href: string, options?: Options) {
+	
+	if (import.meta.env.SSR) return;
```
ematipico, line 380: Can't we use `supportsViewTransitions` here instead?

ematipico (commented):
I would use `supportsViewTransitions` for all the checks. The name of the variable is more descriptive

packages/astro/src/transitions/router.ts
lilnasy: I went with import.meta because it has implications on tree-shaking, but it might extend to variables as well. I'll see.

packages/astro/e2e/fixtures/view-transitions/src/components/ClickToNavigate.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
```
martrapp, line 2:
```suggestion
import { navigate } from "astro:transitions/client";
```

packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
```diff
@@ -0,0 +1,4 @@
+---
+import ClickToNavigate from "../components/ClickToNavigate.jsx"
+---
+<ClickToNavigate id="react-client-load-navigate-button" to="/two" client:load/>
```
martrapp, line 4:
```suggestion
import { ViewTransitions } from "astro:transitions";
---
<html>
	<head>
		<ViewTransitions />
	</head>
	<body>
		<ClickToNavigate id="react-client-load-navigate-button" to="/two" client:load/>
	</body>
</html>
```

martrapp: Client-site router only works with ViewTransitions

packages/astro/src/transitions/router.ts
martrapp: Like ematipico's idea. After `supportsViewTransitions` is once defined using `import.meta.env.SSR` we should be good without further `import.meta.env.SSR` expression and it looks less brutal :)

packages/astro/e2e/fixtures/view-transitions/src/pages/client-load.astro
lilnasy: Thanks! I was having a moment there

martrapp (approved):

ematipico (approved):

packages/astro/src/transitions/router.ts
```diff
@@ -373,6 +384,9 @@ async function transition(
 }
 
 export function navigate(href: string, options?: Options) {
+	
+	if (!supportsViewTransitions) return;
+
```
bluwy, line 389: Should we throw an error on SSR instead? Or maybe remove this completely? I think there's a bigger problem if the user is calling these client APIs in SSR.

If view transitions is not supported in the client, it's already handled below at `if (!transitionEnabledOnThisPage()) {`

lilnasy: The check mostly serves as an opportunity for tree-shaking.

I went back and forth on no-op, error and warning. Ultimately, it's a very unlikely scenario that `navigate()` gets called without user interaction so I stopped thinking about it.

I already wrote the docs saying it does nothing on the server. I'd rather not go back to that PR. Would a warning be acceptable here?

bluwy: Yeah a warning would be fine by me if it's already documented as a no-op 👍

packages/astro/src/transitions/router.ts
```diff
@@ -372,7 +383,21 @@ async function transition(
 	}
 }
 
+let navigateOnSeverWarned = false;
+
 export function navigate(href: string, options?: Options) {
+	
+	if (!supportsViewTransitions) {
```
bluwy, line 390: For treeshaking, I made some tests and I think this condition needs to be `if (import.meta.env.SSR)` for it to work, otherwise the client would include the warning. Since `document.startViewTransition` is non-static.

lilnasy: Good catch!

bluwy: There's still the typo in case you missed it 👀

lilnasy: thanks, corrected

bluwy (approved):

packages/astro/src/transitions/router.ts
```diff
@@ -13,9 +13,14 @@ type Events = 'astro:page-load' | 'astro:after-swap';
...
 const persistState = (state: State) => history.state && history.replaceState(state, '');
-export const supportsViewTransitions = !!document.startViewTransition;
+
+const inBrowser = import.meta.env.SSR === false;
+
+export const supportsViewTransitions = inBrowser && !!document.startViewTransition;
+
 export const transitionEnabledOnThisPage = () =>
-	!!document.querySelector('[name="astro-view-transitions-enabled"]');
+    supportsViewTransitions && !!document.querySelector('[name="astro-view-transitions-enabled"]');
```
martrapp, line 22:
```suggestion
    inBrowser && !!document.querySelector('[name="astro-view-transitions-enabled"]');
```

lilnasy: Wouldn't this skip the check for `'startViewTransition' in document`? Is that intentional?

martrapp: The original code set `transitionEnableOnThisPage` to `!!document.querySelector('[name="astro-view-transitions-enabled"]')` independent of the existence of `document.startViewTransition` 
The new code should do the same (but only in the browser)

packages/astro/src/transitions/router.ts
```diff
@@ -40,13 +45,17 @@ const announce = () => {
...
-const parser = new DOMParser();
+
+let parser: DOMParser
 
 // The History API does not tell you if navigation is forward or back, so
 // you can figure it using an index. On pushState the index is incremented so you
 // can use that to determine popstate if going forward or back.
 let currentHistoryIndex = 0;
+
+if (supportsViewTransitions) {
```
martrapp, line 58:
```suggestion
if (inBrowser) {
```

martrapp (commented):
found another one...

PR #8773, opened 2023-10-07
fix(assets): add the missing `await` to getHTMLAttributes

sumimakito:
## Changes

This pull request adds the missing `await` to properly use values returned by `getHTMLAttribute` in image services. It fixes an issue where HTML attributes do not render if getHTMLAttributes in an image service returns a Promise.

Minimum repro for this issue: [StackBlitz](https://stackblitz.com/edit/repro-astro-image-service-async-gethtmlattributes?file=src%2Fimage-service-async.ts) (You can see that the custom HTML attributes are not rendered in the HTML if `getHTMLAttribute` returns a Promise instead a plain object)

Definition for `getHTMLAttribute`:

https://github.com/withastro/astro/blob/30de324361bc261956eb9fc08fe60a82ff602a9b/packages/astro/src/assets/services/service.ts#L41-L50

## Testing

No tests were added as this is a one-line fix. It should have no impact if `getHTMLAttributes` returns a non-Promise. 

## Docs

No docs were needed as this fix does not change the expected behavior.

### DIFF
--- a/.changeset/young-taxis-battle.md
+++ b/.changeset/young-taxis-battle.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix an issue where HTML attributes do not render if getHTMLAttributes in an image service returns a Promise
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
Awesome, thank you!

PR #8774, opened 2023-10-07
chore(tests): allow mocha to reattempt failing cases

lilnasy:
## Changes
Internal

## Testing
Does not affect behavior.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro-rss/package.json
+++ b/packages/astro-rss/package.json
@@ -24,7 +24,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc",
     "build:ci": "astro-scripts build \"src/**/*.ts\"",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "mocha --exit --timeout 20000"
+    "test": "mocha --exit --retries 3 --timeout 20000"
   },
   "devDependencies": {
     "@types/chai": "^4.3.5",
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -114,7 +114,7 @@
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
     "test:unit": "mocha --exit --timeout 30000 ./test/units/**/*.test.js",
     "test:unit:match": "mocha --exit --timeout 30000 ./test/units/**/*.test.js -g",
-    "test": "pnpm run test:unit && mocha --exit --timeout 30000 --ignore **/lit-element.test.js && mocha --timeout 30000 **/lit-element.test.js",
+    "test": "pnpm run test:unit && mocha --exit --timeout 30000 --retries 3 --ignore **/lit-element.test.js && mocha --timeout 30000 **/lit-element.test.js",
     "test:match": "mocha --timeout 30000 -g",
     "test:e2e": "playwright test",
     "test:e2e:match": "playwright test -g"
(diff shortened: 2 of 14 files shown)


lilnasy:
Might be a better idea to use mocha's `--retry` flag.

natemoo-re (approved):
LGTM! Makes sense for our sometimes flaky tests.

Does make it a little harder to track down flakes, but I suppose it's better if CI just handles flakes properly.

lilnasy:
Yeah, I was hoping mocha would report the tests it had to retry. Doesn't seem like it.

lilnasy:
Apparently, It emits an event on every retry. So there is a way to get the info.

natemoo-re:
> Yeah, I was hoping mocha would report the tests it had to retry. Doesn't seem like it.

Looks like they do have a [`EVENT_TEST_RETRY`](https://mochajs.org/api/tutorial-custom-reporter) event that bubbles up to the reporter, so maybe it's just that our reporter doesn't display that event?

lilnasy:
It works.
image

It requires creating a custom reporter that replaces the default reporter. Maybe there's a way to extend the default, or inherit from it. I'd rather do that in a different PR.

bluwy:
I'm not really sure if this would be a good change. Since mocha tests (integrations/unit) should generally be stable, I think we should judiciously only retry certain tests that we find flaky, instead of enabling it globally. 

We have `this.retries()` littered to handle this. I think a global retry count would make more sense for e2e tests.

lilnasy:
From what I've noticed, pretty much any test that uses fetch is vulnerable. I don't see per-test retry being viable.

bluwy:
We could probably add some code to `fixture.startDevServer` or `fixture.preview` to ensure the server is really ready before before starting the fetches. But otherwise I don't feel comfortable setting a global number of retries.

It also affects local development as maybe there's 10 fails from a refactor, but because of the retries that 10 fails take forever to show up as they keep repeating intentionally failing code.

bluwy (changes requested):
Blocking until we can find a middleground 😬

lilnasy:
Those are good ideas!

They are more investment though. I think we should invest there but without blocking this PR.

With the custom reporter (maybe upstream: https://github.com/mochajs/mocha/issues/5014), we can let the logs guide us to the flaky tests at a convenient time. Currently, they intrude on and interrupt normal development.

bluwy:
These changes also affect normal development in another way though (like I mentioned above). I could be persuaded if it's retries only on CI, and not locally, but at that point the effort would be similar to fixing the core fetching issue.

lilnasy:
`pnpm test:ci`?

bluwy:
Yeah if we want to go that way, it would be `pnpm test:ci`. But I'm not really sure if it's worth the effort. I've also researched this before which the [Mocha docs](https://mochajs.org/#retry-tests) discourages the use of retries, so I think that would be the last resort.

I've added some additional logs for `fetch failed` before at https://github.com/withastro/astro/blob/9cd6a6657b328cec4ad98020e2bcd73c34a869ce/packages/astro/test/test-utils.js#L169-L179 but seems like it always gets buried. (appending directly to `err.message` could be better)

I think there's a few ways we should try before going with this PR. It casts a very wide net that we can easily miss flaky tests to come up again in the future.

ematipico:
@lilnasy is this PR still relevant? it has 3 weeks of inactivity

lilnasy:
Just waiting on @bluwy further comments.

lilnasy:
What would be a middleground you are comfortable with?

lilnasy:
> We probably just need to set some time to debug why the tests are flaky.

I disagree.  Tests are means to an end IMO. I am hesitatnt to consider them a source of maintenance burden in themselves.

PR #8776, opened 2023-10-08
Fix transition attributes on islands

### LINKED ISSUE #8361: Astro 3.0 with ViewTransition and Vue component
### Astro Info

```block
Astro                    v3.0.6
Node                     v20.5.0
Package Manager          pnpm
Adapter                  none
Integrations             @astrojs/vue
```

Chrome

### Describe the Bug

I am following examples from [docs](https://docs.astro.build/en/guides/view-transitions/#maintaining-state) and trying to get a simple `Counter` working where the component persists. However, my custom component disappears after navigation. 

My counter is placed in `Layout`

```
<MyCounter transition:persist transition:name="counter" client:load/>
```

With `MyCounter.vue`

```
<template>
  <div>Counter is {{ count }}</div>
</template>

<script setup>
import { ref } from "vue";
const count = ref(0);
setInterval(() => count.value++, 1000);
</script>

<style scoped>
div {
  color: red;
}
</style>
```

When navigating to another page, this component disappears. 

I expect this component to not disappear, and the `counter` value to persist. 

I have searched other bugs that are open and nothing seems similar. Perhaps this is a simple mistake because my example is from the docs. 

https://stackblitz.com/edit/github-sv58pq

### Participation

martrapp:
## Changes

Transition attributes where set on astro-island and their contents.
data-astro-transition-persist attributes on nested elements cause the elements to be lost on transitions.
The fix ensures that transition attributes are not inserted on island content, only on the island itself.

Closes #8361

## Testing

Added mocha test

## Docs

n/a bug fix

### DIFF
--- a/packages/astro/src/runtime/server/hydration.ts
+++ b/packages/astro/src/runtime/server/hydration.ts
@@ -15,10 +15,12 @@ export interface HydrationMetadata {
 	componentExport: { value: string };
 }
 
+type Props = Record<string | number | symbol, any>;
+
 interface ExtractedProps {
 	isPage: boolean;
 	hydration: HydrationMetadata | null;
-	props: Record<string | number | symbol, any>;
+	props: Props;
 }
 
 const transitionDirectivesToCopyOnIsland = Object.freeze([
@@ -26,10 +28,18 @@ const transitionDirectivesToCopyOnIsland = Object.freeze([
 	'data-astro-transition-persist',
 ]);
 
+export function withoutTransitionAttributes(props: Props): Props {
+	const res: Props = Object.fromEntries(Object.entries(props).filter(([key]) => !transitionDirectivesToCopyOnIsland.includes(key)));
+	for (const sym of Object.getOwnPropertySymbols(props)) {
+		res[sym] = props[sym];
+	}
+	return res;
+}
+
 // Used to extract the directives, aka `client:load` information about a component.
 // Finds these special props and removes them from what gets passed into the component.
 export function extractDirectives(
-	inputProps: Record<string | number | symbol, any>,
+	inputProps: Props,
 	clientDirectives: SSRResult['clientDirectives']
 ): ExtractedProps {
 	let extracted: ExtractedProps = {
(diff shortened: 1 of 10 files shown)


packages/astro/src/runtime/server/render/component.ts
```diff
@@ -93,6 +93,7 @@ async function renderFrameworkComponent(
 	};
 
 	const { hydration, isPage, props } = extractDirectives(_props, clientDirectives);
+	const propsWithoutTransitionAttributes = withoutTransitionAttributes(props);
```
bluwy, line 96: Could we move the transition attributes filtering inside `extractDirectives` instead? That makes it a little and tucks in the transition attribute code in one file.

packages/astro/src/runtime/server/hydration.ts
```diff
@@ -15,21 +15,31 @@ export interface HydrationMetadata {
...
 	'data-astro-transition-persist',
 ]);
 
+export function withoutTransitionAttributes(props: Props): Props {
+	const res: Props = Object.fromEntries(Object.entries(props).filter(([key]) => !transitionDirectivesToCopyOnIsland.includes(key)));
+	for (const sym of Object.getOwnPropertySymbols(props)) {
+		res[sym] = props[sym];
+	}
+	return res;
+}
```
bluwy, line 37: If we move this instead `extractDirectives`, I think we can do the check at here directly? https://github.com/withastro/astro/blob/eaed844ea8f2f52e0c9caa40bb3ec7377e10595f/packages/astro/src/runtime/server/hydration.ts#L97-L99

Using `fromEntries` and `entries` has a higher perf cost.

martrapp: Yes.

packages/astro/src/runtime/server/render/component.ts
bluwy: Yeah I think the `extractDirectives` naming is fine. If we want to change it, something like `extractPropStates` would also be fine by me

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -92,7 +92,7 @@ async function renderFrameworkComponent(
 		displayName,
 	};
 
-	const { hydration, isPage, props } = extractDirectives(_props, clientDirectives);
+	const { hydration, isPage, props, propsWithoutTransitionAttributes } = extractDirectives(_props, clientDirectives);
```
bluwy, line 95: Oh I actually thought `propsWithoutTransitionAttributes` would replace `props`, but I think this is fine too. The props extraction are now all in one spot 👍

bluwy (approved):

PR #8779, opened 2023-10-09
docs: better explanation for `base` configuration 

ematipico:
## Changes

The documentation wasn't extensive enough IMHO. I didn't understand that `config.base` is also manipulated by Astro accordingly, this is important because `base` is read by integrations too.

I also added some practical examples.

## Testing

N/A

## Docs

### DIFF
--- a/.changeset/thin-flies-notice.md
+++ b/.changeset/thin-flies-notice.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Enrich the explanation of the `base` configuration with examples.
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -580,7 +580,28 @@ export interface AstroUserConfig {
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
 	 *
-	 * The value of `import.meta.env.BASE_URL` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
+	 * The value of `import.meta.env.BASE_URL` and configuration value `config.base` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
+	 *
+	 * This means with a configuration like this:
+	 * ```js
+	 * {
+	 * 	 base: '/docs/',
+	 * 	 trailingSlash: "never"
+	 * }
+	 * ```
+	 *
+	 * `import.meta.env.BASE_URL` and `config.base` will be `/docs`.
+	 *
+	 * In a configuration like this:
+	 *
+	 * ```js
+	 * {
+	 * 	 base: '/docs',
+	 * 	 trailingSlash: "always"
+	 * }
+	 * ```
+	 *
+	 * `import.meta.env.BASE_URL` and `config.base` will be `/docs/`.
 	 */
 	base?: string;
 


bluwy (commented):
I don't think we need a changeset for this change, but otherwise lgtm! Would also be good to have a pass from docs before merging too.

packages/astro/src/@types/astro.ts
```diff
@@ -580,7 +580,28 @@ export interface AstroUserConfig {
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
 	 *
-	 * The value of `import.meta.env.BASE_URL` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
+	 * The value of `import.meta.env.BASE_URL` and configuration value `config.base` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
```
sarah11918, line 583:
```suggestion
	 * The value of `import.meta.env.BASE_URL` will be determined by your `trailingSlash` config, no matter what value you have set for `base`.
	 *
	 * A trailing slash is always included if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will include a trailing slash, even if `base` includes one.
	* 
	* Additionally, Astro will internally manipulate the configured value of `config.base` before making it available to integrations. The value of `config.base` as read by integrations will also be determined by your `trailingSlash` configuration in the same way.
```
OK, it seemed weird to me to say the value of "config.base" is *changed* (because if someone looks in their Astro config, the value is not changed. If they are asked, "What value do you have for `base`? They are going to go look in their config file and see/say exactly what they entered.)

After speaking with Chris, we think there needs to be documentation if Astro is literally changing a value on people internally, but still calling it by the same name. That sounds... risky and (as you can see) leads to confusion.  😅 

But, my job is not to judge how Astro works and only to worry about explaining how it does! So, if I have this kind of right, something *like* the above would be my recommendation:

- Don't just lump the two of them together in one explanation. One is an environment variable produced by checking two different config values. The other *is* a configuration variable set by the user that you are changing internally before using it again with the same name.

packages/astro/src/@types/astro.ts
```diff
@@ -580,7 +580,28 @@ export interface AstroUserConfig {
...
+	 * In a configuration like this:
+	 *
+	 * ```js
+	 * {
+	 * 	 base: '/docs',
+	 * 	 trailingSlash: "always"
+	 * }
+	 * ```
+	 *
+	 * `import.meta.env.BASE_URL` and `config.base` will be `/docs/`.
```
sarah11918, line 604:
``````suggestion
	 	 * In the example below, the values of `import.meta.env.BASE_URL` and `config.base` when processed will both be `/docs`:
	 * ```js
	 * {
	 * 	 base: '/docs/',
	 * 	 trailingSlash: "never"
	 * }
	 * ```
	 *
	 * In the example below, the values of `import.meta.env.BASE_URL` and `config.base` when processed will both be `/docs/`:
	 *
	 * ```js
	 * {
	 * 	 base: '/docs',
	 * 	 trailingSlash: "always"
	 * }
	 * ```
``````
- Following our pattern of explaining what the code should demonstrate before the code sample
- Adding an extra nudge towards "stuff has happened to the `config.base` value as you have set it"

**Comment by delucis** [comment]
When does Astro manipulate `config.base`? And can we avoid doing so? It seems bad practice for us to mutate user config like that, but if it is required, these docs should probably say WHEN that happens. I’m guessing in `astro:config:setup` it hasn’t happened yet? Because then it would have happened before another integration has the chance to set `base` itself. So now we potentially have a pretty painful docs problem of “`config.base` can be different depending on when you read it”.

**Review by sarah11918 — COMMENTED** [review]
For the sake of a quick, helpful PR to docs that is an improvement, I've made some suggestions below that I think can work in the short term.

But after talking with @delucis, I do think we need a more extensive look at this to consider this behaviour fully documented!

File: packages/astro/src/@types/astro.ts [hunk 3]
```diff
@@ -580,7 +580,28 @@ export interface AstroUserConfig {
 	 *
 	 * When using this option, all of your static asset imports and URLs should add the base as a prefix. You can access this value via `import.meta.env.BASE_URL`.
 	 *
-	 * The value of `import.meta.env.BASE_URL` respects your `trailingSlash` config and will include a trailing slash if you explicitly include one or if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
+	 * The value of `import.meta.env.BASE_URL` will be determined by your `trailingSlash` config, no matter what value you have set for `base`.
+	 *
+	 * A trailing slash is always included if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will include a trailing slash, even if `base` includes one.
```
sarah11918, line 585:
```suggestion
	 * A trailing slash is always included if `trailingSlash: "always"` is set. If `trailingSlash: "never"` is set, `BASE_URL` will not include a trailing slash, even if `base` includes one.
```
oops  😅

.changeset/thin-flies-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Enrich the explanation of the `base` configuration with examples.
```
sarah11918, line 5:
```suggestion
Enriches the explanation of the `base` configuration with examples.
```

sarah11918 (approved):
One missing word commented in below, but then Docs is happy with this PR! Would be nice to more fully understand exactly what's happening for documenting more fully later, but I think this is a helpful update as is!

PR #8782, opened 2023-10-09
fix(cloudflare): runtime types for Cloudflare caches

helloimalastair:
## Changes

- Add `CacheStorage` type from [@cloudflare/workers-types](https://www.npmjs.com/package/@cloudflare/workers-types), to override default `dom.d.ts` cache types, which do not match.

## Testing

Typechange only, types compile correctly.

## Docs

This update clarifies types for Runtime Cache, which afaict isn't documented.

### DIFF
--- a/.changeset/unlucky-avocados-brake.md
+++ b/.changeset/unlucky-avocados-brake.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-add cloudflare cachestorage types to server entrypoints
+fixes `AdvancedRuntime` & `DirectoryRuntime` types to work woth Cloudflare caches
--- a/.changeset/unlucky-avocados-brake.md
+++ b/.changeset/unlucky-avocados-brake.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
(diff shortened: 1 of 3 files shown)


lilnasy:
I think this would affect client-side types as well. It's fine to intentionally do it as a user, but idk if the adapter should do it across the board. Users might find that surprising. I'll let @alexanderniebuhr decide.

You should know the cloudflare adapter is in the middle of a migration to https://github.com/withastro/adapters. Hopefully, it will be done by the end of the week. This PR would probably be merged in that repo.

alexanderniebuhr:
Oh yeah this was actually what I wanted to ask, maybe I did not make myself clear. Why wouldn't something like this work?

```
export interface DirectoryRuntime<T extends object = object> {
	runtime: {
		waitUntil: (promise: Promise<any>) => void;
		env: EventContext<unknown, string, unknown>['env'] & T;
		cf: CFRequest['cf'];
		caches: CacheStorage;
	};
}
```

helloimalastair:
Also, does this repo qualify for Hacktoberfest? No issues if it doesn't, just curious.

alexanderniebuhr (approved):

Issue #8784, opened 2023-10-09
Cloudflare adapter update for node:crypto


jadbox:
### Astro Info

Cloudflare now has initial support for `node:crypto` package. See:
https://developers.cloudflare.com/workers/platform/changelog/#2023-04-28

### If this issue only occurs in one browser, which browser is a problem?

all

### Describe the Bug

The package node:crypto should be allowlisted in the Cloudflare adapter excludes list. Note, it may be preferable to allowlist all of node:* packages as the module support changes very frequently, and each time the adapter requires an update and waiting period on a new release.

### What's the expected result?

Build fail for node:crypto but it's allowed now by wrangler

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ig3cma?file=src%2Fpages%2Findex.astro

### Participation

- [x] I am willing to submit a pull request for this issue.

jadbox:
PR for fix with additional future proofing for future upstream issues: https://github.com/withastro/astro/pull/8787

alexanderniebuhr:
@jadbox thank you for your contribution! Just a note, `node:crypto` is only supported since 14th September, while it is some time, it is not as long as you linked.

> Note, it may be preferable to allowlist all of node:* packages as the module support changes very frequently, and each time the adapter requires an update and waiting period on a new release.

I'm still against this. I laid out multiple times before, why I don't think it is a good idea to allowlist the `node:*` wildcard. I think it is more likely that we would go for the settings override solution, than for the wildcard.

PR #8787, opened 2023-10-09
feature: add config options to miniflare and esbuild external

jadbox:
## Changes

- node:crypto support added (fixes https://github.com/withastro/astro/issues/8784)
- Added two new configuration options (esbuild.external and miniflare) to allow clients additional esbuild externals or additional Miniflare constructor parameters.

## Testing

- node:crypto can be imported as its been added to the allowlist
- additional packages can be ignored during build time by using esbuild.external config option
- additional parameters can be used to Miniflare using the new k/v parameter miniflare

## Docs

- Docs are added on the parameters

### DIFF
--- a/packages/integrations/cloudflare/src/index.ts
+++ b/packages/integrations/cloudflare/src/index.ts
@@ -49,6 +49,16 @@ type Options = {
 	 */
 	runtime?: 'off' | 'local' | 'remote';
 	wasmModuleImports?: boolean;
+	/**
+	* Additional externals to add to esbuild. Note that the external property adds to the existing list.
+	*/
+	esbuild?: {
+		external?: string[];
+	}
+	/**
+	* Additional config options to provide to the miniflare initialization object. Note that the options property adds to the existing list.
+	*/
+	miniflare?: object;
 };
 
 interface BuildConfig {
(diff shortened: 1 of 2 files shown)


alexanderniebuhr:
@jadbox thank you for your contribution.

I'll review the technical aspect later today. Please also add documentation _(which are mandatory for a change like this)_, These should guidance people on how to use the new feature and these should be clear that this is only an advanced use-case and debugging might get harder. Users should be aware of all potential downsides, if choosing to opt-in for this.

However, I don't expect that we can offer the same level of help/support in Discord or on GitHub if users use this.

packages/integrations/cloudflare/src/index.ts
```diff
@@ -49,6 +49,16 @@ type Options = {
...
+	/**
+	* Additional externals to add to esbuild. Note that the external property adds to the existing list.
+	*/
+	esbuild?: {
+		external?: string[];
+	}
+	/**
+	* Additional config options to provide to the miniflare initialization object. Note that the options property adds to the existing list.
+	*/
+	miniflare?: object;
```
alexanderniebuhr, line 61: Could you give me an example of what to add here?

alexanderniebuhr (changes requested):
I'll ported the `node:crypto` change over to another PR, so we can get that merged asap!

For the other change, it would be great some more opinions in the discussions. So I'm blocking this for some days. If we don't get any more opinions or are fine with exposing the settings. I'll approve this, as it is technically correct! Just the documentation is still missing, as I said before.

alexanderniebuhr:
The adapter is moved to a different repo, and the solution is currently discussed across maintainers. Once we have an agreement I will implement the decided solution. Going to close this for now. Still thank you for the contribution.

PR #8791, opened 2023-10-10
feat(i18n): fallback via middleware

ematipico:
## Changes

This PR rewrite the logic of how the fallback system works in the dev server.

First, the fallback system will be handled by a middleware function. This function will be used in all modes (dev, ssg and ssr). 

The fallback system is an opt-in feature, and this PR makes use of the middleware in dev mode.

Here's the details of the PR:
- at the moment, Astro **can't** handle routes that aren't in the file system. This is a design constraint that will soon change, but it would require some core changes that will be appropriately handled in https://github.com/withastro/roadmap/discussions/681 
- as a workaround to the previous bullet point, I introduced a new `RouteType` called `fallback`. This is a temporary name, and I hope to change it once `Astro.rewrite`/`Astro.render` will land. I updated the logic of `handleRoute` to create a "fake" `RouteData`. This fake type is only used to reach the rendering phase, and Astro eventually calls the middleware. I am not fond of the solution very much, to be honest, but I hope that we will soon settle on a good way to render routes that don't exist in the file system.
- `ComponentInstance` is now `undefined`, because we should be able to bypass the limitation of the file system. I added some errors for the time being.
- The rest of the changes are just a mix of hoisted variables, assignments and `if/else` to keep the exist logic as is, and limit this new logic only for the `i18n` feature.
- I removed some code that I initially created to create "matched routes" on demand.
- I added a new configuration called `fallbackControl`, that tells Astro what it should do when it sees an i18n route that doesn't have a file on file system. At the moment, we can only support the `"redirect"` variant. When `Astro.render`/`Astro.rewrite` will land, we will be able to support the `"render"` variant too.

## Testing
- Updated existing tests to match the new business logic
- I added some unit tests for the configuration validation.
- 

## Docs

None for now


packages/astro/src/@types/astro.ts
```diff
@@ -1379,6 +1379,21 @@ export interface AstroUserConfig {
...
 			fallback?: Record<string, string[]>;
 
+			/**
+			 * @docs
+			 * @name experimental.i18n.fallbackControl
+			 * @type {"none" | "render" | "redirect"}
+			 * @version 3.*.*
+			 * @description
+			 *
+			 * Controls the fallback system of the internalisation:
```
bluwy, line 1389: FYI it's "internationalisation". Saw a couple of these 😄

bluwy (approved):
Logic looks good 👍

PR #8794, opened 2023-10-10
Improve Vue `appEntrypoint` handling

yoyo837:
## Changes

Improve robustness and compatibility for some case.

When vue's appEntrypoint is configured and it does not return the default function, it will throw an error:

astro.config.mjs:
```js
  integrations: [
    vue({
      appEntrypoint: '/src/pages/_app',
      template: {
        compilerOptions: {
          //
        },
      },
    }),
  ],
```

src/app.js:
```js
// Comment it out, so no export default function
// export default async app => {
//   await global(app);
// };
```

An error will be thrown after running:

image
 

## Testing

Check if a function is available before calling it.


.changeset/smart-cameras-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vue': major
```
natemoo-re, line 2:
```suggestion
'@astrojs/vue': patch
```

.changeset/smart-cameras-kneel.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+chore: improve app setup call
```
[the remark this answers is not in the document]
yoyo837: Yes.

`astro.config.mjs`:
```js
  integrations: [
    vue({
      appEntrypoint: '/src/pages/_app',
      template: {
        compilerOptions: {
          //
        },
      },
    }),
  ],
```

In the `/src/pages/_app.js` file, if there is no export defualt, it may cause errors.

ematipico (changes requested):
Blocking, for the following reasons:
- the template is empty and doesn't explain anything about the PR
- we don't know what we are fixing, there isn't a test
- there's no test nor explanation about this fix being tested

yoyo837:
> Blocking, for the following reasons:
> 
> * the template is empty and doesn't explain anything about the PR
> * we don't know what we are fixing, there isn't a test
> * there's no test nor explanation about this fix being tested

There is my first astro PR. Is there any documentation on how to create a standard PR?

ematipico:
> > Blocking, for the following reasons:
> > 
> > * the template is empty and doesn't explain anything about the PR
> > * we don't know what we are fixing, there isn't a test
> > * there's no test nor explanation about this fix being tested
> 
> There is my first astro PR. Is there any documentation on how to create a standard PR?

Edit the description of the PR. Under `Changes` explain what this change is about, you gave a great explanation to Nate before, you can do the same in the PR. Explain the bug and you fixed that.

Under the `Testing` explain how you tested your fix. Ideally, you should create a new test inside the integration. 

No need for documentation, so you can leave that part empty.

yoyo837:
Thanks. I'll try it later.

ematipico:
Hi @yoyo837 ,  are you still interested in this PR?

yoyo837:
> Hi @yoyo837 , are you still interested in this PR?

Yes, I still want to finish this PR, but I also wouldn't mind opening a new PR when I actually do it.

yoyo837:
When I add the test case, it seems that rollup does not allow no export default declaration.

ematipico:
I think the error is correct. Maybe what you want to do is to use `import * as virtualApp` instead

yoyo837:
> I think the error is correct. Maybe what you want to do is to use `import * as virtualApp` instead

Need help.

ematipico:
I am not very familiar with the integration, but that is failing because in case of `appEntrypoint`, we don't export a `default` export

https://github.com/withastro/astro/blob/35cd810f0f988010fbb8e6d7ab205de5d816e2b2/packages/integrations/vue/src/index.ts#L46-L50

packages/integrations/vue/src/index.ts
```diff
@@ -42,18 +51,42 @@ function virtualAppEntrypoint(options?: Options) {
 				return resolvedVirtualModuleId;
 			}
 		},
-		load(id: string) {
+		async load(id: string) {
 			if (id === resolvedVirtualModuleId) {
-				if (options?.appEntrypoint) {
-					return `export { default as setup } from "${options.appEntrypoint}";`;
+				if (options.appEntrypoint) {
```
florian-lefebvre, line 56: Feel free to suggest other variables names, I was running out of ideas 😅

.changeset/smart-cameras-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/vue': patch
+---
+
+Prevents Astro from crashing when no default function is exported from the `appEntrypoint`.
```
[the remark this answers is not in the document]
florian-lefebvre: Exactly, I didn't know how precise the changeset needed to be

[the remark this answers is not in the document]
lilnasy:
```suggestion
Prevents Astro from crashing when no default function is exported from the `appEntrypoint`. Now, the entrypoint will be ignored with a warning instead.
```

packages/integrations/vue/src/index.ts
```diff
@@ -42,18 +49,39 @@ function virtualAppEntrypoint(options?: Options) {
...
+					try {
+						let resolved;
+						if (options.appEntrypoint.startsWith('.')) {
+							resolved = await this.resolve(fileURLToPath(new URL(options.appEntrypoint, options.root)));
+						} else {
+							resolved = await this.resolve(options.appEntrypoint, fileURLToPath(options.root));
+						}
```
natemoo-re, line 62: This is the main logic change. Now we're resolving the `appEntrypoint` with Rollup (Vite) directly so we'll avoid many edge cases like virtual files and aliases.

natemoo-re: Note that relative `appEntrypoint` is now handled correctly by resolving it relative to the project root first.

packages/integrations/vue/src/index.ts
```diff
@@ -42,18 +49,39 @@ function virtualAppEntrypoint(options?: Options) {
...
+						if (!resolved) {
+							throw new Error();
+						}
+						const loaded = await this.load(resolved);
+						if (!loaded.hasDefaultExport) {
```
natemoo-re, line 67: Rollup already has an API for loading a file, and the returned `ModuleInfo` has a handy `hasDefaultExport` property. This avoids us trying to load and parse the file ourselves, again avoiding a bunch of potential edge cases.

packages/integrations/vue/src/index.ts
```diff
@@ -42,18 +49,39 @@ function virtualAppEntrypoint(options?: Options) {
...
+							options.logger.warn(
+								`appEntrypoint \`${options.appEntrypoint}\` does not export a default function. Check out https://docs.astro.build/en/guides/integrations-guide/vue/#appentrypoint.`
+							);
+							return noop;
+						}
+						return `export { default as setup } from "${resolved.id}";`;
```
natemoo-re, line 73: Simplified the export statement to a single line. Now, we know that there is a `default` export so this is safe.

natemoo-re (approved):
Thanks for picking this up again @florian-lefebvre! Just pushed some changes but this LGTM.

packages/integrations/vue/src/index.ts
```diff
@@ -42,18 +49,39 @@ function virtualAppEntrypoint(options?: Options) {
...
+				if (options.appEntrypoint) {
+					try {
+						let resolved;
+						if (options.appEntrypoint.startsWith('.')) {
+							resolved = await this.resolve(fileURLToPath(new URL(options.appEntrypoint, options.root)));
+						} else {
+							resolved = await this.resolve(options.appEntrypoint, fileURLToPath(options.root));
+						}
+						if (!resolved) {
+							throw new Error();
```
florian-lefebvre, line 64: Should we pass a message here?

natemoo-re:
```suggestion
							// This error is handled below, the message isn't shown to the user
							throw new Error('Unable to resolve appEntrypoint');
```

yoyo837:
Thank you everyone for making this PR landed.

PR #8795, opened 2023-10-10
Fix markdown encoding as utf-8 by default

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/8676

Revert to Astro 2 behaviour where markdown pages are `Content-Type` `type/html; charset=utf-8` by default. It was an unintentional breaking change.

We can revisit this again if we want to remove this default and add a warning.

## Testing

Added new test. Testing encoding is a little tricky so I checked them via the headers and html `<meta` tag instead.

## Docs

n/a. bug fix.


packages/astro/src/runtime/server/render/page.ts
```diff
@@ -63,6 +63,11 @@ export async function renderPage(
 		body = encoder.encode(body);
 		headers.set('Content-Length', body.byteLength.toString());
 	}
+	// TODO: Revisit if user should manually set charset by themselves in Astro 4
+	// This code preserves the existing behaviour for markdown pages since Astro 2
+	if (route?.component.endsWith('.md')) {
+		headers.set('Content-Type', 'text/html; charset=utf-8');
+	}
```
natemoo-re, line 70: Makes sense! Should we maybe handle _any_ non-HTML file here? I could see Markdoc running into this problem as well.

natemoo-re (approved):
Not a blocking suggestion, but wondering if we should be more defensive here.

Approved either way!

packages/astro/src/runtime/server/render/page.ts
bluwy: I'm not sure if doing for all non-HTML would be safe. In most cases, they should hit the `if (!isAstroComponentFactory(componentFactory)) {` check above and already apply utf-8. It's a special case here that markdown pages runtime looks like an Astro component.

If Markdoc does also hit the issue we could add another `.endsWith('.mdoc')` check though. (have to check this)

natemoo-re: Ah that is fair! Consider this resolved then.

Issue #6918, opened 2023-04-27
Unable to support TS 5.0 tsconfig extends array


unional:
### What version of `astro` are you using?

2.3.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

`tsconfig-resolver` does not support `extends` with array, which is a feature in TypeScript 5.0

```ts
// tsconfig.json
{
  "extends: ["astro/tsconfigs/strict", "another-config"]
}
```

It errors with:

```sh
The "path" argument must be of type string. Received an instance of Array
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-3dbiun?file=tsconfig.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Not sure if there's anything we can do since it needs to be supported by https://github.com/ifiokjr/tsconfig-resolver, which doesn't seem to have recent activity. Another option is to use https://github.com/dominikg/tsconfck or https://github.com/aleclarson/vite-tsconfig-paths

Princesseuh:
We should probably migrate to `tsconfck` anyway, `tsconfig-resolver` is unmaintained and has some issues we needed workaround for in the past

Princesseuh:
Blocked by https://github.com/dominikg/tsconfck/issues/87, [I started working on a PR for it in this fork](https://github.com/Princesseuh/tsconfck), but since this is kinda low priority for us, I'll circle back to it later unless someone is interested in tackling it.

MichaelDeBoey:
@Princesseuh I would be interested to get your `jsconfig.json` support merged into `tsfonck` for the reasons mentioned in https://github.com/dominikg/tsconfck/issues/87 👀

firxworx:
After having wasted a chunk of time on an obtuse error where I could get Nx + Astro 3 to build on my workstation and not with Dockerfile and finally tracing it to this... My issue: I extended multiple tsconfigs (!!)

I haven't checked but it sort of smells to me like there might be a bit of "find an npm package for anything and everything" at play with that one where dependencies get created on a single maintainer packages that appear abandoned to save a few lines of code.

[...]

PR #8798, opened 2023-10-10
feat: use tsconfck instead of tsconfig-resolver

Princesseuh:
## Changes

`tsconfig-resolver` is unmaintained and doesn't support all of newer TypeScript's features. `tsconfck` is also in use by many reputable project and should be a safer dep to have.

Fix https://github.com/withastro/astro/issues/6918

## Testing

Tests should pass!

## Docs

N/A


packages/astro/src/cli/add/index.ts
```diff
@@ -848,25 +848,31 @@ async function updateTSConfig(
...
+		inputConfig = {
+			tsconfig: defaultTSConfig,
+			tsconfigFile: path.join(cwd, 'tsconfig.json'),
+			rawConfig: { tsconfig: defaultTSConfig, tsconfigFile: path.join(cwd, 'tsconfig.json') },
+		};
+	} else {
+		inputConfigText = JSON.stringify(inputConfig.rawConfig.tsconfig, null, 2);
 	}
 
+	const configFileName = inputConfig.tsconfigFile.split('/').pop();
```
bluwy, line 867:
```suggestion
	const configFileName = inputConfig.tsconfigFile.split(path.sep).pop();
```

Looks like the bug existed before, but we might need to use proper OS separator, or use `path.basename` directly.

bluwy (approved):

natemoo-re (approved):
LGTM. Thanks for pinning the exact pre-release version. We'll need to remember to keep an eye on that when it gets bumped to `latest`.

firxworx:
Looks good this is a nice contribution @Princesseuh :)
If the motivation is to support ts5 extends format / multiple extends do you think there should be a test / fixture that covers that case? Maybe I missed when I skimmed down files changed but I didn't catch. Thanks again!

Princesseuh:
> Looks good this is a nice contribution @Princesseuh :)
> If the motivation is to support ts5 extends format / multiple extends do you think there should be a test / fixture that covers that case? Maybe I missed when I skimmed down files changed but I didn't catch. Thanks again!

We could for sure, but we'd rather avoid testing every feature like this that are ultimately outside of our code. We trust `tsconfck` to do what it says, especially considering it's used upstream from us (Vite uses it)

PR #7597, opened 2023-07-07
Add "allowHTML" option for Markdoc with HTML parsing/processing

alex-sherwin:
## Changes

* Adds an *allowHTML* option for the @astrojs/markdoc integration which enables HTML tokenization/rendering in Markdoc markup

## Testing

With extensive unit tests on parsed HTML DOM elements

## Docs

Roadmap for this discussion: https://github.com/withastro/roadmap/discussions/613

PR #8802, opened 2023-10-11
feat(markdoc): allowIndentation integration option

AndyClifford:
## Changes

- Added allowIndentation to the markdoc integration options which allows indentation of markdoc tags for better readability.
- Updated README to reflect changes

There is an existing `allowHTML` integration option which also enables allowIndentation so this change does not introduce new functionality. This was added in https://github.com/withastro/astro/pull/7597

It's nice to enable better readability without also enabling raw HTML in markdoc files. I think allowIndentation should be enabled by default once it comes out of it's experimental phase.

## Testing

- Added unit tests similar to those that test the `allowHTML` integration option.
- Tested with a local astro project using the markdoc integration. No issues encountered. 

## Docs

This change requires an update to the docs as per the changes in the README. Looks like there is some automation in place to do this.

### DIFF
--- a/packages/integrations/markdoc/README.md
+++ b/packages/integrations/markdoc/README.md
@@ -458,6 +458,40 @@ To achieve a more Markdown-like experience, where HTML elements can be included 
 > **Warning**
 > When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible. Ensure that any HTML markup comes from trusted sources.
 
+### `allowIndentation`
+
+By default, any content that is indented by four spaces is treated as a code block. Unfortunately, this behavior makes it difficult to use arbitrary levels of indentation to improve the readability of documents with complex structure.
+
+When using nested tags in Markdoc, it can be helpful to indent the content inside of tags so that the level of depth is clear. To support arbitrary indentation, we have to disable the indent-based code blocks and modify several other markdown-it parsing rules that account for indent-based code blocks. These changes can be applied by enabling the allowIndentation option.
+
+```diff lang="js" "allowIndentation: true"
+  // astro.config.mjs
+  import { defineConfig } from 'astro/config';
+  import markdoc from '@astrojs/markdoc';
+
+  export default defineConfig({
+    // ...
++   integrations: [markdoc({ allowIndentation: true })],
+    //                       ^^^^^^^^^^^^^^^^^^^^^^
+  });
+```
+
+```md
+# Welcome to Markdoc with indented tags 👋
+# Note: Can use either spaces or tabs for indentation
+
+{% custom-tag %}
+  {% custom-tag %}
+    ### Tags can be indented for better readability
+
+    {% another-custom-tag %}
+      This is easier to follow when there is a lot of nesting
+    {% /another-custom-tag %}
+
+  {% /custom-tag %}
+{% /custom-tag %}
+```
+
 ## Examples
 
 - The [Astro Markdoc starter template](https://github.com/withastro/astro/tree/latest/examples/with-markdoc) shows how to use Markdoc files in your Astro project.
--- a/packages/integrations/markdoc/README.md
+++ b/packages/integrations/markdoc/README.md
@@ -458,21 +458,21 @@ To achieve a more Markdown-like experience, where HTML elements can be included 
 > **Warning**
 > When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible. Ensure that any HTML markup comes from trusted sources.
 
-### `allowIndentation`
+### `ignoreIndentation`
 
 By default, any content that is indented by four spaces is treated as a code block. Unfortunately, this behavior makes it difficult to use arbitrary levels of indentation to improve the readability of documents with complex structure.
 
-When using nested tags in Markdoc, it can be helpful to indent the content inside of tags so that the level of depth is clear. To support arbitrary indentation, we have to disable the indent-based code blocks and modify several other markdown-it parsing rules that account for indent-based code blocks. These changes can be applied by enabling the allowIndentation option.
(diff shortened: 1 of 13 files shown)


packages/integrations/markdoc/README.md
```diff
@@ -458,6 +458,39 @@ To achieve a more Markdown-like experience, where HTML elements can be included
 > **Warning**
 > When `allowHTML` is enabled, HTML markup inside Markdoc documents will be rendered as actual HTML elements (including `<script>`), making attack vectors like XSS possible. Ensure that any HTML markup comes from trusted sources.
 
+### `allowIndentation`
+
+Enables indentation of Markdoc tags for better readability.
+
+The allowIndentation option is experimental and is not enabled by default. When the option is enabled, it is not possible to use indent-based code blocks.
```
lilnasy, line 465: I don't think a call out saying it's experimental is necessary. Both this integration and markdoc itself are pre-1.0, and we are just exposing one of its config options that doesn't seem to be any more experimental than the other features.

lilnasy, line 465: This could use an example for why you may want to configure this option. Is indentation usually not allowed, or is ignored? It isn't documented by markdoc.

lilnasy (commented):
Thanks, this looks promising!

And the PR is thorough as well with test and docs.

packages/integrations/markdoc/test/fixtures/render-with-indented-components/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "@test/markdoc-render-with-indented-components",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
+    "@astrojs/markdoc": "workspace:*",
+    "astro": "workspace:*"
+  },
+  "devDependencies": {
+    "shiki": "^0.14.3"
+  }
```
lilnasy, line 11: Is this intentional?

sarah11918:
Just commenting in here for notifications! Normally I'd review the docs after I know the feature will be accepted, but I haven't yet seen indication from platform here.  Happy to look after the feature has been triaged/reviewed!

packages/integrations/markdoc/README.md
AndyClifford: I found some documentation https://markdoc.dev/docs/faq#indentation - we could copy what's in there or just link to it?

AndyClifford: I found some documentation https://markdoc.dev/docs/faq#indentation - I've updated the README to reflect this

packages/integrations/markdoc/test/render.test.js
```diff
@@ -125,6 +147,26 @@ function renderComponentsChecks(html) {
...
+	const h2 = document.querySelector('h2');
+	expect(h2.textContent).to.equal('Post with indented components');
+
+	// Renders custom shortcode components
+	const marquees = document.querySelectorAll('marquee');
+	expect(marquees.length).to.equal(2);
+
+	// Renders indented h3
+	const h3 = document.querySelector('h3');
+	expect(h3.textContent).to.equal('I am an h3!');
```
ematipico, line 162: How do you know if this is indented? Should we check for spaces/tabs somehow?

packages/integrations/markdoc/test/fixtures/render-with-indented-components/src/content/blog/with-indented-components.mdoc
```diff
@@ -0,0 +1,24 @@
...
+		I'm an indented marquee!
+
+		### I am an h3!
+	{% /marquee-element %}
+
+	And a nested code block:
+
+	```js
+	const isRenderedWithShiki = true;
+	```
```
ematipico, line 22: This could be a nitpick, but I haven't seen anything about it in the documentation. Does the indentation work with tabs and spaces regardless? If so, I believe we should mention it. For example, this code block has **tabs**, but the documentation has **spaces** (which are forced to use in markdown documents)

ematipico (commented):
I still don't want to approve it because I have some questions that need answers, although I think the feature is very welcome!

packages/integrations/markdoc/test/fixtures/render-with-indented-components/src/content/blog/with-indented-components.mdoc
AndyClifford: I didn't consider tabs v spaces. I have tested this now and it works with either. I've ave added a comment at the top of the code example

packages/integrations/markdoc/test/render.test.js
ematipico: Unfortunately this doesn't answer my question. The comment in the test says "we check if text is indented" , but still I can't understand when a text should not be indented. Sorry for those dumb questions , I have little context about markdoc

AndyClifford: @ematipico No worries ! I can see how these comments are a bit confusing - I've updated them.

The `allowIndentation` option is purely to enable better readability of source code and it will not affect the rendered HTML. Check out Markdoc's official documentation on indentation https://markdoc.dev/docs/faq#indentation

- without `allowIndentation` enabled:
  ```
  {% foo %}
  {% bar %}
  This is content inside of nested tags
  {% /bar %}
  {% /foo %}
  ```

- with `allowIndentation` enabled (easier to read, level of depth is clear)

  ```
  {% foo %}
    {% bar %}
      This is content inside of nested tags
    {% /bar %}
  {% /foo %}
  ```

The rendered HTML is exactly the same in both examples. The second example will fail to render if the `allowIndentation` option is not enabled.

I hope this helps!

lilnasy: Do you think `ignoreIndentation: true` might be better here?

It makes sense for it to be called `allowIndentation` in markdoc's tokenizer options. It might be better to hint at its overall effect with the name of the top-level option.

AndyClifford: Yeah `ignoreIndentation` sounds good !

ematipico (approved):
Thank you @AndyClifford for addressing my questions. Happy to merge!

PR #8813, opened 2023-10-11
Fix: Retain focus for persisted input elements during view transitions

martrapp:
## Changes

An element with `transition:persist` survives view transitions and continues its life on the next page. Its state is preserved. For example, a `HTMLTextAreaElement` retains the text entered up to that point. Unfortunately, persisted elements lose focus during the transition when they are removed from the DOM and reinserted.

This fix saves the current focus and cursor position and restores them on the new page.    

## Testing

Added e2e test

## Docs

n/a bug fix


packages/astro/e2e/fixtures/view-transitions/astro.config.mjs
martrapp, line 1: Not really related to focus handling, but as I was going through the tests I noticed that we are running with `output = server`, when in fact almost all tests are static. So I marked the test that needs the server with `prerender = false` and changed to `hybrid`.

packages/astro/e2e/fixtures/view-transitions/src/components/Layout.astro
martrapp, line 1: The file was always called `styles.css`, but I noticed it only now

packages/astro/e2e/view-transitions.test.js
```diff
@@ -788,7 +788,7 @@ test.describe('View Transitions', () => {
 
 	test('replace history', async ({ page, astro }) => {
 		await page.goto(astro.resolveUrl('/one'));
-		// page six loads the router and automatically uses the router to navigate to page 1
```
martrapp, line 791: copy & paste error from another test

natemoo-re (approved):
This seems great, thanks for the fix and detailed test!

PR #8814, opened 2023-10-11
fix(rerouting): attempt without middleware

### LINKED ISSUE #8797: 500.astro content is not showing when throwing an error from a middleware
### Astro Info

```block
Astro                    v3.2.1
Node                     v20.4.0
Package Manager          pnpm
                         @astrojs/tailwind
```

Chrome

### Describe the Bug

I have a sequence of middleware, the `auth` middleware may throw an error if it is not happy with the user credentials.

```ts
  onRequestInterceptor, // prepare the request and logging
  onRequestAuth, // auth
  onResponseInterceptor // report >400 responses
);
```

when the middleware or the page throws it is showing the normal browser 500 error (image below).

Note: to reporduce, please try the 'preview' command rather than 'dev' cause I noticed different behaviours sometimes.

Screenshot 2023-10-10 at 17 38 17

If the middleware throws, I expect the `content` of `500.astro` to appear on the page, that is the nicely designed page with the site identity. 

A few notes:
- I don't expect redirection to happen. 
- I used to do `redirect()` to `/error`, but I want the user to stay on the same url just with a nice view showing the error.

I also have a few questions: 

- How can I pass some custom messages to my `500.astro` (without using cookies, if possible)?
- If I return a `Response` object from the first middleware, does it stop the other middlewares in the sequence?
- How can I make a **one general page** for all `>400` responses? right now I have `400, 404, 500`.astro but I would like a general `error.astro`

https://stackblitz.com/edit/github-qhuzp1?file=package.json

### Participation

lilnasy:
## Changes
- Fixes #8797 

## Testing
_Pending._

## Docs
Does not affect usage.


natemoo-re (approved):
Makes sense! Thanks for the fix

.changeset/shaggy-onions-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixed an issue where 500.astro did not render when the middleware threw an error.
```
ematipico, line 5:
```suggestion
Fix an issue where `500.astro` did not render when the middleware threw an error.
```

packages/astro/src/core/app/index.ts
```diff
@@ -41,6 +41,7 @@ export interface RenderErrorOptions {
 	routeData?: RouteData;
 	response?: Response;
 	status: 404 | 500;
+	runMiddleware?: boolean;
```
ematipico, line 44: I would keep it not optional, otherwise we have tree values: `undefined`, `true` and `false`

ematipico: Can we document this option and explain what it is?

packages/astro/src/core/app/index.ts
```diff
@@ -280,12 +282,21 @@ export class App {
...
+				}
+				if (!runMiddleware) {
+					this.#pipeline.unsetMiddlewareFunction();
 				}
 				const response = await this.#pipeline.renderRoute(newRenderContext, page);
 				return this.#mergeResponses(response, originalResponse);
-			} catch {}
+			} catch {
+				// Middleware may be the cause of the error, so we try rendering 404/500.astro without it.
+				if (runMiddleware && usedMiddleware) {
```
ematipico, line 296: We can remove `usedMiddleware` and use `mod.onRequest`. Or have the `pipeline` give us this information.

ematipico (commented):
I believe this change should be part of the App `pipeline` because it's in charge of running the middleware, so we should only expose the functions to or not run it.

Also, `App` is used by other adapters, so we should bump all those adapters that use `App`:
- node
- vercel
- cloudflare
- netlify

lilnasy:
I was under the impression that the changes made here were all to internal APIs. I'll take another look later.

.changeset/shaggy-onions-try.md
lilnasy: Is this some convention?

ematipico:
> I was under the impression that the changes made here were all to internal APIs. I'll take another look later.

Unfortunately no. https://docs.astro.build/en/reference/adapter-reference/#astroapp

Search for `import { App } from 'astro/app';`, and you'll find that most of our adapters will inherit this fix.

ematipico:
> but I have made no changes to the public interface of app

No, but I think that all the consumers of this fix will benefit from it, so we should bump a patch for them too.

lilnasy:
In its current state, the fix would apply to adapters and all their versions retroactively; a bump souldn't be necessary. But maybe I am misunderstanding.

Is the idea is to allow the adapter to say whether middleware should run?
```ts
app.renderWithOptions(request, { runMiddleware: false })
```
It would expand the scope of this PR from a bugfix to a feature.

FWIW, it is already planned for the reroute API, and you can see the implementation in the exploration branch.

ematipico:
> In its current state, the fix would apply to adapters and all their versions retroactively; a bump souldn't be necessary. But maybe I am misunderstanding.

Yeah you're right, I misunderstood things. Sorry bikeshedding this

lilnasy:
No worries. I'll get to the suggestions in a few hours.

packages/astro/src/core/app/index.ts
lilnasy: I think there really are 3 options: run middleware, prevent middleware, and default.

ematipico: Great! One more reason to document and explain what the three options do :)

lilnasy: yeah, good point

matthewp (approved):

packages/astro/src/core/app/index.ts
```diff
@@ -41,6 +41,10 @@ export interface RenderErrorOptions {
 	routeData?: RouteData;
 	response?: Response;
 	status: 404 | 500;
+	/**
+	 * Whether to call onRequest() while rendering the error page. Defaults to true.
+	*/
+	runMiddleware?: boolean;
```
ematipico, line 47: nit: since this is the default behaviour is to always run the middleware, users that opt-in probably **don't** want to run the middleware. So maybe we can call it `bypassMiddleware` and by default is `false`

packages/astro/src/core/app/index.ts
```diff
@@ -280,12 +284,21 @@ export class App {
...
+					this.#pipeline.unsetMiddlewareFunction();
+				}
 				const response = await this.#pipeline.renderRoute(newRenderContext, page);
 				return this.#mergeResponses(response, originalResponse);
-			} catch {}
+			} catch {
+				// Middleware may be the cause of the error, so we try rendering 404/500.astro without it.
+				if (runMiddleware && mod.onRequest) {
+					return this.#renderError(request, { status, response: originalResponse, runMiddleware: false });
+				}
```
ematipico, line 300: Uhm, I vaguely remember that in dev, we run the middleware for 404, by I might be wrong. If I am correct, wouldn't this cause misalignment in behaviour?

packages/astro/src/core/pipeline.ts
```diff
@@ -55,6 +55,12 @@ export class Pipeline {
 		this.#onRequest = onRequest;
 	}
 
+	/**
+	 * Ensure that a route is rendered without its corresponding middleware.
```
ematipico, line 59:
```suggestion
	 * Removes the current middleware function. Subsequent requests won't trigger any middleware.
```

ematipico (approved):
Looks good. Left some feedback and question.

packages/astro/src/core/app/index.ts
lilnasy: Makes sense, woudn't take long.

packages/astro/src/core/app/index.ts
lilnasy: Not sure, I'll have to check.

natemoo-re (approved):

packages/astro/src/core/app/index.ts
ematipico: It seems to me that here we don't run the middleware when we render a 404. Did I get it wrong?

lilnasy: If there is a 404.astro, we would. Both during production SSR, and dev.

To summarize the difference introduced by this PR, when the 404.astro or 500.astro rendering throws an error...
- dev: the dev overlay with the error is shown immediately.
- prod: the error page is attempted again without middleware.

PR #8816, opened 2023-10-12
feat(i18n): fallback system in SSG

ematipico:
## Changes

This PR sets the foundation of the i18n fallback system in SSG mode. To achieve that, I had to make some changes to some internal data structures and how the build works under the hood.

As explained before, at the moment, we have a design constraint where Astro can't render routes that don't exist on the file system. We bypassed this constraint in Dev, this PR bypasses this constraint in SSG build too. 

Here's the list of changes I did to achieve that:
- `allPages` is not a flat data structure anymore. Before, `allPages` was a 1:1 relation between the component route and the `PageRouteData`; this is not the case anymore. Now, we have a relation where a component route can have multiple routes, hence the `PageRouteData[]`. This is expected because, in a fallback system, a missing translated page will be "handled" - rendered or redirected to - by another locale route.
- I changed the `createBuildManiest`, which is responsible for creating the routes. In the added logic, we create new routes based on `config.experimental.i18n.fallback`. The newly created routes are a copy of their fallback, but we update their `pathname`, `pattern`, `route` and `type: "fallback"`.
- I had to create a new internal data structure called `pagesByComponents` (plural). I would have updated the current `pagesByComponent`, but that's a data structure sent to the integrations, so I couldn't risk creating a breaking change for that. This new type called `pagesByComponents` is then when we collect the pages to render inside the method `retrievePagesToGenerate`, inside the `buildPipeline.ts`
- Then, we register the i18n middleware, which will take care of the fallback. The previous changes will allow us to reach the middleware, which will take care of generating the redirect (`fallbackControl: "redirect")

## Testing

I moved the tests we had in `dev-routing.test.js` inside a newly created file `i18n-routing.test.js`.

In that file I copied the tests cases I had for DEV and adapted them to SSG. They all should pass.

## Docs

N/A for now


packages/astro/src/core/build/index.ts
```diff
@@ -198,7 +198,9 @@ class AstroBuilder {
 		await runHookBuildDone({
 			config: this.settings.config,
 			pages: pageNames,
-			routes: Object.values(allPages).map((pd) => pd.route),
+			routes: Object.values(allPages)
+				.flat()
+				.map((pd) => pd.route),
```
Princesseuh, line 203: This is totally unrelated, but `pd` is a slur in French, also not sure what it stands for, should probably be a more descriptive name.

packages/astro/src/core/build/internal.ts
```diff
@@ -37,9 +38,16 @@ export interface BuildInternals {
 
 	/**
 	 * A map for page-specific information.
+	 * // TODO: Remove in Astro 4.0
```
Princesseuh, line 41: Any reason to not use those directly now? It's internal, so doesn't really have to follow semver (unless it changes something for users downstream)

bluwy: I can't seem to find the type where integrations can access this. I've definitely changed this breakingly in the past because it should be internal for us 🤔 Do you know where it's exposed?

bluwy (approved):
The new logic make sense to me, but would also like Erika's comments addressed first before merging

Princesseuh (approved):
I'm not super familiar with the logic around these parts, so I can't say much, but what I read through made sense to me!

packages/astro/src/core/build/internal.ts
ematipico: https://github.com/withastro/astro/blob/3468c06f56447ee0aa4115b70f658780e6df75c5/packages/astro/src/core/build/static-build.ts#L237-L243

bluwy: Ah ok so it's `pagesByComponent` only. I'm also fine with deprecating it for now then, and maybe we can construct a better abstraction for Astro 4.0 later.

PR #8817, opened 2023-10-12
Fix error overlay syntax highlighting

bluwy:
## Changes

My bad! The error overlay syntax highlighting was broken. I wasn't using the code replacements to make the it work for the overlay, since it calls shiki/ji directly.

Now:

image

image

## Testing

Manually tested in the examples repo

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/components/shiki.js
+++ b/packages/astro/components/shiki.js
@@ -1,8 +1,7 @@
-import { type Highlighter, getHighlighter } from 'shikiji';
+import { getHighlighter } from 'shikiji';
 
-type HighlighterOptions = NonNullable<Parameters<typeof getHighlighter>[0]>;
-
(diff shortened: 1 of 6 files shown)


packages/astro/components/shiki.js
bluwy, line 1: This file needs to be in JS as it's uncompiled and needs to be imported by `vite.ts`. Should be temporary only for now until I refactor all the duplicated shikiji code in one place.

Princesseuh (approved):
Code seems fine to me, but the build failure is concerning

packages/astro/components/shiki.js
bluwy: Nevermind it's moved inside `src` so it can be TypeScript now 😄

ematipico (approved):

PR #8819, opened 2023-10-12
fix: remote `srcset` images not being resized

itsmatteomanf:
## Changes

Fixed the just released `srcset` support for remote images. They are not getting resized, at all.

Something to think about is to maybe avoid duplicating the base `src` with a potential `srcset` of 1x, even though 1x *can* be omitted per the spec:

https://html.spec.whatwg.org/multipage/images.html#creating-a-source-set-from-attributes (point 4)

## Testing

Fixes current functionality.

### DIFF
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -251,21 +251,24 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 				// If the user passed dimensions, we don't want to add it to the srcset
 				const { width: transformWidth, height: transformHeight, ...rest } = options;
 
-				const srcSetValue = {
-					transform: {
-						...rest,
-					},
+				let srcSetValue = {
+					transform: {},
 					descriptor: `${densityValues[index]}x`,
 					attributes: {
-						type: `image/${targetFormat}`,
-					},
+						type: `image/${targetFormat}`
+					}
 				};
 
 				// Only set width and height if they are different from the original image or if the image is remote
 				// to avoid duplicated final images
 				if (!isESMImportedImage(options.src) || maxTargetWidth !== imageWidth) {
-					srcSetValue.transform.width = maxTargetWidth;
-					srcSetValue.transform.height = Math.round(maxTargetWidth / aspectRatio);
+					srcSetValue.transform = {
+						width: maxTargetWidth,
+						height: Math.round(maxTargetWidth / aspectRatio),
+						...rest
+					};
+				} else {
+					srcSetValue.transform = { ...rest };
 				}
 
 				if (targetFormat !== options.format) {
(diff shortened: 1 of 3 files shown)


ematipico:
@itsmatteomanf 

Would it be possible to add more information about testing, please? We usually have that section to **tell us how** you tested the fix.

It would be great to have a new test case to cover the fix

itsmatteomanf:
@ematipico the issue was that the original image was proxied directly and not resized, when remote.

Basically a code block like this, generated 3 identical `srcset` image files, which is clearly wrong. When checking those they were full size.

```astro
<Image
  width={448}
  height={635}
  densities={[1, 2, 3]}
  src=".../assets/5429f0e2-1b5d-4b5b-960a-63ffa01a41f3"
/>
```

```html
<img
  src="/assets/5429f0e2-1b5d-4b5b-960a-63ffa01a41f3_P7NuC.webp"
  srcset="/assets/5429f0e2-1b5d-4b5b-960a-63ffa01a41f3_Z1th2Ml.webp 1x, /assets/5429f0e2-1b5d-4b5b-960a-63ffa01a41f3_Z1th2Ml.webp 2x, /assets/5429f0e2-1b5d-4b5b-960a-63ffa01a41f3_Z1th2Ml.webp 3x"
  width="448" height="635" loading="lazy" decoding="async">
```

By this PR it makes it so the 3 files are different (while technically a `1x` would be different from the base `src` file, it can be omitted by the user, per spec above) and I locally verified the `transform` options passed to the generate command contained correct sizes.

packages/astro/src/assets/services/service.ts
```diff
@@ -229,7 +229,7 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 
 		const aspectRatio = targetWidth / targetHeight;
 		const imageWidth = isESMImportedImage(options.src) ? options.src.width : options.width;
-		const maxWidth = imageWidth ?? Infinity;
+		const maxWidth = isESMImportedImage(options.src) ? imageWidth ?? Infinity : Infinity;
```
ematipico, line 232: nit: we already check `sESMImportedImage(options.src)` at line 231. We could save the result of this check in a local variable and use it in your new code.

Princesseuh: If you save it in a local variable, the type inference doesn't work

ematipico: TypeScript.....

itsmatteomanf: Yeah, it's done 3 times in that function call... I had it as `isLocal` variable initially, but then had to revert it back to this.

packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -19,5 +19,5 @@ export function hashTransform(transform: ImageTransform, imageService: string) {
 	// Extract the fields we want to hash
 	const { alt, class: className, style, widths, densities, ...rest } = transform;
 	const hashFields = { ...rest, imageService };
-	return shorthash(JSON.stringify(hashFields));
+	return shorthash(JSON.stringify(hashFields, Object.keys(hashFields).sort()));
```
itsmatteomanf, line 22: This solves the issue with duplicated `1x` images and probably reduces some duplication issues altogether.

Princesseuh:
Superseded by https://github.com/withastro/astro/pull/8823, thank you @itsmatteomanf for starting this PR!

PR #8821, opened 2023-10-13
feat: Rework image generation to improve performance

Princesseuh:
## Changes

Change how image generation is done to improve performance:
- Images are now grouped under the base image (or URL) they originate from, as such, we can reuse the same buffer to generate all the variants, whereas before we would reload / refetch the image for every variant
- Adds a concurrent promise queue for image generations. Image generation plays well with being done concurrently, so it's a massive performance increase in a lot of cases, especially for uncached runs.
- Previously we'd do a lot of work to setup image generation on every image. This is now only done once and shared across all calls.

On my website, that has 222 local images, here's the result:

**Before**
- Fresh build: 26s~
- Cached build: 2s~

**After**
- Fresh build: 8s~
- Cached: 2s~

Not exactly scientific, but there's definitely an improvement in there.

## Testing

Tests should still pass!

## Docs

N/A.


Princesseuh:
!preview concurrent-assets

itsmatteomanf:
We need to be able to detect when a file is empty as with multiple writers/readers you can get into a situation in which the generation crashes with a file generated and not written.

This in turn causes the next build to fail with the `ENOENT` error. If the file is empty it should be ignored, imho.

packages/astro/src/core/build/generate.ts
```diff
@@ -1,7 +1,9 @@
 import * as colors from 'kleur/colors';
 import { bgGreen, black, cyan, dim, green, magenta } from 'kleur/colors';
 import fs from 'node:fs';
+import OS from 'node:os';
```
bluwy, line 4: nit: I find it a bit odd that this is uppercase 😄

bluwy (approved):

packages/astro/src/core/build/generate.ts
ematipico: I bet @bluwy won't be sleeping at night 👀

ematipico (approved):

itsmatteomanf:
Do we want to consider a way to set the cpu count manually?

In some instances one might want to reduce the parallelisation if the backend doesn't support it for remote images, or something...

.eslintrc.cjs
```diff
@@ -17,7 +17,12 @@ module.exports = {
     '@typescript-eslint/array-type': ['error', { default: 'array-simple' }],
     '@typescript-eslint/no-unused-vars': [
       'error',
-      { argsIgnorePattern: '^_', ignoreRestSiblings: true },
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+        caughtErrorsIgnorePattern: '^_',
+        ignoreRestSiblings: true,
```
Princesseuh, line 24: This is unrelated to the PR, but it was driving me nuts

Princesseuh:
!preview concurrent-assets

Princesseuh:
@bluwy @ematipico I kinda rewrote it all from scratch 🙈. I updated the PR description to reflect what was done and would love to see a re-review from you both! (I made `os` lower case)

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
 
-	const finalFileURL = new URL('.' + filepath, clientRoot);
-	const finalFolderURL = new URL('./', finalFileURL);
+export async function generateImagesForPath(
+	originalFilePath: string,
+	transforms: MapValue<AssetsGlobalStaticImagesList>,
+	env: AssetEnv,
+	queue: PQueue
+) {
+	const originalImageData = await loadImage(originalFilePath, env);
```
Princesseuh, line 93: In theory, the fact that we load every image at the beginning even when all the transforms might be cached already is slower, especially for fully cached builds. However, in practice I found that this didn't change much, probably that the rest of the improvements counteract this, or maybe hard drives are just fast enough now.

If we want to get it of this, what we could probably do is store a list of cached hashes and check first if all the transforms requested are already cached before loading the image, shouldn't be too hard.

packages/astro/src/core/build/generate.ts
```diff
@@ -196,58 +198,35 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
...
+		const cpuCount = os.cpus().length;
+		const assetsCreationEnvironment = await prepareAssetsGenerationEnv(pipeline, totalCount);
+		const queue = new PQueue({ concurrency: cpuCount });
 
-	logger.info(null, dim(`Completed in ${getTimeStat(timer, performance.now())}.\n`));
-}
+		const assetsTimer = performance.now();
+		for (const [originalPath, transforms] of staticImageList) {
+			await generateImagesForPath(originalPath, transforms, assetsCreationEnvironment, queue);
+		}
```
Princesseuh, line 217: A question someone might have: Would make this part concurrent help?

No (at least as far as I can tell), it doesn't really matter if the queue is full of variants from one image vs variants of multiple images.

Princesseuh:
!preview concurrent-assets

.changeset/good-mirrors-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Add concurrency to asset generation
```
ematipico, line 5: I think it would be great to expand this changeset, by mentioning the benchmarks on your website :)

packages/astro/src/assets/build/generate.ts
ematipico: I think we should apply your suggestions, probably in a follow-up PR. An I/O operation is usually more expensive than a memory operation, and even though Node.js is really great in I/O operations, I think we should try to avoid them if we can.

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
 
-	// For remote images, instead of saving the image directly, we save a JSON file with the image data and expiration date from the server
-	const cacheFile = basename(filepath) + (isLocalImage ? '' : '.json');
-	const cachedFileURL = new URL(cacheFile, assetsCacheDir);
+	async function generateImage(
+		originalImage: ImageData,
+		filepath: string,
+		options: ImageTransform
+	) {
+		let timeStart = performance.now();
```
ematipico, line 106: `const`?

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
+		const generationData = await generateImageInternal(originalImage, filepath, options);
 
-	await fs.promises.mkdir(finalFolderURL, { recursive: true });
+		if (!generationData) {
+			return;
```
ematipico, line 110: Why don't we compute the performance in the case of `undefined`? It makes me think that `generateImageInternal` should be responsible for computing the time.

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
-			await fs.promises.copyFile(cachedFileURL, finalFileURL);
+		const timeEnd = performance.now();
+		const timeChange = getTimeStat(timeStart, timeEnd);
+		const timeIncrease = `(+${timeChange})`;
+		const statsText = generationData.cached
+			? `(reused cache entry)`
+			: `(before: ${generationData.weight.before}kB, after: ${generationData.weight.after}kB)`;
+		const count = `(${env.count.current}/${env.count.total})`;
+		env.logger.info(
+			null,
```
ematipico, line 121: Any specific reason for not adding a label?

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
-			}
-		}
-	} catch (e: any) {
-		if (e.code !== 'ENOENT') {
-			throw new Error(`An error was encountered while reading the cache file. Error: ${e}`);
-		}
-		// If the cache file doesn't exist, just move on, and we'll generate it
-	}
+			} else {
+				const JSONData = JSON.parse(readFileSync(cachedFileURL, 'utf-8')) as RemoteCacheEntry;
```
ematipico, line 148: We should handle the error thrown by `JSON.parse`. You never know if someone tampered with the file.

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
 
-	// If the image is local, we can just read it directly, otherwise we need to download it
-	if (isLocalImage) {
-		imageData = await fs.promises.readFile(
-			new URL(
-				'.' + prependForwardSlash(join(config.build.assets, basename(originalImagePath))),
-				serverRoot
+		const finalFolderURL = new URL('./', finalFileURL);
+		await fs.promises.mkdir(finalFolderURL, { recursive: true });
+
```
ematipico, line 168: If the file is present but expired, shouldn't we clear the folder first?

packages/astro/src/assets/build/generate.ts
```diff
@@ -53,113 +72,176 @@ export async function generateImage(
...
+							data: Buffer.from(resultData.data).toString('base64'),
+							expires: resultData.expires,
+						})
+					);
+				}
 			}
+		} catch (e) {
+			env.logger.warn(
+				'astro:assets',
+				`An error was encountered while creating the cache directory. Proceeding without caching. Error: ${e}`
```
ematipico, line 206:
```suggestion
				`An error was encountered while creating the cache directory. Proceeding without caching.\n\nError: ${e}`
```

packages/astro/src/core/build/generate.ts
```diff
@@ -196,58 +198,35 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
...
-	const logger = pipeline.getLogger();
-	let timeStart = performance.now();
-	const generationData = await generateImageInternal(pipeline, transform, path);
+		await queue.onIdle();
+		const assetsTimeEnd = performance.now();
+		logger.info(null, dim(`Completed in ${getTimeStat(assetsTimer, assetsTimeEnd)}.\n`));
 
-	if (!generationData) {
-		return;
+		delete globalThis?.astroAsset?.addStaticImage;
```
ematipico, line 223:
```suggestion
		globalThis?.astroAsset?.addStaticImage = undefined;
```

ematipico (commented):
Can we add some tests with remote images? Especially with the expiration part. It would be great we could log the user when an image is **not** used anymore in case of it's expired.

packages/astro/src/core/build/generate.ts
Princesseuh: I remember you suggesting exactly this before (this is old code) and I don't remember why, but we couldn't do it for some reason... I'd have to check again

packages/astro/src/assets/build/generate.ts
Princesseuh: There's no labels for all those generating things, it follows the same formatting as page generation etc. This is old code, it was just moved.

packages/astro/src/assets/build/generate.ts
Princesseuh: We do, this code is in a `try catch`.

packages/astro/src/assets/build/generate.ts
Princesseuh: I definitely agree, but I think it's not as bad as it might seems because the file is already written in the same process, so it's presumably already in some sort of hot cache.

packages/astro/src/assets/build/generate.ts
Princesseuh: This function actually can't return `undefined`, the type is wrong.

.changeset/good-mirrors-bake.md
Princesseuh: My tests are not exactly scientific, so I'd rather avoid misleading users. I can expand on what it does, though

packages/astro/src/core/build/generate.ts
ematipico: Ah, probably TS saying that `addStaticImage` can't be `undefined`. That's fine, don't worry, it was just a suggestion :)

Here you can find the reasons: https://biomejs.dev/linter/rules/no-delete/

Princesseuh:
> Can we add some tests with remote images? Especially with the expiration part. It would be great if we could log the user when an image is **not** used anymore in case it's expired.

We have tests with remote images, both cached and uncached ones! We also have tests for errors specific to remote images etc.

packages/astro/src/assets/build/generate.ts
bluwy: Maybe we also make `originalImageData` a function to lazily load it, and those who call in parallel would share the same cache? Vite has this [utility](https://github.com/vitejs/vite/blob/9eec4fb3515c797d18686575c3e644da224657f0/packages/vite/src/node/plugins/css.ts#L1232-L1243) (which could be generalized) to simplify a bit.

But also I think we can do it in a follow-up too if there's not a big perf impact currently.

bluwy (approved):
> (I made `os` lower case)

🎉

itsmatteomanf:
@Princesseuh not sure where the delays are, but the cached builds are definitely slower than before... the un-cached ones are incredibly fast in comparison, but the cached builds aren't that much faster than uncached now.

Princesseuh:
> @Princesseuh not sure where the delays are, but the cached builds are definitely slower than before... the un-cached ones are incredibly fast in comparison, but the cached builds aren't that much faster than uncached now.

It's a mix of multiple things
- My comment here: https://github.com/withastro/astro/pull/8821#discussion_r1359954528 (especially for remote images)
- Concurrency doesn't necessarily play well with a lot of short operations like this

For local images on a fast hard drive, there's probably no big differences between this PR and main, but remote images suffer a bit more, I'll fix it

itsmatteomanf:
> It's a mix of multiple things
>
> - My comment here: feat: Rework image generation to improve performance #8821 (comment) (especially for remote images)
> - Concurrency doesn't necessarily play well with a lot of short operations like this

It's somewhat baffling to me though, adding concurrency doesn't only not improve the performance (for what is basically the same process as the uncached images, just without the fetch, which is basically what has always been done and not much different bar the actual copying to what local images do), but slows it down by almost an order of magnitude... there must be something tiny somewhere that causes this.

> I'll fix it.

I have no doubts about that ahah

natemoo-re (approved):
As discussed during today's standup, I'm totally comfortable with the tradeoff being made here! Approved. 

Uncached build are an order-of-magnitude, while cached builds seem to be slightly slower in _some_ cases. Hopefully that can be improved in a follow-up, but this is still worth it.

.changeset/good-mirrors-bake.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': minor
+---
+
+Improved performance of optimized images generation at build time. Astro will now generate optimized images concurrently, which can significantly speed up build times for sites with many images. Additionally, Astro will now reuse the same buffer for all variants of an image, which should improve performance for sites websites with many variants of the same image, especially when using remote images.
```
sarah11918, line 5:
```suggestion
Improved image optimization performance

Astro will now generate optimized images concurrently at build time, which can significantly speed up build times for sites with many images. Additionally, Astro will now reuse the same buffer for all variants of an image. This should improve performance for websites with many variants of the same image, especially when using remote images.
```
For the minor, just made the first line closer to a blog headline! Something like that maybe?

sarah11918 (commented):
Looks good! Since it's a minor, just tried to get a jump on the blog post format with the suggestion below. (And tiny tweaks)

.changeset/good-mirrors-bake.md
Princesseuh: This looks great to me!

PR #8823, opened 2023-10-13
fix: remote srcset images not being resized and deduplication not working in certain cases

Princesseuh:
## Changes

Continuation of https://github.com/withastro/astro/pull/8819

This PR also reworks completely how we do the transforms for `srcset`, putting special care into avoiding regenerating images when unnecessary. It's possible that there's still some edge cases where it doesn't work, nonetheless as far as I could tell, it does work!

This also fixes https://github.com/withastro/astro/issues/8839, since it's in the same area.

## Testing

Added tests!

## Docs

N/A

### DIFF
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -238,7 +238,7 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 		// However, if it's an imported image, we can use the original image's width as a maximum width
 		if (isESMImportedImage(options.src)) {
 			imageWidth = options.src.width;
-			maxWidth = imageWidth ?? Infinity;
+			maxWidth = imageWidth;
 		}
 
 		// If the user passed dimensions, we don't want to add it to the srcset
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -206,6 +206,10 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 			options.format = DEFAULT_OUTPUT_FORMAT;
 		}
 
+		// Sometimes users will pass number generated from division, which can result in floating point numbers
+		if (options.width) options.width = Math.round(options.width);
+		if (options.height) options.height = Math.round(options.height);
+
 		return options;
 	},
 	getHTMLAttributes(options) {
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -251,21 +251,24 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 				// If the user passed dimensions, we don't want to add it to the srcset
 				const { width: transformWidth, height: transformHeight, ...rest } = options;
 
-				const srcSetValue = {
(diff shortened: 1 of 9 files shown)


packages/astro/src/assets/services/service.ts
```diff
@@ -245,62 +255,46 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
...
-					srcSetValue.transform.width = maxTargetWidth;
-					srcSetValue.transform.height = Math.round(maxTargetWidth / aspectRatio);
-				}
+			allWidths.push(...widths.map((width) => ({ width, descriptor: `${width}w` as const })));
+		}
 
-				if (targetFormat !== options.format) {
-					srcSetValue.transform.format = targetFormat;
+		// Caution: The logic below is a bit tricky, as we need to make sure we don't generate the same image multiple times
+		// When making changes, make sure to test with different combinations of local/remote images widths, densities, and dimensions etc.
```
Princesseuh, line 269: Added a "There be dragons" comment here because it's honestly super tricky to completely avoid making transforms that are functionally equivalent to another one, but don't have quite the exact same shape.

Common examples includes:
- A transform where `webp` is specified manually vs one where it's not. Due to the default, both of those results in the same image.
- A transform where `width` is specified, but it's equivalent to the width of the image, so it's not resized.

packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -1,3 +1,4 @@
+import { deterministicString } from 'deterministic-object-hash';
```
Princesseuh, line 1: I found this package in the backwoods somewhere in a random stack overflow thread, but it's honestly quite nice and does everything I wanted.

Princesseuh:
!preview srcset-remote

itsmatteomanf:
Is there *any* chance to know when is this gonna get merged? It's a major bug fix on the newly announced feature...

I know this issue exists so I won't use `srcset` yet, but I'd really like to ahah

Princesseuh:
> Is there *any* chance to know when is this gonna get merged? It's a major bug fix on the newly announced feature...
> 
> I know this issue exists so I won't use `srcset` yet, but I'd really like to ahah

We'll release it Monday morning CET, it needs to get reviewed and we don't typically do releases on the weekend

packages/astro/src/assets/services/service.ts
```diff
@@ -245,62 +259,46 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
...
+				if (options.width && options.height) {
+					transform.width = options.width;
+					transform.height = options.height;
 				}
+			}
 
-				srcSet.push(srcSetValue);
+			srcSet.push({
+				transform: transform,
+				descriptor: descriptor,
```
ematipico, line 296:
```suggestion
				transform,
				descriptor,
```

ematipico (approved):
The logic is very complex, I can't really understand it.

ematipico:
It's okay, you did a great job! Thank you for addressing my comments. Feel free to merge it whenever you want. I don't have big concerns

PR #8824, opened 2023-10-13
Added OUTPUT dir to sitemap build command

silent1mezzo:
## Changes

This changes the CLI output from 

`@astrojs/sitemap: 'sitemap-index.xml' is created` to
`@astrojs/sitemap: 'sitemap-index.xml' created at '{OUTPUT DIRECTORY}'`

## Testing
The CLI output isn't tested anywhere (as far as I could tell).

## Docs
The CLI output isn't mentioned in the docs (as far as I could tell) so there's nothing that needs updating.

### DIFF
--- a/packages/integrations/sitemap/src/index.ts
+++ b/packages/integrations/sitemap/src/index.ts
@@ -177,7 +177,7 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
 						limit: entryLimit,
 						gzip: false,
 					});
-					logger.success(`\`${OUTFILE}\` is created.`);
+					logger.success(`\`${OUTFILE}\` created at \`${fileURLToPath(dir)}`);
 				} catch (err) {
 					if (err instanceof ZodError) {
 						logger.warn(formatConfigErrorMessage(err));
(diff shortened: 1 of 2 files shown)


packages/integrations/sitemap/src/index.ts
```diff
@@ -177,7 +177,7 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
 						limit: entryLimit,
 						gzip: false,
 					});
-					logger.success(`\`${OUTFILE}\` is created.`);
+					logger.success(`\`${OUTFILE}\` created at \`${fileURLToPath(dir)}`);
```
bluwy, line 180: I think we can extract `fileURLToPath(dir)` as its own variable, and maybe we can also do `path.relative(process.cwd(), destDir)` so that the logged path is shorter?

silent1mezzo: Will do!

silent1mezzo:
As far as I can tell the [failed test](https://github.com/withastro/astro/actions/runs/6508888358/job/17679237672?pr=8824#step:8:1248) feels like a flake? Will check back when I update the PR

packages/integrations/sitemap/src/index.ts
silent1mezzo: @bluwy updated.

ematipico:
@silent1mezzo no, we don't have many tests for the CLI, so local test with possible snapshots is very welcome alternative :)

bluwy (approved):
I think no tests for this is fine. It's usually hard to test it reliably.

evadecker:
Thanks for picking up this suggestion @silent1mezzo!

PR #8830, opened 2023-10-13
graceful exit on ctrl+c with the dev server

FredKSchott:
## Changes

- Exiting the dev server with CTRL+C is an expected way to exit the dev server
- This changes the exit code from 1 (error) to 0 (success) when you abort the dev server with CTRL+C
- Appearance-wise, this removes the scary " ELIFECYCLE  Command failed." log from pnpm or other task runners who read `1` as a fatal error that the user should be notified of.

### Before

Screen Shot 2023-10-13 at 1 24 38 PM

### After
Screen Shot 2023-10-13 at 1 25 02 PM

## Testing

- Not needed

## Docs

- Not needed

### DIFF
--- a/.changeset/old-pants-invite.md
+++ b/.changeset/old-pants-invite.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+add a graceful exit on ctrl+c with the dev server
--- a/packages/astro/src/cli/dev/index.ts
+++ b/packages/astro/src/cli/dev/index.ts
@@ -9,6 +9,12 @@ interface DevOptions {
 }
 
 export async function dev({ flags }: DevOptions) {
+	// Gracefully handle ctrl+c as a successful exit.
+	// This prevents tools like PNPM from showing scary
+	// red "ELIFECYCLE  Command failed." error messages.
+	process.on('SIGINT', function () {
+		process.exit(0);
+	});
 	if (flags.help || flags.h) {
 		printHelp({
 			commandName: 'astro dev',


Princesseuh:
I'm -1 on this. If we want to have a proper way to exit the dev server, we should have a keybind dedicated to it.

Hijacking signals and standards is not the proper way to do it (as small as it might seems.

FredKSchott:
> Hijacking signals and standards is not the proper way to do it imo (as small as it might seems)

So as a general statement I agree. But in Astro, we specifically expect (and document?) `CTRL+C` as the correct way to exit the server. There is no other way to exit the dev server.

In that framing, it's actually odd/unexpected to me that this would be considered an error. As a reminder, this is what people see in `pnpm` and Warp when you go to exit the Astro dev server correctly: 

image

image

This is definitely confusing to me as a user, and would be even more so confusing to a more junior developer. Surely this is a valid reason to control the exit code?

For prior art, Next and Nuxt both have this handling to exit with a success code on CTRL+C.

Princesseuh:
I understand the point, but I'm honestly still against it. Exiting Astro with CTRL+C isn't the expected solution because it's the best one, it's because we don't have a proper way to exit right now 🤷‍♀️ 

I can CTRL+C during type gen right now in Astro and end up with an outdated .astro folder, it's not expected, it's a forced interruption!

(I ultimately don't care enough to block this, but on principle, I don't agree with it 😠)

bluwy:
It looks like this PR listens for `SIGINT` directly rather than `Ctrl+C`, which maybe is slightly safer, but I also don't think we should force with exit code 0 in that case. We should only set an exit code if we get an error, otherwise it should be derived "naturally" as other tools may need to set it.

In this case, it seems like pnpm has a good reason (or a bug) to exit with code 1, and if this is confusing we should check and fix it on pnpm first.

ematipico (commented):
If we document/expect the users to use <kbd>Ctrl + C</kbd> to stop the server, we should document why the process exits with an error code/success code.

As for the code to return, I don't have strong opinions, but I would prefer to keep an error code. Although, as long as we document our reasons, I am happy with any change.

FredKSchott:
Considering this a breaking change, will re-open when a future Astro 4.0 is in beta

PR #8832, opened 2023-10-13
make unused variables a warning in eslint

FredKSchott:
## Changes

- as the title says, turn this error into a warning
- reason: Unused imports. Unused imports are a common occurrence during development. But in this repo, unused imports are automatically removed via our auto-formatter post-merge, so this isn't actually an issue for us. Treating these as errors creates distraction in the flow of development where the developer has to stop and resolve the error to make the file "green"/"good".

An alternative is to move this responsibility to `tsc` via `noUnusedLocals` and `noUnusedParameters`, which also treats unused imports as warnings automatically (in VSCode, at least) but keeps the rest as errors. But conceptually I'd prefer to keep linting and building under separate tools (`eslint` and `tsc`) so I'd only recommend that change if others feel uneasy about unused variables not triggering CI.

## Testing

- N/A

## Docs

- N/A


ematipico (approved):

bluwy:
IIUC this also doesn't error on unused variables in general now? e.g. `const something` in the file that's never used, or function parameters that's not used. The formatter doesn't auto-fix it for us so I'm unsure about making this a warning. I've personally not hit by this much myself until I'm ready to push my changes which I'd have to clean up these dangling code anyways 🤔

natemoo-re (approved):
I'm in favor of this! It's mostly an annoyance and often times I end up keeping my unused variable and renaming it to `_oldName` anyway.

bluwy:
If there's two approvals to this, I won't block the PR! I think we can merge it then.

FredKSchott:
Cool, glad this seems useful! Like all things linting you can always revisit this if people end up hating it. My personal preference is always for less linting in OSS due to PR friction, but I'm just a visitor here these days!

PR #8834, opened 2023-10-13
Update the ".astro" meta directory to a general concept

FredKSchott:
## Changes

- Refactor the `.astro` directory handling code for multi-purpose, no longer just for content.
- Refactor the image/assets logic to use this folder as the cache (`.astro/cache`) instead of `node_modules/.astro`.
- Update the ".astro" folder & ".astro/types.d.ts" file to now always exist, created on startup. No more runtime checks, fallbacks, etc.
- Update `env.d.ts` type generation to make Astro's ownership of the file explicit, and documented. No longer supported to make manual changes to this file.

## Testing

- Updated tests.

## Docs

- N/A (will confirm with docs team)


FredKSchott:
@Princesseuh @bholmesdev for your context: I ended up accidentally tackling this as a part of the logging work I started on in #8833. My understanding is that this is the direction you all would already be heading, but I'd love your confirmation before taking this much further. I'm also happy to hand this off to either of you / don't need to own this PR myself.

florian-lefebvre:
I like the `env.d.ts` to be _kinda_ the official place to do env augmentation (although it's possible in any `d.ts` file) and this change looks like a breaking change to me.

Since astro would generate the file from now on, isn't it possible to generate it somewhere else (eg. in cache) and add its location to the `include` field of the astro provided tsconfig.json? (the ones we extend from)

This way, the only thing users have to do is removing the top references in the `env.d.ts` like:
```diff
- /// <reference types="astro/client" />
- /// <reference types="./.astro/types.ts" />
```

Princesseuh:
> I like the `env.d.ts` to be _kinda_ the official place to do env augmentation (although it's possible in any `d.ts` file) and this change looks like a breaking change to me.
> 
> Since astro would generate the file from now on, isn't it possible to generate it somewhere else (eg. in cache) and add its location to the `include` field of the astro provided tsconfig.json? (the ones we extend from)
> 
> This way, the only thing users have to do is removing the top references in the `env.d.ts` like:
> ```diff
> - /// <reference types="astro/client" />
> - /// <reference types="./.astro/types.ts" />
> ```

We can't add it to the presets because the path is relative, so having `./.astro` in it wouldn't work. I guess we could make it so it includes every `.astro` folders though, though that'd probably require enforcing usage of our presets, which isn't necessarily amazing. 

With multiple exports now being possible, probably that we could have it separately from our presets. I'd actually love to work on this myself...

florian-lefebvre:
I'm not sure if this is what you're talking about, but now that tsconfig supports multiple extends, we could do something like nuxt:
```json
{
  "extends": ["./.astro/tsconfig.json", "astro/tsconfig/strict"]
}
```
This way, preset remains optional

FredKSchott:
Closing this PR for now, @bholmesdev or @Princesseuh may pick this back up in the future (and possibly walking back the node_modules cache directory changes, since that would break automatic caching behavior on some of our hosting platforms).

PR #8838, opened 2023-10-15
chore(deps): unpin tsconfck and use 3.0.0 release

dominikg:
3.0.0 is identical to 3.0.0-next.9, this allows astro users to update tsconfck more easily.

## Changes

- What does this change?

## Testing

no tests added because this is just updating a dependency version where both versions have identical code (and i'm not aware of any tests that cover the use of tsconfck in astro)

### DIFF
--- a/.changeset/red-socks-peel.md
+++ b/.changeset/red-socks-peel.md
@@ -1,0 +1,6 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


dominikg:
cc @Princesseuh

Princesseuh (approved):
Looks great, thank you!

PR #8840, opened 2023-10-15
Fix: Persist styles of persistent client:only components during view transitions in DEV mode

martrapp:
## Changes

Yet another attempt ;-) to preserve the stylesheets of `client:only` components during view transitions **in DEV mode**.
Production is not an issue as Vite statically bundles the style files when building for production.
 
### Situation

Stylesheets for client:only components are generated during hydration. They are not part of the static HTML files.
Therefore the comparison of header elements during `swap()` deletes them from the header.
`client:only` components can not simply be re-initialized in the new DOM because the browser won't re-import modules that are already loaded.  In a former approach I used random query strings in the URLs of style elements. While this worked for React-style imported CSS files, it failed for Vue's and Svelte's embedded styles. 

### How the fix works:

Only in DEV (treeshaken in PROD) and only if there are `client:only` components in the new DOM:
Hydrate the components in an environment where the module handler cache is cleared. This way, all style elements are generated. We can now identify style elements that come from hydration and not from the static HTML file. Theses are kept in the head during the transition.

## Testing

Reactivated and extended the existing e2e test case.
New test for Vue and Svelte components

## Documents

N/A, bug fixing.

### DIFF
--- a/.changeset/purple-dots-refuse.md
+++ b/.changeset/purple-dots-refuse.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixes styles of `client:only` components not persisting during view transitions
+Fixes styles of `client:only` components not persisting during view transitions in dev mode
--- a/.changeset/purple-dots-refuse.md
+++ b/.changeset/purple-dots-refuse.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Persist styles of persistent client:only components during view transitions
+Fixes styles of client:only components not persisting during view transitions
--- a/.changeset/purple-dots-refuse.md
+++ b/.changeset/purple-dots-refuse.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixes styles of client:only components not persisting during view transitions
+Fixes styles of `client:only` components not persisting during view transitions
--- a/.changeset/purple-dots-refuse.md
+++ b/.changeset/purple-dots-refuse.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Persist styles of persistent client:only components during view transitions
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -527,33 +527,49 @@ if (inBrowser) {
 
 // Keep all styles that are potentially created by client:only components
 // and required on the next page
+//eslint-disable-next-line @typescript-eslint/no-unused-vars
 async function prepareForClientOnlyComponents(newDocument: Document, toLocation: URL) {
-	const islands = newDocument.body.querySelectorAll("astro-island[client='only']");
-	if (islands) {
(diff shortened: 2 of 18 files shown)


.changeset/purple-dots-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Persist styles of persistent client:only components during view transitions
```
sarah11918, line 5:
```suggestion
Allows styles of `client:only` components to persist during view transitions
```
I can see this is a draft, but also that there's only going to be a changeset here, so I'll just give some pre-advice now then duck out!

For the final version, imagining the context in which these are read, it's often helpful to think of this statement as the answer to the question, "What does this PR do?" That starts with an implied "This PR..."

So, lots of things are possible! I just left one option! But if people are scanning changelogs, it's helpful to know instantly whether something was a fix, vs something new added (to use/learn about), vs a change in behaviour... So another alternative could be something like:

> Fixes styles of `client:only` components not persisting during view transitions

Lots of options! I'll just remind you that describing a particular code behaviour like the original statement isn't the same as telling someone, "This is what this PR gives you/does for you now" while priming them for whether this is a fix to a problem they might have had, or something new to explore, etc.

martrapp: Thank you very much, Sarah, I really appreciate your suggestions!

packages/astro/src/transitions/router.ts
```diff
@@ -519,3 +525,53 @@ if (inBrowser) {
 		markScriptsExec();
 	}
 }
+
+// Keep all styles that are potentially created by client:only components
+// and required on the next page
+async function prepareForClientOnlyComponents(newDocument: Document, toLocation: URL) {
```
matthewp, line 531: have you manually run the build and verified that this code gets removed in prod? I think it will, since the only usage is wrapped in a conditional, but I don't understand how the tree shaking works well enough to know for sure.

martrapp: Sure, 6k unziped after tree shaking, 9k when not guarded by meta.env.DEV. Checked the content of the file manually to understand that it is really this tree shaking

.changeset/purple-dots-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes styles of `client:only` components not persisting during view transitions
```
sarah11918, line 5:
```suggestion
Fixes styles of `client:only` components not persisting during view transitions in dev mode
```
Ooooh, and reading your PR description more closely, you even bolded this, so probably something for here, too!

It continues to be an amusing pattern to me of writing the good stuff in the PR description itself, which only reviewers will read.  :smile:  The PR title itself starts with `fix` to let other devs know what kind of change this PR is, and so it makes perfect sense that `Fixes...` would similarly be a great start for the entry in the changelog that users will read!

PR #8841, opened 2023-10-15
misc(create): Ignore errors removing directory after download failed

### LINKED ISSUE #8785: Failure to cleanup masks real error when creating new project
### Astro Info

```block
Astro                    v3.2.3
Package Manager          pnpm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

My workflow for creating new projects are:

1. Create a new folder
2. Open VSCode at the location
3. Run scaffolding command in VSCode Terminal.

When `pnpm create astro@latest` fails to download the template, it throws a unhelpful error:

```
      ▲  error EBUSY: resource busy or locked, rmdir '.'
```

This error is not the real reason why it failed, it also brings question of "why are you trying to delete my directory?"

Reproduction steps:

1. `mkdir repro && code repro && cd repro`, `code repro` can also be any other ways to lock the directory.
2. Add `127.0.0.1 github.com` to your `hosts` file, OR `$env:HTTPS_PROXY="http://127.0.0.1:123456789"`, OR unplug your ethernet cable.
3. Run `pnpm create astro` or it's equivalent.
4. Type `.` when answering "Where should we create your new project?"

I would expect failure to cleanup be ignored, and the real reason it failed is written to the terminal.

https://github.com/withastro/astro/blob/90d70eb7c4ceddf7246e338100e01de7dc8756f6/packages/create-astro/src/actions/template.ts#L88-L102

N/A

### Participation

Genteure:
## Changes

- Skip cleanup if the path is `.`, `./` or starts with `../`.
- Catches and ignore errors calling `rmdirSync` after template download failed.

This ensures the original "download failed" error is thrown and displayed to the user instead of a "rmdir failed" error which could be confusing.

Closes #8785 

## Testing

No new tests were added.
The changed function is not covered by existing tests.

## Docs

No docs changes were made.
I don't think this behavior detail is worth documenting on `docs.astro.build`.

### DIFF
--- a/.changeset/soft-berries-prove.md
+++ b/.changeset/soft-berries-prove.md
@@ -2,4 +2,4 @@
 'create-astro': patch
 ---
 
-Skip cleanup after template download failed if the path is `.`, `./` or starts with `../`.
+No longer attempts to delete the directory after a template download fails if the path is `.`, `./` or starts with `../`.
--- a/.changeset/soft-berries-prove.md
+++ b/.changeset/soft-berries-prove.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Skip cleanup after template download failed if the path is `.`, `./` or starts with `../`.
--- a/packages/create-astro/src/actions/template.ts
+++ b/packages/create-astro/src/actions/template.ts
@@ -93,7 +93,16 @@ export default async function copyTemplate(tmpl: string, ctx: Context) {
 				dir: '.',
 			});
 		} catch (err: any) {
-			fs.rmdirSync(ctx.cwd);
+			// Only remove the directory if it's most likely created by us.
+			if (ctx.cwd !== '.' && ctx.cwd !== './' && !ctx.cwd.startsWith('../')) {
+				try {
+					fs.rmdirSync(ctx.cwd);
+				} catch (_) {
+					// Ignore any errors from removing the directory,
+					// make sure we throw and display the original error.
+				}
+			}
+
 			if (err.message.includes('404')) {
 				throw new Error(`Template ${color.reset(tmpl)} ${color.dim('does not exist!')}`);
 			} else {


packages/create-astro/src/actions/template.ts
```diff
@@ -93,7 +93,16 @@ export default async function copyTemplate(tmpl: string, ctx: Context) {
 				dir: '.',
 			});
 		} catch (err: any) {
-			fs.rmdirSync(ctx.cwd);
+			// Only remove the directory if it's most likely created by us.
+			if (ctx.cwd !== '.' && ctx.cwd !== './' && !ctx.cwd.startsWith('../')) {
+				try {
+					fs.rmdirSync(ctx.cwd);
+				} catch (_) {
```
ematipico, line 100: We could write the error in a `console.error`

ematipico (approved):

natemoo-re (approved):
Thanks for the PR!

packages/create-astro/src/actions/template.ts
natemoo-re: I'm on the fence about logging it. We'd only hit this codepath if something else goes wrong, so this particular error seems irrelevant (or at least, much less important than the main error).

.changeset/soft-berries-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Skip cleanup after template download failed if the path is `.`, `./` or starts with `../`.
```
sarah11918, line 5:
```suggestion
No longer attempts to delete the directory after a template download fails if the path is `.`, `./` or starts with `../`.
```
Just wondering what you think about an example like this, @Genteure -- if these changelog messages are read by users, what about something more descriptive of the problem seen itself? (I'm inferring this from reading the attached issue. Make the statement correct as necessary!)

Your call, since your change does describe at a technical level what has changed in the code. But often these are meant to be user-facing "docs" of a sort describing effects people *using* the packages will see. If people want to know exactly how the code achieved that, they can look at the PR. So, it's an option to consider!

sarah11918 (commented):
Just left a suggestion for you to consider, @Genteure !  :raised_hands:

PR #8842, opened 2023-10-15
fix(assets): Fallback format not being taken into account properly

Princesseuh:
## Changes

What the title says.

This PR also adds an error message when converting from SVGs to raster formats because it's easier to do by accident on the Picture component

## Testing

Added a test

## Docs

N/A apart from the new error

### DIFF
--- a/.changeset/dull-bikes-decide.md
+++ b/.changeset/dull-bikes-decide.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix Picture component not taking into account the fallback format specified
--- a/.changeset/dull-bikes-decide.md
+++ b/.changeset/dull-bikes-decide.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fix Picture component not taking into account the fallback format specified
+Fixes Picture component not taking into account the fallback format specified
--- a/packages/astro/components/Picture.astro
+++ b/packages/astro/components/Picture.astro
@@ -11,7 +11,7 @@ type Props = (LocalImageProps | RemoteImageProps) & {
 	pictureAttributes?: HTMLAttributes<'picture'>;
 };
 
-const { formats = ['webp'], pictureAttributes = {}, ...props } = Astro.props;
+const { formats = ['webp'], pictureAttributes = {}, fallbackFormat, ...props } = Astro.props;
 
 if (props.alt === undefined || props.alt === null) {
 	throw new AstroError(AstroErrorData.ImageMissingAlt);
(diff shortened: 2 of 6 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -554,6 +554,21 @@ export const UnsupportedImageFormat = {
 		)} are supported by our image services.`,
 	hint: "Using an `img` tag directly instead of the `Image` component might be what you're looking for.",
 } satisfies ErrorData;
+
+/**
+ * @docs
+ * @see
+ * - [Images](https://docs.astro.build/en/guides/images/)
+ * @description
+ * Astro does not currently supporting converting between vector (such as SVGs) and raster (such as PNGs and JPEGs) images.x
```
florian-lefebvre, line 563:
```suggestion
 * Astro does not currently supporting converting between vector (such as SVGs) and raster (such as PNGs and JPEGs) images.
```

packages/astro/components/Picture.astro
```diff
@@ -24,16 +24,14 @@ const optimizedImages: GetImageResult[] = await Promise.all(
...
 );
 
-const fallbackFormat =
-	props.fallbackFormat ?? isESMImportedImage(props.src)
-		? ['svg', 'gif'].includes(props.src.format)
-			? props.src.format
-			: 'png'
-		: 'png';
+let resultFallbackFormat = fallbackFormat ?? 'png';
+if (isESMImportedImage(props.src) && ['gif', 'svg'].includes(props.src.format)) {
```
ematipico, line 28: nit: `['gif', 'svg']` could be moved into a top-level variable

ematipico (approved):

.changeset/dull-bikes-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix Picture component not taking into account the fallback format specified
```
sarah11918, line 5:
```suggestion
Fixes Picture component not taking into account the fallback format specified
```

sarah11918 (approved):
Just standardizing our verb tense for the changeset, otherwise, docs approves!

I realized we don't have this guidance in our actual image page in docs, so I'm making myself a task to add that info in.

packages/astro/components/Picture.astro
itsmatteomanf: I'm gonna agree here, it took me a second to understand what they did while investigating the issue on the previous code, and it's the same here. A tiny bit better, but a separate variable with a descriptive name would go along way.

Princesseuh: Sure, we can do that

PR #8850, opened 2023-10-17
feat(i18n): fallback system in SSR

ematipico:
## Changes

This PR wires the i18n middleware in the SSR system.

To achieve that, I had to update the `SSRManifest` with the information needed to handle the fallback. Considering the fact that `SSRManifest` is used in all three pipelines (DEV, SSG and SSR), the function `createI18nMiddleware` now accepts a `SSRManifest['i18n']` type.

Ultimately, I didn't use any Vite plugin; it wasn't needed.

## Testing

I added new tests. I removed those test cases that use `getStaticPaths`, because they doesn't work in SSR.

## Docs

N/A for now

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -26,6 +26,8 @@ import {
 import { matchRoute } from '../routing/match.js';
 import { EndpointNotFoundError, SSRRoutePipeline } from './ssrPipeline.js';
 import type { RouteInfo } from './types.js';
+import { createI18nMiddleware } from '../../i18n/middleware.js';
+import { sequence } from '../middleware/index.js';
 export { deserializeManifest } from './common.js';
 
 const clientLocalsSymbol = Symbol.for('astro.locals');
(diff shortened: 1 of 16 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -160,8 +162,19 @@ export class App {
...
+					this.#pipeline.setMiddlewareFunction(
+						sequence(i18nMiddleware, mod.onRequest as MiddlewareEndpointHandler)
+					);
+				} else {
+					this.#pipeline.setMiddlewareFunction(i18nMiddleware);
+				}
+			} else {
+				if (mod.onRequest) {
+					this.#pipeline.setMiddlewareFunction(mod.onRequest as MiddlewareEndpointHandler);
+				}
```
matthewp, line 177: Shouldn't there be very similar code in ssg generate and dev? Is the middleware added differently there?

ematipico: Oh, I see what you mean! The only issue I see is that the manifest isn't inside the `Pipeline`, which means that we can't move the `if/else`  blocks to check `manifest.i18n` inside a single place

matthewp: So how *is* the middleware added in those places currently?

matthewp: I think it's ok for now. When we (or I) do the middleware integrations proposal we can revisit.

matthewp (approved):

Issue #8702, opened 2023-09-29
vite.config.js gets used by astro... intentional?


AimForNaN:
### Astro Info

```block
Astro                    v3.1.4
Node                     v20.7.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Not sure if it is intentional, but the astro docs gives the impression that this probably shouldn't be the case. From a practical point of view, it would be nice if these two things were kept separate anyway. `vite.config.js` can store conflicting configuration that could cause astro to fail to start a dev server and even fail to generate static content.

### What's the expected result?

Having two separate vite configurations (e.g. `astro.config.mjs` and `vite.config.js`) that aren't aware of each other's existence.

### Link to Minimal Reproducible Example

https://github.com/AimForNaN/astro-test

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Yeah I think it's unintentional. We should add a `configFile: false` inline Vite config to prevent this. I hope this isn't relied upon by others to be a breaking change though, especially that we already have the `vite` config option.

PR #8852, opened 2023-10-17
fix(astro): build config from astro conflict with vite.config.js

rayriffy:
## Changes

- `configFile: false` added to specifically tell Vite not to automatically find, read, and merge config from `vite.config.js`. which could conflict with Astro's vite config ([docs](https://vitejs.dev/guide/api-javascript.html#inlineconfig))

fixes https://github.com/withastro/astro/issues/8702

## Testing

N/A

## Docs

N/A. Docs already said in the way to modify Vite behavior via `astro.config.mjs`

### DIFF
--- a/.changeset/fresh-lizards-whisper.md
+++ b/.changeset/fresh-lizards-whisper.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Only use Vite config from astro.config.mjs as source of truth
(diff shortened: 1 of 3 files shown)


ematipico:
@rayriffy I think you want to have the configuration [here](https://github.com/withastro/astro/blob/fc838c5564cf84c1340d3568373ad24b8193b7e8/packages/astro/src/core/create-vite.ts#L102).

So it has effects on the dev server too.

rayriffy:
I agree, I forgot to trace back to see where the origin of common config itself is. Change has been revised.

Princesseuh (approved):

Issue #7031, opened 2023-05-08
Typescript starter template doesn't include required commands


CGamesPlay:
### What version of `astro` are you using?

2.4.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

None

### Describe the Bug

I encountered a lot of difficulty using the TypeScript starter template with Astro, so I wanted to file this bug to document some of the things that would make the DX much cleaner (also, for anyone else who wonders why the default TypeScript template for Astro doesn't actually include type checking).

I created a project using `npm create astro@latest -- --skip-houston --typescript strictest`. I see a message saying "TypeScript customized" in the output, and the program has created my site, which builds cleanly. I open [Card.astro](https://github.com/withastro/astro/blob/50bf66e4df8ac9b427e653a48cb8e4e84913d8f2/examples/basics/src/components/Card.astro) and change every instance of "body" to "content".

[As documented](https://docs.astro.build/en/guides/typescript/), "the Astro dev server won’t perform any type checking", so the dev server shows 4 empty cards (on the index page) with no errors in the terminal or browser console. Strangely, `npm run build` also cleanly finishes, resulting in a broken site.

It turns out that the starter template doesn't seem to fully set up TypeScript, and there are at least 2 steps that need to be manually performed, both documented on the same page. First, we need to install and configure `@astrojs/ts-plugin` (it says you need to do this "if you are not using VSCode", which I can only assume also means "if you are a CI environment", but it isn't totally clear). Then, [much further down on the same page](https://docs.astro.build/en/guides/typescript/#type-checking) there is an additional note that we need to replace our build script with `astro check && tsc --noEmit && astro build`. After doing these, the script successfully reports the errors that we introduced.

I'm listing this as a bug because I think these are pretty much required for anyone interested in using TypeScript with Astro, and the setup script, despite advertising that it has configured TypeScript, did not perform these required steps. Separately from this, I also wasn't able to find anything about improving the DX here, like showing the errors in "astro dev" somehow, although I assume this is on the roadmap.

### Link to Minimal Reproducible Example

https://github.com/withastro/astro/blob/50bf66e4df8ac9b427e653a48cb8e4e84913d8f2/examples/basics/

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
`@astrojs/ts-plugin` is a TypeScript plugin, those only run in the editor. So no, it is not required in CI.

[...]

PR #8853, opened 2023-10-17
feat(create-astro): add astro checks to build script when user choose TypeScript

rayriffy:
## Changes

- If user choose TypeScript strictness to `strictest`, the `create-astro` command will also add `astro checks` to build script
- Additionally, `@astrojs/check` and `typescript` will be installed if user previously opt-in to install dependencies
- Internally, some parts of `typescript.ts` uses both async/sync of `node:fs`. I also did some refactoring to be all async

fixes https://github.com/withastro/astro/issues/7031

## Testing

- Writing test to make sure that build script `package.json` is actually changed

## Docs

Maybe need to add extra docs to [TypeScript](https://docs.astro.build/en/guides/typescript/) to explain strictest behavior when use `pnpm create astro`


packages/create-astro/src/actions/typescript.ts
```diff
@@ -71,29 +74,76 @@ export async function typescript(
...
-			const result = Object.assign(templateTSConfig, {
-				extends: `astro/tsconfigs/${value}`,
-			});
+const FILES_TO_UPDATE = {
+	'package.json': async (
+		file: string,
+		options: { value: string; ctx: PickedTypeScriptContext }
+	) => {
+		// do not add astro check command to build script if option is not strictest
+		if (options.value !== 'strictest') return;
```
natemoo-re, line 83: I think we might want to apply this for `strict` as well? The user has said that they do want to use typescript at this point, so I think it's fair!

natemoo-re (approved):
This is great, thanks so much for contributing!

Just wondering about always applying this if the user sets up TypeScript. I usually go with `strict` but I'd still like this to be setup for me.

packages/create-astro/src/actions/typescript.ts
rayriffy: Done! I also made some changes for test to reset fixtures back to original state more consistently.

.changeset/ninety-onions-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': minor
+---
+
+Add `astro-checks` command when select TypeScript option to strictest
```
natemoo-re, line 5:
```suggestion
Automatically setup `astro check` command when configuring TypeScript
```

natemoo-re (approved):
Looks great! Thanks so much for the PR!

.changeset/ninety-onions-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': minor
+---
+
+Automatically setup `astro check` command when configuring TypeScript
```
sarah11918, line 5:
```suggestion
Automatically installs the required dependencies to run the `astro check` command when `strict` or `strictest` 
 TypeScript mode is selected.
```
Since running `astro check` requires both `@astrojs/check` and `typescript`, maybe makes sense to say "required dependencies" (plural) here, because these will be added to your project?

sarah11918 (commented):
Just a quick suggestion from docs re: the wording here! See whether you think the extra context is helpful!

Princesseuh (approved):
Just a comment, but feature looks fine to me. I do feel like we'll get support threads of people complaining that their project doesn't build with various errors, but meh, it's okay, people should and will learn.

.changeset/ninety-onions-bow.md
rayriffy: I do agree. But I have to adjust some wording at the end because `astro check` will be applied to all TypeScript options (`relaxed`, `strict`, `strictest`).

Let me know if this change looks good to you.

packages/create-astro/src/actions/typescript.ts
```diff
@@ -39,7 +42,7 @@ export async function typescript(
 	} else {
 		if (!['strict', 'strictest', 'relaxed', 'default', 'base'].includes(ts)) {
 			if (!ctx.dryRun) {
-				fs.rmSync(ctx.cwd, { recursive: true, force: true });
+				rm(ctx.cwd, { recursive: true, force: true });
```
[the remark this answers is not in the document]
rayriffy: Great catch. Thanks!

.changeset/ninety-onions-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': minor
+---
+
+Automatically installs the required dependencies to run the `astro check` command when TypeScript mode is selected.
```
sarah11918, line 5: So to clarify, `astro-check` is added even when someone says "No I don't intend to write TypeScript?" Which would be every single new project?

In that case, the above is fine, but it's really just:

```suggestion
Automatically installs the required dependencies to run the `astro check` command.
```
Am I missing something? I'm not sure it matters that it happens when you choose which TypeScript mode, since the choice you make doesn't affect whether or not it's installed? If so, I think just THAT it installs the dependencies seems adequate. But, either would be fine!

sarah11918 (approved):
Left a comment to clarify with an optional suggestion, but also fine as is!

.changeset/ninety-onions-bow.md
rayriffy: No, if user don't intend to write TypeScript `astro check` will not be added.

1. If user does not choose to use TypeScript, nothing happens **(no behavior change)**
2. If user choose TypeScript, `astro check` will be added

sarah11918: > I do agree. But I have to adjust some wording at the end because astro check will be applied to all TypeScript options (relaxed, strict, strictest).

This was your reply to my other comment, where I suggested mentioning only `strict` and `strictest`.

natemoo-re: The intent is that the `astro check` code is added only if the user answers "Yes" to "Do you plan to write TypeScript?". Even if they choose "relaxed" they will still get the `astro check` code.

Do you plan to write TypeScript? Yes or No

sarah11918: This is arguably a lot of discussion for something not super important. :sweat_smile: 

Just checking, we have the templates `base`, `strict` and `strictest`?

And my original suggestion was because selecting that the user does intend to write TypeScript is different from "when configuring TypeScript" because you're ALWAYS configuring TS at this step. (You're just not always choosing one of the stricter settings.) Similarly, it's not "when choosing a TS mode" because you're always choosing a TS mode at this point, just not always one that installs `astro check`.

This is why I stand by this option as being the most informative:

> Automatically installs the required dependencies to run the `astro check` command when `strict` or `strictest` TypeScript mode is selected.

However, if it's hidden from the user that their choice selects one of those two modes by name (e.g. by using different words), then a more user-facing alternative could be something like:

> Automatically installs the required dependencies to run the `astro check` command when the user indicates they plan to write TypeScript.

natemoo-re, line 5: Fair. I think that final suggestion is the clearest! One of these days we'll all agree on how the CLI should handle this 😅

```suggestion
Automatically installs the required dependencies to run the astro check command when the user indicates they plan to write TypeScript.
```

natemoo-re (approved):

.changeset/ninety-onions-bow.md
natemoo-re: I had hoped my review would reply to this comment, but it didn't. I like your final suggestion here. Resolved in https://github.com/withastro/astro/pull/8853#discussion_r1369276927

PR #8854, opened 2023-10-17
Refactor content collection generation

natemoo-re:
## Changes

- **DO NOT MERGE.** This PR is meant to be used as a preview release! Mainline release will require moving this behind an experimental flag so it is opt-in.
- As part of exploring https://github.com/withastro/roadmap/issues/698, this PR implements a major refactor to how content collections are built
- Instead of relying on Vite's `import.meta.glob`, which will invalidate a collection every time one of the modules it references changes, this implements a custom build for content collections where individual modules can be built separately.

## Testing

TODO. Lots of manual testing at the moment, but this needs to be automated.

## Docs

Not required until https://github.com/withastro/roadmap/issues/698 is further along.


packages/astro/src/core/build/plugin.ts
```diff
@@ -74,7 +69,7 @@ export function createPluginContainer(options: StaticBuildOptions, internals: Bu
 			const mutations = new Map<
 				string,
 				{
-					build: 'server' | 'client';
+					targets: BuildTarget[];
```
natemoo-re, line 72: Since we now have three build modes, the internal build plugin system has been refactored to accept an array of `'server' | 'content' | 'client'` rather than a single value.

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,105 @@
...
+			this.emitFile({
+				type: 'prebuilt-chunk',
+				code: content,
+				fileName: 'content/index.mjs'
+			})
+		},
+
+		async writeBundle(options, bundle) {
+			const dist = options.dir!
+			const cache = dist.replace('dist', 'node_modules/.astro/content-cache')
```
natemoo-re, line 50: Using `node_modules/.astro/content-cache` since the asset generation uses this directory as well, but this caching logic should be centralized!

packages/astro/src/core/build/plugins/plugin-renderers.ts
```diff
@@ -48,7 +48,8 @@ export function vitePluginRenderers(opts: StaticBuildOptions): VitePlugin {
 
 export function pluginRenderers(opts: StaticBuildOptions): AstroBuildPlugin {
 	return {
-		build: 'ssr',
+		// TODO: handle for content?
+		targets: ['server'],
```
natemoo-re, line 52: TODO: fix this!

packages/astro/src/core/build/static-build.ts
```diff
@@ -424,7 +514,7 @@ async function copyFiles(fromFolder: URL, toFolder: URL, includeDotfiles = false
 			const lastFolder = new URL('./', to);
 			return fs.promises
 				.mkdir(lastFolder, { recursive: true })
-				.then(() => fs.promises.copyFile(from, to));
+				.then(() => fs.promises.copyFile(from, to, fs.constants.COPYFILE_FICLONE));
```
natemoo-re, line 517: Small perf optimization: we can use [copy on write](https://en.wikipedia.org/wiki/Copy-on-write) if available. Node automatically uses a fallback on unsupported systems.

packages/astro/src/core/build/plugins/plugin-content.ts
Princesseuh: Just a heads up that we have a `config.cacheDir`: https://docs.astro.build/en/reference/configuration-reference/#cachedir

natemoo-re: Ahh excellent, thank you!

packages/astro/src/core/build/static-build.ts
Princesseuh: macOS people stay winning (and bearded people on Linux, I guess)

natemoo-re:
!preview content-cache

natemoo-re:
!preview content-cache

natemoo-re:
Working on a refactor to be stacked against this PR https://github.com/withastro/astro/compare/refactor/content-collections...refactor/collections-build

.changeset/lovely-pianos-build.md
```diff
@@ -0,0 +1,22 @@
+---
+'astro': minor
+---
+
+Provides a new, experimental build cache for [Content Collections](https://docs.astro.build/en/guides/content-collections/). This includes multiple refactors to Astro's build process to optimize how Content Collections are handled, which should provide significant performance improvements for users with many collections.
+
+You can enable the build cache by adding the following flag to your Astro config:
```
sarah11918, line 7:
```suggestion
You can opt in to preview the new build cache by adding the following flag to your Astro config:
```
I think something like this implies a bit more that this is the way things will be in the future, and you get to opt in early, while it's experimental.

sarah11918 (approved):
Looks great! Happy to see this one squeak in!  🙌  Tiny suggestion for your consideration below!

.changeset/lovely-pianos-build.md
natemoo-re: Great suggestion, thank you!

.changeset/lovely-pianos-build.md
```diff
@@ -0,0 +1,22 @@
...
+```js
+// astro.config.mjs
+export default {
+  experimental: {
+    contentCollectionCache: true,
+  },
+};
+```
+
+When this experimental feature is enabled, you can bypass the local build cache by passing the `--force` flag to `astro build`.
```
matthewp, line 18: Can you add some of the numbers to this? This will go into the article so it would be good for context.

Even specifically the docs numbers would be good as that is a real world use-case.

matthewp: Separately, would be good to mention a little about how this works as far as where the cache is stored, explaining how it is retained in CI, etc.

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -42,7 +42,7 @@ export default function assets({
 				extendManualChunks(outputOptions, {
 					after(id) {
 						if (id.includes('astro/dist/assets/services/')) {
-							return `astro-assets-services`;
+							return `astro/assets-service`;
```
matthewp, line 45: is this for some convention we have, that everything astro core is in an `astro/` folder?

packages/astro/src/core/build/plugins/plugin-chunks.ts
```diff
@@ -0,0 +1,31 @@
+import type { Plugin as VitePlugin } from 'vite';
+import type { AstroBuildPlugin } from '../plugin.js';
+import { extendManualChunks } from './util.js';
+
+export function vitePluginChunks(): VitePlugin {
```
matthewp, line 5: can you add some comments about what these 2 exports functions are file? It doesn't seem quite clear to me.

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,288 @@
...
+import pLimit from 'p-limit';
+import { extendManualChunks } from './util.js';
+import { isServerLikeOutput } from '../../../prerender/utils.js';
+import { encodeName } from '../util.js';
+
+const CONTENT_CACHE_DIR = './content/';
+const CONTENT_MANIFEST_FILE = './manifest.json';
+// IMPORTANT: Update this version when making significant changes to the manifest format.
+// Only manifests generated with the same version number can be compared.
+const CONTENT_MANIFEST_VERSION = 0;
```
matthewp, line 23: smart!

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,288 @@
...
+
+	return {
+		name: '@astro/plugin-build-content',
+
+		async options(options) {
+			let newOptions = Object.assign({}, options);
+			newManifest = await generateContentManifest(opts, lookupMap);
+			entries = getEntriesFromManifests(oldManifest, newManifest);
+
+			for (const { type, entry } of entries.buildFromSource) {
```
matthewp, line 76: I *think* know what this is doing but I'm not 100%, some comments explaining would be helpful

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,288 @@
...
+				const mod = bundle[key];
+				if (mod.type === 'asset') return;
+				if (mod.facadeModuleId === resolvedVirtualEmptyModuleId) {
+					delete bundle[key];
+				}
+			});
+		},
+
+		async writeBundle() {
+			const clientComponents = new Set([
```
matthewp, line 154: why are these part of the content manifest? Can you explain?

.changeset/lovely-pianos-build.md
natemoo-re: Good call! Updated!

packages/astro/src/assets/vite-plugin-assets.ts
natemoo-re: Not an established convention, but yes, I was thinking it would be nice to organize different parts of the Astro server code in a single `astro/` directory.

packages/astro/src/core/build/plugins/plugin-content.ts
natemoo-re: Thanks! Seemed wise to have a kill-switch if needed.

matthewp (approved):

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -20,98 +21,177 @@ import {
...
+	fs,
 }: AstroContentVirtualModPluginParams): Plugin {
+	let IS_DEV = false;
+	const IS_SERVER = isServerLikeOutput(settings.config);
+	return {
+		name: 'astro-content-virtual-mod-plugin',
+		enforce: 'pre',
+		configResolved(config) {
+			IS_DEV = config.mode === 'development'
+		},
```
ematipico, line 44: `settings` already contains the `config`; why do we need `configResolved` in this case?

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -20,98 +21,177 @@ import {
...
+					} satisfies AstroPluginMetadata
+				};
+			}
+		},
+		renderChunk(code, chunk) {
+			if (!settings.config.experimental.contentCollectionCache) {
+				return;
+			}
+			if (code.includes(RESOLVED_VIRTUAL_MODULE_ID)) {
+				const depth = chunk.fileName.split('/').length - 1;
```
ematipico, line 86: Will this work on Windows?

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -20,98 +21,177 @@ import {
...
-				return {
-					code: virtualModContents.replace(
-						'/* @@LOOKUP_MAP_ASSIGNMENT@@ */',
-						`lookupMap = ${stringifiedLookupMap};`
-					),
-				};
+function getStringifiedCollectionFromLookup(wantedType: 'content' | 'data' | 'render', relContentDir: string, lookupMap: ContentLookupMap) {
+	let str = '{';
+	// In dev, we don't need to normalize the import specifier at all. Vite handles it.
+	let normalize = (slug: string) => slug;
```
ematipico, line 151: This can be `let normalize` because `if/else` both reach the point where we do `normalize =`

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -20,98 +21,177 @@ import {
...
+	// For prod builds, we need to transform from `/src/content/**/*.{md,mdx,json,yaml}` to a relative `./**/*.mjs` import
+	if (process.env.NODE_ENV === 'production') {
+		const suffix = wantedType === 'render' ? '.entry.mjs' : '.mjs';
+		normalize = (slug: string) => `${removeFileExtension(encodeName(slug)).replace(relContentDir, './')}${suffix}`
+	} else {
+		let suffix = '';
+		if (wantedType === 'content') suffix = CONTENT_FLAG;
+		else if (wantedType === 'data') suffix = DATA_FLAG;
+		else if (wantedType === 'render') suffix = CONTENT_RENDER_FLAG;
```
ematipico, line 160: This can be `else` because we checked all types

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -20,98 +21,177 @@ import {
...
+	return virtualModContents;
+}
 
-				return {
-					code: virtualModContents.replace(
-						'/* @@LOOKUP_MAP_ASSIGNMENT@@ */',
-						`lookupMap = ${stringifiedLookupMap};`
-					),
-				};
+function getStringifiedCollectionFromLookup(wantedType: 'content' | 'data' | 'render', relContentDir: string, lookupMap: ContentLookupMap) {
```
ematipico, line 148: It seems we introduced some new logic: `'content' | 'data' | 'render'`. Maybe it's worth having a comment that explains what we do. 

If you think it's too technical, we could also have a `README.md`

packages/astro/src/core/build/plugins/plugin-content.ts
ematipico, line 1: Can we add a description of this new plugin inside the README? https://github.com/withastro/astro/blob/main/packages/astro/src/core/build/plugins/README.md

packages/astro/src/core/build/plugins/plugin-chunks.ts
ematipico, line 1: Can we add a description of this new plugin inside the README? https://github.com/withastro/astro/blob/main/packages/astro/src/core/build/plugins/README.md

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+	return { version: -1, entries: [], serverEntries: [], clientEntries: [] };
+}
+
+function vitePluginContent(opts: StaticBuildOptions, lookupMap: ContentLookupMap, internals: BuildInternals): VitePlugin {
+	const { config } = opts.settings;
+	const { cacheDir } = config;
+	const distRoot = config.outDir;
+	const distContentRoot = new URL('./content/', distRoot);
+	const cachedChunks = new URL('./chunks/', opts.settings.config.cacheDir);
+	const distChunks = new URL('./chunks/', opts.settings.config.outDir);
```
ematipico, line 54: nit: we extract `config`, but then we use `opts.settings.config` later. Or we use `distRoot` AND `opts.settings.config.outDir` which are the same. It can get confusing, so maybe we should try to use the same data everywhere

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+	const contentCacheDir = new URL(CONTENT_CACHE_DIR, cacheDir);
+	const contentManifestFile = new URL(CONTENT_MANIFEST_FILE, contentCacheDir);
+	const cache = contentCacheDir;
+	const cacheTmp = new URL('./.tmp/', cache);
```
ematipico, line 58: Why do we use a `tmp` folder inside the cache? Wouldn't it be more appropriate to use the the temporary folder of the OS?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+	let oldManifest = createContentManifest();
+	let newManifest = createContentManifest();
```
ematipico, line 60:
```js
let oldManifest = createContentManifest();
let newManifest = oldManifest;
```

I think there's no need to call `createContentManifest` twice here.

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+	let entries: ContentEntries;
+	let injectedEmptyFile = false;
+
+	if (fsMod.existsSync(contentManifestFile)) {
+		try {
+			const data = fsMod.readFileSync(contentManifestFile, { encoding: 'utf8' });
+			oldManifest = JSON.parse(data);
+			internals.cachedClientEntries = oldManifest.clientEntries;
```
ematipico, line 68: I think it's fine if we suppress the `readFileSync` error, but a `JSON.parse` error can mean that the cache was tampered with, and we should tell it to the user somehow. And we should **wipe** the current cache because it's unusable.

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+		} catch { }
+	}
+
+	return {
+		name: '@astro/plugin-build-content',
+
+		async options(options) {
+			let newOptions = Object.assign({}, options);
```
ematipico, line 76: We should use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+						const collectionEntry = findEntryFromSrcRelativePath(lookupMap, srcRelativePath);
+						if (collectionEntry) {
+							let suffix = '.mjs';
+							if (flag === PROPAGATED_ASSET_FLAG) {
+								suffix = '.entry.mjs';
+							}
+							id = removeLeadingForwardSlash(removeFileExtension(encodeName(id.replace(srcPath, '/')))) + suffix;
+							return id;
+						}
+					}
```
ematipico, line 126: Can we add a comment that explains what we do here?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+
+		async writeBundle() {
+			// These are stored in the manifest to ensure that they are included in the build
+			// in case they aren't referenced _outside_ of the cached content.
+			// We can use this info in the manifest to run a proper client build again.
+			const clientComponents = new Set([
+				...oldManifest.clientEntries,
+				...internals.discoveredHydratedComponents.keys(),
+				...internals.discoveredClientOnlyComponents.keys(),
+				...internals.discoveredScripts,
```
ematipico, line 170: Since we use a `Set`, I assume the order here is important?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+
+			const cacheExists = fsMod.existsSync(cache);
+			fsMod.mkdirSync(cache, { recursive: true })
+			await fsMod.promises.mkdir(cacheTmp, { recursive: true });
+			await copyFiles(distContentRoot, cacheTmp, true);
+			if (cacheExists) {
+				await copyFiles(contentCacheDir, distContentRoot, false);
+			}
+			await copyFiles(cacheTmp, contentCacheDir);
+			await fsMod.promises.rm(cacheTmp, { recursive: true, force: true });
```
ematipico, line 191: Can we enclose all these FS operations in a `try/catch` block? There might be remote situations where we don't have write access

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+			])
+			// Likewise, these are server modules that might not be referenced
+			// once the cached items are excluded from the build process
+			const serverComponents = new Set([
+				...oldManifest.serverEntries,
+				...internals.discoveredHydratedComponents.keys(),
+			]);
+			newManifest.serverEntries = Array.from(serverComponents);
+			newManifest.clientEntries = Array.from(clientComponents);
+			await fsMod.promises.mkdir(contentCacheDir, { recursive: true });
```
ematipico, line 180: Should we check if `contentCacheDir` exists already?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+		} else {
+			entries.buildFromSource.push(entry);
+		}
+	}
+	return entries;
+}
+
+async function generateContentManifest(opts: StaticBuildOptions, lookupMap: ContentLookupMap): Promise<ContentManifest> {
+	let manifest: ContentManifest = { version: CONTENT_MANIFEST_VERSION, entries: [], serverEntries: [], clientEntries: [] };
+	const limit = pLimit(10);
```
ematipico, line 241: Is this an arbitrary number?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -0,0 +1,302 @@
...
+	const promises: Promise<void>[] = [];
+
+	for (const [collection, { type, entries }] of Object.entries(lookupMap)) {
+		for (const entry of Object.values(entries)) {
+			const key: ContentManifestKey = { collection, type, entry };
+			const fileURL = new URL(encodeURI(joinPaths(opts.settings.config.root.toString(), entry)));
+			promises.push(limit(async () => {
+				const data = await fsMod.promises.readFile(fileURL, { encoding: 'utf8' });
```
ematipico, line 249: Is this operation safe? Should we check if `fileURL` exists and if we can read the file?

packages/astro/src/core/build/util.ts
```diff
@@ -36,3 +36,19 @@ export function i18nHasFallback(config: AstroConfig): boolean {
 
 	return false;
 }
+
+export function encodeName(name: string): string {
```
ematipico, line 40: We should at least explain what kind of encoding this function does. Or maybe an example.

PR #8863, opened 2023-10-18
fix(astro): terminal HMR path (close #8831)

### LINKED ISSUE #8831: CLI paths take the PC Root, not the project's path root.
### Astro Info

```block
"Astro                    v3.2.2\nNode                     v18.16.0\nSystem                   Windows (x64)\nPackage Manager          npm\nOutput                   server\nAdapter                  @astrojs/vercel/serverless\nIntegrations             @custom/update-webflow-js\n                         @astrojs/tailwind\n                         @astrojs/preact\n                         @astrojs/prefetch\n                         astro-robots-txt"
```

_No response_

### Describe the Bug

The Astro CLI during npm run dev is using my entire pc path for outputs, can this be modified to instead use the project's root path?

image

I would think the project's root path is better! instead of the entire PC path !

https://stackblitz.com/edit/github-n2mknz?file=README.md

### Participation

florian-lefebvre:
## Changes

- #8831
- Uses fileURLToPath and nozmalizes the path between OSs

## Testing

Tested by pointing the `@example/basics` astro version to the local one (`workspace:*`), only on Windows.

## Docs

No change required

### DIFF
--- a/.changeset/two-lamps-tell.md
+++ b/.changeset/two-lamps-tell.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes the terminal HMR path
--- a/.changeset/two-lamps-tell.md
+++ b/.changeset/two-lamps-tell.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixes the terminal HMR path
+Fixed an issue where the dev server logged the full file path on updates.
--- a/.changeset/two-lamps-tell.md
+++ b/.changeset/two-lamps-tell.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixed an issue where the dev server logged the full file path on updates.
+Fixes an issue where the dev server logged the full file path on updates.
--- a/packages/astro/src/vite-plugin-astro/hmr.ts
+++ b/packages/astro/src/vite-plugin-astro/hmr.ts
@@ -1,4 +1,5 @@
 import { fileURLToPath } from 'node:url';
+import { relative } from "node:path";
 import type { HmrContext, ModuleNode } from 'vite';
 import type { AstroConfig } from '../@types/astro.js';
 import {
@@ -91,7 +92,7 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	const file = relative(process.cwd(), ctx.file.replace(config.root.pathname, '/'));
 
 	// If only styles are changed, remove the component file from the update list
 	if (isStyleOnlyChange) {
--- a/packages/astro/src/vite-plugin-astro/hmr.ts
+++ b/packages/astro/src/vite-plugin-astro/hmr.ts
@@ -91,9 +91,7 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = fileURLToPath(`file:///${ctx.file}`)
-		.replace(fileURLToPath(config.root), '/')
-		.replace(/\\/g, '/');
+	const file = ctx.file.replace(fileURLToPath(config.root).replace(/\\/g, '/'), '/');
 
 	// If only styles are changed, remove the component file from the update list
 	if (isStyleOnlyChange) {
--- a/packages/astro/src/vite-plugin-astro/hmr.ts
+++ b/packages/astro/src/vite-plugin-astro/hmr.ts
@@ -95,7 +95,7 @@ export async function handleHotUpdate(
 	if (!file.startsWith('/')) {
 		file = `/${file}`
 	}
-	file = file.replace(config.root.pathname, '/')
+	file = file.replaceAll(' ', '%20').replace(config.root.pathname, '/')
 
 	// If only styles are changed, remove the component file from the update list
 	if (isStyleOnlyChange) {
--- a/packages/astro/src/vite-plugin-astro/hmr.ts
+++ b/packages/astro/src/vite-plugin-astro/hmr.ts
@@ -91,11 +91,9 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	let file = ctx.file;
-	if (!file.startsWith('/')) {
-		file = `/${file}`
-	}
-	file = file.replaceAll(' ', '%20').replace(config.root.pathname, '/')
+	const file = fileURLToPath(`file:///${ctx.file}`)
+		.replace(fileURLToPath(config.root), '/')
+		.replace(/\\/g, '/');
 
 	// If only styles are changed, remove the component file from the update list
 	if (isStyleOnlyChange) {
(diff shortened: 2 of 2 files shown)


packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -91,7 +92,7 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	const file = relative(process.cwd(), ctx.file.replace(config.root.pathname, '/'));
```
lilnasy, line 95: I think involving cwd might create unexpected results. Most users would start the dev server from the root of their projects, but they should not see something different if that is not the case.

florian-lefebvre: could `import.meta.url` help?

lilnasy: `import.meta.url` would likely be somewhere in node_modules.

It also seems like `relative()`, and `ctx.file.replace()` are attempting to do the same thing: removing the first few parts of ctx.file.

Why doesn't the `replace()` call work on Windows?

florian-lefebvre: Pushing a fix soon but here is what happens:
```ts
{
  pathname: '/D:/DEV/PROJECTS/withastro/astro/examples/basics/',
  file: 'D:/DEV/PROJECTS/withastro/astro/packages/astro/dist/jsx-runtime/index.js'
}
```
the only thing is to prepend a slash if needed

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -91,7 +91,11 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	let file = ctx.file;
+	if (!file.startsWith('/')) {
+		file = `/${file}`
+	}
+	file = file.replace(config.root.pathname, '/')
```
lilnasy, line 98: Nice! That fixes it.

It looks like it still shows the full path when the project name has a space, like in the case of the issue screenshot. Can you check what happens there?

florian-lefebvre: I handled only spaces, I don't know if there are any other special character that causes the issue

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -91,7 +91,11 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	let file = ctx.file;
+	if (!file.startsWith('/')) {
+		file = `/${file}`
+	}
+	file = file.replaceAll(' ', '%20').replace(config.root.pathname, '/')
```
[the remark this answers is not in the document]
natemoo-re: [`fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl) is probably what you're looking for!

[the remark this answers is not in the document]
florian-lefebvre: Now I understand what you meant @lilnasy! Things are coming together in my head ahah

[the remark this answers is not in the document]
florian-lefebvre: We should be good now! Until I forget something else 😅

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -91,7 +91,9 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	const file = fileURLToPath(`file:///${ctx.file}`)
+		.replace(fileURLToPath(config.root), '/')
+		.replace(/\\/g, '/');
```
lilnasy, line 96:
```suggestion
	const file = ctx.file.replace(fileURLToPath(config.root));
```
Was converting ctx.file path to a URL necessary? I would think converting config.root URL to path is enough.

.changeset/two-lamps-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes the terminal HMR path
```
lilnasy, line 5: Astro users will see this in the changelog, and they may not have full context about what happened.
"Fixed an issue where the dev server logged the full file path on updates."

packages/astro/src/vite-plugin-astro/hmr.ts
florian-lefebvre: I confirm you're suggestion (`ctx.file.replace(fileURLToPath(config.root), '/')`) does not work

.changeset/two-lamps-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixed an issue where the dev server logged the full file path on updates.
```
sarah11918, line 5:
```suggestion
Fixes an issue where the dev server logged the full file path on updates.
```
Just standardizing our changeset tense... Docs out!

packages/astro/src/vite-plugin-astro/hmr.ts
lilnasy, line 96: Thanks for bearing with me so far. This looks great on Windows now!

There is one minor blemish on linux.
image

florian-lefebvre: Do you think the following could be acceptable?
```ts
	const file = fileURLToPath(`file:///${ctx.file.substring(ctx.file.startsWith('/') ? 1 : 0)}`)
		.replace(fileURLToPath(config.root), '/')
		.replace(/\\/g, '/');
```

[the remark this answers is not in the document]
florian-lefebvre: I'll give it a go 👍

lilnasy: It could, but this is becoming kinda long. Maybe if the approach from my other comment does not work.

ematipico, line 96: Can't we make the paths relative to the Astro working directory? It seems more simple...

florian-lefebvre: That was my first attempt, I need to see if it can be simplified first using only one `fileURLToPath`

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -91,7 +91,7 @@ export async function handleHotUpdate(
 	// Bugfix: sometimes style URLs get normalized and end with `lang.css=`
 	// These will cause full reloads, so filter them out here
 	const mods = [...filtered].filter((m) => !m.url.endsWith('='));
-	const file = ctx.file.replace(config.root.pathname, '/');
+	const file = ctx.file.replace(fileURLToPath(config.root).replace(/\\/g, '/'), '/');
```
natemoo-re, line 94: Super minor, but we have a few utilities that might be helpful exported from [`@astrojs/internal-helpers/path`](https://github.com/withastro/astro/blob/main/packages/internal-helpers/src/path.ts). Namely [`slash`](https://github.com/withastro/astro/blob/0d4922dffe7d96b907db89ec502f77efd572cc56/packages/internal-helpers/src/path.ts#L92) could replace the manual regex logic here.

ematipico (approved):

PR #8870, opened 2023-10-19
fix(cli): prevent production install on add cmd

### LINKED ISSUE #8037: `astro add` causes dev dependencies to be removed from node_modules

2.10.5

None

npm

Linux

Firefox

### Describe the Bug

Using `astro add` causes a weird issue, where any existing installed dev dependencies will be removed from `node_modules`.

Stackblitz has repro instructions, but here's some you can also use locally.

1. new project (`npm create astro@latest`)
2. install a dev dependency (e.g. `npm i daisyui -D`)
3. check it exists in node_modules
4. use astro add (e.g. `npx astro add react -y`)
5. check to find the dev dependency missing from node_modules, but still in the package.json

Dev dependencies to still be present, after using `astro add`.

https://stackblitz.com/edit/github-3sk8rq?file=md!README.md&on=stackblitz

### Participation

xstevenyung:
## Changes

- force `NODE_ENV=development` on installation command run when using `astro add` command to prevent installation to remove `devDependencies`

## Testing

no real way to test

## Docs

no need

fix #8037

### DIFF
--- a/.changeset/great-queens-yawn.md
+++ b/.changeset/great-queens-yawn.md
@@ -1,5 +1,5 @@
 ---
-'astro': minor
+'astro': patch
 ---
 
 prevent production install on astro add cmd
--- a/.changeset/great-queens-yawn.md
+++ b/.changeset/great-queens-yawn.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+prevent production install on astro add cmd
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -700,7 +700,11 @@ async function tryToInstallIntegrations({
 						...inheritedFlags,
 						...installCommand.dependencies,
 					],
-					{ cwd }
+					{
+						cwd,
+						// reset NODE_ENV to ensure install command run in dev mode
+						env: { NODE_ENV: undefined },
+					}
 				);
 				spinner.succeed();
 				return UpdateResult.updated;


packages/astro/src/cli/add/index.ts
```diff
@@ -700,7 +700,7 @@ async function tryToInstallIntegrations({
 						...inheritedFlags,
 						...installCommand.dependencies,
 					],
-					{ cwd }
+					{ cwd, env: { NODE_ENV: 'development' } }
```
bluwy, line 703: Interesting! We actually set the `NODE_ENV` here so the config file is loaded correctly:

https://github.com/withastro/astro/blob/0d4922dffe7d96b907db89ec502f77efd572cc56/packages/astro/src/cli/index.ts#L127

On this line, could we set `NODE_ENV: undefined` instead to reset the value? Also would be great to add an additional comment explaining why we need it.

xstevenyung: the reset seems to work with `undefined`.
i also added a comment to explain why we are doing this 👍

jacobthesheep:
Can’t wait for this PR to be merged

ematipico (approved):

ematipico:
@xstevenyung can you add a changeset via `pnpm changeset`?

xstevenyung:
added the changeset as a minor bump

.changeset/great-queens-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
```
bluwy, line 2: I think this should be a patch since it's causing an unexpected behaviour before. We usually reserve minors for new features.

bluwy:
```suggestion
'astro': patch
```

bluwy (approved):

PR #8872, opened 2023-10-19
[ci] Update `!preview` response

natemoo-re:
## Changes

- Updates our snapshot release workflow
- Moves away from the deprecated `::set-output` to the new `>> $GITHUB_OUTPUT` method for passing values between job steps.
- Hopefully fixes some formatting in the response message (but this is difficult to test)

## Testing

Not sure how I can test this, tbh...

## Docs

N/A

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -76,8 +76,13 @@ jobs:
       - name: Bump Package Versions
         id: changesets
         run: |
-          pnpm exec changeset version --snapshot ${{ steps.getSnapshotName.outputs.result }} > changesets.output.txt 2>&1
-          echo ::set-output name=result::`cat changesets.output.txt`
+          pnpm exec changeset status --output changesets.output.json 2>&1
+          pnpm exec changeset version --snapshot ${{ steps.getSnapshotName.outputs.result }} > changesets-version.output.txt 2>&1
+          {
+            echo 'result<<__EOF'
+            cat changesets.output.json
+            echo __EOF
+          } >> "$GITHUB_OUTPUT"
         env:
           # Needs access to run the script
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -88,7 +93,11 @@ jobs:
           pnpm run release --tag next--${{ steps.getSnapshotName.outputs.result }} > publish.output.txt 2>&1
           echo "Release complete"
           cat publish.output.txt
-          echo ::set-output name=result::`cat publish.output.txt`
+          {
+            echo 'result<<__EOF'
+            cat publish.output.txt
+            echo __EOF
+          } >> "$GITHUB_OUTPUT"
         env:
           # Needs access to publish to npm
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
@@ -97,10 +106,23 @@ jobs:
       - name: Pull Request Notification
         uses: actions/github-script@v6
         env:
+          RELEASES: ${{ steps.changesets.outputs.result }}
           MESSAGE: ${{ steps.publish.outputs.result }}
         with:
           script: |
             console.log(process.env.MESSAGE);
+            const changeset = JSON.parse(process.env.RELEASES);
+            const message = 'The following packages have been released:'
+            for (const release of changeset.releases) {
+              if (release.type === 'none') continue;
+              message += `\n- ${release.name}@${release.newVersion}`;
+            }
+            message += '\n';
+            message += '<details><summary>Build Log</summary>\n\n```\n';
+            message += process.env.message;
+            message += '\n```\n\n</details>';
+
+            
             github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -92,7 +92,7 @@ jobs:
           cat version.output.json
 
           echo "status<<$EOF" >> $GITHUB_OUTPUT
-          echo "$(cat version.output.json)" >> $GITHUB_OUTPUT
+          echo "$(cat version.output.txt)" >> $GITHUB_OUTPUT
           echo "$EOF" >> $GITHUB_OUTPUT
         env:
           # Needs access to run the script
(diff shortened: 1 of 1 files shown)


lilnasy:
Tested it by butchering my fork and there seems to be an error.
image

Found a stackoverflow thread for this but I can't make sense of it.
https://stackoverflow.com/questions/74137120/how-to-fix-or-avoid-error-unable-to-process-file-command-output-successfully

.github/workflows/snapshot-release.yml
```diff
@@ -76,8 +76,13 @@ jobs:
...
         run: |
-          pnpm exec changeset version --snapshot ${{ steps.getSnapshotName.outputs.result }} > changesets.output.txt 2>&1
-          echo ::set-output name=result::`cat changesets.output.txt`
+          pnpm exec changeset status --output changesets.output.json 2>&1
+          pnpm exec changeset version --snapshot ${{ steps.getSnapshotName.outputs.result }} > changesets-version.output.txt 2>&1
+          {
+            echo 'result<<__EOF'
+            cat changesets.output.json
+            echo __EOF
+          } >> "$GITHUB_OUTPUT"
```
ematipico, line 85: I don't know this syntax :O (the curly brackets). What does it do?

.github/workflows/snapshot-release.yml
```diff
@@ -97,10 +106,23 @@ jobs:
       - name: Pull Request Notification
         uses: actions/github-script@v6
         env:
+          RELEASES: ${{ steps.changesets.outputs.result }}
           MESSAGE: ${{ steps.publish.outputs.result }}
         with:
           script: |
             console.log(process.env.MESSAGE);
+            const changeset = JSON.parse(process.env.RELEASES);
```
ematipico, line 114: Maybe we should put this operation in a `try/catch`, just in case? But it's more a nitpick, feel free to ignore it if it's not worth it.

ematipico (approved):
Looks good to me

.github/workflows/snapshot-release.yml
natemoo-re: I copied this from the GitHub docs but it doesn't seem to work. It was supposed to run `>> $GITHUB_OUTPUT` after each action, but it seems invalid. Switched this to a more traditional syntax.

.github/workflows/snapshot-release.yml
natemoo-re: Good idea! Can't hurt.

.github/workflows/snapshot-release.yml
```diff
@@ -97,10 +104,25 @@ jobs:
...
+          RELEASES: ${{ steps.changesets.outputs.result }}
           MESSAGE: ${{ steps.publish.outputs.result }}
         with:
           script: |
             console.log(process.env.MESSAGE);
+            let changeset = { releases: [] };
+            try {
+              changeset = JSON.parse(process.env.RELEASES);
+            } catch (e) {}
+            const message = 'The following packages have been released:'
```
lilnasy, line 116:
```suggestion
            let message = 'The following packages have been released:'
```
`message` is a const, but it is being mutated later.

https://github.com/lilnasy/astro/actions/runs/6589260196/job/17903321737

lilnasy:
It works well: https://github.com/lilnasy/astro/pull/5#issuecomment-1772894334

There seems to be something weird going on with encoding still, and really we only care about the last 4 lines. right?

alexanderniebuhr (approved):
Love it. Would like to see it in a shared repo soon, so we can use the same for withastro/adapters 🚀

.github/workflows/snapshot-release.yml
```diff
@@ -76,34 +81,86 @@ jobs:
...
+
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+
+          echo "version<<$EOF" >> $GITHUB_OUTPUT
+          echo "$(cat version.output.json)" >> $GITHUB_OUTPUT
+          echo "$EOF" >> $GITHUB_OUTPUT
+          cat version.output.json
+
+          echo "status<<$EOF" >> $GITHUB_OUTPUT
+          echo "$(cat status.output.json)" >> $GITHUB_OUTPUT
```
lilnasy, line 95:
```suggestion
          echo "$(cat version.output.json)" >> $GITHUB_OUTPUT
```

lilnasy: Maybe that's not right but status.outout.json does not exist

lilnasy: https://github.com/lilnasy/astro/actions/runs/6590567270/job/17907450693

.github/workflows/snapshot-release.yml
```diff
@@ -76,34 +81,86 @@ jobs:
...
-          echo ::set-output name=result::`cat changesets.output.txt`
+          pnpm exec changeset status --output status.output.json 2>&1
+          pnpm exec changeset version --snapshot ${{ steps.getSnapshotName.outputs.result }} > version.output.txt 2>&1
+
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+
+          echo "version<<$EOF" >> $GITHUB_OUTPUT
+          echo "$(cat version.output.json)" >> $GITHUB_OUTPUT
+          echo "$EOF" >> $GITHUB_OUTPUT
+          cat version.output.json
```
lilnasy, line 92: version.output.json did not get created for some reason

https://github.com/lilnasy/astro/actions/runs/6590750901/job/17907965624#step:12:26

lilnasy:
```suggestion
          echo "$(cat version.output.txt)" >> $GITHUB_OUTPUT
          echo "$EOF" >> $GITHUB_OUTPUT
          cat version.output.txt
```

.github/workflows/snapshot-release.yml
```diff
@@ -76,34 +81,86 @@ jobs:
...
+
+          echo "status<<$EOF" >> $GITHUB_OUTPUT
+          echo "$(cat version.output.json)" >> $GITHUB_OUTPUT
```
lilnasy, line 95:
```suggestion
          echo "$(cat version.output.txt)" >> $GITHUB_OUTPUT
```

.github/workflows/snapshot-release.yml
natemoo-re: Good catch!

lilnasy:
Yeah, that was it https://github.com/lilnasy/astro/pull/5#issuecomment-1773133424

natemoo-re:
Thanks for all your help @lilnasy! Think this is working how I want now ☺️

Issue #8558, opened 2023-09-14
CSS Modules inject order


ymatuhin:
### Astro Info

```block
Astro                    v3.0.13
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/preact
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

When I import CSS modules from Preact component, is goes first in order of all styles. Even before global styles. That messes up specificity for components.

### What's the expected result?

They should be treated as scoped css and be applied after globals.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-k1nega-mj5wcr?file=src%2Fcomponents%2FCounter.module.css

### Participation

- [ ] I am willing to submit a pull request for this issue.

<img width="595" alt="Screenshot 2023-09-14 at 17 05 02" src="https://github.com/withastro/astro/assets/3757971/9e0422d7-4164-4337-9694-06c8a50f36b6">

natemoo-re:
Just to confirm, does the issue only occur in dev mode? Is the `astro build` output different?

ymatuhin:
> Just to confirm, does the issue only occur in dev mode? Is the `astro build` output different?

Nope, build output has the same issue.

lilnasy:
Could not reproduce the issue. The text renders green as expected.

<img width="350" alt="image" src="https://github.com/withastro/astro/assets/69170106/ebfc9d25-31c6-459c-ab44-0c082cc18fcf">

I find it really odd that in your screenshot global.css is a script. This is the result I'm getting from building your project.
<img width="350" alt="image" src="https://github.com/withastro/astro/assets/69170106/829e957d-51c2-447a-893e-ce70fef47318">

You might want to open a support thread on our discord: https://astro.build/chat.

duncannah:
I'm having this same problem, only occurs on production build

lilnasy:
@duncannah I can reproduce this issue only in dev mode. Could you open another issue with a minimal reproduction?

ymatuhin:
@lilnasy yeah, looks like it happens only in dev mode. Sorry for misguidance.

PR #8877, opened 2023-10-20
Fix CSS modules ordering

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/8558

Fix CSS modules ordering by rendering styles first before links. This is needed because in `getStylesForURL`, CSS modules are passed as links (`importedCssUrls`) (CSS module has no `ssrModule.default`):

https://github.com/withastro/astro/blob/5ea6ee0ed494c792a4c94928a83c5c85b9b6ac32/packages/astro/src/vite-plugin-astro-server/css.ts#L34-L46

We then put the CSS module links in `links`, and the other styles (global styles) in `styles`.

https://github.com/withastro/astro/blob/5ea6ee0ed494c792a4c94928a83c5c85b9b6ac32/packages/astro/src/vite-plugin-astro-server/route.ts#L301-L330

Ultimately this creates a tricky situation where we shouldn't actually group head elements this way. Scripts/styles/links should all be in a single array but that takes a larger refactor which we could revisit in the future.

---

Also, the content rendering runtime also has styles first before links:

https://github.com/withastro/astro/blob/5ea6ee0ed494c792a4c94928a83c5c85b9b6ac32/packages/astro/src/content/runtime.ts#L316

---

In most case, non-CSS links could be sorted anywhere though, and I can confirm CSS links are only for CSS modules, so I think this change is harmless. During builds, all CSS are links so there's no ordering issue.

## Testing

Added a new integration test.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/runtime/server/render/head.ts
+++ b/packages/astro/src/runtime/server/render/head.ts
@@ -34,6 +34,12 @@ export function renderAllHeadContent(result: SSRResult) {
 		.filter(uniqueElements)
 		.map((link) => renderElement('link', link, false));
 
+	// Order styles -> links -> scripts similar to src/content/runtime.ts
+	// The order is usually fine as the ordering between these groups are mutually exclusive,
+	// except for CSS styles and CSS stylesheet links. However CSS stylesheet links usually
+	// consist of CSS modules which should naturally take precedence over CSS styles, so the
+	// order will still work. In prod, all CSS are stylesheet links.
+	// In the future, it may be better to have only an array of head elements to avoid these assumptions.
 	let content = styles.join('\n') + links.join('\n') + scripts.join('\n');
 
 	if (result._metadata.extraHead.length > 0) {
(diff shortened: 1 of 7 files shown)


packages/astro/src/runtime/server/render/head.ts
```diff
@@ -34,7 +34,7 @@ export function renderAllHeadContent(result: SSRResult) {
 		.filter(uniqueElements)
 		.map((link) => renderElement('link', link, false));
 
-	let content = links.join('\n') + styles.join('\n') + scripts.join('\n');
+	let content = styles.join('\n') + links.join('\n') + scripts.join('\n');
```
ematipico, line 37: Maybe it's worth adding a comment that explains the ordering. In the description, you have a lot of important context, which will be lost once this PR is merged.

sarah11918 (approved):
Docs approves changeset!

matthewp (approved):

packages/astro/src/runtime/server/render/head.ts
bluwy: Added some comments

PR #8878, opened 2023-10-20
feat(i18n): virtual module APIs

ematipico:
## Changes

This PR aligns the name of the APIs exported by the virtual module `astro:i18n`

The `*Absolute*` APIs work using `site` at the moment. Their behaviour will change once we will implement the domain support.

I will tackle `getCurrentLocale` later. It may require more information that we don't currently have, infra speaking.

## Testing

I added new test case and I tested the APIs locally using a pet project

## Docs

N/A for now


packages/astro/src/i18n/index.ts
```diff
@@ -47,6 +78,21 @@ export function getLocalesBaseUrl({ base, locales, trailingSlash, format }: GetL
...
 
+export function getLocaleAbsoluteUrlList({ site, ...rest }: GetLocaleAbsoluteUrl) {
+	const locales = getLocaleRelativeUrlList(rest);
+	return locales.map((locale) => {
+		if (site) {
+			if (site.endsWith('/') || locale.startsWith('/')) {
+				return `${site}${locale}`;
+			} else {
+				return `${site}/${locale}`;
+			}
```
natemoo-re, line 89: We have `@astrojs/internal-helpers/path` for operations like this. Not sure this is exactly what you need, but I'd reccommend using utils from this package or adding the util you need. https://github.com/withastro/astro/blob/main/packages/internal-helpers/src/path.ts#L63-L76

packages/astro/src/i18n/index.ts
```diff
@@ -29,14 +39,35 @@ export function getI18nBaseUrl({ locale, base, locales, trailingSlash, format }:
...
+/**
+ * The absolute URL
+ */
+export function getLocaleAbsoluteUrl({ site, ...rest }: GetLocaleAbsoluteUrl) {
+	const locale = getLocaleRelativeUrl(rest);
+	if (site) {
+		if (site.endsWith('/') || locale.startsWith('/')) {
+			return `${site}${locale}`;
+		} else {
+			return `${site}/${locale}`;
```
natemoo-re, line 51: Same comment!

natemoo-re (approved):

PR #8885, opened 2023-10-21
fix: snapshot release main branch not found

alexanderniebuhr:
## Changes

- fixes main branch not found

## Testing

- tested on the adapters repo: https://github.com/withastro/adapters/pull/34

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -61,6 +61,9 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ steps.refs.outputs.head_ref }}
+          fetch-depth: 0
+
+      - run: git fetch origin main:main
 
       - name: Setup PNPM
         uses: pnpm/action-setup@v2


alexanderniebuhr:
* main
  * **PR #8885** <a href="https://app.graphite.dev/github/pr/withastro/astro/8885" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/withastro/astro/8885?utm_source=stack-comment).

.github/workflows/snapshot-release.yml
```diff
@@ -61,6 +61,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ steps.refs.outputs.head_ref }}
+          fetch-depth: 0
```
ematipico, line 64: This fetches all the history, are you sure about this change??? 
> Set fetch-depth: 0 to fetch all history for all branches and tags

alexanderniebuhr: I think it is necessary to be able to run the next step which is `git checkout main`, since that command needs more than the current ref commit. 

cc @lilnasy do you know by chance if it would work without `fetch-depth: 0`?

ematipico: > since that command needs to have the remote history of main too

You haven't explained why. Sorry I don't have much context

alexanderniebuhr: Sorry about that.

If you didn't set `fetch-depth: 0` using the default, only the specific commit of the ref you're checking out would be fetched. This would mean that the repository on the runner would be in a "shallow" state. In a shallow clone, not all history is available, so if you tried to check out a branch like main that wasn't included in the shallow fetch, the checkout would fail because that branch's history wouldn't be present.

That's at least my understanding.

ematipico: Ok, we had a similar can in Biome, and we solved it like this:

```yml
      - name: Checkout Main Branch
        uses: actions/checkout@v4
        with:
          clean: false
          ref: main
```

alexanderniebuhr: Yeah that make total sense, good point.

I forgot to say, that we need two refs here, because `changeset` tries to compare those two (`main <> pr head`)

So in the action we would need to set, `ref: ${{ steps.refs.outputs.head_ref }}` & `ref: main`, which I don't think would work.

ematipico: Okay, let's test `fetch-dep` and check if that works, while looking for possible alternatives.

alexanderniebuhr: `fetch-depth` works in the `withastro/adapter` repo, used it there already.
But I agree :100: % it is not the best solution.

lilnasy: > cc @lilnasy do you know by chance if it would work without fetch-depth: 0?

It does not. Both `fetch-depth: 0`, and then fetching main specifically afterwards were necessary to make `changeset status` work.

lilnasy:
- Workaround for changesets/changesets#517

matthewp (approved):

.github/workflows/snapshot-release.yml
alexanderniebuhr: Thanks for looking into it. For now I think the only solution is to use `fetch-depth: 0`, hopefully that upstream issue get's resolved

lilnasy (approved):

PR #8889, opened 2023-10-22
Chores: Tidying up the view transition router

martrapp:
## Changes

Chores: View transition router
- Simplification
- Documentation
- Improved comprehensibility and maintainability

No changes in functionality

## Testing

Small changes for less flaky executions

## Docs

n/a not even a change set ;-)

### DIFF
--- a/packages/astro/e2e/view-transitions.test.js
+++ b/packages/astro/e2e/view-transitions.test.js
@@ -1,5 +1,5 @@
 import { expect } from '@playwright/test';
-import { testFactory } from './test-utils.js';
+import { testFactory, waitForHydrate } from './test-utils.js';
 
 const test = testFactory({ root: './fixtures/view-transitions/' });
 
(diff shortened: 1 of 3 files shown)


packages/astro/e2e/view-transitions.test.js
```diff
@@ -696,7 +697,7 @@ test.describe('View Transitions', () => {
 		await page.goto(astro.resolveUrl('/client-only-three'));
 		let msg = page.locator('#name');
 		await expect(msg).toHaveText('client-only-three');
-		await page.waitForTimeout(400); // await hydration
+		await page.waitForTimeout(1000); // await hydration
```
bluwy, line 700: Not a big fan of `waitForTimeout` 😅 In this case for hydration, could it be swapped to the `waitForHydrate` util?

We could also create utils that wait for events like `astro:page-load` and `astro:after-swap` so it runs more reliably.

martrapp: Thank you, good point, I wasn't aware of those!

ematipico (approved):

packages/astro/e2e/view-transitions.test.js
martrapp: All `waitForTimeout`s gone.

bluwy (approved):

PR #8893, opened 2023-10-23
Prevent bumping peer deps

bluwy:
## Changes

The integrations' `astro` peer dep ranges has been updating, which is a breaking change. I changed it to an explicit version range instead of `workspace:` so changeset doesn't bump it now. Works especially in combination of `the onlyUpdatePeerDependentsWhenOutOfRange` experimental option.

The peer dep ranges are all reset to `"astro": "^3.0.0"`

## Testing

Ran `pnpm changeset version` locally.

## Docs

n/a.


natemoo-re (approved):
Oh yikes, good catch!

ematipico (commented):
This will bite us when we are going to do another major release, and the ranges aren't automatically updated. So if you want to go ahead, we must update our internal documentation and remind us that this is a manual step. Why not use `workspace:^3.0.0`?

bluwy:
I'll update the internal docs 👍 `workspace:^3.0.0` is currently what's causing the issue and changesets is bumping it every version.

ematipico (approved):

bluwy:
@ematipico looks like with the `onlyUpdatePeerDependentsWhenOutOfRange` config, changesets is smart enough to also bump the peer dep whenever we enter prerelease mode or cut a new major, so looks like no additional docs is needed.

PR #8903, opened 2023-10-24
Light/dark theming for shikiji's codeblocks

horo-fox:
## Changes

- This passes through `themes` from markdown configuration to the markdown usage in the markdown package.

I would mark this as experimental as shikiji's support is "experimental" but I'm not sure how. I don't think normal versioning guarantees should apply.

## Testing

I've added a new test case.

## Docs

I avoided documentation because this is experimental (according to shikiji) and I don't think Astro wants to look like it's promising any sort of stability.


packages/markdown/remark/src/types.ts
```diff
@@ -42,6 +42,7 @@ export type RemarkRehype = Omit<RemarkRehypeOptions, 'handlers' | 'unknownHandle
 export interface ShikiConfig {
 	langs?: LanguageRegistration[];
 	theme?: BuiltinTheme | ThemeRegistration | ThemeRegistrationRaw;
+	themes?: Record<string, BuiltinTheme | ThemeRegistration | ThemeRegistrationRaw>;
```
ematipico, line 45: Should we call this `experimentalThemes`?

ematipico (commented):
I believe it's OK to update the documentation to document this new feature, although I think it should contain "experimental" in its name and the documentation.

sarah11918 (commented):
Just speaking for docs here, @horo-fox ! As to whether or not new documentation is needed on our Astro Docs site, when we switched to shikiji in the first place, it seemed to be described as just an internal refactor and we did not update any docs. Our docs do talk about setting `shiki` config options etc.

I'll maybe also loop @bluwy in here: do any of our existing docs need to change, or would our users still set `shiki` config options, which internally will be processed using `shikiji`? My big question is, if users want to set their own syntax highlighting options, do they need to know about shikiji or are we still just referring to it and setting it through a user-facing shiki config?

horo-fox:
> My big question is, if users want to set their own syntax highlighting options, do they need to know about shikiji or are we still just referring to it and setting it through a user-facing shiki config?

I suppose they should know about shikiji because the CSS to make this work depends on a bit of implementation detail (the default `defaultColor` (light) since I didn't expose that, the way the CSS vars get created, and having <https://github.com/antfu/shikiji#query-based-dark-mode> linked would help).

But I'm not sure if that's what is best. Maybe it might be better to rename the CSS variables, expose `defaultColor`, and talk about the CSS needed to expose this?

bluwy:
I haven't fully digest the PR and will take a deeper look tomorrow, but the general idea make sense to me. (and also the experimental name). Also from a brief look, the PR isn't covering all the cases yet. I'd suggest looking for files that contain the word `shiki.ts`, `remark-shiki.ts`, and `schema.ts` for additional places to tweak.

It's not great that shiki handling is all over the place in our codebase currently but I'll work on refactoring that soon 😅 

> I'll maybe also loop @bluwy in here: do any of our existing docs need to change, or would our users still set `shiki` config options, which internally will be processed using `shikiji`? My big question is, if users want to set their own syntax highlighting options, do they need to know about shikiji or are we still just referring to it and setting it through a user-facing shiki config?

Yes I think we can keep referring to setting `shiki` options. Technically speaking, this new option is powered by `shikiji` so we should also mention and link to [its docs](https://github.com/antfu/shikiji?tab=readme-ov-file#lightdark-dual-themes). But conceptually, we can keep referring as `shiki` as the new API is still using shiki-compatible types.

[Talking about this page](https://docs.astro.build/en/guides/markdown-content/#shiki-configuration), there's also the `wrap` option which isn't a shiki config, but an Astro-specific addition. So in some ways this PR will be similar too.

`shikiji` also doesn't have documentation for `theme` and `langs` so we still need to link to shiki for the docs.

horo-fox:
I've gone ahead and make a docs PR over at https://github.com/withastro/docs/pull/5191, only PR review comment left is making sure this is complete...

sarah11918 (approved):
Just making my review here official! I've also marked the corresponding docs PR as approved, just waiting for this to be released before it's merged live into docs. :partying_face:

horo-fox:
I've gone ahead and checked the files you mentioned and a bit more (the `Code` element which I held off on updating... but TBH updating it is probably fine? I'd be OK with reverting that) -- @bluwy mind taking another look?

horo-fox:
Bump for @bluwy in case this got dropped somewhere ^^

.changeset/tender-suits-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
+Adds experimental support for multiple themes in `shikiji`.
```
bluwy, line 5:
```suggestion
---
'@astrojs/markdown-remark': minor
'astro': minor
---

Adds experimental support for multiple shiki themes with the new `markdown.shikiConfig.experimentalThemes` option.
```

Extended the changeset a bit to highlight the new schema option, and marked minor as these will be new features. I think we can merge this on next Wednesday when we usually do minors.

bluwy (approved):
Awesome work! The implementation and tests looks clean to me 👍 One small suggestion about the changeset below.

.changeset/tender-suits-glow.md
sarah11918: Love this change, @bluwy !

bluwy:
Looks like a merge conflict. I'll resolve it and merge this one shortly

PR #8904, opened 2023-10-24
feat: options to avoid processing images at build in hybrid mode

y-nk:
## Changes

This adds a `transform: 'eager' | 'lazy'` parameter to astro asset's `getImage`/`<Image />` which allows for image NOT to build processed at build, but rather let it flow to runtime processing.

## Motivation

I'm deploying my website on AWS with a "standard" serverless stack (cloudfront/lambda@edge/s3) and since CloudFront has a good dynamic cache policy, I wanted to let the lambda@edge process my images and update CF's cache instead of building the images at build.

_PS: I can't use a full `output: server` because our CMS does not perform well and TTFB could be > 5s in some cases, so i must prebuild the html pages._

I believe this could also serve a use-case where consumers of this feature would like to have control over what image is transformed at build and what image is transformed at runtime. One could want to transform at build the first 10 pages of a 100 pages product list, but for the "furthest" pages which nobody look at, having runtime transform could help decrease the build time (pre-processing images for potentially no reason).

## Testing

I've a pet project which i toggle between `output: 'static'` and `output: 'hybrid'` and inspect the build manually. The pet project only route is `src/pages/index.astro` and contains:

```tsx
---
import { Image } from 'astro:assets'
---

<html lang="en">
  <body>
    <Image
      src="http://placekitten.com/200/300"
      width={200}
      height={300}
      alt=""
    />
    <Image
      src="http://placekitten.com/200/300"
      width={200}
      height={300}
      alt=""
      transform="lazy" 
    />
  </body>
</html>
```
The acceptance criteria is that when running `npm run build` in static mode, both images are transformed, url is replaced by a `/_astro/` url and image is in the `dist/_astro` folder whereas in hybrid mode, only the 1st one is transformed and the 2nd one has a `/_image?href=` url instead.

## Note

I'm not really sure of what's the best way to code this. I wanted to avoid having the  `globalThis.astroAsset.isStaticImage` but there's a need to get the `settings.config.output` to know if we need to `addStaticImage` or not.

I first thought of modifying `addStaticImage` to `maybeAddStaticImage` but i preferred keeping separation of concern intact and a simple `addStaticImage` with no early exit for this specific case.

If the `config.output` could be brought to `getImage()` directly, there would be no need for this function and we could simply write `option.transform !== 'lazy'` instead.

I also was not sure of what should be the parameter name, should it be a boolean or not. I tried `skipTransformAtBuild` or `prerender` or else, but none were satisfactory when writing code with both `getImage` and `<Image />` so i thought it'd be nicer to follow the same interface than `loading`.

### DIFF
--- a/packages/astro/src/assets/services/service.ts
+++ b/packages/astro/src/assets/services/service.ts
@@ -213,6 +213,10 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 			options.format = DEFAULT_OUTPUT_FORMAT;
 		}
 
+		if (!options.transform) {
+			options.transform = 'eager';
+		}
+
 		// Sometimes users will pass number generated from division, which can result in floating point numbers
 		if (options.width) options.width = Math.round(options.width);
 		if (options.height) options.height = Math.round(options.height);
@@ -221,8 +225,18 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 	},
 	getHTMLAttributes(options) {
 		const { targetWidth, targetHeight } = getTargetDimensions(options);
-		const { src, width, height, format, quality, densities, widths, formats, ...attributes } =
-			options;
+		const {
+			src,
+			width,
+			height,
+			format,
+			quality,
+			densities,
+			widths,
+			formats,
+			transform,
+			...attributes
+		} = options;
 
 		return {
 			...attributes,
@@ -337,7 +351,7 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 		};
 
 		Object.entries(params).forEach(([param, key]) => {
-			options[key] && searchParams.append(param, options[key].toString());
+			key in options && searchParams.append(param, options[key].toString());
 		});
 
 		const imageEndpoint = joinPaths(import.meta.env.BASE_URL, '/_image');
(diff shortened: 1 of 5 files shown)


packages/astro/src/assets/services/service.ts
```diff
@@ -213,6 +213,10 @@ export const baseService: Omit<LocalImageService, 'transform'> = {
 			options.format = DEFAULT_OUTPUT_FORMAT;
 		}
 
+		if (!options.transform) {
```
y-nk, line 216: i thought of writing `options ??= 'eager'` but preferred to stick with `option.format` fashion

ematipico (changes requested):
I am going to block it because this is a new feature. Also, we should have a small RFC.

y-nk:
@ematipico i believe it's not a new feature, rather a fix on feature parity between astro/assets and astro core rendering strategies. here is the RFC which explains it better https://github.com/withastro/roadmap/discussions/760

ematipico:
We are going to close this for the time being. Even though it's a feature we would like to have in the future, the implementation still needs to be discussed. We can follow up in the proposed RFC, and see if there's enough community interest.

PR #8908, opened 2023-10-24
feat(i18n): better APIs

ematipico:
## Changes

This PR changes the APIs of the virtual module after some discussion in the [RFC](https://github.com/withastro/roadmap/pull/734). 

This PR changes the signature of the PRs, by passing a locale mandatory argument, and then a second argument: an object with a set of options.

This type of signature will allow us to add more options in the future.

The reason why this change was needed is that users might need:
- to add a path after the locale, without bothering with trailing slashes
- to add a path before the locale, without bothering with trailing slashes 
- opt-in the normalisation of the locale, not always needed

## Testing

I updated the tests

## Docs

This time I need feedback on the APIs, especially the documentation. I will update the RFC accordingly later

### DIFF
--- a/packages/astro/client.d.ts
+++ b/packages/astro/client.d.ts
@@ -135,7 +135,7 @@ declare module 'astro:i18n' {
 	 *
 	 * ## Errors
 	 *
-	 * It throws an error if the locale doesn't exist in the list of locales defined in the configuration.
+	 * Throws an error if the locale doesn't exist in the list of locales defined in the configuration.
 	 *
 	 * ## Examples
 	 *
@@ -160,7 +160,7 @@ declare module 'astro:i18n' {
 	 *
 	 * ## Errors
 	 *
-	 * It throws an error if the locale doesn't exist in the list of locales defined in the configuration.
+	 * Throws an error if the locale doesn't exist in the list of locales defined in the configuration.
 	 *
 	 * ## Examples
 	 *
@@ -180,14 +180,14 @@ declare module 'astro:i18n' {
 	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
 	 * @return {string[]}
 	 *
-	 * It works like `getLocaleRelativeUrl` but it emits the relative URLs for ALL locales:
+	 * Works like `getLocaleRelativeUrl` but it emits the relative URLs for ALL locales:
 	 */
 	export const getLocaleRelativeUrlList: (options?: GetLocaleOptions) => string[];
 	/**
 	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
 	 * @return {string[]}
 	 *
-	 * It works like `getLocaleAbsoluteUrl` but it emits the absolute URLs for ALL locales:
+	 * Works like `getLocaleAbsoluteUrl` but it emits the absolute URLs for ALL locales:
 	 */
 	export const getLocaleAbsoluteUrlList: (options?: GetLocaleOptions) => string[];
 }
--- a/packages/astro/client.d.ts
+++ b/packages/astro/client.d.ts
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
 declare module 'astro:i18n' {
 	type I18nModule = typeof import('./dist/i18n/index.js');
 
-	// TODO: documentation
-	export const getLocaleRelativeUrl: (locale: string) => string;
-	export const getLocaleAbsoluteUrl: (locale: string) => string;
+	export type GetLocaleOptions = import('./dist/i18n/index.js').GetLocaleOptions;
 
-	// TODO: documentation
-	export const getLocaleRelativeUrlList: () => string[];
-	export const getLocaleAbsoluteUrlList: () => string[];
+	/**
+	 * @param {string} locale A locale
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string}
+	 *
+	 * Returns a _relative_ path with passed locale.
+	 *
+	 * ## Errors
+	 *
+	 * It throws an error if the locale doesn't exist in the
+	 *
+	 * ## Examples
+	 *
+	 * ```js
+	 * import { getLocaleRelativeUrl } from "astro:i18n";
+	 * getLocaleRelativeUrl("es"); // /es
+	 * getLocaleRelativeUrl("es", { path: "getting-started" }); // /es/getting-started
+	 * getLocaleRelativeUrl("es", { path: "getting-started", prependWith: "blog" }); // /blog/es/getting-started
+	 * getLocaleRelativeUrl("es_US", { path: "getting-started", prependWith: "blog", normalizeLocale: true }); // /blog/es-us/getting-started
+	 * ```
+	 */
+	export const getLocaleRelativeUrl: (locale: string, options?: GetLocaleOptions) => string;
+
+	/**
+	 *
+	 * @param {string} locale A locale
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string}
+	 *
+	 * Returns an absolute path with the passed locale. The behaviour is subject to change based on `site` configuration.
+	 * If _not_ provided, the function will return a _relative_ URL.
+	 *
+	 * ## Errors
+	 *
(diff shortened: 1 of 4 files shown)


packages/astro/client.d.ts
```diff
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
...
+	/**
+	 * @param {string} locale A locale
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string}
+	 *
+	 * Returns a _relative_ path with passed locale.
+	 *
+	 * ## Errors
+	 *
+	 * It throws an error if the locale doesn't exist in the
```
martrapp, line 138: Incomplete sentence

packages/astro/src/i18n/index.ts
```diff
@@ -2,16 +2,31 @@ import { AstroError } from '../core/errors/index.js';
...
 	base: string;
 	locales: string[];
 	trailingSlash: AstroConfig['trailingSlash'];
 	format: AstroConfig['build']['format'];
 };
 
+export type GetLocaleOptions = {
+	/**
+	 * If the locale needs to be normalized
+	 */
```
natemoo-re, line 18: It would be nice to explain what "normalized" means in this context! Also, does this have a default value of `true` or `false`?

matthewp (approved):

martrapp (approved):
LGTM!

natemoo-re (approved):

lilnasy (approved):

packages/astro/src/i18n/index.ts
```diff
@@ -2,16 +2,31 @@ import { AstroError } from '../core/errors/index.js';
...
 	locale: string;
 	base: string;
 	locales: string[];
 	trailingSlash: AstroConfig['trailingSlash'];
 	format: AstroConfig['build']['format'];
 };
 
+export type GetLocaleOptions = {
+	/**
+	 * It makes the locale URL-friendly by replacing underscores with dashes, and making the locale lower case.
```
ElianCodes, line 17:
```suggestion
	 * Makes the locale URL-friendly by replacing underscores with dashes, and converting the locale to lower case.
```

packages/astro/client.d.ts
```diff
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
...
+	 * @param {string} locale A locale
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string}
+	 *
+	 * Returns an absolute path with the passed locale. The behaviour is subject to change based on `site` configuration.
+	 * If _not_ provided, the function will return a _relative_ URL.
+	 *
+	 * ## Errors
+	 *
+	 * It throws an error if the locale doesn't exist in the list of locales defined in the configuration.
```
ElianCodes, line 163:
```suggestion
	 * Throws an error if the locale doesn't exist in the list of locales defined in the configuration.
```

packages/astro/client.d.ts
```diff
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
...
+	 * getLocaleAbsoluteUrl("es_US", { path: "getting-started", prependWith: "blog", normalizeLocale: true }); // https://example.com/blog/es-us/getting-started
+	 * ```
+	 */
+	export const getLocaleAbsoluteUrl: (locale: string, options?: GetLocaleOptions) => string;
+
+	/**
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string[]}
+	 *
+	 * It works like `getLocaleRelativeUrl` but it emits the relative URLs for ALL locales:
```
ElianCodes, line 183:
```suggestion
	 * Works like `getLocaleRelativeUrl` but it emits the relative URLs for ALL locales:
```

packages/astro/client.d.ts
```diff
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
...
+	 */
+	export const getLocaleRelativeUrlList: (options?: GetLocaleOptions) => string[];
+	/**
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string[]}
+	 *
+	 * It works like `getLocaleAbsoluteUrl` but it emits the absolute URLs for ALL locales:
```
ElianCodes, line 190:
```suggestion
	 * Works like `getLocaleAbsoluteUrl` but it emits the absolute URLs for ALL locales:
```

packages/astro/client.d.ts
```diff
@@ -124,13 +124,72 @@ declare module 'astro:transitions/client' {
...
+	/**
+	 * @param {string} locale A locale
+	 * @param {import('./dist/i18n/index.js').GetLocaleOptions} options Customise the generated path
+	 * @return {string}
+	 *
+	 * Returns a _relative_ path with passed locale.
+	 *
+	 * ## Errors
+	 *
+	 * It throws an error if the locale doesn't exist in the list of locales defined in the configuration.
```
ElianCodes, line 138:
```suggestion
	 * Throws an error if the locale doesn't exist in the list of locales defined in the configuration.
```

packages/astro/src/i18n/index.ts
```diff
@@ -92,6 +113,9 @@ export function getLocaleAbsoluteUrlList({ site, ...rest }: GetLocaleAbsoluteUrl
  * - replaces the `_` with a `-`;
  * - transforms all letters to be lower case;
  */
-function normalizeLocale(locale: string): string {
+function normalizeTheLocale(locale: string, shouldNormalize: boolean): string {
+	if (!shouldNormalize) {
+		return locale;
+	}
 	return locale.replaceAll('_', '-').toLowerCase();
 }
```
bluwy, line 121: nit: I find passing `shouldNormalize` to determine whether the function is a no-op a bit strange of a pattern. And instead the callee should probably decide if they should call this function in the first place?

Not a big deal for me that needs changing but thought I point it out.

bluwy (approved):

ElianCodes (commented):
Looks good to me, but my gut feeling is to remove the 'it' from the start of the sentences.

packages/astro/src/i18n/index.ts
ematipico: Yeah, I was on the fence on this one. I didn't want to add too much branching in the callee, but I can address this comment in the next PR. Thank you

ElianCodes (approved):
LGTM! but you might want to also want a @sarah11918 review.

delucis (commented):
Nice work @ematipico! I think the docs looks OK.

One thing I’m wondering is if this feels good to use in practice.

Thoughts:

- While there is the edge case of `/` where `path` can be omitted, most uses will provide a path (or will provide one programmatically, so may provide one even when the variable’s value is `'/'` or `''`). That makes requiring `{ path: ... }` feel a bit fiddly.

- What are the most common sources of data people might be feeding into these methods? Should they support things like `getLocaleRelativeUrl('es', Astro.url)`? (i.e. support a `URL` type as input, not just `string`.)

- The names are all `get…Url()`, but the return type is a string rather than a `URL`. Instead of separate relative and absolute methods could we have a single method, which lets you grab `.pathname` for relative or `.href` for absolute?

TheOtterlord (commented):
Docs looks good to me

+1 on the function naming & URL input support.

ematipico:
>    While there is the edge case of / where path can be omitted, most uses will provide a path (or will provide one programmatically, so may provide one even when the variable’s value is '/' or ''). That makes requiring { path: ... } feel a bit fiddly.

I don't see any problem, we could change the signature of the API to `(locale, path, options)`

>    What are the most common sources of data people might be feeding into these methods? Should they support things like getLocaleRelativeUrl('es', Astro.url)? (i.e. support a URL type as input, not just string.)

>    The names are all get…Url(), but the return type is a string rather than a URL. Instead of separate relative and absolute methods could we have a single method, which lets you grab .pathname for relative or .href for absolute?

We could, but I am not sure about the flexibility of the APIs if we decided to have one single function and support for URLs. URLs work only with absolute URLs that have a protocol, and in many cases, users don't need it if they only need relative paths, or we don't have that information (`site` isn't provided). 

Essentially, if the user needs a relative path, and we only have one API, it's technically impossible to return a `URL` (or use one internally).

PR #8911, opened 2023-10-24
fix(create-astro): respect existing `package.json#scripts`

natemoo-re:
## Changes

- There was a regression introduced in https://github.com/withastro/astro/pull/8853 that replaced a template's existing `scripts` with just the new `build` script
- This PR fixes that case by mutating the `packageJsonData.scripts.build` instead of creating a new object.
- Changeset added!

## Testing

Test added. Fixture was updated to verify that this case is handled.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/tame-wasps-exercise.md
+++ b/.changeset/tame-wasps-exercise.md
@@ -1,0 +1,5 @@
+---
+'create-astro': patch
+---
+
+Ensure an existing template's `package.json` `scripts` are respected when modifying `build`.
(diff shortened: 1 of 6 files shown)


delucis (approved):
LGTM! Thanks for the quick work @natemoo-re 🚀

lilnasy (approved):
Thanks for the quick fix! The tests are probably going to be a lifesaver down the line.

PR #8918, opened 2023-10-25
feat(i18n): remove fallbackControl and use `routingStrategy`

ematipico:
## Changes

Requires https://github.com/withastro/astro/pull/8908 to be merged first

After some discussion in the RFC, `fallbackControl` is removed. I added a new option called `routingStrategy` that controls how the default locale routing behaves. 

At the moment, the only strategies that we can support are:
- `prefix-always`, where each route will have a prefix of the locale, even the default locale. A route that doesn't have a locale is a hard 404 (or error, to decide)
- `prefix-expect-default`, where a route that contains the default locale is automatically redirected to a locale-free route

## Testing

I created two fixtures, one for each variant of the strategy and updated their 

## Docs

Feedback appreciated around the `astro.ts` file


packages/astro/src/@types/astro.ts
```diff
@@ -1402,18 +1402,18 @@ export interface AstroUserConfig {
 
 			/**
 			 * @docs
-			 * @name experimental.i18n.fallbackControl
-			 * @type {"none" | "render" | "redirect"}
+			 * @name experimental.i18n.routingStrategy
+			 * @type {'prefix-always' | 'prefix-expect-default'}
```
sarah11918, line 1406: re: naming here: and, bear in mind I'm not entirely sure how this setting is used, so this is just about naming. So, I hope it's relevant! :sweat_smile: 

if there are only two options, and one is `prefix-always`, then `prefix-expect-default` seems not only like a bit of a mouthful, but doesn't itself communicate what "the other thing" is. I'm not even sure what to expect this to mean.

Is this only related to your default locale? or is it what happens for all languages? I could see 'prefix -always` meaning, "any language takes a prefix (so, no redirect necessary? if that's what this code is about?). To me, the opposite would be an idea like `default-root`/ `default-stripped` to indicate that your default local shouldn't ultimately be shown with a prefix, whatever your logic is doing here (internally you use one for routing to make life simpler, but it's stripped before the route is created?)

Is there also a reason this can't be a boolean for your default locale, to show the prefix or not? (I haven't read the code, so this may not be that! I'm only inferring things by looking at the names). If this is trying to convey "Does your default locale show a prefix or not?" it seems like that would be a straightforward config option for the user to set.

ematipico: Thank you, Sarah for the feedback

> if there are only two options,

There are two options **now** because they are the only ones we can support, but there's going to be at least a third one, which will involve domain support. It's not here now because I still need to implement it.

> Is this only related to your default locale?

No, it's related to all locales, depending on the option. Here's an example as a reference: https://i18n.nuxtjs.org/options/routing#strategy

The two options I implemented here are related to the default locale, but when we will add the third option, it will involve all the locales.

>  I could see prefix-always meaning, "any language takes a prefix" (so, no redirect necessary? if that's what this code is about?)

Yes. Routes that don't have a prefix will result into an error.

> To me, the opposite would be an idea like default-root/ default-stripped to indicate that your default local shouldn't ultimately be shown with a prefix, whatever your logic is doing here

I am more keen to use `default-stripped`, because "root" implies that the default locale must be somewhere, but this feature doesn't make restrictions on where the folders/prefixes are.

packages/astro/src/@types/astro.ts
```diff
@@ -1402,18 +1402,18 @@ export interface AstroUserConfig {
...
 			 * @version 3.*.*
 			 * @description
 			 *
-			 * Controls the fallback system of the internationalisation:
-			 *  - `none`: Astro will do nothing and will return `404` if a translated page isn't translated;
-			 *  - `redirect`: Astro will do a redirect to the fallback language if the translated page returns a `404`;
-			 *  - `render`: currently unsupported by Astro
+			 * Controls the routing strategy:
+			 *  - `prefix-expect-default`: This is the default value. When used, Astro will identify those routes that belong to the default locale and execute a redirect a route stripped of that locale.
+			 *  - `prefix-always`: When used,
```
matthewp, line 1413: missing an explanation

packages/astro/src/@types/astro.ts
matthewp: > Yes. Routes that don't have a prefix will result into an error.

You mean a 404, right?

sarah11918: Thanks for that link, and now I see what you mean.

Honestly I don't really like any of Nuxt's naming strategy there (and I swear they have 4 options that only describe 3 distinct possibilities, but their wording is so inconsistent I can't tell which are the overlapping two) , so you can ignore the following if you want to stick to theirs thinking that it's familiar to people, but maybe we can do better! :sweat_smile:  ...

The thing that strikes me is that `i18n routing strategy` doesn't *necessarily* have to be about prefixes, and makes naming specifically about just prefixes harder. (You could be "strategizing" based on any number of factors!)

If you asked me out of the blue, "What's your i18n routing strategy?" I don't think I would automatically think to say, "Oh, prefixes here but not there!" I might think, "Oh, use English file names everywhere as my routes, even when the page is translated!" for example.

I think the naming maybe gets easier if the thing you're trying to name is something more like:

`routePrefix: always | non-default | never` or something like that.

(non-default isn't great, could also be "other-locals" and then documentation defines default vs other as the two categories... but the entire concept is more meaningful, I think) This frees you from having to get "prefix" in the config value itself (and can be an indication that if multiple things are about this prefix, then maybe that's the the thing you really want to define). This also frees up "strategy" in case you want to use it for something else, but honestly, you can probably always just define the thing you're trying to define, without calling it a "strategy."

But, if you're set on keeping `routingStrategy` to only describe prefixes or not, maybe these options make sense?

- `prefix-always`
- `prefix-other-locales`
- `prefix-never`

matthewp: I think the reason that it's not specifically `routePrefix` is because in the future this option will also apply to subdomains (en.example.com). But maybe that still does work, could you call the `.en` a prefix?

ematipico: Yes, @matthewp, I mean 404; sorry for choosing poor words 🤣 

@sarah11918 I don't like Nuxt options, but I wanted to give you an example to help you understand what's this about. It's really about describing how the routing works, and it's not just about prefixes as Matthew pointed out. It's about "strategies", because subdomains will play a part in that too. 

If you prefer, we could keep `routePrefix`, but then we will need to make a matrix (table) to explain the behaviour of the routing when subdomains come into play.

sarah11918: Ah, nope, don't mind me if this isn't properly taking future plans into account! Future proofing is always better!

Like I said, without knowing what the code is doing, or going to do, the names didn't seem very descriptive (and even Nuxt's, which are finalized, don't sound all that great for what they do). BUT, if this is going to hold more eventually, then that is in fact what a more general word like "strategy" is for!

So then future options could include things like:
- use `example.com/lang/content/` for every lang
- use `example.com/lang/content/` for all non-default languages, but `example.com/content` for the default locale
- use `lang.example.com/content` for every lang
- use `lang.example.com/content` for all non-default languages, but `example.com/content` for the default locale
- use `example.com/content` always - never show a subdomain nor file path prefix for any language

sarah11918: Thanks, I was just confirming that I was understanding in that comment. Just to help make clear "If these are the possibilities you are planning for, then do the names you choose now make sense when those are all the options you have to represent?" 

Ultimately, something like `prefix-other-locales` or `prefix-secondary` I think is better than what Nuxt is using. If you have a (better) name for non-default, that would be great to use. I could see something explicitly stated like:

- `prefix-always`
- `prefix-other-locales`
- `subdomain-always`
- `subdomain-other-locales`

Or maybe using includes/excludes is also obvious? And now we don't need a name for non-default?

- `prefix-include-default`
- `prefix-exclude-default`
- `subdomain-include-default`
- `subdomain-exclude-default`

To be clear, you don't need my approval though! I'm just mentioning when things seem like they'd be difficult to infer from just the names, and giving you some alternatives to think about in case it makes you think of something even better!  :smile:

packages/astro/src/core/config/schema.ts
```diff
@@ -311,8 +311,10 @@ export const AstroConfigSchema = z.object({
 						fallback: z.record(z.string(), z.string()).optional(),
 						detectBrowserLanguage: z.boolean().optional().default(false),
 						// TODO: properly add default when the feature goes of experimental
-						fallbackControl: z.enum(['none', 'redirect', 'render']).optional(),
-					})
+                        routingStrategy: z
+                            .enum(['prefix-always', 'prefix-other-locales'])
```
Princesseuh, line 315: This is kinda unrelated to the PR, but what's the difference between `z.enum` and `z.union`? I notice we use both 🤔

Princesseuh (approved):
Code looks great to me from what I understand! I didn't follow all the discussions, but the API looks fun to use

packages/astro/src/@types/astro.ts
```diff
@@ -1420,20 +1420,22 @@ export interface AstroUserConfig {
...
+            /**
+             * @docs
+             * @name experimental.i18n.routingStrategy
+             * @type {'prefix-always' | 'prefix-other-locales'}
+             * @default {'prefix-other-locales'}
+             * @version 3.*.*
+             * @description
+             *
+             * Controls the routing strategy:
+             *  - `prefix-other-locales`: This is the default value. When used, Astro will identify those routes that belong to the default locale and execute a redirect a route stripped of that locale.
```
Princesseuh, line 1432:
```suggestion
             *  - `prefix-other-locales`: This is the default value. When used, Astro will identify those routes that belong to the default locale and execute a redirect to a route stripped of that locale.
```

packages/astro/src/core/config/schema.ts
ematipico: I am not an expert, I just checked their documentation. With unions you do `z.union([z.string, z.numer])`, and you support multiple types. `z.enum` accepts just strings.

matthewp (approved):

bluwy (approved):

Issue #8224, opened 2023-08-25
Do not watch ignored files


FDiskas:
### Astro info

```block
Astro version            v2.10.11
Package manager          npm
Platform                 darwin
Architecture             arm64
Adapter                  Couldn't determine.
Integrations             @astrojs/tailwind, relative-links
```

### What browser are you using?

chrome

### Describe the Bug

![image](https://github.com/withastro/astro/assets/468006/e83825b8-077d-46d0-8b17-c667f17b72a7)

### What's the expected result?

It should not reload the project if some ignored file is updated.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-jqwvp1?file=.DS_Store&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

FDiskas:
I believe that astro could use gitignore file or have his own

natemoo-re:
Hmm, yeah we might need to read the user's `.gitignore` file and then filter out updates. However, things like `node_modules` are traditionally ignored by git but should still affect HMR, so I don't know what the right answer is here.

cornedor:
I have another use case. I'm using ReScript to compile some scripts. When ReScript's watcher is running, it will create a `.bsb.lock` file that is quickly removed. Sometimes so quickly that the watcher will pick it up, but the file will be removed by the time it is handled, causing the Astro watcher to crash.

ematipico:
The `.gitignore` support should be opt-in to avoid unpleasant behaviours.

cornedor:
Perhaps for some cases, it is better to add a configuration option to disable watcher for a list (glob?) of files?

ematipico:
> Perhaps for some cases, it is better to add a configuration option to disable watcher for a list (glob?) of files?

You can customise `vite` behaviour directly form your `astro.config.mjs`

```js
export default defineConfig({
	vite: {
		server: {
			watch: {
				
			}
		}
	}
})
```

Here's the options: https://vitejs.dev/config/server-options.html#server-watch

rishi-raj-jain:
@natemoo-re Per the solution suggested by @ematipico, I think we don't need a resolution for this. IMO it should be done in vite as in the comment above.

bluwy:
[...]

PR #8922, opened 2023-10-26
bugfix: don't watch .gitignore

guobin211:
## Changes

- https://github.com/withastro/astro/issues/8224


packages/astro/src/core/ignore-files.ts
```diff
@@ -0,0 +1,28 @@
...
+export async function ignoreFiles(root: string) {
+	const res: string[] = [];
+	const gitIgnoreFile = joinPaths(root, '.gitignore');
+	const text = await fs.promises
+		.readFile(gitIgnoreFile, 'utf-8')
+		.then((data) => data.toString())
+		.catch(() => '');
+	
+	if (text.length > 0) {
+		const lines = text.split('\n');
```
ematipico, line 17: In windows, you might not have `\n` but you have `\r` or `\r\n`. Use [EOL](https://nodejs.org/dist/latest-v20.x/docs/api/os.html#oseol) instead

packages/astro/src/core/ignore-files.ts
```diff
@@ -0,0 +1,28 @@
...
+		for (const line of lines) {
+			if (!line.startsWith('#')) {
+				const value = line.trim();
+				if (value.length) {
+					res.push(`**/${value}/**`);
```
ematipico, line 22: Doesn't this change the semantics of the paths inside `.gitignore`? If I have a `file.js` inside `.gitignore`, this will become `**/file.js/**`, which doesn't seem to be a valid pattern. I think we should add the paths as they are without tempering them.

packages/astro/src/core/ignore-files.ts
```diff
@@ -0,0 +1,28 @@
+import fs from 'node:fs';
+import { joinPaths } from './path.js';
+
+/**
+ * use .gitignore file to ignore files
+ * @param root astro root path
+ */
+export async function ignoreFiles(root: string) {
```
ematipico, line 8: We could call this function `getGitIgnorePaths`, which is more in line with what the function does

ematipico (changes requested):

bluwy:
The issue mentioned that respecting `.gitignore` isn't correct?
- https://github.com/withastro/astro/issues/8224#issuecomment-1723044876
- https://github.com/withastro/astro/issues/8224#issuecomment-1701808918

packages/astro/src/core/ignore-files.ts
guobin211: fixed

name: getGitignorePaths
file: https://github.com/guobin211/astro/blob/main/packages/astro/src/core/ignore-files.ts

packages/astro/src/core/ignore-files.ts
guobin211: fixed

https://github.com/guobin211/astro/blob/bugfix/watch-ignore-files/packages/astro/src/core/get-gitignore-paths.ts

natemoo-re:
Based on @bluwy's comment above, I think it's safe to assume that we aren't comfortable with the approach this PR uses. We'll have to come up with another solution here, likely plugging into Vite more directly than we do now.

PR #8929, opened 2023-10-26
fix(slots): consume eagerly rendered slot after one use

### LINKED ISSUE #7068: Different behavior with slot vs Astro.slots.render

lilnasy:
## Changes
- Fixes #7068 

## Testing
Added a case to an existing fixture.

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/plenty-spiders-act.md
+++ b/.changeset/plenty-spiders-act.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixes an issue where a slot rendered multiple times resulted in the same output each time.
+Fixes an issue where rendering the same slot multiple times invoked it only once.
--- a/.changeset/plenty-spiders-act.md
+++ b/.changeset/plenty-spiders-act.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue where a slot rendered multiple times resulted in the same output each time.
--- a/packages/astro/src/runtime/server/render/astro/instance.ts
+++ b/packages/astro/src/runtime/server/render/astro/instance.ts
@@ -32,13 +32,13 @@ export class AstroComponentInstance {
 		this.slotValues = {};
 		for (const name in slots) {
 			// prerender the slots eagerly to make collection entries propagate styles and scripts
-			let value: ReturnType<typeof slots[string]> | null = slots[name](result);
+			let didRender = false;
+			let value = slots[name](result);
 			this.slotValues[name] = () => {
 				// use prerendered value only once
-				if (value !== null) {
-					const consumedValue = value
-					value = null
-					return consumedValue
+				if (!didRender) {
+					didRender = true;
+					return value;
 				}
 				// render afresh for the advanced use-case where the same slot is rendered multiple times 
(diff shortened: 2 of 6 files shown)


.changeset/plenty-spiders-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue where a slot rendered multiple times resulted in the same output each time.
```
sarah11918, line 5:
```suggestion
Fixes an issue where rendering multiple slots resulted in the same output each time.
```
super-nit: "a slot" rendered multiple times makes me think it's the same one slot, so the output should be the same! Would this be a correct description? (non-blocking, though)

lilnasy: It is the same one slot.

lilnasy: Now that you bring it up though, I think there is room for clarification. Before this PR, the first render was reused when users tried to manually render the same slot multiple times.

packages/astro/src/runtime/server/render/astro/instance.ts
```diff
@@ -31,8 +31,18 @@ export class AstroComponentInstance {
...
+			let value: ReturnType<typeof slots[string]> | null = slots[name](result);
+			this.slotValues[name] = () => {
+				// use prerendered value only once
+				if (value !== null) {
+					const consumedValue = value
+					value = null
+					return consumedValue
+				}
+				// render afresh for the advanced use-case where the same slot is rendered multiple times 
+				return slots[name](result);
```
natemoo-re, line 44: This code definitely works, but I feel like a simple toggle might be easier to understand? Something like...

```js
let didRender = false;
let value = slots[name](result);
this.slotValues[name] = () => {
  if (!didRender) {
		didRender = true;
    return value;
  }
  return slots[name](result);
```

natemoo-re (approved):
Implementation LGTM! I think it could be refactored to be slightly more clear, but it's up to you. I wouldn't block a merge as-is.

packages/astro/src/runtime/server/render/astro/instance.ts
lilnasy: Much cleaner, thanks!

matthewp (approved):

.changeset/plenty-spiders-act.md
sarah11918: OK, then maybe something like.... (wordy, but trying to see if this is the right idea)

> Fixes an issue where a slot rendered multiple times incorrectly used the same initial rendering for each instance

sarah11918 (approved):
LGTM!

Issue #8921, opened 2023-10-25
Slot name is being rendered as an attribute on Vue component


spacedawwwg:
### Astro Info

```block
Astro                    v3.3.4
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          yarn
Output                   server
Adapter                  none
Integrations             @astrojs/vue
                         astro-purgecss
                         astro-compress
```

### Describe the Bug

When using a slot name on a Vue component, the slot name is being rendered as an attribute

something like:
```VUE
<Parent>
  <Child slot="footer" />
</Parent>
```
renders like:
```HTML
<div class="parent"> 
  <div class="parent-footer">
    <div class="child" slot="footer">I am the child</div>
  </div>
</div>
```
<img width="495" alt="image" src="https://github.com/withastro/astro/assets/1691867/e5b3c55e-8146-41fc-936c-86556b02ca0c">

### What's the expected result?

No slot attribute rendered on component

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-1jemj8

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Ah, good catch, probably a quick fix, let me take a look.

PR #8930, opened 2023-10-26
Prevent passing slot names as props

matthewp:
## Changes

- Ignore the `slot` prop in Vue components
- Fixes https://github.com/withastro/astro/issues/8921

## Testing

- Test case added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/swift-suits-drum.md
+++ b/.changeset/swift-suits-drum.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/vue': patch
+---
+
+Fixes an issue where Astro slot names were being rendered as attributes in components. Astro slot names will no longer be sent as props to framework components.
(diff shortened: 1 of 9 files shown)


natemoo-re (approved):
Works for me!

.changeset/swift-suits-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent passing slot names as props
```
sarah11918, line 5:
```suggestion
Fixes an issue where Astro slot names were being rendered as attributes in components. Astro slot names will no longer be sent as props to framework components.
```
 Just tried to tie it into the issue reported. Something like this?

sarah11918 (commented):
Just a suggestion to tie the changeset description back to the behaviour reported in the original issue!

matthewp:
This actually might be wrong, putting back into draft.

matthewp:
I had to go with a Vue-specific solution because we don't know what slots are going to be used inside of the component providing the slots.

PR #8940, opened 2023-10-28
fix: toStyleString omit nullish values

### LINKED ISSUE #8933: Property with nullish value in inline style should be omitted
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

In JSX, `<div style={{ backgroundColor: condition ? '#d6f9dd' : undefined }} />` will not produce any style. But is astro template, this will produce `style="background-color:undefined"`

behave as convention in JSX

https://stackblitz.com/edit/github-47zegu?file=src%2Fpages%2Findex.astro

### Participation

MarvinXu:
## Changes
fix #8933 . 

- What does this change?

## Testing
I've tested this utility function locally with following inputs:
```javascript
obj = {fontSize: '10px', color: undefined} // -> 'font-size:10px'
obj = {fontSize: null, color: 'red'} // -> 'color:red'
```

## Docs
No need to change as this is for compat with JSX


packages/astro/src/runtime/server/render/util.ts
```diff
@@ -28,6 +28,7 @@ const kebab = (k: string) =>
 	k.toLowerCase() === k ? k : k.replace(/[A-Z]/g, (match) => `-${match.toLowerCase()}`);
 const toStyleString = (obj: Record<string, any>) =>
 	Object.entries(obj)
+		.filter(([k, v]) => v !== undefined && v !== null)
```
Princesseuh, line 31: I'm not familiar with JSX, does it ignore all falsy values or just nullish? (and is there a reason to accept falsy values?)

MarvinXu: I can't find its source code, so I did a little more [tests](https://reactplayground.vercel.app/#N4IglgdgJgpgHgOgFYGcQC5wFsAOB7AJwBcACAQRxxIDMC8sSByBAegp2RTkYB0IxchUgCUYAQwDGpWvSYFxU3v0HESwEhPliiMYXjykAvjToNGWqQFoo9FhIA2YGBCJK+EvBBSkPLsZBgCEgBeEhsJAFcsZyIEAHMYIgBRexholwAhAE8ASSgACnN9VwBKd09vEjoDEI0tHT0DfN8ifwhAsohq2PloQPy+EhJRSVjNcR0UtJiBiCGhkakEAGUiAjApAFk8WAAaQfmICPt7fbn5xbH6mCn0onz2TqHOkpBdkHZOOAwQPgF8VSXEyycwKVzlLykdi1fIlEIAPjUB3kRAiBDms3mJAAPAALACMJG8WVSwWAwAOWJYLBoniIyzAAC8YOgSAAGSnzDz2QisxhMM5Yobc3nswVCkUEVlrCIwcVYyWs6hiewoOWc4V4HlSpgkMQAIwkjHl831kgA1nE6BFoABhLWi-kkLIwE54ADuAo1JEc7QAEjAwHFcUQ+QajZzDIZ4QG3diWAT4QdOoY+Gn2nAAaRYMrjlDKCBDEA).

It seems to have more complicate validation rules than just nullish or falsy values.
```jsx
<h1 style={{
      // falsy value `0` is preserved in fontSize
      // fontSize: 0
     // color omits any value that is not a `color`
      color: ' ',
      color: 0,
      color: true,
      color: false,
      color: ' abc',
      backgroundColor: '  yellow ',
      lineHeight: 'abc'
    }}>Hello</h1>
```

ematipico, line 31: Can't we use `.filter(Boolean)`?

MarvinXu: https://github.com/withastro/astro/pull/8940#discussion_r1376193061 See my test. `font-size: 0` is a valid value

MarvinXu:
Making a complete validation like JSX did (https://github.com/withastro/astro/pull/8940#discussion_r1376193061) requires a lot more effort. So I think we can stick with a quick and simple fix to solve common use cases like:
```jsx
// image may be undefined
<div style={{ backgroundImage: image && `url(${image.src})`}}></div>
```

I updated my code to only accept non-empty string and number

bluwy (approved):

ematipico (approved):

natemoo-re (commented):
Just for clarity, this is implementation-specific behavior that isn't covered in the JSX spec. So the more complex behavior is just what React chooses to do.

I definitely agree that we should filter out `undefined`, `null`, and `false` though. Trying to think of any cases where we'd want to support anything other than strings and numbers, though.

Issue #8926, opened 2023-10-26
A way to disable prefetch with ViewTransitions


fuwaneko:
### Astro Info

```block
Astro                    v3.4.0
Node                     v20.8.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/tailwind
                         @astrojs/solid-js
```

### Describe the Bug

I have an app with cookie-based authentication and a sign-out link. When ViewTransitions are enabled, hovering over the sign-out link causes the sign-out per se. There should be a way to tell Astro to disable prefetch on some links, adding "data-astro-reload" to a link doesn't prevent prefetching.

### What's the expected result?

A data attribute to disable prefetch, perhaps?

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hmcjqs

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
We are in the process of updating prefetch usage in general and plan to have a solution for this as part of it.

matthewp:
An alternative is to use a `<form>` with a `POST` request. Usually you don't do mutations inside of a GET.

bis0072:
I'm having the same issue

PR #8951, opened 2023-10-30
Support prefetch in core

bluwy:
## Changes

close https://github.com/withastro/astro/issues/8926

Add new `prefetch` option. Supersedes `@astrojs/prefetch`. Check the [docs](https://github.com/withastro/docs/pull/5271) and RFC for more information on how it works.

- RFC: https://github.com/withastro/roadmap/pull/755
- NOTE: The `@astrojs/prefetch` integration will be deprecated after this is merged and released.

### Tip for reviewing

- The core logic changes are in `src/core`. `src/core/prefetch/index.ts` contains the client-side script.

- The documentation is in the changeset and `src/@types/astro/ts`. They take a lot of lines, and the docs for `prefetch.prefetchAll` and `prefetch.defaultStrategy` are duplicated. I think it's needed for docs to detect it.

## Testing

Added e2e tests to make sure a request is sent

## Docs

Main docs: https://github.com/withastro/docs/pull/5271

This PR also has changesets and JSDoc for review.


martrapp:
LVGTM: I am very happy that the different prefetch approaches are unified here and the special handling in ViewTransition.astro is replaced by core functions!

packages/astro/src/@types/astro.ts
```diff
@@ -536,6 +536,41 @@ export interface AstroUserConfig {
...
+	 */
+
+	/**
+	 * @docs
+	 * @name prefetch.prefetchAll
+	 * @type {boolean}
+	 * @description
+	 * Enable prefetching for all links by default, similar to all `<a />` links having the `data-astro-prefetch`
+	 * attribute set. You can disable prefetching individually by setting `data-astro-prefetch="false"` on a link.
+	 */
```
ematipico, line 561: Are these blocks intended? They are the same as the ones of `PrefetchOptions`

packages/astro/src/prefetch/index.ts
```diff
@@ -0,0 +1,260 @@
...
+ * Initialize the prefetch script, only works once
+ */
+export function init(opts?: InitOptions) {
+	if (!inBrowser) return;
+
+	prefetchAll ??= opts?.prefetchAll ?? false;
+	defaultStrategy ??= opts?.defaultStrategy ?? 'hover';
+
+	// Init only once
+	if (inited) return;
```
ematipico, line 33: Is it intended? Changing the options before the early return. If so, I think we should leave a comment for future developers, to give context why it's like this.

packages/astro/src/prefetch/index.ts
```diff
@@ -0,0 +1,260 @@
+/*
+  NOTE: Do not add any dependencies or imports in this file so that it can load quickly in dev.
+*/
+
+// eslint-disable-next-line no-console
+const debug = import.meta.env.DEV ? console.debug : undefined;
```
ematipico, line 6: Out of curiosity, do we set `DEV` when using the dev server? Or are there specific astro options?

ematipico (approved):

packages/astro/src/prefetch/index.ts
matthewp: It's from Vite.

packages/astro/src/prefetch/index.ts
```diff
@@ -0,0 +1,260 @@
...
+const inBrowser = import.meta.env.SSR === false;
+const prefetchedUrls = new Set<string>();
+const listenedAnchors = new WeakSet<HTMLAnchorElement>();
+
+// User-defined config for prefetch
+// @ts-expect-error injected by vite-plugin-prefetch
+let prefetchAll: boolean = __PREFETCH_PREFETCH_ALL__;
```
matthewp, line 13: Can you explain how these variables are getting set?

packages/astro/src/prefetch/index.ts
```diff
@@ -0,0 +1,260 @@
...
+function initTapStrategy() {
+	for (const event of ['touchstart', 'mousedown']) {
+		document.body.addEventListener(
+			event,
+			(e) => {
+				if (elMatchesStrategy(e.target, 'tap')) {
+					prefetch(e.target.href, { with: 'fetch' });
+				}
+			},
+			{ passive: true, once: true }
```
matthewp, line 56: once? So after the first tap we no longer prefetch anything? I am probably misunderstanding....

packages/astro/src/prefetch/index.ts
ematipico: Yeah I know, I am asking when it's set :) during the `dev` phase only?

packages/astro/src/@types/astro.ts
bluwy: Yeah it's intentionally duplicated as I think that's how the docs will scan it. The other properties are also duplicated.

packages/astro/src/prefetch/index.ts
bluwy: Yeah it's intended so if you init with different options, they will change. It's written defensively so if both the inject script and View transitions call `init()`, the View transition's custom default will always kick in.

I'll add addition comments here.

packages/astro/src/prefetch/index.ts
bluwy: Yup. But specifically if `NODE_ENV !== 'production'`, which should always be the case in dev. https://github.com/vitejs/vite/blob/ac964de760a455838aeeb5a1e197e5ab77731f59/packages/vite/src/node/config.ts#L767

packages/astro/src/prefetch/index.ts
bluwy: Oh good catch. I thought this was listening on individual links. Thanks I'll remove this.

packages/astro/src/prefetch/index.ts
bluwy: Actually it doesn't make sense the more I think about it. The first `init()` would've set a value which makes the second call and `??=` not work anymore. I'll move this under the `inited` check for now.

packages/astro/src/@types/astro.ts
```diff
@@ -536,6 +536,41 @@ export interface AstroUserConfig {
...
+	 * @type {boolean | PrefetchOptions}
+	 * @description
+	 * Enable prefetching for your site. A prefetch script is automatically added to every page
+	 * in the project. Just add the `data-astro-prefetch` attribute to any `<a />` links on your page
+	 * and you're ready to go!
+	 *
+	 * If you use [View Transitions](https://docs.astro.build/en/guides/view-transitions/), prefetch will
+	 * be enabled automatically with a default of `{ prefetchAll: true }`, which enables prefetching for all links.
+	 * You can configure this option to override the defaults, and even set `false` to disable prefetching
+	 * completely when using View Transitions.
```
sarah11918, line 551:
```suggestion
```
I *think* we can get away without this here. I mentioned up top that it's enabled, and I'll do the same in `prefetchAll`, and we'll mention it both in the guide, and go back and add a note in View Transitions! I think we can keep this config page moving along and still not confuse people!

packages/astro/src/@types/astro.ts
```diff
@@ -536,6 +536,41 @@ export interface AstroUserConfig {
...
+	 */
+
+	/**
+	 * @docs
+	 * @name prefetch.prefetchAll
+	 * @type {boolean}
+	 * @description
+	 * Enable prefetching for all links by default, similar to all `<a />` links having the `data-astro-prefetch`
+	 * attribute set. You can disable prefetching individually by setting `data-astro-prefetch="false"` on a link.
```
sarah11918, line 560:
``````suggestion
	 * Enable prefetching for all links, including those without the `data-astro-prefetch` attribute.
	 * This value defaults to `true` when using the `<ViewTransitions />` router. Otherwise, the default value is `false`.
	 * 
	 * ```js
	 * prefetch: {
	 * 	prefetchAll: true
	 * }
	 *
	 * When set to `true`, you can disable prefetching individually by setting `data-astro-prefetch="false"` on any individual links.
	 *
	 * ```html
	 * <a href="/about" data-astro-prefetch="false">About</a>
	 *```
``````
I'm thinking something like this? What do you think?

File: packages/astro/src/@types/astro.ts [hunk 8]
```diff
@@ -536,6 +536,41 @@ export interface AstroUserConfig {
...
+	 */
+
+	/**
+	 * @docs
+	 * @name prefetch.defaultStrategy
+	 * @type {'tap' | 'hover' | 'viewport'}
+	 * @default `'hover'`
+	 * @description
+	 * The default prefetch strategy to use when the `data-astro-prefetch` attribute (no value) is set on a link.
+	 * You can select a different strategy for individual links by setting, e.g. `data-astro-prefetch="viewport"`.
```
sarah11918, line 570:
```suggestion
	 * The default prefetch strategy to use when the `data-astro-prefetch` attribute is set on a link with no value.
	 * 
	 * - `'tap'`: Prefetch just before you click on the link.
	 * - `'hover'`: Prefetch when you hover over or focus on the link. (default)
	 * - `'viewport'`: Prefetch as the links enter the viewport.
	 * 
	 * You can override this default value and select a different strategy for any individual link by setting a value on the attribute.
	 *
	 * ```html
	 *  `<a href="/about" data-astro-prefetch="viewport">About</a>`
	 * ```
```
I know this is all in the guide, but when there are options, I do think it makes sense to have them in reference for quick "lookupability"!

packages/astro/src/@types/astro.ts
```diff
@@ -2336,3 +2371,25 @@ declare global {
 		'astro-dev-overlay-highlight': DevOverlayHighlight;
 	}
 }
+
+export interface PrefetchOptions {
```
sarah11918, line 2375: I honestly have no idea about this, because when I edit docs in this file, I don't remember ever having to edit them twice before? I am very bad at JSDoc. Someone help.

If these should show up twice, then I guess they should say the same thing? But this seems suspect to me.

sarah11918 (commented):
Love this, @bluwy and I'm also getting into the docs PR for this too!

I left some suggestions, *mostly* about getting explicit options shown etc. as this is a reference page and people might be coming here to look something up or check on it quickly. Duplication of content here and in the guide isn't always a bad thing, since the guide will walk through more and this can be for people who already pretty much know what the feature is about, they just don't know for example all the possible hover, tap etc. 

Also, I do *not* know what's up with documenting the options twice. We're going to need to call in someone else to confirm that.

packages/astro/src/@types/astro.ts
bluwy: Some appeared twice like `server.host` too, and presumably the docs scanner is only able to scan `AstroUserConfig` and not the other types it depends on.

But I went ahead to fix this since it doesn't look like we need a special type for `prefetch`. I inlined the type into `AstroUserConfig` so we only have to document it once.

sarah11918: I trust you!  💜

sarah11918 (approved):
Docs look great!

PR #8953, opened 2023-10-30
feat(underscore-redirects): add `base` to input paths

alexanderniebuhr:
## Changes

- adds base path as prefix for input paths

## Testing

- tested manually calling the function

## Docs

- changeset only

### DIFF
--- a/.changeset/odd-maps-glow.md
+++ b/.changeset/odd-maps-glow.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/underscore-redirects': patch
+---
+
+Adds the base path as prefix for input paths
--- a/packages/underscore-redirects/src/astro.ts
+++ b/packages/underscore-redirects/src/astro.ts
@@ -12,7 +12,7 @@ function getRedirectStatus(route: RouteData): ValidRedirectStatus {
 }
 
 interface CreateRedirectsFromAstroRoutesParams {
-	config: Pick<AstroConfig, 'build' | 'output'>;
+	config: Pick<AstroConfig, 'build' | 'output' | 'base'>;
 	/**
 	 * Maps a `RouteData` to a dynamic target
 	 */
@@ -28,6 +28,12 @@ export function createRedirectsFromAstroRoutes({
 	routeToDynamicTargetMap,
 	dir,
 }: CreateRedirectsFromAstroRoutesParams) {
+	const base =
+		config.base && config.base !== '/'
+			? config.base.endsWith('/')
+				? config.base.slice(0, -1)
+				: config.base
+			: '';
 	const output = config.output;
 	const _redirects = new Redirects();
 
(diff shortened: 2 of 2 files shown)


alexanderniebuhr:
* main
  * **PR #8953** <a href="https://app.graphite.dev/github/pr/withastro/astro/8953" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/withastro/astro/8953?utm_source=stack-comment).

packages/underscore-redirects/src/astro.ts
```diff
@@ -39,7 +45,7 @@ export function createRedirectsFromAstroRoutes({
 				// from the user if provided.
 				_redirects.add({
 					dynamic: false,
-					input: route.pathname,
+					input: `${base}${route.pathname}`,
```
lilnasy, line 48: Could this result in double slashes when `trailingSlashes` is also enabled?

lilnasy: Seems like that's accounted for 👍

florian-lefebvre:
Is that related to https://github.com/withastro/adapters/issues/45 ?

natemoo-re (approved):
Works for me!

alexanderniebuhr:
> Is that related to [withastro/adapters#45](https://github.com/withastro/adapters/issues/45) ?

@florian-lefebvre yes it should fix it _(hopefully)_

packages/underscore-redirects/src/astro.ts
```diff
@@ -94,15 +100,15 @@ export function createRedirectsFromAstroRoutes({
 				}
 				_redirects.add({
 					dynamic: true,
-					input: pattern,
+					input: `${base}${route.pattern}`,
```
florian-lefebvre, line 103: what's the difference with `pattern` on line 89?

.changeset/odd-maps-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/underscore-redirects': minor
```
bluwy, line 2: Since this package is v0, should this be a patch instead? Seems like it should be non-breaking.

ematipico (approved):
The fact that we don't have tests **concerns** me. We really should!

I am OK with merging it, but the team should add some infra to test these utilities.

packages/underscore-redirects/src/astro.ts
alexanderniebuhr: `pattern` is for dynamic routes

alexanderniebuhr:
> the team should add some infra to test these utilities.

I agree, just created https://github.com/withastro/astro/issues/8960, so it does not get out of sight.

.changeset/odd-maps-glow.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Adds the base path in front for input paths
```
sarah11918, line 5:
```suggestion
Adds the base path as prefix for input paths
```
I like the way you said it in the PR description better.  😄  I think "prefix" holds more meaning in context.

sarah11918 (commented):
Left my tiny suggestion!  🙌

Issue #8143, opened 2023-08-18
Exclude original images from output dir


pkolt:
### What version of `astro` are you using?

2.10.9

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

Hi guys!

I'm using Astro with the [assets module](https://docs.astro.build/en/guides/assets/).

After executing the command - `astro build`, my cache folder contains image thumbnails.

But the original images are also in the cache folder. 

Is it possible to add some setting to adjust Astro to remove original images?

The original images weigh a lot and slow down the deployment process to github pages.

### What's the expected result?

There are no original images in the cache folder.

### Link to Minimal Reproducible Example

https://github.com/pkolt/pkolt.ru

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
What do you mean by "your cache folder"? Do you mean the build output or the cache folder in `node_modules`?

pkolt:
Hi @Princesseuh !

Yes, the cache folder in `node_modules`. 
My cache folder is `./cache`, this is specified in my config:

```js
// astro.config.mjs
import { defineConfig, sharpImageService } from 'astro/config';
import react from '@astrojs/react';
import sitemap from '@astrojs/sitemap';
import mdx from '@astrojs/mdx';

/** @type {import('astro/config').AstroUserConfig} */
const config = {
  site: 'https://pkolt.ru',
  integrations: [react(), sitemap(), mdx()],
  compressHTML: true,
  markdown: {
    gfm: true,
    syntaxHighlight: 'prism',
  },
  outDir: 'out',
  cacheDir: 'cache', // <---- CACHE FOLDER
  experimental: {
    assets: true,
  },
  image: {
    service: sharpImageService(),
  },
};

export default defineConfig(config);

```

Princesseuh:
Hmm, we don't cache the original images, only the transformed version. Are you sure you're seeing the original files?

pkolt:
Hi @Princesseuh !

Yes, you can see for yourself.

```
git clone git@github.com:pkolt/pkolt.ru.git
cd pkolt.ru
npm ci
npm run build
```

For example, I have a large image file (3.4Mb) - `src/assets/blog/overview-wemos-d1-mini/wemos_d1_mini_1.jpg`

[...]

PR #8954, opened 2023-10-30
feat(assets): Delete original assets unused outside of the optimization pipeline

Princesseuh:
## Changes

What the title says. On the technical level, this works by making images Proxies, whenever any property of an image is referred to directly we add the image to a list of images to not delete.

Fix https://github.com/withastro/astro/issues/8143

## Testing

Tests should pass!

## Docs

N/A, we didn't document that they stayed there before.


matthewp (approved):
clever

packages/astro/src/assets/internal.ts
```diff
@@ -90,6 +90,15 @@ export async function getImage(
 				: options.src,
 	};
 
+	// Clone the `src` object if it's an ESM import so that we don't refer to any properties of the original object
+	// Causing our generate step to think the image is used outside of the image optimization pipeline
+	const clonedSrc = isESMImportedImage(resolvedOptions.src)
+		? // @ts-expect-error - clone is a private, hidden prop
+		  resolvedOptions.src.clone ?? resolvedOptions.src
```
natemoo-re, line 97: I believe you can use an `/** @internal */` comment to have erasable types for `ImageMetadata`. Properties marked with `@internal` won't be included when `.d.ts` files are generated, assuming [`stripInternal`](https://www.typescriptlang.org/tsconfig#stripInternal) is enabled.

natemoo-re (approved):
LGTM! Very nice use of `Proxy`. You _could_ get rid of the `// @ts-expect-error` but it's not a huge deal.

packages/astro/src/assets/internal.ts
Princesseuh: Oh, I didn't know that, that's neat. Will do that

.changeset/nasty-elephants-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Delete image original that are not used outside of the optimization pipeline
```
sarah11918, line 5:
```suggestion
Updates the Image Services API to now delete original images from the final build that are not used outside of the optimization pipeline. For users with a large number of these images (e.g. thumbnails), this should reduce storage consumption and deployment times.
```
Is this correct?  😅 Just adding a bit more here because this is a minor and Matthew might want a quick line for the blog post content.

sarah11918 (commented):
Just turned this into something that could go in the blog post!

.changeset/nasty-elephants-provide.md
Princesseuh: It is correct! The only nit is that it's not the image service API that does it, but it's around there anyway, so, it's not too bad

PR #8958, opened 2023-10-31
feat(i18n): rework preferred language and new preferredLanguageList

ematipico:
## Changes

This PR adds two new features to i18n routing:
- new property to the `Astro` global, called `preferredLocaleList`, which represents a sorted list of the locales requested by the browser. This list is filtered with `i18n.locales`. 
- reviewed how `preferredLocale` is computed; now it takes into consideration `i18n.locales`

Here are the changes reflected to the RFC: https://github.com/withastro/roadmap/pull/734/commits/e07aef715688703ebcadf297c3f5ed543ed2c14c

## Testing

I added new test cases to cover the new feature.

## Docs

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1436,32 +1436,6 @@ export interface AstroUserConfig {
              *
              */
             routingStrategy: 'prefix-always' | 'prefix-other-locales';
-
-			/**
-			 * @docs
-			 * @name experimental.i18n.redirectToPreferredLanguage
-			 * @type {boolean}
-			 * @default {false}
-			 * @version 3.*.*
-			 * @description
-			 *
-			 * If enabled, Astro will redirect the user to their preferred language. The preferred language is determined by
-			 * reading the header `Accept-Language`, usually sent by the browser.
-			 *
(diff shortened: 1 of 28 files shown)


matthewp (approved):

delucis (commented):
Noting here that @ematipico and I [discussed this a little bit this morning](https://discord.com/channels/830184174198718474/1055240277297733662/1168862833983885392) and I voiced concern that the way this currently behaves would make it easy for Astro developers to go against best practices by overriding user intent:

> I think it’s not an uncommon feature for people to _assume_ they want but there’s a lot of complexity. For example, multilingual sites should offer a language switcher. With the current design enabled, links to the default language in the switcher would no longer work. Generally, if you are going to rely on automatic redirects, you need to do so in combination with cookies so you can manually track user intent more accurately and perhaps have a header to avoid auto-redirects for internal navigations (i.e. you want to perhaps redirect a user on first visit so when they first type `business.com` they end up at `business.com/<preferred-locale>`, but then remember and respect subsequent user decisions about preferred language).
>
> Maybe making it easy to know the preferred locales but leaving the redirect implementations to users could make sense? Maybe later we could consider an API that is more aligned with best practices, but I don’t know if we have the experience and knowledge to know what that would look like today.

matthewp (approved):
lgtm

PR #8959, opened 2023-10-31
Undo the halloween theme

ElianCodes:
## Changes

- Resets the Halloween themed Houston
- Resets the `fancy` behaviour


natemoo-re (approved):
Thanks! This was fun 🫡

PR #8963, opened 2023-10-31
Support form submissions in the ViewTransitions router

matthewp:
## Changes

- Allows the user to enable form submissions to be handled. Currently this is via a `form` option on the ViewTransitions router: `<ViewTransitions form />`. Probably there is a better name.
  - Presumably this would become the default in 4.0.
  - Also allows `data-astro-reload` like with anchors, to opt-out on a per form basis.
- Adds an option to the `navigate()` API `options.init` which is a `RequestInit`. This is needed for forms to provide the body and set the method on the request.
  - If this is undesirable as a public API we could maybe make it a private option.

## Testing

- Tests added in the normal way
- Added a test for when errors are thrown.

## Docs

- https://github.com/withastro/docs/pull/5299


packages/astro/components/ViewTransitions.astro
```diff
@@ -75,6 +84,36 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+				} satisfies RequestInit;
+				if(form.method === 'get') {
+					const url = new URL(action);
+					url.search = params.toString();
+					action = url.toString();
+				} else {
+					init.body = params;
+				}
+				ev.preventDefault();
+				navigate(action, { init });
```
martrapp, line 113: We only need this for method === "post".
The navigate event of the navigation API has a formData property. 

The HTML Standard says:
> event.[formData](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigateevent-formdata)
>
>    The [FormData](https://xhr.spec.whatwg.org/#formdata) representing the submitted form entries for this navigation, if this navigation is a "[push](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationtype-push)" or "[replace](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationtype-replace)" navigation representing a POST [form submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-form-submit); null otherwise.
>
>    (Notably, this will be null even for "[reload](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationtype-reload)" or "[traverse](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigationtype-traverse)" navigations that are revisiting a [session history entry](https://html.spec.whatwg.org/multipage/browsing-the-web.html#session-history-entry) that was originally created from a form submission.)

It would be great if we could do this similarly here, passing the form data as `options.formData` instead of the RequestInit object (only for "post" and nothing for "get")

packages/astro/components/ViewTransitions.astro
```diff
@@ -75,6 +84,36 @@ const { fallback = 'animate' } = Astro.props as Props;
...
+				}
+
+				const form = el as HTMLFormElement;
+				const formData = new FormData(form);
+				const params = new URLSearchParams(formData as any); // TS is wrong here
+				let action = form.action;
+				const init: RequestInit = {
+					method: form.method,
+					redirect: 'follow'
+				} satisfies RequestInit;
```
martrapp, line 104: For redirect, 'follow' is the default value, 'get' is the default method for requests. For forms with method "get" we do not need that object (only for "post)

packages/astro/components/ViewTransitions.astro
```diff
@@ -3,13 +3,18 @@ type Fallback = 'none' | 'animate' | 'swap';
 
 export interface Props {
 	fallback?: Fallback;
+	form?: boolean;
```
martrapp, line 6: I like the idea of using these props to select deviant behavior.
Looking at the uses, `handleForms` seems more intuitive?

packages/astro/src/transitions/router.ts
```diff
@@ -1,6 +1,9 @@
 export type Fallback = 'none' | 'animate' | 'swap';
 export type Direction = 'forward' | 'back';
-export type Options = { history?: 'auto' | 'push' | 'replace' };
+export type Options = {
+	history?: 'auto' | 'push' | 'replace';
+	init?: RequestInit;
+};
```
martrapp, line 6: `formData?: FormData`, see comment above?

packages/astro/src/transitions/router.ts
```diff
@@ -381,7 +385,7 @@ async function transition(
 ) {
 	let finished: Promise<void>;
 	const href = toLocation.href;
-	const response = await fetchHTML(href);
+	const response = await fetchHTML(href, options.init);
 	// If there is a problem fetching the new page, just do an MPA navigation to it.
```
martrapp, line 389: `options.formData ? RequestInit = { method: "post", body: formData } : undefined`

martrapp:
In the Navigate API, the Navigate event has a property for form data, but it is not clear from the specification how this property is set. I don't see a problem if Astro's `navigate()` has more options than the Navigate API's `navigate()`. But I would question the name and type of the option, see comments in code.

packages/astro/components/ViewTransitions.astro
matthewp: Ah yeah, I was confused remembering when we had attempted to handle redirects ourselves and thought I needed to set that.

packages/astro/components/ViewTransitions.astro
matthewp: Yeah I agree. The one downside I see here is that formData has a Content-Type of `multipart/form-data` whereas the default of a form is `application/x-www-form-urlencoded` So we would be changing the default. Probably ok since Astro can handle both.

What happens though if the form has `enctype` set? Are we going to ignore that and just always use the formdata type? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#enctype

packages/astro/components/ViewTransitions.astro
ematipico: Or maybe `animateForms` or something similar?  Astro doesn't handle anything other than catching the submit and passing everything to the endpoint. "Handle" could be misunderstood like validation, error management, etc.

ematipico (commented):
Can we add a test case with a failure when doing a submit? I would like to see/understand what's the expected behaviour in that case

matthewp:
@ematipico You mean if the API responds with a 400/500? I can add that.

ematipico:
> @ematipico You mean if the API responds with a 400/500? I can add that.

Yeah exactly, just to get a sense of the expected behaviour. Cheers!

packages/astro/components/ViewTransitions.astro
matthewp: We don't use the word `animate` as animations are a use-case but not what the feature actually does. You can actually use this router and have no animations at all.

matthewp:
@martrapp I have updated the API per your suggestion to use `formData` as the navigation option.

@ematipico I've added a test which causes regular transitions for non-200 responses.

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,37 @@
+---
+'astro': minor
+---
+
+Form support in ViewTransitions router
```
sarah11918, line 5:
```suggestion
Form support in View Transitions router
```

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,37 @@
...
+
+The `<ViewTransitions />` router can now handle form submissions, allowing you to retain stateful UI. This feature is opt-in for semver reasons and can be enabled by adding the `handleForms` prop to the component like so:
```
sarah11918, line 7:
```suggestion
The `<ViewTransitions />` router can now handle form submissions, allowing you to trigger in-page transitions from a form submission while keeping page state. This feature is opt-in and can be enabled by adding the `handleForms` prop to the routing component:
```
I'm gonna need you to sell forms to me harder than this, Matthew.  😅 

Can you do better than "retain stateful UI"? Something that more walks through the (exciting!) process itself? People are always asking about forms, can we hype this in a more "give the people what they want" kind of way?

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,37 @@
...
+    <!-- ... -->
+    <ViewTransitions handleForms />
+  </head>
+  <body>
+    <!-- ... -->
+  </body>
+</html>
+```
+
+Like with links, you can opt-out of the router handling a submission on a per-form basis with the `data-astro-reload` property:
```
sarah11918, line 26:
```suggestion
Just as with links, if you don't want the routing handling a form submission, you can opt out on a per-form basis with the `data-astro-reload` property:
```

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,37 @@
...
+
+```astro
+---
+// src/components/Contact.astro
+---
+<form class="contact-form" action="/request" method="post" data-astro-reload>
```
sarah11918, line 32:
```suggestion
<form class="contact-form" action="/request" method="POST" data-astro-reload>
```
Should these now be capitalized?

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,37 @@
...
+  <!-- ...-->
+</form>
+```
+
+Form support works on post `method="get"` and `method="post"` forms.
```
sarah11918, line 37:
```suggestion
Form support works on post `method="GET"` and `method="POST"` forms.
```

sarah11918 (commented):
Looks great @matthewp ! Made some suggestions to the changeset below! Basically, I'm gonna need a bit more hypeeeee 🥳

.changeset/many-weeks-sort.md
matthewp: I think it doesn't matter, but MDN documents them as lower case: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method

Also vscode completions suggest lower case.

sarah11918: OK, so we changed *our* case for v3 when writing JS in the frontmatter, but because this is just in HTML, we can use the lower case?  That's fine by me!

bluwy (approved):

martrapp:
I'm happy with the fact that we're adding properties that come later from the Navigation API as additional fields in NavigationNavigateOptions. And the new property in ViewTransitions is now also more descriptive. Everything looks good to me! 

The only thing I was wondering is if the popstate handler needs to be customized as well. After all, it always passes an empty options object. But I think it's fine the way it is. An app would prevent users from returning to the form after the "post". And with a multi-page form, you would disable the "forward" after navigating back.

martrapp (approved):

ematipico (approved):

.changeset/many-weeks-sort.md
matthewp: All right, I needed the pep-talk. Pushed with a bit more, trying to emphasize that this addition completes the circle of everything you need to stay in the client if that's what you want.

.changeset/many-weeks-sort.md
```diff
@@ -0,0 +1,43 @@
...
+
+Form support in View Transitions router
+
+The `<ViewTransitions />` router can now handle form submissions, allowing the same animated transitions and stateful UI retention you already get from `<a>` links to occur on form posts. With this added in Astro you can have animations in all of these scenarios:
+
+- Clicking links between pages.
+- Making stateful changes in forms (ie updating site preferences).
+- Manually triggered navigate via the `navigate()` API.
+
+This feature is opt-in for semver reasons and can be enabled by adding the `handleForms` prop to the component like so:
```
sarah11918, line 13:
```suggestion
The `<ViewTransitions />` router can now handle form submissions, allowing the same animated transitions and stateful UI retention on form posts that are already available on `<a>` links. With this addition, your Astro project can have animations in all of these scenarios:

- Clicking links between pages.
- Making stateful changes in forms (i.e. updating site preferences).
- Manually triggering navigation via the `navigate()` API.

This feature is opt-in for semver reasons and can be enabled by adding the `handleForms` prop to the `<ViewTransitions /> component:
```
Just suggested some tiny word changes. 

And just want to double check: e.g. should be used if the thing in brackets is one example ( of many possibilities), and i.e. is used when it's literally exactly the thing it is, just rephrased. 

So, if "making stateful changes **=** "updating site preferences", then i.e. is fine. Otherwise, that should be e.g.

matthewp: Ok, so this is one example, so I'll change to e.g. after merging.

sarah11918 (approved):

PR #8965, opened 2023-10-31
Move VT route announcer styles to a class

### LINKED ISSUE #8937: ViewTransitions (v3.2.0+) unusable with a content security policy
### Astro Info

```block
Astro                    v3.2.0
Package Manager          npm

Also tested in v3.4.0
```

_No response_

### Describe the Bug

The recent addition of a Route Announcer in `<ViewTransitions />` in [3.2.0](https://github.com/withastro/astro/releases/tag/astro%403.2.0) [appends](https://github.com/withastro/astro/pull/8621/files#diff-d04cc10c291f740d505238a9c81d6190fea48359d7dafa5be0d9c3b21f294a70R27) a div with a style declaration. That inline style breaks content security policy without `unsafe-inline` so makes view transitions unusable with a CSP. While it's possible to declare unsafe-inline for styles in the CSP, that's not optimal from a security standpoint.

To reproduce, run Astro in example below, click on a new page and inspect. You'll find:
```
<div aria-live="assertive" aria-atomic="true" style="position:absolute;left:0;top:0;clip:rect(0 0 0 0);clip-path:inset(50%);overflow:hidden;white-space:nowrap;width:1px;height:1px">About Page</div>
```

Developers should have an option at least to render the styling in a stylesheet, or opt-out so they can implement on their own.

I think Announce should be optional—on by default with an option to not include—so that developers can implement in a manner appropriate for their needs.

https://stackblitz.com/edit/github-zxgdrp-6sx3md?file=src%2Fpages%2Findex.astro

### Participation

matthewp:
## Changes

- Moves the router announcer styles to a class name
- Closes #8937
- This prevents issues with some CSP policies.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/sweet-nails-dream.md
+++ b/.changeset/sweet-nails-dream.md
@@ -1,0 +1,7 @@
+---
+'astro': patch
+---
+
+Move VT route announcer styles to a class
+
+Doing so allows stricter CSP policies.
--- a/packages/astro/components/viewtransitions.css
+++ b/packages/astro/components/viewtransitions.css
@@ -54,3 +54,16 @@
 		animation: none !important;
 	}
 }
+
+/* Route announcer */
+.astro-route-announcer {
+	position:absolute;
+	left:0;
+	top:0;
+	clip:rect(0 0 0 0);
+	clip-path:inset(50%);
+	overflow:hidden;
+	white-space:nowrap;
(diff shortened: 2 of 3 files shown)


martrapp (approved):
LGTM!

PR #8977, opened 2023-11-01
Prevent the route announcer from being visible

matthewp:
## Changes

- Moves the styles from the external viewtransition.css stylesheet into the component itself.
- `viewtransition.css` is only added if there are uses of the `transition:` directives. In apps that do not use those, the styles are never added.
- So to ensure that the route announce styles are on the page, the styles should be in the component. Should have been here in the first place!

## Testing

- Test added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -7,7 +7,20 @@ export interface Props {
 
 const { fallback = 'animate' } = Astro.props as Props;
 ---
-
+<style is:global>
+/* Route announcer */
+.astro-route-announcer {
+	position: absolute;
+	left: 0;
+	top: 0;
+	clip: rect(0 0 0 0);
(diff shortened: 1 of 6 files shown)


packages/astro/components/ViewTransitions.astro
```diff
@@ -7,7 +7,20 @@ export interface Props {
 
 const { fallback = 'animate' } = Astro.props as Props;
```
Princesseuh, line 8: Unrelated to this PR, but is the `as Props` needed?

Princesseuh (approved):

packages/astro/components/ViewTransitions.astro
matthewp: It was when I worked on vscode 😡

natemoo-re (approved):

PR #8978, opened 2023-11-01
feat(dev-overlay): Add a tooltip on plugin hover / focus

Princesseuh:
## Changes

What the title says.

## Testing

Added a test

## Docs

N/A. It was already documented that the name would show in the UI where needed

### DIFF
--- a/packages/astro/e2e/dev-overlay.test.js
+++ b/packages/astro/e2e/dev-overlay.test.js
@@ -15,12 +15,23 @@ test.afterAll(async () => {
 	await devServer.stop();
 });
 
-test.describe('Dev Overlay zzz', () => {
+test.describe('Dev Overlay', () => {
 	test('dev overlay exists in the page', async ({ page, astro }) => {
 		await page.goto(astro.resolveUrl('/'));
 
 		const devOVerlay = page.locator('astro-dev-overlay');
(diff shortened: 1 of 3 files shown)


packages/astro/src/runtime/client/dev-overlay/overlay.ts
```diff
@@ -374,7 +401,7 @@ export class AstroDevOverlay extends HTMLElement {
 	getPluginTemplate(plugin: DevOverlayPlugin) {
 		return `<button class="item" data-plugin-id="${plugin.id}">
 				<div class="icon">${this.getPluginIcon(plugin.icon)}<div class="notification"></div></div>
-				<span class="sr-only">${plugin.name}</span>
```
Princesseuh, line 377: The tooltip replaces the screenreader only element here. I tried it out with VoiceOver and it has the same result.

packages/astro/e2e/dev-overlay.test.js
```diff
@@ -15,14 +15,25 @@ test.afterAll(async () => {
 	await devServer.stop();
 });
 
-test.describe('Dev Overlay zzz', () => {
```
matthewp, line 18: RIP

matthewp (approved):
lgtm

PR #8981, opened 2023-11-01
Increase the scroll restoration throttle time

matthewp:
## Changes

- This increases the throttle time to ever 350ms.
- Safari adds a warning in the console if you exceed 100 calls to `replaceState` ever 30 seconds. 30/100 == 300ms. So our previous throttle would hit the threshold. The new one will not, we only will call it a max of 99 times.

## Testing

- Tested manually scrolling around a lot.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -515,7 +515,7 @@ if (inBrowser) {
 		addEventListener('popstate', onPopState);
 		addEventListener('load', onPageLoad);
 		if ('onscrollend' in window) addEventListener('scrollend', onScroll);
-		else addEventListener('scroll', throttle(onScroll, 300));
+		else addEventListener('scroll', throttle(onScroll, 350));
 	}
 	for (const script of document.scripts) {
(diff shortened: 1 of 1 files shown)


packages/astro/src/transitions/router.ts
```diff
@@ -515,7 +515,7 @@ if (inBrowser) {
 		addEventListener('popstate', onPopState);
 		addEventListener('load', onPageLoad);
 		if ('onscrollend' in window) addEventListener('scrollend', onScroll);
-		else addEventListener('scroll', throttle(onScroll, 300));
+		else addEventListener('scroll', throttle(onScroll, 350));
```
natemoo-re, line 518: Might as well make this a passive event listener at the same time, right? [It's widely supported](https://caniuse.com/passive-event-listener) these days.

```suggestion
		else addEventListener('scroll', throttle(onScroll, 350), { passive: true });
```

natemoo-re (approved):
Not blocking, but a small suggestion!

bluwy (approved):

Issue #8956, opened 2023-10-30
Astro Picture component generates duplicate images when used in multiple pages


bbbbaum:
### Astro Info

```block
Astro                    v3.4.0
Node                     v18.15.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

I have a structure as follows:

```text
/
├── src/
│   └── assets/
│   └── content/
│       └── blogposts/
│       └── config.ts
│   └── pages/
│       └── blogposts/
│              └── [slug].astro
│       └── blogposts.astro
│       └── index.astro
```

The idea is to have:
- a page with a list of all blogposts (`blogposts.astro`) 
- a page for each blogpost entry in collections (`[slug].astro`)

Every entry has an `image` property in its `config.ts` definition as follows:

```javascript
import { z, defineCollection } from 'astro:content';

const blogpostsCollection = defineCollection({
	type: 'content',
	schema: ({ image }) => z.object({
		title: z.string(),
		author: z.string(),
		image: image(),
	})
});

export const collections = {
	'blogposts': blogpostsCollection
};
```

The issue occurs when I build this project, the image that is referenced in the content entry gets generated twice for every specified width.

In this particular case I have the same Picture component on two pages:

```javascript
	<Picture
		src={entry.data.image}
		alt="Papaya"
		loading="eager" 
		sizes="(max-width: 448px) 400px, (max-width: 810px) 750px, 1050px"
		widths={[150, 450, 750, 1050, 1350]}
		formats={['avif', 'webp']}
		fallbackFormat="jpg"
	/>
```

Which should result in 19 images generated:
- 5 widths * 3 formats
- 3 * 1x-density images
- 1 randomly generated image
(More about density and 1 randomly generated image in #8866)

Instead, I get 37 images - everything except that 1 randomly generated image is duplicated.

This doesn't really sound like an intended behaviour, and it might have something to do with the way I'm using collection entry objects but I'm all out of ideas how to circumvent it within my project.

In `[slug].astro` page:

```javascript
// ...
export async function getStaticPaths() {
  const entries = await getCollection('blogposts');
  return entries.map(entry => ({
    params: { slug: entry.slug }, props: { entry },
  }));
}
const { entry } = Astro.props;
const { Content } = await entry.render();

// ...
<Picture
  src={entry.data.image}
  alt="Papaya"
  loading="eager" 
[...]

Issue #8047, opened 2023-08-13
New astro assets images pipeline should deduplicate images with identical props


mastfissh:
### What version of `astro` are you using?

2.8.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

N/A

### Describe the Bug

I have two components that display the same image on different pages.

One of them uses GetImage, and one of them uses the Image component. They both have width=700

When I run build, I get two identical images generated of width 700. 

Note that if I tweak them both to use GetImage, they deduplicate correctly, so worth noting this as a workaround.

See https://github.com/mastfissh/safety_matrix/commit/50c3d244b3950301c4e9893a87cc16577d3a8a2f which is a commit that implements this workaround

### What's the expected result?

I would hope for the generated image result to be deduplicated - ideally the name would be related to the properties somehow, so the image could be cached in between pages.

### Link to Minimal Reproducible Example

https://github.com/mastfissh/safety_matrix/commit/50c3d244b3950301c4e9893a87cc16577d3a8a2f

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
The names are based on the transform content and they are deduplicated, so not quite sure why it's generating two images in your case 🤔

chromakode:
This is happening for me as well. If I remove the `transition:name` directive from my `<Image>` components, I see a single image emitted. It appears that values other than image transform properties are being included in the hash.

PR #8984, opened 2023-11-02
feat(assets): Add property to image services to control which properties to use for hashing

Princesseuh:
## Changes

What the title says. This is a bit advanced, and most image services won't need it so we supply a good enough default while still enabling more advanced use cases.

Fix https://github.com/withastro/astro/issues/8956
Fix https://github.com/withastro/astro/issues/8047

## Testing

The current tests should pass! In particular, we have a test right now that tests image duplication in `srcset` (the most complex case). It already passed before, but it should still catch regressions

## Docs

Will add it to the Image Service API page


bluwy (commented):
I don't quite understand how this feature will fix the linked issues. Are we adding new properties to be hashed to fix them, or would they have to set it themselves? If it's the latter, are we able to do it by default instead?

I can see how this would be useful for third-party image services though since `ImageTransform` has `[key: string]: any`.

packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -16,9 +16,19 @@ export function propsToFilename(transform: ImageTransform, hash: string) {
...
+	transform: ImageTransform,
+	imageService: string,
+	propertiesToHash: string[]
+) {
 	// Extract the fields we want to hash
-	const { alt, class: className, style, widths, densities, ...rest } = transform;
-	const hashFields = { ...rest, imageService };
+	const hashFields = propertiesToHash.reduce(
+		(acc, prop) => {
+			// It's possible for `transform[prop]` here to be undefined, or null, but that's fine because it's still consistent
```
ematipico, line 27: nit: consistent ... to what exactly? Sorry, if I am going to read this comment new few weeks, I might not know what that means

ematipico (commented):
Should we have at least a new test case where **we use** this new option?

packages/astro/src/assets/utils/transformToPath.ts
Princesseuh: Sure, no worries. I actually had a longer message there that I removed because it overexplaining things a bit. I'll add more info

(for information, it's consistent between different transforms, as in, every transform that does not have a `height` explicitly set will end up with a `height: undefined` on it.)

bluwy (approved):

ematipico (approved):

.changeset/new-islands-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Added a property to image services to allow them to specify which properties of `getImage` / `Image` / `Picture` should be used for hashing the result files
```
sarah11918, line 5:
```suggestion
Adds a new property `propertiesToHash` to the Image Services API to allow specifying which properties of `getImage()` / `<Image />` / `<Picture />` should be used for hashing the result files.
```
Could be useful for some to have the property listed here! Also, please double check the wording still makes sense with the slight change!

sarah11918 (commented):
Left a quick suggestion to include the property name!

lerenos:
A small warning (probably related to ths update): I was using the `remotePatterns` key in image config, and it stopped optimizing my remote images after update from 3.0 to 3.5. Now I inserted the `domains` key and everything is working fine 🙌

Princesseuh:
> A small warning (probably related to ths update): I was using the `remotePatterns` key in image config, and it stopped optimizing my remote images after update from 3.0 to 3.5. Now I inserted the `domains` key and everything is working fine 🙌

Shouldn't be related to this, but if you can reproduce the problem, feel free to open an issue!

PR #8986, opened 2023-11-02
fix(assets): Propagate sizes attribute on all sources

Princesseuh:
## Changes

The `sizes` attribute needs to be on every `source` element, much like it is on the `img` one. We were wrongly not including it.

This PR also includes a small refactor to the types of the internal `srcSet` object we pass around, we had two types with the same name when really, only one is needed.

## Testing

Edited the picture test to test for `sizes` as well, since you can't really use `widths` without it anyway.

## Docs

N/A

### DIFF
--- a/.changeset/tasty-oranges-bathe.md
+++ b/.changeset/tasty-oranges-bathe.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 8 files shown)


ematipico:
Changeset missing

Issue #8990, opened 2023-11-02
create-astro: Even if installation of dependencies fails, still prints Dependencies installed


skirianov:
### Astro Info

```block
Astro                    v3.4.3
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

While running `create-astro` CLI on slow/bad network you can end up in Timeout error, but current implementation of `spinner` doesn't handle failed cases and will always print success message, even if error was shown. 

This results in this confusing behaviour:
 
<img width="863" alt="Screenshot 2023-11-02 at 22 17 29" src="https://github.com/withastro/astro/assets/74229951/282480b4-52d1-4c10-ad91-76bd6e67a73c">

`spinner` call from `dependencies.ts` file:
<img width="863" alt="Screenshot 2023-11-02 at 22 25 46" src="https://github.com/withastro/astro/assets/74229951/56111445-4c31-4b6c-9584-f34ba7801264">

Maybe it would be nice to handle such edge cases like slow internet, cause I think for my bad internet { cwd, timeout: 90_000, stdio: 'ignore' } timeout of 90s was not enough.

Also, maybe you could use your `isOnline` function every time before doing any network request, just to confirm that the user still has the internet. 

```
function isOnline(): Promise<boolean> {
     return dns.lookup('[http://github.com](https://t.co/dZiD0COfJD)').then(
         () => true,
         () => false
      );
}
```

### What's the expected result?

If due to my internet speed or internet connection interruption I was not able to install dependencies it should result in error. 

Could also print something like 
"Ah, oh, your internet connection is gone."
"Do you want to retry installing dependencies?"

Basically handling this edge case properly. 

### Link to Minimal Reproducible Example

N/A - CLI issue

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for looking into it!
Do you want to open a PR?

skirianov:
Still digging through the code. It looks like `spinner` itself never meant to be handling errors like this and always returns `end` message which is always success 😅 

```
export async function spinner({ start, end, while: update = () => sleep(100), }, { stdin = process.stdin, stdout = process.stdout } = {}) {
    const act = update();
    const tooslow = Object.create(null);
    const result = await Promise.race([sleep(500).then(() => tooslow), act]);
    if (result === tooslow) {
        const loading = await gradient(chalk.green(start), { stdin, stdout });
[...]

PR #8993, opened 2023-11-03
fix: added onError functionality to spinner and used in the current c…

skirianov:
closing #8990 

This PR depends on https://github.com/withastro/cli-kit/pull/25.

> Merge **only** after `cli-kit` is merged and published.

Added `onError` function to handle errors inside `spinner` if `while` function throws an errors. 

- new `onError` property inside `dependencies.ts` spinner
- `while` function catches and throws an error if installing dependencies fails
- `messages.ts/spinner` accept new args

## Changes

- What does this change?

## Testing

Manual QA

## Docs

No need to update docs, it's just DX fix

### DIFF
--- a/.changeset/cuddly-walls-remember.md
+++ b/.changeset/cuddly-walls-remember.md
@@ -1,5 +1,0 @@
----
-'@astrojs/underscore-redirects': patch
----
-
-Fixes a regression which used the wrong pattern for dynamic pages
--- a/packages/underscore-redirects/CHANGELOG.md
+++ b/packages/underscore-redirects/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @astrojs/underscore-redirects
 
+## 0.3.3
+
+### Patch Changes
+
+- [#8979](https://github.com/withastro/astro/pull/8979) [`0ee7c9aac`](https://github.com/withastro/astro/commit/0ee7c9aac9d0dbe727edb91ce944d607947aa242) Thanks [@alexanderniebuhr](https://github.com/alexanderniebuhr)! - Fixes a regression which used the wrong pattern for dynamic pages
+
 ## 0.3.2
 
 ### Patch Changes
--- a/packages/underscore-redirects/package.json
+++ b/packages/underscore-redirects/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@astrojs/underscore-redirects",
   "description": "Utilities to generate _redirects files in Astro projects",
-  "version": "0.3.2",
+  "version": "0.3.3",
   "type": "module",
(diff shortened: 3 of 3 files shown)


.changeset/silent-experts-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': major
+---
+
+Fixed spinner to handle errors if install dependencies fails
```
sarah11918, line 5:
```suggestion
Fixes an issue where a successful "Dependencies installed" message is displayed even when installing dependencies fails.
```
Something like this maybe?

sarah11918 (commented):
Hi @skirianov! Thanks for this fix.. seems... pretty important!  😅 

Just added a tiny suggestion to the changeset for your consideration, since we think of them as being user-facing, like docs. So, in our case, I often go back and look at any actual issue that was reported and use that as the basis for a fix, vs the more technical implication of it. See what you think!

skirianov:
@sarah11918 haha, yeah, quite an edge case, but still possible :D 

Loved your suggestion, commited. 

Need to first merge https://github.com/withastro/cli-kit/pull/25 and publish package before we can merge this one :)

skirianov:
Hi @sarah11918, updated the `package.json` and `pnpm-lock` to align with `cli-kit: 0.3.1` and resolved merge conflicts. 

If you're fine with it, we can merge the PR. Thanks :)

.changeset/silent-experts-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-astro': major
```
ematipico, line 2: Why is this a `major`? I think it should be a `patch` because **if fixes** a bug. Do you agree?

packages/astro/components/Image.astro
```diff
@@ -25,7 +24,8 @@ if (typeof props.height === 'string') {
 
 const image = await getImage(props);
 
-const additionalAttributes: HTMLAttributes<'img'> = {};
+const additionalAttributes: Record<string, any> = {};
```
ematipico, line 27: This change isn't part of the PR. Can you revert it, please?

packages/astro/components/Picture.astro
ematipico, line 1: We should revert these changes

ematipico (changes requested):
I think some merge/rebase went wrong, because now this PR has changes that don't to this PR. Could you please fix them? I expect to have two/three files changed (changeset and files related to `create-astro`)

skirianov:
@ematipico dang it, I reset the commit history and force pushed it....I'll update the PR

PR #8974, opened 2023-11-01
feat: experimental i18n routing

ematipico:
## Changes

Experimental i18n routing.

RFC: https://github.com/withastro/roadmap/pull/734

## Testing

New tests added. The existing tests not related to i18n routing should pass because this feature is **opt-in**.

## Docs

/cc @withastro/maintainers-docs for feedback!

PR #8995, opened 2023-11-03
chore(i18n): reworked fallback, docs, types and validation

ematipico:
## Changes

There was some issue with the types of `i18n.fallback`, and I added another level of validation to zod.

## Testing

Added a test case to cover the new logic

## Docs

@sarah11918 we can use this PR to land the correct [documentation for fallback](https://github.com/withastro/astro/pull/8974#discussion_r1380679252)

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1449,14 +1449,14 @@ export interface AstroUserConfig {
 			 * @version 3.5.0
 			 * @description
 			 *
-			 * Controls the routing strategyto determine your site URLs.
-             *
+			 * Controls the routing strategy to determine your site URLs.
+			 *
 			 *  - `prefix-other-locales`(default): Only non-default languages will display a language prefix. The `defaultLocale` will not show a language prefix.
-			 *		URLs will be of the form `example.com/[lang]/content/` for all non-default languages, but `example.com/content/` for the default locale.
+			 *    URLs will be of the form `example.com/[lang]/content/` for all non-default languages, but `example.com/content/` for the default locale.
 			 *  - `prefix-always`: All URLs will display a language prefix.
-             *  URLs will be of the form `example.com/[lang]/content/` for every route, including the default language.
-             *
-             * Note: Astro requires all content to exist within a `/[lang]/` folder, even for the default language.
+			 *    URLs will be of the form `example.com/[lang]/content/` for every route, including the default language.
+			 *
+			 * Note: Astro requires all content to exist within a `/[lang]/` folder, even for the default language.
 			 */
 			routingStrategy: 'prefix-always' | 'prefix-other-locales';
 		};
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1422,7 +1422,7 @@ export interface AstroUserConfig {
 			 *
 			 * #### Examples
 			 *
-			 * In this example, we configure Astro to redirect pages that belong to the `pt` locale to their `es` version, and pages that belong to the `fr` locale to their `en` version.
+			 * The following example configures your content fallback strategy to redirect unavailable pages in `/pt/` to their `es` version, and unavailable pages in `/fr/` to their `en` version. Unavailable `/es/` pages will return a 404.
 			 *
 			 * ```js
 			 * export defualt defineConfig({
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1381,7 +1381,7 @@ export interface AstroUserConfig {
 		 * @docs
 		 * @name experimental.i18n
 		 * @type {object}
-		 * @version 3.*.*
+		 * @version 3.5.0
 		 * @type {object}
 		 * @description
 		 *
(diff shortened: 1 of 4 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1412,34 +1412,51 @@ export interface AstroUserConfig {
...
-			 * @version 3.*.*
+			 * @type {Record<string, string>}
+			 * @version 3.5.0
 			 * @description
 			 *
-			 * The fallback system of the locales. By default, the fallback system affect the **content only**, and it doesn't
-			 * do any redirects.
+			 * The content fallback strategy when navigating to pages in non-default languages that do not exist. (e.g. a translated page has not been created).
+			 *
+			 * This a plain object, where a key is a locale that hasn't traslated pages, and the corresponding value is the locale where you want to redirect the missing page to.
```
sarah11918, line 1421:
```suggestion
			 * The content fallback strategy when navigating to pages in non-default languages that do not exist (e.g. a translated page has not been created).
			 *
			 * Use this object to declare a fallback `locale` for each language you support individually. If no fallback is specified, then all languages listed in `locales` will redirect the page to your `defaultLocale`.
```
This looks great! 

I made up that sentence at the end, because I think it helps to say what happens if you don't configure anything. 

I think a common thing to want to do here will be to make the one default fallback, and not have to list every language out individually. So, if the above is true, that's also very useful for doing this! If the above is NOT true, then I think we need to tell them how to set one default language for all other languages, as I expect this to be a common use case.

packages/astro/src/@types/astro.ts
```diff
@@ -1412,34 +1412,51 @@ export interface AstroUserConfig {
...
+			 *
+			 * #### Examples
+			 *
+			 * In this example, we configure Astro to redirect pages that belong to the `pt` locale to their `es` version, and pages that belong to the `fr` locale to their `en` version.
```
sarah11918, line 1425:
```suggestion
			 * The following example configures your content fallback strategy to redirect unavailable pages in `/pt/` to their `es` version, and unavailable pages in `/fr/` to their `en` version. Unavailable `/es/` pages will redirect to the `defaultLocale` (`en`).
```
This is an excellent example! I just tightened up some wording here, but we should also mention what happens for the case that is not explicitly stated. 

Does it default to `en` because it's the `defaultLocale`? Does it give a 404?

packages/astro/src/@types/astro.ts
```diff
@@ -1412,34 +1412,51 @@ export interface AstroUserConfig {
...
+			 * @type {'prefix-always' | 'prefix-other-locales'}
+			 * @default {'prefix-other-locales'}
+			 * @version 3.5.0
+			 * @description
+			 *
+			 * Controls the routing strategy:
+			 *  - `prefix-other-locales`: This is the default value. When used, Astro will identify those routes that belong to the default locale and execute a redirect to a route stripped of that locale.
+			 *		Use `example.com/[lang]/content/` for every route.
+			 *  - `prefix-always`: When used, all URLs need to contains a locale prefix. Astro will return a 404 for any route that doesn't fullfill the requirements.
+			 * 		Use `example.com/[lang]/content/` for all non-default languages, but `example.com/content` for the default locale/.
```
sarah11918, line 1456: I'll leave this one for the other PR where we have a suggestion in progress!

sarah11918 (commented):
These are fabulous updates! Just a few minor suggestions. Thank you for hanging in there with me as I try to get all of these options straight in my head.  :sweat_smile:

packages/astro/src/@types/astro.ts
ematipico: Yes, there aren't any changes here, only a different formatting

bluwy (approved):
LGTM for the logic changes.

packages/astro/src/@types/astro.ts
ematipico: The reason why I haven't put any default fallback is because:
- we do want to avoid redirects as much as possible **as default behaviour**
- it's possible that users don't require this feature (localised market, legal documents, etc.), so it would get more complicated to opt out of the default behaviour 

About this suggestion:

> If no fallback is specified, then all languages listed in `locales` will redirect the page to your `defaultLocale`.

As for now, we won't redirect to the `defaultLocale`, which means that if the user doesn't set a fallback for a locale, the missing pages will return 404.

We can explore a redirect behaviour later on

packages/astro/src/@types/astro.ts
ematipico: As for now, we don't redirect to the default locale, and we return a 404

sarah11918: Ok, that's easy enough to update! Got it!

sarah11918: The above is good to merge as a placeholder for me, but note to self I will check on how we write "return a 404" elsewhere in docs for consistency!

packages/astro/src/@types/astro.ts
sarah11918: Updated the comment (and the other one) with acceptable placeholder language that I'll check on! This is good enough to merge!

sarah11918 (approved):
Approving and my task is to make sure we're saying "returns a 404" properly and consistently before publishing!  Great changes here!

Issue #8970, opened 2023-10-31
Shikiji breaks custom language syntax highlighting


misterorion:
### Astro Info

```block
Astro                    v3.4.0
Node                     v18.18.2
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/mdx
                         @astrojs/react
                         @astrojs/sitemap
```

### Describe the Bug

Astro 3.3.0 introduced a change from the internal `shiki` syntax highlighter to `shikiji`. After this change, providing custom language grammars in `astro.config.mjs` via `path` to a `tmLanguage.json` file no longer works.

After upgrading Astro to version 3.3.0 and making no other changes to my code, syntax highlighting for the custom language is not applied and no warnings are thrown.

For example, prior to 3.3.0 this would work:

```javascript
import { defineConfig } from 'astro/config';
import mdx from '@astrojs/mdx';

import sitemap from '@astrojs/sitemap';

export default defineConfig({
  site: 'https://example.com',
  integrations: [mdx(), sitemap()],
  markdown: {
    shikiConfig: {
      theme: 'dracula',
      langs: [
        'go',
        'yaml',
        {
          id: 'caddy',
          scopeName: 'source.Caddyfile',
          path: '../../src/caddyfile.tmLanguage.json', // works with shiki, not with shikiji
        },
      ],
    },
  },
});
```

In my minimal example (astro 3.4.0, using `shikiji`), you can see that the syntax highlighting for the custom language in `src/pages/index.md` does not work in the preview when running `npm run dev`. Highlighting for the built-in languages does work. 

Changing the astro version to an earlier version (e.g. 3.1.0, using `shiki`) in `package.json` causes the syntax highlighting to work for both the custom language and built-in languages.

Is it a correct statement to say: with `shikiji` passing a custom `tmLanguage.json` file is not possible since `shikiji` grammars are all pure ESM? 

Is there a workaround or way to fix this other than downgrading, perhaps by forcing the internal highlighter to use `shiki` instead of `shikiji`?

### What's the expected result?

Using custom grammars with the latest Astro version should work with minimal or no code changes.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-j1xdb7?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
[...]

PR #8996, opened 2023-11-03
Add shiki lang path compat

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/8970

Shiki supports languages with the `path` prop, which loads language JSON files, relative to the package root (in our case, the Astro package 😬). It's a bit of an ugly compat, but they should be removed in Astro 4.

## Testing

added new test based on the issue repro.

## Docs

n/a

### DIFF
--- a/.changeset/strange-toes-remain.md
+++ b/.changeset/strange-toes-remain.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds compatibility for shiki languages with the `path` property
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -1,4 +1,7 @@
 ---
+import path from 'node:path';
+import fs from 'node:fs';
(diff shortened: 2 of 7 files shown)


Princesseuh (approved):

ematipico (approved):
It looks good to me. Is there any chance that we could add this feature to `shikiji` directly? Are the maintainers willing to accept such change?

bluwy:
I don't think Anthony (`shikiji` author) will be adding this. Its removal is actually the goal of shikiji to rely on ESM imports instead so the code is portable. The compat code we add here will bring the caveats of `shiki` before where it doesn't work too well in SSR.

Ideally they should move away from using `path`. I've commented some additional information at https://github.com/withastro/astro/issues/8970#issuecomment-1794397299

ematipico:
That makes sense, thank you for clarifying!

PR #8997, opened 2023-11-03
refactor(i18n): compute preferred locales lazily

ematipico:
## Changes

From the API bash session, a suggestion was to compute `preferredLocale` and `preferredLocaleList` lazily. To do so, we agreed that `i18n.locales` cab be passed down to the point where we compute the context for the endpoints, and the `Astro` faux-global.

This PR does that by removing the computation of `preferredLocale` and `preferredLocaleList` from the three pipelines, and instead, we simply pass `locales`. The computation is moved inside two new getters.

## Testing

This is a refactor, so existing functionality should not change and the existing tests should still pass.

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/endpoint/index.ts
+++ b/packages/astro/src/core/endpoint/index.ts
@@ -42,6 +42,9 @@ export function createAPIContext({
 	adapterName,
 	locales,
 }: CreateAPIContext): APIContext {
+	let preferredLocale: string | undefined = undefined;
+	let preferredLocaleList: string[] | undefined = undefined;
+
 	const context = {
 		cookies: new AstroCookies(request),
 		request,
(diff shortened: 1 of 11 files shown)


packages/astro/src/core/render/result.ts
```diff
@@ -194,13 +194,23 @@ export function createResult(args: CreateResultArgs): SSRResult {
 					result.cookies = cookies;
 					return cookies;
 				},
+				get preferredLocale(): string | undefined {
+					if (locales) {
+						return computePreferredLocale(request, locales);
```
matthewp, line 199: It would be good to cache the result of the compute, since it's relatively expensive, so the cost only happens on the first time. If you look at how we do this in cookies, we use a closure variable and check if it is already defined inside of the get().

Same for the other places where this compute occurs.

ematipico: Done here https://github.com/withastro/astro/pull/8997/commits/2060d027aaff5d555f183c3c8225e08ed1b5af2f

packages/astro/src/core/endpoint/index.ts
```diff
@@ -59,8 +58,18 @@ export function createAPIContext({
...
+				return computePreferredLocale(request, locales);
+			}
+			return undefined;
+		},
+		get preferredLocaleList(): string[] | undefined {
+			if (locales) {
+				return computePreferredLocaleList(request, locales);
+			}
+			return undefined;
+		},
```
bluwy, line 72: Should we also handle caching for the endpoint API context. However it doesn't seem like `cookies` is also cached here. Not a blocker for the PR for me though.

bluwy (approved):

PR #9004, opened 2023-11-06
refactor(i18n): make `normalizePath` to default to `true`

ematipico:
## Changes

From API bash session, we agreed that it makes sense to have the option `normalizePath` set to `true` as a default value because that's what users would want most of the times.

## Testing

Updated the tests 

## Docs

Update the examples


sarah11918 (approved):
LGTM!

bluwy (approved):

PR #9005, opened 2023-11-06
Handle Vite 5 changes

bluwy:
## Changes

1. Handle type changes and renames in Vite 5.
2. Handle preview server 404 serving middleware.
3. Skip preview server tests for Vite 5 for now. Created a new [linear issue](https://linear.app/withastro/issue/PLT-1148/rework-trailingslash-buildformat) to track it.

The changes are backwards compatible, and should allow Astro to pass in [Vite's ecosystem-ci](https://github.com/vitejs/vite-ecosystem-ci/actions/runs/6766541296/job/18387806208)

## Testing

Ran `pnpm test` locally with Vite 5 and `main`.

## Docs

n/a

### DIFF
--- a/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
+++ b/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
@@ -1,11 +1,13 @@
 import fs from 'node:fs';
 import { fileURLToPath } from 'node:url';
-import type { Plugin } from 'vite';
+import { version } from 'vite';
+import type { Plugin, Connect } from 'vite';
 import type { AstroSettings } from '../../@types/astro.js';
 import { notFoundTemplate, subpathNotUsedTemplate } from '../../template/4xx.js';
 import { stripBase } from './util.js';
 
 const HAS_FILE_EXTENSION_REGEXP = /^.*\.[^\\]+$/;
+const IS_VITE_5 = version.startsWith('5.');
 
 export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
 	const { base, outDir, trailingSlash } = settings.config;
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/preview/vite-plugin-astro-preview.ts
```diff
@@ -61,7 +63,19 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
...
 						res.end(notFoundTemplate(pathname, 'Not Found'));
 					}
-				});
+				};
+
+				// Vite 5 has its own 404 middleware, we replace it with ours instead.
+				if (IS_VITE_5) {
+					for (const middleware of server.middlewares.stack) {
+						// This hardcoded name will not break between Vite versions
+						if ((middleware.handle as any).name === 'vite404Middleware') {
```
ematipico, line 72:
```suggestion
						if ((middleware.handle as any)?.name === 'vite404Middleware') {
```
Is `.handle.name` safe with `any`? Should we have the optional chain operator?

packages/astro/test/preview-routing.test.js
```diff
@@ -1,7 +1,12 @@
 import { expect } from 'chai';
+import { version } from 'vite';
 import { loadFixture } from './test-utils.js';
 
-describe('Preview Routing', () => {
+const IS_VITE_5 = version.startsWith('5.');
+
+// Skip in Vite 5 as it changes how HTML files are served. We may want to review aligning
```
ematipico, line 7: Should we add a `TODO` as a reminder in case we need to review them?

ematipico (approved):

packages/astro/src/core/preview/vite-plugin-astro-preview.ts
bluwy: It should always be defined when a middleware is added. Actually it seems like I can use `as Connect.HandleFunction` instead. Will switch to that so it's more explicit.

packages/astro/test/preview-routing.test.js
bluwy: I've put this on https://linear.app/withastro/issue/PLT-1148/rework-trailingslash-buildformat instead so we don't forget. We should definitely revisit this during Astro 4 though since it will be a breaking change, so I thought a linear issue will help surface and priotitize it.

matthewp (approved):

PR #9007, opened 2023-11-06
Deprecate `@astrojs/prefetch`

bluwy:
## Changes

This should be merged/coordinated after https://github.com/withastro/astro/pull/8951

---

Not now, but when it's time to deprecate, we should do this:

```
npm deprecate @astrojs/prefetch "Use the prefetch feature in Astro 3.5 instead"
```

## Testing

n/a

## Docs

Updated `@astrojs/prefetch` readme.


bluwy:
I think it's easier to just make this change in https://github.com/withastro/astro/pull/8951

PR #9010, opened 2023-11-06
Use jpeg source files as fallback rather than converting to png

jasikpark:
## Changes

- Add `.jpg` and `.jpeg` as image sources that will be represented as themselves in `Picture.astro`

If JPEG is the source for an image, it is likely that was the intention of the creator of the image - PNGs are more effective at compressing simple shapes with lossless quality, while JPEGs are more effective at providing high-noise images like photographs in an effective file size. (https://web.dev/articles/choose-the-right-image-format#features_of_different_raster_image_formats)

## Testing

This is a change in default behavior of an existing api

## Docs

Probably worth updating the docs with the defaults?


florian-lefebvre:
Not sure about this. I personally don't mind about what the input is, but I want the most optimized output (webp if possible). Can this behavior be updated through the config instead of making it the default?

jasikpark:
> Not sure about this. I personally don't mind about what the input is, but I want the most optimized output (webp if possible). Can this behavior be updated through the config instead of making it the default?

This is not a change removing webp, it is a change enforcing `jpg` as the fallback value in the `img`, instead of `png`.

There will still always be a `webp` generated 👍

florian-lefebvre:
Oh ok! I misunderstood, thanks for clarifying!

.changeset/quiet-games-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
```
Princesseuh, line 2:
```suggestion
"astro": minor
```

It's not technically needed, since the component is experimental, but I'd rather have this in the minor just to avoid disturbing people by accident

Princesseuh (approved):
Looks great to me! Will merge on merge day tomorrow.

.changeset/quiet-games-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+`<Picture />`: Use jpeg source files as their own fallback rather than using png as the fallback
```
Princesseuh, line 5:
```suggestion
The `<Picture />` component will now use `jpg` and `jpeg` respectively as fallback formats when the original image is in those formats. 
```

PR #9013, opened 2023-11-06
(feat) return updated config in integration hooks

bayssmekanique:
## Changes

Summary of change in this discussion thread: https://github.com/withastro/roadmap/discussions/762 

- In the `astro:config:setup` hook for integrations, calls to `updateConfig` now return a destructured instance of the updated configuration.
- Changes `updateConfig` property to be strongly typed deep partial of the `AstroConfig` type.

## Testing

- Unit test was added to `astro/test/units/integrations/api.test.js`.
- Tested bundle against another package using `AstroAdapter` in a local environment.

## Docs

- Corresponding doc changes in withastro/docs/pull/5306


florian-lefebvre:
Really cool! I know this is not the initial goal of this PR but it would great to type the newConfig argument

bayssmekanique:
@florian-lefebvre, good point. Added.

packages/astro/src/@types/astro.ts
```diff
@@ -2055,11 +2055,11 @@ export interface SSRLoadedRenderer extends AstroRenderer {
 	};
 }
 
-export type RecursivePartial<T> = {
+export type DeepPartial<T> = {
```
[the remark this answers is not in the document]
bayssmekanique: Fixed in https://github.com/withastro/astro/pull/9013/commits/089f034aea3fc201505c8c9b625a1c4ee08a8bc4

packages/astro/src/integrations/index.ts
```diff
@@ -374,6 +375,7 @@ export async function runHookBuildSetup({
 					target,
 					updateConfig: (newConfig) => {
 						updatedConfig = mergeConfig(updatedConfig, newConfig);
+						return { ...updatedConfig };
```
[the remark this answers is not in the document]
bayssmekanique: It's spread in the original declaration for `updatedConfig`, so I'm just maintaining consistency. The object is pretty deep so this feels like it should either be a true clone or a read-only object, but those both seem a bit outside the scope of this issue.

Ref: https://github.com/withastro/astro/blob/06c51cc00bbdeff34e2b4bedc743eed0efad82b8/packages/astro/src/integrations/index.ts#L72-L77

matthewp:
We're going to do a final patch for Astro 3.x, then once that is done this will be merged in and be part of Astro 4.0.

packages/astro/src/@types/astro.ts
```diff
@@ -2311,7 +2311,7 @@ export interface AstroIntegration {
 			config: AstroConfig;
 			command: 'dev' | 'build' | 'preview';
 			isRestart: boolean;
-			updateConfig: (newConfig: Record<string, any>) => void;
+			updateConfig: (newConfig: DeepPartial<AstroConfig>) => AstroConfig;
```
lilnasy, line 2314: This certainly is a lot better than before, but it could maybe give a false sense of safety. `{}` is technically `Partial<URL>`, for example.

lilnasy: It would be an edge case admittedly, and the runtime logic might actually handle it too. I'll take a closer look when I can, but I'm not blocking a merge.

.changeset/wild-boats-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Returns updated config in integration hooks
```
bluwy, line 5:
```suggestion
Returns the updated config in the integration `astro:config:setup` hook's `updateConfig()` API
```

PR #9014, opened 2023-11-07
feat(overlay): Match the design's animations, shadows and general styling

Princesseuh:
## Changes

A bunch of small style tweaks, mostly around shadows and animations. In order to support exit animations, this PR adds a new "beforeTogglingOff" hook that plugins can have. Outside of animations, it's also useful showing a confirmation window

## Testing

The current tests that test for things to open should work!

## Docs

https://github.com/withastro/docs/pull/5302


natemoo-re (approved):
LGTM!

matthewp (approved):

packages/astro/src/runtime/client/dev-overlay/overlay.ts
```diff
@@ -72,11 +72,10 @@ export class AstroDevOverlay extends HTMLElement {
 				height: 56px;
 				overflow: hidden;
 				pointer-events: auto;
-
 				background: linear-gradient(180deg, #13151A 0%, rgba(19, 21, 26, 0.88) 100%);
-				box-shadow: 0px 0px 0px 0px #13151A4D;
 				border: 1px solid #343841;
 				border-radius: 9999px;
+				box-shadow: 0px 0px 0px 0px rgba(19, 21, 26, 0.30), 0px 1px 2px 0px rgba(19, 21, 26, 0.29), 0px 4px 4px 0px rgba(19, 21, 26, 0.26), 0px 10px 6px 0px rgba(19, 21, 26, 0.15), 0px 17px 7px 0px rgba(19, 21, 26, 0.04), 0px 26px 7px 0px rgba(19, 21, 26, 0.01);
```
ematipico, line 78: Now, this is what I call **shadow**! 🤣

packages/astro/src/runtime/client/dev-overlay/overlay.ts
```diff
@@ -418,24 +407,39 @@ export class AstroDevOverlay extends HTMLElement {
...
 
 	getPluginCanvasById(id: string) {
-		return this.shadowRoot.querySelector(`astro-dev-overlay-plugin-canvas[data-plugin-id="${id}"]`);
+		return this.shadowRoot.querySelector<HTMLElement>(
+			`astro-dev-overlay-plugin-canvas[data-plugin-id="${id}"]`
+		);
 	}
 
-	togglePluginStatus(plugin: DevOverlayPlugin, status?: boolean) {
+	async togglePluginStatus(plugin: DevOverlayPlugin, status?: boolean) {
```
ematipico, line 415: nit: `status` as a boolean is very cryptic. For example what `status = false` mean? Maybe we should choose a different name

ematipico (approved):

packages/astro/src/runtime/client/dev-overlay/overlay.ts
Princesseuh: would `pluginStatus` be a better name for you? Since it's a parameter of `togglePluginStatus`, I feel like it was pretty clear like this.

`status` matches the other API that toggle the plugin's state as well, but it's always done in the context of the plugin, so it felt fine to me

ematipico: Oh, Ok, I understand now. Maybe `hasNewStatus`?

Princesseuh: I changed it to `newStatus`!

PR #9016, opened 2023-11-07
feat(overlay): Add click to go to editor for audits

Princesseuh:
## Changes

> Needs https://github.com/withastro/compiler/pull/375

Much like you can go to the editor for xray, now you can for audits.

## Testing

Will be tested in a separate PR

## Docs

N/A


packages/astro/test/fixtures/middleware space/src/middleware.js
```diff
@@ -24,7 +24,7 @@ const first = defineMiddleware(async (context, next) => {
 		const response = await next();
 		const newResponse = response.clone();
 		const /** @type {string} */ html = await newResponse.text();
-		const newhtml = html.replace('<h1>testing</h1>', '<h1>it works</h1>');
+		const newhtml = html.replace('testing', 'it works');
```
Princesseuh, line 27: So, this is a bit awkward. But since we now add attributes, those kind of naive replacements don't work in dev.

In my humble, informed, opinion, you shouldn't rely on string replaces like this anyway, because whitespaces, minification etc etc could ruin it anyway. Typically, I'd recommend people to use a real tool for this, like https://github.com/cloudflare/lol-html

matthewp: wow, such shade 🌨

matthewp (approved):

packages/astro/src/core/compile/compile.ts
```diff
@@ -46,6 +46,7 @@ export async function compile({
 			scopedStyleStrategy: astroConfig.scopedStyleStrategy,
 			resultScopedSlot: true,
 			transitionsAnimationURL: 'astro/components/viewtransitions.css',
+			annotateSourceFile: viteConfig.mode === 'development' && astroConfig.experimental.devOverlay,
```
bluwy, line 49: I think it would be better to test with something like:

```suggestion
			annotateSourceFile: viteConfig.command === 'serve' && astroConfig.experimental.devOverlay,
```

To check if we're running in dev mode. `mode` in Vite is used to refer to the kind of dev or build, e.g. it could be `staging`.

Princesseuh: Thank you! I got confused because this function doesn't take the usual AstroSettings object

packages/astro/src/core/compile/compile.ts
```diff
@@ -46,6 +46,7 @@ export async function compile({
 			scopedStyleStrategy: astroConfig.scopedStyleStrategy,
 			resultScopedSlot: true,
 			transitionsAnimationURL: 'astro/components/viewtransitions.css',
+			annotateSourceFile: viteConfig.command === 'serve' && astroConfig.experimental.devOverlay,
```
Princesseuh, line 49:
```suggestion
			annotateSourceFile: !viteConfig.isProduction && astroConfig.experimental.devOverlay,
```

bluwy (approved):

PR #9019, opened 2023-11-08
fix(i18n): better filtering and search for preferred locales

ematipico:
## Changes

This PR fixes some edge cases that could happen when looking for the preferred locales.

Since this feature doesn't enforce a restriction on how the locales should be written, this could be a problem when reading the locales coming from the browser. This will become a problem with the four-letter locales, e.g. `pt_BR`.

The `Accept-Language` header has a value like this `pt-BR`, but a user **might** have their routes and folders like this: `src/pages/pt-br/index.astro` and `example.com/pt-br`.

This PR fixes this issue by:
- normalizing both sources: the locales coming from the browser and the locales coming from the configuration, it use the same function we use for the virtual module
- returning the locales of the configuration, and not the locale of the browser

## Testing

Added a new test case

## Docs

 N/A

### DIFF
--- a/packages/astro/src/core/render/context.ts
+++ b/packages/astro/src/core/render/context.ts
@@ -9,6 +9,7 @@ import type {
 import { AstroError, AstroErrorData } from '../errors/index.js';
 import type { Environment } from './environment.js';
 import { getParamsAndProps } from './params-and-props.js';
+import { normalizeTheLocale } from '../../i18n/index.js';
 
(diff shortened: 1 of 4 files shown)


matthewp:
Looks like there are some conflicts, can you update?

ematipico:
@matthewp that's done

bluwy (approved):

matthewp (approved):

PR #9029, opened 2023-11-08
fix(assets): Use uint8arrays instead of Buffer in code that can run outside of Node

Princesseuh:
## Changes

- What does this change?

### DIFF
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -41,7 +41,7 @@ type AssetEnv = {
 	assetsFolder: AstroConfig['build']['assets'];
 };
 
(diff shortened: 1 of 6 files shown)


alexanderniebuhr:
!preview assets-uint8array

alexanderniebuhr (approved):
I just tested this with a known good reproducible example and this fixes the issue, facing in non-node runtimes. 

I also don't think this is breaking, so I would be very happy to get this out sooner than later.

/lgtm

ematipico (approved):
The changes look good to me. I believe this PR should be a `patch` because we used the wrong types (data structures) to handle binary data. For users affected by this wrong assumption from our end, we can provide a migration path to patch their code.

PR #9032, opened 2023-11-09
fix(i18n): fallback index when routing is prefix-always

ematipico:
## Changes

This PR fixes a case where the `routingStrategy` is `prefix-always` and the `src/pages/index.astro` isn't present.

The routing strategy is pretty clear about what it does (now):
> example.com/ will redirect to `example.com/en

This works fine if `src/pages/index.astro` is present but it won't if the file isn't there.

This PR fixes the issue in dev and SSG (SSR is already handled):
- dev was a pretty easy fix
- SSG leverages the same fallback logic to create a fake "fallback" route, so we can emit a `index.html` file that redirects to `/en`

## Testing

I **deleted** the `src/pages/index.astro` from the fixture of `prefix-always` and the assertions should still be valid.

### DIFF
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -485,13 +485,13 @@ export function createRouteManifest(
 		routes.push(routeData);
 	});
 	const i18n = settings.config.experimental.i18n;
-
-	if (i18n && i18n.fallback) {
-		let fallback = Object.entries(i18n.fallback);
+	if (i18n) {
+		// In this block of code we group routes based on their locale
 
 		// A map like: locale => RouteData[]
 		const routesByLocale = new Map<string, RouteData[]>();
-		// We create a set, so we can remove the routes that have been added to the previous map
+		// This type is here only as a helper. We copy the routes and make them unique, so we don't "process" the same route twice.
+		// The assumption is that a route in the file system belongs to only one locale.
 		const setRoutes = new Set(routes);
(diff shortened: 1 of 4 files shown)


packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -524,70 +524,110 @@ export function createRouteManifest(
...
+						...indexDefaultRoute,
+						pathname,
+						route,
+						segments,
+						pattern: getPattern(segments, config),
+						type: 'fallback',
+					});
 				}
+			}
+		}
```
ematipico, line 559: This is the additional code. The rest was just moved to be like this:

```
if (i18n) {

	// new block
	if (i18n.routingStrategy === 'prefix-always') {}

	if (i18n.falklback) {}
}
```

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -187,7 +187,7 @@ export async function handleRoute({
 				.some((segment) => {
 					return locales.includes(segment);
 				});
-			if (!pathNameHasLocale) {
+			if (!pathNameHasLocale && pathname !== '/') {
```
matthewp, line 190: This should account for `base` I think.

matthewp: Thanks for adding the comment!

matthewp (approved):

PR #9045, opened 2023-11-10
fix: Query params trigger the trailingSlash error in preview mode

### LINKED ISSUE #8942: Query params trigger the `trailingSlash` error in preview mode
### Astro Info

```block
Package Manager          pnpm
Adapter                  none
                         @astrojs/solid-js
                         pagefind
```

n/a

### Describe the Bug

When navigating to for example `http://localhost:4321/search/?q=hask` while running `astro preview` an error is shown:

image

The page should be shown normally

https://stackblitz.com/edit/github-rj7etb?file=src%2Flayouts%2FLayout.astro

### Participation

rishi-raj-jain:
## Changes

So this fixes #8942. It basically tweaks the pathname logic to only look at pathname instead of (currently) pathname with query params.

## Testing

I think with the current testing suite this was already covered. I didn't particularly test this change.

## Docs

I don't think we need a documentation change for this as it was an internal configuration that broke the expectation from preview server.


packages/astro/src/core/preview/vite-plugin-astro-preview.ts
```diff
@@ -24,7 +24,8 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
 					return;
 				}
 
-				const pathname = stripBase(req.url!, base);
+				let strippedPathname = stripBase(req.url!, base);
```
florian-lefebvre, line 27:
```suggestion
				const strippedPathname = stripBase(req.url!, base);
```
Can this be a const? Doesn't look like it's reassigned

packages/astro/src/core/preview/vite-plugin-astro-preview.ts
```diff
@@ -24,7 +24,8 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
 					return;
 				}
 
-				const pathname = stripBase(req.url!, base);
+				const strippedPathname = stripBase(req.url!, base);
+        const pathname = new URL(strippedPathname, 'https://a.b').pathname
```
florian-lefebvre, line 28: I'm not familiar with the codebase but what is this `a.b` doing?

rishi-raj-jain: It's anything, we don't care about the domain. We just extract the pathname but need the URL method to correctly parse the pathname.

.changeset/tidy-peas-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix preview server pathname parse
```
bluwy, line 5:
```suggestion
Fixes preview server `trailingSlash` handling for request URLs with query strings
```

bluwy (approved):

Issue #9044, opened 2023-11-10
Cannot read properties of undefined (reading 'referencedImages')


J-env:
### Astro Info

```block
Astro                    v3.5.0
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/preact
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

In Astro3.5.0 version, when using the client:only directive, the image information in the browser will report an error.
<img width="1604" alt="image" src="https://github.com/withastro/astro/assets/25259920/a4188067-7d46-4149-a2a8-f1debf74c0a5">

Is this supposed to be the code generated in ssr mode? Client js code shouldn't be so complicated.
<img width="912" alt="image" src="https://github.com/withastro/astro/assets/25259920/6d661c6c-5fc7-47f8-8cc3-2c7a087a0940">

### What's the expected result?

It would be better if the client code is the same as Astro3.4.4 version.
```js
const s  = { src: '/static-dev/test.a2872d8e.avif',
    width: 626,
    height: 358,
    format: 'avif' }

export {s as t}
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-sdeict-ffgcqm?file=src%2Fcomponents%2FTest2.tsx,src%2Fcomponents%2FTest.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

amxmln:
I can confirm this issue also crops up when generating static pages that contain a framework component which includes an `import.meta.glob` statement that imports images (SVGs in my case).

PR #9049, opened 2023-11-10
fix(assets): Fix image errors when images were used on the client

Princesseuh:
## Changes

We can't track images used on the client, so we need to always assume that they're used and can't be deleted from the final build

Fix https://github.com/withastro/astro/issues/9044

## Testing

Added a test

## Docs

N/A


GiyoMoon:
Thanks a lot for this fix 🙏 Ran into this issue earlier today.

Princesseuh:
> Thanks a lot for this fix 🙏 Ran into this issue earlier today.

No worries! Just in case, this is a 3.5.0 regression, downgrading to the latest 3.4.x version should work in the meantime while we release this fix (likely later today)

ematipico (approved):

bluwy (approved):

Issue #9041, opened 2023-11-10
[astro] params not interpreted as string, when `trailingSlash: 'always'` for API routes


itsmatteomanf:
### Astro Info

```block
Astro                    v3.5.0
Node                     v18.18.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Whenever you have a build that includes a "page" that is not an `.astro` file, but a `.ts` (or I presume a `.js`) file, it will throw an error complaining that the param is not a string, while it definitely is.

This bug appeared with v3.5.0, it worked perfectly fine with all previous versions.

The local error shows as:

```
 error   Expected "lang" to be a string
  File:
    ./node_modules/.pnpm/path-to-regexp@6.2.1/node_modules/path-to-regexp/dist/index.js:237:19
  Code:
    236 |             var typeOfMessage = repeat ? "an array" : "a string";
    > 237 |             throw new TypeError("Expected \"".concat(token.name, "\" to be ").concat(typeOfMessage));
          |                   ^
      238 |         }
      239 |         return path;
      240 |     };
  Stacktrace:
TypeError: Expected "lang" to be a string
    at Object.generate (./node_modules/.pnpm/path-to-regexp@6.2.1/node_modules/path-to-regexp/dist/index.js:237:19)
    at stringifyParams (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/routing/params.js:26:31)
    at findPathItemByKey (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/render/route-cache.js:75:21)
    at getParamsAndProps (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/render/params-and-props.js:25:29)
    at async createRenderContext (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/render/context.js:8:27)
    at async generatePath (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/build/generate.js:402:25)
    at async generatePage (file://./node_modules/.pnpm/astro@3.5.0_typescript@5.2.2/node_modules/astro/dist/core/build/generate.js:252:5)
[...]

Issue #9040, opened 2023-11-10
_image route returns 404 when trailingSlash set to always


drewtownchi:
### Astro Info

```block
Astro                    v3.5.0
Node                     v18.17.1
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

When `trailingSlash: 'always',` is set in the defineConfig in astro.config.mjs the _image route when using the `<Image>` and `<Picture>` tags from `astro:assets` returns a 404

### What's the expected result?

It should return a 200 and the Image

### Link to Minimal Reproducible Example

https://github.com/drewtownchi/image-bug

### Participation

- [ ] I am willing to submit a pull request for this issue.

Boston343:
FWIW, I'm noticing the same. This issue does not occur in astro v3.4.4. I am running Node v18.18.2 on Windows 11.

PR #9051, opened 2023-11-10
fix: regression when calculating params and pattern

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/9041
Closes https://github.com/withastro/astro/issues/9040

The issue comes from a naive refactor of mine when coding i18n stuff.

I reverted the code.

## Testing

I added a new test that failed at the beginning, and now it passes

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -60,9 +60,12 @@ function getParts(part: string, file: string) {
 	return result;
 }
 
-function getPattern(segments: RoutePart[][], config: AstroConfig) {
+function getPattern(
+	segments: RoutePart[][],
+	config: AstroConfig,
+	addTrailingSlash: AstroConfig['trailingSlash']
+) {
 	const base = config.base;
-	const addTrailingSlash = config.trailingSlash;
 	const pathname = segments
 		.map((segment) => {
 			if (segment.length === 1 && segment[0].spread) {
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/routing/manifest/create.ts
ematipico, line 1: This is the file that contains the issue, I reverted the signature of `getPattern`

Princesseuh:
I assume this also fixes https://github.com/withastro/astro/issues/9040, right?

ematipico:
> I assume this also fixes #9040, right?

Yes :) updated the PR description

Princesseuh (approved):

packages/integrations/sitemap/test/staticPaths.test.js
```diff
@@ -30,7 +31,8 @@ describe('getStaticPaths support', () => {
 		expect(urls).to.not.include('http://example.com/de/404/');
 	});
 
-	it('includes numerical pages', () => {
-		expect(urls).to.include('http://example.com/123/');
+	it('includes numerical pages', async () => {
+		const page = await fixture.readFile('./it/manifest');
+		expect(page).to.contain('I\'m a route in the "it" language.');
 	});
```
bluwy, line 37: Should the old test be kept? The new changes replaces it.

ematipico: Good Catch, I'll restore the previous test

bluwy (approved):

itsmatteomanf:
@ematipico note that this broke the root to default locale redirect, it doesn't respect the setting and does redirect to no trailing slash even with `trailingSlash: 'always'`.

PR #9052, opened 2023-11-10
[docs] test config-reference formatting for sidebar

sarah11918:
## Changes

Changes `Prefetch` and `i18n` into `heading`s to see if I can get the docs sidebar nested properly

## Testing

This is a test! The automated PR for docs will let me test the output

## Docs

Just docs!

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1494,7 +1494,7 @@ export interface AstroUserConfig {
 			 *
 			 * Use this object to declare a fallback `locale` route for each language you support. If no fallback is specified, then unavailable pages will return a 404.
 			 *
-			 * #### Example
+			 * ##### Example
 			 *
 			 * The following example configures your content fallback strategy to redirect unavailable pages in `/pt-br/` to their `es` version, and unavailable pages in `/fr/` to their `en` version. Unavailable `/es/` pages will return a 404.
 			 *
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1457,6 +1457,7 @@ export interface AstroUserConfig {
 		i18n?: {
 			/**
 			 * @docs
+		   * @kind h4
 			 * @name experimental.i18n.defaultLocale
 			 * @type {string}
 			 * @version 3.5.0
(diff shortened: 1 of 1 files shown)


ematipico (approved):

delucis (changes requested):
Don’t think this will do what you want!

delucis:
(Speed reviewed in case Ema’s approval made you merge this 😅)

I’m pretty sure our logic in docs only distinguishes between `@heading` and blocks without `@heading`. So these you’ve added would pull these out to be h2s like this:

image image

Breaking for lunch, but we can take a look at a solution when I’m back. (FWIW the current version is potentially OK I think as it matches e.g. how we handle the `build.*` headings.)

sarah11918:
For Prefetch I think that works, if it's moved to other things that work that way! I'd really just want the experimental flag to show up as a one-liner, I think though:

image
image

packages/astro/src/@types/astro.ts
```diff
@@ -538,6 +538,7 @@ export interface AstroUserConfig {
 
 	/**
 	 * @docs
+	 * @kind heading
 	 * @name prefetch
```
sarah11918, line 542:
```suggestion
	 * @name Prefetch Options
```

delucis:
Ah, yes, that makes sense. In that case, we probably also need to move the whole `prefetch` block AFTER the build options?

sarah11918:
Yes, but I'm still not sure how to make the i18n options not show up in the sidebar

ElianCodes (approved):
LGTM!

packages/astro/src/@types/astro.ts
```diff
@@ -1481,6 +1484,7 @@ export interface AstroUserConfig {
 
 			/**
 			 * @docs
+		   * @kind h4
 			 * @name experimental.i18n.fallback
 			 * @type {Record<string, string>}
 			 * @version 3.5.0
```
delucis, line 1490: Now that this is an `<h4>`, the `#### Example` in the `@description` should probably get an extra `#`.

delucis (approved):
Looks great @sarah11918! Glad we’re no longer JSBlocked 🚀

sarah11918:
Thanks for your help with this, @delucis ! This is now `headings` in the right direction!

Issue #9047, opened 2023-11-10
Can't add middleware with integrations


pilcrowOnPaper:
### Astro Info

```block
Astro                    v3.5.0
Node                     v20.2.0
System                   macOS (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             middleware-test
```

### Describe the Bug

Running a basic project with the following config:

```ts
// astro.config.mjs
export default defineConfig({
  integrations: [
    {
      name: "middleware-test",
      hooks: {
        "astro:config:setup": ({ addMiddleware }) => {
          addMiddleware({
            order: "pre",
            entrypoint: "./middleware.js",
          });
        },
      },
    },
  ],
});
```
```ts
// middleware.js
export const onRequest = async (_, next) => {
  console.log("middleware running...");
  return await next();
};
```

leads to an error:

```
[vite] Error when evaluating SSR module @astro-middleware: failed to import "undefined"
|- Error: Cannot find module 'undefined' imported from '@astro-middleware'
    at nodeImport (file:///Users/root/code/astro-integration-middleware/node_modules/.pnpm/vite@4.5.0/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:56088:25)
    at ssrImport (file:///Users/root/code/astro-integration-middleware/node_modules/.pnpm/vite@4.5.0/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:55990:30)
    at eval (@astro-middleware:3:50)
    at instantiateModule (file:///Users/root/code/astro-integration-middleware/node_modules/.pnpm/vite@4.5.0/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:56052:15)

Failed to load url undefined (resolved id: undefined) in @astro-middleware. Does the file exist?
```

### What's the expected result?

No errors

### Link to Minimal Reproducible Example

https://github.com/pilcrowOnPaper/astro-integration-middleware

### Participation

- [ ] I am willing to submit a pull request for this issue.

pilcrowOnPaper:
Same issue when using `@astrojs/node`

pilcrowOnPaper:
I'm not sure if it's helpful, but the logs are different when defining files that exist compared to those that don't

File that exists:

```
Failed to load url undefined (resolved id: undefined) in @astro-middleware. Does the file exist?
```

File that doesn't exist:

```
Failed to load url undefined (resolved id: undefined) in @astro-middleware. Does the file exist?
[...]

PR #9053, opened 2023-11-10
fix(intergration defined middleware): prevent importing non-existent user middleware

lilnasy:
## Changes
- Fixes (partially) #9047
- Refactor middleware/vite-plugin to ignore userMiddleware when it doesnt exist.

## Testing
new fixture

## Docs
Doesn't affect usage


matthewp (approved):
Thanks!

MoustaphaDev (approved):
Nice refactor!

lilnasy:
Do not merge until green.
The tests are not right, but the implementation seems correct.

ematipico (changes requested):
The issue is at line 47. If the user doesn't have a middleware, we bail and we emit `export const onRequest = undefined`.

In case the user doesn't have a middleware, we have to create a new one from scratch and **not attempt** to import the one from the user.

This isn't the correct fix.

lilnasy:
@ematipico could you take it from here?

lilnasy:
!preview

ematipico:
@lilnasy I opened a new PR https://github.com/withastro/astro/pull/9057

lilnasy:
> and not attempt to import the one from the user

@ematipico In this PR, the one from the user **does not** get imported, notice the `.filter(Boolean)`. Were you under the impression that it did?

PR #9055, opened 2023-11-10
Allow same-page navigations on form posts

matthewp:
## Changes

- Normally when the user navigates to the same page we do not trigger a fetch + navigation, instead we just navigate to the hash (if there is one).
- This should not be the case for form posts, which have side-effects that need to occur.
- This also makes it possible to omit the `action`.

## Testing

- New test added

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -465,7 +465,7 @@ export function navigate(href: string, options?: Options) {
 	// We do not have page transitions on navigations to the same page (intra-page navigation)
 	// but we want to handle prevent reload on navigation to the same page
 	// Same page means same origin, path and query params (but maybe different hash)
-	if (location.origin === toLocation.origin && samePage(toLocation)) {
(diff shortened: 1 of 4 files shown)


packages/astro/src/transitions/router.ts
```diff
@@ -465,7 +465,7 @@ export function navigate(href: string, options?: Options) {
 	// We do not have page transitions on navigations to the same page (intra-page navigation)
 	// but we want to handle prevent reload on navigation to the same page
 	// Same page means same origin, path and query params (but maybe different hash)
-	if (location.origin === toLocation.origin && samePage(toLocation)) {
+	if (location.origin === toLocation.origin && samePage(toLocation) && !options?.formData) {
```
natemoo-re, line 468: Might want to call this scenario out in the comment above here!

natemoo-re (commented):
Looks good!

natemoo-re (approved):

PR #9057, opened 2023-11-10
fix(middleware): do not import user middleware if not preset

ematipico:
## Changes

Fix https://github.com/withastro/astro/issues/9047

We need to track if the user doesn't a middleware. If not, we should not attempt to import it.

## Testing

I created a new fixture with not user middleware but with integration middlewares.

## Docs

N/A


matthewp (approved):

lilnasy:
@ematipico  Not sure how this was any different. Line 47 is untouched as well.

PR #9058, opened 2023-11-10
feat(overlay): Add a settings panel

Princesseuh:
## Changes

Add a little settings panel guy and the components it needs (the toggle)
image

Those options are saved locally, using `localStorage` and are intended for Dev Overlay settings.

## Testing

Added a test that it opens and closes, like other plugins

## Docs

Will do docs


bluwy (approved):
LGTM. We'd need a changeset before merging this.

I have a few thoughts about the design/behaviour, but not blocking the PR:
1. When opening the first Astro panel, then the settings panel, should the Astro panel auto-close? So there's only one panel at a time.
2. The settings name e.g. `Verbose logging?` feels a bit weird with the `?`. I don't usually see trailing punctuations in settings names.
3. Each setting controls, the setting description is larger than the setting title. Is it intentional?

image

Princesseuh:
> 3. Each setting controls, the setting description is larger than the setting title. Is it intentional?
> 
> image
> EDIT: The above image is what I see locally in Safari and Chrome, which seems to be slightly different compared to your screenshot.

That's weird, I can't see this no matter what browser I use 🤔

bluwy:
> That's weird, I can't see this no matter what browser I use 🤔

Strange, I can't reproduce this anymore. Sorry for the false alarm 😅

PR #9063, opened 2023-11-11
Allow setting all cookie package serialize/parse options

### LINKED ISSUE #9062: Astro.cookies.set() does not expose underlying cookie serialize options
### Astro Info

```block
Package Manager          npm
Integrations             none
```

_No response_

### Describe the Bug

The `Astro.cookies.set` implementation ultimately delegates final value serialization to the `serialize` function from the **cookie** package (see [serialize](https://github.com/jshttp/cookie#cookieserializename-value-options) docs).

However Astro is only exposing it's own (effectively anonymous/private, since it's not exported) `AstroCookieSetOptions` type which is obfuscating the underlying capabilities that are available.

The `AstroCookieSetOptions` options passed to `Astro.cookies.set` are passed down to **cookie**'s `serialize`, so, if you know this, you can just YOLO it with an `any` cast, etc.

The reason this is a problem for me is it took me awhile to chase down why the cookie value I was setting via `Astro.cookies.set` was getting mangled (url encoded), and it's because by default, the **cookie** `serialize` function's [`encode`](https://github.com/jshttp/cookie#encode) option uses `encodeURIComponent`.

So, short version is, there is nothing in the Astro docs to let you know and/or control the value serialization which is always passed through `encodeURIComponent`.

This came up for me because I was base64 encoding a cookie value, and base64 has a few characters that `encodeURIComponent` will encode, for example the trailing `=` padding characters will get url encoded.

I can YOLO it to customize the `encode` option, for example:

```typescript
const options: any = {
  httpOnly: true,
  encode: (o: string) => o
}
context.cookies.set("REDIRECT_URI", encodedURL, options);
```
But this is obviously unsafe since it's depending on an implementation detail of how Astro utilizes the **cookie** package under-the-covers.

I think Astro should probably:

1. Export the `AstroCookieSetOptions` type so it can be utilized easier by end-users
2. Change `AstroCookieSetOptions` to simply extend `CookieSerializeOptions` from **@types/cookie**

Full control of cookie serialization options is exposed

https://stackblitz.com/edit/github-pqoq26-w1uign?file=src%2Fpages%2Findex.astro

### Participation

alex-sherwin:
This PR fixes #9062 

* Changes `AstroCookieSetOptions` to be based on `CookieSerializeOptions` from the **cookie** package.  All the Astro code was already passing the options provided directly to **cookie**'s `serialize` function, so this is just a typing change.
* Changes `AstroCookieSetOptions` to be exported from the main **astro** package so it can be utilized by end-user code
* Adds an exported `AstroCookieGetOptions` based on `CookieParseOptions` from the **cookie** package.
* Modifies the `Astro.cookies.get()` function to take an optional `options: AstroCookieGetOptions | undefined` which is passed down to the **cookie** package's **parse** function
* `Astro.cookies.has()` gets the same treatment as `get` since it triggers cookie parsing
* Test coverage for existing behavior which covers the fact that the **default** behavior of cookie set/get value serialization is to URI encode the value (the **cookie** package is doing this by default, but no tests were explicitly covering this behavior)
* Test coverage for overriding the cookie set/get behavior for serialization/deserialization with identity functions to prevent URI encoding/decoding


matthewp:
Thanks for the contribution @alex-sherwin! We don't normally like to expose dependency options like this, because it leaks an implementation detail. We would like to be able to swap out the serialization dependency without it being a breaking change in Astro.

So, instead what I would suggest is adding the properties from CookieSerializeOptions that you need directly to AstroCookieSetOptions, so that it's part of our explicit API. Thanks!

matthewp (changes requested):
Instead of inheriting from CookieSerializeOptions, just take the properties it has that we need.

alex-sherwin:
> Thanks for the contribution @alex-sherwin! We don't normally like to expose dependency options like this, because it leaks an implementation detail. We would like to be able to swap out the serialization dependency without it being a breaking change in Astro.
> 
> So, instead what I would suggest is adding the properties from CookieSerializeOptions that you need directly to AstroCookieSetOptions, so that it's part of our explicit API. Thanks!

Happy to make that change

But to clarify, you're receptive to exposing at least the `encode` and `decode` options (and thus commit to supporting these in Astro in the event the underlying serialization package/dependency changes in the future)?

matthewp:
@alex-sherwin Yeah, totally fine with exposing such an option, just didn't want to expose all of the dependencies' options for the backwards compat reason.

matthewp:
Hello @alex-sherwin, was wondering if you planned to update this PR?

alex-sherwin:
> Hello @alex-sherwin, was wondering if you planned to update this PR?

Hi @matthewp, yes, I will plan to update it today

matthewp (approved):

natemoo-re (approved):

matthewp:
Great job getting this over the line @florian-lefebvre. We don't currently have a plan for the next minor, it will likely come around the new year.

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,4 @@
+---
+'astro': minor
+---
+Allows specifying `encode` and `decode` functions for cookies
```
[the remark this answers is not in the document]
Princesseuh: cc @florian-lefebvre

matthewp:
Docs PR: https://github.com/withastro/docs/pull/6042

.changeset/tiny-days-dance.md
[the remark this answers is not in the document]
florian-lefebvre: Will do that tomorrow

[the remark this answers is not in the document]
matthewp: @florian-lefebvre No need, I already redid the changeset. Feel free to improve mine though!

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,26 @@
+---
+'astro': minor
+---
+
+Cookie encoding / decoding can now be customized
+
+New `encode` and `decode` functions allow customizing how cookies are encoded and decoded. For example, cookies are by default encoded via `encodeURIComponent`, but you might want to avoid that in some cases such as when adding a URL as part of a cookie.
```
sarah11918, line 7:
```suggestion
Adds new `encode` and `decode` functions to allow customizing how cookies are encoded and decoded. For example, you can bypass the default encoding via `encodeURIComponent` when adding a URL as part of a cookie:
```

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,26 @@
...
+```astro
+---
+Astro.cookies.set('url', Astro.url.toString(), {
+  // Override the default encoding so that URI components are not encoded
+  encode: o => o
+});
+---
+```
+
+Later, you can get back the URL in the same way:
```
sarah11918, line 18:
```suggestion
Later, you can decode the URL in the same way:
```

sarah11918 (commented):
Thanks so much for providing this, @matthewp ! Looks great! Just left some tiny suggestions for your consideration!

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,26 @@
+---
+'astro': minor
+---
+
+Cookie encoding / decoding can now be customized
```
florian-lefebvre, line 5: What's the convention for ending periods @sarah11918 ? Curious for future changesets and consistency

matthewp: The first line gets used as a heading in the blog post, so that's why I always omit the period there.

florian-lefebvre: Gotcha, thanks!

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,26 @@
...
+
+Adds new `encode` and `decode` functions to allow customizing how cookies are encoded and decoded. For example, you can bypass the default encoding via `encodeURIComponent` when adding a URL as part of a cookie:
+
+```astro
+---
+Astro.cookies.set('url', Astro.url.toString(), {
+  // Override the default encoding so that URI components are not encoded
+  encode: o => o
+});
+---
```
ematipico, line 15:
```suggestion
---
import { encodeCookieValue } from "./cookies";
Astro.cookies.set('url', Astro.url.toString(), {
  // Override the default encoding so that URI components are not encoded
  encode: value => encodeCookieValue(value)
});
---
```

.changeset/tiny-days-dance.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+Later, you can decode the URL in the same way:
+
+```astro
+---
+const url = Astro.cookies.get('url', {
+  decode: o => o
+});
+---
```
ematipico, line 25:
```suggestion
---
import { decodeCookieValue } from "./cookies";
const url = Astro.cookies.get('url', {
  decode: value => decodeCookieValue(value)
});
---
```

ematipico (approved):
Added a better suggestion for the changelog. The arrow function `o => o` didn't tell us very much

.changeset/tiny-days-dance.md
sarah11918: I like Ema's changes to the code sample!

matthewp: Yes! Thanks so much, i was struggling with this, as this is *really* the use-case in mind, so just giving it a function name does help a lot.

PR #9069, opened 2023-11-12
Fix asset propagation regression in 3.5

### LINKED ISSUE #9066: Scoped styles not being applied
### Astro Info

```block
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

Firstly, please forgive my lack of English as it is not my first language :)

After updating to version 3.5.2, the stylesheet is not being applied on some pages.

After taking a closer look at the problem, I realised that **if there is any code that fetches content, such as `getCollection`, the style is not applied on that page.** (I don't know if this is related, but it appears as `<style>undefined</style>` in the rendered html).

This issue appears to be the same in 3.5.1, and doesn't seem to be present in versions prior to 3.4.4.

I've reproduced the same situation in the stackblitz below.
Thanks :)

Scoped styles must be applied

https://stackblitz.com/edit/github-giqnfl?file=package.json

### Participation

natemoo-re:
## Changes

- Resolves PLT-1196
- Closes #9066
- Fixes regression in 3.5.x with asset propagation in content collections
- Had to update a few changes made in #8854 to ensure behavior is properly behind a flag

## Testing

Added a regression test

## Docs

Bug fix only

### DIFF
--- a/packages/astro/src/content/vite-plugin-content-virtual-mod.ts
+++ b/packages/astro/src/content/vite-plugin-content-virtual-mod.ts
@@ -76,7 +76,7 @@ export function astroContentVirtualModPlugin({
 							hydratedComponents: [],
 							clientOnlyComponents: [],
 							scripts: [],
-							containsHead: true,
+							containsHead: false,
 							propagation: 'in-tree',
 							pageOptions: {},
 						},
--- a/packages/astro/src/core/build/plugins/plugin-css.ts
+++ b/packages/astro/src/core/build/plugins/plugin-css.ts
@@ -59,6 +59,8 @@ function rollupPluginAstroBuildCSS(options: PluginOptions): VitePlugin[] {
 	const pagesToCss: Record<string, Record<string, { order: number; depth: number }>> = {};
 	const pagesToPropagatedCss: Record<string, Record<string, Set<string>>> = {};
 
+	const isContentCollectionCache = options.buildOptions.settings.config.output === 'static' && options.buildOptions.settings.config.experimental.contentCollectionCache;
+
 	const cssBuildPlugin: VitePlugin = {
 		name: 'astro:rollup-plugin-build-css',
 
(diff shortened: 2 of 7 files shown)


packages/astro/src/core/build/plugins/plugin-css.ts
```diff
@@ -251,9 +250,16 @@ function rollupPluginAstroBuildCSS(options: PluginOptions): VitePlugin[] {
...
+						let propagatedStyles: Set<StylesheetAsset>;
+						if (isContentCollectionCache) {
+							propagatedStyles =
+								internals.propagatedStylesMap.get(pageInfoId) ??
+								internals.propagatedStylesMap.set(pageInfoId, new Set()).get(pageInfoId)!;
+						} else {
+							propagatedStyles =
+								pageData.propagatedStyles.get(pageInfoId) ??
+								pageData.propagatedStyles.set(pageInfoId, new Set()).get(pageInfoId)!;
+						}
```
lilnasy, line 262: So `internals` is the new place to store propagated assets, but it was read from even without the opt-in, when it should've been the current place `pageData` instead. Is that what happened?

natemoo-re: It was a few things! But that's the main one. Reading from `internals` should have only been behind the flag as the current behavior relies on `pageData.propagatedStyles`. I missed this when refactoring to put the feature back behind a flag and mistakenly assumed we had test coverage on this feature.

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -76,7 +76,7 @@ export function astroContentVirtualModPlugin({
 							hydratedComponents: [],
 							clientOnlyComponents: [],
 							scripts: [],
-							containsHead: true,
+							containsHead: false,
```
natemoo-re, line 79: This, it turns out, was also an important typo. `containsHead` will actually stop our propagation logic.

lilnasy (approved):
Thanks for the quick weekend fix!

PR #9071, opened 2023-11-12
Cancel response stream when connection closes

### LINKED ISSUE #9068: `ReadableStream` in response body not canceled when client closes connection
### Astro Info

```block
Node                     v20.2.0
Package Manager          pnpm
```

_No response_

### Describe the Bug

When implementing server-sent events, `cancel()` is not called when client closes the connection (e.g. closes the tab).

```ts
export async function GET() {
  const stream = new ReadableStream({
    start(controller) {
      // do stuff
    },
    cancel() {
      console.log("cancelled");
    },
  });
  return new Response(stream, {
    headers: {
      "X-Content-Type-Options": "nosniff",
      "Content-Type": "text/event-stream; charset=utf-8",
      "Transfer-Encoding": "chunked",
    },
  });
}
```

`cancelled` is logged to the console

https://github.com/pilcrowOnPaper/astro-readable-stream-bug

### Participation

pilcrowOnPaper:
## Changes

- Closes #9068
- Remove unused code

## Testing

No test added, would love some directions

## Docs

None


lilnasy:
Thanks for the enhancement, the code looks great! Good eye on the unused modules as well.

Do you think this behavior could be easily tested?

pilcrowOnPaper:
@lilnasy Hm, it should be possible, though I'm not sure what's the best way to test step 4

1. Make `fetch()` request to server
2. Return response with stream
3. Abort `fetch()` request
4. Check if `cancel()` was called

lilnasy:
What is the current unwanted behavior?

Is it that the readable stream continues to produce chunks, possibly endlessly?

pilcrowOnPaper:
> Is it that the readable stream continues to produce chunks, possibly endlessly?

Yeah. This is especially an issue for server sent events since there's no definite end to the stream. It continues to push data until the client closes the connection

lilnasy:
Yeah, we could use locals for that. I'll take a look.

.changeset/modern-ways-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/node': patch
+---
+
+Cancel response stream when connection closes
```
lilnasy, line 5:
```suggestion
The node adapter now automatically cancels the `ReadableStream` of your streaming responses when the visitor disconnects, allowing you to gracefully clean up resources.
```
Just making it more descriptive. LMK if this looks helpful.

pilcrowOnPaper: eh, I personally find the original descriptive enough (when compared to other changesets). Gonna defer to the maintainers, I don't have a strong preference

packages/integrations/node/test/api-route.test.js
```diff
@@ -89,4 +89,23 @@ describe('API routes', () => {
...
+			url: '/streaming',
+		});
+
+		let locals = { cancelledByTheServer: false };
+
+		handler(req, res, () => {}, locals);
+		req.send();
+
+		await new Promise((resolve) => setTimeout(resolve, 500));
+		res.emit("close");
```
lilnasy, line 105: This line might need another look. I wasn't able to mock actual disconnection which would have emitted the event realistically. I don't know enough about node to take it further.

pilcrowOnPaper: I think this fine 

>Indicates that the response is completed, or its underlying connection was terminated prematurely (before the response completion).

https://nodejs.org/api/http.html#event-close_2

lilnasy (approved):
LGTM, thanks!

bluwy (approved):
The original PR that added the response iterator was https://github.com/withastro/astro/pull/5418, but it sounds like this PR aligns more with how Response work as a standard, so I think this is the right approach.

.changeset/modern-ways-develop.md
bluwy: I don't really have a hard preference either, but since this is only a patch, I think a short description is also fine by me.

pilcrowOnPaper:
Is it true that the response body can be a `Buffer`? 

> Sometimes Astro sends a ReadableStream or a Buffer as a response body

bluwy:
I think it's possible in the past when we support returning simple objects in endpoints, but now that we enforce and coerce to Response, it shouldn't apply now.

.changeset/modern-ways-develop.md
sarah11918, line 5:
```suggestion
Fixes a bug where the response stream would not cancel when the connection closed
```
(Not blocking) I can meet you halfway! What about something of this style? (Make it accurate.)

(I did like @lilnasy 's version, but can appreciate that everyone wants something shorter for a patch!)

Issue #9002, opened 2023-11-05
passthroughImageService shouldn't output multiple files


stancl:
### Astro Info

```block
Astro                    v3.4.3
Node                     v20.5.1
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/mdx
                         @astrojs/tailwind
                         @astrojs/vue
```

### Describe the Bug

- The noop image service only copies contents into the target file
- Some things enable it implicitly (e.g. `@astrojs/cloudflare`)
- When multiple dimensions are used for a single image (multiple `<Image>` tags for the same image, with different dimensions), it **produces multiple files even if they have the exact same contents (= same as original image)**
- This unnecessarily leads to multiple requests — different `<img>` tags requesting different URLs, even if they all have the exact same content (untransformed content, same as source image)

Reproduction steps (don't have the time to set up a Stackblitz example right now):
1. Add an image (`src/example-image.png`)
2. Add this code to any page:
    ```html
    <Image src={importedImage} width="100" alt="foo" />
    <Image src={importedImage} width="200" alt="bar" />
    ```
3. Use `passthroughImageService()` in your Astro config
4. Run a static build and notice `example-image.png`, `example-image.hash1.webp`, `example-image.hash2.webp` in your `dist/_astro`
5. Browser requests both `example-image.hash1.webp` and `example-image.hash2.webp` even if they have the same contents

### What's the expected result?

Only produce a single `.webp` file, or none at all and just use the original format.

Don't think this can be solved in `packages/astro/src/assets/services/noop.ts` since that only deals with the contents of the output file. The *path* to the output file is probably handled by the code that calls `noopService`.

### Link to Minimal Reproducible Example

See the bug description

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
https://github.com/withastro/astro/pull/8984 should unlock this pretty easily by just including the `src` into the hash and nothing else.

PR #9075, opened 2023-11-12
fix(assets): Only use src to hash files generated by the passthrough service

Princesseuh:
## Changes

Only uses the `src` to hash images on the passthrough service, that way attributes don't cause multiple files to be generated.

Fix https://github.com/withastro/astro/issues/9002

## Testing

We have tests that are testing this hook of the image service API

## Docs

N/A


bluwy (approved):

bluwy:
Looks like we also need a changeset before merging this

PR #9080, opened 2023-11-13
feat(node): add trailingSlash support

### LINKED ISSUE #7808: Bare paths redirect to paths with a trailing slash in production

msxdan:
## Changes

- Add logic to support trailingSlash with `ignore`, `always` and `never` values with nodejs adapter
- Closes #7808

## Testing

Add tests to use each possible value of trailingSlash and check if expected results are good and not breaking other tests

`trailingSlash = always` &#8594; Should always contain the / at the end (will always return host/pathname/?queryifexists)
`trailingSlash = never` &#8594; Should never contain the / at the end (will always return host/pathname?queryifexists)
`trailingSlash = ignore` &#8594; Should ignore the / at the end (will always work without redirect)

New behaviour of `trailingSlash = ignore` will return 200 and the content of the index.html inside the directory
> I had to change the `prerenderer.test.js` to support this new behaviour and not breaking the test

## Docs

Should work as expected in current documentation

---
I know there's currently an open PR (https://github.com/withastro/astro/pull/8593) about this, but I was not able to make it working and needed this for a personal project

### DIFF
--- a/packages/integrations/node/src/types.ts
+++ b/packages/integrations/node/src/types.ts
@@ -1,5 +1,6 @@
 import type { NodeApp } from 'astro/app/node';
 import type { IncomingMessage, ServerResponse } from 'node:http';
+import type { SSRManifest } from 'astro';
 
 export interface UserOptions {
 	/**
@@ -17,6 +18,7 @@ export interface Options extends UserOptions {
 	server: string;
 	client: string;
 	assets: string;
+	trailingSlash?: SSRManifest['trailingSlash'];
 }
 
 export interface CreateServerOptions {
--- a/packages/integrations/node/test/trailing-slash.js
+++ b/packages/integrations/node/test/trailing-slash.js
@@ -41,6 +41,15 @@ describe('Trailing slash', () => {
 				delete process.env.PRERENDER;
 			});
 
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
+			});
+
 			it('Can render prerendered route with redirect', async () => {
 				const res = await fetch(`http://${server.host}:${server.port}/some-base/one`, {
 					redirect : 'manual'
@@ -87,6 +96,15 @@ describe('Trailing slash', () => {
 				await server.stop();
 				await fixture.clean();
 				delete process.env.PRERENDER;
+			});
+
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
 			});
 
 			it('Can render prerendered route with redirect', async () => {
@@ -140,6 +158,15 @@ describe('Trailing slash', () => {
 				delete process.env.PRERENDER;
 			});
 
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
+			});
+
 			it('Can render prerendered route with redirect', async () => {
 				const res = await fetch(`http://${server.host}:${server.port}/some-base/one/`, {
 					redirect : 'manual'
@@ -187,6 +214,15 @@ describe('Trailing slash', () => {
 				await server.stop();
 				await fixture.clean();
 				delete process.env.PRERENDER;
+			});
+
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
 			});
 
 			it('Can render prerendered route with redirect', async () => {
@@ -239,6 +275,15 @@ describe('Trailing slash', () => {
 				await server.stop();
 				await fixture.clean();
 				delete process.env.PRERENDER;
+			});
+
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
 			});
 
 			it('Can render prerendered route with slash', async () => {
@@ -306,6 +351,15 @@ describe('Trailing slash', () => {
 				await server.stop();
 				await fixture.clean();
 				delete process.env.PRERENDER;
+			});
+
+			it('Can render prerendered base route', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}`);
+				const html = await res.text();
+				const $ = cheerio.load(html);
+
+				expect(res.status).to.equal(200);
+				expect($('h1').text()).to.equal('Index');
 			});
 
 			it('Can render prerendered route with slash', async () => {
--- a/packages/integrations/node/test/trailing-slash.js
+++ b/packages/integrations/node/test/trailing-slash.js
@@ -1,0 +1,351 @@
+import nodejs from '../dist/index.js';
+import { loadFixture } from './test-utils.js';
+import { expect } from 'chai';
+import * as cheerio from 'cheerio';
+
+/**
+ * @typedef {import('../../../astro/test/test-utils').Fixture} Fixture
+ */
+
+async function load() {
+	const mod = await import(`./fixtures/trailing-slash/dist/server/entry.mjs?dropcache=${Date.now()}`);
+	return mod;
+}
+
+describe('Trailing slash', () => {
+	/** @type {import('./test-utils').Fixture} */
+	let fixture;
+	let server;
+	describe('Always', async () => {
+		describe('With base', async () => {
+			before(async () => {
+				process.env.ASTRO_NODE_AUTOSTART = 'disabled';
+				process.env.PRERENDER = true;
+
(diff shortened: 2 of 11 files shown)


packages/integrations/node/src/http-server.ts
```diff
@@ -23,12 +25,55 @@ function parsePathname(pathname: string, host: string | undefined, port: number)
 
 export function createServer(
 	{ client, port, host, removeBase }: CreateServerOptions,
-	handler: http.RequestListener
+	handler: http.RequestListener,
+	trailingSlash: AstroUserConfig['trailingSlash']
```
natemoo-re, line 29: I'm not sure that `createServer` is the right place to handle this... Seems like we should be able to read the `trailingSlash` value from the manifest and use that to point requests to the correct route.

packages/integrations/node/src/types.ts
```diff
@@ -15,6 +16,7 @@ export interface Options extends UserOptions {
 	port: number;
 	server: string;
 	client: string;
+	trailingSlash: AstroUserConfig['trailingSlash'];
```
natemoo-re, line 19: I don't think we should change the signature of the Node options. This information is already available in the Astro config, so it should be available to the `app` that is generated by Astro.

packages/integrations/node/test/trailing-slash.js
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * @typedef {import('../../../astro/test/test-utils').Fixture} Fixture
+ */
+
+async function load() {
+	const mod = await import(`./fixtures/trailing-slash/dist/server/entry.mjs?dropcache=${Date.now()}`);
+	return mod;
+}
+
+describe('Trailing slash', () => {
```
natemoo-re, line 15: These tests are very helpful, nice job! 🎉

natemoo-re (commented):
Thank you so much for the PR! This is great work but I'm not sure if it's implemented at the correct layer. I think this might make more sense implemented in the Middleware itself? @matthewp might have the most context here.

packages/integrations/node/src/types.ts
matthewp: Yeah, agree here

ematipico:
I wonder if this behaviour should be implemented inside `astro/app`. The class has already that information - `this.#manifest.trailingSlash` and we could add the logic inside the `match` function.

packages/integrations/node/src/http-server.ts
matthewp: Yeah sounds right. What I'm unsure about is why this needs to happen in the adapter level and not inside of Astro core itself. It seems like we should be handling trailingSlash ourselves. I can't recall why we do not.

It might be the case that we forgot to initially, and thought it would be backwards incompatible to do so. I do have some recollection of handle trailing slash for things like assets at build time (by making sure it was correct in the manifest).

If that's the case, then I thinking having in the Node adapter is ok, and if we ever fix it in core then it will just already be fixed before it gets there.

In any case, fixing it inside of the middleware seems ideal.

florian-lefebvre:
matthewp confirmed that what ematipico suggested is the way to do. Would you be willing to resolve conflicts and make the changes required @msxdan?

msxdan:
> matthewp confirmed that what ematipico suggested is the way to do. Would you be willing to resolve conflicts and make the changes required @msxdan?

Sure, I'll review it and make the changes suggested

msxdan:
@matthewp @ematipico I'm unsure this can be handled inside the `astro/app` nor the middleware, when the server is created, the listener will serve all requests unless there's an error, when an error is reached it will then reach the SSR Handler (the middleware) and call the `match` method inside NodeApp


`http-server.ts`
image


Should we call the handler for each request? that way it will call the match function everytime, however I'm not sure that will work fine, the standalone mode is serving the content by itself unless there's an error

image

matthewp:
@msxdan I'm not sure I understand, that stream code is about serving files. It checks if there's a file and if not, it calls the SSR handler. But I'm not sure what this has to do with trailingSlash. Aren't we *already* calling into astro/app for the trailing slash, but astro/app is just not handling it?

msxdan:
@matthewp The astro/app was not called by default before serving the file, only when there's no existing file, I cannot add the trailingSlash logic inside `match` function if it's not called.

If you can tell me some suggestions about how it should work I could try to implement it.

matthewp:
@msxdan That's how it's supposed to work, it checks for files first, and if no files exist, it then calls Astro to render. Are you expecting Astro render to take precedent over files?

lilnasy:
Behavior for endpoints seems to still be mostly unaffected by trailingSlash.

| msxdan:feat/node-trailing-slash | pages                                        | endpoints                                  |
|------------------------------------|-----------------------------------|-----------------------------------|
| dev                                              | ✅404 without a trailing slash  | ❌ 404 without a trailing slash |
| dev                                              | ✅200 with a trailing slash       | ❌200 with a trailing slash        |
| build                                            | ✅404 without a trailing slash  | ❌ 404 without a trailing slash |
| build                                            | ✅200 with a trailing slash       | ✅200 with a trailing slash        |

| astro@4.0.7 @astrojs/node@7.0.2 | pages                                    | endpoints                                  |
|------------------------------------|-----------------------------------|-----------------------------------|
| dev                                              | ❌404 without a trailing slash  | ❌ 404 without a trailing slash |
| dev                                              | ✅200 with a trailing slash       | ❌200 with a trailing slash        |
| build                                            | ❌404 without a trailing slash  | ❌ 404 without a trailing slash |
| build                                            | ✅200 with a trailing slash       | ❌200 with a trailing slash        |

florian-lefebvre:
@msxdan are you still interested in this PR?

msxdan:
> @msxdan are you still interested in this PR?

Yes I am, sorry I've been busy all these days

msxdan:
> @msxdan That's how it's supposed to work, it checks for files first, and if no files exist, it then calls Astro to render. Are you expecting Astro render to take precedent over files?

@matthewp about our last conversation

The static files are served from the http-server and we have no context about the manifest and hence the trailingSlash, that's why I added the trailingSlash inside Options

I tried to use the render function inside the astro core but it is unable to render static files or I don't know how it works, so since the App or NodeApp has the trailingSlash info I don't know how to implement it inside the http-server, and the nodeMidleware cannot render the files so I'm a bit stuck

matthewp:
@msxdan Let's use an example to talk through. Let's say you have these files:

- public/penguin.png
- src/pages/[...page].astro

If page renders were prioritized then it would be impossible to ever get the image at `/penguin.png`. That's why we check static files first.

If you'd like to change this to make files be checked last, we can talk about doing that through a new option perhaps, but that's different from getting trailingSlash to work.

Could create a reproduction with [astro.new](https://astro.new/latest) that shows `trailingSlash: ignore` not working?

packages/integrations/node/src/http-server.ts
```diff
@@ -23,12 +25,55 @@ function parsePathname(pathname: string, host: string | undefined, port: number)
...
+						res.setHeader('Location', pathname);
+					} else
+						pathname = urlPath;
+				break;
+			}
+			pathname = removeBase(pathname);
+
+			if (urlQuery && !pathname.includes('?')) {
+				pathname = pathname + '?' + urlQuery;
+			}
```
lilnasy, line 84: Can you elaborate on this line?

If we add the query right before `send` handles the pathname, wouldn't it prevent `send` from matching files?

lilnasy:
The node adapter just had a major overhaul. I have rebased this PR on top of the changes introduced.

Original commits:
- [feat(node): add trailing slash support](https://github.com/withastro/astro/commit/ddf320e77b522631ae186d6ca3ab9752194bda71)
- [add changeset](https://github.com/withastro/astro/commit/ccde7820f6c68c0187489effef81e5052664d1c8)
- [test(node): add base route test in trailing-slash.js](https://github.com/withastro/astro/commit/cd13dd653018d7a8ec45a13ab81abec1d155c0d8)
detected an infinite loop in base path when trailingSlash: never
- [fix(node): avoid infinite redirect when trailingSlash: never](https://github.com/withastro/astro/commit/c71720ac96e81b69f0918da4f532b4708c3080fa)

lilnasy:
!preview node-trailing-slash

ematipico:
@lilnasy 

I did some testing:
- In prod (`astro preview` / `node ./dist/server/entry.mjs`) it works as expected
- In dev, there are still issues. The dev server keeps adding a trailing slash regardless of the option passed via configuration

packages/integrations/node/src/http-server.ts
lilnasy: @msxdan I would be ready to approve after clarification on this

lilnasy: After my rebase, it was failing the tests with it.

lilnasy: That's a good point. I hadn't thought of it, but that's behavior I would want as well.

In the code, I see that the query data is added to the path right before performing a redirect. I see that you had written tests for it as well.

So, it looks good to me if it looks good to you.

lilnasy (approved):
Looks great to me! Thanks for the clean fix and thorough tests.

packages/integrations/node/src/http-server.ts
msxdan: Just checked, looks good to me as well!

ematipico (approved):
Happy to merge too. After more testing, I could see things working as expected

PR #9083, opened 2023-11-13
Refactor shikiji syntax highlighting code

bluwy:
## Changes

We used to have 3 places that handle shiki/ji syntax highlighting:

1. `Code.astro`
2. `remark-shiki` (markdown and MDX shared)
3. Markdoc

This PR combines all into a single API - `createShikiHighlighter`. You can now do something like this instead:

```js
const highlighter = await createShikiHighlighter({ /* options object */ })
const html = highlighter.highlight(code, 'js', { /* optional */ inline: true })
```

---

I also removed caching for the remark plugin and markdoc plugin for shiki. I don't think they're as useful as these plugins are usually only inited once and reused. Only `Code.astro` needs caching as the user can pass all kinds of configuration.

## Testing

Existing tests should all pass. And added new test for `createShikiHighlighter`.

I've also manually tests them, including the error overlay and markdoc, that they work.

## Docs

n/a. internal refactoring and shouldn't be visible to users


packages/astro/components/Code.astro
```diff
@@ -94,60 +93,13 @@ if (typeof lang === 'object') {
 
 const highlighter = await getCachedHighlighter({
 	langs: [lang],
-	themes: Object.values(experimentalThemes).length ? Object.values(experimentalThemes) : [theme],
+	theme,
+	experimentalThemes,
+	wrap,
 });
```
bluwy, line 99: With the abstracted function, we can pass the individual options directly, and it'll handle and process them internally.

packages/markdown/remark/src/shiki.ts
bluwy, line 1: Everything is moved to this file. Code is mostly taken from `Code.astro` which uses the `transforms` option to replace certain properties. Compared to other implementations that used regexes.

packages/markdown/remark/src/shiki.ts
```diff
@@ -0,0 +1,135 @@
...
+									innerSpanNode.children.unshift({
+										type: 'element',
+										tagName: 'span',
+										properties: { style: 'user-select: none;' },
+										children: [{ type: 'text', value: start }],
+									});
+								}
+							}
+						}
+					},
```
bluwy, line 101: Most of the code in this file is copied, except this part which is new. I ported it from the regex version.

https://github.com/withastro/astro/blob/5ef89ef33e0dc4621db947b6889b3c563eb56a78/packages/markdown/remark/src/remark-shiki.ts#L84-L90

It's slightly more code now, but with access to the `hast`, this should be a liitle more performant and robust.

ematipico (approved):
Looks good to me!

Issue #9059, opened 2023-11-10
`astro:assets` image service bundle regression


alexanderniebuhr:
```bash
git bisect start
# status: waiting for both good and bad commits
# bad: [1e97708cda779510d638abaefdb4abf707b697e3] fix: added onError functionality to spinner (#9048)
git bisect bad 1e97708cda779510d638abaefdb4abf707b697e3
# status: waiting for good commit(s), bad commit known
# good: [653ad939ce9f0ba12e2ce79a4cf9c154302b0bb9] [ci] release (#8989)
git bisect good 653ad939ce9f0ba12e2ce79a4cf9c154302b0bb9
# good: [4e63467d74e5b0b0eba1c80772e4daf64ff2b2d1] Remove empty changeset (#9030)
git bisect good 4e63467d74e5b0b0eba1c80772e4daf64ff2b2d1
# bad: [83c88706848521922ccd9cbdd914b653e47309db] [ci] release (#9021)
git bisect bad 83c88706848521922ccd9cbdd914b653e47309db
# bad: [3e1239e42b99bf069265393dc359bf967fc64902] Content Collection cache (experimental) (#8854)
git bisect bad 3e1239e42b99bf069265393dc359bf967fc64902
# good: [bd2d494c04ca48c6a496078b650e3c49bf243390] fix(config types): routingStrategy is optional (#9034)
git bisect good bd2d494c04ca48c6a496078b650e3c49bf243390
# good: [5b16619e4ad9bbbf7a1519899b7f5dd1b850562a] fix(i18n): names of the functions should match the RFC (#9035)
git bisect good 5b16619e4ad9bbbf7a1519899b7f5dd1b850562a
# first bad commit: [3e1239e42b99bf069265393dc359bf967fc64902] Content Collection cache (experimental) (#8854)
```

Starting from this [commit](https://github.com/withastro/astro/commit/3e1239e42b99bf069265393dc359bf967fc64902), the image service is (seemingly) getting wrongly bundled into SSR code even though it's unused, using the following "userland" image config.
https://github.com/withastro/adapters/pull/58/files

- the image service is (knowingly) incompatible with Cloudflare runtime,
- but they only use it on prerendered pages, so it should only be there at build
- so they set an empty endpoint (since that should be the only runtime code using the image service)

This worked in <3.4.4, broken in >3.5.0

https://github.com/alexanderniebuhr/astro-bug-debug

alexanderniebuhr:
```bash
❯ pnpm -v
[...]

PR #9087, opened 2023-11-13
fix(assets): image service bundling for pre-rendered pages non-node runtime regression

alexanderniebuhr:
## Changes

- fixes https://github.com/withastro/astro/issues/9059
- see Discord's #dev channel for more context, and [resolution](https://discord.com/channels/830184174198718474/845430950191038464/1173670300567744573)

## Testing

- the fix of the regression was tested manually and compared to `3.4.4` (known good) and `3.5.0` (known bad)
- the content collection cache implications were not tested, hopefully ci catches them

## Docs

- changeset

### DIFF
--- a/.changeset/red-houses-explode.md
+++ b/.changeset/red-houses-explode.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes the regression which broke bundling of image service for pre-rendered pages, which was introduced by [#8854](https://github.com/withastro/astro/pull/8854)
--- a/packages/astro/src/core/build/plugins/plugin-prerender.ts
+++ b/packages/astro/src/core/build/plugins/plugin-prerender.ts
@@ -12,7 +12,11 @@ function vitePluginPrerender(opts: StaticBuildOptions, internals: BuildInternals
 
 		outputOptions(outputOptions) {
 			extendManualChunks(outputOptions, {
-				before(id, meta) {
+				after(id, meta) {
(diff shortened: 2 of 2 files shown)


alexanderniebuhr:
* main
  * **PR #9087** <a href="https://app.graphite.dev/github/pr/withastro/astro/9087?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/withastro/astro/9087?utm_source=stack-comment).

natemoo-re (approved):
Works for me! I think the experimental content collections cache flag already has a number of bugs, so I’d rather prioritize fixing the regression. If we do discover that this breaks the flag, we can find a workaround for that in a follow-up PR.

ematipico (approved):

ematipico:
The thing is, not tests were broken after the regression. How can we prevent that this doesn't happen again?

alexanderniebuhr:
> The thing is, not tests were broken after the regression. How can we prevent this from happening again?

Please correct me if I'm wrong, but I think we would need to add a test using `astro build`, and testing for a special case, where the service is set and the endpoint is set to an empty file, so that image optimization only works on pre-rendered pages and not on SSR runtime?

cc @Princesseuh

alexanderniebuhr:
I will add a test in https://github.com/withastro/adapters/pull/58

PR #9089, opened 2023-11-13
fix(redirects): handle non-verbatim targets

### LINKED ISSUE #8914: Configured Redirects with more than two params don't expand properly
### Astro Info

```block
Astro                    v3.3.2
Package Manager          pnpm
                         @astrojs/mdx
                         @astrojs/react
```

_No response_

#### Context
Redirects configuration in `astro.config.mjs`

Tested usint the following configuration (however certainly not limited to this configuration):
- SSR mode 
- Node adapter

#### Bug
Redirects entries with more than one param (ie: '/ov/[page]/[...user]': '/tests/ov/[page]/[...user]') don't  work.
The expansion does not take place.

#### Example:

Given the following configuration:

```javascript
export default defineConfig({
  redirects: {
    '/ov/[page]/[...user]': '/tests/ov/[page]/[...user]', // does not work
    '/ot/[...user]': '/tests/ot/[...user]',               // works
  },
  integrations: [tailwind(), mdx(), react()],
  output: "server",
  adapter: node({
    mode: "standalone",
  }),
})
```

- http://localhost:4321/ot/some-page/some-user/test 
will redirect properly to
- http://localhost:4321/tests/ot/some-page/some-user/test

**While**
- http://localhost:4321/ov/some-page/some-user/test
will redirect without expanding the params to
- http://localhost:4321/tests/ov/[page]/[...user]

Given the configuration above,

- http://localhost:4321/ov/some-page/some-user/test
Should redirect to 
- http://localhost:4321/tests/ov/some-page/some-user/test
Instead of
- http://localhost:4321/tests/ov/[page]/[...user]

https://github.com/lilnasy/redirect-issue

### Participation

lilnasy:
## Changes
- Fixes #8914

## Testing
- Added 3 cases to [redirects.test.js](https://github.com/withastro/astro/pull/9089/files#diff-f0e8bf856624a391d496c98638e77e91ca1f9ad3a3c30d4d14ff722ff6e6da97)

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/core/redirects/helpers.ts
+++ b/packages/astro/src/core/redirects/helpers.ts
@@ -20,6 +20,7 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 	if (typeof routeData !== 'undefined') {
 		return routeData?.generate(data) || routeData?.pathname || '/';
 	} else if (typeof route === 'string') {
+		console.log({ data, pattern: route});
 		return route;
 	} else if (typeof route === 'undefined') {
 		return '/';
--- a/packages/astro/src/core/redirects/helpers.ts
+++ b/packages/astro/src/core/redirects/helpers.ts
@@ -20,7 +20,16 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 	if (typeof routeData !== 'undefined') {
 		return routeData?.generate(data) || routeData?.pathname || '/';
 	} else if (typeof route === 'string') {
-		console.log({ data, pattern: route});
+		// TODO: this logic is duplicated between here and manifest/create.ts
+		if (Object.keys(data).length > 0) {
+			let target = route
+			for (const param in data) {
+				const paramValue = data[param]!
+				target = target.replace(`[${param}]`, paramValue)
+				target = target.replace(`[...${param}]`, paramValue)
+			}
+			return target
+		}
 		return route;
 	} else if (typeof route === 'undefined') {
 		return '/';
--- a/packages/astro/src/core/redirects/helpers.ts
+++ b/packages/astro/src/core/redirects/helpers.ts
@@ -21,16 +21,13 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 		return routeData?.generate(data) || routeData?.pathname || '/';
 	} else if (typeof route === 'string') {
 		// TODO: this logic is duplicated between here and manifest/create.ts
-		if (Object.keys(data).length > 0) {
-			let target = route
-			for (const param in data) {
-				const paramValue = data[param]!
-				target = target.replace(`[${param}]`, paramValue)
-				target = target.replace(`[...${param}]`, paramValue)
-			}
-			return target
+		let target = route
+		for (const param of Object.keys(data)) {
+			const paramValue = data[param]!
+			target = target.replace(`[${param}]`, paramValue)
+			target = target.replace(`[...${param}]`, paramValue)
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/redirects/helpers.ts
```diff
@@ -20,6 +20,7 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 	if (typeof routeData !== 'undefined') {
 		return routeData?.generate(data) || routeData?.pathname || '/';
 	} else if (typeof route === 'string') {
+		console.log({ data, pattern: route});
```
lilnasy, line 23: image

florian-lefebvre:
@lilnasy what's the status of this PR? Did you encounter some blocker?

lilnasy:
It's just here to explain the code path that goes wrong.

packages/astro/test/redirects.test.js
```diff
@@ -68,10 +75,31 @@ describe('Astro.redirect', () => {
...
+				const app = await fixture.loadTestAdapterApp();
+				const request = new Request('http://example.com/source/x/y/z');
+				const response = await app.render(request);
+				expect(response.headers.get("Location")).to.equal("/not-verbatim/target3/x/y/z");
+			});
 		});
 	});
 
-	describe('output: "static"', () => {
+	describe.skip('output: "static"', () => {
```
ematipico, line 102: Is it intended? If so, could you add a comment or `TODO` to tell us when we should unskip it?

packages/astro/test/redirects.test.js
```diff
@@ -217,7 +245,7 @@ describe('Astro.redirect', () => {
 		});
 	});
 
-	describe('config.build.redirects = false', () => {
+	describe.skip('config.build.redirects = false', () => {
```
ematipico, line 248: Is it intended? If so, could you add a comment or `TODO` to tell us when we should unskip it?

packages/astro/src/core/redirects/helpers.ts
```diff
@@ -20,6 +20,16 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 	if (typeof routeData !== 'undefined') {
 		return routeData?.generate(data) || routeData?.pathname || '/';
 	} else if (typeof route === 'string') {
+		// TODO: this logic is duplicated between here and manifest/create.ts
+		if (Object.keys(data).length > 0) {
+			let target = route
+			for (const param in data) {
+				const paramValue = data[param]!
```
ematipico, line 27: nit: I don't like very much the use of `for..in`, we should use `for..of`

ematipico (approved):

packages/astro/src/core/redirects/helpers.ts
lilnasy: its not an array

packages/astro/test/redirects.test.js
lilnasy: good catch, this was just for testing locally, unskipped

packages/astro/src/core/redirects/helpers.ts
ematipico:
```js
const dataAsArray  =Object.entries(data);

if (dataAsArray.length >) {
	for..of
}
```

lilnasy: oh didnt see this, implemented using object.keys

natemoo-re (approved):
LGTM! Thanks for the fix

PR #9090, opened 2023-11-13
New events for Astro's view transition API

martrapp:
## Changes

Implementation for https://github.com/withastro/roadmap/discussions/740. 
The earlier sections of the discussion document are out of date and merely describe the discussions we have had and the options we have considered. The content from 11/5 onwards is what will be implemented in this PR.

It might be beneficial to look at the larger diff of router.ts with the "hide whitespace options".

No changeset right now. Is this a minor?

## Testing

The current tests still run without failures

## Docs

How I would like to write again that these are just bug fixes, but I suspect that this time there is really something to do with the documentation. Sarah, do you have a week or two please?

Here are some paragraphs (from a technical point of view) describing the structure of the events.

Coming soon: I'm also planning some paragraphs explaining what cool things you can do with these events. (User perspective)

# View Transition Events (Technical View)

## View Transition Processing (bigger picture)

There are currently four types of actions that trigger Astro's View Transitions:
1. cLicking on a link
2. submitting a form
3. triggering history navigation browser user interface buttons, keyboard shortcuts, or calls to `history` functions `back()`, `forward()`, or `go()`
4.  calling the function `navigate()`, which is provided by `astro:transitions/client`

Processing begins with a preparation phase that loads the DOM of the target page. Then the actual view transition begins, which takes a screenshot of the current view (green line), swaps the current DOM with the contents of the loaded DOM and then starts the visual transition from the old to the new view (red line). 

``` mermaid
graph
subgraph nav["Client-Side Navigation"]
  subgraph vtp["View Transition Promises"]
  direction LR
  g>updateCallbackDone]~~~h>ready]~~~i>finished]
  end
  subgraph p["Preparation Phase"]
    a{{astro:before-preparation}}-->
    b[["Loader()"<br>load DOM of target page]]-->
    c{{astro:after-preparation}}
  end
  subgraph vt["View Transition"]
    d([startViewTransition])
    subgraph s["Swap DOM Phase"]
      t{{astro:before-swap}}-->
      u[["Swap()"<br>update current DOM<br>with loaded DOM]]-->
      v([update history state<br>and scroll position])-->
      w{{astro:after-swap}}
    end
    subgraph f["Completion Phase"]
      x([Run scripts])-->
      y{{astro:page-load}}-->
      z([Announce Route])-->
      zz([await finished])-->
      zzz([Cleanup])
    end
  end
  p-->d==>s==>f
  d-.->vtp
end

linkStyle 13 stroke:#faa
linkStyle 12 stroke:#afa

style b stroke:#585,stroke-width:1px,color:#fff
style nav fill:#888,stroke:#555,stroke-width:4px,color:#fff
style p fill:#564,stroke:#555,stroke-width:4px,color:#fff
style vt fill:#456,stroke:#555,stroke-width:4px,color:#fff
style s fill:#457,stroke:#777,stroke-width:4px,color:#fff
style vtp fill:#467,stroke:#777,stroke-width:4px,color:#fff
style f fill:#654,stroke:#777,stroke-width:4px,color:#fff
style x stroke:#855,stroke-width:1px,color:#fff
style z stroke:#855,stroke-width:1px,color:#fff

style a fill:#fea,stroke:#ddd,stroke-width:2px,color:#000
style c fill:#fea,stroke:#ddd,stroke-width:2px,color:#000
style t fill:#fea,stroke:#ddd,stroke-width:2px,color:#000
style w fill:#fea,stroke:#ddd,stroke-width:2px,color:#000
style y fill:#fea,stroke:#ddd,stroke-width:2px,color:#000
```
In the completion phase, the newly added scripts are executed, the page title of the new route is announced for users of assistive technologies and some clean-up work is carried out at the end. 

During the view transition, a view transition object exists with three promises that are resolved or rejected at different points during the visual transition. Users can use these to trigger custom code. For details see the [View Transition documentation](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition) on MDN. 

## Events

The yellow blocks in the diagram above mark the positions in Astro's navigation processing at which events are triggered. There are five events.
* `astro:before-preparation` and `astro:after-preparation` at the beginning and end of the preparation phase.
* `astro:before-swap` and `astro:after-swap` at the beginning and end of the swap phase.
* `astro:page-load` during the closing phase.

The `astro:after-` events and the `astro:page-load` event are standard `Event` objects. Their main purpose is to allow the user code to react to navigation behavior at different points of processing. 

The `astro:before` events provide navigation-specific properties 
that show details of the processing. Some of these properties are even writable to give users control over the behavior of navigation processing.

All five events fire on `window.document`.

### The `astro:before-preparation` Event

``` TypeScript events.ts
readonly from: URL // The page where the navigation started 
to: URL // The destination of the navigation. 
direction: Direction | string // The values directly supported by Astro are 'forward' and 'backward', but this can be extended to other values. This property is writable.
readonly navigationType: NavigationTypeString // 'traverse' | 'push' | 'replace' 
readonly sourceElement: Element // If triggered by a link navigation, the anchor element. If triggered by form submission, the submitter (and if submitter is null, the form element). Can also be set via the sourceElement property of the options parameter on a call to navigate()
readonly info: any // If the transition was initiated by a call to navigate(), the value of options.info. Set to an empty object if undefined.
newDocument: Document // The DOM to be transitioned to. This can be an empty DOM if swap() manipulates the current DOM in place.
readonly formData: FormData // Automatically filled in if the navigation was triggered by a form. If the navigation was triggered by a call to navigate(), the value of options.formData.
loader: () => Promise<void>  // A function that sets event.newDocument to the contents of event.to.
```

While the event listener code must run synchronously, the loader function will perform asynchronous actions. 

The `loader` property initially holds Astro's built-in implementation for loading the contents of `event.to` into `event.newDocument`. An event listener might override the value of the `loader` property to define a completely independent implementation. But more often than not, a listener will keep the original value of `loader` and use it to define its own version. 

``` JavaScript listener.js
const originalLoader = event.loader;
event.loader = async () => {
  doSomeThingBefore(event);
  await originalLoader();
  doSomeThingAfter(event)
}
```

Several listener might cooperate and build a Chain of Responsibility by wrapping the loader in several layers like an onion.

``` mermaid
graph TD
subgraph a["render loading indicator"]
subgraph b["prefetch target images"]
subgraph c["optimize off-viewport transitions"]
subgraph d["original loader"]
end end end end

style a fill:#9bf,stroke:#555,stroke-width:2px,color:#000
style b fill:#8ae,stroke:#555,stroke-width:2px,color:#000
style c fill:#79d,stroke:#555,stroke-width:2px,color:#000
style d fill:#68c,stroke:#555,stroke-width:2px,color:#000
```

|  |  |
|---|---|
|Type|`astro:before-preparation` |
|TypeScript Type|`TransitionBeforePreparationEvent`
|Cancelable| yes|

### The `astro:before-swap` Event
Most of the properties of the `astro:before-swap` event are identical to those of the `astro:before-preparation` event. The differences are:
* `direction` is `readonly` in `astro:before-swap`
* `loader` is not defined in `astro:before-swap`
The `astro:before-swap`event offeres two additional properties:
``` TypeScript event.ts
swap: () => void // Initially, Astro's built-in implementation of the swap() operation. The task of the swap operation is to update the current DOM, typically to reflect the contents of event.newDocument.  
readonly viewTransition: ViewTransition // The object returned by startViewTransition().  
``` 

The `swap()` method of the `astro:before-swap` event is used in a very similar way as `loader()` of `astro:before-preparation` event. It is possible to override the built-in implementation by assigning to the `swap` property and it is possible to build chains with several wrappers. Other tan `loader` swap does *not* support asynchronous actions. Browser implementations of view transition put a rather rigid timout on the code that can run during swap. Long running code should be moved to the loader callback.     

|  |  |
|---|---|
|Type|`astro:before-swap` |
|TypeScript Type|`TransitionBeforeSwapEvent`
|Cancelable| no|

## Data flow between `astro:before-*` events:
The initial state of the `astro:before-swap` event is derived from the final state of the `astro:before-preparation` event of the same transition. Data set by a listener can be seen by subsequent listeners. Data entered during the `astro:before-preparation` dispatch can be seen during the `astro:before-swap` dispatch.  This also applies in particular to the (user) info object that exists in every event.


packages/astro/components/ViewTransitions.astro
```diff
@@ -4,9 +4,10 @@ type Fallback = 'none' | 'animate' | 'swap';
 export interface Props {
 	fallback?: Fallback;
 	handleForms?: boolean;
+	intraPageTransitions?: boolean;
```
martrapp, line 7: Previously, we had two passes through processing: navigation to different pages and a shortcut for navigation to locations on the same page. The routing processing is now unified, but it requires an explicit opt-in for view transitions to destinations on the same page.

packages/astro/components/ViewTransitions.astro
```diff
@@ -94,7 +98,7 @@ const { fallback = 'animate', handleForms } = Astro.props;
 				let action = submitter?.getAttribute('formaction') ?? form.action ?? location.pathname;
 				const method = submitter?.getAttribute('formmethod') ?? form.method;
 
-				const options: Options = {};
+				const options: Options = { sourceElement: submitter ?? form };
```
martrapp, line 101: sourceElement for Anchors and Forms, see https://chromium.googlesource.com/chromium/src/+/3719b4f51e4919413e8eb08dab383c5278c756a1

packages/astro/src/@types/astro.ts
```diff
@@ -82,17 +82,20 @@ export interface TransitionAnimationPair {
...
 }
 
+export type TransitionFreeDirectionalAnimations = Record<string, TransitionAnimationPair>;
+
 export type TransitionAnimationValue =
 	| 'initial'
 	| 'slide'
 	| 'fade'
 	| 'none'
-	| TransitionDirectionalAnimations;
```
martrapp, line 95: There may be a need for discussion here: Currently, animations can only define "forward" and "backward", which are rendered as "" and "back". If we want to be able to support more different navigation directions on a page, we need more names. I've called the old behavior "strict" and defined a "free" variant where you can define animations with any name that will make it into the stylesheet without being changed or deleted.

matthewp:
I can work on the docs for this one.

packages/astro/components/ViewTransitions.astro
matthewp: We currently do not do a VT at all on the same page. I'm thinking maybe that was a mistake. I think we should (in the future, not now) change it so that:

1. Hash clicks do not do VT.
2. Non-hash clicks to same page do.

This mirrors regular page navigation, if you click a link to the same page it does trigger page navigation, unless it's to a hash link.

Let me know if I am forgetting the reason we did this (probably).

packages/astro/components/ViewTransitions.astro
```diff
@@ -26,13 +27,15 @@ const { fallback = 'animate', handleForms } = Astro.props;
 <meta name="astro-view-transitions-enabled" content="true" />
 <meta name="astro-view-transitions-fallback" content={fallback} />
 {handleForms ? <meta name="astro-view-transitions-forms" content="true" /> : ''}
+{intraPageTransitions && <meta name="astro-view-transitions-intra-page" content="true" />}
 <script>
 	import type { Options } from 'astro:transitions/client';
 	import { supportsViewTransitions, navigate } from 'astro:transitions/client';
 	// NOTE: import from `astro/prefetch` as `astro:prefetch` requires the `prefetch` config to be enabled
+	// @ts-ignore
```
matthewp, line 35: Think this is a mistake? If you saw errors in your editor it's probably because you have not build the project since this feature was added.

packages/astro/src/@types/astro.ts
matthewp: What is the reason for the change? Does it relate to the events at all?

I'm not sure what other directions there could be in the future, do you have some in mind?

matthewp: Is it about the fish demo? If so, i don't think that should be part of our API. From out perspective, there's a forward direction (click links, submit forms, forward button) and back (back button / swipe), any other directions are just custom animations that could occur with either of the page directions. You can control the animations like the fish demo by appending your own attributes during the events.

Maybe this is not about the fish demo though, in which case I'll need to understand better :)

packages/astro/components/ViewTransitions.astro
martrapp: I remember we decided that we didn't need VT on "self links". 
With the `viewTransition` object now available via the `astro:before-swap` event, users can call `viewTransition.skipTransition()` to skip transitions. The default should indeed be: No VT to targets on the same page, unless the target has no hash. 

If we want to make this change, it will now be a one line change in `router:transform()`. Currently, I have left the semantics as they are so as not to introduce incompatibilities

packages/astro/src/@types/astro.ts
martrapp: Yes, I would like to have a way to support appending my own attributes :)

Let's look at this example: `... transition:name="main" transition:animate={...} `. 
I could simply write the CSS for a Chrome view transition myself:
```
[data-my-fancy-marker] ::view-transition-new(main) {...} 
```

But I was not able to write the CSS for the VT simulation, e.g. for Firefox, without Astro's help. How can I find the range value `astro-wtmfw64b-2` to write something like the one that is currently generated?
```
[data-my-fancy-marker] [data-astro-transition-fallback="old"][data-astro-transition-scope="astro-wtmfw64b-2"] { ...}
``` 
Is there a way to access the scope identifier from user code? Alternatively I could skip `transition:name` and `transition:animate` and directly set `view-transition-name` on the style-attribute. Then the simulation code could use 
```
[data-my-fancy-marker] [data-astro-transition-fallback="old"][style*="view-transition-name: main"]
```  
That might in fact also be a good way. The idea that the CSS for the real VT and the simulation should be generated from the same definitions might not always lead to good animations.

martrapp:
> I can work on the docs for this one.

I would be very happy if you would do that! It's always so grueling to write good documentation in a foreign language!

packages/astro/components/ViewTransitions.astro
matthewp: So the plan is for 4.0 to come out in ~1 month. Given that this would be released at earliest on 11/23, I'm thinking we shouldn't add this flag to only remove it in a couple of weeks. Instead we can keep intrapage navigation as it has always been and then in 4.0 default to allowing VT in 4.0.

martrapp: Fine with me!

packages/astro/components/ViewTransitions.astro
bluwy: Seems like I'm also getting this locally even after building the project. I don't mind adding the `@ts-ignore` here for now though

sarah11918:
Hey @martrapp and @matthewp ! I'm starting to dig into this to be ready for docs, and just a heads up that in addition to this now-complete picture of things as a whole, some things that will help docs fall into place:

- calling attention to **which of these events that users can tap into are new** (because I believe they are not all new). Highlighting what's new, and relating that to what existed before, is important to convey. (`astro:before-swap`, `astro:after-preparation`, `astro:before-preparation` are the only new events?) And, whether these new events have caused any changes to existing events.

- **a way to frame what this (now) gives you**: 
	- finer control over the whole process with more events during the whole navigation process to tap into (so, more choice of doing things at a more precise time)?
	- More strategically-placed events (so, now the actions you want triggered are at *better* parts in the cycle)?
	- Events specifically to correspond to exact things you want to do? (e.g. not just MORE events to hit, but some CUSTOM events that are intended for very specific use cases/actions?)

	Understanding not just the cycle as one big picture, but what is different/improved is always helpful for docs!

- And then as you already meantioned, Martin, some user examples showing how these new events can be used, which should include:
	- Things you can do now that you couldn't before
	- Things that you *would do differently now* because of the new events
	
The docs site itself will really just document *how things work* as we don't describe changes, or past version there. But for *updating existing docs* (esp. if recommendations are different now), the changeset, blog post, and for helping our support squad understand what's different so they can answer people's questions, paying some attention now to the differences is helpful.

matthewp:
@sarah11918 Yeah there's one case where you can do darkmode with one of the new events that might be cleaner than the existing `after-swap` suggest. I'll take that into account and update the docs accordingly.

matthewp:
I just added a draft PR of the updated docs and @martrapp's explanations above blows away mine :D https://github.com/withastro/docs/pull/5425

packages/astro/src/@types/astro.ts
martrapp: Managed to get it very similar:
```
---
import { ViewTransitions, createAnimationScope } from 'astro:transitions';

const { scope, styles } = createAnimationScope('animationName', { south, north });
---

  <main data-astro-transition-scope={scope}>
  
  <Fragment set:html={styles} />
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -65,6 +74,20 @@ export function renderTransition(
 	return scope;
 }
 
+export function createAnimationScope(
+	transitionName: string,
+	animations: Record<string, TransitionAnimationPair>
+) {
+	const hash = Math.random().toString(36).slice(2, 8);
+	const scope = `astro-${hash}`;
```
martrapp, line 82: How can I create a good identifier for the scope?

matthewp: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID would be ideal but i don't think we can depend on that yet. I think the way you did it is fine for now.

packages/astro/src/runtime/server/transition.ts
```diff
@@ -65,6 +74,20 @@ export function renderTransition(
...
+	const sheet = new ViewTransitionStyleSheet(scope, transitionName);
+
+	addPairs(animations, sheet);
+
+	const styles = `<style>${sheet.toString()}</style>`;
```
matthewp, line 87: I would probably not wrap it in the tag and just pass the raw styles back and let them do what they want with them. With the tags you wouldn't be able to just do `{styles}` because it would be escaped, would have to do `<Fragment set:html={styles} />` which feels a little worse to me than `<style is:inline>{styles}</style>`

martrapp: ha, copied and pasted without giving it much thought. I have to admit that the result looks more natural without style tags.  Nevertheless, `<style is:inline>{styles}</style>` did not work for me. Are there any special spells required here?

matthewp: Ah, looks like you need to do: `<style set:text={styles}></style>`. I'm not sure if not being able to pass an expression as a child is intentional or a bug. 

In any event, I think having the raw styles and then having the flexibility to do it a couple of different ways is still better. Curious what you think.

martrapp: The API is definitely a big improvement now over what I had before. And I'm getting closer :)  `<style set:text={styles} />` now works for me after I removed all quotes from the stylesheet (since set:text escapes quotes). Removing the quotes seems safe to me. It only affects the scope-id, "new" and "old".

packages/astro/src/@types/astro.ts
martrapp: Ok, now it works as suggested by Matthew. Thank you very much for this really improving suggestion!

martrapp:
Two new tech demos on [https://events-3bg.pages.dev/](https://events-3bg.pages.dev/): 
a) Loading indicator and 
b) `swap()` replacement that persists animation and iframe without resetting them.

martrapp:
@matthewp we haven't talked explicitly on the current state of the `options` parameter of `navigate()`, see packages/astro/src/transitions/types.ts:
`state` and `info` are perhaps worth mentioning: together with `history` we now support all properties of the `options` parameter of `navigation.navigate()`. Should we put a `_` in front of `formData` and `sourceElement` to mark them for internal use?

matthewp:
Do you mean like this?

```js
navigate("/url", { _formData: formData })
```

If so, no I don't  think we should do that. The alignment with `navigator.navigate()` is nice for people who know about it, but the majority will not and adding underscores just makes the API confusing imo.

martrapp:
> If so, no I don't think we should do that. The alignment with `navigator.navigate()` is nice for people who know about it, but the majority will not and adding underscores just makes the API confusing imo.

OK, fine with me!

matthewp (approved):

bluwy:
I only skimmed through the code, and I think Matthew has the best knowledge to review this, but I'm slightly concerned of the bundle size increase this feature brings. The code to create events are taking a lot for something that might not be used.

Testing locally with `examples/view-transitions`, the client build JS reports:

```
Before:  9.31 kB │ gzip: 3.42 kB
After : 11.72 kB │ gzip: 4.13 kB
```

It's around 20% size increase. I don't think this is a blocker for the PR, but something I thought to bring up.

.changeset/few-keys-heal.md
```diff
@@ -0,0 +1,16 @@
...
+
+``` javascript
+astro:before-preparation // Control how the DOM and other resources of the target page are loaded 
+astro:after-preparation // Last changes before taking off? Remove that loading indicator? Here you go!
+astro:before-swap // Control how the DOM is updated to match the new page
+```
+
+The `astro:before-*` events allow you to change properties and strategies of the view transition implementation.
+The `astro:after-*` events allow you to make changes to the current DOM before and after the transition. 
+Head over to [View Transition docs](https://docs.astro.build/en/guides/view-transitions/) to find out more about the events!
```
sarah11918, line 16:
```suggestion
Head over to docs to see [the full view transitions lifecycle](https://docs.astro.build/en/guides/view-transitions/#lifecycle-events) including these new events!
```
Maybe something like this? Going directly to the page and implying that they don't have to piece together new and old in their brains, but they'll see a complete picture there?

.changeset/few-keys-heal.md
```diff
@@ -0,0 +1,16 @@
...
+Three new events now complement the existing `astro:after-swap` and `astro:page-load` events:
+
+``` javascript
+astro:before-preparation // Control how the DOM and other resources of the target page are loaded 
+astro:after-preparation // Last changes before taking off? Remove that loading indicator? Here you go!
+astro:before-swap // Control how the DOM is updated to match the new page
+```
+
+The `astro:before-*` events allow you to change properties and strategies of the view transition implementation.
+The `astro:after-*` events allow you to make changes to the current DOM before and after the transition.
```
sarah11918, line 15: I'll note that saying it's for "before and after" is maybe confusing here, since we're trying to distinguish `before-` events from `after-` events?

The explanation on the docs PR refers to after events firing "when a particular phase has finished".

The distinction between these two here makes it sound like "before" is used for changing *how* a process occurs, but "after" sounds like it only changes WHAT gets written to the DOM. Is that correct? Are both of these correct? If so, maybe check that "tip" in the docs section for updating/nuance!

sarah11918 (commented):
Amazing work @martrapp ! The care and detail into this feature is truly outstanding!

Just a question to verify re: connecting this to what's in Docs proper, and suggested updating to a more direct link for the lifecycle events!

matthewp:
@bluwy Yeah I have similar concerns but I don't think that means we shouldn't release this. This is probably the biggest increase we've had to the router, but it's probably the last big one we'll ever do.

The initial version of the router was not configurable, but now with the addition of the events and `navigate()`, we should be able to prevent any future feature creep as the router now allows you to hook into each phase.

There might be some opportunities to trim down the size in the future as well.

.changeset/few-keys-heal.md
martrapp: Argh, 😄 that was really confusing, thanks for pointing it out! I like the wording in the "tip" much better. The after event is not about what is written to the DOM. Making changes to the DOM was just one example of what you can do when you know a phase is over.

sarah11918 (approved):
Just a formality, but Docs is happy!  🥳

PR #9091, opened 2023-11-13
feat(i18n): apply specific routing logic only to pages

ematipico:
## Changes

This PR addresses this new part of the RFC:https://github.com/withastro/roadmap/commit/c93015d44a8dc9e0794c4067899d8f587aaabf47

There have been reports where:
- the assets endpoint doesn't work properly
- users need to have services that aren't part of the routing logic 
- the RSS integration stopped working as expected

After chatting with Matthew, we were more convinced that i18n routing should only work on pages. This PR updates the logic by. 

I created some new logic inside the `Pipeline`, called hooks. Essentially, I created a generic way to run functions before before rendering a route. The reason why I created this generic logic is because I wanted to have a generic function inside my middleware, and then share the same logic among pipelines (dev, SSR and SSG).

This worked out pretty well and I am satisfied with the result.

## Testing

Added:
- a case for assets
- a new case with an endpoint that should render

## Docs

N/A but I will make a PR to reflect the change in logic


packages/astro/src/i18n/middleware.ts
```diff
@@ -83,3 +103,23 @@ export function createI18nMiddleware(
...
+function shouldSkipRoute(pathname: string, pageRoutes: RouteData[], locales: string[]) {
+	if (!pathname.length || pathname === '/') {
+		return false;
+	}
+
+	if (locales.some((locale) => pathname.includes(`/${locale}`))) {
+		return false;
+	}
+
+	return pageRoutes.some((route) => {
```
matthewp, line 122: Won't this return true if any routes are not the pathname?

matthewp (approved):
lgtm!

PR #9092, opened 2023-11-13
Fallback to tap prefetch strategy on slow connection

smitbarmase:
## Changes

1. Fallback to `tap` prefetch strategy on slow connection, instead of bailing.
2. Add `ignoreSlowConnection` option to prefetch API, which lets manually prefetch even on data saver mode or slow connection.

## Testing

Tested on test cases added in this PR https://github.com/withastro/astro/pull/8951

## Docs

Docs: https://github.com/withastro/docs/pull/5393

This PR also has changesets and JSDoc for review.

### DIFF
--- a/.changeset/neat-mangos-judge.md
+++ b/.changeset/neat-mangos-judge.md
@@ -2,5 +2,4 @@
 'astro': minor
 ---
 
-#9092 - Fallback to `tap` prefetch strategy on slow connection, instead of bailing.
-#9092 - Add `ignoreSlowConnection` option to prefetch API, which lets manually prefetch even on data saver mode or slow connection.
+Fallback to the `tap` prefetch strategy on slow connection, instead of bailing.
--- a/.changeset/neat-mangos-judge.md
+++ b/.changeset/neat-mangos-judge.md
@@ -1,0 +1,6 @@
+---
+'astro': minor
+---
+
+#9092 - Fallback to `tap` prefetch strategy on slow connection, instead of bailing.
+#9092 - Add `ignoreSlowConnection` option to prefetch API, which lets manually prefetch even on data saver mode or slow connection.
--- a/.changeset/neat-mangos-judge.md
+++ b/.changeset/neat-mangos-judge.md
@@ -2,4 +2,4 @@
 'astro': minor
 ---
 
-Fallback to the `tap` prefetch strategy on slow connection, instead of bailing.
+Changes the fallback prefetch behavior on slow connections and when data saver mode is enabled. Instead of disabling prefetch entirely, the `tap` strategy will be used.
--- a/packages/astro/src/prefetch/index.ts
+++ b/packages/astro/src/prefetch/index.ts
@@ -104,7 +104,7 @@ function initHoverStrategy() {
 			clearTimeout(timeout);
 		}
 		timeout = setTimeout(() => {
-			prefetch(href, { with: 'fetch', ignoreSlowConnection: true });
+			prefetch(href, { with: 'fetch' });
 		}, 80) as unknown as number;
 	}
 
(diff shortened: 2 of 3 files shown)


packages/astro/src/prefetch/index.ts
```diff
@@ -104,7 +104,7 @@ function initHoverStrategy() {
 			clearTimeout(timeout);
 		}
 		timeout = setTimeout(() => {
-			prefetch(href, { with: 'fetch' });
+			prefetch(href, { with: 'fetch', ignoreSlowConnection: true });
```
bluwy, line 107: I think we shouldn't `ignoreSlowConnection` for hover and viewport strategies. That makes the slow connection detection never work?

The "fallback to tap strategy" should already be handled in the tap strategy iiuc.

packages/astro/src/prefetch/index.ts
```diff
@@ -241,6 +242,12 @@ function elMatchesStrategy(el: EventTarget | null, strategy: string): el is HTML
 	if (attrValue === 'false') {
 		return false;
 	}
+
+	// Fallback to tap strategy if using data saver mode or slow connection
+	if ((attrValue != null || prefetchAll) && isSlowConnection()) {
+		return strategy === 'tap';
+	}
```
bluwy, line 249: I think this check could indirectly affect hover and viewport strategies hooking its initial listeners.

Some events like `mouseenter` and `mouseleave`, and `IntersectionObserver` can only be initialized on startup, which calls this function. And if on startup we found a slow connection, we will want to avoid this check because hover/viewport strategies could continue to work when they later have a stable connection.

I think the fix here is something like this:
```suggestion
	// Fallback to tap strategy if using data saver mode or slow connection
	if (strategy === 'tap' && (attrValue != null || prefetchAll) && isSlowConnection()) {
		return true;
	}
```

.changeset/neat-mangos-judge.md
```diff
@@ -0,0 +1,6 @@
+---
+'astro': minor
+---
+
+#9092 - Fallback to `tap` prefetch strategy on slow connection, instead of bailing.
+#9092 - Add `ignoreSlowConnection` option to prefetch API, which lets manually prefetch even on data saver mode or slow connection.
```
bluwy, line 6: I think you can create two changesets for this. We also usually follow a present tense here, so e.g. this can be splitted as:

```
Fallback to the `tap` prefetch strategy on slow connection, instead of bailing.
```

```
Adds a `ignoreSlowConnection` option to the `prefetch()` API to prefetch even on data saver mode or slow connection.
```

The PR number here is also not needed.

packages/astro/src/prefetch/index.ts
smitbarmase: You are right. I think I misunderstood how strategy works with a link. Fixing it.

bluwy (approved):
Thanks! Implementation looks clean. Fine with not having tests for now as it's tricky to test for slow connections.

.changeset/neat-mangos-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Fallback to the `tap` prefetch strategy on slow connection, instead of bailing.
```
sarah11918, line 5:
```suggestion
Changes the fallback prefetch behavior on slow connections and when data saver mode is enabled. Instead of disabling prefetch entirely, the `tap` strategy will be used.
```
Something like this that indicates this is a change is helpful, I think!

sarah11918 (approved):
Happy with these changesets as long as the one that changes the prefetch fallback starts with "Changes..." and describes a change from old behaviour to new!  :raised_hands:

.changeset/neat-mangos-judge.md
smitbarmase: Looks good to me! 🙌

PR #9098, opened 2023-11-14
Support Svelte 5 (experimental)

bluwy:
## Changes

Add experimental support for Svelte 5. This PR needs more testing, but seems to cover the common cases.

NOTE: When testing this locally, the `svelte` devdep in `@astrojs/svelte` needs to be bumped to 5.0.0-next.1.

## Testing

n/a. manual testing for now until Svelte 5 is stable.

## Docs

Updated README to note supported Svelte versions

### DIFF
--- a/packages/integrations/svelte/src/index.ts
+++ b/packages/integrations/svelte/src/index.ts
@@ -5,13 +5,13 @@ import type { AstroIntegration, AstroRenderer } from 'astro';
 import { fileURLToPath } from 'node:url';
 import type { UserConfig } from 'vite';
 
-const isSvelte5 = Number.parseInt(VERSION.split('.').at(0)) >= 5;
+const isSvelte5 = Number.parseInt(VERSION.split('.').at(0)!) >= 5;
 
 function getRenderer(): AstroRenderer {
 	return {
 		name: '@astrojs/svelte',
 		clientEntrypoint: isSvelte5 ? '@astrojs/svelte/client-v5.js' : '@astrojs/svelte/client.js',
-		serverEntrypoint: '@astrojs/svelte/server.js',
+		serverEntrypoint: isSvelte5 ? '@astrojs/svelte/server-v5.js' : '@astrojs/svelte/server.js',
 	};
 }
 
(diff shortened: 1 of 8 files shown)


packages/integrations/svelte/src/index.ts
```diff
@@ -1,13 +1,16 @@
 import type { Options } from '@sveltejs/vite-plugin-svelte';
 import { svelte, vitePreprocess } from '@sveltejs/vite-plugin-svelte';
+import { VERSION } from 'svelte/compiler';
 import type { AstroIntegration, AstroRenderer } from 'astro';
 import { fileURLToPath } from 'node:url';
 import type { UserConfig } from 'vite';
 
+const isSvelte5 = VERSION.startsWith('5');
```
natemoo-re, line 8: Paranoid future-proofing.

```suggestion
const isSvelte5 = Number.parseInt(VERSION.split('.').at(0)) >= 5;
```

packages/integrations/svelte/src/index.ts
```diff
@@ -37,9 +40,14 @@ async function getViteConfiguration({
 }: ViteConfigurationArgs): Promise<UserConfig> {
 	const defaultOptions: Partial<Options> = {
 		emitCss: true,
-		compilerOptions: { dev: isDev, hydratable: true },
+		compilerOptions: { dev: isDev },
 	};
 
+	if (!isSvelte5) {
+		// @ts-ignore
+		defaultOptions.compilerOptions.hydratable = true;
```
natemoo-re, line 48: Interesting! Hydration is removed in Svelte 5? Is there any way to use hydratable Svelte components (4.0 or earlier) in a Svelte 5 project?

natemoo-re (commented):
Looks good so far! Bummer that slots continue to be hacky, but I guess Astro's use case is quite niche.

packages/integrations/svelte/src/index.ts
bluwy: Hydratable is enabled by default in Svelte 5. If you pass `hydratable: true`, it'll show a warning that you don't have to pass it now. I'll add a note here.

bluwy:
Ready for review! The new `-v5.js` files are mostly copies of the non`-v5.js` files, but with Svelte 5 code. Mainly because Svelte 5 needs to import certain modules that doesn't exist in Svelte 4, so it needs a separate file.

matthewp (approved):

natemoo-re (approved):
Nice! Everything LGTM. This is very similar to how we've handled majors for React in the past.

sarah11918 (approved):
LGTM @bluwy !

PR #9099, opened 2023-11-14
i18n(domains): validation and updated logic

ematipico:
## Changes

This PR starts the work around domain support. Here are the changes:

- updated the schema with proper validation for the `domains` configuration;
- updated the Astro features with the `domains` support capability. I am not aware of any esoteric case, so notifying Astro with `SupportKind` should be enough
- updated the `getAsbsoluteLocaleUrl` function, which now accepts `domains` and a `isBuild` flag. `isBuild` is required because we can only use the URLs with domains when building the website, otherwise the dev server is broken

Closes: 	PLT-994, PLT-992 and PLT-1206

## Testing

I added various cases to test the implemented features

## Docs

Added some basic documentation

### DIFF
--- a/packages/astro/src/i18n/vite-plugin-i18n.ts
+++ b/packages/astro/src/i18n/vite-plugin-i18n.ts
@@ -11,6 +11,8 @@ type AstroInternationalization = {
 export default function astroInternationalization({
 	settings,
 }: AstroInternationalization): vite.Plugin {
+	let isCommandBuild = false;
+
 	return {
 		name: 'astro:i18n',
 		enforce: 'pre',
@@ -18,6 +20,11 @@ export default function astroInternationalization({
 			if (id === virtualModuleId) {
 				return resolvedVirtualModuleId;
 			}
+		},
+
+		config(opts, { command }) {
+			isCommandBuild = command === 'build';
+			return opts;
 		},
 		load(id) {
 			if (id === resolvedVirtualModuleId) {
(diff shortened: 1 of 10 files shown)


packages/astro/src/i18n/vite-plugin-i18n.ts
```diff
@@ -52,13 +60,16 @@ export default function astroInternationalization({
...
+						isBuild,
 						...i18n,
 						...opts 
 					});
 					
 					export const getRelativeLocaleUrlList = (path = "", opts) => _getLocaleRelativeUrlList({ 
 						base, path, trailingSlash, format, ...i18n, ...opts });
-					export const getAbsoluteLocaleUrlList = (path = "", opts) => _getLocaleAbsoluteUrlList({ base, path, trailingSlash, format, site, ...i18n, ...opts });
+					export const getAbsoluteLocaleUrlList = (path = "", opts) => _getLocaleAbsoluteUrlList({ 
+						base, path, trailingSlash, format, site, isBuild, ...i18n, ...opts
```
matthewp, line 71: looks like `getAbsoluteLocaleUrlList` has not changed but is now being passed `isBuild`, what is this flag used for?

ematipico: I addressed your comment. Very silly from me, good catch. 

> isBuild, what is this flag used for?

From the PR description:

> updated the getAsbsoluteLocaleUrl function, which now accepts domains and a isBuild flag. isBuild is required because we can only use the URLs with domains when building the website, otherwise the dev server is broken

matthewp:
Needs a changeset.

ematipico:
> Needs a changeset.

I will add the changeset at the end. The PR is against the `feat/i18n-domain`

bluwy (approved):
One small nit, but looks good to me!

PR #9101, opened 2023-11-14
feat(i18n): add `Astro.currentLocale`

ematipico:
## Changes

RFC change: https://github.com/withastro/roadmap/pull/734/commits/f934880ab5835492e56ddbb69e3f32662310c8bf

This PR adds a new `Astro.currentLocale`, which is computed from `Request#url`, `i18n.locales`, `i18n.routingStrategy` and `i18n.defaultLocale`

If the routing strategy is `prefix-other-locales`, for any URL that doesn't contain a locale, we return the `defaultLocale`.

This PR adds a bunch of data that we pass down the rendering of the page/endpoint, very similar to the same work I did when implementing `preferredLocale`.

Closes PLT-1207

## Testing

I added new test cases that should cover new use cases.

## Docs

https://github.com/withastro/docs/pull/5410


packages/astro/src/core/endpoint/index.ts
```diff
@@ -83,6 +92,19 @@ export function createAPIContext({
...
+				return currentLocale;
+			}
+			if (locales) {
+				currentLocale = computeCurrentLocale(request, locales, routingStrategy, defaultLocale);
+				if (currentLocale) {
+					return currentLocale;
+				}
+			}
+
+			return undefined;
```
natemoo-re, line 106: I believe this could be reduced to the following?

```suggestion
				return currentLocale;
			}
```

natemoo-re (approved):
Code LGTM! Excited to see this API added!

packages/astro/src/@types/astro.ts
```diff
@@ -2112,6 +2112,11 @@ interface AstroSharedContext<
 	 */
 
 	preferredLocaleList: string[] | undefined;
+
+	/**
+	 * The current locale computed from the URL of the request. It matches the locales in `i18n.locales`, and returns `undefined` otherwise.
+	 */
+	currentLocale: string | undefined;
```
sarah11918, line 2119: Just checking that there will eventually be a corresponding docs entry in this file for this, too, like the other ones currently listed in configuration reference?

.changeset/quick-toes-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
sarah11918, line 2: Also checking that this is a patch, so only waiting for the feature to merge but not for the next minor release? If so, please help me remember when the corresponding docs PR should merge, since it's not on the normal release schedule!  :smile:  https://github.com/withastro/docs/pull/5410

packages/astro/src/@types/astro.ts
ematipico: I was hoping that this would be auto-generated, but it's not. I am not **when** we should update this https://docs.astro.build/en/reference/api-reference/#astro-global, but I suppose we should do it right before going out of experimental? 

Please let me know, and I will create a task for me to write the documentation

.changeset/quick-toes-peel.md
ematipico: Once the PR is merged, this new thing will go out in the **next** release (regardless of its entity). 

> remember when the corresponding docs PR should merge

Ideally, we should merge it before the next release. Although we don't have a schedule for that, I think it's also okay if the docs PR is released 1/2 day after.

packages/astro/src/@types/astro.ts
sarah11918: Ah, great point! 

To answer your question, No, these don't get auto-generated.

But, you made me realize that yes, this is not a **configuration** setting, and yes, should go on the `api-reference` instead, so no `@docs` are actually needed here in this file.

And yes, I agree that we'll keep it OFF the api reference page while it's still experimental, so nothing else required. That was my mistake!

bluwy (approved):

PR #9102, opened 2023-11-14
feat(dev-overlay): Hide plugins into a separate menu when there's too many enabled

Princesseuh:
## Changes

Add a new little guy menu when the user has too many plugins in order to not overcrowd the bar 
image

When any of the plugins in it have a notification, it'll show on the three dots, like so: 
image

## Testing

Tested manually for now, will be tested in a different PR

## Docs

N/A, plugins don't have anything to do, all the APIs work the same.


packages/astro/src/runtime/client/dev-overlay/overlay.ts
```diff
@@ -20,6 +23,7 @@ export class AstroDevOverlay extends HTMLElement {
 	plugins: DevOverlayPlugin[] = [];
 	HOVER_DELAY = 750;
 	hasBeenInitialized = false;
+	customPluginsToShow = 3;
```
Princesseuh, line 26: Currently, this is hardcoded. In the future, we plan to make it so it depends on the width of your screen. Unfortunately, this requires quite some refactoring in how the dev overlay UI is built to ensure compatibility with view transitions and stuff, so will tackle it later.

matthewp (approved):

bluwy (approved):

PR #9112, opened 2023-11-15
 feat(i18n): domain with lookup table 

ematipico:
## Changes

Relative RFC change: https://github.com/withastro/roadmap/commit/713fce74bb742f18ad985055758c3818b5eed9aa

This PR adds the basic logic for supporting the new routing strategy called `domain`. We require a new routing strategy because Astro needs to make assumptions** in order to provide the correct user experience.

This assumption is that the localised folders must be at the root - `src/pages` folder. With that, the logic works as follow: during the build phase, we add to the manifest a new information called `domainLookupTable`. This table is a map that will look like this:

```js
const domainLookupTable = {
	"https://fr.example.com": "fr",
	"https://example.pt": "pt",
}
```

It's essentially the configuration of `i18n.domains`, reversed :)

This table maps what we are supposed to get from the server to the correct route. When we extract the correct route, we can load and render the component for that route.

In a server environment, Astro will work with the following headers:
- [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host)
- [Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) as fallback of `X-Forwarded-Host`
- [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)

If Astro can't crate a `URL` from those heasers, it will return a 404. I believe this is an important information and I will make sure to add it to the RFC later.

I will have to do some testing, even on a serverless environment and check if that information is available. I want to do more testing **after** merging this PR.

Closes PLT-1218

## Testing

I added some basic testing to show the feature, although I want to add more tests **in another PR**.

## Docs

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1520,7 +1520,7 @@ export interface AstroUserConfig {
 			 * @docs
 			 * @kind h4
 			 * @name experimental.i18n.routingStrategy
-			 * @type {'prefix-always' | 'prefix-other-locales'}
+			 * @type {'prefix-always' | 'prefix-other-locales' | 'domain'}
 			 * @default 'prefix-other-locales'
 			 * @version 3.5.0
 			 * @description
@@ -1533,9 +1533,12 @@ export interface AstroUserConfig {
 			 *  - `prefix-always`: All URLs will display a language prefix.
 			 *    URLs will be of the form `example.com/[locale]/content/` for every route, including the default language.
 			 *    Localized folders are used for every language, including the default.
+			 *  - `domain`: SSR only, it enables support for different domains. When a locale is mapped to domain, all the URLs won't have the language prefix.
+			 *    You map `fr` to `fr.example.com`, if you want a to have a blog page to look like `fr.example.com/blog` instead of `example.com/fr/blog`.
+			 *    The localised folders be must in the `src/pages/` folder.
 			 *
 			 */
-			routingStrategy?: 'prefix-always' | 'prefix-other-locales';
+			routingStrategy?: 'prefix-always' | 'prefix-other-locales' | 'domain';
 
 			/**
 			 * @docs
@@ -1556,7 +1559,8 @@ export interface AstroUserConfig {
 			 *            locales: ["en", "fr", "pt-br", "es"],
 			 *            domains: {
 			 *                fr: "https://fr.example.com",
-			 *            }
+			 *            },
+			 *            routingStrategy: "domain"
 			 *        }
 			 *    }
 			 * })
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -31,6 +31,7 @@ import { matchRoute } from '../routing/match.js';
 import { EndpointNotFoundError, SSRRoutePipeline } from './ssrPipeline.js';
 import type { RouteInfo } from './types.js';
 import { shouldAppendForwardSlash } from '../build/util.js';
+import { normalizeTheLocale } from '../../i18n/index.js';
 export { deserializeManifest } from './common.js';
 
 const clientLocalsSymbol = Symbol.for('astro.locals');
@@ -136,12 +137,41 @@ export class App {
 		if (this.#manifest.assets.has(url.pathname)) return undefined;
 		let pathname;
 		if (this.#manifest.i18n && this.#manifest.i18n.routingStrategy === 'domain') {
-			const host = request.headers.get('X-Forwarded-Host');
-			if (host) {
+			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
+			let host = request.headers.get('X-Forwarded-Host');
+			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
+			let protocol = request.headers.get('X-Forwarded-Proto');
+			if (!host) {
+				// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host
+				host = request.headers.get('Host');
+			}
+			// If we don't have a host and a protocol, it's impossible to proceed
+			if (host && protocol) {
+				// The header might have a port in their name, so we remove it
+				host = host.split(':')[0];
 				try {
-					const locale = this.#manifest.i18n.domainLookupTable[host];
+					let locale;
+					const hostAsUrl = new URL('', `${protocol}://${host}`);
+					for (const [domainKey, localeValue] of Object.entries(
+						this.#manifest.i18n.domainLookupTable
+					)) {
+						// This operation should be safe because we force the protocol via zod inside the configuration
+						// If not, then it means that the manifest was tampered
+						const domainKeyAsUrl = new URL('', domainKey);
+
+						if (
+							hostAsUrl.host === domainKeyAsUrl.host &&
+							hostAsUrl.protocol === domainKeyAsUrl.protocol
+						) {
+							locale = localeValue;
+							break;
+						}
+					}
+
 					if (locale) {
-						pathname = prependForwardSlash(joinPaths(locale, this.removeBase(url.pathname)));
+						pathname = prependForwardSlash(
+							joinPaths(normalizeTheLocale(locale), this.removeBase(url.pathname))
+						);
 						if (url.pathname.endsWith('/')) {
 							pathname = appendForwardSlash(pathname);
 						}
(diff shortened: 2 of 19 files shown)


packages/astro/src/core/build/plugins/plugin-manifest.ts
```diff
@@ -234,6 +236,32 @@ function buildManifest(
...
+					.filter((segment) => segment !== pathnameLocale)
+					.slice(1)
+					.join('/');
+				if (shouldAppendForwardSlash(settings.config.trailingSlash, settings.config.build.format)) {
+					pathnameWithoutLocale = appendForwardSlash(pathnameWithoutLocale);
+				}
+				// We build a URL object like this: https://es.example.com/guides/getting-started
+				const domainPath = new URL(pathnameWithoutLocale, maybeLocaleDomain);
+				// We map  https://es.example.com/guides/getting-started -> /es/guides/getting-started
+				domainLookupTable[domainPath.toString()] = route.route;
```
matthewp, line 262: Hm, I don't think this will work for dynamic routes, right? In SSR essentially any route is possible, as long as there is a dynamic route that matches.

I was thinking this table would be a mapping of just the domain name part, to the associated locale. So for example:

```js
const domainLookupTable = {
	"https://fr.example.com": "fr",
	"https://en.example.com": "en-US",
}
```

I'm including the protocol because I think you said that it needs to be included.

So then during SSR it uses this lookup to see which locale the domain is for. And once it has the locale it uses that to call `matchRoute` will the locale as part of the pathname to get the correct route.

Does this make sense?

matthewp (commented):
Added a review comment, thinking that having a full lookup table for all routes won't work.

packages/astro/src/core/build/plugins/plugin-manifest.ts
ematipico: This would make the logic much simpler. Sounds good!

packages/astro/src/@types/astro.ts
```diff
@@ -1533,9 +1533,12 @@ export interface AstroUserConfig {
 			 *  - `prefix-always`: All URLs will display a language prefix.
 			 *    URLs will be of the form `example.com/[locale]/content/` for every route, including the default language.
 			 *    Localized folders are used for every language, including the default.
+			 *  - `domain`: SSR only, it enables support for different domains. When a locale is mapped to domain, all the URLs won't have the language prefix.
+			 *    You map `fr` to `fr.example.com`, if you want a to have a blog page to look like `fr.example.com/blog` instead of `example.com/fr/blog`.
+			 *    The localised folders be must in the `src/pages/` folder.
```
sarah11918, line 1538: I have some questions about how this works to be sure I understand:

1. if you select `routingStrategy: "domain"`, what happens to the URLs for your `defaultLocale`? Does it depend on whether you're using a `/[defaultLocale]/` folder structure? Does it make any assumptions by default?
2. Is it correct to assume that for non-default locales, which will all be written in `/locale/` folders, that the URLs will be prefixed if you don't specifically add them to `domains`? So, it's possible for a site to have a mixture of e.g. `example.com/fr/` and `es.example.com`?
3. is `domains` *ignored* if `routingStrategy` is set to one of the two other options? Could it cause problems to have `domains` configured if you do NOT have `domain` set as your routing strategy?
4. Does the order of `domains` and `routingStrategy` matter in the `i18n` configuration? Could someone run into trouble if these are not in the proper order?
5. Looking at the example below, `routingStrategy` is outside of the object containing the other config options. [In the docs](https://docs.astro.build/en/guides/internationalization/#configure-i18n-routing), we show it as inside this object. Can you confirm which should be correct?

Thanks!

sarah11918:
I'm also pointing out that #9099 was merged without my seeing it, which becomes confusing because now *those* docs changes do not show up as new here.

ematipico:
> I'm also pointing out that #9099 was merged without my seeing it, which becomes confusing because now _those_ docs changes do not show up as new here.

Don't worry Sarah, I haven't requested your review for that PR **on purpose**. I will loop you in when I require your review. These PR **aren't** published yet.

sarah11918:
But you did request me on *this* one which has those changes in it already, so I'm just point out that it's very difficult for me to look at these individual pieces in isolation like this. These concepts don't exist independently from each other.

packages/astro/src/@types/astro.ts
ematipico: 1. Yeah, there are assumptions. If you want to map a locale to a domain, the locale folder **must be** at `src/pages/[locale]`. The same goes for the `defaultLocale`. If the default locale **isn't** mapped to a domain, there aren't any assumptions. Although we are still evaluating the behaviour of `defaultLocale`, so it's possible we will change it soon. 
2. Yes, that is correct.
3. No, that will be an error. For example, if you have `i18n.domains` and `routingStategy: "prefix-always"`, Astro will throw a validation nerror.
4. No, the order doesn't matter.
5. Mmmmh, `routingStategy` is inside the `i18n`, same goes for `domains`. Are you referring to something else?

packages/astro/src/core/app/index.ts
```diff
@@ -131,8 +134,31 @@ export class App {
...
 		// ignore requests matching public assets
 		if (this.#manifest.assets.has(url.pathname)) return undefined;
-		const pathname = prependForwardSlash(this.removeBase(url.pathname));
-		const routeData = matchRoute(pathname, this.#manifestData);
+		let pathname;
+		if (this.#manifest.i18n && this.#manifest.i18n.routingStrategy === 'domain') {
+			const host = request.headers.get('X-Forwarded-Host');
+			if (host) {
+				try {
+					const locale = this.#manifest.i18n.domainLookupTable[host];
```
matthewp, line 142: Does the `domainLookupTable` contain the protocol? Does the `host` contain it? If not then they may not match here.

packages/astro/src/core/app/index.ts
```diff
@@ -131,8 +134,31 @@ export class App {
 		const url = new URL(request.url);
 		// ignore requests matching public assets
 		if (this.#manifest.assets.has(url.pathname)) return undefined;
-		const pathname = prependForwardSlash(this.removeBase(url.pathname));
-		const routeData = matchRoute(pathname, this.#manifestData);
+		let pathname;
+		if (this.#manifest.i18n && this.#manifest.i18n.routingStrategy === 'domain') {
+			const host = request.headers.get('X-Forwarded-Host');
```
matthewp, line 139: Is `X-Forwarded-Host` the only header to check here? I think that `X-Forwarded-Host` only occurs when there is a proxy server sitting in front. Should we fallback to `Host` itself?

ematipico: Yeah we really should. I will update. the logic

packages/astro/src/core/app/index.ts
ematipico: Good call out. I believe it doesn't have the protocol

packages/astro/src/core/app/index.ts
```diff
@@ -131,9 +135,61 @@ export class App {
...
 		if (this.#manifest.assets.has(url.pathname)) return undefined;
-		const pathname = prependForwardSlash(this.removeBase(url.pathname));
-		const routeData = matchRoute(pathname, this.#manifestData);
-		// missing routes fall-through, prerendered are handled by static layer
+		let pathname;
+		if (this.#manifest.i18n && this.#manifest.i18n.routingStrategy === 'domain') {
+			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
+			let host = request.headers.get('X-Forwarded-Host');
+			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
+			let protocol = request.headers.get('X-Forwarded-Proto');
```
matthewp, line 143: This is another one coming from the proxy. So if it's not a proxy (or the proxy doesn't forward this error), we should have a fallback. We can use the protocol of the URL here.

packages/astro/src/core/app/index.ts
```diff
@@ -131,9 +135,61 @@ export class App {
 		const url = new URL(request.url);
 		// ignore requests matching public assets
 		if (this.#manifest.assets.has(url.pathname)) return undefined;
-		const pathname = prependForwardSlash(this.removeBase(url.pathname));
-		const routeData = matchRoute(pathname, this.#manifestData);
-		// missing routes fall-through, prerendered are handled by static layer
+		let pathname;
+		if (this.#manifest.i18n && this.#manifest.i18n.routingStrategy === 'domain') {
```
matthewp, line 139: This block is a bit noisy, can it be moved into its own separate function so that the outer function is a little more succinct.

matthewp (approved):
lgtm!

Issue #5512, opened 2022-12-02
Base path not working like the docs said


harrytran998:
### What version of `astro` are you using?

latest

### Are you using an SSR adapter? If so, which one?

NodeJs 18 LTS + SSR nodejs standalone

### What package manager are you using?

yarn 3

### What operating system are you using?

Mac

### Describe the Bug

https://docs.astro.build/en/reference/configuration-reference/#base

The docs said it only affects production. But seem in development, this still have effect.

When I config `astro.config.js` like this

```js
import { defineConfig } from 'astro/config';
import svelte from '@astrojs/svelte';
import AstroNode from '@astrojs/node';

export default defineConfig({
  output: 'server',
  integrations: [svelte()],
  base: '/ok',
  server: {
    port: 4000,
    host: '0.0.0.0',
  },
  adapter: AstroNode({
    mode: 'standalone',
  }),
});

```
In dev mode, it will throw a warning like this

```bash
[serve]  404   /favicon.ico
[dev] Requests for items in your public folder must also include your base. /favicon.svg should be /ok/favicon.svg. Omitting the base will break in production.
```

And the screen shows like this

```
404: Not found
In your site you have your base path set to [/ok/](http://localhost:3000/ok/). Do you want to go there instead?
```

### Link to Minimal Reproducible Example

Stackblitz not work correctly with base path. So your team need to pull this source code 😆

### Participation

- [ ] I am willing to submit a pull request for this issue.

abubakarsohail:
@harrytran998 sorry for commenting on this but I am also facing the same issue. Why did you close this one?

ayush-lal:
@abubakarsohail, @harrytran998 what was the solution here?

lemonpole:
Sorry for posting on such an old thread but I was running into this issue as well and I worked around it by using a conditional.

If I'm in `dev`, I set `/` otherwise, I set my base path:

```js
base: import.meta.env.DEV ? '/' : "/my-sub-path",
```

https://docs.astro.build/en/guides/environment-variables/

PR #9115, opened 2023-11-16
Add new preferences module

natemoo-re:
## Changes

- Resolves PLT-1200, related to `experimental.devOverlay`
- Adds a new "user preferences" module. Whereas the `astro.config.mjs` file manages configuration for a specific project on any machine, user preferences manage configuration for a specific user (machine). 
  - By default, user preferences are scoped to a specific project and stored in the `.astro/settings.json` file.
  - User preferences can also be stored globally and apply to every Astro project. They are stored in an OS-specific location.
- **Why?** The goal here is to allow users to toggle the dev overlay on/off in a way that doesn't affect every other person that works on the project. Since we already automatically `.gitignore` the `.astro/` directory, this is a natural place to put settings for an individual user.
- Adds a new `astro preferences` CLI command, designed to mirror the [`pnpm config`](https://pnpm.io/cli/config) command.

```sh
# Disable the dev overlay for the current user in the current project
npm run astro preferences disable devOverlay
# Disable the dev overlay for the current user in all Astro projects on this machine
npm run astro preferences disable devOverlay --global

# Check if the dev overlay is enabled for the current user
npm run astro preferences reset devOverlay
```

## Open Questions

- [x] Should this module be in `astro` or a separate package? I could see this module being used in `@astrojs/telemetry` and `create-astro` to configure certain default behavior.
  - **Resolution:** We might want to do this in the future, but we don't have to do it now.
- [x] Should the CLI be called via `astro config`? I can see that being hard to document. I considered `astro prefs` or `astro settings` but both felt weird.
  - **Resolution:** `astro preferences` it is! `config` would have been confusing to users because of the `astro.config.mjs` file and `settings` would have been confusing for maintainers because there is an existing `AstroSettings` that means something else.
- [x] This is a new feature that increases the scope of the dev overlay. How do we message it as experimental? Should we flag the CLI command as `astro experimental-config`?
  - **Resolution:** YOLO.

## Testing

Tested manually because I'm not sure how pipe the mocked `fs` through the system just yet

## Docs

[#5512](https://github.com/withastro/docs/pull/5512) and added to the [Astro 4.0 breaking changes tracker](https://github.com/withastro/docs/issues/5390)


Princesseuh:
> Should this module be in astro or a separate package? I could see this module being used in @astrojs/telemetry and create-astro to configure certain default behavior.

I could definitely see this in a separate package, I notable can very easily see `create-astro` being interested in it (it could be neat if it remembered your usual settings? People don't start up projects all the time for sure, but maybe if you're always using strictest, it'd be cool if it defaulted to it, no idea.)

>  Should the CLI be called via astro config? I can see that being hard to document. I considered astro prefs or astro settings but both felt weird.

No particular thoughts, but I assume we don't want users to confuse it with `astro.config.mjs`, so maybe it needs a separate name? 

> This is a new feature that increases the scope of the dev overlay. How do we message it as experimental? Should we flag the CLI command as astro experimental-config?

Since the dev overlay is stable in like two weeks, maybe we just yolo it? I don't see us relying on this module for othe rthings soon anyway.

.changeset/gentle-cobras-wash.md
```diff
@@ -0,0 +1,17 @@
+---
+'astro': minor
+---
+
+Adds the `astro config` command to manage user preferences. User preferences are specific to individual Astro users, unlike the `astro.config.mjs` file which changes behavior for everyone working on a project.
+
+User preferences are scoped to the current project by default, stored in a local `.astro/settings.json` file. Using the `--global` flag, user preferences can also be applied to every Astro project on the current machine. Global user preferences are stored in an operating system-specific location.
```
Princesseuh, line 7: Really like this pattern, awesome work.

packages/astro/src/@types/astro.ts
```diff
@@ -1714,6 +1715,7 @@ export interface AstroAdapterFeatures {
 export interface AstroSettings {
 	config: AstroConfig;
 	adapter: AstroAdapter | undefined;
+	preferences: AstroPreferences;
```
Princesseuh, line 1718: It is incredibly funny to have config and preferences on a settings object, but I have no idea how we could make it clearer. (At least we're consistent, so it's not too bad!)

FredKSchott:
If we're bikeshedding on command name/API:
- I agree `astro config` sounds a lot like "astro config file" or "astro.config.js" and the distinction between project-level and user/machine-level would get totally lost.
- I like `astro settings` and more importantly the idea that "settings" is a term we use for global machine-level configuration. `preferences` or another word could work as well, it really just comes down to us choosing a word that makes sense AND we can say "X means global settings" in our terminology.
- I dislike the complexity of "devOverlay.enabled" and `true|false` in the CLI because it feels like a JSON implementation detail that's leaking. I'd prefer the following API:
  - `astro settings [enable|disable] [telemetry|dev-overlay]`
  - `astro settings set future-complex-item-name "value"`
  - `astro settings unset future-complex-item-name`
  - `astro settings list`
  - to be clear, only the first example in my list needs to be built today! Just thinking through this a bit more than I normally would this early since we're thinking about this as a more long-term, reusable API.

ematipico:
I firmly believe that we **shouldn't** provide a `--global` flag, especially because we suggest users to **not** install `astro` globally.

Princesseuh:
> I firmly believe that we **shouldn't** provide a `--global` flag, especially because we suggest users to **not** install `astro` globally. 

I don't really understand how the global config is related to installing Astro globally, what's the relation?

natemoo-re:
Thank you for the feedback @FredKSchott! I believe I've addressed all of that now.

- `astro settings` would have been a good name, but we have an existing `AstroSettings` object throughout the codebase that means something different. To avoid internal confusion, I settled on `astro preferences`.
- I agree with your thoughts on `[name].enabled true/false` feeling weird! I had originally considered a design like you suggested but wasn't sure if it was the right move. Since you brought it up, I think a subcommand interface will be much easier to interact with.
  - `astro preferences get [key]` will print the current value, or let you know the default if it's not set
  - `astro preferences set [key] [value]` will set the current value
  - `astro preferences enable [key]` will set a `${key}.enabled` value to `true`. This matches the existing `telemetry` command.
  - `astro preferences disable [key]` will set a `${key}.enabled` value to `false`. This matches the existing `telemetry` command.
  - `astro preferences reset [key]` will delete the value. `delete` is an alias. This matches the existing `telemetry` command.
  - `astro preferences list` will list all your current preferences. Using the `--json` flag will give a machine-readable output.

natemoo-re:
@ematipico I understand where you're coming from, but the ability to set a preference for all Astro projects is one of the explicit goals. I don't think users will confuse a `--global` flag with installing `astro` globally, but I could be wrong. 

For some additional context: the `preferences` CLI command mirrors what the existing `astro telemetry [disable|enable|reset]` command does, except `--global` was always implied in that case. The eventual goal is to unify these two commands under one `astro preferences` command, so we're going to need `--global` to make that happen. I also hope to support personalized defaults for `create-astro` in the future (also global).

ematipico:
> > I firmly believe that we **shouldn't** provide a `--global` flag, especially because we suggest users to **not** install `astro` globally. 
> 
> I don't really understand how the global config is related to installing Astro globally, what's the relation?


Having a local and a global configuration is a very risky business, because a **single** change in the configuration, even the addition of a new property, would mean that a previous version wouldn't be able to understand it. 

If this single change goes into the global configuration, the rest of the projects won't work anymore, unless they are upgraded. It would make DX a living hell.

I'm not sure we should allow a global configuration.

natemoo-re:
I'm still not sure I understand your specific concerns. The intention for this feature is to support global user preferences that impact every Astro project. Telemetry already works this way—if a user opts out of telemetry, every Astro project needs to respect that preference. Semver and backwards-compatibility are a requirement with the global preferences as they are with any feature.

sarah11918 (approved):
Docs look great for the changeset, Nate! We'll just need something added to the CLI reference page in docs itself!

ematipico:
I don't know how much we plan to rely on global settings in the future, so please take my takes with a grain of salt. 

> To handle versioning of config file changes I've seen done by having a "version" field in the config file. Since this file is edited by a CLI and not by hand, I think that handles the concern well.

I am aware of it, but this won't fix our issues.

Here's a practical example. You add a global `~/.astro/settings` file that will look like this (it's pseudo code). This property has a typo. This property is released in version 3.7.0.

```
version = 1
devOvelay = true
``` 

Upon revisiting the file, we discovered the issue and published a patch in `3.7.1`. Now, `3.7.1` **won't** be able to understand the contents in `~/.astro/settings` because we had a typo. It will create a new file from scratch, I suppose:

```
version = 1.1
devOverlay = true
```

Now **all** projects with version `3.7.0` can't understand the **new** config file.

___

The way I see it, the only way to land this feature without repercussions, is to make sure that Astro can understand **ALL** the versions of `~/.astro/settings` and provide a way to transition from one way to another, without changing the actual file. This isn't a solution though, it just mitigates the problem.

FredKSchott:
To @ematipico's concern, if Nate committed to long-term full backwards compatibility, would that address your concern? Ex: in your scenario above, the "fixed" version of Astro would write to both `devOvelay` and `devOverlay` indefinitely.

For context, this is not a module that we should expect a lot of usage and changes in. To look at its history, you could say that we introduced the first preference for telemetry in Astro 1.0, and now the second preference for the dev overlay in 4.0. I don't think there's any reason to think that this will accelerate, and if it did I imagine we could revisit this system in the future without being too blocked.

My main point would be that if your concern is with the larger concept of global configuration, that's probably a larger conversation given that we already have the concept of global configuration in Astro today powering the global telemetry opt-out.

natemoo-re:
Thanks for the elaboration @ematipico and for chiming in @FredKSchott! I understand the concern and agree that backwards-compatibility is essential for this to feature's success. Changes to the configuration would be handled with the same concern for proper semantic versioning as the rest of `astro`.

packages/astro/src/cli/preferences/index.ts
```diff
@@ -0,0 +1,150 @@
...
+			usage: 'set [key] [:value]',
+			tables: {
+				Flags: [
+					['--global', 'Change setting value globally.'],
+					['--help (-h)', 'See all available flags.'],
+				],
+			},
+			description: `Starts a local server to serve your static dist/ directory. Check ${cyan(
+				'https://docs.astro.build/en/reference/cli-reference/#astro-preview'
+			)} for more information.`,
```
bluwy, line 41: The `preference` command description here needs an update.

packages/astro/src/cli/preferences/index.ts
```diff
@@ -0,0 +1,150 @@
...
+	if (!isValidSubcommand(subcommand) || flags?.help || flags?.h) {
+		msg.printHelp({
+			commandName: 'astro preferences',
+			usage: 'set [key] [:value]',
+			tables: {
+				Flags: [
+					['--global', 'Change setting value globally.'],
+					['--help (-h)', 'See all available flags.'],
+				],
+			},
```
bluwy, line 38: When I run `pnpm astro preferences --help`, it only shows the flags but not the additional subcommands. Should we also list them here?

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -381,7 +381,8 @@ async function getScriptsAndStyles({ pipeline, filePath }: GetScriptsAndStylesPa
 			children: '',
 		});
 
-		if (settings.config.experimental.devOverlay) {
+		const devOverlayEnabled = await settings.preferences.get('devOverlay.enabled');
+		if (settings.config.experimental.devOverlay && devOverlayEnabled) {
```
bluwy, line 385: Question: If there's a scenario where multiple people are working on a project, a majority prefers to not have the overlay, but I prefer to have it, this condition makes it a little hard to work for me. Since I'd have to enable `experimental.devOverlay` still and inform everyone else to disable it in their preference.

I think I'd expect the inverse, where I can enable it in my preference so no changes are needed from everyone else. In which case, I think a condition like this instead:

```js
if (devOverlayEnabled ?? settings.config.experimental.devOverlay)
```

Where `devOverlayEnabled` is `boolean | undefined`, and `undefined` means unset, would make this work 🤔

packages/astro/src/cli/preferences/index.ts
```diff
@@ -0,0 +1,150 @@
...
+	return 1;
+}
+
+async function listPreferences(settings: AstroSettings, { location, json }: SubcommandOptions) {
+	const store = await settings.preferences.getAll({ location });
+	if (json) {
+		console.log(JSON.stringify(store, null, 2));
+	}
+	// TODO: pretty print
+	console.log(JSON.stringify(store, null, 2));
```
bluwy, line 148: Should this be an else-if?

packages/astro/src/cli/preferences/index.ts
natemoo-re: Yes we should! Thanks for catching that.

packages/astro/src/vite-plugin-astro-server/route.ts
natemoo-re: This feature makes more sense when you realize that the devOverlay will be enabled by default once we remove the experimental flag (happening in another PR)! I believe the preference will be the only way to turn it off, but I could be wrong about that.

packages/astro/src/cli/preferences/index.ts
natemoo-re: Whoops, yes. I think I meant to `return` in the `if`.

packages/astro/src/vite-plugin-astro-server/route.ts
bluwy: In that case, I think we can revisit the `boolean | undefined` approach then for future options 👍  If there will be a request for an explicit config option to disable the dev overlay in the future, we might have to tweak this to work differently.

Princesseuh: Yeah, me and Nate had a discussion about it yesterday, for this first version we're making it so the config always win (the idea is that if your project doesn't support the overlay for some reason, you can disable it for everyone). 

I don't necessarily have particular thoughts on what the behaviour should be, to be honest, open to whatever we deem is best.

natemoo-re: Correct, the resolution for now is that the config setting wins out and preferences can only _disable_ a feature. If we want to revisit that in the future, we definitely can.

PR #9118, opened 2023-11-16
feat: implement overlay main screen redesign

Princesseuh:
## Changes

Implements a redesign of the main screen, now showing the coolest integrations, more links to click on, more neat info (there's a button to copy `astro info` directly, it's cool!)

image

## Testing

Will be tested in details separately, much like other overlay PRs, however the current tests should already make sure that the window does work.

## Docs

WIP


natemoo-re:
CleanShot 2023-11-21 at 12 44 00@2x
Added icon fallbacks when the integration doesn't have an image

matthewp (approved):

Issue #9108, opened 2023-11-15
Setting a i18n.fallback option that maps to the defaultLocale leads to client-side js been removed.


kossidts:
### Astro Info

```block
Astro                    v3.5.4
Node                     v20.8.0
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/node
Integrations             @astrojs/react
                         @astrojs/preact
                         @astrojs/tailwind
                         @astrojs/sitemap
                         @astrojs/mdx
                         @astrojs/solid-js
```

### Describe the Bug

Setting the **i18n.fallback** option to match the **defaultLocale** in the configuration results in the removal of client-side JavaScript during the build process.

 At least client-side scripts are not sent to the browser and hence no client-side interaction possible.

```experimental: {
        i18n: {
            defaultLocale: "en",
            fallback: {
                de: "en", // <-- Issue: Client-side JS is removed
                // de: "fr", // Works
                // fr: "de", // Works
            },
            locales: ["en", "de", "fr"],
        },
    },
```

### What's the expected result?

The configuration should allow setting **i18n.fallback** to match **defaultLocale** without causing the removal of client-side JavaScript during the build. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-chtnz4?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
@kossidts thank you for reporting. I opened your reproduction but I am at a loss because I don't know how to replicate the issue and what I should expect. Could you please add some instructions that tells me how to replicate the issue in the reproduction, and what's the expected result **in the reproduction**?

kossidts:
Thank you, @ematipico, for your prompt response. Apologies for the oversight in my initial report. I've updated the reproduction code and the readme file to include some instructions for replicating the issue.

So please refer to the readme file for detailed instructions. But basically 
- Stop the the mode and run `npm run build && npm run preview`
- Observe that the button with a client-side click event functions correctly, changing the background color as expected.
- Stop the process and modify the astro.config.mjs by setting up the fallback to match the defaultLocale value i.e.
```js
i18n: {
    defaultLocale: 'en',
    fallback: {
      de: 'en', // issue removes client js
      // de: 'fr',
      // fr: 'de',
    },
    locales: ['en', 'de', 'fr'],
  },
```
- Now re-run `npm run build && npm run preview` and notice that the button ceases to function.

Here a record of reproduction.

[...]

PR #9119, opened 2023-11-16
fix(i18n): review fallback system

ematipico:
## Changes

This PR closes https://github.com/withastro/astro/issues/9108

The was a big flaw in the logic of the fallback, so I went back, and restored how the routes were stored before the introduction of the fallback logic.

The new change involves the introduction of `fallbackRoutes` as part of the `RouteData` object, very very similar to how we do it for `redirectRoutes`.

Although, now the generation of the pages is different, and each `RouteData` could create more than one page (itself + `fallbackRoutes`), hence I create a generator function called `eachRouteInRouteData`, which returns all the routes we need to create.

EDIT:

In SSR, the middleware already has the logic for doing the redirect, although we need to tell it somehow. So I updated the logic of `app/index.ts`. 

The `match` function now considers' fallbackRoutes' when matching a route. However, when returning a `Response`, we set a status code 302 because we want the fallback to apply.

Example:
- Fallback `it` to `en`
- Request a `/it/blog/article` route
-  We match the route `/en/blog/article`, because it contains a fallback route that has a pathname that matches `/it/blog/article`
- We then return a `302` status code. The middleware catches the status code and returns a redirect to `/en/blog/article`

## Testing

I added a new case where we check the presence of a hoisted script. 

The other rests should still pass.

### DIFF
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -325,6 +325,9 @@ async function generatePage(
 			: magenta('λ');
 	if (isRelativePath(pageData.route.component)) {
 		logger.info(null, `${icon} ${pageData.route.route}`);
+		for (const fallbackRoute of pageData.route.fallbackRoutes) {
+			logger.info(null, `${icon} ${fallbackRoute.route}`);
+		}
 	} else {
 		logger.info(null, `${icon} ${pageData.route.component}`);
 	}
@@ -346,6 +349,13 @@ async function generatePage(
 	}
 }
 
+function* eachRouteInRouteData(data: PageBuildData) {
+	yield data.route;
+	for (const fallbackRoute of data.route.fallbackRoutes) {
+		yield fallbackRoute;
+	}
+}
+
 async function getPathsForRoute(
 	pageData: PageBuildData,
 	mod: ComponentInstance,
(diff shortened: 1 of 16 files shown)


packages/astro/src/core/build/generate.ts
```diff
@@ -358,56 +368,65 @@ async function getPathsForRoute(
...
 
-				// The path was already built once. Check the manifest to see if
-				// this route takes priority for the final URL.
-				// NOTE: The same URL may match multiple routes in the manifest.
-				// Routing priority needs to be verified here for any duplicate
-				// paths to ensure routing priority rules are enforced in the final build.
-				const matchedRoute = matchRoute(staticPath, opts.manifest);
-				return matchedRoute === route;
-			});
+			paths.push(
```
ematipico, line 398: We chain all the generated routes all together

packages/astro/src/core/build/generate.ts
```diff
@@ -494,128 +513,130 @@ async function generatePath(pathname: string, gopts: GeneratePathOptions, pipeli
 	const manifest = pipeline.getManifest();
 	const { mod, scripts: hoistedScripts, styles: _styles, pageData } = gopts;
 
-	// This adds the page name to the array so it can be shown as part of stats.
-	if (pageData.route.type === 'page') {
-		addPageName(pathname, pipeline.getStaticBuildOptions());
-	}
-
-	pipeline.getEnvironment().logger.debug('build', `Generating: ${pathname}`);
+	for (const route of eachRouteInRouteData(pageData)) {
```
ematipico, line 516: The only relevant change is is that the whole logic is inside this loop now, and instead of doing `pageData.route`, we pass the `route` from this loop

matthewp (approved):

bluwy (approved):

packages/astro/src/i18n/middleware.ts
```diff
@@ -72,11 +72,13 @@ export function createI18nMiddleware(
 					});
 				}
 			}
+			console.log('here', response.status);
```
matthewp: couple of console.logs left in

packages/astro/src/i18n/middleware.ts
```diff
@@ -72,11 +72,13 @@ export function createI18nMiddleware(
...
 			if (response.status >= 300 && fallback) {
 				const fallbackKeys = i18n.fallback ? Object.keys(i18n.fallback) : [];
 
 				const urlLocale = separators.find((s) => locales.includes(s));
 
+				console.log('and here');
```
matthewp: another one

packages/astro/test/i18n-routing.test.js
```diff
@@ -890,11 +890,12 @@ describe('[SSR] i18n routing', () => {
 			expect(await response.text()).includes('Oi essa e start');
 		});
 
-		it('should redirect to the english locale, which is the first fallback', async () => {
+		it.only('should redirect to the english locale, which is the first fallback', async () => {
```
matthewp: remove this

packages/astro/src/core/routing/match.ts
```diff
@@ -2,7 +2,12 @@ import type { ManifestData, RouteData } from '../../@types/astro.js';
 
 /** Find matching route from pathname */
 export function matchRoute(pathname: string, manifest: ManifestData): RouteData | undefined {
-	return manifest.routes.find((route) => route.pattern.test(decodeURI(pathname)));
+	return manifest.routes.find((route) => {
+		return (
+			route.pattern.test(decodeURI(pathname)) ||
+			route.fallbackRoutes.some((fallbackRoute) => fallbackRoute.pattern.test(decodeURI(pathname)))
+		);
+	});
```
bluwy, line 10: While we're updating this, it would be nice if the `decodeURI(pathname)` part is extracted as a variable for a little more perf.

PR #9122, opened 2023-11-17
Support Vite 5

bluwy:
## Changes

- Bump to Vite 5 
- Bump to Rollup 4
- Update Vite plugins to support Vite 5
- Update `vite-plugin-astro-preview` to handle certain URLs that Vite 5 doesn't handle anymore

PS: I'm not sure why pnpm updated te lockfile to be larger. It should be fixed when we bump all the deps to latest later.

## Testing

Tested manually. Ran all tests locally which passed.

## Docs

I think we only need to mention this in the migration guide.


packages/integrations/svelte/src/index.ts
```diff
@@ -73,6 +73,7 @@ async function getViteConfiguration({
 	}
 
 	if (!resolvedOptions.preprocess && !(await svelteConfigHasPreprocess(root))) {
+		// @ts-expect-error there's a bug with the types where the first arg should be optional
 		resolvedOptions.preprocess = vitePreprocess();
```
bluwy, line 77: I've sent a PR upstream to fix it. If it's released, I'll update this part before merging.

packages/astro/src/core/preview/vite-plugin-astro-preview.ts
```diff
@@ -53,29 +52,49 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
...
+					const pathname = cleanUrl(req.url!);
+
+					// Vite doesn't handle /foo/ if /foo.html exists, we handle it anyways
+					if (pathname.endsWith('/')) {
+						const pathnameWithoutSlash = pathname.slice(0, -1);
+						const htmlPath = fileURLToPath(outDir + pathnameWithoutSlash + '.html');
+						if (fs.existsSync(htmlPath)) {
+							req.url = pathnameWithoutSlash + '.html';
+							return next();
 						}
```
bluwy, line 67: The comment above should clarify what this does. But to further explain why we're assigning to `req.url`, this is for a later Vite middleware to load the files. We only need to update the `req.url` so that middleware can detect a HTML file.

packages/astro/src/core/preview/static-preview-server.ts
```diff
@@ -72,8 +72,6 @@ export default async function createStaticPreviewServer(
 		host: getResolvedHostForHttpServer(settings.config.server.host),
 		port: settings.config.server.port,
 		closed,
-		// In Vite 5, `httpServer` may be a `Http2SecureServer`, but we know we are only starting a HTTP server
-		// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
 		server: previewServer.httpServer as http.Server,
```
Princesseuh, line 75: Is there a world where we'd want to support http for the preview server? Unrelated to this, just wondering.

packages/astro/test/fixtures/integration-add-page-extension/astro.config.mjs
```diff
@@ -1,4 +1,4 @@
-import { defineConfig } from 'rollup'
```
Princesseuh, line 1: lol

Princesseuh (approved):

packages/astro/src/core/preview/static-preview-server.ts
bluwy: We could! At the moment we have control over the config passed to Vite's `preview()` so we know this will always be a http server. If we open up to the config from `astro.config.mjs` in the future, they can create https servers.

delucis (approved):
LGTM! Thanks @bluwy 🚀

Issue #9106, opened 2023-11-15
Assets Cache-Control should be more aggressive


belluzj:
### Astro Info

```block
Astro                    v3.5.4
Node                     v18.16.1
System                   Windows (x64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/vue
```

### Describe the Bug

With the node adapter, it looks like static assets, with a hash in their filenames, get by default a max-age=0 cache control policy, which doesn't take advantage of hashing the names in the first place (from what I understand?).

Steps to reproduce:
- Run `astro build` then `node dist/server/entry.mjs`
- Look at assets in the Network Tab (filter by _astro/)
- The Cache-Control says `max-age=0`

![image](https://github.com/withastro/astro/assets/3616772/1370585b-ceb6-4c9a-b523-16b50e230b8f)

### What's the expected result?

The default max-age could be 1 year (or any very high value) because assets have a hash of their contents in the file name.

I tried following the JS code inside the node adapter to find out where the max-age=0 is coming from, but couldn't figure it out, so unfortunately I can't contribute a fix.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-zhox1r?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

belluzj:
This also prevents us from having a longer cache-control on the HTML, because just after deploying a new version of the site, we get in a situation where the HTML is from the previous version (cached), but the assets that go with it have disappeared (because they're not cached), so we get unstyled HTML.

It's also not possible to work around this ourselves easily while using the node adapter because the Astro middleware (were we could set some caching policy) doesn't seem to run around static assets (which is OK because apart from caching there's not much use for middleware on static assets).

matthewp:
Yeah I agree. I believe anything in the `_astro` folder can be given immutable cache headers. Would gladly accept a PR if you wanted to add this to the Node adapter!

belluzj:
Unfortunately I didn't manage to follow the code of the adapter, I tried but couldn't figure out where the max-age=0 comes from. I could have a second look later if no-one else is available to look into it.

matthewp:
I'm taking a look now to see if its a small change.

matthewp:
PR in https://github.com/withastro/astro/pull/9125

belluzj:
Thanks a lot @matthewp!

rishi-raj-jain:
This is great!

[...]

PR #9125, opened 2023-11-17
Support immutable cache headers for _astro assets

matthewp:
## Changes

- Adds an immutable cache header for assets inside of `_astro` in the Node adapter.
- Closes https://github.com/withastro/astro/issues/9106

## Testing

- Test case added
- Test to ensure that only stuff inside `_astro` is given the header.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/twelve-fishes-fail.md
+++ b/.changeset/twelve-fishes-fail.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/node': minor
+---
+
+Support immutable cache headers on \_astro assets
--- a/.changeset/twelve-fishes-fail.md
+++ b/.changeset/twelve-fishes-fail.md
@@ -2,4 +2,4 @@
 '@astrojs/node': minor
 ---
 
-Support immutable cache headers on \_astro assets
+Automatically set immutable cache headers for assets served from the `/_astro` directory
--- a/.changeset/twelve-fishes-fail.md
+++ b/.changeset/twelve-fishes-fail.md
@@ -2,4 +2,4 @@
 '@astrojs/node': minor
 ---
 
-Automatically set immutable cache headers for assets served from the `/_astro` directory
+Automatically sets immutable cache headers for assets served from the `/_astro` directory.
--- a/packages/integrations/node/README.md
+++ b/packages/integrations/node/README.md
@@ -190,6 +190,16 @@ In the case of multiple run-time variables, store them in a seperate file (e.g. 
 export $(cat .env.runtime) && astro build
 ```
 
+#### Assets
+
+In standalone mode assets in your `dist/client/` folder are served via the standalone server. You might be deploying these assets to a CDN, in which case the server will never actually be serving them, but in some cases such as intranet sites it's fine to serve static assets directly from the application server.
+
+Assets in the `dist/client/_astro/` folder are the ones that Astro has built. These assets are all named with a hash and therefore can be given long cache headers. Internally the adapter adds this header for these assets:
+
+```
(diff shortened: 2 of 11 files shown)


.changeset/twelve-fishes-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/node': minor
+---
+
+Support immutable cache headers on \_astro assets
```
natemoo-re, line 5:
```suggestion
Automatically set immutable cache headers for assets served from the `/_astro` directory
```

natemoo-re (approved):
Seems great!

packages/integrations/node/src/http-server.ts
```diff
@@ -54,6 +62,12 @@ export function createServer(
 				// File not found, forward to the SSR handler
 				handler(req, res);
 			});
+			stream.on('headers', (_res: http.ServerResponse<http.IncomingMessage>) => {
+				if(isImmutableAsset(encodedURI)) {
+					// Taken from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable
+					_res.setHeader('Cache-Control', 'public, max-age=604800, immutable')
```
bluwy, line 68: Perhaps we can also align the value as Vercel's  https://github.com/withastro/astro/blob/e3dce215a5ea06bcff1b21027e5613e6518c69d4/packages/integrations/vercel/src/static/adapter.ts#L128

bluwy (approved):

packages/integrations/node/src/http-server.ts
matthewp, line 68:
```suggestion
					_res.setHeader('Cache-Control', 'public, max-age=31536000, immutable')
```

matthewp:
@florian-lefebvre it can be yes, the Vercel adapter already has it.

florian-lefebvre:
Oh ok, looking at the issue I assumed it was concerning every adapter. It may be worth saying it on adapters docs, because I suspect many people (me included) add those headers to vercel/netlify config but if it's done by default, that's even better!

matthewp:
@florian-lefebvre Yeah, Astro core isn't responsible for serving static assets. I agree on the docs, I'll add that to this PR.

matthewp:
Waiting on docs review before merging.

packages/integrations/node/README.md
```diff
@@ -190,6 +190,16 @@ In the case of multiple run-time variables, store them in a seperate file (e.g.
 export $(cat .env.runtime) && astro build
 ```
 
+#### Assets
+
+In standalone mode assets in your `dist/client/` folder are served via the standalone server. You might be deploying these assets to a CDN, in which case the server will never actually be serving them, but in some cases such as intranet sites it's fine to serve static assets directly from the application server.
```
sarah11918, line 195:
```suggestion
In standalone mode, assets in your `dist/client/` folder are served via the standalone server. You might be deploying these assets to a CDN, in which case the server will never actually be serving them. But in some cases, such as intranet sites, it's fine to serve static assets directly from the application server.
```

.changeset/twelve-fishes-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/node': minor
+---
+
+Automatically set immutable cache headers for assets served from the `/_astro` directory
```
sarah11918, line 5:
```suggestion
Automatically sets immutable cache headers for assets served from the `/_astro` directory.
```

PR #9129, opened 2023-11-18
[next] Update error logging

FredKSchott:
## Changes

- New, more concise error log formatting
- Less extremely concise than what I'd originally pitched in #8833, more closely matches the current formatting of an error.
- After discussion with @Princesseuh and other maintainers, agreed to remove: File ID (already included in stacktrace), codeframe (often poorly formatted), full stack trace (now trimmed to just user code, by default).
- Updated one `hint` to be more concise, that @Princesseuh and I had already audited together in #8833 

### This PR

Screen Shot 2023-11-18 at 12 38 06 AM

### Previously

Screen Shot 2023-11-18 at 12 39 56 AM

^ didn't fit in my terminal! Once I dragged to expand and retake the screenshot:

Screen Shot 2023-11-18 at 12 40 09 AM

## Testing

N/A 

## Docs

N/A


.changeset/three-chairs-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Update error log formatting
```
matthewp, line 5: We like to make these changesets a paragraph or two, like what would go in a blog post.

packages/astro/src/cli/throw-and-exit.ts
```diff
@@ -20,7 +20,7 @@ export async function throwAndExit(cmd: string, err: unknown) {
 
 	const errorWithMetadata = collectErrorMetadata(createSafeError(err));
 	telemetryPromise = telemetry.record(eventError({ cmd, err: errorWithMetadata, isFatal: true }));
-	errorMessage = formatErrorMessage(errorWithMetadata);
+	errorMessage = formatErrorMessage(errorWithMetadata, true);
```
natemoo-re, line 23: Non-blocking, but it'd be nice to move to an options object like `{ verbose: true }` or `{ showFullStacktrace: true }` to avoid an unlabelled boolean argument.

natemoo-re (approved):
In favor of this change! Glad we talked through it, thanks for scoping it back a bit.

PR #9131, opened 2023-11-18
feat(examples): make content collections type explicit

LouisGariepy:
## Changes

Explicitly write out the content collection type in example projects. This only affects the `content` project type.

No changeset needed:
> Non-packages (`examples/*`) do not need changesets.

## Testing

No tests required for this as far as I know. Will add if needed.

## Docs

This doesn't affect users in a breaking way. It just makes the configuration explicit. This is particularly useful when generating templated projects from the examples.


sarah11918:
Thank you @LouisGariepy! I do think this is the fix to make, but we'll let the core devs confirm.  🙌

examples/portfolio/src/content/config.ts
```diff
@@ -1,6 +1,7 @@
 import { defineCollection, z } from 'astro:content';
 
 export const collections = {
+	type: 'content',
```
florian-lefebvre, line 4: It should be on line 5 instead

LouisGariepy: Thanks for the review! This should be fixed now.

bluwy (approved):
I suppose there's no harm in making it explicit. IIRC it being nullable was for compat reasons when we add the data type.

florian-lefebvre (approved):

PR #9105, opened 2023-11-15
Logging rewrite 2

FredKSchott:
**Note:** Waiting to enter `4.0` beta period before merging this PR. If it gets out of date with `main`, it can be rebased (by @FredKSchott if the merge conflicts are non-trivial).

## Changes

A re-attempt at https://github.com/withastro/astro/pull/8833, scoped down.  
See this comment for context: https://github.com/withastro/astro/pull/8833#issuecomment-1811977053

- A complete, holistic audit of all messages logged by the Astro CLI
- Clarity: rewrote some info/warn/error messages to be more clear and useful
- Consistency: reduced differences in logged icons, color, formatting, phrasing. For example, the color cyan is now (almost) always used for links. 
- Focus: removed some noisy startup logs that weren't useful
- Focus: removed distracting debug `[label]` information from most logs
- ~~Better Errors:~~ These changes introduced in https://github.com/withastro/astro/pull/8833 have been reverted, and are no longer included in this PR.
- Removed the unnecessary `(x2)`, `(x3)`, etc. effect in the dev server
- Added consistent request logging (including render time)

## Videos

https://github.com/withastro/astro/assets/622227/0f8c8034-f1e5-4696-bc95-e8235c0ffe78

https://github.com/withastro/astro/assets/622227/4a5b16b5-05d4-4ec8-8d25-c094e9391bd8

_Note: All error-related changes have been removed from this PR and are out-of-date in the videos above._

### Example: `(xN)` Formatting Weirdness

Rewriting the CLI output would cause weird formatting issues sometimes. In this example, resizing the CLI window did it.
Screen Shot 2023-10-13 at 2 13 24 PM

#### Before

Screen Shot 2023-10-13 at 2 13 52 PM

#### After

Screen Shot 2023-10-13 at 2 14 52 PM

### Build Output

Mainly stylistic, colors changed intentionally. 

#### Before
Screen Shot 2023-10-13 at 2 27 41 PM

#### After
Screen Shot 2023-10-13 at 2 18 16 PM

## Testing

- Tests updated (caught a small test error in the process)

## Docs

- N/A (to confirm with docs)

PR #9139, opened 2023-11-20
Rework Vite and Astro logger for HMR messages

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/8224

This PR removes the Vite `logLevel: 'warn'` default (now `info` instead), and reuse Vite's logger so we get properly HMR messages for files that actually changed.

This also refactors the logging flow as a follow-up from #9105

### Examples

`framework-vue` (dev). Updated files `src/components/Counter.vue`, `src/pages/index.astro`, `astro.config.mjs`, and then `package.json`, which triggered these logs:

<details>
<summary>Open screenshot</summary>
image
</details>

 > NOTE: there's a leading slash inconcistency for `[watch]` messages, seems to be intended in Vite. Leading slash is the URL, non-leading slash is for local files.

---

`framework-vue` (build):

<details>
<summary>Open screenshot</summary>
image
</details>

---

`core-image-errors` fixture, error is only logged once:

<details>
<summary>Open screenshot</summary>
image
</details>

---

### Notes

This change helps unblock using Vite's `bindCLIShortcut` utility for us to implement CLI shortcuts easily. I think it's good that plugins can log infos by default too, e.g. [vite-plugin-qrcode](https://github.com/svitejs/vite-plugin-qrcode/blob/382a75badabcdd592dfbad23ce6c6a599b32621b/packages/vite-plugin-qrcode/src/index.ts#L46) doesn't log anything today because we block `info` messages.

## Testing

Tested the logging manually. And ran the tests locally.

## Docs

The logging changes shouldn't need to be documented.


packages/astro/src/core/create-vite.ts
```diff
@@ -102,25 +102,13 @@ export async function createVite(
...
-	const viteCustomLogger: ViteLogger = {
-		...vite.createLogger('warn'),
-		// All error log messages are also thrown as real errors,
-		// so we can safely ignore them here and let the error handler
-		// log them for the user instead.
-		error: (msg) => logger.debug('vite', 'ERROR ' + msg),
-		// Warnings are usually otherwise ignored by Vite, so it's
-		// important that we catch and log them here.
-		warn: (msg) => logger.warn('vite', msg),
-	};
```
bluwy, line 114: I overlooked this part while reviewing https://github.com/withastro/astro/pull/9105.

Passing `vite.createLogger('warn')` directly has the issue where users are unable to set a different `logLevel` (even for us), which caused the client build to not log info messages anymore.

The `error` function here could also suppress errors that we don't want to suppress. Vite may sometime call `logger.error()` but doesn't throw any errors.

packages/astro/src/core/logger/vite.ts
```diff
@@ -0,0 +1,88 @@
...
+			if ((m = vitePageReloadMsg.exec(stripped))) {
+				if (isAstroSrcFile(m[1])) return;
+				const extra = m[2] ?? '';
+				astroLogger.info('watch', m[1] + extra);
+			}
+			// Rewrite HMR update message
+			else if ((m = viteHmrUpdateMsg.exec(stripped))) {
+				if (isAstroSrcFile(m[1])) return;
+				astroLogger.info('watch', m[1]);
+			}
```
bluwy, line 44: We log the HMR updates here instead, by rewriting what Vite would log.

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -108,12 +101,6 @@ export async function handleHotUpdate(
 		}
 	}
 
-	// TODO: Svelte files should be marked as `isSelfAccepting` but they don't appear to be
-	const isSelfAccepting = mods.every((m) => m.isSelfAccepting || m.url.endsWith('.svelte'));
-	if (!isSelfAccepting) {
-		logger.debug('watch', 'full page reload triggered');
-	}
```
bluwy, line 115: Not sure if we need this special log check. Besides the Svelte handling, all modules that are self-accepting doesn't necessarily trigger a full page reload 🤔

natemoo-re: I added this a long time ago to workaround a Svelte bug at the time. I'm sure it's out of date and no longer needed.

natemoo-re (approved):
This all looks quite reasonable!

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -197,13 +174,6 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
 		async handleHotUpdate(context) {
 			if (context.server.config.isProduction) return;
 			const filename = context.file;
-			const isSkipLog =
-				/astro\.config\.[cm][jt]s$/.test(filename) ||
-				/(\/|\\)\.astro(\/|\\)/.test(filename) ||
-				isPkgFile(filename);
-			if (!isSkipLog && dedupeHotUpdateLogs(filename)) {
-				logger.info('watch', filename.replace(config.root.pathname, '/'));
-			}
```
bluwy, line 206: This part can be removed as Vite now logs the actual changed files instead. The dedupe should also not be needed now as we log things through the Vite logger.

Princesseuh (approved):

packages/astro/src/core/create-vite.ts
FredKSchott: This was intentional! Every error that I encountered with Vite was also a thrown error, which caused double-logging on almost every error we encounter since Vite is the first one to hit it. 

For example, trying to import a file that doesn't exist results in different 3 error messages now instead of 1, because Vite logs two redundant error messages:

Screen Shot 2023-11-29 at 4 03 57 PM

To summarize the approach I went with:
- Our AstroError information is higher fidelity than Vite's logs, more helpful
- Our handling of Vite errors is good enough at turning them into AstroError
- If all relevant Vite error logs are also reported as AstroErrors, then Vite errors actually are just extra redundent info, useful for debugging but not needed in the console.
- Any Vite errors that aren't thrown / can't be handled by Astro can be handled individually in this logger and turned into error logs. But the model would be an  allowlist / opt-in instead of default logging all Vite errors to the console.

FredKSchott: Also, you should rebase/erge onto `next` which now includes https://github.com/withastro/astro/pull/9129 changes

packages/astro/src/core/logger/vite.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	const logger: ViteLogger = {
+		hasWarned: false,
+		info(msg) {
+			if (!isLogLevelEnabled(viteLogLevel, 'info')) return;
+
+			const stripped = stripAnsi(msg);
+			let m;
+			// Rewrite HMR page reload message
+			if ((m = vitePageReloadMsg.exec(stripped))) {
```
FredKSchott, line 35: The fact that some files are logged by Vite with a leading `/` and some are not is pretty odd and would confuse me as a user (especially on windows where I imagine one would have `/` and the other would have `\`). Does Vite consider this a bug?

I also don't love matching strings like this with regex, since these error messages aren't usually covered by semver and can change unexpectedly across different versions.

I think I prefer the solution that keeps handling this ourselves in `handleHotUpdate` where we can control things ourselves, for those reasons. But I also could be missing some benefits!

FredKSchott (commented):
I won't block this via GitHub, but I'm worried this is a step backwards on some of the improvements of #9105 and #9129.

I left some comments on the two specific spots! If there's a way that you can refactor & fix the original bug without impacting the logging behavior at all, then I'm +1 on merging! 

I can also make time early tomorrow morning my time to talk through this, if it would helpful to not block you. LMK!

packages/astro/src/core/create-vite.ts
bluwy: Thanks! Yeah I made sure that for Vite-turned-Astro errors, we also skip them at https://github.com/withastro/astro/blob/6e46fc0194601008f73614ae3248195b9d265444/packages/astro/src/core/logger/vite.ts#L75-L76

Last I tested it works locally so that it only logged once, so I hope it addresses the concern here.

bluwy: After rebasing, it seems likeit's showing what you have in your screenshot, gonna check this again

packages/astro/src/core/logger/vite.ts
bluwy: I've left a note in the description about the slash issue:

> NOTE: there's a leading slash inconcistency for [watch] messages, seems to be intended in Vite. Leading slash is the URL, non-leading slash is for local files.

I investigated that yesterday and it seems safer this way since you may sometime get `/@fs/some/fs/path` and it would not be safe to remove the initial slash.

If we put back in `handleHotUpdate`, we won't be able to fix https://github.com/withastro/astro/issues/8224 though. Since an update that calls `handleHotUpdate` doesn't always trigger HMR, e.g. if you're editing files that are not part of your app.

packages/astro/src/core/create-vite.ts
bluwy: Just fixed them in the latest commit, it looks like this now:
image

FredKSchott:
@bluwy can you share some examples of errors that are logged, but not thrown as errors? My theory is that all Vite errors are also thrown, so the Vite error log is unnecessary. If that's not true and you can share some situations where that's not true, I'll feel better about moving from `'debug'` to `'error'` as the default. Otherwise I'd still argue to keep `'debug'` as the default.

packages/astro/src/core/create-vite.ts
FredKSchott: Nice! I can reproduce the same, feel much better about that then.

FredKSchott:
Based on the new rebased branch I'm no longer -1. I'm still very nervous about so much string matching and the potential for those to drift in patch/minor releases, but I can stomach it if there's no better alternative.

It would be nice if this could encourage an effort to code errors and/or logs in Vite, so that this would be less flakey without needing to rely on string matching.

bluwy:
> @bluwy can you share some examples of errors that are logged, but not thrown as errors? My theory is that all Vite errors are also thrown, so the Vite error log is unnecessary. If that's not true and you can share some situations where that's not true, I'll feel better about moving from `'debug'` to `'error'` as the default. Otherwise I'd still argue to keep `'debug'` as the default.

The full list can be found with a [`logger.error(` search](<https://github.com/search?q=repo%3Avitejs%2Fvite%20logger.error(&type=code>) in the Vite repo. Some `logger.error()` calls aren't followed up with a throw since it's only logged. In brief, things like optimizer fails, http proxy, and websocket connection errors would be logged and it would be a bit dangerous to suppress them.

One example you can try is something like http://localhost:4321/@fs/Users/bjorn/.gitconfig (or similar) where you're accessing URLs outside your project:

<details>
<summary>Screenshot</summary>

image


</details>

> Frankly it looks like a weird formatting issue. But besides that, I think we shouldn't hide that error.

Another example is with a `@import "./non-existent";` in the Astro style tags:

<details>
<summary>Screenshot</summary>

image


</details>

Vite reports a better error than the default one.

FredKSchott (approved):
LGTM! Thanks for digging into the feedback @bluwy, I feel much better now (maybe even excited?) about the new approach, and the fact that we can always tweak this post 4.0 on a per-error basis.

PR #9140, opened 2023-11-20
Extend the 'click' listener of view transitions to handle links in SVGs and areas in image maps 

### LINKED ISSUE #9001: Using ViewTransitions API from SVG element
### Astro Info

```block
Package Manager          npm
Adapter                  none
                         @astrojs/sitemap
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

I have an SVG element wrapped with `<a href="..."></a>`. It seems that, in this case, a full page refresh happens all the time, instead of Astro intercepting the click event and transitioning between the two pages.

Reproducible code:
```javascript
<div class="p-5 min-w-[100%] md:min-w-[40%] overflow-visible">
  <svg>
    <a href=<some_page>>
      <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
    </a>
  </svg>
</div>  
```

What does fix it is if I intercept the `onclick` event and [call `navigate()`](https://github.com/withastro/astro/blob/main/packages/astro/src/transitions/router.ts#L433C17-L433C26).

```javascript
<script>
import { navigate } from "astro:transitions/client";
document.addEventListener(`click`, e => {
  e.preventDefault();
  const origin = e.target.closest(`a`);
  
  if (origin) {
    navigate(<some_page>);
  }
});
</script>
```

When clicking an `<a>` element, Astro is able to intercept the event and perform the transition between pages.

https://github.com/adrianAzoitei/svg-transitions

### Participation

martrapp:
## Changes

Currently we only trigger view transitions for clicks on or within HTMLAnchorElements. 
This excludes `<a href="" ...>` within SVG elements, which are technically not HTMLElements but SVGAElements, see #9001. This change fixes that issue and now also includes HTMLAreaElements, so you also get view transitions when you click inside your beloved image maps. 

Closes #9001

## Testing

Added e2e test cases 

## Docs

Most users will not recognize the difference between an `<a href=` inside or outside of `<svg>...</svg>`. So this part is not really an extension, but a bug fix. The support for image maps is an extension, but of a more theoretical nature. It is said that they were last used in the 90s.

### DIFF
--- a/.changeset/nine-houses-attend.md
+++ b/.changeset/nine-houses-attend.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+View Transitions: handle clicks on SVGAElements and image maps"
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -30,6 +30,7 @@ const { fallback = 'animate', handleForms } = Astro.props;
 	import type { Options } from 'astro:transitions/client';
 	import { supportsViewTransitions, navigate } from 'astro:transitions/client';
 	// NOTE: import from `astro/prefetch` as `astro:prefetch` requires the `prefetch` config to be enabled
+	// @ts-ignore
 	import { init } from 'astro/prefetch';
 
 	export type Fallback = 'none' | 'animate' | 'swap';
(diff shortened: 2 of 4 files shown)


matthewp (approved):
LGTM!

Issue #9096, opened 2023-11-14
Issues with page routing in `i18n`


itsmatteomanf:
### Astro Info

```block
Astro                    v3.5.3
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

The produced pages in the `dist/` are not what is expected.

If you have `prefix-other-locales`, you are still getting non-prefixed and prefixed files for the default language, which makes sense as they are redirecting to the root.

If you have `prefix-always`, you are getting a weird structure...

```
src/pages/
├─ it/
│  ├─ index.astro
├─ index.astro
```

```
dist/
├─ en/
│  ├─ it/
│  │  ├─ index.html -> /it/
│  ├─ index.html -> /
├─ it/
│  ├─ index.html -> actual content
├─ index.html -> /it

```

Removing the `src/pages/index.astro` file, doesn't solve it, just removes the `dist/en/index.html` file, but retains the `en/it/` folder. Trailing slashes don't seem consistent, either...

### What's the expected result?

```
dist/
├─ en/
│  ├─ index.html -> /it
├─ it/
│  ├─ index.html -> actual content
├─ index.html -> /it

```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-issue-9096

### Participation

- [ ] I am willing to submit a pull request for this issue.

itsmatteomanf:
Adding here, as it is probably related, if you have a nested `it/` folder it won't be even built. While a nested `en/` folder is correctly built. No fallback, anywhere.


https://stackblitz.com/edit/withastro-astro-issue-9096-2

itsmatteomanf:
Additional issue here, fallback seems to behave very weirdly.

I can't seem to get fallback pages in many, especially whenever you have `prefix-always` and some un-prefixed pages as the fallback ones use those and not the actually prefixed content.

This config:
```
experimental: {
	i18n: {
		defaultLocale: "it",
		locales: ["de", "it", "en"],
		routingStrategy: "prefix-always",
		fallback: {
			de: "it",
			en: "it",
		}
	}
}
```
With this page structure:
```
src/pages/
├─ it/
│  ├─ index.astro
├─ 404.astro
├─ index.astro
```
Produces:
```
dist/
├─ de/
│  ├─ 404/
│  │  ├─ index.html -> /404
│  ├─ index.html -> /
├─ en/
│  ├─ 404/
[...]

Issue #9095, opened 2023-11-14
Issues with redirects and page routing with `i18n`


itsmatteomanf:
### Astro Info

```block
Astro                    v3.5.3
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Building with this configuration causes the whole build to fail. The issue is with the `fallback` key, removing it, solves the issue.

```
experimental: {
	i18n: {
		defaultLocale: "it",
		locales: ["it", "en"],
		fallback: {
			en: "it",
		}
	}
}
```

[error.txt](https://github.com/withastro/astro/files/13349242/error.txt)

### What's the expected result?

Two folders, `it/` and `en/`, with all `en/` pages redirecting to `it/` if not present.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-issue-9095

### Participation

- [ ] I am willing to submit a pull request for this issue.

itsmatteomanf:
I think I figured out what the issue is here.

If you have a redirect for `/` it will fail.

I had, commenting that solves the issue:
```
redirects: {
	"/": "/it/",
},
```

PR #9142, opened 2023-11-20
fix(i18n): fallback should create consistent directories

ematipico:
## Changes

Fixes https://github.com/withastro/astro/issues/9096
Fixes https://github.com/withastro/astro/issues/9095
Closes PLT-1230

This PR refactors the fallback system again. There was a business logic flaw in how the routes and paths were computed, plus the flow of things could have been smoother, and that didn't help. 

When rendering a page, the flow **was** like this:
- we extract the styles/links/scripts
- generate the paths using `getStaticPaths`
- loop over the exploded paths
- call the `generatePath` function
- compute more styles and scripts
- create a request and render the response

Now the flow is adapted to the fallback, and a few steps have been moved:
- we extract the styles/links/scripts
- compute more styles and scripts
- **NEW** loop over the exploded routes (itself + fallbacks)
- loop over the exploded paths
- call the `generatePath` function
- create a request and render the response

The function `generatePath` now accepts the styles/scripts/links, and they are already computed; it only needs to render the `Response`.

The function `eachRouteInRouteData` is now called in a different place and its loop is way smaller.

## Testing

I added a new test case that should prevent a regression against the issues we're closing.

The rest of the tests should still pass.

## Docs

N/A

### DIFF
--- a/.changeset/curvy-sheep-lick.md
+++ b/.changeset/curvy-sheep-lick.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Consistely emit fallback routes in the correct folders.
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -13,6 +13,7 @@ import type {
 	MiddlewareEndpointHandler,
 	RouteData,
 	RouteType,
+	SSRElement,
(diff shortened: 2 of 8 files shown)


matthewp (approved):
so much simpler, great job!

PR #9143, opened 2023-11-20
i18n(domains): validation and updated logic (#9099)

ematipico:
## Changes

- What does this change?


.changeset/tidy-carrots-jump.md
```diff
@@ -0,0 +1,31 @@
+---
+'@astrojs/vercel': minor
+'@astrojs/node': minor
+'astro': patch
+---
+
+Add experimental support for i18n domain routing:
```
sarah11918, line 7:
```suggestion
Adds experimental support for a new i18n domain routing strategy (`"domain"`) that allows you to configure different domains for certain locales:
```

.changeset/tidy-carrots-jump.md
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+})
+```
+
+When this routing strategy is enabled and you run `astro build` command, the URLs emitted by `getAbsoluteLocaleUrl` and `getAbsoluteLocaleUrlList` will be different:
+
+- `/fr/about` will become `https://fr.example.com/about`;
+- `/es/about` will become `https://example.es/about`;
+- `/en/about` will **stay** as is because we haven't configured any domain;
```
sarah11918, line 31:
```suggestion
With `routingStrategy: "domain"` configured, the URLs for your built site, including the URLs emitted by `getAbsoluteLocaleUrl()` and `getAbsoluteLocaleUrlList()`, will follow the pattern set for each locale individually.

For any `locales` not configured in `domains`, the fallback URLs created will follow `prefix-other-locales`, and create a `/[locale]/` path for all non-default languages.

For the above configuration:

- The file `/fr/about.astro` will create the URL `https://fr.example.com/about`;
- The file `/es/about.astro` will create the URL `https://example.es/about`;
- the file `/ja/about/astro` will create the URL `https://example.com/ja/about`.
- The file `/en/about.astro` will create the URL `https://example.com/about`.
```

So, I'm not sure whether this is true! But, I think adding `"ja"` to the list of locales and showing what happens would be very helpful. OR, specify that you are only allowed to not specify the defaultLocale, if that's the case. I think only the one example of "en" isn't enough to establish the pattern because it happens to be the default locale, so it could be a special case.

sarah11918 (commented):
Did a quick review of the changeset first! (Have not yet looked at the API docs) because I want to make sure I understand what's happening here to get the API descriptions just right!

packages/astro/src/core/app/index.ts
```diff
@@ -138,13 +141,79 @@ export class App {
...
+				// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host
+				host = request.headers.get('Host');
+			}
+			// If we don't have a host and a protocol, it's impossible to proceed
+			if (host && protocol) {
+				// The header might have a port in their name, so we remove it
+				host = host.split(':')[0];
+				try {
+					let locale;
+					const hostAsUrl = new URL('', `${protocol}//${host}`);
```
matthewp, line 181:
```suggestion
					const hostAsUrl = new URL(`${protocol}//${host}`);
```

packages/astro/src/core/app/index.ts
```diff
@@ -138,13 +141,79 @@ export class App {
...
+					for (const [domainKey, localeValue] of Object.entries(
+						this.#manifest.i18n.domainLookupTable
+					)) {
+						// This operation should be safe because we force the protocol via zod inside the configuration
+						// If not, then it means that the manifest was tampered
+						const domainKeyAsUrl = new URL('', domainKey);
```
matthewp, line 187:
```suggestion
						const domainKeyAsUrl = new URL(domainKey);
```

matthewp (commented):
I think it's better to use the 1 arg form of `new URL` unless there's a reason to do it the other way that I can't think of?

.changeset/tidy-carrots-jump.md
ematipico: Re 1.: good call out, I belive adding one more non-confiured locale in the example is perfect. 

Re 2: 
> I'm not sure whether you'd prefer to describe this as a "redirect" or "a file in this location produces this URL". Either is fine!

I prefer `"a file in this location produces this URL"` because it's closer to the actual behaviour. **We don't redirects**; we make assumptions, and developers must follow those assumptions if they want to have this feature.

matthewp (approved):

matthewp (changes requested):
We're going to release this in 4.x

sarah11918 (commented):
Thanks for your help with this one, @ematipico ! I fixed some silly formatting/typos in the changeset, but I think it's now quite good!

Had a suggestion for each of the config items. They *might* be a bit verbose, but we can always refine them, and once we've finalized the guide, we'll know for sure which info we want there vs here. I think these are still the minimum we might want someone to see when they're looking things up in reference, though? Let's see what we think!

bluwy (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1017,6 +1017,17 @@ export const MissingIndexForInternationalization = {
 	hint: (src: string) => `Create an index page (\`index.astro, index.md, etc.\`) in \`${src}\`.`,
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
```
Princesseuh, line 1022: Missing a little description guy here

.changeset/tidy-carrots-jump.md
```diff
@@ -0,0 +1,46 @@
...
+getAbsoluteLocaleUrl("en", "about"); // will return "https://example.com/en/about"
+getAbsoluteLocaleUrl("fr", "about"); // will return "https://fr.example.com/about"
+getAbsoluteLocaleUrl("es", "about"); // will return "https://example.es/about"
+```
+
+For the above configuration:
+
+- The file `/fr/about.astro` will create the URL `https://fr.example.com/about`.
+- The file `/es/about.astro` will create the URL `https://example.es/about`.
+- The file `/en/about.astro` will create the URL `https://example.com/en/about`.
```
natemoo-re, line 46: I find this language a bit confusing since I don't know what "create the URL" means. 

I assume the intention is to describe that the server will handle requests to `https://fr.example.com/about` by rendering the `/fr/about.astro` file?

packages/astro/src/@types/astro.ts
```diff
@@ -1664,6 +1698,25 @@ export interface AstroUserConfig {
 		 * In the event of route collisions, where two routes of equal route priority attempt to build the same URL, Astro will log a warning identifying the conflicting routes.
 		 */
 		globalRoutePriority?: boolean;
+
+		/**
+		 * @docs
+		 * @name experimental.i18nDomains
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.1.0
```
natemoo-re, line 1707:
```suggestion
		 * @version 4.3.0
```

packages/astro/src/core/app/index.ts
```diff
@@ -171,13 +174,83 @@ export class App {
...
+						pathname = prependForwardSlash(
+							joinPaths(normalizeTheLocale(locale), this.removeBase(url.pathname))
+						);
+						if (url.pathname.endsWith('/')) {
+							pathname = appendForwardSlash(pathname);
+						}
+					}
+				} catch (e) {
+					// waiting to decide what to do here
+					// TODO: What kind of error should we try? This happens if we have an invalid value inside the X-Forwarded-Host and X-Forwarded-Proto headers
```
natemoo-re, line 245: Any thoughts here? Logging seems safest.

packages/astro/src/core/config/schema.ts
```diff
@@ -330,12 +333,22 @@ export const AstroConfigSchema = z.object({
...
 					])
 				),
+				domains: z
+					.record(
+						z.string(),
+						z
+							.string()
+							.url(
+								"The domain value must be a valid URL, and it has to start with 'https' or 'http'."
+							)
```
natemoo-re, line 343: I don't think this will filter out URLs that aren't using `https` or `http`, for the record. Probably fine, but we could add an additional validation if needed.

packages/astro/src/i18n/index.ts
```diff
@@ -100,7 +114,7 @@ export function getLocaleRelativeUrlList({
 	format,
 	path,
 	prependWith,
-	normalizeLocale = false,
+	normalizeLocale = true,
```
natemoo-re, line 117: Just confirming that this in an internal API and this won't break any public APIs we've exposed.

natemoo-re (approved):
A few suggestions but I'm going to preemptively approve! Great work getting this over the line.

packages/astro/src/core/app/index.ts
ematipico: Other than logging, should we render a 404 or a 500 page?

packages/astro/src/i18n/index.ts
ematipico: Yes, this is intentional. This is a bug! I will add a changeset for that

packages/astro/src/integrations/index.ts
```diff
@@ -503,11 +505,3 @@ export function isFunctionPerRouteEnabled(adapter: AstroAdapter | undefined): bo
...
 	}
 }
-
-export function isEdgeMiddlewareEnabled(adapter: AstroAdapter | undefined): boolean {
-	if (adapter?.adapterFeatures?.edgeMiddleware === true) {
-		return true;
-	} else {
-		return false;
-	}
-}
```
ematipico, line 513: This function is not used anywhere

sarah11918:
Heads up that I'd ideally like to polish up the docs here *after* the Docs PR is at least agreed upon in principle! https://github.com/withastro/docs/pull/6655

@ematipico when you're available, I've left feedback there! Let's make sure *that* is good to go, and when we have that, I'll be able to very quickly replicate anything that needs to be adjusted here.

packages/astro/src/@types/astro.ts
```diff
@@ -1530,8 +1530,42 @@ export interface AstroUserConfig {
...
+			 * 		locales: ["en", "fr", "pt-br", "es"],
+			 *		domains: {
+			 *			fr: "https://fr.example.com",
+			 *    },
+			 * 		routing: {
+			 * 			strategy: "domains"
+			 *		}
+			 *  }
+			 * })
+			 * ```
```
sarah11918, line 1563:
``````suggestion
			 * Configures the URL pattern of one or more supported languages to use a custom domain (or sub-domain). When a locale is mapped to a domain, a `/[locale]/` path prefix will not be used. However, localized folders within `src/pages/` are still required, including for your configured `defaultLocale`.
			 *
			 *  Any other locale not configured will default to a localized path-based URL according to your `prefixDefaultLocale` strategy (e.g. `https://example.com/[locale]/blog`).
			 *
			 *
			 * ```js
			 * //astro.config.mjs
			 * 	site: "https://example.com",
			 * 	output: "server", // required, with no prerendered pages
			 *   adapter: node({
			 *     mode: 'standalone',
			 *   }),
			 * 	i18n: {
			 * 		defaultLocale: "en",
			 * 		locales: ["en", "fr", "pt-br", "es"],
			 *     prefixDefaultLocale: false,
			 *		  domains: {
			 *			  fr: "https://fr.example.com",
			 *       es: "https://example.es"
			 *    },
			 *  experimental: {
			 *     i18nDomains: true
			 *  }
			 * })
			 * ```
			 *  
		   * Both page routes built and URLs returned by the i18n helper functions `getAbsoluteLocaleUrl()` and `getAbsoluteLocaleUrlList()` will use the options set in `i18n.domains`.
		   *  
		   * See the [Internationalization Guide](https://docs.astro.build/en/guides/internationalization/#domains-experimental) for more details, including limitations of this experimental feature.
``````

**Review by sarah11918 — COMMENTED** [review]
OK! These update to the agreed-upon configuration set up! I'm happy with the changeset as is, but still want to fine-tune the docs entry, especially because of the experimental-ness of it. (This entry doesn't explicitly mention the experimental flag, for example and we must do that!

I'm thinking that maybe we should put the docs entry under Experimental Flags. I think that makes the most organizational sense, and would make the instruction to enable a flag, then add a `domains` to a pre-configured i18n site easier to include here. Being experimental, this feature might change anyway. And even if it doesn't that means editing more than just the name of this entry to update config-reference here. So I'm not convinced we save much by having it here vs experimental.

Another option could be to write this entry here as if it were NOT experimental, just not expose it to docs. Then, add a complete corresponding entry as if it IS expermental in the Experimental flags section that DOES show up in docs. When it's no longer experimental, we entirely delete the experimental entry and (making sure it's updated), then add back `@docs` here. (If I'm writing this once, copy and paste is free!)

Thoughts?

File: packages/astro/src/@types/astro.ts [hunk 13]
```diff
@@ -1532,6 +1532,37 @@ export interface AstroUserConfig {
 			 * - `"pathanme": The strategy is applied to the pathname of the URLs
 			 */
 			strategy: 'pathname';
+
+			/**
+			 * @docs
```
sarah11918, line 1537:
```suggestion
```
I *think* simply deleting this will hide the entry! That's my intention, anyway, and this will be added back in when the feature is no longer experimental!

packages/astro/src/@types/astro.ts
```diff
@@ -1530,8 +1530,42 @@ export interface AstroUserConfig {
 			 * @description
 			 *
 			 * - `"pathanme": The strategy is applied to the pathname of the URLs
+			 * - `"domain"`: SSR only, it enables support for different domains. When a locale is mapped to domain, all the URLs won't have the language prefix.
+			 *    You map `fr` to `fr.example.com`, if you want to have a blog page to look like `fr.example.com/blog` instead of `example.com/fr/blog`.
+			 *    The localised folders be must in the `src/pages/` folder.
```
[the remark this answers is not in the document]
ematipico: Yeah, let's keep it hidden until it makes sense to document it :)

.changeset/tidy-carrots-jump.md
```diff
@@ -0,0 +1,53 @@
+---
+'@astrojs/node': minor
```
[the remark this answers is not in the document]
ematipico: I reworked the changesets. This particular changeset is for `astro` only. I created a new changeset for `@astrojs/node` and `@astrojs/vercel` with a small sentence.

packages/astro/src/@types/astro.ts
```diff
@@ -1532,6 +1532,45 @@ export interface AstroUserConfig {
 			 * - `"pathanme": The strategy is applied to the pathname of the URLs
 			 */
 			strategy: 'pathname';
+
+			/**
+			 * @name experimental.i18n.domains
```
sarah11918, line 1537:
```suggestion
			 * @name i18n.domains
```
Just confirming that this should (or shouldn't?) have experimental in the name, since it's not actually used/configured under an experimental flag?

packages/astro/src/@types/astro.ts
```diff
@@ -1664,6 +1703,48 @@ export interface AstroUserConfig {
...
+
+		/**
+		 * @docs
+		 * @name experimental.i18nDomains
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.3.0
+		 * @description
+	 	 *
+		 * Enables domain support for the [experimental `domains`(https://docs.astro.build/en/guides/internationalization/#domains-experimental) routing strategy] which allows you to configure the URL pattern of one or more supported languages to use a custom domain (or sub-domain).
```
sarah11918, line 1715:
```suggestion
		 * Enables domain support for the [experimental `domains` routing strategy](https://docs.astro.build/en/guides/internationalization/#domains-experimental) which allows you to configure the URL pattern of one or more supported languages to use a custom domain (or sub-domain).
```

packages/astro/src/@types/astro.ts
```diff
@@ -1532,6 +1532,45 @@ export interface AstroUserConfig {
...
+			 *		  domains: {
+			 *			  fr: "https://fr.example.com",
+			 *       es: "https://example.es"
+			 *    },
+			 * })
+			 * ```
+			 *  
+		   * Both page routes built and URLs returned by the `astro:i18n` helper functions [`getAbsoluteLocaleUrl()`](https://docs.astro.build/en/guides/internationalization/#getabsolutelocaleurl) and [`getAbsoluteLocaleUrlList()`](https://docs.astro.build/en/guides/internationalization/#getabsolutelocaleurllist) will use the options set in `i18n.domains`.
+		   *  
+		   * See the [Internationalization Guide](https://docs.astro.build/en/guides/internationalization/#domains) for more details, including the limitations of this feature.
```
sarah11918, line 1571:
```suggestion
			 * Both page routes built and URLs returned by the `astro:i18n` helper functions [`getAbsoluteLocaleUrl()`](https://docs.astro.build/en/guides/internationalization/#getabsolutelocaleurl) and [`getAbsoluteLocaleUrlList()`](https://docs.astro.build/en/guides/internationalization/#getabsolutelocaleurllist) will use the options set in `i18n.domains`.
			 *
			 * See the [Internationalization Guide](https://docs.astro.build/en/guides/internationalization/#domains) for more details, including the limitations of this feature.
```
tab tab tab, and a space  :P

packages/astro/src/@types/astro.ts
```diff
@@ -1664,6 +1703,48 @@ export interface AstroUserConfig {
...
 		globalRoutePriority?: boolean;
+
+		/**
+		 * @docs
+		 * @name experimental.i18nDomains
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.3.0
+		 * @description
+	 	 *
```
sarah11918, line 1714:
```suggestion
		 *
```
tab tab and a space

.changeset/tidy-carrots-jump.md
```diff
@@ -0,0 +1,52 @@
...
+Adds experimental support for a new i18n domain routing option (`"domains"`) that allows you to configure different domains for individual locales in entirely server-rendered projects.
+
+To enable this in your project, first configure your `server`-rendered project's i18n routing with your preferences if you have not already done so. Then, set the `experimental.i18nDomains` flag to `true` and add `i18n.domains` to map any of your supported `locales` to custom URLs:
+
+```js
+//astro.config.mjs"
+import { defineConfig } from "astro/config"
+export default defineConfig({
+  site: "https://example.com",
+  output: "server",
```
sarah11918, line 14:
```suggestion
  output: "server", // required, with no prerendered pages
```
Might as well do this here, too!

sarah11918 (approved):
ALRIGHT! I am approving for the sake of completion! :partying_face: 

If there are any syntax errors or nits, I'll find them after this is merged on Wed. and I can check and make any quick fixes well before releasing on Thursday. 🫡

jlarmstrongiv:
I’m getting errors, despite not using `i18nDomains`

```
19:01:17 [ERROR] [config] The feature "i18nDomains" is not supported (used by astro-sst).
19:01:17 [ERROR] The adapter astro-sst doesn't support the feature i18nDomains. Your project won't be built. You should not use it.
```

Issue #9110, opened 2023-11-15
Cannot destructure react component props


izmttk:
### Astro Info

```block
Astro                    v3.5.4
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/react
```

### Describe the Bug

When running `astro dev`, I got an error **"Cannot destructure property 'xxx' of 'undefined' as it is undefined"** when using the `map` method on React component props. It seems like this error has occurred before and has been resolved in #660 , but it's happening again. 
If I missed something, please let me know.

### What's the expected result?

no errors

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hrgerw

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Certainly seems straightfoward. Seems bad, but will need to take a closer look.

matthewp:
This is related to not setting a key. Why this happens though is something I'm still investigating. But it's matter of missing the key, and it's giving you a bad error message.

matthewp:
I think the way our code works is shadowing the warning from coming through. Looking into a fix.

izmttk:
After adding a key prop, it works.
Thanks a lot! :heart:

jhedstrom:
This was quite the rabbit hole. Fixing/adding the `key` property anywhere a `map` method was being called resolved the issue, but the undefined error is very misleading 😬

fbzz:
+1, took me a long time to arrive here. Adding the key, worked.

Important mention: I was using on my index.astro, and it was only on the first render.

mattiaz9:
I just tried to migrate from v2 to v4 and i'm getting this error. Using `key` didn't solve it for me

mattiaz9:
> I just tried to migrate from v2 to v4 and i'm getting this error. Using `key` didn't solve it for me

never mind, it was `vite-plugin-svgr` that was messing with astro somehow

ulysses-ck:
> After adding a key prop, it works. Thanks a lot! ❤️

What do you mean by adding a keyprop @izmttk ?

I already try to use a default value to the prop in the react component, but instead of giving the error is saying: " Astro detected an unhandled rejection. Here's the stack trace:
TypeError: Cannot read properties of undefined (reading 'propNameExample')"

matthewp:
Is anyone still running into this issue? It seems cleared up when I use 4.8.

firxworx:
[...]

PR #9144, opened 2023-11-20
Simply React component check

matthewp:
## Changes

- This removes the vnode check logic from the React renderer.
- In the past this was used to check if the component output React or some other library (like Preact) components.
- In 3.0 we changed how multi-jsx works to use the `include` config instead, so this code is no longer needed.
- Additionally, this code could swallow warnings/errors when it looked like check() failed but actually the component just throw.
- Fixes https://github.com/withastro/astro/issues/9110

## Testing

- New test case added that previously failed.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/dull-ways-laugh.md
+++ b/.changeset/dull-ways-laugh.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/react': patch
+---
+
+Simplify React component check
+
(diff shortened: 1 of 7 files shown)


ematipico (approved):
Less code is the best code!

matthewp:
Yep, thanks for the sanity check. I think what I'm going to do instead is provide information about what renderers are available to the renderer, so that it can check if there is only 1, and if so use the more simple approach to checking.

matthewp:
Closing as this implementation is not right.

PR #9147, opened 2023-11-21
Fix RemarkRehype types

bluwy:
## Changes

Supersedes and closes https://github.com/withastro/astro/pull/6211

The linked PR above explains why the types is incorrect today. This PR merges into `next` which includes the latest versions of the remark/rehype plugins, which I think should work now.

## Testing

Existing test should pass. Also tested the types in an Astro config manually.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/khaki-fans-sell.md
+++ b/.changeset/khaki-fans-sell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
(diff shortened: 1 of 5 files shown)


natemoo-re (approved):
Thanks for finally tackling this! ♥️

ematipico (approved):

PR #9159, opened 2023-11-22
Add CLI shortcuts

bluwy:
## Changes

Adds CLI shortcuts as an easter egg for the dev server. (Updated image)

image

I didn't add for the preview server because adapters can implement they're own preview servers, and I don't want to indirectly raise the bar for adapters to also implement shortcuts because of user expectations

## Testing

Tested manually (as in the screenshot)

## Docs

n/a. We're merging this as an easter egg, so I don't think we should document this in the next minor blog or highlight this in the docs for example.


packages/astro/src/core/logger/vite.ts
```diff
@@ -16,6 +16,10 @@ const vitePageReloadMsg = /page reload (.*)( \(.*\))?/;
 const viteHmrUpdateMsg = /hmr update (.*)/;
 // capture "vite v5.0.0 building SSR bundle for production..." and "vite v5.0.0 building for production..." messages
 const viteBuildMsg = /vite.*building.*for production/;
+// capture "\n  Shortcuts" messages
+const viteShortcutTitleMsg = /.*Shortcuts.*/s;
+// capture "press * + enter to ..." messages
+const viteShortcutHelpMsg = /press.*\+ enter.*to.*/s;
```
matthewp, line 22: Definitely is a shame that we have to do it this way.

matthewp:
What's the reasoning for the `+ enter`? Is this a CLI shortcut convention that I'm not aware of?

packages/astro/src/core/logger/vite.ts
bluwy: Yeah, not the best but worst case is that the logs would appear "off" if these regexes failed to match. We're only hacking it to make it cleaner for us for now.

bluwy:
The `+ enter` convention stems from here: https://github.com/vitejs/vite/pull/14342. The issue is that we can only detect keypresses line-by-line, or char-by-char.

Reading by character requires `setRawMode`, but that disables all the default OS shortcuts like ctrl+c, running in background, and any rare OS handling that will be very hard for us to maintain.

Reading by line is much simpler as we don't hijack the shortcuts, but it requires a `enter` press to create a new line. It works like like `cin >> my_name` in C++ for example.

.changeset/old-cherries-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Adds CLI shortcuts for the dev server
```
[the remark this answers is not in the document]
matthewp: Yeah, I can take this on if Bjorn is going to be away.

packages/astro/src/core/dev/restart.ts
```diff
@@ -163,7 +163,35 @@ export async function createContainerWithAutomaticRestart({
 		// Restart the Astro dev server instead of Vite's when the API is called by plugins.
 		// Ignore the `forceOptimize` parameter for now.
 		restart.container.viteServer.restart = () => handleServerRestart();
+
+		// Set up shortcuts, overriding Vite's default shortcuts so it works for Astro
+		restart.container.viteServer.bindCLIShortcuts({
+			customShortcuts: [
+				{
+					key: 'r',
+					description: 'restart the server',
```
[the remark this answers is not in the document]
bluwy: It's in lowercase following Vite's API. When you request "help" for shortcuts later, it'll print something like "press r + enter to ...", so the ... part needs to be lowercase to flow.

.changeset/old-cherries-beg.md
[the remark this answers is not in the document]
bluwy: I figured it's a small thing to have a specific documentation for it, but I can definitely add one 😄

matthewp (dismissed):
Want @FredKSchott to review

.changeset/old-cherries-beg.md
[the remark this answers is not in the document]
bluwy: Updated at https://github.com/withastro/astro/pull/9159/commits/294338c84987dac3ef66ccca51b8654deb09502c

packages/astro/src/core/logger/vite.ts
FredKSchott: It's definitely a relief that you're not worried about the fragility of this given what you know from the Vite side. But I've also been burned by this exact same bug in the past more than once, and it's really difficult to catch when it happens (different dep versions installed, "wait, was THIS new message meant to be ignored or not?", "is this message even still used anymore?" etc.).

Still not blocking, but what's the best way to start a discussion in Vite for a feature where these logs get explicit codes? `info(msg, code)`, etc.

packages/astro/src/core/messages.ts
```diff
@@ -62,6 +62,29 @@ export function serverStart({
...
 	const version = process.env.PACKAGE_VERSION ?? '0.0.0';
+
+	const messages = [
+		'',
+		`${bgGreen(bold(` astro `))} ${green(`v${version}`)} ${dim(`ready in`)} ${Math.round(
+			startupTime
+		)} ${dim('ms')}`,
+		'',
+		serverUrls({ resolvedUrls, host, base }),
+		`${dim('┃')} Press ${bold('h + enter')} for shortcuts`,
```
FredKSchott, line 73: Unfortunately this line doesn't fit well in this `|` element of the terminal UI. This is where the URLs are listed and the `| Label [...] URL` was designed with an intentional layout. The bold also pulls your attention to it more than it should (besides maybe the first time you see it, the URL is always the thing people want to see/copy/click 99% of the time).

For context, I'm going to be more purist than normal about this terminal UI for the next few months just because of how much work went into cleaning up and standardizing how info is presented. So if this sounds like a nit comment gone wrong, that's where I'm coming from :) 

One suggestion that i'd be okay with: what if we merge this PR, but remove all the changes made to `serverStart()`. Then we still get keyboard shortcuts in time for 4.0, but as more of an easter egg. Then, that buys us time to figure out the right way to show the help text in a future 4.1, 4.2, etc.

Otherwise, I'd push to block this PR until post 4.0 when we have some time to revisit this serverStart message formatting.

FredKSchott (dismissed):

packages/astro/src/core/messages.ts
bluwy: Understandable! I tried fiddling a way to put this text but I couldn't find a nicer way too. I can revert this part then, and possibly the `u + enter` shortcut that displays the URLs.

I also don't mind pushing this post 4.0 since it shouldn't break anything, but I'll leave that to your call 😄

packages/astro/src/core/logger/vite.ts
bluwy: I think frankly it would be tedious to assign specific codes on Vite side so we have a reliable way to catch it. These are plain logs from Vite's point of view, so it can technically log anything. The logs API is also indirectly exposed to Vite/Rollup plugins.

One issue for example with HMR messages is that it could contain the HMR path, and a `code` couldn't easily represent that. We would still need some regex or property to capture it. We also have existing regexes like this currently, so compromises for now 😄 
https://github.com/withastro/astro/blob/9cfa8399752f2db2a0e0001446a17d36d0a8aef8/packages/astro/src/core/errors/dev/vite.ts#L44

For me, I'd like to pass these through without rewriting as much as possible. I have some ideas to make it possible in the future though.

packages/astro/src/core/messages.ts
natemoo-re: I agree that the formatting is a bit weird. I'm going to take a pass at this!

natemoo-re:
Okay I took a pass and cleaned up the formatting a bit. Unfortunately I'm not seeing the shortcuts logged automatically, the messages have to be manually triggered with `h + enter`. The server also seems to hang when I press `r + enter`.

I don't love that we have to parse each message to reformat them. Ideally Vite could return the resolved shortcuts (default + custom - disabled) from `bindCLIShortcuts` so we could format them on a single line.

Terminal with Astro dev server output and shortcuts

bluwy:
> I don't love that we have to parse each message to reformat them. Ideally Vite could return the resolved shortcuts (default + custom - disabled) from `bindCLIShortcuts` so we could format them on a single line.

There was also a discussion about it at https://github.com/withastro/astro/pull/9159#discussion_r1412718009

sarah11918 (approved):
Docs approves the changeset! 

Would suggest a PR adding these three commands here in docs: https://docs.astro.build/en/reference/cli-reference/#astro-dev

matthewp:
@bluwy can you give an update on where this PR is, in regards to the things that prevented it from coming out in 4.0? My fuzzy recollection is that we didn't merge it because:

1. @FredKSchott wanted to tweak the UI a little bit
1. Worry about introducing something new right before the release.

Were those the other two reasons? (2) is no longer relevant, did (1) happen?

bluwy:
Yeah I think no1 is the main one. Nate made some commits to the formatting recently and volunteered to take on it on standup today, so perhaps once that's settled, we can have Fred to take another look again.

Once the formatting is resolved, it should be ready.

FredKSchott:
FYI this is unblocked by me assuming that there is no change to the startup UI! Essentially, merged as an easter-egg for now.

For the startup UI changes, I'd like to pair / work 1:1 with someone to design and then add some sort of help message to startup. @matthewp you can either have someone grab time with me on my calendar, or just assign me a Linear ticket, I'd be happy with either path.

matthewp:
The easter egg idea was ok when it was in order to get it out before 4.0, but now we don't have that time pressure so let's do it the right way and design the UI.

bluwy:
I've went ahead and cleanup the PR based on an [internal discussion in team-platform](https://discord.com/channels/830184174198718474/1199776014834864190/1199776111752650852). We decided to keep this as an easter egg as we want to add more helpful shortcuts, and the current list is a little short. We may also implement showing "tips" in the CLI that may reveal this feature in the future.

@sarah11918 from your [approval comment](https://github.com/withastro/astro/pull/9159#pullrequestreview-1809271050), since we're keeping this as an easter egg, I think we want to hold off documenting it for now.

ematipico (approved):

sarah11918:
@bluwy Yup, approval still stands, and changeset looks great! 🫡

natemoo-re (approved):

PR #9161, opened 2023-11-22
Rename entryPoint to entrypoint

bluwy:
## Changes

There's only one instance where we use `entryPoint` which is for the `injectRoute()` API. This PR renames it to `entrypoint`, and emit a warning if the old name is used

## Testing

Existing tests should pass.

## Docs

https://github.com/withastro/docs/pull/5475

I also added a warning at `packages/astro/src/integrations/index.ts`

### DIFF
--- a/.changeset/rude-hairs-whisper.md
+++ b/.changeset/rude-hairs-whisper.md
@@ -1,0 +1,5 @@
+---
+'astro': minor
+---
+
+When inject routes in integrations with the `injectRoute` API, the passed route's `entryPoint` property has been renamed to `entrypoint` to keep consistency with the other APIs
--- a/.changeset/rude-hairs-whisper.md
+++ b/.changeset/rude-hairs-whisper.md
@@ -2,4 +2,4 @@
 'astro': minor
 ---
 
-When inject routes in integrations with the `injectRoute` API, the passed route's `entryPoint` property has been renamed to `entrypoint` to keep consistency with the other APIs
+When injecting routes in integrations with the `injectRoute` API, the passed route's `entryPoint` property is renamed to `entrypoint` to be consistent with other Astro APIs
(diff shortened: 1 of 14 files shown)


packages/astro/src/integrations/index.ts
```diff
@@ -127,6 +127,13 @@ export async function runHookConfigSetup({
 					updatedConfig = mergeConfig(updatedConfig, newConfig) as AstroConfig;
 				},
 				injectRoute: (injectRoute) => {
+					if (injectRoute.entrypoint == null && 'entryPoint' in injectRoute) {
+						logger.warn(
+							null,
+							`The injected route "${injectRoute.pattern}" by ${integration.name} specifies the entry point with the "entryPoint" property. This property is deprecated, please use "entrypoint" instead.`
+						);
+						injectRoute.entrypoint = injectRoute.entryPoint as string;
```
matthewp, line 135: can't wait until 5.0 to remove this :D

.changeset/rude-hairs-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+When injecting routes in integrations with the `injectRoute` API, the passed route's `entryPoint` property is renamed to `entrypoint` to be consistent with other Astro APIs
```
sarah11918, line 5:
```suggestion
Renames the `entryPoint` property of the `injectRoute` integrations API to `entrypoint` for consistency. A warning will be shown prompting you to update your code when using the old name.
```
My prepositions may not be correct here, but I think frontloading the idea that a property has been renamed is a helpful way to go here!

sarah11918 (approved):
Warning looks good! I left a comment re: changeset.

And noticed that @bluwy is way ahead of me and has already added this to the Breaking changes tracker issue in docs!  Perfect!

.changeset/rude-hairs-whisper.md
bluwy: Sounds good to me 👍

PR #9168, opened 2023-11-22
Remove deprecated features from Astro 3.0

bluwy:
## Changes

Remove some deprecated features from 3.0. Including:

- Adapters are now required to pass `supportedAstroFeatures`.
- The `build.split` and `build.excludeMiddleware` options are removed.
- The `markdown.drafts` option and draft feature is removed.
- Lowercase endpoint names are no longer supported.
- `getHeaders()` exported from markdown files is removed.

NOTE: Not all is removed yet, I'll follow up another PR to remove the remaining ones as they're larger and scoped. (done: https://github.com/withastro/astro/pull/9181)

NOTE 2: some diff changes are only formatting changes. the `next` branch doesn't auto-format but I formatted it locally automatically.

## Testing

Ran all tests locally. CI should also pass

## Docs

This should be noted in the migration guide.


.changeset/slow-hornets-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/rss': major
+---
+
+Removes the `drafts` option as the feature is deprecated and removed in Astro 4.0
```
bluwy, line 5:
```suggestion
Removes the `markdown.drafts` option as the feature is deprecated in Astro 3.0
```

.changeset/slow-hornets-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/rss': major
+---
+
+Removes the `markdown.drafts` option as the feature is deprecated in Astro 3.0
```
bluwy, line 5:
```suggestion
Removes the `drafts` option as the feature is deprecated in Astro 3.0
```

PR #9179, opened 2023-11-23
fix(rendering): prevent error when slots is keyed into

### LINKED ISSUE #9178: Using slot in a page leads to an error `Cannot read properties of null (reading 'default')`
### Astro Info

```block
Astro                    v3.6.0
Package Manager          pnpm
Adapter                  none
                         @astrojs/sitemap
```

_No response_

### Describe the Bug

When a page directly in `src/pages` has a slot...

```astro
---
---
<h1>Title</h1>
<slot>Fallback placeholder</slot>
```
...instead of a using the content within the slot as the result, Astro fails with an unhelpful and noisy error message.

image

This was a strange experience for a new user exploring Astro.

Fallback gets used instead of a failure/a helpful error message.

https://stackblitz.com/edit/github-o1zkqe?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes

- Switches a `null` to a `{}`
- Fixes #9178

## Testing
Added a case to `astro-slots.test.js`

## Docs
Does not affect usage

### DIFF
--- a/.changeset/tall-hotels-argue.md
+++ b/.changeset/tall-hotels-argue.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue where the presence of a slot in a page led to an error.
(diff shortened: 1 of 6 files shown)


natemoo-re (approved):
LGTM!

MoustaphaDev (approved):

matthewp (approved):

PR #9181, opened 2023-11-23
Remove support for simple objects in endpoints

bluwy:
## Changes

Also removes `ResponseWithEncoding` and renames types for middlewares.

A good chunk of the changes involved updating tests to not use the deprecated form.

## Testing

Existing tests should pass.

## Docs

Added to migration tracker


packages/astro/test/fixtures/with-endpoint-routes/src/pages/images/hex.ts
```diff
@@ -1,9 +1,6 @@
...
 
-export async function GET({ params, request }) {
-  const buffer = readFileSync(new URL('../../astro.png', import.meta.url));
-  return {
-    body: buffer.toString('hex'),
-    encoding: 'hex',
-  };
+export async function GET() {
+	const buffer = await readFile(new URL('../../astro.png', import.meta.url));
+	return new Response(buffer.buffer);
```
bluwy, line 5: TIL readFile with promises return a correctly encoded buffer compared to the sync version. I'm not sure why.

packages/astro/src/@types/astro.ts
```diff
@@ -2373,20 +2359,16 @@ export interface AstroIntegration {
...
+export type MiddlewareHandler = (
 	context: APIContext,
-	next: MiddlewareNext<R>
-) => Promise<R> | R | Promise<void> | void;
-
-export type MiddlewareResponseHandler = MiddlewareHandler<Response>;
-export type MiddlewareEndpointHandler = MiddlewareHandler<Response | EndpointOutput>;
-export type MiddlewareNextResponse = MiddlewareNext<Response>;
+	next: MiddlewareNext
+) => Promise<Response> | Promise<void> | void;
```
MoustaphaDev, line 2366: Was removing `Response` from this type union intended?
```suggestion
) => Promise<Response> | Response | Promise<void> | void;
```

bluwy: Good catch. It probably got lost during refactors.

MoustaphaDev (approved):

bluwy:
Thanks for the approval. Since this change is slightly big and probably worth a docs pass, I'll hold off merging for now.

matthewp (approved):
wait, so we didn't need response with encoding this whole time?

bluwy:
Yeah I found out over time that it's mostly to help simplify migration, most usecases can convert to `ArrayBuffer` which Response can work with by itself.

.changeset/clever-beds-notice.md
```diff
@@ -0,0 +1,9 @@
+---
+'astro': major
+---
+
+Removes support for returning simple objects from endpoints (deprecated since Astro 3.0). You should return a `Response` instead.
+
+`ResponseWithEncoding` is also removed. You can refactor the code to return a response with an array buffer instead, which is encoding agnostic.
+
+The types for middlewares are also cleaned up. To type a middleware function, you should now use `MiddlewareHandler` instead of `MiddlewareResponseHandler`. If you used `defineMiddleware()` to type the function, no changes are needed.
```
sarah11918, line 9:
```suggestion
The types for middlewares have also been revised. To type a middleware function, you should now use `MiddlewareHandler` instead of `MiddlewareResponseHandler`. If you used `defineMiddleware()` to type the function, no changes are needed.
```
Just a suggestion to imply "have changed" vs "cleaned up" since that might not be obvious what that means. So anything like "types for middleware have changed" "have been revised" etc.

sarah11918 (approved):
Changeset is very nice! One tiny suggestion, but otherwise very helpful!

PR #9183, opened 2023-11-23
Clean up tests

bluwy:
## Changes

- Remove stray `console.log()`
- Make sure tests has `.test.js` postfix/extension
- Remove `node:test` import
- Rename some test names

## Testing

Existing tests should pass.

## Docs

n/a. internal refactor.


packages/astro/test/units/i18n/astro_i18n.test.js
```diff
@@ -636,6 +636,7 @@ describe('getLocaleAbsoluteUrl', () => {
 				...config.experimental.i18n,
 				trailingSlash: 'never',
 				format: 'directory',
+				site: 'https://example.com',
```
bluwy, line 639: This file was never run as it didn't have the `.test.js` postfix. I added it but it failed here, which I fixed it by passing `site` here.

matthewp (approved):

PR #9184, opened 2023-11-23
Remove vercel deprecated analytics option

bluwy:
## Changes

Remove `analytics` option

## Testing

Existing tests should pass.

## Docs

I don't think this needs to be documented in the breaking tracker as this is specific to vercel? The changesets might be enough.


MoustaphaDev (approved):
LGTM!

Issue #9185, opened 2023-11-23
links with the same pathname as the current page but different search params are never prefetched


SpencerWhitehead7:
### Astro Info

```block
Astro                    v3.6.0
Node                     v20.7.0
System                   macOS (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Links with the same pathname as the current page, but different search params, are never prefetched

### What's the expected result?

Links with the same pathname as the current page are prefetched if their search params are different

I don't know whether the current behavior is intended, and just checking search params could have some edge cases (same params, but in a different order in the url string?) But I feel like the current behavior goes against expectations in at least some cases, like pagination, where it's commonly controlled by search query and you'd think it should prefetch the next/previous page of results.

I found this related issue https://github.com/withastro/astro/issues/4963 fixing a bug where it would prefetch the exact same page and I think that introduced the behavior I'm seeing. After the fix, one of the comments raised the same issue I am.

I believe it could be made to also prefetch pages with different search params by adding that case to the check here https://github.com/withastro/astro/blob/04fdc1c613171409ed1a2bd887326e26cdb8b5ef/packages/astro/src/prefetch/index.ts#L227C11-L227C11

### Link to Minimal Reproducible Example

I couldn't reproduce it on stackblitz because it involves the url, but I can make one on github if the description and link to the source code aren't enough

### Participation

- [X] I am willing to submit a pull request for this issue.

bluwy:
Sounds fair to me. Feel free to send a PR to fix that line!

SpencerWhitehead7:
thank you, https://github.com/withastro/astro/pull/9189

PR #9189, opened 2023-11-24
fix: links with same path but different search params not prefetched

SpencerWhitehead7:
## Changes

Links where the path is the same, but search params are not, are prefetched

## Testing

e2e test

## Docs

As far as I can tell from reading the docs, they do not mention prefetching behavior in this case at all. Since this change aligns it more with expected behavior, I don't think more information would need to be added, but I can write something if necessary.

### Open Issue:

fix https://github.com/withastro/astro/issues/9185

### DIFF
--- a/.changeset/brown-jars-lick.md
+++ b/.changeset/brown-jars-lick.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-fix: links with same path but different search params not prefetched
+Fixes an issue where links with the same pathname as the current page, but different search params, were not prefetched.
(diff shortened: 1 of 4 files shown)


bluwy (approved):
Looks great and nice tests!

.changeset/brown-jars-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix: links with same path but different search params not prefetched
```
bluwy, line 5:
```suggestion
Enables prefetching links with the same current path but different search params
```

PR #9193, opened 2023-11-25
feat(astro): use plaintext lang if lang does not exist

florian-lefebvre:
## Changes

See https://github.com/withastro/roadmap/discussions/781 for context.

Prevents the Code component from crashing if the lang isn't supported by falling back to `plaintext`.

## Testing

None, it may require adding a test

## Docs

n/a

cc @Szymon-Wrzos


packages/astro/components/Code.astro
```diff
@@ -92,7 +93,13 @@ if (typeof lang === 'object') {
...
 
 const highlighter = await getCachedHighlighter({
-	langs: [lang],
+	langs: [
+		typeof lang === 'string'
+			? Object.keys(bundledLanguages).includes(lang)
+				? lang
+				: 'plaintext'
+			: lang,
+	],
```
natemoo-re, line 102: This seems okay, but I think the ideal solution would be to not pass it through the highlighter at all if we don't detect a language.

natemoo-re (approved):
Thanks for the PR! I'd prefer if we could avoid the highlighting entirely if a language isn't passed, but this is a decent alternative if that's not possible.

packages/astro/components/Code.astro
florian-lefebvre: I'm not familiar with shikiji, so it langs optional or de we need to pass an empty array?

florian-lefebvre: Looking at https://github.com/withastro/astro/blob/main/packages/markdown/remark/src/shiki.ts#L28, I suppose both work since it will an empty array by default but a confirmation wouldn't hurt!

bluwy (approved):

packages/astro/components/Code.astro
bluwy: I think a fix like now should be fine. We need shikiji to output the specific code/pre HTML elements, so not passing through the highlighter would be a little tricky.

natemoo-re: I'm cool with this, then! Makes sense.

PR #9199, opened 2023-11-27
Switch `app.render` signature

lilnasy:
## Changes
- `app.render(request, undefined, locals)` -> `app.render(request, { locals })`

## Testing
Switched a couple of tests to use the new API, keeping two as-is that still test the current API.

## Docs
- https://github.com/withastro/docs/pull/5510

### DIFF
--- a/.changeset/sharp-starfishes-compete.md
+++ b/.changeset/sharp-starfishes-compete.md
@@ -2,7 +2,7 @@
 'astro': major
 ---
 
-In the Integration API, the `app.render()` method of the `App` class has been simplified.
+This change only affects maintainers of third-party adapters. In the Integration API, the `app.render()` method of the `App` class has been simplified. 
 
 Instead of two optional arguments, it now takes a single optional argument that is an object with two optional properties: `routeData` and `locals`.
 ```diff
--- a/.changeset/sharp-starfishes-compete.md
+++ b/.changeset/sharp-starfishes-compete.md
@@ -1,0 +1,11 @@
+---
+'astro': major
+---
+
+The adapter API now offers a simpler signature for rendering. The `render()` method on App now accepts an `options` object.
+
+```diff
+- app.render(request, undefined, locals)
++ app.render(request, { locals })
+```
+The current signature is deprecated but will continue to function until next major version.
--- a/.changeset/sharp-starfishes-compete.md
+++ b/.changeset/sharp-starfishes-compete.md
@@ -2,9 +2,18 @@
 'astro': major
 ---
 
-The adapter API now offers a simpler signature for rendering. The `render()` method on App now accepts an `options` object.
+In the Integration API, the `app.render()` method of the `App` class has been simplified.
 
+Instead of two optional arguments, it now takes a single optional argument that is an object with two optional properties: `routeData` and `locals`.
 ```diff
+ app.render(request)
+
+- app.render(request, routeData)
++ app.render(request, { routeData })
+
+- app.render(request, routeData, locals)
++ app.render(request, { routeData, locals })
+
 - app.render(request, undefined, locals)
 + app.render(request, { locals })
 ```
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -64,6 +64,7 @@ export class App {
 	#baseWithoutTrailingSlash: string;
 	#pipeline: SSRRoutePipeline;
 	#adapterLogger: AstroIntegrationLogger;
+	#renderOptionsDeprecationWarningShown = false;
 
 	constructor(manifest: SSRManifest, streaming = true) {
 		this.#manifest = manifest;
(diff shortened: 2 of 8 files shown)


packages/astro/src/core/app/node.ts
```diff
@@ -116,11 +117,27 @@ export class NodeApp extends App {
...
+		}
+		else {
+			routeData = routeDataOrOptions as RouteData;
+		}
+		
 		if (!(req instanceof Request)) {
 			req = createRequestFromNodeRequest(req);
 		}
-		return super.render(req, routeData, locals);
+		return super.render(req, { routeData, locals });
```
natemoo-re, line 140: Do any adapters extend from `App`? They might need to be updated as well, but I guess it's not urgent since this is backwards-compatible.

packages/astro/src/core/app/index.ts
```diff
@@ -145,7 +151,23 @@ export class App {
 		return routeData;
 	}
 
-	async render(request: Request, routeData?: RouteData, locals?: object): Promise<Response> {
+	async render(request: Request, options?: RenderOptions): Promise<Response>
+	async render(request: Request, routeData?: RouteData, locals?: object): Promise<Response>
```
natemoo-re, line 155: Adding a `/** @deprecated See https://github.com/withastro/astro/pull/9199 */` comment above the old signature would be good so that users get a hint in their editor.

packages/astro/src/core/app/node.ts
```diff
@@ -116,11 +117,27 @@ export class NodeApp extends App {
...
+		if (routeDataOrOptions && ('routeData' in routeDataOrOptions || 'locals' in routeDataOrOptions)) {
+			if ('routeData' in routeDataOrOptions) {
+				routeData = routeDataOrOptions.routeData;
+			}
+			if ('locals' in routeDataOrOptions) {
+				locals = routeDataOrOptions.locals;
+			}
+		}
+		else {
+			routeData = routeDataOrOptions as RouteData;
```
natemoo-re, line 134: Should we log a deprecation warning here? Maintainers probably won't catch this change unless we have visible log.

.changeset/sharp-starfishes-compete.md
```diff
@@ -0,0 +1,20 @@
+---
+'astro': major
+---
+
+In the Integration API, the `app.render()` method of the `App` class has been simplified.
+
```
natemoo-re, line 6: I would rephrase this or add a note that this change is really only relevant for adapter/integration maintainers, not a something that users of official Astro adapters need to update.

natemoo-re (approved):
Code looks good! Just a few thoughts on how we roll this out and talk about it, but nothing blocking.

packages/astro/src/core/app/node.ts
lilnasy: Only NodeApp extends from App. Cloudflare uses App as-is.

packages/astro/src/core/app/index.ts
lilnasy: You can deprecate an overload?!

packages/astro/src/core/app/node.ts
lilnasy: image

.changeset/sharp-starfishes-compete.md
lilnasy: Updated in the commit titled "[clarify changeset](https://github.com/withastro/astro/pull/9199/commits/07ab063e657c64931961fa7a30ea60231ba5f23a)".

packages/astro/src/core/app/node.ts
```diff
@@ -116,11 +117,34 @@ export class NodeApp extends App {
...
+			if (routeDataOrOptions || locals) {
+				super.logRenderOptionsDeprecationWarning();
+			}
+		}
+		
 		if (!(req instanceof Request)) {
 			req = createRequestFromNodeRequest(req);
 		}
-		return super.render(req, routeData, locals);
+		return super.render(req, { routeData, locals });
```
bluwy, line 147: I think we can pass this as `super.render(req, routeData, locals);` like before so that the warning handling is all handled within `App`, and we don't have to duplicate the validation in `NodeApp`?

We can also make `logRenderOptionsDeprecationWarning` private with a leading `#` then.

bluwy (commented):
I think we also need major changesets for `@astrojs/node` and `@astrojs/vercel` since they're using a new syntax that won't be compatible with previous Astro versions anymore.

packages/astro/src/core/app/node.ts
lilnasy: Wouldn't that make it log the warning everytime?

bluwy: I'm not sure I understand. Wouldn't the duplicate validation here also log the warning everytime? As long as the consumer of `NodeApp` (in this case `@astrojs/node`) uses the new API form, it shouldn't log the warning.

bluwy: I think my initial suggestion was incorrect. I meant `super.render(req, routeDataOrOptions, locals);`. Perhaps that's where the confusion is 😅 

If we simply pass the arguments back to `super.render`, the newer API would also be passed through.

lilnasy: ok yeah that makes sense

bluwy (approved):
Looks great!

matthewp:
@lilnasy can you fix the conflicts so we can get this in?

bluwy:
@lilnasy feel free to merge this if it's ready. or if you're waiting for a docs review?

lilnasy:
Yeah, it's ready to merge. I'm under the impression that docs will review 4.0 stuff at a later point (not in the next branch).

PR #9200, opened 2023-11-27
feat(i18n): refined locales

ematipico:
## Changes

Relevant RFC change:  https://github.com/withastro/roadmap/pull/734/commits/dc277b5d02b565f23265269587178cbd18e4a55f

This PR adds a new way to configure locales. This new configuration allows users to map certain codes to a "path" that is more user-friendly.

Example:

```js
// astro.config.mjs
import {defineConfig} from "astro/config"
export default defineConfig({
    i18n: {
        defaultLocale: 'en',
        locales: ['en', 'es', 'fr', {
            path: "portugues",
            codes: ["pt", "pt-BR", "pt-AO"]
        }]
    }
})
```

With this configuration, the user will have `/portugues/` in their URLs, and a folder named `portugues/`. However, when leveraging the SSR `Accept-Language` header, Astro will use the `codes` when asking for `Astro.preferredLocale`.

As a result of this new configuration, the


packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -571,6 +589,10 @@ export function createRouteManifest(
 			if (fallback.length > 0) {
 				for (const [fallbackFromLocale, fallbackToLocale] of fallback) {
 					let fallbackToRoutes;
+					// const fallbackFromLocale = getPathByLocale(_fallbackFromLocale, i18n.locales);
+					// if (!fallbackFromLocale) {
+					// 	continue;
+					// }
```
martrapp, line 595: Were these left in on purpose?

martrapp (approved):
Otherwise looks good to me!

packages/astro/src/i18n/middleware.ts
```diff
@@ -1,18 +1,31 @@
...
+import { getPathByLocale, normalizeTheLocale } from './index.js';
 
 const routeDataSymbol = Symbol.for('astro.routeData');
 
-// Checks if the pathname doesn't have any locale, exception for the defaultLocale, which is ignored on purpose
-function checkIsLocaleFree(pathname: string, locales: string[]): boolean {
-	for (const locale of locales) {
-		if (pathname.includes(`/${locale}`)) {
-			return false;
+// Checks if the pathname doesn't have any locale, exception for the defaultLocale, which is ignored on purpose.
```
MoustaphaDev, line 13: Nit: Could we reformulate it like this, since it has the opposite result of the function it replaces?
```suggestion
// Checks if the pathname has any locale, exception for the defaultLocale, which is ignored on purpose.
```

matthewp (approved):

.changeset/fluffy-dolls-sleep.md
```diff
@@ -0,0 +1,24 @@
...
+Developers can now assign a custom path that can span multiple language codes:
+
+```js
+// astro.config.mjs
+export default defineConfig({
+  experimental: {
+      i18n: {
+          defaultLocale: "english",
+          locales: [
+            { path: "ensligh", codes: ["en", "en-US"]}
```
MoustaphaDev, line 16:
```suggestion
            { path: "english", codes: ["en", "en-US"]}
```

.changeset/fluffy-dolls-sleep.md
```diff
@@ -0,0 +1,24 @@
+---
+'astro': minor
+---
+
+Add a new way to configure the `i18n.locales` array.
```
MoustaphaDev, line 5:
```suggestion
Adds a new way to configure the `i18n.locales` array.
```

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -520,6 +535,9 @@ export function createRouteManifest(
 
 		// we loop over the remaining routes and add them to the default locale
 		for (const route of setRoutes) {
+			if (route.type !== 'page') {
+				continue;
+			}
```
MoustaphaDev, line 540: Is this necessary?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -502,7 +503,21 @@ export function createRouteManifest(
...
 		// First loop
 		// We loop over the locales minus the default locale and add only the routes that contain `/<locale>`.
-		for (const locale of i18n.locales.filter((loc) => loc !== i18n.defaultLocale)) {
+		const filteredLocales = i18n.locales
+			.filter((loc) => {
+				if (typeof loc === 'string') {
+					return loc !== i18n.defaultLocale;
+				} else {
+					return loc.path !== i18n.defaultLocale;
+				}
```
MoustaphaDev, line 512: Nit suggestion: Consider removing the  `else` block

MoustaphaDev (commented):
Looks good to me overall, I just have a few minor suggestions, and questions.

packages/astro/src/core/routing/manifest/create.ts
ematipico: The i18n routing should be applied only to `page`. Although this check isn't needed here, I will move the logic in a better place

packages/astro/src/@types/astro.ts
```diff
@@ -1473,15 +1473,15 @@ export interface AstroUserConfig {
...
+			 * @type {Locales}
 			 * @version 3.5.0
 			 * @description
 			 *
 			 * A list of all locales supported by the website (e.g. `['en', 'es', 'pt-br']`). This list should also include the `defaultLocale`. This is a required field.
 			 *
 			 * No particular language format or syntax is enforced, but your folder structure must match exactly the locales in the list.
 			 */
-			locales: string[];
+			locales: (string | { path: string; codes: string[] })[];
```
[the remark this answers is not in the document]
ematipico: I think so, good catch!

.changeset/fluffy-dolls-sleep.md
```diff
@@ -0,0 +1,24 @@
...
+
+```js
+// astro.config.mjs
+export default defineConfig({
+  experimental: {
+      i18n: {
+          defaultLocale: "english",
+          locales: [
+            { path: "english", codes: ["en", "en-US"]}
+          ],
```
sarah11918, line 17:
```suggestion
          locales: [
            "de",
            { path: "english", codes: ["en", "en-US"]},
            "fr",
          ],
```
If this is actually true ( :smile: ), then it could be helpful to show the mix of the two formats together?

If not true, and all `locales` entries must be of the same type, I wouldn't bother to actually explain that here, but I would show a second language using the same format to imply that, and nudge people to using one configuration or the other.

.changeset/fluffy-dolls-sleep.md
```diff
@@ -0,0 +1,24 @@
...
+          routingStrategy: "prefix-always"
+      }
+  }
+})
+```
+
+With this setting, the URL of the default locale will be start with `/english`. When computing `Astro.preferredLocale`, Astro will use the `codes`.
```
sarah11918, line 24:
```suggestion
With the above configuration, the URL prefix of the default locale will be `/english/`. When computing `Astro.preferredLocale`, Astro will use the `codes`.
```

.changeset/fluffy-dolls-sleep.md
```diff
@@ -0,0 +1,24 @@
+---
+'astro': minor
+---
+
+Adds a new way to configure the `i18n.locales` array.
+
+Developers can now assign a custom path that can span multiple language codes:
```
sarah11918, line 7:
```suggestion
Developers can now assign a custom URL path prefix that can span multiple language codes:
```

packages/astro/src/@types/astro.ts
```diff
@@ -1473,15 +1473,15 @@ export interface AstroUserConfig {
 			 * @docs
 			 * @kind h4
 			 * @name experimental.i18n.locales
-			 * @type {string[]}
+			 * @type {Locales}
 			 * @version 3.5.0
 			 * @description
 			 *
 			 * A list of all locales supported by the website (e.g. `['en', 'es', 'pt-br']`). This list should also include the `defaultLocale`. This is a required field.
```
sarah11918, line 1480:
```suggestion
			 * A list of all locales supported by the website, include the `defaultLocale`. This is a required field.
```

packages/astro/src/core/routing/manifest/create.ts
MoustaphaDev: Ah ok, I missed that from the RFC

MoustaphaDev (approved):
LGTM!

packages/astro/src/@types/astro.ts
```diff
@@ -1473,15 +1473,15 @@ export interface AstroUserConfig {
...
 			 *
 			 * No particular language format or syntax is enforced, but your folder structure must match exactly the locales in the list.
```
sarah11918, line 1482:
```suggestion
			 * Languages can be listed either as individual codes (e.g. `['en', 'es', 'pt-br']`) or mapped to a shared `path` of codes (e.g.  `{ path: "english", codes: ["en", "en-US"]}`). These codes will be used to determine the URL structure of your deployed site.
			 * 
			 * No particular language code format or syntax is enforced, but your project folders containing your content files must match exactly the locales in the list. In the case of multiple `codes` pointing to a custom URL path prefix, store your content files in a folder that exactly matches the first code listed. 
```
As usual, not sure whether this is true or necessary, but it's explicit enough to provide clear guidance!  :sweat_smile: 

I would say we're probably at the point we need about this level of detail/advice here, so whatever is actually true, let's try to end up with something kind of like this?

sarah11918 (commented):
It's so exciting to see this all take shape, @ematipico ! Here are some comments that I think will help make the docs a little more clear for readers! :raised_hands:

packages/astro/src/@types/astro.ts
ematipico: > In the case of multiple `codes` pointing to a custom URL path prefix, store your content files in a folder that exactly matches the first code listed.

This should be

> In the case of multiple `codes` pointing to a custom URL path prefix, store your content files in a folder that exactly matches the `path`.

The `codes` come into play only when we want to check the **preferred locale(s)** from the browser when we read the `Accept-Language` header.

sarah11918: Sounds good! I updated the example.

packages/astro/src/@types/astro.ts
```diff
@@ -1473,15 +1473,15 @@ export interface AstroUserConfig {
 			 * @docs
 			 * @kind h4
 			 * @name experimental.i18n.locales
-			 * @type {string[]}
+			 * @type {Locales}
 			 * @version 3.5.0
 			 * @description
 			 *
-			 * A list of all locales supported by the website (e.g. `['en', 'es', 'pt-br']`). This list should also include the `defaultLocale`. This is a required field.
+			 * A list of all locales supported by the website, include the `defaultLocale`. This is a required field.
```
sarah11918, line 1480:
```suggestion
			 * A list of all locales supported by the website, including the `defaultLocale`. This is a required field.
```
Oops, I made a typo.

Issue #6947, opened 2023-04-29
Index pages appear to be invalid according to web standards (i.e. Apache MultiViews and others)


voidmain:
### What version of `astro` are you using?

2.0.15

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

nom

### What operating system are you using?

Mac

### What browser are you using?

All

### Describe the Bug

When the build format is set to `file`, it looks like Astro is not generating `index.html` files that would mimic the source. Additionally, this confuses web servers like Apache which does not really love when there is a file with the same name as a directory.

Generally, `index.html` represent the directory listing for the path segment of the URL. This is consistent with the last 30 years of the web or so. I would expect that Astro would generate the appropriate `index.html` file from the corresponding `src/pages/foo/index.astro` source. However, it instead creates `build/foo.html`.

Here's an example that causes some confusion:

```
src/pages/foo.astro -> build/dist/foo.html
src/pages/foo/index.astro -> build/dist/foo.html
src/pages/foo/index.html -> build/dist/foo.html
src/pages/foo/some-page.astro -> build/dist/foo/some-page.html
```

This is a little surprising since I've defined a file named `index.astro` or `index.html` but both get compiled to `foo.html`. This behavior also clobbers other files such as the `foo.astro` in the parent directory. If you run this in most default web servers, the URL `/foo/` won't resolve to anything even though the page `src/pages/foo/index.astro` does exist.

The Stackblitz I created won't illustrate the issue exactly because it looks like Stackblitz is translating URLs. It does illustrate that the `foo.astro` is clobbered though. However, if you clone that project and build it locally, you'll see the full symptom and the issue with directory slashes

A change of this build behavior is likely a breaking change, so I would recommend adding a new build format such as `fileIndex` or something like that.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ztkedn

### Edits

Sunday April 30, 2023 - I added some additional information and update the Stackblitz to include the `foo.astro` file that gets clobbered.

ematipico:
> The Stackblitz I created won't illustrate the issue exactly because it looks like Stackblitz is translating URLs

It's still valuable, we can run the command `astro build` and inspect the generated files!

voidmain:
I don't think this is just Apache. Most web servers (Caddy, Nginx, S3, CloudFront, CloudFlare, Google, etc) and web standards treat directories differently than files. More specifically, these URLs technically point to different resources according to the URL specification:

[...]

PR #1197, opened 2021-08-23
Add config options to control URL behavior

FredKSchott:
## Changes

- New config: `trailingSlash: 'always' | 'never' | 'ignore';` to resolve #452 
- New config: `pageDirectoryUrl: boolean` to resolve a secondary part of #452 
- 404.astro now outputs to `404.html` instead of `404/index.html` to work with Netlify, Vercel, etc.
- Some small cleanup around configuartion

## Testing

- Tests added.

## Docs

- Docs added.

PR #9209, opened 2023-11-28
Write index routes as index.html with build.format: 'file'

matthewp:
## Changes

- Previously we ignored whether a route was an index route (aka `index.astro` or `index.md`) when writing files with `build.format: 'file'`. This is because the logic only looked at the pathname and didn't know it was an index route.
- This was unexpected given the nature of file-based routing. The fix is to keep track of index routes in routing and use that to write out the correct file.
- Fixes https://github.com/withastro/astro/issues/6947

## Testing

- Test case added

## Docs

- https://github.com/withastro/docs/issues/5390#issuecomment-1829850862

### DIFF
--- a/packages/astro/src/core/build/common.ts
+++ b/packages/astro/src/core/build/common.ts
@@ -1,6 +1,6 @@
 import npath from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
-import type { AstroConfig, RouteType } from '../../@types/astro.js';
+import type { AstroConfig, RouteData } from '../../@types/astro.js';
 import { appendForwardSlash } from '../../core/path.js';
 
 const STATUS_CODE_PAGES = new Set(['/404', '/500']);
@@ -17,9 +17,10 @@ function getOutRoot(astroConfig: AstroConfig): URL {
 export function getOutFolder(
 	astroConfig: AstroConfig,
 	pathname: string,
-	routeType: RouteType
+	routeData: RouteData
 ): URL {
 	const outRoot = getOutRoot(astroConfig);
+	const routeType = routeData.type;
 
 	// This is the root folder to write to.
 	switch (routeType) {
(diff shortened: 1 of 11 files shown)


bluwy:
I think the previous behaviour is expected to me:

- `format: 'file'` forces `foo/index.html` to `foo.html`
- `format: 'directory'` forces `foo.html` to `foo/index.html`

The "forcing" style intentionally doesn't respect how you declare your files. With this PR, it also slightly conflicts with our [recommendation here](https://docs.astro.build/en/reference/configuration-reference/#effect-on-astrourl):

> To prevent inconsistencies with trailing slash behaviour in dev, you can restrict the trailingSlash option to 'always' or 'never' depending on your build format:
> - directory - Set trailingSlash: 'always'
> - file - Set trailingSlash: 'never'

---

Practically speaking though, I prefer that Astro doesn't force a certain build format style and always respect how the files were declared. This PR seems to make `build.format: 'file'` that way, but maybe it makes more sense if it's `build.format: 'preserve'` instead 🤔

ematipico (approved):

packages/astro/src/core/build/common.ts
```diff
@@ -36,8 +37,15 @@ export function getOutFolder(
 					return new URL('.' + appendForwardSlash(pathname), outRoot);
 				}
 				case 'file': {
-					const d = pathname === '' ? pathname : npath.dirname(pathname);
-					return new URL('.' + appendForwardSlash(d), outRoot);
+					let dir;
+					// If the pathname is '' then this is the root index.html
+					// If this is an index route, the folder should be the pathname, not the parent
+					if(pathname === '' || routeData.index) {
```
MoustaphaDev, line 43: Nit: Maybe we could omit the check for an empty pathname since, in this scenario, it's guaranteed to represent an index route, isn't it?

matthewp:
@bluwy I, good point. I'll move this back into draft for discussion then.

matthewp:
@bluwy what would you think about adding this as `format: 'preserve'`, keeping `'file`' as it currently is, and then deprecating `'file'` (remove in the next major)?

bluwy:
That makes sense to me 👍  aside from this part:

> and then deprecating `'file'` (remove in the next major)?

I think it might still be worth keeping it. Currently it seems like there's a symmetry with `trailingSlash` that:

| `build.format` | `trailingSlash` |
|----------------|-----------------|
| file           | never           |
| directory      | always          |
| preserve       | ignore          |

So if we plan on keeping `trailingSlash: 'never'`, it might make sense to also keep `build.format: 'file'`. The confusing thing is that we have two options that are now even similar 😅 The last time I tracked this down lead me to https://github.com/withastro/astro/pull/1197 but I don't understand why there's two options.

matthewp:
We're probably going to do this in a minor release as the discussion above about making this a new format means it will not be a breaking change. So deprioritizing, but keeping on my personal board.

matthewp:
Closing the PR, but will reopen later with new implementation.

PR #9213, opened 2023-11-28
fix(upgrade): Remove semver parse

natemoo-re:
## Changes

- Fixes an issue when running `@astrojs/upgrade beta` in the docs repo
- The `semverParse` logic wasn't always correct because it could return `null` in certain scenarios. But it turns out the parse wasn't even needed in this case, so I removed it.

## Testing

Tested manually, it fixes the bug

## Docs

N/A, bug fix only


martrapp (approved):
Looks good to me

PR #9215, opened 2023-11-28
Add a new `overlay-toggled` event that dev overlay plugins can listen to

Princesseuh:
## Changes

What the title says. This is useful for instance if you want your plugin to automatically hide itself when the overlay is toggled, 

## Testing

Will be done separately!

## Docs

https://github.com/withastro/docs/pull/5518

### DIFF
--- a/.changeset/warm-horses-hunt.md
+++ b/.changeset/warm-horses-hunt.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Add a new `overlay-toggled` event that dev overlay plugins can listen to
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
Works for me!

Princesseuh:
Will wait for https://github.com/withastro/astro/pull/9230 to land first to see if this still makes sense.

Princesseuh:
This doesn't make as much sense with https://github.com/withastro/astro/pull/9230 now, because the overlay gets minimized a lot more easily and is often not a conscious choice.

PR #9219, opened 2023-11-28
fix(#9216): fix style HMR in certain cases

### LINKED ISSUE #9216: CSS changes do not always correctly update using HMR
### Astro Info

```block
Astro                    v3.6.2
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

- Start the development server and open the site in your browser
- Open `/src/layouts/Layout.astro`
- Update the `background-color` property in the styles

We're keeping track of a counter value using Javascript, loading a heavy image, and doing some heavy work in JS when the page loads. These are all here purely to demonstrate more clearly the improper behavior that is occurring around style updates.

Updating the color of the background (or any style change in this file) causes the counter to reset and often the image to reload, demonstrating that the styles aren't being applied purely via HMR injection.

This becomes especially noticeable on larger projects. While this demo focuses on refreshing styles within a single file, I've found Astro to generally be very hit-or-miss about when it will update styles via HMR and when it will refresh the entire page, and I haven't been able to track down anything consistent that determines when this will and won't happen.

Incidentally, I just tried using the Astro 4.0 beta and I see the same issue there.

CSS updates should occur without the page reloading or the current Javascript state for the page being reset.

https://github.com/cmalven/astro-css-hmr

### Participation

natemoo-re:
## Changes

- Closes PLT-1261
- Adjusts our style HMR logic so that style only changes will only treat the effected style modules as HMR candidates
- This fixes an edge case where updates inside of Astro files _referenced_ by the main page file would cause a full reload because `pages/index.astro` would remain in the module graph.
- This should close #9216, which is a report about this specific edge case. As reported, if the `style` inside of `src/layouts/Layout.astro` was updated, the whole index page would reload. With this fix, updating the `style` inside of `src/layouts/Layout.astro` does not reload the index page but swaps the new styles in seamlessly.

## Testing

Tested manually, our existing HMR tests aren't capable of catching this problem

## Docs

Bug fix only

### DIFF
--- a/.changeset/wicked-books-appear.md
+++ b/.changeset/wicked-books-appear.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix edge case where `<style>` updates inside of `.astro` files would ocassionally fail to update without reloading the page.
--- a/packages/astro/src/vite-plugin-astro/hmr.ts
+++ b/packages/astro/src/vite-plugin-astro/hmr.ts
@@ -97,8 +97,8 @@ export async function handleHotUpdate(
 	// If only styles are changed, remove the component file from the update list
 	if (isStyleOnlyChange) {
 		logger.info('astro', msg.hmr({ file, style: true }));
-		// remove base file and hoisted scripts
-		return mods.filter((mod) => mod.id !== ctx.file && !mod.id?.endsWith('.ts'));
+		// Only return the Astro styles that have changed!
+		return mods.filter((mod) => mod.id?.includes('astro&type=style'));
 	}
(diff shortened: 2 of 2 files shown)


natemoo-re:
Would love a @bluwy gut check on this one! The goal is to filter the HMR modules down to just the `style` that actually changed. We previously only discarded a module if it was known to be irrelevant but this swaps the logic so that we discard everything we don't recognize as relevant.

bluwy (approved):
Makes sense to me 👍

PR #9222, opened 2023-11-28
Ensure dev-overlay-window is anchored to the bottom

matthewp:
## Changes

https://github.com/withastro/astro/assets/361671/362231af-9cfb-4a21-b3ca-83fe97dc7bc5

- Uses bottom calculation (7.5%, which is the devbar's bottom + a magic number)
- works in all tested apps

## Testing

Manually

## Docs

N/A

### DIFF
--- a/.changeset/famous-eels-trade.md
+++ b/.changeset/famous-eels-trade.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Ensure the dev-overlay-window is anchored to the bottom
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
This is certainly `#NWTWWHB` but I'm curious if we should be using a more robust positioning approach like [`@floating-ui/dom`](https://floating-ui.com/docs/computePosition) would give us.

matthewp:
I'd rather do it in CSS. This could be improved by putting the overlay height and bottom into variables and calculating that.

ematipico (approved):

PR #9226, opened 2023-11-29
[i18n] During fallback, only remove locale if prefix-always is false

### LINKED ISSUE #9223: i18n Fallback Routing Broken for Prefix-Always
### Astro Info

```block
Astro                    v3.6.3
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When using the experimental i18n functionality and the `prefix-always` option, routes that i would expect to be caught by the fallback routing logic 404 instead.

Given the following src:
```
src/pages/
├─ en/
│  ├─ about/
│  │  ├─ who.html
│  ├─ index.html
├─ it/
│  ├─ index.html
├─ index.astro
```

links to `/it/about/who/` return a 404 page.

links to `/it/about/who/` redirect to `/en/about/who/`

https://stackblitz.com/edit/withastro-astro-issue-9096-cgxsxv?file=package.json

### Participation

outofambit:
## Changes

- Closes #9223
- Adds conditional that checks that `prefix-always` is false before removing locale in fallback routing

## Testing

I made this change locally in my project's `node_modules` and added a unit test.

## Docs

No doc changes needed, as this restores functionality to match docs

### DIFF
--- a/.changeset/happy-beans-protect.md
+++ b/.changeset/happy-beans-protect.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix i18n fallback routing with routing strategy of always-prefix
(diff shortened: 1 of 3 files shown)


ematipico (approved):
That's great, thank you for catching that!

PR #9236, opened 2023-11-29
refactor(i18n): breakdown routing strategy

ematipico:
## Changes

This PR addresses RFC change: https://github.com/withastro/roadmap/pull/734/commits/089253d317e4327ee4e3b93eb303ddc28859952b

Closes PLT-1269

Thanks to zod `.transform`, the only changes are user-facing and logically it stays the same.

The code changes I did are:
- renaming `routingStrategy` to `routing`, which is transformed in a string (we can then decide to change the string themselves, but it's only bikeshedding)
- Documentation and the shape of the user-facing configuration

## Testing

All tests should pass after updating the configuration to use the new shape.

## Docs

### DIFF
--- a/.changeset/young-trains-shout.md
+++ b/.changeset/young-trains-shout.md
@@ -10,7 +10,7 @@ export default defineConfig({
       i18n: {
 -          routingStrategy: "prefix-always",
 +          routing: {
-+              prefixDefaultLocale: false,
++              prefixDefaultLocale: true,
 +          }  
       }
   }
@@ -23,7 +23,7 @@ export default defineConfig({
       i18n: {
 -          routingStrategy: "prefix-other-locales",
 +          routing: {
-+              prefixDefaultLocale: true,
++              prefixDefaultLocale: false,
 +          }  
       }
   }
--- a/.changeset/young-trains-shout.md
+++ b/.changeset/young-trains-shout.md
@@ -2,4 +2,30 @@
 'astro': patch
 ---
 
-Break `routingStrategy`
+The configuration `i18n.routingStrategy` has been replaced with an object called `routing`.
+
+```diff
+export default defineConfig({
+  experimental: {
+      i18n: {
+-          routingStrategy: "prefix-always",
++          routing: {
++              prefixDefaultLocale: false,
++          }  
+      }
+  }
+})
+```
+
+```diff
+export default defineConfig({
+  experimental: {
+      i18n: {
+-          routingStrategy: "prefix-other-locales",
++          routing: {
++              prefixDefaultLocale: true,
++          }  
+      }
+  }
+})
+```
--- a/.changeset/young-trains-shout.md
+++ b/.changeset/young-trains-shout.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Break `routingStrategy`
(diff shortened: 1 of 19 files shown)


.changeset/young-trains-shout.md
```diff
@@ -0,0 +1,31 @@
...
+
+The configuration `i18n.routingStrategy` has been replaced with an object called `routing`.
+
+```diff
+export default defineConfig({
+  experimental: {
+      i18n: {
+-          routingStrategy: "prefix-always",
++          routing: {
++              prefixDefaultLocale: false,
```
delucis: These two examples should be flipped I think, right?

```suggestion
+              prefixDefaultLocale: true,
```

.changeset/young-trains-shout.md
```diff
@@ -0,0 +1,31 @@
...
+})
+```
+
+```diff
+export default defineConfig({
+  experimental: {
+      i18n: {
+-          routingStrategy: "prefix-other-locales",
++          routing: {
++              prefixDefaultLocale: true,
```
delucis:
```suggestion
+              prefixDefaultLocale: false,
```

packages/astro/src/@types/astro.ts
```diff
@@ -1519,23 +1519,44 @@ export interface AstroUserConfig {
...
+				 *  URLs will be of the form `example.com/[locale]/content/` for all non-default languages, but `example.com/content/` for the default locale.
+				 *
+				 * When `true`, all URLs will display a language prefix.
+				 * URLs will be of the form `example.com/[locale]/content/` for every route, including the default language.
+				 * Localized folders are used for every language, including the default.
+				 */
+				prefixDefaultLocale: boolean;
+
+				/**
+				 * @docs
```
delucis, line 1549: As this is not intended to be used just yet, I think we can probably exclude it from docs for now:

```suggestion
```

packages/astro/src/core/config/schema.ts
```diff
@@ -346,11 +348,25 @@ export const AstroConfigSchema = z.object({
...
+								switch (routing.strategy) {
+									case 'pathname': {
+										if (routing.prefixDefaultLocale === true) {
+											strategy = 'prefix-always';
+										} else {
+											strategy = 'prefix-other-locales';
+										}
+									}
+								}
+								return strategy as RoutingStrategies;
```
delucis, line 368: Nit: can’t we add the type when you declare `strategy` rather than asserting it in the return statement?

```suggestion
								let strategy: RoutingStrategies;
								switch (routing.strategy) {
									case 'pathname': {
										if (routing.prefixDefaultLocale === true) {
											strategy = 'prefix-always';
										} else {
											strategy = 'prefix-other-locales';
										}
									}
								}
								return strategy;
```

packages/astro/test/i18n-routing.test.js
```diff
@@ -661,7 +663,9 @@ describe('[SSG] i18n routing', () => {
 						fallback: {
 							it: 'en',
 						},
-						routingStrategy: 'prefix-always',
+						routing: {
+							prefixDefaultLocale: false,
```
delucis, line 667: If this should mirror the previous setting, I guess it should be `true`?
```suggestion
							prefixDefaultLocale: true,
```

delucis (commented):
Thanks for updating this @ematipico!

delucis (approved):
I’m not super familiar with the routing code but at least superficially code looks good to me and docs do too!

matthewp (approved):

bluwy (approved):

PR #9241, opened 2023-11-30
Remove @astrojs/prefetch integration

bluwy:
## Changes

Remove the `@astrojs/prefetch` package from the monorepo.

Later we also need to run:

```
npm deprecate @astrojs/prefetch "@astrojs/prefetch is deprecated in favor of the builtin prefetch option. Please see the migration guide for more information: https://docs.astro.build/en/guides/prefetch/#migrating-from-astrojsprefetch"
```

## Testing

Existing tests should still pass.

## Docs

This will remove the `@astrojs/prefetch` readme and docs from https://docs.astro.build


matthewp:
I think in the past we've deprecated before removing, I guess thinking that you need to be in the package's folder in order to deprecate it. So is that not the case?

bluwy:
I think you can deprecate it even after we delete it here, the `npm deprecate` command requires the package name passed. But we can also deprecate it now on npm before merging this PR if we want to.

ematipico:
The safest choice is to deprecate it now and then merge the PR. Worst case scenario we can checkout locally the commit before this one and deprecate it.

ematipico (approved):

PR #9248, opened 2023-11-30
Adds missing name/value of the submit button to the form data of a view transition

### LINKED ISSUE #9240: <ViewTransitions handleForms /> doesn't work with <form method="get">
### Astro Info

```block
Astro                    v3.6.3
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When using a `<form>` without a `method` attribute or with the `method` attribute set to `"get"` along with the new `<ViewTransitions handleForms />` router API, the router does not update the URL with the search parameters submitted by the form. It just seems to request the document at `/` (or whatever the form's `action` attribute is set to).

The router should retrieve the new page from the get request by appending the form keys and values as search parameters to the end of the URL

https://stackblitz.com/edit/github-mxrq8z-x77acz?file=src%2Fpages%2Findex.astro

### Participation

martrapp:
## Changes

FormData() constructor takes the submitter as second parameter and adds its name/value (if there is a name property) to the formData.
This mirrors behavior w/o `<ViewTransitions/>`.
Closes #9240

## Testing

Adds e2e test

## Docs

bug fix. Changeset only. (Right tense, Sarah! 😄)


matthewp (approved):

.changeset/ninety-weeks-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds missing name/value of the submit button to the form data of a view transition
```
sarah11918, line 5:
```suggestion
Adds properties of the submit button (name, value) to the form data of a view transition
```
I'm tempted not to use "missing" for "feature we just hadn't added yet" (could also be interpreted as "something user neglected to add in their code). We're allowed to say "Adds..." so maybe this instead?

sarah11918 (commented):
"Quick" suggestion for the changeset, @martrapp !  ;)

PR #9250, opened 2023-11-30
Fix missing type import from auto format

bluwy:
## Changes

This is not completed yet! For some reason `organize-imports-cli` is stripping away `import type * as vite from 'vite';`, and I couldn't figure out why.

Likely some dependency updates from the `next` branch (now merged into `main`) had triggered this, but I tried to revert versions for `typescript`, `ts-morph`, and `@ts-morph/common>minimatch`, but it still get stripped when I run `pnpm format:ci`

This PR currently only brings in the missing `import type * as vite from 'vite';`. We can't merge it as the auto-formatter would strip it away again.

## Testing

n/a

## Docs

n/a


lilnasy:
Turn it off?

bluwy:
Yeah it's easier to turn it off for now.

ematipico (approved):

PR #9254, opened 2023-11-30
Use fixed positioning on x-ray when inside of fixed containers

matthewp:
## Changes

- When an island is inside of a fixed container, it should not use `scrollY` to position itself, as it will never change.

## Testing

- Tested via the docs site (see video above)

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/runtime/client/dev-overlay/plugins/utils/highlight.ts
+++ b/packages/astro/src/runtime/client/dev-overlay/plugins/utils/highlight.ts
@@ -15,23 +15,34 @@ export function createHighlight(rect: DOMRect, icon?: Icon) {
 	return highlight;
 }
 
-export function getHighlightZIndex(el: Element) {
+// Figures out the element's z-index and position, based on it's parents.
+export function getElementsPositionInDocument(el: Element) {
 	let highestZIndex = 0;
+	let fixed = false;
 	let current: Element | ParentNode | null = el;
 	while (current instanceof Element) {
-		let zIndex = Number(getComputedStyle(current).zIndex);
+		let style = getComputedStyle(current);
+		let zIndex = Number(style.zIndex);
 		if (!Number.isNaN(zIndex) && zIndex > highestZIndex) {
 			highestZIndex = zIndex;
 		}
+		if(style.position === 'fixed') {
+			fixed = true;
+		}
 		current = current.parentNode;
 	}
-	return highestZIndex + 1;
+	return {
+		zIndex: highestZIndex + 1,
+		fixed,
+	};
 }
 
 export function positionHighlight(highlight: DevOverlayHighlight, rect: DOMRect) {
 	highlight.style.display = 'block';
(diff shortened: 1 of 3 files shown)


matthewp:
!preview xray-position

packages/astro/src/runtime/client/dev-overlay/plugins/utils/highlight.ts
```diff
@@ -15,23 +15,34 @@ export function createHighlight(rect: DOMRect, icon?: Icon) {
...
 
-export function getHighlightZIndex(el: Element) {
+// Figures out the element's z-index and position, based on it's parents.
+export function getElementsPositionInDocument(el: Element) {
 	let highestZIndex = 0;
+	let fixed = false;
 	let current: Element | ParentNode | null = el;
 	while (current instanceof Element) {
-		let zIndex = Number(getComputedStyle(current).zIndex);
+		let style = getComputedStyle(current);
```
natemoo-re, line 24: I know this is not _worse_ than what we were doing before but **oof** I have performance concerns. 

IMO this approach is fine for now but I'd love to leave a comment that this is probably going to be one of the first places to look if we see performance issues.

packages/astro/src/runtime/client/dev-overlay/plugins/xray.ts
```diff
@@ -83,8 +83,16 @@ export default {
 				attachTooltipToHighlight(highlight, tooltip, islandElement);
 
 				// Set the z-index to be 1 higher than the greatest z-index in the stack.
-				const zIndex = getHighlightZIndex(islandElement);
+				// And also set the highlight/tooltip as being fixed position if they are inside
+				// a fixed container. We do this so that we don't mistakenly take scroll position
+				// into account when setting their position.
+				// We are only doing both of these things once, as they are relatively expensive
+				// to calculate, and are unlikely to change. If that turns out to be wrong, reconsider this.
+				const { zIndex, fixed } = getElementsPositionInDocument(islandElement);
```
natemoo-re, line 91: Oh hey! You did note this already. Would still love an inline comment on the function itself, too.

natemoo-re (approved):
LGTM! Few comments but you're clearly already thinking about them.

ematipico (approved):

packages/astro/src/runtime/client/dev-overlay/plugins/utils/highlight.ts
matthewp: It only happens the first time. We probably should be wrapping the outer logic in a rAF in which case I think it's probably fine

matthewp: I'll add a comment

matthewp: To be clear though, I don't think there's any other way to know an element's z-index and position other than walking all the way up the DOM. If there's another way I'd definitely want to do it!

PR #9257, opened 2023-11-30
Chore: fix noisy peerDependency warnings (it's fine)

natemoo-re:
## Changes

- Adds some `pnpm` config to ignore noisy peerDependency warnings. They're actually fine and we don't need these!
- Just a chore to make our day-to-day better, no changeset needed

## Testing

`pnpm install`, it is no longer annoying

## Docs

No thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,35 +44,6 @@
     "pnpm": ">=8.6.12"
   },
   "packageManager": "pnpm@8.6.12",
-  "pnpm": {
-    "packageExtensions": {
-      "svelte2tsx": {
-        "peerDependenciesMeta": {
-          "typescript": {
-            "optional": true
-          }
-        }
-      }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,14 +44,51 @@
     "pnpm": ">=8.6.12"
   },
   "packageManager": "pnpm@8.6.12",
+  "dependencies": {
+    "astro-benchmark": "workspace:*"
+  },
+  "devDependencies": {
```
FredKSchott, line 50: took me a second and I thought you were adding all of these for the first time :D

matthewp (approved):

package.json
natemoo-re: Ah, apologies, just organizing a little bit!

Issue #9151, opened 2023-11-21
Bundled JS not included in build HTML (i18n issue?)


AlexVoigt:
### Astro Info

```block
Astro                    v3.5.5
Node                     v18.16.0
System                   macOS (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

Bundled Javascript (hoisted.hash.js) is not included in the html.

I use the experimental feature i18n for a one pager in 3 languages.

The bundled Javascript from the page and components is not included in the build.

If I deactivate the i18n feature, the script is integrated.

### What's the expected result?

Build should include bundled javascript as expected. CSS does work without any issues.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vnvbhe-9gotbn?file=src%2Fcomponents%2FCard.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for the investigation!

What I also notice in your repro is that switching to `routingStrategy: "prefix-other-locales"` includes scripts but deletes the default locale page.

AlexVoigt:
Updated to 3.6.0.

With `routingStrategy: "prefix-always"`
Only /index.html is created with the redirect, no directories for the individual languages.

With `routingStrategy: "prefix-other-locales"`
Directories for other languages created, `<script type="module" src="/_astro/hoisted.hash.js"></script>` included in each file. But /index.html for default language completely missing.

ematipico:
This should be fixed. Can you try the latest Astro version @AlexVoigt ?

PR #9261, opened 2023-12-01
chore: add test case

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/9151

## Testing

I added new test cases to cover the issue, because the logic that fixes the bug is already in `main`

## Docs

N/A


matthewp (approved):

packages/integrations/sitemap/test/fixtures/static/src/pages/endpoint.json.ts
```diff
@@ -1,8 +1,8 @@
...
-      name: 'Astro',
-      url: 'https://astro.build/',
-    }),
-  };
+	return Response.json(
+		JSON.stringify({
+			name: 'Astro',
+			url: 'https://astro.build/',
+		})
+	);
```
bluwy, line 7:
```suggestion
	return Response.json({
		name: 'Astro',
		url: 'https://astro.build/',
	});
```

You can pass the object directly for `Response.json`

natemoo-re (approved):
LGTM!

PR #9266, opened 2023-12-01
Remove remaining simple objects usage

bluwy:
## Changes

Make sure we don't return simple objects from endpoint anywhere else. Also made sure they're uppercase now.

I like `Response.json`

## Testing

Existing tests should pass.

## Docs

n/a. internal change to tests and examples.

### DIFF
--- a/packages/astro/test/fixtures/content-collection-references/src/pages/welcome-data.json.js
+++ b/packages/astro/test/fixtures/content-collection-references/src/pages/welcome-data.json.js
@@ -4,7 +4,7 @@ export async function GET() {
 	const welcomePost = await getEntry('blog', 'welcome');
 
 	if (!welcomePost?.data) {
-		return Response.json({ error: 'blog/welcome did not return `data`.' })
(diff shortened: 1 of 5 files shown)


packages/astro/test/fixtures/content-collection-references/src/pages/welcome-data.json.js
```diff
@@ -4,9 +4,7 @@ export async function GET() {
 	const welcomePost = await getEntry('blog', 'welcome');
 
 	if (!welcomePost?.data) {
-		return { 
-			body: { error: 'blog/welcome did not return `data`.' },
-		}
+		return Response.json({ error: 'blog/welcome did not return `data`.' })
```
natemoo-re, line 7:
```suggestion
		return Response.json({ error: 'blog/welcome did not return `data`.' }, { status: 404 })
```

It's a fixture so I know it doesn't matter but `Response.json` can return a status as well.

natemoo-re (approved):
LGTM!

PR #9273, opened 2023-12-02
chore: also export `DevToolbarApp` type for API

alexanderniebuhr:
## Changes

- If you write your own Dev Toolbar App and want to type it, you need exported types.
- Before it was `interface DevOverlayPlugin`
- I kept the existing one, but added a copy as `interface DevToolbarApp`
- I would probably like to export the same interface as two names, but I'm not sure how to do it with typescript

## Testing

- manually tested

## Docs

- not needed for the change, but I wonder if we should have an `typed` example in the Dev Toolbar App API reference page

cc @sarah11918


packages/astro/src/@types/astro.ts
```diff
@@ -2514,6 +2514,14 @@ export interface DevOverlayPlugin {
 	beforeTogglingOff?(canvas: ShadowRoot): boolean | Promise<boolean>;
 }
 
+export interface DevToolbarApp {
+	id: string;
+	name: string;
+	icon: Icon;
+	init?(canvas: ShadowRoot, eventTarget: EventTarget): void | Promise<void>;
+	beforeTogglingOff?(canvas: ShadowRoot): boolean | Promise<boolean>;
+}
```
natemoo-re, line 2523: This is how you would do an alias of an existing type in TypeScript!

```suggestion
export type DevToolbarApp = DevOverlayPlugin
```

natemoo-re (commented):
You can alias the existing types with the method I commented here! 

I think the best thing to do would be to rename the existing types to match the  new "Toolbar" name and then add exported aliases for the old names. Does that make sense?

packages/astro/src/@types/astro.ts
alexanderniebuhr: Oh I know this, but I thought `type != interface`

MoustaphaDev (approved):

matthewp (approved):

PR #9274, opened 2023-12-03
Fix i18n routing param

TheOtterlord:
## Changes

Fixes usage of `routing` as a param to match the newer config.

TLDR: `getRelativeLocaleUrlList` was flagged as not returning `/en/` when the routing strategy was `prefix-always`. Looks like that originated from `i18n/index.ts` still using `routingStrategy` from before the config was updated

## Testing

Updated existing tests to match the new behaviour

## Docs

N/A

### DIFF
--- a/.changeset/smart-cats-camp.md
+++ b/.changeset/smart-cats-camp.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fix routing prefixes when `prefixDefaultLocale` is `true`


packages/astro/src/i18n/index.ts
lilnasy, line 1: and we didnt get type errors here because it is a virtual module?

ematipico (approved):

PR #9282, opened 2023-12-04
[Toolbar] Final style polish and code cleanup

FredKSchott:
## Changes

- Small updates to some styles
- Changed props display from custom `key="value"` format to JSON (more familiar format for devs = faster to parse, especially when there are many props)
- Cleanup the event handling logic
- Added click handled logic in Audit and Inspect mode. Did this after some user feedback that these modes broke expectations with how browsers implement their "inspect" mode. I intentionally didn't add this globally because some apps will implement their UI differently so that it would be impossible to accurately identify a "page" click vs. an "overlay" click.

Screen Shot 2023-12-03 at 10 03 35 PM

Screen Shot 2023-12-03 at 10 05 37 PM

## Testing

- Tested manually

## Docs

- N/A

### DIFF
--- a/packages/astro/e2e/dev-overlay.test.js
+++ b/packages/astro/e2e/dev-overlay.test.js
@@ -200,7 +200,7 @@ test.describe('Dev Overlay', () => {
 		await expect(settingsWindow).toHaveCount(1);
 		await expect(settingsWindow).toBeVisible();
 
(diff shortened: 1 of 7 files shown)


Princesseuh (approved):
Just added two small tweaks (a missing style and one thing that didn't get renamed), but LGTM.

PR #9287, opened 2023-12-04
fix(vercel): clear artifacts from redirects

lilnasy:
## Changes
- Removes unnecessary forwards and backwards slashes preceding entries.

## Testing
Pending

## Docs
Pending


matthewp:
Hm, I'm not sure if we support external intentionally, let me check.

matthewp:
Yeah, it was an explicit non-goal of the RFC: https://github.com/withastro/roadmap/blob/main/proposals/0041-redirects.md#non-goals

I wonder how many people are hitting this. I wouldn't mind revisiting this decision, but to do so in a more formal way, as redirects have a bit of complexity to them that I wouldn't want to just enable external in a one-off way.

I'm surprised I didn't already do this, but would it make sense to check if the `to` is an http link in Astro core and throw/warn on that?

lilnasy:
Yeah, an error seems the way to go.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -451,6 +451,19 @@ export function createRouteManifest(
...
+		{
+			let destination: string
+			if (typeof to === "string") {
+				destination = to
+			}
+			else {
+				destination = to.destination
+			}
+			if (destination.startsWith("http")) {
+				return logger.error('redirects', `Redirecting to an external URLs is not supported: ${from} -> ${to}`);
```
lilnasy, line 463: image

lilnasy:
- Breaking a test introduced in #8522

ematipico (approved):
I'm not sure about the changes in `redirects.ts`, but the others look good to me

lilnasy:
The changes in `redirects.ts` are the ones fixing the issue.

natemoo-re (approved):
I'm happy with these changes! Would love @matthewp to confirm as well.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -451,6 +451,19 @@ export function createRouteManifest(
...
 
+		{
+			let destination: string
+			if (typeof to === "string") {
+				destination = to
+			}
+			else {
+				destination = to.destination
+			}
+			if (destination.startsWith("http")) {
```
matthewp, line 462: Is 'httpfoobar' a valid pathname destination? Do we enforce that the pathname starts with `/`? I guess not since this gets here. So should we be checking for `http(s)://`?

lilnasy: Makes sense, we don't validate the paths much.

packages/astro/test/redirects.test.js
```diff
@@ -27,6 +28,15 @@ describe('Astro.redirect', () => {
 			expect(response.headers.get('location')).to.equal('/login');
 		});
 
+		// ref: https://github.com/withastro/astro/pull/9287
+		it.skip('Ignores external redirect', async () => {
```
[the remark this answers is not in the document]
lilnasy: No, because of the vt test which tests for the opposite.

matthewp (approved):

PR #9288, opened 2023-12-04
fix: enforce separate type imports

Princesseuh:
## Changes

Bit of a bundling quirk, the following code:

```ts
import {type Something} from "../something.js";
```
becomes the following:

```ts
import {} from "../something.js";
```

which means side effects run and stuff, which can be a source of problem. This PR adds a ESLint rule to enforce separate type statements. I also adjusted the code to reflect this rule.

## Testing

Tests should pass!

## Docs

N/A


packages/astro/src/runtime/client/dev-overlay/entrypoint.ts
```diff
@@ -1,5 +1,5 @@
 import type { DevOverlayPlugin as DevOverlayPluginDefinition } from '../../../@types/astro.js';
-import { type AstroDevOverlay, type DevOverlayPlugin } from './overlay.js';
```
Princesseuh, line 2: This line was the motivation for this, the empty import to `overlay.js` was causing it to load early into the page, causing performance issue even on just my small website.

matthewp (approved):
lgtm

ematipico (approved):

PR #9291, opened 2023-12-04
remove outdated devoverlay config and defaultstate config

FredKSchott:
## Changes

- `devOverlay` config was introduced in the `4.0.0-beta.X` release branch, meaning it is not covered by semver now that it has been renamed to `devToolbar`. The only people who would have been using this were people experimenting with early versions of the Astro 4.0 beta.
- This PR removes that no-longer-used config for the official 4.0 release.
- This PR also fixes a bug where `devOverlay` was being checked, instead of `devToolbar`
- This PR also updates tests to use `devToolbar` instead of `devOverlay`
- This PR also removes an outdated `defaultState` config, which was removed in #9230 but the config was missed. This is now a no-op and doesn't make conceptual sense since the toolbar now shows/hides based on hover/focus.

## Testing

- Updated tests

## Docs

- Updated docs in JSDoc
- @matthewp want your signoff first before I involve docs!

### DIFF
--- a/packages/astro/e2e/astro-envs.test.js
+++ b/packages/astro/e2e/astro-envs.test.js
@@ -3,7 +3,7 @@ import { testFactory } from './test-utils.js';
 
 const test = testFactory({
 	root: './fixtures/astro-envs/',
-	devOverlay: {
+	devToolbar: {
 		enabled: false,
 	},
 });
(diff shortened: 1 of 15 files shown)


matthewp (approved):

FredKSchott:
Confirmed no impact on docs, since this was never added to docs either during the early 4.0.0-beta.1 days!

PR #9293, opened 2023-12-04
Remove audit role that depends on CJS package

matthewp:
## Changes

- The package `axobject-query` is CJS and Vite doesn't ESM-ify it, can't add to `optimizedDeps.include`.
- Maybe there's something we can do to shim but removing for 4.0 release.

## Testing

- N/A

## Docs

N/A

### DIFF
--- a/.changeset/tender-eagles-learn.md
+++ b/.changeset/tender-eagles-learn.md
@@ -1,0 +1,7 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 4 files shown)


matthewp:
The failing script is because it's tries to compare bundle size but installing the old version with the current deps.

natemoo-re (approved):
LGTM! Thanks for the quick fix

Issue #9171, opened 2023-11-22
[@astro/rss] Error thrown when missing required property is unclear.


cdvillard:
### Astro Info

```block
Astro                    v3.6.0
Node                     v20.9.0
System                   Linux (x64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/preact
                         @astrojs/mdx
                         @astrojs/sitemap
```

### Describe the Bug

Updating my RSS feed using `@astro/rss`, I kept receiving the following error when editing the `items` property.

![image](https://github.com/withastro/astro/assets/6250214/b3343531-6360-4786-80fa-0b4c9dba7764)

This error provided no obvious solution to what the problem was. It wasn't until I added a `link` property that it began working.

### What's the expected result?

I expected the error to be a little more clear about what might be missing. I do see the documentation makes note of this, but in the event of a quick fix, it wasn't made apparent that I was missing a critical parameter. I think there would be a benefit to making it clear what property is missing if possible.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vazm4h

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for the reproduction!

Looks like the zod runtime errors aren't as helpful as rss' typescript errors.
<img width="433" alt="image" src="https://github.com/withastro/astro/assets/69170106/014d8cc3-3235-4be0-ba2c-0ecbf68177de">

The logic for creating an error from zod's validation failure is here:
https://github.com/withastro/astro/blob/cfa0c454b5ac6732c538399ea944022fb031d460/packages/astro-rss/src/index.ts#L104-L118

lilnasy:
## Good first issue!
This is an opportunity for astro users to learn about how astro works by fixing a small but real bug. You can expect to get help setting up the dev environment, and understanding the project structure in this channel. **If you have already been contributing for a while, consider guiding an enthusiastic newcomer.**

### What's the bug?
The main function `@astrojs/rss` throws an unhelpful error when things go wrong. cdvillard has provided a very clear minimal here linked in the issue description.

### What's going wrong here?
The `@astro/rss` library performs validation on the input using zod. If the validation fails, the details are gathered to create and throw an error. However, the error (which should still be thrown) has very little helpful information.

### Where do I start looking?
[...]

PR #9299, opened 2023-12-05
[@astrojs/rss] Quality-of-Life Improvement to `items` property-related error

cdvillard:
## Changes
This commit addresses a quality-of-life concern when setting up a RSS feed using collections. Specifically:

- it provides more context to the error message thrown when the object passed to the `items` property is missing any of the three required keys or if one of those keys is mistyped.
- it adds a link to provide additional context and instruction

## Testing

I tested this change similarly to how I encountered #9171. Using the `examples/blog` project, I simply messed around with the properties of `items` to ensure an error would render, and that my changes would build and reflect appropriately.

Before:
Screenshot of items error before my changes

After:
Screenshot of items error after my changes

## Docs

This shouldn't necessitate any changes to the docs as this way entirely done to help improve the error solely. Usage of the package remains unchanged.


cdvillard:
Just to put it out there, I'm not married to the copy language used, nor the method. I was a little confused by the use of `errorMap` on line 105. I tried to define a new conditional there to catch `invalid_union` types, but despite building fine, it never seemed to catch it. I have no deep experience with Zod, so I'm sure it could be something specific to the `rssOptionsValidator` ZodObject alone.

.changeset/thirty-hairs-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/rss': patch
+---
+
+This commit addresses a quality-of-life concern when setting up a RSS feed using collections. Specifically, it provides more context to the error message thrown when the object passed to the `items` property is missing any of the three required keys or if one of those keys is mistyped.
```
sarah11918, line 5:
```suggestion
Improves the `@astrojs/rss` error message thrown when the object passed to the `items` property is missing any of the three required keys or if one of those keys is mistyped.
```
Appreciate the precision of the original statement! Just formatted according to our normal structure (starting with the verb) so the reader will hit the actual PR contributions right away.

sarah11918 (commented):
A docs review only: the proposed error message is fantastic. Thank you for the care and attention to detail @cdvillard! Massaged the changeset a little, so see what you think about that.

.changeset/thirty-hairs-pump.md
cdvillard: @sarah11918 Happy to take the suggestion! Simplifies it to exactly what's necessary.

ematipico (approved):

PR #9301, opened 2023-12-05
[Toolbar] fix z-index on dev overlay tooltips and highlights

FredKSchott:
## Changes

- Sites that set their own `z-index` had trouble display the xray/audit highlights. 
- As a result, audits weren't working on astro.build due to a `z-index: 20` somewhere on the page and the one other template that I tried.
- This PR also simplifies the more complex dynamic z-index calculation in favor of a more simple approach of Large Numbers Always™️. If there's any concern here or missing context, lmk! 

## Testing

- Visual changes only, tested manually

## Docs

- N/A

### DIFF
--- a/packages/astro/src/runtime/client/dev-overlay/plugins/audit/index.ts
+++ b/packages/astro/src/runtime/client/dev-overlay/plugins/audit/index.ts
@@ -3,6 +3,7 @@ import type { DevOverlayHighlight } from '../../ui-library/highlight.js';
 import {
 	attachTooltipToHighlight,
 	createHighlight,
+	getElementsPositionInDocument,
 	positionHighlight,
 } from '../utils/highlight.js';
 import { createWindowElement } from '../utils/window.js';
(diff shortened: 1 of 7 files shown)


packages/astro/src/runtime/client/dev-overlay/plugins/audit/index.ts
```diff
@@ -198,6 +198,18 @@ export default {
...
 			const highlight = createHighlight(rect, 'warning');
 			const tooltip = buildAuditTooltip(rule, originalElement);
+
+			// Set the highlight/tooltip as being fixed position the highlighted element
+			// is fixed. We do this so that we don't mistakenly take scroll position
+			// into account when setting the tooltip/highlight positioning.
+			// 
+			// We only do this once due to how expensive computed styles are to calculate, 
+			// and are unlikely to change. If that turns out to be wrong, reconsider this.
+			const { isFixed } = getElementsPositionInDocument(originalElement);
```
FredKSchott, line 208: this logic was present in the xray logic, so felt like it made sense to add here as well

PR #9305, opened 2023-12-05
[Toolbar] Remove image decode call

FredKSchott:
## Changes

- This ended up waiting forever (and eventually erroring and causing a silent audit failure) when I tested on astro.build/blog. 
- I believe this was due to lazy images never decoding until visible. If I scrolled quickly down the page, it wouldn't error out. If I stayed at the top of the page, it would.
- I tested this change and couldn't tell any difference. The audit seemed to perform correctly without waiting for images to decode.
- @Princesseuh I'd love to know the context of this code before removing! 

## Testing

- N/A

## Docs

- N/A


Princesseuh:
This is actually needed for remote images, not sure what's the best way to do it.

FredKSchott:
Oh got it, are you saying that we check for the image to be fully loaded because we can't draw the rectangle if height/width aren't explicitly set?

In that case I'd say we skip, and don't audit images that aren't loaded (similar to how we don't audit things that aren't visible, in the check right above this one). I'd love to figure out a larger system/approach for handling page changes over time in general, but because this hangs the entire audit as written I think a quick fix here for 4.0.0 is what we need now, that can be improved on later.

Princesseuh:
> Oh got it, are you saying that we check for the image to be fully loaded because we can't draw the rectangle if height/width aren't explicitly set?
> 
> In that case I'd say we skip, and don't audit images that aren't loaded (similar to how we don't audit things that aren't visible, in the check right above this one). I'd love to figure out a larger system/approach for handling page changes over time in general, but because this hangs the entire audit as written I think a quick fix here for 4.0.0 is what we need now, that can be improved on later.

Sure, no worries. 4.0.1 will be the largest Astro release yet 🫡

PR #9308, opened 2023-12-05
Release markdoc integration as v0.x

bluwy:
## Changes

Revert to 0.7.2 (current latest version) so the next release is 0.8.0. It was bumped as 1.0.0 by changesets because of the peer dep range auto-update. I've updated that to `"astro": "^3.0.0 || ^4.0.0"`

## Testing

Verified that it works with `pnpm changeset version` locally, which bumps the next version as 0.8

## Docs

n/a


matthewp:
lgtm, no need to wait for CI.

PR #9317, opened 2023-12-05
Improves `@astrojs/upgrade` dependency handling

natemoo-re:
## Changes

- Previously, `@astrojs/upgrade` would fail when packages weren't found in the NPM registry.
- This PR improves dependency handling so that packages without semver versions are ignored.
- In effect, this change ensures that dependencies using `link:`, `github:`, and `workspace:` protocols are ignored. It's also implemented in a future-proof way by only allowing valid semver identifiers rather than allow-listing known protocols.
- ✅ Changeset added

## Testing

New test suite added for the `verify` step, unit testing this function's behavior specifically

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/poor-otters-doubt.md
+++ b/.changeset/poor-otters-doubt.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/upgrade': patch
+---
+
+Improves dependency handling by ignoring packages that don't use a semver version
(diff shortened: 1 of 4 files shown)


delucis (approved):
LGTM! Thanks @natemoo-re — love the tests too 🙌

Issue #9320, opened 2023-12-05
Closing dialog with form method causes page fetch (reload) with View Transitions


doseofted:
### Astro Info

```block
Astro                    v4.0.1
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

When View Transitions are active in an Astro 4 project, closing a dialog using a `<form method="dialog">` causes a page to be fetched again instead of closing the dialog (of course Astro 4 was just released an hour ago so I expect some bugs 🙂).

When that Astro project is deployed with Vercel, that refetch fails with HTTP error 405 (and the page content is replaced with the pathname).

### What's the expected result?

Forms with a method of "dialog" should not send a network request or page navigation.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-m3pdqc?file=src%2Fpages%2Findex.astro%3AL9

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
This is a bit tricky. Closing a `form[method=dialog]` emits the "submit" event on the form, which is a reasonable hint that a navigation is expected.

lilnasy:
## Good first issue!
This is an opportunity for astro users to learn about how astro works by fixing a small but real bug. You can expect to get help setting up the dev environment, and understanding the project structure in `#dev` channel of our discord server (https://astro.build/chat). **If you have already been contributing for a while, consider guiding an enthusiastic newcomer.**

### What's the bug?
The view transitions router makes a request even when a navigation is not supposed to happen.

### What's going wrong here?
The browser-side code for view transitions listens for the "submit" event on the form. When this event fires, the router intercepts it and runs a transition instead of a normal page navigation.

In this case where the form is using the new `method="dialog"`, the submit event would not have cause a page navigation, but the router attempts a view transition anyway, making an unnecessary request to the server.

### Where do I start looking?
The part where the view transitions router intercepts the submit event on forms is here: https://github.com/withastro/astro/blob/874f68c67f588f37b5ad4a698d5c32e1667292dd/packages/astro/components/ViewTransitions.astro#L95-L118.

PR #9327, opened 2023-12-05
Prevent client-side navigation for method="dialog"

doseofted:
## Changes

- View Transitions no longer cause navigation due to events emitted from `<form method="dialog">`
- Resolves https://github.com/withastro/astro/issues/9320

## Testing

Added new E2E test confirming new requests are not sent when modal is opened and then closed using a button/form.

## Docs

Documentation does not need to be updated. Behavior should now match that of Astro 3.

### DIFF
--- a/.changeset/rich-keys-rescue.md
+++ b/.changeset/rich-keys-rescue.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-View Transitions no longer respond to events emitted from form with method "dialog"
+Fixes an edge case for `<form method="dialog">` when using View Transitions. Forms with `method="dialog"` no longer require an additional `data-astro-reload` attribute.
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
Thanks so much for the PR!

.changeset/rich-keys-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+View Transitions no longer respond to events emitted from form with method "dialog"
```
natemoo-re, line 5:
```suggestion
Fixes an edge case for `<form method="dialog">` when using View Transitions. Forms with `method="dialog"` no longer require an additional `data-astro-reload` attribute.
```

PR #9331, opened 2023-12-05
Update `vitefu` version

natemoo-re:
## Changes

- `vitefu` added support for Vite 5 in a patch but that isn't always picked up by package managers.
- Bumping `vitefu` to the minimum version that supports Vite 5.
- This will avoid `peerDependency` warnings that have been reported

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/beige-gorillas-camp.md
+++ b/.changeset/beige-gorillas-camp.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


matthewp (approved):

delucis (approved):
Thanks @natemoo-re — **L**ess **G**ruesome **T**erminal **M**essages 🚀

PR #9336, opened 2023-12-06
Fix issue where 404/500 status codes were logged as "[200]"

FredKSchott:
## Changes

- Fix issue where 404 & 500 status codes were logged as "[200]"
- This happened because the log wasn't respecting the `status` override

## Testing

- N/A, fairly internal change that only impacts a log message

## Docs

- N/A

### DIFF
--- a/.changeset/modern-mice-shout.md
+++ b/.changeset/modern-mice-shout.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


FredKSchott:
merging this so that Thuy can base off of it tomorrow morning

ematipico (approved):

PR #9339, opened 2023-12-06
Fix log message for `Disable notifications`

morinokami:
## Changes

- Fixed the log message to correctly display 'enabled' and 'disabled' when toggling 'Disable notifications' in the Toolbar.

## Testing

- Ran examples locally and verified that the log message has been corrected.

## Docs

- N/A

### DIFF
--- a/.changeset/strange-parrots-promise.md
+++ b/.changeset/strange-parrots-promise.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Looks good to me

florian-lefebvre (approved):

martrapp (approved):
somewhat confusing due to the double negation but the fix is great!

PR #9343, opened 2023-12-06
Adds source file properties to HTML elements only if devToolbar is enabled

### LINKED ISSUE #9324: data-astro-source-file added when devToolbar disabled
### Astro Info

```block
Astro                    v4.0.1
```

_No response_

### Describe the Bug

Disable devToolbar globally and run dev mode. `data-astro-source-file` and `data-astro-source-loc` still on every astro element. Makes reading source difficult to find what you are looking for in browser dev tool and inspecting plain html.

Attributes shouldn't be added when devToolbar is disabled.

https://stackblitz.com/edit/github-1o5lsw

### Participation

martrapp:
## Changes

When the compiler was called, only astro.config.devToolbar was checked to activate the source annotations, but not AstroPerferences.
Now the code mimics the check that we use in router.ts to activate devToolbar.

Closes #9324 

Note: The astro.config setting currently takes precedence over preferences. If astro.config.devToolbar === false, preferences cannot turn it back on. I think we should not use preferences and astro.config together or at least give astro.config a lower priority between preferences.global and preferences.default.  See thread in #maintainers from this morning (6:40 GMT)

## Testing

Manually tested

## Docs

Depending on how complicate the final solution might get :)

### DIFF
--- a/.changeset/famous-bobcats-vanish.md
+++ b/.changeset/famous-bobcats-vanish.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds source file properties to HTML elements only if devToolbar is enabled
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
Looks great to me! We can revisit the priority thing separately from this. It can definitely be a little bit confusing.

natemoo-re (approved):
Thanks, this seems great!

PR #9355, opened 2023-12-06
[solidjs integration] Update solidjs dependency

marvin-j97:
## Changes

Update SolidJS dependency of its Astro integration to 1.8.7.
You always have to manually update Solid because the version is so outdated in the integration.

## Testing

Tested on Astro 4 using `pnpm astro create` & `pnpm astro add solid`, then manually updating Solid to 1.8.7.

## Docs

For the user nothing changes.

### DIFF
--- a/packages/integrations/solid/package.json
+++ b/packages/integrations/solid/package.json
@@ -35,15 +35,15 @@
     "dev": "astro-scripts dev \"src/**/*.ts\""
   },
   "dependencies": {
-    "vite-plugin-solid": "^2.7.0"
+    "vite-plugin-solid": "^2.8.0"
   },
   "devDependencies": {
     "astro": "workspace:*",
     "astro-scripts": "workspace:*",
-    "solid-js": "^1.8.7"
+    "solid-js": "^1.8.5"
   },
   "peerDependencies": {
-    "solid-js": "^1.8.7"
+    "solid-js": "^1.8.5"
   },
   "engines": {
     "node": ">=18.14.1"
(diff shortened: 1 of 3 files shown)


packages/integrations/solid/package.json
```diff
@@ -40,10 +40,10 @@
   "devDependencies": {
     "astro": "workspace:*",
     "astro-scripts": "workspace:*",
-    "solid-js": "^1.8.5"
+    "solid-js": "^1.8.7"
   },
   "peerDependencies": {
-    "solid-js": "^1.4.3"
+    "solid-js": "^1.8.7"
```
bluwy, line 46: Changing the peer dep range is a breaking change. I don't think we need to update the dependencies manually ourselves, users should update themselves if they want to use the latest version of solid?

marvin-j97:
Maybe. The problem is not the peer dependency but the fact it always installs 1.4 in a new project. Having ^1.4.3 as peer but initializing with create-astro with a more up-to-date version would be much better IMO.

packages/integrations/solid/package.json
```diff
@@ -35,15 +35,15 @@
...
+    "vite-plugin-solid": "^2.8.0"
   },
   "devDependencies": {
     "astro": "workspace:*",
     "astro-scripts": "workspace:*",
     "solid-js": "^1.8.5"
   },
   "peerDependencies": {
-    "solid-js": "^1.4.3"
+    "solid-js": "^1.7.2"
```
bluwy, line 46: We could merge the `vite-plugin-solid` upgrade for the vite 5 support, but this peer dep line should be reverted.

florian-lefebvre (approved):

ematipico (approved):

Issue #9262, opened 2023-12-01
Style-only changes in .astro files don't use HMR consistently


cmalven:
### Astro Info

```block
Astro                    v3.6.4
Node                     v18.17.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Related to #9216 

Normally, if I make a style-only change in a `.astro` file (such as changing a `background-color` property and save the file, the styles in the page will update without a refresh.

However, if I make a style change in one file, and then make another style change in another file, that second change will cause a refresh, and so on for each consecutive change in a different file.

You can see this in the linked repo by doing the following:

- Change the `background-color` in `/src/layouts/Layout.astro` and save
- Change the `color` in `/src/pages/index.astro` and save
- Page will refresh

**IMPORTANT: This issue is not about full HRM support in Astro - it applies only to style-only changes in a .astro file**

### What's the expected result?

Any number of style-only changes in `.astro` files - in any order - should not cause a full page refresh.

### Link to Minimal Reproducible Example

https://github.com/cmalven/astro-css-hmr

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Thanks for opening a new issue and continuing to provide great reproductions! I will look into this one.

natemoo-re:
I've started work on refactoring our HMR logic to make this behavior much more consistent!

https://github.com/withastro/astro/pull/9356

bluwy:
I can confirm that this is fixed by https://github.com/withastro/astro/pull/9706 now, but I'll continue to work on Nate's improvements so we get better style change detection. For now I'll close this as fixed.

cmalven:
Thanks all for your work on this. I just tested using the reduced test case repo I created for this (https://github.com/cmalven/astro-css-hmr) with an updated version of Astro and the immediate problem was indeed fixed, but now I've discovered on another project that any updates made in imported, external `.css` files will still trigger a full refresh. Not sure if this is related or not - I can create a new issue if necessary. To be clear, I'm not saying this was caused by the Astro updates - it more likely has been the case for awhile and just wasn't _fixed_ by these latest updates.

I've updated the linked repo to demonstrate the issue with external files.

You can see the issue in the

bluwy:
No problem @cmalven! This does seem like an issue still so I'll re-open it. Will take a look when possible

Gazook89:
[...]

PR #9356, opened 2023-12-06
Improves HMR logic with proper parser

natemoo-re:
## Changes

- Resolves https://github.com/withastro/astro/issues/9262
- This PR improves our HMR logic by refactoring our style detection to actually use the AST provided by the compiler.
- Previously, we were doing a best-effort attempt at determining if only a style had changed. This approach had some subtly inconsistent behavior due to the order in which modules were invalidated.
- Since our compiler now has a proper `parse` mode that provides a full AST, our HMR logic can be much more exact.

## Testing

Tested manually, but will try to add an E2E test to replicate the issue reported

## Docs

N/A, bug fixes

### DIFF
--- a/packages/astro/src/core/compile/cache.ts
+++ b/packages/astro/src/core/compile/cache.ts
@@ -5,6 +5,16 @@ type CompilationCache = Map<string, CompileResult>;
 
 const configCache = new WeakMap<AstroConfig, CompilationCache>();
 
+interface CachedFile { 
+	timestamp: number;
+	content: string;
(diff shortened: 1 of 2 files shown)


bluwy:
I think my main concern is with performance. Before, we're doing string manipulations from the compile result (that was also cached and used in `transform()` hook). With this PR, we're doing two `parse()` and AST serialization that aren't cached, which could be costly. Maybe it's worth checking the perf implications of this 🤔

PR #8846, opened 2023-10-16
feat: make Houston wear scary hats for spooky season

ElianCodes:
## Changes

- This PR makes Houston wear spooky hats for the Spooky Season
- Also updated phrases Houston uses to start a new project
- Commented out previous phrases and hats for fast restoration
- We can also totally not do this and close this PR

HoustonSpooky

ScaryHouston

## Testing

Test still passing 👻

## Docs

I don't specifically think we need to update the docs, but this will temporarily override the `--fancy` flag

Will still need to add changeset

PR #9358, opened 2023-12-07
feat: make Houston festive for the holiday season

xiBread:
## Changes

- Added new hats (❄️, 🎄, 🎁) and a few new messages

I figured since it was done for Halloween (#8846), it'd be appropriate for the holidays. 😄 

## Testing

N/A

## Docs

N/A


ElianCodes:
It is! But I'm actually planning something bigger 👀 but we might have to override the cli-kit package for that.

xiBread:
Oh! In that case, do you want me to close this?

ematipico (approved):

PR #9362, opened 2023-12-07
Simplify vue appEntrypoint handling

bluwy:
## Changes

Simplify `appEntrypoint` handling. Followup from #8794 and #9333

This change moves the default export check to runtime in dev. While this means we're unable to use the Astro logger, we can avoid the `this.resolve`, `this.load`, and `ssrLoadModule` calls which can be expensive.

With this PR, in dev we will log a warning if `default` is not exported. In build, Rollup will log a warning instead that `default` is not exported.

## Testing

Existing tests should pass.

## Docs

n/a. internal change.


packages/integrations/vue/src/index.ts
```diff
@@ -39,71 +33,55 @@ function getJsxRenderer(): AstroRenderer {
...
-						}
-						return `export { default as setup } from "${resolved.id}";`;
-					} catch {
-						options.logger.warn(
-							`Unable to resolve appEntrypoint \`${options.appEntrypoint}\`. Does the file exist?`
-						);
-					}
+				if (options?.appEntrypoint) {
+					const appEntrypoint = options.appEntrypoint.startsWith('.')
+						? path.resolve(root, options.appEntrypoint)
```
natemoo-re, line 60: `root` is a `URL`, does `path.resolve` automatically convert it to a file path?

packages/integrations/vue/src/index.ts
```diff
@@ -39,71 +33,55 @@ function getJsxRenderer(): AstroRenderer {
...
+						: options.appEntrypoint;
+
+					return `\
```
natemoo-re, line 63: Accidental escape character?

```suggestion
					return `
```

packages/integrations/vue/src/index.ts
```diff
@@ -39,71 +33,55 @@ function getJsxRenderer(): AstroRenderer {
...
+import * as mod from ${JSON.stringify(appEntrypoint)};
+						
+export const setup = (app) => {
+	if ('default' in mod) {
+		mod.default(app);
+	} else {
+		${
+			!isBuild
+				? `console.warn("[@astrojs/vue] appEntrypoint \`" + ${JSON.stringify(appEntrypoint)} + "\` does not export a default function. Check out https://docs.astro.build/en/guides/integrations-guide/vue/#appentrypoint.");`
```
natemoo-re, line 72: There's something weird happening with the quotes here. Pretty sure we can simplify this to the following and avoid needing to escape the double quotes from `JSON.stringify`

```suggestion
				? `console.warn('[@astrojs/vue] appEntrypoint ${JSON.stringify(appEntrypoint)} does not export a default function. Check out https://docs.astro.build/en/guides/integrations-guide/vue/#appentrypoint.')`
```

natemoo-re (commented):
This seems much simpler, thanks for the PR!

packages/integrations/vue/src/index.ts
bluwy: I'm currently retrieving the `root` from Vite's `configResolved` hook (which Astro passes), so it should already be a string path.

packages/integrations/vue/src/index.ts
bluwy: The escape is mainly so the generated code is a little cleaner 😅 It's used in template literals ending char to signal "no new line", so we can avoid an initial new line.

packages/integrations/vue/src/index.ts
bluwy: Yeah the initial code was being a bit more conservative. I figured `appEntrypoint` could contain `'` so using single quotes may break here, e.g. `/Users/Astro's great/project/src/_app.ts`. Maybe backticks work instead? I initially avoided it because I wanted to preserve the backticks when logging the `appEntrypoint`

packages/integrations/vue/src/index.ts
natemoo-re: TIL!

PR #9363, opened 2023-12-07
feat: support setting rootMargin for `client:visible`

forberg:
## Changes

This support adding optional `rootMargin` to the `IntersectionObserver` options.

This gives the developer the optional choice to hydrate a bit before the astro-island enters the viewport.

`<Component client:visible="200px" />`

This is very useful for devices on slower connections, components with some `useEffect` changing DOM structure, to avoid CLS.

## Testing

Tested by scrolling and seeing the hydration starts `200px` before the island appears inscreen.


3ddyBoi:
Heres a suggestion for documentation of this feature:

### `client:visible`

- **Priority:** Low
- **Useful for:** Low-priority UI elements that are either far down the page ("below the fold") or so resource-intensive to load that you would prefer not to load them at all if the user never saw the element.

Load and hydrate the component JavaScript once the component has entered the user's viewport. This uses an `IntersectionObserver` internally to keep track of visibility.

```astro
<HeavyImageCarousel client:visible />
```

🆕 **New Feature - Optional rootMargin:** This feature allows for earlier component hydration in scenarios like slow internet connections or when components alter the page layout after loading. By setting a rootMargin, hydration is initiated when this specified margin around the component enters the viewport, rather than the component itself.
```astro
<HeavyImageCarousel client:visible="200px" />
```
Implementing a rootMargin, such as "200px", is beneficial for reducing layout shifts(CLS) and ensuring components are interactively ready sooner, enhancing the stability and responsiveness of the page.

olekenneth:
I think the action fails because of it's missing a GH token when running from a fork

lilnasy:
Would it be possible to implement this as a custom client directive?

https://docs.astro.build/en/reference/directives-reference/#custom-client-directives

olekenneth:
Absolutely. 

It's just that this is going to be useful for everyone using this directive. Some/most of the time it's better to have done the loading/hydration before the component comes into view.

.changeset/stupid-peas-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
matthewp, line 2: This would be a `minor`

matthewp (approved):
I like this change! A small quality of life improvement that doesn't impact maintainability, I think we should do it.

.changeset/stupid-peas-juggle.md
olekenneth, line 2:
```suggestion
'astro': minor
```

bluwy (approved):

bluwy:
We also need to update the docs at https://docs.astro.build/en/reference/directives-reference/#clientvisible

3ddyBoi:
> We also need to update the docs at [docs.astro.build/en/reference/directives-reference/#clientvisible](https://docs.astro.build/en/reference/directives-reference/#clientvisible)

I have opened a PR for the docs now 🎈

.changeset/stupid-peas-juggle.md
```diff
@@ -0,0 +1,10 @@
+---
+'astro': minor
+---
+
+Supports setting rootMargin for the `client:visible` directive.
```
sarah11918, line 5:
```suggestion
Extends the `client:visible` directive by adding an optional `rootMargin` property. This allows a component to be hydrated when it is close to the viewport instead of waiting for it to become visible.
```
Just suggesting something like the above, to provide a little more context here, given that this is a new feature. 

If you wanted to keep it short, then I'd go with:

> Adds an optional `rootMargin` property for the `client:visible` directive.

"Supports" isn't a word that indicates what *changed* as well as "Adds" does, especially when it's "support this new thing you probably don't know what it is/does". So in the interest of making these more helpful when skimmed, "Adds" is always a great choice for a new feature!

I would lean towards "Extends (or, Enhances, or Improves, if you prefer!) the `client:visible` directive" personally, to again make the change as clear for the reader as possible in terms of things they already understand: they know what this client directive is, and you've made it even better!

ematipico (approved):

sarah11918 (commented):
Thanks for this lovely addition to the client directive, @forberg ! I think people will really like this!

I've just left some thoughts on the changeset message. See what you think!

sarah11918 (approved):
Docs is happy!

Issue #9154, opened 2023-11-21
Build process automatically deletes all *.mjs files


iMoses:
### Astro Info

```block
Using version `2.8.0` but it seems that the issue still exist on the current version.
Using Node version `18.17.1`.
The problem occurs at the cleanup stage of the build process.
```

### Describe the Bug

After the build process is done `cleanServerOutput` is called to clean astro temp files, assuming that all `*.mjs` files are internal.
```
	// The SSR output is all .mjs files, the client output is not.
	const files = await glob('**/*.mjs', {
		cwd: fileURLToPath(out),
	});
```
Well, for us this isn't the case as we include `*.mjs` files as part of our build process, which are then deleted at the end of it.
The way this happens is that we are using a dynamic path similar to this `[...filePath]` and are defining `getStaticPath` which returns file-paths ending with `.mjs`.

### What's the expected result?

I'd expect that user generated files will not be deleted as part of the build cleanup.
Please improve the glob search to exclude user generated files.

NOTE: this bug doesn't happen as part of the `dev` process, only `build` and `preview`, so you should build the attached example repo and look into the `dist/examples` directory to see there's only a `.js` file in there and no `.mjs` file to be found.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-t4kxrz?file=package.json,src%2Fpages%2Fexamples%2F[...filePath].ts

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #9374, opened 2023-12-08
Preserve generating routes that end with .mjs

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/9154

Instead of globbing `**/*.mjs` to remove the unused SSR outputs, we can use the Rollup output information to remove them instead. This way we only remove the `.mjs` files that are part of Rollup, and not the ones generated by pages/endpoitns.

## Testing

Added a test from the issue

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/shiny-glasses-care.md
+++ b/.changeset/shiny-glasses-care.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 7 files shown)


natemoo-re (approved):
I feel like we should eventually align the output structure for `server` and `static` outputs, but this is a fine solution for now!

PR #9295, opened 2023-12-04
Remove aria-query and related rule

matthewp:
## Changes

- Follow-up to https://github.com/withastro/astro/pull/9293 as this package breaks as well.

## Testing

- Doing a preview release

## Docs

- N/A

PR #9377, opened 2023-12-08
Add reverted aria audit rules for dev toolbar

bluwy:
## Changes

Un-revert https://github.com/withastro/astro/pull/9295 and https://github.com/withastro/astro/pull/9293

The main change is adding `optimizeDeps.include: ['astro > aria-query', 'astro > axobject-query']` to fix it.

## Testing

Tested manually with `pnpm link`, and `pnpm pack` + install locally (to emulate when it's published)

## Docs

It doesn't seem like we document the audit rules today.


Princesseuh (approved):

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -489,3 +563,81 @@ const a11y_non_interactive_element_to_interactive_role_exceptions = {
 	td: ['gridcell'],
 	fieldset: ['radiogroup', 'presentation'],
 };
+
+const combobox_if_list = ['email', 'search', 'tel', 'text', 'url'];
```
ematipico, line 567: Can we change snake case names to camel case?

ematipico (approved):
Looks good. I would update the variable names, and use camel case, since that's what we use in the code base

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
bluwy: I can change them to camelCase. It was snake_case as it's borrowed from Svelte and that's the convention they use.

bluwy: Seems like there's a lot that uses snake_case. I think maybe we should do one PR later to convert them all at once.

sarah11918 (approved):
Docs is happy!  🥳

PR #9378, opened 2023-12-08
Fix static build when out dir is not within the cwd

jacobp100:
## Changes

We had an issue where some styles were not loading in the static build. We managed to pinpoint it to this - and noticed `clientBuild` calls this function on `config.outDir` - but the mutations don't. I'm not 100% sure this is the correct fix, but it certainly fixed it for us

## Testing

Local testing & repo tests

## Docs

Fixes static builds when `config.outDir` is located outside of the astro project

### DIFF
--- a/.changeset/lorem-ipsum-dolor
+++ b/.changeset/lorem-ipsum-dolor
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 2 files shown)


matthewp (approved):

matthewp:
Fixes https://github.com/withastro/astro/issues/9415

PR #9381, opened 2023-12-08
makes the output of 'astro preferences list' more verbose

martrapp:
## Changes
Shows whether preferences are project local or global and **warns in cases where astroConfig overrules preferences**.

It would be great if a native speaker could check the wording.
If we want to pursue this suggestion, we could also integrate the table with the defaults into the main table (stating that the value is neither project nor global but the default value). 
If we don't want to proceed with the modified table, I'd still like to keep the warning message, as I still find it very confusing in the case of the devToolbar that user preferences can override a `true` astroConfig value, but that they can't override a `false` astroConfig value.

## Indicates whether the prefrences are set per project or globally.
image

## Issues a warning if the devToolbar preferences are true but the toolbar is disabled in astroConfig
image
image

## Testing

manually tested

### DIFF
--- a/.changeset/selfish-donuts-approve.md
+++ b/.changeset/selfish-donuts-approve.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-makes the output of 'astro preferences list' more verbose
+Improves the CLI output of `astro preferences list` to include additional relevant information
--- a/.changeset/selfish-donuts-approve.md
+++ b/.changeset/selfish-donuts-approve.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+makes the output of 'astro preferences list' more verbose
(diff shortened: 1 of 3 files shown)


natemoo-re:
I will give this a more thorough review after the weekend, but thanks for tackling this @martrapp! I will likely have some suggestions for streamlining the design that I'd be happy to implement, but the functionality is definitely better than what we currently have!

martrapp:
> I will give this a more thorough review after the weekend ... I will likely have some suggestions ... that I'd be happy to implement

Hi Nate, that's great, thanks so much and I'm looking forward to your improvements. I have kept most of the original structure. Once the streamlined vision is clear, some refactoring might be necessary/helpful ;-)

.changeset/selfish-donuts-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+makes the output of 'astro preferences list' more verbose
```
natemoo-re, line 5:
```suggestion
Improves the CLI output of `astro preferences list` to include additional relevant information
```

natemoo-re (approved):
Sorry this took so long! These changes look great to me!

PR #9383, opened 2023-12-08
[docs] Vercel README - update code samples

sarah11918:
## Changes

Corrects code samples where the `output` property did not match the adapter usage. Also removes reference to using the removed `/edge/` adapter.

(I guess this might need a changeset to update at NPM?)

## Testing

No tests, docs.

## Docs

Just docs.


natemoo-re (approved):
It does need a changeset to propagate to NPM, but LGTM otherwise!

PR #9387, opened 2023-12-08
Ensure `astro add` only installs stable versions

natemoo-re:
## Changes

- We didn't have an issue to track this but it's come up a few times in Discord
- `astro add svelte` would install `5.0.0-next.22`, which is not a stable version
- This PR refactors `astro add` installation logic to actually resolve the most recent stable version of packages before installing. We do this by querying the registry for all versions and filtering out any prereleases.
- The custom `peerDependency` ~~parser~~ regex has also been removed in favor of `semver`! It was already a dependency, so we might as well use it.
- If users do want to install a prerelease, they can do so manually.

## Testing

Tested manually because `astro add` has ~~almost~~ no test coverage. Woooooo!

## Docs

Bug fix only

### DIFF
--- a/.changeset/young-rats-sin.md
+++ b/.changeset/young-rats-sin.md
@@ -1,0 +1,11 @@
+---
+'astro': patch
+---
+
+Fixes an edge case with `astro add` that could install a prerelease instead of a stable release version. 
+
+**Prior to this change**
+`astro add svelte` installs `svelte@5.0.0-next.22`
+
+**After this change**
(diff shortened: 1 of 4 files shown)


lilnasy:
!preview add-stable

lilnasy (approved):
Looks great to me! Thanks for the much needed cleanup.

matthewp (approved):

bluwy (approved):

ematipico (approved):

Issue #9389, opened 2023-12-09
Dev Toolbar Apps failed to import on Windows


66Leo66:
### Astro Info

```block
Astro                    v4.0.3
Node                     v20.10.0
System                   Windows (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             unocss
                         @astrojs/markdoc
                         @astrojs/sitemap
                         @sentry/astro
                         @spotlightjs/astro
```

### If this issue only occurs in one browser, which browser is a problem?

This happens on nodejs/vite side.

### Describe the Bug

After installing Spotlight Astro Integration, dev toolbar failed to load and console throws error.

I believe this is something related to path issues on Windows platform which I'm using. I also tested [buk0vec/astro-lighthouse](https://github.com/buk0vec/astro-lighthouse) and got similar results. That's why I report here rather than to spotlightjs.

Same result when using npm instead of pnpm.

<details>
<summary>Console output</summary>

```
12:02:02 [200] / 236ms
12:02:02 [ERROR] [vite] Internal server error: Failed to resolve import "C:UsersJust6Repositorielog-new
ode_modules.pnpm@spotlightjs+astro@1.0.0_patch_hash=eywty7qcayhhoywlcu2agrdkti_@sentry+astro@7.86.0_astro@4.0.3
ode_modules@spotlightjsastrosrcindex.ts" from "node_modules\.pnpm\vite@5.0.6\node_modules\vite\dist\client\client.mjs". Does the file exist?
  Plugin: vite:import-analysis
  File: C:/Users/Just6/Repositories/blog-new/node_modules/.pnpm/vite@5.0.6/node_modules/vite/dist/client/client.mjs:1475:29
  1473|
  1474|
  1475|  import * as Spotlight from 'C:\Users\Just6\Repositories\blog-new\node_modules\.pnpm\@spotlightjs+astro@1.0.0_patch_hash=eywty7qcayhhoywlcu2agrdkti_@sentry+astro@7.86.0_astro@4.0.3\node_modules\@spotlightjs\astro\src\index.ts';
     |                              ^
  1476|
  1477|  if (enableOverlay) {
      at formatError (file:///C:/Users/Just6/Repositories/blog-new/node_modules/.pnpm/vite@5.0.6/node_modules/vite/dist/node/chunks/dep-YJaePtkC.js:63007:46)
[...]

PR #9400, opened 2023-12-11
Fix import dev toolbar apps on windows

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/9389

Use `JSON.stringify` when injecting arbitrary paths so it works on Windows (backslashes wont cause problems)

## Testing

I'm not sure if this will fix it, might need help testing from someone on Windows, or add a test (marking draft)

## Docs

n/a. bug fix.


bluwy:
Awesome, thank you for testing the fix! I'm away from my laptop until next week, but I'll open this PR for review for now so we could get it out, and we could add tests later.

Also feel free to submit a PR to `@apotlightjs/Astro` to fix that! I didn't notice they did something similar too.

ematipico (approved):

PR #9403, opened 2023-12-11
Fix Preact Forwarded Ref components not working in astro files when React integration is enabled

### LINKED ISSUE #9402: Cannot use Preact component with forwarded ref in an astro file if the project uses both React and Preact components
### Astro Info

```block
Astro                    v4.0.4
Package Manager          npm
Adapter                  none
                         @astrojs/preact
```

_No response_

### Describe the Bug

In a project configured to use both Preact and React components as such:

```mjs
import react from '@astrojs/react';
import preact from '@astrojs/preact';

export default defineConfig({
  integrations: [
    react({ include: ['**/react/**'] }),
    preact({ inlcude: ['**/preact/**'] }),
  ],
});
```

The following Preact component is able to be used in as an island:

```jsx
export default function (props) {
  return <h1 {...props}>Hello, Preact!</h1>;
}
```

However, if you attempt to use `forwardRef` from `preact/compat`:

```jsx
import { forwardRef } from "preact/compat";

export default forwardRef<HTMLHeadingElement>(function (props, ref) {
  return (
    <h1 ref={ref} {...props}>
      Hello, Preact Ref!
    </h1>
  );
});
```

An error occurs:
image

For search:

```
instance.render is not a function
```

```
cjs/react-dom-server-legacy.node.development.js:5667:31
```

```
finishClassComponent@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5667:31
renderClassComponent@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5700:23
renderElement@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:5947:27
renderNodeDestructiveImpl@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6108:24
renderNodeDestructive@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6080:14
retryTask@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6532:26
performWork@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6580:16
startWork/<@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6904:12
scheduleWork@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:78:3
startWork@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:6903:15
renderToNodeStreamImpl@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:7048:12
renderToStaticNodeStream@file:///home/projects/jvmiwpnmp.github/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:7061:10
renderToStaticNodeStreamAsync/<@/home/projects/jvmiwpnmp.github/node_modules/@astrojs/react/server.js:159:46
renderToStaticNodeStreamAsync@/home/projects/jvmiwpnmp.github/node_modules/@astrojs/react/server.js:158:9
```

It seems like when both React and Preact are used as islands in .astro files, ref-forwarded Preact components attempt to be hydrated as React components. At least that's what that error would indicate to me.

If you remove the React component and integration the forwarded-ref Preact component works just fine as an Astro island (just comment out the imported React component and the react integration configuration in the linked example).

Any mixture of React and Preact islands should work, with or without `client:*` directives, with or without `forwardRef`.

https://stackblitz.com/edit/github-9xdqvd

### Participation

knpwrs:
## Changes

Fixes #9402

Preact's `forwardRef` from `preact/compat` can return function components which forward refs, which React does not support (React forward ref components are always objects, never functions, even when `forwardRef` is wrapping a function component).

## Testing

I have not tested this change. What would qualify as a sufficient test as far as the astro maintainers are concerned?

## Docs

No need for docs, this just prevents the React integration from picking up Preact components that it cannot render.

### DIFF
--- a/.changeset/breezy-hairs-yell.md
+++ b/.changeset/breezy-hairs-yell.md
@@ -2,4 +2,4 @@
 '@astrojs/react': patch
 ---
 
-Prevent React integration from picking up forward-ref Preact function components that it cannot render.
+Prevents unsupported `forwardRef` components created by Preact from being rendered by React
--- a/.changeset/breezy-hairs-yell.md
+++ b/.changeset/breezy-hairs-yell.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+Prevent React integration from picking up forward-ref Preact function components that it cannot render.
--- a/packages/integrations/react/server.js
+++ b/packages/integrations/react/server.js
@@ -24,7 +24,7 @@ async function check(Component, props, children) {
 	if (typeof Component !== 'function') return false;
 
 	// Preact forwarded-ref components can be functions, which React does not support
-	if (typeof Component === 'function' && Component['$$typeof'] === Symbol.for('react.forward_ref') return false;
+	if (typeof Component === 'function' && Component['$$typeof'] === Symbol.for('react.forward_ref')) return false;
 
 	if (Component.prototype != null && typeof Component.prototype.render === 'function') {
 		return React.Component.isPrototypeOf(Component) || React.PureComponent.isPrototypeOf(Component);
--- a/packages/integrations/react/server.js
+++ b/packages/integrations/react/server.js
@@ -24,7 +24,7 @@ async function check(Component, props, children) {
 	if (typeof Component !== 'function') return false;
 
 	// Preact forwarded-ref components can be functions, which React does not support
-	if (typeof Component === 'function' && Component['$$typeof'].toString() === 'Symbol(react.forward_ref)') return false;
+	if (typeof Component === 'function' && Component['$$typeof']?.toString() === 'Symbol(react.forward_ref)') return false;
 
 	if (Component.prototype != null && typeof Component.prototype.render === 'function') {
 		return React.Component.isPrototypeOf(Component) || React.PureComponent.isPrototypeOf(Component);
--- a/packages/integrations/react/server.js
+++ b/packages/integrations/react/server.js
@@ -23,6 +23,9 @@ async function check(Component, props, children) {
 	}
 	if (typeof Component !== 'function') return false;
 
+	// Preact forwarded-ref components can be functions, which React does not support
+	if (typeof Component === 'function' && Component['$$typeof'].toString() === 'Symbol(react.forward_ref)') return false;
+
 	if (Component.prototype != null && typeof Component.prototype.render === 'function') {
 		return React.Component.isPrototypeOf(Component) || React.PureComponent.isPrototypeOf(Component);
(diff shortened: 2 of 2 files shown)


packages/integrations/react/server.js
```diff
@@ -23,6 +23,9 @@ async function check(Component, props, children) {
 	}
 	if (typeof Component !== 'function') return false;
 
+	// Preact forwarded-ref components can be functions, which React does not support
+	if (typeof Component === 'function' && Component['$$typeof'].toString() === 'Symbol(react.forward_ref)') return false;
```
natemoo-re, line 27: Two things here:
1. `Component['$$typeof']` might be `undefined` which would cause the `toString()` to throw. That's what's causing the current test failures.
2. Serializing a Symbol to check the stringified value seems like it kinda defeats the purpose of using a `Symbol`? I'm guessing React does not expose the `forward_ref` symbol?

knpwrs: Preact exports their symbol here: https://github.com/preactjs/preact/blob/aa53ecb182eb1f5cfff397778150e01b66473f5e/compat/src/forwardRef.js#L13-L17

React, as best I can tell, only exports the symbol from an internal/private shared package here: https://github.com/facebook/react/blob/0cdfef19b96cc6202d48e0812b5069c286d12b04/packages/shared/ReactSymbols.js#L25

natemoo-re (approved):
Perfect, thanks for addressing my feedback!

.changeset/breezy-hairs-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/react': patch
+---
+
+Prevent React integration from picking up forward-ref Preact function components that it cannot render.
```
natemoo-re, line 5:
```suggestion
Prevents unsupported `forwardRef` components created by Preact from being rendered by React
```

ematipico (approved):

Issue #9371, opened 2023-12-07
"part" is missing from HTMLAttributes


ricogallo:
### Astro Info

```block
Astro                    v4.0.3
Node                     v20.6.1
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

Any element with a [part](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/part) global attribute throws a TS error

### What's the expected result?

part should be ok

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-53gxxa-jb3vfg?file=src%2Fcomponents%2FMyComponent.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #9404, opened 2023-12-11
Add missing global attributes to JSX definition

Princesseuh:
## Changes

Some newer attributes were missing from our definition, this fixes that.

Fix https://github.com/withastro/astro/issues/9371

## Testing

Tested manually!

## Docs

N/A

### DIFF
--- a/packages/astro/astro-jsx.d.ts
+++ b/packages/astro/astro-jsx.d.ts
@@ -517,7 +517,8 @@ declare namespace astroHTML.JSX {
 			| 'search'
 			| 'send'
 			| undefined
-			| null;
+		| null;
+		exportparts?: string | undefined | null;
 		hidden?: boolean | string | undefined | null;
 		id?: string | undefined | null;
 		inert?: boolean | string | undefined | null;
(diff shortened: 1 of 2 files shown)


bluwy (approved):

packages/astro/astro-jsx.d.ts
```diff
@@ -533,18 +534,23 @@ declare namespace astroHTML.JSX {
...
 		lang?: string | undefined | null;
+		part?: string | undefined | null;
+		popover?: boolean | string | undefined | null;
 		slot?: string | undefined | null;
 		spellcheck?: 'true' | 'false' | boolean | undefined | null;
 		style?: string | StyleObject | undefined | null;
 		tabindex?: number | string | undefined | null;
 		title?: string | undefined | null;
-		translate?: 'yes' | 'no' | undefined | null;
+		translate?: 'yes' | 'no' | string | undefined | null;
```
Princesseuh, line 553: Unrelated to the additions, but `translate` can also be an empty string (equivalent in behaviour to `yes`, weird)

natemoo-re: Would `string` be the correct type here, in that case? Seems like it should only accept `''`?

natemoo-re (approved):
Looks good! One question on `translate`

packages/astro/astro-jsx.d.ts
Princesseuh: Ah, this is a good catch! I actually purposely used `string` because you used to not be able to properly type check empty strings in this context, but it seems like the issue was fixed in a recent version of TypeScript. I'll try it out

PR #9407, opened 2023-12-11
Allow file URLs to be used as import specifiers

matthewp:
## Changes

- Wanted to support this pattern in integrations: `addDevToolbarApp(new URL('./plugin.ts', import.meta.url))`
- This doesn't currently work because we directly pass this value into the virtual module.
- Vite does not support file URLs.
- A Vite plugin is the answer! Maybe we can talk them into adding this to core.

## Testing

- Test added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/smart-colts-think.md
+++ b/.changeset/smart-colts-think.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Allows file URLs as import specifiers
--- a/packages/astro/src/vite-plugin-fileurl/index.ts
+++ b/packages/astro/src/vite-plugin-fileurl/index.ts
@@ -2,7 +2,7 @@ import type { Plugin as VitePlugin } from 'vite';
 
 export default function vitePluginFileURL({}): VitePlugin {
 	return {
-		name: 'astro:vite-plugin-env',
+		name: 'astro:vite-plugin-file-url',
 		resolveId(source, importer) {
 			if(source.startsWith('file://')) {
 				const rest = source.slice(7);
--- a/packages/astro/src/vite-plugin-fileurl/index.ts
+++ b/packages/astro/src/vite-plugin-fileurl/index.ts
@@ -1,0 +1,13 @@
+import type { Plugin as VitePlugin } from 'vite';
+
+export default function vitePluginFileURL({}): VitePlugin {
+	return {
+		name: 'astro:vite-plugin-env',
+		resolveId(source, importer) {
+			if(source.startsWith('file://')) {
+				const rest = source.slice(7);
+				return this.resolve(rest, importer);
+			}
+		}
+	}
+}
--- a/packages/astro/test/fixtures/astro-basic/src/pages/fileurl.astro
+++ b/packages/astro/test/fixtures/astro-basic/src/pages/fileurl.astro
@@ -1,0 +1,10 @@
+---
+import {capitalize} from 'file://../strings.js';
+---
+
+<html>
+	<head><title>Testing</title></head>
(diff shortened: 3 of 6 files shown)


packages/astro/src/vite-plugin-fileurl/index.ts
```diff
@@ -0,0 +1,13 @@
+import type { Plugin as VitePlugin } from 'vite';
+
+export default function vitePluginFileURL({}): VitePlugin {
+	return {
+		name: 'astro:vite-plugin-env',
```
natemoo-re, line 5:
```suggestion
		name: 'astro:vite-plugin-file-url',
```

packages/astro/test/fixtures/astro-basic/src/pages/fileurl.astro
```diff
@@ -0,0 +1,10 @@
+---
+import {capitalize} from 'file://../strings.js';
```
natemoo-re, line 2: Is this supposed to be a thing you can do? I've always assumed `file:` required an absolute path

natemoo-re (commented):
I'm not sure how I feel about this... seems like integration hooks should handle conversions from whatever formats we want to support for entrypoints into a valid ID for Vite.

packages/astro/test/fixtures/astro-basic/src/pages/fileurl.astro
matthewp: This is how npm suggests doing it: https://docs.npmjs.com/cli/v10/configuring-npm/package-json although there they are omitting the `//` part, either works, try `new URL('file://../strings.js')` in your console.

Although in practice this feature is for full URLs passed via `new URL('./rel', import.meta.url)`, I think this added bonus is kind of cool, but it was really only done this way because it's easier to test.

natemoo-re: Hmm, TIL! Thanks for clarifying

matthewp:
@natemoo-re I thought about doing it that way, but this is a lower-level solution and enables any vite plugin (in Astro) to use file URLs which is a super convenient way to support this sort of pattern (having some config value that gets piped into a virtual module). So I felt that was better.

We already allow full file paths as module specifiers, why not file URLs? It's the same thing, but more universal.

.changeset/smart-colts-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Allows file URLs as import specifiers
```
lilnasy, line 5: everywhere?

matthewp: Yes.

matthewp:
@natemoo-re can you explain more? Which integration APIs are inconsistent? I can maybe fix it up here. I just did a quick scan and it looks like all of them save what is passed in directly. Maybe they mutate it further down the stack?

natemoo-re:
> @natemoo-re can you explain more? Which integration APIs are inconsistent? I can maybe fix it up here. I just did a quick scan and it looks like all of them save what is passed in directly. Maybe they mutate it further down the stack?

IMO that's the inconsistency? Some of them throw helpful errors if they can't resolve the file, some of them don't. Some of them support `URL` because they convert it to a path before storing it, some of them don't. Some support relative specifiers, some don't. My thought was to have a util that would normalize them or some way to ensure that they're all stored as a valid Vite ID.

matthewp:
@natemoo-re Do you know which ones behave differently? From looking at the code I don't see a difference.

PR #9409, opened 2023-12-11
refactor(assets): Move things around to avoid importing image service unnecessarily

Princesseuh:
## Changes

In the way of trying to solve https://github.com/withastro/adapters/pull/58, I realized there's some instances where the image service could get imported by accident because some utils were next to the code that loads it. This refactor move away everything unrelated to that elsewhere, so that people and adapters can import the utilities without anything else coming in by accident. 

This doesn't unblock the linked PR just yet, some other bundling nonsense is in the way, however it's a step forward!

## Testing

Tests should pass, this is just a refactor!

## Docs

N/A


ematipico (approved):
Looks good to me

PR #9411, opened 2023-12-11
Prevent malformed  redirect URLs from crashing the build

matthewp:
## Changes

- We track page data during the build, but redirects do not need this. If there is a malformed URL it would throw expecting a valid ViteID. This avoids that.

## Testing

- Test case added to the redirects test.

## Docs

N/A, bug fix


natemoo-re (approved):

matthewp:
This fix is not correct, so closing.

PR #9414, opened 2023-12-12
fix: mention which feature is experimental/deprecated

Skn0tt:
## Changes

Adds the feature name to logs about feature deprecation / experimental status, so users know what feature the log is about.

## Testing

Didn't add any tests, because I couldn't find any test for this part of the code. Let me know where to add a test, and i'm happy to write one!

## Docs

No docs update needed, because this doesn't change any behaviour that applications depend on.


ematipico (approved):
Thank you for the addition!

Issue #8862, opened 2023-10-18
Partytown breaks VT fallback


florian-lefebvre:
### Astro Info

```block
Astro                    v3.3.2
Node                     v18.14.1
System                   Windows (x64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         @astrojs/partytown
```

### Describe the Bug

When creating a partytown script and navigating to a new page in an unsupported browser (like firefox), VT breaks: url is not updated, event listeners are not triggered etc but page swaps. See video.

![image](https://github.com/withastro/astro/assets/69633530/4bc66997-3318-4ecb-b285-ccb64bfcc4a1)

https://github.com/withastro/astro/assets/69633530/1fdfd8c6-fc04-447b-bb69-755d8bce3b31

### What's the expected result?

Shouldn't break

### Link to Minimal Reproducible Example

https://github.com/florian-lefebvre/astro-vt-fallback-bug

### Participation

- [ ] I am willing to submit a pull request for this issue.

martrapp:
Help wanted! Obviously the minimal example does not work for @florian-lefebvre, but I was not able to reproduce the error with Firefox. Who can help to identify the source of this behavior?

lilnasy:
I can reproduce the issue. I could take a closer look later this week.


https://github.com/withastro/astro/assets/69170106/fb16a206-39e0-4a0b-9e9f-5af91f76f45a

lilnasy:
Affects firefox only, not Safari.

alan-barzilay:
I also got the same issue, firefox + partytown + VT

lilnasy:
This one might be hard to debug. The Firefox-specific error does not provide a stacktrace. I'm pretty sure it has to do with Partytown's usage of low-level APIs (Atomics, SharedArrayBuffer) that exposes a memory-management bug in Firefox, but there is no straightforward way to confirm.

lilnasy:
Workaround: turn off minification.
```diff
import { defineConfig } from 'astro/config';
import mdx from '@astrojs/mdx';
import sitemap from '@astrojs/sitemap';

import partytown from "@astrojs/partytown";

// https://astro.build/config
export default defineConfig({
  site: 'https://example.com',
  integrations: [mdx(), sitemap(), partytown()],
+ vite: {
+   build: {
+     minify: false
+    }
+  }
});
```

florian-lefebvre:
[...]

PR #9419, opened 2023-12-12
Prevent Partytown from hijacking history API

matthewp:
## Changes

- Binds the `pushState` and `replaceState` methods which break Partytown with Firefox.
- Fixes https://github.com/withastro/astro/issues/8862

## Testing

- Manually

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -11,8 +11,9 @@ type Events = 'astro:page-load' | 'astro:after-swap';
 
 // Create bound versions of pushState/replaceState so that Partytown doesn't hijack them,
 // which breaks Firefox.
-const pushState = history.pushState.bind(history);
-const replaceState = history.replaceState.bind(history);
+const inBrowser = import.meta.env.SSR === false;
+const pushState = (inBrowser && history.pushState.bind(history)) as typeof history.pushState;
+const replaceState = (inBrowser && history.replaceState.bind(history)) as typeof history.replaceState;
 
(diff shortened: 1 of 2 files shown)


matthewp:
!preview vt-partytown

packages/astro/src/transitions/router.ts
```diff
@@ -9,12 +9,17 @@ type State = {
 };
 type Events = 'astro:page-load' | 'astro:after-swap';
 
+// Create bound versions of pushState/replaceState so that Partytown doesn't hijack them,
+// which breaks Firefox.
+const pushState = history.pushState.bind(history);
+const replaceState = history.replaceState.bind(history);
+
```
martrapp, line 16: Looks good to me. I have no idea, how party town scripts are injected. But these assignments should be early enough as `router.ts` is loaded through `ViewTransitions.astro` from the `<head>`.

martrapp (approved):
LGTM

packages/astro/src/transitions/router.ts
martrapp: Should be guarded with `if (inBrowser)...`

matthewp: Does the router ever run outside of a browser?

martrapp: not run but loaded in SSR, see #8771

martrapp: .. and in that failing test :)

PR #9423, opened 2023-12-13
Error when getImage() is passed an undefined src

matthewp:
## Changes

- `getImage({ src: undefined })` should throw a good error. Now it does.

## Testing

- Existing test case was skipped, now its not.

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/assets/internal.ts
+++ b/packages/astro/src/assets/internal.ts
@@ -79,6 +79,12 @@ export async function getImage(
 			message: AstroErrorData.ExpectedImageOptions.message(JSON.stringify(options)),
 		});
 	}
+	if(typeof options.src === 'undefined') {
+		throw new AstroError({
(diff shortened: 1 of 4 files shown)


packages/astro/src/assets/internal.ts
```diff
@@ -79,6 +79,12 @@ export async function getImage(
 			message: AstroErrorData.ExpectedImageOptions.message(JSON.stringify(options)),
 		});
 	}
+	if(typeof options.src === 'undefined') {
+		throw new AstroError({
+			...AstroErrorData.ExpectedImage,
+			message: AstroErrorData.ExpectedImage.message(options.src, 'undefined', JSON.stringify(options))
```
Princesseuh, line 85: This is a bit of a duplicate check from the one in `baseService`, I do think it's okay though because other services should probably have this check as well.

Princesseuh (approved):

PR #9424, opened 2023-12-13
Handle unhandledrejections in the dev server

matthewp:
## Changes

- Unhandled rejections, caused by having a promise that rejects but has no `.catch()`, will crash the dev server. We can prevent it using `process.on('unhandledRejection')`.

Screen Shot 2023-12-13 at 9 35 02 AM

---

Screen Shot 2023-12-13 at 5 59 39 PM

## Testing

- Manually in demo project. See above screenshot. Server does not crash, good error message is displayed.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/curvy-lobsters-crash.md
+++ b/.changeset/curvy-lobsters-crash.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent dev server from crashing on unhandled rejections
--- a/.changeset/curvy-lobsters-crash.md
+++ b/.changeset/curvy-lobsters-crash.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Prevent dev server from crashing on unhandled rejections
+Prevents dev server from crashing on unhandled rejections, and adds a helpful error message
--- a/packages/astro/src/vite-plugin-astro-server/error.ts
+++ b/packages/astro/src/vite-plugin-astro-server/error.ts
@@ -7,7 +7,7 @@ import { createSafeError } from '../core/errors/index.js';
 import { formatErrorMessage } from '../core/messages.js';
 import { eventError, telemetry } from '../events/index.js';
 
-export function recordServerError(loader: ModuleLoader, config: AstroConfig, pipeline: DevPipeline, _err: unknown): Error {
+export function recordServerError(loader: ModuleLoader, config: AstroConfig, pipeline: DevPipeline, _err: unknown) {
 	const err = createSafeError(_err);
 
 	// This could be a runtime error from Vite's SSR module, so try to fix it here
@@ -26,5 +26,8 @@ export function recordServerError(loader: ModuleLoader, config: AstroConfig, pip
 		formatErrorMessage(errorWithMetadata, pipeline.logger.level() === 'debug')
 	);
 
-	return err;
+	return {
+		error: err,
+		errorWithMetadata
+	};
 }
--- a/packages/astro/src/vite-plugin-astro-server/error.ts
+++ b/packages/astro/src/vite-plugin-astro-server/error.ts
@@ -1,0 +1,30 @@
+import type { ModuleLoader } from '../core/module-loader/index.js'
+import type { AstroConfig } from '../@types/astro.js';
+import type DevPipeline from './devPipeline.js';
+
+import { collectErrorMetadata } from '../core/errors/dev/index.js';
+import { createSafeError } from '../core/errors/index.js';
+import { formatErrorMessage } from '../core/messages.js';
+import { eventError, telemetry } from '../events/index.js';
+
+export function recordServerError(loader: ModuleLoader, config: AstroConfig, pipeline: DevPipeline, _err: unknown): Error {
+	const err = createSafeError(_err);
+
+	// This could be a runtime error from Vite's SSR module, so try to fix it here
+	try {
+		loader.fixStacktrace(err);
+	} catch {}
+
+	// This is our last line of defense regarding errors where we still might have some information about the request
+	// Our error should already be complete, but let's try to add a bit more through some guesswork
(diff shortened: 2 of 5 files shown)


matthewp:
!preview handle-unhandled

ematipico (approved):

packages/astro/src/vite-plugin-astro-server/plugin.ts
```diff
@@ -43,6 +44,16 @@ export default function createVitePluginAstroServer({
 			viteServer.watcher.on('unlink', rebuildManifest.bind(null, true));
 			viteServer.watcher.on('change', rebuildManifest.bind(null, false));
 
+			function handleUnhandledRejection(rejection: any) {
+				const error = new AstroError({
+					...AstroErrorData.UnhandledRejection,
+					message: AstroErrorData.UnhandledRejection.message(rejection?.stack || rejection)
+				})
+				logger.error(null, error.message);
+			}
```
Princesseuh, line 53: It's a bit of a edge case, but presumably this should pass through our normal-ish error pipeline so we get a nicely formatted error. How does this look in the browser too?

matthewp: It can't pass through the normal error pipeline, it happens outside of the context of a request. You could do: `setTimeout(() => Promise.reject('foo'), 5000)` and there is no ongoing request, the rejection still happens.

It doesn't currently show up in the browser, do we have a way to send an error from the server to the client outside of a request?

.changeset/curvy-lobsters-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Prevent dev server from crashing on unhandled rejections
```
sarah11918, line 5:
```suggestion
Prevents dev server from crashing on unhandled rejections, and adds a helpful error message
```

sarah11918 (approved):
Error message looks good to me! :partying_face:

packages/astro/src/vite-plugin-astro-server/plugin.ts
Princesseuh: I would think that from where you are in the code, you'd be able to send to the websocket since you're so close to the server creation, here's how it works in the rendering: https://github.com/withastro/astro/blob/f8b27b07c064e84a9b0c51c16fe7fdb72b2ef5f6/packages/astro/src/vite-plugin-astro-server/response.ts#L32

You should just need to send the error payload through the websocket and it should work!»

> It can't pass through the normal error pipeline, it happens outside of the context of a request. 

If you can't make it go through the entire pipeline, at least formatting the error message would be great: https://github.com/withastro/astro/blob/f8b27b07c064e84a9b0c51c16fe7fdb72b2ef5f6/packages/astro/src/core/messages.ts#L257

matthewp: I think AsyncLocalStorage can maybe allow this, going to give that try.

matthewp:
!preview handle-unhandled

packages/astro/src/vite-plugin-astro-server/plugin.ts
natemoo-re: And so it begins... I have _many_ ideas for AsyncLocalStorage!

matthewp: Updated the PR and it should work now. Updated with a screenshot of what the browser receives. Would appreciate another review @Princesseuh

packages/astro/src/vite-plugin-astro-server/plugin.ts
```diff
@@ -57,13 +80,15 @@ export default function createVitePluginAstroServer({
...
 						return;
 					}
-					handleRequest({
-						pipeline,
-						manifestData,
-						controller,
-						incomingRequest: request,
-						incomingResponse: response,
-						manifest,
+					localStorage.run(request, () => {
```
matthewp, line 83: Passing in the `request` which acts as the context for the local storage. When an unhandled rejection occurs, we can use this to know which request it was associated with.

packages/astro/src/vite-plugin-astro-server/plugin.ts
```diff
@@ -43,6 +50,22 @@ export default function createVitePluginAstroServer({
...
 
+			function handleUnhandledRejection(rejection: any) {
+				const error = new AstroError({
+					...AstroErrorData.UnhandledRejection,
+					message: AstroErrorData.UnhandledRejection.message(rejection?.stack || rejection)
+				});
+				const store = localStorage.getStore();
+				if(store instanceof IncomingMessage) {
+					const request = store;
+					setRouteError(controller.state, request.url!, error);
```
matthewp, line 61: This marks the route as being in an error state.

packages/astro/src/vite-plugin-astro-server/plugin.ts
ematipico: First time I asse `AsyncLocalStorage` in action!

packages/astro/src/vite-plugin-astro-server/error.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	telemetry.record(eventError({ cmd: 'dev', err: errorWithMetadata, isFatal: false }));
+
+	pipeline.logger.error(
+		null,
+		formatErrorMessage(errorWithMetadata, pipeline.logger.level() === 'debug')
+	);
+
+	return {
+		error: err,
```
ematipico, line 30: It seems we don't use this at the moment

ematipico (approved):

Princesseuh (approved):
Not 100% satisfied with the stack trace in the error message, but since we don't show `cause` super well and it's a bit of a edge case, it's okay, we can improve it later if we get requests.

Issue #8627, opened 2023-09-22
Content-Type missing when returning new Response with 404 status


wishrd:
### Astro Info

```block
Astro                    v3.1.2
Node                     v18.15.0
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             none
```

### Describe the Bug

Declaring a dynamic route like `/[param1]/[param2]`, if there is a return `new Response(null, { status: 404 })` due to any reason in that route, the server side rendering of the `404.astro` page happens but the Content-Type is missing.

I've the security header `X-Content-Type-Options: nosniff` which check the presence of the Content-Type, breaking the rendering of the 404 by the browser. Removing the header, the 404 page is rendered successfully.

Also, I've noticed the middleware I have adding the security header is executed twice in this case:
```
SECURITY MIDDLEWARE http://localhost:4321/es/asdasd HeadersList {
  cookies: null,
  [Symbol(headers map)]: Map(1) {
    'x-content-type-options' => { name: 'X-Content-Type-Options', value: 'nosniff' }
  },
  [Symbol(headers map sorted)]: null
}

SECURITY MIDDLEWARE http://localhost:4321/es/asdasd HeadersList {
  cookies: null,
  [Symbol(headers map)]: Map(2) {
    'content-type' => { name: 'content-type', value: 'text/html' },
    'x-content-type-options' => { name: 'X-Content-Type-Options', value: 'nosniff' }
  },
  [Symbol(headers map sorted)]: null
}
```
The first time, it doesn't contains the Content-Type but the second one yes. I think the browser is receiving the first one.

This is the use case we have in our application but probably it can be reproduced with other routes if a `new Response` is returned.

It has been tested building and running the server with a NodeJS adapter, all routes with SSR: `npm run build && node dist/server/entry.mjs`. Stackblitz is responding with error too for this use case but the response is a bit different:
* Local: HTML rendered as string
* Stackblitz: Error without rendering

I think the right fix is having the Content-Type and/or stop processing the request twice. If anyone prefer a Github repository instead of the stackblitz one, let me know. Thank you for your help!

### What's the expected result?

The http response after returning a `Response` object by any route should have the Content-Type header.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-g7nwcp

- Go to /whatever, it should display 404
- Go to /test/test, it should display the page
- Go to /test/whatever, it should display 404 but it doesn't

### Participation

[...]

PR #9433, opened 2023-12-14
fix: merge headers from the original response in error pages

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8627

## Testing

Added a new test case from the original reproduction

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -411,12 +411,18 @@ export class App {
 			  ? newResponse.status
 			  : originalResponse.status;
 
+		try {
+			// this function could throw an error...
+			originalResponse.headers.delete('Content-type');
+		} catch {}
 		return new Response(newResponse.body, {
 			status,
 			statusText: status === 200 ? newResponse.statusText : originalResponse.statusText,
 			// If you're looking at here for possible bugs, it means that it's not a bug.
 			// With the middleware, users can meddle with headers, and we should pass to the 404/500.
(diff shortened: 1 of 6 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -398,21 +402,25 @@ export class App {
...
-			statusText: status === 200 ? newResponse.statusText : statusText,
-			headers: new Headers(Array.from(headers)),
+			statusText: status === 200 ? newResponse.statusText : originalResponse.statusText,
+			// If you're looking at here for possible bugs, it means that it's not a bug.
+			// With the middleware, users can meddle with headers, and we should pass to the 404/500.
+			// If users see something weird, it's because they are setting some headers they should not.
+			headers: new Headers([
+				...Array.from(newResponse.headers),
+				...Array.from(originalResponse.headers),
+			]),
```
lilnasy, line 423: This will have an odd effect when the originalResponse is `new Response("whatever string", { status: "404" })`

ematipico: What do you suggest to do?

lilnasy: I think the solution would be to remove `Content-type` from originalResponse.

We are replacing content, it's natural we replace content type too.

ematipico: Done, I also updated the test case where we attempt to change the content type header

lilnasy (approved):
Looks great to me!

PR #9434, opened 2023-12-14
fix: improve error message when a middleware doesn't return a `Response`

ematipico:
## Changes

The error message was old and didn't explain well the solution to solve the error.

## Testing

N/A

## Docs

### DIFF
--- a/.changeset/selfish-parents-leave.md
+++ b/.changeset/selfish-parents-leave.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Better error message when a middleware doesn't return a `Response`
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -733,8 +733,9 @@ export const ResponseSentError = {
  */
 export const MiddlewareNoDataOrNextCalled = {
 	name: 'MiddlewareNoDataOrNextCalled',
-	title: "The middleware didn't return a response or call `next`.",
-	message: 'The middleware needs to either return a `Response` object or call the `next` function.',
+	title: "The middleware didn't return a response or call and return `next`.",
+	message:
+		'The middleware needs to either return a `Response` object or call the `next` function and return the `Response` that it yields.',
 } satisfies ErrorData;
 
 /**


.changeset/selfish-parents-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Better error message when a middleware doesn't return a `Response`
```
sarah11918, line 5:
```suggestion
Improves the error message when a middleware doesn't return a `Response`
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -733,8 +733,9 @@ export const ResponseSentError = {
  */
 export const MiddlewareNoDataOrNextCalled = {
 	name: 'MiddlewareNoDataOrNextCalled',
-	title: "The middleware didn't return a response or call `next`.",
-	message: 'The middleware needs to either return a `Response` object or call the `next` function.',
+	title: "The middleware didn't return a response or call and return `next`.",
```
sarah11918, line 736:
```suggestion
	title: "The middleware didn't return a `Response`.",
```
The title as written is fine, BUT, I'm wondering whether actually saying this in the message is more helpful. The problem is there's no Response returned (however that gets returned, either directly or by calling `next()`). And I think this could actually be useful when combined with my suggestion for the next line, which really emphasizes the need for A RESPONSE (even if it's the one from `next`)

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -733,8 +733,9 @@ export const ResponseSentError = {
...
+	message:
+		'The middleware needs to either return a `Response` object or call the `next` function and return the `Response` that it yields.',
```
sarah11918, line 738:
```suggestion
		'Make sure your middleware returns a `Response` object, either directly or by returning the `Response` from calling the `next` function.
```
If this is accurate, then I think this can help make the emphasis you wanted? What do you think?

sarah11918 (commented):
I would be OK with the error message as written, *but*, if my suggestion is technically correct, then I think it might help get the point across!  Whatever you think!

ematipico:
All good suggestions, thank you @sarah11918

ElianCodes (approved):
Looking all sharp!

MoustaphaDev (approved):

PR #9435, opened 2023-12-14
feat: add tunneling option for dev server

G3root:
## Changes

this PR adds the ability to tunnel dev server to the internet using cloudflare.

Screenshot 2023-12-15 00:25:47

## Testing
run the dev server with the `--tunnel `flag

## Docs

TODO


ematipico (changes requested):
Blocking because it's a `minor`. We will review it later, thank you!

bluwy:
I'm not really sure this should be builtin. It's relying on Cloudflare and isn't something the user usually want. Couldn't this be created as a Vite plugin or Astro integration instead?

matthewp:
I agree, not comfortable with adding a core feature that relies on one specific vendor.

Issue #9425, opened 2023-12-13
Incorrect script url when rollup output options are configured


dkobierski:
### Astro Info

```block
Astro                    v4.0.4
Node                     v20.9.0
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/vercel/serverless
Integrations             none
```

### Describe the Bug

Generated URLs for scripts in astro files are incorrect when using SSR and rollup output options - [customizing output file names](https://docs.astro.build/en/guides/configuring-astro/#customising-output-filenames). I have checked node and vercel adapters and this issue occurs for both of them, it works in SSG though.

You can check that the generated script src [here](https://astro-script-issue.vercel.app/test/test) is using relative path instead of an absolute one. I've added moment package just to make the script bigger, so it goes to a separate file (smaller ones are inlined in html) - it could by anything in the script. Additionally when you configure assetsPrefix in astro options it is not added to the script path as well.

My current workaround is to make an empty Preact component and put the scripts in that file, so it's loaded as an island - islands script paths are working correctly.

### What's the expected result?

The src of the script is generated properly using absolute path. When assetsPrefix is specified in the astro options it should be taken into account.

### Link to Minimal Reproducible Example

https://github.com/dkobierski/astro-script-issue

### Participation

- [X] I am willing to submit a pull request for this issue.

ematipico:
Fixed by https://github.com/withastro/astro/pull/9437

PR #9437, opened 2023-12-14
fix: handle external hoisted scripts correctly

dkobierski:
## Changes

Handles #9425 

## Testing

Confirmed while debugging building of a sample app that assumption of a `.js` file extension is the root cause of wrong asset paths. Unit tests added.

## Docs

N/A


ematipico (approved):
Lots of tests! That's great, thank you

Issue #6221, opened 2023-02-11
custom integration: injected catch-all route overwrites local routes


nowaythatworked:
### What version of `astro` are you using?

2.0.10

### Are you using an SSR adapter? If so, which one?

Vercel (error on dev)

### What package manager are you using?

yarn

### What operating system are you using?

mac

### Describe the Bug

I´m injecting a "catch-all" route `[...slug]` inside a custom integration. When doing this all other routes I create inside my pages folder can not be found.
I guess something is wrong in the prioritisation.

I´m willing to help fix this if someone can point me to the right direction in code.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-g758sr?file=astro.config.ts

### Participation

- [ ] I am willing to submit a pull request for this issue.

nowaythatworked:
@Princesseuh are you sure this is a duplicate? 
The entrypoint is not pointing to files inside src/pages in fact it's in its own package. Also this is not about not catching two of the same routes but wrong priorisation. 

Usually, inside src/pages you can create a file called '[...args].Astro' and the same directory 'index.Astro'. the index will be served when requesting '/' and this is correct. But as soon as an integration registers the pattern '[...args]' this system breaks. Index is no longer served.

nowaythatworked:
Hi @Princesseuh, since this is quite important to me I would love to help fixing this. Could you or somebody else in the team point me towards where the logic for this sits in the code? I´d be happy to create a PR :)

Princesseuh:
I believe @bluwy has the necessary context to help with this, though you might have to wait until the work week start for an answer from him.

bluwy:
It looks like injected route taking the highest priority is an intentional decision:

https://github.com/withastro/astro/blob/869197aafd9802d059dd8db1ef23794fdd938a91/packages/astro/src/core/routing/manifest/create.ts#L430-L443

The PR that added it is https://github.com/withastro/astro/pull/4087. I think what we need is a middleground where injected routes are sorted together with user routes, but injected routes take higher priority than user routes when they have equal priority.

I'm not sure how that can be achieved. Looks like it'll take some leg work to combine both sorted routes, but if you're interested in taking a look, the linked code above (injected routes sorting) and [this user routes sorting code](https://github.com/withastro/astro/blob/869197aafd9802d059dd8db1ef23794fdd938a91/packages/astro/src/core/routing/manifest/create.ts#L306) would be the place to start tweaking.

[...]

PR #9439, opened 2023-12-15
Implement priority overrides for injected routes and redirects

### LINKED ISSUE #9392: Injected SSR routes not following routing priority order
### Astro Info

```block
$ astro info
Astro                    v4.0.3
Package Manager          yarn
Integrations             test-integration
```

_No response_

### Describe the Bug

Non-prerendered routes added by integrations override any existing matching route.

The injected route should follow the same [route priority order](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) as if it were part of the project.

https://stackblitz.com/edit/github-lukzfz-wradsh

### Participation

Fryuni:
## Changes

Refactor route sorting logic to follow the priority order as specified in the docs in order:
- Static routes without path parameters will take precedence over all other routes
- Dynamic routes using named parameters take precedence over rest parameters
- Pre-rendered dynamic routes take precedence over server dynamic routes
- Rest parameters have the lowest priority
- Endpoints always take precedence over pages
- Ties are resolved alphabetically

Injected routes and redirects will now be sorted alongside routes from project files, with an option to switch to the previous "legacy" behavior.

The "legacy" behavior for redirects is to be sorted below any other route and for injected routes is to be sorted above.

### Collision detection

The previous collision detection relied on routes being sorted in steps, which is exactly what this PR intends to remove, so this also changes the collision detection. Cases like `/[foo] (redirect)` and `/[bar] (ssr route)` are now reported as collisions.

Routes that may result in a collision depending on their returns on `getStaticPaths` are not reported as colliding. Routes that only partially collide are also not reported since that may be intentional (like `/[...slug]` and `/api/[operation]`).

### Previous divergence from the docs

Static redirects and dynamic pages did not resolve as documented in the previous code.

Example: `/foo (redirect)` and `/[slug] (page)`.  
According to the docs, the redirect should be shadowed by the page route.  
But by the previous code, the redirect would be sorted above the page.

### New divergence from the docs

The docs do not specify how to handle scenarios that differ only in specificity, like `/[slug]` and `/foo/[slug]`. So this PR includes an extra rule to sort more specific routes higher than less specific routes. So:

- `/blog/[...slug] (page)` is sorted above `/[...slug] (page)`. This was sorted the same way previously due to how the items were sorted as they were crawled.
- `/blog/[...slug] (injected page)` is sorted above `/[...slug] (injected page)`. This was previously kept in the order that `injectRoute` was called (BREAKING CHANGE).

### Breaking changes... kinda

- `/blog/[...slug] (injected page)` is sorted above `/[...slug] (injected page)`. This was previously kept in the order that `injectRoute` was called.
- `/foo (redirect)` is now sorted below `/[slug] (page)` by default, as documented.

Since those changes in behavior were not tested and either not documented or behaving differently from what was documented, I considered them more bug fixes than breaking changes, so I didn't set the changeset to major.

I can change it to major if that is the prevailing decision.

### Summary of the changes

- Better collision reporting when the collision is certain
- Injected routes default to be sorted alongside project routes
- Redirects default to be sorted alongside project routes
- Routes sorted using the following rules in order of relevance:
  - More specific routes (with more segments) take precedence over less specific routes
  - Static routes (without any path parameter) take precedence over dynamic routes
  - Dynamic routes using named parameters take precedence over rest parameters
  - Pre-rendered routes take precedence over server routes (not just for dynamic routes)
  - Endpoints take precedence over pages
  - Lastly, routes are sorted alphabetically

## Testing

Unit tests were added for all the cases mentioned above. If the team thinks it is necessary, I can also add extra integration tests.

## Docs

The changes in behavior are described above.
The docs should be updated to use the rules from the summary above in the "Route Priority Order" section.
The new `priority` option should be added in the reference documentation for redirects and `injectRoute`.

I can send the docs PR as soon as the new behavior is approved.

## Relations

- Closes #9392


Fryuni:
I'm not sure about the names of the priorities. They are technically correct but might also be confusing. I also considered the following alternatives:

- `override`/`normal`/`fallback`
- `higher`/`normal`/`lower`

Any other ideas?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -476,35 +448,132 @@ export function createRouteManifest(
...
+		...[
+			...projectRoutes,
+			...injectedRoutes['same-as-project'],
+			...redirectRoutes['same-as-project'],
+		].sort(comparator),
+		...[
+			...injectedRoutes['below-project'],
+			...redirectRoutes['below-project'],
+		].sort(comparator),
+	];
```
ematipico, line 557: I don't think using spread operators is the correct choice here. It seems we always create new arrays, which is a hit memory-wise. 

I believe we should mutate the ones we have and create only one.

E.g.
```js
const routes = [];
routes.concat(
	injectedRoutes['above-project'].concat(
		redirectRoutes['above-project']
	).sort(comparator)
);
```

I think this is safe because we aren't going to use `injectedRoutes` and `redirecteRoutes` anymore.

packages/astro/src/@types/astro.ts
```diff
@@ -1555,10 +1555,19 @@ export interface AstroUserConfig {
  */
 export type InjectedScriptStage = 'before-hydration' | 'head-inline' | 'page' | 'page-ssr';
 
+/**
+ * IDs for different priorities of injected routes and redirects:
+ * - "above-project": Override any project route in case of conflict.
+ * - "same-as-project": Behave the same as if the route was defined in the project, following the same priority rules.
+ * - "below-project": Only match if no project route matches.
+ */
+export type RoutePriorityOverride = 'above-project' | 'same-as-project' | 'below-project';
```
ematipico, line 1564: If this is only applied to injected routes, we should choose a name that tells us that, e.g. `InjectedRoutePriorityOverride`

packages/astro/src/core/config/schema.ts
```diff
@@ -173,6 +173,7 @@ export const AstroConfigSchema = z.object({
 						z.literal(308),
 					]),
 					destination: z.string(),
+					priority: z.enum(['above-project', 'same-as-project', 'below-project']).optional(),
```
ematipico, line 176: I believe we should at least provide a default.

packages/astro/src/@types/astro.ts
```diff
@@ -1555,10 +1555,19 @@ export interface AstroUserConfig {
  */
 export type InjectedScriptStage = 'before-hydration' | 'head-inline' | 'page' | 'page-ssr';
 
+/**
+ * IDs for different priorities of injected routes and redirects:
+ * - "above-project": Override any project route in case of conflict.
+ * - "same-as-project": Behave the same as if the route was defined in the project, following the same priority rules.
+ * - "below-project": Only match if no project route matches.
```
ematipico, line 1562: @withastro/maintainers-docs You should focus here. Naming is very important here

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -363,69 +318,83 @@ export function createRouteManifest(
...
+			: null;
+		const params = segments
+			.flat()
+			.filter((p) => p.dynamic)
+			.map((p) => p.content);
+		const route = `/${segments
+			.map(([{dynamic, content}]) => (dynamic ? `[${content}]` : content))
+			.join('/')}`.toLowerCase();
+
+		routes[priority ?? 'above-project'].push({
```
ematipico, line 370: Why is the default here `above-project` and at line 441 is `below-project`?

packages/astro/src/@types/astro.ts
```diff
@@ -1555,10 +1555,19 @@ export interface AstroUserConfig {
  */
 export type InjectedScriptStage = 'before-hydration' | 'head-inline' | 'page' | 'page-ssr';
 
+/**
+ * IDs for different priorities of injected routes and redirects:
```
ematipico, line 1559: I am not really sure redirects should be configured. I think redirects should have the highest priority above all routes.

ematipico (commented):
Some preliminary comment

packages/astro/src/core/routing/manifest/create.ts
Fryuni: I don't think it will make much of a difference. Each `.concat` is a possible allocation while multiple spread operations inside the same `[]` are always a single allocation with the correct size (the sum of the length of all the arrays being spread).

Safety-wise you are correct, using concat is not a safety problem.

I'll try to run some benchmarks for that

packages/astro/src/core/config/schema.ts
Fryuni: The default is added where it is used, but I can also add it here.

 I didn't do it because then there would be a risk of the defaults becoming out of sync. If the redirect uses just the destination string instead of an object, there is no priority in the configuration, so the default has to be where it is used.

packages/astro/src/core/routing/manifest/create.ts
Fryuni: Injected routes default to be above the project, and redirects default to be below the project.

Redirects defaulting to being below is what was documented and previously implemented.

Injected routes defaulting to being above were not documented, but they were already implemented like that, and from my quick test, it seems some integrations rely on that behavior, so this is to minimize breaking changes to just the edge cases described in the PR.

packages/astro/src/@types/astro.ts
Fryuni: It is documented that they have the lowest priority of all the routes, so I didn't change that default. I personally think both of them should have the same priority as the project routes by default, but that would break way too many things.

I wasn't going to make them configurable, but it was pointed out on Discord that there are similar problems for them as there are for injected routes (albeit inversed since they go below the project, not above), so I fixed them together here.

packages/astro/src/core/routing/manifest/create.ts
Fryuni: I stand corrected. `concat` takes half the time of the spread operator, unless the function is called enough times for all JIT passes, then it is just 2% faster (but still consistently faster).

Although I just tested this for 3.5k routes, 500 each of (injected above, redirect before, project, inject with project, redirect with project, injected below, redirect below) and it took about 0.5ms even with JIT disabled, so it might not be a worth optimization at the cost of readability.

I think the code with the spreads is more clear about what would be the final ordering than this:
```ts
const routes = injectedRoutes['above-project']
  .concat(redirectRoutes['above-project'])
  .sort(comparator)
  .concat(
    projectRoutes
      .concat(injectedRoutes['same-as-project'], redirectRoutes['same-as-project'])
      .sort(),
    injectedRoutes['below-project']
      .concat(redirectRoutes['below-project'])
      .sort()
  );
```

packages/astro/src/@types/astro.ts
matthewp: I think the naming should involve file-based routing somehow, rather than project. Took me a second to realize that's what project meant.

delucis: I also wonder if “above”/“below” is ideal. I understand why it was chosen, but I associate high/low more with priority.

Maybe something like this could work? @sarah11918 is also a good person to bounce these off. I don’t _love_ “file-system” here — not sure it’s super clear that this means “file-based route” or if there’s something clearer.

Could ever be other kinds of priority? If not, it could be redundant to mention that in the enum values and the options could just be `high`/`default`/`low` or something? With the relationship explained in the documentation.

```suggestion
 * IDs for different priorities of injected routes and redirects:
 * - "higher-than-file-system": Override any project route in case of conflict.
 * - "same-as-file-system": Behave the same as if the route was defined in the project, following the same priority rules.
 * - "lower-than-file-system": Only match if no project route matches.
```

florian-lefebvre:
Do you think it can close https://github.com/withastro/astro/issues/6221 ?

Fryuni:
> Do you think it can close https://github.com/withastro/astro/issues/6221 ?

Seems like the same problem to me.

It would require change on the integration registering the catch-all route to use a lower priority than the default since this keeps the current behavior by default

packages/astro/src/@types/astro.ts
Fryuni: Another option would be to use a number, with all project routes having priority zero, injected routes defaulting to a priority of -1 and redirects defaulting to a priority of 1.

Then the other rules apply for when this number is the same.

This may be easier to document and even more flexible. Maybe too flexible and would make things confusing for integration developers.

delucis: Yes, I’d caution against that I think. Even if internally the enum maps to `-1 | 0 | 1`, allowing people to potentially try to outbid each other with ever higher numbers seems really risky. (And doesn’t actually help with clarity `high` or `higher-than-file-system` tell you something quite explicit, where `1` is pretty meaningless on its own.)

Fryuni: Agreed, I also don't like the numbers idea.

packages/astro/src/core/config/schema.ts
ematipico: > so the default has to be where it is used.

From this phrase, I understand "where it is used" and *this configuration* aren't the same. Did I understand correctly?

packages/astro/src/core/routing/manifest/create.ts
ematipico: It would be great to leave a comment to explain that

Fryuni: Added comments for both defaults

packages/astro/src/core/config/schema.ts
Fryuni: Yes, I mean the code on `create.ts` that uses the value from the configuration to do something.

packages/astro/src/@types/astro.ts
sarah11918: @Fryuni OK, so my gut reaction was to look at your definitions to see what terms might make sense. The first thing that comes to mind is something like:

Injected/Redirect Route Priorities:
- `override` - takes higher priority than other defined routes in your project
- `default` - integrates/conforms with the other routes defined in your project
- `defer` - route is only prioritized after all other normally-defined routes in your project

(Admittedly, I had the hardest time thinking of the last one) Maybe that gives you another direction to think in?

packages/astro/src/core/config/schema.ts
delucis: Is the configuration ever **not** passed through this schema?

I think what @ematipico is suggesting is this:

```suggestion
					priority: z
						.enum(['above-project', 'same-as-project', 'below-project'])
						.default('above-project'),
```

This ensures `priority` is always defined and no checks need to be made later.

packages/astro/src/@types/astro.ts
delucis: I like that direction @sarah11918!

packages/astro/src/core/config/schema.ts
Fryuni: Redirects can be just a string:
```ts
{
  redirects: {
    '/foo': '/bar'
  }
}
```


So the code code on `create.ts` would have to duplicate the default anyway. We transform it here... I'll send an option for that

delucis: You could use a Zod transform to do that upfront too, iiuc!

packages/astro/src/@types/astro.ts
Fryuni: I also liked @sarah11918's suggestion. I'll send a commit with those and change from just "project routes" to "project file-based routes" to make it clear what they are relative to.

packages/astro/src/core/config/schema.ts
Fryuni: Using a transform now!

packages/astro/src/@types/astro.ts
Fryuni: Just sent a commit with a new suggestion. I first used the terms by @sarah11918, but it felt weird to have a priority named "default" that is not the default, so I went with "merge".

Also changed the "project routes" to clarify that those are the routes inferred from files.

What do you all think? Better, worse, still confusing?

sarah11918: Right, I agree not to use default as an option that is not the default.  😀 

`merge` could work. Synonyms would be like `integrate` or `incorporate` or maybe even `include`.

If you wanted the "conform" (follow the rules) angle, then other synonyms could be `follow`, `obey`, `observe`, `match`, `comply`.

I'm not sure I have a strong preference. Just highlighting some options to see if one immediately jumps out at you!

Fryuni: Out of those, I think I still prefer `merge`, with `incorporate` as a close second.

sarah11918 (commented):
Just adding to remove the request for review since I've given my feedback below

packages/astro/src/@types/astro.ts
```diff
@@ -1555,10 +1555,22 @@ export interface AstroUserConfig {
...
+ * - "override": Override any file-based project route in case of conflict. Conflicts within override
+ *   routes are resolved by the same rules as file-based routes.
+ * - "merge": Merge with discovered file-based project routes, behaving the same as if the route 
+ *   was defined as a file in the project.
+ * - "defer": Defer to any file-based project route in case of conflict. Conflicts within defer
+ *   routes are resolved by the same rules as file-based routes.
+ */
+export type RoutePriorityOverride = 'override' | 'normal' | 'defer';
```
sarah11918, line 1567: Is the plan for this to be user-configurable @Fryuni ? If so, where/how? 

If in `astro.config.mjs` then this file will also need a corresponding `@docs` entry.

Also reminder that you will also need an accompanying PR to update: https://docs.astro.build/en/core-concepts/routing/#route-priority-order 

(FYI It's OK if this update doesn't entirely duplicate the config option, but instead just links to it, if you're also adding it to config reference. It might be most helpful if this sections lists out the default, then links to the configurable option for "fine-tuning")

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,20 @@
+---
+'astro': minor
+---
+
+Reworks route priority processing to allow for more flexible and intuitive redirects and route injection
```
sarah11918, line 5: Let me know when this is considered finalized behaviour/naming etc, and please `/ptal` me for a review at that time!

packages/astro/src/@types/astro.ts
Fryuni: On redirects this is user-configurable, on routes this is configurable for integrations.

matthewp: I don't think injected routes should have the ability to decide how they should be prioritized. I think the only options should be "normal" (ie, the same as if they were file-based routes) and maybe something called "legacy" (the old behavior, only for backwards compatibility, to be removed in next major).

Fryuni: I think the customization is positive for integrations.

For example, Starlight injects a 404 route. With this change it could be injected with `priority: 'defer'` allowing projects using Starlight to define a custom 404 page. As it is now this is impossible, Starlight will always override the 404 page.

Also, if we change to have a "normal" and "legacy", I'd expect the "normal" to be the default. But that would be a breaking change. I'm not sure how much of an impact this would be on integrations in the wild, but IMHO it would require a major version by itself.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -137,77 +132,40 @@ function validateSegment(segment: string, file = '') {
 	}
 }
 
-function comparator(a: Item, b: Item) {
-	if (a.isIndex !== b.isIndex) {
-		if (a.isIndex) return isSpread(a.file) ? 1 : -1;
-
-		return isSpread(b.file) ? -1 : 1;
+function comparator(a: RouteData, b: RouteData) {
```
ematipico, line 135: Can we have a comment that explains the business logic of the sorting logic?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -137,77 +132,40 @@ function validateSegment(segment: string, file = '') {
...
-	{ config, cwd }: { config: AstroConfig; cwd?: string },
-	{ pattern, entrypoint }: InjectedRoute
-): Item {
-	let resolved: string;
-	try {
-		resolved = require.resolve(entrypoint, { paths: [cwd || fileURLToPath(config.root)] });
-	} catch (e) {
-		resolved = fileURLToPath(new URL(entrypoint, config.root));
+	// endpoints take precedence over pages
+	if ((a.type === 'endpoint') !== (b.type === 'endpoint')) {
```
ematipico, line 163: If the endpoints take precedence, wouldn't make more sense to move this check at the beginning of the function?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -137,77 +132,40 @@ function validateSegment(segment: string, file = '') {
...
-		}
+	// rest parameters go last
+	if (aHasSpread !== bHasSpread) {
+		return aHasSpread ? 1 : -1;
 	}
 
-	// endpoints are prioritized over pages
-	if (a.isPage !== b.isPage) {
-		return a.isPage ? 1 : -1;
+	// prerendered routes go first
```
ematipico, line 157: Go first? Even before the endpoints?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -363,69 +319,85 @@ export function createRouteManifest(
...
-				pathname: pathname || void 0,
-				prerender: prerenderInjected ?? prerender,
-				fallbackRoutes: [],
+	const routes: PrioritizedRoutesData = {
+		'override': [],
+		'normal': [],
+		'defer': [],
+	};
+
+	settings.injectedRoutes.forEach(({pattern: name, entrypoint, prerender: prerenderInjected, priority}) => {
```
ematipico, line 337: While we are at it, can we remove the `forEach` call for a simple `for..of` loop?

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -476,35 +441,134 @@ export function createRouteManifest(
...
 
-		const redirBase = path.posix.dirname(route);
-		const dynamicRedir = lastSegmentIsDynamic(routeData);
-		let i = 0;
-		for (const existingRoute of routes) {
-			// An exact match, prefer the page/endpoint. This matches hosts.
-			if (existingRoute.route === route) {
-				routes.splice(i + 1, 0, routeData);
-				return;
+function routesCollide(a: RouteData, b: RouteData) {
```
ematipico, line 458: It's worth adding a comment to the function that explains what this function is in charge of, and a brief explanation of its business logic.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -476,35 +441,134 @@ export function createRouteManifest(
...
+		return false;
+	}
+
+	if (a.type === 'fallback' || b.type === 'fallback') {
+		// If either route is a fallback route, they don't collide.
+		// Fallbacks are always added below other routes exactly to avoid collisions.
+		return false;
+	}
+
+	const length = a.segments.length;
```
ematipico, line 469: Why do we choose the lengh of `a` and not `b`? Now I question why we call the two parameters `a` and `b`. Shouldn't we choose the great lenght between `a` and `b`? Otherwise if `b.length` is lower than `a.length`, we overflow.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -476,35 +441,134 @@ export function createRouteManifest(
...
+
+	for (let i = 0; i < length; i++) {
+		const segmentA = a.segments[i];
+		const segmentB = b.segments[i];
+
+		if (segmentA.length !== segmentB.length) {
+			return false;
+		}
+
+		const segmentLength = segmentA.length;
```
ematipico, line 480: Again, we should explain why we use `a` and not `b`. Surely there's a reason, but it's not documented.

ematipico (changes requested):
I finally had some time to look into this PR. Thank you for waiting. 

As hinted by Matthew, we would like to reduce the goal of the PR to **predictable** results. While giving the ablity to integration developers to change the priority of routes is tempting, it doens't make that predictable, especially when multiple integrations start injecting their routes, which eventually **could** collide.

What we could do, instead, is to error in case there are multiple routes that collide, and explain to the user **why** they collide.

I believe we can tag a route with the integration name, so we know who injected the routes , and tell it to the user. In case this occurs, there isn't much we can do.

packages/astro/src/@types/astro.ts
ematipico: > I think the customization is positive for integrations.

That's a fact, and I agree with your point. However, we risk not providing predictability to the end users, who need to learn what integrations do. While it's OK to change. In the prioritization logic, we also make sure that the user is protected from what integrations do. 

I believe we can progressively provide integrations more control, although we should first fix the predictability of the prioritization.

matthewp: @Fryuni It should be possible for Starlight to define 404 via middleware. I think a lot of what Starlight does might be better via middleware. Currently middleware is a little limited and can't do something like 404, we are in the process of fixing that though.

packages/astro/src/core/config/schema.ts
```diff
@@ -173,8 +173,15 @@ export const AstroConfigSchema = z.object({
 						z.literal(308),
 					]),
 					destination: z.string(),
+					priority: z.enum(['override', 'normal', 'defer'])
```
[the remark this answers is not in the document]
Fryuni: Changing the default would be a breaking change and IMO require a major version. It was documented that redirects are a lower priority than the project files, so I think it is fair for projects to rely on that.

The tests for Node adapter for example rely on those defaults.

I'd prefer to not break many things with this PR, but I'm fine with this decision for a breaking change.

packages/astro/src/@types/astro.ts
Fryuni: Rethinking about this, if an integration wants to provide a route that can be overridden by the project it is on, it could just allow disabling the route. If Starlight wants to allow overriding the 404 page it could have a `notFoundRoute: false` and not even call `injectRoute` in that case.

Considering that I don't see much advantage of allowing these different priorities.

Fryuni:
> What we could do, instead, is to error in case there are multiple routes that collide, and explain to the user why they collide.

From the point where this is being checked it is possible to fail the build in case of conflicts between static routes (i.e. `/foo (project file)`, `/foo (injected)` and `/foo (redirect)`) and dynamic routes that are not pre-rendered. But for pre-rendered dynamic routes we can't say for sure that there will be a conflict. You could have `/foo/[bar]` and `/foo/[baz]` and as long as the `getStaticPaths` of both routes don't return the same thing they won't conflict with each other.

Unless we decide to fail on any case that _can_ conflict. In that case having two dynamic routes in the same level would fail even if pre-rendered with different params.

What are the thoughts on that @ematipico and @matthewp?

Fryuni:
Another option would be to split this logic, failing earlier (where it is now) on cases that we know that will conflict and then validate the pre-rendered dynamic routes later on after the static params are present.

packages/astro/src/core/config/schema.ts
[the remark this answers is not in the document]
Fryuni: Correction, there is no test in any package in this repo that rely on the current (now "legacy") behavior. It was triggering an error due to a race condition that is being fixed by https://github.com/withastro/astro/pull/9582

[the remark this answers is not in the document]
ematipico: Apologies, the default should be `"legacy"`, you're correct!

ematipico:
Maybe we can't avoid creating a hard error, but we could at least log an error/warning to notify the user. 

> Another option would be to split this logic, failing earlier (where it is now) on cases that we know will conflict and then validate the pre-rendered dynamic routes later on after the static params are present.

Yeah, that's what I was thinking. If a dynamic route generates a path that was generated before, we could ... not generate the route that just generated a double path and log a warning? And **not** fail the build. Or maybe fail the build? 

I am a bit conflicted. Failing would mean that users can't build their websites anymore. I believe a warning should be more appropriate, and then once we ship the breaking change - Astro v6 - we can hard-error. This should give enough time to users and integration developers to move to the new logic.

Fryuni:
To not change the scope of this PR what about failing on the build on routes that we know at this point that will conflict (static routes and server-rendered dynamic routes) and log a warning for routes that may conflict, but depends on the actual use case in those routes?

I can tackle the option to fail on conflicting pre-rendered dynamic routes on a separate PR since that is on a different stage of the build. Also, it would not cover _all_ possible conflicts, there is one scenario that I think we should log a warning but could very well be intentional:

```
/src/pages/
  - [...slug].astro
  - blog/
    - [post].astro
```

If both of those are pre-rendered, `[...slug]` could have a param `blog/foo` and `blog/[post]` could have the param bar and it would render both pages correctly and without a conflict (although we would warn about a possible conflict).

If both are _not_ pre-rendered the behavior is not the same, `blog/foo` would be handled by `blog/[post]` and not `[...slug]`, possibly with an undesired result. So in this case it _is_ a conflict.

But we can't know if that is intentional or not, if `[...slug]` was never intended to handle `blog/` routes this is the expected behavior.

In that scenario I'd just log the warning

Fryuni:
We could (again, separately and for a future major version) make those two scenarios behave the same and fail if a spread route is generated with a parameter that would be routed into a different route. So the `[...slug]` page in my example would fail if it attempted to pre-render `blog/foo` because there is a `blog/[post]` route.

ematipico:
Yeah, I believe we can make an error in this case. I wonder if you can suggest some potential resolution paths in the error? Probably it's too much, so it's just a nice-to-have :D  

After all, these routes are controlled by the user, so they can find a solution that suits their needs and makes sense.

matthewp:
I'm happy with that plan too!

sarah11918:
Just reminding everyone that there are going to be a *lot* of docs review involved with this, none of which are available yet. :smile:  I'll have to review changeset(s)?, there needs to be some JSDoc for anything user configurable, and an update to the Routing page in docs.

Please do notify me loudly when these are available for me to review, because 4.2.0 is fast approaching!

packages/astro/src/core/routing/manifest/create.ts
Fryuni: They take precedence after the previous rules. So if both an endpoint and a page are static/dynamic/rest and both have the same number of segments, the endpoint will go first.

packages/astro/src/core/routing/manifest/create.ts
Fryuni: Im rewriting all the comments to hopefully make it more clear

sarah11918:
So, just a comment now that I've seen the proposed docs:

I understand that `legacy` makes sense as a name to you *right now* because you're changing something that already exists. But if you were just starting a framework fresh and you saw there were two routing options, "normal" and "legacy"... those tell you absolutely nothing about what is happening, or what they mean.

I would *strongly* suggest that you see if you can actually describe what these two options are, what their difference is. 

It's like calling something "new"... it's not actually "new" for very long. Calling something "legacy" only makes sense for this exact moment in time, really. Imagine seeing these options for the first time: what are you really choosing between?

packages/astro/src/@types/astro.ts
delucis: I was hoping to avoid exactly that kind of option with the priority system.

UX with a `notFoundRoute: false` option:

1. User creates a `404.astro` 
2. They get an error about route collisions (see https://github.com/withastro/starlight/issues/1080)
3. They need to work out Starlight’s option exists — likely not something I can add helpful logging for in Starlight because the error is Astro’s
4. Maybe they set `notFoundRoute: false` if they find it, maybe they assume this isn’t supported.

UX with a lower route priority when injecting Starlight’s 404:

1. User creates a `404.astro`. It works!

ematipico:
> So, just a comment now that I've seen the proposed docs:
> 
> I understand that `legacy` makes sense as a name to you _right now_ because you're changing something that already exists. But if you were just starting a framework fresh and you saw there were two routing options, "normal" and "legacy"... those tell you absolutely nothing about what is happening, or what they mean.
> 
> I would _strongly_ suggest that you see if you can actually describe what these two options are, what their difference is.
> 
> It's like calling something "new"... it's not actually "new" for very long. Calling something "legacy" only makes sense for this exact moment in time, really. Imagine seeing these options for the first time: what are you really choosing between?

@sarah11918 

Then let's use the approach that endpoints usually use: `v1` and `v2`. `v1` will be the default, and `v2` will be the new approach.

Also, I would like to say that these options are only temporary and they will be removed in the next major versions, so I don't see particular value in finding a **good** name.

packages/astro/test/units/routing/manifest.test.js
```diff
@@ -29,7 +37,173 @@ describe('routing - createRouteManifest', () => {
...
+
+		const manifest = createRouteManifest({
+			cwd: fileURLToPath(root),
+			settings,
+			fsMod: fs,
+		});
+
+		expect(getManifestRoutes(manifest)).to.deep.equal([
+			{
+				route: '/[...slug]',
```
ematipico, line 126: Is this route at the first place because it's a "legacy" route?

ematipico:
It seems there are some regressions, some tests are now emitting errors, even though they should not because the new logic should be opt-in.

sarah11918:
Am adding here that now that this is an experimental flag, documentation would normally be:

- an introduction-y changeset, including what the config option looks like in `astro.config.mjs`
- a `@docs` entry in the "experimental flags" section of the `types.ts` file containing ALL the info we're choosing to document about this option.

Experimental flag features only get documented otherwise in docs if they get their own "experimental" page because it's a brand new feature (e.g. when view transitions was newly added and experimental). Since this is a change to routing, not a new feature itself, it should NOT go in docs otherwise so as not to confuse people who have not opted in.

I suggest looking at the format/structure/style of #9644 to model the docs after this other PR that is introducing a new experimental flag. That's pretty much what we want here! And, no other PR submitted to the docs repo is necessary.

packages/astro/test/units/routing/manifest.test.js
Fryuni: Yes, now it is the second because both injected routes are legacy

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -139,77 +138,60 @@ function validateSegment(segment: string, file = '') {
...
+	if (a.segments.length !== b.segments.length) {
+		return a.segments.length > b.segments.length ? -1 : 1;
 	}
 
-	const max = Math.max(a.parts.length, b.parts.length);
-
-	for (let i = 0; i < max; i += 1) {
-		const aSubPart = a.parts[i];
-		const bSubPart = b.parts[i];
+	const aIsStatic = a.segments.every((segment) => segment.every((part) => !part.dynamic && !part.spread));
```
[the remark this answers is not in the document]
Fryuni: We could. Few options:
- Pre-compute this and add to the route data object
- Memoize this on a WeakMap keyed by the route data object
- Pre-compute all the related checks keyed by the route string and pass it around so it gets dropped after all this logic while the route data lives

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -139,77 +138,60 @@ function validateSegment(segment: string, file = '') {
...
+ * - Static routes are sorted before dynamic routes.
+ *   For example, `/foo/bar` is sorted before `/foo/[bar]`.
+ * - Dynamic routes with single parameters are sorted before dynamic routes with rest parameters.
+ *   For example, `/foo/[bar]` is sorted before `/foo/[...bar]`.
+ * - Prerendered routes are sorted before non-prerendered routes.
+ * - Endpoints are sorted before pages.
+ *   For example, a file `/foo.ts` is sorted before `/bar.astro`.
+ * - If both routes are equal regarding all previosu conditions, they are sorted alphabetically.
+ *   For example, `/bar` is sorted before `/foo`.
+ *   The definition of "alphabetically" is dependent on the default locale of the running system.
```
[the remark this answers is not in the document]
Fryuni: Indeed, that is what happens (at least with most default locales, but I don't know about all of them). Same as before.

ematipico (approved):
Great work @Fryuni! Thank you very much for the time you spent on this one, and thank you for your immense patience, considering the scope change from the initial implementation.

Now that we ship this as an experimental feature, we can slowly add new APIs to allow better control of routes from integrations.

.changeset/short-keys-bow.md
```diff
@@ -0,0 +1,15 @@
+---
+"astro": patch
+---
+
+Fix inconsistency between route priorities of file routes, redirects and injected routes
```
sarah11918, line 5:
```suggestion
Updates [Astro's routing priority rules](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) to prioritize the most specifically-defined routes. 

Now, routes with **more defined path segments** will take precedence over less-specific routes.

For example, `/blog/posts/[pid].astro` (3 path segments) takes precedence over `/blog/[...slug].astro` (2 path segments). This means that:

- `/pages/blog/posts/[id].astro` will build routes of the form `/blog/posts/1` and `/blog/posts/a`
- `/pages/blog/[...slug].astro` will build routes of a variety of forms, including `blog/1` and `/blog/posts/1/a`, but will not build either of the previous routes.

For a complete list of Astro's routing priority rules, please see the [routing guide](https://docs.astro.build/en/core-concepts/routing/#route-priority-order).
```

.changeset/short-keys-bow.md
```diff
@@ -0,0 +1,15 @@
...
+
+
+Now all groups are prioritized following the same rules:
+
+- Routes with more path segments will take precedence over less specific routes. E.g. `/blog/post/[pid].astro` takes precedence over `/blog/[...slug].astro`.
+- Static routes without path parameters will take precedence over dynamic routes. E.g. `/posts/create.astro` takes precedence over all the other routes in the example.
+- Dynamic routes using named parameters take precedence over rest parameters. E.g. `/posts/[page].astro` takes precedence over `/posts/[...slug].astro`.
+- Pre-rendered dynamic routes take precedence over server dynamic routes.
+- Endpoints take precedence over pages. E.g. `/posts/[pid].ts` takes precedence over `/posts/[pid].astro`.
+- If none of the rules above decide the order, routes are sorted alphabetically based on the default locale of your Node installation.
```
sarah11918, line 15:
```suggestion
```
Most of these have not changed, so having all the rules listed out here feels more like overwhelming noise that detracts from the main change. I think a link to docs with the rules is fine here!

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,34 @@
+---
+'astro': minor
+---
+
+Adds an experimental flag `stableRoutePriority` to make all routes be prioritized following the same stable rules.
```
sarah11918, line 5:
```suggestion
Adds an experimental flag `globalRoutePriority` to prioritize redirects and injected routes equally alongside file-based project routes, following the same [route priority order rules](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) for all routes.
```

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,34 @@
...
+// astro.config.mjs
+{
+  experimental: {
+    stableRoutePriority: true,
+  },
+}
+```
+
+Enabling this feature makes redirects and injected routes be prioritized along with discovered file-based project
+routes, behaving the same as if all routes were defined as files in the project.
```
sarah11918, line 17:
```suggestion
Enabling this feature ensures that all routes in your project follow the same, predictable route priority order rules. In particular, this avoids an issue where redirects or injected routes (e.g. from an integration) would always take precedence over locally-defined route definitions, making it impossible to override some routes locally.
```

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,34 @@
...
+
+For the following scenario in SSR mode:
```
sarah11918, line 19:
```suggestion
The following table shows which route builds certain page URLs when file-based routes, injected routes, and redirects are combined as shown below:
```

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,34 @@
...
+
+- File-based route: `/blog/post/[pid]`
+- File-based route: `/[page]`
+- Injected route: `/blog/[...slug]`
+- Redirect: `/blog/tags/[tag]` -> `/[tag]`
+- Redirect: `/posts` -> `/blog`
+
+URLs are handled with the following routes:
```
sarah11918, line 27:
```suggestion
URLs are handled by the following routes:
```

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,34 @@
...
+
+| Page               | Current Behavior                 | Stable Routing Priority Behavior   |
+|--------------------|----------------------------------|------------------------------------|
+| `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`          |
+| `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid] |
```
sarah11918, line 32:
```suggestion
| `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid]` |
```

packages/astro/src/@types/astro.ts
```diff
@@ -1566,6 +1566,33 @@ export interface AstroUserConfig {
...
+
+		/**
+		 * @docs
+		 * @name experimental.globalRoutePriority
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.2.0
+		 * @description
+		 *
+		 * Prioritize redirects and injected routes along with discovered file-based project routes, behaving the same as if all routes were defined as files in the project.
```
sarah11918, line 1578:
```suggestion
		 * Prioritizes redirects and injected routes equally alongside file-based project routes, following the same [route priority order rules](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) for all routes.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1566,6 +1566,33 @@ export interface AstroUserConfig {
...
+		 * - Redirect: `/blog/tags/[tag]` -> `/[tag]`
+		 * - Redirect: `/posts` -> `/blog`
+		 *
+		 * URLs are handled with the following routes:
+		 *
+		 * | Page               | Current Behavior                 | Stable Routing Priority Behavior   |
+		 * | ------------------ | -------------------------------- | ---------------------------------- |
+		 * | `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`          |
+		 * | `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid] |
+		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                |
```
sarah11918:
```suggestion
		 * The following table shows which route builds certain page URLs when file-based routes, injected routes, and redirects are combined as shown below:
		 * - File-based route: `/blog/post/[pid]`
		 * - File-based route: `/[page]`
		 * - Injected route: `/blog/[...slug]`
		 * - Redirect: `/blog/tags/[tag]` -> `/[tag]`
		 * - Redirect: `/posts` -> `/blog`
		 *
		 * URLs are handled by the following routes:
		 *
		 * | Page               | Current Behavior                 | Stable Routing Priority Behavior   |
		 * | ------------------ | -------------------------------- | ---------------------------------- |
		 * | `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`          |
		 * | `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid]` |
		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                |
```

sarah11918 (commented):
OK, assuming that I have the content right, here's my suggestion for both changesets and docs!

packages/astro/src/@types/astro.ts
```diff
@@ -1566,6 +1566,33 @@ export interface AstroUserConfig {
...
+		 * - Redirect: `/blog/tags/[tag]` -> `/[tag]`
+		 * - Redirect: `/posts` -> `/blog`
+		 *
+		 * URLs are handled with the following routes:
+		 *
+		 * | Page               | Current Behavior                 | Global Routing Priority Behavior   |
+		 * | ------------------ | -------------------------------- | ---------------------------------- |
+		 * | `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`          |
+		 * | `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid] |
+		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                |
```
sarah11918, line 1593:
```suggestion
		 * The following table shows which route builds certain page URLs when file-based routes, injected routes, and redirects are combined as shown below:
		 * - File-based route: `/blog/post/[pid]`
		 * - File-based route: `/[page]`
		 * - Injected route: `/blog/[...slug]`
		 * - Redirect: `/blog/tags/[tag]` -> `/[tag]`
		 * - Redirect: `/posts` -> `/blog`
		 *
		 * URLs are handled by the following routes:
		 *
		 * | Page               | Current Behavior                 | Global Routing Priority Behavior   |
		 * | ------------------ | -------------------------------- | ---------------------------------- |
		 * | `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`          |
		 * | `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid]` |
		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                |
```

.changeset/short-keys-bow.md
```diff
@@ -0,0 +1,16 @@
...
+Updates [Astro's routing priority rules](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) to prioritize the most specifically-defined routes. 
+
+Now, routes with **more defined path segments** will take precedence over less-specific routes.
+
+For example, `/blog/posts/[pid].astro` (3 path segments) takes precedence over `/blog/[...slug].astro` (2 path segments). This means that:
+
+- `/pages/blog/posts/[id].astro` will build routes of the form `/blog/posts/1` and `/blog/posts/a`
+- `/pages/blog/[...slug].astro` will build routes of a variety of forms, including `blog/1` and `/blog/posts/1/a`, but will not build either of the previous routes.
+
+For a complete list of Astro's routing priority rules, please see the [routing guide](https://docs.astro.build/en/core-concepts/routing/#route-priority-order).
```
sarah11918, line 14:
```suggestion
For a complete list of Astro's routing priority rules, please see the [routing guide](https://docs.astro.build/en/core-concepts/routing/#route-priority-order). This should not be a breaking change, but you may wish to inspect your built routes to ensure that your project is unaffected.
```

.changeset/smooth-cobras-help.md
```diff
@@ -0,0 +1,33 @@
...
+- Redirect: `/posts` -> `/blog`
+
+URLs are handled by the following routes:
+
+| Page               | Current Behavior                 | Global Routing Priority Behavior    |
+|--------------------|----------------------------------|-------------------------------------|
+| `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`           |
+| `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid]` |
+| `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                 |
+
```
sarah11918, line 33:
```suggestion

In the event of route collisions, where two routes of equal route priority attempt to build the same URL, Astro will log a warning identifying the conflicting routes.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1566,6 +1566,33 @@ export interface AstroUserConfig {
...
+		 * - Redirect: `/posts` -> `/blog`
+		 *
+		 * URLs are handled by the following routes:
+		 *
+		 * | Page               | Current Behavior                 | Global Routing Priority Behavior    |
+		 * | ------------------ | -------------------------------- | ----------------------------------- |
+		 * | `/blog/tags/astro` | Injected route `/blog/[...slug]` | Redirect to `/tags/[tag]`           |
+		 * | `/blog/post/0`     | Injected route `/blog/[...slug]` | File-based route `/blog/post/[pid]` |
+		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                 |
+		 */
```
sarah11918, line 1594:
```suggestion
		 * | `/posts`           | File-based route `/[page]`       | Redirect to `/blog`                 |
		 *
		 *
		 * In the event of route collisions, where two routes of equal route priority attempt to build the same URL, Astro will log a warning identifying the conflicting routes.
		 */
```

.changeset/short-keys-bow.md
```diff
@@ -0,0 +1,16 @@
+---
+"astro": patch
+---
+
+Updates [Astro's routing priority rules](https://docs.astro.build/en/core-concepts/routing/#route-priority-order) to prioritize the most specifically-defined routes. 
+
+Now, routes with **more defined path segments** will take precedence over less-specific routes.
```
Yan-Thomas, line 7: Small nit
```suggestion
Now, routes with **more defined path segments** will take precedence over less specific routes.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1587,6 +1617,15 @@ export interface AstroUserConfig {
  */
 export type InjectedScriptStage = 'before-hydration' | 'head-inline' | 'page' | 'page-ssr';
 
+/**
+ * IDs for different priorities of injected routes and redirects:
+ * - "normal": Merge with discovered file-based project routes, behaving the same as if the route
+ *   was defined as a file in the project.
+ * - "legacy": Use the old ordering of routes. Inject routes will override any file-based project route,
+ *   and redirects will be overriden by any project route on conflict.
```
Yan-Thomas, line 1625: typo: `overriden` -> `overridden`
```suggestion
 *   and redirects will be overridden by any project route on conflict.
```

Yan-Thomas (commented):
Found two small nits, overall the docs LGTM.

sarah11918 (approved):
Giving my final, official approval, while *still* pointing out that since this is a minor release *anyway*, maybe `minor` makes more sense than `patch` if we think this could change how some people's projects behave?

nowaythatworked:
> > Do you think it can close #6221 ?
> 
> Seems like the same problem to me.
> 
> It would require change on the integration registering the catch-all route to use a lower priority than the default since this keeps the current behavior by default

Hi, sorry for commenting so late. I created the mentioned issue, did a lot of other work on the integration and am now coming back to this.
Can you elaborate on what I'd need to change? I don't see an option to influence the priority when calling injectRoute. I'm using the current astro version but so far, the issue is still there.

PR #9441, opened 2023-12-15
feat: Expose markdoc config to astro components

Thodor12:
Use case: Astro components currently are unable to access the markdoc config. Normally you do this in the transform function, but the transform function + astro component does not really work. Therefore having access to the config as props to the component is useful in case you want access to variables (or other parts of the config).

## Changes

- Modifies the markdoc renderer to include the markdoc config to the tree nodes, which allows astro components to obtain the markdoc config as props.

## Testing

- Tested in the "with-markdoc" example by checking if the config variable was not undefined in the props.

## Docs

N/A, if people do or don't define the config as part of the props has no impact, if it's defined the user will have access to it, if not it will still exist but just not be shown.


bholmesdev:
Hey @Thodor12! Thanks for the PR. I'll admit I'm not a fan of magic props that we apply to components, since this can cause naming collisions and doesn't _fully_ address the problem. Example: what if you need to access the markdoc config outside of an embedded component? Maybe at the page level to render external markdoc files?

I have a couple (unexplored) ideas on how to retrieve the config from components:
1. **Use a Markdoc transform to pass `config` as a prop to your component.** This is a strategy we use today [to pass properties to custom headings](https://docs.astro.build/en/guides/integrations-guide/markdoc/#custom-headings) like the generated slug. You can explore this yourself using a Markdoc transform ([reference to our headings plugin](https://github.com/withastro/astro/blob/main/packages/integrations/markdoc/src/heading-ids.ts#L40) in case this is wrong!):

```ts
import { defineMarkdocConfig, nodes, component } from '@astrojs/markdoc/config';

  tags: {
    myComponent: {
      transform(node, config) {
	const children = node.transformChildren(config);
	// pass the `config` as props
        const props = { config };
        const render = config.tags.myComponent.render;
       
	return new Markdoc.Tag(render, props, children);
      },
      render: component('./src/components/myComponent.astro'),
    },
  },
});
```

2. **Just import the config file via ESM.** From what I understand, this _should_ work. I'd need to double check that styles and scripts load as expected though. aka `import markdocConfig from '../markdoc.config.mjs'`

Hope this helps! Lmk if either of these strategies work. Closing this issue since I believe those would be better recommended paths.

bholmesdev:
I'll also leave this open in case there's a code change to explore. Though for now, I'd encourage us to explore a different solution

Thodor12:
Now that you say it, magic props may indeed not be a good idea due to possible conflicts, did not consider that.

Just tested your first example with the transform function, never tried `const render = config.tags.myComponent.render;`, that's what did it.

Am able to implement my use-case right now, (default arguments coming from variables in case an argument was missing), thanks.

bholmesdev:
Woo! Glad that worked. May be a little niche for a docs update but if we get another request we can add that 👍

PR #9445, opened 2023-12-15
Upgrade compiler

natemoo-re:
## Changes

- Upgrades `@astrojs/compiler` to `^2.3.4`.
- In practice, this fixes an issue where multibyte characters would cause the sourcemapper to crash
- A few other `index out of range` issues were fixed as well!
- Also fixed a fun new bug that this test suite uncovered related to expressions being the final node in a file

## Testing

CI

## Docs

Bug fix only

### DIFF
--- a/.changeset/many-dogs-rest.md
+++ b/.changeset/many-dogs-rest.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


matthewp (approved):

matthewp (changes requested):
Looks like there are some test failures.

matthewp (approved):

PR #9446, opened 2023-12-15
Toggle dev toolbar hitbox height when toolbar is visible

alexnguyennz:
## Changes

Sets the `dev-bar-hitbox-above` div height to 0 when the overlay is active so it doesn't block elements in that space e.g. links from being clickable (like footer links or any links that come into that space when you're scrolling up and down).

The tooltip's top value for the tooltip then includes the fixed `42px` height of the hitbox.

### Before (hovering over Twitter link)

before-link

### Before (hitbox)

before-inspect

### After (hovering over Twitter link)

after-link

### After (hitbox)

after-inspect

### After (showing tooltip)

after-tooltip

## Testing

`pnpm run test:match "runtime"`

## Docs

N/A


alexnguyennz:
> Thank you @alexnguyennz , could you post some screenshots of **before** and **after** the change?

Yep sure, have updated the comment.

bluwy (approved):
Tested locally and works great!

packages/astro/src/runtime/client/dev-overlay/overlay.ts
```diff
@@ -492,16 +492,20 @@ export class AstroDevOverlay extends HTMLElement {
...
 			barContainer?.removeAttribute('inert');
 			devBar?.removeAttribute('tabindex');
+			if (devBarHitboxAbove) devBarHitboxAbove.style.height = '0';
 			return;
 		}
 		if (newStatus === false) {
 			this.devOverlay?.setAttribute('data-hidden', '');
 			barContainer?.setAttribute('inert', '');
 			devBar?.setAttribute('tabindex', '0');
+			if (devBarHitboxAbove) devBarHitboxAbove.style.height = '42px';
```
Princesseuh, line 508: Could we move the `42` value to a const somewhere so we can change the hitbox height in one go if needed? Especially now that there's a calculated value from it (the -38)

alexnguyennz: Yep that makes sense. I've updated things - more like that?

Princesseuh (approved):

Issue #9444, opened 2023-12-15
Middleware fails to work without showing an error


lilnasy:
### Astro Info

```block
Astro                    v4.0.5
Node                     v20.9.0
System                   Linux (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

<img width="444" alt="image" src="https://github.com/withastro/astro/assets/69170106/57fe3150-cb40-4142-82d2-f5cd6ebad4c4">
<img width="354" alt="image" src="https://github.com/withastro/astro/assets/69170106/59c0e112-4774-4e13-b90b-cbb8e116d68f">

### What's the expected result?
The top-level error gets logged somewhere.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-nwcwrn?file=src%2Fmiddleware.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #9458, opened 2023-12-18
fix: handle middleware loading error

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/9444

## Testing

I locally tested the bug using the reproduction of the issue:

Screenshot 2023-12-18 at 14 05 11

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -782,6 +782,26 @@ export const LocalsNotAnObject = {
 
 /**
  * @docs
+ * @description
+ * Thrown in development mode when middleware throws an error while attempting to loading it.
+ *
+ * For example:
+ * ```ts
+ * import {defineMiddleware} from "astro:middleware";
+ * throw new Error("Error thrown while loading the middleware.")
+ * export const onRequest = defineMiddleware(() => {
+ *   return "string"
+ * });
+ * ```
+ */
+export const MiddlewareCantBeLoaded = {
+	name: 'MiddlewareCantBeLoaded',
+	title: "Can't load the middleware.",
+	message: 'The middleware thrown an error while Astro was trying to loading it.',
+} satisfies ErrorData;
+
+/**
+ * @docs
  * @see
  * - [Images](https://docs.astro.build/en/guides/images/)
  * @description
--- a/packages/astro/src/core/middleware/loadMiddleware.ts
+++ b/packages/astro/src/core/middleware/loadMiddleware.ts
@@ -1,16 +1,22 @@
 import type { ModuleLoader } from '../module-loader/index.js';
 import { MIDDLEWARE_MODULE_ID } from './vite-plugin.js';
+import type { Logger } from '../logger/core.js';
 
 /**
  * It accepts a module loader and the astro settings, and it attempts to load the middlewares defined in the configuration.
  *
  * If not middlewares were not set, the function returns an empty array.
  */
-export async function loadMiddleware(moduleLoader: ModuleLoader) {
+export async function loadMiddleware(moduleLoader: ModuleLoader, logger: Logger) {
 	try {
 		const module = await moduleLoader.import(MIDDLEWARE_MODULE_ID);
 		return module;
-	} catch {
+	} catch (e: any) {
+		logger.error(
+			'middleware',
+			"Astro couldn't load the middleware because the module contains an error."
+		);
+		logger.error('middleware', e.stack || e.message);
 		return void 0;
 	}
 }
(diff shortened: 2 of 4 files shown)


MoustaphaDev (approved):
LGTM!

packages/astro/src/core/middleware/loadMiddleware.ts
```diff
@@ -1,16 +1,22 @@
...
  * If not middlewares were not set, the function returns an empty array.
  */
-export async function loadMiddleware(moduleLoader: ModuleLoader) {
+export async function loadMiddleware(moduleLoader: ModuleLoader, logger: Logger) {
 	try {
 		const module = await moduleLoader.import(MIDDLEWARE_MODULE_ID);
 		return module;
-	} catch {
+	} catch (e: any) {
+		logger.error(
```
Princesseuh, line 15: Is there a specific reason for this to be a log instead of an actual throw?

ematipico: No particular reason. Should we throw an astro error? What do you think?

Princesseuh: I'd say so! Presumably if the middleware fails to load, we'd like the build to fail? I'd just wonder in dev if it would work correctly? Would it crash the dev server? If so, maybe it'd be possible to handle it correctly using the method Matthew added not too long ago.

ematipico:
Updated, now we get the overlay: 

Screenshot 2023-12-18 at 17 20 29

Princesseuh (approved):
Optimally, we should catch the error, rethrow an error (Astro couldn't load the middleware...) and put the original error in the `cause`. However, I think this is still a major improvement over silently ignoring the error, ha!

lilnasy (approved):
Much needed fix, thanks!

lilnasy:
Using cause may require bit of a touch up because it is ignored in the terminal.

Princesseuh:
> Using cause may require bit of a touch up because it is ignored in the terminal.

Hmm, it shouldn't be: https://github.com/withastro/astro/blob/047d285be1ab764bc82f88b8553b46429c37efca/packages/astro/src/core/messages.ts#L283

lilnasy:
You're right! I'm thinking about prod errors, dev and build show causes.

packages/astro/src/core/middleware/loadMiddleware.ts
```diff
@@ -8,9 +10,10 @@ import { MIDDLEWARE_MODULE_ID } from './vite-plugin.js';
...
 export async function loadMiddleware(moduleLoader: ModuleLoader) {
 	try {
-		const module = await moduleLoader.import(MIDDLEWARE_MODULE_ID);
-		return module;
-	} catch {
-		return void 0;
+		return await moduleLoader.import(MIDDLEWARE_MODULE_ID);
+	} catch (error: any) {
+		const astroError = new AstroError(MiddlewareCantBeLoaded);
+		astroError.cause = error.cause;
```
lilnasy, line 16:
```suggestion
		const astroError = new AstroError(MiddlewareCantBeLoaded, { cause: error });
```
We missed the actual error here.

ematipico: Weird, I couldn’t find an example like this

lilnasy: Actually, worth looking whether AstroError passes the cause to `super()`.

lilnasy: It does, but in a roundabout way.
```suggestion
		const astroError = new AstroError(MiddlewareCantBeLoaded, undefined, { cause: error });
```

ematipico:
@Princesseuh would it be possible to document https://github.com/withastro/astro/pull/9458#discussion_r1430685828 somehow? Or even making it type aware, it was a guess game 😅

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -780,6 +780,26 @@ export const LocalsNotAnObject = {
...
+ * throw new Error("Error thrown while loading the middleware.")
+ * export const onRequest = defineMiddleware(() => {
+ *   return "string"
+ * });
+ * ```
+ */
+export const MiddlewareCantBeLoaded = {
+	name: 'MiddlewareCantBeLoaded',
+	title: "Can't load the middleware.",
+	message: 'The middleware thrown an error while Astro was trying to loading it.',
```
sarah11918, line 800: Hi @ematipico ! There's a tiny language error here! Can you please update to whichever feels right to you?

> message: 'The middleware threw an error while Astro was trying to loading it.',

Issue #9345, opened 2023-12-06
Cannot build when using Sharp 0.33.0


Boston343:
### Astro Info

```block
Astro                    v4.0.2
Node                     v18.18.2
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             auto-import
                         @astrojs/mdx
                         @astrojs/tailwind
                         @astrojs/sitemap
                         astro-compress
```

### Describe the Bug

If I upgrade the Sharp dependency to 0.33.0 for image optimization with astro:assets, I am unable to build locally (`npm run build`). Development mode still works fine.

This happens with both Astro v3.6.4 and v4.0.2.

The error logged to the console looks like:

```bash
MissingSharp: Could not find Sharp. Please install Sharp (`sharp`) manually into your project or migrate to another image service.
    at loadSharp (file:///.../dist/chunks/astro/assets-service_x_nToqfa.mjs:352:11)
    at async Object.transform (file:///.../dist/chunks/astro/assets-service_x_nToqfa.mjs:364:15)
    at async generateImageInternal (file:///.../node_modules/astro/dist/assets/build/generate.js:119:24)
    at async generateImage (file:///.../node_modules/astro/dist/assets/build/generate.js:66:28)
    at async file:///.../node_modules/p-queue/dist/index.js:118:36 {
  loc: undefined,
  title: 'Could not find Sharp.',
  hint: "See Sharp's installation instructions for more information: https://sharp.pixelplumbing.com/install. If you are not relying on `astro:assets` to optimize, transform, or process any images, you can configure a passthrough image service instead of installing Sharp. See https://docs.astro.build/en/reference/errors/missing-sharp for more information.\n" +
    '\n' +
    'See https://docs.astro.build/en/guides/images/#default-image-service for more information on how to migrate to another image service.',
  frame: undefined,
  type: 'AstroError'
}
```

**Additional notes:** It seems to be an interaction between astro:assets and astro-compress. If I even have the line `import compress from "astro-compress"` inside `astro.config.mjs` the build will fail (I don't even have to add it to the integrations arrray). However, if I remove the compress() integration, and comment out the import statement, it will build fine. THEN I can add the astro-compress import and integration back into the integrations array and it will build and compress fine. But if I then go and add an image or otherwise cause the images to need re-generated (since they aren't cached), the issue will occur again. And then you'll have to do the comment out compress and what-not all over again.

If I downgrade sharp back to 0.32.6 it builds just fine.

[...]

PR #9463, opened 2023-12-18
fix(deps): Update Sharp optionalDep version

Princesseuh:
## Changes

I'm not too sure, but I think this is the cause of all the recents Sharp-related issues like https://github.com/withastro/astro/issues/9345

Since Sharp changed its installation method in its latest version, mixing versions can get weird and certain package managers seems to have issues handling it. I've never been able to reproduce any issues, so I can't really tell what the problem is, nonetheless I believe this should fix it!

## Testing

Can't really test if it fixes the issue or not, but it should still work.

## Docs

N/A

### DIFF
--- a/.changeset/shy-knives-unite.md
+++ b/.changeset/shy-knives-unite.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Update Sharp version to ^0.33.1
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -175,7 +175,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "^0.32.5"
+    "sharp": "^0.33.1"
   },
   "devDependencies": {
     "@astrojs/check": "^0.3.1",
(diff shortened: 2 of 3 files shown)


ematipico (approved):

bluwy (approved):

jlarmstrongiv:
@Princesseuh unfortunately, the latest version of sharp still breaks when bundling deployments, see https://github.com/vercel/ncc/issues/1153 

Would you allow the installation of an older version of sharp (specifically, at least `0.32.6`) for the time being?

Princesseuh:
> @Princesseuh unfortunately, the latest version of sharp still breaks when bundling deployments, see [vercel/ncc#1153](https://github.com/vercel/ncc/issues/1153)
> 
> Would you allow the installation of an older version of sharp (specifically, at least `0.32.6`) for the time being?

It's an optionalDep, so you should be able to install whatever version of Sharp you prefer, I think!

jlarmstrongiv:
image

It seems `sharp@0.33.1` is still being installed for me, but I have a different workaround to force `0.32.6` for the time being

PR #9464, opened 2023-12-18
Fix transitions with non-recommended headers

Fryuni:
## Changes

The syntax of the Content-Type header specified by the [HTTP Semantics RFC](https://www.rfc-editor.org/rfc/rfc9110#appendix-A) is as follows:
```
Content-Type = media-type

OWS = *( SP / HTAB )

media-type = type "/" subtype parameters

subtype = token
type = token

parameter = parameter-name "=" parameter-value
parameter-name = token
parameter-value = ( token / quoted-string )
parameters = *( OWS ";" OWS [ parameter ] )

tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
 "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
token = 1*tchar
```

The `;` that marks the start of the parameters may, even though not recommended, be preceded by any number of white spaces. The previous implementation of the `fetchHTML` function would not accept a header in that form and do full page reload.

This PR takes that into account and replaces the media type extraction from the header with an implementation that works for any spec-compliant server, even without using recommended practices.

## Testing

A new integration test was added for a page replying with the following header:
```
Content-Type: text/html ; charset=utf-8
```

## Docs

I don't think this needs any docs change

### DIFF
--- a/.changeset/khaki-ducks-give.md
+++ b/.changeset/khaki-ducks-give.md
@@ -1,0 +1,9 @@
+---
+'astro': patch
+---
+
+Fixes edge case on view transitions not working for some valid responses
+
+If a page was returned by the server with blank spaces before the `;` separator in the `Content-Type` header,
+the view transition would not work and a full page reload would occur.
+Now such scenarios will have the correct view transition.
--- a/.changeset/khaki-ducks-give.md
+++ b/.changeset/khaki-ducks-give.md
@@ -2,8 +2,4 @@
 'astro': patch
 ---
 
-Fixes edge case on view transitions not working for some valid responses
-
-If a page was returned by the server with blank spaces before the `;` separator in the `Content-Type` header,
-the view transition would not work and a full page reload would occur.
-Now such scenarios will have the correct view transition.
+Fixes an edge case with view transitions where some spec-compliant `Content-Type` headers would cause a valid HTML response to be ignored.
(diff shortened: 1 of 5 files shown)


packages/astro/src/transitions/router.ts
```diff
@@ -119,8 +119,9 @@ async function fetchHTML(
 ): Promise<null | { html: string; redirected?: string; mediaType: DOMParserSupportedType }> {
 	try {
 		const res = await fetch(href, init);
+		const contentType = res.headers.get('content-type') ?? '';
 		// drop potential charset (+ other name/value pairs) as parser needs the mediaType
-		const mediaType = res.headers.get('content-type')?.replace(/;.*$/, '');
+		const mediaType = contentType.split(';', 1)[0].trim();
```
natemoo-re, line 124: This is pedantic in the _best_ way. Thanks for contributing a fix!

.changeset/khaki-ducks-give.md
```diff
@@ -0,0 +1,9 @@
+---
+'astro': patch
+---
+
+Fixes edge case on view transitions not working for some valid responses
+
+If a page was returned by the server with blank spaces before the `;` separator in the `Content-Type` header,
+the view transition would not work and a full page reload would occur.
+Now such scenarios will have the correct view transition.
```
natemoo-re, line 9: Proposing a slightly simplified changelog message! If anyone is interested in the nitty-gritty details, the changelog is automatically linked to this PR.

```suggestion
Fixes an edge case with view transitions where some spec-compliant `Content-Type` headers would cause a valid HTML response to be ignored.
```

natemoo-re (approved):

martrapp (approved):
I read about it in your article ;-) 
Many thanks for this improvement!

PR #3744, opened 2022-06-28
Fix: prevent sourcemap warnings in monorepo "dev" mode

bholmesdev:
## Changes

- Stop generating sourcemaps when building the monorepo in dev mode - prevents Vite console warnings for missing sourcemap files
- Resolves #3403 

## Testing

N/A

## Docs

N/A

PR #9465, opened 2023-12-18
Enable source map on dev builds

Fryuni:
## Changes

Debugging and following errors on Astro internals when contributing currently requires looking at transpiled JS and manually finding the original TS code. Errors are reported with their transpiled positions and debug points only work on transpiled JS.

This PR just enables sourcemaps for the dev builds, allowing for debuggers (in IDEs or using Chrome's debug tools) to refer to the original TS files when stepping through code and also for Errors to report the position in the original code (so long as the option `--enable-source-maps` is passed to Node)

How an error looks without source maps:
```
Error: Example error
    at file:///home/lotus/IsoWorkspaces/OSS/astro/astro/packages/astro/dist/core/routing/manifest/create.js:325:11
    at createRouteManifest (file:///home/lotus/IsoWorkspaces/OSS/astro/astro/packages/astro/dist/core/routing/manifest/create.js:300:45)
    at Context.<anonymous> (file:///home/lotus/IsoWorkspaces/OSS/astro/astro/packages/astro/test/units/routing/manifest.test.js:49:20)
```

How it looks with source maps:
```
Error: Example error
    at <anonymous> (src/core/routing/manifest/create.ts:469:9)
    at createRouteManifest (src/core/routing/manifest/create.ts:430:44)
    at Context.<anonymous> (file:///home/lotus/IsoWorkspaces/OSS/astro/astro/packages/astro/test/units/routing/manifest.test.js:49:20)

```

Debugging without source maps, debug points have to be on JS files without comments:
image

Debugging with source maps, debug points can be on the TS file with all the context around it:
image

## Testing

This doesn't change any behavior, so also no test changed.

## Docs

Same for docs.

### DIFF
--- a/scripts/cmd/build.js
+++ b/scripts/cmd/build.js
@@ -108,6 +108,7 @@ export default async function build(...args) {
 		entryPoints,
 		outdir,
 		format,
+		sourcemap: 'linked',
 		plugins: [
 			rebuildPlugin,
 			svelte({ isDev }),


Fryuni:
Enabling it for the release builds could be beneficial if we made the astro script added to `node_modules/.bin/astro` inject that `--enable-source-maps` options. Errors from users would become much cleaner.

But I didn't do that here

natemoo-re:
I knew we had this at one point! Looks like it got disabled due to some overly aggressive warnings from Vite since the `src` files are not installed along with the package. I'm not sure if we're doing something wrong, but there's probably a way around it. I'll ask around. https://github.com/withastro/astro/pull/3744

Fryuni:
Packaging would need to change to have it enabled on release builds without warnings, that is true.

But I think this might be a confusion between two dev modes. The warnings mention in the issue liked in that PR, which I have also seen, are on dev mode for an astro project (`astro dev`) when the package is built with source maps and doesn't include the TS files.

This PR is for dev mode of Astro itself, (`pnpm dev` while working on the astro package). And since it is for local development if you link the package into a project it will have the original files, so no warnings.

ematipico:
What's the impact in terms of performance? Especially after a change. Do you see any visible regression?

Fryuni:
> What's the impact in terms of performance? Especially after a change. Do you see any visible regression?

Nothing noticeable in my notebook with a `Intel(R) Core(TM) i7-9750H CPU`.

Rebuilds everything faster than I can click to run a test again

natemoo-re:
> But I think there might be a confusion between two dev modes. The warnings mention in the issue liked in that PR, which I have also seen, are on dev mode for an astro project (`astro dev`) when the package is built with source maps and doesn't include the TS files.
> 
> This PR is for dev mode of Astro itself, (`pnpm dev` while working on the astro package). And since it is for local development if you link the package into a project it will have the original files, so no warnings.

Ah, that is true, but when I tried to run one of the examples in the monorepo with this change, I saw the same warnings. This might be more of a Vite problem than an Astro problem, it's hard to tell.

bluwy:
> Ah, that is true, but when I tried to run one of the examples in the monorepo with this change, I saw the same warnings. This might be more of a Vite problem than an Astro problem, it's hard to tell.

Is there a way to replicate this? I tested locally and it seems to work fine. I'm +1 for this PR as well.

About the sourcemaps in prod, I think we should keep it without for now so we don't bloat up the npm bundle. Plus the transpiled code is usually easily mapped to the raw source code, so it's easy to find where it exists later in the repo.

ematipico (approved):

PR #9466, opened 2023-12-19
Respect forms with enctype set for view transitions

### LINKED ISSUE #9447: ViewTransitions form submission behavior does not match browser-native behavior
### Astro Info

```block
Astro                    v4.0.6
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When submitting a form with `method="POST"` with the `<ViewTransitions />` component, Astro submits the form as `Content-Type: multipart/form-data`, whereas the browser default behavior without JavaScript is to submit as `Content-Type: application/x-www-form-urlencoded`.

In the linked repo, I have the server log out `await Astro.request.text()` on the server side. It looks like this:

image

When you disable JavaScript, or just do this same thing on a non-astro project, the log looks like this:

image

See also the request header with JavaScript enabled:

image

Vs JavaScript disabled:

image

The `ViewTransitions` progressive enhancement should submit forms as `Content-Type: application/x-www-form-urlencoded`.

Here is where Remix handles this in their codebase: https://github.com/remix-run/remix/blob/e26ddc724e78b9cf48110cd4f854516a7f1a87a2/packages/remix-react/data.ts#L95

Note that it might be hard to disable JavaScript while using StackBlitz. The easiest way to reproduce this bug is to clone the repo locally: https://github.com/knpwrs/astro-view-transition-post-bug

https://stackblitz.com/~/github.com/knpwrs/astro-view-transition-post-bug

### Participation

knpwrs:
## Changes

This makes progressively-enhanced forms (via `ViewTransitions`) respect the `enctype` attribute.

Closes #9447.

## Testing

I built the repo and added a form to the view transitions example and tried submitting the form with both `enctype` set and not set.

## Docs

I will submit a PR to update the docs.

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -117,7 +117,18 @@ const { fallback = 'animate' } = Astro.props;
 				url.search = params.toString();
 				action = url.toString();
 			} else {
-				options.formData = formData;
+				// Form elements without enctype explicitly set default to application/x-www-form-urlencoded.
+				// In order to maintain compatibility with Astro 4.x, we need to check the value of enctype
+				// on the attributes property rather than accessing .enctype directly. Astro 5.x may
+				// introduce defaulting to application/x-www-form-urlencoded as a breaking change, and then
+				// we can access .enctype directly.
+				//
+				// Note: getNamedItem can return null in real life, even if TypeScript doesn't think so.
+				const enctype = form.attributes.getNamedItem('enctype')?.value;
+				options.body =
+					enctype === 'application/x-www-form-urlencoded'
+						? new URLSearchParams(formData as any)
+						: formData;
 			}
 
 			ev.preventDefault();
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -449,7 +449,7 @@ async function transition(
 		navigationType,
 		options.sourceElement,
 		options.info,
-		options.formData,
+		options.body,
 		defaultLoader
 	);
 	if (prepEvent.defaultPrevented) {
@@ -460,9 +460,9 @@ async function transition(
 	async function defaultLoader(preparationEvent: TransitionBeforePreparationEvent) {
 		const href = preparationEvent.to.href;
 		const init: RequestInit = {};
-		if (preparationEvent.formData) {
+		if (preparationEvent.body) {
 			init.method = 'POST';
-			init.body = preparationEvent.formData;
+			init.body = preparationEvent.body;
 		}
 		const response = await fetchHTML(href, init);
 		// If there is a problem fetching the new page, just do an MPA navigation to it.
@@ -488,7 +488,7 @@ async function transition(
 		// Unless this was a form submission, in which case we do not want to trigger another mutation.
 		if (
 			!preparationEvent.newDocument.querySelector('[name="astro-view-transitions-enabled"]') &&
-			!preparationEvent.formData
+			!preparationEvent.body
 		) {
 			preparationEvent.preventDefault();
 			return;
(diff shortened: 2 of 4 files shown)


martrapp (changes requested):
Hi @knpwrs, thanks for your quick PR!

While it solves the original problem and maintains compatibility for the encoding behavior, it still leads to incompatibilities in the client-side router and transition events.
   
Could you please try to base your solution on the existing `sourceElement` property? I have left details in the comments in the code. That way you could access the form directly or via the event's handler from the `defaultLoader()`. If this works, no changes to the `navigate()` options or the events would be required.

packages/astro/components/ViewTransitions.astro
```diff
@@ -117,7 +117,18 @@ const { fallback = 'animate' } = Astro.props;
...
+				// on the attributes property rather than accessing .enctype directly. Astro 5.x may
+				// introduce defaulting to application/x-www-form-urlencoded as a breaking change, and then
+				// we can access .enctype directly.
+				//
+				// Note: getNamedItem can return null in real life, even if TypeScript doesn't think so.
+				const enctype = form.attributes.getNamedItem('enctype')?.value;
+				options.body =
+					enctype === 'application/x-www-form-urlencoded'
+						? new URLSearchParams(formData as any)
+						: formData;
```
martrapp, line 131: Might be possibl to push this down into `defaultLoader()` in router.ts

knpwrs:
@martrapp thank you for the feedback, I think these changes should address what you've said.

martrapp:
Great! Much more concise! Were you able to test it? Does it sill work?
I guess, `sourceElement` is not always the `<form>` element but might also be a "submitter" `<button>` or similar. In that case we first need to identify the form.

knpwrs:
The latest commit is tested and working with both forms and submitters.

packages/astro/src/transitions/router.ts
```diff
@@ -462,7 +462,25 @@ async function transition(
...
+			//
+			// Note: getNamedItem can return null in real life, even if TypeScript doesn't think so, hence
+			// the ?.
+			const form =
+				preparationEvent.sourceElement instanceof HTMLFormElement
+					? preparationEvent.sourceElement
+					: preparationEvent.sourceElement instanceof HTMLElement &&
+					    'form' in preparationEvent.sourceElement
+					  ? (preparationEvent.sourceElement.form as HTMLFormElement)
+					  : preparationEvent.sourceElement?.closest('form');
```
knpwrs, line 479: I'm open to better ideas here than `'form' in element`. I wanted to be able to use `.form` and not rely on `closest('form')` since people can set the `form` attribute. This ternary falls back to `closest('form')` though I can't think of a scenario off the top of my head where that would happen.

martrapp:
So far everything looks good and it was very pleasant and productive how you solved the problem! I'm very happy with where we are now. As you wrote, we can simply fall back to the standard of form.enctype for 5.0. I like to quote a dear colleague: definitely much "better than what we had before".

I don't want to overshoot the mark, but we need additional e2e tests now. Since we don't have a long history together yet, I'll ask very carefully: do you want to add them or are you satisfied with what we've achieved so far ;-)

packages/astro/src/transitions/router.ts
martrapp: I think that is really good!

martrapp (approved):

knpwrs:
I'm open to giving it a go tonight. It'll be a good opportunity to try out playwright test and I can see there's some prior-art for view transitions that I can model after.

.changeset/rude-geckos-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Respect forms with enctype set for view transitions
```
natemoo-re, line 5:
```suggestion
Updates view transitions behavior with forms to respect `enctype`
```

natemoo-re (approved):
Works for me! Would be great to have tests, as @martrapp suggested!

knpwrs:
Tests have been pushed!

martrapp:
Very good solution! Thank you again, it was a pleasure!

PR #9476, opened 2023-12-19
Adds automatic, seasonally appropriate messages from Houston

ElianCodes:
## Changes

- Depends on https://github.com/withastro/cli-kit/pull/29
- Adds a new `data/seasonal.ts` file that implements seasonally appropriate messages and decorations
- Automates Houston's decorations (hats, ties) based on the local date / time of year

## Testing

Tested locally and regressions are covered by the existing tests!

## Docs

No docs


packages/create-astro/src/data/festive.ts
```diff
@@ -0,0 +1,66 @@
+import { random } from '@astrojs/cli-kit/utils';
+
+export default function getFestiveHouston(fancy?: boolean) {
+	const date = new Date();
+	if (date.getMonth() === 11) {
+		// Christmas season
+		return {
+			clothes: {
+				hat: random(['🎁', '🎄', '🌲']),
+				tie: '🧣'
```
natemoo-re, line 10: Technically these should all respect the `fancy` option, which is a good sign that we don't actually need to do that logic in this function. We can do it in the `actions/context.ts` file after we call `getFestiveHouston`/`getSeasonalHouston`.

natemoo-re (commented):
This is really fun! I'm glad we won't have to manually manage these messages anymore.

A few suggested refactors to organize things a bit better.

packages/create-astro/src/data/seasonal.ts
```diff
@@ -0,0 +1,69 @@
+import { random } from '@astrojs/cli-kit/utils';
+
+export default function getSeasonalHouston(fancy?: boolean) {
+	const date = new Date();
+	if ((date.getMonth() === 11 && date.getDate() === 21) || (date.getMonth() === 0 && date.getDate() === 1)) {
```
ElianCodes, line 5:
```suggestion
	if (date.getMonth() === 0 && date.getDate() === 1) {
```

natemoo-re (approved):
Looks great, thanks for implementing this!

MoustaphaDev (approved):

PR #9478, opened 2023-12-19
chore: improve type fidelity for internal error class

lilnasy:
## Changes

- Add types for the parameters that AstroError passes onto built-in Error
- Fix misuse of AstroError that the types exposed

## Testing

None as this is a maintainer QOL change mostly.

## Docs

Does not affect usage.

### DIFF
--- a/packages/astro/src/assets/build/generate.ts
+++ b/packages/astro/src/assets/build/generate.ts
@@ -226,7 +226,6 @@ export async function generateImagesForPath(
 					...AstroErrorData.CouldNotTransformImage,
 					message: AstroErrorData.CouldNotTransformImage.message(originalFilePath),
 				},
-				undefined,
 				{ cause: e }
 			);
 
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -492,7 +492,7 @@ export async function getEntrySlug({
 		contents = await fs.promises.readFile(fileUrl, 'utf-8');
 	} catch (e) {
 		// File contents should exist. Raise unexpected error as "unknown" if not.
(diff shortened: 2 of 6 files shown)


packages/astro/src/content/utils.ts
```diff
@@ -492,7 +492,7 @@ export async function getEntrySlug({
 		contents = await fs.promises.readFile(fileUrl, 'utf-8');
 	} catch (e) {
 		// File contents should exist. Raise unexpected error as "unknown" if not.
-		throw new AstroError(AstroErrorData.UnknownContentCollectionError, { cause: e });
```
ematipico, line 495: I still don't understand why the second argument is `undefined` 🤔

ematipico (approved):

packages/astro/src/content/utils.ts
lilnasy: That is the message.
```ts
const error = new Error("error message", { cause })
```
AstroError's constructor overrides the message (`this.message = errorData.message`) anyway so it might be possible to remove the option altogether.

lilnasy: I was a bit worried but it was relatively simple.

Princesseuh (approved):

packages/astro/src/assets/build/generate.ts
```diff
@@ -226,7 +226,6 @@ export async function generateImagesForPath(
 					...AstroErrorData.CouldNotTransformImage,
 					message: AstroErrorData.CouldNotTransformImage.message(originalFilePath),
 				},
-				undefined,
```
lilnasy, line 229: how did tsc miss this?

lilnasy: oh its from 9352

Issue #9456, opened 2023-12-07
`a11y-label-has-associated-control` audit produces false positives


at-the-vr:
### What version of `starlight` are you using?

0.15.0

### What version of `astro` are you using?

4.0.1

### What package manager are you using?

pnpm

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

According to `Dev Toolbar Audit` the labels used for `ThemeSelect`& `LanguageSelect` are missing its controlled text
Complete Error according to Dev Toolbar ⬇
```js
"label" tag should have an associated control and a text content.`
The label tag must be associated with a control using either for or having a nested input. Additionally, the label tag must have text content.
```
Link to component found from Audit: [Select Component](https://github.com/withastro/starlight/blob/main/packages/starlight/components/Select.astro)

Screenshots ⬇️

![image](https://github.com/withastro/starlight/assets/88548999/bc9a2d0b-4cea-4c0d-8525-dffd8fd99239)
![image](https://github.com/withastro/starlight/assets/88548999/0e7d3345-b00d-43ca-87d9-16429438a6fa)

> Note: The Stackblitz instance for [Starlight](https://astro.new/starlight-basics?on=stackblitz) does not highlights this issue but any local starlight repo will show this issue

### Link to Minimal Reproducible Example

Not applicable

### Participation

- [ ] I am willing to submit a pull request for this issue.

delucis:
Quick summary of why this was moved from the Starlight repo:

The `a11y-label-has-associated-control` audit produces warnings on Starlight’s `<select>` components, which are wrapped in a `<label>` because it requires labels to always include an `<input>` element, even though other element types are supported inside `<label>`.

See @HiDeoo’s [comment](https://github.com/withastro/astro/issues/9456#issuecomment-1859284148) for links to a similar ESLint rule, which accounts for this.

Issue #9422, opened 2023-12-13
Astro Dev Toolbar Audit: False positive on anchors


pauchiner:
### Astro Info

```block
Astro                    v4.0.4
Node                     v21.1.0
System                   macOS (arm64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/preact
```

### Describe the Bug

In the minimal reproducible example, i have two anchors, within these anchors there is an svg, in the first case the indentation of the html itself generates a false positive in the audit panel of the dev toolbar:

<img width="408" alt="Captura de pantalla 2023-12-13 a las 14 46 10" src="https://github.com/withastro/astro/assets/49660697/ce46f289-5db6-4741-99f9-a5499a1096f6">

## Element in red
``` html
<a aria-label="Astro" href="https://astro.build" target="_blank"><svg width="48" height="49" viewBox="0 0 48 49" fill="none">...</svg></a>
```
## Element in green
``` html
<a aria-label="Astro" href="https://astro.build" target="_blank"><svg width="48" height="49" viewBox="0 0 48 49" fill="none">...</svg> </a>
```

The only thing that changes between this anchors is the space at the end of the anchor tag ` </a>`

I also check lighthouse and gives a `10/10` in accessibility.

### What's the expected result?

The expected result is that both elements have the same audit, not conditioned by the parser.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-dqtpmf?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #9483, opened 2023-12-19
fix(a11y): Fix some audit false positives

Princesseuh:
## Changes

Some small false positives. Additionally rewrote a message to have the same form as the other messages.

Fix https://github.com/withastro/astro/issues/9456
Fix https://github.com/withastro/astro/issues/9422

## Testing

Tested manually!

## Docs

N/A


packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -280,14 +282,20 @@ export const a11y: AuditRuleWithSelector[] = [
...
-			const inputChild = element.querySelector('input');
-			if (!inputChild?.textContent) return true;
+			// Label must be associated with a control
+			const hasFor = element.hasAttribute('for');
+			const nestedFormElement = element.querySelector(`${formElements.join(', ')}`);
+			if (!hasFor && !nestedFormElement) return true;
+
+			// Label must have text content
+			const innerText = (element as HTMLLabelElement).innerText.trim();
+			if (innerText === undefined || innerText.length === 0) return true;
```
Princesseuh, line 298: This is not perfect, because some form elements might have text that are not actually labels and the audit can't tell that that label element is then potentially unnecessary (though there's no actual a11y issue, I think)

It's still an improvement, though!

natemoo-re: The check is fine with me, but the logic seems wrong? If we're doing `.trim()` we already have a string value ad don't need to check for `undefined`.

```suggestion
			// Label must have text content
			const innerText = (element as HTMLLabelElement).innerText.trim();
			if (innerText === '') return true;
```

natemoo-re: I'm also a bit confused if `innerText` was an intentional choice vs the standard `textContent`. https://kellegous.com/j/2013/02/27/innertext-vs-textcontent/

`innerText` does take layout into account, so I can see how that would be relevant here, but just wanted to double check.

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -348,7 +356,8 @@ export const a11y: AuditRuleWithSelector[] = [
 		message: 'Headings and anchors must have content to be accessible.',
 		selector: a11y_required_content.join(','),
 		match(element) {
-			if (!element.textContent) return true;
+			const innerText = (element as HTMLElement).innerText.trim();
+			if (innerText === undefined || (innerText && innerText.length === 0)) return true;
```
natemoo-re, line 360: Same comment here.

natemoo-re (commented):
Changes look mostly good, just one question on implementation

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -280,14 +282,20 @@ export const a11y: AuditRuleWithSelector[] = [
...
 		message:
-			'The `label` tag must be associated with a control using either `for` or having a nested input. Additionally, the `label` tag must have text content.',
-		selector: 'label:not([for])',
+			'The `label` element must be associated with a control using either `for` or having a nested form element. Additionally, the `label` element must have text content.',
+		selector: 'label',
 		match(element) {
-			const inputChild = element.querySelector('input');
-			if (!inputChild?.textContent) return true;
+			// Label must be associated with a control
+			const hasFor = element.hasAttribute('for');
```
delucis, line 292: Should we also be checking that the `for` attribute element actually exists? Or does another audit take care of that?

delucis (approved):
Updated label audit makes sense to me! Left a couple of small comments, but LGTM 🙌

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
Princesseuh: Yes, `innerText` is an intentional choice, I figured that for accessibility it was better since it takes into account CSS and stuff

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
Princesseuh: Not sure I understand what you mean. The line you highlighted checks for `for`?

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
natemoo-re: Cool, that's valid. A comment to that effect would be helpful!

My original suggestion about the conditional logic still stands. `innerText` (seemingly) won't ever be `undefined`.

natemoo-re (approved):

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
Princesseuh: Oh, I just got it, that the element the for is for exists, that's neat. We could do another rule for that, this rule is already a bit overloaded!

PR #9490, opened 2023-12-20
Ensure `appEntrypoint` is referenced in Vue components

### LINKED ISSUE #6827: Astro doesn't bundle imports from Vue appEntrypoint during build

2.2.2

None

yarn

Mac

Chrome

### Describe the Bug

This is a strange issue and took me a long time to track down. The issue seems to occur when "globally" importing CSS and global Vue components in a custom `appEntrypoint`. The bundler does not include the CSS files (maybe JS files too, I'm not sure) from these things, but _only when running `build`_. When running `dev`, it behaves as I would expect.

**Minimal Reproducible Example**:

[astro-bundle-issue.zip](https://github.com/withastro/astro/files/11214839/astro-bundle-issue.zip)
(Sorry for not making a CodeSandbox, but I tend to delete those later. And this was small enough to attach. This will be more permanent).

**Workaround**:

The "solution" is to do these global imports in my `/layouts/Layout.astro` file, as you can see commented out in the attached example. Uncomment those lines to watch the appropriate styles actually be included in the `build`.

**Expected behavior**:

I would expect imports in the `appEntrypoint` to be properly included in the resulting bundle, and not have to do all global imports in an `.astro` file. At the very least, the behavior should be consistent between `build` and `dev`, and _well documented_, as this will cause very insidious deployment bugs for people.

See attached zip file for example

### Participation

Unfortunately I would have no idea where to start debugging this in the actual source code. Probably can't dedicate the time to it anyway.

natemoo-re:
## Changes

- Closes #6827, closes PLT-1346
- Fixes bundling for styles referenced in the Vue integration's `appEntrypoint` file.
- Rather than hacking Astro's internals to treat the virtual entrypoint as a special file, this PR adds a side-effectual import to every `.vue` component. It may not be the cleanest method, but this does allow Astro to associate styles referenced by the `appEntrypoint` with pages where Vue components are rendered.
- This is probably fine to release as a patch

## Testing

New test suite added for dev and build behavior

Tested manually, too

## Docs

N/A, bug fix only

### DIFF
--- a/packages/integrations/vue/src/index.ts
+++ b/packages/integrations/vue/src/index.ts
@@ -4,6 +4,7 @@ import vue from '@vitejs/plugin-vue';
 import type { Options as VueJsxOptions } from '@vitejs/plugin-vue-jsx';
 import type { AstroIntegration, AstroRenderer } from 'astro';
 import type { Plugin, UserConfig } from 'vite';
+import { MagicString } from '@vue/compiler-sfc';
 
 interface Options extends VueOptions {
 	jsx?: boolean | VueJsxOptions;
@@ -39,6 +40,7 @@ function virtualAppEntrypoint(options?: Options): Plugin {
 
 	let isBuild: boolean;
 	let root: string;
+	let appEntrypoint: string | undefined;
 
 	return {
 		name: '@astrojs/vue/virtual-app',
@@ -47,6 +49,11 @@ function virtualAppEntrypoint(options?: Options): Plugin {
 		},
 		configResolved(config) {
 			root = config.root;
+			if (options?.appEntrypoint) {
+				appEntrypoint = options.appEntrypoint.startsWith('.')
+					? path.resolve(root, options.appEntrypoint)
+					: options.appEntrypoint;
+			}
 		},
 		resolveId(id: string) {
 			if (id == virtualModuleId) {
@@ -55,11 +62,7 @@ function virtualAppEntrypoint(options?: Options): Plugin {
 		},
 		load(id: string) {
 			if (id === resolvedVirtualModuleId) {
-				if (options?.appEntrypoint) {
-					const appEntrypoint = options.appEntrypoint.startsWith('.')
-						? path.resolve(root, options.appEntrypoint)
-						: options.appEntrypoint;
-
+				if (appEntrypoint) {
 					return `\
 import * as mod from ${JSON.stringify(appEntrypoint)};
 						
@@ -78,6 +81,20 @@ export const setup = async (app) => {
 }`;
 				}
 				return `export const setup = () => {};`;
+			}
+		},
+		// Ensure that Vue components reference appEntrypoint directly
+		// This allows Astro to assosciate global styles imported in this file
+		// with the pages they should be injected to
+		transform(code, id) {
+			if (!appEntrypoint) return;
+			if (id.endsWith('.vue')) {
+				const s = new MagicString(code);
+				s.prepend(`import "${appEntrypoint}";\n`);
+				return {
+					code: s.toString(),
+					map: s.generateMap()
+				}
 			}
 		},
 	};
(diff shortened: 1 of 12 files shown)


packages/integrations/vue/src/index.ts
```diff
@@ -80,6 +83,20 @@ export const setup = (app) => {
...
+			if (!appEntrypoint) return;
+			if (id.endsWith('.vue')) {
+				const s = new MagicString(code);
+				s.prepend(`import "${appEntrypoint}";\n`);
+				return {
+					code: s.toString(),
+					map: s.generateMap()
+				}
+			}
+		},
```
florian-lefebvre, line 99: Does this mean if there are 10 components, a css file will be imported 10 times?

lilnasy: Imported 10 times, but not included in the parent page 10 times.

florian-lefebvre: Is that some vite magic that deduplicates those? Genuinely curious

lilnasy: Yes, it's the same as if you imported A.astro and B.astro into your page, while A in turn also imported B.

vincerubinetti:
Nice! Thanks for the fix. Looked at the code but wouldn't be able to make sense of it without more study.

Just double check that it works/errors consistently between dev and prod, as I think that's the most important thing 😇

lilnasy:
!preview 9490

lilnasy:
@vincerubinetti You can help us confirm by trying this preview release.

ematipico (approved):
Looks good to me!

packages/integrations/vue/src/index.ts
```diff
@@ -47,6 +49,12 @@ function virtualAppEntrypoint(options?: Options): Plugin {
 		},
 		configResolved(config) {
 			root = config.root;
+			if (options?.appEntrypoint) {
+				appEntrypoint = options.appEntrypoint.startsWith('.')
+					? path.resolve(root, options.appEntrypoint).replace(root, '')
+					: options.appEntrypoint;
+				appEntrypoint = appEntrypoint?.replaceAll('\\', '/')
+			}
```
bluwy, line 57: Is there a reason the path resolving code is changed here? e.g. the `.replace(root, '')` part and the`replaceAll('\\', '/')` part

packages/integrations/vue/src/index.ts
```diff
@@ -80,6 +84,20 @@ export const setup = (app) => {
...
 			}
 		},
+		// Ensure that Vue components reference appEntrypoint directly
+		// This allows Astro to assosciate global styles imported in this file
+		// with the pages they should be injected to
+		transform(code, id) {
+			if (!appEntrypoint) return;
+			if (id.endsWith('.vue')) {
+				const s = new MagicString(code);
+				s.prepend(`import "${appEntrypoint}";\n`);
```
bluwy, line 94: If the change of resolving is related to here, I think it's safer to use `JSON.stringify` here instead like above so it's safer.

packages/integrations/vue/src/index.ts
```diff
@@ -80,6 +84,20 @@ export const setup = (app) => {
...
+				return {
+					code: s.toString(),
+					map: s.generateMap()
```
bluwy, line 97:
```suggestion
					map: s.generateMap({ hires: 'boundary' })
```

We can generate a higher resolution mapping here to improve the final merged sourcemap resolution.

packages/integrations/vue/src/index.ts
natemoo-re: Ah you're right, I do not need to do either of these.

packages/integrations/vue/src/index.ts
natemoo-re: Good catch!

bluwy (approved):
Looks great!

PR #9493, opened 2023-12-21
Compress banner image

altano:
## Changes

I compressed the banner image (that shows up on npm, for example) from a 2.3 MB PNG to a 161 KB JPG.

Compression was performed on https://squoosh.app/ using MozJPG @ 80% quality. I can't tell the difference between the compressed and original, even at full size.

## Testing

I don't really know how to test this short of publishing.

## Docs

No doc updates should be required.


ematipico (approved):
Thank you @altano !

PR #9497, opened 2023-12-21
qol(routing): warn when api route method doesn't match the casing of an export

### LINKED ISSUE #9496: Upgrading to v4 make tRPC api requests fail in 404 error
### Astro Info

```block
Astro                    v4.0.7
Node                     v21.2.0
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

Trying to integrate a fresh Astro 4 instalation with tRPC, following the official [docs here](https://trpc.io/docs/server/adapters/fetch#example-apps) is not working.

This issue has been also mentioned [here](https://discord.com/channels/830184174198718474/830184175176122389/1187470627356024993) and [here](https://discord.com/channels/830184174198718474/830184175176122389/1187470978058563654)

Return a successful response when accessing 
http://localhost:4321/trpc/test or any other method under the endpoint.

https://github.com/cmnstmntmn/astro-trpc

### Participation

lilnasy:
## Changes

- Closes #9496
- improve messaging for getting the case wrong

## Testing

Does not affect behavior, just messaging - existing tests should pass.

## Docs

Does not affect usage.

### DIFF
--- a/.changeset/tidy-dogs-remain.md
+++ b/.changeset/tidy-dogs-remain.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds helpful message for when the API Route export is not uppercase.
--- a/.changeset/tidy-dogs-remain.md
+++ b/.changeset/tidy-dogs-remain.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Adds helpful message for when the API Route export is not uppercase.
+Adds a helpful warning message for when an exported API Route is not uppercase.
--- a/packages/astro/src/runtime/server/endpoint.ts
+++ b/packages/astro/src/runtime/server/endpoint.ts
@@ -2,18 +2,6 @@ import { bold } from 'kleur/colors';
 import type { APIContext, EndpointHandler, Params } from '../../@types/astro.js';
 import type { Logger } from '../../core/logger/core.js';
 
-function getHandlerFromModule(mod: EndpointHandler, method: string) {
-	// If there was an exact match on `method`, return that function.
-	if (mod[method]) {
-		return mod[method];
-	}
-	if (mod['ALL']) {
-		return mod['ALL'];
-	}
-	// Otherwise, no handler found.
-	return undefined;
-}
-
 /** Renders an endpoint request to completion, returning the body. */
 export async function renderEndpoint(
 	mod: EndpointHandler,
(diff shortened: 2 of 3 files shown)


packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -23,38 +11,33 @@ export async function renderEndpoint(
...
 			)} requests are not available for a static site. Update your config to \`output: 'server'\` or \`output: 'hybrid'\` to enable.`
 		);
 	}
-	if (!handler || typeof handler !== 'function') {
+	if (typeof handler !== 'function') {
+		logger.warn(
+			'router',
+			`No API Route handler exists for the method "${method}" for the route ${url.pathname}.\n` +
+			`Found handlers: ${Object.keys(mod).map(method => JSON.stringify(method)).join(', ')}\n`
+		);
```
lilnasy, line 30: image

.changeset/tidy-dogs-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds helpful message for when the API Route export is not uppercase.
```
VoxelMC, line 5:
```suggestion
Adds a helpful warning message for when an exported API Route is not uppercase.
```
Possible rewording for the changeset, I hope it helps :)

packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -23,38 +11,34 @@ export async function renderEndpoint(
...
 			)} requests are not available for a static site. Update your config to \`output: 'server'\` or \`output: 'hybrid'\` to enable.`
 		);
 	}
-	if (!handler || typeof handler !== 'function') {
+	if (typeof handler !== 'function') {
+		logger.warn(
+			'router',
+			`No API Route handler exists for the method "${method}" for the route ${url.pathname}.\n` +
+			`Found handlers: ${Object.keys(mod).map(exp => JSON.stringify(exp)).join(', ')}\n` +
+			('all' in mod ? `One of the exported handlers is "all" (lowercase), did you mean to export 'ALL'?\n` : '')
```
VoxelMC, line 30: Just wondering why `all` is the only case to handle, and not any of the other verbs?

I would assume that, if the user reads "get" in the `Found handlers:` log, it should be enough.

VoxelMC (commented):
Hello! I have never reviewed here, so I am sorry if anything I have said is out of scope.

These changes look very helpful for someone who has just upgraded to v4.0! I especially like the `Found handlers:` log, which will be great now that there are no deprecation notices on lowercase methods anymore.

I have left a couple of comments on my opinions. I hope they are helpful :)

.changeset/tidy-dogs-remain.md
lilnasy: Much better, thanks!

packages/astro/src/runtime/server/endpoint.ts
lilnasy: The most recent user to run into this was exporting "all". I agree this message should expand to "get", "post" depending what the current request's method is.

bluwy (approved):
I like the new warnings! Looks great

packages/astro/src/runtime/server/endpoint.ts
ematipico, line 30: We **removed** the support lower-case functions in v4 (they were deprecated in v3), so we shouldn't show a warning anymore, otherwise it sounds like we still support it.

lilnasy: This message would only be shown when `pages/x.ts` existed but `/x` failed to match a handler. I think at that point, it's safe to say the user could use a helpful nudge.

ematipico: To be more precise, I talk about the `all` check. We don't support that anymore, so we shouldn't write **new** code around that. New developers would see this line and won't understand why it is there.

matthewp (approved):

PR #9501, opened 2023-12-22
feat: export JSX types from jsx-runtime

Princesseuh:
## Changes

We're migrating our language-tooling to use JSX pragmas, but this require exporting `astroHTML.JSX` as `JSX` from something named `jsx-runtime` because it's hardcoded in TypeScript, very annoying, but nonetheless we can do that!

For users on previous versions of Astro, our language server will fake this export until a majority of users are on whatever version this is released in.

## Testing

N/A. Tested manually that it works in the language-server

## Docs

N/A


packages/astro/jsx-runtime.d.ts
```diff
@@ -0,0 +1,6 @@
+// Q: Why this file?
+// A: Our language tooling needs to access the JSX types from `astro/jsx-runtime`, due to TS limitations, however we
+// can't import `astro-jsx` types inside the actual `jsx-runtime/index.js` file due to circular dependency issues.
+import './astro-jsx.js';
+export * from './dist/jsx-runtime/index.js';
+export import JSX = astroHTML.JSX;
```
Princesseuh, line 6: Yes, this is actual syntax that works.

jasikpark (commented):
pretty, horrifying code. lgtm

bluwy (approved):

ematipico (approved):

PR #9502, opened 2023-12-22
[docs] update build.format with note an adapter may set this value

sarah11918:
## Changes

Adds a short note (`This value may be set by your adapter for you.`) to the `build.format` docs.

## Testing

Not tested. In response to a [Docs issue](https://github.com/withastro/docs/issues/4807) where someone was surprised this value had been changed by installing an adapter.

## Docs

Only docs!

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -722,7 +722,7 @@ export interface AstroUserConfig {
 		 * @typeraw {('file' | 'directory')}
 		 * @default `'directory'`
 		 * @description
-		 * Control the output file format of each page.
+		 * Control the output file format of each page. This value may be set by your adapter for you.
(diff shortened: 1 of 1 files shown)


TheOtterlord (approved):
lgtm! although haven't tested yet either

packages/astro/src/@types/astro.ts
```diff
@@ -722,7 +722,7 @@ export interface AstroUserConfig {
 		 * @typeraw {('file' | 'directory')}
 		 * @default `'directory'`
 		 * @description
-		 * Control the output file format of each page.
+		 * Control the output file format of each page. This value may be set by your adapter for you.
```
sarah11918, line 725:
```suggestion
		 * Control the output file format of each page. This value may be set by an adapter for you.
```

PR #9507, opened 2023-12-23
Update a11y.ts

OliverSpeir:
Allow for aria-label and aria-labelledby in a11y-missing-content rule

## Changes

- What does this change?
  - dev overlay axe rule for required content
  - https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA14 
  - https://www.w3.org/TR/2008/REC-WCAG20-20081211/#content-structure-separation-programmatic
  - This also will align with https://github.com/dequelabs/axe-core rules

  - not sure how to handle this, I tried it but didn't seem correct hopefully someone can help

## Testing

unfamiliar with testing honestly, sorry in advance 

## Docs

docs update unnecessary


packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -358,7 +358,18 @@ export const a11y: AuditRuleWithSelector[] = [
 		match(element: HTMLElement) {
 			// innerText is used to ignore hidden text
 			const innerText = element.innerText.trim();
-			if (innerText === '') return true;
+			// Check for aria-label
+			const ariaLabel = element.getAttribute('aria-label')?.trim();
+			// Check for valid aria-labelledby
+			let ariaLabelledbyValid = false;
+			const ariaLabelledby = element.getAttribute('aria-labelledby')?.trim();
```
Princesseuh, line 365: Hmm, should this account for `aria-describedby` as well?

OliverSpeir: I don't think so. I think the distinction is that it needs specifically an accessible name, and that accessible description and accessible name are separate things. This is also how deque axe-core tests things, which is what powers Lighthouse accessibility tests.


https://www.w3.org/TR/accname/#dfn-accessible-name

OliverSpeir:
After researching deeper to answer the question about `aria-describedby` I realized that if a heading or anchor has an img inside it and that img has an alt property then that alt will become the "accessible name".

I also realized that previously it was only going to check if the element that the `aria-labelledby` property referenced existed, but what it really should be checking if any one of the IDs have valid text 

I did test this by editing my node_modules on a local repo fwiw but if someone can verify it all works that would probably be best 😅

more reference: https://www.w3.org/TR/accname/#dfn-accessible-name

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -358,7 +358,33 @@ export const a11y: AuditRuleWithSelector[] = [
...
+    	}
+			// Check for <img> with valid alt attribute as a descendant
+			let imgWithAlt = false;
+      const imgElements = element.querySelectorAll('img');
+      for (const img of imgElements) {
+          if (img.getAttribute('alt') && img.getAttribute('alt').trim() !== '') {
+              imgWithAlt = true;
+              break;
+          }
+      }
```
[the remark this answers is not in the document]
OliverSpeir: Yes it should 😅 , updated.

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -358,7 +358,47 @@ export const a11y: AuditRuleWithSelector[] = [
...
+				}
+			}
+			// Return true if innerText, ariaLabel, ariaLabelledby, imgWithAlt and svgWithTitle are all empty or invalid
+			return (
+				innerText === '' &&
+				(!ariaLabel || ariaLabel === '') &&
+				!ariaLabelledbyValid &&
+				!imgWithAlt &&
+				!svgWithTitle
+			);
```
fflaten, line 401: Thanks for including svg title 👏 

I'd suggest returning early in each check. Unnecessary processing for most elements. Personally I find this larger return statement with negates harder to read.

OliverSpeir: That's a good point thank you

OliverSpeir:
Tried to improve readability, as well as return early as @fflaten suggested.

Also changed the error message.

Issue #9508, opened 2023-12-24
Svelte island does not expect unknown props, causes warning


nikhilk:
### Astro Info

```block
Astro                    v4.0.7
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/svelte
                         @astrojs/tailwind
```

### Describe the Bug

I have a very simple counter svelte component included into an Astro component.

When viewing the page after running "astro dev", I see the following warning in the console.

```
<Counter> was created with unknown prop 'data-astro-cid-j7pv25f6'
```

I see this is a warning generated from the stackblitz sample for Svelte as well (linked as reproducible example)

### What's the expected result?

Ideally this warning should not appear, as it adds noise to the console. With a warning per component, this can easily make the experience poor.

### Link to Minimal Reproducible Example

https://stackblitz.com/github/withastro/astro/tree/latest/examples/framework-svelte?file=src%2Fcomponents%2FCounter.svelte

### Participation

- [x] I am willing to submit a pull request for this issue.

Indyandie:
:relieved: Phew… I'm glad I found this! I thought it had something to do with one of my dependencies. 

```
Card.svelte:82 <Card> was created with unknown prop 'data-astro-cid-bqi7seht'
```

I'm in the of upgrading to astro 4

PR #9510, opened 2023-12-24
Filter out Svelte's unknown data prop warnings

nikhilk:
## Changes

- What does this change?
Fixes https://github.com/withastro/astro/issues/9508

The existing console filter looked for class attributes. This change adds a check for warnings related to data attributes as well.

## Testing

Tested by validating the change in a simple Astro app with a Svelte counter component.

Before the fix, the console output contains the following warning:

```
<Counter> was created with unknown prop 'data-astro-cid-j7pv25f6'
```

After patching client.js in node_modules and clearing the cache:

```
rm -rf ./node_modules/.vite
```

Run the app again

```
npm run dev
```

The console output is now clear, without any warnings.

## Docs

N/A - no docs need to be updated. This only removes console warnings, which was already happening automatically.

### DIFF
--- a/packages/integrations/svelte/client.js
+++ b/packages/integrations/svelte/client.js
@@ -104,9 +104,11 @@ function finishUsingConsoleFilter() {
  */
 function filteredConsoleWarning(msg, ...rest) {
 	if (consoleFilterRefs > 0 && typeof msg === 'string') {
-		// Astro passes a `class` prop to the Svelte component, which
+		// Astro passes `class` and `data-astro-cid` props to the Svelte component, which
 		// outputs the following warning, which we can safely filter out.
-		const isKnownSvelteError = msg.endsWith("was created with unknown prop 'class'");
+		const isKnownSvelteError =
+			msg.endsWith("was created with unknown prop 'class'") ||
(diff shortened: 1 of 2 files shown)


packages/integrations/svelte/client.js
```diff
@@ -104,9 +104,15 @@ function finishUsingConsoleFilter() {
...
+		// Astro passes `class` and `data-astro-cid` props to the Svelte component, which
 		// outputs the following warning, which we can safely filter out.
-		const isKnownSvelteError = msg.endsWith("was created with unknown prop 'class'");
+
+		// NOTE: In practice data-astro-cid props have a hash suffix. Hence the use of a
+		// quoted prop name string without a closing quote.
+
+		const isKnownSvelteError =
+			msg.endsWith("was created with unknown prop 'class'") ||
+			msg.indexOf("was created with unknown prop 'data-astro-cid") > 0;
```
bluwy, line 115: Can we use a `.includes` instead of `.indexOf > 0`? It's a bit more succinct.

nikhilk: Updated to use includes.

bluwy (approved):
Thanks!

Issue #9526, opened 2023-12-27
Tailwind nesting does not work


dctalbot:
### Astro Info

```block
Astro                    v4.0.7
Node                     v18.15.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

- Unable to nest styles when using Tailwind integration. 
- It seems like there is some kind of conflict with postcss-nesting. 
- Local build works correctly but with lots of warning logs in the console (see below)
- Production build fails to apply the styles correctly, notably in my astro layout files

```
01:07:38 [WARN] [vite] [vite:css] Nested CSS was detected, but CSS nesting has not been configured correctly.
Please enable a CSS nesting plugin *before* Tailwind in your configuration.
See how here: https://tailwindcss.com/docs/using-with-preprocessors#nesting
6  |    background: red;
7  |  
8  |    h1 {
   |    ^^^^
9  |      font-style: italic;
   |  ^^^^^^^^^^^^^^^^^^^^^^^
10 |    }
   |  ^^^
01:07:38 [200] / 8ms
01:07:39 [WARN] [vite] [vite:css] Nested CSS was detected, but CSS nesting has not been configured correctly.
Please enable a CSS nesting plugin *before* Tailwind in your configuration.
See how here: https://tailwindcss.com/docs/using-with-preprocessors#nesting
6  |    background: red;
7  |  
8  |    h1 {
   |    ^^^^
9  |      font-style: italic;
   |  ^^^^^^^^^^^^^^^^^^^^^^^
10 |    }
```

### What's the expected result?

- No warnings in the console
- Production builds match behavior of local builds
- Ideally the integration "just works", but some amount of configuration is also valid as long as it's documented

### Link to Minimal Reproducible Example

https://github.com/dctalbot/astro-repro-issue-9526

### Participation

- [X] I am willing to submit a pull request for this issue.

dctalbot:
I'm aware of https://github.com/withastro/astro/issues/6202, but the suggested workaround of changing the file extension did not work. I also tried a variety of configurations with a custom `postcss.config.js/cjs/mjs` file, as well as updating the `vite.css.postcss` config option in `astro.config.mjs`.

bluwy:
I'm able to get it to work with this config:

```js
import { defineConfig } from 'astro/config'

import tailwind from '@astrojs/tailwind'
import tailwindcssNesting from 'tailwindcss/nesting'

[...]

PR #9529, opened 2023-12-27
Add tailwindcss nesting support

bluwy:
## Changes

close https://github.com/withastro/astro/issues/9526

Add `tailwindcss/nesting` support via an option the user can enable. https://tailwindcss.com/docs/using-with-preprocessors#nesting

## Testing

Added new tests

## Docs

TBA in docs repo. And there's also the added jsdoc.

### DIFF
--- a/packages/integrations/tailwind/src/index.ts
+++ b/packages/integrations/tailwind/src/index.ts
@@ -1,3 +1,4 @@
+import { fileURLToPath } from 'node:url';
 import type { AstroIntegration } from 'astro';
 import autoprefixerPlugin from 'autoprefixer';
 import tailwindPlugin from 'tailwindcss';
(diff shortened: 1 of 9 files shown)


packages/integrations/tailwind/src/index.ts
```diff
@@ -23,15 +24,22 @@ async function getPostCssConfig(
 
 async function getViteConfiguration(
 	tailwindConfigPath: string | undefined,
-	viteConfig: UserConfig
+	nesting: boolean,
```
florian-lefebvre, line 27: We're starting to have lots of options here, could it be better to have a single object instead?

matthewp: Just an internal function so not a big deal. Going to merge this in but feel free to refactor if you'd like!

matthewp (approved):

Issue #6885, opened 2023-04-21
Duplicated css import when hydrating Solid-js component in dev mode


predaytor:
### What version of `astro` are you using?

2.3.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

pnpm

### What operating system are you using?

Ubuntu (WSL)

### What browser are you using?

Chrome

### Describe the Bug

I use css-modules for shared styles in astro and solid-js components. When using hydration directives for a solid-js component, imported styles are duplicated and imported twice in the head after all previous styles, which breaks the layout for the entire page.

In the example, it's replacing astro styles due to import order.

The problem occurred only in dev mode. Also tested with SSR adapter, same behavior. Production on SSR/SSG works fine.

```
<link rel="stylesheet" href="/src/styles/container.module.css">
```
```
<style type="text/css" data-vite-dev-id="/home/predaytor/projects/astro-solid-duplicated-css/src/styles/container.module.css">
  ._container_19k93_1 {
      background-color: pink;
      inline-size: 300px;
      block-size: 300px;
      display: grid;
      place-items: center;
  }
</style>
```

Dev (hydration-off):
![dev hydration-off](https://user-images.githubusercontent.com/32335730/233717302-f25407c4-2304-4c80-8d30-16fe62bc8b53.png)

Dev (hydration-on):
![dev hydration-on](https://user-images.githubusercontent.com/32335730/233717306-80b03f9b-adae-4d4c-9136-90482d673321.png)

Production (hydration-on):
![production](https://user-images.githubusercontent.com/32335730/233717308-725ef7d3-2444-4cba-b3cc-1e5af737c908.png)

### Link to Minimal Reproducible Example

https://github.com/predaytor/astro-solid-duplicated-css

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #9531, opened 2023-12-27
Fix duplicated CSS modules inlining

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/6885

When a `.module.css` file is imported by both an Astro file and framework component, it gets inlined by Astro to prevent a FOUC. However, the way we inline it before (using a `link`) is not how Vite handles it when it's loaded in the client (via the framework component). Vite expects the CSS module to be a `style` tag inlined.

This PR uses a `style` tag instead to fix it, which potentially causes different ordering with other style tags now (since CSS modules used to be links and they're rendered separately), but since we had changed the ordering before at https://github.com/withastro/astro/pull/8877 (which I think worked fine?), I think we can tweak this a little bit more.

## Testing

Updated tests

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/css.ts
+++ b/packages/astro/src/vite-plugin-astro-server/css.ts
@@ -1,4 +1,3 @@
-import type { RuntimeMode } from '../@types/astro.js';
 import type { ModuleLoader } from '../core/module-loader/index.js';
 import { viteID } from '../core/util.js';
 import { isBuildableCSSRequest } from './util.js';
@@ -13,8 +12,7 @@ interface ImportedStyle {
 /** Given a filePath URL, crawl Vite’s module graph to find all style imports. */
 export async function getStylesForURL(
 	filePath: URL,
-	loader: ModuleLoader,
-	mode: RuntimeMode
+	loader: ModuleLoader
 ): Promise<{ urls: Set<string>; styles: ImportedStyle[] }> {
 	const importedCssUrls = new Set<string>();
 	// Map of url to injected style object. Use a `url` key to deduplicate styles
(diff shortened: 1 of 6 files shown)


packages/astro/src/vite-plugin-astro-server/css.ts
```diff
@@ -13,37 +12,48 @@ interface ImportedStyle {
...
-				});
-			} else {
-				// NOTE: We use the `url` property here. `id` would break Windows.
-				importedCssUrls.add(importedModule.url);
+			// Else try to load it
+			else {
+				const url = new URL(importedModule.url, 'http://localhost');
+				// Mark url with ?inline so Vite will return the CSS as plain string, even for CSS modules
+				url.searchParams.set('inline', '');
+				const modId = `${decodeURI(url.pathname)}${url.search}`;
```
bluwy, line 34: We used to load the CSS file without the `?inline` query, which is fine for plain CSS files. But now we need `?inline` so it also works for CSS modules. `?inline` added to CSS files is also fine and should also lead to lesser processing.

packages/astro/src/vite-plugin-astro-server/css.ts
```diff
@@ -13,37 +12,48 @@ interface ImportedStyle {
...
-				// be inlined into other modules if it happens to be in the graph.
-				continue;
+			// In dev, we inline all styles if possible
+			let css = '';
+			// If this is a plain CSS module, the default export should be a string
+			if (typeof importedModule.ssrModule?.default === 'string') {
+				css = importedModule.ssrModule.default;
 			}
-			if (
-				mode === 'development' && // only inline in development
```
bluwy, line 35: I removed this check (and the `mode` param entirely) as it seems like we only use this function in dev now.

matthewp (approved):

Issue #7374, opened 2023-06-12
Route parameter with uri-encoded percent sign character (%) causes error in SSR


jwulf0:
### What version of `astro` are you using?

2.6.3

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### What browser are you using?

Firefox (the problem is with SSR though)

### Describe the Bug

When calling a path with an **uri-encoded** route parameter which contains the `%` character, an error is rendered instead of the expected page. Please see the repo with minimal reproducable example for more details.

I would expect to be able to provide such a route parameter as long as it's properly URI-encoded. Please let me know if I am wrong to expect that.

Looking at the stack trace and following the code, I am rather sure that this is caused by `decodeURIComponent` being called with a value that has been previously decoded via `decodeURIComponent` or `decodeURI`. This causes a string containing a percent sign without two following digits to be provided to the `decodeURIComponent` function.

### Link to Minimal Reproducible Example

https://github.com/jwulf0/astro-uri-decode-bug-minimal

### Participation

- [ ] I am willing to submit a pull request for this issue.

jwulf0:
Note: Just updated the minimal reproducible example to make it "more minimal".

PR #9532, opened 2023-12-27
Prevent double uri decoding

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/7374

Prevent unnecessary decoding in certain places. I'll add comments below to explain it.

## Testing

Added a test

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/request.ts
+++ b/packages/astro/src/vite-plugin-astro-server/request.ts
@@ -37,7 +37,7 @@ export async function handleRequest({
 	if (config.trailingSlash === 'never' && !incomingRequest.url) {
 		pathname = '';
 	} else {
-		pathname = decodeURI(url.pathname);
+		pathname = url.pathname;
 	}
 
 	// Add config.base back to url before passing it to SSR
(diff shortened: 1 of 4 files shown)


packages/astro/src/vite-plugin-astro-server/request.ts
```diff
@@ -37,7 +37,7 @@ export async function handleRequest({
 	if (config.trailingSlash === 'never' && !incomingRequest.url) {
 		pathname = '';
 	} else {
-		pathname = decodeURI(url.pathname);
+		pathname = url.pathname;
```
bluwy, line 40: It doesn't look like we need to decode it here. It also doesn't work well with the line below:

https://github.com/withastro/astro/blob/cf993bc263b58502096f00d383266cd179f331af/packages/astro/src/vite-plugin-astro-server/request.ts#L43-L44

(URLs are always encoded)

Removing this `decodeURI` seems to not break things. It affects dev only. The prod SSR "app" also doesn't decode the url, so I think this makes it more consistent.

matthewp (approved):

PR #9533, opened 2023-12-27
fix(node): prevent crash on stream error

lilnasy:
## Changes

- Fix for the server crashing on a promise rejection within the stream.

## Testing

- `node packages/integrations/node/test/fixtures/errors/dist/server/entry.mjs`
- `curl -i localhost:4321/in-stream`

## Docs

Does not affect usage

### DIFF
--- a/packages/integrations/node/src/nodeMiddleware.ts
+++ b/packages/integrations/node/src/nodeMiddleware.ts
@@ -2,6 +2,7 @@ import type { NodeApp } from 'astro/app/node';
 import type { ServerResponse } from 'node:http';
 import { createOutgoingHttpHeaders } from './createOutgoingHttpHeaders.js';
 import type { ErrorHandlerParams, Options, RequestHandlerParams } from './types.js';
+import type { AstroIntegrationLogger } from 'astro';
 
 // Disable no-unused-vars to avoid breaking signature change
 export default function (app: NodeApp, mode: Options['mode']) {
@@ -29,12 +30,14 @@ export default function (app: NodeApp, mode: Options['mode']) {
 			}
 		}
 
+		const logger = app.getAdapterLogger();
+
 		try {
 			const routeData = app.match(req);
 			if (routeData) {
 				try {
 					const response = await app.render(req, { routeData, locals });
-					await writeWebResponse(app, res, response);
+					await writeWebResponse(app, res, response, logger);
 				} catch (err: unknown) {
 					if (next) {
 						next(err);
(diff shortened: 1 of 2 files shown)


packages/integrations/node/src/nodeMiddleware.ts
```diff
@@ -60,34 +62,39 @@ export default function (app: NodeApp, mode: Options['mode']) {
...
+			const reader = body.getReader();
 			res.on('close', () => {
-				reader.cancel();
+				// Cancelling the reader may reject not just because of
+				// an error in the ReadableStream's cancel callback, but
+				// also because of an error anywhere in the stream.
+				reader.cancel().catch(err => {
+					logger.error(`There was an uncaught error in the middle of the stream while rendering ${res.req.url}.`);
+					console.error(err);
+				});
```
lilnasy, line 90: the fix is here, the rest is clerical

matthewp (approved):

packages/integrations/node/src/nodeMiddleware.ts
```diff
@@ -60,34 +62,39 @@ export default function (app: NodeApp, mode: Options['mode']) {
...
 			});
 			let result = await reader.read();
 			while (!result.done) {
 				res.write(result.value);
 				result = await reader.read();
 			}
 		} catch (err: any) {
-			console.error(err?.stack || err?.message || String(err));
 			res.write('Internal server error');
```
bluwy, line 98: This could now potentially swallow errors without logging. Should keep the logging here as before?

lilnasy: That would end up logging twice, because stream errors are thrown on cancelling the reader.

bluwy: I guess I was worried that `res.on('close'` isn't always guaranteed to be called, which means `reader.cancel()` may not be called. So it feels safer to log it in this catch block instead. Maybe we can do `reader.cancel().catch(() => {})` instead since we know the error should already be logged here?

lilnasy: Wouldn't `res.end()` below this catch block gurantee that?

bluwy: True. In that case, maybe we should leave a comment here instead about why we're swallowing the error (also remove the `(err: any)` part)

packages/integrations/node/src/nodeMiddleware.ts
```diff
@@ -60,34 +62,39 @@ export default function (app: NodeApp, mode: Options['mode']) {
...
+					logger.error(`There was an uncaught error in the middle of the stream while rendering ${res.req.url}.`);
+					console.error(err);
+				});
 			});
 			let result = await reader.read();
 			while (!result.done) {
 				res.write(result.value);
 				result = await reader.read();
 			}
 		} catch (err: any) {
```
lilnasy, line 97:
```suggestion
        // the error will be logged by the "on end" callback above
		} catch {
```

bluwy (approved):

PR #9537, opened 2023-12-27
fix: handle srcset local image paths with spaces

### LINKED ISSUE #9535: Failed parsing 'srcset' attribute value since it has an unknown descriptor.
### Astro Info

```block
Astro                    v4.0.7
Package Manager          npm
Adapter                  none
                         @astrojs/sitemap
                         @astrojs/react
                         @astrojs/tailwind
```

Chrome (haven't tested others)

### Describe the Bug

When an image path name has a space in it, `srcset` doesn't work properly in chrome.

For example, when doing the following:
```astro
---
import { Image } from "astro:assets";
import image1 from "../assets/image 1.jpg";
---

<!doctype html>
<html lang="en">
  <body>
    <Image
      src={image1}
      alt="test-image"
      widths={[240, 540, 720, image1.width]}
      sizes={`(max-width: 360px) 240px, (max-width: 720px) 540px, (max-width: 1600px) 720px, ${image1.width}px`}
      loading={"eager"}
    />
  </body>
</html>
```

I get the following output in the console in chrome after the build (this prevents the correct behaviour).
image

We can see this is because the space hasn't been correctly parsed into `%20` in the HTML `img` element:
```html
<img src="/_astro/image 1.yeyrKOPY_Z6ITw9.webp" srcset="/_astro/image 1.yeyrKOPY_Zmcymq.webp 240w, /_astro/image 1.yeyrKOPY_Z2i6AKl.webp 540w, /_astro/image 1.yeyrKOPY_Zg2MlG.webp 720w, /_astro/image 1.yeyrKOPY_Z6ITw9.webp 3072w" alt="test-image" sizes="(max-width: 360px) 240px, (max-width: 720px) 540px, (max-width: 1600px) 720px, 3072px" loading="eager" width="3072" height="3560" decoding="async">
```

The build step should replace the srcset path spaces with `%20` like so:

```html
<img src="/_astro/image 1.yeyrKOPY_Z6ITw9.webp" srcset="/_astro/image%201.yeyrKOPY_Zmcymq.webp 240w, /_astro/image%201.yeyrKOPY_Z2i6AKl.webp 540w, /_astro/image%201.yeyrKOPY_Zg2MlG.webp 720w, /_astro/image%201.yeyrKOPY_Z6ITw9.webp 3072w" alt="test-image" sizes="(max-width: 360px) 240px, (max-width: 720px) 540px, (max-width: 1600px) 720px, 3072px" loading="eager" width="3072" height="3560" decoding="async">
```

This produces no errors and it works properly now in Google Chrome.

https://stackblitz.com/edit/github-mdbw9p?file=src%2Fpages%2Findex.astro

### Participation

walter9388:
## Changes

fix #9535 

`srcset` now replaces ` ` (spaces) with `%20` (see issue #9535 for details).

## Testing

Added testing which looks at the `` element after build and checks `srcset` is correct.

## Docs

N/A (bug fix)


packages/astro/src/assets/internal.ts
```diff
@@ -110,7 +110,7 @@ export async function getImage(
 		src: imageURL,
 		srcSet: {
 			values: srcSets,
-			attribute: srcSets.map((srcSet) => `${srcSet.url} ${srcSet.descriptor}`).join(', '),
+			attribute: srcSets.map((srcSet) => `${srcSet.url.replaceAll(" ","%20")} ${srcSet.descriptor}`).join(', '),
```
Princesseuh, line 113: should `encodeUri` be used instead?

walter9388: Yeah I think so. I was just thinking that it is a bit specific. I will update momentarily...

.changeset/weak-oranges-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+srcset now parses image paths with spaces correctly
```
florian-lefebvre, line 5:
```suggestion
`<Image />` srcset now parses encoded paths correctly
```

walter9388:
The following test is failing with the size being equal to 2 instead of 1:
https://github.com/withastro/astro/blob/b34bd2b2d9f85a4de397191bf19fadd3af284be6/packages/astro/test/core-image.test.js#L255-L261

I undid my changes and found the test was still failing on main. It seems to be related to when densities = 1:
https://github.com/withastro/astro/blob/b34bd2b2d9f85a4de397191bf19fadd3af284be6/packages/astro/test/fixtures/core-image/src/pages/srcset.astro#L27-L33

I don't think this has anything to do with this PR?! Any insight into why this test is suddenly failiing?

Princesseuh:
No, the test failure is valid because it checks that `srcset` and `src` are the same URL, which with this change they're not anymore since one is encoded and the other isn't. I'll take a look!

matthewp (approved):

matthewp (approved):

walter9388:
> No, the test failure is valid because it checks that `srcset` and `src` are the same URL, which with this change they're not anymore since one is encoded and the other isn't. I'll take a look!

Ah that makes sense. Thanks for all your help!

Issue #9511, opened 2023-12-24
Astro Image remote url is not working with Firebase Storage


zenshindm:
### Astro Info

```block
NPM 
Astro 4.0.6
Node 20.10
Yes, i did included image authority 
image: {
    domains: ["firebasestorage.googleapis.com", "placehold.co", "res.cloudinary.com"],
  },
```

### Describe the Bug

Astro Image remote url with firebase storage Work in npm run dev but fail in npm run build

build error:
file:///D:/webproject/astro/ecliptic-equinox/node_modules/astro/dist/assets/build/generate.js:109
        throw new Error(`An error was encountered while reading the cache file. Error: ${e}`);
              ^

Error: An error was encountered while reading the cache file. Error: TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must not include encoded \ or / characters

main issue is the generate.js does not allow these character "%2F" in firebase storage url 
eg: https://firebasestorage.googleapis.com/v0/b/hari-ecom.appspot.com/o/documents%2FProductCoverPhotos%2Fdisruptivo%20(Small).jpg?alt=media&token=529e428d-0eaf-4618-aed3-c6b9fe28b6d2

meanwhile, other remote url from placehold or cloudinary work fine 
 <Image
      src="https://placehold.co/600x400"
      alt=""
      width="500"
      height="500"
    />

Can you make it accept the firebase storage url ? thanks 
and firebase storage unable to customize url structure 

### What's the expected result?

` <Image
      src="https://firebasestorage.googleapis.com/v0/b/hari-ecom.appspot.com/o/documents%2FProductCoverPhotos%2Fdisruptivo%20(Small).jpg?alt=media&token=529e428d-0eaf-4618-aed3-c6b9fe28b6d2"
      alt="test"
      width="500"
      height="500"
    />`
    
    expected no build error

### Link to Minimal Reproducible Example

https://codesandbox.io/p/devbox/quizzical-montalcini-kxxyvm

### Participation

- [ ] I am willing to submit a pull request for this issue.

davidjss04:
Hello friend, I have the same problem, when I build the project, the upload image in Firebase shows an error.

![image](https://github.com/withastro/astro/assets/62025422/0bc60cfe-e286-4743-ac36-c1251e72a52a)

PR #9540, opened 2023-12-28
Support remote images with encoded characters

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/9511

## Testing

- Test added

## Docs

N/A, bug fix


packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -10,7 +10,7 @@ export function propsToFilename(transform: ImageTransform, hash: string) {
 		isESMImportedImage(transform.src) ? transform.src.src : transform.src
 	);
 	const ext = extname(filename);
-	filename = basename(filename, ext);
+	filename = decodeURIComponent(basename(filename, ext));
```
Princesseuh, line 13: Hmm, while this does fix the problem I'm sure, I wonder what kind of filepath does it generate? Doesn't that mean that for non-encoded slashes it would generate nested directories?

.changeset/small-snakes-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Support remote images with encoded characters
```
florian-lefebvre, line 5:
```suggestion
Fixes remote images with encoded characters
```

packages/astro/src/assets/utils/transformToPath.ts
matthewp: A non-encoded slash would be removed in the `basename()` operation.

Princesseuh: Oh yeah, that's true. Awesome then!

Princesseuh (approved):
Looks good!

Issue #8886, opened 2023-10-22
Alllow setting limitInputPixels with Sharp


orf:
### Astro Info

```block
Astro                    v3.3.2
Node                     v18.16.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         @astrojs/preact
                         @astrojs/tailwind
```

### Describe the Bug

I have a large image that I would like to embed in one of my pages.

Locally this fails with:

```
Server Error: Error: Input image exceeds pixel limit
```

Which is related to https://github.com/lovell/sharp/issues/2373

In the sharp image service here https://github.com/withastro/astro/blob/d5dc5ce04acd19d9c9b7a29c2519408f6bdc3b83/packages/astro/src/assets/services/sharp.ts#L46 there is not  away to add `limitInputPixels: false` to the sharp constructor.

### What's the expected result?

I should be able to add `limitInputPixels: false` to the sharp constructor

Note: I can't upload the gif to the stackblitz editor, so the example below has the image link embedded directly in it. This does not work - you need to copy it locally.

The full gif can be found here: https://tomforb.es/visualizing-how-s3-deletes-1-billion-objects-with-athena-and-rust/animation_5.gif

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-arpawb?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
I'd be okay with adding this to `image.service.config` somehow (maybe `sharpImageService` from `astro/config` could take a param to set a config), this is an easy enough

DerTimonius:
I would like to work on this, please 🙂

lilnasy:
I am getting a different error.
![image](https://github.com/withastro/astro/assets/69170106/0bbbdc57-caa7-4ced-996b-33454271c6ee)

Fixed by adding `jsx: preseve` to `tsconfig.json`.

lilnasy:
What do you think the API would look like?

PR #9546, opened 2023-12-29
Add limitInputPixels option for sharp image service

bluwy:
## Changes

close https://github.com/withastro/astro/issues/8886

## Testing

Tested manually

## Docs

[The documentation](https://docs.astro.build/en/guides/images/#default-image-service) doesn't quite mention how options are passed. I suppose we should add one? cc @withastro/maintainers-docs


packages/astro/src/assets/services/sharp.ts
```diff
@@ -43,7 +50,11 @@ const sharpService: LocalImageService = {
 		// TODO: Sharp has some support for SVGs, we could probably support this once Sharp is the default and only service.
 		if (transform.format === 'svg') return { data: inputBuffer, format: 'svg' };
 
-		let result = sharp(inputBuffer, { failOnError: false, pages: -1 });
+		const result = sharp(inputBuffer, {
+			failOnError: false,
+			pages: -1,
+			limitInputPixels: config.service.config.limitInputPixels,
```
Princesseuh, line 56: Does Sharp handle it properly when this is `undefined`? The default for this is `268402689`, or `true`.

bluwy: Yeah I tested it locally, and when undefined (which is default) seems to still work fine like before. Not seeing any limit errors.

Princesseuh: The default is for limit errors to appear, what users are asking for is to be able to disable this

bluwy: Yeah if the image is too big that it hits the default, this configuration still works and would error. Also checked the source code and it seems to handle `undefined` fine:

https://github.com/lovell/sharp/blob/0499f59e710f8338779f7a29fec277a7b31da98f/lib/input.js#L40

https://github.com/lovell/sharp/blob/0499f59e710f8338779f7a29fec277a7b31da98f/lib/input.js#L113-L123

https://github.com/lovell/sharp/blob/0499f59e710f8338779f7a29fec277a7b31da98f/lib/is.js#L10-L12

Princesseuh: Ah okay, perfect then. Awesome work!

Princesseuh (approved):

ematipico (approved):

.changeset/poor-apes-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+Adds `limitInputPixels` option for the Sharp image service to be passed on to Sharp.
```
sarah11918, line 5:
``````suggestion
Adds an option for the Sharp image service to allow large images to be processed. Set `limitInputPixels: false` to bypass the default image size limit:

```
// astro.mjs.config
example showing this config
```

``````

(I'm assuming this is a config-wide option set, and not e.g. per-image.)

Depending on the appropriate code sample here, the same will need to be added to the astro types file. Since sharp is the default image service, someone might not have this in their config already, but they'd need to add it in order to pass the option. So, I'm thinking somewhere here: https://docs.astro.build/en/reference/configuration-reference/#imageservice but it would have to account for the fact that someone is only adding this NOT to actually change the image service used, but rather to configure the default one. (Since, this option is JUST for Sharp, and not the other services)

Depending on whatever's decided here, for an actual docs PR, we can even get away with just linking to the config option from here: https://docs.astro.build/en/guides/images/#default-image-service and don't need to explain it. We really should have a link there anyway!  😅 

Sorry this requires a bit of figuring out, but I'll make sure I get up early tomorrow and am happy to help with it!

**Review by sarah11918 — COMMENTED** [review]
Thanks @bluwy! This one is going to require a little bit of figuring out, but I'll turn up early tomorrow in case anything doesn't feel obvious!

File: .changeset/poor-apes-cheat.md (hunk 2 above)
Comment (bluwy) [reply .changeset/poor-apes-cheat.md:5]: Thanks for the review! Yeah I have a rough idea to improve this now.

Thanks for linking me to the docs. I actually would've have documented something like:

```js
import { defineConfig, sharpImageService } from 'astro/config'

export default defineConfig({
  image: {
    service: sharpImageService({ limitInputPixels: false })
  }
})
```

But I think it's easier to keep the object format for now, so it applies more generally to other kinds of image services.

bluwy: I've updated the changesets and jsdoc. Let me know what you think!

sarah11918 (approved):
Looks great, @bluwy ! Thank you!

jlarmstrongiv:
It seems that the release notes
```
// astro.config.mjs

export default defineConfig({
  image: {
    service: {
      entrypoint: 'astro/assets/services/sharp',
      config: {
        limitInputPixels: false,
      },
    },
  },
});
```

and config

```
export interface SharpImageServiceConfig {
    /**
     * The `limitInputPixels` option passed to Sharp. See https://sharp.pixelplumbing.com/api-constructor for more information
     */
    limitInputPixels?: number;
}
```

Have different option types

bluwy:
@jlarmstrongiv you're right! I mistyped that. Will send a followup PR to fix it

Issue #8274, opened 2023-08-29
"Internal server error: URI malformed" when visiting dynamic param (SSR) with encoded symbol %


artmsv:
### Astro info

```block
❯ astro info

Astro version            v2.10.14
Package manager          npm
Platform                 linux
Architecture             x64
Adapter                  @astrojs/node
Integrations             None or couldn't determine.
```

### What browser are you using?

Chrome

### Describe the Bug

If I pass escape symbols to the page with dynamic param `[slug].astro` app will throw 500 in `astro dev`, `astro preview`, and after build with node integration 

To reproduce visit `/blog/%2F` page in stackblitz

```
3:58:03 PM [vite] Internal server error: URI malformed
      at decodeURI (<anonymous>)
      at devBaseMiddleware (file:///home/projects/github-oiqdwp-nnenjt/node_modules/astro/dist/vite-plugin-astro-server/base.js:32:22)
      at call (file:///home/projects/github-oiqdwp-nnenjt/node_modules/vite/dist/node/chunks/dep-df561101.js:50665:7)
      at next (file:///home/projects/github-oiqdwp-nnenjt/node_modules/vite/dist/node/chunks/dep-df561101.js:50609:5)
      at Function.handle (file:///home/projects/github-oiqdwp-nnenjt/node_modules/vite/dist/node/chunks/dep-df561101.js:50612:3)
      at Server.app (file:///home/projects/github-oiqdwp-nnenjt/node_modules/vite/dist/node/chunks/dep-df561101.js:50477:37)
      at EventEmitter.emit (https://githuboiqdwpnnenjt-exbq.w-credentialless.staticblitz.com/blitz.33975f1d.js:45:8318)
      at parserOnIncoming (https://githuboiqdwpnnenjt-exbq.w-credentialless.staticblitz.com/blitz.33975f1d.js:126:13762)
      at _0x27379b.parserOnHeadersComplete (https://githuboiqdwpnnenjt-exbq.w-credentialless.staticblitz.com/blitz.33975f1d.js:85:1175)
      at https://githuboiqdwpnnenjt-exbq.w-credentialless.staticblitz.com/blitz.33975f1d.js:334:392568
```

### What's the expected result?

should return 404 page.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-oiqdwp-nnenjt?file=package.json,src%2Fpages%2Fblog%2F[slug].astro,astro.config.mjs

### Participation

[...]

PR #9548, opened 2023-12-29
Fix error overlay display on malformed uri

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/8274

The issue is that Vite would signal an error [here](https://github.com/vitejs/vite/blob/56ae92c33cba6c86ab5819877c19b9ea39f7121b/packages/vite/src/node/server/middlewares/error.ts#L54-L57), which error doesn't have the `loc` property. However the error overlay expected so as it assume all errors are processed by Astro.

I fixed it by making `loc` undefined-able for now, and handling that in the error overlay.

## Testing

With the repro, I'm seeing a proper error message page now:
image

## Docs

n/a. bug fix

### DIFF
--- a/.changeset/great-beans-check.md
+++ b/.changeset/great-beans-check.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes error overlay display on URI malformed error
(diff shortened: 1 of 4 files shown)


Princesseuh (approved):
Fine with me!

PR #9558, opened 2024-01-01
Fix asynchronous `appEntrypoint` support

### LINKED ISSUE #9557: Vue `appEntrypoint` cannot be asynchronous
### Astro Info

```block
Astro                    v4.0.8
Package Manager          pnpm
Adapter                  none
Integrations             @astrojs/vue
                         astro-global
```

_No response_

### Describe the Bug

Asynchronous `setup` scripts in `appEntrypoint` do not work as expected. It seems [this is intended to work](https://github.com/withastro/astro/blob/22f42d11a4fd2e154a0c5873c4f516584e383b70/packages/integrations/vue/server.js#L24), but due to the Vite virtual module being synchronous, the `setup` script is not properly awaited. This bug appears to have been introduced in #9362.

Asynchronous `setup` scripts are executed before the Vue application is mounted, enabling async code to be run in the `appEntrypoint` file.

https://stackblitz.com/edit/github-wwumrz?file=src%2Fvue-entrypoint.ts

### Participation

michaelhthomas:
## Changes

Fixes #9557, a regression which has broken the (admittedly niche) use case of an asynchronous `appEntrypoint` function. The changes introduced in #9362 led the potentially asynchronous setup method to be wrapped in a synchronous function, making the `await setup()` in `[server,client].js` more-or-less useless. This PR makes the virtual module's wrapper method `async` as well, bringing back support for asynchronous `appEntrypoint`s.

## Testing

A simple test for this scenario was added, which leverages two components registered globally in the `appEntrypoint` script separated by a simulated asynchronous call to ensure that registration works correctly following an async call within an async entrypoint script.

## Docs

Not sure if docs will be needed as this is a pretty niche use case, but it is certainly nice to have. I'll be happy to add to the docs if it would be appropriate.

### DIFF
--- a/.changeset/ninety-buses-occur.md
+++ b/.changeset/ninety-buses-occur.md
@@ -1,5 +1,5 @@
 ---
-"@astrojs/vue": minor
+"@astrojs/vue": patch
 ---
 
 Fixes support for async `appEntrypoint`
--- a/.changeset/ninety-buses-occur.md
+++ b/.changeset/ninety-buses-occur.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/vue": minor
+---
+
+Fixes support for async `appEntrypoint`
--- a/packages/integrations/vue/src/index.ts
+++ b/packages/integrations/vue/src/index.ts
@@ -63,9 +63,9 @@ function virtualAppEntrypoint(options?: Options): Plugin {
 					return `\
 import * as mod from ${JSON.stringify(appEntrypoint)};
 						
-export const setup = (app) => {
+export const setup = async (app) => {
 	if ('default' in mod) {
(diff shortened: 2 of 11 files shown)


.changeset/ninety-buses-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/vue": minor
```
bluwy, line 2:
```suggestion
"@astrojs/vue": patch
```

bluwy (approved):
Thanks for the fix! The PR that regresses this certainly looked unintended, so I made this a patch fix instead. I don't think we need a specific documentation for this too. Sync/Async should work ootb without documenting.

PR #9562, opened 2024-01-01
`yarn add` should be used to install packages instead of `yarn install`

DET171:
`yarn install` is used to install all packages for a project, and does nor accept arguments.

## Changes
- Checks if `ctx.packageManager` is `yarn`, and if so, uses `add` instead of `install`

## Docs

Not sure if this needs to be explicitly stated in the upgrade section.

### DIFF
--- a/packages/upgrade/src/actions/install.ts
+++ b/packages/upgrade/src/actions/install.ts
@@ -123,6 +123,8 @@ async function runInstallCommand(
 	const cwd = fileURLToPath(ctx.cwd);
 	if (ctx.packageManager === 'yarn') await ensureYarnLock({ cwd });
 
+	const installCmd = ctx.packageManager === 'yarn' ? 'add' : 'install'; 
+
 	await spinner({
 		start: `Installing dependencies with ${ctx.packageManager}...`,
 		end: `Installed dependencies!`,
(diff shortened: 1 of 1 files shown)


.changeset/poor-cars-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/upgrade": patch
+---
+
+Change the command used for upgrading packages for yarn from `install` to `add`
```
florian-lefebvre, line 5:
```suggestion
Fixes packages upgrading when using yarn
```

packages/upgrade/src/actions/install.ts
```diff
@@ -123,6 +123,8 @@ async function runInstallCommand(
 	const cwd = fileURLToPath(ctx.cwd);
 	if (ctx.packageManager === 'yarn') await ensureYarnLock({ cwd });
 
+	const installCmd = ctx.packageManager === 'yarn' ? 'add' : 'install';
```
bluwy, line 126:
```suggestion
	const installCmd = ctx.packageManager === 'yarn' || ctx.packageManager === 'pnpm' ? 'add' : 'install'; 
```

While we're at it, we could also perhaps to the same for pnpm. pnpm works with `install` just fine too, but `add` is usually what they recommend instead.

.changeset/poor-cars-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/upgrade": patch
+---
+
+Fixes packages upgrading when using yarn
```
florian-lefebvre, line 5:
```suggestion
Updates the command used for installing packages with pnpm and yarn
```

florian-lefebvre (approved):

bluwy (approved):

PR #9563, opened 2024-01-01
Fix history navigation to hash fragment on Firefox when using view transitions

martrapp:
## Changes

Firefox clears the history state on same page navigation, which later breaks history navigation.
This fix prevents that.
From discord user mcoms on #support > View Transitions back/forward not working in Firefox for fragment links:
> In Firefox, on all ViewTransition fallbacks but none, going back to a fragment link (#about) from another page does not work. It stays on the current page without navigating back. For example, if I have a page with a link /#about and then I go to /faq, I press back and the URL changes to /#about but the content is still /faq. This happens until I reach that does not contain a # in the URL, or if the # was initially navigated to. It even doesn't work if the # link had data-astro-reload={true}.

SvelteKit had a [similar issue](https://github.com/sveltejs/kit/issues/8725). 
Their analysis put me on the right track.

## Testing

Tested manually on Firefox as we do not run automated e2e tests on Firefox.

## Docs

n/a bug fix.

### DIFF
--- a/.changeset/breezy-eyes-teach.md
+++ b/.changeset/breezy-eyes-teach.md
@@ -2,4 +2,7 @@
 "astro": patch
 ---
 
-Fix history navigation to hash fragment on Firefox when using view transitions
+Fixes back navigation to fragment links (e.g. `#about`) in Firefox when using view transitions
+
+Co-authored-by: Florian Lefebvre <69633530+florian-lefebvre@users.noreply.github.com>
+Co-authored-by: Sarah Rainsberger <sarah@rainsberger.ca>
--- a/.changeset/breezy-eyes-teach.md
+++ b/.changeset/breezy-eyes-teach.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes history navigation to hash targets on Firefox when using view transitions
+Fix history navigation to hash fragment on Firefox when using view transitions
--- a/.changeset/breezy-eyes-teach.md
+++ b/.changeset/breezy-eyes-teach.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes history navigation to hash targets on Firefox when using view transitions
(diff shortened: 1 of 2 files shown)


.changeset/breezy-eyes-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix history navigation to hash fragment on Firefox when using view transitions
```
florian-lefebvre, line 5:
```suggestion
Fixes history navigation to hash fragment on Firefox when using View Transitions
```

packages/astro/src/transitions/router.ts
```diff
@@ -213,7 +213,9 @@ const moveToLocation = (to: URL, from: URL, options: Options, historyState?: Sta
 			// ... what comes next is a intra-page navigation
 			// that won't reload the page but instead scroll to the fragment
 			history.scrollRestoration = 'auto';
-			location.href = to.href;
+			const savedState = history.state;
```
florian-lefebvre, line 216: I think `state` is an object right? So shouldn't we make a copy instead of a reference using `{ ...history.state }`?

.changeset/breezy-eyes-teach.md
martrapp: I thought we were deliberately using the imperative form?

packages/astro/src/transitions/router.ts
martrapp: Yes, it is an object. It looks like Firefox just clears the reference and leaves the object alone. I think it's safe to keep the reference alive and restore it after the assignment. Why would Firefox bother changing the attributes of a custom object that it doesn't even know what attributes it has?

.changeset/breezy-eyes-teach.md
florian-lefebvre: I thought the opposite 🤔 we need Sarah (on phone rn, I can't ping for some reason)

martrapp: @sarah11918 to the rescue!

bluwy (approved):

matthewp (approved):

.changeset/breezy-eyes-teach.md
sarah11918, line 5:
```suggestion
Fixes issues with navigation history in Firefox when attempting to navigate back to fragment links (e.g. `#about`) when using view transitions
```
It is indeed "fixes" (aka. "This PR.... fixes!) but I'd go a step further and say you could describe this in a tiny bit more user-friendly way! (Make it correct, but I'd go with a style like this)

PR #9566, opened 2024-01-02
Allow remark plugins to affect getImage call for .md files

OliverSpeir:
## Changes

- adds hProperties into getImage call for images that should be optimized in .md files
- current pipeline is 
  - @markdown/remark/remark-collect-images --> @markdown/remark/rehype-images --> @astro/src/vite-plugin-markdown/images.ts
- this pr adds any hProperties a user's remark plugin might have added into the `__ASTRO_IMAGE__` property in rehype-images  then extracts those and passes them to the getImage call in images.ts

## Testing

- Tested manually by adding an optimized img and a remark plugin to the @examples/blog, looked into adding a test in core-images.test.js but because it gets the html I didn't see a simple way to run a remark plugin. 

## Docs

- This shouldn't need documenting, but if accepted a guide or tutorial about creating a remark plugin to make images eager loaded or change their width could be made later.


OliverSpeir:
This has limitation of, if you're using the same image twice it will only let one set of hProps be passed 

I think it has to do with imagePaths being a set, but I just realized it and haven't thought too much about it.

the markdown-remark test is passing locally for me as well with macos node@18, I'm not sure why it would be failing here but I guess that's a problem 😳

packages/astro/src/vite-plugin-markdown/images.ts
```diff
@@ -1,34 +1,43 @@
...
+          const decodedImagePath = JSON.parse(imagePath.replace(/&#x22;/g, '"'));
+          return spreadAttributes({
+            src: imageSources[decodedImagePath.src].src,
+            ...imageSources[decodedImagePath.src].attributes,
+          });
+        });
+      });
+    }
+    
+    const html = await updateImageReferences(${JSON.stringify(html)});
```
Princesseuh, line 40:
```suggestion
	  // NOTE: This causes a top-level await to appear in the user's code, which can break very easily due to a Rollup
	  // bug and certain adapters not supporting it correctly. See: https://github.com/rollup/rollup/issues/4708
	  // Tread carefully!
    const html = await updateImageReferences(${JSON.stringify(html)});
```

Comment here is quite important, this `await` being there can break the build randomly with no error messages due to the Rollup bug.

OliverSpeir: Ah yeah sorry I didn't mean to delete

OliverSpeir:
@Princesseuh think this should add srcset if widths or densities is passed?

Princesseuh:
> @Princesseuh think this should add srcset if widths or densities is passed?

Sorry, I had actually started a review, but then I had a meeting and a network outage and completely forgot to continue 🤦‍♀️

I think that would be great, yes! I don't think it's super hard, should just be a check like the one done here: https://github.com/withastro/astro/blob/bd3f36e6aba779b3bbe645a7cfee939021da786c/packages/astro/components/Image.astro#L29-L31, this can be even shorter in here because `spreadAttributes` should normally already handle undefined arguments.

OliverSpeir:
Alright yeah you're right, I'll add it looking back not sure why I didn't do it initially

OliverSpeir:
If this gets accepted this plugin will make it easy to add these attributes https://www.npmjs.com/package/remark-imgattr

.changeset/calm-socks-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@astrojs/markdown-remark": minor
+"astro": minor
+---
+
+Allow remark plugins to add hProperties to imgs that will be included in the getImage call if they are optimized.
```
sarah11918, line 6: @OliverSpeir , could you instead describe what the user-facing benefit/feature is here? What does this actually do or unlock? Sounds like it improves using remark plugins? How so? to do what? That's the kind of changeset message that would be helpful to have here.

sarah11918 (commented):
No accompanying docs necessary here (though, we wouldn't turn down an interesting recipe that makes use of this new functionality!)

Would ask that the changeset present a user benefit, or what will be seen/experienced though!

.changeset/calm-socks-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@astrojs/markdown-remark": minor
+"astro": minor
+---
+
+Users can now create remark plugins that will be able to pass options for how images in `.md` files will be optimized
```
sarah11918, line 6:
```suggestion
Allows remark plugins to pass options specifying how images in `.md` files will be optimized
```
Love this change, @OliverSpeir ! Something like this, maybe?

sarah11918 (approved):
Docs is happy with the changeset!

Princesseuh (changes requested):
I would love to see a test in core for this, because I feel like this can quickly become a source of problem.

Here's how I would do it:
- Add a fixture in `packages/astro/test/fixtures` with a remark plugin adding attributes
- Add a test either in `core-image.test.js` or in a new file (`core-image.test.js` is getting a bit big) in `package/astro/test`
- Confirm the attributes result in different images, you can test this in dev by making sure the `_image` URLs have the proper params.

packages/markdown/remark/src/rehype-images.ts
```diff
@@ -3,17 +3,28 @@ import type { MarkdownVFile } from './types.js';
...
-				}
-			});
-		};
+					visit(tree, (node) => {
+							if (node.type !== 'element') return;
+							if (node.tagName !== 'img') return;
+
+							if (node.properties?.src) {
+									if (file.data.imagePaths?.has(node.properties.src)) {
+											const { alt, ...otherProps } = node.properties;
```
Princesseuh, line 15: You can pass the `alt` through as well, `getImage` and `Image` will handle it!

ematipico:
Hi @OliverSpeir, do you think you'll have some to address the comments?

packages/markdown/remark/src/rehype-images.ts
```diff
@@ -3,17 +3,28 @@ import type { MarkdownVFile } from './types.js';
...
+
+											// Initialize or increment occurrence count for this image
+											const index = imageOccurrenceMap.get(node.properties.src) || 0;
+											imageOccurrenceMap.set(node.properties.src, index + 1);
+
+											node.properties['__ASTRO_IMAGE_'] = JSON.stringify({ ...otherProps, index });
+											
+											Object.keys(otherProps).forEach((prop) => {
+													delete node.properties[prop];
+											});
```
Princesseuh, line 25: This is a little bit concerning, but normally it's fine because image services **should** re-output everything they received (but didn't understand) as attributes. It's non-breaking for the built-in image services at least.

In the future, we should most likely offer an option to resolve the img's `src`, but otherwise leave the `img` tag alone, just so that people who want/need full control can do so.

packages/markdown/remark/src/rehype-images.ts
OliverSpeir: Ah yeah good point will do

OliverSpeir:
> - Add a fixture in `packages/astro/test/fixtures` with a remark plugin adding attributes

I agree, I didn't realize this was possible will do

packages/markdown/remark/src/rehype-images.ts
OliverSpeir: Yeah you're right I thought the same what if it doesn't return all the random extra stuff so far it has but yeah adding in ability to opt out or use a service that wouldn't return extra stuff 

I guess this would happen in the remark-collect-images shouldOptimize function 

But you're right it's pretty ugly and weird to stuff everything inside __ASTRO_IMAGE__ and assume it comes out of getImage()

OliverSpeir:
> Hi @OliverSpeir, do you think you'll have some to address the comments? 

I can tonight after EST work hours, but if you want to get this in on a timeline I wouldn't be sad if someone else did 

I'm not unsure about adding every attribute into '__ASTRO_IMAGE__' so if there's not a time crunch I might look at that again

packages/markdown/remark/src/rehype-images.ts
Princesseuh: Yeah, ideally we'd do it in Node, but last time I tried it didn't work. Maybe something to revisit in the future. It works at the moment, and that's what's important so far, so it's ok

Princesseuh (approved):
Awesome, thank you!

PR #9567, opened 2024-01-02
Update a11y.ts

OliverSpeir:
Accidentally deleted [PR #9507](https://github.com/withastro/astro/pull/9507)

## Changes

- Aligns missing content rule with WCAG and deque/axe-core ( used by lighthouse )
- Allows for aria-label, aria-labelledby, img with alt, svg with title
- Changes Error title and Message to be more clear and provide info how to solve the error

## Testing

Manually tested with @examples/blog 

## Docs

- No doc update needed

### DIFF
--- a/.changeset/orange-trainers-learn.md
+++ b/.changeset/orange-trainers-learn.md
@@ -1,0 +1,5 @@
+---
+"astro": minor
+---
+
+Improve a11y-missing-content rule for audit feature of dev-overlay
--- a/packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
+++ b/packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
@@ -352,13 +352,45 @@ export const a11y: AuditRuleWithSelector[] = [
 	},
 	{
 		code: 'a11y-missing-content',
-		title: 'Missing content on element important for accessibility',
-		message: 'Headings and anchors must have content to be accessible.',
+		title: 'Missing content',
+		message:
+			'Headings and anchors must have an accessible name, which can come from: inner text, aria-label, aria-labelledby, an img with alt property, or an svg with a title.',
 		selector: a11y_required_content.join(','),
 		match(element: HTMLElement) {
 			// innerText is used to ignore hidden text
 			const innerText = element.innerText.trim();
-			if (innerText === '') return true;
+			if (innerText !== '') return false;
+
+			// Check for aria-label
+			const ariaLabel = element.getAttribute('aria-label')?.trim();
+			if (ariaLabel && ariaLabel !== '') return false;
+
+			// Check for valid aria-labelledby
+			const ariaLabelledby = element.getAttribute('aria-labelledby')?.trim();
+			if (ariaLabelledby) {
+				const ids = ariaLabelledby.split(' ');
+				for (const id of ids) {
+					const referencedElement = document.getElementById(id);
+					if (referencedElement && referencedElement.innerText.trim() !== '') return false;
+				}
+			}
+
+			// Check for <img> with valid alt attribute
+			const imgElements = element.querySelectorAll('img');
+			for (const img of imgElements) {
+				const altAttribute = img.getAttribute('alt');
+				if (altAttribute && altAttribute.trim() !== '') return false;
+			}
+
+			// Check for <svg> with valid title
+			const svgElements = element.querySelectorAll('svg');
+			for (const svg of svgElements) {
+				const titleText = svg.querySelector('title');
+				if (titleText && titleText.textContent && titleText.textContent.trim() !== '') return false;
+			}
+
+			// If all checks fail, return true indicating missing content
+			return true;
 		},
 	},
 	{


packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,45 @@ export const a11y: AuditRuleWithSelector[] = [
 	},
 	{
 		code: 'a11y-missing-content',
-		title: 'Missing content on element important for accessibility',
-		message: 'Headings and anchors must have content to be accessible.',
+		title: 'Missing content',
+		message:
+			'Headings and anchors must have an accessible name, which can come from: inner text, aria-label, aria-labelledby, an img with alt property, or an svg with a title.',
```
ematipico, line 357:
```suggestion
			'Headings and anchors must have an accessible name, which can come from: inner text, aria-label, aria-labelledby, an img with alt property, or an svg with a tag <title></title>.',
```

ematipico (approved):
The fix looks good to me, although it seems that the rule doesn't check for missing `alt` for `input` tag with type image, for example.

Should this be implemented in this PR or another? Here's a reference: https://biomejs.dev/linter/rules/use-alt-text/

OliverSpeir:
> The fix looks good to me, although it seems that the rule doesn't check for missing `alt` for `input` tag with type image, for example.
> 
> Should this be implemented in this PR or another? Here's a reference: https://biomejs.dev/linter/rules/use-alt-text/

Good catch didn't even think about inputs although I don't know why someone would have a <a> or <h1> input it is valid and accessible HTML and this also aligns with deque/axe-core

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,72 @@ export const a11y: AuditRuleWithSelector[] = [
...
+			if (ariaLabelledby) {
+				const ids = ariaLabelledby.split(' ');
+				for (const id of ids) {
+					const referencedElement = document.getElementById(id);
+					if (referencedElement && referencedElement.innerText.trim() !== '') return false;
+				}
+			}
+
+			// Check for <img> with valid alt attribute
+			const imgElements = element.querySelectorAll('img');
```
natemoo-re, line 379: `match` is a callback for a single `HTMLElement`, so this will run for every `img` on the page. Shouldn't this check if `element.localName === 'img'` and not use `querySelectorAll`?

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,72 @@ export const a11y: AuditRuleWithSelector[] = [
...
+			for (const img of imgElements) {
+				const altAttribute = img.getAttribute('alt');
+				if (altAttribute && altAttribute.trim() !== '') return false;
+			}
+
+			// Check for <svg> with valid title
+			const svgElements = element.querySelectorAll('svg');
```
natemoo-re, line 386: Same for `svg`

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,72 @@ export const a11y: AuditRuleWithSelector[] = [
...
+			for (const svg of svgElements) {
+				const titleText = svg.querySelector('title');
+				if (titleText && titleText.textContent && titleText.textContent.trim() !== '') return false;
+			}
+
+			const inputElements = element.querySelectorAll('input');
```
natemoo-re, line 392: Same for `input`

natemoo-re (commented):
Logic looks good other than the `querySelectorAll` parts!

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
OliverSpeir: I think it just gets the imgs that are descendants of the element, because if you have multiple images inside the `<a>` only 1 of them needs to have an alt property

OliverSpeir: but I guess all imgs should have alt property but I don't know if that's the scope of this rule?

fflaten: > `match` is a callback for a single `HTMLElement`, so this will run for every `img` on the page.

This rule is only applied to anchors and headings, so not an issue atm.

> Shouldn't this check if `element.localName === 'img'` and not use `querySelectorAll`?

It needs to inspect child elements to work for e.g. links with svgs or images only.

Princesseuh: Yeah, this looks fine to me!

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,72 @@ export const a11y: AuditRuleWithSelector[] = [
...
+			for (const input of inputElements) {
```
Princesseuh, line 393: I'm not sure I understand this input part, what kind of HTML is this testing? This rule only checks headings and anchors, so this would test inputs inside those elements?

fflaten: https://github.com/withastro/astro/pull/9567#pullrequestreview-1800115529

@ematipico do we expect image-type input inside a heading or anchor? Sounds like an image-based form button in which case it would be in a form and as a result out of scope for the selector.

OliverSpeir: Yeah it's new to me as well, 

`<a href="/"><input type="image" src="/image.png" alt="accessible name"></a>` the accessible name of the a tag will become test 

similarly for 
`<a href="/"><input type="text" title="axe name"></a>`

also applies to h1s in my testing with VoiceOver and deque/axe-core

Princesseuh: Okay, that's weird, but if it exists, we must support it

.changeset/orange-trainers-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Improve a11y-missing-content rule for audit feature of dev-overlay
```
sarah11918, line 5:
```suggestion
Improves the a11y-missing-content rule and error message for audit feature of dev-overlay. This also fixes an error where this check was falsely reporting accessibility errors.
```

sarah11918 (commented):
Just a suggestion for your consideration!

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
```diff
@@ -352,13 +352,72 @@ export const a11y: AuditRuleWithSelector[] = [
 	},
 	{
 		code: 'a11y-missing-content',
-		title: 'Missing content on element important for accessibility',
-		message: 'Headings and anchors must have content to be accessible.',
+		title: 'Missing content',
+		message:
+			'Headings and anchors must have an accessible name, which can come from: inner text, aria-label, aria-labelledby, an img with alt property, or an svg with a tag <title></title>.',
```
sarah11918, line 357: Just checking because I don't think these error messages currently get pulled into docs, so this should be fine. But for docs, we would be using back ticks for inline code formatting on all values, but especially any components like `<title>` or else docs will break.

@Princesseuh - tangentially, do we *want* these dev tool error messages available in docs somewhere? Eventually? We don't really have any "using dev toolbar" docs at all. Is this something we want/care about?

sarah11918 (approved):
Docs is happy with the changeset!

natemoo-re (approved):
Looks great, thank you!

packages/astro/src/runtime/client/dev-overlay/plugins/audit/a11y.ts
Princesseuh: Sorry, missed this message (blame GitHub notifications not working properly since a few weeks 😠). Yes, eventually I would like to have the audits in the docs, similar to the error reference.

PR #9571, opened 2024-01-02
Remove telemetry for unhandled errors

bluwy:
## Changes

The telemetry for unhandled errors are a LOT, so removing them for now. I didn't uncomment it out as I don't quite see another way we could enable it in the future.

## Testing

It's a one-line removal, so should still work like before.

## Docs

n/a. not docs related.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/error.ts
+++ b/packages/astro/src/vite-plugin-astro-server/error.ts
@@ -5,7 +5,6 @@ import type DevPipeline from './devPipeline.js';
 import { collectErrorMetadata } from '../core/errors/dev/index.js';
 import { createSafeError } from '../core/errors/index.js';
 import { formatErrorMessage } from '../core/messages.js';
-import { eventError, telemetry } from '../events/index.js';
 
 export function recordServerError(
 	loader: ModuleLoader,
@@ -23,8 +22,6 @@ export function recordServerError(
 	// This is our last line of defense regarding errors where we still might have some information about the request
 	// Our error should already be complete, but let's try to add a bit more through some guesswork
 	const errorWithMetadata = collectErrorMetadata(err, config.root);
-
-	telemetry.record(eventError({ cmd: 'dev', err: errorWithMetadata, isFatal: false }));
 
 	pipeline.logger.error(
 		null,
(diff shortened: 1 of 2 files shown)


packages/astro/src/vite-plugin-astro-server/error.ts
```diff
@@ -24,8 +23,6 @@ export function recordServerError(
 	// Our error should already be complete, but let's try to add a bit more through some guesswork
 	const errorWithMetadata = collectErrorMetadata(err, config.root);
 
-	telemetry.record(eventError({ cmd: 'dev', err: errorWithMetadata, isFatal: false }));
```
matthewp, line 27: this is going to remove telemetry for all errors that happen during a request. Do we just want to remove the unhandled rejections? if so we can check that `_err?.name === 'UnhandledRejection'`

bluwy: I didn't know we assigned a specific name to it. Yeah I think that works too 👍

matthewp (approved):

packages/astro/src/vite-plugin-astro-server/error.ts
```diff
@@ -24,7 +24,10 @@ export function recordServerError(
...
+	// Ignore unhandled rejection errors as they appear A LOT and we cannot record the amount to telemetry
+	if (errorWithMetadata.name !== AstroErrorData.UnhandledRejection.name) {
```
Princesseuh, line 28: I don't remember, do we need to be defensive against the error not having a name here?

Princesseuh (approved):
Just one comment, jsut wondering about it., but looks fine to me otherwise. Would love to figure out what issues people were hitting though, but we don't have much info to go on from, unfortunately

packages/astro/src/vite-plugin-astro-server/error.ts
matthewp: If it doesn't have a name it should follow through to telemetry, shouldn't it? We would only need to be defensive if the error was null or undefined, which seems unlikely.

Princesseuh: Oh yeah, for sure. I was just wondering if it can be `undefined` at all there or not

bluwy:
I think Fred has the data if we filter the errors by `UnhandledRejection`, and we can check the kinds of errors we're recording before.

PR #9573, opened 2024-01-02
Support --open with url string

bluwy:
## Changes

close https://github.com/withastro/astro/issues/8192

Supports passing a url string to `server.open` and `--open` to open a specific URL in the browser on dev startup.

## Testing

Tested manually with the basic example. e.g. `pnpm dev --open /test`

## Docs

Updated the JSDoc for `server.open`.

Might need to also update: https://docs.astro.build/en/reference/cli-reference/#--open

`server.open` seems to not be documented, but probably should: https://docs.astro.build/en/reference/configuration-reference/#server-options

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -153,7 +153,7 @@ export interface CLIFlags {
 	host?: string | boolean;
 	port?: number;
 	config?: string;
-	open?: boolean;
+	open?: string | boolean;
 }
 
 /**
@@ -371,11 +371,12 @@ type ServerConfig = {
 
 	/**
 	 * @name server.open
-	 * @type {boolean}
+	 * @type {string | boolean}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
 	 * Control whether the dev server should open in your browser window on startup.
+	 * A URL string can also be passed to specify the URL to open.
 	 *
 	 * ```js
 	 * {
@@ -383,7 +384,7 @@ type ServerConfig = {
 	 * }
 	 * ```
 	 */
-	open?: boolean;
+	open?: string | boolean;
 };
 
 export interface ViteUserConfig extends vite.UserConfig {
@@ -1021,11 +1022,12 @@ export interface AstroUserConfig {
 
 	/**
 	 * @name server.open
-	 * @type {boolean}
+	 * @type {string | boolean}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
 	 * Control whether the dev server should open in your browser window on startup.
+	 * A URL string can also be passed to specify the URL to open.
 	 *
 	 * ```js
 	 * {
--- a/.changeset/cyan-seals-bathe.md
+++ b/.changeset/cyan-seals-bathe.md
@@ -1,0 +1,5 @@
+---
+"astro": minor
+---
+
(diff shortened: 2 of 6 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -371,19 +371,20 @@ type ServerConfig = {
...
+	 * @type {string | boolean}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
 	 * Control whether the dev server should open in your browser window on startup.
+	 * A URL string can also be passed to specify the URL to open.
 	 *
 	 * ```js
 	 * {
 	 *   server: { open: true }
```
ematipico, line 383: I believe this documentation should be updated and provide an example of how to use the new variant string. Plus, we should explain how this works. Can a user provide any kind of string? Should a user expect an error if a string doesn't have a certain format?

I noticed that you added a new paragraph, but it goes in conflict with your testing block, where you used `/test`, which is not a full URL

bluwy: You can pass either a pathname (so the origin is based on `host`), or the full URL path here. I'll update this part to clarify it 👍

bluwy: I added some examples, but I didn't explain what would happen if you e.g. pass an invalid URL. Since the `site` option also doesn't explain it, I think it should implicitly mean that it'll fail with `new URL` if you pass an invalid URL.

ematipico (approved):

ematipico:
We should probably double check with @withastro/maintainers-docs , however the PR it's fine with me!

packages/astro/src/@types/astro.ts
```diff
@@ -371,19 +371,21 @@ type ServerConfig = {
...
 	/**
 	 * @name server.open
-	 * @type {boolean}
+	 * @type {string | boolean}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
 	 * Control whether the dev server should open in your browser window on startup.
+	 * A full URL string (e.g. "http://example.com") or a pathname (e.g. "/about") can also be passed
+	 * to specify the URL to open.
```
sarah11918, line 380:
```suggestion
	 * Controls whether the dev server should open in your browser window on startup.
	 * 
	 * Pass a full URL string (e.g. "http://example.com") or a pathname (e.g. "/about") to specify the URL to open.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1020,16 +1022,19 @@ export interface AstroUserConfig {
...
+	 * @docs
 	 * @name server.open
-	 * @type {boolean}
+	 * @type {string | boolean}
 	 * @default `false`
 	 * @version 2.1.8
 	 * @description
 	 * Control whether the dev server should open in your browser window on startup.
+	 * A full URL string (e.g. "http://example.com") or a pathname (e.g. "/about") can also be passed
+	 * to specify the URL to open.
```
sarah11918, line 1033:
```suggestion
	 * Controls whether the dev server should open in your browser window on startup.
	 * 
	 * Pass a full URL string (e.g. "http://example.com") or a pathname (e.g. "/about") to specify the URL to open.
```

sarah11918 (commented):
:baby_bottle: suggestion, @bluwy! Thanks for adding this!

sarah11918 (approved):
Looks great!

PR #9579, opened 2024-01-02
quality of life updates for `App`

lilnasy:
## Why?
This came from working on 5 adapters over the past month. The logic for dealing with cookies and `node:http` is always duplicated. `NodeApp` takes care of converting the request, but the logic of converting response is duplicated across the node and the vercel adapters. Cookies had to be handled manually for legacy reasons (I think), but all adapters we maintain are now doing the same thing.

## Changes
- Exposes well-known symbols as a static property on App. (`App.Symbol.locals`, `App.Symbol.clientAddress`)
- Adds helper functions for node-based runtimes to manage WHATWG Request and Response objects as static methods on `NodeApp`.
- `App.getSetCookieFromResponse()` in favor of `app.setCookieHeaders()` (confusing name)
- `app.render(request, { addCookieHeader: true })`

## Testing
Tests for most of the functionality here are already present. This PR only exposes the APIs on App and NodeApp to prevent duplication across adapters.

## Docs
_Pending_

### DIFF
--- a/.changeset/cool-foxes-talk.md
+++ b/.changeset/cool-foxes-talk.md
@@ -1,0 +1,34 @@
+---
+"astro": minor
+---
+
+Adds new helper functions for integration developers.
+
+- Symbols used to represent `Astro.locals` and `Astro.clientAddress` are now available as static properties on the `App` class: `App.Symbol.locals` and `App.Symbol.clientAddress`.
+
+- `Astro.clientAddress` can now be passed directly to the `app.render()` method.
+```ts
+const response = await app.render(request, { clientAddress: "012.123.23.3" })
+```
+
+- Helper functions for converting node http request and response objects to web-compatible `Request` and `Response` objects are now provided as static methods on the `NodeApp` class.
+```ts
+http.createServer((nodeReq, nodeRes) => {
+    const request: Request = NodeApp.createRequest(nodeReq)
+    const response = await app.render(request)
+    NodeApp.writeResponse(response, nodeRes)
+})
+```
+
+- Cookies added via `Astro.cookies.set()` can now be automatically added to the `Response` object by passing the `addCookieHeader` option to `app.render()`.
+```diff
+-const response = await app.render(request)
+-const setCookieHeaders: Array<string> = Array.from(app.setCookieHeaders(webResponse));
+
+-if (setCookieHeaders.length) {
+-    for (const setCookieHeader of setCookieHeaders) {
+-        headers.append('set-cookie', setCookieHeader);
+-    }
+-}
++const response = await app.render(request, { addCookieHeader: true })
+```
--- a/.changeset/cool-foxes-talk.md
+++ b/.changeset/cool-foxes-talk.md
@@ -2,7 +2,7 @@
 "astro": minor
 ---
 
-Adds new helper functions for integration developers.
+Adds new helper functions for adapter developers.
 
 - Symbols used to represent `Astro.locals` and `Astro.clientAddress` are now available as static properties on the `App` class: `App.Symbol.locals` and `App.Symbol.clientAddress`.
 
@@ -11,7 +11,7 @@ Adds new helper functions for integration developers.
 const response = await app.render(request, { clientAddress: "012.123.23.3" })
 ```
 
-- Helper functions for converting node http request and response objects to web-compatible `Request` and `Response` objects are now provided as static methods on the `NodeApp` class.
+- Helper functions for converting Node.js HTTP request and response objects to web-compatible `Request` and `Response` objects are now provided as static methods on the `NodeApp` class.
 ```ts
 http.createServer((nodeReq, nodeRes) => {
     const request: Request = NodeApp.createRequest(nodeReq)
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -198,11 +198,24 @@ export class App {
 	): Promise<Response> {
 		let routeData: RouteData | undefined;
 		let locals: object | undefined;
+		let clientAddress: string | undefined;
+		let addCookieHeader: boolean | undefined;
 
 		if (
 			routeDataOrOptions &&
-			('routeData' in routeDataOrOptions || 'locals' in routeDataOrOptions)
+			(
+				'addCookieHeader' in routeDataOrOptions ||
+				'clientAddress' in routeDataOrOptions ||
+				'locals' in routeDataOrOptions ||
+				'routeData' in routeDataOrOptions
+			)
 		) {
+			if ('addCookieHeader' in routeDataOrOptions) {
+				addCookieHeader = routeDataOrOptions.addCookieHeader;
+			}
+			if ('clientAddress' in routeDataOrOptions) {
+				clientAddress = routeDataOrOptions.clientAddress;
+			}
 			if ('routeData' in routeDataOrOptions) {
 				routeData = routeDataOrOptions.routeData;
 			}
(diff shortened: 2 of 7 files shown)


.changeset/cool-foxes-talk.md
```diff
@@ -0,0 +1,34 @@
...
+Adds new helper functions for integration developers.
+
+- Symbols used to represent `Astro.locals` and `Astro.clientAddress` are now available as static properties on the `App` class: `App.Symbol.locals` and `App.Symbol.clientAddress`.
+
+- `Astro.clientAddress` can now be passed directly to the `app.render()` method.
+```ts
+const response = await app.render(request, { clientAddress: "012.123.23.3" })
+```
+
+- Helper functions for converting node http request and response objects to web-compatible `Request` and `Response` objects are now provided as static methods on the `NodeApp` class.
```
ematipico, line 14:
```suggestion
- Helper functions for converting Node.js HTTP request and response objects to web-compatible `Request` and `Response` objects are now provided as static methods on the `NodeApp` class.
```

> HTTP request and response objects

I suggest being more explicit. I suppose you meant `http.IncomingMessage`? Or did you mean something else?

ematipico (commented):
Here's some preliminary feedback: since there isn't an issue nor an RFC where reviewers can understand the **reasons** and **use cases** of these additions, I suggest expanding the description of the PR, and telling us why these APIs were added, with possible use cases.

This will help us reviewers to have a better context and help to land these new features.

.changeset/cool-foxes-talk.md
lilnasy: You're right, I meant to refer to the classes from `node:http`.

lilnasy:
Thanks, added a section to the description.

ematipico:
> but the logic of converting response is duplicated across the node and the vercel adapters.

Could you add a couple of examples?

packages/astro/src/core/app/index.ts
```diff
@@ -29,15 +29,44 @@ import { EndpointNotFoundError, SSRRoutePipeline } from './ssrPipeline.js';
...
 export interface RenderOptions {
-	routeData?: RouteData;
+	/**
+	 * Whether to automatically add all cookies written by `Astro.cookie.set()` to the response headers.
+	 * 
+	 * When set to `true`, they will be added to the `Set-Cookie` header as comma-separated key=value pairs. You can use the standard `response.headers.getSetCookie()` API to read them individually.
+	 * 
+	 * When set to `false`, the cookies will only be available from `App.getSetCookieFromResponse(response)`.
+	 * 
+	 * Default: `false`
```
ematipico, line 48:
```suggestion
	 * @default {false}
```

packages/astro/src/core/app/index.ts
```diff
@@ -51,6 +80,15 @@ export interface RenderErrorOptions {
...
 
 export class App {
+
+	/**
+	 * Symbols that the Astro app reads on the passed Request instance. Use these when you can't directly provide these values to `app.render()`.
+	 */
+	static readonly Symbol = Object.freeze({
+		locals: Symbol.for('astro.locals'),
+		clientAddress: Symbol.for('astro.clientAddress'),
+	})
```
ematipico, line 90: If we want to **expose** these symbols to user-land, I think we should do it a stricter way. We have been using `Symbol.for` because we know that this is internal information.

Although, this API adds a layer fo contract with the adapters. I think it's time to use  e.g.`Symbol("astro.locals")` and use this one symbol across the board. 

Also, I would expose these values via a getter function:

```js
import { LocalsSymbol } from "../constants.js"

class App {
	get localsSymbol() {
		return LocalsSymbol;
	}
}
```

packages/astro/src/core/app/node.ts
```diff
@@ -151,3 +168,36 @@ export async function loadApp(rootFolder: URL): Promise<NodeApp> {
 	const manifest = await loadManifest(rootFolder);
 	return new NodeApp(manifest);
 }
+
+/**
+ * Takes in a nullable WebAPI Headers object and produces a NodeJS OutgoingHttpHeaders object suitable for usage
+ * with ServerResponse.writeHead(..) or ServerResponse.setHeader(..)
+ *
+ * @param webHeaders WebAPI Headers object
```
ematipico, line 176:
```suggestion
 * @param headers WebAPI Headers object
```

packages/astro/src/core/app/node.ts
```diff
@@ -151,3 +168,36 @@ export async function loadApp(rootFolder: URL): Promise<NodeApp> {
...
+ * @returns NodeJS OutgoingHttpHeaders object with multiple set-cookie handled as an array of values
```
ematipico, line 177:
```suggestion
 * @returns {OutgoingHttpHeaders} NodeJS OutgoingHttpHeaders object with multiple set-cookie handled as an array of values
```

packages/astro/src/core/app/node.ts
```diff
@@ -151,3 +168,36 @@ export async function loadApp(rootFolder: URL): Promise<NodeApp> {
...
+ */
+export const createOutgoingHttpHeaders = (
```
ematipico, line 179: Does it need to be exported? If so, can we add an example in the JSDoc to show how to use it?

lilnasy: It needs to be exported for unit testing. However, it can be exported from another file that is not exposed from the astro package.

packages/astro/src/core/app/index.ts
lilnasy: I am guessing the getter is for preventing mutation. Object.freeze should have us covered there.

ematipico: > I think it makes sense to keep the symbol registered for compatibility. Adapters may also have constraints that prevent them from importing anything in the relevant code path, using the registered symbol directly could help there.

I can't understand your point here.

> Exposing it here serves the purpose of not having to remember the casing etc ("did it have dots or underscores?")

And that's why we should have the casing written in one single place :)

Feel free to keep using `Symbol.for`, but the actual string now should be saved somewhere as constant and shared.

packages/astro/src/core/app/node.ts
ematipico: Then maybe it's having this function inside the test utilities

lilnasy: Moved away from public API in the latest commit.

packages/astro/src/core/app/index.ts
ematipico: Again, I don't understand why this is an issue. An adapter is already doing `import { App } from "astro/app"`, so it's importing code from astro

lilnasy: That is true, it is for cases where the adapter may want to load locals in another file which does not import `App`, and can't. Basically, `Symbol.for` is versatile.

lilnasy: > And that's why we should have the casing written in one single place :)

I agree. After having looked at the our use of registered symbols, it's widespread enough to warrant its own PR.

ematipico (commented):
I would like to see some tests to make sure that the new APIs behave as expected. We can add these tests in a separate PR if you want, but I think it's best to have them there, together with the new logic

lilnasy:
Yeah, that makes sense.

ematipico (approved):
Looks good to me, let's merge it and move to the next one :)

ematipico:
@lilnasy I rebased this PR and the `adapter-enhancements` branch

lilnasy:
Thanks for the rebase!

PR #9582, opened 2024-01-02
quality of life updates for the node adapter

lilnasy:
## Changes
- Depends on #9579.
- `createOutgoingHeaders` is moved into `NodeApp` - its logic was duplicated between node and vercel.
- the entrypoint is now called `server-entrypoint.ts` (previously "server.ts")
- the part responsible for serving static files is now called `serve-static.ts` (previously "http-server.ts").
- the part responsible for serving on-demand rendered pages is now in `serve-app.ts` (previously `http-server.ts` and `nodeMiddleware.ts`)
- the part responsible for gathering local and remote network addresses for logging is now exclusively in `log-listening-on.ts` (previously `get-network-address.ts`, `standalone.ts`, and `preview.ts`)
- the part responsible for creating an express compatible middleware is now called `middleware.ts` (previously nodeMiddleware.ts)

## Testing
No new functionality is introduced. Existing tests should pass.

## Docs
Does not affect behavior.


packages/integrations/node/package.json
```diff
@@ -19,8 +19,8 @@
   "homepage": "https://docs.astro.build/en/guides/integrations-guide/node/",
   "exports": {
     ".": "./dist/index.js",
-    "./server.js": "./dist/server.js",
-    "./preview.js": "./dist/preview.js",
+    "./server.js": "./dist/server-entrypoint.js",
+    "./preview.js": "./dist/preview-entrypoint.js",
```
natemoo-re, line 23: This is the only change I don't love! I think we should probably remove the `-entrypoint` suffix so that the exported path and the actual file path continue to match.

packages/integrations/node/src/preview-entrypoint.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AstroError } from 'astro/errors';
+import logListeningOn from './log-listening-on.js';
+import { createServer } from './standalone.js';
+import type { CreatePreviewServer } from 'astro';
+import type { createExports } from './server-entrypoint.js';
+
+type ServerModule = ReturnType<typeof createExports>;
+type MaybeServerModule = Partial<ServerModule>;
+
+const preview1: CreatePreviewServer = async function (preview) {
```
natemoo-re, line 11: Why not `createPreviewServer`?

```suggestion
const createPreviewServer: CreatePreviewServer = async function (preview) {
```

packages/integrations/node/src/preview-entrypoint.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+	const host = preview.host ?? "127.0.0.1"
+	const port = preview.port ?? 4321
+	const server = createServer(ssrHandler, host, port);
+	logListeningOn(preview.logger, server.server, options)
+	server.server.listen(port, host);
+	return server;
+};
+
+export { preview1 as default }
```
natemoo-re, line 44:
```suggestion
export { createPreviewServer as default }
```

packages/integrations/node/src/standalone.ts
```diff
@@ -1,75 +1,90 @@
...
-	const rel = path.relative(fileURLToPath(serverURLRaw), fileURLToPath(clientURLRaw));
-
-	const serverEntryURL = new URL(import.meta.url);
-	const clientURL = new URL(appendForwardSlash(rel), serverEntryURL);
-
+export default function standalone(app: NodeApp, options: Options) {
+	const port = process.env.PORT ? Number(process.env.PORT) : options.port ?? 8080;
+	// Allow to provide host value at runtime
+	const host_ = typeof options.host === "boolean" ? "127.0.0.1" : options.host
+	const host = process.env.HOST ?? host_;
```
natemoo-re, line 16: Naming adjustment for clarity

```suggestion
	const hostOptions = typeof options.host === "boolean" ? "127.0.0.1" : options.host
	const host = process.env.HOST ?? hostOptions;
```

natemoo-re (approved):
Refactors look great! A few nits on naming

packages/integrations/node/package.json
lilnasy: This was mostly because every other file had its own value that it called a "server", but yeah for entrypoints, it may have been an over-correction.

packages/integrations/node/src/preview.ts
```diff
@@ -33,49 +26,19 @@ const preview: CreatePreviewServer = async function ({
...
-		);
-	} else {
-		logger.info(`Preview server listening on ${address.local[0]}`);
-	}
-
+	const host = preview.host ?? "127.0.0.1"
+	const port = preview.port ?? 4321
+	const server = createServer(ssrHandler, host, port);
+	logListeningOn(preview.logger, server.server, options)
+	server.server.listen(port, host);
```
Fryuni, line 40: This causes a race in the tests on some cases and make environment problems preventing the server from starting fail with timeouts hiding the error.

```suggestion
	await new Promise<void>((resolve, reject) => {
	    server.once('listening', resolve);
	    server.once('error', reject);
	    server.server.listen(port, host);
	});
```

lilnasy: That's perfect, thanks!

packages/integrations/node/src/log-listening-on.ts
```diff
@@ -0,0 +1,84 @@
...
+		);
+	} else {
+		logger.info(`Server listening on ${address.local[0]}`);
+	}
+}
+
+function getResolvedHostForHttpServer(host: string | boolean) {
+	if (host === false) {
+		// Use a secure default
+		return '127.0.0.1';
```
lilnasy, line 30:
```suggestion
		return 'localhost';
```

packages/integrations/node/src/preview.ts
```diff
@@ -33,49 +26,23 @@ const preview: CreatePreviewServer = async function ({
...
-
-	if (host === undefined) {
-		logger.info(
-			`Preview server listening on \n  local: ${address.local[0]} \t\n  network: ${address.network[0]}\n`
-		);
-	} else {
-		logger.info(`Preview server listening on ${address.local[0]}`);
-	}
-
+	const host = preview.host ?? "127.0.0.1"
```
lilnasy, line 36:
```suggestion
	const host = preview.host ?? "localhost"
```

Fryuni:
There is a bug-fix in here now. Maybe mention it on the changeset?

The race condition bug and silent error happens when calling [`preview`](https://docs.astro.build/en/reference/cli-reference/#preview) with the Node adapter.
This is used in the tests for the adapter itself but is public for any project to use so its not just internal

packages/integrations/node/src/serve-app.ts
```diff
@@ -0,0 +1,25 @@
+import { NodeApp } from "astro/app/node"
+import type { RequestHandler } from "./types.js";
+
+/**
+ * Intended to be used in both standalone and middleware mode.
```
ematipico, line 5: We could document what it does, other than where :)

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
+import path from "node:path";
+import url from "node:url";
+import send from "send";
+import type { IncomingMessage, ServerResponse } from "node:http";
+import type { Options } from "./types.js";
+import type { NodeApp } from "astro/app/node";
+
+/**
+ * Intended to be used only in the standalone mode.
+ * This is queried first, and if a static file is not found, it relegates to the SSR handler.
```
ematipico, line 10: "This is queried first" => we miss the "what" is queried first

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export default function createListener(app: NodeApp, options: Options) {
+	const client = resolveClientDir(options);
+    return (req: IncomingMessage, res: ServerResponse, ssr: () => unknown) => {
```
ematipico, line 14: Can we add a small comment that explains what the third argument is?

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+function resolveClientDir(options: Options) {
+	const clientURLRaw = new URL(options.client);
+	const serverURLRaw = new URL(options.server);
+	const rel = path.relative(url.fileURLToPath(serverURLRaw), url.fileURLToPath(clientURLRaw));
+	const serverEntryURL = new URL(import.meta.url);
+	const clientURL = new URL(appendForwardSlash(rel), serverEntryURL);
+	const client = url.fileURLToPath(clientURL);
+	return client;
```
ematipico, line 74: All these `URL` could fail; should we at least add a `try/catch`?

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
...
+				}
+			});
+			stream.on('directory', () => {
+				// On directory find, redirect to the trailing slash
+				let location: string;
+				if (req.url!.includes('?')) {
+					const [url1 = '', search] = req.url!.split('?');
+					location = `${url1}/?${search}`;
+				} else {
+					location = req.url + '/';
```
ematipico, line 50:
```suggestion
					location = appendForwardSlash(req.url);
```

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
...
+					// Taken from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable
+					_res.setHeader('Cache-Control', 'public, max-age=31536000, immutable');
+				}
+			});
+			stream.on('directory', () => {
+				// On directory find, redirect to the trailing slash
+				let location: string;
+				if (req.url!.includes('?')) {
+					const [url1 = '', search] = req.url!.split('?');
+					location = `${url1}/?${search}`;
```
ematipico, line 48: Is it possible that `url1` already has a forward slash? Should we cover that case? Or let's just use `appendForwardSlash`

packages/integrations/node/src/serve-static.ts
```diff
@@ -0,0 +1,79 @@
...
+					res.end('Internal server error');
+					return;
+				}
+				// File not found, forward to the SSR handler
+				ssr();
+			});
+			stream.on('headers', (_res: ServerResponse) => {
+                // assets in dist/_astro are hashed and should get the immutable header
+				if (pathname.startsWith(`/${options.assets}/`)) {
+					// Taken from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable
```
ematipico, line 39: The link doesn't explain the `max-age`, so we should add a comment that explains why we decided this number. Do you know why?

packages/integrations/node/src/preview.ts
```diff
@@ -33,49 +26,23 @@ const preview: CreatePreviewServer = async function ({
...
-	if (host === undefined) {
-		logger.info(
-			`Preview server listening on \n  local: ${address.local[0]} \t\n  network: ${address.network[0]}\n`
-		);
-	} else {
-		logger.info(`Preview server listening on ${address.local[0]}`);
-	}
-
+	const host = preview.host ?? "localhost"
+	const port = preview.port ?? 4321
```
ematipico, line 37: Can we know if `4321` **is** already in use?

.changeset/weak-apes-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/node": minor
```
ematipico, line 2: Shouldn't this be a `major` because it uses the new integration `astro/app` APIs?

packages/integrations/node/src/serve-static.ts
lilnasy: `client` and `server` are URLs before being serialized for the adapter, so I'd be surprised. If a failure does happen, the server will fail to start, logging the error anyway.

packages/integrations/node/src/serve-static.ts
lilnasy: I am not sure. This code was the same before the refactor.

packages/integrations/node/src/serve-static.ts
lilnasy: This is the "far future" cache header, used for static whose name/path includes their digest hash. 1 year (31,536,000 seconds) is convention.

packages/integrations/node/src/preview.ts
lilnasy: It can be done with a mild amount of code, but there is no ready-made API, no.

.changeset/weak-apes-add.md
lilnasy: You're right. I'll change that.

packages/integrations/node/src/serve-app.ts
lilnasy: Added :)

https://github.com/withastro/astro/pull/9582/files#diff-e93f38dcd3505fbeb1c5979bca34bc90e48b1fe9b1f3ead1102c31993e5ef286R5-R6

packages/integrations/node/src/serve-static.ts
lilnasy: Added

https://github.com/withastro/astro/pull/9582/files#diff-867b6a94d71f10420472afd23eedb1ed912505336f406d23ac93e502e0810ad4R17

packages/integrations/node/src/serve-static.ts
lilnasy: This is not explicitly stated, but from send's documented example, it seems like it emits this event when it finds a directory instead of a file - a case it wants us to handle. Here, we handle it by redirecting with a trailing slash, with the assumption that the second request will match `index.html` within the directory.

packages/integrations/node/src/serve-static.ts
lilnasy: Added

https://github.com/withastro/astro/pull/9582/files/#diff-867b6a94d71f10420472afd23eedb1ed912505336f406d23ac93e502e0810ad4R10

packages/integrations/node/src/serve-static.ts
ematipico: No, but that function is at the bottom of the file

lilnasy: You're right, I missed that

ematipico (approved):

PR #9545, opened 2023-12-29
Fallback node standalone to localhost

bluwy:
## Changes

close https://github.com/withastro/astro/issues/9133

NOTE: this is a major bump for `@astrojs/node`

The default host is now `localhost` instead of `172.0.0.1`. This aligns with the dev and preview server, and allows it to use ipv6 if the OS chooses so.

## Testing

Existing tests should pass.

## Docs

Added a changeset explaining the breaking change

PR #9583, opened 2024-01-02
chore(vercel): delete request response conversion logic

lilnasy:
## Changes
- Depends on #9579.
- Removes request response conversion logic in favor of `NodeApp.writeResponse`

## Testing

Existing tests should pass.

## Docs
Does not affect behavior.


natemoo-re (approved):

packages/integrations/vercel/src/serverless/entrypoint.ts
```diff
@@ -1,35 +1,21 @@
...
-			if (localsAsString) {
-				locals = JSON.parse(localsAsString);
-			}
-		}
-		await setResponse(app, res, await app.render(request, { routeData, locals }));
+		const clientAddress = req.headers['x-forwarded-for'] as string | undefined;
+		const localsHeader = req.headers[ASTRO_LOCALS_HEADER]
+		const locals =
+			typeof localsHeader === "string" ? JSON.parse(localsHeader)
+			: Array.isArray(localsHeader) ? JSON.parse(localsHeader[0])
```
ematipico, line 15: Can you add a comment that explains why the locals header is an array?

ematipico (approved):

packages/integrations/vercel/src/serverless/entrypoint.ts
lilnasy: That's just its type. Node.js headers yield an array when the same header appears multiple times.

packages/integrations/vercel/src/serverless/entrypoint.ts
```diff
@@ -1,35 +1,21 @@
 import type { SSRManifest } from 'astro';
-import { App } from 'astro/app';
-import { applyPolyfills } from 'astro/app/node';
+import { applyPolyfills, NodeApp } from 'astro/app/node';
```
bluwy, line 2: Question: I've not been following the refactor changes, but with this change, it looks like it'll depend on https://github.com/withastro/astro/pull/9579. So does this mean that it's a breaking change for `@astrojs/vercel` that it relies on a new version of Astro to work?

packages/integrations/vercel/src/serverless/entrypoint.ts
ematipico: Then why do we check if it's a string using `typeof`? Sorry, I might not know what's going on

packages/integrations/vercel/src/serverless/entrypoint.ts
ematipico: We are progressively pushing these refactors to the branch `adapter-enhancements`, eventually all the changes in the adapter will rely on the PR you linked. Eventually, all changes will be shipped together. However, I don't think `@astrojs/vercel` will be a major

bluwy: But won't it only work after that PR is released, aka the latest Astro version released then? An older version of Astro would not work with this new `@astrojs/vercel` changes?

ematipico: You're right. `@astronjs/vercel` will result in a `major` release.

cc @lilnasy

packages/integrations/vercel/src/serverless/entrypoint.ts
lilnasy: It yields a string when the header only appears a single time.

packages/integrations/vercel/src/serverless/entrypoint.ts
lilnasy: That is correct. Peer dependency range will need to be updated to astro 4.2+.

bluwy: Cool. If we're also doing similar things for other adapters, maybe we can group it with this `@astrojs/node` pending major: https://github.com/withastro/astro/pull/9545 (if it's relevant).

lilnasy: That makes sense to me. Do you think it would be easier if #9545 pointed to `adapter-enhancements`. It will be merged into main for astro 4.2.

bluwy: Yup I don't mind changing the base. If it makes it easier for you to make the changes later, feel free to merge that PR into your branch.

PR #9585, opened 2024-01-02
fix(vercel): edge middleware

### LINKED ISSUE #9357: Middleware works locally, doesn't run at all when deployed to Vercel
### Astro Info

```block
Astro                    v4.0.3
Node                     v21.1.0
Package Manager          yarn
Integrations             none
```

_No response_

### Describe the Bug

Trying to deploy my Astro app to vercel. I wrote a middleware that implements basic HTTP authentication. 

When I run my app locally on the astro dev server, everything works as expected—the user/password prompt appears and if I enter the correct credentials, I am forwarded to the correct page. 

When I deploy to Vercel, there is no authentication at all, as if the middleware isn't being run.

Middleware should behave identically whether I am running the project locally or when deploying to Vercel.
**Stackblitz**: https://stackblitz.com/~/github.com/bstro/github-pw6m1w (note the Auth required response)
**Deployment**: https://github-pw6m1w.vercel.app/ (note that the middleware doesn't seem to run at all)

https://stackblitz.com/~/github.com/bstro/github-pw6m1w

### Participation

lilnasy:
## Changes

- Depends on #9583
- Closes #9357
- Creates an edge middleware according to Vercel's Build Output API.

## Testing
Added cases to edge-middleware.test.js

## Docs
Does not affect usage.


lilnasy:
@gislerro Will be ready for review soon after astro 4.2 (wednesday). The work is done, it is just based on a future branch.

Late this week/early next week.

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -68,12 +65,12 @@ export default async function middleware(request, context) {
...
-		const response = await fetch(url, {
+		return new Response(null, {
 			headers: {
-				${JSON.stringify(ASTRO_LOCALS_HEADER)}: trySerializeLocals(ctx.locals)
+				'x-middleware-next': '1',
+				'x-astro-locals': trySerializeLocals(ctx.locals)
 			}
-		});
-		return response;
+		})
```
lilnasy, line 73: Need to make this work without returning an empty response somehow. Maybe it could fetch an internal-only route that leaves out `middlewarePath`.

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -61,20 +65,20 @@ function edgeMiddlewareTemplate(middlewarePath: string, vercelEdgeMiddlewareHand
...
 			headers: {
-				${JSON.stringify(ASTRO_LOCALS_HEADER)}: trySerializeLocals(ctx.locals)
+				...Object.fromEntries(request.headers.entries()),
+				'x-astro-path': request.url.replace(origin, ''),
+				'x-astro-locals': trySerializeLocals(ctx.locals)
 			}
-		});
-		return response;
-	};
+		})
```
lilnasy, line 81: The edge function calls the node server at `/_render`, with the original path as a header.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -314,7 +312,9 @@ export default function vercelServerless({
 							? [
 									{
 										src: '/.*',
-										dest: fourOhFourRoute.prerender ? '/404.html' : 'render',
+										dest: fourOhFourRoute.prerender ? '/404.html'
+											: _middlewareEntryPoint ? '_middleware'
+											: 'render',
```
bluwy, line 317: Should this use `dest` instead? Or the `render` should be `_render`

ematipico (commented):
Sorry to bother you, but this PR has many changes that aren't related to the edge middleware and aren't explained in the PR description.

Could you:
1. Describe the entity to the changes in the PR description. For example, we don't know why you changed `render` to `_render`.
2. Move them in a separate PR

packages/integrations/vercel/src/serverless/adapter.ts
ematipico: I think we should move this strings (`_render`, `_middleware`) to some constants, so we don't do any typo

packages/integrations/vercel/src/serverless/middleware.ts
```diff
@@ -36,17 +31,26 @@ export async function generateEdgeMiddleware(
...
 		format: 'esm',
 		bundle: true,
 		minify: false,
+		plugins: [{
+			name: 'esbuild-namespace-node-built-in-modules',
+			setup(build) {
+				const filter = new RegExp(builtinModules.map((mod) => `(^${mod}$)`).join('|'));
+				build.onResolve({ filter }, (args) => ({ path: 'node:' + args.path, external: true }));
+			},
+		}]
```
ematipico, line 45: Could you add a comment that explains this plugin and what it's for?

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -278,24 +272,28 @@ export default function vercelServerless({
...
-						routeDefinitions.push({
-							src: route.pattern.source,
-							dest: 'render',
-						});
+						if (!route.prerender) routeDefinitions.push({ src: route.pattern.source, dest });
 					}
 				}
+				if (_middlewareEntryPoint) {
+					await createMiddlewareFolder({
+						functionName: '_middleware',
```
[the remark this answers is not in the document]
lilnasy: Vercel routes the folder names to a path on the deployed website so we avoid interfering this way.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -393,7 +418,7 @@ async function createFunctionFolder({
 	// https://vercel.com/docs/build-output-api/v3#vercel-primitives/serverless-functions/configuration
 	await writeJson(vcConfig, {
 		runtime,
-		handler,
+		handler: handler.replaceAll("\\","/"),
```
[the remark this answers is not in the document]
lilnasy: `copyDependenciesToFunction` returns a relative path with backslashes on windows, that breaks vercel.

packages/integrations/vercel/src/serverless/entrypoint.ts
```diff
@@ -10,6 +10,10 @@ export const createExports = (manifest: SSRManifest) => {
 	const handler = async (req: IncomingMessage, res: ServerResponse) => {
 		const clientAddress = req.headers['x-forwarded-for'] as string | undefined;
 		const localsHeader = req.headers[ASTRO_LOCALS_HEADER];
+		const realPath = req.headers['x-astro-path'];
+		if (typeof realPath === 'string') {
+			req.url = realPath;
+		}
```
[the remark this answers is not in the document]
lilnasy: This is so that the edge function can call the serverless function in `next()`.

packages/integrations/vercel/src/serverless/adapter.ts
lilnasy: You're right, it is supposed to be `_render`. Moved to constants.

packages/integrations/vercel/src/serverless/middleware.ts
ematipico: I can see why. Wouldn't this affect the runtime, though? Here, we use the user's Node.js version to create the edge function. This means that there could be a situation where the user's node.js version will be different from the edge middleware version.

What happens if a user uses a module in Node.js v20 but not in Node.js v18 - which is the version of the edge middleware?

packages/integrations/vercel/src/serverless/entrypoint.ts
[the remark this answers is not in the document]
ematipico: I still think it's worth adding a comment because the read and write of the header happens in two different places, so it could be helpful to give some context.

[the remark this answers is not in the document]
lilnasy: Added

https://github.com/withastro/astro/pull/9585/files#diff-071ccfbdfd7b9bc803219179f88ae0bdb227a399bc4a481b7831c82b801d7a17R32-R43

packages/integrations/vercel/src/serverless/middleware.ts
lilnasy: That is a real corner case. Currently, `inspector/promises` is the one module absent from node 18 but available in node 20. I'm guessing the error message from vercel in that case would point the users towards the right node version they should configure.

ematipico (approved):

PR #9586, opened 2024-01-02
Fixes titles in the browser's dropdown for back / forwards traversals through the browser history

martrapp:
## Changes

It's strange that nobody has complained yet:
The page titles displayed in the browser UI drop-downs for the history navigation were off by one because we call `pushState()` only after the DOM swap during view transitions. 

One way to fix this would be to update the history before the swap. However, I prefer to keep history management and scroll restoration together as they are interdependent. And reaching the scroll target position may not be possible until after the swap. 

Therefore, the original page title is now saved before the swap and used later when updating the history.

# Testing
Manually tested.
No idea how to automatically test this. 

## Docs
n.a. bug fix

### DIFF
--- a/.changeset/silent-rabbits-cross.md
+++ b/.changeset/silent-rabbits-cross.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes page titles in the browser's drop-down for back / forward navigation when using view transitions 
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Looks good to me!

PR #9591, opened 2024-01-03
fix(vercel): include 404.html as fallback in the static adapter

### LINKED ISSUE #9578: Custom 404 page not working on Vercel (output mode static)
### Astro Info

```block
Astro                    v4.0.9
Package Manager          npm
Adapter                  @astrojs/vercel
                         astro-icon
                         @astrojs/mdx
```

_No response_

### Describe the Bug

- Create a custom [404.astro](https://github.com/devmatteini/website/blob/280e7ef1914025fc514a38eef23f4dd51867502b/src/pages/404.astro) page.
- Deploy to Vercel with Astro output mode on `static`
  I use vercel cli:
  ```
  vercel pull --yes --environment production --token "$VERCEL_TOKEN"
  astro build
  vercel deploy --prebuilt --prod --token "$VERCEL_TOKEN"
  ```
- Open your website in a non-existing route and the custom 404 page is not shown. The vercel 404 default page is shown instead.
  vercel default 404 page
 
This is the repository: https://github.com/devmatteini/website
I just ported this repository from Next.js to Astro, so I might have done something wrong somewhere.

Just to give more information, I can see from Vercel deployment `Source` tab that the deployed files and the 404.html exists (`.vercel/output/static/404.html`).

This is the rest of the content:
vercel deployed source

Interestingly, if you go to `https://<domain>/404.html` the custom 404 page is rendered correctly.  

The custom defined 404 page should be used instead of Vercel's default one when reaching a non-existing route

I'm not sure a minimal example with Vercel is doable since it would require a personal Vercel account

### Participation

lilnasy:
## Changes
- Closes #9578

## Testing
Added `static.test.js` along with a fixture.

## Docs
Does not affect behavior.

### DIFF
--- a/packages/integrations/vercel/src/static/adapter.ts
+++ b/packages/integrations/vercel/src/static/adapter.ts
@@ -118,6 +118,11 @@ export default function vercelStatic({
 							continue: true,
 						},
 						{ handle: 'filesystem' },
+						{
+							src: `/*`,
+							dest: `/404.html`,
+							status: 404,
+						}
 					],
 					...(imageService || imagesConfig
 						? {
--- a/packages/integrations/vercel/src/static/adapter.ts
+++ b/packages/integrations/vercel/src/static/adapter.ts
@@ -118,11 +118,11 @@ export default function vercelStatic({
 							continue: true,
 						},
 						{ handle: 'filesystem' },
-						{
-							src: `/*`,
+						...routes.find(route => route.component.endsWith("/pages/404.astro")) ? [{
+							src: `/.*`,
 							dest: `/404.html`,
 							status: 404,
-						}
+						}] : [],
 					],
 					...(imageService || imagesConfig
 						? {
--- a/packages/integrations/vercel/src/static/adapter.ts
+++ b/packages/integrations/vercel/src/static/adapter.ts
@@ -118,7 +118,7 @@ export default function vercelStatic({
 							continue: true,
 						},
 						{ handle: 'filesystem' },
-						...routes.find(route => route.component.endsWith("/pages/404.astro")) ? [{
+						...routes.find(route => route.pathname === "/404") ? [{
 							src: `/.*`,
 							dest: `/404.html`,
 							status: 404,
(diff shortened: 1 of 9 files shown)


lilnasy:
!preview 9591

lilnasy:
! preview 9591

lilnasy:
!preview 9591

packages/integrations/vercel/src/static/adapter.ts
```diff
@@ -118,6 +118,11 @@ export default function vercelStatic({
 							continue: true,
 						},
 						{ handle: 'filesystem' },
+						...routes.find(route => route.component.endsWith("/pages/404.astro")) ? [{
```
natemoo-re, line 121: Not blocking at all: is `route.component` the right thing to check here? Seems like `route.pathname?.endsWith('/404')` might be better?

Maybe we should expose if the site has a custom 404 page to adapters...

natemoo-re (approved):
I can't believe you finally figured this out! Great job!

lilnasy:
Was this a known issue? It was reported just yesterday.

packages/integrations/vercel/src/static/adapter.ts
lilnasy: Yeah we definitely should. I would've been the happiest with an exact check for `src/pages/404.astro`, but `src` is configurable so this is the next best option. Do you know off the top of your head whether `pathname` is immune from configuration options, like `trailingSlash`?

packages/integrations/vercel/src/static/adapter.ts
```diff
@@ -118,6 +118,11 @@ export default function vercelStatic({
 							continue: true,
 						},
 						{ handle: 'filesystem' },
+						...routes.find(route => route.pathname === "/404") ? [{
```
ematipico, line 121: I think the safest way to handle this case is by using the `.pattern` property. Also, this check doesn't take into consideration the `base` option, so we can't redirect to `/404.html` because it could be outside of Astro jurisdiction

lilnasy: `pattern`, after being converted from RegExp to a string, would be the most difficult to match against because it includes base and trailingSlash.

To clarify the intent here, we only need to know whether there will be a 404.html in the filesystem.

ematipico (approved):

PR #9595, opened 2024-01-03
[docs] update config-reference for 4.1.0 docs release

sarah11918:
## Changes

- Updates the wrong version in a `<Since />` component showing the version added
- Adds docs (we'll see whether I got the indentation correct!) for the new `limitInputPixels` config option that was only shown in an example, not documented.

## Testing

no tests, only docs

## Docs

Docs update


TheOtterlord (approved):
Indentation looks great!

packages/astro/src/@types/astro.ts
```diff
@@ -1112,6 +1112,20 @@ export interface AstroUserConfig {
...
+			 * @type {boolean}
+			 * @default `true`
+			 * @version 4.1.0
+			 * @description
+			 *
+		   * Whether or not to limit the size of images that the Sharp integration will process.
+		   *
+			 * Set `false` to bypass the default image size limit for the Sharp integration and process large images.
+			 */
+			prefixDefaultLocale: boolean;
```
Princesseuh, line 1128:
```suggestion
```

packages/astro/src/@types/astro.ts
```diff
@@ -1112,6 +1112,19 @@ export interface AstroUserConfig {
...
+			/**
+			 * @docs
+			 * @name image.service.config.limitInputPixels
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `true`
+			 * @version 4.1.0
+			 * @description
+			 *
+		   * Whether or not to limit the size of images that the Sharp integration will process.
```
Princesseuh, line 1124:
```suggestion
		   * Whether or not to limit the size of images that the Sharp image service will process.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1112,6 +1112,19 @@ export interface AstroUserConfig {
...
+		   *
+			 * Set `false` to bypass the default image size limit for the Sharp integration and process large images.
```
Princesseuh, line 1126:
```suggestion
			 * Set `false` to bypass the default image size limit for the Sharp image service and process large images.
```

Wish we could've linked to Sharp's docs for this option, unfortunately there's no way to make a permalink to it...

PR #9596, opened 2024-01-03
feat: update extended client:visible to use an object instead of a string

Princesseuh:
## Changes

Revert #9363 and redo it using an object instead of a string, that way it can supports options we'll add in the future.

## Testing

Tested manually, like the original PR.

## Docs

Will update docs!


packages/astro/src/runtime/client/visible.ts
```diff
@@ -11,8 +11,11 @@ const visibleDirective: ClientDirective = (load, options, el) => {
 		await hydrate();
 	};
 
-	const ioOptions = {
-		rootMargin: typeof options.value === 'string' ? options.value : undefined,
+	const rawOptions =
+		typeof options.value === 'object' ? (options.value as ClientVisibleOptions) : undefined;
```
natemoo-re, line 15: Just checking, do we already use `JSON.stringify`/`JSON.parse` on `options.value`? If not, we probably get a string here that needs to be deserialized.

natemoo-re (commented):
Looks good! Can we add a test that these arguments are properly encoded in the HTML?

packages/astro/src/runtime/client/visible.ts
Princesseuh: We do! I tested the change locally and everything works. I'm not sure where we do that, though. I'll add a E2E test.. somewhere

packages/astro/e2e/custom-client-directives.test.js
```diff
@@ -89,4 +89,13 @@ function testClientDirectivesShared() {
 		// Hydrated, this should be 1
 		await expect(counterValue).toHaveText('1');
 	});
+
+	test('Client directives should be passed options correctly', async ({ astro, page }) => {
```
Princesseuh, line 93: This doesn't test it for `client:visible` exactly... But still, it test that options are passed correctly to client directives (custom ones too, in this case)

packages/astro/e2e/fixtures/custom-client-directives/client-options.js
```diff
@@ -0,0 +1,6 @@
+// Hydrate on first click on the window
+export default async (load, options) => {
+		const hydrate = await load()
+		document.write(`<div id="options">Passed options are: ${JSON.stringify(options.value)}</div>`)
```
natemoo-re, line 4: Lmao I have not seen a `document.write` in quite a while!

natemoo-re (approved):
Nice test! Thank you!

packages/astro/src/runtime/client/visible.ts
natemoo-re, line 15: Final nit: this _technically_ adds support for all `IntersectionObserverInit` values, but our types are narrowed to only support `Pick<IntersectionObserverInit, 'rootMargin'>`. Is that desired?

Princesseuh: It doesn't, see the code below

PR #9600, opened 2024-01-03
chore: Update`possibleConfigFiles`

jacobdalamb:
## Changes

- What does this change?

## Testing

n/a

## Docs

https://github.com/withastro/docs/pull/6093


lilnasy (approved):
Quick context for reviewers, `astro add` checks whether any one of `possibleConfigFiles` exists to determine whether to create a fresh one.

packages/astro/src/cli/add/index.ts
```diff
@@ -160,6 +160,7 @@ export async function add(names: string[], { flags }: AddOptions) {
 					possibleConfigFiles: [
 						'./tailwind.config.cjs',
 						'./tailwind.config.mjs',
+						'./tailwind.config.ts',
```
bluwy, line 163: Should we also add .mts and .cts here?

florian-lefebvre: I think so since we already support cjs and mjs

ematipico (commented):
A changeset is required here, would it be possible to add one?

packages/astro/src/cli/add/index.ts
jacobdalamb: Added!

jacobdalamb:
> A changeset is required here, would it be possible to add one?

Added!

bluwy (approved):

ematipico (approved):

PR #9606, opened 2024-01-04
Fix `.html` file escaping

### LINKED ISSUE #9480: 🐛 BUG: `.html` pages unescape escaped entities in JS strings
### What version of `@astrojs/compiler` are you using?

0.27.2 (astro@1.5.2)

npm

macOS / Stackblitz

### Describe the Bug

It appears that `.html` pages unescape escaped entities in JS strings causing issues when Vite tries to load the output.

For example:

```html
<script>
  const count = 6;
  console.log(`There are \`${count}\`!`);
  console.log('Multiline\n string.');
</script>
```

- The escaped backticks cause a `Parse failure: Expecting Unicode escape sequence \uXXXX` error
- The `\n` gets converted into a literal new line character, breaking the JS string syntax.

https://stackblitz.com/edit/github-prairu?file=src/pages/index.html

eryue0220:
## Changes

Close #9480 

Correctly parse in `.html`
```js
`There are \`${count}\` things!`
```

- What does this change?

## Testing

Add `test/html-escape-bug.test.js` along with fixture.

## Docs

n/a. bug fix to the template.


.changeset/metal-garlics-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": major
```
ematipico, line 2:
```suggestion
"astro": minor
```

ematipico (approved):

packages/astro/src/vite-plugin-html/transform/utils.ts
```diff
@@ -25,5 +25,5 @@ export function needsEscape(value: any): value is string {
 	return typeof value === 'string' && (value.includes('`') || value.includes('${'));
 }
 export function escape(value: string) {
-	return value.replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
+	return value.replace(/(\\*)\`/g, '\\`').replace(/\$\{/g, '\\${');
```
natemoo-re, line 28: This is definitely better but might fail if the backslash itself is escaped. I think we might need to use a different approach entirely? I'm going to look into some options.

natemoo-re (commented):
Thanks for the PR! There might be a different approach for this, but this is definitely a step in the right direction.

packages/astro/src/vite-plugin-html/transform/utils.ts
eryue0220: Yeah. But I have a question, is there any cases in this scenario that the backslash itself is escaped. On the other hand, before the `escape` function call, it should pass `needEscape` check.

.changeset/metal-garlics-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
```
MoustaphaDev, line 2: Should this be a patch maybe?
```suggestion
"astro": patch
```

packages/astro/src/vite-plugin-html/transform/escape.ts
```diff
@@ -3,19 +3,21 @@ import type MagicString from 'magic-string';
...
-					s.overwrite(node.position!.start.offset!, node.position!.end.offset!, escape(node.value));
+					s.overwrite(node.position!.start.offset!, node.position!.end.offset!, escapeTemplateLiteralCharacters(node.value));
 				}
 			} else if (node.type === 'element') {
-				for (const [key, value] of Object.entries(node.properties ?? {})) {
-					const newKey = needsEscape(key) ? escape(key) : key;
-					const newValue = needsEscape(value) ? escape(value) : value;
+				if (!node.properties) return;
+				for (let [key, value] of Object.entries(node.properties)) {
+					key = key.replace(/([A-Z])/g, '-$1').toLowerCase()
```
MoustaphaDev, line 18: I'm curious about the decision to convert attributes to kebab-case. Could you share the reasoning behind it?
Shouldn't it be left to the browser?

MoustaphaDev (approved):
Looks great, I'm just curious about the reasoning behind some code

packages/astro/src/vite-plugin-html/transform/escape.ts
natemoo-re: Oh this was working around an interesting behavior of one of the rehype parsers. They normalize `kebab-case` attributes into a `camelCase` key on the `node.properties` object, but we need to match the key back to the source code, which was probably using the `kebab-case` format.

This code allows us to match up the parsed AST with the source code for sourcemaps.

MoustaphaDev: Ah gotcha, my concern was that an attribute like `MyAttribute` would end up being `-my-attribute`, but that won't happen since the value is always camel case 👍️

packages/astro/src/vite-plugin-html/transform/index.ts
```diff
@@ -5,7 +5,7 @@ import escape from './escape.js';
 import slots, { SLOT_PREFIX } from './slots.js';
 
 export async function transform(code: string, id: string) {
-	const s = new MagicString(code, { filename: id });
+	let s = new MagicString(code, { filename: id });
```
eryue0220, line 8: maybe it's no need to use `let` here.

natemoo-re: Good catch!

natemoo-re, line 8:
```suggestion
	const s = new MagicString(code, { filename: id });
```

PR #9603, opened 2024-01-04
perf: disable streaming in SSG

ematipico:
## Changes

Disable streaming in case we do an SSG build.

Internally, we noticed disabling streaming in certain projects yields to better build times.

## Testing

Manual testing and launch the benchmark job

## Docs

N/A

PR #9614, opened 2024-01-04
Improve Node.js performance using an AsyncIterable

matthewp:
## Changes

- In Node.js, use an AsyncIterable instead of a ReadableStream, as the latter is slow.

## Testing

- Benchmarks should be better.

## Docs
N/A, bug fix


matthewp:
!bench render

matthewp:
!bench render

.changeset/hip-cherries-behave.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": patch
+---
+
+Improve Node.js streaming performance
```
florian-lefebvre, line 5:
```suggestion
Improves Node.js streaming performance
```

delucis:
FWIW I don’t expect `!bench` to reflect the benefits here — also saw the same in #9603 — as they only kick in beyond a certain threshold.

With the #9603 preview release the benefit kicked in once list items in one `Array#map` passed around 2000:

chart showing build time increasing exponentially as list size increases for current Astro and a more or less flat curve as list size increases with a preview release with streaming disabled

That benchmark also suggests a similar slight slowdown for smaller list sizes seen in the benchmark above:

| sidebar items | build time | build time (preview release) | % change |
| ------------: | ---------: | ---------------------------: |---------:|
|             0 |   5,488 ms |                     6,981 ms | +27% |
|          1000 |   7,086 ms |                     8,202 ms | +15% |
|          2000 |  16,888 ms |                     8,770 ms | -48% |

Not sure if those are statistically significant though — just 10 runs each using `hyperfine`.

matthewp:
@delucis I'm going to update the benchmark (or create a new one, not sure), which shows the benefit.

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -100,13 +61,87 @@ export async function renderToReadableStream(
 					}
 
 					const bytes = chunkToByteArray(result, chunk);
-					controller.enqueue(bytes);
+					handler.push(bytes);
```
bluwy, line 64: Perf-wise, I'm a bit concerned that we're consolidating on handling plain `UInt8Array` here. `chunkToByteArray` and `chunkToString` are implemented this way:

https://github.com/withastro/astro/blob/47b951b3888a5a8a708d2f9b974f12fba7ec9ed3/packages/astro/src/runtime/server/render/common.ts#L131-L142

https://github.com/withastro/astro/blob/47b951b3888a5a8a708d2f9b974f12fba7ec9ed3/packages/astro/src/runtime/server/render/common.ts#L123-L129

Which means that with this PR, `renderToString`'s `string` chunks would convert to `UInt8Array`, and then back as `string` again:

https://github.com/withastro/astro/blob/6b4f77c100da3401f1749184e15b5218c947c70c/packages/astro/src/runtime/server/render/astro/render.ts#L119

It's the reason why the initial render functions has slightly duplicated code, because strings can be fast-path-ed. Perhaps we should keep the fast path like before?

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -125,6 +160,54 @@ export async function renderToReadableStream(
...
 }
 
+export async function renderToAsyncIterable(
+	result: SSRResult,
+	componentFactory: AstroComponentFactory,
+	props: any,
+	children: any,
+	isPage = false,
+	route?: RouteData
+): Promise<AsyncIterable<Uint8Array>> {
```
bluwy, line 170: Question: Since we know strings as faster to deal with, could the async iterable return strings instead? Not sure if Node supports that.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Yeah totally. I didn't feel great about that either. The reason I did it was because the `renderToString`, `renderToReadableStream` (and the new) `renderToAsyncIterator` are 90% duplicate code and wanted to avoid having to add any new code in 3 places. So this refactor was meant to make it so that the functions only handle their different outputs. But for `renderToString` I could either revert back to the old way, or figure out some way to make the abstraction handle either the typed array way or the string way.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: It might, but why do you think strings are faster to deal with? That's not obvious to me at least. In SSG everything gets turned into an array buffer (we call `response.arrayBuffer()`), and in dev + ssr prod everything gets turned into bytes too. So it makes sense to me to stay in bytes. Unless the string concatenation part is much faster.

Once the PR is ready I'd be happy to test a second implementation to see.

bluwy: I mentioned strings are faster in relation to my comment above, where the strings don't have to be encoded and decoded again. The default/base type of the chunks are strings. And I guess theoretically speaking concatenating strings will be faster than with bytes: https://github.com/withastro/astro/blob/6b4f77c100da3401f1749184e15b5218c947c70c/packages/astro/src/runtime/server/render/astro/render.ts#L80-L92

If we want to deal with bytes, we could concat the strings and then encode them as bytes in one step.

.changeset/hip-cherries-behave.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": patch
+---
+
+Improves Node.js streaming performance
+
+This uses an AsyncIterable instead of a ReadableStream to do streaming in Node.js. This is a non-standard enhancement by Node, so this is done only in that environment.
```
ematipico, line 7:
```suggestion
Improves Node.js streaming performance.

This uses an `AsyncIterable` instead of a `ReadableStream` to do streaming in Node.js. This is a non-standard enhancement by Node, which is done only in that environment.
```

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +166,101 @@ async function bufferHeadContent(result: SSRResult) {
...
+
+			if(error) {
+				throw error;
+			}
+
+      // Get the total length of all arrays.
+      let length = 0;
+      chunks.forEach(item => {
+        length += item.length;
+      });
```
ematipico, line 208: nit: let's use a `const length = chunks.map()` instead

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +166,101 @@ async function bufferHeadContent(result: SSRResult) {
...
+
+      // Create a new array with total length and merge all source arrays.
+      let mergedArray = new Uint8Array(length);
+      let offset = 0;
+      chunks.forEach(item => {
```
ematipico, line 213: Here we can use a `for..of` loop

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +166,101 @@ async function bufferHeadContent(result: SSRResult) {
...
+
+      return returnValue;
+    }
+  };
+
+  const destination: RenderDestination = {
+    write(chunk) {
+      if (isPage && !renderedFirstPageChunk) {
+        renderedFirstPageChunk = true;
+        if (!result.partial && !/<!doctype html/i.test(String(chunk))) {
```
ematipico, line 234: Have this regex in a top-level constant, we don't need to create it every time.

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -196,3 +196,25 @@ export function renderToBufferDestination(bufferRenderFunction: RenderFunction):
...
+export const isNode = typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]';
+
+// We can get rid of this when Promise.withResolvers() is ready
+export type PromiseWithResolvers<T> = {
+	promise: Promise<T>
+	resolve: (value: T) => void;
+	reject: (reason?: any) => void;
+}
+
+export function promiseWithResolvers<T = any>(): PromiseWithResolvers<T> {
```
ematipico, line 209: Can you add some documentation that explains the utility?

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +166,101 @@ async function bufferHeadContent(result: SSRResult) {
...
+          const doctype = result.compressHTML ? "<!DOCTYPE html>" : "<!DOCTYPE html>\n";
+          chunks.push(encoder.encode(doctype));
+        }
+      }
+      if (chunk instanceof Response) {
+        throw new AstroError({
+          ...AstroErrorData.ResponseSentError
+        });
```
ematipico, line 242:
```suggestion
        throw new AstroError(AstroErrorData.ResponseSentError);
```

No need to create a new object

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +166,101 @@ async function bufferHeadContent(result: SSRResult) {
 		}
 	}
 }
+
+export async function renderToAsyncIterable(
```
ematipico, line 170: I am a bit worried that we don't have any comments that explain what this function is doing. We should explain both high-level and low-level stuff, so other developers can understand the logic in case there's some bugfix to do.

.changeset/hip-cherries-behave.md
sarah11918, line 7: What's the implication of this being "a non-standard enhancement by Node" and why does the reader care about this? 

"Non-standard" doesn't always imply something positive, (is this a hack? are we jumping on to something experimental?) and in fact might make people worry/wonder. So this feels maybe a little too jargon-y when I'm assuming that this could actually be framed as something cool and helpful!

Just a thought!

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: You mean reduce? Why though? it's equivalent code.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: I want to avoid for/of since this is perf critical code. I can change to a tranditional for loop if you prefer though.

.changeset/hip-cherries-behave.md
matthewp: It just means that it won't be run non-Node environments, for ex. in Cloudflare. Those will take the slower path. 

Yeah it can be tricky, I'm not always sure who the target of these changesets are. A lot of changes are implementation details that most won't understand or care about, but some will. And they can be useful for understanding the change on the part of maintainers. But we don't usually do that.

Thinking... do you have a suggested way to rephrase this?

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Switched to a plain for loop, let me know what you think.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Switched to a regular for loop.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Moved it to the top

packages/astro/src/runtime/server/render/util.ts
matthewp: Added

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Added comments throughout the function and tried to explain how it works. Also happy to explain on a call if you'd like!

packages/astro/src/runtime/server/render/astro/render.ts
ematipico: Ah sorry I missed the `+=`. Yeah for loop was the other suggestion

ematipico (approved):
Thank you! Looks good

bluwy (approved):

ematipico:
!preview render-nodejs

delucis:
Ran the same benchmark as mentioned in https://github.com/withastro/astro/pull/9614#issuecomment-1878540695 using the prerelease and saw a similar improvement to benchmarks for #9603: build time roughly doubles from 5s to 10s as sidebar size increases from 0 to 7,000 items. With the public release, build time increased c. 20x for the same sidebar size.

graph showing prerelease build times almost flat while public release build time increases exponentially as sidebar size increases

.changeset/hip-cherries-behave.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": patch
```
matthewp, line 2:
```suggestion
"astro": minor
```

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +168,119 @@ async function bufferHeadContent(result: SSRResult) {
...
+			}
+
+      // Get the total length of all arrays.
+      let length = 0;
+			for(let i = 0, len = buffer.length; i < len; i++) {
+				length += buffer[i].length;
+			}
+
+      // Create a new array with total length and merge all source arrays.
+      let mergedArray = new Uint8Array(length);
```
Jarred-Sumner, line 223: Did you consider using Buffer.concat? In Bun, we've optimized Buffer.concat a lot, it should be meaningfully faster than this approach.

If for some reason `Buffer.concat` cannot be used, `Buffer.allocUnsafe` should also be faster here - there's no risk of uninitialized memory leaking since the full length is counted and will not change

packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -165,3 +168,119 @@ async function bufferHeadContent(result: SSRResult) {
...
+        done: length === 0,
+        value: mergedArray
+      };
+
+      return returnValue;
+    }
+  };
+
+  const destination: RenderDestination = {
+    write(chunk) {
```
Jarred-Sumner, line 245: Would you be open to a PR that makes this fast in Bun as well? 

It would use `[Symbol.asyncIterator]` and you can yield either `ArrayBufferView` or string literals, so you can skip the encode/decode step (if I'm reading this code correctly). This would make faster and use a lot less memory.

Then the iterator function would essentially be:

```ts
// earlier
let { promise, resolve, reject } = Promise.withResolvers();
let pendingChunks = [];

// inside the iterator
async function* iterator() {
  while (true) {
    await promise;
    ({ promise, resolve, reject } = Promise.withResolvers());
    const chunks = pendingChunks;
    pendingChunks = [];
    yield* chunks;
    if (chunks.length === 0) return;
  }
}

const destination: RenderDestination = {
   write(chunk) {
       // ....
```

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: No I did not know that it would be faster. Would be happy to take a PR on this, we have benchmarks we can compare.

packages/astro/src/runtime/server/render/astro/render.ts
matthewp: Thanks, assuming this would need to conditionally run only in Node we'd like to hold off on that for now. I didn't particularly want to do the change in this PR but the difference was too big to ignore.

mb21: I've not looked into performance, but seems you could even return an [async generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator) of strings directly. I don't see why you would need to construct the async iterator yourself, and why you need to mess around with that buffer at all, but perhaps I'm missing something? Returning the generator would be a lot less code, and the following works in Node.js:

```js
async function* renderToAsyncIterable() {
  yield await Promise.resolve('a');
  yield await Promise.resolve('b');
  yield await Promise.resolve('c');
}

const res = new Response(renderToAsyncIterable())
await res.text()
```

Note that you can also use [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) to iterate over async iterables when needed.

bluwy: In my past refactors, I found that generators tend to be slow. So I think we should avoid it even if it may be easier to read.

mb21: @bluwy right... seems like generators are 1.8-2x slower than doing it manually (at least in node.js), but seems that they scale equally. But that still leaves the question why not to return a `AsyncIterable<string | Uint8Array>` and avoid creating another buffer.

Issue #9617, opened 2024-01-05
Hydration error with Vue component on Astro 4.1.0


maddsua:
### Astro Info

```block
Astro                    v4.1.0
Node                     v20.10.0
System                   Linux (x64)
Package Manager          npm
Output                   hybrid
Adapter                  none
Integrations             @astrojs/vue
```

### If this issue only occurs in one browser, which browser is a problem?

Firefox

### Describe the Bug

Not having `is:global` style tag directive in parent Astro component and assigning any class in children Vue component causes `Hydration class mismatch`

![image](https://github.com/withastro/astro/assets/102395937/8e0958f7-a60d-4646-96ae-36d3bb8cbefb)

### What's the expected result?

No hydration error shall occur

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-gdvsnr

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #1406, opened 2021-09-21
🚀 Astro Next (0.21.0)

drwpow:
# 🚀 Astro Next (0.21.0)

Hello! This PR showcases a huge change to Astro, namely the new Go/WASM compiler, and switch to Vite. So this will work a little differently from most PRs. This will be a **long-running PR that other PRs get merged into** (no pushes directly, please!).

## 👀 For Viewers

If you’re using this to track the status of `astro@next--compiler`

### ✨ Try it today

⚠️ We recommend starting with a new project! Trying to use on an existing Astro project will likely not work until we have full feature parity.

```
npm install astro@next--compiler
```

### <a id="examples-linked">Examples Checklist - Complete, Linked</a>

Status: started

- [x]  **blog** (jn)
- [x]  **blog-multiple-authors** (dp)
- [x]  **docs** (jn)
- [x]  **framework-preact** (mp)
- [x]  **framework-react** (mp)
- [x]  **framework-svelte** (mp)
- [x]  **framework-vue** (mp)
- [x]  **framework-solid** (mp)
- [x]  **framework-lit** (mp)
- [x]  **framework-multiple** (mp)
- [x]  **minimal** (mp)
- [x]  **portfolio** (jn)
- [x]  **starter** (mp)
- [x]  **with-markdown** (nm)
- [x]  **with-markdown-plugins** (mp)
- [x]  **with-nanostores** (mp)

### <a id="examples-unlinked">Examples Checklist - Complete, Installed/Unlinked</a>

Status: finished

- [x] ***shared/general:** (mp) https://github.com/stackblitz/webcontainer-core/issues/335 (also: https://github.com/snowpackjs/astro/pull/1406#issuecomment-930506551)
- [x]  **blog** (jn)
- [x]  **blog-multiple-authors** (jn)
- [x]  **docs** (jn
- [x]  **framework-preact** (nm)
- [x]  **framework-react** (nm)
- [x]  **framework-svelte** (nm)
- [x]  **framework-vue** (nm)
- [x]  **framework-solid** (mp)
- [x]  **framework-lit** (mp)
- [x]  **framework-multiple** (dp)
- [x]  **minimal** (jn)
- [x]  **portfolio** (dp)
- [x]  **starter** (mp)
- [x]  **with-markdown** (dp)
- [x]  **with-markdown-plugins** (nm)
- [x]  **with-nanostores** (dp)

### 🧩 Stability

_Last updated: 2021-10-11_

- [ ] Enable Windows tests
[...]

PR #9620, opened 2024-01-05
Fix class property Vue hydration error

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/9617

The `delete props['class']` was introduced in https://github.com/withastro/astro/pull/1406/commits/1103cb1f9d2a2077f65a0d08aee67ee9bf0e9540, but I'm not sure why. But this should work fine today without the deletion.

## Testing

Existing tests should pass. Also tested this fix with the repro.

## Docs

n/a. bug fix.

### DIFF
--- a/.changeset/dirty-monkeys-bow.md
+++ b/.changeset/dirty-monkeys-bow.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/vue": patch
+---
+
+Fixes `class` property hydration error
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):

maddsua:
I wonder why did it break things now, considering that the commit in question is like 2 years old

matthewp (approved):

PR #9622, opened 2024-01-05
Fix limitInputPixels type

bluwy:
## Changes

ref https://github.com/withastro/astro/pull/9546#issuecomment-1877736238

I mistyped `limitInputPixels`. It should be `number | boolean` as accepted by Sharp as well.

## Testing

n/a. Should still work

## Docs

Updated jsdoc as well.

### DIFF
--- a/packages/astro/src/assets/services/sharp.ts
+++ b/packages/astro/src/assets/services/sharp.ts
@@ -12,7 +12,7 @@ export interface SharpImageServiceConfig {
 	/**
 	 * The `limitInputPixels` option passed to Sharp. See https://sharp.pixelplumbing.com/api-constructor for more information
 	 */
-	limitInputPixels?: number;
+	limitInputPixels?: number | boolean;
 }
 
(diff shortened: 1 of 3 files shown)


packages/astro/src/assets/services/sharp.ts
```diff
@@ -12,7 +12,7 @@ export interface SharpImageServiceConfig {
 	/**
 	 * The `limitInputPixels` option passed to Sharp. See https://sharp.pixelplumbing.com/api-constructor for more information
 	 */
-	limitInputPixels?: number;
+	limitInputPixels?: number | boolean;
```
Princesseuh, line 15: Is there any way we could somehow use the type from Sharp's directly?

bluwy: Good point. For some reason I never thought about it. Updated to use the type from Sharp.

Princesseuh (approved):

PR #9635, opened 2024-01-08
Allow img attributes for Picture component

Apisathan:
## Changes

Adds a option allowing for attributes for the Picture component's img tag.

## Testing

By importing the Picture componet and add the following: 
```
<Picture width="50" height="50" src="/favicon.svg" alt="test" pictureAttributes={{
	class: "picture-class",
}} imgAttributes={{
	class: "image-class",
}}/>
```
Output:
image

## Docs

Yes should be added as a option under: https://docs.astro.build/en/guides/images/#picture-

### DIFF
--- a/.changeset/bright-waves-scream.md
+++ b/.changeset/bright-waves-scream.md
@@ -1,0 +1,5 @@
+---
+"astro": major
+---
+
+Allowing for img attributes for the Picture component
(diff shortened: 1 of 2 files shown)


Apisathan:
Stupid me, found i should just use class on the Picture component itself..

PR #9637, opened 2024-01-08
Improve env var handling in SSR

bluwy:
## Changes

Some env var replacement can be relayed to Vite to replace instead, which is more accurate as we're using plain string replacements. I'm currently working on improving Astro's env replacement too and notice this stop gap improvement.

## Testing

Existing env vars tests should pass.

## Docs

n/a. internal refactor.

### DIFF
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -149,9 +149,9 @@ export async function createVite(
 		root: fileURLToPath(settings.config.root),
 		envPrefix: settings.config.vite?.envPrefix ?? 'PUBLIC_',
 		define: {
-			'import.meta.env.SITE': settings.config.site
-				? JSON.stringify(settings.config.site)
-				: 'undefined',
+			'import.meta.env.SITE': stringifyForDefine(settings.config.site),
+			'import.meta.env.BASE_URL': stringifyForDefine(settings.config.base),
+			'import.meta.env.ASSETS_PREFIX': stringifyForDefine(settings.config.build.assetsPrefix),
 		},
 		server: {
 			hmr:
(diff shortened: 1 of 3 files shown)


packages/astro/src/vite-plugin-env/index.ts
```diff
@@ -48,12 +48,6 @@ function getPrivateEnv(
 			}
 		}
 	}
-	privateEnv.SITE = astroConfig.site ? JSON.stringify(astroConfig.site) : 'undefined';
-	privateEnv.SSR = JSON.stringify(true);
-	privateEnv.BASE_URL = astroConfig.base ? JSON.stringify(astroConfig.base) : 'undefined';
-	privateEnv.ASSETS_PREFIX = astroConfig.build.assetsPrefix
-		? JSON.stringify(astroConfig.build.assetsPrefix)
-		: 'undefined';
```
bluwy, line 56: We don't need to re-specify these here as Vite will already handle them. If we specify them here, we'd do a plain string replacement instead, which is less accurate than Vite's implementation.

packages/astro/src/core/create-vite.ts
```diff
@@ -308,3 +308,7 @@ function isCommonNotAstro(dep: string): boolean {
 		)
 	);
 }
+
+function stringifyForDefine(value: string | undefined): string {
```
ematipico, line 312: Should it be `stringifyOrDefine`?

ematipico (approved):

packages/astro/src/core/create-vite.ts
bluwy: It is "For" since we're stringifying the `value` for the `define` config passed to Vite.

PR #9638, opened 2024-01-08
feat(i18n): disable redirect to default language

ematipico:
## Changes

Closes PLT-1318

This PR adds a new option to the `i18n.routing` object, called `redirectToDefaultLocale`, which comes into play when the option `prefixDefaultLocale` comes into play (it's `true`). 

This new option is `true` by default to maintain the current behaviour of the routing, where Astro does a redirect from `/`  to `/<defaultLocale>`.

When this option is set to `false`, the redirect isn't triggered anymore. 

List of technical changes inside the PR:
- I added a new schema validation, where we throw an error when `redirectToDefaultLocale` is `false` and `prefixDefaultLocale` is `false`. The reason why I added an error is that this combination of options doesn't have any effect, so I want to enforce a sense of correctness when configuring the i18n.
- I refactored some types. Before, the routing name types were copied all over the code base, and it was unmaintainable. I created a type that is now imported everywhere. 
- The core of the logic occurs in the middleware.

## Testing

I added new test cases that should cover DEV, SSG and SSR

## Docs

I will create a PR in the docs repository to update the page.


packages/astro/src/core/config/schema.ts
```diff
@@ -322,17 +325,31 @@ export const AstroConfigSchema = z.object({
...
 						strategy: z.enum(['pathname']).default('pathname'),
 					})
 					.default({})
+					.refine(
+						({ prefixDefaultLocale, redirectToDefaultLocale }) => {
+							return !(prefixDefaultLocale === false && redirectToDefaultLocale === false);
+						},
+						{
+							message:
+								'The option `i18n.redirectToDefaultLocale` has effects only when the `i18n.prefixDefaultLocale` is set to `true`. Remove the option `i18n.redirectToDefaultLocale`, or change the value of `i18n.prefixDefaultLocale` to `true`.',
```
bluwy, line 338:
```suggestion
								'The option `i18n.redirectToDefaultLocale` is only useful when the `i18n.prefixDefaultLocale` is set to `true`. Remove the option `i18n.redirectToDefaultLocale`, or change its value to `true`.',
```

Maybe this is a bit more concise.

packages/astro/src/i18n/middleware.ts
```diff
@@ -54,30 +55,52 @@ export function createI18nMiddleware(
...
+						});
+					}
+					break;
+				}
+
+				case 'pathname-prefix-always-no-redirect': {
+					// We return a 404 if:
+					// - the current path isn't a root. e.g. / or /<base>
+					// - the URL doesn't contain a locale
+					const isRoot = url.pathname.startsWith(base) || url.pathname === '/';
```
bluwy, line 75: In line 86 below, it's using `url.pathname === base + '/' || url.pathname === base` to check the root. Is that more accurate? Since the `startsWith` here could match against deep paths.

packages/astro/src/i18n/middleware.ts
```diff
@@ -54,30 +55,52 @@ export function createI18nMiddleware(
...
 
-				// Astro can't know where the default locale is supposed to be, so it returns a 404 with no content.
-				else if (!pathnameHasLocale(url.pathname, i18n.locales)) {
-					return new Response(null, {
-						status: 404,
-						headers: response.headers,
-					});
+				case 'pathname-prefix-always': {
+					if (url.pathname === base + '/' || url.pathname === base) {
+						if (trailingSlash === 'always') {
```
bluwy, line 87: Unrelated: It looks like the prefix is only added when `trailingSlash` is set to `always`. Shouldn't the prefix also be added for other `trailingSlash` configurations?

packages/astro/src/i18n/middleware.ts
ematipico: Yeah, it makes sense

packages/astro/src/i18n/middleware.ts
ematipico: Right, although to check the other variants, we need `build.format`, which is a piece of information that we don't have in SSR (in the SSR manifest), that's why I only check `always`. Should store `build.format` in the SSRManifest?

.changeset/cyan-grapes-suffer.md
```diff
@@ -0,0 +1,22 @@
+---
+"astro": minor
+---
+
+Adds a new property inside `i18n.routing`, called `redirectToDefaultLocale`.
+
+When `false`, Astro doesn't do a redirect from `/` to `/<defaultLocale>`. This option comes into play only
+when `prefixDefaultLocale` is `true`.
```
sarah11918, line 8:
```suggestion

Adds a new `i18n.routing` config option `redirectToDefaultLocale` to disable automatic redirects of the root URL (`/`) to the default locale when `prefixDefaultLocale: true` is set.

In projects where every route, including the default locale, is prefixed with `/[locale]/` path, this property allows you to control whether or not `src/pages/index.astro` should automatically redirect your site visitors from `/` to `/[defaultLocale]`.

You can now opt out of this automatic redirection by setting `redirectToDefaultLocale: false`:
```
What do you think about something like this, @ematipico ? (Also see my comments in the docs PR to explain some of my choices)

packages/astro/src/@types/astro.ts
```diff
@@ -1493,6 +1493,20 @@ export interface AstroUserConfig {
...
+			 * @docs
+			 * @name i18n.routing.redirectToDefaultLocale
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `true`
+			 * @version 4.2.0
+			 * @description
+			 *
+			 * When `false`, Astro doesn't do a redirect from `/` to `/<defaultLocale>`. This option comes into play only
+			 * when `prefixDefaultLocale` is `true`.
```
sarah11918, line 1506:
``````suggestion
			 * Configures whether or not the home URL (`/`) generated by `src/pages/astro.index`
			 * will redirect to `/[defaultLocale]` when `prefixDefaultLocale: true` is set.
			 * 
			 * Set `redirectToDefaultLocale: false` to disable this automatic redirection at the root of your site:
			 * ```js
			 * // astro.config.mjs
			 *   i18n:{
			 *     defaultLocale: "en",
			 * 		locales: ["en", "fr"],
			 *     routing: {
			 *       prefixDefaultLocale: true,
			 *       redirectToDefaultLocale: false
			 *     }
			 *   }
			 * })
			 *```
``````
We could include a code sample of the config here, too maybe?

**Review by sarah11918 — COMMENTED** [review]
Thanks, @ematipico ! Made some suggestions here, based on the docs PR suggestions I made. See what you think of these, and edit to taste!

File: .changeset/cyan-grapes-suffer.md (hunk 4 above)
Comment (ematipico) [reply .changeset/cyan-grapes-suffer.md:8]: Awesome!

File: .changeset/cyan-grapes-suffer.md [hunk 6]
```diff
@@ -0,0 +1,24 @@
+---
+"astro": minor
+---
+
+
+Adds a new `i18n.routing` config option `redirectToDefaultLocale` to disable automatic redirects of the root URL (`/`) to the default locale when `prefixDefaultLocale: true` is set.
+
+In projects where every route, including the default locale, is prefixed with `/[locale]/` path, this property allows you to control whether or not `src/pages/index.astro` should automatically redirect your site visitors from `/` to `/[defaultLocale]`.
```
sarah11918, line 8:
```suggestion
In projects where every route, including the default locale, is prefixed with a `/[locale]/` path, this property allows you to control whether or not `src/pages/index.astro` should automatically redirect your site visitors from `/` to `/[defaultLocale]`.
```
Sorry, my original typo!

.changeset/long-mangos-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix the passthrough image service not generating `srcset` values properly
```
sarah11918, line 5:
```suggestion
Fixes the passthrough image service not generating `srcset` values properly
```

sarah11918 (approved):
Docs is happy! Looking forward to this, @ematipico !  :partying_face:

packages/astro/src/i18n/middleware.ts
bluwy: Ah ok that makes sense. I think it's fine to store `build.format` then if it helps improve things here. We can definitely improve this in a later PR.

packages/astro/src/@types/astro.ts
```diff
@@ -1493,6 +1493,35 @@ export interface AstroUserConfig {
...
+			/**
+			 * @docs
+			 * @name i18n.routing.redirectToDefaultLocale
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `true`
+			 * @version 4.2.0
+			 * @description
+			 *
+			 * Configures whether or not the home URL (`/`) generated by `src/pages/astro.index`
```
bluwy, line 1505:
```suggestion
			 * Configures whether or not the home URL (`/`) generated by `src/pages/index.astro`
```

bluwy (approved):
Looks great 👍

PR #9642, opened 2024-01-08
Fixes an issue where view transitions to the 404-page did not work

### LINKED ISSUE #9629: View Transition to 404 loses state
### Astro Info

```block
> green-galaxy@0.0.1 astro
> astro info

Astro                    v4.1.1
Node                     v18.14.2
Package Manager          npm
Adapter                  none
```

only tried on Firefox

### Describe the Bug

when I use ViewTransitions and:
1. open a non existing url and get redirected to the 404 page
2. or open the 404 page directly
I lose the state of my `transition:persist` island.

The real life usage is the login state, which I want to maintain also on the 404 page, but in the minimal example I use a simple counter.

A `transition:persist` island should persist its state on ViewTransition also to non existing/404 pages.

https://github.com/aminabromand/astro-404-view-transition-state

### Participation

martrapp:
## Changes

Changed the cloning of a Response in route.ts to retain the original headers.
With the headers the content type of the 404 page was lost.
The view transition loader expected text/html and did a full reload due to the missing header.
Closes #9629 

## Testing

new e2e test to confirm that we can transition to 404 / undefined pages without reloads

## Docs

n.a. / bug fix

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -370,7 +370,11 @@ export async function handleRoute({
 	// Apply the `status` override to the response object before responding.
 	// Response.status is read-only, so a clone is required to override.
 	if (status && response.status !== status && (status === 404 || status === 500)) {
-		response = new Response(response.body, { ...response, status });
+		response = new Response(response.body, {
+			status: status,
+			statusText: status === 404 ? 'Not Found' : 'Internal Server Error',
+			headers: response.headers
+		});
 	}
 	await writeSSRResult(request, response, incomingResponse);
 }
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -372,7 +372,6 @@ export async function handleRoute({
 	if (status && response.status !== status && (status === 404 || status === 500)) {
 		response = new Response(response.body, {
 			status: status,
-			statusText: status === 404 ? 'Not Found' : 'Internal Server Error',
 			headers: response.headers
 		});
 	}
--- a/.changeset/big-knives-own.md
+++ b/.changeset/big-knives-own.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue where view transitions to 404 page did not work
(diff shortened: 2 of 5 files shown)


lilnasy (approved):
Looks great to me! Thanks for adding the simple to understand test as well.

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -370,7 +370,11 @@ export async function handleRoute({
 	// Apply the `status` override to the response object before responding.
 	// Response.status is read-only, so a clone is required to override.
 	if (status && response.status !== status && (status === 404 || status === 500)) {
-		response = new Response(response.body, { ...response, status });
+		response = new Response(response.body, {
+			status: status,
+			statusText: status === 404 ? 'Not Found' : 'Internal Server Error',
+			headers: response.headers
+		});
```
lilnasy, line 377: Makes sense why `...response` has no effect, spreading only gives you enumerable properties, `"headers"` (and all others) are instance properties instead.

```suggestion
		response = new Response(response.body, {
			status: status,
			headers: response.headers
		});
```

This is minor and a little subjective, but we would want to avoid setting `statusText`. It is not part of http 2, so it will be error prone if users depend on it being a certain value.

martrapp: Good point!

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -370,7 +370,10 @@ export async function handleRoute({
 	// Apply the `status` override to the response object before responding.
 	// Response.status is read-only, so a clone is required to override.
 	if (status && response.status !== status && (status === 404 || status === 500)) {
-		response = new Response(response.body, { ...response, status });
+		response = new Response(response.body, {
+			status: status,
+			headers: response.headers
+		});
```
bluwy, line 376: Can confirm that `...response` seems to do nothing here 👍 We could also copy over `statusText`, but not a big deal I think

lilnasy: That was it originally. I advised against it.

https://github.com/withastro/astro/pull/9642#discussion_r1444741971

bluwy: Ah, looks like you're already one step ahead 😄

PR #9645, opened 2024-01-08
fix(vercel): handle error pages using vercel APIs

### LINKED ISSUE #9028: Infinite 404 loop in prod on Vercel
### Astro Info

```block
Astro                    v3.4.4
Node                     v21.1.0
Integrations             none
```

_No response_

### Describe the Bug

When vercel's "cleanUrls" config is enabled and `404.astro` is prerendered - either explicitly or because of hybrid mode - visiting any path not routed would lead to an infinite loop which gets intervened by Vercel's infrastructure.

`404.html` gets served.

https://infinite-rerouting.vercel.app/afsfasbj
https://github.com/lilnasy/vercel-infinite-loop/commit/087871b115227025b76cc3faef2d183d7daf55c6

### Participation

lilnasy:
## Changes
- Fixes #9028
- [x] Introduces an opt-out for implicit rerouting with `app,render(request, { handleErrorPages: false })`
- [ ] Makes vercel perform the rewrites instead.

## Testing
_Pending_

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -32,7 +32,7 @@ export { deserializeManifest } from './common.js';
 const responseSentSymbol = Symbol.for('astro.responseSent');
 
 /**
- * A response with one of these status codes will be rewritten
+ * A response with one of these status codes may be rewritten
  * with the result of rendering the respective error page.
  */
 const REROUTABLE_STATUS_CODES = new Set([404, 500]);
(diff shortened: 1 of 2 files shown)


lilnasy:
Turns out there isn't a vercel API that could be used here.

Configured rewrites can only act if a route hasn't already matched. In this case, the route for the serverless function did.

[`NextResponse.rewrite()`](https://github.com/vercel/next.js/blob/v14.0.4/packages/next/src/server/web/spec-extension/response.ts#L96-L105), which couldve been used to rewrite to `/404` only works within a middleware.

PR #9649, opened 2024-01-09
Use node:test and node:assert/strict

bluwy:
## Changes

For `@astrojs/upgrade` only for now as an experiment. There are quite some quirks with `node:test`:

1. Passing test glob patterns in Node <21 is broken. You can't pass trailing arguments like `--test-name-pattern` with it. https://github.com/nodejs/node/issues/50287
2. Running `.only` tests requires the `--test-only` flag. If you have nested tests, e.g. describe > it, you need both to be `describe.only` and `it.only` :((
3. Test arguments are really long and hard to use/remember

Hence I created a wrapper `astro-scripts test` CLI which should simplify these. Except no2's nested --only issue.

## Testing

Tested manually with node 18.18.1 and node 20.6.1. Node 20 has slightly better test report description.

Performance-wise, running `time pnpm test`:

- Before: `pnpm test  0.89s user 0.21s system 167% cpu 0.659 total`
- After:  `pnpm test  0.54s user 0.08s system 89% cpu 0.687 total`

## Docs

n/a. internal change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,11 +88,23 @@ DEBUG=vite:[name] astro dev   # debug specific process, e.g. "vite:deps" or "vit
 # run this in the top-level project root to run all tests
 pnpm run test
 # run only a few tests in the `astro` package, great for working on a single feature
-# (example - `pnpm run test:match "cli"` runs `cli.test.js`)
+# (example - `pnpm run test:match "cli"` runs tests with "cli" in the name)
 pnpm run test:match "$STRING_MATCH"
 # run tests on another package
 # (example - `pnpm --filter @astrojs/rss run test` runs `packages/astro-rss/test/rss.test.js`)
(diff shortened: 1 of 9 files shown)


Princesseuh (approved):
I'm for this! I think `node:test` is likely to become better "quickly" and will in the future have a fair ecosystem around it, by being the built-in option. I trust in Node 🫡

ematipico (approved):
How would it work if I wanted to run a single test? `node test ../path/to/file.test.js`?

bluwy:
Almost yeah! It's `--test`, so it should be `node --test ../path/to/file.test.js`. It's optional to use `astro-scripts test` since it's just a wrapper around it to workaround quirks and set proper defaults. But you can also if you want to, e.g. `astro-scripts test ../path/to/file.test.js --match "filter test name"`, or shorter `pnpm test -- -m "filter test name"`. I think I'll document this in the contributing guide before merging.

PR #9651, opened 2024-01-09
[docs] fix some error message syntax and add some lint rules to help

sarah11918:
## Changes

This updates a malformed link in an error message that snuck in.

## Testing

No testing!

## Docs

Only updates a link for the docs!

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -117,5 +117,13 @@ module.exports = {
         'no-console': 'off',
       },
     },
+    {
+      files: ['packages/astro/src/core/errors/errors-data.ts'],
+      rules: {
+        // This file is used for docs generation, as such the code need to be in a certain format, we can somewhat ensure this with these rules
+        'object-shorthand': ['error', 'methods', { avoidExplicitReturnArrows: true }],
+        'arrow-body-style': ['error', 'never'],
+      },
+    },
   ],
 };
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -440,7 +440,7 @@ export const NoMatchingImport = {
 export const InvalidPrerenderExport = {
 	name: 'InvalidPrerenderExport',
 	title: 'Invalid prerender export.',
-	message: (prefix: string, suffix: string, isHydridOuput: boolean) => {
+	message(prefix: string, suffix: string, isHydridOuput: boolean) {
 		const defaultExpectedValue = isHydridOuput ? 'false' : 'true';
 		let msg = `A \`prerender\` export has been detected, but its value cannot be statically analyzed.`;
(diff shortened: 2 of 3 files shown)


TheOtterlord (approved):
lgtm

sarah11918:
Realizing that these errors were also added to the bottom of the page, not in a particular section, so waiting for guidance as to where exactly to put them so they are all categorized properly! cc @Princesseuh

florian-lefebvre (approved):

.eslintrc.cjs
```diff
@@ -117,5 +117,13 @@ module.exports = {
         'no-console': 'off',
       },
     },
+    {
+      files: ['packages/astro/src/core/errors/errors-data.ts'],
+      rules: {
+        // This file is used for docs generation, as such the code need to be in a certain format, we can somewhat ensure this with these rules
```
Princesseuh, line 123: Those two rules are unfortunately not perfect, for a real perfect dx, we'd need to take into account the JSDoc comments and stuff. Nonetheless, this prevents `message: (something) => {return something}`, which is what I mostly wanted.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -987,12 +988,51 @@ export const FailedToFindPageMapSSR = {
...
+	name: 'MissingLocaleError',
+	title: 'The provided locale does not exist.',
+	message: (locale: string) =>
+		`The locale/path \`${locale}\` does not exist in the configured \`i18n.locales\`.`,
+} satisfies ErrorData;
+
+/**
+ * @docs
+ * @description
+ * Astro could not find an associated file with content while trying to render the route. This is an Astro error and not a user error.  If restarting the dev server does not fix the problem, please file an issue.
```
sarah11918, line 1007:
```suggestion
 * Astro could not find an associated file with content while trying to render the route. This is an Astro error and not a user error. If restarting the dev server does not fix the problem, please file an issue.
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -987,12 +988,51 @@ export const FailedToFindPageMapSSR = {
...
+	title: "Astro can't render the route.",
+	message:
+		'Astro cannot find any content to render for this route. There is no file or redirect associated with this route.',
+	hint: 'If you expect to find a route here, this may be an Astro bug. Please file an issue/restart the dev server',
+} satisfies ErrorData;
+
+/**
+ * @docs
+ * @description
+ * Astro could not find any code to handle a rejected  `Promise`. Make sure all your promises have an `await` or `.catch()` handler.
```
sarah11918, line 1020:
```suggestion
 * Astro could not find any code to handle a rejected `Promise`. Make sure all your promises have an `await` or `.catch()` handler.
```

Issue #9012, opened 2023-11-06
Builds fail when MDX souremaps are enabled on files with `import.meta.env.*`


bayssmekanique:
### Astro Info

```block
Astro                    v3.4.3
Node                     v18.18.2
System                   Windows (x64)
Package Manager          npm
Output                   server
Adapter                  astro-sst
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         @astrojs/prefetch
                         astro-robots-txt
```

### If this issue only occurs in one browser, which browser is a problem?

n/a

### Describe the Bug

When enabling `sourcemaps` in the `vite.build` configuration with the `MDX` integration, the build will fail if any of the `.mdx` files references environment variables.

Example MDX Files:
```mdx
---
title: 'Using ENVs'
description: 'Lorem ipsum dolor sit amet'
pubDate: 'Oct 25 2023'
heroImage: '/img/placeholder-hero.jpg'
---

export const SECRET_KEY = import.meta.env.SECRET_KEY
export const VITE_KEY = import.meta.env.VITE_KEY
export const PUBLIC_KEY = import.meta.env.PUBLIC_KEY

Let's see what environment variables are available to us:

- `SECRET_KEY`: { SECRET_KEY }
- `VITE_KEY`: { VITE_KEY }
- `PUBLIC_KEY`: { PUBLIC_KEY }
```

Resulting error:
```sh
> @example/basics@0.0.1 dev
> astro dev

  🚀  astro  v3.4.3 started in 300ms
  
  ┃ Local    http://localhost:4321/
  ┃ Network  use --host to expose
  
02:18:22 PM [astro] update /src/pages/test.mdx (x6)
TypeError: Cannot destructure property 'length' of 'code' as it is undefined.
    at State.map (file:///home/projects/github-jmjzta/node_modules/astring/dist/astring.mjs:1166:13)
    at State.writeAndMap (file:///home/projects/github-jmjzta/node_modules/astring/dist/astring.mjs:1123:10)
    at Object.Literal (file:///home/projects/github-jmjzta/node_modules/astring/dist/astring.mjs:1057:13)
    at Object.VariableDeclarator (file:///home/projects/github-jmjzta/node_modules/astring/dist/astring.mjs:530:27)
    at formatVariableDeclaration (file:///home/projects/github-jmjzta/node_modules/astring/dist/astring.mjs:253:15)
[...]

Issue #7503, opened 2023-06-27
Vue SSR - data-v attributes are not stable, due to hash mismatch


ffxsam:
### What version of `astro` are you using?

2.7.1

### Are you using an SSR adapter? If so, which one?

Node.js

### What package manager are you using?

pnpm

### What operating system are you using?

Mac

### What browser are you using?

Chrome

### Describe the Bug

I discovered an issue with the Vue integration when using SSR.

Vue components have a unique + stable `data-v-*` attribute that's used for scoping styles (and maybe other purposes I'm not aware of). If a component references `import.meta.env.SSR`, during Astro's build process, Vite generates two separate `data-v-*` hashes because of the source code differences in a particular component. In one component (the client-side build):

```ts
const test = import.meta.env.SSR;
```

but in the server-side version:
```ts
const test = true;
```

See test reproduction link. Instructions:

```shell
$ pnpm i
$ pnpm build && pnpm preview
```
Then go to `http://localhots:3000` and toggle the component. Notice that the style fails to apply when the component is revealed after being hidden, due to the `data-v-*` attribute change.

### What's the expected result?

Not sure if it's possible, but perhaps the `import.meta.env.SSR` value should not evaluate on build.

### Link to Minimal Reproducible Example

https://github.com/ffxsam/repro-vue-attr-change

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Ah yikes. It makes a lot of sense that this happens! Vite always inlines the values of `import.meta.env.SSR` (and, in fact, any `import.meta.env.*` values).

You mentioned that you found a workaround for this on Discord, so I wanted to include that here for future reference.

```ts
// Create new `is-ssr.ts` file, use stable `isSSR` function everywhere
export const isSSR = () => import.meta.env.SSR
```

natemoo-re:
I'm almost hesitant to mark this as a bug because this is a design limitation of Vite's environment variable processing and we probably won't be revisiting this pattern in the near term.

Maybe this is best left as a thing to document? Alternatively, we could potentially expose an `astro:env` module that exports `BROWSER` and `DEV` for clarity. I'd like to consider that for Astro 3.0

bluwy:
I think this is caused by our env handling instead. We have this:

https://github.com/withastro/astro/blob/1a9b644e20b9ff1e8484aae3e97a9ee2fa81238c/packages/astro/src/vite-plugin-env/index.ts#L88-L91

[...]

PR #9652, opened 2024-01-09
Use esbuild for env replacement

bluwy:
## Changes

fixes https://github.com/withastro/astro/issues/7503
fixes https://github.com/withastro/astro/issues/9012

In dev, we can use a fast path to assign all private env vars to `import.meta.env` on the server-side runtime, which is a lot faster and is what Vite does too.

In build, we use esbuild to perform the replacements so that we don't accidentally replace env vars inside strings, same as Vite.

## Testing

Existing tests should pass. I also added test for the MDX issue, but not for the Vue one. It's a bit harder to test that, but I can confirm it's fixed with the repro manually.

NOTE: The changes in `astro` and `@astrojs/mdx` are mutually exclusive, which means the user can use the latest version of `astro` and older version of `@astrojs/mdx` at the same time just fine. (and vice versa)

## Docs

n/a. Ideally this should be transparent to users.


packages/integrations/mdx/src/index.ts
```diff
@@ -129,7 +129,7 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 										const compiled = await processor.process(vfile);
 
 										return {
-											code: escapeViteEnvReferences(String(compiled.value)),
+											code: String(compiled.value),
```
bluwy, line 132: We can safely remove this in a non breaking way as the plugin calls `fs.readFile`, bypassing Vite's transform. The only continuing env replacement left after this transform is Vite's define, which we know is fixed in Vite 5.

Luckily Astro's env replacements (which didn't work well in plain strings) happen before the `fs.readFile`, so the broken code was thrown away.

packages/astro/src/vite-plugin-env/index.ts
```diff
@@ -29,7 +33,7 @@ function getPrivateEnv(
 	const privateEnv: Record<string, string> = {};
 	for (const key in fullEnv) {
 		// Ignore public env var
-		if (envPrefixes.every((prefix) => !key.startsWith(prefix))) {
+		if (isValidIdentifierRe.test(key) && envPrefixes.every((prefix) => !key.startsWith(prefix))) {
```
bluwy, line 36: Added a valid identifier check because keys could be `ProgramFiles(x64)` which wouldn't work as `import.meta.env.ProgramFiles(x64)`. This worked before because we do plain string replacements, but I don't think people were relying on this to work.

packages/astro/src/vite-plugin-env/index.ts
```diff
@@ -61,71 +65,136 @@ function getReferencedPrivateKeys(source: string, privateEnv: Record<string, any
...
 	return {
 		name: 'astro:vite-plugin-env',
-		enforce: 'pre',
+		config(_, { command }) {
+			isDev = command !== 'build';
+		},
 		configResolved(resolvedConfig) {
 			viteConfig = resolvedConfig;
+
+			// HACK: move ourselves before Vite's define plugin to apply replacements at the right time (before Vite normal plugins)
```
bluwy, line 128: Copied this hack from `@astrojs/mdx`.

packages/integrations/mdx/src/recma-inject-import-meta-env.ts
```diff
@@ -11,7 +11,7 @@ export function recmaInjectImportMetaEnv({
 			if (node.type === 'MemberExpression') {
 				// attempt to get "import.meta.env" variable name
 				const envVarName = getImportMetaEnvVariableName(node);
-				if (typeof envVarName === 'string') {
+				if (typeof envVarName === 'string' && importMetaEnv[envVarName] != null) {
```
bluwy, line 14: This check was the main reason for the error in https://github.com/withastro/astro/issues/9012, since if it's undefined, we accidentally generate a broken AST.

After fixing this, it relies on other fixes in this PR so that it works correctly in runtime.

packages/astro/src/vite-plugin-env/index.ts
```diff
@@ -1,13 +1,17 @@
...
+import { transform } from 'esbuild';
+import MagicString from 'magic-string';
 import type { AstroConfig, AstroSettings } from '../@types/astro.js';
 
 interface EnvPluginOptions {
 	settings: AstroSettings;
 }
 
+const importMetaEnvOnlyRe = /import\.meta\.env\b(?!\.)/;
+const isValidIdentifierRe = /^[_$a-zA-Z][_$a-zA-Z0-9]*$/;
```
ematipico, line 13: Can you add a small comment that explains the regex?

ematipico (approved):

matthewp (approved):

PR #9653, opened 2024-01-09
fix: pin Sharp at old version

### LINKED ISSUE #9345: Cannot build when using Sharp 0.33.0

Princesseuh:
## Changes

Sharp's `0.33.0` is much better in every single regard to the 0.32.x version, however its semver checks are incompatible with ours, causing it to not install properly for a lot of users using outdated versions of Node. This PR pins it at the old version until Astro 5.0, where we'll presumably raise the Node requirements. Oh well.

For reference, Sharp's semver check is `^18.17.0 || ^20.3.0 || >=21.0.0` and ours is `>=18.14.1`

Fix #9345 

## Testing

Tested manually, we were at that version previously so, it definitely does work.

## Docs

N/A

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -179,7 +179,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "0.32.6"
+    "sharp": "^0.32.6"
   },
   "devDependencies": {
     "@astrojs/check": "^0.3.1",
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -179,7 +179,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "0.32.6 || ^0.33.1"
+    "sharp": "0.32.6"
   },
   "devDependencies": {
     "@astrojs/check": "^0.3.1",
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -179,7 +179,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "0.32.6"
+    "sharp": "0.32.6 || ^0.33.1"
   },
   "devDependencies": {
     "@astrojs/check": "^0.3.1",
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -179,7 +179,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "^0.33.1"
+    "sharp": "0.32.6"
   },
   "devDependencies": {
     "@astrojs/check": "^0.3.1",
--- a/.changeset/twenty-stingrays-hear.md
+++ b/.changeset/twenty-stingrays-hear.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Pin Sharp to 0.32.6 until we can raise our semver requirements. To use the latest version of Sharp, you can add it to your project's dependencies.
--- a/examples/starlog/package.json
+++ b/examples/starlog/package.json
@@ -12,6 +12,6 @@
   "dependencies": {
     "astro": "^4.1.1",
     "sass": "^1.69.5",
-    "sharp": "0.32.6"
+    "sharp": "^0.32.6"
   }
 }
--- a/examples/starlog/package.json
+++ b/examples/starlog/package.json
@@ -12,6 +12,6 @@
   "dependencies": {
     "astro": "^4.1.1",
     "sass": "^1.69.5",
-    "sharp": "^0.32.5"
+    "sharp": "0.32.6"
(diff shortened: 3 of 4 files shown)


Princesseuh:
!preview pin-sharp

Princesseuh:
!preview pin-sharp

Princesseuh:
!preview pin-sharp

packages/astro/package.json
```diff
@@ -179,7 +179,7 @@
     "zod": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "^0.33.1"
+    "sharp": "0.32.6"
```
Princesseuh, line 182: This sucks, because users have to do a manual override to use newer versions, but unfortunately, I couldn't figure it out. The semver checks in Sharp cannot really be relaxed, because it (very understandably so) depends on the newer Node API 9. This is a shame because `0.33.0` is otherwise a very very good upgrade making installation much much less troublesome.

Now that Node 20 is the LTS, it's likely that when we'll release 5.0 a majority of our users will be on it, which should make this much easier. In 5.0 we can update our requirements to be `^18.17.0 || >= 20.3.0` or similar, which should cover everything + be understandable by our users.

bluwy, line 182: Is the caret (`^`) removed here intentional? (Same as for the starlog example update). `^0.32.6` should still restrict within 0.32.x if a new patch is ever released in the future.

ematipico: I think it's intentional because if users have a version of Node.js lower than the one requested by sharp 0.33.0, it won't work. That's why Erika said that users are forced to install and override their locale version.

ematipico (approved):
You did great Erika, let's merge it and get over it

packages/astro/package.json
bluwy: Even with the caret, it won't use 0.33.0 by default. Users have to override to use 0.33.0 too, so I'm not sure if we need to omit it.

Princesseuh: Ah no, it wasn't necessarily intentional, but Sharp never goes back and fixes previous versions, especially now that 0.33.0 completely changed how the installation works. 

Happy to change it just to be safe, though

bluwy: Thanks for changing it 👍 I think in case there's a security fix it would be nice to unblock it.

bluwy (approved):

Issue #9568, opened 2024-01-02
custom status code does not seem to be respected in configured redirects


JulioGrajales:
### Astro Info

```block
Astro                    v4.0.8
Node                     v21.5.0
System                   Windows (x64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/tailwind
                         @astrojs/svelte
```

### Describe the Bug

[Configured redirects](https://docs.astro.build/en/core-concepts/routing/#configured-redirects) don't seem to respect a custom status code when provided using an object. using the following configuration: 
```js
redirects: {
    '/hello': {
      status: 302,
      destination: '/other',
    },
  },
```

does not use 302 status code and instead uses 301 status code (the default). I'm in SSR mode, using node adapter.
![redirect-astro](https://github.com/withastro/astro/assets/29667371/bd3bb64b-bddc-4fee-aaf4-90a8225c5f79)

### What's the expected result?

It should redirect me using 302 status code or any other valid custom status code provided in the configured redirect object.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-9cfd3y?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

florian-lefebvre:
An interesting to check would be to see if it's node specific or if it also happens with other adapters like vercel

ematipico:
> An interesting thing to check would be to see if it's node specific or if it also happens with other adapters like vercel

I can replicate it in dev using the stackblitz example, so it's probably an issue in core

Runeloon:
Appears this fix doesn't work in the latest 4.2.4 release. Bit difficult to work out why, as the create.ts appears to have gone through a bit of a rework.
Have verified at least that upon hitting redirectRouteStatus, the redirectRoute.redirectRoute property is null, meaning it never bothers checking the redirectRoute.redirect.status, which is set.
Testing example of input to redirectRouteStatus in 4.2.4:
`{ type: 'redirect', isIndex: false, route: '/test1', pattern: /^\/test1\/?$/, segments: [ [ [Object] ] ],       params: [], component: '/test1', generate: [Function (anonymous)], pathname: '/test1', prerender: false, redirect: { status: 302, destination: '/test2' }, redirectRoute: undefined, fallbackRoutes: [] }`

ematipico:
@Runeloon can you provide a reproduction?

lilnasy:
[...]

PR #9657, opened 2024-01-10
fix(redirects): correctly compute the status code in redirects

ematipico:
## Changes
Closes PLT-1371
Closes https://github.com/withastro/astro/issues/9568

## Testing

I tested it locally with the use case provided

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/core/redirects/helpers.ts
+++ b/packages/astro/src/core/redirects/helpers.ts
@@ -36,8 +36,8 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 
 export function redirectRouteStatus(redirectRoute: RouteData, method = 'GET'): ValidRedirectStatus {
 	const routeData = redirectRoute.redirectRoute;
-	if (typeof routeData?.redirect === 'object') {
-		return routeData.redirect.status;
+	if (routeData && typeof redirectRoute.redirect === 'object') {
+		return redirectRoute.redirect.status;
 	} else if (method !== 'GET') {
 		return 308;
 	}
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -477,7 +477,13 @@ export function createRouteManifest(
 			pathname: pathname || void 0,
 			prerender: false,
 			redirect: to,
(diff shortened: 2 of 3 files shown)


packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -477,7 +477,13 @@ export function createRouteManifest(
 			pathname: pathname || void 0,
 			prerender: false,
 			redirect: to,
-			redirectRoute: routes.find((r) => r.route === to),
+			redirectRoute: routes.find((r) => {
+				if (typeof to === 'object') {
+					return r.route === to.destination;
+				} else {
+					return r.route === to;
+				}
```
ematipico, line 485: We forgot to find the route in case `to` is the object

packages/astro/src/core/redirects/helpers.ts
```diff
@@ -36,8 +36,8 @@ export function redirectRouteGenerate(redirectRoute: RouteData, data: Params): s
 
 export function redirectRouteStatus(redirectRoute: RouteData, method = 'GET'): ValidRedirectStatus {
 	const routeData = redirectRoute.redirectRoute;
-	if (typeof routeData?.redirect === 'object') {
-		return routeData.redirect.status;
+	if (routeData && typeof redirectRoute.redirect === 'object') {
+		return redirectRoute.redirect.status;
```
ematipico, line 40: We were checking the incorrect data

bluwy (approved):

Runeloon:
Appears this fix doesn't work in the latest 4.2.4 release. Bit difficult to work out why, as the create.ts appears to have gone through a bit of a rework.
Have verified at least that upon hitting redirectRouteStatus, the redirectRoute.redirectRoute property is null, meaning it never bothers checking the redirectRoute.redirect.status, which is set.
Testing example of input to redirectRouteStatus in 4.2.4:
`{
  type: 'redirect',
  isIndex: false,
  route: '/test1',
  pattern: /^\/test1\/?$/,
  segments: [ [ [Object] ] ],      
  params: [],
  component: '/test1',
  generate: [Function (anonymous)],
  pathname: '/test1',
  prerender: false,
  redirect: { status: 302, destination: '/test2' },
  redirectRoute: undefined,
  fallbackRoutes: []
}`

Issue #9133, opened 2023-11-19
Port conflict


cdtut:
### Astro Info

```block
Node                     v20.8.0
System                   Linux (x64)
```

### Describe the Bug

Start `astro dev` and it start on 4321. Start `astro dev` again and it start on next available port 4322.

Run `astro preview` and there is error `EADDRINUSE: address already in use ::1:4321`.

Run `node dist/server/entry.mjs` and it say `[@astrojs/node] Server listening on http://127.0.0.1:4321` even though process already running on 4321.

### What's the expected result?

1. `astro preview` should open on next available port 4322.
2. `node dist/server/entry.mjs` should have error `EADDRINUSE: address already in use ::1:4321`.
3. `astro preview` should have command line option `--port` and astro config file option to change port to try first.
4. node adapter should have host and port options that are used as default if environment variable not given. Not every build process can provide environment variables (https://docs.astro.build/en/guides/integrations-guide/node/#runtime-environment-variables).

```
adapter: node({
    host: "0.0.0.0",
    port: "4321"
})
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-1o5lsw

### Participation

- [ ] I am willing to submit a pull request for this issue.

cdtut:
There is `server.host` and `server.port` which is used by dev server and node adapter. `host` and `port` should be possible to pass to node adapter to be different from dev which is needed if running dev and build together.

Another problem is passing `host: "0.0.0.0"` prints `[@astrojs/node] Server listening on http://localhost:4321` but should print `0.0.0.0` instead of `localhost`. To be consistent with how it's printed in other places `Error: listen EADDRINUSE: address already in use 0.0.0.0:4321`.

matthewp:
@bluwy surprising to me, we are fully using Vite for preview server now, right?

bluwy:
We use Vite's preview server for `astro preview` only (mostly for static sites). The nodejs adapter's preview server is rather simple and doesn't handle it specially, but maybe we're swallowing the `EADDRINUSE` error somewhere.

bluwy:
> 1. `astro preview` should open on next available port 4322.

When using an adapter, `astro preview` would route to their implementation, and it's not always possible to expect them to auto-increment the port if it's already occupied. So I don't think this behaviour should be expected by default.

> 2. `node dist/server/entry.mjs` should have error `EADDRINUSE: address already in use ::1:4321`.

[...]

PR #9661, opened 2024-01-11
Adapter enhancements

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/9133

## Docs
- https://github.com/withastro/docs/pull/6315


sarah11918 (approved):
For the sake of completeness, approving the changeset for docs as long as the little tense grammar nit is fixed, one way or the other!

porl:
For some reason when I use this merged version `8.0.0` rather than `7.0.4` I get an error on my docker server with Traefik of `bad gateway`. Rolling back to `7.0.4`  works as expected. No errors in any logs I can find and the node server claims it is running fine on the correct IP and port. Running the `node` server on my local dev machine has no problems, so I assume it is something specific to the way Traefik tries to access the port.

Not sure how I can replicate this outside of my production server I'm afraid, so I don't know how helpful this is.

Issue #9370, opened 2023-12-07
HMR error after astro v4 upgrade


abolajibisiriyu:
### Astro Info

```block
Astro                    v4.0.3
Node                     v18.17.1
System                   macOS (x64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
```

### Describe the Bug

- I upgraded to astro v4
- I made changes to my `Footer` component (referenced in `Layout`); no HMR happened
- Then manually saved(`CMD` + `S`) `Layout` file (I did this to trigger HMR); got this behaviour 

`console`
<img width="1387" alt="Screenshot 2023-12-08 at 00 06 03" src="https://github.com/withastro/astro/assets/8493924/cd7538ed-945c-4b0d-a675-f1ae6be930cb">

`browser`
<img width="1678" alt="Screenshot 2023-12-08 at 00 06 23" src="https://github.com/withastro/astro/assets/8493924/3433c772-402d-4a2b-8874-459845480402">

The link to the example(basic astro v4) doesn't seem to throw any errors, but doesn't reflect the changes(💡 making changes to `Footer`)

### What's the expected result?

I expected my changes to reflect in the browser with no errors

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ztyzx4?file=src%2Flayouts%2FLayout.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

jacobthesheep:
Yeah, is anyone else noticing this? I have to refresh the page to see changes.

jtag05:
Also experiencing this issue.

natemoo-re:
Thanks for the report! Presumably this is an issue affecting any Vite 5 users since Astro's HMR logic hasn't changed at all. We'll take a look at this as soon as we can!

natemoo-re:
I was able to reliably reproduce this and track it to one spot in our code! I've never had such luck with an HMR issue before. Hoping to get this fixed ASAP!

jacobthesheep:
Does have anything to do with `data-astro-source-file` on all the elements in dev mode?

bluwy:
> Does this have anything to do with `data-astro-source-file` on all the elements in dev mode?

No, it's related to how Vite 5 handles HMR with circular imports as Tailwind tends to do that :(

---

[...]

PR #9664, opened 2024-01-11
Add fallback compile for astro script and style load

bluwy:
## Changes

Helps https://github.com/withastro/astro/issues/9370 but does not fix it yet. This should reduce the occasional parsing error due to missing Astro compilation cache for its scripts and styles modules, which can often happen in HMR.

It does not fix the issue due to a Vite bug explained in https://github.com/withastro/astro/issues/9370#issuecomment-1887366837. I'm still trying to figure that out, but if I can't, I can probably add a workaround in a later PR too.

## Testing

Tested manually with the issue's repro. I didn't add a test as the bug still exists, but it should not break other tests.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/vite-plugin-astro/index.ts
+++ b/packages/astro/src/vite-plugin-astro/index.ts
@@ -27,6 +27,7 @@ interface AstroPluginOptions {
 export default function astro({ settings, logger }: AstroPluginOptions): vite.Plugin[] {
 	const { config } = settings;
 	let resolvedConfig: vite.ResolvedConfig;
+	let server: vite.ViteDevServer;
 
 	// Variables for determining if an id starts with /src...
 	const srcRootWeb = config.srcDir.pathname.slice(config.root.pathname.length - 1);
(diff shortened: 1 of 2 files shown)


packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -46,9 +50,18 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
 			}
 			// For CSS / hoisted scripts, the main Astro module should already be cached
 			const filename = normalizePath(normalizeFilename(parsedId.filename, config.root));
-			const compileResult = getCachedCompileResult(config, filename);
+			let compileResult = getCachedCompileResult(config, filename);
 			if (!compileResult) {
-				return null;
+				// In dev, HMR could cause this compile result to be empty, try to load it first
+				if (server) {
+					await server.transformRequest('/@fs' + filename);
```
matthewp, line 57: this is inside load, this doesn't cause a loop?

bluwy: The `id` of this `load` is something like `/Users/bjorn/project/src/pages/index.astro/?astro&lang.css`, but the filename here would be `/Users/bjorn/project/src/pages/index.astro`. So it won't cause a loop and should only transform the main Astro file so we have the cache ready.

ematipico (approved):

matthewp (approved):

PR #9666, opened 2024-01-11
Force re-execution of Partytown's head snippet on view transitions

martrapp:
## Changes

Just before the view transition swap(), remove the Partytown `<head>` snippet.
This forces a re-execution on the target page.

## Testing

Manually tested

## Docs
n.a. / bug fix

### DIFF
--- a/packages/integrations/partytown/src/index.ts
+++ b/packages/integrations/partytown/src/index.ts
@@ -21,7 +21,7 @@ export default function createPlugin(options?: PartytownOptions): AstroIntegrati
 	let partytownSnippetHtml: string;
 	const partytownEntrypoint = resolve('@builder.io/partytown/package.json');
 	const partytownLibDirectory = path.resolve(partytownEntrypoint, '../lib');
-	const SELF_DESTRUCT_ON_VIEW_TRANSITION = `;((d,s)=>(s=d.currentScript,d.addEventListener('astro:before-swap',()=>s?.remove(),{once:true})))(document);`
+	const SELF_DESTRUCT_ON_VIEW_TRANSITION = `;((d,s)=>(s=d.currentScript,d.addEventListener('astro:before-swap',()=>s.remove(),{once:true})))(document);`
 	return {
 		name: '@astrojs/partytown',
 		hooks: {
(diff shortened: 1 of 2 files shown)


matthewp:
Looks good!

matthewp (approved):

matthewp:
Does this fix an open issue?

packages/integrations/partytown/src/index.ts
```diff
@@ -21,6 +21,7 @@ export default function createPlugin(options?: PartytownOptions): AstroIntegrati
 	let partytownSnippetHtml: string;
 	const partytownEntrypoint = resolve('@builder.io/partytown/package.json');
 	const partytownLibDirectory = path.resolve(partytownEntrypoint, '../lib');
+	const SELF_DESTRUCT_ON_VIEW_TRANSITION = `;((d,s)=>(s=d.currentScript,d.addEventListener('astro:before-swap',()=>s?.remove(),{once:true})))(document);`
```
lilnasy, line 24: The logic seems correct. Is this an IIFE to save on code size?

martrapp:
> Does this fix an open issue?

@OliverSpeir had an inspiring (pun intended) example which got stuck in view transitions or didn't work at all once view transitions were enabled. This, and the discussions with him, helped a lot as minimal reproduction, but to my knowledge, we currently do not have an official issue.

packages/integrations/partytown/src/index.ts
martrapp: Yes, I started with Typescript, chrunched it a bit and then had fun with it. Mainly it fits the style of the original Partytown snippet.

martrapp: Also I needed a var to save currentScript for later. Always better than storing in window.

lilnasy: It's pretty neat!

lilnasy (approved):
Bit of a hack but that's okay because that's what partytown itself is.
Looks great to me!

martrapp:
> Bit of a hack but that's okay because that's what partytown itself is.

I have not yet found another part for Code Golf in the codebase. 
In the end, we all prefer to have maintainable code ;-)

PR #9671, opened 2024-01-11
[Content collections] Remove "unsupported file type" warning

bholmesdev:
We decided to remove the "unsupported file type" warning from Content Collections [during our roadmap discussions](https://github.com/withastro/roadmap/discussions/805#discussioncomment-8006117). This requirement led to several bugs and special cases overtime, including an "ignore list" for `.DS_Store` and special handling for image assets.

Now, all files that are _not_ content (`.md`) or data (`.json`) are ignored by default. This means underscores are no longer required for these files. You can still use underscores to hide content or data entries from collection queries.

## Changes

- Remove `unsupported` file type from warning logs and type-generation internals

## Testing

- Update tests that expect `unsupported` to now expect `ignored`
- Remove one-off tests for `DS_Store` and image assets

### DIFF
--- a/.changeset/itchy-clouds-invite.md
+++ b/.changeset/itchy-clouds-invite.md
@@ -1,0 +1,14 @@
+---
+"astro": minor
+---
+
+Remove underscore prefix requirement for assets inside content collections. For colocated files like `.js` or `.css`, you can now remove underscores without raising a warning:
+
+```diff
+src/content/blog/
+post.mdx
+- _styles.css
+- _Component.astro
++ styles.css
++ Component.astro
+```
--- a/.changeset/itchy-clouds-invite.md
+++ b/.changeset/itchy-clouds-invite.md
@@ -2,7 +2,7 @@
 "astro": minor
 ---
 
-Remove underscore prefix requirement for assets inside content collections. For extensions like `.astro` or `.css`, you can now remove underscores without raising a warning:
+Removes the requirement for non-content files and assets inside content collections to be prefixed with an underscore. For files with extensions like `.astro` or `.css`, you can now remove underscores without seeing a warning in the terminal.
 
 ```diff
 src/content/blog/
--- a/.changeset/itchy-clouds-invite.md
+++ b/.changeset/itchy-clouds-invite.md
@@ -12,3 +12,5 @@ post.mdx
 + styles.css
 + Component.astro
 ```
+
+Continue to use underscores in your content collections to exclude individual content files, such as drafts, from the build output.
--- a/.changeset/itchy-clouds-invite.md
+++ b/.changeset/itchy-clouds-invite.md
@@ -2,7 +2,7 @@
 "astro": minor
 ---
 
-Remove underscore prefix requirement for assets inside content collections. For colocated files like `.js` or `.css`, you can now remove underscores without raising a warning:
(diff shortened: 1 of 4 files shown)


.changeset/itchy-clouds-invite.md
```diff
@@ -0,0 +1,14 @@
+---
+"astro": minor
+---
+
+Remove underscore prefix requirement for assets inside content collections. For extensions like `.astro` or `.css`, you can now remove underscores without raising a warning:
```
sarah11918, line 5:
```suggestion
Removes the requirement for non-content files and assets inside content collections to be prefixed with an underscore. For files with extensions like `.astro` or `.css`, you can now remove underscores without seeing a warning in the terminal.
```
For a minor, I'm not sure we need to be quite so terse here, and this gives more words for the people writing blog posts.  :smile:  I would go for "removes the requirement for" to frontload the *nature of the change* for people scanning.

(e.g. "Remove underscore..." could trigger, "Wait, are they removing support for using underscores to prevent building? Do I need to remove my underscores and do something different now?")

.changeset/itchy-clouds-invite.md
```diff
@@ -0,0 +1,14 @@
...
+
+```diff
+src/content/blog/
+post.mdx
+- _styles.css
+- _Component.astro
++ styles.css
++ Component.astro
+```
```
sarah11918, line 14:
``````suggestion
```

Continue to use underscores in your content collections to exclude individual content files, such as drafts, from the build output.
``````
Totally not necessary, BUT, thinking of beefing this up for the blog post, and for adding a tiny bit more context/reassurance for a minor release, this is a friendly, helpful reminder of when you still *do* want/need underscores. 

Your call!

sarah11918 (commented):
Yay! Makin' people happy, @bholmesdev !  :partying_face: 

Left a couple of thoughts on the changeset, and I'll just remind you of [what we say in docs](https://docs.astro.build/en/guides/content-collections/#what-are-content-collections) in case you think we should amend anything there: 

> A collection entry is any piece of content stored inside of your content collection directory. Entries can use content authoring formats including Markdown (.md) and MDX (.mdx using the [MDX integration](https://docs.astro.build/en/guides/integrations-guide/mdx/)) or as one of two supported data formats: YAML (.yaml) and JSON (.json). We recommend using a consistent naming scheme (lower-case, dashes instead of spaces) for your files to make it easier to find and organize your content, but this is not required. You can also [exclude entries from being built](https://docs.astro.build/en/core-concepts/routing/#excluding-pages) by prefixing the filename with an underscore (_).

We make it sound like you can't even add non-content files at all, so no need to "correct" the need for underscore. But, if a lot of people are doing this, we could possibly choose to mention this pattern if you think it makes collections more compelling.

.changeset/itchy-clouds-invite.md
bholmesdev: Love it! Happy to add

bholmesdev:
@sarah11918 Thanks for the edits! Yes, I noticed that paragraph scanning our docs. I think we're okay leaving the docs as-is, and letting users discover asset colocation naturally. I expect most who want it will just... try it! And see that it works as expected 😄

sarah11918 (approved):
Docs happy with changeset! Signing off!

PR #9672, opened 2024-01-11
feat(transitions router): data-astro-transition

### LINKED ISSUE #9670: Web component links do not trigger view transitions
### Astro Info

```block
Package Manager          npm
Adapter                  none
                         @astrojs/lit
                         @astrojs/mdx
                         @astrojs/react
```

Any

### Describe the Bug

Because Astro View Transitions only listen for navigation events on `a` tags, they flat out do not work with web components that either act like a link/contain internal `a` tags.

Example:

Say we have a custom button web component that renders an internal `a` tag when an `href` prop is passed.

```html
<my-custom-button href="/some/link"></my-custom-button>
```

This element will never trigger a navigation event in Astro with view transitions. When clicked it will always trigger a full load of the new page.

---

I included a stackblitz that showcases this behavior. It also showcases that when a view-transition does occur via a normal `a` tag element, the web components don't seem to hydrate properly (and thus are not visible)? I'm not sure what is going on there, but it could be a separate issue.

Perhaps we should be able to specify a custom selector that is used to locate elements from which click events should be treated like navigation events. The assumption would be they must have an `href` attribute like `a` tags do. 

Something like this in the config would work wonderfully:
```typescript
export default defineConfig({
  // ...
  viewTransitions: {
    linkSelector: 'a, [href]'
  }
});
```

Really this same issue prevents page prefetching from working for the same reasons, so if there was a more generic setting for this beyond just view transitions that would be even better.

https://stackblitz.com/edit/withastro-astro-cwuduu?file=src%2Fpages%2Findex.astro&title=Astro%20Starter%20Kit:%20Blog

### Participation

lilnasy:
## Changes

- Closes #9670
- Introduces `data-astro-transition` to allow any element to opt-in to view transitions.
- The element must emit `click` event.
- The element must have an `href` attribute.

## Testing

image

## Docs
_Pending_


lilnasy:
!preview data-astro-transition

lilnasy:
@reecemcd Could you try it out?

reecemcd:
> @reecemcd Could you try it out?

Sure thing, I should have time to test this out in the next 2-3 days. Will report back results. Thanks!

reecemcd:
Appears to be working as intended now in our project!

Works well in the [repro stackblitz](https://stackblitz.com/edit/withastro-astro-cwuduu?file=src%2Fpages%2Findex.astro&title=Astro%20Starter%20Kit:%20Blog) as well now too. However, the hydrate issue of some sort still persists there when the normal link is used for some reason? That seems like a separate issue we can open up if we run into that elsewhere. This specific change looks good to go from our perspective.

lilnasy:
- Closing in favor of #9685

PR #9675, opened 2024-01-12
[docs] Update Current Example for `site:` in the Configuration Reference

VoxelMC:
## Changes

**Docs Update**
The currently displayed setting for the `site:` configuration property does not contain a trailing slash, despite the change to this behaviour in `v2` (https://github.com/withastro/astro/issues/5604, https://github.com/withastro/astro/pull/5608).

This closes https://github.com/withastro/docs/issues/6255.

## Testing

This is only a docs change

## Docs

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -563,7 +563,7 @@ export interface AstroUserConfig {
 	 *
 	 * ```js
 	 * {
-	 *   site: 'https://www.my-site.dev'
+	 *   site: 'https://www.my-site.dev/'
 	 * }
 	 * ```
 	 */


florian-lefebvre (approved):

bluwy (changes requested):
I don't think we need to change this. The suggestion to add a trailing slash was because if you want to keep compat like before, you can add a trailing slash to do so. For new projects, you can choose to have a trailing slash or not.

ronan-smith:
@bluwy I don't believe the issue can be solved by adding or removing the trailing slash, however,the documentation shows an example without one but the language in the migration guide suggests that one should be added.

Surely there should be some additional guidance in the documentation regarding the implications of including/excluding the trailing slash?

bluwy:
The only implication if you have or not have the trailing slash is that the value of `import.meta.env.SITE` is with or without the trailing slash that you specified. There aren't any downsides of using either that's worth mentioning IMO.

For new projects, you wouldn't be reading the v2 migration guide either, so I don't think it is confusing. The suggestions in the v2 migration guide are only for projects migrating to v2, not a general suggestion for all projects.

ronan-smith:
@bluwy Thank you for taking the time to explain.

ematipico:
@bluwy should we close this PR? I don't understand what are the changes you requested from OP

bluwy:
Yeah we can close this 👍

PR #9678, opened 2024-01-12
fix(i18n): emit an error when the index isn't found

ematipico:
## Changes

We've had more than one report where users didn't add a index page when they use the option `i18n.routing.prefixDefaultLocale` to `true`.

The index page is required to create a redirect from `/` to `/<defaultLocale>`.

This PR adds a check during the build, and emits an error if the index is missing.

## Testing

There was a case where a fixture didn't have the index, and the test cases were still passing. 
These fixtures were now raising an error, and I added the index page.

## Docs

### DIFF
--- a/.changeset/proud-guests-bake.md
+++ b/.changeset/proud-guests-bake.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Adds an error during the build phase incase `i18n.routing.prefixDefaultLocale` is set to `true` and the index page is missing.
+Adds an error during the build phase in case `i18n.routing.prefixDefaultLocale` is set to `true` and the index page is missing.
--- a/.changeset/proud-guests-bake.md
+++ b/.changeset/proud-guests-bake.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Adds an error during the build phase incase `i18n.routing.prefixDefaultLocale` is set to `true` and the index page is missing.
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -1003,7 +1003,7 @@ export const MissingLocale = {
 
 export const MissingIndexForInternationalization = {
 	name: 'MissingIndexForInternationalizationError',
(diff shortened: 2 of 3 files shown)


.changeset/proud-guests-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Adds an error during the build phase incase `i18n.routing.prefixDefaultLocale` is set to `true` and the index page is missing.
```
sarah11918, line 5:
```suggestion
Adds an error during the build phase in case `i18n.routing.prefixDefaultLocale` is set to `true` and the index page is missing.
```

sarah11918 (approved):
Docs is happy with the error message! (c @Princesseuh to confirm the syntax/formatting!)

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1001,6 +1001,13 @@ export const MissingLocale = {
 		`The locale/path \`${locale}\` does not exist in the configured \`i18n.locales\`.`,
 } satisfies ErrorData;
 
+export const MissingIndexForInternalisation = {
```
florian-lefebvre, line 1004: Not specifically related to this PR but shouldn't it be internationalization? AFAIK z is more American and s more British. I'm quite curious!

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1001,6 +1001,13 @@ export const MissingLocale = {
 		`The locale/path \`${locale}\` does not exist in the configured \`i18n.locales\`.`,
 } satisfies ErrorData;
 
+export const MissingIndexForInternationalization = {
+	name: 'MissingIndexForInternationalizationError',
+	title: 'Index URL not found.',
```
florian-lefebvre, line 1006:
```suggestion
	title: 'Index page not found.',
```
Since it's the index, it's always `/` so maybe saying "page" is less confusing?

Princesseuh (approved):
Format looks great!

MoustaphaDev (approved):

PR #9680, opened 2024-01-12
Fix for #9673

### LINKED ISSUE #9673: Wrong import path for types.d.ts ContentConfig
### Astro Info

```block
Node                     v21.5.0
Package Manager          pnpm
Adapter                  @astrojs/vercel
                         @astrojs/tailwind
                         astro-expressive-code
```

_No response_

### Describe the Bug

The types generated from `packages/astro/src/content/types-generator.ts` generate a `ContentConfig` type that looks like this:

```ts
type ContentConfig = typeof import("../src/content/config");
```

The issue with it is that if our `tsconfig.json` it has `verbatimModuleSyntax` set to `true`, that `import` is invalid because it doesn't have the file extension, so we end up with `any` instead of the strong type we wanted. If instead, that `import` was written like this:

```ts
type ContentConfig = typeof import("../src/content/config.js");
```

Then it would work in all cases.

Get the proper type for my content, right now I'm getting `any` for it.

https://stackblitz.com/edit/verbatim-module-issue?file=src%2Fpages%2Frss.xml.js

### Participation

loucyx:
## Changes

Closes #9673

-   Added a `.js` to the end of the `configPath` for types, making it work for verbatim modules in TypeScript.
-   I also moved the logic to its own function, making it more readable.

## Testing

Ran all the tests with no errors reported. I didn't add tests because I didn't think they were needed, but if you think otherwise, I'll need guidance on testing different TSConfigs on a given file.

## Docs

Docs don't need to be updated. This is a fix for users with verbatim modules in TypeScript, but it works the same for users without that setting.

### DIFF
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -362,6 +362,28 @@ function invalidateVirtualMod(viteServer: ViteDevServer) {
 	viteServer.moduleGraph.invalidateModule(virtualMod);
 }
 
+/**
+ * Takes a configPath and returns a normalized relative version:
+ *  -   If is not relative, it adds `./` to the beginning.
+ *  -   If it ends with `.ts`, it removes the extension.
+ *  -   It adds `.js` to the end.
+ *  -   It stringifies the result (adds `""` around it).
+ * @param from Config path from path.
+ * @param to Config path to path.
+ * @returns Normalized config path.
+ */
+function normalizeConfigPath(from: string, to: string) {
+	const normalizedConfigPath = path.relative(from, to);
+
+	return JSON.stringify(
+		`${isRelativePath(normalizedConfigPath) ? '' : './'}${
+			normalizedConfigPath.endsWith('.ts')
+				? normalizedConfigPath.replace(/\.ts$/, '')
+				: normalizedConfigPath
+		}.js`
+	);
+}
+
 async function writeContentFiles({
 	fs,
 	contentPaths,
@@ -444,17 +466,10 @@ async function writeContentFiles({
 		fs.mkdirSync(contentPaths.cacheDir, { recursive: true });
 	}
 
-	let configPathRelativeToCacheDir = normalizePath(
-		path.relative(contentPaths.cacheDir.pathname, contentPaths.config.url.pathname)
+	const configPathRelativeToCacheDir = normalizeConfigPath(
+		contentPaths.cacheDir.pathname,
+		contentPaths.config.url.pathname
 	);
-
-	if (!isRelativePath(configPathRelativeToCacheDir))
-		configPathRelativeToCacheDir = './' + configPathRelativeToCacheDir;
-
-	// Remove `.ts` from import path
-	if (configPathRelativeToCacheDir.endsWith('.ts')) {
-		configPathRelativeToCacheDir = configPathRelativeToCacheDir.replace(/\.ts$/, '');
-	}
 
 	for (const contentEntryType of contentEntryTypes) {
 		if (contentEntryType.contentModuleTypes) {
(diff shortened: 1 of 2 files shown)


packages/astro/src/content/types-generator.ts
```diff
@@ -362,6 +362,28 @@ function invalidateVirtualMod(viteServer: ViteDevServer) {
...
+ * @param to Config path to path.
+ * @returns Normalized config path.
+ */
+function normalizeConfigPath(from: string, to: string) {
+	const normalizedConfigPath = path.relative(from, to);
+
+	return JSON.stringify(
+		`${isRelativePath(normalizedConfigPath) ? '' : './'}${
+			normalizedConfigPath.endsWith('.ts')
+				? normalizedConfigPath.replace(/\.ts$/, '')
```
loucyx, line 381: If this always resolves to a file that ends with `.ts` or `.js` then we can simplify `normalizeConfigPath` a bit by moving that `.js` at the end of the returned string here, so the line ends up being:

```ts
? normalizedConfigPath.replace(/\.ts$/, ".js")
```

packages/astro/src/content/types-generator.ts
```diff
@@ -362,6 +362,28 @@ function invalidateVirtualMod(viteServer: ViteDevServer) {
...
+ *  -   If is not relative, it adds `./` to the beginning.
+ *  -   If it ends with `.ts`, it removes the extension.
+ *  -   It adds `.js` to the end.
+ *  -   It stringifies the result (adds `""` around it).
+ * @param from Config path from path.
+ * @param to Config path to path.
+ * @returns Normalized config path.
+ */
+function normalizeConfigPath(from: string, to: string) {
+	const normalizedConfigPath = path.relative(from, to);
```
ematipico, line 376: nit: I would call this variable `relativeConfigPath`

packages/astro/src/content/types-generator.ts
```diff
@@ -362,6 +362,28 @@ function invalidateVirtualMod(viteServer: ViteDevServer) {
...
+
+	return JSON.stringify(
+		`${isRelativePath(normalizedConfigPath) ? '' : './'}${
+			normalizedConfigPath.endsWith('.ts')
```
ematipico, line 380: There are also `.d.ts` files. Are they covered by this check on purpose?

florian-lefebvre: AFAIK it should not be .d.ts? So is it worth handling this case?

loucyx: If we never get `.d.ts` and is always `.ts` then we can update this to replace `.ts` with `.js` simplifying the code a bit:

```ts
function normalizeConfigPath(from: string, to: string) {
	const normalizedConfigPath = path.relative(from, to);

	return JSON.stringify(
		`${isRelativePath(normalizedConfigPath) ? '' : './'}${
			normalizedConfigPath.replace(/\.ts$/, '.js')
		}`
	);
}
```

packages/astro/src/content/types-generator.ts
loucyx: I used "normalize" because the function is an "action" and the result would be the "relative" path. Maybe something like `configPathToRelative`?

packages/astro/src/content/types-generator.ts
ematipico: > If we never get .d.ts

That's what I **not** sure about (I am not familiar with this part of the code). So maybe keeping the code as is better.

ematipico: cc @bholmesdev

bholmesdev: Ah, this looks like the loader for `src/content/config.ts` correct? If so, we only support `.ts` and `.js`. `.d.ts` shouldn't be possible 👍

bholmesdev (approved):
Good find! Like leaning on the `.js` convention for TypeScript. `.js`ust ship it 👍

packages/astro/src/content/types-generator.ts
bholmesdev: Fair point. I believe `.mjs` would be possible, and this change would silently leave the `.mjs` extension without changing it

loucyx: I’ll update the PR with the simplified code, then. Thank you!

PR #9685, opened 2024-01-12
fix(transitions router): attempt to find the clicked element within an open shadow root

### LINKED ISSUE #9670: Web component links do not trigger view transitions

lilnasy:
## Changes

- Closes #9670

## Testing
Added a test case.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -53,6 +53,9 @@ const { fallback = 'animate' } = Astro.props;
 	if (supportsViewTransitions || getFallback() !== 'none') {
 		document.addEventListener('click', (ev) => {
 			let link = ev.target;
+			if (ev.composed) {
+			    link = ev.composedPath()[0];
+			}
 			if (link instanceof Element) {
 				link = link.closest('a, area');
 			}
--- a/packages/astro/e2e/fixtures/view-transitions/src/pages/one.astro
+++ b/packages/astro/e2e/fixtures/view-transitions/src/pages/one.astro
@@ -12,6 +12,11 @@ import Layout from '../components/Layout.astro';
 	<a id="click-redirect-two" href="/redirect-two">go to redirect 2</a>
 	<a id="click-redirect-external" href="/redirect-external">go to a redirect external</a>
 	<a id="click-404" href="/undefined-page">go to undefined page</a>
+	<custom-a id="custom-click-two">
+		<template shadowrootmode="open">
+			<a href="/two">go to 2</a>
+		</template>
+	</custom-a>
 
 	<div id="test">test content</div>
 </Layout>
--- a/.changeset/weak-planes-help.md
+++ b/.changeset/weak-planes-help.md
@@ -1,0 +1,4 @@
+---
+"astro": patch
+---
+Fixes an issue where anchor elements within a custom component could not trigger a view transition.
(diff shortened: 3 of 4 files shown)


lilnasy:
!preview 9685

lilnasy:
!preview 9685

natemoo-re (approved):
Looks great, nice test!

packages/astro/e2e/fixtures/view-transitions/src/pages/one.astro
```diff
@@ -12,6 +12,11 @@ import Layout from '../components/Layout.astro';
 	<a id="click-redirect-two" href="/redirect-two">go to redirect 2</a>
 	<a id="click-redirect-external" href="/redirect-external">go to a redirect external</a>
 	<a id="click-404" href="/undefined-page">go to undefined page</a>
+	<custom-a id="custom-click-two">
+		<template shadowrootmode="open">
```
MoustaphaDev, line 16: I'm not familiar with custom elements, but is it required to have the ` shadowrootmode` set to "open" for this to work? If that's the case it would be a good idea documenting it

lilnasy: In the changeset, you mean?

MoustaphaDev: I meant somewhere in the VT docs as a note/warning maybe

packages/astro/components/ViewTransitions.astro
```diff
@@ -53,6 +53,9 @@ const { fallback = 'animate' } = Astro.props;
 	if (supportsViewTransitions || getFallback() !== 'none') {
 		document.addEventListener('click', (ev) => {
 			let link = ev.target;
+			if (ev.composed) {
```
martrapp, line 56: This may be superfluous, but on the other hand it can't hurt either.

packages/astro/e2e/fixtures/view-transitions/src/pages/one.astro
lilnasy: You're right. I could make a PR to docs after this.

Issue #7962, opened 2023-08-04
Routing error when navigating to path that includes an encoded forward slash


svensigmond:
### What version of `astro` are you using?

2.10.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Windows

### What browser are you using?

Firefox

### Describe the Bug

Given this folder structure:
```
src
-pages
--currencies
---[id].astro
```

When navigating to a route with an encoded forward slash, for example: `http://foo.com/currencies/eur%2Fusd`, the server returns a type error:

```
Expected "id" to be a string

dist/index.js:237:19


Stack Trace
TypeError: Expected "id" to be a string
    at Object.generate (project_path\node_modules\path-to-regexp\dist\index.js:237:19)
    at stringifyParams (project_path/packages/client/node_modules/astro/dist/core/routing/params.js:23:31)
    at findPathItemByKey (project_path/packages/client/node_modules/astro/dist/core/render/route-cache.js:74:21)
    at getParamsAndProps (project_path/packages/client/node_modules/astro/dist/core/render/params-and-props.js:19:29)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async matchRoute (project_path/packages/client/node_modules/astro/dist/vite-plugin-astro-server/route.js:35:7)
    at async run (project_path/packages/client/node_modules/astro/dist/vite-plugin-astro-server/request.js:52:28)
    at async runWithErrorHandling (project_path/astro-boilerplate/packages/client/node_modules/astro/dist/vite-plugin-astro-server/controller.js:65:5)
    at async handleRequest (project_path/packages/client/node_modules/astro/dist/vite-plugin-astro-server/request.js:48:3)
```

This bug seems to have been introduced in the update from version **2.8.4** to **2.8.5**.

It is possible to reach the slugged page in version 2.8.4, although the id will not be present in `Astro.params`, which will return an empty object instead of  and object containing the id and value for the slug.

Please see the repo on Stackblitz and click the link on the home page to see the error.

### What's the expected result?

I would expect to be able to navigate to urls that have been encoded using `encodeURI()` or `encodeURIComponent()`. I also expect these url params to be present in `Astro.params`.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-mt9ply

### Participation

- [ ] I am willing to submit a pull request for this issue.

rishi-raj-jain:
[...]

PR #9696, opened 2024-01-14
fix(routing): supports encoded forward slashes in URI params

asdfjkalsdfla:
## Changes
Fix for #7962 

## Testing
Ran unit test & added new unit test

## Docs
If the user had a URI encoded `, ? : @ & = + $ # /` in the path, it will no longer be decoded automatically.


packages/astro/src/core/render/params-and-props.ts
```diff
@@ -64,7 +64,7 @@ function getRouteParams(route: RouteData, pathname: string): Params | undefined
 	if (route.params.length) {
 		// The RegExp pattern expects a decoded string, but the pathname is encoded
 		// when the URL contains non-English characters.
-		const paramsMatch = route.pattern.exec(decodeURIComponent(pathname));
+		const paramsMatch = route.pattern.exec(decodeURI(pathname));
```
natemoo-re, line 67: We've had many URI encoding issues in the past, so we tend to be very cautious when modifying any logic related to this.
 
Seems like using `decodeURI` here would break users who have encoded `, ? : @ & = + $ #` but potentially fix just users who have encoded `/`?

natemoo-re (changes requested):
I think #7962 is a valid issue, but it's not totally clear to me if this is the most robust fix. Would you mind adding some additional tests for the cases that are potentially broken by this change?

packages/astro/src/core/render/params-and-props.ts
asdfjkalsdfla: Yes, the change the behavior for users who have those symbols encoded in their parameter. But, it prevents encoded components from being interrupted as query params, uri params, location hash, etc... My gut is the change here is the right long-term strategy to prevent that class of bugs. (I also think may be issues with cache misses for these cases.)

I did think about adding automatic uri component decoding when accessing the parameter itself. However, I couldn't find a great spot to do that. The params object are used as part of the URI cache today, so, I can't encode it there and still have the cache work. There might be a spot to add this when accessing the value, but I don't know if that would cause other inconsistencies.

The behavior around this one was changed in a [minor release](https://github.com/withastro/astro/commit/4c7ba4da084d7508df91cbac03c2b099a8301e2b) in the past (i.e. from not decoded to decoded), so I leaned towards not making a more major refactor to keep backwards compatibility.

What do you think?

asdfjkalsdfla:
Also, happy to add the additional test coverage for these other cases. Will try to get those over tomorrow.

packages/astro/src/core/render/params-and-props.ts
ematipico: > Yes, the change the behavior for users who have those symbols encoded in their parameter

This will cause a breaking change for users who rely on `, ? : @ & = + $ #`. Not sure we can land this PR as is 🤔

natemoo-re:
Thanks again for the PR! I need to do a bit more digging on the original issue. The bug is totally valid but after testing this PR's `decodeURI` logic, I'm confident that this is not the correct fix.

We are technically decoding a URI Component rather than an entire URI here. The problem is probably higher up in the routing logic, decoding the `/` before we are able to match it to a route.

I'm going to close this PR, but I have put #7962 on my board and am hoping to investigate and fix it very soon.

PR #9712, opened 2024-01-17
Refactor style-only HMR cache and performance

bluwy:
## Changes

Supersedes and closes https://github.com/withastro/astro/pull/9356

This PR changes how we detect if only the styles are changed in an Astro file, so we can only apply HMR on those styles. In contrary to https://github.com/withastro/astro/pull/9356, this uses regexes to perform the comparisons, which is technically a bit brittle, but I think covers a fair amount of cases enough that it's worth the perf gain.

It also simplifies our caching, so it's less about _caching_ but more of _recording_ additional metadata from the compilation result. This way we use lesser data and memory, and avoid unintentional invalidation causing scripts and styles virtual modules to fail. 

@natemoo-re Would love your thoughts on the regex approach for style-change detection. I'm happy to refactor that to something else if anyone doesn't feel comfortable about it 😄 
 
## Testing

The existing tests should pass as it's a refactor.

## Docs

n/a. internal refactor.


matthewp (approved):

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -43,35 +41,57 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
...
 		},
 		buildStart() {
 			astroFileToCssAstroDeps = new Map();
+			astroFileToCompileMetadata = new Map();
+
+			// Share the `astroFileToCompileMetadata` across the same Astro config as Astro performs
+			// multiple builds and its hoisted scripts analyzer requires the compile metadata from
+			// previous builds. Ideally this should not be needed when we refactor hoisted scripts analysis.
+			if (astroFileToCompileMetadataWeakMap.has(config)) {
+				astroFileToCompileMetadata = astroFileToCompileMetadataWeakMap.get(config)!;
```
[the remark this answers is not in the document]
bluwy: I'm not really sure about `@ts-expect-error` since we only need to suppress a specific error instead of the whole line. I think once TS improves this, our eslint config should flag it: 
image

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -43,35 +41,57 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
...
-					throw new Error('No cached compile result found for ' + id);
-				}
+			// Astro scripts and styles virtual module code comes from the main Astro compilation
+			// through the metadata from `astroFileToCompileMetadata`. It should always exist as Astro
+			// modules are compiled first, then its virtual modules. If the virtual modules are somehow
+			// compiled first, throw an error and we should investigate it.
+			const filename = normalizePath(normalizeFilename(parsedId.filename, config.root));
+			const compileMetadata = astroFileToCompileMetadata.get(filename);
+			if (!compileMetadata) {
+				throw new Error('No cached compile metadata found for ' + id);
```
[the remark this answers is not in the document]
bluwy: Sounds good 👍 I can add a little more hint about this, but I think I'll keep the more extensive description as a comment above so we can return to it in code.

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -68,25 +58,39 @@ export async function handleHotUpdate(
...
+
+	// Before we can regex-capture style tags, we remove the frontmatter and scripts
+	// first as they could contain false-positive style tag matches. At the same time,
+	// we can also compare if they have changed and early out.
+
+	// Strip off and compare frontmatter
+	let oldFrontmatter = '';
+	let newFrontmatter = '';
+	oldCode = oldCode.replace(frontmatterRE, (m) => ((oldFrontmatter = m), ''));
+	newCode = newCode.replace(frontmatterRE, (m) => ((newFrontmatter = m), ''));
```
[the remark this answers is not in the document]
bluwy: If we do that, I think we'll miss out on the early checks in this function for perf though 🤔 Maybe I'm missing something for what you have in mind.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -43,35 +41,60 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
...
+			// through the metadata from `astroFileToCompileMetadata`. It should always exist as Astro
+			// modules are compiled first, then its virtual modules. If the virtual modules are somehow
+			// compiled first, throw an error and we should investigate it.
+			const filename = normalizePath(normalizeFilename(parsedId.filename, config.root));
+			const compileMetadata = astroFileToCompileMetadata.get(filename);
+			if (!compileMetadata) {
+				throw new Error(
+					`No cached compile metadata found for "${id}". The main Astro module "${filename}" should have ` +
+						`compiled and filled the metadata first, before its virtual modules can be requested.`
+				);
```
natemoo-re, line 97: We should always use the AstroError subclass if we can. I think this would fit under `UnknownCompilerError`?

packages/astro/src/vite-plugin-astro/hmr.ts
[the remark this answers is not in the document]
natemoo-re: I agree that this `replace()` + mutation pattern is quite difficult to read, but I do see how it's a useful one-liner in this case. I'm okay with it if you think this is the cleanest approach!

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -68,25 +58,39 @@ export async function handleHotUpdate(
...
+	if (oldFrontmatter !== newFrontmatter) return false;
 
-const astroStyleImportRE = /import\s*"[^"]+astro&type=style[^"]+";/g;
-const sourceMappingUrlRE = /\/\/# sourceMappingURL=[^ ]+$/gm;
+	// Strip off and compare scripts
+	const oldScripts: string[] = [];
+	const newScripts: string[] = [];
+	oldCode = oldCode.replace(scriptRE, (m) => (oldScripts.push(m), ''));
+	newCode = newCode.replace(scriptRE, (m) => (newScripts.push(m), ''));
```
natemoo-re, line 83: `scriptRE` is a single global RegExp outside of the `isStyleChanged()` scope. Multiple executions of `isStyleChanged()` will continue from the previous `lastIndex`.

You need to manually reset `scriptRE.lastIndex = 0` at the top of `isStyleChanged` so the regex matches from the beginning of the file.

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -68,25 +58,39 @@ export async function handleHotUpdate(
...
- * Remove style-related code and sourcemap from the final astro output so they
- * can be compared between non-style code
- */
-function normalizeCode(code: string) {
-	return code.replace(astroStyleImportRE, '').replace(sourceMappingUrlRE, '').trim();
+	// Finally, we can compare styles
+	const oldStyles: string[] = [];
+	const newStyles: string[] = [];
+	oldCode.match(styleRE)?.forEach((m) => oldStyles.push(m));
+	newCode.match(styleRE)?.forEach((m) => newStyles.push(m));
```
natemoo-re, line 90: Same comment here for `styleRE`!

.changeset/eight-turtles-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Improves Astro file style-only HMR caching and performance
```
natemoo-re, line 5:
```suggestion
Improves HMR behavior for style-only changes in `.astro` files
```

natemoo-re (commented):
Very clean refactor! Big fan of this approach and reducing the complexity of our previous caching system.

packages/astro/src/vite-plugin-astro/hmr.ts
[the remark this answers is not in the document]
bluwy: > If I were to read the code in the future, it would a surprise to see that `newCode` is manipulated inside a function that returns a boolean.

This function that mutates `newCode` won't affect the value of `newCode` when used in `compile`. JavaScript passes strings to parameters as a "copy", so mutating it inside the function only stays within it. It's kinda a trick, but I found it's slightly more performant as you don't have to allocate another variable.

packages/astro/src/vite-plugin-astro/hmr.ts
bluwy: I think `replace` will auto-reset it to 0, usually when it can't find anymore replacement, it'll set to 0 internally. I made a test to confirm this:

```js
const re = /a/g
const str = 'abc_abc_abc'.replace(re, () => '')
console.log(re.lastIndex) // 0
```

packages/astro/src/vite-plugin-astro/index.ts
bluwy: I used an error here since the other ones are also an error 😬 I think they are more of unexpected errors as our normal code paths should never trigger the error in the first place, so `UnknownCompilerError` might not be the most accurate.

I could switch all the errors to a new Astro error in a separate PR though so it's more consistent. Hoping to keep the PR scope down a bit!

ematipico (approved):

packages/astro/src/vite-plugin-astro/hmr.ts
natemoo-re: > Because `@@replace` would keep calling `exec()` until it returns null, and `exec()` would automatically reset the regex's [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) to 0 when the last match fails, `@@replace` would typically not have side effects when it exits. However, when the regex is [sticky](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) but not global, `lastIndex` would not be reset. In this case, each call to `replace()` may return a different result.
> - [MDN `@@replace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@replace)

TIL. The nuances of regex continue to be wild.

natemoo-re (approved):

PR #9713, opened 2024-01-17
[docs] fix syntax error, avoid a Markdown table

sarah11918:
## Changes

Fixes a JSDoc syntax error for config reference docs.

Also, reworks content to avoid a Markdown table. Not entirely sure how this will look, but we are trying to avoid tables in this file:
image

## Testing

No tests: corrects formatting errors in docs site

## Docs
All docs!


Princesseuh (approved):
Looks fine to me! JSDoc are complicated

PR #9716, opened 2024-01-17
Reduce work done by `astro sync`

### LINKED ISSUE #9711: Malformed entry filename is not reported
### Astro Info

```block
Astro                    v4.1.3
Package Manager          npm
Adapter                  none
                         @astrojs/sitemap
```

_No response_

### Describe the Bug

When an entry has an invalid frontmatter, the parsing library throws an error but the associated Astro error is irrelevant: it points to the `content/config.*` instead of the faulty file. For example, the following mdx file
```md
---
# example.mdx
title: 'Here's a wrong pair'
---
```
will throw this:
```
[GenerateContentTypesError] `astro sync` command failed to generate content collection types: can not read an implicit mapping pair; a colon is missed
  Hint:
    Check your `src/content/config.*` file for typos.
  Error reference:
    https://docs.astro.build/en/reference/errors/generate-content-types-error/
  Stack trace:
```

Have a new error that specifies what file is failing. I don't think it's realistic to also want the exact location since it depends on the parsing library

https://stackblitz.com/edit/github-se3vcm?file=src%2Fcontent%2Fblog%2Ffirst-post.md

### Participation

lilnasy:
## Changes
- Closes #9711 
image

## Testing

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/new-tigers-kiss.md
+++ b/.changeset/new-tigers-kiss.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Reduces the amount of work done by `astro sync`.
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -107,7 +107,7 @@ export async function createContentTypesGenerator({
 		return { typesGenerated: true };
 	}
 
-	async function sync() {
+	async function sync({ loadContentConfig }: { loadContentConfig?: boolean }) {
 		await writeContentFiles({
 			fs,
 			collectionEntryMap,
@@ -116,6 +116,7 @@ export async function createContentTypesGenerator({
 			contentEntryTypes: settings.contentEntryTypes,
 			viteServer,
 		});
+		if (loadContentConfig) await reloadContentConfigObserver({ fs, settings, viteServer });
 	}
 
 	async function handleEvent(event: ContentEvent): Promise<{ shouldGenerateTypes: boolean }> {
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -107,6 +107,17 @@ export async function createContentTypesGenerator({
 		return { typesGenerated: true };
 	}
 
+	async function sync() {
+		await writeContentFiles({
+			fs,
+			collectionEntryMap,
+			contentPaths,
+			typeTemplateContent,
+			contentEntryTypes: settings.contentEntryTypes,
+			viteServer,
+		});
+	}
+
 	async function handleEvent(event: ContentEvent): Promise<{ shouldGenerateTypes: boolean }> {
 		if (event.name === 'addDir' || event.name === 'unlinkDir') {
 			const collection = normalizePath(
@@ -321,7 +332,7 @@ export async function createContentTypesGenerator({
 			}
 		}
 	}
-	return { init, queueEvent };
+	return { init, sync, queueEvent };
 }
 
 // The virtual module contains a lookup map from slugs to content imports.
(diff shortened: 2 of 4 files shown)


sarah11918 (approved):
Assuming that this now also says *which* file doesn't satisfy the content collections type, I'd say that's an improvement! The only way it could be better is if it identified the specific way in which the file did not conform.

matthewp (approved):
lgtm

florian-lefebvre:
I guess we need a changeset?

lilnasy:
> Assuming that this now also says which file doesn't satisfy the content collections type, I'd say that's an improvement! The only way it could be better is if it identified the specific way in which the file did not conform.

Apparently it already does, **but only during build**.
image

florian-lefebvre (approved):

Princesseuh:
> > Assuming that this now also says which file doesn't satisfy the content collections type, I'd say that's an improvement! The only way it could be better is if it identified the specific way in which the file did not conform.
> 
> Apparently it already does, **but only during build**. image

Those are essentially two different errors, one is because it couldn't generate the types at all (entry malformed), whereas the other one is that a type didn't match

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1190,8 +1190,8 @@ export const UnknownCLIError = {
 export const GenerateContentTypesError = {
 	name: 'GenerateContentTypesError',
 	title: 'Failed to generate content types.',
-	message: (errorMessage: string) =>
-		`\`astro sync\` command failed to generate content collection types: ${errorMessage}`,
+	message: (errorMessage: string, file?: string) =>
```
Princesseuh, line 1193: You're gonna need a `@message` property here, because the docs won't be able to parse a message from the ternary.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1182,6 +1182,7 @@ export const UnknownCLIError = {
 } satisfies ErrorData;
 /**
  * @docs
+ * @message `astro sync` command failed to generate content collection types.
  * @description
  * `astro sync` command failed to generate content collection types.
```
Princesseuh, line 1187: Not blocking, but this description is kinda terrible

lilnasy:
> Those are essentially two different errors, one is because it couldn't generate the types at all (entry malformed), whereas the other one is that a type didn't match.

I would've expected it to validate if it parses the content entries. Maybe the odd part is the fact that `astro sync` reads the markdown files in the first place.

packages/astro/src/core/errors/errors-data.ts
lilnasy: Now that you bring attention here, it is odd that a syntax error in markdown is one of the reasons, considering that you mentioned there are two different concerns here.

natemoo-re:
> I would've expected it to validate if it parses the content entries. Maybe the odd part is the fact that `astro sync` reads the markdown files in the first place.

Honestly that is surprising. I don't think anything about `sync` _needs_ to perform eager validation.

natemoo-re (dismissed):
#NWTWWHB

lilnasy:
Sorry for stealing all the reviews, but I think we can not even error in the first place.

lilnasy:
This is annoying, `build` depends on `sync` via shared global state.

image

Princesseuh:
All of the `astro x` commands run `astro sync` at their start, so I assume this is true for many commands

lilnasy:
That makes sense. I should clarify what I found problematic working on this.

`build` depends not just on the functionality of `sync`, but also on it being performed in the same way `npm exec astro sync` would have. It involves starting its own vite server, 
doing it in the same process, and mutating imported global state so `build` can consume the result.

The result is errors appearing from unexpected places, and tracking them down being a guessing game.

lilnasy:
I did not feel good about the enhacing `GenerateContentTypesError` because it is a catch-all error. It does not make sense for it to report a specific file name.

The solution I want is where a new `AstroError` wraps the `YamlException` close to where it is thrown. That can be in  a new PR.

PR #9643, opened 2024-01-08
Bump shikiji, use transformers API, expose transformers API

blackmann:
## Changes

- Bump `shikiji` version
- Use [transformers](https://shikiji.netlify.app/guide/transformers) API. `tranforms` is deprecated.
- Expose the `transformers` API so that some [tranformer presets](https://shikiji.netlify.app/packages/transformers) or community transformers can be used.

## Testing

There's are existing tests for this and they pass.

## Docs

We'll need to update the docs to indicate that `transformers` can be supplied to `astro.config.js` > `markdown.shikiConfig.transformers`

/cc @withastro/maintainers-docs for feedback!

PR #9723, opened 2024-01-18
normalize class property

blackmann:
## Changes

Attempting to use some of the [common transformers](https://shikiji.netlify.app/packages/transformers) raises an exception:

```
MDXError: classValue.replace is not a function
    at TransformContext.transform (node_modules/@astrojs/mdx/dist/index.js:93:27)
    at async Object.transform (node_modules/vite/dist/node/chunks/dep-V3BH7oO1.js:64060:30)
    at async loadAndTransform (node_modules/vite/dist/node/chunks/dep-V3BH7oO1.js:49741:29)
    at async instantiateModule (node_modules/vite/dist/node/chunks/dep-V3BH7oO1.js:50759:10)
```

This is because `addClassToHast` returns an array of strings. But `Array`s don't have a `.replace`. See: https://github.com/antfu/shikiji/blob/31241ba4dc47c2d4d7809ee08c6573b26b07b812/packages/shikiji-core/src/utils.ts#L44

## Testing

I couldn't figure out how to write a unit test for this. But I edited a build version with this logic and it works.

### DIFF
--- a/.changeset/nervous-dodos-compete.md
+++ b/.changeset/nervous-dodos-compete.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/markdown-remark": patch
+---
+
+Normalize class property
--- a/.changeset/nervous-dodos-compete.md
+++ b/.changeset/nervous-dodos-compete.md
@@ -2,4 +2,4 @@
 "@astrojs/markdown-remark": patch
 ---
 
-Normalize class property
+Fixes a case where transformers wouldn't work on the `class` property
--- a/packages/markdown/remark/src/shiki.ts
+++ b/packages/markdown/remark/src/shiki.ts
@@ -62,10 +62,8 @@ export async function createShikiHighlighter({
 								node.tagName = 'code';
 							}
 
-							const classValue = normalizeMaybeArray(node.properties.class) ?? '';
-
-							// Cast to string as shikiji will always pass them as strings instead of any other types
-							const styleValue = (node.properties.style as string) ?? '';
+							const classValue = normalizePropAsString(node.properties.class) ?? '';
+							const styleValue = normalizePropAsString(node.properties.style) ?? '';
 
 							// Replace "shiki" class naming with "astro-code"
 							node.properties.class = classValue.replace(/shiki/g, 'astro-code');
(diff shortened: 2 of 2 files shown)


packages/markdown/remark/src/shiki.ts
```diff
@@ -129,6 +131,10 @@ export async function createShikiHighlighter({
 	};
 }
 
+function normalizeMaybeArray(value: Properties[string]): string | null {
+	return Array.isArray(value) ? value.join(' ') : (value as string);
```
ematipico, line 135: If `value` is cast to `string`, when does the function return `null`?

blackmann: Updated @ematipico

matthewp:
What does this fix? Is there an issue or code example?

blackmann:
I gave a description with the PR; to elaborate, if you added a codeblock like the following to an `.md` content file:

```rust
  // &mut means mutable reference
  fn next_question(&mut self) -> &Question {
    let count = self.questions.len() - 1; // [!code --]
    if self.current_question().is_answered() && self.current_index < count {
      self.current_index += 1;
    }

    self.current_question()
  }
```

Notice on line 3, we do `// [!code --]`, this triggers the [transformerNotationDiff](https://shikiji.netlify.app/packages/transformers#transformernotationdiff) transformer. In that process, `addClassToHast` is called which converts the `class` from string to an array of string.

This PR converts it back to a string by just concatenating them.

Here's a screenshot of the error

Screenshot 2024-01-18 at 7 10 30 PM

@matthewp

blackmann:
This is a `patch` follow-up to #9643 which exposes the `transformers` API from shikiji.

packages/markdown/remark/src/shiki.ts
```diff
@@ -61,8 +62,9 @@ export async function createShikiHighlighter({
 								node.tagName = 'code';
 							}
 
+							const classValue = normalizeMaybeArray(node.properties.class) ?? '';
+
```
bluwy, line 66: Thanks for the fix! This makes sense to me. Could we also handle this for `style`? I assume transformers could also change that to a non-string now.

Maybe we can rename and generalize the function as `normalizePropAsString`

blackmann:
@bluwy @ematipico 

ready

.changeset/nervous-dodos-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/markdown-remark": patch
+---
+
+Normalize class property
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where transformers wouldn't work on the `class` property
```

blackmann:
@florian-lefebvre @bluwy @ematipico 

Can we get this merged soon?

PR #9726, opened 2024-01-18
Fix regression in the routing priority of index routes

### LINKED ISSUE #9724: Dynamic routes take precedence over index.astro in 4.2
### Astro Info

```block
Astro                    v4.2.0
Package Manager          pnpm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

I've hit a potentially breaking change when updating from 3.5.5 to 4.2.0: `[...dynamic].astro` routes seem to have higher priority than `index.astro` after the upgrade. Apologies if this is an expected change, but I was unable to find 
it in the changelog.

There is a minimal stackblitz reproduction - with 3.x this would have rendered the Index page.

In the stackblitz example I would expect the **Index** page to be rendered, but instead the **dynamic** route is rendered.

https://stackblitz.com/edit/github-pqgx5b?file=src%2Fpages%2F%5B...dynamic%5D.astro

### Participation

Fryuni:
## Changes

- Fix regression that causes pages ending with `index.astro` to be overridden by dynamic routes with rest parameters.
- Closes #9724

## Testing

There was an incorrect test for the index priority, that was fixed and a new test for the full combination of routes in the same directory was added.

## Docs

No changes in docs since this was an unintended regression. The docs already reflect the correct behavior.

### DIFF
--- a/.changeset/smart-rules-train.md
+++ b/.changeset/smart-rules-train.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix regression in routing priority between `index.astro` routes and dynamic routes with rest parameters
--- a/.changeset/smart-rules-train.md
+++ b/.changeset/smart-rules-train.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fix regression in routing priority between `index.astro` routes and dynamic routes with rest parameters
+Fixes a regression in routing priority between `index.astro` and dynamic routes with rest parameters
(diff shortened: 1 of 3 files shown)


.changeset/smart-rules-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix regression in routing priority between `index.astro` routes and dynamic routes with rest parameters
```
florian-lefebvre, line 5:
```suggestion
Fixes a regression in routing priority between `index.astro` and dynamic routes with rest parameters
```

matthewp (approved):
looks great!

PR #9730, opened 2024-01-19
fix getRelativeLocaleUrl when redirectToDefaultLocale is false

Blede2000:
## Changes

When using getLocaleRelativeUrl or getLocaleAbsoluteUrl while having both prefixDefaultLocale set to true and redirectToDefaultLocale set to false, should also return the locale path in the url.

For example `getLocaleRelativeUrl('en', '/posts')` should return `/en/posts` but right now it returns `/posts`

## Testing

Not sure if any tests should change, there's probably an opportunity for a test but I couldn't find tests for these utilities in i18n-routing.test.js

## Docs

This should be the expected behaviour no doc updates

### DIFF
--- a/.changeset/sweet-owls-trade.md
+++ b/.changeset/sweet-owls-trade.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Allow i18n routing utilities like getRelativeLocaleUrl to also get the default local path when redirectToDefaultLocale is false
--- a/packages/astro/src/i18n/index.ts
+++ b/packages/astro/src/i18n/index.ts
@@ -109,7 +109,7 @@ export function getLocaleRelativeUrlList({
 		const pathsToJoin = [base, prependWith];
 		const normalizedLocale = normalizeLocale ? normalizeTheLocale(locale) : locale;
 
-		if (routing === 'pathname-prefix-always') {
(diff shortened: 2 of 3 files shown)


ematipico:
We have a bunch of unit tests in these two places:
- https://github.com/withastro/astro/blob/main/packages%2Fastro%2Ftest%2Funits%2Fi18n%2Fastro_i18n.test.js
- https://github.com/withastro/astro/blob/main/packages%2Fastro%2Ftest%2Fi18n-routing.test.js#L6

We should add some add test case

dallyh:
@Blede2000 @ematipico Just want to mention that this should also be done for `getAbsoluteLocaleUrl()` as with the same config it does not return the prefixed locale, possibly also for other routing functions.

Blede2000:
@ematipico Thank you! I've added tests, they follow the same format as the one with routing strategy `pathname-prefix-always` because they should behave the same.

@dallyh That's a good point, I updated the PR this should resolve `getRelativeLocaleUrl()` `getLocaleAbsoluteUrl()` `getLocaleRelativeUrlList()` `getLocaleAbsoluteUrlList()`

ematipico (approved):
Awesome work! thank you ❤️

Issue #9740, opened 2024-01-19
Dot files not copied over from `public` folder if not running build from root folder


taktran:
### Astro Info

```block
✗ npx astro info
Astro                    v3.6.4
Node                     v20.11.0
System                   macOS (arm64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/react
                         @astrojs/markdoc
                         @astrojs/sitemap
```

### Describe the Bug

I have a `public/.htaccess` file and when running `astro build`, it doesn't get copied over to the dist folder. I also have a custom `outDir` set to something like `../../dist/packages/folder`.

~~The strange thing is, I can't quite replicate it on stackblitz. This is my attempt, where the `.htaccess` **does** get generated when you run `npm run build` in the terminal: https://stackblitz.com/edit/github-es2mdg-rajrjb?file=README.md~~ (it can be reproduced when run from another directory - see stackblitz below)

However, looking at the code and patching the file locally, it does fix it on my local machine:

https://github.com/withastro/astro/blob/main/packages/astro/src/core/build/static-build.ts#L439-L445

From

```
if (out.toString() !== opts.settings.config.outDir.toString()) {
  // Copy assets before cleaning directory if outside root
  await copyFiles(out, opts.settings.config.outDir);
  await fs.promises.rm(out, { recursive: true });
  return;
}

// Where
copyFiles(fromFolder: URL, toFolder: URL, includeDotfiles = false) 
```

it looks like `copyFiles` should set `includedDotfiles` parameter to true ie,

```
await copyFiles(out, opts.settings.config.outDir, true);
```

The `out.toString() !== opts.settings.config.outDir.toString()` check seems to match with my scenario where `outDir` is set, and I can't think of a reason why it wouldn't copy dot files (or a least make it configurable somewhere).

The `includedDotfiles` parameter seems to have been added in https://github.com/withastro/astro/pull/5832, if that is useful.

Am I missing something?

### What's the expected result?

Expect `.htaccess` to appear in the dist folder when running `astro build`. ~~Don't see the bug in the attached stackblitz, but can replicate locally and looking at the code, it does seem like it's a legitimate bug.~~

### Link to Minimal Reproducible Example

~~https://stackblitz.com/edit/github-es2mdg-rajrjb (not a reproducible example, but it's the gist of the steps - see above)~~

https://stackblitz.com/edit/github-es2mdg-xzyijr?file=README.md (actual reproducible example)

### Participation

[...]

PR #9741, opened 2024-01-19
Fix dot files not being copied over on non-root builds (#9740)

taktran:
## Changes

Fixes https://github.com/withastro/astro/issues/9740

When running `astro build` outside of where the astro files are, dotfiles are not copied over because the `includeDotfiles` parameter is not passed in. This PR adds a `true` `includeDotfiles` value to `copyFiles` to fix this.

## Testing

1. Check out this PR locally
1. Download astro bug example https://stackblitz.com/edit/github-es2mdg-tnkbfm?file=README.md
1. Check bug

   ```bash
   cd another-astro
   pnpm install
   npm run astro build -- --root ../
   ls -la ../dist
   # Notice that there is no `.htaccess` file

   pnpm link [PR-root]/astro/packages/astro
   npm run astro build -- --root ../
   ls -la ../dist
   # Now there is a `.htaccess` file
   ```

Not sure if this requires an e2e test, but if so, please point me in the direction of an example of how one is set up.

## Docs

Fixes a bug with build, so shouldn't affect any of the docs

### DIFF
--- a/.changeset/slimy-jobs-smash.md
+++ b/.changeset/slimy-jobs-smash.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix dot files not being copied over on non-root builds
--- a/.changeset/slimy-jobs-smash.md
+++ b/.changeset/slimy-jobs-smash.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fix dot files not being copied over on non-root builds
+Fixes an issue where dot files were not copied over from the public folder to the output folder when build command was run inside a subfolder.
--- a/.changeset/slimy-jobs-smash.md
+++ b/.changeset/slimy-jobs-smash.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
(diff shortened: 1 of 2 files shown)


lilnasy (approved):
Tried this earlier. Looks great to me!

Thanks for fixing it 💜

.changeset/slimy-jobs-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix dot files not being copied over on non-root builds
```
lilnasy, line 5:
```suggestion
Fixes an issue where dot files were not copied over from the public folder to the output folder when build command was run inside a subfolder.
```

taktran: It's not quite just a subfolder, it's any folder that is not the root of the project. I've updated the changeset description to reflect this.

natemoo-re (approved):
LGTM! I think with this change we might always be copying dotfiles and can probably remove the third argument? No need to take care of that in this PR, though.

florian-lefebvre (approved):

PR #9746, opened 2024-01-20
fix(rss): rssSchema definition to allow calling standard zod object methods

### LINKED ISSUE #9744: astro/rss: type of rssSchema changed so it can no longer be extended
### Astro Info

```block
Package Manager          npm
Adapter                  none
```

_No response_

### Describe the Bug

Hello 👋 

#9610 introduced a [change to `rssSchema`](https://github.com/withastro/astro/pull/9610/files#diff-01273eb750ee014202811acc2c006387a1f903dd4d33ea6994f0d2748ceed582) where it went from being a Zod object to being a Zod union.

I don't remember where I picked up this pattern, but my goal is to ensure I have a collection that conforms to the RSS spec. At the same time I have some custom fields in the frontmatter that I use in my blog post template `.astro`.

This collection works with `@astro/rss@4.0.1`. In `@astro/rss@4.0.2` it causes an error: `__vite_ssr_import_1__.rssSchema.extend is not a function`

```js
import { z, defineCollection } from "astro:content";
import { rssSchema } from "@astrojs/rss";

const blogCollection = defineCollection({
	schema: ({ image }) =>
		rssSchema.extend({
			category: z.string().optional(),
			cover: z
				.object({
					image: image(),
					alt: z.string(),
				})
				.optional(),
		}),
});

export const collections = {
	blog: blogCollection,
};
```

My expectation when upgrading was for `rssSchema.extend` to continue to work, but I'll freely admit I might just be doing something silly here 🙈 

Should I add the other fields some different way?

https://stackblitz.com/edit/github-yuurwk?file=src%2Fcontent%2Fconfig.ts

### Participation

florian-lefebvre:
## Changes

- Close #9744
- Uses `object` and `refine` instead of an union

## Testing

Tests are still passing

### DIFF
--- a/.changeset/mighty-icons-try.md
+++ b/.changeset/mighty-icons-try.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/rss": patch
+---
+
+Fixes `rssSchema` definition to allow calling standard zod object methods (like `extend`)
--- a/packages/astro-rss/src/schema.ts
+++ b/packages/astro-rss/src/schema.ts
@@ -22,7 +22,7 @@ export const rssSchema = z.object({
 		.optional(),
 	link: z.string().optional(),
 	content: z.string().optional(),
-}).refine(val => !val.title && !val.description, {
+}).refine(val => val.title || val.description, {
 	message: "At least title or description must be provided.",
 	path: ["title", "description"]
 })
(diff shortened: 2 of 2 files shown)


MoustaphaDev (approved):
Looks good to me code wise!

ematipico (approved):

PR #9749, opened 2024-01-20
Feature: add image size to schema metadata

madcampos:
## Changes

Add new property to schema metadata for images returning the optimized image file size in bytes.

This change makes it easier to add images to RSS feed posts. The `<enclosure>` tag in the RSS spec requires a non 0 size in bytes for the media item.

Without that change a user needs to implement the logic to get the file size and it is brittle and prone to breaking as it depends on different file paths for both the dev server and build process.

Here is an example of the code implemented on the user side: https://github.com/madcampos/madcampos.github.io/blob/772084331d6c9eb822405d5af14c4337c64c5a90/src/pages/blog/feed.xml.ts#L26-L39

## Testing

Tests passed and no new tests were added.

## Docs

It should need an update on the docs for [RSS feeds](https://docs.astro.build/en/guides/rss/) on how to include images for posts.

### DIFF
--- a/.changeset/fresh-onions-hide.md
+++ b/.changeset/fresh-onions-hide.md
@@ -1,0 +1,5 @@
+---
+"astro": minor
+---
+
+Adds image file size to content collection schema
--- a/packages/astro/src/assets/types.ts
+++ b/packages/astro/src/assets/types.ts
@@ -37,6 +37,7 @@ export interface ImageMetadata {
 	height: number;
 	format: ImageInputFormat;
 	orientation?: number;
+	size?: number;
 	/** @internal */
 	fsPath: string;
 }
--- a/packages/astro/src/assets/utils/emitAsset.ts
+++ b/packages/astro/src/assets/utils/emitAsset.ts
@@ -27,6 +27,7 @@ export async function emitESMImage(
 	const emittedImage: Omit<ImageMetadata, 'fsPath'> = {
 		src: '',
 		...fileMetadata,
+		size: (await fs.stat(url)).size,
 	};
 
 	// Private for now, we generally don't want users to rely on filesystem paths, but we need it so that we can maybe remove the original asset from the build if it's unused.


Princesseuh:
Your change does not return the size of optimized images, it returns the size of the originals.

ematipico (changes requested):
We must add new tests for new features.

Princesseuh:
Are you still interested in this feature considering it does not solve your use case?

madcampos:
@Princesseuh I still think it is a valid issue, yes, but to be honest I don't know how to solve it properly. 😕

Having the image size pre-optimization "works" in the sense that a larger size is okay, my concern is if after optimizing the image ends up being larger.
E.g.: in the optimization it converts a photo from jpg to png so the file size ends up balloing.

Princesseuh:
> @Princesseuh I still think it is a valid issue, yes, but to be honest I don't know how to solve it properly. 😕
> 
> Having the image size pre-optimization "works" in the sense that a larger size is okay, my concern is if after optimizing the image ends up being larger. E.g.: in the optimization it converts a photo from jpg to png so the file size ends up balloing.

There's unfortunately no real way to get the size of the image post-optimization, especially in SSR the images are rendered at runtime. I'll note that the RSS spec specify that you can use `length="0"` for files of unknown length:

> When an enclosure's size cannot be determined, a publisher SHOULD use a length of 0.

madcampos:
> There's unfortunately no real way to get the size of the image post-optimization, especially in SSR the images are rendered at runtime. I'll note that the RSS spec specify that you can use `length="0"` for files of unknown length:
> 
> > When an enclosure's size cannot be determined, a publisher SHOULD use a length of 0.

I cannot get the feed to build when setting the image's `length` to 0, it errors out. But that maybe is better fixed on the `@astrojs/rss` package to allow for a length of 0.

Issue #8311, opened 2023-08-30
Astro 3.0 Markdown/MDX auto image asset imports ignore Remark plugins


i-bsd:
### Astro info

```block
Astro version            v3.0.3
Package manager          pnpm
Platform                 linux
Architecture             x64
Adapter                  Couldn't determine.
Integrations             @astrojs/mdx, @astrojs/partytown, @astrojs/prefetch, @astrojs/sitemap, @astrojs/react, @astrojs/svelte, client:click, astro-purgecss, astro-imagetools
```

### What browser are you using?

Chrome

### Describe the Bug

Upgraded to Astro 3.0. 

Using Remark plugin `remark-unwrap-images` to remove parent `<p>` tags from images.

Because of the way Astro is now automatically processing relative path images in Markdown/MDX (e.g. `![alt](./filename.jpg)`), the Remark plugin is ignored.

If I use `/public/` folder images or remote images, there is no issue.

```
import remarkUnwrapImages from 'remark-unwrap-images';

...

integrations: [mdx(
    {
      remarkPlugins: [
        remarkUnwrapImages
      ]
    }
  )... ]
```

Can the automatic image asset import in Markdown/MDX be fixed to check for the existence of Remark plugins?

### What's the expected result?

Outputted `<img>` should not be wrapped by `<p>` tags.

Astro's new automatic image asset imports in Markdown should take into account Remark plugins if they exist.

### Link to Minimal Reproducible Example

https://github.com/i-bsd/astro-3-img-test

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Hmm, I'm not exactly sure *how* we would check for remark / rehype plugins existing for that. Nonetheless I think a way we could make this work for MDX at least would be a way to disable the rehype part that uses the Image component and instead have a remark plugin just adding the attributes to the node using `getImage`

I'm not sure why it doesn't work in vanilla Markdown however, there we just add attributes to the existing `img` tag.

epsseniyer:
1. First, if you're using Astro version 3.0.3, navigate to the root directory of your Astro project and run the command `pnpm install remark-unwrap-images` to install the `remark-unwrap-images package`. This package helps remove the `<p>` tags wrapping the image tags in Markdown files.

2. Next, open the `astro.config.mjs` file in your Astro project. If such a file doesn't exist, create a file named `astro.config.mjs` in the root directory of your project.

[...]

PR #9753, opened 2024-01-21
MDX remark image props

OliverSpeir:
## Changes

- What does this change?

Allows for remark plugins to affect images in mdx files simliar to https://github.com/withastro/astro/pull/9566 which does it for .md files

## Testing

- adds fixture and tests in astro/packages/integrations/mdx/test

## Docs

Docs don't need update but similiar to how mentioned in #9566 there could be a recipe in future


OliverSpeir:
Closes https://github.com/withastro/astro/issues/8311

packages/integrations/mdx/src/plugins.ts
```diff
@@ -54,7 +54,7 @@ function getRemarkPlugins(mdxOptions: MdxOptions): PluggableList {
 		}
 	}
 
-	remarkPlugins = [...remarkPlugins, ...mdxOptions.remarkPlugins];
+	remarkPlugins = [...mdxOptions.remarkPlugins, ...remarkPlugins];
```
Princesseuh, line 57: Hmm, I'm not sure if this can cause issues or not. The image plugins being after should be fine, however maybe gfm and smartypants can cause issues? cc @bholmesdev

packages/integrations/mdx/src/remark-images-to-component.ts
```diff
@@ -89,6 +95,55 @@ export function remarkImageToComponent() {
 					});
 				}
 
+				if (node.data && node.data.hProperties) {
+					const { widths, densities, ...prop } = node.data.hProperties as imgArgs;
```
Princesseuh, line 99: Is there a way we could make this more generic? `widths` and `densities` are attributes our image services supports, but other image services might have more (or less) attributes. Maybe doing some sort of switch with a typeof and automatically making them in the first format would be a better approach

packages/integrations/mdx/src/plugins.ts
OliverSpeir: Ah yeah sorry we did talk about this you're right. I rearrange them again so its gfm/smartyparts -> user defined -> image related -> syntax highlighters

packages/integrations/mdx/src/remark-images-to-component.ts
OliverSpeir: That is much better you're right

packages/integrations/mdx/test/remark-imgattr.test.js
```diff
@@ -0,0 +1,66 @@
...
+					level: 'error',
+					dest: new Writable({
+						objectMode: true,
+						write(event, _, callback) {
+							logs.push(event);
+							callback();
+						},
+					}),
+				}),
+			});
```
Princesseuh, line 36: Since you're not using the logs, you don't need all that part (and the definition on line 18)

Princesseuh (approved):

PR #9758, opened 2024-01-22
chore(@astrojs/node): use Node.js for testing

ematipico:
## Changes

This PR moves `@astrojs/node` to use Node.js for testing

## Testing

Tests should pass

## Docs

N/A

### DIFF
--- a/packages/integrations/node/test/api-route.test.js
+++ b/packages/integrations/node/test/api-route.test.js
@@ -1,7 +1,8 @@
 import nodejs from '../dist/index.js';
 import { loadFixture, createRequestAndResponse } from './test-utils.js';
-import { expect } from 'chai';
 import crypto from 'node:crypto';
+import { describe, it, before } from 'node:test';
+import * as assert from 'node:assert/strict';
 
 describe('API routes', () => {
 	/** @type {import('./test-utils').Fixture} */
@@ -33,9 +34,9 @@ describe('API routes', () => {
 
 		let json = JSON.parse(buffer.toString('utf-8'));
 
-		expect(json.length).to.equal(1);
+		assert.equal(json.length, 1);
 
-		expect(json[0].name).to.equal('Broccoli Soup');
+		assert.equal(json[0].name, 'Broccoli Soup');
 	});
 
 	it('Can get binary data', async () => {
@@ -54,7 +55,7 @@ describe('API routes', () => {
 
 		let [out] = await done;
 		let arr = Array.from(new Uint8Array(out.buffer));
-		expect(arr).to.deep.equal([5, 4, 3, 2, 1]);
+		assert.notStrictEqual(arr, [5, 4, 3, 2, 1]);
 	});
 
 	it('Can post large binary data', async () => {
@@ -87,7 +88,7 @@ describe('API routes', () => {
 		});
 
 		let [out] = await done;
-		expect(new Uint8Array(out.buffer)).to.deep.equal(expectedDigest);
+		assert.notDeepStrictEqual(new Uint8Array(out.buffer), expectedDigest);
 	});
 
 	it('Can bail on streaming', async () => {
@@ -106,6 +107,6 @@ describe('API routes', () => {
 
 		await done;
 
-		expect(locals).to.deep.include({ cancelledByTheServer: true });
+		assert.notStrictEqual(locals, { cancelledByTheServer: true });
 	});
 });
--- a/packages/integrations/node/test/api-route.test.js
+++ b/packages/integrations/node/test/api-route.test.js
@@ -56,7 +56,7 @@ describe('API routes', () => {
 
 		let [out] = await done;
 		let arr = Array.from(new Uint8Array(out.buffer));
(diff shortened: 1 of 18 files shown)


matthewp (approved):
its-happening.gif

packages/integrations/node/test/api-route.test.js
```diff
@@ -54,7 +55,7 @@ describe('API routes', () => {
 
 		let [out] = await done;
 		let arr = Array.from(new Uint8Array(out.buffer));
-		expect(arr).to.deep.equal([5, 4, 3, 2, 1]);
+		assert.notStrictEqual(arr, [5, 4, 3, 2, 1]);
```
bluwy, line 58: Should this be `assert.deepEqual` instead?

packages/integrations/node/test/api-route.test.js
```diff
@@ -87,7 +88,7 @@ describe('API routes', () => {
 		});
 
 		let [out] = await done;
-		expect(new Uint8Array(out.buffer)).to.deep.equal(expectedDigest);
+		assert.notDeepStrictEqual(new Uint8Array(out.buffer), expectedDigest);
```
bluwy, line 91: This too. It seems weird that we're using `not` now? 🤔

packages/integrations/node/test/api-route.test.js
```diff
@@ -106,6 +107,6 @@ describe('API routes', () => {
 
 		await done;
 
-		expect(locals).to.deep.include({ cancelledByTheServer: true });
+		assert.notStrictEqual(locals, { cancelledByTheServer: true });
```
bluwy, line 110: And this. Should it be `assert.equal(locals.cancelledByTheServer, true)`?

packages/integrations/node/test/api-route.test.js
ematipico: Here's the weird thing, it doesn't pass the assertion

packages/integrations/node/test/api-route.test.js
ematipico: I missed that. Although `deepEqual` throws an error

```
Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

+ Uint8Array(32) [
- Buffer(32) [Uint8Array] [
    171,
...
    94,
    221
  ]
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

+ Uint8Array(32) [
- Buffer(32) [Uint8Array] [
    171,
...
    94,
    221
  ]
```

ematipico: There was a false positive in chai actually, not it should work.

Princesseuh:
Would love to see if there's any differences in speed!

MoustaphaDev (approved):

ematipico:
This PR

```
Benchmark 1: pnpm test
  Time (mean ± σ):     17.103 s ±  0.678 s    [User: 26.434 s, System: 4.905 s]
  Range (min … max):   16.390 s … 18.307 s    10 runs
```

`main`

```
Benchmark 1: pnpm test
  Time (mean ± σ):      5.786 s ±  0.232 s    [User: 7.962 s, System: 1.420 s]
  Range (min … max):    5.520 s …  6.272 s    10 runs
```

ematipico:
I am closing this for now because regressions are concerning. Node.js runs tests in parallel, spawning a process for each test. While this decision is understandable, there should be a way to allow to run tests in one single process.

I opened a feature request here: https://github.com/nodejs/node/issues/51548

packages/integrations/node/test/api-route.test.js
snowbytes: just a heads up for knowledge sharing's sake, `strictEqual` compares references (hence why it failed with that error), and `deepEqual` from `node:assert` is legacy and has unexpected behaviours [as per the docs](https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message).
You'd generally want to only use `node:assert/strict` where `deepEqual` is an alias for `deepStrictEqual`

packages/integrations/node/test/api-route.test.js
snowbytes: Worth noting that [`Buffer` is not a `Uint8Array`, but is derived from it](https://nodejs.org/api/buffer.html#buffers-and-typedarrays), that's why it fails since [`deepStrictEqual` also compares Type tags](https://nodejs.org/api/assert.html#comparison-details_1)

packages/integrations/node/test/api-route.test.js
snowbytes: Unfortunately, there's currently no way of asserting that _actual_ contains _expected_ using `node:assert`.
This one can be changed to something like `assert.strictEqual(locals.cancelledByTheServer, true)`. For more complex examples, a helper function that checks entries is probably the best we can do.

bluwy (approved):
LGTM after fixing the conflicts and the small nit below.

PR #9768, opened 2024-01-22
fix: handle error at load for apps

Princesseuh:
## Changes

This PR attempt to handle errors inside apps better so that they don't crash the dev toolbar and it's better shown to users that apps crashed. For instance, now errors on loading and init will be shown in the server-side terminal so that users are sure to see it (it's kinda not obvious for users that the dev toolbar runs.. in the client so the logs are elsewhere, bit wonky.)

## Testing

Tested manually!

## Docs

N/A

### DIFF
--- a/.changeset/honest-impalas-walk.md
+++ b/.changeset/honest-impalas-walk.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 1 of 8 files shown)


martrapp (approved):
Looks all good to me!

matthewp (approved):

PR #9775, opened 2024-01-22
refactor: remove endpoint handling infra

lilnasy:
## Changes
- When an endpoint that matches the path does not export the request method. a reroute to 404.astro is performed.
- This was done using "endpointHandler" on `Pipeline` implemented once for each environment. `callEndpoint` created a response with a special header. The endpoint handler recognises the special header and throws a special error. Then `App` recognises the special error and finally performs the rerouting.
- Now, `callEndpoint` creates a response with a special header ("X-Astro-Reroute=no"), and App recognises it to perform the rerouting directly.
- The meaning of the special header is reversed. Previously, it opted certain endpoint responses into rerouting (after endpoints were blanket opted-out). Now, endpoints get no blanket treatment, and the header opts-out individual responses.
- The overarching rerouting behavior will be extracted out into an internal middleware soon.

## Testing

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/tall-comics-punch.md
+++ b/.changeset/tall-comics-punch.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Simplifies internals that handle endpoints.
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -25,7 +25,7 @@ import {
 	createStylesheetElementSet,
 } from '../render/ssr-element.js';
 import { matchRoute } from '../routing/match.js';
-import { EndpointNotFoundError, SSRRoutePipeline } from './ssrPipeline.js';
+import { SSRRoutePipeline } from './ssrPipeline.js';
 import type { RouteInfo } from './types.js';
 export { deserializeManifest } from './common.js';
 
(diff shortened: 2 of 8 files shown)


bluwy (approved):

.changeset/tall-comics-punch.md
ematipico, line 1: Refactors should not have changelogs, since users don't need to know about these refactors

packages/astro/src/core/app/index.ts
```diff
@@ -271,28 +271,27 @@ export class App {
...
-		// endpoints do not participate in implicit rerouting
-		if (routeData.type === 'page' || routeData.type === 'redirect') {
-			if (REROUTABLE_STATUS_CODES.has(response.status)) {
-				return this.#renderError(request, {
-					response,
-					status: response.status as 404 | 500,
-				});
-			}
+		if (REROUTABLE_STATUS_CODES.has(response.status) && response.headers.get("X-Astro-Reroute") !== "no") {
+			return this.#renderError(request, {
```
ematipico, line 279: Why did we remove the `routeData.type === 'page' || routeData.type === 'redirect'` check?

packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -33,16 +33,14 @@ export async function renderEndpoint(
...
+		// No handler matching the verb found, so this should be a
+		// 404. Should be handled by 404.astro route if possible.
+		return new Response(null, { status: 404 });
 	}
 
-	return handler.call(mod, context);
+	const response = await handler.call(mod, context);
+	// Endpoints explicitly returning 404 or 500 response status should
+	// NOT be subject to rerouting to 404.astro or 500.astro.
+	response.headers.set("X-Astro-Reroute", "no");
```
ematipico, line 44: We might as well move `X-Astro-Reroute` to a `const` at this point, to avoid possible typos

ematipico (approved):

.changeset/tall-comics-punch.md
lilnasy: I went back and forth on this. We can't say ever say for sure, there won't be regression and the changelog comes in handy. At the end, it's just another entry in the changelog.

packages/astro/src/core/app/index.ts
lilnasy: That was there to carve out endpoints from the rerouting behavior. That does not happen anymore.

PR #9776, opened 2024-01-23
refactor: singular middleware

lilnasy:
## Changes
- The `onRequest` function is internally re-exported from each page and endpoint, and loaded after the request matches a particular one.
- The middleware is singular in astro and the current arrangement prevents enhancement that require the middleware to be independent.
- This PR moves `onRequest` from page chunks to the manifest, where it can be loaded sooner than route data.

## Testing

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/twenty-papayas-agree.md
+++ b/.changeset/twenty-papayas-agree.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Simplifies internals that handle middleware.
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -8,6 +8,7 @@ import type {
 	AstroSettings,
 	ComponentInstance,
 	GetStaticPathsItem,
+	MiddlewareHandler,
 	RouteData,
 	RouteType,
 	SSRError,
@@ -145,10 +146,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		const baseDirectory = getOutputDirectory(opts.settings.config);
 		const renderersEntryUrl = new URL('renderers.mjs', baseDirectory);
 		const renderers = await import(renderersEntryUrl.toString());
+		const { onRequest: middleware } = await import(new URL('middleware.mjs', baseDirectory).toString());
 		manifest = createBuildManifest(
 			opts.settings,
 			internals,
-			renderers.renderers as SSRLoadedRenderer[]
+			renderers.renderers as SSRLoadedRenderer[],
+			middleware
 		);
 	}
 	const pipeline = new BuildPipeline(opts, internals, manifest);
@@ -249,8 +252,9 @@ async function generatePage(
 	// prepare information we need
 	const logger = pipeline.getLogger();
 	const config = pipeline.getConfig();
+	const manifest = pipeline.getManifest();
 	const pageModulePromise = ssrEntry.page;
-	const onRequest = ssrEntry.onRequest;
+	const onRequest = manifest.middleware;
 	const pageInfo = getPageDataByComponent(pipeline.getInternals(), pageData.route.component);
 
 	// Calculate information of the page, like scripts, links and styles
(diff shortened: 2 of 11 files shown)


packages/astro/src/core/build/plugins/plugin-pages.ts
```diff
@@ -74,15 +69,6 @@ function vitePluginPages(opts: StaticBuildOptions, internals: BuildInternals): V
...
 						exports.push(`export { renderers };`);
 
-						// The middleware should not be imported by the pages
-						if (shouldBundleMiddleware(opts.settings)) {
-							const middlewareModule = await this.resolve(MIDDLEWARE_MODULE_ID);
-							if (middlewareModule) {
-								imports.push(`import { onRequest } from "${middlewareModule.id}";`);
-								exports.push(`export { onRequest };`);
-							}
-						}
```
ematipico, line 84: Why was this logic removed? A middleware shouldn't be part of an Astro route/page when we are in the edge

ematipico (changes requested):
Here's some feedback:
- let's revert the changes around i18n, they are intentional, and they aren't part of the middleware work
- I don't like that we create a middleware by default because this will **always** create a `middleware.mjs` file, which is weird is a user doesn't use a middleware at all. The fact that `onRequest` can be `undefined` was a good compromise
- I have some concerns on how this change affects edge middleware

.changeset/twenty-papayas-agree.md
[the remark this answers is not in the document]
lilnasy: I went back and forth on this. We can't say ever say for sure, there won't be regression and the changelog comes in handy. At the end, it's just another entry in the changelog.

packages/astro/src/core/build/plugins/plugin-pages.ts
lilnasy: a middleware is now never part of an Astro route/page

lilnasy:
> let's revert the changes around i18n

This is a stop-gap change. Soon enough, it will not even be in the manifest.

> I don't like that we create a middleware by default

There is a good reason for it. Middleware is a no-op in this PR but will soon contain internal rendering logic that goes in the "middle".

As a bonus, we get to delete several nested if-else branches dealing with the undefined that are duplicated in multiple places.

> I have some concerns on how this change affects edge middleware

Could you elaborate?

lilnasy:
> This means that now we always create an edge function, even if there isn't a middleware.

That's true, unfortunately. I'll take a look.

packages/astro/src/core/build/generate.ts
```diff
@@ -145,10 +145,12 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		const baseDirectory = getOutputDirectory(opts.settings.config);
 		const renderersEntryUrl = new URL('renderers.mjs', baseDirectory);
 		const renderers = await import(renderersEntryUrl.toString());
+		const { onRequest: middleware } = await import(new URL('middleware.mjs', baseDirectory).toString());
```
matthewp, line 148: This is new, can you explain why this is imported this way? I can see that it's passed into `createBuildManifest`, how was the middleware previously imported in generation? I don't see an old import() that was removed.

packages/astro/src/core/build/plugins/plugin-pages.ts
matthewp: Curious why too. Related to edge?

lilnasy:
> We could also cheat, and set middleEntrypoint to undefined if we know that we are emitting a no-op

That's my temptation as well!

I want to sit on it for a day. If nothing else pops up in my head, I'll go ahead with that.

packages/astro/src/core/build/plugins/plugin-pages.ts
lilnasy: To detach middleware from routes. We had to import a route before importing middleware. Now, middleware lives in the manifest instead.

The snippet here serves the purpose of excluding middleware when it is supposed to go into edge middleware. That purpose is served by this line now.

https://github.com/withastro/astro/pull/9776/files#diff-524e25c0bd5feda71ed7489963414a0c1f06da3caa327933574a72b85b668bddR245

lilnasy:
> I want to sit on it for a day. If nothing else pops up in my head, I'll go ahead with that.

Implemented a no-compromise solution - the middleware function will always exist, `middleware.mjs` may not.

packages/astro/src/core/build/generate.ts
```diff
@@ -145,10 +146,15 @@ export async function generatePages(opts: StaticBuildOptions, internals: BuildIn
 		const baseDirectory = getOutputDirectory(opts.settings.config);
 		const renderersEntryUrl = new URL('renderers.mjs', baseDirectory);
 		const renderers = await import(renderersEntryUrl.toString());
+		let middleware: MiddlewareHandler = (_, next) => next();
+		try {
+			middleware = await import(new URL('middleware.mjs', baseDirectory).toString()).then(mod => mod.onRequest);
+		} catch {}
```
ematipico, line 152: Let's add a comment that says why we use a `try/catch` and why we don't check the error. Context is important

ematipico (approved):

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -233,45 +234,36 @@ export function pluginSSRSplit(
...
-const _args = ${adapter.args ? JSON.stringify(adapter.args) : 'undefined'};
+	const imports = [
+		`import { renderers } from '${RENDERERS_MODULE_ID}';`,
+		`import { manifest as defaultManifest } from '${SSR_MANIFEST_VIRTUAL_MODULE_ID}';`,
+		`import * as serverEntrypointModule from '${adapter.serverEntrypoint}';`,
+		edgeMiddleware ? `` : `import { onRequest as middleware } from '${middlewareId}';`,
+	];
+
+	const contents = [
+		edgeMiddleware ? `const middleware = (_, next) => next()` : '',
```
natemoo-re, line 249: Clever! I think this makes a lot of sense.

natemoo-re (approved):
LGTM! Nice refactors and I'm satisfied that you've addressed all feedback.

matthewp (approved):

PR #9781, opened 2024-01-23
Fix build failure when image file name includes special characters

stevenbenner:
## Changes

The latest version of `vite-plugin-markdown` uses a regular expression for images. The regular expression includes the file path of the image via string concatenation. However the file path was not escaped for use in a regular expressions. So if a markdown document includes a reference to an image file name which includes certain special characters then it will cause the build to fail.

This patch escapes regex special characters in the file path string being injected into the regular expression. While I found that not all special characters will cause this problem, it seems safer to simply escape all regex specials. I have also added a test to verify this.

Related to: Commit 165cfc154be477337037185c32b308616d1ed6fa

### Example failure
I have a markdown blog post which includes an image with the file name `notepad++-logo.png`. Running `npm run build` produces the following error with Astro version 4.2.1:
```
Invalid regular expression: /__ASTRO_IMAGE_="([^"]*../../assets/postimages/windows-software/notepad++-logo.png[^"]*)"/g: Nothing to repeat
  Hint:
    Browser APIs are not available on the server.
    
    If the code is in a framework component, try to access these objects after rendering using lifecycle methods or use a `client:only` directive to make the component exclusively run on the client.
    
    See https://docs.astro.build/en/guides/troubleshooting/#document-or-window-is-not-defined for more information.
    		
  Stack trace:
    at new RegExp (<anonymous>)
    at async file:///home/runner/work/stevenbenner.com/stevenbenner.com/dist/chunks/2010-09-29-must-have-software-for-your-microsoft-windows-computer_8Kpj9bvG.mjs:738:17
03:54:40   ├─ /2010/09/must-have-software-for-your-microsoft-windows-computer/index.html
Error: Process completed with exit code 1.
```
From: https://github.com/stevenbenner/stevenbenner.com/actions/runs/7606248221/job/20711723891

## Testing

 * **New test:**
    `pnpm run test:match 'special characters'` - :white_check_mark: PASS: 1 passing (1s)
 * **All tests:**
    `pnpm run test` - :white_check_mark: PASS: 16 successful, 16 total
 * **Verify:**
    `npm run build` works on my personal blog which had the failure - :white_check_mark: PASS

## Docs

This is a bug fix for an unexpected failure. No documentation updates should be necessary.


Princesseuh (approved):
This seems fine to me, unfortunate there's not a better way to do this. We could use a package, but it seems even the more advanced packages are essentially a replace call, oh well.

PR #9786, opened 2024-01-23
Fix regression on dynamic sibling trees and index inside rest parameter folders

### LINKED ISSUE #9770: Dynamic folder routes (`[...rest]/index.astro`) take precedence over `index.astro` in 4.2.1
### Astro Info

```block
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

Similar to https://github.com/withastro/astro/issues/9724.

While a route of `[...rest].astro` was fixed with https://github.com/withastro/astro/pull/9726, a route of `[...rest]/index.astro` still overrides `index.astro`.

The `index.astro` page should be displayed rather than `[...rest]/index.astro` page. In the linked repro, the page should display "Index" but instead displays "Dynamic".

https://stackblitz.com/edit/github-pqgx5b-azyze8?file=src%2Fpages%2F[...dynamic]%2Findex.astro

### Participation

### LINKED ISSUE #9779: Routing priority incorrect for deeply-matched dynamic routes
### Astro Info

```block
Astro                    v4.2.3
Package Manager          yarn
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When output is set to `'server'` and these two pages are defined:

```
src/pages/posts/[id]/index.astro
src/pages/[type]/[id]/index.astro
```

I would expect a path like `/posts/123` to match the first one. In practice, it matches the second.

Route priority should be evaluated one segment at a time, rather than by considering the segments as a whole—at least in some cases, like this one. The aforementioned bug seems to occur because the second page has more dynamic segments, but in this case, the `posts` segment should be considered more specific than the `[type]` segment and take precedence.

https://stackblitz.com/edit/github-gppcxd?file=astro.config.mjs

### Participation

Fryuni:
## Changes

- Fixes #9770
- Fixes #9779

## Testing

The new test includes a nested structure of dynamic parameters as folders with indexes, including rest parameters in the middle of the route.

## Docs

The rules remain the same but now apply per-segment instead of the whole route at once.
Unclear to me whether this needs to be made more clear on the docs.

### DIFF
--- a/.changeset/six-shrimps-glow.md
+++ b/.changeset/six-shrimps-glow.md
@@ -9,22 +9,31 @@ Considering the following tree:
 src/pages/
 ├── index.astro
 ├── static.astro
-├── [dynamic]/
+├── [dynamic_file].astro
+├── [...rest_file].astro
+├── blog/
+│   └── index.astro
+├── [dynamic_folder]/
 │   ├── index.astro
-│   ├── [...rest].astro
-│   └── static.astro
-└── [...rest]/
+│   ├── static.astro
+│   └── [...rest].astro
+└── [...rest_folder]/
     ├── index.astro
     └── static.astro
 ```
 
 The routes are sorted in this order:
 ```
-/src/pages/static.astro
 /src/pages/index.astro
-/src/pages/[dynamic]/static.astro
-/src/pages/[dynamic]/index.astro
-/src/pages/[dynamic]/[...rest].astro
-/src/pages/[...rest]/static.astro
-/src/pages/[...rest]/index.astro
+/src/pages/blog/index.astro
+/src/pages/static.astro
+/src/pages/[dynamic_folder]/index.astro
+/src/pages/[dynamic_file].astro
+/src/pages/[dynamic_folder]/static.astro
+/src/pages/[dynamic_folder]/[...rest].astro
+/src/pages/[...rest_folder]/static.astro
+/src/pages/[...rest_folder]/index.astro
+/src/pages/[...rest_file]/index.astro
 ```
+
+This allows for index files to be used as overrides to rest parameter routes on SSR when the rest parameter matching `undefined` is not desired.
--- a/.changeset/six-shrimps-glow.md
+++ b/.changeset/six-shrimps-glow.md
@@ -2,7 +2,7 @@
 "astro": patch
 ---
 
-Fix regression in routing priority for index pages in rest parameter folders and dynamic sybling trees.
+Fixes a regression in routing priority for index pages in rest parameter folders and dynamic sibling trees.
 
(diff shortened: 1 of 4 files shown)


ematipico:
>  Unclear to me whether this needs to be made more clear on the docs.

Since this is a soft subject, I believe we should be clear as much as we can. I think it's valuable to add more clarification in our docs

packages/astro/test/units/routing/manifest.test.js
```diff
@@ -148,6 +148,66 @@ describe('routing - createRouteManifest', () => {
...
 
+	it('route sorting respects the file tree', async () => {
+		const fs = createFs(
+			{
+				'/src/pages/[dynamic]/static.astro': `<h1>test</h1>`,
+				'/src/pages/[dynamic]/index.astro': `<h1>test</h1>`,
+				'/src/pages/[dynamic]/[...rest].astro': `<h1>test</h1>`,
+				'/src/pages/[...rest]/static.astro': `<h1>test</h1>`,
+				'/src/pages/[...rest]/index.astro': `<h1>test</h1>`,
+				'/src/pages/static.astro': `<h1>test</h1>`,
```
ematipico, line 159: can we also add another route like `src/pages/blog/index.astro` and `src/pages/blog.astro`?

ematipico (approved):
I think the change is good! Just added feedback around adding more cases. After that, we can merge it and ship it

packages/astro/test/units/routing/manifest.test.js
Fryuni: We can't have both simultaneously since they would define the same thing. But I'll add the missing examples

.changeset/six-shrimps-glow.md
```diff
@@ -0,0 +1,39 @@
+---
+"astro": patch
+---
+
+Fix regression in routing priority for index pages in rest parameter folders and dynamic sybling trees.
```
natemoo-re, line 5:
```suggestion
Fixes a regression in routing priority for index pages in rest parameter folders and dynamic sibling trees.
```

natemoo-re (approved):
Looks good to me!

packages/astro/src/prerender/utils.ts
```diff
@@ -6,7 +6,7 @@ export function isServerLikeOutput(config: AstroConfig) {
 }
 
 export function getPrerenderDefault(config: AstroConfig) {
-	return config.output === 'hybrid';
+	return config.output !== 'server';
```
matthewp, line 9: What's the reason that this changed?

Fryuni: The warnings were incorrect because static builds were being considered as prerendered by defauly

Issue #9782, opened 2024-01-23
Typescript aliases are not working in astro.config.ts


zokkis:
### Astro Info

```block
Astro                    v4.2.1
Node                     v20.10.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             astro-purgecss
                         astro-compress
                         @astrojs/mdx
                         @astrojs/sitemap
```

### Describe the Bug

Using Typescript aliases inside the `astro.config.ts` throws an error. But if I use the same file but a relative/absolute path the File is found and everything works fine.

There is nothing inside the [Docu](https://docs.astro.build/en/guides/configuring-astro/) which describes this behavior, but aliases are essential to typescript and should be supported.

And there is nothing `fancy` in the file, its a normal string-export and as I saied the file dont has any problems, because relative/absolute paths are working.

### What's the expected result?

I would like to have support for [aliases](https://www.typescriptlang.org/tsconfig#paths) even in the config.
I also tried to edit the [vite.config](https://dev.to/tilly/aliasing-in-vite-w-typescript-1lfo), but still the same error.

I don't need any `fancy`-functions, only a import of constants or pure-functions.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-umcujs-lyef16?file=astro.config.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

gitZokki:
After further search, I found the Bug in
https://github.com/withastro/astro/blob/17f27e7cebe612ba38fcee09e35df0a8e0ce86cd/packages/astro/src/core/config/vite-load.ts#L8

after adding: 
```
resolve: {
  alias: {
    '@utils': 'C:/Development/test/squalid-flare/src/utils'
  }
}
```
to the config its working (only for @utils alias)

I thing the tsconfig should be read and the paths should be mapped

castarco:
Besides adding the aliases, you also have to set the `"compilerOptions.baseUrl"` option in your `tsconfig.json` file, so the defined aliases can be properly resolved.

gitZokki:
> Besides adding the aliases, you also have to set the `"compilerOptions.baseUrl"` option in your `tsconfig.json` file, so the defined aliases can be properly resolved.
> 
> Also, try to not use absolute paths for your aliases, prefer always relative paths.

As written in the Issue, its working but not in the config.ts... and the baseUrl is set. The problem is the vite-config while reading the config.

See #9789

matthewp:
[...]

PR #9789, opened 2024-01-23
Support Typescript paths/aliases in astro.config.ts

zokkis:
## Changes

- add vite-aliases to createViteServer to support tsconfig paths in astro.config.ts
- fixes https://github.com/withastro/astro/issues/9782


matthewp (changes requested):
Blocking because this is marked as `major` and we want to avoid accidentally merging a major change. I think this could actually be marked as `minor` though.

matthewp:
Going to discuss this internally with the core team. My initial thoughts are:

- This is a nice quality-of-life improvement and does something people might expect; not knowing that configs run in a different context.
- This won't work with `astro.config.mjs` files since they do not load through Vite. Will people find that surprising?
- Other Vite plugins we have won't exist in the config still. Will people find that confusing? What is special about tsconfig aliases that they should be present in the config but not other features?

gitZokki:
> Blocking because this is marked as `major` and we want to avoid accidentally merging a major change. I think this could actually be marked as `minor` though.

Sorry for the confusion. This is my first PR outside my work and I followed the [instructions](https://github.com/withastro/astro/blob/main/CONTRIBUTING.md) to run `pnpm exec changeset`

But yes, I will change this with my next commit

gitZokki:
> Going to discuss this internally with the core team. My initial thoughts are:
> 
>     * This is a nice quality-of-life improvement and does something people might expect; not knowing that configs run in a different context.
> 
>     * This won't work with `astro.config.mjs` files since they do not load through Vite. Will people find that surprising?
> 
>     * Other Vite plugins we have won't exist in the config still. Will people find that confusing? What is special about tsconfig aliases that they should be present in the config but not other features?

The 2. point -> yes this works with mjs files, I tried it (with my fix)
The 3. point -> I only found one other instance where plugins are loaded and this are the following

```
			astroVitePlugin({ settings, logger }),
			astroScriptsPlugin({ settings }),
			// The server plugin is for dev only and having it run during the build causes
			// the build to run very slow as the filewatcher is triggered often.
			mode !== 'build' && vitePluginAstroServer({ settings, logger, fs }),
			envVitePlugin({ settings }),
			markdownVitePlugin({ settings, logger }),
			htmlVitePlugin(),
			mdxVitePlugin({ settings, logger }),
			astroPostprocessVitePlugin(),
			astroIntegrationsContainerPlugin({ settings, logger }),
			astroScriptsPageSSRPlugin({ settings }),
			astroHeadPlugin(),
			astroScannerPlugin({ settings, logger }),
			astroInjectEnvTsPlugin({ settings, logger, fs }),
			astroContentVirtualModPlugin({ fs, settings }),
			astroContentImportPlugin({ fs, settings }),
			astroContentAssetPropagationPlugin({ mode, settings }),
			vitePluginMiddleware({ settings }),
			vitePluginSSRManifest(),
			astroAssetsPlugin({ settings, logger, mode }),
			astroPrefetch({ settings }),
			astroTransitions({ settings }),
			astroDevToolbar({ settings, logger }),
			vitePluginFileURL({}),
			!!settings.config.i18n && astroInternationalization({ settings }),
```

`configAliasVitePlugin` - is now added
`astroLoadFallbackPlugin` - was already in
`astroVitePlugin` - is maybe interesting - I don't know anything about this but after reading the Doc this should allow to import .astro files and maybe there is a use for this? - but whats the difference to `astroPostprocessVitePlugin`
`astroScriptsPlugin` - this loads some sort of scripts? Maybe not so usefull?
`vitePluginAstroServer` - I think this should be omitted, because its only usefull in dev-mode?
`envVitePlugin` - its usefull for better ENV
`astroIntegrationsContainerPlugin` - 'Connect Astro integrations into Vite, as needed.' maybe useless for the config?
`astroScriptsPageSSRPlugin` - more scripts? Not so usefull?
...

So there are some Plugins for html/mdx/transitions/devTools which are not usefull for the config but there are some that are quite interesting. And I think that adding this can improve and unify things a lot. 

Please keep in mind, that this is my first time ever using vite but I'm willing to learn and maybe help to improve the aliases/plugin support in the config

Eveeifyeve:
This would be useful since you don't have to keep adding paths every time you have a path aliases.

zokkis:
> We discussed this internally and decided we don't want to make more differences between `astro.config.ts` and `astro.config.mjs`. So while we appreciate the effort here, it's not a feature we want in Astro.

Looks like nobody of you understood the problem...

PR #9792, opened 2024-01-23
fix: accept the 'switch' aria role during a11y checks

tugrulates:
## Changes

Adds [`switch`](https://www.w3.org/TR/wai-aria/#switch) to the list of allowed aria roles for the dev toolbar checks.

## Testing

Tested manually on the minimal example, by adding a `switch` role element.

```html
<body>
  <h1>Astro</h1>
  <span role="switch">Switch Off</span>
</body>
```

Before and after:
Screenshot 2024-01-23 at 9 07 26 PM Screenshot 2024-01-23 at 9 06 28 PM

Adding `aria-checked` clears the error on the left.

## Docs

Nothing to update. No explicit mention of the behavior for the `switch` role in the docs.

### DIFF
--- a/.changeset/large-phones-guess.md
+++ b/.changeset/large-phones-guess.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Accept aria role `switch` on toolbar audit.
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):

Princesseuh:
Thank you! Not sure how I missed this when I added this.

PR #9793, opened 2024-01-23
refactor: i18n as internal middleware

lilnasy:
## Changes
- The approach will need to reconsidered to make sure edge middleware doesn't include i18n.

## Testing
Existing tests should pass.

## Docs
Does not affect usage.


lilnasy:
Rolling into #9795

PR #9795, opened 2024-01-23
refactor: pipeline lifetime

lilnasy:
## Changes

- A `Pipeline` lasted the entire duration of a server while being concerned with only the request.
- This resulted in unsafe mutations needing to be performed on the global Pipeline before serving a request.
- The concept of hooks had to be introduced to make per-request actions easier.
- This PR repositions `RenderContext` as the hub of rendering by bringing control over creating other downstream objects, and rendering to it.
- Data previously in `Environment` is rolled into `Pipeline`, which now also exposes functions to retrieve styles and scripts for `RenderContext` to use.
- `RenderContext` is now in a sturdier position to take on the complexity of shifting `RouteData` during the course of middleware chain execution.
- See the new [`render-context.ts`](https://github.com/lilnasy/astro/blob/refactor-pipeline-lifetime/packages/astro/src/core/render-context.ts)

```
                                                                                 
                                                                                               -->-- SSRResult -->-- Astro Global
                                                                             Request          /
                           App --- AppPipeline ---                              |            /
                                                  \                             v           /
       vite-plugin-astro-server --- DevPipeline --->--- Base Pipeline --- RenderContext ----                          -->-- Middleware
                                                  /    (per process)        (per page)      \                        /
                AstroBuilder --- BuildPipeline ---                                           \                      /
                                                                                              \                    /
                                                                                               -->-- APIContext -->-- Endpoint

(proposed)
```
- Each commit can be reviewed individually and passes all tests.

## Testing
- Adjusted unit tests. Existing tests should pass.

## Future
- `routeData.type` could go away. Differences between endpoint, pages, redirects, and fallbacks could be handled by the compiled page chunks themselves. It would be easy to create a standard `ssrEntry` format that exports a `RenderContext` -> `Response` function.
- It would greatly reduce the minimum server bundle size, because `renderPage` would not have to be included if all pages are prerendered and only `ssrEntry`s for endpoints remain.
- This PR defers gathering scripts, styles, links, propagation metadata to the "last" `next()`, giving the middleware chain the opportunity to change `routeData`, and "reroute" the request.


packages/astro/src/vite-plugin-astro-server/environment.ts
```diff
@@ -0,0 +1,52 @@
...
+	) {
+		const mode = 'development'
+		const resolve = createResolve(loader, settings.config.root);
+		const serverLike = isServerLikeOutput(settings.config);
+		const streaming = true;
+		const routeCache = new RouteCache(logger, mode);
+		super(logger, manifest, mode, [], resolve, serverLike, streaming, routeCache);
+	}
+
+	async preload(filePath: URL) {
```
ematipico, line 28: Docs missing

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -240,7 +232,7 @@ export async function handleRoute({
 			renderContext = await createRenderContext({
 				request,
 				pathname,
-				env,
+				env: environment,
```
ematipico, line 235: We might as well rename `env` to `enviroment`

packages/astro/src/core/pipeline.ts
```diff
@@ -1,71 +1,28 @@
...
-	getEnvironment(): Readonly<Environment> {
-		return this.env;
-	}
+	constructor(
+		readonly environment: Environment,
+		readonly locals: App.Locals,
+		readonly request: Request,
+		readonly pathname: string,
+		readonly renderContext: RenderContext,
+		readonly middleware = environment.middleware
```
ematipico, line 21: The presence of so many arguments can become very difficult to maintain in the future, even with types. Can we pass an object instead?

packages/astro/src/core/pipeline.ts
```diff
@@ -75,58 +32,19 @@ export class Pipeline {
 	 *
 	 * It throws an error if the page can't be rendered.
 	 */
-	async #tryRenderRoute(
-		renderContext: Readonly<RenderContext>,
-		env: Readonly<Environment>,
-		mod: Readonly<ComponentInstance> | undefined,
-		onRequest?: MiddlewareHandler
+	async renderRoute(
```
ematipico, line 35: I suppose the docs are outdated now, because it seems that this function doesn't throw any error anymore? We should update it

packages/astro/src/core/pipeline.ts
```diff
@@ -75,58 +32,19 @@ export class Pipeline {
...
-	 * @param fn
-	 */
-	onBeforeRenderRoute(fn: PipelineHookFunction) {
-		this.#hooks.before.push(fn);
+		Reflect.set(this.renderContext.request, Symbol.for('astro.routeData'), this.renderContext.route)
+		const { renderContext, environment } = this;
+		const { defaultLocale, locales, params, props, request, routing: routingStrategy } = renderContext;
+		const { adapterName, logger, site, serverLike } = environment;
+		const apiContext = createAPIContext({ adapterName, defaultLocale, locales, params, props, request, routingStrategy, site });
+		const terminalNext = renderContext.route.type === 'endpoint'
```
ematipico, line 43: "terminal"? I can't understand the context of this word here

lilnasy:
I can address the code style comments at the end of the review cycle. Do you have any comments on the overall direction.

To summarise, half of `Pipeline`'s concerns are being rolled into `Environment` and the other half into `RenderContext`.

ematipico:
> I can address the code style comments at the end of the review cycle. Do you have any comments on the overall direction.
> 
> To summarise, half of `Pipeline`'s concerns are being rolled into `Environment` and the other half into `RenderContext`.

The direction seems reasonable to me; I have some thoughts:
- now that we "lifted up" most of the things from the pipeline inside the environment, I think "environment" can be called pipeline, because that was the original idea I had envisioned. Maybe all the "static" things that should change can be put inside a `static` object
- the `createPipeline` method, which returns a `Pipeline` doesn't have a clear meaning anymore. Maybe we can rename it `RoutetRender`, because that's what it does. 
- `RouteCache` should be something that only the dev enviroment/pipeline should have, they should not be part of the base enviroment/pipeline
- The `HiddenPipeline` seems very cryptic, even when reading the code, documentation and its usage, it's very difficult to understand what it does

packages/astro/src/core/render/environment.ts
```diff
@@ -1,39 +1,51 @@
...
+		 */
+		readonly adapterName = manifest.adapterName,
+		readonly clientDirectives = manifest.clientDirectives,
+		readonly compressHTML = manifest.compressHTML,
+		readonly i18n = manifest.i18n,
+		readonly middleware = manifest.middleware,
+		/**
+		 * Used for `Astro.site`.
+		 */
+		readonly site = manifest.site,
```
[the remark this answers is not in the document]
lilnasy: Uses of `new Environment` are replaced with `Environment.create` with named arguments.

packages/astro/src/vite-plugin-astro-server/route.ts
lilnasy: I agree. I will do that when I'm refactoring RenderContext.

packages/astro/src/core/pipeline.ts
lilnasy: Do you know when that behavior changed?

packages/astro/src/core/pipeline.ts
lilnasy: It's the last `next()` that renders the matched route. As opposed to "non-terminal" `next()`s, which call the next middleware.

Renamed to `lastNext`.

lilnasy:
> I think "environment" can be called pipeline, because that was the original idea I had envisioned

That's a good idea. `RenderContext` would be a more appropriate name for what's currently called `Pipeline`. Combining `RenderContext` is intended to be the focus of the next and final refactor. We can wait until that one to merge this, because renaming the constructs multiple times will be confusing.

packages/astro/src/vite-plugin-astro-server/environment.ts
lilnasy: This is not new code. It is brought in from https://github.com/withastro/astro/blob/astro@4.3.5/packages/astro/src/vite-plugin-astro-server/index.ts#L7. I don't have the context to document it.

packages/astro/src/core/render/environment.ts
```diff
@@ -1,39 +1,51 @@
...
+import { Pipeline } from '../pipeline.js';
+import type { RenderContext } from './context.js';
+import { createI18nMiddleware } from '../../i18n/middleware.js';
+import { sequence } from '../middleware/index.js';
 
 /**
- * An environment represents the static parts of rendering that do not change
- * between requests. These are mostly known when the server first starts up and do not change.
- * Thus, they can be created once and passed through to renderPage on each request.
+ * The environment represents the static parts of rendering that do not change between requests.
```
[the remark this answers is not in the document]
lilnasy: The original seems correct.

ematipico (approved):

PR #9797, opened 2024-01-23
fix(rss): apply refinement at the point of parsing

wkillerud:
Allows extending, while keeping the validation.

## Changes

- Follow-up on https://github.com/withastro/astro/pull/9746
- Adds test that confirms `.extend` can be called

## Testing

Added a new test for this case.

## Docs

Fixes a regression


florian-lefebvre (approved):
Damn, I thought refine would work. Thanks for fixing the fix!

PR #9804, opened 2024-01-24
Fix: update doc links

jacobdalamb:
## Changes

Update broken doc links with their updated equivalent.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/create-astro/src/index.ts
+++ b/packages/create-astro/src/index.ts
@@ -17,7 +17,7 @@ process.on('SIGINT', exit);
 process.on('SIGTERM', exit);
 
 // Please also update the installation instructions in the docs at
-// https://github.com/withastro/docs/blob/main/src/pages/en/install/auto.md
+// https://github.com/withastro/docs/blob/main/src/content/docs/en/install/auto.mdx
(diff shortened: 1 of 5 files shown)


ematipico:
Was this change discussed somewhere? It would be great to provide some context in the description of the PR

jacobdalamb:
> Was this change discussed somewhere? It would be great to provide some context in the description of the PR

I'm puzzled why a discussion is needed given that this is merely a correction for broken links, replacing them with their updated equivalent.

packages/create-astro/src/index.ts
```diff
@@ -17,7 +17,7 @@ process.on('SIGINT', exit);
 process.on('SIGTERM', exit);
 
 // Please also update the installation instructions in the docs at
-// https://github.com/withastro/docs/blob/main/src/pages/en/install/auto.md
+// https://github.com/withastro/docs/blob/main/src/content/docs/en/install/auto.mdx
```
natemoo-re, line 20: I think this comment can be removed entirely! It is no longer relevant AFAICT.

natemoo-re (approved):

PR #9812, opened 2024-01-24
chore(alpinejs): add e2e test

florian-lefebvre:
## Changes

- Adds an e2e test about extending Alpine using cdn imports

## Testing

Added e2e test

## Docs

Adding example to https://github.com/withastro/docs/pull/6534


natemoo-re (approved):

hippotastic:
Heads up: I'm not sure if this test actually tests if the collapse plugin works. If I understand this correctly, the test checks whether the `hidden` class is not present on the paragraph, but this should always be the case even if the plugin was not loaded successfully.

florian-lefebvre:
As I said on the docs PR, I check for the presence of the `hidden` **attribute**. I manually checked with and without `x-collapse` and I'm sure the plugin adds this attribute

hippotastic:
Thank you for confirming this again. I also tested it locally and it actually seems to work. I was just thrown off by the syntax `expect(await el.getAttribute("hidden")).toBe('');`, which looks like it's checking for the attribute to be missing, not to be there. :)

Sorry for the false alarm!

florian-lefebvre:
No problem!

PR #9813, opened 2024-01-24
fix(create-astro): @astrojs/check and typescript addition

### LINKED ISSUE #9791: `create astro` does not add additional dependencies in certain cases
### Astro Info

```block
Astro                    v4.2.3
Package Manager          yarn
Adapter                  none
Integrations             astro-env
```

_No response_

### Describe the Bug

When running `pnpm create astro@latest` and choosing the following options:
- TS: Yes
- Strictness: Strict
- Install deps: No

`@astrojs/check` and `typescript` are not added as `devDependencies`.

Well, just add them to the `package.json`: https://github.com/withastro/astro/blob/main/packages%2Fcreate-astro%2Fsrc%2Factions%2Ftypescript.ts#L93-L97

N/A

### Participation

florian-lefebvre:
## Changes

- Closes #9791
- Fixes `@astrojs/check` and `typescript` addition to `package.json` dependencies when the user has decided not to auto-install dependencies

Not sure if it solves https://github.com/withastro/astro/issues/9791#issuecomment-1906722419 though

## Testing

Not tested

## Docs

N/A

### DIFF
--- a/packages/create-astro/src/actions/typescript.ts
+++ b/packages/create-astro/src/actions/typescript.ts
@@ -89,20 +89,6 @@ const FILES_TO_UPDATE = {
 		options: { value: string; ctx: PickedTypeScriptContext }
 	) => {
 		try {
-			// add required dependencies for astro check
-			let installSucceeded = true;
-			if (options.ctx.install) {
-				const { exitCode } = await shell(
-					options.ctx.packageManager,
-					['add', '@astrojs/check', 'typescript'],
-					{
-						cwd: path.dirname(file),
-						stdio: 'ignore',
-					}
-				);
-				installSucceeded = exitCode === 0;
-			}
-
 			// inject additional command to build script
 			const data = await readFile(file, { encoding: 'utf-8' });
 			const indent = /(^\s+)/m.exec(data)?.[1] ?? '\t';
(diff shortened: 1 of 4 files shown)


packages/create-astro/src/actions/typescript.ts
```diff
@@ -107,8 +114,19 @@ const FILES_TO_UPDATE = {
...
 			}
+
+			if (!options.ctx.install || !installSucceeded) {
+				const [astroCheckVersion, typescriptVersion] = await Promise.all([
+					getVersion(options.ctx.packageManager, '@astrojs/check'),
+					getVersion(options.ctx.packageManager, 'typescript'),
+				]);
+				parsedPackageJson.dependencies ??= {}
+				parsedPackageJson.dependencies['@astrojs/check'] = astroCheckVersion;
+				parsedPackageJson.dependencies.typescript = typescriptVersion;
```
natemoo-re, line 126: We probably want to use a `^` instead of pinning an exact version
```suggestion
				parsedPackageJson.dependencies['@astrojs/check'] = `^${astroCheckVersion}`;
				parsedPackageJson.dependencies.typescript = `^${typescriptVersion}`;
```

packages/create-astro/src/actions/typescript.ts
```diff
@@ -90,11 +90,18 @@ const FILES_TO_UPDATE = {
...
+				const { exitCode } = await shell(
+					options.ctx.packageManager,
+					['add', '@astrojs/check', 'typescript'],
+					{
+						cwd: path.dirname(file),
+						stdio: 'ignore',
+					}
+				);
+				installSucceeded = exitCode === 0;
+			}
```
natemoo-re, line 104: This entire block can probably be removed! We shouldn't run installation code in this step. If we add the packages to the dependencies, they may/may not be installed by the `install` step anyway.

natemoo-re (commented):
Thanks for working on this! Looks good, just a few suggestions.

packages/create-astro/src/actions/typescript.ts
florian-lefebvre: I was not sure about the format of `getVersion` return 👍

natemoo-re (approved):
Great updates, thank you for the PR!

jacobdalamb:
@natemoo-re does this pr solve solve my [comment](https://github.com/withastro/astro/issues/9791#issuecomment-1906722419)? If not I will open a new issue.

natemoo-re:
> @natemoo-re does this pr solve solve my [comment](https://github.com/withastro/astro/issues/9791#issuecomment-1906722419)? If not I will open a new issue.

@jacobdalamb thanks for double checking! Yes, this PR does also fix the behavior with the `-y` or `-n` flags.

PR #9815, opened 2024-01-24
[docs] reorder items for configuration reference page

sarah11918:
## Changes

This reorders the `@docs` items for building the Docs Configuration Reference Page. There is **no content change**, only organizational changes. This should only affect how the docs page is generated (the order of items).

(File diffs may imply otherwise, because of similar text repeated throughout. But there have been no changes to any text. Only moving items around and attempting to get the indentation correct.)

- It reorganizes similar items together in an order that more reflects what people are asking about and looking for.

- It moves `integrations` and `vite` into the top-level category instead of making them standalone headers at the bottom of the page, mixed in with config items that *do* have several sub-options, taking care to:
	- [x]  remove the `@heading` property
	- [x] change the name to lower case

## Testing

No tests, only docs. Not sure why tests are failing.

## Docs

I hope I got all the indentations correct

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -409,11 +409,208 @@ export interface ImageServiceConfig<T extends Record<string, any> = Record<strin
  * Docs: https://docs.astro.build/reference/configuration-reference/
  */
 export interface AstroUserConfig {
+
 	/**
 	 * @docs
 	 * @kind heading
 	 * @name Top-Level Options
 	 */
+
+	/**
+	 * @docs
+	 * @name site
+	 * @type {string}
+	 * @description
+	 * Your final, deployed URL. Astro uses this full URL to generate your sitemap and canonical URLs in your final build. It is strongly recommended that you set this configuration to get the most out of Astro.
+	 *
+	 * ```js
+	 * {
+	 *   site: 'https://www.my-site.dev'
+	 * }
+	 * ```
+	 */
+	site?: string;
+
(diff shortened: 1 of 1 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -675,48 +745,36 @@ export interface AstroUserConfig {
...
-	 */
-	adapter?: AstroIntegration;
-
-	/**
-	 * @docs
-	 * @name output
-	 * @type {('static' | 'server' | 'hybrid')}
-	 * @default `'static'`
-	 * @see adapter
+	 * @name vite
```
TheOtterlord, line 748:
```suggestion
	 * @name vite
	 * @typeraw {ViteUserConfig}
```

packages/astro/src/@types/astro.ts
```diff
@@ -556,104 +546,184 @@ export interface AstroUserConfig {
...
 	 * ```
 	 */
-	compressHTML?: boolean;
+	adapter?: AstroIntegration;
 
 	/**
 	 * @docs
-	 * @name base
-	 * @type {string}
+	 * @name integrations
```
TheOtterlord, line 595:
```suggestion
	 * @name integrations
	 * @typeraw {AstroIntegration[]}
```

TheOtterlord (approved):
Adding types for `integrations` & `vite`, but looks good otherwise

natemoo-re (approved):
LGTM!

Issue #9695, opened 2024-01-14
& encoding with server rendered strings affects head URLs with multiple params


alexnguyennz:
### Astro Info

```block
Astro                    v4.1.2
Node                     v20.9.0
System                   Windows (x64)
Package Manager          unknown
Output                   static
Adapter                  none
```

### Describe the Bug

Seems intentional but when strings with `&` are used in a meta head tag, they're encoded as `&#38;` during development and build which affects URLs with a query string with multiple search params e.g. an `og:image` link pointing to an API that dynamically generates the image with a title and description.

```html
<meta  
  name="og:image"  
  content={'https://example.com/api/og?title=hello&description=somedescription'}  
/>

<!-- <meta name="og:image" content="https://example.com/api/og?title=hello&#38;description=somedescription"> --> 
<!-- title=hello -->
```

For the Stackblitz example, when built, it is transformed into `&#38;` .

### What's the expected result?

Not escaped in this case?

```html
<meta  
  name="og:image"  
  content={'https://example.com/api/og?title=hello&description=somedescription'}  
/>

<!-- <meta name="og:image" content="https://example.com/api/og?title=hello&description=somedescription"> --> 
<!-- title=hello description=somedescription -->
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-2f3s26?file=src%2Flayouts%2FLayout.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

florian-lefebvre:
I can confirm this, I had the issue in a project of mine and added `// TODO: fix in Astro` (but never even reported 😂)

alexnguyennz:
Thanks Florian, should I make a PR?

alexnguyennz:
I've made it here: https://github.com/withastro/astro/pull/9820

PR #1789, opened 2021-11-10
Markdown bug bash!

natemoo-re:
## Changes

- @jonathantneal and I uncovered a handful of bugs while migrating the Docs site for #1738.
- This fixes them.

## Testing

Manually. Before it did not work, now it does work.

## Docs

Bug fixes

PR #9820, opened 2024-01-25
prevent URLs in attributes being escaped

alexnguyennz:
## Changes

[Issue](https://github.com/withastro/astro/issues/9695)

Adds check for server rendered URLs attributes with any `&` so they're not escaped. For edge cases where multiple params are needed e.g. for an API that generates OG images.

Before:
`<meta name="og:image" content="https://example.com/api/og?title=hello&#38;description=somedescription">`

After:
`<meta name="og:image" content="https://example.com/api/og?title=hello&description=somedescription">`

I think this change is safe - let me know if there's another way this should be done.

## Testing

`pnpm run test`

And manually with `minimal` demo in dev and build with:

```
<head>
  <meta
    name="og:image"
    content={'https://example.com/api/og?title=hello&description=somedescription'}
  />
</head>
```

## Docs

N/A


.changeset/quick-islands-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Prevent URLs in `content` attributes being escaped for multiple params to work
```
florian-lefebvre, line 5:
```suggestion
Prevents URLs in attributes from being escaped in static builds
```
I think it concerns any attribute, eg. hrefs

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -104,6 +104,11 @@ Make sure to use the static attribute syntax (\`${key}={value}\`) instead of the
 		return markHTMLString(` class="${toAttributeString(value, shouldEscape)}"`);
 	}
 
+	// prevent URLs in `content` attributes being escaped for multiple params to work
+	if (key === 'content' && typeof value === 'string' && URL.canParse(value)) {
```
florian-lefebvre, line 108: Not familiar with this part of the codebase, but can  you for hrefs attributes as well?

bluwy:
Do we need to escape the `&` in the first place? It should be fine to be used in attribute values 🤔  Seems like it's added since https://github.com/withastro/astro/pull/1789. Perhaps @natemoo-re has some context about it.

.changeset/quick-islands-ring.md
alexnguyennz: Yeah I think you're right. I've just set it as a general URL check for any attribute for now.

natemoo-re:
> Do we need to escape the `&` in the first place? It should be fine to be used in attribute values 🤔 Seems like it's added since #1789. Perhaps @natemoo-re has some context about it.

Good question! Escaping `&` is best practice. Seems like [Preact](https://github.com/preactjs/preact-render-to-string/blob/22cb8461a70d86e0e3e4bf36320812a50591814a/src/util.js#L8) and [Svelte](https://github.com/sveltejs/svelte/blob/d57eff76ed24ae2330f11f3d3938761ae4e14b4b/packages/svelte/src/internal/server/index.js#L27) escape it by default.

I'm not sure that we should hard-code the `content` attribute... Maybe we could check for the existence of `&` and then check for a URL? I'm not sure of a simple way we can determine if the & should be escaped or if it's potentially needed for a URL.

alexnguyennz:
> I'm not sure that we should hard-code the `content` attribute... Maybe we could check for the existence of `&` and then check for a URL? I'm not sure of a simple way we can determine if the & should be escaped or if it's potentially needed for a URL.

Not sure either - it's now just a general URL check (with any &) for all attributes if that works.

bluwy:
> Good question! Escaping `&` is best practice. Seems like [Preact](https://github.com/preactjs/preact-render-to-string/blob/22cb8461a70d86e0e3e4bf36320812a50591814a/src/util.js#L8) and [Svelte](https://github.com/sveltejs/svelte/blob/d57eff76ed24ae2330f11f3d3938761ae4e14b4b/packages/svelte/src/internal/server/index.js#L27) escape it by default.

Thanks! It seems like they encode `&` as `&amp;` and `"` as `&quot;`, while we encode as `&#38;` and `&#34;` respectively. Perhaps we need to change how we encode this as well? Technically both should be equivalent, but maybe browsers are able to parse `&amp;` better because the `#` is not tripping it up (?).

ematipico:
I also think that using HTML entities is better.

ematipico:
This PR is sitting for a while without too much activity. cc @alexnguyennz  and @natemoo-re

natemoo-re:
Thanks for the ping @ematipico! I'm okay with this solution if everyone else is.

We can follow-up with switching to `&amp;` and `&quot;` in another PR.

.changeset/quick-islands-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Prevents URLs in attributes from being escaped in static builds
```
natemoo-re, line 5:
```suggestion
Prevents fully formed URLs in attributes from being escaped
```

natemoo-re (approved):

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -104,6 +104,11 @@ Make sure to use the static attribute syntax (\`${key}={value}\`) instead of the
 		return markHTMLString(` class="${toAttributeString(value, shouldEscape)}"`);
 	}
 
+	// Prevents URLs in attributes from being escaped in static builds
+	if (typeof value === 'string' && value.includes('&') && URL.canParse(value)) {
```
natemoo-re, line 108: I noticed that [`URL.canParse`](https://nodejs.org/api/url.html#urlcanparseinput-base) was added in Node v18.17.0, but our minimum required version is still v18.14.1.

We should either implement a fallback for `URL.canParse` (`try/catch`) or wait on merging this until we bump our engines to v.18.17.0 (possibly in the next minor cc @Princesseuh)

natemoo-re (changes requested):
Just blocking until we resolve the `URL.canParse` discrepancy.

Nothing for you to do @alexnguyennz, the maintainers just need to decide when we're updating our Node version.

packages/astro/src/runtime/server/render/util.ts
lilnasy: [`URL.canParse`](https://url.spec.whatwg.org/#dom-url-canparse) uses the same internal method as the [`URL constructor`](https://url.spec.whatwg.org/#constructors), so a try-catch should be sufficient to unblock.

lilnasy (approved):
Looks great to me! It must have taken effort to find the relevant code path, thanks for the clean fix!

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -224,3 +229,12 @@ export function promiseWithResolvers<T = any>(): PromiseWithResolvers<T> {
 		reject,
 	};
 }
+
+function urlCanParse(url: string) {
+	try {
+		new URL(url);
```
Fryuni, line 235: Are relative URLs expected to be unescaped as well? If the OG image in the example were only `/api/og?title=hello&description=somedescription` to be relative to the current browser origin, it wouldn't be escaped as that is not a valid URL.

If we want to validate relative URLs as well this should be:
```suggestion
		new URL(url, 'http://example.com');
```

Any base there would work.

lilnasy: URL.canParse wouldn't have allowed it but this seems like a great enhancement to me, thanks!

lilnasy: On second thought, it may be too permissive.
I don't think there's a case anymore where urlCanParse will fail. If the attribute value has a "&", it always prints unescaped.

Fryuni: `URL.canParse` can also receive a base as a second argument, so this applies after upgrading the Node version as well

lilnasy: You're right, and we might want to be careful with it as well. I'm not sure how valid it is, but the threat is that we let an attribute like "&\\"" through and break everything.

Fryuni: > On second thought, it may be too permissive. I don't think there's a case anymore where urlCanParse will fail. If the attribute value has a "&", it always prints unescaped.

Indeed, it is very hard for a string not to be a valid URL when there is a base. It was hard to find a test case that didn't pass it when we had a similar use case.

Makes this check do almost nothing.

lilnasy: You're right. Undoing the change.

natemoo-re (approved):
Thanks for getting this over the line, @lilnasy!

PR #9822, opened 2024-01-25
fix: fix `transition:name` can be unicode

### LINKED ISSUE #9803: `transition:name` can't use Chinese
### Astro Info

```block
Astro                    v4.1.0
Package Manager          npm
Adapter                  none
                         @astrojs/react
                         @astrojs/sitemap
                         astro-expressive-code
                         @astrojs/mdx
```

Chrome

### Describe the Bug

Set a element `transition:name` to `transition:name={name}`
The name is Chinese text. 

Read Astro code:

```ts
// Ensure animationName is a valid CSS identifier
function toValidIdent(name: string): string {
	return name.replace(/[^a-zA-Z0-9\-\_]/g, '_').replace(/^\_+|\_+$/g, '');
}
```

This will cause the Chinese name to be replaced with empty, resulting in multiple duplicates of the name.

image

## My website as example:
https://liruifengv.com/tags/

image

I think the same will be true for Japanese and Korean, etc.

Expect Chinese to be used correctly as transition:name.

CSS identifiers can be Chinese、Japanese and Korean, etc.

https://liruifengv.com/tags/

### Participation

liruifengv:
## Changes

- Close #9803
- escape `transition:name` use `cssesc` lib

## Testing

Added e2e test.
## Docs

Not need update.


packages/astro/src/runtime/server/transition.ts
```diff
@@ -39,15 +39,16 @@ function toValidIdent(name: string): string {
 		null,
 		`Your transition:name ${bold(name)} is not a valid CSS identifier. It will be escaped.`
 	);
-	const result = [];
-	for (const char of name) {
-		if (/[a-zA-Z0-9_-]/.test(char)) {
-			result.push(char);
+	let result = "";
+	for (let i = 0; i < name.length; i++)  {
```
martrapp, line 43: That last change is a bad idea, please revert to const char of name

.changeset/cold-bobcats-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+fix(view-transition): escaped unicode `transition:name`
```
martrapp, line 5:
```suggestion
Applies the correct escaping to identifiers used with `transition:name`.

**Comment by martrapp** [comment]
Great PR! Don't let my many comments put you off, I really appreciate your contribution!

**Comment by yisibl** [comment]
Also, need to consider emoji, additionally is there a test case to test them?

**Comment by martrapp** [comment]
> Also, need to consider emoji, additionally is there a test case to test them?

An idea for a e2e test case would be to read back the `view-transition-name` style property, e.g. test `el.style.viewTransitionName`. If the `view-transition-name` in the generated style sheet contains illegal chars, the `viewTransitionName` of the elements styles will be completely empty. In other cases it will yield the string as shown on the right side of the screenshots above, i.e. human readable. Escaped ascii characters would would be shown with there escaping slash but hex encoded code points will be shown as the character they represent.

File: packages/astro/src/runtime/server/transition.ts [hunk 3]
```diff
@@ -25,7 +28,26 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+	logger.warn(
+		null,
+		`Your transition:name ${bold(name)} is not a valid CSS identifier. It will be escaped.`
+	);
+	const result = [];
+	for (const char of name) {
+		if (/[a-zA-Z0-9_-]/.test(char)) {
+			result.push(char);
+		} else {
+			result.push(`\\${char.charCodeAt(0).toString(16)}-`);
```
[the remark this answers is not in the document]
martrapp: Sorry, i was wrong for the last example. As space isn't a valid character here, it has also to be escaped and the result of the last example must be `"\\5f00\\ \\6e90"`

packages/astro/src/runtime/server/transition.ts
```diff
@@ -32,6 +32,7 @@ function toValidIdent(name: string): string {
 	if (/^[0-9]|^-[0-9]|^--/.test(name)) {
     throw new Error("Your transition:name is not a valid CSS identifier.");
 	}
+	// can't be a CSS keyword
 	if (['unset', 'initial', 'inherit', 'none' ].includes(name)) {
```
martrapp, line 33: Top! As mdn says, you can't define those. And it would definitively be a mistake to use them with `transition:name` in combination with `transition-animate` to define a new transition animation.

But you could use them as `view-transition-names` as they have a predefined meaning. At least `none` might come in handy if you want to exclude an element from the default animation.

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +28,26 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+	});
+	logger.warn(
+		null,
+		`Your transition:name ${bold(name)} is not a valid CSS identifier. It will be escaped.`
+	);
+	const result = [];
+	for (const char of name) {
+		if (/[a-zA-Z0-9_-]/.test(char)) {
+			result.push(char);
+		} else {
```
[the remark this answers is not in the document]
martrapp: will take me a bit ...

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
 
 // Ensure animationName is a valid CSS identifier
 function toValidIdent(name: string): string {
-	return name.replace(/[^a-zA-Z0-9\-\_]/g, '_').replace(/^\_+|\_+$/g, '');
+	// can't start with a number, - + number, or --
```
martrapp, line 28:
```suggestion
	// can't start with a number, - + number
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+	if (/^[0-9]|^-[0-9]/.test(name)) {
+    throw new Error(`Your transition:name ${name} is not a valid CSS identifier.`);
```
martrapp, line 30:
```suggestion
    throw new Error(`Your transition:name ${name} is not a valid CSS identifier as it starts with ${name.substring(0,2)}`);
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+	}
+	if (/^[a-zA-Z0-9_\\-]*$/.test(name)) {
+		return name;
+	}
+	let result = "";
```
martrapp, line 35:
```suggestion
	let result = "";
	let wasHexEscaped = false;
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+	for (const char of name)  {
+		if (/[a-zA-Z0-9_\\-]/.test(char)) {
```
martrapp, line 37:
```suggestion
		if (/[0-9a-fA-F]/.test(char) && wasHexEscaped) {
			result += " ";
		}
		wasHexEscaped = false;
		if (/[a-zA-Z0-9_\\-]/.test(char)) {
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+			result += char;
```
martrapp, line 38:
```suggestion
			result += char;
			continue;
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+		} else {
+			// TODO handle $ and other special characters
+			const suffix = char === name[name.length - 1] ? "" : "\ ";
+			result += `\\${char.codePointAt(0)!.toString(16)}${suffix}`;
```
martrapp, line 42:
```suggestion
		const code = char.codePointAt(0)
		if (code === undefined) {
			throw new Error(`Your transition:name ${name} is not a valid CSS identifier as it contains undefined code points`);
		}
		if (code < 0x20 || code >= 0x7f) {
			result += `\\${code.toString(16)}`;
			wasHexEscaped = true;
		} else {
			result += `\\${char}`;
		}
```

packages/astro/src/runtime/server/transition.ts
```diff
@@ -25,7 +25,24 @@ export function createTransitionScope(result: SSRResult, hash: string) {
...
+		}
```
martrapp, line 43:
```suggestion
```

natemoo-re:
I did find a potential package to handle this! [`cssesc`](https://github.com/mathiasbynens/cssesc) was written by [Mathias Bynens](https://github.com/mathiasbynens), who is very involved in standards work. 

Although the package is quite old, I'm confident that the code is stable!

I'd suggest that we try to use a package for this, and fallback to implementing our own logic only if there isn't a package out there that meets our needs.

martrapp:
> I did find a potential package to handle this!

Excellent catch! As you can see from the linked issue, we were looking for a lib like this but failed.
```
cssesc(name, {'isIdentifier': true});
```
seems to be exactly what we need here.

liruifengv:
> I did find a potential package to handle this! [`cssesc`](https://github.com/mathiasbynens/cssesc) was written by [Mathias Bynens](https://github.com/mathiasbynens), who is very involved in standards work. He also [wrote an article](https://mathiasbynens.be/notes/css-escapes) and built [a hosted tool](https://mothereff.in/css-escapes) to explore this topic.
> 
> Although the package is quite old, I'm confident that the code is stable!
> 
> I'd suggest that we try to use a package for this, and fallback to implementing our own logic only if there isn't a package out there that meets our needs.

Thank you! I used the package. It works!

martrapp:
Wow over 20 conversation and at the end a one line change (+ tests and dependencies of course)
That was a fun trip, I enjoyed it!

martrapp (approved):
Looks very good to me now!

PR #9825, opened 2024-01-25
fix(audit): mark widget roles as interactive

tugrulates:
## Changes

- passes valid element/role combinations like button/switch ([example](https://headlessui.com/react/switch))
- adds all subclasses of [`widget`](https://www.w3.org/TR/wai-aria/#widget) to the interactive roles (by exclusion from non-interactive roles in code), including composite widgets
- I used the list from https://www.w3.org/TR/wai-aria/#widget_roles
- a programmatic subclass query, using [aria-query](https://www.npmjs.com/package/aria-query) is possible but omitted in this PR (please let me know if this would be preferred)
- looks like the [`a11y_non_interactive_element_to_interactive_role_exceptions`](https://github.com/withastro/astro/blob/57ab578bc7bdac6c65c2315365c0e94bc98af2b3/packages/astro/src/runtime/client/dev-toolbar/apps/audit/a11y.ts#L617) list was setup to handle composite widgets, but it was defunct before
- this reverts behavior for all of these roles, i.e., now the audit will fail if they are on non-interactive elements

## Testing

Manually on minimal example:

```html
<button role="switch" aria-checked="false">Switch</button>
```

Audit fails before, passes after.

Also tested on `blog`, `framework-react` and `with-mdx` examples to make sure new false positives are not introduced.

## Docs

No mention of this audit in the docs (as far as I could find).


Princesseuh (approved):
This looks great to me!

ematipico (approved):
I wonder we could add a new test to avoid possible future regressions

natemoo-re (approved):
Thanks for the PR!

PR #9828, opened 2024-01-25
fix(core): don't noop shared modules

ematipico:
## Changes

Closes PLT-1445
Closes https://github.com/withastro/astro/issues/9688

To fix the issue, I check if the modules imported by a pre-rendered page are imported somewhere else in the user's project. If so, I mark that page with a new flag.

## Testing

Added a new test

## Docs
N/A


packages/astro/src/@types/astro.ts
```diff
@@ -2553,6 +2553,7 @@ export interface RouteData {
 	segments: RoutePart[][];
 	type: RouteType;
 	prerender: boolean;
+	hasSharedModules: boolean;
```
matthewp, line 2556: I don't think I'd use routeData for this, given that it's not a property of the route itself, it's more like internal state for the build. Normally we do this sort of thing either using the build internals or the PageData (I'd probably go with PagaData here).

packages/astro/src/core/build/plugins/plugin-prerender.ts
```diff
@@ -18,10 +18,31 @@ function vitePluginPrerender(opts: StaticBuildOptions, internals: BuildInternals
...
+							// Here, we check if this page is importing modules that are shared among other modules e.g. middleware, other SSR pages, etc.
+							// we loop the modules that the current page imports
+							for (const moduleId of infoMeta.importedIds) {
+								// we retrieve the metadata of the module
+								const moduleMeta = meta.getModuleInfo(moduleId)!;
+								if (
+									// a shared modules should be inside the `src/` folder, at least
+									moduleMeta.id.startsWith(opts.settings.config.srcDir.pathname) &&
+									// and have at least two importers: the current page and something else
+									moduleMeta.importers.length > 1
```
natemoo-re, line 37: This check might be insufficient? It's close, but we technically want to trace all of these imported modules back to each entrypoint that references it. If all of the entrypoints (pages, endpoints) are prerendered, we don't need to move the module to a separate chunk.

natemoo-re (commented):
This solves the issue but it might introduce another one—if an imported module is shared by multiple routes, but all of those routes are prerendered, it should still be replaced in the final server build. 

Currently this will keep any imports that are referenced by another file no matter what.

packages/astro/src/core/build/plugins/plugin-prerender.ts
ematipico: >  If all of the entrypoints (pages, endpoints) are pre-rendered, 

Ironically, my first implementation had this check 😆

natemoo-re (approved):
The new approach seems reasonable to me!

PR #9830, opened 2024-01-25
fix(dev): dont reroute endpoint responses

### LINKED ISSUE #9829: Custom error pages are not served in production
### Astro Info

```block
Astro                    v4.2.4
Integrations             astro-i18next
                         @astrojs/react
```

_No response_

### Describe the Bug

Given an endpoint returning `return new Response(null, { status: 404 })`,
I've noticed different behaviours when serving the site with `astro dev` or `astro preview`.

astro dev: custom 404 page is shown
astro preview: blank page is shown

It also works as expected, when a 404 Response is returned in the code-fence of Astro page, but not in an endpoint.

If in production the endpoint returns a response with 404 status and the custom 404 page is present, it should show up

https://stackblitz.com/edit/github-s7uxf1?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes
- fix #9829 

## Testing
Added a case to dev-routing.test.js

## Docs
Does not affect behavior

### DIFF
--- a/packages/astro/src/runtime/server/consts.ts
+++ b/packages/astro/src/runtime/server/consts.ts
@@ -1,0 +1,1 @@
+export const REROUTE_DIRECTIVE_HEADER = 'X-Astro-Reroute';
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -35,6 +35,7 @@ import { preload } from './index.js';
 import { getComponentMetadata } from './metadata.js';
 import { handle404Response, writeSSRResult, writeWebResponse } from './response.js';
 import { getScriptsForURL } from './scripts.js';
+import { REROUTE_DIRECTIVE_HEADER } from '../runtime/server/consts.js';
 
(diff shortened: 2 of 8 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -342,7 +343,7 @@ export async function handleRoute({
 			})
 		);
 	}
-	if (response.status === 404 && has404Route(manifestData)) {
+	if (response.status === 404 && has404Route(manifestData) && response.headers.get(REROUTE_DIRECTIVE_HEADER) !== "no") {
```
lilnasy, line 346: The fix is here, the rest is clerical.

packages/astro/src/runtime/server/consts.ts
```diff
@@ -0,0 +1 @@
+export const REROUTE_DIRECTIVE_HEADER = 'X-Astro-Reroute';
```
natemoo-re, line 1: I like this!

natemoo-re (approved):
LGTM! Thanks for the fix.

ematipico (approved):

Issue #9824, opened 2024-01-25
Dev toolbar App with absolute file path does not work on Windows


florian-lefebvre:
### Astro Info

```block
Astro                    v4.2.4
Node                     v18.19.0
System                   Windows (x64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         @sentry/astro
                         @spotlightjs/astro
```

### Describe the Bug

When provided an absolute file path to a dev toolbar app (for example using the Spotlight integration) on Windows, it fails to load:
```
10:32:50 [ERROR] [toolbar] Failed to load dev toolbar app from D:\DEV\PROJECTS\florian-lefebvre\astro-spotlight-repro\node_modules\@spotlightjs\astro\src\overlay\index.ts: Failed to fetch dynamically imported module: file:///D:/DEV/PROJECTS/florian-lefebvre/astro-spotlight-repro/node_modules/@spotlightjs/astro/src/overlay/index.ts
```

### What's the expected result?

It should load properly

### Link to Minimal Reproducible Example

https://github.com/florian-lefebvre/astro-spotlight-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Ah, I didn't mean to remove anything in that PR, I'll send a PR. Sorry for the inconvenience!

KyTiXo:
Welp.. wasted a whole hour trying to figure out why my toolbar wouldn't import in a fresh project haha

PR #9834, opened 2024-01-26
fix(toolbar): Fix loading third-party apps using absolute paths

Princesseuh:
## Changes

Bring back the fix for absolute paths for dev toolbar apps that was removed by accident in my error handling PR.

Fix https://github.com/withastro/astro/issues/9824

## Testing

Added a test, I think it cover the use case? Not sure with the monorepo

## Docs

N/A


florian-lefebvre:
Looks good! Can we make a preview release to test it out? I don't remember the command

packages/astro/src/vite-plugin-dev-toolbar/vite-plugin-dev-toolbar.ts
```diff
@@ -65,13 +65,16 @@ export default function astroDevToolbar({ settings, logger }: AstroPluginOptions
 				return `
 					export const loadDevToolbarApps = async () => {
 						return (await Promise.all([${settings.devToolbarApps
-							.map((plugin) => `safeLoadPlugin(${JSON.stringify(plugin)})`)
+							.map(
+								(plugin) =>
+									`safeLoadPlugin(async () => (await import(${JSON.stringify(plugin)})).default)`
```
bluwy, line 70: I think we also need to pass second parameter of `JSON.stringify(plugin)`, since on line 85 the `safeLoadPlugin` function references the `entrypoint` variable string

Princesseuh: This is why programming in strings was a mistake

bluwy: We can also do this:

```js
async function something() {
  // ...
}

// ...
return `
  ${something.toString()}
`
```

But either way works for me 😄

ematipico:
!preview toolbar-absolute-paths

bluwy (approved):

ematipico (approved):

.changeset/real-lamps-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix third-party apps not loading correctly when using absolute paths on Windows
```
natemoo-re, line 5:
```suggestion
Fixes third-party dev toolbar apps not loading correctly when using absolute paths on Windows
```

natemoo-re (approved):

PR #9835, opened 2024-01-26
Fix release announcement script (redemption arc)

natemoo-re:
## Changes

- Turns out that we should probably return the message instead of just logging it to the console 😵‍💫

## Testing

We'll do it live

## Docs

Nooooooooooooo


ematipico (approved):

delucis (approved):
This is beautiful :D 💖

PR #9838, opened 2024-01-26
fix(i18n): use `define` to deliver config to virtual module

lilnasy:
## Changes
- Switches the approach used to deliver i18n config to the `astro:i18n` module.
- The virtual module approach broke in mysterious ways.

## Testing
Existing tests should pass.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/i18n/vite-plugin-i18n.ts
+++ b/packages/astro/src/i18n/vite-plugin-i18n.ts
@@ -31,7 +31,7 @@ export default function astroInternationalization({
 		config(config) {
 			const i18nConfig: I18nInternalConfig = { base, format, site, trailingSlash, i18n };
 			config.define ??= {}
-			config.define["import.meta.env._ASTRO_INTERNAL_I18N_CONFIG"] = JSON.stringify(i18nConfig);
+			config.define.__ASTRO_INTERNAL_I18N_CONFIG__ = JSON.stringify(i18nConfig);
 		},
 		resolveId(id) {
 			if (id === virtualModuleId) {
--- a/packages/astro/src/i18n/vite-plugin-i18n.ts
+++ b/packages/astro/src/i18n/vite-plugin-i18n.ts
@@ -30,8 +30,11 @@ export default function astroInternationalization({
 		enforce: 'pre',
 		config(config) {
 			const i18nConfig: I18nInternalConfig = { base, format, site, trailingSlash, i18n };
-			config.define ??= {}
-			config.define.__ASTRO_INTERNAL_I18N_CONFIG__ = JSON.stringify(i18nConfig);
+			return {
+				define: {
+					__ASTRO_INTERNAL_I18N_CONFIG__: JSON.stringify(i18nConfig)
+				}
+			}
 		},
 		resolveId(id) {
 			if (id === virtualModuleId) {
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Not my favorite solution, but if it works, it works!

ematipico (commented):
Can we add a test? As far as I understood, there was a way to reproduce it

lilnasy:
The error happens on dev startup, during some work being performed by a plugin called vite:dep-scan. It would be simpler if it happened on request. I'm not confident a test can catch but I'll take a look.

lilnasy:
No luck.

The fact that a fetch request itself will succeed with expected html aside, the bug won't even happen in our monorepo. It seems to be one of these cases where vite is sensitive to whether astro is linked or copied to node_modules.

ematipico:
Thank you for taking the time to explain and double check.

ematipico (approved):

bluwy:
Can you explain how this fixes the issue? Using `import.meta.env` this way may bloat the size if the user references `import.meta.env` itself in runtime.

lilnasy:
AFAIK, it will not appear on import.meta.env. `define` only does a string replace, doesn't it?

lilnasy:
I'm guessing astro-env-inject picks it up.

packages/astro/src/i18n/vite-plugin-i18n.ts
```diff
@@ -29,28 +28,16 @@ export default function astroInternationalization({
 	return {
 		name: 'astro:i18n',
 		enforce: 'pre',
-		async resolveId(id) {
+		config(config) {
+			const i18nConfig: I18nInternalConfig = { base, format, site, trailingSlash, i18n };
+			config.define ??= {}
+			config.define.__ASTRO_INTERNAL_I18N_CONFIG__ = JSON.stringify(i18nConfig);
```
bluwy, line 34: Tiny nit: I'd do something like:

```js
return {
  define: {
    __ASTRO_INTERNAL_I18N_CONFIG__: ...
  }
}
```

so it's a little cleaner. Though they'd still work the same.

bluwy:
> AFAIK, it will not appear on import.meta.env. `define` only does a string replace, doesn't it?

Vite would use `import.meta.env.*` in `define`: https://github.com/vitejs/vite/blob/f9c33add88e8edcd54316629724730d8049447f1/packages/vite/src/node/plugins/define.ts#L50-L53. The entire object may get inlined for `import.meta.env` as it represents an object of all the keys and values.

packages/astro/src/i18n/vite-plugin-i18n.ts
lilnasy: That looks cleaner to me as well!

Could you make a suggestion so it can be committed directly?

lilnasy:
I didn't know that, thanks!

packages/astro/src/i18n/vite-plugin-i18n.ts
bluwy:
```suggestion
			return {
				define: {
					__ASTRO_INTERNAL_I18N_CONFIG__: JSON.stringify(i18nConfig)
				}
			}
```

PR #9841, opened 2024-01-27
fix: use publicDir & base for 'include your base in url' error message

### LINKED ISSUE #9759: Error message about missing baseUrl should reflect publicDir config
### Astro Info

```block
Package Manager          pnpm
Adapter                  none
which: no xclip (didn't even know I don't have that installed)
```

unrelated

### Describe the Bug

After creating a repo using the framework-solid template and changing the baseUrl and publicDir configs according to the Gitlab Deployment guide I got the following error message:

```
[ERROR] [router] Request URLs for public/ assets must also include your base. "/favicon.svg" expected, but received "/favicon.svg".
```

1. This error message should probably reference the publicDir config instead of having `public/` hardcoded
2. The paths in the second half of the error message are the same. It appears the first should read something like `/{baseUrl}/favicon.svg`

The error message should read:

```
[ERROR] [router] Request URLs for static/ assets must also include your base. "/my-page/favicon.svg" expected, but received "/favicon.svg".
```

for the following astro config:

```
import solid from "@astrojs/solid-js";
import { defineConfig } from "astro/config";

// https://astro.build/config
export default defineConfig({
  site: "https://gitlab-user.gitlab.io",
  base: "/my-page",
  integrations: [solid()],
  outDir: "public",
  publicDir: "static"
});
```

https://stackblitz.com/edit/github-xhj8bw-sw2zrn?file=astro.config.mjs

### Participation

kristianbinau:
## Changes

Closes #9759 

Fixes error message when requesting a public url without your base.
Now the error message will correctly show the correct publicUrl  and will add your base to the expected.

### Before / After
_With the following config:_
```ts
export default defineConfig({
  base: "/my-page",
  publicDir: "static"
});
```

_Before:_
```
[ERROR] [router] Request URLs for public/ assets must also include your base. "/favicon.svg" expected, but received "/favicon.svg".
```

_After:_
```
[ERROR] [router] Request URLs for static/ assets must also include your base. "/my-page/favicon.svg" expected, but received "/favicon.svg".
```

## Testing

Didn't feel like this small "text" change needed it's own test : )

## Docs

Doesn't change current behavior only makes the error message more clear.

### DIFF
--- a/.changeset/eight-windows-kiss.md
+++ b/.changeset/eight-windows-kiss.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes error message when requesting a public url without your base.
+Makes the warning clearer when having a custom `base` and requesting a public URL without it
--- a/.changeset/eight-windows-kiss.md
+++ b/.changeset/eight-windows-kiss.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes error message when requesting a public url without your base.
--- a/packages/astro/src/vite-plugin-astro-server/base.ts
+++ b/packages/astro/src/vite-plugin-astro-server/base.ts
@@ -3,11 +3,11 @@ import type { AstroSettings } from '../@types/astro.js';
 
 import { bold } from 'kleur/colors';
 import * as fs from 'node:fs';
+import path from 'node:path';
+import { appendForwardSlash } from '@astrojs/internal-helpers/path';
 import type { Logger } from '../core/logger/core.js';
 import notFoundTemplate, { subpathNotUsedTemplate } from '../template/4xx.js';
 import { writeHtmlResponse } from './response.js';
-import { appendForwardSlash, removeTrailingForwardSlash } from '@astrojs/internal-helpers/path';
-import path from 'node:path';
 
 export function baseMiddleware(
 	settings: AstroSettings,
--- a/packages/astro/src/vite-plugin-astro-server/base.ts
+++ b/packages/astro/src/vite-plugin-astro-server/base.ts
@@ -55,7 +55,7 @@ export function baseMiddleware(
 		fs.stat(publicPath, (_err, stats) => {
 			if (stats) {
 				const publicDir = appendForwardSlash(
-					path.relative(config.root.pathname, config.publicDir.pathname)
+					path.posix.relative(config.root.pathname, config.publicDir.pathname)
 				);
 				const expectedLocation = new URL(devRootURL.pathname + url, devRootURL).pathname;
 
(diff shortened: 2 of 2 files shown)


.changeset/eight-windows-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes error message when requesting a public url without your base.
```
florian-lefebvre, line 5:
```suggestion
Makes the warning clearer when having a custom `base` and requesting a public URL without it
```

packages/astro/src/vite-plugin-astro-server/base.ts
```diff
@@ -52,11 +52,13 @@ export function baseMiddleware(
 		const publicPath = new URL('.' + req.url, config.publicDir);
 		fs.stat(publicPath, (_err, stats) => {
 			if (stats) {
-				const expectedLocation = new URL('.' + url, devRootURL).pathname;
+				const publicDir = config.publicDir.pathname.replace(/\/$/, '').split('/').pop() + '/';
```
florian-lefebvre, line 55: I think there is a utility for slashes somewhere, where is it located @natemoo-re?

ematipico: Here's an example: https://github.com/withastro/astro/blob/main/packages%2Fastro%2Fsrc%2Fi18n%2Findex.ts#L1

kristianbinau: Like this? : ) https://github.com/withastro/astro/pull/9841/commits/a9afa2bd577f83af06eb717c5f4cd482aeedc628

florian-lefebvre (approved):

packages/astro/src/vite-plugin-astro-server/base.ts
```diff
@@ -52,11 +53,15 @@ export function baseMiddleware(
 		const publicPath = new URL('.' + req.url, config.publicDir);
 		fs.stat(publicPath, (_err, stats) => {
 			if (stats) {
-				const expectedLocation = new URL('.' + url, devRootURL).pathname;
+				const publicDir = appendForwardSlash(
+					removeTrailingForwardSlash(config.publicDir.pathname).split('/').pop()!
+				);
```
bluwy, line 58: If `publicDir` is `some/static/path`, wouldn't this only show the last part `path`? I think we could try to construct a relative path using `path.relative(fileUrlToPath(config.publicDir), fileUrlToPath(config.root))` perhaps so we can get the other directories. `URL` doesn't seem to make this easy.

kristianbinau: That seems like a great idea, I'll try tomorrow, I should have time then👍

kristianbinau: Done, Great idea! : )

bluwy (approved):
Thanks!

PR #9850, opened 2024-01-28
Draft: allow creating schemas for json files

godalming123:
Note: closes https://github.com/withastro/roadmap/discussions/806

## Changes

- [X] Allow the user to specify JSON files to be validated in the astro config
- [X] Type check these JSON files (**TODO**: when this fails the build/debug process continues)
- [-] Allow for images to be processed so that they can be used with the astro image components
- [ ] Add changeset `pnpm exec changeset`

## Testing

- [x] Run a manual test, since so far the code has been guess work
- [ ] Add automatic tests

## Docs

- [ ] Update docs


godalming123:
I'm ready to start receiving feedback on this PR (despite it not being finished). Here's a list of todos/potential improvements:
- Currently when JSON data is invalid, astro just continues building/debugging/running as if nothing happened (https://github.com/withastro/astro/pull/9850/files#diff-3aeabaefb8af152384488c68cc022bc10f3ff327683568298cd6865e795700bdR35)
- Allow the use of optimized images with a syntax like content collections
- Type check the zod schema that is parsed via `jsonDataFiles[n].schema` (https://github.com/withastro/astro/pull/9850/files#diff-cec1de65bb8dd2f8330cb37fe3698afd66162bc0f311aa11e0b3dd573f6a2b9fR409)
- Instead of requiring the user to import zod, use an alias like content collections?? (`import { z } from 'astro:content';`)
- Use a JS object for the ZOD schemas?? EG:
  ```js
  export default defineConfig({
    jsonDataFiles: {
      "src/content.json": z.object({
        websiteName: z.string(),
        websiteDescription: z.string(),
        websiteFrontImage: image(),
      })
    }
  });
  ```
  **NOTE:** I think that this would use https://github.com/colinhacks/zod?tab=readme-ov-file#catchall

I might add to this list as people report problems/improvements. Knowledge as to how these could be implemented is also appreciated.

examples/validate-json-with-zod/astro.config.mjs
```diff
@@ -0,0 +1,20 @@
...
+				name: z.string(),
+				images: z.array(
+					z.object({
+						imageSrc: z.string(), // TODO: use an optimized image type
+						imageAlt: z.string(),
+					})
+				),
+			}),
+		},
+	],
```
natemoo-re, line 19: Ah, this direction surprised me! 

We already have a pattern for type-safe/validated [data collections](https://docs.astro.build/en/guides/content-collections/#defining-a-collection-schema) using the `astro:content` module.

natemoo-re (commented):
Thanks for opening a PR, I really appreciate your enthusiasm to jump in and implement this!

Unfortunately, I think we already support this feature with a slightly different API! This functionality is covered by our [Content Collection](https://docs.astro.build/en/guides/content-collections/) feature when using the [`type: 'data'`](https://docs.astro.build/en/guides/content-collections/#defining-a-collection-schema) value.

Looking through [your proposal](https://github.com/withastro/roadmap/discussions/806), I don't think there was quite enough discussion about how this should be implemented. While singleton support would be nice, I don't think we should diverge from the established conventions for _where_ these files live.

examples/validate-json-with-zod/astro.config.mjs
godalming123: Don't you think that having a data collection with 1 JSON file for a name, description, and heroImage of a website would be confusing? The alternative is that I take the same code, but put the vite plugin code that I've added (in astro-postprocess) in my own astro project, which seems like something that astro would/should handle 🤔.  If that option does end up being taken (IE no singular JSON file validation with zod), then I will still need to know how images get processed, as I can't find anything that exports `z` within `/packages/astro/content/index.ts`, so don't know where to start looking when faced with this line of code:
```ts
import { z, defineCollection } from 'astro:content';
```

natemoo-re: I'm not really suggesting that you should use a full collection to handle a single file! I'm suggesting that if we want to implement this feature, the API should align with the existing ones. In this case, having configuration for validating a single file in the `astro.config.mjs` file but configuration for collection validation in the `src/content/config.ts` seems pretty confusing.

We can bikeshed on the exact naming, but it might be able to support a single argument version of `getEntry()`? So something like `getEntry("site-info")` would correspond to "src/content/site-info.json".

natemoo-re: I'd also say that we _probably_ don't need to modify the existing content collections logic that much. If we can just wire up the existing collection resolution logic, we won't need to implement a whole new Vite plugin for loading these files.

godalming123:
> It might be able to support a single argument version of `getEntry()`? So something like `getEntry("site-info")` would correspond to "src/content/site-info.json".

I definitely like this approach as it's much more explicit, but I'm curious about how you would define single JSON files to validate with zod:

> I'm not really suggesting that you should use a full collection to handle a single file!

Maybe something similar to the `jsonDataFiles` config, but in the `src/content/config.ts` would work?

As far as I see, these are the 2 details of a proposal that achieves this (defining zod schemas for JSON files, and using the JSON files that have been processed by the zod schemas).

florian-lefebvre:
Hey @godalming123 thank for submitting this PR. Unfortunately since this is a new feature, it needs to go through our RFC process (see https://github.com/withastro/roadmap). I'm closing this PR since it's only a Stage 1 proposal for now

Issue #9852, opened 2024-01-28
ASTRO_ASSET (string) field definition missing from type ImageFunction return value and interface ImageMetadata


aripalo:
### Astro Info

```block
Astro                    v4.2.1
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/netlify
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         astro-icon
```

### Describe the Bug

`ASTRO_ASSET: string` field definition missing from [`type ImageFunction` return value](https://github.com/withastro/astro/blob/70fdf1a5c660057152c1ca111dcc89ceda5c8840/packages/astro/content-types.template.d.ts#L23) and [interface ImageMetadata](https://github.com/withastro/astro/blob/70fdf1a5c660057152c1ca111dcc89ceda5c8840/packages/astro/src/assets/types.ts#L34).

**Having access to `ASTRO_ASSET` is useful: For example one could use it to create blurhash string (or other image placeholder values) _centrally_ within content collection schema definitions – without having to do it in various views/layouts.**

In comparison [`ImageMetadata["fsPath"]`](https://github.com/withastro/astro/blob/70fdf1a5c660057152c1ca111dcc89ceda5c8840/packages/astro/src/assets/types.ts#L40-L41) has been marked `@internal`, but `ASTRO_ASSET` is not marked internal, it's just missing from type definitions.

Of course I can workaround the issue by doing something like:
```ts
import {  type ImageFunction } from 'astro:content';

type CtxImage = z.infer<ReturnType<ImageFunction>>;

type CtxImageAsset = CtxImage & {
  ASTRO_ASSET: string;
};

function isCtxImageAsset(img: CtxImage): img is CtxImageAsset {
  return 'ASTRO_ASSET' in img;
}
```

... but I would feel much more comfortable with depending on that value if it were present in Astro's own type definitions as well (not just runtime JavaScript).

### What's the expected result?

I would like to see `ASTRO_ASSET` as part of `type ImageFunction` and/or `interface ImageMetadata`.

I'm willing to make a pull request, but I'd require some guidance if it's okay to add it and if yes, then for both or just to `ImageFunction` for example.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-avs1k5?file=src%2Fcontent%2Fconfig.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

aripalo:
[...]

PR #9853, opened 2024-01-28
feat: Expose ASTRO_ASSET field for return type of ImageFunction

aripalo:
## Changes

Solves [#9852](https://github.com/withastro/astro/issues/9852) by adding `ASTRO_ASSET` (string) field to type ImageFunction return value.

## Testing

- I ran `pnpm run test`
- Other tests success, but `pnpm run test -- -m "@astrojs/node#test"` fails with:
    ```ansi
    Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:test/reporters
    ```
    ... but this test fails already with `main` branch even without any changes. So that leads me to think that test regression has nothing to do with my changes.

## Docs

I don't think docs update is necessary as [Images in Content Collections](https://docs.astro.build/en/guides/images/#images-in-content-collections) documentation does not contain detailed documentation about the return values of `image()` function anyway.

### DIFF
--- a/.changeset/good-beers-deliver.md
+++ b/.changeset/good-beers-deliver.md
@@ -1,0 +1,5 @@
+---
+"astro": minor
+---
+
+Expose ASTRO_ASSET (string) field for return type of ImageFunction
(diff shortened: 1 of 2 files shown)


aripalo:
See https://github.com/withastro/astro/issues/9852#issuecomment-1913654748

PR #9857, opened 2024-01-29
Fixes false positives in the dev overlay audit when multiple `role` values exist.

### LINKED ISSUE #9856: When I use astro-diagram to render the mermaid sequence diagram, in developer mode, an error is reported！
### Astro Info

```block
> blog@0.0.1 dev /Users/bullub/个人中心/项目/blog
> astro dev

Debugger attached.

 astro  v4.2.6 ready in 1375 ms

┃ Local    http://localhost:4321/
┃ Network  use --host to expose

11:45:13 watching for file changes...
11:45:32 [200] /blog/hybrid/ios-wkwebview-ajax-lost/ 5438ms
11:45:32 [200] /blog/hybrid/ios-wkwebview-ajax-lost/ 3957ms
11:45:49 [ERROR] [toolbar] Failed to initialize dev toolbar app Audit (astro:audit):
TypeError: Cannot destructure property 'requiredProps' of 'roles.get(...)' as it is undefined.
    at Object.match (http://localhost:4321/node_modules/.pnpm/astro@4.2.6_less@4.2.0_typescript@5.3.3/node_modules/astro/dist/runtime/client/dev-toolbar/apps/audit/a11y.js?v=956fd8b0:454:15)
    at lint (http://localhost:4321/node_modules/.pnpm/astro@4.2.6_less@4.2.0_typescript@5.3.3/node_modules/astro/dist/runtime/client/dev-toolbar/apps/audit/index.js?v=956fd8b0:68:22)
    at async Object.init (http://localhost:4321/node_modules/.pnpm/astro@4.2.6_less@4.2.0_typescript@5.3.3/node_modules/astro/dist/runtime/client/dev-toolbar/apps/audit/index.js?v=956fd8b0:28:5)
    at async AstroDevToolbar.initApp (http://localhost:4321/node_modules/.pnpm/astro@4.2.6_less@4.2.0_typescript@5.3.3/node_modules/astro/dist/runtime/client/dev-toolbar/toolbar.js?v=956fd8b0:323:7)
```

Chrome

### Describe the Bug

I found through breakpoint debugging that the issue is due to the role attribute of the SVG node rendered by mermaid being set to `role="graphics document"`, resulting in `const {required Props}=roles. get (role)` in a11y.js; The code encountered an error during execution.

I understand that it may be necessary to parse the role attribute. If multiple roles are set, the value with index 0 will be prioritized.

Don't make an error.

https://stackblitz.com/edit/github-ankjbg?file=package.json

### Participation

iamyunsin:
When multiple values are set in the role, only the first value is returned.

fix #9856

## Changes

When multiple roles exist, check each role. Fixing the issue of errors caused by multiple values in the role.  #9856

## Testing

- No need to test, as this change will not affect the original behavior.

## Docs

- Only modified the way the role is valued, compatible with the situation of multiple role values.


Princesseuh:
Hmm, I think it would be better to fix the destructuring code here. I also think getRole should actually properly process multiple roles at once.

iamyunsin:
> Hmm, I think it would be better to fix the destructuring code here. I also think getRole should actually properly process multiple roles instead of choosing just the first one.

Okay, indeed I also think there should be restoration where deconstructed. But I may not be sure how to handle multiple roles correctly.

iamyunsin:
> Hmm, I think it would be better to fix the destructuring code here. I also think getRole should actually properly process multiple roles instead of choosing just the first one.

@Princesseuh I tried it again, please help me correct it. Thanks！

packages/astro/src/runtime/client/dev-toolbar/apps/audit/a11y.ts
```diff
@@ -496,13 +496,17 @@ export const a11y: AuditRuleWithSelector[] = [
...
 			}
-			const { requiredProps } = roles.get(role)!;
-			const required_role_props = Object.keys(requiredProps);
-			const missingProps = required_role_props.filter((prop) => !element.hasAttribute(prop));
-			if (missingProps.length > 0) {
-				(element as any).__astro_role = role;
-				(element as any).__astro_missing_attributes = missingProps;
-				return true;
+			
+			const elementRoles = role.split(/\s+/) as ARIARoleDefinitionKey[];
```
ematipico, line 500: nit: it would benefit the performance if we moved this regex to a top-level const, so we don't need to create a new regex every single time. 

```js
const WHITESPACE_REGEX = /s+/;
// ...

role.split(WHITESPACE_REGEX);
```

iamyunsin: Thank you for your review. I have made the necessary changes. Thank you very muchI have made the modifications as required, please help me review again, thank you.

.changeset/eight-flowers-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+When multiple roles exist, check each role. Fixing the issue of errors caused by multiple values in the role.
```
natemoo-re, line 5:
```suggestion
Fixes false positives in the dev overlay audit when multiple `role` values exist.
```

natemoo-re (approved):
Thanks for the PR!

packages/astro/src/runtime/client/dev-toolbar/apps/audit/a11y.ts
```diff
@@ -496,13 +498,17 @@ export const a11y: AuditRuleWithSelector[] = [
...
-			const required_role_props = Object.keys(requiredProps);
-			const missingProps = required_role_props.filter((prop) => !element.hasAttribute(prop));
-			if (missingProps.length > 0) {
-				(element as any).__astro_role = role;
-				(element as any).__astro_missing_attributes = missingProps;
-				return true;
+			
+			const elementRoles = role.split(WHITESPACE_REGEX) as ARIARoleDefinitionKey[];
+			for(const elementRole of elementRoles) {
+				const { requiredProps } = roles.get(elementRole)!;
```
Princesseuh, line 504: This will still fail if the role isn't a real one.

iamyunsin: > This will still fail if the role isn't a real one.

@Princesseuh Could you please provide the code snippet where you reported an error so that I can verify it again.

I have verified through my blog that when there is a multi role SVG generated by mermaid, errors are no longer reported. The code is as follows:
```html
<svg aria-roledescription="sequence" role="graphics-document document" viewBox="-50 -10 1146 825"
  style="max-width: 1146px;" xmlns="http://www.w3.org/2000/svg" width="100%" id="diagram">
...
</svg>
```
Content written in markdown:
```markdown
```mermaid
sequenceDiagram
...
```.
```

iamyunsin: Yes, but I think this situation should be modified by the application layer to the real role.

ematipico (approved):
Awesome!

packages/astro/src/runtime/client/dev-toolbar/apps/audit/a11y.ts
Princesseuh: Well, I sure hope they do, but I think I'd prefer the entire app to not crash when they don't, ha

iamyunsin: > Well, I sure hope they do, but I think I'd prefer the entire app to not crash when they don't, ha

@Princesseuh In fact, I am quite confused, because I think that although such code will increase robustness and fault tolerance, in fact, I think that wrong usage should be corrected, not acquiesced.

Princesseuh: Yes, we can tackle this in a separate PR for sure.

iamyunsin: Okay, thank you for your guidance.

Issue #9847, opened 2024-01-27
i18n bug: build format "file" & prefixDefaultLocale: false -> Astro.currentLocale of the non-default-locale indexes is wrong


philipschm1tt:
### Astro Info

```block
❯ astro info
Astro                    v4.2.6
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

build format "file" & prefixDefaultLocale: false
-> the `[lang]/index.astro` compiles to [lang].html and uses the wrong Astro.currentLocale.

In my minimum example, the default locale is German and there is also an `en/index.astro`.
The two index files display {Astro.currentLocale} in the headline.

After running `npm run build`, the dist folder contains index.html and en.html.
Both files show "de" in the headline.

### What's the expected result?

`dist/en.html` should contain "en" instead of "de" in the headline.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-rnbqdz?file=dist%2Fen.html

### Participation

- [ ] I am willing to submit a pull request for this issue.

philipschm1tt:
@ematipico, here you go.

PR #9865, opened 2024-01-29
fix(i18n): compute current locale from route instead of request

ematipico:
## Changes
Closes https://github.com/withastro/astro/issues/9847
Closes PLT-1511

During the build, the `url.pathname` contains the the extension e.g. `en.html`, so Astro could never find the correct locale.

With this fix, we use `RouteData` instead, which is an object that we attach to the request using a pipeline hook. We attach this information here:

https://github.com/withastro/astro/blob/f27f790b922247c168eb167394cea8dd4c3e3390/packages/astro/src/core/build/generate.ts#L284

And this is the value of the function

https://github.com/withastro/astro/blob/f27f790b922247c168eb167394cea8dd4c3e3390/packages/astro/src/i18n/middleware.ts#L137-L139

## Testing

I modified and existing test to render the current locale

## Docs

N/A

### DIFF
--- a/.changeset/large-kangaroos-camp.md
+++ b/.changeset/large-kangaroos-camp.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug in `Astro.currentLocale`, where the returned was incorrectly computed during the build.
--- a/.changeset/large-kangaroos-camp.md
+++ b/.changeset/large-kangaroos-camp.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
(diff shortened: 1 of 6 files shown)


.changeset/large-kangaroos-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug in `Astro.currentLocale`, where the returned was incorrectly computed during the build.
```
bluwy, line 5:
```suggestion
Fixes a bug in `Astro.currentLocale` where the value was incorrectly computed during the build.
```

Is the "returned" a typo?

PR #9867, opened 2024-01-29
chore: add rule `switch-exhaustiveness-check`

ematipico:
## Changes

This PR adds the rule [switch-exhaustiveness-check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/)

I find this rule very important because we don't want to "forget" to handle variants inside a `switch` statement.

## Testing

Current CI should pass

## Docs

N/A

### DIFF
--- a/packages/astro/src/preferences/index.ts
+++ b/packages/astro/src/preferences/index.ts
@@ -66,7 +66,12 @@ export function coerce(key: string, value: unknown) {
 		case 'boolean': {
 			if (value === 'true' || value === 1) return true;
 			if (value === 'false' || value === 0) return false;
+			break;
 		}
+		// @ematipico: what should we do with the other types?
+		//  "bigint" | "symbol" | "undefined" | "object" | "function"
+		default:
+			break;
 	}
 	return value as any;
 }
--- a/packages/astro/src/preferences/index.ts
+++ b/packages/astro/src/preferences/index.ts
@@ -68,10 +68,8 @@ export function coerce(key: string, value: unknown) {
 			if (value === 'false' || value === 0) return false;
 			break;
 		}
-		// @ematipico: what should we do with the other types?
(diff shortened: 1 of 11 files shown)


packages/astro/src/preferences/index.ts
```diff
@@ -66,7 +66,12 @@ export function coerce(key: string, value: unknown) {
 		case 'boolean': {
 			if (value === 'true' || value === 1) return true;
 			if (value === 'false' || value === 0) return false;
+			break;
 		}
+		// @ematipico: what should we do with the other types?
+		//  "bigint" | "symbol" | "undefined" | "object" | "function"
+		default:
+			break;
```
bluwy, line 74: Maybe we can throw an error here just in case, but it seems unlikely that we'll hit it.

packages/astro/src/runtime/client/dev-toolbar/apps/settings.ts
```diff
@@ -161,6 +161,10 @@ export default {
 						astroToggle.input.addEventListener('change', setting.changeEvent);
 						astroToggle.input.checked = settings.config[setting.settingKey];
 						label.append(astroToggle);
+						break;
+					}
+					default: {
+						break;
 					}
```
bluwy, line 168: I don't think we need a `{ }` here, but just a tiny nit.

packages/astro/src/runtime/server/scripts.ts
```diff
@@ -42,6 +42,9 @@ export function getPrescripts(result: SSRResult, type: PrescriptType, directive:
 			)};${islandScript}</script>`;
 		case 'directive':
 			return `<script>${getDirectiveScriptText(result, directive)}</script>`;
+		case null: {
+			break;
+		}
```
bluwy, line 47: and here too.

bluwy (approved):
I like this rule!

PR #9868, opened 2024-01-29
fix: signal request close

Gabrielbdd:
## Changes

The "signal" property on "APIContext.request" is supposed to inform when the request was aborted/canceled. Right now, it is not working and this PR fixes it.

image

## Testing

Well, this is my first time contributing to Astro, and I did not find where the code that I changed was being tested.

## Docs

I think its more like an internal behavior, and not sure where it could fit in the docs.

### DIFF
--- a/.changeset/eight-goats-peel.md
+++ b/.changeset/eight-goats-peel.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Signals request close on nodejs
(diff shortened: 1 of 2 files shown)


ematipico:
> Well, this is my first time contributing to Astro, and I did not find where the code that I changed was being tested.

This means you want to create a **new test case** that mimics the behaviour we are trying to solve. Then, your fix should confirm that we fixed the issue.

You can create a new test here: https://github.com/withastro/astro/tree/main/packages/integrations/node/test

We don't have tests against `AbortController`, so you'll have to create a new file and a new fixture.

ematipico:
Closing for inactivity. Feel free to open a new PR with the addressed comments if you're still interested.

Issue #950, opened 2021-07-30
💡 RFC: Log full snowpack errors to command line


kaleidawave:
### Background & Motivation

Spent a while trying to figure out why my site wasn't building (from `astro build`). Was getting this error message: `[static] Not found /styles.css`.

Turns out I should have looked at the line above `[snowpack] Build Result Error: There was a problem with a file build result.` to realise there was a issue building *some part of the site*.

The issue was the `public/styles.css` stylesheet was malformed and wasn't being transpiled/compiled, thus the server could not *get* it. 

The error is thrown by snowpack's server and is displayed in the browser console logs. But for building there is no way to see this issue. This is especially important when on a CI server.

It would be good for snowpack errors like this to come through in the CLI logs! 

### Possible solutions

Astro should be able to intercept these kind of issues from snowpack. But I understand snowpack does not expose full issues yet:

https://github.com/snowpackjs/snowpack/blob/ab4ddec39c02dbd126e420e510fbda694408a3c5/snowpack/src/commands/dev.ts#L709-L719

It would be good to add *some* way for snowpack to expose full error messages. Probably through a CLI flag.

Then some receiver on the astro side could repeat these out.  

### Help make it happen!

- [ ] I am willing to submit a PR to implement this change.
- [X] I am willing to submit a PR to implement this change, but would need some guidance.
- [ ] I am not willing to submit a PR to implement this change.

matthewp:
Did you run `astro --verbose`? It should log everything that snowpack does.

jasikpark:
I think maybe their point is that Snowpack isn't providing descriptive enough errors regardless of whether they're in verbose mode, though I might be wrong.

kaleidawave:
Hmm even with `--verbose` it still doesn't explain the problem in the same details as is coming through from snowpack's hmr-client script.

From CLI:

```
[17:24:53] [@snowpack/postcss-transform] transform() starting… [public\styles.css]
[access] /_snowpack/hmr-client.js
[dev] loaded /_snowpack/hmr-client.js [2ms]
[17:24:53] [snowpack] Build Result Error: There was a problem with a file build result.
[dev] loaded /styles.css [136ms]
[static] Not found /styles.css
```

From browser console:

```
[ESM-HMR] ...\styles.css
 Build Result Error: There was a problem with a file build result.
CssSyntaxError: ...\styles.css:4568:20: Unclosed string
```

drwpow:
[...]

PR #9872, opened 2024-01-29
Add regression test #950

MoustaphaDev:
## Changes

Depends on https://github.com/withastro/compiler/pull/952.
Adds a regression test for https://github.com/withastro/compiler/issues/950

## Testing

Ensure that all the slots are rendered

## Docs

N/A only tests were added


packages/astro/package.json
```diff
@@ -114,7 +114,7 @@
     "test:e2e:match": "playwright test -g"
   },
   "dependencies": {
-    "@astrojs/compiler": "^2.5.0",
+    "@astrojs/compiler": "0.0.0-mk-slot-expr-fix-20240130083013",
```
MoustaphaDev, line 117: We don't need to update the compiler version here, this was changed just to publicly validate the compiler bug fix.

Issue #9871, opened 2024-01-29
Endpoint Response.redirect header modification


friedemannsommer:
### Astro Info

```block
Astro                    v4.2.6
Node                     v20.11.0
System                   Linux (x64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/tailwind
                         @astrojs/react
```

### Describe the Bug

When returning [`Response.redirect`](https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect_static) from an endpoint, the `renderEndpoint` function tries to [set an HTTP response header](https://github.com/withastro/astro/blob/44c957f893c6bf5f5b7c78301de7b21c5975584d/packages/astro/src/runtime/server/endpoint.ts#L42-L46) on the [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object of the `handler`. This will throw a `TypeError: immutable` and return an HTTP 500 error to the client.

Note that [`new Response(..)`](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) and [`Response.json`](https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static) are not affected by this issue, even though they technically should be :thinking:.

---

This is a relatively new change introduced by [`075706f26d2e11e66ef8b52288d07e3c0fa97eb1`](https://github.com/withastro/astro/commit/075706f26d2e11e66ef8b52288d07e3c0fa97eb1).
These changes are included in Astro from version `4.2.4` and later.

### What's the expected result?

The client should receive a HTTP 30X response with a `Location` header.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-2fktbw?file=README.md

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Sad. I'll look into it.

lilnasy:
Thanks for looking into it, this is all spot on!

I didn't know Response.redirect would have immutable headers, but it would make sense. The directive would have to be something other than a header.

lilnasy:
I have written a test based on your repro here: https://github.com/withastro/astro/compare/main...lilnasy:astro:fix/9871

Feel free to take that and attempt a fix. I would guess putting the `.set()` in a try-catch block helps.

PR #9876, opened 2024-01-30
Fix endpoint response with immutable header

friedemannsommer:
## Changes

Updated the implementation of `renderEndpoint` to only append the reroute header if the response has the HTTP status code 404 or 500.
This should resolve the issue mentioned in https://github.com/withastro/astro/issues/9871.

## Testing

Added tests based on the work of @lilnasy ([cad43ea5c11e58b022df82907cd110e56dd4efed](https://github.com/withastro/astro/commit/cad43ea5c11e58b022df82907cd110e56dd4efed)).
These tests check that the response returned from an endpoint contains the expected headers and doesn't error with an HTTP 500 error.

## Docs

I don't think there is a need to update the documentation because of this change, since it's only internal response rewriting.

### DIFF
--- a/.changeset/silver-peaches-pump.md
+++ b/.changeset/silver-peaches-pump.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Updated `renderEndpoint` implementation to only append the reroute directive if the response HTTP status code is 404 or 500.
+Fixes an issue where using `Response.redirect` in an endpoint led to an error.
--- a/.changeset/silver-peaches-pump.md
+++ b/.changeset/silver-peaches-pump.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Updated `renderEndpoint` implementation to handle `Response` with immutable headers.
--- a/.changeset/silver-peaches-pump.md
+++ b/.changeset/silver-peaches-pump.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Updated `renderEndpoint` implementation to handle `Response` with immutable headers.
+Updated `renderEndpoint` implementation to only append the reroute directive if the response HTTP status code is 404 or 500.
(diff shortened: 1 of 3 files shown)


packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -42,6 +42,27 @@ export async function renderEndpoint(
...
+// This is separated into a dedicated function,
+// to allow the JavaScript runtime to optimize
+// the `renderEndpoint` implementation.
+// For more details: https://web.dev/articles/speed-v8#therefore_5
+function setRerouteDirective(response: Response): Response {
+	try {
+		response.headers.set(REROUTE_DIRECTIVE_HEADER, 'no');
+
+		return response;
+	} catch (_) {
```
lilnasy, line 57: Do you think the catch block could removed altogether?

It only fails with a 300 response, which wouldn't have rerouted anyways.

Now that I think about it, we could only set the header when response status is 404 or 500 which might be even less code.

friedemannsommer: Yes, I think you are right, in most cases it is unnecessary to set the reroute directive header.

Should i change it to only add the header if the HTTP status code is 404 or 500?

ematipico (approved):

.changeset/silver-peaches-pump.md
```diff
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Updated `renderEndpoint` implementation to handle `Response` with immutable headers.
+Updated `renderEndpoint` implementation to only append the reroute directive if the response HTTP status code is 404 or 500.
```
lilnasy, line 5:
```suggestion
Fixes an issue where using `Response.redirect` in an endpoint led to an error.
```
That perfectly describes what maintainers need to know about this change!

The changeset also gets merged into [the changelog](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md), which end-users would read. They don't need to know that a function called `renderEndpoint` exists, for example.

lilnasy (approved):
Looks great to me! The tests are a welcome addition, thanks.

I will merge soon as the tests pass.

PR #9877, opened 2024-01-30
Fix/content type gen windows

### LINKED ISSUE #9869: ContentCollections type generation generates faulty .d.ts on windows
### Astro Info

```block
Astro                    v4.2.6
Package Manager          yarn
Adapter                  none
Integrations             astro-i18next
                         @astrojs/prefetch
                         @astrojs/sitemap
                         @astrojs/vue
                         astro-compress
                         @astrojs/mdx
```

_No response_

### Describe the Bug

When running `yarn astro sync` on windows the following line will be generated at the end of the `.astro/types.d.ts` file:

```ts
type ContentConfig = typeof import("./..\src\content\config.js");
```

The dynamic import statement contains a faulty file path which results in all `CollectionEntry<xy>['data']` properties having the `any` type.

The path to the `content/config.ts` should either use only normal slashes `/` or when using backslaches they need to be escaped like so `\\`.

```ts
type ContentConfig = typeof import("./../src/content/config.js");
// or
type ContentConfig = typeof import(".\\..\\src\\content\\config.js");
```

not possible

### Participation

fabiankachlock:
## Changes

Closes #9869 

- Replaces backslashes generated by using `path.replace` on windows with forward slashes,

## Testing

All tests ran without errors. A test confirming this behaviour would be great, but I would need guidance on how to run this function with a mocked `node:path` module.

## Docs

No docs updates needed.

### DIFF
--- a/.changeset/funny-mice-help.md
+++ b/.changeset/funny-mice-help.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix content type reference to content config on windows.
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -345,8 +345,10 @@ function invalidateVirtualMod(viteServer: ViteDevServer) {
  */
 function normalizeConfigPath(from: string, to: string) {
 	const configPath = path.relative(from, to).replace(/\.ts$/, '.js');
+	// on windows `path.relative` will use backslashes, these must be replaced with forward slashes
+	const normalizedPath = configPath.replaceAll('\\', '/');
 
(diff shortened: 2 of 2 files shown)


.changeset/funny-mice-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix content type reference to content config on windows.
```
florian-lefebvre, line 5:
```suggestion
Fixes the content config type path on windows
```

ematipico (approved):

florian-lefebvre (approved):

Issue #9873, opened 2024-01-30
☂️ Move Astro tests to `node:test` 


ematipico:
## Description

Internally, the team decided to move our testing infrastructure to use `node:test` instead of Mocha.

While Mocha **is good**, it has its own quirks. Also, we believe that the Node.js test runner will improve with time. 

This is an **excellent opportunity** to contribute to Astro and help us with the migration because you don't need to understand Astro logic. So if you have been looking for an opportunity to contribute to Astro, this is the **perfect chance for you**!

## Plan

**Comment here which integration/adapter** you want to migrate first. Failing to comment might nullify your contribution if someone else did the work and **commented** on the issue. This will help us to avoid having multiple people working on the same thing.

The plan should be as follows:
1. Migrate tests of our `@astrojs/*` packages. This can be done in one PR, because we only have a few tests inside our integrations/adapters. You can use this PR as a blueprint of what we're looking for: https://github.com/withastro/astro/pull/9758
2. **Progressively** migrate the tests we have in core, starting with the unit tests, then integrations tests. This means that we need to have **two** scripts, one that will run the tests using Mocha, and one that will run the tests using Node.js

### Integrations/adapters

List of the integrations/adapters

- [x] Migrate tests of [`@astrojs/vue`](https://github.com/withastro/astro/tree/main/packages/integrations/vue)
- [x] Migrate tests of [`@astrojs/alpinejs`](https://github.com/withastro/astro/tree/main/packages/integrations/alpinejs)
- [x] Migrate tests of [`@astrojs/lit`](https://github.com/withastro/astro/tree/main/packages/integrations/lit) @xMohamd 
- [x] Migrate tests of [`@astrojs/markdoc`](https://github.com/withastro/astro/tree/main/packages/integrations/markdoc)
- [x] Migrate tests of [`@astrojs/mdx`](https://github.com/withastro/astro/tree/main/packages/integrations/mdx)
- [x] Migrate tests of [`@astrojs/partytown`](https://github.com/withastro/astro/tree/main/packages/integrations/partytown)
- [x] Migrate tests of [`@astrojs/preact`](https://github.com/withastro/astro/tree/main/packages/integrations/preact)
- [x] Migrate tests of [`@astrojs/react`](https://github.com/withastro/astro/tree/main/packages/integrations/react)
- [x] Migrate tests of [`@astrojs/sitemap`](https://github.com/withastro/astro/tree/main/packages/integrations/sitemap)
[...]

PR #9886, opened 2024-01-30
chore(telemetry): Migrate tests to `node:test`

VoxelMC:
## Changes

- Migrates `chai` tests to `node:test` assertions
- Part of #9873 

> The original test command still runs mocha tests, and a new test command: `test:node` will run the node tests.
I have renamed the mocha tests to `*.mocha.test.js` and added node tests as `*.node.test.js`. An alternative to this might just be using different directories, which I can fix if needed
The above is no longer true. I have removed the progressive tests!

## Testing

Tests should pass! Twice!

## Docs


ematipico:
Thank you @VoxelMC !

Can you give us a bit of context, and explain why a test still needs to run via Mocha?

VoxelMC:
> Thank you @VoxelMC !
> 
> Can you give us a bit of context, and explain why a test still needs to run via Mocha?

I added the mocha tests back to align with the notion of Progressively migrating the tests to `node:test`. Do these packages fall under that umbrella (RSS, this, etc.)?

ematipico:
Yeah, the idea was to progressively migrate the tests of the `astro` package. But we can migrate all the tests of the other packages because they aren't that many, so it's doable.

Plus, with this PR you gave me a nice idea of how to progressively migrate `astro` tests!

VoxelMC:
> Yeah, the idea was to progressively migrate the tests of the `astro` package. But we can migrate all the tests of the other packages because there aren't that many with one PR (per package), so it's doable.
> 
> Plus, with this PR you gave me a nice idea of how to progressively migrate `astro` tests!

Ah, I see it clearly now!
Okay, I will remove `mocha` tests from `telemetry` and then it should be ready to merge?

VoxelMC:
Done!

bluwy (approved):

ematipico (approved):

PR #9890, opened 2024-01-31
chore: add test for withastro/compiler#899

MoustaphaDev:
## Changes

Added regression test for withastro/compiler#899, which was fixed in https://github.com/withastro/compiler/pull/947.

## Testing

Added an e2e test to make sure there's no regression in the future.

## Docs

N/A this PR only adds tests


ematipico:
I tried to update the compiler version to `2.5.1`, but it seems that there are still regressions

MoustaphaDev:
Hmm, not sure what this could be, but I'll take a look later today. Thanks for checking!

Issue #9700, opened 2024-01-15
Vite config `base` and `experimental` is all not work


peng:
### Astro Info

```block
Astro                    v4.0.8
Node                     v18.18.1
System                   macOS (arm64)
Package Manager          yarn
Output                   hybrid
Adapter                  @astrojs/node
Integrations             @astrojs/react
                         @astrojs/vue
```

### If this issue only occurs in one browser, which browser is a problem?

build

### Describe the Bug

It possible not a bug.

I want to config two cdn url, js file is https://a.com/static prefix url, css file is https://b.com/static prefix url. I found vite has renderBuiltUrl mehthod.

```
import { defineConfig } from 'astro/config';

const config = {
  vite: {
    experimental: {
      renderBuiltUrl(filename, { hostType }) {
            return 'https://a.com/static'
      }
    }
  }
}

export default defineConfig(config);
```

But vite.experimental.renderBuiltUrl is not work.

Is support vite.experimental.renderBuiltUrl  config ?

### What's the expected result?

Js, css, ... file can use different cdn url

### Link to Minimal Reproducible Example

https://github.com/peng/astro-demo/

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
It's possible that Astro's version of `vite` is outdated and doesn't have that option. Make sure you are using the correct version of vite.

peng:
vite/5.0.10 darwin-arm64 node-v18.18.1, but vite `base` options is not work either.

> It's possible that Astro's version of `vite` is outdated and doesn't have that option. Make sure you are using the correct version of vite.

peng:
> In what way does it not work? I guess the feature does work partially, but it's not taking effect on all files that you expected, hence it isn't working well. Astro also uses that config internally (but you can override it) with its [assetsPrefix](https://docs.astro.build/en/reference/configuration-reference/#buildassetsprefix) option, but it's opinionated that it applies to both JS and CSS files. But that option should also cover cases where it would fail if you only used `renderBuiltUrl` yourself.

Is `vite.experimental.renderBuiltUrl` can't replace `build.assetsPrefix` options ?

peng:
> Yeah there is not support for this. Astro only supports the prefix for both JS and CSS, and you can't choose between it. If you want to see support for it, you can open a discussion in https://github.com/withastro/roadmap. Closing this for now as it's not a bug.
Ok, I have opened a discussion.  And I want open a PR to reslove it.
[...]

PR #9898, opened 2024-01-31
Add multiple cdn support

peng:
## Changes

This is not final PR. If my code is ok, later I will write test and docs.

- What does this change?

Add `build.assetsPrefix` is also function.

Astro not support multiple cdn. We disscuss is this issues: https://github.com/withastro/astro/issues/9700#issuecomment-1907226632

I have opened a discussion here: https://github.com/withastro/roadmap/discussions/818

## Docs

If my idea and code is ok,  later I will write test and docs. If not, please tell me how to modity it. Thanks!

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -13,7 +13,7 @@ import type { AddressInfo } from 'node:net';
 import type * as rollup from 'rollup';
 import type * as vite from 'vite';
 import type { RemotePattern } from '../assets/utils/remotePattern.js';
-import type { SerializedSSRManifest } from '../core/app/types.js';
+import type { SerializedSSRManifest, AssetsPrefix } from '../core/app/types.js';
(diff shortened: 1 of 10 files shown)


ematipico (changes requested):
This approach, unfortunately, doesn't work in SSR. My suggestion is to have an object like this:

```js
{
	"default": "https://cdn.example.com",
	"js": "https://cdn-js.example.com",
}
```

The new logic will match the file extension with the key of the object. If nothing is in there, you can use the default.

peng:
> This approach, unfortunately, doesn't work in SSR. My suggestion is to have an object like this:
> 
> ```js
> {
> 	"default": "https://cdn.example.com",
> 	"js": "https://cdn-js.example.com",
> }
> ```
> 
> The new logic will match the file extension with the key of the object. If nothing is in there, you can use the default.

OK, I will solve it.

PR #9899, opened 2024-01-31
chore(@astrojs/integrations/react) Migrate tests to `node:test` for `@astrojs/react` 

VoxelMC:
## Changes

- Migrates from `mocha` to `node:test` for testing assertions
- Part of #9873 
- Removes `chai` and `mocha` from devDependencies

## Testing

Tests pass!
Some skipped tests may not pass, such as the `Throws helpful error message on window SSR` test. This did not pass when using `mocha` either.

## Docs

This PR only affects testing.


ematipico (approved):
Awesome!

Issue #9854, opened 2024-01-28
Atro Build: 404.Astro only handles paths with the pathname https://example.com/404, rather than all non-existing paths.


karannakra:
### Astro Info

```block
Astro                    v4.2.6
Node                     v18.18.2
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/react
```

### Describe the Bug

Create 2 pages 

1. 404.astro
2. index.astro

Now, try entering a path for a page that does not exist. In development mode, 404 handles all unknown paths. However, while running with Astro preview, only 404 is captured. 

The 404.astro page only captures routes with the pathname 404, although it should capture all pathnames for which the path does not exist. It was working great in Astro 3. When I switched to Astro 4, this fault appeared.

**Steps to reproduce the bug**

1. Run astro build
2. Run astro preview
3. Go to some unknown path

### What's the expected result?

404 should catch all pathnames to pages that do not exist. It works properly in development mode. It does not work after astro build and astro preview.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-vf7clf?file=src%2Fpages%2F404.astro,src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #9907, opened 2024-01-31
Fix: 404.html load correctly on preview

ktym4a:
## Changes

Fix: #9854 

https://github.com/ktym4a/astro/blob/520be8b113bbb3189033e2952b28a6cda2a676a1/packages/astro/src/core/preview/vite-plugin-astro-preview.ts#L81-L98

I think previous `vite404Middleware` checks were never happen because `next();` is called before that.

so, Moved to be called.

## Testing

If this is good, I will add test cases.
but I just need to know where to add test cases.

## Docs

don't need to add.

### DIFF
--- a/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
+++ b/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
@@ -60,6 +60,7 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
 					}
 				}
 
+				// TODO: look into why the replacement needs to happen here
 				for (const middleware of server.middlewares.stack) {
 					// This hardcoded name will not break between Vite versions
 					if ((middleware.handle as Connect.HandleFunction).name === 'vite404Middleware') {
--- a/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
+++ b/packages/astro/src/core/preview/vite-plugin-astro-preview.ts
@@ -12,6 +12,18 @@ const HAS_FILE_EXTENSION_REGEXP = /^.*\.[^\\]+$/;
 export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
 	const { base, outDir, trailingSlash } = settings.config;
 
+	function handle404(req: IncomingMessage, res: ServerResponse) {
+		const errorPagePath = fileURLToPath(outDir + '/404.html');
+		if (fs.existsSync(errorPagePath)) {
+			res.statusCode = 404;
+			res.setHeader('Content-Type', 'text/html;charset=utf-8');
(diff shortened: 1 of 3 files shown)


ematipico:
@ktym4a 

I believe this file is a good place to start for testing: https://github.com/withastro/astro/blob/main/packages/astro/test/preview-routing.test.js

You might want to create a use case very similar to what was filed in the issue, and make sure it works

ktym4a:
Thanks for your kind support!
I added a changeset.

ematipico:
@lilnasy I would love your review here

packages/astro/src/core/preview/vite-plugin-astro-preview.ts
```diff
@@ -48,6 +60,13 @@ export function vitePluginAstroPreview(settings: AstroSettings): Plugin {
...
 				}
 
+				for (const middleware of server.middlewares.stack) {
+					// This hardcoded name will not break between Vite versions
+					if ((middleware.handle as Connect.HandleFunction).name === 'vite404Middleware') {
+						middleware.handle = handle404;
+					}
+				}
+
 				next();
```
lilnasy, line 70: Was it necessary to overwrite vite's middleware _during_ the execution of our other middleware?

lilnasy (approved):
Looks great to me! Just one question, non-blocking.

packages/astro/src/core/preview/vite-plugin-astro-preview.ts
ktym4a: @lilnasy 
Thanks, I'll do a little research to see if I can move somewhere properly handle vite404Middleware.

lilnasy: It's fine if you don't find it or if it takes too long. This is already quite an improvement!

lilnasy: We could add a "// TODO: look into why the replacement needs to happen here" in the meanwhile.

ktym4a: Added TODO comment.
I will research and send another PR if I think it can be improved.
Thank you very much.

ktym4a: If we do, we may need to set appType to 'custom' and add HTML middlewares, I think.

References:
https://github.com/vitejs/vite/blob/775bb5026ee1d7e15b75c8829e7f528c1b26c493/packages/vite/src/node/preview.ts#L213-L219
https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/middlewares/notFound.ts#L3
https://vitejs.dev/config/shared-options.html#apptype

PR #9908, opened 2024-01-31
fix(NodeApp): end with "Internal server error" on mid-stream error

### LINKED ISSUE #8714: Node adapter error handling in http stream: throwing async iterable shouldn't result in last HTTP/1.1 chunk being zero-length
### Astro Info

```block
Astro                    v3.2.0
Package Manager          npm
Integrations             none
```

### Describe the Bug

It's great that Astro handles [JS iterables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) in its templates and converts them to an HTTP stream, as mentioned in the [docs](https://docs.astro.build/en/guides/server-side-rendering/#streaming).

However, async iterables can throw/reject, and that case isn't handled very gracefully. (Of course, the 200 status-headers have already been sent, so that's not an option.) The Astro Node adapter only supports HTTP/1.1 (if I understand correctly), so there it's actually [fairly clear what to do](https://stackoverflow.com/a/77207289/214446): simply not terminate with the zero-length end chunk. Because (from https://github.com/marko-js/community/pull/1/):

> If a chunked response doesn’t terminate with the zero-length end chunk, the client must assume that the response was incomplete — which at the very least, [means a cache should double-check with the server before reusing the stored incomplete response](https://httpwg.org/http-core/draft-ietf-httpbis-cache-latest.html#rfc.section.3.1).

But Astro currently returns a zero-length chunk.

To reproduce:

```
---
function * generator () {
  yield 1
  throw Error('ohnoes')
}
---
<body>
  {generator()}
</body>
```

and run `npm run build && npm run preview`, then in another terminal `curl --raw -i http://localhost:4321`. This prints.

```
HTTP/1.1 200 OK
Transfer-Encoding: chunked
[...]
<body>
1
1
15
Internal server error
0

```

```
HTTP/1.1 200 OK
Transfer-Encoding: chunked
[...]
<body>
1
1
15
Internal server error
```

https://stackblitz.com/edit/github-v33sgm?file=src%2Fpages%2Findex.astro

(Note that the `curl` built into stackblitz somehow doesn't have a `--raw` option, so you need to download the example.)

lilnasy:
## Changes
- Closes #8714 
- `curl --raw -i http://localhost:4321`
<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

## Testing
Added a test case to errors.test.js

## Docs
Does not affect usage.

### DIFF
--- a/packages/integrations/node/test/errors.test.js
+++ b/packages/integrations/node/test/errors.test.js
@@ -1,4 +1,4 @@
-import * as assert from 'node:assert/strict';
+import assert from 'node:assert/strict';
 import { describe, it, before, after } from 'node:test';
 import nodejs from '../dist/index.js';
 import { loadFixture } from './test-utils.js';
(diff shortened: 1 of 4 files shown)


packages/integrations/node/test/errors.test.js
```diff
@@ -22,11 +22,26 @@ describe('Errors', () => {
 	after(async () => {
 		await devPreview.stop();
 	});
-	it('when mode is standalone', async () => {
+
+	it.skip('rejected promise in template', async () => {
```
lilnasy, line 26:
```suggestion
	it('rejected promise in template', async () => {
```

natemoo-re (approved):

ematipico (approved):

PR #9911, opened 2024-01-31
Silently ignore adapters that don't export `start()`

natemoo-re:
## Changes

- Fixes a regression introduced by https://github.com/withastro/astro/pull/9776/files#diff-524e25c0bd5feda71ed7489963414a0c1f06da3caa327933574a72b85b668bddR264
- We got a report on Discord about Cloudflare failing to build in Astro 4.2.5
- Turns out that the obfuscated way this was originally written was very intentional because it bypasses some static analysis that bundlers try to do to warn against missing exports
- I added a comment to this effect so that this doesn't happen again

## Testing

We unfortunately don't have automated adapter tests

## Docs

N/A, bug fix only

### DIFF
--- a/.changeset/tiny-moose-melt.md
+++ b/.changeset/tiny-moose-melt.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue where some adapters that do not include a `start()` export would error rather than silently proceed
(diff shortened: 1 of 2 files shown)


lilnasy (approved):
Thanks for reverting this!

alexanderniebuhr (approved):
lgtm

PR #9912, opened 2024-01-31
react: with experimentalReactChildren, convert style css string to string

nemosupremo:
## Changes

Motivation: I am use shadcn tabs which make use of React's Context. Astro does not support this as [context's are not available between islands](https://github.com/withastro/roadmap/discussions/742). I want to have tabs in `mdx` but the content of those tabs should be astro generated code blocks:

Screenshot 2024-01-31 at 12 03 19 PM

This means I want to mix React and MDX, so just using a React component island doesn't work for me. So I created a custom React component and thought I could just parse the children and build the component tree manually. The `experimentalReactChildren` solves this problem for me. However I was unable to get it to work because the props that were passed to React were valid JSX. Specifically it was sending the `style` prop as a string instead of an object.

This PR modifies the react component to convert the style prop into an object before `React.createElement` is called.

## Testing

The existing tests should suffice for this feature.

## Docs

This feature is technically a fix and should not need any documentation updates.

### DIFF
--- a/.changeset/swift-eggs-approve.md
+++ b/.changeset/swift-eggs-approve.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/react": patch
+---
+
+experimentalReactChildren: convert style string to style object compatible with react
--- a/packages/integrations/react/client.js
+++ b/packages/integrations/react/client.js
@@ -1,5 +1,6 @@
 import { createElement, startTransition } from 'react';
 import { createRoot, hydrateRoot } from 'react-dom/client';
+import styleToObject from 'style-to-object';
 import StaticHtml from './static-html.js';
 
 function isAlreadyHydrated(element) {
@@ -13,7 +14,11 @@ function isAlreadyHydrated(element) {
 function createReactElementFromDOMElement(element) {
 	let attrs = {};
 	for (const attr of element.attributes) {
-		attrs[attr.name] = attr.value;
+		if (attr.name === 'style' && typeof attr.value === 'string') {
+			attrs['style'] = styleToObject(attr.value);
+		} else {
+			attrs[attr.name] = attr.value;
+		}
 	}
 
 	return createElement(
--- a/packages/integrations/react/package.json
+++ b/packages/integrations/react/package.json
@@ -46,10 +46,13 @@
     "test": "astro-scripts test \"test/**/*.test.js\""
   },
   "dependencies": {
+    "html-escaper": "^3.0.3",
+    "style-to-object": "^1.0.5",
     "@vitejs/plugin-react": "^4.2.0",
     "ultrahtml": "^1.3.0"
   },
   "devDependencies": {
+    "@types/html-escaper": "^3.0.2",
     "@types/react": "^18.2.37",
     "@types/react-dom": "^18.2.15",
     "astro": "workspace:*",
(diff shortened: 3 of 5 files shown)


matthewp:
can you show me an example `mdx` that uses this?

nemosupremo:
Here is a sample mx that produces something similar to the screenshot I posted. The reason this can't live in it's own React component is I want to leverage remark for code block generation and syntax highlighting,  we all as having an easily editable md file

```mdx
# GET /v1/products/search

## Example

<CodeTab
client:load
  langs={["curl", "python", "javascript", "java", "go"]}
>

<div client:load>
```sh
curl --get https://api.corporate.com/v1/products/search \
  -H "Authorization: Api-Key dx_xxxxxxxxx" \
  -d q="screws"
```x
</div>

<div client:load>
```python
import requests

response = requests.get('https://api.corporate.com/v1/products/search',
  headers={
    'Authorization': 'Api-Key dx_xxxxxxxxx'
  },
  params={
    'q': 'screws'
  }
)
print(response.json())
```x
</div>
</CodeTab>
```

`CodeTab` is my own custom component that uses shadcn tabs:

```tsx
export default ({ langs, children }: Props) => {
  const content = Children.toArray(children);
  return (
    <Tabs defaultValue={langs[0]}>
      <TabsList
        className="mt-6 rounded-none rounded-t"
        style={{ backgroundColor: "#22272e", color: "#adbac7" }}
      >
        {langs.map((lang, i) => (
          <TabsTrigger key={i} value={lang} className="rounded-none">
            {lang}
          </TabsTrigger>
        ))}
      </TabsList>

      {content.map((child, i) => (
        <TabsContent key={i} value={langs[i]} className="code-tab-content mt-0">
          {child}
        </TabsContent>
      ))}
    </Tabs>
  );
};
```

Without this PR, astro crashes because React throws an exception about the style attribute (which `rehypePrettyCode` uses) is a string instead of an object.

matthewp:
Hey @nemosupremo I want to thank you for taking the time to submit the PR and explain, however we'll not be accepting this change for a couple of reasons I want to explain:

* `style-to-object` is 80kb install size and we don't want to inflate the size of client bundles unless absolutely necessary. This is for an experimental feature.
* `experimentalReactChildren` is an experiment that we're not 100% sure about. The main use-case that has triggered it is that people expect MDX to be React MDX, but it's not, even with this change, it's still Astro under the hood. Thus the need to add more and more code, like you're doing, to convert it over to React.

Since we are not sure about `experimentalReactChildren` I don't know if I would suggest investing more of your time on this, however if you really need this then I would suggest forking the integration, in which case you can make any changes you need/want.

Another possibility would be if you could reimplement this without adding the extra dependencies then we could possibly consider it.

In any event, I want to thank you for taking your time.

nemosupremo:
@matthewp 

`style-to-object` can be replaced with a simply function (ex [this ~10 liner](https://stackoverflow.com/a/68853314/807701)); I opted to use a library but that can be replaced to minimize dependencies. I can redo the PR with that, but your second point is more concerning; I don't want to burden you guys with extra code that you aren't interested in maintaining - I opened the PR because it seemed like this project was interesting in going down this avenue and supporting this use case, but if its something that astro isn't really open to supporting then it's better to not implement at all. There's no point in mucking up your roadmap for a handful of loud react users.

PR #9927, opened 2024-02-01
Fix indentation formatting in types JSDoc

delucis:
## Changes

- Tabs instead of spaces in a JSDoc type
- Noticed this in https://github.com/withastro/docs/pull/6737

## Testing

n/a docs only

## Docs

n/a docs fix

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1734,7 +1734,7 @@ export interface AstroUserConfig {
 		 *     prefixDefaultLocale: false,
 		 *		  domains: {
 		 *			  fr: "https://fr.example.com",
-		 *       es: "https://example.es"
+		 *			  es: "https://example.es"
 		 *    },
 		 *  experimental: {
 		 *     i18nDomains: true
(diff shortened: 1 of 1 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -1734,7 +1734,7 @@ export interface AstroUserConfig {
 		 *     prefixDefaultLocale: false,
 		 *		  domains: {
 		 *			  fr: "https://fr.example.com",
-		 *       es: "https://example.es"
+		 *			  es: "https://example.es"
 		 *    },
```
TheOtterlord, line 1738: it looks like this whole section might be misaligned? although that could be the mix of tabs/spaces.
also noticing that the `i18n` key is missing a closing tag?

delucis: Oh yeah… it’s all kind of wonky! Thanks — I hadn’t noticed. Will fix and push another commit.

delucis: Fixed in 0dcb036 🤞

sarah11918 (approved):
Thank you!

PR #9931, opened 2024-02-01
fix(NodeApp): fix responses with null bodies never completing

### LINKED ISSUE #9925: Node.js SSR redirect timeout
### Astro Info

```block
Astro                    v4.3.0
Package Manager          npm
                         @astrojs/react
```

_No response_

### Describe the Bug

When returning an HTTP 30X (for example via an `Astro.redirect` in a `.astro` component), the HTTP response is not flushed and therefore not completed, resulting in a timeout while waiting for it.

This behavior was introduced by 2f6d1faa6f2d6de2d4ccd2a48adf5adadc82e593 in #9908.

The HTTP response should be flushed immediately since (in most cases) there is no HTTP body.

https://stackblitz.com/edit/withastro-astro-y5eicp?file=README.md

### Participation

lilnasy:
## Changes
- Fixes #9925

## Testing
- Added a test case for redirects, thanks @friedemannsommer.
- Adjusted a test introduced in #9908 to address flakiness.
- Cherry-picked #9930 to unbreak remaining tests.

## Docs
Does not affect usage.

### DIFF
--- a/.changeset/sour-seas-hang.md
+++ b/.changeset/sour-seas-hang.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a regression where a response created with `Response.redirect` or containing `null` as the body never completed in node-based adapters.
--- a/packages/astro/src/core/app/node.ts
+++ b/packages/astro/src/core/app/node.ts
@@ -98,7 +98,7 @@ export class NodeApp extends App {
 	static async writeResponse(source: Response, destination: ServerResponse) {
 		const { status, headers, body } = source;
 		destination.writeHead(status, createOutgoingHttpHeaders(headers));
-		if (body) {
+		if (!body) return destination.end();
 			try {
 				const reader = body.getReader();
 				destination.on('close', () => {
@@ -123,7 +123,6 @@ export class NodeApp extends App {
 			} catch {
 				destination.end('Internal server error');
 			}
-		}
 	}
 }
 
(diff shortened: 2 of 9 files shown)


lilnasy:
Do you mean for the tests?

friedemannsommer:
> Do you mean for the tests?

@lilnasy Yes correct, I mean the test implementation.

lilnasy:
The test looks great to me, I want to add that. Although adding a new fixture also means adding its build-time to the CI. If you could make it use an existing fixture, I will add it.

friedemannsommer:
> The test looks great to me, I want to add that. Although adding a new fixture also means adding its build-time to the CI. If you could make it use an existing fixture, I will add it.

@lilnasy The only other fixture with the "SSR" prefix that has "@astrojs/node" in it's dependencies is "ssr-api-route". Should I migrate the tests into this fixture then?

lilnasy:
Could you use the one used in the current test?

friedemannsommer:
> Could you use the one used in the current test?

@lilnasy Sure i'll use your implementation as a base then, give me ~5 minutes.

friedemannsommer:
@lilnasy It took me a bit more than 5 minutes, sorry for that.
I've added the examples i used to your changes here: https://github.com/friedemannsommer/astro/commit/713d0191a344050a7cbc706aba8f93df17e17934

lilnasy:
I need to adjust errors.test.js. I can cherry pick your commit afterwards, thanks!

PR #9932, opened 2024-02-01
fix: don't log warning if `i18nDomains` isn't enabled

ematipico:
## Changes

This PR fixes a bug where an error log was incorrectly logged inside an adapter

## Testing

Manually tested

## Docs

N/A

### DIFF
--- a/.changeset/mean-pugs-nail.md
+++ b/.changeset/mean-pugs-nail.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a case where a warning was logged even when the feature `i18nDomains` wasn't enabled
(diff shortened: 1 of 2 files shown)


TotalLag:
Hi, thanks for this fix. Just wanted to confirm this is the same as:

```
14:54:49 [ERROR] [config] The feature "i18nDomains" is not supported (used by astro-sst).
14:54:49 [ERROR] The adapter astro-sst doesn't support the feature i18nDomains. Your project won't be built. You should not use it.
```
Is it simply logging or is it really not building either?

lilnasy:
@TotalLag That is only a log. If your build fails, it would be for another reason.

Issue #9933, opened 2024-02-01
Dev server can't handle non-UTF-8 file name


ktym4a:
### Astro Info

```block
Astro                    v4.3.0
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Return 404 on dev server when Non-UTF-8 is used for file name.
If I build and check in preview, it is works.

### What's the expected result?

html is rendered

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-zl85am?file=src%2Fpages%2Findex.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #9934, opened 2024-02-01
fix: route matching for encoded pathnames

ktym4a:
fix: #9933 

## Changes

Decode `pathnames` to check.

## Testing

I don't know where to add testcases.
Should I add in `dev-routing.test.js`? and also add in `preview-routing.test.js`?

## Docs

I don't think to update.


matthewp:
`dev-routing.test.js` would be great.

The CI failing is unrelated to your change. Sorry about that, we're working on it.

PR #9939, opened 2024-02-01
chore(`@astrojs/rss`): Migrate tests to `node:test`

VoxelMC:
## Changes

- Migrates Mocha tests to `node:test` assertions.
- Part of #9873 

I have added a comment in `rss.test.js` regarding some code additions that should be addressed before merging!
- [x] Comment resolved

## Testing

Tests pass!

## Docs

No docs changes, just for testing.


packages/astro-rss/test/rss.test.js
```diff
@@ -44,10 +46,28 @@ describe('rss', () => {
...
+
+		/*
+		parseXmlString(str, (err, actual) => {
+			assert.strictEqual(err, null);
+			parseXmlString(validXmlResult, (err, predicted) => {
+				assert.strictEqual(err, null);
+				assert.deepStrictEqual(actual, predicted);
+			});
+		});
+		*/
```
VoxelMC, line 65: Comment is here, let me know what you think about putting the function and import in `test-utils.js`! I think it works well, but I'm not sure :)

ematipico: I think it works well too! A suggestion: the function should check for errors, and if there are any, it should throw before asserting that the two results are the same. Also, we should assert the equality only of the results, not the errors too.

VoxelMC: > I think it works well too! A suggestion: the function should check for errors, and if there are any, it should throw before asserting that the two results are the same. Also, we should assert the equality only of the results, not the errors too.

Alright, can do! If you have any input regarding what it should throw, I would appreciate it! Otherwise, I will come up with something :)

VoxelMC:
@ematipico, I have made the requested changes; please let me know if this is what you had in mind!
Thank you for all of your help :)

PR #9940, opened 2024-02-01
Update package.json to fix shikiji-core error

### LINKED ISSUE #9921: astro tried to access shikiji-core, but it isn't declared in its dependencies
### Astro Info

```block
Astro                    v4.2.8
Package Manager          yarn
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When I use create astro CLI to create a new project，by step 2, yarn/npm/pnpm run dev, I got the Error like this:

Error: astro tried to access shikiji-core, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: shikiji-core (via "shikiji-core/package.json")
Required by: astro@npm:4.2.8 (via /Users/xxx/.yarn/berry/cache/astro-npm-4.2.8-b761f9a737-10c0.zip/node_modules/astro/dist/core/config/schema.js)

    at makeError (/Users/xxx/Documents/learn/astro/hello-astro/.pnp.cjs:11499:34)
    at resolveToUnqualified (/Users/xxx/Documents/learn/astro/hello-astro/.pnp.cjs:13147:21)
    at Object.resolveToUnqualified (/Users/xxx/Documents/learn/astro/hello-astro/.pnp.cjs:13327:26)
    at resolve$1 (file:///Users/xxx/Documents/learn/astro/hello-astro/.pnp.loader.mjs:1993:31)
    
When I install "shikiji-core" myself, the error was resolved.

Start server normally

add "shikiji-core" into the package.json

https://github.com/buffalo-thb/hello2

### Participation

marwan-mohamed12:
fix #9921

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -177,7 +177,8 @@
     "vitefu": "^0.2.5",
     "which-pm": "^2.1.1",
     "yargs-parser": "^21.1.1",
-    "zod": "^3.22.4"
+    "zod": "^3.22.4",
+    "shikiji-core": "^0.9.18"
   },
   "optionalDependencies": {
     "sharp": "^0.32.6"
@@ -226,7 +227,6 @@
     "remark-code-titles": "^0.1.2",
     "rollup": "^4.5.0",
     "sass": "^1.69.5",
-    "shikiji-core": "^0.9.18",
     "srcset-parse": "^1.1.0",
     "unified": "^11.0.4"
   },
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -168,6 +168,7 @@
     "semver": "^7.5.4",
     "server-destroy": "^1.0.1",
     "shikiji": "^0.9.18",
+    "shikiji-core": "^0.9.18",
     "string-width": "^7.0.0",
     "strip-ansi": "^7.1.0",
     "tsconfck": "^3.0.0",
@@ -177,8 +178,7 @@
     "vitefu": "^0.2.5",
     "which-pm": "^2.1.1",
     "yargs-parser": "^21.1.1",
-    "zod": "^3.22.4",
-    "shikiji-core": "^0.9.18"
+    "zod": "^3.22.4"
   },
   "optionalDependencies": {
     "sharp": "^0.32.6"
--- a/packages/create-astro/test/fixtures/not-empty/package.json
+++ b/packages/create-astro/test/fixtures/not-empty/package.json
@@ -5,9 +5,5 @@
     "dev": "astro dev",
     "build": "astro check && astro build",
     "preview": "astro preview"
-  },
-  "dependencies": {
-    "@astrojs/check": "^0.4.1",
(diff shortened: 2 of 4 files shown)


packages/astro/package.json
djsiddz, line 1: @marwan-mohamed12 you might want to sort the dependencies list. `shikiji-core` should come after `shikiji`.

marwan-mohamed12: Thank you, I did it and updated the pull request

djsiddz: I'm sorry, I should have been more specific. If you see the `package.json` file, the dependencies are sorted in alphabetical order. That's where you need to make the correct order. You need not change it in the '*-lock' files manually, ever. When you do `pnpm install` after making a change in the `package.json` file, it will generate the equivalent lock file.

marwan-mohamed12: No problem, I got your point thank you

packages/create-astro/test/fixtures/not-empty/package.json
```diff
@@ -5,5 +5,9 @@
     "dev": "astro dev",
     "build": "astro check && astro build",
     "preview": "astro preview"
+  },
+  "dependencies": {
+    "@astrojs/check": "^0.4.1",
+    "typescript": "^5.3.3"
```
ematipico, line 11: I don't think this change is needed

packages/astro/package.json
```diff
@@ -168,6 +168,7 @@
     "semver": "^7.5.4",
     "server-destroy": "^1.0.1",
     "shikiji": "^0.9.18",
+    "shikiji-core": "^0.9.18",
```
ematipico, line 171: You didn't update the version at all; it's the same as before: `0.9.18`

ematipico (changes requested):
There are some changes to revert and the version of `shikiji-core` is still the same

packages/astro/package.json
bluwy: I don't think the version needs to change. It's only moving it from devdeps to deps.

packages/create-astro/test/fixtures/not-empty/package.json
marwan-mohamed12: @ematipico I reverted it

packages/astro/package.json
marwan-mohamed12: Do you want to update the shikiji-core ?

xMohamd:
@ematipico As I understand from #9921, it's an easy issue to just move a package from `devDependencies` to `dependencies`  without updating, it according to @bluwy comment.
Now, the author, what should he do to merge this PR?

packages/astro/package.json
ematipico: > I don't think the version needs to change. It's only moving it from devdeps to deps.

Ah. The PR doesn't mention that. I wished there was more context in the description.

marwan-mohamed12:
Ignore this request 
Check the new one #10063

PR #9944, opened 2024-02-02
chore(@astrojs/integrations/lit): use Node.js for testing

xMohamd:
## Changes

- This PR moves @astrojs/integrations/lit to use Node.js for testing
- See #9873 for details

## Testing

Tests should pass!

## Docs


ematipico (approved):
Great contribution, thank you

PR #9952, opened 2024-02-02
feat: injectDts and codegenDir

florian-lefebvre:
## Changes

- Adds a new internal setting `codegenDir`
- Adds a new integration option `injectDts`
- Adds a new integration option `codegenDir`
- Rework astro content to use the new internal `injectDts`
- Reworks a script to use `tsconfck`

This PR initially meant to tackle https://github.com/withastro/roadmap/discussions/826 but it proved too difficult for me. I decided to reduce the PR content so it does not handle the autogenerated tsconfig. We'll have to discuss this later on, because it's really tricky!

Codegen seems to become a quite common usecase for integrations (for example: `@astrojs/db`, `astro-env`, `@keystatic/astro`, `astro-typed-api`...). Let's look at 2 cases:

### Adding files to the `.astro` dir

Before this PR, integrations had to make sure themselves that the `.astro` dir existed before adding files to it, and also had to compute the path from the root. Now, the `.astro` directory is always created before and its path can conveniently be accessed as `codegenDir` (URL). This change is also convenient for Astro core as it allows any core feature to access the `codegenDir` (through settings), not only `content` related code.

### Adding types

Before this PR, integrations authors had to create a `.d.ts` file inside the `.astro` directory, then update `src/env.d.ts` to add a relative reference to this file. Now, `.astro/types.d.ts` is always created and creates references to other .d.ts files so that there can only be one reference in `src/env.d.ts`. For now, this file is called `types.d.ts` for backward compatibility reasons (that's the former filename of `.astro/content.d.ts`) but `astro.d.ts` is reserved as it seems to be a nicer convention. That can be an Astro 5 breaking change. Also worth noting that in Astro 5, this allows us to remove the `astro/client` reference in `env.d.ts` as it's already present in `.astro/types.d.ts`.

## Testing

- [x] Existing tests should pass
- [x] New tests
	- [x] codegenDir created before integrations
	- [x] types.d.ts content
	- [x] injectDts (internal + integration)

## Docs

- [x] Changeset
- [x] Docs PR https://github.com/withastro/docs/pull/7058


packages/astro/src/@types/astro.ts
```diff
@@ -705,6 +705,23 @@ export interface AstroUserConfig {
...
+	 *
+	 * The value can be either an absolute file system path or a path relative to the project root.
+	 *
+	 * ```js
+	 * {
+	 *   typegenDir: './my-custom-cache-directory'
+	 * }
+	 * ```
+	 */
+	typegenDir?: string;
```
Princesseuh, line 723: It'd be totally okay with this not being customizable if it makes the logic more complex everywhere, fwiw.

packages/astro/src/core/dev/restart.ts
```diff
@@ -30,6 +30,7 @@ async function createRestartedContainer(
 }
 
 const configRE = new RegExp(`.*astro\.config\.((mjs)|(cjs)|(js)|(ts))$`);
+// TODO: use typegenDir
```
florian-lefebvre, line 33: I need help for this one, I guess we should generate the regex using a string and getting a relative path to root?

packages/astro/src/core/dev/restart.ts
```diff
@@ -30,6 +30,7 @@ async function createRestartedContainer(
 }
 
 const configRE = /.*astro.config.(?:mjs|cjs|js|ts)$/;
+// TODO: use codegenDir
```
florian-lefebvre, line 33: I need help for this one, I guess we should generate the regex using a string and getting a relative path to root?

packages/astro/src/config/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	ensureCodegenDirExists(codegenDir);
+	writeFileSync(new URL(filename, codegenDir), content);
+}
+
+export function ensureCodegenDirExists(codegenDir: URL) {
+	if (!existsSync(codegenDir)) {
+		mkdirSync(codegenDir, { recursive: true });
+	}
```
Fryuni, line 29: An exists check is not necessary when using `recursive: true`.

packages/astro/src/integrations/index.ts
```diff
@@ -174,6 +174,20 @@ export async function runHookConfigSetup({
...
 				logger: integrationLogger,
+				injectDts: (dts) => {
+					if (
+						updatedSettings.injectedDts.some(
+							(_dts) => _dts.filename === dts.filename && _dts.source === 'core'
+						)
+					) {
+						logger.warn(
+							null,
+							`The integration ${integration.name} is overriding a core dts: ${dts.filename}`
```
Fryuni, line 185: I was perfectly ok with the name until I read this error. Should we call them "ambient declarations", which is what they are, instead of "dts", which is a simplification of their extension?

I imagine a user would not know what a "core dts" is. Searching for "core dts" or "dts file" also doesn't help since there is a `.dts` extension. Searching for "ambient declarations" on the other hand immediately directs to TypeScript.

florian-lefebvre: I'm open to any naming! All the messages are really WIP and can be improved a lot for sure! (i'll focus on that in a 2nd phase, with documentation) For reference, `@nuxt/kit` has a [`addTypeTemplate` utility](https://github.com/nuxt/nuxt/blob/main/packages/kit/src/template.ts#L39)

scripts/smoke/check.js
```diff
@@ -64,30 +76,61 @@ function checkExamples() {
...
+		if (parseResult.tsconfig == null) {
+			return 'missing-config';
+		}
 
-		if (!tsconfig.config.compilerOptions) {
-			tsconfig.config.compilerOptions = {};
+		return parseResult;
+	} catch (e) {
+		// TODO: remove
+		console.error(e)
```
bluwy, line 95: Do we want to remove this? Personally I think it's also fine to leave it in for debugging, and if so we can remove the todo comment.

.changeset/tender-pens-sell.md
```diff
@@ -0,0 +1,46 @@
...
+
+```ts
+import type { AstroIntegration } from "astro"
+
+export function integration(): AstroIntegration {
+    return {
+        name: "my-integration",
+        hooks: {
+            "astro:config:setup": ({ codegenDir }) => {
+                const filePath = new URL("./data.json", codegenDir)
```
bluwy, line 17: Is there a reason that we expose `codegenDir` here? I think we'd want to "own" this, so not allowing or encouraging integrations to put files here. In the future if Astro wants to write a new file/folder here it could conflict with other integrations.

bluwy: Hmm I just realized that `injectDts` writes an exact `<name>.d.ts` file inside `.astro`, so this would already be an issue 🤔 I thought the types are all concat into a single `.astro/types.d.ts`. I'm not sure if that's also possible.

.changeset/tender-pens-sell.md
```diff
@@ -0,0 +1,46 @@
...
+    return {
+        name: "my-integration",
+        hooks: {
+            "astro:config:setup": ({ injectDts }) => {
+                injectDts({
+                    name: "my-integration.d.ts",
+                    content: `declare module "virtual:my-integration/config" {
+                        export const param: string;
+                    }`
+                })
```
ematipico, line 41: I have some reservations about this API. I don't think integrations should be allowed to specify a file name because there isn't a need:
- they might end up creating folders e.g. `folder/folder/folder/file.d.ts`. We don't want that.
- Astro should own file system operations, even file names
- with this approach, integrations can decide to change file names from one version to another, and the old files might end up there forever, unused

packages/astro/src/config/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	InjectedDts & { bypassValidation?: boolean; fs: typeof fsMod }) {
+	if (
+		!bypassValidation &&
+		(!filename.endsWith('.d.ts') || RESERVED_FILE_NAMES.includes(filename))
+	) {
+		throw new AstroError(AstroErrorData.InvalidInjectTypesFilename);
+	}
+
+	ensureCodegenDirExists({ codegenDir, fs });
+	fs.writeFileSync(new URL(filename, codegenDir), content);
```
ematipico, line 25: I/O operations are expensive, and we should minimise them whenever possible. 

This means that `injectDts` is going write to disk every single time we call it. Instead, we could save the contents in some internal object and do writing operations at the very end.

The object could be something like:
```js
const dts = {
	["<INTEGRATION_NAME>"]: ["contentA", "contentB"]
}
```

.changeset/tender-pens-sell.md
florian-lefebvre: About the `codegenDir` being exposed to integrations: I think it's useful to have it because integrations are going to use it anyway (eg. `@astrojs/db` or `@keystatic/astro`). If we want to avoid conflicts, I think we can define some conventions to make all of this safe. For example, we could have this structure:
```
.astro
	astro.d.ts // reference any other registered types
	astro
		content.d.ts
		env.d.ts // for example for the types rfc
	integrations
		whatever
			some-types.d.ts
```
If we go that way, that would make things safer by ensuring each integration has its own space, not conflicting with core. In that case, `codegenDir` would be different for any integration, basically `${root}/.astro/integrations/${integrationName}` and `injectDts` would then inject in this folder.
---
About `injectDts` and exact .d.ts files, this is actually on purpose. I think it's bettern mainly for debuggability. Like, I wouldn't want to debug my injected types, lost somewhere after 2k lines of astro content and before 5 other integrations. Also, it can be unsafe to have them all in one file if there are some name conflicts, eg.
```ts
// injected by integration a
type Test = 'a'

declare module 'virtual:a' {
	export const a: Test;
}

// injected by integration b
type Test = 'b'

declare module 'virtual:b' {
	export const b: Test;
}
``` 
`Test` would conflict here

packages/astro/src/core/build/index.ts
```diff
@@ -144,7 +144,10 @@ class AstroBuilder {
 		);
 		await runHookConfigDone({ settings: this.settings, logger: logger });
 
+		const { ensureCodegenDirExists } = await import('../../config/types.js')
```
ematipico, line 147: It's weird that a function that contains **logic** inside a file is called `types`. Usually, a `types` file contains only... types :D

packages/astro/src/config/types.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function ensureCodegenDirExists({ codegenDir, fs }: { codegenDir: URL; fs: typeof fsMod }) {
+	fs.mkdirSync(codegenDir, { recursive: true });
+}
```
ematipico, line 30: This function should do more things:
- it should handle errors. Sometimes, permissions issues might occur during FS operations, and we have to handle them. We should add a `try/catch`. Or, we should handle the error in the places where we call this function.
- it should **check** that the folders don't exist already. I believe Node.js errors if we try to create something that already exists

.changeset/tender-pens-sell.md
florian-lefebvre: If we go with my suggestion on Blu's comment, that would solve this issue. And to also address Fry naming concerns, it could look like this for integrations:
```ts
injectTypes(`declare module ....`)
```
That would always create a file at `${root}/.astro/integrations/${integrationName}/types.d.ts`

packages/astro/src/config/types.ts
florian-lefebvre: Actually, injectDts is only called in 2 places:
1. At the end of syncInternal, where we loop over each injected dts (both core and integrations)
2. When calling `createContentTypesGenerator`. That one is a bit unfortunate but basically, `injectDts` is called by this function after we write the `types.d.ts` file with the references. To work around this, We first inject an empty `content.d.ts` so that its part of the references and then, we call `injectDts` whenever we want in the content plugin

That's truly the only places we're we doing file writes.

However, calling `ensureCodegenDirExists` each time may not be optimal. The thing, I spent quite some time debugging and probably because I'm misunderstanding in what order parts run, I could only fix the issue by calling this function in there :/. Ideally we'd basically call it in 1 place and call it a day but I couldn't manage to find where.

packages/astro/src/core/build/index.ts
florian-lefebvre: Ahah right! The place was originally suggested by Erika but things changed quite a bit! If you have any suggestion on where to put those, let me know! The repo is so big that I struggled to find the most sensible place

packages/astro/src/config/types.ts
florian-lefebvre: Agreed concerning your first point. Regarding 2, see this review by Fry and let me know https://github.com/withastro/astro/pull/9952#discussion_r1501255184

Fryuni: [For reference](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options):

> Calling `fsPromises.mkdir()` when `path` is a directory that exists results in a rejection only when `recursive` is false.

The same applies for the callback-based API and the sync API

.changeset/tender-pens-sell.md
florian-lefebvre: Actually, yesterday i was working on an integration that required several dts files, to avoid conflicts. So I wonder how that would go with such API 🤔

ematipico: Ideally, the API will save the string inside an array. At the end everything will go inside the same file

packages/astro/src/core/build/index.ts
ematipico: I'm not good at names. Maybe `typeGen.ts`?

florian-lefebvre: Still under `config`?

ematipico: Yeah I think so

.changeset/tender-pens-sell.md
florian-lefebvre: I've made more testing and there are cases where several dts are required because stuff would conflict between them. For example:
```ts
declare module "virtual:foo" {
	export const bar: string;
}

import type _i18next0 from "../src/locales/en/test.json";
declare module "i18next" {
  interface CustomTypeOptions {
    defaultNS: "test";
    resources: {
      test: typeof _i18next0;
    };
  }
}
```
here because there's an import, the first module declaration is broken. I've tried many things but it can't work without being 2 files. So I think it could be useful to be able to create several dts per integration

sarah11918 (changes requested):
Just putting a "Docs Block" here because there still seems to be a lot of discussion going on in the changeset, which means I'm not sure that stuff has been resolved here yet. Ping me when this content is agreed-upon and finalized by everyone, then I'll review this with the docs PR together to make sure they align!

.changeset/tender-pens-sell.md
bluwy: Sorry for the late reply, I think if we want to have `codegenDir` be `.astro`, then perhaps we should rename it to `tempDir`/`astroDir`/`tempAstroDir`/`dotAstroDir`/`cacheDir` since it has multiple uses. Otherwise I think reading `config.root` would also work.

I don't think we need to take measures to avoid name conflicts for now, but perhaps putting all of the types from `injectDts` inside `.astro/types/*` should be enough, and document that integrations should use a name unique to its integration.

Good point about the `injectDts` conflict, I didn't thought of that.

.changeset/tender-pens-sell.md
ematipico: That's... unsettling. But as long as we control the name of the files, I believe it's fine and we can work with this requirement

.changeset/tender-pens-sell.md
florian-lefebvre: `cacheDir` is already taken (see [docs](https://docs.astro.build/en/reference/configuration-reference/#cachedir)). tbh I feel like internally we could call this `dotAstroDir` and for integrations still `codegenDir` because if we head into the direction of my proposal, such URL will not technically point to `.astro`. I am going to post a new comment with the updated proposal based on all the reviews

florian-lefebvre:
Thanks everyone for the amazing feedback! Based on all those, here is a proposal that I think addresses your concerns and reviews.

### `.astro` structure rework

Currently, the `.astro` dir contains only `types.d.ts`, injected by the content plugin. With this PR, here is how the structure of this folder would look:

```
.astro/
	types.d.ts // only references to other .d.ts. We keep this name for backward compatibility,
				  can be changed to something like astro.d.ts in Astro 5
	astro/
		content.d.ts // formerly content.d.ts
	integrations/
		<integration_name>/
			<dts_name>.d.ts
```

### Internal changes

- Internally, a new `dotAstroDir` (URL) is accessible from the settings to access `.astro` in an unified way
- Internally, a new `injectTypes` takes a `name` (enforced by a regex to only allow letters, dashes and underscores) and creates a `.astro/astro/<name>.d.ts`

### `astro sync` changes

This PR means that the `sync` command is now not only for core types, but for codegen in general (even from integrations). It means it's more tied to `astro:config:setup` so `syncInternal` probably needs to be called in more places.

syncInternal now creates the `.astro/types.d.ts` with all the references to any other created .d.ts file.

### Integrations API changes for `astro:config;setup`

- New property `codegenDir` (URL), available at `.astro/integrations/<integration_name>/`. This means we probably need to normalize the name somehow to avoid creating folders. Can be probably enforced by replacing problematic characters by like `_` or something
- New function `injectTypes`. Basically the same as core but inject types at `codegenDir`.

---

Let me know what you think!

.changeset/tender-pens-sell.md
Princesseuh: You need to support multiple declarations files anyway because you need to be able to output both ambient and non-ambient types.

florian-lefebvre:
I'm closing this PR in favor of a proper RFC, I didn't think this PR would get this big

Issue #9794, opened 2024-01-23
CSS updates require a dev server restart in Starlight


Princesseuh:
### Astro Info

```block
Astro                    v4.2.3
Node                     v18.18.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/starlight
```
### If this issue only occurs in one browser, which browser is a problem?

N/A

### Describe the Bug

When updating CSS inside Starlight components, the styles do not update until the dev server is restarted. Reloading the page does not fix the issue, only fully restarting the dev server does.

This has been going on in Starlight since Astro 4.0.0.

### What's the expected result?

For the CSS to update

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-trknvj?file=src%2Fcomponents%2FMyComponent.astro (change red to blue to see no changes)

### Participation

- [X] I am willing to submit a pull request for this issue.

heycassidy:
I am also seeing this same issue outside of Starlight with MDX components in a content collection. Here's a reproducible example: https://stackblitz.com/edit/github-1sy2xt?file=src%2Fcomponents%2FMyComponent.astro

0livare:
I am also seeing this behavior with MDX components.  With CSS modules, SCSS, and SCSS modules.

The CLI does pick up that the file was changed, but doesn't update the CSS contents:
```bash
12:03:27 [watch] /src/foobar.scss
```

Interestingly, I noticed that if the component is included both in the MDX & directly in the `.astro` file, then the css **WILL** update.

fisherfrontier:
Seeing the same issue in MDX files (opened #9917).

hippotastic:
The issue also seems to happen when changing the styles inside of Vue components used on a regular `.astro` page. We just had a user on Discord report this.

Princesseuh:
Bumping this issue to a p5 since it seems to be affecting more and more people, unusually.

spacedawwwg:
> The issue also seems to happen when changing the styles inside of Vue components used on a regular `.astro` page. We just had a user on Discord report this.

It's worth noting that the Vue component in question is imported from an external package in my monorepo

matthewp:
This PR contains what I think is the fix: https://github.com/withastro/astro/pull/9956

I tested it against @heycassidy's reproduction. I haven't tested against Starlight projects but I think it is the same issue. 

[...]

PR #9956, opened 2024-02-02
Fix HMR in MDX deps in Content Collections

matthewp:
## Changes

- Hopefully fixes https://github.com/withastro/astro/issues/9794
  - Definitely fixes a test case provided there
- The underlying issue is that the virtual propagated asset modules depend on the underlying MDX modules, but they also depend on their underlying assets which we get with getStylesForURL, etc. But these things are not "reactive" to changes in the module.
- This is all caused because Vite has "soft invalidation" of dependencies, the propagated asset module is not invalidated when parents change. This is correct and expected.
- The fix is to make grabbing the assets a function call, rather than a static thing that's compiled into the module definitions. This means every time `entry.render()` is called in dev, it gets fresh set of styles/scripts, so it should always be right.

## Testing

- Added a new E2E test

## Docs

N/A, bug fix


matthewp:
!preview cc-hmr

HiDeoo:
Just tested on the few Starlight related scenarios I had in my notes having that issue, and just like Hippo, it now works perfectly with the preview release :thumbsup:

.changeset/soft-bags-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix HMR in MDX deps in Content Collections
```
florian-lefebvre, line 5:
```suggestion
Fixes HMR for MDX dependencies in Content Collections
```

packages/astro/src/content/vite-plugin-content-assets.ts
```diff
@@ -48,52 +42,59 @@ export function astroContentAssetPropagationPlugin({
...
+					code = `
+async function getMod() {
+	return import(${JSON.stringify(basePath)});
+}
+
+const collectedLinks = ${stringifiedLinks};
+const collectedStyles = ${stringifiedStyles};
+const collectedScripts = ${stringifiedScripts};
+
+function getPropagatedAssets() {
```
natemoo-re, line 85: Dang. Clever solution! Not _too_ clever of course, just the right amount of clever. 👌

natemoo-re (approved):
It’s wild that this works, but I’m really glad it does! Nice one.

packages/astro/src/virtual-modules/dev-helpers.ts
```diff
@@ -0,0 +1,29 @@
+
+// TODO
```
Princesseuh, line 2: Was there something to be done here?

packages/astro/test/units/test-utils.js
```diff
@@ -62,6 +62,9 @@ class VirtualVolume extends Volume {
 class VirtualVolumeWithFallback extends VirtualVolume {
 	// Fallback to the real fs
 	readFile(p, ...args) {
+		if(p?.toString().includes('astro:dev-module-loader')) {
+			throw new Error('This module can\'t be loaded threw here. blah');
```
Princesseuh, line 66: Not sure how likely that error is, but if it's real, a good error message would be better so we don't eventually have to explain to a new contributor why that error happens, ha

Princesseuh (approved):
Just two small comments, thank you very much for taking this one! Gonna be a big productivity boost, especially for Starlight users.

bluwy (changes requested):
It seems like this could break markdoc? https://github.com/withastro/astro/blob/6d273bed13e1ba6e69f8e0c2228ce461dceed148/packages/integrations/markdoc/components/TreeNode.ts#L126-L128

I'm not very comfortable with the `.ssrModule` assignment because it means crossing an explicitly separate boundary between the "Vite-plugin runtime" and "SSR app runtime". It can also make supporting other envs in the future, like workerd in dev, harder.

<details>
<summary>(Collapsing my comments because it's outdated)</summary>


It looks like the core issue here is that if `/my/component/Foo.astro` is invalidated, we should also invalidate `/my/component/Foo.astro?astroPropagatedAssets` hardly. I can't think of a cleaner fix today, but if we need to workaround it, maybe we should directly fix the invalidation, e.g. patching `server.moduleGraph.invalidateModule`, check if we're invalidating an `.astro` module, then try to invalidate the `.astro?astroPropagatedAssets` module as well?

I think what Vite can fix, which I can try to implement, is that the `/my/component/Foo.astro?astroPropagatedAssets` transform can call `this.addWatchFile('/my/component/Foo.astro')`, which implies that even if we import it in code, whenever the watched file is changed, we should still hard-invalidate it.

</details>

EDIT: I got a better idea of the problem now today. The issue is that when any imported modules recursively down from the propagated module is updated, the propagated module also needs to be invalidated. Because `getStylesForURL` and `getScriptsForURL` crawls all of its imported modules recursively to retrieve that.

The proper way is to register all of its imported modules with `this.addWatchFile`, maybe `get...ForURL` can return that based on the walked `mod.file`. Another workaround is to simply invalidate all propagated modules whenever `handleHotUpdate` is invoked, so that they are re-transformed on the next request. Or another workaround is this PR, so I could see this PR being something worth considering too 🤔  I think I'd still lean on the first way (in this paragraph) though.

bluwy:
!preview cc-hmr

matthewp:
lgtm, willing to try this approach.

bluwy (approved):

Issue #9798, opened 2024-01-23
inline scripts in layout with view transitions not re-executing


OliverSpeir:
### Astro Info

```block
Astro                    v4.2.3
Node                     v18.19.0
System                   macOS (arm64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Before 3.5 (I think or at least around there) `<script is:inline>`s that existed on both the original and the page being transitioned to would re execute after transition and this is documented [here](https://docs.astro.build/en/guides/view-transitions/#script-behavior-during-page-navigation). 

>If you have code that sets up global state, this state will need to take into account that the script might execute more than once. Check for the global state in your <script> tag, and conditionally execute your code where possible:
```
<script is:inline>
  if(!window.SomeGlobal) {
    window.SomeGlobal = {} // ....
  }
</script>
```

Although this is not explicitly stating that inline scripts in your layout will re-execute after transition, I assume it is the spirit of providing the global check. 

Inline scripts do execute after transition (obviously), the only change is that if there is an identical script on the page that is being transitioned to, it remains marked as "already been ran" 

### What's the expected result?

I believe inline scripts should re-execute, as I think that is the point of an inline script. If not, I think there should be a way to specify built in similar to `astro-data-reload`. I'm creating this issue as a means of sparking discussion about why this behavior is happening. I might be missing something in terms of why this is important behavior. 

Side note: Stackblitz has a seperate behavior for this issue in their preview so if you do use the minimal reproduction it's better to run it locally. The `<h1>` should be turned green after moving between pages, in my opinion. 

### Link to Minimal Reproducible Example

https://github.com/OliverSpeir/vt-inline

### Participation

- [X] I am willing to submit a pull request for this issue.

martrapp:
Current behavior: Scripts of the new page are executed if they have not yet been seen (since the page was last fully loaded). We should update the documents to be consistent. 

Users often had problems when scripts that were not intended to be automatically executed multiple times. Typical errors were complaints about re-declared constants and the like.

Users should use the lifecycle events to re-execute the code where needed and adapt it where necessary to avoid re-declarations.

On the other hand: For re-execution robust inline scripts, having a way to opt-in on re-execution doesn't sound too bad to me.

morja:
I am having the exact opposite problem. See https://github.com/withastro/astro/issues/8466#issuecomment-1907084538

OliverSpeir:
[...]

PR #9977, opened 2024-02-05
Reload scripts

OliverSpeir:
## Changes

- What does this change?

Adds ability to add `data-astro-reload` to inline scripts so they will be re-executed after view transition

Issue https://github.com/withastro/astro/issues/9798 is related, in it both Mathew and Martin agreed this would be acceptable change

## Testing

Didn't add tests because this is very small change, but happy to add if someone thinks its needed

## Docs

This should be documented


martrapp:
@OliverSpeir, I'm sure you are aware that adding `data-astro-reload` turns TypeScript modules into failing inlined scripts.

```
<script>
  (function (a: string) {
    console.log(a);
  })("iife test");
</script>
```
... works just fine with Astro. After adding `data-astro-reload` to the script it only outputs 
image

This could be very confusing for users. In fact, we can't reload bundled scripts anyway. Imho we should better warn users when they try to reload a bundled script. What do you think? 

Up to now, we do not have error messages in the browser. Is there a chance to detect this server side?

OliverSpeir:
> (function (a: string) {
>     console.log(a);
>   })("iife test");

I thought about this, adding a check for this is possible and probably could be done for both `is:inline` as well. I'm not sure where I'd get started on adding that though, I assume we'd want that printed in the terminal, it could even throw an astro error 

Do you think there's a way we could inform the user better that adding this property makes it an `is:inline` script because currently that script will create the browser error with is:inline without informing the user either so what I think would be good enough for now, or at least *not worse than what we had before* would be clearly letting the user know this is only for `is:inline` scripts and will also serve the exact same purpose as the `is:inline` attribute if you don't add that 

@martrapp

martrapp:
At this time of day, my brain is too slow. You are correct that any attribute other than `src` makes an astro script inline. There is an explicit warning in the documentation. Also, with the current settings it's probably impossible to force astro to output a script with `src && data-astro-reload && type="module"`, which is the only case where we can't reload.

So yes, it's not an astro-data-reload problem, it's an auto-inlining/bundling-is-confusing-and-may-break-things problem. 
I just remember a discussion on discord the other day that we're thinking about making exceptions, like for `type="module"` that don't trigger inlining. But we can worry about it when data-astro-reload will no longer trigger inlining.

.changeset/rare-coins-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Add option to add `data-astro-reload` attribute to scripts so that they will be re-executed after transition
```
martrapp, line 5:
```suggestion
Add option to add `data-astro-reload` attribute to scripts so that they will be re-executed after view transitions
```

OliverSpeir:
> it's probably impossible to force astro to output a script with `src && data-astro-reload && type="module"`, which is the only case where we can't reload.

Ah this is something I forgot about, I should check this, does the src need to be remote for this condition? I honestly forgot about  `<script type="module" data-astro-reload>`  when writing the docs for this so I should reread them and make sure that's clear. I did test that  `<script type="module" data-astro-reload>`  will re-load

> So yes, it's not an astro-data-reload problem, it's an auto-inlining/bundling-is-confusing-and-may-break-things problem. I just remember a discussion on discord the other day that we're thinking about making exceptions, like for `type="module"` that don't trigger inlining. But we can worry about it when data-astro-reload will no longer trigger inlining.

Yeah this will be something to be aware of for sure, although 

Also something I just thought about is there's intellisense auto complete for `is:inline` on script tags, I'm not sure where that comes from but that should probably happen for this attribute

martrapp (changes requested):
Sorry for blocking, but I'm having doubts about the name right now. I would like to discuss with Erika/Matthew whether it is perfect or not so good to use `reload` as a name for both opt-in on re-execution of  scripts and opt-out from view transitions on links.

Pro for same term: 
- less cluttered name space
- no risk of confusion: opt-out only for links, re-execution only for scripts

Contra:
- not so simple to describe, e.g. in VS-Code popups "on <a>, <area>, and <form>: opt out from view transitions, on <script> elements opt-in to re-execution during view transitions"

OliverSpeir:
I see your point that makes sense! Never thought of it like that, definitely happy to change it to anything else.

I went with data-astro-reload because you opt out of view transitions' default behavior with it, but I'm not tied to it.

matthewp:
Yeah I think it makes sense to give this a different name given that it's not the same thing. `data-astro-reexecute` or something like that (hopefully we can think of something better than that!)

matthewp:
@bluwy brought up a good point, do we know if scripts (I'm thinking) 3rd party that really need this? Or do we have support threads of people who needed it? Although it sounds fine, and like something people *might* run into, I'm not super comfortable with releasing a feature until we know for sure that people have ran into a problem.

OliverSpeir:
> @bluwy brought up a good point, do we know if scripts (I'm thinking) 3rd party that really need this? Or do we have support threads of people who needed it? Although it sounds fine, and like something people _might_ run into, I'm not super comfortable with releasing a feature until we know for sure that people have ran into a problem.

It's hypothetical, I mean I'll use it but I could also work around it. I'm sure there are some 3rd party packages out there though. 

In the initial issue @matthewp  and @martrapp agreed to it, and thats why I submit this PR. I'll understand of course if we decide not to add it, but given that this is restoring a previous functionality of Astro's I think it would be nice to have. Polluting the window, and not having functions execute after every transitions unnecessarily seems like a valid use case to me. I haven't done the math, but wouldn't there be some concern about consuming memory or slowing down mobile devices if they have transitioned to many pages on a site (and thus accumulated many scripts listening for life cycle events)?

brandonedley:
> > @bluwy brought up a good point, do we know if scripts (I'm thinking) 3rd party that really need this? Or do we have support threads of people who needed it? Although it sounds fine, and like something people _might_ run into, I'm not super comfortable with releasing a feature until we know for sure that people have ran into a problem.
> 
> It's hypothetical, I mean I'll use it but I could also work around it. I'm sure there are some 3rd party packages out there though.
> 
> In the initial issue @matthewp and @martrapp agreed to it, and thats why I submit this PR. I'll understand of course if we decide not to add it, but given that this is restoring a previous functionality of Astro's I think it would be nice to have. Polluting the window, and not having functions execute after every transitions unnecessarily seems like a valid use case to me. I haven't done the math, but wouldn't there be some concern about consuming memory or slowing down mobile devices if they have transitioned to many pages on a site (and thus accumulated many scripts listening for life cycle events)?

Im currently running into this issue. Need to include an iframe on my site. Iframe is giving to me by a 3rd party. 3rd party gives me a script to include that finds an <a> link they provided and injects iframe. Then it uses iframe resizer to size the iframe. Without this fix, if a user is on the page that contains the iframe and then clicks on a link that takes them to the same page, the script does not load.

matthewp:
We discussed it and are fine with releasing this feature. We want a different name though. I think presently the best we've thought of is `data-astro-rerun`.

OliverSpeir:
rerun is nice 👍 will do

martrapp:
It's good to give different things different names.

My main motivation for #10094 and https://github.com/withastro/compiler/pull/965 was to improve errors/warnings to inform users that (what is now called) data-astro-rerun will not work with external ESM scripts.

What would be the best approach to just error when data-astro-rerun is set on a script which has its `src` attribute defined and `type="module"`?

OliverSpeir:
One nice piece of this could be a linting rule @martrapp 

https://github.com/withastro/compiler/pull/970

martrapp:
> One nice piece of this could be a linting rule

Hello Oliver, thank you very much for pointing this out! I wasn't aware of https://github.com/withastro/compiler/pull/970. The warning from https://github.com/withastro/compiler/pull/965 could well belong there!

.changeset/rare-coins-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Add option to add `data-astro-reload` attribute to scripts so that they will be re-executed after view transitions
```
bluwy, line 5:
```suggestion
Supports adding the `data-astro-rerun` attribute on script tags so that they will be re-executed after view transitions
```

bluwy (approved):
LGTM! I forgot to reply here, but yeah I've talked with Matthew that the usecase you explained makes sense to me.

martrapp (approved):
LGTM!

MoustaphaDev (approved):
LGTM!

natemoo-re (dismissed):
Blocking until minor release merge day since GitHub Actions isn't working on the fork 😭

sarah11918 (approved):
Docs approves!

PR #9967, opened 2024-02-03
FIX: allow rss feeds to have an enclosure with length of 0

madcampos:
## Changes

It allows for an `enclosure` in an RSS feed to have a length of `0`.

Reference on this change: https://www.rssboard.org/rss-profile#element-channel-item-enclosure:~:text=When%20an%20enclosure%27s%20size%20cannot%20be%20determined%2C%20a%20publisher%20SHOULD%20use%20a%20length%20of%200.

## Testing

Added unit test.

## Docs

It would be nice to point people to use a value of `0` for `enclosure` `length` if they are processing the images with Astro.

/cc @withastro/maintainers-docs

PR #9980, opened 2024-02-05
fix(rss): use node assert instead of chai

florian-lefebvre:
## Changes

- Issue introduced in #9967

## Testing

Ran `pnpm test`

## Docs

N/A


martrapp (approved):
LGTM!

Issue #9970, opened 2024-02-04
Version 4.2.7 breaks internacionalization of URLs when the locale is within rest parameters


Frondor:
### Astro Info

```block
Astro                    v4.3.2
Node                     v18.18.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

This used to work prior to 4.2.7 when this [PR ](https://github.com/withastro/astro/pull/9865) got merged.
`Astro.currentLocale` is no longer getting configured correctly since we are not splitting the path segments to find it anymore.

Provide a fallback, maybe?

### What's the expected result?

I have a variety of routes under `/blog`, for instance, these are valid routes

- `/blog` (spanish version / default locale)
- `/blog/en` (english version)
- `/blog/en/tags/some-tag`
- `/blog/tags/algun-tag`
- etc

Since we are not splitting the path anymore, Astro can't get currentLocale from URLs.

### Link to Minimal Reproducible Example

https://stackblitz.com/~/edit/withastro-astro-dtu37f?file=README.md&initialPath=%2Fblog%2Fen

### Participation

- [ ] I am willing to submit a pull request for this issue.

philipschm1tt:
I observed the same behavior for my `[...page].astro`.

I verified that the issue started exactly with Astro 4.2.7.

philipschm1tt:
Looping in @ematipico

Frondor:
I suggest to get back to the original implementation. The approach at #9865 looks a bit _invasive_ to me and as @philipschm1tt said, it may brake some i18n things.

@ematipico what do you thing about it?
I think what you wanted to fix in that PR could also be done by passing a regex to the split.
For instance:
```diff
	const requestUrl = new URL(request.url);
-	for (const segment of requestUrl.pathname.split('/')) {
+       const segments = requestUrl.pathname.split(/\/|\.\w+$/).filter(Boolean);
+	for (const segment of segments) {
```

jurajkapsz:
> I observed the same behavior for my [...page].astro.

For me the same. `Astro.currentLocale` is no longer defined.

I went with a [Middleware](https://docs.astro.build/en/guides/middleware/) solution for the time being, where I create `currentLocale` from request `pathname`, actually it works pretty well for my use case (though I guess I don't utilize the i18n suite features in my project in general).

ematipico:
@Frondor could you be more clear about the "expected result", please?

[...]

PR #9983, opened 2024-02-05
fix(i18n): match default locale correctly when it is removed because of `prefixDefaultLocale: false`

espy:
## Changes

Fixes the following i18n middleware bug:

**Bug description:** when the i18n router is in its default setting of `prefixDefaultLocale: false`, the default locale is detected and removed from the `url.pathname` too agressively, breaking URLs in which the default locale string appears as part of non-locale URL fragments, eg. `/en` appearing in `/enigma`.

**Examples** current buggy behaviour with a default locale of `en`:

- `/en/test` becomes `/test`, this is correct ✅
- `/en/random/encounter` becomes `/random/encounter`, this is also correct ✅
- `/random/encounter` becomes `/randomcounter`, this is obviously false ❌
- `/de/test/entwicklung` becomes `/de/testtwicklung`, this URL should not be touched at all, since it doesn’t actually contain the `/en` locale ❌

This can be demonstrated with a curl request against such a route:

```bash
▸ curl -vI http://localhost:4321/de/softwareentwicklung/entwicklung
*   Trying 127.0.0.1:4321...
* connect to 127.0.0.1 port 4321 failed: Connection refused
*   Trying [::1]:4321...
* Connected to localhost (::1) port 4321 (#0)
> HEAD /de/softwareentwicklung/entwicklung HTTP/1.1
> Host: localhost:4321
> User-Agent: curl/8.1.2
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Access-Control-Allow-Origin: *
< content-type: text/html
< location: /de/softwareentwicklungtwicklung
< Date: Wed, 31 Jan 2024 09:30:58 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```

The response location value of `/de/softwareentwicklungtwicklung` indicates that the default locale is matched incorrectly, since the same method is used to determine whether to respond with `404` (`pathnameContainsDefaultLocale`) and how to modify the `location` in the repsonse (`newLocation`).

The fix requires modifying two parts of the i18n middleware:
- The detection of whether the `url.pathname` actually contains the default locale _as a locale_, the simplest way was to do `pathname.split('/').includes(localeString)`. There is another method doing a similar thing, `pathnameHasLocale`, but this explicitly _shouldn’t_ match the default locale, not exactly sure why or whether it actually does. So not touching that for now.
- Correct removal of the default locale for the response `location`, in this case, using a Regex was easier.

Since Astro explicitly allows locales to be in arbitrary positions in a route ("The localized folders do not need to be at the root of the /pages/ folder.", [source](https://docs.astro.build/en/guides/internationalization/#configure-i18n-routing)), simply matching the default locale at the start of the `url.pathname` is insufficient. We only want to match the default locale when it is followed by `/` or nothing, regardless of where in the pathname it is. 

This fix does that.

## Testing

Adds tests for the i18n middleware (and modifies a fixture), and also adds unit tests for the new `removeOrReplaceLocaleFromPathname()` utility. 

## Docs

No docs changed or added because it just works as you would expect it to now.

### DIFF
--- a/packages/astro/src/i18n/index.ts
+++ b/packages/astro/src/i18n/index.ts
@@ -279,6 +279,28 @@ export function toPaths(locales: Locales): string[] {
 		}
 	});
 }
+/**
+ * Returns the pathname with a given locale either removed
+ * or optionally replaced. Does not touch parts of the
+ * pathname that look like a locale but aren’t
+ * @param pathname 
+ * @param localeToRemove 
+ * @param replacement 
+ */
+export function removeOrReplaceLocaleFromPathname(locale: string, pathname: string, replacement?: string) {
+	// We want to remove the default locale only when it is followed
+	// by `/` or nothing, not when it is part of a URL fragment, eg. 
+	// `/en/crypto/enigma` should not become 
+	// `/cryptoigma` if the default locale of `en` is stripped
+	const regex = new RegExp(`/${locale}(/|$)`)
+	let result
+	if (replacement) {
+		result = pathname.replace(regex, `$1${replacement}$1`)
+	} else {
+		result = pathname.replace(regex, '$1')
+	}
+	return result
+}
 
 function peekCodePathToUse(locales: Locales, locale: string): undefined | string {
 	for (const loopLocale of locales) {
--- a/packages/astro/src/i18n/index.ts
+++ b/packages/astro/src/i18n/index.ts
@@ -287,17 +287,12 @@ export function toPaths(locales: Locales): string[] {
  * @param localeToRemove 
  * @param replacement 
  */
-export function removeOrReplaceLocaleFromPathname(locale: string, pathname: string, replacement?: string) {
-	// We want to remove the default locale only when it is followed
-	// by `/` or nothing, not when it is part of a URL fragment, eg. 
-	// `/en/crypto/enigma` should not become 
-	// `/cryptoigma` if the default locale of `en` is stripped
-	const regex = new RegExp(`/${locale}(/|$)`)
+export function removeOrReplaceLocaleFromPathname(localeRegex: RegExp, pathname: string, replacement?: string) {
 	let result
 	if (replacement) {
-		result = pathname.replace(regex, `$1${replacement}$1`)
+		result = pathname.replace(localeRegex, `$1${replacement}$1`)
 	} else {
-		result = pathname.replace(regex, '$1')
+		result = pathname.replace(localeRegex, '$1')
 	}
 	return result
 }
@@ -324,9 +319,9 @@ class Unreachable extends Error {
 	constructor() {
 		super(
 			'Astro encountered an unexpected line of code.\n' +
-				'In most cases, this is not your fault, but a bug in astro code.\n' +
-				"If there isn't one already, please create an issue.\n" +
-				'https://astro.build/issues'
+			'In most cases, this is not your fault, but a bug in astro code.\n' +
+			"If there isn't one already, please create an issue.\n" +
+			'https://astro.build/issues'
 		);
 	}
 }
--- a/packages/astro/test/i18n-routing.test.js
+++ b/packages/astro/test/i18n-routing.test.js
@@ -218,6 +218,14 @@ describe('[DEV] i18n routing', () => {
 			expect(await response2.text()).includes('Hello world');
 		});
 
+		it('should only match the default locale string in URLs where it is actually used as a locale', async () => {
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			const response = await fixture.fetch('/new-site/spanish/enigma');
+			expect(response.status).to.equal(200);
+			expect(await response.text()).includes('Astro Enigma');
+		});
+
 		it('should return 404 when route contains the default locale', async () => {
 			const response = await fixture.fetch('/new-site/en/start');
 			expect(response.status).to.equal(404);
@@ -699,6 +707,19 @@ describe('[SSG] i18n routing', () => {
 				// success
 				return true;
 			}
+		});
+
+		it('should only match the default locale string in URLs where it is actually used as a locale', async () => {
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			let html = ''
+			try {
+				html = await fixture.readFile('/spanish/enigma/index.html');
+			} catch {
+				// catch block just here so the test can fail cleanly
+			}
+			let $ = cheerio.load(html);
+			expect($('body').text()).includes('Astro Enigma');
 		});
 	});
 
@@ -1215,6 +1236,15 @@ describe('[SSR] i18n routing', () => {
 			let request = new Request('http://example.com/new-site/fr/start');
 			let response = await app.render(request);
 			expect(response.status).to.equal(404);
+		});
+
+		it('should only match the default locale string in URLs where it is actually used as a locale', async () => {
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			let request = new Request('http://example.com/new-site/spanish/enigma');
+			let response = await app.render(request);
+			expect(response.status).to.equal(200);
+			expect(await response.text()).includes('Astro Enigma');
 		});
 	});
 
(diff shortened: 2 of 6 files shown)


packages/astro/test/i18n-routing.test.js
```diff
@@ -219,6 +219,14 @@ describe('[DEV] i18n routing', () => {
 			expect(await response2.text()).includes('Hello world');
 		});
 
+		it('should only match the default locale string in URLs where it is actually used as a locale', async () => {
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			const response = await fixture.fetch('/new-site/spanish/enigma');
+			expect(response.status).to.equal(200);
+			expect(await response.text()).includes('Astro Enigma');
```
ematipico, line 227: I believe you'd want to add an assertion for "Current Locale: es", not "Astro Enigma"

packages/astro/test/i18n-routing.test.js
```diff
@@ -701,6 +709,19 @@ describe('[SSG] i18n routing', () => {
...
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			let html = ''
+			try {
+				html = await fixture.readFile('/spanish/enigma/index.html');
+			} catch {
+				// catch block just here so the test can fail cleanly
+			}
+			let $ = cheerio.load(html);
+			expect($('body').text()).includes('Astro Enigma');
```
ematipico, line 723: I believe you'd want to add an assertion for "Current Locale: es", not "Astro Enigma"

packages/astro/test/i18n-routing.test.js
```diff
@@ -1156,6 +1177,15 @@ describe('[SSR] i18n routing', () => {
...
 			expect(response.status).to.equal(404);
 		});
+
+		it('should only match the default locale string in URLs where it is actually used as a locale', async () => {
+			// It should _not_ match `/en` from `/spanish/enigma` 
+			// and should respond with a 200
+			let request = new Request('http://example.com/new-site/spanish/enigma');
+			let response = await app.render(request);
+			expect(response.status).to.equal(200);
+			expect(await response.text()).includes('Astro Enigma');
```
ematipico, line 1187: I believe you'd want to add an assertion for "Current Locale: es", not "Astro Enigma"

packages/astro/src/i18n/index.ts
```diff
@@ -279,6 +279,28 @@ export function toPaths(locales: Locales): string[] {
...
+ * @param pathname 
+ * @param localeToRemove 
+ * @param replacement 
+ */
+export function removeOrReplaceLocaleFromPathname(locale: string, pathname: string, replacement?: string) {
+	// We want to remove the default locale only when it is followed
+	// by `/` or nothing, not when it is part of a URL fragment, eg. 
+	// `/en/crypto/enigma` should not become 
+	// `/cryptoigma` if the default locale of `en` is stripped
+	const regex = new RegExp(`/${locale}(/|$)`)
```
ematipico, line 295: This is a costly operation, because we create a new regex every single time, without "caching" regexes that we already use. Can we achieve the same result with simple string operations?

Frondor:
Hello! In my humble opinion, I believe this is a bit over-engineered.
Please, see the discussion and my _KISS_ proposal at #9970

espy:
> Hello! In my humble opinion, I believe this is a bit over-engineered. Please, see the discussion and my _KISS_ proposal at #9970

Hi @Frondor, I always appreciate a simpler solution, but I’m not sure that the things discussed in #9970 would fix this problem. If I understand this correctly, these two aren’t exactly related. That issue begins with a "this used to work in `4.2.7`", but the issue fixed in this PR has been in the i18n middleware since [the very first commit](https://github.com/withastro/astro/commit/143bacf3962f7b0ed3efe2bdfea844e72e10d288#diff-e97808f4a89c7fe4c3311be86b085986ee6dd5087f30ac288f4a1786ad16bc5cR36). Even if #9865 was reverted and implemented in a different manner, the matching/replacement logic in the style of
```js
const pathnameContainsDefaultLocale = url.pathname.includes(`/${defaultLocale}`)
```
would still be present in the middleware in several places, and still lead to the problems described above. At least that’s my understanding of the code, happy to be proved wrong, would be less work for me 😅

Frondor:
Because if we split by the pathname segments and have the file extension in consideration, this would never be the case. Right?
I think I understood your problem, but I may still be wrong. 
Please, check version `4.3.3`  if this is still a thing

espy:
Updated to `4.3.3`, and the issue persists:

```bash
(astro-i18n-routing) ▸ npm list astro
astro-i18n-routing@1.0.0 /Users/espy/Sites/astro-i18n-routing
└── astro@4.3.3

(astro-i18n-routing) ▸ curl -vI http://localhost:4321/de/softwareentwicklung/entwicklung
*   Trying 127.0.0.1:4321...
* connect to 127.0.0.1 port 4321 failed: Connection refused
*   Trying [::1]:4321...
* Connected to localhost (::1) port 4321 (#0)
> HEAD /de/softwareentwicklung/entwicklung HTTP/1.1
> Host: localhost:4321
> User-Agent: curl/8.1.2
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Access-Control-Allow-Origin: *
< content-type: text/html
< location: /de/softwareentwicklungtwicklung 👈 marf
< Date: Wed, 07 Feb 2024 09:48:17 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```

And that makes sense, since [the matcher in the i18n middleware has not been fixed](https://github.com/withastro/astro/blob/main/packages/astro/src/i18n/middleware.ts#L75) by #9998.

Ok, so given this, I’ll implement @ematipico’s change requests now.

packages/astro/test/i18n-routing.test.js
espy: Trying to do this has revealed that the `Astro.currentLocale` is sometimes not set correctly (tested on main just now, so this includes #9998).

Example: adding

```js
expect($('body').text()).includes('Current Locale: es');
```

to a test that loads a spanish page results in:

```bash
  1) [SSG] i18n routing
       i18n routing with routing strategy [prefix-other-locales]
         should render localised page correctly when it has codes+path:
     AssertionError: expected '\nEspanol\nCurrent Locale: en ' to include 'Current Locale: es'
      at Context.<anonymous> (file:///Users/espy/Sites/sandbox/astro/packages/astro/test/i18n-routing.test.js:677:29)
```

My PR and its tests are only concerned whether the requested page is loaded or erroneously `404`ed by the i18n middleware, not whether the `currentLocale` is set correctly. Would you accept my original tests, since they do that, and we can worry about `currentLocale` in a separate issue?

packages/astro/src/i18n/index.ts
espy: Wasn’t quite as simple, the fallback change added last week also needs a regex for at least one other locale, but either way, the regexes are now cached.

espy:
Well this was all a complete waste of my time then 👉 https://github.com/withastro/astro/commit/037e4f12dd2f460d66f72c9f2d992b95e74d2da9

ematipico:
> Well this was all a complete waste of my time then 👉 [037e4f1](https://github.com/withastro/astro/commit/037e4f12dd2f460d66f72c9f2d992b95e74d2da9)

Apologies @espy, that wasn't intentional. 

I want to highlight that I didn't understand that you were trying to fix https://github.com/withastro/astro/issues/10005 because you didn't link the issue. Usually, when a PR closes a specific issue, we link it in the PR description.

Also, we comment on the original issue by saying something like "I am working on this", so we can assign the issue to the person, and avoid "double work" like it happened this time.

janl:
This PR predates the issue, tho, so this doesn’t make much sense.

ematipico:
> This PR predates the issue, tho, so this doesn’t make much sense. 

I'm human, I make mistakes, I apologise and explain how I made the mistake.

espy:
@ematipico Alright, that didn’t go well, it happens. Thanks for the apology, and I could also have made things clearer by opening an issue first and not making assumptions about you seeing all the stuff I did in here. It’s a big, fast moving project, we're all busy, stuff goes wrong, nobody died, let’s keep going 🤜 🤛

In any case, `4.3.5` fixes the issue for me, so yay, I guess 🥳

PR #9985, opened 2024-02-05
Fix flaky streaming test

bluwy:
## Changes

Because components that are sibilings are rendered in parallel, the slot streaming tests we use runs with a timeout of `50` and `60`, making only a 10ms gap for the chunk separation, which I _think_ is what makes the macos-14 test keep failing. I extended the `60` to `100` so each wait has a 50ms gap.

I hope this fixes it 😄 

## Testing

Ran the local tests a bunch of times.

## Docs

n/a. test fix.


natemoo-re (approved):
Sure looks like it works!

PR #9991, opened 2024-02-06
Change view transitions rule

ktym4a:
## Changes

Change View Transition Rule.

References: https://github.com/withastro/roadmap/discussions/770#discussioncomment-8368617

## Testing

add test cases

## Docs

Might need to add. Discussion still needed


ktym4a:
@martrapp The rule change is already here.

> I guess we should not change the default prior to Astro V5 as for some sites this might be a breaking change.

Am I correct in understanding that add a parameter to `<ViewTransitions />`?

packages/astro/src/transitions/router.ts
```diff
@@ -453,9 +453,11 @@ async function transition(
 	if (navigationType !== 'traverse') {
 		updateScrollPosition({ scrollX, scrollY });
 	}
-	if (samePage(from, to) && !!to.hash) {
-		moveToLocation(to, from, options, document.title, historyState);
-		return;
+	if (samePage(from, to)) {
+		if ((direction === 'forward' && !!to.hash) || (direction === 'back' && !!from.hash)) {
```
martrapp, line 457:
```suggestion
		if ((direction !== 'back' && to.hash) || (direction === 'back' && from.hash)) {
```

martrapp: When changing `event.direction` in a listener, there might exist other values with the same meaning as `forward`

martrapp (changes requested):

martrapp:
> Am I correct in understanding that add a parameter to <ViewTransitions />? Or is it internally switched by ASTRO_VERSION?

Yes, I thought of a new `<meta>` in `ViewTransitions.astro`, which is only inserted when we switch to the new behavior. Later with V5  we can flip this and have a `<meta>` that is only there if you want the legacy behavior. Same for the prop on `<ViewTransitions />`. Something like `<ViewTransitions good-name-for-new-behavior />` and with V5 `<ViewTransitions good-name-for-legacy-behavior />`, what ever those good names may be :)

ktym4a:
>  Something like <ViewTransitions good-name-for-new-behavior /> and with V5 <ViewTransitions good-name-for-legacy-behavior />, what ever those good names may be :)

Thanks, and I understand what you mean. I will do that.

martrapp:
Impressive test btw!

packages/astro/e2e/view-transitions.test.js
```diff
@@ -413,10 +413,7 @@ test.describe('View Transitions', () => {
 		await expect(locator).toBeInViewport();
 
 		// Scroll back to top
-		// back returns immediately, but we need to wait for navigate() to complete
-		const waitForReady = page.waitForEvent('console');
 		await page.goBack();
-		await waitForReady;
 		locator = page.locator('#longpage');
 		await expect(locator).toBeInViewport();
```
martrapp, line 418: We should keep this test for now to test for current behavior

packages/astro/e2e/view-transitions.test.js
```diff
@@ -426,6 +423,85 @@ test.describe('View Transitions', () => {
...
 
+	test('View Transitions Rule', async ({ page, astro }) => {
+		let consoleCount = 0;
+		page.on('console', (msg) => {
+			// This count is used for transition events
+			if (msg.text() === 'ready') consoleCount++;
+		});
+		// Don't test back and forward '' to '', because They are not stored in the history.
+		// click '' to '' (transition)
+		await page.goto(astro.resolveUrl('/long-page'));
```
martrapp, line 434: needs some parameter (or a copy of the fixture) once the "feature toggle" is im place

ktym4a:
@martrapp 
What do you think of like this design?
```
type Trigger = 'request' | 'non-hash';

const { trigger = 'non-hash' } = Astro.props;

<ViewTransitions trigger='...' />
```

martrapp:
> What do you think of like this design?
For me it sounds a bit technical (who wonders). I trust in @matthewp who is always very good in naming things. Maybe we can even deny this is a breaking change and view it as a necessary bug fix. Matthew what do you think?

martrapp:
Once we know how to deal with "feature toggle" we also need a changeset

packages/astro/e2e/view-transitions.test.js
matthewp: Yep, agreed

ktym4a: Does that mean adding prop to <ViewTransitions />?

matthewp (approved):

packages/astro/e2e/view-transitions.test.js
martrapp: We fix the bug without adding a prop. We know how to introduce a compatibility toggle if this change would raise problems. But I guess people would rather be happy to get rid of the bug than to keep the status quo, if they care at all about this detail.

martrapp (approved):

ktym4a:
Thanks for Review.
I added changeset. So is it correct that this will be a patch?

martrapp:
> haha, Great! was just about to do that let me 👀

.changeset/wild-singers-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Change behavior of View Transitions when browser back.
```
martrapp, line 5:
```suggestion
Increases compatibility with standard browser behavior by changing where view transitions occur  on browser back navigation.  
```

martrapp:
> I added changeset.

I have tweaked it a little. 
Patch is OK now.

PR #9998, opened 2024-02-06
fix(i18n): fix regression in current locale

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/9970

I reverted the change to use `pathname` instead for `RouteData::route`.`.route` contains the actual name of the file e.g. `/blog/[...locale`, which is not what I needed.

So, I reverted the code to use `RouteData::pathname` first and then `URL::pathname` in the former isn't present. `RouteData::pathname` covers SSR cases.

This PR also fixes another issue where `Astro.currentLocale` wasn't returning the correct value in case a locale is configured using the `path` property

## Testing

I added new test cases.

## Docs

N/A

### DIFF
--- a/.changeset/shy-wolves-ring.md
+++ b/.changeset/shy-wolves-ring.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes bug in `Astro.currentLocale` wasn't returning the correct locale when a locale is configured via `path`
+Fixes a bug in `Astro.currentLocale` that wasn't returning the correct locale when a locale is configured via `path`
--- a/.changeset/shy-wolves-ring.md
+++ b/.changeset/shy-wolves-ring.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes bug in `Astro.currentLocale` wasn't returning the correct locale when a locale is configured via `path`
--- a/.changeset/six-fishes-beg.md
+++ b/.changeset/six-fishes-beg.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an regression in `Astro.currentLocale`, where it stopped working properly with dynamic routes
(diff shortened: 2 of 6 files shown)


.changeset/shy-wolves-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes bug in `Astro.currentLocale` wasn't returning the correct locale when a locale is configured via `path`
```
ematipico, line 5:
```suggestion
Fixes a bug in `Astro.currentLocale` that wasn't returning the correct locale when a locale is configured via `path`
```

packages/astro/test/i18n-routing.test.js
```diff
@@ -1525,6 +1585,15 @@ describe('[SSR] i18n routing', () => {
 				expect(await response.text()).includes('Current Locale: en');
 			});
 
+			it('should return the default locale when rendering a route with spread operator', async () => {
+				let request = new Request('http://example.com/blog/es', {});
+				let response = await app.render(request);
+				let text = await response.text();
+				console.log('text', text);
```
bluwy, line 1592:
```suggestion
```

bluwy (approved):
Fix looks good to me. A bit worried about the perf implication though since it has 3 nested loops, but we could revisit them next time. Also, is the two changesets intentional or should there only be one?

ematipico:
> Also, is the two changesets intentional or should there only be one?

It's intentional, because the two bugs are unrelated (two different use cases), even though the function is the same.

.changeset/six-fishes-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an regression in `Astro.currentLocale`, where it stopped working properly with dynamic routes
```
bluwy, line 5:
```suggestion
Fixes a regression in `Astro.currentLocale` where it stopped working properly with dynamic routes
```

PR #10002, opened 2024-02-06
chore: move i18n routing tests

ematipico:
## Changes

- new `test:node` command to run only `*.nodetest.js` files
- moved `i18n-routing.test.js` to use Node.js test runner

> [!WARNING]
> It seems that Node.js test runner can't very deep nested `describe` functions. If your test hangs, try to lift those `describe` functions.

## Testing

CI should pass


packages/astro/test/i18n-routing.nodetest.js
```diff
@@ -22,14 +23,14 @@ describe('astro:i18n virtual module', () => {
...
-		expect(response.status).to.equal(200);
+		assert.equal(response.status, 200);
 		const text = await response.text();
-		expect(text).includes("Virtual module doesn't break");
-		expect(text).includes('About: /pt/about');
-		expect(text).includes('About spanish: /spanish/about');
-		expect(text).includes('Spain path: spanish');
-		expect(text).includes('Preferred path: es');
-		expect(text).includes('About it: /it/about');
+		assert.equal(text.includes("Virtual module doesn't break"), true);
```
VoxelMC, line 28: I would favour using `assert.strictEqual` because it `assert.equal` is an alias for it, but it's probably not that big of a deal and we can cross that bridge when it deprecates. Just a thought :)

VoxelMC (commented):
LGTM!

packages/astro/test/i18n-routing.nodetest.js
ematipico: If it's an alias, then `assert.equal` should be fine too. Also, less typing 😆

bluwy: We're importing from `node:assert/strict`, so I don't think we have to use `strictEqual` here

PR #10003, opened 2024-02-06
Fix: support `strict()` on content collection schemas when `slug` is present

### LINKED ISSUE #8871: Content collections doesn't allow `.strict()`
### Astro Info

```block
❯ pnpm astro info

Astro version            v2.10.15
Package manager          pnpm
Platform                 darwin
Architecture             arm64
Integrations             @astrojs/react, @astrojs/image, @astrojs/mdx, my-global-css, @astrojs/sitemap
```

_No response_

### Describe the Bug

Discord context: https://discordapp.com/channels/830184174198718474/872579324446928896/1162124756213059594

@delucis describes the problem well:

> `z.object({ foo: z.string() }).strict()` will throw if it finds a slug in frontmatter.
> But `z.object({ foo: z.string(), slug: z.string().optional() }).strict()` will also throw, because Astro doesn’t led you add slug to your own schema

I would suggest adding an additional custom schema provided by the `defineCollection` callback, similar to the `image` schema: `({ image, slug }) => {`

It would default to `slug: z.string().optional()`, but would allow the end user to then append `.refine()` and other steps to the schema to ensure a stricter shape of their slugs, like `/YYYY-MM-DD/lowecase_title_with_underscores` as a pattern.

Then loosening the astro requirement of "no slug schemas" to "only the astro built slug schema w/ refinements is allowed" which would enable setting `.strict()` on the collection schema

I would expect that I can set `.strict()` on my content collection schemas, and that they will not error with "unexpected key `slug` found", since Astro currently prevents me from setting a `slug` property key in my content collection.

https://stackblitz.com/edit/github-wtddxv?file=src%2Fcontent%2Fposts%2Ftest.md

### Participation

bholmesdev:
## Changes

Resolves #8871

Support `strict()` extension on content collection schemas when `slug` is present. Before, we would accidentally parse `slug` with the user's content schema.

## Testing

- Add `.strict()` to the content collection slug fixture

## Docs

N/A

### DIFF
--- a/.changeset/strong-pets-rhyme.md
+++ b/.changeset/strong-pets-rhyme.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Support `.strict()` on content collection schemas when a custom `slug` is present.
--- a/.changeset/strong-pets-rhyme.md
+++ b/.changeset/strong-pets-rhyme.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Support `.strict()` on content collection schemas when a custom `slug` is present.
+Adds support for `.strict()` on content collection schemas when a custom `slug` is present.
--- a/packages/astro/src/content/utils.ts
+++ b/packages/astro/src/content/utils.ts
@@ -126,7 +126,7 @@ export async function getEntryData(
 
 		// Use `safeParseAsync` to allow async transforms
 		let formattedError;
-		const parsed = await (schema as z.ZodSchema).safeParseAsync(entry.unvalidatedData, {
+		const parsed = await (schema as z.ZodSchema).safeParseAsync(data, {
 			errorMap(error, ctx) {
 				if (error.code === 'custom' && error.params?.isHoistedAstroError) {
 					formattedError = error.params?.astroError;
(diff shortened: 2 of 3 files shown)


.changeset/strong-pets-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Support `.strict()` on content collection schemas when a custom `slug` is present.
```
florian-lefebvre, line 5:
```suggestion
Adds support for `.strict()` on Content Collection schemas when a custom `slug` is present
```

delucis (approved):
Looks good to me! Thanks for following up on this one @bholmesdev 🚀

.changeset/strong-pets-rhyme.md
delucis: We don’t usually capitalize “content collection”! But I do agree with “Adds support for…” vs “Support…” as per https://contribute.docs.astro.build/guides/new-feature-docs/#changesets

```suggestion
Adds support for `.strict()` on content collection schemas when a custom `slug` is present.
```

florian-lefebvre: Ah didn't know that! I'll remember it

PR #10015, opened 2024-02-07
feat(audits): Add initial perf audits

Princesseuh:
## Changes

This PR adds a beginning of perf audits and infrastructure for it, most notably, this includes a `server-render-time` and `client-render-time` attribute on islands, allowing us to provide audits when an island take a suspicious amount of time to render.

## Testing

Will do

## Docs

N/A

### DIFF
--- a/packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
+++ b/packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
@@ -157,10 +157,10 @@ export default {
 						}
 					</style>
 					<header>
-						<h1><astro-dev-toolbar-icon icon="check-circle"></astro-dev-toolbar-icon>No accessibility issues detected.</h1>
+						<h1><astro-dev-toolbar-icon icon="check-circle"></astro-dev-toolbar-icon>No accessibility or performance issues detected.</h1>
 					</header>
 					<p>
-						Nice work! This app scans the page and highlights common accessibility issues for you, like a missing "alt" attribute on an image.
+						Nice work! This app scans the page and highlights common accessibility and performance issues for you, like a missing "alt" attribute on an image, or a image not using performant attributes.
 					</p>
 					`
 				);
--- a/packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
+++ b/packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
@@ -8,6 +8,7 @@ import {
 } from '../utils/highlight.js';
 import { createWindowElement } from '../utils/window.js';
 import { a11y } from './a11y.js';
+import { perf } from './perf.js';
 
 const icon =
 	'<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 1 20 16"><path fill="#fff" d="M.6 2A1.1 1.1 0 0 1 1.7.9h16.6a1.1 1.1 0 1 1 0 2.2H1.6A1.1 1.1 0 0 1 .8 2Zm1.1 7.1h6a1.1 1.1 0 0 0 0-2.2h-6a1.1 1.1 0 0 0 0 2.2ZM9.3 13H1.8a1.1 1.1 0 1 0 0 2.2h7.5a1.1 1.1 0 1 0 0-2.2Zm11.3 1.9a1.1 1.1 0 0 1-1.5 0l-1.7-1.7a4.1 4.1 0 1 1 1.6-1.6l1.6 1.7a1.1 1.1 0 0 1 0 1.6Zm-5.3-3.4a1.9 1.9 0 1 0 0-3.8 1.9 1.9 0 0 0 0 3.8Z"/></svg>';
@@ -28,10 +29,20 @@ export interface ResolvedAuditRule {
 
 export interface AuditRuleWithSelector extends AuditRule {
 	selector: string;
-	match?: (element: Element) => boolean | null | undefined | void;
+	match?: (
+		element: Element
+	) =>
+		| boolean
+		| null
+		| undefined
+		| void
+		| Promise<boolean>
+		| Promise<void>
+		| Promise<null>
+		| Promise<undefined>;
 }
 
-const rules = [...a11y];
+const rules = [...a11y, ...perf];
 
 const dynamicAuditRuleKeys: Array<keyof AuditRule> = ['title', 'message'];
 function resolveAuditRule(rule: AuditRule, element: Element): ResolvedAuditRule {
@@ -93,7 +104,7 @@ export default {
 					matches = Array.from(elements);
 				} else {
 					for (const element of elements) {
-						if (rule.match(element)) {
+						if (await rule.match(element)) {
 							matches.push(element);
 						}
 					}
--- a/packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
+++ b/packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
@@ -1,0 +1,73 @@
+import type { AuditRuleWithSelector } from './index.js';
+
+// A regular expression to match external URLs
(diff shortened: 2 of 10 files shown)


packages/astro/src/runtime/server/astro-island.ts
```diff
@@ -179,8 +179,9 @@ declare const Astro: {
 						throw e;
 					}
 					let hydrationTimeStart;
+					const hydrator = this.hydrator(this);
 					if (process.env.NODE_ENV === 'development') hydrationTimeStart = performance.now();
-					await this.hydrator(this)(this.Component, props, slots, {
+					await hydrator(this.Component, props, slots, {
```
Princesseuh, line 184: I'm not sure if this *actually* does something, but when testing locally, I was getting times that made a bit more sense and were generally more stable 🤷‍♀️

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -327,6 +330,8 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 		metadata as Required<AstroComponentMetadata>
 	);
 
+	island.props['server-render-time'] = componentServerRenderEndTime;
```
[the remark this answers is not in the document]
Princesseuh: Ah, yes! Thank you

packages/astro/e2e/dev-toolbar.test.js
```diff
@@ -90,25 +90,26 @@ test.describe('Dev Toolbar', () => {
 		await expect(xrayWindow.locator('astro-dev-toolbar-icon[icon=lightbulb]')).toBeVisible();
 	});
 
-	test('audit shows higlights and tooltips', async ({ page, astro }) => {
+	test('audit shows higlights and tooltips zzz', async ({ page, astro }) => {
```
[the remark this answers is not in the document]
Princesseuh: Oops, one day I'll stop commit my Erika-ism nonsenses.

packages/astro/e2e/fixtures/dev-toolbar/src/pages/audits-perf.astro
```diff
@@ -0,0 +1,9 @@
+---
+import { Image } from "astro:assets";
+---
+
+<Image src="https://source.unsplash.com/random/800x600" loading="lazy" width={800} height={600} alt="Random Image" />
+
+<div style="height: 9000px;">
+
+<Image src="https://source.unsplash.com/random/800x600" loading="eager" width={800} height={600} alt="Random Image" />
```
bluwy, line 9: Is it safe to use unsplash in tests? Also, the div here isn't working as expected I think because you didn't close its tag, so the second image is inside the div and becomes in the fold.

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
```diff
@@ -0,0 +1,126 @@
...
+		selector: 'img:not([data-image-component])',
+		async match(element) {
+			const src = element.getAttribute('src');
+			if (!src) return false;
+
+			// Don't match data URIs, they're typically used for specific use-cases that the image component doesn't help with
+			if (src.startsWith('data:')) return false;
+
+			// Ignore images that are smaller than 20KB, most of the time the image component won't really help with these, or they're used for specific use-cases (pixel tracking, etc.)
+			// Ignore this test for remote images for now, fetching them can be very slow and possibly dangerous
```
ematipico, line 20: Will we document these exceptions somewhere?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
```diff
@@ -0,0 +1,126 @@
...
+			if (!externalUrlRe.test(src)) {
+				const imageData = await fetch(src).then((response) => response.blob());
+				if (imageData.size < 20 * 1024) return false;
```
ematipico, line 23: `20 * 1024` could be a top-level constant

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
```diff
@@ -0,0 +1,126 @@
...
+	{
+		code: 'perf-use-loading-eager',
+		title: 'Use the loading="eager" attribute',
+		message: (element) =>
+			`This ${element.nodeName} tag is above the fold and could be eagerly-loaded to improve performance.`,
+		selector: 'img[loading="lazy"], iframe[loading="lazy"]',
+		match(element) {
+			const htmlElement = element as HTMLImageElement | HTMLIFrameElement;
+			console.log(htmlElement);
+			console.log({ offsetTop: htmlElement.offsetTop, innerHeight: window.innerHeight });
```
ematipico, line 53: To delete

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
```diff
@@ -0,0 +1,126 @@
...
+			if (htmlElement.offsetTop > window.innerHeight) return false;
+
+			return true;
+		},
+	},
+	{
+		code: 'perf-use-videos',
+		title: 'Use videos instead of GIFs for large animations',
+		message:
+			'This GIF could be replaced with a video to reduce its file size and improve performance.',
```
ematipico, line 64: We should tell the user what "large" means, maybe by explicitly tell the user what's the threshold of this audit

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
```diff
@@ -0,0 +1,126 @@
...
+			return true;
+		},
+	},
+	{
+		code: 'perf-slow-component-server-render',
+		title: 'Server-rendered component took a long time to render',
+		message: (element) =>
+			`This component took an unusually long time to render on the server (${getCleanRenderingTime(
+				element.getAttribute('server-render-time')
+			)}.). This might be a sign that it's doing too much work on the server, or something is blocking rendering.`,
```
ematipico, line 88:
```suggestion
			)}). This might be a sign that it's doing too much work on the server, or something is blocking the rendering.`,
```

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -28,10 +29,20 @@ export interface ResolvedAuditRule {
...
+		element: Element
+	) =>
+		| boolean
+		| null
+		| undefined
+		| void
+		| Promise<boolean>
+		| Promise<void>
+		| Promise<null>
+		| Promise<undefined>;
```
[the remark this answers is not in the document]
Princesseuh: Wouldn't be breaking, but would require a whole lot of `async` added. This is an internal thing, users don't interact with this

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
Princesseuh: Eventually, yes! We're planning to have a list of audit page on the docs, like we have for errors. Right now it's mostly about avoiding common false positives

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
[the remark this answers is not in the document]
ematipico: Makes sense. It's just that it's a lot of types to maintain

sarah11918 (approved):
For the sake of a visible checkmark, docs is approving!

packages/astro/src/runtime/client/dev-toolbar/apps/audit/perf.ts
Princesseuh: In my opinion, the sentence makes more sense with "blocking rendering" than "blocking the rendering", the latter naturally ask the question of "blocking the rendering of what?" whereas the former implies that it's blocking rendering in general. Could be wrong though

ematipico (approved):

PR #10016, opened 2024-02-07
fix(i18n): don't consider URLs that start with the name of the defaut locale

### LINKED ISSUE #10005: Dynamic URI starting with `/{AstroConfig.i18n.defaultLocale}` will result in a browser 404
### Astro Info

```block
Astro                    v4.3.3
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

On a clean installation, if using i18N Routing from Astro 3.5.0, the following will happen:

If you have one URI starting with /{AstroConfig.i18n.defaultLocale}, the page will failed to resolve. 

So, given this config: 

```

// https://astro.build/config
export default defineConfig({
  i18n: {
    defaultLocale: 'fr',
    locales: ['en', 'fr'],
    routing: {
      prefixDefaultLocale: false,
    },
  },
});
```

and this "catch-all" page : 

``` [...uri].astro
---
  export function getStaticPaths() {
  return [{ params: { uri: '/test' }}, { params: { uri: '/france' }}]
}
---
Test Page
```

the following examples URL would fails to resolve:
- /france
- /fr 

the following would work:
- /test

anything else would display the Astro 404.

As a note : 
- On the Stacblitz link, only `/france` and `/fr`  would get this browser 404 blank page - all others page (like `/frwhaever` or `/tag/france`) would get the Astro 404. 
- On my Vercel, all the page including `/fr` anywhere in their URL would get the Astro 404 even `/fr and /france`. 
- And locally any pages with `/fr` anywhere in the URL (example : `/fr`, `/france`, `/frwhatever`, `/tag/france`) would get the blank page/browser 404 treatment.

The page needs to resolve correctly even if it's starting with `/fr`

https://stackblitz.com/edit/withastro-astro-iqt5vg?file=src%2Fpages%2F%5B...uri%5D.astro

### Participation

ematipico:
## Changes

Closes  #10005

The logic that checks the presence of the default locale was very broad and poor. This new logic should be more robust.

## Testing

I added new test cases 

## Docs

N/A

### DIFF
--- a/.changeset/tough-socks-change.md
+++ b/.changeset/tough-socks-change.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug where routes with a name that start with the name of the `i18n.defaultLocale` were incorrectly returning a 404 response.
--- a/packages/astro/src/i18n/middleware.ts
+++ b/packages/astro/src/i18n/middleware.ts
@@ -65,7 +65,13 @@ export function createI18nMiddleware(
 	};
 
 	const prefixOtherLocales = (url: URL, response: Response): Response | undefined => {
-		const pathnameContainsDefaultLocale = url.pathname.includes(`/${i18n.defaultLocale}`);
+		let pathnameContainsDefaultLocale = false;
+		for (const segment of url.pathname.split('/')) {
+			if (normalizeTheLocale(segment) === normalizeTheLocale(i18n.defaultLocale)) {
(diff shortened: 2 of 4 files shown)


bluwy (approved):

janl:
heads up that we had already fixed this here https://github.com/withastro/astro/pull/9983 and now have nothing to show for but wasted time. Not very encouraging to ever contribute again here. ✌️

PR #10021, opened 2024-02-07
chore: Migrate `astro-markdown-remarkRehype.test.js` to `node:test`

VoxelMC:
## Changes

- Part of #9873 
- Progressively migrates `packages/astro/test/astro-markdown-remarkRehype.nodetest.js`

## Testing

Tests all pass.

## Docs

Test-only change.


ematipico (changes requested):
We should remove the old `astro-markdown-remarkRehype.test.js` or at least remove the case that is now covered by the new test.

We don't want to test the same use cases twice

VoxelMC:
> We should remove the old `astro-markdown-remarkRehype.test.js` or at least remove the case that is now covered by the new test.
> 
> We don't want to test the same use cases twice

Will do for all three of my open PRs, then get back to you :)

VoxelMC:
> We should remove the old `astro-markdown-remarkRehype.test.js` or at least remove the case that is now covered by the new test.
> 
> We don't want to test the same use cases twice

Done!

PR #10060, opened 2024-02-08
fix(integration api): improve module resolution

lilnasy:
## Changes

image

## Testing
Passes on a third party package without `chdir` after these changes. Existing tests should pass.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/integrations/index.ts
+++ b/packages/astro/src/integrations/index.ts
@@ -153,7 +153,8 @@ export async function runHookConfigSetup({
 							`The "${integration.name}" integration is trying to add the "${name}" client directive, but it already exists.`
 						);
 					}
-					addedClientDirectives.set(name, buildClientDirectiveEntrypoint(name, entrypoint));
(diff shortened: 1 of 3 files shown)


packages/astro/src/integrations/index.ts
```diff
@@ -153,7 +153,8 @@ export async function runHookConfigSetup({
 							`The "${integration.name}" integration is trying to add the "${name}" client directive, but it already exists.`
 						);
 					}
-					addedClientDirectives.set(name, buildClientDirectiveEntrypoint(name, entrypoint));
+					// TODO: this should be performed after astro:config:done
+					addedClientDirectives.set(name, buildClientDirectiveEntrypoint(name, entrypoint, settings.config.root));
```
lilnasy, line 157: That's a corner case for another day (year).

natemoo-re (approved):
Good catch!

bluwy (approved):

ematipico (approved):

PR #10074, opened 2024-02-09
feat(audits): Add UI with list of audits

Princesseuh:
## Changes

Add a small window showing the list of problems to the audit app. Clicking on an audit will scroll the page and focus the problem in question, showing you more information on what happened through the tooltip.

## Testing

Added a test that the window exists, we'll be expanding those tests in the coming weeks when the content of the window will change consequentially

## Docs

N/A


packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -111,6 +116,119 @@ export default {
...
+						  cursor: pointer;
+						}
+					</style>`;
+
+					card.clickAction = () => {
+						audit.highlightElement.scrollIntoView();
+						audit.highlightElement.focus();
+					};
+					const h3 = document.createElement('h3');
+					h3.innerText = finder(audit.auditedElement);
```
Princesseuh, line 218: Used a package for this, because it turns out that this is super complicated to do correctly

ematipico (approved):

PR #10078, opened 2024-02-10
chore: Migrate some `astro-*.test.js` to `node:test`

xMohamd:
## Changes

- See #9873
- Migrates some  `packages/astro/test/astro-*.test.js`

## Testing

All tests pass.

## Docs

N/A


packages/astro/test/astro-class-list.nodetest.js
```diff
@@ -0,0 +1,63 @@
...
+	before(async () => {
+		fixture = await loadFixture({ root: './fixtures/astro-class-list/' });
+		await fixture.build();
+	});
+
+	it('Passes class:list attributes as expected to elements', async () => {
+		const html = await fixture.readFile('/index.html');
+		const $ = cheerio.load(html);
+
+		assert.strictEqual($('[class="test control"]').length, 1, '[class="test control"]');
```
ematipico, line 18: `assert.equal` is an alias of `assert.strictEqual`, because we are importing `node:assert/strict`. So we can use `assert.equal`, like we are doing in all tests. Would you mind changing it?

PR #10079, opened 2024-02-10
Fix to get Integration Data even when a page is transitioned by ViewTransition

ktym4a:
## Changes

Fix to get Integration Data even when a page is transitioned by ViewTransition.

## Testing

add test case

## Docs

N/A

### DIFF
--- a/packages/astro/e2e/dev-toolbar.test.js
+++ b/packages/astro/e2e/dev-toolbar.test.js
@@ -51,6 +51,28 @@ test.describe('Dev Toolbar', () => {
 		await expect(astroWindow).not.toBeVisible();
 	});
 
+	test('show integration data app', async ({ page, astro }) => {
+		await page.goto(astro.resolveUrl('/view-transition-a'));
+
+		let toolbar = page.locator('astro-dev-toolbar');
+		let appButton = toolbar.locator('button[data-app-id="astro"]');
+		await appButton.click();
+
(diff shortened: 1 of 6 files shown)


florian-lefebvre:
Can you add a changeset?

ktym4a:
added!

packages/astro/e2e/dev-toolbar.test.js
```diff
@@ -51,6 +51,28 @@ test.describe('Dev Toolbar', () => {
...
+	test('show integration data app', async ({ page, astro }) => {
+		await page.goto(astro.resolveUrl('/view-transition-a'));
+
+		let toolbar = page.locator('astro-dev-toolbar');
+		let appButton = toolbar.locator('button[data-app-id="astro"]');
+		await appButton.click();
+
+		let astroAppCanvas = toolbar.locator('astro-dev-toolbar-app-canvas[data-app-id="astro"]');
+		let astroIntegrationSkelton = astroAppCanvas.locator('astro-dev-toolbar-card');
+		await expect(astroIntegrationSkelton).toHaveCount(2);
```
Princesseuh, line 63: The counts here are reliable right now, but won't be in the future. It's intended that eventually, a more complete list will be shown. Could this test be adapted to perhaps just check that some elements exists, but not a specific count?

ktym4a: Thanks,
There was also a mistake in the test case, which has also been corrected.

PR #10084, opened 2024-02-11
chore: Migrate all `packages/create-astro/test` to node:test

shoaibkh4n:
## Changes
Migrates all mocha tests to node:test of `packages/create-astro/test`
Part of ☂️ Move Astro tests to node:test #9873 

## Testing
All tests passed!

## Docs


packages/create-astro/test/project-name.test.js
```diff
@@ -9,25 +9,22 @@ describe('project name', async () => {
 	it('pass in name', async () => {
 		const context = { projectName: '', cwd: './foo/bar/baz', prompt: () => {} };
 		await projectName(context);
-
-		expect(context.cwd).to.eq('./foo/bar/baz');
-		expect(context.projectName).to.eq('baz');
+		assert.strictEqual(context.cwd, './foo/bar/baz');
```
ematipico, line 12: `assert.equal` and `assert.strictEqual` are the same because we import `node:assert/strict`. Can we use `assert.equal` like we are doing in the rest of the code base?

packages/create-astro/test/git.test.js
```diff
@@ -12,21 +13,20 @@ describe('git', () => {
 		const context = { cwd: '', dryRun: true, prompt: () => ({ git: false }) };
 		await git(context);
 
-		expect(fixture.hasMessage('Skipping Git initialization')).to.be.true;
+		assert.ok(fixture.hasMessage('Skipping Git initialization'));
 	});
 
 	it('yes (--dry-run)', async () => {
-		const context = { cwd: '', dryRun: true, prompt: () => ({ git: true }) };
+		const context = { cwd: '', dryRun: true, prompt: () => ({ git: false }) };
```
ematipico, line 20: Why was this changed?

ematipico (changes requested):

packages/create-astro/test/project-name.test.js
shoaibkh4n: hey, yeah you're right, `.equal` will be better as it maintains the consistancy through out the codebase. Give me a minute I'll update this.

packages/create-astro/test/git.test.js
shoaibkh4n: Ohh, Seriously I also dont know 😅😅. I'll update it wait a minute. May be I'm trying some experiment but forgot to roll back.

packages/create-astro/package.json
```diff
@@ -22,7 +22,7 @@
     "build": "astro-scripts build \"src/index.ts\" --bundle && tsc",
     "build:ci": "astro-scripts build \"src/index.ts\" --bundle",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "mocha --exit --timeout 20000 --parallel --ignore **/*.nodetest.js"
+    "test": "astro-scripts test \"test/**/*.test.js\" --parallel"
```
ematipico, line 25:
```suggestion
    "test": "astro-scripts test \"test/**/*.test.js\""
```

shoaibkh4n:
Hey @ematipico , there is a problem with the nodejs test runner, sometimes it just stuck and lead to timeout. Even I made the description async with a low nesting. Any idea Why thou?

ematipico:
> Hey @ematipico , there is a problem with the nodejs test runner, sometimes it just stuck and lead to timeout. Even I made the description async with a low nesting. Any idea Why thou?

No idea, let's see if the concurrency testing allows us to see which test times out and why (I just pushed a commit in your PR)

packages/astro/package.json
```diff
@@ -111,7 +111,7 @@
     "test:match": "mocha --timeout 30000 -g",
     "test:e2e": "playwright test",
     "test:e2e:match": "playwright test -g",
-    "test:node": "astro-scripts test \"test/**/*.nodetest.js\""
+    "test:node": "astro-scripts test \"test/**/*.nodetest.js\" --parallel"
```
ematipico, line 114:
```suggestion
    "test:node": "astro-scripts test \"test/**/*.nodetest.js\""
```

ematipico (approved):

PR #10086, opened 2024-02-12
run "format:imports" script

FredKSchott:
## Changes

- 63 files changed, ~100 imports organized or removed entirely!
- Not sure when the last time we ran this was, but I'm going to create a PR to add this back to CI (in a more efficient way than before) because I find automating this really useful.

## Testing

- Covered by CI

## Docs

- Internal only


FredKSchott:
looks like we rely on that `vite` import that it's removing. If the `format:imports` script is no longer useful/working, we can remove it instead!

Closing since this is not likely to merge

bluwy:
We've also talked about using biome to format the imports, eg https://github.com/withastro/adapters/pull/77, maybe we can also start with something like that before moving more of the formatting to biome. And since it's fast, we could include it as part of the normal format script.

Princesseuh:
Closing this since it's massively outdated and seems like it didn't work anyway. We're currently working toward using Biome in certain parts of the codebase progressively and will eventually achieve what this PR set out to do 🫡

PR #10087, opened 2024-02-12
add back organize imports on a nightly job

FredKSchott:
## Changes

- https://github.com/withastro/astro/pull/10086 seemed to have a big impact
- I know that this CI job was mainly removed due to perf
- This new job only runs once, nightly. Previously, this ran after every push to main.
- Running nightly should solve the performance issue while retaining 99% of the value.

## Testing

- N/A

## Docs

- N/A


FredKSchott:
More discussion needed, see #10086

PR #10094, opened 2024-02-12
Adds types for transition:reload to <a>, <area>, <form>, <script>

martrapp:
## Changes

Related to introduction of transition:reload in https://github.com/withastro/compiler/pull/965

## Testing

manually so far

## Docs

see https://github.com/withastro/docs/pull/6856


packages/astro/astro-jsx.d.ts
```diff
@@ -1,4 +1,7 @@
 /// <reference lib="dom" />
+
+import { AstroReloadAttribute } from './dist/@types/astro.js';
```
bluwy, line 3: I don't think we can import the types this way in this file. You'd need to use `import()` like the below pattern so that the ambient types don't break.

martrapp: Right! But since we go with data-astro- attributes now I'll revert.

PR #10095, opened 2024-02-12
chore: use Biome to sort imports

ematipico:
## Changes

This PR uses Biome to sort imports, and it replaces `organize-imports-cli` to do the job.

```
> root@0.0.0 format:imports /Users/ema/www/withastro/astro
> biome check --apply .

Fixed 5509 file(s) in 1018ms
```

## Testing

CI should pass


matthewp:
The diff is a bit noisy with all of the changes it caused. I assume the only relevant change is the package.json part, right?

PR #10006, opened 2024-02-07
cssesc investigation

lilnasy:
## Changes
- https://github.com/withastro/astro/issues/9880

Issue #9880, opened 2024-01-30
 Issue with cssesc import in Transition.ts causing compatibility problems with postcss-preset-env


nick-noaa:
### Astro Info

```block
Astro                    v4.2.7
Node                     v20.5.1
System                   Linux (x64)
Package Manager          bun
Output                   static
Adapter                  none
Integrations             @astrojs/react
                         @astrojs/svelte
```

### Describe the Bug

Hello,

I've encountered an issue with Astro version 4.2.5 (and possibly later versions) where the removal of the `import cssesc from '/node_modules/cssesc/cssesc.js?v=8e4d546f';` statement in `Transition.ts` is causing compatibility problems with the `postcss-preset-env` package.

As a temporary workaround, I've  downgraded to  Astro 4.2.4.

Could you please look into this issue? Thank you.

### What's the expected result?

Astro package to be compatible with `postcss-preset-env` without requiring manual modifications to the `node_modules` directory, finding an alternative postcss package or downgrading Astro to 4.2.4. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-g3bsbv

### Participation

- [ ] I am willing to submit a pull request for this issue.

natemoo-re:
Could you please share more information about _what_ the compatability issue you're experiencing is? I'm unable to reproduce this and will need more details to help.

You should not have to modify anything inside of `node_modules` for Astro to work with another package. Astro now relies on the `cssesc` package to properly escape the `transition:name` directive.

I'd also try running this on Node and seeing if you have the same problem? Despite Bun's claims of Node compatibility, they are in fact using different runtime heuristics that could cause issues.

dreadjr:
I am seeing a similar error (using node) only on 1 page of my project that uses a client side script to utilize @antv/g6.

Console errors.
```
Uncaught SyntaxError: The requested module '/node_modules/cssesc/cssesc.js?v=e7df1019' does not provide an export named 'default' (at transition.js?v=e7df1019:3:8)
```

If i revert back to previous version of astro, it works again.

nick-noaa:
Yes, same with NPM and PNPM. I am getting the same error message as @dreadjr. Also, a lot of my JS (React, Svelte) isn't working. I believe it's just my components with animations.

pijusz:
Same issue, had to downgrade to 4.2.4.
Using: Transitions, React. React does not load at all.

florian-lefebvre:
[...]

PR #10099, opened 2024-02-13
Re-encode view-transition-names

### LINKED ISSUE #10081: `transition:name` no longer works with special characters after the introduction of cssesc
### Astro Info

```block
Astro                    v4.2.5
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

After the merge of https://github.com/withastro/astro/pull/9822, the usage of special characters in the `transition:name` prop breaks transitions. I use the page title to create a unique transition name on my website that includes whitespaces. After upgrading Astro to v4.2.5, I noticed that this implementation no longer worked.

I have reproduced the problem in a minimal reproducible example. Here are two versions of the same implementation, one working with Astro v4.2.4 and one broken with v4.2.5:
[Working (Astro v4.2.4)](https://stackblitz.com/edit/withastro-astro-eqnptw?file=src%2Fpages%2Findex.astro)
[Broken (Astro v4.2.5)](https://stackblitz.com/edit/withastro-astro-73q1em?file=src%2Fpages%2Findex.astro)

While preparing the examples, I noticed that other special characters like `.` or `/` also break the transitions.

It is possible to use special characters in the `transition:name` prop. If this is not an intended use case, it should be documented in the docs that special characters will not work.

https://stackblitz.com/edit/withastro-astro-73q1em?file=src%2Fpages%2Findex.astro

### Participation

martrapp:
## Changes

We have observed regressions with the cssesc view-transition-name encoding as Chrome seems to have issues with some specification-compliant view transition names. Specifically with backslash-escaped, printable ASCII letters like `\.`, `\ `, `\$` but alos `\30` which would encode a '0'. 

We fix this on the Astro side. Closes #10081 

Instead of removing all problematic characters or replacing them with `_` (as we did in our original solution), this fix re-encodes them in a way that works for cssesc/Chrome, still supports all Unicode characters > 0x7f, and doesn't create collisions for different inputs. 

Problematic characters are replaced by their character code. This is somewhat more difficult to read in the generated CSS, but is more robust. For example `Welcome to Astro.` is now encoded as `Welcome20_to20_Astro2e_`.

 
## Testing

Adapted and extended existing e2e tests
~~(Which only shows that Chrome accepts the encoding but not whether it acts on them. 
We could add animation event listeners to be sure.)~~ **Did that.**

## Docs

bug fix, no implication on documentation of transition:name
/cc @withastro/maintainers-docs for feedback on changeset


.changeset/many-lobsters-accept.md
```diff
@@ -0,0 +1,8 @@
+---
+"astro": minor
```
ematipico, line 2: What's the reason of a `minor`? Should this be a patch?

packages/astro/src/runtime/server/transition.ts
```diff
@@ -45,6 +45,29 @@ const addPairs = (
...
+// Chrome (121) accepts custom-idents for view-transition-names as generated by cssesc,
+// but it just ignores them during view transitions if they contain escaped 7-bit ASCII characters
+// like \<space> or \. A special case are digits and minus at the beginning of the string,
+// which cssesc also encodes as \xx
+const reEncodeValidChars: string[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-_"
+	.split('').reduce((v, c) => (v[c.charCodeAt(0)] = c, v), [] as string[]);
+const reEncodeInValidStart: string[] = "-0123456789"
+	.split('').reduce((v, c) => (v[c.charCodeAt(0)] = c, v), [] as string[]);
+
+function reEncode(s: string) {
```
ematipico, line 57: Can you add some comments that explain the re-encoding logic?

.changeset/many-lobsters-accept.md
martrapp: I always prefer patch level ;-), but here I am unsure whether the fix should be a minor as it might break code in rare cases. Astro users who only use `transition:name` will not notice any negative effects. In niche cases there may be surprises, e.g. if you use `transition:name` in an .astro file and `view-transition-name` CSS property in an .mdx file and both should match. Then it may be that they differ after we have changed the coding scheme. 

As I just wanted to construct an example that will a) differ and has b) worked before this fix, I see that those cases might be really rare. One would be "a_b" which now encodes as "a__b".

What do you think? Should I revert to "patch"?

packages/astro/src/runtime/server/transition.ts
martrapp: Sure, I can't understand why I didn't do that from the start ;-)

.changeset/many-lobsters-accept.md
ematipico: I see! We can go with a `minor` then. Considering your comment, I believe the changelog should highlight the possibility of a "breaking change", meaning that users who were "leveraging the bug" might be impacted.

matthewp: Minor is ok in this case since the release is only 2 days from now. If it were next week I'd rather fix the known regression than worry about unknown corner cases. But 2 days is ok.

matthewp (approved):

martrapp:
Hi @lilnasy, How far have you got with your cssesc investigation? (#10006 / #9880)

.changeset/many-lobsters-accept.md
```diff
@@ -0,0 +1,8 @@
...
+---
+
+Fixes an issue with view transition names containing spaces or punctuation.
+
+This fix could be a breaking change if you leverage details about how Astro translates `transition:name` directives into values of the underlying CSS `view-transition-name` property.
+This mainly affects spaces and punctuation marks but no unicode characters with codes >= 128.
```
sarah11918, line 8:
```suggestion
Regular use naming your transitions with `transition: name` is unaffected.

However, this fix may result in breaking changes if your project relies on the particular character encoding strategy Astro uses to translate `transition:name` directives into values of the underlying CSS `view-transition-name` property. For example `Welcome to Astro` is now encoded as `Welcome20_to20_Astro2e_`.

This mainly affects spaces and punctuation marks but no Unicode characters with codes >= 128.
```

OK, this may be a little much, and of course, may not be entirely accurate. But, would something like this be helpful? It is a minor, and this *might* affect people, so I grabbed some context from the error and PR description in case we liked this.

sarah11918 (commented):
Some thoughts on the changeset here, for your consideration @martrapp ! :raised_hands:

lilnasy:
I didn't find anything that explains why it's being requested by the browser.

.changeset/many-lobsters-accept.md
martrapp: Perfect! Thanks a lot!

ematipico (approved):

Issue #7857, opened 2023-07-28
`walkParentInfos` handles circular imports too naively


ottomated:
### What version of `astro` are you using?

2.9.5

### Are you using an SSR adapter? If so, which one?

N/A

### What package manager are you using?

npm

### What operating system are you using?

Linux

### What browser are you using?

Chrome

### Describe the Bug

When analyzing hoisted script imports, the following situation can happen:

```mermaid
graph TD;
    Component["/src/components/Component.astro"]
    Layout["/src/components/Layout.astro"]
    index["/src/components/index.ts"]
    bad["/src/pages/bad.astro"]
    good["/src/pages/good.astro"]
    Component-->Layout;
    Layout-->index;
    index-->bad;
    index-->good;
```

This causes an issue when `bad.astro` has `import { Component }  from '../components';`. This is because the path from `Component.astro` -> `index.ts` -> `bad.astro` is never traversed, meaning Astro can't analyze the import.

### What's the expected result?

The expected traversal is:

```mermaid
graph TD;
    Component["/src/components/Component.astro"]
    Layout["/src/components/Layout.astro"]
    index["/src/components/index.ts"]
    bad["/src/pages/bad.astro"]
    good["/src/pages/good.astro"]
    Component-->Layout;
    Layout-->index;
    Component-->index;
    index-->bad;
    index-->good;
```

Since walkParentInfo's implementation prevents circular imports by ensuring each module is only visited once, we can't to visit `index.ts` twice like we need to.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xidexe?file=dist%2Fbad%2Findex.html

### Participation

- [X] I am willing to submit a pull request for this issue.

natemoo-re:
Thanks for opening an issue, @ottomated! Is this something you're planning to tackle in https://github.com/withastro/astro/pull/7733 ?

natemoo-re:
> @natemoo-re I think it should be done on that branch, but I don't think I'm the best person for the job – I assume there's an existing algorithm for resolving circular imports that someone else would know.

Cool, thanks for the context! @bluwy hopefully might know something of the technique Vite uses to handle this...

bluwy:
I took a look at the issue in the past few days. I think I got a good grasp of it now, and what proposed here would indeed improve the optimization.

But looking into all the edge case surrounding hoisted scripts, it's a lot complex than expected.

[...]

PR #10102, opened 2024-02-13
Add `experimental.directRenderScript` option

bluwy:
## Changes

Relies on https://github.com/withastro/compiler/pull/969, The PR should be merged and released and updated here before merging.

Closes https://github.com/withastro/astro/issues/7857 (Styles will be tackled separately)

This introduces `experimental.directRenderScript` and removes `experimental.optimizeHoistedScript`. Both work differently but should achieve the same result. When directly rendering scripts, it will no longer be hoisted and will be rendered where it's declared. The compiled code will call `${renderScript($$result, "/src/pages/...")}` to render the script.

---

This option is introduced due to the current caveats of our script handling. Today, it works by starting from each Astro pages, and crawling the module graph to find the imported Astro components's scripts, and include it to the Astro page. However, this is not ideal in many cases if Astro components are only used conditionally, or through barrel files, resulting in scripts added to a page when it shouldn't.

This new option prevents the caveat by only including the scripts when the components are actually rendered. That way it won't be possible for scripts to be included in a page that it shouldn't.

## Testing

Lend the tests from the `experimental.optimizeHoistedScript` option. I also tested enabling the option by default, and it seems to mostly work besides the known caveats,

## Docs

Added a changeset and jsdoc. Needs a review on this!


packages/astro/package.json
MoustaphaDev, line 1: I noticed that estree isn't used anymore in astro so we can probably remove it from dev dependencies

packages/astro/src/core/build/plugins/plugin-analyzer.ts
```diff
@@ -317,6 +184,16 @@ export function vitePluginAnalyzer(
...
+
+				// When directly rendering scripts, we don't need to group them together when bundling,
+				// each script module is its own entrypoint, so we directly assign each script modules to
+				// `discoveredScripts` here, which will eventually be passed as inputs of the client build.
+				if (options.settings.config.experimental.directRenderScript && astro.scripts.length) {
+					for (let i = 0; i < astro.scripts.length; i++) {
+						const hid = `${id.replace('/@fs', '')}?astro&type=script&index=${i}&lang.ts`;
+						internals.discoveredScripts.add(hid);
+					}
+				}
```
MoustaphaDev, line 196: I wonder if we could reuse the values in [`hoistedScripts`](https://github.com/withastro/astro/pull/10102/files#diff-2761aafa73765fcf9ead8be801bbe5eb6da3cb305c7b14917299c9fab004e6feR41-R45) by returning it in the `scan` function, looks like they would be identical.

packages/astro/test/hoisted-imports.test.js
MoustaphaDev, line 1: It would be great to have a test for the [script inlining behavior](https://github.com/withastro/astro/pull/10102/files#diff-e741b21916fb6d560d1d19eb436ec16e87cb59065d6f80dc8499701c4712bb48R8-R37) if possible

.changeset/cool-jobs-fetch.md
MoustaphaDev, line 1: It's probably worth mentioning that this feature also enables conditional rendering of processed scripts, that wasn't possible before

packages/astro/src/runtime/server/render/script.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export async function renderScript(result: SSRResult, id: string) {
+	const inlined = result.inlinedScripts.get(id);
+	if (inlined) {
+		return markHTMLString(`<script type="module">${inlined}</script>`);
+	}
+
+	const resolved = await result.resolve(id);
+	return markHTMLString(`<script type="module" src="${resolved}"></script>`);
+}
```
lilnasy, line 16: Would it be possible to avoid duplicates by keeping track of already added scripts on `result`?

bluwy: Oh you're right! I actually wanted to implement this and totally forgot.

MoustaphaDev (approved):
LGTM, excited for this!

ematipico:
We should update the PR, and replace the compiler dependency with the correct one

bluwy:
!preview direct-render-script

ematipico (approved):

.changeset/cool-jobs-fetch.md
```diff
@@ -0,0 +1,11 @@
+---
+"astro": minor
+---
+
+Adds a new `experimental.directRenderScript` option.
+
+This option replaces the `experimental.optimizeHoistedScript` option to use a more reliable strategy to prevent scripts being executed in pages they are not used. The scripts are now directly rendered as declared in Astro files (with features like TypeScript, importing `node_modules`, and deduplicating scripts still working), and should result in scripts running in the correct pages compared to the previous static analysis approach. You can also now conditionally render scripts in your Astro file.
+
+However, as scripts are now directly rendered, they are no longer hoisted to the `<head>` and multiple scripts on a page are no longer bundled together. If you enable this option, you should check if it affects your site's behaviour.
+
```
sarah11918, line 10:
``````suggestion
Adds a new `experimental.directRenderScript` configuration option which provides a more reliable strategy to prevent scripts from being executed in pages where they are not used. 

This replaces the `experimental.optimizeHoistedScript` flag introduced in v2.10.4 to prevent unused components' scripts from being included in a page unexpectedly. That experimental option no longer exists and must be removed from your configuration, whether or not you enable `directRenderScript`:

```diff
// astro.config.mjs
import { defineConfig } from 'astro/config';

export default defineConfig({
	experimental: {
-		optimizeHoistedScript: true,
+		directRenderScript: true
	}
});
```

With `experimental.directRenderScript` configured, scripts are now directly rendered as declared in Astro files (including existing features like TypeScript, importing `node_modules`, and deduplicating scripts). You can also now conditionally render scripts in your Astro file.

However, this means scripts are no longer hoisted to the `<head>` and multiple scripts on a page are no longer bundled together. If you enable this option, you should check that all your `<script>` tags behave as expected.
``````
Just a suggestion mostly to show removing the old config, and adding the new. Please update as you see fit to make this accurate!

File: packages/astro/src/@types/astro.ts [hunk 4]
```diff
@@ -1601,25 +1601,29 @@ export interface AstroUserConfig {
...
 		 * @type {boolean}
 		 * @default `false`
-		 * @version 2.10.4
+		 * @version 4.5.0
 		 * @description
-		 * Prevents unused components' scripts from being included in a page unexpectedly.
-		 * The optimization is best-effort and may inversely miss including the used scripts. Make sure to double-check your built pages
-		 * before publishing.
-		 * Enable hoisted script analysis optimization by adding the experimental flag:
+		 * Directly render scripts as declared in Astro files (with features like TypeScript, importing `node_modules`, and deduplicating
```
sarah11918, line 1609: Will review this after the changeset is final, just in case any wording used there makes you want to update something here first!

sarah11918 (commented):
Thanks, @bluwy ! Took a pass at the changeset, mostly because I wanted to make sure it explicitly addressed removing the old option (whether or not they added the new one) since it's deleted from the docs (and, I'm assuming from the codebase? But could be wrong about that! Even if not, just as well to tell people the option is deleted and force them on to the new one!  :smile: )

When we have the wording from the changeset finalized, I will take a look at the JSdoc again, just for consistency in case anything changes!  :raised_hands:

sarah11918 (approved):
Looks good to me, @bluwy ! :raised_hands:

PR #10104, opened 2024-02-13
fix(markdown): don’t generate mdast html nodes

### LINKED ISSUE #9909: Astro wrongly process markdown code
### Astro Info

```block
❯ astro info
Re-optimizing dependencies because vite config has changed
Astro                    v4.2.7
Adapter                  none
Integrations             astro-expressive-code
```

_No response_

### Describe the Bug

It looks like Astro preprocesses the [hast](https://github.com/syntax-tree/hast) in a way that’s incompatible with this plugin (other rehype plugins). [rehype-mermaidjs](https://github.com/remcohaszing/rehype-mermaidjs) searches for `<pre>` and `<code>` elements, but Astro turns them into raw nodes. This is a bug in Astro. I suggest you open an issue there. Maybe you can work around it using [rehype-raw](https://github.com/rehypejs/rehype-raw).

This bug was initially coded as https://github.com/remcohaszing/remark-mermaidjs/issues/23 and later identified in https://github.com/remcohaszing/rehype-mermaid/issues/10

(Citing @remcohaszing )

The processed hast should allow for a further processing by any `rehype` plugin.

https://stackblitz.com/edit/withastro-astro-ekqekg

### Participation

remcohaszing:
## Changes

`html` nodes from mdast are converted to `raw` hast nodes. These nodes are then not processed by proper rehype plugins. Typically if a remark plugin generates `html` nodes, this indicates it should have actually been a rehype plugin.

This changes the remark plugins that generate `html` nodes into rehype nodes. These were `remarkPrism` and `remarkShiki`.

## Testing

```sh
pnpm build -- --force
pnpm test
```

No tests were added, because it mostly changes internals.

## Docs

This affects a user’s behaviour by adding better compatibility with some existing remark / rehype plugins. E.g. [`rehype-mermaid`](https://github.com/remcohaszing/rehype-mermaid) will work. I don’t think a docs update is needed.

Closes #9909


packages/integrations/mdx/src/plugins.ts
```diff
@@ -79,18 +64,25 @@ function getRehypePlugins(mdxOptions: MdxOptions): PluggableList {
...
-		rehypeApplyFrontmatterExport,
-	];
+		rehypePlugins.push(rehypeHeadingIds, rehypeInjectHeadingsExport);
+
+		// Apply syntax highlighters after user plugins to match `markdown/remark` behavior
+		if (mdxOptions.syntaxHighlight === 'shiki') {
+			rehypePlugins.push([rehypeShiki, mdxOptions.shikiConfig]);
+		} else if (mdxOptions.syntaxHighlight === 'prism') {
+			rehypePlugins.push(rehypePrism);
+		}
```
remcohaszing, line 81: You could change the order here. I.e. move these plugins before or after user provided rehype plugins.

packages/markdown/remark/src/index.ts
```diff
@@ -103,6 +96,15 @@ export async function createMarkdownProcessor(
 		...remarkRehypeOptions,
 	});
 
+	if (!isPerformanceBenchmark) {
+		// Syntax highlighting
+		if (syntaxHighlight === 'shiki') {
+			parser.use(rehypeShiki, shikiConfig);
+		} else if (syntaxHighlight === 'prism') {
+			parser.use(rehypePrism);
+		}
```
remcohaszing, line 105: You could change the order here. I.e. move these plugins before or after user provided rehype plugins.

packages/markdown/remark/package.json
```diff
@@ -46,7 +48,9 @@
     "remark-smartypants": "^2.0.0",
     "shikiji": "^0.9.19",
     "unified": "^11.0.4",
+    "unist-util-remove-position": "^5.0.0",
     "unist-util-visit": "^5.0.0",
+    "unist-util-visit-parents": "^6.0.0",
```
remcohaszing, line 53: These seem like new dependencies, but they were already indirect dependencies.

bluwy:
The code looks good to me, but isn't this a breaking change as the order is changed, and code is transformed differently? Or do you foresee that it shouldn't break much?

remcohaszing:
I see why this can be considered a breaking change. On the other hand, it only really breaks if people rely on the bad AST that Astro produced earlier. In that case I would say it breaks people’s workarounds.

I don’t expect this to break people’s code, but I’m not sure how people are using this. I do know this **fixes** compatibility with `rehype-mermaid` and potentially other rehype plugins.

Princesseuh (approved):
I have the same concerns as Bjorn, but I do think it's okay!

ematipico:
As middle ground, we could release these changes as `minor` and add a warning for a possible breaking change if integrations were relying on broken AST. What do you think @bluwy  and @Princesseuh ?

Princesseuh:
> As middle ground, we could release these changes as `minor` and add a warning for a possible breaking change if integrations were relying on broken AST.

Fine with me! We'll mention it in the article.

.changeset/thirty-beds-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@astrojs/mdx": patch
+"@astrojs/markdown-remark": patch
```
ematipico, line 3:
```suggestion
"@astrojs/mdx": minor
"@astrojs/markdown-remark": minor
```

remcohaszing:
Don't forget about this comment: https://github.com/withastro/astro/pull/10104#discussion_r1488274899. Whatever you choose is fine, just make it explicit :)

bluwy:
Marking as minor is fine by me too 👍 Let's released it in 4.5 then.

packages/integrations/mdx/src/plugins.ts
bluwy: Thanks for flagging this (and the other comment as well). I suppose if we move it before the user-provided rehype plugins, it would be safer compat-wise as the previous remark plugin was added last. I think I'll lean on moving it before then.

packages/markdown/remark/src/index.ts
bluwy: It looks like this code is already moved before user-defined rehype plugins, but the [other comment](https://github.com/withastro/astro/pull/10104#discussion_r1488274899) was after.

OliverSpeir:
Having both before user rehype plugins I think is important otherwise great change I know rehype mermaid has been requested

ematipico:
It would be great if we could craft a good changelog explaining the possible breaking change. @remcohaszing would you mind doing that? cc @bluwy might help, he has more context on the matter.

remcohaszing:
Do you mean that as part of the changeset?

ematipico:
Yeah sorry. That's what I meant

packages/markdown/remark/src/index.ts
```diff
@@ -24,8 +24,8 @@ import { rehypeImages } from './rehype-images.js';
 export { InvalidAstroDataError, setVfileFrontmatter } from './frontmatter-injection.js';
 export { rehypeHeadingIds } from './rehype-collect-headings.js';
 export { remarkCollectImages } from './remark-collect-images.js';
-export { remarkPrism } from './remark-prism.js';
-export { remarkShiki } from './remark-shiki.js';
```
bluwy, line 28: I think we still need to preserve exporting `remarkPrism` and `remarkShiki` here to prevent a breaking change, and the potential where an older version of Astro or `@astrojs/mdx` uses this newer package version (somehow).


We can add a TODO to remove it in the next major of this package.

packages/integrations/mdx/src/plugins.ts
```diff
@@ -79,18 +64,27 @@ function getRehypePlugins(mdxOptions: MdxOptions): PluggableList {
...
+	// computed from `astro.data.frontmatter` in VFile data
+	rehypePlugins.push(rehypeApplyFrontmatterExport);
+
+	if (!isPerformanceBenchmark) {
+		// Apply syntax highlighters after user plugins to match `markdown/remark` behavior
+		if (mdxOptions.syntaxHighlight === 'shiki') {
+			rehypePlugins.push([rehypeShiki, mdxOptions.shikiConfig]);
+		} else if (mdxOptions.syntaxHighlight === 'prism') {
+			rehypePlugins.push(rehypePrism);
+		}
```
bluwy, line 86: IIUC this still pushes `rehypeShiki` and `rehypePrism` after the user rehype plugins? I think we should be putting it before.

I noticed there were some CI fails though while you were working on this, was it failing when moving it before the user rehype plugins?

remcohaszing: Yes, my bad. It was failing, because it shiki was throwing for `math` code blocks. Those are now skipped.

bluwy (approved):
Thanks for following up with the requested changes! Appreciate the refactor and cleanup.

.changeset/thirty-beds-smoke.md
```diff
@@ -0,0 +1,8 @@
+---
+"@astrojs/mdx": minor
+"@astrojs/markdown-remark": minor
+---
+
+Handle syntax highlighting using rehype plugins instead of remark plugins. This provides better interoperability with other [rehype plugins](https://github.com/rehypejs/rehype/blob/main/doc/plugins.md#list-of-plugins) that deal with code blocks, in particular with third party syntax highlighting plugins and [`rehype-mermaid`](https://github.com/remcohaszing/rehype-mermaid).
```
sarah11918, line 6:
```suggestion
Changes Astro's internal syntax highlighting to use rehype plugins instead of remark plugins. This provides better interoperability with other [rehype plugins](https://github.com/rehypejs/rehype/blob/main/doc/plugins.md#list-of-plugins) that deal with code blocks, in particular with third party syntax highlighting plugins and [`rehype-mermaid`](https://github.com/remcohaszing/rehype-mermaid).
```

Given that this is a *change to an existing thing*, I like phrasing this so it makes it clear upfront that something has changed! (and especially given that below we're going to say your project might actually break)

Make sure the nuance is correct here, and correct as necessary, but I like the "in your face-ness" of "We changed how Astro's syntax highlighting works!"

.changeset/thirty-beds-smoke.md
```diff
@@ -0,0 +1,8 @@
...
+
+This may break your code if you are using either a remark plugin that relies on nodes of type `html`, or a rehype plugin that depends on nodes of type `raw`. If you are, you should consider using a rehype plugin that deals with the generated `element` nodes instead. You can use [`hast-util-to-html`](https://github.com/syntax-tree/hast-util-to-html) to get a string from a `raw` node, but better is to transform the AST instead of raw HTML strings.
```
sarah11918, line 8:
```suggestion
This may be a breaking change if your are currently using:
- a remark plugin that relies on nodes of type `html`
- a rehype plugin that depends on nodes of type `raw`. 

Please review your rendered code samples carefully, and if necessary, consider using a rehype plugin that deals with the generated `element` nodes instead. You can transform the AST of raw HTML strings, or alternatively use [`hast-util-to-html`](https://github.com/syntax-tree/hast-util-to-html) to get a string from a `raw` node.
```
Just a suggestion because clarity is always good when someone's project code might break because of a change!

So I've called attention to what the problem could be and put the desired solution first in the list (rather than "do this... or better, do this).

Please correct if this is now inaccurate because of moving words around, but I think something that is visually easier to see the breaking change is helpful here!

sarah11918 (commented):
Thank you for this contribution @remcohaszing -- seems like this will ensure some greater compatibility, and we at Astro love our code samples!

I've made some small suggestions below because this is not only a change, but potentially a *breaking* change, and this changeset message is a great way to call attention to that and provide the most warning, guidance and support to people who might need to make a change because of this.

See what you think and correct to be accurate if I've lost any of the particular meaning by editing! :raised_hands:

.changeset/thirty-beds-smoke.md
```diff
@@ -0,0 +1,12 @@
+---
+"@astrojs/mdx": minor
+"@astrojs/markdown-remark": minor
+---
+
+Changes Astro's internal syntax highlighting to use rehype plugins instead of remark plugins. This provides better interoperability with other [rehype plugins](https://github.com/rehypejs/rehype/blob/main/doc/plugins.md#list-of-plugins) that deal with code blocks, in particular with third party syntax highlighting plugins and [`rehype-mermaid`](https://github.com/remcohaszing/rehype-mermaid).
+
+This may be a breaking change if your are currently using:
```
sarah11918, line 8:
```suggestion
This may be a breaking change if you are currently using:
```

sarah11918: tiny typo fix!

PR #10105, opened 2024-02-13
fix(cli): load polyfill before loading config during `astro add`

negativems:
I tried to import Vue Astro adapter and I got an error. The error is the next one:

image

And I tried to import any other adapter and I get the same error when using the Vercel serverless adapter.

After seeing the problem, I added the node crypto import into the Vercel adapter and the error gone:

image

Edit:

This is my astro.config.mjs

image

### DIFF
--- a/.changeset/itchy-monkeys-fail.md
+++ b/.changeset/itchy-monkeys-fail.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes an issue where some astro commands failed if an integration used the global `crypto` object.
+Fixes an issue where some astro commands failed if the astro config file or an integration used the global `crypto` object.
--- a/.changeset/itchy-monkeys-fail.md
+++ b/.changeset/itchy-monkeys-fail.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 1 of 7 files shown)


florian-lefebvre (commented):
Must have been introduced with the edge middleware verification, good catch! Can you please add a changeset?

ematipico:
This is weird, though, we have a polyfill for crypto: 

https://github.com/withastro/astro/blob/e24db1d8a683897d4c179f3511f8e38950884e7f/packages/astro/src/core/polyfill.ts#L6-L10

ematipico:
@negativems, what's your local version of Node.js? What's the version of the runtime in Vercel?

ematipico:
@negativems  can you try this version of the vercel adapter, and see if that fixes your issue?

`@astrojs/vercel@experimental--vercel-crypto-fix`

lilnasy:
!preview node-crypto-error

lilnasy (approved):
Thanks for taking this on!

bluwy (approved):

bluwy (approved):

ematipico (approved):

PR #10120, opened 2024-02-15
Update to Vite 5.1

bluwy:
## Changes

Update to Vite 5.1. Also handle the type changes for `vite.build.assetsInlineLimit` that now accepts a function. Other than that, it's just a version bump.

Vite 5.1 has other features that should trim down our code too that I'll send as separate PRs later.

## Testing

Existing tests should pass.

## Docs

n/a. version upgrade.

### DIFF
--- a/packages/astro/src/core/build/plugins/util.ts
+++ b/packages/astro/src/core/build/plugins/util.ts
@@ -1,5 +1,5 @@
 import { extname } from 'node:path';
-import type { Rollup, Plugin as VitePlugin } from 'vite';
+import type { BuildOptions, Rollup, Plugin as VitePlugin } from 'vite';
 
 // eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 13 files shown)


packages/astro/src/core/build/plugins/util.ts
```diff
@@ -69,3 +69,20 @@ export function getPathFromVirtualModulePageName(virtualModulePrefix: string, id
...
+	if (typeof assetsInlineLimit === 'number') {
+		return Buffer.byteLength(assetContent) < assetsInlineLimit;
+	}
+
+	const result = assetsInlineLimit(assetPath, Buffer.from(assetContent));
+	if (result != null) {
+		return result;
+	}
+
+	return Buffer.byteLength(assetContent) < 4096; // Fallback to 4096kb by default (same as Vite)
```
ematipico, line 87: It would be nice if Vite would expose this default. Something to think about :)

ematipico (approved):

PR #10125, opened 2024-02-15
chore: move test to node

ktym4a:
## Changes

This PR moves the files that start with `g*` and `h*` to node:test

Part of https://github.com/withastro/astro/issues/9873

## Testing

should pass.

### DIFF
--- a/packages/astro/test/get-static-paths-pages.nodetest.js
+++ b/packages/astro/test/get-static-paths-pages.nodetest.js
@@ -17,12 +17,12 @@ describe('getStaticPaths with trailingSlash: ignore', () => {
 	it('includes index page', async () => {
 		let html = await fixture.readFile('/index.html');
 		let $ = cheerio.load(html);
-		assert.strictEqual($('h1').text(), 'Page 1');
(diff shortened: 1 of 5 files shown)


packages/astro/test/get-static-paths-pages.nodetest.js
```diff
@@ -16,12 +17,12 @@ describe('getStaticPaths with trailingSlash: ignore', () => {
 	it('includes index page', async () => {
 		let html = await fixture.readFile('/index.html');
 		let $ = cheerio.load(html);
-		expect($('h1').text()).to.equal('Page 1');
+		assert.strictEqual($('h1').text(), 'Page 1');
```
ematipico, line 20: We do use `assert.equal` in all tests, that's because we already import `assert` from `node:assert/strict`. This means `assert.equal` is `assert.strictEqual`.

Would you mind changing that?

ematipico: Yes. If there are other instances of `strictEqual`, we can do that in another PR

ematipico (approved):

PR #10136, opened 2024-02-15
Allow islands to be re-rendered with new props on page transition

matthewp:
https://github.com/withastro/astro/assets/361671/e1ae6e4d-ac20-4dab-a840-6c2c00e536d2

## Changes

- Copies over an islands props that are preserved via `transition:persist
- Fixes the React renderer so it doesn't create a new root when one exists, allowing state to be preserved.

## Testing

- Added a test case

## Docs

- https://github.com/withastro/docs/pull/6915


packages/astro/e2e/view-transitions.test.js
```diff
@@ -782,7 +786,7 @@ test.describe('View Transitions', () => {
 	});
 
 	test('client:only styles are retained on transition (1/2)', async ({ page, astro }) => {
-		const totalExpectedStyles = 9;
+		const totalExpectedStyles = 8;
```
matthewp, line 789: Need to confirm this is correct, just want to see if the rest of CI passes with this change first.

packages/astro/e2e/view-transitions.test.js
```diff
@@ -801,7 +805,7 @@ test.describe('View Transitions', () => {
 	});
 
 	test('client:only styles are retained on transition (2/2)', async ({ page, astro }) => {
-		const totalExpectedStyles_page_three = 11;
+		const totalExpectedStyles_page_three = 10;
```
matthewp, line 808: Same as above

ematipico (approved):

bluwy (approved):

martrapp:
Wow, that looks really cool! Preserved state and updated props!

Three questions came to my mind:
- If we update the properties of persisted island with the values of the new page, shouldn't we offer something similar for other persisted (non-island) elements, e.g. have a way to update their attributes with the values from the new page?
- Do we need a way to opt-out (or explicitly opt-in)?
- Have you had a chance to review the other UI frameworks? Are they robust to hydration errors?

martrapp:
Wow @matthewp, that looks really cool! Preserved state and updated props!

Three questions came to my mind:
- If we update the properties of persisted island with the values of the new page, shouldn't we offer something similar for other persisted (non-island) elements, e.g. have a way to update their attributes with the values from the new page?
- Do we need a way to opt-out (or explicitly opt-in)?
- Have you had a chance to review the other UI frameworks? Are they robust to hydration errors?

matthewp:
@martrapp I think for non-islands it's a bit trickier to determine what to do. For example, on a `<video>` you would not want to just copy over the `src` attribute because even if it is the same value, it would trigger the video to replay. So you'd have to do "set if not the same". And even then, what is the intent exactly? If two videos are different and you use persist, is the intent to change the video or keep the current one? I'm not sure.

This is true of islands to, of course. My assumption here is that islands are more adept to respond to the changes and keep the current state if that's the right thing to do. So I do think we should do a preview release here and have some people use it to see if it matches their expectations or not. We could provide an option in / out directive in either case.

martrapp:
Bad example :), but i know what you mean
> Dynamically modifying a [source](https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element) element's [src](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-source-src) or [type](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-source-type) attribute when the element is already inserted in a [video](https://html.spec.whatwg.org/multipage/media.html#the-video-element) or [audio](https://html.spec.whatwg.org/multipage/media.html#the-audio-element) element will have no effect.

A nice example for non-islands would be updating the class attribute as transition:persist copies over the element with the current class attribute, but not the CSS referenced from the class attribute. Updating the class attribute with the values from the next page would guarantee that the referenced classes are found in the style sheets of the next page.

Of course, this can all be done using lifecycle events and no need to include it in the core implementation.

But there may be a point to give users some control over whether their island props being updated automatically ;-)

matthewp:
@martrapp What about something like:

- `transition:persist-props` for islands. This defaults to `false` and props are updated. You can add this attribute to keep the props.
- `transition:persist-attrs` for non-islands (any regular element). This defaults to `true`. If you want to swap attributes use `transition:persist-attrs="false"`

Not sold on this tbh, just getting the conversation moving.

martrapp:
Hey @matthewp 
> What about something like ...

Yes, great! 

* transition:persist-props: This allows users to correct the default if it doesn't suit them. 
* transition:persist-attrs: We can do this, but we don't have to. We haven't had any requests for updating attributes from outside so far. I just thought it made sense to discuss it when we were looking at the properties of islands.

matthewp:
@martrapp Ok, I'll update the PR to implement `transition:persist-props`. I'll like the attr version alone for now, if we get requests it should be easy enough to add.

sarah11918 (commented):
Hi @matthewp ! Following the conversation in this PR, I also wanted to add mention here of `transition:persist-props` as I think that's important to call out!

Also, the PR title says "islands" which we've always taken to mean a UI framework component. (e.g. not an Astro component) but doesn't mean our audience is necessarily going to have that distinction top of mind. 

This changeset description uses "island" and "component" interchangeably, and no where screams "YOU CAN'T USE THIS ON AN ASTRO COMPONENT". I didn't *add* anything to that effect, but it could be worth a stronger, explicit mention somewhere if that's the case. It can be as simple as updating a couple of the times where the word "component" is used to "framework component" for example. So, you can decide whether you think that's necessary.

So, see what you think of my suggestions for your consideration!

sarah11918 (approved):
Approving for docs, so it's clear!

Fakerko:
I think it works differently now than described in the documentation.

**transition:persist-props**
---

The documentation says:

This allows you to control whether or not an island’s props should be persisted upon navigation.

By default, when you add transition:persist to an island, the state is retained upon navigation, but your component will re-render with new props. This is useful, for example, when a component receives page-specific props such as the current page’s title.

You can override this behavior with transition:persist-props. Adding this directive will keep an island’s existing props (not re-render with new values) in addition to maintaining its existing state.

---

Now it works exactly the other way around. When I have transition:persist, all props are preserved, but when I use transition:persist-props, they are reset when the page changes.

martrapp:
Hi @Fakerko there might be a miss-understanding:

For a navigation from "from" to "to",
use `transition:persist` + `transition:name` (or short `transition:persist="name"`) on the "from" page (and the same properties on the "to" page) to get the properties reevaluated on the "to" page.
Use `transition:persist` + `transition:name` (or short `transition:persist="name"`) + `transition:persist-props` on the "from" page (and transition:persist` + `transition:name` (or short `transition:persist="name"`) on the "to" page) to prevent reevaluation of properties on the "to" page.
  
Do you have a counterexample that I should have a look at?

Fakerko:
Hi @martrapp, I made an example on [Codesandbox](https://codesandbox.io/p/devbox/transition-persist-islands-v4tc9f?file=%2Fsrc%2Fpages%2Findex.astro%3A12%2C29&layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522cly8w60gh00073b6k15lrv4uu%2522%252C%2522sizes%2522%253A%255B100%252C0%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522cly8w60gh00023b6k345i872k%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522cly8w60gh00043b6krdft7jaw%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522cly8w60gh00063b6kgpnrbml6%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522cly8w60gh00023b6k345i872k%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522cly8w60gg00013b6k2d649auy%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252FREADME.md%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522id%2522%253A%2522cly8whjtc00023b6j0400tn6f%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A12%252C%2522startColumn%2522%253A29%252C%2522endLineNumber%2522%253A12%252C%2522endColumn%2522%253A29%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252Fpages%252Findex.astro%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522id%2522%253A%2522cly8wm3gi00023b6jlasy0ibs%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A21%252C%2522startColumn%2522%253A22%252C%2522endLineNumber%2522%253A21%252C%2522endColumn%2522%253A22%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252Fcomponents%252FTest.tsx%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%252C%2522id%2522%253A%2522cly8w60gh00023b6k345i872k%2522%252C%2522activeTabId%2522%253A%2522cly8whjtc00023b6j0400tn6f%2522%257D%252C%2522cly8w60gh00063b6kgpnrbml6%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522cly8w60gh00053b6khg24dy3u%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TASK_PORT%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522port%2522%253A4321%252C%2522path%2522%253A%2522%252F%2522%257D%255D%252C%2522id%2522%253A%2522cly8w60gh00063b6kgpnrbml6%2522%252C%2522activeTabId%2522%253A%2522cly8w60gh00053b6khg24dy3u%2522%257D%252C%2522cly8w60gh00043b6krdft7jaw%2522%253A%257B%2522id%2522%253A%2522cly8w60gh00043b6krdft7jaw%2522%252C%2522activeTabId%2522%253A%2522cly8w60gh00033b6kokaisubq%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522cly8w60gh00033b6kokaisubq%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TASK_LOG%2522%252C%2522taskId%2522%253A%2522dev%2522%257D%252C%257B%2522id%2522%253A%2522cly8w8b9q005p3b6ktkoqgdfu%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TERMINAL%2522%252C%2522shellId%2522%253A%2522cly8w8bc3000od8hi17ba7afs%2522%257D%255D%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Afalse%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D).

The first component uses transition:persist and the second uses transition:persist-props.

Once you click on "Update values" in the components and go to the "About" page, for the component with transition:persist the component values are preserved, but for the component with transition:persist-props they are reset.

martrapp:
Yes.

- The first component does not refresh props because the Test component does not use any. (The signature is `Test = (props: any)` but it does not render any props.) `Test1` ... `test3` is state, and as the docs say: "the state is retained upon navigation".
- The second component does not persist anything because there is no `transition:persist` directive.

Fakerko:
Oh, I get it. So it's working as it should, I'm sorry. 

My understanding was that transition:persist-props preserves the component values, and transition:persist preserves the component but without the values.

Thanks a lot!

martrapp:
Let's see if we can make that more clear in the docs. ;-)

Fakerko:
Maybe add an example at the end of transition:persist-props block, something like this:

**Component will be re-rendered with new props:**
`<MyComponent myProp="someProp" transition:name="someName" transition:persist />`

**Component will keep the existing props:**
`<MyComponent myProp="someProp" transition:name="someName" transition:persist-props />`

martrapp:
At least we have to better point out, that `transition:persist-props` does not work without `transition:persist`!

Thank you very much for your feedback. Obviously this is complicated.

PR #10138, opened 2024-02-16
Update Vite ws property to hot

bluwy:
## Changes

Migrate from `ws` to `hot`, which provides the same interface, but allows signalling all HMR clients (not just the browser but also SSR) about certain events. `ws` was deprecated in Vite 5.1.

This unblocks one of the hurdles when using the new Vite Runtime API. cc @alexanderniebuhr 

## Testing

Existing tests should pass.

## Docs

n/a. Didn't add a changeset as there shouldn't be any visible changes.

### DIFF
--- a/.changeset/flat-rabbits-pump.md
+++ b/.changeset/flat-rabbits-pump.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Updates Vite calls to use `hot` instead of deprecated `ws`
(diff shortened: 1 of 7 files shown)


bluwy:
@florian-lefebvre Ive noted in the PR description for why I didn't add a changeset. I don't think we need one.

florian-lefebvre:
My bad! I didn't see it 😅

Issue #9859, opened 2024-01-29
Exclude server endpoints from Vite's dependency optimization


pilcrowOnPaper:
### Astro Info

```block
Astro                    v4.2.6
Node                     v20.2.0
System                   macOS (x64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/node
Integrations             none
```

### Describe the Bug

With Astro's current configuration, Vite optimizes all dependencies imported in `src/**`. That causes issues when you import packages that use `.node` files such as `@node-rs/argon2`:

```
✘ [ERROR] No loader is configured for ".node" files: node_modules/.pnpm/@node-rs+argon2-darwin-x64@1.7.0/node_modules/@node-rs/argon2-darwin-x64/argon2.darwin-x64.node

    node_modules/.pnpm/@node-rs+argon2@1.7.0/node_modules/@node-rs/argon2/index.js:159:36:
      159 │             nativeBinding = require('@node-rs/argon2-darwin-x64')
          ╵                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] No loader is configured for ".node" files: node_modules/.pnpm/@node-rs+bcrypt-darwin-x64@1.9.0/node_modules/@node-rs/bcrypt-darwin-x64/bcrypt.darwin-x64.node

    node_modules/.pnpm/@node-rs+bcrypt@1.9.0/node_modules/@node-rs/bcrypt/binding.js:153:36:
      153 │             nativeBinding = require('@node-rs/bcrypt-darwin-x64')
```

We should probably exclude at least server endpoints, which we know won't be used in the client, from Vite's optimization. This should lead to better performance as well.

### What's the expected result?

Server only modules won't be optimized.

### Link to Minimal Reproducible Example

https://github.com/pilcrowOnPaper/astro-deps-optimization-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

pilcrowOnPaper:
In the future, I'd love to see a way to mark modules as server-only, like `.server.ts` in SvelteKit

pixelmund:
@pilcrowOnPaper 

Currently running into some issues using oslo with astro & cloudflare pages
``` bash
[ERROR] Could not resolve "fs"

    node_modules/.pnpm/@node-rs+argon2@1.7.0/node_modules/@node-rs/argon2/index.js:5:45:
      5 │ const { existsSync, readFileSync } = require('fs')
        ╵                                              ~~~~

[...]

PR #10143, opened 2024-02-16
Improve `optimizeDeps.entries` to avoid server endpoints

bluwy:
## Changes

close https://github.com/withastro/astro/issues/9859

Previously, `optimizeDeps.entries` set does not respect the `srcDir` option and globs into server endpoints, which we know should not contain browser code. `optimizeDeps` should only be used on browser code.

This PR fixes by respecting `srcDir`, and make sure server endpoints are not included in the glob. While this is not perfect in the bigger picture, and the ideal solution is to add a custom esbuild plugin to transform Astro files to allow the scanner to scan the appropriate browser files, this PR is an easier and quicker fix to the issue for now.

## Testing

It's a bit hard to test the overall behaviour, but I ran `DEBUG="vite:deps" pnpm dev` in the example projects to check that it logs the correct files globbed by `optimizeDeps.entries`.

## Docs

n/a. bug fix.


packages/astro/src/core/create-vite.ts
```diff
@@ -112,7 +115,13 @@ export async function createVite(
 		customLogger: createViteLogger(logger, settings.config.vite.logLevel),
 		appType: 'custom',
 		optimizeDeps: {
-			entries: ['src/**/*'],
+			// Scan all files within `srcDir` except for known server-code (e.g endpoints)
+			entries: [
+				`${srcDirPattern}!(pages)/**/*`, // All files except for pages
```
bluwy, line 120: I made some tests and it seems that `srcDirPattern` will always have a trailing slash. The only case I'm not sure is on Windows.

packages/astro/src/core/create-vite.ts
```diff
@@ -112,7 +115,13 @@ export async function createVite(
...
+				`${srcDirPattern}pages/**/!(*.js|*.ts)`, // All pages except for endpoints
```
ematipico, line 121: Ehehe, unfortunately `mjs` and `cjs` should be included, even though it's a low chance that users use them. Still, we should

ematipico (approved):

packages/astro/src/core/create-vite.ts
bluwy: Good catch! I'll add this too

ematipico (approved):
Looks good!

PR #10145, opened 2024-02-16
feat: experimental content collection JSON schemas

alexanderniebuhr:
## Changes

- initial implementation of https://github.com/withastro/roadmap/discussions/696
- the implementation does not handle automatic inserting of schemas into files for lsp to pick that up
- using this implementation JSON should be supported, and yaml should be supported if user used redhats yaml vscode extension 
- implementation using an experimental flag, to test things out
```diff
import { defineConfig } from 'astro/config';

export default defineConfig({
	experimental: {
+		contentCollectionJsonSchema: true
	}
});

```

- first caveat, the used library has known issues for advanced zod schemas (https://github.com/StefanTerdell/zod-to-json-schema#known-issues)
- second caveat, users have to allow `$schema` property in their zod schema, otherwise they get a warning
```diff
import { defineCollection, z } from 'astro:content';

const test = defineCollection({
	type: 'data',
	schema: z.object({
+		"$schema": z.string().optional(),
		test: z.string()
	}),
});

export const collections = { test };
```
- third caveat, users have to manually reference the schema in their content collection data files
```diff
{
+  "$schema": "../../../.astro/schemas/collections/i18n.json",
  "homepage": {
    "greeting": "ME",
    "preamble": "ME"
  }
}

```

## Testing

- not added any tests yet, since we would most likely need some snapshot tests 🤔 

## Docs

- I think the config reference is pulled automatically for the experimental flag
- unsure whether we need docs &or where we would document experimental features (cc @sarah11918)


packages/astro/package.json
```diff
@@ -176,7 +176,8 @@
     "vitefu": "^0.2.5",
     "which-pm": "^2.1.1",
     "yargs-parser": "^21.1.1",
-    "zod": "^3.22.4"
+    "zod": "^3.22.4",
+    "zod-to-json-schema": "^3.22.4"
```
alexanderniebuhr, line 180: The library looks to be actively maintained, and is officially linked from zod's docs. I personally think it is safe to use for now, but we could look into our own implementation in the future.

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,34 @@
+---
+"astro": minor
```
alexanderniebuhr, line 2: Not sure if this really needs to be a minor or if we could get it out sooner, cause it is behind an experimental flag. I would prefer the second one.

packages/astro/src/@types/astro.ts
```diff
@@ -1640,6 +1640,25 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 * 	experimental: {
+		 * 		contentCollectionJSONSchema: true,
+		 * 	},
+		 * }
+		 * ```
+		 */
+		contentCollectionJSONSchema?: boolean;
```
alexanderniebuhr, line 1660: How do we feel about this experimental flag name?

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +423,29 @@ async function writeContentFiles({
 				for (const entryKey of Object.keys(collection.entries).sort()) {
 					const dataType = collectionConfig?.schema ? `InferEntrySchema<${collectionKey}>` : 'any';
 					dataTypesStr += `${entryKey}: {\n	id: ${entryKey};\n  collection: ${collectionKey};\n  data: ${dataType}\n};\n`;
+					if(settings.config.experimental.contentCollectionJSONSchema) {
+						if (collectionConfig) {
+							// writing inside loop makes sense for the json schema, since we want to have a file per collection
+							const collectionSchemasDir = new URL('./schemas/collections/', contentPaths.cacheDir);
```
alexanderniebuhr, line 429: How do we feel about `.astro/schemas/collections/NAME.json`?

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +423,29 @@ async function writeContentFiles({
 				for (const entryKey of Object.keys(collection.entries).sort()) {
 					const dataType = collectionConfig?.schema ? `InferEntrySchema<${collectionKey}>` : 'any';
 					dataTypesStr += `${entryKey}: {\n	id: ${entryKey};\n  collection: ${collectionKey};\n  data: ${dataType}\n};\n`;
+					if(settings.config.experimental.contentCollectionJSONSchema) {
+						if (collectionConfig) {
```
ematipico, line 427: nit: you can have one single `if` that contains these two checks

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +423,29 @@ async function writeContentFiles({
...
+							// writing inside loop makes sense for the json schema, since we want to have a file per collection
+							const collectionSchemasDir = new URL('./schemas/collections/', contentPaths.cacheDir);
+							if (!fs.existsSync(collectionSchemasDir)) {
+								fs.mkdirSync(collectionSchemasDir, { recursive: true });
+							}
+	
+							if (collectionConfig?.schema)
```
ematipico, line 434: We should add this check to line 426

packages/astro/src/content/types-generator.ts
ematipico, line 429: There's no need to create the same `URL` at every instance of the loop. This variable can be created once before the loop.

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +423,29 @@ async function writeContentFiles({
...
+								await fs.promises.writeFile(
+									new URL(`./${collectionKey.replace(/"/g, '')}.json`, collectionSchemasDir),
+									JSON.stringify(
+										zodToJsonSchema(collectionConfig?.schema, {
```
ematipico, line 438: `writeFile`, `JSON.stringify` and maybe `zodToJsonSchema` could throw errors. We should catch them.

ematipico, line 438: nit: `collectionConfig?.schema` -> `collectionConfig.schema`

packages/astro/src/@types/astro.ts
ematipico: Looks good. Nit for me: `contentCollectionJsonSchema`

packages/astro/src/content/types-generator.ts
ematipico: Looks good. I would remove `collections`, for brevity, unless we predict to have more schemas. If so, they I would invert the names: `.astro/collections/<NAME>.schema.json`

alexanderniebuhr: That's a good question. I'm not sure if we predict to have more schemas, but I guess we might, so maybe keeping it safer now? 🤔

packages/astro/src/content/types-generator.ts
ematipico: `writeFile` is an I/O operation, so you should always expect the impossible. Even permission issues. It's always good to use a `try/catch`. Here, I suppose `JSON.stringify`, here, is safe if it's that `zodToJsonSchema` returns a serializable JSON. If you use `JSON.stringify` when reading from disk, a `try/catch` is advised because the file might have been tempered.

packages/astro/src/content/types-generator.ts
ematipico: I don't have particular preferences, even your initial suggestion is fine

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,36 @@
...
+});
+```
+
+To be able to use the schemas, you need to adapt your zod definition and reference it inside the json manually:
+
+```diff
+const test = defineCollection({
+	type: 'data',
+	schema: z.object({
++		"$schema": z.string().optional(),
```
natemoo-re, line 25: Ah this would be a nice limitation to avoid... Are you up to try to automatically allow the `$schema` property? It should probably _not_ be exposed in the resolved data, but even if it is, every data collection having `$schema?: string` sounds like a fair tradeoff.

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,36 @@
...
+    test: z.string()
+	}),
+});
+```
+
+```diff
+{
++  "$schema": "../../../.astro/schemas/collections/test.json",
```
natemoo-re, line 33: This is valid and straight-forward, but you can also adjust your workspace settings in VSCode to avoid needing to set `$schema` on every entry. Might be worth mentioning!

```
"json.schemas": [
  {
    "fileMatch": [
      "/src/content/test/**"
    ],
    "url": "../../../.astro/schemas/collections/test.json"
  }
]
```

It would be nice to link directly to the [VS Code JSON docs](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings)

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +431,27 @@ async function writeContentFiles({
...
+										name: collectionKey.replace(/"/g, ''),
+										markdownDescription: true,
+										errorMessages: true,
+									}),
+									null,
+									2
+								)
+							);
+						} catch (error) {
+							// new AstroError(error)
```
natemoo-re, line 452: Still needs to be handled!

natemoo-re (approved):
Looks great! Have a few suggestions still, but not blocking so I'm preemptively approving.

packages/astro/src/content/types-generator.ts
alexanderniebuhr: I was thinking about logging a warning using our logger, but it seems that is not available here.

Throwing an error seems a bit severe, as a dev I would expect to continue even if this experimental step fails?

natemoo-re: I think we should have access to the logger since `settings` is passed in.

Agree that throwing an unhandled error would be too severe here. It seems intentional to gracefully degrade the experience if we run into problems with this step.

packages/astro/src/content/types-generator.ts
alexanderniebuhr: I like your suggestion of having `collections` first, but I'm unsure about `.schema.json`. Any other opinions? 
cc @natemoo-re

.changeset/slow-items-heal.md
alexanderniebuhr: @natemoo-re I guess we should only add the `$schema` property if it is an object anyway, otherwise the user wouldn't have a way to set that anyway

ematipico (approved):

packages/astro/src/content/types-generator.ts
natemoo-re: I think `collections/*` and `*.schema.json` are great!

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +435,36 @@ async function writeContentFiles({
...
+					if (
+						settings.config.experimental.contentCollectionJsonSchema &&
+						collectionConfig?.schema
+					) {
+						let zodSchemaForJson = collectionConfig.schema;
+						if (zodSchemaForJson instanceof z.ZodObject) {
+							zodSchemaForJson = zodSchemaForJson.extend({
+								$schema: z.string().optional(),
+							});
+						}
```
natemoo-re, line 447: Nice, this seems like a reasonable way to handle it!

natemoo-re (approved):

packages/astro/src/content/types-generator.ts
ematipico: Sounds good to me!

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,38 @@
+---
+"astro": minor
+---
+
+Adds experimental JSON Schema support for content collections of with `type: 'data'`.
+
+To enable this feature, add the experimental flag:
```
sarah11918, line 7:
```suggestion
Adds experimental JSON Schema support for content collections of with `type: 'data'`.

This feature will auto-generate a content collection JSON `$schema` for TypeScript-style autocompletion/hints in tools like VSCode.

To enable this feature, add the experimental flag:
```
Make this sentence accurate, but I think some kind of line describing what this is or why it's useful would work here! (I just pulled this from the [roadmap discussion description](https://github.com/withastro/roadmap/discussions/696), but of course this feature might not do exactly that! Write whatever makes sense here, and I'll come back and look at it!)

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,38 @@
...
+
+To make sure VSCode picks up the schema file, you have multiple option. You can either reference the file in every file of your content collection:
+```diff
+{
++  "$schema": "../../../.astro/collections/test.schema.json",
+  "test": "test"
+}
+```
+
+Or you can set your VSCode settings to match them, read more in the [VSCode docs](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings):
```
sarah11918, line 27:
``````suggestion
Then, add a `$schema` property to any data collection in `src/content.config.ts`:

```diff
// src/content/config.ts
import { defineCollection, z } from 'astro:content';

const test = defineCollection({
	type: 'data',
	schema: z.object({
+		"$schema": z.string().optional(),
		test: z.string()
	}),
});

export const collections = { test };
```

This experimental implementation requires you to manually reference the schema in each data entry file of the collection:

```diff
// src/content/test/entry.json
{
+  "$schema": "../../../.astro/collections/test.schema.json",
  "test": "test"
}
```

Alternatively, you can set this in your [VSCode `json.schemas` settings](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings):
``````

File: .changeset/slow-items-heal.md [hunk 14]
```diff
@@ -0,0 +1,38 @@
...
+```diff
+"json.schemas": [
+  {
+    "fileMatch": [
+      "/src/content/test/**"
+    ],
+    "url": "../../../.astro/collections/test.schema.json"
+  }
+]
+```
```
sarah11918, line 38:
``````suggestion
```

Note that this initial implementation uses a library with [known issues for advanced Zod schemas](https://github.com/StefanTerdell/zod-to-json-schema#known-issues), so you may wish to consult these limitations before enabling the experimental flag.
``````

File: packages/astro/src/@types/astro.ts [hunk 15]
```diff
@@ -1644,6 +1644,25 @@ export interface AstroUserConfig {
...
 		contentCollectionCache?: boolean;
 
+		/**
+		 * @docs
+		 * @name experimental.contentCollectionJsonSchema
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.5.0
+		 * @description
+		 * Enables generation of JSON Schema files for content collections of type `data`.
```
sarah11918, line 1654: "Once the changeset is figured out, those instructions, limitations etc should also be here! This is the place in docs where people will go to figure out what this does and how it works!

When we like the changeset, it can be a mostly copy-paste here, except that this can be less "here's a new feature!" and more just boring, "here's what you do."

Reminder to self: this should also link to the content collections page in docs.

sarah11918 (commented):
Thanks for this @alexanderniebuhr ! Left some things to think about for the changeset, which will then take us on to the actual docs content!

Because this is experimental, we typically would *not* mention it in docs on e.g. the content collections page. So, we don't need another docs PR, because this should be the only place it appears while experimental!  :raised_hands:

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +450,36 @@ async function writeContentFiles({
...
+										errorMessages: true,
+									}),
+									null,
+									2
+								)
+							);
+						} catch (err) {
+							logger.warn(
+								'content',
+								`An error was encountered while creating the Json schema. Proceeding without it. Error: ${err}`
```
alexanderniebuhr, line 479: Not sure if this warning also needs rewording. If so please feel free to commit that yourself.

sarah11918: JSON is probably all caps when referred to as a name/noun?

.changeset/slow-items-heal.md
```diff
@@ -0,0 +1,44 @@
+---
+"astro": minor
+---
+
+Adds experimental JSON Schema support for content collections.
+
+This feature will auto-generate a JSON Schema for content collection of `type: 'data'`, which can be used as the `$schema` value for TypeScript-style autocompletion/hints in tools like VSCode.
```
sarah11918, line 7:
```suggestion
This feature will auto-generate a JSON Schema for content collections of `type: 'data'` which can be used as the `$schema` value for TypeScript-style autocompletion/hints in tools like VSCode.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1644,6 +1644,54 @@ export interface AstroUserConfig {
...
 		contentCollectionCache?: boolean;
 
+		/**
+		 * @docs
+		 * @name experimental.contentCollectionJsonSchema
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.5.0
+		 * @description
+		 * This feature will auto-generate a JSON Schema for content collection of `type: 'data'`, which can be used as the `$schema` value for TypeScript-style autocompletion/hints in tools like VSCode.
```
sarah11918, line 1654:
```suggestion
		 * This feature will auto-generate a JSON schema for content collections of `type: 'data'` which can be used as the `$schema` value for TypeScript-style autocompletion/hints in tools like VSCode.
```

packages/astro/src/content/types-generator.ts
```diff
@@ -419,6 +451,36 @@ async function writeContentFiles({
...
```
alexanderniebuhr, line 480:
```suggestion
								`An error was encountered while creating the JSON schema. Proceeding without it. Error: ${err}`
```

sarah11918:
```suggestion
								`An error was encountered while creating the JSON schema. Proceeding without it. Error: ${err}`
```

packages/astro/src/content/types-generator.ts
alexanderniebuhr: According to https://json-schema.org/ yes, so I've updated the warning message

sarah11918 (approved):
I think just some tiny nits to clean up here, otherwise, LGTM!

Issue #5083, opened 2022-10-13
[Build] vanilla js is not putted into separeted file like css


ampmonteiro:
### What version of `astro` are you using?

1.5.0

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Linux, Windows

### Describe the Bug

Template: 'Just the Basics'

in index.astro file, inside of script tag jus add:
```js
<script>
  fetch('/api.json')
	  .then((r) => r.json())
	  .then((d) => console.log('data', d))
	  .catch((er) => console.log('er', er));
 </script>
```

after that run build
``` npm run build  ```

dist folder:
![image](https://user-images.githubusercontent.com/31895754/195711260-3d915353-6156-4d82-bf3e-9b9086ba4827.png)

index.html:
![image](https://user-images.githubusercontent.com/31895754/195711406-b86ed603-704b-42f6-8f66-746505aa8078.png)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-tshpt6?file=public%2Fapi.json,src%2Fpages%2Findex.astro&on=stackblitz

### Participation

- [ ] I am willing to submit a pull request for this issue.

**Comment by ampmonteiro** [comment]
Hi,

you should mention how and point out to this in your docs too, beside vite, before close the issue
https://docs.astro.build/en/reference/configuration-reference/#vite 
``` js  
import { defineConfig } from 'astro/config';

// https://astro.build/config
export default defineConfig({
  vite: {
    build: {
      assetsInlineLimit: '0',
    },
  },
});

```

PR #10154, opened 2024-02-18
fix: better assetsInlineLimit runtime type checking

### LINKED ISSUE #10153: Regression with string `assetsInlineLimit` in 4.4.0
### Astro Info

```block
Astro                    v4.4.0
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

With the following config, the basic starter site would build just fine in `4.3.7`, but fails in `4.4.0`.

```typescript
export default defineConfig({
    vite: {
        build: {
            assetsInlineLimit: '0',
        }
    }
});
```

Using the basic template from `npm create astro@latest`:

- `npm run build` with `4.3.7` - no issues
- `npm run build` with `4.4.0` - errors with the following:

```
18:05:41 [ERROR] [vite] x Build failed in 658ms
[astro:rollup-plugin-inline-stylesheets] assetsInlineLimit is not a function
  Stack trace:
    at shouldInlineAsset (file:///E:/GitHub/astro-testing/verdant-visual/node_modules/astro/dist/core/build/plugins/util.js:44:18)
    at file:///E:/GitHub/astro-testing/verdant-visual/node_modules/rollup/dist/es/shared/node-entry.js:19579:40
```

This seems to have been introduced with the bump to Vite 5.1 in https://github.com/withastro/astro/pull/10120.

`4.4.0` should build the site just fine, as it did in <= `4.3.7`.

https://stackblitz.com/edit/github-nunp67

### Participation

Cherry:
## Changes

Fixes #10153

This now mimics the behaviour in Vite core directly, which passes the config option to `Number` first: https://github.com/vitejs/vite/blob/71dc6a6b7d41c27133f04b92256bead74b8f2127/packages/vite/src/node/plugins/asset.ts#L419

## Testing

I added a test to the `css-assets` test file, which was essentially a copy/paste of the existing one there, but sets the config option to a string instead of a number. Before my changes, this was failing, but now it's passing as per previous behaviour in 4.3.x. 

If there's a better place to add a test for this, please let me know.

## Docs

I don't think any additional docs are needed here, as it's just a minor regression from 4.3.x.


packages/astro/src/core/build/plugins/util.ts
```diff
@@ -75,8 +75,8 @@ export function shouldInlineAsset(
 	assetPath: string,
 	assetsInlineLimit: NonNullable<BuildOptions['assetsInlineLimit']>
 ) {
-	if (typeof assetsInlineLimit === 'number') {
-		return Buffer.byteLength(assetContent) < assetsInlineLimit;
+	if (typeof assetsInlineLimit === 'number' || typeof assetsInlineLimit === 'string') {
+		return Buffer.byteLength(assetContent) < Number(assetsInlineLimit);
```
ematipico, line 79: This operation can return a `NaN`. Should we check for that too, and use the default if that's the case?

bluwy:
But `assetsInlineLimit` doesn't accept a string? I think it only happened to work (https://vitejs.dev/config/build-options.html#build-assetsinlinelimit).

Cherry:
While true, it's documented to only accept a number, passing `'0'` especially has been used and is even still recommended every now and then in Astro's discord.

- https://github.com/withastro/astro/issues/5083#issuecomment-1279449649
- https://canary.discord.com/channels/830184174198718474/1197000366752026645/1197000809318203424
- https://canary.discord.com/channels/830184174198718474/1195099820839407707/1195099822911389818

As well as a handful of projects using it around GitHub: https://github.com/search?q=%22assetsInlineLimit%3A+%5C%22%22+OR+%22assetsInlineLimit%3A+%27%22&type=code
 
etc.

Considering how it worked in 4.3.x, but then stopped in 4.4.x, I think it would be good to handle this at least in 4.x, and officially dropped in 5.x, perhaps with the next major version of Vite?

packages/astro/src/core/build/plugins/util.ts
Cherry: Good catch. I've refactored this function a bit to better handle `NaN` and simplified the logic a bit to match Vite's.

bluwy:
I don't think it being misused or misrecommended today means that it is a stable API that we can't break. We had always documented it to accept numbers, so Vite should be able to fix that in a patch even. Hence I don't think we should allow strings here.

matthewp:
How many reports of this breaking peoples apps have we gotten? I would agree to take some action if it were a large number.

Cherry:
With Astro 4.4 releasing on the tail end of last week, I believe I'm the first to officially hit and report this. But with the AI bot in Discord still recommending this as recently as just 2 days ago in the Discord, I won't be the only one.

For the small addition this adds to the codebase, with much better runtime type-guarding (passing `null` or some other data-type would also error in 4.4.x since the code now assumes anything that's not a number is a function), I figure it would be a good idea to merge this and fix the edge-cases, as well as improve DX.

Cherry:
If explicitly supporting a `string` isn't something that's wanted to be supported, would you accept the `typeof function` check at least, as like in Vite, to improve the DX people get when passing something they shouldn't?

matthewp:
Thanks for arguing your position with passion @Cherry! 

One thing I'm not understanding is, why is so much code needed to support strings in the first place? Can't we just do `Number(assetsInlineLimit)` without checking the type? If it's NaN it will fail the check, everything should be good, right?

Princesseuh:
The only thing I'm getting from this thread is that AI chat bots were a mistake. I understand your point completely, and it's ultimately neither your or our fault really, but I cannot help but be extremely annoyed at the fact that we're accommodating for an inaccurate technology 🤦‍♀️ But I guess in the first place we're accommodating for JavaScript not having types, everything sucks!

ematipico (approved):

bluwy (changes requested):
If others are fine with this change, I won't block this though. However, I'd prefer if we check as `typeof` function only and unconditionally coerce it with `Number(...)` like Vite. (so no `NaN` checks). I'd also update the changeset to mention that its not a regression and remove the tests as I don't think it's something we should guarantee no to break again.

While we're indeed using JS and types are only a first-line of defence, it would be taxing to have to validate every option exposed on the config so that everything works as expected. Hence, types helps draw that line to prevent perf implications.

matthewp:
@Cherry Yes, agree with @bluwy's compromise here. Thanks for working with us!

Cherry:
Thanks folks, I've updated to simplify the check now, and tweaked the changeset. Let me know if there's anything further you'd like to see.

packages/astro/src/core/build/plugins/util.ts
```diff
@@ -75,13 +75,14 @@ export function shouldInlineAsset(
...
+		const result = assetsInlineLimit(assetPath, Buffer.from(assetContent));
+		if (result != null) {
+			return result;
+		}
 	}
-
-	const result = assetsInlineLimit(assetPath, Buffer.from(assetContent));
-	if (result != null) {
-		return result;
+	if (typeof assetsInlineLimit === 'number' || typeof assetsInlineLimit === 'string') {
```
bluwy, line 84: I think we can change this to
```suggestion
	} else {
```

Since if it's not a function, it should already be either a number or string

bluwy: Yeah I think we should change it.

Cherry: If I simply change this to an `else`, the `4096` default fallback at the bottom of the function will never fire. I would have to at least check `else if (typeof assetsInlineLimit !== 'undefined')`. It might not be set at all.

Would you like me to do that instead of the string/number check?

bluwy: It will fire if `assetsInlineLimit` is a function but the function returns undefined. `assetsInlineLimit` should already be set by default because the param has `NonNullable`.

Maybe to make it less confusing, we can move the `4096` fallback inside the `typeof assetsInlineLimit === 'function'` block so it's less confusing.

bluwy: We currently type it as `assetsInlineLimit: NonNullable<BuildOptions['assetsInlineLimit']>`. That means it can only be either a string or a function. It can't be nullable because we marked it as `NonNullable` (the consumers will always set a value). In which case, we don't need to cover the 3rd case. We should able to trust the types here.

Overall, my suggestion would be this kind of flow:

```js
	if (typeof assetsInlineLimit === 'function') {
		const result = assetsInlineLimit(assetPath, Buffer.from(assetContent));
		if (result != null) {
			return result;
		} else {
			return Buffer.byteLength(assetContent) < 4096; // Fallback to 4096kb by default (same as Vite)
		}
	}

	return Buffer.byteLength(assetContent) < Number(assetsInlineLimit);
```

Cherry: Sure. I'd consider my solution safer (not everyone uses TypeScript), but I understand your want to simplify, so I've updated that now.

bluwy (approved):
Thanks for making the change!

.changeset/late-bears-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Improves runtime type-checking with the vite.build.assetsInlineLimit configuration option, which could result in errors when passing invalid options like strings. Astro now automatically casts this to a number to match the Vite behaviour.
```
lilnasy, line 5:
```suggestion
Fixes an issue where `config.vite.build.assetsInlineLimit` could not be set as a function.
```

Issue #9905, opened 2024-01-31
Bug/Tests: some tests that expect errors to be thrown can never fail


espy:
### Astro Info

```block
This is a bug in the Astro tests on `main`, so this field does not apply.
```

### Describe the Bug

There’s a specific style of test that will always pass, for example in [packages/astro/test/i18n-routing.test.js#L555-L564](https://github.com/withastro/astro/blob/main/packages/astro/test/i18n-routing.test.js#L555-L564)`, where the expected outcome is that an error _is_ thrown:

```js
it("should render a 404 because the route `fr` isn't included in the list of locales of the configuration", async () => {
  try {
    await fixture.readFile('/fr/start/index.html');
    // failed
    return false;
  } catch {
    // success
    return true;
  }
});
```

~~However, if you switch the `return false;` and the `return true;`, the test still passes.~~
**Correction:** that ☝️ is not the correct way to reproduce this problem. The correct reproduction is to have the `try` block _not_ throw an error, which should make the test fail:

```js
it("This test should pass, success means an error was thrown", async () => {
  try {
    throw new Error('assume readFile failed')
    // failed
    return false;
  } catch {
    // success
    console.log('expected')
    return true;
  }
});

it("This test should fail, success means an error was thrown", async () => {
  try {
    const something = "unexpected"
    console.log(something)
    // failed
    return false;
  } catch {
    // success
    return true;
  }
});

```

However, both tests pass:

```bash
expected
      ✔ This test should pass, success means an error was thrown
unexpected
      ✔ This test should fail, success means an error was thrown
```

 I assume this requires the use of something like `assert.throws`, but I’m not an expert here.

### What's the expected result?

The tests that expect caught errors actually fail if no errors are caught.

### Link to Minimal Reproducible Example

Does not apply, bug is in the Astro tests

### Participation

- [ ] I am willing to submit a pull request for this issue.

espy:
Updated the issue, I had made a mistake by simplifying the reproduction too far. Switching the return statements only demonstrates that this test format will never fail if you’re using it to test than an error should _not_ be thrown, which doesn’t actually cover the described problem. Apologies if this caused any confusion.

ematipico:
Wanna send a PR? It should be fairly easy to update it. 

Make that a function, and use the `.to.throw`  function from chai

espy:
[...]

PR #10156, opened 2024-02-19
chore: fix tests in `packages/astro/test/i18n-routing` which use `try/catch`

abidjappie:
## Changes

- Add missing asserts in the tests for `packages/astro/test/i18n-routing` which use `try/catch` (related: https://github.com/withastro/astro/issues/9905)
- This was previously not implemented as it relied on moving tests to `nodejs`

## Testing

To run the test case:
```zsh
pnpm astro-scripts test "test/**/i18n-routing.test.js"  
```

## Docs

This only affects the tests, so no docs.


ematipico (approved):
Thank you!

PR #10166, opened 2024-02-20
Improve Astro style HMR for imported styles

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/9262

This PR removes some hacks needed before Vite 5.1. Now with Vite 5.1, we can register style dependencies directly to the Astro style virtual module instead, e.g. `/src/pages/index.astro?astro&type=style&index=0&lang.css`. We can more better prevent circular deps that caused https://github.com/vitejs/vite/pull/15585 and some tailwind issues.

The result is that CSS HMR should be more granular and simpler, but in this PR I had to leverage `astroFileToCompileMetadata` still in HMR to invalidate the metadata correctly, so that the Astro style virtual module can re-compile the new CSS.

## Testing

Added a new test and made sure HMR don't refresh for styles.

## Docs

m/a. bug fix.


packages/astro/src/core/compile/style.ts
```diff
@@ -11,18 +12,21 @@ export function createStylePreprocessor({
...
-		const id = `${filename}?astro&type=style&lang${lang}`;
+		const id = `${filename}?astro&type=style&index=${index}&lang${lang}`;
 		try {
 			const result = await preprocessCSS(content, id, viteConfig);
 
-			result.deps?.forEach((dep) => {
-				cssDeps.add(dep);
-			});
+			if (result.deps) {
+				cssDeps[index] = [...result.deps].map((dep) => normalizePath(dep));
```
ematipico, line 28:
```suggestion
				cssDeps[index] = result.deps.map((dep) => normalizePath(dep));
```

Not sure the spread operator is needed..

packages/astro/src/vite-plugin-astro/hmr.ts
```diff
@@ -1,62 +1,44 @@
...
+	// so the next transform will re-generate it.
+	for (const [astroFile, compileData] of astroFileToCompileMetadata) {
+		const isUpdatedFileCssDep = compileData.css.some((css) => css.dependencies?.includes(ctx.file));
+		if (isUpdatedFileCssDep) {
+			astroFileToCompileMetadata.delete(astroFile);
+		}
+	}
+
+	// Bail if not Astro file, we only handle them below
+	if (compileMetadata == null || !ctx.file.endsWith('.astro')) return;
```
ematipico, line 27: Why do we bail here not before the loop? The loop assumes `ctx.file` is an astro file, but here we bail. 

Also, line 15 can be moved after the loop because we don't use it before the loop.

packages/astro/src/vite-plugin-astro/index.ts
```diff
@@ -86,10 +82,18 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
...
+			// modules are compiled first, then its virtual modules.
 			const filename = normalizePath(normalizeFilename(parsedId.filename, config.root));
-			const compileMetadata = astroFileToCompileMetadata.get(filename);
+			let compileMetadata = astroFileToCompileMetadata.get(filename);
+			// If `compileMetadata` doesn't exist in dev, that means the virtual module may have been invalidated.
+			// We try to re-compile the main Astro module (`filename`) first before retrieving the metadata again.
+			if (!compileMetadata && server) {
+				const code = await loadId(server.pluginContainer, filename);
+				if (code != null) await compile(code, filename);
+				compileMetadata = astroFileToCompileMetadata.get(filename);
```
ematipico, line 93: Who does add data to `astroFileToCompileMetadata`? The function `loadId` or `compile`?

packages/astro/src/core/compile/style.ts
bluwy: `result.deps` is a `Set` and doesn't have `.map`

packages/astro/src/vite-plugin-astro/index.ts
bluwy: `compile` will populate `astroFileToCompileMetadata`.

packages/astro/src/vite-plugin-astro/hmr.ts
bluwy: The for-loop doesn't assume `ctx.file` to be an Astro file. It could be any file, hence we check against the list there. Line 15 is called before the loop, because line 22 could have deleted it. We still want to narrow down Astro file HMR even after the for-loop matches it. The for-loop is kinda doing its own handling regarding CSS dependencies HMR.

bluwy: Re line 15, thinking about it more I think you're right too. The only case where line 22 could mis-delete it is when `ctx.file` has a dependency of `ctx.file` itself, which is highly unlikely. I'll move that down then.

bluwy:
@ematipico I added some comments since your last review explaining some things better. Appreciate if you can take another look!

ematipico (approved):
Thank you for adding more context to how hmr internally works!

matthewp (approved):

PR #10175, opened 2024-02-20
unused constructs cleanup

lilnasy:
## Changes

- The resolve function is an environment-specific function used for island component URLs, it naturally fit as a method on the environment-specific `Pipeline` classes.
- Simplified `callMiddleware`, which had a hard to follow control flow, with a repeating pattern of checks.
- I found the existence of `ManifestData` confusing - it is an object with array of `RouteData` with no strong connection to the manifest. Replaced with `routes: RouteData[]`
- Moved `createResult()` fully into `renderContext,createResult`, allowing the creation of the `Astro` global to share code with the creation of `APIContext`.
- Removec `createAPIContext()`, which has been replaced with `renderContext.createAPIContext()`.

## Testing
- Does not affect behavior.

## Docs
- Does not affect usage.


packages/astro/src/core/render-context.ts
```diff
@@ -226,23 +258,18 @@ export class RenderContext {
...
+		const { pipeline: { i18n }, request, routeData, url } = this;
+		if (!i18n) return {
+			currentLocale: undefined,
+			preferredLocale: undefined,
+			preferredLocaleList: undefined
+		}
+		const { defaultLocale, locales, routing } = i18n
 		return (this.#i18nData = {
-			currentLocale: computeCurrentLocale(routeData.route, locales, routing, defaultLocale),
+			currentLocale: computeCurrentLocale(url.pathname, locales, routing, defaultLocale) ?? computeCurrentLocale(routeData.route, locales, routing, defaultLocale),
```
lilnasy, line 270: We are making two calls to `computeCurrentLocale()`. Previously, these two were made by `createApiContext()` and `createResult()` separately.

I would expect the `pathname` passed to `RenderContext.create()` to be the one true "file-system-matchable" path, but that isn't the case yet - something to tackle with the reroute/rewrite API.

packages/astro/src/@types/astro.ts
```diff
@@ -2687,29 +2683,29 @@ export type SSRComponentMetadata = {
 };
 
 export interface SSRResult {
+	/**
```
ematipico, line 2686: The move around of properties makes it difficult to understand what actually changed. We removed only `cookies` did we? Can you explain why those properties were removed?

packages/astro/src/core/base-pipeline.ts
```diff
@@ -52,9 +52,20 @@ export abstract class Pipeline {
...
-	abstract componentMetadata(routeData: RouteData): Promise<SSRResult['componentMetadata']> | void;
+
+	/**
+	 * Resolves specifiers in the inline hydrated scripts, such as
+	 * - \@astrojs/preact/client.js
+	 * - \@/components/Foo.vue
+	 * - /Users/macos/project/src/Foo.vue
+	 * - C:/Windows/project/src/Foo.vue (normalized slash)
+	 */
+	abstract resolve(s: string): Promise<string> | string;
```
ematipico, line 67: Finally, some documentation! Thank you! 🙏

packages/astro/src/core/middleware/callMiddleware.ts
```diff
@@ -42,54 +42,37 @@ export async function callMiddleware(
...
-	let middlewarePromise = onRequest(apiContext, next);
+	if (response && response instanceof Response) {
+		return response;
+	}
 
-	return await Promise.resolve(middlewarePromise).then(async (value) => {
-		// first we check if `next` was called
+	if (response === undefined) {
+		/**
+		 * This seems unintentional, but it might be breaking to backtrack.
```
ematipico, line 58: This comment is very cryptic, I am not sure about its meaning. It might be a comment for yourself, but other people need to understand too. Could you provide more explanation?

packages/astro/src/core/render/slots.ts
```diff
@@ -0,0 +1,82 @@
+import type { SSRResult } from '../../@types/astro.js';
```
ematipico, line 1: The entity of this change is missing from the description; could you please add it?

packages/astro/src/core/render-context.ts
ematipico, line 270: This change will break people's code. `url.pathname` in SSG with contain the path of the file, e.g. `/dist/en/blog.html` that's not what we want. We want the actual "remote" path e.g. `/en/blog`

Also, we don't even want the "file system" path, because it can contain things like `[...dynamic].astro`, and we can't compute the locale anymore.

Hence why I passed the actual `Route` type. We should revert it.

ematipico: The fact that we are making two calls without any inline comment that explains this "anti-pattern" seems like a bug to me.

I know you are explaining this in a comment on Github, but this comment will eventually disappear and the context of this "anti-pattern" won't be explained anymore. Please leave a comment that explains this.

Personally, I would use `routeData.pathname` if it exists, and `routeData.route` otherwise. `Request#url` isn't reliable because in SSG has this value: `/en.html` and we won't be able to pick the current locale anymore.

ematipico (commented):
Looking good. I left some questions and feedback. The entity of some changes isn't described in the PR, which makes the review a bit harder because there isn't any context.

packages/astro/src/@types/astro.ts
lilnasy: Yes, only cookies were removed - nothing used them anymore. The others were moved to keep them alphabetical - it keeps them scannable and is the general pattern of most interfaces.

packages/astro/src/core/middleware/callMiddleware.ts
lilnasy: A middleware returning undefined is not a documented pattern. Here, we attempt to correct that misuse by directly using the response returned by renderPage/renderEndpoint. That seems unintentional.

packages/astro/src/core/render/slots.ts
lilnasy: This is not new. This file is `result.ts` renamed after createResult was removed from it.

packages/astro/src/core/middleware/callMiddleware.ts
ematipico: That's probably a leftover of the initial logic. Initially, we designed the feature in a way where a user could call just `next()` without returning anything from the middleware.

We changed that behaviour to be more in line with how other middleware in the JS ecosystem works. So yeah, we can remove that check

lilnasy: That makes sense. Now, all that `callMiddleware()` needs to do is throw `MiddlewareNotAResponse`. That seems simple enough that it can be done directly, instead being extracted out into a function. What do you think?

lilnasy (commented):
I'll take a look at the merge conflicts soon.

packages/astro/src/core/middleware/callMiddleware.ts
ematipico: Yeah that sounds good to me!

ematipico (approved):

matthewp (changes requested):
This change is too big and seems to mostly move code around. It definitely does more than remove unused code. I'm sure some of the changes are valuable and good, but it's hard to judge given the number of them.

Given the amount of time it takes to review PRs, I don't think those that move code around without either fixing a bug or making a user-facing feature is something we should keep doing. An occasional small refactor is ok, but in general code improvements can/should happen in service of an improvement to the framework.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -748,27 +748,6 @@ export const ResponseSentError = {
 	message: 'The response has already been sent to the browser and cannot be altered.',
 } satisfies ErrorData;
 
-/**
- * @docs
- * @description
```
Princesseuh, line 753: Irrelevant of if this should be merged or not (agree with Matthew's comment), errors should never be deleted (at least, not in patches or minors), they should be deprecated using the `@deprecated` JSDoc tag

lilnasy: Oh right, it's also used for docs.

Issue #10171, opened 2024-02-20
Sitemap sometimes not generated correctly


ktym4a:
### Astro Info

```block
Astro                    v4.4.1
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/sitemap
```

### Describe the Bug

After https://github.com/withastro/astro/pull/9846, sometimes the structure is incorrect when generating the sitemap. (even not set `prefix`)
but `${pefix}-0.xml` is always the correct structure.

https://github.com/withastro/astro/assets/51779800/2407030f-c4d3-410a-ad10-b0743cba9b4d

### What's the expected result?

Always generated with the correct structure

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

ktym4a:
cc @ematipico 
I currently have no idea how to fix this.

ematipico:
It's definitely a regression of the latest changes, we'll have to debug them

ktym4a:
Is there a function that is called after `cli` function?

After some debugging, I think this is more of a build-side issue than a sitemap issue. (Not sure)

Throwing an error after runCommand for debugging, It seems that the sitemap structure is always correct.

```diff
export async function cli(args: string[]) {
	const flags = yargs(args, { boolean: ['global'], alias: { g: 'global' } });
	const cmd = resolveCommand(flags);
	try {
		await runCommand(cmd, flags);
+		throw new Error('Just testing...');
	} catch (err) {
		const { throwAndExit } = await import('./throw-and-exit.js');
		await throwAndExit(cmd, err);
	}
}
```

fisherfrontier:
Also seeing this problem as of @astrojs/sitemap 3.1.0. I'm getting this output in `sitemap-index.xml` (missing closing tag `</sitemapindex>`):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <sitemap>
        <loc>https://www.fisherfrontier.com/sitemap-0.xml</loc>
    </sitemap>

```

Per [Google Search Central](https://developers.google.com/search/docs/crawling-indexing/sitemaps/large-sitemaps#example-sitemap-index), should include closing tag `</sitemapindex>`, like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #9846, opened 2024-01-27
Add option to prefix sitemap

ktym4a:
## Changes

Google Serach Console sometimes not work by `sitemap-*.xml`, so add an option to change the file name.
https://github.com/ekalinin/sitemap.js/pull/395

Follow this one with `Want to customize that?` section.
https://github.com/ekalinin/sitemap.js?tab=readme-ov-file#create-sitemap-and-index-files-from-one-large-list

## Testing

Add prefix.test.js.
Test the content is the same with and without the prefix option.
Test the link is in `*-index.xml` correct.

## Docs

Docs need to be updated, because there are more options

PR #10179, opened 2024-02-21
Revert "Add option to prefix sitemap"

ematipico:
Reverts withastro/astro#9846

Closes https://github.com/withastro/astro/issues/10171


.changeset/slow-cats-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/sitemap": patch
+---
+
+Rever https://github.com/withastro/astro/pull/9846
```
Princesseuh, line 5: A description here to explain why it was reverted would be great in case any users are using the feature!

Princesseuh (approved):

PR #10180, opened 2024-02-21
chore: use biome to sort imports - only test files

ematipico:
## Changes

That's how I would like to start introducing Biome in the repository. 

Import sorting of the test files. They should not change the logic of the source code.

## Testing

CI should work.

After merging this PR, I will make a new PR to ignore the commit from git.

## Docs

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "format": "pnpm run format:code",
     "format:ci": "pnpm run format:code",
     "format:code": "prettier -w \"**/*\" --ignore-unknown --cache",
-    "format:imports": "organize-imports-cli ./packages/*/tsconfig.json ./packages/*/*/tsconfig.json",
+    "format:imports": "biome check --apply .",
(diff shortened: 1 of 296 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "format": "pnpm run format:code",
     "format:ci": "pnpm run format:code",
```
bluwy, line 17: Should we also have the format scripts run the import format? e.g. appending a `&& pnpm run format:imports`

I guess we also don't need `format:ci` as it's the same as `format`, and we can update that in `format.yml`

```suggestion
    "format": "pnpm run format:code && pnpm run format:imports",
    "format:ci": "pnpm run format:code && pnpm run format:imports",
```

bluwy (approved):
I like this start! Better to merge it as soon as possible to prevent conflicts. I only have one (optional) comment below.

package.json
ematipico: Good call! I didn't think this through enough. Goo catch

PR #10185, opened 2024-02-21
feat: middleware and virtual routes

ematipico:
## Changes

This is just a PoC. The PR adds a test that will need to pass eventually.

The test case is a middleware that is able to "catch" a route that doesn't exist and return a `Response` out of it.

### DIFF
--- a/packages/astro/test/fixtures/middleware space/src/middleware.js
+++ b/packages/astro/test/fixtures/middleware space/src/middleware.js
@@ -2,7 +2,7 @@ import { defineMiddleware, sequence } from 'astro:middleware';
 
 const first = defineMiddleware(async (context, next) => {
 	if (context.request.url.includes('/virtual')) {
-		return new Response('<span>New content!!</span>', {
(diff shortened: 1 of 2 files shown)


packages/astro/test/fixtures/middleware space/src/middleware.js
```diff
@@ -1,6 +1,11 @@
 import { defineMiddleware, sequence } from 'astro:middleware';
 
 const first = defineMiddleware(async (context, next) => {
+	if (context.request.url.includes('/virtual')) {
+		return new Response('<span>New content!!</span>', {
```
lilnasy, line 5:
```suggestion
		return new Response('<span>Virtual!!</span>', {
```

bluwy:
Is this PR ready. The description says:

> The PR adds a test that will need to pass eventually.

It looks like the test already passes?

ematipico:
@bluwy No, it's not ready yet. Sorry, I got caught by other stuff and forgot to follow up

PR #10186, opened 2024-02-21
feat(toolbar): Allow every element to have every color

Princesseuh:
## Changes

Previously only a few elements could have different colors, buttons and badges. This PR makes it so every element can be customized in the same fashion. Additionally, it is now fully reactive, so you can change it at runtime however you want.

A notable exception however is Tooltip, because that one doesn't use standard colors and need a bit more design work.

image

(the card in this screenshot are all hovered, their neutral state has no colors.)

## Testing

We don't really test the precise style of everything at this time, nonetheless everything should be clickable and everything like before, so tests should pass!

## Docs

https://github.com/withastro/docs/pull/7191


bluwy (approved):

packages/astro/src/runtime/client/dev-toolbar/ui-library/badge.ts
```diff
@@ -2,21 +2,35 @@ type BadgeSize = 'small' | 'large';
...
+		this.updateStyle();
+	}
+
+	get badgeStyle() {
+		return this._badgeStyle;
+	}
+
+	set badgeStyle(value) {
+		this._badgeStyle = value;
+		this.updateStyle();
```
natemoo-re, line 23: Sorry to be super annoying, but what if `badgeStyle` isn't one of the enum values? Can we validate at runtime here? Maybe a fallback to gray is better than an error?

packages/astro/src/runtime/client/dev-toolbar/ui-library/badge.ts
```diff
@@ -2,21 +2,35 @@ type BadgeSize = 'small' | 'large';
...
+	_badgeStyle: BadgeStyle = 'purple';
+
+	get size() {
+		return this._size;
+	}
+
+	set size(value) {
+		this._size = value;
+		this.updateStyle();
+	}
```
natemoo-re, line 15: Same comment with `size`.

natemoo-re (commented):
Looks great! I worry about users accidentally passing incorrect attribute values, then losing the element because everything is invisible, so validation would be nice to have, but maybe not that critical.

Also, we have a `blue` palette if you want to expose that option too! Base is `#3645D9`, light is `#BDC3FF`, and dark is `#0B1466`.

packages/astro/src/runtime/client/dev-toolbar/ui-library/badge.ts
Princesseuh: Oh no, that's definitely a valid worry to have. I'll add some validation

Princesseuh: Added validation in the coolest way I know.

ematipico (approved):

sarah11918 (approved):
Docs approves!

PR #10189, opened 2024-02-22
Add multiple cdn v2

peng:
## Changes

- What does this change?
Add build.assetsPrefix is also Object.

I also think `build.assetsPrefix` is also Object is good.

Previous PR: https://github.com/withastro/astro/pull/9898

## Testing

Add `packages/astro/test/astro-assets-prefix-mult-cdn.test.js` file for test add multiple cdn option.

## Docs

Docs add how to use build.assetsPrefix with object Record<string, string> type.

Docs PR: https://github.com/withastro/docs/pull/7183


packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -94,9 +96,15 @@ export default function assets({
...
-						isESMImportedImage(options.src) ? options.src.src : options.src
-					).replace(settings.config.build.assetsPrefix || '', '');
+					const ESMImportedImageSrc = isESMImportedImage(options.src) ? options.src.src : options.src;
+					let pf = ''
+					const fileExtension = getFileExtension(ESMImportedImageSrc)
+					if (settings.config.build.assetsPrefix && typeof settings.config.build.assetsPrefix === 'string')  {
+						pf = settings.config.build.assetsPrefix
+					} else if (isObject(settings.config.build.assetsPrefix)) {
+						pf = getAssetsPrefix(fileExtension, settings.config.build.assetsPrefix)
+					}
```
ematipico, line 106: There's a lot of repeated code here and in other places. The prefix can be easily computed like this:

```js
const fileExtension = getFileExtension(ESMImportedImageSrc)
const assetPrefix = getAssetsPrefix(fileExtension, settings.config.build.assetsPrefix);
```

The function `getAssetsPrefix` already checks whether `config.build.assetsPrefix` is a string/object, so we don't need to check this *here, _again_**. Can you update all instances similar to this one, and simplify the code?

peng: OK, I will simplify the code.

peng: Done!
I have simplified the code.

.changeset/fluffy-readers-add.md
```diff
@@ -0,0 +1,6 @@
+---
+"@astrojs/internal-helpers": minor
+"astro": minor
+---
+
+Add multiple CDN option, `build.assetsPrefix` also add object type for can config multiple CDN assets prefix.
```
VoxelMC, line 6:
```suggestion
Adds the option to pass an object to `build.assetsPrefix`. This allows for the use of multiple CDN prefixes based on the target filetype.
```
Great first starting changeset!

The changesets in Astro usually follow the following format:
`this PR` \<verb>s `...`
This is just to ensure consistency in the changesets!
Thank you for writing this well to start out with! :)

ematipico (changes requested):

packages/astro/src/core/create-vite.ts
```diff
@@ -309,6 +310,6 @@ function isCommonNotAstro(dep: string): boolean {
 	);
 }
 
-function stringifyForDefine(value: string | undefined): string {
+function stringifyForDefine(value: string | undefined | object): string {
```
ematipico, line 313: Why was this changed?

packages/internal-helpers/src/path.ts
```diff
@@ -80,6 +80,11 @@ export function removeFileExtension(path: string) {
 	return idx === -1 ? path : path.slice(0, idx);
 }
 
+export function getFileExtension(path: string) {
+	let idx = path.lastIndexOf('.');
+	return idx === -1 ? '' : path.slice(idx+1);
+}
```
ematipico, line 86: Node provides a functionality to retrieve the file extension from a string: https://nodejs.org/dist/latest-v20.x/docs/api/path.html#pathextnamepath

sarah11918 (dismissed):
Just putting a "docs block" on this one because the documentation changes are needed to the config reference source file in this repo, not in a separate docs PR.

I will address the content of the docs material in https://github.com/withastro/docs/pull/6982 initially, but in fact those changes will need to be made to https://github.com/withastro/astro/blob/main/packages/astro/src/%40types/astro.ts and that Docs PR will be closed.

sarah11918:
DOCS UPDATE:

The docs content has been initially reviewed in the docs PR, and is ready to be added here as an edit to `types.ts`. It may still require more editing (and technical review from devs evaluating this PR), but it is a good format to now bring from the other Docs PR to here to continue working on it.

packages/astro/src/core/create-vite.ts
```diff
@@ -205,9 +206,9 @@ export async function createVite(
 	const assetsPrefix = settings.config.build.assetsPrefix;
 	if (assetsPrefix) {
 		commonConfig.experimental = {
-			renderBuiltUrl(filename, { type }) {
+			renderBuiltUrl(filename, { type, hostType }) {
```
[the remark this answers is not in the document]
peng: File type, `hostType` is vite `experimental. renderBuiltUrl` option function param.
Docs: https://v4.vitejs.dev/guide/build.html#advanced-base-options

packages/astro/src/@types/astro.ts
```diff
@@ -838,7 +838,7 @@ export interface AstroUserConfig {
 		 * }
 		 * ```
 		 */
-		assetsPrefix?: string;
+		assetsPrefix?: AssetsPrefix;
```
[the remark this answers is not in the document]
peng: I have completed the document.
Docs PR: https://github.com/withastro/docs/pull/6982

packages/astro/src/core/create-vite.ts
peng: image
Here stringify `assetsPrefix`

.changeset/fluffy-readers-add.md
peng: Thanks I will solve it

packages/internal-helpers/src/path.ts
peng: I have replaced it with `extname`

packages/astro/src/core/config/schema.ts
```diff
@@ -138,7 +138,7 @@ export const AstroConfigSchema = z.object({
 				.default(ASTRO_CONFIG_DEFAULTS.build.server)
 				.transform((val) => new URL(val)),
 			assets: z.string().optional().default(ASTRO_CONFIG_DEFAULTS.build.assets),
-			assetsPrefix: z.string().optional(),
+			assetsPrefix: z.string().optional().or(z.record(z.string()).optional()),
```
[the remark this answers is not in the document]
peng: Write like this `assetsPrefix: z.string().or(z.record(z.string()).optional())` or finish docs in file `packages/astro/src/@types/astro.ts` ?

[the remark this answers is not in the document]
ematipico: The thing is, when `assetsPrefix` is defined as an object, the key `defaultAssetsPrefix` must be mandatory otherwise the logic breaks. 

If you're not able to define a schema with zod, then you should use `refine` to check the presence of `defaultAssetsPrefix`. If the isn't, we should emit an error.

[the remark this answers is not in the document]
peng: `defaultAssetsPrefix` is optional, If key `defaultAssetsPrefix ` does not exist，default value is empty string.(If `defaultAssetsPrefix` not set and no other keys are matched, an empty string will be returned.)

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -95,9 +96,10 @@ export default function assets({
 					}
 
 					// Rollup will copy the file to the output directory, this refer to this final path, not to the original path
-					const finalOriginalImagePath = (
-						isESMImportedImage(options.src) ? options.src.src : options.src
-					).replace(settings.config.build.assetsPrefix || '', '');
+					const ESMImportedImageSrc = isESMImportedImage(options.src) ? options.src.src : options.src;
+					const fileExtension = extname(ESMImportedImageSrc);
+					const pf = getAssetsPrefix(fileExtension, settings.config.build.assetsPrefix);
```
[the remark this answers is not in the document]
peng: I think would not be prone to bugs.

`defaultAssetsPrefix` is default assets prefix cdn value set. You can set the default values of other files by setting the value of `defaultAssetsPrefix`. If `defaultAssetsPrefix` is not set, when the file extension is not found in the `assetsPrefix` object, it will be set to the empty string `''` by default.

packages/astro/src/@types/astro.ts
```diff
@@ -881,11 +881,13 @@ export interface AstroUserConfig {
 		/**
 		 * @docs
 		 * @name build.assetsPrefix
-		 * @type {string}
+		 * @type {string | Record<string, string>}
```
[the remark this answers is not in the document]
peng: This requires listing all file extension, and I'm afraid some will be missed.

packages/astro/src/core/config/schema.ts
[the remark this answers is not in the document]
peng: Or write like this？

```
assetsPrefix: z.string().optional().or(z.record(z.enum(['defaultAssetsPrefix']).or(z.string()), z.string()).optional())
```

packages/astro/src/@types/astro.ts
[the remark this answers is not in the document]
bluwy: I don't think that requires listing all the file extensions. Following my comment below, if we hardcode and know the type of asset we're dealing with, e.g. `js`, then the user only needs to specify `{ js: '...', ... }` to set the prefix for a specific type.

The idea would be that, for nonjs/non-css assets, we'll fallback to the `fallback` key for now. In the future, we can add more keys like `image` for example, if we want to make it more flexible.

packages/astro/src/assets/vite-plugin-assets.ts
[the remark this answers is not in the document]
bluwy: I think it would. Let's say we decide to rename the output from `.js` to `.mjs` today, that would mean the existing configuration won't work anymore. If someone wants to be robust, they have to configure `{ js: '...', mjs: '...', cjs: '...' }` which I don't think is a nice DX. We can help here by grouping them up as a single asset type.

packages/astro/src/@types/astro.ts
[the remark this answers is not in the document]
peng: Do you mean that we currently only provide less file extension support and will support more flexible configuration methods in the future? 

eg: Currently we only support js, css file have their own CDN address. Other file use `fallback` key address ?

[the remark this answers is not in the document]
ematipico: Yes, that's the idea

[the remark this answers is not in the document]
bluwy: Yeah I was thinking that.  I guess I see your point now. Allowing any extensions gives the most flexibility if for example, they want to put fonts on a different prefix too.

In that case, let's go with your idea then. Thanks for discussing it with me! I think we should also document that the keys are based on extensions, and users should set multiple `js` or `mjs` etc if needed.

---

Also, another idea might be to make it function-form, e.g. `(ext: string) => string`. It could be easier to debug if the user missed configuring certain extensions, but it's up to you if you think whichever is better.

ematipico:
Heads up @peng and @bluwy 

I pushed a commit that applies the following changes:
- refines the schema to have `fallback` as a mandatory key inside the object
- renames `defaultSchemaPrefix` to `fallback`

packages/astro/src/@types/astro.ts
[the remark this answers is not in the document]
ematipico: > Also, another idea might be to make it function-form, e.g. (ext: string) => string

That was the initial implementation, but that wouldn't work in SSR. We can revisit it once we ship the object configuration. What do you think? I think the feature is fine as is for now

ematipico (approved):

peng:
> I pushed a commit that applies the following changes:
> 
> * refines the schema to have `fallback` as a mandatory key inside the object
> * renames `defaultSchemaPrefix` to `fallback`

OK, Thank you ! I change the docs.

packages/astro/src/@types/astro.ts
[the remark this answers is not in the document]
bluwy: Hmm good point. Let's keep it as is then.

peng:
When will it be merged?

ematipico:
> When will it be merged?

Monday

packages/astro/src/@types/astro.ts
```diff
@@ -881,12 +881,14 @@ export interface AstroUserConfig {
...
 		 * @default `undefined`
 		 * @version 2.2.0
 		 * @description
 		 * Specifies the prefix for Astro-generated asset links. This can be used if assets are served from a different domain than the current site.
 		 *
+		 * If this value is a `string`,
+		 *
 		 * For example, if this is set to `https://cdn.example.com`, assets will be fetched from `https://cdn.example.com/_astro/...` (regardless of the `base` option).
```
bluwy, line 892: This comment isn't complete. I think it could be:
```
If a string is passed, e.g. `https://cdn.example.com`, assets will be ...
```

packages/astro/src/@types/astro.ts
```diff
@@ -899,8 +901,29 @@ export interface AstroUserConfig {
...
+		 *     assetsPrefix: {
+		 *       'js': 'https://js.cdn.example.com',
+		 *       'css': 'https://js.cdn.example.com',
+		 *       'fallabck': 'https://cdn.example.com'
+		 *     }
+		 *   }
+		 * }
+		 * ```
+		 *
+		 * Object key is file extension, value is CDN assets prefix path. `fallabck` is default assets prefix cdn value set and it is a mandatory key. You can set the default values of other files by setting the value of `fallabck`.
```
bluwy, line 923:
```suggestion
		 * If an object of `Record<string, string>` is passed, you can individually map certain file extensions to use a specific assets prefix. This can be useful for multiple CDN scenarios, for example:
		 *
		 * ```js
		 * {
		 *   build: {
		 *     assetsPrefix: {
		 *       js: 'https://js.cdn.example.com',
		 *       mjs: 'https://js.cdn.example.com',
		 *       css: 'https://css.cdn.example.com',
		 *       fallback: 'https://cdn.example.com'
		 *     }
		 *   }
		 * }
		 * ```
		 *
		 * The `fallback` key must be set to so unspecified file extensions will fallback to this value.
```

I think this jsdoc will be slightly clearer and concise.

packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -33,3 +34,15 @@ export function hashTransform(
 	);
 	return shorthash(deterministicString(hashFields));
 }
+
+export function getAssetsPrefix(fileType: string, assetsPrefix?: AssetsPrefix): string {
+	if (!assetsPrefix) return '';
+	if (typeof assetsPrefix === 'string') return assetsPrefix;
+	fileType = fileType[0] === '.' ? fileType.slice(1) : fileType;
```
bluwy, line 41: nit: It looks like every usage of `getAssetsPrefix` passes `fileType` as `.js` or `.css` (for example). Perhaps we should name the param `fileExt` instead and assume that the string will always have a leading `.`? (I usually prefer doing the least work possible unless really needed.

I think we should be consistent how the "file type" is computed too, in this PR, there's different patterns:

```js
const fileExtension = extname(...)
getAssetsPrefix(fileExtension, ...)
```

```js
const fileType = extname(...)
getAssetsPrefix(fileType, ...)
```

I think we should stick with the `fileExtension` name, or `fileExt`.

packages/astro/src/assets/utils/transformToPath.ts
```diff
@@ -33,3 +34,15 @@ export function hashTransform(
...
+	if (assetsPrefix[fileType]) {
+		return assetsPrefix[fileType];
+	} else if (assetsPrefix.fallback) {
+		return assetsPrefix.fallback;
+	}
+	return '';
```
bluwy, line 47: I think we can directly return `assetsPrefix.fallback` here because we had validated in the `schema.ts` that it should always exist.

bluwy (changes requested):

packages/astro/src/@types/astro.ts
```diff
@@ -881,12 +881,14 @@ export interface AstroUserConfig {
 		/**
 		 * @docs
 		 * @name build.assetsPrefix
-		 * @type {string}
+		 * @type {string | Record<string, string>}
 		 * @default `undefined`
 		 * @version 2.2.0
 		 * @description
 		 * Specifies the prefix for Astro-generated asset links. This can be used if assets are served from a different domain than the current site.
```
sarah11918, line 888:
```suggestion
		 * Specifies the prefix for Astro-generated asset links. This can be used if assets are served from a different domain than the current site.
		 *
		 * This requires uploading the assets in your local `./dist/_astro` folder to a corresponding `/_astro/` folder to the remote domain.
```

packages/astro/src/@types/astro.ts
```diff
@@ -881,12 +881,14 @@ export interface AstroUserConfig {
...
 		 *
+		 * If this value is a `string`,
+		 *
 		 * For example, if this is set to `https://cdn.example.com`, assets will be fetched from `https://cdn.example.com/_astro/...` (regardless of the `base` option).
 		 * You would need to upload the files in `./dist/_astro/` to `https://cdn.example.com/_astro/` to serve the assets.
```
sarah11918, line 893:
```suggestion
		 * To fetch all assets uploaded to the same domain (e.g. `https://cdn.example.com/_astro/...`), set `assetsPrefix` to the root domain as a string (regardless of your `base` configuration):
```

packages/astro/src/@types/astro.ts
```diff
@@ -881,12 +881,14 @@ export interface AstroUserConfig {
...
 		 * The process varies depending on how the third-party domain is hosted.
```
sarah11918, line 894:
```suggestion
```
(Also, I can't suggest editing the line below this, so in fact what I would prefer is that the two existing lines are removed. Then, only one line after the code sample with a short note about the "this might vary". Something like:
```
* To fetch all assets uploaded to the same domain (e.g. `https://cdn.example.com/_astro/...`), set `assetsPrefix` to the root domain as a string (regardless of your `base` configuration):
[code sample]
* (Note: This process may vary depending on how the third-party domain is hosted.)
```

packages/astro/src/@types/astro.ts
sarah11918, line 923:
``````suggestion
		 * **Added in 4.5.0**
		 *
		 * You can also pass an object to `assetsPrefix` to specify a different domain for each file type.
		 * In this case, a `fallback` property is required and will be used by default for any other files.
		 *
		 * ```js
		 * {
		 *   build: {
		 *     assetsPrefix: {
		 *       'js': 'https://js.cdn.example.com',
		 *       'css': 'https://js.cdn.example.com',
		 *       'fallback': 'https://cdn.example.com'
		 *     }
		 *   }
		 * }
		 * ```
``````
And here's my suggestion!  :smile:

**Review by sarah11918 — COMMENTED** [review]
I know we're all in this file at the same time, so here are some suggestions from me to consider for the docs!

File: packages/astro/src/@types/astro.ts (hunk 10 above)
Comment (bluwy) [reply packages/astro/src/@types/astro.ts:892]: Resolving this in favour of Sarah's review below

File: packages/astro/src/@types/astro.ts (hunk 11 above)
Comment (bluwy) [reply packages/astro/src/@types/astro.ts:923]: Also resolving this in favour of Sarah's review below

Comment (bluwy) [reply packages/astro/src/@types/astro.ts:923]: Moving my review here so it's less confusing 😄 I think we should also show the `'mjs': 'https://js.cdn.example.com/'` example here. So it implies that you'd set `js` and `mjs` if needed.

Comment (peng) [reply packages/astro/src/@types/astro.ts:923]: > And here's my suggestion! 😄

Your suggestion is great !  I agree !

**Comment by bluwy** [comment]
@ematipico It doesn't look like my comments above have been addressed yet.

**Comment by ematipico** [comment]
> @ematipico It doesn't look like my comments above have been addressed yet.

Which ones? Sorry, it's becoming confusing 😅 I would be grateful if you can tell me what I have need to address

**Review by bluwy — APPROVED** [review]
Thanks for picking this up Ema!

**Review by sarah11918 — COMMENTED** [review]
Looks great! Just some suggestions where I think we can use more user-friendly language in the changeset. See what you think!

File: .changeset/fluffy-readers-add.md [hunk 17]
```diff
@@ -0,0 +1,26 @@
...
+```js
+// astro.config.mjs
+import { defineConfig } from "astro/config"
+
+export default defineConfig({
+    build: {
+        assetsPrefix: {
+            'js': "https://js.cdn.example.com",
+            'mjs': "https://js.cdn.example.com", // if you have .mjs files, you must add a new entry like this
+            "png": "https://images.cdn.example.com",
```
sarah11918, line 21:
```suggestion
            'png': "https://images.cdn.example.com",
```
Not sure why this one was different :joy:

sarah11918 (approved):
Approving for docs!

PR #10190, opened 2024-02-22
chore: import sorting for test folder and e2e folder

ematipico:
## Changes

This PR applies import sorting for the `test` folders and the `e2e` folders.

I will make a later PR to ignore the commit from git.

## Testing

Current CI should pass

### DIFF
--- a/packages/astro/e2e/fixtures/astro-component/astro.config.mjs
+++ b/packages/astro/e2e/fixtures/astro-component/astro.config.mjs
@@ -1,5 +1,5 @@
-import { defineConfig } from 'astro/config';
 import preact from '@astrojs/preact'
+import { defineConfig } from 'astro/config';
 
 // https://astro.build/config
 export default defineConfig({
--- a/biome.json
+++ b/biome.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://biomejs.dev/schemas/1.5.3/schema.json",
   "files": {
-    "include": ["*.test.js"],
+    "include": ["test/**", "e2e/**"],
     "ignore": ["vendor"]
   },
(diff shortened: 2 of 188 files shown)


packages/astro/e2e/fixtures/astro-component/astro.config.mjs
```diff
@@ -1,5 +1,5 @@
-import { defineConfig } from 'astro/config';
 import preact from '@astrojs/preact'
+import { defineConfig } from 'astro/config';
```
bluwy, line 2: I think the only slightly odd part of the formatting is that `defineConfig` is now below, but I'm not sure if it's feasible to make a rule to fix this. I'm not too bummed by it but bringing this up in case others might think the same.

bluwy (approved):
One thing I noticed is that `import { defineConfig } from 'astro/config';` is sorted below integrations likek

packages/astro/e2e/fixtures/astro-component/astro.config.mjs
ematipico: Makes sense. Here's the algorithm: https://biomejs.dev/analyzer/#how-imports-are-sorted

bluwy: I think the sorting algorithm is fine, but if there's an option to tweak certain orderings, I think that's a great bonus. 
e.g. https://github.com/IanVS/prettier-plugin-sort-imports?tab=readme-ov-file#1-put-specific-dependencies-at-the-top. I used to use this prettier plugin and it has a few options to configure the ordering.

matthewp (approved):

PR #10193, opened 2024-02-22
feat(i18n): manual routing

ematipico:
## Changes

Closes PLT-1865

This PR adds **i18n manual routing**. This enabled using the the following configuration:

```diff
export default defineConfig({
	i18n: {
		defaultLocale: "fr",
		locales: ["fr", "en"],
+		routing: "manual"
	}
})
```

With this kind of configuration, the `astro:i18n` virtual modules **exports** new functions that users can use in their pages, and specifically in their middleware:

- `redirectToDefaultLocale`, to redirect to the default locale that was configured
- `notFound`, to return a 404
- `redirectToFallback`, to manually trigger the Astro i18n fallback

Another function that is exported is `middleware`. This function is actually useful in case the user wants to use the **Astro i18n middleware**, but still being able to *hack* it and provide some escape hatches/exceptions to some routes. Also, this `middleware` function accepts the majority of the option of the Astro i18n middleware. Eventually, users can do something like this, in their `middleware.js` file:

```js
import { middleware } from "astro:i18n";
import { defineMiddleware, sequence } from "astro:middleware"

const myMiddleware = defineMiddleware(async (context, next) => {
	// custom logic here
	const response = await next();
	
	// fallback logic that is run **after** the astro i18n middleware
	return response;
})

export const onRequest = sequence(myMiddleware, middleware({
	redirectToDefaultLocale: false,
	prefixDefaultLocale: true
}))
```

Technical changes:
- the new functions export a `noop` function that emits an error if they didn't opt-in the manual routing
- the middleware plugin emits an error if the user opt-in the feature, but they don't have a middleware file. However,  I can easily change the idea in case you think this is too much. Integrations can inject their middleware functions after all
- when the user opt-in the function, we don't inject our internal middleware
- I refactored our internal middleware. I took shared functions and lifted them, so they can be used inside our internal middleware and the user middleware

## Testing

I added new tests

## Docs

https://github.com/withastro/docs/pull/7749


packages/astro/test/i18n-routing-manual-with-default-middleware.test.js
```diff
@@ -0,0 +1,113 @@
...
+	});
+
+	it('should return a 200 because the custom middleware allows it', async () => {
+		const response = await fixture.fetch('/about');
+		assert.equal(response.status, 200);
+		const text = await response.text();
+		assert.equal(text.includes('ABOUT ME'), true);
+	});
+	//
+	// it('should render a i18n route', async () => {
```
matthewp, line 39: should these be removed?

packages/astro/src/virtual-modules/i18n.ts
```diff
@@ -191,3 +201,161 @@ export const getPathByLocale = (locale: string) => I18nInternals.getPathByLocale
...
+ * export const onRequest = sequence(customLogic, middleware({
+ * 	prefixDefaultLocale: true,
+ * 	redirectToDefaultLocale: false
+ * }))
+ *
+ * ```
+ */
+export const middleware =
+	i18n?.routing === 'manual'
+		? (customOptions?: Options) => {
```
matthewp, line 347: I find this ternary and nesting to be a little difficult to read. Would it make sense instead to have this function be defined as `manualMiddleware` or something and then make the conditional a regular `if` statement?

packages/astro/src/i18n/index.ts
```diff
@@ -244,3 +276,109 @@ class Unreachable extends Error {
...
+	return function (context: APIContext, statusCode?: ValidRedirectStatus) {
+		if (shouldAppendForwardSlash(trailingSlash, format)) {
+			return context.redirect(`${appendForwardSlash(joinPaths(base, defaultLocale))}`, statusCode);
+		} else {
+			return context.redirect(`${joinPaths(base, defaultLocale)}`, statusCode);
+		}
+	};
+}
+
+export function notFound({ base, locales }: MiddlewarePayload) {
```
matthewp, line 306: are these new exported functions intended to be used as helpers?

ematipico: Yes, as per description:
> I refactored our internal middleware. I took shared functions and lifted them, so they can be used inside our internal middleware and the user middleware

I thought it was a good way to dogfooding the helpers that we expose to the users, because we use them too.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1067,6 +1067,18 @@ export const MissingIndexForInternationalization = {
 	hint: (src: string) => `Create an index page (\`index.astro, index.md, etc.\`) in \`${src}\`.`,
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * Some internationalization functions can't be exposed unless the default routing is disabled
+ */
+export const IncorrectStrategy = {
+	name: 'IncorrectStrategy',
```
Princesseuh, line 1076: I would suggest using a name more specific to i18n, perhaps just `Incorrecti18nStrategy` would work

.changeset/little-hornets-give.md
```diff
@@ -0,0 +1,34 @@
...
+        defaultLocale: "fr",
+        routing: "manual"
+    }
+})
+```
+
+```js
+// middleware.js
+import { redirectToDefaultLocale } from "astro:i18n";
+export const onRequest = defineMiddleawre(async (context, next) => {
```
Princesseuh, line 22:
```suggestion
export const onRequest = defineMiddleware(async (context, next) => {
```

packages/astro/src/core/config/schema.ts
```diff
@@ -386,21 +386,24 @@ export const AstroConfigSchema = z.object({
...
-					.refine(
-						({ prefixDefaultLocale, redirectToDefaultLocale }) => {
-							return !(prefixDefaultLocale === false && redirectToDefaultLocale === false);
-						},
-						{
-							message:
-								'The option `i18n.redirectToDefaultLocale` is only useful when the `i18n.prefixDefaultLocale` is set to `true`. Remove the option `i18n.redirectToDefaultLocale`, or change its value to `true`.',
-						}
-					),
+					.enum(['manual'])
```
Princesseuh, line 389:
```suggestion
					.literal('manual')
```
maybe? Though I figure you used enum because you think there might be more choices in the future.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1067,6 +1067,18 @@ export const MissingIndexForInternationalization = {
 	hint: (src: string) => `Create an index page (\`index.astro, index.md, etc.\`) in \`${src}\`.`,
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * Some internationalization functions can't be exposed unless the default routing is disabled
```
Princesseuh, line 1073: Additional context in both the error messages would be great, at least some links in the `@see` thing (see other errors for examples)

packages/astro/src/i18n/index.ts
```diff
@@ -244,3 +276,109 @@ class Unreachable extends Error {
...
+				});
+			}
+		}
+
+		return undefined;
+	};
+}
+
+export type UseFallback = (context: APIContext, response: Response) => Response | undefined;
+export function useFallback({ fallback, locales, defaultLocale, routing }: MiddlewarePayload) {
```
bluwy, line 337: Question about `useFallback`: The `use` prefix is kinda overloaded by React to mean its React hooks. It seems like what this function does is that it redirects to a fallback page, would it be more accurate as `redirectToFallback` so it's also consistent with `redirectToDefaultLocale`?

packages/astro/src/i18n/index.ts
Princesseuh: Might be useful to add a comment to say those functions are public to users

packages/astro/src/i18n/index.ts
```diff
@@ -244,3 +276,109 @@ class Unreachable extends Error {
...
+	return function (context: APIContext, response: Response): Response | undefined {
```
bluwy, line 338: I noticed the `response` param here is required, while in `virtual-modules/i18n.ts`, it's marked as optional. Is there a typo here?

packages/astro/src/virtual-modules/i18n.ts
```diff
@@ -191,3 +209,184 @@ export const getPathByLocale = (locale: string) => I18nInternals.getPathByLocale
...
+	});
+} else {
+	useFallback = noop('useFallback');
+}
+
+// @ematipico: This is a helper type because, because I don't know how to extract from TS
+type Options = {
+	prefixDefaultLocale?: boolean;
+	redirectToDefaultLocale?: boolean;
+};
```
Princesseuh, line 342: I'm not 100% sure what you're trying to extract here, can't you just do `AstroConfig['i18n'][option]`?

packages/astro/src/core/config/schema.ts
ematipico: Oh, no. I wasn't aware there was a `literal` API! Thank you

packages/astro/src/virtual-modules/i18n.ts
ematipico: No, because the type is more complex than that, and it's above my skills. 

`AstroConfig['i18n']['routing']` can be undefined, and I can manage by using `NonNullable`, although `routing` is now an object OR a literal `"manual"`. I just want to extract the object.

packages/astro/src/i18n/index.ts
ematipico: It was a typo, good catch. It should be mandatory.

packages/astro/src/i18n/index.ts
ematipico: Done in [`e75b46c` (#10193)](https://github.com/withastro/astro/pull/10193/commits/e75b46c01fe5a3284a59e91c15a2cb79cd5dfc6b)

packages/astro/src/i18n/middleware.ts
```diff
@@ -153,58 +132,47 @@ export function createI18nMiddleware(
...
-				}
-
-				return context.redirect(newPathname);
-			}
-		}
-
-		return response;
+		return _redirectToFallback(context, response);
+
+		// if (response.status >= 300 && fallback) {
```
bluwy, line 137: This part is commented out

packages/astro/src/core/base-pipeline.ts
```diff
@@ -48,9 +48,13 @@ export abstract class Pipeline {
 		 */
 		readonly site = manifest.site ? new URL(manifest.site) : undefined
 	) {
-		this.internalMiddleware = [
-			createI18nMiddleware(i18n, manifest.base, manifest.trailingSlash, manifest.buildFormat),
-		];
+		this.internalMiddleware = [];
+		// We do use our middleware only if the user isn't using the manual setup
+		if (i18n?.strategy !== 'manual') {
```
lilnasy, line 53: We also have a similar guard within `createI18nMiddleware`. Does it make sense to colocate the two?

matthewp (approved):

.changeset/little-hornets-give.md
```diff
@@ -0,0 +1,34 @@
+---
+"astro": minor
+---
+
+Astro now allows to create your know i18n middleware:
```
sarah11918, line 5:
```suggestion
Adds a new i18n routing option `manual` to allow you to write your own i18n middleware:
```

.changeset/little-hornets-give.md
```diff
@@ -0,0 +1,34 @@
...
+// astro.config.mjs
+export default defineConfig({
+    i18n: {
+        locales: ["en", "fr"],
+        defaultLocale: "fr",
+        routing: "manual"
+    }
+})
+```
+
```
sarah11918, line 18:
```suggestion

Adding `routing: "manual"` to your i18n config disables Astro's own i18n middleware and provides you with helper functions to write your own: `redirectToDefaultLocale`, `notFound`, and `redirectToFallback`:

```

.changeset/little-hornets-give.md
```diff
@@ -0,0 +1,34 @@
...
+    } else {
+        return redirectToDefaultLocale(context, 302);  
+    }
+})
+```
+
+When `routing: "manual"` is provided, the virtual module `astro:i18n` exports new functions:
+- `redirectToDefaultLocale`
+- `notFound`
+- `redirectToFallback`
```
sarah11918, line 34:
``````suggestion
Also adds a `middleware` function that manually creates Astro's i18n middleware. This allows you to extend Astro's i18n routing instead of completely replacing it. Run `middleware` in combination with your own middleware, using the `sequence` utility to determine the order:

```js title="src/middleware.js"
import {defineMiddleware, sequence} from "astro:middleware";
import { middleware } from "astro:i18n"; // Astro's own i18n routing config

export const userMiddleware = defineMiddleware();

  userMiddleware,
  middleware({
    redirectToDefaultLocale: false,
    prefixDefaultLocale: true
  })
)
```
``````
This is a super stripped down version of the example in docs, but I think is worth showing here too as it's a feature of this PR!

File: packages/astro/src/core/errors/errors-data.ts [hunk 16]
```diff
@@ -1067,6 +1067,21 @@ export const MissingIndexForInternationalization = {
 	hint: (src: string) => `Create an index page (\`index.astro, index.md, etc.\`) in \`${src}\`.`,
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * Some internationalization functions can't be exposed unless the i18n routing is disabled by assigning to `i18n.routing` the value "manual"`.
```
sarah11918, line 1073:
```suggestion
 * Some internationalization functions are only available when Astro's own i18n routing is disabled by the configuration setting `i18n.routing: "manual"`.
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1067,6 +1067,21 @@ export const MissingIndexForInternationalization = {
...
+ *
+ * @see
+ * - [`i18n` routing](https://docs.astro.build/en/guides/internationalization/#routing)
+ */
+export const IncorrectStrategyForI18n = {
+	name: 'IncorrectStrategyForI18n',
+	title: "You can't use the current function with the current strategy",
+	message: (functionName: string) =>
+		`The function \`${functionName}\' can't be used because it can only be used when the \`i18n.routing.strategy\` is set to \`"manual"\`.`,
```
sarah11918, line 1082:
```suggestion
		`The function \`${functionName}\' can only be used when the \`i18n.routing.strategy\` is set to \`"manual"\`.`,
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1079,6 +1094,18 @@ export const NoPrerenderedRoutesWithDomains = {
 		`Static pages aren't yet supported with multiple domains. If you wish to enable this feature, you have to disable prerendering for the page ${component}`,
```
sarah11918, line 1094:
```suggestion
		`Static pages aren't yet supported with multiple domains. To enable this feature, you must disable prerendering for the page ${component}`,
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1079,6 +1094,18 @@ export const NoPrerenderedRoutesWithDomains = {
...
+/**
+ * @docs
+ * @description
+ * Astro throws an error if the user enables manual routing, but it doesn't have a middleware file.
+ */
+export const MissingMiddlewareForInternationalization = {
+	name: 'MissingMiddlewareForInternationalization',
+	title: 'Enabled manual internationalization routing without having a middleware.',
+	message:
+		"You enabled manual routing for internationalization via `i18n.routing: 'manual'`, but you must have a `middleware` file in order to use this feature.",
```
sarah11918, line 1106:
```suggestion
		"Your configuration setting `i18n.routing: 'manual'` requires you to provide your own i18n `middleware` file.",
```

packages/astro/src/@types/astro.ts
```diff
@@ -1497,66 +1497,68 @@ export interface AstroUserConfig {
...
-			 * @default `"pathname"`
-			 * @version 3.7.0
-			 * @description
-			 *
-			 * - `"pathname": The strategy is applied to the pathname of the URLs
-			 */
-			strategy?: 'pathname';
-		};
+		routing?:
+			| 'manual'
```
sarah11918, line 1501: Just checking, are we waiting for a docs entry for "manual" here? This should have its own entry that shows up in the docs config reference file.

For reference, this is the text from the docs guide page, to help craft something:

> When this option is enabled, Astro will **disable** its i18n middleware so that you can implement your own custom logic. No other `routing` options (e.g. `prefixDefaultLocale`) may be configured with `routing: "manual"`.
> 
> You will be responsible for writing your own routing logic, or executing Astro's i18n middleware manually alongside your own.

sarah11918 (commented):
What a cool feature and exciting release! Most of these are small suggestions @ematipico but I've noted that I think we're missing some JSDoc for the configuration reference page for "manual"?

Looking so good though!

packages/astro/src/@types/astro.ts
ematipico: That's done

packages/astro/src/@types/astro.ts
```diff
@@ -1497,66 +1497,80 @@ export interface AstroUserConfig {
...
-			prefixDefaultLocale?: boolean;
-
-			/**
 			 * @docs
-			 * @name i18n.routing.redirectToDefaultLocale
-			 * @kind h4
-			 * @type {boolean}
-			 * @default `true`
-			 * @version 4.2.0
+			 * @name i18n.routing
```
sarah11918, line 1505: Just checking:
```suggestion
			 * @name i18n.routing.manual
```

sarah11918 (approved):
Admittedly, I have the hardest time figuring out what the JSDoc is going to look like ahead of time (I'm hoping the syntax is correct to get the new `manual` option nested under `i18n.routing` along with the other object!  :sweat_smile: )

But all the text is approved by docs! Great work, @ematipico ! :partying_face:

packages/astro/src/@types/astro.ts
```diff
@@ -1497,66 +1497,80 @@ export interface AstroUserConfig {
...
+					 * @name i18n.routing.strategy
+					 * @type {"pathname"}
+					 * @default `"pathname"`
+					 * @version 3.7.0
+					 * @description
+					 *
+					 * - `"pathname": The strategy is applied to the pathname of the URLs
+					 */
+					strategy?: 'pathname';
+			  };
```
sarah11918, line 1573: I can't make the suggestion in this file, but **line 1609** should also be updated to remove links to the guide page:

```
Both page routes built and URLs returned by the `astro:i18n` helper functions [`getAbsoluteLocaleUrl()`](https://docs.astro.build/en/reference/api-reference/#getabsolutelocaleurl) and [`getAbsoluteLocaleUrlList()`](https://docs.astro.build/en/reference/api-reference/#getabsolutelocaleurllist) will use the options set in `i18n.domains`.
```

Issue #10187, opened 2024-02-21
Bundled `<script>`s do not work with Vercel ISR in development


bholmesdev:
### Astro Info

```block
Astro                    v4.4.0
Node                     v18.18.0
System                   macOS (arm64)
Package Manager          unknown
Output                   server
Adapter                  @astrojs/vercel/serverless
Integrations             none
```

### Describe the Bug

Bundled `<script>` tags (without using `is:inline`) fail to load in development when using Vercel serverless with ISR. Changing the `prerender` property and changing the `mode` to `hybrid` have no effect.

The following is logged to the client console:
```ts
transition.js?v=c0e8ccae:3 Uncaught SyntaxError: The requested module '/node_modules/.pnpm/cssesc@3.0.0/node_modules/cssesc/cssesc.js?v=c0e8ccae' does not provide an export named 'default' (at transition.js?v=c0e8ccae:3:8)
```

Note: this error is _also_ logged when bundled `<style>` tags are used (though the styles load successfully). This makes me think any file with `.astro?lang=js|css` are pulling in runtime internals to the client unexpectedly. 

### What's the expected result?

`<script>` and `<style>` load successfully without a console error.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-e9exm5?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #10194, opened 2024-02-22
Fix cssesc from breaking browser code

matthewp:
## Changes

- Include `astro > cssesc` as an optimized dep
- Fixes https://github.com/withastro/astro/issues/9880
- Fixes https://github.com/withastro/astro/issues/10187
- Loading `astro:content` loads a lot of Astro's internal server-side code. This includes the server-side portion of View Transitions. This code depends on `cssesc` which is used to correctly escape `view-transition-name` properties to what is valid in CSS. This code is CommonJS and doesn't work in the browser unless optimized. This change makes it so that this module can be optimized.
  - Note that you should not be using `astro:content` in the browser. This is only being fixed because a significant number of people are doing so. This PR will be updated to warn.

## Testing

- Can't be tested in the monorepo, where dependencies are loaded differently.
- Tested manually with a test project: https://stackblitz.com/edit/github-gqosvq?file=src%2Fcontent%2Fconfig.ts,src%2Fpages%2Findex.astro

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/quick-bottles-march.md
+++ b/.changeset/quick-bottles-march.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Allow content collections to run in browser
--- a/.changeset/quick-bottles-march.md
+++ b/.changeset/quick-bottles-march.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Allow content collections to run in browser
+Fixes an issue related to content collections usage in browser context caused by `csssec`
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -122,11 +122,7 @@ export async function createVite(
 				`${srcDirPattern}pages/**/_*.{js,mjs,ts,mts}`, // Remaining JS/TS files prefixed with `_` (not endpoints)
 				`${srcDirPattern}pages/**/_*/**/*.{js,mjs,ts,mts}`, // Remaining JS/TS files within directories prefixed with `_` (not endpoints)
 			],
-			exclude: [
-				'astro:dev-toolbar',
-				'astro/virtual-modules/prefetch.js',
-				'node-fetch'
-			],
+			exclude: ['astro', 'node-fetch'],
 		},
 		plugins: [
 			configAliasVitePlugin({ settings }),
--- a/packages/astro/src/core/create-vite.ts
+++ b/packages/astro/src/core/create-vite.ts
@@ -122,7 +122,11 @@ export async function createVite(
 				`${srcDirPattern}pages/**/_*.{js,mjs,ts,mts}`, // Remaining JS/TS files prefixed with `_` (not endpoints)
 				`${srcDirPattern}pages/**/_*/**/*.{js,mjs,ts,mts}`, // Remaining JS/TS files within directories prefixed with `_` (not endpoints)
 			],
-			exclude: ['astro', 'node-fetch'],
+			exclude: [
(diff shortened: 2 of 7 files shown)


packages/astro/src/core/create-vite.ts
```diff
@@ -122,7 +122,11 @@ export async function createVite(
 				`${srcDirPattern}pages/**/_*.{js,mjs,ts,mts}`, // Remaining JS/TS files prefixed with `_` (not endpoints)
 				`${srcDirPattern}pages/**/_*/**/*.{js,mjs,ts,mts}`, // Remaining JS/TS files within directories prefixed with `_` (not endpoints)
 			],
-			exclude: ['astro', 'node-fetch'],
+			exclude: [
+				'astro:dev-toolbar',
```
Princesseuh, line 126: Is it potentially an issue that this only exclude parts of the dev toolbar?

matthewp: I'm not sure, what gets imported in the browser? is it a single entrypoint? Are there multiple? I tested this with the toolbar enabled and it works and does not work if I don't exclude this. Is there some scenarios I should try?

matthewp: Actually I think I can do this with a single `include` instead, testing now.

Princesseuh: The main entrypoint is at `astro/runtime/client/dev-toolbar/entrypoint.js`, that entrypoint imports `astro:dev-toolbar`

matthewp: Updated this so that `astro` is still excluded by default.

.changeset/quick-bottles-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Allow content collections to run in browser
```
ematipico, line 5: Is the changeset correct? I don't think this PR is around content collections

matthewp: Yeah it is, look at the stackblitz example.

ematipico: But title and description of the PR talk about csssec, I'm confused.

ematipico (commented):
I think the changeset should be more detailed. Sorry, I'm very confused by the whole PR

.changeset/quick-bottles-march.md
florian-lefebvre, line 5:
```suggestion
Fixes an issue related to content collections usage in browser context caused by `csssec`
```
Maybe something like this?

matthewp: View Transitions has both a server and client part to it. The client part has no issues, works fine. The server part depends on `cssesc`. You don't load that code in the client. Unless you import `astro:content`, in which that comes along for the ride.

matthewp:
> I think the changeset should be more detailed. Sorry, I'm very confused by the whole PR

What do you think it should say? I can give a better description of what is going on in the PR comment, but I don't think the intent of changesets is to explain how the bug is fixed, just what is fixes.

bluwy (approved):
Will have to sign soon, so I'll preemptively approve this first! I think making it work and adding a warning is a good middleground for now. And we can add on Linear to remove this for 5.0.

matthewp:
!preview allow-cc

matthewp:
Tested the preview release and it works:

Screen Shot 2024-02-22 at 11 08 14 AM

ematipico (approved):

PR #10196, opened 2024-02-22
prevent warning: `Astro.request.headers` is not available in "static" output mode

lilnasy:
## Changes
- Makes computing i18n attributes lazy again. Computing `preferredLanguage`, for example, involves reading the `Accept-Language` header.

## Testing
Existing tests should pass.

## Docs
Does not affect usage.

### DIFF
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -230,7 +230,7 @@ export class RenderContext {
 		// TODO: we are making two calls to computeCurrentLocale(). In various cases, one works and the other doesn't.
 		// Ideally, we could use `renderContext.pathname` which is intended to be the one true "file-system-matchable" path.
 		// - Arsh
-		return this.#currentLocale ??= computeCurrentLocale(url.pathname, locales, strategy, defaultLocale) ?? computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
+		return this.#currentLocale ??= computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
 	}
 
 	#preferredLocale: APIContext['preferredLocale'];
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -36,7 +36,8 @@ export class RenderContext {
 		readonly routeData: RouteData,
 		public status: number,
 		readonly cookies = new AstroCookies(request),
-		readonly params = getParams(routeData, pathname)
+		readonly params = getParams(routeData, pathname),
+		readonly url = new URL(request.url)
 	) {}
 
 	static create({
@@ -124,20 +125,18 @@ export class RenderContext {
 
 	createAPIContext(props: APIContext['props']): APIContext {
 		const renderContext = this;
-		const { cookies, i18nData, params, pipeline, request } = this;
-		const { currentLocale, preferredLocale, preferredLocaleList } = i18nData;
+		const { cookies, params, pipeline, request, url } = this;
 		const generator = `Astro v${ASTRO_VERSION}`;
 		const redirect = (path: string, status = 302) =>
 			new Response(null, { status, headers: { Location: path } });
 		const site = pipeline.site ? new URL(pipeline.site) : undefined;
-		const url = new URL(request.url);
 		return {
 			cookies,
-			currentLocale,
+			get currentLocale() { return renderContext.computeCurrentLocale() },
 			generator,
 			params,
-			preferredLocale,
-			preferredLocaleList,
+			get preferredLocale() { return renderContext.computePreferredLocale() },
+			get preferredLocaleList() { return renderContext.computePreferredLocaleList() },
 			props,
 			redirect,
 			request,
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/render-context.ts
```diff
@@ -282,42 +262,51 @@ export class RenderContext {
...
+			props,
+			redirect,
+			request,
+			response,
+			slots,
+			url,
+		};
+
+		// `Astro.self` is added by the compiler
+		return Object.setPrototypeOf(astroDynamicPartial, astroStaticPartial) as unknown as typeof astroStaticPartial & typeof astroDynamicPartial & Pick<AstroGlobal, 'self'>;
```
matthewp, line 280: Using setPrototypeOf causes code to be deoptimized. I don't know if that matters in this case, but this is a hot code-path so it very well might. Can we go back to setting `__proto__` instead?

lilnasy: That required unsafe assertions. I'll take another look.

matthewp:
Closes https://github.com/withastro/astro/issues/10212

packages/astro/src/core/render-context.ts
```diff
@@ -223,26 +222,28 @@ export class RenderContext {
...
 		const { defaultLocale, locales, strategy } = i18n;
-		return (this.#i18nData = {
-			currentLocale: computeCurrentLocale(routeData.route, locales, strategy, defaultLocale),
-			preferredLocale: computePreferredLocale(request, locales),
-			preferredLocaleList: computePreferredLocaleList(request, locales),
-		});
+		// TODO: we are making two calls to computeCurrentLocale(). In various cases, one works and the other doesn't.
+		// Ideally, we could use `renderContext.pathname` which is intended to be the one true "file-system-matchable" path.
+		// - Arsh
+		return this.#currentLocale ??= computeCurrentLocale(url.pathname, locales, strategy, defaultLocale) ?? computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
```
ematipico, line 233: As mentioned in the previously, we can't use `url.pathname` because in SSG we get a URL like `/en.html`, and we won't be able to provide the current locale.

We need to use `routeData.pathname` if it exists, and `routeData.route` as a fallback. Also. this code be cleaned up:

```js
return this.#currentLocale = computeCurrentLocale(url.pathname ?? routeData.route, locales, strategy, defaultLocale)
```

ematipico (changes requested):

packages/astro/src/core/render-context.ts
lilnasy, line 233:
```suggestion
		return this.#currentLocale ??= computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
```
```suggestion
		return this.#currentLocale ??= computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
```

packages/astro/src/core/render-context.ts
```diff
@@ -223,26 +222,28 @@ export class RenderContext {
...
+		if (!i18n) return;
 		const { defaultLocale, locales, strategy } = i18n;
-		return (this.#i18nData = {
-			currentLocale: computeCurrentLocale(routeData.route, locales, strategy, defaultLocale),
-			preferredLocale: computePreferredLocale(request, locales),
-			preferredLocaleList: computePreferredLocaleList(request, locales),
-		});
+		// TODO: we are making two calls to computeCurrentLocale(). In various cases, one works and the other doesn't.
+		// Ideally, we could use `renderContext.pathname` which is intended to be the one true "file-system-matchable" path.
+		// - Arsh
```
lilnasy, line 232:
```suggestion
```

packages/astro/src/core/render-context.ts
```diff
@@ -223,26 +222,25 @@ export class RenderContext {
...
+	computeCurrentLocale() {
+		const { url, pipeline: { i18n }, routeData } = this;
+		if (!i18n) return;
 		const { defaultLocale, locales, strategy } = i18n;
-		return (this.#i18nData = {
-			currentLocale: computeCurrentLocale(routeData.route, locales, strategy, defaultLocale),
-			preferredLocale: computePreferredLocale(request, locales),
-			preferredLocaleList: computePreferredLocaleList(request, locales),
-		});
+		return this.#currentLocale ??= computeCurrentLocale(routeData.route, locales, strategy, defaultLocale);
```
ematipico, line 230:
```suggestion
		return this.#currentLocale ??= computeCurrentLocale(routeData.pathname ?? routeData.route, locales, strategy, defaultLocale);
```

ematipico: Mmmh, I do recall using `.route`, and that caused of some issues... But if all tests pass, we should be fine. Weird though

ematipico (approved):

packages/astro/src/core/render-context.ts
lilnasy: They would fail if this code was used for the Astro global as well.  The tests for `Astro.currentLocale` require `url.pathname`.

PR #10197, opened 2024-02-22
fix(errors): Remove duplicate logging of errors

Princesseuh:
## Changes

Since about a week ago, we were logging errors twice wrongly

## Testing

Tested manually

## Docs

N/A


lilnasy (approved):
Looks great to me!

PR #10201, opened 2024-02-22
Introduce the db integration (prerelease)

matthewp:
## Changes

- This introduces a new package `@astrojs/db` that we've been working on. It's a prerelease and not ready for usage. It will change rapidly in breaking ways until the release.

## Testing

- Included!

## Docs

- Documentation in progress and will go out with the release.


.changeset/twenty-bears-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Adds a new top-level configuration option to the Astro configuration
```
matthewp, line 5: this isn't true any more, we removed that and instead augment in the integration.

Issue #8244, opened 2023-08-27
Middleware isn't invoked before rendering the default `404` page


ohansemmanuel:
### Astro info

```block
Astro version            v2.10.14
Package manager          npm
Platform                 darwin
Architecture             arm64
Adapter                  Couldn't determine.
Integrations             None or couldn't determine.
```

### What browser are you using?

Chrome

### Describe the Bug

Middleware isn't invoked before rendering the default `404` page. 

NB: I'm uncertain if this is a bug or the intended behaviour 

1. Consider the basic middleware below: 

```ts
import { defineMiddleware } from "astro/middleware";

export const onRequest = defineMiddleware((context, next) => {
  console.log("in middleware");
  return next();
});
```

2. Visit a random page route: e.g., `/uhoihoijoijoijoijoi` 

Result: The middleware is never invoked. No logs. 

3. Add a custom 404.astro page 

```ts 
// src/pages/404.astro
---
console.log("404 page");
---
```

Now, the middleware will be invoked before rendering the custom `404.astro` page. 

```bash 
#logs 
in middleware
404 page
```

### What's the expected result?

Consistently invoke the middleware for both custom and default 404 pages. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-i9px2m?file=src%2Fmiddleware%2Findex.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

syahzuan:
Similar to #8226 .

ematipico:
Duplicate of https://github.com/withastro/astro/issues/8226

The origin of the issue is the same.

ematipico:
Duplicate of https://github.com/withastro/astro/issues/8226

The origin of the issue is the same.

Issue #8226, opened 2023-08-25
Middleware doesn't redirect and serves 404 when a page doesn't exist


syahzuan:
### Astro info

```block
Astro version            v2.10.14
Package manager          npm
Platform                 linux
Architecture             x64
Adapter                  Couldn't determine.
Integrations             @astrojs/mdx, @astrojs/sitemap
```

### What browser are you using?

Firefox, Chrome

### Describe the Bug

This is the same issue in #7752, released in #7851 #7854, but the issue isn't fixed.

I am trying to redirect `/` (which does not exist) to the localized content (`/en` or `/zh` etc), but Astro returns 404 instead of running the middleware.

The bug is reproduced at the link below.

### What's the expected result?

Middleware should run before Astro returns 404.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-stgbcz?file=src%2Fmiddleware.js

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
@syahzuan 

I am looking into this. Here's a couple of questions:

- I think it's possible to fix this **in development**, but there's a high chance this won't work on Cloudflare, because Cloudfare runs first before astro can (even in SSR). Is this the expected result for you?
- Is there a reason why you don't use [astro redirects](https://docs.astro.build/en/core-concepts/routing/#redirects)?

syahzuan:
Here is my use case. 
My pages are all localized, placed under a [lang] folder.
To illustrate, example.com/about does not exist, rather it is redirected to its localized content e.g. example.com/en/about or example.com/zh/about, based on the users' accept-language header.
This solution is typically done through middleware.

Now, I could create a [...slug].astro directly under `pages` folder just to capture url without locale prefix and use Astro.redirect (only redirect logic without content). 
But two issues with this solution:
1. While it does the job, I think it's not elegant and seems hacky. Because I am using a page that does what middleware is supposed to do.
2. In my use case, the redirection rules would lead to a large (>100 entries) _routes.json which is [not allowed by Cloudflare](https://developers.cloudflare.com/pages/platform/functions/routing/#limits). I could create my own customized _routes.json as a workaround, but again, it's not an elegant solution.

Let me know if I could provide more info on this.

And really appreciate your quick reply on this @ematipico! Cheers.

ohansemmanuel:
[...]

PR #10206, opened 2024-02-22
feat: middleware and virtual routes

lilnasy:
## Changes

- Allows middleware to run for 404 responses without needing a 404.astro in the project.
- Fixes https://github.com/withastro/astro/issues/8244
- Fixes https://github.com/withastro/astro/issues/8226

## Testing
- Adapted from https://github.com/withastro/astro/pull/10185/files

## Docs
Mentions of `404.astro` or spread routes needing to exist to run middleware should be removed.

### DIFF
--- a/.changeset/brown-pets-clean.md
+++ b/.changeset/brown-pets-clean.md
@@ -1,0 +1,7 @@
+---
+"astro": minor
+---
+
+The astro middleware now runs when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
+
+Note that some adapters may need to update to let Astro's SSR code handle the not-found case.
--- a/.changeset/brown-pets-clean.md
+++ b/.changeset/brown-pets-clean.md
@@ -2,6 +2,6 @@
 "astro": minor
 ---
 
-The astro middleware now runs when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
+Allows middleware to run when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
 
 When a route does not match in SSR deployments, your adapter may show a platform-specific 404 page instead of running Astro's SSR code. In these cases, you may still need to add a `404.astro` or fallback route with spread params, or use a routing configuration option if your adapter provides one.
--- a/.changeset/brown-pets-clean.md
+++ b/.changeset/brown-pets-clean.md
@@ -4,4 +4,4 @@
 
 The astro middleware now runs when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
 
-Note that some adapters may need to update to let Astro's SSR code handle the not-found case.
+When a route does not match in SSR deployments, your adapter may show a platform-specific 404 page instead of running Astro's SSR code. In these cases, you may still need to add a `404.astro` or fallback route with spread params, or use a routing configuration option if your adapter provides one.
--- a/packages/astro/src/core/routing/astro-designed-error-pages.ts
+++ b/packages/astro/src/core/routing/astro-designed-error-pages.ts
@@ -9,7 +9,7 @@ export function ensure404Route(manifest: ManifestData) {
 			pattern: /\/404/,
 			prerender: false,
 			segments: [],
-			type: 'endpoint',
+			type: 'page',
 			route: '/404',
 			fallbackRoutes: [],
 			isIndex: false,
--- a/packages/astro/src/core/routing/astro-designed-error-pages.ts
+++ b/packages/astro/src/core/routing/astro-designed-error-pages.ts
@@ -1,0 +1,19 @@
+import type { ManifestData } from "../../@types/astro.js";
+
+export function ensure404Route(manifest: ManifestData) {
+	if (!manifest.routes.some(route => route.route === '/404')) {
+		manifest.routes.push({
+			component: 'astro-default-404',
+			generate: () => '',
+			params: [],
+			pattern: /\/404/,
+			prerender: false,
+			segments: [],
+			type: 'endpoint',
+			route: '/404',
+			fallbackRoutes: [],
+			isIndex: false,
+		})
+	}
+	return manifest;
+}
--- a/packages/astro/src/core/routing/astro-designed-error-pages.ts
+++ b/packages/astro/src/core/routing/astro-designed-error-pages.ts
@@ -1,9 +1,10 @@
 import type { ManifestData } from "../../@types/astro.js";
+import { DEFAULT_404_COMPONENT } from "../constants.js";
 
 export function ensure404Route(manifest: ManifestData) {
 	if (!manifest.routes.some(route => route.route === '/404')) {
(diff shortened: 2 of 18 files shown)


packages/astro/test/astro-dev-headers.test.js
```diff
@@ -31,10 +31,10 @@ describe('Astro dev headers', () => {
 			assert.equal(Object.fromEntries(result.headers)['x-astro'], headers['x-astro']);
 		});
 
-		it('does not return custom headers for invalid URLs', async () => {
+		it('returns custom headers in the default 404 response', async () => {
 			const result = await fixture.fetch('/bad-url');
 			assert.equal(result.status, 404);
-			assert.equal(Object.fromEntries(result.headers).hasOwnProperty('x-astro'), false);
+			assert.equal(Object.fromEntries(result.headers).hasOwnProperty('x-astro'), true);
 		});
```
lilnasy, line 38: The test checks that configured headers added in `config.server.headers` are NOT added in the 404 responses.

This behavior was limited to astro's default 404 page - if the user added a custom `404.astro`, `server.headers` are still added.

The new behavior is that `server.headers` apply to all astro-generated responses - the distinction between astro's default 404 page and user's `404.astro` is gone.

matthewp:
lgtm, still in draft though and missing a changeset.

lilnasy:
Manually adding `semver: minor` because label workflow doesn't work on forks.

OliverSpeir (changes requested):
blocking per request

OliverSpeir (approved):

.changeset/brown-pets-clean.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": minor
+---
+
+The astro middleware now runs when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
+
+Note that some adapters may need to update to let Astro's SSR code handle the not-found case.
```
ematipico, line 7: nit: "to update.." what? It feels like we should hint at something or maybe rephrase the sentence. 

It would be great if the changelog actually explains what an adapter maintainer should update in order to "fix" this.

packages/astro/src/core/routing/astro-designed-error-pages.ts
```diff
@@ -0,0 +1,19 @@
+import type { ManifestData } from "../../@types/astro.js";
+
+export function ensure404Route(manifest: ManifestData) {
+	if (!manifest.routes.some(route => route.route === '/404')) {
+		manifest.routes.push({
+			component: 'astro-default-404',
```
ematipico, line 6: Let's export this `'astro-default-404'` as a constant, so we use this constant in all the files that check for this string.

packages/astro/src/vite-plugin-astro-server/pipeline.ts
```diff
@@ -133,6 +134,9 @@ export class DevPipeline extends Pipeline {
 
 	async preload(filePath: URL) {
 		const { loader } = this;
+		if (filePath.href === new URL('astro-default-404', this.config.root).href) {
+			return { default: default404Page } as any as ComponentInstance
```
ematipico, line 138: Should we use `unknown` instead of `any`? Or is the `as any` necessary?

packages/astro/src/vite-plugin-astro-server/response.ts
```diff
@@ -23,6 +23,18 @@ export async function handle404Response(
...
+	{ pathname }: { pathname: string }
+) {
+	return new Response(notFoundTemplate({
+		statusCode: 404,
+		title: 'Not found',
+		tabTitle: '404: Not Found',
+		pathname,
+	}), { status: 404, headers: { 'Content-Type': 'text/html; charset=utf-8' } });
+}
+default404Page.isAstroComponentFactory = true;
```
ematipico, line 36: Let's add a comment that explains what this is doing and why

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -95,6 +95,19 @@ export async function matchRoute(
 	}
 
 	const custom404 = getCustom404Route(manifestData);
+	
+	if (custom404 && custom404.component === 'astro-default-404') {
+		const component: any = {
```
ematipico, line 100: Can we avoid the `any` and use `@ts-expect-error`, so we can add a comment that explains why we're doing this?

ematipico:
@lilnasy What about the documentation? The template is empty. We should have a PR for that, right?

.changeset/brown-pets-clean.md
lilnasy: I feel it adequately explains the change necessary.

Adapters may look at the list of routes given by astro to create their platform-specific routing file. And the platform may then run SSR code only for routes known to be on-demand rendered. The website would then fallback to a platform-designed 404 page. SST, vercel and cloudflare (in one routing strategy) work like this.

packages/astro/src/vite-plugin-astro-server/pipeline.ts
lilnasy: There wouldn't be any difference between the two in terms of safety.

packages/astro/src/vite-plugin-astro-server/response.ts
lilnasy:
```suggestion
// mark the function as an AstroComponentFactory for the rendering internals
default404Page.isAstroComponentFactory = true;
```

lilnasy:
Not necessarily. We don't explicitly document that middleware won't run in this case. We should also hold explicitly saying that will , until we know it works this way across adapters.

packages/astro/src/vite-plugin-astro-server/route.ts
lilnasy: The types around these parts are wishy washy, so you'd see a lot of any's. In this case, the correct type error'd. I'll  take another look.

ematipico:
> Not necessarily. We don't explicitly document that middleware won't run in this case. 

Sorry for being pedantic, but I am a bit concerned by this kind of approach: 
1. Can you update the PR template and explain why docs don't need to be updated? 
2. Since this is a** feature**, we should actually document this **new** behaviour.

> We should also hold explicitly saying that will , until we know it works this way across adapters.

And how do we plan to do that? The PR doesn't say anything about it.

lilnasy:
updated the description

ematipico (approved):
Thank you, Arsh, for addressing my concerns! Another `minor`ready for the next release. Let's wait for docs :)

.changeset/brown-pets-clean.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": minor
+---
+
+The astro middleware now runs when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
```
sarah11918, line 5:
```suggestion
Allows middleware to run when a matching page or endpoint is not found. Previously, a `pages/404.astro` or `pages/[...catch-all].astro` route had to match to allow middleware. This is now not necessary.
```

sarah11918 (approved):
Tiny suggestion from docs!

.changeset/brown-pets-clean.md
lilnasy: Looks great!

ematipico:
We should update the documentation: https://docs.astro.build/en/guides/middleware/

We need to mention that now the middleware will run on 404 pages

Issue #10149, opened 2024-02-17
Static File Endpoints broken with `trailingSlash: "always"` in development


kripod:
### Astro Info

```block
Astro                    v4.4.0
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             astro-icon
                         @astrojs/mdx
                         @astrojs/sitemap
                         @astrojs/tailwind
```

### Describe the Bug

Since `astro@4.1.0`, [Static File Endpoints](https://docs.astro.build/en/guides/endpoints/#static-file-endpoints) return 404 in development (running `astro dev`) when `trailingSlash: "always"` is set in the Astro config.

### What's the expected result?

Static File Endpoints should work as they did before merging https://github.com/withastro/astro/pull/9597.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-fy7uro?file=src%2Fpages%2Ftest.json.js

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Are they broken only in development (just to understand, obviously this is still very important).

kripod:
Yes, the breakage only occurs in dev mode.

ematipico:
I'll look into. it

ematipico:
Hi @kripod, we talked about this internally.

This isn't a bug. We actually that bug, and now the routing behaves consistently. As you set `trailingSlash: always`, it's expected that routes should respond to a trailing slash, endpoints too. Essentially, you were leveraging a bug. 

We will monitor the situation and see if this bugfix impacts too many users.

kripod:
@ematipico I think my issue report may have been misunderstood.

Trailing slashes are a valid requirement for most API endpoint URLs, except for the ones which are meant to serve static files, e. g. `.png` images. Having to refer to those assets via `/something.png/` instead of `/something.png` may be outright invalid in production, even when trailing slashes are preferred.

drewtownchi:
We've been having this issue for months.  We have a search.json.ts static file endpoint in the pages dir.  In prod we need to not add a trailing slash to the end of the route and in dev we need to make sure there is a trailing slash. It's not an optimal situation.

diraneyya:
This is really one of those silly things that can waste so much time and energy. It deserves a revisit. I mean: revisiting the option itself, the documentation around it, and specifically the impact of this setting on serving images locally through the `/_image` endpoint.

PR #9597, opened 2024-01-03
fix(routing): trailingSlash on endpoints

lilnasy:
## Changes
- Closes #9206 
- trailingSlash settings now apply to endpoints same as pages.

## Testing
Added a unit test.
image

## Docs
Does not affect behavior.

PR #10216, opened 2024-02-23
fix: endpoints in dev server should never have trailing slash

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/10149

## Testing

Added a new test case. It seems we didn't have one.

### DIFF
--- a/.changeset/dry-suns-protect.md
+++ b/.changeset/dry-suns-protect.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a regression where the dev server was returning a 404 when hitting an endpoint, when the app was configured with `trailingSlash: "always"`
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -383,7 +383,8 @@ function createFileBasedRoutes(
 			} else {
 				components.push(item.file);
 				const component = item.file;
-				const { trailingSlash } = settings.config;
+				// Endpoints should NOT have trailing slash
+				const trailingSlash = item.isPage ? settings.config.trailingSlash : 'never';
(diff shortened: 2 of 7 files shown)


.changeset/dry-suns-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a regression where the dev server was returning a 404 when hitting an endpoint, when the app was configured with `trailingSlash: "never"`
```
lilnasy, line 5: Did you mean "always"?

ematipico: I did!

matthewp (approved):

lilnasy (changes requested):
Assuming the changeset needs to be corrected, I think this is the wrong call.

Users should configure "trailingSlash" to inform the dev server that the production server will behave a certain way so it should too. If the behavior doesn't match, they may need to configure something else. `trailingSlash` is doing what the config is telling it to.

matthewp:
Hm, that's a good point, let's chat about it.

lilnasy:
Test cases already exist in [test/units/routing/trailing-slash.test.js](https://github.com/withastro/astro/pull/9597/files#diff-859cc2635b0ceffafe57a59e8c20d9c4fb5b5cb70b9245c430e5335b1b2b46ee)

kripod:
I‘m afraid this issue still persists: https://github.com/withastro/astro/issues/10149#issuecomment-1961767578

[Static File Endpoints](https://docs.astro.build/en/guides/endpoints/#static-file-endpoints) exposing e.g. `.png` assets for dynamic social image generation can‘t be accessed as `/img.png` in dev mode, as only `/img.png/` would pass the overly strict trailing slash requirements.

PR #10222, opened 2024-02-23
Warn when view transitions run on a prefere-reduced-motion device

martrapp:
## Changes

In DEV mode warns in the browser console when view transitions are used on a prefer-reduced-motion device.
A user at discord complained that view transitions wouldn't work at all, and @FredKSchott suggested that this could be the reason. Always fun to use an argument from authority!

***Question: *** I would even remove the restriction for DEV mode as it is really hard to find. What do the approvers think? 

## Testing

Manually tested on Windows with Chrome

## Docs

Basically the warning message copies https://docs.astro.build/en/guides/view-transitions/#prefers-reduced-motion. So docs is more than up-to-date.

### DIFF
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -51,6 +51,9 @@ const { fallback = 'animate' } = Astro.props;
 	}
 
 	if (supportsViewTransitions || getFallback() !== 'none') {
+		if (import.meta.env.DEV && window.matchMedia('(prefers-reduced-motion)').matches) {
+			console.warn(`Astro ViewTransitions: all view transition animations, including fallback animation, are disabled as this device has the prefer-reduced-motion setting enabled.`);
+		}
 		document.addEventListener('click', (ev) => {
 			let link = ev.target;
 			if (ev.composed) {
--- a/.changeset/wicked-eyes-prove.md
+++ b/.changeset/wicked-eyes-prove.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Adds a warning in DEV mode when using view transitions on a device with prefer-reduced-motion enabled.


OliverSpeir:
Hmm ignoring in dev might be nice, but I assume there’s at least one dev who actually does want to prefer reduce motion for their own sake. 

Could it be a config option ?

martrapp:
> Hmm ignoring in dev might be nice

Ah @OliverSpeir , I'm afraid, I didn't make myself clear: If someone has set up prefer-reduced-motion, we should definitely respect that. But I suspect that many developers only come across users who have set up their system this way in PROD. I therefore believe that the new warning message could also be useful in PROD.

martrapp:
Might need a different text then: "This page was created with view transitions. Because your device ...." or similar. I'm usually the worst candidate to get English text right :)

matthewp:
A lot of people will dislike logging happening in their frontend application. I think in dev mode it is ok, but wouldn't want it in prod.

martrapp:
> A lot of people will dislike logging happening in their frontend application. I think in dev mode it is ok, but wouldn't want it in prod.

Then fine with me as is.

packages/astro/components/ViewTransitions.astro
```diff
@@ -51,6 +51,9 @@ const { fallback = 'animate' } = Astro.props;
 	}
 
 	if (supportsViewTransitions || getFallback() !== 'none') {
+		if (import.meta.env.DEV && window.matchMedia('(prefers-reduced-motion)').matches) {
+			console.warn(`Astro ViewTransitions: all view transition animations, including fallback animation, are disabled as this device has the prefer-reduced-motion setting enabled.`);
```
ematipico, line 55:
```suggestion
			console.warn(`[transitions]: all view transition animations, including fallback animation, are disabled as this device has the prefer-reduced-motion setting enabled.`);
```

The format is more in line with our server logging

ematipico (approved):

Princesseuh (approved):

PR #10227, opened 2024-02-25
refactor: fix typo in index.ts

eltociear:
## Changes

- fixed typo

## Docs

colum -> column

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -175,7 +175,7 @@ export class App {
 			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
 			let protocol = request.headers.get('X-Forwarded-Proto');
 			if (protocol) {
-				// this header doesn't have the colum at the end, so we added to be in line with URL#protocol, which has it
+				// this header doesn't have the column at the end, so we added to be in line with URL#protocol, which has it
 				protocol = protocol + ':';
 			} else {
 				// we fall back to the protocol of the request


packages/astro/src/core/app/index.ts
```diff
@@ -175,7 +175,7 @@ export class App {
 			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
 			let protocol = request.headers.get('X-Forwarded-Proto');
 			if (protocol) {
-				// this header doesn't have the colum at the end, so we added to be in line with URL#protocol, which has it
+				// this header doesn't have the column at the end, so we added to be in line with URL#protocol, which has it
```
bluwy, line 178:
```suggestion
				// this header doesn't have the colon at the end, so we added to be in line with URL#protocol, which has it
```

I think this means colon (`:`)

packages/astro/src/core/app/index.ts
```diff
@@ -175,7 +175,7 @@ export class App {
 			// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
 			let protocol = request.headers.get('X-Forwarded-Proto');
 			if (protocol) {
-				// this header doesn't have the colum at the end, so we added to be in line with URL#protocol, which has it
+				// this header doesn't have the colon at the end, so we added to be in line with URL#protocol, which has it
```
Princesseuh, line 178:
```suggestion
				// this header doesn't have a colon at the end, so we add to be in line with URL#protocol, which does have it
```

PR #10228, opened 2024-02-25
Fallback URL

delucis:
## Changes

- Fixes PLT-1712
- Uses same fallback URL technique in `vite-plugin-db` as used here:
	https://github.com/withastro/astro/blob/7e34bdcc1806ece58bee04c099be885f9e5ad4f7/packages/db/src/core/utils.ts#L11-L14

## Testing

- Used this same change as a patch in a working project

## Docs

n/a


delucis:
Fixed in #10223

PR #10229, opened 2024-02-26
Update LICENSE

emjio:
## Changes
Just update LICENSE year

## Docs
n/o


jacobdalamb:
License isn't something that is updated every year.
As mentioned here: https://github.com/withastro/starlight/pull/1559

ematipico:
@jacobdalamb is right.

The year in the license means when it's created

PR #10230, opened 2024-02-26
perf: pre-calculate loop consts

castarco:
## Changes

- perf: pre-calculate normalised values for double-nested loop

## Docs

No changes in behavior


castarco:
it introduces a regression, and the performance benefit would be minimal at best.

PR #10242, opened 2024-02-27
chore: import sort source code, exception for the `astro` package

ematipico:
## Changes

This PR applies import sorting to our source code, exception for the `astro` package.

The only file that I manually changed was `packages/integrations/lit/server.js`, where I "pinned" the shim at the top of the file, by adding a space.

## Testing

The CI should pass

## Docs

N/A, internal code change


packages/integrations/lit/server.js
```diff
@@ -1,4 +1,6 @@
+// It must stay at the top
```
Princesseuh, line 1: Explaining why would be great for anyone unaware of this pattern
```suggestion
// Separate import from the rest so it doesn't get re-organized after other imports
```
something like that

packages/integrations/vercel/src/lib/nft.ts
```diff
@@ -1,6 +1,6 @@
-import type { AstroIntegrationLogger } from 'astro';
-import { relative, relative as relativePath } from 'node:path';
```
Princesseuh, line 2: Weird

bluwy (approved):

packages/integrations/vercel/src/lib/nft.ts
ematipico: I submitted a bug in the Biome repository. The code, even though weird, is correct. Biome forgets the `as` import.

Hopefully this change won't break the adapter logic!

PR #10244, opened 2024-02-27
Fix Svelte editor tooling not working under relaxed TypeScripts projects

Princesseuh:
## Changes

Svelte's tool for editor tooling (`svelte2tsx`) require TypeScript to work. A long time ago, we were able to provide it from our language server directly, but this hasn't been the case for a long time now.

An alternative fix upstream would be for `svelte2tsx` to take in an instance of TypeScript, however at the moment it imports it from a bunch of top level places and it would be a significant refactor.

Svelte's own tooling does not suffer from this because it's all bundled together, unlike us where Svelte's specific tooling is only in its own integration.
  
Fix https://github.com/withastro/language-tools/issues/652

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/gold-baboons-drum.md
+++ b/.changeset/gold-baboons-drum.md
@@ -1,5 +1,5 @@
 ---
-"@astrojs/svelte": patch
+"@astrojs/svelte": minor
 ---
 
 Adds TypeScript as a required peer dependency as it is required for Svelte's editor tooling to work correctly.
--- a/.changeset/gold-baboons-drum.md
+++ b/.changeset/gold-baboons-drum.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/svelte": patch
+---
+
+Adds TypeScript as a required peer dependency as it is required for Svelte's editor tooling to work correctly.
--- a/packages/integrations/svelte/package.json
+++ b/packages/integrations/svelte/package.json
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@sveltejs/vite-plugin-svelte": "^3.0.0",
-    "svelte2tsx": "^0.6.25"
+    "svelte2tsx": "^0.6.27"
   },
   "devDependencies": {
     "astro": "workspace:*",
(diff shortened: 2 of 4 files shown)


Princesseuh:
!preview svelte-editor-fix

packages/integrations/svelte/package.json
```diff
@@ -53,7 +53,8 @@
   },
   "peerDependencies": {
     "astro": "^4.0.0",
-    "svelte": "^4.0.0 || ^5.0.0-next.56"
+    "svelte": "^4.0.0 || ^5.0.0-next.56",
+    "typescript": "^5.3.3"
```
Princesseuh, line 57: This is not as bad as it might seem when you consider that two of our presets install TypeScript for you anyway.

ematipico (approved):
We have done our best so far; I think it's OK to add it as peer dependency!

packages/integrations/svelte/package.json
bluwy: Should we mark it as an optional peer dep? Otherwise it seems like a breaking change

Princesseuh: I feel like it was more of an oversight to not have it, since not having it result in bugs, so it's like a bug fix in some ways? I'm happy to move it to an optionalDep, though

bluwy: The line is definitely blur to me especially that this dep is only needed for editor tooling, so technically dev/build is not affected. However, package manager installs we get a warning of a missing dep now.

If we're ok with the warnings that users might get, then I think it's not crucial to have it optional. I also don't mind not making this optional if you think this is better overall!

Princesseuh: Maybe we can do a minor as a middleground? 😄

bluwy: Works for me 👍

.changeset/gold-baboons-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/svelte": patch
```
Princesseuh, line 2:
```suggestion
"@astrojs/svelte": minor
```

bluwy (approved):

Issue #8241, opened 2023-08-26
Astro v3: Recent `injectRoute` changes broke integration `astro-i18n-aut`


jlarmstrongiv:
### Astro info

```block
Astro version            v2.10.14
Package manager          npm
Platform                 darwin
Architecture             x64
Adapter                  Couldn't determine.
Integrations             @astrojs/mdx, astro-i18n-integration, @astrojs/sitemap
```

### What browser are you using?

Chrome

### Describe the Bug

Has the behavior of `injectRoute` changed substantially in the last month or so?  It seems to have broken my translation integration https://github.com/jlarmstrongiv/astro-i18n-aut/issues/19

Recent changes @natemoo-re:
- https://github.com/withastro/astro/pull/7986
- https://github.com/withastro/astro/pull/7943

It seems that the astro dev server no longer supports using `injectRoute()` with the same `entryPoint` multiple times, specifically when the `entryPoint` has dynamic paths like `[slug].astro`.  

For example, `src/pages/blog/[slug].astro` is used as the `entryPoint` for with the default page, but is also injected with a `pattern` that has a locale prefix like `/es/`  

It works fine when the `entryPoint` and `pattern` doesn’t have any dynamic path portions, like `src/pages/about.astro`

On the first load, whichever `pattern` is loaded always works.  But, trying to use another `pattern` that uses the same `entryPoint` results in an error:
```
03:44:10 PM [getStaticPaths] A `getStaticPaths()` route pattern was matched, but no matching static path was found for requested path `/blog/using-mdx/`.

Possible dynamic routes being matched: src/pages/blog/[slug].astro.
03:44:10 PM [serve]    404                         /blog/using-mdx/
03:44:12 PM [getStaticPaths] A `getStaticPaths()` route pattern was matched, but no matching static path was found for requested path `/fr/blog/using-mdx/`.

Possible dynamic routes being matched: src/pages/blog/[slug].astro.
03:44:12 PM [serve]    404                      /fr/blog/using-mdx/
```

Would this be a core astro bug?  Are there any alternatives to get the same functionality?

> Sounds like a bug to me! A GitHub issue and reproduction would be super helpful.

### What's the expected result?

The correct page loads instead of returning 404.

### Link to Minimal Reproducible Example

https://github.com/jlarmstrongiv/astro-i18n-aut/issues/19

### Participation

- [ ] I am willing to submit a pull request for this issue.

jlarmstrongiv:
Broken in the latest v3 as well:

```
[...]

PR #10248, opened 2024-02-27
fix: change strategy for route caching

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/8241
Closes PLT-1716

This PR changes the key of our internal route caching. Previously, we were using `RouteData#component`, which used to cover most of the cases, although it seems too restrictive, because the same component can emit multiple paths.

This means that, given the [minimum use case](https://github.com/withastro/astro/issues/8241#issuecomment-1863928168), the first time we hit `/blog/world`, we correctly render the route and cache its params, props and static paths using the `/blog/[post].astro` as key. When we attempt to render `/fr/blog/world`, we hit the cache because this path is rendered by the same component, although the static paths don't match, so we emit an error and return a 404.

With this change, we use `RouteData#route` as key. Given the previous use case, we cache the entry using the `/blog/[post]` key when rendering `/blog/world`. Then, when we attempt to render `/fr/blog/world`, `RouteData#route` value is `/fr/blog/[post]`, which isn't inside the cache and we correctly resolve the route and its static paths.

## Testing

I tested in a local project, and it works. The current tests should pass.

I tried to add an integration test, but I was receiving a strange error, I will investigate later. 

## Docs

N/A, bugfix

### DIFF
--- a/packages/astro/src/core/render/route-cache.ts
+++ b/packages/astro/src/core/render/route-cache.ts
@@ -102,14 +102,14 @@ export class RouteCache {
 		// NOTE: This shouldn't be called on an already-cached component.
 		// Warn here so that an unexpected double-call of getStaticPaths()
 		// isn't invisible and developer can track down the issue.
-		if (this.mode === 'production' && this.cache[route.component]?.staticPaths) {
-			this.logger.warn(null, `Internal Warning: route cache overwritten. (${route.component})`);
+		if (this.mode === 'production' && this.cache[route.route]?.staticPaths) {
+			this.logger.warn(null, `Internal Warning: route cache overwritten. (${route.route})`);
 		}
-		this.cache[route.component] = entry;
+		this.cache[route.route] = entry;
 	}
 
 	get(route: RouteData): RouteCacheEntry | undefined {
-		return this.cache[route.component];
+		return this.cache[route.route];
 	}
 }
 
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/render/route-cache.ts
```diff
@@ -102,14 +102,14 @@ export class RouteCache {
 		// NOTE: This shouldn't be called on an already-cached component.
 		// Warn here so that an unexpected double-call of getStaticPaths()
 		// isn't invisible and developer can track down the issue.
-		if (this.mode === 'production' && this.cache[route.component]?.staticPaths) {
-			this.logger.warn(null, `Internal Warning: route cache overwritten. (${route.component})`);
+		if (this.mode === 'production' && this.cache[route.route]?.staticPaths) {
+			this.logger.warn(null, `Internal Warning: route cache overwritten. (${route.route})`);
```
Fryuni, line 106: I think this will switch one conflict for another. Imagine a project with two integrations such that:
- Integration A injects route `/[...slug]` pre-rendered that generates the paths `/foo` and `/bar`
- Integration B injects route `/[...slug]` pre-rendered that generates the paths `/baz` and `/qux`
- A `src/pages/[...slug].astro` file in the project.

Those three are the same route, so they would conflict in the cache. Two integrations may define the same route for different components, and a single integration may define the same component for different routes.

Maybe the cache key should be the combination of both

ematipico: > Maybe the cache key should be the combination of both

I thought about it too! And thank you for the comment, it's good to see that I'm not the only one :) I'll apply the change

Princesseuh (approved):
Seems good to me

PR #10252, opened 2024-02-27
Add new levels of notification for dev toolbar apps

Princesseuh:
## Changes

Some users asked for the ability to show a little pill that wasn't as threatening as the red dots we have right now. This adds that.

image

## Testing

Added a test

## Docs

Will do!


ematipico:
Nit: I would make the `info` level green or blue :)

delucis (commented):
This is exciting stuff! Nice work @Princesseuh 🚥 

We should probably not rely only on colour for accessibility reasons — colour vision variations are really common. It's definitely a bit tricky at this scale though.

Let me know if it would be helpful for me to actually suggest something rather than just noting this!

Princesseuh:
> This is exciting stuff! Nice work @Princesseuh 🚥 
> 
> We should probably not rely only on colour for accessibility reasons — colour vision variations are really common. It's definitely a bit tricky at this scale though.
> 
> Let me know if it would be helpful for me to actually suggest something rather than just noting this!

Sure, feel free to suggest improvements! We definitely don't have a lot of room, so we can't really like, add visual text, but I'm sure we can figure something out

natemoo-re:
Great suggestion @delucis! @Princesseuh here's a concept for shape + color indicators.

three icons in a black bar, the first has a red square to indicate an error, the second has a yellow triangle to indicate a warning, and the third has a blue circle to indicate info

| **type** | color     | SVG                                                                                                                                                                             |
|-----------------------|-----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **error**             | `#B33E66` | `<svg viewBox="0 0 10 10"><rect width="9" height="9" x=".5" y=".5" fill="currentColor" stroke="#13151A" rx="4.5"/></svg>`                                                            |
| **warning**           | `#B58A2D` | `<svg viewBox="0 0 12 10"><path fill="currentColor" stroke="#13151A" d="m10.76 7.25-3.46-6c-.58-1-2.02-1-2.6 0l-3.46 6a1.5 1.5 0 0 0 1.3 2.25h6.92a1.5 1.5 0 0 0 1.3-2.25Z"/></svg>` |
| **info**              | `#3645D9` | `<svg viewBox="0 0 10 10"><rect width="9" height="9" x=".5" y=".5" fill="currentColor" stroke="#13151A" rx="1.5"/></svg>`                                                            |

natemoo-re (commented):
Code changes look great, left a suggestion for a slightly updated UI to improve a11y.

delucis:
+1 on Nate’s suggestion! This is pretty much what I was thinking too.

Not sure if the suggested blue has quite high enough contrast with the dark background? So exact colour shade might want checking, but for the general idea: two thumbs up!

(Aside: I hadn’t seen GitHub displaying the colour in hex codes before — that’s cool!)

natemoo-re (approved):

PR #10255, opened 2024-02-27
`create-astro`: Add fallback values when registry fails to fetch package version

natemoo-re:
## Changes

- Fixes an issue reported in Discord where a user's `package.json` ended up having `"@astrojs/check": "^undefined"` and `"typescript": "^undefined"` in the dependencies
- Updates our internal build script with some new logic to extract a few current package versions at build time. These are exposed as `process.env.ASTRO_VERSION`, `process.env.ASTRO_CHECK_VERSION`, and `process.env.TYPESCRIPT_VERSION`.
- Instead of erroring or setting to `latest` when a network version check fails, we use the most recent version (at build-time) of workspace packages as a fallback. Since we already use the `^` operator, this will result in more stable behavior than `latest`.

## Testing

Updates the TypeScript logic test to check that there are no `undefined` values in `package.json`. Not testing the exact version value since that isn't a stable value.

## Docs

N/A, bug fix.


scripts/cmd/build.js
```diff
@@ -138,3 +141,44 @@ async function clean(outdir) {
...
+	const matches = JSON.parse(stdout);
+	for (const pkg of matches) {
+		for (const [key, value] of Object.entries(pkg)) {
+			if (DEPENDENCY_KEY_RE.test(key) && value[packageName] !== undefined) {
+				const { version } = value[packageName];
+				return version;
+			}
+		}
+	}
+}
```
bluwy, line 184: For `@astrojs/check` and `typescript`, maybe we can read them from the root `package.json` instead. I think it'll be safer and more stable than running through `execa`. Also, I think we should error if we're unable to find any version, otherwise the defined vars won't get replaced and they fallback to `undefined`.

packages/create-astro/test/fixtures/not-empty/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "dev": "astro dev",
-    "build": "astro check && astro build",
```
florian-lefebvre, line 6: Why did you remove astro check? (and the extended tsconfig?)

florian-lefebvre (commented):
LGTM overall, just a little question about the fixture

packages/create-astro/test/fixtures/not-empty/package.json
natemoo-re: Ah, these files were actually reflecting an incorrect state _after_ some test runs.

We were previously not calling `after(() => resetFixtures())`, but we should have been. These files are now aligned with the "reset" state.

scripts/cmd/build.js
natemoo-re: Yep, great suggestion. Refactored to read from the workspace root instead!

As for the missing behavior, I also updated that to throw a more helpful error.

bluwy (approved):
Thanks for updating it. This looks great!

florian-lefebvre (approved):

PR #10257, opened 2024-02-28
improve link command

FredKSchott:
## Changes

- Improves the `astro link` command to handle new project creation
- Improves the `astro link` command to handle existing project linking by name (vs. currently, only supporting an id passed as argument)

## Testing

- N/A

## Docs

- N/A

### DIFF
--- a/packages/db/src/core/cli/commands/link/index.ts
+++ b/packages/db/src/core/cli/commands/link/index.ts
@@ -1,36 +1,60 @@
-import { mkdir, writeFile } from 'node:fs/promises';
 import type { AstroConfig } from 'astro';
+import { slug } from 'github-slugger';
 import { bgRed, cyan } from 'kleur/colors';
+import { mkdir, writeFile } from 'node:fs/promises';
+import { homedir } from 'node:os';
+import { basename } from 'node:path';
+import ora from 'ora';
 import prompts from 'prompts';
 import type { Arguments } from 'yargs-parser';
 import { MISSING_SESSION_ID_ERROR } from '../../../errors.js';
 import { PROJECT_ID_FILE, getSessionIdFromFile } from '../../../tokens.js';
 import { getAstroStudioUrl } from '../../../utils.js';
 
-export async function cmd({ flags }: { config: AstroConfig; flags: Arguments }) {
-	const linkUrl = new URL(getAstroStudioUrl() + '/auth/cli/link');
+export async function cmd({}: { config: AstroConfig; flags: Arguments }) {
 	const sessionToken = await getSessionIdFromFile();
 	if (!sessionToken) {
 		console.error(MISSING_SESSION_ID_ERROR);
 		process.exit(1);
 	}
-	let body = { id: flags._[4] } as {
-		id?: string;
-		projectIdName?: string;
-		workspaceIdName?: string;
-	};
-	if (!body.id) {
-		const workspaceIdName = await promptWorkspaceName();
-		const projectIdName = await promptProjectName();
-		body = { projectIdName, workspaceIdName };
+	const getWorkspaceIdAsync = getWorkspaceId();
+	await promptBegin();
+	const isLinkExisting = await promptLinkExisting();
+	if (isLinkExisting) {
+		const workspaceId = await getWorkspaceIdAsync;
+		const existingProjectData = await promptExistingProjectName({workspaceId});
+		return await linkProject(existingProjectData.id);
 	}
+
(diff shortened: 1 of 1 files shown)


packages/db/src/core/cli/commands/link/index.ts
```diff
@@ -1,36 +1,76 @@
...
+
+	const isLinkNew = await promptLinkNew();
+	if (isLinkNew) {
+		const newProjectName = await promptNewProjectName();
+		const newProjectRegion = await promptNewProjectRegion();
+		const newProjectId = await createNewProject({name: newProjectName, region: newProjectRegion});
+		// TODO(fks): Actually listen for project creation before continuing
+		// This is just a dumb spinner that roughly matches database creation time.
+		const spinner = ora('Creating new project...');
+		await new Promise((r) => setTimeout(r, 4000));
```
bholmesdev, line 35: "First make it work, then make it good."
~ Confucius, probably

bholmesdev: When this is addressed, it would be great to provide a link to your project dashboard!

packages/db/src/core/cli/commands/link/index.ts
```diff
@@ -1,36 +1,76 @@
...
+	const response = await fetch(linkUrl, {
+		method: 'POST',
+		headers: {
+			Authorization: `Bearer ${await getSessionIdFromFile()}`,
+			'Content-Type': 'application/json',
+		},
+		body: JSON.stringify({ name, region }),
+	});
+	if (!response.ok) {
+		console.error(`Failed to create project: ${response.status} ${response.statusText}`);
```
bholmesdev, line 58: Can we standardize to the `{ success: boolean, value: JSON }` format for reading errors? I worry `statusText` could log errors that aren't human-readable

packages/db/src/core/cli/commands/link/index.ts
```diff
@@ -1,36 +1,76 @@
...
+export async function promptExistingProjectName(): Promise<string> {
+	const linkUrl = new URL(getAstroStudioUrl() + '/api/cli/projects.list');
 	const response = await fetch(linkUrl, {
 		method: 'POST',
 		headers: {
 			Authorization: `Bearer ${await getSessionIdFromFile()}`,
 			'Content-Type': 'application/json',
 		},
-		body: JSON.stringify(body),
+		body: JSON.stringify({}),
```
bholmesdev, line 73: nit: should this be a GET? Seems odd to post an empty body to get a list of projects

packages/db/src/core/cli/commands/link/index.ts
```diff
@@ -47,33 +87,95 @@ export async function cmd({ flags }: { config: AstroConfig; flags: Arguments })
...
+		console.log('Canceled.')
 		process.exit(0);
 	}
-	return projectName;
+	return projectId;
+}
+
+export async function promptBegin(): Promise<void> {
+	// Get the current working directory relative to the user's home directory
+	const prettyCwd = process.cwd().replace(homedir(), '~');
```
bholmesdev, line 107: Please work on windows please work on windows please work on

bholmesdev (commented):
Thanks for adding the CLI so `prompt`ly! Left comments mostly on API signature.

packages/db/src/core/cli/commands/link/index.ts
FredKSchott: it *should...* right? `homedir()` is platform agnostic so I think this should work. We'll bug @itsMapleLeaf after merging to test :)

packages/db/src/core/cli/commands/link/index.ts
FredKSchott: ah good catch, I think that already is the response format and this is just checking for the case where our entire API fails and you just get back an unhandled 500. But, we should also have proper handling for the handled error case.

packages/db/src/core/cli/commands/link/index.ts
FredKSchott: oh nice, that's a great call. Just refactored the code a bit to make this even easier in the future

packages/db/src/core/cli/commands/link/index.ts
FredKSchott: Because we're already maintaining TRPC, we follow an RPC-like approach to this API as well, which effectively says that request method doesn't matter (so POST is okay even if you're just getting data).

PR #10258, opened 2024-02-28
improve login command 

FredKSchott:
## Changes

- Improves the login flow, fixing a bug for Brave users who were blocked from talking to localhost in the browser. Now, we redirect the user directly to localhost to pass the session token via query param.

## Testing

- N/A

## Docs

- N/A


natemoo-re (approved):
Nice! LGTM!

PR #10259, opened 2024-02-28
Update Vite to latest

bluwy:
## Changes

The main focus of the PR is to bump `@astrojs/db` Vite version from v4 to v5. I figured to bump all the Vite versions anyway.

## Testing

Existing tests should pass.

## Docs

Not really sure if a changeset is needed. It doesn't affect how `@astrojs/db` works as far as I can tell.


packages/db/src/core/integration/load-astro-config.ts
```diff
@@ -69,8 +69,8 @@ async function loadConfigWithVite(configPath: string): Promise<Record<string, un
 
 async function createViteServer(): Promise<ViteDevServer> {
 	const viteServer = await createServer({
-		server: { middlewareMode: true, hmr: false, watch: { ignored: ['**'] } },
-		optimizeDeps: { disabled: true },
+		server: { middlewareMode: true, hmr: false, watch: null },
+		optimizeDeps: { noDiscovery: true },
```
bluwy, line 73: This change keeps it in line with the current code in Astro core, however, it doesn't look like this file is used at all. Is it safe to remove it?

Princesseuh (approved):

ematipico (approved):

packages/db/src/core/integration/load-astro-config.ts
bluwy: Ben mentioned on Discord that the file is no longer used as the CLI attaches to the `astro` commend. So I've removed it in the last commit.

natemoo-re (approved):

PR #8826, opened 2023-10-13
fix(lit): pin last lit 2.x compatible versions

lilnasy:
## Changes
- Fixes #8825

## Testing
Existing tests should pass. I would've expected them to be failing but maybe the lockfile kept CI passing.

## Docs
Does not affect usage.

PR #10267, opened 2024-02-28
fix: update lit@3.1.2

aleksandrjet:
## Changes
While working on [PR](https://github.com/withastro/astro/pull/10202) I noticed that the `lit@3.1.2` update breaks the tests and the build. A similar problem was fixed here [pr:8826](https://github.com/withastro/astro/pull/8826).

The current solution is to add dependencies from `lit` to `integrations/lit`:
``` json
"@lit/reactive-element": "^2.0.4",
"lit-element": "^4.0.4",
"lit-html": "^3.1.2"
```
Without adding dependencies, tests fail, after adding everything is ok

## Testing
Existing tests are sufficient to verify

## Docs
I don't think documentation is needed for this. This does not have any side-effects, because these new dependencies will be installed with `lit`

### DIFF
--- a/.changeset/wet-games-drive.md
+++ b/.changeset/wet-games-drive.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/lit": patch
+---
+
+Fixing `lit` update error to version 3.1.2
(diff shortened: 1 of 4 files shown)


florian-lefebvre:
Looks like lit's version is now `3.1.3` (https://github.com/withastro/astro/blob/main/packages/integrations/lit/package.json#L59) and tests are not failing so I'm closing this PR as well as the error demo PR. Feel free to reopen if you think I'm wrong!

PR #10268, opened 2024-02-28
feat(audits): Handle mutations

Princesseuh:
## Changes

Set up a MutationObserver to rerun auditing the page after changes. There's a bunch of perf concerns here for sure, but according to Google, in 2024 MutationObserver's performance isn't that bad? Also, I tried to set up a bunch of reasonable timeouts.

Note that this does not handle images and iframes that haven't loaded yet, that requires a bit more changes.

## Testing

Added a bunch of tests, that was the most complicated part of this PR.

## Docs

N/A


packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -69,8 +70,46 @@ export default {
...
+		});
+		document.addEventListener('astro:after-swap', async () => {
+			lint();
+		});
+		document.addEventListener('astro:page-load', async () => {
+			refreshLintPositions();
+
+			// HACK: View transitions add a route announcer after this event, so we need to wait for it to be added
+			setTimeout(() => {
+				setupObserver();
```
ematipico, line 110: Query: we call `setupObserver` twice (here and at line 97), does that mean we end up observing `document.body` twice?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -69,8 +70,46 @@ export default {
...
+							lint();
+						},
+						{ timeout: 300 }
+					);
+				} else {
+					// Fallback for old versions of Safari
+					setTimeout(() => {
+						lint();
+					}, 150);
+				}
```
ematipico, line 93: Why doe we have two different timeouts?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
Princesseuh: No, because we disconnect the observer on page preparation. (line 100) Additionally, it's not possible for a single observer to watch the same element twice

Per MDN:
> If you call observe() on a node that's already being observed by the same MutationObserver, all existing observers are automatically removed from all targets being observed before the new observer is activated.

The final `expect` in the `handle mutations properly during view transitions` test also test this, by checking that there's only one instance of "Rerunning lints..."

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -69,8 +70,46 @@ export default {
...
 		await lint();
 
-		document.addEventListener('astro:after-swap', async () => lint());
-		document.addEventListener('astro:page-load', async () => refreshLintPositions);
+		let mutationDebounce: ReturnType<typeof setTimeout>;
+		const observer = new MutationObserver(() => {
+			if (mutationDebounce) {
+				clearTimeout(mutationDebounce);
+			}
+			mutationDebounce = setTimeout(() => {
```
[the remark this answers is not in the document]
Princesseuh: Will do!

packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
Princesseuh: We only want to rerun the lints after the final page mutation, so we debounce it with the first `setTimeout`, but we also want to audit only if the browser is idle (in order to avoid affecting the page's performance). Alas, Safari doesn't support `requestIdleCallback` so we have to "fake it" (we assume that after 150ms, the page is hopefully idle)

Princesseuh:
@ematipico Added comments explaining the points you were confused about!

ematipico (approved):
Thank you Erika! Looks good to me

bluwy (approved):

sarah11918 (approved):
Docs approval!

PR #10272, opened 2024-02-29
Updates another Node.js 16 github action

martrapp:
## Changes

second helping for #10270 

## Testing

n.a.

## Docs
n.a.


ematipico (approved):

Princesseuh:
That one we purposely left at the lower version because they changed the syntax of the labelling file

Princesseuh (changes requested):
Syntax of the label file needs to be updated if this action is to be updated.

Issue #10066, opened 2024-02-09
Using MDX and building for production outputs both optimized and original image assets


ddotlic:
### What version of `starlight` are you using?

0.17.1

### What version of `astro` are you using?

4.2.6

### What package manager are you using?

yarn 1.x

### What operating system are you using?

macOS 14.3

### What browser are you using?

Not browser specific

### Describe the Bug

When the documentation is in "plain" Markdown files `.md` then the images get optimized and copied out into `dist/_astro`.

As soon as we renamed all the documentation files to `.mdx` (was a necessity, needed to use Astro/Starlight components), building for production yields both optimized and original assets. Specifically, am talking about images placed in `src/assets` (those in `public` are IIUC not optimized anyway).

To repro, simply create a new project, add a `.png` image to assets and reference it from a `.md` file and then `build` - you should get a single `.webp` image. Renaming to `.mdx` and changing nothing else, will yield the same screenshot being output twice - once as `.webp` and once as `.png` (renamed, but otherwise identical).

Yes, I am aware that I _could_ post-process the build result and remove the `.png` files from the output, but it's not in the spirit of this project IMO where the tooling is **so good** and nothing needs silly tweaking.

**Important**: was unable to figure out how to add an image to StackBlitz, the repro needs a single `Screenshot.png` in the assets folder.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-bwwxgj-kxfzr1?file=src%2Fcontent%2Fdocs%2Freference%2Fcool.mdx

### Participation

- [X] I am willing to submit a pull request for this issue.

Princesseuh:
What I can only assume to be our JSX renderer is trying to access `$$slot` on the image import, since we keep the image on any access, it causes the original image to stay wrongly.

ddotlic:
@Princesseuh Thank you! Can confirm that this does indeed resolve this issue. Cheers!

PR #10278, opened 2024-02-29
fix(markdoc & mdx): Proxy crimes

Princesseuh:
## Changes

In both MDX and Markdoc we had issues with the proxy. Well, actually it's only in Markdoc, in MDX the issue is that our JSX renderer tries to access some stuff it shouldn't, I tried to fix that, but couldn't figure it out, so I went another way by making the proxy only trigger if the property exists.

In Markdoc we unfortunately just lose the proxy completely, so we can't track references. Oh well, we'll just assume images are always used, it's not the end of the world.

Fix https://github.com/withastro/astro/issues/10066 (the Markdoc issue doesn't have a corresponding issue, noticed the problem on my website)

## Testing

Added a test! For Markdoc I additionally tried on my own website that has a complex Markdoc setup (custom component and everything) to see if it fixes the issue and everything worked.

## Docs

N/A


packages/astro/package.json
```diff
@@ -108,6 +108,7 @@
     "dev": "astro-scripts dev --copy-wasm --prebuild \"src/runtime/server/astro-island.ts\" --prebuild \"src/runtime/client/{idle,load,media,only,visible}.ts\" \"src/**/*.{ts,js}\"",
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
     "test": "pnpm run test:node",
+    "test:match": "pnpm run test:node --match",
```
Princesseuh, line 111: `node:test`'s `match` is unfortunately terrible, it takes multiple minutes on my computer since it runs through every file still. However, the command in the root repo was broken, so this fixes that.

ematipico (approved):

Issue #10164, opened 2024-02-20
Middleware dependencies are incorrectly bundled


danielo515:
### Astro Info

```block
Astro                    v4.4.1
Node                     v18.18.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/vercel/serverless
Integrations             @astrojs/tailwind
                         @astrojs/svelte       
```

### Describe the Bug

I am using a middleware to handle auth. In local dev mode, everything works as expected. However, when deployed to vercel, one of the imported constants becomes `noop` instead of being an array.

To give you an idea of how it looks.
Here is the original middleware

```ts
import { errors, jwtVerify, type JWTPayload } from "jose";
import { defineMiddleware } from "astro/middleware";
import { TOKEN, PUBLIC_ROUTES } from "./constants";

const secret = new TextEncoder().encode(import.meta.env.JWT_SECRET);

//... etc
```

Then, the constants files contains:

```ts
export const TOKEN = "token";
export const PUBLIC_ROUTES = [
  "/",
  "/login",
  "/api/login",
];
```

However, this is the generated middleware (part)

```js
import { jwtVerify, errors } from 'jose';
import './astro_BxjT5Sqv.mjs';
import { P as PUBLIC_ROUTES, T as TOKEN } from './prerender_BwN9fO_L.mjs'; //<- key part

// bla bla bla
```

And this is the file where PUBLIC_ROUTES is being imported from:

```js
const noop = () => {};
export const P = noop;
export const T = noop;
export const _ = noop;
export const a = noop;
export const b = noop;
export const c = noop;
export const d = noop;
export const i = noop;
export const l = noop;

```

### What's the expected result?

I expect the constants to be inlined in the generated middleware so it can properly work

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-t5xhyw

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
@danielo515, please replace the `Astro info` block with the information returned by the command `Astro info`. This information is vital to us so we can triage your issue.

mikenicklas:
I am seemingly running into this issue out of nowhere as well and I can't even run `astro info`.

```
mikenicklas@Mikes-MacBook-Pro fe % npm run astro info

> fe@0.0.1 astro
> astro info

6:02:24 AM [vite] Error when evaluating SSR module /Users/mikenicklas/Sites/Work/fe/astro.config.mjs:
[...]

PR #10279, opened 2024-02-29
fix(build): skip only the configured redirects

lilnasy:
## Changes

- Reported here: https://github.com/withastro/astro/issues/10164#issuecomment-1957016426.
- A middleware returned redirects which made files disappear in build only with the vercel adapter.

## Testing
- Added a case in `test/redirects.test.js`.

## Docs
- Does not affect usage.

### DIFF
--- a/.changeset/fifty-buttons-clean.md
+++ b/.changeset/fifty-buttons-clean.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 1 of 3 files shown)


ematipico (approved):

ematipico:
@lilnasy does this PR fix https://github.com/withastro/astro/issues/10164#issuecomment-1957016426?

PR #10280, opened 2024-02-29
finalize WIP API

bholmesdev:
## Changes

- Move to unified `defineTable()` config
- Move all logic to the `db/` directory for config, seeding, and migrations
- Remove all logic and test suites for "readable" tables
- Temporary: remove "unsafe disable" flag for studio connections in prod builds

## Testing

- Update basics tests to be development-only. Should create a proxy server to mimic a remote connection for production builds.

## Docs

They're coming, Sarah. They're coming.


packages/db/src/core/integration/typegen.ts
```diff
@@ -28,19 +28,7 @@ ${Object.entries(tables)
 function generateTableType(name: string, collection: DBTable): string {
 	let tableType = `	export const ${name}: import(${RUNTIME_IMPORT}).Table<
 		${JSON.stringify(name)},
-		${JSON.stringify(
-			Object.fromEntries(
-				Object.entries(collection.columns).map(([columnName, column]) => [
-					columnName,
-					{
-						// Only select columns Drizzle needs for inference
```
bholmesdev, line 36: This destructuring actually introduced bugs in generated types! Pass through everything that's JSON serializable for inference

packages/db/src/core/integration/index.ts
```diff
@@ -31,14 +33,13 @@ function astroDBIntegration(): AstroIntegration {
 		hooks: {
 			'astro:config:setup': async ({ updateConfig, config, command: _command, logger }) => {
 				command = _command;
+				root = config.root;
 				if (_command === 'preview') return;
 
 				let dbPlugin: VitePlugin | undefined = undefined;
-				const studio = config.db?.studio ?? false;
 
-				if (studio && command === 'build' && process.env.ASTRO_DB_TEST_ENV !== '1') {
```
bholmesdev, line 39: The test suite was updated to test the dev server, so this env flag was removed. For production builds, we should instead stand up a proxy URL and override using the `ASTRO_STUDIO_REMOTE_DB_URL` variable. Testing a local sqlite file in production was not mirroring the expected user flow.

packages/db/src/core/types.ts
```diff
@@ -318,69 +265,11 @@ interface IndexConfig<TColumns extends ColumnsConfig> extends z.input<typeof ind
...
-	return {
-		type,
-		/**
-		 * @internal
-		 */
-		schema,
-	};
-}
-
-export const column = {
```
bholmesdev, line 370: moved to `runtime/config.ts`

packages/db/src/core/cli/commands/gen/index.ts
```diff
@@ -9,12 +10,23 @@ import {
 	getMigrationStatus,
 	initializeMigrationsDirectory,
 } from '../../migrations.js';
+import { getMigrationsDirUrl } from '../../../utils.js';
```
FredKSchott, line 13: nit: `Directory` over `Dir`

packages/db/src/core/load-file.ts
```diff
@@ -0,0 +1,141 @@
...
+import {
+	getConfigVirtualModContents,
+	getStudioVirtualModContents,
+} from './integration/vite-plugin-db.js';
+import type { DBTables } from './types.js';
+import { writeFile, unlink } from 'node:fs/promises';
+import { getDbDirUrl } from './utils.js';
+import { existsSync } from 'node:fs';
+
+export async function executeFile({
```
FredKSchott, line 13: nit, but it's unclear what kind of file we're talking about here. Is this for loading the `db/config.ts` file? `db/seed.ts`? A custom seed file for `astro db execute`? 

Renaming the file to be less generic might help, or at the very least adding a comment to the file or the individual functions.

packages/db/src/runtime/db-client.ts
```diff
@@ -1,58 +1,66 @@
...
-	const tableName = getTableName(Table);
-	const collection = tables[tableName];
-	if (!collection.writable) {
-		throw new Error(`The [${tableName}] collection is read-only.`);
+/**
+ * Sorted by precedence.
+ * Ex. If both "seed.dev.ts" and "seed.ts" are present,
+ * "seed.dev.ts" will be used.
+ */
+export const SEED_DEV_FILE_NAMES_SORTED = [
```
FredKSchott, line 27: do we need this feature for launch? I'd rather pick one of these and document it, vs. the complexity of the user reading docs and thinking "wait, is seed.development.ts different from seed.ts?"

packages/db/src/core/cli/index.ts
```diff
@@ -1,50 +1,56 @@
...
-			const { cmd } = await import('./commands/logout/index.js');
-			return await cmd({ config, flags });
-		}
-	}
-
-	if (!config.db?.studio) {
-		console.log(STUDIO_CONFIG_MISSING_CLI_ERROR);
-		process.exit(1);
-	}
+	const { mod } = await loadConfigFile(astroConfig.root);
```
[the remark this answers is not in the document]
bholmesdev: This is needed yes! We are now loading the `db/config.ts` file.

FredKSchott (approved):
LGTM!

packages/db/src/core/load-file.ts
bholmesdev: This file hold the general logic to load a JS file. There are two domain-specific functions:
- executeFile for `db execute`
- `loadConfigFile` for `db/config.ts`

I can move these domain-specific functions out of this file for clarity, or add jsdoc

packages/db/src/runtime/db-client.ts
bholmesdev: Was planning on a `seed.prod.ts` that runs with a `pnpm astro db seed` command. This would be a shortcut to running `pnpm astro db execute db/seed.whatever.ts`. If you don't think this is worth pursuing, I will remove!

packages/db/src/core/load-file.ts
bholmesdev: Update: moved executeFile() logic inline with the execute cli command. Also renamed `loadConfigFile()` to `loadDbConfigFile()`

packages/db/src/runtime/db-client.ts
bholmesdev: Update: decided `astro db seed` isn't very compelling. Just seed.ts, and `astro db execute` for any other file. Easy to doc

packages/db/src/core/cli/index.ts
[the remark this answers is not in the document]
FredKSchott: perfect, thanks!

.changeset/rich-turtles-live.md
```diff
@@ -0,0 +1,6 @@
+---
+"astro": patch
+"@astrojs/db": patch
```
FredKSchott, line 3: nit: lets make this the forcing `minor` release

PR #10281, opened 2024-02-29
fix(i18n): prevent overwriting 404.astro

### LINKED ISSUE #10276: `404.astro` Is Not Rendered When Unsupported Locale In `/[lang]/` Is Typed
### Astro Info

```block
Astro                    v4.4.6
Package Manager          pnpm
                         @astrojs/sitemap
```

Chrome

### Describe the Bug

Hello, 

My website is multilingual, in Astro configuration I've written the following:
```
  i18n: {  
  defaultLocale: "en",
  locales: ["ku", "en", "ar"],
  routing: {
    prefixDefaultLocale: true,
	redirectToDefaultLocale: false
   }
}
```
And this is my directory structure:
image

If I type "/en/contact/" in my browser URL, the page loads successfully, if I type "/en/category/" my 404.astro page loads as expected because that page doesn't exist, but if I type "/fr/contact/" I get the browser's own 404 error page, not my own 404.astro page, I do want it to return 404 because "fr" locale is not supported by my website, but I want it to use my 404.astro not the browser's own 404 page.

P.S.
I'm sorry I don't know how to use CodeSanbox, but I have provided link contain my project source code as zip, please use it for recreating the issue. 

Load my 404.astro page when an unsupported locale is typed in /[lang]/.

https://filebin.net/lauq80leuinyfohc

### Participation

lilnasy:
## Changes
- Fixes #10276 

## Testing
- Added cases to `test/i18n-routing.test.js`

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -274,7 +274,7 @@ export async function handleRoute({
 		response.headers.get(REROUTE_DIRECTIVE_HEADER) !== 'no'
 	) {
 		const fourOhFourRoute = await matchRoute('/404', manifestData, pipeline);
-		if (options && fourOhFourRoute?.route !== options.route)
+		if (options)
 			return handleRoute({
 				...options,
 				matchedRoute: fourOhFourRoute,
@@ -288,6 +288,12 @@ export async function handleRoute({
 				incomingResponse,
 			});
 	}
+
+	// We remove the internally-used header before we send the response to the user agent.
+	if (response.headers.has(REROUTE_DIRECTIVE_HEADER)) {
+		response.headers.delete(REROUTE_DIRECTIVE_HEADER);
+	}
+
 	if (route.type === 'endpoint') {
 		await writeWebResponse(incomingResponse, response);
 		return;
--- a/.changeset/curvy-trees-compare.md
+++ b/.changeset/curvy-trees-compare.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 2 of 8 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -288,6 +288,12 @@ export async function handleRoute({
 				incomingResponse,
 			});
 	}
+
+	// We remove the internally-used header before we send the response to the user agent.
+	if (response.headers.has(REROUTE_DIRECTIVE_HEADER)) {
+		response.headers.delete(REROUTE_DIRECTIVE_HEADER);
+	}
+
```
ematipico, line 296: I wonder, should we also do this during the build? If so, at this point we should have this code inside the pipeline, because it seems we have this logic in SSR too:

https://github.com/withastro/astro/blob/87a3d51f2ca8661babbb76956e54bf389eb86d8f/packages/astro/src/core/app/index.ts#L323-L325

Of course, we can land this possible improvement in another PR

ematipico (approved):

packages/astro/src/vite-plugin-astro-server/route.ts
lilnasy: You are exactly right! It makes sense for this to be managed by `Pipeline`/`RenderContext`.

Error page rerouting is currently the jurisdiction of `vite-plugin-astro-server` and `App`. When it is turns into a middleware, the header could be removed in a single place. It would be similar to how `X-Astro-Route-Type` is handled after the i18n middleware has read it.

Regarding doing this for static builds, I'm not sure if it would have an effect since headers are ignored altogether.

bluwy (approved):

PR #10282, opened 2024-03-01
Node Adapter server.host: true Issue

### LINKED ISSUE #10264: Using the `true` option in `server.host` doesn't work
### Astro Info

```block
Astro                    v4.4.5
Package Manager          npm
Integrations             @astrojs/alpinejs
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

The `server.host` option with `output: "server"` option doesn't build the correct output when used with the `true` value. The resulting code just checks for `typeof options.host === "boolean"` and uses `"localhost"` in this case, not considering the actual value of the (boolean) option. The documentation states that true should instead result in listening to all addresses. Our current workaround is to just use `"0.0.0.0"` instead of `true`.
 https://github.com/withastro/astro/blob/e56c428f3bfc3a4a8ed799993f80a29a580cb820/packages/integrations/node/src/standalone.ts#L15

Setting the `server.host` option should result in a server listening to all addresses.

https://stackblitz.com/edit/github-pbnej2

### Participation

SatanshuMishra:
## Changes

**Problem**: In Astro's "server" output mode, setting `server.host: true` incorrectly binds the server to `localhost`, preventing connections from other devices.

**Cause**: The existing code always assigns `localhost` if the `options.host` value is boolean, regardless of whether it's `true` or `false`.

**Proposed Fix**: The code now correctly handles both string and boolean values for `options.host`:

- String: If `options.host` is a string, its value is used directly.
- `True`: Binds to `0.0.0.0`.
- `False`: Binds to `localhost`.

image

## Testing

A new test was created to ensure this change behaved as expected. All tests were run and passed.

image

## Docs

This change simply fixes a bug. No affect to user's behavior or doc changes are expected.

Thank you to @kevinzunigacuellar & @lilnasy for their contributions to this PR. This PR closes #10264.


lilnasy:
The changes look great!

There seems to be an extra `.changeset` folder. We have that in the root of the monorepo: https://github.com/withastro/astro/tree/main/.changeset

SatanshuMishra:
> The changes look great!
> 
> There seems to be an extra `.changeset` folder. We have that in the root of the monorepo: https://github.com/withastro/astro/tree/main/.changeset

Oh! I realize my mistake now. Apologies! Will fix it!

.changeset/smooth-singers-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/node": patch
+---
+
+Modified node package to properly assign server.host option set to `true` to listen on all network interfaces. Added test to test this functionality.
```
lilnasy, line 5: This is great! It describes perfectly what this PR does.

This changeset would become part of [node changelog](https://github.com/withastro/astro/blob/main/packages/integrations/node/CHANGELOG.md), and the primary reader would be astro users. With that in mind, we would want leave out the statement about tests, as that's between maintainers and adequately communicated with the PR itself.
```suggestion
Modified node package to properly assign server.host option set to `true` to listen on all network interfaces.
```

lilnasy (approved):
Looks great to me! Thanks @kevinzunigacuellar for the clear tests!

kevinzunigacuellar:
Thanks to @lilnasy for all the guidance. Couldn't have done it without his help.

packages/integrations/node/src/standalone.ts
```diff
@@ -9,11 +9,17 @@ import { createAppHandler } from './serve-app.js';
 import { createStaticHandler } from './serve-static.js';
 import type { Options } from './types.js';
 
+// Used to get Host Value at Runtime
+export const hostOptions = (host: string | boolean): string => {
```
kevinzunigacuellar, line 13: Perhaps we can use the type from `Options`

```suggestion
export const hostOptions = (host: Options["host"]): string => {
```

kevinzunigacuellar (commented):
Just a tiny comment to reuse some types. Thanks for putting together the PR @SatanshuMishra

packages/integrations/node/src/standalone.ts
SatanshuMishra: Good suggestion! LGTM!

.changeset/smooth-singers-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/node": patch
+---
+
+Modified node package to properly assign server.host option set to `true` to listen on all network interfaces.
```
florian-lefebvre, line 5:
```suggestion
Fixes the `server.host` option to properly listen on all network interfaces when set to `true`
```

florian-lefebvre (approved):
LGTM! A little changeset suggestion ;)

PR #10283, opened 2024-03-01
fix: fixed an error in the vue component editor(https://github.com/wi…

phk422:
…thastro/language-tools/issues/758)

## Changes

- Modify the regular expression for matching `defineProps`
- Removes the condition where comments exist when working with code
- Fixed an issue with 'defineProps' import error

### DIFF
--- a/packages/integrations/vue/package.json
+++ b/packages/integrations/vue/package.json
@@ -49,7 +49,6 @@
     "@types/chai": "^4.3.10",
     "astro": "workspace:*",
     "astro-scripts": "workspace:*",
-    "chai": "^4.3.10",
     "cheerio": "1.0.0-rc.12",
     "linkedom": "^0.16.4",
     "vite": "^5.1.4",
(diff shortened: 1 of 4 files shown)


Princesseuh:
Awesome, thank you! Would it be possible to add tests? Essentially, add a new `x.test.js` file in the `test` folder of the package, import `toTSX` and make sure the code generated is valid.

packages/integrations/vue/package.json
```diff
@@ -49,6 +49,7 @@
     "@types/chai": "^4.3.10",
     "astro": "workspace:*",
     "astro-scripts": "workspace:*",
+    "chai": "^4.3.10",
```
ematipico, line 52: We don't use `chai` anymore, you'll have to use `node:assert/strict`

ematipico (changes requested):
The new tests use an assertion library that we don't use anymore.

pnpm-lock.yaml
ematipico, line 1: You have to revert this file too @phk422

PR #6176, opened 2023-02-08
Take dynamic imports into account for CSS ordering

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/5676

## Testing

- New test case added

## Docs

N/A, bug fix

PR #10291, opened 2024-03-01
Treeshake unused Astro scoped styles

bluwy:
## Changes

Unused scoped styles from Astro components are removed by checking if the main Astro module (that owns the scoped styles) is also removed/treeshaken. The PR copies some code from https://github.com/vitejs/vite/pull/16058 to fix this at the meantime.

There's currently a tradeoff with this approach (https://github.com/vitejs/vite/pull/16058#issuecomment-1972860379), but it should be better than before, and also it could be better.

## Testing

Added a test. Also tested on a starlight project.

## Docs

n/a. internal improvement


packages/astro/src/core/compile/compile.ts
```diff
@@ -42,7 +35,7 @@ export async function compile({
 	// Because `@astrojs/compiler` can't return the dependencies for each style transformed,
 	// we need to use an external array to track the dependencies whenever preprocessing is called,
 	// and we'll rebuild the final `css` result after transformation.
-	const cssDeps: CompileCssResult['dependencies'][] = [];
+	const cssPartialCompileResults: PartialCompileCssResult[] = [];
```
bluwy, line 38: Note: I refactored this from `string[][]` to `{ ... dependencies: string[] }[]` because I needed to keep an extra `isGlobal` metadata whenever preprocessing is called.

packages/astro/test/fixtures/css-order-dynamic-import/src/pages/one.astro
```diff
@@ -1,7 +1,7 @@
 ---
 import '../components/One.astro';
 import '../components/Two.astro';
-await import('../components/Three.astro');
+await import('../components/Three.js');
```
bluwy, line 4: This PR caused this test to fail because the `Three.astro` component was imported but it's never used, hence the scoped styles in `Three.astro` weren't injected. Previously it was testing if the scoped styles were injected which I don't think it's quite right.

This test comes from https://github.com/withastro/astro/pull/6176, so to stay true to what it's testing, I changed it to a `.js` file so that the CSS is no longer scoped.

bluwy:
!preview treeshake-scoped-css

matthewp (approved):
love it!

PR #10293, opened 2024-03-01
fix: fixed an error in the vue component editor

phk422:
fix https://github.com/withastro/language-tools/issues/758

## Changes

- Modify the regular expression for matching defineProps
- Removes the condition where comments exist when working with code
- Fixed an issue with 'defineProps' import error

### DIFF
--- a/packages/integrations/vue/src/editor.cts
+++ b/packages/integrations/vue/src/editor.cts
@@ -23,7 +23,8 @@ export function toTSX(code: string, className: string): string {
 		const { scriptSetup } = parsedResult.descriptor;
 
 		if (scriptSetup) {
-			const definePropsType = scriptSetup.content.match(/defineProps<([\s\S]+?)>\s?\(\)/);
+			const codeWithoutComments = scriptSetup.content.replace(/\/\/.*|\/\*[\s\S]*?\*\//g, '');
+			const definePropsType = codeWithoutComments.match(/defineProps<([\s\S]+?)>\s?\(\)/);
 			const propsGeneric = scriptSetup.attrs.generic;
 			const propsGenericType = propsGeneric ? `<${propsGeneric}>` : '';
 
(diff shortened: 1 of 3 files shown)


phk422:
Hi @ematipico! Could you please review the code again? If everything is fine, could you merge it? I would like to resolve this issue as soon as possible.

ematipico:
@bluwy may I have a review from you too?

packages/integrations/vue/src/editor.cts
```diff
@@ -40,20 +41,19 @@ export function toTSX(code: string, className: string): string {
 				// TODO. Find a way to support generics when using defineProps without passing explicit types.
 				// Right now something like this `defineProps({ prop: { type: Array as PropType<T[]> } })`
 				//  won't be correctly typed in Astro.
-				const defineProps = scriptSetup.content.match(/defineProps\([\s\S]+\)/);
-
+				const defineProps = codeWithoutComments.match(/defineProps\([\s\S]+?\)/);
 				if (defineProps) {
 					result = `
-					import { defineProps } from '@vue/runtime-core';
+						import { defineProps } from 'vue';
```
bluwy, line 47: Is there a reason `@vue/runtime-core` is changed to `vue`. We should also remove the indent changes

packages/integrations/vue/test/toTsx.test.cts
bluwy, line 1: This file doesn't run as it's in `.cts`. It should be in `.js` and import from '../dist/editor.cjs`.

packages/integrations/vue/test/toTsx.test.cts
```diff
@@ -0,0 +1,45 @@
+import { toTSX } from '../src/editor.cts';
+import {  describe, it } from 'node:test';
+import { strictEqual } from 'node:assert'
```
bluwy, line 3: Our tests uses `import * as assert from 'node:assert/strict';`, I think we should refactor to that for consistency.

bluwy (commented):
Left a couple comments and the idea seems fine, but I think @Princesseuh has better insights on the tooling side of changes 😅

packages/integrations/vue/src/editor.cts
phk422: In my case with pnpm, I found that `@vue/runtime-core` might not be installed, In order to ensure the correct import of defineProps, I changed it to `vue` because it will be installed.

packages/integrations/vue/test/toTsx.test.cts
phk422: Thanks for your suggestion, I have changed it.

bluwy (approved):

ematipico (approved):
Thank you @phk422 for the fix 🙏

PR #10295, opened 2024-03-01
fix: client prerender fallback

rossrobino:
## Changes

Prerendering with the Speculation Rules API can fail---for example, certain Chrome extensions like LastPass can block prerendering, or browsers like Arc have disabled the feature. Currently in this case, neither prerendering or prefetching occurs. 

Adding a `link` tag in addition to the speculation rules script does not create an extra request. This change will append the speculation rules script, in addition to prefetching with `link` or `fetch` to ensure at least prefetching will occur. 

## Testing

No modifications needed.

## Docs


bluwy:
If those extensions block prerendering, wouldn't they also block prefetching? And if Arc disables the feature, wouldn't `HTMLScriptElement.supports('speculationrules')` be `false`? I'm a bit worried about this approach, ideally we trust prerendering as _the_ way for snappier navigations in the future, and wouldn't fallback to the old approach again.

rossrobino:
Thanks @bluwy, agreed that it doesn't seem like it should be necessary if the API is supported. 

Discussing this issue [on the Speculation Rules spec repo](https://github.com/WICG/nav-speculation/issues/162#issuecomment-1977066562) to ensure the best approach is implemented.

rossrobino:
Of course! Thanks for looking this over. [Domenic suggested](https://github.com/WICG/nav-speculation/issues/162#issuecomment-1977818473) still using the Speculation Rules API to prefetch in this case instead of a link tag due to the [benefits stated here](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API#using_prefetching). I have made those changes, so I think it's all set to fallback using the recommended approach now.

.changeset/warm-spoons-fetch.md
```diff
@@ -0,0 +1,7 @@
+---
+"astro": patch
+---
+
+Prerendering with the Speculation Rules API can fail---for example, [certain Chrome extensions can block prerendering](https://developer.chrome.com/blog/speculation-rules-improvements#chrome-limits). Currently in this case, neither prerendering or prefetching occurs. 
+
+Adding `prefetch` in addition to `prerender` within the speculation rules script does not create an extra request. This change will allow browsers to fallback to prefetching if prerendering is not supported.
```
bluwy, line 7:
```suggestion
Adds a prefetch fallback when using the `experimental.clientPrerender` option. If prerendering fails, which can happen if [Chrome extensions block prerendering](https://developer.chrome.com/blog/speculation-rules-improvements#chrome-limits), it will fallback to prefetching the URL. This works by adding a `prefetch` field to the `speculationrules` script, but does not create an extra request.
```

Tweaked this a bit to be more concise.

bluwy (approved):

ematipico (approved):

PR #10300, opened 2024-03-02
small fix to an incorrect .length check

FredKSchott:
## Changes

- Previous code was failing if you didn't have an assets directory at all

## Testing

- N/A

## Docs

- N/A

### DIFF
--- a/packages/db/src/core/integration/file-url.ts
+++ b/packages/db/src/core/integration/file-url.ts
@@ -72,12 +72,12 @@ export function fileURLIntegration(): AstroIntegration {
 						unlinks.push(fs.promises.unlink(url));
 					}
 					await Promise.all(unlinks);
+					// Delete the assets directory if it is empty.
+					// NOTE(fks): Ignore errors here because this is expected to fail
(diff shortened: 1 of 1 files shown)


packages/db/src/core/integration/file-url.ts
```diff
@@ -73,8 +73,8 @@ export function fileURLIntegration(): AstroIntegration {
 					}
 					await Promise.all(unlinks);
 					const assetDir = new URL(config.build.assets, config.outDir);
-					const assetFiles = await fs.promises.readdir(assetDir);
-					if (!assetFiles.length) {
+					const assetFiles = await fs.promises.readdir(assetDir).catch(() => []);
```
FredKSchott, line 76: this was also causing an issue if I didn't have an assets folder.

@bholmesdev I'm surprised that this integration is responsible for cleaning up the `_astro` folder? I'd assume this is astro's responsibility and already handled.

FredKSchott: nvm ignore me, fixed this a different way

PR #10312, opened 2024-03-04
Revamp db migrations

FredKSchott:
## Changes

- Currently, we ask you to manage a `migrations/` folder with the `astro db gen` command. There are some benefits to this system (control, explicit files written to disk) but the downsides are significant (it's an entire thing you need to learn and then manage, and for us it's a fairly complex thing that we need to maintain that can cause really nasty bugs for our users).
- This new system removes the migrations folder entirely, and instead syncs your schema on-demand. This is much less complex for both user and for us (see the ~400 lines of application logic removed in the diff). 
- The only case that this cannot handle naively -- and the main reason that we didn't pursue this originally -- is when you add and remove a column or table in the same `db push` command. We'll solve for this in other ways that don't require an entire migrations system.

### Progress
- [x] Remove migrations and update `verify`/`push`
- [x] Add `--force-reset` flag on `db push`
- [x] Add `deprecated` flag on tables and columns

## Testing

- Tests updated

## Docs

- Docs not yet written


FredKSchott:
Merging this without an approval so that we have more time to test and review the package starting on Monday.

PR #10316, opened 2024-03-04
fix(rendering): allow render instructions to propagate while rendering slots

### LINKED ISSUE #9996: Fragment + nested client:only components + slots problem
### Astro Info

```block
Astro                    v4.3.2
Package Manager          npm
                         @astrojs/react
                         simple-stream
```

_No response_

### Describe the Bug

below is the "not working" case, which is webpage display nothing.
```astro
---
import { Hello, Island, World } from "@components/demo.tsx";
---

<Island client:only>
  <Fragment slot="footer"><Hello client:only /></Fragment>
  <Fragment slot="header"><World client:only /></Fragment>
</Island>
```

but, if don't use Fragment, it works
```astro
<Island client:only>
  <a slot="footer"><Hello client:only /></a>
  <a slot="header"><World client:only /></a>
</Island>
```

also, below is work if nested components not defined `client:only`
```astro
<Island client:only>
  <Fragment slot="footer"><Hello /></Fragment>
  <Fragment slot="header"><World /></Fragment>
</Island>
```

I expected this should work
```
<Island client:only>
  <Fragment slot="footer"><Hello client:only /></Fragment>
  <Fragment slot="header"><World client:only /></Fragment>
</Island>
```

https://stackblitz.com/edit/github-qczw9r-srdto2?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes

- Fixes #9996
- The slot rendered first eagerly and stringified the island script with it. The slot was then put inside a template tag because the parent component is `client:only`.
- Island script should have been rendered right before the parent component itself. The `RenderDestination`-`RenderInstance` system makes sure that while things execute out of sequence, they get "stringified" in sequence.
- The reason was that slots are rendered in a temporary render destination and some chunk types are not handled there, falling back to stringifying immediately.
- The fix is to defer stringifying to the real destination, accumulating instructions in the `SlotString` untouched.

## Testing
- Added a case in an existing E2E fixture.

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/playwright.config.js
+++ b/packages/astro/playwright.config.js
@@ -27,16 +27,6 @@ const config = {
 		/* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
 		trace: 'on-first-retry',
 	},
-	projects: [
-		{
-			name: 'Chrome Stable',
-			use: {
-				browserName: 'chromium',
-				channel: 'chrome',
-				args: ['--use-gl=egl'],
-			},
-		},
-	],
 };
 
 export default config;
--- a/packages/astro/playwright.config.js
+++ b/packages/astro/playwright.config.js
@@ -27,6 +27,16 @@ const config = {
 		/* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
 		trace: 'on-first-retry',
 	},
+	projects: [
+		{
+			name: 'Chrome Stable',
+			use: {
+				browserName: 'chromium',
+				channel: 'chrome',
+				args: ['--use-gl=egl'],
+			},
+		},
+	],
 };
 
 export default config;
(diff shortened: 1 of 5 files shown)


packages/astro/playwright.config.js
```diff
@@ -27,16 +27,6 @@ const config = {
...
-	projects: [
-		{
-			name: 'Chrome Stable',
-			use: {
-				browserName: 'chromium',
-				channel: 'chrome',
-				args: ['--use-gl=egl'],
-			},
-		},
-	],
```
ematipico, line 39: Why was this removed?

ematipico (approved):
I would ask @bluwy and @matthewp to look at the PR. It seems good to me though

matthewp (approved):

packages/astro/playwright.config.js
lilnasy: I need this to run E2E tests locally. It shouldn't have been committed. I'll undo it.

bluwy (approved):

PR #10317, opened 2024-03-04
fix(rendering): prevent removal of necessary `<astro-slot>` elements

### LINKED ISSUE #8247: slot content disappears upon hydration when using polymorphic tags
### Astro info

```block
v2.10.14
```

Chrome, Firefox

### Describe the Bug

WHEN using polymorphic components (or even just assigning a tag name to a variable), AND
WHEN this polymorphic Astro component has a slot, AND
WHEN that slot is filled with a framework (Svelte) component which also has a slot, AND
WHEN the framework component has a client hydration directive:

The slot content doesn't appear.

If you refresh the page, you might be able to see the slot appear in prerendering and then disappear upon hydration.

Slot content should stay in the DOM after hydration.

https://stackblitz.com/edit/github-stsn4j?file=src%2Fpages%2Findex.astro,src%2Fcomponents%2FStuff.svelte,src%2Fcomponents%2FPoly.astro

### Participation

lilnasy:
## Changes

- Fixes #8247
- The check for whether to remove `<astro-slot>` happens multiple times - once per parent.
- When a distant parent is _not_ to be hydrated, `<astro-slot>` gets removed, even though a direct parent needs it.
- ~~The solution is to mark the tag as "definitely needed", so that distant parents don't evaluate it.~~
- ~~This is done by replacing `<astro-slot>` with `<astro-preserved-slot>` in the HTML temporarily.~~
- ~~`<astro-preserved-slot>` is converted back into `<astro-slot>` once there's no potential of being evaluated for removal - in the code for `RenderDestination`.~~
- Solution: https://github.com/withastro/astro/pull/10317#issuecomment-1983775885

## Testing
- Added a case to a pre-existing E2E fixture.

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -393,7 +393,7 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 				);
 			}
 			const renderedElement = renderElement('astro-island', island, false);
-			destination.write(renderedElement);
+			destination.write(markHTMLString(renderedElement));
 		},
 	};
 }
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -66,32 +66,11 @@ function isHTMLComponent(Component: unknown) {
 }
 
 const ASTRO_SLOT_EXP = /<\/?astro-slot\b[^>]*>/g;
-// same as ASTRO_SLOT_EXP, but only includes the tag name in the match while requiring that it still be surrounded by "<" and ">"
-const ASTRO_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-slot(?=\b[^>]*>)/g;
-// used to match temporary tags that will be replaced back with astro-slot
-const ASTRO_PRESERVED_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-preserved-slot(?=\b[^>]*>)/g;
-
 const ASTRO_STATIC_SLOT_EXP = /<\/?astro-static-slot\b[^>]*>/g;
-// same as ASTRO_STATIC_SLOT_EXP, but only includes the tag name in the match while requiring that it still be surrounded by "<" and ">"
-const ASTRO_STATIC_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-static-slot(?=\b[^>]*>)/g;
-// used to match temporary tags that will be replaced back with astro-static-slot
-const ASTRO_PRESERVED_STATIC_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-preserved-static-slot(?=\b[^>]*>)/g;
 
-function removeStaticAstroSlot(html: string, supportsAstroStaticSlot: boolean) {
+function removeStaticAstroSlot(html: string, supportsAstroStaticSlot = true) {
 	const exp = supportsAstroStaticSlot ? ASTRO_STATIC_SLOT_EXP : ASTRO_SLOT_EXP;
 	return html.replace(exp, '');
-}
-
-// An HTML string may be processed by the parent of a parent, and if it isn't to be hydrated, astro-slot tags will be incorrectly removed.
-// We rename them here so that the regex doesn't match.
-function preserveHydratableAstroSlot(html: string) {
-	return html.replaceAll(ASTRO_STATIC_SLOT_TAGNAME_EXP, 'astro-preserved-static-slot')
-		.replaceAll(ASTRO_SLOT_TAGNAME_EXP, 'astro-preserved-slot');
-}
-
-export function restoreHydratableAstroSlot(html: string) {
-	return html.replaceAll(ASTRO_PRESERVED_STATIC_SLOT_TAGNAME_EXP, 'astro-static-slot')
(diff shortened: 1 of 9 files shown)


packages/astro/test/astro-slot-with-client.test.js
```diff
@@ -18,9 +18,9 @@ describe('Slots with client: directives', () => {
 		assert.equal($('script').length, 1);
 	});
 
-	it('Astro slot tags are cleaned', async () => {
+	it('Astro slot tags are kept', async () => {
 		const html = await fixture.readFile('/index.html');
 		const $ = cheerio.load(html);
-		assert.equal($('astro-slot').length, 0);
+		assert.equal($('astro-slot').length, 1);
 	});
```
lilnasy, line 25: This test was checking for the conditions that _cause_ the bug. We need `<astro-slot>` elements to exist to allow slots to render after hydration.

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -66,12 +66,34 @@ function isHTMLComponent(Component: unknown) {
...
+const ASTRO_STATIC_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-static-slot(?=\b[^>]*>)/g;
+// used to match temporary tags that will be replaced back with astro-static-slot
+const ASTRO_PRESERVED_STATIC_SLOT_TAGNAME_EXP = /(?<=<\/?)astro-preserved-static-slot(?=\b[^>]*>)/g;
+
 function removeStaticAstroSlot(html: string, supportsAstroStaticSlot: boolean) {
 	const exp = supportsAstroStaticSlot ? ASTRO_STATIC_SLOT_EXP : ASTRO_SLOT_EXP;
 	return html.replace(exp, '');
 }
 
+// An HTML string may be processed by the parent of a parent, and if it isn't to be hydrated, astro-slot tags will be incorrectly removed.
```
ematipico, line 85: "parent of a parent" -> "grandparent"?

ematipico (approved):
The fix seems reasonable to me, but it would be great if @matthewp  or @bluwy could weigh in

matthewp (approved):
nice explanation! 

Given that this is a sticky part of the code, I think we should release this in its own patch so it's easier to spot regressions if they happen.

bluwy (commented):
Alternating the slots feels like working around the issue and could have a performance impact compared to if we can render it directly correctly. Could this be fixed instead by only stripping away `<astro-slot>` for non-slot/child templates? e.g.

https://github.com/withastro/astro/blob/1d9fb15a2d8642ac6c9f2547a213b91019038381/packages/astro/src/runtime/server/render/component.ts#L281-L287

Could we only replace `<astro-slot>` for non `childSlots` rendered strings?

lilnasy:
@bluwy Do you mean to make an exception for the "tagname as component" case?

bluwy:
I mean that `html` can be derived from several ways, e.g. `renderToStaticMarkup`, "tagname as component", etc. And I noticed the tests only tested for "tagname as component", and I wonder if the issue only arises in this way?

If so, I was thinking we can make the fix directly in the if block for the "tagname as component" handling. Or if not, then I think this PR is probably the easiest fix, since I'm not sure if there's a reliable way to say "this chunk belongs to a slot, and we skip replacing `astro-slots` in it. For other chunks, we do the replacement"

lilnasy:
> I noticed the tests only tested for "tagname as component", and I wonder if the issue only arises in this way?

I don't think the issue is limited to it. but I'll take another look.

natemoo-re (approved):
Works for me! If there's a cleaner way to do the transformations as Blu suggested, I'm all for it. But this solution is aligned with our current hacky code, so I'm cool if the RegExp approach is the one that gets merged.

lilnasy:
@bluwy 

First-party framework components are not affected because they signal using "supportsAstroStaticSlots" that they will render `astro-static-slot` for the non-hydration case.

A much simpler 90% fix could be assuming that renderers do that by default.

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -391,7 +392,8 @@ If you're still stuck, please open an issue on GitHub or join us at https://astr
 					})
 				);
 			}
-			destination.write(markHTMLString(renderElement('astro-island', island, false)));
+			const renderedElement = renderElement('astro-island', island, false);
+			destination.write(renderedElement);
```
bluwy, line 396: I think we still need to call `markHTMLString` since `renderElement` doesn't

lilnasy:
The simpler fix works.

Required slots could still be removed if there's a third-party renderer that explicitly sets the "supportsAstroStaticSlots" flag to `false`. Qwik sets the flag to `true`, and analog doesn't set it, so it's more like a 99% fix.

For Astro 5.0, we should remove the flag to assume renderers that will always distinguish between hydratable and non-hydratable slots on their own. This was originally intended for Astro 3.0, but wasn't tracked.

bluwy (approved):
Awesome! Thanks for looking into it. This looks a ton simpler and makes sense to me too.

packages/astro/src/runtime/server/render/component.ts
lilnasy: That's right! I left it out accidentally.

PR #10319, opened 2024-03-04
workaround(node ssr): cancellation support for renderToAsyncIterable

### LINKED ISSUE #10301: There was an uncaught error in the middle of the stream while rendering (TypeError: iterator.return is not a function)
### Astro Info

```block
Astro                    v4.4.8
Package Manager          npm
                         @astrojs/react
```

### Describe the Bug

The problem as described below, was first mentioned in #10184, but seems to be still an issue.

The following error is being logged sometimes:

```
There was an uncaught error in the middle of the stream while rendering /. TypeError: iterator.return is not a function
    at Object.cancel (node:internal/deps/undici/undici:666:34)
    at readableStreamDefaultControllerCancelSteps (node:internal/webstreams/readablestream:2281:39)
    at [kCancel] (node:internal/webstreams/readablestream:1057:12)
    at ensureIsPromise (node:internal/webstreams/util:192:19)
    at readableStreamCancel (node:internal/webstreams/readablestream:1882:5)
    at readableStreamReaderGenericCancel (node:internal/webstreams/readablestream:2022:10)
    at ReadableStreamDefaultReader.cancel (node:internal/webstreams/readablestream:880:12)
    at ServerResponse.<anonymous> (file:///app/dist/server/entry.mjs:1934:16)
    at ServerResponse.emit (node:events:517:28)
    at emitCloseNT (node:_http_server:1020:10)
```

```js
// dist/server/entry.mjs (lines: 1922-1947)

 static async writeResponse(source, destination) {
    const { status, headers, body } = source;
    destination.writeHead(status, createOutgoingHttpHeaders(headers));
    if (!body)
      return destination.end();
    try {
      const reader = body.getReader();
      destination.on("close", () => {
        reader.cancel().catch((err) => {
          console.error(
            `There was an uncaught error in the middle of the stream while rendering ${destination.req.url}.`,
            err
          );
        });
      });
      let result = await reader.read();
      while (!result.done) {
        destination.write(result.value);
        result = await reader.read();
      }
      destination.end();
    } catch {
      destination.end("Internal server error");
    }
  }
}
```

I have not been able to reproduce it consistently yet.
The stack trace doesn't point to any custom code, so it seems like it is a problem in the node code?

What is the `TypeError: iterator.return is not a function` it is referring to?

Anyone else experiencing this?

Could be related to #9614
Supposedly fixed by #10197 - but the error is still thrown for unknown reasons.

lilnasy:
## Changes

- Fixes #10301

## Testing
- Not sure about the conditions required for the error. Existing tests should pass.

## Docs
- Does not affect usage.

### DIFF
--- a/.changeset/nice-pets-tie.md
+++ b/.changeset/nice-pets-tie.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes an issue where a streaming SSR responses sometimes failed with "`iterator.result`" is not a function" on node-based adapters.
+Fixes an issue where streaming SSR responses sometimes failed with "`iterator.result` is not a function" on node-based adapters.


.changeset/nice-pets-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue where a streaming SSR responses sometimes failed with "`iterator.result`" is not a function" on node-based adapters.
```
lilnasy, line 5:
```suggestion
Fixes an issue where streaming SSR responses sometimes failed with "`iterator.result` is not a function" on node-based adapters.
```

matthewp (approved):
great fix, thanks!

ematipico (approved):

PR #10321, opened 2024-03-04
Add integrations API for db config/seed files

delucis:
## Changes

- Adds support for integrations adding db config and seed files via an `astro:db:setup` hook. This hook provides an `extendDb()` method to register the extra entrypoints.
- The hook is not typed in Astro core yet. Integration authors should use a `defineDbIntegration()` helper to get typing support.

## Testing

- Added a fixture that tests an integration. It’s a local “in-project integration”, so I also manually tested locally with a separate packed integration with npm-style module identifier.

## Docs

- First pass in the changeset, will liaise with @sarah11918 about getting this into the actual docs


florian-lefebvre:
Will astro db be part of core eventually or remain as a separate integration?

matthewp:
@florian-lefebvre Will astro db be part of core eventually or remain as a separate integration?

The goal would be for it to eventually be part of core, but we still have several things to work out before that's ready, so it's going to remain an integration for the near future at least.

matthewp:
@delucis can you explain why this needs to be a separate hook? I'm not against it, per se, but would like to understand the technical reasons for it.

delucis:
> @delucis can you explain why this needs to be a separate hook? I'm not against it, per se, but would like to understand the technical reasons for it.

The main issue is that `astro db` has components that run outside of the usual Astro environment — primarily its dedicated CLI commands — that need to access the tables for a project. If the `extendDb` API is in `astro:config:setup`, we need to call that hook for all integrations, providing (or mocking out) the full setup API in an environment where those APIs don't really exist. This leads to an issue on both ends:

- on db's end we're maintaining some faux options object in order to call the existing hook.

- on integrations' end, they're tricked into thinking they can maybe do regular integration stuff (most obviously perhaps, creating a virtual module), which they can't really, because the tables need to be available in a non-Vite environment.

That said, I think ideally this PR wouldn't make the types change included here. Will pick it up in the morning.

packages/db/src/core/integration/vite-plugin-db.ts
```diff
@@ -88,18 +94,27 @@ export function getConfigVirtualModContents() {
...
 	shouldSeed: boolean;
 }) {
-	const seedFilePaths = SEED_DEV_FILE_NAME.map(
+	const userSeedFilePaths = SEED_DEV_FILE_NAME.map(
 		// Format as /db/[name].ts
 		// for Vite import.meta.glob
 		(name) => new URL(name, getDbDirectoryUrl('file:///')).pathname
 	);
-
+	const resolveId = (id: string) => (id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
```
matthewp, line 111: Try using `normalizePath` from `'vite'` rather than `node:path`'s resolve

packages/db/src/core/integration/vite-plugin-db.ts
```diff
@@ -88,18 +94,27 @@ export function getConfigVirtualModContents() {
...
+	const integrationSeedFilePaths = seedFiles.map((pathOrUrl) =>
+		typeof pathOrUrl === 'string' ? resolveId(pathOrUrl) : pathOrUrl.pathname
+	);
+	const integrationSeedImports = integrationSeedFilePaths.map(
+		(filePath) => `() => import(${JSON.stringify(filePath)})`
+	);
```
matthewp, line 117: Very likely the problem is in this block of code. I'd try using `normalizePath` like about. Otherwise you can also pass in a file URL which is easier to work with.

matthewp:
API overall looks good to me. Could see some adjustment in the future but not anything important enough to do now.

delucis:
!preview db-integrations-support

delucis:
!preview db-integrations-support

.changeset/purple-poets-sin.md
```diff
@@ -0,0 +1,58 @@
+---
+"@astrojs/db": patch
```
natemoo-re, line 2: Might as well be a minor since `0.7.0` is slated!
```suggestion
"@astrojs/db": minor
```

natemoo-re (approved):
LGTM 🙌

matthewp (approved):

PR #10323, opened 2024-03-04
[ci] set `--tag` on release

natemoo-re:
## Changes

- We currently don't have a process for backporting fixes to previous release branches
- We _do_ currently release from the `main`, `next`, and `1-legacy` / `2-legacy` / `3-legacy` branches
- When we recently cut a release from `3-legacy`, it was published to `latest` on NPM! Thankfully we released from `main` shortly thereafter, but it could be annoying to debug if we forget to do that in the future.
- This PR automates the process so we don't have to think about it again. The code comments hopefully provide enough context.
- Closes PLT-1750

## Testing

Contrary to my typical YOLO approach, I actually minimized this workflow and tested it in a private repo. Everything works as expected!

## Docs

Currently, our actions are super undocumented. Currently attempting to kick off an effort to get more of these decisions documented, but until then, code comments should suffice.


.github/workflows/release.yml
```diff
@@ -21,7 +21,6 @@ env:
 jobs:
   changelog:
     name: Changelog PR or Release
-    if: ${{ github.repository_owner == 'withastro' }}
     runs-on: ubuntu-latest
```
natemoo-re, line 24:
```suggestion
		if: ${{ github.repository_owner == 'withastro' }}
    runs-on: ubuntu-latest
```

.github/workflows/release.yml
```diff
@@ -21,7 +21,6 @@ env:
 jobs:
   changelog:
     name: Changelog PR or Release
-    if: ${{ github.repository_owner == 'withastro' }}
```
bluwy, line 24: Is there a reason this is moved into a step below? Shouldn't we skip this workflow entirely on forks?

ematipico (approved):
Looks good to me

.github/workflows/release.yml
natemoo-re: Didn't mean to change this, but yes, can totally skip the whole workflow.

PR #10324, opened 2024-03-04
chore: delete `core/endpoint/index.ts`

lilnasy:
## Changes

- Removes a file that is not used anywhere since the pipeline refactor.

## Testing
- Chore. Does not affect behavior.

## Docs
- Does not affect usage.


bluwy (approved):
We may also need to update this document in the future: https://github.com/withastro/astro/blob/main/packages/astro/src/core/render/README.md

PR #10325, opened 2024-03-04
fix(Astro global): parity with `ctx.site`

### LINKED ISSUE #10303: context.site different from Astro.site (missing base / pathname)
### Astro Info

Astro                    v4.4.8
Package Manager          npm
Adapter                  none

_No response_

### Describe the Bug

I have my site settings set to:
```
{
    site: 'http://localhost:4321',
    base: '/dolphin'
}
```

logging `  const site = context.site` from within GET endpoint
```
{
  siteFromEndpoint: URL {
    href: 'http://localhost:4321/dolphin',
    origin: 'http://localhost:4321',
    protocol: 'http:',
    username: '',
    password: '',
    host: 'localhost:4321',
    hostname: 'localhost',
    port: '4321',
    pathname: '/dolphin',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  }
}
```

from within astro comopnent

```
const site = Astro.site
console.log({siteFromAstroSite:site})
```

```
{
  siteFromAstroSite: URL {
    href: 'http://localhost:4321/',
    origin: 'http://localhost:4321',
    protocol: 'http:',
    username: '',
    password: '',
    host: 'localhost:4321',
    hostname: 'localhost',
    port: '4321',
    pathname: '/',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  }
}
```

I'd expect these to be the same (containing the pathname dolphin

xxx

### Participation

lilnasy:
## Changes

- Fixes #10303
- Colocates code for the `Astro` global alongside that of `APIContext`.
- Removes `base` out of the picture while generating `manifest.site`.

## Testing
- Added to `astro-global.test.js` and `ssr-api-route.test.js`

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/core/base-pipeline.ts
+++ b/packages/astro/src/core/base-pipeline.ts
@@ -45,7 +45,7 @@ export abstract class Pipeline {
 		/**
 		 * Used for `Astro.site`.
 		 */
-		readonly site = manifest.site
+		readonly site = manifest.site ? new URL(manifest.site) : undefined,
 	) {
 		this.internalMiddleware = [
 			createI18nMiddleware(i18n, manifest.base, manifest.trailingSlash, manifest.buildFormat),
--- a/packages/astro/src/core/render/result.ts
+++ b/packages/astro/src/core/render/result.ts
@@ -1,60 +1,9 @@
-import type {
-	AstroGlobal,
-	AstroGlobalPartial,
-	Locales,
-	Params,
-	SSRElement,
-	SSRLoadedRenderer,
-	SSRResult,
-} from '../../@types/astro.js';
-import {
-	type RoutingStrategies,
-	computeCurrentLocale,
-	computePreferredLocale,
-	computePreferredLocaleList,
-} from '../../i18n/utils.js';
+import type { SSRResult } from '../../@types/astro.js';
 import { type ComponentSlots, renderSlotToString } from '../../runtime/server/index.js';
 import { renderJSX } from '../../runtime/server/jsx.js';
 import { chunkToString } from '../../runtime/server/render/index.js';
-import { clientAddressSymbol, responseSentSymbol } from '../constants.js';
-import { AstroCookies } from '../cookies/index.js';
 import { AstroError, AstroErrorData } from '../errors/index.js';
 import type { Logger } from '../logger/core.js';
-
-export interface CreateResultArgs {
-	/**
-	 * Used to provide better error messages for `Astro.clientAddress`
-	 */
-	adapterName: string | undefined;
-	/**
-	 * Value of Astro config's `output` option, true if "server" or "hybrid"
-	 */
-	ssr: boolean;
-	logger: Logger;
-	params: Params;
-	pathname: string;
-	renderers: SSRLoadedRenderer[];
-	clientDirectives: Map<string, string>;
-	compressHTML: boolean;
-	partial: boolean;
-	resolve: (s: string) => Promise<string>;
-	/**
-	 * Used for `Astro.site`
-	 */
-	site: string | undefined;
-	links: Set<SSRElement>;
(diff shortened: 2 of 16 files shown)


matthewp:
We're not supposed to modify the user's `site` config in any way. We don't add the base or do any other modifications. It's intended to be exactly what's in the user's config, only validated to be a URL. It looks like that's what you're doing here, I just want to confirm that's your expectation as well.

packages/astro/test/astro-global.test.js
```diff
@@ -2,14 +2,15 @@ import assert from 'node:assert/strict';
...
+import testAdapter from './test-adapter.js';
 
 describe('Astro Global', () => {
 	let fixture;
 
 	before(async () => {
 		fixture = await loadFixture({
 			root: './fixtures/astro-global/',
-			site: 'https://mysite.dev/blog/',
+			site: 'https://mysite.dev/subsite/',
```
lilnasy, line 13: A subpath to the site seems unnatural. It is possible it was not the intention of the orignal PR in the first place.
- https://github.com/withastro/astro/pull/3959.

packages/astro/src/core/endpoint/index.ts
```diff
@@ -85,6 +85,7 @@ export function createAPIContext({
 				return currentLocale;
 			}
 			if (locales) {
+				// @ts-ignore
```
ematipico, line 88: Use `ts-expect-error` instead. Also, since we own the code base, it's not good we need to suppress our errors. This is a  symptom that something isn't right

ematipico (commented):
Regarding the `TODO`s, what's the plan? Will create tickets for those or will you solve them in this PR?

lilnasy:
I don't know enough about how computeCurrentLocale works to tackle it myself. I can open a p1.

packages/astro/src/core/endpoint/index.ts
lilnasy: This file is not referenced by anything.
- #10324

ematipico:
> I don't know enough about how computeCurrentLocale works to tackle it myself. I can open a p1.

Linear tickets should be fine

packages/astro/src/core/render/result.ts
```diff
@@ -131,157 +80,3 @@ class Slots {
 		return outHTML;
 	}
 }
-
-export function createResult(args: CreateResultArgs): SSRResult {
```
matthewp, line 135: Why was this function moved out?

lilnasy: So that the code for the Astro global could live alongside that of APIContext and share the implementation details.

packages/astro/src/core/base-pipeline.ts
```diff
@@ -45,7 +45,7 @@ export abstract class Pipeline {
 		/**
 		 * Used for `Astro.site`.
 		 */
-		readonly site = manifest.site
+		readonly site = manifest.site ? new URL(manifest.site) : undefined,
```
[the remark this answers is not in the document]
lilnasy: You are right, it will throw. However, I don't think it needs extra code to handle the case. If it fails at runtime, it would also necessarily fail in dev and in build.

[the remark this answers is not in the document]
bluwy: Isn't site already validated at https://github.com/withastro/astro/blob/5fe98599a3fc6f912995d94ce783c32c06c0b44f/packages/astro/src/core/config/schema.ts#L95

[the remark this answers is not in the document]
lilnasy: Oh right, it is validated!

ematipico (approved):
Happy to merge once https://github.com/withastro/astro/pull/10325/files#r1513004270 is addressed

PR #10343, opened 2024-03-06
fix(toolbar): add anchor element to interactive elements

ematipico:
## Changes

This PR adds `a` to the list of element that can be interactive.

Interactive elements are usually marked with the `"widget"` role. [`link`](https://www.w3.org/TR/wai-aria-1.1/#link) role is a `command` role, which is a superclass that includes `"widget"`.

## Testing

I am happy to add a test with some guidance 

## Docs

N/A

### DIFF
--- a/.changeset/tough-moose-prove.md
+++ b/.changeset/tough-moose-prove.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes some false positive in the dev toolbar a11y audits, by adding the `a` element to the list of interactive elements.
--- a/packages/astro/e2e/dev-toolbar-audits.test.js
+++ b/packages/astro/e2e/dev-toolbar-audits.test.js
@@ -43,4 +43,18 @@ test.describe('Dev Toolbar - Audits', () => {
 		// Toggle app off
 		await appButton.click();
 	});
+
+	test('does not warn for non-interactive element', async ({ page, astro }) => {
+		await page.goto(astro.resolveUrl('/a11y-exceptions'));
+
+		const toolbar = page.locator('astro-dev-toolbar');
+		const appButton = toolbar.locator('button[data-app-id="astro:audit"]');
+		await appButton.click();
+
+		const auditCanvas = toolbar.locator('astro-dev-toolbar-app-canvas[data-app-id="astro:audit"]');
+		const auditHighlights = auditCanvas.locator('astro-dev-toolbar-highlight');
+
+		const count = await auditHighlights.count();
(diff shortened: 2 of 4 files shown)


Princesseuh (approved):
LGTM, just needs a changeset! You can add a test in `e2e/dev-toolbar-audits.test.js` if you want to, but I'm totally okay with not having one right now. We can make a full test suite for audits later

HiDeoo:
Following the discussion from earlier today, I opened #10346.

According to this [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#interactive_content), an anchor element is only considered interactive content if it has an `href` attribute:

> Some elements belong to this category only under specific conditions:
>
> `<a>`, if the `href` attribute is present
> `<audio>`, if the `controls` attribute is present
> ``, if the `usemap` attribute is present
> `<input>`, if the `type` attribute is not in the hidden state
> `<object>`, if the `usemap` attribute is present
> `<video>`, if the `controls` attribute is present

If my understanding is correct, the rule should check for the presence of the `href` attribute before considering the anchor element as interactive content (and same for the other elements with specific conditions). Any thoughts on this?

Princesseuh:
> Following the discussion from earlier today, I opened #10346.
> 
> According to this [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#interactive_content), an anchor element is only considered interactive content if it has an `href` attribute:
> 
> > Some elements belong to this category only under specific conditions:
> > `<a>`, if the `href` attribute is present
> > `<audio>`, if the `controls` attribute is present
> > ``, if the `usemap` attribute is present
> > `<input>`, if the `type` attribute is not in the hidden state
> > `<object>`, if the `usemap` attribute is present
> > `<video>`, if the `controls` attribute is present
> 
> If my understanding is correct, the rules should check for the presence of the `href` attribute before considering the anchor element as interactive content (and same for the other elements with specific conditions). Any thoughts on this?

Ah yeah, the optimal solution would be to check in the `match` property. An helper could be made to make that easier to re-use.

ematipico:
Changeset added. I suppose we can create the helper in another PR?

Princesseuh:
> Changeset added. I suppose we can create the helper in another PR?

Well it's not just the helper, the behaviour in this PR is incorrect since a `a` is only interactable under conditions

ematipico:
All right, I'll add the logic tomorrow

Princesseuh (approved):
Seems fine!

PR #10349, opened 2024-03-06
db: scaffold db files when using `astro add db`

natemoo-re:
## Changes

- Updates `astro add` CLI to scaffold out the `db/` directory
- Refactors the stub logic in `astro add` for improved maintainability
- We'll eventually want to expose scaffolding functionality for _any_ integration, but that will take some API design work. Right now, we just need to improve the `astro add db` experience. Baby steps!
- Bonus: removes default logger formatting when running any `astro add` command (AS INTENDED!)

## Testing

Tested manually because `astro add` isn't tested

## Docs

Will be documented in preparation for launch day

### DIFF
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -49,26 +49,44 @@ const ALIASES = new Map([
 	['solid', 'solid-js'],
 	['tailwindcss', 'tailwind'],
 ]);
-const ASTRO_CONFIG_STUB = `import { defineConfig } from 'astro/config';\n\nexport default defineConfig({});`;
-const TAILWIND_CONFIG_STUB = `/** @type {import('tailwindcss').Config} */
+
+const STUBS = {
+	ASTRO_CONFIG: `import { defineConfig } from 'astro/config';\n// https://astro.build/config\nexport default defineConfig({});`,
+	TAILWIND_CONFIG: `/** @type {import('tailwindcss').Config} */
 export default {
 	content: ['./src/**/*.{astro,html,js,jsx,md,mdx,svelte,ts,tsx,vue}'],
 	theme: {
 		extend: {},
 	},
 	plugins: [],
-}\n`;
-const SVELTE_CONFIG_STUB = `\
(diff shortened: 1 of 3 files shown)


packages/astro/src/cli/add/index.ts
```diff
@@ -49,26 +49,44 @@ const ALIASES = new Map([
...
+}\n`,
+	LIT_NPMRC: `\
 # Lit libraries are required to be hoisted due to dependency issues.
 public-hoist-pattern[]=*lit*
-`;
+`,
+	DB_CONFIG: `\
+import { defineDB } from 'astro:db';
+
+// https://astro.build/db/config
```
bholmesdev, line 76: TIL we have these redirects! Agreed we can hold off until a docs preview is ready

bholmesdev (approved):
You are DaBest 👏

packages/astro/src/cli/add/index.ts
```diff
@@ -183,9 +201,32 @@ export async function add(names: string[], { flags }: AddOptions) {
...
 			}
+			if (integrations.find((integration) => integration.id === 'db')) {
+				if (!existsSync(new URL('./db/', root))) {
+					logger.info(
+						'SKIP_FORMAT',
+						`\n  ${magenta(`Astro will scaffold ${green('./db/config.ts')}${magenta(' and ')}${green('./db/seed.ts')}${magenta(' files.')}`)}\n`
+					);
+
+					if (await askToContinue({ flags })) {
+						await fs.mkdir(new URL('./db/', root), { recursive: true });
```
ematipico, line 215: nit: `recursive` isn't needed because we already checked if the folder doesn't exist

ematipico (approved):

packages/astro/src/cli/add/index.ts
natemoo-re: They _will be added_ by https://github.com/withastro/astro.build/pull/1000! Just need the actual URLs we'll be using.

matthewp (approved):

PR #10350, opened 2024-03-06
Include virtual declaration in package

Fryuni:
## Changes

The module declaration was added but is not being included in the package. This fixes that.

## Testing

No change in logic

## Docs

N/A

### DIFF
--- a/.changeset/good-rats-bathe.md
+++ b/.changeset/good-rats-bathe.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Include missing file for packaging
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
Thanks!

.changeset/good-rats-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Include missing file for packaging
```
natemoo-re, line 5:
```suggestion
Includes `./virtual.d.ts` file that was previously unpublished
```

PR #10352, opened 2024-03-06
Add `--remote` flag for remote connection

bholmesdev:
## Changes

Require the `--remote` flag to dev or build with a studio connection. Use a local db with local seeding for standard `astro build` call.

- Add `--remote` check to dev and build
- Add `--remote` support to `shell` and `execute`
- Move table recreation from runtime to buildtime. Prevents build hanging failure and unblocks SSR app usage
- Refactor in-memory databases to a generate db hash. Solves the clobbering problem and remove all special logic for in-memory data
- Misc: remove bad `author3` and `author4` config in ticketing example

## Testing

- Add in-memory db patch for seeding to ensure tests still run

## Docs

https://docs-git-sarah-patch-6-astrodotbuild.vercel.app/en/guides/astro-db/#production-builds

### DIFF
--- a/packages/db/src/core/consts.ts
+++ b/packages/db/src/core/consts.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from 'node:crypto';
 import { readFileSync } from 'node:fs';
 
 export const PACKAGE_NAME = JSON.parse(
@@ -11,7 +12,7 @@ export const DB_TYPES_FILE = 'db-types.d.ts';
 
 export const VIRTUAL_MODULE_ID = 'astro:db';
 
-export const DB_PATH = '.astro/content.db';
+export const DB_PATH = `.astro/${process.env.TEST_RANDOM_ID ? randomUUID() : 'content.db'}`;
 
 export const CONFIG_FILE_NAMES = ['config.ts', 'config.js', 'config.mts', 'config.mjs'];
 
--- a/packages/db/src/core/consts.ts
+++ b/packages/db/src/core/consts.ts
@@ -12,7 +12,9 @@ export const DB_TYPES_FILE = 'db-types.d.ts';
 
 export const VIRTUAL_MODULE_ID = 'astro:db';
 
-export const DB_PATH = `.astro/${process.env.TEST_RANDOM_ID ? randomUUID() : 'content.db'}`;
+export const DB_PATH = `.astro/${
+	process.env.ASTRO_TEST_RANDOM_DB_ID ? randomUUID() : 'content.db'
+}`;
 
 export const CONFIG_FILE_NAMES = ['config.ts', 'config.js', 'config.mts', 'config.mjs'];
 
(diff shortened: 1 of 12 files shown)


FredKSchott:
LGTM! Can you add to other commands as well? 
- `astro dev --remote`
- `astro shell --remote`
- `astro execute --remote`

If there is no thing as local `astro shell` or local `astro execute` then it's okay to error if no `--remote` flag is found, asking you to add the `--remote` to remove the error.

packages/db/src/core/consts.ts
```diff
@@ -1,17 +1,19 @@
...
-export const RUNTIME_DRIZZLE_IMPORT = JSON.stringify(`${PACKAGE_NAME}/runtime/drizzle`);
 export const RUNTIME_CONFIG_IMPORT = JSON.stringify(`${PACKAGE_NAME}/runtime/config`);
 
 export const DB_TYPES_FILE = 'db-types.d.ts';
 
 export const VIRTUAL_MODULE_ID = 'astro:db';
 
-export const DB_PATH = '.astro/content.db';
+export const DB_PATH = `.astro/${
+	process.env.ASTRO_TEST_RANDOM_DB_ID ? randomUUID() : 'content.db'
```
ematipico, line 16: Is this regenerated when some HMR change is triggered or when we restart the server? Is it fine if it is regenerated?

packages/db/src/runtime/index.ts
```diff
@@ -16,7 +16,8 @@ export { sql };
 export type SqliteDB = LibSQLDatabase;
 export type { Table } from './types.js';
 export { createRemoteDatabaseClient, createLocalDatabaseClient } from './db-client.js';
-export { seedLocal } from './queries.js';
+// only used for in-memory db in test fixtures
+export { recreateTables } from './queries.js';
```
ematipico, line 20: Isn't it worth investigating if we can pull this function using dynamic import?

packages/db/src/core/consts.ts
bholmesdev: This _must_ be regenerated for our test runners! See comment on the basics test. In short, the test runner starts multiple dev servers in parallel if it finds this optimal (does not occur locally for some reason). If there's multiple servers pointing to the same db file, seed data will be clobbered. We also can't inject a db name as an environment variable because then the environment variable would get clobbered by each test run 😅 This is the best compromise I could find

packages/db/src/runtime/index.ts
bholmesdev: Good catch, this can be removed now that in-memory dbs are no longer used

packages/db/src/core/consts.ts
bholmesdev: Ah, if you're wondering if the ID is stable for the lifetime of the test, it seems to be. Can't think of an HMR trigger

ematipico: Awesome, thank you for the detailed explanation :) just making sure if that's what we need

ematipico (approved):

bholmesdev:
@FredKSchott done! Added the `--remote` treatment to dev, shell, and execute. Should make local and remote debugging much more powerful 👏

matthewp (approved):

PR #10361, opened 2024-03-07
Add batch support

bholmesdev:
## Changes

Add support for `db.batch()` to run multiple queries in a single transaction.
- Bump Drizzle to 0.29.5 (minor bump)
- Implement batching for the remote adapter. Nothing to do for local

## Testing

- Add `db.batch()` call to basics test fixture

## Docs

N/A

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -62,7 +62,7 @@
     "@libsql/client": "^0.4.3",
     "async-listen": "^3.0.1",
     "deep-diff": "^1.0.2",
-    "drizzle-orm": "^0.28.6",
+    "drizzle-orm": "0.29.5-7888ad6",
     "kleur": "^4.1.5",
     "nanoid": "^5.0.1",
     "open": "^10.0.3",
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -62,7 +62,7 @@
     "@libsql/client": "^0.4.3",
     "async-listen": "^3.0.1",
     "deep-diff": "^1.0.2",
-    "drizzle-orm": "0.29.5-7888ad6",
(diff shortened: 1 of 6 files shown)


bholmesdev:
!preview db-batch

packages/db/package.json
```diff
@@ -62,7 +62,7 @@
     "@libsql/client": "^0.4.3",
     "async-listen": "^3.0.1",
     "deep-diff": "^1.0.2",
-    "drizzle-orm": "^0.28.6",
+    "drizzle-orm": "0.29.5",
```
delucis, line 65: Any specific notes as to why this is pinned to an exact version?

delucis (commented):
I’m not familiar with the code, so hard to review, but left one question.

In the PR description docs says “N/A” but if this is something we support, wouldn’t we need to document it?

packages/db/package.json
bholmesdev: Ah, it shouldn't be pinned! Good catch

bholmesdev:
@delucis re-documentation: I'm using Sarah's suggestion to outsource to the Drizzle docs as much as possible to avoid explaining all of their features. There _is_ an argument to document `db.batch()` as a best practice for seeding though. There were mixed thoughts on recommending batch early, so I'll raise in the docs discussion thread.

delucis (approved):
> I'm using Sarah's suggestion to outsource to the Drizzle docs

Ah, OK — assumed as _we_ were adding support, there might be something Astro-specific to the API separate from Drizzle’s batch API. If that’s not the case, then all’s good!

Approving from a place of ignorance — a `batch` made in heaven 🙌

PR #10374, opened 2024-03-08
export DB types

itsMapleLeaf:
We need to use these types in the dashboard. Since this technically makes the internal-ish types a part of the package's public interface, I'm open to alternate approaches if we want to avoid that

### DIFF
--- a/.changeset/silver-foxes-protect.md
+++ b/.changeset/silver-foxes-protect.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Expose DB utility types from @astrojs/db/types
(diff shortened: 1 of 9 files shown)


bholmesdev:
Yeah this is a tough one... Because exposing these types could open VS Code completions we'd rather avoid. Is this worth a separate package, or a copy / paste in the meantime?

itsMapleLeaf:
> Yeah this is a tough one... Because exposing these types could open VS Code completions we'd rather avoid. Is this worth a separate package, or a copy / paste in the meantime?

Making a separate package sounds like the better solution long-term, so I'd rather do that from the start

bholmesdev (approved):
Alright, I moved all schema and utility types to a separate schemas.js file. Comfortable exposing types as an export now! DBetter merge soon 🙏

PR #10376, opened 2024-03-09
Adds Tailwind v4 support

Eveeifyeve:
# PR ONHOLD: [commit about it here](https://github.com/withastro/astro/pull/10376#issuecomment-1991422308)

This PR is for the offical support for [tailwind v4](https://tailwindcss.com/blog/tailwindcss-v4-alpha)

## Changes

- Adds changes to `package.json` to export the new v4 changes and adds `@tailwind/vite` package.
- Adds `v4.ts` and `base-v4.css` for tailwind v4 support.

## Testing

Testing done. Only type errors for `AstroIntergration`, `config`, `updateConfig` and `injectScript`.

## Docs

You will need to update the docs to tell people how to use the v4 edition of the TailwindCSS to change the import to`@astrojs/tailwind/v4`


packages/integrations/tailwind/src/v4.ts
```diff
@@ -0,0 +1,37 @@
...
+    const applyBaseStyles = options?.applyBaseStyles ?? true;
+
+    return {
+        name: '@astrojs/tailwind',
+        hooks: {
+            'astro:config:setup': async ({ updateConfig, injectScript }) => {
+                // Inject the Tailwind postcss plugin
+                updateConfig({
+                    vite: {
+                      plugins: [...tailwindcss()],
```
jdtjenkins, line 25: This doesn't need to be spread, it can just be

```suggestion
                      plugins: [tailwindcss()],
```

packages/integrations/tailwind/src/v4.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export default function tailwindIntegration(options?: TailwindOptions): AstroIntegration {
+    const applyBaseStyles = options?.applyBaseStyles ?? true;
+
+    return {
+        name: '@astrojs/tailwind',
+        hooks: {
+            'astro:config:setup': async ({ updateConfig, injectScript }) => {
+                // Inject the Tailwind postcss plugin
```
jdtjenkins, line 22:
```suggestion
                // Inject the Tailwind vite plugin
```

packages/integrations/tailwind/src/v4.ts
```diff
@@ -0,0 +1,37 @@
...
+				updateConfig({
+					vite: {
+						plugins: [tailwindcss()],
+						css: { transformer: 'lightningcss' },
+					},
+				});
+
+				if (applyBaseStyles) {
+					// Inject the Tailwind base import
+					injectScript('page-ssr', `import '@astrojs/tailwind/base-v4.css';`);
```
alexanderniebuhr, line 32: Couldn't this be problematic, because it would inject the import onto every page?

The import should only happen on the most top level of any given page. So if you have a layout the import should happen inside the layout and not the page.

I'm unsure if vite optimizes that, but I don't think it does.

Eveeifyeve: Well vite does optimise by default unless you specify to not minify not only that you can also have the ability to disable it by doing applybasestyles to false.

alexanderniebuhr: Yeah. However that is not the point, I tried to make. Just because it currently is build that way it doesn't mean it is the best solution. However I really haven't tested if it has an impact - if the global stylesheet is imported on every page, instead of just once.

Just my personalfeeling but I guess it does matter, and I personally would mind. And if it does I would at least expect an note about that in docs, so users are aware of it. 

But I'll leave it with that. :)

Eveeifyeve: Well with v4 they are now using @import "tailwindcss" instead of `@tailwind` so it has to be separate since v4 is different to v3.

Eveeifyeve: I felt like this change is completely useless since V4 has different stuff now. You Check is out [here](https://tailwindcss.com/blog/tailwindcss-v4-alpha)

Eveeifyeve: And also a lot of people don't want to apply the base styles so we give them the choice of that. So Really unnecessary and not best pratices to use the same css file when v4 uses import.

Eveeifyeve:
I have talked it out the the discord under `#contribute` channel it has come to a conclusion that @alexanderniebuhr was correct about the sizing, but:

A. I can't measure the size clearly
B. It's a issue with vite and rollup

Eveeifyeve:
Well all of the code LGTM I can't wait till I can use tailwind v4 in astro!

bluwy:
This was discussed in the [#contribute channel](https://discord.com/channels/830184174198718474/845430950191038464/1216725391587414016) yesterday, and there was discussion of maybe deprecating/removing `applyBaseStyles`, which makes even more sense for tailwind v4 since the config is in a CSS file.

In which case, do we need an Astro integration that simply applies the `@tailwind/vite` plugin? Users can do that themselves in the `vite.plugins` config option which requires equal amount of code.

Eveeifyeve:
Well we still need the extensions cause less code = less cluttered and most people perfer to use the integrations I felt like we could just make it add a tw.css file to css and make index page import it.

Eveeifyeve:
Yeah I get the reason why but at the same time wouldn't users rather have a command to add the part that does it for them cause remember developers are lazy. Lets take this in a scenario where you are a developer using a framework do you want to use a simple command that does for you or do you want to hard code it by your self which takes more time. Which one would you choose?

Eveeifyeve:
Pluss compatibility with sass requires one file for my end for every project to add tailwind support.

Eveeifyeve:
I have started a [thread\(https://discord.com/channels/830184174198718474/845430950191038464/1217061142812168213) in the discord to talk about this extensively but I really agree to a comment from the [#contibute channel](https://discord.com/channels/830184174198718474/845430950191038464/1217059777822068767) from @alexanderniebuhr saying maybe we should still keep it there but tell the user to import it

Eveeifyeve:
The PR Is ON HOLD since it's not really required as Tailwind v4 Is in alpha when Tailwind V4 is in schedule for it's beta or stable release we would reconsider the implementation. For now if you want to use tailwind v4 I would use the [astro-tailwind-v4 community intergration](https://github.com/jdtjenkins/astro-tailwind-v4/) for now.

Any more discussions can be located on the [discord thread here](https://discord.com/channels/830184174198718474/1217061142812168213) for any issues, questions or other things about tailwind v4 support

bluwy:
Let's wait for tailwind v4 to stabilise, or until there's a reason v4 requires additional Astro setup before add some support for it. Closing this for now.

PR #10377, opened 2024-03-09
WIP adds POC for injectAssets

alexanderniebuhr:
**:warning: Looking for Guadiance / Feedback, before investing more time into this! https://discord.com/channels/830184174198718474/845430950191038464/1216122199518482662**

## Changes

- this is a POC for `injectAssets` inside the `astro:config:setup` hook
- goal is to allow integrations define entrypoints which are copied as is to the `build.assets` folder
- `injectRoute` & `injectScripts` are not suficient for such use-case
- see Discord #integration channel for some more reference and at least 3-4 people being in favor of this feature
- **it won't be a good assumption to say all assets should be injected into _astro, he final solution should have a config to allow configurable target folders for each injected asset**
- **the final solution should also have some opt-in hashing mechanism**
- **the final solution should also be able to return the URL / path for the injected asset**

some quote from Discord: 
 
> hey, I'm an integration, and this asset here will be a part of the site that I'll need to be able to access from the content I generate - can you let me know the URL this will be accessible under from the client side?

### DIFF
--- a/packages/astro/src/core/build/index.ts
+++ b/packages/astro/src/core/build/index.ts
@@ -1,6 +1,6 @@
 import fs from 'node:fs';
 import { performance } from 'node:perf_hooks';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { blue, bold, green } from 'kleur/colors';
 import type * as vite from 'vite';
 import type {
@@ -33,6 +33,8 @@ import { collectPagesData } from './page-data.js';
 import { staticBuild, viteBuild } from './static-build.js';
 import type { StaticBuildOptions } from './types.js';
 import { getTimeStat } from './util.js';
+import { basename } from 'node:path';
+import { appendForwardSlash } from '../path.js';
 
 export interface BuildOptions {
 	/**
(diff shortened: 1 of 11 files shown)


packages/astro/src/core/build/index.ts
```diff
@@ -53,6 +55,42 @@ export interface BuildOptions {
...
+	settings: AstroSettings;
+	/** Current working directory */
+	cwd?: string;
+}) {
+	const { config } = settings;
+	for (const injectedAsset of settings.injectedAssets) {
+		const { entrypoint } = injectedAsset;
+		let resolved: string;
+		try {
+			resolved = require.resolve(entrypoint, { paths: [cwd || fileURLToPath(config.root)] });
```
natemoo-re, line 72: Hmm… I think we want to resolve these relative to the integration rather than the cwd or project root? That’s my understanding at least.

packages/astro/src/core/build/index.ts
```diff
@@ -33,6 +33,8 @@ import { collectPagesData } from './page-data.js';
 import { staticBuild, viteBuild } from './static-build.js';
 import type { StaticBuildOptions } from './types.js';
 import { getTimeStat } from './util.js';
+import { basename } from 'node:path';
```
natemoo-re, line 36: If it is at all possible to avoid working with paths, I would highly recommend sticking with URLs for the implementation. Windows always bites us.

packages/astro/src/core/build/index.ts
```diff
@@ -210,6 +248,20 @@ class AstroBuilder {
 		});
 		this.logger.debug('build', timerMessage('Additional assets copied', this.timer.assetsStart));
 
+		createInjectedAssets({ settings: this.settings }, this.logger);
+		// console.log('DEBUG', this.settings.resolvedInjectedAssets);
+		this.logger.info('assets', 'Injecting assets...');
+		const assetsPath = this.settings.config.build.assets;
+		for (const resolvedAsset of this.settings.resolvedInjectedAssets) {
+			const assetURL = new URL(
+				`./${assetsPath}/${basename(fileURLToPath(resolvedAsset.resolvedEntryPoint))}`,
```
natemoo-re, line 257: If the `assetsPath` is used, it will likely always be under `_astro`. I think that’s probably correct, but goes against one of your stated goals that the URL could be anything.

natemoo-re: Also should we rely on the resolved filepath to determine the asset name? I’d expect integration authors to refine the hosted path.

natemoo-re (commented):
Extremely in favor of adding this to unblock y’all! Some initial thoughts to hopefully help move you along.

alexanderniebuhr:
Checking in.. Haven't forgot this, will come back to this soon :)

alexanderniebuhr:
Played a bit more with it, and I don't think this is the correct approach.. Need to take this back to think about this a little more.. We need something which has a good DX, and maybe fixing `injectRoutes` is the better way..

PR #10378, opened 2024-03-10
[@astrojs/db] handle undefined snapshot from API

FredKSchott:
## Changes

- handle undefined snapshot from API
- This is an upcoming change to the API response format that will go out on Monday, along with the release of this PR

## Testing

- N/A (tested manually)

## Docs

- N/A


FredKSchott:
Merging so that it ends up in tomorrow morning's release

PR #10379, opened 2024-03-10
fix(routing): partially dynamic segments are being truncated

### LINKED ISSUE #10122: Redirects for dynamic routes break when files with same name but different extensions exist
### Astro Info

```block
Astro                    v4.3.7
Package Manager          bun
Adapter                  none
                         @astrojs/sitemap
```

_No response_

### Describe the Bug

I am developing a blog based on the [Astro Starter Kit: Blog](https://github.com/withastro/astro/tree/main/examples/blog). The articles can be fetched at `/blog/[...slug]`, so I've made it possible to fetch the OGP images of the articles at `/blog/[...slug].png`. 

The problem starts here. I add a redirect from `/posts/[...slug]` to `/blog/[...slug]` in the Astro Config. However, when I access `/posts/[...slug]`, it redirects to `/blog/[...slug].png` instead of `/blog/[...slug]`.

Access to `/posts/[...slug]` is redirected to `/blog/[...slug]`.

https://stackblitz.com/edit/github-ndv9vp

### Participation

1574242600:
## Changes
fix #10122
> partially dynamic segments are being truncated and affecting routing.
## Testing
units/routing/manifest.test.js


packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -899,3 +893,17 @@ function computeRoutePriority(config: AstroConfig): RoutePriorityOverride {
...
+
+function joinSegments(segments: RoutePart[][]): string {
+	const arr = segments.map(function d(segment): string {
+		const [{ dynamic, content }] = segment;
+
+		return segment.length > 1
+			? segment.map((s) => d([s])).join('')
+			: dynamic
+				? `[${content}]`
+				: content;
```
Fryuni, line 905: Spread segments should be `[...${content}]`

Fryuni:
I reviewed it from my phone, so it wat just a quick look over. I'll review it more carefully later today.

Thanks for the PR!

.changeset/five-colts-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix the issue of partially truncated dynamic segments.
```
florian-lefebvre, line 5:
```suggestion
Fixes a routing issue with partially truncated dynamic segments.
```
Would be great to add an example as well maybe?

packages/astro/src/core/routing/manifest/create.ts
1574242600: `content` of rest parameter is `...xxx`,i'm not entirely certain whether it requires separate handling.

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -899,3 +893,17 @@ function computeRoutePriority(config: AstroConfig): RoutePriorityOverride {
 	}
 	return 'legacy';
 }
+
+function joinSegments(segments: RoutePart[][]): string {
+	const arr = segments.map(function d(segment): string {
+		const [{ dynamic, content }] = segment;
+
+		return segment.length > 1
+			? segment.map((s) => d([s])).join('')
```
lilnasy, line 902: Could you confirm I have got this right?

This one line is the fix.

Previously, we ignored all but the first `RoutePart`.

Now, each `RoutePart` gets passed to this `d()` function. And the function ultimately still wraps the dynamic `RouteParts` in square brackets same as it was done before.

1574242600: Sure thing! You've got it right.
Would there be any issues with this?

lilnasy: No, this is great - just took me a while to understand the recursive map, so I thought I'd double check.

lilnasy (approved):
Looks great to me! Great work on the test as well.

packages/astro/test/units/routing/manifest.test.js
```diff
@@ -144,8 +144,7 @@ describe('routing - createRouteManifest', () => {
 
 		assertRouteRelations(getManifestRoutes(manifest), [
 			['/', '/[...rest]'],
-			['/static', '/static-'],
-			['/static-', '/[dynamic]'],
```
Fryuni, line 148: This check shouldn't be removed, it should no longer be truncated but still be there

Fryuni (changes requested):
I think the recursive mapping is more confusing than needed and either two normal for loops or two nested maps would make it easier to comprehend.

There is one test check that was removed instead of updated.

Besides that, looking good! Sorry for the delay in the review

packages/astro/test/units/routing/manifest.test.js
lilnasy: It is combined into a single rule, isn't it?

Fryuni: No, this is checking two different rules
- `/static` comes before `/static-[dynamic]` (fully static vs mixed static + dynamic)
- `/static-[dynamic]` comes before `/[dynamic]` (fully dynamic vs mixed static + dynamic)

bluwy (approved):

Fryuni (approved):
LGTM

ematipico (approved):

arjendevos:
Issue still exists in version `"astro": "^4.7.1"`

Using it like this:
```ts
redirects: {
    "/nl/blog/[slug]": "/blog/[slug]",
  },
```

but it redirects to the `.png` version

PR #10381, opened 2024-03-11
Use static imports for integration seed files in `astro:db`

delucis:
## Changes

- Switches from dynamic to static imports for integration seed files in the `astro:db` virtual module.
- Fixes a bug where builds would fail silently when using integrations.

## Testing

Added a simple test with build output for the integrations fixture and also tested `astro build` manually locally.

## Docs

N/A — bug fix

### DIFF
--- a/packages/db/src/core/integration/vite-plugin-db.ts
+++ b/packages/db/src/core/integration/vite-plugin-db.ts
@@ -114,6 +114,9 @@ export function getLocalVirtualModContents({
 		(name) => new URL(name, getDbDirectoryUrl('file:///')).pathname
 	);
 	const resolveId = (id: string) => (id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
+	// Use top-level imports to correctly resolve `astro:db` within seed files.
+	// Dynamic imports cause a silent build failure,
+	// potentially because of circular module references.
 	const integrationSeedImportStatements: string[] = [];
 	const integrationSeedImportNames: string[] = [];
 	seedFiles.forEach((pathOrUrl, index) => {
--- a/packages/db/src/core/integration/vite-plugin-db.ts
+++ b/packages/db/src/core/integration/vite-plugin-db.ts
@@ -114,17 +114,20 @@ export function getLocalVirtualModContents({
 		(name) => new URL(name, getDbDirectoryUrl('file:///')).pathname
 	);
 	const resolveId = (id: string) => (id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
-	const integrationSeedFilePaths = seedFiles.map((pathOrUrl) =>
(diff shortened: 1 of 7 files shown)


packages/db/src/core/integration/vite-plugin-db.ts
```diff
@@ -114,17 +114,20 @@ export function getLocalVirtualModContents({
 		(name) => new URL(name, getDbDirectoryUrl('file:///')).pathname
 	);
 	const resolveId = (id: string) => (id.startsWith('.') ? resolve(fileURLToPath(root), id) : id);
-	const integrationSeedFilePaths = seedFiles.map((pathOrUrl) =>
-		typeof pathOrUrl === 'string' ? resolveId(pathOrUrl) : pathOrUrl.pathname
-	);
-	const integrationSeedImports = integrationSeedFilePaths.map(
-		(filePath) => `() => import(${JSON.stringify(filePath)})`
-	);
+	const integrationSeedImportStatements: string[] = [];
```
bholmesdev, line 117: Maybe leave a comment on why we avoided dynamic imports? Something like: "Use top-level imports to correctly resolve astro:db. Dynamic imports cause a silent build failure, potentially because of circular module references.'

bholmesdev (approved):
How integracious of you to make this fix 🙇

packages/db/src/core/integration/vite-plugin-db.ts
delucis: Good idea! Will add this:

```suggestion
	// Create static imports block to correctly resolve `astro:db` in integration seed files.
	// Dynamic imports cause a silent build failure, potentially because of circular module references.
	const integrationSeedImportStatements: string[] = [];
```

delucis: Oop, see you’ve got it! Go Ben 🙌

packages/db/test/integrations.test.js
```diff
@@ -10,10 +10,10 @@ describe('astro:db with integrations', () => {
...
 	});
 
-	// Note(bholmesdev): Use in-memory db to avoid
-	// Multiple dev servers trying to unlink and remount
-	// the same database file.
-	process.env.TEST_IN_MEMORY_DB = 'true';
+	// Note (@bholmesdev) generate a random database id on startup.
+	// Ensures database connections don't conflict
+	// when multiple dev servers are run in parallel on the same project.
+	process.env.ASTRO_TEST_RANDOM_DB_ID = 'true';
```
ematipico, line 16: Where was this changed?

bholmesdev: Ah, we added the check here when we generate the database URL: https://github.com/withastro/astro/blob/main/packages/db/src/core/consts.ts#L16

Is this what you are asking?

bholmesdev: Update: I just removed the need for this code 😅 No need for the comment!

ematipico (approved):

PR #10385, opened 2024-03-11
Fix db for projects without a seed file or with integrations

delucis:
## Changes

- Fixes `astro:db` integration support when an Astro project doesn’t include a user-defined seed file.
- Source of the bug was this change: https://github.com/withastro/astro/pull/10352/files#diff-4d4325b62adab978b9a1910a7ae9742c5269220b88f3b4372ff1a2a7d7fc9704R64-R70
- This moved the table creation to be triggered when a user’s seed file is loaded. This meant in the absence of a user seed file, tables are never created.
- The fix is moving `recreateTables()` back to runtime, called each time we `seedLocal()`. @bholmesdev had some concerns about this initially but I think these were resolved in #10381.
- To help keep things organised, I’ve split `seedLocal()` into its own file.

## Testing

- Added a fixture where an integration configures and seeds tables and no user config is present.
- Added a fixture where there is no seed file at all.
- Both tests include testing the build output.

## Docs

N/A — bug fix.

### DIFF
--- a/.changeset/calm-roses-camp.md
+++ b/.changeset/calm-roses-camp.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Fixes support for integrations configuring `astro:db` and for projects that use `astro:db` but do not include a seed file.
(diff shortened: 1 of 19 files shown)


bholmesdev (approved):
No notes. Exseeded my expectations!

PR #10386, opened 2024-03-11
fix: cloudflare issue

alexanderniebuhr:
## Changes

- this should fix the IIRC only issue when deploying to Cloudflare SSR
- https://discord.com/channels/830184174198718474/1134657186786725928/1215959050886447194

### DIFF
--- a/packages/db/src/runtime/db-client.ts
+++ b/packages/db/src/runtime/db-client.ts
@@ -5,7 +5,8 @@ import { drizzle as drizzleLibsql } from 'drizzle-orm/libsql';
 import { drizzle as drizzleProxy } from 'drizzle-orm/sqlite-proxy';
 import { z } from 'zod';
 
-const isWebContainer = !!process.versions?.webcontainer;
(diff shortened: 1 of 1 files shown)


packages/db/src/runtime/db-client.ts
```diff
@@ -5,7 +5,8 @@ import { drizzle as drizzleLibsql } from 'drizzle-orm/libsql';
 import { drizzle as drizzleProxy } from 'drizzle-orm/sqlite-proxy';
 import { z } from 'zod';
 
-const isWebContainer = !!process.versions?.webcontainer;
+let isWebContainer = false;
+if (process) isWebContainer = !!process.versions?.webcontainer;
```
Fryuni, line 9:
```suggestion
const isWebContainer = typeof process !== 'undefined' && !!process.version?.webcontainer;
```

alexanderniebuhr:
turns out it's only an issue with v10 of the cloudflare adapter, which is not released yet.

PR #10387, opened 2024-03-11
[db] Add success=false handling to API entrypoints

FredKSchott:
## Changes

- We had an issue where the server was incorrectly returning a `200` on error responses.
- This is a bit redundant if everything is working well, but a good line of defense in case anything isn't.

## Testing

- n/a

## Docs

- n/a

### DIFF
--- a/packages/db/src/core/cli/commands/push/index.ts
+++ b/packages/db/src/core/cli/commands/push/index.ts
@@ -82,11 +82,24 @@ async function pushSchema({
 		return new Response(null, { status: 200 });
 	}
 	const url = new URL('/db/push', getRemoteDatabaseUrl());
-	return await fetch(url, {
+	const response = await fetch(url, {
 		method: 'POST',
 		headers: new Headers({
 			Authorization: `Bearer ${appToken}`,
 		}),
 		body: JSON.stringify(requestBody),
 	});
+	if (!response.ok) {
+		console.error(`${url.toString()} failed: ${response.status} ${response.statusText}`);
+		console.error(await response.text());
+		throw new Error(`/db/push fetch failed: ${response.status} ${response.statusText}`);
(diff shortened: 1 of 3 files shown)


packages/db/src/core/cli/commands/push/index.ts
```diff
@@ -82,11 +82,16 @@ async function pushSchema({
...
 	const url = new URL('/db/push', getRemoteDatabaseUrl());
-	return await fetch(url, {
+	const response = await fetch(url, {
 		method: 'POST',
 		headers: new Headers({
 			Authorization: `Bearer ${appToken}`,
 		}),
 		body: JSON.stringify(requestBody),
 	});
+	const result = await response.json();
```
ematipico, line 92: `.json()` could throw an error, we should use a `try/catch`

packages/db/src/core/cli/commands/push/index.ts
```diff
@@ -82,11 +82,24 @@ async function pushSchema({
...
 		method: 'POST',
 		headers: new Headers({
 			Authorization: `Bearer ${appToken}`,
 		}),
 		body: JSON.stringify(requestBody),
 	});
+	if (!response.ok) {
+		console.error(`${url.toString()} failed: ${response.status} ${response.statusText}`);
+		console.error(await response.text());
+		throw new Error(`/db/push fetch failed: ${response.status} ${response.statusText}`);
```
bholmesdev, line 95: Yikes... think I found our "polish" ticket after launch 😉

bholmesdev (approved):
Definitely need an abstraction for handling Studio errors and some refinement on status codes. Won't block for this PR though. That's the entry-point of polish week!

FredKSchott:
100% agreed

PR #10391, opened 2024-03-11
fix: remove runtime/config suggestion for db config exports

bholmesdev:
## Changes

Before, your LSP would recommend importing `defineDb()` from `@astrojs/runtime/config` before recommending `astro:db`. This change removes `runtime/config` from public types to only surface `astro:db`

## Testing

Manual `pnpm link` to check intellisense

## Docs

N/A

### DIFF
--- a/.changeset/breezy-seals-begin.md
+++ b/.changeset/breezy-seals-begin.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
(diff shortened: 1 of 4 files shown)


bholmesdev:
!preview db-config-types

matthewp (approved):

delucis (approved):
LGTM!

PR #10400, opened 2024-03-12
Fix an issue where dev toolbar x-ray didn't escape props content

### LINKED ISSUE #10375: (Non-security) Toolbar renders HTML included in props, potentially causing unintended script execution in dev mode
### Astro Info

```block
> astro "info"

Package Manager          pnpm
Adapter                  none
                         @astrojs/mdx
                         @astrojs/svelte
```

_No response_

### Describe the Bug

If you insert an element with a prop containing HTML, the HTML will get rendered in the "inspect" overlay, potentially causing a harmless XSS.

This not an issue most of the time, but I am writing an article demonstrating an XSS (a simple `alert(1)`) and it is a bit annoying.

Notice in the example that I am not rendering the XSS-causing `img` tag in my code, but it is still getting executed. Stackblitz sandbox doesn't show it as well, but the element is getting loaded, erroring out, and alerting.

image

The inspect overlay does not render HTML if it is in the props

https://stackblitz.com/edit/github-1mq73i?file=src%2Fpages%2Findex.astro

### Participation

mingjunlu:
## Changes

- Escapes props content when building island tooltips
- Closes #10375

## Testing

- A test case is added
- Current tests should pass

## Docs

- N/A

### DIFF
--- a/.changeset/gentle-gorillas-sit.md
+++ b/.changeset/gentle-gorillas-sit.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue where dev toolbar x-ray didn't escape props content.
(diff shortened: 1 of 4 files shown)


Princesseuh (approved):
Looks great, thank you!

PR #10405, opened 2024-03-12
chore(db): Add missing github-slugger dependency & tests

43081j:
## Changes

This wasn't building locally thanks to `github-slugger` missing from the dependencies.

I've added it to the package, and a couple of tests while I was in there.

## Testing

New unit tests were added (though unrelated to github-slugger)

## Docs

N/A

### DIFF
--- a/.changeset/yellow-ducks-greet.md
+++ b/.changeset/yellow-ducks-greet.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Added github-slugger as a direct dependency of the db package
--- a/.changeset/yellow-ducks-greet.md
+++ b/.changeset/yellow-ducks-greet.md
@@ -2,4 +2,4 @@
 "@astrojs/db": patch
 ---
 
-Added github-slugger as a direct dependency of the db package
+Added github-slugger as a direct dependency
--- a/packages/db/test/unit/column-queries.test.js
+++ b/packages/db/test/unit/column-queries.test.js
@@ -1,5 +1,6 @@
 import { expect } from 'chai';
 import { describe, it } from 'mocha';
+import * as color from 'kleur/colors';
 import {
 	getCollectionChangeQueries,
 	getMigrationQueries,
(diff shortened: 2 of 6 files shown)


lilnasy (approved):
Looks great to me! Thanks for fixing it.

The changes seem obvious, but I'm surprised this wasn't an issue before. It may have to do with the specific package manager you use.

@43081j Could you share the error and the package manager it occurs with?

packages/db/test/unit/column-queries.test.js
```diff
@@ -105,6 +106,76 @@ describe('column queries', () => {
...
+			const { queries, confirmations } = await userChangeQueries(blogInitial, blogFinal);
+			expect(queries).to.deep.equal([
+				'DROP TABLE "Users"',
+				'CREATE TABLE "Users" (_id INTEGER PRIMARY KEY, "date" text NOT NULL)',
+			]);
+			expect(confirmations).to.deep.equal([
+				`You added new required column '${color.bold(
+					`${TABLE_NAME}.date`
+				)}' with no default value.` + '\n      This cannot be executed on an existing table.',
+			]);
```
lilnasy, line 154: We probably shouldn't test this. The messages may be updated at any time and updating the tests could be an inconvenience.

ematipico:
Yeah, this issue was completely unsound, but the package must be added. Thank you for fixing this

.changeset/yellow-ducks-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Added github-slugger as a direct dependency of the db package
```
ematipico, line 5:
```suggestion
Adds `github-slugger` as a direct dependency.
```

ematipico (approved):

packages/db/test/unit/column-queries.test.js
43081j: in eslint, we would have message IDs rather than asserting on the message itself

maybe we can do similar here, or just assert `confirmations.length` is `1`?

43081j:
i didn't do anything odd in particular, i basically did a clean clone then `pnpm i` at the root

i was using node v21, and pnpm 8.15.5 iirc

i'll try again locally later today if i can, to be sure it was those versions

43081j:
just tried it on my macbook (was on WSL on windows before i think), with a clean clone

```sh
pnpm i
pnpm run build
```

> @astrojs/db:build: src/core/cli/commands/link/index.ts:4:22 - error TS2307: Cannot find module 'github-slugger' or its corresponding type declarations.

node 21.7.1
pnpm 8.15.5

packages/db/test/unit/column-queries.test.js
lilnasy: Asserting the count sound perfect!

43081j:
have done the PR comments FYI 👍 

let me know if all is good

PR #10407, opened 2024-03-13
Add filter and conditional operators and aggregation helpers to exports

alexnguyennz:
## Changes

@astrojs/db - Add filter and conditional operators and aggregation helpers to exports [based on docs](https://docs.astro.build/en/guides/astro-db/#drizzle-utilities)

## Testing

Unable to test locally - get `Cannot find package 'shikiji'` error

## Docs

N/A


bholmesdev (commented):
Appreciate the catch here @alexnguyennz! Sorry your PR was missed, but we added a majority of these helpers in a separate PR. We also avoided any utilities that were marked as "not supported" for SQLite in the Drizzle documentation. This includes `ilike` and array operators. Hope this helps!

onnenon:
Any update on this? Just picked up AstroDB and really could use `ilike`

alexnguyennz:
Not sure - thought that they were added in the other PR.

@bholmesdev

onnenon:
Nope, no `ilike` on main.

see https://github.com/withastro/astro/blob/7b09c62b565cd7b50c35fb68d390729f936a43fb/packages/db/src/runtime/virtual.ts#L59-L90

bholmesdev:
Hey @onnenon! This is because we followed Drizzle's docs to decide which helpers are compatible with SQLite and which are not. According to their docs, `ilike` is incompatible: https://orm.drizzle.team/docs/operators#ilike

I'm surprised it would be incompatible while `notIlike` _is_ compatible though. I'll message their team to confirm.

onnenon:
@bholmesdev Hey thanks for the update! If there is anything I can do let me know. Would you like for me to open an issue to track this?

bholmesdev:
@onnenon Ah, I spoke with the Drizzle team and checked the SQLite docs to confirm. It seems `ilike` isn't supported in SQLite, so it isn't something we should expose from Astro DB either. `notIlike` support was a mistake in their documentation. 

That said, SQLite text search is case-insensitive by default. So there shouldn't be a need!

onnenon:
Awesome! Thank you so much for looking into that. I'm loving AstroDB so far.

PR #10414, opened 2024-03-13
Update VSCode Guidance for JSON Schemas

joe-bell:
## Changes

VSCode will search from the root, unlike the inline `$schema` definition

_Note: this feature is amazing, thank you!_

## Testing

Tested locally in my Astro project

## Docs

Documentation-only change


ematipico:
The fact that VSCode searches from the root, doesn't mean that other editors would do the same. I think the current usage is actually correct.

Princesseuh (approved):

ematipico:
Ooh I see, thank you!

PR #10415, opened 2024-03-13
Fix check example CI fail

bluwy:
## Changes

`main` is currently failing on the examples check fail. Seems like `@astrojs/check` after the lockfile upgrade is able to properly typecheck `.tsx` files now. This PR improves the typings so it passes.

- I also updated `@astrojs/check` to latest (0.5.8) from 0.3.1 (this updates our version only and not a breaking change)
- I tweaked the `check.yml` workflow so it'll run whenever we do deps upgrades again. 

## Testing

Ran `pnpm test:check-examples` locally.

## Docs

n/a

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,6 +12,7 @@ on:
       - ".github/workflows/check.yml"
       - "scripts/smoke/check.js"
       - "packages/astro/src/@types/astro.ts"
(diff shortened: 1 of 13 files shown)


Princesseuh (approved):
Tests failures are unrelated, this looks good to me!

ematipico (approved):

PR #10422, opened 2024-03-13
Use renovate

bluwy:
## Changes

This PR is opened as a discussion if we want to use renovate! If so, we can merge this PR, then someone with repo perms can setup the [renovate github app](https://github.com/marketplace/renovate) following [this instruction](https://docs.renovatebot.com/getting-started/installing-onboarding/#repository-installation). I'd suggest giving the github app access to this repo only as a start.

---

This PR removes the `nightly.yml` workflow in favor of [renovate](https://github.com/marketplace/renovate). The renovate github app is [free to use](https://github.com/renovatebot/renovate/discussions/23768#discussioncomment-6676357) for public/private repos. I'm currently using it for Vite and Svete orgs and I find it helpful. Some benefits:

1. Easier to review deps that have updated. It provides a table of updated deps and links to changelog. Example: https://www.github.com/vitejs/vite/pull/16131
2. It can help update major/breaking deps so we constantly keep them fresh. Example: https://www.github.com/vitejs/vite/pull/15234
3. If we don't want to update a dep, it can remember it (by closing the PR or via config). Example: https://www.github.com/vitejs/vite/pull/12880

Notes:

1. It will create one long-standing issue as a "dashboard" that we cannot close. Example: https://www.github.com/vitejs/vite/issues/4790
2. It's configured to send a PR weekly on Monday like our current workflow for non-major dependencies.
3. When a major dep upgrade is available, it'll send a PR to try and upgrade the dep (every Monday too iirc). There may likely be CI fails, but we can commit to the PR to fix it (renovate won't override changes).

## Testing

I copied the config from Vite and tweaked a bit.

## Docs

n/a


.github/renovate.json5
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:base", "schedule:weekly", "group:allNonMajor"],
+  "labels": ["dependencies"],
+  "rangeStrategy": "bump",
```
bluwy, line 5: This line will bump the dependency range in `package.json`. I think it's nice, but if we don't want to and prefer to update the lockfile only like before, it's also possible by removing this line.

Eveeifyeve: I would agree this would be best practices to me since the lock file is going to get updated anyways from someone building the library

Eveeifyeve:
Renovate would be a great thing but I would want to keep a tab and make sure it's on manual pr acceptance.

ematipico:
@bluwy can give us a brief summary of how renovate will work?

matthewp (approved):

ematipico (approved):

ematipico:
I suppose we need to coordinate the merging of this PR with @matthewp so he can grant the permissions to renovate bot to run in the repository

matthewp:
Yes, I'll add this to my board.

Eveeifyeve:
> @ematipico I've updated the description (notes) to add a bit more details. Let me know if I'm missing something. I think the linked PR examples are the best way to see how it works in practice.
> 
> @Eveeifyeve I've not used dependabot in the past because it didn't support grouped PRs (grouping multiple deps upgrade in one PR) until only recently. So I'm not really familiar with configuring it and can't quite vouch if there's any caveats in the long run. Happy to give it a shot though if you'd like to send a PR setting up `dependabot.yml` and we can compare them.

I guess I could look at doing a pr.

PR #10426, opened 2024-03-13
Fix issue with JSON schema generation when schema is a function

markgaze:
## Changes

When testing out the new experimental JSON schema generation, some of my schemas weren't generated because they were using functions as the `schema` (which is required to use the `image` function from the `SchemaContext`). I've changed the code to test whether the `schema` is a `Function` and then called it to get the actual schema (with a "stub" of `z.string()` for the `image` function).

## Testing

I've added 3 tests to `data-collections-schema-test.js` that failed for this use case before I made the changes.

I've also tested it in the repo that it was failing in and that fixed my issues.

## Docs

Nothing changes for the end user, so no new docs required


packages/astro/src/content/types-generator.ts
```diff
@@ -455,7 +455,10 @@ async function writeContentFiles({
 						settings.config.experimental.contentCollectionJsonSchema &&
 						collectionConfig?.schema
 					) {
-						let zodSchemaForJson = collectionConfig.schema;
+						let zodSchemaForJson =
+							collectionConfig.schema instanceof Function
```
bluwy, line 459: I think we usually check for functions this way:
```suggestion
							typeof collectionConfig.schema === 'function'
```

bluwy (commented):
I'm not familiar if a stub `image` API is fine, but otherwise only one nit about the changes. Great tests!

markgaze:
> I'm not familiar if a stub `image` API is fine, but otherwise only one nit about the changes. Great tests!

Thanks for the review! 🙂 I've changed the function check to your suggestion and all the tests still pass (as they should! 🙂).

I _think_ the stubbed `image` should be fine because, as far as the JSON is concerned, an image is just a string path. We could potentially make it use a regex so that the `pattern` property is populated in the schema but the types of `src` in `ImageMetadata` and `ImageFunction` are both strings anyway so I think it should be ok! If not, let me know and I'll change it 🙂

ematipico (approved):
Great effort in creating a good test! Thank you 🙏

Princesseuh (approved):
Looks good to me. The stub is fine, it does properly represent what that function is!

PR #10428, opened 2024-03-13
[db] Pass preferred region for token management

FredKSchott:
## Changes

- Currently, we create a short-lived Studio token through the Studio API.
- This is creating a race condition where the token is created in the API's region (`iad`) but not yet propagated to the region closest to the user (`sjc`, `ams`, etc). This presents as a "TOKEN_UNUAUTHORIZED" error because the region closest to the user assume's it's a bad token.
- This PR fixes the issue by capturing the region code closest to the user, and then passing that to the Studio API. The Studio API then makes sure that the token created/renewed/deleted  will be instantly up-to-date in that `region` before returning.

## Testing

- Manually 

## Docs

- N/A


natemoo-re (approved):
Can't speak to the Studio side of this code, but this all seems reasonable to me!

matthewp (approved):

matthewp (changes requested):
Needs API change, so blocking for the moment.

matthewp (approved):

PR #10431, opened 2024-03-13
[db] wait for token to propagate

FredKSchott:
## Changes

- A new attempt to solve this problem: https://github.com/withastro/astro/pull/10428
- That PR did solve the problem when each region had a single machine running, but failed to account for replication across multiple machines in the same region. 
- In practice, Machine A in region `sjc` could create your token and Machine B (also in region `sjc`) could attempt to use it before it's been propagated. Pinning to region `sjc` did not solve the problem in that case.
- The new solution removes the complexity of the original one, and now just embraces replication as something to wait for vs. trying to build around and solve for. It adds a delay on token creation greater than the maximum time our tokens take to propagate (2sec of wait vs. 1600ms of maximum token propagation). In the near future, we should be able to get this down closer to 1sec of wait time. 

https://github.com/withastro/astro/assets/622227/fa16bcad-3396-4617-8edf-009344202c5c

## Testing

- Tested manually

## Docs

- N/A

### DIFF
--- a/.changeset/breezy-dancers-give.md
+++ b/.changeset/breezy-dancers-give.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Add wait time for the db token to propagate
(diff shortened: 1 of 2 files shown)


matthewp:
Approved to get this fix out the door. Would love to discuss trying to figure out a deterministic solution for the future.

matthewp (approved):

Issue #10412, opened 2024-03-13
[Astro DB]: attempt to renew the same token indefinitely


FredKSchott:
### Astro Info

```block
(not my machine)
```

### Describe the Bug

https://github.com/withastro/astro/blob/2dd00a002482995469bf510d5b9fd0fa24bd023b/packages/db/src/core/tokens.ts#L85

This line will re-run the renew attempt indefinitely, failing indefinitely. 

### What's the expected result?

If `ManagedRemoteAppToken` is aware of your token's `ttl` and the time of every successful renewal, then it should be able to track when the token has become completely expired and is no longer valid. If this happens, it should likely fetch you a new token from scratch. If that fails, it should then throw an error and exit.

### Link to Minimal Reproducible Example

N/A

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #10435, opened 2024-03-14
Create new token when current has expired

matthewp:
## Changes

- Uses the `ttl` returned from the service to get an `expires` date.
- When renewing the token, the `expires` tells us if the token is completely expired and cannot be renewed, in which case we fetch a new token instead.
- Create a new renewal timer whether refreshing the current token or creating a new one.
- Fixes https://github.com/withastro/astro/issues/10412

## Testing

- Needs to be tested manually.

## Docs

- N/A, bug fix


matthewp:
!preview token-renewal

matthewp:
!preview token-renewal

packages/db/src/core/tokens.ts
```diff
@@ -95,24 +102,35 @@ class ManagedRemoteAppToken implements ManagedAppToken {
...
 				throw new Error(`Unexpected response: ${response.status} ${response.statusText}`);
-			}
-		} catch (error: any) {
-			const retryIn = (60 * this.ttl) / 10;
-			// eslint-disable-next-line no-console
-			console.error(`Failed to renew token. Retrying in ${retryIn} seconds.`, error?.message);
-			this.renewTimer = setTimeout(() => this.renew(), retryIn * 1000);
+			}	
+		} else {
+			const { token, ttl } = await ManagedRemoteAppToken.createToken(this.session, this.projectId);
```
FredKSchott, line 129: if token renewal has failed I think you can assume getting a new token would fail as well and just exit instead, with a helpful error message like "token has expired and attempts to renew it have failed, please try again"

That would prevent the class of errors where this handling logic fails and swallows the original error (I think in this case, the user would get a "could not create a new token" error if this fails when really the error was around renewal and not creation).

FredKSchott (approved):
LGTM! One comment as a recommendation about how to simplify the logic and improve the error case for the user

packages/db/src/core/tokens.ts
matthewp: done, thanks

matthewp:
!preview token-renewal

matthewp:
Hey @FredKSchott, I'm not sure I can properly test this. The previous code would only fail if:

1. Renewal continually failed past the `ttl` (probably because the API was failing?)
2. But the token create API was still working.

Does this sound right? If so I'm included just to test that this branch is working past the `ttl`.

PR #10442, opened 2024-03-14
fix(toolbar): Fixes small misc issues

Princesseuh:
## Changes

This fixes two small issues that got reported after 4.5 with the audit panel:

- If the lints refreshed while the panel was open, the panel wouldn't update correctly
- If a lint contained HTML tags, they would get wrongly stripped from the description in the panel. We now have a small Markdown support like we do for the error overlay, so we can show things nicely while still generally escaping HTML

image

## Testing

Updated mutation test to test for the audit UI as well

## Docs

N/A


packages/astro/src/runtime/client/dev-toolbar/apps/audit/index.ts
```diff
@@ -23,6 +23,8 @@ try {
 	customElements.define('astro-dev-toolbar-audit-list-item', DevToolbarAuditListItem);
 } catch (e) {}
 
+let state = false;
```
ematipico, line 26: nit: `showState`?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/ui/audit-ui.ts
```diff
@@ -147,18 +147,32 @@ function buildAuditCard(
...
 	return card;
 }
+
+/**
+ * Render a very small subset of Markdown to HTML or a CLI output
+ */
+function simpleRenderMarkdown(markdown: string) {
+	const linkRegex = /\[([^[]+)\]\((.*)\)/g;
+	const boldRegex = /\*\*(.+)\*\*/g;
+	const codeRegex = /`([^`]+)`/g;
```
ematipico, line 172: I would move these regexes outside of this function. We create them only once and not every time we run the function.

ematipico (approved):

PR #10445, opened 2024-03-14
Prefer getting app token from runtime env

matthewp:
## Changes

- Use the runtime `process.env.ASTRO_STUDIO_APP_TOKEN` before falling back to one that's in CI.

## Testing

- manually

## Docs

N/A, bug fix


lilnasy (approved):
Makes sense, `.env` files are often not included during docker/CI builds.

PR #10448, opened 2024-03-14
fix(rendering): allow framework renders to be cancelled

### LINKED ISSUE #10367: Server crashes due to multiple runtime errors in Astro component (Astro + React)
### Astro Info

```block
Package Manager          npm
```

All browsers

### Describe the Bug

When rendering an Astro component (`index.astro`), if there is an error in a React component preceded by another runtime error, it will result in a server crash.

Example:
```
---
// index.astro
import ReactComp from '../components/react-comp';

const foo = { bar: null } as unknown as { bar: { baz: string[] } };
---

<!-- React component has to error first -->
<ReactComp foo={foo} />
{ foo.bar.baz.length > 0 && (
  <div />
)}
```

### Steps to reproduce
1. Run `npm run build`
2. Run `npm run preview`
3. Go to localhost
4. Go to `/` page, server crashes.

Should not crash the server no matter how the error happen.

https://github.com/frendyguo/astro-react-runtime-error-crash

### Participation

lilnasy:
## Changes

- Fixes #10367 in a runtime-agnostic way, but only for ~70% of the similar cases.
- `renderTemplate()` did not complete - an error was thrown before we could get references to the expressions.
- Luckily, `renderComponent()` allowed us to get a reference the promise that would reject, and we could handle it.

## Testing
- Adapter the repro case into `node --test test/streaming.test.js`.

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -89,15 +89,14 @@ export class RenderContext {
 		const apiContext = this.createAPIContext(props);
 		const { type } = routeData;
 
-		const lastNext =
-			type === 'endpoint'
-				? () => renderEndpoint(componentInstance as any, apiContext, serverLike, logger)
-				: type === 'redirect'
-					? () => renderRedirect(this)
-					: type === 'page'
-						? async () => {
+		const lastNext = async () => {
+			if (routeData.type === 'endpoint') return renderEndpoint(componentInstance as any, apiContext, serverLike, logger);
+			if (routeData.type === 'redirect') return renderRedirect(this);
+			if (routeData.type === 'page') {
 								const result = await this.createResult(componentInstance!);
-								const response = await renderPage(
+								let response: Response;
+								try {
+								response = await renderPage(
 									result,
 									componentInstance?.default as any,
 									props,
@@ -105,6 +104,12 @@ export class RenderContext {
 									streaming,
 									routeData
 								);
+								} catch (e) {
+									// If the instantiation of the RenderTemplate fails midway,
+									// we signal to ignore the results of existing renders and avoid kicking off more of them.
+									result.abortController.abort();
+									throw e;
+								}
 								// Signal to the i18n middleware to maybe act on this response
 								response.headers.set(ROUTE_TYPE_HEADER, 'page');
 								// Signal to the error-page-rerouting infra to let this response pass through to avoid loops
@@ -112,11 +117,10 @@ export class RenderContext {
 									response.headers.set(REROUTE_DIRECTIVE_HEADER, 'no');
 								}
 								return response;
-							}
-						: type === 'fallback'
-							? () =>
+				}
+				if (routeData.type === 'fallback') return (
 									new Response(null, { status: 500, headers: { [ROUTE_TYPE_HEADER]: 'fallback' } })
-							: () => {
+				)
 									throw new Error('Unknown type of route: ' + type);
 								};
 
@@ -196,10 +200,13 @@ export class RenderContext {
 			},
 		} satisfies AstroGlobal['response'];
 
+		const abortController = new AbortController;
+
 		// Create the result object that will be passed into the renderPage function.
 		// This object starts here as an empty shell (not yet the result) but then
 		// calling the render() function will populate the object with scripts, styles, etc.
 		const result: SSRResult = {
+			abortController,
 			clientDirectives,
 			inlinedScripts,
 			componentMetadata,
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -120,7 +120,7 @@ export class RenderContext {
 					}
 					case 'fallback': {
 									return (
-										new Response(null, { status: 500, headers: { [ROUTE_TYPE_HEADER]: 'fallback' } })
+									new Response(null, { status: 500, headers: { [ROUTE_TYPE_HEADER]: 'fallback' } })
 									)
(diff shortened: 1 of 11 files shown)


packages/astro/src/runtime/server/render/component.ts
```diff
@@ -478,7 +478,11 @@ export async function renderComponent(
 		return renderAstroComponent(result, displayName, Component, props, slots);
 	}
 
-	return await renderFrameworkComponent(result, displayName, Component, props, slots);
+	return await renderFrameworkComponent(result, displayName, Component, props, slots)
+		.catch(e => {
+			if (result.cancelled) return { render() {} };
+			throw e;
+		});
```
lilnasy, line 485: The fix is here, the rest is clerical.

packages/astro/test/streaming.test.js
```diff
@@ -2,11 +2,9 @@ import assert from 'node:assert/strict';
 import { after, before, describe, it } from 'node:test';
 import * as cheerio from 'cheerio';
 import testAdapter from './test-adapter.js';
-import { isWindows, loadFixture, streamAsyncIterator } from './test-utils.js';
+import { loadFixture, streamAsyncIterator } from './test-utils.js';
 
 describe('Streaming', () => {
-	if (isWindows) return;
-
```
lilnasy, line 9: Theses tests do not need to be skipped.

lilnasy, line 9: Theses tests do not need to be skipped.

packages/astro/test/streaming.test.js
```diff
@@ -2,11 +2,9 @@ import assert from 'node:assert/strict';
 import { after, before, describe, it } from 'node:test';
 import * as cheerio from 'cheerio';
 import testAdapter from './test-adapter.js';
-import { isWindows, loadFixture, streamAsyncIterator } from './test-utils.js';
+import { loadFixture, streamAsyncIterator } from './test-utils.js';
 
 describe('Streaming', () => {
-	if (isWindows) return;
```
lilnasy, line 8: The test suite passes on windows. Was there another reason these were skipped?

packages/astro/src/runtime/server/render/component.ts
bluwy: Should we handle this catch for the entire code in `renderComponent` instead? Other components may error asynchronously too.

One of the caveats I figured with this technique is that we now have to ensure every async render APIs used in the template needs this special handling, but perhaps it's not too bad, we don't have many async APIs.

lilnasy: > Should we handle this catch for the entire code in renderComponent instead?

Yeah, that's a good idea.

> One of the caveats I figured with this technique is that we now have to ensure every async render APIs used in the template needs this special handling

That isn't sufficient either - there could be offshoot promises created by the user or a library. The blanket fix would be handling unhandled rejections similar to the dev mode. The problem there is listening for them doesn't seem to have a standard way across runtimes - it will be per adapter. I'll open a PR for node tomorrow.

bluwy: Yeah that's true. I think for those cases, we should recommend not passing in promises instead. I think it's rare for someone to do that, so I'm not too concerned if it happens.

I'm not too sure yet if adapters should be injecting the unhandledrejections by default though. It can happen anywhere in the code (e.g. third-party libraries), so it should be up to users to handle it themselves.

lilnasy: That's true, and we should prevent crashes in those situation - nothing more.

packages/astro/test/streaming.test.js
ematipico: No idea, to be honest. But if it passes, happy days!

packages/astro/src/core/render-context.ts
```diff
@@ -89,34 +89,38 @@ export class RenderContext {
...
 
-		const lastNext =
-			type === 'endpoint'
-				? () => renderEndpoint(componentInstance as any, apiContext, serverLike, logger)
-				: type === 'redirect'
-					? () => renderRedirect(this)
-					: type === 'page'
-						? async () => {
+		const lastNext = async () => {
+			if (routeData.type === 'endpoint') return renderEndpoint(componentInstance as any, apiContext, serverLike, logger);
```
ematipico, line 93: We should use a `switch (routeData.type)` here. If we add a new variant to `type`, TS and ESlint will be both errors and prevent the build.

ematipico (approved):

packages/astro/src/runtime/server/render/component.ts
lilnasy: > Should we handle this catch for the entire code in renderComponent instead?

Added cancellation handling for all async work in `renderComponent()`.

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -459,26 +459,35 @@ export async function renderComponent(
...
 	}
 
-	return await renderFrameworkComponent(result, displayName, Component, props, slots);
+	return await renderFrameworkComponent(result, displayName, Component, props, slots)
+		.catch(handleCancellation);
+
+	function handleCancellation(e: unknown) {
+		if (result.cancelled) return { render() {} };
+		throw e;
+	}
```
bluwy, line 490: I was thinking we can wrap this entire function body with a `try..catch`, but this also works for me.

bluwy (approved):

PR #10454, opened 2024-03-15
fix(node): handle offshoot promise rejections

### LINKED ISSUE #10367: Server crashes due to multiple runtime errors in Astro component (Astro + React)

lilnasy:
## Changes
- Fixes #10367 thoroughly but only for the node adapter.

## Testing
- Added

## Docs
- Does not affect usage.

### DIFF
--- a/packages/integrations/node/test/errors.test.js
+++ b/packages/integrations/node/test/errors.test.js
@@ -1,11 +1,16 @@
+import { spawn } from 'node:child_process';
+import { Worker } from 'node:worker_threads';
 import assert from 'node:assert/strict';
 import { after, before, describe, it } from 'node:test';
 import * as cheerio from 'cheerio';
 import nodejs from '../dist/index.js';
 import { loadFixture } from './test-utils.js';
+import { fileURLToPath } from 'node:url';
 
 describe('Errors', () => {
+	/** @type {import('./test-utils.js').Fixture} */
 	let fixture;
+
 	before(async () => {
 		fixture = await loadFixture({
 			root: './fixtures/errors/',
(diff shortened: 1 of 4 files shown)


packages/integrations/node/test/errors.test.js
```diff
@@ -17,10 +22,30 @@ describe('Errors', () => {
 	let devPreview;
 
 	before(async () => {
-		devPreview = await fixture.preview();
+		// devPreview = await fixture.preview();
```
ematipico, line 25: To remove

ematipico (approved):

packages/integrations/node/test/errors.test.js
lilnasy: It isnt being used yet

ematipico: What do you mean with yet? Can you be more precise, please?

lilnasy: The other two tests in the suite are skipped. The one I added starts a server within a worker. The preview server being started here is not used.

lilnasy: and it listens on the same port

ematipico: Usually, if you want to keep commented code around, it's customary to leave a comment explains that :)

lilnasy: Sure ill add that

lilnasy:
```suggestion
		// The two tests that need the server to run are skipped
		// devPreview = await fixture.preview();
```

bluwy (approved):
Still have a little concern of handling `unhandledRejection` for users by default, but I think it's worth trying it out.

lilnasy:
> Still have a little concern of handling unhandledRejection for users by default, but I think it's worth trying it out.

That's fair. I think it makes sense for the adapter to do it because there is no natural place for the developer to add the listener - api route top-level, middleware top-level?

PR #10455, opened 2024-03-15
qol(endpoints): helpful error message when a response is not provided

lilnasy:
## Changes
- From a bad experience in a support thread.
- Misuse of endpoints results in `cant read status of undefined` (we do a check on `response.status`)

## Testing
- Does not affect behavior.

## Docs
- The error will show up in error reference on docs.

### DIFF
--- a/packages/astro/src/runtime/server/endpoint.ts
+++ b/packages/astro/src/runtime/server/endpoint.ts
@@ -2,6 +2,8 @@ import { bold } from 'kleur/colors';
 import type { APIContext, EndpointHandler } from '../../@types/astro.js';
 import { REROUTABLE_STATUS_CODES, REROUTE_DIRECTIVE_HEADER } from '../../core/constants.js';
 import type { Logger } from '../../core/logger/core.js';
+import { AstroError } from '../../core/errors/errors.js';
+import { EndpointDidNotReturnAResponse } from '../../core/errors/errors-data.js';
 
 /** Renders an endpoint request to completion, returning the body. */
 export async function renderEndpoint(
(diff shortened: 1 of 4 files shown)


packages/astro/src/runtime/server/endpoint.ts
```diff
@@ -49,6 +51,11 @@ export async function renderEndpoint(
 	}
 
 	const response = await handler.call(mod, context);
+
+	if (!response || response instanceof Response === false) {
+		throw new AstroError(EndpointDidNotReturnAResponse)
+	}
```
ematipico, line 57: Isn't this handled by the `renderPage` function of the pipeline? If not, then a test should be cover this edge case

lilnasy: The `renderPage` function wouldn't run for endpoints. I'm guessing you meant `RenderContext.render()` - it checks the return of the middleware. Failures from the absence of a response occur before control flow returns to it.

An error here is not new, it was just thrown by accessing `response.status` before, now it thrown explicitly. A test would work the same before and after this PR.

lilnasy: A different error would be logged, but the tests will have access to a 500 response either way.

ematipico: Still, I believe it's worth expanding our test suite in this case. Any `throw` could return a 500, although this is a specific error that Astro decides to handle a meaningful way with a proper message.

Such cases should to be tested. We do this in our middleware, I don't see any reason why we shouldn't do the same here.

lilnasy: Okay, that makes sense. I had thought you meant a regression test.

sarah11918 (approved):
All the text looks great by docs! :raised_hands:

PR #10470, opened 2024-03-18
improves `client:only` error message

liruifengv:
## improves `client:only` error message

When set `client:only="react"` correctly, but not use `@astrojs/react` Integration, will get the error: https://docs.astro.build/en/reference/errors/no-client-only-hint/

- added clientOnlyValues set for check passedname
- throw different error by whether provide correct client:only directive
- Give a warning if provide incorrect client:only directive but find the renderer by guess


packages/astro/src/runtime/server/render/component.ts
```diff
@@ -80,7 +80,7 @@ async function renderFrameworkComponent(
 	_props: Record<string | number, any>,
 	slots: any = {}
 ): Promise<RenderInstance> {
-	if (!Component && !_props['client:only']) {
+	if (!Component && !_props['client:only'] && _props['client:only'] !== "") {
```
liruifengv, line 83: Add `_props['client:only'] !== ""` so that the following logic can be entered

packages/astro/src/runtime/server/render/component.ts
```diff
@@ -164,13 +164,12 @@ async function renderFrameworkComponent(
...
-				: passedName;
-			renderer = renderers.find(
-				({ name }) => name === `@astrojs/${rendererName}` || name === rendererName
-			);
+			const rendererName = metadata.hydrateArgs;
+			if (clientOnlyValues.has(rendererName)) {
+				renderer = renderers.find(
+					({ name }) => name === `@astrojs/${rendererName}` || name === rendererName
+				);
+			}
```
lilnasy, line 172: I haven't run this, but I think the changes here will prevent `client:only="solid"` from having an effect anymore.

We will try to find a renderer by the name of `@astrojs/solid` but the package name is `@astrojs/solid-js`.

Is it possible to reduce the changes here? I'm asking because it will also make reviewing easier since we could be rest assured that behavior doesn't change - only the messaging.

packages/astro/src/runtime/server/render/component.ts
lilnasy: If I understand correctly, we are making this condition more specific so that we do not throw an error here, and instead go on to guess the renderer.

We previously checked whether `props["client:only"]` was falsy (null, undefined, 0, ""), but really what we care about is whether or not its provided. Could `"client:only" in _props === false` be a better check?

liruifengv: Yeah. You are right! Thank you!

packages/astro/src/runtime/server/render/component.ts
liruifengv: Thanks. I'm keeping the same logic as before.

Issue #10158, opened 2024-02-19
Astro + Svelte 5 + MDX - Shows "undefined" instead of component


endigo9740:
### Astro Info

```block
Astro                    v4.4.0
Node                     v21.1.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/svelte
                         @astrojs/mdx
```

### If this issue only occurs in one browser, which browser is a problem?

All browsers

### Describe the Bug

I originally reported this issue on Discord, but was asked to file it here by @delucis:
https://discord.com/channels/830184174198718474/1197638002764152843/1208128538096701530

TL;DR: I am the core maintainer for Skeleton (https://skeleton.dev/). We're currently porting our component library forward to support Svelte 5 and React. As such, we're moving to Astro for our documentation website so we can run these multiple framework components in parallel.

## The Bug

Unfortunately we've run into a scenario where if we try to render even barebones Svelte 5 components in MDX pages, the component will show as `undefined`. However, we've discovered a partial workaround, which involves added an empty `$props` definition within the component's script tag, like so:

```svelte
<script lang="ts">
  let name = "john";

  // The Workaround
  const {} = $props<any>();
</script>

<p>Hello {name}, this is a Svelte test component.</p>

```

This enables support in a local dev server only. Once deployed (we use Vercel/SSR) we see `undefined` once again.

## Screenshots

Local dev server with props enabled:

![Screenshot 2024-02-19 at 12 26 46 PM](https://github.com/withastro/astro/assets/1509726/0166476d-f6a5-4cff-88fa-b5fe2af9ea41)

Otherwise we see this:

![Screenshot 2024-02-19 at 12 26 59 PM](https://github.com/withastro/astro/assets/1509726/ffa7a935-3615-44b7-8820-7da1b28a4500)

We see this odd error when hovering the component in the MDX file:

![Screenshot 2024-02-19 at 1 41 44 PM](https://github.com/withastro/astro/assets/1509726/b5b5e79b-2acc-454c-af37-96baedd37754)

### What's the expected result?

Svelte 5 support within MDX is critical for our project and MDX support is the gold standard for documentation. So support of the two in combination is make or break for us. We're happy to jump in an test and troubleshoot to help make this work.

---

A link to a minimal reproduction has been linked below. The Stackblitz link wasn't picking up the latest changes so beware (the root page must be `.mdx`)

### Link to Minimal Reproducible Example

[...]

Issue #4894, opened 2022-09-27
Server hangs until OOM if an MDX file imports something that throws an error


RobertAKARobin:
### What version of `astro` are you using?

1.3.1

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

npm

### What operating system are you using?

Mac

### Describe the Bug

If an `.mdx` file imports something that throws an error, the server just keeps spinning until it runs out of memory and the whole Astro process throws e.g:

`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

See example below.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/astro-mdx-errors

### Participation

- [ ] I am willing to submit a pull request for this issue.

RobertAKARobin:
Related to #4533?

PR #10473, opened 2024-03-18
Improve Astro JSX rendering

bluwy:
## Changes

Fixes https://github.com/withastro/astro/issues/10158

---

This fix requires some context, so here it is 😄:

When rendering JSX nodes, there's currently a problem where it can call `renderComponentToString` to render React/Vue/MDX component from JSX nodes, but it can also render JSX nodes again if it hits the `astro:jsx` renderer (for MDX). Creating a loop.

### **Before**
The solution to this problem is using "Skips" to make sure this order is ran (Worst case):

1. Try to render JSX nodes.
2. If fail, try to render components with React/Vue/MDX renderers.
3. Detected MDX, try to render its JSX nodes again.
4. Try to render components with React/Vue/MDX renderers.
5. Detected MDX, try to render its JSX nodes again.
6. Try to render JSX nodes again.
8. Finished. Returns rendered string or error.

(I don't know why no4 and no5 repeated. Perhaps a [typo in the logic](https://github.com/withastro/astro/blob/fdceed58894f34c81db46b104da662daf0807cb4/packages/astro/src/runtime/server/jsx.ts#L31)?)

### **After**
We use a single Symbol only to check if it had tried to render with React/Vue/MDX renderers. New order (Worst case):

1. Try to render components with React/Vue/MDX renderers.
2. Detected MDX, try to render its JSX nodes.
3. Finished. Returns rendered string or error.

This refactor allows us to simplify the skips logic, and remove the console filter as we no longer unconditionally render `vnode.type`. `renderComponentToString` will check for it first.

---

## Testing

Existing tests should pass. Also tested the repro manually from https://github.com/withastro/astro/issues/10158 (didn't add test because Svelte 5 is still experimental and hard to setup test). Also tested repro from https://github.com/withastro/astro/issues/4894.

## Docs

n/a. bug fix.

### DIFF
--- a/packages/astro/src/runtime/server/render/component.ts
+++ b/packages/astro/src/runtime/server/render/component.ts
@@ -515,7 +515,7 @@ export async function renderComponentToString(
 	// Handle head injection if required. Note that this needs to run early so
 	// we can ensure getting a value for `head`.
 	let head = '';
-	if (nonAstroPageNeedsHeadInjection(Component)) {
+	if (isPage && !result.partial && nonAstroPageNeedsHeadInjection(Component)) {
 		for (const headChunk of maybeRenderHead()) {
 			head += chunkToString(result, headChunk);
 		}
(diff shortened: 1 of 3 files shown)


packages/astro/src/runtime/server/render/component.ts
```diff
@@ -515,7 +515,7 @@ export async function renderComponentToString(
 	// Handle head injection if required. Note that this needs to run early so
 	// we can ensure getting a value for `head`.
 	let head = '';
-	if (nonAstroPageNeedsHeadInjection(Component)) {
+	if (isPage && !result.partial && nonAstroPageNeedsHeadInjection(Component)) {
```
bluwy, line 518: We now only try to render the head if this component is a page and is non-partial. Otherwise there's chance we could render a component but not use it (with how the new `renderJSX` flow works)

bluwy:
!bench render

ematipico:
> !bench render
> 
> EDIT: It doesn't look like there's much gain

Performance-wise, maybe not, but memory-wise, maybe yes. We don't construct a `Skip` class anymore, which is definitely some good nonetheless.

ematipico (approved):
Less code is the best code! 💪

Princesseuh (approved):
Looks fine to me!

PR #10474, opened 2024-03-18
Add indicator for inline code blocks

spcbfr:
## Changes

- currently there is no way to distinguish between inline and block-level code blocks.
- this results in inline code blocks getting styled in the same way as block-level ones.
- which may or may not be what you want.
- this change is non-breaking, the user is free to implement inline-code specific styles at their own pace

## Testing
no tests implemented. pretty simple change

## Docs
Developer can style inline code blocks on their own. just like what's done with other aspects of shiki. although if requested I can write documentation

### DIFF
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -84,6 +84,13 @@ const highlighter = await getCachedHighlighter({
 	],
 	theme,
 	themes,
+  transformers: [
+    {
+      code (node) {
+        this.addClassToHast(node, inline ? "inline-block" : "")
+      },
+    },
+  ]
 	wrap,
 });
 


bluwy:
If you use `inline`, it will be rendered as a `code` instead of `pre`. I think we can use this to differentiate?

https://github.com/withastro/astro/blob/c12666166db724915e42e37a048483c99f88e6d9/packages/markdown/remark/src/shiki.ts#L62-L65

spcbfr:
> If you use `inline`, it will be rendered as a `code` instead of `pre`. I think you can use this to differentiate?
> 
> https://github.com/withastro/astro/blob/c12666166db724915e42e37a048483c99f88e6d9/packages/markdown/remark/src/shiki.ts#L62-L65

@bluwy that is true, inline blocks do indeed use `<code>` tags, the gotcha is that non-inline blocks also use `<code>` they just wrap it inside a `<pre>`. if CSS had a way of styling an element based on it's parent I would have used that in this case but sadly CSS doesn't implement that as of yet

bluwy:
You can use `pre.astro-code`, `code.astro-code` or `code:not(.astro-code)` to target them specifically. The `astro-code` class is always added to the top-level element.

spcbfr:
@bluwy fair enough, I can handle that. but since I have you here, could we please get inline code blocks to respect the shiki theme the developer picked in the astro config? currently it defaults to github-dark unless you change it on a per-code-block basis.

bluwy:
Alright, I'll close this then. About respecting the shiki config, currently it's not implemented since it's not always possible to serialize the config to use in the Code component. They live in two "worlds": one in nodejs and one possibly in workerd/vercel-edge/etc in prod. So you have to manually specify the options to the Code component.

You can create a wrapper component for Code though that applies the defaults.

PR #10478, opened 2024-03-18
db: expose LibsqlError

bholmesdev:
## Changes

Expose `LibsqlError` to handle database errors.

Misc: remove unused `drizzle.ts` file

## Testing

Add foreign key constraint test. Tests that foreign keys work and whether `instanceof LibsqlError` resolves correctly.

## Docs

N/A


bholmesdev:
Working with the Turso team to understand the Libsql error codes available. Codes are set, but they are undocumented.

ematipico (commented):
I don't like exposing internal errors to users, because we expose something that is not handled by Astro.

Personally, I think we should use `AstroError` to wrap `LibsqlError`.

In fact, we risk exposing something that is undocumented, and I wouldn't want to do that (it would lower the quality of the docs).

Once the code errors are documented, we can expand this new error and expose the codes too.

bholmesdev:
Closing in favor of a different implementation

PR #10491, opened 2024-03-19
fix: do not append trailing slash to subresource urls

### LINKED ISSUE #10481: astro preview appends `/` to stylesheet URLs in some circumstances
### Astro Info

```block
Astro                    v4.5.6
Package Manager          npm
Integrations             none
```

_No response_

### Describe the Bug

Given the following circumstances:
- `output` is set to `hybrid`
- We use the `node` adapter
- `trailingSlash` is set to `always`

Then, when running our web app in preview mode with `astro preview`:
1. many static resources loaded by the page (such as .css or .js files) return a 301 status code
2. redirecting us to an URL with an appended `/` (that's OK for pages, but certainly not for .css or .js files)
3. which in turn returns us a 404

There shouldn't be any forced redirection to routes with an appended trailing slash for subresources such as .css or .js files.

https://stackblitz.com/edit/withastro-astro-qrbb9r?file=astro.config.mjs

### Participation

castarco:
## Changes

- What does this change? It fixes #10481 .
- How: By not appending a trailing slash to subresource urls

## Testing

- Added a new couple of tests to ensure that subresources such as .css files are properly served with a 200 status response.

## Docs

This change is a bugfix, no documentation changes are needed.

### DIFF
--- a/.changeset/tender-snails-accept.md
+++ b/.changeset/tender-snails-accept.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/node": patch
+---
+
+fix: do not append traling slash to subresource urls
--- a/.changeset/tender-snails-accept.md
+++ b/.changeset/tender-snails-accept.md
@@ -2,4 +2,4 @@
 "@astrojs/node": patch
 ---
 
-fix: do not append traling slash to subresource urls
+Fixes a bug where the preview server wrongly appends trailing slashes to subresource URLs.
--- a/packages/integrations/node/src/serve-static.ts
+++ b/packages/integrations/node/src/serve-static.ts
@@ -6,6 +6,8 @@ import type { NodeApp } from 'astro/app/node';
 import send from 'send';
 import type { Options } from './types.js';
 
+const isSubresourceRegex = /.+\.[a-z]+$/i
+
 /**
  * Creates a Node.js http listener for static files and prerendered pages.
  * In standalone mode, the static handler is queried first for the static files.
@@ -49,7 +51,7 @@ export function createStaticHandler(app: NodeApp, options: Options) {
 					break;
 				case 'always':
 					// trailing slash is not added to "subresources"
-					if (!hasSlash && !urlPath.match(/.+\.[a-z]+$/i)) {
+					if (!hasSlash && !urlPath.match(isSubresourceRegex)) {
 						pathname = urlPath + '/' + (urlQuery ? '?' + urlQuery : '');
 						res.statusCode = 301;
 						res.setHeader('Location', pathname);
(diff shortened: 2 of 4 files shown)


packages/integrations/node/test/trailing-slash.test.js
```diff
@@ -76,6 +76,14 @@ describe('Trailing slash', () => {
 				expect(res.status).to.equal(200);
 				expect($('h1').text()).to.equal('One');
 			});
+
+			it('Does not add trailing slash to subresource urls', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}/some-base/one.css`);
+				const css = await res.text();
+
+				expect(res.status).to.equal(200);
```
ematipico, line 84: I'm not sure what's gonig here, I think your local fork is outdated and it's using the old test file. We don't use `chai` anymore. Please update your fork, and use `assert` from `node:assert/strict`

ematipico (changes requested):

packages/integrations/node/test/trailing-slash.test.js
castarco: I think the test was never updated, it was left as .js, not .test.js , I just pushed a rebase trying to avoid relying on `chai`.

packages/integrations/node/src/serve-static.ts
```diff
@@ -48,7 +48,8 @@ export function createStaticHandler(app: NodeApp, options: Options) {
 					}
 					break;
 				case 'always':
-					if (!hasSlash) {
+					// trailing slash is not added to "subresources"
+					if (!hasSlash && !urlPath.match(/.+\.[a-z]+$/i)) {
```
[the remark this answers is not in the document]
castarco: I extracted the regex.

packages/integrations/node/test/trailing-slash.test.js
castarco: Replaced `expect` with `assert.equal` 👍🏻

.changeset/tender-snails-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/node": patch
+---
+
+fix: do not append traling slash to subresource urls
```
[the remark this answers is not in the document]
castarco: Applied suggestion

packages/integrations/node/src/serve-static.ts
```diff
@@ -6,6 +6,8 @@ import type { NodeApp } from 'astro/app/node';
 import send from 'send';
 import type { Options } from './types.js';
 
+const isSubresourceRegex = /.+\.[a-z]+$/i
```
lilnasy, line 9:
```suggestion
// check for a dot followed by a extension made up of lowercase characters
const isSubresourceRegex = /.+\.[a-z]+$/i
```

lilnasy (approved):
Looks great to me! Thanks for fixing the test as well.

ematipico (approved):
Thank you!

PR #10493, opened 2024-03-19
fix: void elements

firefoxic:
## Changes

- What does this change?
	HTML is not XML. It doesn't have self-closing tags, it has void element tags that don't need closing slashes.

	Generated void elements (e.g. `link` with path to style file) do not pass validation, which can be easily fixed by simply removing two characters.
- FYI: https://jakearchibald.com/2023/against-self-closing-tags-in-html/
	Screenshot from 2024-03-19 16-14-53
	Screenshot from 2024-03-19 16-17-11
	I'm not sure, but I hope I did it right.

## Testing

I tested this change in my astro-project — it fixed the validity of the markup.

## Docs

Correcting the documentation is most likely not necessary.

### DIFF
--- a/.changeset/thick-games-thank.md
+++ b/.changeset/thick-games-thank.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fix void elements
+`<link>` tags created by astro for optimized stylesheets now do not include the closing forward slash. This slash is optional for void elements such as link, but made some html validation fail.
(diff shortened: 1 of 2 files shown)


matthewp (approved):

Princesseuh (approved):

.changeset/thick-games-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix void elements
```
lilnasy, line 5:
```suggestion
`<link>` tags created by astro for optimized stylesheets now do not include the closing forward slash. This slash is optional for void elements such as link, but made some html validation fail.
```

PR #10496, opened 2024-03-19
feat: make 3rd party integration warning less scary

florian-lefebvre:
## Changes

- Removes "Use at your own risk!" when using `astro add` with a 3rd party integration. It does not remove any check

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/light-bikes-study.md
+++ b/.changeset/light-bikes-study.md
@@ -1,5 +1,5 @@
 ---
-"@fake-scope/fake-pkg": patch
+"astro": patch
 ---
 
 Makes the warning less scary when adding 3rd-party integrations using `astro add`
--- a/.changeset/light-bikes-study.md
+++ b/.changeset/light-bikes-study.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+Makes the warning less scary when adding 3rd-party integrations using `astro add`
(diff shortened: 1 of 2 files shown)


ematipico:
I believe that `Use at your own risk!` is a valid message, and it's not scary at all.

bholmesdev (approved):
Thanks for this! And to your comment @ematipico, I agree we should have a warning prompt before a user installs a third party integration. However, the addition of "Use at your own risk!" implies the user is doing something _wrong_ by installing a third party integration. They are not! We have a large integration community, and we should keep the warning to just "This is not an official Astro integration. Continue?"

.changeset/light-bikes-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
jdtjenkins, line 2: Is this scope right?

Adammatthiesen:
> We have a large integration community, and we should keep the warning to just "This is not an official Astro integration. Continue?"

100% agree, I personally kinda felt the "use at your own risk" meant your now "outside of the scope of the withAstro viewpoint" even if they were packages that could be common use among MANY of the community members (Expressive Code for example)

.changeset/light-bikes-study.md
florian-lefebvre: oops

ematipico (approved):

PR #10497, opened 2024-03-19
Removing embedding appToken from CI

matthewp:
## Changes

- Removes fallbacks that grab the app token from the CI env. 
- This is because we require this token be set in your host env instead.

## Testing

N/A

## Docs

N/A


bholmesdev (approved):
I'd do a sanity check that a Vercel deploy sets the secret correctly. If it works, take this approval as a `token` of my appreciation

PR #10501, opened 2024-03-19
Fix: remove `db.transaction()` from types until supported

bholmesdev:
## Changes

Users can call `db.transaction()` today, even though it is not supported by our remote DB adapter. This removes `db.transaction()` from our type definitions until this is supported, and raises a runtime error if called.

## Testing

Manual `pnpm pack` to ensure types work.

## Docs

N/A


Fryuni:
Sad, but better than failing weirdly at runtime!

ematipico (approved):

Issue #10451, opened 2024-03-15
Unable to create indexes on astro db


arch-fan:
### Astro Info

```block
Astro                    v4.5.4
Node                     v20.10.0
System                   Linux (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             astro:db
                         @astrojs/db/file-url
```

### Describe the Bug

Astro db push fails at creating the indexes. It doesn't matter the way I name the index, it throws the same error:

```
✔ Connected to remote database.
Database schema is out of date.
Pushing database schema updates...
https://db.services.astro.build/db/push failed: 400 Bad Request
{"success":false,"error":{"code":"SQL_QUERY_FAILED","details":"SQLITE_UNKNOWN: SQLite error: index id_idx already exists"}}
/db/push fetch failed: 400 Bad Request
```

### What's the expected result?

Indexes being created

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-deerzq?file=db%2Fconfig.ts

### Participation

- [ ] I am willing to submit a pull request for this issue.

arch-fan:
It's on the studio

matthewp:
@arch-fan Did this happen the first time you pushed or had you previously pushed and made some schema changes and then pushed again?

arch-fan:
It was the first time I pushed to studio

matthewp:
Can you try to push again and let me know if you still get this error?

arch-fan:
Still getting the same error:

```
✔ Connected to remote database.
Database schema is out of date.
Pushing database schema updates...
https://db.services.astro.build/db/push failed: 400 Bad Request
{"success":false,"error":{"code":"SQL_QUERY_FAILED","details":"SQLITE_UNKNOWN: SQLite error: index id_idx already exists"}}
/db/push fetch failed: 400 Bad Request
```

Sure, these are the versions:

```
dependencies:
@astrojs/db 0.8.5
astro 4.5.4
```

arch-fan:
Here you have the whole definition of the tables:

```ts
import { defineDb, defineTable, column, NOW } from "astro:db";

const Categories = defineTable({
  columns: {
    id: column.number({ primaryKey: true }),
    name: column.text({ unique: true }),
    slug: column.text({ unique: true }),
  },
  indexes: {
    slug_idx: {
      on: "slug",
      unique: true,
    },
    id_idx: {
      on: "id",
      unique: true,
    },
  },
});

const Authors = defineTable({
[...]

PR #10506, opened 2024-03-20
Ensure --force-reset drops previous tables

matthewp:
## Changes

- Previously `--force-reset` did not include DROP TABLE statements
- Now it does, and they are prepended before creation statements.
- Fixes https://github.com/withastro/astro/issues/10451

## Testing

- Added new unit tests for this.

## Docs

N/A, bug fix

### DIFF
--- a/packages/db/src/core/cli/migration-queries.ts
+++ b/packages/db/src/core/cli/migration-queries.ts
@@ -40,12 +40,23 @@ const genTempTableName = customAlphabet('abcdefghijklmnopqrstuvwxyz', 10);
 export async function getMigrationQueries({
 	oldSnapshot,
 	newSnapshot,
+	reset = false
 }: {
 	oldSnapshot: DBSnapshot;
 	newSnapshot: DBSnapshot;
+	reset?: boolean;
 }): Promise<{ queries: string[]; confirmations: string[] }> {
-	const queries: string[] = [];
+	let queries: string[] = [];
 	const confirmations: string[] = [];
+
+	// When doing a reset, first create DROP TABLE statements, then treat everything
+	// else as creation.
+	if(reset) {
+		const curSnapshot = oldSnapshot;
+		oldSnapshot = createEmptySnapshot();
+		queries = getDropTableQueriesForSnapshot(curSnapshot);
+	}
+
 	const addedCollections = getAddedCollections(oldSnapshot, newSnapshot);
 	const droppedTables = getDroppedCollections(oldSnapshot, newSnapshot);
 	const notDeprecatedDroppedTables = Object.fromEntries(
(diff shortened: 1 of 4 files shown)


packages/db/src/core/cli/migration-queries.ts
```diff
@@ -40,12 +40,23 @@ const genTempTableName = customAlphabet('abcdefghijklmnopqrstuvwxyz', 10);
...
+	reset?: boolean;
 }): Promise<{ queries: string[]; confirmations: string[] }> {
-	const queries: string[] = [];
+	let queries: string[] = [];
 	const confirmations: string[] = [];
+
+	// When doing a reset, first create DROP TABLE statements, then treat everything
+	// else as creation.
+	if(reset) {
+		const curSnapshot = oldSnapshot;
```
FredKSchott, line 55: nit `current` > `cur`

packages/db/src/core/cli/migration-queries.ts
```diff
@@ -40,12 +40,23 @@ const genTempTableName = customAlphabet('abcdefghijklmnopqrstuvwxyz', 10);
...
 	oldSnapshot,
 	newSnapshot,
+	reset = false
 }: {
 	oldSnapshot: DBSnapshot;
 	newSnapshot: DBSnapshot;
+	reset?: boolean;
 }): Promise<{ queries: string[]; confirmations: string[] }> {
-	const queries: string[] = [];
+	let queries: string[] = [];
```
FredKSchott, line 49: nit: can we keep this as a const, and instead of overwriting the array do `queries.push(...getDropTableQueriesForSnapshot(curSnapshot))`? My fear is that future devs blow away existing queries accidentally by re-assigning

packages/db/test/unit/reset-queries.test.js
```diff
@@ -0,0 +1,56 @@
...
+				reset: true,
+			});
+
+			expect(queries).to.deep.equal([
+				`DROP TABLE "${TABLE_NAME}"`,
+				`CREATE TABLE "${TABLE_NAME}" (_id INTEGER PRIMARY KEY, "name" text NOT NULL, "age" integer NOT NULL, "email" text NOT NULL UNIQUE, "mi" text)`,
+			]);
+		});
+
+		it('should not drop table when previous snapshot did not have it', async () => {
```
FredKSchott, line 42: good test 👍

packages/db/src/core/cli/migration-queries.ts
FredKSchott: actually, this name is a bit confusing, since it's not really the current snapshot right? Maybe like `originalOldSnapshot`?

FredKSchott (approved):
overall strong +1 !!! Left a few nit comments but resolve them at will

Issue #10492, opened 2024-03-19
Allow dynamic route segments in @astrojs/vercel isr.exclude array


tk04:
### Astro Info

```block
Astro                    v4.5.6
Node                     v20.11.0
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/vercel/serverless
Integrations             none
```

### Describe the Bug

When using the @astrojs/vercel integration, I want to exclude a dynamic route from using ISR. Currently, there seems to be no way to do this. All the routes that are passed into the `isr.exclude` array are matched exactly as is.

https://github.com/withastro/astro/blob/ad57a02c330b544770ab853fe0521eb784421016/packages/integrations/vercel/src/serverless/adapter.ts#L354-L357

### What's the expected result?

The processing of the `isr.exclude` array should offer a way to match dynamic route segments. Therefore, when the code is built, the route `src` values in the`.vercel/output/config.json` file should reflect dynamic route segments accordantly.

The strings are currently processed like this:
https://github.com/withastro/astro/blob/ad57a02c330b544770ab853fe0521eb784421016/packages/integrations/vercel/src/lib/redirects.ts#L79-L81

It would be nice to enable dynamic values using `/:path` or `/[path]`. A solution might look something like this:
```typescript
function escapeRegex(content) {
  return `^${getMatchPattern(
    content.split("/").map((value) => [
      {
        content: value,
        dynamic: value.startsWith(":"),
        spread: false,
      },
    ]),
  )}$`;
}


```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-alhqnq?file=astro.config.mjs

### Participation

- [X] I am willing to submit a pull request for this issue.

ematipico:
This sounds more like a feature request, not a bug report. 

As far as I remember, ISR accepts regex too.

tk04:
The route `src` values in the `config.json` file allow regex, but the `exclude.isr` array doesn't, the values are always escaped:

https://github.com/withastro/astro/blob/ad57a02c330b544770ab853fe0521eb784421016/packages/integrations/vercel/src/serverless/adapter.ts#L354-L357

lilnasy:
The exclude field currently only accepts verbatim paths, but I think it makes sense for it to handle dynamic and spread segments like `/[path]`.

It would work similarly to how configured redirects are processed. We have utilities somewhere that turn `/api/[path]` into the regular expression `/api/([^/]+?)`.

[...]

PR #10513, opened 2024-03-21
feat: allow dynamic route segments in isr.exclude array

tk04:
## Changes
Adds support for dynamic route segments in the isr.exclude array. See #10492

## Testing

Manually tested: the implementation transforms the route segments similar to how redirects are transformed.
## Docs

N/A

### DIFF
--- a/packages/integrations/vercel/src/lib/redirects.ts
+++ b/packages/integrations/vercel/src/lib/redirects.ts
@@ -15,6 +15,8 @@ interface VercelRoute {
 }
 
 // Copied from astro/packages/astro/src/core/routing/manifest/create.ts
+// Disable eslint as we're not sure how to improve this regex yet
+// eslint-disable-next-line regexp/no-super-linear-backtracking
 const ROUTE_DYNAMIC_SPLIT = /\[(.+?\(.+?\)|.+?)\]/;
 const ROUTE_SPREAD = /^\.{3}.+$/;
 function getParts(part: string, file: string) {
--- a/packages/integrations/vercel/src/lib/redirects.ts
+++ b/packages/integrations/vercel/src/lib/redirects.ts
@@ -1,5 +1,5 @@
 import nodePath from 'node:path';
-import { appendForwardSlash } from '@astrojs/internal-helpers/path';
+import { appendForwardSlash, removeLeadingForwardSlash } from '@astrojs/internal-helpers/path';
 import type { AstroConfig, RouteData, RoutePart } from 'astro';
 
 const pathJoin = nodePath.posix.join;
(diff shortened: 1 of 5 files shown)


packages/integrations/vercel/test/isr.test.js
```diff
@@ -33,10 +33,18 @@ describe('ISR', () => {
 				src: '^/two$',
 				dest: '_render',
 			},
+			{
+				src: '^\\/excluded\\/([^/]+?)$',
```
lilnasy, line 37: I noticed that the `src` fields created did not escape `/`, unlike core. The escaping may not be necessary, but it's worth confirming. Do you know if vercel handles this correctly unescaped?

lilnasy, line 37: I noticed that the `src` fields created did not escape `/`, unlike core. The escaping may not be necessary, but it's worth confirming. Do you know if vercel handles this correctly unescaped?

lilnasy (commented):
This looks great. Incredible work putting this together!

I was hoping the code for creating segments in core would be more modular, but oh well.

lilnasy (approved):
Looks great to me! Thanks for the clean implementation (especially all things considered 😅) .

packages/integrations/vercel/src/lib/redirects.ts
```diff
@@ -14,6 +14,33 @@ interface VercelRoute {
 	continue?: boolean;
 }
 
+// Copied from astro/packages/astro/src/core/routing/manifest/create.ts
+// Disable eslint as we're not sure how to improve this regex yet
+// eslint-disable-next-line regexp/no-super-linear-backtracking
+const ROUTE_DYNAMIC_SPLIT = /\[(.+?\(.+?\)|.+?)\]/;
+const ROUTE_SPREAD = /^\.{3}.+$/;
+function getParts(part: string, file: string) {
```
ematipico, line 22: I've seen this very function used in the Cloudflare adapter too. Maybe we should consider exposing it

tk04: That's right, it would make sense to expose it in the core package. Maybe also the code for parsing the segments: 
```typescript
const segments = removeLeadingForwardSlash(content)
  .split(nodePath.posix.sep)
  .filter(Boolean)
  .map((s: string) => {
	  return getParts(s, content);
  });
``` 
I've seen it used many times in the core package and the Cloudflare adapter.

matthewp:
If we're adding a new feature that we're going to have to support going forward I want it to be tested.

ematipico (changes requested):
Blocking the PR because it requires to update the documentation: https://docs.astro.build/en/guides/integrations-guide/vercel/#isr

ematipico (approved):

matthewp (approved):

PR #10529, opened 2024-03-22
Avoid generators when crawling module graph

bluwy:
## Changes

Avoid generators when crawling the module graph. Instead calculate the entire array directly and return it. In most cases, we're using the entire array so benefits of generators like _able to bail out of iterating_ wasn't used.

## Testing

existing tests should pass

In the astro docs repo, the Rollup build time went from 3m to 2m40s.

## Docs

n/a. internal refactor.


matthewp:
Needs a changeset. And this upsets me a lot.

matthewp (approved):

matthewp:
!bench memory

PR #10532, opened 2024-03-22
Fixes a style issue of `client:only` components in DEV mode during view transitions.

### LINKED ISSUE #10466: CSS imported into (solid) islands is not loaded on subsequent page loads when using view transitions and client:only in dev
### Astro Info

```block
Astro                    v4.5.5
Package Manager          npm
                         @astrojs/solid-js
                         @astrojs/mdx
                         @sentry/astro
```

I have seen this happen in both Chrome and Firefox.

### Describe the Bug

When importing CSS into a solid island (may be other islands too, I haven't tested other frameworks), the CSS loads fine on the initial page load, but when you navigate away and come back (with view transitions) the CSS is not loaded properly and the page looks broken.

This only appears to affect `client:only` (at least when using `client:visible` the CSS always loads correctly).

The included example sets builds on the official solid example. Click on the `Another page` link and then click the `Counter` link.

The fix is to move the CSS import from `Counter.tsx` to `index.astro`, then the CSS is always loaded properly under every circumstance, including with view transitions and `client:only`.

CSS imported in islands should load properly when navigating around the site using view transitions and `client:only`.

https://stackblitz.com/edit/github-eqw3jl?file=src%2Fpages%2Findex.astro

### Participation

martrapp:
## Changes

fixes #10466 

Originally fixed for client:only components that use transition:persist in #8840.
It turns out that the styles were also missing when navigating back to pages with non-persisted `client:only` components.

## Testing

e2e tests successful

## Docs

bug fix, n.a.

### DIFF
--- a/packages/astro/src/transitions/router.ts
+++ b/packages/astro/src/transitions/router.ts
@@ -696,18 +696,13 @@ async function prepareForClientOnlyComponents(newDocument: Document, toLocation:
 
 		const nextHead = nextPage.contentDocument?.head;
 		if (nextHead) {
-			// Clear former persist marks
-			document.head
-				.querySelectorAll(`style[${PERSIST_ATTR}=""]`)
-				.forEach((s) => s.removeAttribute(PERSIST_ATTR));
-
 			// Collect the vite ids of all styles present in the next head
 			const viteIds = [...nextHead.querySelectorAll(`style[${VITE_ID}]`)].map((style) =>
 				style.getAttribute(VITE_ID)
 			);
 			// Copy required styles to the new document if they are from hydration.
 			viteIds.forEach((id) => {
-				const style = document.head.querySelector(`style[${VITE_ID}="${id}"]`);
+				const style = nextHead.querySelector(`style[${VITE_ID}="${id}"]`);
 				if (style && !newDocument.head.querySelector(`style[${VITE_ID}="${id}"]`)) {
 					newDocument.head.appendChild(style.cloneNode(true));
 				}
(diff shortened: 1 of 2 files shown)


packages/astro/src/transitions/router.ts
```diff
@@ -696,18 +696,13 @@ async function prepareForClientOnlyComponents(newDocument: Document, toLocation:
...
-				.forEach((s) => s.removeAttribute(PERSIST_ATTR));
-
 			// Collect the vite ids of all styles present in the next head
 			const viteIds = [...nextHead.querySelectorAll(`style[${VITE_ID}]`)].map((style) =>
 				style.getAttribute(VITE_ID)
 			);
 			// Copy required styles to the new document if they are from hydration.
 			viteIds.forEach((id) => {
-				const style = document.head.querySelector(`style[${VITE_ID}="${id}"]`);
+				const style = nextHead.querySelector(`style[${VITE_ID}="${id}"]`);
```
martrapp, line 705: Instead of copying the styles from the last page, we no copy from the iframe. This also works for non-persisted `client:only` components.

packages/astro/src/transitions/router.ts
```diff
@@ -696,18 +696,13 @@ async function prepareForClientOnlyComponents(newDocument: Document, toLocation:
 
 		const nextHead = nextPage.contentDocument?.head;
 		if (nextHead) {
-			// Clear former persist marks
-			document.head
-				.querySelectorAll(`style[${PERSIST_ATTR}=""]`)
-				.forEach((s) => s.removeAttribute(PERSIST_ATTR));
-
```
martrapp, line 703: This was a remnant of an earlier approach.

matthewp (approved):

PR #10533, opened 2024-03-22
Require that ASTRO_STUDIO_REMOTE_DB_URL is defined at runtime

matthewp:
## Changes

- When we build we always get the app token from `process.env.ASTRO_STUDIO_APP_TOKEN` at runtime.
- `createRemoteDatabaseClient` now checks that it has an app token and throws without one.

## Testing

- SSR test added in the basics fixture. Using the mock services this tests that production works, can test other production things here.
- Added test specifically for when there is not a runtime ASTRO_STUDIO_APP_TOKEN, to check that it throws.

## Docs

N/A, bug fix

### DIFF
--- a/packages/db/src/core/cli/commands/execute/index.ts
+++ b/packages/db/src/core/cli/commands/execute/index.ts
@@ -44,6 +44,7 @@ export async function cmd({
 		virtualModContents = getStudioVirtualModContents({
 			tables: dbConfig.tables ?? {},
 			appToken: appToken.token,
+			isBuild: false,
 		});
 	} else {
 		virtualModContents = getLocalVirtualModContents({
--- a/.changeset/curly-donkeys-sleep.md
+++ b/.changeset/curly-donkeys-sleep.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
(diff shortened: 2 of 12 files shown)


packages/db/src/core/cli/commands/execute/index.ts
```diff
@@ -44,6 +44,7 @@ export async function cmd({
 		virtualModContents = getStudioVirtualModContents({
 			tables: dbConfig.tables ?? {},
 			appToken: appToken.token,
+			isBuild: false,
```
bholmesdev, line 47: Is there a reason for an `isBuild` parameter, when `execute` could set the app token in your environment to match how production builds work? That way, we could remove both the `appToken` and `isBuild` args from this function, and avoid maintaining two separate code paths.

bholmesdev (commented):
Not even `remote`ly worried about support issues now 👍 

Left a question on simplifying `db execute`.

packages/db/src/core/cli/commands/execute/index.ts
matthewp: That would be nice, but does `execute` run in `dev` and `build` and all of the other commands?

bholmesdev (approved):

PR #10538, opened 2024-03-23
Adds dataLanguage property to the replacement <pre> element.

604qgc:
## Changes
- Passes the language info used by the highlighter (`languageMatch?.[1] || "plaintext"`) as `dataLanguage` property to the replacement `<pre>` element.

## Testing
No testing other than verifying that the property is passed down properly; unsure what/how to test for such thing.

## Docs
When using Astro's built-in markdown processing, the built-in code highlighter processes the data first, before any other remark/rehype plugin can. The Shiki highlighter does not pass around the language information in its output.

This change fixes this and will allow users using the built-in Shiki highlighter to collect the coding language info after Shiki has passed.

Since it standardises the way the language info can be retrieved between Astro's two built-in highlighters, it will allow switching between the two while preserving this feature with no intervention (currently only Prism supports passing down the language, using `className`).

This should not break anything for users, unless they already use a `dataLanguage` property on `<pre>` elements as well as the built-in markdown processing.

The language info can then be retrieved in a rehype plugin like so:
```js
// myRehypePre.js
import { visit } from "unist-util-visit";
export default function myRehypePre() {
  return (tree) => {
    visit(tree, { tagName: "pre" }, (node) => {
      const lang = node.properties.dataLanguage;
      [...]
    });
  };
}
```


packages/markdown/remark/src/highlight.ts
```diff
@@ -60,6 +60,10 @@ export function highlightCodeBlocks(tree: Root, highlighter: Highlighter) {
 		const html = highlighter(code, languageMatch?.[1] || 'plaintext', { meta });
 		// The replacement returns a root node with 1 child, the `<pr>` element replacement.
 		const replacement = fromHtml(html, { fragment: true }).children[0] as Element;
+
+		// Adds dataLanguage property to replacement `<pre>`
+		replacement.properties.dataLanguage = languageMatch?.[1] || "plaintext"
```
bluwy, line 65: Can we extract the `languageMatch?.[1] || "plaintext"` as a variable so it can be shared on line 60 too?

bluwy (commented):
Sounds good to me. We also have a roadmap request for this: https://github.com/withastro/roadmap/discussions/276

bluwy (approved):

OliverSpeir (approved):

bluwy:
I'll put this in the 4.6 milestone since we're near it. It should be merged on this Wednesday.

.changeset/cold-snakes-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/markdown-remark": minor
+---
+
+Adds a `data-language` attribute on the rendered `pre` and `code` elements to expose the highlighted syntax language
```
matthewp, line 5: can you add a usage example?

matthewp (changes requested):
Before merging we would like a usage example in the changeset.

bluwy:
Good catch 👍 Now that I see it, maybe we can do the fix differently. We can remove the current changes, but add the `data-language` at:

https://github.com/withastro/astro/blob/799f6f3f29a3ef4f76347870a209ffa89651adfa/packages/markdown/remark/src/rehype-prism.ts#L11-L13

https://github.com/withastro/astro/blob/799f6f3f29a3ef4f76347870a209ffa89651adfa/packages/markdown/remark/src/shiki.ts#L103-L104

`Code.astro` currently uses the second linked code so the changes would auto-propagate there. That way I think we can continue use the same changeset.

.changeset/cold-snakes-train.md
sarah11918, line 5:
``````suggestion
Adds a `data-language` attribute on the rendered `pre` and `code` elements to expose the highlighted syntax language.

This allows retrieving the language in a rehype plugin:

```js
// myRehypePre.js
import { visit } from "unist-util-visit";
export default function myRehypePre() {
  return (tree) => {
    visit(tree, { tagName: "pre" }, (node) => {
      const lang = node.properties.dataLanguage;
      [...]
    });
  };
}
```
``````

Hi @604qgc ! I was thinking of including something like this in the changeset, since for minor releases of new features we do like to show them off a bit and really help people understand why your thing is cool! :nail_care: 

I took this from your PR description, but if there's something else you'd prefer to use, or would change this in some way, please do! I think a changeset of this style/length is helpful to people to understand the new feature, so anything like this you can suggest would be great.

File: packages/markdown/remark/src/rehype-prism.ts [hunk 3]
```diff
@@ -9,7 +9,7 @@ export const rehypePrism: Plugin<[], Root> = () => {
 			let { html, classLanguage } = runHighlighterWithAstro(language, code);
 
 			return Promise.resolve(
-				`<pre class="${classLanguage}"><code is:raw class="${classLanguage}">${html}</code></pre>`
+				`<pre data-language="${language}" class="${classLanguage}"><code is:raw class="${classLanguage}">${html}</code></pre>`
```
bluwy, line 12:
```suggestion
				`<pre class="${classLanguage}" data-language="${language}"><code is:raw class="${classLanguage}">${html}</code></pre>`
```

small nit

bluwy (approved):
Looks great!

.changeset/cold-snakes-train.md
```diff
@@ -0,0 +1,42 @@
...
+      [...]
+    });
+  };
+}
+```
+
+Note: if using the built-in `<Code>` component, the output of the component being flattened html, the replacement `<pre>` cannot be accessed using `{ tagName: "pre" }`.
+
+
+It also allows to use the `data-language` attribute in css rules:
```
ematipico, line 32:
```suggestion
It also allows the use of the `data-language` attribute in CSS rules:
```

.changeset/cold-snakes-train.md
```diff
@@ -0,0 +1,42 @@
...
+  return (tree) => {
+    visit(tree, { tagName: "pre" }, (node) => {
+      const lang = node.properties.dataLanguage;
+      [...]
+    });
+  };
+}
+```
+
+Note: if using the built-in `<Code>` component, the output of the component being flattened html, the replacement `<pre>` cannot be accessed using `{ tagName: "pre" }`.
```
ematipico, line 29:
```suggestion
Note: if using the built-in `<Code>` component, the output of the component being flattened HTML, the replacement `<pre>` cannot be accessed using `{ tagName: "pre" }`.
```

ematipico (approved):

.changeset/cold-snakes-train.md
```diff
@@ -0,0 +1,42 @@
+---
+"@astrojs/markdown-remark": minor
+---
+
+Adds a `data-language` attribute on the rendered `pre` elements to expose the highlighted syntax language.
+
+For example, the following markdown fenced-codeblock will expose `data-language="python"`:
```
sarah11918, line 7:
```suggestion
For example, the following Markdown code block will expose `data-language="python"`:
```

.changeset/cold-snakes-train.md
```diff
@@ -0,0 +1,42 @@
...
+```
+\```python
+def func():
+    print('Hello Astro!')
+\```
+```
+
+This allows retrieving the language in a rehype plugin:
```
sarah11918, line 15:
```suggestion
This allows retrieving the language in a rehype plugin from `node.properties.dataLanguage` by accessing the `<pre>` element using `{ tagName: "pre" }`:
```
Would this be the correct way to say this? I'd like to be able to call attention to the significant parts of the example.

.changeset/cold-snakes-train.md
sarah11918, line 29:
```suggestion
Note: The `<pre>` element is not exposed when using Astro's `<Code />` component which outputs flattened HTML.
```
I think we could keep this shorter here, especially since I called attention to `{ tagName: "pre" }` before the code snippet!

"outputs"? "produces"? "renders"?

.changeset/cold-snakes-train.md
sarah11918, line 32:
```suggestion
The `data-language` attribute may also be used in css rules:
```

sarah11918 (commented):
Thank you so much @604qgc! Admittedly, I think this feature *does* seem cooler know that you've been able to share more about it... and that's what we want! :smile:  The addition of the CSS mention is fantastic!

I just left a couple of wording suggestions for your consideration. Now that this is longer with examples, I tried to make the text itself more concise and direct. See what you think!

.changeset/cold-snakes-train.md
matthewp: Sounds right.

PR #10543, opened 2024-03-24
Fix inline stylesheets in content collection cache

matthewp:
## Changes

- Can't walk up the parent tree in propagation flag modules with CCC because `astro:content` is marked as external. So walk parents of that instead.

## Testing

- Unskipped most of the inline stylesheet tests, one remains.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/tricky-cobras-mate.md
+++ b/.changeset/tricky-cobras-mate.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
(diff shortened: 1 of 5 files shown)


matthewp:
!preview inline-ccc

Princesseuh (approved):

ematipico (approved):

PR #10547, opened 2024-03-25
feat: add option to parse `.md` files as mdx

duanwilliam:
## Motivation

Some people don't like having to change their file extensions to `.mdx` to make use of MDX.

For a sample use case, I personally mostly use mdx/astro as an IR of sorts, using directives for anything complex, and plugins to basically transform the directives into JSX. This way the content is technically valid plain markdown with directives as opposed to MDX, and in turn I find the `.md` extension more appropriate.

One user's feedback from the discord:
> I also am a .md purist so the concept sounds very cool

## Changes

- adds an `allowMd` boolean option to `@astrojs/mdx`, which if true makes `.md` files parsed as mdx
- updates `astro/vite-plugin-mdx` to run the transform on `.md` files as well. previously it would only run on `.mdx` files, but the associated comment with the check for this mentions both MD and MDX anyway

### Points of discussion

- open to feedback regarding what the option name should be, if another name is preferred
- currently only applies to `.md` files. `astro/src/core` has a `SUPPORTED_MARKDOWN_FILE_EXTENSIONS` and `isMarkdownFile()` but those are not exported and so aren't accessible to `@astrojs/mdx`. not sure how this should be best resolved
- I'm not too familiar with how the dataflow goes for markdown and `.mdx` files. I'm aware of `astro/vite-plugin-markdown` which appears to do some processing on markdown files already during the vite `load` step, but from what I can tell the data is the same going into the `transform` step between `.md` and `.mdx` files anyway? not too familiar with vite plugins either though so don't too much about it, but not sure if there's some behavior I'm overlooking due to whatever other processing plain markdown files typically go through that `.mdx` files don't.

## Testing

- added test fixtures to `@astrojs/mdx`, using substitution of MDX expressions as a proxy for parsing as MDX

## Docs

will probably submit docs pr once accepted and discussion is mostly settled

### DIFF
--- a/.changeset/hip-worms-marry.md
+++ b/.changeset/hip-worms-marry.md
@@ -1,0 +1,8 @@
+---
+"astro": minor
+"@astrojs/mdx": minor
+---
+
+Add `allowMd` boolean option, which makes `.md` files parsed as MDX as well.
+
+vite-plugin-mdx no longer early returns on `.md` files, as they may need to be parsed as JSX.
--- a/packages/astro/src/vite-plugin-mdx/index.ts
+++ b/packages/astro/src/vite-plugin-mdx/index.ts
@@ -102,7 +102,8 @@ export default function mdxVitePlugin({ settings }: AstroPluginJSXOptions): Plug
 			}
 			id = removeQueryString(id);
 			// Shortcut: only use Astro renderer for MD and MDX files
-			if (!id.endsWith('.mdx')) {
+			const markdownExtensions = ['.md', '.mdx'];
+			if (markdownExtensions.every((ext) => !id.endsWith(ext))) {
 				return null;
 			}
 			const { code: jsxCode } = await transformWithEsbuild(code, id, {
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -26,6 +26,7 @@ export type MdxOptions = Omit<typeof markdownConfigDefaults, 'remarkPlugins' | '
 	rehypePlugins: PluggableList;
 	remarkRehype: RemarkRehypeOptions;
 	optimize: boolean | OptimizeOptions;
+	allowMd: boolean;
 };
 
 type SetupHookParams = HookParameters<'astro:config:setup'> & {
@@ -36,6 +37,8 @@ type SetupHookParams = HookParameters<'astro:config:setup'> & {
 };
 
 export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroIntegration {
+	const extensions = partialMdxOptions.allowMd ? ['.md', '.mdx'] : ['.mdx'];
+
 	return {
 		name: '@astrojs/mdx',
 		hooks: {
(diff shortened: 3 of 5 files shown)


packages/astro/src/vite-plugin-mdx/index.ts
```diff
@@ -102,7 +102,8 @@ export default function mdxVitePlugin({ settings }: AstroPluginJSXOptions): Plug
 			}
 			id = removeQueryString(id);
 			// Shortcut: only use Astro renderer for MD and MDX files
-			if (!id.endsWith('.mdx')) {
+			const markdownExtensions = ['.md', '.mdx'];
+			if (markdownExtensions.every((ext) => !id.endsWith(ext))) {
```
duanwilliam, line 106: not sure if written as `not any` would be easier to understand than current `all not`, i.e.

```ts
if (!markdownExtensions.some((ext) => id.endsWith(ext)))
```

the former more closely resembles `vite-plugin-markdown` (`if (!isMarkdownFile(...))`), the latter more closely resembles the original code in this file.

packages/integrations/mdx/src/index.ts
```diff
@@ -46,7 +49,7 @@ export default function mdx(partialMdxOptions: Partial<MdxOptions> = {}): AstroI
 				addRenderer(astroJSXRenderer);
 				addPageExtension('.mdx');
 				addContentEntryType({
-					extensions: ['.mdx'],
+					extensions: extensions,
```
duanwilliam, line 52: don't know whether codebase prefers to use shorthand properties, erred on the side of explicitness here and didn't seem to be elided from linter/formatter in any case

ematipico:
While I understand the frustration of changing the extension of the file, this change will bring more harm than positive things.

Unfortunately mdx **ISN'T** a super language of Markdown, which means there's a chance markdown syntax will throw an error when it's converted to a mdx file.

.changeset/hip-worms-marry.md
```diff
@@ -0,0 +1,8 @@
+---
+"astro": minor
```
duanwilliam, line 2: not sure what type of change the change to `vite-plugin-mdx` constitutes as

duanwilliam:
> While I understand the frustration of changing the extension of the file, this change will bring more harm than positive things.
> 
> Unfortunately mdx **ISN'T** a super language of Markdown, which means there's a chance markdown syntax will throw an error when it's converted to a mdx file.

I agree it's a potential footgun, but it having to be explicitly set ideally somewhat mitigates that (+ potential documentation warning). Although I do realize that won't stop some people.

While I do sort of understand the stance, and with not adding options from any odd request for fear of bloat, it is a bit disappointing. For these situations where it is a sense of "I have valid MDX content, or content that will be transformed into it, but I want it in a `.md` file", the only alternatives are either changing extensions (breaking the abstraction, MDX is an implementation detail) or vendoring/forking (extremely high maintenance, given the use of internal Astro information in the unified plugins). A stretch, but is there any chance the likes of `createMdxProcessor()` could be exported in any case then, to avoid having to vendor all the unified plugins?

PR #10549, opened 2024-03-25
Fix: Type error on `<Image />` component when strict TypeScript rules are used with library type checking.

admirsaheta:
## Changes

- What does this change?
Fixes the following typescript issue, with it now being explicitly defined to avoid type mismatch on the runtime interpreter
```
Element implicitly has an 'any' type because expression of type '"data-image-component"' can't be used to index type 'HTMLAttributes<"img">'.
  Property 'data-image-component' does not exist on type 'HTMLAttributes<"img">'.
```
image

image

### DIFF
--- a/.changeset/blue-dolls-melt.md
+++ b/.changeset/blue-dolls-melt.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes `<Image />` type error when strict TypeScript rules are used with library type checking
+Updated the `HTMLAttributes` type exported from `astro` to allow data attributes
--- a/.changeset/blue-dolls-melt.md
+++ b/.changeset/blue-dolls-melt.md
@@ -1,5 +1,5 @@
 ---
-"astro": major
+"astro": patch
 ---
 
-hotfix:explicit-typesafe-fix
+Fixes <Image /> type error when strict TypeScript rules are used with library type checking.
--- a/.changeset/blue-dolls-melt.md
+++ b/.changeset/blue-dolls-melt.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Updated the `HTMLAttributes` type exported from `astro` to allow data attributes
+Updates the `HTMLAttributes` type exported from `astro` to allow data attributes
--- a/.changeset/blue-dolls-melt.md
+++ b/.changeset/blue-dolls-melt.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes <Image /> type error when strict TypeScript rules are used with library type checking.
+Fixes `<Image />` type error when strict TypeScript rules are used with library type checking
--- a/.changeset/blue-dolls-melt.md
+++ b/.changeset/blue-dolls-melt.md
@@ -1,0 +1,5 @@
+---
+"astro": major
+---
+
+hotfix:explicit-typesafe-fix
--- a/packages/astro/types.d.ts
+++ b/packages/astro/types.d.ts
@@ -4,11 +4,14 @@ import type { Simplify } from './dist/type-utils.js';
 
 /** Any supported HTML or SVG element name, as defined by the HTML specification */
 export type HTMLTag = keyof astroHTML.JSX.DefinedIntrinsicElements;
+
 /** The built-in attributes for any known HTML or SVG element name */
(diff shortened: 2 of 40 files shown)


.changeset/blue-dolls-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": major
+---
+
+hotfix:explicit-typesafe-fix
```
Fryuni, line 5: Fixes `<Image />` type error when strict TypeScript rules are used with library type checking.

Fryuni (changes requested):
Hey @admirsaheta, thanks for the PR! We are always glad to have new contributors.

There are just some things to improve on the changeset:
- Those messages get added to the [release notes](https://github.com/withastro/astro/releases/tag/astro%404.5.9) and to the [changelog](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#459), so we use clear descriptions of what was changed or fixed for people to understand how it might affect them.
- The changeset also indicates how the change relates to [SemVer](https://semver.org/), so a "major" there would mean a breaking change that would have to wait until Astro 5 to be included. That is not the case here. Since this PR doesn't change any Astro behavior and fixes a problem it can be a patch.

Fryuni (approved):

.changeset/blue-dolls-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes <Image /> type error when strict TypeScript rules are used with library type checking.
```
florian-lefebvre, line 5:
```suggestion
Fixes `<Image />` type error when strict TypeScript rules are used with library type checking
```

Princesseuh:
Hmm, I think there's a better fix here. `HTMLAttributes` not allowing data attributes is a problem

admirsaheta:
> Hmm, I think there's a better fix here. `HTMLAttributes` not allowing data attributes is a problem

You could extend the interface and provide custom attributes there e.g 

```ts
interface AdditionalAttributes extends HTMLAttributes<HTMLImageElement> {
    'data-image-component'?: string;
}
```
Let me know which approach fits better

florian-lefebvre:
We have been discussing on discord to update the `HTMLAttributes`. So at https://github.com/withastro/astro/blob/1cd2a740221ee14267f2889c4eb200bbcecb08aa/packages/astro/types.d.ts#L8 allow such type:
```ts
[key: `${string}-${string}`]: any
```
Do you want to handle it?

florian-lefebvre:
There's no rush don't worry!

admirsaheta:
Should handle it correctly here now 👍

packages/astro/types.d.ts
```diff
@@ -4,11 +4,14 @@ import type { Simplify } from './dist/type-utils.js';
...
 export type HTMLTag = keyof astroHTML.JSX.DefinedIntrinsicElements;
+
 /** The built-in attributes for any known HTML or SVG element name */
 export type HTMLAttributes<Tag extends HTMLTag> = Omit<
 	astroHTML.JSX.IntrinsicElements[Tag],
-	keyof Omit<AstroBuiltinAttributes, 'class:list'>
->;
+	keyof Omit<AstroBuiltinAttributes, 'class:list'> 
+> & {
+	[key: string]: string | number | boolean | undefined;
```
florian-lefebvre, line 13: @Princesseuh Do you prefer `any`?

.changeset/blue-dolls-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes `<Image />` type error when strict TypeScript rules are used with library type checking
```
florian-lefebvre, line 5:
```suggestion
Updated the `HTMLAttributes` type exported from `astro` to allow data attributes
```

florian-lefebvre (approved):
LGTM

packages/astro/types.d.ts
Princesseuh: Without necessarily using `any`, allowing also `null` would be great

admirsaheta: > Without necessarily using `any`, allowing also `null` would be great

Done

Princesseuh (approved):

PR #10550, opened 2024-03-25
fix: ensure outDir doesn't lead to bundlers bloating Lambda size

Skn0tt:
## Changes

Fixes https://github.com/withastro/adapters/issues/185, which is a bug around Lambda bundles containing unneeded `dist` files on Netlify. This bug presumably also exists on Vercel, as they use the same tracer under the hood (`@vercel/nft`).

## Testing

Added a test to ensure this isn't introduced again to the Vercel integration test suite. I picked that test suite, since we know they also have their own image integration, and that it likely has the same bug as the Netlify adapter.

## Docs

It's a bug fix, I don't see a need to change docs.

### DIFF
--- a/packages/astro/src/assets/vite-plugin-assets.ts
+++ b/packages/astro/src/assets/vite-plugin-assets.ts
@@ -70,7 +70,7 @@ export default function assets({
 					export const imageConfig = ${JSON.stringify(settings.config.image)};
 					// This is used by the @astrojs/node integration to locate images.
 					// It's unused on other platforms, but on some platforms like Netlify (and presumably also Vercel)
-					// "new URL("dist/...") is interpreted by the bundler as a signal to include that directory
+					// new URL("dist/...") is interpreted by the bundler as a signal to include that directory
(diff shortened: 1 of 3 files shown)


Skn0tt:
cc @Princesseuh for a review!

packages/astro/src/assets/vite-plugin-assets.ts
```diff
@@ -68,14 +68,20 @@ export default function assets({
 					export { default as Picture } from "astro/components/Picture.astro";
 
 					export const imageConfig = ${JSON.stringify(settings.config.image)};
-					export const outDir = new URL(${JSON.stringify(
+					// This is used by the @astrojs/node integration to locate images.
+					// It's unused on other platforms, but on some platforms like Netlify (and presumably also Vercel)
+					// "new URL("dist/...") is interpreted by the bundler as a signal to include that directory
```
ematipico, line 73:
```suggestion
					// new URL("dist/...") is interpreted by the bundler as a signal to include that directory
```

ematipico (approved):
Looks good!

PR #10561, opened 2024-03-26
fix(types): Fixes HTMLAttribute extend to allow everything

Princesseuh:
## Changes

The type in https://github.com/withastro/astro/pull/10549 didn't allow enough stuff and the key was too wide. This fixes that

## Testing

Tested manually + astro check should pass

## Docs

N/A

### DIFF
--- a/packages/astro/types.d.ts
+++ b/packages/astro/types.d.ts
@@ -10,7 +10,8 @@ export type HTMLAttributes<Tag extends HTMLTag> = Omit<
 	astroHTML.JSX.IntrinsicElements[Tag],
 	keyof Omit<AstroBuiltinAttributes, 'class:list'>
 > & {
-	[key: string]: string | number | boolean | null | undefined;
+	// This is not needed in the actual JSX definitions, because TypeScript implicitly adds it in JSX, however we need it here in case someone is using HTMLAttributes directly
+	[key: `${string}-${string}`]: any;
 };
 
 /**
--- a/packages/astro/types.d.ts
+++ b/packages/astro/types.d.ts
@@ -9,19 +9,17 @@ export type HTMLTag = keyof astroHTML.JSX.DefinedIntrinsicElements;
 export type HTMLAttributes<Tag extends HTMLTag> = Omit<
 	astroHTML.JSX.IntrinsicElements[Tag],
 	keyof Omit<AstroBuiltinAttributes, 'class:list'>
-> & {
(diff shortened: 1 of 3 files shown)


packages/astro/types.d.ts
```diff
@@ -10,7 +10,8 @@ export type HTMLAttributes<Tag extends HTMLTag> = Omit<
 	astroHTML.JSX.IntrinsicElements[Tag],
 	keyof Omit<AstroBuiltinAttributes, 'class:list'>
 > & {
-	[key: string]: string | number | boolean | null | undefined;
+	// This is not needed in the actual JSX definitions, because TypeScript implicitly adds it in JSX, however we need it here in case someone is using HTMLAttributes directly
+	[key: `${string}-${string}`]: any;
```
Fryuni, line 14: Isn't this just for `data-` attributes? Can't we limit to `data-${string}`?

Princesseuh: TypeScript actually has a funny behaviour here that we can't really reproduce in type, they end up allowing *every* invalid identifiers, from their docs

> Note: If an attribute name is not a valid JS identifier (like a data-* attribute), it is not considered to be an error if it is not found in the element attributes type.

It's quite wonky, but they somehow decided that would be how they'd support data attributes. I'd be okay with restricting it to `data-${string}` though

florian-lefebvre: Agreed

Princesseuh: I think I'll adjust everything to not use the implicit behaviour... let me rework things

packages/astro/types.d.ts
```diff
@@ -10,7 +10,8 @@ export type HTMLAttributes<Tag extends HTMLTag> = Omit<
 	astroHTML.JSX.IntrinsicElements[Tag],
 	keyof Omit<AstroBuiltinAttributes, 'class:list'>
 > & {
-	[key: string]: string | number | boolean | null | undefined;
+	// This is not needed in the actual JSX definitions, because TypeScript implicitly adds it in JSX, however we need it here in case someone is using HTMLAttributes directly
```
ematipico, line 13: I don't understand the comment in relation to the implementation. Why is it a string with a `-` in the middle? Reading at @Fryuni comment, it seems we're defining some data attributes?

Princesseuh:
> Does this PR require a changeset?

No, it's a fix to an unreleased thing.

Issue #10560, opened 2024-03-26
Type Errors in `examples/framework-react` on init in VS Code


apetta:
### Astro Info

```block
$ astro info
Astro                    v4.5.9
Node                     v18.18.2
System                   macOS (arm64)
Package Manager          yarn
Output                   static
Adapter                  none
Integrations             @astrojs/tailwind
                         @astrojs/react
```

### Describe the Bug

After setting up the example via pnpm/yarn `create astro@latest -- --template framework-react`, if you open a `*.tsx` file, VS Code type checking immediately begins complaining about type mismatches in .astro files in the node_modules folder from the astro package:
![Screenshot 2024-03-26 at 10 52 28](https://github.com/withastro/astro/assets/33671231/ba92f89b-2f4a-485e-ad8c-cf580adefbf8)

I have tried adding node_modules to the exclude field in tsconfig.json, but no luck.

It doesn't seem to hinder astro dev or build commands in any way, purely an annoyance in VS Code. 

### What's the expected result?

No errors.

### Link to Minimal Reproducible Example

https://github.com/withastro/astro/tree/main/examples/framework-react

### Participation

- [X] I am willing to submit a pull request for this issue.

apetta:
Uninstalling the Astro VS Code extension does make the errors go away...

apetta:
##10562 seems to make these errors go away whilst having the Astro VS Code extension installed.

I'm quite new to Astro, so I'm not entirely sure what's causing this or why it's even type checking the node_modules directory.

Just documenting my experience in case someone more knowledgeable on the topic can chime in. 

For now, I'm just applying the above changes as a patch via patch-package so I don't have to look at these errors 🙃

PR #10562, opened 2024-03-26
Patch for Astro VS Code Plugin type errors

apetta:
## Changes

Closes https://github.com/withastro/astro/issues/10560

Suggestive changes to typing that seems to silence the errors generated by the Astro VS Code plugin in a fresh project setup with the `/examples/framework-react` example as outlined in #10560 

## Testing

Tested in project, type annotation changes only

## Docs

N/A

### DIFF
--- a/packages/astro/components/Code.astro
+++ b/packages/astro/components/Code.astro
@@ -80,7 +80,7 @@ const highlighter = await getCachedHighlighter({
 			? Object.keys(bundledLanguages).includes(lang)
 				? lang
 				: 'plaintext'
-			: lang as any,
+			: (lang as any),
 	],
 	theme,
 	themes,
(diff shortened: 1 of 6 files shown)


Princesseuh:
Thank you for contributing! I think the cast here are actually unsafe, and this PR has some overlap with https://github.com/withastro/astro/pull/10561. I'll commit just a few small changes, and we'll get it merged.

packages/astro/components/Code.astro
```diff
@@ -89,7 +89,7 @@ const highlighter = await getCachedHighlighter({
 
 const html = highlighter.highlight(code, typeof lang === 'string' ? lang : lang.name, {
 	inline,
```
Princesseuh, line 91: For reviewers, those cast look suspicious, but there's actually already a bunch of said casts in the file. I don't really know why, perhaps @bluwy has more context having worked on Shiki recently?

Princesseuh: Okay, not sure what happened, GitHub moved my comment.. but I'm talking about the `as any` casts on `lang`.

bholmesdev (approved):
Should have had these fixes all a`lang`

PR #10568, opened 2024-03-26
Prevent importing db core from db runtime

matthewp:
## Changes

- Importing things in `src/core` from `src/runtime` will cause production to include rollup, vite and other unwanted things.
- This eslint rule will error if you do this. 

## Testing

- Verified locally

## Docs

N/A, bug fix

### DIFF
--- a/packages/db/src/core/errors.ts
+++ b/packages/db/src/core/errors.ts
@@ -37,44 +37,12 @@ export const SHELL_QUERY_MISSING_ERROR = `${red(
 	'▶ Please provide a query to execute using the --query flag.'
 )}\n`;
 
-export const SEED_ERROR = (error: string) => {
-	return `${red(`Error while seeding database:`)}\n\n${error}`;
-};
-
 export const EXEC_ERROR = (error: string) => {
 	return `${red(`Error while executing file:`)}\n\n${error}`;
 };
 
-export const SEED_DEFAULT_EXPORT_ERROR = (fileName: string) => {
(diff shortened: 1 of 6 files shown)


packages/db/src/core/errors.ts
```diff
@@ -37,46 +37,14 @@ export const SHELL_QUERY_MISSING_ERROR = `${red(
...
 
-export const SEED_DEFAULT_EXPORT_ERROR = (fileName: string) => {
-	return SEED_ERROR(`Missing default function export in ${bold(fileName)}`);
-};
-
 export const EXEC_DEFAULT_EXPORT_ERROR = (fileName: string) => {
 	return EXEC_ERROR(`Missing default function export in ${bold(fileName)}`);
 };
 
-export const REFERENCE_DNE_ERROR = (columnName: string) => {
```
bholmesdev, line 56: Nit: maybe we can move `errors.ts` to the base of the `src` directory instead of splitting into two modules?

bholmesdev (commented):
Excel-lint change! Left a comment on organizing our errors

packages/db/src/core/errors.ts
matthewp: We definitely can, it's interesting though that none of the errors are shared between. in which case, why have them together?

bholmesdev: Huh, honestly expected at least _one_ to be shared. Guess we can keep separate if that's the case!

bholmesdev (approved):

ematipico (approved):

PR #10579, opened 2024-03-27
Provide guidance when --remote is omitted

matthewp:
## Changes

- Expect a `DATABASE_FILE` during the build, which points to the production database, or error and explain that they probably wanted `--remote`.
- Note that this restriction does not apply to static mode.

## Testing

- New tests added, testing both cases.

## Docs

N/A, bug fix

### DIFF
--- a/packages/db/src/core/integration/index.ts
+++ b/packages/db/src/core/integration/index.ts
@@ -2,6 +2,7 @@ import { existsSync } from 'fs';
 import { dirname } from 'path';
 import { fileURLToPath } from 'url';
 import type { AstroConfig, AstroIntegration } from 'astro';
+import { AstroError } from 'astro/errors';
 import { mkdir, writeFile } from 'fs/promises';
 import { blue, yellow } from 'kleur/colors';
 import parseArgs from 'yargs-parser';
@@ -114,8 +115,10 @@ function astroDBIntegration(): AstroIntegration {
 				});
 			},
 			'astro:build:start': async ({ logger }) => {
-				if(!connectToStudio && !databaseFileEnvDefined() && output === 'server') {
-					throw new Error(`Attempting to build without the --remote flag or the DATABASE_FILE environment variable defined. You probably want to pass --remote to astro build.`)
+				if(!connectToStudio && !databaseFileEnvDefined() && (output === 'server' || output === 'hybrid')) {
+					const message = `Attempting to build without the --remote flag or the DATABASE_FILE environment variable defined. You probably want to pass --remote to astro build.`;
+					const hint = 'Learn more connecting to Studio: https://docs.astro.build/en/guides/astro-db/#connect-to-astro-studio';
+					throw new AstroError(message, hint);
 				}
 
 				logger.info('database: ' + (connectToStudio ? yellow('remote') : blue('local database.')));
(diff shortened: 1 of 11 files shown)


packages/db/src/core/integration/index.ts
```diff
@@ -111,6 +114,10 @@ function astroDBIntegration(): AstroIntegration {
 				});
 			},
 			'astro:build:start': async ({ logger }) => {
+				if(!connectToStudio && !databaseFileEnvDefined() && output === 'server') {
```
bholmesdev, line 117:
```suggestion
				if(!connectToStudio && !databaseFileEnvDefined() && (output === 'server' || output === 'hybrid')) {
```
Don't forget hybrid output!

packages/db/src/core/integration/index.ts
```diff
@@ -120,6 +127,11 @@ function astroDBIntegration(): AstroIntegration {
 	};
 }
 
+function databaseFileEnvDefined() {
+	const env = loadEnv('', process.cwd());
+	return env.DATABASE_FILE != null || process.env.DATABASE_FILE != null;
```
bholmesdev, line 132: nit: Any reason to use `DATABASE_FILE` instead of `ASTRO_DATABASE_FILE`?

I'm also pretty sure `loadEnv` covers the `process.env` check for us, but would need to test

bholmesdev (commented):
Some comments on error handling and naming. Re-mostly good to merge 👍

packages/db/src/core/integration/index.ts
matthewp: It doesn't cover it, the test I have needed the fallback.

No particular reason for the name, happy to change it to `ASTRO_DATABASE_FILE`.

packages/db/src/core/integration/index.ts
matthewp: I think we should error on hybrid, because the server routes might be using db, so we should err on the side of caution so they don't accidentally deploy something where only some routes fail.

bholmesdev: Wait, I thought the code _did not_ error on hybrid unless we make this change. Are you saying we **should** error if you use hybrid mode? Or saying it **shouldn't** error?

matthewp: Oh you're right, i'm saying it should, but it wasn't currently. I'll update.

packages/db/src/core/integration/index.ts
```diff
@@ -111,6 +114,10 @@ function astroDBIntegration(): AstroIntegration {
 				});
 			},
 			'astro:build:start': async ({ logger }) => {
+				if(!connectToStudio && !databaseFileEnvDefined() && output === 'server') {
+					throw new Error(`Attempting to build without the --remote flag or the DATABASE_FILE environment variable defined. You probably want to pass --remote to astro build.`)
```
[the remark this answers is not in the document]
matthewp: done

packages/db/src/core/integration/index.ts
matthewp: Updated

bholmesdev (approved):
TIL we expose Astro error

Issue #10581, opened 2024-03-27
Creating a new Response with a Status Code of 404 leads to an infinite loop when the /404.astro page is absent.


gitcaduff:
### Astro Info

```block
Astro                    v4.5.6
Node                     v20.5.1
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/svelte
                         @astrojs/tailwind
```

### Describe the Bug

In the past it was feasible to generate Responses bearing the 404 status code. However, in the current setup (version 4.5.x and above), this process triggers an infinite loop due to Astro's subsequent search for the missing /404.astro page. To circumvent this loop, I opt not to furnish a /404.astro page. This choice stems from my desire to refrain from redirecting users to a dedicated 404 page. Instead, I intend to furnish a 404 Code directly while retaining the requested URL's context.

The problem exists only in dev mode and was already reported and fixed here: https://github.com/withastro/astro/issues/8100
But with the new version 4.5.x the bug came back.

### What's the expected result?

I would expect an Response with the status code 404 without redirecting to the missing /404.astro 

`return new Response("custom 404 html content", {
   status: 404
});`

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/stackblitz-starters-jcjvxf?file=response-with-404-code%2Fsrc%2Fpages%2F[...path].astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for the clean reproduction!

PR #10582, opened 2024-03-27
fix(dev): consider pages' route priority when matching request to a page

lilnasy:
## Changes
- #10581
- Fixes infinite loop when an astro page with a dynamic or spread path returned a 404 response.
- The problem was that the default error page appeared lower in the `manifest.routes` array, leading to the dynamic normal page - which also matched the `/404` path - always being matched as the reroute destination.
- The solution is to sort the routes whose regex matched, using the normal sorting rules used in prod.

## Testing
- Added fixture.

## Docs
- Does not affect behavior

### DIFF
--- a/packages/astro/src/prerender/routing.ts
+++ b/packages/astro/src/prerender/routing.ts
@@ -1,5 +1,6 @@
 import type { AstroSettings, ComponentInstance, RouteData } from '../@types/astro.js';
 import { RedirectComponentInstance, routeIsRedirect } from '../core/redirects/index.js';
+import { routeComparator } from '../core/routing/priority.js';
 import type { DevPipeline } from '../vite-plugin-astro-server/pipeline.js';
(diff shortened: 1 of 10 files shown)


ematipico (approved):
Looks good! Thank you catching this

packages/astro/src/prerender/routing.ts
```diff
@@ -19,7 +20,9 @@ export async function getSortedPreloadedMatches({
 			matches,
 			settings,
 		})
-	).sort((a, b) => prioritizePrerenderedMatchesComparator(a.route, b.route));
+	)
+	.sort((a,b) => routeComparator(a.route, b.route))
+	.sort((a, b) => prioritizePrerenderedMatchesComparator(a.route, b.route))
```
lilnasy, line 25: The fix is here. The rest is clerical.

matthewp (approved):

PR #10584, opened 2024-03-27
fix: dont error on nullish prop values in jsx runtime

### LINKED ISSUE #10575: Astro MDX fails to render framework components and components defined with MDX file, when explicit `null` or `undefined` prop passed in
### Astro Info

(from stackblitz environment)

```block
Astro                    v4.5.10
Adapter                  none
```

_No response_

### Describe the Bug

If a framework component is used within an MDX file, and one of its props gets passed in an explicit `undefined` or `null`, Astro crashes with an error trying to render the component.
This issue also occurs for components defined within the MDX file.

e.g.

```mdx
import A from 'foo.jsx'

export function B() { return <div /> }

{/* both of these cause an error */}
<A foo={undefined} />
<B bar={undefined} />
```

The problematic line is https://github.com/withastro/astro/blob/bb7707e1715fc3f496e4c3f07834221c8da62ff6/packages/astro/src/runtime/server/jsx.ts#L118-L119

in these cases,
- `vnode` is an object, so `renderJSX` delegates to `renderJSXNode` (this seems to be true of all the jsx inside an mdx file?)
- `vnode['astro:jsx']` is `true` (i.e. it is in fact a vnode)
- `vnode.type` is the component, i.e. a function
- `vnode.type` does _not_ have `isAstroComponentFactory` set
- there is no truthy `server:root` prop

which leads to aforementioned line being called, resulting in attempting to do a property access on `null`/`undefined` which results in an error.

Component renders properly.

https://stackblitz.com/edit/github-mbbqqu?file=src%2Fpages%2Fa.mdx

### Participation

duanwilliam:
resolves #10575

## Changes

- jsx runtime no longer errors on non-`.astro` jsx function component instances represented as vnode objects with nullish prop values
    - only try property access on non-nullish prop values when checking if a prop is a slot

## Testing

tested locally (ran dev server for `@examples/with-mdx` and added reproduction sample from linked issue)

not sure where a test file would best go or how to construct it, since particular reproduction from linked issue uses mdx, but don't think mdx is actually necessary, merely an environment where the issue could naturally occur.

## Docs

not needed, internal bugfix

### DIFF
--- a/.changeset/forty-turkeys-tap.md
+++ b/.changeset/forty-turkeys-tap.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug where JSX runtime would error on components with nullish prop values in certain conditions.
--- a/packages/astro/src/runtime/server/jsx.ts
+++ b/packages/astro/src/runtime/server/jsx.ts
@@ -116,7 +116,7 @@ Did you forget to import the component or is it possible there is a typo?`);
 			}
 			extractSlots(children);
 			for (const [key, value] of Object.entries(props)) {
-				if (value['$$slot']) {
+				if (value?.['$$slot']) {
 					_slots[key] = value;
 					delete props[key];
 				}


lilnasy (approved):
Looks great to me! Excellent work finding the fix, I was pleasantly surprised by the investigation in the original report.

ematipico (approved):

PR #10585, opened 2024-03-27
refactor: Drop Preact compat hack, remove incorrect alias

rschristian:
## Changes

- Swaps out the config to disable `preact/compat` aliases now that `@preact/preset-vite` supports this natively (https://github.com/preactjs/preset-vite/pull/79)
- Disables incorrect `react/jsx-runtime` -> `preact/jsx-runtime` alias. `@preact/preset-vite` already covers this (correctly) and so it did not need a replacement here.

## Testing

No functional changes, beyond fixing a theoretical edge case with the `jsx-runtime` alias. I don't think it's worth adding a test here, this is pretty unlikely behavior to run across and is addressed by `@preact/preset-vite` already.

## Docs

N/A, not user facing.

## Misc

Some of the other config bits looks to be something we can upstream, so I'll take a look soon-ish and see if we can't reduce the config here further.

### DIFF
--- a/.changeset/proud-books-hope.md
+++ b/.changeset/proud-books-hope.md
@@ -2,4 +2,4 @@
 "@astrojs/preact": patch
 ---
 
-Fix (theoretical) edge case in Preact integration's JSX aliases
+Fixes (theoretical) edge case in Preact integration's JSX aliases
--- a/.changeset/proud-books-hope.md
+++ b/.changeset/proud-books-hope.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/preact": patch
+---
+
+Fix (theoretical) edge case in Preact integration's JSX aliases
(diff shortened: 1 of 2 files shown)


packages/integrations/preact/src/index.ts
```diff
@@ -20,6 +20,7 @@ export default function ({ include, exclude, compat }: Options = {}): AstroInteg
 		hooks: {
 			'astro:config:setup': ({ addRenderer, updateConfig, command }) => {
 				const preactPlugin = preact({
+					reactAliasesEnabled: compat,
```
lilnasy, line 23:
```suggestion
					reactAliasesEnabled: compat ?? false,
```
From the typedocs, the default for `reactAliasesEnabled` is `true`. It would be problematic for projects using both react and preact components, such as several of our test fixtures.

rschristian: Ah, did this from my phone, didn't notice `compat` isn't always set. Cheers

matthewp (approved):

.changeset/proud-books-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/preact": patch
+---
+
+Fix (theoretical) edge case in Preact integration's JSX aliases
```
bluwy, line 5:
```suggestion
Fixes (theoretical) edge case in Preact integration's JSX aliases
```

PR #10589, opened 2024-03-27
db: Rework index config with generated index names

bholmesdev:
## Changes

This introduces generated index names when passing an array to `indexes`. The object configuration syntax is now deprecated. [See changelog](https://github.com/withastro/astro/pull/10589/files#diff-64f66d18b0d560e74e3df56636df8a3dd2c564e9bc453080615f2208d5a50732)
- Update `dbConfig` with a transformation step to generate index names by table and column names
- Introduce a `ResolvedDbConfig` type for migration queries
- Misc: add a `mapObject` utility to remap object values with type safety

## Testing

Add tests for new array format, and move old tests to "legacy"

## Docs

https://github.com/withastro/docs/pull/7639

### DIFF
--- a/.changeset/blue-ghosts-rule.md
+++ b/.changeset/blue-ghosts-rule.md
@@ -1,0 +1,31 @@
+---
+"@astrojs/db": patch
+---
+
+Update the table indexes configuration to allow generated index names. The `indexes` object syntax is now deprecated in favor of an array.
+
+## Migration
+
+You can update your `indexes` configuration object to an array like so:
+
+```diff
+import { defineDb, defineTable, column } from 'astro:db';
+
+const Comment = defineTable({
+  columns: {
+    postId: column.number(),
+    author: column.text(),
+    body: column.text(),
+  },
+- indexes: {
+-   postIdIdx: { on: 'postId' },
+-   authorPostIdIdx: { on: ['author, postId'], unique: true },
+- },
++ indexes: [
++   { on: 'postId' /* 'name' is optional */ },
++   { on: ['author, postId'], unique: true },
++ ]
+})
+```
+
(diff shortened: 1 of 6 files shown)


bholmesdev:
Kept to a `patch` since the old config object is deprecated, but still works! Made a ticket to remove in the next minor.

.changeset/blue-ghosts-rule.md
```diff
@@ -0,0 +1,31 @@
+---
+"@astrojs/db": patch
```
matthewp, line 2: Just to check, this is non-breaking, the old syntax is still supported for now, right?

matthewp: What will happen to people's existing indices? Will they be dropped and recreated?

bholmesdev: Old syntax is supported! And it _shouldn't_ drop and recreate even when refactoring to the new syntax. We handle everything with a Zod transform that matches the original object format. I'll add a unit test to confirm

matthewp: I guess what i'm asking is, currently the user can name them whatever they want. For example `foobar_index`. But now with the automated index names, is it not going to try to add the new index and not know about the previous index?

I guess you're saying that when you do push, it will see the old index and the new index and drop the old one (assuming it doesn't match) and create a new one?

matthewp (approved):

.changeset/blue-ghosts-rule.md
bholmesdev: @matthewp Yep, it will drop and recreate the indexes if you _don't_ pass a `name` option. If you pass a `name` to the new indexes that matches the old name, it'll run without any migrations 👍

PR #10599, opened 2024-03-28
db: Seed on dev server startup

bholmesdev:
## Changes

This PR loads your seed file whenever the dev server startups. To make it unblocking and preserve HMR on file changes, we use `vite.ssrLoadModule()`.
- Add a `ssrLoadModule()` call for the `astro:db:seed` module on server startup
- Add a file watcher to `ssrLoadModule()` again whenever a seed file changes. This does **not** cause duplicate runs in the Vite server; it eagerly loads and allows us to track the promise to log another "seeded database" message. This gives user feedback when your seed file changes and keeps your db up-to-date even when your browser is closed.

## Testing

Manual testing to ensure logs fire and the seed file is only loaded/reloaded once.

## Docs

N/A

### DIFF
--- a/packages/db/src/core/integration/index.ts
+++ b/packages/db/src/core/integration/index.ts
@@ -13,8 +13,9 @@ import { type ManagedAppToken, getManagedAppTokenOrExit } from '../tokens.js';
 import { type VitePlugin, getDbDirectoryUrl } from '../utils.js';
 import { fileURLIntegration } from './file-url.js';
 import { typegenInternal } from './typegen.js';
-import { type LateSeedFiles, type LateTables, vitePluginDb } from './vite-plugin-db.js';
+import { type LateSeedFiles, type LateTables, vitePluginDb, resolved } from './vite-plugin-db.js';
 import { vitePluginInjectEnvTs } from './vite-plugin-inject-env-ts.js';
+import { SEED_DEV_FILE_NAME } from '../../runtime/queries.js';
 
 function astroDBIntegration(): AstroIntegration {
 	let connectToStudio = false;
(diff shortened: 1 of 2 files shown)


packages/db/src/core/integration/index.ts
```diff
@@ -113,6 +106,36 @@ function astroDBIntegration(): AstroIntegration {
 						server.restart();
 					}
 				});
+				// Wait for dev server log before showing "connected".
+				await new Promise((resolve) => setTimeout(resolve, 100));
+				logger.info(
+					connectToStudio ? 'Connected to remote database.' : 'New local database created.'
+				);
+				const seedFileUrls = SEED_DEV_FILE_NAME.map(
```
delucis, line 114: Does this also need to account for integration seed files? Right now it looks like it’s only our static array of project ones?

delucis (commented):
Excited to see this! Left a query about integrations.

Will save my best puns for an approving `seed`quel to this review.

packages/db/src/core/integration/index.ts
bholmesdev: Ah, that's a good point! This is only for _eagerly_ loading when a seed file changes, so it shouldn't break existing behavior. Happy to add if it's easy enough

bholmesdev: Update: this will work for relative paths but _not_ package ids. Since eagerly reloading is a nice-to-have, I think this is a fine compromise

packages/db/src/core/integration/index.ts
```diff
@@ -113,6 +106,51 @@ function astroDBIntegration(): AstroIntegration {
...
+					if (seedFiles.get().length || localSeedPaths.find((path) => existsSync(path))) {
+						loadSeedModule();
+					}
+					const eagerReloadIntegrationSeedPaths = seedFiles
+						.get()
+						// Map integration seed paths to URLs, if possible.
+						// Module paths like `@example/seed` will be ignored
+						// from eager reloading.
+						.map((s) => (typeof s === 'string' && s.startsWith('.') ? new URL(s, root) : s))
+						.filter((s): s is URL => s instanceof URL);
```
ematipico, line 130: nit: you should flip them :)

ematipico (approved):
Looks good, I left a minor nitpick

packages/db/src/core/integration/index.ts
bholmesdev: Ah, I see where you're coming from but we actually cannot flip these. The `map()` is used to convert relative path strings to URLs, then the `filter()` consolidates to all URL objects. If we put the `filter()` first, those relative path strings would get filtered out

PR #10605, opened 2024-03-29
reset history title after push/replace but prior to assignment to location

martrapp:
## Changes

We change the history title while we push / replace the history during view transition.
It got changes back at the end of the function which turned out to be too late when using text fragments (`#:~:text=...`) as Chrome pushes another state.

## Testing

**Edit:** ~~n./a.~~ Tested manually.

## Docs

bug fix n./a.

### DIFF
--- a/.changeset/eighty-olives-destroy.md
+++ b/.changeset/eighty-olives-destroy.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue with outdated page titles in browser history when using text fragments in view transition navigation.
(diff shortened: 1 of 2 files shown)


matthewp:
lgtm, but can you add a changeset?

martrapp:
> lgtm, but can you add a changeset?

Looked so small and innocent. Done.

ematipico (approved):
I suppose you tested the change manually?

martrapp:
> I suppose you tested the change manually?

Hi @ematipico, yes, sue ;-)  Single stepped through the code and was really surprised that Chrome pushes another state to the history when assigning a text fragment for the same page. The "moveToLocation()" code could certainly do with a reworking, but at least it's not wrong now.

Issue #9103, opened 2023-11-14
static route in subfolder gets overridden by Rest-parameter


frankbits:
### Astro Info

```block
Astro                    v3.5.4
Node                     v18.15.0
System                   Windows (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

I had the files `index.astro` and `[slug].astro` in `pages/`, as well as a static route in a subfolder.

As the first two were pretty much the same, I wanted to combine them.
As Astro doesn't provide an optional route-parameter (eg. `[[slug]]` or `[slug]` with `slug: undefined`) I switched to the Rest-parameter.
In `npm run dev` all work as intended, but `npm run build` overrides a file generated by a static route in a subfolder as it could also be catched by `[...slug].astro`.

This contradicts the documentation on routing:
> Statische Routen ohne Pfadparameter haben Vorrang vor allen anderen Routen

### What's the expected result?

`[..slug].astro` only generates pages, which are not already generated by a static route.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xykwes

### Participation

- [X] I am willing to submit a pull request for this issue.

jbmoelker:
I'm experiencing a similar issue with dynamic route matching in `output: server` mode. Not sure if it has the same cause.

```
Astro                    v4.2.3
Node                     v20.9.0
System                   macOS (x64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/sitemap
```

I have the following file structure:

```
pages/
  [locale]/
    [...path]/index.astro
    index.astro
```
In `output: hybrid` mode I get:

```
dist/en/index.html
dist/en/page-1.html
dist/nl/index.html
dist/nl/pagina-1.html
etc
```
So visiting `/en/` returns the page rendered with the template matching `pages/[locale]/index.astro`.

However in `output: server` mode, that same `/en/` URL now matches `pages/[locale]/[...path]/index.astro` where the `path` parameter is `undefined`. While I'd expect it to match the same route as in the other output modes.

Also happy to work on a solution if someone can point me where to start :)

jbmoelker:
[My issue above](https://github.com/withastro/astro/issues/9103#issuecomment-1918767021) has been resolved by a [patch in Astro v4.2.4](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#424). Thanks @Fryuni 🙌

[...]

PR #10607, opened 2024-03-29
Fix leading slash creating incorrect conflict resolution between pages generated from static routes and rest parameters

frankbits:
## Changes

Following the PR https://github.com/withastro/astro/pull/10298 to fix issue https://github.com/withastro/astro/issues/9103, my specific problem was still not fixed, as mentioned in https://github.com/withastro/astro/issues/9103#issuecomment-1986952264.
In https://github.com/withastro/astro/issues/9103#issuecomment-1986965476 I explained my following findings:
- having a "static path"-object without a leading slash works fine (trailing slash is fine):
```
{
	slug: 'test/ing/',
	title: 'Rest TestIng',
},
```
- but adding a trailing slash if a Rest-parameter can generate the same page as a static page, it will overwrite it in build-mode, while everything works as expected in dev-mode:
```
{
	slug: '/test/ing/',
	title: 'Rest TestIng',
},
```
- this is related to the generated route being `//test/ing/`
    - first hotfix was to just replace leading `//` with `/` https://github.com/frankbits/astro/commit/737d40717749807088cfdabb2ba53a8ffb44802b 
- to fix this, I replaced the direct call to `route.generate()` with `stringifyParams()` which calls `route.generate()` internally, after trimming slashes https://github.com/frankbits/astro/commit/78d604cf01ed531540e9b503e6ea1af2a9678ee5
    - I also removed the `JSON.stringify()` of the return-value of `stringifyParams()`, as the returned value of `route.generate()` is already a string, and I had to parse it back to a string without `""` and the other 2 places where the function is used should not have any problem anywhere without the `JSON.stringify()`, as the return-value i used as an array-key to store/retrieve "static path"-objects and only have to fit to each other.

## Testing

I added https://github.com/frankbits/astro/commit/a913d0484e66739629134256eb4d4083346eff85 ( edited https://github.com/frankbits/astro/commit/7cba77b4fd72eac24044bbc313aa9ace232fda08, https://github.com/frankbits/astro/commit/22fb67e4decbd44b6a43a9bbdc72df3c98870b7f) a test with the slug having a leading slash (similar to the failing path in my [MRE](https://stackblitz.com/edit/github-xykwes) of my original issue https://github.com/withastro/astro/issues/9103)

Also added a test for leading slash in a dynamic route https://github.com/withastro/astro/pull/10607/commits/d4bd99e6a55a588d2acf21c93d1d430fb728bec9 (see https://github.com/withastro/astro/pull/10607#issuecomment-2028441036)

All test in `packages/astro/test/dynamic-route-collision.test.js` finish successfully.

I couldn't run all test of the project, as `pnpm run test` ran into errors setting up tests.

```
@astrojs/telemetry:test: cache miss, executing cfa2dc085a22837a
@astrojs/telemetry:test: 
@astrojs/telemetry:test: > @astrojs/telemetry@3.0.4 test C:\Users\Frank\PhpstormProjects\astro\packages\telemetry
@astrojs/telemetry:test: > astro-scripts test "test/**/*.test.js"
@astrojs/telemetry:test: 
@astrojs/telemetry:test: node:internal/errors:490
@astrojs/telemetry:test:     ErrorCaptureStackTrace(err);
@astrojs/telemetry:test:     ^
@astrojs/telemetry:test: 
@astrojs/telemetry:test: Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:test/reporters
@astrojs/telemetry:test:     at new NodeError (node:internal/errors:399:5)
@astrojs/telemetry:test:     at ESMLoader.builtinStrategy (node:internal/modules/esm/translators:259:11)
@astrojs/telemetry:test:     at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14) {
@astrojs/telemetry:test:   code: 'ERR_UNKNOWN_BUILTIN_MODULE'
@astrojs/telemetry:test: }
@astrojs/telemetry:test: 
@astrojs/telemetry:test: Node.js v18.15.0
@astrojs/telemetry:test: ERROR: command finished with error: command (C:\Users\Frank\PhpstormProjects\astro\packages\telemetry) C:\Users\Frank\AppData\Roaming\npm\pnpm.cmd run test exited (1)
```

## Docs

I have not updated the docs, as my change is not in conflict with the documented behaviour.
But it could probably be sensible to add information to the routing-doc about leading and trailing slashes in dynamic and rest-parameters. (as mentioned in https://github.com/withastro/astro/issues/9103#issuecomment-1986967627)


frankbits:
I just noticed, that calling `removeTrailingForwardSlash()` in packages/astro/src/core/build/generate.ts:362, packages/astro/src/core/build/generate.ts:377 and possibly other places would be redundant with this change, as trailing slashes should already be trimmed by `stringifyParams()`.

frankbits:
Maybe the Changeset should be updated, to include dynamic-routes too, I don't know if this issue also exists for those, or if yes, then if this fix would fix it too.

frankbits:
> Maybe the Changeset should be updated, to include dynamic-routes too, but I don't know if this issue also existed for those, or if yes, then if this fix would fix it too.

Added second test for dynamic route with leading slash, which I locally ran with and without my fix.
Dynamic Routes normally don't allow any slashes, with my change leading and trailing slashes can now be added, as they get trimmed before the validation.
I don't think this would be a problem, as only adding leading/trailing slashes still describes one directory-level.

ematipico (approved):

PR #10611, opened 2024-03-29
fix: update vercel node support status

jacobdalamb:
## Changes

- Rename `current` to `default`
- Add `retiring` status with `removal: Date | string`
- Update the current supported node version to 20
- Add info for `retiring` status


packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -522,9 +522,15 @@ function getRuntime(process: NodeJS.Process, logger: AstroIntegrationLogger): Ru
...
-	if (support.status === 'current') {
+	if (support.status === 'default') {
 		return `nodejs${major}.x`;
 	}
+	if (support.status === 'retiring') {
+	  logger.info(
+      `Your project is being built for Node.js ${major} as the runtime, which is retiring by ${support.removal}.`
+    );
+    return `nodejs${major}.x`;
+	}
```
lilnasy, line 533: I don't think we need to log this because there is no user action to be taken. 18 has a long road of support ahead of it.

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -59,11 +59,11 @@ const ISR_PATH = `/_isr?${ASTRO_PATH_PARAM}=$0`;
 // https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version
 const SUPPORTED_NODE_VERSIONS: Record<
 	string,
-	{ status: 'current' } | { status: 'beta' } | { status: 'deprecated'; removal: Date }
+	{ status: 'default' } | { status: 'beta' } | { status: 'retiring'; removal: Date | string } | { status: 'deprecated'; removal: Date }
 > = {
 	16: { status: 'deprecated', removal: new Date('February 6 2024') },
```
lilnasy, line 64: We can remove this. Astro doesn't support 16 to begin with.

lilnasy (approved):
Looks great to me! Just a couple of non-blocking comments.

packages/integrations/vercel/src/serverless/adapter.ts
jacobdalamb: Well for security support it does, which ends in a year (https://endoflife.date/nodejs)

bluwy: I agree with Arsh, 18 is still under maintenance LTS so it should be fine to keep using it. I think it's helpful to have a retiring status, but maybe only warn if it's around 3 months away?

jacobdalamb: Like this? https://github.com/withastro/astro/pull/10611/commits/5131f126d3a9cf23a3f573f56e7bd7e3155b501e

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -522,9 +521,15 @@ function getRuntime(process: NodeJS.Process, logger: AstroIntegrationLogger): Ru
...
-	if (support.status === 'current') {
+	if (support.status === 'default') {
 		return `nodejs${major}.x`;
 	}
+	if (support.status === 'retiring' && support.warnDate && new Date() >= support.warnDate) {
+	  logger.warn(
+      `Your project is being built for Node.js ${major} as the runtime, which is retiring by ${support.removal}.`
+    );
+    return `nodejs${major}.x`;
+	}
```
bluwy, line 532: Should this be something like:
```suggestion
	if (support.status === 'retiring') {
	  if (support.warnDate && new Date() >= support.warnDate) {
	    logger.warn(
        `Your project is being built for Node.js ${major} as the runtime, which is retiring by ${support.removal}.`
      );
    }
    return `nodejs${major}.x`;
	}
```

Otherwise currently (before the warn date) it won't return the string.

.changeset/clever-houses-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/vercel": patch
+---
+
+Update supported Node version
```
ematipico, line 5:
```suggestion
Updates the supported version of Node.js
```

The correct name is `Node.js`

packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -59,11 +59,10 @@ const ISR_PATH = `/_isr?${ASTRO_PATH_PARAM}=$0`;
 // https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version
 const SUPPORTED_NODE_VERSIONS: Record<
 	string,
-	{ status: 'current' } | { status: 'beta' } | { status: 'deprecated'; removal: Date }
+	{ status: 'default' } | { status: 'beta' } | { status: 'retiring'; removal: Date | string; warnDate: Date } | { status: 'deprecated'; removal: Date }
 > = {
-	16: { status: 'deprecated', removal: new Date('February 6 2024') },
-	18: { status: 'current' },
-	20: { status: 'beta' },
+	18: { status: 'retiring', removal: "Early 2025", warnDate: new Date('October 1 2024') },
```
ematipico, line 64: nit: can we use ISO dates? e.g. `2024-10-1`

ematipico (approved):

bluwy (approved):

packages/integrations/vercel/src/serverless/adapter.ts
lilnasy: counternit 😆: I would have to think for a second whether 10 is the month or the day.

lilnasy (approved):

PR #10618, opened 2024-03-30
Lazy loaded shiki languages during syntax highlighting

43081j:
## Changes

There's a few changes in this PR, each contributing to memory usage improvement. Bundled into one PR since all but one are close to one-liners.

- Lazily load shiki languages if they're not already loaded
  - Rather than falling back to plain text, we attempt to load the language first
  - We default to `['plaintext']` rather than all languages, leading to all others being lazily loaded by default
  - This results in a chunky reduction of memory since we're no longer holding shiki's many language modules in memory for no reason
- Discarding contexts after build end
  - Some contexts were being held around permanently (i.e. retained in memory). By discarding them on build end, we allow them to be garbage collected
- Move static image helper into a factory
  - To avoid capturing local scopes, it has been moved into a factory
  - Since we store it on `globalThis`, we were holding deep references to large objects before
  - Now we no longer hold those refs, they can be garbage collected

### Memory benchmark

Before:

|                 | Elapsed time (s) | Memory used (MB) | Final memory (MB) |
| :-------------- | ---------------: | ---------------: | ----------------: |
| SSR build       |             3.11 |            91.74 |            128.05 |
| Client build    |             0.11 |            -0.04 |            128.26 |
| Static generate |             0.91 |             4.88 |            132.53 |

After:

|                 | Elapsed time (s) | Memory used (MB) | Final memory (MB) |
| :-------------- | ---------------: | ---------------: | ----------------: |
| SSR build       |             3.50 |            61.52 |             97.82 |
| Client build    |             0.10 |           -41.66 |             56.20 |
| Static generate |             0.90 |             5.02 |             51.29 |

I did a few runs and got roughly these results each time.

## Testing

- Extra test added for shiki highlighting via a lazily loaded language
- Memory benchmarking run to observe benefit of changes
- Full test run for regression

## Docs

We may need to update docs on the fact that the default shiki languages are now `['plaintext']`, although it shouldn't make much difference to a user since the rest will be lazily loaded anyway if they use them.

## For reviewers

Each commit describes its own change, so it may be worth reading the commit messages and/or the diff of the individual commit if you want to understand better.


43081j:
making this a draft for now as its possible we can't have these nice things unless rehype lets us do async 👀

bluwy:
Thanks! All these changes make sense to me. About shiki lazy load, we actually used to do the same but I removed it because I thought shikiji/shiki 1.0 already did that by default, but in hindsight it probably didn't because its highlighter functions are all sync anyways (which you pointed out).

About your message on Discord on version bumps, this will require a major for `@astrojs/markdown-remark` as it's published. Adding a major changeset for it should be enough and changesets will auto bump the rest (I hope). There's already many things within `@astrojs/markdown-remark` that had been marked deprecated, and this seems like the right time to clean it all up.

Would you be able to add a major changeset for`@astrojs/markdown-remark`? I'll handle the other major changes in another PR.

packages/markdown/remark/src/shiki.ts
```diff
@@ -43,18 +48,24 @@ export async function createShikiHighlighter({
...
-		langs: langs.length ? langs : Object.keys(bundledLanguages),
+		langs: ['plaintext', ...langs],
 		themes: Object.values(themes).length ? Object.values(themes) : [theme],
 	});
 
-	const loadedLanguages = highlighter.getLoadedLanguages();
-
 	return {
-		highlight(code, lang = 'plaintext', options) {
+		async highlight(code, lang = 'plaintext', options) {
```
bluwy, line 56: I think this line in `Code.astro` will also need an await:

https://github.com/withastro/astro/blob/b3028caecf1a96310e2ee7a65075a81c231d87be/packages/astro/components/Code.astro#L90-L93

Not sure why the tests aren't failing for this.

43081j:
have fixed the broken test

FYI it was working because of this it seems:
https://github.com/withastro/astro/blob/0be26d8b10fb19efaa9c1d2f88a4950d9550994a/packages/astro/src/runtime/server/escape.ts#L99-L103

i suppose `<Fragment set:html=${promise}>` results in it awaiting the promise internally via the above, so it was working fine

bluwy (approved):
Thanks!

.changeset/real-rabbits-bake.md
```diff
@@ -0,0 +1,7 @@
+---
+"@astrojs/markdown-remark": major
+---
+
+This changes the `markdown-remark` package to lazily load shiki languages by
+default (only preloading `plaintext`). Additionally, highlighting is now an
+async task due to this.
```
bluwy, line 7:
```suggestion
Updates Shiki syntax highlighting to lazily load shiki languages by default (only preloading `plaintext`). Additionally, the `createShikiHighlighter()` API now returns an asynchronous `highlight()` function due to this.
```

minor tweaks

PR #9720, opened 2024-01-18
Improve rendering performance for common data types

bluwy:
## Changes

Matthew and Chris mentioned before that we could optimize the rendering for common data types, e.g. strings. This PR moves those data types earlier in the if condition checks to improve perf. It also:

[...]

PR #10624, opened 2024-03-31
feat: rework child rendering to use class

43081j:
Quite a lot of garbage collection happens when rendering due to the fact we create anonymous functions when buffering writes (the `catch`, the destination/write function, etc).

By extracting this logic into a class, we can instead rely on prototype methods which exist once in memory.

## Testing

Existing tests should cover this.

Manual testing has happened via the benchmark script to see if any benefit was gained.

Before:

| Page   | Avg (ms) | Stdev (ms) | Max (ms) |
| :----- | -------: | ---------: | -------: |
| /astro |   203.76 |      38.25 |   315.99 |
| /md    |     2.73 |       1.99 |    22.16 |
| /mdx   |   118.25 |      17.55 |   220.04 |

After:

| Page   | Avg (ms) | Stdev (ms) | Max (ms) |
| :----- | -------: | ---------: | -------: |
| /astro |   133.88 |      25.51 |   313.56 |
| /md    |     2.44 |       1.99 |    22.00 |
| /mdx   |   113.79 |      17.49 |   206.58 |

## Docs

N/A

## Notes to reviewers

This isn't a huge gain so i'm happy to throw it away if you think the numbers are lying, or the code is worse.

the whole `renderChild` buffering is interesting. thanks to the buffering, we do a _lot_ of garbage collection (which slows execution). however, its probably still faster than doing them in serial.

i do wonder if we can rework to avoid allocating/processing individual buffers per child

### DIFF
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -181,7 +181,7 @@ class BufferedRenderer implements RenderDestination {
 			destination.write(chunk);
 		}
 
-		// NOTE: We don't empty `bufferChunks` after it's written as benchmarks show
+		// NOTE: We don't empty `this.chunks` after it's written as benchmarks show
 		// that it causes poorer performance, likely due to forced memory re-allocation,
 		// instead of letting the garbage collector handle it automatically.
 		// (Unsure how this affects on limited memory machines)
--- a/packages/astro/src/runtime/server/render/util.ts
+++ b/packages/astro/src/runtime/server/render/util.ts
@@ -150,29 +150,34 @@ export function renderElement(
 }
 
 const noop = () => {};
+
+/**
+ * Renders into a buffer until `renderToFinalDestination` is called (which
+ * flushes the buffer)
+ */
 class BufferedRenderer implements RenderDestination {
-	private __chunks: RenderDestinationChunk[] = [];
-	private __renderPromise: Promise<void> | void;
-	private __destination?: RenderDestination;
+	private chunks: RenderDestinationChunk[] = [];
+	private renderPromise: Promise<void> | void;
+	private destination?: RenderDestination;
 
 	public constructor(bufferRenderFunction: RenderFunction) {
-		this.__renderPromise = bufferRenderFunction(this);
+		this.renderPromise = bufferRenderFunction(this);
 		// Catch here in case it throws before `renderToFinalDestination` is called,
 		// to prevent an unhandled rejection.
-		Promise.resolve(this.__renderPromise).catch(noop);
+		Promise.resolve(this.renderPromise).catch(noop);
 	}
 
 	public write(chunk: RenderDestinationChunk): void {
-		if (this.__destination) {
-			this.__destination.write(chunk);
+		if (this.destination) {
+			this.destination.write(chunk);
 		} else {
-			this.__chunks.push(chunk);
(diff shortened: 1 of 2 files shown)


bluwy:
!bench render

bluwy (approved):
Awesome! The results look great and I think this is worth the change. I tried to improve the rendering part too but didn't get far (https://github.com/withastro/astro/pull/9720)

> the whole renderChild buffering is interesting. thanks to the buffering, we do a lot of garbage collection (which slows execution). however, its probably still faster than doing them in serial.
>
> i do wonder if we can rework to avoid allocating/processing individual buffers per child

Do you mean sharing a single buffer between all children? It could be faster but I'm not sure how to queue the second buffer-part until the first is done that way. It's not possible to know the size of a buffer-part ahead of time.

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -149,6 +149,46 @@ export function renderElement(
 	return `<${name}${internalSpreadAttributes(props, shouldEscape)}>${children}</${name}>`;
 }
 
+const noop = () => {};
+class BufferedRenderer implements RenderDestination {
+	private __chunks: RenderDestinationChunk[] = [];
+	private __renderPromise: Promise<void> | void;
+	private __destination?: RenderDestination;
```
bluwy, line 156: tiny nit, but is there a reason to prefix them with `__`? I think TypeScript already guarantees private fields to not be accessible.

43081j: Just habit really, it is convention in most of the repos I work in and eslint is often configured to pick up on it

If you prefer different, I'm happy to drop the prefix 👍

bluwy: Yeah I prefer without the prefix so it's consistent with the codebase, e.g.

https://github.com/withastro/astro/blob/b3028caecf1a96310e2ee7a65075a81c231d87be/packages/astro/src/runtime/server/transition.ts#L130-L132

https://github.com/withastro/astro/blob/b3028caecf1a96310e2ee7a65075a81c231d87be/packages/astro/src/runtime/server/render/astro/instance.ts#L15-L22

43081j:
Basically we spend a lot of time waiting for garbage collection each time we finish a `renderChild`

Mostly because we discard the buffer and the promise we were waiting for. 

By itself, that isn't much. But when you have thousands of promises in memory, the clean up of them takes a while

I feel like there's a solution that will just click one day in my head 😅 something to do with having one big queue and sorting it at the end before flushing it

43081j:
have updated the private properties and rebased onto main FYI

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -149,6 +149,51 @@ export function renderElement(
...
+		}
+	}
+
+	public async renderToFinalDestination(destination: RenderDestination) {
+		// Write the buffered chunks to the real destination
+		for (const chunk of this.chunks) {
+			destination.write(chunk);
+		}
+
+		// NOTE: We don't empty `bufferChunks` after it's written as benchmarks show
```
bluwy, line 184:
```suggestion
		// NOTE: We don't empty `this.chunks` after it's written as benchmarks show
```

matthewp:
This looks great, thanks so much @43081j !

PR #10625, opened 2024-03-31
fix(page-data): add key to allPages

### LINKED ISSUE #10622: BUG: injectRoute merge routes in production when entrypoint is already used
Hi everyone 👋 

Duplicate my message from #3602 , with two minimal reproductions.

I'm currently making an [Astro integration](https://github.com/goulvenclech/astropi/) where my `injectRoute`s **works perfectly in dev mode**, but where all my paths are weirdly merged in production, even with different pattern. This is caused by using the same `.astro` entrypoint for two or more `injectRoute`s.

### Astro Info

```block
Package Manager          pnpm
Adapter                  none
Integrations             astropi
```

### Simple example with static path

Link to Minimal Reproducible Example -> https://stackblitz.com/edit/github-l3a17a?file=astro.config.mjs

```ts
    injectRoute({
      pattern: `/blog`,
      entrypoint: './src/entrypoint.astro',
      prerender: true,
    });
    injectRoute({
      pattern: `/another-blog`,
      entrypoint: './src/entrypoint.astro',
      prerender: true,
    });
```

^ This will work in dev. But in prod, only `/blog/index.html` is generated.

### Complete example with static & spread paths

To clarify the bug, I have made another (not so) minimal repro -> https://stackblitz.com/edit/github-l3a17a?file=src%2F%5B...slug%5D.astro

This one have spread routes like : 
```ts
          injectRoute({
            pattern: `/blog`,
            entrypoint: './src/blog.astro',
            prerender: true,
          });
          injectRoute({
            pattern: `/blog/[...slug]`,
            entrypoint: './src/[...slug].astro',
            prerender: true,
          });
          injectRoute({
            pattern: `/another-blog`,
            entrypoint: './src/blog.astro',
            prerender: true,
          });
          injectRoute({
            pattern: `/another-blog/[...slug]`,
            entrypoint: './src/[...slug].astro',
            prerender: true,
          });
```

In dev, everything work as intended, with 5 pages generated with the correct URLs : 
```
/ <- src/pages/index.astro
/blog <- src/blog.astro
/another-blog <- src/blog.astro
/blog/test <- src/[...slug].astro
/another-blog/another-test  <- src/[...slug].astro
```

Here how it's merged in build :
```
generating static routes 
12:06:19 ▶ src/blog.astro
12:06:19   └─ /blog/index.html (+24ms)
12:06:19 ▶ src/[...slug].astro
12:06:19   ├─ /blog/test/index.html (+29ms)
12:06:19   └─ /blog/another-test/index.html (+6ms)
12:06:19 ▶ src/pages/index.astro
12:06:19   └─ /index.html (+1ms)
12:06:19 ✓ Completed in 176ms.
```

We loose one the blog's index. And one article is merged into the wrong URL.

Firstly I don't see why the `entrypoints` influence the generation of routes, when the pattern and the collections are different. Secondly, because the behaviour differs between development and production, so if this issue is closed "as not planed" we should at least limit the behaviour in development with a warning.

### Participation

- [X] I am working on a pull request for this issue, see #10625

goulvenclech:
## Current changes

**As now :**
Fixes #10622 , by changing the logic of "getting all pages by their component" by "getting all pages by a key" in the build part. Inspired by how the PR #10248 implemented the same solution in the route caching part. Also, remove a bunch of generators (verbose and inneficients).

Since this doesn't change anything in the public API, I guess it's a `patch`? But because I did deep changes in the build process, I'll understand if you prefer this to be a `minor`. I leave that to reviewers 🙂

**For Astro 5 :** 
`internals` has a new function `getPageDatasWithPublicKey()`. It only exist to avoid breaking change in the integration API, but maybe we should remove it at the next major update?

## Testing

@Fryuni provided me with an integration test, see `packages/astro/test/reuse-injected-entrypoint.test.js`. I don't plan to change the tests apart from that one.

## Docs

Should this be documented? I feel like this is just the expected behaviour.


goulvenclech:
Okay @Fryuni @Princesseuh , the good news is: I fixed the key problem in `allPages` without breaking existing tests.

The bad news is: it didn't solve my problem. So I guess the problem runs deeper... 🕵🏻‍♂️

goulvenclech:
Current status : 
The problem is probably inside `packages/astro/src/core/build/static-build.ts` in the `viteBuild()` definition.
At `	const { allPages, settings } = opts;` (line 39) , `allPages` contains every pages.
At the end of the function line 132, `internals.pageToBundleMap` does not.

goulvenclech:
On an unrelated (but kind of related) note: 
I find some naming in this part of the code REALLY confusing. For example,  for `allPages` et `internal.pagesByComponents`, the key "component" is sometimes called "path", "pageName", "page" or "component" indistinctively.

goulvenclech:
There is still a `pagesByKeys.get()` to fix in `packages/astro/src/core/build/plugins/plugin-pages.ts`, but this bad boy will require a bit more work...

goulvenclech:
Still a lot of work to get rid of the "getting all pages by their component" logic everywhere. But I'm still thinking : should this use a key (like the PR mentioned above), or should this use path ? (Could be more useful + two routes can't have the same path).

goulvenclech:
Struggling to understand `internals.entrySpecifierToBundleMap` & `virtualModuleName`/`virtualModulePageName` and their usage... If someone can provide me some help / informations / context, it could help me a lot.

Fryuni:
`internals.entrySpecifierToBundleMap` is a mapping from the ID of the virtual module that represents an entrypoint to the Astro application (aka, what should be imported to render a route) to the name of the file on disk, part of the bundle, that exports that entrypoint. It is populated at the final stage of a Vite build before Astro starts pre-rendering pages.

`virtualModuleName`/`virtualModulePageName` is the name of a Virtual Module that will be populated with compiled information from an Astro file. If it were the name of a regular file, Rollup (the underlying bundler used by Vite) would try to read the file, and other plugins might attempt to process it as if it were a regular file. It isn't a regular file, it is used for things like:
- The styles extracted from a `.astro` component.
- The client-side scripts from a `.astro` component that need to be bundled only for the client (contrary to the code for the Astro component itself that is only for the server).

packages/astro/src/core/build/plugins/plugin-manifest.ts
```diff
@@ -189,7 +189,10 @@ function buildManifest(
 	}
 
 	for (const route of opts.manifest.routes) {
-		const pageData = internals.pagesByComponent.get(route.component);
+		let pageData;
+		for (const page of internals.pagesByKeys.values()) {
+			if (page.route.route == route.route) pageData = page;
+		}
```
Fryuni, line 195:
```suggestion
		const key = `${route.route}_${route.component}`;
		const pageData = internals.pagesByKeys.get(key);
```

Maybe have a `computeRouteKey` or similarly named function to generate the key consistently everywhere it might be needed.

Fryuni:
> I find some naming in this part of the code REALLY confusing. For example, for `allPages` et `internal.pagesByComponents`, the key "component" is sometimes called "path", "pageName", "page" or "component" indistinctively.

The value from `component` is called different things in different places because it means different things in those different places.

A "Route", conceptually in Astro, can be either:
- An entry from the filesystem routing in the project
- An injected route from an integration
- A redirect

When it comes from the filesystem or from an integration, the "component" is the entrypoint defining the route. It is called `page`/`pageName` in those cases because it is the name under which the pages built from that route will be displayed.

When it is a redirect, the "component" really _is_ a path. It is the target of the redirect.

---

That being confusing is great feedback, thank you! We have been talking recently exactly about where things could be confusing for new contributors.

matthewp:
Great example of some of the more complicated internals and how they sort of just grew and never got a proper refactor / rethinking. We're unlikely to do that, but renaming properties to make them make more sense, and adding inline documentation to explain what they are for, is something that could be done more easily.

goulvenclech:
@Fryuni Thanks a lot for the first review. This week I have a lot of work, that's why I didn't continue this PR, but I'll come back to it soon.

For the key I'm wondering : I've choose this key pattern because it's the same one as @ematipico in #10248 . But if we want to do something different, maybe we could just use the path (`getPageByPath`) or the pattern (`getPageByPattern`) since, unlike the component, it doesn't seem possible to have two pages with the same path? 

Or maybe we shouldn't use a Record here ? And I should change that to a Set or an Array ? But I'm afraid this will break even more things.

goulvenclech:
@matthewp I won't do any refactor/ rethinking since I'm not in the Astro team and I would like this to be a minor fix. But I'll simplify some over-engineered parts (I already deleted some generators), add documentation, and change some naming. Here is my course of action:

1. Fix the bug and get the behavior I want, as a POC
2. Make sure this hasn't broken anything (I don't plan to change the tests apart from the one I added), especially in the public API for integrations
3. Clean up after myself, including improving naming and documentation.

Fryuni:
> it doesn't seem possible to have two pages with the same path?

It is. Two integrations can define the same dynamic route, like `[slug]` and render different slugs each. That is perfectly fine.

goulvenclech:
Currently on : 
```
  for (const pageData of Object.values(opts.allPages)) {
	  if (routeIsRedirect(pageData.route)) {
		  continue;
	  }
	  inputs.add(getVirtualModulePageNameFromPath(ASTRO_PAGE_MODULE_ID, pageData.component));
  }
```

Where `Object.values(opts.allPages))` has everything, but it's merged in `inputs` at the end... `getVirtualModulePageNameFromPath/2`? 👀

goulvenclech:
Btw, I don't think it's really useful to have `getVirtualModulePageIdFromPath/1` function declared in `packages/astro/src/core/build/plugins/plugin-pages.ts` to be only used once in `packages/astro/src/core/build/internal.ts`, inside a `getEntryFilePathFromComponentPath/1` function that is only used twice in `packages/astro/src/core/build/generate.ts`.

Those kinds of over-engineered and low-documented situations are pretty confusing for a new contributor. I'll try to simplify them.

goulvenclech:
@ematipico @Fryuni Sorry but I may need help to understand this :
```
  const path = getPathFromVirtualModulePageName(RESOLVED_SPLIT_MODULE_ID, id);
  const virtualModuleName = getVirtualModulePageNameFromPath(ASTRO_PAGE_MODULE_ID, path);
  let module = await this.resolve(virtualModuleName);
  if (module) {
	  // we need to use the non-resolved ID in order to resolve correctly the virtual module
	  imports.push(`import * as pageModule from "${virtualModuleName}";`);
  }
```

goulvenclech:
Stuck on a error, from [9306c71](https://github.com/withastro/astro/pull/10625/commits/9306c7113fbc1b8f958f299603908b70604b5a95) or older :
```
19:51:20 [ERROR] [vite] x Build failed in 10ms
[vite:load-fallback] Could not load @astro-page:another-blog@_@src/entrypoint@_@astro: The argument 'path' must be a string, Uint8Array, or URL without null bytes. Received '\x00@astro-page:another-blog@_@src/entrypoint@_@astro'
  Stack trace:
    at open (node:internal/fs/promises:600:10)
    at Object.load (file:///Users/goulvenclech/Documents/astro/node_modules/.pnpm/vite@5.1.6_@types+node@18.19.24_sass@1.72.0/node_modules/vite/dist/node/chunks/dep-jvB8WLp9.js:66717:43)
    at async file:///Users/goulvenclech/Documents/astro/node_modules/.pnpm/rollup@4.13.0/node_modules/rollup/dist/es/shared/node-entry.js:18634:33
 ELIFECYCLE  Command failed with exit code 1.
```

I'll investigate that, then there is four `// TODO: change that` where I've identified things to fix.

goulvenclech:
I stop there for tonight.

About the `The argument 'path' must be a string, Uint8Array, or URL without null bytes` error. Here what returns `console.log('vitePluginPages – loaded id', JSON.stringify(id));` putted just after `async load(id) {` in packages/astro/src/core/build/plugins/plugin-pages.ts.

In my branch : 
```
vitePluginPages – loaded id "\u0000@astro-page:another-blog@_@src/entrypoint@_@astro"
vitePluginPages – loaded id "\u0000@astro-page:blog@_@src/entrypoint@_@astro"
vitePluginPages – loaded id "\u0000@astro-page:@_@src/pages/index@_@astro"
vitePluginPages – loaded id "\u0000@astrojs-manifest"
20:39:36 [ERROR] [vite] x Build failed in 12ms
vitePluginPages – loaded id "/Users/goulvenclech/Documents/astro/packages/astro/dist/core/app/index.js"
vitePluginPages – loaded id "\u0000astro:ssr-manifest"
vitePluginPages – loaded id "\u0000noop-middleware"
```

In the `main` branch : 
```
vitePluginPages – loaded id "\u0000@astro-page:src/entrypoint@_@astro"
vitePluginPages – loaded id "\u0000@astro-page:src/pages/index@_@astro"
vitePluginPages – loaded id "\u0000@astrojs-manifest"
vitePluginPages – loaded id "/Users/goulvenclech/Documents/github-l3a17a/src/entrypoint.astro"
vitePluginPages – loaded id "/Users/goulvenclech/Documents/github-l3a17a/src/pages/index.astro"
vitePluginPages – loaded id "/Users/goulvenclech/Documents/astro/packages/astro/dist/core/app/index.js"
vitePluginPages – loaded id "\u0000astro:ssr-manifest"
vitePluginPages – loaded id "\u0000noop-middleware"
```

I may have broken something there.

goulvenclech:
Last commit : 

1. drop the null byte in the key for a `&` separator, and drop the `@_@` in the virtualModuleName also for a `&` separator. Mainly for consistency and to avoid confusion with other separators.
2. fix `getVirtualModulePageName` and `retrieveRoutesToGenerate`, now we correctly get the `pageData` from the `virtualModulePageName` of every pages.

TLDR: everything works for SSG!

goulvenclech:
**Next steps :** 

1. understanding the usage of `getVirtualModulePageName` in `packages/astro/test/test-utils.js`
2. understand why do we need to get PageData from PageData in `packages/astro/src/core/build/generate.ts`
3. refactor `getPageDatasByHoistedScriptId` so it's not a generator anymore + can work with multiples PageData
4. after 2 & 3, remove unused `getPageDataByComponent` from `Internals`

I'll see from there if the test will run.

**Problem with SSG mode :** 

Go a cryptic error, I need to investigate more.

```
23:50:25 
 finalizing server assets 

23:50:25 [build] Rearranging server assets...
23:50:25 [@astrojs/vercel/serverless] Bundling function ../../_functions/entry.mjs
File /Users/goulvenclech/Documents/github-l3a17a/.vercel/output/_functions/entry.mjs does not exist.
  Stack trace:
    at Job.emitDependency (/Users/goulvenclech/Documents/astro/node_modules/.pnpm/@vercel+nft@0.26.4/node_modules/@vercel/nft/out/node-file-trace.js:297:23)
    at async nodeFileTrace (/Users/goulvenclech/Documents/astro/node_modules/.pnpm/@vercel+nft@0.26.4/node_modules/@vercel/nft/out/node-file-trace.js:52:5)
    at async VercelBuilder.buildServerlessFolder (file:///Users/goulvenclech/Documents/astro/packages/integrations/vercel/dist/serverless/adapter.js:293:25)
    at async withTakingALongTimeMsg (file:///Users/goulvenclech/Documents/astro/packages/astro/dist/integrations/index.js:23:18)
    at async AstroBuilder.build (file:///Users/goulvenclech/Documents/astro/packages/astro/dist/core/build/index.js:145:5)
 ELIFECYCLE  Command failed with exit code 1.
 ```

**Comment by goulvenclech** [comment]
Error when mixing route `prerender: true,` and route `prerender: false,` in SSR : 

```
12:11:51 [ERROR] NoMatchingRenderer: Unable to render `noop`.

No valid renderer was found for this file extension.
    at renderFrameworkComponent (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/chunks/astro_Dy1b-4yy.mjs:2062:15)
    at async renderComponent (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/chunks/astro_Dy1b-4yy.mjs:2271:10)
    at async renderComponentToString (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/chunks/astro_Dy1b-4yy.mjs:2314:28)
    at async renderPage (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/chunks/astro_Dy1b-4yy.mjs:2480:17)
    at async lastNext (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/entry.mjs:1471:25)
    at async callMiddleware (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/entry.mjs:1005:10)
    at async RenderContext.render (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/entry.mjs:1494:22)
    at async NodeApp.render (file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/entry.mjs:2053:18)
    at async file:///Users/goulvenclech/Documents/github-l3a17a/dist/server/entry.mjs:2404:24
```

**Comment by goulvenclech** [comment]
Last commit add a function to avoid breaking change in the Integration API... But maybe it should be removed at the next big update? (like Astro 6)

I've added a TODO on it. I'll let you decide.

**Comment by goulvenclech** [comment]
In tests, `smoke/docs` failed to build because of those injectRoutes :
https://github.com/expressive-code/expressive-code/blob/9eb861924c0e28228f18314f9f32590151f6264d/packages/astro-expressive-code/src/index.ts#L66-L85

Investigating.

**Comment by goulvenclech** [comment]
Update: I can't reproduce it on MacOS, maybe it's related to a different behaviour between Linux and MacOS regarding files pathing? 🧐

**Comment by Princesseuh** [comment]
!preview route-key

**Comment by Princesseuh** [comment]
!preview route-key

File: packages/astro/src/core/build/generate.ts [hunk 2]
```diff
@@ -259,7 +256,8 @@ async function generatePage(
 	// prepare information we need
 	const { config, internals, logger } = pipeline;
 	const pageModulePromise = ssrEntry.page;
-	const pageInfo = getPageDataByComponent(internals, pageData.route.component);
+	// TODO: Why do we need to get PageData from PageData?
+	const pageInfo = getPageData(internals, pageData.route.component, pageData.route.route);
```
matthewp, line 260: Is it the same exact object?

goulvenclech: @matthewp Hum...

First, it looks like it's the same exact object.

Secondly, this code is broken haha `getPageDataByComponent(internals, pageData.route.component);` will always return `undefined`.

Luckily, It looks like `pageInfo` is only used by `const scripts = pageInfo?.hoistedScript ?? null;`, and then `scripts` is gived to : 
```ts
const generationOptions: Readonly<GeneratePathOptions> = {
    pageData,
    linkIds,
    scripts,
    styles,
    mod: pageModule,
};
```

And then never used anywhere? Tbh, none of those params (except `mod`) seems used anywhere... Legacy code?

Princesseuh (approved):
I was with @goulvenclech in a call pretty much the entire time he made this PR, and can confirm that it works locally. Not familiar enough to say if it has deeper complications, though.

packages/astro/src/core/build/static-build.ts
```diff
@@ -374,12 +369,17 @@ async function cleanStaticOutput(
...
-		if (pageData.route.prerender && !pageData.hasSharedModules) {
-			const { moduleSpecifier } = pageData;
-			const pageBundleId = internals.pageToBundleMap.get(moduleSpecifier);
-			const entryBundleId = internals.entrySpecifierToBundleMap.get(moduleSpecifier);
+	for (const pageData of internals.pagesByKeys.values()) {
+		const { moduleSpecifier } = pageData;
+		const pageBundleId = internals.pageToBundleMap.get(moduleSpecifier);
+		const entryBundleId = internals.entrySpecifierToBundleMap.get(moduleSpecifier);
+		if (pageData.route.prerender && !pageData.hasSharedModules && allStaticFiles.has(pageBundleId ?? entryBundleId)) {
 			allStaticFiles.add(pageBundleId ?? entryBundleId);
```
Fryuni, line 377: This doesn't seem right...

Part of the condition is `allStaticFiles.has(pageBundleId ?? entryBundleId)`, which means that adding it to the set will do nothing since it is already in the set.

Also, since the set starts off empty, it seems like `allStaticFiles` will always remain empty as this block will never be reached.

packages/astro/src/core/build/plugins/plugin-ssr.ts
```diff
@@ -167,12 +171,10 @@ function vitePluginSSRSplit(
...
 				const contents: string[] = [];
 				const exports: string[] = [];
-
-				const path = getPathFromVirtualModulePageName(RESOLVED_SPLIT_MODULE_ID, id);
-				const virtualModuleName = getVirtualModulePageNameFromPath(ASTRO_PAGE_MODULE_ID, path);
+				const componentPath = getComponentFromVirtualModulePageName(RESOLVED_SPLIT_MODULE_ID, id);
+				const virtualModuleName = getVirtualModulePageName(ASTRO_PAGE_MODULE_ID, componentPath);
 				let module = await this.resolve(virtualModuleName);
 				if (module) {
-					// we need to use the non-resolved ID in order to resolve correctly the virtual module
```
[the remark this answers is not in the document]
goulvenclech: Idk why I deleted it, I'll restore it.

packages/astro/src/core/build/static-build.ts
goulvenclech: Good catch! Should be fixed now.

packages/astro/src/core/build/static-build.ts
```diff
@@ -367,19 +362,27 @@ async function runPostBuildHooks(
...
 	internals: BuildInternals,
 	ssrOutputChunkNames: string[]
 ) {
 	const allStaticFiles = new Set();
-	for (const pageData of eachPageData(internals)) {
-		if (pageData.route.prerender && !pageData.hasSharedModules) {
-			const { moduleSpecifier } = pageData;
-			const pageBundleId = internals.pageToBundleMap.get(moduleSpecifier);
-			const entryBundleId = internals.entrySpecifierToBundleMap.get(moduleSpecifier);
+	const allSSRFiles = new Set();
```
Fryuni, line 373: I'd just rename this since the files added here are not necessarily "all SSR" like the "all static" above, they are not entirely static in opposition to the const above.

Maybe renaming them to `staticOnlyFiles` and `nonStaticOnlyFiles` would be more clear.

Fryuni (approved):
Looking GREAT to me!

I left just a naming suggestion that I think would improve clarity; I got a bit confused about what it really meant for a file to be in that set.

bluwy (approved):
Awesome work! The logic all looks good to me.

packages/astro/src/core/build/internal.ts
```diff
@@ -221,16 +216,77 @@ export function* getPageDatasByClientOnlyID(
...
+	const pageDatas: PageBuildData[] = [];
+		internals.pagesByKeys.forEach((pageData) => {
+			if (component === pageData.component) pageDatas.push(pageData);
+	})
+	return pageDatas;
+}
+
+/**
+ * TODO: This function is used to avoid breaking changes in the Integrations API.
+ * Should be removed in the future (in Astro 5 ?).
```
ematipico, line 253: nit: A `TODO` shouldn't be part of the JSDoc, we should move it in a simple `//` comment. Also, would you mind adding some context? Like, unfortunately, input types and output types are the same, so we should explain **what's the breaking** and how integrations will be affected in Astro 5.

The reason I am asking this is because if we decided to go ahead with the breaking change, we need to provide a migration path to the integrations, and if you're not around, I would like to have this information stored in a comment so we can copy it and give it to downstream users.

packages/astro/src/core/build/plugins/util.ts
```diff
@@ -63,13 +74,34 @@ export function getVirtualModulePageNameFromPath(virtualModulePrefix: string, pa
...
+export function getPagesFromVirtualModulePageName(internals: BuildInternals, virtualModulePrefix: string,  id: string): PageBuildData[]
+{
+	const path = getComponentFromVirtualModulePageName(virtualModulePrefix, id);
+
+	const pages: PageBuildData[] = [];
+	internals.pagesByKeys.forEach(pageData => {
+		if (pageData.component === path) {
+			pages.push(pageData);
+		}
+	});
```
ematipico, line 91: nit: can you use a `for..of` instead of a `forEach`?

packages/astro/src/core/build/static-build.ts
ematipico: In the docs, we use the term "on-demand". So maybe `onDemandFiles` is a better fit.

packages/astro/src/core/build/static-build.ts
```diff
@@ -367,19 +362,27 @@ async function runPostBuildHooks(
...
+	for (const pageData of internals.pagesByKeys.values()) {
+		const { moduleSpecifier } = pageData;
+		const pageBundleId = internals.pageToBundleMap.get(moduleSpecifier);
+		const entryBundleId = internals.entrySpecifierToBundleMap.get(moduleSpecifier);
+		if (pageData.route.prerender && !pageData.hasSharedModules && !allSSRFiles.has(pageBundleId ?? entryBundleId)) {
 			allStaticFiles.add(pageBundleId ?? entryBundleId);
+		} else {
+			allSSRFiles.add(pageBundleId ?? entryBundleId);
+			// Check if the component was not previously added to the static build by a statically rendered route
+			if (allStaticFiles.has(pageBundleId ?? entryBundleId)) {
```
ematipico, line 383: We are computing **four times** `pageBundleId ?? entryBundleId`. Let's do it once and save it in a variable.

packages/astro/src/core/build/plugins/util.ts
Princesseuh: This is a Map object, on which `forEach` is typically recommended unlike with arrays

packages/astro/src/core/build/internal.ts
goulvenclech: I don't see the point of putting it in a simple comment rather than a JSdoc? The entire function is affected.

But I've added some context about the old behavior, and why this feature is useful :
```ts
/**
 * TODO: This function is used to avoid breaking changes in the Integrations API.
 * Should be removed in the future (in Astro 5 ?).
 * Parse internals.pagesByKeys to get the page data with the public key.
 * If the page component is unique -> the public key is the component.
 * This match the old behavior of the Integrations API, where pages were identified by the component.
 * If the page component is shared -> the public key is the internal key.
 * This should be the new behavior, since we identify pages by their internal key now.
 * Until then, this is not a breaking change, because a shared entrypoint was not supported before.
 * @param pagesByKeys A map of all page data by their internal key
 */
export function getPageDatasWithPublicKey(pagesByKeys: Map<string, PageBuildData>): Map<string, PageBuildData> {
```

ematipico: >I don't see the point of putting it in a simple comment rather than a JSdoc?

IDEs will render JSDoc, and the `TODO` will be part of the documentation of the function. `TODO` aren't documentation, but internal things.

Princesseuh: This doesn't really matter here, the entire file (and this function) is internal (as its name suggest). If anything, I'd argue that the `TODO` is part of the documentation of the function, as it signifies that it is temporary and shouldn't be used without considerations.

There's already a few places in Astro with this, in similar contexts (and not, I actually added a TODO to a user-facing method by accident, looking at the code, ha)

goulvenclech: I'm not an Astro maintainer, so perhaps I don't fully understand your documentation culture. Nevertheless, I am surprised to read that _« [internals documentation] needs the same treatment as public APIs »_ when this part of the code had so little documentation before I contributed, especially almost no JSdoc on the exported functions (and even one empty JSdoc).

Moreover, while I believe that internal documentation should receive as much effort as public documentation (as demonstrated by my PR), I think these two types of documentation do not share the same purpose and should not be treated the same. In this specific case, the `TODO:` purpose was : 
1. assist you during the major update.
2. indicate that it's just a temporary function, which probably should not be used outside its current reference, and definitely should not be used without understanding the context of its inception. That's why I was OK with it showing in the IDE.

But no worries, I have removed the `TODO:` from the JSdoc and took the opportunity to refine the comment a bit. There's no reason to spend more time on this detail, let's move forward 👍

Princesseuh: If a public function was planned to be removed in a future version, we would publicly document it, like deprecations, otherwise you risk leading users down the wrong path for the future.

As such, in my mind the planned impending removal of something is definitely a relevant part to include in the contract you show to the user, especially when the user here are the other developers of the project. (Of course, in a public function it'd be written more eloquently than a TODO, but this is fine for internal code)

Anyway, this is the nittiest of nit at this point, and not really a good usage of the time of anyone involved 😅 If we feel strongly about this, we can change it later ourselves.

.changeset/great-turtles-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Multiples pages can now have the same component as entrypoint. This is purely internal, this aligns the build process with the behaviour in development server, no public API changes.
```
sarah11918, line 5:
```suggestion
Multiple pages can now use the same component as an `entrypoint`. This change is purely internal, and aligns the build process with the behaviour in the development server.
```
just some tiny language edits! Please adjust if anything is inaccurate!

sarah11918 (commented):
Just reviewing for docs, and have some tiny language edits for the changeset!

goulvenclech:
Thanks @sarah11918 !

BTW, I just remembered that @ematipico suggested adding a usage example in the docs (See in the issue -> https://github.com/withastro/astro/issues/10622#issuecomment-2036803965 ).

Is this still relevant? Should I take care of this, and what should I have in my doc PR?

sarah11918:
Hi @goulvenclech ! I checked out the docs in question, and I think we only need a docs PR if something in that section/instruction now works differently. As you wrote in your issue, this may simply have been an unexpected limitation, and people just assumed it would have already worked this way. I will also tag in @Fryuni since he understands the underlying feature better than I do!  😄 

Is anything on the [Integrations API reference page](https://docs.astro.build/en/reference/integrations-reference/) now incorrect or misleading as a result of your feature?

.changeset/great-turtles-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Multiple pages can now use the same component as an `entrypoint`. This change is purely internal, and aligns the build process with the behaviour in the development server.
```
sarah11918, line 5:
```suggestion
Adds the ability for multiple pages to use the same component as an `entrypoint` when building an Astro integration. This change is purely internal, and aligns the build process with the behaviour in the development server.
```
@goulvenclech - Another thought:

Also reading back through the old issue and comments there, perhaps it is worth calling out (if true!) that this relates exclusively to building integrations? I know when people get deep into their own features, it can be easy to have a lot of context in one's head. But, this will be read by people simply updating their own site's version of Astro, and providing some guidance as to whether or not this is something that affects them is often helpful!

(Revisiting this also gives me the opportunity to use the format I *really* want, which is to start with my present tense verb like "Adds"/"Fixes"/"Removes"... 😄 )

goulvenclech:
> Is anything on the [Integrations API reference page](https://docs.astro.build/en/reference/integrations-reference/) now incorrect or misleading as a result of your feature?

Not incorrect nor misleading, the doc didn't talk about since it wasn't decided (different behavior between dev and prod). I think there are two options:

1. In [the Astro Integration API documentation at the "InjectRoute" section](https://docs.astro.build/en/reference/integrations-reference/#injectroute-option), add a paragraph clarifying that we can use the same Astro component as an entrypoint for several pages. Illustrate with the use case -> inject several blogs (using the same layout component blog.astro and [...entry].astro) for several paths provided by the user.

2. Acknowledge that this would fill the documentation with a marginal use case, and that it is not worth it.

Your call 🙂 both suit me, and I can quickly do the PR doc.

.changeset/great-turtles-greet.md
goulvenclech: > when building an Astro integration.

I'm not sure we want to add this... I think currently the `injectRoute` function is the only way to use this change (maybe I'm wrong) so it's not wrong.... BUT this PR removed the technical limitation in the build process, not in the integration API. So if there is another way to inject routes or there will be one in the future, everyone will be able to use this, integration or not.

> (Revisiting this also gives me the opportunity to use the format I really want, which is to start with my present tense verb like "Adds"/"Fixes"/"Removes"... 😄 )

Noted for next time!

sarah11918: @goulvenclech 

> Adds the ability for multiple pages to use the same component as an `entrypoint` when building an Astro integration.
 
Thanks for the reply and would love to get this just right! Here's my thinking, and tell me where I'm off:

This does say "when building an Astro integration" as a *use case* (not that the Integration API itself was actually changed), so if we're aiming at helping readers figure out whether or not the change affects them, I think this might be helpful even if it doesn't describe *what* internal change now allows them this new flexibility when building an integration?

I will also note that this is gonna get buried in a changelog soon, so I think it's OK to write for what this unlocks and allows *now* vs thinking about potential uses in the future. If and when it becomes possible for someone to use this in another situation, presumably we will have docs for that new situation?

goulvenclech:
@sarah11918 Just commited your changeset. Noted for "when building an Astro integration" as an use case, makes sense. And noted for "marginal use case", I think this is the expected behavior anyway.

Thanks for your time 🙂

.changeset/great-turtles-greet.md
Fryuni: I agree with Sarah regarding this being a "marginal use case". I don't see any need to add this on Astro Docs.

sarah11918 (approved):
Approving on behalf of docs! Thank you @goulvenclech ! 🙌

ematipico (approved):

PR #10635, opened 2024-04-01
Give proper error when seed missing default export

matthewp:
## Changes

- Was logging the function and not calling it, oops!

## Testing

Tested manually

## Docs

N/A bug fix


bholmesdev (approved):
Yikes, `exp`ert catch!

PR #10636, opened 2024-04-01
db: Better error messages when querying remote

bholmesdev:
## Changes

Introduce well-formatted errors for remote db client exceptions.
- Add `AstroDbError` class to set "Astro DB Error" as the error name instead of `AstroUserError`.
- Format an error message using the expected error "code" and "details" from Studio.
- Special case "no such table" issues to show a hint to push the latest schema changes.
- Misc: fix "unhandled promise rejection" getting logged for eager seeding by adding a `.catch()` with manual logging.

### Before
image

### After

Suspected "missing db push" errors:

image

General SQLite errors:

image

### DIFF
--- a/packages/db/src/runtime/db-client.ts
+++ b/packages/db/src/runtime/db-client.ts
@@ -66,7 +66,7 @@ export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string
 				const json = await res.json();
 				remoteResult = remoteResultSchema.parse(json);
 			} catch (e) {
-				throw new AstroDbError(await getUnexpectedErrorMessage(res));
+				throw new AstroDbError(await getUnexpectedResponseMessage(res));
 			}
 
 			if (method === 'run') return remoteResult;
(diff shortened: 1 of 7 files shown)


packages/db/src/runtime/db-client.ts
```diff
@@ -149,3 +138,36 @@ export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string
...
+	} catch (e) {
+		return new AstroDbError(await getUnexpectedResponseMessage(response));
+	}
+	// Strip LibSQL error prefixes
+	let details =
+		error.details?.replace(/.*SQLite error: /, '') ??
+		`(Code ${error.code}) \nError querying remote database.`;
+	let hint = `See the Astro DB guide for query and push instructions: https://docs.astro.build/en/guides/astro-db/#query-your-database`;
+	if (error.code === KNOWN_ERROR_CODES.SQL_QUERY_FAILED && details.includes('no such table')) {
+		hint = `Did you run \`astro db push\` to push your latest table schemas?`;
```
matthewp, line 170: nice addition

matthewp (approved):

PR #10638, opened 2024-04-01
refactor: rename "config" file to "virtual"

bholmesdev:
## Changes

The `runtime/config` file has slowly evolved from holding the config helper (`defineDb`) to holding all static exports in the virtual module (`sql`, `NOW`, etc). Rename to `runtime/virtual` for clarity.

## Testing

`pnpm pack` and install in another project to ensure types still work.


.changeset/plenty-owls-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/db": minor
```
bholmesdev, line 2: Made `minor` since this renames an `exports` entry to `runtime/virtual.js`. This export is undocumented and is only used by Astro DB internally (not Astro core). Just being extra cautious with versioning.

ematipico (approved):

PR #10642, opened 2024-04-02
Move format option from LocalImageProps to ImageSharedProps

OliverSpeir:
## What does this change?

Moves format and quality options from `LocalImageProps` to `ImageSharedProps` so it doesn't cause a type error when used with remote images, these always should have been shared

### DIFF
--- a/.changeset/lovely-trees-wait.md
+++ b/.changeset/lovely-trees-wait.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Moves format option from LocalImageProps to ImageSharedProps to solve type error when using format with remote images
--- a/.changeset/lovely-trees-wait.md
+++ b/.changeset/lovely-trees-wait.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Moves format option from LocalImageProps to ImageSharedProps to solve type error when using format with remote images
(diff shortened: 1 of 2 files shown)


packages/astro/src/assets/types.ts
```diff
@@ -153,15 +162,6 @@ export type LocalImageProps<T> = ImageSharedProps<T> & {
...
-	/**
-	 * Desired output format for the image. Defaults to `webp`.
-	 *
-	 * **Example**:
-	 * ```astro
-	 * <Image src={...} format="avif" alt="..." />
-	 * ```
-	 */
-	format?: ImageOutputFormat;
 	/**
```
Princesseuh, line 165: `quality` should also be on shared, it's also relevant for remote images. Not sure why everything is on Local...

.changeset/lovely-trees-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Moves format option from LocalImageProps to ImageSharedProps to solve type error when using format with remote images
```
florian-lefebvre, line 5:
```suggestion
Fixes typing issues when using `format` and `quality` options with remote images
```

Princesseuh (approved):

PR #10646, opened 2024-04-02
fix(db): isolate AstroDbError from core utils

lilnasy:
## Changes

- What does this change?

### DIFF
--- a/packages/db/src/runtime/seed-local.ts
+++ b/packages/db/src/runtime/seed-local.ts
@@ -3,7 +3,7 @@ import { type SQL, sql } from 'drizzle-orm';
 import type { LibSQLDatabase } from 'drizzle-orm/libsql';
 import { SQLiteAsyncDialect } from 'drizzle-orm/sqlite-core';
 import { type DBTables } from '../core/types.js';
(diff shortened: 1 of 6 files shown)


lilnasy:
!preview 10646

packages/db/src/runtime/seed-local.ts
```diff
@@ -3,7 +3,7 @@ import { type SQL, sql } from 'drizzle-orm';
 import type { LibSQLDatabase } from 'drizzle-orm/libsql';
 import { SQLiteAsyncDialect } from 'drizzle-orm/sqlite-core';
 import { type DBTables } from '../core/types.js';
-import { AstroDbError } from '../utils.js';
```
bholmesdev, line 6: Wow, I'm surprised this would cause an issue. `utils.js` is at the project root, _not_ in the `core` directory

bholmesdev (approved):
Hadn't considered that `src/utils` exposes core utilities, making this unsafe _and_ uncaught by our eslint rule. Good reason to avoid shared files in `src` I suppose. Even if this does not resolve the issue, I approve of the change. Thanks for utelling us

PR #10648, opened 2024-04-02
feat(toolbar): Allow using Preact easily

Princesseuh:
## Changes

This PR adds an API to use Preact easily inside dev toolbar apps. Apps can now be `.(t|j)sx` files and use Preact directly, components can be imported and additionally, a `definePreactToolbarApp` helper is provided to easily start.

Basic examples:
```tsx
// my-app.tsx - using the normal defineToolbarApp
import { defineToolbarApp } from "astro:toolbar";
import { render } from "astro:toolbar:preact";

export default defineToolbarApp({
	init(canvas) {
		render(<h1>Hello, Preact!</h1>, canvas);
	},
})
```

```tsx
// my-app.tsx - using definePreactToolbarApp
import { definePreactToolbarApp, type App } from "astro:toolbar:preact";

const MyCoolApp: App = () => {
	return <h1>Hello, Preact!</h1>;
};

export default definePreactToolbarApp(MyCoolApp);
```
(for brevity sake, only the JSX versions are shown here, but you can of course use `h` directly or any of the no build -step ways that Preact offer)

`astro:toolbar:preact` exports all of Preact, its hooks and signals. A separate module had to be used to ensure we don't end up conflicting with a Preact version in the user's project.

## Testing

Added a test with a Preact app


florian-lefebvre:
Sorry I missclicked from gh mobile! I can't set it back to draft :/

packages/astro/src/vite-plugin-dev-toolbar/vite-plugin-dev-toolbar.ts
```diff
@@ -96,7 +120,62 @@ export default function astroDevToolbar({ settings, logger }: AstroPluginOptions
...
+
+				if (!shouldTryTransform) {
+					return;
+				}
+
+				// TODO: Seems like a fairly old school way to detect JSX, maybe there's a better way?
+				const isJSXToolbarHeuristic =
+					code.includes('astro:toolbar:preact') ||
+					id.endsWith('?toolbar-app') ||
+					code.includes('astro/toolbar-preact');
```
Princesseuh, line 147: I tried a few ways before landing on this, most notably, bundling the entire toolbar app from its entrypoint. That worked, but it meant that apps had to be bundling compatible and on more complex apps, you instantly ran into warnings or errors (not to mention needing to fiddle with `external` and stuff for virtual modules etc)

This heuristic works just fine, however it is technically possible for a component to not trigger it:
- If it has no imports to `astro:toolbar:preact` (which means it doesn't use any features from Preact)
- Is not the entrypoint of the app
- Does not have a pragma comment

If anyone has a better solution, I'll take gladly. It's a bit tough to not conflict with Vite or worse, our renderers

Princesseuh:
This doesn't work and won't actually look like this, so, closing.

PR #10649, opened 2024-04-02
feat: automatic Markdoc partial resolution

### LINKED ISSUE #10418: Custom markdoc tag to include markdoc file does not render react component
### Astro Info

```block
Astro                    v4.5.2
Package Manager          npm
Adapter                  none
Integrations             @astrojs/markdoc
                         @astrojs/react
```

_No response_

### Describe the Bug

I'm trying to create a custom markdoc tag to include other markdoc files, however it doesn't seem to work when those included files use markdoc tags that are rendered using react.

I've implemented my [`include-markdoc` custom tag](https://stackblitz.com/edit/github-mkxshb-tsleyv?file=src%2Fmarkdoc%2Finclude-markdoc.ts) using a transform function, based on how [markdoc implements partials](https://github.com/markdoc/markdoc/blob/a13cc583fa3b0420538e772a75628dd15d72b793/src/tags/partial.ts) ([docs](https://markdoc.dev/docs/partials)).

It does work with included files that use markdoc tags that are rendered using astro though.

The error I get is:

```
  TypeError: Cannot read properties of undefined (reading 'toString')
    at Object.check (/home/projects/github-mkxshb/node_modules/@astrojs/react/server.js:29:32)
    at renderFrameworkComponent (/home/projects/github-mkxshb/node_modules/astro/dist/runtime/server/render/component.js:112:33)
    at async Module.renderComponent (/home/projects/github-mkxshb/node_modules/astro/dist/runtime/server/render/component.js:374:10)
    at async AstroComponentInstance.render (/home/projects/github-mkxshb/node_modules/astro/dist/runtime/server/render/astro/instance.js:44:7)
    at async Object.render (/home/projects/github-mkxshb/node_modules/astro/dist/runtime/server/render/component.js:356:7)
    at async renderChild (/home/projects/github-mkxshb/node_modules/astro/dist/runtime/server/render/any.js:36:5)
```

Which points to https://github.com/withastro/astro/blob/main/packages/integrations/react/server.js#L22
It seems like a `check` function to see whether the node passed in is a legitimate React component or not. From console logging, the `Component passed in is something like

```
{
  type: 'local',
  path: './src/components/alert/Warning',
  namedExport: undefined,
  [Symbol(@astrojs/markdoc/component-config)]: true
}
```

So my issue is, how would I pass in a react renderer that `@astro/react` understands? Is there some exposed function somewhere I can use to transform the object above into a valid React component?

A side issue is that, maybe the `check` function needs to accommodate for this situation and not throw an error? It currently crashes the dev server and fails to build.

I can write a custom markdoc tag to include other markdoc files which contain React rendered components.

https://stackblitz.com/edit/github-mkxshb-tsleyv

### Participation

bholmesdev:
Resolves #10418 

## Changes

This adds automatic resolution for Markdoc partials by relative file path.

```md
<!--src/content/blog/post.mdoc-->

{% partial file="./_diagram.mdoc" /%}

<!--src/content/blog/_diagram.mdoc-->

## Diagram

<figure>
<img src="..." />
<figcaption>Diagram embedded into <code>post.mdoc</cod>
</figure>
```

- **Crawl Markdoc AST to discover any `partial` tags.** When one is found, render the partial in-place and assign the node. Flattening lets us catch validation errors at buildtime and discover nested component tags for hoisted scripts.
- Ensure manually configured `partials` are still respected to make this a non-breaking feature
- Track file changes to related partials using `addWatchFile()`
- Update validator to ignore any "partial not found" errors.

## Testing

Add tests to render components and HTML within a partial.

## Docs

https://github.com/withastro/docs/pull/7743

### DIFF
--- a/.changeset/moody-spies-learn.md
+++ b/.changeset/moody-spies-learn.md
@@ -1,0 +1,19 @@
+---
+"@astrojs/markdoc": minor
+---
+
+Add automatic resolution for Markdoc partials. This allows you to render other Markdoc files inside of a given entry. Reference files using the `partial` tag with a `file` attribute for the relative file path:
+
+```md
+<!--src/content/blog/post.mdoc-->
+
+{% partial file="my-partials/_diagram.mdoc" /%}
+
+<!--src/content/blog/my-partials/_diagram.mdoc-->
+
+## Diagram
+
+This partial will render inside of `post.mdoc.`
+
+![Diagram](./diagram.png)
+```
--- a/.changeset/moody-spies-learn.md
+++ b/.changeset/moody-spies-learn.md
@@ -1,5 +1,5 @@
 ---
-"@astrojs/markdoc": minor
+"@astrojs/markdoc": patch
 ---
 
 Add automatic resolution for Markdoc partials. This allows you to render other Markdoc files inside of a given entry. Reference files using the `partial` tag with a `file` attribute for the relative file path:
--- a/packages/integrations/markdoc/package.json
+++ b/packages/integrations/markdoc/package.json
@@ -59,7 +59,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc",
     "build:ci": "astro-scripts build \"src/**/*.ts\"",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "astro-scripts test --timeout 40000 \"test/**/*.test.js\""
+    "test": "astro-scripts test --timeout 40000 --parallel \"test/**/*.test.js\""
   },
   "dependencies": {
     "@astrojs/internal-helpers": "workspace:*",
--- a/packages/integrations/markdoc/package.json
+++ b/packages/integrations/markdoc/package.json
@@ -59,7 +59,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc",
     "build:ci": "astro-scripts build \"src/**/*.ts\"",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "astro-scripts test --timeout 40000 \"test/**/*.test.js\""
+    "test": "astro-scripts test --timeout 60000 \"test/**/*.test.js\""
   },
   "dependencies": {
     "@astrojs/internal-helpers": "workspace:*",
--- a/packages/integrations/markdoc/package.json
+++ b/packages/integrations/markdoc/package.json
@@ -59,7 +59,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc",
     "build:ci": "astro-scripts build \"src/**/*.ts\"",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "astro-scripts test --timeout 40000 --parallel \"test/**/*.test.js\""
+    "test": "astro-scripts test --timeout 40000 \"test/**/*.test.js\""
   },
   "dependencies": {
     "@astrojs/internal-helpers": "workspace:*",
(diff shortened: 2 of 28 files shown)


bholmesdev:
Seems the test file contains too many prod build calls and times out. Will try to reuse fixtures and split the `render.test.js` file

matthewp (approved):

sarah11918:
@bholmesdev Just wanted to check on the naming convention here Ben, because we are currently referring to "page partials" as just plain old "partials" (I guess, the only kind we had so it didn't much matter): https://docs.astro.build/en/basics/astro-pages/#page-partials

So, I'm wondering whether in conjunction with your PR documenting Markdoc partials we should update wording on the Pages page to always say "page partial" and whether we should be clear to always say "Markdoc partial" when referring to these files.

The joys of adding the *second* thing that use the same word.  😄

packages/integrations/markdoc/src/content-entry-type.ts
```diff
@@ -142,6 +138,124 @@ export const Content = createContentComponent(
...
+			if (!file) {
+				throw new MarkdocError({
+					// Should be caught by Markdoc validation step.
+					message: `(Uncaught error) Partial tag requires a 'file' attribute`,
+				});
+			}
+
+			if (markdocConfig.partials?.[file]) continue;
+
+			const partialPath = path.resolve(path.dirname(fileURLToPath(fileUrl)), file);
```
Princesseuh, line 177: Can't this use `ctx.pluginContext.resolve` instead? That way you could use aliases, partials from packages etc.

.changeset/moody-spies-learn.md
```diff
@@ -0,0 +1,19 @@
+---
+"@astrojs/markdoc": patch
+---
+
+Add automatic resolution for Markdoc partials. This allows you to render other Markdoc files inside of a given entry. Reference files using the `partial` tag with a `file` attribute for the relative file path:
+
+```md
+<!--src/content/blog/post.mdoc-->
+
+{% partial file="my-partials/_diagram.mdoc" /%}
```
ematipico, line 10: It's worth mentioning here or in the docs from **which folder** the partials are solved from. `srcDir`?

packages/integrations/markdoc/package.json
```diff
@@ -59,7 +59,7 @@
     "build": "astro-scripts build \"src/**/*.ts\" && tsc",
     "build:ci": "astro-scripts build \"src/**/*.ts\"",
     "dev": "astro-scripts dev \"src/**/*.ts\"",
-    "test": "astro-scripts test --timeout 40000 \"test/**/*.test.js\""
+    "test": "astro-scripts test --timeout 40000 --parallel \"test/**/*.test.js\""
```
ematipico, line 62: Reminder that we should remove this before merging the PR

ematipico (commented):
Can we have at least **one** test that raises an error in case we attempt to render a partial that doesn't exist or it's an incorrect folder? This would help to set down expectations.

packages/integrations/markdoc/package.json
bholmesdev: I'm still unsure how to resolve my testing issue without this adjustment. Will try to discuss in team-platform

.changeset/moody-spies-learn.md
bholmesdev: Ah, it is a relative path from whatever file you're currently in. So no folder specifically! This is clearer in the docs, so I'll update the changeset to match

packages/integrations/markdoc/package.json
ematipico: Does that mean that without `--parallel` your test will always fail?

PR #10655, opened 2024-04-02
Fix building static sites with Astro DB

matthewp:
## Changes

- When building a site with `output: 'static'` (or hybrid), the temporary app token is needed to build pages.
- Pass through in this condition so this is used if the env doesn't exist (which won't most of the time).

## Testing

- New test fixture added.

## Docs

N/A, bug fix


bholmesdev (approved):
thanks for stattacking this bug quickly!

Issue #10620, opened 2024-03-31
`Astro.preferredLocale` is always `undefined` in "hybrid" output mode with `prerender = false`


stefanprobst:
### Astro Info

```block
Astro                    v4.5.12
Node                     v20.12.0
System                   Linux (x64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/node
Integrations             none
```

### Describe the Bug

i want to use `Astro.preferredLocale` to [redirect the "/" root route](https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid/blob/main/src/pages/index.astro) to the user's preferred locale (or the default locale). i am using `hybrid` output mode, and have [set `export const prerender = false` in `src/pages/index.astro`](https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid/blob/main/src/pages/index.astro#L2), and [configured `redirectToDefaultLocale: false` in `astro.config.mjs`](https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid/blob/main/astro.config.mjs#L11).

When logging both `Astro.preferredLocale` and `Astro.request.headers.get("Accept-Language")`, i see that `Astro.preferredLocale` is `undefined`, even though the `Accept-Language` headers are logged correctly.

## how to reproduce

0. clone the repo at https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid
1. `pnpm run build`
2. `pnpm run preview`
3. open http://localhost:3000
4. see server logs:
  ```
  Preferred locale: undefined
  Accept-Language header: de,en-US;q=0.7,en;q=0.3
  ```

## observations

- when changing "output" mode from `hybrid` to `server`, it correctly prints: "Preferred locale: de"
- when not using any import from `astro:i18n`, i.e. commenting out [this line](https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid/blob/main/src/layouts/page.astro#L5), it also works correctly in hybrid mode
- changing `src/pages/index.astro` to a `src/pages/index.ts` GET route does not make a difference

### What's the expected result?

see above

### Link to Minimal Reproducible Example

https://github.com/stefanprobst/issue-astro-i18n-preferredlocale-hybrid

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
It seems that the prerender logic is transforming the function `computePreferredLocale` in a `noop` function. I'll have to understand why.

[...]

PR #10662, opened 2024-04-03
fix(i18n): don't make virtual modules no-op

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/10620
Closes PLT-1920

This was another tricky issue to fix. The `plugin-prerender` plugin was marking `computePreferredLocale` as a no-op function. 

In order to prevent that, I decided to go for a more generic route, in case something like this happens in the future. I created a new constant that we can use to "mark" our virtual modules as "Astro internal virtual modules", and these virtual modules won't be transformed by the plugin.

As for now, I marked the i18n plugin as internal. Not sure if we need to do this also for others. 

## Testing

I tested this one locally, using the repository provided by OP. It works as expected now.

## Docs

N/A


packages/astro/src/core/build/static-build.ts
```diff
@@ -374,7 +374,10 @@ async function cleanStaticOutput(
 ) {
 	const allStaticFiles = new Set();
 	for (const pageData of eachPageData(internals)) {
-		if (pageData.route.prerender && !pageData.hasSharedModules) {
+		if (
+			pageData.route.prerender &&
+			!(pageData.hasSharedModules || pageData.hasAstroVirtualModule)
```
matthewp, line 379: is this part the actual fix of the issue? Can you explain why this is needed if so?

ematipico: Yes this is part of the fix. This is where we collect the chunks, few lines below we clean them up by making them no-op:

https://github.com/withastro/astro/blob/317d18ef8c9cf4fd13647518e3fd352774a86481/packages/astro/src/core/build/static-build.ts#L397-L413

packages/astro/src/core/build/plugins/plugin-prerender.ts
```diff
@@ -70,6 +76,7 @@ function vitePluginPrerender(opts: StaticBuildOptions, internals: BuildInternals
 
 							opts.allPages;
 							pageInfo.hasSharedModules = hasSharedModules;
+							pageInfo.hasAstroVirtualModule = hasAstroVirtualModules;
```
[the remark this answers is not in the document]
ematipico: That's the other way around, virtual modules need special treatment for prerendered pages.

It seems that the static build takes the chunk that contains the virtual double and makes it a no-op.

I'm not very familiar with the specifics of business logic though.

bluwy:
I'm not really sure if this is the best fix for the issue. If there is `ASTRO_VIRTUAL_MODULE` to mark virtual modules, why is it only added for the `astro:i18n` virtual module. And I also don't think virtual modules should have to manually mark this.

I don't understand the issue well to suggest another fix, but I think there would be a different "bigger-picture" fix to prevent the no-op in the first place.

ematipico:
@bluwy 

All your concerns are valid, as they are my very own concerns. I was hoping that you guys had more context and knowledge about the prerendering logic and how the no-op strategy works.

Not sure if you checked the source code of `plugin-prerenderer.ts`, but there's also some dirty logic where modules that were used in the middleware AND a pre-rendered page were made no-op. 

This is even worse because the functions that are made no-op are functions inside the Astro core.

natemoo-re:
Great discussion @bluwy and @ematipico! Totally agree that this original prerendering logic has not held up as Astro's capabilities have grown. This was all written shortly after we initially added SSR support, so things like middleware, virtual modules, and our more complex bundling setup were never properly accounted for.

If we're looking for a long-term refactor, I have some context to share on how and why things work as they currently do. Our build process is currently one single Rollup build (via Vite). For features like **prerender** and **content collection caching**, we are hitting the limits of manually mutating specific chunks before/after that build process. 

The prerender `noop` replacement was alway a hacky approach intended to avoid a secondary build step just for prerendered pages. It has become increasingly apparent that we would reduce complexity by separating prerendered pages from dynamically rendered ones, so that there are no shared chunks that need special logic like this PR adds. 

The main challenges I foresee with this multi-build approach are that:
1. All routes must be compiled in order to determine which ones are prerendered vs dynamically rendered. Not 100% sure how to pull that off.
2. Certain moudles/logic will likely be duplicated when sent to two separate Rollup builds. Seems fine, tbh! 

Those tradeoffs and challenge still seem the most correct, given the number of edge cases we've had to fix with the current approach. As part of these refactors, we could even unify the divergent logic for static/hybrid/server modes since they're all essentially the same? That would definitely be a maintainability win.

lilnasy:
- Are you suggesting splitting the production build into two manifests - one containing only the `routeData` for prerendered pages, one only for on-demand rendered pages?

- Could you elaborate on "unifying" logic for static, hybrid, and server modes?

natemoo-re:
@lilnasy good questions! The manifest generation is a big open question. But yes, the gist of my recommendation would be to run pre-rendered pages in one bundler process and on-demand pages in another. The other idea would be to disable Rollup’s chunking behavior entirely for the first stage to avoid any shared chunks, then bundle just the code for the routes that should be rendered on-demand.

As for “unifying” the logic between build modes, I only mean internally. The API would remain the same. We currently have quite a few conditional branches that use an `isServerLike()` check. We probably could eliminate most of them, since “static” mode is really just “server” + all pre-rendered.

PR #10671, opened 2024-04-03
Accept common cookie attributes when deleting a cookie

### LINKED ISSUE #10480: AstroCookies.delete does not support __Host- and __Secure- prefixed cookies
### Astro Info

```block
Astro version is irrelevant as you’ll see in the bug description, but here it is:

Astro                    v3.6.5
Package Manager          pnpm
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

The `AstroCookies.delete` API does not work for cookies with `__Host-` and `__Secure-` prefixes.

From [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie_prefixes):

> `__Host-`
If a cookie name has this prefix, it's accepted in a [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) header only if it's also marked with the Secure attribute, was sent from a secure origin, does not include a Domain attribute, and has the Path attribute set to /. This way, these cookies can be seen as "domain-locked".
> `__Secure-`
If a cookie name has this prefix, it's accepted in a [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) header only if it's marked with the Secure attribute and was sent from a secure origin. This is weaker than the __Host- prefix.

The key takeaway is the `Secure` attribute requirement, which based on the main branch of the repo, is not recognized by the `AstroCookies.delete` method:

https://github.com/withastro/astro/blob/0ddc327ee85535b1a3b415ae49c664f9848b87f4/packages/astro/src/core/cookies/cookies.ts#L20

As an example, this is how the Chrome treats **Set-Cookie** headers that violate these requirements:

Screenshot 2024-03-18 at 6 38 00 PM

The `AstroCookies.delete` method should recognize all standard cookie attributes, except for the `maxAge` and `expires`, since these are the only attributes that set a valid cookie apart from its expired version. 

My current workaround is to use the `AstroCookies.set` method as follows:

```js
Astro.cookies.set("cookie-name", "deleted", {
    expires: new Date(0),
    secure: true,
    path: "/",
    httpOnly: true, 
})
```

https://stackblitz.com/edit/github-avarao?file=README.md

### Participation

fshafiee:
## Changes

Allows cookie attributes to be specified when deleting a cookie. The attributes were limited to `domain` and `path`.

See #10480 for context.

Closes #10480

## Testing

The changes are validated by unit tests.

## Docs

The [Astro.Cookie.delete API reference](https://docs.astro.build/en/reference/api-reference/#delete) must be updated, since following statement will no longer be true:

>Options allow setting the domain and path of the cookie to delete.

### DIFF
--- a/.changeset/shaggy-cats-film.md
+++ b/.changeset/shaggy-cats-film.md
@@ -1,5 +1,11 @@
 ---
-"astro": patch
+"astro": minor
 ---
 
-Accept standard cookie attributes when deleting a cookie
+Extends the acceptable cookie options when deleting a cookie, to the following attribute:
+
+- `domain`
+- `path`
+- `httpOnly` *(Added)*
+- `sameSite` *(Added)*
+- `secure` *(Added)*
--- a/.changeset/shaggy-cats-film.md
+++ b/.changeset/shaggy-cats-film.md
@@ -2,10 +2,4 @@
 "astro": minor
 ---
 
-Extends the acceptable cookie options when deleting a cookie, to the following attribute:
-
-- `domain`
-- `path`
-- `httpOnly` *(Added)*
-- `sameSite` *(Added)*
-- `secure` *(Added)*
+Adds the `httpOnly`, `sameSite`, and `secure` options when deleting a cookie
--- a/.changeset/shaggy-cats-film.md
+++ b/.changeset/shaggy-cats-film.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Accept standard cookie attributes when deleting a cookie
--- a/packages/astro/src/core/cookies/cookies.ts
+++ b/packages/astro/src/core/cookies/cookies.ts
@@ -17,7 +17,7 @@ export interface AstroCookieGetOptions {
 	decode?: (value: string) => string;
 }
 
-type AstroCookieDeleteOptions = Pick<AstroCookieSetOptions, 'domain' | 'path'>;
+type AstroCookieDeleteOptions = Omit<CookieSerializeOptions, 'expires' | 'maxAge'>;
 
 interface AstroCookieInterface {
 	value: string;
@@ -78,16 +78,21 @@ class AstroCookies implements AstroCookiesInterface {
 	 * @param options Options related to this deletion, such as the path of the cookie.
 	 */
 	delete(key: string, options?: AstroCookieDeleteOptions): void {
+		/**
+		 * The `@ts-expect-error` is necessary because `maxAge` and `expires` properties
+		 * must not appear in the AstroCookieDeleteOptions type.
+		 */
+		const {
+			// @ts-expect-error
+			maxAge: _ignoredMaxAge,
+			// @ts-expect-error
+			expires: _ignoredExpires,
+			...sanitizedOptions
+		} = options || {};
 		const serializeOptions: CookieSerializeOptions = {
 			expires: DELETED_EXPIRATION,
+			...sanitizedOptions,
 		};
-
-		if (options?.domain) {
-			serializeOptions.domain = options.domain;
-		}
-		if (options?.path) {
-			serializeOptions.path = options.path;
-		}
 
 		// Set-Cookie: token=deleted; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT
 		this.#ensureOutgoingMap().set(key, [
--- a/packages/astro/src/core/cookies/cookies.ts
+++ b/packages/astro/src/core/cookies/cookies.ts
@@ -11,7 +11,7 @@ export interface AstroCookieGetOptions {
 	decode?: (value: string) => string;
 }
 
-type AstroCookieDeleteOptions = Omit<CookieSerializeOptions, 'expires' | 'maxAge'>;
+type AstroCookieDeleteOptions = Omit<AstroCookieSetOptions, 'expires' | 'maxAge' | 'encode'>;
 
 interface AstroCookieInterface {
 	value: string;
--- a/packages/astro/src/core/cookies/cookies.ts
+++ b/packages/astro/src/core/cookies/cookies.ts
@@ -2,16 +2,10 @@ import type { CookieSerializeOptions } from 'cookie';
 import { parse, serialize } from 'cookie';
 import { AstroError, AstroErrorData } from '../errors/index.js';
 
-export interface AstroCookieSetOptions {
-	domain?: string;
-	expires?: Date;
-	httpOnly?: boolean;
-	maxAge?: number;
-	path?: string;
-	sameSite?: boolean | 'lax' | 'none' | 'strict';
-	secure?: boolean;
-	encode?: (value: string) => string;
-}
+export type AstroCookieSetOptions = Pick<
+	CookieSerializeOptions,
+	'domain' | 'path' | 'expires' | 'maxAge' | 'httpOnly' | 'sameSite' | 'secure' | 'encode'
+>;
 
 export interface AstroCookieGetOptions {
 	decode?: (value: string) => string;
--- a/packages/astro/test/units/cookies/delete.test.js
+++ b/packages/astro/test/units/cookies/delete.test.js
@@ -47,26 +47,57 @@ describe('astro/src/core/cookies', () => {
 			assert.equal(cookies.has('foo'), false);
 		});
 
-		it('can provide a path', () => {
+		it('deletes a cookie with attributes', () => {
 			let req = new Request('http://example.com/');
 			let cookies = new AstroCookies(req);
+
 			cookies.delete('foo', {
+				domain: 'example.com',
 				path: '/subpath/',
+				priority: 'high',
+				secure: true,
+				httpOnly: true,
+				sameSite: 'strict',
 			});
+
 			let headers = Array.from(cookies.headers());
 			assert.equal(headers.length, 1);
+			assert.equal(/foo=deleted/.test(headers[0]), true);
+			assert.equal(/Expires=Thu, 01 Jan 1970 00:00:00 GMT/.test(headers[0]), true);
+			assert.equal(/Domain=example.com/.test(headers[0]), true);
 			assert.equal(/Path=\/subpath\//.test(headers[0]), true);
+			assert.equal(/Priority=High/.test(headers[0]), true);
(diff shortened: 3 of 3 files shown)


lilnasy (approved):
Looks great to me! Thanks for the clear tests as well.

Do you want to create a PR to [the docs page](https://github.com/withastro/docs/edit/main/src/content/docs/en/reference/api-reference.mdx#L235) as well? I can do it if not.

packages/astro/src/core/cookies/cookies.ts
```diff
@@ -17,7 +17,7 @@ export interface AstroCookieGetOptions {
 	decode?: (value: string) => string;
 }
 
-type AstroCookieDeleteOptions = Pick<AstroCookieSetOptions, 'domain' | 'path'>;
+type AstroCookieDeleteOptions = Omit<CookieSerializeOptions, 'expires' | 'maxAge'>;
```
matthewp, line 20: We intentionally don't just support all options from `cookie` so that we know what we support and don't start supporting something accidentally because `cookie` adds new options in a minor. So I'd rather we not do this and instead explicitly add the new options we mean to support.

lilnasy: I was under the impression that our semver range would prevent that.

.changeset/shaggy-cats-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Accept standard cookie attributes when deleting a cookie
```
florian-lefebvre, line 5: After you address matthew's review, can you update the changeset to state what new properties are available?

packages/astro/src/core/cookies/cookies.ts
fshafiee: @matthewp would swapping `CookieSerializeOptions` with `AstroCookieSetOptions` address your concern?

Also, do you think it makes sense to construct `AstroCookieSetOptions` using `Pick<CookieSerializeOptions, ...>` instead of manually rewriting it? Currently, it’s missing the IDE annotations form the `cookie` package which is nice to have.

matthewp: Yes it would. Fine with using Pick as well, as long as we limit the options to what we knowingly support. Thanks!

.changeset/shaggy-cats-film.md
matthewp: Yeah we need a list of which properties we are supporting. Also this would be a `minor`.

.changeset/shaggy-cats-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
```
matthewp, line 2: Should be minor.

packages/astro/src/core/cookies/cookies.ts
fshafiee: Taken care of.

.changeset/shaggy-cats-film.md
fshafiee: Updated the change log to include the final set.

.changeset/shaggy-cats-film.md
fshafiee: Done.

.changeset/shaggy-cats-film.md
```diff
@@ -0,0 +1,11 @@
...
+"astro": minor
+---
+
+Extends the acceptable cookie options when deleting a cookie, to the following attribute:
+
+- `domain`
+- `path`
+- `httpOnly` *(Added)*
+- `sameSite` *(Added)*
+- `secure` *(Added)*
```
florian-lefebvre, line 11:
```suggestion
Adds the `httpOnly`, `sameSite` ans `secure` options when deleting a cookie
```

ematipico (approved):

ematipico:
I added this PR to the current milestone, so it will be shipped next release

packages/astro/src/core/cookies/cookies.ts
```diff
@@ -2,22 +2,16 @@ import type { CookieSerializeOptions } from 'cookie';
...
-	maxAge?: number;
-	path?: string;
-	sameSite?: boolean | 'lax' | 'none' | 'strict';
-	secure?: boolean;
-	encode?: (value: string) => string;
-}
+export type AstroCookieSetOptions = Pick<
+	CookieSerializeOptions,
+	'domain' | 'path' | 'expires' | 'maxAge' | 'httpOnly' | 'sameSite' | 'secure' | 'encode'
+>;
```
lilnasy, line 8: Partitioned should also be here. Support was added in `cookie@0.6`, which Astro has been using since 4.0.

.changeset/shaggy-cats-film.md
sarah11918: Nicely done @florian-lefebvre ! Just a typo for "and" missing Oxford comma in there, but otherwise, I approve this! :purple_heart: 

>Adds the `httpOnly`, `sameSite`, and `secure` options when deleting a cookie

sarah11918 (commented):
Florian's changeset comment is good!

Also, just to make it easy, here's where the docs change needs to be made:

https://docs.astro.build/en/reference/api-reference/#delete

And all that needs to happen is something like this:

> Marks the cookie as deleted. Once a cookie is deleted Astro.cookies.has() will return false and Astro.cookies.get() will return an [AstroCookie](https://docs.astro.build/en/reference/api-reference/#astrocookie) with a value of undefined. Options available when deleting a cookie are: `domain`, `path`, `httpOnly`, `sameSite`, and `secure`.

.changeset/shaggy-cats-film.md
florian-lefebvre: Updated! Didn't know about this Oxford thing, we don't put a coma there in French

packages/astro/src/core/cookies/cookies.ts
fshafiee: This would also change the types `AstroCookies.set` API. And I think this is what @matthewp meant by [his comment](https://github.com/withastro/astro/pull/10671#discussion_r1551744943) about being selective when it comes to cookie attribute, probably since `partitioned` is still experimental?

I personally agree with supporting it, but I believe this to be a separate conversation amongst the maintainers. 😅

lilnasy: It already has been working, I am only suggesting aligning the types with the actual current behavior.

As for its status, `cookie` doesn't consider it to be experimental - not anymore than everything else at least, the package itself is pre-1.0.

fshafiee:
Here’s the docs PR: https://github.com/withastro/docs/pull/7814

CC: @sarah11918 @ematipico

florian-lefebvre (approved):

sarah11918 (approved):
Docs approves! :partying_face:

matthewp (approved):

jlarmstrongiv:
This commit broke Astro cookie types for me.  Now, the types require all properties to be set, when previously they were optional.  In this PR, see `packages/astro/src/core/cookies/cookies.ts`

jlarmstrongiv:
For example:
```
error ts(2739): Type '{ httpOnly: boolean; maxAge: number; path: string; secure: boolean; }' is missing the following properties from type 'AstroCookieSetOptions': domain, expires, sameSite, encode
45   const defaultCookieOptions: AstroCookieSetOptions = {
```

lilnasy:
@jlarmstrongiv It seems to be fine. Could you share the code snippet?

image

nanek:
I also see the types for cookie options are all required now, as @jlarmstrongiv notes.
Running the sample @lilnasy provided above, throws the type error. Perhaps `astro info` is showing a different version that what is npm installed in that folder?

Screenshot 2024-04-11 at 9 13 25 PM


https://github.com/withastro/astro/pull/10671/files#diff-d3e7e8fe6551080102ade434b01c0baaf4f3be04289c1174294b114e466f21a0R5 is using Pick, which makes them required. 

I believe the intent was to use: 
````
export type AstroCookieSetOptions = Partial<Pick<
	CookieSerializeOptions,
	'domain' | 'path' | 'expires' | 'maxAge' | 'httpOnly' | 'sameSite' | 'secure' | 'encode'
>>;
````

lilnasy:
Thanks, I see it in a separate project, just not within the repo.

fshafiee:
`Pick` and `Omit` don’t make optional fields required, unless using an ancient version of Typescript (<3.5.1). See https://github.com/microsoft/TypeScript/issues/20722.

Here’s the [proof](https://www.typescriptlang.org/play?ts=5.4.5#code/JYOwLgpgTgZghgYwgAgILIN4Chm+VCARwC5kBnMKUAcxzwHsAHARgH5SKqRa9kmAmduUo0sAXyxYwAT0YpU-ZAF5kABWAIA1gB5UAGmQByAoUPIAPkabNDAPkkJ6ICsgBupBcszI6uE6UM4M1xfZAB6MOQAA2so5GAyYWAAG2S+RjBgJzhkgDpkADl6ZBk5ZGgoeigycUlS+QBmLwB5AFtgMF1+A2MiOyxHZzA3fg8mlWxeCOiBOISk1PTM7LzC4vryqErq2qwgA):

TypeScript TS Playground - An online editor for exploring TypeScript and JavaScript

jlarmstrongiv:
@lilnasy here is the code snippet!  Though it seems like you were able to reproduce it

@fshafiee here is proof that the types are definitely broken with `typescript@5.4.5`

https://github.com/jlarmstrongiv/astro-cookie-types

ematipico:
Can anyone create an issue so we track it and fix it?

fshafiee:
@jlarmstrongiv I understand that there’s an issue. I just meant to point out that it’s not due to the `Pick` syntax and the problem is probably somewhere else. 😅 

I checked your repro and when I tried to follow the types, I realized that the `cookie` types are in fact unresolved, simply because they are missing from the `dependencies` (listed as `devDependencies` at the moment).

You can fix the issue by `npm install @types/cookie`, and you’d see the type-hints to appear, and the `astro check` will also pass.
Screenshot 2024-04-12 at 12 44 28 PM

@lilnasy @ematipico I'll open a PR and we can continue the conversation there.

PR #10672, opened 2024-04-03
Fix test runner issues by upgrading nodejs version to 18.17.1

fshafiee:
## Changes

Specifies Node.js v18.17.1  as the minimum requirements for contributors, since `node:test/reporters` is not exposed in v18.14.1

## Testing

I was unable to run test using and it kept complaining about `node:test/reporters` not being a built-in module.

Screenshot 2024-04-03 at 7 03 21 PM

After consulting the [Node.js documentation](https://nodejs.org/docs/latest-v18.x/api/test.html#test-reporters), it became apparent that this version is outdated and v18.17.x should be considered as the minimum required version for development purposes.

Screenshot 2024-04-03 at 7 19 34 PM

I switched to v18.17.1 and the tests ran successfully.

## Docs


jacobdalamb:
I'm wondering if this would require updating the required Node.js version for users which would put this PR on hold until Astro 6

bluwy (approved):

bluwy:
No, we only need to bump it for contributors, which isn't a breaking change.

ematipico (approved):

Issue #9946, opened 2024-02-02
Generating a sitemap for [...slug].astro pages with slugs starting with / results in broken URLs


moose96:
### Astro Info

```block
Astro                    v4.3.1
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/sitemap
```

### Describe the Bug

I am creating my own integration with CMS and I've decided to store all slugs inside content management system. Thus, my `pages` folder contains only one page - `[...slug].astro`. I have complicated logic in `getStaticPaths` but it migth be simplified to this:
```ts
export async function getStaticPaths() {
  return [
    {
      params: {
        slug: undefined,
      }
    },
    {
      params: {
        slug: '/blog'
      }
    },
    {
      params: {
        slug: '/test'
      }
    }
  ];
}
```
Let's assume my `astro.config.mjs` looks like:
```js
export default defineConfig({
  site: 'https://example.com',
  integrations: [sitemap()],
});
``` 
In this case I get sitemap:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9"
        xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1"
        xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
    <url>
        <loc>https://blog/</loc> <!-- <- wrong url here -->
    </url>
    <url>
        <loc>https://example.com/</loc>
    </url>
    <url>
        <loc>https://test/</loc><!-- <- wrong url here -->
    </url>
</urlset>
```

### What's the expected result?

I expect sitemap to be:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9"
        xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1"
        xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
    <url>
        <loc>https://example.com/blog/</loc>
    </url>
    <url>
        <loc>https://example.com/</loc>
    </url>
    <url>
[...]

PR #10676, opened 2024-04-04
pathToParam utility

matthewp:
## Changes

- Adds a `pathToParam(path)` utility for `getStaticPaths`. This normalizes paths that want to be converted to params, removing leading and trailing slashes.
- Closes https://github.com/withastro/astro/issues/9946

## Testing

- New tests cases added

## Docs

- https://github.com/withastro/docs/pull/7763


packages/astro/src/core/render/route-cache.ts
```diff
@@ -58,6 +58,21 @@ export async function callGetStaticPaths({
 		// Q: Why the cast?
 		// A: So users downstream can have nicer typings, we have to make some sacrifice in our internal typings, which necessitate a cast here
 		paginate: generatePaginateFunction(route) as PaginateFunction,
+		pathToParam(path: string) {
```
Princesseuh, line 61: No intention to blow up the scope, but do you think users might expect this to do more normalization than it does? Like remove invalid characters or something

matthewp: I wouldn't be surprised but I can't anticipate what those expectations are. Should we not do this until we have a better idea?

bluwy, line 61: Is there a case where one would not want to use `pathToParam`, and pass the string directly? And if they do so, wouldn't they hit issues as described in the linked issue? If we see that the params should always be normalized, maybe we should do that for all params automatically.

ematipico: The assumption is that every param should be a valid URL, that's the same assumption we're using as pillar before discussing any bug that affects these issues e.g. encoding. 

Regarding the _why_, it's exactly why Matthew decided to create this utility in the first place. We can't predict the usage of users, and felt that it's better to provide an API instead of forcing our way around (always regarding the first assumption)

ematipico: I also expected that. I think we should call this function `unstable_pathToParam` and see how it's received, and we figure out what it's meant for (should it encode stuff? I mean, `/folder name` is a valid path but not a valid param because the space should be encoded. 

Having it called `unstable_` (or `experimental_*`) would allow us to opt in to our experimentation contract and break stuff whenever we see fit.

bluwy: If we don't force our way around, or to put it differently, if the original issue author decided that they don't want to use `pathToParam` because that's what they want, wouldn't that not fix the issue?

The root issue in https://github.com/withastro/astro/issues/9946 was that the sitemap integration wasn't constructing the paths right because of the `//`. I'm guessing it's this issue:

```js
new URL('//blog/', 'http://example.com') // -> http://blog/
```

So if we want to allow users to use slashes still and fix the root issue, the sitemap integration would need a fix. And in that case, I'm not sure if there's any other use for `pathToParam`.

matthewp: The more I think about it the more I question this helper as being useful for SSG. The example use-case of `/about//en` is a valid URL, so I wouldn't want to block anyone from using it (most likely as a legacy URL). But in the context of SSG, is that even possible to have. I think we need to do some more investigation.

matthewp:
Closing this PR as I think it's needs more thought and investigation before we pursue. Will keep around the code in case it's needed in the future.

Issue #10668, opened 2024-04-03
Astro DB build fails with Cloudflare adapter


casungo:
### Astro Info

```block
Astro                    v4.5.15
Node                     v20.7.0
System                   Windows (x64)
Package Manager          pnpm
Output                   hybrid
Adapter                  none
Integrations             astro:db
                         @astrojs/db/file-url
```

### Describe the Bug

I am logged into Astro Studio, linked, and able to work with production data using the --remote flag in dev mode.
When attempting to build with the --remote flag, I get the following error:

> [ERROR] [vite] x Build failed in 1.30s
> [commonjs--resolver] [plugin vite:resolve] Cannot bundle Node.js built-in "url" imported from "node_modules\.pnpm\@astrojs+db@0.10.0\node_modules\@astrojs\db\dist\runtime\index.js". Consider disabling ssr.noExternal or remove the built-in dependency.

### What's the expected result?

`astro build --remote` should build using the remote data.

(forked the minimal reproducible example from #10617 since the issue is in the db integration)

(for the minimal reproducible example to work you would need to download the project, login, link and then execute a build with the remote flag)

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-6icpvd-oyrvsw

### Participation

- [ ] I am willing to submit a pull request for this issue.

aklevecz:
I'm having the same issue.

I tried updating my config to include: `  vite: {
    ssr: {
      noExternal: ["@astrojs/db"]
    }
  }`, but it still fails with the same message

aklevecz:
Running into this same issue. Appreciate the urgency!

matthewp:
I'm not seeing the issue here. Will try on my Windows computer to make sure it's not specific to Windows.

Could you try force reinstalling? Maybe something didn't get installed correctly.

aklevecz:
For myself I was able to successfully build and deploy once I drank more coffee and realized that I'm using the Cloudflare adapter, and followed the docs on [Node compatability](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#nodejs-compatibility) to enable the Node `url` library.

My config looks like this now:
```
import { defineConfig } from "astro/config";
import cloudflare from "@astrojs/cloudflare";

import db from "@astrojs/db";

export default defineConfig({
  output: "server",
  adapter: cloudflare({
    platformProxy: {
      enabled: true
    }
  }),
  integrations: [db()],
  vite: {
    ssr: {
      external: ["url"]
    }
[...]

PR #10677, opened 2024-04-04
Fix compatibility for @astrojs/db with cloudflare

matthewp:
## Changes

- Fixes https://github.com/withastro/astro/issues/10668
- We can't use `url` or `node:url` if we want Cloudflare compat.

## Testing

- Not wanting to test Cloudflare directly.
- Existing tests should cover Node compat, both Windows and Unix.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/eleven-eyes-sip.md
+++ b/.changeset/eleven-eyes-sip.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/db": patch
+---
+
+Fix compatibility of @astrojs/db with Cloudflare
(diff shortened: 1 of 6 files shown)


alexanderniebuhr (commented):
The 3rd party package `url-file-url` uses `path` in their source code. So this will change the error during build from `url` to `path`. But it will still be an error for users without further configuration. 

While Cloudflare supports `node:path` imports when enabled the `nodejs_compat` flag, the package still doesn't use `node:` import prefixes.

alexanderniebuhr (approved):
Not tested with a live deployment, but given my CF experience, I'm confident this should work.

PR #10678, opened 2024-04-04
feat: add origin check for CSRF protection

ematipico:
## Changes

This PR adds the **origin header check** for the CSRF protection.

Closes PLT-1872

Here's the RFC for more details: https://github.com/withastro/roadmap/pull/879

This is a feature that is supposed to work only for on-demand pages. It should not work during the build. To enable the feature, I **had** considered some generic way to change the `internalMiddleware` that is present in the `BasePipeline` class, but after having a chat with @lilnasy, we decided that it felt a bit premature and decided to go for a more custom way.

The `manifest` inside the pipeline contains the middleware, and if the feature is enabled, the `App` class (the one that all adapters use) will modify the middleware to add the origin check middleware function.

The logic of the origin check is heavily inspired from SvelteKit, and adapted to match Astro.

## Testing

I added new test cases 

## Docs

I documented new options. I am not sure about the creation of a new page. We can still create a new page later on, considering that the RFC is still open.

### DIFF
--- a/.changeset/fair-jars-behave.md
+++ b/.changeset/fair-jars-behave.md
@@ -2,7 +2,7 @@
 "astro": minor
 ---
 
-Adds a new security - and experimental - option to prevent CSRF attacks. This feature is available only for on-demand pages:
+Adds a new experimental security option to prevent [Cross-Site Request Forgery (CSRF) attacks](https://owasp.org/www-community/attacks/csrf). This feature is available only for pages rendered on demand:
 
 ```js
 import { defineConfig } from "astro/config"
@@ -17,9 +17,8 @@ export default defineConfig({
 })
 ```
 
-When enabled, it checks that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
+Enabling this setting performs a check that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
 
-The "origin" check is executed only on-demand pages, and only for the requests `POST, `PATCH`, `DELETE` and `PUT`, only for those requests that
-the followin `content-type` header: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
+This experimental "origin" check is executed only for pages rendered on demand, and only for the requests `POST, `PATCH`, `DELETE` and `PUT` with one of the following `content-type` headers: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
 
 It the "origin" header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
--- a/.changeset/fair-jars-behave.md
+++ b/.changeset/fair-jars-behave.md
@@ -1,0 +1,26 @@
+---
+"astro": minor
+---
+
+Adds a new security - and experimental - option to prevent CSRF attacks. This feature is available only for on-demand pages:
+
+```js
+import { defineConfig } from "astro/config"
+export default defineConfig({
+  security: {
+    csrfProtection: {
+      origin: true
+    }
+  },
+  experimental: {
+    csrfProtection: true
+  }
+})
+```
+
+When enabled, it checks that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
+
+The "origin" check is executed only on-demand pages, and only for the requests `POST, `PATCH`, `DELETE` and `PUT`, only for those requests that
+the followin `content-type` header: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
+
+It the "origin" header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
--- a/.changeset/fair-jars-behave.md
+++ b/.changeset/fair-jars-behave.md
@@ -7,13 +7,12 @@ Adds a new security - and experimental - option to prevent CSRF attacks. This fe
 ```js
 import { defineConfig } from "astro/config"
 export default defineConfig({
-  security: {
-    csrfProtection: {
-      origin: true
-    }
-  },
   experimental: {
-    csrfProtection: true
+    security: {
+      csrfProtection: {
+        origin: true
+      }
+    }
   }
 })
 ```
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1611,6 +1611,49 @@ export interface AstroUserConfig {
 	legacy?: object;
 
 	/**
+	 * @name security
+	 * @type {object}
+	 * @version 4.6.0
+	 * @type {object}
+	 * @description
+	 *
+	 * It allows to opt-in various security measures for Astro applications.
+	 */
+	security?: {
+		/**
+		 * @name security.csrfProtection
+		 * @type {object}
+		 * @default '{}'
+		 * @version 4.6.0
+		 * @description
+		 */
+
+		csrfProtection?: {
+			/**
+			 * @name security.csrfProtection.origin
+			 * @type {boolean}
+			 * @default 'false'
+			 * @version 4.6.0
+			 * @description
+			 *
+			 * Something
+			 */
+			origin?: boolean;
+
+			/**
+			 * @name security.csrfProtection.token
+			 * @type {boolean}
+			 * @default 'false'
+			 * @version 4.6.0
+			 * @description
+			 *
+			 * Something
+			 */
+			token?: boolean | string;
+		};
+	};
+
+	/**
 	 * @docs
 	 * @kind heading
 	 * @name Experimental Flags
@@ -1821,6 +1864,29 @@ export interface AstroUserConfig {
 		 * See the [Internationalization Guide](https://docs.astro.build/en/guides/internationalization/#domains-experimental) for more details, including the limitations of this experimental feature.
 		 */
 		i18nDomains?: boolean;
+
+		/**
+		 * @docs
+		 * @name experimental.csrfProtection
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.6.0
+		 * @description
+		 *
+		 * It enables the CSRF protection for Astro websites.
+		 *
+		 * The CSRF protection works only for on-demand pages (SSR) and hybrid pages where prerendering is opted out.
+		 *
+		 * ```js
+		 * // astro.config.mjs
+		 * export default defineConfig({
+		 *   experimental: {
+		 *     csrfProtection: true,
+		 *   },
+		 * })
+		 * ```
+		 */
+		csrfProtection?: boolean;
 	};
 }
 
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -1830,9 +1830,9 @@ export interface AstroUserConfig {
 		 * @version 4.6.0
 		 * @description
 		 *
-		 * It enables the CSRF protection for Astro websites.
+		 * Enables CSRF protection for Astro websites.
 		 *
-		 * The CSRF protection works only for on-demand pages (SSR) and hybrid pages where prerendering is opted out.
+		 * The CSRF protection works only for pages rendered on demand (SSR) using `server` or `hybrid` mode. The pages must opt out of prerendering in `hybrid` mode.
 		 *
 		 * ```js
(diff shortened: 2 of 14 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -137,6 +146,42 @@ export class App {
...
+	 * Returns a middleware function in charge to check the `origin` header.
+	 *
+	 * @private
+	 */
+	#createOriginCheckMiddleware(): MiddlewareHandler {
+		return defineMiddleware((context, next) => {
+			const { request, url } = context;
+			const contentType = request.headers.get('content-type');
+			if (contentType) {
+				if (this.#formContentTypes.includes(contentType)) {
```
bluwy, line 167: I think the `content-type` here may not be thoroughly checked. There was a [SvelteKit vulnerability](https://github.com/sveltejs/kit/security/advisories/GHSA-gv7g-x59x-wf8f) where `content-type` value could have different casing, which can bypass the check.

bluwy (commented):
Only one comment below, but others seem simple enough! I feel like it shouldn't need the experimental flag since it's rather straight-forward, but works for me anyways.

.changeset/fair-jars-behave.md
```diff
@@ -0,0 +1,26 @@
...
+```js
+import { defineConfig } from "astro/config"
+export default defineConfig({
+  security: {
+    csrfProtection: {
+      origin: true
+    }
+  },
+  experimental: {
+    csrfProtection: true
```
matthewp, line 16: This sugar version is not explained in the RFC. I'm wondering, should we even have it? I could see arguments that the cookie version is better and should be the default. It might be good to just let them specify which version to use manually.

matthewp: Oh sorry, i this is to enable the protection as an experimental feature. Ignore!

packages/astro/src/core/app/index.ts
matthewp: > Can this be fixed by doing `contentType.toLowerCase()`?

matthewp:
I don't think we need the experimental flag for the `origin` property, but if the plan is to follow up with `cookie` then I think it makes sense to still have the flag, as that will be a little more complicated and have potential for bugs.

matthewp:
Also we have to have experimental flag if the RFC is not merged, and that one is not, yet.

bluwy (approved):

packages/astro/src/core/app/index.ts
```diff
@@ -137,6 +146,42 @@ export class App {
...
+	 * @private
+	 */
+	#formContentTypes = ['application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'];
+
+	/**
+	 * Returns a middleware function in charge to check the `origin` header.
+	 *
+	 * @private
+	 */
+	#createOriginCheckMiddleware(): MiddlewareHandler {
```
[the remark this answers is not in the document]
ematipico: A fair request; it makes sense! :)

sarah11918:
I haven't *reviewed* the docs here yet, but just popping in to say that I think @ematipico 's choice of where to put docs here are perfect. The changeset will have the example, then config-reference will have both the `security` configuration entry and the experimental flag will be documented with the others.

I don't think this calls for an entire new page *yet*, but I will add it to my task list (I'm reworking our sidebar/content structure this quarter) to keep an eye out for "security related things" so we can eventually have a page. Right now, the entry in config will be our documenting, if everyone's OK with that!

packages/astro/src/@types/astro.ts
```diff
@@ -1610,6 +1610,45 @@ export interface AstroUserConfig {
...
+	 */
+	security?: {
+		/**
+		 * @name security.csrfProtection
+		 * @type {object}
+		 * @default '{}'
+		 * @version 4.6.0
+		 * @description
+		 *
+		 * It enables some security measures to prevent CSRF attacks: https://owasp.org/www-community/attacks/csrf
```
sarah11918, line 1630:
```suggestion
		 * It enables some security measures to prevent CSRF attacks: https://owasp.org/www-community/attacks/csrf
```
I think we'll need to provide *some* more detail here. Not heavy implementation details, but like a sentence or so on HOW are you doing this or WHAT MEASURES are you taking, in lay speak. (cookies, tokens, other random words Sarah knows?) This doesn't exactly inspire trust. :sweat_smile: 

Ideally, think about what people will find on web pages as advice of "things they should do" in their project to protect against this, and be able to say, "Astro enables security measures recommended to protect against CSRF attacks, like [thing that people will recognize as stuff they're supposed to want/need/do] -- if this is at all possible!

packages/astro/src/@types/astro.ts
```diff
@@ -1610,6 +1610,45 @@ export interface AstroUserConfig {
...
+
+		csrfProtection?: {
+			/**
+			 * @name security.csrfProtection.origin
+			 * @type {boolean}
+			 * @default 'false'
+			 * @version 4.6.0
+			 * @description
+			 *
+			 * When enabled, it checks that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
```
sarah11918, line 1641:
```suggestion
			 * When enabled, performs a check that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
```
Either something like above, or be more explicit about what "it" actually is. (I think you can do above and it's fine, though.)

packages/astro/src/@types/astro.ts
```diff
@@ -1610,6 +1610,45 @@ export interface AstroUserConfig {
...
+			 *
+			 * The "origin" check is executed only on-demand pages, and only for the requests `POST, `PATCH`, `DELETE` and `PUT`, only for those requests that
+			 * the followin `content-type` header: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
+			 *
+			 * It the "origin" header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
```
sarah11918, line 1646:
```suggestion
			 * The "origin" check is executed only for pages rendered on demand, and only for the requests `POST, `PATCH`, `DELETE` and `PUT` with
			 * the following `content-type` header: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
			 *
			 * If the "origin" header doesn't match the `pathname` of the request, Astro will return a 403 status code and will not render the page.
```
Had to clean up a bit here, so please make sure this is the sentence you intended!

packages/astro/src/@types/astro.ts
```diff
@@ -1821,6 +1860,29 @@ export interface AstroUserConfig {
...
+
+		/**
+		 * @docs
+		 * @name experimental.csrfProtection
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.6.0
+		 * @description
+		 *
+		 * It enables the CSRF protection for Astro websites.
```
sarah11918, line 1872:
```suggestion
		 * Enables CSRF protection for Astro websites.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1821,6 +1860,29 @@ export interface AstroUserConfig {
...
+		 *
+		 * The CSRF protection works only for on-demand pages (SSR) and hybrid pages where prerendering is opted out.
```
sarah11918, line 1874:
```suggestion
		 * The CSRF protection works only for pages rendered on demand (SSR) using `server` or `hybrid` mode. The pages must opt out of prerendering in `hybrid` mode.
```

packages/astro/src/@types/astro.ts
```diff
@@ -1821,6 +1860,29 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
```
sarah11918, line 1876: Is this code sample all that's needed? Or do you have to ALSO enable the `security` described above?

sarah11918 (commented):
The only thing I have not yet reviewed here is the changeset, @ematipico !

First, I want to double check: ONLY `csrfProtection: true` is experimental, but `security` is not?

It's not clear to me which of these things is doing what, nor how they work together. Do I need to enable the experimental flag AND the `security` config? Can I just enable security and not the experimental flag? What extra feature am I getting from the experimental flag in that case?

So, all the individual words seem OK, but I really don't get a good handle on how these pieces fit together. THAT's what the changeset should ideally make clear!

ematipico:
@sarah11918 Thank you for the feedback. Based on your thoughts, I reworked the configuration and now the experimental flags is like this:
```
defineConfig({
	experimental: {
		security: {
			csrfProtection: { origin: true }
		}
	}
})
```

Since we are still debating if it makes sense to have a top-level `security` object, it's best to have everything under the experimental flag to avoid confusion, and to allow us to apply any change during the experimental phase.

sarah11918:
@ematipico This change looks great! I was even wondering whether it might be possible to just enable `security` by default if you opted in to the experimental flag, but you beat me to it! 🫡

packages/astro/src/@types/astro.ts
```diff
@@ -1821,6 +1821,62 @@ export interface AstroUserConfig {
...
+		 */
+		security?: {
+			/**
+			 * @name security.csrfProtection
+			 * @type {object}
+			 * @default '{}'
+			 * @version 4.6.0
+			 * @description
+			 *
+			 * It enables some security measures to prevent CSRF attacks: https://owasp.org/www-community/attacks/csrf
```
sarah11918, line 1859:
```suggestion
			 * Allows you to enable security measures to prevent CSRF attacks: https://owasp.org/www-community/attacks/csrf
```
I'm assuming the plan is for there to be more than just "origin" here, so I think this wording is future-proof for when there are more available options here.

.changeset/fair-jars-behave.md
```diff
@@ -0,0 +1,25 @@
+---
+"astro": minor
+---
+
+Adds a new security - and experimental - option to prevent CSRF attacks. This feature is available only for on-demand pages:
```
sarah11918, line 5:
```suggestion
Adds a new experimental security option to prevent CSRF attacks. This feature is available only for pages rendered on demand:
```

.changeset/fair-jars-behave.md
```diff
@@ -0,0 +1,25 @@
...
+  }
+})
+```
+
+When enabled, it checks that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.
+
+The "origin" check is executed only on-demand pages, and only for the requests `POST, `PATCH`, `DELETE` and `PUT`, only for those requests that
+the followin `content-type` header: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.
+
+It the "origin" header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
```
sarah11918, line 25:
```suggestion
Enabling this setting performs a check that the "origin" header, automatically passed by all modern browsers, matches the URL sent by each `Request`.

This experimental "origin" check is executed only for pages rendered on demand, and only for the requests `POST, `PATCH`, `DELETE` and `PUT` with one of the following `content-type` headers: 'application/x-www-form-urlencoded', 'multipart/form-data', 'text/plain'.

It the "origin" header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
```

sarah11918:
Just left a few tiny nit suggestions, but otherwise, Docs is happy with this one! Great job @ematipico!

packages/astro/src/@types/astro.ts
ematipico: Yes, although this is still under evaluation and discussion. Let's see how this turns out :)

sarah11918 (approved):
Approving for the green check!

matthewp (approved):

septatrix:
This will break form submissions when using `curl` for example or pretty much any other tool which is not a modern browser such as http clients in most programming languages. Another example would be load-testing tools. While some like k6 support setting custom headers this is cumbersome and not supported by everything.

A cookie based implementation would at least work with everything which has a sense of session. So not curl but at least the other stuff I listed.

ematipico:
Hey @septatrix, thank you for your feedback, although we would prefer to channel the discussions in the relative RFC: https://github.com/withastro/roadmap/pull/879

A cookie solution has been pondered, but we are still discussing if it's worth the effort considering how Astro works.

PR #10681, opened 2024-04-04
Prevent errors in finding workspaceId from interrupting link prompts

matthewp:
## Changes

- When running `astro db link` we don't want to exit in the middle of a question, but instead do so when the workspace is needed.

## Testing

Manual testing in example project

## Docs

N/A, bug fix


bholmesdev (approved):
(german accept) I sync zis ist a good PR!

PR #10684, opened 2024-04-04
Update sharp to 0.33 to fix issue with Alpine Linux

PeterDraex:
## Changes

This PR updates the version of the `sharp` dependency to `^0.33` and thus makes Astro compatible with Alpine Linux containers.

It was impossible for me to use the Astro image service in an Alpine Linux docker container. Even though I would install sharp `0.33.3` in my app, pnpm would download version `0.32.6`, which doesn't work in Alpine Linux container. I have to override the downloaded version in my package.json. 

For more details about the issue see:
- https://discord.com/channels/830184174198718474/1224861729792458803/1224861729792458803
- https://github.com/lovell/sharp/issues/4054

## Testing

- Automatic tests
- Override of sharp with the current version in local project

## Docs

This resolves existing bug. No documentation needed.


Princesseuh (changes requested):
Blocking this because it's a breaking change. Sharp's 0.33.0 `engines` contract is stricter than Astro's, requiring a higher version of Node 18 and 20 than Astro, and not supporting Node 19 at all. 

As such, for a certain percentage of users, Sharp would just not get installed ever unless they forced the install. Last time we updated to 0.33.0 (because I didn't know its checks were stricter than us, ha), we ended up with a bunch of support threads from people running outdated versions of Node.

packages/astro/package.json
```diff
@@ -179,7 +179,7 @@
     "zod-to-json-schema": "^3.22.4"
   },
   "optionalDependencies": {
-    "sharp": "^0.32.6"
+    "sharp": "^0.33"
```
Princesseuh, line 182:
```suggestion
    "sharp": "^0.33.3"
```

Princesseuh (approved):
Thank you, this looks great!

pnpm-lock.yaml
```diff
@@ -694,8 +694,8 @@ importers:
         version: 3.22.5(zod@3.22.4)
     optionalDependencies:
       sharp:
-        specifier: ^0.32.6
-        version: 0.32.6
+        specifier: ^0.33
```
Princesseuh, line 697:
```suggestion
        specifier: ^0.33.3
```

ematipico (approved):

.changeset/famous-mugs-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Update sharp to 0.33 to fix issue with Alpine Linux
```
sarah11918, line 5:
```suggestion
Updates sharp to version 0.33 to fix issues when using an Alpine Linux container.
```

sarah11918 (commented):
Thanks for this helpful contribution @PeterDraex !  I've just updated the changeset message to be a tiny bit more explicit. Especially since this is being released with a minor version of Astro, we try to give the changesets a little extra love and attention as people are checking more closely!  :raised_hands:

PeterDraex:
> @PeterDraex would you mind rebasing with `main` and rerunning install to update the lockfile?

@matthewp Done. I've also updated sharp version in `examples/starlog/package.json`. Is that okay?

PR #10688, opened 2024-04-05
Refactor MDX transformJSX handling

bluwy:
## Changes

- Mark renderer `jsxImportSource` and `jsxTransformOptions` as deprecated as they're not used since Astro 3.0 (https://github.com/withastro/astro/pull/7924)
- Refactor `transformJSX` to use the `astro:jsx` transform options directly. This makes `transformJSX` less generic and make it easier to port it to `@astrojs/mdx` in the future.
- Altogether should improve the performance of transforming MDX files.

Before:

image

After:

image

Execution time of `transformJSX` dropped from 1.16s to 0.33s.

## Testing

Existing tests should pass

## Docs

Added a changeset about `jsxImportSource` and `jsxTransformOptions` being deprecated, and in the jsdoc.

### DIFF
--- a/packages/astro/src/jsx/renderer.ts
+++ b/packages/astro/src/jsx/renderer.ts
@@ -1,21 +1,11 @@
 import type { AstroRenderer } from '../@types/astro.js';
+import { jsxTransformOptions } from './transform-options.js';
 
 const renderer: AstroRenderer = {
 	name: 'astro:jsx',
 	serverEntrypoint: 'astro/jsx/server.js',
 	jsxImportSource: 'astro',
-	jsxTransformOptions: async () => {
-		// @ts-expect-error types not found
-		const plugin = await import('@babel/plugin-transform-react-jsx');
-		const jsx = plugin.default?.default ?? plugin.default;
-		const { default: astroJSX } = await import('./babel.js');
-		return {
-			plugins: [
-				astroJSX(),
-				jsx({}, { throwIfNamespace: false, runtime: 'automatic', importSource: 'astro' }),
-			],
-		};
-	},
+	jsxTransformOptions,
 };
 
 export default renderer;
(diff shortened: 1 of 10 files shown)


packages/astro/src/jsx/renderer.ts
```diff
@@ -1,19 +1,11 @@
-const renderer = {
+import type { AstroRenderer } from '../@types/astro.js';
+import { jsxTransformOptions } from './transform-options.js';
+
+const renderer: AstroRenderer = {
 	name: 'astro:jsx',
 	serverEntrypoint: 'astro/jsx/server.js',
 	jsxImportSource: 'astro',
-	jsxTransformOptions: async () => {
```
bluwy, line 5: This function is moved to `./transform-options.js` without changes. This allows using the `jsxTransformOptions` function directly.

packages/astro/src/vite-plugin-mdx/index.ts
```diff
@@ -1,99 +1,19 @@
...
-interface TransformJSXOptions {
-	code: string;
-	id: string;
-	mode: string;
-	renderer: AstroRenderer;
-	ssr: boolean;
-	root: URL;
-}
-
-async function transformJSX({
```
bluwy, line 22: This function is moved to `./transform-jsx.ts` and is updated to retrieve the transform options differently.

packages/astro/src/vite-plugin-mdx/tag.ts
```diff
@@ -9,115 +12,100 @@ import * as t from '@babel/types';
...
-									t.callExpression(t.identifier('__astro_tag_component__'), [
-										t.identifier(id),
-										t.stringLiteral(rendererName),
-									])
-								)
-							);
-						}
-					}
-				},
+export const tagExportsPlugin: PluginObj = {
```
bluwy, line 15: I converted the function to a simple object so the instance can be shared. Nothing else within the function body is changed.

Princesseuh (approved):
Hugo been real quiet since this dropped

PR #10689, opened 2024-04-05
feat: upgrade the minimum Node.js maintainance LTS version

ematipico:
## Changes

This PR updates the minimum Node.js extended LTS version from `v18.14.1` to `v18.20.1`.

This change is in line with our Node.js support contract: 
- support https://docs.astro.build/en/upgrade-astro/#support
- upgrade https://docs.astro.build/en/upgrade-astro/#upgrade

## Testing

CI should pass

## Docs

I will open a PR to update the docs too


package.json
```diff
@@ -42,7 +42,7 @@
     "packages/*"
   ],
   "engines": {
-    "node": ">=18.14.1",
+    "node": ">=18.20.1",
```
Princesseuh, line 45: Should this also drop support for 19 (1.49% of users on 4.x in the last 30 days) and `20.0.0`, `20.1.0` and `20.2.0` (1% of users on 4.x in the last 30 days) so that we can finally upgrade Sharp?

packages/astro/package.json
```diff
@@ -224,7 +224,7 @@
     "unified": "^11.0.4"
   },
   "engines": {
-    "node": ">=18.14.1",
+    "node": ">=18.20.1",
     "npm": ">=6.14.0"
```
Princesseuh, line 228: I realize we forgot to mention this in our documentation, but maybe we should outline our policy for updating package managers versions as well. I'm actually not quite sure Astro really installs on npm 6.14.0, and last I heard, some deps don't install correctly on Yarn 1.

(For context, Sharp requires `>=9.6.5` for `npm`, `>=3.2.0` for Yarn and `>=7.1.0` for `pnpm`, probably due to bugs with installing native deps.)

ematipico: You're making a good point. If we look at the [Node.js document](https://nodejs.org/en/about/previous-releases#release-schedule), the npm minimum version shipped by the maintenance LTS is `v10.5.0`, so I think it's perfectly safe to bump the `npm` version..

package.json
ematipico: > Should this also drop support for 19 (1.49% of users on 4.x in the last 30 days) and 20.0.0, 20.1.0 and 20.2.0 (1% of users on 4.x in the last 30 days) so that we can finally upgrade Sharp?

I think dropping support for `19` is fine and in line with our Node.js support policy. I am not sure about `20`, but if we consider that we never set a restriction, I think we should be careful about this change. Personally, I would just set `>=20`

Princesseuh: 20.3.0 added support for node-api v9, which is required for not only Sharp, but also other common native binaries (it was equivalent to 18.17.0, came out at the same time)

packages/astro/package.json
```diff
@@ -224,8 +224,8 @@
     "unified": "^11.0.4"
   },
   "engines": {
-    "node": ">=18.14.1",
-    "npm": ">=6.14.0"
+    "node": ">=18.20.1",
+    "npm": ">=10.5.0"
```
Princesseuh, line 228: Only 18% of our users are compatible with this change 😅

ematipico: I'll have to understand the logic of that... if Node.js whips `18` ships npm `10.5.0`, how can users use a lower version of npm? 😮

Princesseuh: On certain systems they don't ship together, you download (and upgrade) npm separately. In addition to that, it's the *latest* of Node 18 that ships with 10.5.0, the vast vast vast majority of people are not on the latest of 18, they're a version or two behind (reminder: last version of Node 18 came out 2 days ago)

ematipico: Thank you. I reverted the change

package.json
Princesseuh:
```suggestion
    "node": "^18.17.1 || ^20.3.0 || >=21.0.0",
    "npm": ">=9.6.5",
    "pnpm": ">=7.1.0"
```

I know this isn't the latest of 18, however the vast majority of our users are not on it (like, 98% of Node 18 users are not on it). With 17 we cover the 3 most used versions of Node 18 at this time for Astro 4.x users and make ourselves compatible with Sharp - addition of being compatible with our policy. We can raise again later on once needed / people are a bit more upgraded

Dropping 19 is part of our policy, and almost no one is on it (around 1%).

For Node 20, I know this isn't part of our policy, but in my opinion, this is safe to do because virtually no one is on those versions (it's less than 1% of users), I doubt Astro runs perfectly on those versions anyway and it makes us compatible with Sharp.

All together, these changes should affect around 9% of users, 99.99% (quite literally 20 users are on incompatible versions in CI, not sure how, might even be just a wrong detection of CI) of which are in dev.

For `npm`, older versions of `npm` already cannot install Astro correctly due to native dependencies or just modern packages, 91% of our npm users already matches this.

For `pnpm`, PNPM users are really good at keeping themselves updated, less than 1% of pnpm users don't match this.

I decided against putting a restriction for yarn just yet, but we'll add documentation to the Sharp error page about what to do for Yarn 1.

ematipico: I addressed the comment in [`d5df786` (#10689)](https://github.com/withastro/astro/pull/10689/commits/d5df7860d54f05d2e58c074f9c17c6ecbebb9a41)

bluwy (approved):

.changeset/empty-rules-type.md
```diff
@@ -0,0 +1,17 @@
...
+"@astrojs/mdx": minor
+"@astrojs/vue": minor
+"create-astro": minor
+"@astrojs/prism": minor
+"@astrojs/telemetry": minor
+"@astrojs/upgrade": minor
+"astro": minor
+---
+
+Upgrades the minimum version of Node.js to `v18.20.1`. This change is in line with [Node.js support policy](https://docs.astro.build/en/upgrade-astro/#support)
```
sarah11918, line 16:
```suggestion
Upgrades the minimum version of Node.js to `v18.20.1`. This change is in line with Astro's [Node.js support policy](https://docs.astro.build/en/upgrade-astro/#support).
```

sarah11918 (approved):
Just changeset nits. Otherwise, good to go! Thank you @ematipico

matthewp (approved):

PR #10693, opened 2024-04-05
Add disableremoteplayback attribute to VideoHTMLAttributes interface

apetta:
## Changes

Addition of `disableremoteplayback` to the `VideoHTMLAttributes` interface for TypeScript support.
No code changes needed as it's already handled by the regex in `packages/astro/src/runtime/server/render/util.ts`

## Testing

N/A - Type change only

## Docs

N/A - Type change only

### DIFF
--- a/packages/astro/astro-jsx.d.ts
+++ b/packages/astro/astro-jsx.d.ts
@@ -873,7 +873,8 @@ declare namespace astroHTML.JSX {
 		playsinline?: boolean | string | undefined | null;
 		preload?: string | undefined | null;
 		src?: string | undefined | null;
-		disableremoteplayback?: boolean | string | undefined | null;
+		// https://www.w3.org/TR/remote-playback/#the-disableremoteplayback-attribute
+		disableRemotePlayback?: boolean | string | undefined | null;
 	}
 
 	interface MetaHTMLAttributes extends HTMLAttributes {
--- a/packages/astro/astro-jsx.d.ts
+++ b/packages/astro/astro-jsx.d.ts
@@ -1057,6 +1057,7 @@ declare namespace astroHTML.JSX {
 		poster?: string | undefined | null;
 		width?: number | string | undefined | null;
 		disablepictureinpicture?: boolean | string | undefined | null;
+		disableremoteplayback?: boolean | string | undefined | null;
 	}
 
 	// this list is "complete" in that it contains every SVG attribute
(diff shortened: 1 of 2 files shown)


packages/astro/astro-jsx.d.ts
```diff
@@ -1057,6 +1057,7 @@ declare namespace astroHTML.JSX {
 		poster?: string | undefined | null;
 		width?: number | string | undefined | null;
 		disablepictureinpicture?: boolean | string | undefined | null;
+		disableremoteplayback?: boolean | string | undefined | null;
```
Princesseuh, line 1060: I think this should be on the media interface instead, as audio can also use this

Princesseuh:
Do you have a reference for this attribute existing? I can only find reference to the `disableRemotePlayback` property on JavaScript HTMLMediaElement, but not to the corresponding HTML attribute.

apetta:
> Do you have a reference for this attribute existing? I can only find reference to the `disableRemotePlayback` property on JavaScript HTMLMediaElement, but not to the corresponding HTML attribute.

Yes here's one from [mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/disableRemotePlayback).

I use it fairly often to get rid of the cast/chromecast button on videos.

apetta:
@Princesseuh You're right it's only a JS property. 

Is there a better way to handle the typing here given that?

Princesseuh:
Closing since it isn't an attribute, feel free to re-open if it does become one, though!

apetta:
@Princesseuh Actually, I found a [reference](https://www.w3.org/TR/remote-playback/#the-disableremoteplayback-attribute). It's in draft _but stable_ status as of March 2024.

Princesseuh:
> @Princesseuh Actually, I found a [reference](https://www.w3.org/TR/remote-playback/#the-disableremoteplayback-attribute). It's in draft _but stable_ status as of March 2024.
> 
> Should I re-open? Seems like a harmless but useful addition given the stability

Ah, interesting! It's weird that it's `disableRemotePlayback` and not `disableremoteplayback`, HTML attributes are typically not in camel case 🤔 That works for me though, at least we have a document to refer to!

Princesseuh (approved):
Thank you! A rare attribute, but it apparently does exists!

apetta:
> > @Princesseuh Actually, I found a [reference](https://www.w3.org/TR/remote-playback/#the-disableremoteplayback-attribute). It's in draft _but stable_ status as of March 2024.
> > Should I re-open? Seems like a harmless but useful addition given the stability
> 
> Ah, interesting! It's weird that it's `disableRemotePlayback` and not `disableremoteplayback`, HTML attributes are typically not in camel case 🤔 That works for me though, at least we have a document to refer to!

Awesome! I believe html attributes are not case sensitive, but yes a strange choice

PR #10694, opened 2024-04-05
test(@astrojs/node) wait for server listening in trailing-slash tests

rluvaton:
## Changes

Fix flaky trailing slash tests when the server is not ready by the time the test run


rluvaton:
thank you, can we merge now?

Princesseuh (approved):

Issue #10485, opened 2024-03-19
Out of memory during build


agamm:
### Astro Info

```block
Astro                    v4.5.6
Node                     v18.18.2
System                   Linux (x64)
Package Manager          unknown (edit: pnpm)
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

I have a website with > 100K markdown files I need statically built.

When I played with testing the limits of astro I got:
```
01:49:57 [build] Building static entrypoints...

<--- Last few GCs --->

[10615:0x71cf750]    39539 ms: Scavenge 4002.0 (4127.2) -> 4002.8 (4144.0) MB, 911.3 / 0.0 ms  (average mu = 0.715, current mu = 0.134) allocation failure; 
[10615:0x71cf750]    42292 ms: Mark-sweep 4002.9 (4144.0) -> 4000.1 (4156.7) MB, 2752.1 / 0.0 ms  (average mu = 0.533, current mu = 0.251) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb85bc0 node::Abort() [node]
 2: 0xa94834  [node]
 3: 0xd66d10 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd670b7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf447c5  [node]
 6: 0xf56cad v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 7: 0xf313ae v8::internal::HeapAllocator::AllocateRawWithLightRetrySlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 8: 0xf32777 v8::internal::HeapAllocator::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 9: 0xf1394a v8::internal::Factory::NewFillerObject(int, v8::internal::AllocationAlignment, v8::internal::AllocationType, v8::internal::AllocationOrigin) [node]
[...]

PR #10695, opened 2024-04-05
Skip prerender chunk in static output

bluwy:
## Changes

close https://github.com/withastro/astro/issues/10485 (This PR helps with the issue's repro, but fundamentally rendering >100k is not possible with Rollup)

This PR skips a prerender chunk edge case where if you put too many modules within a single chunk, Rollup has a hard time deconflicting the variable names. For example, all Markdown modules export `compiledContent`/`frontmatter`. Those exist as top-level variables in the prerender chunk, and Rollup has to deconflict the variables which proportionately gets slower when iterating over the variables.

The main bottleneck is [`getSafeName` in Rollup](https://github.com/rollup/rollup/blob/459bfe1b336814199b5c643e212390f6979de83d/src/utils/safeName.ts#L4-L16), which would be nice to fix directly, but I made this different fix within Astro for now. 

Before:

image

After:

image

(The execution time for `getSafeName` dropped from 36.7s to 0.033s)

For the astro docs, there's no significant time change as `getSafeName` wasn't the bottleneck)

## Testing

Existing tests should pass.

## Docs

n/a. internal change.

### DIFF
--- a/.changeset/plenty-bugs-hunt.md
+++ b/.changeset/plenty-bugs-hunt.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Skips prerender chunk if building with static output
--- a/packages/astro/src/core/build/plugins/plugin-prerender.ts
+++ b/packages/astro/src/core/build/plugins/plugin-prerender.ts
@@ -68,7 +68,6 @@ function vitePluginPrerender(opts: StaticBuildOptions, internals: BuildInternals
 								}
 							}
 
-							opts.allPages;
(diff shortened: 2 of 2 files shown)


Fryuni (approved):
4 line change improving performance by a factor of 1000.

Looks AWESOME to me!

agamm:
Will we have to do anything for this to take effect or is it automatic?

bluwy:
As long as you use [`output: "static"`](https://docs.astro.build/en/reference/configuration-reference/#output), this will work automatically.

PR #10705, opened 2024-04-06
fix: MDX cannot find relative image path without leading `./`

### LINKED ISSUE #10641: MDX cannot find relative image path without leading `./`
### Astro Info

```block
Node                     v21.6.1
Package Manager          pnpm
Adapter                  none
                         @astrojs/sitemap
                         @astrojs/mdx
```

_No response_

### Describe the Bug

```
13:05:42 [ERROR] Cannot find module 'banner.png' imported from '/Users/tison/Brittani/nightsky-blog/src/content/posts/greptimedb-community-report/index.mdx'
  Stack trace:
    at /Users/tison/Brittani/nightsky-blog/src/content/posts/greptimedb-community-report/index.mdx?astroPropagatedAssets
    [...] See full stack trace in the browser, or rerun with --verbose.
```

```
![GitHub Banner](./banner.png) works
![GitHub Banner](banner.png) failed
```

Both compiled

https://github.com/tisonkun/repro-astro-gldfcz

### Participation

rishi-raj-jain:
## Changes

fixes #10641

- What does this change?

Ensures prefix if not present in the referenced image.

## Testing

- Added a test that optimizes a png and checks for presence of webp in the built project.

## Docs

https://github.com/withastro/docs/pull/7856


OliverSpeir (approved):
Beautiful

OliverSpeir:
I think we called this a minor when we did it for .md https://github.com/withastro/astro/pull/9755 

and I guess https://semver.org/ specifies that when you add functionality it becomes a minor 
> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes

What do you think @rishi-raj-jain?

.changeset/perfect-ties-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/mdx": patch
+---
+
+ensure relative imports if not present in referenced images
```
OliverSpeir, line 5:
```suggestion
Fixes an issue where images in MDX required a relative specifier (e.g. `./`)

Now, you can use the standard `![](img.png)` syntax in MDX files for images colocated in the same folder: no relative specifier required!

There is no need to update your project; your existing images will still continue to work. However, you may wish to remove any relative specifiers from these MDX images as they are no longer necessary:

```diff
- ![A cute dog](./dog.jpg)
+ ![A cute dog](dog.jpg)
<!-- This dog lives in the same folder as my article! -->
```
```

**Comment by rishi-raj-jain** [comment]
@OliverSpeir 

Appreciate the thorough review. Does make sense.

> add functionality it becomes a minor

I think I still need to learn where to mark it as a minor / patch as sometimes I just end up confusing between them. The PR can be seen as fixing a bug because it's "expected" to be able to use the paths without slashes, or be seen as an added functionality from the creators perspective.

**Review by ematipico — CHANGES_REQUESTED** [review]
> The PR can be seen as fixing a bug because it's "expected"

A good way to understand that is if the docs mention this feature. If not, than it's not a bug. In fact, why don't we update the [docs](https://docs.astro.build/en/guides/images/#images-in-mdx-files) to show the new feature? 

@rishi-raj-jain I ask you the following tasks, if you don't mind:
- update the test section of the template, and tell us how you tested the feature. Ideally, we want to add a new test
- update the docs section of the template, and link the docs PR where we add a new example

File: packages/integrations/mdx/src/remark-images-to-component.ts [hunk 2]
```diff
@@ -20,7 +20,7 @@ export function remarkImageToComponent() {
 			// checking if an image should be imported or not
 			if (file.data.imagePaths?.has(node.url)) {
 				let importName = importedImages.get(node.url);
-
+				const prefixedNodeUrl = (node.url.includes('/') ? '' : './') + node.url;
```
bluwy, line 23: This doesn't work when accessing files relatively like `foo/img.png`. I think the correct fix is to do a secondary `this.resolve` at https://github.com/withastro/astro/blob/799f6f3f29a3ef4f76347870a209ffa89651adfa/packages/astro/src/vite-plugin-markdown/index.ts#L76-L84 if it doesn't start with a `/`, which is kinda unfortunate for perf, but I suppose it's not something users would hit often.

rishi-raj-jain: How about if it does not start with a / or ./, we append a `./` by default? That doesn't require use to resolve I think?

rishi-raj-jain: Oh wait in case of `foo/img.png` it looks for it in the present directory of the markdown?

bluwy: If it's `foo/img.png`, it's synonymous to `./foo/img.png`. We can't add `./` by default because users can do `@/img.png` and expect the alias to resolve correctly.

rishi-raj-jain: Right that's another case I didn't think of. Alright, I'll push the resolve thing. Makes sense. Thank you.

rishi-raj-jain: @bluwy,

This works as expected with prefixedNodeUrl as verified using tests.

rishi-raj-jain:
> update the docs section of the template, and link the docs PR where we add a new example

https://github.com/withastro/docs/pull/7856

Issue #9306, opened 2023-12-05
`getCollection()` method returning referance of the cached collection


uksarkar:
### Astro Info

```block
Astro                    v3.6.4
Node                     v20.10.0
System                   Linux (x64)
Package Manager          yarn
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/mdx
                         @astrojs/sitemap
                         @astrojs/tailwind
```

### Describe the Bug

In production, the `getCollection()` method caches the entries, when the first call to the function happens, it's caching and returns the same array reference from the caching logic. As a result, if we do some modifications like `.slice()` on the returned array it's impacting the cached collection.

### What's the expected result?

It should return a deep copy of the cached collection array.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-15q7pu?file=src%2Fpages%2Findex.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
Thanks for opening this issue! This is a limitation that come up a few times before.

Is there a case where mutation can't be avoided?

uksarkar:
> Thanks for opening this issue! This is a limitation that come up a few times before.
> 
> Is there a case where mutation can't be avoided?

Well, we can avoid mutation but it's a bit confusing. Like in the first block you can see it's returning a copy of the array but why not from the fallback one?

matthewp:
This is no longer the case, see https://github.com/withastro/astro/blob/1e33cd85ad42c978404e9e3b33f588a2e9c4a5ea/packages/astro/src/content/runtime.ts#L83

PR #10709, opened 2024-04-07
fix: Prod build collections cache can be modified

### LINKED ISSUE #10700: Prod build collections cache can be modified
### Astro Info

```block
Astro                    v4.5.16
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When building for production, the first time a collection is loaded it can be manipulated in a way that affects the cache and results in the manipulations carrying over to subsequent calls to getCollection. This doesn't happen when running or building dev.

The cached collection should not be affected by operations performed after loading it.
I believe I found the cause is simply line 113 here returning the original array rather than a copy like it does when the cache is used on line 81:
https://github.com/withastro/astro/blob/8214bacfeaf478ec8ff57a92655963441f3f45e9/packages/astro/src/content/runtime.ts#L79-L114

https://github.com/SStembridge/astro-collections-cache-prod-build

### Participation

rishi-raj-jain:
fixes #10700 

## Changes

- What does this change?

## Docs

I do not think docs are required as this is a patch for the existing functionality?


florian-lefebvre:
Do you think you can tackle #9306 as well? Seems similar

rishi-raj-jain:
@florian-lefebvre

I think this PR addresses that as well (already) as it's not returning a reference anymore. What do you think?

ematipico:
Can we have a changeset? Also, can you update the template with the relevant information on how you tested the fix?

packages/astro/src/content/runtime.ts
```diff
@@ -107,6 +104,8 @@ export function createGetCollection({
 			);
 			cacheEntriesByCollection.set(collection, entries);
 		}
+		// Always return a new instance so consumers can safely mutate it
+		entries = [...cacheEntriesByCollection.get(collection)!];
```
matthewp, line 108: This can be expensive for large collections. Given this is corner case, can we use `Object.freeze` when we set it instead? That would have similar effect (can't push, splice, etc) but without doing it on ever request.

matthewp:
@rishi-raj-jain do you still plan on working on this?

rishi-raj-jain:
@matthewp 

Hey yeah, let me get on to this by tomorrow.

rishi-raj-jain:
I am unable to locate the tests for content collection, may you help me with that, @ematipico?

packages/astro/src/content/runtime.ts
rishi-raj-jain: - Slice method https://measurethat.net/Benchmarks/Show/2667/0/arrayprototypeslice-vs-spread-operator#latest_results_block seems to be much faster.
- Can not freeze the cached collection as individual pages can still sort the collection based on it's attributes as done in https://github.com/withastro/astro/packages/astro/test/content-collections-render.test.js fixture: https://github.com/withastro/astro/blob/main/packages/astro/test/fixtures/content/src/pages/sort-blog-collection.astro#L7.

ematipico:
There's a `content-collections.test.js` file. You could use that one. We just need to create a new test case, and make sure it's fixed by this PR

rishi-raj-jain:
> There's a `content-collections.test.js` file. You could use that one. We just need to create a new test case, and make sure it's fixed by this PR

Yup, the tests already cover this as recorded in my observation at https://github.com/withastro/astro/pull/10709#discussion_r1576908837

ematipico:
> > There's a `content-collections.test.js` file. You could use that one. We just need to create a new test case, and make sure it's fixed by this PR
> 
> Yup, the tests already cover this as recorded in my observation at [#10709 (comment)](https://github.com/withastro/astro/pull/10709#discussion_r1576908837)

I am not sure I understand. We want to make sure that the original content collection isn't modified and that test case doesn't do that.

packages/astro/src/content/runtime.ts
ematipico: > Can not freeze the cached collection as individual pages can still expect to sort the collection

Where did you read this? I just tried to use `Object.freeze`, and it works

rishi-raj-jain:
> > > There's a `content-collections.test.js` file. You could use that one. We just need to create a new test case, and make sure it's fixed by this PR
> > 
> > 
> > Yup, the tests already cover this as recorded in my observation at [#10709 (comment)](https://github.com/withastro/astro/pull/10709#discussion_r1576908837)
> 
> I am not sure I understand. We want to make sure that the original content collection isn't modified and that test case doesn't do that.

I aimed to prevent modifications to the cached collection, but I think my understanding is not correct. I'm going to close this so that the issue doesn't end up being stalled due to me.

PR #10719, opened 2024-04-08
fix(devtool): do not trigger interaction check for `div` and `span`

ematipico:
## Changes

This PR fixes the interaction check by bypassing `div` and `span` elements, which don't have a real interaction role.

## Testing

I added a new test

## Docs

N/A

### DIFF
--- a/.changeset/swift-coats-teach.md
+++ b/.changeset/swift-coats-teach.md
@@ -2,9 +2,9 @@
 "astro": patch
 ---
 
-Fixes a false positive for the elements `div` and `span`. `div` and `span` are special elements that don't have an interaction assigned, instead it is assigned based on the role assigned via attributes.
+Fixes a false positive for `div` and `span` elements when running the Dev Toolbar accessibility audits.
 
-This means that cases like the following are deemed correct by the a11y audits:
+Those are special elements that don't have an interaction assigned by default. Instead, it is assigned through the `role` attribute. This means that cases like the following are now deemed correct:
 
 ```html
 <div role="tablist"></div>
--- a/.changeset/swift-coats-teach.md
+++ b/.changeset/swift-coats-teach.md
@@ -2,4 +2,11 @@
 "astro": patch
 ---
 
-The devtool bar doesn't report `div` and `span` when check their role. `div` and `span` are special elements that don't have an interaction assigned, instead it is assigned based on the role assigned via attributes.
+Fixes a false positive for the elements `div` and `span`. `div` and `span` are special elements that don't have an interaction assigned, instead it is assigned based on the role assigned via attributes.
+
+This means that cases like the following are deemed correct by the a11y audits:
+
+```html
+<div role="tablist"></div>
+<span role="button" onclick="" onkeydown=""></span>
+```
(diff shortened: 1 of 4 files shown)


matthewp:
I think this sounds right but would be good for @Princesseuh and @yanthomasdev to confirm.

.changeset/swift-coats-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+The devtool bar doesn't report `div` and `span` when check their role. `div` and `span` are special elements that don't have an interaction assigned, instead it is assigned based on the role assigned via attributes.
```
florian-lefebvre, line 5: It's unclear to me what the PR does from the changeset only. Is that the toolbar **won't** report from now on?

ematipico: Yeah exactly. Let me reword the changeset by adding some examples, then please review it and suggest possible changes.

ematipico: Done in [`782940e` (#10719)](https://github.com/withastro/astro/pull/10719/commits/782940eb29eedac822bc8857bd5ad3fab65198ac)

.changeset/swift-coats-teach.md
```diff
@@ -0,0 +1,12 @@
+---
+"astro": patch
+---
+
+Fixes a false positive for the elements `div` and `span`. `div` and `span` are special elements that don't have an interaction assigned, instead it is assigned based on the role assigned via attributes.
+
+This means that cases like the following are deemed correct by the a11y audits:
```
florian-lefebvre, line 7:
```suggestion
Fixes a false positive for `div` and `span` elements when running the Dev Toolbar accessibility audits.

Those are special elements that don't have an interaction assigned by default. Instead it is assigned through the `role` attribute. This means that cases like the following are now deemed correct:
```

packages/astro/src/runtime/client/dev-toolbar/apps/audit/rules/a11y.ts
```diff
@@ -107,6 +107,9 @@ const aria_non_interactive_roles = [
 	'tooltip',
 ];
 
+// These elements aren't interactive and aren't non-interactive. Their interaction changes based on the role assigned to them
+const roleless_elements = ['div', 'span'];
```
Princesseuh, line 111: Is there a documentation we can link to here?

ematipico: Done in [`69dc9a8` (#10719)](https://github.com/withastro/astro/pull/10719/commits/69dc9a8325328825dfe14390003455babef3e06b)

Princesseuh: Awesome, thank you

Princesseuh (approved):

yanthomasdev (approved):
LGTM.

Both `<div>` and `<span>` receive the [`generic` role](https://www.w3.org/TR/wai-aria-1.2/#generic) that basically has no semantic meaning and is meant for implementators only. Both these elements according to [ARIA's conformance](https://www.w3.org/TR/html-aria/#docconformance) can assume [any role](https://www.w3.org/TR/html-aria/#dfn-any-role), therefore they can be interactive depending on the role associated.

Issue #10712, opened 2024-04-07
Cannot seed Astro DB when seed file is under `srcDir`


kidonng:
### Astro Info

```block
Astro                    v4.5.16
Node                     v18.18.0
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             astro:db
                         @astrojs/db/file-url
```

### Describe the Bug

I prefer to set `srcDir: '.'` to avoid an extra directory level. However, when using Astro DB, seeding will fail with the following error:

```
00:13:11 [ERROR] [astro:db] Cannot read properties of undefined (reading 'Symbol(drizzle:Columns)')
```

As mentioned in https://github.com/withastro/astro/issues/10458#issuecomment-2004571015, if I unset `srcDir` and move the seed file (`seed.ts`, along with the entire `db` folder) outside of `src`, seeding will succeed.

I digged in the code and found if I change it to return `resolved.virtual` even if the seed file is inside `srcDir`, the seeding will succeed as well, without "recursive seed calls" as mentioned in the comments. This may have unexpected consequences though since I'm not familiar with the internals.

https://github.com/withastro/astro/blob/ecb44351e98cfba575447699689b0ace49cd3c9d/packages/db/src/core/integration/vite-plugin-db.ts#L57-L62

### What's the expected result?

Seeding should success whether the seed file is under `srcDir` or not.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-tx22uf

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #10720, opened 2024-04-08
Fix db seeding when srcDir is root

matthewp:
## Changes

- When checking if the import for `astro:db` comes within `src/`, also check that it's not within `db/`, as it's possible for db to be inside src (when src is configured to root).
- Fixes https://github.com/withastro/astro/issues/10712

## Testing

- New test added

## Docs

Bug fix


bholmesdev (approved):
Guess you found the `src` of our problems!

bluwy (approved):

PR #10732, opened 2024-04-09
fix: use assetsDir in creating vite config

### LINKED ISSUE #10626: Worker scripts not output in correct assets directory
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

The Astro build is picking up the Worker script file and transforming it and placing it in the `dist` directory during builds, but it is always placed in `dist/assets` regardless of what the `build.assets` configuration is set to. 

I would expect the worker script to be output to the same directory as other client scripts as specified by the `build.assets` property in the configuration.

https://stackblitz.com/edit/github-knx6rb?file=src%2Fpages%2Findex.astro

### Participation

rishi-raj-jain:
## Changes

fixes #10626 

Sets `assetsDir` (if) `assets` property is passed by user in their Astro configuration.

## Testing

Added tests to match the `custom_dir_1` absolute path to match with the generated directory's `custom_dir_1`.

## Docs

It's a bug so I do not think this needs to be updated, as it now behaves as it was expected to.


packages/astro/src/core/create-vite.ts
```diff
@@ -239,6 +239,8 @@ export async function createVite(
 	// We also need to filter out the plugins that are not meant to be applied to the current command:
 	// - If the command is `build`, we filter out the plugins that are meant to be applied for `serve`.
 	// - If the command is `dev`, we filter out the plugins that are meant to be applied for `build`.
+	const assetsDir = settings.config.build.assets;
+	if (assetsDir) result = vite.mergeConfig(result, { build: { assetsDir } });
```
bluwy, line 243: Is there a reason why we only set when `build.assets` is specified? Shouldn't we always pass it?

.changeset/afraid-laws-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+set build.assets directory during vite config setup
```
bluwy, line 5: Do you see some potential behaviour changes that this is marked a minor? If everything else still works as expected, this seems like a patch / bug fix to me.

rishi-raj-jain: Oh right, per https://github.com/withastro/astro/pull/10705#pullrequestreview-1986028561 so the docs mention it it's a bug fix. Fixing.

rishi-raj-jain:
Ready for your review, @bluwy.

.changeset/afraid-laws-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+set build.assets directory during vite config setup
```
ematipico, line 5:
```suggestion
Correctly sets `build.assets` directory during `vite` config setup
```

packages/astro/test/astro-assets-dir.test.js
```diff
@@ -0,0 +1,36 @@
...
+						'astro:build:done': ({ dir }) => {
+							checkDir = dir;
+						},
+					},
+				},
+			],
+		});
+		await fixture.build();
+	});
+	it('generates the assets directory per the build.assets configuration', async () => {
```
ematipico, line 27:
```suggestion
	it('generates the assets directory as per build.assets configuration', async () => {
```

ematipico (approved):

bluwy (approved):

PR #10734, opened 2024-04-09
feat: add a new message telling the user that a new version of Astro is available

Princesseuh:
## Changes

Check for update on startup of the dev command. This will appear in both the console and the dev toolbar's home app and can be disabled if unwanted. 

There's a few heuristics to make it less annoying:
- Only once every 10 days
- Only if you're more than 5 patches behind

We can make this more or less strict, it's two constants.

image

image

## Testing

Tested manually

## Docs

There's some docs in there!


matthewp:
Can you add screenshots of both?

bluwy:
Hmm personally I'm not a fan of tools that yap about new versions (turbo, pnpm) as I'd upgrade dependencies myself when I have a reason to, but maybe that's just me 😅 (dev toolbar only is fine though)

Princesseuh:
> Hmm personally I'm not a fan of tools that yap about new versions (turbo, pnpm) as I'd upgrade dependencies myself when I have a reason to, but maybe that's just me 😅 (dev toolbar only is fine though)

I can agree! Though I think the message here isn't quite as visually bold as pnpm's or turbo's, so it doesn't bother me as much personally. I also cannot deny that `pnpm` is the most updated package manager in our stats, ha.

ematipico (approved):

matthewp:
Core discussed async and this is the current status:

- Fewer options, probably getting rid of the option to check the semver range and to just check latest always.
- Desire to not show the message all of the time. Need to investigate some other commands here.
- Don't want the toolbar notification to be shown.
- Unsure if there is consensus on the devtoolbar part at all, need to follow-up async.

packages/astro/src/cli/install-package.ts
```diff
@@ -161,3 +162,56 @@ async function installPackage(
 		return false;
 	}
 }
+
+export async function fetchPackageJson(
```
Princesseuh, line 166: Fished this out of `astro add`, code is unchanged. `astro add` is insanely big so fishing things out of it is honestly just good.

packages/astro/src/cli/install-package.ts
```diff
@@ -161,3 +162,56 @@ async function installPackage(
...
+	name: string,
+	tag: string
+): Promise<Record<string, any> | Error> {
+	const packageName = `${scope ? `${scope}/` : ''}${name}`;
+	const registry = await getRegistry();
+	const res = await fetch(`${registry}/${packageName}/${tag}`);
+	if (res.status >= 200 && res.status < 300) {
+		return await res.json();
+	} else if (res.status === 404) {
+		// 404 means the package doesn't exist, so we don't need an error message here
```
Princesseuh, line 177: This is a quite odd pattern that `astro add` does everywhere where functions return errors, kinda neat

packages/astro/src/core/dev/update-check.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	const version = packageJson?.version;
+
+	if (!version) {
+		throw new Error('Failed to fetch latest Astro version');
+	}
+
+	if (preferences) {
+		await preferences.set('checkUpdates._lastCheck', Date.now(), { reloadServer: false });
```
Princesseuh, line 28: So, I had a choice here:
- I add one more file
- I re-use one of the files we have

To me, it didn't feel really wrong to use preferences for this, since users are not supposed to change it by hand. So I went with putting it in preferences, that way I don't have to essentially redo preferences again, but let me know if I should add a third store somewhere for this kind of data, happy to do either (and can be changed any time anyway, not like it's breaking)

packages/astro/src/core/dev/update-check.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	const version = packageJson?.version;
+
+	if (!version) {
+		throw new Error('Failed to fetch latest Astro version');
+	}
+
+	if (preferences) {
+		await preferences.set('_variables.lastUpdateCheck', Date.now(), { reloadServer: false });
```
Princesseuh, line 28: So, I had a choice to make here:
- Use a new file, a new system, etc to store variables like this
- Re-use preferences under a private namespace

I went with the later when I realized adding a new store with different values to preferences would require change a lot of type utils and stuff. I think it's not a problem... You're not supposed to change the config file manually, and the CLI won't show this private namespace.

packages/astro/src/core/messages.ts
```diff
@@ -104,6 +104,13 @@ export function serverShortcuts({ key, label }: { key: string; label: string }):
 	return [dim('  Press'), key, dim('to'), label].join(' ');
 }
 
+export function newVersionAvailable({ latestVersion }: { latestVersion: string }) {
+	const badge = bgYellow(black(` update `));
+	const headline = yellow(`▶ New version of Astro available: ${latestVersion}`);
+	const details = `  Run ${cyan('npx @astrojs/upgrade')} to update`;
```
florian-lefebvre, line 110: Is it possible to update the command based on the package manager? eg `pnpm dlx @astrojs/upgrade` for pnpm

alexanderniebuhr:
Quick comment from my side, would the update checker fail for users who don't have `npm` installed? _It sounds like yes based on the description of this PR `Check for update using npm at launch`_
I explicitly set my `node` install to not install the bundled version of `npm`.

Princesseuh:
> Quick comment from my side, would the update checker fail for users who don't have `npm` installed? _It sounds like yes based on the description of this PR `Check for update using npm at launch`_ I explicitly set my `node` install to not install the bundled version of `npm`, so I don't think we can take having `npm` installed as a requirement.

Description was outdated, doesn't use `npm` anymore. It uses whichever package manager is used to find the registry, with a fallback to the default registry for special cases, same as `astro add`!

packages/astro/src/core/messages.ts
Princesseuh: It is possible yes!

matthewp (approved):

.changeset/pink-rivers-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Astro will now notify you on start and in the dev toolbar when a new version of Astro is available. This behaviour can be configured using the new `checkUpdates` setting in the Astro config.
```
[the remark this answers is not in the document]
Princesseuh: Ah yes, forgot to update the changeset, will do!

packages/astro/src/cli/install-package.ts
```diff
@@ -161,3 +162,56 @@ async function installPackage(
...
+	} else {
+		return new Error(`Failed to fetch ${registry}/${packageName} - GET ${res.status}`);
+	}
+}
+
+// Users might lack access to the global npm registry, this function
+// checks the user's project type and will return the proper npm registry
+//
+// A copy of this function also exists in the create-astro package
+let _registry: string;
```
natemoo-re, line 203: Can this default to `process.env.npm_config_registry` or is that value just totally irrelevant?

packages/astro/src/core/messages.ts
natemoo-re: Agree that this would be nice to handle!

packages/astro/src/core/dev/update-check.ts
```diff
@@ -0,0 +1,44 @@
...
+		await preferences.set('checkUpdates._lastCheck', Date.now(), { reloadServer: false });
+	}
+
+	_latestVersion = version;
+	return version;
+}
+
+export async function shouldCheckForUpdates(preferences: AstroPreferences): Promise<boolean> {
+	const timeSinceLastCheck = Date.now() - (await preferences.get('checkUpdates._lastCheck'));
+	const hasCheckUpdatesEnabled = await preferences.get('checkUpdates.enabled');
```
natemoo-re, line 37: That's very nice! Not sure if we have a preferences reference page yet, but this would definitely be something to document.

packages/astro/src/preferences/defaults.ts
```diff
@@ -3,6 +3,12 @@ export const DEFAULT_PREFERENCES = {
 		/** Specifies whether the user has the Dev Overlay enabled */
 		enabled: true,
 	},
+	checkUpdates: {
+		/** Specifies whether the user has the update check enabled */
+		enabled: true,
+		/** Time since last check */
+		_lastCheck: 0,
```
natemoo-re: Is this the time **since** the last check or the timestamp **of** the last check?

packages/astro/src/@types/astro.ts
```diff
@@ -2106,6 +2106,14 @@ export interface AstroSettings {
...
 	watchFiles: string[];
 	timer: AstroTimer;
+	/**
+	 * Latest version of Astro, will be undefined if:
+	 * - unable to check
+	 * - the user has disabled the check
+	 * - the check has not completed yet
+	 * - the user is on the latest version already
+	 */
+	latestAstroVersion?: string | undefined;
```
natemoo-re, line 2116: Should this be marked as optional (`?`) if we expect an explicit `undefined` value?

natemoo-re (approved):
Look great, awesome work! Just a few small comments, but nothing blocking!

packages/astro/src/core/dev/update-check.ts
natemoo-re: That definitely seems reasonable. We have this same problem with telemetry, too, so probably worth generalizing a pattern at somepoint.

packages/astro/src/cli/install-package.ts
Princesseuh: The value seems to be irrelevant in many cases, I've even seen npm issues saying that it doesn't respect it. No idea how that works.

packages/astro/src/core/dev/dev.ts
```diff
@@ -34,6 +40,45 @@ export default async function dev(inlineConfig: AstroInlineConfig): Promise<DevS
 	const restart = await createContainerWithAutomaticRestart({ inlineConfig, fs });
 	const logger = restart.container.logger;
 
+	const currentVersion = process.env.PACKAGE_VERSION ?? '0.0.0';
+	const isPrerelease = currentVersion.includes('-');
+
+	if (!isPrerelease) {
+		try {
+			// Don't await this, we don't want to block the dev server from starting
+			shouldCheckForUpdates(restart.container.settings.preferences).then(async (shouldCheck) => {
```
bluwy, line 49: One minor thing I just remembered, could we skip the version check when running tests or in CI? So we don't spam npm everytime we run tests 😅

.changeset/pink-rivers-knock.md
sarah11918, line 5: Just checking up on this, since I don't see a config reference update to docs in this PR. Also, the blog post mentions running `astro preferences disable checkUpdates` to disable, so we should probably have a docs PR to the CLI reference page here: https://docs.astro.build/en/reference/cli-reference/#astro-preferences

Princesseuh: Updated the changeset with the latest info! I'll send a PR to that page.

sarah11918 (approved):
Approving for docs!

Issue #10735, opened 2024-04-10
Infinite loops and runaway billing when middleware creates a 404 response


lilnasy:
### Astro Info

```block
Astro                    v4.5.17
Node                     v20.11.1
System                   Windows (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Quoting https://github.com/withastro/astro/issues/8100#issuecomment-2046201539

> ...there is no way now to return any 404 in the middleware, it always ends up in an infinite loop, even if I create a static 404 page. I'm now forced to diverge from security best practices and return a 403 instead. This way there is no endless loop.

### What's the expected result?

Server does NOT recursively call itself. Context: "Recursive rerouting" section of https://github.com/withastro/roadmap/discussions/681.

### Link to Minimal Reproducible Example

https://github.com/withastro/astro/blob/2acea4defdeb2bc2e43dfb5b339be38d99828885/packages/astro/test/fixtures/custom-404-loop-case-3/src/middleware.js#L4 (skipped in #10084)
https://github.com/withastro/astro/blob/2acea4defdeb2bc2e43dfb5b339be38d99828885/packages/astro/test/custom-404-implicit-rerouting.test.js#L35-L36

### Participation

- [X] I am willing to submit a pull request for this issue.

lilnasy:
@kyr0 Did you encounter the loop in production with the node adapter or just the dev server?

lilnasy:
The tests passed with the test adapter. Confirming that the node adapter is unaffected before closing.

kyr0:
This happens in development with the Node adapter and affects a production deployment with the Vercel adapter, but only if the `output` is set to `hybrid`. I changed to `server` now and everything is nice and shiny again.

PR #10737, opened 2024-04-10
fix(dev): break implicit rerouting loop

lilnasy:
## Changes
- Fixes in dev #10735 

## Testing
- Unskipped tests for implicit rerouting, added prod tests.

## Docs
- Does not affect usage.


matthewp (approved):

lilnasy:
Adding a changeset.

PR #10738, opened 2024-04-10
test for bug reproduction

alexanderniebuhr:
@matthewp this adds a test to reproduce the issue:

1. `pnpm i && pnpm run build`
2. `node --test "packages/astro/test/ssr-prerender-chunk.test.js"`
3. Open `packages/astro/test/fixtures/ssr-prerender-chunks/dist/_worker.js/renderers.mjs`
4. Follow the imports from that file `import { r as reactExports, R as React } from './chunks/prerender_dvgq6Avf.mjs';`
5. Open `packages/astro/test/fixtures/ssr-prerender-chunks/dist/_worker.js/chunks/prerender_dvgq6Avf.mjs`
6. See that the chunking tries to import from a file with `noop` content

**It is not possible to test this with the exisiting `testAdapter`, since that relies on `fs` and also doesn't use a real `serverEntrypoint` but a Vite virtual import. I expect this is an issue with `manualChunks` settings of Astro core, but it might could be also an issue with some settings for the `serverless-test-adapter`. Still I'm very lost and would love to get help & or a fix!**

_(Maybe @bluwy has also an idea 🤔)_

As my debugging goes the following issues are downstream issues caused by this:

- https://github.com/withastro/adapters/issues/224
- https://github.com/withastro/adapters/issues/211
- https://github.com/withastro/adapters/issues/233
- https://github.com/withastro/adapters/issues/214

### DIFF
--- a/packages/astro/test/fixtures/ssr-prerender-chunks/astro.config.mjs
+++ b/packages/astro/test/fixtures/ssr-prerender-chunks/astro.config.mjs
@@ -1,0 +1,10 @@
+import serverlessAdapter from '@test/serverless-test-adapter';
+import { defineConfig } from 'astro/config';
+import react from "@astrojs/react";
+
+// https://astro.build/config
+export default defineConfig({
+  adapter: serverlessAdapter(),
+  output: 'server',
+  integrations: [react()]
+});
--- a/packages/astro/test/fixtures/ssr-prerender-chunks/package.json
+++ b/packages/astro/test/fixtures/ssr-prerender-chunks/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@test/ssr-prerender-chunks",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
+    "@astrojs/react": "^3.2.0",
+    "@test/serverless-test-adapter": "workspace:*",
+    "@types/react": "^18.2.75",
+    "@types/react-dom": "^18.2.24",
+    "astro": "workspace:*",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  }
+}
--- a/packages/astro/test/fixtures/ssr-prerender-chunks/src/components/Counter.tsx
+++ b/packages/astro/test/fixtures/ssr-prerender-chunks/src/components/Counter.tsx
@@ -1,0 +1,26 @@
+import React, { useState } from "react";
+
+const Counter: React.FC = () => {
+    const [count, setCount] = useState<number>(0);
+
+    const increment = () => {
+        setCount((prevCount) => prevCount + 1);
+    };
+
+    const decrement = () => {
+        setCount((prevCount) => prevCount - 1);
+    };
+
+    return (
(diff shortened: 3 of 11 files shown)


bluwy:
I looked into this in the past couple days, and I think it's hard to fix this. I've tried:

1. Use `this.emitFile()` with `preserveSignature: 'strict'` so that the prerender chunk doesn't contain and export unrelated modules, but it wasn't possible to ensure all prerendered pages to go into this chunk. `manualChunks` could've kept some chunks out of it.
2. Instead of making the entire prerender chunk exports a noop, try to analyze the Astro component exports and noop them only. However, it's not possible to easily analyze this with Rollup or `es-module-lexer` without making assumptions of the chunk code and manually treeshake it.

At the end, both approaches aren't perfect either, and I'm not sure if it's worth making them work. I think the more proper solution for now is to:

1. Remove the prerender chunk and noop flow. (Accept that this brings a larger bundle size)
2. Each page should correspond to a Rollup input and contain its page code within it. The former is true today, but the latter is not because each input correspond to (yet another?) dynamic entry chunk: https://github.com/withastro/astro/blob/f95ace9b37c027164ec945329f38bc89b0ec446e/packages/astro/src/core/build/plugins/plugin-pages.ts#L47-L56
3. That way we can safely delete the input chunks of the prerendered pages only. These input chunks should never be used again.

In the future, an even proper solution would be to run two rollup builds: SSG/prerendered pages and SSR/dynamic pages. It would be tricky to do that now because:

1. `export const prerender` can be tied to env vars, we don't know ahead of time if the page should be prerendered.
2. Two Rollup builds may need different configuration by users and adapters (Vite 6 environment API may help).
3. Two Rollup builds is likely to be slow.

alexanderniebuhr:
Thanks @bluwy for looking into this. I do agree with your outcome since I discovered something similar.

I like both ideas, not using the prerender chunk and having two builds. Do you think it would be possible to add those workarounds now and then come back to this with vite 6 and environment api.

alexanderniebuhr:
Also very interested how @matthewp thinks about this, and which approach he would like to go, I just really think we should find a solution now and then look at a long term fix later :)

matthewp:
I'd be fine with getting rid of the noop for now and thinking about another build later. There's a chance we'll need another build when we do server islands (but i hope not) so i could see changing how it works then.

bluwy:
I think we can definitely do what's described at `I think the more proper solution for now is to: ...` for sure. I don't think they will take as much work but still need a little grease. I can try to attempt step 1-3 next week, but if they take more work, doing step 1 only should be fine.

The two builds idea is something we would/could do far in the future though, so not something we should do now but can think about.

alexanderniebuhr:
Sounds great for me. If you're trying to do Step 1-3 next week, I'll be around and be able to test a preview release with the Cloudflare adapter to see if works :) ❤️

bluwy:
Test is included in https://github.com/withastro/astro/pull/11245. Thanks for the test!

PR #10745, opened 2024-04-10
fix(cli): call path.replace only if it is a function

### LINKED ISSUE #10741: TypeError: possibleFilePath?.replace is not a function
### Astro Info

```block
Astro                    v4.5.18
Package Manager          npm
                         @astrojs/vue
                         astro-auth
                         @astrojs/react
                         @sentry/astro
```

_No response_

### Describe the Bug

I get the following error when building on CI server:

```
12:39:27 [ERROR] [vite] x Build failed in 11.04s
TypeError: possibleFilePath?.replace is not a function
    at collectInfoFromStacktrace (file:///app/node_modules/astro/dist/core/errors/dev/utils.js:109:38)
    at file:///app/node_modules/astro/dist/core/errors/dev/utils.js:17:25
    at collectErrorMetadata (file:///app/node_modules/astro/dist/core/errors/dev/utils.js:15:7)
    at throwAndExit (file:///app/node_modules/astro/dist/cli/throw-and-exit.js:16:29)
    at cli (file:///app/node_modules/astro/dist/cli/index.js:172:11)
```

Unfortunately, I can't reproduce it locally or on stackblitz. It actually masks another error I'm debugging. However, looking at the [code](https://github.com/withastro/astro/blob/ba3af206b53d8ea9aca78361c66e9806c660ad2a/packages/astro/src/core/errors/dev/utils.ts#L177), I see a potential mistake that the second `replace` is not called defensively:

```ts
const source = possibleFilePath?.replace(/^[^(]+\(([^)]+).*$/, '$1').replace(/^\s+at\s+/, '');
```

In any case, `collectInfoFromStacktrace` function is not robust against potentially missing data.

Astro's error handling should be robust. Showing raw error is always better than masking the error with own failure.

N/A

### Participation

lilnasy:
## Changes

- Fixes #10741 

## Testing
- Minor change. Existing tests should pass.

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/core/errors/dev/utils.ts
+++ b/packages/astro/src/core/errors/dev/utils.ts
@@ -174,7 +174,7 @@ function collectInfoFromStacktrace(error: SSRError & { stack: string }): StackIn
 			stackText.split('\n').find((ln) => ln.includes('src') || ln.includes('node_modules'));
 		// Disable eslint as we're not sure how to improve this regex yet
 		// eslint-disable-next-line regexp/no-super-linear-backtracking
-		const source = possibleFilePath?.replace(/^[^(]+\(([^)]+).*$/, '$1').replace(/^\s+at\s+/, '');
+		const source = possibleFilePath?.replace?.(/^[^(]+\(([^)]+).*$/, '$1').replace(/^\s+at\s+/, '');
 
 		let file = source?.replace(/:\d+/g, '');
 		const location = /:(\d+):(\d+)/.exec(source!) ?? [];
--- a/.changeset/late-spoons-knock.md
+++ b/.changeset/late-spoons-knock.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes an issue where CLI commands could not report the reason for failure before exiting.


lilnasy:
!preview 10745

packages/astro/src/core/errors/dev/utils.ts
```diff
@@ -174,7 +174,7 @@ function collectInfoFromStacktrace(error: SSRError & { stack: string }): StackIn
 			stackText.split('\n').find((ln) => ln.includes('src') || ln.includes('node_modules'));
 		// Disable eslint as we're not sure how to improve this regex yet
 		// eslint-disable-next-line regexp/no-super-linear-backtracking
-		const source = possibleFilePath?.replace(/^[^(]+\(([^)]+).*$/, '$1').replace(/^\s+at\s+/, '');
+		const source = possibleFilePath?.replace?.(/^[^(]+\(([^)]+).*$/, '$1').replace(/^\s+at\s+/, '');
```
ematipico, line 177: This is very weird... was JS calling `.replace` on `undefined`?? 🤔

ematipico (approved):

packages/astro/src/core/errors/dev/utils.ts
lilnasy: Yes, because possibleFilePath was defined but its type was something other than a string.

PR #10746, opened 2024-04-10
fix some i18n config docs

sarah11918:
## Changes

Updates the config reference for docs for i18n

## Testing

no tests, just docs

## Docs

- adds the necessary`h4` tag to the new `routing.manual` entry
- improves some existing descriptions for other i18n by adding code examples (did not uniformly have these)

STILL TO FIX:

- currently, `manual` shows up BEFORE the other config options. This should appear last, as the "least common/most manual" option. I did not want to play around with the syntax to get the `|` just right by reversing the object and the manual entry.


TheOtterlord (approved):
LGTM!
image

PR #10747, opened 2024-04-10
fix(rendering): remove render instructions from slot expressions

### LINKED ISSUE #7610: Props not being passed to some slots

2.8.0

None

yarn

Mac

Chrome

### Describe the Bug

I am attaching a stripped-down example of what I am experiencing.

Basically, at least one of the slots I'm trying to render does not receive props ( hence the error ), usually the one named after. I found out about this when trying to build a component that receives multiple slots, all were passed props, except the one named after (or anything starting with A, weirdly enough) .
If I remove the props destructuring from the slotted function the slot is rendered.

By debugging the code into node_modules/astro/dist/core/render/result.js, I found out that that slot's expressions sometimes has an Object [Generator] {} as its first member (therefore the getFunctionExpression returns undefined because the expressions length is greater than 1) and I don't know where it comes from. However, sometimes, by changing the slot's name from after to something else the  Object [Generator] {} disappears

I would like the props to be passed to all slots. It seems like the issue has something to do with my example not having a <head> tag and the `after` slot being the first in alphabetical order ( but I have seen it happening even if the <head> tag is included in other examples )

https://stackblitz.com/edit/github-927mjd?file=src%2Fpages%2Findex.astro

### Participation

lilnasy:
## Changes

- Closes #7610 

## Testing
- Added

## Docs
- Does not affect usage.


matthewp (approved):
Awesome fix!

bluwy (approved):

Issue #10724, opened 2024-04-09
When setting the prefixDefaultLocale = false, fallback to default is not working at homepage.


dande-sun:
### Astro Info

```block
Astro                    v4.5.12
Node                     v20.8.1
System                   Linux (x64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/react
                         @astrojs/vue
                         @astrojs/tailwind
                         @astrojs/mdx
                         @astrojs/partytown
                         client:eager
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

when setting below options, the fallback for SG homepage is not working, the `Location` in response is a null string.
`i18n: {
    defaultLocale: "default",
    locales: [
      "default",
      "fr",
    ],
    fallback: {
      "fr": 'default'
    },
    routing: {
      prefixDefaultLocale: false
    }
  },
  trailingSlash: "never"`

This is the response, 
`Request URL:
https://astro.avepoint.com/uk
Request Method:
GET
Status Code:
302 Found
Remote Address:
10.1.0.182:443
Referrer Policy:
strict-origin-when-cross-origin
Date:
Tue, 09 Apr 2024 07:31:48 GMT
Last-Modified:
Tuesday, 09-Apr-2024 07:31:48 UTC
Location:
Referrer-Policy:
no-referrer-when-downgrade
`

### What's the expected result?

It's should be fallback to default homepage.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xfj9kv?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
Can you please create a valid reproduction? The reproduction uses the `fr` language, but then the `pages/` folder contains a `es/` folder instead.

Also, when running `astro build`, I can see the emitted files as expected.

dande-sun:
@ematipico, the `fr` only an example, I set the fallback `fr` to default, but there not have the `fr` homepage, it should be fallback to default homepage, the bug does not need `astro build`, in the `dev` mod is same too.
Anyway I updated the reproduction, https://stackblitz.com/edit/github-xfj9kv

ematipico:
@dande-sun: thank you

PR #10755, opened 2024-04-11
fix(i18n): fallback SSR

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/10724
Closes PLT-1955

There was a small issue when there's so `base` provided

## Testing

I added a test case, and tested manually using the reproduction provided

## Docs

N/A

### DIFF
--- a/.changeset/old-pugs-jog.md
+++ b/.changeset/old-pugs-jog.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes a case where the i18n fallback faild to correctly redirect to the index when the fallback is enabled in SSR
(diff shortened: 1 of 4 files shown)


matthewp (approved):

.changeset/old-pugs-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a case where the i18n fallback faild to correctly redirect to the index when the fallback is enabled in SSR
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where the i18n fallback failed to correctly redirect to the index page with SSR enabled
```

PR #10762, opened 2024-04-11
Enable `allowJs` for `strictest.json`

### LINKED ISSUE #10507: astro:db db import is not typed when using `strictest` typescript preset and a .js config file
### Astro Info

```block
Astro                    v4.5.8
Package Manager          npm
Adapter                  none
                         astro:db
                         @astrojs/db/file-url
```

Not browser specific

### Describe the Bug

*If* I create a brand new project using `npm create astro@latest`: 
1. Select the `strictest` option for typescript 
2. Add astro db via `npx astro add db` 
3. Attempt to use `import { db } from "astro:db"` in a `.astro` file

*Then* typescript will not pick up on DrizzleORM types for that import. User defined tables are available from `astro:db`, but everything else like `db`, `eq`, `gt`, `count`, and `sql` will result in a type error saying those are not exported from `astro:db`

This type error can be seen when using the VSCode astro extension, or when running  `npx astro check`. 

This seemingly only happens for `.astro` files. Other imports like `defineTable` and `defineDb` work just fine in `db/config.ts` and `db/seed.ts`.

Changing `"extends": "astro/tsconfigs/strictest"` to `"extends": "astro/tsconfigs/strict"` in `tsconfig.json` fixes this issue.

Screen capture of changing strictest to strict and seeing the types populate correctly

I would expect that the generated db types from astro would work, even when using the strictest typescript preset.

https://stackblitz.com/edit/astro-db-strictest-bug?file=db%2Fseed.ts

### Participation

bholmesdev:
## Changes

Resolves #10507

This allows typechecking for JS files when using the `strictest` TS config. This is consistent with our other TS configs, and fixes virtual module types provided by integrations like Astro DB (https://github.com/withastro/astro/issues/10507)

### DIFF
--- a/.changeset/thirty-poems-chew.md
+++ b/.changeset/thirty-poems-chew.md
@@ -2,4 +2,6 @@
 "astro": minor
 ---
 
-Enable type checking for JavaScript files for the `strictest` TS config. This ensures consistency with Astro's other TS configs, and fixes type checking for integrations like Astro DB when using an `astro.config.mjs`.
+Enables type checking for JavaScript files when using the `strictest` TS config. This ensures consistency with Astro's other TS configs, and fixes type checking for integrations like Astro DB when using an `astro.config.mjs`.
+
+If you are currently using the `strictest` preset and would like to still disable `.js` files, set `allowJS: false` in your `tsconfig.json`.
--- a/.changeset/thirty-poems-chew.md
+++ b/.changeset/thirty-poems-chew.md
@@ -1,0 +1,5 @@
+---
+"astro": minor
+---
+
+Enable type checking for JavaScript files for the `strictest` TS config. This ensures consistency with Astro's other TS configs, and fixes type checking for integrations like Astro DB when using an `astro.config.mjs`.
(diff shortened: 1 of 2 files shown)


bholmesdev:
!preview js-strictest

Princesseuh (approved):
Fine with me, probably won't break too many people.

.changeset/thirty-poems-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Enable type checking for JavaScript files for the `strictest` TS config. This ensures consistency with Astro's other TS configs, and fixes type checking for integrations like Astro DB when using an `astro.config.mjs`.
```
sarah11918, line 5:
```suggestion
Enables type checking for JavaScript files when using the `strictest` TS config. This ensures consistency with Astro's other TS configs, and fixes type checking for integrations like Astro DB when using an `astro.config.mjs`.

If you are currently using the `strictest` preset and would like to still disable `.js` files, set `allowJS: false` in your `tsconfig.json`.
```
Just snagged this from the blog post, as I think it's helpful!

sarah11918 (approved):
Approving the changeset here, just suggesting that we add the helpful tip from the blog post here in the changeset too, for your consideration!

PR #10763, opened 2024-04-11
Invalidate CC cache manifest when lockfile or config changes

matthewp:
## Changes

- Changes to lockfiles means that the contents of packages have likely changed. When this happens the cache needs to be invalidated because it is based on specific version of modules.
- Logic is to check each known lockfile, since it's possible people are using more than 1 (odd but possible). This happens in parallel to make it as fast as possible.
- Same for the Astro config, except there's only 1 of those so checks in order.
- If there is no lockfile then the person could very well have no versions of packages any time they run `npm install`. For now anyways I'm not sure how to resolve this while staying cached. This is an edge case nowadays as most people use lockfiles.

## Testing

- New tests added that check for invalidation.
- Tested extensively against docs

## Docs

N/A, bug fix


matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

matthewp:
!preview cc-invalid

natemoo-re (approved):
Conceptually LGTM from a quick scan. I'm sure you'll get more thorough reviews from others, tho ☺️

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -290,13 +355,65 @@ async function generateContentManifest(
...
+
+async function configHash(root: URL) {
+	const configFileNames = configPaths;
+	for(const configPath of configFileNames) {
+		try {
+			const fileURL = new URL(encodeURI(joinPaths(root.toString(), configPath)));
+			const handle = await fsMod.promises.open(fileURL, 'r');
+			const data = await handle.readFile();
+			const hash = checksum(data);
+			await handle.close();
```
bluwy, line 403: This code looks the same as `pushBufferInto()`, should we use it instead?

Also, why not use `await fsMod.promises.readFile(fileURL)` instead? It returns a buffer by default

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -321,10 +445,10 @@ export function pluginContent(
 				if (isServerLikeOutput(opts.settings.config)) {
 					return { vitePlugin: undefined };
 				}
-
+				opts.settings.config.build.assets
```
bluwy, line 448:
```suggestion

```

Unused

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -39,40 +41,44 @@ interface ContentManifest {
 	// Tracks components that should be passed to the client build
 	// When the cache is restored, these might no longer be referenced
 	clientEntries: string[];
+	// Hash of the lockfiles, pnpm-lock.yaml, package-lock.json, etc.
+	// Kept so that installing new packages results in a full rebuild.
+	lockfiles: string;
```
ematipico, line 46: nit: the name of the property is plural, but it's only one string. Should it be `lockfile` instead?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -39,40 +41,44 @@ interface ContentManifest {
...
+	// Hash of the Astro config. Changing options results in invalidating the cache.
+	configs: string;
```
ematipico, line 48: nit: he name of the property is plural, but it's only one string. Should it be `config` instead?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -308,8 +425,15 @@ export function pluginContent(
 	opts: StaticBuildOptions,
 	internals: BuildInternals
 ): AstroBuildPlugin {
-	const cachedChunks = new URL('./chunks/', opts.settings.config.cacheDir);
-	const distChunks = new URL('./chunks/', opts.settings.config.outDir);
+	const { cacheDir, outDir } = opts.settings.config;
+
+	const chunksFolder = './chunks/';
```
ematipico, line 430: Where is this `chunks` folder coming from? Is it a rollup thing? If so, I think we should compute some setting because it can change if the build changes

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -290,13 +355,65 @@ async function generateContentManifest(
...
 
 	await Promise.all(promises);
 	return manifest;
 }
 
-function checksum(data: string): string {
-	return createHash('sha1').update(data).digest('base64');
+async function pushBufferInto(fileURL: URL, buffers: Uint8Array[]) {
+	try {
+		const handle = await fsMod.promises.open(fileURL, 'r');
```
ematipico, line 373: What does `'r'` mean?

packages/astro/src/core/build/plugins/plugin-content.ts
```diff
@@ -290,13 +355,65 @@ async function generateContentManifest(
...
+		buffers.push(data);
+		await handle.close();
+	} catch {
+		// File doesn't exist, ignore
+	}
+}
+
+async function lockfilesHash(root: URL) {
+	// Order is important so don't change this.
+	const lockfiles = ['package-lock.json', 'pnpm-lock.yaml', 'yarn.lock', 'bun.lockb'];
```
ematipico, line 384: Why do we hash multiple lock files?

packages/astro/src/core/config/config.ts
```diff
@@ -78,15 +78,17 @@ export function resolveRoot(cwd?: string | URL): string {
 	return cwd ? path.resolve(cwd) : process.cwd();
 }
 
+export const configPaths = Object.freeze([
```
ematipico, line 81: It seems these paths are in order of possible appearance, i.e. `astro.config.mjs` is most likely to appear first. Can we leave a comment to say that?

packages/astro/test/experimental-content-collections-cache-invalidation.test.js
```diff
@@ -0,0 +1,98 @@
...
+		restore() {
+			fs.rmSync(this.cacheDir, { recursive: true });
+			copyFiles(this.tmpDir, this.cacheDir);
+		}
+		rmTmp() {
+			fs.rmSync(this.tmpDir, { force: true, recursive: true });
+		}
+	}
+
+	class ManifestUsed {
```
ematipico, line 27: nit: `TestPlugin`? It took me a while to understand that

packages/astro/src/core/build/plugins/plugin-content.ts
matthewp: I went with `configs` because we might add other config checks in the future (like Tailwind config).

packages/astro/src/core/build/plugins/plugin-content.ts
matthewp: It's possible to use multiple package managers in the same project. So we hash all of them in case this happens and one of them changes. ie you might use `pnpm` and I use `npm`, `pnpm-lock.yaml` is unchanged but `package-lock.json` is changed. This should result in an invalidation.

matthewp: Not even necessarily that you use multiple package managers, but that you have multiple in the same project. You might switch from `npm` to `pnpm`, but the package-lock.json is still in the project. We can't know which one is the "right" one to use, therefore hash them all.

packages/astro/test/experimental-content-collections-cache-invalidation.test.js
matthewp: sure

packages/astro/src/core/build/plugins/plugin-content.ts
matthewp: This is the same hardcoded value in `static-build.ts`. Should it be moved somewhere as a constant and used in both places?

packages/astro/src/core/build/plugins/plugin-content.ts
ematipico: Oh I see, thank you

packages/astro/src/core/build/plugins/plugin-content.ts
ematipico: Maybe it could be useful, so we can inline some comment that explains that's that folder for

ematipico (approved):

packages/astro/src/core/build/plugins/plugin-content.ts
matthewp: Moved it to a constant.

PR #10764, opened 2024-04-11
Add useful links to `@astrojs/db` package.json

delucis:
## Changes

- Adds repo, homepage, and issues links to the DB package.json
- These are surfaced on npm but also in places like editor tooltips
- Adds a `description` field to package.json too. Currently this is an empty string causing the first link of the README to be used in places, e.g.
   npm:  
   image
   VS Code:  
   image

## Testing

n/a — copied the format used in other integrations

## Docs

n/a

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@astrojs/db",
   "version": "0.10.3",
-  "description": "Add libSQL and Astro Studio support to your Astro site.",
+  "description": "Add libSQL and Astro Studio support to your Astro site",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@astrojs/db",
   "version": "0.10.3",
-  "description": "Add SQLite and Astro Studio support to your Astro site.",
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):

packages/db/package.json
```diff
@@ -1,8 +1,15 @@
 {
   "name": "@astrojs/db",
   "version": "0.10.3",
-  "description": "",
+  "description": "Add SQLite and Astro Studio support to your Astro site.",
```
TheOtterlord, line 4: I think we can mention LibSQL over SQLite specifically here, like we do in the docs?

TheOtterlord (approved):
Left a small comment, but looks good to me!

packages/db/package.json
matthewp: Yep, +1

sarah11918 (approved):
Thanks Chris! LGTM!

packages/db/package.json
```diff
@@ -1,8 +1,15 @@
 {
   "name": "@astrojs/db",
   "version": "0.10.3",
-  "description": "",
+  "description": "Add libSQL and Astro Studio support to your Astro site.",
```
delucis, line 4: Micro-nit: just realised other packages don’t end the description with a `.`
```suggestion
  "description": "Add libSQL and Astro Studio support to your Astro site",
```

PR #10767, opened 2024-04-12
fix(types): manually Pick and Omit from `CookieSerializeOptions` interface

lilnasy:
## Changes

- Fixes an issue where all options of `Astro.cookies.set` and `Astro.cookies.delete` were incorrectly reported as required.

## Testing
- Existing tests should pass.

## Docs
- Does not affect usage.

### DIFF
--- a/packages/astro/src/core/cookies/cookies.ts
+++ b/packages/astro/src/core/cookies/cookies.ts
@@ -2,16 +2,155 @@ import type { CookieSerializeOptions } from 'cookie';
 import { parse, serialize } from 'cookie';
 import { AstroError, AstroErrorData } from '../errors/index.js';
 
-export type AstroCookieSetOptions = Pick<
-	CookieSerializeOptions,
-	'domain' | 'path' | 'expires' | 'maxAge' | 'httpOnly' | 'sameSite' | 'secure' | 'encode'
->;
+// `AstroCookieSetOptions` and `AstroCookieDeleteOptions` are manually copied subsections of `CookieSerializeOptions` because `Omit` and `Pick` result in all fields being required.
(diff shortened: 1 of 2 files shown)


florian-lefebvre (commented):
Could we use a partial + a mapped type to allow undefined (maybe there's one built-in TS?) instead? Having to maintain the jsdoc annotations is not ideal imo

lilnasy:
We could but users lose typedocs that way.

I don't think maintainability is much of a concern for this part of code, it's very low traffic.

ematipico (approved):
I think it's a good comprise, especially if the file is rarely modified.

packages/astro/src/core/cookies/cookies.ts
```diff
@@ -2,17 +2,156 @@ import type { CookieSerializeOptions } from 'cookie';
 import { parse, serialize } from 'cookie';
 import { AstroError, AstroErrorData } from '../errors/index.js';
 
-export type AstroCookieSetOptions = Pick<
-	CookieSerializeOptions,
-	'domain' | 'path' | 'expires' | 'maxAge' | 'httpOnly' | 'sameSite' | 'secure' | 'encode'
->;
+// `AstroCookieSetOptions` and `AstroCookieDeleteOptions` are manually copied subsections of `CookieSerializeOptions` because `Omit` and `Pick` result in all fields being required.
```
Princesseuh, line 5: Any reason it cannot be `Partial<Pick...`?

florian-lefebvre: I think it's because jsdoc annotations are lost anyway

PR #10765, opened 2024-04-11
Performance improvement in createAstro function

LeanderG:
## Changes

- I ran a cpu profile and noticed the createAstro function was taking up a lot of time. For some reason the object spreading is very slow.
- this change improves the server stress latency when running `pnpm run benchmark` by about 25%

#### Server stress

|         |        Avg |     Stdev |     Max |
| :------ | ---------: | --------: | ------: |
| Latency | 2401.63 ms | 529.43 ms | 5495 ms |

|           |     Avg |   Stdev |     Min |   Total in 30s |
| :-------- | ------: | ------: | ------: | -------------: |
| Req/Sec   |   400.8 |  171.83 |     300 | 12.0k requests |
| Bytes/Sec | 70.6 MB | 30.2 MB | 52.8 MB |   2.12 GB read |

#### Server stress

|         |       Avg |     Stdev |     Max |
| :------ | --------: | --------: | ------: |
| Latency | 1908.7 ms | 383.24 ms | 3237 ms |

|           |     Avg |   Stdev |     Min |   Total in 30s |
| :-------- | ------: | ------: | ------: | -------------: |
| Req/Sec   |  508.77 |  160.08 |     398 | 15.3k requests |
| Bytes/Sec | 89.6 MB | 28.2 MB | 70.1 MB |   2.69 GB read |
## Testing

This PR doesn't change any behaviour so no tests are needed.

## Docs

This PR doesn't change any behaviour so no docs are needed.

PR #10773, opened 2024-04-12
optimization(runtime): create smaller objects for each Astro global

lilnasy:
## Changes

- Piggybanking on the bottleneck identified by @LeanderG in #10765

## Testing
- Existing tests should pass.

## Docs
- Does not affect behavior.

### DIFF
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -242,12 +242,39 @@ export class RenderContext {
 		return result;
 	}
 
+	/**
+	 * The Astro global is sourced in 3 different phases:
+	 * - **Static**: `.generator` and `.glob` is printed by the compiler, instantiated once per process per astro file
+	 * - **Page-level**: `.request`, `.cookies`, `.locals` etc. These remain the same for the duration of the request.
+	 * - **Component-level**: `.props`, `.slots`, and `.self` are unique to each _use_ of each component.
+	 * 
+	 * The page level partial is used as the prototype of the user-visible `Astro` global object, which is instantiated once per use of a component.
+	 */
 	createAstro(
 		result: SSRResult,
-		astroGlobalPartial: AstroGlobalPartial,
+		astroStaticPartial: AstroGlobalPartial,
 		props: Record<string, any>,
 		slotValues: Record<string, any> | null
 	): AstroGlobal {
+		const slots = new Slots(result, slotValues, this.pipeline.logger) as unknown as AstroGlobal['slots'];
+
+		const astroPagePartial = this.#astroPagePartial ??= this.createAstroPagePartial(result, astroStaticPartial) 
+		const astroComponentPartial = { props, self: null, slots }
+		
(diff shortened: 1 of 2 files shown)


lilnasy:
!bench server-stress

lilnasy:
!bench server-stress

lilnasy:
!bench server-stress

lilnasy:
Odd, I saw a much drastic difference both locally and on gitpod.

packages/astro/src/core/render-context.ts
```diff
@@ -242,12 +242,39 @@ export class RenderContext {
...
 	createAstro(
 		result: SSRResult,
-		astroGlobalPartial: AstroGlobalPartial,
+		astroStaticPartial: AstroGlobalPartial,
 		props: Record<string, any>,
 		slotValues: Record<string, any> | null
 	): AstroGlobal {
+		const slots = new Slots(result, slotValues, this.pipeline.logger) as unknown as AstroGlobal['slots'];
+
+		const astroPagePartial = this.#astroPagePartial ??= this.createAstroPagePartial(result, astroStaticPartial)
```
bluwy, line 261: I think at most `createAstro` is only called once per file, do we need a cache?

bluwy (commented):
@lilnasy I measured the perf for this locally and I can see the improvement. `createAstro()` dropped from 7s to 0.35s with this PR. I think we can go ahead with this, but I have some questions below.

Something else we can also look into is to only call `createAstro` if we use the `Astro` identifier. We'd have to update the compiler.

packages/astro/src/core/render-context.ts
bluwy: Nevermind, it seems like we call `createAstro` per instantiation too.

bluwy:
I pushed a commit to make `Astro.slots` lazy too since most components don't always use them. The `defineProperty` causes the total `createAstro` run about 200ms slower (0.35s to 0.52s), but it creates less garbage and reduces the garbage collection time by 5-7s. So I'll take the smaller hit on it for now.

In which case, the total perf margin is 6.5s + 5s, so 11.5s saved.

packages/astro/src/core/render-context.ts
```diff
@@ -242,12 +242,57 @@ export class RenderContext {
...
 	): AstroGlobal {
+		// Create page partial with static partial so they can be cached together.
+		const astroPagePartial = (this.#astroPagePartial ??= this.createAstroPagePartial(
+			result,
+			astroStaticPartial
+		));
+		// Create component-level partials. `Astro.self` is added by the compiler.
+		const astroComponentPartial = { props, self: null };
+
+		// Create final object. `Astro.slots` will be lazily created.
```
Princesseuh, line 268:
```suggestion
		// Create final object. `Astro.slots` will be lazily created and `Astro.self` is added by the compiler.
```

Princesseuh (approved):

ematipico (approved):

Issue #10752, opened 2024-04-11
Images in `src` folder are not working in vercel branch preview with deployment protection enabled


Trombach:
### Astro Info

```block
Astro                    v4.5.18
Node                     v21.6.2
System                   Linux (x64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/vercel/serverless
Integrations             @astrojs/tailwind
                         @astrojs/svelte
                         @astrojs/mdx
                         @astrojs/react
                         adds-to-head
```

### Describe the Bug

I have not fully understood the cause of this bug, but I am able to reproduce it with a minimal example. It is possible this is entirely on vercel's side but I'm hoping to gain some insight to what's causing this issue here.

It is possible that this bug is also caused under different circumstances, but this is the specific set of conditions I found (also see linked stackblitz). The project must be set to "hybrid" mode and the vercel integration must be added. On pages that are not prerendered, i.e `export const prerender = false;` is set, images imported won't show on vercel preview deployments when deployment protection (vercel authentication) is enabled for that environment (only alt text is shown). The network tab shows 404 responses for the image requests. Prerendered pages, local development and when deployed to vercel's production environment, everything works as expected.

When I disable vercel authentication in the project settings the 404 responses disappear.

![image](https://github.com/withastro/astro/assets/19306765/f3e7716d-7a27-4d5f-83c0-7cabe5e9df05)
![image](https://github.com/withastro/astro/assets/19306765/0fda3060-026a-4c7e-9eb7-91a2fddd24b4)

Please let me know if this issue is entirely on Vercel's end so I can file a bug report with them.

### What's the expected result?

Images should work when vercel deployment protection is active.

### Link to Minimal Reproducible Example

https://stackblitz.com/~/github.com/Trombach/astro-image-test

### Participation

- [ ] I am willing to submit a pull request for this issue.

Princesseuh:
Our image endpoint cannot actually return a 404, so this is either Astro or Vercel returning a 404. Since this only happens when a specific setting in the UI is enabled and only in previews, I would assume that this is on Vercel's end

Trombach:
Thanks @Princesseuh. I have started a [discussion](https://github.com/orgs/vercel/discussions/6480) on the vercel Github

Trombach:
I see, that makes sense. I didn't mention this in my original post, but this actually all used to work just a few weeks ago. I wonder if something changed on Vercel's end :thinking: 
Thanks for looking into this and making a PR!

PR #10775, opened 2024-04-12
fix(assets): Forward headers from the original request to the internal request to the image

Princesseuh:
## Changes

This is similar to something we do in our edge middleware thingy where we forward the headers from the original request to the self-request (is that a thing) so that it works properly when the website is protected. For instance, this fixes image requests in preview website on Vercel that have their preview behind a auth.

Fixes https://github.com/withastro/astro/issues/10752

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/.changeset/rich-spoons-fold.md
+++ b/.changeset/rich-spoons-fold.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Looks good to me!

matthewp (approved):

lilnasy (approved):

Issue #10778, opened 2024-04-12
sitemap integration removes pages ending with 404 or 500


adrianlyjak:
### Astro Info

```block
Astro                    v4.2.3
Node                     v20.11.1
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/react
                         @astrojs/sitemap
```

### Describe the Bug

I am using numerical IDs for items on a static site. One of the items (a store location) has an ID of "404". Its getting stripped from the sitemap, since 404 pages are a somewhat special identifier to be used as a not found page in some cases for astro.

It seems that perhaps the exclusionary logic is a little too strict. Can this be made to instead only match this type of pattern?

```
/404
/500
/[locale]/404
/[locale]/500
```

In my case the page follows this pattern

```
/store/[state]/[city]/[store-id]
```

### What's the expected result?

Not all pages that end in 404 should be excluded from the site map

### Link to Minimal Reproducible Example

https://github.com/adrianlyjak/astro/tree/404-fix/packages/integrations/sitemap/test/fixtures/static

### Participation

- [X] I am willing to submit a pull request for this issue.

matthewp:
@ascorbic I think this one shouldn't be too hard, suspect the logic here is too strict: https://github.com/withastro/astro/blob/main/packages/integrations/sitemap/src/index.ts#L50

adrianlyjak:
@matthewp I think this can actually be closed? I had an PR (#10779) that was merged last month. The updated sitemap plugin is working correctly for my use case now

ematipico:
@adrianlyjak you can close it :)

PR #10779, opened 2024-04-12
Make status code check more strict for sitemap plugin

adrianlyjak:
This is a potential fix for #10778

## Changes

- Makes status code exclusion from sitemap more strict. Only `[locale]/[error-code]` and `[error-code]` style pages are excluded
- requires locales to be configured for locale page exclusion

## Testing

Updated existing test fixture and added test for this scenario

## Docs

The current exclusion logic isn't documented, so it seems unnecessary to add further documentation about this edge case. IMO, this seems like a bug


matthewp:
Can you rebase against `main`?

adrianlyjak:
@matthewp 👌 rebased and all checks pass

bholmesdev (approved):
This fix looks good to me! Confirmed that 404 and 500 should only be detected from `src/pages` in our docs, and should include localized version as well per the i18n middleware. We Adri-can merge I think @adrianlyjak

.changeset/nasty-turtles-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/sitemap": patch
+---
+
+Site map status code exclusion now only excludes pages like /[error-code] or /[locale]/[error-code]
```
bholmesdev, line 5:
```suggestion
Fixed false positives for status code routes like `404` and `500` when generating sitemaps.
```

nit: I'd restructure to describe the bug solved a bit more clearly

PR #10787, opened 2024-04-13
Make viewTransition.finished wait for animations triggered by viewTransition.ready

martrapp:
## Changes

Ensures that the `finished` promise of view transition simulation not only waits for animations triggered by the data-astro-transition-fallback attribute, but also waits for JavaScript animations triggered by the viewTransition.ready promise.

## Testing

Manually tested

## Docs

n.a./bug fix

### DIFF
--- a/.changeset/stupid-countries-warn.md
+++ b/.changeset/stupid-countries-warn.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a timing issue in the view transition simulation.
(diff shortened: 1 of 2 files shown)


bluwy (approved):
It would be nice if there's an example or tests for future reference, but I trust that this works for now 😄

martrapp:
Will add a test ;-) Thank you for merging, bluwy!

martrapp:
Sorry @bluwy, I've promised too much. The fix is for the simulated view transitions and our e2e tests only run with Chrome. Can only be tested with Firefox/Safari. 

But of course I have an example: Please go to https://events-3bg.pages.dev/jotter/starlight/guide/ using Firefox. The PR was necessary so that in the sidebar when changing the category, the old one only closes after the animation that moves the marker is through.

bluwy:
No problem! No worries if a test is hard for this, was mostly curious to see what this fixes so there's a reference we can compare in the future. I can observe the bug now too 👍

PR #10788, opened 2024-04-13
chore: use Biome to format JS files

ematipico:
## Changes

This PR changes the formatter to Biome for the following files: `.js`, `.jsx`, `.ts`, `.tsx` and `.json`.

The rest of the files are still formatted with Prettier.

Here's the format command run locally:

```
 ~/www/withastro/astro/ [chore/faster-formatter] pnpm exec biome format ./ --write
Formatted 1058 files in 181ms. Fixed 45 files.
```

The formatting of `package.json` is different because Prettier actually uses `JSON.stringify(content, null, 2)`; instead, Biome uses the same heuristic as all the other JSON files.

## Testing

CI should still pass

## Docs

N/A


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -168,9 +168,7 @@ export const NoMatchingRenderer = {
 
 ${
 	validRenderersCount > 0
-		? `There ${plural ? 'are' : 'is'} ${validRenderersCount} renderer${
-				plural ? 's' : ''
-			} configured in your \`astro.config.mjs\` file,
+		? `There ${plural ? 'are' : 'is'} ${validRenderersCount} renderer${plural ? 's' : ''} configured in your \`astro.config.mjs\` file,
```
bluwy, line 171: I like this change 😲 Does Biome preserve the single line even if the interpolated expression contains multiple identifiers? That has always bugged me from Prettier.

packages/db/package.json
```diff
@@ -38,29 +38,14 @@
...
-    "index.d.ts",
-    "virtual.d.ts",
-    "dist"
-  ],
-  "keywords": [
-    "withastro",
-    "astro-integration"
-  ],
+  "files": ["index.d.ts", "virtual.d.ts", "dist"],
+  "keywords": ["withastro", "astro-integration"],
```
bluwy, line 48: I feel like making this single-line is less readable. Maybe we should disable formatting for all `package.json` by default? I think package managers already format it on install.

Not a dealbreaker though if others are fine with this change. In that case, we can go ahead with it.

The other formatting looks great 👍

packages/astro/src/core/errors/errors-data.ts
bluwy: Looks like a feature to me 😄

Princesseuh (approved):
Same thought regarding the package.json changes, I think it was more readable before. Everything else looks great!

bluwy (approved):

PR #10793, opened 2024-04-15
feat(toolbar): Add a toolbar example

Princesseuh:
## Changes

Adds a starter example of an integration with a dev toolbar app. This will be coupled with a recipe on the docs.

## Testing

Our smoke test try to build this, so they'll catch if it doesn't build anymore because of a type issue or something.

## Docs

N/A


bluwy (approved):

scripts/smoke/check.js
```diff
@@ -6,9 +6,11 @@ import * as path from 'node:path';
 import pLimit from 'p-limit';
 import { tsconfigResolverSync } from 'tsconfig-resolver';
 
+const skippedExamples = ['toolbar-app', 'component']
```
Princesseuh, line 9: `astro check` is able to handle those examples, because at the end of the day it's kinda just `tsc` + Astro support, but it's kinda not the purpose, so figured it's safe to skip them.

ematipico (approved):

PR #10795, opened 2024-04-16
Improve sitemap generate performance

bluwy:
## Changes

The `generateSitemap()` function was taking a while to run if passed the `i18n` config. Before it took 11.5s:

image

With this PR, it takes 0.055s:

image

I refactored how it handles parsing URLs for i18n purposes, it was calling the `parseUrl` utility O(n^2) times. Now it's only O(n) times as I parse each URLs once, and implement caching for same i18n url paths.

Reading the code may be a bit hard, but the [docs for i18n for sitemap](https://docs.astro.build/en/guides/integrations-guide/sitemap/#i18n) may help give an idea of the kind of work we need to do.

## Testing

Added a new test for `generateSitemap()` to prevent regressions.

## Docs

n/a. internal refactor.


packages/integrations/sitemap/src/utils/parse-i18n-url.ts
```diff
@@ -0,0 +1,42 @@
+interface ParsedI18nUrl {
+	locale: string;
+	path: string;
+}
+
+// NOTE: The parameters have been schema-validated with Zod
+export function parseI18nUrl(
```
bluwy, line 7: This utility is pretty much the same as the `parseUrl` utility (deleted in `parse.url.ts`). But I refactored the code to work slightly faster.

ematipico: I bet one of the bottlenecks was the double `.some` over `localCodes` 🤔

ematipico (approved):
This is great! I am speechless; excellent work, @bluwy!

Princesseuh (approved):

packages/integrations/sitemap/src/utils/parse-i18n-url.ts
bluwy: Yeah that was also very likely too. The first can be skipped because we already validated with Zod.

PR #10796, opened 2024-04-16
Disable streaming for SSG

bluwy:
## Changes

Since we need the entire string before we can write to a file anyway, we don't need streaming. Streaming keeps the state in an array buffer, while non-streaming keeps it as a string. The latter is more performant as most of the input is already a string, preventing transformations.

Credits go to @delucis for the initial idea & benchmark. And @ematipico for the initial PR (https://github.com/withastro/astro/pull/9603)

---

Locally building the docs repo, the render time went from 56s to 44.3s.

I marked this as a patch because I wanted to get this out, but if anyone feels being more cautious, happy to mark this as a minor too 😄 

## Testing

Existing tests should pass

## Docs

n/a. internal change.

### DIFF
--- a/packages/astro/src/runtime/server/render/astro/render.ts
+++ b/packages/astro/src/runtime/server/render/astro/render.ts
@@ -31,6 +31,10 @@ export async function renderToString(
 	let str = '';
 	let renderedFirstPageChunk = false;
 
+	if (isPage) {
+		await bufferHeadContent(result);
+	}
+
 	const destination: RenderDestination = {
 		write(chunk) {
 			// Automatic doctype insertion for pages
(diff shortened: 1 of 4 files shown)


packages/astro/src/runtime/server/render/astro/render.ts
```diff
@@ -31,6 +31,10 @@ export async function renderToString(
 	let str = '';
 	let renderedFirstPageChunk = false;
 
+	if (isPage) {
+		await bufferHeadContent(result);
+	}
```
bluwy, line 36: This was the missing piece to make the non-streaming render work. It exists in `renderToReadableStream` but wasn't ported here too.

ematipico: So simple!

ematipico (approved):

Princesseuh (approved):

PR #10799, opened 2024-04-16
Fixes an issue with persisted non-text input fields that have the focus during view transition navigation.

martrapp:
TIL: setting selectionStart on an HTMLInputElement throws an error on types other than text, search, URL, tel, and password. 

## Changes

Luckily, in all other cases reading the property yields null. 
And we only restore focus if the activeElement did not change.
To prevent the errors in the console it is sufficient to just not assign if we have no numbers.
    
## Testing

Added test for a checkbox input.
Also added explicit test for the selectionStart and selectionEnd on a text input field as an extension to the existing focus test.

## Docs
n.a. / bug fix

### DIFF
--- a/.changeset/itchy-donuts-destroy.md
+++ b/.changeset/itchy-donuts-destroy.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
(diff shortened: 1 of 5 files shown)


ematipico (approved):
Great fix!

PR #10708, opened 2024-04-07
Limit imports in flight for `getCollection`

horo-fox:
## Changes

This is like https://github.com/withastro/astro/pull/7125 but for another place accessing many files at once: `getCollection`. This might address some of https://github.com/withastro/astro/issues/7241 -- some of the errors don't look like they're from `getCollection` but I got `EMFILE` and this fixed my own project.

## Testing

I modified the distributed `runtime.ts` (well, `runtime.js` cause it's in `dist`) in my `node_modules` install of `astro` and my project stopped erroring. Then I did those same edits to the file here.

## Docs

I don't think this needs to be documented as it is a bugfix. Should I include a changelog? I did not for now.

PR #10800, opened 2024-04-17
Batch instead of using p-limit

horo-fox:
cc @jlarmstrongiv

## Changes

Potentially fix https://github.com/withastro/astro/pull/10708#issuecomment-2060010637. We use `p-limit` two other places that could probably use a similar strategy, if we want to remove a dependency.

## Testing

If possible, I would like to try installing this branch as a package to ensure it still works. I'm not familiar enough with package managers to know how to do that.

## Docs

Should just fix a potential import issue.

### DIFF
--- a/.changeset/fresh-plants-worry.md
+++ b/.changeset/fresh-plants-worry.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Batch imports rather than use `p-limit`
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -1,5 +1,4 @@
 import type { MarkdownHeading } from '@astrojs/markdown-remark';
-import pLimit from 'p-limit';
 import { ZodIssueCode, string as zodString } from 'zod';
 import { AstroError, AstroErrorData } from '../core/errors/index.js';
 import { prependForwardSlash } from '../core/path.js';
@@ -81,10 +80,12 @@ export function createGetCollection({
 			// Always return a new instance so consumers can safely mutate it
 			entries = [...cacheEntriesByCollection.get(collection)!];
 		} else {
-			const limit = pLimit(10);
-			entries = await Promise.all(
-				lazyImports.map((lazyImport) =>
-					limit(async () => {
+			
+			entries = []
+
+			// process files 1k at a time to prevent EMFILE
+			for (let i = 0; i < lazyImports.length; i += 1000) {
+				entries.push(...await Promise.all(lazyImports.slice(i * 1000, (i + 1) * 1000).map(async (lazyImport) => {
 						const entry = await lazyImport();
 						return type === 'content'
 							? {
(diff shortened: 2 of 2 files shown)


bluwy:
I don't think we should revert it. Our build setup (Vite) already respects subpath imports, it seems like a different tool from that comment that's not supporting it. @jlarmstrongiv do you have a repro?

jlarmstrongiv:
Here’s the reproduction https://github.com/jlarmstrongiv/astro-p-limit-error it appears the error occurs with the `sst` adapter.  

I suppose I’m in favor of reverting it because breaking `sst` builds in a patch release isn’t ideal

bluwy:
I can't get the project to build. I get `CredentialsProviderError: Could not load credentials from any providers` which perhaps I'm missing some env vars to get this work.

I don't think we're intentionally breaking things here. The package is using subpath imports which should be supported by any tooling that also supports subpath exports. It seems like SST is using esbuild to bundle things:

- https://github.com/sst/sst/blob/0acf9c68e9731755105899e20cad1fba1b118d3c/packages/sst/src/runtime/handlers/node.ts#L150
- https://github.com/sst/sst/blob/0acf9c68e9731755105899e20cad1fba1b118d3c/packages/sst/src/constructs/SsrFunction.ts#L343

And AFAIK esbuild also supports subpath imports. So it seems like there's a bug in SST somewhere that prevents it working.

jlarmstrongiv:
@bluwy I’ve let SST know in the Discord, I’m assuming they’ll ( @bayssmekanique ) follow up here.  Other users have started to confirm astro + sst breaking in Discord too.

> I can't get the project to build. I get CredentialsProviderError: Could not load credentials from any providers which perhaps I'm missing some env vars to get this work.

Here’s instructions on ensuring your AWS credentials are set up properly https://docs.sst.dev/advanced/iam-credentials

---

> I don't think we're intentionally breaking things here. The package is using subpath imports which should be supported by any tooling that also supports subpath exports. It seems like SST is using esbuild to bundle things:
> And AFAIK esbuild also supports subpath imports. So it seems like there's a bug in SST somewhere that prevents it working.

You’re reasoning is sound.  SST does use esbuild behind the scenes (so does Vite).  And it certainly wasn’t international.  Nonetheless, it broke all astro + sst builds in a patch update.  Hopefully, this issue will be resolved from either the sst side or astro side.

bayssmekanique:
This does appear to be an issue that will affect more than just `sst` users. Anyone using Deno, Bun, tsx, or esbuild will face this issue according to the issue comments: https://github.com/sindresorhus/p-limit/issues/75

I was not able to reproduce

Princesseuh:
> This does appear to be an issue that will affect more than just `sst` users. Anyone using Deno, Bun, tsx, or esbuild will face this issue according to the issue comments: [sindresorhus/p-limit#75](https://github.com/sindresorhus/p-limit/issues/75)
> 
> ~My understanding is that esbuild doesn't implement the resolution, it relies on Node's implementation, so this could be an issue of the build-time environments node version not being compatible.~ nvm, I misunderstand Evan's comment but looking at the code he does perform his own module resolution.

Looking at esbuild's issues, support for subpath imports was added multiple years ago. Are you saying that it doesn't work even in esbuild's latest version? Bun also supports subpath imports (though there's apparently bugs in it on some OS)

bluwy:
Closing as we haven't heard of similar reports.

PR #10805, opened 2024-04-17
feat(env): add schema, types and envField

florian-lefebvre:
## Changes

- Updates `AstroConfigSchema`
- Updates `AstroUserConfig`
- Adds `envField`

Related RFC section: https://github.com/withastro/roadmap/blob/feat/astro-env-rfc/proposals/0046-astro-env.md#astro-config

## Testing

Updates a unit test to test custom zod errors

## Docs

N/A


matthewp (dismissed):
Needs to be under experimental.

ematipico:
@matthewp FYI this PR isn't against `main`.

We're using stacked PRs to progressively work on the feature.

florian-lefebvre:
Moreover, I checked and everything is under `experimental` in both types and schema. I'm going to convert to draft for now since it looks like we must make a change to this part of the rfc

packages/astro/src/core/config/schema.ts
```diff
@@ -525,6 +587,12 @@ export const AstroConfigSchema = z.object({
 				.optional()
 				.default(ASTRO_CONFIG_DEFAULTS.experimental.security),
 			i18nDomains: z.boolean().optional().default(ASTRO_CONFIG_DEFAULTS.experimental.i18nDomains),
+			env: z
+				.object({
+					schema: envSchemas.EnvSchema.optional(),
+				})
+				.optional()
+				.default({}),
```
ematipico, line 595: nit: for defaults, we usually use `ASTRO_CONFIG_DEFAULTS`

ematipico (approved):

packages/astro/src/core/config/schema.ts
florian-lefebvre: Saw that but also saw some config use an empty object when there's actually no default. What do you recommend here?

packages/astro/src/core/env/config.ts
```diff
@@ -0,0 +1,48 @@
...
+				}),
+			}),
+		}),
+	};
+};
+
+/**
+ * TODO:
+ */
+export const envField = {
```
florian-lefebvre, line 40: Types don't match actually because of the methods but I'm waiting until we're a bit more sure about what to support and how to name things

ematipico (approved):
I think we can merge it :)

bluwy (approved):
Looks great!

PR #10818, opened 2024-04-19
feat: implement reroute in dev

ematipico:
## Changes

> [!NOTE]
> This is a series of stacked PRs. This PR targets `feat/reroute`

This PR implements the rerouting logic in the development environment.

Technical changes:
- added the experimental flag `experimental.rerouting`
- The `BasePipeline` has now two methods:
	- `tryReroute` which accepts the payload coming from a `reroute` function, and **tries** to find the `RouteData` that matches that path, and the component to render
	- `getComponentInstance`, a function that is called to retrieve the module of the component based on its `RouteData`
- The `DevPipeline` now accepts the `ManifestData`. In dev this is the real type that contains the up to date information about the current routes. It has also a method to update the `ManifestData`, which is called every time the dev server updates
- The core logic is inside `RenderContext`. When rerouting occurs, the new `reroute` function needs to update the `RenderContext` with new data. Information such as `cookies`, `url`, and `params` depend on the new request. 
- The `RenderContext` has new fields:
	- `isRerouting`, used to skip the middleware 
	- `counter`, to detect possible loops. If a loop is detected, a 508 is returned

## Testing

Just one test for now, but I plan to add more of them

## Docs

N/A

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -246,6 +246,10 @@ export interface AstroGlobal<
 	 */
 	redirect: AstroSharedContext['redirect'];
 	/**
+	 * TODO add documentation
+	 */
+	reroute: AstroSharedContext['reroute'];
+	/**
 	 * The <Astro.self /> element allows a component to reference itself recursively.
 	 *
 	 * [Astro reference](https://docs.astro.build/en/guides/api-reference/#astroself)
@@ -1918,6 +1922,18 @@ export interface AstroUserConfig {
 				origin?: boolean;
 			};
 		};
+
+		/**
+		 * @docs
+		 * @name experimental.rerouting
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.6.0
+		 * @description
+		 *
+		 * TODO
+		 */
+		rerouting: boolean;
 	};
 }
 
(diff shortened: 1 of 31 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -245,6 +245,10 @@ export interface AstroGlobal<
 	 * [Astro reference](https://docs.astro.build/en/guides/server-side-rendering/)
 	 */
 	redirect: AstroSharedContext['redirect'];
+	/**
+	 * TODO add documentation
```
Princesseuh, line 249: Are all the TODOs here intentional for this PR?

Princesseuh (approved):

packages/astro/test/middleware.test.js
```diff
@@ -377,3 +377,26 @@ describe(
 		});
 	}
 );
+
+describe.only('middleware with virtual routes', () => {
```
Princesseuh, line 381: `only` 😄

packages/astro/src/@types/astro.ts
ematipico: Yes, they are all intentional and I plan to resolve them in a later PR

matthewp (approved):

PR #10821, opened 2024-04-19
Optimize minor hot code

bluwy:
## Changes

Some minor changes to some hot code paths. During testing, they help improve a few ms only, but I think it should helps memory usage altogether.

## Testing

Existing tests should pass

## Docs

internal change only. no changeset.


packages/astro/src/runtime/server/render/astro/instance.ts
```diff
@@ -49,23 +49,21 @@ export class AstroComponentInstance {
...
-		let value: Promise<AstroFactoryReturnValue> | AstroFactoryReturnValue | undefined =
-			this.returnValue;
-		if (isPromise(value)) {
-			value = await value;
-		}
-		if (isHeadAndContent(value)) {
-			await value.content.render(destination);
+		const returnValue = await this.init(this.result);
+		if (isHeadAndContent(returnValue)) {
+			await returnValue.content.render(destination);
```
bluwy, line 64: This change helps prevent double `await` being called if `this.returnValue` is a promise.

packages/astro/src/runtime/server/render/util.ts
```diff
@@ -22,7 +25,9 @@ const toIdent = (k: string) =>
 	});
 
 export const toAttributeString = (value: any, shouldEscape = true) =>
-	shouldEscape ? String(value).replace(/&/g, '&#38;').replace(/"/g, '&#34;') : value;
+	shouldEscape
+		? String(value).replace(ampersandRegex, '&#38;').replace(doubleQuoteRegex, '&#34;')
+		: value;
```
bluwy, line 30: hoisting the regex prevents the regex instance from creating everytime. When testing I didn't quite find a big perf improvement though, but I feel like it might still be worth having.

I think the most significant cost is with `String()` but I haven't figured a refactor for it yet from the consumer side.

bluwy:
Marking draft for now until I figure out the fails.

bluwy:
Going to merge this with the approval before. Only tweaked with an additional empty `.catch()` call to prevent the false unhandledrejection errors.

PR #10827, opened 2024-04-20
feat(env): add validators

florian-lefebvre:
> Based on #10805

## Changes

- Adds custom validators for strings, numbers and booleans

## Testing

Unit test

## Docs

N/A


packages/astro/test/units/env/env-validators.test.js
```diff
@@ -0,0 +1,228 @@
...
+		});
+
+		it('Should not fail if the optional variable is missing', () => {
+			const input = fixture.createInput(undefined, {
+				type: 'boolean',
+				optional: true,
+			});
+			fixture.givenInput(input);
+			fixture.whenValidating();
+			fixture.thenResultShouldBeValid(undefined);
```
ematipico, line 205: I would change the fixture, considering how you're using them:
```js
      fixture.createInput(undefined, {
				type: 'boolean',
				optional: true,
			});
		  fixture.thenResultShouldBeValid(undefined);
```

The input you create is always the input that you pass in `.giveInput`. 

The `whenValidating()` can be part of `thenResultShouldBeValid` and `thenResultShouldBeInvalid`

florian-lefebvre: I was trying to follow arrange/act/assert but that makes sense, I'll update it

ematipico (approved):

PR #10828, opened 2024-04-20
fix: resolve syntax errors in fixtures

virchau13:
I parsed every `.astro` file in the Astro repository (to check if [tree-sitter-astro](https://github.com/virchau13/tree-sitter-astro) had any bugs), and found three files with syntax errors that don't seem intentional.

## Changes

Fixes those syntax errors.

## Testing

Executed `pnpm run test` at top-level and all 17 tasks passed.

## Docs

It's a simple syntax error fix, so no docs are required.


matthewp:
Thanks!

PR #10829, opened 2024-04-20
feat(env): add vite plugin

florian-lefebvre:
> Based on #10827 

## Changes

- Adds a vite plugin
- Updates zod schema

## Testing

Tests should still pass

## Docs

N/A


matthewp:
@florian-lefebvre should this be in draft? Doesn't look ready for review yet.

florian-lefebvre:
It is, it's a really small pr to make sure I didn't put things in the wrong place

ematipico (approved):

packages/astro/src/core/create-vite.ts
```diff
@@ -155,6 +156,7 @@ export async function createVite(
 			astroDevToolbar({ settings, logger }),
 			vitePluginFileURL({}),
 			astroInternationalization({ settings }),
+			astroEnvVirtualModPlugin({ settings, logger })
```
bluwy, line 159: Only a minor nit, but since this is a Vite pre plugin, it would be nice to put this higher up the list as I think its hook should get higher priority?

florian-lefebvre: i looked at the plugin above (i18n) and it's also a pre plugin so I assumed there was fine. Happy to move it higher but how much?

PR #10832, opened 2024-04-21
Refactor MDX postprocess plugin

bluwy:
## Changes

Refactor the `@astrojs/mdx-postprocess` Vite plugin to its own file and separate its inner content to functions. This makes the code a bit more readable, and prepares for some upcoming changes I'd like to add to this plugin.

I had a few iterations of this refactor locally. Initially using `MagicString` so that we can prepend strings with sourcemap intact, however I noticed it has a significant perf impact, so I switched back to plain strings. Here's the Rollup build time of the Astro docs.

- Main: 113s
- MagicString w/ sourcemap: 132s
- MagicString w/o sourcemap: 111s
- Plain strings (this PR): 110s

## Testing

Existing tests should pass.

## Docs

Internal refactor. No changeset.

### DIFF
--- a/packages/integrations/mdx/src/index.ts
+++ b/packages/integrations/mdx/src/index.ts
@@ -3,19 +3,14 @@ import { fileURLToPath } from 'node:url';
 import { markdownConfigDefaults, setVfileFrontmatter } from '@astrojs/markdown-remark';
 import type { AstroIntegration, ContentEntryType, HookParameters, SSRError } from 'astro';
 import astroJSXRenderer from 'astro/jsx/renderer.js';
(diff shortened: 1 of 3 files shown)


ematipico:
I don't understand. Does this PR move code, or does it change too? If it does change the code, could you please tell us what the changes are? Since you moved the file, it's very difficult to view the change in logic

bluwy:
It only moves the code. There shouldn't be any logic change, other than how it updates the `code` string in the `transform` hook.

matthewp (approved):

ematipico (approved):

ematipico:
Thank you, I was confused by the perf comparison

PR #10848, opened 2024-04-22
feat(env): client/public variables

florian-lefebvre:
> Depends on #10829

## Changes

- Implements client/public variables handling
- Refactors how `src/env.d.ts` references are generated
- Moves `envField` to the right place

## Testing

- Updates the `astro sync` test
- Adds an integration test
- Updates the validators test to also include `getType`

## Docs

N/A


packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -1,28 +1,117 @@
...
-	// TODO: client / public
+	const schema = settings.config.experimental.env.schema ?? {};
+	const loadedEnv = loadEnv(
+		mode === 'dev' ? 'development' : 'production',
+		fileURLToPath(settings.config.root),
+		''
+	);
+
+	const { clientContent, clientDts } = handleClientModule({ schema, loadedEnv });
+	handleDts({ settings, fs, content: clientDts });
```
ematipico, line 42: We are generating code before the `load` function; is that correct? Shouldn't we generate code inside `load`? Types and client code will be emitted only if the user attempts to use the virtual module.

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -1,28 +1,117 @@
...
+	content,
+	settings,
+	fs,
+}: {
+	content: string;
+	settings: AstroSettings;
+	fs: typeof fsMod;
+}) {
+	fs.mkdirSync(settings.dotAstroDir, { recursive: true });
+	fs.writeFileSync(new URL(ENV_TYPES_FILE, settings.dotAstroDir), content, 'utf-8');
```
ematipico, line 72: nit: It's a long shot, but I would wrap these in a `try/catch` in case we can't write files on disk

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -1,28 +1,117 @@
...
+			continue;
+		}
+		const variable = loadedEnv[key];
+		const result = validateEnvVariable(variable === '' ? undefined : variable, options);
+		if (!result.ok) {
+			throw new AstroError({
+				...AstroErrorData.EnvInvalidVariable,
+				message: AstroErrorData.EnvInvalidVariable.message(key, result.type),
+			});
+		}
```
ematipico, line 95: nit: I think it would be nice if we were able to group all the validation errors together and emit one single message at the end. For example, if the user has 4 invalid variables out of 6, they will see all the errors at once.

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -1,28 +1,117 @@
...
+${data.map((e) => `const ${e.key} = data.${e.key};`).join('\n')}
+
+export {
+	${data.map((e) => e.key).join(',\n')}
+}
+	`;
+
+	const clientDts = `declare module "astro:env/client" {
+    ${data.map((e) => `export const ${e.key}: ${e.type};`).join('\n    ')}
+}`;
```
ematipico, line 111: We are looping `data` **four times** (four `.map`). I think we need four arrays defined at the beginning, fill them using the first loop (the `for..of`) and eventually just print them using a `join`.

packages/astro/src/vite-plugin-inject-env-ts/index.ts
```diff
@@ -41,32 +58,49 @@ export async function setUpEnvTs({
...
-	const dotAstroDir = getContentPaths(settings.config).cacheDir;
-	const dotAstroTypeReference = getDotAstroTypeReference(settings.config);
 	const envTsPathRelativetoRoot = normalizePath(
 		path.relative(fileURLToPath(settings.config.root), fileURLToPath(envTsPath))
 	);
 
+	const injectedTypes: Array<InjectedType> = [
+		{
+			filename: CONTENT_TYPES_FILE,
+			condition: () => fs.existsSync(new URL(CONTENT_TYPES_FILE, settings.dotAstroDir)),
```
ematipico, line 68: nit: `condition` -> `meetsCondition`

packages/astro/src/vite-plugin-inject-env-ts/index.ts
```diff
@@ -41,32 +58,49 @@ export async function setUpEnvTs({
...
-			typesEnvContents = `${expectedTypeReference}\n${typesEnvContents}`;
-			await fs.promises.writeFile(envTsPath, typesEnvContents, 'utf-8');
-			logger.info('types', `Added ${bold(envTsPathRelativetoRoot)} type declarations`);
+				if (!typesEnvContents.includes(expectedTypeReference)) {
+					typesEnvContents = `${expectedTypeReference}\n${typesEnvContents}`;
+				}
+			}
 		}
+
+		logger.info('types', `Added ${bold(envTsPathRelativetoRoot)} type declarations`);
```
ematipico, line 93:
```suggestion
		logger.info('types', `Added ${bold(envTsPathRelativetoRoot)} type declarations.`);
```

ematipico (commented):
Left some comments

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1156,6 +1156,18 @@ export const i18nNotEnabled = {
...
+/**
+ * @docs
+ * @description
+ * The failing environment variable does not match the type and constraints your defined in `experimental.env.schema`.
+ */
+export const EnvInvalidVariable = {
+	name: 'EnvInvalidVariable',
+	title: 'Invalid Environment variable',
+	message: (key: string, type: string) => `Variable "${key}" is not of type: ${type}.`,
+	hint: 'The failing environment variable does not match the type and constraints your defined in `experimental.env.schema`.',
```
bluwy, line 1168:
```suggestion
	hint: 'The failing environment variable does not match the type and constraints defined in `experimental.env.schema`.',
```

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1156,6 +1156,18 @@ export const i18nNotEnabled = {
 	hint: 'See https://docs.astro.build/en/guides/internationalization for a guide on setting up i18n.',
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * The failing environment variable does not match the type and constraints your defined in `experimental.env.schema`.
```
bluwy, line 1162:
```suggestion
 * The failing environment variable does not match the type and constraints defined in `experimental.env.schema`.
```

Seems like a typo.

packages/astro/src/env/constants.ts
```diff
@@ -0,0 +1,11 @@
+function resolveVirtualModuleId<T extends string>(id: T): `\0${T}` {
+	return `\0${id}`;
+}
+
+export const VIRTUAL_CLIENT_MODULE_ID = 'astro:env/client';
+export const RESOLVED_VIRTUAL_CLIENT_MODULE_ID = resolveVirtualModuleId(VIRTUAL_CLIENT_MODULE_ID);
```
bluwy, line 6: Personally I think it's overkill to have a function that only prepends `\0`. I think it's fine to hardcode it 😅

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -1,28 +1,117 @@
...
+import type { EnvSchema } from './schema.js';
+import { validateEnvVariable } from './validators.js';
+import { AstroError, AstroErrorData } from '../core/errors/index.js';
+import type fsMod from 'node:fs';
+import { fileURLToPath } from 'node:url';
 
 interface AstroEnvVirtualModPluginParams {
 	settings: AstroSettings;
 	logger: Logger;
+	mode: 'dev' | 'build' | string;
```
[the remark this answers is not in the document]
bluwy: I think it matches this: https://github.com/withastro/astro/blob/fc097e81a85795701c21eaab8a53582e7de0fb45/packages/astro/src/core/create-vite.ts#L41-L48

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
bluwy: Yeah, I also think we should be lazy loading the virtual module code. One thing to also check though is if new env vars get picked up across server restarts. As a Vite plugin, it's best to cache or reset the cache with the `buildStart`/`buildEnd`/`config`/`configResolved` hooks etc so that we re-compute the env when needed.

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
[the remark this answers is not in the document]
ematipico: Thank you for the clarification!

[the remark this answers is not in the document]
florian-lefebvre: Yes exactly! I don't know if it's better to extract to a shared type or something?

ematipico (approved):
Great work here @florian-lefebvre! 

Let's sheep it

bluwy (approved):

PR #10856, opened 2024-04-23
Prevents inputs from changing ViewTransitions' form method or action

### LINKED ISSUE #10849: VIewTransitions break on presence of <input name="action">
### Astro Info

```block
Astro                    v4.6.1
Node                     v18.20.2
Package Manager          pnpm
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

When using ViewTransitions and having semantically valid HTML Inputs with a `name` attribute that's either `method` or `action`, submitting a form results in a broken route: `404 /[object%20HTMLInputElement] not found`

This is caused by a concept called [DOM clobbering](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#naming-form-controls:-the-name-attribute:~:text=DOM%20clobbering%20is,in%20method%20property.) and by incorrectly (but totally intuitively and understandably) assuming `form.action` is always the same a `<form action="">`

Where usually you'd except `form.action` to be `foo`, the following DOM results in `form.action` to point to the `HTMLInputElement` instead.
```html
<form action="foo">
 <input name="action" value="bar">
</form>
``` 

When trying to submit the form, it will read this element as the destination url instead.

The submit handler in View Transitions reading `method` and `action` should not assume reading `form.action` and `form.method` are safe to read in user land. Instead is should explicitly read the attribute. This follows the same pattern browser behavior without `ViewTranstions` enabled.

Suggested changes in workaround.

https://github.com/withastro/astro/blob/8d5f3e8656027023f9fda51c66b0213ffe16d3a5/packages/astro/components/ViewTransitions.astro#L112-L113

```diff
const form = el as HTMLFormElement;
const submitter = ev.submitter;
const formData = new FormData(form, submitter);
// Use the form action, if defined, otherwise fallback to current path.
- 	let action = submitter?.getAttribute('formaction') ?? form.action ?? location.pathname;
+ let action = submitter?.getAttribute('formaction') ?? form.getAttribute('action') ?? location.pathname;
-  const method = submitter?.getAttribute('formmethod') ?? form.method;
+  const method = submitter?.getAttribute('formmethod') ?? form.getAttribute('method') ?? 'get'
```

Happy to create my first PR to Astro for this.

https://stackblitz.com/edit/github-3nkr3h?file=src%2Fpages%2Findex.astro

### Participation

robertvanhoesel:
## Changes

Fixes #10849. Prevents inputs with a name attribute of action or method to break ViewTransitions' form submission

### Before
```html
<form action="foo">
 <input name="action" value="bar">
</form>
```
Submission would result in a broken route `GET /[object%20HTMLInputElement]` since `form.action` now points to an HTML element.

### After
Form submission handler tests action to be a string, and if not reads the action using `form.getAttribute`

## Testing

- Added two additional e2e tests of scenarios that would break without this fix.
- Existing tests enhanced my fix to have more rigorous checking of attributes

## Docs

No docs changed needed

### DIFF
--- a/.changeset/tasty-swans-refuse.md
+++ b/.changeset/tasty-swans-refuse.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Prevents inputs with a name attribute of action or method to break ViewTransitions' form submission
--- a/packages/astro/components/ViewTransitions.astro
+++ b/packages/astro/components/ViewTransitions.astro
@@ -108,9 +108,16 @@ const { fallback = 'animate' } = Astro.props;
 			const form = el as HTMLFormElement;
 			const submitter = ev.submitter;
 			const formData = new FormData(form, submitter);
+			// form.action and form.method can point to an <input name="action"> or <input name="method">
+			// in which case should fallback to the form attribute
+			const formAction =
+				typeof form.action === 'string' ? form.action : form.getAttribute('action');
+			const formmethod =
+				typeof form.method === 'string' ? form.method : form.getAttribute('method');
 			// Use the form action, if defined, otherwise fallback to current path.
-			let action = submitter?.getAttribute('formaction') ?? form.action ?? location.pathname;
-			const method = submitter?.getAttribute('formmethod') ?? form.method;
+			let action = submitter?.getAttribute('formaction') ?? formAction ?? location.pathname;
+			// Use the form method, if defined, otherwise fallback to "get"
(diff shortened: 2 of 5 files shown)


martrapp:
Thank you so much ;-)! 
Will give @matthewp a chance to approve, if he finds time.

matthewp (approved):
wonderful tests

PR #10858, opened 2024-04-23
Actions experimental release

bholmesdev:
## Changes

Our experimental release for astro actions. This brings type-safe backend handlers to Astro, and implements the **full design** [outlined in the RFC](https://github.com/withastro/roadmap/pull/912).

## Testing

- Add unit tests for form data validation
- Add integration test for fetching actions
- Add e2e test for client-side components calling actions

## Docs

Kept to configuration reference only for this release. We link to the RFC for complete documentation. Thanks for the early review Sarah!


bholmesdev:
!preview actions

florian-lefebvre:
A few comments after watching the video:
- It's really cool! (and not only because it uses AIK)
- I wonder if it could be possible to provide a helper like `safe` to handle errors client side
- What happen if you throw from within the action? Because `ActionError` seemed really input validation specific
- Would be great to also be able to provide `output` zod schema for more optional strictness. Like if it's not provided, infer the handler return type otherwise take it's type

bholmesdev:
Hey thanks for checking this out @florian-lefebvre!
- Ah, I forgot to mention: `safe()` works on the client too 😄 For example, calling `const result = await safe(actions.like(...))` will give you the union type of error and success.
- I didn't get to discuss exception throwing either. I intended `ActionError` to be the generic error object you throw from handler code if you want an explicit exception. I was thinking we follow the same signature as [TrpcError](https://trpc.io/docs/server/error-handling), which lets you set a human-readable `code` (maps to REST statuses) and a `message`. Validation errors use a special `ValidationError` class that extends `ActionError` to include a Zod payload as well. From your feedback, it sounds like "errors are just for validation" was the gut reaction. Worth some discussion on API refinements.
- Good to know `output` is of interest to you! This is [also a feature of tRPC](https://trpc.io/docs/server/validators#output-validators). Were there any use cases brought this feature to mind for you?

bholmesdev:
!preview actions

bholmesdev:
!preview actions

matthewp:
!preview actions

bholmesdev:
!preview actions

Princesseuh:
!preview actions

matthewp:
!preview actions

packages/astro/src/@types/astro.ts
```diff
@@ -1691,6 +1691,105 @@ export interface AstroUserConfig {
 		 */
 		directRenderScript?: boolean;
 
+		/**
```
bholmesdev, line 1694: Paging @sarah11918 for early docs review!

bholmesdev:
!preview actions

bholmesdev:
!preview actions

.changeset/shaggy-moons-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Add experimental support for the actions API. Actions let you define type-safe endpoints you can query from client components with progressive enhancement built-in.
```
sarah11918, line 5:
```suggestion
Adds experimental support for the Actions API. Actions let you define type-safe endpoints you can query from client components with progressive enhancement built in.
```
`Actionally`, we *do* capitalize API names, so let's give you a big ol' flashy capital A!

(And if we don't, we should, so consider it so!)

bholmesdev: Ah cool, wasn't sure if we did! Definitely think a capital will make Actions easier to call out in the docs

packages/astro/src/@types/astro.ts
```diff
@@ -1691,6 +1691,105 @@ export interface AstroUserConfig {
...
+		 *     handler: async ({ postId }, context) => {
+		 *       // insert comments in db
+		 *
+		 *       return comment;
+		 *     },
+		 *   }),
+		 * };
+		 * ```
+		 *
+		 * Then, call an action from your client components using the `actions` object from `astro:actions`. You can pass a type-safe object when using JSON, or a `FormData` object when using `accept: 'form'` in your action definition:
```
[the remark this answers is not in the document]
bholmesdev: Ah, I'm clarifying this is _the_ FormData object that's standard in the browser. I can add an MDN reference, and probably remove the code fence:https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects

packages/astro/src/@types/astro.ts
```diff
@@ -1691,6 +1691,105 @@ export interface AstroUserConfig {
...
+		 *     actions: true,
+		 *   },
+		 * }
+		 * ```
+		 *
+		 * All actions are declared in a global actions handler: the `src/actions/index.ts` file.
+		 *
+		 * You can define an action using the `defineAction()` utility from the `astro:actions` module. This accepts the `handler` property to define your server-side request handler. If your action accepts arguments, apply the `schema` property to validate parameters with Zod.
+		 *
+		 * This example defines two actions: `like` and `comment`. The `like` action accepts a JSON object with a `postId` string, while the `comment` action accepts [form data](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects) with `postId`, `author`, and `body` strings.
```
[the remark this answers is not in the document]
bholmesdev: Sounds good! Added a sentence along those lines

.changeset/shaggy-moons-peel.md
sarah11918: Yes, as long as we are naming the "Actions API" then A is capitalized. When we're just saying "actions", it's not.

packages/astro/package.json
```diff
@@ -54,6 +54,13 @@
     "./components": "./components/index.ts",
     "./components/*": "./components/*",
     "./toolbar": "./dist/toolbar/index.js",
+    "./actions/runtime/virtual/server.js": {
+      "types": "./dist/actions/runtime/virtual/server.d.ts",
+      "default": "./dist/actions/runtime/virtual/server.js"
+    },
+    "./actions/runtime/virtual/client.js": "./dist/actions/runtime/virtual/client.js",
+    "./actions/runtime/route.js": "./dist/actions/runtime/route.js",
+    "./actions/runtime/middleware.js": "./dist/actions/runtime/middleware.js",
```
matthewp, line 63: do you need to add these individually or would a wildcard work?

packages/astro/src/@types/astro.ts
```diff
@@ -1691,6 +1707,103 @@ export interface AstroUserConfig {
...
+		 *     actions: true,
+		 *   },
+		 * }
+		 * ```
+		 *
+		 * Declare all your actions in `src/actions/index.ts`. This file is the global actions handler.
+		 *
+		 * Define an action using the `defineAction()` and `defineFormAction()` utilities from the `astro:actions` module. These accept the `handler` property to define your server-side request handler. If your action accepts arguments, apply the `input` property to validate parameters with Zod.
+		 *
+		 * This example defines two actions: `like` and `comment`. The `like` action accepts a JSON object with a `postId` string, while the `comment` action accepts [FormData](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects) with `postId`, `author`, and `body` strings. Each `handler` may update your database and return a type-safe response.
```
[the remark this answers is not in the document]
bholmesdev: Yeah, just used "may" because the example is contrived. The handler does whatever you tell it to! I'll just say "updates"

packages/astro/src/@types/astro.ts
```diff
@@ -238,6 +238,22 @@ export interface AstroGlobal<
 	response: ResponseInit & {
 		readonly headers: Headers;
 	};
+	/**
+	 * get an action result on the server when using a form POST.
```
sarah11918, line 242:
```suggestion
	 * Get an action result on the server when using a form POST.
```
I don't usually review these, but just checking that you intended lower case here!

packages/astro/actions-module.template.mjs
```diff
@@ -0,0 +1,49 @@
+import { ActionError, callSafely } from 'astro:actions';
+
+function toActionProxy(actionCallback = {}, aggregatedPath = '/_actions/') {
+	return new Proxy(actionCallback, {
+		get(target, objKey) {
+			const path = aggregatedPath + objKey.toString();
+			if (objKey in target) {
+				return target[objKey];
+			}
+			async function action(param) {
```
matthewp, line 10: can you lift this function out of the scope? Preferably to the top-level? So that way it's not recreated (consuming memory) on every invocation.

.changeset/shaggy-moons-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Adds experimental support for the Actions API. Actions let you define type-safe endpoints you can query from client components with progressive enhancement built in.
```
sarah11918, line 5: This is perfectly fine as is, @bholmesdev , but if you want, you can compare to Ema's experimental rerouting changeset. We *do* have a precedent of making hefty minor changesets for experimental flag: at least showing the config option but also optionally stuffing some examples in.

If you'd *like* to make this bigger, faster, stronger, you are welcome to do so!

sarah11918 (approved):
Approving for docs transparency, noting that I've given Ben extra leeway to add showing the experimental flag configuration in a code block and potentially more.

packages/astro/actions-module.template.mjs
```diff
@@ -0,0 +1,49 @@
+import { ActionError, callSafely } from 'astro:actions';
+
+function toActionProxy(actionCallback = {}, aggregatedPath = '/_actions/') {
```
ematipico, line 3: Why is this file here and not inside `src/`? Also, I think it's a top-level comment that explains this file

packages/astro/actions-module.template.mjs
ematipico, line 10: Can you add a comment that explain what `param` is?

packages/astro/actions-module.template.mjs
```diff
@@ -0,0 +1,49 @@
...
+						code: 'BAD_REQUEST',
+						message:
+							'Action unexpectedly called on the server. If this error is unexpected, share your feedback on our RFC discussion: https://github.com/withastro/roadmap/pull/912',
+					});
+				}
+				const headers = new Headers();
+				headers.set('Accept', 'application/json');
+				let body = param;
+				if (!(body instanceof FormData)) {
+					body = param ? JSON.stringify(param) : undefined;
```
ematipico, line 22: Is it safe to stringify `param`? Are you sure that it won't fail?

packages/astro/actions-module.template.mjs
```diff
@@ -0,0 +1,49 @@
...
+					throw new ActionError({
+						code: 'BAD_REQUEST',
+						message:
+							'Action unexpectedly called on the server. If this error is unexpected, share your feedback on our RFC discussion: https://github.com/withastro/roadmap/pull/912',
+					});
+				}
+				const headers = new Headers();
+				headers.set('Accept', 'application/json');
+				let body = param;
+				if (!(body instanceof FormData)) {
```
ematipico, line 21: Should we force the `application/x-www-form-urlencoded` header when **it's a `FormData`**?

packages/astro/actions-module.template.mjs
```diff
@@ -0,0 +1,49 @@
...
+			}
+			action.toString = () => path;
+			action.safe = (input) => {
+				return callSafely(() => action(input));
+			};
+			action.safe[Symbol.for('astro:action:safe')] = true;
+			action.safe.toString = () => path;
+			// recurse to construct queries for nested object paths
+			// ex. actions.user.admins.auth()
+			return toActionProxy(action, path + '.');
```
ematipico, line 44: Recursion! 😱 Can't we achieve the same using a loop?

packages/astro/src/actions/index.ts
```diff
@@ -0,0 +1,87 @@
+import { existsSync } from 'node:fs';
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
+import type { AstroIntegration } from '../@types/astro.js';
+import { ACTIONS_TYPES_FILE, RESOLVED_VIRTUAL_MODULE_ID, VIRTUAL_MODULE_ID } from './consts.js';
+import type { Plugin as VitePlugin } from 'vite';
+
+const name = 'astro:actions';
```
ematipico, line 7: Duplicate of `VIRTUAL_MODULE_ID`

packages/astro/src/actions/index.ts
```diff
@@ -0,0 +1,87 @@
...
+					stringifiedActionsPath,
+					root: params.config.root,
+				});
+			},
+		},
+	};
+}
+
+const vitePluginActions: VitePlugin = {
+	name: 'astro:actions',
```
ematipico, line 47: Duplicate of `VIRTUAL_MODULE_ID`

packages/astro/actions-module.template.mjs
bholmesdev: Good point, can do!

packages/astro/src/actions/index.ts
```diff
@@ -0,0 +1,87 @@
...
+	},
+	async load(id, opts) {
+		if (id !== RESOLVED_VIRTUAL_MODULE_ID) return;
+
+		let code = await readFile(
+			new URL('../../actions-module.template.mjs', import.meta.url),
+			'utf-8'
+		);
+		if (opts?.ssr) {
+			code += `\nexport * from 'astro/actions/runtime/virtual/server.js';`;
```
ematipico, line 62: Genuine question: this is a barrel file. Is it safe in terms of performance to use a barrel file inside a virtual module?

packages/astro/src/actions/index.ts
```diff
@@ -0,0 +1,87 @@
...
+	type Actions = typeof import(${stringifiedActionsPath})["server"];
+
+	export const actions: Actions;
+}`;
+
+	const dotAstroDir = new URL('.astro/', root);
+
+	if (!existsSync(dotAstroDir)) {
+		await mkdir(dotAstroDir);
+	}
```
ematipico, line 84: nit: use `mkdir(dotAstroDir, {recursive: true})`, you don't need `existsSync`: https://nodejs.org/docs/v20.12.1/api/fs.html#fspromisesmkdirpath-options

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+	};
+};
+
+export const onRequest = defineMiddleware(async (context, next) => {
+	const locals = context.locals as Locals;
+	const { request } = context;
+	const contentType = request.headers.get('Content-Type');
+	locals._actionsInternal = {
+		getActionResult: () => undefined,
+	};
```
ematipico, line 19: nit: you might want to protect this object with a `Proxy` or `Object.create` from being rewritten from users.

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	if (!formContentTypes.some((f) => contentType?.startsWith(f))) return next();
```
ematipico, line 21: You have to `.toLowerCase` the value of the headers. The values aren't case-sensitive. We do this in the check origin too. E.g. you might get `APPLICATION/JSON` as header value, and it's valid.

Also, why do we use`startsWith`?

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+	if (typeof actionPath !== 'string') return next();
+
+	const actionPathKeys = actionPath.replace('/_actions/', '').split('.');
+	const action = await getAction(actionPathKeys);
+	const result = await ApiContextStorage.run(context, () => callSafely(() => action(formData)));
+
+	locals._actionsInternal = {
+		getActionResult: (actionFn) => {
+			if (actionFn.toString() !== actionPath) return Promise.resolve(undefined);
+			if (Symbol.for('astro:action:safe') in actionFn) {
```
ematipico, line 34: Any chance to have `Symbol.for('astro:action:safe')` inside a constant instead of repeating it?

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+				// @ts-ignore It's expected that `action` may not match the generic passed in
```
ematipico, line 35: nit: `ts-ignore` -> `ts-expect-error`

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	if (!formContentTypes.some((f) => contentType?.startsWith(f))) return next();
+
+	const formData = await request.clone().formData();
+	const actionPath = formData.get('_astroAction');
+	if (typeof actionPath !== 'string') return next();
+
+	const actionPathKeys = actionPath.replace('/_actions/', '').split('.');
+	const action = await getAction(actionPathKeys);
+	const result = await ApiContextStorage.run(context, () => callSafely(() => action(formData)));
```
ematipico, line 29:
```suggestion
	const result = await ApiContextStorage.run(_, () => callSafely(() => action(formData)));
```

packages/astro/package.json
bluwy, line 1: Could this be formatted back to spaces? pnpm will change this to spaces when installing a new dependency, so would be nice to avoid this diff.

packages/astro/src/actions/runtime/route.ts
```diff
@@ -0,0 +1,41 @@
+import type { APIRoute } from '../../@types/astro.js';
+import { ApiContextStorage } from './store.js';
+import { formContentTypes, getAction } from './utils.js';
+import { ActionError } from './virtual/shared.js';
+
+export const POST: APIRoute = async (context) => {
+	const { request, url } = context;
+	if (request.method !== 'POST') {
+		return new Response(null, { status: 405 });
+	}
```
ematipico, line 10: Is this possible? To call this function with a method that isn't a `POST`?

packages/astro/src/actions/runtime/route.ts
```diff
@@ -0,0 +1,41 @@
...
+		args = await request.clone().formData();
+	} else if (contentType === 'application/json') {
+		args = await request.clone().json();
+	} else {
+		return new Response(null, { status: 415 });
+	}
+	let result: unknown;
+	try {
+		result = await ApiContextStorage.run(context, () => action(args));
+	} catch (e) {
```
ematipico, line 25: I am seeing some repeated code between this function and the middleware. Can't we create some shared function?

packages/astro/src/actions/runtime/route.ts
```diff
@@ -0,0 +1,41 @@
...
+	const actionPathKeys = url.pathname.replace('/_actions/', '').split('.');
+	const action = await getAction(actionPathKeys);
+	const contentType = request.headers.get('Content-Type');
+	let args: unknown;
+	if (formContentTypes.some((f) => contentType?.startsWith(f))) {
+		args = await request.clone().formData();
+	} else if (contentType === 'application/json') {
+		args = await request.clone().json();
+	} else {
+		return new Response(null, { status: 415 });
```
ematipico, line 20: nit: maybe let's add the MDN link to status code 415

packages/astro/src/actions/runtime/route.ts
```diff
@@ -0,0 +1,41 @@
...
+	}
+	let result: unknown;
+	try {
+		result = await ApiContextStorage.run(context, () => action(args));
```
ematipico, line 24: Why don't we call `callSafely` here?

packages/astro/src/actions/runtime/virtual/client.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function defineAction() {
+	throw new Error('[astro:action] `defineAction()` unexpectedly used on the client.');
+}
+
+export function getApiContext() {
+	throw new Error('[astro:action] `getApiContext()` unexpectedly used on the client.');
+}
+
+export const z = new Proxy(
```
ematipico, line 11: nit: can we have a better name instead of `z`?

packages/astro/src/actions/runtime/virtual/server.ts
```diff
@@ -0,0 +1,141 @@
...
+		if (!inputSchema) return await handler(unparsedInput);
+		const parsed = await inputSchema.safeParseAsync(unparsedInput);
+		if (!parsed.success) {
+			throw new ActionInputError(parsed.error.issues);
+		}
+		return await handler(parsed.data);
+	};
+}
+
+export function upgradeFormData<T extends z.AnyZodObject>(
```
ematipico, line 103: Let's comment on this function. What does upgrading a form mean?

packages/astro/src/actions/runtime/virtual/server.ts
```diff
@@ -0,0 +1,141 @@
...
+		} else if (validator instanceof z.ZodArray) {
+			obj[key] = upgradeArray(key, formData, validator);
+		} else {
+			obj[key] = upgradeBase(key, formData, validator, baseValidator);
+		}
+	}
+	return obj;
+}
+
+function upgradeArray(key: string, formData: FormData, validator: z.ZodArray<z.ZodUnknown>) {
```
ematipico, line 124: Same here. The "upgrading" concept requires some clarification.

ematipico (commented):
Sorry for the many questions and comments. Overall it seems good, however, the changes on the `Content-Type` header are crucial.

packages/astro/actions-module.template.mjs
bholmesdev: Oh, I would actually prefer to force `multipart/form-data` to ensure `File` objects are encoded correctly. Is there a reason to prefer `urlencoded`?

packages/astro/src/actions/runtime/virtual/shared.ts
```diff
@@ -0,0 +1,145 @@
...
+				return 'UNSUPPORTED_MEDIA_TYPE';
+			case 500:
+				return 'INTERNAL_SERVER_ERROR';
+			default:
+				return 'INTERNAL_SERVER_ERROR';
+		}
+	}
+
+	static async fromResponse(res: Response) {
+		if (res.status === 400 && res.headers.get('Content-Type')?.startsWith('application/json')) {
```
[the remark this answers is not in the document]
bholmesdev: Ah, this is a check specifically for validation errors. These are owned internally and should raise the `400` status code. For all other codes, we fall to a base `ActionError` object.

packages/astro/src/actions/index.ts
bholmesdev: Hm, I hadn't considered performance for `*` exports. Enumerating each export individually would mean another place to update exports, which could cause future bugs. How would we improve this?

packages/astro/src/actions/index.ts
bholmesdev: TIL!

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Yep, worth a const

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Wait, wouldn't this _remove_ the context we're trying to provide? Otherwise `ApiContextStorage` will be empty.

packages/astro/src/actions/runtime/route.ts
bholmesdev: Ah, in the Response body or as a comment?

packages/astro/src/actions/runtime/route.ts
bholmesdev: Because I'm dumb. I'll see if we can refactor.

packages/astro/src/actions/runtime/virtual/client.ts
bholmesdev: Wanted to stay consistent with content collections exposing `z`, since it _is_ the entirety of Zod. A name change would imply this is some superset of Zod.

Still, there's a general question here: **why can you import Zod from multiple modules in Astro?** It's now exposed from `astro:content` and `astro:actions`. Seemed out-of-scope to fix for the experimental release, but I want to discuss a change.

packages/astro/src/actions/runtime/virtual/server.ts
bholmesdev: Fair. I'll try a clearer function name, or a comment if I can't come up with one.

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Good point! definitely need a `.toLowerCase` check. I also found the need for `startsWith` when using web-standard form POST requests. For whatever reason, Chrome may add a suffix to urlencoded forms like `urlencoded---other-info`. The `startsWith` parses correctly. Happy to add a comment explaining

packages/astro/src/actions/index.ts
```diff
@@ -0,0 +1,88 @@
...
+import { ACTIONS_TYPES_FILE, RESOLVED_VIRTUAL_MODULE_ID, VIRTUAL_MODULE_ID } from './consts.js';
+import type { Plugin as VitePlugin } from 'vite';
+
+export default function astroActions(): AstroIntegration {
+	return {
+		name: VIRTUAL_MODULE_ID,
+		hooks: {
+			async 'astro:config:setup'(params) {
+				const stringifiedActionsPath = JSON.stringify(
+					new URL('actions', params.config.srcDir).pathname
```
matthewp, line 13: It's hard to tell how this is being used, but should this be using `fileURLToPath` instead, otherwise I think this will not work on Windows.

packages/astro/actions-module.template.mjs
matthewp: I think FormData has to be `multipart/form-data`.

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	locals._actionsInternal = {
+		getActionResult: (actionFn) => {
+			if (actionFn.toString() !== actionPath) return Promise.resolve(undefined);
+			if (Symbol.for('astro:action:safe') in actionFn) {
+				// It's expected that `action` may not match the generic passed in
+				return result as any;
+			}
+			if (result.error) throw result.error;
```
matthewp, line 38: From https://github.com/withastro/roadmap/pull/912#discussion_r1592571393 it sounds like you are onboard with making `getActionResult` always the "safe" version, if so I would go ahead and make the change now.

packages/astro/src/actions/runtime/route.ts
matthewp: Pretty sure he means in a comment.

packages/astro/src/actions/runtime/route.ts
```diff
@@ -0,0 +1,38 @@
...
+	} catch (e) {
+		if (e instanceof ActionError) {
+			return new Response(JSON.stringify(e), {
```
matthewp: Does this include the stack trace? Presumably you don't want that included in the client.

packages/astro/actions-module.template.mjs
bholmesdev: Gotcha. I know the `fetch()` function automatically applies that header when a `FormData` object is used.

packages/astro/actions-module.template.mjs
bholmesdev: Fair, it _might_ fail to stringify. I'll add a try / catch with a readable error.

packages/astro/src/actions/index.ts
bholmesdev: this is embedding the actions path as an ESM import path. Since this is static and resolved by Vite, I thought `pathname` would be the way to go. Am I correct if that's the intent?

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Sounds good! Would remove the `Symbol` check as well.

packages/astro/src/actions/runtime/route.ts
bholmesdev: Good catch. I'll ensure stack traces are dev-only. I noticed that's the practice tRPC uses

packages/astro/actions-module.template.mjs
bholmesdev: @florian-lefebvre That's right. I noticed you created a `templates/` directory for these files in your env PR, so I'll preemptively move there

packages/astro/actions-module.template.mjs
bholmesdev: Hm, I can't see a way on first impression. We're building a URL with each dot chain `.` you add to a function. We can't predict how deep the function chaining will be, so we don't have an upper bound for a loop. To you have a suggestion if that's our constraint?

matthewp (changes requested):
From https://github.com/withastro/roadmap/pull/912#discussion_r1592571393, please remove non-safe version of `Astro.getActionResult`

packages/astro/src/actions/runtime/middleware.ts
ematipico: Oh sorry, I though it was a closure providing a param 😆

packages/astro/src/actions/runtime/route.ts
ematipico: Yeah just a comment. 415 isn't a very common code that you see around, so a link would be helpful

packages/astro/src/actions/runtime/virtual/client.ts
ematipico: I didn't know it was something around Zod, sorry about that

packages/astro/src/actions/runtime/middleware.ts
ematipico: Oh really? That's... new to me 🤔 maybe it's worth leaving a comment for posterities

packages/astro/actions-module.template.mjs
ematipico: > We can't predict how deep the function chaining will be, so we don't have an upper bound for a loop

That's unsettling :( no worries, we can review it later if that becomes an hot path

bholmesdev:
@matthewp Non-safe version has been removed 👍

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Ah ha, this is what I was seeing. Chrome may apply a `boundary` property to the content-type. We could split on `;` for completeness from reading the spec
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#content-type_in_html_forms

matthewp (approved):

packages/astro/src/actions/runtime/route.ts
bholmesdev: I did a side-by-side comparison, and they each seem pretty unique with the other comments addressed. Middleware ignores JSON checking and defines `locals` instead of returning a `415` for instance. I'll leave for now

ematipico (approved):
I am really hyped for this release! Thank you for the massive effort!

packages/astro/src/actions/index.ts
bholmesdev: Renamed to `stringifiedActionsImport` for clarity

bholmesdev:
!preview actions

PR #10864, opened 2024-04-24
Tag MDX component for faster checks when rendering

bluwy:
## Changes

> Merges into the `mdx-v3` branch. Technically this is non-breaking, but it's easier to merge here to prevent new releases for `@astrojs/mdx`.

This PR does two things:
1. Inject `__astro_tag_component__(Content, 'astro:jsx')` so that we can render MDX components faster. No need to call `check()` to test if it's `astro:jsx` as we already tagged it.
2. Add enhanced error handling for `astro:jsx` `renderToStaticMarkup()`. Because of the above, `check()` wasn't enhancing the error anymore, which failed [a test](https://github.com/withastro/astro/blob/f513ba602c9d9bc2f1984bf994381d30cdea6a0d/packages/integrations/mdx/test/invalid-mdx-component.test.js#L31-L37).

Performance:

I tested this in the docs repo but there isn't any significant improvements nor regressions, so this would be harmless. Part of the reason I added this is to port a faster version of https://github.com/withastro/astro/blob/main/packages/astro/src/vite-plugin-mdx/tag.ts.

## Testing

Existing tests should pass. 

## Docs

Added changesets about the improved error message, but otherwise this shouldn't have a significant impact on normal usage.

### DIFF
--- a/packages/astro/src/jsx/server.ts
+++ b/packages/astro/src/jsx/server.ts
@@ -4,6 +4,8 @@ import { renderJSX } from '../runtime/server/jsx.js';
 
 const slotName = (str: string) => str.trim().replace(/[-_]([a-z])/g, (_, w) => w.toUpperCase());
 
+// NOTE: In practice, MDX components are always tagged with `__astro_tag_component__`, so the right renderer
+// is used directly, and this check is not often used to return true.
 export async function check(
 	Component: any,
 	props: any,
@@ -19,18 +21,7 @@ export async function check(
 		const result = await Component({ ...props, ...slots, children });
 		return result[AstroJSX];
 	} catch (e) {
-		const error = e as Error;
-		// if the exception is from an mdx component
-		// throw an error
-		if (Component[Symbol.for('mdx-component')]) {
-			throw new AstroError({
-				message: error.message,
-				title: error.name,
-				hint: `This issue often occurs when your MDX component encounters runtime errors.`,
-				name: error.name,
-				stack: error.stack,
-			});
-		}
+		throwEnhancedErrorIfMdxComponent(e as Error, Component);
 	}
 	return false;
 }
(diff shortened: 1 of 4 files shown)


packages/integrations/mdx/src/vite-plugin-mdx-postprocess.ts
```diff
@@ -103,13 +95,52 @@ export default Content;`;
...
+			!isSpecifierImported(
+				code,
+				imports,
+				astroTagComponentImportRegex,
+				'astro/runtime/server/index.js'
+			)
+		) {
+			code += `\nimport { __astro_tag_component__ } from 'astro/runtime/server/index.js';`;
+		}
+		code += `\n__astro_tag_component__(Content, 'astro:jsx');`;
```
bluwy, line 123: This mimics `tag.ts` but specific to the MDX component so it's much simpler. We only need to tag the default exported `Content` component. I don't think there will be other exports that are MDX components.

packages/astro/src/jsx/server.ts
```diff
@@ -48,8 +39,27 @@ export async function renderToStaticMarkup(
...
+	} catch (e) {
+		throwEnhancedErrorIfMdxComponent(e as Error, Component);
+		throw e;
+	}
+}
+
+function throwEnhancedErrorIfMdxComponent(error: Error, Component: any) {
+	// if the exception is from an mdx component
+	// throw an error
+	if (Component[Symbol.for('mdx-component')]) {
```
ematipico, line 54: We started to move those symbols inside [`constants.ts`,](https://github.com/withastro/astro/blob/main/packages/astro/src/core/constants.ts) so we could document them. Does it make sense for this symbol to be placed together with the rest?

packages/integrations/mdx/src/vite-plugin-mdx-postprocess.ts
```diff
@@ -103,13 +95,52 @@ export default Content;`;
...
+	if (ssr) {
+		if (
+			!isSpecifierImported(
+				code,
+				imports,
+				astroTagComponentImportRegex,
+				'astro/runtime/server/index.js'
+			)
+		) {
+			code += `\nimport { __astro_tag_component__ } from 'astro/runtime/server/index.js';`;
```
ematipico, line 121: I can't understand one thing. Where do we define this `__astro_tag_component__` binding/variable? Is it already defined somewhere in the codebase?

ematipico (approved):

PR #10869, opened 2024-04-24
Use unified plugin only for MDX transform

bluwy:
## Changes

- Port `jsx/babel.ts` to `jsx/rehype.ts`. The file logic are essentially the same thing but written as a rehype plugin.
- The builtin `vite-plugin-mdx` plugin is deleted by `@astrojs/mdx`, the integration will takeover MDX transformation instead, using the rehype plugin.

Performance:

Rollup build time of Astro docs dropped from **108s to 96s (12s or 11% faster)**

How this perf improvement works is by removing the hoops needed to compile MDX as JS:
- Before: MDX string -> mdast -> hast -> esast -> JSX string -> esbuild -> babel ast -> JS string
- After: MDX string -> mdast -> hast -> esast -> JS string

## Testing

Existing tests should pass, we have extensive coverage for MDX usecases and edgecases.

## Docs

This is a breaking change. I've added a changeset explaning the change, but the only change needed is to upgrade to Astro v4.8.0 (to be released together)

### DIFF
--- a/packages/astro/src/jsx/rehype.ts
+++ b/packages/astro/src/jsx/rehype.ts
@@ -1,6 +1,10 @@
 import type { RehypePlugin } from '@astrojs/markdown-remark';
 import type { RootContent } from 'hast';
-import type { MdxJsxFlowElementHast, MdxJsxTextElementHast } from 'mdast-util-mdx-jsx';
+import type {
+	MdxJsxAttribute,
+	MdxJsxFlowElementHast,
+	MdxJsxTextElementHast,
+} from 'mdast-util-mdx-jsx';
 import { visit } from 'unist-util-visit';
 import type { VFile } from 'vfile';
 import { AstroError } from '../core/errors/errors.js';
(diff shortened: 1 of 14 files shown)


bluwy:
Strangely lots of fail on CI 😢 Marking as draft

packages/astro/src/jsx/rehype.ts
bluwy, line 1: This file is a port of `./babel.ts` so I understand if this is a bit raw to review 😄 Let me know if there's part you don't understand or grasp and I can add additional comments.

packages/integrations/mdx/src/plugins.ts
```diff
@@ -30,7 +31,6 @@ export function createMdxProcessor(mdxOptions: MdxOptions, extraOptions: MdxProc
 		rehypePlugins: getRehypePlugins(mdxOptions),
 		recmaPlugins: mdxOptions.recmaPlugins,
 		remarkRehypeOptions: mdxOptions.remarkRehype,
-		jsx: true,
```
bluwy, line 33: This is removed so the default of `jsx` is `false`, so MDX will compile directly to JS instead of needing a babel step.

packages/integrations/mdx/src/vite-plugin-mdx-postprocess.ts
```diff
@@ -37,9 +37,9 @@ export function vitePluginMdxPostprocess(astroConfig: AstroConfig): Plugin {
 /**
  * Inject `Fragment` identifier import if not already present.
  */
-function injectFragmentImport(code: string, imports: readonly ImportSpecifier[]) {
-	if (!isSpecifierImported(code, imports, fragmentImportRegex, 'astro/jsx-runtime')) {
-		code += `\nimport { Fragment } from 'astro/jsx-runtime';`;
+function injectUnderscoreFragmentImport(code: string, imports: readonly ImportSpecifier[]) {
+	if (!isSpecifierImported(code, imports, underscoreFragmentImportRegex, 'astro/jsx-runtime')) {
```
bluwy, line 41: I changed the `Fragment` detection to `_Fragment` as it's often already generated by MDX, which should reduce the code generation.

bluwy:
@ematipico This is merging into the `mdx-v3` branch, so I'd like to have this reviewed and merged soon if possible, until everything's ready then I'll send the PR for `mdx-v3` into `main` with the 4.8 milestone 😅

packages/integrations/mdx/src/plugins.ts
Princesseuh: That must be quite faster I assume!

ematipico:
I didn't see that, sorry @bluwy 

Feel free to remove the milestone and merge it whenever you want

PR #10871, opened 2024-04-24
Respect "--mode" flag when running dev command

acupofspirt:
## Changes

This PR allows us to fully utilize Vite's **--mode** flag and its ability to load appropriate **.env.[mode]** files for local development.
So this is similar to https://github.com/withastro/astro/pull/4850 but for `astro dev`

## Testing

Unit tests were added.

## Docs

It's important to note that the current documentation doesn't cover **--mode** flag.

The passed value will be reflected in `import.meta.env.MODE` Someone may use `import.meta.env.MODE === 'development'` to check if it's **not production** mode. However, setting the flag to "staging" or anything else will disrupt these checks. In such cases, it's recommended to use `import.meta.env.DEV` instead.


bluwy:
There's a past PR for this that was closed https://github.com/withastro/astro/pull/6735

I'm not sure this is sufficient to support mode in Astro. Even if it works in dev for now, there's an in-balance and may not work in other places, e.g. builds.

I think it's better to properly solve different modes if we want to support this, which will take some work.

acupofspirt:
@bluwy, you are right. I did it just to get this functional parity between dev and build commands.

ematipico:
> There's a past PR for this that was closed #6735
> 
> EDIT: Well I just noticed that you linked #4850, so we already somewhat support this mode. I don't think supporting custom modes will work everywhere in Astro, but if this is to only align with #4850 then I guess it's fine for now.

If what you say is correct, maybe we should add some documentation (considering that we don't have it) and explain expectations. Like, that it won't work everywhere.

acupofspirt:
@bluwy @ematipico After a more thorough analysis of the code base I decided to use a new flag instead of the existing one. It will be a cleaner and safer solution. It won't affect internal Astro logic (I've checked all possible places). It just passes the flag value to Vite config.
I've updated the PR description and forcepushed new changes. Later I'll open a PR for the docs.

.changeset/bright-zebras-end.md
```diff
@@ -0,0 +1,6 @@
+---
+"astro": minor
+"@astrojs/vercel": patch
+---
+
+add new --viteMode cli flag for dev and build commands
```
sarah11918, line 6:
```suggestion
Adds a new `--viteMode` CLI flag for `dev` and `build` commands to allow you to override the default `development` and `production` modes.
```
Just added some code formatting here, but also I think a statement describing the *usefulness* of this feature can go here! Say what's good about it! (Suggestion above, but correct for accuracy/hype mode. )

You could even add a "This is useful for/when..." second sentence, if you really wanted to.  (For minor releases, we tend to put a little more in the changesets, and then this info is useful to have when we go to write the blog post about the new features.)

sarah11918 (commented):
Just a quick changeset edit from docs, and in case wanted, giving you permission to "sell" this feature a little more!  :smile:

matthewp:
I don't like leaking Vite in this way. I agree that `--mode` is weird, why does it exist and what is it for? Well we never documented so who knows. I would be in favor of replacing `--mode` if there was a very clear definition of its usage. Even if that mirrors exactly what you are aiming for here.

matthewp:
After discussing we decided that we don't want to take on this feature at this time:

- We don't want to expose an option called `viteMode`, we should have our own mode concept and make that configurable.
- There are related issues around `async sync` where people ant to conditionally render differently in these environments. It feels like we should instead do a rethinking of what `mode` is and take all of this into account.

For now I suggest using regular environment variables if you want to have different behavior in dev / build / staging.

PR #10875, opened 2024-04-25
docs: updated comment from Accept-Header to Accept-Language

W1M0R:
## Changes

This change corrects a typo in the comments. More specifically, Accept-Header is changed to Accept-Language.

## Testing

No tests, since this is affects comments only.

## Docs

No docs, because nothing noteworthy was changed.


ematipico (approved):
Good catch. Can you add a changeset? This is official documentation that we provide to the users, so changes should be tracked

florian-lefebvre (approved):

PR #10877, opened 2024-04-25
Allow remark/rehype plugins added after mdx to work

bluwy:
## Changes

This PR moves the markdown options generation to the `astro:config:done` hook so that remark/rehype plugins added by integrations after `@astrojs/mdx` during `astro:config:setup` hook works.

This should unblock `astro-expressive-code` needing `@astrojs/mdx` in a specific order to work, which should remove the hack in https://github.com/withastro/docs/pull/7972

cc @hippotastic does this work for you?
cc @delucis 

---

NOTE: The implementation would be simpler if I added the Vite plugin in the `astro:config:done`, but other integrations also add Vite plugins in `astro:config:setup`. I figured it's more predictable if we also do so.

## Testing

Added a new test.

## Docs

I don't think this will affect many projects in practice, but I added a changeset explaining the migration path if so.

### DIFF
--- a/packages/integrations/mdx/src/vite-plugin-mdx.ts
+++ b/packages/integrations/mdx/src/vite-plugin-mdx.ts
@@ -16,12 +16,9 @@ export function vitePluginMdx(mdxOptions: MdxOptions): Plugin {
 			processor = undefined;
 		},
 		configResolved(resolved) {
-			// The first time we access `mdxOptions`, make sure it's populated (just in case)
-			if (Object.keys(mdxOptions).length === 0) {
-				throw new Error(
-					'`mdxOptions` is not populated. This is an internal error. Please file an issue.'
-				);
-			}
+			// `mdxOptions` should be populated at this point, but `astro sync` doesn't call `astro:config:done` :(
+			// Workaround this for now by skipping here. `astro sync` shouldn't call the `transform()` hook here anyways.
+			if (Object.keys(mdxOptions).length === 0) return;
 
 			processor = createMdxProcessor(mdxOptions, {
 				sourcemap: !!resolved.build.sourcemap,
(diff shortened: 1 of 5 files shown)


hippotastic:
This looks like a great improvement that will allow me to eliminate a bunch of code to check for integration ordering of EC vs MDX, and should also prevent the `rehype-optimize-static`-related hack. Love it!

packages/integrations/mdx/src/vite-plugin-mdx.ts
```diff
@@ -16,6 +16,10 @@ export function vitePluginMdx(mdxOptions: MdxOptions): Plugin {
 			processor = undefined;
 		},
 		configResolved(resolved) {
+			// `mdxOptions` should be populated at this point, but `astro sync` doesn't call `astro:config:done` :(
+			// Workaround this for now by skipping here. `astro sync` shouldn't call the `transform()` hook here anyways.
```
delucis, line 20: Interesting. @bholmesdev mentioned an issue with `astro sync` and Astro DB the other day, which I wonder if it might be related — it also uses `astro:config:done` in a similar pattern to this to assign settled integration state for use by plugins.

https://github.com/withastro/astro/blob/5bf5dd496bf55c0713754f2a4f7726693b759744/packages/db/src/core/integration/index.ts#L88-L89

Maybe something that will need fixing in `sync` itself?

delucis (commented):
Thank you @bluwy! Agree with Hippo that this will be a nice user QoL thing to not need to worry about order any more.

packages/integrations/mdx/src/vite-plugin-mdx.ts
bluwy: Yeah I think it would be nice if the hook was called in Astro sync too. I initially did that, but walked back a bit because I'm unsure if this breaks the db integration.

I noticed the db integration is running the same startup code in both `astro sync` and `astro:config:done`, and I'm unsure what happens in `astro sync`  if it runs that code twice. It should be something we can look into fixing in the future though.

delucis (approved):
LGTM!

PR #10879, opened 2024-04-25
chore: upgrade biome and renovate

ematipico:
## Changes

Upgrades Biome to the latest, and adds a new extension to renovate. With that extension, renovate will be able to migrate the configuration file alone.

## Testing

Current CI should pass

## Docs

N/A


.github/renovate.json5
```diff
@@ -4,7 +4,8 @@
     "config:recommended",
     "schedule:weekly",
     "group:allNonMajor",
-    ":disablePeerDependencies",
+    ":disablePeerDependencies", 
+    "regexManagers:biomeVersions"
```
bluwy, line 8: Since we pin `@biomejs/biome`, do we want renovate to auto update the biome schema version? it might not work between versions.

ematipico: There's a chance a person updates biome and forgets to update the schema version.

In case that happens, renovate will update it when it raises the usual PR. That's my understanding from the person that added the feature

bluwy: Ah so the version in `biome.json` will always match the version that's installed in `node_modules`? That sounds great then 👍

bluwy (approved):

PR #10881, opened 2024-04-25
feat(env): server/public variables

florian-lefebvre:
> Depends on #10848

## Changes

- Handles the server module id
- Throws if module is imported client-side

## Testing

- Adds a fixture and some tests

## Docs

N/A


packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -57,11 +61,23 @@ export function astroEnvVirtualModPlugin({
...
+				return RESOLVED_VIRTUAL_SERVER_MODULE_ID
+			}
 		},
-		load(id) {
+		load(id, options) {
 			if (id === RESOLVED_VIRTUAL_CLIENT_MODULE_ID) {
 				return clientTemplates!.content;
 			}
+			if (id === RESOLVED_VIRTUAL_SERVER_MODULE_ID) {
+				if (options?.ssr) {
```
ematipico, line 73: I think you should use `isServerLikeOutput(options.settings.config);` where `options` is `StaticBuildOptions`.

ematipico (commented):
Should we have some test case for `output: "hybrid"` with prerendered pages (opt-in and opt-out)?

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
florian-lefebvre: I think this is irrelevant here. We just don't want the module to be imported clients-side, no matter the output. Or am I missing something?

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1164,7 +1164,19 @@ export const i18nNotEnabled = {
...
+
+/**
+ * @docs
+ * @description
+ * Module is only available server-side
+ */
+export const EnvServerOnlyModule = {
+	name: 'EnvServerOnlyModule',
+	title: 'Module is only available server-side',
+	message: (name: string) => `The "${name}" module is only available server-side.`,
```
bluwy, line 1179: Some thoughts for myself: I know we have a few virtual modules that currently only meant to work in the server-side only. Maybe we can consider making this more generic later. We don't have to change this now.

bluwy (approved):

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
bluwy: I think using `options?.ssr` should be fine here since it's contextual to the current `load`.

matthewp (approved):

ematipico (approved):

PR #10882, opened 2024-04-25
Improve `asDrizzleTable()` types

delucis:
## Changes

- Improves the typing of the `asDrizzleTable()` utility

- Fixes a type error when passing the output of `defineTable()` to the utility and returns a more detailed type inferred from the columns of the passed table config.

## Testing

Tested manually in an integration

## Docs

n/a — bug fix

### DIFF
--- a/packages/db/src/runtime/index.ts
+++ b/packages/db/src/runtime/index.ts
@@ -8,9 +8,10 @@ import {
 	sqliteTable,
 	text,
 } from 'drizzle-orm/sqlite-core';
-import { type DBColumn, type DBTable } from '../core/types.js';
-import { type SerializedSQL, isSerializedSQL } from './types.js';
+import { type ColumnsConfig, type DBColumn, type DBTable, type TableConfig } from '../core/types.js';
+import { type SerializedSQL, isSerializedSQL, type Table } from './types.js';
 import { pathToFileURL } from './utils.js';
+import { tableSchema } from '../core/schemas.js';
 
 export type { Table } from './types.js';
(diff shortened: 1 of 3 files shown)


packages/db/src/runtime/index.ts
```diff
@@ -8,8 +8,14 @@ import {
 	sqliteTable,
 	text,
 } from 'drizzle-orm/sqlite-core';
-import { type DBColumn, type DBTable } from '../core/types.js';
-import { type SerializedSQL, isSerializedSQL } from './types.js';
+import { tableSchema } from '../core/schemas.js';
```
bholmesdev, line 11: Ah ha, it looks like `tableSchema` would pull a core module into the runtime. We have a lint check against this. Perhaps we can move `asDrizzleTable` from `runtime/` to `core/` and still expose as a package export?

delucis: Took a quick look, but not sure if that would work? `asDrizzleTable` is used in the Vite plugin:

https://github.com/withastro/astro/blob/cc83367c4fd48ec8bf493d620ef5615780388420/packages/db/src/core/integration/vite-plugin-db.ts#L122

delucis: OK, tried a different approach, which seems to work. I think the `asDrizzleTable()` we expose to users needs to be different from the one we use internally, so in the latest commit I wrapped the internal version with the tweaks we made for use _only_ by integration authors.

bholmesdev (approved):
Glad we could addr`izz` the issue!

PR #10883, opened 2024-04-26
Add web-vitals integration

delucis:
## Changes

- Adds a new `@astrojs/web-vitals` integration
- ~~This branch currently includes some changes to `@astrojs/db` from #10882 that will go away once that is merged~~

## Testing

Added some basic tests for dev mode. Not quite sure what a build test would look like or if we should have end-to-end tests for the client-side script?

## Docs

Minimal docs in the README for now cc @withastro/maintainers-docs


packages/integrations/web-vitals/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+							'Please add an SSR adapter: https://docs.astro.build/en/guides/server-side-rendering/'
+					);
+				}
+
+				// Middleware that adds a `<meta>` tag to each page.
+				addMiddleware({ entrypoint: '@astrojs/web-vitals/middleware', order: 'post' });
+				// Endpoint that collects metrics and inserts them in Astro DB.
+				injectRoute({
+					entrypoint: '@astrojs/web-vitals/endpoint',
+					pattern: '/_/astro-vitals',
```
delucis, line 33: Currently injecting the data collection endpoint to `/_/astro-vitals/`.

Should we discuss whether we want a default namespacing approach? For example, the image service uses `/_image`.

packages/integrations/web-vitals/src/middleware.ts
delucis, line 1: This middleware preserves streaming and injects a `<meta>` tag to each page’s `<head>`. Wouldn’t mind someone taking a quick look to make sure there are no footguns here.

delucis (commented):
Ready for review! Testing is a little bit minimal. Not sure if we’d want an end-to-end test for the client script.

packages/integrations/web-vitals/src/env.d.ts
bluwy, line 1: Is this file intended to be committed?

bluwy (approved):
Looks great! The logic seems good to me.

packages/integrations/web-vitals/src/index.ts
bluwy: `/_/` sounds nice, I suppose we can only change `/_image` to it in a major since it's somewhat exposed?

Also bikeshedding 😅 Should we drop the `astro` and do `/_/web-vitals` or `/_/vitals` instead?

packages/integrations/web-vitals/src/env.d.ts
delucis: Yes — currently Astro DB types aren’t working great for integrations that don’t have an Astro project environment to refer to, so we’re kind of working around that with this triple-slash reference.

Hopefully something @bholmesdev or a TS magician might be able to improve in the future?

packages/integrations/web-vitals/src/index.ts
delucis: Yeah, there are a bunch of potential approaches. I guess most extreme would be vendor-style namespacing matching the package name:

```
/_/astrojs/web-vitals/
```

Or alternatively, we don’t worry too much, and eventually support users customising the endpoint location to handle any conflicts.

packages/integrations/web-vitals/src/client-script.ts
```diff
@@ -0,0 +1,36 @@
...
+		pathname,
+		route,
+		name,
+		id,
+		value,
+		rating,
+	}));
+	const body = JSON.stringify(rawBody);
+	const endpoint = '/_/astro-vitals';
+	if (navigator.sendBeacon) navigator.sendBeacon(endpoint, body);
```
bholmesdev, line 24: TIL about `sendBeacon` 💡

packages/integrations/web-vitals/src/index.ts
bholmesdev: Naming is always tough! Maybe we could match the namespacing expected by `astro add`:
- use vendor prefixing for third party packages
- drop the "astro" for first party packages

So use either `/_/web-vitals/` or `/_web-vitals/`. I bias towards the latter to match `_images` and avoid introducing a new competing standard.

packages/integrations/web-vitals/src/middleware.ts
```diff
@@ -0,0 +1,60 @@
...
+	}
+	route = miniEncodeAttribute(stripTrailingSlash(route));
+	return `<meta name="x-astro-vitals-route" content="${route}" />`;
+}
+
+function stripTrailingSlash(str: string) {
+	return str.length > 1 && str.at(-1) === '/' ? str.slice(0, -1) : str;
+}
+
+function miniEncodeAttribute(str: string) {
```
bholmesdev, line 54: We sure this covers all necessary cases?

bholmesdev (approved):
Implementation looks robust and well-tested. A `vital` addition to Astro

packages/integrations/web-vitals/src/index.ts
delucis: Thanks for the feedback! Renamed to `/_web-vitals` and moved to a shared constant, so we can easily rename in the future if we want.

packages/integrations/web-vitals/src/middleware.ts
delucis: I’m pretty sure for attributes it does? I think in fact even `<` and `>` would be safe unencoded — just tried in an Astro component and this input:
```jsx
content={'/blog/&"<>\\\'/'}
```
output to
```html
content="/blog/&#38;&#34;<>\'/"
```
So only the `&` and the `"` were escaped.

PR #10887, opened 2024-04-26
Improve MDX optimize with sibling nodes

bluwy:
## Changes

There was a flaw with how `rehype-optimize-static.ts` worked before. Take this page for example https://docs.astro.build/en/concepts/why-astro/, which is a static page with no components.

Before, the rehype plugin will only transform to something like this:
```jsx
<_components.h1 set:html="Why Astro?" /> 
<_components.p set:html="<strong>Astro</strong> is the web framework..." />
<_components.h1 set:html="Features" /> 
// ...
```

With this PR, it can go one step further and transform as:

```jsx
<Fragment set:html="<h1>Why Astro?</h1>\n<strong>Astro..." />
```

Which is only a single line and a single string, which simplifies the AST nodes and runtime rendering.

Performance:

<table>
<tr>
 <td>
 <td>Rollup build
 <td>Page rendering
<tr>
 <td>Before
 <td>86s
 <td>46s
<tr>
 <td>After
 <td>83s
 <td>34s
<tr>
 <td>Comparison
 <td><b>-3s (3.5% faster)</b>
 <td><b>-12s (26% faster)</b>
</table>

A bit dissapointed of the Rollup build time change, but I'll take the rendering improvement 😄 

## Testing

The `mdx-optimize` playground that should still pass. And I tested in the Astro docs repo.

## Docs

Added a changeset and updated the internal README explaining the new code.

### DIFF
--- a/packages/integrations/mdx/src/rehype-optimize-static.ts
+++ b/packages/integrations/mdx/src/rehype-optimize-static.ts
@@ -8,6 +8,11 @@ export interface OptimizeOptions {
 	ignoreComponentNames?: string[];
 }
 
+interface ElementMetadata {
+	parent: Node;
+	index: number;
+}
+
 const exportConstComponentsRe = /export\s+const\s+components\s*=/;
 
 /**
@@ -45,9 +50,11 @@ export function rehypeOptimizeStatic(options?: OptimizeOptions) {
 		const allPossibleElements = new Set<Node>();
 		// The current collapsible element stack while traversing the tree
 		const elementStack: Node[] = [];
+		// Metadata used by `findElementGroups` later
+		const elementMetadatas = new WeakMap<Node, ElementMetadata>();
 
 		visit(tree, {
-			enter(node, key) {
+			enter(node, key, index, parents) {
 				// `estree-util-visit` may traverse in MDX `attributes`, we don't want that. Only continue
 				// if it's traversing the root, or the `children` key.
 				if (key != null && key !== 'children') return SKIP;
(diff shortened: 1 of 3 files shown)


packages/integrations/mdx/src/rehype-optimize-static.ts
```diff
@@ -29,10 +34,11 @@ export function rehypeOptimizeStatic(options?: OptimizeOptions) {
 		for (const child of tree.children) {
 			if (child.type === 'mdxjsEsm' && exportConstComponentsRe.test(child.value)) {
 				// Try to loosely get the object property nodes
-				const objectPropertyNodes = child.data.estree.body[0]?.declarations?.[0]?.init?.properties;
+				const objectPropertyNodes =
+					child.data.estree?.body[0]?.declaration?.declarations?.[0]?.init?.properties;
```
bluwy, line 38: The previous access was incorrect and never actually worked 😅 I should look into properly typing this in the future.

Princesseuh (commented):
Logic seems fine to me, but would it be possible to have a test? Perhaps a unit test using the plugin and checking the result.

bluwy:
Thanks, I'll add a test 👍 There's currently the `mdx-optimize.test.js` file but it's indeed a bit simple.

bluwy:
I added a unit test. It now reveals that it generates some cursed JSX code that has attribute values with quotes extending multiple lines, but it still seems to parse properly and we _don't have_ to manually escape the new lines, so leaving it like so for now.

Princesseuh (approved):

PR #10889, opened 2024-04-26
Preserve content modules properly in cache

matthewp:
## Changes

- In `contentCollectionsCache`, in order to preserve linkage to propagated styles we need to wait until after the build to copy the content modules to `node_modules/.astro/content/`. This is because the styles/scripts are set as a post:build step, so copying the files over in `writeBundle` is too late.
- Also fixes hoisted scripts. These had to be implemented separately.

## Testing

- New test added to run the build twice.
- Fixes the hoisted scripts tests. Previously these tests *were not* running under `experimental.contentCollectionsCache` by mistake. Fixing that mistake revealed that hoisted scripts had not been implemented, so they are now and tests updated.
- Got rid of dev mode and ssr tests since the feature is SSG build only.
- Works in docs:
  - https://github.com/withastro/docs/pull/8162
  - https://docs-3gxhogy2w-astrodotbuild.vercel.app/en/getting-started/

## Docs

N/A, bug fix


matthewp:
!preview cc-preserve-cache

packages/astro/src/content/runtime.ts
```diff
@@ -44,15 +44,16 @@ export function createCollectionToGlobResultMap({
 	return collectionToGlobResultMap;
 }
 
-const cacheEntriesByCollection = new Map<string, any[]>();
```
matthewp, line 47: This was creating global state making it impossible to run the same fixture test multiple times with different config.

PR #10894, opened 2024-04-28
Add config option to serve all stylesheets in all pages (for SPA-like apps)

rellfy:
## Changes

For context, see https://github.com/withastro/roadmap/discussions/909

## Testing

This feature was only tested manually.

## Docs

Comments were added to the new config option.


rellfy:
> What about disabling [vite.build.cssCodeSplit](https://vitejs.dev/config/build-options.html#build-csscodesplit)? I think it should also work in Astro.

@bluwy That does work! but only for builds, it does not seem to work with the `astro dev` server.

I was able to get it working for builds previously as an integration (calling the same  `mergeSpaStylesheets` function in this PR), but it's great that a Vite setting can be set to handle this instead. I definitely missed that from the docs.

I could modify the PR to remove the custom post-build functionality, instead setting the Vite build config based on the new config setting. Also, might be a good idea to rename from `build.mergeSpaStylesheets` to `build.cssCodeSplit` so that it matches Vite's, but also applies to the local dev server. 

i.e. if `build.cssCodeSplit` is set to true, then `vite.build.cssCodeSplit` would also be set to true during builds.

rellfy:
I've modified the PR as per my comments above

bluwy:
Thanks for trying that out. I'd recommend that we discuss a bit more about the feature first before making more commits, at the risk that this might not be accepted 😅

For me, I don't quite understand why we also need to merge them in dev. Currently the crawl is only meant to prevent FOUC in dev. The links/styles need to be in a specific format in the HTML for Vite to not double-inject the links/styles.

I think the issue described in your proposal could be resolved with a different way? It's similar to the View Transition router we have now, where when we navigate to a new page, we fetch the new HTML and load the styles in the current page. Which gives an SPA-like experience too. Could enabling View Transitions in your app also help with this?

rellfy:
@bluwy thanks for the suggestion.

> For me, I don't quite understand why we also need to merge them in dev. Currently the crawl is only meant to prevent FOUC in dev. The links/styles need to be in a specific format in the HTML for Vite to not double-inject the links/styles.

If they are not merged in dev, then when I run `astro dev` in my Astro/HTMX app, and have a HTMX component load a different page for me (without reloading the entire  page in the browser) that page will not be styled at all. So for me to test the correct behaviour I would have to run `astro build && astro preview` instead, which is disappointing.

I think it is overkill to have to enable a client-side library (View Transition) just to get this to work in the `astro dev` server. With this PR I can simply enable an option and get this use case to work in both the dev server and for production builds.

I understand this might not get merged but I do believe it is a legitimate use case for HTMX users that don't want to deal with global CSS styles and managing them manually. I will prepare a template / minimum reproducible example to showcase this later and I'll link it here.

If there is an existing option in Astro to achieve this in both dev and build by simply toggling some setting, that would be great and exactly what I'm looking for. But I'd rather not add user code like View Transitions if it is not strictly necessary.

bluwy:
I'm not familiar with HTMX, but about:

> have a HTMX component load a different page for me (without reloading the entire page in the browser) that page will not be styled at all

If it's loading the entire page, shouldn't it also inject the related style/links from the other page, into the current page? Usually scripts that makes MPA feel like SPAs would handle this too. I think this issue can be solved externally, and without needing to change Astro.

rellfy:
@bluwy I worded myself poorly regarding HTMX. HTMX does not load or redirect to a new page, it actually fetches some HTML snippet from the server and replaces that without ever leaving the page. With this functionality, parts of the page can be swapped similarly to how a client-side router would behave, except there is no virtual DOM as all of the data comes from the server as HTML. But this requires all of the CSS to be pre-loaded in the initial page load, which is something that the Astro dev server would have to allow, otherwise it only works when previewing a production build -- hence my PR / fork to get this to work for my use case.

rellfy:
I made a minimal reproducible example for facilitating the discussion: https://github.com/rellfy/astro-htmx-css-mre

ematipico (changes requested):
Blocking for now, because we usually are a bit conservative about adding new features without a discussion and an RFC. 

Also, the PR doesn't have a changeset.

florian-lefebvre:
We think you can probably reproduce this behavior in userland, through an integration and/or vite plugin. It would be odd to add this feature to core for a single not so common usecase at the moment, so we have to decline your PR unfortunately.

Thanks for the time your put into this and I'm sure you'll be able to reuse most of it for your specific usecase! If your roadmap proposal gets more traction, we may implement it in core fyi.

PR #10898, opened 2024-04-28
chore(astro): move error-map

florian-lefebvre:
## Changes

- Moves the error map from content to core errors to prepare #10892

## Testing

Tests should still pass

## Docs

N/A, no changeset


Princesseuh:
Could we maybe rename it to `zod-error-map`? It's not immediately obvious from the name of the file what it does.

PR #10906, opened 2024-04-29
feat: Add new component to toolbar library

Princesseuh:
## Changes

Adds a new radio checkbox component and a way to make rounded buttons, for things like floating pluses UI and stuff

These two were some of the components that were supposed to be added when the toolbar was first released, but fell through as they were not needed for the built-in components.

## Testing

Tested manually

## Docs

Working on it!


bluwy (approved):

sarah11918 (approved):
Approving for docs!  🥳

PR #10917, opened 2024-04-30
fix: don't include port twice from x-forwarded-host and x-forwarded-port headers

### LINKED ISSUE #10916: Node SSR breaks when `X-Forwarded-Host` includes port and `X-Forwarded-Port` is also sent
### Astro Info

```block
4.5.14 still works, 4.5.15 and greater doesn't. Regression happens in https://github.com/withastro/astro/pull/10666.

Node                     v21.7.3
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

When astro is behind a reverse proxy that sets `X-Forwarded-Host: somehost:8080` and also `X-Forwarded-Port: 8080`, astro will internally build a URL that looks like `somehost:8080:8080` and fail with a 500 internal server error. This was introduced in https://github.com/withastro/astro/pull/10666.

Handle the request normally, without the 500.

stackblitz doesn't load for me right now

```sh
npm create astro@latest
cd repro
pnpm astro add node
pnpm run build
node dist/server/entry.mjs &

curl localhost:4321 -H 'X-Forwarded-Host: hostname:1234' -H 'X-Forwarded-Port: 1234'
Internal Server Error
```

### Participation

jakobhellermann:
fixes #10916 

## Changes

- Fix internal server error when bot `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` are set

## Testing

add another test next to the existing URL tests for `X-Forwarded-*`

## Docs

The docs shouldn't need to be changed since this is just a bugfix and now works as expected.

### DIFF
--- a/.changeset/forty-wolves-turn.md
+++ b/.changeset/forty-wolves-turn.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix internal server error when bot `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` are set
--- a/.changeset/forty-wolves-turn.md
+++ b/.changeset/forty-wolves-turn.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fix internal server error when bot `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` are set
+Fixes a case where the local server would crash when the host also contained the port, eg. with `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` headers
--- a/packages/astro/src/core/app/node.ts
+++ b/packages/astro/src/core/app/node.ts
@@ -66,7 +66,11 @@ export class NodeApp extends App {
 		const hostname =
 			req.headers['x-forwarded-host'] ?? req.headers.host ?? req.headers[':authority'];
 		const port = req.headers['x-forwarded-port'];
-		const url = `${protocol}://${hostname}${port ? `:${port}` : ''}${req.url}`;
+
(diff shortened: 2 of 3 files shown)


matthewp (approved):
Great, thanks!

matthewp:
Looks like this does break some tests 🤔

packages/astro/src/core/app/node.ts
```diff
@@ -66,7 +66,11 @@ export class NodeApp extends App {
 		const hostname =
 			req.headers['x-forwarded-host'] ?? req.headers.host ?? req.headers[':authority'];
 		const port = req.headers['x-forwarded-port'];
-		const url = `${protocol}://${hostname}${port ? `:${port}` : ''}${req.url}`;
+
+		const portInHostname = typeof hostname === "string" && typeof port === "string" && hostname.endsWith(port);
+		const hostnamePort = portInHostname ? hostname : `${hostname}:${port}`;
```
davidborzek, line 71: This should be ```
const hostnamePort = portInHostname ? hostname : `${hostname}${port ? `:${port}` : ''}`
``` because `port` can still be undefined.

File: .changeset/forty-wolves-turn.md [hunk 2]
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix internal server error when bot `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` are set
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where the local server would crash when the host also contained the port, eg. with `X-Forwarded-Host: hostname:8080` and `X-Forwarded-Port: 8080` headers
```

florian-lefebvre (approved):

matthewp (approved):

PR #10918, opened 2024-04-30
Provide a better error for when the app token is missing in CI

matthewp:
## Changes

- Uses `ci-info` which seems to be the most popular library for detecting CI environments.
- Provide a message about add the ASTRO_STUDIO_APP_TOKEN, rather than the previous message which said to run `astro login`.

## Testing

- None yet

## Docs

N/A, bug fix


bholmesdev (approved):
ci think this is a good change!

PR #10919, opened 2024-04-30
[db] Fix duplicate calls to recreate tables on startup

bholmesdev:
## Changes

This simplifies table initialization by pulling table creation and seeding out of Vite.

**Before**
- Buildtime: `astro:db` is loaded
- Runtime: `astro:db` calls `recreateTables` with top-level `await`.
- Runtime: `astro:db` calls `seedLocal()` with top-level `await`. Seed files are loaded as Vite modules.

**After**
- Buildtime: `astro:db` is loaded
- Buildtime: `recreateTables` is called.
- Buildtime: seeding is called. Seed files are loaded using esbuild, the same as our `db execute` CLI command.

### Changes
- Move seeding from the Vite runtime to a buildtime call using esbuild
- Remove special `astro:db:seed` virtual module used for deduping. This dramatically simplifies the DB vite plugin by removing the `resolveId` special case.
- **Remove eager _reloading_ for seed files.** This code included an `invalidateModule()` call on the `astro:db` module, which introduced a race condition when unit testing. This removal means that, if you edit the seed file, we no longer "eagerly" re-seed the database; you still need to refresh your dev server to trigger a seed. We still have eager seeding on _dev startup_ though, so SQLite inspectors should still have a nice experience.
- Move `queries.js` out of runtime and into core. Less code shipped to SSR!


Princesseuh:
> Buildtime: seeding is called. Seed files are loaded using esbuild, the same as our db execute CLI command.

Does this mean you can't access things from the Vite part anymore inside seeding? ex: content collections using `astro:content`?

bholmesdev:
@Princesseuh This is correct, though this was in an inconsistent state before as it did not work with `db execute`. We can restore that functionality if we spin up a vite dev server similar to the content collection config. Do you consider that a blocker? Do you know if users are relying on astro content for local db?

Princesseuh:
> @Princesseuh This is correct, though this was in an inconsistent state before as it did not work with `db execute`. We can restore that functionality if we spin up a vite dev server similar to the content collection config. Do you consider that a blocker? Do you know if users are relying on astro content for local db?

I am myself 😅 Both for `astro:content` and `astro:assets`.

delucis (approved):
Approving in principle as #NWTWWHB — happy to let you figure out what makes sense w/r/t availability of other virtual modules. Seems nice to do if possible (maybe in a follow-up PR?)

Spotted a bit of unused code which I’m not sure why linting isn’t flagging?

packages/db/src/runtime/index.ts
```diff
@@ -139,7 +138,7 @@ export function normalizeDatabaseUrl(envDbUrl: string | undefined, defaultDbUrl:
 		if (envDbUrl.startsWith('file://')) {
 			return envDbUrl;
 		}
-		return new URL(envDbUrl, pathToFileURL(process.cwd())).toString();
+		return new URL(envDbUrl, pathToFileURL(addTrailingSlash(process.cwd()))).href;
```
matthewp, line 141: Didn't you say you found out that `.href` excluded `file://` in some cases? According to https://developer.mozilla.org/en-US/docs/Web/API/URL/toString that should be the same...

bholmesdev: Ah yes, I also found that they are the same. The change to `.href` wa just left from testing and I can remove. However, that “add trailing slash” WAS needed for relative paths to resolve correctly. Added a unit test for that.

Princesseuh: Note that a previous PR of mine already added a trailing slash, though I didn't use the util

bholmesdev: @Princesseuh oh, I didn't see a fix in Astro DB for this. Was this somehow removed?

Princesseuh: I think your branch is just out of date 😄

delucis: The PR in question: https://github.com/withastro/astro/pull/10925

bholmesdev: OH lol, you're right. I'll rebase

PR #10924, opened 2024-05-01
add error message for svg without dimensions

### LINKED ISSUE #10913: Astro glob throw unsupported files for svg
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

### Describe the Bug

Astro glob throw unsupported files for svg

```js
const svgs = await Astro.glob("../files/*.svg")
```

throws

`unsupported file type: undefined`

Detect and import svgs

https://stackblitz.com/edit/astro-vexf1b

Its-Just-Nans:
May fix #10913


Its-Just-Nans:
> I would suggested contributing those changes upstream in https://github.com/image-size/image-size, as opposed to modifying the vendored version we have.

Done. https://github.com/image-size/image-size/pull/404/files

but it's not fixing the current astro bug with svg

Its-Just-Nans:
Okay the major problem was that `lookup()` was throwing an error without a good explanation.

This PR fixes that

Its-Just-Nans:
@Princesseuh 
This PR now only modify the error messages!

packages/astro/src/assets/utils/remoteProbe.ts
```diff
@@ -37,8 +37,7 @@ export async function probe(url: string): Promise<ISize> {
 					return dimensions;
 				}
 			} catch (error) {
-				// This catch block is specifically for `sizeOf` failures,
-				// which might occur if the accumulated data isn't yet sufficient.
+				throw new Error(`Failed to determine the type or the size of the picture (${url}): ${error}`);
```
Princesseuh, line 40: This would break `inferSize` in some cases, the error is ignored on purpose here as the comment previously there said.

Its-Just-Nans: done

matthewp (approved):
lgtm

Its-Just-Nans (commented):
looks like @github-actions[bot] is blocking the merge

ematipico (approved):

PR #10931, opened 2024-05-02
change: NotificationPayload type

### LINKED ISSUE #10930: `NotificationPayload` type is not correct.
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

https://github.com/withastro/astro/blob/5248ed1fb35bc1407660a189f23155812787f9e4/packages/astro/src/runtime/client/dev-toolbar/helpers.ts#L1-L8

The `level` should be `warning`, not `warn`.
Now, if set `warn`, it will be the same notification as `error`.

type should be `warning`.

N/A

### Participation

ktym4a:
## Changes

close #10930 

## Testing

Tests should still pass.

## Docs

N/A

### DIFF
--- a/.changeset/silent-zoos-work.md
+++ b/.changeset/silent-zoos-work.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes `toggleNotification()`'s parameter type for the notification level not using the proper levels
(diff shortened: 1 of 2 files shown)


Princesseuh (approved):
Looks great, thank you!

Issue #10932, opened 2024-05-02
toolbar: `toggleState` is not working.


ktym4a:
### Astro Info

```block
Astro                    v4.7.0
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

If `toggleState` is the same as before this,
```ts
app.dispatchEvent(
    new CustomEvent("toggle-app", {
        detail: {
            state: false,
        },
    }),
);
```

App does not close when I call the following
```ts
app.toggleState({
    state: false,
});
```

### What's the expected result?

If the same behavior is assumed, app should close.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-um5tv1?file=app.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

If my understanding is correct, I will send PR.

ktym4a:
added repro

PR #10933, opened 2024-05-02
Fixes toggleState

Princesseuh:
## Changes

Oops, not using the right event.

Fix https://github.com/withastro/astro/issues/10932

## Testing

Tested manually

## Docs

N/A

### DIFF
--- a/.changeset/lovely-feet-greet.md
+++ b/.changeset/lovely-feet-greet.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes `app.toggleState` not working correctly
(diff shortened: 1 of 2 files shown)


ktym4a:
@Princesseuh 
I would like to ask one related question.
https://docs.astro.build/en/reference/dev-toolbar-app-reference/#togglestate
Is there a reason why the docs are `changeAppState`? Or should it be fixed?

Princesseuh:
> @Princesseuh I would like to ask one related question. https://docs.astro.build/en/reference/dev-toolbar-app-reference/#togglestate Is there a reason why the docs are `changeAppState`? Or should it be fixed?

Yes, looks like it's outdated! It was changed close to release, so we might've missed this.

ktym4a:
Okay, Thanks.
I will send PR for that to docs!

PR #10934, opened 2024-05-02
Simplify plain MDX nodes in optimize option

bluwy:
## Changes

Some MDX nodes are simply `<kbd>something</kbd>` which the optimizer bails out because it thinks that it's non-static. However, it works like a plain HTML element and should be static.

This PR tries to simplify these MDX nodes as normal hast `element` nodes so we can further optimize the strings.

## Testing

Added a new test

## Docs

Added a changeset


delucis (approved):
LGTM!

PR #10938, opened 2024-05-02
feat(preact): add support for devtools

florian-lefebvre:
## Changes

- Adds a flag to enable https://preactjs.github.io/preact-devtools/ in dev mode

## Testing

Manually in the preact example

## Docs

[TBD](https://github.com/withastro/docs/pull/8164)


Princesseuh:
Should all these tools be opt-in? Is there a reason to not want them?

PR #10939, opened 2024-05-02
feat(react): add support for devtools

florian-lefebvre:
I feel like this is very hacky so I understand if we don't want it in the official integration

## Changes

- Adds a flag to enable https://react.dev/learn/react-developer-tools in dev mode

## Testing

Manually in the react example

## Docs

TBD

### DIFF
--- a/packages/integrations/react/src/index.ts
+++ b/packages/integrations/react/src/index.ts
@@ -1,14 +1,16 @@
 import react, { type Options as ViteReactPluginOptions } from '@vitejs/plugin-react';
-import type { AstroIntegration } from 'astro';
+import type { AstroConfig, AstroIntegration } from 'astro';
 import { version as ReactVersion } from 'react-dom';
-import type * as vite from 'vite';
+import { type Plugin } from 'vite';
+import { type ChildProcess, spawn } from 'node:child_process';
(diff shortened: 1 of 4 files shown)


packages/integrations/react/src/index.ts
```diff
@@ -94,18 +96,51 @@ export default function ({
...
-				if (command === 'dev') {
+				if (isDev) {
 					const preamble = FAST_REFRESH_PREAMBLE.replace(`__BASE__`, '/');
 					injectScript('before-hydration', preamble);
+					if (devtools) {
+						injectScript(
+							'head-inline',
+							`
+						const script = document.createElement("script");
+						script.src = "http://localhost:8097";
```
bholmesdev, line 122: Erm, is this an expected port for the devtools chrome extension? Very surprised to see this 😅 Definitely worth a detailed comment if so

florian-lefebvre: Yes I believe it's expected for non supported browsers: https://react.dev/learn/react-developer-tools#safari-and-other-browsers. The docs mention having to insert a script tag but since we can't currently with the integration API, I create the script programmatically. Again, I don't mind not shipping this thing at all because it looks hacky to me

Issue #10940, opened 2024-05-02
React 19 Beta throws error when starting astro dev server


oemer-aran:
### Astro Info

```block
astro info
Astro v4.7.1
Noda v18.18.0
System Linux (x64)
Package Manager unknown
Output static
Adapter none
Integrations @astrojs/react
```

### Describe the Bug

Using React 19 Beta results in following error:
```bash
[ERROR] Missing "./server.js" specifier in "react-dom" package
  Stack trace:
    at C:\dev\project-name\node_modules\.pnpm\vite@5.2.11_@types+node@20.12.8\node_modules\vite\dist\node\chunks\dep-cNe07EU9.js:47597:25
    [...] See full stack trace in the browser, or rerun with --verbose.
```
I followed the official React 19 Beta installation guide: https://react.dev/blog/2024/04/25/react-19-upgrade-guide

If the stackblitz doesn't not show the error and just a blank screen, try `npm install --force` or `yarn install`.

### What's the expected result?

Astro working with React 19 without any error.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-2hppzz?file=package.json,.vscode%2Flaunch.json,src%2Fpages%2Findex.astro&title=Astro%20Starter%20Kit:%20Basics

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
We don't currently support 19 which is currently in beta. I'll see what it takes to allow it.

zhondori:
@matthewp hello Matthew. I'm also encountering this issue with my project. But error is little different.

```
 error   Missing "./errors" specifier in "astro" package
 ```

tparsons-godaddy:
@zhondori I have the same error. Have you seen any updates?

PR #10942, opened 2024-05-03
Support React 19

matthewp:
## Changes

- Support React 19
- Fixes https://github.com/withastro/astro/issues/10940

## Testing

- https://stackblitz.com/edit/withastro-astro-yerxxk

## Docs

N/A, bug fix


matthewp:
!preview react-19

packages/integrations/react/package.json
```diff
@@ -62,8 +62,8 @@
   "peerDependencies": {
     "@types/react": "^17.0.50 || ^18.0.21",
```
bholmesdev, line 63: Should we add the `@types/react` preview package as well? They are listed as `npm:types-react@beta` and `npm:types-react-dom@beta`. Not sure if the `npm:` prefix works with peer deps though

bholmesdev (commented):
Note on React 19 types. From my testing, I also found that we use a server rendering API that is no longer available in the base React 19 package: `renderToReadableStream()`. I needed to update [this import](https://github.com/withastro/astro/blob/main/packages/integrations/react/server.js#L3) from `react-dom/server` to `react-dom/server.browser` to access. Otherwise, React defaults to Node and only exposes `rendertoPipeableStream()` from this package. Can you confirm you see the same issue in a test React 19 project?

packages/integrations/react/package.json
matthewp: I don't know how to semver this, is it:

```
"@types/react": "^17.0.50 || ^18.0.21 || npm:types-react@beta"
```

I admit this is outside of my wheelhouse

matthewp:
@bholmesdev The logic seems to fallback to `renderToPipeableStreamAsync` so I think it works fine, see: https://github.com/withastro/astro/blob/befbda7fa3d712388789a5a9be1e0597834f86db/packages/integrations/react/server.js#L109-L113

.changeset/short-phones-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/react": patch
+---
+
+Support React 19
```
matthewp, line 5:
```suggestion
Updates package to support React 19 beta
```

matthewp:
!preview react-19

bholmesdev:
@matthewp Ah okay, my guess is the react 17 conditional was triggering the wrong function. I see that was refactored here

packages/integrations/react/package.json
bholmesdev: Yes, that looks correct. I tested locally to confirm whether npm will raise an exception resolving peer deps, and it unfortunately does. I think we need to add this for everything to install correctly

bluwy (approved):

ematipico (approved):

ematipico:
I added this to the 4.8 milestone @matthewp , I hope it's fine

bholmesdev (approved):
Pushed a fix for the `@types` peer dep. LGTM now

matthewp:
Thanks @bholmesdev !

PR #10946, opened 2024-05-03
Avoid importing vite in db utils

delucis:
## Changes

- Reorganises some Astro DB code to avoid Vite being imported by `@astrojs/db/utils`

## Testing

- Existing tests should pass
- Is there a way we can enforce or check Vite is not imported?

## Docs

n/a — bug fix

### DIFF
--- a/packages/db/src/core/schemas.ts
+++ b/packages/db/src/core/schemas.ts
@@ -4,7 +4,6 @@ import { type ZodTypeDef, z } from 'zod';
 import { SERIALIZED_SQL_KEY, type SerializedSQL } from '../runtime/types.js';
 import { errorMap } from './integration/error-map.js';
 import type { NumberColumn, TextColumn } from './types.js';
-import { mapObject } from './utils.js';
 
 export type MaybeArray<T> = T | T[];
 
@@ -245,3 +244,17 @@ export const dbConfigSchema = z
 			}),
 		};
 	});
+
+
+/**
+ * Map an object's values to a new set of values
+ * while preserving types.
+ */
+function mapObject<T, U = T>(
+	item: Record<string, T>,
+	callback: (key: string, value: T) => U
+): Record<string, U> {
+	return Object.fromEntries(
+		Object.entries(item).map(([key, value]) => [key, callback(key, value)])
+	);
+}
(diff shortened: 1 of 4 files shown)


packages/db/src/utils.ts
```diff
@@ -1,8 +1,12 @@
-export { defineDbIntegration } from './core/utils.js';
+import type { AstroIntegration } from 'astro';
 import { tableSchema } from './core/schemas.js';
```
matthewp, line 2: This is still importing from `core/schemas.js` which imports `core/utils.js` which imports vite.

The rule is that `runtime` should not import inside in `core`. I don't know if this problem is solvable in this way.

delucis: `core/schemas` no longer imports Vite — I removed that.

packages/db/src/core/schemas.ts
```diff
@@ -4,7 +4,6 @@ import { type ZodTypeDef, z } from 'zod';
 import { SERIALIZED_SQL_KEY, type SerializedSQL } from '../runtime/types.js';
 import { errorMap } from './integration/error-map.js';
 import type { NumberColumn, TextColumn } from './types.js';
-import { mapObject } from './utils.js';
```
delucis, line 7: `core/utils` imports Vite so we want to avoid importing it — moved `mapObject` inline into this file as it is only used here

packages/db/src/core/utils.ts
```diff
@@ -28,21 +27,4 @@ export function getDbDirectoryUrl(root: URL | string) {
 	return new URL('db/', root);
 }
 
-export function defineDbIntegration(integration: AstroDbIntegration): AstroIntegration {
-	return integration;
-}
```
delucis, line 33: Moved this straight to the user-facing utils library to avoid importing it from here where Vite is also imported

packages/db/src/utils.ts
matthewp: Yeah, I see that now, but this could easily be broken by a new PR that does need something else from utils. We try to stick with the rule that runtime code only imports runtime code and not core, to prevent mistakes.

PR #10954, opened 2024-05-05
feat(env): server/secret variables

florian-lefebvre:
> Depends on #10881

## Changes

- Moves templates (stubs) and some
- Add feature validation
- Add `getEnv` to `RenderContext`
- Rework how astro:env templates are managed
- Add runtime code

## Testing

Cleans a test i forgot and add one to test `getSecret` with the `testAdapter`

## Docs

N/A


packages/astro/test/test-adapter.js
```diff
@@ -57,15 +64,14 @@ export default function (
 															? `request[Symbol.for('astro.clientAddress')] = '0.0.0.0';`
 															: ''
 													}
-													return super.render(request, routeData, locals);
+													return super.render(request, { routeData, locals, getEnv });
```
florian-lefebvre, line 67: This make a lof of tests fail

packages/astro/test/test-adapter.js
```diff
@@ -37,35 +39,36 @@ export default function (
...
 											class MyApp extends App {
 												#manifest = null;
 												constructor(manifest, streaming) {
 													super(manifest, streaming);
 													this.#manifest = manifest;
 												}
 
-												async render(request, routeData, locals) {
+												async render(request, {routeData, locals} = {}) {
+												console.log("test adapter");
```
bholmesdev:
```suggestion

```

bholmesdev (commented):
Approved since none of my comments are blocking. This API will be 4.8's `secret` sauce

bholmesdev (approved):

packages/astro/src/core/app/index.ts
```diff
@@ -65,6 +67,9 @@ export interface RenderOptions {
 	 * Default: `app.match(request)`
 	 */
 	routeData?: RouteData;
+
+	/** TODO: */
```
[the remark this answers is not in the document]
florian-lefebvre: I've let a bunch of todos already regarding docs so i'll tackle this in the docs pr if you think it's okay?

packages/astro/src/env/constants.ts
```diff
@@ -1,7 +1,13 @@
 export const VIRTUAL_CLIENT_MODULE_ID = 'astro:env/client';
-export const RESOLVED_VIRTUAL_CLIENT_MODULE_ID = '\0astro:env/client';
+export const RESOLVED_VIRTUAL_CLIENT_MODULE_ID = `\0${VIRTUAL_CLIENT_MODULE_ID}`;
 export const VIRTUAL_SERVER_MODULE_ID = 'astro:env/server';
-export const RESOLVED_VIRTUAL_SERVER_MODULE_ID = '\0astro:env/server';
+export const RESOLVED_VIRTUAL_SERVER_MODULE_ID = `\0${VIRTUAL_SERVER_MODULE_ID}`;
+export const VIRTUAL_INTERNAL_MODULE_ID = 'virtual:astro:env/internal';
+export const RESOLVED_VIRTUAL_INTERNAL_MODULE_ID = `\0${VIRTUAL_INTERNAL_MODULE_ID}`;
```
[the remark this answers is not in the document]
florian-lefebvre: Tried something, let me know if that sounds better to you!

packages/astro/src/@types/astro.ts
```diff
@@ -2428,6 +2428,11 @@ export type AstroFeatureMap = {
 	 * List of features that orbit around the i18n routing
 	 */
 	i18nDomains?: SupportsKind;
+
+	/**
+	 * The adapter is able to support `astro:env`
```
ematipico, line 2433: We might want to be more specific. It's not a blocker, however the support is for `getEnv`, right?

packages/astro/src/core/app/index.ts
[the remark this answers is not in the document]
ematipico: I think what Ben is saying is that we don't know what the `TODO` is about, is having something like `TODO: add docs in a later PR` is more helpful

packages/astro/src/core/app/index.ts
```diff
@@ -307,6 +317,9 @@ export class App {
 		if (!routeData) {
 			return this.#renderError(request, { locals, status: 404 });
 		}
+		if (getEnv) {
+			setGetEnv(getEnv);
```
ematipico, line 321: Shouldn't we do this only if the experimental flag is turned on? In order to retrieve the experimental flag from SSR, you'll have to add a new property to the `SSRManifest` type

packages/astro/src/integrations/astroFeaturesValidation.ts
```diff
@@ -88,6 +90,14 @@ export function validateSupportedFeatures(
 		}
 	}
 
+	validationResult.env = validateSupportKind(
+		env,
+		adapterName,
+		logger,
+		'astro:env',
+		() => config?.experimental?.env?.schema !== undefined
+	);
```
ematipico, line 99: The validation should be triggered only `config?.experimental?.env` isn't `undefined`. Otherwise, all adapters will start logging the warning even though a user didn't enable the feature.

packages/astro/src/runtime/server/astro-env.ts
```diff
@@ -0,0 +1,22 @@
...
+
+let _getEnv: GetEnv = (key) => process.env[key];
+
+export function setGetEnv(fn: GetEnv) {
+	_getEnv = fn;
+}
+
+export function getEnv(...args: Parameters<GetEnv>) {
+	return _getEnv(...args);
+}
```
ematipico, line 13: Maybe I should comment on the RFC, so take this as a suggestion and not a blocker: you should consider making `getEnv` an async function.

There are some environments where retrieving a secret is an async operation, such as AWS. Their SDK provides a package, and their APIs are callbacks/promises.

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -127,29 +152,75 @@ function validatePublicVariables({
...
-	const contentParts: Array<string> = [];
-	const dtsParts: Array<string> = [];
+	let module = '';
+	let types = '';
 
-	for (const { key, type, value } of validatedVariables.filter((e) => e.context === context)) {
-		contentParts.push(`export const ${key} = ${JSON.stringify(value)};`);
-		dtsParts.push(`export const ${key}: ${type};`);
+	for (const { key, type, value } of validatedVariables.filter((e) => e.context === 'client')) {
+		module += `export const ${key} = ${JSON.stringify(value)};	\n`;
```
bluwy, line 183: Really small nit: Would be nice to trim out the whitespaces before the `\n` and in the other code below.

packages/astro/src/runtime/server/astro-env.ts
bluwy: I'm not familiar with AWS, but would it be possible to retrieve those variables before hand, so we can have a sync version of this? e.g. they can be retrieved before calling `setGetEnv()`

ematipico: I suppose we should think of examples on how users can do that  using an adapter

florian-lefebvre: Should I make it async or not?

ematipico: As I mentioned at the beginning, it's not a blocker, so no need to change anything. I will add a comment in the RFC Let's keep it as is for now.

FYI, here's the secret manager SDK: https://www.npmjs.com/package/@aws-sdk/client-secrets-manager

packages/astro/src/@types/astro.ts
ematipico: `envGetter`? Even `env` is fine, as long as we're good at documenting it

ematipico (approved):

packages/astro/src/@types/astro.ts
florian-lefebvre: Will spend some time working on jsdoc 👍

bluwy (approved):

PR #10955, opened 2024-05-05
fix(astro): handle AstroUserError during sync and exports types

florian-lefebvre:
## Changes

- Handles `AstroUserError` hint during CC syncing
- Exports a few content types

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/sync/index.ts
+++ b/packages/astro/src/core/sync/index.ts
@@ -17,7 +17,7 @@ import { createNodeLogger } from '../config/logging.js';
 import { createSettings } from '../config/settings.js';
 import { createVite } from '../create-vite.js';
 import { collectErrorMetadata } from '../errors/dev/utils.js';
-import { AstroError, AstroErrorData, createSafeError, isAstroError } from '../errors/index.js';
+import { AstroError, AstroErrorData, AstroUserError, createSafeError, isAstroError } from '../errors/index.js';
 import type { Logger } from '../logger/core.js';
 import { formatErrorMessage } from '../messages.js';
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/sync/index.ts
```diff
@@ -159,9 +159,11 @@ export async function syncContentCollections(
 		if (isAstroError(e)) {
 			throw e;
 		}
+		const hint = AstroUserError.is(e) ? e.hint : undefined
```
Fryuni, line 162:
```suggestion
		const hint = AstroUserError.is(e) ? e.hint : AstroErrorData.GenerateContentTypesError.hint;
```

packages/astro/types/content.d.ts
```diff
@@ -26,7 +26,7 @@ declare module 'astro:content' {
 		| import('astro/zod').ZodDiscriminatedUnion<string, import('astro/zod').AnyZodObject[]>
 		| import('astro/zod').ZodIntersection<BaseSchemaWithoutEffects, BaseSchemaWithoutEffects>;
 
-	type BaseSchema =
+	export type BaseSchema =
```
Fryuni, line 29: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

Fryuni (approved):

bluwy (approved):
Seems fine to me, but I'd also like Erika's review on this if passing the hint from `AstroUserError` is normal. Seems to be the first place where we handle it specially.

ematipico (approved):

Princesseuh (approved):
It's a bit odd perhaps, but it's fine with me. The end result is better for the user!

PR #10959, opened 2024-05-06
Improve content collection styles and scripts build perf

bluwy:
## Changes

This PR refactors how we handle propagated styles and scripts in the Rollup `generateBundle` phase.

### Context
A propagated module, e.g.`/Users/.../blog.mdx?astroPropagatedAsset`, contains styles and scripts. The styles and scripts are only attached to the head if that `blog.mdx` is actually rendered, prevent other content styles and scripts from leaking.

### Before

 The data structure we used to track propagated modules to it styles and scripts is something like this:
```js
Map {  
  "/src/pages/index.astro" => Map {
    "/Users/.../blog1.mdx?astroPropagatedAsset" => { styles: Set {}, scripts: Set {} }
    "/Users/.../blog2.mdx?astroPropagatedAsset" => { styles: Set {}, scripts: Set {} } 
  },
  "/src/pages/other.astro" => Map {
    "/Users/.../blog1.mdx?astroPropagatedAsset" => { styles: Set {}, scripts: Set {} }
  }
}
```

This is great because we have information of which top-level pages uses what propagated modules and its styles and scripts. However, **(correct me if I'm wrong)** I noticed we're not using this information at all.

The only place we consume the information [is here](https://github.com/withastro/astro/blob/fd508a0fbb5148aafc180f1b14d3e47974777248/packages/astro/src/content/vite-plugin-content-assets.ts#L188-L210), but you can notice that both [propagated styles searching](https://github.com/withastro/astro/blob/fd508a0fbb5148aafc180f1b14d3e47974777248/packages/astro/src/core/build/plugins/plugin-css.ts#L141-L158) and [propagated scripts searching](https://github.com/withastro/astro/blob/fd508a0fbb5148aafc180f1b14d3e47974777248/packages/astro/src/core/build/plugins/plugin-analyzer.ts#L52-L70) do the same logic, so in practice the work done in all places are kinda negated.

### After
This PR simplifies and flattens it to:

```js
// internals.propagatedStylesMap
Map {
    "/Users/.../blog1.mdx?astroPropagatedAsset" => Set {}
    "/Users/.../blog2.mdx?astroPropagatedAsset" => Set {}
 }

// internals.propagatedScriptsMap
Map {
    "/Users/.../blog1.mdx?astroPropagatedAsset" => Set {}
    "/Users/.../blog2.mdx?astroPropagatedAsset" => Set {}
 }
```

This way there's only a single "propagated module to styles/scripts" map, each key deduplicated, and we can save the work to figure out which page uses what propagated module, which was very expensive.

This is actually the pattern that Content Collection Cache used, and seems like it's on the right track abstracting this. So this PR also unifies how it's handled with CCC.

### Performance
The goal of this refactor is to reduce calls to `getParentModuleInfos`, before it took 7.5s, now it takes 0.05s. Seems like making the work O(n^2) to O(n) is quite significant.

## Testing

Existing tests should pass. I think we have some comprehensive tests around this. I also clicked around the local docs build and all looks good as before.

## Docs

Added a changeset because I'm a bit afraid this may break some edge cases.


natemoo-re:
Oooh that's great, nice work identifying this! I wouldn't be surprised if there were a few other inefficient data structures lurking around this part of the codebase...

ematipico (approved):

matthewp (approved):

.changeset/grumpy-pillows-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Refactors internal styles and scripts handling for content collection to improve build performance
```
sarah11918, line 5:
```suggestion
Refactors internal handling of styles and scripts for content collections to improve build performance
```

sarah11918 (approved):
Looks great, @bluwy , and happy to have content collections improvements! Tiny tweak to the changeset mostly so I didn't have to do any grammar research.  😅  (And, collections plural). Otherwise, approved for docs!  🥳

Issue #8516, opened 2023-09-12
url encoded dynamic params not properly handled


y-nk:
### Astro Info

```block
~/projects/withastro-astro-vzlytn
❯ npx astro info
Astro                    v3.0.12
Node                     v16.20.0
System                   Linux (x64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/node
Integrations             none
```

### Describe the Bug

i've a url such as `/%5Bpage%5D` coming from a 3rd party place, but handling this valid url in `getStaticPaths` breaks the build.

```tsx
---
export function getStaticPaths() {
  return [{
    params: { path: undefined },
    props: {}
  }, {
    params: { path: '%5Bpage%5D' }, // ← minimal repro
    props: {}
  }]
}
---

<ul>
  <li>hello world</li>
  <li>
    <a href="/%5Bpage%5D">url encoded (should work)</a>
  </li>
  <li>
    <a href="/[page]">url decoded (should not work)</a>
  </li>
</ul>
```

Error log:
```
 prerendering static routes 
▶ src/pages/[...path].astro
  ├─ index.html (+7ms)
 error   A `getStaticPaths()` route pattern was matched, but no matching static path was found for requested path `/%5Bpage%5D`.
  Hint:
    Possible dynamic routes being matched: src/pages/[...path].astro.
  Error reference:
    https://docs.astro.build/en/reference/errors/no-matching-static-path-found/
  File:
    /home/projects/withastro-astro-vzlytn/node_modules/astro/dist/core/errors/errors.js:36:5
  Code:
    35 |   setFrame(source, location) {
    > 36 |     this.frame = codeFrame(source, location);
         |     ^
      37 |   }
      38 |   static is(err) {
      39 |     return err.type === "AstroError";
```

### What's the expected result?

it should not matter what chars are in a dynamic param as long as it computes to a valid url, and this one seems to be valid according to `new URL('https://example.com/%5Bpage%5D')`.

<img width="299" alt="image" src="https://github.com/withastro/astro/assets/411625/f535aa9c-a1ca-4a38-ad92-fd51c5fcb205">

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-vzlytn?file=src/pages/%5B...path%5D.astro

### Participation

- [X] I am willing to submit a pull request for this issue.

rsk700:
[...]

PR #10965, opened 2024-05-07
Update generator.ts to allow %23 (#) in dynamic urls

Elias-Chairi:
## Changes

- Allows the server to retrive # in dynamic paths (will need to be %23 on client to be recived)
- I do have a case in my project where this makes sense.

## Testing

- Tested locally and seems to be fine. I dont know why these checks are in place well enough.
- Didnt see this beeing a problem in my older project where I used react.

## Docs

- Dont think this need any new doc changes

### DIFF
--- a/packages/astro/src/core/routing/manifest/generator.ts
+++ b/packages/astro/src/core/routing/manifest/generator.ts
@@ -2,6 +2,11 @@ import type { AstroConfig, RoutePart } from '../../../@types/astro.js';
 
 import { compile } from 'path-to-regexp';
 
+/**
+ * Sanitizes the parameters object by normalizing string values and replacing certain characters with their URL-encoded equivalents.
+ * @param {Record<string, string | number | undefined>} params - The parameters object to be sanitized.
+ * @returns {Record<string, string | number | undefined>} The sanitized parameters object.
+ */
 function sanitizeParams(params: Record<string, string | number | undefined>): Record<string, string | number | undefined> {
 	return Object.fromEntries(
 		Object.entries(params).map(([key, value]) => {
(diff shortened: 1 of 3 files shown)


matthewp:
Please provide a test, thank you!

packages/astro/src/core/routing/manifest/generator.ts
```diff
@@ -2,6 +2,20 @@ import type { AstroConfig, RoutePart } from '../../../@types/astro.js';
 
 import { compile } from 'path-to-regexp';
 
+function sanitizeParams(params: Record<string, string | number | undefined>): Record<string, string | number | undefined> {
+	return Object.entries(params).reduce((acc, [key, value]) => {
```
ematipico, line 6: You could use `.map` instead of `.reduce`. It's easier to read, and you don't need to maintain an accumulator

Elias-Chairi: changed it to using map now. Considered simplifying like this, but final thought is that its not worth making an intermediate object:
```
function sanitizeParams(params: Record<string, string | number | undefined>): Record<string, string | number | undefined> {
	const sanitizedParams: Record<string, string | number | undefined> = {};

	Object.entries(params).forEach(([key, value]) => {
		if (typeof value === 'string') {
			sanitizedParams[key] = value
				.normalize()
				.replace(/#/g, '%23')
				.replace(/\?/g, '%3F');
		} else {
			sanitizedParams[key] = value;
		}

	})

	return sanitizedParams
}
```

y-nk:
@ematipico do you want me to close my #8516 issue?

PR #10974, opened 2024-05-08
feat(astro): experimental astro:env

florian-lefebvre:
## Changes

- An experimental release for `astro:env`. View RFC https://github.com/withastro/roadmap/pull/894
- Related testing PR for Cloudflare and Netlify https://github.com/withastro/adapters/pull/258

## What to review

I'm mostly looking for reviews on

- `astro:env/setup`: https://github.com/withastro/astro/pull/10974/files#diff-d08d0989a863f87bfceda052de858b8d03124b6bfbff8e084ffbef31240bb446, https://github.com/withastro/astro/pull/10974/files#diff-a95d41705ded14985c748b640b26f3226a7c37d714a8f2fe6171252a59e9941eR79-R81, https://github.com/withastro/astro/pull/10974/files#diff-f982c118ef9afa9c2d69b474b2e206e460b62068cccb0061a5b4eccfe81c708b
- Pipelines/manifest usage: https://github.com/withastro/astro/pull/10974/files#diff-484131dc8cd8613b15500c8d5920a253fd8a70ef55db44c11de7ae35e3805792, https://github.com/withastro/astro/pull/10974/files#diff-53b52353e46ee04d0f157485e6511510f0b8779e7d6e8b0653d733ce4b56f9fa, https://github.com/withastro/astro/pull/10974/files#diff-faa73d40645eda14a766bb44509025ac3a3758731ae7f63d657b3cd230961e9f, https://github.com/withastro/astro/pull/10974/files#diff-0a19d4e7d5ce97887629902a09e7ecb9f4c5591c15ebcec95c28ceffc61cc3f7, https://github.com/withastro/astro/pull/10974/files#diff-0a19d4e7d5ce97887629902a09e7ecb9f4c5591c15ebcec95c28ceffc61cc3f7, https://github.com/withastro/astro/pull/10974/files#diff-449e05a2f96764b550a296f6016fa30a0f6727b679a7330497e9233c4fea9ac9, https://github.com/withastro/astro/pull/10974/files#diff-4dbd8dabc703c23edaa57597427f67214fd01a957888457a76a525808a252099
- Tests
- Docs: a bit everywhere with jsdocs and errors, hard to link specific files

Also worth reviewing the adapters pr to see how it will be used for real.

## Testing

Unit and integration tests

## Docs

Added changeset and config reference


packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -0,0 +1,226 @@
...
+// merged with the static types/content.d.ts
+
+interface AstroEnvVirtualModPluginParams {
+	settings: AstroSettings;
+	logger: Logger;
+	mode: 'dev' | 'build' | string;
+	fs: typeof fsMod;
+}
+
+export function astroEnvVirtualModPlugin({
```
bluwy, line 29: Small nit: I think we can drop the `VirtualMod` here and from the plugin name. Vite plugins don't usually need to share that.

packages/astro/src/env/vite-plugin-env-virtual-mod.ts
```diff
@@ -0,0 +1,226 @@
...
+	settings,
+	logger,
+	mode,
+	fs,
+}: AstroEnvVirtualModPluginParams): Plugin | undefined {
+	if (!settings.config.experimental.env) {
+		return;
+	}
+	logger.warn('env', 'This feature is experimental.');
```
bluwy, line 38: Do we need a warning here? If they opt into it in the config, I think it should be enough that they know it's experimental.

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
...
+// astro.config.mjs
+import { defineConfig, envField } from "astro/config"
+
+export default defineConfig({
+    experimental: {
+        env: {
+            schema: {
+                PUBLIC_API_URL: envField.string({ context: "client", access: "public", optional: true }),
+                PUBLIC_PORT: envField.number({ context: "server", access: "public", default: 4321 }),
+                API_SECRET: envField.string({ context: "server", access: "secret" }),
```
bluwy, line 19: Re-reading this, since `PUBLIC_` already hinted what the `access` would be, maybe we should remove `access` altogether from the expose API, and we can auto set it internally?

packages/astro/src/core/app/index.ts
```diff
@@ -300,6 +312,9 @@ export class App {
 		if (!routeData) {
 			return this.#renderError(request, { locals, status: 404 });
 		}
+		if (this.#manifest.experimentalEnv && getEnv) {
+			setGetEnv(getEnv);
+		}
```
bluwy, line 317: Is `setGetEnv` only possible to be called within `render()`? I feel like env vars should be global by default and this is only for Cloudflare? Runtimes like node and deno adapter should be able to set this once.

packages/astro/src/runtime/server/astro-env.ts
```diff
@@ -0,0 +1,22 @@
+import type { GetEnv } from '../../env/types.js';
+import { AstroError, AstroErrorData } from '../../core/errors/index.js';
+export { validateEnvVariable } from '../../env/validators.js';
+
+let _getEnv: GetEnv = (key) => process.env[key];
```
bluwy, line 5: I also think that this should try by default if `setGetEnv` wasn't called. Otherwise there's a risk that it's called in non-node runtimes and makes it hard to debug if the env are working.

ematipico: That's actually a great catch. Maybe we shouldn't default to `process.env` at all, and let it define in the adapter regardless

packages/astro/src/core/app/index.ts
florian-lefebvre: I looked a little bit at the adapter api and I guess we could also expose something on `App`? So adapters like node would be able to do:
```ts
const app = new App(manifest);
app.setGetEnv((key) => {
	// whatever logic
})
```
This would call the runtime internally. Wdyt @ematipico?

packages/astro/src/runtime/server/astro-env.ts
florian-lefebvre: Well I thought about that and I think it makes sense to keep it as is by default. In particular for static sites because we want all astro sites, no matter the output, to be able to use any kind of variable. If we don't provide a default then people could not use it. Alternatively, we could default to `process.env` only when the output is static from within the virtual module

.changeset/poor-berries-occur.md
florian-lefebvre: I quite like forcing people to be super explicit but I don't mind inferring it as well. I'd like other opinions before refactoring

ematipico: I would also prefer expressiveness over "magic"

packages/astro/src/@types/astro.ts
```diff
@@ -2601,6 +2688,11 @@ export type AstroFeatureMap = {
 	 * List of features that orbit around the i18n routing
 	 */
 	i18nDomains?: SupportsKind;
+
+	/**
+	 * The adapter is able to support `getSecret` exported from `astro:env/server`
+	 */
+	env?: SupportsKind;
```
florian-lefebvre, line 2695: Should rename to `envGetSecret`

.changeset/poor-berries-occur.md
bluwy: Alright then. Let's see how this goes.

packages/astro/src/runtime/server/astro-env.ts
bluwy: Shouldn't all kinds of build be responsible for calling `setGetEnv` so that they all work? Which means for static builds, we can also call it manually before rendering the pages.

florian-lefebvre:
!preview astro-env

florian-lefebvre:
!preview astro-env

florian-lefebvre:
!preview astro-env

florian-lefebvre:
!preview astro-env

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
+---
+"astro": minor
+---
+
+Adds experimental support for the `astro:env` API. It lets you define a type-safe schema for your environments variables, and where they should be available (on the server or the client).
```
dreyfus92, line 5:
```suggestion
Adds experimental support for the `astro:env` API. It lets you define a type-safe schema for your environment variables and specify their availability on the server or client.
```

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+})
+```
+
+There are currently 3 data types supported: strings, numbers and booleans.
+
+There are also 3 kinds of variables:
+
+- **Public client variables**: Those variables end up in your final client and server bundle and can be accessed both client and server through the `astro:env/client` module:
```
dreyfus92, line 30:
```suggestion
- **Public client variables**: These variables are included in both the client and server bundles and can be accessed from either via the astro:env/client module:
```

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
...
+
+    ```js
+    import { PUBLIC_API_URL } from "astro:env/client"
+    ```
+
+- **Public server variables**: Those variables end up in your final server bundle and can be accessed on the server through the `astro:env/server` module:
```
dreyfus92, line 36:
```suggestion
- **Public server variables**: These variables are included only in the server bundle and can be accessed on the server via the astro:env/server module:
```

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
...
+
+    ```js
+    import { PUBLIC_PORT } from "astro:env/server"
+    ```
+
+- **Secret server variables**: Those variables are not part of your final bundle and can be accessed on the server through the `astro:env/server` module:
```
dreyfus92, line 42:
```suggestion
- **Secret server variables**: These variables are not included in the final bundle and can only be accessed on the server via the astro:env/server module:
```

dreyfus92 (commented):
This looks great @florian-lefebvre! I've made a few rewording suggestions. Feel free to usr or modify as needed! 😁

florian-lefebvre:
!preview astro-env

sarah11918 (commented):
Yay @florian-lefebvre! This looks so great and is so cool!

Note: I have *not* edited the changeset (since so much of it will follow from the config docs). Let's get the docs part done and polished, then we'll know what we want represented in the changelog!

Some thoughts below!

sarah11918 (commented):
Yay @florian-lefebvre! This looks so great and is so cool!

Note: I have *not* edited the changeset (since so much of it will follow from the config docs). Let's get the docs part done and polished, then we'll know what we want represented in the changelog!

Some thoughts below!

packages/astro/src/@types/astro.ts
```diff
@@ -2054,6 +2055,101 @@ export interface AstroUserConfig {
...
+		 * Enables experimental `astro:env` features .
+		 *
+		 * The `astro:env` API lets you configure a type-safe schema for your environment variables, and indicate whether they should be available on the server or the client. Import and use your defined variables from the appropriate `/client` or `/server` module:
+		 *
+		 * ```js
+		 * ---
+		 * import { PUBLIC_POKEAPI } from "astro:env/client"
+		 * 
+		 * const data = fetch(`${PUBLIC_POKEAPI}/pokemon/squirtle`);
+		 * ---
```
florian-lefebvre, line 2076:
```suggestion
		 * ---
		 * import { PUBLIC_APP_ID } from "astro:env/client"
		 * import { PUBLIC_API_URL, getSecret } from "astro:env/server"
		 * const API_TOKEN = getSecret("API_TOKEN")
		 * 
		 * const data = await fetch(`${PUBLIC_API_URL}/users`, {
		 * 	method: "POST",
		 * 	headers: {
		 * 		"Content-Type": "application/json",
		 * 		"Authorization": `Bearer ${API_TOKEN}`
		 * 	},
		 * 	body: JSON.stringify({ appId: PUBLIC_APP_ID })
		 * })
		 * ---
```
@sarah11918 this would more realistic and show the usage of every type of variable, wdyt?

packages/astro/src/@types/astro.ts
```diff
@@ -2054,6 +2055,101 @@ export interface AstroUserConfig {
...
+		 *     import { PUBLIC_API_URL } from "astro:env/client"
+		 *     ```
+		 * 
+		 * - **Public server variables**: These variables end up in your final server bundle and can be accessed on the server through the `astro:env/server` module:
+		 * 
+		 *     ```js
+		 *     import { PUBLIC_PORT } from "astro:env/server"
+		 *     ```
+		 * 
+		 * - **Secret server variables**: These variables are not part of your final bundle and can be accessed on the server through the `astro:env/server` module:
```
sarah11918, line 2112:
```suggestion
		 * - **Secret server variables**: These variables are not part of your final bundle and can be accessed on the server through the `getSecret()` helper function available from the `astro:env/server` module:
```
Since this one is a little different, it's probably worth mentioning that you don't import it directly, but rather use the helper.

packages/astro/src/@types/astro.ts
```diff
@@ -2054,6 +2055,101 @@ export interface AstroUserConfig {
...
+		 * 
+		 *     ```js
+		 *     import { getSecret } from "astro:env/server"
+		 * 
+		 *     const API_SECRET = getSecret("API_SECRET") // typed
+		 *     const SECRET_THAT_MAY_EXIST = getSecret("SECRET_THAT_MAY_EXIST") // string | undefined
```
sarah11918, line 2118:
```suggestion
		 *     const SECRET_THAT_MAY_EXIST = getSecret("SECRET_THAT_MAY_EXIST") // string | undefined
```
I have to admit, I'm not clear what this does, or what "that may exist" means. Checking the RFC, it looks like this is what happens when something is not specified in the schema.

a) Are people going to be using `getSecret` for something NOT defined in their schema? If so when/why? What would this actually DO for people? I'm not sure of the use case for this.
b) Do we need to SHOW using `getSecret` for something not in their schema? Is this a case of "not everything that is true belongs in docs" or is this a very common thing people need to know how to do?

packages/astro/src/@types/astro.ts
```diff
@@ -2054,6 +2055,101 @@ export interface AstroUserConfig {
...
+		 *     ```
+		 * 
+		 * **Note:** Secret client variables are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
+		 * For a complete overview, and to give feedback on this experimental API, see the [Astro Env RFC](https://github.com/withastro/roadmap/blob/feat/astro-env-rfc/proposals/0046-astro-env.md).
```
sarah11918, line 2122:
```suggestion
		 * **Note:** Secret client variables are not supported because there is no safe way to send this data to the client. Therefore, it is not possible to configure both `context: "client"` and `access: "secret"` in your schema.
		 *
		 * For a complete overview, and to give feedback on this experimental API, see the [Astro Env RFC](https://github.com/withastro/roadmap/blob/feat/astro-env-rfc/proposals/0046-astro-env.md).
```
Just nicer spacing!

packages/astro/src/@types/astro.ts
florian-lefebvre: Yeah the naming is probably bad, let me reexplain.

Basically in most cases, you'll want to define a server/secret var in your config schema for types, validation etc. But in some cases, it's also fine to get those purely dynamically. Maybe you allow a user to specify the key, maybe the key comes from a database...For all those cases, you'd want to use something like `process.env`.

But one of the point of Astro Env is to abstract that part so you can switch adapters without worrying. That's why you'll want to use `getSecret` with a key not defined in a schema

sarah11918: So what about something like

```
const DYNAMIC_SECRET = getSecret("DYNAMIC_SECRET") // string | undefined
```
Is that sufficient enough to make sense?

```
const SECRET_NOT_IN_SCHEMA = getSecret("SECRET_NOT_IN_SCHEMA") // string | undefined
```
??

Anything along those lines?

florian-lefebvre: I like the 2nd suggestion more since a secret is "dynamic" 😅

sarah11918: That looks good, ship it!

sarah11918 (approved):
Approving for docs with this one update!  🥳 

Nooooo wait! We have to do the changeset to match now! UNAPPROVE! lol

packages/astro/src/env/validators.ts
```diff
@@ -0,0 +1,82 @@
...
+	return {
+		valid: typeof input === 'string',
+		parsed: input,
+	};
+};
+
+const numberValidator: ValueValidator = (input) => {
+	const num = parseFloat(input ?? '');
+	return {
+		valid: !isNaN(num),
```
matthewp, line 36: Tested this implementation quite a bit and couldn't find a bug, good job!

packages/astro/src/env/vite-plugin-env.ts
```diff
@@ -0,0 +1,226 @@
...
+					serverPublic: serverTemplates.types.public,
+					serverSecret: serverTemplates.types.secret,
+				}),
+			});
+		},
+		buildEnd() {
+			templates = null;
+		},
+		resolveId(id) {
+			for (const moduleId of Object.values(VIRTUAL_MODULES_IDS)) {
```
matthewp, line 74: This could be optimized, probably should given how often this code is called. In the constants module:

```js
export const VIRTUAL_MODULE_ID_VALUES = new Set(Object.values(VIRTUAL_MODULES_IDS));
```

Then check with `VIRTUAL_MODULE_ID_VALUES.has(id)`, which is O(1).

packages/astro/src/env/vite-plugin-env.ts
```diff
@@ -0,0 +1,226 @@
...
+}: AstroEnvVirtualModPluginParams): Plugin | undefined {
+	if (!settings.config.experimental.env) {
+		return;
+	}
+	const schema = settings.config.experimental.env.schema ?? {};
+
+	let templates: { client: string; server: string; internal: string } | null = null;
+
+	return {
+		name: 'astro-env-virtual-mod-plugin',
```
bluwy, line 41: Perhaps we can rename this as `astro-env-plugin` to match the function name.

packages/astro/src/runtime/server/astro-env.ts
bluwy, line 1: I'm not sure if this file is in the right place. Maybe we can move it to `src/env/runtime.ts`? I think we keep this folder for the runtime of rendering Astro components only.

packages/astro/test/test-adapter.js
```diff
@@ -37,14 +50,27 @@ export default function (
 											import { App } from 'astro/app';
 											import fs from 'fs';
 
+											${
+												env
+													? `
+											await import('astro:env/setup')
+												.then(mod => mod.setGetEnv((key) => {
+													console.log({ key });
```
bluwy, line 58: Could we remove this `console.log`?

bluwy (commented):
Some small nits, but this looks great!

packages/astro/src/content/types-generator.ts
```diff
@@ -382,7 +382,7 @@ async function writeContentFiles({
 	let contentTypesStr = '';
 	let dataTypesStr = '';
 
-	const collectionSchemasDir = new URL('./collections/', contentPaths.cacheDir);
+	const collectionSchemasDir = new URL('./collections/', settings.dotAstroDir);
```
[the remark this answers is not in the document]
florian-lefebvre: All codegen related features need to create files in `.astro`. The path is not content specific so I extracted it to settings to be able to reuse it across content, actions and env. It will also be helpful when I get to work on https://github.com/withastro/roadmap/discussions/887

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,51 @@
+---
+"astro": minor
+---
+
+Adds experimental support for the `astro:env` API. It lets you define a type-safe schema for your environment variables and specify their availability on the server or client.
```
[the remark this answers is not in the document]
florian-lefebvre: Updated the changeset, ptal!

ematipico (approved):

.changeset/poor-berries-occur.md
```diff
@@ -0,0 +1,72 @@
...
+                API_SECRET: envField.string({ context: "server", access: "secret" }),
+            }
+        }
+    }
+})
+```
+
+There are currently 3 data types supported: strings, numbers and booleans.
+
+There are three kinds of variables, determined by the combination of `context` (`client` or `server`) and `access` (`private` or `public`) settings defined in your [`env.schema`](#experimentalenvschema):
```
sarah11918, line 47:
```suggestion
There are three kinds of environment variables, determined by the combination of `context` (`client` or `server`) and `access` (`private` or `public`) settings defined in your [`env.schema`](#experimentalenvschema):
```
I think for the changeset, we don't need to repeat the three kinds of data types, since we said them above. Also, I like explicitly "environment variables" instead of just variables here.

sarah11918 (approved):
Made one small suggestion, but I think otherwise the docs pass my checks! Nicely done @florian-lefebvre !

If either @ematipico or @matthewp has suggestions or comments for the technical accuracy/completeness of the docs, feel free to comment! But from an editing standpoint, I'm happy with what's there!

matthewp (approved):

dreyfus92 (approved):
Looking great so far! 😁

Issue #10307, opened 2024-03-03
Styles contained in SCSS class passed to <Picture /> component's pictureAttributes.class not being applied


Lper258:
### Astro Info

```block
Astro                    v4.4.8
Node                     v18.17.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

When passing `pictureAttributes` prop to Astro's `<Picture />` component, the passed class DOES get applied to the generated HTML`<picture />` tag, but it seems that the styles are not generate i.e. the class has no styles.

WARNING: Minimal reproducible example on StackBlitz works for me only in Chrome. Sometimes the image in the local src/assets/ directory is loaded, some times it is not, but the reported issue is still visible.

```astro
---
import headerImage from "../assets/homepage/header.png";

 headerImageStyles = {
  width: "100%",
  display: "flex",
  justifyContent: "center",
};
---

 <Picture
      src={headerImage}
      alt='Homepage title'
      class='header-picture' // This is applied correctly to the underlying <img /> tag
      formats={["avif", "webp"]}
      loading='eager'
      fallbackFormat='png'
      pictureAttributes={{
        style: headerImageStyles,
        class: "this-class-is-the-problem", // The issue is here
      }}
    />


<style lang="scss">

    .header-picture {
        width: 17rem;
        height: auto;

        margin-bottom: 16rem;
    }

    .header-picture-test {
        background-color: red;
    }
</style>
``` 

### What's the expected result?

The class gets applied to the generated `<picture />` tag, AND it contains the correct styles.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qrod97

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
I think this is expected since the scoped class/attribute is passed to the component directly, so `pictureAttributes` didn't received the extra scoped class/attribute that would've made it work. Seems like the scoped attribute works for the fallback `img` by default though. Maybe it's worth making this usecase work, but we'd need to special case some special attributes or classes to pass on to the `picture` or `img`.

matthewp:
I don't think we should do this, this is an issue any time you use a component and attempt to set its classes. The solution, like with markdown, is to use global CSS for that.

matthewp:
After talking about it, @bluwy thinks he might know of a way to fix this.

PR #10975, opened 2024-05-08
Support scope styles for picture element in Picture component

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/10307

The PR uses some tricks to get the scoped attributes or class to pass it to the `<picture>` element, instead of only to the `img` element.

The trick uses regexes for `astro-*` class and `data-astro-cid-*` attributes, which perhaps is not the most elegant, but I think it should be fine?

Question: Do we want to check for `data-astro-cid` specifically?

## Testing

Added test

## Docs

Added changeset

### DIFF
--- a/packages/astro/components/Picture.astro
+++ b/packages/astro/components/Picture.astro
@@ -27,6 +27,21 @@ if (props.alt === undefined || props.alt === null) {
 	throw new AstroError(AstroErrorData.ImageMissingAlt);
 }
 
+// Picture attribute inherit scoped styles from class and attributes
+const scopedStyleClass = props.class?.match(/\bastro-\w{8}\b/)?.[0]
+if (scopedStyleClass) {
+	if (pictureAttributes.class) {
+		pictureAttributes.class = `${pictureAttributes.class} ${scopedStyleClass}`;
+	} else {
+		pictureAttributes.class = scopedStyleClass;
+	}
+}
+for (const key in props) {
+	if (key.startsWith('data-astro-')) {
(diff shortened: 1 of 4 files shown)


ematipico (approved):
I am not sure this solution will be future-proof, especially in the future we will land https://github.com/withastro/roadmap/discussions/243

However, I don't know if there's a better solution, so I leave the final call to you

packages/astro/components/Picture.astro
```diff
@@ -27,6 +27,21 @@ if (props.alt === undefined || props.alt === null) {
...
+const scopedStyleClass = props.class?.match(/\bastro-\w{8}\b/)?.[0]
+if (scopedStyleClass) {
+	if (pictureAttributes.class) {
+		pictureAttributes.class = `${pictureAttributes.class} ${scopedStyleClass}`;
+	} else {
+		pictureAttributes.class = scopedStyleClass;
+	}
+}
+for (const key in props) {
+	if (key.startsWith('data-astro-cid')) {
```
matthewp, line 40: 😔

matthewp (approved):

bluwy:
Yeah if we want to support the feature request and not break this, we would probably need to support some kind of `Astro.scopedProps` API and restrict the hash config to string-form like `cssHash: 'myhash-[hash:8]'` in order to continue making this work.

I don't think this is perfect too, but I think it enables a nice usecase and we could try to revisit this again. I'll take the approvals as a green light though 😄

PR #10981, opened 2024-05-09
fix(types): Add name attribute to AnchorHTMLAttributes

mo:
## Changes

Add "name" attribute in the type definition for AnchorHTMLAttributes, allowing the creating of <a name="foo"></a> in astro jsx

## Testing

not tested

## Docs

no docs, already documented in HTML docs


mo:
@Princesseuh I hadn't realized that people used id="" these days :facepalm: ... but I guess this PR could still be merged since name="" still works and it's annoying to get warnings for valid code

jacobdalamb:
> @Princesseuh I hadn't realized that people used id="" these days 🤦 ... but I guess this PR could still be merged since name="" still works and it's annoying to get warnings for valid code

You can extend it yourself https://docs.astro.build/en/guides/typescript/#built-in-html-attributes

Princesseuh:
I'd normally suggest adding it yourself, but since this is not actually the first time someone asks for it, I think it's useful. Since it still works in browsers, I assume there's also probably some packages out there that use it too.

Anyway, it's free to add 😄

ematipico (approved):

PR #10986, opened 2024-05-09
feat(package.json): removed wrong react types

emish89:
## Changes

removed wrong react types that break install

### DIFF
--- a/packages/integrations/react/package.json
+++ b/packages/integrations/react/package.json
@@ -16,7 +16,7 @@
     "astro-component",
     "renderer",
     "react"
-  ],  
+  ],
   "bugs": "https://github.com/withastro/astro/issues",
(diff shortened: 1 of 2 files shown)


packages/integrations/react/package.json
```diff
@@ -11,12 +11,7 @@
     "url": "https://github.com/withastro/astro.git",
     "directory": "packages/integrations/react"
   },
-  "keywords": [
-    "astro-integration",
-    "astro-component",
-    "renderer",
-    "react"
-  ],
+  "keywords": ["astro-integration", "astro-component", "renderer", "react"],
```
bluwy, line 14: Can this be reverted?

emish89: yes, I don't know why it created it

emish89: reverted

ematipico (approved):

Issue #10984, opened 2024-05-09
`@astrojs/react` is broken on `npm install` command


mika-f:
### Astro Info

```block
Astro                    v4.8.0
Node                     v18.19.0
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Failed to install `@astrojs/react` of latest release.

```bash
npm run astro add react

> green-galaxy@0.0.1 astro
> astro add react

✔ Resolving packages...

  Astro will run the following command:
  If you skip this step, you can always run it yourself later

 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ npm install @astrojs/react@^3.3.3 @types/react@^null @types/react-dom@^null react@^18.3.1 react-dom@^18.3.1  │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

√ Continue? ... yes
✖ Installing dependencies...

 Command failed with exit code 1: npm install @astrojs/react@^3.3.3 @types/react@^null @types/react-dom@^null react@^18.3.1 react-dom@^18.3.1
npm ERR! code EINVALIDTAGNAME
npm ERR! Invalid tag name "^null" of package "@types/react@^null": Tags may not have any characters that encodeURIComponent encodes.

npm ERR! A complete log of this run can be found in: C:\Users\REDACTED\AppData\Local\npm-cache\_logs\2024-05-09T11_58_24_625Z-debug-0.log

Astro could not update your astro.config.js file safely.
Reason: Unable to install dependencies

You will need to add these integration(s) manually.
Documentation: https://docs.astro.build/en/guides/integrations-guide/
  Stack trace:
    at add (file:///C:/a/green-galaxy/node_modules/astro/dist/cli/add/index.js:218:31)
    at async runCommand (file:///C:/a/green-galaxy/node_modules/astro/dist/cli/index.js:117:7)
```

### What's the expected result?

Successfully install it without any errors.

### Link to Minimal Reproducible Example

https://github.com/mika-sandbox/astro-add-react-broken

### Participation

- [ ] I am willing to submit a pull request for this issue.

Reikon95:
Also just started getting this error whenever adding react.

emish89:
> Likely due to
> 
> https://github.com/withastro/astro/blob/770b9f06c3d67f89de527d8b3a842555e596c01f/packages/integrations/react/package.json#L62-L64

[...]

PR #10988, opened 2024-05-09
Fix astro add react

bluwy:
## Changes

Fix https://github.com/withastro/astro/issues/10984

Alternative to https://github.com/withastro/astro/pull/10986 to not remove the `npm:` peer deps, if it works.

## Testing

Ran `astro add react` manually.

## Docs

n/a.

### DIFF
--- a/packages/astro/src/cli/add/index.ts
+++ b/packages/astro/src/cli/add/index.ts
@@ -693,7 +693,16 @@ async function resolveRangeToInstallSpecifier(name: string, range: string): Prom
 	if (versions instanceof Error) return name;
 	// Filter out any prerelease versions
 	const stableVersions = versions.filter((v) => !v.includes('-'));
+	// Peer dependencies may sometimes use the `npm:` specifier, but `semver` package panics with it.
+	// Filter it out if so.
+	if (range.includes('npm:')) {
+		range = range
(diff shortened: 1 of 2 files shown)


packages/astro/src/cli/add/index.ts
```diff
@@ -693,7 +693,16 @@ async function resolveRangeToInstallSpecifier(name: string, range: string): Prom
 	if (versions instanceof Error) return name;
 	// Filter out any prerelease versions
 	const stableVersions = versions.filter((v) => !v.includes('-'));
+	// Peer dependencies may sometimes use the `npm:` specifier, but `semver` package panics with it.
+	// Filter it out if so.
+	if (range.includes('npm:')) {
```
ematipico, line 698: nit: `startsWith`

ematipico (approved):

bluwy:
Closing as `npm:` specifiers don't work in peer deps.

CGQAQ:
It is not that `npm:` don't work in peer deps, it's because that you can not mix alias with other specifiers,
for example, npm tries to treat the whole part after `npm:` (which is `types-react-dom@beta || npm:types-react-dom@beta`) as the  package was aliased to.
image

PR #10990, opened 2024-05-09
fix:  fix incorrect actions path on windows

liruifengv:
## Changes

On Windows, the path `stringifiedActionsImport` will get `/D:/XX/src/actions`, this cause the `Actions` type to be `any`.

image

### DIFF
--- a/.changeset/fluffy-zoos-hammer.md
+++ b/.changeset/fluffy-zoos-hammer.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+fix: remove actions path leading slash on windows
--- a/.changeset/fluffy-zoos-hammer.md
+++ b/.changeset/fluffy-zoos-hammer.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-fix: remove actions path leading slash on windows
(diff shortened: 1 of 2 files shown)


matthewp:
What happens if instead of doing this we remove the `new URL('actions', srcDir).pathname` with `fileURLToPath(new URL('actions', srcDir))`? Does that fix it?

liruifengv:
> What happens if instead of doing this we remove the `new URL('actions', srcDir).pathname` with `fileURLToPath(new URL('actions', srcDir))`? Does that fix it?

It seems to work! I will try again

matthewp:
@liruifengv actually we have this function already for this use case: https://ygithub.com/withastro/astro/blob/main/packages/astro/src/core/util.ts#L90. can you try using it? I think that's exactly what you want here.

liruifengv:
> @liruifengv actually we have this function already for this use case: https://ygithub.com/withastro/astro/blob/main/packages/astro/src/core/util.ts#L90. can you try using it? I think that's exactly what you want here.

It's perfect!

matthewp (approved):

PR #10991, opened 2024-05-09
fix: new version message when running astro dev

ArmandPhilippot:
The `getExecCommand` is an async function but it wasn't awaited in the `newVersionAvailable` function so the message in the console was:

```
Run [object Promise] @astrojs/upgrade to update
```

Instead of (for example):

```
Run pnpx @astrojs/upgrade to update
```

## Changes

I transformed `newVersionAvailable` to an async function and added two `await`:
* one for `getExecCommand()`
* another where `newVersionAvailable` was called

Should I add a changeset for this?

## Testing

Not sure how to test. If I use `pnpm link` on a project, it will use the latest version so the message will not be displayed. But I'm not sure a test is necessary.

I also run `pnpm test` and everything was ok.

## Docs

No new behavior, only a minor fix so no need to update the doc.

### DIFF
--- a/packages/astro/src/core/dev/dev.ts
+++ b/packages/astro/src/core/dev/dev.ts
@@ -67,7 +67,7 @@ export default async function dev(inlineConfig: AstroInlineConfig): Promise<DevS
 
 						logger.warn(
 							'SKIP_FORMAT',
-							msg.newVersionAvailable({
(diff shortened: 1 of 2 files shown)


matthewp (approved):
amazing, thanks!

Issue #10464, opened 2024-03-16
Astro's `prefetch` doesn't work in non-Chromium-based browsers


smastrom:
### Astro Info

```block
Astro                    v4.5.5
Node                     v21.6.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             none
```

### If this issue only occurs in one browser, which browser is a problem?

Firefox, Safari

### Describe the Bug

According to the [docs](https://docs.astro.build/en/guides/prefetch/), Astro prefetch should work on any browser whether or not View Transitions are enabled (as it is not mentioned otherwise).

I virutally tried any combination under my `astro.config.mjs`: `prefetch: true`, `prefetch.prefetchAll: true`, etc. And also to manually append the `data-astro-prefetch` attribute as explained in the docs and enable/disable ViewTransitions. 

Under the "Network" tab of both Firefox 124 (latest) and Safari 17.4 (latest) there is no way I can see the incoming documents.

The only way to see them in that tab is to manually call `prefetch()` in a script tag using `{ with: 'fetch' }`, however, those pages are then requested again to the server once I click on the correspondent links.

On Chrome and Edge (and I suppose any other Chromium-based browser) everything works as per docs.

I am not sure if this expected, if it's a bug or if I (hopefully) misinterpreted the docs and actually doing something wrong. Thank you.

### What's the expected result?

To work on Firefox and Safari as it works on Chrome. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-rdn9rn?file=src%2Flayouts%2FLayout.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
I used Safari exclusively when implementing this, so it should work. Are you testing this in dev or prod? The prefetch behaviour also relies on a proper `Cache-Control` header set, so depending on your deployment platform, they may set a short Cache-Control so that it's cached on the second request. In dev, `Cache-Control` is usually not set so you might not see the benefits.

IIRC Chrome implements some unique heuristics, e.g. checking date of request and guessing a proper cache time if `Cache-Control` is not set. So it can sometime work for Chrome. https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_caching

If you're unsure if the prefetching is actually done by the script, you can enable debug/verbose logging in the browser console to see the prefetch logs (only in dev).

smastrom:
[...]

PR #10999, opened 2024-05-10
Improve prefetch behaviour for browsers

bluwy:
## Changes

Closes https://github.com/withastro/astro/issues/10464

- This PR makes all prefetch strategies to use `<link rel="prefetch">` if supported, or else fall back to `fetch()`.
- It also applies the same for the programmatic `prefetch()` API, unless the `with: 'link' | 'fetch'` option is passed. The `with` option is deprecated as it's better to use link whenever possible.

More explanation for why this change is made at https://github.com/withastro/astro/issues/10464#issuecomment-2104238799

## Testing

The existing tests should pass. Testing is a bit tricky because this mainly matters for Firefox and Safari but we don't have those setup. I think manual testing should be enough for now as the prefetch code is isolated and doesn't changed often.

## Docs

https://github.com/withastro/docs/pull/8246


ematipico:
@bluwy what should we do to test it on Firefox?

bluwy:
We probably need to update `playwright.config.js` `projects` to also test on Firefox etc. But we don't want to run all tests on Firefox, so we may need a second playwright config that only runs specific tests? And I think that should be enough, we can also run it for the view transitions tests.

We also need to update this to install the firefox etc binaries:
https://github.com/withastro/astro/blob/1c5905940241931d93330993f7f62798d19284fd/package.json#L33-L34

ematipico:
Awesome, I can make a PR for that if you're happy

packages/astro/src/prefetch/index.ts
```diff
@@ -215,28 +219,27 @@ export function prefetch(url: string, opts?: PrefetchOptions) {
...
+	// Prefetch with speculationrules if `clientPrerender` is enabled and supported
+	// NOTE: This condition is tree-shaken if `clientPrerender` is false as its a static value
+	if (clientPrerender && HTMLScriptElement.supports?.('speculationrules')) {
+		debug?.(`[astro] Prefetching ${url} with <script type="speculationrules">`);
 		appendSpeculationRules(url);
-	} else if (priority === 'link') {
+	}
+	// Prefetch with link if supported
+	else if (
+		document.createElement('link').relList?.supports?.('prefetch') &&
```
Princesseuh, line 230: Isn't it somewhat expensive to do this on every link? Could it be done once?

bluwy: I benchmarked this locally and there isn't a big difference in perf from what I can tell compared to caching, so I kept this direct approach, and to save some bytes.

Princesseuh (approved):

bholmesdev (approved):

bholmesdev (approved):
The conditional checks look right to me. Glad we're getting `with` the times!

ematipico:
I added Firefox to the e2e tests, however some of the tests fail e.g. view transitions

PR #10484, opened 2024-03-19
Adding support for toml data entry

q1b:
## Changes

- This will add support for `.toml` extension 
- using [js-toml](https://github.com/sunnyadn/js-toml) package

## Testing

I checked the repository, yet unable to see existing tests for JSON, YAML and TOML data entry points,
If so, there is need for adding tests for these data entry points let me know, I am happy to contribute

## Docs

Adding Support for .toml file in src/content folder with existing data formats json, yaml

PR #11004, opened 2024-05-10
added toml support

q1b:
It's a file format for configuration files that's intended to be easy to read and write. TOML is an alternative to YAML and JSON, and is designed to be more human friendly than JSON and simpler than YAML. 

Added a example for showing how `src/content` we can use `.yaml` and `.toml` for site seo


q1b:
Earlier PR: https://github.com/withastro/astro/pull/10484

@ematipico toml file, comes as same level as yaml, json it's just alternative!

Is it necessary to write documentation about it, if so let me know I will happily make a pr

florian-lefebvre:
Now that the Content Layer is in experimental, we think toml should be implemented as a community [loader](https://github.com/withastro/roadmap/blob/content-layer/proposals/0047-content-layer.md#loaders) at the moment. We may change our position in the future so feel free to watch the roadmap repo, thanks for the attempt!

PR #11008, opened 2024-05-11
docs: fix arguments in experimental.actions code example

k1350:
## Changes

fix arguments in experimental.actions code example

### DIFF
--- a/packages/astro/CHANGELOG.md
+++ b/packages/astro/CHANGELOG.md
@@ -117,7 +117,7 @@
   export const server = {
     like: defineAction({
       input: z.object({ postId: z.string() }),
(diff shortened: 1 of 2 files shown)


ematipico:
Since it's an example, I believe that was intentional: to show that `context` is the second argument

bholmesdev:
@k1350 Ah, this has been changed to a `getApiContext()` helper! You can call this function within your action `handler` to access the context. See here: https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#access-api-context

bholmesdev (approved):
Thanks for catching this! Missed a spot it seems

PR #11013, opened 2024-05-12
Prevents unhandledrejection error when checking for latest Astro version 

### LINKED ISSUE #11012: Check latest astro version fail cause astro render error page at first time visit.
### Astro Info

```block
Astro                    v4.8.2
Node                     v18.18.1
Package Manager          pnpm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

Astro cannot catch [fetchLatestAstroVersion](https://github.com/withastro/astro/blob/main/packages/astro/src/core/dev/dev.ts#L51) error when fetch failed, and it will create a error layout when first time fetch the page.

This bug may only appear in some country that cannot connect to npmjs.org normally.
It cannot reproduction in the stackblitz. I only can put some screenshots to show it.

The error will cause my test fail when I fetch the page first time.

Screenshots:
图片
图片

Stackblitz reproduction can follow this:
- Modify `node_modules/astro/dist/cli/install-package.js` and let `fetchPackageVersions` always return Error.
- Modify `node_modules/astro/dist/core/dev/dev.js#L31` and made shouldCheck to `true`.

图片

Expect cli do nothing, keep server run normally.

https://stackblitz.com/edit/github-82f59z

### Participation

QingXia-Ela:
## Changes

Fix #11012

## Testing

Running test locally.

## Docs

It doesn't need to update docs.

### DIFF
--- a/.changeset/warm-schools-applaud.md
+++ b/.changeset/warm-schools-applaud.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-fix: prevent error appear when check latest version fail(#11012)
+Prevents unhandledrejection error when checking for latest Astro version 
--- a/.changeset/warm-schools-applaud.md
+++ b/.changeset/warm-schools-applaud.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+fix: prevent error appear when check latest version fail(#11012)
--- a/packages/astro/src/core/dev/dev.ts
+++ b/packages/astro/src/core/dev/dev.ts
@@ -48,29 +48,33 @@ export default async function dev(inlineConfig: AstroInlineConfig): Promise<DevS
 			// Don't await this, we don't want to block the dev server from starting
 			shouldCheckForUpdates(restart.container.settings.preferences).then(async (shouldCheck) => {
 				if (shouldCheck) {
-					const version = await fetchLatestAstroVersion(restart.container.settings.preferences);
+					try {
+						const version = await fetchLatestAstroVersion(restart.container.settings.preferences);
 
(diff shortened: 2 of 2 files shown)


packages/astro/src/core/dev/dev.ts
```diff
@@ -48,29 +48,33 @@ export default async function dev(inlineConfig: AstroInlineConfig): Promise<DevS
...
+								await msg.newVersionAvailable({
+									latestVersion: version,
+								})
+							);
+						}
+					} catch (e) {
+						// Do nothing, check version fail shouldn't block server running.
 					}
 				}
 			});
```
bluwy, line 80: I think the fix is to append a `.catch(() => {})` here instead. The outer `try..catch` intent was to ignore the error, but it doesn't work when the code uses `.then()`.

.changeset/warm-schools-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+fix: prevent error appear when check latest version fail(#11012)
```
bluwy, line 5: Updated this to match the usual formatting.
```suggestion
Prevents unhandledrejection error when checking for latest Astro version
```

packages/astro/src/core/dev/dev.ts
QingXia-Ela: Done in 3c53f9593ab87a82afc913a20dea274212c3c3ac.

ematipico (approved):

florian-lefebvre (approved):

matthewp:
Thank you!

PR #11021, opened 2024-05-13
feat: make CSRF protection stable

ematipico:
## Changes

This PR makes CSRF protection stable. As for now, we decided to **not** implement a cookie strategy because it requires a strategy that isn't very Astro-y.

The configuration has been changed, and it was make less deep: 

```diff
export default defineConfig({
-  experimental: {
-    security: {
-      csrfProtection: {
-        origin: true
-      }
-    }
-  },
+  security: {
+    checkOrigin: true
+  }
})
```

## Testing

The current tests should pass.

## Docs


packages/astro/src/core/build/generate.ts
```diff
@@ -284,7 +284,7 @@ async function getPathsForRoute(
 		const label = staticPaths.length === 1 ? 'page' : 'pages';
 		logger.debug(
 			'build',
-			`├── ${bold(green('✔'))} ${route.component} → ${magenta(`[${staticPaths.length} ${label}]`)}`
+			`├── ${bold(green('√'))} ${route.component} → ${magenta(`[${staticPaths.length} ${label}]`)}`
```
matthewp, line 287: why did this change?

matthewp:
In addition to unflagging the RFC should be updated to account for the fact that we're not going to be doing the cookie check for now. You could either remove that section or move it to the alternative ideas section: https://github.com/withastro/roadmap/pull/879

packages/astro/src/core/build/generate.ts
ematipico: I didn't, Biome did (bug)

matthewp: Ah ok, can you save it without the formatter in this case then?

bluwy (approved):

packages/astro/src/@types/astro.ts
```diff
@@ -1686,6 +1686,47 @@ export interface AstroUserConfig {
 	 */
 	legacy?: object;
 
+	/**
+	 * @docs
+	 * @name experimental.security
```
sarah11918, line 1691:
```suggestion
	 * @name security
```

packages/astro/src/@types/astro.ts
```diff
@@ -1686,6 +1686,47 @@ export interface AstroUserConfig {
...
+	 */
+	security?: {
+		/**
+		 * @name security.checkOrigin
+		 * @type {boolean}
+		 * @default 'false'
+		 * @version 4.6.0
+		 * @description
+		 *
+		 * The CSRF protection works only for pages rendered on demand (SSR) using `server` or `hybrid` mode. The pages must opt out of prerendering in `hybrid` mode.
```
sarah11918, line 1717: I'm wondering about something like this...

``````suggestion
	 * Enables security measures, such as Cross-Site Request Forgery (CSRF) protection, for an Astro website.
	 *
	 * These features only exist for pages rendered on demand (SSR) using `server` or `hybrid` mode. The pages must opt out of prerendering in `hybrid` mode.
	 *
	 * ```js
	 * // astro.config.mjs
	 *   output: "server",
	 *   security: {
	 *     checkOrigin: true
	 *   }
	 * })
	 * ```
	 */
	security?: {
		/**
		 * @name security.checkOrigin
		 * @type {boolean}
		 * @default 'false'
		 * @version 4.6.0
		 * @description
``````
If it's true that all of the security measures that will eventually be here will require server mode, then I think it makes sense to front-load the SSR requirement at the top, then it won't need to be in each item below.

Plus, since the value is no longer named `csrfProtection`, we can't assume people will make the connection to CSFR protection and will have to spell it out a bit more somewhere. It's probably OK to have the description below be directly about checking origin, since that's the new name of the feature.

**Review by sarah11918 — COMMENTED** [review]
Just a quick thought here for the docs, @ematipico !

Also, wondering whether we might want to link here from any other docs pages like our "Authentication" guide as a "thing to consider enabling".

Is there any reason why someone might have authentication on their Astro site and NOT want this set?

File: packages/astro/src/@types/astro.ts (hunk 3 above)
Comment (ematipico) [reply packages/astro/src/@types/astro.ts:1717]: I think your suggestion should be belong to `security.checkOrigin`. In the future, the `security` object might contain other security measures unrelated to CSRF.

Comment (sarah11918) [reply packages/astro/src/@types/astro.ts:1717]: OK, what about the part about requiring SSR? Do you already know/think these will all require server-rendering? That was the thing I was mostly thinking about putting up front if everything was going to require it, so we wouldn't have to duplicate it for more items.

If not, it's totally fine to leave it where it was!

**Review by matthewp — CHANGES_REQUESTED** [review]
Blocked pending the call for consensus passing.

File: packages/astro/src/@types/astro.ts (hunk 3 above)
Comment (ematipico) [reply packages/astro/src/@types/astro.ts:1717]: > OK, what about the part about requiring SSR? Do you already know/think these will all require server-rendering?

I think, yeah, I doubt that a static build could have some security flaws; in the end, nothing is exposed other than some static file.

File: .changeset/chatty-experts-smell.md [hunk 4]
```diff
@@ -0,0 +1,20 @@
+---
+"astro": minor
+---
+
+The CSRF protection check is out of experimental, and its configuration is changed:
```
sarah11918, line 5: @ematipico if you get a chance to beef this up a bit more before I get here, great! You can use the `domains` PR as an example of ideally what goes here! If not, I will make this match before it's merged. (Just want to spend my time in the docs PRs first!)

sarah11918 (commented):
Other than a more detailed changeset for this unflagged feature, just a couple of quick suggestions for the actual docs to take a look at! (Note, one earlier comment was updated based on feedback).

.changeset/chatty-experts-smell.md
ematipico: Done

ematipico:
> Just a quick thought here for the docs, @ematipico !
> 
> Also, wondering whether we might want to link here from any other docs pages like our "Authentication" guide as a "thing to consider enabling".
> 
> Is there any reason why someone might have authentication on their Astro site and NOT want this set?

If users are using SSR and plan to expose some endpoints that should be reachable from external domains (e.g. posting a form), then they should not enable this feature, otherwise all requests coming from external domains will be blocked.

.changeset/chatty-experts-smell.md
```diff
@@ -0,0 +1,27 @@
+---
+"astro": minor
+---
+
+The CSRF protection feature that was introduced behind a flag in [v4.6.0](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#460) is no longer experimental and is available for general use. If you were using the experimental feature, you'll have to update your `astro.config.mjs` as follows:
```
sarah11918, line 5:
```suggestion
The CSRF protection feature that was introduced behind a flag in [v4.6.0](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#460) is no longer experimental and is available for general use. 

To enable the stable version, add the new top-level `security` option in `astro.config.mjs`. If you were previously using the experimental version of this feature, also delete the experimental flag:
```
This one is a little different from some other cases because they may or may not have been using the experimental flag, but might still want to add it new to the same config now!

.changeset/chatty-experts-smell.md
```diff
@@ -0,0 +1,27 @@
...
++  }
+})
+```
+
+Enabling this setting performs a check that the `"origin"` header, automatically passed by all modern browsers, matches the URL sent by each Request.
+
+This check is executed only for pages rendered on demand, and only for the requests `POST`, `PATCH`, `DELETE` and `PUT` with one of the following `"content-type"` headers: `'application/x-www-form-urlencoded'`, `'multipart/form-data'`, `'text/plain'`.
+
+If the `"origin"` header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
+*
```
sarah11918, line 27:
```suggestion

For more information, see the [`security` configuration docs](https://docs.astro.build/en/reference/configuration-reference/#security)
```

packages/astro/src/@types/astro.ts
```diff
@@ -1686,6 +1686,47 @@ export interface AstroUserConfig {
 	 */
 	legacy?: object;
 
+	/**
+	 * @docs
```
sarah11918, line 1690: I think this should be moved much higher up in the file? It loos like this will now appear under "Legacy Flags" in the sidebar.

This is a top-level option, right? So it should probably move up maybe between `scopedStyleStrategy` and `vite`? These aren't in alphabetical order, but manually by common usage first, and then whenever just seems right. I think just above `vite` makes sense, but open to suggestions if you don't like that!

sarah11918 (commented):
This looks great! A few minor tweaks, and checking closely, I think we need to move up the JSDoc content to the "Top-level Options" section.

Otherwise, ready to go!

packages/astro/src/@types/astro.ts
ematipico: Done!

.changeset/chatty-experts-smell.md
```diff
@@ -0,0 +1,30 @@
...
+})
+```
+
+Enabling this setting performs a check that the `"origin"` header, automatically passed by all modern browsers, matches the URL sent by each Request.
+
+This check is executed only for pages rendered on demand, and only for the requests `POST`, `PATCH`, `DELETE` and `PUT` with one of the following `"content-type"` headers: `'application/x-www-form-urlencoded'`, `'multipart/form-data'`, `'text/plain'`.
+
+If the `"origin"` header doesn't match the pathname of the request, Astro will return a 403 status code and won't render the page.
+
+For more information, see the [`security` configuration docs](https://docs.astro.build/en/reference/configuration-reference/#security)
```
sarah11918, line 30:
```suggestion
For more information, see the [`security` configuration docs](https://docs.astro.build/en/reference/configuration-reference/#security).
```
Oops!

packages/astro/src/@types/astro.ts
```diff
@@ -779,6 +779,47 @@ export interface AstroUserConfig {
 	 */
 	scopedStyleStrategy?: 'where' | 'class' | 'attribute';
 
+	/**
+	 * @docs
+	 * @name security
+	 * @type {boolean}
+	 * @default `{}`
+	 * @version 4.6.0
```
sarah11918, line 787:
```suggestion
	 * @version 4.9.0
```

sarah11918 (approved):
Great! I left two tiny fixes in comments, but I am happy to consider this good to go! :partying_face:

matthewp (approved):

PR #11028, opened 2024-05-13
actions: Throw error on missing server output

bholmesdev:
## Changes

Throw error when `output` is missing when using actions.

## Testing

Manual testing

## Docs

Already documented!

### DIFF
--- a/packages/astro/src/actions/index.ts
+++ b/packages/astro/src/actions/index.ts
@@ -1,18 +1,16 @@
 import { mkdir, readFile, writeFile } from 'node:fs/promises';
 import type { Plugin as VitePlugin } from 'vite';
 import type { AstroIntegration } from '../@types/astro.js';
-import { viteID } from '../core/util.js';
+import { viteID, isServerLikeOutput } from '../core/util.js';
 import { ACTIONS_TYPES_FILE, RESOLVED_VIRTUAL_MODULE_ID, VIRTUAL_MODULE_ID } from './consts.js';
 import { AstroUserError } from '../core/errors/errors.js';
-
-const serverOutputs = ['server', 'hybrid'];
 
 export default function astroActions(): AstroIntegration {
 	return {
 		name: VIRTUAL_MODULE_ID,
 		hooks: {
 			async 'astro:config:setup'(params) {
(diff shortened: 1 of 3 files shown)


packages/astro/src/actions/index.ts
```diff
@@ -3,12 +3,24 @@ import type { Plugin as VitePlugin } from 'vite';
...
+import { AstroUserError } from '../core/errors/errors.js';
+
+const serverOutputs = ['server', 'hybrid'];
 
 export default function astroActions(): AstroIntegration {
 	return {
 		name: VIRTUAL_MODULE_ID,
 		hooks: {
 			async 'astro:config:setup'(params) {
+				if (!serverOutputs.includes(params.config.output)) {
```
Princesseuh, line 15: You might want to use https://github.com/withastro/astro/blob/461cd42a0652a6675e72619962a1c0cee902af5c/packages/integrations/vercel/src/lib/prerender.ts#L3 for this

packages/astro/src/actions/index.ts
```diff
@@ -3,12 +3,24 @@ import type { Plugin as VitePlugin } from 'vite';
...
+					const error = new AstroUserError(
```
Princesseuh, line 16: You could use a normal `AstroError` here. If you fear that the error won't last, then you can use a not defined one, but if you think it will last definitely document it! I think it'll be a fairly common wonder of users why actions need output server.

ematipico (commented):
What @Princesseuh said, you should use `isServerLikeOutput`

bholmesdev:
@Princesseuh @ematipico made those adjustments and added an `errors-data` entry! Ready for re-review.

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1493,6 +1493,13 @@ export const DuplicateContentEntrySlugError = {
 	},
 } satisfies ErrorData;
 
+export const ActionsWithoutServerOutputError = {
```
Princesseuh, line 1496: You need docs for the error, or it'll lead to a 404 link in the overlay.

Princesseuh (approved):
I think the error message could be improved, but nothing we can't do later. Awesome!

PR #11032, opened 2024-05-13
prompt for workspace in db link

itsMapleLeaf:
Should be merged and released _after_ the Studio PR to list all workspaces


packages/db/src/core/cli/commands/link/index.ts
```diff
@@ -16,20 +16,16 @@ export async function cmd() {
 		console.error(MISSING_SESSION_ID_ERROR);
 		process.exit(1);
 	}
-	const getWorkspaceIdAsync = getWorkspaceId().catch((err) => {
-		return err as Error;
-	});
+	const workspaceId = await promptWorkspace(sessionToken);
```
FredKSchott, line 19: This is prompting too early in the flow, if you run this code you'll see it prompt for the workspace before it prompts to ask if you even want to link the project in the first place (`promptBegin()`)

I imagine, as a user, you'd want:

- `promptBegin`
- `promptLinkExisting`
- `promptWorkspace`
- then, fork between `isLinkExisting`, `isLinkNew`,  etc.

Basically, just move this down to like 22 and you should be good to go

delucis:
Updated this PR based on @FredKSchott’s feedback. The flow is now:

- Link "directory/" with Astro Studio? y/n
   - **Yes:** Link with an existing project in Astro Studio? y/n
      - **Yes:** Select your workspace: [assuming there are multiple, skipped if there is only one]
         - What is your project name?
   - **No:** Create a new project in Astro Studio? y/n
      - **Yes:** Select your workspace: [as above]
         - What is your new project's name?

PR #11041, opened 2024-05-14
Actions: support empty args and empty response

bholmesdev:
## Changes

We raised a 500 when an action's arguments or return value were `undefined`. This follows the HTTP spec to respect these cases.
- Return a 204 when the response is empty.
- Include the `Content-Length` on JSON args to check for a length of zero.
- Misc: return a 404 instead of a 500 when requesting an action that doesn't exist.

## Testing

Add test for 204 status.

## Docs

N/A

### DIFF
--- a/.changeset/many-guests-yell.md
+++ b/.changeset/many-guests-yell.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Actions: fix 500 errors when sending empty params or returning an empty response from an action.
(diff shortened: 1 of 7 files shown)


bluwy (approved):

.changeset/many-guests-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Actions: fix 500 errors when sending empty params or returning an empty response from an action.
```
florian-lefebvre, line 5:
```suggestion
Fixes 500 errors when sending empty params or returning an empty response from an action.
```

florian-lefebvre (approved):

PR #11050, opened 2024-05-15
fix(rss): fix an issue where trailing slash is not removed even if `trailingSlash` is set to `false`

### LINKED ISSUE #10994: @astrojs/rss uses trailing slash on home page when it shouldn't
### Astro Info

```block
Astro                    v4.8.2
Node                     v20.12.1
Package Manager          npm
```

_No response_

### Describe the Bug

In my astro.config.mjs I set:

export default defineConfig({
...
trailingSlash: 'never',
})

and in my vercel.json I set:

{
"cleanUrls": true,
...
"trailingSlash": false
}

and in my rss.xml.js I set:

export async function GET(context) {
  return rss({
    trailingSlash: false,
  })
}

and yet, the RSS feed still puts a slash on the home page, viewable here:

http://astro-rss-trailing-slash.patrickreiner.com/rss.xml

RSS home page should not have a trailing slash.

Should be:

<link>https://astro-rss-trailing-slash.patrickreiner.com</link>

Not:

<link>https://astro-rss-trailing-slash.patrickreiner.com/</link>

https://github.com/patrickreiner/astro-rss-trailing-slash

### Participation

mingjunlu:
## Changes

- Makes `createCanonicalURL` function return string instead of URL object
- Fixes an issue where trailing slash is not removed even if `trailingSlash` option is set to `false` when generating RSS string
- Closes #10994

## Testing

- Related test case updated
- `pnpm --filter @astrojs/rss run test`

## Docs

- N/A

### DIFF
--- a/.changeset/tame-otters-destroy.md
+++ b/.changeset/tame-otters-destroy.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/rss": patch
+---
+
+Fixes an issue where trailing slash is not removed even if the `trailingSlash` option is set to `false`.
--- a/packages/astro-rss/src/index.ts
+++ b/packages/astro-rss/src/index.ts
@@ -202,7 +202,7 @@ async function generateRSS(rssOptions: ValidatedRSSOptions): Promise<string> {
 	root.rss.channel = {
 		title: rssOptions.title,
 		description: rssOptions.description,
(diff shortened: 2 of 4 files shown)


ematipico (approved):
Oh, wow, this is a great fix! Thank you @mingjunlu

bluwy (approved):

PR #11051, opened 2024-05-15
feat: container APIs

ematipico:
## Changes

Relative RFC: https://github.com/withastro/roadmap/pull/916 (this RFC will be soon updated based on the APIs and options crafted in this PR)

As for now, we publicly expose the following methods:

- `create`: creates a new instance of the container
- `renderToString`: renders a component to string
- `renderToResponse`: returns the response emitted by the rendering phase

The methods `insertRoute` and `createFromManifest` were left private using TS on purpose because I still want them available so I can still play with them without breaking the compatibility with end users.

The goal is to provide, **for now**, the ability to test components with `vitest`. This would allow us to quickly release the APIs and receive feedback.

## Testing

Added unit tests and example tests

## Docs

https://github.com/withastro/docs/pull/8281


examples/container-with-vitest/README.md
```diff
@@ -0,0 +1,11 @@
+# Astro + [Vitest](https://vitest.dev/) Example
+
+```sh
+npm create astro@latest -- --template with-vitest
```
Princesseuh, line 4:
```suggestion
npm create astro@latest -- --template container-with-vitest
```

Princesseuh: Probably that the rest of the README needs to be updated too, but at least making the code snippet correct is the most important 😅

examples/container-with-vitest/test/ReactWrapper.test.ts
```diff
@@ -0,0 +1,20 @@
+import { experimental_AstroContainer as AstroContainer } from 'astro/container';
+import { expect, test } from 'vitest';
+// @ts-ignore
+import ReactWrapper from '../src/components/ReactWrapper.astro';
```
Princesseuh, line 4: Was there an error here? This is supported in our TypeScript plugin

examples/container-with-vitest/test/ReactWrapper.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+test('ReactWrapper with react renderer', async () => {
+	const container = await AstroContainer.create({
+		renderers: [
+			{
+				name: '@astrojs/react',
+				clientEntrypoint: "@astrojs/react/client.js",
+				serverEntrypoint: "@astrojs/react/server.js",
+			}
```
Princesseuh, line 13: For the future, but will we be providing an easier way to build this for users?

examples/container-with-vitest/test/[locale].test.ts
```diff
@@ -0,0 +1,18 @@
+import { experimental_AstroContainer as AstroContainer } from 'astro/container';
+import { expect, test } from 'vitest';
+// @ts-ignore
+import Locale from '../src/pages/[locale].astro';
+
+test('Dynamic route', async () => {
+	const container = await AstroContainer.create();
+	// @ts-ignore
```
Comment (Princesseuh) [inline examples/container-with-vitest/test/[locale].test.ts:8]: What does this one do?

packages/astro/src/container/index.ts
```diff
@@ -0,0 +1,468 @@
...
+	request?: Request;
+	/**
+	 * Useful for dynamic routes. If your component is something like `src/pages/blog/[id]/[...slug]`, you'll want to provide:
+	 * ```js
+	 * container.renderToString(Component, { params: ["id", "...slug"] });
+	 * ```
+	 */
+	params?: Record<string, string | undefined>;
+	/**
+	 * Useful if your component needs to access some locals without the use a middleware.
```
Princesseuh, line 63: I'll let @sarah11918 go over these, but these comments don't seem really useful (ha) to me. It's, in my opinion, important to explain what something is in addition to explain why something might be useful.

packages/astro/src/container/index.ts
```diff
@@ -0,0 +1,468 @@
...
+	 */
+	route?: string
+};
+
+/**
+ * @param renderers
+ * @param config
+ * @param manifest
+ * @param middleware
+ */
```
Princesseuh, line 95: I see you making those kinds of empty JSDoc comments often, but I don't find them really useful. I'd suggest adding a description or not adding a JSDoc at all

packages/astro/src/container/index.ts
```diff
@@ -0,0 +1,468 @@
...
+				}
+				return undefined;
+			})
+		);
+		const finalRenderers = loadedRenderers.filter((r): r is SSRLoadedRenderer => Boolean(r));
+		
+		return new experimental_AstroContainer({ streaming, renderers: finalRenderers, config });
+	}
+
+	// NOTE: we keep this private via TS instead via `#` so it's still available on the surface, so we can play with it.
```
Princesseuh, line 300: What does "play with it" mean?

packages/astro/src/container/index.ts
```diff
@@ -0,0 +1,468 @@
...
+	 * 
+	 * const container = await AstroContainer.create();
+	 * const result = await container.renderToString(Card);
+	 * 
+	 * console.log(result); // it's a string
+	 * ```
+	 * 
+	 * 
+	 * @param {AstroComponentFactory} component The instance of the component.
+	 * @param {ContainerRenderOptions=} options Possible options to pass when rendering the component.
```
Princesseuh, line 370: What's this `=` at the end of the type? Saw it in a few places

bluwy: With TSDoc, we should also be able to remove the type altogether so we don't have to declare it in two places.

ematipico: @Princesseuh It's a JSDoc thing to mark something as optional

packages/astro/src/container/index.ts
ematipico: I plan to use it to create some integration that leverages it. However, I don't know yet how an integration would look like, so I wanted to keep it "available" to us maintainers until I have a clear idea on how we can use it.

What should I change in the comment?

packages/astro/src/container/index.ts
ematipico: Force of the habit :)

packages/astro/src/container/index.ts
ematipico: Do you mean that the wording isn't correct? Or the whole comment is superfluous?

examples/container-with-vitest/test/[locale].test.ts
Comment (ematipico) [reply examples/container-with-vitest/test/[locale].test.ts:8]: I don't know. I was hoping someone like yourself could help: https://github.com/withastro/astro/actions/runs/9093410164/job/24992249153?pr=11051

Comment (ematipico) [reply examples/container-with-vitest/test/[locale].test.ts:8]: Screenshot 2024-05-15 at 14 11 59

examples/container-with-vitest/test/ReactWrapper.test.ts
ematipico: Definitely. At the moment no integration exposes something like that :(

examples/container-with-vitest/test/ReactWrapper.test.ts
ematipico: Yes: https://github.com/withastro/astro/pull/11051#discussion_r1601620312

examples/container-with-vitest/test/[locale].test.ts
Comment (Princesseuh) [reply examples/container-with-vitest/test/[locale].test.ts:8]: Ah, I see what's happening. Your function expects the internal, runtime, type of an Astro component, but the editor tooling has completely different types specific to editor tooling. 

I'm not quite sure what's the best way to do this, maybe faking supporting the type just so users have a better experience is okay, not sure.

For your screenshot, I assume this is webstorm? So you might not have our TypeScript plugin installed: https://docs.astro.build/en/guides/typescript/#setup

packages/astro/src/container/index.ts
Princesseuh: I think the wording is correct, it's just feels incomplete, which makes them seems superfluous. I think just a short description saying *what* something is before saying what it is useful for would do wonder!

Assume that the user has no context, so for instance, they might not know at all what `params` are.

packages/astro/src/container/index.ts
```diff
@@ -0,0 +1,443 @@
...
+	/**
+	 * Useful in case you're attempting to render an endpoint:
+	 * ```js
+	 * container.renderToString(Endpoint, { routeType: "endpoint" });
+	 * ```
+	 */
+	routeType?: RouteType;
+};
+
+function createContainerManifest(
```
matthewp, line 78: is this naming right? it looks like it creates an SSR Manifest *from* a container manifest (among other things).

packages/astro/src/container/pipeline.ts
```diff
@@ -0,0 +1,143 @@
...
+import { AstroError } from '../core/errors/index.js';
+import { RouteNotFound } from '../core/errors/errors-data.js';
+import type { SinglePageBuiltModule } from '../core/build/types.js';
+
+export class TestPipeline extends Pipeline {
+	/**
+	 *
+	 * @private
+	 */
+	#componentsInterner: WeakMap<RouteData, SinglePageBuiltModule> = new WeakMap<
```
matthewp, line 22: missing a comment, what is this?

packages/astro/test/container.test.js
```diff
@@ -0,0 +1,193 @@
...
+					`,
+					}
+				)}`;
+			},
+			'Component2.astro',
+			undefined
+		);
+
+		const container = await experimental_AstroContainer.create();
+		container.insertRoute({
```
matthewp, line 100: This API isn't in the RFC. I would suggest adding it and explaining what its for.

packages/astro/src/container/index.ts
ematipico: Would `createManifest` be a better name?

packages/astro/src/container/pipeline.ts
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	async #retrieveSsrEntry(route: RouteData, filePath: string): Promise<SinglePageBuiltModule> {
+		if (this.#componentsInterner.has(route)) {
+			// SAFETY: it is checked inside the if
+			return this.#componentsInterner.get(route)!;
+		}
+		const ssrEntryURLPage = this.#createEntryURL(filePath, new URL(process.cwd()));
+		const entry = await import(ssrEntryURLPage.toString());
```
matthewp, line 135: this scares me a bit and probably won't work when this is being used in Vite.

Instead of us importing components what do you think about adding an API to allow you to register components inside of a container. That way you can register the ones that might be rewritten to.

packages/astro/test/container.test.js
ematipico: I removed the test and made it private

packages/astro/src/container/index.ts
ematipico: I reworded the comment, let me know what you think

packages/astro/src/container/index.ts
sarah11918: Just a quick note here that I did some editing of descriptions for these in the Container API page PR for docs. These could always be revisited after that is finalized, and/or these could have links to docs (not sure what the proper/usual M.O. is here with these code comments as I don't usually edit them).

But, I have made some suggestions in the docs PR itself for these!

.changeset/brave-colts-cover.md
```diff
@@ -0,0 +1,25 @@
...
+	const result = await container.renderToString(Card, {
+		slots: {
+			default: 'Card content',
+		},
+	});
+
+	expect(result).toContain('This is a card');
+	expect(result).toContain('Card content');
+});
+```
```
sarah11918, line 25:
``````suggestion
```

For a complete reference, see the [Container API docs](/en/reference/container-reference/).

For a feature overview, and to give feedback on this experimental API, see the [Container API roadmap discussion](https://github.com/withastro/roadmap/pull/916).
``````
I would at least include links to the docs, but I think it would also be good to link to the discussion!

File: .changeset/brave-colts-cover.md [hunk 14]
```diff
@@ -0,0 +1,25 @@
+---
+"astro": minor
+---
+
+It's now possible the render components in isolation with the new Astro Container APIs:
+
+If you use `vitest`, you can quickly write a test like this:
```
sarah11918, line 7:
```suggestion
Introduces an experimental Container API to render `.astro` components in isolation.

This API introduces three new functions to allow you to create a new container and render an Astro component returning either a string or a Response:

- `create()`: creates a new instance of the container.
- `renderToString()`: renders a component and return a string.
- `renderToResponse()`: renders a component and returns the `Response` emitted by the rendering phase.

The first supported use of this new API is to enable unit testing. For example, with  `vitest`, you can create a container to render your component with test data and check the result:
```
OK, make all this correct and make sense, but I'm thinking maybe something like this here? Something approx this length and structure would work, I think?

examples/container-with-vitest/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "scripts": {
+    "dev": "astro dev",
+    "start": "astro dev",
+    "build": "astro build",
+    "preview": "astro preview",
+    "astro": "astro",
+    "test": "vitest run"
+  },
+  "dependencies": {
+    "astro": "experimental--container",
```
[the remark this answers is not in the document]
ematipico: This is required for the time being, otherwise `astro check` will fail. We will change this straight after we will release v4.9

sarah11918 (approved):
Changeset looks good to me, so if everyone is happy with its content, consider it approved by docs!

packages/astro/src/container/pipeline.ts
```diff
@@ -0,0 +1,115 @@
...
+			renderers,
+			resolve,
+			serverLike,
+			streaming
+		);
+	}
+
+	componentMetadata(_routeData: RouteData): Promise<SSRResult['componentMetadata']> | void {}
+
+	headElements(routeData: RouteData): Promise<HeadElements> | HeadElements {
```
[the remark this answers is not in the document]
matthewp: Ok, I see now that this is called when the render context is created.

matthewp (approved):

packages/astro/src/container/pipeline.ts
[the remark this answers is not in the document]
ematipico: I used some code we already use in dev/build/ssr. We will probably need to review it in case it causes bugs.

Princesseuh (approved):

PR #11052, opened 2024-05-15
fix: conflict between rewrite and actions middleware

florian-lefebvre:
## Changes

- When using rewrite with actions, a case where `locals._actionsInternal` is reassigned can occur. However, there's a protection to disable reassigning it so it fails. This PR does not attempt to reassign if it already exists

## Testing

Manual patch in a private project

## Docs

N/A, bug fix

### DIFF
--- a/packages/astro/src/actions/runtime/middleware.ts
+++ b/packages/astro/src/actions/runtime/middleware.ts
@@ -12,6 +12,12 @@ export type Locals = {
 
 export const onRequest = defineMiddleware(async (context, next) => {
 	const locals = context.locals as Locals;
+	// Actions middleware may have run already after a path rewrite.
+	// See https://github.com/withastro/roadmap/blob/feat/reroute/proposals/0047-rerouting.md#ctxrewrite
+	// `_actionsInternal` is the same for every page,
+	// so short circuit if already defined.
+	if (locals._actionsInternal) return next();
+
 	const { request, url } = context;
 	const contentType = request.headers.get('Content-Type');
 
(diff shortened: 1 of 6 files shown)


ematipico (approved):

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -42,11 +42,13 @@ export const onRequest = defineMiddleware(async (context, next) => {
 });
 
 function nextWithLocalsStub(next: MiddlewareNext, locals: Locals) {
-	Object.defineProperty(locals, '_actionsInternal', {
-		writable: false,
-		value: {
-			getActionResult: () => undefined,
-		},
-	});
+	if (!locals._actionsInternal) {
```
bholmesdev, line 45: This covers the stub, but do we need a conditional check for [the assignment from middleware](https://github.com/withastro/astro/blob/main/packages/astro/src/actions/runtime/middleware.ts#L42) as well?

bholmesdev (commented):
Question about a conditional check for the success case. Also curious what is triggering the reassignment? Is it an object spread?

packages/astro/src/actions/runtime/middleware.ts
florian-lefebvre: yes probably

florian-lefebvre:
Well I think the case is kinda tricky but basically, I have this middleware
```ts
(context, next) => {
	context.locals.data = null
	context.locals.foo = async (data) => {
		context.locals.data = data
		return await next("/bar")
	}

	return next()
}
```
then I can call it in an astro route:
```ts
Astro.locals.bar("abc")
```
I need to call `next` to be able to keep the data in the rewritten page

PR #11054, opened 2024-05-15
fix: set status on error with progressive fallback

bholmesdev:
## Changes

Respect error status when using a progressive fallback. This will overwrite any `status` set from Astro frontmatter with `response.status`. Curious if we think this makes sense?

## Testing

Already have e2e test for progressive fallback validation errors. It doesn't check the status, but this is difficult to assert.

## Docs

N/A

### DIFF
--- a/packages/astro/src/actions/runtime/middleware.ts
+++ b/packages/astro/src/actions/runtime/middleware.ts
@@ -51,7 +51,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 	if (result.error) {
 		return new Response(response.body, {
 			status: result.error.status,
-			statusText: result.error.message,
+			statusText: import.meta.env.DEV ? result.error.message : 'Action error',
 			headers: response.headers,
 		})
 	}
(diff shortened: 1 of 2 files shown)


matthewp (approved):

ematipico:
Tests are failing consistently  🤔

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -47,7 +47,15 @@ export const onRequest = defineMiddleware(async (context, next) => {
...
 	};
 	Object.defineProperty(locals, '_actionsInternal', { writable: false, value: actionsInternal });
-	return next();
+	const response = await next();
+	if (result.error) {
+		return new Response(response.body, {
+			status: result.error.status,
+			statusText: result.error.message,
+			headers: response.headers,
+		})
```
bluwy, line 56: Maybe we should return a simplified `statusText` for prod? In https://github.com/withastro/astro/pull/10284, it was considered a vulnerability that the error message was shown to the users.

bholmesdev: Good point. I also realized `statusText` does not support special characters like newlines. To keep things simple, I'll just out the `error.name` here, which is just the status code written out like `NOT_FOUND`. No need for prod security checks I don't think

ematipico (approved):

PR #11058, opened 2024-05-15
Fix streaming in Node.js fast path

matthewp:
## Changes

- Rendering occurs before the response starts asking for data.
- This resulted in a scenario where the first render would resolve a promise that the iterator was not listening to.
- This resulting in waiting until the *next* render to actually start streaming.
- The fix is to prevent rendering from creating the promises and instead only resolve them. Now the iterator is in charge of the promise and creates them on each pull.

## Testing

- Updated existing test which only checked for more than 1 chunks when there should be several.
- Added a new assertion to verify that the first chunk does not contain the promise-blocked content.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/fluffy-pears-press.md
+++ b/.changeset/fluffy-pears-press.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fix streaming in Node.js fast path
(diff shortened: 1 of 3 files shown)


matthewp:
!preview node-streaming

bholmesdev (approved):
Looks right to me! Unsure if the `renderingComplete` is needed, since I'd assume `next()` won't be triggered again if rendering _has_ completed. But I don't know the wider context.

ematipico (approved):

Issue #11011, opened 2024-05-12
Page 404.html is not built when using internationalization


luciano-schirmer:
### Astro Info

```block
Astro                    v4.8.2
Node                     v22.0.0
System                   macOS (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             none
```

### If this issue only occurs in one browser, which browser is a problem?

build

### Describe the Bug

Custom `404.html` page is not built to `dist` folder when using internationalization.

This is my `astro.config.mjs`:

```node
import { defineConfig } from 'astro/config';

// https://astro.build/config
export default defineConfig({
  i18n: {
    defaultLocale: "en",
    locales: ["en", "es", "pt-br"],
    routing: {
      prefixDefaultLocale: true,
      redirectToDefaultLocale: true,
    },
  },
});
```

And this is the relevant file structure:

<img width="192" alt="image" src="https://github.com/withastro/astro/assets/9817215/7b9a383d-1b96-4c66-a0e2-8de999c7f7a9">

Steps to reproduce the error:

```shell
npm run build
```

The 404.html is not produced in the `dist` folder.

### What's the expected result?

`404.html` file should be present in `dist` folder.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-plpmkk?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

devacame:
I have the same problem but it's not just the 404 page but any top level pages in page/ 
however, in the build log, it does say that the 404.html page was generated

luciano-schirmer:
My workaround:

Create a `postbuild.js` file in the root of the project (along with `package.json`) with the following content:

```js
// postbuild.js
const fs = require("fs").promises;

async function moveFiles() {
  try {
    await fs.rename("dist/en/404/index.html", "dist/404.html");
    console.log("Files moved successfully!");
  } catch (err) {
    console.error("Error moving files:", err);
  }
  try {
    await fs.rmdir("dist/en/404");
    console.log("Directory removed successfully!");
  } catch (err) {
    console.error("Error removing directory:", err);
  }
}

moveFiles();
```

Then add the following line to `package.json` under `scripts` section:


```json
  "scripts": {
    "postbuild": "node postbuild.js",
  }
```

Works for me.

devacame:
> My workaround:
> 
> Create a `postbuild.js` file in the root of the project (along with `package.json`) with the following content:

actually my files are like below
[...]

PR #11062, opened 2024-05-16
fix(i18n): allow to create 404.html and 500.html

ematipico:
## Changes

Fixes https://github.com/withastro/astro/issues/11011

## Testing

Added a new test case with two new assertions

## Docs

N/A

### DIFF
--- a/.changeset/lazy-rockets-raise.md
+++ b/.changeset/lazy-rockets-raise.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Fixes a bug where `astro build` didn't create custom `404.html` and `500.html` when a certain combination of i18n was applied.
+Fixes a bug where `astro build` didn't create custom `404.html` and `500.html` when a certain combination of i18n options was applied
--- a/.changeset/lazy-rockets-raise.md
+++ b/.changeset/lazy-rockets-raise.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug where `astro build` didn't create custom `404.html` and `500.html` when a certain combination of i18n was applied.
--- a/packages/astro/src/i18n/index.ts
+++ b/packages/astro/src/i18n/index.ts
@@ -19,7 +19,7 @@ export function requestHasLocale(locales: Locales) {
 	};
 }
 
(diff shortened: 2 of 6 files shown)


.changeset/lazy-rockets-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes a bug where `astro build` didn't create custom `404.html` and `500.html` when a certain combination of i18n was applied.
```
florian-lefebvre, line 5:
```suggestion
Fixes a bug where `astro build` didn't create custom `404.html` and `500.html` when a certain combination of i18n options was applied
```

packages/astro/src/i18n/index.ts
```diff
@@ -19,6 +19,17 @@ export function requestHasLocale(locales: Locales) {
 	};
 }
 
+export function requestIs404Or500(request: Request, base?: string) {
+	const url = new URL(request.url);
+
+	return (
+		url.pathname.startsWith('/404') ||
+		url.pathname.startsWith(base + '/404') ||
```
florian-lefebvre, line 27: I think this is wrong if base is undefined? i think it will return `undefined/404`. Maybe this could simplified by making the default value of `base` an empty sting and always checking with the base

packages/astro/src/i18n/index.ts
```diff
@@ -19,6 +19,17 @@ export function requestHasLocale(locales: Locales) {
...
+		url.pathname.startsWith('/500') ||
+		url.pathname.startsWith(base + '/500')
+	);
+}
+
```
florian-lefebvre, line 32:
```suggestion
export function requestIs404Or500(request: Request, base: string = '') {
	const url = new URL(request.url);

	return (
		url.pathname.startsWith(`${base}/404`) ||
		url.pathname.startsWith(`${base}/500`)
	);
}

```

florian-lefebvre (changes requested):

packages/astro/src/i18n/index.ts
ematipico: I hate JS

florian-lefebvre (approved):

PR #11069, opened 2024-05-16
fix: improve logging for on-demand pages

ematipico:
## Changes

This PR improves debug logging for on-demand pages.

I was chatting with a person on Discord, trying to debug some issue they were having, and realised we don't have any logging that could help the debug.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/app/index.ts
+++ b/packages/astro/src/core/app/index.ts
@@ -280,6 +280,10 @@ export class App {
 				this.#logRenderOptionsDeprecationWarning();
 			}
 		}
+		if (routeData) {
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/app/index.ts
```diff
@@ -296,8 +300,12 @@ export class App {
 		}
 		if (!routeData) {
 			routeData = this.match(request);
+			this.#logger.debug("router", "Astro matched the following route for "+ request.url);
```
florian-lefebvre, line 303: Just curious why in some calls you use concatenation and others several arguments?

ematipico: Readability, mostly

florian-lefebvre (approved):

Issue #11044, opened 2024-05-14
Inspect Popup exceeds window


zaaakher:
### Astro Info

```block
Astro                    v4.8.3
Node                     v18.19.0
System                   Windows (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/react
                         @astrojs/tailwind
                         @astrojs/partytown
```

### If this issue only occurs in one browser, which browser is a problem?

Chrome

### Describe the Bug

When I try the inspect feature the popup exceeds the current window. I'm not sure if this happens because my site is RTL or not
(notice the scrollbar at the bottom is due to the popup)
![image](https://github.com/withastro/astro/assets/46135573/75c6799b-9f11-4ef5-bf8e-ea4199e9bcc4)

### What's the expected result?

It should stay within the window and pop to the left/right depending on conditional awareness of it's position

![image](https://github.com/withastro/astro/assets/46135573/ee07fa74-7f7e-489f-9d98-9cd061e42644)

### Link to Minimal Reproducible Example

n/a

### Participation

- [ ] I am willing to submit a pull request for this issue.

zaaakher:
My apologies @matthewp, here you go https://stackblitz.com/edit/github-5k1mkz?file=src%2Fpages%2Findex.astro

Here's a quick gif showing it:
![astro-rtl-inspector-popup](https://github.com/withastro/astro/assets/46135573/d8a29a96-cfb8-4a0f-95ca-b743472a4103)

V3RON:
@bluwy There is a check for right-side overflow, but not for left-side overflow. I'll open a PR adding one shortly.

PR #11081, opened 2024-05-17
fix(astro): positioning of inspection tooltip in RTL mode

V3RON:
## Changes

In RTL mode (`<body dir="rtl">` the inspection tooltip tend to overflow on the left side of the page.
This pull request contains a fix for that, which will make sure the tooltip stays in the viewport.

Before:

https://github.com/withastro/astro/assets/8137511/2815dd56-8f8e-4d7f-85ae-72a4f8026ee0

After:

https://github.com/withastro/astro/assets/8137511/4fbc3fb6-79f8-4b9f-a949-043568d7e18f

Closes https://github.com/withastro/astro/issues/11044

## Testing

Playwright test was added to make sure the tooltip stays in the viewport in all edge-cases (corners).
Note: There is one thing I am not fully satisfied with. I couldn't use visibility ratio of `1` (fully visible), because for some reason elements positioned absolutely overflow plus scrollbar is being displayed without reason. If anyone has an idea how to prevent that, please share it 🙏 

## Docs

No update needed.


bluwy (approved):
Looks great to me. I think another way to test it is doing something like:

https://github.com/withastro/astro/blob/4d32a8077b0d8b34581df3a81329a41a9f618420/packages/astro/e2e/dev-toolbar.test.js#L220-L226

But the current one is also fine by me.

Princesseuh (approved):
Looks great, thank you!

Issue #11033, opened 2024-05-13
Partytown scripts break ViewTransition 


dsegovia90:
### Astro Info

```block
Astro                    v4.8.3
Node                     v18.19.0
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/partytown
```

### Describe the Bug

Scripts with partytown break when transitioning between routes with ViewTransitions.

Console error (on multiple browsers):
```
Uncaught SyntaxError: redeclaration of const t
    <anonymous> index2 line 102 > injectedScript:1
    runScripts router.js:102
    transition router.js:335
    transition router.js:334
    navigate router.js:366
    <anonymous> ViewTransitions.astro:97
    EventListener.handleEvent* ViewTransitions.astro:59
[index2 line 102 > injectedScript:1:1](http://localhost:4321/index2%20line%20102%20%3E%20injectedScript)
```

This can be shown in the Minimal Reproducible Example attached.

### What's the expected result?

Partytown webworker should not break when transitioning routes with ViewTransition

### Link to Minimal Reproducible Example

https://astro-viewtransition-partytown-bug.vercel.app/

### Code

Repo: https://github.com/dsegovia90/astro-viewtransition-partytown-bug
Partytown script: https://github.com/dsegovia90/astro-viewtransition-partytown-bug/blob/57ac9a6a624284e57bd4163feb553d563a7772dd/src/layouts/Layout.astro#L22C3-L27C12

### Participation

- [ ] I am willing to submit a pull request for this issue.

V3RON:
The Partytown snippet is supposed to be loaded only once, but when the View Transitions API gets activated, the <script> element containing this snippet is recreated on every transition.

https://github.com/withastro/astro/blob/6f948903fcd33c660641203507240a8b5e99e7ac/packages/integrations/partytown/src/index.ts#L28-L38

I decided to comment out the line adding SELF_DESTRUCT_ON_VIEW_TRANSITION to the script and see what happens. The outcome is:
a) No error while transitioning between pages.
b) Partytown scripts are not executed after the transition.

That's probably expected. We should let Partytown know there may be new scripts added by dispatching new CustomEvent('ptupdate'). I did so, and nothing happened.

Under certain conditions, Partytown may need an <iframe> element to be present on the page. It's true for my local environment, and when a view transition happens, this <iframe> is removed from the DOM. 

We should persist it somehow 🤔

matthewp:
[...]

PR #11083, opened 2024-05-17
fix(@astrojs/partytown): prevent crashes when View Transitions are enabled

V3RON:
## Changes

This pull request intend to resolve the incompatibility of Partytown integration and View Transitions.

The Partytown snippet is supposed to be loaded only once, but when the View Transitions API gets activated, the <script> element containing this snippet is recreated on every transition. It has been changed and now it's executed only once.

Under certain conditions, Partytown may need an <iframe> element to be present on the page. When a view transition happens, this <iframe> is removed from the DOM. There has been a change added, which will move `<iframe>` element from the old document to the new one. The side-effect is all Partytown scripts will get re-executed. That behavior is aligned with other frameworks like Nuxt.

Closes https://github.com/withastro/astro/issues/11033

## Testing

There is no testing suite present for Partytown. I believe we should work on implementing one, but that's probably out of scope for this pull request.

## Docs

I'm not sure if docs update is needed, because Partytown never really worked with View Transitions 😅 On the other hand, it's worth mentioning that scripts will get re-executed on each transition.

### DIFF
--- a/.changeset/friendly-eagles-arrive.md
+++ b/.changeset/friendly-eagles-arrive.md
@@ -1,0 +1,9 @@
+---
+"@astrojs/partytown": patch
+---
+
+Prevent Partytown from crashing when View Transitions are enabled
+
+When View Transitions are turned on, Partytown executes on every transition.
+It's not meant to be like that, and therefore it breaks the integration completely.
+Starting from now, Partytown will be executed only once.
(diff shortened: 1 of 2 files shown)


dsegovia90:
Anything we can do to move this along? Looking forward to using it on one of our projects.

ematipico (approved):

PR #11088, opened 2024-05-17
Actions: Allow actions to be called on the server

bholmesdev:
## Changes

Allow actions to be called on the server.

```astro
---
// src/pages/blog/[postId].astro

await actions.like({ postId: Astro.params.postId });
---
```

- Ensure API context exists by wrapping the `next()` call from middleware with a context provider.
- Add an SSR code path to the `actions.mjs` template. This imports the `getAction()` utility to retrieve the appropriate action by path name.

## Testing

Add e2e test for calling the like action from astro frontmatter.

## Docs

TODO: RFC update

### DIFF
--- a/.changeset/eighty-taxis-wait.md
+++ b/.changeset/eighty-taxis-wait.md
@@ -1,5 +1,5 @@
 ---
-"astro": minor
+"astro": patch
 ---
 
 Allow actions to be called on the server. This allows you to call actions as utility functions in your Astro frontmatter, endpoints, and server-side UI components.
--- a/.changeset/eighty-taxis-wait.md
+++ b/.changeset/eighty-taxis-wait.md
@@ -1,0 +1,16 @@
+---
+"astro": minor
+---
+
+Allow actions to be called on the server. This allows you to call actions as utility functions in your Astro frontmatter or server-side components.
+
+Import and call directly from `astro:actions` as you would for client actions:
+
+```astro
+---
+// src/pages/blog/[postId].astro
+import { actions } from 'astro:actions';
+
+await actions.like({ postId: Astro.params.postId });
+---
+```
--- a/.changeset/eighty-taxis-wait.md
+++ b/.changeset/eighty-taxis-wait.md
@@ -2,7 +2,7 @@
 "astro": minor
 ---
 
-Allow actions to be called on the server. This allows you to call actions as utility functions in your Astro frontmatter or server-side components.
+Allow actions to be called on the server. This allows you to call actions as utility functions in your Astro frontmatter, endpoints, and server-side UI components.
 
 Import and call directly from `astro:actions` as you would for client actions:
 
--- a/packages/astro/src/actions/runtime/middleware.ts
+++ b/packages/astro/src/actions/runtime/middleware.ts
@@ -28,7 +28,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 	// See https://github.com/withastro/roadmap/blob/feat/reroute/proposals/0047-rerouting.md#ctxrewrite
 	// `_actionsInternal` is the same for every page,
 	// so short circuit if already defined.
-	if (locals._actionsInternal) return next();
+	if (locals._actionsInternal) return ApiContextStorage.run(context, () => next());
 
 	const { request, url } = context;
 	const contentType = request.headers.get('Content-Type');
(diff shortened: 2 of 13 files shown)


.changeset/eighty-taxis-wait.md
```diff
@@ -0,0 +1,16 @@
+---
+"astro": minor
+---
+
+Allow actions to be called on the server. This allows you to call actions as utility functions in your Astro frontmatter or server-side components.
```
florian-lefebvre, line 5: Does it also allow using them in endpoints?

packages/astro/src/actions/runtime/middleware.ts
florian-lefebvre, line 1: Does it mean `getApiContext` could technically be called outside of actions?

packages/astro/src/core/render-context.ts
```diff
@@ -8,7 +8,7 @@ import type {
 	RouteData,
 	SSRResult,
 } from '../@types/astro.js';
-import type { ActionAPIContext } from '../actions/runtime/store.js';
+import { type ActionAPIContext } from '../actions/runtime/store.js';
```
florian-lefebvre, line 11: Why this change?

.changeset/eighty-taxis-wait.md
bholmesdev: Yup! Just didn’t expect that use case as much. I’ll update the changeset

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Technically yes… but I’m not a huge fan that. Relying on local storage can be brittle for custom functions. I’m considering moving the `APIContext` back to the second handler parameter, and making this function internal only. Thoughts?

packages/astro/src/core/render-context.ts
bholmesdev: My editor being dumb

packages/astro/src/actions/runtime/middleware.ts
florian-lefebvre: I think it would be better yeah, I wonder if it should be the first or second parameter, I guess it depends on what we expect people to use more

bholmesdev: Erm, well if people are reaching for context more than the input we designed something by wrong 😅 But yeah open to what feels natural to you

florian-lefebvre: I think `(input, context)` makes sense 👍

.changeset/eighty-taxis-wait.md
```diff
@@ -0,0 +1,16 @@
+---
+"astro": minor
```
ematipico, line 2: Since actions are experimental, I think we can flag this as `patch`. But I leave the decision to you

packages/astro/src/actions/runtime/middleware.ts
```diff
@@ -47,15 +46,15 @@ export const onRequest = defineMiddleware(async (context, next) => {
 		},
 	};
 	Object.defineProperty(locals, '_actionsInternal', { writable: false, value: actionsInternal });
-	return next();
+	return ApiContextStorage.run(context, () => next());
```
ematipico, line 49: Just a word of warning... when users use the rewrites APIs inside the middleware, under the hood, we create a new `APIContext`; there's a chance that this `context` could be outdated in case users decide to use the rewrite APIs:

```js
next("/new-url") // new API, it will generate a new context with a new `Request`, and the URL is `/new-url`
```

ematipico (approved):

packages/astro/src/actions/runtime/middleware.ts
bholmesdev: Ah, that's a good catch! I _think_ applying the `APIContext` to this case would resolve our issue, since it only gets called when a rewrite is detected: https://github.com/withastro/astro/pull/11088/files#diff-a07f25105bdef494ce45893f063230e218d44b36cd7db7c9c8d6447cb03db136R19

I'll need a unit test to confirm

ematipico: It's also true that having a rewrite inside an action feels weird, not sure there's a valid case for that. I could create a stackblitz after the release and see how that works

bholmesdev: Added a test. It works!

Issue #10970, opened 2024-05-08
`astro:db`: `db.run` queries are giving different responses from Astro Studio and from local files


timsexperiments:
### Astro Info

```block
Astro                    v4.7.1
Node                     v22.0.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             astro:db
                         @astrojs/db/file-url
                         @astrojs/tailwind
```

### Describe the Bug

When running queries with `db.run`, the result returned from the query differs on file and Astro DB databases. When running a query to a file database (local / using the `ASTRO_DATABASE_FILE` env var) the query rows are properly aliased with both the index and the alias for the selected column.

Another observation I have noticed is that with the file-based database response, the `result.rows` values are not iterable where they are iterable from the Astro Studio version. This means that destructuring a row by `[first, second, ...] = result.rows[0]` will not work on file-based database but will work with an Astro Studio DB.

Assuming the following code was run:

```TypeScript
import { db, sql } from  'astro:db';

const result = db.run(sql`SELECT 1 as value`);
```

The following is an example of what works with the different configurations:

|                         | Example Code                  | Astro Studio Database `value` Result | File-Based Database`value` Result                                       |
|-------------------------|-------------------------------|-----------------------------|-----------------------------------------------------------------|
| Access through alias    | `const { value } = result`    | undefined                   | 1                                                               |
| Access through index    | `const { 0: value } = result` | 1                           | 1                                                               |
| Access through iterable | `const [value] = result`      | 1                           | Throws: TypeError: object is not iterable (cannot read property |

### What's the expected result?

I would expect:

1. The values are to be accessible through both the alias and the index, no matter what type of database is being used through the Astro DB API.

2. Either both results to be iterable or neither to be iterable.

### Link to Minimal Reproducible Example

https://github.com/timsexperiments/astro-sql-file-vs-remote

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #11091, opened 2024-05-17
Fix inconsistent results from raw SQL

matthewp:
## Changes

- Locally drizzle turns each row into an object when creating the result. Remote, which runs regular SQL, does not.
- So this changes how remote deserializes the result, only for raw SQL, so that it is also an object.
- Fixes https://github.com/withastro/astro/issues/10970

## Testing

- New test case added

## Docs

N/A, bug fix

### DIFF
--- a/packages/db/src/runtime/db-client.ts
+++ b/packages/db/src/runtime/db-client.ts
@@ -71,7 +71,20 @@ export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string
 				});
 			}
 
-			if (method === 'run') return remoteResult;
+			if (method === 'run') {
+				// Using `db.run()` drizzle massages the rows into an object.
+				// So in order to make dev/prod consistent, we need to do the same here.
+				// This creates an object and loops over each row replacing it with the object.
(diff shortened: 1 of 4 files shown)


packages/db/src/runtime/db-client.ts
```diff
@@ -71,7 +71,23 @@ export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string
 				});
 			}
 
-			if (method === 'run') return remoteResult;
+			if (method === 'run') {
+				const rawRows = Array.from(remoteResult.rows);
+				// in order to serialize properly much implement toJSON
```
bholmesdev, line 76:
```suggestion
				// Implement basic `toJSON()` for Drizzle to serialize properly
```

Cleaning up this comment a little. Am I right this the reason for `toJSON()`? For Drizzle to call internally?

bholmesdev (approved):
On a `row`ll!

packages/db/src/runtime/db-client.ts
matthewp: Yep, thanks!

ematipico (approved):

PR #11092, opened 2024-05-17
fix type definition of IntrinsicAttributes's slot attribute

### LINKED ISSUE #11072: IntrinsicAttributes and HTMLAttributes have conflicts
### Astro Info

```block
Astro                    v4.8.4
Node                     v21.6.2
Package Manager          npm
Adapter                  none
                         astro:db
                         @astrojs/db/file-url
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

astroHTML.JSX.IntrinsicAttributes and HTMLAttributes conflicting type definitions (I found one at least):
```
astroHTML.JSX.IntrinsicAttributes {
...
    slot: string | undefined
...
}

HTMLAttributes{
...
    slot: string | undefined | null
...
}
```

If I were to define for two components `MyComponent` and `MyComponentWrapper` a Props interface extending from HTMLAttributes<'div'> for example, I can't forward props to `MyComponent` by spreading the props received by MyComponentWrapper, even tho they should be of the same type: I get the error "Type { .... } is not assignable to type 'IntrinsicAttributes & Props'. " 

Error on local machine but not on StackBlitz example, please try it locally by using the example provided inside StackBlitz; I get the same error both on Windows (WSL) and MacOS:
```
Type '{ children: any; class: string; 'class:list'?: string | Record<string, boolean> | Record<any, any> | Iterable<string> | Iterable<any> | undefined; ... 187 more ...; onfullscreenerror?: string | ... 1 more ... | undefined; }' is not assignable to type 'IntrinsicAttributes & Props'.
  Type '{ children: any; class: string; 'class:list'?: string | Record<string, boolean> | Record<any, any> | Iterable<string> | Iterable<any> | undefined; ... 187 more ...; onfullscreenerror?: string | ... 1 more ... | undefined; }' is not assignable to type 'IntrinsicAttributes'.ts(2322)
```
```

// this gives error, most likely because of the type difference of slot between IntrinsicAttributes and HTMLAttributes
---
interface Props extends HTMLAttributes<'html'>{};
const props = Astro.props;
---

<MyComponent {...props}> 
...


// by guarding from slot equal to null the error goes away indeed
---
interface Props extends HTMLAttributes<'html'>{};
const props = Astro.props;
const slot = props.slot || undefined // type of slot is then string | undefined
---

<MyComponent {...props} slot={cleanedSlot}> 
...


// or even like this, but this removes the slot prop altogether
---
interface Props extends HTMLAttributes<'html'>{};
const {slot, ...props} = Astro.props;
---

<MyComponent {...props}> 
...
```

I would expect HTMLAttributes and IntrinsicAttributes to not have conflicting definitions, if I'm not missing something, else I would have to always type guard the slot attribute every time I extend from HTMLAttributes and pass props to another component whose Props interface also extends from HTMLAttributes

https://stackblitz.com/edit/withastro-astro-qnckaa?file=src%2Flayouts%2FDivWrapper.astro

### Participation

duckycoding-dev:
## Changes

Closes #11072 
- What does this change?
Changed IntrinsicAttributes's slot attribute type, adding `null` as a possible type
 ```
interface IntrinsicAttributes {
....
slot?: string | undefined | null
....
}
```
Now `slot` is consistent in IntrinsicAttributes and HTMLAttributes

## Testing

None, since I only widened the type choice

## Docs

None, change doesn't affect docs

### DIFF
--- a/.changeset/new-mice-occur.md
+++ b/.changeset/new-mice-occur.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Change `slot` attribute of `IntrinsicAttributes` to match the definition of `HTMLAttributes`'s own `slot` attribute of type `string | undefined | null`
--- a/packages/astro/astro-jsx.d.ts
+++ b/packages/astro/astro-jsx.d.ts
@@ -22,7 +22,7 @@ declare namespace astroHTML.JSX {
 		extends AstroBuiltinProps,
 			AstroBuiltinAttributes,
 			AstroClientDirectives {
-		slot?: string;
+		slot?: string | undefined | null;
 		children?: Children;
 	}
 


Princesseuh (approved):
Looks good to me! Thank you for contributing this.

PR #11096, opened 2024-05-17
Support deprecating Web Vitals table

delucis:
## Changes

- Adds an option to the web vitals integration that allows users to “deprecate” the Astro DB tables web vitals adds.
- Currently, Astro DB requires a user to set `deprecated: true` on a table and push that before it will allow you to remove it completely. Because integrations control the tables themselves this is not something users can do themselves, so this PR reflects that up to the integration options.
- See the README changes in this PR for a full description of the “uninstall” process. (It’s not amazing, might be something for us to be aware of/think about.)
 
## Testing

Tested manually that the `deprecated` property is being set properly on the table config.

## Docs

/cc @withastro/maintainers-docs for feedback on the README changes!

### DIFF
--- a/packages/integrations/web-vitals/src/index.ts
+++ b/packages/integrations/web-vitals/src/index.ts
@@ -2,7 +2,8 @@ import { defineDbIntegration } from '@astrojs/db/utils';
 import { AstroError } from 'astro/errors';
 import { WEB_VITALS_ENDPOINT_PATH } from './constants.js';
 
-export default function webVitals() {
+export default function webVitals({ deprecated }: { deprecated?: boolean } = {}) {
+	process.env.DEPRECATE_WEB_VITALS = deprecated ? 'true' : undefined;
 	return defineDbIntegration({
 		name: '@astrojs/web-vitals',
 		hooks: {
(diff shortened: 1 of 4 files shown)


packages/integrations/web-vitals/src/index.ts
```diff
@@ -2,7 +2,8 @@ import { defineDbIntegration } from '@astrojs/db/utils';
 import { AstroError } from 'astro/errors';
 import { WEB_VITALS_ENDPOINT_PATH } from './constants.js';
 
-export default function webVitals() {
+export default function webVitals({ deprecated }: { deprecated?: boolean } = {}) {
+	process.env.DEPRECATE_WEB_VITALS = deprecated ? 'true' : undefined;
```
ematipico, line 6: Shouldn't we set this only when running a `db push`? Or is it intentional?

ematipico (approved):
I wish npm would allow some `preunitinstall` scripts, so libraries can clean after themselves. I suppose that's the best we can do for now.

matthewp (approved):

packages/integrations/web-vitals/src/index.ts
delucis: I guess you might want it in `dev` etc. too? If you happened to run it while deprecated. Probably doesn’t matter either way though — I just wanted it to match how a user-defined table might behave really.

sarah11918 (approved):
README and changeset LGTM, Chris!

PR #11098, opened 2024-05-18
Expose `inferRemoteSize` function

itsmatteomanf:
## Changes

This change exposes the old `inferSize()`, renamed to `inferRemoteSize()` per https://discord.com/channels/830184174198718474/1240790957591367701.

## Testing

No additional testing needed, it was never tested directly.

## Docs

https://github.com/withastro/docs/pull/8808


packages/astro/src/assets/utils/remoteProbe.ts
```diff
@@ -1,7 +1,7 @@
 import { lookup } from './vendor/image-size/lookup.js';
 import type { ISize } from './vendor/image-size/types/interface.ts';
 
-export async function probe(url: string): Promise<ISize> {
+export async function inferRemoteSize(url: string): Promise<ISize> {
```
Princesseuh, line 4: I don't really like that we're returning a type from `image-size` here, it means that `image-size` updates might require a minor or major for us in case it changes. Not to mention, it restricts us to `image-size`.

Maybe what should be exported is a wrapper of this function that return some other interface.

itsmatteomanf: What would you suggest? Just separate out the interface and duplicate the `image-size` types for now? Like a `RemoteISize` type?

itsmatteomanf: I presume I'd manually parse out the important bits so that it can be changed freely? Or just straight up return the same thing now, with the other type and allow for changes?

Princesseuh: Yeah, you can fish out the import bits. This is similar to how imageMetadata works here: https://github.com/withastro/astro/blob/main/packages/astro/src/assets/utils/metadata.ts

itsmatteomanf: I'm confused though, I do want `width`, `height`, and `type`, and you told that `orientation` sometimes shows up, these are all the results I want.

It'd duplicate the exact same type... I did remove the `orientation` property as i never saw it exposed, but the rest is the same...

Princesseuh: It's about not exposing to users a type we don't own unnecessarily. The content of the type is correct, it's just it's provenance that's annoying.

If `image-size` release an update tomorrow that change, I don't know, `type` to `format`, we can't update it in Astro without releasing a new major version of Astro.

itsmatteomanf: Ohhh, never mind, I didn't realize it wasn't fully owned by Astro.

So just probably use ImageMetadata, with the `Omit<>`...

itsmatteomanf: It's so annoying though, ImageMetadata uses format, ISize uses type 🙈

Princesseuh:
Refactored it a bit, the only thing missing will be adding better error messages now that this is exposed and stuff. I'll do that later, thank you for contributing!

itsmatteomanf:
Thank you, as always, @Princesseuh!

ematipico:
@itsmatteomanf please ping me when this PR is ready to be shipped. Do you think it will be ready for the next minor?

itsmatteomanf:
@ematipico as far as I am concerned it's ready, I know @Princesseuh had some changes in mind though, but said she was going to do them herself...

Princesseuh:
Yeah, will tackle when I'm back from holidays if no one does. The changes required are just making the errors better, now that the feature is exposed. Docs might be great as well!

.changeset/large-geese-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": minor
+---
+
+Exposes the `inferRemoteSize` function used by `inferSize`
```
sarah11918, line 5: Can we have a description here that describes how this affects user behaviour? What it now allows them to do, or why this change was made that will help someone in their project?

With new features introduced in a minor we often do have slightly longer changelog messages, so even an example of using this would be helpful here!

sarah11918 (changes requested):
Blocking this while we figure out what docs are needed!

The current docs for `inferSize` are here: https://docs.astro.build/en/guides/images/#infersize so this would be a place to check.

Also checking that we don't need any new error messages associated with making this available to users now?

Additionally, adding the necessary context (why this PR was made, what this allows you to do) inside the PR description rather than linking to a Discord message (not publicly available) is helpful! This would have helped me suggest an appropriate changeset message.

.changeset/large-geese-play.md
Princesseuh: Done! 👍

Princesseuh:
Hmm, okay there's a problem with exporting it from `astro:assets`, i'll keep investigating and see if I can fix this before release.

.changeset/large-geese-play.md
```diff
@@ -0,0 +1,16 @@
+---
+"astro": minor
+---
+
+Adds a new `inferRemoteSize` function that can be used to infer the dimensions of a remote image outside of the Image and Picture components. This is useful for when you need to know the dimensions of an image for styling purposes or to calculate different densities for responsive images.
```
sarah11918, line 5:
```suggestion
Adds a new `inferRemoteSize()` function that can be used to infer the dimensions of a remote image.

 Previously, the ability to infer these values was only available by adding the [`inferSize`] attribute to the `<Image>` and `<Picture>` components. Now, you can access this data outside of these components. 
 
 This is useful for when you need to know the dimensions of an image for styling purposes or to calculate different densities for responsive images.
```

What do you think about something like this? I think maybe it highlights for people skimming and thinking, wait, I thought I could already infer remote dimensions?" and emphasizes that this is a function, and "not just in the components anymore" aspect.

sarah11918 (commented):
Fantastic example, thank you @Princesseuh ! One quick thought for your consideration, but happy for you to make the final call.

.changeset/large-geese-play.md
Princesseuh: Fine with me! One note is that inferSize can also be used with getImage, not just the components

packages/astro/src/assets/utils/index.ts
```diff
@@ -1,4 +1,4 @@
-export { emitESMImage } from './emitAsset.js';
+export { emitESMImage } from './node/emitAsset.js';
```
Princesseuh, line 1: So this is annoying, `emitAsset` (obviously) isn't non-Node safe, but I exported it from the utils without thinking a while ago, and now it means that `astro:assets` can't actually import from `astro/assets/utils` otherwise it breaks runtime.

For now, I moved `emitAsset` to its own folder to reduce damage, and in the future I'll update the Markdoc integration and stuff to import it from the proper place.

Princesseuh (approved):

.changeset/large-geese-play.md
```diff
@@ -0,0 +1,20 @@
+---
+"astro": minor
+---
+
+Adds a new `inferRemoteSize()` function that can be used to infer the dimensions of a remote image.
+
+ Previously, the ability to infer these values was only available by adding the [`inferSize`] attribute to the `<Image>` and `<Picture>` components or `getImage()`. Now, you can access this data outside of these components. 
+ 
+ This is useful for when you need to know the dimensions of an image for styling purposes or to calculate different densities for responsive images.
```
sarah11918, line 9:
```suggestion
Previously, the ability to infer these values was only available by adding the [`inferSize`] attribute to the `<Image>` and `<Picture>` components or `getImage()`. Now, you can also access this data outside of these components. 
 
This is useful for when you need to know the dimensions of an image for styling purposes or to calculate different densities for responsive images.
```
stupid spaces

sarah11918 (approved):
Approving for Docs!

PR #11120, opened 2024-05-22
Let web vitals route handle all requests under that path

delucis:
## Changes

- Fixes requests to the web vitals endpoint in setups like Vercel’s `trailingSlash: true` that redirect from `/web-vitals` to `/web-vitals/`
- I’m not 100% sure if this is a larger bug in Astro’s route handling (I’d kind of expect a route injected at `/web-vitals` to handle `/web-vitals/`), or just Vercel being fussy — I’ve definitely found their `trailingSlash` handling to be much more fragile than other platforms

## Testing

Applied as a patch in some repos using the current integration version.

## Docs

n/a — bug fix

### DIFF
--- a/.changeset/three-bikes-stare.md
+++ b/.changeset/three-bikes-stare.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/web-vitals": patch
+---
+
+Fixes requests to the web vitals endpoint in setups like Vercel’s `trailingSlash: true` that redirect from `/web-vitals` to `/web-vitals/`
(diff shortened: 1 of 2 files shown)


florian-lefebvre (approved):
I think I encountered the same issue a while ago and added a fix here https://github.com/withastro/astro/blob/main/packages/integrations/vercel/src/serverless/adapter.ts#L235-L253

delucis:
> I think I encountered the same issue a while ago

Ah interesting. That’s setting _Astro’s_ `trailingSlash` though right? The issue here appears to be that when Vercel’s `trailingSlash` is enabled:

- request `/web-vitals` => Vercel redirects to `/web-vitals/` => that’s a 404 for some reason

florian-lefebvre:
What value for `trailingSlash` are you using in the Astro config?

ematipico (approved):
I am not sure this is the proper fix, as Florian suggested. I think we can accept this as a quick patch, but we should circle back and properly fix it.

delucis:
> What value for `trailingSlash` are you using in the Astro config?

None!

PR #11134, opened 2024-05-23
feat: 500.astro improvements

florian-lefebvre:
## Changes

- RFC: https://github.com/withastro/roadmap/discussions/925
- Pass an error to `500.astro` props
- Render the custom 500 error page in dev if `ASTRO_CUSTOM_500` env var is set to `'true'`

## Testing

Adds tests to check:
1. If the custom 500.astro is rendered in dev/preview
2. If the default error page (error overlay in dev, nothing in preview) is rendered if the custom 500 throws

## Docs

https://github.com/withastro/docs/pull/8392


packages/astro/src/core/render-context.ts
```diff
@@ -93,18 +93,22 @@ export class RenderContext {
...
+			props,
+			await getProps({
+				mod: componentInstance,
+				routeData: this.routeData,
+				routeCache,
+				pathname,
+				logger,
+				serverLike,
+			})
+		);
```
ematipico, line 111: `props` are now part of the class `RenderContext`, so we should update this code

packages/astro/src/core/app/index.ts
```diff
@@ -68,6 +68,7 @@ export interface RenderErrorOptions {
 	 * Whether to skip middleware while rendering the error page. Defaults to false.
 	 */
 	skipMiddleware?: boolean;
+	error?: unknown;
```
ematipico, line 71: Let's start some docs whenever we can, at least when we add new stuff

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
...
+ASTRO_CUSTOM_500=true
+```
+
+Or inline:
+
+```sh
+ASTRO_CUSTOM_500=true astro dev
+```
+
+Note that if an error occur in this file, it will default to the host 500 error page in SSR, and the error overlay in development.
```
ematipico, line 34:
``````suggestion
Additionally, you can now use your custom 500 error page in development by setting the `ASTRO_CUSTOM_500` environment variable to `'true'`. For example, using a `.env` file:

```ini
ASTRO_CUSTOM_500=true
```

Or inline:

```sh
ASTRO_CUSTOM_500=true astro dev
```

Note that if an error occurs in this file, it will default to the host 500 error page in SSR and the error overlay in development.
``````

File: .changeset/rich-dolls-compete.md [hunk 4]
```diff
@@ -0,0 +1,34 @@
+---
+"astro": minor
+---
+
+Improves DX around `500.astro`
+
+The special `src/pages/500.astro` page now accepts an error as a prop. It can be anything so make sure to handle it properly:
```
ematipico, line 7:
```suggestion
Improves the developer experience of the `500.astro` file.

The special `src/pages/500.astro` page now accepts an error as a prop. It can be anything, so make sure to handle it properly:
```

ematipico (approved):

packages/astro/test/custom-500.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+			const html = await response.text();
+			const $ = cheerio.load(html);
+
+			assert.equal($('h1').text(), 'Server error');
+			assert.equal($('p').text(), 'some error');
+		});
+	});
+
+	describe('preview', () => {
```
ematipico, line 54:
```suggestion
	describe('SSR', () => {
```

packages/astro/src/core/app/index.ts
```diff
@@ -335,6 +340,7 @@ export class App {
 				locals,
 				response,
 				status: response.status as 404 | 500,
+				error: response.status === 500 ? null : undefined,
```
ematipico, line 343: What's the difference between `null` and `undefined` **here**? It's a weird case. Maybe we should leave a comment that explains it

ematipico (approved):

packages/astro/src/core/app/index.ts
florian-lefebvre: Basically at this point of the code, we don't really have an error to report. So passing `null` means we pass nothing intentionally while `undefined` means there's no error. Hope that makes sense 😅 but ideally we can pass an error there, whatever it is

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
+---
+"astro": minor
+---
+
+Improves the developer experience of the `500.astro` file.
```
sarah11918, line 5:
```suggestion
Improves the developer experience of the `500.astro` file by passing it a new `error` prop.
```

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
...
+
+The special `src/pages/500.astro` page now accepts an error as a prop. It can be anything, so make sure to handle it properly:
```
sarah11918, line 7:
```suggestion
When an error is thrown, the special `src/pages/500.astro` page now automatically receives the error message as a prop. This allows you to display more specific information about the error on a custom 500 page.
```
For the changeset, (vs full complete docs), I wonder if we can get away with not having the second paragraph? I don't like "handle it properly" without getting into an explanation of what they need to worry about/do, and we're writing that in docs anyway. This is just the taste of what's new and cool!

(Also, is the above correct? I like focusing on the use of it)

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
...
+ASTRO_CUSTOM_500=true
+```
+
+Or inline:
+
+```sh
+ASTRO_CUSTOM_500=true astro dev
+```
+
+Note that if an error occurs in this file, it will default to the host 500 error page in SSR and the error overlay in development.
```
sarah11918, line 34:
```suggestion
If an error occurs rendering this page, your host's default 500 error page will be shown to your visitor in production, and Astro's default error overlay will be shown in development.
```
This is the wording we took from docs, so we can reuse that here!

sarah11918 (commented):
Quick suggestions for your consideration here, @florian-lefebvre ! See what you think, and adjust any of them as necessary.

sarah11918 (approved):
Approving for docs!! :partying_face:

packages/astro/src/core/app/index.ts
ematipico: This is valuable information! We should have this very comment in the code :)

florian-lefebvre:
```suggestion
				// We don't have an error to report here. Passing null means we pass nothing intentionally
				// while undefined means there's no error
				error: response.status === 500 ? null : undefined,
```

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
+---
+"astro": minor
+---
+
+Improves the developer experience of the `500.astro` file by passing it a new `error` prop.
+
+When an error is thrown, the special `src/pages/500.astro` page now automatically receives the error as a prop. This allows you to display more specific information about the error on a custom 500 page.
```
bluwy, line 7: Should we document that the `error` prop only works when dynamically rendering `500.astro`? Since we can't get an error after it's prerendered.

Also, perhaps docs-specific but we should also caution about printing the error entirely as stuff like `.stack` may expose the server implementation.

florian-lefebvre: True, I'll update both changeset and docs

.changeset/rich-dolls-compete.md
```diff
@@ -0,0 +1,34 @@
...
+    error: unknown
+}
+
+const { error } = Astro.props
+---
+
+<div>{error instanceof Error ? error.message : 'Unknown error'}</div>
+```
+
+Additionally, you can now use your custom 500 error page in development by setting the `ASTRO_CUSTOM_500` environment variable to `'true'`. For example, using a `.env` file:
```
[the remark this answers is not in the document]
florian-lefebvre: This is not true anymore after a PR from Ema, I need to update the PR

PR #11139, opened 2024-05-24
fix(upgrade): use preferred-pm instead of which-pm-runs

Princesseuh:
## Changes

@astrojs/upgrade wrongfully used the current package manager used to run the command instead of the package manager used to install the project its run in. This normally would be fine, however many people uses `npx` even when using pnpm and other package managers, so using the package manager of the project ends up working better

image

Notice it now using pnpm despite using npx in this screenshot.

## Testing

Tested manually, see screenshot

## Docs

N/A

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5691,22 +5691,22 @@ importers:
       '@astrojs/cli-kit':
         specifier: ^0.4.1
         version: 0.4.1
+      preferred-pm:
+        specifier: ^3.1.3
+        version: 3.1.3
       semver:
         specifier: ^7.6.2
         version: 7.6.2
       terminal-link:
         specifier: ^3.0.0
         version: 3.0.0
-      which-pm-runs:
-        specifier: ^1.1.0
-        version: 1.1.0
     devDependencies:
+      '@types/preferred-pm':
+        specifier: ^3.0.0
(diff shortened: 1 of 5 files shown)


delucis (approved):
Makes sense to me 👍

ematipico (approved):

pnpm-lock.yaml
```diff
@@ -8771,6 +8771,13 @@ packages:
     resolution: {integrity: sha512-37i+OaWTh9qeK4LSHPsyRC7NahnGotNuZvjLSgcPzblpHB3rrCJxAOgI5gCdKm7coonsaX1Of0ILiTcnZjbfxA==}
     dev: true
 
+  /@types/preferred-pm@3.0.0:
+    resolution: {integrity: sha512-Ub1de7EkdavsyM1KNrTb1K1QL+ISepEELELh2QWccyDcVEcyUDiGoYzzOJfonpGNwpymYXY13oRFpXQluGOC5w==}
+    deprecated: This is a stub types definition. preferred-pm provides its own type definitions, so you do not need this installed.
```
gacek1123, line 8776: Seems that the types package isn't needed?

Princesseuh: Good catch, I'll update this when possible!

PR #11141, opened 2024-05-24
fix(runtime): check head+content component before throwing an error

ematipico:
## Changes

This PR fixes an issue in our runtime engine. This fix allows the `AstroContainer` to render the `Container` component.

## Testing

I tested it manually against the source code that Chris created yesterday.

## Docs

We could potentially create a PR to update the RSS page and use the container to render the content collection.

### DIFF
--- a/.changeset/silver-bananas-move.md
+++ b/.changeset/silver-bananas-move.md
@@ -1,0 +1,13 @@
+---
+"astro": patch
+---
+
+Fixes an internal error that prevented the `AstroContainer` to render the `Content` component.
+
+You can now write code similar to the following to render content collections:
+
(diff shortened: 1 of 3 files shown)


matthewp (approved):

Princesseuh:
A test would be nice imo

ematipico:
> A test would be nice imo. With people using this a bunch for RSS, it's important that it works and stay working.

Ok, let me see if I can pull it off

sarah11918:
I'm going to cc @delucis in here re: the docs question. It might not be worth changing the RSS docs just yet, unless this *prevents* the current method from working.

PR #11146, opened 2024-05-25
feat: upgrade pnpm

### LINKED ISSUE #11130: Repo can't be opened on GitHub codespaces
### Astro Info

```block
-
```

_No response_

### Describe the Bug

The dev container fails to build because the monorepo requires an outdated version of pnpm.

The repo can be opened on GitHub codespaces.

https://github.com/codespaces/new?hide_repo_select=true&ref=main&repo=348060227&machine=standardLinux32gb

### Participation

florian-lefebvre:
## Changes

- Closes #11130
- Upgrade pnpm to latest
- Allows renovate to update pnpm

## Testing

Tests should still pass. Smoke tests are failing because docs pnpm version need to match (https://github.com/withastro/docs/pull/8394)

## Docs

N/A, no changeset


packages/astro/package.json
```diff
@@ -234,7 +234,7 @@
   "engines": {
     "node": "^18.17.1 || ^20.3.0 || >=21.0.0",
     "npm": ">=9.6.5",
-    "pnpm": ">=7.1.0"
+    "pnpm": ">=9.1.2"
```
bluwy, line 237: I don't think we should change this as the `engines` field is for users installing this package, and we say that "you can only download `astro` with pnpm >=7.1.2". We don't want to restrict this to latest.

(Personally I'd remove this line altogether as I'm not sure how useful this way, but probably another topic)

package.json
```diff
@@ -45,9 +45,9 @@
   ],
   "engines": {
     "node": "^18.17.1 || ^20.3.0 || >=21.0.0",
-    "pnpm": ">=8.6.12"
+    "pnpm": ">=9.1.2"
```
bluwy, line 48: We could probably remove this altogether as we don't publish this `package.json`. The `packageManager` hint below should be enough.

bluwy (commented):
One of the big change with pnpm v9 is that it will validate `packageManager` matching your current running pnpm version before executing, and error if not matching. This means everyone working on this monorepo should be using corepack.

I'd probably like to hear thoughts on this if this affects anyone. Otherwise we can disable this restriction with https://pnpm.io/npmrc#engine-strict. I heard corepack can be tricky to use if you use pnpm to manage Node.js versions.

Also, we should update `CONTRIBUTING.md` about the latest pnpm version. There's a lot of logistics when upgrading pnpm hence why I disabled it in renovate 😅

florian-lefebvre:
I'm using corepack to manage pnpm and never had any issues tbh, curious to know what issues people have been facing

packages/astro/package.json
florian-lefebvre: Do you know if there's a way with renovate to prevent auto-updating pnpm for this package?

bluwy:
I'm ok with recommending corepack, and even requiring it if others are fine with it.

bluwy (approved):
I forgot about this. LGTM

ematipico (approved):
Let's 🐑 it

PR #11147, opened 2024-05-25
fix(assets): ensure valid mime types in picture component

### LINKED ISSUE #11135: Invalid MIME type in Picture component
### Astro Info

```block
Astro                    v4.9.0
Package Manager          pnpm
Adapter                  none
                         @astrojs/sitemap
                         @astrojs/svelte
```

_No response_

### Describe the Bug

The Picture component calculates a `type` attribute for the `<source>` element as a simple concatenation of a format string from `getImage()`.

Here's the relevant section from Picture.astro:

https://github.com/withastro/astro/blob/4c7949d1ae28f5597ec08cab788fe683197b0c3c/packages/astro/components/Picture.astro#L100-L104

This can lead to cases where the MIME type is invalid:

| Image Extension | Picture MIME Type | Valid? | Valid MIME type |
| --------------- | ----------------- | ------ | --------------- |
| `avif`          | `image/avif`      | ✅     |                 |
| `png`           | `image/png`       | ✅     |                 |
| `webp`          | `image/webp`      | ✅     |                 |
| `jpeg`          | `image/jpeg`      | ✅     |                 |
| `jpg`           | `image/jpg`       | ❌     | [`image/jpeg`](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#jpeg_joint_photographic_experts_group_image)    |
| `svg`           | `image/svg`       | ❌     | [`image/svg+xml`](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#svg_scalable_vector_graphics) |

(Plus, hypothetically, other extensions since `ImageOutputFormat`'s type allows arbitrary `string` values.)

An invalid MIME type means that the affected `<source>` element will be _always_ be skipped over by the browser, leading to use of the fallback image in circumstances where a `<source>` option was more appropriate.

Valid values in the `type` attribute, ensuring correct selection of `<source>` elements.

https://stackblitz.com/edit/github-spdq1j?file=src%2Fpages%2Findex.astro

### Participation

kitschpatrol:
## Changes

- Closes #11135 
- Use existing  `mrmime` dependency to ensure valid mapping between file name extensions and image MIME types

## Testing

Added MIME type test to `core-image.test.js` alongside existing Picture component validation tests.

## Docs

N/A, the Picture component should simply now work as expected.


packages/astro/components/Picture.astro
```diff
@@ -99,7 +100,7 @@ if (import.meta.env.DEV) {
 			return (
 				<source
 					srcset={srcsetAttribute}
-					type={'image/' + image.options.format}
+					type={mime.lookup(image.options.format ?? '') ?? `image/${image.options.format}`}
```
Princesseuh, line 103: Instead of `''`, you can pass the src of the image. mrmime should be able to handle it

Princesseuh:
This is fine with me, I think that ultimately there's a potential better fix in the image service API itself, but that's for me to think about in the future - this is great!

packages/astro/components/Picture.astro
kitschpatrol: Ok! That was to appease TypeScript, since `lookup`'s definition only accepts a string, but I've confirmed that mrmime correctly returns `undefined` if given `undefined`.

kitschpatrol: Oh I see pass the source, read too quickly.

kitschpatrol: Hmm passing `src` works in production builds, but in dev `mrmime` doesn't know what to make of the src URL:

```ts
import * as mrmime from 'mrmime'

// Build
mrmime.lookup('http://localhost:4321/_astro/penguin1.wSAOfMN3_ZYmXno.jpg')
// image/jpeg


// Dev
mrmime.lookup('http://localhost:4321/_image?href=%2F%40fs%2FUsers%2Fmika%2FCode%2Fastro%2Fpackages%2Fastro%2Ftest%2Ffixtures%2Fcore-image%2Fsrc%2Fassets%2Fpenguin1.jpg%3ForigWidth%3D207%26origHeight%3D243%26origFormat%3Djpg&f=jpg') 
// undefined
```

I'm not sure how common an undefined `image.options.format` is? Presumably not very since it wasn't handled in the previous implementation?

kitschpatrol: Pushed a commit with this, just noting the above caveat. It doesn't materially change the undefined edge case.

Princesseuh (approved):
This looks great, thank you very much! As I said, we'll do a proper overhaul of this in the future, one day, maybe. But this works perfectly fine in the meantime!

kitschpatrol:
Thanks!

Issue #11149, opened 2024-05-26
Updating to svelte@5.0.0-next.141 throws errors


amit13k:
### Astro Info

```block
Astro                    v4.9.1
Node                     v18.20.3
System                   Linux (x64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/svelte
```

### Describe the Bug

Updating to svelte@5.0.0-next.141 throws the following error,
```ts
✘ [ERROR] No matching export in "node_modules/svelte/src/internal/client/index.js" for import "add_snippet_symbol"

    node_modules/@astrojs/svelte/client-v5.js:2:9:
      2 │ import { add_snippet_symbol } from 'svelte/internal/client';
```

### What's the expected result?

No errors

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-w4fbdg?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

und3fined:
From next.134 can't access to `svelte/internal/client` (https://github.com/sveltejs/svelte/releases/tag/svelte%405.0.0-next.134)

Currently, i downgrade to next.133 and wait Astro fix it

mwc:
Svelte checks whether the import path starts with 'svelte/internal', so it is possible to import from the file path to bypass error checking; or directly 'fake' the 'add_snippet_symbol' function, which merely adds a 'snippet' identifier to the function, but neither approach seems to be a long-term solution.

mateothegreat:
Downgrading `@astrojs/svelte` to `4.0.4` and changing to `static` for `output: "static"` will get you up and running for now.

package.json:

```json
{
  "name": "blog",
  "type": "module",
  "version": "0.0.1",
  "scripts": {
    "dev": "astro check --watch & astro dev",
    "start": "astro dev",
    "build": "astro check && astro build",
    "preview": "astro preview",
    "astro": "astro"
  },
  "dependencies": {
    "@astrojs/check": "^0.7.0",
    "@astrojs/mdx": "^3.0.1",
    "@astrojs/rss": "^4.0.6",
    "@astrojs/sitemap": "^3.1.5",
    "@astrojs/svelte": "^4.0.4",
    "@astrojs/tailwind": "^5.1.0",
    "@astrojs/vercel": "^7.6.0",
    "astro": "^4.9.2",
    "prettier-plugin-astro": "^0.14.0",
    "prettier-plugin-svelte": "^3.2.3",
[...]

PR #11161, opened 2024-05-28
fix(svelte): update add_snippet_symbol import

mass8326:
## Changes

- What does this change?

Fixes https://github.com/withastro/astro/issues/11149

[`svelte@5.0.0-next-134`](https://github.com/sveltejs/svelte/pull/10800) removed the `add_snippet_symbol` export in `svelte/internal/client`. This changes the import from  `svelte/internal/client` to `svelte/internal/server`.

 ---

Looking at the svelte source code, `add_snippet_symbol` from both [client](https://github.com/sveltejs/svelte/blob/4c0b723c177063e94ed6beecc7e6a9b5ef806749/packages/svelte/src/internal/client/index.js#L140) and [server](https://github.com/sveltejs/svelte/blob/4c0b723c177063e94ed6beecc7e6a9b5ef806749/packages/svelte/src/internal/server/index.js#L661) were simply re-exports of a [shared file](https://github.com/sveltejs/svelte/blob/4c0b723c177063e94ed6beecc7e6a9b5ef806749/packages/svelte/src/internal/shared/validate.js#L8-L11).

Though it feels wrong to import a server file into a client file, this might be acceptable since this is for dev validation only and the underlying function is the same.

The re-export from `svelte/internal/server` was originally added in [`svelte@5.0.0-next.88`](https://github.com/sveltejs/svelte/pull/10972). Since the minimum version required is already `svelte@5.0.0-next.90`, this should not be a breaking change.

## Testing

Changing the import from `svelte/internal/client` to `svelte/internal/server` using `pnpm patch` in pre-existing projects allows build to complete. 

```diff
diff --git a/client-v5.js b/client-v5.js
index 755a6aa5381ee602ba8ece2f91f6439c8b19dd3f..896be0982ee3047cdc182345bad818733e06be92 100644
--- a/client-v5.js
+++ b/client-v5.js
@@ -1,5 +1,5 @@
 import { hydrate, mount, unmount } from 'svelte';
-import { add_snippet_symbol } from 'svelte/internal/client';
+import { add_snippet_symbol } from 'svelte/internal/server';
 
 // Allow a slot to be rendered as a snippet (dev validation only)
 const tagSlotAsSnippet = import.meta.env.DEV ? add_snippet_symbol : (s) => s;
```

## Docs

This fix requires no changes on the part of the user and shouldn't need to be documented.

### DIFF
--- a/.changeset/lovely-kids-knock.md
+++ b/.changeset/lovely-kids-knock.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/svelte": patch
+---
+
+Support internal breaking changes in svelte@5.0.0-next.134
--- a/packages/integrations/svelte/client-v5.js
+++ b/packages/integrations/svelte/client-v5.js
@@ -1,5 +1,5 @@
 import { hydrate, mount, unmount } from 'svelte';
-import { add_snippet_symbol } from 'svelte/internal/client';
+import { add_snippet_symbol } from 'svelte/internal/server';
 
 // Allow a slot to be rendered as a snippet (dev validation only)
 const tagSlotAsSnippet = import.meta.env.DEV ? add_snippet_symbol : (s) => s;


bluwy (commented):
I don't think we can import from `svelte/internal/server`. It implies that it'll only work on the server side. From what I can tell, Svelte refactored it to `wrap_snippet` in https://github.com/sveltejs/svelte/pull/11401, which caused `add_snippet_symbol` to be unused and subsequently removed in https://github.com/sveltejs/svelte/pull/10800. 

I think we need to use `wrap_snippet` in this case, but I'm not sure if there's any downsides to this. Otherwise it may be easier to remove calling the function in the first place if it works.

mass8326:
I took a look into this and tried replacing the call with `wrap_snippet` and it seems like the integration silently fails.

```svelte
<script lang="ts">
  import type { Snippet } from "svelte";
  let { children }: { children: Snippet } = $props();
</script>

{@render children()}
```

```astro
---
import Test from "test.svelte";
---

<Test client:load>Hello world!</Test>
```

The above code will end up displaying nothing without throwing any errors.

It seems like the [component context handling](https://github.com/sveltejs/svelte/blob/b13322cc345cb913ec0587f9e0b0dc08e1365c98/packages/svelte/src/internal/client/dom/blocks/snippet.js#L44-L53) added to `wrap_snippet` won't work with what Astro passes in.

Should an issue be made with the svelte repo asking for `add_snippet_symbol` to be exposed from 'svelte/internal/client' again?

mwc:
Is there any progress?

Issue #11160, opened 2024-05-28
lit client-shim uses deprecated apis, throws page speed error


scottnath:
### Astro Info

```block
Astro                    v4.9.2
Node                     v20.12.2
System                   macOS (x64)
Package Manager          npm
Output                   static
Adapter                  none
Integrations             @astrojs/lit
                         @astrojs/mdx
                         @astrojs/react
```

### If this issue only occurs in one browser, which browser is a problem?

chrome + pagespeed

### Describe the Bug

The lit plugin's client-shim is using a deprecated API, which is causing a big hit to page insights.

Chrome's deprecation info:
https://chromestatus.com/feature/5116094370283520

The fix:

```
Instead of:
  (new DOMParser()).parseFromString(html,'text/html',{includeShadowRoots: true});

this can be used instead:
  document.parseHTMLUnsafe(html);
```

The file: https://github.com/withastro/astro/blob/2d93902f4c51dcc62b077b0546ead688e6f32c63/packages/integrations/lit/client-shim.js#L11

The warning:

```
[Deprecation] The includeShadowRoots parameter to parseFromString has been deprecated, and will be removed very soon. Please use parseHTMLUnsafe() or setHTMLUnsafe() instead.
```

Example of page speed error:
https://pagespeed.web.dev/analysis/https-scottnath-com/d3d6jmz94w?form_factor=desktop

The difference:
If I remove the `@astrojs/lit` plugin from `astro.config` the warning goes away and Lighthouse/Best Practices goes from `81` to `100`

### What's the expected result?

Proposed fix:
Change client-shim to use the correct api

### Link to Minimal Reproducible Example

https://stackblitz.com/github/withastro/astro/tree/latest/examples/framework-lit?file=astro.config.mjs

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
Would you be willing to submit a PR?

scottnath:
hi @matthewp submitted here:
https://github.com/withastro/astro/pull/11164

Not sure what the rules are to fix it coming from a fork though?

cheers, scott

matthewp:
No issue at all, thanks!

PR #11164, opened 2024-05-29
:bug: removed deprecated items from lit integration

scottnath:
### Deprecated API
- Removes a deprecated API in lit plugin's client-shim which is causing a big hit to page insights.

Chrome's deprecation info:
https://chromestatus.com/feature/5116094370283520

Related story:
https://github.com/withastro/astro/issues/11160

### Deprecated Attribute

- Removes the `shadowroot` attribute which was deprecated by chrome and not needed by other browsers

Deprecation info:
https://chromestatus.com/feature/6239658726391808

## Testing

The original file did not include tests so I installed `client-shim` and client-shim-min` locally and tested them with the latest astro install I have, which has `lit` components in it

## Docs

Should not change the user experience. Worked for me with both local dev server and build/preview

### DIFF
--- a/packages/integrations/lit/client-shim.js
+++ b/packages/integrations/lit/client-shim.js
@@ -7,7 +7,7 @@ async function polyfill() {
 	});
 }
 
-const polyfillCheckEl = document
+const polyfillCheckEl = Document
 	.parseHTMLUnsafe(`<p><template shadowrootmode="open"></template></p>`)
 	.querySelector('p');
 
--- a/packages/integrations/lit/client-shim.js
+++ b/packages/integrations/lit/client-shim.js
@@ -7,14 +7,8 @@ async function polyfill() {
 	});
 }
 
-const polyfillCheckEl = new DOMParser()
-	.parseFromString(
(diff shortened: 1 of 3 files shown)


matthewp (approved):
Thank you!

packages/integrations/lit/client-shim.js
```diff
@@ -7,14 +7,8 @@ async function polyfill() {
...
-const polyfillCheckEl = new DOMParser()
-	.parseFromString(
-		`<p><template shadowroot="open" shadowrootmode="open"></template></p>`,
-		'text/html',
-		{
-			includeShadowRoots: true,
-		}
-	)
+const polyfillCheckEl = Document
+	.parseHTMLUnsafe(`<p><template shadowrootmode="open"></template></p>`)
```
jrencz, line 11: Hi,

May I ask about the comparability? 

`parseFromString` is there in basically every browser (https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString) while `parseHTMLUnsafe` is not yet we'll supported (https://caniuse.com/mdn-api_document_parsehtmlunsafe_static today reports global of just under 58%)

Wont that be a problem?

PR #11173, opened 2024-05-31
chore: enable @typescript-eslint/no-unused-vars internally

JoshuaKGoldberg:
## Changes

> _(please forgive me for wiping the template for this internal-only chore)_

Upgrades `@typescript-eslint/no-unused-vars` from `'warn'` to `'error'` in the ESLint config, then resolves the roughly 20 complaints that were hanging around from it. It's a nice amount of deletions, I think. 🙂 

I don't believe any of the changed types impact users. But just to be safe, added comments inline.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -62,7 +62,7 @@ export default [
 			// These off/configured-differently-by-default rules fit well for us
 			'@typescript-eslint/switch-exhaustiveness-check': 'error',
 			'@typescript-eslint/no-unused-vars': [
-				'warn',
+				'error',
 				{
 					argsIgnorePattern: '^_',
 					varsIgnorePattern: '^_',
--- a/packages/astro/e2e/view-transitions.test.js
+++ b/packages/astro/e2e/view-transitions.test.js
@@ -168,8 +168,6 @@ test.describe('View Transitions', () => {
 	});
 
 	test('Moving from a page without ViewTransitions w/ back button', async ({ page, astro }) => {
-		const loads = collectLoads(page);
-
 		// Go to page 1
 		await page.goto(astro.resolveUrl('/one'));
 		let p = page.locator('#one');
--- a/packages/astro/src/actions/runtime/virtual/shared.ts
+++ b/packages/astro/src/actions/runtime/virtual/shared.ts
@@ -42,7 +42,7 @@ const statusToCodeMap: Record<number, ActionErrorCode> = Object.entries(codeToSt
 
 export type ErrorInferenceObject = Record<string, any>;
 
-export class ActionError<T extends ErrorInferenceObject = ErrorInferenceObject> extends Error {
+export class ActionError extends Error {
 	type = 'AstroActionError';
 	code: ActionErrorCode = 'INTERNAL_SERVER_ERROR';
 	status = 500;
@@ -84,19 +84,19 @@ export class ActionError<T extends ErrorInferenceObject = ErrorInferenceObject> 
 }
 
 export function isInputError<T extends ErrorInferenceObject>(
-	error?: ActionError<T>
+	error?: ActionError
 ): error is ActionInputError<T> {
 	return error instanceof ActionInputError;
 }
 
-export type SafeResult<TInput extends ErrorInferenceObject, TOutput> =
+export type SafeResult<TOutput> =
 	| {
 			data: TOutput;
 			error: undefined;
 	  }
 	| {
 			data: undefined;
-			error: ActionError<TInput>;
+			error: ActionError;
 	  };
 
 export class ActionInputError<T extends ErrorInferenceObject> extends ActionError {
@@ -128,7 +128,7 @@ export class ActionInputError<T extends ErrorInferenceObject> extends ActionErro
 
 export async function callSafely<TOutput>(
 	handler: () => MaybePromise<TOutput>
-): Promise<SafeResult<z.ZodType, TOutput>> {
+): Promise<SafeResult<TOutput>> {
 	try {
 		const data = await handler();
 		return { data, error: undefined };
--- a/packages/astro/src/content/vite-plugin-content-assets.ts
+++ b/packages/astro/src/content/vite-plugin-content-assets.ts
@@ -1,6 +1,6 @@
 import { extname } from 'node:path';
 import { pathToFileURL } from 'node:url';
-import type { Plugin, Rollup } from 'vite';
+import type { Plugin } from 'vite';
 import type { AstroSettings, SSRElement } from '../@types/astro.js';
 import { getAssetsPrefix } from '../assets/utils/getAssetsPrefix.js';
 import type { BuildInternals } from '../core/build/internal.js';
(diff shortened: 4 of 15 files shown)


packages/astro/src/actions/runtime/virtual/shared.ts
```diff
@@ -42,7 +42,7 @@ const statusToCodeMap: Record<number, ActionErrorCode> = Object.entries(codeToSt
 
 export type ErrorInferenceObject = Record<string, any>;
 
-export class ActionError<T extends ErrorInferenceObject = ErrorInferenceObject> extends Error {
+export class ActionError extends Error {
```
JoshuaKGoldberg, line 45: This `T` was unused 🔪 . And, as a result, `SafeResult`'s `TInput` later on becomes unused as well. 🔪

packages/astro/src/content/vite-plugin-content-assets.ts
```diff
@@ -129,22 +129,9 @@ export function astroConfigBuildPlugin(
...
-					vitePlugin: {
-						name: 'astro:content-build-plugin',
-						generateBundle() {
-							if (target === 'server') {
-								ssrPluginContext = this;
-							}
-						},
-					},
-				};
-			},
```
JoshuaKGoldberg, line 147: All this `'build:before'` hook as doing was assigning to the `ssrPluginContext` variable... which was then never used. So I believe removing it altogether is safe, right?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/rules/a11y.ts
```diff
@@ -125,12 +125,6 @@ const a11y_required_content = [
 
 const a11y_distracting_elements = ['blink', 'marquee'];
 
-// Unused for now
```
JoshuaKGoldberg, line 128: > for now

These are almost half a year old. May I suggest removing them? 🙂

packages/astro/e2e/view-transitions.test.js
```diff
@@ -168,8 +168,6 @@ test.describe('View Transitions', () => {
 	});
 
 	test('Moving from a page without ViewTransitions w/ back button', async ({ page, astro }) => {
-		const loads = collectLoads(page);
```
florian-lefebvre, line 171: @martrapp do you know if this was required for the test (maybe doing a side-effect?)

packages/astro/src/actions/runtime/virtual/shared.ts
florian-lefebvre: lgtm, @bholmesdev can you confirm?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/rules/a11y.ts
florian-lefebvre: @Princesseuh do we want to keep them as comments for future usage or something?

packages/integrations/sitemap/src/utils/is-valid-url.ts
```diff
@@ -4,8 +4,7 @@ export const isValidUrl = (s: any) => {
 		return false;
 	}
 	try {
-		// eslint-disable-next-line @typescript-eslint/no-unused-vars
-		const dummy = new URL(s);
+		new URL(s);
```
florian-lefebvre, line 7: Could we even use `URL.canParse`?

packages/astro/src/runtime/client/dev-toolbar/apps/audit/rules/a11y.ts
Princesseuh: Nah, it's fine to remove.

eslint.config.js
```diff
@@ -62,7 +62,7 @@ export default [
 			// These off/configured-differently-by-default rules fit well for us
 			'@typescript-eslint/switch-exhaustiveness-check': 'error',
 			'@typescript-eslint/no-unused-vars': [
-				'warn',
+				'error',
```
Princesseuh, line 65: We've historically been against this as some of us didn't like the experience of having errors while working, so I guess we should probably discuss it a bit

packages/astro/e2e/view-transitions.test.js
martrapp: Hi @florian-lefebvre, thank you for asking! The loads value was indeed unused. Now it has a purpose ;-)

eslint.config.js
bluwy: Given what the PR has caught if we had a stricter lint, I think this is a nice change to me. It makes the code more readable too, which improves the experience for others working on the code.

bluwy (approved):

bluwy:
We've discussed this among the platform team and think this is a good change. Would need the conflicts resolved first before merging.

packages/astro/src/actions/runtime/virtual/shared.ts
bholmesdev: I've documented the reason for this generic type [here](https://github.com/withastro/astro/blob/main/packages/astro/src/actions/runtime/utils.ts#L37-L46). The generic is used by `ActionInputError` (a descendant of `ActionError`). However, I've found that if you remove this generic from `ActionError`, all type inference breaks for our `isInputError()` utility. If you have a suggestion to avoid this, please let me know @JoshuaKGoldberg. We now have a [type test](https://github.com/withastro/astro/blob/main/packages/astro/test/types/is-input-error.ts) to assert whether removing this generic causes any issue

bholmesdev (changes requested):
I've responded to the comment above on the unused generic for the `ActionError` object. I would vote to preserve this generic with a clear comment on why it is present. If we can find a TypeScript refactor for `isInputError()` to preserves the generic, that would be best.

packages/astro/src/actions/runtime/virtual/shared.ts
JoshuaKGoldberg: Aha! Thanks, that's helpful.

The root problem here is that your `isInputError` type is not type-safe. It violates the ["Golden Rule of Generics"](https://effectivetypescript.com/2020/08/12/generics-golden-rule). Nothing in the body of `isInputError` actually uses the type parameter `T`. You could call it like...

```ts
const example = new ActionError({ code: "BAD_REQUEST" });

if (isInputError<{ gotcha: "whoops" }>(example) {
	example.fields.gotcha;
	//             ^? string[] | undefined

}
```

...and TypeScript wouldn't know to complain.

> Aside: which we recently codified that golden rule in typescript-eslint as [`@typescript-eslint/no-unnecessary-type-parameters`](https://typescript-eslint.io/rules/no-unnecessary-type-parameters)... but the lint rule doesn't (yet?) have the ability to check type parameters passed as type arguments - see its [Limitations](https://typescript-eslint.io/rules/no-unnecessary-type-parameters/#limitations)). Pity. This wouldn't be caught by linting right now.

I don't have enough context to refactor things to not violate the golden rule. Sometimes the right approach actually is to just `eslint-disable-next-line` the unused thing and/or only-used-once type parameter. So I'll revert these removals for now with a comment.

florian-lefebvre (approved):

bholmesdev (approved):

PR #11176, opened 2024-06-02
feat: add first and last to the Page interface

tsawada:
## Changes

Add `page.url.first`, `page.url.last` to the Page interface.
Closes https://github.com/withastro/roadmap/discussions/862

## Testing
I couldn't find relevant tests for `page.url`. Let me know if there is one already.

## Docs
This needs to be updated
https://github.com/withastro/docs/blob/main/src/content/docs/en/guides/routing.mdx#complete-api-reference

### DIFF
--- a/.changeset/twenty-maps-glow.md
+++ b/.changeset/twenty-maps-glow.md
@@ -1,0 +1,5 @@
+---
+'astro': major
+---
+
+feat: Add urls to the first and the last page in the Page interface
--- a/.changeset/twenty-maps-glow.md
+++ b/.changeset/twenty-maps-glow.md
@@ -2,4 +2,4 @@
 'astro': minor
 ---
 
-feat: Add urls to the first and the last page in the Page interface
+Adds two new values to the [pagination `page` prop](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop): `page.first` and `page.last` for accessing the URLs of the first and last pages.
--- a/.changeset/twenty-maps-glow.md
+++ b/.changeset/twenty-maps-glow.md
@@ -1,5 +1,5 @@
 ---
-'astro': major
+'astro': minor
 ---
 
 feat: Add urls to the first and the last page in the Page interface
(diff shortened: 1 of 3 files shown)


.changeset/twenty-maps-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
```
bluwy, line 2:
```suggestion
'astro': minor
```
It's a minor as it introduces a feature but doesn't break anything.

bluwy (approved):

bluwy:
Looks like we forgot to followup on this PR. I queued this up for the next minor in 4.12.0

.changeset/twenty-maps-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+feat: Add urls to the first and the last page in the Page interface
```
sarah11918, line 5: This will need updating, will wait to see what the docs PR is to know what to suggest.

sarah11918 (changes requested):
Just placeholding for docs here! When I get a docs PR, then I'll be able to review!

matthewp:
Will need docs for this. @tsawada will you be able to write them?

sarah11918:
Added a docs PR with my suggestion here, since this appears to be similar to other existing items. Please read carefully for accuracy, as I inferred e.g. when the value is undefined based on the previous and next entries!

https://github.com/withastro/docs/pull/8751

.changeset/twenty-maps-glow.md
sarah11918, line 5:
```suggestion
Adds two new values to the [pagination `page` prop](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop): `page.first` and `page.last` for accessing the URLs of the first and last pages.
```
Maybe something like this? When new options are added, I do always think it's helpful if the changeset message mentions them by name!

sarah11918 (commented):
Thank you for this very handy addition to the `page` prop, @tsawada ! I think people will make very good use of this!  :raised_hands: 

I've gone ahead and drafted a Docs PR to accompany this, but I will need you to review and make sure it's accurate, as well as my suggestion here for the message that will show up in the CHANGELOG.

So, docs aren't ready to go but I've tried to move them along with a suggestion. Once you've approved or corrected them, then Docs is happy!

sarah11918 (approved):
Approving for docs!

PR #11178, opened 2024-06-03
fix: better isPromise check for proxy objects

theoephraim:
## Changes

Adjusts the `isPromise` utility to first check for the presence of the `'then'` key before accessing it to check the type.

The absence of this check on a Proxy object was causing an error to be thrown  - as my Proxy object throws on access of an unknown keys.

## Testing

This should be a totally transparent change, and not quite worth setting up a test IMO.
Note that a few tests were failing for me locally, but they are failing on main as well, so seem to be unrelated.

## Docs

N/A - This should hopefully be a totally transparent change that affects nothing other than this very specific error scenario.

### DIFF
--- a/.changeset/green-moose-fetch.md
+++ b/.changeset/green-moose-fetch.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Improves `isPromise` utility to check the presence of `then` on an object before trying to access it - which can cause undesired side-effects on Proxy objects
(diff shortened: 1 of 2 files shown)


bluwy (approved):

.changeset/green-moose-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+fix: make isPromise work on proxy objects
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where a `Promise` on a `Proxy` would not be handled correctly
```
I'm not sure about the wording "on a" mainly because idk much about proxies. Any suggestions?

florian-lefebvre (approved):

Issue #11127, opened 2024-05-23
Experimental rewriting of POST request does not work


gersomvg:
### Astro Info

```block
Astro                    v4.8.4
Node                     v20.12.2
System                   macOS (arm64)
Package Manager          npm (this should be pnpm)
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/tailwind
```

### If this issue only occurs in one browser, which browser is a problem?

Not specific to a browser

### Describe the Bug

Just trying out the new rewriting feature as this came just in time for a use case that I had. Question: is this supposed to work with a POST request?

Let's say you post an `email` field to `/page-a`, I want to rewrite to `/page-b` while keeping the FormData.

It currently does not work, as the rewrite results in a 404. Any ideas?

### What's the expected result?

I'd expect `/page-b` to render and receive the `email` field in FormData

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-oyssmw?file=src%2Fpages%2Fpage-a.astro

ematipico:
Downgraded to P3, as it's not critical and it doesn't affect many users. Plus, it's experimental and bugs are expected.

PR #11182, opened 2024-06-04
fix(rewrite): copy body from the request

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11127

I created a new utility function called `copyRequest` inside `core/routing/`, I thought it fits there.

## Testing

Added a new test case

## Docs

N/A


packages/astro/src/core/routing/request.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param newUrl The new `URL`
+ * @param oldRequest The old `Request`
+ */
+export async function copyRequest(newUrl: URL, oldRequest: Request): Promise<Request> {
+	const body = oldRequest.headers.get('content-type') ? oldRequest.blob() : Promise.resolve(undefined);
+	return body.then((requestBody) => {
+		return new Request(newUrl, {
+			method: oldRequest.method,
+			headers: oldRequest.headers,
+			body: requestBody,
```
bluwy, line 13: I was wondering why we need to copy all these manually, and it seems like the spec never quite handled `body`. It also seems similar to https://stackoverflow.com/a/34641566 where there's a comment that we should only copy the body if it's not GET or HEAD instead? Or do you have a reason to base on content-type?

Also, I wonder if `body: oldRequest.body` would be enough, since the body is already a valid type for the input 🤔

matthewp: It would be if it has not been ready.

ematipico: The Stackoverflow comment was old and not up to date with the specs. I found a way to copy the body without incurring into errors: [`e6b975e` (#11182)](https://github.com/withastro/astro/pull/11182/commits/e6b975ef208d11bb054f04f0bc599b717100f222)

matthewp (approved):

PR #11184, opened 2024-06-04
Fix redirect logic to be able to handle dynamic parts in target paths

Joelkang:
## Context
There are a few related changes in this PR, but this was the use case I was trying to fix:
`[category]/page/[page].astro` is the route handler for `myCategory/page/1`.
 
I'm trying to redirect from `/[category]` to `/[category]/page/1`. Right now, since there's no RouteData with key `/[category]/page/1` in the manifest, my redirect target falls back to the server-rendered path, which in my case for cloudflare, is the [empty string ''](https://github.com/withastro/adapters/blob/be0b27de38418330898205d506fad87b7da563be/packages/cloudflare/src/index.ts#L378), causing the entry in `_redirect` to be invalid. The line that gets generated is `/:category   301`

There *is* however, a RouteData in the manifest that's able to handle this path, so it should resolve that route at the key `/[category]/page/[page]`.

## Changes
1. Adds validation for the target such that target params should also be in source params, as per the current docs
2. Resolve partially static redirect target paths to their route handler during manifest creation
3. Fix `underscore-redirect` to generate the right _redirect for targets with dynamic parts

This should allow [this comment](https://github.com/withastro/adapters/blob/be0b27de38418330898205d506fad87b7da563be/packages/netlify/src/index.ts#L198) and hack to be removed 

## Questions
Are there edge cases that I'm missing? Its not clear to me if this is actually a backwards compatible change. It feels like my changes align with what the code is *supposed* to do, but this might break some folks?

## Testing

Added test for each of the 3 cases above that did not exist before.

## Docs

/cc @withastro/maintainers-docs Not sure if we need to add more docs around what is permissible in the redirect config, but this really threw me off guard that I couldn't do what I thought I was able to do.

### DIFF
--- a/packages/astro/src/core/routing/manifest/create.ts
+++ b/packages/astro/src/core/routing/manifest/create.ts
@@ -140,6 +140,16 @@ export function validateSegment(segment: string, file = '') {
 	}
 }
 
+export function getSegmentsFromRoutePath(routePath: string, file = routePath) {
+	return removeLeadingForwardSlash(routePath)
+		.split(path.posix.sep)
+		.filter(Boolean)
+		.map((s: string) => {
+			validateSegment(s);
+			return getParts(s, file);
+		});
+}
+
 /**
  * Checks whether two route segments are semantically equivalent.
  *
@@ -350,13 +360,7 @@ function createInjectedRoutes({ settings, cwd }: CreateRouteManifestParams): Pri
 		}
 		const component = slash(path.relative(cwd || fileURLToPath(config.root), resolved));
 
-		const segments = removeLeadingForwardSlash(name)
-			.split(path.posix.sep)
-			.filter(Boolean)
-			.map((s: string) => {
-				validateSegment(s);
-				return getParts(s, component);
-			});
+		const segments = getSegmentsFromRoutePath(name, component);
 
 		const type = resolved.endsWith('.astro') ? 'page' : 'endpoint';
 		const isPage = type === 'page';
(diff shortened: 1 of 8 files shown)


packages/underscore-redirects/test/astro.test.js
```diff
@@ -29,4 +29,282 @@ describe('Astro', () => {
 
 		assert.equal(_redirects.definitions.length, 2);
 	});
+
+	describe('Supports dynamic routes', () => {
```
Joelkang, line 33: There's a lot of branching in this file that isn't covered by these tests, but I'm going to leave that for later

packages/underscore-redirects/src/astro.ts
```diff
@@ -125,15 +138,15 @@ export function createRedirectsFromAstroRoutes({
...
+function generateDynamicPattern(route: RouteData, forTarget = false) {
 	const pattern =
 		'/' +
 		route.segments
 			.map(([part]) => {
 				//(part.dynamic ? '*' : part.content)
 				if (part.dynamic) {
 					if (part.spread) {
-						return '*';
+						return forTarget ? ':splat': '*';
```
Joelkang, line 149: As per:
https://developers.cloudflare.com/pages/configuration/redirects/#splats and https://docs.netlify.com/routing/redirects/redirect-options/#splats

packages/astro/src/core/routing/manifest/create.ts
```diff
@@ -443,6 +462,18 @@ function createRedirectRoutes(
...
+		validateRedirectPaths(from, destination);
+
+		const redirectRoute =
+			routeMap.get(destination) ??
+			[...routeMap.values()]
+				.sort(routeComparator)
+				.find(
+					(route) =>
+						route.pattern.test(destination) ||
+						route.fallbackRoutes.some((fallbackRoute) => fallbackRoute.pattern.test(destination))
```
Joelkang, line 474: This is basically the same as matchRoute, but the current matchRoute code in match.js is for handling requests one level higher, and thus requires the entire Manifest

Princesseuh:
Are you still interested in working on this?

florian-lefebvre:
Looks like there's a bunch of conflict so it may be worth starting over to be honest, I'm sorry about that! I don't think it's a bug but rather a new feature. It doesn't need to go through the RFC process because it's fairly small tho. Thanks for trying and feel free to re-open a PR

Issue #11100, opened 2024-05-19
i18n Issue: Astro encountered an unexpected line of code.


bapspatil:
### Astro Info

```block
Astro                    v4.8.6
Node                     v18.19.0
System                   macOS (arm64)
Package Manager          unknown
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/react
                         @astrojs/tailwind
                         @astrojs/mdx
                         @astrojs/partytown
                         @astrojs/sitemap
                         @swup/astro
```

### Describe the Bug

I tried getting the locale like this:
```const lang = getLocaleByPath(Astro.url.pathname);```

But this was the error I got
PS: I am using Bun package manager, but running Astro in NodeJS environment via Bun's NodeJS compat.

### What's the expected result?

`lang` should contain the locale. "en" or "fr" in this case.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-hgf9mt?file=src%2Fpages%2Findex.astro

### Participation

- [x] I am willing to submit a pull request for this issue.

matthewp:
@ematipico I think this is an inaccurate error message. There is no locale in the pathname, but we are throwing an unreachable error. https://github.com/withastro/astro/blob/464b0a108ecd0b77e6e34829381ccd1335d525ce/packages/astro/src/i18n/index.ts#L215

I'm not sure if we should throw when not finding a locale or not, but if we do it should be a different error message.

PR #11189, opened 2024-06-05
fix(i18n): improve error message

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11100

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -1205,7 +1205,7 @@ export const i18nNotEnabled = {
 export const i18nNoLocaleFoundInPath = {
 	name: 'i18nNoLocaleFoundInPath',
 	title: 'The path doesn\'t contain any locale',
-	message: 'You tried to use an utility on a path that doesn\'t contain any locale. You should use `pathHasLocale` to determine, first, if the path has a locale.',
+	message: 'You tried to use an i18n utility on a path that doesn\'t contain any locale. You can use \`pathHasLocale\` first to determine if the path has a locale.',
 } satisfies  ErrorData;
 
 /**
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -1199,6 +1199,18 @@ export const i18nNotEnabled = {
 /**
  * @docs
  * @description
+ * The path doesn't contain any locale. The user should use `pathHasLocale` first.
+ * 
+ */
+export const i18nNoLocaleFoundInPath = {
+	name: 'i18nNoLocaleFoundInPath',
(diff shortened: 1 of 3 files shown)


packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1196,6 +1196,18 @@ export const i18nNotEnabled = {
 	hint: 'See https://docs.astro.build/en/guides/internationalization for a guide on setting up i18n.',
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * The path doesn't contain any locale. The user should use `pathHasLocale` first.
```
sarah11918, line 1202:
```suggestion
 * An i18n utility requires a `locale` property, but the path doesn't contain one. Use `pathHasLocale` first to check for a `locale` before using i18n utilities.
```

sarah11918 (commented):
Some tiny suggestions!

packages/astro/src/core/errors/errors-data.ts
ematipico: The suggestion is misleading. The function accepts a `path` and the path must contain a locale .e.g `/en/slug`. The user doesn't provide locale directly

bluwy (approved):

packages/astro/src/core/errors/errors-data.ts
sarah11918: I think we're saying the same thing? I'm just trying to better understand. I'm not suggesting that a user provide a locale directly, but (if I understand this correctly) they should add some logic with `pathHasLocal` to perform a check for whether or not a path in included before running logic that takes `path` and assumes it has a locale?

ematipico: > An i18n utility requires a `locale` property

This is the part that is confusing to me. There's no property involved 😅 I will make an example, maybe it will help

```js
`getLocaleFromPath("/slug")` // error, because there's not locale in the path, this error is thrown
`getLocaleFromPath("/en/slug")` // correct, no error, the user gets the correct value
```

sarah11918: Ah, I think I get why my version sounded confusing! I was jumping ahead to explain the later problem that someone would have, that an i18n utility wasn't going to work. (from the `message`)

I was trying to provide a bit more content about the situation in which this problem of missing locale would come up, but maybe that's too far a jump. This content shows up under the heading "What went wrong?" in docs, so I was trying to make this explain a little more.

Here's another attempt, and if I still fail, then we keep the original until someone complains it's too short and doesn't explain enough!  :sweat_smile: 

> An i18n utility tried to use the locale from a URL path that does not contain one. You can prevent this error by using `pathHasLocale` to check URLs for a `locale` first before using i18n utilities.

PR #11195, opened 2024-06-06
feat(astro): improved validators

florian-lefebvre:
## Changes

- Adds support for enums
- Add more validation rules
- Simplifies `envField` types

## Testing

Added unit tests

## Docs

N/A


matthewp:
Can you update the RFC with these types?

.changeset/red-rings-tell.md
```diff
@@ -0,0 +1,38 @@
...
+            schema: {
+                FOO: envField.string({
+                    // ...
+                    max: 32,
+                    min: 3,
+                    length: 12,
+                    url: true,
+                    includes: 'foo',
+                    startsWith: 'bar',
+                    endsWith: 'baz'
```
matthewp, line 24: I guess the reason we can't have a `check(value: string) => boolean` sort of thing is because this needs to happen at runtime, right?

.changeset/red-rings-tell.md
```diff
@@ -0,0 +1,38 @@
...
+export default defineConfig({
+    experimental: {
+        env: {
+            schema: {
+                FOO: envField.string({
+                    // ...
+                    max: 32,
+                    min: 3,
+                    length: 12,
+                    url: true,
```
matthewp, line 21: What does this do, does it turn it into a `URL` object? If so I feel like a new type of URL, `envField.url()` might be better.

florian-lefebvre: It's the equivalent of `z.string().url()`, so it only returns a string that is a valid url. Happy to add `envField.url` if some users request it

.changeset/red-rings-tell.md
florian-lefebvre: Yeah exactly, everything has to be serializable

florian-lefebvre:
I updated the RFC

ematipico (approved):

packages/astro/src/env/schema.ts
```diff
@@ -66,61 +108,32 @@ export const EnvSchema = z
...
-		SecretServerEnvFieldMetadata,
-	]),
-	BooleanSchema.omit({ type: true })
-);
-export type BooleanFieldInput = z.infer<typeof BooleanFieldInput>;
+export type EnumField = _Field<typeof EnumSchema>;
+export type EnumFieldInput<T extends string> = Prettify<
+	_FieldInput<typeof EnumSchema, 'type' | 'values' | 'default'> & {
+		values: Array<T>;
+		default?: NoInfer<T> | undefined;
```
florian-lefebvre, line 137: This has been released in TS 5.4, could that be an issue?

ematipico: No, it should not

ematipico:
Can we merge this one @florian-lefebvre ?

florian-lefebvre:
Yes, I was waiting in case Matthew wanted to follow up on the reviews

PR #11199, opened 2024-06-06
feat: add support for astro:env to node and vercel

florian-lefebvre:
## Changes

- Depends on #11206
- Adds support for `astro:env`

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.changeset/wet-chefs-develop.md
+++ b/.changeset/wet-chefs-develop.md
@@ -1,6 +1,6 @@
 ---
-'@astrojs/vercel': patch
-'@astrojs/node': patch
+'@astrojs/vercel': minor
+'@astrojs/node': minor
 ---
 
 Adds support for experimental `astro:env` released in Astro 4.10
--- a/.changeset/wet-chefs-develop.md
+++ b/.changeset/wet-chefs-develop.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/vercel': patch
+'@astrojs/node': patch
+---
+
+Adds support for experimental `astro:env` released in Astro 4.10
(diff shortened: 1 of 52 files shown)


packages/integrations/vercel/src/static/adapter.ts
```diff
@@ -32,6 +32,7 @@ function getAdapter(): AstroAdapter {
 			staticOutput: 'stable',
 			serverOutput: 'unsupported',
 			hybridOutput: 'unsupported',
+			envGetSecret: 'unsupported',
```
florian-lefebvre, line 35: I think it's fine since `getSecret` will never be called at runtime?

matthewp (approved):

.changeset/wet-chefs-develop.md
```diff
@@ -0,0 +1,6 @@
+---
+'@astrojs/vercel': patch
+'@astrojs/node': patch
+---
```
ematipico, line 4: Mmmmh, shouldn't this be a `minor`? We are adding a new feature :)

ematipico (approved):

.changeset/wet-chefs-develop.md
florian-lefebvre: IIRC patches are fine for experimental features in core but idk what should be done for other official packages relying on experimental features! Is that documented anywhere?

ematipico: I think It should be a minor because they both require a new feature exposed by Astro that was available in 4.10 (Astro features). Also, we released the experimental feature in a minor, not a patch, hence adapters should do the same. 

However, **changes in _existing_** experimental features are accepted in patches.

Hence, I'd say these changes should be shipped as a minor because we ship a new user facing feature.

PR #11202, opened 2024-06-06
fix(node): add missing error logging

### LINKED ISSUE #11200: No logs on Internal Server Error in node adapter
### Astro Info

```block
Astro                    v4.9.2
Package Manager          npm
Integrations             @astrojs/preact
                         @astrojs/mdx
```

_No response_

### Describe the Bug

#10285 introduced a try catch which swallows some of the errors making it very hard to know what was the root cause of 500. You should always either log or rethrow in a catch block.

All errors should be logged.

Any app with node adapter

```
npm create astro@latest
cd repro
npx astro add node
npm run build
node dist/server/entry.mjs

# in second terminal
curl localhost:4321 -H 'X-Forwarded-Host: example.com, example2.com' 
# no logs in first terminal
```

### Participation

dkobierski:
## Changes

Fixes #11200
Adds missing error logging in catch block.

## Testing

Testing not needed

## Docs

N/A

### DIFF
--- a/packages/integrations/node/src/serve-app.ts
+++ b/packages/integrations/node/src/serve-app.ts
@@ -25,6 +25,8 @@ export function createAppHandler(app: NodeApp): RequestHandler {
 		try {
 			request = NodeApp.createRequest(req);
 		} catch (err) {
+			logger.error(`Could not render ${req.url}`);
+			console.error(err);
 			res.statusCode = 500;
 			res.end('Internal Server Error');
 			return;
--- a/.changeset/nervous-worms-sort.md
+++ b/.changeset/nervous-worms-sort.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/node': patch
+---
+
+Fixes suppressed logs when error occurs


packages/integrations/node/src/serve-app.ts
```diff
@@ -25,6 +25,8 @@ export function createAppHandler(app: NodeApp): RequestHandler {
 		try {
 			request = NodeApp.createRequest(req);
 		} catch (err) {
+			logger.error(`Could not render ${req.url}`);
+			console.error(err);
```
ematipico, line 29: Why don't we use our logger to log the actual error?

dkobierski: I've just used the same approach as [here](https://github.com/withastro/astro/blob/368e4ced10b7a629c9ad94800ccf2a1243038921/packages/integrations/node/src/middleware.ts#L33).  I'm happy to change it.

ematipico (approved):

PR #11203, opened 2024-06-06
feat: env leak detection

florian-lefebvre:
## Changes

- A new `leakDetectionMiddleware` is exported from `astro/env/middleware`. If a server var ends up on the client, an error will be thrown. [Inspired by DMNO](https://github.com/dmno-dev/dmno/blob/main/packages/integrations/astro/src/astro-middleware.ts#L59)

## Testing

Added a test

## Docs

Changeset, I'll update the RFC after approval


.changeset/forty-crabs-judge.md
```diff
@@ -0,0 +1,30 @@
...
+
+Adds an optional middleware for usage with `astro:env`
+
+If you're using `astro:env`, you can now use a middleware to detect server envrionment variables leaks on the client:
+
+```ts
+import { leakDetectionMiddleware } from 'astro/env/middleware'
+
+export const onRequest = leakDetectionMiddleware()
+```
```
ematipico, line 13: nit: I think a better example would to be show how to enable it inside an existing middleware, and tell users that they should put this middleware in the back:
``````suggestion
```ts
// src/middleware.js

import { leakDetectionMiddleware } from 'astro/env/middleware'
import { sequence, defineMiddleware } form 'astro:middleware'

const userMiddleware = defineMiddleware((_, next) => {
	return next()
})

	// have the leak detection utility to run before the user middleware
	userMiddleware,
	leakDetectionMiddleware()
);
```
``````

File: .changeset/forty-crabs-judge.md [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+
+An error will be thrown instead of rendering the page if a leak is detected.
+
+You can pass 2 options:
+
+1. `filterContentType`: filters what response content type should trigger the check. Defaults to the content type starting with `/text` or `application/json`
```
ematipico, line 19: Do you mean `text/*`? I am not aware of content type headers that are named `/text`

packages/astro/src/@types/astro.ts
```diff
@@ -2160,7 +2160,7 @@ export interface AstroUserConfig {
 			 * })
 			 * ```
 			 */
-			schema?: EnvSchema;
```
ematipico, line 2163: Is it intentional? If so, you should update the docs too

ematipico (commented):
It took me a while to understand that this middleware needs to inspect the `Response`. I would highlight this important detail, because putting this middleware on the front could allow users to change the secrets before the check.

packages/astro/src/env/runtime/middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+ * environment variable.
+ */
+export function leakDetectionMiddleware({
+	filterContentType = (v) => v.startsWith('text/') || v.startsWith('application/json'),
+	excludeKeys = [],
+}: MiddewareOptions = {}): MiddlewareHandler {
+	return async (_, next) => {
+		const response = await next();
+
+		const contentType = response.headers.get('Content-Type');
```
ematipico, line 29: Oh, use `content-type`. Headers are not case sensitive

.changeset/forty-crabs-judge.md
```diff
@@ -0,0 +1,30 @@
+---
+'astro': patch
+---
+
+Adds an optional middleware for usage with `astro:env`
+
+If you're using `astro:env`, you can now use a middleware to detect server envrionment variables leaks on the client:
+
+```ts
+import { leakDetectionMiddleware } from 'astro/env/middleware'
```
bluwy, line 10: I wonder if we should make this `astro:env/middleware` instead?

florian-lefebvre: I think it makes things a bit more complicated. There's no codegen involved for the middleware code so I'm a bit reluctant to move it to a virtual import. We were also doing this for `astro:env/setup` and ended up moving it to `astro/env/setup`

matthewp:
Let's set up a call and discuss this feature. Although it's very cool, I worry that the maintenance burden might be too high and not sure if we should take it on. But open to discussing for sure. We can set up a maintainers call to go over it.

florian-lefebvre:
We need a proper RFC, this is a sensitive feature

Issue #11204, opened 2024-06-06
Astro.rewrite not working with SSR dynamic routes


Dovakeidy:
### Astro Info

```block
Astro                    v4.10.0
Node                     v20.11.0
System                   Docker / macOS (arm64)
Package Manager          yarn
Output                   hybrid
Adapter                  @astrojs/node
```

### Describe the Bug

**IMPORTANT NOTE: In the previous version (astro@4.9.2), the rewrite was sending a 404 Not found error ( without showing my custom 404.astro route, just a Not found error )** 

I'm using the experimental `rewrite()` method from an Astro but i get unexpected results.

My goal is to rewrite to an other SSR dynamic route defined like this:

<img width="398" alt="Capture d’écran 2024-06-06 à 23 22 16" src="https://github.com/withastro/astro/assets/26869114/c03f83fb-0e25-4178-89d0-8477ef2a7e25">

### Example 1 ( using Astro.params.slug in the desired page )

```
---
export const prerender = false;

return Astro.rewrite("/agences/angers"); // Should rewrite to the [subdomain]/index.astro
---
```
**Error:**
<img width="892" alt="Capture d’écran 2024-06-06 à 23 33 30" src="https://github.com/withastro/astro/assets/26869114/dcd9f28b-8b33-427c-92f7-d5de80dec65c">

### Example 2
```
---
export const prerender = false;

return Astro.rewrite("/agences/angers/page-1"); // Should rewrite to the [subdomain]/[slug].astro
---
```

**Error:**
<img width="1110" alt="Capture d’écran 2024-06-06 à 23 56 21" src="https://github.com/withastro/astro/assets/26869114/4ebfd4d2-397d-44fa-b72a-0fe9481d5758">

_As the above image show, i'm using the `export const prerender = false;` in my page_

The page is correctly working when I use the URL in my browser ( http://localhost:3002/agences/angers OR http://localhost:3002/agences/angers/page-1 ), but not with the `rewrite()` method. 

### What's the expected result?

The expected result is to render the same page that appear in the browser while using the full URL ( http://localhost:3002/agences/angers OR http://localhost:3002/agences/angers/page-1 ).

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-wtauce?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #11207, opened 2024-06-07
fix(rewrite): purge old data when rewriting

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11204

This PR fixes two issues:
- Inside `RenderContext`, `pathname` is now updated during the rewriting. To correctly infer the correct pathname (`RouteData.pathame` sometimes is undefined, and `RouteData.route` **ins't** correct e.g. `/some-slug/title` VS `/[slug]/title`, and we need the former)
- The `tryRewrite` now returns the new `URL`, since we had the logic here in all the pipelines
- We recompute the `AstroPartial` object when rewriting because during rewriting, we were sending old information, such as `params`

## Testing

Added two new test cases

## Docs

N/A

### DIFF
--- a/.changeset/fuzzy-eggs-kneel.md
+++ b/.changeset/fuzzy-eggs-kneel.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue in the rewriting logic, where old data weren't puged during the rewrite flow, causing some false positiving during the rendering phase.
--- a/.changeset/fuzzy-eggs-kneel.md
+++ b/.changeset/fuzzy-eggs-kneel.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Fixes an issue in the rewriting logic, where old data weren't puged during the rewrite flow, causing some false positiving during the rendering phase.
+Fixes an issue in the rewriting logic where old data was not purged during the rewrite flow. This caused some false positives when checking the validity of URL path names during the rendering phase.
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -20,7 +20,6 @@ import { renderEndpoint } from '../runtime/server/endpoint.js';
 import { renderPage } from '../runtime/server/index.js';
 import {
 	ASTRO_VERSION,
-	DEFAULT_404_COMPONENT,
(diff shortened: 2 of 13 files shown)


packages/astro/src/core/render-context.ts
```diff
@@ -388,6 +392,7 @@ export class RenderContext {
 			},
 		});
 
+		console.log('Astro.params', Astro.params);
```
matthewp, line 395: oops

packages/astro/src/core/render-context.ts
```diff
@@ -410,23 +415,22 @@ export class RenderContext {
 
 		const rewrite = async (reroutePayload: RewritePayload) => {
 			pipeline.logger.debug('router', 'Calling rewrite: ', reroutePayload);
-			const [routeData, component] = await pipeline.tryRewrite(
+			const [routeData, component, newURL] = await pipeline.tryRewrite(
 				reroutePayload,
 				this.request,
 				this.originalRoute
 			);
+			console.log(routeData);
```
matthewp, line 423: another console.log

.changeset/fuzzy-eggs-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue in the rewriting logic, where old data weren't puged during the rewrite flow, causing some false positiving during the rendering phase.
```
sarah11918, line 5:
```suggestion
Fixes an issue in the rewriting logic where old data was not purged during the rewrite flow. This caused some false positives when checking the validity of URL path names during the rendering phase.
``` 
We probably don't even need the second sentence (is it correct? I tried to infer from the PR description.) But if we do keep it, I think it's helpful to explain *what* the false positives were, so that people can compare it to what was happening in their project if they experienced this error!

(Would they know that whatever problem they were experienced was because data wasn't being purged? Is that meaningful to them as they use it? Or, did they have some problem and that was the underlying cause? If so, then connecting it to the problem they experienced is helpful!)

sarah11918 (commented):
Quick note on the changeset! :raised_hands:

sarah11918 (approved):
approving changset for docs!

matthewp (approved):

Issue #11103, opened 2024-05-20
Partials rendering regression in 4.8.5


Igloczek:
### Astro Info

```block
Astro                    v4.8.6
Node                     v20.12.2
System                   macOS (arm64)
Package Manager          pnpm
Output                   static (but it's broken with server + node.js adapter too)
Adapter                  none
Integrations             none
```

### Describe the Bug

Partials containing nested conditionally rendered elements are not properly rendering since 4.8.5

This is the minimal reproducible example I came with 

```astro
---
export const partial = true
---

{
  true && (
    <>
      {true && <div>test</div>}
      {false && <div>test</div>}
      {true && <div>test</div>}
    </>
  )
}
```
Replacing those `true` and `false` with some real variables and functions leads to broken output too.

Astro >=4.8.5 returns 200 OK, with empty body, no errors in the console. 

The same code, but not as partial works fine.
The same code, but on 4.7.1 and 4.8.4 works fine.

### What's the expected result?

I want to get either 200 with a proper HTML in the body, or 500 with error.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-xu4kcf?file=src%2Fpages%2Fpartial.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

alex-sherwin:
Seeing same, as of 4.8.5 partial rendering doesn't work.

A response is generated with any headers etc, but body is never sent over the wire in the response.

But the render code for the body runs, it's just not used in the response seemingly.

alex-sherwin:
Just looking at the diff for 4.8.4 to 4.8.5 https://github.com/withastro/astro/compare/astro%404.8.4...astro%404.8.5

The most likely change, I would think, is in the "Fix streaming in Node.js fast path" change here https://github.com/withastro/astro/commit/749a7ac967146952450a4173dcb6a5494755460c

In terms of actual, concrete behavioral differences made, this line stands out to me: https://github.com/withastro/astro/commit/749a7ac967146952450a4173dcb6a5494755460c#diff-d312d8fce663b920b083d795af0cf0ca4f20374f3467998906ec00ef2b85f6eaL280

Specifically noting that

```typescript
next = promiseWithResolvers<void>();
```

Is no longer run in the new code as of 4.8.5

I'm not really familiar with this area of code at all, but, purely speculating from the overall changes this feels like the most likely culprit

alex-sherwin:
[...]

PR #11210, opened 2024-06-07
Close iterator after rendering is complete and no more chunks remain

matthewp:
## Changes

- Our async iterable was closing too fast because with partials we skip the head parts. So there would be a pull of chunks that were empty even though rendering was still ongoing. The fix is to only end the iterator after both rendering is done and there are no more chunks to flush out.
- Closes https://github.com/withastro/astro/issues/11103

## Testing

- Test case added

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/purple-pianos-greet.md
+++ b/.changeset/purple-pianos-greet.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Close the iterator only after rendering is complete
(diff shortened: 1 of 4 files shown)


ematipico (approved):

ematipico:
@matthewp The CI failed three times on Node.js v20, which means it isn't flaky but it's a valid regression

matthewp:
@ematipico yes, i can recreate locally. It needs the exact version in our CI, so something might have changed in a patch release. I'll figure it out.

PR #11228, opened 2024-06-10
fix(astro): always return cloned content collection

### LINKED ISSUE #10700: Prod build collections cache can be modified

ascorbic:
## Changes

Currently, the first time `getCollection` is called in prod it returns the same array that is then cached, meaning that any mutations are applied to the cached value. Subsequent calls return a clone of the cached value (so these can be safely changed), but any changes made on the first call are still present. The original bug was fixed in #8022, but it missed the issue with the first invocation. This PR ensures that the array is always cloned before returning.

Fixes #10700

## Testing

Adds a test fixture based on the #10700 repro. Thanks @SStembridge.

### DIFF
--- a/.changeset/early-melons-thank.md
+++ b/.changeset/early-melons-thank.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Always return a cloned array from getCollection
+Updates `getCollection()` to always return a cloned array
--- a/.changeset/early-melons-thank.md
+++ b/.changeset/early-melons-thank.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Always return a cloned array from getCollection
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -79,8 +79,7 @@ export function createGetCollection({
 		// Cache `getCollection()` calls in production only
 		// prevents stale cache in development
 		if (!import.meta.env?.DEV && cacheEntriesByCollection.has(collection)) {
-			// Always return a new instance so consumers can safely mutate it
-			entries = [...cacheEntriesByCollection.get(collection)!];
+			entries = cacheEntriesByCollection.get(collection)!;
 		} else {
 			const limit = pLimit(10);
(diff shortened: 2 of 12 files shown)


.changeset/early-melons-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Always return a cloned array from getCollection
```
sarah11918, line 5:
```suggestion
Updates `getCollection()` to always return a cloned array
```
We're not super strict about patch PRs, but in general, if we can start with a present tense verb like Adds/Removes/Fixes/Refactors/Deprecates etc. then this tends to create a helpful descriptive sentence for readers who are inspecting the changelog to know what has changed, and whether they need to do anything.

Suggestion above, but as you'll get used to me saying, check for accuracy/nuance since you know the change better than I do!

sarah11918 (commented):
Congrats on your first core PR @ascorbic !  :partying_face: 

Might as well start you off on the right foot! Just made a small suggestion to the changeset for your consideration to give you a taste of what you're gonna get from me going forward... :smile: 

Welcome to core! :purple_heart:

matthewp (approved):

.changeset/early-melons-thank.md
ascorbic: Brilliant, thanks! That's just the sort of detail I need.

Issue #11221, opened 2024-06-09
Regression: `getViteConfig` doesn't work after 4.9.3 release


vchirikov:
### Astro Info

```block
4.9.2 - works
4.9.3+ (tested on 4.10.1) doesn't
```

### Describe the Bug

Regression: `getViteConfig` is broken after 4.9.3+ (including the latest 4.10.1) and doesn't return the correct config to vitest. 

In the gif I showed that the regression: 

`getViteConfig` works fine (it returns correct `environment` to vitest) and doesn't work at 4.10.1
(`ReferenceError: document is not defined` shows us that vitest's `environment` is not defined, I also re-checked this with `// @vitest-environment happy-dom`)

![bug](https://github.com/withastro/astro/assets/40987940/8a111d7a-1f24-4199-b02e-1b6ce576ceb8)

the vitest.config.ts:

```ts
/// <reference types="vitest" />

import { getViteConfig } from 'astro/config';

export default getViteConfig({
  test: {
    testTimeout: 10_000,
    css: false,
    globals: false,
    environment: 'happy-dom',
    include: ['./tests/**/*.test.{ts,tsx}'],
    setupFiles: [
      '@testing-library/react/dont-cleanup-after-each',
      './tests/vitest.setup.ts'
    ],
    pool: 'threads',
    poolOptions: {
      threads: {
        isolate: false,
        minThreads: 4,
      }
    },
    isolate: false,
    typecheck: {
      enabled: true,
      tsconfig: 'tsconfig.json',
    }
  },
}, {
  site: 'http://localhost:3000/base',
},
);

```
### Repro

https://github.com/vchirikov/astro_issue_11221

### What's the expected result?

`getViteConfig` returns the correct config with defined `environment`

vchirikov:
Related issues & PRs:

https://github.com/withastro/astro/issues/11175
https://github.com/withastro/astro/pull/11194

vchirikov:
@ematipico added repository with repro - https://github.com/vchirikov/astro_issue_11221

vchirikov:
Yep, but according the docs we should use the first arg. For now I use the same workaround:

```ts
import { getViteConfig, type ViteUserConfig } from 'astro/config';

// ...
const config = {
  test: { /* the actual vitest config*/} 
} satisfies { test: ViteUserConfig['test']; };

export default getViteConfig(
  { ...config },
  { vite: { ...config }},
);
```

PR #11231, opened 2024-06-11
fix: regression for getViteConfig

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11221

## Testing

I will create a preview release and test it

## Docs

N/A

### DIFF
--- a/packages/astro/src/config/index.ts
+++ b/packages/astro/src/config/index.ts
@@ -1,4 +1,4 @@
-import type { UserConfig } from 'vite';
+import type { UserConfig as UserViteConfig } from 'vite';
 import type { AstroInlineConfig, AstroUserConfig } from '../@types/astro.js';
 import { Logger } from '../core/logger/core.js';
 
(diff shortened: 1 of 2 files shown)


ematipico:
!preview vitest-config

ematipico:
Tested locally using one of the reproductions, and it works now

packages/astro/src/config/index.ts
```diff
@@ -1,12 +1,12 @@
...
+import type { UserConfig as UserViteConfig } from 'vite';
 import type { AstroInlineConfig, AstroUserConfig } from '../@types/astro.js';
 import { Logger } from '../core/logger/core.js';
 
 export function defineConfig(config: AstroUserConfig) {
 	return config;
 }
 
-export function getViteConfig(inlineConfig: UserConfig, inlineAstroConfig: AstroInlineConfig = {}) {
+export function getViteConfig(inlineUserViteConfig: UserViteConfig, inlineAstroConfig: AstroInlineConfig = {}) {
```
bluwy, line 9: nit, but could we rename this as `viteUserConfig: ViteUserConfig`? The order will also match Astro's internal type, like `AstroUserConfig` this way.

I also think that we could remove the "inline" because in Vite (and Astro), "inline" and "user" have two separate meanings. "inline" extends from the "user" interface.

ematipico: Done

bluwy (approved):

Issue #11181, opened 2024-06-04
The Astro got crashed when use the Container API and deploy on vercel with server mode


syhily:
### Astro Info

```block
Astro                    v4.9.2
Node                     v22.2.0
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/vercel/serverless
Integrations             none
```

### Describe the Bug

When the Astro project use the Container API internally. The build on vercel platform works as expected but soon crashed when access the production link. The error log is attached below:

```
u. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
at requireWithFriendlyError (/var/task/vercel/path0/node_modules/rollup/dist/native.js:59:9)
at Object.<anonymous> (/var/task/vercel/path0/node_modules/rollup/dist/native.js:68:76)
at Module._compile (node:internal/modules/cjs/loader:1358:14)
at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
at Module.load (node:internal/modules/cjs/loader:1208:32)
at Module._load (node:internal/modules/cjs/loader:1024:12)
at /opt/rust/nodejs.js:1:11497
at Function.Lt (/opt/rust/nodejs.js:1:11867)
at G.e.<computed>.z._load (/opt/rust/nodejs.js:1:11467)
at cjsLoader (node:internal/modules/esm/translators:348:17) {
[cause]: Error: Cannot find module '@rollup/rollup-linux-x64-gnu'
Require stack:
- /var/task/vercel/path0/node_modules/rollup/dist/native.js
at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
at Module._load (node:internal/modules/cjs/loader:986:27)
at /opt/rust/nodejs.js:1:11497
at Function.Lt (/opt/rust/nodejs.js:1:11867)
at G.e.<computed>.z._load (/opt/rust/nodejs.js:1:11467)
at Module.require (node:internal/modules/cjs/loader:1233:19)
at require (node:internal/modules/helpers:179:18)
at requireWithFriendlyError (/var/task/vercel/path0/node_modules/rollup/dist/native.js:41:10)
at Object.<anonymous> (/var/task/vercel/path0/node_modules/rollup/dist/native.js:68:76)
at Module._compile (node:internal/modules/cjs/loader:1358:14) {
code: 'MODULE_NOT_FOUND',
[...]

PR #11234, opened 2024-06-11
feat(@astrojs/react): export renderer for easy loading

ematipico:
## Changes

The change in API is **minimal**, but it allows to use the container API and renderer in SSR mode.

It addresses https://github.com/withastro/astro/issues/11181, but I will make another PR to add `import { renderer }` to the rest of the renderers 

## Testing

Created a new test case

## Docs

N/A 

I will update docs later once we agreed in this API


packages/integrations/react/server-v17.js
ematipico, line 1: I tried to create a `runtime.js` entry point, and re-export these values, but rollup has some issues with it. For the time being, this works

packages/integrations/react/server.js
```diff
@@ -230,3 +230,11 @@ export default {
 	renderToStaticMarkup,
 	supportsAstroStaticSlot: true,
 };
+
+export const renderer = {
+	ssr: {
```
matthewp, line 235: This seems non-ideal, these things are exported already, why do it a second time with a different shape?

matthewp: I think this might be because this is the shape that the container accepts. But the `{ ssr: ... }` shape is completely internal Astro code. Renderers should not need be aware of it.

I would change the types of container's `renderers` prop to accept the non-nested shape as well, and then "fix" it internally. That way only container's internal code is aware of this issue.

ematipico: Instead of "fixing" it, I think the container should do just a simple check about the "shape" of the renderer passed. If it doesn't contain `check` and `renderToStaticMarkup`, then we can throw an error with some hints on how to fix the issue.

matthewp (approved):
like this new approach.

PR #11236, opened 2024-06-11
fix(astro): handle symlinked content collection directories

### LINKED ISSUE #9088: Unknown Content Collection Error when `content` directory is a symlink
### Astro Info

```block
Astro                    v3.5.3
Package Manager          pnpm
Adapter                  none
                         @astrojs/react
```

_No response_

### Describe the Bug

When `./src/content` is a symlink and via `getEntry` data is retrieved from any collection, a `Unknown Content Collection Error` is thrown:

```
  Error reference:
    https://docs.astro.build/en/reference/errors/unknown-content-collection-error/
  File:
    /home/projects/github-tvxvbc/content/navigation/top.yml?astroDataCollectionEntry=true
  Stacktrace:
UnknownContentCollectionError
    at AstroError (/home/projects/github-tvxvbc/node_modules/astro/dist/core/errors/errors.js:33:5)
    at getEntryModuleBaseInfo (/home/projects/github-tvxvbc/node_modules/astro/dist/content/vite-plugin-content-imports.js:243:11)
```

It should be possible to retrieve collection data through symlinked `content` directories.

https://stackblitz.com/edit/github-tvxvbc

### Participation

```[tasklist]
### Tasks
```

ascorbic:
## Changes

Currently if you try to use a symlink in a content collection directory it will break the build. This is because Vite resolves the path to the symlink target, which is then outside of the content directory.

This PR tracks symlinks in the content directory, and then reverses them when transforming the content. This works out as a more robust way than intercepting the module resolution, which seemed to break many things in various places.

Fixes #9088

## Testing

Adds fixtures and test cases with symlinks

## Docs

This should be documented. I'm not sure if we document the limitation currently. This only supports symlinks at the top level (i.e. the link is `src/content/blog -> /Path/to/real/content/blog`).


ascorbic:
Hm. Looks like it's broken with caching.

.changeset/fifty-clouds-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Supports symlinked content collection directories
```
ematipico, line 5:
```suggestion
Fixes a case where symlinked content collection directories were not correctly resolved
```

A suggestion to hint that this is a bug fix. The previous phrase seemed more like a new feature

packages/astro/src/content/utils.ts
```diff
@@ -167,6 +167,50 @@ export function getEntryConfigByExtMap<TEntryType extends ContentEntryType | Dat
...
+	}
+
+	const contentDirEntries = await fsMod.promises.readdir(contentDir, { withFileTypes: true });
+	for (const entry of contentDirEntries) {
+		if (entry.isSymbolicLink()) {
+			const entryPath = path.join(contentDirPath, entry.name);
+			const realPath = await fsMod.promises.realpath(entryPath);
+			contentPaths.set(normalizePath(realPath), entry.name);
+		}
+	}
```
ematipico, line 187: nit: file system operations can come with runtime errors (permissions, etc.), I tend to wrap these operations in a `try/catch` to be more safe

packages/astro/src/content/utils.ts
```diff
@@ -167,6 +167,50 @@ export function getEntryConfigByExtMap<TEntryType extends ContentEntryType | Dat
...
+	return contentPaths;
+}
+
+export function reverseSymlinks({
```
ematipico, line 191: nit: `reverseSymlinks` -> `reverseSymlink` singular, because it returns a `string`, and not a `stringp[]`

ematipico (approved):
Just a few nits, otherwise it looks good!

packages/astro/src/content/utils.ts
ascorbic: The reason I did the plural is because there are multiple symlinks. But I can see what you mean.

packages/astro/src/content/utils.ts
ascorbic: I'd lean towards silently ignoring them here, because we're just looking for symlinks and a failure probably means it's not a symlink.

ematipico: It really depends on the use case, really. In this case we could just ignore the error, and maybe log and error/warning using the logger (you should have access to it).

ascorbic: Where would I find a logger instance? I'm calling this in [vite-plugin-content-imports](https://github.com/withastro/astro/blob/ab8746eea7dcbfadd7210c2fbfcb8b7dfca0ac3c/packages/astro/src/content/vite-plugin-content-imports.ts).

ematipico: It seems we don't pass it down to this plugin (the logger arrived *after* this plugin was created). You'll have to pass it like in this case:

https://github.com/withastro/astro/blob/2851b0aa2e2abe80ea603b53c67770e94980a8d3/packages/astro/src/core/create-vite.ts#L149

ascorbic: Aha, thanks. I just pushed an update with error handling and renamed funciton, but no logging. I'll add the logs.

ascorbic: OK, I've added logging now

matthewp (approved):

PR #11240, opened 2024-06-12
fix(astro): ignore query params when matching .astro extension

### LINKED ISSUE #11230: Virtual module ids with query parameters that end in `.astro` throw errors
### Astro Info

```block
Astro                    v4.10.1
Adapter                  none
Integrations             astro-test-invalid-transform
```

_No response_

### Describe the Bug

The [`transform` hook](https://github.com/withastro/astro/blob/734b98fecf0212cd76be3c935a49f84a9a7dab34/packages/astro/src/vite-plugin-astro/index.ts#L205) inside the [vite plugin `astro:build`](https://github.com/withastro/astro/blob/734b98fecf0212cd76be3c935a49f84a9a7dab34/packages/astro/src/vite-plugin-astro/index.ts) matches module ids that are ***not*** Astro files but contain paths to Astro files inside query parameters, for example:

```
virtual:test?importer=%2Fhome%2Fprojects%2Fgithub-vdjlmf%2Fsrc%2Fpages%2Findex.astro
```

This issue occurs because Astro uses `.endsWith` to check for Astro files
https://github.com/withastro/astro/blob/734b98fecf0212cd76be3c935a49f84a9a7dab34/packages/astro/src/vite-plugin-astro/index.ts#L205

```diff
- if (!id.endsWith('.astro') || parsedId.query.astro) {
-   return
- } 
+ if (!parsedId.filename.endsWith('.astro')) {
+   return
+ }
+ if (parsedId.query.astro) {
+   return
+ }
```

One workaround for this is to add an empty query parameter as a suffix:
```
virtual:test?importer=%2Fhome%2Fprojects%2Fgithub-vdjlmf%2Fsrc%2Fpages%2Findex.astro&end
```

Astro does not attempt to transform virtual module's that have `.astro` paths inside query parameters

https://stackblitz.com/edit/github-vdjlmf?file=astro.config.mjs,src%2Fpages%2Findex.astro

### Participation

ascorbic:
## Changes

Checks the filename for the `.astro`, rather than the full ID. Fixes #11230, where a query that ended in the string `.astro` would incorrectly match.

## Testing

I wasn't able to find a place where there were tests of this part of the code. If anyone has suggestions for a good place to test it then that would be welcome.

## Docs
Nothing changed

### DIFF
--- a/.changeset/pink-experts-count.md
+++ b/.changeset/pink-experts-count.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Ignores query strings in module identifiers when matching ".astro" file extensions in Vite plugin
--- a/packages/astro/src/vite-plugin-astro/index.ts
+++ b/packages/astro/src/vite-plugin-astro/index.ts
@@ -202,7 +202,7 @@ export default function astro({ settings, logger }: AstroPluginOptions): vite.Pl
 		async transform(source, id) {
 			const parsedId = parseAstroRequest(id);
 			// ignore astro file sub-requests, e.g. Foo.astro?astro&type=script&index=0&lang.ts
-			if (!id.endsWith('.astro') || parsedId.query.astro) {
+			if (!parsedId.filename.endsWith('.astro') || parsedId.query.astro) {
 				return;
 			}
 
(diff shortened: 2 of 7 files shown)


ematipico:
> I wasn't able to find a place where there were tests of this part of the code. If anyone has suggestions for a good place to test it then that would be welcome.

Can't you create a new fixture by replicating the reproduction provided in the original issue?

ascorbic:
@ematipico good suggestion

ematipico (approved):
Awesome!

Issue #11232, opened 2024-06-11
SSR behaviour different with `hybrid` mode and `node` adapter between Linux and MacOS (ARM)


LunaticMuch:
### Astro Info

```block
Astro                    v4.10.1
Node                     v20.11.1
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/node
Integrations             @astrojs/starlight
                         @astrojs/tailwind
```

### Describe the Bug

I recently opened an issue with [Astro Starlight](https://github.com/withastro/starlight/issues/1989) because I was seeing a weird behaviour with SSR.
In building the test case, I've noticed that the issue can be reproduced on MacOS (ARM) but not on codesandbox (or stackblitz) with Linux. It pointed me out to further debug and potentially rule out Starlight, and consider that the issue is within the adapter or compiler somewhere 😕  

None of the pages included in starlight are pre-rendered when the build runs on MacOS. They should be prerendered. 
However, when the build runs on Linux, they all get pre-rendered.

I tried different version of nodes too, as the default one in Stackblitz is v18 and I was running on v20, but this has not change anything.

In trying to reproduce, I collected logs and committed into my repo. Look for the following
- [MacOS](https://github.com/LunaticMuch/fuzzy-giggle/tree/main/logs/macos)
- [Codesandbox](https://github.com/LunaticMuch/fuzzy-giggle/tree/main/logs/codesandbox)

### What's the expected result?

The result is that the pages in starlight plugin get pre-rendered too. There's no reason for not rendering then and, given the `hybrid` configuration, by default they should be rendered.

### Link to Minimal Reproducible Example

https://stackblitz.com/~/github.com/LunaticMuch/fuzzy-giggle

### Participation

- [ ] I am willing to submit a pull request for this issue.

V3RON:
That's weird. I cloned you repository, try building it myself and experienced no issues at all, no matter which Node version I used (v18, v20, v21).

```
Astro                    v4.10.1
Node                     v18.20.3
System                   macOS (arm64)
Package Manager          unknown
Output                   hybrid
Adapter                  @astrojs/node
Integrations             @astrojs/starlight
                         @astrojs/tailwind
```
[See full logs here](https://github.com/user-attachments/files/15792177/log.txt)

Is there anything special about your environment? Could you purge `node_modules`, reinstall and try again? 🤔

LunaticMuch:
[...]

PR #11243, opened 2024-06-12
fix(astro): prerendering issue when path contains underscore

V3RON:
https://github.com/withastro/astro/blob/769f11f23f3dd32da912d0688008f9233ae13dd4/packages/astro/src/core/util.ts#L122-L129

Notice the condition in line 124. It checks for the presence of `_` in any part of the URL. The problem is that file is an absolute path, so `_GITHUB` is picked up by this condition, causing the page to be considered 'not public'.

## Changes

Updated isPublicRoute to use relative paths, ensuring correct handling of underscore folders outside the repository. This change ensures the function correctly returns 'true' when an underscore folder is present outside the repository folder, as intended.

Closes https://github.com/withastro/astro/issues/11232

## Testing

Added automated test to cover edge cases involving underscores in paths.

## Docs

No update needed


V3RON:
Quick note: packages/astro/test/fixtures/_underscore in folder name/node_modules/fake-astro-library/ is there on purpose to simulate case when page is exposed by a third-party library.

packages/astro/src/core/util.ts
```diff
@@ -120,11 +120,23 @@ function isInjectedRoute(file: URL, settings: AstroSettings) {
...
 
 function isPublicRoute(file: URL, config: AstroConfig): boolean {
-	const pagesDir = resolvePages(config);
-	const parts = file.toString().replace(pagesDir.toString(), '').split('/').slice(1);
+	const rootDir = config.root.toString();
+	const pagesDir = resolvePages(config).toString();
+	const fileDir = file.toString();
+
+	// Normalize the file directory path by removing the pagesDir prefix if it exists,
+	// otherwise remove the rootDir prefix.
```
ematipico, line 128: The comment should also explain **why** this is necessary. The code itself isn't too complicated to understand, so a comment that *explains* seems superfluous. However, it's worth adding an explain of the motive, because if you left a comment here, it means there's a special case to handle :)

.changeset/honest-ravens-double.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Fixes prerendering issue for libraries in node_modules when a folder with an underscore is in the path.
```
ematipico, line 5:
```suggestion
Fixes a prerendering issue for libraries in `node_modules` when a folder with an underscore is in the path.
```

ematipico (approved):

packages/astro/src/core/util.ts
V3RON: > It might happen that a page is located in `node_modules`, and the Astro project itself is in a path containing the `_` character. In such cases, we should use a relative path to bypass the segment containing `_`.

WDYT?

matthewp (approved):
great test, thanks!

kirso:
I am facing the same issue, after changing to `hybrid`, I don't get any html pages in my .dist folder at all. Although I wonder if its the same reason. 

The way how I found out is by adding a server endpoint (with `prerender=false`) and my Pagefind now fails the build because it doesn't index any .html files.

I checked the solution here, but I don't have any folders with the `_` before it so I wonder if it is what is causing an issue.

The repo is here: https://github.com/kirso/KirillSo.com

Issue #11208, opened 2024-06-07
Experimental rewrite: Response with status:500 does not trigger 500.astro in production


arty-name:
### Astro Info

```block
Astro                    v4.10.0
Node                     v20.13.1
System                   Linux (x64)
Package Manager          unknown
Output                   server
Adapter                  @astrojs/node
Integrations             @sentry/astro
```

### Describe the Bug

I conditionally rewrite a page to one of two other pages, and both branches work fine.

Sometimes these pages throw errors, and the handling of this condition has been improved in #11110 :+1: 

In production mode I show a nice error page provided in 500.astro and this works. It works even when I return `new Response(null, { status: 500 })`. 

However if rewriting took place during the processing of the request, 500.astro is not rendered and the plain Response#status=500 is rendered.

Background for returning Response#status=500 is this. I use JSON logs in production to collect them in Kibana and process them automatically: each log prints just one line which is a JSON object. By default Astro prints the multiple-line stack trace for uncaught exceptions, so I catch exceptions in middleware and log them. Then I return `new Response(null, { status: 500 })` so that Astro renders 500.astro.

### What's the expected result?

When there’s 500.astro page it should be rendered every time Astro receives Response#status=500, regardless of rewriting.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-h5n839-ifnpfx?file=src%2Fpages%2Findex.astro

### Participation

- [ ] I am willing to submit a pull request for this issue.

arty-name:
Thank you for the quick response @ematipico!

I just wanted to quickly confirm with you: will this change also fix the behavior in production mode? It’s the production mode I had the issue with, but maybe haven’t stressed this enough in the ticket description.

PR #11244, opened 2024-06-12
fix: better DX for 500.astro in local development

ematipico:
## Changes

This PR enhance the usage of the `500.astro` page when rewriting.

Before, in dev mode, we were not even displaying the custom `500.astro`, probably because we wanted to show the original error. However, this was causing more issues when using the rewriting, because users needed to rewrite using the `500.astro`, and see the results.

This PR catches the error in dev mode, and renders the custom 500 if present. I also logs the original error, because that won't be visible in the default `500.astro`

Closes https://github.com/withastro/astro/issues/11208

## Testing

Added new test cases to test the custom `500.astro` and the default 500 page (none)

## Docs

N/A


packages/astro/src/core/constants.ts
```diff
@@ -32,6 +32,11 @@ export const ROUTE_TYPE_HEADER = 'X-Astro-Route-Type';
  */
 export const DEFAULT_404_COMPONENT = 'astro-default-404.astro';
 
+/**
+ * The value of the `component` field of the default 404 page, which is used when there is no user-provided 404.astro page.
```
florian-lefebvre, line 36:
```suggestion
 * The value of the `component` field of the default 500 page, which is used when there is no user-provided 404.astro page.
```

packages/astro/src/core/routing/astro-designed-error-pages.ts
```diff
@@ -16,6 +16,20 @@ export const DEFAULT_404_ROUTE: RouteData = {
 	isIndex: false,
 };
 
+export const DEFAULT_500_ROUTE: RouteData = {
+	component: DEFAULT_500_COMPONENT,
+	generate: () => '',
+	params: [],
+	pattern: /\/500/,
+	prerender: false,
+	pathname: '/404',
```
florian-lefebvre, line 25:
```suggestion
	pathname: '/500',
```

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -273,7 +278,21 @@ export async function handleRoute({
...
-	let response = await renderContext.render(mod);
+	let response;
+	try {
+		response = await renderContext.render(mod);
+	} catch (err: any) {
+		const custom500 = getCustom500Route(manifestData);
+		if (!custom500) {
+			throw err;
+		}
+		logger.error('router', err.stack || err.message);
```
florian-lefebvre, line 289: If I understand correctly, the goal is to have great error logs since the page will probably display less data than the default error overlay?

florian-lefebvre (commented):
I think we need a changeset for this?

packages/astro/src/vite-plugin-astro-server/route.ts
ematipico: Yes exactly. Your 500 props feature will allow users to better handle the error, but since it's a feature, for now I thought a log could be enough

florian-lefebvre: Make sense! Even after the prop introduction, keeping this log makes sense.
```suggestion
		// Log useful information that the custom 500 page may not display unlike the default error overlay
		logger.error('router', err.stack || err.message);
```

florian-lefebvre:
Related docs PR https://github.com/withastro/docs/pull/8495

florian-lefebvre (approved):
LGTM! Don't forget to update the PR I linked, specifically the part about enabling in development

PR #11249, opened 2024-06-13
Fix a performance issue with JSON schema generation

markgaze:
## Changes

I have a large number of items in my content collections and I've noticed a considerable slowness when starting the dev server, having looked into it (with `--verbose`), I noticed that the JSON schema files were being changed repeatedly which made the HMR spend approximately 90ms reloading it. With a lot of content files, this quickly adds up and made my startup times approach a minute in some cases 🙈 

This PR changes the behaviour so that the JSON schema is only generated once rather than in the `for` loop like it was before, I can't think of any reason that it needed to be in that `for` loop but other people may know better than me!

## Testing

I've tested this change in the project that was having the performance issue and it only generates the JSON schemas once and the dev server starts instantly, as it did before.

I also ran the unit tests locally and they still generate the same JSON schema files

I haven't added a test for this scenario, as it could be quite difficult to prove that it was slow but isn't now but I'm happy to look into it, if it's of value!

If nothing else, I could throw together a StackBlitz example that shows the issue, if needs be 🙂 

## Docs

Nothing changes for the end user, so no new docs required

### DIFF
--- a/packages/astro/src/content/types-generator.ts
+++ b/packages/astro/src/content/types-generator.ts
@@ -449,41 +449,42 @@ async function writeContentFiles({
 				for (const entryKey of Object.keys(collection.entries).sort()) {
 					const dataType = collectionConfig?.schema ? `InferEntrySchema<${collectionKey}>` : 'any';
 					dataTypesStr += `${entryKey}: {\n	id: ${entryKey};\n  collection: ${collectionKey};\n  data: ${dataType}\n};\n`;
-					if (
-						settings.config.experimental.contentCollectionJsonSchema &&
-						collectionConfig?.schema
-					) {
-						let zodSchemaForJson =
-							typeof collectionConfig.schema === 'function'
-								? collectionConfig.schema({ image: () => z.string() })
-								: collectionConfig.schema;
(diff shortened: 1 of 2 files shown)


alexanderniebuhr (approved):

ematipico (approved):

packages/astro/src/content/types-generator.ts
```diff
@@ -449,41 +449,42 @@ async function writeContentFiles({
...
-									2
-								)
-							);
-						} catch (err) {
-							logger.warn(
-								'content',
-								`An error was encountered while creating the JSON schema for the ${entryKey} entry in ${collectionKey} collection. Proceeding without it. Error: ${err}`
-							);
-						}
+					dataTypesStr += `};\n`;
```
syhily, line 452: I think this line shouldn't be moved to here. Right?

markgaze: You're right. My bad, sorry! I was having some formatting issues in VS Code and I think this happened when I was trying to fix it. I'll see if I can add some tests around this to make sure it doesn't happen again!

PR #11250, opened 2024-06-13
fix(markdown-remark): mark prism package as an internal dependency

V3RON:
## Changes

This package wasn't marked as an internal dependency correctly, which introduced a race condition in the build process. Starting from now, `@astrojs/prism` will always be built first.

See discussion on Discord (contribute) for details.

## Testing

Clear Turborepo cache, try building `@astrojs/markdown-remark` - you will see an error regarding `@astrojs/prism`.

Checkout to this branch, do the same steps - you will see no error, `@astrojs/prism` will be built first.

## Docs

No changes needed

### DIFF
--- a/.changeset/few-beans-tickle.md
+++ b/.changeset/few-beans-tickle.md
@@ -1,0 +1,9 @@
+---
+'@astrojs/markdown-remark': patch
+---
+
+Mark @astrojs/prism as an internal dependency
+
+
+This package wasn't marked as an internal dependency correctly, which introduced a race condition in the build process.
+Starting from now, @astrojs/prism will always be built first.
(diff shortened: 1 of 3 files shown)


.changeset/few-beans-tickle.md
bluwy, line 1: I don't think we need a changeset here as the published version should still be the same for `@astrojs/prism`. So there's no change if we do another publish.

V3RON: I thought so, but wasn't sure if it's OK to have a PR without a changeset. Should I simply remove it?

bluwy: Yeah you can remove it. We only need a changeset if it requires publishing a release and it affects users.

V3RON: Done!

bluwy (approved):

PR #11253, opened 2024-06-14
Add base to paginate

### LINKED ISSUE #10358: The "page" prop in paginated dynamic routes doesn’t handle base URLs
### Astro Info

```block
Astro                    v4.4.13
Adapter                  none
Integrations             astro-imagetools
                         @astrojs/vue
```

_No response_

### Describe the Bug

When having a `base` option configured, the `page` prop resulting from having a dynamic route with paginated content doesn’t take that base into account.

This means that the URLs in `page.url.prev` and `page.url.next` are not prefixed with the configured `base`. This (strangely) doesn't cause issues in development mode, but when running a preview after build, the links don’t work correctly.

I believe `page.url.prev` and `page.url.next` should be correctly prefixed with `base` if it is defined in the configuration. If there are issues with that, the behaviour should at least be documented so developers can manually handle the case.

https://stackblitz.com/edit/github-kzsh3y?file=src%2Fpages%2Findex.astro

### Participation

kevinzunigacuellar:
## Changes

- Closes #10358
- Adds base url to next and prev links in `paginate()`

## Testing

Updated tests

## Docs

~~https://github.com/withastro/docs/pull/8569~~

https://github.com/withastro/docs/pull/9200


packages/astro/src/core/render/paginate.ts
```diff
@@ -77,12 +79,11 @@ export function generatePaginateFunction(
 	};
 }
 
-function correctIndexRoute(route: string) {
+function addRouteBase(route: string, base: string) {
```
ematipico, line 82: I would type `base` using `AstronConfig['base']`, so if the type of `base` changes, this one changes too

packages/astro/src/core/render/paginate.ts
```diff
@@ -77,12 +79,11 @@ export function generatePaginateFunction(
...
 	// `routeMatch.generate` avoids appending `/`
 	// unless `trailingSlash: 'always'` is configured.
 	// This means an empty string is possible for the index route.
-	if (route === '') {
-		return '/';
-	}
-	return route;
+	let routeWithBase = base === '/' ? route : removeTrailingForwardSlash(base) + route;
```
ematipico, line 86: You can use the `joinPaths` utility here: 

https://github.com/withastro/astro/blob/d4336a5470b1e4265e8329407efcd2898fa7d368/packages/internal-helpers/src/path.ts#L63

ematipico (commented):
A changeset is missing

bluwy:
There's no changeset, but presumably this would be a major as existing users would've manually added the base now, and we might break them?

ematipico:
But isn't this a bug? Users were using a workaround for the bug. Or do we document that those URLs were meant to be without base?

packages/astro/src/core/render/paginate.ts
kevinzunigacuellar: I thought about using `joinPaths` but wouldn't it remove the `trailingSlash` in case it is set to `"always"`?. My assumption was that both base and route urls were coming in already formatted so I didn't want override them.

packages/astro/src/core/render/paginate.ts
kevinzunigacuellar: good observation! maybe I could also update other types that were changed when prop drilling

packages/astro/src/core/render/paginate.ts
kevinzunigacuellar: disregard my last comment, updated in https://github.com/withastro/astro/pull/11253/commits/67a81bbf52b8861b8a8b2b69f520242ebc86e98a

bluwy:
@ematipico The [docs](https://docs.astro.build/en/reference/api-reference/#paginate) doesn't explicit mention about the base. I think it has always worked this way for very long and it's risky to fix it now even if it may be a bug.

I was thinking of documenting that you need to add the base manually. Or maybe `paginate()` can accept an option to prepend base, which we will turn on automatically in Astro 5?

ematipico:
> @ematipico The [docs](https://docs.astro.build/en/reference/api-reference/#paginate) doesn't explicit mention about the base. I think it has always worked this way for very long and it's risky to fix it now even if it may be a bug.
> 
> I was thinking of documenting that you need to add the base manually. Or maybe `paginate()` can accept an option to prepend base, which we will turn on automatically in Astro 5?

Yeah, that sounds like a reasonable approach!

kevinzunigacuellar:
I am leaning towards the most simple solution to just edit docs and mention that base is not prepended by default. If it's okay with everyone, I will close the PR an open one in docs. Thanks for all the support @bluwy and @ematipico

ematipico:
> I am leaning towards the most simple solution to just edit docs and mention that base is not prepended by default. If it's okay with everyone, I will close the PR an open one in docs. Thanks for all the support @bluwy and @ematipico

Make sure to **not** delete the branch. We will recreate this PR once we start the works towards Astro 5.0

.changeset/twenty-cobras-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
+---
+
+Adds base url to next and prev url in `paginate()` utility
```
sarah11918, line 5:
``````suggestion
Changes the data returned for `page.url.next` and `page.url.prev` to include the value set for `base` in your Astro config. 

Previously, you had to manually prepend your configured value for `base`  to the URL path.  Now, Astro automatically includes your `base` value to `next` and `prev` URLs.

If you are using the `paginate()` function for "previous" and "next" URLs, remove any existing `base` value as it is now added for you:

```diff

show a line of code plz!
```

``````
Assuming this is the change for 5.0, would love to see an example and explicit instruction here showing people a line of code they'll have to change!

Something kind of like this? (Pls make it accurate)

**Review by sarah11918 — COMMENTED** [review]
Suggestion for a style of changeset here! Also, I think @bluwy is on the case, but if this is NOT currently associated with a Linear ticket tagged "breaking change", then it's not in my system!  :smile:

File: .changeset/twenty-cobras-push.md (hunk 3 above)
Comment (ematipico) [reply .changeset/twenty-cobras-push.md:5]: I'll commit the suggestion and follow up with a commit that shows the diff

**Comment by ematipico** [comment]
@sarah11918 Code snippet added here [`9403870` (#11253)](https://github.com/withastro/astro/pull/11253/commits/9403870e1d84c07c2b4fca3f17e262acef46a619)

**Comment by sarah11918** [comment]
Quick reminder (as will be the process going forward) that this will require an update to the `5.0.0-beta` branch v5 upgrade guide under the appropriate heading there!

@kevinzunigacuellar is a maintainer, so can commit directly there. But, might not make the obvious connection to the Linear task. So especially @ematipico and @bluwy whom I suspect will be helping out a lot with non-core PRs, please help everyone out who doesn't use Linear!  🙌

**Review by delucis — COMMENTED** [review]
Thanks for the PR @kevinzunigacuellar!

How does this change impact pages not directly in `src/pages/`?

As an example:

- I have `base: "blog"` in my config
- I have `src/pages/categories/[...category].astro` which uses `paginate()`

The correct inclusion of `base` would be `/blog/categories/2`. 

Does this PR do that or does it generate `/categores/blog/2`?

**Review by matthewp — APPROVED** [review]

File: .changeset/twenty-cobras-push.md [hunk 4]
```diff
@@ -0,0 +1,32 @@
+---
+'astro': major
+---
+
+Changes the data returned for `page.url.next` and `page.url.prev` to include the value set for `base` in your Astro config. 
+
+Previously, you had to manually prepend your configured value for `base`  to the URL path.  Now, Astro automatically includes your `base` value to `next` and `prev` URLs.
```
sarah11918, line 7:
```suggestion
Previously, you had to manually prepend your configured value for `base` to the URL path.  Now, Astro automatically includes your `base` value in `next` and `prev` URLs.
```

.changeset/twenty-cobras-push.md
```diff
@@ -0,0 +1,32 @@
...
+    astronaut: 'Buzz Aldrin',
+  }, {
+    astronaut: 'Sally Ride',
+  }, {
+    astronaut: 'John Glenn',
+  }];
+  return paginate(astronautPages, { pageSize: 1 });
+}
+const { page } = Astro.props;
+// "docs" is the `base` configured in astro.config.mjs
```
sarah11918, line 26:
```suggestion
//  `base: /docs` configured in `astro.config.mjs` 
```

sarah11918:
Noting that the place to update this information in the upgrade guide is here:

Preview: https://5-0-0-beta--astro-docs-2.netlify.app/en/guides/upgrade-to/v5/#changed-add-base-to-paginate-plt2204
https://github.com/withastro/docs/edit/5.0.0-beta/src/content/docs/en/guides/upgrade-to/v5.mdx (line 223)

Will also require an update to: 
https://docs.astro.build/en/reference/api-reference/#pageurlprev and following entries that all say "If a value is set for base, prepend the base path to the URL"

sarah11918:
@kevinzunigacuellar Does this also affect `page.url.first` and `page.url.last`? If so, then the changeset shouldn't only mention previous and next. If not, isn't that kind of confusing? :sweat_smile: 

Note: I think first and last are very recent additions, maybe even before you made this PR!

ematipico:
> Thanks for the PR @kevinzunigacuellar!
> 
> How does this change impact pages not directly in `src/pages/`?
> 
> As an example:
> 
>     * I have `base: "blog"` in my config
> 
>     * I have `src/pages/categories/[...category].astro` which uses `paginate()`
> 
> 
> The correct inclusion of `base` would be `/blog/categories/2`.
> 
> Does this PR do that or does it generate `/categores/blog/2`?

The `base` is always prefixed to any route, regardless of their position the file system. In your example, the emitted route is `/blog/categories/2`

delucis (approved):
> The `base` is always prefixed to any route, regardless of their position the file system. In your example, the emitted route is `/blog/categories/2`

Fantastic, thanks for the clarification @ematipico. In that case approving conceptually 😁 

Thanks again for the PR @kevinzunigacuellar and for taking it to the finish line Ema 💜

PR #11261, opened 2024-06-14
Fix backwards compat with Astro <= 4.9 in Node and Vercel adapters

matthewp:
## Changes

- Even though the code uses a dynamic import with a catch, this still breaks in the build because Rollup tries to build that import.
- Moving the import specifier to a variable lets us hide from Rollup, while still dynamically importing this module at runtime.

## Testing

- Can't create unit tests for this since it requires Astro <= 4.9. Tested manually.

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/serious-humans-obey.md
+++ b/.changeset/serious-humans-obey.md
@@ -1,0 +1,6 @@
+---
+'@astrojs/vercel': patch
+'@astrojs/node': patch
+---
+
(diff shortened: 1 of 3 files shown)


Fryuni (approved):

florian-lefebvre (approved):
Thanks! We'll also need this for netlify and cloudflare

PR #11216, opened 2024-06-09
Export type Database from db/runtime

OliverSpeir:
This type used to be exported, was moved around a bit starting in https://github.com/withastro/astro/pull/10457 and ended up no longer being exported

This PR moves the import and export to runtime/index.ts to ensure it ends up exported from `@astrojs/db/runtime`

PR #11262, opened 2024-06-14
Import Database type from correct file

nezouse:
## Changes
Following #11216 `Database` type was moved between files, but dts wasn't updated. In @astrojs/db 0.11.15 db is always type any as it can't be found in `'./dist/_internal/runtime/virtual.js'`

Sorry if there is anything wrong with the change, just started using the lib today.

### DIFF
--- a/.changeset/nice-pillows-teach.md
+++ b/.changeset/nice-pillows-teach.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/db': patch
+---
+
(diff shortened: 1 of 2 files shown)


OliverSpeir (approved):

OliverSpeir:
Good catch thank you, any chance you can make a changeset as a patch?

matthewp (approved):

PR #11263, opened 2024-06-14
chore: logging cleanup

wackbyte:
## Changes

- Use the logger provided by Astro more consistently in integrations.
- Remove an old duplicate check in the sitemap integration.

## Testing

No tests added because there are no integration log tests (to my knowledge).

## Docs

No docs added because this is strictly stylistic.

### DIFF
--- a/.changeset/thin-icons-yell.md
+++ b/.changeset/thin-icons-yell.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/sitemap': patch
+'@astrojs/node': patch
+'@astrojs/mdx': patch
+---
+
+Refactor to use Astro's integration logger for logging
--- a/packages/integrations/sitemap/src/index.ts
+++ b/packages/integrations/sitemap/src/index.ts
@@ -88,15 +88,7 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
 
 					const { filter, customPages, serialize, entryLimit } = opts;
 
-					let finalSiteUrl: URL;
-					if (config.site) {
-						finalSiteUrl = new URL(config.base, config.site);
-					} else {
-						console.warn(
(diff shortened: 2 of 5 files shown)


packages/integrations/sitemap/src/index.ts
```diff
@@ -88,15 +88,7 @@ const createPlugin = (options?: SitemapOptions): AstroIntegration => {
...
-					let finalSiteUrl: URL;
-					if (config.site) {
-						finalSiteUrl = new URL(config.base, config.site);
-					} else {
-						console.warn(
-							'The Sitemap integration requires the `site` astro.config option. Skipping.'
-						);
-						return;
-					}
+					let finalSiteUrl = new URL(config.base, config.site);
```
florian-lefebvre, line 91:
```suggestion
					if (!config.site) {
						logger.warn('The `site` astro.config option is required. Skipping.');
						return;
					}
					let finalSiteUrl = new URL(config.base, config.site);
```

florian-lefebvre (commented):
Can you add a changeset (patch)? You can get one by running `pnpm changeset` from the root

packages/integrations/sitemap/src/index.ts
wackbyte: This is already checked a couple lines above this :)

.changeset/thin-icons-yell.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/sitemap': patch
+'@astrojs/node': patch
+'@astrojs/mdx': patch
+---
+
+chore: logging cleanup
```
bluwy, line 7:
```suggestion
Refactors to use Astro's integration logger for logging
```

Perhaps this is clearer of what it changes.

bluwy (approved):

bluwy (approved):

PR #11264, opened 2024-06-15
fix(astro): type generation for empty collections

### LINKED ISSUE #11075: Empty Content Collections with defined scema don't generate typescript types
### Astro Info

```block
Astro                    v4.8.4
Node                     v22.0.0
Package Manager          pnpm
Adapter                  none
                         @astrojs/svelte
```

_No response_

### Describe the Bug

I want to publish a blank vesion of my website with no content. I already have a schema set up on config.ts, and doing a dev build complains but works no problem. However, running a full production builds throws many typescript errors (specifically that something's of a type unknown) because the type wasn't properly generated. The example I provided should be enough to get the bug to trigger, but I wasn't able to.

In this scenario, I would expect a build with no pages, similar to how a dev build works

https://stackblitz.com/edit/github-2n2i41

### Participation

Fryuni:
## Changes

- Generate a placeholder type to allow for type-safety, intellisense and autocomplete on empty collections.
- Closes #11075

## Testing

Expected types added on tests for `astro sync`

## Docs

I don't think this needs any docs change


.changeset/light-bugs-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes type generation for empty collections
```
florian-lefebvre, line 5:
```suggestion
Fixes type generation for empty content collections
```

Fryuni: Also applies for data collections

florian-lefebvre: True but I think the term "content collections" cover the usage of defineCollection basically. I mean iirc in the docs content collections is always used

packages/astro/src/content/types-generator.ts
```diff
@@ -429,12 +429,17 @@ async function writeContentFiles({
 					collectionConfig?.type ?? 'data'
 				: collection.type;
 
+		const collectionEntryKeys = Object.keys(collection.entries).sort();
+		const dataType = collectionConfig?.schema ? `InferEntrySchema<${collectionKey}>` : 'any';
 		switch (resolvedType) {
 			case 'content':
+				if (collectionEntryKeys.length === 0) {
+					contentTypesStr += `${collectionKey}: Record<string, {\n  id: string;\n  slug: string;\n  body: string;\n  collection: ${collectionKey};\n  data: ${dataType};\n  render(): Render[".md"];\n}>;\n`;
```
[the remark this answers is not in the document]
Fryuni: Good point, I'll try to save some time to do a little refactor around that.

ematipico (approved):

PR #11271, opened 2024-06-17
feat(underscore-redirects): add support for force

florian-lefebvre:
## Changes

- Add support for `force`, which adds a `!` after the status

## Testing

Added one test, existing ones should pass

## Docs

N/A


.changeset/funny-cats-sell.md
```diff
@@ -0,0 +1,16 @@
+---
+'@astrojs/underscore-redirects': minor
```
bluwy, line 2: The package is v0, so features should be a patch
```suggestion
'@astrojs/underscore-redirects': patch
```

bluwy (approved):
LGTM. Seems like a netlify only feature: https://docs.netlify.com/routing/redirects/redirect-options/#force-redirects

ematipico (approved):

Issue #11235, opened 2024-06-11
Rewriting to the root doesn't work with trailingSlash: never & rewrite to non-existed page/404 issues


vchirikov:
### Astro Info

```block
astro 4.10.2
```

### Describe the Bug

1. rewriting to `/` doesn't work with `trailingSlash: 'never'` config setting
1. rewriting to `/404` doesn't work at all (with or without context (`next('/404')` and `ctx.rewrite('/404')`), might be the same behavior for 500 (?)
1. astro's pipeline throws the same error type for a non-existed page or rendering error (so we don't know which error page should show - 404 or e.g. 500, [it's bad for i18n](https://github.com/withastro/roadmap/discussions/950) )

![bug](https://github.com/withastro/astro/assets/40987940/e9527556-7fa5-416a-8e60-7d36ed0d8bc7)

### What's the expected result?

rewriting works fine with `/404` and `/` with or without trailingSlash setting

### Link to Minimal Reproducible Example

https://github.com/vchirikov/astro_issue_01

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
Downgrading to a P3 because this is an experimental feature, and I doubt we have a lot of users using it

vchirikov:
@ematipico also I found that rewriting doesn't change the status code of response, e.g. I want to rewrite from `/en/about` to `/pages/about.astro` I do something like this:

```ts
const response = await next('/about');
console.log('return response from rewrite handler', response)
return response;
```
see nodejs in logs:

```js
16:36:49 [watch] src/middleware.ts
return response from rewrite handler Response {
  status: 200,
  statusText: 'OK',
  headers: Headers {
    'content-type': 'text/html',
    'X-Astro-Route-Type': 'page',
    'X-Astro-Reroute': 'no'
  },
  body: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
  bodyUsed: false,
  ok: true,
  redirected: false,
  type: 'default',
  url: ''
}
16:36:49 [404] /en/about 187ms
```

note the `response.status` is 200, but astro returns 404 to the browser:

![image](https://github.com/withastro/astro/assets/40987940/e69db9e3-4b67-4b7d-8c95-a3edc2701d13)

vchirikov:
@ematipico I've tested 4.11 and [404 on rewriting is still the issue](https://github.com/withastro/astro/issues/11235#issuecomment-2163478342)

[...]

PR #11272, opened 2024-06-17
fix(rewrite): allow to rewrite 404 and take `base` into consideration

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11235

This PR fixes few bugs and reverts some changes:
- It's now possible to rewrite `404` (again). I believe it makes sense to rewrite `/404` even in SSG
- There were few bugs when calculating the URL to rewrite when `trailingSlash` is `"never"` and the user uses `Astro.rewrite("/")`
- I refactored the logic of `tryRewrite` and moved inside a generic function, since there was shared logic
- 

## Testing

I created new test cases with custom `404.astro` and default 404.
I created new test cases with `trailingSlash: "never"`

## Docs

N/A

### DIFF
--- a/.changeset/honest-shirts-trade.md
+++ b/.changeset/honest-shirts-trade.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Reverts a logic where it was possible to rewrite `/404` in static mode. It's **now possible** - again - to rewrite `/404`.
--- a/.changeset/honest-shirts-trade.md
+++ b/.changeset/honest-shirts-trade.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Reverts a logic where it was possible to rewrite `/404` in static mode. It's **now possible** - again - to rewrite `/404`.
+Reverts a logic where it wasn't possible to rewrite `/404` in static mode. It's **now possible** again
--- a/packages/astro/src/container/pipeline.ts
+++ b/packages/astro/src/container/pipeline.ts
@@ -14,6 +14,7 @@ import {
 	createStylesheetElementSet,
 } from '../core/render/ssr-element.js';
 import { DEFAULT_404_ROUTE } from '../core/routing/astro-designed-error-pages.js';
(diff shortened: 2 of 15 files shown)


packages/astro/src/container/pipeline.ts
```diff
@@ -77,22 +77,26 @@ export class ContainerPipeline extends Pipeline {
...
-			}
 			if (route.routeData.pattern.test(decodeURI(finalUrl.pathname))) {
 				foundRoute = route.routeData;
 				break;
-			} else if (finalUrl.pathname === '/404') {
+			}
+		}
+
+		if (!foundRoute) {
+			if (finalUrl.pathname.includes('/404')) {
```
matthewp, line 95: should this be `startsWith` instead?

vchirikov: I think `/404` might be used with locale, e.g. `/de/404` -> 404 + `Astro.currentLocale == de`

ematipico: Good catch Matthew, I should update this pipeline and use the same logic of other pipelines.

ematipico: > I think `/404` might be used with locale, e.g. `/de/404` -> 404 + `Astro.currentLocale == de`

This isn't allowed, a 404 is always at root level.

What you want is to "mark" a route, and assign a locale, which is a feature we don't have yet. However you can use the middleware+locals

matthewp: That's not a special route though, is it?

.changeset/honest-shirts-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Reverts a logic where it was possible to rewrite `/404` in static mode. It's **now possible** - again - to rewrite `/404`.
```
florian-lefebvre, line 5:
```suggestion
Reverts a logic where it wasn't possible to rewrite `/404` in static mode. It's **now possible** again
```

matthewp (approved):

Issue #10785, opened 2024-04-13
manual i18n routing not working with prerendered 404 page


stefanprobst:
### Astro Info

```block
Astro                    v4.6.1
Node                     v20.12.2
System                   Linux (x64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/node
Integrations             none
```

### Describe the Bug

with the recently added `manual` i18n routing strategy, and [this middlwware](https://github.com/stefanprobst/issue-astro-i18n-manual/blob/main/src/middleware.ts), i want to redirect the "/" root route to the default locale "/en".

in a production build, when the `src/pages/404.astro` file has `export const prerender = false`, i get the expected 307 redirect. however, when `404.astro` has `export const prerender = true` i get a 404.

---

to reproduce:

1. `git clone git@github.com:stefanprobst/issue-astro-i18n-manual.git && cd issue-astro-i18n-manual && pnpm install`
2. `pnpm run build`
3. `pnpm run preview`
4. `curl -i http://localhost:3000` which prints: "HTTP/1.1 404 Not Found"

now, change the prerender mode in 404.astro to `false`:

1. `sed -i "s/prerender = true/prerender = false/" src/pages/404.astro`
2. `pnpm run build`
3. `pnpm run preview`
4. `curl -i http://localhost:3000` which prints: "HTTP/1.1 307 Temporary Redirect"

### What's the expected result?

root route "/" should be redirected to default locale "/en" when `404.astro` has `export const prerender = true` (or with `output: "hybrid"`).

### Link to Minimal Reproducible Example

https://github.com/stefanprobst/issue-astro-i18n-manual

### Participation

- [ ] I am willing to submit a pull request for this issue.

matthewp:
@ascorbic this could be a fun one to understand the i18n logic. @ematipico can help you with this one possibly.

ascorbic:
Hey! It looks like the issue here is that middleware isn't being invoked on the request for `/`, so the manual redirect isn't happening. For prerendered pages, middleware is run at build time by default. It's only run at request time for SSR routes, or if you are using edge middleware. You have no `index.astro` in the root, so it's not prerendered - meaning the middleware is not run. For that reason, when a request comes for it at runtime it's not found. Because the 404 is a prerendered route, the middleware isn't invoked at runtime, meaning it can't intercept the request and redirect it.

[...]

PR #11273, opened 2024-06-17
fix(astro): don't run middleware in dev for prerendered 404 routes

ascorbic:
## Changes

When a 404 page is prerendered, middleware is not run in production. This is because middleware is only run at build time for prerendered pages. Currently this behaves differently in dev, which was running middleware for all 404 requests. This is confusing, causing reports such as #10785 

As part of the work making this change it became clear there was a lot of unused logic in the dev 404 handling. Currently it checks if there is no matching route to handle the 404 case. However there is always a matched route in practice, because `ensure404Route` means a 404 route will always match it, so those code branches are never hit. This PR removes a lot of this unused code.

## Testing

Added new tests, and fixed bugs in other tests.

## Docs

Just fixes a bug, though maybe there could be docs explaining that middleware doesn't run for prerendered 404s?

### DIFF
--- a/.changeset/silly-pens-fetch.md
+++ b/.changeset/silly-pens-fetch.md
@@ -1,0 +1,7 @@
+---
+'astro': patch
+---
+
+Corrects an inconsistency in dev where middleware would run for prerendered 404 routes.
+Middleware is not run for prerendered 404 routes in production, so this was incorrect.
+
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -162,11 +162,10 @@ export async function handleRoute({
 }: HandleRoute): Promise<void> {
 	const timeStart = performance.now();
 	const { config, loader, logger } = pipeline;
-	if (!matchedRoute && !config.i18n) {
-		if (isLoggedRequest(pathname)) {
-			logger.info(null, req({ url: pathname, method: incomingRequest.method, statusCode: 404 }));
-		}
-		return handle404Response(origin, incomingRequest, incomingResponse);
+
+	if (!matchedRoute) {
+		// This should never happen, because ensure404Route will add a 404 route if none exists.
+		throw new Error('No route matched, and default 404 route was not found.');
 	}
 
 	let request: Request;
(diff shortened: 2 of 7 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -177,106 +176,47 @@ export async function handleRoute({
...
+	const isPrerendered404 = matchedRoute.route.route === '/404' && matchedRoute.route.prerender;
+
+	renderContext = RenderContext.create({
+		locals,
+		pipeline,
+		pathname,
+		middleware: isPrerendered404 ? undefined : middleware,
+		request,
+		routeData: route,
+	});
```
ascorbic, line 219: This is the actual changed logic

packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -177,106 +176,47 @@ export async function handleRoute({
 	const middleware = (await loadMiddleware(loader)).onRequest;
 	const locals = Reflect.get(incomingRequest, clientLocalsSymbol);
 
-	if (!matchedRoute) {
```
ascorbic, line 180: `matchedRoute` is always truthy because of `ensure404Route`, so this branch is never hit

.changeset/silly-pens-fetch.md
```diff
@@ -0,0 +1,7 @@
+---
+'astro': patch
+---
+
+Corrects an inconsistency in dev where middleware would run for prerendered 404 routes.
+Middleware is not run for prerendered 404 routes in production, so this was incorrect.
+
```
florian-lefebvre, line 7:
```suggestion
Corrects an inconsistency in development where middlewares would run for prerendered 404 routes. It doesn't happen anymore to match production's behavior
```

ascorbic:
@matthewp yeah, I was a bit unclear there. Middleware is of course run for prerendered pages at build time. This is specific to 404 routes. The issue with prerendered 404 pages is that when we prerender it there's no way of knowing what the actual route will be, so there's no way that middleware can be used for things like i18n redirects.

matthewp (approved):

Issue #11274, opened 2024-06-17
Duplicated brackets are generated in data collections.


syhily:
### Astro Info

```block
Astro                    v4.10.3
Node                     v22.3.0
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/mdx
```

### Describe the Bug

The latest release astro doesn't generate the JSON schema correctly. A duplicated `};` is introduced for each yaml data file. I think this bug is introduced by the https://github.com/withastro/astro/pull/11249

<img width="812" alt="image" src="https://github.com/withastro/astro/assets/1761698/79f37eb7-3f48-4ade-a43e-eb865b7445b3">

### What's the expected result?

Valid schema json.

### Link to Minimal Reproducible Example

https://github.com/syhily/yufan.me

### Participation

- [x] I am willing to submit a pull request for this issue.

syhily:
IIUC, the line in the belowing PR shouldn't be moved into the for-loop. It should be kept in the original place.

https://github.com/withastro/astro/pull/11249/files#diff-7e9b96084af9c6f79604dd4aaba392dc11ebb3e64f5a47f2d5c1ce67cb8abb79R452

markgaze:
Apologies, this was my fault! Like I said on the PR comment, I'll have a look to see if I can add some tests around this because I made an assumption that the tests passing meant everything was fine with the types but that wasn't the case, unfortunately!

PR #11277, opened 2024-06-17
Add a test to ensure that `types.d.ts` is valid

markgaze:
## Changes

In #11249, I inadvertently introduced a bug in the types generation, which has since been fixed but I thought I'd add a test to make sure that the output is valid TypeScript (as mentioned in https://github.com/withastro/astro/issues/11274#issuecomment-2174185752). This PR adds that check to the already existing test that checked the types were generated.

I've moved the offending line of code back closer to the `for` loop (but not inside it!) so that it's easier to see that it's related to that and not the JSON schema generation.

## Testing

I moved the line that broke the types generation back into the `for` loop, ran the tests and it failed. When I undid that change and ran the tests again, it passed 🙂 

## Docs

Does not affect the end user so no docs have been written.


packages/astro/test/astro-sync.test.js
```diff
@@ -1,6 +1,7 @@
 import assert from 'node:assert/strict';
 import * as fs from 'node:fs';
 import { before, describe, it } from 'node:test';
+import ts from 'typescript';
```
alexanderniebuhr, line 4: I think this is fine, since it is only a test, but I'll let other comment, if we want to import the full `typescript` package here.

alexanderniebuhr (approved):
@markgaze Thanks for this contribution. No worries on the introduced bug, I didn't catch it in the review either, but very good to have a test for this now. I commented a little question for others to comment on, but otherwise this looks good to me.

packages/astro/test/astro-sync.test.js
```diff
@@ -62,6 +63,20 @@ const createFixture = () => {
 			const expectedPath = new URL(path, astroFixture.config.root).href;
 			assert.equal(writtenFiles[expectedPath].includes(content), true, error);
 		},
+		thenFileShouldBeValidTypescript(path) {
```
matthewp, line 66: I don't think I understand the naming of this. Did you mean `typesFileShouldBeValidTypeScript`?

packages/astro/test/astro-sync.test.js
Princesseuh: This is fine and is the typical way to import from TS

packages/astro/test/astro-sync.test.js
markgaze: I was trying to keep consistency with the functions above 🙂

packages/astro/src/content/types-generator.ts
```diff
@@ -490,6 +490,7 @@ async function writeContentFiles({
 						);
 					}
 				}
+
```
ematipico, line 493: Unrelated change, can you revert?

ematipico (approved):

packages/astro/src/content/types-generator.ts
markgaze: Will do!

PR #11279, opened 2024-06-18
fix(astro): prevent ESM imports being passed directly to getImage

### LINKED ISSUE #10126: Discrepancy in getImage in dev vs build
### Astro Info

```block
Astro                    v4.3.2
Package Manager          pnpm
Adapter                  none
Integrations             @astrojs/svelte
                         @astrojs/tailwind
```

_No response_

### Describe the Bug

When accidentally passing the imported object to `getImage` (and not as `{ src: ImportedImage }`) then when using `astro dev` we get a `LocalImageUsedWrongly` error. But the image is used without issue and included as expected in the `astro build` output.

When using `getImage` from `astro:assets` and accidentally passing the imported object like this:

```astro
import { getImage } from 'astro:assets';
import MyCoolImage from '@assets/component/image.jpeg';
const optimizedImage = await getImage(MyCoolImage);

<img src={optimizedImage.src} />
```

You will get different behaviour during `astro dev` and `astro build`.

To get an error explaining that the image has to be passed as `{ src: MyCoolImage }` and that it's not an import issue.

https://file.io/hmwoBKrygvT6

### Participation

ascorbic:
## Changes

Currently it is possible to pass an ESM image import directly to `getImage` without any type error. There will be an error at build or runtime, but it is more generic and doesn;t make it clear what the problem is.

This PR includes two changes. First it adds runtime error checks for this specific case. It detects if an ESM image has been passed directly and throws a more helpful error.

Second it uses branded types to prevent the ESM image object being passed. This was previously allowed because an ESM import is the correct shape. This is handled now by adding an internal field to the type returned by the ESM import which isn't permitted by getImage.

Fixes #10126 

## Testing

Adds tests

## Docs

@withastro/maintainers-docs includes a new AstroError

### DIFF
--- a/.changeset/old-walls-report.md
+++ b/.changeset/old-walls-report.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Improves type-checking and error handling to catch case where an image import is passed directly to getImage
+Improves type-checking and error handling to catch case where an image import is passed directly to `getImage()`
--- a/.changeset/old-walls-report.md
+++ b/.changeset/old-walls-report.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Improves type-checking and error handling to catch case where an image import is passed directly to getImage
--- a/packages/astro/test/fixtures/core-image-base/tsconfig.json
+++ b/packages/astro/test/fixtures/core-image-base/tsconfig.json
@@ -1,9 +1,11 @@
 {
-  "extends": "astro/tsconfigs/base",
+  "extends": "astro/tsconfigs/strict",
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "~/assets/*": ["src/assets/*"]
+      "~/assets/*": [
+        "src/assets/*"
+      ]
     },
   }
 }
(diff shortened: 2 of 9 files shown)


packages/astro/test/fixtures/core-image-base/tsconfig.json
```diff
@@ -1,9 +1,11 @@
 {
-  "extends": "astro/tsconfigs/base",
+  "extends": "astro/tsconfigs/strict",
```
ascorbic, line 2: When not strict, the content collection fields are nullable, which makes them invalid for images

matthewp:
This current breaks in dev and build, but just without a type error, am I understanding that right?

.changeset/old-walls-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Improves type-checking and error handling to catch case where an image import is passed directly to getImage
```
sarah11918, line 5:
```suggestion
Improves type-checking and error handling to catch case where an image import is passed directly to `getImage()`
```
Nicely written! 🙌

sarah11918 (commented):
This is some pretty fine changeset writing, @ascorbic ! 🫡

ascorbic:
> This current breaks in dev and build, but just without a type error, am I understanding that right?

Yes, and the runtime/build-time error is confusing because it's saying that `src` cannot be a string.

matthewp (approved):

PR #11280, opened 2024-06-18
fix: allow cookies to be set in rewritten responses

### LINKED ISSUE #11265: Cookie not set when response is rewritten (experimental.rewriting)
### Astro Info

```block
Astro                    v4.10.2
Node                     v20.13.1
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

Using `Astro.cookie.set` does not actually set a cookie for a rewritten request.

A cookie should be set.

https://stackblitz.com/edit/github-xcsrpa?file=src%2Fpages%2F1.astro

### Participation

ascorbic:
## Changes

Ensures that `set-cookie` headers are set when rewriting a request.

Fixes #11265

## Testing

Adds a test case for a rewritten request

## Docs
Just a bugfix

### DIFF
--- a/packages/astro/src/core/cookies/cookies.ts
+++ b/packages/astro/src/core/cookies/cookies.ts
@@ -192,6 +192,19 @@ class AstroCookies implements AstroCookiesInterface {
 	}
 
 	/**
+	 * Merges a new AstroCookies instance into the current instance. Any new cookies
+	 * will be added to the current instance, overwriting any existing cookies with the same name.
+	 */
+	merge(cookies: AstroCookies) {
+		const outgoing = cookies.#outgoing;
+		if (outgoing) {
+			for (const [key, value] of outgoing) {
+				this.#ensureOutgoingMap().set(key, value);
+			}
+		}
+	}
+
+	/**
 	 * Astro.cookies.header() returns an iterator for the cookies that have previously
 	 * been set by either Astro.cookies.set() or Astro.cookies.delete().
 	 * This method is primarily used by adapters to set the header on outgoing responses.
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -27,6 +27,7 @@ import {
 	responseSentSymbol,
 } from './constants.js';
 import { AstroCookies, attachCookiesToResponse } from './cookies/index.js';
+import { getFromResponse } from './cookies/response.js';
 import { AstroError, AstroErrorData } from './errors/index.js';
 import { callMiddleware } from './middleware/callMiddleware.js';
 import { sequence } from './middleware/index.js';
@@ -164,10 +165,9 @@ export class RenderContext {
 							this.routeData
 						);
 
-						if (result.cookies) {
-							for (const cookie of AstroCookies.consume(result.cookies)) {
-								response.headers.append('Set-Cookie', cookie);
-							}
+						const responseCookies = getFromResponse(response);
+						if (result.cookies && responseCookies) {
+							result.cookies?.merge(responseCookies);
 						}
 					} catch (e) {
 						// If there is an error in the page's frontmatter or instantiation of the RenderTemplate fails midway,
--- a/packages/astro/src/core/render-context.ts
+++ b/packages/astro/src/core/render-context.ts
@@ -147,14 +147,17 @@ export class RenderContext {
 					);
 				}
 			}
+			let response: Response;
+
 			switch (this.routeData.type) {
-				case 'endpoint':
-					return renderEndpoint(componentInstance as any, ctx, serverLike, logger);
+				case 'endpoint': {
+					response = await renderEndpoint(componentInstance as any, ctx, serverLike, logger);
+					break;
+				}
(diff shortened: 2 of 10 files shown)


packages/astro/src/core/render-context.ts
```diff
@@ -163,6 +163,12 @@ export class RenderContext {
 							streaming,
 							this.routeData
 						);
+
+						if (result.cookies) {
+							for (const cookie of AstroCookies.consume(result.cookies)) {
+								response.headers.append('Set-Cookie', cookie);
```
matthewp, line 169: I don't think we should append the headers here. They are already appended further up in the response, here: https://github.com/withastro/astro/blob/cb4d07819f0dbdfd94bc4f084edf7720ada01323/packages/astro/src/core/app/index.ts#L346-L350

What I think is happening in this bug is that the `Response` has changed because of rewriting, so the AstroCookies object is not attached to it anymore. So what I think you can do is call [attachCookiesToResponse](https://github.com/withastro/astro/blob/main/packages/astro/src/core/cookies/response.ts#L5-L7) here and then when the cookies are consumed further up the chain they will be appended to the response there.

ascorbic: Looks like they actually need to be merged, so I took a different approach

packages/astro/src/core/cookies/cookies.ts
```diff
@@ -191,6 +191,19 @@ class AstroCookies implements AstroCookiesInterface {
 		}
 	}
 
+	/**
+	 * Merges a new AstroCookies instance into the current instance. Any new cookies
+	 * will be added to the current instance, overwriting any existing cookies with the same name.
+	 */
+	merge(cookies: AstroCookies) {
```
ematipico, line 198: I have mixed feelings about this method, I would prefer to **return** a new instance instead of manipulating the existing one. Essentially, try to use a pure/functional approach and avoid side-effects.

I would called this method `fork` instead

packages/astro/src/core/render-context.ts
```diff
@@ -163,6 +164,11 @@ export class RenderContext {
 							streaming,
 							this.routeData
 						);
+
+						const responseCookies = getFromResponse(response);
+						if (result.cookies && responseCookies) {
+							result.cookies?.merge(responseCookies);
```
ematipico, line 170:
```suggestion
							result.cookies.merge(responseCookies);
```

Not needed I think, due to the previous check

packages/astro/test/fixtures/astro-cookies/src/pages/from.astro
```diff
@@ -0,0 +1,4 @@
+---
+console.log('cookie value: ', Astro.cookies.get('my_cookie')?.value);
```
ematipico, line 2: Should we add another test where in the first page we set a `lorem` value, then in the rewritten page we set a `ipsum` value, and assert that the cookie has `ipsum` as value?

packages/astro/src/core/cookies/cookies.ts
matthewp: Why? In this case there are already 2 AstroCookies instances created (once for each response), what does adding a 3rd accomplish? This is not immutable style code, and this function is not mutating the argument, it's mutating itself.

ematipico: I don't think this is the correct fix. I think the fix should be done here:

https://github.com/withastro/astro/blob/bc3c3296bfd44f6ab7cad38dc47a39355a6af4cb/packages/astro/src/core/render-context.ts#L237

And here

https://github.com/withastro/astro/blob/bc3c3296bfd44f6ab7cad38dc47a39355a6af4cb/packages/astro/src/core/render-context.ts#L432

And the code should be 

```js
this.cookies = this.cookies.fork(this.request)
```

`fork` will return a new instance of cookies, and the creation of these new cookies is confined inside the `rewrite` function, while the proposed solution seems to be more broad

matthewp: Ok, that plan makes sense then. 👍

ascorbic: I don't think will help with cookies that are set on the response. It only merges the request cookies.

ascorbic: I'm happy to change it to a pure function, but I don't think it can be done at the point the cookies are first assigned.

ematipico: What confuses me is the term "rewriting a response" from the initial issue. `Astro.rewrite`, actually, rewrites a `Request`, and you'll get a new `Response`. So I wonder how the issue is related to rewriting? I am a bit confused 😅

ascorbic: It refers to responses to a rewritten request. If page `/a` has `Astro.rewrite('/b')`, and `/b` sets a cookie, currently the cookie from b is not returned.

ematipico (approved):

ascorbic:
@ematipico and I had a chat and agreed this approach

PR #11284, opened 2024-06-19
fix: support HTTP/2 in `astro dev`

### LINKED ISSUE #10238: Using HTTP/2 with astro dev
### Astro Info

```block
Astro                    v4.4.3
Package Manager          npm
Adapter                  none
Integrations             none
```

Chrome

### Describe the Bug

By default, astro is setting `vite.server.proxy` internally, which causes http/2 to be disabled according to the vite [documentation](https://vitejs.dev/config/server-options#server-https). Hence to use http/2 we need to set `proxy: false`. However, this makes it so that `Astro.request.url='https://undefined/'`.

NOTE: The provided minimal reproducible example works on stackblitz, but not locally. To reproduce - download it locally, install deps, run astro dev and check the console logs.

I would like to be able to use http/2 with astro dev.

https://stackblitz.com/edit/github-j1fdfd-tvtekj?file=src%2Fpages%2Findex.astro

### Participation

ascorbic:
## Changes

The Vite dev server suppports HTTP/2 when the `server.https` option is set. However this doesn't work in `astro dev` by default, because we set the `vite.server.proxy` option (to an empty object). If a user manually enables it by unsetting the proxy option, we generate a malformed `Astro.request.url` and SSR pages break because we try to set invalid headers.

This PR removes the default proxy config so it is possible to enable HTTP/2 by setting the `vite.server.https` options. It handles  URL correctly now, by chedking the `:authority` pseudo-header as well as `host`, and ensures that we don't attempt to create invalid headers when we copy the headers object.

Fixes #10238

## Testing
It adds a test fixture based on the repro in #10238 (thanks @alexvuka1). In order to do this, I also added support for TLS and HTTP/2 in our test utils. If the fixture's Astro config enables https, the test utils configure fetch to trust self-signed certs and enable HTTP/2 in our tests.

## Docs
I don't think this needs to be documented


packages/astro/test/test-utils.js
```diff
@@ -171,6 +177,17 @@ export async function loadFixture(inlineConfig) {
...
 		resolveUrl,
 		fetch: async (url, init) => {
+			if (config.vite?.server?.https) {
+				init = {
+					dispatcher: new Agent({
+						connect: {
+							rejectUnauthorized: false,
+						},
+						allowH2: true,
+					}),
```
ematipico, line 187: What's this for? I have no knowledge of this

ematipico (approved):

packages/astro/src/core/request.ts
```diff
@@ -50,7 +50,10 @@ export function createRequest({
 		? undefined
 		: headers instanceof Headers
 			? headers
-			: new Headers(Object.entries(headers as Record<string, any>));
+			: new Headers(
+					// Filter out H2 pseudo-headers, as these can't be created with Headers
```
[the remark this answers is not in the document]
ascorbic: They're fake headers that are added to all HTTP/2 requests. They're only allowed to be generated internally though: it's an error to add one to Headers yourself. https://httpwg.org/specs/rfc7540.html#HttpRequest

packages/astro/test/test-utils.js
ascorbic: This tells undici fetch to use a custom dispatcher. This lets you customise some of the HTTP options for the request. `rejectUnauthorized` lets you ignore invalid server certificates, which is needed here as the dev server will always use self-signed certs. `allowH2` enables HTTP/2 requests in fetch, if the server supports it.

ascorbic: I'll add comments for both of these

ascorbic:
@ematipico thanks for the review. I've added comments answering both of your questions

PR #11286, opened 2024-06-19
fix: warn for missing head

ascorbic:
## Changes

This PR warns if a page has either hoisted scripts or inline styles, but does not have a head element. This can cause bugs and is probably a mistake.


ematipico:
> For historical reasons we don't require it

I'm curious now 🤣 what are the reasons?

PR #11289, opened 2024-06-19
unify preview server config

oe:
## Changes
Merge preview server config by:
```js
const previewConfig = Object.assign({},
  settings.config.vite.server,
  settings.config.vite.preview,
  settings.config.server,
)
```

Make the preview server can access all server config from `vite.server` / `vite.preview` / `server`.

Especially, this make preview server access `proxy` settings from these objects which allow preview proxy api request just like dev server!

## Testing
1. add proxy config to `astro.config.ts`
2. build your project
3. run `astro preview` and check network

## Docs
This PR adds new feature to `astro preview`, docs should be updated


.changeset/famous-dancers-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': major
```
ematipico, line 2: Why is this a `major`?

.changeset/famous-dancers-tan.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+unify preview server config
```
ematipico, line 5: Can you leave a better explanation? Developers will read the changelog, and I doubt they will understand what "unify" means

ematipico:
> This PR adds new feature to astro preview, docs should be updated

Developers are usually in charge of updating the docs. Could you send a PR to update the docs? Or at least, which docs should be updated?

packages/astro/src/core/preview/index.ts
```diff
@@ -66,16 +66,19 @@ export default async function preview(inlineConfig: AstroInlineConfig): Promise<
 	if (typeof previewModule.default !== 'function') {
 		throw new Error(`[preview] ${settings.adapter.name} cannot preview your app.`);
 	}
-
+	const previewConfig = Object.assign({},
+		settings.config.vite.server,
+		settings.config.vite.preview,
+		settings.config.server,
+	)
```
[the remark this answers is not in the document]
oe: According to the [official Vite documentation](https://vitejs.dev/config/preview-options.html),  `vite preview` uses the Vite Server configuration by default. Astro is built on Vite, but doesn't use Vite's Server configuration, which is very strange. (The proxy configuration in Dev and Preview is very useful for this purpose.

Anyway, the above code just keeps Astro Preview behavior consistent with Vite Preview, and is technically a bug fix.

oe:
> > This PR adds new feature to astro preview, docs should be updated
> 
> Developers are usually in charge of updating the docs. Could you send a PR to update the docs? Or at least, which docs should be updated?

The [Server Options](https://docs.astro.build/en/reference/configuration-reference/#server-options) may need update according to [Vite Preview Options](https://vitejs.dev/config/preview-options.html)

.changeset/famous-dancers-tan.md
oe: I don't know, after running `pnpm exec changeset`, `major` is the only option left

.changeset/famous-dancers-tan.md
oe: A more accurate description would be to keep Astro Preview behaving the same way as [Vite Preview](https://vitejs.dev/config/preview-options.html), supporting Vite Preview settings.

According to the Astro code, static previews are supported by default, but third-party adapter previews require additional support.

packages/astro/src/core/preview/index.ts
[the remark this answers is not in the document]
bluwy: Astro preview does use the Vite preview server internally, but we don't document that the Vite server/preview options will strictly work when configured. Editing this part of the code is also sensitive because you're passing all the Vite-specific server/preview options to `previewModule.default` directly.

We're currently intentionally limiting the options here because it would be expected for all adapters to support the same set of features, but not all adapters use the Vite preview server. That way, the adapters need to only concern of supporting features from `config.server`.

---

If this is about the `proxy` config, it leans towards a feature request, not a bug fix. I could see perhaps passing that only to `previewModule.default` and update the types to be optional, and adapters could decide to support it. But I'd like to hear what others think of this.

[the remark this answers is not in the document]
oe: Got it. What I need is `proxy`, it can be approached by configuring vite.server on astro dev, but doesn't work on astro preview.

florian-lefebvre:
@oe are you still interested in this PR?

oe:
> @oe are you still interested in this PR?

Yes, what do I need to do?

florian-lefebvre:
@bluwy do you have advices?

florian-lefebvre:
Looks like this is a feature request, you can create a proposal at https://github.com/withastro/roadmap/discussions

PR #11292, opened 2024-06-19
bug: removes type override from defineAction

jdtjenkins:
## Changes

This PR removes the type override for the `accept` parameter in the `defineAction` Actions function.

Fernando Herrera pointed out earlier that the `accept` param for `defineAction` only suggested `json` in the intellisense instead of both `json` and `form`.

Now I'm not going to pretend I understand _everything_ going on in the typing for the `accept` param here. But I'm pretty sure the `= 'json'` here is redundant.

https://discord.com/channels/830184174198718474/1252972140161142816/1252996832238506056

Before:
image

After:
image

## Testing

No tests are necessary here I think?

## Docs

No docs are necessary as I think this is a bug fix?

### DIFF
--- a/.changeset/quiet-doors-eat.md
+++ b/.changeset/quiet-doors-eat.md
@@ -2,4 +2,4 @@
 "astro": patch
 ---
 
-Removes type override from defineAction accept paramter
+Fixes a case where `defineAction` autocomplete for the `accept` prop would not show `"form"` as a possible value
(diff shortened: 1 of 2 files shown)


ematipico (approved):

.changeset/quiet-doors-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"astro": patch
+---
+
+Removes type override from defineAction accept paramter
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where `defineAction` autocomplete for the `accept` prop would not show `"form"` as a possible value
```

florian-lefebvre (approved):

PR #11295, opened 2024-06-20
ci: don't build inside the linting job

ematipico:
## Changes

We were building the project in the linting job, but we don't need to build the project for linting the source code.

## Testing

The CI should pass, and we should save around 1m each time we run the linting step

## Docs

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,9 +91,6 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
-      - name: Build Packages
-        run: pnpm run build
-
       - name: Lint
         run: pnpm run lint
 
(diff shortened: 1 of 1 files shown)


bluwy:
It was added (and explained) at https://github.com/withastro/astro/pull/7425#discussion_r1234707203, we need it for typescript-eslint to typecheck or they become `any`?

ematipico:
I wish there were a comment explaning it... Well, at least should should reduce the builds. The lint job `needs` build, so it's safe to remove it anyway?

bluwy:
I always use git blame for things like this 😄 I think the `needs: build` is different though, it specifies that it needs the build to complete first before executing this job, not that it would reuse the build environment, so I think we still need the build here. If turbopack kicks in for our PRs, the build time should also be fast.

PR #11296, opened 2024-06-20
fix: astro:env getSecret

florian-lefebvre:
## Changes

- Uses the rollup trick

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/integrations/node/src/index.ts
+++ b/packages/integrations/node/src/index.ts
@@ -42,6 +42,11 @@ export default function createIntegration(userOptions: UserOptions): AstroIntegr
 						ssr: {
 							noExternal: ['@astrojs/node'],
 						},
+						build: {
+							rollupOptions: {
+								external: ['astro/env/setup'],
+							},
+						},
 					},
 				});
 			},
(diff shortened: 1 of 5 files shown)


packages/integrations/vercel/src/serverless/adapter.ts
```diff
@@ -265,6 +265,11 @@ export default function vercelServerless({
 						ssr: {
 							external: ['@vercel/nft'],
 						},
+						build: {
+							rollupOptions: {
+								external: ['astro/env/setup'],
+							},
+						},
```
bluwy, line 272: Questions:
1. Could we put this in `ssr.external`? (I think they work the same)
2. Could we only add this if `astro/env/setup` exists? We can try importing it here, then decide adding the config.

packages/integrations/node/src/index.ts
```diff
@@ -24,6 +24,16 @@ export function getAdapter(options: Options): AstroAdapter {
 	};
 }
 
+// TODO: remove once we don't use a TLA anymore
+async function shouldExternalizeAstroEnvSetup() {
+	try {
+		await import('astro/env/setup');
+		return false;
+	} catch {
+		return true;
```
florian-lefebvre, line 33: @bluwy I think we need to externalize only if it can't be imported?

bluwy: Yup 👍

ematipico (approved):

bluwy (approved):

PR #7502, opened 2023-06-27
Don’t post Discord congrats for auto-formatter commits

delucis:
## Changes

Adds a condition to the Discord congrats bot action to not post if the commit message matches the auto-formatter commit message of `[ci] format`.

Should help reduce some unnecessary noise:

image

## Testing

We do something like this in docs already.

## Docs

N/A.

PR #11298, opened 2024-06-20
Update CompressHTML Whitespace issue #7502

Dillonpw:
## Changes
Clarification to the docs in regards to compressHTML top-level option

## Docs

Clarification to the docs in regards to the compressHTML option
https://docs.astro.build/en/reference/configuration-reference/#compresshtml


packages/astro/src/@types/astro.ts
```diff
@@ -750,8 +750,11 @@ export interface AstroUserConfig {
 	 * @type {boolean}
 	 * @default `true`
 	 * @description
-	 * This is an option to minify your HTML output and reduce the size of your HTML files. By default, Astro removes all whitespace from your HTML, including line breaks, from `.astro` components. This occurs both in development mode and in the final build.
-	 * To disable HTML compression, set the `compressHTML` flag to `false`.
+	 *
+	 * This is an option to minify your HTML output and reduce the size of your HTML files, occuring both in development mode and in the final build. 
+	 * By default, Astro removes whitespace from your HTML, including line breaks, from .astro components in a lossless manner.
+	 * It is important to note that some whitespace may be kept to preserve the visual rendering of your HTML.  
+	 * To disable HTML compression, set the compressHTML flag to false.
```
sarah11918, line 757:
```suggestion
	 * This is an option to minify your HTML output and reduce the size of your HTML files.
	 *
	 * By default, Astro removes whitespace from your HTML, including line breaks, from .astro components in a lossless manner. Some whitespace may be kept as needed to preserve the visual rendering of your HTML. This occurs both in development mode and in the final build.
	 *
	 * To disable HTML compression, set the compressHTML flag to false.
```
While we're here, a good chance to add some nicer formatting to an entry that hasn't been updated in a while! I always like separating out the definition at the top, and a statement like "To disable..." so they get their own paragraphs!

sarah11918 (commented):
Thank you for grabbing this docs issue, @Dillonpw !

Just made a tiny edit for readability, but looking forward to having this explained in the docs!

PR #11302, opened 2024-06-20
Fall back to page reload when redirected to a cross origin URL

### LINKED ISSUE #11301: View transition router fails for cross origin redirects
### Astro Info

```block
Astro                    v4.11.0
Node                     v18.20.0
Adapter                  none
Integrations             astro-d2-integration
                         astro-vtbot
                         @astrojs/starlight
                         @astrojs/react
```

_No response_

### Describe the Bug

From Discord thread https://discord.com/channels/830184174198718474/1253123090045599745
mcomsOP
 — 
I think Astro View Transitions don't expect this case and don't bail correctly, causing the URL in the browser to end up broken. See browser console log below. There's a problem when I have configured certain pages on my site to redirect to a new domain. So, when a user is on domain1.com/ and clicks a link to domain1.com/page, they get redirected to domain2.com/page instead. This errors out in Astro View Transitions due to the domain handoff, and interrupts any URL updates. So the browser URL bar stays at domain1.com/, and navigation history does not update. The user can navigate normally unless they refresh or click the back button.

history update not broken, no errors.

see new test case in PR

### Participation

martrapp:
## Changes

Fixes #11301 
when redirected to a different origin, interrupt loading and do a full page reload the redirected URL 

## Testing

Added new e2e testcase

## Docs

n.a.

### DIFF
--- a/.changeset/small-badgers-juggle.md
+++ b/.changeset/small-badgers-juggle.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes an issue with the view transition router when redirecting to an URL with different origin.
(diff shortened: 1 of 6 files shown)


matthewp (approved):
🙌

PR #11304, opened 2024-06-20
feat: Refactor integration hooks type

Fryuni:
### `astro`

- Extract the `AstroIntegration['hooks']` type into a global `Astro.IntegrationHooks` interface to allow for third-party extensions.

As discussed on Discord, this breaks only a very narrow use case that wasn't documented as being supported and relied on the hooks not being extensible in order to create an extension workaround on top of it. So it is not being considered as a breaking change.

### `@astrojs/db`

- Remove the `AstroDbIntegration` type (BREAKING)
- Kept the `defineDbIntegration` just so people don't have to change their code for that, it is no longer needed

## Testing

This doesn't change any runtime behavior

## Docs

We can now document how integrations can add their own hooks 😄

## Future scope

This PR only allows integrations to declare the types for their hooks; it doesn't provide any higher-level mechanism to trigger those hooks. Triggering them is already possible with the following code:
```ts
export default myIntegration: (): AstroIntegration => {
  let integrations: AstroIntegration[] = [];

  return {
    name: 'myLib',
    hooks: {
	  'astro:config:setup': ({config}) => {
	    integrations = config.integrations;
	  },
	  // Wherever some other logic runs
	  'astro:build:done': () => {
	    for (const integration of integrations) {
	      integrations.hooks['myLib:eventHappened']?.(hookParameters);
	    }
	  },
    },
  };
}
```

In the future, we might provide some way for integrations to trigger their hooks with less boilerplate


packages/astro/src/@types/astro.ts
```diff
@@ -3012,83 +3012,92 @@ export type HookParameters<
 	Fn = AstroIntegration['hooks'][Hook],
 > = Fn extends (...args: any) => any ? Parameters<Fn>[0] : never;
 
+declare global {
+	// eslint-disable-next-line  @typescript-eslint/no-namespace
+ 	namespace Astro {
```
florian-lefebvre, line 3017: What is the advantage of a namespace over a module declaration?

Fryuni: Two:
- Makes a distinction between types we are intentionally allowing third-party libraries to extend vs types that just happen to be interfaces at the root level and, as such, are extendable.
- We already use namespaces for this one the `Astro.locals`, though there it uses the `App` namespace instead of the `Astro` namespace.

florian-lefebvre: Would be nice to use the same namespace everywhere, I'd be curious to know why it was named App

Fryuni: I suppose so it wouldn't be ambiguous on `.astro` files when using the compiler-injected constant vs the namespace

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,18 @@
+---
+'astro': minor
+'@astrojs/db': minor
+---
+
+Refactor type for integration hooks to be extensible from an interface on the global `Astro` namespace.
+
+To declare your own hooks for integrations extend the interface like so:
```
ematipico, line 8: We should tell users **where** this snippet should be pasted/used

packages/astro/src/@types/astro.ts
ematipico: I believe it makes sense to separate `Astro` and `App`. `App` for end-users, while `Astro` works across all Astro projects that use a particular integration. 

In fact. `Locals` change **per project**. However, `IntegrationHooks` change **per integration**, which means multiple projects will have the same type.

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,18 @@
+---
+'astro': minor
+'@astrojs/db': minor
```
ematipico, line 3: We should create a new changeset only for `@astrojs/db` that shows the migration path of the new type. Since this is a breaking change, I believe this should stand out in the changelog

sarah11918: Just noting that I'll review here when I have a docs PR!

sarah11918 (changes requested):
Just leaving a placeholder for docs! When I have a docs PR submitted, I can review it and this!

.changeset/curvy-otters-jog.md
Fryuni: It is barely even a breaking change for `@astrojs/db` though. It breaks something we never said was there (the `AstroDbIntegration` type). The helper `defineDbIntegration` method [shown on the docs](https://docs.astro.build/en/guides/astro-db/#building-astro-db-integrations) continues to work with no changes.

The main difference is requiring the new version of `astro`.

But I'll write that up either way

.changeset/blue-colts-film.md
```diff
@@ -0,0 +1,23 @@
+---
+'@astrojs/db': minor
+---
+
+Remove `AstroDbIntegration` type
```
sarah11918, line 5: Just pointing out that it looks like docs will need updating here for this, too? (I currently have a docs PR for the change to astro core!)

https://docs.astro.build/en/guides/astro-db/#building-astro-db-integrations

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,56 @@
+---
+'astro': minor
+---
+
+Refactor type for integration hooks to be extensible from an interface on the global `Astro` namespace.
```
sarah11918, line 5:
```suggestion
Adds support for writing custom integration hooks to allow runtime interactions between your integration and other integrations.
```
Something of this style maybe, to more explicitly describe the use case? (Vs the change to the code)

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,56 @@
...
+
+To declare your own hooks for integrations extend the interface by adding the following code in your integration:
```
sarah11918, line 7:
```suggestion
To declare your own hooks for your integration, extend the `Astro.IntegrationHooks` interface:
```
nit: we only say "add the following code" if it's something they should actually copy and paste. This is a generic model of the shape of code to provide, and "adding" it wouldn't help them.

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,56 @@
...
+declare global {
+  namespace Astro {
+    interface IntegrationHooks {
+      'myLib:eventHappened': (your: string, parameters: number) => Promise<void>;
+    }
+  }
+}
+```
+
+Call your hooks on all integrations when you want to trigger them:
```
sarah11918, line 20:
```suggestion
Call your hooks on all other integrations installed in a project at the appropriate time. For example, you can call your hook on initialization before before either the Vite or Astro config have resolved:
```

.changeset/curvy-otters-jog.md
```diff
@@ -0,0 +1,56 @@
...
+        for (const integration of config.integrations) {
+          await integration.hooks['myLib:eventHappened'].?('your values', 123);
+        }
+      },
+    }
+  }
+}
+```
+
+Other integrations can now declare your hooks:
```
sarah11918, line 40:
```suggestion
Other integrations can also now declare your hooks:
```

sarah11918 (commented):
Quick review for docs! Looks great, I am mostly suggesting a more "use case" description to kick this off!

Additionally, there is another change in here to `@astrojs/db` that will need a corresponding docs update, as this type being removed is currently documented.

Fryuni:
> Additionally, there is another change in here to @astrojs/db that will need a corresponding docs update, as this type being removed is currently documented.

It is not documented Sarah, `AstroDbIntegration` is nowhere in the docs. The `defineDbIntegration` utility is documented and that is why I kept it around.

.changeset/blue-colts-film.md
florian-lefebvre, line 5:
```suggestion
Removes the `AstroDbIntegration` type
```

.changeset/curvy-otters-jog.md
sarah11918: Coming back to this again, and thinking that because the change primarily is a refactor, it would be nice to assure people that this will not break existing code. So I think we can get both a use case and an implementation change in here. What do you think? (I updated my suggestion

packages/db/src/core/load-file.ts
```diff
@@ -31,7 +28,6 @@ export async function resolveDbConfig({
 	const integrationDbConfigPaths: Array<{ name: string; configEntrypoint: string | URL }> = [];
 	const integrationSeedPaths: Array<string | URL> = [];
 	for (const integration of integrations) {
-		if (!isDbIntegration(integration)) continue;
```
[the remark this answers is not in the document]
Fryuni: It was needed as a type guard to turn `integration` from `AstroIntegration` to `AstroDbIntegration`. That guard is no longer needed since it is a single augmented type now.

.changeset/blue-colts-film.md
```diff
@@ -0,0 +1,23 @@
...
+
+Astro integration hooks can now be extended and as such `@astrojs/db` no longer needs to declare it's own integration type. Using `AstroIntegration` will have the same type.
+
+If you were using the `AstroDbIntegration` type, apply this change to your integration code:
+
+```diff
+- import { defineDbIntegration, type AstroDbIntegration } from '@astrojs/db/utils';
++ import { defineDbIntegration } from '@astrojs/db/utils';
```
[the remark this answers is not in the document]
Fryuni: I left it for backwards compatibility. No migration steps are needed for anyone who didn't look under the hood and just followed the docs.

Since this is for integration authors making integrations for a still beta platform, I'm fine with removing it completely (those are power users that should be fine with a small migration)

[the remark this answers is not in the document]
florian-lefebvre: Got it, I'll ask core about it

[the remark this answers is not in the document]
florian-lefebvre: Let's keep it for now, removal can happen in a later PR

sarah11918:
@Fryuni Sorry, I thought I checked closely, but I was concerned about this example which you're right, is not the type itself. As long as this example doesn't need updating, then no docs are needed, you're correct!

image

sarah11918:
Just pinging @Fryuni to look at my changeset suggestions! The Docs PR has already been approved and is ready for next week, so this is just tiny details now!  :raised_hands:

packages/db/src/core/types.ts
```diff
@@ -88,13 +87,16 @@ interface LegacyIndexConfig<TColumns extends ColumnsConfig>
...
-		'astro:db:setup'?: (options: {
-			extendDb: (options: {
-				configEntrypoint?: URL | string;
-				seedEntrypoint?: URL | string;
-			}) => void;
-		}) => void | Promise<void>;
-	};
-};
+declare global {
+	// eslint-disable-next-line  @typescript-eslint/no-namespace
```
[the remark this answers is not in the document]
Fryuni: Maybe...
That would be a separate discussion since it would disable it for the whole repo. Not the scope of this PR

sarah11918 (approved):
Approving for docs!

matthewp (approved):

PR #11307, opened 2024-06-21
refactor(plugin-analyzer): code refactor

Simon-He95:
## Changes

This PR extracts the data that is not related to the variable in the cycle to the outer layer. This can optimize the conditional judgment and repeated data reading of the multiple times in the cycle

### DIFF
--- a/.changeset/kind-dodos-love.md
+++ b/.changeset/kind-dodos-love.md
@@ -1,5 +1,0 @@
----
-'astro': minor
----
-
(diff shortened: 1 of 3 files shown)


.changeset/kind-dodos-love.md
bluwy, line 1: I think we can remove this changeset as the changes doesn't fix a bug or introduce a new feature. Great work on the optimization though 👍

Simon-He95: ok :)

bluwy (approved):

ematipico (approved):

Issue #11306, opened 2024-06-20
Rewriting doesn't use HTTP status code from rewritten response


vchirikov:
### Astro Info

```block
4.11.0
```

### Describe the Bug

If we rewrite from non-existed page we always get 404 status even if middleware rewrite to the valid page.

Looks like the response object from rewriting isn't used. This is the repro:

![bug](https://github.com/withastro/astro/assets/40987940/9109b9a0-d081-42ec-becf-e007e7c3f4e1)

we should get http status 200 (because we rewrite to the /about page), but we see 404.

p.s. The issue was originally described [here](https://github.com/withastro/astro/issues/11235#issuecomment-2163478342), this is separate issue for the bug with repro.
p.s. looks similar to the [cookie issue](https://github.com/withastro/astro/issues/11265) (fixed by @ascorbic )

### What's the expected result?

rewriting should use status code from rewritten page pipeline.

### Link to Minimal Reproducible Example

https://github.com/vchirikov/astro_issue_02

ematipico:
@vchirikov do you experience the same issue with `output: "server"` (SSR)?

vchirikov:
Yes, btw the repro uses `output: "server"`:

https://github.com/vchirikov/astro_issue_02/blob/f7c9532f3e31798f6acbbae748c076fe9fe185d9/astro.config.mjs#L7

vchirikov:
To be clear, the issue isn't about `/404` rewriting, it's about wrong `404` status from rewritten page (`/about` in the gif), which should return 200, but returned 404 instead...

vchirikov:
> could you reduce the branching of the middleware in your repro?

done in https://github.com/vchirikov/astro_issue_02/tree/issue-11306

> Can you be more clear, please?

As a developer I expect that successfully rewritten page should return status 200 instead of 404:

![image](https://github.com/withastro/astro/assets/40987940/5d958b4f-b6bb-4bc3-a966-bf31eef55d43)

ematipico:
> As a developer I expect that successfully rewritten page should return status 200 instead of 404:

You haven't told me why though, because in this case you're intentionally rewriting with special route 404, which always returns a status code 404. Why should it return a 200? IMHO, it should not.

vchirikov:
I rewrite all routes to `next('/')` (see middleware.ts on the screenshot), I don't get your `you're intentionally rewriting with special route 404, which always returns a status code 404`

[...]

PR #11308, opened 2024-06-21
fix(routing): return correct status code for `500.astro` and `404.astro`

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11306

This PR hardcodes the status code of the rendered pages `404.astro` and `500.astro`.

This issue was never evident because users were never able to see/catch the response emitted by these custom pages, but with the rewriting this was evident.

I updated the `renderPage` function to set a different `status` in case we attempt to render these particular routes.

## Testing

Testing here is particularly difficult, because we are inside a rewriting cycle.

I locally tested that in `const response = await next()` the response holds a `404` or a `500` depending on where we're redirecting. 

## Docs

N/A


packages/astro/src/runtime/server/render/page.ts
```diff
@@ -85,6 +85,18 @@ export async function renderPage(
 	if (route?.component.endsWith('.md')) {
 		headers.set('Content-Type', 'text/html; charset=utf-8');
 	}
-	const response = new Response(body, { ...init, headers });
-	return response;
+	let status = init.status;
+	// Custom 404.astro and 500.astro are particular routes that must return a fixed status code
+	// The check uses the `/` to avoid possible false negatives such as custom_404.astro
+	if (route?.component.endsWith("/404.astro")) {
```
florian-lefebvre, line 91: What if the rewrite aims at `src/pages/foo/404.astro`? This would be also matched but shouln't right?

ematipico: I thought about it. I was thinking to pass `config.srcDir`, or check `pages/404.astro`. Which one should I go for?

matthewp: What are the valid values of `route.component`? It's doesn't include `src/pages` does it?

ematipico: They can differ, especially if the user changes the `srcDir` to something different from `src/`. However, it's safe to assume that these files are at the root for `pages/`, correct? If so, the value of `route.component` should always end with `pages/404.astro`

florian-lefebvre: Maybe not necessarily because of integrations?

ematipico: I am not sure what you mean, can you expand what you're talking about?

florian-lefebvre: I mean what if an integration does that?
```ts
injectRoute({
	pattern: '/404'
	entrypoint: './foo/Bar.astro'
})
```
Is `route.component` computed based on the entrypoint?

matthewp: @ematipico I don't think that's a safe assumption, you can have `src/pages/blog/pages/404.astro`, a `pages` folder as part of the URL.

ematipico: So what do guys suggest? Passing `srcDir` and concat it with `pages/404.astro`

ematipico: I updated the code to use `route` instead

florian-lefebvre (approved):

PR #11312, opened 2024-06-21
refactor(plugin-content): code refactor

Simon-He95:
## Changes

Concurrent operation to copy files


ematipico:
Can you explain to use this refactor a bit more? What does it change? Why is it needed? etc.

Please argument your decisions

Simon-He95:
> Can you explain to us this refactor a bit more? What does it change? Why is it needed? etc.
> 
> Please argument your decisions

I made some adjustments and waited until the promises had values, because there was no order or dependency in copying files, and I thought parallel processing would be more efficient.

bluwy (approved):

ematipico (approved):

PR #11323, opened 2024-06-24
chore: extract fs helpers into shared internal-helpers package

ascorbic:
## Changes

The `@astrojs/vercel` package includes some filesystem helpers that would be useful in other packages. This PR moves them into the `internal-helpers` package.

## Testing

There should be no change in behaviour


ematipico (commented):
Does the vercel integration have the `@astrojs/internal-helpers` as dependency?

ascorbic:
> Does the vercel integration have the `@astrojs/internal-helpers` as dependency?

Yup

ematipico (approved):

Issue #11128, opened 2024-05-23
Container API usage within an Astro project fails to build when using pnpm


TheOtterlord:
### Astro Info

```block
Astro                    v4.9.0
Node                     v20.12.2
System                   Linux (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             none
```

### Describe the Bug

Building a project using the container API fails with the following stacktrace:
```
14:19:09 [ERROR] [vite] x Build failed in 3.64s
[vite]: Rollup failed to resolve import "fsevents" from "/home/otterlord/Documents/Repos/container-repro/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/node-entry.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
  Stack trace:
    at viteWarn (file:///home/otterlord/Documents/Repos/container-repro/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:67612:27)
    at onwarn (file:///home/otterlord/Documents/Repos/container-repro/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:67356:13)
    at Object.logger [as onLog] (file:///home/otterlord/Documents/Repos/container-repro/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/node-entry.js:20162:9)
    at ModuleLoader.resolveDynamicImport (file:///home/otterlord/Documents/Repos/container-repro/node_modules/.pnpm/rollup@4.18.0/node_modules/rollup/dist/es/shared/node-entry.js:19162:58)
```
This only occurs when using pnpm. Using npm the project builds correctly.

**NOTE**: The GitHub repo link is intentional, as you'll need to use gitpod/local or something that actually uses pnpm. Stackblitz works fine as it uses something else iirc.

### What's the expected result?

Builds correctly when using pnpm

### Link to Minimal Reproducible Example

https://github.com/TheOtterlord/container-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

1v3n:
+1

ematipico:
@1v3n please use the reactions to upvote the issue. No need to add comment if it doesn't any additional information

PR #11327, opened 2024-06-24
fix(container): breakdown files to avoid importing `vite`

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11128

The container was indirectly pulling and importing some modules that were importing `vite`.

This PR attempts to breakdown the files in order to not import `vite` inside `astro/container`

## Testing

I tested locally the reproduction provided.

## Docs

N/A

### DIFF
--- a/packages/astro/src/container/index.ts
+++ b/packages/astro/src/container/index.ts
@@ -329,10 +329,10 @@ export class experimental_AstroContainer {
 	private static async createFromManifest(
 		manifest: SSRManifest
 	): Promise<experimental_AstroContainer> {
-		// const astroConfig = await validateConfig(ASTRO_CONFIG_DEFAULTS, process.cwd(), 'container');
+		const astroConfig = await validateConfig(ASTRO_CONFIG_DEFAULTS, process.cwd(), 'container');
 		const container = new experimental_AstroContainer({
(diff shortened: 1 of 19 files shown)


packages/astro/src/container/index.ts
```diff
@@ -329,10 +329,10 @@ export class experimental_AstroContainer {
 	private static async createFromManifest(
 		manifest: SSRManifest
 	): Promise<experimental_AstroContainer> {
-		const astroConfig = await validateConfig(ASTRO_CONFIG_DEFAULTS, process.cwd(), 'container');
+		// const astroConfig = await validateConfig(ASTRO_CONFIG_DEFAULTS, process.cwd(), 'container');
```
matthewp, line 332: if this code isn't needed anymore than we should remove it rather than comment it out

ematipico: Good catch, I was looking for it. I have to re-enable the code

bluwy (approved):
This fix seem ok to me, but it would be nice in the future if the container API code (or shared code) is more isolated into its own directory so we can be careful not to add imports that would break the container API environment bundling, as we also don't have tests for this.

PR #11331, opened 2024-06-25
Remove resolve package and refactor db & studio exports

bluwy:
## Changes

We can use `require.resolve()` instead of the `resolve` package for resolving.

Also simplifies the resolving of `@astrojs/db` by hoisting it up so it can be resolved from `packages/astro/dist/cli/install-package.js`. While we could also install `@astrojs/db` as a devdep in `packages/astro/package.json` to fix it, it would however create a cyclic workspace dependency between `astro`, `@astrojs/db`, and `@astrojs/studio`.

In order to simplify the code, I had to tweak the exports condition of `@astrojs/db` and `@astrojs/studio` to allow importing ESM from CJS (by `require.resolve`). This would also unblock a future Node.js feature where you could import ESM from CJS as long as there's no top-level await. In practice, I also find that it's best practice to use the `import` condition only if you have a corresponding `require` condition.

## Testing

Tested manually in an external project and in one of the fixtures.

## Docs

n/a. refactor only. I added a changeset as it removes a dependency (and would like to document this for users)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -4,11 +4,11 @@ link-workspace-packages=true
 save-workspace-protocol=false # This prevents the examples to have the `workspace:` prefix
 auto-install-peers=false
 
-# `github-slugger` is used by `vite-plugin-markdown-legacy`.
-# Temporarily hoist this until we remove the feature.
-public-hoist-pattern[]=github-slugger
 # Vite's esbuild optimizer has trouble optimizing `@astrojs/lit/client-shim.js`
 # which imports this dependency.
 public-hoist-pattern[]=@webcomponents/template-shadowroot
 # There's a lit dependency duplication somewhere causing multiple Lit versions error.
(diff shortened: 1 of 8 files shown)


.npmrc
```diff
@@ -4,11 +4,11 @@ link-workspace-packages=true
 save-workspace-protocol=false # This prevents the examples to have the `workspace:` prefix
 auto-install-peers=false
 
-# `github-slugger` is used by `vite-plugin-markdown-legacy`.
-# Temporarily hoist this until we remove the feature.
-public-hoist-pattern[]=github-slugger
```
bluwy, line 9: This is no longer needed as we had removed the `vite-plugin-markdown-legacy` plugin

ematipico (approved):

florian-lefebvre (approved):

PR #11334, opened 2024-06-25
feat: add Content Layer loader

ascorbic:
This is the work-in-progress branch implementing the [Content Layer RFC](https://github.com/withastro/roadmap/issues/946).

## Testing

There is an example site in [the content-layer fixture](https://github.com/withastro/astro/tree/content-layer-loader/packages/astro/test/fixtures/content-layer), which shows the use of a custom loader and the `file()` loader. Currently the API for end user looks like this:

```typescript
import { defineCollection, file, z } from 'astro:content';
import { loader } from '../loaders/post-loader.js';

	type: "experimental_data",
	loader: loader({ url: "https://jsonplaceholder.typicode.com/posts" }),
});

const dogs = defineCollection({
	type: "experimental_data",
	// The built in `file` loader loads multiple items from a JSON file
	loader: file("_data/dogs.json"),
	schema: z.object({
		breed: z.string(),
		id: z.string(),
		size: z.string(),
		origin: z.string(),
		lifespan: z.string(),
		temperament: z.array(z.string())
	}),
})
export const collections = { blog, dogs };
```

The `blog` collection shows a custom loader. These can be defined like this:

```typescript
import type { Loader } from 'astro:content';

export interface PostLoaderConfig {
	url: string;
}

export function loader(config:PostLoaderConfig): Loader {
	return {
		name: "post-loader",
		load: async ({
			store, meta, logger
		}) => {
			logger.info('Loading posts');

			const lastSynced = meta.get('lastSynced');

			// Don't sync more than once a minute
			if (lastSynced && (Date.now() - Number(lastSynced) < 1000 * 60)) {
					logger.info('Skipping sync');
					return;
			}

			const posts = await fetch(config.url)
				.then((res) => res.json());

			store.clear();

			for (const post of posts) {
				store.set(post.id, post);
			}
			meta.set('lastSynced', String(Date.now()));
		},
	};
}
```

The API is currently defined like this:

```typescript

export interface Loader<S extends ZodSchema = ZodSchema> {
	/** Unique name of the loader, e.g. the npm package name */
	name: string;
	/** Do the actual loading of the data */
	load: (context: LoaderContext) => Promise<void>;
	/** Optionally, define the schema of the data. Overridden by user-defined schema */
	schema?: S | Promise<S> | (() => S | Promise<S>);
	// Not yet implemented
	render?: (entry: any, options: any) => any;
}

export interface LoaderContext {
	collection: string;
	/** A database abstraction to store the actual data */
	store: ScopedDataStore;
	/**  A simple KV store, designed for things like sync tokens */
	meta: MetaStore;
	logger: AstroIntegrationLogger;

	settings: AstroSettings;

	/** Validates and parses the data according to the schema */
	parseData<T extends Record<string, unknown> = Record<string, unknown>>(
		props: ParseDataOptions
	): T;
}

export interface ParseDataOptions {
	/** The ID of the entry. Unique per collection */
	id: string;
	/** The raw, unvalidated data of the entry */
	data: Record<string, unknown>;
	/** An optional file path, where the entry represents a local file */
	filePath?: string;
}
```
Currently there is no support for loading md or mdx files using `glob()`, or for defining renderers.

Querying is currently limited to `getCollection` and `getDataEntryById`, with no filter options.

To test this, install the experimental version with `npm install astro@experimental--content-layer`


ascorbic:
!preview content-layer

ascorbic:
!preview content-layer

ascorbic:
!preview content-layer

packages/astro/src/content/loaders.ts
```diff
@@ -0,0 +1,115 @@
...
+export interface Loader<S extends ZodSchema = ZodSchema> {
+	/** Unique name of the loader, e.g. the npm package name */
+	name: string;
+	/** Do the actual loading of the data */
+	load: (context: LoaderContext) => Promise<void>;
+	/** Optionally, define the schema of the data. Will be overridden by user-defined schema */
+	schema?: S | Promise<S> | (() => S | Promise<S>);
+	render?: (entry: any) => any;
+}
+export async function syncDataLayer({
```
ematipico, line 42: We could document this function, and explain how to use it and when/where.

packages/astro/src/content/loaders.ts
```diff
@@ -0,0 +1,115 @@
...
+	settings,
+	logger: globalLogger,
+	store,
+}: { settings: AstroSettings; logger: Logger; store?: DataStore }) {
+	const logger = globalLogger.forkIntegrationLogger('content');
+	if (!store) {
+		store = await DataStore.fromDisk(new URL('data-store.json', settings.config.cacheDir));
```
ematipico, line 49: What's `data-store.json`? Can we document it somewhere?

packages/astro/src/content/runtime.ts
```diff
@@ -153,6 +163,21 @@ export function createGetEntryBySlug({
 
 export function createGetDataEntryById({ getEntryImport }: { getEntryImport: GetEntryImport }) {
 	return async function getDataEntryById(collection: string, id: string) {
+		const store = await globalDataStore.get();
+
+		if (store.hasCollection(collection)) {
+			const data = store.get(collection, id);
+			if (!data) {
+				throw new Error(`Entry ${collection} → ${id} was not found.`);
```
ematipico, line 171: Should we consider an `AstroError` for the future, or was this intentional?

packages/astro/src/content/types-generator.ts
```diff
@@ -46,6 +46,10 @@ type CollectionEntryMap = {
 		| {
 				type: 'data';
 				entries: Record<string, DataEntryMetadata>;
+		  }
+		| {
+				type: 'experimental_data';
+				entries: Record<string, DataEntryMetadata>;
```
ematipico, line 52: Is this how we plan to ship the feature as experimental?

packages/astro/src/content/utils.ts
```diff
@@ -81,10 +105,10 @@ export async function getEntryData(
 	},
 	collectionConfig: CollectionConfig,
 	shouldEmitFile: boolean,
-	pluginContext: PluginContext
+	pluginContext?: PluginContext
```
ematipico, line 108: Why was this changed?

packages/astro/test/content-layer.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+describe('Content Collections', () => {
+	describe('Query', () => {
+		let fixture;
+		before(async () => {
+			fixture = await loadFixture({ root: './fixtures/content-layer/' });
+			await fixture.build();
+		});
+
+		describe('Collection', () => {
```
ematipico, line 13: Do you envision so many nested test cases (e.g. `describe`)? If not, we could have 1/2 describe maximum, because running `--test-only` would mean to add `.only` to all the `describe` functions (that's how `node:test` works)

packages/astro/src/content/file.ts
```diff
@@ -0,0 +1,51 @@
...
+			const json = JSON.parse(data);
+
+			const filePath = fileURLToPath(url);
+
+			if (Array.isArray(json)) {
+				if (json.length === 0) {
+					logger.warn(`No items found in ${fileName}`);
+				}
+				for (const rawItem of json) {
+					const id = rawItem.id ?? rawItem.slug;
```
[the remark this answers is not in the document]
ascorbic: Does this apply to both incoming data and queried data?

packages/astro/src/content/utils.ts
ascorbic: So that it can be used outside of a Vite plugin

packages/astro/src/content/types-generator.ts
ascorbic: This is something we should discuss. It has the benefit of meaning it's backward-compatible, as it can be used alongside the existing collection types.

packages/astro/src/content/runtime.ts
ascorbic: I've not created any new AstroErrors yet. For this one I just used the same approach as the one below, but it would make sense to use an AstroError for both.

packages/astro/test/content-layer.test.js
ascorbic: Yeah, I grabbed the structure from the content collections test, which has far more cases. For this I do expect there to be more tests, but not so many now to need so many levels

packages/astro/src/content/data-store.ts
```diff
@@ -0,0 +1,135 @@
+import { promises as fs, type PathLike, existsSync } from 'fs';
+export class DataStore {
+	#collections = new Map<string, Map<string, any>>();
```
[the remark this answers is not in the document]
ascorbic: I've not, because the store is a singleton and could be instantiated in several different ways and then shared. If it does turn out that it's needed then we can work it out, but for now I think it's ok without. Of course when we get to using libsql, we'll definitely need it.

packages/astro/src/content/file.ts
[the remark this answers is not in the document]
ematipico: We discussed this offline. Marking it as resolved

ematipico (approved):
Let's get the party started!

PR #11335, opened 2024-06-25
Revert "bug: removes type override from defineAction"

ematipico:
Reverts withastro/astro#11292

### DIFF
--- a/.changeset/afraid-rivers-begin.md
+++ b/.changeset/afraid-rivers-begin.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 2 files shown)


bholmesdev:
This was reverted due to incorrect types when the `input` parameter is omitted. The input should default to type `unknown`, but instead became `unknown | FormData`. This is because `json` was no longer assigned as the default. 

We’ll need to take a different path to fix the bug in the original PR.

bholmesdev (approved):
js-on with the show!

florian-lefebvre (approved):

PR #11337, opened 2024-06-25
feat: astro:env validateSecrets

florian-lefebvre:
## Changes

- Adds a new `experimental.env.validateSecrets` flag. It checks secrets on `buildStart` (dev and build, maybe more?)

## Testing

Adds a test

## Docs

Updates config reference and adds changeset


packages/astro/src/@types/astro.ts
```diff
@@ -2164,6 +2164,35 @@ export interface AstroUserConfig {
...
+			/**
+			 * @docs
+			 * @name experimental.env.validateSecretsOnStart
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `false`
+			 * @version 4.11.2
+			 * @description
+			 *
+			 * If enabled, secrets will also be checked on start (dev/build modes). This is useful for example in some CIs to make sure all your secrets are correctly set before deploying.
```
ematipico, line 2177: The description implies that these secrets aren't validated at the beginning. Maybe we should mention what's the default behaviour of `astro:env`, because - for now - this is our only documentation.

packages/astro/test/env-secret.test.js
```diff
@@ -1,77 +1,92 @@
...
+			experimental: {
+				env: {
+					validateSecretsOnStart: true,
+				},
+			},
 		});
+
+		let error = null;
+		try {
+			devServer = await fixture.startDevServer();
```
ematipico, line 83:
```suggestion
			devServer = await fixture.startDevServer();
			assert.fail();
```

Just making sure that we fail the test if no error was thrown

ematipico (approved):

packages/astro/test/env-secret.test.js
florian-lefebvre: If no error is thrown, the asserts below will catch that since error will not be an instance of `Error` right?

packages/astro/src/@types/astro.ts
```diff
@@ -2164,6 +2164,35 @@ export interface AstroUserConfig {
...
 			 */
 			schema?: EnvSchema;
+
+			/**
+			 * @docs
+			 * @name experimental.env.validateSecretsOnStart
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `false`
+			 * @version 4.11.3
```
florian-lefebvre, line 2174: I need to update it once we're confident it can be merged (not the case yet)

packages/astro/test/env-secret.test.js
ematipico: Ideally, you would want to run the assertions inside the `catch` block, and `.fail` inside the `try` block. There's so need to run assertions if you know something is failing to begin with.

But if you want to run the assertion **regardless**, I suggest to use the `finally` block

packages/astro/src/@types/astro.ts
sarah11918: It's as good as it can be until I see it live! This looks right to me, though.

packages/astro/src/@types/astro.ts
```diff
@@ -2166,6 +2166,35 @@ export interface AstroUserConfig {
...
+			 * @description
+			 *
+			 * By default, public variables are validated on start (dev/build modes) and secrets at runtime only. If enabled, secrets will also be checked on start. This is useful for example in some CIs to make sure all your secrets are correctly set before deploying.
```
sarah11918, line 2179:
```suggestion
			 * Whether or not to validate secrets on the server when starting the dev server or running a build.
			 *
			 * By default, only public variables are validated on the server at `buildStart`. Private variables are validated on the client at runtime only. If enabled, secrets will also be checked on start. This is useful in some continuous integration (CI) pipelines to make sure all your secrets are correctly set before deploying.
```

OK, I said a lot of things that I'm not sure are totally accurate ( :sweat_smile:  ) but if they are, I think this is a helpful amount of information to give! Can you make this correct?

.changeset/slow-roses-call.md
```diff
@@ -0,0 +1,23 @@
+---
+'astro': patch
+---
+
+Adds a new optional `experimental.env.validateSecretsOnStart` property
+
+If enabled, secrets will also be checked on start (dev/build modes). This is useful for example in some CIs to make sure all your secrets are correctly set before deploying.
```
sarah11918, line 7: Once the docs description is finalized, I'll suggest a shorter version for this line!

sarah11918 (commented):
Thanks @florian-lefebvre , looking really good!

Just wanted to discuss the final wording of the docs description, so I've given an example of maybe a scope/quantity that I think is helpful for you to work from and correct. :raised_hands:

packages/astro/src/@types/astro.ts
sarah11918: 1. Ok, so what I want to clarify with "on dev and build" is whether it's correct to say, "when starting the dev server or running a build"

I know with the dev server running, it's *like* continuously updating, so I'm not sure whether it's correct to say "when STARTING the dev server" or "when RUNNING the dev server". Can you confirm EXACTLY, being overly pedantic, when/what is happening so I can consider my options here?

2. I chose "private variables" to contrast with "public variables" because we'd already used the word "secret" and I figured this was "another way for people to understand." Not opposed to changing it, but using both interchangeably can help add variety to the text people have to read... IF they mean the same thing!

Can you confirm whether there's a different meaning or nuance in saying "secret variables" vs "private variables", or something that you feel "secret" conveys that "private" doesn't? If I know how private is sub-optimal, then maybe there's another phrase I can use as an alternative description when I want some word variety to avoid repetition.

florian-lefebvre: 1. I think it's technically "when STARTING the dev server"
2. I was not clear enough my bad! I don't really mind secret or private, my point was that "on the client" is wrong

sarah11918: D'oh, gotcha!

sarah11918: Ok, @florian-lefebvre - updated my suggestion, see whether you think that works!

sarah11918:
@florian-lefebvre re: naming -- *I* don't have a problem with the name, as it's clear! Chris might also be a good person to ask!

delucis (commented):
Firstly, before I make any suggestions, let me make sure I understand this correctly.

The context:

1. There are two different kinds of environment variables:
   1. public
   2. private (aka secret)

2. Public variables can end up inlined in client code or the server bundle.
3. Private variables (secrets) are never inlined this way. Instead they must be available at runtime (during either static prerender or on-demand server render).
4. In dev, all code runs in one single place:
   1. a local process where all required variables will need to be provided (`.env` file, command line etc.)
5. For 100% static builds, all code runs in one single place:
   1. a build process where all required variables will need to be set (CI provider settings, local `astro build`, etc.)
6. For builds with server or hybrid output, some code potentially runs in two environments:
   1. At build time in the CI provider/local build etc.
   2. On-demand when a request is made that needs to render something in a hosted server somewhere.
7. By default, `astro:env` only checks **public** environment variables are set when the dev server or build process starts.
8. If a required **secret** is missing, a user will discover by hitting a runtime error at one of these times:
   - In dev, visiting a page that needs the secret to render.
   - In builds, when statically prerendering a page that needs the secret to render.
   - On a server, when an on-demand route is requested that needs the secret to render.
9. The default behaviour means runtime errors are not protected against, but allows support for secrets that are **only** available in the on-demand render server context.

This PR:

- Enables checking that private environment variables (secrets) are set **when the dev server or build process starts**.

Questions:

- Is that an accurate summary?
- Does this option mean runtime errors for on-demand rendered requests can no longer happen? Or even if you check secrets are present during a build, can they still be missing in the final on-demand server context?

florian-lefebvre:
Excellent summary!

> Does this option mean runtime errors for on-demand rendered requests can no longer happen? Or even if you check secrets are present during a build, can they still be missing in the final on-demand server context?

I don't think so. Technically, you could update a secret between a build and the deployment.

---

Also, something I've been thinking about having `start` in the option name: some people could think it means they are validated when the production servers starts, which is not the case

delucis:
Thanks for confirming @florian-lefebvre! OK, so given your answer, I’m wondering a bit **who** this new option is intended to help?

In dev and static builds (points 4 & 5 in my summary), this option means they can fail fast, so failures happen quicker but the builds would have failed anyway eventually.

In production builds with on-demand rendering (point 6), this option does the same, but provides no protection on deployment for the reasons you mention.

So is a separate option needed to just _move_ when in the process the failure happens? What was the user request that led to this? It feels like the more important failure to protect against would be in the production environment, but this is not adding extra protection there.

If the option does seem valuable, maybe even something like `env.failFast` or similar could be an idea for naming? Although not sure it does much to explain the **when**.

florian-lefebvre:
fyi that's a feature i had in `astro-env` and that saved me quite a few times. A common thing is you add a new required secret var, you make a PR and you forget to add it to your host (and ofc it crashes once in prod). This feature allows to fail early.

I agree that the issue is to convey the "when" 😅

delucis:
> that's a feature i had in `astro-env`

You mean this PR? Or the behaviour I was describing?

delucis:
> I want my build to fail if any variable (public or secret) is not valid. I don't want to discover it at runtime

Isn’t that already the case?

Based on my summary above the only case where currently it won’t fail during a build is:

- A variable only used in the on-demand SSR runtime

And you said above that this PR won’t change that part — on-demand SSR can still fail at runtime because it runs in a different environment from the build.

florian-lefebvre:
yes but some hosts expose the same secrets on both build and runtime (eg. netlify) so that's why it's interesting imo.
Basically
- Before this PR, invalid secret is only discovered at build time with a 500 error
- After this PR and the option enabled, build fails because variable is invalid. Of course an error can still occur at runtime if the secret is updated between the build and the deployment

Idk if that clarifies anything

delucis:
Thanks, yeah that confirms what I suspected:

- moves the check earlier in the build process
- catches issues **IF** you happen to deploying in a way where the environment is consistent between build and prod

OK, last question (thanks for your patience 😅):

**Does “validation” already only happen at “start”?**  
In other words: public variables get validated at dev/build start, but private variables are not ever currently “validated”, they just fail at runtime?

If that’s the case, maybe this option just gets simplified to

```js
env: {
  validateSecrets: true,
}
```

delucis:
Nice! OK, then that concludes my questions — thanks again for going through them all 🙌

packages/astro/src/@types/astro.ts
```diff
@@ -2166,6 +2166,37 @@ export interface AstroUserConfig {
...
+			 * // astro.config.mjs
+			 * import { defineConfig, envField } from "astro/config"
+			 *
+			 * export default defineConfig({
+			 *   experimental: {
+			 *     env: {
+			 *       schema: {
+			 * 			// ...
+			 *       },
+			 * 		 validateSecrets: true
```
florian-lefebvre, line 2193:
```suggestion
			 *       validateSecrets: true
```

packages/astro/src/@types/astro.ts
```diff
@@ -2166,6 +2166,37 @@ export interface AstroUserConfig {
...
+			 * 
+			 * ```js
+			 * // astro.config.mjs
+			 * import { defineConfig, envField } from "astro/config"
+			 *
+			 * export default defineConfig({
+			 *   experimental: {
+			 *     env: {
+			 *       schema: {
+			 * 			// ...
```
florian-lefebvre, line 2191:
```suggestion
			 *         // ...
```

.changeset/slow-roses-call.md
```diff
@@ -0,0 +1,23 @@
+---
+'astro': patch
+---
+
+Adds a new optional `experimental.env.validateSecrets` property
```
sarah11918, line 5:
```suggestion
Adds a new property `experimental.env.validateSecrets` to allow validating private variables on the server.
```

packages/astro/src/@types/astro.ts
```diff
@@ -2166,6 +2166,37 @@ export interface AstroUserConfig {
...
 			 */
 			schema?: EnvSchema;
+
+			/**
+			 * @docs
+			 * @name experimental.env.validateSecrets
+			 * @kind h4
+			 * @type {boolean}
+			 * @default `false`
+			 * @version 4.11.6
```
sarah11918, line 2176: just pointing out this looks right to me if this is merged now, but if it misses the next patch, you'll need to update the version number here!

.changeset/slow-roses-call.md
```diff
@@ -0,0 +1,23 @@
...
+
+If enabled, secrets will also be checked on start (dev/build modes). This is useful for example in some CIs to make sure all your secrets are correctly set before deploying.
```
sarah11918, line 7:
```suggestion
By default, this is set to `false` and only public variables are checked on start. If enabled, secrets will also be checked on start (dev/build modes). This is useful for example in some CIs to make sure all your secrets are correctly set before deploying.
```

Haha, so didn't get shorter....

sarah11918 (approved):
Left a couple of suggestions, mostly for word variety and to make an "also" explicit  e: what it was in comparison to!

But looks great, and whatever you choose from here should be fine!

PR #11348, opened 2024-06-26
fix: prevent client hydration when rendering via Container API

ematipico:
## Changes

This PR fixes a bug inside the container API, where an error was thrown when rendering a client component using `client:*` directives.

Closes https://github.com/withastro/astro/issues/11407

This issue was brought up on discord: https://discord.com/channels/830184174198718474/1255408685275021342

To fix this bug, I did the two following things:
- updated the container API to bootstrap its manifest with the default directives. This fixes an error thrown saying "invalid directive"
- updated the `SSResult` to add a new attribute called `skipHydration`, which is used inside the conditions that do work for client hydration. This attributed is also added in `RenderContext` and passed down to the renderer. In this way, the container API can turn off the logic related to client hydration, because it isn't needed

## Testing

Added a new test case

## Docs

N/A


packages/integrations/react/src/index.ts
```diff
@@ -60,6 +60,7 @@ function getViteConfiguration(
 				'react/jsx-runtime',
 				'react/jsx-dev-runtime',
 				'react-dom',
+				``,
```
florian-lefebvre, line 63: What is this for?

ematipico: for nothing, good catch

florian-lefebvre (approved):

matthewp:
Does this prevent using client directives in all Container rendered content or just some? If this prevents all then that means the PHP / Rails use-case doesn't work any more.

ematipico:
> Does this prevent using client directives in all Container rendered content or just some? If this prevents all then that means the PHP / Rails use-case doesn't work any more.

You're probably right. Do you think we should provide an option when we call the render function?

ematipico:
I added an option to the `render*` functions.

matthewp:
I don't think we should have such an option. Why is the user not passing through the directives? Is the problem that we don't have a good way to import those into the test? If so that's what we should fix, not add an option.

ematipico:
> I don't think we should have such an option. Why is the user not passing through the directives? Is the problem that we don't have a good way to import those into the test? If so that's what we should fix, not add an option.

Users have components like this:

```astro
---
// Button.astro
import VueButton from "../components/Button.vue"
---

<VueButton client:idle>Text<VueButton>
```

So, I am not sure there's a good way to "turn off" the hydration checks inside the renderer. Do you have any concrete solution in mind?

matthewp:
You should be able to pass in the directive config via AstroContainer. It's on the SSR manifest. Can we expose this? https://github.com/withastro/astro/blob/main/packages/astro/src/core/client-directive/default.ts#L7

```js
import {getDefaultClientDirectives, AstroContainer} from 'astro:container';

let container = await AstroContainer.create({
  clientDirectives: getDefaultClientDirectives()
});
```

ematipico:
> You should be able to pass in the directive config via AstroContainer. It's on the SSR manifest. Can we expose this? [`main`/packages/astro/src/core/client-directive/default.ts#L7](https://github.com/withastro/astro/blob/main/packages/astro/src/core/client-directive/default.ts?rgh-link-date=2024-06-27T14%3A37%3A46Z#L7)
> 
> ```js
> import {getDefaultClientDirectives, AstroContainer} from 'astro:container';
> 
> let container = await AstroContainer.create({
>   clientDirectives: getDefaultClientDirectives()
> });
> ```

This is already present in the current PR, but it isn't enough to avoid errors.

aloverso:
Hi - my team is also facing this issue and we need this fix to move forward with our testing strategy. Is there any timeline on when this will be merged?

matthewp:
@aloverso I don't think disabling client hydration is the right fix here. Need to figure out why it's breaking.

PR #11349, opened 2024-06-26
fix: check experimental flag when using the rewrite function

ematipico:
## Changes

Closes https://github.com/withastro/docs/issues/8660

## Testing

Manually tested

## Docs

N/A

### DIFF
--- a/.changeset/curvy-elephants-fry.md
+++ b/.changeset/curvy-elephants-fry.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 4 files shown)


florian-lefebvre (approved):
Looks good but since it's duplicated, couldn't it be extracted to a shared function?

bluwy (approved):
Same thoughts as Florian too, and perhaps if can do a

```js
if (!this.pipeline.manifest.rewritingEnabled) {
  /// log error
  return
}
```

To prevent the indentation here and when we stabilize the feature, but not a huge blocker.

PR #11352, opened 2024-06-26
fix(rewrite): correctly update the status code during a rewrite

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11306

The issue was caused by `this.status` inside the `RenderContext` class. This status code is then passed to the `SSResult` and used as initial status of `Response`. 

This status code was passed by `route.ts`. It was a 404 because `/foo` didn't match any route, however we still call the middleware because the user chose `routing: 'manual'`.

When we arrived to the middleware, we executed the rewrite, and we need update this status code in case we find a `RouteData` and a `ComponentInstance`.

I moved the shared logic inside a new, private, function called `#executeRewrite` as per @florian-lefebvre suggestion.

## Testing

Added a new test case

## Docs

N/A

### DIFF
--- a/packages/astro/src/runtime/server/render/page.ts
+++ b/packages/astro/src/runtime/server/render/page.ts
@@ -92,6 +92,7 @@ export async function renderPage(
 	} else if (route?.route === '/500') {
 		status = 500;
 	}
+	console.log(status, init.status);
 	if (status) {
 		return new Response(body, { ...init, headers, status });
 	} else {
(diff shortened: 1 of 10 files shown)


packages/astro/src/runtime/server/render/page.ts
```diff
@@ -92,6 +92,7 @@ export async function renderPage(
 	} else if (route?.route === '/500') {
 		status = 500;
 	}
+	console.log(status, init.status);
```
bluwy, line 95: Should be removed?

bluwy (approved):

PR #11356, opened 2024-06-26
fix: add missing href to canonicalURL

ColinWttt:
the `new URL(Astro.url.pathname, Astro.site)` return `URL`, only the `href` is needed
```
<!-- Canonical URL -->
<link rel="canonical" href={canonicalURL.href} />
```
## Changes

- add `.href`to `canonicalURL`

## Docs

https://docs.astro.build/en/reference/api-reference/#astrocanonicalurl


bluwy:
I don't think it matters in practice? Astro cast it to a string during render (which will be the same as `.href`), so there isn't a difference here. The docs also document passing `URL` directly: https://docs.astro.build/en/reference/api-reference/#astrourl

bluwy:
Closing this for now

PR #11358, opened 2024-06-27
feat: add typegen for loaders

ascorbic:
## Changes

Adds typegen for loaders. Unlike collections where the schema is defined by the user, these can't be inferred. This is because they can be generated dynamically by the loader during sync. This may happen for example when introspecting a CMS. Because of this, we use `zod-to-ts` to generate the actual types during sync.


packages/astro/src/content/loaders.ts
```diff
@@ -30,13 +30,13 @@ export interface LoaderContext {
 	): T;
 }
 
-export interface Loader<S extends ZodSchema = ZodSchema> {
+export interface Loader {
```
ascorbic, line 33: Making this generic messed up inference for top level schemas

packages/astro/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "declarationDir": "./dist",
     "outDir": "./dist",
     "jsx": "preserve",
-    "types": ["@types/dom-view-transitions"]
+    "types": ["@types/dom-view-transitions"],
+    "rootDir": "src"
```
ascorbic, line 10: The compiler complained about ambiguous rootDir for some reason

florian-lefebvre: I remember facing this as well, I think you need to revert the import change from `astro/zod` to `zod`

ascorbic: Thanks! That did the trick. Well remembered

ematipico:
Before merging this one, can you create a PR for these changes https://github.com/withastro/astro/pull/11334 ?

I know we are going to use that PR for preview releases and the final release, but it would be useful to have a PR where we can actually add comments without littering the original PR, which will be used for docs

ascorbic:
@ematipico sorry, I don't understand. A PR from which branch to which branch?

ematipico:
> @ematipico sorry, I don't understand. A PR from which branch to which branch?

Something like this
- final PR: `main` <- `contect-collections-v2` (not needed right now, but you can open it anyway as a draft)
- then a PR `contect-collections-v2` <- `content-layer-loader`, so we can review and eventually merge it
- then this PR will be `contect-collections-v2` <- `loader-types` **after** we merged the previous PR

ascorbic:
OK, so the new PR would be a branch that's currently identical to main?

ematipico:
Yes exactly

packages/astro/src/content/types-generator.ts
```diff
@@ -5,6 +5,7 @@ import glob from 'fast-glob';
 import { bold, cyan } from 'kleur/colors';
 import { type ViteDevServer, normalizePath } from 'vite';
 import { z } from 'zod';
+import { zodToTs, printNode } from 'zod-to-ts';
```
florian-lefebvre, line 8: We'll have to document the limitations of this library, iirc some types can't be generated

PR #11360, opened 2024-06-27
Content Layer

ascorbic:
This is the work-in-progress branch for [the Content Layer](https://github.com/withastro/roadmap/blob/content-layer/proposals/content-layer.md)
The Content Layer API is a new way to handle content and data in Astro. It builds upon [content collections](https://docs.astro.build/en/guides/content-collections/), taking them beyond local files in `src/content` and allowing you to fetch content from anywhere, including remote APIs, or files anywhere in your project. As well as being more powerful, the Content Layer is designed to be more performant, helping sites scale to thousands of pages. Data is cached between builds and updated incrementally. Markdown parsing is also 5-10x faster, with similar scale reductions in memory. While the feature is experimental and subject to breaking changes, we invite you to try it today and let us know how it works for you.

#### Enabling content layer

To enable, add the `contentLayer` flag to the `experimental` object in your Astro config:

```js
{
	experimental: {
		contentLayer: true,
	}
}
```

#### Using the content layer

>[!TIP]
>
> The content layer APIs are based on content collections, so you can learn more about them in the [content collection docs](https://docs.astro.build/en/guides/content-collections/). Any differences are highlighted below.

To use content layer, create a collection in `src/content/config.ts` with a `loader` property. For local files where there is one entry per file, use the `glob()` loader. You can put your content files anywhere, but *not* in `src/content` because these would be handled by the current content collections instead. In this example the files are in `src/data`.

```ts
import { glob } from 'astro/loaders';

  // By default the ID is a slug, generated from the path of the file relative to `base`
  loader: glob({ pattern: "**/*.md", base: "./src/data/blog" }),
  schema: z.object({
    title: z.string(),
    description: z.string(),
    pubDate: z.coerce.date(),
    updatedDate: z.coerce.date().optional(),
  }),
});

export const collections = { blog };
```

You can load multiple entries from a single JSON file using the `file()` loader. In this case the data must either be an array of objects, which each contain an `id` property, or an object where each key is the ID.

**Array syntax:**

```json
[
  {
    "id": "labrador-retriever",
    "breed": "Labrador Retriever",
    "size": "Large",
    "origin": "Canada",
    "lifespan": "10-12 years",
    "temperament": [
      "Friendly",
      "Active",
      "Outgoing"
    ]
  },
  {
    "id": "german-shepherd",
    "breed": "German Shepherd",
    "size": "Large",
    "origin": "Germany",
    "lifespan": "9-13 years",
    "temperament": [
      "Loyal",
      "Intelligent",
      "Confident"
    ]
  }
]
```

**Object syntax:**

```json
{
  "labrador-retriever": {
    "breed": "Labrador Retriever",
    "size": "Large",
    "origin": "Canada",
    "lifespan": "10-12 years",
    "temperament": [
      "Friendly",
      "Active",
      "Outgoing"
    ]
  },
  "german-shepherd": {
    "breed": "German Shepherd",
    "size": "Large",
    "origin": "Germany",
    "lifespan": "9-13 years",
    "temperament": [
      "Loyal",
      "Intelligent",
      "Confident"
    ]
  }
}
```

The collection is then defined using the `file()` loader:

```ts
import { file } from 'astro/loaders';

const dogs = defineCollection({
	loader: file('src/data/dogs.json'),
	schema: z.object({
		id: z.string(),
    breed: z.string(),
		size: z.string(),
		origin: z.string(),
		lifespan: z.string(),
		temperament: z.array(z.string()),
	}),
});

export const collections = { dogs };
```

The collection can be queried in the same way as content collections:

```ts
import { getCollection, getEntry } from 'astro:content';

// Get all entries from a collection.
// Requires the name of the collection as an argument.
const allBlogPosts = await getCollection('blog');

// Get a single entry from a collection.
// Requires the name of the collection and ID
const labradorData = await getEntry('dogs', 'labrador-retriever');
```

#### Rendering content

Entries generated from markdown or MDX can be rendered directly to a page using the `render()` function. 

> [!WARNING]  
>
> The syntax for rendering content layer collections is different from existing content collections syntax.

```astro
---
import { getEntry, render } from 'astro:content';

const post = await getEntry('blog', Astro.params.slug);

const { Content, headings } = await render(entry);
---

<Content />
```

#### Creating a content layer loader

Content layer loaders aren't restricted to just loading local files. You can also use loaders to load or generate content from anywhere. The simplest type of loader is an async function that returns an array of objects, each of which has an `id`:

```ts
const countries = defineCollection({
  loader: async () => {
    const response = await fetch("https://restcountries.com/v3.1/all");
    const data = await response.json();
    // Must return an array of entries with an id property, or an object with IDs as keys and entries as values
    return data.map((country) => ({
      id: country.cca3,
      ...country,
    }));
  },
  // optionally add a schema
  // schema: z.object...
});

export const collections = { countries };
```

For more advanced loading logic, you can define an object loader. This allows incremental updates and conditional loading, and gives full access to the data store. See the API in [the draft RFC](https://github.com/withastro/roadmap/blob/content-layer/proposals/content-layer.md#loaders).

### Migrating a markdown or MDX collection to content layer

If you have used [content collections](https://docs.astro.build/en/guides/content-collections/), content layer will be very familiar as most of the APIs are the same. You can try converting an existing content collection to content layer if it uses markdown, MDX or JSON, with these steps:

1. **Move the collection folder out of `src/content`.** This is so it won't be handled as a current content collection. This example assumes the content has been moved to `src/data`. The `config.ts` file must remain in `src/content`.
2. **Edit the collection definition**. The collection should not have `type` set, and needs a `loader` defined.

```diff
import { defineCollection, z } from 'astro:content';
+ import { glob } from 'astro/loaders';

const blog = defineCollection({
  // For content layer you do not define a `type`
- type: 'content',
+ loader: glob({ pattern: "**/*.md", base: "./src/data/blog" }),	
  schema: z.object({
		title: z.string(),
		description: z.string(),
		pubDate: z.coerce.date(),
		updatedDate: z.coerce.date().optional(),
	}),
});
```

3. **Change references from `slug` to `id`**. Content layer collections do not have a `slug` field. You should use `id` instead, which has the same syntax.

```diff
---
export async function getStaticPaths() {
	const posts = await getCollection('blog');
	return posts.map((post) => ({
-   params: { slug: post.slug },
+   params: { slug: post.id },
    props: post,
	}));
}
---
```

4. **Switch to the new `render()` function**. Entries no longer have a `render()` method, as they are now serializable plain objects. Instead, import the `render()` function from `astro:content`. 

```diff
---
- import { getEntry } from 'astro:content';
+ import { getEntry, render } from 'astro:content';

  const post = await getEntry('blog', params.slug);

- const { Content, headings } = await post.render();
+ const { Content, headings } = await render(entry);
---

<Content />
```

The `getEntryBySlug` and `getDataEntryByID` functions are deprecated and cannot be used with content layer collections. Instead, use `getEntry`, which is a drop-in replacement for both.

#### Learn more

To see the full API look at [the draft RFC](https://github.com/withastro/roadmap/blob/content-layer/proposals/content-layer.md) and [share your feedback on the feature and API](https://github.com/withastro/roadmap/pull/982).


ascorbic:
!preview content-layer

ascorbic:
!preview contentlayer

ascorbic:
!preview contentlayer

ascorbic:
!preview contentlayer

toxsick:
@ascorbic awesome first implementation. I was just playing around with it in combination with statamic content and i was wondering if the glob loader could also get a optional `generateSlug` fn with the same api as generateId. The reason is the statamic content yaml files have an id field and the slug needs to be generated based on the file path. So I would like to do something like this:

```
const pagesCollection = defineCollection({
  type: 'experimental_content',
  loader: glob({
    pattern: '**/*.md',
    base: '../../content/collections/pages',
    generateId: ({ data }) => data.id as string,
    // here I would like to do the same as in generateIdDefault
    generateSlug: () => 'slug'
  }),
  schema: z.object({
    id: z.string(),
    title: z.string().optional(),
  }),
});
```
Is this possible somehow right now?

ascorbic:
@toxsick the glob loader doesn't create a slug - you should use ID instead. By default the ID *is* the slug.

toxsick:
> @toxsick the glob loader doesn't create a slug - you should use ID instead. By default the ID _is_ the slug.

Jeah alright, but could the glob loader behave like the current implementation in [Defining custom slugs](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs), that when a slug field is present it will also be added to the entry, so resolving the entry with [getEntry](https://docs.astro.build/en/reference/api-reference/#getentry) works by both id and slug?

ascorbic:
Having special behaviour for `slug` isn't something that we'll be doing. You could implement it with a combination of `generateId` and schema though.

toxsick:
@ascorbic thanks for the info, I guess my situation is an edge case here. I already have the slug in the data schema and I will resolve the entry by slug using `getCollection` with a filter then.

Thanks for the clarification and I look forward to this being released!

ematipico (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1469,6 +1469,32 @@ export const UnknownContentCollectionError = {
...
 } satisfies ErrorData;
+
+/**
+ * @docs
+ * @description
+ * Astro encountered an unexpected error when writing your content layer data to disk.
+ *
+ * If you can reliably cause this error to happen, we'd appreciate if you could [open an issue](https://astro.build/issues/)
+ */
+export const ContentLayerWriteError = {
```
Princesseuh, line 1480: A similar error was added for injectType, I think we make it use teh same one

ascorbic: Do you want a include any renaming/rewording in this PR, or will you do it in yours?

Princesseuh: I'll merge them post merge

Princesseuh: Done!

packages/astro/src/content/data-store.ts
```diff
@@ -157,10 +157,7 @@ export class DataStore {
 			try {
 				await fs.writeFile(filePath, 'export default new Map();');
 			} catch (err) {
-				throw new AstroError({
-					message: (err as Error).message,
-					...AstroErrorData.ContentLayerWriteError,
-				});
+				throw new AstroError(AstroErrorData.UnknownFilesystemError, { cause: err });
```
ascorbic, line 160: @Princesseuh Will this surface the actual FS error message to the user?

PR #11371, opened 2024-06-28
fix: watch for content layer changes

ascorbic:
## Changes

- adds a `watcher` to the loader props, which can be used to update data. I've not tried to do anythign with scoping the watcher, because I think it's best to allow flexibility there
- uses this watcher in the `file()` loader to reload data when the file changes
- because we now can change data in the store outside of the main lifecycle, I've needed to add support for automatically saving changes to disk. I did this with a debounced save function that's automatically called whenever the store is changed. This will save to disk 500ms after the last data change
- during dev, it watches the data store file and invalidates the virtual module when it changes

## Testing

Adds `dev` tests, including one that edits the data file

### DIFF
--- a/packages/astro/test/test-utils.js
+++ b/packages/astro/test/test-utils.js
@@ -36,7 +36,7 @@ process.env.ASTRO_TELEMETRY_DISABLED = true;
  * @property {(path: string) => Promise<boolean>} pathExists
  * @property {(url: string, opts: Parameters<typeof fetch>[1]) => Promise<Response>} fetch
  * @property {(path: string) => Promise<string>} readFile
- * @property {(path: string, updater: (content: string) => string) => Promise<void>} writeFile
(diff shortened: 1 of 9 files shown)


packages/astro/test/test-utils.js
```diff
@@ -36,7 +36,7 @@ process.env.ASTRO_TELEMETRY_DISABLED = true;
  * @property {(path: string) => Promise<boolean>} pathExists
  * @property {(url: string, opts: Parameters<typeof fetch>[1]) => Promise<Response>} fetch
  * @property {(path: string) => Promise<string>} readFile
- * @property {(path: string, updater: (content: string) => string) => Promise<void>} writeFile
+ * @property {(path: string, updater: (content: string) => string) => Promise<void>} editFile
```
ascorbic, line 39: This was an error in the jsdoc types

matthewp (approved):

PR #11372, opened 2024-06-28
Add support for the build to Server Islands

matthewp:
## Changes

- Adds support for the build.
- Discovered islands are built using `this.emitFile` in the Vite/Rollup plugin.
- The name / path pair are stored in a map and saved on the manifest.
- From here it's the same as in dev mode.

## Testing

- Not yet, adding tests in a follow-up

## Docs

N/A


packages/astro/src/core/server-islands/vite-plugin-server-islands.ts
```diff
@@ -2,44 +2,92 @@ import type { AstroPluginMetadata } from '../../vite-plugin-astro/index.js';
...
+				const islandName = settings.serverIslandNameMap.get(resolvedPath)!;
+				mapSource += `\n\t['${islandName}', () => import('./${fileName}')],`
+			}
+			mapSource += '\n]);';
+			referenceIdMap.clear();
+
+			for (const [fileName, output] of Object.entries(bundles)) {
+				if(output.type !== 'chunk') continue;
+
+				//console.log("OUTPUT", output.code);
```
ematipico, line 85: to remove

packages/astro/src/core/server-islands/vite-plugin-server-islands.ts
```diff
@@ -2,44 +2,92 @@ import type { AstroPluginMetadata } from '../../vite-plugin-astro/index.js';
...
+
+								// Append the name map, for prod
+								settings.serverIslandNameMap.set(comp.resolvedPath, name);
+
+								settings.serverIslandMap.set(name, () => {
+									return viteServer?.ssrLoadModule(comp.resolvedPath) as any;
+								});
+
+								// Build mode
+								if(!viteServer) {
```
ematipico, line 56: We usually use `command` to check if we are inside a build; why was this changed here?

packages/astro/src/core/server-islands/vite-plugin-server-islands.ts
```diff
@@ -2,44 +2,92 @@ import type { AstroPluginMetadata } from '../../vite-plugin-astro/index.js';
...
+			referenceIdMap.clear();
+
+			for (const [fileName, output] of Object.entries(bundles)) {
+				if(output.type !== 'chunk') continue;
+
+				//console.log("OUTPUT", output.code);
+
+				if(output.code.includes(serverIslandPlaceholder)) {
+					output.code = output.code.replace(serverIslandPlaceholder, mapSource);
+				}
```
bluwy, line 89: Editing the code in place wouldn't update the file hash anymore, which could lead to subtle bugs. I think you can use the `augmentChunkHash` rollup hook to add the relevant reference ids as keys for Rollup to hash to fix this.

matthewp: I used `renderChunk`

PR #11379, opened 2024-06-30
chore: unflag contentCollectionJsonSchema

alexanderniebuhr:
## Open questions

- I need help with the changeset
- We should finalize the naming of this flag and how to write it `Json` or `JSON`, etc.. (cc @sarah11918 for the config naming)

## Changes

- based on discussions in Discord this moves the feature to stable (experimental PR https://github.com/withastro/astro/pull/10145)
- introduced a new section for `contentCollection` inside the config, because we might habe other options being stabilized in the future, such as cache, etc. 
- An alternative thought was to move those configs into the `content.ts` file, but that doesn't felt right
- renamed the flag to `JsonSchema` to make use of new config section

## Testing

- updated existing tests to new setting

## Docs

- https://github.com/withastro/docs/pull/8705


matthewp:
Can this wait until 4.13?

matthewp:
The reason to wait is that some people are on vacation that have more context on this feature and would like to get their input on whether it's missing anything or if there's any other reason to delay dropping the flag.

sarah11918 (commented):
Just noting that I'll come in here before 4.13 to help with the changeset!

matthewp:
Hey @alexanderniebuhr, talked a bit about this async and we were wondering:

- Is there ever a reason to turn this feature off? Since users have to use the schema manually it's not something that should really be a bother, I don't think. Just a file in the `.astro` folder that we already own.
  - The only reason I can think to make it configurable would be if there was a performance issue; like if generating the schemas took a while, but I very much doubt it.
  
If there's not a reason to disable it, we'd probably favor not having the option at all in config and just always doing it.

alexanderniebuhr:
> Since users have to use the schema manually it's not something that should really be a bother, I don't think.

This is the current state, the longterm goal is to improve our language-tools and extension, so users don't need to configure it manually.

> Just a file in the .astro folder that we already own.

It's one file for each data collection, so if you have multiple, you get multiple files

> The only reason I can think to make it configurable would be if there was a performance issue;

I don't see any performance issue here, but we use a 3rd party dependency to generate schemas, if that fails for any reason, a config flag would allow users to opt-out while we investigate the bug and fix it, otherwise they would need for our fix


> we'd probably favor not having the option at all in config and just always doing it.

Your call, I'm happy with both, above is just more context and not any arguments for any sides.

sarah11918:
Glad this is making 4.13, @alexanderniebuhr !

To get you started on the changeset, here are [my standard docs on what to write for unflagging an experimental feature](https://contribute.docs.astro.build/docs-for-code-changes/experimental-feature-docs/#unflagging-an-experimental-feature)!

.changeset/tasty-spoons-double.md
```diff
@@ -0,0 +1,48 @@
+---
+'astro': minor
+---
+
+The `experimental.contentCollectionJsonSchema` feature introduced behind a flag in [v4.5.0](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#450) is no longer experimental and is available for general use.
+
+If you are working with collections of type `data`, Astro will auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema#known-issues).
```
sarah11918, line 7:
```suggestion
If you are working with collections of type `data`, Astro will auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema#known-issues).
```
Just checking that you want this to go directly to the "known issues" section on that page? From the text, it doesn't sound like that's where I'd expect to go. Probably safe to just link to the main page?

sarah11918, line 7:
```suggestion
If you are working with collections of type `data`, Astro can now auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema#known-issues).
```
Just realizing that we say it "will" do something, but you have to do some manual steps. Is "can" more accurate?

And, I wouldn't put this in official documentation, but I do like using "now" for this changeset message, drawing attention to what's new.

sarah11918 (commented):
Just some quick thoughts, but otherwise looks good! Congrats on unflagging this feature! :partying_face:

.changeset/tasty-spoons-double.md
```diff
@@ -0,0 +1,48 @@
+---
+'astro': minor
+---
+
+The `experimental.contentCollectionJsonSchema` feature introduced behind a flag in [v4.5.0](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#450) is no longer experimental and is available for general use.
+
+If you are working with collections of type `data`, Astro will auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema).
```
alexanderniebuhr, line 7:
```suggestion
If you are working with collections of type `data`, Astro will now auto-generate JSON schema files for your editor to get IntelliSense and type-checking. A separate file will be created for each data collection in your project based on your collections defined in `src/content/config.ts` using a library called [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema).
```

sarah11918 (approved):
Approving for docs! :partying_face:

bluwy (commented):
There seems to be many unnecessary whitespace/indent changes in this PR. Is it possible to revert them to not affect the git history?

alexanderniebuhr:
I can revert them, but I got those automaticity based on the format configuration 🤔

bluwy:
I hope you didn't format those back manually, but we do have `pnpm format` at the root which should auto-fix everything 😅 Maybe you can compare if your editor setup isn't aligning with it. It should use Biome to format by default.

alexanderniebuhr:
I guess it does not pick Biome by default in my editor, because there is not such configuration in `.vscode/settings.json`, but anyways the formatting in this PR should be fixed now...

P.S.: I can't make Biome the default for my VSCode instance completely, so it either has to be configured inside the repo or I wouldn't be able to use "Format on Save" at all...

bluwy:
You can set it per-workspace. For example, I have this in `.vscode/settings.json`:

```json
{
  "typescript.tsdk": "node_modules/typescript/lib",
  "references.preferredLocation": "peek",
  "[typescript]": {
    "editor.defaultFormatter": "biomejs.biome"
  },
  "[javascript]": {
    "editor.defaultFormatter": "biomejs.biome"
  }
}
```

alexanderniebuhr:
Yeah that's what I meant, but IMO this should be committed to the repository, so I don't have to configure it, but the correct configuration for the repository is already defined by maintainers.

bluwy:
I don't think it should be committed. If others have different workspace config, we don't want to accidentally commit them either. These settings are for personal preference.

bluwy (approved):

PR #11385, opened 2024-07-01
Add support for LibSQL remote

Fryuni:
## Changes

- Allows `ASTRO_STUDIO_REMOTE_DB_URL` to point to a LibSQL server instance (using any supported LibSQL protocols). This enables Astro DB to be used with self-hosting and air-gapped deployments.
- Accept extra LibSQL options as query parameters in the remote URL.
  - For example, `memory:?syncUrl=libsql%3A%2F%2Fdb-server.example.com` would create an in-memory embedded replica for the LibSQL DB on `libsql://db-server.example.com`.

## Testing

Replicated the test for Studio remote using the LibSQL remote

## Docs

I'll open the docs PR if this is of interest :)

### DIFF
--- a/.changeset/healthy-boxes-poke.md
+++ b/.changeset/healthy-boxes-poke.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/db': minor
+---
+
+Adds support for self-hosted LibSQL
--- a/packages/db/src/runtime/db-client.ts
+++ b/packages/db/src/runtime/db-client.ts
@@ -1,5 +1,5 @@
 import type { InStatement } from '@libsql/client';
-import { createClient } from '@libsql/client';
+import { createClient, type Config as LibSQLConfig } from '@libsql/client';
 import type { LibSQLDatabase } from 'drizzle-orm/libsql';
 import { drizzle as drizzleLibsql } from 'drizzle-orm/libsql';
 import { type SqliteRemoteDatabase, drizzle as drizzleProxy } from 'drizzle-orm/sqlite-proxy';
@@ -23,7 +23,9 @@ export function createLocalDatabaseClient({ dbUrl }: { dbUrl: string }): LibSQLD
 	const client = createClient({ url });
 	const db = drizzleLibsql(client);
 
-	applyTransactionNotSupported(db);
+	if (!process.env.ASTRO_DB_ENABLE_TRANSACTIONS) {
+		applyTransactionNotSupported(db);
+	}
 	return db;
 }
 
@@ -36,6 +38,41 @@ const remoteResultSchema = z.object({
 });
 
 export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string) {
+	const remoteUrl = new URL(remoteDbURL);
+
+	switch (remoteUrl.protocol) {
+		case 'http:':
+		case 'https:':
+			return createStudioDatabaseClient(appToken, remoteUrl)
+    case 'libsql+http:':
+    	remoteUrl.protocol = 'http:';
+    case 'libsql+https:':
+    	remoteUrl.protocol = 'https:';
+    case 'ws:':
+    case 'wss:':
+    case 'file:':
+    case 'libsql:':
+			return createRemoteLibSQLClient(appToken, remoteUrl);
+		default:
+			throw new Error(`Unsupported remote DB: ${remoteDbURL}`);
+	}
+}
+
+function createRemoteLibSQLClient(appToken: string, remoteDbURL: URL) {
+	const options: Partial<LibSQLConfig> = Object.fromEntries(remoteDbURL.searchParams.entries());
+	remoteDbURL.search = '';
+
+	const client = createClient({ 
+		...options,
+		authToken: appToken,
+		url: remoteDbURL.toString(),
+	});
+	const db = drizzleLibsql(client);
+
+	return db;
+}
+
+function createStudioDatabaseClient(appToken: string, remoteDbURL: URL) {
 	if (appToken == null) {
 		throw new Error(`Cannot create a remote client: missing app token.`);
 	}
--- a/packages/db/src/runtime/db-client.ts
+++ b/packages/db/src/runtime/db-client.ts
@@ -1,5 +1,5 @@
 import type { InStatement } from '@libsql/client';
-import { createClient, type Config as LibSQLConfig } from '@libsql/client';
+import { type Config as LibSQLConfig, createClient } from '@libsql/client';
 import type { LibSQLDatabase } from 'drizzle-orm/libsql';
 import { drizzle as drizzleLibsql } from 'drizzle-orm/libsql';
 import { type SqliteRemoteDatabase, drizzle as drizzleProxy } from 'drizzle-orm/sqlite-proxy';
@@ -43,15 +43,16 @@ export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string
 	switch (remoteUrl.protocol) {
 		case 'http:':
 		case 'https:':
-			return createStudioDatabaseClient(appToken, remoteUrl)
-    case 'libsql+http:':
-    	remoteUrl.protocol = 'http:';
-    case 'libsql+https:':
-    	remoteUrl.protocol = 'https:';
-    case 'ws:':
-    case 'wss:':
-    case 'file:':
-    case 'libsql:':
+			return createStudioDatabaseClient(appToken, remoteUrl);
+		case 'libsql+http:':
+			remoteUrl.protocol = 'http:';
+		case 'libsql+https:':
+			remoteUrl.protocol = 'https:';
+		case 'ws:':
+		case 'wss:':
+		case 'file:':
+		case 'memory:':
+		case 'libsql:':
 			return createRemoteLibSQLClient(appToken, remoteUrl);
 		default:
 			throw new Error(`Unsupported remote DB: ${remoteDbURL}`);
@@ -62,10 +63,10 @@ function createRemoteLibSQLClient(appToken: string, remoteDbURL: URL) {
 	const options: Partial<LibSQLConfig> = Object.fromEntries(remoteDbURL.searchParams.entries());
 	remoteDbURL.search = '';
 
-	const client = createClient({ 
+	const client = createClient({
 		...options,
 		authToken: appToken,
-		url: remoteDbURL.toString(),
+		url: remoteDbURL.protocol === 'memory:' ? ':memory:' : remoteDbURL.toString(),
 	});
 	const db = drizzleLibsql(client);
 
--- a/packages/db/test/static-remote.test.js
+++ b/packages/db/test/static-remote.test.js
@@ -39,4 +39,31 @@ describe('astro:db', () => {
 			assert.match($('#row').text(), /1/);
 		});
 	});
+
+	describe('static build --remote with custom LibSQL', () => {
+		let remoteDbServer;
+
+		before(async () => {
+			process.env.ASTRO_STUDIO_REMOTE_DB_URL = `memory:`;
+			await fixture.build();
+		});
+
+		after(async () => {
+			await remoteDbServer?.stop();
+		});
+
+		it('Can render page', async () => {
+			const html = await fixture.readFile('/index.html');
+			const $ = cheerioLoad(html);
+
+			assert.equal($('li').length, 1);
+		});
+
+		it('Returns correct shape from db.run()', async () => {
+			const html = await fixture.readFile('/run/index.html');
+			const $ = cheerioLoad(html);
+
+			assert.match($('#row').text(), /1/);
+		});
+	});
 });


Adammatthiesen (commented):
Amazing how simple the changes that are required to unlock the potential of this library, I could see many people very happy about this change! since it will also allow more "Self-hosted-production" database options!

syhily:
I'm very happy and confident that this PR unlocks new use cases for Astro DB. I hope it gets merged soon.

ematipico:
Thank you @Fryuni. I'll make sure that Core is aware of this PR :)

matthewp:
How is this working? Is this because the libsql db has an https URL? Any links you can share? Seems exciting.

packages/db/src/runtime/db-client.ts
```diff
@@ -36,6 +38,42 @@ const remoteResultSchema = z.object({
...
 export function createRemoteDatabaseClient(appToken: string, remoteDbURL: string) {
+	const remoteUrl = new URL(remoteDbURL);
+
+	switch (remoteUrl.protocol) {
+		case 'http:':
+		case 'https:':
+			return createStudioDatabaseClient(appToken, remoteUrl);
+		case 'libsql+http:':
+			remoteUrl.protocol = 'http:';
+		case 'libsql+https:':
```
ascorbic, line 49: The fallthrough here will mean the `http:` case will always be overwritten with `https:`. This is a great example of how easy it is for bugs to sneak in when using switch fallthrough like this. I think it's best if you can refactor this.

Fryuni:
@matthewp 
> How is this working?

It uses the DB itself to manage the snapshots, like Studio did back in the beta.

> Is this because the libsql db has an https URL?

LibSQL has two APIs, one exclusively over [HTTP](https://github.com/tursodatabase/libsql/blob/main/docs/http_api.md) and one using the [Hrana](https://github.com/tursodatabase/libsql/blob/main/docs/HRANA_3_SPEC.md) protocol, which can work both over HTTP or over WebSockets or a combination of both.

The Hrana protocol allows for stateful connections, so transactions are supported.

> Any links you can share? Seems exciting.

Besides the specs above, the supported URL protocols are described in [LibSQL's type docs](https://github.com/tursodatabase/libsql-client-ts/blob/dab55ad1cfed1d90b46dd970c9897f435eb3adcb/packages/libsql-core/src/api.ts#L3-L10). It links to [a section of the README](https://github.com/tursodatabase/libsql-client-ts/blob/60c1d397dcc66729a2bea5770b0a63962bfd7938/README.md#supported-urls) that was removed during some reorganization of the repo a few months ago.

The LibSQL client uses the Hrana protocol. The given URL can have `http`, `https`, `ws` or `wss` as the protocol to select which underlying transport should be used or can be `libsql` and let the library decide which transport to use (currently selects HTTP).

Since Studio already uses `http`/`https` I added the variation of `libsql+http` and `libsql+https` to select a LibSQL remote using an explicit HTTP transport

matthewp:
@Fryuni Fred is out this week, he has the most context for how migrations work, so when he gets back he'll review this and leave any feedback.

matthewp:
@Fryuni Plan is still to have this in for 4.13

sarah11918:
Just noting that if this goes through, we will want docs for how someone can do this in the DB guide!

Fryuni:
!preview libsql

matthewp:
!preview db-remote

.changeset/healthy-boxes-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/db': minor
+---
+
+Adds support for self-hosted LibSQL
```
sarah11918, line 5: I'll come back to this after we know what's going in docs, but for a new feature released in a minor, we typically would hype up the feature a bit here! Something covering the PR description like:

``````suggestion
Adds Astro DB support for a remote LibSQL server instance using any supported LibSQL protocols. This enables Astro DB to be used with self-hosting and air-gapped deployments.

To configure, set `ASTRO_STUDIO_REMOTE_DB_URL`.... (code block example showing where this is set*)
``````
(We'd also need to note whether this is a Studio feature or DB feature)

Comment (Fryuni) [reply .changeset/healthy-boxes-poke.md:5]: > We'd also need to note whether this is a Studio feature or DB feature

It is a DB feature, not a Studio feature. For environments that can't use Studio, it can be replaced with a LibSQL server.

> To configure, set `ASTRO_STUDIO_REMOTE_DB_URL`.... (code block example showing where this is set*)

There is no code block for this. No code needs to be changed at all, just the environment variables.

Comment (sarah11918) [reply .changeset/healthy-boxes-poke.md:5]: OK, then we should show setting/changing this as an environment variable e.g. in a `.env` file? Does the environment variable need to be named `ASTRO_STUDIO_REMOTE_DB_URL` ?

It would be nice to show someone what they need to do to take advantage of this, I think! If you'd prefer to make the docs PR first, so that we have that to pull from to write this changeset, that's fine too!

Comment (Princesseuh) [reply .changeset/healthy-boxes-poke.md:5]: I confirm as the blog writer that an example would be neat.

File: packages/db/src/core/utils.ts [hunk 3]
```diff
@@ -15,6 +15,11 @@ export function getRemoteDatabaseUrl(): string {
 	return env.ASTRO_STUDIO_REMOTE_DB_URL || 'https://db.services.astro.build';
 }
 
+export function isRemoteStudio(url: string): boolean {
+	const protocol = new URL(url).protocol;
+	return protocol === 'http:' || protocol === 'https:';
+}
```
Princesseuh, line 21: Isn't this a bit naive? I thought you could connect to http(s) databases using LibSQL.

Fryuni: I was prioritizing compatibility with any existing internal Studio tooling.

sarah11918:
Hey @Fryuni ! Friendly ping to remind you that I'd normally want to have a docs PR reviewed before approving an astro PR, so I know how to make helpful suggestions to the changeset here! And, this content is what helps Erika write the release blog post so she knows how to hype your feature!

Since PRs will be merged tomorrow for the minor, can I help you in any way to get a docs PR together today?

Fryuni:
I got caught up with some things recently, but I'll have a PR for docs today 😁

.changeset/healthy-boxes-poke.md
sarah11918: Updated my suggestion! What would you like to see different here @Fryuni ?

sarah11918 (approved):
Approving for docs!  🥳

.changeset/healthy-boxes-poke.md
Fryuni: LGTM, @Princesseuh do you think the changeset and docs are good for the blog post or would you prefer to have a project deployed with it to have a link as an example?

Princesseuh: I'll take a bigger look tomorrow, but this looks fine to me!

packages/db/src/core/utils.ts
Princesseuh: Does this mean that to connect to http / https LibSQL instances, you need to do something special specific to Astro? I tried to search for `libsql+https` only but couldn't find any references to anything.

packages/db/src/core/cli/migration-queries.ts
```diff
@@ -422,12 +424,46 @@ function hasRuntimeDefault(column: DBColumn): column is DBColumnWithDefault {
...
+			sql`select snapshot from _astro_db_snapshot order by id desc limit 1;`
+		);
+
+		console.log('DB res:', res);
+
+		return JSON.parse(res.snapshot);
+	} catch (error: any) {
+		if (error.code === 'SQLITE_UNKNOWN') {
+			// Snapshots table was not created yet,
+			return;
```
ematipico, line 455: Can you expand the current comment and explain why this error doesn't need to be handled? Still, I think it's valuable to have a log of the error.

packages/db/src/core/cli/migration-queries.ts
```diff
@@ -422,12 +424,46 @@ function hasRuntimeDefault(column: DBColumn): column is DBColumnWithDefault {
...
+): Promise<DBSnapshot | undefined> {
+	const client = createRemoteDatabaseClient(appToken, remoteUrl);
+
+	try {
+		const res = await client.get<{ snapshot: string }>(
+			// Latest snapshot
+			sql`select snapshot from _astro_db_snapshot order by id desc limit 1;`
+		);
+
+		console.log('DB res:', res);
```
ematipico, line 449: Remove

packages/db/src/core/utils.ts
```diff
@@ -15,6 +15,11 @@ export function getRemoteDatabaseUrl(): string {
 	return env.ASTRO_STUDIO_REMOTE_DB_URL || 'https://db.services.astro.build';
 }
 
+export function isRemoteStudio(url: string): boolean {
+	const protocol = new URL(url).protocol;
```
ematipico, line 19: Is `url` **always** a valid URL?

packages/db/src/runtime/db-client.ts
```diff
@@ -36,6 +38,44 @@ const remoteResultSchema = z.object({
...
+			remoteUrl.protocol = 'https:';
+			return createRemoteLibSQLClient(appToken, remoteUrl);
+		case 'ws:':
+		case 'wss:':
+		case 'file:':
+		case 'memory:':
+		case 'libsql:':
+			return createRemoteLibSQLClient(appToken, remoteUrl);
+		default:
+			throw new Error(`Unsupported remote DB: ${remoteDbURL}`);
```
ematipico, line 60: Why not a `AstroError`

packages/db/src/runtime/db-client.ts
```diff
@@ -36,6 +38,44 @@ const remoteResultSchema = z.object({
...
+}
+
+function createRemoteLibSQLClient(appToken: string, remoteDbURL: URL) {
+	const options: Partial<LibSQLConfig> = Object.fromEntries(remoteDbURL.searchParams.entries());
+	remoteDbURL.search = '';
+
+	const client = createClient({
+		...options,
+		authToken: appToken,
+		url: remoteDbURL.protocol === 'memory:' ? ':memory:' : remoteDbURL.toString(),
```
ematipico, line 71: It's a bit weird to check for `memory:` **again**, since we already do that already inside `createRemoteDatabaseClient`, in the `switch` statement you added. I wonder if we can do the check just once.

packages/db/src/runtime/db-client.ts
```diff
@@ -36,6 +38,44 @@ const remoteResultSchema = z.object({
...
+	});
+	const db = drizzleLibsql(client);
+
+	return db;
```
ematipico, line 75:
```suggestion
 return drizzleLibsql(client);
```

packages/db/src/runtime/db-client.ts
Princesseuh: It's in an integration, can't use `AstroError`

matthewp (dismissed):

packages/db/src/runtime/db-client.ts
ematipico: We can import `AstroUserError` though

packages/db/src/core/utils.ts
matthewp: If there was a different environment variable for this couldn't we avoid this problem?

Princesseuh:
We'll be releasing this outside of the minor, as there's still some concerns to be addressed.

Fryuni:
> It does seem a bit odd to me to reuse the environment variables for Studio to connect to something non-studio. Can we not have different variables for this use-case? Would that be a problem?

I explained that on Discord to Sarah, but it might have gotten lost there:

Since the Studio side is closed source, I don't know what would break internally if I changed the name of the env var. If it were just DB, I'd have renamed it to something like `ASTRO_DB_REMOTE_URL` and `ASTRO_DB_TOKEN`.
So I kept the same name to not break anything in Studio.

I'll refactor the code to accept different values, here is what I'm thinking:
- Add a new `ASTRO_DB_REMOTE_URL`, which would be used just for LibSQL support.
	- If only `ASTRO_STUDIO_REMOTE_URL` is present it will use that URL to connect as a Studio backend.
	- If only `ASTRO_DB_REMOTE_URL` is present, it will use that URL to connect to LibSQL. It would no longer need to modify the scheme.
	- If neither env is present, connector to the prod Studio URL.
	- If both env vars are present we use the Studio one.
- When connecting to Studio, the token is read from `ASTRO_STUDIO_TOKEN`
- When connecting to LibSQL, the token is read form `ASTRO_DB_TOKEN`

This way, it should not change anything about configuring the Studio connection while also adding LibSQL support.

What do you guys thing?

ematipico:
@Fryuni I like the proposal! :)

matthewp:
@Fryuni sorry for the late reply, yes that sounds good! Basically just treat this as a new feature that's unrelated to the Astro Studio flow. So your proposal sounds good.

.changeset/healthy-boxes-poke.md
```diff
@@ -0,0 +1,14 @@
+---
+'@astrojs/db': minor
+---
+
+Adds support for connecting Astro DB to any remote LibSQL server. This allows Astro DB to be used with self-hosting and air-gapped deployments.
+
+To connect Astro DB to a remote LibSQL server instead of Studio, set the following environment variables:
+
+- `ASTRO_STUDIO_REMOTE_DB_URL`: the connection URL to your LibSQL server
+- `ASTRO_STUDIO_APP_TOKEN`: the auth token to your LibSQL server
```
sarah11918, line 10: Just commenting that this will need updating with the new proposal!

sarah11918 (approved):
Still approving for docs, and wondering whether this will be included in 4.15!

.changeset/healthy-boxes-poke.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Adds support for connecting Astro DB to any remote LibSQL server. This allows Astro DB to be used with self-hosting and air-gapped deployments.
+
+To connect Astro DB to a remote LibSQL server instead of Studio, set the following environment variables:
+
+- `ASTRO_DB_REMOTE_URL`: the connection URL to your LibSQL server
+- `ASTRO_DB_APP_TOKEN`: the auth token to your LibSQL server
+
+Details of the LibSQL connection can be configured using the connection URL.
```
sarah11918, line 12:
```suggestion
Details of the LibSQL connection can be configured using the connection URL. For example, `memory:?syncUrl=libsql%3A%2F%2Fdb-server.example.com` would create an in-memory embedded replica for the LibSQL DB on `libsql://db-server.example.com`.
```
I think if you wanted, an example here would be fine to add! I pulled this from the PR description, but you could choose one that you think works for a changeset/blog post example!

matthewp (approved):

ematipico (approved):

PR #11386, opened 2024-07-01
feat: adds simple loader

ascorbic:
## Changes

Adds a simplified syntax for loaders, for cases where no granular updates or delta sync is needed. The loader returns an array of objects, which regenerates the collection from scratch, replacing the existing content. This is meant to be available in addition to the lower level loader API, but I don't know how best to explain this, or if this is the right approach so would welcome feedback. This is the API as implemented in this PR:

```ts
const cats = defineCollection({
	type: "experimental_data",
	loader: async function() {
		return  [{
				"breed": "Siamese",
				"id": "siamese",
				"size": "Medium",
				"origin": "Thailand",
				"lifespan": "15 years",
				"temperament": ["Active", "Affectionate", "Social", "Playful"]
			},
			 {
				"breed": "Persian",
				"id": "persian",
				"size": "Medium",
				"origin": "Iran",
				"lifespan": "15 years",
				"temperament": ["Calm", "Affectionate", "Social"]
			},
			 {
				"breed": "Tabby",
				"id": "tabby",
				"size": "Medium",
				"origin": "Egypt",
				"lifespan": "15 years",
				"temperament": ["Curious", "Playful", "Independent"]
			},
			{
				"breed": "Ragdoll",
				"id": "ragdoll",
				"size": "Medium",
				"origin": "United States",
				"lifespan": "15 years",
				"temperament": ["Calm", "Affectionate", "Social"]
			}
		];
	},
	schema: z.object({
		breed: z.string(),
		id: z.string(),
		size: z.string(),
		origin: z.string(),
		lifespan: z.string(),
		temperament: z.array(z.string())
	}),
})
```

You can compare it with the lower-level loader API here: https://github.com/withastro/astro/blob/content-layer/packages/astro/test/fixtures/content-layer/src/loaders/post-loader.ts

## Testing

Adds a test and updates the fixture

### DIFF
--- a/packages/astro/src/content/loaders.ts
+++ b/packages/astro/src/content/loaders.ts
@@ -86,7 +86,7 @@ export async function syncContentLayer({
 
 			let { schema } = collection;
 
-			if (!schema) {
+			if (!schema && typeof collection.loader === 'object') {
 				schema = collection.loader.schema;
 			}
 
@@ -129,7 +129,7 @@ export async function syncContentLayer({
 				return simpleLoader(collection.loader, payload);
 			}
 
-			if(!collection.loader.load) {
+			if (!collection.loader.load) {
 				throw new Error(`Collection loader for ${name} does not have a load method`);
 			}
 
@@ -144,8 +144,11 @@ export async function syncContentLayer({
 	logger.info('Synced content');
 }
 
-export async function simpleLoader(handler: () => Array<DataWithId>, context: LoaderContext) {
-	const data = handler();
+export async function simpleLoader(
+	handler: () => Array<DataWithId> | Promise<Array<DataWithId>>,
+	context: LoaderContext
+) {
+	const data = await handler();
 	context.store.clear();
 	for (const raw of data) {
 		const item = await context.parseData({ id: raw.id, data: raw });
--- a/packages/astro/src/content/loaders.ts
+++ b/packages/astro/src/content/loaders.ts
@@ -16,6 +16,11 @@ export interface ParseDataOptions {
 	filePath?: string;
 }
 
+export type DataWithId = {
+	id: string;
+	[key: string]: unknown;
+};
+
 export interface LoaderContext {
 	/** The unique name of the collection */
 	collection: string;
@@ -28,9 +33,7 @@ export interface LoaderContext {
 	settings: AstroSettings;
 
 	/** Validates and parses the data according to the collection schema */
-	parseData<T extends Record<string, unknown> = Record<string, unknown>>(
-		props: ParseDataOptions
-	): T;
+	parseData(props: ParseDataOptions): Promise<DataWithId>;
 
 	/** When running in dev, this is a filesystem watcher that can be used to trigger updates */
 	watcher?: FSWatcher;
@@ -97,12 +100,8 @@ export async function syncContentLayer({
 
 			const collectionWithResolvedSchema = { ...collection, schema };
 
-			function parseData<T extends Record<string, unknown> = Record<string, unknown>>({
-				id,
-				data,
-				filePath = '',
-			}: { id: string; data: T; filePath?: string }): T {
-				return getEntryData(
+			const parseData: LoaderContext['parseData'] = ({ id, data, filePath = '' }) =>
+				getEntryData(
 					{
 						id,
 						collection: name,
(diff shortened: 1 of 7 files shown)


packages/astro/test/fixtures/content-layer/src/content/config.ts
```diff
@@ -18,4 +18,52 @@ const dogs = defineCollection({
 		temperament: z.array(z.string())
 	}),
 })
-export const collections = { blog, dogs };
+
+const cats = defineCollection({
+	type: "experimental_data",
+	loader: async function() {
```
ematipico, line 24: Why do we define `async` if the function doesn't return any promise?

packages/astro/src/content/loaders.ts
```diff
@@ -16,6 +16,11 @@ export interface ParseDataOptions {
 	filePath?: string;
 }
 
+export type DataWithId = {
+	id: string;
+	[key: string]: unknown;
```
ematipico, line 21: Here, we define the data as `unknown`, but the zod schema is `z.any`. Is it intentional?

packages/astro/test/fixtures/content-layer/src/content/config.ts
```diff
@@ -18,4 +18,52 @@ const dogs = defineCollection({
...
+				"breed": "Ragdoll",
+				"id": "ragdoll",
+				"size": "Medium",
+				"origin": "United States",
+				"lifespan": "15 years",
+				"temperament": ["Calm", "Affectionate", "Social"]
+			}
+		];
+	},
+	schema: z.object({
```
ematipico, line 59: The schema shapes an entry, and the loader returns an array. I wonder why this is the case. Wouldn't it make more sense to match the schema with the data returned by the loader? Why are we forced to return an array and not a single object?

packages/astro/test/fixtures/content-layer/src/content/config.ts
ascorbic: Just so the fixture tests that promises work

packages/astro/test/fixtures/content-layer/src/content/config.ts
ascorbic: The schema is for each item though. The array is for the whole collection

packages/astro/src/content/loaders.ts
ascorbic: I've tightened up the schema

packages/astro/test/fixtures/content-layer/src/content/config.ts
ematipico: Do you know if it is mandatory? I saw that the types accept returning data that don't return a promise

packages/astro/test/fixtures/content-layer/src/content/config.ts
ascorbic: Do you mean returning the collection as an object with the ids as keys, or something else?

packages/astro/test/fixtures/content-layer/src/content/config.ts
ascorbic: We `await` it [here](https://github.com/withastro/astro/pull/11386/files#diff-fff398c3bcb68ee6e2e9161735922902d26b20550bf701ca1dc745ce7baaef79R151), so it's supported but not mandatory.

packages/astro/test/fixtures/content-layer/src/content/config.ts
ematipico: IMHO, if a `schema` defines an object, I would expect the `loader` to return an object. If a `schema` defines an array, I would expect the `loader` to return an array.

However, if we decide to go with the current design, I think we should rename `schema` and `loader`, for example, `entrySchema` or `entitySchema`, and `entitiesLoader` or `entriesLoader`. 

Anyway, not a blocker, we can merge this PR any time

ematipico (approved):

packages/astro/test/fixtures/content-layer/src/content/config.ts
ascorbic: This is how collections work now though. The schema is for the individual items, not the whole collection array

bluwy (approved):

matthewp (approved):

PR #11398, opened 2024-07-02
feat: add glob loader

ascorbic:
## Changes
Adds a new `glob()` loader to the content layer. Usage is:

```ts
const spacecraft = defineCollection({
	type: "experimental_data",
	loader: glob({ pattern: "*.md", base: "content/space" })
})
```
By default it auto-generates an ID based on a slug of the path, relative to the base. e.g. `content/space/columbia.md` would have the ID `columbia`. However you can pass a `generateId` function and it will allow you to generate your own IDs. The function receives the `base` directory, relative `entry` path and parsed `data` from the entry. e.g.

```ts
const staff = defineCollection({
	type: 'experimental_data',
	loader: glob({
		pattern: '**/*.md',
		base: 'data/staff',
		generateId: ({ entry, base, data }) => {
			if (data.slug) {
				return data.slug as string;
			}
			return `profile/${entry.slice(0, -3)}`;
		},
	}),
});
```

In order to support these changes, there have been a few refactors and additions to the content layer API. One of these is moving the loaders so they are no longer impoirted from `astro:content`, but instead from `astro/loaders`. The loaders use node builtins that shouldn't be used at runtime. Having a separate import makes it easier to keep them separate.


florian-lefebvre:
The base is a string relative to the root right? I think it would be nice to allow (for the file loader as well) absolute paths as string and urls, that may be useful for monorepos or integrations for example

ascorbic:
I _think_ absolute paths should work too. It generates a file URL relative to the root, but an absolute path should work as expected.

florian-lefebvre:
Awesome! Worth adding some tests then I guess

ascorbic:
@florian-lefebvre I've updated a test to use an absolute path and confirmed it works.

matthewp (approved):

PR #11415, opened 2024-07-04
feat(astro): clean sync

### LINKED ISSUE #11380: astro:env does not generate .astro/env.d.ts in develoment mode
### Astro Info

- astro: 4.11.3
- node: 18.18.2

Using last experimental type-safe environment variables features, types are not generated automatically in development mode.

```typescript
import { getSecret } from "astro:env/server";
```

Cannot find module 'astro:env/server' or its corresponding type declarations.ts(2307)

I have to execute

```bash
npm run build
````

to remove the ts error.

_No response_

### Describe the Bug

Code works but intellisense show an error.

Captura de pantalla 2024-06-30 a las 9 34 13

Looking the file `env.d.ts`

```typescript
/// <reference path="../.astro/env.d.ts" />
/// <reference types="astro/client" />
```

new line is added but the file is not generated under  `.astro` folder.

Same happens for backend vars.

Captura de pantalla 2024-06-30 a las 9 45 17

but if you execute `npm run build`, the error dissapears because typings file is generated.

Generate automatically `env.d.ts` file on develoment mode.

```typescript
declare module "astro:env/client" {
  export const API_URL: string;
}

declare module "astro:env/server" {
  export const DB_HOST: string;
  export const DB_NAME: string;
  export const DB_PASS: string;
  export const DB_PORT: number;
  export const DB_USER: string;

  export const getSecret: (key: string) => string | undefined;
}
```

https://stackblitz.com/edit/withastro-astro-ey4hsq?file=.env

### Participation

florian-lefebvre:
## Changes

- Closes #11380
- Supersedes #11419
- Refactors `sync` to run correctly no matter the command

## Testing

Existing tests should still pass

## Docs

N/A


florian-lefebvre:
In the monorepo, running `pnpm --filter @example/basics astro sync` now generates astro db types although the example does not use db...is that because of how `getPackage` works?

packages/astro/src/core/index.ts
```diff
@@ -22,5 +22,5 @@ export const build = (inlineConfig: AstroInlineConfig) => _build(inlineConfig);
  *
  * @experimental The JavaScript API is experimental
  */
-// Wrap `_sync` to prevent exposing the second internal options parameter
-export const sync = (inlineConfig: AstroInlineConfig) => _sync(inlineConfig);
+// Wrap `syncInlineConfig` to prevent exposing internal options
+export const sync = async (inlineConfig: AstroInlineConfig) => syncInlineConfig({ inlineConfig });
```
florian-lefebvre, line 26:
```suggestion
export const sync = (inlineConfig: AstroInlineConfig) => syncInlineConfig({ inlineConfig });
```

matthewp (approved):

packages/astro/src/core/sync/index.ts
```diff
@@ -28,73 +22,92 @@ import {
...
+		telemetry.record(eventCliSession('sync', userConfig));
+	}
+	let settings = await createSettings(astroConfig, inlineConfig.root);
+	settings = await runHookConfigSetup({
+		command: 'build',
+		settings,
+		logger,
+	});
+	return await sync({ settings, logger, fs });
+}
```
bluwy, line 66: IIUC this is the main function to call from the programmatic API (`sync()`). In which case, I think this should be kept as `sync()` name and shouldn't be renamed as `syncInlineConfig()`.

The other default exported `sync()` function should be renamed as something else in that case. And should no longer be default exported since it's not the main `sync` function (the `syncInlineConfig` would be default exported).

packages/astro/src/core/sync/index.ts
```diff
@@ -28,73 +22,92 @@ import {
...
-	inlineConfig: AstroInlineConfig,
-	options?: SyncOptions
-): Promise<ProcessExit> {
+export default async function sync({
+	logger,
+	fs = fsMod,
+	settings,
+	skip,
+}: SyncOptions): Promise<void> {
 	ensureProcessNodeEnv('production');
```
bluwy, line 80: This should be called at the very first before we do any work (as soon as we identify `sync()` or `astro sync` is being called. So this should be moved into the above function.

ematipico (approved):

bluwy (approved):
Amazing. Thanks for updating!

PR #11431, opened 2024-07-08
Pass slots to server islands

matthewp:
## Changes

- Slots are statically rendered into a string and included in the original HTML.
- When an island posts back it includes those slots to the server environment so that it can use them.

## Testing

- E2E test added to verify the full flow works.

## Docs

N/A


Princesseuh (approved):
This looks fine to me as far as I can tell!

PR #11432, opened 2024-07-08
Require the experimental flag to use server islands

matthewp:
## Changes

- Create the experimental flag and only enable the vite plugin when its used.

## Testing

- Updated tests

## Docs

N/A

### DIFF
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -2163,6 +2163,34 @@ export interface AstroUserConfig {
 			 */
 			schema?: EnvSchema;
 		};
+
+		/**
+		 * @docs
+		 * @name experimental.serverIslands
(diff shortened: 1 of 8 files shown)


packages/astro/src/@types/astro.ts
```diff
@@ -2163,6 +2163,34 @@ export interface AstroUserConfig {
...
+		 *   experimental: {
+		 *     serverIslands: true,
+		 *   },
+		 * }
+		 * ```
+		 * 
+		 * Use the `server:defer` directive on any component, Astro or framework, and the component will not render initially.
+		 *
+		 * ```astro "server:defer"
+		 * <Avatar server:defer />
```
florian-lefebvre, line 2188: What happens if the directive is used without the flag? Will there be an error saying users should enable it?

Issue #11430, opened 2024-07-08
`astro:env` not allowing numbers in variable names


NuroDev:
### Astro Info

```block
Astro                    v4.11.3
Node                     v20.12.2
System                   macOS (arm64)
Package Manager          bun
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/tailwind
```

### Describe the Bug

I originally opened this [as a discussion](https://github.com/withastro/roadmap/discussions/972) as I assumed this was intentional but apparently this seems more like a bug.

In short: Currently if you try to add an environment variable with any kind of number in the name it will throw the following error:
```bash
[config] Astro found issue(s) with your configuration:
  ! experimental.env.schema.S3_ACCESS_KEY_ID  A valid variable name can only contain uppercase letters and underscores..
```

For the most part numbers in variables names do not make sense. However in some cases, like the one above with S3 credentials, it would be good to still all numbers in variable names.

### What's the expected result?

For the `A valid variable name can only contain uppercase letters and underscores..` validation error above not to get thrown whenever a environment variable name includes a number.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-pudfny?file=astro.config.ts

### Participation

- [X] I am willing to submit a pull request for this issue.

florian-lefebvre:
Thanks! Yes actually we only want to prevent the first character being a number because it's not a valid identifier. Code happens here:

- https://github.com/withastro/astro/blob/main/packages/astro/src/env/schema.ts#L84
- https://github.com/withastro/astro/blob/main/packages/astro/src/env/schema.ts#L87-L89

It's probably worth
1. Creating a dedicated schema for the key
2. Adding a few tests here https://github.com/withastro/astro/blob/main/packages/astro/test/units/config/config-validate.test.js

NuroDev:
Good idea. For a dedicated Zod schema would it be better here to stick with a Regex pattern, which would likely be faster but performance isn't as critical here, or adding multiple refinements (`.refine()`) to add granular error messages for each requirement?

florian-lefebvre:
Lets go with refine calls! I'm in favor of clearer errors

PR #11437, opened 2024-07-09
feat: astro:env allow schema keys to include numbers

NuroDev:
This PR closes & fixes the bug reported in #11430 

## Changes

- Updated the `EnvSchema` validators regex pattern to also allow for numbers
- Updated the `EnvSchema` validator to reject if the first character is a number (Not a valid identifier)

## Testing

 - Added 2 new unit tests to account for the above changes
	 - `Should allow schema variables with numbers`
	 - `Should not allow schema variables starting with a number`

## Docs

Added changeset & did not update any docs as currently there is no documentation to modify informing users that numbers were not previously allowed.

Though I am unsure if it is worth adding a comment for users to be aware of this, or just let the edge case error message inform them?

### DIFF
--- a/.changeset/loud-socks-doubt.md
+++ b/.changeset/loud-socks-doubt.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Allow env schema variables to include numbers, except as a first character
+Fixes a case where Astro's config `experimental.env.schema` keys did not allow numbers. Numbers are still not allowed as the first character to be able to generate valid JavaScript identifiers
--- a/.changeset/loud-socks-doubt.md
+++ b/.changeset/loud-socks-doubt.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Allow env schema variables to include numbers, except as a first character
--- a/packages/astro/src/env/schema.ts
+++ b/packages/astro/src/env/schema.ts
@@ -81,18 +81,18 @@ const EnvFieldMetadata = z.union([
 	SecretServerEnvFieldMetadata,
 ]);
 
-const KEY_REGEX = /^[A-Z0-9_]+$/;
+const EnvSchemaKey = z
+	.string()
+	.min(1)
+	.refine(([firstChar]) => isNaN(Number.parseInt(firstChar)), {
+		message: 'A valid variable name cannot start with a number.',
+	})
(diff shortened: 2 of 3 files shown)


.changeset/loud-socks-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Allow env schema variables to include numbers, except as a first character
```
florian-lefebvre, line 5:
```suggestion
Fixes a case where Astro's config `experimental.env.schema` keys did not allow numbers. Numbers are still not allowed as the first character to be able to generate valid JavaScript identifiers
```

packages/astro/src/env/schema.ts
```diff
@@ -81,12 +81,18 @@ const EnvFieldMetadata = z.union([
 	SecretServerEnvFieldMetadata,
 ]);
 
-const KEY_REGEX = /^[A-Z_]+$/;
+const KEY_REGEX = /^[A-Z0-9_]+$/;
```
florian-lefebvre, line 84:
```suggestion
const EnvSchemaKey = z
	.string()
	.min(1)
	.refine(([firstChar]) => isNaN(Number.parseInt(firstChar)), {
		message: 'A valid variable name cannot start with a number.',
	})
	.refine((str) => /^[A-Z0-9_]+$/.test(str), {
		message: 'A valid variable name can only contain uppercase letters, numbers and underscores.',
	});
```

packages/astro/src/env/schema.ts
```diff
@@ -81,12 +81,18 @@ const EnvFieldMetadata = z.union([
...
-	}),
+	z
+		.string()
+		.min(1)
+		.refine(([firstChar]) => isNaN(Number.parseInt(firstChar)), {
+			message: 'A valid variable name cannot start with a number.',
+		})
+		.refine((str) => KEY_REGEX.test(str), {
+			message: 'A valid variable name can only contain uppercase letters, numbers and underscores.',
+		}),
```
florian-lefebvre, line 95:
```suggestion
	EnvSchemaKey
```

florian-lefebvre (commented):
Looking good thank you! Left a small nitpicks

florian-lefebvre (approved):

PR #11439, opened 2024-07-09
Actions: expand `isInputError` to accept `unknown`

bholmesdev:
## Changes

Expand `isInputError` to accept `unknown` as a type, so it can narrow any error object to an `ActionInputError`. This is [per feedback in the RFC.](https://github.com/withastro/roadmap/pull/912#issuecomment-2211537388)

- Add type override for `unknown`
- Document the purpose of `ErrorInferenceObject`. The reason for this utility type was unclear originally

## Testing

Manually tested that `isInputError` still type narrows and preserves the `error.fields` types

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -120,6 +120,7 @@
     "test:e2e:match": "playwright test -g",
     "test:e2e:chrome": "playwright test",
     "test:e2e:firefox": "playwright test --config playwright.firefox.config.js",
+    "test:types": "tsc --project tsconfig.tests.json",
     "test:node": "astro-scripts test \"test/**/*.test.js\""
   },
   "dependencies": {
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -215,6 +215,7 @@
     "astro-scripts": "workspace:*",
     "cheerio": "1.0.0-rc.12",
     "eol": "^0.9.1",
+    "expect-type": "^0.19.0",
     "mdast-util-mdx": "^3.0.0",
     "mdast-util-mdx-jsx": "^3.1.2",
     "memfs": "^4.9.3",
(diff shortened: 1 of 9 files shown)


matthewp (approved):
lgtm. It can also be nice to still put these in `define/it` blocks so you can have multiple tests in the same file. Even though it can feel a little weird given that the code doesn't run, it can be nice for organization

bholmesdev:
@matthewp Yeah, I wasn't sure if that would be misleading for new contributors given the tests don't run? I agree the comment separation isn't break. I'll try that since it sounds like you'd prefer it

packages/astro/package.json
```diff
@@ -114,12 +114,13 @@
...
     "postbuild": "astro-scripts copy \"src/**/*.astro\" && astro-scripts copy \"src/**/*.wasm\"",
-    "test": "pnpm run test:node",
+    "test": "pnpm run test:node && pnpm run test:types",
     "test:match": "pnpm run test:node --match",
     "test:e2e": "pnpm test:e2e:chrome && pnpm test:e2e:firefox",
     "test:e2e:match": "playwright test -g",
     "test:e2e:chrome": "playwright test",
     "test:e2e:firefox": "playwright test --config playwright.firefox.config.js",
+    "test:types": "tsc --project tsconfig.tests.json",
     "test:node": "astro-scripts test \"test/**/*.test.js\""
```
matthewp, line 124: Can you update this so that node:test doesn't run these tests?

matthewp (changes requested):

packages/astro/package.json
bholmesdev: Oh, do you mean exclude the `test/types` folder from what `node:test` crawls?

bholmesdev: Or do you mean remove the call from here? https://github.com/withastro/astro/pull/11439/files#diff-31a85c0b4c5526a896f02b675a49f1f942ec15fa815f6326edc4f74248aa0048R117

I'd like our CI to run these tests, and I thought including in `test` (but excluding in `node:test`) was the right move.

PR #11440, opened 2024-07-09
feat: add markdown rendering to content layer

ascorbic:
Adds support for rendering content. Initially just supports markdown.

This works by rendering the markdown before adding it to the store. A `ContentEntryType` gets a new optional field `getRenderFunction`, which returns a reusable render function that accepts an `entry` object and returns an object with `html` and `metadata` fields. The data store `set` method now accepts `rendered` as an optional field, which stores these with the rest of the entry.

Users can access the HTML and metadata directly on `entry.rendered`. Calling `entry.render()` will return a `Content` component like now, but in this case the markdown is already rendered to HTML, so its only transform at that point is to replace image references.

To make it easier to optimise rendering, the data store gains a `digest` field. This is considered an opaque value, but if passed it will be stored and returned in the entry, and if it matches an existing entry in the store then it will skip setting it in the DB. The digest is also used to skip rendering markdown if the text is unchanged. 

I have added benchmarks to compare rendering large numbers of markdown docs using content collections v1, and content layer. 

This PR includes all of the changes from #11469, so those who reviewed that will see some familiar parts. I asked for that one to be reviewed first because it had some changes to the behaviour of this one.


benchmark/bench/memory.js
```diff
@@ -18,7 +18,7 @@ export async function run(projectDir, outputFile) {
 	const outputFilePath = fileURLToPath(outputFile);
 
 	console.log('Building and benchmarking...');
-	await execaCommand(`node --expose-gc --max_old_space_size=256 ${astroBin} build`, {
+	await execaCommand(`node --expose-gc --max_old_space_size=10000 ${astroBin} build`, {
```
ascorbic, line 21: Needs more memory to render 10k pages with cc v1.

packages/astro/src/content/runtime.ts
```diff
@@ -315,6 +318,15 @@ type RenderResult = {
...
 
+async function renderEntry(entry?: DataEntry) {
+	const Content = createComponent({
+		factory() {
+			return createHeadAndContent('', renderTemplate`${entry?.rendered?.html ?? ''}`);
+		},
+	});
+	return { Content };
+}
+
```
ascorbic, line 329: I have no idea if this is a good way of doing this!

packages/astro/src/content/loaders/glob.ts
```diff
@@ -5,9 +5,10 @@ import { green } from 'kleur/colors';
 import micromatch from 'micromatch';
 import pLimit from 'p-limit';
 import { relative } from 'path/posix';
-import type { ContentEntryType } from '../../@types/astro.js';
+import type { ContentEntryType, RenderFunction } from '../../@types/astro.js';
 import { getContentEntryIdAndSlug, getEntryConfigByExtMap } from '../utils.js';
 import type { Loader, LoaderContext } from './types.js';
+import xxhash from 'xxhash-wasm';
```
ascorbic, line 11: It might be worth providing a `generateDigest` utility, based on this.

packages/astro/src/content/loaders/glob.ts
```diff
@@ -63,41 +64,75 @@ export function glob(globOptions: GlobOptions): Loader {
...
-
-		fileToIdMap.set(filePath, id);
-	}
-
 	return {
 		name: 'glob-loader',
 		load: async (options) => {
-			const { settings, logger, watcher } = options;
+			const { settings, logger, watcher, parseData, store } = options;
+			const { h64ToString } = await xxhash();
```
ascorbic, line 71: It might be worth providing a `generateDigest` utility, based on this.

bholmesdev: Unfamiliar with xxhash. Is that preferred over web crypto or node crypto for cross-platform compat?

packages/astro/src/content/loaders/glob.ts
```diff
@@ -63,41 +64,75 @@ export function glob(globOptions: GlobOptions): Loader {
...
+				const filePath = fileURLToPath(fileUrl);
+
+				const parsedData = await parseData({
+					id,
+					data,
+					filePath,
+				});
+
+				if (entryType.getRenderFunction) {
+					let render = renderFunctionByContentType.get(entryType);
```
bholmesdev, line 110: It seems we have two ways to access the render function:
- `renderFunctionByContentType.get(entryType)`
- `entryType.getRenderFunction(settings)`

What was the motivation to have both?

packages/astro/src/content/runtime.ts
bholmesdev: Ha, seems like a minimal implementation to me! This would remove support for the `layout` frontmatter option in content collections, but that has been a maintenance burden for us. Happy to simplify for v2

packages/astro/src/vite-plugin-markdown/content-entry-type.ts
```diff
@@ -15,4 +16,22 @@ export const markdownContentEntryType: ContentEntryType = {
 	},
 	// We need to handle propagation for Markdown because they support layouts which will bring in styles.
 	handlePropagation: true,
+
+	async getRenderFunction(settings) {
+		const processor = await createMarkdownProcessor(settings.config.markdown);
```
bholmesdev, line 21: Wow, this is simpler than I expected. This works great.

Also want to put this on your radar: we have a specialized field called `remarkPluginFrontmatter` that includes any data that's computed by a remark plugin. [A common use case is "reading time,"](https://docs.astro.build/en/recipes/reading-time/) which extracts text nodes from remark to compute minutes read. Is this covered by `result.metadata` potentially? Or is this a consideration we would need to tackle separately, if we support it?

packages/astro/src/content/loaders/glob.ts
ascorbic: It's a non-cryptographic hash, which is orders of magnitude faster than crypto hash functions. It's designed specifically for this kind of case. https://github.com/sebastian-software/node-hash-comparison

bholmesdev:
!preview contentlayer

packages/astro/src/content/data-store.ts
```diff
@@ -183,12 +294,46 @@ export class DataStore {
 export interface ScopedDataStore {
 	get: (key: string) => DataEntry | undefined;
 	entries: () => Array<[id: string, DataEntry]>;
-	set: (key: string, data: Record<string, unknown>, body?: string, filePath?: string) => void;
+	/**
+	 * Adds a new entry to the store. If an entry with the same ID already exists,
+	 * it will be replaced.
```
ematipico, line 299: We need to document the returned value and explain what `true` and `false` mean. Especially `false`, where the `digest` seems involved.

packages/astro/src/content/loaders/glob.ts
ematipico: I am not sure I understand the implications of this choice. Can we add some tech documentation somewhere that explains when to use one and when to use the other?

packages/astro/src/content/loaders/glob.ts
```diff
@@ -63,41 +62,87 @@ export function glob(globOptions: GlobOptions): Loader {
...
+
+			const untouchedEntries = new Set(store.keys());
+
+			async function syncData(entry: string, base: URL, entryType?: ContentEntryType) {
+				if (!entryType) {
+					logger.warn(`No entry type found for ${entry}`);
+					return;
+				}
+				const fileUrl = new URL(entry, base);
+				const contents = await fs.readFile(fileUrl, 'utf-8');
```
ematipico, line 78: Possible runtime error to catch (file doesn't exist, or no rights to read the file)

packages/astro/src/content/loaders/glob.ts
```diff
@@ -63,41 +62,87 @@ export function glob(globOptions: GlobOptions): Loader {
...
+						// Cache the render function for this content type, so it can re-use parsers and other expensive setup
+						renderFunctionByContentType.set(entryType, render);
+					}
+					const rendered = await render({
+						id,
+						data: parsedData,
+						body,
+						filePath,
+						digest,
+					});
```
ematipico, line 128: A render function could throw some errors that we should catch/log, correct?

packages/astro/src/content/runtime.ts
```diff
@@ -315,6 +337,87 @@ type RenderResult = {
...
+		}
+	});
+}
+
+async function renderEntry(entry?: DataEntry) {
+	const html =
+		entry?.rendered?.metadata?.imagePaths?.length && entry.filePath
+			? await updateImageReferencesInBody(entry.rendered.html, entry.filePath)
+			: entry?.rendered?.html;
+
```
ematipico, line 416: The value of `html` can be `undefined`. Is it possible to have this value? If not, should we throw and error? If yes, should we optimise it? (it feels weird to call `createComponent`, `serverRender` and `unescapeHTML` on something that doesn't need it)

ascorbic: What would be the best way to return a component that renders nothing?

ematipico: It's possible to shortcircuit it by returning an object like this:

```js
const ComponentInstance = {
	default() {
		return new Response(null, {
			status: 200,
		});
	},
}
```

However, here we are inside the `Content` component, so I need to think about it more. Consider this as an optimisation we can do later

ascorbic: Thanks

packages/astro/src/content/loaders/glob.ts
ascorbic: To be clear: this isn't two different APIs. The API is `entryType.getRenderFunction(settings)`, and I'm just memoising it in a Map.

ematipico (approved):
I am a bit confused on how the PRs have been handled. I was under the impression that this PR would have been merged **before** the reference PR and the assets PR.

Anyway, looks good to me. I wonder if we could avoid the `console.warn` and use our logger

ascorbic:
@ematipico While normally I'd merge the parent branch first, in this case it had been open for quite a while with a lot of work in the child branches since. I needed the assets one merging first because it included some changes to the logic in here that would mean that otherwise you'd be reviewing outdated code.

PR #11443, opened 2024-07-09
feat: `ActionReturnType`

bholmesdev:
## Changes

Expose an `ActionReturnType` utility to infer the result of an action. This is simpler than the usual `Awaited<ReturnType<typeof action>>`.

## Testing

Added type test

## Docs

See changeset


florian-lefebvre (approved):
Love it!

PR #11444, opened 2024-07-10
chore: don't log more than 100 pages per route

ascorbic:
## Changes

Currently we log every page generated, which is fine in most cases, but for very large sites is useless, makes the logs hard to use and can affect performance. This PR caps this to 100 pages per route, after which it displays a summary instead.

## Testing

Most easily tested with the benchmark scripts.

### DIFF
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -237,6 +237,10 @@ async function generatePage(
 		const paths = await getPathsForRoute(route, pageModule, pipeline, builtPaths);
 		let timeStart = performance.now();
 		let prevTimeEnd = timeStart;
+
+		const shouldTruncate =
+			pipeline.logger.options.level !== 'debug' && paths.length > maxPathsToLog;
+
 		for (let i = 0; i < paths.length; i++) {
 			const path = paths[i];
 			pipeline.logger.debug('build', `Generating: ${path}`);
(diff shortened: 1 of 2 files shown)


packages/astro/src/core/build/generate.ts
```diff
@@ -223,6 +223,9 @@ async function generatePage(
 		styles,
 		mod: pageModule,
 	};
+
+	const maxPathsToLog = 100;
```
Princesseuh, line 227: You might want to disable this when verbose mode is enabled. I'm not 100% sure how you get it from here, but a search for `verbose` should hopefully find an example

florian-lefebvre (commented):
Idk, I found it useful when having some weird errors to log on each page and see what slug was failing so not showing them all would make this harder. Maybe a possibility would be to not log more than 100 by default and have an option to have the full log (using `--verbose` or equivalent). Don't forget the changeset!

packages/astro/src/core/build/generate.ts
ascorbic: Done!

ematipico:
This is "potentially" a breaking change because it changes a lot the DX for people. It could potentially be for the docs website, and for people that rely on this log for checking the progress of the build. (I do, in the Biome website)

Also, I wouldn't want to use `--verbose` because that flag literally litters the console, making it difficult to see where the log of the page occurred.

packages/astro/src/core/build/generate.ts
```diff
@@ -234,18 +237,39 @@ async function generatePage(
 		const paths = await getPathsForRoute(route, pageModule, pipeline, builtPaths);
 		let timeStart = performance.now();
 		let prevTimeEnd = timeStart;
+
+		const shouldTruncate =
+			pipeline.logger.options.level !== 'debug' && paths.length > maxPathsToLog;
```
ascorbic, line 242: Log all pages if when verbose mode is enabled

ascorbic:
@ematipico do you think this shouldn;t be done at all, or that it should be in a major, or that there should be a specific config setting to enable or disable it?

ascorbic:
Sounds like this is a whole can of worms and not really worth the change!

ematipico:
I think it's a valuable change, as you said logging can slow down the build and I agree 100%.

I think we should evaluate this change for 5.0, and worth considering a new flag to log all pages.

I know that Fred has strong feelings around logging, so we should make sure to run the suggestion through him

bholmesdev:
I also agree logs can get very noisy about a certain threshold, but I'd want to see which page failed if possible during a build. If the current page could be bubbled into the error logs, I'd be +1 to this change!

bluwy:
I'd lean on not changing this at all to prevent the risk of possibly not logging the current page if an error occurred. Error handling could be tricky and happen anywhere, so keeping the logging as simple as possible would be safest for me.

Princesseuh:
Closing since we won't be doing this, I'm sure improvements could be considered outside of this, though.

PR #11469, opened 2024-07-15
feat: image support for content layer

ascorbic:
## Changes

Adds support for images to content layer. They are implemented both in markdown body and frontmatter.

## Testing

Added various image types to the content layer fixture. Also added an image to the benchmark.


packages/astro/package.json
```diff
@@ -169,6 +169,7 @@
     "magic-string": "^0.30.10",
     "micromatch": "^4.0.7",
     "mrmime": "^2.0.0",
+    "neotraverse": "^0.6.9",
```
ascorbic, line 172: A you-know-who-free traverse lib

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,89 @@ type RenderResult = {
...
+			.join(' ');
+	});
+}
+
+function updateImageReferencesInData<T extends Record<string, unknown>>(
+	data: T,
+	fileName: string,
+	imageAssetMap: Map<string, ImageMetadata>
+): T {
+	return new Traverse(data).map(function (ctx, val) {
```
ascorbic, line 395: I benchmarked this with 10k entries and it didn't add any noticable time to the processing.

packages/astro/src/content/utils.ts
```diff
@@ -195,6 +205,14 @@ export async function getEntryData(
 			throw formattedError;
 		}
 	}
+	if (imageImports.size > 0) {
+		Object.defineProperties(data, {
+			_internal_imageImports: {
```
ascorbic, line 210: Bit of a hack, admittedly, but it avoids needing to change the usage of `getEntryData`

packages/astro/src/assets/utils/resolveImports.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function imageSrcToImportId(imageSrc: string, filePath: string): string | undefined {
+	// If the import is coming from the data store it will have a special prefix to identify it
+	// as an image import. We remove this prefix so that we can resolve the image correctly.
+	if (imageSrc.startsWith(IMAGE_IMPORT_PREFIX)) {
+		imageSrc = imageSrc.slice(IMAGE_IMPORT_PREFIX.length);
+	}
+
+	// We only care about local imports
+	if (['http', 'https'].includes(imageSrc.split(':')[0]) || imageSrc.startsWith('/')) {
```
Princesseuh, line 20: There's a `isRemotePath` util you can use here

https://github.com/withastro/astro/blob/16ed760f99bfd8da05473548ccc15a1a045b238c/packages/internal-helpers/src/path.ts#L88****

packages/astro/src/assets/utils/resolveImports.ts
```diff
@@ -0,0 +1,40 @@
+import { resolve } from 'node:path/posix';
```
Princesseuh, line 1: Could you move this file to a `node` folder? Either in this folder or at the root of `src/assets`. We somewhat try to do this pattern where possible to avoid someone importing something that imports Node stuff by accident. Some of the utils in this folder are exposed to users, so the distinction is useful

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,91 @@ type RenderResult = {
...
+
+		if (!image) {
+			return full;
+		}
+
+		return Object.entries({
+			...image.attributes,
+			src: image.src,
+			srcset: image.srcSet.attribute,
+			alt: decodedImagePath.alt,
```
Princesseuh, line 385: Won't the `alt` be part of `image.attributes`?

ascorbic: No, because getImage is only passed the ESM import. Is that likely to be a problem? I wanted to be able to share one import between all of the instances of the image on the page, even if they had different alts.

ascorbic:
!preview content-layer-images

packages/astro/src/content/data-store.ts
```diff
@@ -77,7 +87,59 @@ export class DataStore {
 		return this.#collections;
 	}
 
-	#saveToDiskDebounced = () => {
+	addAssetImport(assetImport: string, filePath: string) {
+		const id = imageSrcToImportId(assetImport, filePath);
```
Princesseuh, line 91: Is there a world where this is "asset kind"-agnostic? We sometimes get requests for supporting things like videos and general related assets

Princesseuh (approved):
Looks great to me! Some things are hacky, sure, but they were already hacky. If both of our brains couldn't make it not hacky, perhaps that's just how it has to be. 

Awesome comments everywhere too, great work!

packages/astro/src/content/runtime.ts
```diff
@@ -64,11 +68,19 @@ export function createGetCollection({
...
 		} else if (store.hasCollection(collection)) {
-			return [...store.values<DataEntry>(collection)].map((entry) => ({
-				...entry,
-				collection,
-				render: () => renderEntry(entry),
-			}));
+			// @ts-expect-error	virtual module
+			const { default: imageAssetMap } = await import('astro:asset-imports');
+
+			return [...store.values<DataEntry>(collection)].map((entry) => {
```
ematipico, line 74: nit: maybe we should consider using a `for..of` loop instead of creating a temporary array via spread, and then map it.

packages/astro/src/content/runtime.ts
```diff
@@ -64,11 +68,19 @@ export function createGetCollection({
...
+				if (entry.filePath) {
+					entry.data = updateImageReferencesInData(entry.data, entry.filePath, imageAssetMap);
```
ematipico, line 76: Isn't this mutating the original entry?

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,84 @@ type RenderResult = {
 	remarkPluginFrontmatter: Record<string, any>;
 };
 
+const IMAGE_REGEX = /__ASTRO_IMAGE_="([^"]+)"/g;
```
ematipico, line 337: nit: `CONTENT_LAYER_IMAGE_REGEX`

Mainly to give a more descriptive name, so we know it's not a "generic" image.

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,84 @@ type RenderResult = {
...
+
+	const imageObjects = new Map<string, GetImageResult>();
+
+	// @ts-expect-error Virtual module resolved at runtime
+	const { getImage } = await import('astro:assets');
+
+	// First load all the images. This is done outside of the replaceAll
+	// function because getImage is async.
+	for (const [_full, imagePath] of html.matchAll(IMAGE_REGEX)) {
+		const decodedImagePath = JSON.parse(imagePath.replace(/&#x22;/g, '"'));
```
ematipico, line 351: A couple of things:
- the regex `/&#x22;/g` can be placed in a top-level constant; no need to create a new regex inside a loop. 
- `JSON.parse` *can fail* and we should handle the error

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,84 @@ type RenderResult = {
...
+		const id = imageSrcToImportId(decodedImagePath.src, fileName);
+
+		if (!id || imageObjects.has(id)) {
+			continue;
+		}
+
+		const imported = imageAssetMap.get(id);
+		if (!imported) {
+			continue;
+		}
```
ematipico, line 361:
```suggestion
		const imported = imageAssetMap.get(id);
		if (!id || imageObjects.has(id) || !imported) {
			continue;
		}
```

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +334,84 @@ type RenderResult = {
...
+		const image: GetImageResult = await getImage({ ...decodedImagePath, src: imported });
+		imageObjects.set(imagePath, image);
+	}
+
+	return html.replaceAll(IMAGE_REGEX, (full, imagePath) => {
```
ematipico, line 366: Is there a reason why loop over `html` twice - `matchAll` and `replaceAll`?  If there's a valid reason, I believe we should document it with a comment.

packages/astro/src/content/utils.ts
ematipico: Can you leave a comment that explains the hack and why?

packages/astro/src/assets/utils/resolveImports.ts
```diff
@@ -0,0 +1,33 @@
+import { isRemotePath, removeBase } from '@astrojs/internal-helpers/path';
+import { CONTENT_IMAGE_FLAG, IMAGE_IMPORT_PREFIX } from '../../content/consts.js';
+import { shorthash } from '../../runtime/server/shorthash.js';
+import { VALID_INPUT_FORMATS } from '../consts.js';
+
+/**
+ * Resolves an image src from a content file (such as markdown) to an import id.
```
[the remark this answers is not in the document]
ascorbic: It's the module ID in the rollup/Vite meaning

packages/astro/src/content/data-store.ts
```diff
@@ -77,7 +87,59 @@ export class DataStore {
...
+		// Import the assets, with a symbol name that is unique to the import id. The import
+		// for each asset is an object with path, format and dimensions.
+		// We then export them all, mapped by the import id, so we can find them again in the build.
+		const imports: Array<string> = [];
+		const exports: Array<string> = [];
+		this.#assetImports.forEach((id) => {
+			const symbol = importIdToSymbolName(id);
+			imports.push(`import ${symbol} from '${id}';`);
+			exports.push(`[${JSON.stringify(id)}, ${symbol}]`);
+		});
```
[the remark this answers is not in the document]
ascorbic: This is a Set, so there's no `.reduce` method. Of course I could convert it to an array, but a hill that I will die on is that if you're not using a resusable reducer function or something like that, pushing onto an array is almost always easier to read than using array.reduce.

packages/astro/src/content/runtime.ts
ascorbic: Yeah, the comment above: replaceAll can't have an async replacer, so first I get all matches and use that to asynchrounously get the image objects. I then synchronously replace them. I'm not sure of a way to do that in one step.

packages/astro/src/content/data-store.ts
ascorbic: Yeah, I think it could, but I think we'd need to work out a way to separate them, as there's quite a bit of Astro Image-specific code in all of the asset handling stuff

packages/astro/src/content/data-store.ts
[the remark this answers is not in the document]
ematipico: What?? `Set` supports `forEach` but not `reduce`?? TIL (and it's sad 😅 )

packages/astro/src/assets/utils/resolveImports.ts
ascorbic: I realised this is needed in the runtime, so I've removed all the node imports

packages/astro/src/content/data-store.ts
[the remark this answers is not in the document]
ascorbic: Not map either! Just forEach, entries, keys and values iirc

packages/astro/src/content/data-store.ts
```diff
@@ -77,7 +87,59 @@ export class DataStore {
...
+		const code = /* js */ `
+${imports.join('\n')}
+export default new Map([${exports.join(', ')}]);
+		`;
+		await fs.writeFile(filePath, code);
+	}
+
+	#writeAssetsImportsDebounced() {
+		this.#assetsDirty = true;
```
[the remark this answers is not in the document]
ascorbic: :facepalm:

packages/astro/src/content/runtime.ts
ascorbic: I think this could actually just use `replaceAll` with a string. I nicked this from the Astro Image code, and probably missed something that needed it before.

packages/astro/src/content/runtime.ts
ascorbic: Ah, I was thinking that Map.get wouldn't like undefined, but of course that's not a problem

packages/astro/src/content/data-store.ts
```diff
@@ -216,6 +297,21 @@ export interface ScopedDataStore {
...
+	 */
+	addAssetImports: (assets: Array<string>, fileName: string) => void;
+	/**
+	 * Adds a single asset to the store. This asset will be transformed
+	 * by Vite, and the URL will be available in the final build.
+	 * @param assetImport
+	 * @param fileName
+	 * @returns
+	 */
+	addAssetImport: (assetImport: string, fileName: string) => void;
```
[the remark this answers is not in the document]
ascorbic: The only reason I did it is because adding several at once is goign to be the most common case, but I get your point.

packages/astro/src/content/runtime.ts
ascorbic: No, it uses `Traverse.map`, which returns a new object. I could use the method that mutates it in place, but that gives me an instinctive ick. It might probably save memory though, but it's probably marginal.

packages/astro/src/content/utils.ts
ascorbic: Looking at the hack again made me sad, so I've removed it and made a new method.

packages/astro/src/content/data-store.ts
```diff
@@ -77,7 +87,59 @@ export class DataStore {
...
+			exports.push(`[${JSON.stringify(id)}, ${symbol}]`);
+		});
+		const code = /* js */ `
+${imports.join('\n')}
+export default new Map([${exports.join(', ')}]);
+		`;
+		await fs.writeFile(filePath, code);
+	}
+
+	#writeAssetsImportsDebounced() {
```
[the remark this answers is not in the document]
ascorbic: I've added a comment above, at the point it's called

packages/astro/src/content/runtime.ts
ascorbic: Ah, got it. I'll change that.

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +337,84 @@ type RenderResult = {
 	remarkPluginFrontmatter: Record<string, any>;
 };
 
+const CONTENT_LAYER_IMAGE_REGEX = /__ASTRO_IMAGE_="([^"]+)"/g;
```
ematipico, line 340: nit: we could use `IMAGE_IMPORT_PREFIX` to reduce repetition

packages/astro/src/content/runtime.ts
```diff
@@ -319,9 +337,84 @@ type RenderResult = {
...
+
 async function renderEntry(entry?: DataEntry) {
-	console.log('entry:', entry?.rendered?.metadata);
-	const Content = createComponent(() => serverRender`${unescapeHTML(entry?.rendered?.html)}`);
+	const html =
+		entry?.rendered?.metadata?.imagePaths?.length && entry.filePath
+			? await updateImageReferencesInBody(entry.rendered.html, entry.filePath)
+			: entry?.rendered?.html;
+
+	const Content = createComponent(() => serverRender`${unescapeHTML(html)}`);
```
ematipico, line 417: nit: it does seem to be an overkill to create a component factory if `html` is `undefined`

ematipico (approved):

packages/astro/src/content/runtime.ts
ascorbic: What would be the right approach for an empty component?

Issue #11480, opened 2024-07-17
Build broken in Astro 4.11.6


wheresrhys:
### Astro Info

```block
Astro                    v4.11.6
Node                     v18.20.4
System                   macOS (arm64)
Package Manager          unknown
Output                   static
Adapter                  none
Integrations             @astrojs/starlight

I'm also seeing the issue in github actions, where I am unable to run `astro info`
```

### Describe the Bug

Some dependencies appear to be missing. See this build output https://github.com/wheresrhys/fetch-mock/actions/runs/9975537393/job/27565585726

### What's the expected result?

When pinning to 4.11.5 I get a successful build https://github.com/wheresrhys/fetch-mock/actions/runs/9975667348/job/27565992625

### Link to Minimal Reproducible Example

https://github.com/wheresrhys/fetch-mock/actions/runs/9975537393/job/27565585726

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
I had a quick look at your action, and it seems that was triggered after you update the package manager. Is it possible that some lock file wasn't updated? Or that the action is buggy? https://github.com/wheresrhys/fetch-mock/commit/ffab4b6f5795de6179ad425bc9fc273ef457f9fd

I did update a repo of mine that relies on starlight with that latest of Astro and I didn't see any issue.

Piotr-Aueternum:
> I had a quick look at your action, and it seems that was triggered after you updated the package manager. Is it possible that some lock file wasn't updated? Or that the action is buggy? [wheresrhys/fetch-mock@ffab4b6](https://github.com/wheresrhys/fetch-mock/commit/ffab4b6f5795de6179ad425bc9fc273ef457f9fd)
> 
> I did update a repo of mine that relies on starlight with that latest of Astro and I didn't see any issue.

I started new project and I'm having same issue with 4.11.6. In other hand 4.11.5 works fine.
windows 11
node 18/20
npm 9.6.7
```
{
  "name": "test",
  "type": "module",
  "version": "0.0.1",
  "scripts": {
    "dev": "astro dev",
    "start": "astro dev",
    "build": "astro check && astro build",
    "preview": "astro preview",
    "astro": "astro"
  },
  "dependencies": {
    "@astrojs/check": "^0.8.2",
    "@astrojs/mdx": "^3.1.2",
    "@astrojs/rss": "^4.0.7",
    "@astrojs/sitemap": "^3.1.6",
    "astro": "^4.11.5",
[...]

PR #11483, opened 2024-07-17
fix: pin boxen to avoid JSON import

Princesseuh:
## Changes

Boxen 8 has a dependency that does a JSON import using an import attribute. This is fine, however our lowest supported versions of Node 18 and 20 don't support import attributes just yet.

Fix https://github.com/withastro/astro/issues/11480

## Testing

Tested manually

## Docs

N/A


ematipico (approved):

ematipico (commented):
We should update the renovate bot to not provide another PR to update it

Semkoo:
> We should update the renovate bot to not provide another PR to update it

I think by not placing "^" in `"boxen": "7.1.1",` the bot should ignore no?

bluwy:
Ah thanks for figuring this out. Looks like boxen's new version ranges isn't correct 🤔

PR #11489, opened 2024-07-18
fix: make root serializable

ematipico:
## Changes

- Cloudflare doesn't support `import.meta.url`. This functionality is only used in dev mode, so the URL doesn't matter in prod.
- Moved the logic for determining the root into the manifest so it has default values in the build. This value is "wrong" in prod but is ignored anyways.

## Testing

- Existing tests should pass

## Docs

N/A, bug fix

### DIFF
--- a/.changeset/cuddly-days-relate.md
+++ b/.changeset/cuddly-days-relate.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Move root inside the manifest and make serialisable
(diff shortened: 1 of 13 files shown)


ematipico:
!preview server-islands

ematipico:
!preview server-islands

matthewp:
This approach should work fine. Not sure why the tests are currently breaking though.

PR #11494, opened 2024-07-18
feat(content-layer): support references in content layer

ascorbic:
## Changes

Adds support for the `reference()` in content layer schemas. This does unfortunately mean that we can't error out at validation time if the colleciotn name doesn't exist. This is because the content layer loads collections in parallel, and a colleciotn may not have been created at the time the other is syncing. Instead, what we do is introduce a check at runtime, which logs a warning if the colleciton doesn't exist.

## Testing

Adds tests


packages/astro/src/content/runtime.ts
```diff
@@ -277,7 +294,9 @@ export function createGetEntry({
 		if (store.hasCollection(collection)) {
 			const entry = store.get<DataEntry>(collection, lookupId);
 			if (!entry) {
-				throw new Error(`Entry ${collection} → ${lookupId} was not found.`);
+				// eslint-disable-next-line no-console
+				console.warn(`Entry ${collection} → ${lookupId} was not found.`);
```
ascorbic, line 298: There seems to be inconsistency in different parts of the existing code about what the correct behaious should be when `getEntry` is called and there's no match. Should it throw or just warn? I'm inclined to suggest that it throw if there's no matching collection, but just return undefined if there's no matching entry.

packages/astro/src/content/runtime.ts
```diff
@@ -208,6 +218,11 @@ export function createGetDataEntryById({ getEntryImport }: { getEntryImport: Get
 				collection,
 			};
 		}
+		if (!collectionNames.has(collection)) {
+			// eslint-disable-next-line no-console
+			console.warn(`The collection ${JSON.stringify(collection)} does not exist.`);
```
ascorbic, line 223: Is this the right behaviour? It's not clear whether a missing collection should throw or just warn.

ascorbic:
!preview contentlayer

packages/astro/src/content/runtime.ts
bluwy: It also warns in `main` now so I think warning is fine https://github.com/withastro/astro/blob/026e8baf3323e99f96530999fd32a0a9b305854d/packages/astro/src/content/runtime.ts#L64-L72

ascorbic: Yeah, that's the only reason I used it. I was trying to follow the current behaviour.

packages/astro/src/content/runtime.ts
```diff
@@ -168,6 +170,11 @@ export function createGetEntryBySlug({
 				render: () => renderEntry(entry),
 			};
 		}
+		if (!collectionNames.has(collection)) {
+			// eslint-disable-next-line no-console
+			console.warn(`The collection ${JSON.stringify(collection)} does not exist.`);
```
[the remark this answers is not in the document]
ascorbic: Do you have a suggestion of how best to do that? I can't find any other usage of the AstroLogger in the content runtime, so I'm not sure how best to get one in scope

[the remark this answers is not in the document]
matthewp: Yeah anything inside of runtime code is difficult to have a logger for.

PR #11496, opened 2024-07-18
Hide the dev toolbar on print

alfawal:
## Changes

The dev toolbar shows in every page when running `window.print()` (<kbd>CTRL</kbd> + <kbd>P</kbd>), which is useless and disturbing when working on styling a page for printing. This change hides the dev toolbar on printing.

If you see a point in having it on each page when printing, then I suggest adding a settings option to hide it.

Before:
image

After:
image

## Testing

Tested by emulating media print and checking visibility with Playwright.

## Docs

No usage instructions or important details to be mentioned.

### DIFF
--- a/.changeset/orange-boats-trade.md
+++ b/.changeset/orange-boats-trade.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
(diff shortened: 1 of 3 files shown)


Princesseuh (approved):
This is great, thank you!

PR #11505, opened 2024-07-19
fix: better logging for rewrites

ematipico:
## Changes

This PR enhances the logging of the dev server when there's a rewrite

## Testing

I tested locally

Screenshot 2024-07-19 at 08 27 39

The logic of the colours stays the same, which means the colour reflects the **final** status code.

### DIFF
--- a/packages/astro/src/vite-plugin-astro-server/route.ts
+++ b/packages/astro/src/vite-plugin-astro-server/route.ts
@@ -130,7 +130,7 @@ type HandleRoute = {
 	manifestData: ManifestData;
 	incomingRequest: http.IncomingMessage;
 	incomingResponse: http.ServerResponse;
-	status?: 404 | 500;
+	status?: 404 | 500 | 200;
 	pipeline: DevPipeline;
 };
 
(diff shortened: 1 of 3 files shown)


packages/astro/src/vite-plugin-astro-server/route.ts
```diff
@@ -130,7 +130,7 @@ type HandleRoute = {
 	manifestData: ManifestData;
 	incomingRequest: http.IncomingMessage;
 	incomingResponse: http.ServerResponse;
-	status?: 404 | 500;
+	status?: 404 | 500 | 200;
```
florian-lefebvre, line 133: Can't it be anything else?

ematipico: The type must match what's returned by `getStatus`, I just updated to make TS happy. I don't think it can be something else because `getStatus` checks our own physical routes

florian-lefebvre:
```suggestion
	status?: ReturnType<typeof getStatus>;
```

florian-lefebvre (approved):

ascorbic:
I'd find it confusing to have the status of what the request would have been without a rewrite. The target is the one I'm interested in. Showing that it has been rewritten is good though.

ematipico:
Screenshot 2024-07-19 at 12 13 04

Feedback applied

ascorbic (approved):

delucis (approved):
Lovely stuff. Thanks @ematipico!

PR #11507, opened 2024-07-19
feat: log slow pages in red

ematipico:
## Changes

A small change in DX that renders times that exceed the 500ms in red.

## Testing

Manually tested

Screenshot 2024-07-19 at 13 40 54

### DIFF
--- a/.changeset/spotty-rice-shake.md
+++ b/.changeset/spotty-rice-shake.md
@@ -1,5 +1,5 @@
 ---
-'astro': patch
+'astro': minor
 ---
 
 During the build, pages that take more than 500 milliseconds to render, will have their time increased logged in red.
--- a/.changeset/spotty-rice-shake.md
+++ b/.changeset/spotty-rice-shake.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+During the build, pages that take more than 500 milliseconds to render, will have their time increased logged in red.
--- a/.changeset/spotty-rice-shake.md
+++ b/.changeset/spotty-rice-shake.md
@@ -2,4 +2,6 @@
 'astro': minor
 ---
 
-During the build, pages that take more than 500 milliseconds to render, will have their time increased logged in red.
+Adds color-coding to the console output during the build to highlight slow pages. 
+
+Pages that take more than 500 milliseconds to render will have their build time logged in red. This change can help you discover pages of your site that are not performant and may need attention.
--- a/packages/astro/src/core/build/generate.ts
+++ b/packages/astro/src/core/build/generate.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import os from 'node:os';
 import { fileURLToPath } from 'node:url';
-import { bgGreen, black, blue, bold, dim, green, magenta } from 'kleur/colors';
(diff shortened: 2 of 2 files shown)


packages/astro/src/core/build/generate.ts
```diff
@@ -192,14 +192,16 @@ export async function generatePages(options: StaticBuildOptions, internals: Buil
 	await runHookBuildGenerated({ config, logger });
 }
 
+const THRESHOLD_SLOW_RENDER_TIME_MS = 500;
```
matthewp, line 195: is this number based on any data? I'm not really sure what to expect the value to be.

packages/astro/src/core/build/generate.ts
```diff
@@ -244,7 +246,13 @@ async function generatePage(
 			const timeEnd = performance.now();
 			const timeChange = getTimeStat(prevTimeEnd, timeEnd);
 			const timeIncrease = `(+${timeChange})`;
-			logger.info('SKIP_FORMAT', ` ${dim(timeIncrease)}`);
+			let timeIncreaseLabel;
+			if (Math.round(timeEnd - prevTimeEnd) > THRESHOLD_SLOW_RENDER_TIME_MS) {
```
bluwy, line 250: Do we need to call `Math.round` when comparing numbers? I think it should also work without

packages/astro/src/core/build/generate.ts
ematipico: No, it isn't based on any data.

bluwy (approved):

Princesseuh (approved):

packages/astro/src/core/build/generate.ts
matthewp: Is there a way that we can collect some data to have a better understanding what the expected render time is before deploying this kind of change? My only worry is that 500 is actually not that long and users start reporting that they are getting what they think is an error in their console.

.changeset/spotty-rice-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
```
matthewp, line 2: I think this should be a `minor` as it's a new "feature"

packages/astro/src/core/build/generate.ts
matthewp: Given outside discussions no issue with going with this number for now. We should adjust it if it turns out a lot of people are receiving false-positives on this, however.

.changeset/spotty-rice-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': minor
+---
+
+During the build, pages that take more than 500 milliseconds to render, will have their time increased logged in red.
```
sarah11918, line 5:
```suggestion
Adds color-coding to the console output during the build to highlight slow pages. 

Pages that take more than 500 milliseconds to render will have their build time logged in red. This change can help you discover pages of your site that are not performant and may need attention.
```
Was trying to think of how to make this sound helpful and exciting! :sweat_smile: Just a thought for a direction you could take!

sarah11918 (commented):
Superflous change, but it is a minor so we try to get people excited for new things! :partying_face:

PR #11508, opened 2024-07-19
Fix an XSS in Server Islands.

cramforce:
Discussed with @FredKSchott that this is OK to disclose since Server Islands are still experimental.

It's generally not safe to use `JSON.stringify` to interpolate potentially attacker controlled data into `<script>` tags as JSON doesn't escape `<>"'` and so one can use it to break out of the script tag and e.g. make a new one with controlled content.

See https://pragmaticwebsecurity.com/articles/spasecurity/json-stringify-xss

## Changes

No behavior change expected. All tests should still pass unless they trigger the encoding change.

## Testing

Not at all. I couldn't get tests to run.

### DIFF
--- a/.changeset/smart-snakes-promise.md
+++ b/.changeset/smart-snakes-promise.md
@@ -1,0 +1,5 @@
+---
+"astro": patch
+---
+
+Escapes HTML in serialized props
--- a/packages/astro/src/runtime/server/render/server-islands.ts
+++ b/packages/astro/src/runtime/server/render/server-islands.ts
@@ -2,6 +2,7 @@ import type { SSRResult } from '../../../@types/astro.js';
 import { renderChild } from './any.js';
 import type { RenderInstance } from './common.js';
 import { type ComponentSlots, renderSlotToString } from './slot.js';
+import {escape} from 'html-escaper';
 
 const internalProps = new Set([
 	'server:component-path',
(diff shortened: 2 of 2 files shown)


matthewp:
Wonderful, thank you very much.

ascorbic:
Thansk! Unfortunately this doesn't work, because it escapes the quotes too, leading to this in the source:

```js
let componentId = &quot;Island&quot;;
let componentExport = &quot;default&quot;;
let script = document.querySelector('script[data-island-id="ecac82eb-69f6-4e0a-896c-3ddde2c01f79"]');
let data = {
	componentExport,
	props: {},
	slots: {},
};
```

We'll need a different approach. I think it's only the actual props that could be vulnerable, because the component names aren't user-generated, but it is probably a legitimate issue.

cramforce:
I'll fix the encoding of quotes

ascorbic:
Reproduced the issue though: 
image
image

ascorbic:
I think we would be safe to just escape `</script>`. This seems to be enough:

```js
function safeJsonStringify(obj: any) {
	return JSON.stringify(obj).replaceAll('</script>', '<\\/script>');
}
```

cramforce:
@ascorbic You can circumvent that with all kinds of tricks. I can't for the life of me find a good npm package to do this. But it might exist. Added manual encoding for now

ascorbic:
That looks good. I just tried it and it does the job. Thanks for your help on this.

ascorbic:
Of course all of these will be fixed when we start encrypting the props as planned

matthewp:
Thanks! Patch should go out in a minute.

PR #11513, opened 2024-07-20
feat(server-islands): only encode ETAGO delimiter

kurtextrem:
## Changes

As per https://mathiasbynens.be/notes/etago, to make JSON safe inside a `<script>` literal, you need to only encode end-open tag (ETAGO) delimiters and `<!--`. Thus, we can avoid some work. I've also hoisted the regexps so that we only initialize them once.

Regarding the removal of 0x2029 and 0x2028 added in https://github.com/withastro/astro/pull/11508, to me it does not seem security related to escape those, [jsesc](https://github.com/mathiasbynens/jsesc) notes:

> [avoid errors](https://twitter.com/annevk/status/380000829643571200) when passing JSON-formatted data (which may contain U+2028 LINE SEPARATOR, U+2029 PARAGRAPH SEPARATOR, or [lone surrogates](https://esdiscuss.org/topic/code-points-vs-unicode-scalar-values#content-14)) to a JavaScript parser or an UTF-8 encoder.

Which does not seem relevant in the server-islands case. Did I miss anything here?

## Testing

I used the following snippet by @ascorbic: 
image

## Docs

Only a small perf related change, so no further docs needed.


matthewp:
Thanks @kurtextrem! Would you mind adding a test for this? You can add one here: https://github.com/withastro/astro/blob/main/packages/astro/test/server-islands.test.js (see the fixtures being used). The last PR was a hot-fix so a test didn't get added for urgency but now since the problem is fixed it would be good to have a test to prevent regressions.

florian-lefebvre:
@kurtextrem are you still interested in this PR? It would just need a test to move forward

kurtextrem:
@florian-lefebvre Sorry, yeah - I am. The only unclear thing is how extensive we want to test this.

ematipico:
@kurtextrem, are you still interested in finishing the PR?

.changeset/fifty-socks-end.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Updates the server islands encoding logic to only escape the script tag ETAGO delimiter and opening HTML comment syntax
```
ascorbic, line 5:
```suggestion
Updates the server islands encoding logic to only escape the script end tag open delimiter and opening HTML comment syntax
```
Make the changeset clearer

packages/astro/src/runtime/server/render/server-islands.ts
```diff
@@ -15,13 +15,19 @@ export function containsServerDirective(props: Record<string | number, any>) {
 	return 'server:component-directive' in props;
 }
 
+const scriptRegex = /<\/script/giu;
+const commentRegex = /<!--/gu;
+const scriptReplacer = '<\\/script';
+const commentReplacer = '\\u003C!--';
```
ematipico, line 21: We usually use upper-case for top-level constants like this ones

ematipico (approved):
Just a nit. Looks good!

packages/astro/test/fixtures/server-islands/ssr/src/pages/index.astro
```diff
@@ -1,12 +1,14 @@
 ---
 import Island from '../components/Island.astro';
+
+const xssMe ="</script><script>alert('xss')</script>"
```
ascorbic, line 4: It would be good to have a couple more examples in the test, particularly one that uses a comment

florian-lefebvre: I'm not really familiar with what could through, examples welcome

kurtextrem: Probably something like `"<!--"</script>`. Thank you for taking this over.

PR #11516, opened 2024-07-20
add ability to pass middleware to @astrojs/node when using standalone…

nm123github:
Add option to pass middleware to `@astrojs/node` when using standalone mode.

## Changes

- Make changes in the @astrojs/node to accept `standaloneMiddleware` file as a parameter
- If the `standaloneMiddleware` param is passed, @astrojs/node (in standalone mode) will attempt to load the middleware's and run them.
- The middleware file is expected to return the `combined` list of middleware's that need to run.
- The standalone server will work as it works today, if no `standaloneMiddleware` option is passed.

## Testing
I've tested this using a standalone app. Also added tests. Below are the sample changes.

astro.config.mjs
```
import { defineConfig } from "astro/config";
import node from "@astrojs/node";

// https://astro.build/config
export default defineConfig({
  output: "server",
  adapter: node({
    mode: "standalone",
    standaloneMiddleware: `${import.meta.dirname}/src/node_middleware.js`,
  }),
  integrations: [{
    name: "astro-lifecycle-logs",
    hooks: {
      'astro:server:setup': async ({ server }) => {
        // used during `pnpm run dev`
        const middleware = await import(`${import.meta.dirname}/src/node_middleware.js`);
        server.middlewares.use(middleware.default);
      }
    }
  }],
});
```

node_middleware.js
```
const combineMiddleware = (...middlewares) => {
    return (req, res, next) => {
        let index = 0;

        const run = (err) => {
            if (err) {
                return next(err);
            }
            if (index >= middlewares.length) {
                return next();
            }
            const middleware = middlewares[index++];
            try {
                middleware(req, res, run);
            } catch (error) {
                next(error);
            }
        };

        run();
    };
};

const middleware_one = (req, res, next) => {
    if (req.url === '/middleware-one') {
        res.end('This is middleware one');
    } else {
        next();
    }
};

const middleware_two = (req, res, next) => {
    if (req.url === '/middleware-two') {
        res.end('This is middleware two');
    } else {
        next();
    }
};

export default combineMiddleware(middleware_one, middleware_two);
```

## Docs
Opened PR for updating docs in `@astrojs/node` [integrations guide](https://docs.astro.build/en/guides/integrations-guide/node/)

https://github.com/withastro/docs/pull/8789


florian-lefebvre:
I'm not sure I understand the usecase. Why can't you use `middleware` mode and pass your middleware to something like express?

nm123github:
Its mostly for use-cases where we benefit from using (just)`middleware` (e.g. [http-proxy-middleware](https://www.npmjs.com/package/http-proxy-middleware)), and want to avoid overhead of adding ExpressJS.

florian-lefebvre:
We're not convinced by the usecase to be honest. We think middleware mode is exactly designed for what you want. You said you don't want the overhead of express but standalone mode is not meant to be a full fledge web server either. I'm sure you can find lighter alternatives than express if the bloat is what you're afraid of. Unless you have more arguments, I'm afraid we'll have to close the PR

Issue #11354, opened 2024-06-26
Using Astro Actions causes Build Error


hkbertoson:
### Astro Info

```block
Astro                    v4.11.3
Node                     v20.15.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/tailwind
                         astro-icon
                         @astrojs/react
```

### Describe the Bug

When actions are enabled. Project does not build

### What's the expected result?

Project should build

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-tpp1b5

### Participation

- [ ] I am willing to submit a pull request for this issue.

ematipico:
cc @bholmesdev

bholmesdev:
Hi @hkbertoson, thanks for reporting! I see you're using the Cloudflare adapter. Did you try the following?

1. [Add the `nodejs_als` flag](https://developers.cloudflare.com/workers/runtime-apis/nodejs#enable-only-asynclocalstorage) to your wrangler config. This allows a Node-specific API that actions relies on

```toml
# wrangler.toml
compatibility_flags = [ "nodejs_als" ]
```

2. [Mark this Node API as external](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#nodejs-compatibility) in your vite config:

```ts
import {defineConfig} from "astro/config";
import cloudflare from '@astrojs/cloudflare';

export default defineConfig({
  adapter: cloudflare({}),
  output: 'server',
  vite: {
    ssr: {
      external: ['node:async_hooks'],
    },
  },
})
```

I was also unable to reproduce this bug with the node adapter as @alexanderniebuhr noted above. If you can share a minimal reproduction with the node adapter that displays this issue, please share!

alexanderniebuhr:
@bholmesdev I just took that assumption based on screenshots from here https://discord.com/channels/830184174198718474/1255555330646216785

ematipico:
I am going to raise the priority of this issue, because the project doesn't even build with the Node.js adapter.

dgurns:
Just confirming that I ran into the original error in this issue -- failed build when using Astro Actions, related to `node:async_hooks`, and @bholmesdev's fix [here](https://github.com/withastro/astro/issues/11354#issuecomment-2194365696) solved it for me.

PR #11525, opened 2024-07-22
fix(actions): resolve actions when there aren't user actions

ematipico:
## Changes

Closes https://github.com/withastro/astro/issues/11354

The issue was caused by the static analyzer of rollup, where it was trying to "understand" if `import(import.meta.env.ACTIONS_PATH)` was a valid import. The function `getAction` wasn't even called, at all, during the build, because there were not actions to begin with. 

This issue was caused by the previous code's assumption that `actions/index.ts` was always present, which was not the case. This issue would have surfaced once the experimental flag was removed. 

I fixed the issue by using the same approach we use in the middleware, where we have an internal virtual module, and we return a noop code in case the user didn't specify any action.

This PR also removes the need to have an environment variable.

## Testing

Current CI should pass. I tested it locally by using the reproduction provided by the user.

### DIFF
--- a/packages/astro/src/actions/index.ts
+++ b/packages/astro/src/actions/index.ts
@@ -1,12 +1,25 @@
 import fsMod from 'node:fs';
 import type { Plugin as VitePlugin } from 'vite';
-import type { AstroIntegration } from '../@types/astro.js';
+import type { AstroIntegration, AstroSettings } from '../@types/astro.js';
 import { ActionsWithoutServerOutputError } from '../core/errors/errors-data.js';
 import { AstroError } from '../core/errors/errors.js';
 import { isServerLikeOutput, viteID } from '../core/util.js';
-import { ACTIONS_TYPES_FILE, RESOLVED_VIRTUAL_MODULE_ID, VIRTUAL_MODULE_ID } from './consts.js';
+import {
+	ACTIONS_TYPES_FILE,
+	NOOP_ACTIONS,
+	RESOLVED_VIRTUAL_INTERNAL_MODULE_ID,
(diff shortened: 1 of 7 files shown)


packages/astro/src/integrations/hooks.ts
```diff
@@ -21,6 +21,7 @@ import { mergeConfig } from '../core/config/index.js';
 import type { AstroIntegrationLogger, Logger } from '../core/logger/core.js';
 import { isServerLikeOutput } from '../core/util.js';
 import { validateSupportedFeatures } from './features-validation.js';
+import { vitePluginUserActions } from '../actions/index.js';
```
bholmesdev, line 24:
```suggestion
```

Unused import

packages/astro/src/actions/index.ts
```diff
@@ -50,6 +60,42 @@ export default function astroActions({ fs = fsMod }: { fs?: typeof fsMod }): Ast
...
+	let resolvedActionsId: string;
+	return {
+		name: '@astro/plugin-actions',
+		async resolveId(id) {
+			if (id === NOOP_ACTIONS) {
+				return NOOP_ACTIONS;
+			}
+			if (id === VIRTUAL_INTERNAL_MODULE_ID) {
+				const resolvedModule = await this.resolve(
+					`${decodeURI(settings.config.srcDir.pathname)}actions`
```
bholmesdev, line 78:
```suggestion
decodeURI(new URL('actions', settings.config.srcDir).pathname)
```

Nit: could use a URL constructor instead of string concat? Not certain if this would work without testing

bholmesdev (approved):
Thanks for taking `action` on this! Couple nits, non-blocking

PR #11535, opened 2024-07-23
Encrypt server islands props

matthewp:
## Changes

- Astro now creates a `CryptoKey` on dev startup. This key can be used for anything we need a key for, but specifically here it's for encrypting props. This is for 2 primary reasons:
  1. To prevent accidental leaking of secrets.
  2. To prevent CSRF attacks, you need the key to send a request.
- The key is stored in the manifest during the build, so any islands created at build-time use the same key as the one that will ship to product.
- We probably want a way to override the key with an environment variable, so that people can reuse the same key to avoid skew issues. Most won't use or need that.

## Testing

- Added test of passing a secret through the props.

## Docs

Part of the RFC

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -168,6 +168,7 @@
     "magic-string": "^0.30.10",
     "mrmime": "^2.0.0",
     "ora": "^8.0.1",
+    "oslo": "^1.2.1",
     "p-limit": "^6.1.0",
     "p-queue": "^8.0.1",
     "path-to-regexp": "^6.2.2",
--- a/packages/astro/src/core/encryption.ts
+++ b/packages/astro/src/core/encryption.ts
@@ -1,0 +1,68 @@
+import { base64, decodeHex, encodeHex } from 'oslo/encoding';
+
+export async function createKeyBytes() {
+	const key = await crypto.subtle.generateKey(
+		{
+			name: 'AES-GCM',
+			length: 256,
+		},
+		true,
+		['encrypt', 'decrypt']
+	);
+	const exported = await crypto.subtle.exportKey('raw', key);
+	return new Uint8Array(exported);
+}
+
+export async function createKey() {
+  const bytes = await createKeyBytes();
+  return importKey(bytes);
+}
+
+export async function importKey(bytes: Uint8Array) {
+  const key = await crypto.subtle.importKey('raw', bytes, 'AES-GCM', true, ['encrypt', 'decrypt']);
+  return key;
+}
+
+export async function encodeKey(key: CryptoKey) {
+	const exported = await crypto.subtle.exportKey('raw', key);
+	const encodedKey = base64.encode(new Uint8Array(exported));
+	return encodedKey;
+}
+
+export async function decodeKey(encoded: string) {
+  const bytes = base64.decode(encoded);
+	return crypto.subtle.importKey('raw', bytes, 'AES-GCM', true, ['encrypt', 'decrypt']);
+}
+
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+const IV_LENGTH = 24;
+
+export async function encryptData(key: CryptoKey, raw: string) {
+	const iv = crypto.getRandomValues(new Uint8Array(IV_LENGTH / 2));
+	const data = encoder.encode(raw);
+	const buffer = await crypto.subtle.encrypt(
+		{
+			name: 'AES-GCM',
+			iv,
+		},
+		key,
+		data
+	);
+	return encodeHex(iv) + base64.encode(new Uint8Array(buffer));
+}
+
+export async function decryptData(key: CryptoKey, encoded: string) {
+  const iv = decodeHex(encoded.slice(0, IV_LENGTH));
+	const dataArray = base64.decode(encoded.slice(IV_LENGTH));
+	const decryptedBuffer = await crypto.subtle.decrypt(
+		{
+			name: 'AES-GCM',
+			iv,
+		},
+		key,
+		dataArray
+	);
+	const decryptedString = decoder.decode(decryptedBuffer);
+	return decryptedString;
+}
--- a/packages/astro/src/core/server-islands/endpoint.ts
+++ b/packages/astro/src/core/server-islands/endpoint.ts
@@ -11,6 +11,7 @@ import {
 	renderTemplate,
 } from '../../runtime/server/index.js';
 import { createSlotValueFromString } from '../../runtime/server/render/slot.js';
+import { decryptData } from '../encryption.js';
 import { getPattern } from '../routing/manifest/pattern.js';
 
 export const SERVER_ISLAND_ROUTE = '/_server-islands/[name]';
@@ -48,7 +49,7 @@ export function ensureServerIslandRoute(config: ConfigFields, routeManifest: Man
 
 type RenderOptions = {
 	componentExport: string;
-	props: Record<string, any>;
+	props: string;
 	slots: Record<string, string>;
 };
 
@@ -74,7 +75,11 @@ export function createEndpoint(manifest: SSRManifest) {
 			});
 		}
 
-		const props = data.props;
+		const key = await manifest.key;
+		const encryptedProps = data.props;
+		const propString = await decryptData(key, encryptedProps);
+		const props = JSON.parse(propString);
+		
 		const componentModule = await imp();
 		const Component = (componentModule as any)[data.componentExport];
 
--- a/packages/astro/src/runtime/server/render/server-islands.ts
+++ b/packages/astro/src/runtime/server/render/server-islands.ts
@@ -1,7 +1,9 @@
+import { encryptData } from '../../../core/encryption.js';
 import type { SSRResult } from '../../../@types/astro.js';
 import { renderChild } from './any.js';
 import type { RenderInstance } from './common.js';
 import { type ComponentSlots, renderSlotToString } from './slot.js';
+
 
 const internalProps = new Set([
 	'server:component-path',
@@ -59,6 +61,9 @@ export function renderServerIsland(
 				}
 			}
 
+			const key = await result.key;
+			const propsEncrypted = await encryptData(key, JSON.stringify(props));
+
 			const hostId = crypto.randomUUID();
 
 			destination.write(`<script async type="module" data-island-id="${hostId}">
@@ -67,7 +72,7 @@ let componentExport = ${safeJsonStringify(componentExport)};
 let script = document.querySelector('script[data-island-id="${hostId}"]');
 let data = {
 	componentExport,
-	props: ${safeJsonStringify(props)},
+	props: ${JSON.stringify(propsEncrypted)},
 	slots: ${safeJsonStringify(renderedSlots)},
 };
 
--- a/packages/astro/e2e/fixtures/server-islands/src/components/Island.astro
+++ b/packages/astro/e2e/fixtures/server-islands/src/components/Island.astro
@@ -1,4 +1,6 @@
 ---
+const { secret } = Astro.props;
 ---
 <h2 id="island">I am an island</h2>
 <slot />
+<h3 id="secret">{secret}</h3>
--- a/packages/astro/e2e/fixtures/server-islands/src/pages/index.astro
+++ b/packages/astro/e2e/fixtures/server-islands/src/pages/index.astro
@@ -7,7 +7,7 @@ import Island from '../components/Island.astro';
 		<!-- Head Stuff -->
 	</head>
 	<body>
-		<Island server:defer>
+		<Island server:defer secret="test">
 			<h3 id="children">children</h3>
 		</Island>
 	</body>
--- a/packages/astro/e2e/server-islands.test.js
+++ b/packages/astro/e2e/server-islands.test.js
@@ -37,6 +37,12 @@ test.describe('Server islands', () => {
 
 			await expect(el, 'element rendered').toBeVisible();
 		});
+
+		test('Props are encrypted', async ({ page, astro }) => {
+			await page.goto(astro.resolveUrl('/'));
+			let el = page.locator('#secret');
+			await expect(el).toHaveText('test');
+		});
 	});
 
 	test.describe('Production', () => {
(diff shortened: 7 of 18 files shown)


packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,68 @@
...
+			length: 256,
+		},
+		true,
+		['encrypt', 'decrypt']
+	);
+	const exported = await crypto.subtle.exportKey('raw', key);
+	return new Uint8Array(exported);
+}
+
+export async function createKey() {
```
matthewp, line 16: i don't think i need separate functions here, can just combine with createKeyBytes and not export in the former.

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,68 @@
...
+	const data = encoder.encode(raw);
+	const buffer = await crypto.subtle.encrypt(
+		{
+			name: 'AES-GCM',
+			iv,
+		},
+		key,
+		data
+	);
+	return encodeHex(iv) + base64.encode(new Uint8Array(buffer));
```
matthewp, line 52: iv is length 12, encoding gets it to IV_LENGTH, I think.

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export async function decryptData(key: CryptoKey, encoded: string) {
```
matthewp, line 55: should rename these to `encryptString` / `decryptString`

packages/astro/src/core/encryption.ts
bholmesdev: Yep, that could make the internals simpler 👍

packages/astro/src/core/encryption.ts
bholmesdev: Yep, that's my understanding. Worth a code comment?

packages/astro/src/runtime/server/render/server-islands.ts
```diff
@@ -67,7 +72,7 @@ let componentExport = ${safeJsonStringify(componentExport)};
 let script = document.querySelector('script[data-island-id="${hostId}"]');
 let data = {
 	componentExport,
-	props: ${safeJsonStringify(props)},
+	props: ${JSON.stringify(propsEncrypted)},
```
bholmesdev, line 75: Any reason to stop using `safeJsonStringify()`?

bholmesdev (commented):
See you left some refactor notes already. Commented where I could

packages/astro/src/runtime/server/render/server-islands.ts
matthewp: nope, i should be.

Tc-001:
> To prevent CSRF attacks, you need the key to send a request

I wouldn't count on it because if the key is global and static then the attacker can generate the same encrypted props. Probably doesn't have any real-world danger (hopefully)

Overall looks good!

matthewp:
@Tc-001 how can they generate the same props without the key? The key is not static to all of Astro if that's what you're thinking. It's generated per build.

Tc-001:
But all the attacker has to do is make a top-level request that produces an island with the same props. So for example `<AccountAction server:dynamic action="delete"/>` that (for some reason) immediately deletes your account

This would be CSRF-able because the actual auth would be from the cookies as the key is not per-user.

But I don't think it's a big deal because I *hope* islands are used as read-only and anything important is an action / POST frontmatter / API route etc.

matthewp:
@Tc-001 Whether it's a big deal or not, if there's a vulnerability I would like to try and prevent it, so let's talk through this.

I still don't think I understand the scenario here. You say that the attacker can produce an island with the same props; how can they do that? The props are encrypted with a key that only the server knows about. You can't create a request that mimics the real request because your key will not be what is used to decrypt the props.

I believe you here! Just not following, can you walk me through it some more?

Tc-001:
@matthewp 
> You say that the attacker can produce an island with the same props; how can they do that?

If the server serves an attacker an island with encrypted props, the attacker can then redirect the user (with some CSRF or a straight up redirect) to that island and it will render with the user's cookies.

So the attacker cannot change what props the island has, but if the site serves a dangerous island and the attacker can get to it, there is nothing preventing the same encrypted props working on the user (unless the username is one of the props for example)

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,69 @@
+import { base64, decodeHex, encodeHex } from 'oslo/encoding';
+
+export async function createKeyBytes() {
+	const key = await crypto.subtle.generateKey(
+		{
+			name: 'AES-GCM',
```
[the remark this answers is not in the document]
matthewp: yep, changed it.

packages/astro/package.json
```diff
@@ -168,6 +168,7 @@
     "magic-string": "^0.30.10",
     "mrmime": "^2.0.0",
     "ora": "^8.0.1",
+    "oslo": "^1.2.1",
```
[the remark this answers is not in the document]
matthewp: Buffer is not cross-platform (Deno, Cloudflare) I believe, right?

matthewp:
@Tc-001 I understand what you mean now. An attacker can't change the props but if you had a destructive island that say, deleted a database row upon load, they could just repeat the same request to delete more rows potentially.  We have https://docs.astro.build/en/reference/configuration-reference/#securitycheckorigin for that.

pilcrowOnPaper:
If I'm understanding server islands correctly, the JSON-serialized props are passed to the client and are re-sent to render the island with `fetch()`? I *think* this should be sufficient for protecting secrets but I agree with @Tc-001 that this part isn't 100% (though I'm not sure how viable attacks are):

> To prevent over-trusting of island props; they are encrypted so an attacker can't forge a request.

I would put a strong emphasis on doing auth checks on a per-server-island basis. Basic example but this wouldn't be bad if the user used to have admin access (but got revoked):

```ts
---
if (!Astro.props.user || Astro.props.user.role !== "admin") {
  return new Response(null, { status: 403 })
}
const data = getSuperPrivateUserData();
---
```

pilcrowOnPaper:
To be honest, given the uncertainty from a security perspective and the performance implications, I'm not fully onboard with it. I would prefer if it was made clear that the props are from the client and not directly from the server like regular components. 

(As a side note, if Astro didn't need to support serverless platforms, it could just store the props in a session)

pilcrowOnPaper:
A bit off-topic, but for skew protection, wouldn't a simple hash or timestamp suffice?

bholmesdev:
@pilcrowOnPaper I understand the hesitancy. Still, I want to consider the wider application of an encryption primitive in Astro. A common use case shared by Laravel is to [encrypt session data as cookies](https://laravel.com/docs/11.x/session#configuration) for flashing user inputs and preserving information after a redirect. Astro could benefit from mirroring this approach for Astro Actions.

Do you have an opinion on that use case?

matthewp:
@pilcrowOnPaper Thanks for the review. Yes, you absolutely need to still do auth checks in server islands and should not rely on props. My thinking writing that was that you might want to prevent forgeries outside of auth, but I can't really thinking of any that would matter. So if we exclude that as a reason for this, it still helps prevent accidental leaking of secrets passed via props. Do you not think that's a good reason for this?

----

> I'm not sure if it's already implemented, but I would also add an id unique to the target server island to the prop if possible. This would ensure that an attacker can't reuse encrypted props from other server islands. I believe the encrypted props for these 2 server islands would be interchangeable.

I don't think I follow what you're saying here. We serialize all of the props into a single string, not just the values, so an attacker can't create a similar shaped request. If the props were exactly the same then they could reuse that in a request to a different island. I'm not exactly sure what that would achieve, however.

ascorbic:
I guess the main argument _against_ including this is that it provides a false sense of security, because even if it does mitigate most issues, it's not a full CSRF-proof solution. I do think it's better than nothing though, and we'd just need to document well the parts that can't be trusted.

pilcrowOnPaper:
@bholmesdev 

I personally don't understand why you'd need to encrypt cookies, but unlike props, it's much more clear that cookies are stored in and sent from the client. 

@matthewp 

>  If the props were exactly the same then they could reuse that in a request to a different island.

Yeah that's my point, though there needs to be other vulnerabilities (like the insufficient auth check) for the attack to work.

matthewp:
It should be easy to include the island as part of the encrypted props, we can then verify on the receiving end that the request is for that island. I'll add that to this PR.

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,69 @@
...
+export async function importKey(bytes: Uint8Array) {
+  const key = await crypto.subtle.importKey('raw', bytes, 'AES-GCM', true, ['encrypt', 'decrypt']);
+  return key;
+}
+
+export async function encodeKey(key: CryptoKey) {
+	const exported = await crypto.subtle.exportKey('raw', key);
+	const encodedKey = base64.encode(new Uint8Array(exported));
+	return encodedKey;
+}
```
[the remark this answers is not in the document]
Fryuni: Supporting a JWK would be nice for the idea of setting the key from an environment variables. Key management systems don't usually expose the raw key materials so users would need to add conversion steps.

packages/integrations/node/src/server.ts
```diff
@@ -4,14 +4,14 @@ import createMiddleware from './middleware.js';
 import { createStandaloneHandler } from './standalone.js';
 import startServer from './standalone.js';
 import type { Options } from './types.js';
+applyPolyfills();
```
matthewp, line 7: Need to run this first

.changeset/hip-toys-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/node': patch
+---
+
+Apply polyfills before awaiting the env module
```
ematipico, line 5: Can you add a better description for this change? I really don't understand what it is

packages/astro/src/core/build/static-build.ts
```diff
@@ -255,6 +255,8 @@ async function ssrBuild(
 							return 'renderers.mjs';
 						} else if (chunkInfo.facadeModuleId === RESOLVED_SSR_MANIFEST_VIRTUAL_MODULE_ID) {
 							return 'manifest_[hash].mjs';
+						} else if (chunkInfo.facadeModuleId === settings.adapter?.serverEntrypoint) {
+							return 'adapter_[hash].mjs';
```
ematipico, line 259: What's this change about?

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
+import { base64, decodeHex, encodeHex } from '@oslojs/encoding';
+
+const ALGORITHM = 'AES-GCM';
```
ematipico, line 3: Can you add a link that explains the algorithm? Maybe Wikipedia or MDN

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export async function createKey() {
```
ematipico, line 5: Can you please add some documentation that explains what kind of key this function generates?

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
...
+			name: ALGORITHM,
+			length: 256,
+		},
+		true,
+		['encrypt', 'decrypt']
+	);
+	return key;
+}
+
+export async function importKey(bytes: Uint8Array) {
```
ematipico, line 17: The documentation is missing. Also, the function's name suggests that it imports the key from somewhere, so we should explain that part, too. We should explain also what are the `bytes`, and what they are meant for

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
...
+  const key = await crypto.subtle.importKey('raw', bytes, ALGORITHM, true, ['encrypt', 'decrypt']);
+  return key;
+}
+
+export async function encodeKey(key: CryptoKey) {
```
ematipico, line 22: Same here, we should explain the function. Having an example is also good.

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export async function decodeKey(encoded: string) {
+  const bytes = base64.decode(encoded);
+	return crypto.subtle.importKey('raw', bytes, ALGORITHM, true, ['encrypt', 'decrypt']);
+}
+
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+const IV_LENGTH = 24;
```
ematipico, line 35: Let's document this constant please. I don't know what a `IV length` is in this context.

packages/astro/src/core/encryption.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export async function encryptString(key: CryptoKey, raw: string) {
```
ematipico, line 37: Docs. Let's explain when to use this function, maybe with an example

packages/astro/src/core/server-islands/endpoint.ts
```diff
@@ -74,7 +75,11 @@ export function createEndpoint(manifest: SSRManifest) {
 			});
 		}
 
-		const props = data.props;
+		const key = await manifest.key;
+		const encryptedProps = data.props;
```
ematipico, line 79: Let's rename `props` to `encriptedProps` at this point so it's more visible to us.

packages/integrations/node/src/server.ts
ematipico: Let's leave a comment so devs know why it's there.

packages/astro/src/core/encryption.ts
bholmesdev: This is the document I followed for the first implementation if it helps: https://gist.github.com/rpivo/2904951ee31ce9146d32321f12c941bd

We can also link to the Next.js source, but it's a bit too spread out to be helpful in the future.

packages/astro/src/core/build/static-build.ts
matthewp: This changes depends on `globalThis.crypto` which isn't available in Node 18. We have a polyfill for it, but currently the manifest is created before the adapter runs, so the adapter isn't able to polyfill it in time.

So the fix was to make the adapter an entrypoint, causing it to be built to a separate file and therefore it would run before the manifest. This line here just names it a certain way to make debugging easier.

matthewp:
This needs some cleanup. As I'll be out for a week I'll do this when I get back.

.changeset/hip-toys-shake.md
matthewp: Done, thanks!

ematipico (approved):

PR #11541, opened 2024-07-24
Clean content layer with `--force`

ascorbic:
## Changes
Cleans the content layer data store when `astro build --force` is run. This matches the behavior for content collection caching.

## Testing

Added tests for builds with and without the flag.

## Docs

This will need to be included in the docs when content layer is added


ematipico (changes requested):
We have to add the new flag to the help section. I don't know why it wasn't documented, but we really should:

https://github.com/withastro/astro/blob/536209aa74924dbde4084026e3b72832961813ac/packages/astro/src/cli/build/index.ts#L16-L18

Also, I would consider an environment variable too. In this PR or another, totally up to you

ematipico (approved):

PR #11550, opened 2024-07-25
refactor: dynamically import middleware

ematipico:
## Changes

This PR refactors how we import the middleware file.  This was a refactor I wanted to do regardless, but the Cloudflare issues put some pressure on it, so it's a good change regardless because it allows us to add more variables other than `onRequest` inside the middleware. Potential uses cases: errors, and more.

Before, Astro always imported the `onRequest` method from the middleware file.

Now, Astro imports the entire module, and then resolve the `onRequest` method from it.

To achieve this, I used the `Pipeline` abstract class to define a shared method called `getMiddleware()`, which is used inside the `RenderContext` class. 

The `RenderContext::create` method is now `async`, because during the creation of it, we need to resolve the middleware too.

## Testing

The existing tests should pass. I didn't add any changeset because this is an internal refactor, and the business logic didn't change.

## Docs

It's an internal refactor, the functionality hasn't changed

### DIFF
--- a/packages/astro/src/container/index.ts
+++ b/packages/astro/src/container/index.ts
@@ -22,13 +22,13 @@ import { validateConfig } from '../core/config/validate.js';
 import { createKey } from '../core/encryption.js';
 import { Logger } from '../core/logger/core.js';
 import { nodeLogDestination } from '../core/logger/node.js';
+import { NOOP_MIDDLEWARE_FN } from '../core/middleware/noop-middleware.js';
 import { removeLeadingForwardSlash } from '../core/path.js';
 import { RenderContext } from '../core/render-context.js';
 import { getParts, validateSegment } from '../core/routing/manifest/create.js';
 import { getPattern } from '../core/routing/manifest/pattern.js';
 import type { AstroComponentFactory } from '../runtime/server/index.js';
 import { ContainerPipeline } from './pipeline.js';
-import { NOOP_MIDDLEWARE_FN } from '../core/middleware/noop-middleware.js';
 
 /**
  * Options to be passed when rendering a route
@@ -474,7 +474,6 @@ export class experimental_AstroContainer {
 			routeType === 'endpoint'
 				? (component as unknown as ComponentInstance)
 				: this.#wrapComponent(component, options.params);
-		const middleware = await this.#pipeline.getMiddleware();
 		const routeData = this.#insertRoute({
 			path: request.url,
 			componentInstance,
@@ -485,7 +484,6 @@ export class experimental_AstroContainer {
 			pipeline: this.#pipeline,
 			routeData,
 			status: 200,
-			middleware,
 			request,
 			pathname: url.pathname,
 			locals: options?.locals ?? {},
--- a/packages/astro/src/container/index.ts
+++ b/packages/astro/src/container/index.ts
@@ -2,6 +2,7 @@ import './polyfill.js';
 import { posix } from 'node:path';
 import type {
 	AstroConfig,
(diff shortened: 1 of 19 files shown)


florian-lefebvre:
!preview dynamic-middleware

alexanderniebuhr:
This doesn't fix the issue fully yet, but I wound the following.

If we could separate the helper functions `AstroCookies`, `sequence`, `defineMiddleware` in a different chunk, and remove them from `_astro-internal_middleware` chunk, it could work. All those helpers should stay statically imported, while everything else "user middleware", should be dynamically imported.

https://discord.com/channels/830184174198718474/1228360509418504332/1266257419164450976

packages/astro/src/vite-plugin-astro-server/plugin.ts
```diff
@@ -152,8 +152,12 @@ export function createDevelopmentManifest(settings: AstroSettings): SSRManifest
 		checkOrigin: settings.config.security?.checkOrigin ?? false,
 		experimentalEnvGetSecretEnabled: false,
 		key: createKey(),
-		middleware(_, next) {
-			return next();
+		middleware() {
+			return {
+				onRequest(_, next) {
```
florian-lefebvre, line 157: could we reuse the noop middleware here?

ematipico: Good eye. Done

alexanderniebuhr:
!preview dynamic-middleware

packages/astro/src/container/pipeline.ts
```diff
@@ -88,7 +109,7 @@ export class ContainerPipeline extends Pipeline {
 				return Promise.resolve(componentInstance);
 			},
 			renderers: this.manifest.renderers,
-			onRequest: this.manifest.middleware,
+			onRequest: this.resolvedMiddleware,
```
florian-lefebvre, line 112: Are we sure that `getMiddleware()` has been called before?

packages/astro/src/core/app/pipeline.ts
```diff
@@ -11,9 +12,28 @@ import type { SinglePageBuiltModule } from '../build/types.js';
...
 import { findRouteToRewrite } from '../routing/rewrite.js';
+import { sequence } from '../middleware/index.js';
+import { createOriginCheckMiddleware } from './middlewares.js';
+import { NOOP_MIDDLEWARE_FN } from '../middleware/noop-middleware.js';
 
 export class AppPipeline extends Pipeline {
 	#manifestData: ManifestData | undefined;
+	resolvedMiddleware: MiddlewareHandler | undefined = undefined;
+
+	async getMiddleware(): Promise<MiddlewareHandler> {
```
florian-lefebvre, line 23: Is there a reason why this logic is duplicated with the container pipeline? Could it be part of `Pipeline`?

florian-lefebvre (commented):
I think we may need a changeset since it changes the build output, that could have an impact on adapters

florian-lefebvre (approved):

packages/astro/src/container/index.ts
```diff
@@ -469,17 +473,18 @@ export class experimental_AstroContainer {
 			routeType === 'endpoint'
 				? (component as unknown as ComponentInstance)
 				: this.#wrapComponent(component, options.params);
+		const middleware = await this.#pipeline.getMiddleware();
```
bluwy, line 476: Do we need this line? `RenderContext.create` also seem to do the same internally.

bluwy (approved):

PR #11551, opened 2024-07-25
feat: injectTypes

florian-lefebvre:
## Changes

- Adds `injectTypes` utility to `astro:config:done`
- Refactors how typegen works internally so it's also used by core
- Update `astro:db` to use this new  API
- Refactors `.astro` structure
	- It's backward compatible
	- Now, a `.astro/types.d.ts` file is generated with references to any injected type + `astro/client`
	- Core types are moved to `.astro/astro`
	- Integration types are written to `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts`
	- `src/env.d.ts` only needs a reference to the newly generated `.astro/types.d.ts`. Goal in the future is to remove the need for this file by generating a `tsconfig.json` file
- Uses `settings.dotAstroDir` when possible

## Testing

- Updates sync tests
- All tests should pass
- Added unit tests for types path normalization

## Docs

https://github.com/withastro/docs/pull/9012

## Additional info

- RFC: https://github.com/withastro/roadmap/discussions/887
- First disastrous attempt: https://github.com/withastro/astro/pull/9952
- [addDts](https://astro-integration-kit.netlify.app/utilities/add-dts/) from AIK ([source](https://github.com/florian-lefebvre/astro-integration-kit/blob/main/package/src/utilities/add-dts.ts))


.changeset/fresh-fans-study.md
```diff
@@ -0,0 +1,18 @@
...
+```diff
+- .astro/db-types.d.ts
++ .astro/astro/db.d.ts
+```
+
+The update made to `src/env.d.ts` can be undone:
+
+```diff
+- /// <reference path="../.astro/db-types.d.ts" />
+```
```
delucis, line 18: This sounds to me like I need to undo something **in this update**, but I think you mean something else by “update” right?

Something like this?

````suggestion
The following line can now be removed from `src/env.d.ts`:

```diff
- /// <reference path="../.astro/db-types.d.ts" />
```
````

File: .changeset/fresh-fans-study.md [hunk 2]
```diff
@@ -0,0 +1,18 @@
+---
+'@astrojs/db': minor
+---
+
+Changes how typegen works
```
delucis, line 5: We should avoid abbreviating this for people less familiar with the concept:

```suggestion
Changes how type generation works
```

.changeset/fresh-fans-study.md
```diff
@@ -0,0 +1,18 @@
...
+
+The generated dts file is now at a new location:
```
delucis, line 7: And here should probably use the correct extension:

```suggestion
The generated `.d.ts` file is now at a new location:
```

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
+---
+'astro': minor
+---
+
+Adds a new `injectTypes` utility to the Integration API and refactors how typegen works
```
delucis, line 5:
```suggestion
Adds a new `injectTypes()` utility to the Integration API and refactors how type generation works
```

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
...
+
+Integrations can now use a new `injectTypes` utility exposed on the `astro:config:done` hook. It allows to inject types easily in the user's project. `filename` must ends with `".d.ts"` and `content` must be valid TypeScript (it will be formatted).
+
+Under the hood, it will create a file at `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts` and create references to it. `injectTypes` returns a URL to the normalized path.
```
delucis, line 9: Suggesting a couple tweaks here. In general in docs we show a `()` after function names to distinguish them from non-function variables:

```suggestion
Adds a new `injectTypes()` utility to the Integration API and refactors how type generation works

Integrations can now use a new `injectTypes()` utility in the `astro:config:done` hook. This utility allows an integration to inject types into a user's project. `filename` must end with `".d.ts"` and `content` must be valid TypeScript (it will be formatted).

Under the hood, it will create a file at `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts` and create references to it. `injectTypes()` returns a URL to the normalized path.
```

delucis (commented):
Thanks @florian-lefebvre! Quickly jotting down my initial questions:

- Is there a particular reason this needs to happen in `astro:config:done` rather than `setup`? So much of the user-facing integration APIs happen in `setup` so I tend to think of that as the main hook vs `done`.
- Is formatting the generated types important? It feels like an added complexity that could be skipped in theory.

Not needed in this context where we’re just dealing with a changeset, but for docs themselves, it’s probably important to document the shape of a type declaration expected by the hook — i.e. you can’t just inject `interface Foo {}` and expect it to work.

florian-lefebvre:
Smoke test is failing and I have no clue why. It seems completely unrelated

florian-lefebvre:
@delucis some answers!

> Is there a particular reason this needs to happen in `astro:config:done` rather than `setup`? So much of the user-facing integration APIs happen in `setup` so I tend to think of that as the main hook vs `done`.

There are a few reasons to this:
1. It's common to do some typegen based on the final astro config, exposed on `config:done`
2. `sync` is always executed after both `config:setup` and `config:done` so adding `injectTypes` to the latter unlocks more things imo
3. Astro DB does typegen in `config:done` already
4. In the future, we'll expose some `RouteData` related data on a hook and it's likely to be on `config:done`. That will enable interesting things for codegen (eg. my integration for typed links, i18n etc)

> Is formatting the generated types important? It feels like an added complexity that could be skipped in theory.

I would not say it's important but is nice to have. It doesn't add much complexity since we're reusing existing dependencies. I'd say it has a few benefits:
1. It's easy to debug, even if you write your whole content in a string etc
2. actually checks if valid TS is provided since the code has to be parsed (may need some error handling now that I'm thinking about it)

> Not needed in this context where we’re just dealing with a changeset, but for docs themselves, it’s probably important to document the shape of a type declaration expected by the hook — i.e. you can’t just inject `interface Foo {}` and expect it to work.

Definitely

delucis:
Thanks @florian-lefebvre — makes sense!

> > Is formatting the generated types important? It feels like an added complexity that could be skipped in theory.
> 
> I would not say it's important but is nice to have. It doesn't add much complexity since we're reusing existing dependencies. I'd say it has a few benefits:
> 
> 1. It's easy to debug, even if you write your whole content in a string etc
> 2. actually checks if valid TS is provided since the code has to be parsed (may need some error handling now that I'm thinking about it)

Yeah, I didn’t mean so much maintenance complexity, but runtime complexity:

- We are now erroring on badly formatted TypeScript instead of just passing it through and letting the TS server deal with it however it likes. (There are pros and cons here admittedly, but it could be annoying for something to be entirely broken because of a small type gen error, vs just types being broken while the functionality still works.)

- We create a disconnect between the authored input and the consumed output. I don’t know how likely it is, but if that format step ever messed with functionality, that wouldn’t be great. And in terms of debugging, it could be harder for integration authors to link the output back to their code if the formatter radically changes line numbers etc. (The pro I see here is that messy input can still generate readable output in case someone clicks through to the `.d.ts` I guess.)

florian-lefebvre:
Yeah I get your point as well regarding formatting. Another possibility would be to add it as an option

BryceRussell:
My 2 cents on formatting:

I think automatically formatting the types in some way is important. I have experience with the `astro-integration-kit` implementation before it had formatting and it was painful to use. With formatting, I was able simplify my code a lot and create complicated types using only string concatenation. I did have issues with errors being emited like Chris mentioned, but I always figured that this would be something that Astro could improve with better errors messages that explain it is a type gen error, give the file name, etc.

packages/astro/src/core/sync/write-files.ts
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		// Otherwise, inject the `env.d.ts` file
+		await fs.promises.mkdir(settings.config.srcDir, { recursive: true });
+		await fs.promises.writeFile(envTsPath, expectedTypeReference, 'utf-8');
+		logger.info('types', `Added ${bold(envTsPathRelativetoRoot)} type declarations`);
+	}
+}
+
+function formatContent(source: string) {
+	const ast = parse(source, {
```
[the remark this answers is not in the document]
delucis: I guess we could also log an error without throwing potentially for type gen? (If we agree there’s value in letting things keep working even when type gen fails.)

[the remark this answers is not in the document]
florian-lefebvre: Yeah logging may make more sense, suggesting to report to the right integration

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
...
+
+```js
+const path = injectTypes({
+  filename: "types.d.ts",
+  content: "declare module 'virtual:integration' {}"
+})
+console.log(path) // URL
+```
+
+Codegen has been refactored. As a user, you just need to update `src/env.d.ts`:
```
ematipico, line 19: I believe it's worth mentioning that the old values are still valid

packages/astro/src/core/sync/write-files.ts
```diff
@@ -0,0 +1,88 @@
...
+	}
+	fs.writeFileSync(
+		normalizePath(fileURLToPath(new URL(REFERENCE_FILE, settings.dotAstroDir))),
+		astroDtsContent,
+		'utf-8'
+	);
+}
+
+async function setUpEnvTs(settings: AstroSettings, fs: typeof fsMod, logger: Logger) {
+	const envTsPath = normalizePath(fileURLToPath(new URL('env.d.ts', settings.config.srcDir)));
```
ematipico, line 45: We should have a big `try/catch` block to catch all the possible file system errors that could be thrown

packages/astro/src/integrations/hooks.ts
```diff
@@ -100,6 +100,17 @@ export function getToolbarServerCommunicationHelpers(server: ViteDevServer) {
 	};
 }
 
+const SAFE_CHARS_RE = /[^\w.-]/g
```
ematipico, line 103: A small comment that explains the regex and what's that for will be useful

packages/astro/src/integrations/hooks.ts
```diff
@@ -327,6 +338,28 @@ export async function runHookConfigDone({
 						}
 						settings.adapter = adapter;
 					},
+					injectTypes(injectedType) {
+						if (integration.name === 'astro:db' && injectedType.filename === 'db-types.d.ts') {
+							const filename = './astro/db.d.ts';
```
ematipico, line 343: This feels like code smell, because we are handling something for `@astrojs/db`, an integration, inside core. Not sure why it's needed, but if we can't do otherwise, we should add a comment that explains this exception

packages/astro/src/core/sync/write-files.ts
florian-lefebvre: Where should such try/catch be? maybe wrap everything inside `writeFiles`?

.changeset/mean-horses-kiss.md
florian-lefebvre:
```suggestion
Codegen has been refactored. Although `src/env.d.ts` will continue to work as is, we recommend you update it:
```

packages/astro/src/integrations/hooks.ts
```diff
@@ -100,6 +100,17 @@ export function getToolbarServerCommunicationHelpers(server: ViteDevServer) {
 	};
 }
 
+const SAFE_CHARS_RE = /[^\w.-]/g;
```
florian-lefebvre, line 103:
```suggestion
// Will match any invalid characters (will be converted to _). We only allow a-zA-Z0-9.-_
const SAFE_CHARS_RE = /[^\w.-]/g;
```

packages/astro/src/integrations/hooks.ts
florian-lefebvre: If we get rid of the code, the file will be generated at `.astro/integrations/astro_db/db-types.d.ts` which is not super pretty. Since `astro:db` is a core integration, there are already a few code smells here and there (eg. calling `getPackage`) so I think it's fine for now. I can add a comment to remember to remove it once `astro:db` is really part of core. Wdyt?

delucis: Does the path need to be pretty? I’d think the path doesn’t really matter too much.

packages/astro/src/core/sync/write-files.ts
ematipico: The entire body of the function `setUpEnvTs`, because I see various FS operations here and in other non-exported function. In the end, the reason is the same, so we can handle all of them with only one single try/catch block

ematipico:
Regarding formatting:

I thought the formatting was actually running a somewhat prettification, but instead, we use Babel. With Babel, there's a parsing phase and a transformation phase, which are expensive. If a user has 20 integrations, and these integrations generate big files, that risk causes Astro to be slow.

I believe formatting should be left to the integrations developers.

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
...
+
+```diff
+- /// <reference types="astro/client" />
++ /// <reference path="../.astro/types.d.ts" />
+- /// <reference path="../.astro/env.d.ts" />
+- /// <reference path="../.astro/actions.d.ts" />
+```
```
bluwy, line 26: Question: Would this be a breaking change for users who run `astro check`? I think some people run that before `astro build`, and that may break their build pipeline.

florian-lefebvre: I think it would not be breaking. To be clear:
- it's fine to keep `astro/client`, it would just be duplicated
- `../.astro/types.d.ts` is not breaking since it already exists for CC
- As for any other core reference, it may break but since those features are marked as experimental I think it's fine
- As for `astro:db`, it may break as well?

@Princesseuh do you know if astro check would complain?

packages/astro/src/core/sync/write-files.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { Logger } from '../logger/core.js';
+import { REFERENCE_FILE } from './constants.js';
+import { AstroError, AstroErrorData } from '../errors/index.js';
+
+export async function writeFiles(settings: AstroSettings, fs: typeof fsMod, logger: Logger) {
+	try {
+		writeInjectedTypes(settings, fs);
+		await setUpEnvTs(settings, fs, logger);
+	} catch (e) {
+		throw new AstroError(AstroErrorData.UnknownError, { cause: e });
```
florian-lefebvre, line 16: @ematipico what error should I throw here?

.changeset/mean-horses-kiss.md
Princesseuh: `astro check` runs the sync step first, so if the syncing "fix" the types then it'll be fine

florian-lefebvre: Outdated declarations are currently not removed

packages/astro/src/core/sync/write-files.ts
ematipico: I think a generic error that says that Astro couldn't read or write files to disk should be fine. The `cause` should give us the stacktrace of the source, which should be enough for us. 

You can create a new Astro error if you want.

.changeset/mean-horses-kiss.md
Princesseuh: Ah, but okay. TypeScript does not report an error for triple slashes references that point to non-existing files, so there shouldn't be an error.

bluwy: Thanks for the explanation. Then it seems fine then. Since the db package is also marked a minor (for v0), it's already a breaking release so changing the types path should be fine.

ematipico (approved):

packages/astro/src/content/types-generator.ts
```diff
@@ -515,8 +515,16 @@ async function writeContentFiles({
 		contentConfig ? `typeof import(${configPathRelativeToCacheDir})` : 'never',
 	);
 
-	await fs.promises.writeFile(
-		new URL(CONTENT_TYPES_FILE, settings.dotAstroDir),
-		typeTemplateContent,
-	);
+	if (settings.injectedTypes.some((t) => t.filename === CONTENT_TYPES_FILE)) {
```
Princesseuh, line 518: What does this do?

florian-lefebvre: If it's the first time, we inject types the usual way. `sync` will handle creating files and references. If it's not the first time, we just override the dts content. Does it make sense to you?

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
+---
```
Princesseuh, line 1: Would love to see an example of an integration using the hook in here

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
...
+
+Adds a new `injectTypes()` utility to the Integration API and refactors how type generation works
+
+Integrations can now use a new `injectTypes()` utility in the `astro:config:done` hook. This utility allows an integration to inject types into a user's project. `filename` must end with `".d.ts"` and `content` must be valid TypeScript (it will be formatted).
+
+Under the hood, it will create a file at `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts` and create references to it. `injectTypes()` returns a URL to the normalized path.
+
+```js
+const path = injectTypes({
```
Princesseuh, line 12: What's the point of it returning the path?

packages/astro/src/content/types-generator.ts
Princesseuh: I think a comment would be great on top just explaining that would be great! This file is already quite tough to read IMO

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1291,6 +1291,16 @@ export const RewriteWithBodyUsed = {
 		'Astro.rewrite() cannot be used if the request body has already been read. If you need to read the body, first clone the request.',
 } satisfies ErrorData;
 
+/**
+ * @docs
+ * @description
+ * An unknown error occured while reading or writing files to disk.
```
Princesseuh, line 1297: Some examples of why this can happen would be great, just to give a hint to the user. You can also mention that more information is available in the `cause`

packages/astro/src/content/types-generator.ts
florian-lefebvre:
```suggestion
	// If it's the first time, we inject types the usual way. sync() will handle creating files and references. If it's not the first time, we just override the dts content
	if (settings.injectedTypes.some((t) => t.filename === CONTENT_TYPES_FILE)) {
```

.changeset/mean-horses-kiss.md
florian-lefebvre: See what we're doing in core for content sync (ie. overriding the dts file with content updates)? I can imagine some people wanting to do this in integrations. Since we normalize the filename, people can't easily construct the path themselves so we give it to them

packages/astro/src/core/errors/errors-data.ts
florian-lefebvre: Have examples?

.changeset/mean-horses-kiss.md
florian-lefebvre: Isn't it already the case with the snippet below?

packages/astro/src/integrations/hooks.ts
```diff
@@ -100,6 +100,18 @@ export function getToolbarServerCommunicationHelpers(server: ViteDevServer) {
...
 
+// Will match any invalid characters (will be converted to _). We only allow a-zA-Z0-9.-_
+const SAFE_CHARS_RE = /[^\w.-]/g;
+
+export function normalizeInjectedTypeFilename(filename: string, integrationName: string): string {
+	if (!filename.endsWith('.d.ts')) {
+		throw new Error(
+			`Integration ${bold(integrationName)} is injecting a type that does not end with "${bold('.d.ts')}"`,
+		);
+	}
```
Princesseuh, line 111: Is there a world where this API should handle all kinds of filegen? I'm just thinking for instance of the new content intellisense stuff, none of the files being generated are d.ts, right.

packages/astro/src/core/errors/errors-data.ts
Princesseuh: There's a few common errors that can happen, like missing permissions, or a file not existing when it's supposed to.

Notably permissions are sometimes an head scratcher for Windows users (ex: they put their Astro project in a weird folder)

packages/astro/src/integrations/hooks.ts
florian-lefebvre: I had been thinking about an `injectAsset()` API but that will be completely distinct. I think it's fine to keep `injectTypes` focused on dts/references/types and have something else later for the rest (eg. astro db)

matthewp (approved):

.changeset/mean-horses-kiss.md
Princesseuh: I mean in the actual integration hook, the snippet doesn't show where that function comes from. You can amend the snippet below too, no need for a new example necessarily

packages/astro/src/core/dev/restart.ts
```diff
@@ -40,17 +39,19 @@ function shouldRestartContainer(
...
 	// Otherwise, watch for any astro.config.* file changes in project root
 	else {
-		const normalizedChangedFile = vite.normalizePath(changedFile);
 		shouldRestart = configRE.test(normalizedChangedFile);
-
-		if (preferencesRE.test(normalizedChangedFile)) {
+		const settingsPath = vite.normalizePath(
+			fileURLToPath(new URL('settings.json', settings.dotAstroDir)),
+		);
+		if (settingsPath.match(normalizedChangedFile)) {
```
bluwy, line 54: What is this trying to do here? It's calling `str1.match(str2)`

bluwy (approved):

packages/astro/src/core/dev/restart.ts
florian-lefebvre: I tried a few things and it was not that easy. Basically, I think it's more reliable to use `dotAstroDir` in case it becomes configurable, or changes. But then, I had to find a way to make both paths match but some had to normalize stuff etc. Anyways, I tried a bunch of combinations but that's the only one I found was working. If you have suggestions, happy to apply them!

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,26 @@
...
+  filename: "types.d.ts",
+  content: "declare module 'virtual:integration' {}"
+})
+console.log(path) // URL
```
florian-lefebvre, line 16:
```suggestion
// my-integration/index.js
export default {
  name: 'my-integration',
  'astro:config:done': ({ injectTypes }) => {
    injectTypes({
      filename: "types.d.ts",
      content: "declare module 'virtual:my-integration' {}"
    })
  }
};
```

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,31 @@
+---
+'astro': minor
+---
+
+Adds a new `injectTypes()` utility to the Integration API and refactors how type generation works
+
+Integrations can now use a new `injectTypes()` utility in the `astro:config:done` hook. This utility allows an integration to inject types into a user's project. `filename` must end with `".d.ts"` and `content` must be valid TypeScript (it will be formatted).
+
+Under the hood, it will create a file at `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts` and create references to it. `injectTypes()` returns a URL to the normalized path.
```
sarah11918, line 9:
```suggestion
Use `injectTypes()` in the `astro:config:done` hook to inject types into your user's project by adding a new a `*.d.ts` file.

The `filename` property will be used to generate a file at `/.astro/integrations/<normalized_integration_name>/<normalized_filename>.d.ts` and must end with `".d.ts"`.

The `content` property will create the body of the file, and must be valid TypeScript.

Additionally, `injectTypes()` returns a URL to the normalized path so you can overwrite its content later on, or manipulate it in any way you want.
```
How about pulling this from docs?

sarah11918 (commented):
Assuming we're happy with how the docs PR looks, we can pull that in here for the changeset! (Adjust to taste.)

.changeset/mean-horses-kiss.md
```diff
@@ -0,0 +1,35 @@
+---
+'astro': minor
+---
+
+Adds a new `injectTypes()` utility to the Integration API and refactors how type generation works
```
sarah11918, line 5:
```suggestion
Adds a new [`injectTypes()` utility](https://docs.astro.build/en/reference/integrations-reference/#injecttypes-options) to the Integration API and refactors how type generation works
```
Just adding a link where documentation will be!

packages/astro/src/core/dev/restart.ts
florian-lefebvre: Both are not exactly equal, I think one of them has a prefix like `file:///` or something

bluwy: I don't think it's possible for either to have `file://` from reading the code, but anyways it would be nice to move away from `.match` and use `.includes` or `.endsWith` if needed.

Issue #11458, opened 2024-07-12
Special characters excluded from content collection slug


nicolabovolato:
### Astro Info

```block
Astro                    v4.11.5
Node                     v20.15.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/preact
                         @astrojs/tailwind
                         @astrojs/mdx
                         @astrojs/partytown
                         @astrojs/sitemap
```

### Describe the Bug

On my personal website there's among others a basic tutorial for C and one for C++.

I'm currently migrating from pages to content collections, but il looks like the `+` sign gets stripped in the slug.

```
09:13:48 [ERROR] [DuplicateContentEntrySlugError] blog contains multiple entries with the same slug: `it/tutorial-c/introduzione`. Slugs must be unique.
Entries: 
- /src/content/blog/it/tutorial-c++/introduzione/index.mdx
- /src/content/blog/it/tutorial-c/introduzione/index.mdx
  Error reference:
    https://docs.astro.build/en/reference/errors/duplicate-content-entry-slug-error/
  Stack trace:
    at REDACTED/node_modules/astro/dist/content/vite-plugin-content-virtual-mod.js:213:19
```

### What's the expected result?

From what I found in the docs, there's no mention of any character stripping when generating the slug.
I think these character should be kept as it also provides better compatibility with Astro pages.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-ywpupp

### Participation

- [ ] I am willing to submit a pull request for this issue.

bgentry:
I believe I actually just ran into this as well. I'm trying to build a content collection for a software library that includes versioned URLs (so I can have docs for all versions). A `pkg` collection with content files like `src/content/pkg/v0.0.1/sub/index.mdoc` will have its slug returned as `v001/sub`. I can of course override that slug back to its "true" form of `v0.0.1/sub`, but I'll have to do that for every single file—definitely not ideal!

I would love to be able to opt out of sanitizing actually-supported characters out of my slugs in some fashion.

matthewp:
You can set your own custom slug on individual entries: https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs

I don't think we want to change our default slugging algorithm.

bgentry:
[...]

PR #11554, opened 2024-07-25
fix(astro): remove github slugger in content utils

syi0808:
## Changes
I removed the github slugger from content utils and astro, leaving only the behavior left in the comment.

## Testing
I tested whether getContentEntryIdAndSlug does not delete special characters.

## Docs
I don't know if this is something I should understand from a user's point of view. So I don't add docs.

Fix: #11458

### DIFF
--- a/.changeset/big-rules-jam.md
+++ b/.changeset/big-rules-jam.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Remove github slugger in content slug util
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -159,7 +159,6 @@
     "execa": "^8.0.1",
     "fast-glob": "^3.3.2",
     "flattie": "^1.1.1",
-    "github-slugger": "^2.0.0",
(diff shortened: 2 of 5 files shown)


syi0808:
> What is the reason for this change? What is wrong about github-slugger?

I think people want to design url with special characters or uft-8 characters ([see also issue](https://github.com/withastro/astro/issues/11458)), but github-slugger prevent it.
Additionally, the comment that the purpose of using github-slugger is for blank and case-sensitive purposes only, so we deleted it because we thought we didn't need to do anything unintended.

matthewp:
You can override the slug: https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs I don't think we should change our slugging algorithm, it can't accommodate every desire which is why it's configurable.

syi0808:
I agree, and I'm also not sure about the side effects, but I think the most important thing is whether we are intended to exclude special characters.

I think the slug override was intended to be set separately from the file path, so even though it's like the file path as in the current issue, it might feel a little hacky or not understanding the intent.

PR #11559, opened 2024-07-26
actions: check result.data is not undefined instead of truthy

### LINKED ISSUE #11549: Server actions cause an error when returning a value of 0.
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

When returning the number 0 from a server action an error occurs:

`Response constructor: Invalid response status code 204`

Cause appears to be this falsy check:
https://github.com/withastro/astro/blob/7f26de906e87f1e8973a1f84399f23e36e506bb3/packages/astro/src/actions/runtime/route.ts#L43

Should this check for only `null` and `undefined` values?

A work around is to wrap the value in an object literal e.g. `{ number: 0 }`

The action returns the value 0.

https://stackblitz.com/edit/github-fozayp?file=src%2Fpages%2Findex.astro

### Participation

bryanwood:
## Changes

Allows actions to return falsy values by changing the result.data check to a strict inequality check for undefined. Fixes #11549

## Testing

- Added tests for 0 and false return values.

## Docs

I don't think this requires docs, reading them I'd expect returning a value of 0 or false to work.

### DIFF
--- a/.changeset/tasty-rockets-jog.md
+++ b/.changeset/tasty-rockets-jog.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Let actions return falsy values without an error
+Allows actions to return falsy values without an error
--- a/.changeset/tasty-rockets-jog.md
+++ b/.changeset/tasty-rockets-jog.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Let actions return falsy values without an error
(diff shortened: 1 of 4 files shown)


bholmesdev (approved):
Ah, good catch! `define`tly ready to merge

PR #11546, opened 2024-07-24
fix(types): update Astro.redirect JSDoc comment

ArmandPhilippot:
## Changes

Update the inline documentation (JSDoc comment) for `Astro.redirect`:
* Closes #11544

I haven't added a changeset since it is a small change. Should I?

## Testing

No test. I only reworded a comment.

## Docs

No doc needed since it does not change the behavior. However it could be a nice addition to clarify the difference between `server` and `static` with `Astro.redirect` (I can post a suggestion on `astro/docs` separately of this PR).

PR #11561, opened 2024-07-26
fix(types): change page.size to 10 in JSDoc comment

### LINKED ISSUE #11560: Inaccurate JSdoc comment for `page.size` in astro/@types
### Astro Info

```block
Package Manager          npm
Adapter                  none
Integrations             none
```

_No response_

### Describe the Bug

I was reading the [Routing guide](https://docs.astro.build/en/guides/routing/#complete-api-reference) and I noticed an error which leads me to `astro/@types` package.

The Page interface in [`astro/@types`](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/%40types/astro.ts#L2707-L2708) indicates that the default for `page.size` is `25`. It seems to be wrong:
* see [paginate.ts#L19](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/core/render/paginate.ts#L19) and [paginate.ts#L80](https://github.com/withastro/astro/blob/02c85b541241a07db45bf9e15717e111104898e5/packages/astro/src/core/render/paginate.ts#L80): `page.size` uses the given `pageSize` option in `paginate()` function or `10`.
* [Stackblitz reproduction](https://stackblitz.com/edit/astro-paginate-pagesize?file=src%2Fpages%2Fpokemon%2F[page].astro&on=stackblitz) (click on `Pokemon - Page 1` in the preview)

```diff
/**
 * Represents a single page of data in a paginated collection
 *
 * [Astro reference](https://docs.astro.build/en/reference/api-reference/#the-pagination-page-prop)
 */
export interface Page<T = any> {
	/** result */
	data: T[];
	/** metadata */
	/** the count of the first item on the page, starting from 0 */
	start: number;
	/** the count of the last item on the page, starting from 0 */
	end: number;
	/** total number of results */
	total: number;
	/** the current page number, starting from 1 */
	currentPage: number;
-	/** number of items per page (default: 25) */
+	/** number of items per page (default: 10) */
	size: number;
	/** number of last page */
	lastPage: number;
	url: {
		/** url of the current page */
		current: string;
		/** url of the previous page (if there is one) */
		prev: string | undefined;
		/** url of the next page (if there is one) */
		next: string | undefined;
		/** url of the first page (if the current page is not the first page) */
		first: string | undefined;
		/** url of the next page (if the current page in not the last page) */
		last: string | undefined;
	};
}
```

https://stackblitz.com/edit/astro-paginate-pagesize?file=src%2Fpages%2Fpokemon%2F[page].astro&on=stackblitz

### Participation

ArmandPhilippot:
## Changes

- Update the JSDoc comment for `page.size` to use the accurate `pageSize` default (`10` and not `25`).
- Closes #11560

## Testing

No test needed, it's just a comment update. You can check the [reproduction](https://stackblitz.com/edit/astro-paginate-pagesize?file=src%2Fpages%2Fpokemon%2F[page].astro&on=stackblitz) to see that `10` is the correct number.

## Docs

An update to the [Routing guide](https://docs.astro.build/en/guides/routing/#complete-api-reference) will be needed since it reuse the `Page` interface (and maybe also an update to API reference). This is already reported in https://github.com/withastro/docs/issues/8929.

### DIFF
--- a/.changeset/good-bats-add.md
+++ b/.changeset/good-bats-add.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Use the correct pageSize default in `page.size` JSDoc comment
--- a/packages/astro/src/@types/astro.ts
+++ b/packages/astro/src/@types/astro.ts
@@ -2704,7 +2704,7 @@ export interface Page<T = any> {
 	total: number;
 	/** the current page number, starting from 1 */
 	currentPage: number;
-	/** number of items per page (default: 25) */
+	/** number of items per page (default: 10) */
 	size: number;
 	/** number of last page */
 	lastPage: number;


.changeset/good-bats-add.md
bluwy, line 1: I don't think we need a changeset for this as it's usually reserve for code behaviour changes. It should be fine to remove this here and we can release the fix in the next release.

ArmandPhilippot: @bluwy I don't mind removing it but yesterday it was needed for a similar change. See https://github.com/withastro/astro/pull/11546#pullrequestreview-2197046078 ... So should I remove it?

bluwy: Hmm, looks like we have different ideas for changesets 😄 Nevermind then let's keep the changeset, I don't think it's a big deal

.changeset/good-bats-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Use the correct pageSize default in `page.size` JSDoc comment
```
bluwy, line 5:
```suggestion
Uses the correct pageSize default in `page.size` JSDoc comment
```

bluwy (approved):

sarah11918:
Thanks @ArmandPhilippot @bluwy @ematipico  for handling this! Just to weigh in:

- we usually only use a changeset for docs when the content is corrected (as it was here) because this text occurs in the editor on hover, people might be consulting it to figure out how to use it etc. So, fixing actually WRONG information is considered helpful to ship in a patch and get people's editors updated and have a chance for the changeset to point out that the previous guidance was wrong.

- for something like a typo that doesn't exactly *correct* incorrect information, I tend not to use one. (And then Chris tells me when it seems important enough to actually do have one.)  :smile: 

For the changeset content itself, calling out that this is specifically a DOCS FIX is helpful, I think. See the last changeset I made, for example:

> Fixes typo in documenting the `slot="fallback"` attribute for Server Islands experimental feature.`

That way, people who are scanning can CLEARLY tell this is just a docs fix. In this situation, I would have suggested something like: 

> Fixes an error where docs incorrectly stated the default value of `page.size`

(Will note for future reference that "Uses" isn't a particularly descriptive/active verb for letting people know whether e.g. something was "fixed", "added", "removed", "improved", "refactored". Used is about what YOU did. Readers need to know what this change means for THEM, and the new state of the project as it relates to their code! :wink: )

ArmandPhilippot:
Thanks @sarah11918 for taking the time to write this detailed explanation! 👏🏽  I better understand when to use a changeset now and your note on the wording makes perfect sense. Your suggestion is indeed much better in a changelog. I'll try to be more descriptive the next time!

PR #11562, opened 2024-07-26
feat: use devalue to serialize content layer data

ascorbic:
## Changes

Switches content layer to use devalue for the data store serialization, so it supports Dates etc

## Testing
Added a test

### DIFF
--- a/packages/astro/src/content/consts.ts
+++ b/packages/astro/src/content/consts.ts
@@ -25,5 +25,5 @@ export const CONTENT_FLAGS = [
 
 export const CONTENT_TYPES_FILE = 'types.d.ts';
 
-export const DATA_STORE_FILE = 'data-store.json';
+export const DATA_STORE_FILE = 'data-store-module.mjs';
 export const ASSET_IMPORTS_FILE = 'assets.mjs';
--- a/packages/astro/src/content/consts.ts
+++ b/packages/astro/src/content/consts.ts
@@ -25,5 +25,5 @@ export const CONTENT_FLAGS = [
 
 export const CONTENT_TYPES_FILE = 'types.d.ts';
 
-export const DATA_STORE_FILE = 'data-store-module.mjs';
+export const DATA_STORE_FILE = 'data-store.json';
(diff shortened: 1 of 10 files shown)


matthewp:
Why can't we serialize Dates as ISO strings? I'm worried that the big perf gain of this feature is going to be erased if we start making it be JS again.

ascorbic:
@matthewp the problem is that zod transform and coerce allows values to be converted to `Date` objects. If we serialise these as strings, we won't be able to convert them back and the types will be wrong.

ascorbic:
I've benchmarked it, and it makes no appreciable difference to the performance when building 10k pages. It's all in one module, so it doesn't have the problem with the old implementation which was having one module per entry.

matthewp:
Won't we run into this problem when we do a sqlite backed storage anyways? If so it's worth tackling it now, imo.

matthewp:
Could we save the raw, untransformed value so that when we rehydrate with Zod it gets us back to where we want to be?

ascorbic:
Do we have access to the schema at runtime? I transformed before insertion because I didn't think we have access to the schema then. For the sqlite backend we'd want to preserve types too if possible.

ascorbic:
If you're concerned about the use of JS, we could serialise to devalue JSON instead, though I think that's a bit slower.

ematipico:
My gut tells me we should prefer a JSON format over JavaScript

matthewp:
Yes, avoiding JS is my biggest issue here. So if using devalue JSON unblocks then that works for me for now. It does feel like a little bit of architectural problem if we don't have full control over serialization though. Maybe that's something we can fix later.

ascorbic:
OK, makes sense. I'll benchmark that and assuming it's ok then I'll update the PR to use it.

ascorbic:
So it turns out it's actually slightly _faster_ to use devalue stringify (i.e. write to JSON rather than JS). Just a matter of 150ms or so with 10k entries, but that will teach me to not do my own benchmarks.

matthewp (approved):

Issue #11538, opened 2024-07-24
Dev Server: Astro tries to rewrite Error messages and sometimes fails at it


rburgst:
### Astro Info

```block
Astro                    v4.11.5
Node                     v20.11.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/preact
                         astro-compressor
```

### Describe the Bug

In `pnpm dev` when a DOMException occurs (e.g. fetch timeout), then astro tries to run this code

https://github.com/withastro/astro/blob/536209aa74924dbde4084026e3b72832961813ac/packages/astro/src/core/errors/dev/utils.ts#L74-L75

For `DOMExceptions` the field `message` is a readonly property which produces the following error
```
09:23:36 [ERROR] [UnhandledRejection] Astro detected an unhandled rejection. Here's the stack trace:
TypeError: Cannot set property message of  which has only a getter
    at eval (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/core/errors/dev/utils.js:62:21)
    at Array.forEach (<anonymous>)
    at Module.collectErrorMetadata (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/core/errors/dev/utils.js:34:7)
    at Module.recordServerError (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/vite-plugin-astro-server/error.js:26:55)
    at onError (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/vite-plugin-astro-server/request.js:68:54)
    at Module.runWithErrorHandling (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/vite-plugin-astro-server/controller.js:85:19)
    at async Module.handleRequest (file:///home/projects/xoriwgxvly.github/node_modules/astro/dist/vite-plugin-astro-server/request.js:49:3)
  Hint:
    Make sure your promises all have an `await` or a `.catch()` handler.
  Error reference:
    https://docs.astro.build/en/reference/errors/unhandled-rejection/
  Stack trace:
    at /home/projects/xoriwgxvly.github/node_modules/astro/dist/core/errors/dev/utils.js:62:21
    [...] See full stack trace in the browser, or rerun with --verbose.
09:23:39 [watch] src/pages/index.astro
09:24:05 [watch] src/pages/index.astro
09:24:06 [watch] src/pages/index.astro

```

### What's the expected result?

[...]

PR #11566, opened 2024-07-29
fix: check before writing to errors

Princesseuh:
## Changes

Unusually, some errors have some of their properties read-only. Since this has only happened once in the history of Astro, I figure it wasn't worth it to shield *everything*, but just adding checks for the common properties should be enough

Fixes https://github.com/withastro/astro/issues/11538

## Testing

Added a test

## Docs

N/A

### DIFF
--- a/packages/astro/e2e/errors.test.js
+++ b/packages/astro/e2e/errors.test.js
@@ -125,4 +125,10 @@ test.describe('Error display', () => {
 		const message = (await getErrorOverlayContent(page)).message;
 		expect(message).toMatch('can only be used in');
 	});
+
+	test('can handle DomException errors zzz', async ({ page, astro }) => {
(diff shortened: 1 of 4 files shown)


packages/astro/e2e/errors.test.js
```diff
@@ -125,4 +125,10 @@ test.describe('Error display', () => {
 		const message = (await getErrorOverlayContent(page)).message;
 		expect(message).toMatch('can only be used in');
 	});
+
+	test('can handle DomException errors zzz', async ({ page, astro }) => {
```
ascorbic, line 129: Are these zzzs meant to be here?

Princesseuh: No, you've just found a Erika-ism.

PR #11570, opened 2024-07-29
Actions: New fallback behavior with `action={actions.name}`

bholmesdev:
## Changes

Introduce the new pattern `action={actions.name}` for form fallbacks. This replaces the previous `getActionProps()` pattern.

```astro
---
---

<form action={actions.logOut}>
  <button>Log Out</button>
</form>
```

- Deprecate `getActionProps()`
- Allow actions to return a string
- Introduce `.queryString` as an action property for more explicit URL construction
- Update React 19 handler to also use a query string for form actions

## Testing

- Add integration tests for new pattern, marking old tests as legacy
- Check that e2e tests still pass

## Docs

RFC updates to be pushed on minor release


bholmesdev:
Reason for the minor: this is a breaking change for the React integration. It's otherwise non-breaking since `getActionProps()` still has a legacy handler.

matthewp:
My only pause here is will people understand that the POST will return to this current Astro page? I might find that confusing without knowing. I guess if people do then we could add the `onSuccess` / `onError` options like previously discussed.

bholmesdev:
@matthewp Totally understand the concern. This API was chosen due to the precedent React and SvelteKit have set for actions returning to the current page. I'd expect the pattern to feel nicer for our majority userbase.

We are waiting another 2 weeks for a minor release, so that should give us time to get user feedback on the change. I will also rework the RFC docs to better explain the pattern for those that (inevitably) use this in a `.astro` file.

matthewp (approved):

Issue #11515, opened 2024-07-20
Astro Actions .safe() function required empty object argument for Vercel deployment


Grahf0085:
### Astro Info

```block
Astro                    v4.12.2
Node                     v18.20.3
System                   Linux (x64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/vercel/serverless
Integrations             @astrojs/solid-js
```

### Describe the Bug

I have this logout action:
`const { data, error } = await actions.logoutUser.safe()`

This woks fine on local host when running dev server and when using the node adapter and running the app using entry.mjs.

But when deployed to vercel I get this error when the logout function runs:

> 17:03:11 [ERROR] SyntaxError: Unexpected end of JSON input at JSON.parse (<anonymous>) at parseJSONFromBytes (node:internal/deps/undici/undici:5584:19) at successSteps (node:internal/deps/undici/undici:5555:27) at fullyReadBody (node:internal/deps/undici/undici:1665:9) at process.processTicksAndRejections (node:internal/process/task_queues:95:5) at async specConsumeBody (node:internal/deps/undici/undici:5564:7) at async Module.POST (file:///var/task/vercel/path0/.vercel/output/_functions/chunks/route_Yu3iUWSv.mjs:17:12) at async renderEndpoint (file:///var/task/vercel/path0/.vercel/output/_functions/chunks/astro/server_CBvJsTK-.mjs:45:20) at async lastNext (file:///var/task/vercel/path0/.vercel/output/_functions/entry.mjs:1031:23) at async callMiddleware (file:///var/task/vercel/path0/.vercel/output/_functions/entry.mjs:480:10) 

It's fixed by adding an empty object as a argument to the .safe() function:
`const { data, error } = await actions.logoutUser.safe({})`

Not sure if it's document somewhere that an empty object needs to be passed to .safe() when it has no other arguments or if this is a bug.

To reproduce you will need to go to my site (https://sharper-than-time-c1ti0fvty-grahf0085s-projects.vercel.app/), make an account, and click the logout button. I can only produce this on vercel so.....

This is my logout action if it matters:
  ```
logoutUser: defineAction({
    handler: (_, context) => {
      const TOKEN = import.meta.env.TOKEN

      try {
        context.cookies.set(TOKEN, '', {
          httpOnly: true,
          maxAge: 0,
          path: '/',
          secure: true,
          sameSite: 'lax',
        })

        return { cookieDeleted: true }
      } catch (error) {
[...]

PR #11571, opened 2024-07-29
Actions: Make `.safe()` the default return value

bholmesdev:
## Changes

Make `.safe()` the default return value for actions. This means `{ data, error }` will be returned when calling an action directly. If you prefer to get the data while allowing errors to throw, chain the `.orThrow()` modifier.

```ts

// Before
const { data, error } = await actions.like.safe();
// After
const { data, error } = await actions.like();

// Before
const newLikes = await actions.like();
// After
const newLikes = await actions.like.orThrow();
```

- Update types and implementation to reverse this pattern
- Remove support for using unsafe action calls with React 19. Chaining `.orThrow()` will remove progressive enhancement, since exceptions must be caught on the client.

## Testing

Update test fixtures to use the new API pattern.

## Docs

RFC updates to be pushed on minor release

### DIFF
--- a/.changeset/light-chairs-happen.md
+++ b/.changeset/light-chairs-happen.md
@@ -1,7 +1,9 @@
 ---
-'@astrojs/react': minor
-'astro': minor
+'@astrojs/react': patch
+'astro': patch
 ---
+
+**BREAKING CHANGE to the experimental Actions API only**
 
 Make `.safe()` the default return value for actions. This means `{ data, error }` will be returned when calling an action directly. If you prefer to get the data while allowing errors to throw, chain the `.orThrow()` modifier.
 
--- a/.changeset/light-chairs-happen.md
+++ b/.changeset/light-chairs-happen.md
@@ -3,7 +3,7 @@
 'astro': patch
 ---
 
-**BREAKING CHANGE to the experimental Actions API only**
+**BREAKING CHANGE to the experimental Actions API only.** Install the latest `@astrojs/react` integration as well if you're using React 19 features.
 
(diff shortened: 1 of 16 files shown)


matthewp (approved):

.changeset/light-chairs-happen.md
```diff
@@ -0,0 +1,27 @@
+---
+'@astrojs/react': minor
+'astro': minor
```
ematipico, line 3: This one can be a `patch`, since this is a change to an experimental flag. Probably, the same should be for `@astrojs/react`

.changeset/light-chairs-happen.md
```diff
@@ -0,0 +1,27 @@
...
+---
+
+Make `.safe()` the default return value for actions. This means `{ data, error }` will be returned when calling an action directly. If you prefer to get the data while allowing errors to throw, chain the `.orThrow()` modifier.
```
ematipico, line 6: We should use this template for breaking changes inside experimental features 

https://contribute.docs.astro.build/docs-for-code-changes/experimental-feature-docs/#breaking-changes-while-still-experimental

bholmesdev: Ah ha, this is super helpful! I thought experimental packages followed the pre-1.0 rule:
- Major -> minor
- Minor -> patch

Didn't realize it was _always_ a patch. That makes life a bit simpler.

ematipico: Yes, experimental features are semver-free, so we quickly ship changes and fixes without problems.

ematipico (approved):

Princesseuh:
Does this interact in any way with this issue? https://github.com/withastro/astro/issues/11515

bholmesdev:
@Princesseuh Taking a look at that issue. This is switching a default, but wouldn't affect parsing behavior, so I don't believe so.

Princesseuh (approved):

PR #11574, opened 2024-07-30
fix: highlight line with error in the error overlay

Princesseuh:
## Changes

Shiki 1.0 removed the `lineOptions` property, so our line highlighting stopped working. For some reason, this didn't cause a type error.

## Testing

Added a test

## Docs

N/A

### DIFF
--- a/packages/astro/e2e/errors.test.js
+++ b/packages/astro/e2e/errors.test.js
@@ -131,4 +131,12 @@ test.describe('Error display', () => {
 		const message = (await getErrorOverlayContent(page)).message;
 		expect(message).toMatch('The operation was aborted due to timeout');
 	});
+
+	test('properly highlight the line with the error zzz', async ({ page, astro }) => {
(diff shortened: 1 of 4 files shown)


packages/astro/e2e/errors.test.js
```diff
@@ -131,4 +131,12 @@ test.describe('Error display', () => {
 		const message = (await getErrorOverlayContent(page)).message;
 		expect(message).toMatch('The operation was aborted due to timeout');
 	});
+
+	test('properly highlight the line with the error zzz', async ({ page, astro }) => {
```
bluwy, line 135:
```suggestion
	test('properly highlight the line with the error', async ({ page, astro }) => {
```

bluwy (approved):
Ah thanks for catching it. Didn't realize that `lineOptions` no longer worked.

packages/astro/e2e/errors.test.js
Princesseuh: I left it in there on purpose to make sure this PR was properly reviewed. (I did not)

PR #11579, opened 2024-07-31
Separate render function and merge content layer types

ascorbic:
## Changes

Changes `entry.render()` to `render(entry)`. This has several benefits:

- makes `entry` serialisable, allowing it to be passed to islands or returned from an endpoint
- makes it easier for libraries etc to provide their own render function
- simplifies types - no need to separate `experimental_data` from `experimental_content`

Because of the final point, this now also removes `experimental_data`. This will also provide a migration path when we leave experimental: if there is a `loader` defined, then it can set the type automatically. In the meantime, it now throws an error if the type is wrong when loader is present.

Fixes PLT-2370
Fixes PLT-2371

## Testing

Updated fixtures to use the new syntax

## Docs

Adds a new AstroError

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -17,7 +17,7 @@ import {
 	render as serverRender,
 	unescapeHTML,
 } from '../runtime/server/index.js';
-import { IMAGE_IMPORT_PREFIX } from './consts.js';
+import { CONTENT_LAYER_TYPE, IMAGE_IMPORT_PREFIX } from './consts.js';
 import { type DataEntry, globalDataStore } from './data-store.js';
 import type { ContentLookupMap } from './utils.js';
 type LazyImport = () => Promise<any>;
@@ -26,6 +26,13 @@ type CollectionToEntryMap = Record<string, GlobResult>;
 type GetEntryImport = (collection: string, lookupId: string) => Promise<LazyImport>;
 
 export function defineCollection(config: any) {
+	if (
+		('loader' in config && config.type !== CONTENT_LAYER_TYPE) ||
+		(config.type === CONTENT_LAYER_TYPE && !('loader' in config))
+	) {
+		// TODO: when this moves out of experimental, we will set the type automatically
+		throw new AstroError(AstroErrorData.ContentLayerTypeError);
+	}
 	if (!config.type) config.type = 'content';
 	return config;
 }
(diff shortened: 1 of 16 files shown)


packages/astro/src/core/errors/errors.ts
```diff
@@ -27,7 +27,7 @@ type ErrorTypes =
 	| 'AggregateError';
 
 export function isAstroError(e: unknown): e is AstroError {
-	return e instanceof AstroError;
+	return e instanceof AstroError || typeof e === 'object' && (e as AstroError).type === 'AstroError';
```
ascorbic, line 30: I'm not sure why, but it wasn't recognising the error thrown as an AstroError.

packages/astro/src/content/runtime.ts
```diff
@@ -26,6 +26,13 @@ type CollectionToEntryMap = Record<string, GlobResult>;
 type GetEntryImport = (collection: string, lookupId: string) => Promise<LazyImport>;
 
 export function defineCollection(config: any) {
+	if (
+		('loader' in config && config.type !== CONTENT_LAYER_TYPE) ||
+		(config.type === CONTENT_LAYER_TYPE && !('loader' in config))
+	) {
+		// TODO: when this moves out of experimental, we will set the type automatically
+		throw new AstroError(AstroErrorData.ContentLayerTypeError);
```
ematipico, line 34: If you already know if this error is going to be removed when we go out of the experimental, I suggest throwing a normal `Error`. Unfortunately, Astro errors can't be deleted once they are public (that's how the codegen of docs work). I made this very mistake with the rewrite feature

packages/astro/src/content/runtime.ts
```diff
@@ -433,7 +440,14 @@ function updateImageReferencesInData<T extends Record<string, unknown>>(
 	});
 }
 
-async function renderEntry(entry?: DataEntry) {
+export async function renderEntry(
+	entry?: DataEntry | { render: () => Promise<{ Content: AstroComponentFactory }> }
```
ematipico, line 444: It feels weird to have a public function called `renderEntry` where the `entry` is optional

ematipico (approved):

packages/astro/src/content/runtime.ts
matthewp: Yeah same, if we are making this public we can refactor the API just a tad.

packages/astro/src/content/runtime.ts
ascorbic: Changed to an `AstroUserError` so that when it's caught it propagates our custom message and hint.

PR #11580, opened 2024-07-31
Update errors-data.ts for Astro.rewrite release in 4.13

sarah11918:
## Changes

This updates a link in an error message to point to the new docs for `Astro.rewrite()`, as part of 4.13 docs.

## Testing

no tests!

## Docs

all docs!


Princesseuh (changes requested):
Errors should not be removed, they should be deprecated using the `@deprecated` tag

> When an error is no longer relevant, it should be deprecated, not removed.

bluwy (approved):

Princesseuh (approved):

PR #11581, opened 2024-07-31
Support --force flag in sync and dev

ascorbic:
## Changes

Adds support for the `--force` flag to `dev` and `sync`. It was already supported in `build`. 

Fixes PLT-2369


bluwy (approved):
Clean use of inline config!

PR #11584, opened 2024-07-31
Actions: Remove async local storage dependency

bholmesdev:
## Changes

Remove async local storage dependency from Astro Actions. This simplifies setup for environments that either do not support async local storage, or require an opt-in for the functionality:
- Stackblitz environments now support `getActionResult()` calls
- Cloudflare no longer requires special flags or `vite.ssr.external` patches

This update requires one breaking change: actions can no longer be called directly from server code. They must now be wrapped with `Astro.callAction()`. This applies the API context using `.bind()` on the function passed in.
- Update internals to use `.bind()` instead of relying on async local storage
- add `callAction()` to the API context type
- add new error classes for misuse of actions on the server

## Testing

- Add type test for new `Astro.callAction()` function
- Update server function tests to use `Astro.callAction()` function

## Docs

Updated roadmap with new `callAction()` function


ematipico (approved):

packages/astro/src/core/errors/errors-data.ts
```diff
@@ -1663,6 +1663,17 @@ export const UnsupportedConfigTransformError = {
...
 
+/**
+ * @docs
+ * @description
+ * Action called from a server page or endpoint without using `Astro.callAction()`.
+ */
+export const ActionCalledFromServerError = {
+	name: 'ActionCalledFromServerError',
+	title: 'Action unexpected called from the server.',
+	message: `Action called from a server page or endpoint without using \`Astro.callAction()\`. See the RFC section on server calls for usage instructions: https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#call-actions-directly-from-server-code`,
```
Princesseuh, line 1674: I'd suggest moving the "See the RFC section" part in a hint

packages/astro/test/types/call-action.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, it } from 'node:test';
+import { expectTypeOf } from 'expect-type';
+import { type ActionReturnType, defineAction } from '../../dist/actions/runtime/virtual/server.js';
+import { z } from '../../zod.mjs';
+import type { APIContext } from '../../src/@types/astro.js';
```
Princesseuh, line 5: I suspect that your imports here are the reason why the type tests don't pass, importing from the project tend to result in the TypeScript project's scope adding things it shouldn't

bholmesdev: Ah ha, that's probably it. Thanks!

PR #11591, opened 2024-08-01
fix: clear content layer cache if config has changed

ascorbic:
## Changes

- calculates a digest of the content config file and adds it to the config object
- when the data store is instantiated, checks if the value has changed, and if it has then it clears the data store
- stores the new digest value in the meta store

Fixes PLT-2372
Fixes PLT-2374

## Testing

Added a test

### DIFF
--- a/packages/astro/src/content/vite-plugin-content-imports.ts
+++ b/packages/astro/src/content/vite-plugin-content-imports.ts
@@ -156,8 +156,12 @@ export const _internal = {
 						const entryType = getEntryType(entry, contentPaths, contentEntryExts, dataEntryExts);
 						if (!COLLECTION_TYPES_TO_INVALIDATE_ON.includes(entryType)) return;
 
+						// The content config could depend on collection entries via `reference()`.
 						// Reload the config in case of changes.
-						await reloadContentConfigObserver({ fs, settings, viteServer });
+						// Changes to the config file itself are handled in types-generator.ts, so we skip them here
+						if (entryType === 'content' || entryType === 'data') {
+							await reloadContentConfigObserver({ fs, settings, viteServer });
+						}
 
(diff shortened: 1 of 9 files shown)


packages/astro/src/content/vite-plugin-content-imports.ts
```diff
@@ -156,11 +156,8 @@ export const _internal = {
 						const entryType = getEntryType(entry, contentPaths, contentEntryExts, dataEntryExts);
 						if (!COLLECTION_TYPES_TO_INVALIDATE_ON.includes(entryType)) return;
 
-						// The content config could depend on collection entries via `reference()`.
 						// Reload the config in case of changes.
-						if (entryType === 'content' || entryType === 'data') {
-							await reloadContentConfigObserver({ fs, settings, viteServer });
-						}
+						await reloadContentConfigObserver({ fs, settings, viteServer });
```
ascorbic, line 160: I'm not sure why this was previously set to not reload on config changes, but this is needed now.

bluwy: Looks like it was being handled at https://github.com/withastro/astro/blob/2abc38babfe5a647882251e578f48450565b0988/packages/astro/src/content/types-generator.ts#L146-L150, but I don't know why it's split out.

ascorbic: Weird. Thanks!

ascorbic: Added the conditional back, and added a comment explaining where it is

bluwy (approved):
Looks great! I think it'd be nice if the content sync stuff can be kept in a single place, but we can refactor that in the future.

Issue #11423, opened 2024-07-06
Astro Actions: "Error: Unable to get API context." when is sended arguments to an action.


ynhhoJ:
### Astro Info

```block
Astro                    v4.11.5
Node                     v18.20.3
System                   Linux (x64)
Package Manager          unknown
Output                   server
Adapter                  @astrojs/node
Integrations             @astrojs/preact
```

### If this issue only occurs in one browser, which browser is a problem?

Firefox

### Describe the Bug

I followed example from [Actions Proposal](https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md) and everytime when is called an action with arguments like:
```js
await actions.like({ postId })
```

Is returned error: ` Uncaught (in promise) Error: Unable to get API context.`  and `500` HTTP Status

But if we do:
```js
await actions.like()
```

We get `400` HTTP Status and validation input error.
![image](https://github.com/withastro/astro/assets/22500212/47347d56-3c1e-4b90-b39d-8dc5883c2cfd)

### What's the expected result?

There is no error when action:

```js
await actions.like({ postId })
```

is called

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-pdghi3?file=src%2Fcomponents%2FLike.tsx

### Participation

- [ ] I am willing to submit a pull request for this issue.

florian-lefebvre:
Yeah I think the 500 error should not occur here but be a 400 instead. @bholmesdev wdyt?

bholmesdev:
Thanks for reporting @ynhhoJ! I just confirmed this a Stackblitz issue, but does not occur locally. Is that true for you as well? That 500 error is indeed unexpected

ynhhoJ:
@bholmesdev , originally, on on my machine:
```
Astro                    v4.11.5
Node                     v20.15.0
System                   Linux (x64)
Package Manager          bun
Output                   server
Adapter                  @astrojs/node
Integrations             i18n:custom
                         @astrojs/tailwind
                         @astrojs/preact
```
I got another error `Unexpected end of JSON input`, every time when `action` has a body payload. So this is the cause why I opened this issue.

It's very strange:

![image](https://github.com/withastro/astro/assets/22500212/7fce9205-5abd-4bbc-8642-957c43ce62b9)
![image](https://github.com/withastro/astro/assets/22500212/04a83252-ff62-4122-b3ca-1079d5cc3f0c)

[...]

PR #11599, opened 2024-08-02
Actions: remove `request.clone()` to fix Bun issue

bholmesdev:
## Changes

- Resolves https://github.com/withastro/astro/issues/11423

This remove an unnecessary `request.clone()` on the Actions route handler. It seems the Bun runtime incorrectly loses the body when calling `.clone()` on a request.

I don't want to open PRs for bundlers we don't _officially_ support. Still, the `request.clone()` was unnecessary and could add additional latency, so it felt safe to remove.

## Testing

Manual testing on a sample app using `bun --bun dev`

## Docs

N/A

### DIFF
--- a/.changeset/nervous-mirrors-invite.md
+++ b/.changeset/nervous-mirrors-invite.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes "Unexpected end of JSON input" error when using Actions with the Bun runtime.
--- a/packages/astro/src/actions/runtime/route.ts
+++ b/packages/astro/src/actions/runtime/route.ts
@@ -15,9 +15,9 @@ export const POST: APIRoute = async (context) => {
 	if (contentLength === '0') {
 		args = undefined;
 	} else if (contentType && hasContentType(contentType, formContentTypes)) {
-		args = await request.clone().formData();
+		args = await request.formData();
 	} else if (contentType && hasContentType(contentType, ['application/json'])) {
-		args = await request.clone().json();
+		args = await request.json();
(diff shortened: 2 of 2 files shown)


bluwy:
IIUC we need `.clone()` so that we consume the body separately, and the following middlewares after the action middleware can later interact with the original response body if needed? In this PR, wouldn't that mean that the following middlewares won't be able to read the body? Or maybe I'm missing the flow.

ematipico:
They wouldn't be able to read the body anyway, because the actions middleware is at the last position.

bluwy:
That's good to know 👍 Then consuming the body directly should be safe then

bluwy (approved):

ascorbic:
Is there a scenario where a rewrite could occur after this? If so, the body may be needed for that.

bholmesdev:
@ascorbic that's a good question. This is only affecting the compiled `_actions` route that we control. The only way to return a rewrite would be from an action handler, and you currently cannot return a `Response` from an action (only JSON serializable output). We also do not have plans to allow this behavior.

It's a good note to revisit if we allow `Response` to be returned from an action though 👍

bholmesdev:
Well... investigating that failed windows e2e test, I realized the error of my ways. I hadn't considered that action handlers may also want to read `ctx.request.json()` or `ctx.request.formData()` directly. This may be useful for one-off handling of form inputs, for example. It seems our React utility to retrieve the current state value relies on this being supported.

Based on this, I'm closing the issue. Hope the Bun runtime can resolve the issue on their end!

PR #11600, opened 2024-08-02
fix(astro:content): add deprecated to `getEntryBySlug` & `getDataEntryById`

ArmandPhilippot:
## Changes

As [noted in the documentation](https://docs.astro.build/en/reference/api-reference/#getentrybyslug), `getEntryBySlug` is deprecated in favor of `getEntry`. @delucis [confirmed to me](https://github.com/withastro/docs/pull/8943#issuecomment-2263447606) that this also applied to `getDataEntryById`.

* Adds a `@deprecated` warning on these two functions to help consumers migrate to `getEntry`.

## Testing

Locally, the JSDoc comment is correctly displayed when hovering the functions so I don't think another test is needed.

## Docs

Docs is needed for `getDataEntryById` but this is already covered up by https://github.com/withastro/docs/pull/8943


.changeset/dull-taxis-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Adds deprecation warning on `getEntryBySlug` and `getDataEntryById` functions exported by `astro:content`.
```
ArmandPhilippot, line 5: After thinking about it, maybe this would be better:
```suggestion
Deprecates `getEntryBySlug` and `getDataEntryById` functions exported by `astro:content` in favor of `getEntry`.
```
This would let users know directly that they should look for `getEntry`.

delucis: Yes, I like this suggestion!

delucis (approved):
+1 on your changeset suggestion. Otherwise LGTM!

PR #11601, opened 2024-08-02
[error docs] adds section heading for Actions

sarah11918:
## Changes

Adds a new heading in `errors-data.ts` so that Action errors don't show up under Content Collection errors on the Docs Error reference index page.

Note: We'll probably add more headings/organization in the future so that "Astro Errors" isn't a dumping ground, but this clears up an actual problem with errors being in the wrong section!

## Testing

No tests, only docs

## Docs

Only updates the docs site!

### DIFF
--- a/packages/astro/src/core/errors/errors-data.ts
+++ b/packages/astro/src/core/errors/errors-data.ts
@@ -1620,6 +1620,12 @@ export const UnsupportedConfigTransformError = {
 
 /**
  * @docs
+ * @kind heading
+ * @name Action Errors
+ */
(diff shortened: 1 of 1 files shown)


delucis (approved):
LGTM! ✅ 

Do we document how the headings here work anywhere?

sarah11918:
We do not. :sweat_smile: 

But, I think we'll do a more thoughtful analysis of this whole page and get everything nicely sorted. Just dealing with the new Action Error added to docs yesterday.

PR #11603, opened 2024-08-02
Actions: auto-redirect action response to avoid "confirm form resubmission" dialog

bholmesdev:
## Changes

This change adds automatic redirects when an action is called. This follows Laravel's POST/redirect/GET pattern to avoid a "confirm form resubmission" dialog and remove the `?_astroAction` param from the rendered result.

[See the changeset](https://github.com/withastro/astro/blob/d3ba81a8a6d990dd0411dd18cb60d3383fdec7c9/.changeset/rotten-months-report.md) for usage instructions.

- Persist action data in a cookie. **This cookie is not encrypted.** We are investigating a general encryption strategy for Astro as part of Astro Islands: #11535. If Astro adopts an encryption key as part of the runtime, we will revisit encryption of the action result cookie.
- Redirect to referrer on error, and the `url.pathname` on success.
- If an action cookie is present, populate `getActionResult` with the cookie JSON. Delete after the request has been fulfilled.

## Testing

Update integration tests to follow an expected redirect.

## Docs

Updated RFC docs with new guide: https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#call-actions-from-an-html-form

### DIFF
--- a/.changeset/rotten-months-report.md
+++ b/.changeset/rotten-months-report.md
@@ -9,7 +9,6 @@ Improves user experience when render an Action result from a form POST request:
 
 Also improves the DX of directing to a new route on success. Actions will now redirect to the route specified in your `action` string on success, and redirect back to the previous page on error.
 
-
 For example, say you want to redirect to a `/success` route when `actions.signup` succeeds. You can add `/success` to your `action` string like so:
 
 ```astro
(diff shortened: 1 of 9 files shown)


bluwy (approved):
The code looks good to me. I suppose we want to wait for the docs change before merging?

ematipico:
@bholmesdev is there a chapter in the RFC I can look at before reviewing the code?

bholmesdev:
@ematipico Thanks for flagging that! I just finished a new guide on the RFC docs: https://github.com/withastro/roadmap/blob/actions/proposals/0046-actions.md#call-actions-from-an-html-form

Writing out examples actually helped me find a design flaw. I decided the cookie should _not_ be cleared on success. Otherwise, the action result data would disappear on a refresh.

ematipico (approved):

PR #11622, opened 2024-08-05
fix: skip glob files in content dir

ascorbic:
## Changes

Files in `src/content` are automatically processed by content collections, so the glob loader should not be used for these. This might not be immediately obvious, and it will be tempting to use that folder for data. This PR avoids that by skipping any globbed files in the content dir, and logging a warning.

Fixes PLT-2373

### DIFF
--- a/packages/astro/src/content/loaders/glob.ts
+++ b/packages/astro/src/content/loaders/glob.ts
@@ -229,9 +229,9 @@ export function glob(globOptions: GlobOptions): Loader {
 			if (skipCount > 0) {
 				logger.warn(`The glob() loader cannot be used for files in ${bold('src/content')}.`);
 				if (skipCount > 10) {
-					logger.warn(`Skipped ${green(skippedFiles.length)} files.`);
+					logger.warn(`Skipped ${green(skippedFiles.length)} files that matched ${green(globOptions.pattern)}.`);
 				} else {
-					logger.warn(`Skipped:`);
+					logger.warn(`Skipped the following files that matched ${green(globOptions.pattern)}:`);
 					skippedFiles.forEach((file) => logger.warn(`• ${green(file)}`));
 				}
 			}
--- a/packages/astro/src/content/loaders/glob.ts
+++ b/packages/astro/src/content/loaders/glob.ts
@@ -199,28 +199,29 @@ export function glob(globOptions: GlobOptions): Loader {
 				return fileUrl.href.startsWith(contentDir.href);
 			}
 
+			const configFiles = new Set(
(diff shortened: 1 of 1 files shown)


packages/astro/src/content/loaders/glob.ts
```diff
@@ -190,14 +190,51 @@ export function glob(globOptions: GlobOptions): Loader {
...
+			function isInContentDir(file: string) {
+				const fileUrl = new URL(file, baseDir);
+				return fileUrl.href.startsWith(contentDir.href);
+			}
+
+			function isConfigFile(file: string) {
+				const configFiles = ['config.js', 'config.ts', 'config.mjs'];
+				const fileUrl = new URL(file, baseDir);
+				return configFiles.some(
+					(configFile) => new URL(configFile, contentDir).href === fileUrl.href
```
bluwy, line 206: Could we hoist the `configFiles` beside `contentDir` and pre-compute their `URL`? I'm a bit worried that otherwise there's perf issues.

packages/astro/src/content/loaders/glob.ts
```diff
@@ -190,14 +190,51 @@ export function glob(globOptions: GlobOptions): Loader {
...
 			await Promise.all(
 				files.map((entry) =>
 					limit(async () => {
+						if (isConfigFile(entry)) {
+							return;
+						}
+						if (isInContentDir(entry)) {
+							skippedFiles.push(entry);
+							return;
+						}
```
bluwy, line 219: Can we move this one block upwards, e.g.:

```ts
files.map((entry) => {
  if (isConfigFile(entry)) {
    return
  }
  if (isInContentDir(entry)) {
    skippedFiles.push(entry);
    return;
  }
  return limit(async () => { ... })
})
```

That way there's less pressure for `p-limit` to track promises and lesser promises gets created too.

packages/astro/src/content/loaders/glob.ts
```diff
@@ -190,14 +190,51 @@ export function glob(globOptions: GlobOptions): Loader {
...
 						const entryType = configForFile(entry);
 						await syncData(entry, baseDir, entryType);
 					})
 				)
 			);
+
+			const skipCount = skippedFiles.length;
+
+			if (skipCount > 0) {
+				logger.warn(`The glob() loader cannot be used for files in ${bold('src/content')}.`);
```
bluwy, line 229: Would it also make sense to show the glob pattern used here?

packages/astro/src/content/loaders/glob.ts
ascorbic: Actually, that means I can use a `Set` instead of `some()` too.

bluwy (approved):
Looks great!

Issue #11615, opened 2024-08-04
Using `bind:` in Svelte 5 in conjunction with React causes error in MDX files.


Hugos68:
### Astro Info

```block
Astro                    v4.13.1
Node                     v20.14.0
System                   Windows (x64)
Package Manager          pnpm
Output                   static
Adapter                  none
Integrations             @astrojs/svelte
                         @astrojs/react
                         @astrojs/mdx
```

### If this issue only occurs in one browser, which browser is a problem?

Not relevant

### Describe the Bug

When running `astro build` the build unexpectedly fails and spits out the following error:
```txt
> astro build

2:29:22 PM [vite-plugin-svelte] You are using Svelte 5.0.0-next.208. Svelte 5 support is experimental, breaking changes can occur in any release until this notice is removed.
work in progress:
 - svelte-inspector is disabled until dev mode implements node to code mapping

14:29:22 [types] Generated 187ms
14:29:22 [build] output: "static"
14:29:22 [build] directory: C:\Users\hugok\repositories\repro-astro-svelte\dist\
14:29:22 [build] Collecting build info...
14:29:22 [build] ✓ Completed in 230ms.
14:29:22 [build] Building static entrypoints...
14:29:23 [vite] ✓ built in 1.45s
14:29:23 [build] ✓ Completed in 1.49s.

 building client (vite) 
14:29:24 [vite] ✓ 101 modules transformed.
14:29:24 [vite] dist/_astro/Foo.BqSHm6A9.js          0.11 kB │ gzip:  0.12 kB
14:29:24 [vite] dist/_astro/client-v5.2QKS7mx9.js    8.13 kB │ gzip:  3.66 kB
14:29:24 [vite] dist/_astro/client.5I5BMcNS.js     142.41 kB │ gzip: 45.91 kB
14:29:24 [vite] ✓ built in 1.08s

 generating static routes 
14:29:24 ▶ src/pages/index.mdx
14:29:24   └─ /index.html[TypeError] Cannot read properties of undefined (reading 'title')
  Hint:
    This issue often occurs when your MDX component encounters runtime errors.
  Stack trace:
    at copy_payload (file:///C:/Users/hugok/repositories/repro-astro-svelte/dist/chunks/_@astro-renderers_VtnIBQy8.mjs:25:16)
    at Tester (file:///C:/Users/hugok/repositories/repro-astro-svelte/dist/chunks/_@astro-renderers_VtnIBQy8.mjs:293:18)
[...]

PR #11624, opened 2024-08-05
Prevent throwing in react and solid component checks

bluwy:
## Changes

fix https://github.com/withastro/astro/issues/11615

Svelte 5 is a function component, so it messes some of our existing renderers like React and Solid where their `check()` functions assumes JSX and wouldn't throw, but Svelte 5 trips them up because it's not JSX functions.

Anyways, I think it's more correct that these check functions do not throw since they only do checking, like how Preact works now.

## Testing

No tests as we don't have them for Svelte 5, and it's a little tricky to emulate a component that would error 😬 Hopefully the current passing tests are enough though

EDIT: I have to tweak one test regarding runtime errors from Preact components, because previously it relied on the React renderer's check to throw the error if it failed to render, which I don't think is necessarily correct. For now, renderers with `check()` that simply runs the render will not be able to detect actual runtime errors to display on the error overlay.

## Docs

n/a. bug fix.


packages/integrations/preact/src/server.ts
```diff
@@ -27,19 +27,17 @@ async function check(
 	useConsoleFilter();
 
 	try {
-		try {
```
bluwy, line 30: I only refactored this part as the nested `try` isn't necessary. Before it's `try { try..catch } finally {}`. It can be flatten as `try..catch..finally` and work the same.

Princesseuh:
> For now, renderers with check() that simply runs the render will not be able to detect actual runtime errors to display on the error overlay.

How many frameworks does this affect? This seems like a pretty consequent drawback

bluwy:
They affect the react and solid integrations only. Preact already handles it correctly now. However, this only affects if you have more than one UI framework integration, e.g. if you have React and Svelte 4, you get this error with this PR:

<details>
<summary>Screenshot</summary>

image

</details>

If you only have the React integration, Astro will fallback to it and shows the error correctly (same as before this PR):

<details>
<summary>Screenshot</summary>

image


</details>

Princesseuh:
Oh okay, fine with me then!

Princesseuh (approved):

PR #11625, opened 2024-08-05
Refactor content layer into shared instance

ascorbic:
## Changes

As suggested by @bluwy, this refactors content layer into a central location. This is most useful in dev, where variosu events can trigger a re-sync. Currently this includes the content config changing, but in future will also include hot keys and integrations.


bluwy (approved):
Nice refactor!

PR #11626, opened 2024-08-05
Add hot key to reload content layer

ascorbic:
## Changes

Adds support for an "r + enter" hotkey to reload the content layer during dev. Also logs a hint at startup, showing how to display the hotkeys

image


packages/astro/src/core/messages.ts
```diff
@@ -99,6 +99,8 @@ export function serverStart({
 		...localUrlMessages,
 		...networkUrlMessages,
 		'',
+		`${dim('Press')} ${bold('h + enter')} ${dim('to')} ${bold('show hot keys')}`,
+		''
```
bluwy, line 103: I don't think we want to show this message yet as proposed by Fred before, but an alternate tip display. ([Discord discussion](https://discord.com/channels/830184174198718474/1070208455329198190/1268119257263116401))

ascorbic: Doesn't that just apply to persistent footers though? This is just logged at the same time the server URLs are. 
image

bluwy: Yeah, it was also where I added the logs in the past, but we later discussed to not want to log it here for now.

ascorbic: Aah. I will remove it from this PR (which is probably sensible anyway as it'll be experimental) and will make the case with Fred that content layer tips the balance in favour of adding it once it's stable!

bluwy (approved):
I don't know if we would expand `r` in the future to also reload other things, or restart the server, but in any case I don't think it would be a breaking change.

ematipico:
I would go for `g` instead of `r`. G would stand for "generate", in line with the sync command (which we will rename to gen)

ascorbic:
But it's not generating anything - it's reloading the data

ascorbic:
I've changed the key to "s" for "sync"

PR #11628, opened 2024-08-05
Consistent file hashes

madbook:
## Changes

Astro's css plugin generates chunk ids that include a hash of all of the chunk's parent ids.  These ids are currently the absolute file paths of the parent files.  The generated chunk ids are then inserted into those pages as import statements.

Because these import statements include a hash based on these absolute file paths, this causes rollup to generate different hashes for those pages when a build is run in different environments.  The exact same project will produce identical assets with different filenames when built on different machines, or when built from different directories on the same machine, etc.

To fix this, I've stripped out the working directory of these file paths before they are added to the hash.  This means that the hash will still change if the files referencing it chacnge (which I believe is the intended behavior), but will be stable if the entire project is built in different environments.

## Testing

I tested this by manually patching the copy of astro in my project to include this change.  I did not include any new tests here, mostly because I'm not very familiar with astro's codebase and wasn't sure where/how to best add one, but it's also a fairly small change.

## Docs

I don't think this would need any additional docs; I think that either this behavior doesn't matter to most users, or they would expect it to already work this way.

### DIFF
--- a/.changeset/young-pillows-shave.md
+++ b/.changeset/young-pillows-shave.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Generated CSS chunk ids are now stable across different environments
--- a/.changeset/young-pillows-shave.md
+++ b/.changeset/young-pillows-shave.md
@@ -2,4 +2,4 @@
 'astro': patch
 ---
 
-Generated CSS chunk ids are now stable across different environments
+Ensures consistent CSS chunk hashes across different environments
--- a/packages/astro/src/core/build/css-asset-name.ts
+++ b/packages/astro/src/core/build/css-asset-name.ts
@@ -3,6 +3,7 @@ import type { GetModuleInfo, ModuleInfo } from 'rollup';
 import crypto from 'node:crypto';
 import npath from 'node:path';
 import process from 'node:process';
+import { normalizePath } from 'vite';
 import type { AstroSettings } from '../../@types/astro.js';
 import { viteID } from '../util.js';
 import { getTopLevelPageModuleInfos } from './graph.js';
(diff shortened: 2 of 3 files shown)


packages/astro/src/core/build/css-asset-name.ts
```diff
@@ -24,8 +25,13 @@ export function shortHashedName(id: string, ctx: { getModuleInfo: GetModuleInfo
...
 	const baseName = baseId ? prettifyBaseName(npath.parse(baseId).name) : 'index';
 	const hash = crypto.createHash('sha256');
+	const cwd = process.cwd();
+
 	for (const id of hashIds) {
-		hash.update(id, 'utf-8');
+		// Strip the project directory from the paths before they are hashed, so that assets
+		// that import these css files have consistent hashes when built in different environments.
+		const relativePath = npath.relative(cwd, id);
+		hash.update(relativePath, 'utf-8');
```
bluwy, line 34: I think it'd be more accurate if it's relative from the project root instead of cwd. You should be able to get that from `settings.config.root` in `plugin-css.ts` which uses this function.

Also, the slashes could be backslashes on Windows. We could use `normalizePath()` from `vite` to make sure that the slashes are always forward slashes.

bluwy (commented):
Good catch! No worries about the test. I think as long as it doesn't break anything it should be fine.

packages/astro/src/core/build/css-asset-name.ts
```diff
@@ -13,19 +14,32 @@ const confusingBaseNames = ['404', '500'];
...
 
-export function createNameHash(baseId: string | undefined, hashIds: string[]): string {
+export function createNameHash(
+	baseId: string | undefined,
+	hashIds: string[],
+	settings: AstroSettings
+): string {
 	const baseName = baseId ? prettifyBaseName(npath.parse(baseId).name) : 'index';
 	const hash = crypto.createHash('sha256');
+	const root = settings.config.root.pathname;
```
bluwy, line 35: This should use `fileUrlToPath(settings.config.root)` to handle Windows path.

.changeset/young-pillows-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'astro': patch
+---
+
+Generated CSS chunk ids are now stable across different environments
```
bluwy, line 5:
```suggestion
Ensures consistent CSS chunk hashes across different environments
```

Minor update to follow the usual formatting we use

bluwy (approved):
Looks great!

ematipico (approved):

PR #11631, opened 2024-08-06
fix: support filters in content layer `getCollection`

ascorbic:
## Changes
Adds support for the `filter` arg in `getCollection`

## Testing

Updated fixtures to use the filter.

### DIFF
--- a/packages/astro/src/content/runtime.ts
+++ b/packages/astro/src/content/runtime.ts
@@ -71,6 +71,7 @@ export function createGetCollection({
 	cacheEntriesByCollection: Map<string, any[]>;
 }) {
 	return async function getCollection(collection: string, filter?: (entry: any) => unknown) {
+		const hasFilter = typeof filter === 'function';
 		const store = await globalDataStore.get();
 		let type: 'content' | 'data';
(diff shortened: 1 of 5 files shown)


packages/astro/src/content/runtime.ts
```diff
@@ -82,15 +83,19 @@ export function createGetCollection({
...
+					: rawEntry.data;
+				const entry = {
+					...rawEntry,
 					data,
 					collection,
-				});
+				};
+				if(hasFilter && !filter(entry)) {
+					continue;
+				}
```
ascorbic, line 97: We're looping over these already, so no need for a separate `.filter()` pass.

packages/astro/test/fixtures/content-layer/src/pages/collections.json.js
```diff
@@ -1,19 +1,20 @@
 import { getCollection, getEntry } from 'astro:content';
 import * as devalue from 'devalue';
-import { stripAllRenderFn, stripRenderFn } from '../utils';
```
ascorbic, line 3: No longer needed, as `render` function has been removed from content layer entries.

PR #11633, opened 2024-08-06
feat: handle simple mdx rendering

ematipico:
## Changes

This PR adds some infrastructure to handle compiled languages such as`mdx`. In particular, this PR addresses `mdx`, but we will explore an architecture that would allow to render Markdoc documents too.

The idea is to introduce a flag called `isModule` (better names are welcome). If an entry is treated as a module, then its rendering phase is deferred at runtime. 

The store exposes new APIs to create a temporary file called `modules.mjs` where people can create a map that will look like this:

```js
{
	"content/file.mdx": () => import("virtual-module")
}
```

This logic is implemented in our `glob` loader, which will generate a file like this:

```js
{
	"content/file.mdx": () => import('astro:content-module-imports?astro%3Acontent-module-imports=&importer=content-outside-src-mdx%2Figuana.mdx&astroContentModuleFlag=true')
}
```

In the runtime, via virtual modules, we retrieve the correct file to render and we execute the dynamic import. The vite system will transform the import, and it will return to us an Astro component, which we will map to the collection

## Testing

For now, I added a test. I plan to add more of them while I will explore a more stable way to re-use our existing architecture


packages/astro/src/content/runtime.ts
```diff
@@ -87,12 +88,36 @@ export function createGetCollection({
...
+						body: rawEntry.body,
+						data: rawEntry.data,
+						async render() {
+							return {
+								Content: resolvedComponent.Content,
+								headings: resolvedComponent.getHeadings?.() ?? [],
+								remarkPluginFrontmatter: resolvedComponent.frontmatter ?? {},
+							};
+						},
+					};
```
ascorbic, line 118: I think there's arguments for and against making this more like existing MDX (so like this one) or more like new markdown (which puts headings and frontmatter inside `rendered.metadata`).

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -72,6 +77,29 @@ export function astroContentVirtualModPlugin({
...
+				const importerParam = params.get('importer');
+				const importerPath = importerParam
+					? fileURLToPath(new URL(importerParam, settings.config.root))
+					: importer;
+				if (importerPath) {
+					const result = await this.resolve(importerPath);
+					console.log(result);
+					return result;
+				}
+			}
```
ascorbic, line 101: Am I missing something, or will this not reolve the example `astro:content-module-imports?astro%3Acontent-module-imports=&importer=content-outside-src-mdx%2Figuana.mdx&astroContentModuleFlag=true` because it will match the first condition instead

packages/astro/src/content/loaders/glob.ts
```diff
@@ -102,6 +103,10 @@ export function glob(globOptions: GlobOptions): Loader {
 				const digest = generateDigest(contents);
 
 				if (existingEntry && existingEntry.digest === digest && existingEntry.filePath) {
+					if (entryType.extensions.includes('.mdx')) {
+						store.addModuleImport(existingEntry.filePath, 'astro:content-module-imports');
```
ascorbic, line 107: Rather than using the virtual module here, could you instead use the filepath as the specifier? You're not actually checking the virtual module ID inside the Vite plugin, and are actually just using the flag to match it. You can then save `astro:content-module-imports` as just the virtual module to import the map itself.

packages/astro/src/content/data-store.ts
```diff
@@ -371,5 +468,13 @@ function dataStoreSingleton() {
 	};
 }
 
+// TODO: find a better place to put this image
+export function contentModuleToId(specifier: string, filePath: string) {
+	const params = new URLSearchParams(specifier);
+	params.set('importer', filePath);
```
ascorbic, line 474: As mentioned in the other comment: rather than using an `importer` param, could you use the filePath as the specifier? So `data/iguana.mdx?astroContentModuleFlag`. I used `improter` for images because they needed to be resolved relative to the source file, but that's not eneded here.

packages/astro/src/content/runtime.ts
```diff
@@ -87,11 +88,39 @@ export function createGetCollection({
...
+								Content: resolvedComponent.Content,
+								rendered: {
+									metadata: {
+										headings: resolvedComponent.getHeadings?.() ?? [],
+										remarkPluginFrontmatter: resolvedComponent.frontmatter ?? {},
+									}
+								}
+							};
+						},
+					};
```
ascorbic: To match the markdown behaviour, this could be moved into the `renderEntry` function. That also means the module doesn't need to be resolved unless it's actually rendered.

packages/astro/src/content/runtime.ts
```diff
@@ -87,11 +88,39 @@ export function createGetCollection({
...
+					const module = contentModules.get(rawEntry.filePath);
+					const resolvedComponent = await module();
+
+					entry = {
+						...rawEntry,
+						collection,
+						id: rawEntry.id,
+						slug: rawEntry.id,
+						body: rawEntry.body,
+						data: rawEntry.data,
```
ascorbic: You're spreading rawEntry, so don't need to repeat it here

packages/astro/src/content/vite-plugin-content-virtual-mod.ts
```diff
@@ -72,6 +76,29 @@ export function astroContentVirtualModPlugin({
...
 				return RESOLVED_DATA_STORE_VIRTUAL_ID;
 			}
+
+			if (hasContentModuleFlag(id)) {
+				const [, query] = id.split('?');
+				const params = new URLSearchParams(query);
+				const importerParam = params.get('importer');
+				const importerPath = importerParam
+					? fileURLToPath(new URL(importerParam, settings.config.root))
+					: importer;
```
ascorbic, line 86: This is resolving the wrong thing. The `importer` arg passed to `resolveId` isn't the specifier: it's the file that is doing the import.

packages/astro/test/fixtures/content-layer/src/pages/reptiles/[slug].astro
```diff
@@ -0,0 +1,25 @@
...
+import type { GetStaticPaths } from "astro";
+import { getCollection, render } from "astro:content"
+import { Image } from "astro:assets"
+
+export const getStaticPaths = (async () => {
+	const collection = await getCollection("reptiles");
+	if(!collection) return []
+	return collection.map((reptile) => ({
+		params: {
+			slug: `/${reptile.id}`
```
Comment (ascorbic) [inline packages/astro/test/fixtures/content-layer/src/pages/reptiles/[slug].astro:11]: I'm not sure this needs the leading slash

packages/astro/test/fixtures/content-layer/src/content/config.ts
```diff
@@ -86,6 +85,21 @@ const spacecraft = defineCollection({
 		}),
 });
 
+const reptiles = defineCollection({
+	type: 'experimental_content',
+	loader: glob({
+		pattern: '*.{mdx, md}',
+		base: new URL('../../content-outside-src-mdx', import.meta.url),
```
ascorbic, line 92:
```suggestion
		base: './content-outside-src-mdx'
```
Base can be relative to the site root

packages/astro/src/content/loaders/glob.ts
ematipico: I think the bug is that we don't check for `astro:content-module-imports` **and** the flag, which we should. I can test your suggestion, I am not sure if using the file path as a specifier could work on the long run, because it could trigger other rollup plugin

ascorbic: If we do use the virtual module id, then I think it should be a different one from the one used to import the actual manifest.

packages/astro/src/content/data-store.ts
```diff
@@ -371,5 +470,13 @@ function dataStoreSingleton() {
 	};
 }
 
+// TODO: find a better place to put this image
+export function contentModuleToId(fileName: string, specifier: string) {
+	const params = new URLSearchParams(specifier);
+	params.set('fileName', fileName);
+	params.set(DEFERRED_MODULE, 'true');
```
ascorbic, line 477: I think this is meant to be CONTENT_MODULE_FLAG

packages/astro/src/content/data-store.ts
```diff
@@ -371,5 +470,13 @@ function dataStoreSingleton() {
 	};
 }
 
+// TODO: find a better place to put this image
+export function contentModuleToId(fileName: string, specifier: string) {
```
ascorbic, line 474: I don't think we need to have specifier as an argument. Just hard code it to `DEFERRED_MODULE`

packages/astro/src/content/data-store.ts
```diff
@@ -33,6 +35,10 @@ export interface DataEntry<TData extends Record<string, unknown> = Record<string
 	digest?: number | string;
 	/** The rendered content of the entry, if applicable. */
 	rendered?: RenderedContent;
+	/**
+	 * If an entry is a deferred, its rendering phase is delegated to a virtual module during the runtime phase when calling `renderEntry`.
+	 */
+	isDeferred?: boolean;
```
ascorbic, line 41: How about `renderingDeferred` or `deferredRender`?

packages/astro/src/content/runtime.ts
```diff
@@ -442,6 +445,29 @@ export async function renderEntry(
...
+			const resolvedComponent = await module();
+			
+			return {
+				Content: resolvedComponent.Content,
+				rendered: {
+					metadata: {
+						headings: resolvedComponent.getHeadings?.() ?? [],
+						remarkPluginFrontmatter: resolvedComponent.frontmatter ?? {},
+					},
+				},
```
ascorbic, line 462: I don't think this needs to be this shape. The `rendered` object is a property of the top-level `entry`, so we don't need to use it here. We should probably just use the same shape as the old implementation, and maybe implement that for markdown too.

packages/astro/src/content/loaders/glob.ts
```diff
@@ -124,7 +129,10 @@ export function glob(globOptions: GlobOptions): Loader {
 					filePath,
 				});
 
-				if (entryType.getRenderFunction) {
+				if (entryType.extensions.includes('.mdx')) {
+					store.addModuleImport(relativePath);
```
ascorbic, line 133: Could this be called internally when `set` is called with `deferredRender` set to true?

ascorbic (approved):

PR #11634, opened 2024-08-06
Actions: fix 500 on empty object for Vercel serverless

### LINKED ISSUE #11515: Astro Actions .safe() function required empty object argument for Vercel deployment

bholmesdev:
## Changes

Resolves #11515. We discovered that Actions sets the `content-type` to `application/json` even when the body is `undefined`. We set `content-length` to `0` for this case, but it seems some deployment platforms (Vercel) may remove this header if it detects an empty body.

This fix removes the `content-type` when no body is present.

## Testing

- Add test for empty content type
- Add test for unexpected content type

## Docs

N/A

### DIFF
--- a/packages/astro/templates/actions.mjs
+++ b/packages/astro/templates/actions.mjs
@@ -65,15 +65,18 @@ async function handleAction(param, path, context) {
 	let body = param;
 	if (!(body instanceof FormData)) {
 		try {
-			body = param ? JSON.stringify(param) : undefined;
+			body = JSON.stringify(param);
 		} catch (e) {
 			throw new ActionError({
 				code: 'BAD_REQUEST',
 				message: `Failed to serialize request body to JSON. Full error: ${e.message}`,
 			});
 		}
-		headers.set('Content-Type', 'application/json');
-		headers.set('Content-Length', body?.length.toString() ?? '0');
+		if (body) {
+			headers.set('Content-Type', 'application/json');
+		} else {
+			headers.set('Content-Length', '0');
+		}
 	}
 	const rawResult = await fetch(`/_actions/${path}`, {
 		method: 'POST',
--- a/.changeset/tiny-poems-battle.md
+++ b/.changeset/tiny-poems-battle.md
@@ -1,0 +1,5 @@
+---
+'astro': patch
+---
+
+Fixes internal server error when calling an Astro Action without arguments on Vercel.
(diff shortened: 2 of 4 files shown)


packages/astro/templates/actions.mjs
```diff
@@ -65,15 +65,18 @@ async function handleAction(param, path, context) {
 	let body = param;
 	if (!(body instanceof FormData)) {
 		try {
-			body = param ? JSON.stringify(param) : undefined;
```
bholmesdev, line 68: Fixes edge case where `null` is considered an empty body. `null` is valid JSON!

bluwy (approved):
I think it's missing a changeset

PR #11635, opened 2024-08-06
Improve regex performance

bluwy:
## Changes

Adds the [regexp/prefer-regexp-test](https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-regexp-test.html#regexp-prefer-regexp-test) and [regexp/prefer-regexp-exec](https://ota-meshi.github.io/eslint-plugin-regexp/rules/prefer-regexp-exec.html) lint rules, which helps with perf. I've benchmarked it in the past, like in https://github.com/lukeed/polka/pull/210, where preferring `regex.exec` over `string.match` is faster.

## Testing

Existing tests should pass

## Docs

I didn't add a changeset as it's mostly a refactor, but if anyone feels like the changes are a bit big, I'm happy to add a changeset too.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -111,6 +111,8 @@ export default [
 
 			// In some cases, using explicit letter-casing is more performant than the `i` flag
 			'regexp/use-ignore-case': 'off',
+			'regexp/prefer-regexp-exec': 'warn',
+			'regexp/prefer-regexp-test': 'warn',
 		},
 	},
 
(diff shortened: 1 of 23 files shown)


ematipico:
Can you run the benchmark job to see if there's a difference?

bluwy:
!bench

bluwy:
It seems like there isn't a clear improvement overall, mostly within small fluctuations, but I think it's expected that the regex changes are only prominent at a macro level than from the top level.

ematipico (approved):

PR #11636, opened 2024-08-06
Remove `#astro/*` subpath imports

bluwy:
## Changes

The `#astro/*` subpath imports was added in https://github.com/withastro/astro/pull/161 but it doesn't look like we're using it much today. I'm not sure if we still want to continue pushing to use the subpath imports, we can discuss here if so!

If not, I figured to remove it so it doesn't interfere with the auto-imports path completion so we always stick with relative paths.

## Testing

n/a

## Docs

n/a

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -83,9 +83,6 @@
     },
     "./virtual-modules/*": "./dist/virtual-modules/*"
   },
-  "imports": {
-    "#astro/*": "./dist/*.js"
-  },
   "bin": {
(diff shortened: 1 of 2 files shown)


bholmesdev (approved):
Yes, great to see this removed! My editor always suggested the `#` import first, and it raises a typescript build error requesting we add a `rootDir` configuration.

florian-lefebvre (approved):
Awesome thanks! It bothered me quite a lot

PR #11639, opened 2024-08-06
Add file generation and flag for content intellisense

Princesseuh:
## Changes

Powers https://github.com/withastro/language-tools/pull/915

This PR adds a new experimental flag (`contentCollectionIntellisense`) that when enabled, make Astro generates two new things: a collection manifest (`.astro/collections/collections.json`) and JSON schemas for every content collection.

The manifest is then used by the language server to map from file -> collection, which then use that to associate JSON schemas to the appropriate files and generate a virtual typescript file for every entry.

All of that powers the _**Content Collections Intellisense**_:tm: story.

## Testing

Added tests!

## Docs

Part of the flag JSdoc


packages/astro/src/content/types-generator.ts
```diff
@@ -391,6 +392,8 @@ async function writeContentFiles({
 			entries: {},
 		};
 	}
+
+	let contentCollectionsMap: CollectionEntryMap = {};
```
Princesseuh, line 396: There's nothing wrong for language tooling to also include data collections in the manifest (it just won't take them into account), but it caused crashes here because data collections are defined differently

Princesseuh:
!preview content-collections-intellisense

Princesseuh:
!preview content-collections-intellisense

Princesseuh:
This is now on top of Content Layer, it doesn't quite work yet because in Content Layer, `collection.entries` is always empty, but at least the JSON schemas work. Should be figurable before minor time cc @ascorbic

packages/astro/src/@types/astro.ts
```diff
@@ -2186,6 +2186,30 @@ export interface AstroUserConfig {
...
+		 * @docs
+		 * @name experimental.contentCollectionIntellisense
+		 * @type {boolean}
+		 * @default `false`
+		 * @version 4.14.0
+		 * @description
+		 *
+		 * Enables the generation of files required for content intellisense for Astro content files.
+		 *
+		 * When enabled, this feature will add JSON schemas to the `.astro` directory in your project, which can be used by the Astro language server to provide intellisense inside content files (`.md`, `.mdx`, `.mdoc`).
```
sarah11918, line 2199:
```suggestion
		 * Enables Intellisense features (e.g. code completion, quick hints) for your content collection entries in compatible editors.
		 *
		 * When enabled, this feature will generate and add JSON schemas to the `.astro` directory in your project. These files can be used by the Astro language server to provide Intellisense inside content files (`.md`, `.mdx`, `.mdoc`).
```
Intellisense seems to be capitalized as a proper noun. 

Just a thought: also thinking we could choose to highlight what the feature unlocks, vs the fact that files are generated. (I made up the "compatible editors" because now we're claiming not just the file generation, but actually promising Intellisense. If this isn't the right way to frame it, could be something like "in editors that support the Astro language server.")

packages/astro/src/@types/astro.ts
```diff
@@ -2186,6 +2186,30 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 *   experimental: {
+		 *     contentIntellisense: true,
+		 *   },
+		 * }
+		 * ```
+		 *
+		 * To use this feature with the Astro VS Code extension, make sure to also enable the `astro.content-intellisense` option in your VS Code settings. For editor using the Astro language server directly, pass the `contentIntellisense: true` initialization parameter to enable this feature.
```
sarah11918, line 2209:
```suggestion
		 * To use this feature with the Astro VS Code extension, you must also enable the `astro.content-intellisense` option in your VS Code settings. For editors using the Astro language server directly, pass the `contentIntellisense: true` initialization parameter to enable this feature.
```

sarah11918 (commented):
JSDoc looking good! Made a suggestion in case you like it!

The changeset can be as close to this as you want, even copied!

Princesseuh:
Added some tests to the PR, they won't pass until Content Layer supports this

sarah11918 (approved):
Approving for docs! Fabulous job, Erika!

.changeset/serious-pumas-run.md
```diff
@@ -0,0 +1,19 @@
...
+export default defineConfig({
+	experimental: {
+		contentIntellisense: true
+	}
+})
+```
+
+When enabled, this feature will generate and add JSON schemas to the `.astro` directory in your project. These files can be used by the Astro language server to provide Intellisense inside content files (`.md`, `.mdx`, `.mdoc`).
+
+Note that at this time, this also require enabling the `astro.content-intellisense` option in your editor, or passing the `contentIntellisense: true` initialization parameter to the Astro language server for editors using it directly.
```
sarah11918, line 19:
```suggestion
Note that at this time, this also require enabling the `astro.content-intellisense` option in your editor, or passing the `contentIntellisense: true` initialization parameter to the Astro language server for editors using it directly.

See the [experimental content Intellisense docs](https://docs.astro.build/en/reference/configuration-reference/#experimentalcontentintellisense) for more information updates as this feature develops.
```
I know there really isn't any *more* information there right now. :sweat_smile:  But in theory, any developments or changes would go there!

packages/astro/src/@types/astro.ts
```diff
@@ -2186,6 +2186,30 @@ export interface AstroUserConfig {
...
+		 *
+		 * ```js
+		 * {
+		 *   experimental: {
+		 *     contentIntellisense: true,
+		 *   },
+		 * }
+		 * ```
+		 *
+		 * To use this feature with the Astro VS Code extension, you must also enable the `astro.content-intellisense` option in your VS Code settings. For editors using the Astro language server directly, pass the `contentIntellisense: true` initialization parameter to enable this feature.
```
sarah11918, line 2209: If there *were* a link to a discussion or RFC where you want people to leave feedback, or where there's more helpful information, we'd typically include it at the end here. If not, then no worries! We don't have this for some of our longer-standing experimental flags.

packages/astro/test/fixtures/content-layer/src/content/config.ts
```diff
@@ -99,11 +99,13 @@ const increment = defineCollection({
 				},
 			});
 		},
+		// Example of a loader that returns an async schema function
+		schema: async () => z.object({
+			lastValue: z.number(),
+			lastUpdated: z.date(),
+		}),
```
Princesseuh, line 106: Did you mean to change this (and the other test changes). We can test this isolated in the content-intellisense fixture instead

ascorbic: Yeah, I was testing it with this fixture. I'm fine with leaving these in though, as they catch more cases in the content-layer tests too.

packages/astro/test/content-intellisense.test.js
```diff
@@ -0,0 +1,80 @@
...
+	let fixture;
+
+	/** @type {string[]} */
+	let collectionsDir = [];
+
+	/** @type {{collections: {hasSchema: boolean, name: string}[], entries: Record<string, string>}} */
+	let manifest = undefined;
+
+	before(async () => {
+		fixture = await loadFixture({ root: './fixtures/content-Intellisense/' });
```
Princesseuh, line 16:
```suggestion
		fixture = await loadFixture({ root: './fixtures/content-intellisense/' });
```

packages/astro/src/@types/astro.ts
Princesseuh: I don't expect it to be flagged for too long (it might get unflagged before 5.x), so I'd rather people report issues directly if there's bugs vs a discussion, I think

packages/astro/src/@types/astro.ts
```diff
@@ -2186,6 +2186,30 @@ export interface AstroUserConfig {
 		 */
 		serverIslands?: boolean;
 
+		/**
+		 * @docs
+		 * @name experimental.contentCollectionIntellisense
```
Princesseuh, line 2191:
```suggestion
		 * @name experimental.contentIntellisense
```

packages/astro/src/content/content-layer.ts
```diff
@@ -230,6 +230,42 @@ export class ContentLayer {
...
+				const collections = await fs.readFile(collectionsManifest, 'utf-8');
+				const collectionsJson = JSON.parse(collections);
+				collectionsJson.entries ??= {};
+
+				for (const { hasSchema, name } of collectionsJson.collections) {
+					if (!hasSchema) {
+						continue;
+					}
+					const entries = this.#store.values(name);
+					if (!entries?.[0]?.filePath || entries[0].filePath.endsWith('.json')) {
```
Princesseuh, line 252: What does this `endsWith('.json')` condition prevents?

matthewp (approved):

packages/astro/src/content/content-layer.ts
ascorbic: The intellisense doesn't support JSON does it?

Princesseuh: You're correct (though there's no problems with them being in the manifest, tooling just ignore them), but I figure a whitelist would potentially be better here? Not sure if it's still the case in Content Layer, but YAML is supported inside data collections, for instance

ascorbic: OK, I've removed it

bluwy (approved):

PR #11640, opened 2024-08-06
update formatter config

itsMapleLeaf:
## Changes

Makes the following formatter changes:

- `trailingComma: "all"`

  Rationale:
  - Less tedious line reordering
  - The default was changed to `all` in prettier v3
  - [Environment support is very green](https://caniuse.com/mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions) (Astro potentially uses other features that are _less_ broadly supported than this)

<details>
<summary>Reverted changes</summary>

- `printWidth: 80`
  
  Rationale:
  - More readable on narrow viewports, e.g. when putting two editors side-by-side
  - [Shorter lines are more readable, generally](https://prettier.io/docs/en/options.html#print-width)
  - Can encourage more readable sequential code: seeing an awkwardly-formatted statement at 80 lines might make you feel more inclined to break it up. No hard evidence on this, but this has been my anecdotal observation with prettier in general.

- `singleQuote: false`

  Rationale:
  - [Double quotes result in less escapes than single quotes](https://prettier.io/docs/en/rationale#strings)

</details>

## Testing

Tests passed after running format. I got a local test failure at `packages\markdown\remark\test\shiki.test.js`, but it doesn't seem related.

## Docs

No docs needed, internal change.


.git-blame-ignore-revs
```diff
@@ -6,3 +6,5 @@
 59e8c71786fd1c154904b3fefa7d26d88f4d92d2
 # Change formatting (#10180)
 062623438b5dfd66682a967edc7b7c91bd29e888
+# Prettier config update
```
Princesseuh, line 9: This won't work because of the squash, you have to do it in two steps.

Princesseuh:
I'll admit that I, very subjectively of course, don't like the 80 line lengths. I use a condensed font and so it kinda result in 70% of my screen just being empty now

Going through the changes and testing 80 in the language-tools, I also (subjectively) feel like some code now takes a bigger cognitive effort to parse due to being (somewhat unnecessarily) spread over multiple lines

I could probably be convinced, just not feeling it on first look

Everything else is fine, don't care

FredKSchott:
> `trailingComma: all` 
Big fan of making refactoring code easier, where reordering items in an array/object/function definition leads to fewer syntax errors around missing commas where the entire file breaks!

> [Double quotes result in less escapes than single quotes](https://prettier.io/docs/en/rationale#strings)

I don't see this stated in the link. The link only states that prettier will choose the option that will result in the fewest escapes for that specific string, but will blindly defaulting to whatever `singleQuote` is set to. Can you confirm where this came from? That's a really interesting fact if true! I don't care too much about this one personally but loosely -1 if this is just entirely subjective.

> `printWidth: 80`

Torn on this one. I personally always code with 100 columns visible and prefer the extra space, having used both `80` and `100` in different projects in the past, so this is personally a change I won't enjoy. But if others code with two columns visible or zoomed to a level where >80 columns aren't visible on their screen, I think you can make an objective accessibility argument to drop down to 80.

FredKSchott:
For context: we were discussing these changes originally in another Astro repo, but I set the hard hard rule that all Astro repos should use the same formatter config, copy-pasting from the `astro` monorepo to make it easy for team members to contribute and feel familiar across repos. So the result of this PR will also impact the prettier config we use it some/most/all other Astro repos as well.

Thanks for taking the time to kick this off in the main repo, Darius!

bluwy:
- trailingComma: in favor
- printWidth: against. Having worked with 80 (vite) and 100 (astro). I actually quite like 100 as it doesn't break up the code as often. I have more opportunity to break up code semantically than stylistically with a larger width.
- singleQuote: against. Agree with Fred that "if it ain't broke don't fix it"

ematipico:
- trailing commas: yes
- print width: neutral 
- quotes: neutral

ascorbic:
- trailingComma: neutral
- printWidth: against
- quotes: against

I'm neutral on trailingComma because I think there are decent arguments in favour of it that balance my strong bias against any changes to the config.

I'm against the other two because there are arguments both for and against, and we should only be changing these if there's a strong reason for doing it.

.git-blame-ignore-revs
itsMapleLeaf: I'll convert the team away from squash merging one of these days 😅

itsMapleLeaf:
Looks like we're not strongly against trailingComma - kept.
I don't feel as strongly about printWidth than the others here - reverted.

I'll elaborate more on double quotes: anecdotally, I frequently find myself needing to change from single to double quotes due to adding an apostrophe. I'll admit I misinterpreted what Prettier was saying there 😅 thinking that it was basically saying this.

Still, reverted quotes too. I'll bet that's a me thing (even though I already count several apostrophes in this comment alone 😂). `trailingComma` is arguably the only important one here.

ascorbic:
I'm sorry if my comments came across as dismissive. I'm not suggesting that your suggestions are arbitrary. The point I was trying to make is that even when there are objective arguments for the different options, few are as cut and dried as to avoid having arguments on both sides. My main objection is that these sort of things generate disagreements that are way out proportion to the actual impact of a change, which is why I am so biased against changing any option that's already in use.

itsMapleLeaf:
> I'm sorry if my comments came across as dismissive. I'm not suggesting that your suggestions are arbitrary. The point I was trying to make is that even when there are objective arguments for the different options, few are as cut and dried as to avoid having arguments on both sides. My main objection is that these sort of things generate disagreements that are way out proportion to the actual impact of a change, which is why I am so biased against changing any option that's already in use.

I agree, and no worries. I only initially included the `printWidth` and quote changes _because_ I had objective arguments for them, even if anecdotal, and I was ready to pull back if there was strong pushback nonetheless.

And, not specifically to you, but generally, I want to make clear that my motivation for `trailingComma`, aside from reducing microfriction, is modernizing the project. There's just no good reason to use es5. In retrospect, I think that would have made sense to put in a PR separate from other opinionated changes.

ematipico (approved):

Princesseuh (approved):
It's crazy that we're in 2024 and opening the "files" tab on this PR crashed my browser twice, what are we doing with our life really, is the web really the best platform to build on?

Princesseuh:
I'll take care of adding this to other repos, thank you for kicking the discussion! Apologies if any comments came as dismissive and/or rude.

